PR #2657, opened 2017-06-21
enhancement: Logbook selection wraps around

tehhowch:
Pressing Up while at the top entry will select the bottommost entry.
Pressing Down while at the bottom entry will select the topmost entry.

Right now, the logbook does not wrap its scroll. This seems weird, considering every other interface with selectable elements does wrap. This PR adds scrolling wraparound to the logbook.

### DIFF
--- a/source/LogbookPanel.cpp
+++ b/source/LogbookPanel.cpp
@@ -183,7 +183,7 @@ bool LogbookPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command)
 		{
 			++i;
 			if(i >= contents.size())
-				return true;
+				i = 0;
 		}
 		else if(i)
 		{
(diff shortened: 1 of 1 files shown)


endless-sky:
The difference is that the logbook has a linear order (i.e. you're reading a story from start to finish) rather than a random one (searching through a list for a particular outfit or ship). The conversation panel doesn't loop around either.

tehhowch:
I think it's a pretty common operation to - when reading a historical record - be able to jump from beginning to end, or end to beginning. We can do this with the mouse, so why not the keyboard? As it is not "actionable" like the conversation panel, there is no possible harm in over-scrolling.

jafdy:
Where dose the selection start?
If it starts at the top then you wan't a quick way to get to the end and if it starts at the end the then you want a quick way to get back to the top.
And as @tehhowch says there is no harm in having the feature.

Issue #2663, opened 2017-06-22
Visiting Danforth on Farpoint text doesn't change based on checkmate branch


jlnxr:
Played through the checkmate branch recently, was surprised when Danforth says this:
![screenshot from 2017-06-21 22-34-36](https://user-images.githubusercontent.com/20472439/27413563-8c6e0d9a-56d2-11e7-8a4d-6f703f8df5ab.png)
![screenshot from 2017-06-21 22-34-27](https://user-images.githubusercontent.com/20472439/27413564-8c6ede8c-56d2-11e7-894a-1fe6cb557312.png)

Danforth shouldn't know about the Syndicate's involvement with nukes if you've taken the checkmate line, and the player only has a vague idea/hints about their exact involvement based on what the defector said he mad before they turned him over.

MessyMix:
#2665 good catch.

Amazinite:
On the topic of epilogue, it might be worth giving Tomek an epilogue mission for Checkmate players (and maybe even for Reconciliation players so we can know that he isn't under perpetual house arrest).

jlnxr:
I agree with that- seems odd that the situation that leads the Senate to reinstate him- the battle over Thule, and the escalation of the war, albeit against a different enemy- happens in both story lines and yet in the reconciliation one he is just left rotting away in house arrest as far as the player is concerned.

Amazinite:
Finished in 8123622 by having Danforth not mention the Syndicate at the beginning of the epilogue, and near the end instead of talking about Raven finding the extremists, he talks about the nukes that the Free Worlds still have.

PR #2665, opened 2017-06-22
Branched Danforth Epilogue for Reconciliation Branch (@jlnxr #2663)

MessyMix:
#2663 Danforth's epilogue conversation had a paragraph that suggested the Navy had been and was still actively monitoring Syndicate activity regarding nuclear activity. I changed it thanks to @jlnxr such that Danforth would only say this when the player takes the reconciliation route.

### DIFF
--- a/data/free worlds epilogue.txt
+++ b/data/free worlds epilogue.txt
@@ -173,6 +173,11 @@ mission "FW Epilogue: Danforth"
 			label end
 			`	"Are the pirates giving you any trouble?" you ask.`
 			`	"A bit," he says. "We still have far too many pirates in this sector, in part due to the Free Worlds driving them out of the south. The pirates grew stronger and more bold as the war dragged on, when all our fleets were busy elsewhere, but now we're beginning to get them under control again."`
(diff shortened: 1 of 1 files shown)


jafdy:
```
branch checkmate
   not "FW Defend New Tibet: done"
`<reconciliation text>`
label checkmate
`<end text>`
```
I can't put the tabs in here but this would save one line.

endless-sky:
He also mentions the Syndicate in the first paragraph.

I wonder if it'd make more sense to just have a completely different mission depending on what branch you took, because the player really has much less of a relationship with Danforth if they took the checkmate branch. (I'd suggest removing this mission entirely if the player took the checkmate branch, except that Danforth becomes important in the Wanderer plot so it's worth noting where he's stationed after the war ends even if you aren't friends with him.)

MessyMix:
@endless-sky I would be happy to help with making a new one. Do you think the player should be talking to Danforth? iirc there's already a log entry that makes clear that Danforth is on Farpoint, but the player could also be reminded in the epilogue..

tehhowch:
@MessyMix special logs are supported, but none exist yet in vanilla. The Danforth log you saw was a teaser from MZ

MessyMix:
Yes, but as I said the player could be reminded in the epilogue about Danforth's placement at Farpoint, but whether or not this is necessary is really up to @endless-sky.

MessyMix:
Mm, I'm just going to go ahead and re-write it anyways. @endless-sky stop me if this is unnecessary. :)

PR #2644, opened 2017-06-18
Small change in the intro conversation

Amazinite:
Just noticed this. Then it started to irk me.

PR #2672, opened 2017-06-24
Prettify more credits displays with Format::Number

TurkeyMcMac:
I have added commas to the displays of planet tribute, player salary, fines, mortgage payments, and, crew salaries.

Note: I have zero experience with C++, so sorry for any mistakes.

### DIFF
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -15,6 +15,8 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "DataNode.h"
 #include "DataWriter.h"
 
+#include "Format.h"
+
 #include <algorithm>
 #include <sstream>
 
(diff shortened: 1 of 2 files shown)


TurkeyMcMac:
I am sorry if I am not making myself clear. What I mean is, "you pay 100000 credits in crew salaries" will instead be written as "you pay 100,000 credits in crew salaries."

Bladewood:
but, https://github.com/endless-sky/endless-sky/pull/2644!

Bladewood:
relevant: https://github.com/endless-sky/endless-sky/pull/2644

source/Account.cpp
```diff
@@ -15,6 +15,8 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "DataNode.h"
 #include "DataWriter.h"
 
+#include "Format.h"
```
tehhowch: Should be grouped with the existing header includes

source/Account.cpp
```diff
@@ -211,19 +213,22 @@ string Account::Step(int64_t assets, int64_t salaries)
 	
 	// If you made payments of all three types, the punctuation needs to
 	// include commas, so just handle that separately here.
+	string salariesPaidDisplay = Format::Number(salariesPaid),
```
tehhowch: Rather than define new variables you can simply wrap the numbers when they are being printed. While not expensive to do, this means only the numbers that need to be formatted (e.g. non-zero numbers) will be formatted.

e.g.
`out << Format::Number(salariesPaid) << ....`

source/Account.cpp
TurkeyMcMac: Changed accordingly.

source/Account.cpp
TurkeyMcMac: Changed accordingly.

Issue #2607, opened 2017-06-08
Outfitter enhancement: Change ship color outline based on installed / not installed status of selected outfit


tehhowch:
If you select multiple ships, you will see various for-sale outfits switch to "installed: 0 - N".
I propose that when you select such an outfit - where not all ships have it installed - the ships that do have it installed obtain a differently colored outline. A first choice: green

I could also understand always doing this behavior (e.g. even if all ships have it, all outlines become \<color\>) to avoid the "same action different results" possibility of confusion.

This condition-dependent coloring could be extended to highlight ships in red if they do not have it installed and also do not have the space to install it.

Current game behavior:
![image](https://user-images.githubusercontent.com/20871346/26944629-e1901ed2-4c4e-11e7-9473-2f32b567a25a.png)

Proposition:
![screenshot from 2017-06-08 13-31-49](https://user-images.githubusercontent.com/20871346/26945315-f9d1de98-4c50-11e7-99fc-752797b588d0.png)

KiLEdEnNis:
what if some ships have none of the outfits, some have 1 or 2 and some have like 6? although that's a possible "next step", this shows which ships have the outfit which is good

tehhowch:
Using color to indicate how many is certainly possible (see commodities on map), but more convoluted. Possoble 3 color schemes: 
1) green = has at least 1 installed , white = zero installed, and red = no room / can't install.
2) green = zero and can install, red = zero and can't install, and white = has at least 1 installed

Elyssaen:
Prefer your first scheme, as it keeps the white default in the default case. But definitely a nice idea.

endless-sky:
In commit cf9ff67, I've got what I think is a more intuitive and less intrusive approach to this: when you hover over the "buy" or "sell" button, highlighting which ships the buy or sell will apply to. Because the highlight is in response to hovering over the button, its meaning is immediately clear (unlike the icons proposed here), and it provides an easily discoverable way to check which ships have a given outfit.

tehhowch:
Awesome

PR #2675, opened 2017-06-26
Convey per-ship install status when multi-selecting ships in the Outfitter

tehhowch:
Ref issue #2607 

Summary (as of ea10da4):
Opt-in preference: When not active, current drawing scheme is used. If active, when multiple ships are selected in the Outfitter:
 - regular brightness checkbox - these ships cannot install the selected outfit
 - very bright checkbox - these ships can install the selected outfit
 - checked box - this ship has at least one of the selected outfit installed
 - unchecked box - this ship has none of the selected outfit installed

image
Both the Carrier and Osprey have the selected outfit installed. The Carrier is at maximum capacity, while the Osprey can install at least one more. The Finch is at its maximum installed number (currently 0).

The Preferences Panel, as of ea10da4:
image


jafdy:
So you need to consider people who are color blind. I imagine that this will be a bit useless for people with red/green color blindness.

tehhowch:
I have (quite mild) partial red-green myself, and chose to use the absolutes rather than mixed colors. The game already uses fixed red (.5, .15, 0, .25) and green (0, .5, .0, .25) to convey friend or foe information in the status overlays.

[Here is the above image after simulating the worst possible extent of the most common red-green colorblindness](http://www.vischeck.com/vischeck/vischeckImage.php?&img1=uploads%2Fvc_Vuc4r8_orig.jpg&img2=uploads%2Fvc_Vuc4r8_sim.jpg&simType=deuteranope&imageCached=vc_Vuc4r8_orig.jpg&imEngIP=ec2-46-51-141-72.eu-west-1.compute.amazonaws.com) 

I can see the appeal of allowing user-selected colors and will do so in the next day.

tehhowch:
@jafdy this is now a toggled feature with user-definable colors (no color wheel so the color picking must be done by editing data/interfaces.txt, just like the other game color pickings, except status overlays).
I also added a touch of blue to the default "can't install" color.

New Preferences->Settings view:
pr

I struggled a bit to come up with a suitably descriptive yet succinct phrase, so any improvements are welcomed.

Bladewood:
The O in "Outfitter" should be lower-case.

Pointedstick:
Honestly I don't think you even need to clutter up the settings with an option to turn this off. Make it the default! Good stuff.

Bladewood:
It can be the default, but we've got options for plenty other minor things.

tehhowch:
I have my doubts about referring to it in lowercase from the context of game menus and settings, but you are correct that game speech text refers to it as "outfitter" and not "Outfitter".

Bladewood:
Compare with the other preferences: the first word is capitalized, while all others (excepting abbreviations like HUD) are lower-case.

tehhowch:
And none of the other preferences refer to a game menu by name, so they are not directly relevant.

It would be good to have input from MZ on this style convention (and this PR in general).

endless-sky:
I try to use color really sparingly in the UI, especially bright colors like red and green. In my opinion, having all the ships in the sidebar light up different colors depending on what outfits can be installed looks rather tacky.

tehhowch:
Red and green were chosen due to their ubiquitous use as "not allowed / allowed" visual signals, as I felt this would make understanding the information very intuitive. I am obviously biased towards this pr's method of conveying the currently-missing information of "which of the selected ships have the selected outfit, and which will install it if I click buy," but am happy to change to a more muted/less distracting color scheme (suggestions welcome), or to refactor to use an alternate indicator method.
One such method is to draw checkmarked, empty boxes, and x'd out / fuzzy boxes in the corners of the ship outlines for the has-can have-can't have status. This has the downside (or upside) of hinting that these boxes could be interactive, e.g. "click the empty box to buy the outfit only for this ship," or "click the checkmarked box to sell this outfit from this ship."

marphod:
How about a small box with a count, color coded based on install-ability?

A red 0 for none & can't install, single or double digit number of installed count, in white if more can be installed or green if at capacity, and 'Lots' color coded if 100+. 


If you want to avoid color, use a number in white if room, number in grey if at capacity.

endless-sky:
I'm still not too fond of the Christmas tree look, but if you want to brainstorm alternatives that wouldn't result in such big swatches of bright color it's possible this could be useful. (I also feel like just the colors without any explanation would be confusing to new players.)

tehhowch:
I'll swap to WIP for the time being, as I haven't tried out any of the alternatives mentioned above, and the checkbox / number route may be more promising and intuitive.
Plus the christmas trees may not play too well with the new warning/error icons, and a graphics update to Outfitter Panel needs to play well with those. :smile:

tehhowch:
@marphod numbers aren't going to play well without additional font sizes -  the only available sizes are 14 and 18 (e.g. the same size as the rest of the text on the Outfitter screen). The checkbox elements are only 12x12 px from edge to edge (20x20 for the whole checkbox sprite), so putting a 14pt digit inside this is not exactly pretty:
image

The simple fix (to still show the 4 states of install status) would be to use OutlineShader to change the outline of the existing checkbox sprites, but as additive sprites (`ui/checked+.png` indicates this) they don't actually outline. There is the simple "repeated-draw" method with SpriteShader, but this feels a bit kludgy (but does mean no new assets are needed):
image
In the above image, the checkbox is drawn 3 times if the selected outfit can be installed, or just once if it cannot The Carrier and Osprey are able to install the outfit, the Osprey has the outfit, and the fighters cannot install it and do not have it. higher draw counts are possible, to better distinguish the status, but can make the checkbox seem pixelated if the value is too large.

tehhowch:
@endless-sky the Christmas tree is gone, just in time for the holidays :grinning: :christmas_tree: 
The extra information conveyed with the sprite change should help convey the meanings without a tooltip.

jafdy:
I might be easier to just have a new sprite.

endless-sky:
To me, a checkbox that is not a clickable UI element seems confusing. Can you think of any other icon options?

tehhowch:
I too was hesitant about using a checkbox (see my 2 July comment above). In addition to the interactivity, checkboxes are also good at conveying have / do not have in comparisons.
I also think they are easier to understand than using something like a filled, unfilled, or partially-filled circle, or a set of triangles and a stubby rectangle.

E.g. a filled circle generally means "has" - here it would mean "has maximum". An unfilled circle means "does not have", and a partially filled circle means "has, but not maximum." Repeated draws could be used to convey the remaining number of the outfit that can be installed, e.g. the brighter the icon the more can be installed.
Alternately, depending on how much of the circle is "dedicated sprite" vs "shader magic," it could go from unfilled to filled as outfits are installed. A ship unable to install even 1 would have a dim, unfilled circle, while a ship that can install, or has at least 1, would have a medium or bright circle in varied degrees of fill.

endless-sky:
I'm just trying to brainstorm here, but maybe a check or X without a box would work better.

tehhowch:
Brainstorm away - I'm not (and won't ever claim to be) great with UI/UX, so any and all input is welcomed :grin: 

With no unchecked-but-lit box, would an unfilled checkmark be appropriately intuitive for "when you click Buy, this ship is going to install the outfit?" And to indicate "can't install any more", a dim filled checkmark?
My initial thought is the x sprite should only be used to show that a ship can't install the outfit at all, but it could be shown alongside / below the checkmark for when at maximum installed capacity.

KiLEdEnNis:
what if selected ships that don't have the outfit installed do have the brighter background but the ship outline is darker than ships that have it installed?

tehhowch:
@KiLEdEnNis care to show a mockup? Would that be in conjunction with some icon in the lower left corner? Or entirely replacing the use of an icon?

KiLEdEnNis:
basically: left one has them installed, middle one doesn't but is also selected, right one is just unselected
temp8

Amazinite:
I think that highlighting the border around the square that contains the ship outline instead of highlighting the ship outline would be more obvious.

KiLEdEnNis:
I thought of something like that; it would be more obvious but I think it wouldn't look good visually

endless-sky:
I think it might be worth taking a step back and asking, in what situations would this actually be useful? The only one I can think of would be if you want all your ships to have a certain outfit and you're trying to find the one that doesn't. But if they're all the same ship type, an alternative solution is to just select them all and sell, then buy, the outfit you want them all to have.

In general, to avoid overwhelming the player with information, I don't want to always provide information that is almost never going to be useful.

endless-sky:
In commit cf9ff67, I've got what I think is a more intuitive and less intrusive approach to this: when you hover over the "buy" or "sell" button, highlighting which ships the buy or sell will apply to. Because the highlight is in response to hovering over the button, its meaning is immediately clear (unlike the icons proposed here), and it provides an easily discoverable way to check which ships have a given outfit.

PR #2677, opened 2017-06-26
enhancement: Use user-defined colors for status overlays and escort attributes

tehhowch:
Status overlays (shields, hull, scanning progress) were previously fixed
 - New colors are added to data/interfaces.txt to allow custom colors

Escort attributes (shields, hull, fuel, heat, energy) were previously hard-coded to the default values assigned to the flagship.
 - Escort colors now mimic the custom values given to the user's flagship's attributes.
 - The existing "overlapping bar" behavior for displaying attributes that differ between like-model escorts is preserved.

These changes to user-colors allow players (and mods) to define more visually accessible color schemes as needed.

Acknowledgements to @jafdy for the PR request.

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -42,6 +42,13 @@ color "escort not ready" .9 .8 0. 1.
 color "escort blocked" .9 .2 0. 1.
 color "escort selected" .2 .8 0. 1.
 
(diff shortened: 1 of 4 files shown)


jafdy:
I'll add setting up full color-blindness mods on my to-do list once this gets merged.

PR #2678, opened 2017-06-27
feat(mechanics): Issue out-of-system travel orders to escorts via the galaxy map

tehhowch:
This PR extends the current fleet movement orders to enable (via an opt-in preference setting) sending escorts to alternate star systems.
Only those star systems which have been visited are valid targets of a fleet move command, and no fleet commands will be issued if the player is landed on a planet (e.g. can add/remove ships from play).

 - Right-clicking within the planets / orbits scene issues a MOVE_TO order
   - the position is highlighted in the orbits scene: 
image

 - The currently selected escorts will go to the relative position within the system.
 - A message is shown in the ticker when the order is issued (after the player closes the map)
 - If the escorts cannot route to the system, they will ignore the order until a route is available (current behavior)

The player can cancel the order just like they can cancel other right-click move orders.


MessyMix:
Holy crap. If this works, nice! :)

tehhowch:
Of course it works :) I test all my PRs extensively

endless-sky:
I'm a little worried that the confusion caused by accidentally telling your escorts to go to some random system (because you right clicked when you meant to left click) will more than outweigh the usefulness for the few people who would want this.

endless-sky:
(I also feel like if the system map is going to be changed to allow it to be used to issue escort commands, it might also be worth changing it to show any of your escorts currently in that system. But again, I'm not sure if the added complexity would be worth it.)

tehhowch:
Preference-guarding (defaulted off) this is probably a good idea, and would prevent that accidental order.

Regarding updates to the design of the system map: it could indeed by worthwhile, and as you mention, would definitely add both complexity and usefulness.
I'm thinking maybe a clickzone underneath the system map which defaults to "You have no escorts in this system", and clicking this zone colors the galactic map by a new commodity value, indicating which systems you have escorts in. Selecting a system that has escorts within it changes the clickzone text to read "You have N escort(s) within this system," and also draws the position of those escorts (with some appropriately-colored marker). A graphics update should also probably include visual feedback of any about-to-issue fleet movement order (in the form of some other appropriate marker), and a button to cancel that pending order.
I can work on that as a separate PR, if this merges.

tehhowch:
I'm addressing https://github.com/endless-sky/endless-sky/pull/2678#issuecomment-312155956 in part with #3070.

Unaddressed in 3070 is an indication of the MoveTo order location, and a way to clear the order. Those will have to be added here since 3070 does not have the features of this PR.

tehhowch:
Marking as WIP as it still needs a way to display and / or cancel the issued order

petervdmeer:
Commenting on a quite old pull request, but this is a feature that I (and I expect many other players) really would like to have.

> Marking as WIP as it still needs a way to display and / or cancel the issued order

#3070 will indeed help in displaying the locations of escorts that use this feature.

But, don't we already have an interface to cancel the order? The current game allows me to halt ships at a fixed location in the current system (by issuing a move using the mouse, or by giving an explicit halt command using the keyboard), then allows me to jump to another system and if I then want to have the ships fly to me in this another system, then I can select them and give the halt-command again through the keyboard to have those ships return to me in the another system.

I would expect ships that forward-move to a different location to also end-up in the same state as the ships that I left behind in the system I visited before (halt at a fixed location in the system they moved to). If that is the case, then selecting them and giving the halt-command again should bring them back.

kikotheexile:
This PR has been suggested as a needed feature in the Endless Sky Stories fork. With your permission, may I use this? I can either grab the change myself and commit it for testing, or you can PR it yourself, but I will not take this without your permission, either way you would be credited.

tehhowch:
> This PR has been suggested as a needed feature in the Endless Sky Stories fork. With your permission, may I use this? I can either grab the change myself and commit it for testing, or you can PR it yourself, but I will not take this without your permission, either way you would be credited.

Probably it needs a good reevaluation before merging anywhere. I'll look at updating it this weekend.

kikotheexile:
Ill give it a look over and ask more code inclined people who are working on ESS to check it out, too. This is a nice feature, and i would really like to see it implemented somewhere.

source/PreferencesPanel.cpp
```diff
@@ -448,6 +448,7 @@ void PreferencesPanel::DrawSettings()
 		"Rehire extra crew when lost",
 		SCROLL_SPEED,
 		"Show escort systems on map",
+		"System map sends move orders",
```
petervdmeer, line 451: I don't really see a need to hide this feature behind a preference. Right-clicking on the orbits panel is a very explicit action which I wouldn't expect to happen by accident. (But I would expect support questions from players that have the preference disabled and wonder why this feature doesn't work.)

tehhowch: The default game behavior is that clicking on "radar"-like displays doesn't do anything. Instead of a new preference, we could infer from that preference whether this functionality is enabled.

petervdmeer: Okay, I just re-read the earlier discussions from 2017. Let's hide this behind a preference indeed.

source/AI.cpp
```diff
@@ -1256,7 +1258,12 @@ bool AI::FollowOrders(Ship &ship, Command &command) const
 		// way to reach that system (via wormhole or jumping). This may
 		// result in the ship landing to refuel.
 		SelectRoute(ship, it->second.targetSystem);
-		return false;
+		
+		// Travel there even if your parent is not planning to travel.
+		if(ship.GetTargetSystem())
+			MoveIndependent(ship, command);
```
petervdmeer, line 1264: I just did some testing (in Shaula) and the ships I order to move appear to prioritize finishing fighting enemies in the current system before they start moving to the desired destination in the system that I ordered them to go to.

If I give an order to a ship to move somewhere, then I would expect the move to be executed immediately, not after all enemies are gone.

source/MapDetailPanel.h
```diff
@@ -57,7 +59,12 @@ class MapDetailPanel : public MapPanel {
 	int governmentY = 0;
 	int tradeY = 0;
 	
+	// Default display scaling for orbits within the currently displayed system.
+	double scale = .03;
```
petervdmeer, line 63: Should this be const? It looks like the value is never changed.

source/MapDetailPanel.cpp
```diff
@@ -271,6 +274,35 @@ bool MapDetailPanel::Click(int x, int y, int clicks)
...
+		if(uiClick.Length() > 130)
+			return true;
+		
+		// Only issue movement orders if the player is in-flight.
+		if(player.GetPlanet())
+			GetUI()->Push(new Dialog("You cannot issue fleet movement orders while docked."));
+		else if(!player.HasVisited(selectedSystem))
+			GetUI()->Push(new Dialog("You must visit this system before you can send your fleet there."));
+		else
+			player.SetEscortDestination(selectedSystem, uiClick / scale);
```
petervdmeer, line 298: It's a pity that this goes through the playerInfo class, but I guess that PlayerInfo is the only common shared information between the MapDetailPanel and the engine. Maybe add a comment to explain why the orders don't go to the Engine/AI directly?

source/MapDetailPanel.h
Amazinite: MapDetailPanel.cpp L616: `scale *= 115. / maxDistance;`

tux2603:
One thing to note is that this makes a certain exploit a little easier to accomplish

source/AI.cpp
tehhowch: That's an unfortunate side effect of how player-given orders work, and how we currently tell AI ships to act like an escort (i.e. pay attention to your parent and be near them) vs make their own decisions (e.g. fighting a target, following player-given orders, or performing intersystem travel operations without regard to a parent ship).

Refactoring that sequence of the AI class is something that could be done here, but could also be done in a separate PR. As @tux2603 mentions, making your escorts promptly ignore combat to jump away further expands the "combat only happens when the player can see it" exploit.

source/MapDetailPanel.cpp
tehhowch: Yep, the only way to reach AI currently is to go through its owner, Engine, which is owned by the main in-flight panel (as opposed to the map panels).

An opportunity for future architectural changes is to abstract some of the large AI class into e.g. `ShipAI` classes that are owned by any given ship (or group of ships). One would then be able to update a ship's goals, like a desired destination system, without needing the orchestrating AI class.

source/AI.cpp
petervdmeer: I'm okay with leaving this for a separate PR (and for leaving it until after fixing the "combat only happens when the player can see it" exploit).

source/MapDetailPanel.cpp
petervdmeer: Sounds good. But let's keep this for a future PR.

petervdmeer:
@Amazinite I can think of a number of ways to make the feature more discoverable, for example:
- Popping up a hint (once) when the map-panel gets opened when the player has more than 1 ship.
- Popping up a hint (once) when the map-panel gets opened when the player has ships selected.
- Showing selected ships (a small icon and numbers) when the map-panel gets opened with ships selected.
- Creating a new map-panel that is specifically for moving ships to other systems.
- Adding the feature to the game manual.

I would be fine if we keep discoverability for a follow-up PR since that could (and probalby should) trigger quite some discussion by itself on which option is best.

petervdmeer:
@Amazinite There is a chance that this conflicts with #4710, but my first impression is that #4710 only affects commands with a target-ship while this PR uses the move command which doesn't use a target-ship. Maybe we should merge one the PRs first and then merge master in the other PR and re-test the combination in the other PR.

tehhowch:
> 
> 
> @Amazinite There is a chance that this conflicts with #4710, but my first impression is that #4710 only affects commands with a target-ship while this PR uses the move command which doesn't use a target-ship. Maybe we should merge one the PRs first and then merge master into the other PR and re-test the combination in the other PR.

No issue that I see, testing this against latest `master`
This may also help those whose escorts' AI gets "stuck" after certain wormhole travel, e.g. #5541 et al., avoid needing to reload / save edit to un-stick their escorts, but I haven't attempted to repro & verify.

Issue #2211, opened 2017-03-03
Alternate Target Selection Method


VonZorch:
Select a target by clicking in the radar screen (upper left corner).  It may not be very important but I frequently want to know who that idiot is on the other side of the system.  Referring here to a ship in my fleet.
This would also be useful for missile duels against a distant ship.

Tadrix:
And, while at it, would be nice to be able to select a planet this way as well instead if flying towards it, manually selecting, bribing and praying ship will be able to land before silly pirates will blast it out of the sky (morons are frequently using up their missiles, forfeiting any chance of actually disabling and plundering instead of destroying their target)

Bladewood:
Pretty sure this is a duplicate. When I go home, I'll try to find the Steam thread for it.

jafdy:
#1748

Bladewood:
Damn, he beat me to it.

jafdy:
It was a pain to find the right key words for that one.

VonZorch:
I spent several hours doing broad searches and looking through issues before posting.
It would still be useful when there is a pure missile ship in an enemy fleet.  You could click on it and engage in a missile duel while your escorts deal with closer enemies.
If not then a modifier key such as <ctrl> that would select the next enemy  instead of the nearest enemy.  I'm using ~ to select the nearest enemy so in my case <ctrl>~would cycle through enemy ships.  I know you can use select next ship but if there are many friendly ships that can take too long to be useful.

DingusShingleton:
I knew it had been suggested before but I couldn't find it either.

Anyway I think targeting with the minimap is a bandaid for the targeting system which is not particularly fleshed out. The single use-case of targeting a distant missile boat hardly seems pressing enough to add a whole new method of targeting... I should think that revamping the existing targeting system to be more fluid and feature rich would be time more well spent.

jafdy:
This is an out and out duplicate. Please close.

Issue #1748, opened 2016-10-22
Mouse selectable ships on minimap


Admiral-Danforth:
![screen shot 2016-10-22 at 6 34 46 pm](https://cloud.githubusercontent.com/assets/22384290/19623193/417c480c-9886-11e6-8b77-b47e1e74e2fe.png) selectable on this thing

Bladewood:
I don't really think this will work, because dots on the minimap are INSANELY small.

LocalGod79:
and when there are lots of ships piled together, which one does it select?

Bladewood:
Well, the same thing would happen if you were up close and tried to select a ship in a pile without the minimap.

LocalGod79:
okay; it would probably be handled the same way, but adding the feature would only make that particular problem worse, wouldn't it?

Admiral-Danforth:
It would be useful if you have a ship that is ten miles away spewing missiles and your to lazy to click R or something.

PR #2679, opened 2017-06-27
Left- and right-clicking the Radar display issues appropriate click commands

tehhowch:
Refs:
#1748 #2211
https://steamcommunity.com/app/404410/discussions/2/1471966894886573827/

Summary:
1) Adds a preference to enable and disable this PR's features
2) Any click that starts within the radar display results in reinterpreting that click to the position referenced by the radar display, rather than the point underneath the radar display:
    - target selection
    - landing autopilot activation
    - drag-selecting escorts
    - shift-clicking escorts
    - issue Move / Follow / Attack orders

To select/target via the radar display requires the player to be quite precise, but with a mouse this should not be difficult (I can do so with a trackpad without difficulty). Movement orders do not require such precision, and are very easy to achieve.
________
original OP:
Current behavior is to allow clicking "behind" or through the radar display, e.g. a right-click acts as though the radar display is not present.

With this PR, right-clicking the radar display will send a MOVE_TO order to your selected escorts, with the scaled target position. If you are extremely nimble and judicious with your mouse, you can issue "Follow ship" or "Attack Ship" orders, depending on the government of the ship whose radar dot you are clicking.

This allows ordering your escorts up to 128 / 0.025 = 5120 units away from your flagship's current position in the system, without requiring you to zoom out.


ericedem:
Out of curiosity: What is a use case for this?

MessyMix:
Say you want to send your escorts somewhere, but it's off your screen / beyond your zoom level. This lets you right-click the radar to order your escorts to move to the position located on the radar screen. the current behavior iirc just pretends that the radar screen doesn't exist, so if you right click onto the radar screen your escorts just end up moving behind it.

ericedem:
Sorry I should have been more specific. I was wondering what in game tactical strategy this would make possible. When would you want to move your escorts off to some location? Is this useful for conquering / blockading planets?

jafdy:
To get them out of the way. You don't overkill as much or they are safer.

tehhowch:
Yes, primarily a tactical response. The radar shows things with blast radii such as nuclear missiles, so this allows skilled captains to both fly out of the way and redirect escorts away from the flagship's relative position. One is also able to tell specific ships to leave combat whilst intervening, and not telling that escort to remain relatively close by.

gunqqer:
Oh thank you for this. I've always was slightly annoyed why it didn't do it before.

endless-sky:
This should really be querying the radar position in the `"targets"` interface definition, rather than hard-coding the position of the radar.

tehhowch:
I'm wondering if this should also reinterpret left-clicks on the radar, as in its current form if a right-click is performed accurately enough, the ship at the clicked position is the target of a follow/attack order. With a right-click redirecting, it seems logical that a left click should as well. Drag selections would remain fixed to the player's screen.

This could also easily be gated behind a preference if desired.

tehhowch:
I made a preference to control this behavior, and included left-clicks and dragging in the rescaled behavior. 
 - When the preference is inactive (default behavior), the radar display is purely informative and non-interactive.
 - When the preference is active, any click/drag that starts within the radar display will result in a reinterpreted point, and any click/drag that starts elsewhere is handled normally.
    - Single-left-clicking will select the target of the click
    - Double-left-clicking a planet will land
    - Drag-selecting escorts works
    - Shift-clicking escorts works
    - Right-clicking issues a move order (or follow/attack order, if the radar position of a ship was clicked)

Obviously, to select/target via the radar display requires the player to be quite precise, but with a mouse this should not be difficult (I can do so with a trackpad without difficulty). Movement orders do not require such precision, and are very simple to achieve.

source/Engine.cpp
```diff
@@ -790,8 +790,20 @@ void Engine::Click(const Point &from, const Point &to, bool hasShift)
 	doClickNextStep = true;
 	this->hasShift = hasShift;
 	isRightClick = false;
-	clickPoint = from;
-	clickBox = Rectangle::WithCorners(from / zoom + center, to / zoom + center);
+	
+	// Determine if the left-click was within the radar display.
+	const Point radarCenter = GameData::Interfaces().Get("targets")->GetPoint("radar");
+	static const double radarRadius = GameData::Interfaces().Get("targets")->GetSize("radar").Y() * .5;
```
endless-sky: As with other UI values, it's probably worthwhile to have this be non-static in case future game changes make it possible to reconfigure the heads-up-display interface while in-game.

source/PreferencesPanel.cpp
```diff
@@ -438,11 +438,12 @@ void PreferencesPanel::DrawSettings()
 		"Show hyperspace flash",
 		"\n",
 		"Other",
+		"Interactive radar display",
```
endless-sky: I feel like "*Clickable* radar display" would make the functionality more obvious than "Interactive."

tehhowch: :grinning: I had almost changed it to that when merging yesterday, and then second guessed myself because clickable didn't relay back that the game would respond to the user clicking it.

source/Engine.cpp
tehhowch: Addressed

PR #2684, opened 2017-06-30
Marked the Syndicate Test Dummy ships with the new "target" personality

Pointedstick:


tehhowch:
I almost included them in my espionage PR, but wasn't sure if we should reserve the target personality for non hostile targets

Pointedstick:
The Test Dummy ships are most definitely hostile. :-)

PR #2691, opened 2017-07-02
Fixing number of passenger for "Terraforming 2"

Amazinite:
It was pointed out to me by the user fgntlucas that "Terraforming 2" has 2 passengers, despite the fact that you pick up Amy on Glory, which means you should have 3 passengers.

### DIFF
--- a/data/transport missions.txt
+++ b/data/transport missions.txt
@@ -1690,7 +1690,7 @@ mission "Terraforming 2"
 	description "Return to Rand with a student who thinks she can terraform the planet cheaply."
 	source Glory
 	destination Rand
-	passengers 2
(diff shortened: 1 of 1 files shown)


endless-sky:
The reason it has two passengers instead of 3 is to avoid the situation where you arrive on the planet and are unable to continue the mission. That seemed more important than absolute accuracy.

Amazinite:
Would adding a blocked dialog be sufficient enough, or would you rather the player always be able to continue the mission string with no interruptions?

tehhowch:
Or set it and its predecessor to 3 with the blocked dialog for the predecessor

PR #2692, opened 2017-07-02
Fixing a missing quotation mark

Amazinite:


Bladewood:
It's also possible that that last quotation mark that should not be there.

tehhowch:
The changed sentence definitely feels as though it is being stated by Katya.

PR #2699, opened 2017-07-03
Allow events to freeze the player's reputation with a given government

Hadron1776:
This would allow events to prevent the player from changing his/her reputation with a government with one attribute. Then, if you want the player in a story where he/she is fighting another faction, you wouldn't need to repeatedly tank the player's reputation with that faction. With this PR, you could just set it to -1, set `unwavering 1`, and forget about it until he/she reaches the end of the storyline, at which point, just set `unwavering 0` to unlock the reputation.

Example:
`event "SomeEvent"`
`    government SomeGov`
`        unwavering 1`

...unset:
`event "SomeOtherEvent"`
`    government SomeGov`
`        unwavering 0`

You could theoretically use the attitude matrix to do this in the `master` branch, but you would need to specify each enemy of the faction whose reputation you want to "lock".

This PR doesn't prevent the reputation drop from committing an atrocity, though. (Maybe it should, but I kinda wanted to prevent an exploit due to the fact that you are no longer penalized for killing stuff with the lock in place)

### DIFF
--- a/source/Government.cpp
+++ b/source/Government.cpp
@@ -321,4 +321,3 @@ void Government::SetReputation(double value) const
 {
 	GameData::GetPolitics().SetReputation(this, value);
 }
-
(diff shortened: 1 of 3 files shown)


endless-sky:
I'm not convinced that this is the best way to keep a player's reputation from changing during a mission string. I'd prefer to think about other options before committing to this one.

PR #2702, opened 2017-07-04
Fix Skybound personality

tehhowch:
In the current master, the `skybound` personality does not do what it is intended to do, per 51c79f9ca419
1. If not also `uninterested`, skybound did not prevent landing.
2. if uninterested, NPCs would never hail for fuel, as they would think they could refuel in friendly systems. 
3. if uninterested, NPCs would hover over only landable planets, even if non-landable planets were present, because they were always given a target StellarObject

With this PR, the `skybound` NPC personality indicates it:
 - Will not land, even if it is an escort/has a parent, or even if it needs refueling
 - Will always be in space rather than launching from the planet
 - Will hail friendly ships to obtain fuel if out of fuel

This was accomplished by placing the relevant `IsSkybound` checks in AI::CanRefuel and Planet::CanLand, so that the NPC recognizes when it is stranded, and so that AI::MoveIndependent will not give it a random planet to try to land on.

The refactor of MoveIndependent restructured it such that any ship which cannot jump (for whatever reason), cannot land (for whatever reason), or is a stranded fighter/drone will patrol the available system's objects until a hostile target, a ship which will refuel it (if it needs fuel), or a ship which can carry it appears in system.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -970,18 +970,15 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
 		int systemTotalWeight = totalWeight;
 		
 		vector<const StellarObject *> planets;
-		if(!ship.GetPersonality().IsSkybound())
-		{
-			// Anywhere you can land that has a port has the same weight. Ships will
-			// not land anywhere without a port.
-			for(const StellarObject &object : ship.GetSystem()->Objects())
-				if(object.GetPlanet() && object.GetPlanet()->HasSpaceport()
-						&& object.GetPlanet()->CanLand(ship))
-				{
-					planets.push_back(&object);
-					totalWeight += planetWeight;
-				}
-		}
+		// Anywhere you can land that has a port has the same weight. Ships will
+		// not land anywhere without a port.
+		for(const StellarObject &object : ship.GetSystem()->Objects())
+			if(object.GetPlanet() && object.GetPlanet()->HasSpaceport()
+					&& object.GetPlanet()->CanLand(ship))
+			{
+				planets.push_back(&object);
+				totalWeight += planetWeight;
+			}
 		// If there are no ports to land on and this ship cannot jump, consider
 		// landing on uninhabited planets.
 		if(!totalWeight)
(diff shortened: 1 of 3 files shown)


tehhowch:
Here are a few test missions that illustrate the problems with the rendition of `skybound` as of the current `master`:
1: This ship will jump to wherever it desires (due to `uninterested`). When it runs out of fuel, it will hover over whatever planet in its current system is landable in general.
2: This ship will follow you around, and if it runs out of fuel, land.
3: This ship will KeepStation, unless you cloak, in which case it will perform a patrol of the system's stellar objects. If while navigating to your current system it happens to run out of fuel, it will land. Because it is `fleeing`, it will not respawn until you lift off from a planet, nor will it target you and try to fight.
4: This ship will jump to wherever it desires (due to `uninterested`) if you become untargetable (jump away or cloak). When it runs out of fuel, it will hover over whatever planet in its current system is landable in general, as in 1.

With this PR
1. This ship will jump to wherever it desires (due to `uninterested`). When it runs out of fuel, it will hail a ship to refuel it. If there are no ships to hail, it will patrol the system.
2. This ship will folow you around, and if it runs out of fuel, hail a ship to be refueled. If there are no ships, it will KeepStation with your flagship, as expected.
3. This ship will KeepStation, unless you cloak, in which case it will perform a patrol of the system's stellar objects. If it runs out of fuel and you are not targetable, it will hail a friendly ship to be refueled.
4. This ship will jump to wherever it desires (due to `uninterested`) if you become untargetable (jump away or cloak). When out of fuel, it will hail a friendly ship to be refueled.

[npc_skybound.txt](https://github.com/endless-sky/endless-sky/files/1129836/npc_skybound.txt)

tehhowch:
The `skybound` Korath Worldship happily takes off from the planet you landed on in `master`. 
screenshot from 2017-07-12 21-05-17
The worldship also will not hail Remnant ships for fuel in the 3 Remnant systems with planets (as it simply hovers on the planet as mentioned above).
With this PR, that takeoff behavior isn't possible. The `skybound` ship will also not hover over the Remnant planets until it has enough fuel to jump, but rather hails Remnant ships to be refueled (which is what it does on `master` if it is out of fuel in the uninhabited systems).

endless-sky:
That worldship is not supposed to hail Remnant ships for fuel. It is supposed to be a hostile ship that the Remnant is leaving alone because you agreed to fight it rather than them. So, it would be really weird for them to be refueling it.

tehhowch:
It already happily hails them for fuel, because of the government relationship between Bounty and Remnant. I'm not sure how you can make a hostile-in-theory ship not be fought by a government, yet also not be assisted by that government, but that is an issue with something other than the Skybound personality.

endless-sky:
The only point of "skybound" was to create a ship that would not be able to refuel on Remnant worlds even though it wouldn't be attacked by Remnant ships. So, I've taken it out and replaced it with something that makes more sense: a "marked" personality that means a ship will only fight the player and only player ships will fight it. (68cd8d0)

PR #2708, opened 2017-07-04
Allow modders to override default crew attack/defense values per government

Hadron1776:
Adds 2 new optional attributes to governments: `crew attack` and `crew defense`. This would be the strength per crew member of that government's ships. (If one or both attributes are omitted, the default values of 1 for attacking and 2 for defending will be used.) This allows one to create alien governments whose ships are easier or harder to capture than normal.

### DIFF
--- a/source/CaptureOdds.cpp
+++ b/source/CaptureOdds.cpp
@@ -169,7 +169,7 @@ vector<double> CaptureOdds::Power(const Ship &ship, bool isDefender)
 	
 	// Check for any outfits that assist with attacking or defending:
 	const string attribute = (isDefender ? "capture defense" : "capture attack");
-	const double crewPower = (isDefender ? 2. : 1.);
+	const double crewPower = (isDefender ? ship.GetGovernment()->CrewDefense() : ship.GetGovernment()->CrewAttack());
 	
 	// Each crew member can wield one weapon. They use the most powerful ones
 	// that can be wielded by the remaining crew.
--- a/source/CaptureOdds.cpp
+++ b/source/CaptureOdds.cpp
@@ -14,6 +14,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include "Outfit.h"
 #include "Ship.h"
+#include "Government.h"
 
 #include <algorithm>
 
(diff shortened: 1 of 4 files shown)


source/CaptureOdds.h
```diff
@@ -14,8 +14,10 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #define CAPTURE_ODDS_H_
 
 #include <vector>
+#include "Government.h"
```
tehhowch: This should be the new line 16, and the std include should then be line 18, to follow the style guide.

source/CaptureOdds.cpp
```diff
@@ -14,6 +14,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include "Outfit.h"
 #include "Ship.h"
+#include "Government.h"
```
tehhowch: Alphabetize per style guide

endless-sky:
Can't this just as easily be done by arming the crews? Because currently the player is the only one who tries to capture ships, you can just give the aliens a weapon that provides defense but not attack power, and there's no risk that that weapon will be overpowered in the player's hands.

Hadron1776:
At least this PR would cater to people who want weaker crews than the default, as you can't have a total "capture attack/defense" value of less than 0 (which would be a requirement when you want a crew to have a defense value less than 2. In other words, you can't replicate this effect in the master branch:
`	"crew attack" .7`
`	"crew defense" 1.4`

Hadron1776:
...and if you want to change the crew attack/defense values later (in the `master` branch), all you need to do is change the ship outfits around... for _every_ ship owned by that government. So... Can it be done in master? Probably. Just as easily? No. IMO

Elyssaen:
I do like this, though it touches on two other areas in ways that leave me musing:

1. People have talked before about the idea of having different types of crew, and that might include different species. This draws attention to that, as the (perhaps unimportant) shortcoming of your approach is it fixes on a government basis something that is really a question of what _type_ of crew a ship has.

2. When I think about a commit we need to address capture attack/defence, what it really seems like we need is some way to make ships reasonably hard to capture that don't fall into the problems of `capture defence`. Even with Pug Biodefences, you really need a high crew to be hard to capture, as it's just how the numbers play out. Then we end up with the Wardragon crew being higher than makes sense to me... unless there's a secret reason I'm missing.

Issue #2704, opened 2017-07-04
Heroic ships clustering around the system center


Amazinite:
If there is a ship with the heroic personality whos only target is outside of the boundry of the system, then that ship will wait at the system center for the target to enter back into the boundry of the system. 
![screenshot_1](https://user-images.githubusercontent.com/17688683/27811386-d4e54aec-6033-11e7-9b86-e91f167efd61.png)
My ships are currently holding position far from the system center, meaning that these Wanderer ships would stay here for as long as my ships remain where they are.
This was brought up [on Steam](http://steamcommunity.com/app/404410/discussions/1/2592234299570997332/), and I'm wondering if it's intended behavior or not for heroic ships to wait indefinitely for their target to become viable. 
The same behavior is not experienced when all hostile ships in the system are cloaked (i.e. if my flagship is the only one in the system and is cloaked).

tehhowch:
This is from AI::MoveIndependent
```
	if(target && !ship.IsYours() && !ship.GetPersonality().IsUnconstrained())
	{
		Point extrapolated = target->Position() + 120. * (target->Velocity() - ship.Velocity());
		if(extrapolated.Length() >= MAX_DISTANCE_FROM_CENTER)
		{
			MoveTo(ship, command, Point(), Point(), 40., .8);
			if(ship.Velocity().Dot(ship.Position()) > 0.)
				command |= Command::FORWARD;
			return;
		}
	}
```
which does the movement, and AI::FindTarget()
```
// Find the closest enemy ship (if there is one). If this ship is "heroic,"
	// it will attack any ship in system. Otherwise, if all its weapons have a
	// range higher than 2000, it will engage ships up to 50% beyond its range.
	// If a ship has short range weapons and is not heroic, it will engage any
	// ship that is within 3000 of it.
	double closest = person.IsHeroic() ? numeric_limits<double>::infinity() :
		(minRange > 1000.) ? maxRange * 1.5 : 4000.;
```
There are two ways to fix this... make `heroic` equivalent to `heroic unconstrained` by changing `!ship.GetPersonality().IsUnconstrained()` to `!(ship.GetPersonality.IsUnconstrained() || ship.GetPersonality().IsHeroic()`,
or make heroic only target ships within the max_distance_from_center range, rather than infinite range.
[...]

PR #2713, opened 2017-07-06
Heroic personality no longer targets any hostile in system

tehhowch:
`Heroic` ships will instead target any hostile within 5000 of themselves, or an ally's target that is within 5000 of it
`Unconstrained` ships will target any hostile, so long as it is not too powerful.
`Heroic unconstrained` ships will target any hostile no matter how strong, and no matter where it is in the system.

This should solve #2704 while also making the wiki text on `heroic` accurate.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -717,13 +717,16 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 	if(parentTarget && !parentTarget->IsTargetable())
 		parentTarget.reset();
 
-	// Find the closest enemy ship (if there is one). If this ship is "heroic,"
-	// it will attack any ship in system. Otherwise, if all its weapons have a
-	// range higher than 2000, it will engage ships up to 50% beyond its range.
-	// If a ship has short range weapons and is not heroic, it will engage any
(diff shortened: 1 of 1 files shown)


Amazinite:
Resolves http://steamcommunity.com/app/404410/discussions/1/2592234299570997332/

tehhowch:
@endless-sky do you view the current behavior for `heroic` in the linked cases as appropriate? Or are there some remaining changes that need to be made in this PR before it fully addresses the intent of the heroic personality?

endless-sky:
The whole point of the "heroic" modifier is that ships won't stop trying to fight just because you run away from them.

If the automata are getting stuck waiting around in a system just because the player is hovering outside the "invisible fence," a much simpler fix would be to switch those fleets to not being "heroic" (226bdef). But, I can't duplicate the behavior from the Steam screenshots, so I'm not sure if that's what was going on.

tehhowch:
That description of `heroic` doesn't match up with the wiki text on heroic: 
```
heroic: goes out of its way to join fights when an ally is threatened.
```

The wiki text on `heroic` was in large part the reason why I had structured this PR's targeting logic in the way that I did.

PR #2719, opened 2017-07-06
New outfit attributes: disruption, ion, piercing and slowing resistance

Elyssaen:
```
Adds outfit attributes that resist the effect of disruption, ion,
piercing and slowing damage, respectively. Each functions by 'shaking
off' an amount of damage each second, immediately if possible.
```

@endless-sky, given your commit 2bf7c0f089d72aef14a301bbdd09b9639f17c1e0 in the last half an hour, I thought I should finally PR this piece of work. It includes piercing resistance but not heat resistance. It's been tested (by myself and a few others in the community) and quite a bit of work's gone into it.

It follows a slightly different approach to your fractional scaling of the damage. In this PR, the resistance is an amount of that type of damage you ignore each second. This means your resistance values are pegged to the strength of the effect. You can be completely immune to a weak disruptive effect, but that same resistance might not be worth much against a very strong (Tier 3/4) disruption. I playtested other approaches, but this one seemed to be most resilient across tiers.

Working on the 'resisted value' approach rather than 'percentage reduction' approach, `piercing resistance` was a little more complicated, as I didn't want to make a piercing beam too weak or a piercing missile too strong. The proposed method here is that your piercing resistance covers a second of play and slowly restores, so a continuous beam can wear it down.

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -63,8 +63,8 @@ tip "cost:"
 	`Cost, in credits. If you have sold any used (depreciated) items today, you can buy them back for the same price you sold them for.`
 
 tip "disruption resistance:"
-	`Resistance to weapons that have a disruption effect. A value of 1 cuts the effect in half, and 2 cuts it to a third of what it would otherwise be.`
-
+	`The amount of disruption damage, per second, that your ship will resist.`
+	
 tip "drag:"
 	`Limits the maximum speed of a ship.`
 
@@ -129,7 +129,7 @@ tip "illegal:"
 	`Amount of credits you can be fined for carrying this outfit.`
 
 tip "ion resistance:"
-	`Resistance to weapons that do ion damage. A value of 1 cuts the effect in half, and 2 cuts it to a third of what it would otherwise be.`
+	`The amount of ion damage, per second, that your ship will resist.`
 
 tip "jump speed:"
 	`How slow the ship must be moving to make a jump.`
@@ -167,6 +167,9 @@ tip "radar jamming:"
 tip "ramscoop:"
 	`Replenishes fuel by harvesting the stellar wind. Ramscoops are more effective when closer to a star. If you add more than one ramscoop, each additional one is less effective: you need four ramscoops to achieve double the recharge rate that one ramscoop provides.`
 
+tip "piercing resistance:"
+	`The amount of piercing hull damage, per second, that your ship will resist.`	
+	
 tip "required crew / bunks:"
 	`How many crew members are required to operate this ship, and how many total bunks are available. Extra bunks can be used for passengers or extra crew.`
 
@@ -210,7 +213,7 @@ tip "shields:"
 	`Total shield strength. Most weapons cannot damage your hull until your shields are depleted. Some weapon effects, such as heat damage, become more severe once shields are down.`
 
 tip "slowing resistance:"
-	`Resistance to weapons that have a slowing effect. A value of 1 cuts the effect in half, and 2 cuts it to a third of what it would otherwise be.`
+	`The amount of slowing damage, per second, that your ship will resist.`
 
 tip "solar collection:"
 	`Produces a variable amount of energy depending on how far this ship is from the star at the center of the system.`
@@ -218,6 +221,9 @@ tip "solar collection:"
 tip "max speed:"
 	`The fastest this ship can travel when using ordinary thrusters (as opposed to an afterburner).`
 
+tip "slowing resistance:"
+	`The amount of slowing damage, per second, that your ship will resist.`
+	
 tip "acceleration:"
 	`How quickly this ship gains speed. The higher a ship's mass (including the mass of any cargo or fighters it is carrying), the slower it accelerates.`
 
--- a/source/OutfitInfoDisplay.cpp
+++ b/source/OutfitInfoDisplay.cpp
@@ -173,6 +173,12 @@ void OutfitInfoDisplay::UpdateAttributes(const Outfit &outfit)
 		double scale = 1.;
 		if(it.first == "thrust" || it.first == "reverse thrust" || it.first == "afterburner thrust")
 			scale = 60. * 60.;
+		// As resisted ion/etc. damage decays with *= .99, it follows a geometric series with convergence
+		// to 100 times the initial value. The 'value' of resisting 1 point is therefore eventually 100.
+		else if(it.first == "ion resistance" || 
+				it.first == "slowing resistance" || 
+				it.first == "disruption resistance")
+			scale = 60. * 100.;
 		else if(ATTRIBUTES_TO_SCALE.count(it.first))
 			scale = 60.;
 		
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2118,10 +2118,10 @@ int Ship::TakeDamage(const Projectile &projectile, bool isBlast)
 	double shieldDamage = weapon.ShieldDamage();
 	double hullDamage = weapon.HullDamage();
 	double hitForce = weapon.HitForce();
-	double heatDamage = weapon.HeatDamage() / (1. + attributes.Get("heat resistance"));
-	double ionDamage = weapon.IonDamage() / (1. + attributes.Get("ion resistance"));
-	double disruptionDamage = weapon.DisruptionDamage() / (1. + attributes.Get("disruption resistance"));
-	double slowingDamage = weapon.SlowingDamage() / (1. + attributes.Get("slowing resistance"));
+	double heatDamage = weapon.HeatDamage();
+	double ionDamage = weapon.IonDamage();
+	double disruptionDamage = weapon.DisruptionDamage();
+	double slowingDamage = weapon.SlowingDamage();
 	bool wasDisabled = IsDisabled();
 	bool wasDestroyed = IsDestroyed();
 
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -709,20 +709,38 @@ bool Ship::Move(list<Effect> &effects, list<shared_ptr<Flotsam>> &flotsam)
 	// Handle ionization effects, etc.
 	if(ionization)
 	{
+		if(ionResistance)
+			ionization = max(ionization - ionResistance, 0.);
 		ionization *= .99;
 		CreateSparks(effects, "ion spark", ionization * .1);
 	}
 	if(disruption)
 	{
+		if(disruptionResistance)
+			disruption = max(disruption - disruptionResistance, 0.);
 		disruption *= .99;
 		CreateSparks(effects, "disruption spark", disruption * .1);
 	}
 	if(slowness)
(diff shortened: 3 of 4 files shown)


tehhowch:
Is it possible to have both types of resistance?

I can imagine some highly advanced outfits that convey percentage resistance, while cheaper outfits convey a static damage per second resistance.

Amazinite:
Was about to ask that, @tehhowch. Having both resistance types would be great.

Elyssaen:
Hypothetically, yes. I could see the case that that might be confusing to some, that maybe having both features is a 'total conversion support' matter rather than something you'd see both of in one universe.

My concern about percentage resistances was what the scaling would look like at higher tiers. I've seen games where the following happens:

1. Resistances like 50%, 75%, etc. cut hard into enormously powerful effect damages of higher tier weapons.
2. To make those higher tier weapons 'punch through', their effect damages then need to be extremely high, and don't always balance well against zero resistance.
3. You end up with the assumption that everyone will need to have resistances like 95%, and it becomes hard to track numbers: am I doing 5000 damage? Probably not, they have 98% resistance, so that's... 100.

The other bit was, immunity to slowing or disruption _at a certain level_ is actually cool and useful. I like the idea that you could plausibly be utterly immune to Kor Mereti disruption in normal numbers, while still being vulnerable to hypothetical stronger disruptors.

tehhowch:
@Amazinite probably better to have 
`damage resistance`: % reduction in damage type
`damage absorption`: static per-second reduction
resistance and resilience are too similar IMO.

jafdy:
I think I prefer this way of doing it.
All of these effects can be applied by both beam and projectile weapons so it may be best to implement them all in the same way you implemented the piercing value.

Elyssaen:
With disruption, slowing and ionisation, the weapon might apply the effect in just one frame, but the status lingers into additional frames. The way I was proposing implementing resistances to them, the resistance will immediately try to cancel out any damage taken, and if any of that frame's resistance is left, it will try to wipe out any lingering status effect.

These three were my starting point and it seems to play well. It does mean that, yes, a missile 'spiking' disruption damage will get most of it through in that first frame, and though it will be resisted away in the successive frames it might be a little more effective than an equivalent beam.

At the time, I thought okay, that's not a problem. Then I looked at piercing resistance and had the problem that piercing _isn't_ a lasting status effect. It forces either going for % resistance (which as said bothers me from a scalability perspective) or what I've tried here, aggregating the resistance over a second.

I guess that same method could apply to the others, even if it's not needed in quite the same way.

endless-sky:
I can see several different options here, but they all come down to two questions:

1. Should resistance be an amount per second, or a percentage?
2. Should there be some sort of resistance to piercing effects?

If resistance is a rate per second, it means that a ship could, for example, completely ignore ion effects if it's getting hit by one ion cannon but not if it's getting hit by two. That's the same as ordinary damage, where you can ignore it if the damage you're taking is less than your shield recharge rate. This would also mean that mechanically, cooling is equivalent to heat resistance and energy generation is equivalent to ion resistance, except that cooling can also be used on heat from other sources and energy generation can be used on energy drains from other sources.

If resistance is a percentage, it means that the same ship getting hit by one ion cannon will be slightly affected; there will be no sudden inflection point where the ion cannon goes from doing almost nothing to having a significant effect. On the other hand, if resistance is conferred by an outfit it means that any large ship can afford to buy a lot of resistance, using only a small portion of its outfit space.

But the question of whether there should be piercing resistance, and if so how it should operate, seems a lot more complicated than that. I'd prefer to hold off on making a decision of how that should work, rather than committing to one implementation before thinking through alternatives.

tehhowch:
Per 1) I would argue that there is room in the engine for both mitigation methods (just not with both being called "resistance"). Leave it to the balancing skills of the content creator to make it work - after all, the current "static resistance to damage from outfits" works for the base game re: shield and hull damage, so allowing the extension of this technique to non shield/hull(/heat) damage makes sense.

endless-sky:
Another option for piercing resistance would be for it to follow exactly the same mechanics as ionization, etc., except in reverse. That is, every frame:

```
piercingBuffer += piercingResistance
piercingBuffer *= .99
```

And every time a ship takes piercing damage:

```
piercingDamage = piercing * hullDamage
shieldFraction = 1 - piercing * (1 - min(1, piercingBuffer / piercingDamage))
piercingBuffer -= shieldFraction * piercingDamage
```

It would still have the same effect that "piercing resistance" is the most piercing hull damage you can resist per frame, but would make the change in resistance over time be a smoother line, recovering full strength asymptotically instead of racing right up to full strength and then stopping.

Amazinite:
My thoughts on the topic:

I like the idea of having both `resistance` and `absorption` attributes (resistance being what has just been added to the game and absorption being @Elyssaen's take on it). Resistance seems like very high level tech to me, to be able to shrug off an entire chunk of damage from a certain source no matter how much you're getting hit by. This absorption idea though feels like it could be used on a lower level. Having both attributes would certainly be interesting and a treat for plugin makers and their alien races.

As for the topic of `piercing absorption` (`resistance` in this PR), I don't really think it makes much sense. As @endless-sky pointed out, this absorption idea is similar in nature to how you can shrug off shield or hull damage with your shield generators or hull repair outfits. Piercing damage is nothing more than hull damage that goes through the shields, so hull repair outfits already act as piercing absorption in a way. Although, a `piercing resistance` that is similar to the way that @endless-sky currently has resistances in vanilla seems like it would make sense, ignoring a percentage of all hull damage that bleeds through to the hull.

Speaking of shield and hull repair, how about some sort of `absorption energy` and `absorption heat` attributes? Since any outfit with an absorption attribute is pretty much acting like a shield generator but for a different damage type, I think it makes sense that they would require energy and produce heat when they are active. (Resistance outfits on the other hand seem like they would have constant energy and heat to them, if any at all.)

Also:
>This would also mean that mechanically, cooling is equivalent to heat resistance and energy generation is equivalent to ion resistance, except that cooling can also be used on heat from other sources and energy generation can be used on energy drains from other sources.

This is indeed the case. Ion absorption is in a way a dumbed down energy generation, and the same for heat absorption to cooling. Because of this, these attributes would allow for very niche ship setups.
Right now, the most cooling provided by any outfit per frame is 69 at 40 tons from the Liquid Helium Cooler. The largest (obtainable) reactor outfits all produce between 18.2 (Fusion Reactor) and 43.6 (Triple Plasma) energy per frame, most of which being in the 20s for their energy per frame (the Blue Sun and Triple Plasma reactors being the only ones above 30 energy/frame), but all of which being over 90 tons, going all the way up to 169. An outfit with `heat absorption` could provide 70 absorption/frame at a size smaller than 40 tons, or an outfit with `ion absorption` could provide around 20 absorption/frame at something much smaller than 90 to 169 tons. These outfits would then provide a purely combat bonus at an efficiency much better than if the player had installed more cooling or more reactors, but in turn they do not get the benefit that cooling outfits or reactors give, which is also combating heat production and energy drain from the player's own ship. The dynamic of such outfits really seems interesting to me.

Elyssaen:
Got to dash, but just to say: there is a key gameplay difference between my `piercing resistance` proposal and hull regeneration. `Piercing resistance` doesn't _negate_ the damage: it pushes the damage that would pierce through to the hull _back to the shield_. It's actually resisting the piercing effect itself, not the damage done as a result of it.

It always felt unfinished that even a seemingly low-level piercing effect like the Railgun can penetrate _any_ shield. A shield's job is to stop the attack going through: shouldn't a sufficiently advanced shield be able to stop a weak piercing effect from going through?

tehhowch:
In addition to energy and heat cost-of-use, a general effect cost would make intricate defensive outfits possible.
E.g. a damage absorption outfit that protects against slowing damage, but causes ionization damage + heat + fuel

TheUnfetteredOne:
Why stop at absorption? Shield fuel, please.

Ealb:
As far as names go, I'd say I like resistance and dissipation (or recovery, or even deflection, though that I like the least) more than resistance and absorption. If I may superficially nitpick.

Elyssaen:
> Another option for piercing resistance would be for it to follow exactly the same mechanics as ionization, etc., except in reverse. That is, every frame:
> 
> piercingBuffer += piercingResistance
> piercingBuffer *= .99

The other way around, right? So it will tend to `100 * piercingResistance`, rather than 99 times.

Does this mean that your ship takes a couple of hundred frames to build `piercingBuffer` up on launching? Or do we set `piercingBuffer = 100 * piercingResistance` on Recharge or Launch to avoid that?

> 
> And every time a ship takes piercing damage:
> 
> piercingDamage = piercing * hullDamage
> shieldFraction = 1 - piercing * (1 - min(1, piercingBuffer / piercingDamage))
> piercingBuffer -= shieldFraction * piercingDamage
> 
> It would still have the same effect that "piercing resistance" is the most piercing hull damage you can resist per frame, but would make the change in resistance over time be a smoother line, recovering full strength asymptotically instead of racing right up to full strength and then stopping.

Would play fairly similarly I guess. My way linearly recovers up to one second's worth of resistance, yours recovers asymptomatically hitting about 45% in 1 sec, 70% in 2, 84% in 3, and 90% in 5. Gameplay-wise I don't personally have a preference, but your way is probably more in keeping stylistically.

What do you think at the moment? I hadn't posted this thinking there was room for both methodologies, but perhaps that _is_ the best answer. Would you go in for `resistance` and `absorption` separately, if I tweaked this?

> Speaking of shield and hull repair, how about some sort of absorption energy and absorption heat attributes?

That... probably does make sense. I'd like more input on it. It would involve writing a fair few more lines to support it, as we'd need `disruption absorption energy,` `slowing absorption energy`, and perhaps we're adding `heat absorption` too...

> In addition to energy and heat cost-of-use, a general effect cost would make intricate defensive outfits possible.
> E.g. a damage absorption outfit that protects against slowing damage, but causes ionization damage + heat + fuel consumption

Different topic I guess? I've wondered about this but I quickly get into ideas that are too much of an overhaul. i.e. moving toward an outfit having a whole `passive effect` / `active effect` handler.

endless-sky:
Well, here's a compromise (ab5fd68). I switched ion, disruption, and slowing resistance to use the mechanics that you proposed, because that does seem to scale better. I left out heat and piercing resistance for now. (They could be added somewhere down the road, if there's ever a clear consensus on how they could be implemented.)

Amazinite:
It'd still much rather have both systems be in place.

@Elyssaen 
>That... probably does make sense. I'd like more input on it. It would involve writing a fair few more lines to support it, as we'd need `disruption absorption energy`, `slowing absorption energy`, and perhaps we're adding `heat absorption` too...

Could they perhaps all be under two attributes? i.e. instead of `dirsuption absorption heat`, `slowing absorption heat`, `heat absorption heat`, and so on, a single attribute of `absorption heat` handles all situations of heat being applied to the player's ship when absorption of any kind is activated, and likewise for `absorption energy`?

Issue #2716, opened 2017-07-06
Long-range AI doesn't use reverse thrusters


McNutty145:
I haven't had a chance to confirm this behavior in-game and won't be able to for a while, but I was reading the AI.cpp file and I think I discovered a minor bug. An AI-controlled ship with long-range weapons will maintain a minimum distance from its target by fully turning around and using forward thrust. This occurs even if reverse thrusters are installed, which unnecessarily wastes time taking weapons off-target when it could keep firing while reversing.

I could also be reading it wrong. Relevant code is in AI.cpp lines 1518-1527.

tehhowch:
If the reverse thrusters are weak compared to the primary thrusters, this may result in the target catching the missile boat when the target could be kept at bay via turn + thrust + turn back.

I've got a thought, though.

endless-sky:
It's kind of a moot point because there aren't any ship variants in the game that use reverse thrusters. Reverse thrusters are basically just included as a crutch for players who can't figure out how to control a ship without one.

tehhowch:
(yet)

Any mods that use reversing variants would certainly benefit

Amazinite:
>Reverse thrusters are basically just included as a crutch for players who can't figure out how to control a ship without one.

Slow down, what? 
You don't start out with reverse thrusters, and reverse thrusters aren't exactly cheap, so how is it that a player is going to struggle at the beginning of their playthrough trying to control their ship but then somehow get to the point where they both discover and have the credits to purchase reverse thrusters for their ship? Said player then needs a ship with the weapon space to fit the reverse thrusters, and none of the starter ships can do that, which means said player would also have had to play long enough to upgrade their ship. 
I'm really just confused at how a player could get to that point not knowing how to fly well, then find reverse thrusters and consider them far better to use.

Startome:
I'm reading this as him meaning that players who have played for an extended period of time and just simply cannot get the hang of flying, can get this crutch when the time comes. Not early players, just mid gamers who still can't get the hang of it. While I don't agree with him, as I think it can have strategic uses as well, I do read that as the original intent of the outfit.

Amazinite:
There are quite a few (closed) issues about reverse thrusters not being used in this situation or that situation, etc. etc. (#1170, #1377, #1583, #1838) I'd say this is practically a duplicate given that the general response on the topic of reverse thrusters has always been it's not something worth working on given how limited reverse thruster use currently is.

McNutty145:
[...]

PR #2721, opened 2017-07-06
Simple logic to determine if missile boat AI will reverse or turn around

tehhowch:
Use reverse thrust if its capability exceeds:
 - the target's ability to close the distance
 - the current closing rate of the target
 - the ship's forward thrust capability

Otherwise, turn around to use the better thrusters.

Ref #2716

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1520,10 +1520,23 @@ void AI::Attack(Ship &ship, Command &command, const Ship &target)
 	Point d = target.Position() - ship.Position();
 	if(shortestRange > 1000. && d.Length() < .5 * shortestRange)
 	{
-		command.SetTurn(TurnToward(ship, -d));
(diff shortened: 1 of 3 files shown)


tehhowch:
Here are 2 test missions to verify the behavior.

[npc_reversemissileboat.txt](https://github.com/endless-sky/endless-sky/files/1129418/npc_reversemissileboat.txt)

tehhowch:
I know that currently the base game content treats reverse thrust as nothing more than a handicap mechanic, but that does not mean that the reverse thrust implementation should be broken for mods as well.

Would it be better for this to go into a larger PR that addresses all movement scenarios to consider whether the ship is better off using forward thrust or turning around & reverse thrusting? Or is this PR's decision algorithm too simple to be useful? I recall seeing a mention that the missile boat AI may be reworked to use a circling behavior, but in the meantime this would serve as a drop - in fix for player escorts that have reverse thrusters.

PR #2722, opened 2017-07-07
bugfix: Restore follow-parent behavior for hostile NPCs which began patrolling

tehhowch:
In 4ab74b5bc8 hostile, non-staying NPCs which could not target any ships in a system were instructed to patrol it instead of jump away and then jump back. However, if the player left the system while remaining untargetable, the NPCs would get stuck landing if they were targeting a non-planet at the time of your jump.
This commit resolves the logic bug in MoveEscort that assumed the ship's stellar target was a landable planet by verifying that it is indeed a landable planet. If it is not, the ship will clear the target and resume tracking down the player.

This simple mission illustrates the issue: accept the job, then cloak and jump away once the bounty ship is targeting a non-landable object in the system.
[hostile_shouldfollow_stuck.txt](https://github.com/endless-sky/endless-sky/files/1129776/hostile_shouldfollow_stuck.txt)

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1081,8 +1081,12 @@ void AI::MoveEscort(Ship &ship, Command &command) const
 		Refuel(ship, command);
 	else if(!parentIsHere && !isStaying)
 	{
-		// Check whether the ship has a target system and is able to jump to it.
+		// Check whether the ship has a target system and is able to jump to it,
+		// and that the targeted stellar object can be landed on.
 		bool hasJump = (ship.GetTargetSystem() && ship.JumpFuel(ship.GetTargetSystem()));
+		if(ship.GetTargetStellar() && (!ship.GetTargetStellar()->GetPlanet()
+				|| !(ship.GetTargetStellar()->GetPlanet() && ship.GetTargetStellar()->GetPlanet()->CanLand(ship))))
+			ship.SetTargetStellar(nullptr);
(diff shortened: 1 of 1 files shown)


source/AI.cpp
```diff
@@ -1081,8 +1081,12 @@ void AI::MoveEscort(Ship &ship, Command &command) const
 		Refuel(ship, command);
 	else if(!parentIsHere && !isStaying)
 	{
-		// Check whether the ship has a target system and is able to jump to it.
+		// Check whether the ship has a target system and is able to jump to it,
+		// and that the targeted stellar object can be landed on.
 		bool hasJump = (ship.GetTargetSystem() && ship.JumpFuel(ship.GetTargetSystem()));
+		if(ship.GetTargetStellar() && (!ship.GetTargetStellar()->GetPlanet()
+				|| !(ship.GetTargetStellar()->GetPlanet() && ship.GetTargetStellar()->GetPlanet()->CanLand(ship))))
+			ship.SetTargetStellar(nullptr);
```
endless-sky: The second check of `ship.GetTargetStellar()->GetPlanet()` is redundant. This could simply be done with:

```
ship.GetTargetStellar()
&& !(ship.GetTargetStellar()->GetPlanet() && ship.GetTargetStellar()->GetPlanet()->CanLand(ship))
```

tehhowch:
Redundancy removed, thanks for the correction.

PR #2723, opened 2017-07-07
bugfix: NPC turrets not autofiring

tehhowch:
Extra ! in a boolean condition during the maxRange setup prevented some turrets from autofiring.
https://github.com/endless-sky/endless-sky/commit/957f6c077b78e0ea6a674ef19b83c9e6b20176fa#diff-80825e2b1f90be6b565f227d9ddb18ccL2107
As seen in the above commit, the condition did not change despite now being &&'d instead of ||'d.

In `master`, a hostile NPC which is given `fleeing` (no ship target) will not fire its turrets despite them clearly aiming at a hostile and taking no ammunition.
fail to fire
With this commit, the `fleeing` NPC will happily fire:
now can fire
Even homing turrets will work:
even grabstrikes

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2111,7 +2111,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 	for(const Hardpoint &weapon : ship.Weapons())
 		if(weapon.IsReady()
 				&& !(weapon.IsHoming() && currentTarget && weapon.GetOutfit()->Ammo())
-				&& !(secondary && !weapon.GetOutfit()->Icon())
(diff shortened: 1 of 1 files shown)


endless-sky:
It should actually be the other way around (d392a51). `(secondary || !weapon.GetOutfit()->Icon())` is equivalent to `!(!secondary && weapon.GetOutfit()->Icon())`, not `!(secondary && !weapon.GetOutfit()->Icon())`

tehhowch:
Confirmed fixed with d392a51 for both homing turrets (Grab-Strikes) and non-homing turrets (Blaster Turrets) when the ship has no official target.

PR #2725, opened 2017-07-07
Mission NPCs added to the escort display are drawn with hostile colors as appropriate.

tehhowch:
If a mission adds a hostile ship to the escort display so that the player can track its whereabouts, this ship will appear under the player's colors (green) on the radar and in the target reticle, rather than the orange-red hostile colors:
single fired turret

This commit prevents drawing hostile escorts as assumed friendlies:
hostile escort radar

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -49,6 +49,14 @@ color "overlay hostile hull" .5, .3, 0., .25
 color "overlay outfit scan" .5, .5, .5, .25
 color "overlay cargo scan" .7, .7, .7, .25
 
+color "radar player" .2, 1., 0., 0.
+color "radar friendly" .4, .6, 1., 0.
+color "radar unfriendly" .8, .8, .4, 0.
+color "radar hostile" 1., .6, .4, 0.
+color "radar inactive" .4, .4, .4, 0.
+color "radar special" 1., 1., 1., 0.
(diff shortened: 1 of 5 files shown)


endless-sky:
I wonder if it would also make sense for it to show up in red in the escort status icon. (An added complication would be that if you had an escort of the same model, they could be merged together into a single icon.)

tehhowch:
It would make sense, indeed. I'll add the bits to do that.

jafdy:
Red is also the "can't jump colour.

tehhowch:
I've got the escort stacks working fine, using the "hostile on radar / in target reticle" color
non-merged escorts

I noticed that the radar colors were not taken from a file, but instead declared statically. My past approach doesn't work (see the radar in the screenshot), as GameData::Colors() doesn't have any information about the actual colors in data/interfaces.txt when the colors object in the Radar.cpp namespace is being built. I'm going to leave those as statically defined but add the colors to the interface file, so that MZ can implement his preferred method for using user-defined colors in the Radar display.

tehhowch:
With radar colors restored:
default radar

tehhowch:
test mission: 
[escortMerging.txt](https://github.com/endless-sky/endless-sky/files/1131555/escortMerging.txt)

Issue #2274, opened 2017-03-17
Make the mining personality deploy fighters and drones


Amazinite:
Right now, fleets with the mining personality (or it might just be a side effect of human miners also having timid) will hold on to their fighters and drones until they're in combat. Suggested making them deploy their fighters and drones to help in mining asteroids (and picking them up if they have any cargo).

As a side note, it would also be cool if drones and fighters would transfer cargo that they picked up to the mothership whenever they get full.

zwparchman:
How should that be weighed against filling the mothership's cargo bay? Just fill it up? Don't drop cargo if the mothership can't hold it? I like the idea in general though.

LocalGod79:
i think that drones giving their cargo to their mothership might conflict with existing coding if their mothership is the player's flag...

endless-sky:
I'm pretty sure the only fighter or drone that comes with cargo space is the Boxwing, so this would have very limited usefulness.

KiLEdEnNis:
future fighters could have some cargo too

tehhowch:
As @LocalGod79 mentions, cargo transfer from carried to carrier is complicated by the situation where the carrier is the player's flagship. This is, however, the only case I see where the child->parent transfer would be unwanted, as it prevents the player from controlling what is in their cargo: 

Imagine if the player has a Bactrian, Carrier, or Aerie, with Boxwing fighters, and is doing some cargo + mining runs from A->B: 
1. The player buys `totalCargoSpace - flagshipCargoSpace` in commodities before takeoff
2. Then launches fighters while mining to benefit from improved mobility
3. Then recalls the fighters to jump to the next system after destroying all the high-value asteroids.

The game previously stored all the commodity cargo on the player's escorts, upon takeoff, but simple `carry->transfer` code would move the Boxwing's commodities into the flagship's cargo, possibly filling it up, and also leaving the largely uncontrollable boxwings with empty holds and no way to instruct to harvest mining flotsam.


All other cases do not impact the player, but allow for neat AI behavior, such as nimble mining fighters which hunt down minerals for a lumbering carrier ship, and also increase the likelihood that critical mission cargo is not stored on weak fighters with cargo space, but on their more durable carrier.

tehhowch:
Updating @zwparchman 's #1971 over the next day or so to better cover situational docking (cargo full, low health, ammo) and relaunching.
WIP branch https://github.com/endless-sky/endless-sky/compare/master...tehhowch:SituationalFighterDocking 
[...]

PR #2726, opened 2017-07-07
Carried ships assist in NPC mining operations

tehhowch:
The mining personality will now deploy any carried ships while actively mining.
These carried ships inherit the mining personality, so if they have cargo space free, they will mine independently.
After filling their cargo space, carried ships will assist their parent ship in destroying its target asteroid.

The restriction to carried ships is arbitrary, and could be removed, in which case any member of a fleet that is not actively mining or in combat, but has a fleet leader with the `mining` personality, would choose to assist that fleet leader in destroying its target asteroid. Removing the carried ship restriction would necessitate adding a check for a shipToAssist.

If the DoMining routine is updated to allow changing the target asteroid if it gets too far away from the ship that is mining, then the distance check used here can be dropped. Without the distance check, nimble fighters will engage target asteroids at distances too great for the parent to pick up the flotsam.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -472,7 +472,18 @@ void AI::Step(const PlayerInfo &player)
 		if(isPresent && personality.IsMining() && !target && !it->GetShipToAssist()
 				&& it->Cargo().Free() >= 5 && ++miningTime[&*it] < 3600 && ++minerCount < 9)
 		{
+			if(it->HasBays())
+				command |= Command::DEPLOY;
 			DoMining(*it, command);
+			it->SetCommands(command);
+			continue;
+		}
+		else if(isPresent && !target && it->CanBeCarried() && parent && miningTime[&*parent] < 3601
+				&& parent->GetTargetAsteroid() && parent->GetTargetAsteroid()->Position().Distance(parent->Position()) < 800.)
+		{
+			// Assist your parent in mining its nearby targeted asteroid.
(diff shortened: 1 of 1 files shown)


Amazinite:
Resolves #2274

tehhowch:
[mining_fleet.txt](https://github.com/endless-sky/endless-sky/files/1130210/mining_fleet.txt)
A simple mining fleet and test case for Imo Dep courtesy of @Kryes-Omega
It doesn't touch the bit in #2274 about transferring cargo

Amazinite:
The bit about transfering cargo was an afterthought. The main concern was deplyoing drones and fighters to help mine, which this does.

tehhowch:
It's pretty easy to add cargo transfer, actually.
Ship::Carry()
```
			heat += ship->heat;
			// If this ship collected anything in space, try to unload it into the parent's cargo hold.
			if(this->Cargo().Free() && !ship->Cargo().IsEmpty())
				ship->Cargo().TransferAll(&this->Cargo());
			return true;
```
(that doesn't prevent the "oops" of a fighter unloading into the player's flagship, which is almost certainly not desired)
Adding this to the PR would making mining NPCs with cargo-space fighters first mine separately (if they're in the independent minerCount), and then help their parent if/when they fill up, or if they aren't in the indepedent minerCount.

tehhowch:
(This PR is ready to merge, and deals only with making the mining personality use any fighters that it may have)

If requested, I will include some new mining fleets, but they will likely better be served in a separate PR.

Issue #595, opened 2016-01-05
Suggestion: Normal Maps for Ships


dplepage:
If we rendered normal maps for ships we could do dynamic lighting pretty easily, giving the ships a much more 3D look without requiring dozens of images - you just need an unshaded color image and a normal map, e.g.:
![color](https://cloud.githubusercontent.com/assets/107468/12117497/249061f6-b38f-11e5-86c6-6d6c7641b1f8.png) ![norm](https://cloud.githubusercontent.com/assets/107468/12117498/274fa19a-b38f-11e5-8320-93fa34508f88.png)

And then you can render the ship under any lighting:

![bactlit](https://cloud.githubusercontent.com/assets/107468/12117435/bcdfaa1c-b38e-11e5-8b80-d8834820e5d9.gif)

The main downside of this is that you need a normal map for every ship image, which means you can't just draw a ship in photoshop or whatever and drop it in. It looks like most people are using Blender, though, where it's really easy to render normal maps (it's how I made the one above, and I could probably make a small Blender plugin to automate that if there's interest).

It'd also be pretty easy to include multiple light sources, so that e.g. ships flying in binary/ternary star systems could be lit from multiple directions.

frameworld:
I'd love the possibility.

Have normal-map?
 * Yes? Render in the new way
 * No? Render in the old way

Want consistency?
 * create an old sprite from a new sprite with normal-map and put it into the cache folder.

dplepage:
I suspect it'll look kind of weird to see ships with static lighting flying alongside dynamically-lit ones, but I agree that making it optional is the best way to go, so that plugin authors can still make ships without normal maps. Possibly it should also be something you can disable in the settings.

Pointedstick:
That is reeeeeeeeally sweet. I like it!

Gkramme:
Yes, I was annoyed that the shadows were always pointed in one direcrion, no matter the ship's position relitive to the sun

knope:
+1

bobafetthotmail:
Agree. This is a rather clever trick.

+1 also for the blender plugin to automate this.

Would this increase graphic load or be annoying to keep working in a cross-platform game?

>"It looks like most people are using Blender, "

Afaik, old 2D ship images are being migrated to renderings, and it's mildly recommended in the wiki to use blender, then render it, then post-process it in GIMP for more grit and stuff.

dplepage:
[...]

PR #2729, opened 2017-07-07
Lighting effects (corrected)

857b:
Create a new shader "LightSpriteShader" which is based on SpriteShader. It can take as parameters the positions and the light emissions (power and color, fixed for now) of the system stars and the ambient light of the system.
If there are multiple stars in the system the light received by each pixel will be the sum of the incoming lights from the different stars.
The objects far from the stars will be displayed as if they receive less light.
The shader also compute the normal for each pixel to create a shadow on the parts which aren't facing the light sources. It assume that the object is spherical and is now only activated for the StellarObjects.
If the shader is not supported by the player hardware, the classic SpriteShader is used instead.
Corrected the many code-style errors spotted by @tehhowch in my previous PR.
Also raised the minimal luminosity so the ships will not be too dark if they are far from the stars.
light2


source/DrawList.cpp
```diff
@@ -43,85 +58,99 @@ void DrawList::SetCenter(const Point &center, const Point &centerVelocity)
...
+	return true;
+}
 
 // Draw all the items in this list.
-void DrawList::Draw() const
+void DrawList::Draw(bool lights) const
 {
 	bool showBlur = Preferences::Has("Render motion blur");
-	SpriteShader::Bind();
+	if(!lights || !LightSpriteShader::IsAvailable()){
```
tehhowch:
```
if(.....)
{
```

source/DrawList.cpp
```diff
@@ -43,85 +58,99 @@ void DrawList::SetCenter(const Point &center, const Point &centerVelocity)
...
+		SpriteShader::Bind();
+
+		for(const Item &item : items)
+			SpriteShader::Add(
+				item.tex0, item.tex1, item.position, item.transform,
+				item.Swizzle(), item.Clip(), item.Fade(), showBlur ? item.blur : nullptr);
+
+		SpriteShader::Unbind();
+		
+	}else{
```
tehhowch:
```
}
else
{
```

source/DrawList.cpp
```diff
@@ -214,14 +261,22 @@ float DrawList::Item::Clip() const
...
+	flags &= ~FADE_MASK;
+	flags |= static_cast<uint32_t>(cloak * FADE_FACTOR) << FADE_SHIFT;
+}
+
+bool DrawList::Item::Lighted() const{
+	return flags&LIGHT_MASK;
+}
+
+float DrawList::Item::NormalUse() const{
+	return ( (flags&NORMAL_MASK) >> NORMAL_SHIFT) / NORMAL_FACTOR;
```
tehhowch: `return ((flags & NORMAL_MASK) ....`

source/DrawList.cpp
```diff
@@ -214,14 +261,22 @@ float DrawList::Item::Clip() const
...
 void DrawList::Item::Cloak(double cloak)
 {
 	tex1 = SpriteSet::Get("ship/cloaked")->Texture();
-	flags &= 0xFF;
-	flags |= static_cast<uint32_t>(cloak * 256.f) << 8;
+	flags &= ~FADE_MASK;
+	flags |= static_cast<uint32_t>(cloak * FADE_FACTOR) << FADE_SHIFT;
+}
+
+bool DrawList::Item::Lighted() const{
```
tehhowch:
```
const
{
```

source/LightSpriteShader.cpp
```diff
@@ -0,0 +1,259 @@
...
+		lightPosI = shader.Uniform("lightPos");
+		lightEmitI = shader.Uniform("lightEmit");
+		lightAmbiantI = shader.Uniform("lightAmbiant");
+		angCoeffI = shader.Uniform("angCoeff");
+	
+		glUseProgram(shader.Object());
+		glUniform1i(shader.Uniform("tex0"), 0);
+		glUniform1i(shader.Uniform("tex1"), 1);
+		glUseProgram(0);
+	} catch(runtime_error& e){
```
tehhowch: } and statements on same lines

source/LightSpriteShader.cpp
```diff
@@ -0,0 +1,259 @@
...
+	glUseProgram(shader.Object());
+	glBindVertexArray(vao);
+	glActiveTexture(GL_TEXTURE0);
+	
+	GLfloat scale[2] = {2.f / Screen::Width(), -2.f / Screen::Height()};
+	glUniform2fv(scaleI, 1, scale);
+}
+
+
+void LightSpriteShader::Add(uint32_t tex0, uint32_t tex1, const float position[2], const float transform[4], int swizzle, float clip, float fade, const float blur[2], const float posGS[2], const float transformGS[2], int nbLight, const float lightAmbiant[3], const float *lightPos, const float *lightEmit, float angCoeff){
```
tehhowch:
```
)
{
```

source/LightSpriteShader.h
```diff
@@ -0,0 +1,46 @@
...
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+#ifndef LIGHT_SPRITE_SHADER_H_
+#define LIGHT_SPRITE_SHADER_H_
+
+class Sprite;
+class Point;
```
tehhowch: P < S

jafdy:
It is making the sprites on the right of the screen very hard to see.
I also have concerns over how this will interact with a ships own light (running lights or the Skylarks lighting).

tehhowch:
Based on the effect with projectiles, they will be drawn bright. This image is from the closed PR:

zoomed

jafdy:
That image is enough to make me say that this is not a good thing, I can only just see the ship if I zoom in and squint a little.

tehhowch:
Here is the current PR, with my ship sprite swapped to the Skylark (e.g. wrong hardpoints)
image

jafdy:
A reduced impact version of this could work well if all ships had running lights that would show up as normal, but they would need to light up part of the ship to avoid looking tacky. And that would just generate complexity, so unless a very complete version of this containing all lighting effects comes up I still don't think this is a goo idea.

tehhowch:
This PR harkens back to #595 in a way: if changes to global sprites are done, then light emitters need proper handling (e.g. the current game method of an animated sprite is not sufficient)

jafdy:
Except this assumes all ships are spheres. While it keeps the low entry requirement for new content it is not representative of the shape of a ship.

tehhowch:
No normal shading is applied to the ships yet, and as I mentioned in the other PR, the effect of the spherical normal shading is quite difficult to distinguish due to the current (very successful) method of predefined sprite orientation, prerendered spherical shadowing, and orbit-specific sprite rotation.

857b:
I raised the default ambient so the ship will be displayed with a 0.5 luminosity factor in the worst case.
I disabled the normals effect on the planets, the current system is pretty good even if it constrains the objects orientations. I activated this for the asteroids instead, but it look a bit strange since there are already some shadows in the sprites (but these are not correctly oriented).
If we want to add this normal effect to the ships we would have to create normal maps (as suggested in #595) but it's not the idea for now. This shader is more intended to create a different ambiance in each system.
I agree that we should avoid code duplication between the LightSpriteShader and the SpriteShader, but I thinks that we should keep the SpriteShader as it is for the menus and compatibility reason. But since the LightSpriteShader can't simply inherit from the SpriteShader (because it uses some local variables), the code structure has to be modified:
- we can remove the static aspect of the SpriteShader class and just keep an instance as static member / local variable. But all the call to the shader functions will have to be reformatted: SpriteShader::Draw(...) will become SpriteShader::Inst().Draw(...).
- or we keep the SpriteShader as an interface to a nested class which can be inherited.

image
In this screen shot I manually changed the star color to red to create a "dead star" ambiance:
image

tehhowch:
.5 is a much better luminosity than the .2:
here i am, far from the center
image

Even if we do not do normal mapping for ship shadows, you will need a solution that works for animated lights and ship light effects, as no matter how far one is from the star, those should have the same brightness.

857b:
I've added a light sprite to the Body class. From this light sprite the LightShader can determine which amount of light he has to add to the external lights. I've reused the Skylark light texture published in the #595 PR but I've raised the color saturation since I use a method a bit different because I add the internal light to the light color instead of adding it directly to the final color (in case of color swizzle). By the way, does anyone knows where I could find the .blend models mentioned in the #595 PR ?

Here is a screen shot of a skylark with his light texture (the ambient light has been decreased to view the ship's own lights)
skylark light

tehhowch:
Can that work with animated lights?  @jafdy is adding running lights to a large number of human craft, and @Elyssaen has some seriously awesome Trin ships with artwork by @comnom.

857b:
The light sprite can be animated as long as it has the same number of frame
than the main sprite. But it could have it's own animation cycle if we
duplicate the Body's attributes.

MessyMix:
Now that #2741 may (or may not) be a thing, I think we can revisit this idea. :)

Personally, I like it a lot, although we don't want to get too dark. The running lights should pair with this beautifully.

jafdy:
The running lights would get dimmed with everything else.

endless-sky:
I think this falls on the wrong side of the balance between what looks "realistic" vs. what actually looks good and makes the game fun to play. If everything on screen just gets murky when you're far from the star, that's just going to be annoying.

Issue #2746, opened 2017-07-12
Placement of mission cargo/passengers in the player's fleet


Amazinite:
A player on the Discord was attempting to bomb Zenith by flying into the system cloaked and landing on the planet while leaving their escorts holding position in a different system. This led to quite a bit of confusion when the player recieved no conversation for the next mission and no pop-up dialog explaining why. The player could not have parked their ships without landing again because the previous mission forces the player to launch.

Another player expressed having the same issue in the past during the Checkmate campaign when they had to bring cargo out of Pug occupied space, only to find out half an hour later that it was in their non-jump drive equipped escort the entire time.

Suggested that mission cargo be placed in the flagship first before it is placed in escorts as to avoid such situations where the player does not have their escorts in the destination system, either through leaving them behind or being unable to bring them.

tehhowch:
I'm thinking this should be a preference setting, because the majority of times I want my escorts holding that cargo.

This is already done for passengers, so configuring it for cargo will not be too bad. (I also had that issue with Checkmate)

tehhowch:
![image](https://user-images.githubusercontent.com/20871346/28123214-00eca91a-66e6-11e7-8140-7ff94bef6c23.png)

Do note that this will include bog-standard mission cargo, such as that 148 tons of paper towels you said you'd deliver to Dancer.

jlnxr:
I've run into this problem, although I've also run into situations where I want to keep passengers and cargo off my flagship, to make room for plundering and the largest boarding party possible. I think the ideal solution would be to put things on the flagship first, but have some mechanism for moving things from ship to ship as you see fit. That way  I if it's a surprise launch, it's  a on the flagship, but you aren't stuck with that for your fish head/paper towel/ industrial waste jobs. Not sure if that would be easily possible though

tehhowch:
@jlnxr #2747 would do exactly that for cargo, if the player turns on that preference.
Any cargo from a one-time-only mission (e.g. not repeatable jobs), a mission with `autosave` (dangerous jobs, perhaps), that is illegal, or a mission that is `stealth` would get placed on your flagship instead of escorts (until you have no room). After your escorts run out of room, then what's left gets put on your flagship (normal behavior)

Amazinite:
#4315/#4342 effectively solve the problem of this issue by making it so that the player knows that the cargo/passengers are on a different ship.

PR #2747, opened 2017-07-12
enhancement: Store "critical" cargo in flagship first

tehhowch:
**Refs #2746**
Introduce a preference setting "Special cargo fills flagship first" that
 - stores cargo from missions with 'autosave'
 - stores cargo from missions marked 'priority'
 - stores cargo from one-time-only missions
 - stores cargo with illegal fines
 - stores cargo from missions with 'stealth'

into the flagship, and then if/when full, into your available escorts.

The default behavior of filling available escort holds before the flagship is preserved when the preference is off.

image

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -28,6 +28,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Person.h"
 #include "Planet.h"
 #include "Politics.h"
+#include "Preferences.h"
 #include "Random.h"
 #include "SavedGame.h"
 #include "Ship.h"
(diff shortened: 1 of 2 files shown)


tehhowch:
If players use this setting, it would prevent situations such as having only one jump drive (possibly quite common) on the FW: Checkmate branch, and also having escorts (very common), and jumping out of Pug space to set up the 'graviton reflector', only to find that the 'graviton reflector' is held by a linklocked escort and not your JD flagship.

The boolean checks prevent this from offering for run-of-the-mill missions such as cargo deliveries from A to B. I chose to include illegal cargo and stealth mission cargo as the player can generally outmanuever surveillance ships, while their escorts (if so foolish as to try these missions with escorts) just think "how do i get to the flagship?".

AdamMil:
Is there a downside to disabling this feature? If not, maybe no option is needed and it should simply become the new behavior.

tehhowch:
It can result in the flagship having less space than without the preference active. For missions which trigger the checks but aren't so critical (e.g. your escorts can route to you), you can end up with some special cargo in your bay when it could be safely carried in your escort instead.

It's possible that this should indeed be default behavior, and able to be turned off, rather than opt-in behavior. I'll change it if @endless-sky agrees (or he'll merge it and make the change himself), otherwise I'll keep it as-is.

Jeremy-E-Johnson:
My biggest concern with this PR is in whether or not a player could figure out exactly what this setting means in game. "special cargo" is pretty vague. A tool-tip would probably solve this problem, but, as far as I know, there are no tool-tips in the settings so that may be a bit out of scope.

I am also on the side of making this the default behavior. It prevents some weird issues with being unable to complete missions due to your escorts not keeping up. Maybe if such failure to complete a mission at least brought up a helpful message with the gist of "your escort with the critical cargo is not here" then I could see keeping the default as is.

endless-sky:
One of my basic principles for the game's design is that the rules for situations like this need to be straightforward and predictable enough that it won't look to the player like the game is doing things randomly. For example, "Cargo gets loaded into your flagship last, so it has space free for plunder" is a simple rule and makes sense. "Mission cargo from one of these five classes of missions gets loaded into your flagship first" is not a simple rule (especially because some of those mission classes are not visible to the player).

Note that not all special missions have "autosave"; rather, autosave is used periodically in special mission strings at points where it's guaranteed that the player hasn't irredeemably messed things up.

I think a better approach to the problem of a mission not completeing because the cargo is in another ship would just be to add "on visit" conditions to most jobs and missions saying something like, "You have arrived on <planet>, but the [passengers / cargo] is on another ship that is not here yet."

PR #2749, opened 2017-07-12
Allow events to modify a fleet's personality

Hadron1776:
Add the ability for events to modify a fleet's personality without replacing the fleet's definition. 

With this PR, if you want an event to change a fleet's personality, you can simply put
`event SomeEvent`
`    fleet SomeFleet`
`        personality add <whatever> `
or `personality remove <whatever>` instead of rewriting the personality definition.

### DIFF
--- a/source/Fleet.cpp
+++ b/source/Fleet.cpp
@@ -48,6 +48,8 @@ void Fleet::Load(const DataNode &node)
 		// clearing the rest of the existing definition.
 		bool add = (key == "add" && hasValue && child.Token(1) == "variant");
 		bool remove = (key == "remove" && hasValue && child.Token(1) == "variant");
+		bool addPersonality = (key == "add" && hasValue && child.Token(1) == "personality");
+		bool removePersonality = (key == "remove" && hasValue && child.Token(1) == "personality");
 		
 		if(key == "government" && hasValue)
 			government = GameData::Governments().Get(child.Token(1));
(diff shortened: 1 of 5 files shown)


tehhowch:
i think the preferred add/remove syntax is add/remove before the thing, and 
```
add personality <value1> ... <valueN>
add personality
    <value1> <value2>
    <value3>
    <confusion> <N>
remove personality <value1> .... <valueN>
```
A simpler method (well, fewer file changes) is probably to load the new personality into a temporary Personality object
```
Personality changes;
changes.Load(child);
```
and then define what it means to add or subtract personalities (`+` or `+=` means |= the flags, while `-` or `-=` means &= ~ the flags, and both adding and removing returns the confusion multiplier of the second) in Personality.cpp

Finally, in the fleet personality section, use add/remove to inform the operator (e.g. add ->  `personality += changes`, and remove -> `personality -= changes`

This would let you avoid needing to change calls to personality.Load, at least.

Hadron1776:
So you're saying I should do an operator overload?

tehhowch:
Instead of changing how Personality::Load is called, yes.
(right now you can't do person1 + person2, since the game what the definition of adding means for the Personality class.)
Then you only have to handle add/remove logic in places that make sense, such as outside the personality base class.
You won't even have to strip the add/remove/personality token from the child, as Personality::Parse only does something when it can do something with that token (e.g. `TOKEN.find("add")` will be `TOKEN.end()`)

PR #2751, opened 2017-07-12
Allow events to easily alter the personality of a fleet

Hadron1776:
This PR adds `add personality` and `remove personality`, to be used when re-defining a fleet in an event. These parameters can be used to add or remove individual flags from a fleet's definition without re-writing the whole `fleet` block.


Hadron1776:
I'm pretty sure you can pass a negative "confusion" to `add personality` to reduce the confusion as well...

endless-sky:
After thinking about this for a while, I decided that letting the user "add" or "remove" confusion would just be too error-prone. So, I implemented this in a different way (561d875) and also added some more specific error messages to Fleet and Personality.

Issue #2754, opened 2017-07-13
Enhancement Request: Better Cargo Packing


marphod:
I'm not sure what the cargo packing algorithm is, but it does not seem to be one that optimizes for the storage of large items.  At a guess, it places items in the first available cargo hold with enough space.

A better packing algorithm would be:
For items to store (largest to smallest) 
If there is a cargo hold with exactly item.size space left, use that
otherwise, place in hold with currently the most available space

This will avoid cases where my 160 ton cargo ships had a 90 ton object and random trade goods, rather than the 150 ton objects i wanted them to have.

tehhowch:
<details><summary>Current packing is done as mission cargo -> spare outfits -> commodities
</summary>


PlayerInfo::TakeOff (PlayerInfo.cpp#L1168)
```
for(const shared_ptr<Ship> &ship : ships)
		if(!ship->IsParked() && ship->GetSystem() == system && !ship->IsDisabled())
		{
			ship->Recharge(hasSpaceport);
			if(ship != flagship)
			{
				ship->Cargo().SetBunks(ship->Attributes().Get("bunks") - ship->RequiredCrew());
				cargo.TransferAll(&ship->Cargo());
			}
			else
			{
				// Your flagship takes first priority for passengers but last for cargo.
				ship->Cargo().SetBunks(ship->Attributes().Get("bunks") - ship->Crew());
				for(const auto &it : cargo.PassengerList())
					cargo.TransferPassengers(it.first, it.second, &ship->Cargo());
			}
		}
	// Load up your flagship last, so that it will have space free for any
	// plunder that you happen to acquire.
	cargo.TransferAll(&flagship->Cargo());
```
The relevant bits TransferAll routine (CargoHold.cpp#L457)
```
	for(const auto &it : passengers)
		TransferPassengers(it.first, it.second, to);
	// Handle zero-size mission cargo correctly. For mission cargo, having an
	// entry in the map, but with a size of zero, is different than not having
	// an entry at all.
	auto mit = missionCargo.begin();
	while(mit != missionCargo.end())
	{
		Transfer(mit->first, mit->second, to);
		if(!mit->second)
			mit = missionCargo.erase(mit);
		else
			++mit;
	}
	for(const auto &it : outfits)
		Transfer(it.first, it.second, to);
	for(const auto &it : commodities)
		Transfer(it.first, it.second, to);
```
and lastly, Transfer (outfits version):
```
{
	double mass = outfit->Get("mass");
	
[...]

PR #2757, opened 2017-07-13
Transfer outfits in reverse size order

tehhowch:
When performing a TransferAll(), first obtain a descending-by-mass sorted vector of the outfits to be transferred, and then move outfits by iterating over that vector. If an outfit does not have a specified mass, outfit->Get("mass") will return 0 and there is no limit to the number that can be in a given cargo hold.

This means that the largest outfits that can fit in escort holds are guaranteed to be in escort holds.
edit: this is a First-Fit-Decreasing bin-packing implementation, minus the bits where new bins can be added and the bins are all the same size.

Photo proof:
My cargo while landed:
total_cargo
1 Jump Drive, 1 Large Heat Shunt, 2 System Core (Medium), 1 System Core (Small).
This order is also the order of the items in the `outfits` variable used by CargoHold.

To takeoff, on `master`, this list is iterated left-to-right, beginning in my 2nd escort (the Solifuge), until it can no longer hold the next outfit:
master_2ndescort

My other ships have insufficient space to hold the outfits left, so my flagship gets the remaining outfits:
master_flagship

With this PR, the System Core (Medium) are added first:
pr_2ndescort

And my flagship again receives the rest:
pr_flagship

But by maximizing the storage of large items in my escorts, I now have more free space in my flagship's hold.

Refs #2754

### DIFF
--- a/source/CargoHold.cpp
+++ b/source/CargoHold.cpp
@@ -468,8 +468,9 @@ void CargoHold::TransferAll(CargoHold *to)
 		else
 			++mit;
 	}
-	for(const auto &it : outfits)
-		Transfer(it.first, it.second, to);
+	const std::vector<const Outfit *> outfitOrder = GetOutfitOrder();
+	for(const auto &outfit : outfitOrder)
+		Transfer(outfit, outfits[outfit], to);
 	for(const auto &it : commodities)
 		Transfer(it.first, it.second, to);
 }
@@ -569,4 +570,28 @@ int CargoHold::IllegalCargoFine() const
 		worst = max(worst, fine);
 	}
 	return worst;
+}
+
+
+
+// Retrieve vector of pointers to the outfits, sorted descending by size.
+const vector<const Outfit *> CargoHold::GetOutfitOrder() const
+{
+	vector<const Outfit *> sortedOutfits;
+	if(!outfits.size())
+		return sortedOutfits;
+	
+	for(const auto &outfit : outfits)
+		sortedOutfits.emplace_back(outfit.first);
+	sort(sortedOutfits.begin(), sortedOutfits.end(),
+		[] (const Outfit *lhs, const Outfit *rhs)
+		{
+			if(lhs->Get("mass") == rhs->Get("mass"))
+				return lhs->Name() < rhs->Name();
+			else
+				return lhs->Get("mass") > rhs->Get("mass");
+		}
+	);
+	
+	return sortedOutfits;
 }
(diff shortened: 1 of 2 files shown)


source/CargoHold.cpp
```diff
@@ -570,3 +571,27 @@ int CargoHold::IllegalCargoFine() const
...
+	if(!outfits.size())
+		return sortedOutfits;
+	
+	for(const auto &outfit : outfits)
+		sortedOutfits.emplace_back(outfit.first);
+	sort(sortedOutfits.begin(), sortedOutfits.end(),
+		[] (const Outfit *lhs, const Outfit *rhs)
+		{
+			if(lhs->Get("mass") == rhs->Get("mass"))
+				return lhs->Name() < rhs->Name();
```
endless-sky: Note that the comparison function given to sort() only has to be a weak ordering, so this step isn't necessary; if two outfits have the same mass it doesn't matter which one comes first.

endless-sky:
This is essentially a [bin-packing problem](https://en.wikipedia.org/wiki/Bin_packing_problem). Adding the largest objects first will help in some situations, but will cause worse results in others.

tehhowch:
I got (too) curious about if we could decide if one is better, and wrote a little [comparison](https://gist.github.com/tehhowch/8460ad86e300d78dabf3f8e57f589267) to find out.


<details><summary>The result: sorting first is better in 91-94% of cases, depending on the number of ships and outfits.</summary>

It implements the current packing system for a random collection of outfits (with outfit sizes based on the frequency of outfit sizes in the base game) and a random number of ships, each of which has a different cargo size. I have arbitrarily biased the cargo size of the ships to favor smaller cargo sizes, such as what would be found in a battle fleet where Outfits Expansions are used.
After the ships' cargo sizes are set, I remove the most recently added outfits until the total outfit mass is less than the total cargo capacity (unless there is only a single ship, in which no outfits are removed). This ensures that only a handful of outfits will be "left over" after the filling method completes.
For each pairing of an outfit set and fleet, I used the current filling method of "emplace outfits as they are naturally encountered", and recorded the remaining mass of outfits that did not fit in the cargo holds of the ship set.
After sorting the outfit list by size, decreasing, the filling method is called again and the remaining mass of outfits is recorded. The "savings" is defined as the leftover amount from the current method - the leftover amount from the method used by this PR. A positive number means that this PR more efficiently packed the same items in the same space than the current method, and a negative number means it failed to pack as efficiently as the current method.


There are a few booleans in the main() function that can be turned on to enable printing of the packing details for each ship in each iteration (I don't recommend doing many simulations with those on).

Build instructions: ubuntu 16.04, `g++ -std=c++11 -O3 *.cpp && ./a.out`
In addition to the .gist, you will need the Random.cpp and Random.h files from the ES repo in the same directory as the gist.
</details>

PR #2758, opened 2017-07-13
Add support for custom "demand tribute" messages

Hadron1776:
This allows modders to override the various messages a planet gives when you demand tribute:
`tribute ignored message` -> You demand tribute from a planet when your CR isn't high enough
`tribute relinquished message` -> You release your tribute from a planet
`tribute in progress message` -> You demand tribute when you're in the middle of fighting the defense fleet
`tribute battle message` -> The planet takes your demand for tribute seriously and sends its defense fleet
`tribute surrendered message` -> The planet surrenders and (hopefully) tells you how many credits you'll be getting*

You can use these tokens:
`<tribute>` -> shows up as " (_whatever the planet's tribute payment is_) credits"
`<first>` -> The player's first name
`<last>` -> The player's last name
`<origin>` -> The planet you're demanding tribute from
`<ship>` -> The name of the player's flagship

Edit: Forgot to mention that these are defined per government. If these hails are left out, they get default values.

*_The default message does this, but the_ `<tribute>` _token is not explicitly required. As such, modders may or may not actually have their factions tell you how much you're getting._


tehhowch:
You should coordinate with @Amazinite to include some default Pirate planet tribute hails. This way the game can ship with examples of properly extending the hails on a per-government basis.

\cc @Disiuze : tribute+

Elyssaen:
You are a gentleman and a scholar. This is a great little enhancement.

A picky point: I'm not sure about some of the names. For me, `denied` and `accepted` are confusing ones, as what you mean is they they deny or accept your _challenge_, not that they deny or accept giving you tribute. I think I'd prefer something more like `ignored` and `resist`.

tehhowch:
I do agree with Elyssaen here:
TributeAccepted
TributeGranted
those are pretty darn similar.
Perhaps 
`tribute granted` -> `tribute succeeded`
With that changed, then the relationship between `tribute accepted` and `tribute denied` is more clear (to me at least)

Hadron1776:
What about:
`tribute accepted hail` -> `tribute battle message`
`tribute granted hail` -> `tribute surrendered message`
`tribute denied hail` -> `tribute ignored message`?

Elyssaen:
I'd go for that. I was just thinking `surrendered` would be good wording as well.

source/Government.h
```diff
@@ -49,6 +49,13 @@ class Government {
 	// Get the color to use for displaying this government on the map.
 	const Color &GetColor() const;
 	
+	// Get the government's tribute hails
+	const Phrase *TributeIgnoredHail() const;
```
tehhowch: It may be worthwhile to take a page from the Government::GetHail() book, and use a single method with an int type specifier to determine which hail to return.

e.g. 
```
Government::GetTributeHail(int tributeStatus)

static const tribute[5] = { *tributeIgnoredHail, ..... *tributeSurrenderedHail }
return tribute[tributeStatus]->Get()
```
(or however)

source/Government.cpp
```diff
@@ -169,37 +169,31 @@ const Color &Government::GetColor() const
...
-{
-	return tributeCombatHail;
-}
-
-
-
-const Phrase *Government::TributeSurrenderedHail() const
-{
-	return tributeSurrenderedHail;
+	switch(condition)
```
tehhowch: fun fact: this would be the only use of `switch` in all of es.
I was thinking of something like what is used for colors:
https://github.com/endless-sky/endless-sky/blob/89259410371ab124580751c98fd1e332b35e6a58/source/Radar.cpp#L89

source/Government.cpp
```diff
@@ -169,31 +169,17 @@ const Color &Government::GetColor() const
...
-	}
+	static const vector<const Phrase *> phrases = {
+	tributePresentHail,
+	tributeIgnoredHail,
+	tributeInProgressHail,
+	tributeCombatHail,
+	tributeSurrenderedHail
+	};
+	
+	return phrases[condition];
```
tehhowch: could you make this `return phrases[condition]->Get()`? then you change it to a std::string return value, and position it in the `.h` and `.cpp` right next to Government::GetHail() for supreme logical consistency :1st_place_medal: 

You may also want bounds checking on the input, though in your current code there's no variable inputs to this functions caller so you are pretty secure in knowing you are in-bounds.

source/Government.cpp
```diff
@@ -223,6 +243,25 @@ string Government::GetHail(bool isDisabled) const
 
 
 
+// Get the government's tribute-related hails
+const string *Government::GetTributeHail(int condition) const
+{
+
+	static const vector<const Phrase *> phrases = {
+	tributePresentHail,
```
tehhowch: Indent once

source/Planet.cpp
```diff
@@ -472,10 +473,23 @@ void Planet::Bribe(bool fullAccess) const
...
+	subs["<tribute>"] = Format::Number(tribute) + " credits";
+	subs["<first>"] = player.FirstName();
+	subs["<last>"] = player.LastName();
+	subs["<origin>"] = Name();
+	subs["<ship>"] = player.Flagship()->Name();
+	static const string tributeHail = "";
 	if(player.GetCondition("tribute: " + name))
-		return "We are already paying you as much as we can afford.";
+	{
+		tributeHail = GetGovernment()->GetTributeHail(0);
```
tehhowch: MZ generally defines a constant value and then refers to that constant, rather than the value of the constant, to improve code readability.
So instead of `(0)`, you'd do `(TRIBUTE_PRESENT)` here, and before your `subs` declaration you'd declare the values of your constants:
```
static const int TRIBUTE_PRESENT = 0x01
static const int TRIBUTE_IGNORED = 0x02
...
```

solardawning:
Honestly, while it's nice to be able to customize them per government, it'd be even better to also be able to override them per planet (probably in the map data, then?).

tehhowch:
@solardawning I think that would be a 2nd PR to add a new feature, rather than de-static-ify an existing feature.

You're not wrong, though :+1:

Hadron1776:
@solardawning Maybe later. 
But I'm thinking, instead of forcing modders to define custom tribute messages for _every_ planet (because there are a lot of them), maybe take the planet's tribute messages, if any are missing, then pull them from the government, and if THOSE are missing, there would be defaults.

tehhowch:
I would suggest that method as well for a followup PR.

Here's a simple txt file to drop into a plugin (or es/data...) that rewrites the message that the Syndicate gives when you demand tribute. All the other phrases are the default.
[custom_tribute_hail.txt](https://github.com/endless-sky/endless-sky/files/1146981/custom_tribute_hail.txt)

Amazinite:
>`tribute present message` -> You demand tribute from a planet which is already paying you tribute

Shouldn't this be entitled `tribute relinquished message` given that you can't demand for tribute again once you've already conquered the planet? The `Demand Tribute` button changes to a `Relinquish Tribute` button.

And why stop at tribute hails? What about when you normally hail a planet (currently `You are cleared to land <ship>.`) or when you need to pay to land (currently `If you want to land here, it'll cost you <bribe>` and when the bribe is offered `It's a pleasure doing business with you.` or denied `Sorry, but you don't have enough money to be worth my while.`) or when landing access is restricted (`I'm afraid we can't permit you to land here.` and `I don't want your money.`).

`planet hailed message` for normal hails
`planet restricted message` for when you hail and can't land
`planet bribe message` for when you hail and can land after bribing
`planet bribed message` for after you bribe the planet
`planet denied message` for after you try to bribe but don't have the money for it

Not sure what I would call the message for when you try to bribe and unbribable planet.

Amazinite:
How do these sound for pirate tribute messages?
`ignored`: We're not going to waste our time on a nobody like you.
`battle`: You `<ship>` is about to become an empty husk floating through space!
`in progress`: Did you think we'd go that easily? We still have plenty of fight left in us!
`surrender`: We surrender to you, great Warlord <last>. You will receive `<tribute>` per day from our spoils for your greatness.
`relinquished`: We are under the thumb of a warlord no more! Anarchy forever!

Amazinite:
And for the Quarg. The only Quarg planet that you can currently demand tribute from is Humanika, which is mostly inhabited by humans. As such, the `battle` message assumes that you're demanding tribute on Humanika. This could be changed so that it works for all future Quarg planets (or ringworlds) that you can demand tribute from, then making per-planet tribute messages an interesting addition.

`ignored`: We do not play war games with lesser species. Move along now.
`battle`: We do this only because the humans that we harbor ask that we do so. Prepare to pay for your mistake.
`in progress`: You have earned nothing for you have yet to defeat our defense fleet. 
`surrender`: We will play this game no longer. You shall be given `<tribute>` per day and in return you must leave us be.
`relinquished`: You have made a wise choice. You shall no longer receive credits from this planet.

data/hails.txt
```diff
@@ -3056,3 +3056,23 @@ phrase "tribute battle message"
...
 	word
 		"We surrender. We will pay you <tribute> per day to leave us alone."
+		
+phrase "pirate tribute ignored"
+	word
+		"We're not going to waste our time on a nobody like you."
+		
+phrase "pirate tribute battle"
+	word
+		"<last>, you have taunted us for the last time! Prepare to die!"
```
Amazinite: Define taunting, and what if the player has never "taunted" pirates before? This just doesn't really make sense to me.

data/hails.txt
```diff
@@ -3056,3 +3056,23 @@ phrase "tribute battle message"
...
+		
+phrase "pirate tribute in progress"
+	word
+		"The warriors of <origin> will not give up so easily!"
```
Amazinite: I've never seen pirates call themselves "warriors."

data/hails.txt
Hadron1776: To be fair, if the player has never taunted pirates before, then he/she has only done it once, and first = last, in that case.

Amazinite: The wording really seems to imply there has been previous taunting involved, though.

source/Government.cpp
```diff
@@ -111,15 +111,35 @@ void Government::Load(const DataNode &node)
...
-	// Default to the standard disabled hail messages.
+	// Default to the standard disabled and tribute hail messages.
 	if(!friendlyDisabledHail)
 		friendlyDisabledHail = GameData::Phrases().Get("friendly disabled");
 	if(!hostileDisabledHail)
 		hostileDisabledHail = GameData::Phrases().Get("hostile disabled");
+	if(!tributeIgnoredHail)
+		tributeIgnoredHail = GameData::Phrases().Get("tribute ignored message");
+	if(!tributeCombatHail)
+		tributeCombatHail = GameData::Phrases().Get("tribute battle message");
```
Elyssaen: Can we keep battle/combat consistent?

source/Government.cpp
```diff
@@ -223,6 +243,25 @@ string Government::GetHail(bool isDisabled) const
 
 
 
+// Get the government's tribute-related hails
```
Elyssaen: Petty point: full stop please!

source/Planet.cpp
```diff
@@ -472,10 +472,22 @@ void Planet::Bribe(bool fullAccess) const
 // Demand tribute, and get the planet's response.
 string Planet::DemandTribute(PlayerInfo &player) const
 {
-	if(player.GetCondition("tribute: " + name))
-		return "We are already paying you as much as we can afford.";
+	static const int TRIBUTE_IGNORED = 1;
```
Elyssaen: I don't know enough to be sure, but should these be defined in `namespace` at the top of the file instead? Arguably they should also be `static constexpr`, but I don't think this is generally used in the code right now.

data/hails.txt
```diff
@@ -3035,3 +3035,44 @@ phrase "wormhole hail"
...
+	word
+		"Your <ship> is about to become an empty husk floating through space!"
+		
+phrase "pirate tribute in progress"
+	word
+		"Did you think we'd go that easily? We still have plenty of fight left in us!"
+		
+phrase "pirate tribute surrender"
+	word
+		"We surrender to you, great Warlord <last>. You will receive <tribute> per day from our spoils for your greatness."
```
Elyssaen: I don't think warlord should be capitalised here. If this was an official 'title' amongst pirates, then perhaps it'd be Warlord. As it is I think it's just warlord.

source/Planet.cpp
tehhowch: They are not used by any other function in Planet.cpp, so I would say they are fine where they are. They will get set in the first call, and any subsequent calls to DemandTribute() will reuse them without re-setting them.

source/Government.cpp
tehhowch: All comments should be sentences.

source/Planet.cpp
Hadron1776: There's also one in HailPanel.cpp

endless-sky:
As I've mentioned before, tribute is really just meant to be a tacked-on feature for bored end-game players, so the way it's implemented is intentionally simple. This seems like a whole lot of added complexity for a relatively minor benefit.

qntf:
While tribute may be a "tacked-on" feature, I still think this change adds value without really adding much complexity. If anything, it just adds more variety to the same old hail messages.

Hadron1776:
> `planet bribe message` for when you hail and can land after bribing
`planet bribed message` for after you bribe the planet

That has the same issue as with `accepted` and `granted` - they're too similar. Also, typos happen.

tehhowch:
The base impact is minimal at worst, and greatly enhances the worldbuilding __[mods](https://github.com/Disiuze/TributePlus/)__ can do. Aliens should probably respond to tribute in a different manner than the Republic does, after all.

This PR also helps move more fixed strings out of `source` and into `data`, enhancing any future localization work.
Default inheritance of the current values means we do not have to change or add __anything__ to  `governments.txt` -- only those governments which are special will merit an addition, and only for bits that should not be default.

Hadron1776:
> And these are just singular hails that could be attached with this PR. If this PR gets merged then I could do multiple other hails so that demanding tribute doesn't give the same messages each time.

^ This is another strong point of this PR - the "demand tribute" messages don't have to be the same every time.

Amazinite:
Y tho

Hadron1776:
Mainly because I'm not sure how to resolve the merge conflicts that commits 1ce173d and 7675eb7 made...

endless-sky:
Rather than this approach, I think it'd be cleaner and more extensible to have Government store a single `map<string, const Phrase *>` for all sorts of different messages. The syntax could then be something like:

```
message "tribute challenge rejected" "pirate tribute challenge rejected"
```

and the parsing would be as simple as:

```C++
if(key == "message" && child.Size() >= 3)
    message[child.Token(1)] = GameData::Phrases().Get(child.Token(2));
```

A `Government::GetMessage(const string &type)` function could store a static map defining the Phrase to use for anything not custom-defined in the government spec. And, it could easily be expanded to bribes, requests for help, etc.

Issue #2759, opened 2017-07-13
Enhancement: Make ramscoops and solar panels perform differently depending on the type of star system


Jeremy-E-Johnson:
Currently the amount of fuel gained with a ramscoop only depends on your ramscoop value and proximity to the center of the system. The same is true of solar panels. I propose that this be modified to take into account the number of stars in a system and their type. (note that #2508 is somewhat relevant)

I see there being three major decisions to be made in the implementation:

- What exactly does this modification effect?
- What are the values for the different star types?
- What is the schema for dealing with multiple stars in a single system?

Handling each respectively:

Currently there is a value called "scale" which depends linearly on a ship's distance from the center of the system that determines how effective solar panels and ramscoops are. 
`double scale = .2 + 1.8 / (.001 * position.Length() + 1);`
This could be modified directly to depend on the system, in which case it is useful as a value that describes how much the stars of the system are effecting a ship.
Or modifiers can be implemented separately for ramscoops and solar panels. This has the added value that stars do not have to emit light in direct correlation to how much gas they emit. (I see this being most notable in the supernova systems where the stars are small but there are plenty of gas clouds)

I have no clue how to appropriately choose values for star types, but further research into actual stellar emissions may provide a good starting place. They would probably be centered on making ramscoops and solar panels work pretty much the same as they do now in single star systems where the star is similar to Sol.

Dealing with multiple stars can be handled in a couple ways. We could just add their modifiers, but I feel like this may be a bit much. My gut tells me that a binary star system with two stars like Sol should probably have a modifier more like 1.5 than 2.
The only thing that comes to mind is to order the stars by their modifiers and from largest to smallest sum up the values times 1/n where n represents where in the ordered list the modifier resided. (for example [1, 3., .2] -> 3 + 1*(1/2) + .2*(1/3) )

MessyMix:
I thought this was a duplicate, but turns out they didn't PR this.

It was addressed in https://github.com/endless-sky/endless-sky/issues/2508#issuecomment-300492186. This, and see @Elyssaen's comment.

Elyssaen:
Hey. That issue (#2508) led to a PR (#2618), but it's about scaling the heat generated by solar collectors. @Jeremy-E-Johnson's idea has been talked about too, but I think MZ would suggest it's a little too 'simulationist' and doesn't really add much fun. 

(Note that PR #2618 hasn't had any attention either: if we can't answer the question "How does this make the game more fun?", it's not likely to be a high priority.)

[...]

PR #2509, opened 2017-05-10
enhancement: Let System class report StellarObject counts

tehhowch:
While not immediately useful, these seem like basic quantities that can be useful for more advanced things with the game and mission engines.

For example, if a system is able to report that it has no stars, ramscoop and solar collection could be turned off entirely (except for perhaps some tiny background amount), or optical tracking can be degraded due to lack of light... If a system has multiple stars, the collection rate of ramscoops / solar collectors can scale appropriately due to the higher luminosity or solar wind, optical tracking becomes more effective while radar and infrared tracking slightly decrease (due to higher background levels).
Coded example usage: https://github.com/tehhowch/endless-sky/commit/f3f26a6abda14fd524b57ee1106974f04f24797b

For missions, (if the current filtering system supports it) a system with two planets could be randomly picked to have a civil war, or a system with two stars can be randomly chosen for a special science expedition, or some warring faction can target a random nearby systems having space stations...

PR #2765, opened 2017-07-15
Added system dependency for ramscoop and solar panel effectiveness.

Jeremy-E-Johnson:
This is in reference to issue #2759. The solution presented is pretty much exactly as I had outlined in the issue.  ~~(Including how modifiers are added together for multiple star, star systems.)~~

EDIT: Modifiers are now summed without any weighting.

The modifiers that are implemented in this pull request are just numbers that I came up with after about an hour on Wikipedia. That is to say that they try to be a little realistic and hopefully still fit into the scope of the game. For example, red dwarfs have low brightness and wind, and W-R stars have large wind and brightness (apparently they are some of the brightest objects in the sky). All of the numbers are trying to be relative to the G0 class stars (the ones similar to Sol) which have default values for both wind and brightness. 

Here is a complete table of the modifiers as they currently are in this pull request:

| Star Type  | Luminosity | Stellar Wind |
| --- | --- | --- |
| A0  | 1.8  | .8 |
| A5  | 1.6  | .85 |
| B5  | 2  | .75 |
| F0  | 1.5  | .9 |
| F5  | 1.2  | .95 |
| G0  | 1  | 1 |
| G5  | .9  | .9 |
| Giant  | 1.3  | 2 |
| K0 | .85  | .8 |
| K5  | .8  | .7 |
| M0  | .7  | .6 |
| M4  | .6  | .5 |
| M8  | .5  | .4 |
| Nova  | .5  | 4 |
| W-R  | 2  | 4 |

The modifiers all live in system.cpp right now and it's a little messy. I am open to changing where the this code lives and updating the pull request with those changes and any good suggestions on changes to the modifiers.

### DIFF
--- a/source/System.cpp
+++ b/source/System.cpp
@@ -23,6 +23,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Random.h"
 #include "Sprite.h"
 #include "SpriteSet.h"
+#include "StarType.h"
 
 #include <cmath>
 
@@ -116,7 +117,7 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 	// For the following keys, if this data node defines a new value for that
 	// key, the old values should be cleared (unless using the "add" keyword).
 	set<string> shouldOverwrite = {"link", "asteroids", "fleet", "object"};
-	
+		
 	for(const DataNode &child : node)
 	{
 		// Check for the "add" or "remove" keyword.
@@ -261,6 +262,11 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 			child.PrintTrace("Skipping unrecognized attribute:");
 	}
 	
+	// Reset stellarWindStrength and luminosity before recalculating their values in
+	// the following loop.
+	stellarWindStrength = 0;
+	luminosity = 0;
+	
 	// Set planet messages based on what zone they are in, and calculate the luminosity
 	// and stellar wind values for the stars in the system.
 	for(StellarObject &object : objects)
@@ -286,7 +292,9 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 		
 		double fraction = root->distance / habitable;
 		if(object.IsStar())
+		{
 			object.message = &STAR;
+		}
 		else if (object.IsStation())
 			object.message = &STATION;
 		else if (object.IsMoon())
@@ -308,7 +316,6 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 				object.message = &UNINHABITEDPLANET;
 		}
 	}
-
 	// Make sure that the stellarWindStrength and luminosity are reasonable.
 	stellarWindStrength = std::max(0.25, stellarWindStrength);
 	luminosity = std::max(0.5, luminosity);
@@ -700,79 +707,12 @@ void System::LoadObject(const DataNode &node, Set<Planet> &planets, int parent)
 
 
 // Update stellarWindStrength and luminosity to include the star's effects in the system.
-void System::AddStar(const std::string starName)
+void System::AddStar(const string starName)
 {
-	static const double A0_LUMINOSITY = 1.8;
-	static const double A0_WIND = 0.8;
-	static const double A5_LUMINOSITY = 1.6;
-	static const double A5_WIND = 0.85;
-	static const double B5_LUMINOSITY = 2.0;
-	static const double B5_WIND = 0.75;
-	static const double F0_LUMINOSITY = 1.5;
-	static const double F0_WIND = 0.9;
-	static const double F5_LUMINOSITY = 1.2;
-	static const double F5_WIND = 0.95;
-	static const double G0_LUMINOSITY = 1.0;
-	static const double G0_WIND = 1.0;
-	static const double G5_LUMINOSITY = 0.9;
-	static const double G5_WIND = 0.9;
-	static const double GIANT_LUMINOSITY = 1.3;
-	static const double GIANT_WIND = 2.0;
-	static const double K0_LUMINOSITY = 0.85;
-	static const double K0_WIND = 0.8;
-	static const double K5_LUMINOSITY = 0.8;
-	static const double K5_WIND = 0.7;
-	static const double M0_LUMINOSITY = 0.7;
-	static const double M0_WIND = 0.6;
-	static const double M4_LUMINOSITY = 0.6;
-	static const double M4_WIND = 0.5;
-	static const double M8_LUMINOSITY = 0.5;
-	static const double M8_WIND = 0.4;
-	static const double NOVA_LUMINOSITY = 0.5;
-	static const double NOVA_WIND = 4.0;
-	static const double WR_LUMINOSITY = 2.0;
-	static const double WR_WIND = 4.0;
-	
-	if(starName == "star/a0")
-		ApplyNewStar(A0_WIND, A0_LUMINOSITY);
-	else if(starName == "star/a5")
-		ApplyNewStar(A5_WIND, A5_LUMINOSITY);
-	else if(starName == "star/b5")
-		ApplyNewStar(B5_WIND, B5_LUMINOSITY);
-	else if(starName == "star/f0")
-		ApplyNewStar(F0_WIND, F0_LUMINOSITY);
-	else if(starName == "star/f5" || starName == "star/f5-old")
-		ApplyNewStar(F5_WIND, F5_LUMINOSITY);
-	else if(starName == "star/g0" || starName == "star/g0-old")
-		ApplyNewStar(G0_WIND, G0_LUMINOSITY);
-	else if(starName == "star/g5" || starName == "star/g5-old")
-		ApplyNewStar(G5_WIND, G5_LUMINOSITY);
-	else if(starName == "star/giant")
-		ApplyNewStar(GIANT_WIND, GIANT_LUMINOSITY);
-	else if(starName == "star/k0" || starName == "star/k0-old")
-		ApplyNewStar(K0_WIND, K0_LUMINOSITY);
-	else if(starName == "star/k5" || starName == "star/k5-old")
-		ApplyNewStar(K5_WIND, K5_LUMINOSITY);
-	else if(starName == "star/m0")
-		ApplyNewStar(M0_WIND, M0_LUMINOSITY);
-	else if(starName == "star/m4")
-		ApplyNewStar(M4_WIND, M4_LUMINOSITY);
-	else if(starName == "star/m8")
-		ApplyNewStar(M8_WIND, M8_LUMINOSITY);
-	else if(starName == "star/nova")
-		ApplyNewStar(NOVA_WIND, NOVA_LUMINOSITY);
-	else if(starName == "star/wr")
-		ApplyNewStar(WR_WIND, WR_LUMINOSITY);
-	else
-		ApplyNewStar(1.0, 1.0);
-}
+	const StarType *starType = GameData::Stars().Get(starName.substr(5));
 
-
-
-void System::ApplyNewStar(double starWind, double starLuminosity)
-{
-	stellarWindStrength += starWind;
-	luminosity += starLuminosity;
+	stellarWindStrength += starType->GetStellarWind();
+	luminosity += starType->GetLuminosity();
 }
 
 
--- a/source/System.cpp
+++ b/source/System.cpp
@@ -22,6 +22,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Planet.h"
 #include "Random.h"
 #include "SpriteSet.h"
+#include "Sprite.h"
 
 #include <cmath>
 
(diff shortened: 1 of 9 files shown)


Amazinite:
So the default star sprites have hard coded values, but can those be overwritten in the map? And what about custom star sprites?

source/System.cpp
```diff
@@ -303,6 +304,155 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+        solarPanelEffectiveness += (solarPanelModifiers.back()/(i + 1));
+        
+        ramscoopModifiers.pop_back();
+        solarPanelModifiers.pop_back();
+    }
+    
+    // Make sure that the effectiveness modifiers are not game brakeingly low or high.
+    ramscoopEffectiveness = std::max(0.25, ramscoopEffectiveness);
+    solarPanelEffectiveness = std::max(0.5, solarPanelEffectiveness);
+    solarPanelEffectiveness = std::min(2.0, solarPanelEffectiveness);
```
Amazinite: Why have a hardcoded min/max value? This restricts modders.

Jeremy-E-Johnson:
@Amazinite The sprites have hard-coded values that can not be overridden in this PR. This functionality could be added, or new sprites with new values can be added.

Also min/max values are added for a couple reasons. The ramscoop min value is to stop the player from getting stranded without fuel (the same reason why a ship with no ramscoops will still get a trickle of fuel from sitting around a system). The solar panel min/max values were suggested by MZ to make sure that no ships become over-powered or derelict based on the system that they are in.

Amazinite:
>Also min/max values are added for a couple reasons.

I'd say that we can stand by these minimum and maximum values in the vanilla content, but I see to reason why plugin makers should be restricted by these values; if a plugin maker wants to make a system that gives no fuel from ramscoops then that should be up to them.

source/System.cpp
```diff
@@ -303,6 +304,155 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+	// Get list of modifiers associated with each star in the system and remember how many.
+	int starCount = 0;
+	for(StellarObject &object : objects)
+    {
+        if(!object.IsStar())
+            continue;
+
+        string starName = object.GetSprite()->Name();
+        
+        // Add the correct value to the lists.
```
Amazinite: Maybe bring these out into a data file.
```
star a0
	brightness 1.8
	wind 0.8
```
That would allow plugin makers to give their own sprites brightness and wind values.

Jeremy-E-Johnson:
Hmm. That's a fair point. My greatest concern is that the invisible deep space system near Kor Men would have 0 solar wind without the minimum ramscoop value.

jafdy:
You have different values for almost every star, and from a game play point of view I am struggling to see how such a mix of values adds to the game. My feel is that it would be better to restrict yourself (but not modders) to a few values (0.5, 1, 1.5, 2, 4) so that it is easier for players to classify the input from the stars and base this on what the sprites look like over their scientific definitions.

Perhaps setting a minimum value of 0.1 or something simply because some of the gas is sitting around in space without the need for a star (although very little. I can't so easily justify setting bounds on solar power but again some kind of minimum would be good.

Amazinite:
Maybe if the star count is greater than 0 and the star(s) in the system has no assigned brightness or wind then set a default minimum value, but otherwise I don't think that stars should be restricted in how bright or dim they can be.

source/System.cpp
```diff
@@ -303,6 +304,155 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+        else
+        {
+            ramscoopModifiers.push_back(1);
+            solarPanelModifiers.push_back(1);
+            // TOTO(Jeremy) get rid of this message statement.
+            Messages::Add("there is an unacounted for star type!");
+            Messages::Add(starName);
+        }
+        
+        starCount++;
```
tehhowch: ++thing, not thing++

source/System.cpp
```diff
@@ -303,6 +304,155 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
 		}
 	}
+	
+	// Calculate efficiency of ramscoops and solar panels.
+	std::list<double> ramscoopModifiers;
+	std::list<double> solarPanelModifiers;
+	
+	// Get list of modifiers associated with each star in the system and remember how many.
+	int starCount = 0;
+	for(StellarObject &object : objects)
```
tehhowch: There is already a loop on StellarObjects in System::Load(). See #2509

source/System.cpp
```diff
@@ -303,6 +304,155 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+	}
+	ramscoopModifiers.sort();
+	solarPanelModifiers.sort();
+	
+	// Sum up modifiers (currently using 1/n to normalize the sum).
+	ramscoopEffectiveness = 0;
+	solarPanelEffectiveness = 0;
+	for(int i = 0; i < starCount; i++)
+    {
+        ramscoopEffectiveness += (ramscoopModifiers.back()/(i + 1));
```
tehhowch: i don't think this is correct.

a better loop is:
```
for(int i = 0; i < starCount; ++i)
{
    rsE += rsM.back();
    spE += spM.back();
    rsM.pop_back();
    spM.pop_back();
}
rsE /= starCount;
spE /= starCount;
```
The loop as you have it takes 100% of the first value, then 50% of the 2nd value, and so on.

Furthermore:
I'm also not sure why you are even dividing here, since stellar outputs are directly additive (2x 100 lumens sources = 1x 200 lumen source). Perceived brightness is indeed not linear, but solar panels and matter collectors don't work on perception.

Not needing to do this loop (e.g. you correctly just add the values) means you do not even need the rsM and spM lists/maps/whatever, and in your if/elseif tree above you directly increment the solarWindStrength and luminosity variables.

Amazinite:
Maybe rename `wind` to `solar wind` or `stellar wind` as that seems like a better descriptor than just `wind`.

source/System.cpp
Hadron1776: > Why have a hardcoded min/max value? This restricts modders.

^^ In my opinion, if people _really_ wanted to make their values game-breakingly OP, it's their copy- let them.

KiLEdEnNis:
personally I hate this; yes it's more realistic that different stars give different ramscoop levels, but it's really annoying that in some systems you have to wait way longer to get the same ramscoop amount

Amazinite:
I'd say for vanilla content, the lowest we should go for solar wind and luminosity is 0.75 (25% lower than normal).

As for having to wait longer in some systems than others, those systems that you have to wait longer in aren't the majority, and probably aren't clumped together in any region of space either. There are 369 stars in the game that under this system would give less ramscoop and solar energy than currently and 130 that would give the same or more. That is 499 total stars. There are only 390 systems in the game (excluding what's in the Pug galaxy). This means that there are 109 systems with binary stars, which will yeild more ramscoop and solar power than normal (unless it's a double m8 system, in which case you're getting the same power and 10% less ramscoop under the current numbers from this PR). Of the remaining 281 singular star systems, 130 are single stars that yield more or the same power and ramscoop as currently. That means only 151 out of the 390 systems will give you less power and ramscoop under the current numbers of this PR, or about 38.7%. 

If g5 stars were used as a baseline instead of g0 (which I think should be the case), then an additional 36 singular star systems would give the same power and ramscoop as currently, meaning that only 115 of 390 systems, or about 29.5%, would give the player less power and ramscoop as currently, while 239 systems would give more and 36 systems would give the same amount (assuming all binary systems give increased power and/or fuel; I haven't checked for binary m8 stars), which is about 61.3% and 9.2% respectively.

source/System.cpp
```diff
@@ -260,9 +261,13 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 			child.PrintTrace("Skipping unrecognized attribute:");
 	}
 	
-	// Set planet messages based on what zone they are in.
+	// Set planet messages based on what zone they are in, and calculate the luminosity
+	// and stellar wind values for the stars in the system.
 	for(StellarObject &object : objects)
 	{
+		if(object.IsStar())
+            AddStar(object.GetSprite()->Name());
```
tehhowch: use tabs, not spaces

source/System.cpp
```diff
@@ -303,6 +308,11 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 				object.message = &UNINHABITEDPLANET;
 		}
 	}
+	
+    // Make sure that the stellarWindStrength and luminosity are not game brakeingly low or high.
```
tehhowch: `are reasonable`

source/System.cpp
```diff
@@ -303,6 +308,11 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+    stellarWindStrength = std::max(0.25, stellarWindStrength);
+    luminosity = std::max(0.5, luminosity);
+    luminosity = std::min(2.0, luminosity);
```
tehhowch: I would clamp the upper range to 4 or 5 - several stars already have luminosity 2.
You should allow for 0 luminosity and 0 wind (a system without a star, nebulae, or perhaps one in which something is already capturing the entire output of the star(s)). A safety catch for purely solar-powered ships could be `luminosity = std::max(.001, luminosity);`, as this would allow them a tiny bit of generation (from the galactic background light).
I would not force positive values of stellarWindStrength, as a system that drains hyperspace fuel would be a neat tool for modders (a black hole system, or some alien artifact). I'd suggest enforcing a sane number though, e.g. `stellarWindStrength = std::max(-.05, stellarWindStrength);` so that it is not insanely draining.

source/System.h
```diff
@@ -136,9 +136,19 @@ class System {
...
 	
+	// Get the effectiveness of ramscoops and solar panels in this system.
+	double GetStellarWindStrength() const;
+	double GetLuminosity() const;
+	
 	
 private:
 	void LoadObject(const DataNode &node, Set<Planet> &planets, int parent = -1);
+
+    // Update stellarWindStrength and luminosity to include the star's effects in the system.
```
tehhowch: spaces should be tabs

Jeremy-E-Johnson:
So with these commits I tried to address the style concerns that @tehhowch brought up. (thanks for reminding me about the style guide by the way)

Also, I have switched over to adding the attributes of the stars with no quantity based weights. This is a straight buff to values for wind and luminosity in multi-star systems and now they should always be pretty good for solar panels and refueling. I wanted to model some sort of diminishing returns but hopefully this system will be more intuitive to the player.

@KiLEdEnNis I agree with not wanting to get stranded in bad ramscoop systems. However, my hope is that this is far overshadowed by how nice it feels to plan a route where you refuel in a nova, wr or large binary system.

@Amazinite I decided to use g0 stars as baseline because it seemed at the time that they were most similar to Sol. Moving the baseline down is an option, or we could consider Jafdy's suggestion of moving to a stricter tier system, in which case they could both be the baseline. (I don't think this is too important as the numbers are all behind the scenes so players probably won't notice the difference)

As for the minimum and maximum values, would setting them in the data file (when I implement it) alongside the values for the sprites allow modders enough freedom?

source/System.cpp
```diff
@@ -632,6 +642,22 @@ double System::Danger() const
 
 
 
+// Get the modifier for ramscoop effectiveness.
+double System::GetStellarWindStrength() const
+{
+    return stellarWindStrength;
```
tehhowch: spaces -> tabs

source/System.cpp
```diff
@@ -632,6 +642,22 @@ double System::Danger() const
...
+}
+
+
+
+// Get the modifier for solar panel effectiveness.
+double System::GetLuminosity() const
+{
+    return luminosity;
+}
```
tehhowch: spaces -> tabs

source/System.cpp
jafdy: If you allow negatives then there is no point in bounding it.  A minimum wind makes sense as there is some gas anywhere in galactic space, and black holes are a _good_ place to collect gas because their gravity condenses it.

tehhowch: You are correct.. Even with -.05 stellar wind, with 5 wanderer ramscoops at the center of the stem the player would only be losing around .8 fuel per second. The drain can generally be as high as 20 fuel/s before the player can not notice in time (much like how the player can be eaten by nanobots if too much time is spent looking around)

source/System.cpp
```diff
@@ -260,9 +261,13 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 			child.PrintTrace("Skipping unrecognized attribute:");
 	}
 	
-	// Set planet messages based on what zone they are in.
+	// Set planet messages based on what zone they are in, and calculate the luminosity
+	// and stellar wind values for the stars in the system.
 	for(StellarObject &object : objects)
 	{
+        if(object.IsStar())
+            AddStar(object.GetSprite()->Name());
```
tehhowch: spaces -> tabs

source/System.cpp
```diff
@@ -673,6 +699,84 @@ void System::LoadObject(const DataNode &node, Set<Planet> &planets, int parent)
 
 
 
+// Update stellarWindStrength and luminosity to include the star's effects in the system.
+void System::AddStar(const std::string starName)
+{
+    static const double A0_LUMINOSITY = 1.8;
```
tehhowch: these are all using spaces instead of tabs as well

source/System.cpp
```diff
@@ -673,6 +699,84 @@ void System::LoadObject(const DataNode &node, Set<Planet> &planets, int parent)
...
+    else
+        ApplyNewStar(1.0, 1.0);
+}
+
+
+
+void System::ApplyNewStar(double starWind, double starLuminosity)
+{
+    stellarWindStrength += starWind;
+    luminosity += starLuminosity;
```
tehhowch: all the way to here was done with spaces

source/System.h
```diff
@@ -136,11 +136,21 @@ class System {
...
 	
 private:
 	void LoadObject(const DataNode &node, Set<Planet> &planets, int parent = -1);
-	
-	
+
+    // Update stellarWindStrength and luminosity to include the star's effects in the system.
+    void AddStar(const std::string starName);
+    // Helper function for AddStar() that does the actual value 
+    // incrementation of luminosity and stellarWindStrength.
```
tehhowch: spaces -> tabs here

source/System.h
```diff
@@ -178,6 +188,11 @@ class System {
 	
 	// Commodity prices.
 	std::map<std::string, Price> trade;
+		
+	// Physical attributes of the space in the system. They effect things
+	// such as ramscoops and solar panels.
+    double stellarWindStrength;
```
tehhowch: spaces -> tabs in this section too

endless-sky:
This is a reasonable idea, but I'd prefer to consider other possible implementations (and game-play implications) before settling on one.

In particular, this could probably be done with `map<const Sprite *, double>`s rather than a complete new class. And, the "old" stars should have a lot less luminosity and more wind than their younger counterparts.

PR #2766, opened 2017-07-16
update weapon table: turret turn and special damage

comnom:
Add turret turn rates and special damage values to PrintWeaponTable().

Made this for personal use, but I figured I'd share with the class. 

The base game will likely add more and more special damage weapons, and I know they're heavily used in plugins so it's more of an important balancing factor than before. (plus resistance is now a thing)

### DIFF
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -992,7 +992,7 @@ void GameData::PrintWeaponTable()
 		
 		cout << outfit.Homing() << '\t';
 		double strength = outfit.MissileStrength() + outfit.AntiMissile();
-		cout << strength << '\n';
+		cout << strength << '\t';
 		
 		double turretTurn = outfit.TurretTurn() * 60.;
 		cout << turretTurn << '\t';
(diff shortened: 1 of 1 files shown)


source/GameData.cpp
```diff
@@ -990,7 +992,22 @@ void GameData::PrintWeaponTable()
...
+		cout << heatDmg << '\t';
+		double ion = outfit.IonDamage() * 6000. / outfit.Reload();
+		cout << ion << '\t';
+		double disruption = outfit.DisruptionDamage() * 6000. / outfit.Reload();
+		cout << disruption << '\t';
+		double slowing = outfit.SlowingDamage() * 6000. / outfit.Reload();
+		cout << slowing << '\t';
+		
+		double piercing = outfit.Piercing() * 100.;
+		cout << piercing << '\n';
```
endless-sky: It'd make more sense to just leave piercing as a value between 0 and 1. (If you have a "0.1" in a spreadsheet and you format that cell as a percentage, it'll correctly display as "10%." And if you want to do math with that value, you don't need to divide it by 100 first.)

tehhowch:
Maybe add damage resistances, shield regeneration, and hull regeneration to the ship table while you're at it. Those are functions of the default loadout, but so are thrust, speed, and energy consumption

tehhowch:
Needs fuel damage too i think.

endless-sky:
I feel like it's going to get unwieldy if the table keeps growing every time new damage types, etc. are added. Note that OutfitInfoDisplay is running into the same trouble.

I wonder if there's a way to refactor OutfitInfoDisplay to have a list of attributes and scale factors to display for weapon info, and then to use that same list and scale factors here.

comnom:
Yea, that's a more reasonable approach for the long term. Gonna go ahead and close this since a refactor would make it obsolete.

PR #2768, opened 2017-07-17
added new images

jafdy:
The result of [this contest (steam)](http://steamcommunity.com/app/404410/discussions/3/135514507320943175/?tscn=1500263486), two of the images where found from Wikimedia commons.


tehhowch:
Yellowstone is a very nice place :+1:

endless-sky:
I'm just not convinced that the game needs even more landscapes right now. The landscapes are already about a third of the total size of the graphics, and there are enough that they almost never need to get doubled up (273 landscape images for 325 planets).

jafdy:
I don't follow, this is the last set (the 4th contest flopped) and it would be nice to see some of the in the game. I don't intend to run any more of these and the big attraction of them for a lot of people was the chance that they would see their own work in the game.

KiLEdEnNis:
even if it's not added now, I think it should be stored somewhere to add when the need rises

PR #2772, opened 2017-07-17
Slight Dialog change

Ajm219:
Changed because while playing threw that the line "The Syndicate hate my my guts right now" does not seem like something a Professional would say, especially in the year 3014. Personal opinion does accompany this, take it with a grain of salt.

### DIFF
--- a/data/free worlds intro.txt
+++ b/data/free worlds intro.txt
@@ -790,7 +790,7 @@ mission "FW Katya 2B"
 				`	"How does that even work?"`
 				`	"Is that even legal?"`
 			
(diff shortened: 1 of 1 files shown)


solardawning:
Why?
Seriously- reading the change, why? It changes the characterization of Ijs, in a way that doesn't seem to be necessary, or a good change.
The existing line reads as a witty, snarky comment- exactly what a self-described rogue nuclear inspector/activist would say. The proposed change loses that.

Downvoted.

Rob59er:
Yeah there's no point to this.

jafdy:
He is not strictly a professional . . .

Please-Close-This-Issue:
Quickly

MessyMix:
As I said in the Discord, this is how to make a PR, and the game is open source so if you feel like there is something you should add, do so. I did not specifically recommend this change, however. Just clarification.

PR #2778, opened 2017-07-19
Replacing duplicate landscape images

Amazinite:
I've checked the descriptions of all the affected planets and I believe that all the replacement images fit the descriptions as well as or better than the previous image.

The main purpose of this was to use some unused landscape images, but this only uses half of what is currently unused. There's another 22 images that are unused either because there aren't enough planets with the type of landscape that the image is or I didn't think that any of the unused images could replace a duplicate image and still fit the description (e.g. the planets with lava0 or lava6 didn't feel like they could be replaced with lava7).

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -24035,7 +24035,7 @@ planet "Aresepru Nat"
 
 planet Asgard
 	attributes deep factory urban rich
-	landscape land/city1
+	landscape land/city13-sfiera
 	description `Asgard was one of the first worlds to be settled in the Deep. It is now home to many crowded cities, with soaring skyscrapers that are made of lightweight composite materials rather than the steel and cement so commonly used elsewhere.`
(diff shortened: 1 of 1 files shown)


tehhowch:
One bit that was missed in my initial check was that I only pulled data from map.txt for the initial count - two other files do change the landscape images of planets:

There's another use of `hills3` in pug.txt, for Ruin, and it's also used for Desi Seledrak and Gemstone.
The Wanderer campaign also gives `beach1` to Varu K'prai after the Sestor occupation. (`sea4` originally in map.txt); `beach1` is also used by Shorebreak and Turquoise Four, while `sea4` is not used by any other planet.

Amazinite:
Hills3 was another one of those landscapes that I didn't feel could be replaced with any of the unused hill images (hills5, 6, and 7).

PR #2752, opened 2017-07-12
Add more mining variants and fleets

tehhowch:
1. Add mining fleets for Deep, Hai, Unfettered Hai, Korath, Pirates, Remnant, and Wanderers
2. Wanderer's Earth Shaper now is capable of actually mining, and is the primary component of the Wanderer mining fleet.
3. Added new mining ship + mining drone combo, courtesy of @Kryes-Omega 
4. Events which introduce ships will add to those definitions as appropriate
   - Nest, Roost for Pirate, Human mining fleets
   - Mereti miners for Korath Exiles
5. Added fleets to map.txt where appropriate
   - considered normal fleet periods and also the number of minables when setting the mining fleet periods
   - Updated Wanderer events that alter minable systems as necessary to maintain the new mining fleets

Notes:
1. Remnant do not have many places to mine (yet). Not sure how their territory is going to evolve.
2. Included 2 jump-drive capable variants of the Model 8 and a "bodyguard" Model 16 so that Kor Ak'Mari is not mined by just the Raider+Chaser.
3. I did not go through the FW events to check how many Human Miners get removed by the many system fleet respecifications (best done in a different PR) - the answer here is undoubtedly "many / all" for Republic / FW / Syndicate space. I have a PR ready to switch the entirety of Free Worlds story from the static respecifications into `add/remove` format, which will address this persistence.
4. Mining fleet behavior will benefit from #1971 (cargo transfer) and #3064

Issue #2742, opened 2017-07-11
Remove ______ only removes 1st instance


tehhowch:
It can happen that a system's fleets definition obtains more than one instance of a particular fleet, either with the same period or different periods, e.g.:
```
system "Testbed"
    fleet "Small Southern Pirates" 2000
    fleet "Small Southern Pirates" 800
    fleet "Southern Pirate Miners" 2000
    fleet "Large Security Force" 500
```
If one wants to use the `remove fleet` syntax to edit this system, in order to remove fully remove the "Small Southern Pirates" fleet, an event will need to specify `remove fleet "Small Southern Pirates"` twice. For players with multiple mods installed, the only way a mod creator can guarantee that a specific fleet is removed is to perform a number of removals in any one event (or play 'not nice' and just redefine the system's fleets entirely).

There are a couple of ways to address this
1) forbid multiple definitions by merging successively-defined fleets (e.g. 2000 + 800 -> 571)
2) use a more complex iterative removal solution.
3) keep the current behavior

Some notes:
 - Merging instances of the same fleet would allow for partial reductions in that fleet's arrival rate
 - Merging instances prevents "flavor" behavior such as defining one of the fleet's entries with a low period (for a constant presence) and another with a rather high period (a chance opportunity).
 - Partial reductions and merging behavior simplify event definitions considerably
 - Complex iterative removals may make the code harder to read
 - Mod creators should probably know what exactly is going on w.r.t. fleets in systems at each point of their story, and thus should know how many times they need to call `remove ____` to fully remove a fleet from a system. However, they won't know if there are extras that their mod did not add.

tehhowch:
My vote is either to implement auto-merging and partial reductions, or switch to a more complex iterative removal (an outer while loop that repeats the existing range-based for loops)

jafdy:
The fleet arrival rates should be a Poisson distribution, one over the period can be added up and the result should be indistinguishable from leaving them separate.

endless-sky:
To me, the existing behavior seems fine. Saying that if you add an item twice you also have to remove it twice makes at least as much sense as saying that "remove" automatically clears any number of copies of an item.

tehhowch:
My preference is for an auto-merge to maintain the average period (e.g. 2000 + 800 -> 571) so that a single remove will do the removal, and an additional system fleet keyword, `"set period"`, which simply overwrites the existing fleet period, or adds the fleet to the system with the desired period.

tehhowch:
[...]

PR #2779, opened 2017-07-19
Implement add/remove syntax for the Wanderers campaign

tehhowch:
This work is taken from my PR that adds more mining variants (to capitalize on #2726 ) and extends mining from just humans and the Kor Mereti to almost all factions. When it came time to add to the Wanderers, I noticed that while we had updated the fleet definitions with add/remove, we had entirely neglected to do the same for system definitions. 

This PR splits the work of changing how the fleets are defined in Wanderer events from the tangentially related "update fleets and races to utilize mining" that is #2752

### DIFF
--- a/data/wanderers middle.txt
+++ b/data/wanderers middle.txt
@@ -231,12 +231,11 @@ event "wanderers: kaliptari battle begins"
 
 event "wanderers: kaliptari battle ends"
 	system Kaliptari
-		fleet "Large Quarg" 2000
-		fleet "Small Kor Mereti" 8000
-		fleet "Large Kor Mereti" 3600
-		fleet "Wanderer Defense" 1000
-		fleet "Wanderer Drones" 500
-		fleet "Wanderer Flycatchers" 500
+		add fleet "Large Quarg" 2000
+		add fleet "Small Kor Mereti" 8000
+		add fleet "Large Kor Mereti" 3600
+		add fleet "Wanderer Defense" 1000
+		add fleet "Wanderer Drones" 500
 
 mission "Wanderers: First Mereti Attack"
 	landing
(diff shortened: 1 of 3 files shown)


data/wanderers middle.txt
```diff
@@ -1094,11 +1093,10 @@ mission "Wanderers: Mind 6"
 
 event "wanderers: kor mereti transformation"
 	system Chimitarp
-		fleet "Small Kor Mereti" 200
-		fleet "Large Kor Mereti" 400
+		add fleet "Small Kor Mereti" 250
```
tehhowch: These have a larger period because there is an existing definition of small/large Kor Mereti fleets in Chimitarp, and the combined result has average period 200 / 400.

If we change how add/remove works for fleet periods ( #2742 ), this should be revisited as well.

data/wanderers middle.txt
```diff
@@ -3039,8 +3025,8 @@ event "wanderers: sestor shutdown"
 		fleet "Small Kor Sestor" 3200
 		fleet "Large Kor Sestor" 4800
 	system "Mesuket"
-		fleet "Large Kor Mereti" 900
-		fleet "Large Kor Sestor" 10000
+		remove fleet "Large Kor Sestor"
+		add fleet "Large Kor Sestor" 10000
```
tehhowch: If we change how add/remove works for preexisting fleets ( #2742), this subsection (and L3137) should be revisited too.

data/wanderers middle.txt
```diff
@@ -3158,9 +3144,9 @@ event "wanderers: sabira eseskrai colony"
 		fleet "Wanderer Drones" 800
 		fleet "Kor Efret Home" 800
 	system "Furmeliki"
-		fleet "Large Quarg" 1000
-		fleet "Kor Efret Home" 800
+		add fleet "Kor Efret Home" 800
 	planet "Sabira Eseskrai"
+		add attributes wanderer
```
tehhowch: This gives Sabira Eseskrai `attributes korath wanderer` which seems very appropriate given that the Efret and Wanderer populations are rehabilitating this world together. Without this line, Sabira remains `attributes korath`.

data/wanderers.txt
```diff
@@ -676,26 +656,19 @@ event "wanderers: unfettered invasion starts"
...
-		fleet "Large Unfettered" 1200
+		add fleet "Wanderer Defense" 500
+		add fleet "Wanderer Flycatchers" 480
+		add fleet "Large Unfettered" 1200
 	"reputation: Hai (Unfettered)" <?= -100
 	government "Hai (Unfettered)"
 		"attitude toward"
 			"Wanderer" -.01
-	outfitter "Wanderer Advanced"
-		"Wanderer Anti-Missile"
```
tehhowch: Wanderer A/M was already added to this outfitter in a previous event.

data/wanderers.txt
```diff
@@ -947,17 +863,13 @@ event "wanderers: derecho mass production"
 
 event "wanderers: spera anatrusk colony"
 	system Ap'arak
-		fleet "Wanderer Freight" 2000
-		fleet "Wanderer Drones" 4000
-		fleet "Wanderer Defense" 600
-		fleet "Wanderer Flycatchers" 400
-		fleet "Large Unfettered" 8000
+		remove fleet "Large Unfettered"
+		add fleet "Large Unfettered" 8000
```
tehhowch: Same bit as before about #2742

PR #2786, opened 2017-07-21
Add configuration files for CI tools

tehhowch:
.travis.yml for build verification of linux executables by Travis CI
.appveyor.yml for build verification of Windows executables by AppVeyor
  +	.winmake for "headless" builds of Endless Sky using only MinGW

These CI tools are free for FOSS such as Endless Sky. For those who do not have a 2nd machine of a different architecture, these services allow one to verify that their commits do not break the build process on other operating systems. The makefile is included mostly for AppVeyor, but it will build Endless Sky for Windows without the need for Code::Blocks configuration, provided the DIR_ESLIB and DIR_MINGW64 directories are pointed at the right folders.
Even if @endless-sky does not enroll in the services, by distributing working configuration files all that remains for those who do wish to use the services is to sign up on an individual basis. (It's about 3 clicks each if you use GitHub authentication).


kikotheexile:
Should make all manner of projects easier.

MCOfficer:
so we're finally getting a CI aswell? awesome

tehhowch:
No, this PR just makes it possible for anyone with a fork to use a free CI service to check that they are making changes on their branches that still build on other platforms. It doesn't require anyone to use a CI service, not even MZ.

By including these config files in the base repo, no special hoops need to be jumped through in order for someone who uses a CI service to check that their new changes on a given branch are working as expected. Right now those hoops pretty much require a terminal (unless one doesn't mind superfluous commits), and consist of 
1. work normally
2. merging/adding these files in a commit
3. pushing that commit to the web,
4. after verifying a successful build on all platforms, reverting that commit
    - either via interactive rebase, dropping the commit that added the config files
    - or via `git revert` which adds another commit
5. pushing the reversion commit (or force-pushing the new public history

With this PR, the process to get the same results is:
1. work normally

MCOfficer:
i know CI is optional, i was just hoping that, since everything is set up with this PR, MZ would just use it... whatever, @ MZ consider this a vote ;)

kikotheexile:
Wont this also clear up some of those "muh special settings when compiling from source break the game and its everyone else's fault!" threads to some degree? Or did i misread?

tehhowch:
This PR would not alleviate or nullify any difficulties that players may have with compiling from source - it is already well-known that others are able to build from source, so a CI service also being able to build the same source files they are working with will do nothing (the CI service does not use their environment).

kikotheexile:
Shucks. I misread. I guess that means people will still have to read and follow instructions then.

PR #2788, opened 2017-07-21
Cache friend/foe list for each government to limit ship iteration

tehhowch:
This PR takes the government roster added in #3531 and persists it outside of the `UpdateStrengths` method, so the list of foes and non-foes each frame for all in-flight governments can be used by other class methods. (Limiting the number of `Ship` objects to interrogate when needing to find a target for some purpose.)

An accessor function `GetShipsList` is added that applies common exclusionary criteria to the government rosters:
 - ship must be targetable (not cloaked, landing, taking off, dead, or in a non-player system)
 - ship must not be near hyperspace velocity 
 - ship must be in the same system
 - within a desired range
 - `marked` (or not `marked`) as appropriate

Consuming functions (like `FindTarget`, `AimTurrets`, or `DoSwarming`) then need only apply their task-specific logic to this subset, instead of determining whether the same ships meet the above criteria for each ship that executes the function.


Hadron1776:
This PR seems to lower CPU usage _very_ slightly when I'm just sitting around with a cloaked fleet in Mesuket. In big battles, this PR cuts CPU use from ~160% to ~130%. Still, I think the projectile spam from the ships is much more of a CPU hog than the AI targeting routines.

tehhowch:
In addition to large battles, there should be cpu savings when running large freighter fleets, as each ship will only need to check the available govts rather than every other ship, including it's own

solardawning:
As a large fleet player, I love this.
The recent turret changes to give them orientation and independent targeting led to combat hitching with much smaller fleet numbers than previous, so any efficiency changes here will help.

tehhowch:
@solardawning,

It slipped my mind to do this for AI::AimTurrets as well! This roster would be a good fit there as well, and would certainly help reduce some of the impact of self aimed turrets.
I'll add that when I get a chance on Sunday (but if someone wants to beat me to it and PR to this branch that's totally ok!)

tehhowch:
@endless-sky I'm curious about performance implications of the following:
1) as-is implementation
2) clearing rosters at the end of every AI::Step (instead of only in AI::Clean())
3) use maintenance function to erase expired weak_ptr from each government map every couple of AI::Step calls (e.g. `!Random::Int(180)`)
4) use maintenance function to erase expired weak_ptr from each government map every AI::Step

1) accumulates expired weak_ptrs which have to be iterated over and checked, and player ships must be checked to see if they changed government (and erased from the other govts if true).
2) has minimum 2 * ships.size() call which modify the roster every step, but ensures only non-expired pointers are in the list, since it is only valid for the current step. This would mean one less check in GetShipsList, possibly changing from weak_ptr to shared_ptr (no need to try locking), and no need to walk the non-player governments to handle captured ships (both now and if capturing is extended to other governments)
3) is a mix of 1 and 2 - there could be a handful of expired pointers due to ship deaths / forgets since the last maintenance, player ships would still need to be checked for reassignment, but in general there would not be many `emplace` or erases on any given Step, nor would their be many extra weak_ptr that have to be checked and skipped during target acquistion). There would be at minimum ships.size() [] updates of the roster map.
4) means walking the rosters every step and calling `erase` if `expired()` is true. Fewer erase calls, but have to use the [] operator instead of emplace.

I'm leaning towards 2, or perhaps 4, because if the player stays in a busy system for a long time, this can result in a lot of expired pointers from ships being destroyed, or from jumping away and being forgotten.

tehhowch:
I did some profiling with gprof of the above tweaks. The takeaway is that, at minimum, some removal of expired weak_ptr needs to be done, as the accumulation of expired pointers in the roster will otherwise  remove any possible benefit to having a list of only enemies / non-enemies instead of the total ship list (which is every NPC for every mission, + any ships in your current system).

This summary considers the reported normalized self time from the 6 variations listed below. I normalized the reported self time by scaling each such that all methods had the same reported self time for Outfit::Get()
I then summed the normalized self time from the methods AimTurrets, Autofire, and FindTarget ("Method Time"), and then add the normalized self time for the GetShipsList method to create Total Time.

Version | Method Time | Total Time | raw Outfit::Get calls
-- | -- | -- | --
clear+emplace shared_ptr | 9.710 | 21.445 | 2.962E+09
occasional purge | 9.404 | 22.827 | 1.139E+09
0.9.7 | 26.840 | 26.840 | 2.200E+09
always purge | 15.368 | 45.609 | 2.409E+09
clear+emplace weak_ptr | 18.940 | 52.880 | 3.360E+09
PR as-is | 12.820 | 145.236 | 2.063E+09




<details><summary>Profiles (same fleet, master as of prior to the commits in this PR, travel from Viminal to Mesukent, and then spending a reasonable amount of time there. All mission NPCs were removed from the save, leaving only ships in the fleet and those that spawned in Mesukent. Fleet used is that of Travion Expectra (75 ships)</summary>

[Travion Expectra.txt](https://github.com/endless-sky/endless-sky/files/1181945/Travion.Expectra.txt)


0 - master: 
[v0.9.7.txt](https://github.com/endless-sky/endless-sky/files/1181566/v0.9.7.txt)

1 - PR: 
[as-is.txt](https://github.com/endless-sky/endless-sky/files/1181533/as-is.txt)

2a - clear + emplace w/weak_ptr: 
[clear_emplace_weakptr.txt](https://github.com/endless-sky/endless-sky/files/1181530/clear_emplace_weakptr.txt)

2b - clear + emplace w/shared_ptr:
[clear_emplace_sharedptr.txt](https://github.com/endless-sky/endless-sky/files/1181531/clear_emplace_sharedptr.txt)

3. Using `!Random::Int(180)` to control the rate of cleaning the list. This can safely be larger, probably once a minute is OK.
[purge_expired_sometimes.txt](https://github.com/endless-sky/endless-sky/files/1181679/purge_expired_sometimes.txt)

4. Like option 2, clean out expired weak_ptr every time AI::Step runs.
[purge_expired_always.txt](https://github.com/endless-sky/endless-sky/files/1181680/purge_expired_always.txt)

</details>

tehhowch:
Here are two profiles for the same battle, the one in Farpoint when you are given the bomb for Zenith. I ran both battles for 2 minutes of user time, with no capslocking. For both, this amount of wall time corresponds to the battle being finished, though there are still numerous disabled ships strewn about.
[Travion Expectra.txt](https://github.com/endless-sky/endless-sky/files/1185119/Travion.Expectra.txt)

1: `v0.9.7`:
altalnilam2-v0 9 7
[AltAlnilam2-v0.9.7.txt](https://github.com/endless-sky/endless-sky/files/1185126/AltAlnilam2-v0.9.7.txt)

2: a826ed8:
altalnilam2-a826ed8
[AltAlnilam2-a826ed8.txt](https://github.com/endless-sky/endless-sky/files/1185127/AltAlnilam2-a826ed8.txt)

endless-sky:
This definitely looks promising, but it's a big enough change that I'd prefer to put it off until the next unstable release just in case. (I'm hoping to release 0.9.8 in a week or two.)

tehhowch:
I would have been shocked if this merged prior to a stable release 😄

tehhowch:
I'm thinking the next obvious step for this PR is to remove the range argument to GetShipsList and cache the ally and enemy vectors at the start of each AI::Step. Given that the probability of at least one ship of each government in the player's system having a turret that can be aimed is essentially 1, precomputing the vector of enemy ships and storing it each Step should be an almost-free performance improvement. The "almost-free" comes from each government's vector of allies will only being used if a ship is doing surveillance / swarming (which is "free" in Republic and Wanderer space).

tehhowch:
With the latest 2 commits:
Testing `master` vs this vs this with caching:
1. Using the Travion Expectera file above, for the Defend Farpoint mission
2. Jumping to Mesuket afterwards
3. Idling Mesukent for minimum 5 minutes
4. Jumping a bit more in Korath space
5. Back to Mesuket

<details><summary>master @6ec39691fc9bccfa </summary>

[master.txt](https://github.com/endless-sky/endless-sky/files/1487388/master.txt)
master
</details>
<details><summary>this</summary>

[NoCaching.txt](https://github.com/endless-sky/endless-sky/files/1487361/NoCaching.txt)
nocaching
</details>
<details><summary>this with caching</summary>

[cached2.txt](https://github.com/endless-sky/endless-sky/files/1488947/cached2.txt)
cached2
</details>
Based solely on the reported total time in the nodes it looks like the no-caching methods results in about 5% more total time spent in AI::Step and all its children, and about 5% less when using the caching method.

I need to update a few things given that FindTarget and AimTurret are looking for Body *, but should have something soonish.

tehhowch:
As of 0449fdb for the above test case:
<details><summary> master @ ba4d1b4849967121939ca4c55eead444b7a5a984 </summary>

master
[master.txt](https://github.com/endless-sky/endless-sky/files/1489425/master.txt)
</details>
<details><summary> 0449fdb </summary>

cached
[cached.txt](https://github.com/endless-sky/endless-sky/files/1489463/cached.txt)
</details>

tehhowch:
Kind of a reductio ad absurdum situation, but for a fleet of 3901 Aphids performing a single jump from Fah Soom to Due Yoot: [Mr.Test.txt](https://github.com/endless-sky/endless-sky/files/1573964/Mr.Test.txt)

master @ https://github.com/endless-sky/endless-sky/commit/3c4e13947b9200983dd35754cb5d6f6fd61dc392
megafleet
[megafleet.txt](https://github.com/endless-sky/endless-sky/files/1573959/megafleet.txt)

this PR @ 
https://github.com/endless-sky/endless-sky/pull/2788/commits/7617fefb5981a6f5ad260300153cc4e3d6ec1cfe
megafleet_pr
[megafleet_PR.txt](https://github.com/endless-sky/endless-sky/files/1573961/megafleet_PR.txt)

Same operation, same sampling rate, and with master we sample around 122s of program operation, while with this PR we only sample 35.9s.

Sinsling:
Based on my (admittedly limited) understanding of the changes being made, I have some questions;

1) In PR a826ed8, does 'target nearest' still find the **closest** enemy ship if the nearest enemy ship is in the shared government list of the *current* target? Example: does target properly change from pirate argosy '1' @ 600 range to pirate argosy '2' @ 400 range?

2) In PR 7617fef, it appears as though ships will only auto-fire if their current target is viable, and not any time a hostile ship is in their fore section (think merchants taking pot-shots at disabled pirates). Is this not the case?

tehhowch:
> does 'target nearest' still find the closest enemy ship if the nearest enemy ship is in the shared government list of the current target? 

I'm not sure what that's asking. AI::GetShipsList applies the common exclusion criteria from each of the methods it is called from - it doesn't alter the biases any of its calling methods use to choose the particular ship from the list of enemies, it only shortens that list from "every ship in the game" to "(friendly/hostile) ships in the player's system."
> it appears as though ships will only auto-fire if their current target is viable, and not any time a hostile ship is in their fore section (think merchants taking pot-shots at disabled pirates). Is this not the case?

I didn't change any of the "should I fire" code, only limited the scope of the container that is checked for "should I fire." GetShipsList does not discriminate on a target's disabled status, so Autofire will still consider disabled ships for "potshots".

(Due to changes in some code's indentation level, it may be easier to compare the before / after for this PR in the "Split" viewing mode rather than "Unified".)

PR #2803, opened 2017-07-22
Make restricted planets a separate color from hostile ones

Hadron1776:
This PR makes restricted planets show up yellow on the radar, instead of red. (Red is still used for hostile planets).

I felt like there should be a difference between "you can't land here because this planet is restricted" and "you can't land here because we hate your guts".

To be fair, there _is_ a difference in the `master` branch: The little spikes on the planet label. You'd have to be really attentive to notice it, though.

EDIT: I just now noticed how this change is nicely consistent with the way restricted planets appear in the map view :)

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -59,6 +59,8 @@ namespace {
 			return Radar::PLAYER;
 		if(object.GetPlanet()->CanLand())
 			return Radar::FRIENDLY;
+		if(!object.GetPlanet()->GetGovernment()->IsEnemy())
+			return Radar::UNFRIENDLY;
 		return Radar::HOSTILE;
 	}
 	


Pointedstick:
Awesome! How about orange instead, just for nostalgic consistency with Escape Velocity?

tehhowch:
That can be done locally via a plugin that changes the "radar unfriendly"
color 👍

endless-sky:
I do wonder if orange would make more sense, just because yellow means "hostile but not targeting you" in the current radar.

Or, would that just be confusing to people because orange is in between yellow and red?

endless-sky:
On the other hand, the "red" color used in the radar is close enough to orange that it's not really possible to create a "restricted" color that's clearly different from it.

tehhowch:
If you add a restricted = orange color, then i would ask for a more-clearly-red hostile color.

PR #2806, opened 2017-07-23
Press R or U in Outfitter to uninstall selected outfit directly to cargo.

tehhowch:
Pressing `r` or `u` will uninstall the selected outfit and place it in to your cargo hold, if there is sufficient room. If there is not sufficient room for the outfit in cargo, it will be sold.
Excess ammunition that needed to be stored but cannot fit in the remaining space will be sold.
As expected, the normal modifier keys work to multiply the quantity moved to cargo.

This makes it possible to move an outfit installed on a ship directly into cargo, or to more simply sell an outfit from a ship when you have some in cargo that you want to keep (uninstall once + sell once vs sell every copy in cargo, then the one from the ship, then buy back into cargo).

### DIFF
--- a/source/OutfitterPanel.cpp
+++ b/source/OutfitterPanel.cpp
@@ -604,7 +604,7 @@ void OutfitterPanel::FailSell() const
 
 
 
-bool OutfitterPanel::CanRemove() const
+bool OutfitterPanel::CanUninstall() const
 {
 	if(!planet || !selectedOutfit)
 		return false;
@@ -626,7 +626,7 @@ bool OutfitterPanel::CanRemove() const
 			else
 			{
 				// Not all of the ship's ammo necessarily needs to
-				// be removed along with remove the outfit, but this
+				// be removed along with uninstalling the outfit, but this
 				// is not a straightforward check and requires modifying
 				// the ship's attributes outfit after removing the outfit.
 				continue;
@@ -641,9 +641,9 @@ bool OutfitterPanel::CanRemove() const
 
 
 
-void OutfitterPanel::Remove()
+void OutfitterPanel::Uninstall()
 {
-	// Find the ships that have the most of this outfit to remove.
+	// Find the ships that have the most of this outfit to uninstall.
 	int most = 0;
 	vector<Ship *> shipsToOutfit;
 	for(Ship *ship : playerShips)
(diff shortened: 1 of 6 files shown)


Bladewood:
"Remove" is not the right term to describe this.

tehhowch:
Could be switched to `u` for `U`ninstall or `U`nequip

tehhowch:
I've updated the terminology to use 'Uninstall'. Both `r` and `u` trigger the behavior

tehhowch:
a6ba578 fixes a logic error that I found in CanUninstall(), in addition to properly checking if a ship can remove an outfit and a portion of its ammo if there isn't room for the removed outfit and all of the installed ammo, and moving common code from Buy, Sell, and Uninstall into a single function.
0be7046 replaces very simple `auto` uses with the requisite substitutions, adds `#includes` I believe were missing but covered by other imported headers, and replaces the Refill() `while` loop with a once-through transfer of the required amount.

endless-sky:
There's a lot of copy-pasted code here. I feel like it would make more sense to implement removal as a sell followed by a buy into cargo.

It might also make sense to not apply cargo space limits when uninstalling, because most players are just going to use this (if at all) for uninstalling from one ship and then installing in another. (But of course, if that's all they're using it for, there would be no reason for them to do that instead of just using sell / buy.)

In general, I'm also just not sure that this is useful enough to be worth adding. The one use case that you cite is one that almost never comes up. I think the main reason people have asked for an "uninstall" function is because they want to be able to "uninstall" an outfit from one ship and install it on another, because they haven't figured out that they can just sell and buy and get the same effect.

tehhowch:
I did follow the structure of the existing Buy/Sell code, and the restriction of working within the player's available cargo space is likely what led me to make a standalone method instead of reworking Buy and Sell to support a remove-type action. If I understand your comment, you are suggesting it is OK to leave the player with negative cargo space free, as this will be caught in the Planet::TakeOff code. Given that, I can rework this PR to be more streamlined.


I think another reason why players don't immediately learn that they can use buy & sell to accomplish this is because they are working with an outfit that is available in the store as well - they see "sells for" and a low price, and the buying price (before they sell) is full price. They don't want to lose the money, and don't realize that the Outfitter's shopkeep is actually nice to them and will let them buy the used outfit at cost. They see the possibility to lose money in a situation where they feel they shouldn't, even though the reality is that they will not lose any money. Being able to press a single key and observe no change to their money, and the successful transfer of something from their ship to their cargo seems to be an adequate way to address that fear.

On the subject of usefulness: 
 - This PR adds the complement of the "Install" functionality. Having one implies having the other, and not having both but having one seems incomplete. 
 - The checkboxes do reduce the level of perceived need for this functionality, as it is visually apparent to the player that there is a way to alter what is being shown/sold and where it goes.

tehhowch:
@endless-sky 7883317b should resolve the issue with duplicated code. Since the cargo hold will not accept over-capacity transfers, any excess is sold off instead.

Issue #2801, opened 2017-07-22
Parked Flagship should not hide buttons, and should auto-unpark


MessyMix:
Currently, if you switch your flagship to a parked ship, and then sell your original flagship, you end up with this screen:

![image](https://user-images.githubusercontent.com/20450076/28483273-01966f20-6e9f-11e7-9d21-5c66f67857f3.png)

It's really confusing and it looked very much like a bug to me, since some functions of the planet were hidden. It would be great if a flagship auto-unparks once you assign it, otherwise you can't access anything, which looks like a bug. Any other solutions are welcome too. :)

Bladewood:
The same thing happens if you end up without a flagship in any way.
It even happens at the start of the game, if you refuse to purchase a ship.

Hadron1776:
I think this issue is when you _have_ a potential flagship, but it's parked...

MessyMix:
My opinion: the potential flagship should auto unpark once you assign it, because otherwise you can't do anything on the ground.

AdamMil:
Yes, I think when you select a new flagship it should unpark.

endless-sky:
This is the intended behavior. Choosing whether or not to unpark a ship (and if so, which ship to unpark) should be the player's choice, not something the game chooses for you.

For example, a common situation where this could occur is when the player is upgrading to a new flagship but needs to get a loan. So, they sell their old ship, then visit the bank, then return to the shipyard.

MessyMix:
It would be a lot better if it didn't seem like a bug, especially since the game crashes at exactly this point—when rearranging the ships via the info panel. Perhaps give the full UI, but "block" the player from functions, with a dialogue that says you need to unpark the ship first.

Side note: why can you access the outfitter without a ship?

PR #2811, opened 2017-07-24
Unpark the would-be flagship if it is parked and so are all others

tehhowch:
Refs #2801

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -747,12 +747,22 @@ const shared_ptr<Ship> &PlayerInfo::FlagshipPtr()
 {
 	if(!flagship)
 	{
+		// Choose the first unparked, eligible ship to be the flagship.
 		for(const shared_ptr<Ship> &it : ships)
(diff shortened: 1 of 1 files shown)


endless-sky:
I'm not sure if it makes sense to automatically unpark a ship without giving the player a choice of which ship. The player might want to unpark a different one, or may in fact want to buy a new flagship instead of unparking a ship.

tehhowch:
So you're suggesting that instead of the extra bit in FlagshipPtr(), the PlayerInfoPanel is opened when there are no eligible unparked ships, but there are eligible parked ships? UI window manipulation is not my strong suit, so I'll leave that up to you.

Feel free to close this PR if you do not plan to merge it, as I don't plan to extend this branch to do anything more than what it already does. As is, it nicely rectifies the issue raised in #2801, but as you indicate, it does choose for the player.

Hadron1776:
>  As is, it nicely rectifies the issue raised in #2801, but as you indicate, it does choose for the player.

Specifically, it chooses the ship that _would have_ been the flagship if it weren't parked.

PR #2822, opened 2017-07-25
Allow blinking of mission targets to be disabled in the Preferences menu

Hadron1776:
In the `master` branch, ships with the `target` personality will double-blink on the radar. This PR adds a preference to disable this if players want to do so.

highlight mission targets

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -64,7 +64,7 @@ namespace {
 	
 	int RadarType(const Ship &ship, int step)
 	{
-		if(ship.GetPersonality().IsTarget() && !ship.IsDestroyed())
+		if(ship.GetPersonality().IsTarget() && !ship.IsDestroyed() && Preferences::Has("Highlight mission targets"))
 		{
 			// If a ship is a "target," double-blink it a few times per second.
 			int count = (step / 6) % 7;
--- a/source/Preferences.cpp
+++ b/source/Preferences.cpp
@@ -53,6 +53,7 @@ void Preferences::Load()
 	settings["Draw background haze"] = true;
 	settings["Hide unexplored map regions"] = true;
 	settings["Turrets focus fire"] = true;
+	settings["Highlight mission targets"] = true;
 	
 	DataFile prefs(Files::Config() + "preferences.txt");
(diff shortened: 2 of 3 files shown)


Bladewood:
@gunqqer I personally think a flash is good, but it flashes too often or too quickly or something.

gunqqer:
Maybe in the future when you have to clear an entire system it would be annoying it would be more useful. But everyone like things differently

Hadron1776:
Keep in mind that this PR doesn't touch the part of the radar that actually _says_ `mission target` - it just allows one to turn off the flashing.

endless-sky:
I don't think this will be a widely-enough-used feature to be worth adding. There are dozens of UI frills like this, and if there were a setting for every one of them it'd become impossible to find the few settings that people actually care about.

A better option would be to add Radar::BLINK as a distinct color (white by default) that can be configured in the data files. That way someone who really wanted to disable or de-emphasize the blink could do it via a plugin, but it wouldn't be cluttering the settings page.

gunqqer:
Yeah I myself won't use it, maybe a preference in preferances.txt but not in the settings pane in game

tehhowch:
A second option would be to decrease the blink rate by 20-50%

gunqqer:
I feel like a lot of preferances should be just hidden inside the .txt so we can have more customisable things and not clutter up the pane

tehhowch:
That is indeed possible. Preferences need not be defined on the preferences panel, but it makes sense that situational ones are.

Bladewood:
I assume you're going to work on Michael's suggestion?

endless-sky:
I could also do it myself, @Hadron1776; just let me know what you'd prefer.

Hadron1776:
@endless-sky I can handle it :)

MessyMix:
Awe. This was definitely something I was going to use; the blinking target was fine on the radar but gave me epilepsy in the "selected target" thing... I look forward to seeing how the new solution turns out :)

Bladewood:
@MessyMix https://github.com/endless-sky/endless-sky/pull/2825?

Issue #2820, opened 2017-07-25
Crash-to-Desktop when disowning all ships ordered ahead of flagship


tehhowch:
Windows 10 v1607 (have not checked other OS)
Prebuilt 0.9.7 binary from downloads page (have not checked Steam / self-compiled versions)

To reproduce:
1. Have ships in other system (unparked, perhaps also parked):
![image](https://user-images.githubusercontent.com/20871346/28586729-5c843b22-713a-11e7-9f3f-644d3a50fe32.png)

2. Sell all ships in current system
![image](https://user-images.githubusercontent.com/20871346/28586776-7ebb7f2a-713a-11e7-80e5-95cc1ab9eeb2.png)

3. Buy new ship in current system
![image](https://user-images.githubusercontent.com/20871346/28586811-8e2a7178-713a-11e7-9b8f-26a91d26f809.png)

4. Disown ships in other system
![image](https://user-images.githubusercontent.com/20871346/28586843-a705d44e-713a-11e7-9275-d4633610ab36.png)

5. CTD.
![image](https://user-images.githubusercontent.com/20871346/28586852-b5a9316c-713a-11e7-96f2-4b455f0de0d4.png)

I can look into the cause / PR a fix later, but figured I'd post the bug in case someone can act more quickly than I

gunqqer:
I would test this on OS X but over here we're having the game crash if you try to reorder ships (unless it way fixed and I just haven't seen it today)

tehhowch:
Confirmed on latest self-compiled `master` when disowning all ships ahead of the flagship:
```
Thread 1 "endless-sky" received signal SIGSEGV, Segmentation fault.
0x00007ffff68407b0 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::size() const ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007ffff68407b0 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::size() const ()
   from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00000000004f1793 in Font::TruncateMiddle (this=0x1b07904, 
    str=<error reading variable: Cannot access memory at address 0x110>, 
    width=200) at source/Font.cpp:287
#2  0x000000000065e162 in ShipInfoPanel::DrawShipStats (this=0x3d75ee0, 
[...]

PR #2824, opened 2017-07-26
Fix crash when disowning all ships ahead of the flagship

tehhowch:
If the flagship was the last ship in the player's list of ships, and all ships ahead of it were disowned, the ShipInfoPanel would attempt to access invalid data.

Refs #2820


endless-sky:
What's actually going on is that I switched the player's ships from a list to a vector, and that made the iterators work differently. In a list, if you delete one item all the iterators keep pointing to the same thing (unless they pointed to the item you deleted), but in a vector all the iterators after the thing you deleted shift over one item. So, the correct solution is 5848f35.

Issue #2836, opened 2017-07-27
Health-based recloaking causes "pop-in / pop-out" behavior


tehhowch:
Reported on Steam: https://steamcommunity.com/app/404410/discussions/1/1458455461497401563/#c1471966894862283146

Video of the cloak/decloak/cloak loop.
https://youtu.be/IL0KMPFwp7k

Since the Archon (Cloaked) has free cloaking, it does not care about distance to target when deciding to de-cloak.
```
if(ship.Attributes().Get("cloak"))
	{
		// Never cloak if it will cause you to be stranded.
		if(ship.Attributes().Get("cloaking fuel") && !ship.Attributes().Get("ramscoop"))
		{
			double fuel = ship.Fuel() * ship.Attributes().Get("fuel capacity");
			fuel -= ship.Attributes().Get("cloaking fuel");
			if(fuel < ship.JumpFuel())
				return;
		}
		// Otherwise, always cloak if you are in imminent danger.
		static const double MAX_RANGE = 10000.;
		double nearestEnemy = MAX_RANGE;
		for(const auto &other : ships)
			if(other->GetSystem() == ship.GetSystem() && other->IsTargetable()
					&& other->GetGovernment()->IsEnemy(ship.GetGovernment())
					&& !other->IsDisabled())
				nearestEnemy = min(nearestEnemy,
					ship.Position().Distance(other->Position()));
		
		// If this ship has started cloaking, it must get at least 40% repaired
		// or 40% farther away before it begins decloaking again.
		double hysteresis = ship.Cloaking() ? 1.4 : 1.;
		double cloakIsFree = !ship.Attributes().Get("cloaking fuel");
		if(ship.Hull() + .5 * ship.Shields() < hysteresis
				&& (cloakIsFree || nearestEnemy < 2000. * hysteresis))
			command |= Command::CLOAK;
		
		// Also cloak if there are no enemies nearby and cloaking does
		// not cost you fuel.
		if(nearestEnemy == MAX_RANGE && cloakIsFree && !ship.GetTargetShip())
			command |= Command::CLOAK;
	}
```
[...]

PR #2837, opened 2017-07-27
bugfix: Ships which cloaked to repair should fully decloak after repairing

tehhowch:
**Refs #2836**
Instead of checking `ship.Cloaking()`, access the ships Commands to determine if it is cloaking/maintaining cloak, or is decloaking. This allows the higher hysteresis value to only apply when the ship is trying to cloak, rather than while the ship has any degree of cloak, preventing the ship from immediately recloaking if it takes any damage while decloaking.

Test save (courtesy of Steam user): 
On `master`, the Archon will cloak to repair after being reduced to health = 1, but then after repairing to health = 1.4, it will rapidly cycle its cloak if it is in your fleet's weapons range and takes damage before decloaking.
With this PR, the Archon has to be damaged back to health = 1 once it has started decloaking before it will cloak to repair.
[Travion Expectra.txt](https://github.com/endless-sky/endless-sky/files/1180121/Travion.Expectra.txt)

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1836,7 +1836,7 @@ void AI::DoCloak(Ship &ship, Command &command)
 		
 		// If this ship has started cloaking, it must get at least 40% repaired
 		// or 40% farther away before it begins decloaking again.
-		double hysteresis = ship.Cloaking() ? 1.4 : 1.;
+		double hysteresis = ship.Commands().Has(Command::CLOAK) ? 1.4 : 1.;
 		double cloakIsFree = !ship.Attributes().Get("cloaking fuel");
 		if(ship.Hull() + .5 * ship.Shields() < hysteresis
 				&& (cloakIsFree || nearestEnemy < 2000. * hysteresis))


source/AI.cpp
```diff
@@ -1836,7 +1836,7 @@ void AI::DoCloak(Ship &ship, Command &command)
 		
 		// If this ship has started cloaking, it must get at least 40% repaired
 		// or 40% farther away before it begins decloaking again.
-		double hysteresis = ship.Cloaking() ? 1.4 : 1.;
+		double hysteresis = ship.IsCloaking() ? 1.4 : 1.;
```
endless-sky: Would it work to just use `ship.Commands().Has(Command::CLOAK)` instead of adding a whole new function and member variable to Ship?

tehhowch: Probably. Though if ships have a disrupted cloak, they may have the cloaking command but not actually be cloaking.

endless-sky: I think that would be preferable, though. That is, if a ship has decided to cloak, it shouldn't stop unless it's done repairing, even if something does interrupt it.

endless-sky: The Ship class has such a huge collection of functions already that I just prefer not to add even more unless absolutely necessary.

tehhowch: Indeed, there are a great many there. For a game about spaceships though, it does kinda make sense!

tehhowch:
@endless-sky updated to use the ship's existing Commands().

PR #2844, opened 2017-07-28
Add cloaking heat attribute

Hadron1776:
This PR adds a `cloaking heat` attribute for further customization of outfits, (specifically, cloaking devices).

Requested by @TheUnfetteredOne

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -49,6 +49,9 @@ tip "cloaking energy:"
 
 tip "cloaking fuel:"
 	`Fuel per second required to maintain cloaking.`
+	
+tip "cloaking heat:"
+	`Heat generated per second while cloaked.`
 
 tip "cooling:"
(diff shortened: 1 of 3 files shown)


TheUnfetteredOne:
Thx bby

Hadron1776:
Yeah, totally forgot about that :P

tehhowch:
Question: should cloaking heat be added to idle heat display or not? I don't know if the energy it uses is, but I imagine including both could be warranted

Bladewood:
`cloaking energy` isn't.

Nor is something like `afterburner energy` and `afterburner heat`.

Hadron1776:
@tehhowch It doesn't look like the cloaking energy is added to the idle energy display.

tehhowch:
Maybe we need/want an "Activated outfit" line? For a different PR though.

Issue #2840, opened 2017-07-27
Unable to select wormhole not on jump path


Amazinite:
http://steamcommunity.com/app/404410/discussions/1/1471966894867277108/

To reproduce
1) Go to Cardea
2) Map a route that would lead you through one of the wormholes
3) Try to select the other wormhole

It would be expected that the wormhole you click on would be selected, but instead the game rapidly switches to selecting the wormhole on your jump path.

tehhowch:
Interestingly enough, you can hover over the "wrong" wormhole and then rapidly press L... you will start going to land on the "right" wormhole, but if you are fast enough, you will start shrinking (and rapidly accelerate too!), and travel through the "wrong" wormhole (ending up in its proper destination).

This has to do with the opening bits of AI::MovePlayer(), as it checks only that you have a travel plan defined, with no regard for if you have selected a TargetStellar that does not coincide with it.

There are no planets in the base game that define a wormhole that you can make travel plans through, and also have a planet you can land on, but I imagine this bug would prevent you from landing on said planet as well.

tehhowch:
(almost ready with a fix... just checking for unintended bugs)

tehhowch:
Fixed by #2845

PR #2845, opened 2017-07-29
Allow landing on planets or wormholes that are not in the TravelPlan

tehhowch:
Refs #2840 
If the player selects a landable object that is not part of the travel plan, it will have priority over the automatic landing target needed for completing the TravelPlan.

I also added a QoL feature, where if you press Jump when your next step is to land on a wormhole, a message about landing on the wormhole is displayed instead of the erroneous message about jumping to the system you are facing.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2375,7 +2375,8 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 				&& object.GetPlanet()->WormholeDestination(ship.GetSystem()) == system && player.HasVisited(system))
 			{
 				isWormhole = true;
-				ship.SetTargetStellar(&object);
+				if(!ship.GetTargetStellar() || keyStuck.Has(Command::JUMP))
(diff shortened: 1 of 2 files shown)


tehhowch:
There is one small issue remaining - if the player lands on a wormhole that is not a part of the existing travel plan, the current game logic does not update the travel plan to reflect that there is now a missing step between the player's current system, and their previously-desired travel route. Thus, if a player tries to jump without opening the map, they receive the "You cannot jump to the selected system" error message.

I plan to address this by performing a check in Engine::EnterSystem() that will clear the now-invalid travel plan, with the understanding that the player chose to not follow their travel plan, and thus it should be removed.

tehhowch:
As mentioned in #2840, this fix will also let the player land on a planet in the same system as a wormhole being used by the travel plan.
Without this PR, the player is unable to land on a planet if the player's travel plan is using a wormhole in that same system.

PR #2850, opened 2017-07-29
Ships that flash no longer repeatedly trigger the warning siren

Hadron1776:
Something I've noticed is that if one has the "warning siren" preference on, then hostile ships that flash on the radar (i.e. are overheated or have the `target` personality) will also trigger the warning siren... repeatedly. This PR stops that behavior by querying the ship's target, rather than its radar type.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1275,8 +1275,10 @@ void Engine::CalculateStep()
 			
 			double size = sqrt(ship->Width() + ship->Height()) * .14 + .5;
 			bool isYourTarget = (flagship && ship == flagship->GetTargetShip());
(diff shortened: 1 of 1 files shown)


tehhowch:
I wonder if instead you should include `Radar::Blink && ship->getgovernment()->is enemy()` rather than check the ship's target.

Hadron1776:
Checking the ship's target also stops the spam from being overheated... I could do
```
hasHostiles |= (type == Radar::HOSTILE || (ship->GetGovernment()->IsEnemy() && (type == Radar::BLINK || type == Radar::INACTIVE)))
```
But that would silence the alarm if there were any _disabled_ hostile ships in the system..

Issue #2855, opened 2017-07-29
NPC dialogs get offered multiple times if labeled evade


tehhowch:
When capturing an NPC labeled `evade` that has a dialog associated with its npc block, the dialog is shown every time,  provided the rest of the NPCs are disabled or destroyed.
The dialog is also not shown automatically if all NPCs are killed.

Contrast with `npc kill`, which only prompts the dialog when all of the npc block have been killed / captured, and not when any ship is captured.

Reference mission is "Wanderers: Sestor: Farpoint Attack 2" - if you defeat the fleet and don't capture one of them, you are not informed about the possibility of not using the bomb. However, if you capture one, you are (immediately) given the dialog.

tehhowch:
Note that for evade, the dialog should also be offered once all the ship are disabled, and not just destroyed. This dialog never appears.

tehhowch:
To the above point about the dialog never appearing when you complete the required flagged action for the npc block (in this case, `evade`): the dialog is shown when the act of recording the corresponding ShipEvent in the map of actions performed (aka past ShipEvents) for these NPCs causes the NPC block to go from failed to succeeded (or not failed to failed). To determine this, the success state is checked before adding the new event to the actions map, and then again after.
The success state is calculated by NPC::HasSucceeded:
```
	// Check what system each ship is in, if there is a requirement that we
	// either evade them, or accompany them. If you are accompanying a ship, it
	// must not be disabled (so that it can land with you). If trying to evade
	// it, disabling it is sufficient (you do not have to kill it).
	if(mustEvade || mustAccompany)
		for(const shared_ptr<Ship> &ship : ships)
		{
			// Special case: if a ship has been captured, it counts as having
			// been evaded.
			auto it = actions.find(ship.get());
			bool isCapturedOrDisabled = ship->IsDisabled();
			if(it != actions.end())
				isCapturedOrDisabled |= (it->second & ShipEvent::CAPTURE);
			bool isHere = (!ship->GetSystem() || ship->GetSystem() == playerSystem);
			if((isHere && !isCapturedOrDisabled) ^ mustAccompany)
				return false;
		}
```
spot the issue?

For `evade`, the npc is in a succeed state if every ship falls into one of these categories: 
1) in a different system than the player
2) destroyed
3) disabled

[...]

PR #2859, opened 2017-07-30
Fix NPC HasSucceeded() to use only recorded ShipEvent information

tehhowch:
Refs #2855 

NPC::Do calculates whether a given ShipEvent causes success or failure by comparing the value of HasSucceeded before and after recording the ShipEvent.
 - The previous implementation inherited knowledge about the ship's Disabled status from the ship itself, and not from the recorded information.
This meant that `evade` missions could never determine that the incoming ShipEvent::DISABLE was the one that allowed for success criteria.
 - Learning of a new ShipEvent::ASSIST will appropriately remove a ship's known ShipEvent::DISABLE flag.
 - It should now be possible to `evade` a ship which is `never disabled` by destroying it (in addition to being in a different system)

This PR resolves all the issues I mentioned in 2855
 - `npc evade` dialogs are shown when all required evaded npcs are evaded.
 - capturing each npc ship when all have been disabled / destroyed will not throw the dialog again
 - 'derelict' ships are correctly counted as unable to accompany if they have not been assisted.

I do not address the bug in which an assisted `derelict` NPC is spawned as `derelict` again if it was previously assisted and the player landed when it was in a `not disabled` state, then restarted the game, or that a derelict ship's hull/shields do not regenerate if it lands with you and then you take off again. These should be solvable by referencing the NPC actions table when doing ship placements / personality assignments (in a different PR).

### DIFF
--- a/source/NPC.cpp
+++ b/source/NPC.cpp
@@ -249,6 +249,10 @@ void NPC::Do(const ShipEvent &event, PlayerInfo &player, UI *ui, bool isVisible)
 		if(bay.ship)
 			actions[bay.ship.get()] |= type;
 	
+	// If this event was "ASSIST", the ship is now known as not disabled.
+	if(type == ShipEvent::ASSIST)
+		actions[ship.get()] &= ~(ShipEvent::DISABLE);
+	
 	// Check if the success status has changed. If so, display a message.
 	if(HasFailed() && !hasFailed && isVisible)
 		Messages::Add("Mission failed.");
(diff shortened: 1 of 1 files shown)


endless-sky:
The "evade" condition requires that all NPCs leave the system or be disabled. That means that it's possible a ship will leave the system (meaning, you've successfully evaded it) but then return (so, you have no longer evaded it). Similarly, it's possible (although much less likely) that a ship will be disabled but later repaired.

tehhowch:
Are you saying I should update this to compare before/after on the system too? I thought only the flagship generates jump events, so NPC::Do (and thus HasSucceeded) would not get called.

Regarding the assistance of a disabled ship - is that not already addressed?

tehhowch:
Here is a save file that allows one to reproduce the issue described by #2855 and fixed by this PR: 
[Travion Expectra-Defend Farpoint.txt](https://github.com/endless-sky/endless-sky/files/1203302/Travion.Expectra-Defend.Farpoint.txt)

When using `master`, every time you capture a ship when the rest of the ships are disabled/destroyed, you will get the notification dialog for NPC evade success. You will not receive the notification dialog if every ship is destroyed instead of being captured. 
When using this PR, you will get the notification dialog only upon disabling every ship (capture not necessary), and you will not receive it upon capturing any of the ships.
I've disabled the self-destruct on each of the automata in the mission so that they can be more easily captured and expose any existing issues.

tehhowch:
Here is a test mission suite that allows one to test evade, accompany, and assist:
[accompany_test_mission.txt](https://github.com/endless-sky/endless-sky/files/1203459/accompany_test_mission.txt)

Issue #2846, opened 2017-07-29
How many ships should a stranded / disabled ship ask for help?


tehhowch:
Currently, they will ask as many as possible, 1 per frame.
If the player has a large number of escorts with 2 fuel amounts (say 50% 600 and 50% 1000), when the 600-fuel ships are down to a single jump they will each spam the rest of your fleet with assistance requests. If your ships arrive asynchronously (pretty common since fleet jump is not the default behavior), then the first low-fuel escorts will ask the majority of your escorts for help, even though only 1 needs to help.

This can really quickly murder the frame time, as 50% of your escorts are checking every escort to see if the other has fuel for them, and also if they are not helping anyone else.
(with 75 escorts, this "time to refuel" issue spikes CPU time from 4% to 60%+)

A simple solution is to add a counter during the "can anyone help me" part of AI::Step, and anytime a ship is found to already be helping the stranded ship, to increment this counter. After N ships are found to be helping, no more ships should be asked or checked to be asked.

What is a reasonable value of N? Bear in mind that the AI will move to help anyone it said it would help before going into combat (and also that a ship won't ask for help if **it** has enemies to fight, and if disabled, only same-government allies will be asked to help).

I think 3 is probably good.

tehhowch:
Here's a test save which illustrates the CPU load from refueling: 
[Travion Expectra.txt](https://github.com/endless-sky/endless-sky/files/1184618/Travion.Expectra.txt)
Launch from the planet, hit jump. When you arrive in Cardea, take note of the CPU load. It spikes to about 67% on my laptop (i7-4500U 4x@1.8GHz)

tehhowch:
It may be that a better solution is an entire rework of the "can anyone help me" block, as if the majority/all of the ships that can be asked are incapable of helping, implementing a limit on the number of ships to ask - even if the knowledge of this value persists across AI::Step calls - will not reduce the number of checks being made, since the count is still 0 (or 1 or 2).

tehhowch:
With #2861 merged, the answer is "1". It's also now a bit more likely to be a fast ship rather than a slow freighter 👍

Issue #2835, opened 2017-07-27
Notably slower/higher CPU load (but playable after dropkicking my laptop): 0.9.7 v 0.9.6 (Windows64 bit)


marphod:
Updated to 0.9.7, and the game is nearly unplayable f i encounter more than a handful of hostile ships,

My system is a rather modest  windows laptop (Celeron N3050@1.6 GHz, 4G RAM, integrated graphics), but 0.9.6 was playable.  large combats had a slowdown, which was annoying but live-able.  the internal CPU monitor would spike to 90-120% (I assume that 100% is 1 core fully utilized, but i'm not sure) and GPU would spike to the 70-80% range (normal was 40% CPU and 30% GPU).  Only Performance option were hyperspace flash and showing the cpu/gpu status.  

with 0.9.7, the baseline is 70-80% cpu and 40% gpu. and the game feels sluggish. Modest fleet combats (one kor mereti, my fleet of 25) become ridiculously slow, (effectively sub-20 fps, choppy, some delays in sound, running at what feels like 1/3rd speed) and reported CPU of over 600% (i'm not sure what that  even means on a dual-core system) and 180% GPU, even when most of the ships were off-screen.   Turning on time warp returned to only slightly slower than the new normal, but was in single-digit Fps.  For what it is worth, speed slightly improved, i think, with 'focused' turrets rather than independent.  Turning off the minimap or labels did not appear to change anything.

Background apps and os activity did not change between runs.

What stats/debug logs can i generate to help?

tehhowch:
CPU / GPU time indicate how many 60ths of a second is needed to compute / draw 1 frame. so 100% utilization means you need 1 frame to calculate or draw 1 frame. If you are playing self-compiled, please be sure you are not running a debug exe. Expected x64 Windows 10 EndlessSky.exe sizes are in the ballpark of 1700 - 2100 kB, while a debug build will be in the 120 MB range. (`scons` output for me on linux gives a 2.4 MB binary and 46 MB debug binary).

You may obtain some improvements by changing/adding a compiler flag, `-march=native` to the build, but by and large without improvements to the game engine your choices are limited.

You can test out my changes in #2788 to see if they improve performance, but I think for such small lists there will not be much improvement.

MCOfficer:
[...]

PR #2861, opened 2017-07-30
Refactor AI::Step's "Find someone to assist me" section

tehhowch:
Ref #2835 , #2846, perhaps some steam issues.

This PR dramatically reduces the game lockup seen when entering an uninhabited system with a large number of out-of-fuel escorts, to the point where it is almost indistinguishable from that of the normal game.

Summary
 - Move "can anyone help me" code into into separate function
 - Use std::vector to store possible choices instead of 2 pointers.
 - Randomly choose a possible helper (must check all ships in the system in order to determine if there is an enemy present)
 - Reduce frequency of calls to avg. 2 per second rather than every step for every stranded/disabled ship.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -36,6 +36,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include <cmath>
 #include <limits>
 #include <set>
+#include <vector>
 
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


endless-sky:
Would this make it so that twice a second, each ship can potentially change which ship it's asking for help? The advantage of the previous system was that in general, a ship will ask the same ship for help over and over gain.

Also, this would make it so that a ship that's out of fuel will come to a stop as if waiting to be boarded even if it has not succeeded in requesting help from anyone (because 29 frames out of 30, it skips the check and just assumes someone is coming to help it).

tehhowch:
I believe those concerns are now addressed:
 - the ship that gets asked is stored in a map, and removed when that ship decides that it cannot help the requestor.
 - a ship with no one helping it will not stand still until it finds someone to help it.

Also, the current implementation on master will never set `selectNext` to true, because at that point in the code `ship` and `it` can not be the same ("Never ask yourself for help"). This bit hadn't been updated since its original addition in https://github.com/endless-sky/endless-sky/commit/d87f009b55b6faeb092a8ed92dc57e26c3f150ab :)

Issue #2866, opened 2017-07-31
FIrst Contact: Remnant log entry should be in the on decline


Amazinite:
The information in the log is learned after you take the blood test, but should the player run then they wouldn't have learned the information which is written in the faction log on the Remnant. To avoid this, the log should be moved from the `on offer` to the `on decline`.

Hadron1776:
> but should the player run then they wouldn't have learned the information which is written in the faction log on the Remnant.

Seeing as "the player runs" triggers `accept`, I don't think this would solve the problem. I thought `on offer` blocks were triggered immediately when the mission is offered, anyway?

Amazinite:
>Seeing as "the player runs" triggers accept, I don't think this would solve the problem. I thought on offer blocks were triggered immediately when the mission is offered, anyway?

Running triggers `accept`. Accepting the blood test and learning the history of the Remnant triggers `decline`. As such, the log should be in the `on decline` and not the `on offer`, as placing it in the `on offer` will create the log whether you actually learn the history of the Remnant or not.

Hadron1776:
Oh, ok. I think I misread your OP then :)

tehhowch:
There should probably be an `on accept` log for a hostile Remnant faction if the `on offer` log becomes the `on decline` log.

PR #2873, opened 2017-08-01
Update Remnant faction knowledge based on friendly/hostile relationship

tehhowch:
Refs #2866 

 - Move the existing faction log for the Remnant to the `on decline` so that it is learned only if the player could logically have learned it.
 - Created a replacement faction log with the observed information at the time of flight, if the player refuses.
 - Added a regular log message so that the player can remember when the Remnant was first met.

If a method to improve Remnant rep is added, a mission which adds the missing information to the special logs can be added when the player is able to learn of the Remnant's history.

### DIFF
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -56,8 +56,8 @@ mission "First Contact: Remnant"
 	on accept
 		"reputation: Remnant" <?= -10
 		log "Refused to give a blood sample to a group of strange-looking humans that live south of the Core. Upon takeoff, their ships attacked."
-		log "Factions" "Remnant" `A group of unusual-looking humans that live in a secluded section of the galaxy. They are openly hostile to those who refuse to submit blood samples.`
-			`The buildings of the Remnant people seem to be alien in origin, and they communicate amongst each other primarily with sign language. When they do use their voices, they sing or chant, rather than speak.`
+		log "Factions" "Remnant" `The Remnant are a group of unusual-looking humans that live in a secluded section of the galaxy. It is unclear how long they have lived apart from the rest of humanity, but they have been free of human influences for long enough that their culture is wildly different from any found in human record. Their buildings seem to be alien in origin, and they communicate amongst each other primarily with sign language. When they do use their voices they sing or chant, rather than speak.`
+			`The Remnant seem fearful - or at least especially wary - of contact with humans. Visitors that do not undergo medical examination are not allowed to remain in their territory.`
 	on decline
 		event "ember waste label"
 		log "Met a group of strange-looking humans that live in a secluded section of the galaxy south of the Core. After submitting a blood sample, they became very friendly."
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -14,8 +14,6 @@ mission "First Contact: Remnant"
 	source
 		government "Remnant"
 	on offer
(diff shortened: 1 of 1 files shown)


data/remnant missions.txt
```diff
@@ -57,8 +55,14 @@ mission "First Contact: Remnant"
 				decline
 	on accept
 		"reputation: Remnant" <?= -10
+		log "Refused to give a blood sample to a group of strange-looking humans that live south of the Core. Upon takeoff, their ships attacked."
```
Amazinite: So tall, dark, and handsome is strange-looking to you? :) /s

marphod:
I'd suggest going with describing the Remnant as 'human-like' rather than 'human' on the refusal -- we know that not everyone who looks human is 100% Homo sapien as it is.

tehhowch:
"ancestral tongue" + communicates in the player's language, ergo Remant are the player's species, and the player is human in the base game.

marphod:
The ancestral tongue of the Alphas and so-called 'Betas' would be the same, and at least the Betas (and one presumes some Alpha -- given the bloodtest here and the Naval Intel test earlier) look Human.

tehhowch:
What noun would you use instead of human? Humanoid? Alien?

marphod:
If writing in a void, I'd probably go with 'strange, human-like sentients', but that would be unprecedented in the game.  'Inhabitants' would be the most consistent with other descriptions in the game.

jafdy:
I would describe the Alphas and Betas as human, and as tehhowch says they use the same language. 
The blood test presumably will do the same thing they do now except they are looking for something characteristic about Beta blood instead of just blood type.

data/remnant missions.txt
Bladewood: ~~yes~~

Bladewood:
sentient - can feel and experience instincts
sapient - actual higher thinking

data/remnant missions.txt
KiLEdEnNis: I don't think changing humans to negros is any better lol

Amazinite: Oy vey, calm down Grand Wizard Kile.

KiLEdEnNis: it's just a matter of using a decent fire spell...

data/remnant missions.txt
```diff
@@ -57,8 +55,14 @@ mission "First Contact: Remnant"
...
+		log "Factions" "Remnant" `A group of unusual-looking humans that live in a secluded section of the galaxy. They are openly hostile to those who refuse to submit blood samples.`
```
endless-sky: The last part of this - "They are openly hostile to those who refuse to submit blood samples" - seems a little too specific. It would be like saying, "The Wanderers are hostile toward people who land starships in their gardens": it's probably true, but isn't necessarily one of the faction's most defining characteristics.

It might make more sense to just not get a faction log for the Remnant if you make them hostile.

tehhowch: It may seem a bit strange for the player to have labeled planets/systems, but no faction log for this group, while every other faction - hostile or not - has a faction log (automata excluded, I think). With that in mind, I wanted to offer some kind of faction log to the player. A single sentence indeed seemed too short, and there isn't much else the player knows about the Remnant to include (including the motivation behind the the blood test), if the player runs.

I'm open to suggestions for edits, and if you've got a particular design in mind, I welcome the touchup merge commit 👍  After all, you have a better idea of what's next for a player who has a hostile rep with the Remnant than we do 👼

tehhowch:
Updated the hostile mission branch's "Factions" log to be a little less pointed.

PR #2875, opened 2017-08-01
Extend GameEvents via MissionAction capabilities

tehhowch:
~~This PR would allow a game event to add information to a player's logbook, under both the timestamped logbook, and the Faction/People section.~~

~~This makes it possible to insert entries into a player's logbook without requiring the player to be in a position to have a mission offer. For example, timed events associated with a "reminder mission" can also insert a notice into the player's logbook, so if they are not in a region of space that allows the reminder / followup mission to activate, the entry in the player's logbook can help them remember that they should return.~~

This PR
 - adds a MissionAction to every GameEvent, allowing actions typically reserved for missions to be performed in a game event.
    - logs
    - special logs
    - outfit transfer (gifting)
    - payment
    - scheduling more events
    - failing other missions
 - adds reputation update checking to conditions set via GameEvent (a bugfix, I believe)


endless-sky:
The logbook is not the right place for reminders; there's no way for the player to know when something new has been added there. In general, it seems weird to add something to the log when no message has been shown to the player to give them a clue that something has happened.

Missions can trigger any time you land, so the only way this would be useful is if you wanted to log something when the player is flying long-distance without landing. And that's a small enough use case that it doesn't seem worth adding.

tehhowch:
I do admit that there is a low use case count for using an event to add a log or special log, given the current state of logs.  I'll keep this open in case any modders wish to articulate how they would use this feature, or you find a case where adding a log or special log via event makes more sense than piggybacking it onto a mission.

If the log feature gets extended to incorporate text substitutions, I'll keep this up-to-date as needed.

endless-sky:
I'm not sure there's much of a use case for this, but if you *were* to do it this would result in a lot of code duplication between GameEvent and MissionAction - it might be cleaner to just have GameEvent contain its own MissionAction instead of just its own ConditionSet and logs. I could imagine some other uses for that, e.g. "gifting" a tracking device or a bomb to the player's flagship right as it takes off so there's no way the player can see it beforehand.

tehhowch:
That's a neat idea, and as of da9b7e5, is now done.

In order for conversations and dialog to be displayed via the GameEvent's MissionAction, PlayerInfo::IncrementDate() needs a UI reference. Adding one would (I think) require piping a pointer through Engine::CalculateStep() and Engine::EnterSystem(), or adding the pointer directly to PlayerInfo.
Given the lack of detailed control over event occurrence (e.g. LocationFilters and `to offer` ConditionSets), it's probably best to keep the current mission system for displaying conversations and messages.

endless-sky:
As I think about it more, I just don't think this is useful. Not being able to show dialogs or conversations also means you will almost never want to give payments or outfits (because there would be no way to tell the player they had been given). That means the only MissionAction functionality that might be useful is logging, triggering other events, and failing missions. And those could almost always be done directly by the mission that triggered the event.

Amazinite:
>allowing actions typically reserved for missions to be performed in a game event.
>* logs
>* special logs

>I'm not sure there's much of a use case for this
>I just don't think this is useful

An event could be used to trigger logs similar to how conversations can be stored on their own and triggered by a mission. The last mission of each James branch triggers the same conversation so that there doesn't need to be a lot of copy-pasted dialog. An event could be used in the same way to trigger a lot of log entries to reduce the need to copy and paste.

Issue #2277, opened 2017-03-19
Have missiles disengage once you bribe a faction


Amazinite:
A problem I've noticed right now with bribing is that if you (or the enemy) use missiles (and some slower firing projectile weapons) then once you bribe them, the missiles will continue to track their target and eventually do damage if they hit (and projectiles will do damage as well if they are aimed right at a ship that just became friendly). This causes two problems.
1) If it's the enemy using the missiles and slower moving projectiles, then it can kill you or ships in your fleet after you've bribed the enemy to become friendly.
2) If you're the one using missiles, then bribing the enemy while missiles are still flying can cause them to become hostile again if any of them hit, requiring you to bribe again, but in that time more missiles were fired, so sometimes it becomes a perpetual loop where you're unable to bribe the enemy to stop attacking you.

Given this, it would be helpful if missiles disengaged their targets and projectiles would no longer collide with what were enemies (or your own ships) after they've been bribed, or at least have the enemies not become hostile to projectiles that were fired before the bribe.

Bladewood:
Yeah, I've noticed the second one several times with some weapons while attempting to bribe pirates in Durax.

KiLEdEnNis:
projectile weapons in general;
once I bribed a pirate like 100x for fun because they kept becoming angry & loaded game when I got bored

LocalGod79:
or you could tap "u" so that your fleet stops shooting missiles before you try to bribe...

Bladewood:
Again, that is just for missiles and other limited ammo weapons. It still exists with other weapons.

Amazinite:
And usually when I'm bribing, it's in a situation where I'm close to death, so I don't have the time to stop my fleet from firing then wait for all the missiles to hit or run out of lifetime, and then bribe. This issue was mainly brought up for the first point anyways, which the player has no control over. 
I enter a system in my little Star Barge, I see a pirate ship fire a few torpedoes, I bribe him so we're friendly, but then the torpedoes still chase and eventually hit me, killing me, despite me having bribed the pirate.

endless-sky:
That would only make sense if all projectiles were somehow remotely controlled by your ship, which they're not. If a missile is in the air, it's going to keep trying to do what it was originally told to do.

Amazinite:
[...]

PR #2878, opened 2017-08-01
enhancement: Disable missile tracking for bribed government

tehhowch:
Bribing a government causes all homing projectiles that are targeting player ships to stop *tracking* the player's ships (e.g. become ballistic). If they hit the player's ship, they will still cause damage, but the missiles will not actively attempt to hit the player.

Refs #2277 
Also refs https://steamcommunity.com/app/404410/discussions/0/1471966894884358249/

Given that this method is only called when disengaging a blocking UI panel, there should be no impact on normal game load during combat.


endless-sky:
There's still no good explanation for why all missiles would magically know to stop tracking you just because you bribed the ship that fired them. As it is, bribery is way too powerful - I don't think it's a problem to expect someone to also learn how to dodge missiles, instead of just bribing their way out of every encounter with pirates.

tehhowch:
I think the tech to disarm/disable the targeting circuits/computer in a launched missile in the year 3013+ is a fraction of the cost of the explosives in the weapon. We know that ships in the ES universe can communicate across the entire system in an instant (and even conduct banking transactions instantaneously). 
Surely it is not hard to believe that paying money for your safety could result in the bribed government using a very low tech failsafe in its munitions? 

In regards to this PR adding on to the perceived overpowered nature of bribing a hostile government to spare your life, perhaps bribing needs a chance to fail? You've expressed that the characters in the game do not always tell the whole truth (sometimes willingly, sometimes not), so it seems sensible to extend this deceit to interactive NPCs, and especially pirates.
Hail a pirate -> pay a bribe -> roll against `(2 * bribefraction)` -> fail roll -> "thanks for the money, chump. Care to part with a little more, since you seem so eager to live?`

tehhowch:
With bde378c I address the increased survivability, and make NPCs a bit more realistic, by having the bribe action fail randomly, in accordance with `sqrt( bribefraction / 2 )`. Most bribe fractions are 0.05, which means about 16% of the time, those bribes would fail, and the bribe will go up. as well. Safety can expensive 👍

Amazinite:
Maybe have the % chance for a bribe to fail be its own government attribute? `sqrt( bribefraction / 2 )` means that militia/FW ships have a 22.3% chance to fail the bribe, which I find a little odd because that makes pirates (at 16%) more trustworthy than the militia/FW. Ever failing a bribe with the Hai (which wouldn't be too uncommon at its current 10% value) also seems a little out of place, especially since the player usually needs to bribe their way onto Hai-home when they first entire Hai space. It would suck for them to have to pay twice for a single landing.

tehhowch:
It could make sense to do that, though it may make more sense to revisit the current bribefraction values to achieve the same thing without using a new value.

Hadron1776:
Also: New Alexandria would need to be bribed if one does the Wanderer missions right after finishing the FW storyline. It seems wrong to have bribing the Republic fail.

tehhowch:
> It seems wrong to have bribing the Republic fail.

True enough, that would seem odd, given that they are supposed to [always follow through on their promises](https://github.com/endless-sky/endless-sky/wiki/MajorOrganizations#the-navy). Having only ship bribes possibly fail would solve the landing clearance issues, and, given the threat made in FWC, make more sense than a planet trying to stiff-arm a ship in orbit. But of course if we think too much about that, we should never be restricted from landing on any planet, since we could just bombard it...

Combining ship-only chances to fail bribing and specifying the failure chance in the government definition would allow covering all the bases regarding what is logical for a faction. Thoughts?

marphod:
The Republic isn't the Navy, but that's a side point.

I can go with only ship bribes might fail, and having a separate stat for failure chance.  If you were going to avoid a new value, I would suggest that the higher the bribe rate, the less likely a bribe will fail. `bribe rate <.2 ? sqrt(.2-bribe rate) : 0` could be more reasonable.

FWIW, most guided munitions these days can abort (usually, self destruct) upon a signal from the launching weapon system.  It would be weird for that technology to go missing in the future.

tehhowch:
- I've changed the Planet and Government Load() methods to pull in a second value on the `bribe` token to serve as the chance to fail when bribing the relevant target. The default is 0, but this allows specifying "untrustworthy" planets or ships for which bribes may fail.
 - I've increased the bribe fractions for the Pirate and Unfettered factions to match that of the Syndicate (8% net worth), and given them failure rates of 50%, 30%, and 20% respectively.
 - For other governments which accept bribing, I have given them a failure rate of 10%.
 - Only pirate planets were given bribe failure rates, of 25%.
 - The bribe increases by the twice the bribe fraction - e.g. default Pirate bribe would be 8% of your net worth, then it will be 8% * 1.16 of your net worth, then 8% * 1.16^2 .... and so on.

data/hails.txt
```diff
@@ -3023,17 +3023,47 @@ phrase "hostile free worlds"
...
+		"You seem pretty gullible."
+		"Wow, you paid up!"
+		"Credits? What credits? Your last transmission cut off early."
+		"Your pockets would look even better if they were a just little lighter."
+		"Sadly, <old bribe> credits was a little low."
+		"Great, <old bribe> credits will really help pay for this ship. We've got other bills to pay though."
+	word
+		" "
+	word
+		"Tell you what, if you really want to be left alone, you need to pay at least <new bribe> credits."
```
Bladewood: > if you really want to be left alone

Sounds a bit weird. Perhaps we can replace it with the standard
> if you really want us to leave you alone

?

data/hails.txt
```diff
@@ -3023,17 +3023,47 @@ phrase "hostile free worlds"
...
+# <new bribe> -> the amount that is now required
+# <old bribe> -> the amount that was just paid
+phrase "failed ship bribe"
+	word
+		"It's so nice to work with an honest captain."
+		"I can't believe that worked!"
+		"You seem pretty gullible."
+		"Wow, you paid up!"
+		"Credits? What credits? Your last transmission cut off early."
+		"Your pockets would look even better if they were a just little lighter."
```
KiLEdEnNis: "a just little lighter"
*just a little lighter

marphod:
A 50% failure rate seems really high.  It makes bribes not about avoiding conflict and saving effort but about a desperation move.  It makes the early game harder with little to no impact on the late game, when the player has spare cash.

(Also, from a real world perspective, that high a failure rate is unlikely to encourage more bribes and would end up being self-defeating)

It also means that >6% of the time, 4 bribes in a row will fail.  Which seems really problematic.

Lineth:
Yeah, 50% is far too high. I'd say 30% should be the absolute upper limit. Probably more like 20%.

tehhowch:
And bribing probably shouldn't be about avoiding conflict and saving effort, but a last ditch effort to stay alive. 
As stated above it is already viewed by MZ as an overpowered tactic, so to balance it requires increasing the cost of the action. If one wants to also add the disabling of incoming missile tracking, then the cost must increase quite a bit.
The cost of bribery could also be increased by expiring them in-flight after 5 minutes.

You'll note that Pirates - the most loosely defined faction in the game - are the only ones with such a penchant for taking more money. To me it certainly seems realistic - that pirate is almost certainly going to be killed in conflict very soon. Why not earn a little extra dough?
Every other faction has a bit more structure, a bit more guarantee of stability and life expectancy. And as such, they behave a little more sensibly.

Rob59er:
30% "accidents happen"
"we need to increase rates in order to cover the costs"
For the syndicate

tehhowch:
>30% "accidents happen"
"we need to increase rates in order to cover the costs"
For the syndicate

There's no government-specific bribery failure message, and I'm not sure that adding one would be sensible - I certainly wouldn't be able to come up with enough flavor text to justify it. I can include suggestions, but they should make sense with regards to the existing bribe fail messages I've added - either an introductory clause that comments on the player attempting to pay them off, or the closing clause which asks for more money.

endless-sky:
I'm not convinced that this would really make the game more fun or interesting. Because it's a random effect that the player can do nothing to alter, it's more likely to just be annoying.

KiLEdEnNis:
what if bribes have a fail chance after a certain negative reputation, and the failure chance goes up the lower your reputation goes?
say if:
rep = -5000 ya have 100% success rate
rep = -10000 ya have 75% success rate
rep = -15000 ya have a 30% success rate etc.
if that'd be a thing, it should go in some smooth curve, but modders should also have the option to change the chances of failure, and I don't know if that'll be do-able at all

tehhowch:
Making the failure chance influenced by the player reputation with the affected party could work. The better your reputation, the less likely a bribe is to fail - after all, you have good standing with them - while the worse it is, the more likely they are to just take your money (since their allies won't care if you are "taken care of")

I'll look into something along those lines.

KiLEdEnNis:
it can go like 4 ways basically
- does not want your money, and said ship will not accept it if you retry enough
- takes your money once, then gives some message if you retry that basically means "you got scammed, I'm not gonna let you go)
- takes your money, and you need to retry to succeed (I think this would be a bit annoying)
- takes your money infinitely (even more annoying)

another idea that can go along with this; the worse your reputation, the higher the bribe, but it depends on which of those 4 "paths" you wanna go

Pointedstick:
> I'm not convinced that this would really make the game more fun or interesting. Because it's a random effect that the player can do nothing to alter, it's more likely to just be annoying.

I disagree. It adds flavor, and communicates something about the game world and people's trustworthiness. Definitely an improvement for Firefly/Han Solo mode gameplay IMHO.

endless-sky:
This is really two separate changes: 1. giving bribes a chance of failing and 2. disabling homing the moment you bribe a government.

Disabling homing could be done just by clearing the projectile's targetShip, without any need for introducing a new variable in Projectile.

tehhowch:
I'll trim the Projectile tweak to clear out the targetShip. The dual nature of this PR originates from offering some balance - making bribing more effective at ensuring your safety by disabling projectile tracking, yet also making bribing less effective as it might fail, and thus the enhanced safety costs more.

PR #2881, opened 2017-08-02
Add a Minelayer version of the Model 8 to the Kor Mereti fleets

Rob59er:
Add rare to medium spawning minelayer based Model 8s as well as rare Slicer and Minelayer variants fleets. Allows the player to gain mine racks without hacks or potential Mereti outfitters. Also gives those pesky Sestor fighters something to chase. Smaller engines were given to avoid too much random mines.

### DIFF
--- a/data/korath ships.txt
+++ b/data/korath ships.txt
@@ -902,6 +902,21 @@ ship "Model 8"
 	"final explode" "final explosion small"
 	description "The primary role of the smallest Kor Mereti drones is asteroid mining, but their lasers can also be employed as offensive weapons when necessary."
 
+ship "Model 8" "Model 8 (Minelayer)"
+	outfits
+		Hyperdrive
+		"Generator (Candle Class)"
+		"Generator (Furnace Class)"
+		"Systems Core (Small)"
+		"Reasoning Node"
+		"Thruster (Asteroid Class)"
+		"Steering (Lunar Class)"
+		"Korath Minelayer"
+		"Korath Mine" 26
+		"Korath Mine Rack"
+	
+	gun "Korath Minelayer"
+	gun 
 
 ship "Model 16"
 	sprite "ship/model 16"
--- a/data/korath.txt
+++ b/data/korath.txt
@@ -456,6 +456,12 @@ fleet "Small Kor Mereti"
 		"Model 32"
 	variant 5
 		"Model 64"
+	variant 4
+		"Model 16" 3
+		"Model 8 (Minelayer)" 5
+	variant 1
+		"Model 8" 5
+		"Model 8 (Minelayer)" 3
 
 fleet "Large Kor Mereti"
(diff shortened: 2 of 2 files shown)


solardawning:
"Allows the player to gain mine racks without hacks or potential Mereti outfitters."

^just wanted to point out that the Kor Efreti outfitters sell mine and piercer missile racks.

Please-Close-This-Issue:
oh huh

Rob59er:
Oops, but at least the Mereti now make use of it.

data/korath ships.txt
```diff
@@ -902,6 +902,21 @@ ship "Model 8"
...
+		Hyperdrive
+		"Generator (Candle Class)"
+		"Generator (Furnace Class)"
+		"Systems Core (Small)"
+		"Reasoning Node"
+		"Thruster (Asteroid Class)"
+		"Steering (Lunar Class)"
+		"Korath Minelayer"
+		"Korath Mine" 26
+		"Korath Mine Rack"
```
KiLEdEnNis: the 1st ammo storage NPC's use
not gonna make another PR to add them to everything else though

endless-sky:
Model 8's were basically just intended to be mining ships. (That is, when a Mereti Drone is very small, it practices on asteroids rather than getting involved in real fights.) I kind of feel like they're more interesting as a separate function of ships rather than just one more tier in their warship hierarchy.

MessyMix:
While it isn't a bad idea imo, what MZ just said is actually pretty interesting! I would stick with that instead of trying to make minelayer variants, but it would've been a nice touch if it weren't for the (newly introduced) lore.

KyroFirehammer:
While on the subject shouldn't the line and piercer  ammo be purchaseable from the regular friendly korath if they sell the storage racks for them.

tehhowch:
If you'd like to discuss that, open an issue about it and provide good
arguments for why the ammunition of weapons exclusive to the Mereti and
Sestor factions should be sold at a station for the Kor Efreti.
This pull request is not the place to discuss that.

data/korath ships.txt
marphod: I assume this uses the Asteroid Thrusters due to space concerns; I was hoping that this would be another ship that used the Comet Thrusters (as the standard Model 8). Currently, the (standard) model 8 is the only ship to use them.

tehhowch:
As I recall, this was done because the Minelayer-only armament triggers the
missile boat AI behavior. Making its thrust low results in it being closer
to the action, and actually drop mines that hit the enemy :)
With only 1 weapon space left there isn't room for a non-homing, non-long
range weapon to avoid the missile boat AI.

data/korath ships.txt
Bladewood: It uses asteroid thrusters because otherwise the minelayer doesn't work properly.

Issue #2884, opened 2017-08-03
Compile failure with latest build (2017-08-03)


Bladewood:
```
||=== Build: Release in EndlessSky (compiler: GNU GCC Compiler) ===|
C:\Users\litot_000\Documents\GitHub\endless-sky\endless-sky\source\AI.cpp||In member function 'void AI::AskForHelp(Ship&, bool&, const Ship*)':|
C:\Users\litot_000\Documents\GitHub\endless-sky\endless-sky\source\AI.cpp|698|error: no matching function for call to 'std::vector<Ship*>::insert(std::vector<Ship*>::const_iterator, double, Ship*)'|
C:\Users\litot_000\Documents\GitHub\endless-sky\endless-sky\source\AI.cpp|698|note: candidates are:|
C:\Program Files\mingw64\x86_64-w64-mingw32\include\c++\bits\vector.tcc|107|note: std::vector<_Tp, _Alloc>::iterator std::vector<_Tp, _Alloc>::insert(std::vector<_Tp, _Alloc>::iterator, const value_type&) [with _Tp = Ship*; _Alloc = std::allocator<Ship*>; std::vector<_Tp, _Alloc>::iterator = __gnu_cxx::__normal_iterator<Ship**, std::vector<Ship*> >; typename std::_Vector_base<_Tp, _Alloc>::pointer = Ship**; std::vector<_Tp, _Alloc>::value_type = Ship*]|
C:\Program Files\mingw64\x86_64-w64-mingw32\include\c++\bits\vector.tcc|107|note:   candidate expects 2 arguments, 3 provided|
C:\Program Files\mingw64\x86_64-w64-mingw32\include\c++\bits\stl_vector.h|988|note: std::vector<_Tp, _Alloc>::iterator std::vector<_Tp, _Alloc>::insert(std::vector<_Tp, _Alloc>::iterator, std::vector<_Tp, _Alloc>::value_type&&) [with _Tp = Ship*; _Alloc = std::allocator<Ship*>; std::vector<_Tp, _Alloc>::iterator = __gnu_cxx::__normal_iterator<Ship**, std::vector<Ship*> >; typename std::_Vector_base<_Tp, _Alloc>::pointer = Ship**; std::vector<_Tp, _Alloc>::value_type = Ship*]|
C:\Program Files\mingw64\x86_64-w64-mingw32\include\c++\bits\stl_vector.h|988|note:   candidate expects 2 arguments, 3 provided|
[...]

PR #2885, opened 2017-08-03
Fix bugs in two recent commits

tehhowch:
#2884 

I used `cend()` instead of `end()`, which fails on some platforms but not others, in #2861 

@endless-sky 's touchup in 381b01fa

This is a subtle nudge for #2786, which would have flagged my PR

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -695,7 +695,7 @@ void AI::AskForHelp(Ship &ship, bool &isStranded, const Ship *flagship)
 				continue;
 			
 			// Prefer fast ships over slow ones.
-			canHelp.insert(canHelp.cend(), 1 + .3 * helper->MaxVelocity(), helper.get());
+			canHelp.insert(canHelp.end(), 1 + .3 * helper->MaxVelocity(), helper.get());
 		}
 		
 		if(!hasEnemy && canHelp.size())
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1292,7 +1292,7 @@ void Engine::CalculateStep()
 			
 			double size = sqrt(ship->Width() + ship->Height()) * .14 + .5;
 			bool isYourTarget = (flagship && ship == flagship->GetTargetShip());
-			hasHostiles |= (!ship.IsDisabled() && ship->GetGovernment()->IsEnemy()
+			hasHostiles |= (!ship->IsDisabled() && ship->GetGovernment()->IsEnemy()
 				&& ship->GetTargetShip() && ship->GetTargetShip()->GetGovernment()->IsPlayer());
 			int type = RadarType(*ship, step);
 			radar[calcTickTock].Add(isYourTarget ? Radar::SPECIAL : type, ship->Position(), size);


marphod:
For what it is worth, I believe the requiring vector::end() rather than vector::cend() issue is either from using a pre-c++11 version of vector, or is a compiler bug.  In the latter case, it might be worth filing a bug report against the compiler(mingw).

tehhowch:
It is not an issue with only MinGW, but g++ as well Interestingly, VS2017's clang/c2 toolchain does not complain about using `cend()` in this manner.

marphod:
Actually, on closer inspection, this looks like a compiler flag issue.  

The correct version of vector::insert() is in include/c++/bits/stl_vector.h (line 1060 of mingw 5.4.0); it is just not being used as the pre-processor directive `#if __cplusplus >= 201103L` is failing.

I really should set up my build environment and track this down, but at a guess, the std=c++11 flag is getting lost between the SConstruct file and the call to g++

tehhowch:
Could be version related.  The makefile I made for Windows x64 builds, using mingw32-make (v6.3.0) to call g++ with std=c++11 directly, compiled AI.cpp using `cend()` just fine. My scons build (gcc 4.8.4) however, does not. And at the same time, the code blocks build that @Bladewood uses (also MinGW) failed on cend().

marphod:
The gcc standards compliance doc (https://gcc.gnu.org/projects/cxx-status.html) has all c++11 features in by 4.8.1, so that shouldn't be it.  I'd need to check the headers to be sure.

Does scons and code::blocks invoke gcc or g++? if gcc, it might be worth changing to `g++` (or trying adding `-x c++` early in the flags).

tehhowch:
https://github.com/endless-sky/endless-sky/blob/master/SConstruct
That will tell you more than I can 👍

marphod:
That looks like the compiler is taken from the environment and not explicitly set.  Ew.

And code::blocks explicitly uses gcc.  Let me see if I can track down if this is a not-explicitly-stating-the-language item and open an issue against those files.

tehhowch:
Also, scons on my laptop (as opposed to the scons build above, which is via the Travis CI system), compiled the `.cend()` version just fine too:
```
 g++ -v
Using built-in specs.
COLLECT_GCC=g++
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-linux-gnu/5/lto-wrapper
Target: x86_64-linux-gnu
Configured with: ../src/configure -v --with-pkgversion='Ubuntu 5.4.0-6ubuntu1~16.04.4' --with-bugurl=file:///usr/share/doc/gcc-5/README.Bugs 
--enable-languages=c,ada,c++,java,go,d,fortran,objc,obj-c++ --prefix=/usr --program-suffix=-5 --enable-shared 
--enable-linker-build-id --libexecdir=/usr/lib --without-included-gettext --enable-threads=posix --libdir=/usr/lib 
--enable-nls --with-sysroot=/ --enable-clocale=gnu --enable-libstdcxx-debug --enable-libstdcxx-time=yes 
--with-default-libstdcxx-abi=new --enable-gnu-unique-object --disable-vtable-verify --enable-libmpx --enable-plugin 
--with-system-zlib --disable-browser-plugin --enable-java-awt=gtk --enable-gtk-cairo --with-java-home=/usr/lib/jvm/java-1.5.0-gcj-5-amd64/jre 
--enable-java-home --with-jvm-root-dir=/usr/lib/jvm/java-1.5.0-gcj-5-amd64 --with-jvm-jar-dir=/usr/lib/jvm-exports/java-1.5.0-gcj-5-amd64 
--with-arch-directory=amd64 --with-ecj-jar=/usr/share/java/eclipse-ecj.jar --enable-objc-gc --enable-multiarch 
--disable-werror --with-arch-32=i686 --with-abi=m64 --with-multilib-list=m32,m64,mx32 --enable-multilib --with-tune=generic 
--enable-checking=release --build=x86_64-linux-gnu --host=x86_64-linux-gnu --target=x86_64-linux-gnu
Thread model: posix
gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4) 
```

tehhowch:
From Bladewood:
[build-info.txt](https://github.com/endless-sky/endless-sky/files/1199275/build-info.txt)

Both this and the falied travis build are gcc 4.8.4... so there may have been a now-resolved issue.

PR #2896, opened 2017-08-06
Visual Studio 2017 Solution

kingmooselord:
Hello,

I've made some edits so that the project can be built purely with MSVC. In this case, I am using VS2017.

In general the edits were minor.

- included algorithms to resolve ::min and ::max
- included functional to resolve greater<double>
- renamed data to es_data in Files.cpp to resolve ambiguity. Probably a MSVC bug here. 
- version string needed c_str() for some reason.

I decided to #ifdef via _MSC_VER as Visual Studio defines this variable and I figured using _WIN32 could break existing stuff. 

For now this only builds endless-sky directly. Third party libs still need to be downloaded, compiled, etc.

### DIFF
--- a/readme-developer.txt
+++ b/readme-developer.txt
@@ -64,6 +64,20 @@ You will probably need to adjust the paths to your compiler binaries, and you sh
 You will also need libmingw32.a and libopengl32.a. Those should be included in the MinGW g++ install. If they are not in C:\Program Files\mingw64\x86_64-w64-mingw32\lib\ you will have to adjust the paths in the Code::Blocks file.
 
 
+Windows (Visual Studio):
+
+To build Endless Sky with Visual Studio (64bit) download this prebuilt dependency pack:
+
+<fix-me-url-to-dev64-vs.zip>
+
+Unpack the zip file in the same parent folder that contains the Endless Sky repo.
+
+Example:
+\dev\endless-sky\
+\dev\dev64-vs\
+
+Then open the Visual Studio solution located at endless-sky\vs\2017\endless-sky.sln. Compile the solution in either Debug or Release mode. You will then need to copy all of the .dll files in dev64-vs\bin\ to the output folder endless-sky\vs\2017\x64\Debug\ or endless-sky\vs\2017\x64\Release\.
+
 
 Mac OS X:
 
--- a/readme-developer.txt
+++ b/readme-developer.txt
@@ -79,7 +79,7 @@ Example:
 
 3) Open the Visual Studio solution located at 'endless-sky\vs\2017\endless-sky.sln'
 
-4) Compile the solution in either Release or Debug mode. A Release compile will result in machine optimations which will improve in-game performance.
+4) Compile the solution in either Release or Debug mode. A Release compile will result in machine optimizations which will improve in-game performance compared to the Debug compile.
 Successful compilation may require changing the Windows SDK version in the Project -> Properties submenu, to match the locally available SDK version. As an alternative to retargeting, newer Visual Studio installers can be used to obtain the missing Windows 8.1 SDK.
 
 5) Copy all of the .dll files from the 'dev64-vs\bin\' directory into the folder containing the freshly-made executable (default location is the main 'endless-sky' folder).
--- a/readme-developer.txt
+++ b/readme-developer.txt
@@ -66,17 +66,23 @@ You will also need libmingw32.a and libopengl32.a. Those should be included in t
 
 Windows (Visual Studio):
 
-To build Endless Sky with Visual Studio (64bit) download this prebuilt dependency pack:
+To build 64-bit Endless Sky using Microsoft's Visual Studio
 
+1) Download the 64-bit development libraries, prebuilt specifically for use with Visual Studio:
 <fix-me-url-to-dev64-vs.zip>
 
-Unpack the zip file in the same parent folder that contains the Endless Sky repo.
+2) Unpack the zip file's contents into the same parent folder that contains your download of the Endless Sky repository.
 
 Example:
 \dev\endless-sky\
 \dev\dev64-vs\
 
-Then open the Visual Studio solution located at endless-sky\vs\2017\endless-sky.sln. Compile the solution in either Debug or Release mode. You will then need to copy all of the .dll files in dev64-vs\bin\ to the output folder endless-sky\vs\2017\x64\Debug\ or endless-sky\vs\2017\x64\Release\.
+3) Open the Visual Studio solution located at 'endless-sky\vs\2017\endless-sky.sln'
+
+4) Compile the solution in either Release or Debug mode. A Release compile will result in machine optimations which will improve in-game performance.
+Successful compilation may require changing the Windows SDK version in the Project -> Properties submenu, to match the locally available SDK version. As an alternative to retargeting, newer Visual Studio installers can be used to obtain the missing Windows 8.1 SDK.
+
+5) Copy all of the .dll files from the 'dev64-vs\bin\' directory into the folder containing the freshly-made executable (default location is the main 'endless-sky' folder).
 
 
 Mac OS X:
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,7 @@
 /endless-sky
+/endless-sk*exe
+/endless-sk*pdb
+/*.dll
 /build/
 /.sconsign.dblite
 .DS_Store
@@ -9,4 +12,5 @@
 /obj/Release/
 /obj/Win32/
 /bin/
-
+vs/*/.vs/*
+vs/*/*.vcxproj.user
--- a/data/pirate jobs.txt
+++ b/data/pirate jobs.txt
@@ -1416,7 +1416,7 @@ mission "Bulk Slave Transport [0]"
 	name "Transfer slaves to <planet>"
 	job
 	repeat
-	description description "A pirate on <planet> has bought slaves from a local warlord. Transfer these <bunks> slaves and <cargo> to <destination> for <payment>. Slavery is punishable by death throughout the galaxy, so this may be a dangerous job to take."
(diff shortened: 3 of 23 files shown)


endless-sky:
The missing #includes were definitely a mistake, so I've added them in in commit e0ad03a. I also fixed the NaN thing, which was leftover code that was doing nothing. But some of that other stuff really looks like bugs in Visual Studio's compiler: cerr should be able to print a string, and I have no idea why a member variable would not be allowed to be called "data".

tehhowch:
I can confirm that the `data` name clash was an issue in VS2015 as well.
The compiler complains that it is an ambiguous reference, so perhaps it can
be solved by excluding the relevant header on the project level.

I'll try to do a checkout and build of this branch later today - does your
vsproject make assumptions about where the relevant external libraries are?

kingmooselord:
Ah. I figured out the version thing. #include <string> fixes it.

I think the data thing is probably a MSVC bug with the nameless namespace. I'll look into it some more and maybe file a bug with microsoft.

tehhowch:
Were you able to use the prebuilt resources (`dev64.zip` on the downloads page), or does MSVC throw a fit unless you build the libraries with MSVC as well?

kingmooselord:
I had to rebuild everything. The game would crash using the dev64.zip stuff, though it would link fine. 

Could be calling convention related, release/debug mismatch, or perhaps I just merely overlooked something.

tehhowch:
I would suggest then expanding the "readme-developer" file to indicate the steps needed to build using Visual Studio, with links to the official resources and/or links to a github repo that contains known-working versions of the source, with relevant VS projects to build them.
That way, it would be a two-step process to build via VS rather than a many-step, where most steps involve figuring out how to deal with dependencies. :+1: 
And the vsproject for ES can then assume a 'shared-parent-dir' relationship between ES and the library/binary dependencies

kingmooselord:
Yes, I could do that. The dependencies were definitely a pain to track down and compile. 

Though, we could also just offer a VS prebuilt package as well I'd imagine? Otherwise people will need to go additionally fetch cmake/nasm to build at least one of the dependencies. I believe it was libjpeg-turbo.

tehhowch:
There is already a prebuilt endless-sky binary for Windows, so I'm not sure what a VS prebuilt package would add, unless you are talking about prebuilding the dependencies (e.g. the `dev64.zip` resource) and shipping that as a standalone resource. In that case, I would definitely support that.

kingmooselord:
Yeah I meant prebuilt dependencies. ie dev64-VS.zip

kingmooselord:
Alright I've updated the project to use my new prebuilt stuff.

Just need to put endless-sky and the prebuilt folder in the same parent folder.

Uploaded a dev64-vs.zip here: https://drive.google.com/file/d/0B-EdEdbBhIaRX2J0c1AwR3VWb2s/view?usp=sharing

tehhowch:
(you have merge conflicts that need to be resolved)

endless-sky:
What's the error message you get for the "data" variable?

kingmooselord:
1>------ Build started: Project: endless-sky, Configuration: Debug x64 ------
1>Files.cpp
1>c:\users\ken\documents\dev\endless-sky\source\files.cpp(144): error C2872: 'data': ambiguous symbol
1>c:\users\ken\documents\dev\endless-sky\source\files.cpp(144): note: could be 'data'
1>c:\users\ken\documents\dev\endless-sky\source\files.cpp(42): note: or       'std::string `anonymous-namespace'::data'

warp-core:
Compiling it on visual studio 2015 works perfectly fine for me.
I did have to change, in the project properties, under configuration properties and then general, the platform toolset from v141 to v140 because it was complaining that it didn't have v141 installed.

tehhowch:
VS will warn about data loss when casually assigning a double to an int datatype, which is almost all of those warnings IIRC.

tehhowch:
Alright, had a chance to pull down your branch:
Steps taken:
 - Unzip dev64-vs.zip into the parent directory of the `endless-sky` folder
 - Checkout branch
 - Open the .sln file in vs/2017
 - Retarget the solution from Win 8.1 to Win 10 (10.0.15063.0), as I do not have the 8.1 SDK.
 - Build.

This is the dev64-vs : endless-sky folder relationship: 
image

The compile and link complete with minimal warnings (no errors). This is the warning list I get for the Debug x64 target:
image
I do obtain a fair number of "PDB not found" warnings when linking (see log).
Build log: 
[es_vs_Debug_x64.txt](https://github.com/endless-sky/endless-sky/files/1208166/es_vs_Debug_x64.txt)

The binary is created, though attempting to run the it results in a failure due to 'missing .dlls' (OpenAL32, SDL2, glew32): 
image

The same error set appears for the Release x64 version: 
[es_vs_Release_x64.txt](https://github.com/endless-sky/endless-sky/files/1208319/es_vs_Release_x64.txt)

Adding `..\..\..\dev64-vs\bin` to the VC++ Directories executable path & rebuilding all did not resolve the .dll errors. I'll look into this again later.

I would also suggest that this PR update the `.gitignore` file, output the executable into the ./endless-sky directory, and use the `build/` directory to store the compiled object files (rather than nesting further down into the `vs` folder.

warp-core:
Those errors upon running can be solved by copying the libraries from wherever they are into the same folder as the executable, as is necessary with the release versions of the game.
That list of warnings looks like what I was getting as well.

kingmooselord:
Yeah, you need to manually copy the binaries in dev64-vs/bin to your output folder. I suppose I could have a xcopy post build event that does it or something if we must automate it. I had somewhat assumed most people would know to do that since Visual Studio typically has insane behavior with DLL dependencies.

Changing the output location is fine. I wasn't sure where XCode and friends dumped all their stuff so I figured to just go with the VS defaults.

tehhowch:
Consider adding it to the updated readme-developer file - I'm sure I won't be the last one for whom Endless Sky is the first VS project that has DLL dependencies.

warp-core:
Am I missing something because I have to copy those libraries into the output folder when I compile in codebloocks as well.

tehhowch:
The developer readme that ships with the repo makes no mention of needing to also copy the DLLs to the application's directory in order for it to run on Windows. I've also only built ES on ubuntu before, where this particular insanity doesn't happen 😇

warp-core:
Glad we're all on the same page. Building stuff on Linux based systems is better than on windows based systems.
Maybe the need to copy the libraries into the output folder for builds done on Windows should be added to the readme.

kingmooselord:
Added a blurb explaining it to the developer readme file. I left the url as a FIXME for the dependency zip since we probably want something more official than my random google drive link ;)

tehhowch:
Do you have any thoughts on how we can appropriately handle multiple target platforms? As I mentioned, I do not have the 8.1 SDK and thus I had to modify the project file to build. As this is a tracked file, testing code on feature branches and making PRs will be rather annoying due to the perpetual need to stash the change and then reapply it when swapping branches.

I'd also suggest updating the .gitignore to ignore the user-specific files (and the build directory if it isn't moved to `endless-sky/build/*`)

warp-core:
Also, you might want to change the folder name to 2015 because that solution file contains the VS 2015 version numbers and so can be opened on VS 2015 without any complaints. Just to make that clearer to future users.

kingmooselord:
Eh, not sure about the SDK thing. 8.1 is needed to usually build anything useful since half the useful headers and examples are or used to be missing from the 10.X ones. Plus there's multiple 10.X SDK's which could or could not be installed depending on how/when VS was installed. I know 8.1 was needed to build those dependencies, though moot since we are providing them prebuilt. We could bump the target version but it's very likely everyone is going to have different versions installed at some point. 

We could just have everyone need to install 8.1 and call it a day. It's a checkbox in the Visual Studio 2017 installer under C++, so nothing groundbreaking to track down there. 

I feel like Visual Studio usually prompted the user if the SDK/Toolkit version doesn't match, but I usually have them all installed so I might be dreaming of that. Perhaps we just need to add a blurb about it. It's a simple right click -> retarget to fix.

I don't know about renaming it to 2015. It was created in 2017, it just happens to be the same as 2015's format. Most future/new people are probably going to have 2017 anyway since it's free.

I feel like both issues could just be mentioned in the developer readme?

tehhowch:
If you do not have the correct SDK, you are not instructed to change the build target until you attempt to build:
image

I've made the updates I mention (minus retargeting to the latest windows SDK) and opened a PR to this branch.

The 'fixme' link can probably be replaced with a link to a new github repository created to house all the prebuilt dependencies that are needed.

kingmooselord:
Merged your PR, thank you.

readme-developer.txt
```diff
@@ -64,6 +64,26 @@ You will probably need to adjust the paths to your compiler binaries, and you sh
...
+
+2) Unpack the zip file's contents into the same parent folder that contains your download of the Endless Sky repository.
+
+Example:
+\dev\endless-sky\
+\dev\dev64-vs\
+
+3) Open the Visual Studio solution located at 'endless-sky\vs\2017\endless-sky.sln'
+
+4) Compile the solution in either Release or Debug mode. A Release compile will result in machine optimations which will improve in-game performance.
```
tehhowch: optimations -> optimizations (oops 😇 )
Perhaps add ` compared to the Debug compile.` to the end of the sentence.

endless-sky:
It's already enough of a pain for me to maintain the XCode and CodeBlocks projects that I'm not sure I want to add yet another project type that needs updating every time a new class or library is added. I did go through and fix the `#include` errors, but the `data` one is entirely VS's fault.

KiLEdEnNis:
from the few weeks I've worked with VS2017 in college I know it has some really dumb bugs that make no sense, if those pop up everytime you make a VS2017 Solution you'd basically spend more time fixing those than on actual content
so doesn't even seem worth to do this

jafdy:
I've seen VS jump code around in ways that should never ever happen. It's not the best editor.

Issue #2910, opened 2017-08-10
Repaired "Derelict" ships do not recharge shield/hull/energy when landed


tehhowch:
Currently, if you repair a `derelict` personality ship and then land with it at a planet with a spaceport, it will not restore any shields, hull, or energy, but will get refueled.

This seems counterintuitive, as you have already repaired it (otherwise it would not have been able to land), and every other ship type in the game is repaired upon landing at a spaceport.

The [cause of this behavior is](https://github.com/endless-sky/endless-sky/blame/master/source/Ship.cpp#L1798):
```
	if(!personality.IsDerelict())
	{
		if(atSpaceport || attributes.Get("shield generation"))
			shields = attributes.Get("shields");
		if(atSpaceport || attributes.Get("hull repair rate"))
			hull = attributes.Get("hull");
		if(atSpaceport || attributes.Get("energy generation"))
			energy = attributes.Get("energy capacity");
	}
```
The fix is to change the first line to
```
if(!personality.IsDerelict() || atSpaceport)
```
This means that when Ship::Recharge is called for any ship that has landed at a friendly spaceport, it will result in a full health ship.

A check of the existing issues about derelict ships didn't turn up any indications that the current behavior is explicitly intended, but perhaps I've missed one.
The relevant derelict / Ship::Recharge commits are
2b5bc71 - introduce derelict personality
e9ee546 - landing on uninhabited planets restores attributes for which ships have regeneration

tehhowch:
Moving this here so it doesn't checklist the issue

<details><summary>Code Review</summary>

The Ship::Recharge method is declared with the default value of `true` for the parameter `atSpaceport`, and is called from these points in the code
 - [x] [Engine::Place, line 185](https://github.com/endless-sky/endless-sky/blob/2e813777c01e36af5f6e5d35caa2056d7193432d/source/Engine.cpp#L184)
Here the function is only called for ships which are not disabled (an unrepaired derelict ship is disabled and thus does not receive this call)
 - [x] [Engine::CalculateStep, line 1540](https://github.com/endless-sky/endless-sky/blob/2e813777c01e36af5f6e5d35caa2056d7193432d/source/Engine.cpp#L1540)
Here the function is called for a special `person` that has been selected to appear in the game. This call to Recharge precedes the call to SetPersonality, which is the method that would make a derelict ship actually become a derelict, disabled ship (i.e. not applicable).
[...]

PR #2912, opened 2017-08-10
Repaired derelicts spawn as repaired, and repair fully when landing at spaceports. 

tehhowch:
Refs #2910 

If you have not repaired a derelict, it cannot land -> no change from master
If you have repaired a derelict, it can land:
 - landing at a friendly spaceport -> "extra tools available" -> derelict ships can be fully repaired
 - landing on an uninhabited planet / planet with no spaceport -> no "extra tools available" -> no change from master
 - reopening the game after you have repaired the derelict ship will not require you to repair it again

### DIFF
--- a/source/NPC.cpp
+++ b/source/NPC.cpp
@@ -138,7 +138,7 @@ void NPC::Load(const DataNode &node)
 	{
 		ship->FinishLoading();
 		ship->SetGovernment(government);
-		ship->SetPersonality(personality);
+		ship->SetPersonality(personality, actions[ship.get()]);
 		ship->SetIsSpecial();
 	}
 }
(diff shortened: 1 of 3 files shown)


endless-sky:
I thought about this for a while before realizing that every time Ship::FinishLoading() is called, the game knows whether this is a newly created instance of a ship, or a previously created instance being loaded from a saved game.

If it's loaded from a saved game, its hull, shields, etc. are loaded from the file (so if it was disabled, it will stay disabled). If it's being newly created the game can just give it full health or make it disabled based on whether or not it's a derelict (as long as the personality is set first). So, it's sufficient to just move the IsDerelict() check into Ship::FinishLoading(). (ee9d339)

tehhowch:
I do like that implementation better. I wonder if there is any spillover to previous issues, as I recall some past discussions about spawning ships with specific crew required specifying the hull value... These folks know more about that:
@Elyssaen @EndrosG

PR #2913, opened 2017-08-10
NPC "accompany" requires failIf value

tehhowch:
This effectively makes `npc accompany` become `npc save accompany`.

If it was the intent of `accompany` to allow ships to be destroyed and still be able to be completed, then this is currently not working as intended (and a different fix is required).

By adding a failIf value, the player is immediately shown the "Mission failed" message and the mission will be removed upon landing when the mission becomes impossible to complete.

### DIFF
--- a/source/NPC.cpp
+++ b/source/NPC.cpp
@@ -56,7 +56,10 @@ void NPC::Load(const DataNode &node)
 		else if(node.Token(i) == "evade")
 			mustEvade = true;
 		else if(node.Token(i) == "accompany")
+		{
 			mustAccompany = true;
+			failIf |= ShipEvent::DESTROY;
+		}
 	}
 	
 	for(const DataNode &child : node)


endless-sky:
Currently in the game data, every "accompany" npc also has the "save" requirement, so this would make no difference for them.

I'm not sure if it makes more sense to say that "accompany" implies "save," or to keep them separate so it's clear(er) when you're reading a mission definition that the ship has to be saved.

tehhowch:
It is probably better to continue the practice of explicitly specifying both `save` and `accompany`, but at the same time this is only relevant to those who are browsing the game's datafiles - the in-game mission text is where the success criteria need to be effectively conveyed to the player. Ensuring that the failIf value gets set even if `save` is omitted in the mission specification is more or less housekeeping to guarantee that the player is notified of failure, and not given `on visit` messages when the player can't even complete the mission anymore.
(I accept that this is not going to be an issue for the base game missions, and is more or less an enhancement to make sure plugins play a little better if their author is not quite fully versed on the code internals)

marphod:
> I accept that this is not going to be an issue for the base game missions, and is more or less an enhancement to make sure plugins play a little better if their author is not quite fully versed on the code internals

Well, it isn't for the FW campaign.  I could see a mission in another campaign, where keeping the accompanying ship alive is a secondary goal, but not required for mission success.

tehhowch:
@marphod such a mission cannot use accompany - accompany requires every ship marked accompany to be in system and not dead or disabled when you land

marphod:
In retrospect, you'd probably do a base mission, and an invisible accompany mission as the kicker.

Asaccompany requires the ship in system and not disabled, save is inherently redundant. Standardizing on both being in the file seems unnecessarily verbose.  Kind of like using `signed long int` instead of `long`.

PR #2933, opened 2017-08-14
Removed compiler warning.

Hacklin:
Remove the only warning given by GCC v7.1.

### DIFF
--- a/source/Politics.cpp
+++ b/source/Politics.cpp
@@ -100,7 +100,7 @@ void Politics::Offend(const Government *gov, int eventType, int count)
 				provoked.insert(other);
 			}
 		}
-		else if(abs(weight) >= .05 && (count * weight) != 0.)
+		else if(count && abs(weight) >= .05)
 		{
 			// Weights less than 5% should never cause permanent reputation
 			// changes. This is to allow two governments to be hostile or
--- a/source/Politics.cpp
+++ b/source/Politics.cpp
@@ -100,7 +100,7 @@ void Politics::Offend(const Government *gov, int eventType, int count)
 				provoked.insert(other);
 			}
 		}
-		else if(abs(weight) >= .05 && count * weight)
+		else if(abs(weight) >= .05 && (count * weight) != 0.)
 		{
 			// Weights less than 5% should never cause permanent reputation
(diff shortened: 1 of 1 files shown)


source/Politics.cpp
```diff
@@ -100,7 +100,7 @@ void Politics::Offend(const Government *gov, int eventType, int count)
 				provoked.insert(other);
 			}
 		}
-		else if(abs(weight) >= .05 && count * weight)
+		else if(abs(weight) >= .05 && (count * weight) != 0.)
```
endless-sky: I think it could actually just be `... && count)`, because the first check already guarantees that `weight` is nonzero.

tehhowch:
Are you going to update this PR based on the feedback received?
I have additional feedback: perhaps the `count` check should come first (before the float operation and compare).

Hacklin:
Sorry.
Done.

Hacklin:
The 83e0483 ref (83e04833484e1279cafecba8eeeebcc17b249ebf) was erroneous.
It didn't do anything.

Cloned my fork locally.

So did a reset locally to previous ref to get rid of 83e0483 ref.
Did a git push to my fork.
Got errors (rejected, non-fast-forward). [???]

So did did a push to my fork to remove the 83e0483 ref.
Got the error that this ref did not exist.

Checked and double checked: the 83e0483 ref is gone.
No idea why I got the [???] errors.

tehhowch:
Non fast forward means you have diverged histories (we said do this here, they said do that here) which generally require editing the changed files to keep the desired changes. When pushing to a fork it means you are changing the history of a public document, which (if you have collaborators that base their work off of yours) is just about the most heinous git sin you can commit. For a single-author feature branch, it's common practice.

```
git checkout fix-compilation-warning-1
git reset --hard 8748370e
```
Add a new commit with your incorporation of the feedback received.
```
git push -f <remote>
```
(Replace \<remote\> with whatever your GitHub fork's alias is - for me, my fork is `origin` and this repo is `upstream`)
This should result in a PR with 2 commits, both yours.

Hacklin:
Thanx.
Still pretty new to git.

Issue #2936, opened 2017-08-15
Flagship cannot plan wormhole travel without hyper/jump drive using mouse clicks


tehhowch:
Per the title. If the player's flagship has no Hyperdrive or Jump Drive, the mouse cannot be used to plan travel over otherwise valid wormhole routes.
Ref: https://steamcommunity.com/app/404410/discussions/1/1471967615874008822/#c1471967615881475355

Detail: 
MapPanel initializes DistanceMap with the [current player](https://github.com/endless-sky/endless-sky/blob/v0.9.8/source/MapPanel.cpp#L58), but due to the [lack of a hyper/jump drive, the initialization of routes is aborted](https://github.com/endless-sky/endless-sky/blob/v0.9.8/source/DistanceMap.cpp#L157), and the "You have not yet mapped a route to this system" [message is displayed](https://github.com/endless-sky/endless-sky/blob/v0.9.8/source/MapPanel.cpp#L102) on top of the map, even if the player has visited the system / used the wormhole.
The system can still be [selected via the J key](https://github.com/endless-sky/endless-sky/blob/v0.9.8/source/MapDetailPanel.cpp#L162), which shortcuts [the check in Select](https://github.com/endless-sky/endless-sky/blob/v0.9.8/source/MapPanel.cpp#L446) (but of course the "You have not yet mapped a route to this system" message is still displayed).

This is currently only a mod issue, as there are no systems in the base game with an Outfitter and also a wormhole.

tehhowch:
I think the simplest fix for this is changing DistanceMap::Init line 157 from
```
if(!hyperspaceFuel && !jumpFuel)
```
to
```
if(!hyperspaceFuel && !jumpFuel && !player)
```
This shouldn't affect any of the AI escort pathfinding while allowing the player to design wormhole-only travel plans via both click / J key and not receive the "You have not..." message.

tehhowch:
This is fixed by #2938, which checks the local system for a wormhole when initializing DistanceMap, and only exits early if there is no wormhole and no jump method.
(will close with milestone)

PR #2938, opened 2017-08-15
Update allowed travel plan creation conditions

tehhowch:
Refs #2936 

 - Removes incorrect/vague MapPanel error message when the player's flagship has no hyperdrive/jump drive installed.
 - Wormhole-only travel plan creation is possible for the player when no hyperdrive/jump drive is installed.

If a player has no hyperdrive or jump drive installed, this PR will enable them to make appropriate & allowed travel plans that use only wormholes.
The error message shown when attempting to route to an unrouteable system now properly indicates if you know how to get to that system but do not have the correct equipment which makes said travel possible, or if you actually do not know how to get to that system.

### DIFF
--- a/source/DistanceMap.cpp
+++ b/source/DistanceMap.cpp
@@ -153,9 +153,21 @@ void DistanceMap::Init(const System *center, const Ship *ship)
 		if(hyperspaceFuel == jumpFuel)
 			hyperspaceFuel = 0.;
 		
-		// If this ship has no mode of hyperspace travel, bail out.
-		if(!hyperspaceFuel && !jumpFuel && !player)
-			return;
+		// If this ship has no mode of hyperspace travel, and no local
+		// wormhole to use, bail out.
+		if(!jumpFuel && !hyperspaceFuel)
+		{
+			bool hasWormhole = false;
+			for(const StellarObject &object : ship->GetSystem()->Objects())
+				if(object.GetPlanet() && object.GetPlanet()->IsWormhole())
+				{
+					hasWormhole = true;
+					break;
+				}
+			
+			if(!hasWormhole)
+				return;
+		}
 	}
 	
 	// Find the route with lowest fuel use. If multiple routes use the same fuel,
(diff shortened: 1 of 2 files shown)


source/MapPanel.cpp
```diff
@@ -102,14 +102,18 @@ void MapPanel::Draw()
...
-		
-		static const string NO_ROUTE = "You have not yet mapped a route to this system.";
 		Color black(0., 1.);
 		Color red(1., 0., 0., 1.);
-		Point point(-font.Width(NO_ROUTE) / 2, Screen::Top() + 40);
 		
-		font.Draw(NO_ROUTE, point + Point(1, 1), black);
-		font.Draw(NO_ROUTE, point, red);
+		string errormsg = "You have no";
+		if(player.HasSeen(selectedSystem))
```
endless-sky: HasSeen() returns true any time you know the system's name (e.g. because it's a mission destination), so this will lead to confusing messages in cases where the player has not actually mapped a route yet.

Also, splitting the string in the middle of a word is rather confusing, and `errormsg` doesn't follow the coding conventions for variable names (camelCase and avoiding abbreviations).

tehhowch: I had wanted to use `HasSeen` because then the player would get told about having no available route to a system that they had mapped (e.g. visited a neighbor at some point and thus knew a route to it), but hadn't fully considered the implication from the mission standpoint.

Rather than make a custom function just so the error message can be a touch more accurate, I switched to `HasVisited`. This way, only a system the player has actually routed to (and for some reason, can no longer route to anymore) will trigger the "You have no available route" message.

Fixed the style clash as well.

source/DistanceMap.cpp
```diff
@@ -154,7 +154,7 @@ void DistanceMap::Init(const System *center, const Ship *ship)
 			hyperspaceFuel = 0.;
 		
 		// If this ship has no mode of hyperspace travel, bail out.
```
marphod: Comment no longer accurate.

tehhowch: Good point. I almost wonder if the better fix is to (instead of `!player`) verify that there are no wormholes in the relevant ship's system, as wormholes (while not explicitly 'hyperspace' travel, are an intersystem transportation mechanism.

tehhowch:
Rather than the clunky "if it's the player make an exception" check, the more universal check for a wormhole in the calling ship's system is used as the final kicker in exiting the DistanceMap initialization early.

PR #2949, opened 2017-08-21
Honor environement

AMDmi3:
Pass environment through SCons. For instance, this is required to pass CCACHE_* variabled to ccache.


tehhowch:
I notice below the env initialization, specific flags are addressed. I think that may be the preferred method rather than importing every value in system env.

How many flags does ccache set?

AMDmi3:
There's no way to list all possible individual flags, any of CCACHE_* mentioned in the ccache manpage may be set, and more flags may be added in future. There's also distcc and a lot of other tools which may need environment. So the environemnt cleaning SCons does is broken, and the only clean way to fix it is to always pass whole env.

tehhowch:
Rather than whole environment import, wouldn't doing something similar to this work?
https://github.com/endless-sky/endless-sky/blob/785bb501f353e13498b146a19341394fdb7be10d/SConstruct#L61

AMDmi3:
I've already explained - you don't and cannot ever know the full list of variables to pass. Whole idea of tampering with environment is broken and should be avoided.

PR #2955, opened 2017-08-24
Keep partially collected flotsam

flaviojs:
The leftover units would vanish when only part of the flotsam fits the
cargo hold. This means that you can lose cargo just by transfering it
to another ship of your fleet (jettison cargo, then guide other ships to
the flotsam).

This commit changes that behaviour, only removing the flotsam when there
are no more units. Only 1 ship can collect from the flotsam on each step.

### DIFF
--- a/source/Flotsam.cpp
+++ b/source/Flotsam.cpp
@@ -154,7 +154,7 @@ int Flotsam::TransferTo(Ship *collector)
 		collector->Cargo().Add(commodity, count);
 	
 	Point relative = collector->Velocity() - velocity;
-	double proportion = (double)amount / count;
+	double proportion = static_cast<double>(amount) / count;
 	velocity += relative * proportion;
 	
 	count -= amount;
--- a/source/Flotsam.cpp
+++ b/source/Flotsam.cpp
@@ -142,3 +142,14 @@ double Flotsam::UnitSize() const
 {
 	return outfit ? outfit->Get("mass") : 1;
 }
+
+
+
+// Remove "units" from the flotsam.
+void Flotsam::Remove(int amount)
+{
+	if(amount < this->count)
+		this->count -= amount;
+	else
+		this->count = 0;
+}
--- a/source/Flotsam.cpp
+++ b/source/Flotsam.cpp
@@ -145,11 +145,18 @@ double Flotsam::UnitSize() const
 
 
 
-// Remove "units" from the flotsam.
-void Flotsam::Remove(int amount)
+// Transfer contents to the collector ship. The flotsam velocity is
+// stabilized in proportion to the amount being transferred.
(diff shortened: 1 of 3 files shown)


tehhowch:
What you have so far looks good. As I read it, though, the flotsam will continue on the same trajectory as before, despite having been collected, stripped, and then thrown back out.

Visually, you can polish this a lot by re-placing the flotsam to come from the collecting ship. I'd probably use the (body, point) version of Place, so that you have better control of the emitted angle and velocity, i.e. redirect it by 15 - 90d and reduce its velocity by 5-50%.

flaviojs:
Since conceptually the flotsam is first being brought into the ship, I preferred to keep consistency and place it the same way as a jettison.

If conceptually I was only bringing in what fit into the cargo hold, then I wouldn't mind adjusting the angle and velocity.

tehhowch:
You should also adjust the behavior of `harvests` NPCs - right now they only pursue flotsam that can fit in their available space. I'm also curious if you are able to maintain the initial lifetime of the flotsam, or if it gets "renewed" each time it gets shuffled from ship to ship.

flaviojs:
The lifetime of the flotsam is unchanged. I consider the flotsam lifetime to be the amount of time the contents can handle space radiation and even considered shortening it from the stripping, but refrained from doing so because the lifetime is already short.

The ```harvests``` behavior targets any flotsam that has **a unit** that can fit the available space, it will chase flotsam that has other units that don't fit so there is nothing to change.

endless-sky:
Visually, it looks kind of weird to me for the flotsam to "teleport" to the center of your ship and change its movement, in a single frame, when you pick up part of it. Even though it's less realistic, the behavior of the original commit (just leaving the flotsam where it is) looks better.

flaviojs:
I tried delaying the placement but it didn't look right. I also tried what @tehhowch suggested and it was a bit better but still didn't look right.

Finally, I changed the concept to stripping the flotsam in space (only bringing in what can fit) and came up with the velocity being stabilized in proportion to the stripped amount.
I like how it looks, you can clearly see that something happened to the flotsam and you get a visual hint of who collected and how much.

source/Flotsam.cpp
```diff
@@ -145,11 +145,18 @@ double Flotsam::UnitSize() const
...
-	if(amount < this->count)
-		this->count -= amount;
-	else
-		this->count = 0;
+	int amount = outfit ?
+		collector->Cargo().Add(outfit, count) :
+		collector->Cargo().Add(commodity, count);
+	
+	Point relative = collector->Velocity() - velocity;
+	double proportion = (double)amount / count;
```
tehhowch: I think `static_cast<double>` is preferred over `(double)` casting.

flaviojs: Done.

PR #2956, opened 2017-08-24
Merge "add fleet" and "add variant" declarations with existing fleet/variant definition

tehhowch:
1) This prevents the need for multiple "remove fleet" or "remove variant" uses in a game event, as there will  be at most one of any specific fleet definition in System::Fleets(), or at most one of any specific variant in Fleet::Variants().
(There are currently no duplicated fleet variant specifications, but to preserve conceptual conformity - "remove ___" means the ____ is gone from its container - the Fleet.cpp change is required)

2) For System fleet definitions, the "set period" keyword is defined. It is used exactly as "add" or "remove" would be, e.g. `"set period" fleet "Small Southern Merchants" 1000`. If the `Small Southern Merchants` fleet exists, its period will become 1000, no matter if it was previously 100 or 20000. If the `Small Southern Merchants` fleet does not exist in the system, it will be inserted with period 1000.
This means "set period" can effectively replace the use of "add" as a keyword for changes to a system's fleet definitions, and the keyword "remove" is needed only for situations in which a specific fleet is completely removed from the system's fleets, rather than paired with "add" to create situations in which the fleet's presence is to be reduced.

Combined, these two changes minimize the number of lines needed to implement incremental changes to system fleets and fleet variants.
Example: 
Applying add/remove to the FW campaign, largely in full, is +1353 / -1233: https://github.com/tehhowch/endless-sky/pull/21
Applying add/remove with the features in this PR, and also making every FW event incremental (except "pug flee" and the final Deneb specification), is +1359 / -1339: https://github.com/tehhowch/endless-sky/pull/23


tehhowch:
Per https://github.com/endless-sky/endless-sky/pull/2959#issuecomment-326304063, replaced `"set period"` with `set`

endless-sky:
To me it makes a lot more sense to just say that every "add" must be balanced with a "remove," rather than one "remove" eliminating all fleets of a given type.

(Note that the current behavior might actually be useful in some situations - for example, you could create a system with five copies of a pirate fleet, and every time you complete some mission one of them disappears, slowly reducing the rate of piracy.)

tehhowch:
I hadn't considered that kind of campaign, and it is easier to implement with the current behavior (though still achievable with the behavior from this PR, just requiring the writer to make 1 event for each stage, since the period has to be set explicitly).

I'll finish up with tehhowch#21 then, to have parity with tehhowch#23, and then open it here for comment.

PR #2962, opened 2017-08-25
Fix missing jettisoned commodities

flaviojs:
The last box of commodities was not being jettisoned unless it was a multiple of 5, which is the maximum number of commodities per box.

Note that this will also allow players to send out more flotsam to distract pirates (select commodity, dump a quantity lower than 5, repeat).

### DIFF
--- a/source/Flotsam.cpp
+++ b/source/Flotsam.cpp
@@ -20,7 +20,13 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Ship.h"
 #include "SpriteSet.h"
 
+#include <cmath>
+
 using namespace std;
+
+
+
+const int Flotsam::TONS_PER_BOX = 5;
 
 
 
@@ -29,6 +35,9 @@ Flotsam::Flotsam(const string &commodity, int count)
 	: commodity(commodity), count(count)
 {
 	lifetime = Random::Int(300) + 360;
+	// Scale lifetime in proportion to the expected amount per box.
+	if(count != TONS_PER_BOX)
+		lifetime = sqrt(count / TONS_PER_BOX) * lifetime;
 }
 
 
--- a/source/Flotsam.h
+++ b/source/Flotsam.h
@@ -63,6 +63,11 @@ public:
 	double UnitSize() const;
 	
 	
+public:
+	// Amount of tons that is expected per box.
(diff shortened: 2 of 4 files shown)


endless-sky:
The current behavior is intentional, for exactly that reason - it makes it so you can't distract an entire pirate fleet with 100 credits worth of food.

flaviojs:
Hmm, what if I also reduce the flotsam lifetime in proportion to how incomplete it is?

tehhowch:
Add a check to the player's method of jettisoning that fails for quantities less than perbox, but allow AI NPCs to jettison what they have.

Or add a comment in the code that identifies this as an intentionally reduced iteration space.

flaviojs:
The main reason why I want to keep partial flotsam of the player is so I can pass it along to my fleet by jettisoning it over the ship that will carry it.

As for how to scale lifetime so partial flotsam does not distract pirates better than full flotsam, I liked the square root lifetime: `lifetime * sqrt( count / 5 )`
This is only for commodities since outfits are already being scaled by value.

| Tons | Raw Lifetime | Linear Lifetime | Square Root Lifetime |
| --- | --- | --- | --- |
| 1 | 300 to 660 | 60 to 132 | 134 to 295 |
| 2 | 300 to 660 | 120 to 264 | 189 to 417 |
| 3 | 300 to 660 | 180 to 396 | 232 to 511 |
| 4 | 300 to 660 | 240 to 528 | 268 to 590 |
| _5_ | _300 to 660_ | _300 to 660_ | _300 to 660_ |
| 10 | 300 to 660 | 600 to 1320 | 424 to 933 |
| 15 | 300 to 660 | 900 to 1980 | 519 to 1143 |
| 20 | 300 to 660 | 1200 to 2640 | 600 to 1320 |

flaviojs:
Tried and liked the square root lifetime.

Miners didn't seem to be affected by the decreased lifetime.

When distracting pirates with 15 tons of cargo in Algorel with a fresh pilot, both 1 ton and 5 ton flotsams were effective distractions.
Usually the 1 ton flotsams were only good enough to make pirates change to an already nearby target.
Usually the 5 ton flotsams were good enough to make pirates take damage from allies and change target.
I was able to run away more often by using 5 ton flotsams.

@endless-sky
It felt balanced.

Issue #2964, opened 2017-08-25
Should Remnant space's raid fleet be the Korath?


tehhowch:
Lore says that Korath occasionally raid their system. This is set via the fleet definitions for their systems. 

However, the Remnant government has no raid fleet defined, so a player is free to run a trade-only fleet amongst the (currently added) Remnant worlds. Based on the worlds available now, setting the Remnant government's `raid fleet` to be `Korath Raid` (Raider + 2 Chasers) is pretty sensible, but if the Remnant worlds get expanded significantly, this may no longer make sense as the new worlds could be pretty far from Korath-accessible space.

I'd like to hear feedback on this proposed change

Bladewood:
Problem with your pull request is that these Korath have jump drives.
Which makes farming them *far* easier.

tehhowch:
Indeed, the Remnant fleets would engage the raiding Korath ships alongside your own. To use the existing Raider (Hyperdrive) variant would need a new fleet as well (since you can't define a fleet directly in the raid fleet specifier).

PR #2965, opened 2017-08-25
Set Remnant space's "raid fleet"

tehhowch:
#2964 : This PR sets the Remnant's raid fleet (for having too much unguarded cargo space) to be the Korath Raider + 2x Chaser fleet.


Bladewood:
Problem is, these Korath have jump drives.
Which makes farming them *far* easier.

PR #2966, opened 2017-08-25
Enhancement: Allow government color in interface strings

flaviojs:
I'm currently playing with a Mule and a permanent fleet of 7 Headhunters (mid-game). I'm no longer interested in some of the small pirate ships, so when one of those ships appears as my target for boarding (after pressing B) I tell my fleet to focus fire on it to destroy it. Unfortunately sometimes I end up killing disabled allies by mistake.

When a ship is disabled the only way to know if it's an ally is by reading the government text, which forces me to take focus away from ongoing battles. To be able to **know if the disabled target is an ally** at a glance, this pull request is coloring the **government name** with the **government color**.

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -354,7 +354,7 @@ interface "targets" top left
 		color "medium"
 	string "target government"
 		center 75 435
-		color "medium"
+		color "medium" government
 	string "mission target"
 		center 75 455
 		color "medium"
--- a/source/Interface.cpp
+++ b/source/Interface.cpp
@@ -16,6 +16,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Font.h"
 #include "FontSet.h"
 #include "GameData.h"
+#include "Government.h"
 #include "Information.h"
 #include "LineShader.h"
 #include "OutlineShader.h"
@@ -487,11 +488,23 @@ bool Interface::TextElement::ParseLine(const DataNode &node)
 	if(node.Token(0) == "size" && node.Size() >= 2)
 		fontSize = node.Value(1);
 	else if(node.Token(0) == "color" && node.Size() >= 2)
+	{
 		color[Element::ACTIVE] = GameData::Colors().Get(node.Token(1));
+		if(node.Size() >= 3 && node.Token(2) == "government")
+			useGovernmentColor[Element::ACTIVE] = true;
+	}
 	else if(node.Token(0) == "inactive" && node.Size() >= 2)
+	{
 		color[Element::INACTIVE] = GameData::Colors().Get(node.Token(1));
+		if(node.Size() >= 3 && node.Token(2) == "government")
+			useGovernmentColor[Element::INACTIVE] = true;
+	}
 	else if(node.Token(0) == "hover" && node.Size() >= 2)
+	{
 		color[Element::HOVER] = GameData::Colors().Get(node.Token(1));
+		if(node.Size() >= 3 && node.Token(2) == "government")
+			useGovernmentColor[Element::HOVER] = true;
+	}
 	else
 		return false;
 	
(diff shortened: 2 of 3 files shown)


KiLEdEnNis:
in case you didn't know
use R to only target hostile enemies, B never lets you board friendlies unless there are no disabled hostiles (if I remember)
so either ya press R, then B, or you just press B, if you want to board an ally press Shift+B

tehhowch:
I'm really not sure about the way you've done that. In general, the UI and text elements are monochrome. you could perhaps blink the ship outline more slowly than the "temporary disabled due to overheating".
You could also change the target text in general - allies get the allied blue tint, and hostiles get the hostile tint, even if they aren't targeting you. Then you don't even need to decipher if the government color you're seeing is hostile or allied, you simply have your default "blue = no fire", orange = fire" reaction.

flaviojs:
@KiLEdEnNis

I'm usually boarding mid-battle while my fleet is kicking ass and the local "police" (navy or militia or syndicate) is chasing targets (instead of actively destroying leftover pirates). This means R will target an active enemy, not a disabled enemy, and using R then B will be the same as just using B.

@tehhowch

Indeed I noticed most of the interface is monochrome, that's why I left alone the other places that show the government name.

Coloring this particular text gives useful information in other live situations. For example, it allows me to quickly cycle through all ships (with N) and see at a glance if there is any local "police" present. This would probably be useful if I try to play as a pirate or if I become the enemy of a government and want to see if there are enough pirates to distract them for a while.

Bladewood:
What you should do is press [R] to check if there are any hostile ships in the system. If nothing appears, you can safely [B]oard your friend. If there are hostiles, kill them off, then [B]oard your friend.

KiLEdEnNis:
if ya board enemies mid-battle just plainly pressing B should give no trouble, I do that all the time

tehhowch:
Op uses b to target disabled ship models he doesn't want to cap and tells
his fleet to fire at them with f.
He is coloring the text in the target ship display to help the visual
recognition of friend vs foe so he does not provoke allies mistakenly.

(Pressing f again will toggle the attack command btw)

KiLEdEnNis:
maybe this should be a setting then; not everyone enjoys a colored government text that much

tehhowch:
That is why I suggested the simpler friend vs foe coloring for the text,
the same colors that are used for the ship's outline when it is not
disabled.

Perhaps a different solution, one that keeps the monochrome ui, is to keep
the outline color even when disabled, but simply blend it with the inactive
color.  it is still visually evident that the ship has been disabled, but
so is the friend and foe status.

flaviojs:
In the Almach system I experienced enemies of different factions (Pirates and Korath) fighting each other and found this coloring to be useful, the only thing seriously missing was a way to target the previous ship after pressing N too many times (which I'll make a separate Pull Request for).

Having different factions with different colors in the minimap would be even more useful, but then I would lose the enemy vs ally visual information, so that's not desirable.

tehhowch:
image image image
image image image 

Because governments can be blue or blue tinted - a color which generally means "good" or at least "not harmful", and also hostile, or orange/red and not hostile, I am wary of approving this as-is.

flaviojs:
After using it for some time, I found it much more intuitive than I thought it would be. Now I know exactly who I'm offending when I try to capture/plunder and haven't made a single mistake (even when rushing). I don't even need to take my eyes off the ship outline (while repeating N) to know the government.

As far as I can see, the enemy/ally/disabled color scheme is only used in the minimap and the target outline, so I don't think it's confusing at all.

endless-sky:
I'm really against using colored text in the HUD. To me it just looks tacky, and some colors (like the Pirate and Syndicate ones above) end up hard to read..

That said, a much cooler alternative might be to have an icon for each government. Or, to have little icons that are colored using the government swizzle rather than the government color:

fw merch pir rep synd

(I'm not in love with those particular icons; I just wanted to sketch out how it might look.)

jafdy:
It would be interesting to create a symbol for the factions. I don't know who would do that though.

KiLEdEnNis:
I really like those color markers though (if edited a bit of course, but the idea is great)

but not teasing with a new ship outline? feelsbad

PR #2902, opened 2017-08-06
Syndicate Target Practice fixes

Pointedstick:
Made the syndicate target ships into self-destructing automata, imposed a penalty for failure, and added some appropriate hostile and disabled hails, thereby resolving the following issues:
- https://github.com/endless-sky/endless-sky/issues/2151
- https://github.com/endless-sky/endless-sky/issues/2268
- https://github.com/endless-sky/endless-sky/issues/2890

I also wanted add a fine on failure, but that currently doesn't show up right in the description. I have opened issue https://github.com/endless-sky/endless-sky/issues/2900 to track that.

PR #2969, opened 2017-08-26
Always display a mission's <payment> substitution as a positive number

tehhowch:
Otherwise "you are fined \<payment\>" becomes "you are fined -200,000".

Refs #2902 
image

### DIFF
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -25,6 +25,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Ship.h"
 #include "UI.h"
 
+#include <cstdlib>
 #include <vector>
 
 using namespace std;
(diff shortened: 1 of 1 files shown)


puk3:
any thoughts regarding always displaying negatives in red?

It is a basic accounting practice, and also tells you at a glance if you are looking at a debit or credit, while still addressing the linguistic/syntactical problem that this pull is fixing.

mootootwo:
any thoughts regarding always displaying negatives in red?

It is a basic accounting practice, and also tells you at a glance if you are looking at a debit or credit, while still addressing the linguistic/syntactical problem that this pull is fixing.

KiLEdEnNis:
if it says you are fined, any player that reads the message will know they lose that money, so it's not really needed / worth the effort of adding colored text support

MCOfficer:
as long as it's just in a list yes. in a full sentence, like Dennis said, it's kinda unnecessary and probably distracting.

PR #2973, opened 2017-08-28
Enhancement: Select previous ship with Ctrl

flaviojs:
This commit extends the "Select next ship" command, N key by default, to get the previous ship when Ctrl is being held.

With N you get the next targetable ship.
With Ctrl+N you get previous targetable ship.
With Shift+N you get the next player ship.
With Ctrl+Shift+N you get the previous player ship.

This allows a player to quickly repeat N until the intended target is selected and go back if N was pressed to many times by mistake.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -116,6 +116,7 @@ void AI::IssueMoveTarget(const PlayerInfo &player, const Point &target, const Sy
 void AI::UpdateKeys(PlayerInfo &player, Command &clickCommands, bool isActive)
 {
 	shift = (SDL_GetModState() & KMOD_SHIFT);
+	ctrl = (SDL_GetModState() & KMOD_CTRL);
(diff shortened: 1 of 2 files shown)


Pointedstick:
How about shift instead of ctrl? Shift is a more standard UI convention for inverting the direction of something discrete selected via the keyboard (example: in web browsers, ctrl-tab = select next tab and shift-ctrl-tab = select previous tab).

tehhowch:
Shift n is already bound as next player ship

Hacklin:
Well, you could bind Control-n as the next player ship.

endless-sky:
I think it would make more sense to re-think all the ship selection keys, rather than just adding new ones piecemeal. (Basically, trying to figure out what the most common tasks are that players want to perform, and making sure the key bindings for those actions are the simplest.)

PR #2975, opened 2017-08-28
Change to free worlds middle.txt to allow completion of electron beam mission with turret.

endlesssteve:
The mission FW Southern Recon 1B ( turret) is not offered and so the electron beam mission does not finish if you return with an electron turret. It seems for some reason the two mission will not be offered in parallel. I have changed the turret mission to be offered at Longjump if FW southern diplomacy 1D is done with completion at Trinket. I have tested this with several runs  from before FW Southern Diplomacy 1D is done. Longjump must be visited as part of the storyline leading to the electron beam mission so the (turret) version should not be missed.   issue 2961


Amazinite:
You might want to double check this. 121 files changed and added a duplicate of fw middle.

endlesssteve:
Too much to expect this would work first time. Sorry but this learning curve is rather steep!
Now I know why it's called Git.

endlesssteve:
The extra mission file is my fault for trying git add when it wasn't needed. It didn't seem to do anything but obviously did.
the 121 files are all some rubbish init message which came with a suggestion for a working push command from a forum.
Is there a way of repairing the push request or do I kill it and try again?

endlesssteve:
Replaced by 2977.

PR #2982, opened 2017-08-30
Reverse Thrust while turning.

Hacklin:
Key Press   Command
[Up]        Engage Forward Thrust (and disable Reverse Thrust).
[Left]      Turn Left.
[Right]     Turn Right.
[Down]      If Reverse Thrust installed and not engaging Forward Thrust,
            engage Reverse Thrust,
            else if not turning Left nor Right,
            turn to the reverse direction the ship is moving to.

Behavior Changes:
- Can no longer engage both Forward and Reverse Thrust at the same time.
- Turning Left or Right, will no longer disable Reverse Thrust.
- Can Reverse Thrust while turning Left or Right.
- Engaging Forward Thrust will always behave as if Reverse Thrust is not installed.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2774,18 +2774,18 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 		AutoFire(ship, command, false);
 	if(keyHeld)
 	{
+		if(keyHeld.Has(Command::FORWARD))
+			command |= Command::FORWARD;
 		if(keyHeld.Has(Command::RIGHT | Command::LEFT))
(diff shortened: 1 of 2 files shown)


Hacklin:
Note:
Either the Forward Thrust's Energy and Heat cost is applied
or the Reverse Thrust's  Energy and Heat cost,
but never both (see Ship.cpp line 1100).
Even if you currently engage both Forward and Reverse Thrusters.

So I suggest to cherry-pick the "Not using Forward and Reverse Thrust together." patch.

PR #2983, opened 2017-08-30
Fix memory leak.

Hacklin:


Hacklin:
You're correct.
And those malformed png images are not found in ES.
But then, no program survives contact with the user system ;)

endless-sky:
It's pretty much universally accepted that goto statements are a bad idea and should never be used. The best alternative would be to just repeat the cleanup code.

Hacklin:
No, you're wrong.
Unrestricted  Goto is considered harmful (like in old BASIC: GOTO <line-no>).
However, in C the goto is restricted to jumping to a label inside the same function.

In C you should use goto:
- in error handling by jumping to the clean-up code in order to remove code duplication,
  (in C you use `if (error) goto catchLabel;` instead of `try ... catch (error)` or RAII)
- to remove/reduce code duplication,
- for optimizations (speed and/or size)
  (and if goto doesn't do the job, you use assembler which has the branch/jump instruction).
- to break out of multi-nested loops (e.g. like a break 2).

Personally, I also used goto once to add a feature to 3rd party source code.
Without the goto the source would not automatically merge
causing unnecessary work with each update of the 3rd party sources.

Notes:
- Only teachers consider goto to be evil.
  C programmers consider goto an advanced feature
  (one with which you can shoot yourself the foot).
- This is not universally accepted by the likes of
  Brian Kernighan and Dennis Ritchie (C), Donald Knuth (Pascal, Modula)
  and Linus Torvalds (Linux).
- The libpng code uses setjmp/longjmp for error handling from its library.
  (setjmp/longjmp is considered to be worse than goto.)
- Code duplication is also considered a bad idea ;-)
- Some people even consider break from a loop, continue
  and multiple returns from a function bad.

The ES code handling libpng is basically C not C++.
If you want to do it the C++ way, you should throw errors,
wrap the function in a try/catch and do the error clean-up in the catch,
I guess.

Note that I use goto to jump to the error clean-up code
and not to the clean-up code at the end of the function,
so that one could add additional error handling/reporting in one place.

abenkovskii:
@endless-sky I think you can replace this goto with a call to `longjmp()` if it better matches your taste. 😄

abenkovskii:
@endless-sky You could also do this https://github.com/endless-sky/endless-sky/compare/master...abenkovskii:pr-2983-alt if you are more into RAII. This is all kinda moot right now though because the rest of the game doesn't appear to handle `ImageBuffer::Read` returning `nullptr` gracefully. I tried truncating a png file to the first 20 bytes. `ImageBuffer::Read` correctly returned `nullptr` but then the game hang.

abenkovskii:
You are also missing a `volatile` qualifier on `buffer` and `rows`. Both this variables have automatic storage duration, are local to the function and can change between a call to `setjmp()` and a call to `longjmp()`
http://www.cplusplus.com/reference/csetjmp/setjmp/
> The state of the calling environment includes the values of all accessible objects, **except those of automatic duration local to the function which do not have volatile-qualified types and which change before the call to longjmp; these have indeterminate values.**

Hacklin:
I've looked into setjmp/longjmp and
distilled it to the following:


## setjmp / longjmp
================

The exact behavior of setjmp and longjmp and
their surrounding code in the function that calls them,
depends on the OS and compiler used.

- **int setjmp(jmp_buf env)**
    a) Setup:   Save    the stack context in   env and return  0,
    b) Longjmp: Restore the stack context from env and return !0.
    The stack context contains a.o. the PC (Program Counter),
    the SP (Stack Pointer) and the FP (stack Frame Pointer).
    The stack context may or may not contain the process signal mask.
    A setjmp() can only be called as:
        1) a single statement while ignoring its returned value,
                `(void)setjmp(env);`
        2) the condition to an if, switch, while, or for statement
                `switch (setjmp(env))`
        2a) a condition compared to a literal integer value,
                `if (setjmp(env) < 9)`
        2b) a condition negated by an unary not,
                `if (!setjmp(env))`
    In practice setjmp() can also be called as:
        1a) a single assignment statement,
                `e = setjmp(env);`
        2c) the condition while being assigned to a variable,
                `switch ((e = setjmp(env)))`

- **void longjmp(jmp_buf env, int val)**
    Jump to the setjmp() which saved the given env, and
    have this setjmp() restore the stack context env
    and return (val != 0)? val : 1 .
    The stack will be rewind but not unwind.
    A longjmp() MUST NOT be called with an invalid jmp_buf.

- **Lifetime of function calling setjmp()**
    A longjmp() MUST NOT transfer execution to a non-existing setjmp().
    When the function (or method) that calls setjmp() no longer exists
    (e.g. the function returned/exited, the method's object was deleted,
     the function's thread/process was destroyed),
    the jmp_buf setup by setjmp() is no longer valid and
    longjmp() MUST NOT be called with an invalid jmp_buf.

- **Variable definition**
    All automatic (stack) variables that are to be used (or destructed)
    in multiple setjmp branches, must be defined before the setjmp().
    All variables declared after setjmp() will cease to exist
    without calling their destructors,
    when setjmp() returns due to a transfer from longjmp().

- **Volatile**
    All automatic (stack) variables that are defined before a setjmp()
    and modified after the setjmp() must be declared as volatile
    to force the compiler to write the variables to the stack
    example:

```            int* i_ptr = new int(5);
            volatile int i = 1;
            if (setjmp(env)) { /* return from longjmp() */ }
            *i_ptr = 4;
            i = 2;
```
        The variable i must be volatile,
        because we change its value on the stack.
        The pointer i_ptr does not need to be volatile,
        because the compiler will write i_ptr on the stack before setjmp()
        and we do not change the pointer itself after setjmp();
        we only change the value pointed to by i_ptr.

- **C++ temporary objects**
    Do not use a C++ construct that causes a temporary object to exist,
    when a longjmp call may interrupt the execution flow
    before the destructor of the temporary object could be called.
    example:
           ` fn( Object(initValue) );`
        if the longjmp() is called from fn(),
        the destructor of Object(initValue) will _not_ be called.

- **C++ try/catch**
    You cannot use a C++ try/catch and a C setjmp/longjmp in the same function.
    Doing so can cause problems with stack unwinding and object destruction.

- **C++ noexcept**
    C++ functions/methods that call longjmp() cannot have the noexcept specifier.

- **VC++ __try / __catch / __finally**
    Non-portable Microsoft extension to catch structured exceptions.
    [Microsoft recommends not to use these in (portable) C++.]



**Notes**:
- Setjmp branch = execution path taken based on the return value of setjmp().
    example:
```
        if (setjmp(env)) {
            // setjmp branch !0 (return from longjmp())
        } else {
            // setjmp branch  0 (setjmp setup)
        }
```
- SE = Structured Exception = Microsoft name for asynchronous C errors,
    like setjmp/longjmp or math devide-by-zero errors.
    SEH = Structured Exception Handling,
    EH  = Exception Handling (C++ try/catch)
    [see also VC++ option: /EH]
- Asynchronous Exceptions = Exceptions for math/float or C setjmp/longjmp.
- Synchronous  Exceptions = Exceptions like C++ try/catch.

Hacklin:
So, if I'm wright.

The ImageBuffer pointer and the vector need to defined before setjmp().
However, if you initialize the ImageBuffer to an ImageBuffer object,
the ImageBuffer pointer need not be volatile.

abenkovskii:
@Hacklin good find. I'll need to fix my patch. There is an issue with your solution though: if `longjmp` can be called from `png_get_image_width`, `png_get_image_height`, `png_read_info`, `png_set_sig_bytes`, or `png_init_io`, we can't initialize `ImageBuffer` before calling `setjmp` because we would need to know the width and the height of the image first. We can however avoid using `volatile` by having two places where `setjmp` is called: one before `png_init_io` and another after both `rows` and `ImageBuffer` are initialized.

Hacklin:
libpng will not call longjmp for:
- `png_get_*` functions as these are simple accessors
  except for `png_get_*inch*` functions,
- `png_set_sig_bytes`.

but libpng can call longjmp for:
- `png_get_*inch*` functions which require a to/per inch calculation
  that could cause a math error or warning.
- `png_init_io`,
- `png_read_info`,
- `png_read_update_info`,
- `png_read_image`.


`png_set_sig_bytes`?
We don't read the signature bytes ourselves,
so we don't have to tell libpng that we have not read them.
Meaning the `png_set_sig_bytes(png, 0)` is superfluous,
but also harmless (it just adds a little unnecessary overhead).

However, we're not (officially) optimizing the code in this PR,
so I don't know if we need a separate PR just to remove
the call to `png_set_sig_bytes()`.

PR #2984, opened 2017-08-30
Fix not being able to take back orders that require a target

flaviojs:
To take back an order like "gather" or "focus fire" you issue it again.

When one of your ships is unparked in another system it will first travel to your system. While it is traveling, orders that require a target could be given but not taken back.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -3010,6 +3010,10 @@ void AI::IssueOrders(const PlayerInfo &player, const Orders &newOrders, const st
 		if(ship == newTarget)
 			continue;
 		
+		// Never issue orders that target a ship in another system.
+		if(newTarget && ship->GetSystem() != newTarget->GetSystem())
+			continue;
+		
 		gaveOrder = true;
 		hasMismatch |= !orders.count(ship);
 		


tehhowch:
I'm not sure about this fix - it's kind of nice to jump into a system, target a ship, and have all of your escorts target that ship as they enter the system, rather than needing you to continually press F (or G) each time a new escort enters the system.

Also, with this PR as of 0a912f4, if you issue a target to your escorts, a new one enters, and you change your mind, you have to double-tap F to cancel it - the first press enforces the targeting by the newly entered escort, and the second one cancels it.

flaviojs:
To cancel an order the selected escorts (or all) need to have orders that match the order I'm giving. Without this PR the escorts outside the current system would not retain the same target orders, so when they all arrive you'd still have to double-tap F/G to cancel it.

To fix this there were 2 options, either mess with the AI so that the player's orders are always retained while it moves independently (as a secondary objective) or don't include them in orders they can't keep.
The current AI code is not easy to understand or change (really needs some restructuring), and I'm not familiar enough with it  so I chose the other option.

endless-sky:
I'm really not certain if this improves on the current behavior. It seems like it would help in some use cases but make things more complicated in others.

flaviojs:
Current behavior means I'm unable to cancel orders when a ship is in another system, I can only assign different orders. Can you give an example of how it would be more complicated?

PR #2985, opened 2017-08-30
Enhancement: Show outfit mass in the cargo of the ship panel.

flaviojs:
This commit will show the cargo size as "(count) x (unit mass)".
If the mass is 1, only the count is reported.
Commodities have of mass of 1 so no change is required.

This allows a player to better decide what to jettison from existing cargo to make space for a big outfit which is about to be plundered.

### DIFF
--- a/source/ShipInfoPanel.cpp
+++ b/source/ShipInfoPanel.cpp
@@ -549,7 +549,14 @@ void ShipInfoPanel::DrawCargo(const Rectangle &bounds)
 			
 			bool isSingular = (it.second == 1 || it.first->Get("installable") < 0.);
 			table.Draw(isSingular ? it.first->Name() : it.first->PluralName(), dim);
-			table.Draw(to_string(it.second), bright);
+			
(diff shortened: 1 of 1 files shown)


tehhowch:
May be useful to insert some sort of non-alphabetic sort by total held mass (or resale value), if the count x unit format is kept:
image

flaviojs:
Sorting by total mass seems useful and appropriate.
Sorting by resale value only makes sense if I show the value too.

Showing the total mass seems like too much text. Hmm, maybe show a tooltip with info about value and mass? But then it might be better to go back to the old format... I'll try it out and see how it looks.

flaviojs:
Tooltips were useful but made me hunger for more information.
I found both tons and resale value to be valuable information.
Not sure what would be the best thing to display...

endless-sky:
After thinking about this, I decided that what the player is most interested in is probably the total mass, and the number of outfits is the second most useful piece of information. The mass per outfit is not as important. So, I made a change (0570847) to show those two pieces of information, instead of merging this.

cargo

PR #2988, opened 2017-08-31
Stabilize escort order and extend escort display

flaviojs:
The escort list was unusable if I unparked lots of ships from different systems. The ships would jump all over the list and I would have no idea how many were not being displayed.

Escorts are ordered first by being in the current system, second by system name, and third by the biggest cost.

The system name is only displayed below his first ship stack.

There can be aditional columns of escorts up to the center.

The number of hidden escorts is displayed.

Screenshot with a smaller window to showcase this:
endless_sky_2017-08-31_14-46-37

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -687,6 +687,7 @@ void Engine::Draw() const
 		Screen::Bottom() - 20. * messages.size());
 	auto it = messages.begin();
 	double firstY = Screen::Top() - font.Height();
+	int visibleMessages = 0;
 	if(messagePoint.Y() < firstY)
 	{
 		int skip = (firstY - messagePoint.Y()) / 20.;
@@ -699,6 +700,8 @@ void Engine::Draw() const
 		Color color(alpha, 0.);
 		font.Draw(it->message, messagePoint, color);
 		messagePoint.Y() += 20.;
+		if(visibleMessages || alpha > 0.f)
+			++visibleMessages;
 	}
 	
 	// Draw crosshairs around anything that is targeted.
@@ -767,7 +770,7 @@ void Engine::Draw() const
 	}
 	
 	// Draw escort status.
-	escorts.Draw();
+	escorts.Draw(visibleMessages);
 	
 	// Upload any preloaded sprites that are now available. This is to avoid
 	// filling the entire backlog of sprites before landing on a planet.
--- a/source/EscortDisplay.cpp
+++ b/source/EscortDisplay.cpp
@@ -19,6 +19,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Government.h"
 #include "LineShader.h"
 #include "Point.h"
+#include "Preferences.h"
 #include "OutlineShader.h"
 #include "Screen.h"
 #include "Ship.h"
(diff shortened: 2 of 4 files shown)


tehhowch:
I don't think a horizontal expansion is going to merge. To paraphrase the de facto response, the game allows large fleets, but is designed with small fleets in mind.

flaviojs:
The code is still trying to fit everything in a single column by stacking. In this example the whole list will reduce to a column and a half when they all arrive (the 23 different sprites used by my ships).

Most people will still see a single column.

flaviojs:
Simplified the merging process since this is called one every engine draw. Looks faster than the original.

Screenshot with all unsold ships in the same system:
endless_sky_2017-09-01_03-20-36

@endless-sky 
Can you confirm if showing multiple columns is a no go?

tehhowch:
You can get a profile-ready binary for gdb/valgrind/\<tool of choice\> via `scons mode=profile`

One other issue I have (aside from the column sprawl) is that a system-then-cost sort results in the same escort being nearly impossible to follow in a large stack such as your own, because it as it changes system, it will jump up, down, left, and right in accordance with the name of its current system. In the current `master`, where the cost sort has priority, an escort stack can only move up or down, and its relative position never changes.

flaviojs:
That only works well if the ships in each stack have a similar cost. In my case most ships that can stack have different costs. When part of a stack jumped system it would create a stack in another position (different system and different cost), causing an avalanche when combined with other stacks and making it impossible to follow most stacks (the first few stacks jump position a bit, the others jump position almost every frame in unpredictable ways).
After ordering by system name it becomes much more stable, making it possible to see if they're still in the same system and try to see where they are. It saves space by only writing the system name once and makes it possible to (manually) make ships hold position when they reach a particular system, so they can arrive at the same time when you cancel the hold order.

I'll try profiling it.

flaviojs:
After profiling, there was a slight improvement from 0.13% to 0.11% (according to callgrind).

Also, I noticed the icon code restricts "is here" to ships that aren't disabled (misleading variable name), which means first is active ships in the current system, then disabled ships in the current system and then each system in alphabetical order. (ordered by cost inside each group)

endless-sky:
Personally I think that extending the escort display to multiple columns looks rather ugly, and this has the additional issue that messages shown on the bottom of the screen will overlap the escort icons.

flaviojs:
Made multiple columns optinal (off by default) and it's not overlapping messages anymore.
endless_sky_2017-09-02_01-02-01

MessyMix:
Nice, but what's the behavior like when you receive multiple messages in
rapid succession? Also, for the sake of consistency and OCDness in your PR,
why don't you elevate the first column of escorts as well? That would be a
lot easier on the eyes. :)

flaviojs:
The bottom height of the extra columns will adapt itself when messages become visible or invisible (same space between icons and the topmost message). When there are no messages it will be like the other images.

The main reference point is still the bottom-left corner, changing that is not a good idea. I guess people with OCD tendencies will keep this setting off.

flaviojs:
I noticed that the current height limit is already hitting the government text (depends on game window height), so the "X more escorts" text will be over it. Needs to be adjusted.

Have been testing playing without multiple columns. Found reordering by system and showing the number of extra escorts to be pleasing. Need to fix the placement of fighters/drones that are not deployed and have the parent is in another system.

There was only one thing I found lacking. Selected ships were being grouped with unselected ships, which makes group control a bit frustrating. (assign selected ships to group with "ctrl+number", select group ships with "number")

PR #2993, opened 2017-09-02
Use fleet multiplier to simplify data/*txt files

tehhowch:
Implement the feature from #2957


Amazinite:
Damn. 712 less lines.

PR #2994, opened 2017-09-02
Re-ordered the turret on the heavy cruiser variant

warp-core:
In the old configuration, the turrets were arranged:
              [heavy laser]
[anti-missile]     [heavy laser]
             [quad blaster]
Which was inconsistent with the other cruiser variants.
The new configuration puts the heavy laser turrets in the left and right positions, the quad blaster in the front and the anti-missile at the back:
          [quad blaster]
[heavy laser] [heavy laser]
           [anti-missile]


warp-core:
Also moved the heavy rocket launchers into the innermost gun ports like the sidewinder missile launchers on the normal Cruiser.

PR #2996, opened 2017-09-03
Enhancement: Show depreciated stock of items that can be bought.

flaviojs:
An outfitter has a unlimited stock of outfits that can be bought. When you bought and sold back an outfit it would be placed with the depreciated stock which would give incorrect stock numbers (if displayed).

Now new outfits are sent back to the infinite stock and the depreciated stock is shown.

### DIFF
--- a/source/Depreciation.cpp
+++ b/source/Depreciation.cpp
@@ -169,10 +169,10 @@ void Depreciation::Buy(const Ship &ship, int day, Depreciation *source)
 
 
 // Add a single outfit to the depreciation record.
-void Depreciation::Buy(const Outfit *outfit, int day, Depreciation *source)
+bool Depreciation::Buy(const Outfit *outfit, int day, Depreciation *source, int unlimitedStockDay)
 {
 	if(outfit->Get("installable") < 0.)
-		return;
+		return false;
 	
 	if(source)
 	{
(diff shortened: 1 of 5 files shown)


flaviojs:
It wasn't my original intention, but now what you mined from asteroids will not be shown in the outfitter after you sell everything off. This seems more intuitive so I rather keep it.

Bladewood:
You can buy it into cargo, though. No point to doing that, but I think it fine for that to remain possible.

tehhowch:
Being unable to buy back mined asteroids will make mining jobs rather difficult (once added). This can be good or bad, depending on the desired difficulty (e.g. a rush mining mission could be better served by buying previously sold stock, but no-deadline missions would be better solved by mining.

flaviojs:
Indeed I didn't test the awkward buy to cargo. If you can buy it back then the amount of stock should be shown for them too.

flaviojs:
Rebased changes against master to resolve conflicts.

endless-sky:
The existing behavior is necessary to make it so that if the player buys a ship and then accidentally sells an outfit on that ship that is not in stock, they can always buy that outfit back.

flaviojs:
I already reverted the change in behavior for mined asteroids, so everything should be equivalent to the current behavior (including the trick that allows you to get outfits not sold in the outfitter "buy new ship, then sell target outfit, then sell ship").

Issue #2163, opened 2017-02-20
Make escorts refuel in the last available (and friendly) inhabited system


Tadrix:
Would be awesome if those escorts which are going to run out of fuel in uninhabited or hostile systems would try to refuel at the last possible place before it, if not already full. Or at least give the player option to command them to refuel now.

(Another thought, although perhaps warranting another issue, would be to put a tanker/resupply ship in the game with practically no cargo capacity but huuuge fuel hold. Differences between that and converted ship with fuel pods? Efficiency, appearance and AI behavior (also, good shields+hull, but crappy weapons capacity, some small self-defense turrets, but no guns — no missiles either). Namely, escorts are trying to help each other by boarding other escorts and giving them enough for another jump, and do it only when no hostiles are around. To resupply, other ships (interceptors, other small ships) would themselves dock and get a full tank, or at least half-full, if others are in need and fuel is running low. This fits nicely in the narrative of far expeditions and better ramscoops. But I guess such thing isn't something that would become a priority, so just spoke my mind.)

Amazinite:
Currently, they actually will do this if they have 1 jump left and the next syst they are going into is uninhabited while the current is inhabited, but not beyond that.

Rakete1111:
They will also refuel if there are no friendly planets in the next system.

Amazinite:
Related to the second point, I would like to see the AI be smarter in how they ask for fuel and get it. The faster ship should always fly to the slower ship, not the requesting ship waiting for the offering ship.

jafdy:
What counts as "faster"? What if one ship could turn faster and one could travel faster in a strait line.

Tadrix:
@jafdy @Amazinite probably meant it as "which ship can faster make it to the other, performing all necessary turns in the process". And it's probably possible for the game to take this into account, seeing as it manager to calculate unparked fleets getting to the player. It's an awesome idea.

tehhowch:
It's possible that a simple Ship::MaxVelocity comparison could be used for non-Player Flagship requestors. For example: If the hailed ship has twice the speed of the requestor, the requestor issues itself a MOVE_TO order for a point that is 1/3 of the way to to the other ship. No change in behavior of the hailed ship would be needed, as it sets a rendezvous target with the hailing ship rather than the position of the hailing ship at the time of the request.
This would be a simple quality of life improvement in AI-to-AI refueling operations, but note that this is the only time such a comparison would be used - the requestor can't move at all if it's disabled.

[...]

PR #2997, opened 2017-09-03
Enhancement: Consider next refuel stop when traveling to parent ship's system

flaviojs:
An escort traveling to the parent ship's system could get stuck when it encountered two consecutive system where it could not refuel because it only checked if it had enough fuel for the next system.

Now it checks how much fuel is required to reach the next refuel stop and uses that to decide if it should refuel in the current system.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1246,7 +1246,7 @@ void AI::MoveEscort(Ship &ship, Command &command) const
 						break;
 					}
 				}
-				// Is next refuel system?
+				// Stop tallying fuel use once the next system with fuel is located.
 				if(to->HasFuelFor(ship))
 					break;
 				// Try with the next system in the route.
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1194,40 +1194,65 @@ void AI::MoveEscort(Ship &ship, Command &command) const
 		Refuel(ship, command);
 	else if(!parentIsHere && !isStaying)
 	{
-		// If this ship has already refuelled, and its parent has left the
-		// system, no need to land on a planet again.
-		if(ship.GetTargetStellar() && ship.Fuel() == 1. && ship.GetTargetStellar()->GetPlanet()
-				&& !ship.GetTargetStellar()->GetPlanet()->IsWormhole())
-			ship.SetTargetStellar(nullptr);
-		
-		// Check whether the ship has a target system and is able to jump to it,
-		// and that the targeted stellar object can be landed on.
-		bool hasJump = (ship.GetTargetSystem() && ship.JumpFuel(ship.GetTargetSystem()));
-		if(ship.GetTargetStellar()
-				&& !(ship.GetTargetStellar()->GetPlanet() && ship.GetTargetStellar()->GetPlanet()->CanLand(ship)))
-			ship.SetTargetStellar(nullptr);
-		if(!hasJump && !ship.GetTargetStellar())
+		if(ship.GetTargetStellar())
 		{
-			// If we're stranded and haven't decided where to go, figure out a
-			// path to the parent ship's system.
+			const Planet *targetPlanet = ship.GetTargetStellar()->GetPlanet();
+			if(!targetPlanet || !targetPlanet->CanLand(ship))
+				ship.SetTargetStellar(nullptr);
+			// If this ship has already refuelled, and its parent has left the
+			// system, no need to land on a planet again.
+			else if(!targetPlanet->IsWormhole() && ship.Fuel() == 1.)
+				ship.SetTargetStellar(nullptr);
+		}
+		
+		if(!ship.GetTargetStellar() && !ship.GetTargetSystem())
+		{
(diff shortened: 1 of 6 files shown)


source/AI.cpp
```diff
@@ -1194,40 +1194,65 @@ void AI::MoveEscort(Ship &ship, Command &command) const
...
+			// ship should refuel, land on a wormhole or jump to the next system.
 			DistanceMap distance(ship, parent.GetSystem());
 			const System *from = ship.GetSystem();
-			const System *to = distance.Route(from);
-			for(const StellarObject &object : from->Objects())
-				if(object.GetPlanet() && object.GetPlanet()->WormholeDestination(from) == to)
-				{
-					ship.SetTargetStellar(&object);
+			double requiredFuel = 0.;
+			while(from != parent.GetSystem())
```
tehhowch: What happens if the ship cannot route to the parent's system?
You should probably break as soon as requiredFuel is >= the ship's fuel capacity.

source/AI.cpp
```diff
@@ -1194,40 +1194,65 @@ void AI::MoveEscort(Ship &ship, Command &command) const
...
+				for(const StellarObject &object : from->Objects())
+					if(object.GetPlanet() && object.GetPlanet()->WormholeDestination(from) == to)
+					{
+						useWormhole = true;
+						if(from == ship.GetSystem())
+							ship.SetTargetStellar(&object);
+						break;
+					}
+				
+				if(!systemHasFuel || !hasFuelCapacity)
```
tehhowch: You can probably move the `to->HasFuelFor` check to this line.

source/AI.cpp
```diff
@@ -1194,40 +1194,65 @@ void AI::MoveEscort(Ship &ship, Command &command) const
...
+				if(!useWormhole)
+				{
+					if(from->Links().count(to))
+						requiredFuel += ship.HyperdriveFuel();
+					else if(from->Neighbors().count(to))
+						requiredFuel += ship.JumpDriveFuel();
+					else
+					{
+						// How do I get to the next system? Refuel just in case...
+						requiredFuel = ship.Attributes().Get("fuel capacity");
```
tehhowch: Is a ship that can't reach the parent's system going to refuel after every jump it makes?

tehhowch:
As of #2967, "stranded" escort ships will maximize their ramscoop generation, instead of drifting aimlessly - that alone should really mitigate some issues with stuck escorts.

source/AI.cpp
flaviojs: This is a failsafe for when this code doesn't know how DistanceMap decided you can reach the next system in the route. It should never happen for now, but might happen when DistanceMap is modified.

source/AI.cpp
flaviojs: Nope, the ```to->HasFuelFor``` is to know the next refuel system. This check is to avoid checking the next refuel system when I can't refuel in the current system.

source/AI.cpp
flaviojs: Will do.

source/AI.cpp
tehhowch: Fair. I had been thinking more of just the first time through - "if the next system has fuel, we don't need to care how much fuel we need to get there if we can't refuel here"

source/AI.cpp
tehhowch: If the route is empty it throws nullptr back, so you will fall down to this, e.g. non jump escorts capped in FW space and you jump to Coalition space.

flaviojs:
That speeds up recovery after being stranded, this avoids being stranded. :)

I'm curious as to why they don't ask for help and always seem to be safe. Do ships spawn in remote systems?

source/AI.cpp
```diff
@@ -1237,17 +1240,19 @@ void AI::MoveEscort(Ship &ship, Command &command) const
...
 					{
-						// How do I get to the next system? Refuel just in case...
-						requiredFuel = ship.Attributes().Get("fuel capacity");
+						// This is a failsafe for when DistanceMap decides we can travel to
+						// the next system in a way that this code doesn't know about.
+						requiredFuel = fuelCapacity;
 						break;
 					}
 				}
+				// Is next refuel system?
```
tehhowch: `// Stop tallying fuel use once the next system with fuel is located.`

tehhowch:
Ships do not spawn in remote systems, only your own, or an immediately adjacent system if a fleet that was picked to spawn in yours is also chosen to `Enter` from that system.

One of my eventual goals is to extend combat to any system in which a ship that is either yours, or of the `escort` personality, is located. Such an extension would prevent the stranded NPC/escort in otherwise-populated systems (like Wei).

tehhowch:
Ref #2163

flaviojs:
Hmm, maybe required fuel calculations should be moved to DistanceMap so it can be reused in other "should I refuel now" decisions. It could also be used to plan better routes to the target system.

flaviojs:
DistanceMap is already tallying fuel so I used that instead.

PR #3001, opened 2017-09-03
Out-of-system escort tweaks

tehhowch:
- Carried ships that are in remote systems when the player takes off will start out loaded into their carrier ship, just as they would in the player's system, rather than needing to board.
    - This prevents any possible "dropped fighters" in case 1) the carrier is Stepped before the fighter is able to ask it to become its parent, and 2) the carrier is Placed in such a manner that it can jump immediately
 - Ships in remote systems will recharge their shields, hull, and energy, if they have the equipment to do so, rather than starting out with whatever shields/hull they had a day ago, and 0 energy (if loading save) or their previous energy.

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -934,15 +934,22 @@ void PlayerInfo::Land(UI *ui)
 	for(const shared_ptr<Ship> &ship : ships)
 		ship->UnloadBays();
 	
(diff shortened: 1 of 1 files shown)


tehhowch:
Refs https://steamcommunity.com/app/404410/discussions/1/1474221865192461789/

Issue #2188, opened 2017-02-25
recieving mission cargo in space makes other cargo invisible on the cargo list


Kryes-Omega:
This may be rather low priority, but I thought I would mention it. I was working on assisting cargo missions when I noticed that when you receive cargo from a mission while you are in space, such as assisting or boarding, it will remove other cargo you have on board from the 'cargo' list on the ship info screen, but keep it on your ship.

tehhowch:
Interesting bug.
Looks like boarding missions that are not designed to be declined are not currently supported - missions offer even if you do not have space for their cargo or passengers, which can give you negative capacities.
I've fixed that issue, and added support for blocked messages to be displayed if the boarding/assisting mission was chosen to offer but cannot be accepted. For boarding/assisting missions, it does not *currently* make a lot of sense to have blocked messages - the NPC will be either boarded for plundering or assisted and no longer disabled, so it doesn't make much sense to hint to the player that they can get a mission by "returning" to this ship, but by receiving a blocked message the player learns that a new mission is possible if the player boards other similar ships with more available space/bunks.

Still working on the cargo "offset" issue raised here

tehhowch:
Should be addressed by the merging of #3005

Issue #1663, opened 2016-10-02
Spawning mission ships through boarding


OmegaLokkes:
New issue, dealing with assisting and boarding again. Almost all the normal mission characteristics can be used when starting missions through boarding/assisting, except for spawning mission ships, which only happens upon departing a planet/station. Perhaps changing it so ships can be spawned mid-flight?

Amazinite:
A related issue: spawning missions on mission NPCs. Pretty much mission-ception. You can't do it right now. Ex: If a merchant is part of an NPC mission, you will never be able to get the merchant assisting mission.

tehhowch:
This is achieved in #3005

Issue #58, opened 2015-07-03
Negative cargo


johndh:
When switching flagships and parking your escorts, it's possible to end up with a negative amount of cargo space.  When you have negative cargo space and try to strip a disabled enemy of their outfits, those outfits instead get multiplied.  You can capture this ship and it will have lots of negative outfit space and lots of mass.
![screenshot from 2015-07-02 22 59 17](https://cloud.githubusercontent.com/assets/692397/8492889/3101efb4-2117-11e5-90ed-d74365a6e2cd.jpg)
![screenshot from 2015-07-02 23 02 08](https://cloud.githubusercontent.com/assets/692397/8492888/31018fe2-2117-11e5-82d3-2e3406bf1a66.jpg)

endless-sky:
Well, the second half of that bug (the boarding panel) is easy enough to fix: 34d21dc. And I found another bug in the process, where by swapping flagships you can end up with a "parked" flagship (bda6105).

But, I'm having trouble reproducing the condition where your ship has negative cargo space. If I buy enough cargo to fill my holds, then swap flagships, then park the non-flagship, when I take off from the planet all the excess cargo is sold off and I end up with zero cargo space free, not a negative amount.

Are you running the latest release?

johndh:
At the time, yes, I was running the latest release.  As I recall, I think I had just bought a bunch of mass expansions.

endless-sky:
I can't find a way to end up with negative cargo space using the latest code, and the ability to exploit that situation in the boarding panel has been removed by commit 34d21dc, so I'm going to close this for now.

azekeil:
You can get negative cargo space if a mission gives you cargo when you are already full... this is running 0.84.

PR #3005, opened 2017-09-05
Improve support for missions offered and actions performed while assisting/boarding

tehhowch:
@Kryes-Omega mentioned that there were some issues with missions that offered on boarding, and I took a look into it.

This PR fixes a host of issues with missions that offer from ships, primarily
 - ~~Offering when the player has insufficient cargo space / bunks~~ (ref https://github.com/endless-sky/endless-sky/commit/1d780317b9cfb4e1afb42bcd288dde344667a480)
 - ~~Shifted / missing displayed cargo~~
 - No "blocked" messages would be displayed
 - Mission NPCs would not spawn until after the player next took off from a planet
 - Gifting of outfits assumes the player is landed

With this PR
 - Missions chosen to be offered, but unable to be offered due to insufficient space, will show a blocked message if one was specified.
 - ~~If having accepted a mission from boarding/assisting a ship requires the player to jettison some cargo, the ShipInfoPanel is displayed and cannot be closed until the player has non-negative free space remaining.~~
 - ~~When jettisoning excess cargo, the minimum amount to jettison is pre-filled, instead of however much of the selected item was present (e.g. if only 1 ton of space is needed, don't jettison 30 Jump Drives).~~
 - Any NPCs spawned by accepting the mission will appear immediately, without requiring the player to land and depart again.
Diligent and mindful mission writers will avoid in-flight spawning of ships in the player's current system, preferring instead `system + distance 1 N` or `personality entering`, or other methods of locating the NPCs in a non-origin system.
 - Gifting (outfit transfer in Mission Actions) will use the proper `CargoHold` based on if the flagship is boarding or landed.
 - `require 0` means actually 0 - you can't have the outfit on a ship that is parked, disabled, or out-of-system either.

Notes:
 - ~~If making missions which offer from hostile targets, use `on offer` conversations ending with `launch` or `flee` unless the missions are sufficiently low probability (or not `repeat`), lest the player receive the mission again upon re-boarding the same target.~~ (once boarded, a ship becomes incapable of offering a mission upon reboarding).
   - ~~If making a "silent" mission (no `on offer` conversation or dialog) that offers for hostile targets, the player will have to issue the `board` command a second time, as the first command is consumed by spawning the mission. This means the player can always notice if a call for reinforcements is made, as the first boarding attempt appears to fail. If the mission always offers & accepts, then the ship cannot be boarded.~~

### DIFF
--- a/source/MainPanel.cpp
+++ b/source/MainPanel.cpp
@@ -24,6 +24,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "LineShader.h"
 #include "MapDetailPanel.h"
 #include "Messages.h"
+#include "Mission.h"
 #include "Phrase.h"
 #include "Planet.h"
 #include "PlanetPanel.h"
@@ -469,8 +470,7 @@ void MainPanel::StepEvents(bool &isActive)
 				&& !event.Target()->IsDestroyed() && flagship && event.Actor().get() == flagship)
 		{
 			Mission *mission = player.BoardingMission(event.Target());
-			const CargoHold &cargo = flagship->Cargo();
-			if(mission && mission->CargoSize() <= cargo.Free() && mission->Passengers() <= cargo.BunksFree())
+			if(mission && mission->HasSpace(*flagship))
 				mission->Do(Mission::OFFER, player, GetUI());
 			else if(mission)
 				player.HandleBlockedMissions((event.Type() & ShipEvent::BOARD)
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -28,6 +28,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Mask.h"
 #include "Messages.h"
 #include "Minable.h"
+#include "Mission.h"
 #include "NPC.h"
 #include "OutlineShader.h"
 #include "Person.h"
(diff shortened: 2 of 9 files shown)


tehhowch:
Issue references: 
#2188 
https://github.com/endless-sky/endless-sky/issues/58#issuecomment-153521198

Does not address #1663 yet.

tehhowch:
#1663 should be addressed now as well.

tehhowch:
Test missions: 
[boarding_fixes.txt](https://github.com/endless-sky/endless-sky/files/1286508/boarding_fixes.txt)

tehhowch:
With #3371 , the last caveat I noted:
>If making a "silent" mission (no `on offer` conversation or dialog) that offers for hostile targets, the player will have to issue the board command a second time, as the first command is consumed by spawning the mission. This means the player can always notice if a call for reinforcements is made, as the first boarding attempt appears to fail. If the mission always offers & accepts, then the ship cannot be boarded.

will no longer be true: truly silent reinforcement requests will be able to be made, and the player won't have to press the board key twice to plunder the NPC that created the mission. For non-silent missions, the player will have to board again if they want to plunder / assist (which allows the mission's `on offer` to kill that NPC if desired).

tehhowch:
Finally found a solution to adding the ships to the game mid-flight that I'm happy with - the previous method was pretty kludgy and essentially hoped for the best by calling Engine::Place from MainPanel::Step.
With d8cdda4, the NPCs to be added mid-flight are added by Engine during its normal "add ships to the universe" method, rather than at some random point in time.

source/MainPanel.cpp
```diff
@@ -109,8 +111,13 @@ void MainPanel::Step()
 		{
 			// Boarding events are only triggered by your flagship.
 			Mission *mission = player.BoardingMission(event.Target());
-			if(mission)
+			if(mission && mission->HasSpace(player))
```
endless-sky: It would make more sense for boarding missions to never offer if they would make you jettison cargo.

Also note that the check actually needs to be done based on whether the player's flagship has space; Mission::HasSpace() is checking the CargoHold object in the player info.

endless-sky: I've addressed the latter point in commit 1d78031, since it was a bug in the existing code.

tehhowch: This PR updates Mission::HasSpace to check the appropriate CargoHold. However, by not allowing the missions to offer if any jettison, even of commodities, is required, I can drop the forced ShipInfoPanel bits.

Pointedstick:
What's the status of this?

tehhowch:
Ready for merge, does what it says on the tin :)

tehhowch:
@Pointedstick Feedback on my interpretation of what it means for an outfit to labeled with `require 0` would be nice. Otherwise playtest feedback would be good. Content that needs this would be even better :)

Amazinite:
The implementation of `require 0` makes sense to me. If a mission has `require "Jump Drive" 0`, then you can't have a jump drive. Is there any other sort of implementation that it could possibly be?

tehhowch:
The current one 😀  :
https://github.com/endless-sky/endless-sky/blob/810b50e302430e056db764f4179786e6154c754d/source/MissionAction.cpp#L317-L332

In the current implementation, the game starts with the number of that outfit in the player's `Cargo()` (which is the collective cargo and supposed to be empty if you're flying, or whatever outfits and commodities you had in the holds of the ships that landed on your current planet), then checks the in-flight cargo of each ship in your fleet (including your flagship). Parked ships can't hold any cargo, so their inclusion here doesn't matter. Installed outfits are only considered for the flagship.

Here the primary difference is in the last sentence - require 0 is a more exhaustive check, looking at the installed and carried outfits of every single ship in your fleet (unless that ship was destroyed). There is also a difference in standard "require N" when flying, in that the check of in-flight cargo is done only for ships in the same system that aren't disabled.

Issue #3014, opened 2017-09-07
Can sell disabled ships by scrolling past the end


abenkovskii:
v0.9.8
Linux
I know a similar issue https://github.com/endless-sky/endless-sky/issues/2870 was already fixed but I can reproduce this one in v0.9.8.

To reproduce:
1. Disable a ship you own in a system with a shipyard
1. Land
1. Enter shipyard
1. Click on one of your ships in the top right
1. Scroll through your ships using keyboard arrow keys until you select the disabled ship
1. Sell the ship
1. Profit

Expected:
Can't sell disabled ships

Actual:
Disabled ships are not visible in the shipyard but you can still select them using arrow keys

tehhowch:
Should be fixed by #3015 

(good find)

endless-sky:
Fixed by merging #3015.

PR #3015, opened 2017-09-07
Arrow keys should not select disabled, in-system ships

tehhowch:
Refs #3014


abenkovskii:
It looks like this commit did fix the bug.

Issue #3016, opened 2017-09-08
Escorts travel through unexplored systems


flaviojs:
I captured a ship and jumped to an unexplored system (with jump drive). That system didn't have any connection with known systems or their linked systems.

Since there is no known hyperspace connection shown in the map, I was expecting the new escort to be stranded. Instead, I saw it happily traversing unexplored systems as if it knew about them, not even revealing them to me when he "discovers" them.

What is the correct behavior escorts should have in this situation?

Bladewood:
I think there was another (ancient) issue about this, but I can't find it.

tehhowch:
I think the escorts using systems unknown to the player should probably be avoided. This would be an `IsYours` check in DistanceMap, and then probably a HasSeen requirement in the routes' systems

Amazinite:
Would this only apply to the escort government or would it apply to all escorts? I'd prefer the former as the latter could become a nuisance with missions.

tehhowch:
IsYours means exactly what you think it does. :)

`personality escort` is generally what is used for mission NPCs that tag
along with you. There are a few FW missions that directly use the Escort
government, but these aren't exactly using unknown paths or destinations.

flaviojs:
Fixed by #3020.

endless-sky:
This is intentional, because the small loss of realism involved in an escort's captain knowing about a system you don't know about seems less important than allowing escorts to pick the shortest path to come to you. (For example, because you just unparked them or told them to stop holding position.)

It would also add complexity if player escorts had different rules for following their parent ships than all other escorts do.

flaviojs:
Hmm, then I suppose this is a ```wontfix``` and should be closed? The alternative of having your escorts discover the system/wormhole for you would complicate events that would trigger if you discovered a system or wormhole yourself.

PR #3020, opened 2017-09-08
Escorts avoid travel through unexplored systems.

flaviojs:
Reference: #3016


flaviojs:
I detected some unexpected issues so this PR is not ready.

flaviojs:
The fact that it can search in both directions is not clear in the code, which is confusing.

flaviojs:
Issue marked as ```wontfix```.

PR #3024, opened 2017-09-09
Use IsAccessible rather than CanLand to ID unalterably unlandables

tehhowch:
CanLand also checks politics, and the player may be able to bribe for clearance.


endless-sky:
I'm not sure why you're reversing the order of the checks. For nearly all planets, the IsInSystem() check will be quicker than the IsAccessible() check, so it makes sense to keep that first.

tehhowch:
Addressed

Issue #3002, opened 2017-09-03
Changing illumination direction?


endless-sky:
While working on the Remnant ships, I played around with the ship illumination and found that part of why ES ships tend to look flat and cartoony is because they are rendered with the illumination coming from nearly overhead. I've created a [plugin](https://endless-sky.github.io/dark-ships-plugin.zip) with the ships re-rendered with the light coming in at a sharper angle, and I'm interested in what others think of the results.

For example, here are the Kor Sestor ships using the current sprites:

![before](https://user-images.githubusercontent.com/11450262/30006476-490750aa-90c7-11e7-8f64-319fb8015987.jpg)

Here's what they would look like with a sharper illumination angle:

![after](https://user-images.githubusercontent.com/11450262/30006480-575ab3f4-90c7-11e7-8e43-29373d4e0dcf.jpg)

And, here's a big fleet of human ships, using the plugin:

![fleet](https://user-images.githubusercontent.com/11450262/30006509-dd52d93c-90c7-11e7-9bde-8f0b86057487.jpg)

The exact illumination angle and the strength of the "fill lights" added to keep the shadows from being pure black might need some tweaking, but I wanted to get more feedback before putting too much more time into this.

Lineth:
Yes please.

Bladewood:
OH GOD.

Bladewood:
Uh, link broken.

Rob59er:
Should make this a DLC

endless-sky:
Fixed the link. Sorry about that.

Kryes-Omega:
The ships definitely look a lot more 3D, the shadows are really nice on the Sestor ships, but my concern would be that the ships would become _too_ dark, and some of the shadows would blend the ship with the black of space.

tehhowch:
@comnom @Elyssaen 
Care to share the before/after for the Trin assets with these possible changes? I know the existing sun angle had caused some rather prominent reflections on some of the larger ships.

Personally, I'm for any changes that remove "cartoony" and replace it with "gritty" or "realistic."

The last cartoony bit is of course the issue of statically rendered shadows, and a discussion on changing the default graphics would be remiss without questioning if/how we could add support for, but not require, dynamic renders of the ship shadows and lighting (e.g. 595 plus more).

comnom:
[...]

PR #3033, opened 2017-09-11
Normal mapping

dplepage:
As requested in the comments on #3002, here's a rough sketch of what ES would look like with normal-mapping (as discussed in #595).

This PR is not ready to be merged, as it has a number of flaws (listed below), but I thought people would want to see it as soon as there was something to see, and provide feedback on whether it's worth taking this any further.

**Limitations of this PR:**
 * Ships without blender models aren't normal-mapped (they still have their old images). These include all the "modified X" ships (which are all hand-drawn variants of rendered ships), the Korath Raider (which was drawn in some SVG editor), the Void Sprites (which are doctored photographs of sea swallows), and the two Hallucinations (which are photographs of flowers). If we want these to be dynamically lit, we'd need to model them in blender (or hand-draw normal maps for them).
 * The normal-mapped ships are missing all the details that were painted on by hand in the GIMP. These details aren't hard to add back, it'll just take some time to do it. This is particularly noticeable on the Pug ships, where the yellow color was added to the models after the fact.
 * For reasons I don't yet understand, a handful of the ships are completely messed up - they seem to be getting masks and normal maps either from other ships or maybe from random data. This affects the Albatross, the Starling, and all Coalition ships. I have no idea what's different about those ships that makes this happen; I'm still trying to figure it out. If anyone has any insight I'd love to hear it.
 * Right now the light always comes from the upper right of the screen. I'd like to tweak it to come from the actual position of the sun (or suns, in systems with more than one).
 * I didn't implement emission maps yet, so ships that normally light themselves via their own glow (e.g. the Quarg Skylark) currently lack those.
 * At some point I'd like to try implementing [horizon mapping](http://www.ppsloan.org/publications/bs.pdf) or something similar for dynamic shadowing, but that also requires a lot more textures per sprite than we may be comfortable with (and generating them is a somewhat technical process, raising the barrier for new people wanting to add custom sprites).


Killerhurtz:
Just going to point out: there are software solutions for generating most types of maps with little to no know-how or work, especially no work involving creating something from scratch in Blender. 

There's http://cpetry.github.io/NormalMap-Online/ that does not seem to have any type of license limits (though I AM investigating whether it does or not with the owner and developer of the tool), and for those who want an alternate solution, I do believe that CrazyBump would be usable in all licenses (including free trial, within the specified time) for our own purposes (the only limitation I can see being that the educational license disallows commercial usage - and since this project is open-source and free, I do believe it does not qualify as commercial). 

So even for non-Blender assets, creating a normal map may be trivial.

jafdy:
https://drive.google.com/open?id=0B0mWx-Fhhhz7azFWemxBMXcwRmc 
This has the marauders along with pointed's vanguard (with some of Wrzl's other experiments) in it.

tehhowch:
Horizon mapping sounds interesting.

I compiled the PR and i think it is looking pretty dang good. I do have the dark-ships-plugin active as well:
image
image
image
image
image
image
image
image
image
image
image

comnom:
@dplepage The masking problems may come from the fact that your rerenders have no margin around the ship:

collision

I've tried this out, and still maintain that static shadows produce a more realistic looking ship, but in fairness I'll reserve any harsh critiques until at least a couple ships have the post-processing added back in, as that's where a great deal of the current realism comes from.

One thing while/if you're still working on it, the different shading is a bit difficult to notice, is there anyway to make the light v. shadow a bit more contrasty/noticable? In the following (sorry for the mass of ships it was a borrowed save) I find it hard to discern the difference in shading on the bactrians even though they're all at different orientations.

cap

jafdy:
I am seeing a difference, although the mass of ships and the fact they are in almost the same direction isn't helping.
I think the Bactrian is one of the ships that A: could use a remodel, and B: kinda needs shadows.

dplepage:
Yeah, a number of these ships are designed specifically to look interesting under the existing lighting model but in ways that make they look more boring under dynamic lighting. I suspect the bactrian's vertical fins, for example, were added specifically to make a shadow that would look more interesting; the entire bactrian design is mostly horizontal surfaces and vertical walls, which are the most boring things to look at if you have dynamic lighting but no shadows.

The ships that'll look best under dynamic lighting are probably going to be the ones with a wider variety of curved or slanted surfaces, so once I have the time to do some image editing I'll probably start with those (unfortunately the saryd diagonal ships and the starling/albatross are two of the best candidates for this, so I'll need to fix that issue first).

@comnom I didn't see anything in the docs about needing a 1px buffer around the image, though it's easy enough to do. Why is that needed? Is this a bug in the mask generator?

jafdy:
The buffer is in bold in the last paragraph of the sprites section of the creating ships section of the wiki. It's not a bug it's simply because of the way the sprites are drawn.

dplepage:
Oh, I totally missed that. Adding that doesn't fix the problem with the coalition ships, though. I think the actual issue is a bug in the png loader - the failing ships are precisely those for which the flat image is a grayscale PNG instead of a color one. I'll file a bug report.

dplepage:
The patch for that is #3043.

endless-sky:
@jafdy, redoing the Bactrian model (and the Mule) is on my to-do list. The original is from 3 years ago, and my modeling skills were pretty rudimentary back then...

Amazinite:
I'm interested to see what comes of this remodel.

Lineth:
I do as well. Would prefer for them to maintain the asymmetrical look.

dplepage:
I merged in master to get the grayscale fix, and also applied the existing GIMP modifications to the flat models for all the coalition ships. So if you want a better sense of what this will look like, go hang out in any coalition system and watch the ships coming and going.

Amazinite:
You seem to be resizing the borders on a lot of the images (removing a pixel or two off the borders but not resizing the actual ships), and some images (see Spindle, Transport, and Bastion) are even changing the size of the ships.

dplepage:
This is true, and is because I don't actually have a lot of time to work on this so I'm automating as much as possible.

When I first did this I didn't notice the transparent borders, so I took my (borderless and automatically generated) renders and just resized them to match the size of the sprites used by the game. This makes every ship a few pixels larger.

I was more careful when I redid the coalition ships yesterday to actually put a transparent border, but since a lot of my pipeline is still automated it's hard - the transparent borders on the existing ships seem to be kind of arbitrary and hence difficult to replicate automatically (and of course the transparent padding has to be the same on the flat image and the normal image).

So instead I took the borderless images, resized them to be 4 pixels narrower & shorter than the existing sprites, and then adding a 2-pixel transparent border around them.

Much of this would be simpler if instead of requiring transparent borders we used borderless images and had the image loader add buffers as needed. I may implement that at some point.

tehhowch:
Having the image loader add buffers to things that need buffers seems like it makes a lot of sense.

I think the resize (and the warning about reordered constructor arguments in SpriteQueue) can be ignored in this proof-of-concept pull.

tehhowch:
<details><summary>Updated photos (made sure to remove the dark-ships plugin first this time):</summary>


image
image
image
image
image
image
image
</details>

jafdy:
It's not perfect yet but it is stunning. OK, it could use some improvements but if looks like a great step in the right direction.

dplepage:
I applied existing texture touch-ups to a bunch of ships, nearly fully-automating this with a gimp plugin (all my tools are in https://github.com/dplepage/es-normal-ships). This restores things like pug coloration and the various metallic textures that were added in post-production. It doesn't fix a handful of ships that had more manual intervention in their creation (any ship defined without blender like the raider or the void sprite, or ships with more complex gimp edits like the shuttle (flashing lights) or the tek far 109 (a composite of the blender render with a bunch of other ships, or ships where the gimp files just weren't laid out the same way as most of them).

Someone could obviously go through manually and touch all of these up, I'm just not an artist so I'm not going to try.

Next step for me: getting the actual sun position(s) loaded in the shader so that the lighting will be in the direction of the sun instead of always from the upper right. In theory this should also allow for suns with color, and I want to experiment with allowing other bodies in the system to provide colored lighting as well (e.g. red wormholes should actually light you in red).

tehhowch:
Raider got a blend file in the latest commit. Not sure about any others
that had odd resources, but @kiledennis can maybe chime in as he worked on
the Marauder files

(PS awesome work and we are glad to have you back!)

jafdy:
The lights would need to be an animated lamp layer. I will likely sort that once I know what kind of shadows I will be working with.

comnom:
So I've tried the latest version out, cruising around and checking out the shading, then grabbed a screenshot with shading and without. While cruising around the dynamic lighting was pretty cool, and on alot of the ships pretty good looking. It's (naturally) pretty dull looking in a static screenshot compared to what we have, so the real beauty comes from lots of ships on screen running around in combat and such.

With:
w_shading
Without:
wo_shading

Some pros and cons:

* Pros
	* With lots of action on the screen, the dynamic shading really shines
	* Ships with a lot of depth on the zed look good sitting still or moving
	* It eliminates the awkwardness of conflicting shadows with the current lighting
	* It opens up new avenues for other lighting effects

* Cons
	* When nothing much is going on the old way with carefully designed lighting looks better (also true of some ship designs)
	* If not directly lit, grays are flat and muddy and colors are dull
	* The shadow can become pure black. Realistic perhaps, but ugly when half a ship disappears
	* There's no distance falloff (fixable I believe)
	* Really needs something like horizon mapping to be a clear improvement over the status quo

The last con deserves it's own paragraph, and that is that it increases the barrier to entry on creating artwork for ES. 

The amount of amateur artists/coders/writers ES attracts is a clear testament to how easy it is to start creating content for the game. Now alot of what's going to be needed sounds trivial to the tech minded, or to people already experienced in CG (ie render it flat, generate a normal map), but to someone unexperienced looking to try something new it could be the difference as to whether they try at all. We should be sure to make the process as accessible as possible if we go this route.

Some food for thought: Other tools for creating normal maps aren't going to work here. We'll be relying on generating them in blender. (Crazybump, which I have, doesn't handle the alpha channel and the online tool noted above doesn't work at all when I tried it) As such, all we have right now are dplepage's scripts and a command line toolset is going to turn alot of people off.

dplepage:
> When nothing much is going on the old way with carefully designed lighting looks better (also true of some ship designs)

Yeah, there's probably no way to avoid this except to have someone carefully revisit the ship designs with the new lighting in mind. I'm not an artist (which is why my file conversion pipeline is almost entirely automated); I'm just doing my best to approximate what they'd look like if someone qualified actually spent time on it.

> If not directly lit, grays are flat and muddy and colors are dull

Yeah, I think all my ship images are darker in general than the lit ones were. This is also something that would benefit from an actual artist's intervention.

> The shadow can become pure black. Realistic perhaps, but ugly when half a ship disappears

Good point. I can add a small ambient component to fix that.

> There's no distance falloff (fixable I believe)

This'll be fixable once I have the light coming from the suns instead of from an unspecified infinitely-far-away point light. However, I also don't want ships to become pure black when you get too far from the system center (though this would be realistic).

> Really needs something like horizon mapping to be a clear improvement over the status quo

Yeah, horizon mapping turns out to be pretty complicated and possibly not worth it. The big downside is that while producing bump-maps from blender is pretty straightforward, producing horizon maps is not, and also involves a lot more images per sprite than bump-mapping did (probably extra 10-15 grayscale images per sprite). I haven't worked through the implementation yet, but it might not be worth the effort, and might raise the barrier to new content creation too high.

> Some lighting-dependant materials in blender will be unuseable with a flat render. I do quite a bit with procedural materials, and would have to switch solely to texturing in gimp

What sort of materials are you using that are lighting-dependent? Is this mainly for getting e.g. specular highlights? I've already been considering adding specular mapping to the `SpriteShader`, for what that's worth.

> As such, all we have right now are dplepage's scripts and a command line toolset is going to turn alot of people off.

Once I've got the cpp side looking good, I can probably turn my scripts into a Blender addon, so that as long as people are making their ships in Blender they can just hit some "Render Bumpmap" button to get a bumpmap out.

For reference, here are the steps I see moving forwards:
- [x] Get basic normal mapping working.
- [x] Render sample flat/normal pairs for most ships, so we can see what it looks like. This is NOT intended to be a complete, deployable version of all graphics - other people will probably need to get involved to get that finished. This is just so that as the later steps are done it'll be obvious when they're working.
- [ ] Render flat/normal pairs for asteroids because it's probably really easy and I'll bet it'd look cool.
- [ ] Teach the shader to track a real-world light source, allowing light to shine in different directions and have falloff (initially this'll just be the system center)
- [ ] See about having multiple light sources (e.g. a system with two suns and a glowing wormhole). This'll also require adding light intensity and color as a property of system bodies. There'll probably be a hard limit on lights per system because the shader's much easier to write that way.
- [ ] Add emission maps (for things like the quarg's internal glow, wing lights on shuttles, etc.).
- [ ] Add specular maps (so things can have custom shininess).
- [ ] Come up with a better way to package images - it's awkward if `images/ship` contains e.g. `shuttle-0.png`, `shuttle-0-bump.png`, `shuttle-0-emit.png`, `shuttle-0-spec.png`, and then the same for each other frame. I suspect the ideal would be to have one folder per sprite, which would contain all the different maps and frames for that sprite; it could also contain a small data file for things like hardpoints and animation durations, rather than requiring those to be in a separate place? Needs more thinking.
- [ ] Convert my blender hacks into a proper blender addon.
- [ ] Strip all the image changes out of this PR and make it just a code change.
- [ ] Package all the enhanced graphics as a plugin.

The last two items are especially important, I think - the way I'm setting up this code, it'll do fancy rendering if the normal maps etc. are there, but will render normally otherwise. This means that the code itself could be merged into the main project without the core game changing at all; advanced graphics would be an optional plugin.

jafdy:
Unless it is really bad for the code things like the shuttle could use the same base, bump, spec and horizon images as only the emit will be changing.

You are going to need to look at horizon as a possibility simply because the shadows from fins are important.

KiLEdEnNis:
> Not sure about any others
that had odd resources, but @kiledennis can maybe chime in as he worked on
the Marauder files

the drive link for them is at https://drive.google.com/drive/u/0/folders/0B5-qbs9LtB26ZGRnaTlVLXlaMlk

but if ya wanna be sure, use https://drive.google.com/drive/folders/0B9aK8dG39P29bmNQZlQxZ0pBZEU (the actual ES google drive, since idk if MZ edited them after me, but I think not)

comnom:
> The big downside is that while producing bump-maps from blender is pretty straightforward, producing horizon maps is not, and also involves a lot more images per sprite than bump-mapping did (probably extra 10-15 grayscale images per sprite)

Ouch. Yea that's a bit much.

> Is this mainly for getting e.g. specular highlights?

In a nutshell yes. Specular maps would cover the most common use cases. I think there's even a way to do anisotropic highlighting (for brushed metals) with spec maps.

> Once I've got the cpp side looking good, I can probably turn my scripts into a Blender addon, so that as long as people are making their ships in Blender they can just hit some "Render Bumpmap" button to get a bumpmap out.

This would be ideal. With an addon and a template for the flat render, I don't think we could make it any easier. 

All in all it's looking pretty cool; with some polish it could be great. Looking forward to seeing how it turns out.

jafdy:
Any more news on this?

tehhowch:
Busy busy. Just like MZ had been busy the past week, it's likely OP is too. While I can suss out where the vector to the light source is supposed to go, I haven't the foggiest idea how to pass the light source's position variable (s) into shader code so that the GPU can compute the vector. 

OP wrote a checklist and has published his pipeline tools so anyone who can help is likely welcome to do so.

My thought on the light source limit is 4. A ternary star system and a wormhole? That's a busy system. Unless this count is also including each ship as an emitter, in which case 48 or 64 is likely a better choice (e.g. running lights on ship A light up ship B). Some players sure have more than 48 ships in a system, but with that many you can't really tell what's lighting what.

jafdy:
I'm keen to start work on some emission maps as the moving shadows will make the running lights a bit easier to see. I don't want to pressure anyone.

Issue #2944, opened 2017-08-19
Enhancement Suggestions: installed outfits and ships unfit to fly


flaviojs:
The short story, 4 suggestions:

1. ~~_(visual)_ show visual hints for outfits that are installed in the selected ship(s) in the Outfitter, maybe filled-in versions of "outfitter selected+.png" and "outfitter unselected+.png" or maybe change the color and/or width of the lines~~ (the frustration it's trying to solve has already been handled by the "installed only" view);
2. _(restriction)_ allow the player to leave the Outfitter when a ship is unfit to fly (making it possible to fully strip outfits from a ship and sell the empty hull in the Shipyard another day) and instead prevent unfit ships from departing the planet (unfit ships must be parked, the current conversations appear when you try to depart);
3. _(visual)_ show visual hints for ships that are unfit to fly, maybe a red ```!``` on the corner of the ship icon in the Shipyard/Outfitter (top right corner below "Your Ships:") and a red line in the "Player Info" tab;
4. _(information)_ list why the ship is unfit to fly in the details (in the Outfitter/Shipyard ship details and in the "Ship Info" tab that is next to "Player Info").

--

The long story:

I'm a newish player playing v0.9.6 in steam. I started a new pilot to try out mining while carrying people around. After upgrading my Shuttle to a Scout, I tried capturing disabled pirate ships and found it very profitable! From then on I always had a full crew on takeoff and captured what I could before landing and parking the captured ships. I got lucky in "The Deep" and captured a Corvette that was fast, had range (4 particle cannons) and had more crew bunks.

The corvette has served me very well, but I've had to pass up on capturing some big ships. Having fragmentation grenades and swapping 2 particle cannons for 12 crew bunks is not enough to capture a Falcon or a Leviathan. When I got the "Pirate Reconnaissance" mission to visit the 4 pirate systems in "The South" my combat rating was almost enough to demand tribute from Bloodsea in Antares, so I decided to try it out, but first I wanted to upgrade my ship and recycle the parked ships.

After arriving at my target shipyard system, I land, unpark all ships, depart from the planet, wait for everything to arrive, and land again (there were 126 ships in the save file). The shipyard is straightforward, but the outfitter is frustrating with many ships:

- it's hard to see what is installed on a ship with a glance because other text can appear over the outfit ("in stock" or "in cargo"), which means I have to stop and actually read the text (**see suggestion 1**);
[...]

PR #3040, opened 2017-09-12
Rework flight check

flaviojs:
Reference: #2944

**Changes in this PR**
- add warning and error icons (by MZ)
- reorganize ship warnings
- display a packed ship warning icon with a multi-warning tooltip per ship in the list of ships of the ShopPanel and the PlayerInfoPanel.
- display all ship warnings as icons with tooltips in the ship details of the ShipPanel and the ShipInfoPanel.
- add new class ShipWarnings.
- add new class TextTooltip
- move flight checks from the outfitter to take off
- show 1 conversation per ship warning and allow 4 actions:
  - go to the outfitter: available if you can access the outfitter and the ship is in the system
  - abort the take off: always available
  - park the ship: available if it's not be the flagship
  - dismiss the warning: available if it's not a serious warning


flaviojs:
Screenshot of the flight check line and tooltip (with engines removed):
flight_check

source/ShipInfoDisplay.cpp
```diff
@@ -303,6 +318,27 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 	heatTable.push_back(Format::Number(60. * emptyMass * .1 * attributes.Get("heat dissipation")));
 	// Pad by 10 pixels on the top and bottom.
 	attributesHeight += 30;
+	
+	// Flight check warnings
+	if(ship.IsYours())
```
tehhowch: Is this check needed? Surely one would want to know if there are flight warnings for any ship they can see the information of?

source/ShipInfoDisplay.cpp
```diff
@@ -303,6 +318,27 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
...
+	{
+		vector<string> warnings = ship.FlightCheck();
```
tehhowch: Why make FlightCheck() a Ship class member, rather than a ShipInfoDisplay member? The Ship class is pretty large already, and the game only uses FlightCheck() for informational purposes - the rest of Ship class members are used to instantiate ships or help them make decisions in flight (e.g. used by the Ship, AI, and Engine classes).

source/ShipInfoDisplay.cpp
flaviojs: The check makes ships for sale in the shipyard (or the map's shipyard) not show that particular line.

source/ShipInfoDisplay.cpp
flaviojs: Because it will be used in other places (this is a work in progress) and it just makes sense.
The Ship class has methods that provide information about the ship. I consider this function to be the same kind of function and the consequences of this information is left to outside code.

source/ShipInfoDisplay.cpp
tehhowch: Yes, I'm aware. If a stock ship being sold is not ready to fly, that should be communicated to the player. This is beneficial for plugin-makers, as a content sanity check, or as a gameplay mechanic (e.g. selling discount ships which are missing parts).

flaviojs:
Just to make it clear, the WIP in the title means this is a work in progress and is not ready to be merged. The reason it's already a PR is so I can get feedback.

I want this PR to address all the flight check stuff I mentioned in #2944.

source/ShipInfoDisplay.cpp
tehhowch: I suppose if you are planning to reference this data from PlayerInfoPanel (which ships have warnings), PlanetPanel/PlayerInfo (can we take off?) or OutfitterPanel/ShopPanel (can we leave the outfitter?), then it being a Ship class member makes sense.

Bladewood:
I'd suggest indenting the individual tasks or bulleting them (somehow).

source/ShipInfoDisplay.cpp
flaviojs: Good point, will remove the check.

endless-sky:
It might be better to have a warning icon appear above the ship's icon in the outfitter, rather than putting this at the bottom of the info display, for two reasons:

1. If it's at the bottom of the info display, it will overflow the height of the ship info panel.
2. If you have a lot of ships, an icon would make it easier to see which ship has issues without clicking through all of them.

flaviojs:
A warning icon (or another visual cure) above the ship icon is also planned.

I checked ShipInfoPanel before committing this and it looked fine. Originally I wanted to show the problems too but it didn't fit ShipInfoPanel so I moved them to the tooltip.

Screenshot of left edges in ShipInfoPanel:
flight_check_shipinfopanel

flaviojs:
@Bladewood
WrappedText doesn't support indenting paragraphs so I'll do simple bulleting.
flight_check_2

Amazinite:
I'd definitely prefer this over the current system if finalized.

tehhowch:
As of 1e1a6fc, the conversations are still shown when leaving the outfitter: https://github.com/endless-sky/endless-sky/pull/3040/files#diff-b6734b3a6bc3d5b29d3c6a3cd3af1a75R622

I think the intent is to move those to when trying to leave the planet, which will allow the player to park empty hulls, or decide to sell a hull after selling engines/critical systems (with the original intent to upgrade) without needing to rebuy all missing critical systems.

flaviojs:
Indeed, I intend to move those conversations to when trying to leave the planet.
Before that I'll need to thing about how multiple ship problems and multiple problematic ships should be handled by the conversation.
And at an even more basic level, I'll need to learn about the capabilities of conversations. xD

Found that the flight check line overflows in ShipInfoPanel when there are fighter/drone bays so I'll need to revisit it.

flaviojs:
For the list with your ships in the shop, I think this is enough:
flight_check_your_ships

For the flight check details, I tried a bunch of stuff to see what would work on the various places and I liked best putting icons bellow the ship image and more details in tooltips.

Here are the prototypes with text instead of icons:
flight_check_3 flight_check_shipinfopanel_2

The text maps to these flight check warnings:
* E - "no energy"
* T - "no thrusters"
* TE - "no thruster energy"
* S - "no steering"
* SE - "no steering energy"
* SP - "solar power warning"
* O - "overheating"

If this placement is fine, can someone make an icon for each flight check warning? xD

flaviojs:
@endless-sky
In order to use text tooltips in other places I'm moving relevant code from ItemInfoDisplay to a new class TextTooltip. What should I put in the copyright, keep MZ there and update to 2017?

data/tooltips.txt
```diff
@@ -289,6 +289,37 @@ tip "max:"
 
 
 
+# Ship warnings:
+tip "flight check: no energy"
+	`Insufficient energy is being generated or stored.`
+	`Install energy sources or batteries`
```
Bladewood: missing period

flaviojs:
I updated the project files but they are untested (I don't use either).

data/tooltips.txt
flaviojs: Fixed.

flaviojs:
Forgot to mention this in the log, ShipWarnings also detects when no energy is being produced (normally and far away from the sun).

endless-sky:
There are so many changes involved here that it's a bit hard to review them and decide if they all make sense or not. It might make sense to figure out the design details first, then implement the changes as simply as possible.

Note that commit ba06025 made it possible for thrusters and steering to operate, but at reduced effectiveness, if there's not enough energy to drive them at full power. In general, I think the checks could be re-organized into a set of warnings and errors:

* Errors:
  * **(no energy!)** No energy source (batteries or generator)
  * **(no steering!)** No steering.
  * **(no thruster!)** No thruster (ordinary thruster or afterburner).
  * **(overheating!)** Overheating at idle.
* Warnings:
  * **(afterburner only?)** Only mode of thrust is an afterburner.
  * **(battery only?)** Only energy source is a battery (will be stranded if battery wears out).
  * **(limited movement?)** No battery, and energy production insufficient for full thrusting or steering.
  * **(solar power?)** No battery, solar power only, and solar power far from the star will be below **{some criterion}**.

In the shops, errors and warnings could be shown with just a single icon above the ship icon:

warning icons

Instead of adding a whole new class for ship warnings, a simpler approach would be to only report the most urgent error or warning for each ship. In that case, you wouldn't need a bitmask - you could just have the FlightCheck() function return a string, and the conversation names could be changed to the same strings. It would probably make sense for FlightCheck() to be a member of Ship.

To make things even simpler, the string could end in **!** or **?** to indicate whether it should use the error icon or the warning icon.

I'm not at all sure how XCode picks those GUID-like identifiers for new files you add, so I'm not sure if your manual edits to the project file to add the new classes will work or not.

Here are normal and high-dpi error icons:

error
error 2x
warning
warning 2x

flaviojs:
Since there is a distinction between errors and warnings, I presume the errors should prevent the ship from taking off, but not the warnings?

Only knowing one issue is a problem for me, it was one of the frustrations I had with the original conversation. The order of errors/warnings might be good for some builds but it will be bad for others (ex: my build giving priority to thrusting over steering VS the other way around). Since I can only safely store outfits by leaving them installed in parked ships, I end up having to shuffle around stuff from multiple ships when I want to change one of my ships.

I've been using the labels. By knowing the problematic ships and all the problems it has I can save a lot of time. I'll try putting all the issues in the tooltip again.

---

How is high-dpi handled by the interface? I can't find a example.

flaviojs:
Implemented the reorganized errors and warnings and used the icons.

Showing the icons separately in the ship details allows me to know immediately if I fixed a problem or created a new one. To know the warning details I need to hover over the icon, which is not a problem.

This looks good so next I'm gonna work on the conversations.

Ship list in the shop (hovering multi-warning icon and dragging ship):
endless_sky_20170924_1 endless_sky_20170924_2

Ship details in the shop (hovering warning icon):
endless_sky_20170924_3

ShipInfoPanel (hovering warning icon):
endless_sky_20170924_4

PlayerInfoPanel ("ship" column is shifted to the right):
endless_sky_20170924_5

data/conversations.txt
```diff
@@ -19,37 +19,175 @@ conversation "intro"
...
+	branch batteryonly
+		"ship warning: battery only" == 1
+	branch limited_movement
+		"ship warning: limited movement" == 1
+	branch solarpower
+		"ship warning: solar power" == 1
+	
+	label unknown
+	scene "outfit/unknown"
+	`You've heard enough horror stories about the sort of accidents that happen when starship systems malfunction to be completely convinced of the necessity of a pre-flight safety check. As you listen to the complex symphony of hums, rattles, and clicks of "<ship>" warming up, you run through a mental checklist. Computers, life support, engine...`
```
Bladewood: `Computers, life support, engine...`

engines*?

flaviojs:
Moved flight checks from the outfitter to take off.

This is as far as I can take the conversation with the current capabilities.
Player conditions are used as input variables.
Each ship warning shows a conversation and it allow 4 actions:

- go to the outfitter: available if you can access the outfitter and the ship is in the system
- abort the take off: always available
- park the ship: available if it's not be the flagship
- dismiss the warning: available if it's not a serious warning

data/conversations.txt
flaviojs: Fixed.

flaviojs:
@endless-sky 
I don't know what to do with the high dpi images.
All the stuff I mentioned in #2944 has been addressed, so it's no longer WIP.
Want me to rework the commits to remove discarded solutions and make the PR easier to evaluate?

Amazinite:
>I don't know what to do with the high dpi images.

Here: https://github.com/endless-sky/endless-sky-high-dpi

endless-sky:
I decided to just re-implement this in a simpler manner (abf8b78). Flight checks are a minor enough part of the game that I just don't think it makes sense to dedicate nearly a thousand lines of code to such a detailed presentation of them.

PR #3042, opened 2017-09-12
Fix a missing quotation mark in "Ask the Hai about the Unfettered"

awfulworldkid:


jafdy:
The original is correct.

PR #3043, opened 2017-09-12
Load grayscale pngs properly.

dplepage:
This patch fixes an issue where the png loader feeds grayscale png data to OpenGL as BGR data.

To reproduce, replace images/_menu/side panel.png with this image:
side panel

This image is visually identical to the one that currently ships, but since it's one-channel it appears in the game like this:

sidebar

The fix is just to make sure to call `png_set_gray_to_rgb` if needed.


dplepage:
Instead of downloading that image you can also just run `mogrify images/_menu/*` - ImageMagick uses 1-channel pngs by default if the image is grayscale.

PR #3044, opened 2017-09-13
Use selected ship/outfit as scroll reference.

flaviojs:
~~In the outfitter, when you have an outfit selected and choose a different ship it will auto-scroll the outfit description back to the same place it was relative to the screen.~~ In the outfitter, when you have an outfit selected and choose a different ship it will keep the selected outfit in the same screen height.
This is needed because when a new ship is selected it can add or remove outfits before the selected outfit, causing the selected outfit to be placed in a different position.

It is also able to auto-scroll down and tries to make the ~~outfit~~ image visible.

You can deselect the ~~outfit~~ item and cancel the auto-scroll by clicking on other areas of the main panel.

### DIFF
--- a/source/ShopPanel.cpp
+++ b/source/ShopPanel.cpp
@@ -66,8 +66,8 @@ void ShopPanel::Step()
 	// Perform autoscroll to bring item details into view.
 	if(scrollDetailsIntoView && mainDetailHeight > 0)
 	{
-		int mainTopY = Screen::Height() / -2;
-		int mainBottomY = Screen::Height() / 2 - 40;
+		int mainTopY = Screen::Top();
+		int mainBottomY = Screen::Bottom() - 40;
 		double selectedBottomY = selectedTopY + TileSize() + mainDetailHeight;
 		// Scroll up until the bottoms match.
 		if(selectedBottomY > mainBottomY)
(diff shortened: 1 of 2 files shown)


flaviojs:
Sometimes it gets stuck going up and down, need to fix that.

flaviojs:
As I used this I found the constant auto-scrolling to be slightly less annoying than old behavior, making me wish it just stayed in the same position.

Recreated PR from scratch, but instead of auto-scrolling it keeps the same height.

source/ShopPanel.cpp
```diff
@@ -64,11 +64,18 @@ void ShopPanel::Step()
...
 		DoHelp("multiple ships");
 	// Perform autoscroll to bring item details into view.
-	if(scrollDetailsIntoView && selectedBottomY > 0.)
+	if(scrollDetailsIntoView && mainDetailHeight > 0)
 	{
-		double offY = Screen::Bottom() - selectedBottomY;
-		if(offY < 0.)
-			DoScroll(max(-30., offY));
+		int mainTopY = Screen::Height() / -2;
+		int mainBottomY = Screen::Height() / 2 - 40;
```
endless-sky: `Screen::Height() / -2` is the same as `Screen::Top()`, but the latter would be easier to read.

flaviojs: Done.

PR #3047, opened 2017-09-13
Updated the Finisher Pod description to improve readability

warp-core:
This is to improve readability as the current way seems awkward to read.

### DIFF
--- a/data/coalition weapons.txt
+++ b/data/coalition weapons.txt
@@ -112,7 +112,7 @@ outfit "Finisher Pod"
 		"radar tracking" .8
 		"missile strength" 50
 		"submunition" "Active Finisher" 1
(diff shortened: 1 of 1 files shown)


tehhowch:
Maybe also strip the first `for`?

"...of their other weapons merely exist to hold a ship in place ~~for~~ long enough for a barrage of Finishers to destroy it."

`
Finisher Torpedoes are one of the Heliarchs' most dreaded weapons. In fact, in one way or another, most of their other weapons merely exist to hold a ship in place long enough for a barrage of Finishers to destroy it.
`

Bladewood:
I've said this before, but I believe another problem is in the use of the phrase `one of` with a plural `Finisher Torpedoes`, rather than using Torpedo singular or using `some of` instead.

I have a feeling this will be written off as "personal style", or something along those ends...

endless-sky:
I think this is just a matter of personal preference, not a definite improvement or a grammatical error fix.

MessyMix:
Good snipe, @Bladewood. :P

PR #3062, opened 2017-09-19
Refactor DoSurveillance

tehhowch:
There are some checks that are no longer required due to the updated order of AI::Step, making personality-based actions such as Surveillance come after combat.
 - If the ship's target ship is not targetable, FindTarget gives it a new one that is.
 - This target ship may have left the system since being targeted on a previous iteration of AI::Step, so this must still be checked.
 
 - Since the start of DoSurveillance checks that the ship is in-system (and staying in-system), this does not need to be checked again before performing ship-based surveillance actions. (This section existed before the checks were originally added to the start of DoSurveillance in https://github.com/endless-sky/endless-sky/commit/eb149c335b22021af6cb317a8c20ea23ce191995).
 - Removed now-unnecessary `isInSystem` boolean (was added in https://github.com/endless-sky/endless-sky/commit/42f361ba0c0908ba0a089d91c3686fb66b97f249 when no other in-system checks were used).
 - Don't consider ships for scanning unless this ship has the ability to perform a scan that has not already been performed (e.g. no cargo scan ability and already did an outfit scan -> don't consider that ship).

I considered replacing the `Stop` action for when the ship has no ability to scan planets, no ability to jump (currently), and no new ships to scan with a call to MoveEscort or MoveIndependent, but those can result in the ship deciding to land and stop surveilling new ships that enter.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1715,9 +1715,11 @@ void AI::PickUp(Ship &ship, Command &command, const Body &target)
 
 void AI::DoSurveillance(Ship &ship, Command &command) const
 {
-	const shared_ptr<Ship> &target = ship.GetTargetShip();
-	if(target && (!target->IsTargetable() || target->GetSystem() != ship.GetSystem()))
-		ship.SetTargetShip(shared_ptr<Ship>());
+	// Since DoSurveillance is called after target-seeking and firing, if this
+	// ship has a target, that target is guaranteed to be targetable.
+	shared_ptr<Ship> target = ship.GetTargetShip();
+	if(target && (target->GetSystem() != ship.GetSystem() || target->IsEnteringHyperspace()))
+		target.reset();
 	if(target && ship.GetGovernment()->IsEnemy(target->GetGovernment()))
 	{
 		// Automatic aiming and firing already occurred.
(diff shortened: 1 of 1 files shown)


source/AI.cpp
```diff
@@ -1715,9 +1715,11 @@ void AI::PickUp(Ship &ship, Command &command, const Body &target)
...
 void AI::DoSurveillance(Ship &ship, Command &command) const
 {
-	const shared_ptr<Ship> &target = ship.GetTargetShip();
-	if(target && (!target->IsTargetable() || target->GetSystem() != ship.GetSystem()))
-		ship.SetTargetShip(shared_ptr<Ship>());
+	// Since DoSurveillance is called after target-seeking and firing, if this
+	// ship has a target, that target is guaranteed to be targetable.
+	shared_ptr<Ship> target = ship.GetTargetShip();
+	if(target && (target->GetSystem() != ship.GetSystem() || target->IsEnteringHyperspace()))
+		target.reset();
```
endless-sky: You should probably also call `ship.SetTargetShip(shared_ptr<Ship>())` here so the ship clears the bad target instead of holding on to it frame after frame.

tehhowch: Addressed.

Issue #1819, opened 2016-11-12
NPC Afterburner behavior


Amazinite:
When an NPC has an afterburner that does not use the `"afterburner fuel"` attribute, it will use it 100% of the time during combat, regardless of how much energy is consumed or heat is produced. 

So obviously the afterburner mechanic is made on the idea of the `"afterburner fuel"` attribute also being used along side it and the NPC conserving fuel, but this makes it somewhat annoying when creating an afterburner that instead of using fuel consumes large amounts of energy and produces a lot of heat. The NPCs end up either running themselves out of energy to be destroyed or overheating themselves to also be destroyed.

Requested having the NPCs also take energy and heat into account when using their afterburners. (Maybe something like don't use the afterburner if below 25% energy or above 75% heat.)

jafdy:
Wouldn't the energy mess up with perpetual combat designs and one the only have super capacitors for energy storage (to fire projectile weapons).

Amazinite:
What do you mean by that?

jafdy:
If I have 150 energy storage and if I have a gun that uses 125 energy per shot then every time that gun is fired the ship would stop using the afterburner. This would happen even if you had more energy production then you could possibly need.

And what would the AI do with ships without energy storage. You would work out energy proportion with current-energy divided by total-storage. If you have no storage then you would get a divide by 0 error.

KiLEdEnNis:
> then you would get a divide by 0 error.

I can already see escorts disappearing :P

Amazinite:
> I can already see escorts disappearing :P

Actually, the game would crash. 

>  This would happen even if you had more energy production then you could possibly need.

What I requested in OP is just a suggestion. If the 25% energy capacity thing doesn't work, other things can be done. 

> And what would the AI do with ships without energy storage.

Do any NPC ships even exist like that? At that point it's a player-fault kind of thing.

tehhowch:
This is addressed now, courtesy of #3063

Issue #1746, opened 2016-10-22
Hyperspace jump with only an afterburner equipped: arriving in the system pushes you backwards


Bladewood:
Got bored.
Removed my A520 thruster and I was left with a hyperdrive, A865 steering and an ionic afterburner.
Jumped to an adjacent system.
When I arrived in the system (and lost my "hyperspacey" effect), I was immediately blasted backwards by some unknown force.

Hyperspace is not afterburner-friendly, apparently.

Amazinite:
#1170

Bladewood:
rip my personal entertainment.

Issue #1170, opened 2016-05-30
Auto Pilot w/ only Afterburners or Reverse Thrusters


Amazinite:
If you decide to for some reason place only afterburners on your ship and then warp to another system, land on a planet, or really just do anything that requires the auto pilot, you have to manually move yourself into position as the auto pilot does not use the afterburners.

Not sure if this is something worth fixing, but given that we just got the ability to launch with only afterburners it makes sense to have the auto pilot use them.

Edit: and when you do manually slow down to go to another system, you spawn into the system moving backwards...

Edit: It also can't use only the reverse engines when landing and boarding, but it can when jumping to another system.

Gkramme:
Reverse thrusters are used, they are just so weak that it is easier to use the main engines instead.

Amazinite:
I'm talking about when you only have reverse engines installed. The auto pilot won't do anything when attempting to board or land because it doesn't recognize how to use the reverse engines.

endless-sky:
This seems like really low priority: none of the stock ships are configured that way, and if the player chooses to make a ship with no ordinary thrusters, that's their own fault. :)

Amazinite:
Rip my dream of flying around the galaxy in reverse.

KiLEdEnNis:
@Amazinite it would never work out, you don't have guns facing backwards :(

KevinKorzekwa:
It might be worth noting that I only put afterburners on my escort ships (not my flagship) because the AI on escorts actually uses it while the AI for landing/jumping with the flagship does not use the afterburners.  So I can't use them without excessive micromanaging that ends up taking longer than letting the much weaker engines installed just do their thing.

Granted it's a bit different than having only reverse thrusters and afterburners, but it is still afterburner behaviour on the flagship :/

Issue #842, opened 2016-03-08
Content/Balance: (Ionic) Afterburners


LocalGod79:
further to #597, both outfit sprites imply that they could be standalone engines, but the game engine does not allow this (requires a standard thruster to be installed) - i feel like their use would be greater if this were the case (as their provided thrust is often much less than the efficiency gains between larger engine steps).

masterpierround:
Perhaps they could be standalone engines for non-hyperspace ships? That could be the reason you can't install them on your bigger ships: they can't handle hyperspace jumps.

LocalGod79:
why would an afterburner generating 99,000+ thrust be incapable of handling hyperspace, but an X1700 ion thruster generating only 21,000 is?

i guess i'm proposing that either they take up less or no engine capacity (hyperdrive and fuel pods don't - this is a modification to those systems as well, isn't it?), or they should be counted when the check for primary thrusters happens on leaving the outfitter.

endless-sky:
The purpose of the flight check is mostly to keep new players from getting stuck. I suppose I could do something like allowing afterburners in place of normal engines as long as the player has a ramscoop. But, the AI would also have to be updated to recognize that a ship has no thrusters and use the afterburners instead, which is a lot of added complication for something that 99.9% of players will never attempt anyway.

Note that the game *will* allow you to create engines that serve both as a normal thruster and as an afterburner, so you could always do that instead.

LocalGod79:
would you be averse to adding tiny thrust figures to the afterburners so that extra checks/AI programming were not necessary?

PR #3063, opened 2017-09-19
Update Afterburner Logic

tehhowch:
- Support fully automated ship movement for ships with afterburners but no other forward thrust mechanism.
   - If no forward thrusters are installed, the displayed MaxVelocity and Acceleration values are determined by using the afterburner values. If forward thrusters are installed, adding or removing afterburners do not alter their values (since they are no longer used by default).
 - Afterburner heat and energy usage are added to the ShipInfo display table.
 - In addition to fuel usage, NPCs consider the afterburner's energy use and thermal output before deciding to use it (endless-sky#1819), when usage is optional. If usage is not optional (no forward thrust), these checks are bypassed.
 - Consider afterburner usage whenever MoveToAttack is used (ships, asteroids)
 - Consider afterburner usage if picking up flotsam.

The dot product and distance relationships in AI::Pickup were extrapolated from the existing afterburner relationship, at a distance of 1000 and an angle mismatch of ~45deg (dp = 0.9). For each halving of the distance to the flotsam, the allowed angular mismatch is also roughly cut in half, preventing the ship from over-thrusting before it can effectively point at the target flotsam (i.e. don't use the afterburner if using it means you'll miss your target because you weren't pointing at it closely enough).

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1706,12 +1706,9 @@ void AI::PickUp(Ship &ship, Command &command, const Body &target)
 		command |= Command::FORWARD;
 	
 	// Use the afterburner if it will not cause you to miss your target.
-	double distance = p.Length();
+	double squareDistance = p.LengthSquared();
 	if(command.Has(Command::FORWARD) && ShouldUseAfterburner(ship))
-		if(dp > .9999 || (distance > 1000. && dp > .9)
-				|| (distance > 500. && dp > .975)
-				|| (distance > 250. && dp > .995)
-				|| (distance > 100. && dp > .999))
+		if(dp > max(.9, min(.9999, 1. - squareDistance / 10000000.)))
 			command |= Command::AFTERBURNER;
 }
 
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -648,16 +648,6 @@ void AI::Step(const PlayerInfo &player)
 		else
 			MoveEscort(*it, command);
 		
-		// Apply the afterburner if you're in a heated battle and it will not
-		// use up your last jump worth of fuel.
-		if(it->Attributes().Get("afterburner thrust") && target && !target->IsDisabled()
-				&& target->IsTargetable() && target->GetSystem() == it->GetSystem())
-		{
-			double fuel = it->Fuel() * it->Attributes().Get("fuel capacity");
-			if(fuel - it->Attributes().Get("afterburner fuel") >= it->JumpFuel())
-				if(command.Has(Command::FORWARD) && targetDistance < 1000.)
-					command |= Command::AFTERBURNER;
-		}
 		// Your own ships cloak on your command; all others do it when the
 		// AI considers it appropriate.
 		if(!it->IsYours())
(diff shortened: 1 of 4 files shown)


Amazinite:
[Cool](https://youtu.be/KpcmfjFN8OI)

tehhowch:
I'm adding afterburner-only autopilot too, as it is pretty simple.

tehhowch:
Refs
https://github.com/endless-sky/endless-sky/issues/842
https://github.com/endless-sky/endless-sky/issues/1170
https://github.com/endless-sky/endless-sky/issues/1746

As of d33779b, the player can remove all normal thrusters, install only afterburners + steering, and successfully autopilot around, landing on planets, fighting/boarding ships, making hyperspace jumps, just as though normal thrusters were installed. The same goes for player escorts, and NPCs.

AI ships with only afterburner + steering will need a fair bit of ramscoop to not run out of fuel, especially during combat, but that is a task for the content creator.

Amazinite:
I'd certainly have fun with this, not only in testing newly viable outfit setups but also revisiting the afterburner-thruster hybrid engines I was working on when I discovered #1819.

source/AI.cpp
```diff
@@ -1707,8 +1701,44 @@ void AI::PickUp(Ship &ship, Command &command, const Body &target)
...
 		command |= Command::FORWARD;
+	
+	// Use the afterburner if it will not cause you to miss your target.
+	double distance = p.Length();
+	if(command.Has(Command::FORWARD) && ShouldUseAfterburner(ship))
+		if(dp > .9999 || (distance > 1000. && dp > .9)
+				|| (distance > 500. && dp > .975)
+				|| (distance > 250. && dp > .995)
+				|| (distance > 100. && dp > .999))
+			command |= Command::AFTERBURNER;
```
endless-sky: Rather than doing this piecewise, I'd probably use a continuous function: `if(dp > max(.9, min(.9999, 1. - distance * distance / 10000000.)))`  would match these numbers almost exactly.

source/Ship.cpp
```diff
@@ -2075,7 +2076,8 @@ double Ship::TurnRate() const
 
 double Ship::Acceleration() const
 {
-	return attributes.Get("thrust") / Mass();
+	return (attributes.Get("thrust") ? attributes.Get("thrust")
+			: attributes.Get("afterburner thrust")) / Mass();
```
endless-sky: Just because this function gets called a *lot* and attribute lookups by name are time-consuming, it would make sense here and in MaxVelocity() to store the thrust value in a local variable, i.e.:

```c++
double thrust = attributes.Get("thrust");
return (thrust ? thrust : attributes.Get("afterburner thrust")) / Mass();
```

source/AI.cpp
tehhowch: Added

source/Ship.cpp
tehhowch: Added

Amazinite:
Awesome.

PR #1894, opened 2016-12-13
Improving the range calculation (Ref #1826)

Elyssaen:
1. Random lifetime and random velocity are now supported for range calculations (Outfitter Panel, AI firing decisions).
2. The range calculation now uses full, correct submunition velocities.
3. As @Amazinite suggested, I've added an "optimal range" weapon attribute that can be specified which will override these calculations for AI firing decisions only, particularly useful for weapons with high inaccuracy / randomness where the AI would do better with a custom range.

Repost of #1857 from my rebuilt fork, original issue #1826.

@endless-sky, I hope it's okay that I've tried out building a fix for this myself. I'm really enthusiastic about this game and when I had some problems with the existing calculation it was much too tempting to try to figure out how to resolve it. Is this a realistic, reasonable PR? Would it be better if I cut out the "optimal range" part?

PR #3064, opened 2017-09-20
Refactor Mining Logic

tehhowch:
- If the target asteroid is sufficiently out-of-reach, the mining ship selects a new minable rather than embarking on a hopeless chase.
   - either double the engagement range away (1600.)
   - or moving faster than the ship can
 - Mining ships utilize turrets to mine, if they do not have a ship to aim their turrets at.

I updated Autofire to account for the random velocity of each weapon's projectile (which is accounted for in other aim/fire code, but was not here). I did not do any updates for RandomLifetime, as those changes are being addressed in #1894. 

There is a fair bit of duplicated aiming code in AimTurrets now, but I'm not sure of the best way to reduce it, as there are small details specific to each implementation (target asteroid, aim straight ahead, sweep the sky, or aim at the best target ship).

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -516,9 +516,12 @@ void AI::Step(const PlayerInfo &player)
 		else if(isPresent && !target && it->CanBeCarried() && parent && miningTime[&*parent] < 3601 && !isStranded
 				&& parent->GetTargetAsteroid() && parent->GetTargetAsteroid()->Position().Distance(parent->Position()) < 800.)
 		{
-			// Assist your parent in mining its nearby targeted asteroid.
-			MoveToAttack(*it, command, *parent->GetTargetAsteroid());
-			AutoFire(*it, command, *parent->GetTargetAsteroid());
+			// Assist your parent in mining its nearby targeted asteroid
+			// since you do not have the cargo space needed to mine for yourself.
+			const shared_ptr<Minable> &minable = parent->GetTargetAsteroid();
+			it->SetTargetAsteroid(minable);
+			MoveToAttack(*it, command, *minable);
+			AutoFire(*it, command, *minable);
 			it->SetCommands(command);
 			continue;
 		}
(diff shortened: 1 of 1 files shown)


endless-sky:
I'm a bit hesitant to allow turret-based mining. It might make mining way too easy, and it will also make it harder for the AI to pick up flotsam because it no longer guarantees the ship is pointed straight at the asteroid when it explodes.

tehhowch:
The turret based mining will only apply for AI miners, so I'm not sure about the `too easy` qualification. The Hai's Aphid is well-suited to serve as an asteroid miner, with the sole exception that it has only a rear-facing turret. This means it can never attack its targeted asteroid, as the turret faces opposite the direction to the asteroid (and is why I removed the Aphid (Mining) variant from https://github.com/endless-sky/endless-sky/pull/2752 in https://github.com/endless-sky/endless-sky/pull/2752/commits/ba035172985317a4c14fac235ee01ad791e3228b ). It is certainly a little odd to see a ship which the player knows has turrets use only a tiny fraction of its firepower to attack its target. With turret-based mining active, it is the mining fleet creator's responsibility to not use ships with an inordinate amount of high-precision hardpoints in mining operations.

I don't notice ships missing more flotsam than normal - ships with high hull damage output and high speed relative to their turn rate already have some difficulty picking up flotsam, but in general the ships don't target asteroids that are off their path, and the turrets don't autoseek asteroids - they only seek the targeted asteroid.
To limit the impact of aiming with turrets on missed flotsam, I've added relative position checks that limit the aiming of turrets at asteroids to only those in front of the ship, and only while the ship is moving towards the target.

endless-sky:
I'm still hesitant about allowing AI ship turrets to track minables, but the rest of this looks good. My worry is that players will be annoyed that AI ships have an ability that player ships do not.

tehhowch:
I can split that to a separate PR so people can contribute feedback and you can mull it over. Personally, I wouldn't fret about players being annoyed that the AI gets to do more than they can - mining players are generally better at making sure they don't miss flotsam, or ignoring trivial pirate attacks to keep mining (before this PR, players were at least better at not chasing minables they couldn't catch :smile: ).


I think before turret mining for players, people would want a "minables scanner", (query the system to see what kinds of minables exist). What follows after "minables scanner" is a "minables targeter".
An example implementation: 
 - if the player's flagship has an attribute (via an outfit or perhaps a special ship model has a base value), minables within a given range (dependent on the attribute magnitude) get highlighted for a given duration (also dependent on attribute magnitude) when the scanner completes a sweep. 
   - Sweeping would naturally consume energy/power or emit heat as a balancing point. 
   - Sweeping might also trigger pirate raids
 - Minables tagged by the most advanced versions of this kind of outfit could possibly be set as the flagship's target, at least until the tag wears off.

PR #3065, opened 2017-09-20
Independent ships with escorts wait for escorts to be ready to jump

tehhowch:
Rather than jumping immediately when able, a parent ship now waits for the rest of its fleet to be ready to jump before jumping away. Previously, only escorts would wait, and this would only happen if the flagship was less maneuverable than the escorts.
This enforced behavior would enhance the added behavior from #3052 

This PR also fixes a critical bug where an independent carrier that has unfilled bays (i.e. capacity 6 but only 4 carryable escorts) refuses to jump, due to a duplicated "mustWait" check in MoveIndependent.
   - This check was made obsolete in 7936ebc9508787569e365d497af27528e7f20ec5
   - This bug was hard to observe since sole carrier + fighter escort fleets are generally mission NPCs (which means they are escorts themselves and not independent), or are `personality surveillance` (which skips MoveIndependent for non-combat operations), or are not the chosen flagship in their randomly instantiated fleet.


tehhowch:
With 327db04 I moved the escorts-ready-for-jump check to a namespaced function, and reference it in MoveIndependent and MoveEscort. Currently it is rather uncommon for a ship to have non-carried escorts and also have a parent, but with #3052 it would become a common occurrence. Adding the checks to MoveEscort ensure that the mission NPC's ships will move in unison even if their chosen flagship is more maneuverable than the rest of their ships.

tehhowch:
This PR introduced a bug #3243 which is fixed in #3246 : if you hold J while at least two ships from the same (friendly) NPC block mission are in-system with you, only the NPC flagship will prepare to jump, since this PR results its escorts not seeing your flagship's JUMP command.
#3246 adds a check to the "grandparent" (i.e. the player) to resolve it (while also better communicating escort readiness to the NPC flagship).

PR #3070, opened 2017-09-22
Display ships in the map's "orbits" scene

tehhowch:
I developed this as a response to https://github.com/endless-sky/endless-sky/pull/2678#issuecomment-312155956

Summary:
 - This PR adds ship position and heading information to the orbits scene. Ships are drawn as pointers, with the color they would have when viewed in the radar interface or targeted with `R` or `N`. 
Selecting one of these ships displays its name, government, sprite (with hardpoints), shields, and hull, and sets the ship as the player's flagship's targeted ship. If the player has performed an outfit or cargo scan, then the results of the scan are available behind a button. If the player is aware of the ship's description, it is also available with the press of a button (this requires the player to have obtained the ship, boarded it to plunder, or otherwise be able to view the ship in a shipyard).
 - Clicking a ship pointer switches to a new map coloring scheme, where the systems with escorts are colored green, and all others are colored by the system's `System::Danger` level (i.e. the strength of hostile fleets).
 - If you have selected ships previously (e.g. the map was being drawn with the new scheme), closing and reopening the map will - if you know where the ship is - return your view to it. If you do not know where the ship is (e.g. no ships of yours active and in it system, you are shown your current system.
 - Solar Wind and Solar Luminosity are displayed when a system is selected, next to the stars of the system.

Caveats:
1. Given the multitude of ways to open the map, rather than initialize every ___Panel with the ships list, only when the player opens the map in-flight will ships be drawn on the orbits scene. No ships are drawn if the map is opened while landed.
2. Sprites are not drawn rotated, nor are the hardpoints. This is acceptable as-is (to me).

My TODO:
- [x] Prevent description overflow for verbose ship descriptions 
- [x] Improve thumbnail's background UI
- [x] Engine.Ships exports every ship that is spawned (when the map is opened in-flight)
- [x] Fix bug with erroneously recentered map when the flagship's target ship is in now in a system for which no ship data is available
- [x] Address https://github.com/endless-sky/endless-sky/pull/3070#issuecomment-341292180
- [x] Display solar wind and solar power levels for the selected system
- [x] Split "System::Danger" coloring into own PR
- [x] Split "This system has escorts" into own PR
- [ ] Sprite and alignment fixes
<details><summary>The current UI</summary>

At main zoom 100, with the mouse hovering over the Outfits row:
image

</details>
Thanks to @comnom for adapting the art assets from those that MZ originally made.


tehhowch:
With 0e1bf89 all that remains are UI improvements before this is ready to merge.
Aside from a general improvement, there is this issue:
image
@endless-sky any tips or suggestion for dealing with the overflow?

czartrak:
You should somehow compress descriptions or make the text smaller, so longer descriptions do not spill over and block your view. This would be good if it's possible.

Bladewood:
There *is* the solution of *not making such ridiculously long descriptions in the first place...* [which is already used in planet and plugin descriptions.]

tehhowch:
@czartrak Right now the game has only 14pt and 18pt fonts, and above is drawn with 14pt. If a smaller size is available, switching to it would be simple:
```
for(int fontSize = 14; fontSize > 1; --fontSize)
{
	text.SetFont(FontSet::Get(fontSize));
	text.Wrap(fillText);
	if(text.Height() <= MAX_TEXT_HEIGHT)
		break;
}

@Bladewood This description fits just fine if I don't add the 2 lines that tell you what you are looking at. Avoiding long descriptions is good in theory but limiting lore solely because of technical specs is IMO not desired.

jafdy:
Put a scroll bar in.

tehhowch:
@comnom happened to have some time to spare to make a more appropriate box to draw the selected ship inside.
image
 With 463a538 the ship display is now available whenever the player opens the map in flight, as opposed to only when the player *directly* opened the map. That is, the player can open the `i` menu, then the map, and the ship positions are available in the MapDetailView.

tehhowch:
76e4ac0 adds to the displayed information (and further cements the need for a ScrolledText class or subclass of WrappedText) by adding outfit and cargo scan results (in sentence format). It also makes hiding the description possible, by requiring the player to own, have boarded, or be able to purchase the selected ship model on a planet they have visited. If the player does not meet that criteria, the description is not displayed.

tehhowch:
Alright, given the change in to the orbits and key, I'm retargeting this to look OK on the minimum window (1024 x 768), even though that size has overlap in the MissionPanel (Abort & Done). First thought is to place the ship thumbnail in the marqueed area, perhaps with doodads to help tie it to the map key (horizontally) and orbits (diagonally):
image

comnom:
With the changes from https://github.com/endless-sky/endless-sky/commit/746bc212cd40950efa22089f4ef53ebea43a8ad8 there really isn't a nice place to put the ship box element without moving something. 

So what we're going to try to do is: instead of only displaying the ship sprite we display a planet sprite in the box if the player has a planet selected. 

That should be enough justification to have the element always displayed, at which point the new "orbits selection" box can be attached to the bottom of the orbits element where the keys element is now, and the keys element(since it has nothing to do with the orbits) can get pushed down below that.

Ill rework the sprites when I can so we can see what it looks like in practice.

tehhowch:
Without changing the underlying art assets, just approximating positions of things:
1024 x 768
image

1855 x 1056 (windowed full-screen)
image
Per comnom's post, we would shift the map key background down, and enlarge the box used to "contain" the sprite.
I've gone a step beyond the planet art, and if no planet or ship is selected, the system's stars are shown.

Once the updated art is ready, this PR waits on a resolution to the data overflow. I can either use a custom-sized FillShader (and accept that this may draw over a non-trivial portion of the middle of the screen), or try to figure out how to implement a scrollbar  or tabs within the description panel that is shown.

tehhowch:
1024x768:
<details><summary>generic "clicked a new system"</summary>


image
</details>
<details><summary>clicked a ship that has been scanned (or is a playership)</summary>


image
</details>
<details><summary>clicked a landable object</summary>

image
</details>

I'm thinking I might try to turn the ui description panel (where the planet description / ship data is stored) into an `interface`. That'll at least give it flexibility to be moved without source changes, and perhaps some of its hide / clickable functions can be used to consolidate things.

tehhowch:
Alright, this is ready for some more commentary and feedback.
Here's a video of some map interactions: https://youtu.be/jxPfSZYaMHs
Please ignore the flickering - you don't see that during gameplay.

To reconcile the possible overflow from long descriptions, long cargo scan results, and long outfit lists, each is hidden behind a button. Activating the button by either clicking it or pressing the (case sensitive) shortcut (Shift + C, O, or D) opens a dialog box with the known text. 

Screenshots at minimum resolution (1024 x 768) and default ui zoom:
<details><summary>generic "clicked a new system"</summary>

image
</details>
<details><summary>clicked a ship that has been scanned (or is a playership)</summary>

image

</details>
<details><summary>clicked a landable object</summary>

image
</details>
_________________

And on a "real" screen size:
<details><summary>Windowed fullscreen & default ui zoom:</summary>

image
</details>

endless-sky:
To me this interface (especially the ship info) looks way too cluttered. So much of the new information that you're displaying (pictures of planets / stars, systems colored based on whether escorts are present, ship descriptions, etc.) are not particularly useful, and adding information that isn't useful makes it harder for the player to locate the actual useful information. And, the ship's shield / hull status indicator doesn't really fit into the box, adding to the impression that too much is being crammed into too small an interface.

tehhowch:
- The planet / star display behavior could come out, but it definitely looked weird to have the unused UI element there. 
I think a way to view planet sprites without needing to fly to the specific system and then hover around the planet is useful. Ship sprites are easily viewed in the shipyard, but stars and planets have no such existing mechanism. 
   - Perhaps the selected scene defaulting to display the player's current targeted ship / the last selected ship, or the player's flagship (if no ships selected) would make more sense than showing the system's stars though.
   - If stars become more important, there is currently no way to determine what star type is in a system without jumping to that system. While no text is used to describe the star here, extending the interface with a label for the selected scene is trivial.
 - The system coloration based on known ships is pretty worthless at the moment; you are absolutely correct. 
Coloring systems based on known ships, however, allows the player to visually find systems in which they have escorts, rather than forcing them to know about the undocumented and buttonless `Find`, or remember which system is in which sector of the galaxy, to avoid painstakingly panning around the map trying to find where that stranded escort really is.
   - With #2678 in mind, this means selecting escorts and giving them orders to travel to other systems is simpler.
 - Ship descriptions are currently only available to view for ships that can be purchased in a Shipyard. 
With this PR, every ship that has a description can have its lore exposed - and even then, only if the player presses a button to display that description (after having learned it). 

I reused the ui/descriptionpanel sprite for the selected ship's info to help minimize UI shifts when the player is clicking between planets or between ships - if they click nearer to a planet while clicking ships, I didn't want the display to change significantly. The layout of the information could certainly be improved (I've asked around for some inputs but so far have only been told to add even more information like fuel, shield/hull numeric values, etc).

I'll see about using a skinnier ship info background sprite, tucking in the rings, and repositioning the selected ship's information elements.

tehhowch:
The latest set of commits has 
 - downsized the shield/hull rings
 - downsized the displayed ship/planet sprite
 - corrected an issue with multi-star systems' star sprites drawing at large separations
 - swapped to a more appropriate width description panel
 - reordered the contents of said description panel
 - Altered the ship map coloring routine to be much more useful
 - Removed a deprecated fine prevention action

The last bit actually caused me a lot of hidden frustration - When I was developing this (and ships were shown while landed), I couldn't figure out how IsEnemy was returning false for "known" hostiles (which led to using both "IsEnemy()" and "rep < 0" checks). While swapping to use System::Danger in the map coloring scheme, I noticed this issue on a massive scale - all because I'm landed on Greenrock, which resulted in auto-bribing the Pirate government upon game load.

tehhowch:
Given that systems now have solar and wind strengths, how should those best be conveyed? I'm thinking two vertical bar elements, where 100% relative solar/wind strength is in drawn as 50% of the bar height. These bar elements would be in the UI element where the ship sprite is drawn, either both on one side or one on the left and the other on the right. I'd probably need to do a tooltip as well, to avoid needing to put a label on the bar.
(The interface currently doesn't support changing the color of the drawn bar based on the value, so I'm not able to shift the color red/blue based on the deviation from 100%. )

tehhowch:
I have two possible orientations for displaying the solar power and solar wind information:
possible orientations
please use the "Hooray" reaction to vote for the left, and the "Heart" reaction to vote for the right (and :-1: to vote for neither).

KiLEdEnNis:
I'd prefer the right one but with the icons on the left side

tehhowch:
A Discord poll showed strong preference for the vertical orientation, so I've gone with that for now.

Current UI @ ui zoom 100%
<details><summary>Landed, select a system</summary>

landed with star info
</details>
<details><summary>In-flight, no ship or planet selected</summary>

inflight star info
</details>
<details><summary>In-flight, non-player ship selected</summary>

The map is now colored by the results of System::Danger
inflight selected ship
</details>

endless-sky:
I feel that coloring the systems by "danger level" is just going to confuse people, given that there's a different coloring mode that shows friendly vs. hostile systems. And, the UI makes it sound like you know there are "hostiles" there, but you don't. It's also confusing that in all other cases, you click on something on the left to change the map coloring, and a pointer appears on that item in the sidebar. So, if you want to be able to sort by danger level, it should be included in the sidebar. And, it should probably use the standard blue to yellow gradient rather than having a sharp transition between the color for perfectly safe and almost perfectly safe.

It might make more sense for the presence of escorts to be indicated in some other way rather than coloring the systems.

The ship info panels seems like far more information than is necessary, and in any case it would be better for the information to be laid out in table form (like everywhere else in the game) rather than having each field at a different X position.

It's confusing for the header above the orbits panel to show the name of the system or planet that is selected but not the name of a ship if it is selected.

The "light level" icon has an aliasing problem - the vertical lines are one pixel wide but the horizontal ones are spread over two pixels, making them look dimmer and blurrier.

It seems a bit weird to use the actual sprite (instead of an outline) for the selected ship, because everywhere else target ships are drawn in outline mode. On the other hand, it wouldn't work to use outlines for the planets and stars, so I'm not sure what the right approach is there.

Using the "status overlay" colors for the ship status is likely to confuse players who have never turned that option on.

I'm also not convinced that the tracking of known descriptions, outfit lists, and cargo is worth the added complexity. It is also likely to confuse players that they can't learn a ship's description by seeing it in a shipyard.

In general, I feel like the only piece of information that will be helpful to nearly all players would be having some indication of where their escorts are. Being able to view solar power and ramscoop levels for each system will help players who have those outfits, and coloring systems by danger level may be useful too (although it's hard to explain how the player knows the danger level so precisely). Everything else seems like information overload.

tehhowch:
For danger level, perhaps a categorical classification would be better.
 - the player can only sense a general level of danger rather than detailed comparison between systems
 - would decrease the usefulness to a modder that is playtesting event fleet changes
 - would serve well as the appropriate left-hand side labeling / pointer

I think the full ship sprite is OK here on the map. In flight, it's important to see as much of the surroundings as possible - you wouldn't want your cockpit arbitrarily covering things up in a fight. In the ship info panel, there isn't much color in use anyway - using the full sprite doesn't fit, so the muted swizzle 8 is used. However, the map is full of colors - a black swatch or a ship outline would look weirder than the actual, full sprite. Consider that the player is not actively piloting - they are at a terminal, reviewing plans and records - there isn't time-sensitive information behind the ship sprite being lost.
 - I chose the overlay colors so that the player can easily determine that this ship is hostile or friendly, and it looks basically like it does in flight. For the in-flight HUD, this hostility is shown by the outline color, but with the full sprite that isn't viable.

For cargo/outfit scans, the player has no way to track what they've scanned other than rote memorization of the ship name, and what was on it. Not having a record of scan logs seems really terrible for 30th-century shipboard computers, so I really feel that this is a necessary addition.

I'm willing to drop ship descriptions, but having them here adds a currently absent feature (view all ships' descriptions).
 - I'll make sure the descriptions are known for any visible Shipyard ship, not just purchaseable ones.

I haven't tried to embed a button in a Table (or really used the Table class much at all), but if that can't be done I think only the name and government stuff would work in a table and the buttons will need to stay.

Escort presence could be shown all the time, perhaps by drawing a double-radius ring around the relevant systems. Thoughts on this?
 - For color I might go for just white / grey to avoid color changes due to player swizzle updates.
 - Would require shifting mission pointers too, or drawing the "escort ring" underneath them

I'll work on these items and other the feedback you mention, over the next week or so. I think danger level/escort presence is deserving of a separate PR, so I'll split those changes out from this PR.

Amazinite:

Issue #3034, opened 2017-09-11
Remnant Keystones mission only taking flagship's keystone if escorts are out of system


Hadron1776:
If I buy 50 Quantum Keystones to my cargo hold for the Remnant keystones mission and turn it in normally, the keystones get removed from cargo. _However_, if the keystones are bought into an escort's cargo hold, and that escort is outside the system when you turn in the mission, you will be paid the 6 million credits, but only the keystones from the ships that are in the system will be taken.

1. Have at least one ship (other than your flagship) with space to hold 50 Quantum Keystones
2. Install Quantum Keystones on both ships
3. Accept Remnant Keystones mission
4. Buy 50 Quantum Keystones to cargo. The keystones should go to the escort's cargo hold. 
5. Have your escort hold position just outside Arculus
6. Turn in the mission

Expected result: I get paid 6 million credits and 50 Quantum Keystones are taken from my cargo hold

The result I got: I still got paid 6 million credits, but only the keystone that was installed on my flagship got taken away.

tehhowch:
While #3036 will solve this issue, it will not also fix problems with boarding or assisting missions which gift outfits.  #3005 's latest commit (a6b157322be0e959d50b571a5a70b351a98c0599) will both fix this and fix outfit gifting by boarding/assisting missions.

endless-sky:
Fixed by merging #3036; #3005 is a much bigger change so I'll have to look at that separately.

PR #3075, opened 2017-09-25
enhancement: Visual feedback for outfit transfer missions while in-flight

tehhowch:
With 3bfa5eee6e0a4942857e240a4614e1f2c1975bb2 players lost the ability to observe in-flight if any current missions they had that use outfit gifting were able to be completed.

This PR changes the display state of the mission in the MissionPanel view based on whether or not the player has the required outfits to perform the exchange.

For example, if the player takes a mission to mine 100 tons of Aluminum and lands in-between harvests, some of the haul is on an escort ship (potentially multiple). The only way to determine if this mission can be completed is to manually toggle through all of the player's ships and keep a running total of stored Aluminum. With this PR, the game will only color the mission in white when the player has 100 Aluminum  in ships that are present in-system.

The guards added to the cargo check should prevent #3034 from occurring again, while making it simple for the player to check if they have harvested enough of a given mineral and can now land.

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -574,10 +574,7 @@ bool Mission::CanComplete(const PlayerInfo &player) const
 	if(!toComplete.Test(player.Conditions()))
 		return false;
 	
-	if(!IsSatisfied(player))
-		return false;
-	
-	return true;
+	return IsSatisfied(player);
 }
 
 
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -568,16 +568,34 @@ bool Mission::HasSpace(const PlayerInfo &player) const
 
 bool Mission::CanComplete(const PlayerInfo &player) const
 {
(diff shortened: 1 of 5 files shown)


source/Mission.cpp
```diff
@@ -568,16 +568,34 @@ bool Mission::HasSpace(const PlayerInfo &player) const
...
+	if(player.GetPlanet() != destination)
 		return false;
 	
 	if(!toComplete.Test(player.Conditions()))
 		return false;
 	
+	if(!IsSatisfied(player))
+		return false;
+	
+	return true;
```
endless-sky: It might be simpler and more readable to just `return IsSatisfied(player);` instead of the above four lines, just to avoid the "double negative" of "if not satisfied return false..."

tehhowch: Addressed

PR #2449, opened 2017-04-26
map : shipyard attributes ; jobs : mining jobs and a few more large bulk delivery jobs in human space

EndrosG:
majors
+ infile documentation of currently used attributes, including some comments on their usage
+ shipyard attributes to all relevant human, coalition, hai, wanderer, as well as Kessel Sepret, Sapira Mereti, Sestor Ikfar - i.e. no shipyards so far for Kor Efreti, Quarg and Pug --- part of #2359

+ added various mining jobs, for human space only (so far) --- part of #2101
+ added large bulk and large rush deliveries to major shipyards, for human space only (so far)

minors
+ attributes hai to planet Darkwaste (did not have any attributes yet)
+ attributes pug to planets Pugglemug and Pugglequat (did not have any attributes yet, and did never receive any by events, should happen in "pug territory liberated")
+ korath attr. to "Korati Efreti"
+ urban attr. to Hai-home and Stonebreak
+ shipyard attributes in Geminus related events (war begins, geminus rebuilt)
- put moon / station as 2nd attribute after the region attribute
- reordered the blocks in the job file a little bit (moved large bulk and large rush deliveries upwards)
+ added two additional empty lines between the blocks in the job file (i.e. 3 empty lines now between blocks)
+ in hai space, I changed "manufacturing" to "factory" on 4 planets (Dustmaker, Heartvalley, Makerplace, and Stonebreak), because Icelake used "factory" and `hai jobs.txt` uses "factory", while "manufacturing" is never used.

notes
- checked all event files, actually checked all data files in master repository
- I did not change a few things which seemed odd to me:
  - in coalition space, it is "wealthy" instead of "rich"
  - planet Hope is not volcanic

Issue #2101, opened 2017-02-03
Feature Request: Salvage and mining missions


Tadrix:
Mining missions are self explanatory: Suitable enough for beginners. Perhaps some kind of tractor beam would be nice? Missions could be given by mining planets.

Salvage missions: A mission to retrieve a wrecked vessel. Payment is 50% of its stock cost, but the wreckage is always in a dangerous place (pirate system, empty system with large amount of pirates). It could even require to retake without destroying a certain ship from the pirates.

KiLEdEnNis:
tractor beam? Korath Grab-Strike & Heliarch Repulsor, humans are probably not advanced enough for a tractor beam

salvage: it'd just be a disabled ship that you repair, what do you mean by wrecked? someone is working on "destroyed" ship sprites but they're just made to be destroyed to get the damaged outfits to sell, adding; 1. how would a tractor beam get a ship through hyperspace? 2. "towing" has been brought up before 3. how would you only shoot the tractor beam on the ship? why not just board them and repair them?
mining: why would someone pay you for getting them ore when you can sell them yourself as a side job?

Tadrix:
I'll answer.
1. It wouldn't. Tractor beam is for asteroid chunks, not for ships.
2. For ships one would need some kind of tug (fighters?). For example, 1 for smaller ships (drone, fighter, interceptor, light warship, light freighter), 2 for medium warships, 5 for heavy warships and freighters. And they would only bring the disabled ship to the flagship for boarding, not do something else (for players which prefer their flagship to be a big-ass mother with weak engines).
3. I believe 1. and 2. answered this one.

Mining: Same as rush deliveries. They need certain amount of certain ore to meet their quota on time. They are willing to pay premium to get them, lest they'll have to pay bigger penalty for missing a delivery.

jafdy:
Mining can already be done. No one has got round to doing it yet.

Salvage, repair a ship and escort it back. There is a mission like that in the FWEscort storyline.

MessyMix:
@Tadrix I'm currently working on flavor missions for the galaxy. Hostage ship / Salvage missions in dangerous parts are included.

KiLEdEnNis:
asteroids behave similarly to ships, being able to pull asteroids but not ships would make no sense & would probably be hard to code, for hardly any results: you can destroy the asteroids anyway

[...]

PR #3076, opened 2017-09-25
Add Mining jobs

tehhowch:
Closes #2449 
Refs #2101 

Given the latest mission/job toolset in https://github.com/endless-sky/endless-sky/commit/11a30e0efc309d67f18d5e7212c81eac89fce8cd, I've updated @EndrosG 's PR to use the automatic shipyard attributes.
I've expanded upon her work and also clarified the file diffs, to simply the pre-merge review needed. To prevent the player from starting a bunch of high-quantity or rush jobs without knowing where to find some of the needed outfits, the player has to complete smaller jobs for a given minable first (which will place a marker for those minables in the player's map file when viewing the map->Outfitter and selecting the minable).

Summary:
 - Mining jobs for
    - Human
    - Hai
    - Coalition
    - Wanderer
 - Shipyard-based Delivery missions

While working on this, I came across a deprecated function declaration, and also determined that any missions that completed based on the ability to gift outfits do not display this dependence visually. Those two issues are rectified in #3075

### DIFF
--- a/data/jobs.txt
+++ b/data/jobs.txt
@@ -2180,7 +2180,7 @@ mission "Large Bulk Delivery [0]"
 	description "Deliver <cargo> to <destination>. Payment is <payment>."
 	cargo random 40 2 .02
 	to offer
-		random < 70
+		random < 65
 		"cargo space" > 160
 	source
 		government "Republic" "Free Worlds" "Syndicate" "Neutral"
@@ -2199,7 +2199,7 @@ mission "Large Bulk Delivery [1]"
 	description "Deliver <cargo> to <destination>. Payment is <payment>."
 	cargo random 40 2 .02
 	to offer
-		random < 60
+		random < 55
 		"cargo space" > 160
 	source
 		government "Republic" "Free Worlds" "Syndicate" "Neutral"
@@ -2218,7 +2218,7 @@ mission "Large Bulk Delivery [2]"
 	description "Deliver <cargo> to <destination>. Payment is <payment>."
 	cargo random 40 2 .02
 	to offer
-		random < 50
+		random < 45
 		"cargo space" > 160
 	source
 		attributes "mining" "textiles" "factory" "farming" "fishing" "oil"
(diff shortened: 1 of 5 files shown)


tehhowch:
Given the multi-author nature of this PR, I feel it should be merged 'normally' and not via squash-merge, as that would inappropriately label me as the original author of some of EndroG's work.

endless-sky:
I'm a bit worried that adding so many more "large bulk" and "large rush" delivery missions will make them swamp the job list. (And, limiting them to planets with shipyards will concentrate the destinations on a few planets, making those missions even more lucrative compared to others.)

Also, all the formatting changes make this hard to review. I know the existing code isn't totally consistent about this, but I often try to use quotation marks for free-form text to distinguish it from keywords, even if the free-form text is one word (e.g. "Republic"). That's why the government names in the jobs are in quotes.

tehhowch:
I can revert that token quotation change and decimate the `random` condition on the new non-mining jobs. I think @EndrosG was looking to make use of the new keywords she added.

I agree with the bit about the oops-too-lucrative nature of specialized missions which is why I had reduced the mining job rates, and limited their concurrency somewhat.

tehhowch:
"Replace All" was a bit greedy, but normalizes the Escort & Bounty jobs to quote single-word government names in their location filters, along with the rest of the generic jobs.

Pointedstick:
IMHO more flavor jobs is always a huge gain. If there are too many jobs in the list, we should merge this and then just adjust the frequencies down for all the jobs.

tehhowch:
This will also be the first set of missions that use complex/"arbitrary" conditions, so it can't merge until #3128 or an equivalent merges (lest we flood users' consoles with "Unrecognized condition expression" 😁 ).

endless-sky:
I'm still not too enthusiastic about this. First, because mining is enough of a niche activity that it doesn't seem like there need to be special missions for it. And second, because players who do mine will likely prefer to just fill up their holds with the most valuable materials they can mine rather than having to be limited to searching out a particular material.

tehhowch:
I'll add the mining missions into a plugin i develop around mining, drawing other changes for it from #2752

Issue #3083, opened 2017-09-27
Wormhole travel sends the flagship to the wrong coordinates


tehhowch:
The wormhole's position advances by one day when the player changes position, but the flagship's position is set by the previous spot of the wormhole.

To reproduce, turn a moon of a planet near the system center into a wormhole and enter the system via it.

tehhowch:
The finnicky bit is that you can't currently write
`ship->Position() = object->Position()`
since the Position information is returned as a constant value - e.g. read-only:
https://github.com/endless-sky/endless-sky/blob/a5a5cfec95eb2623167a8e703533aba45cf2fd12/source/Body.cpp#L134-L138

Existing methods which modify a ship's position call a member function of that class (e.g. `Ship->Place( newPos ....)`), within which the privately kept variable (only the Ship class can manipulate it) is changed directly (rather than accessed publicly via `Position()`). Unfortunately, Ship->Place() does a host of additional things that are not desired in this use case:
https://github.com/endless-sky/endless-sky/blob/a5a5cfec95eb2623167a8e703533aba45cf2fd12/source/Ship.cpp#L539-L565

This suggests to me that the only way to fix this bug is to either:
1) add a StellarObject method that returns its position on the next day, 
gate the wormhole position update on if this ship is the player flagship
if so use the TomorrowPosition() method
else use the existing Position() method
2) add a Body method `SetPosition` and call this from Engine::EnterSystem()
3) add a boolean argument to Ship::Place that defaults to performing the extra bits, but allows not performing them, then call this method from Engine::EnterSystem()  

I'm choosing method 3.

endless-sky:
Fixed by merging #3085.

PR #3085, opened 2017-09-27
Use a wormhole's actual position instead of the previous day's position

tehhowch:
Refs #3083  

While no wormhole-like objects in the base game move far enough each day that the flagship's position is noticeably incorrect, mods do exist and the game might eventually have something like this.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -819,9 +819,8 @@ void Engine::SelectGroup(int group, bool hasShift, bool hasControl)
 void Engine::EnterSystem()
 {
 	ai.Clean();
-	grudge.clear();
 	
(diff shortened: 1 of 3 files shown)


endless-sky:
To me it would make more sense to add a Ship::SetPosition() function that only modifies the ship's position, rather than modifying the Place() function.

tehhowch:
Fair enough. I liked the modification to Place as it easily allows for future scenarios where a ship's velocity and/or angle should also be changed. Granted, those future scenarios all require event code that doesn't exist yet so it's a moot point.

tehhowch:
Addressed the comments in e7ce120

Issue #2889, opened 2017-08-04
Some way to convey a player's fleet attraction to them


Amazinite:
Either by showing the player's fleet cargo and DPS in their info panel or just straight up showing their attraction, there should be something in game to convey to the player that their fleet is attracting pirate raids, and some sort of information on what exactly pirate raids are and how they appear.

gunqqer:
It could just be me being stupid but that title doesn't really convay what you said in the body.

tehhowch:
Rather than hard numbers for the chance to spawn a raid fleet, we probably
want text categories based on the value... "Unlikely .... Possible...
Likely.... Very likely... Near-certain"

Hadron1776:
> Rather than hard numbers for the chance to spawn a raid fleet, we probably
want text categories based on the value... "Unlikely .... Possible...
Likely.... Very likely... Near-certain"

Just a nitpick: "Unlikely" coming before "possible"?

Bladewood:
How about this, then:
> Nigh Impossible -> Unlikely -> Likely -> Very Likely -> Near Certain

gunqqer:
Nigh Impossible -> Very Unlikely -> Unlikely -> Possible -> Likely -> Very Likely -> Near Certain

-> You're screwed, go home, don't try it.

tehhowch:
>Just a nitpick: "Unlikely" coming before "possible"?
We can just drop possible :)

endless-sky:
Your fleet's attractiveness to pirates is already indicated via a user interface when pirate ships shoot things at you if they find you attractive.

If I did add a fleet rating, it'd be more interesting if it was along the liens of a combat rating, something like:
* not worth the risk
* slim pickings
* juicy target
* easy prey
* pirate bonanza

gunqqer:
as always MZ comes up with much better stuff than we do.

Amazinite:
>Your fleet's attractiveness to pirates is already indicated via a user interface when pirate ships shoot things at you if they find you attractive.

The only reason I ask for something like this is because there are a couple of people on Steam (who I told this to) asking for some sort of reasoning given to the player as to why they are attracting so many pirates and something telling them how they could fix it. i.e. making the idea of pirate raid attraction visible in text instead of just in the fact that you spawn pirate raids.

marphod:
Juicy target before easy prey?  I'd assume the other order, but it is ambiguous.

[...]

Issue #2223, opened 2017-03-05
Overhaul freighter raider system


Tadrix:
Searched for it in issues, didn't find. (I'm notoriously bad at searching.) So I'm making a new issue.

There are alternatives to avoid freighter raiders. Any Berserker or Bounder can be disarmed and modded to be better at hauling cargo than a Star Barge — yet, they will not attract pirate raiders. So, instead of exponential Star Barge problem we can have exponential Bounder problem. Or cargo Bactrian problem.

Also, if there are many freighters, but even more heavy warships guarding them, then suicidal pirates don't make much sense.

My proposition is to replace fixation on amount of freighters to evaluating fleet's total cargo space to total firepower (number of weapons and total dps capability). Then no matter what ship category, better guarded convoys will be left alone in preference to poorly guarded ones, which is what real pirates would do, and which would resolve these balance problems nicely.

Amazinite:
This was talked about a while back, but never on GitHub. Good change. 👍

jafdy:
I see a problem at the start where you fly an unarmed ship around but this can be fixed by doing (maxcargo-abit)/firepower.

Tadrix:
@jafdy As an example, this system could be turned on if any of the following are true:
1. Player has more than 7 ships (incl. flagship);
2. Player's total cargo capacity exceeds 760 t;
3. Player's net worth is at least 25M.
So, the player could have relatively untroubled start of the game, only to make things interesting later.

Amazinite:
`(<fleet cargo>/(<fleet firepower> + 1))/1000 = raid spawn chance`?
This would make it so that an unarmed Star Barge has a (50/1)/10000 = 0.5% chance of spawning a raid.

Tadrix:
Number of guns seems to me as poor choice. Then savvy players would outfit their ships with ordinary blasters.
I'd say total firepower is fleets total dps vs. shields.

MessyMix:
I agree with this, and this is something I supported when it was first brought up.

But just to nitpick with you :P
>Any Berserker or Bounder can be disarmed and modded to be better at hauling cargo than a Star Barge

hmmmm...... not sure about that one ;)

Tadrix:
Challenge accepted.

Reductio ad absurdum proof — remove anything, put a fuel cell and smallest engines available, fill rest with cargo expansions.

[...]

PR #3087, opened 2017-09-28
Convey raid fleet spawn likelihood via Player Info menu

tehhowch:
Refs #2223 #2889 

 - Updates the attraction calculation to require that ships are not destroyed and have ammunition for any weapons that require them. The IsDestroyed check means that a fleet that takes heavy losses to its freighters will have an easier time escaping from pirates, while a fleet that takes heavy losses to its escorts is nearly guaranteed to take heavy damage to its freighters if more jumps are needed to reach safety. I considered adding checks for in-system ships, but this opens up more ways to game the calculation.
 - Adds a section underneath the player's "combat rating" in which a categorical label and an estimate of the number of raid fleets that will spawn are displayed:
image
 - If a raid fleet is spawned, the player is notified that their fleet attracted the raiders' interest.

10 categorical labels were chosen and roughly ordered to indicate how the player is regarded by the raid fleet. The respective probability numbers indicate the chance that the stated number of raid fleets are instantiated during Engine::EnterSystem for the given `raw attraction` value.

raid threat level | raw attraction | Prb 0 | Prb >= 1 | Prb >= 2 | Prb >= 5 | Prb >= 9 | Avg
-- | -- | -- | -- | -- | -- | -- | --
"not worth the risk" | 2 | 100.00% | 0.0% | 0.0% | 0.0% | 0.0% | 0
“chopped liver” | 3 | 90.44% | 9.6% | 0.4% | 0.0% | 0.0% | 0.1
“slim pickings” | 4 | 85.97% | 14.0% | 0.9% | 0.0% | 0.0% | 0.15
“small potatoes” | 6 | 77.63% | 22.4% | 2.5% | 0.0% | 0.0% | 0.25
“fresh meat” | 9 | 66.48% | 33.5% | 5.8% | 0.0% | 0.0% | 0.4
"juicy prey" | 15 | 48.40% | 51.6% | 15.2% | 0.0% | 0.0% | 0.7
"easy target" | 27 | 24.84% | 75.2% | 38.0% | 0.5% | 0.0% | 1.3
“sitting duck” | 51 | 5.63% | 94.4% | 75.6% | 7.8% | 0.0% | 2.5
“fish in a barrel” | 100 | 0.11% | 99.9% | 98.8% | 61.1% | 1.0% | 4.95
"pirate bonanza" | 197 | 0.00% | 100.0% | 100.0% | 100.0% | 98.4% | 9.8

I'm open to changing the text labels and menu header, if better suggestions are offered. I started with those given in #2889


nylki:
I like the general idea! However some of the labels seem a bit too cryptic or slangy for my taste. As a non-native speaker I am having issues especially with “chopped liver” and “fish in a barrel”.

tehhowch:
I admit, the phrases are generally English idioms.
Like the combat ratings that appear above it, the text is not meant to be absolutely indicative of the order - the number associated with the value provides a much clearer indication of the threat imposed.

ghoulavenger:
I hope this gets in just to see how bad my fleet really is. I have over 100 deep rivers, 52 hurricanes, get huge pirate attack fleets every time I'm in human space. I suspect I'm at pirate bonanza level.

endless-sky:
I feel like the "raid threat level" label is somewhat unclear, although I'm having a hard time thinking up a better alternative. I also think it would help to show negative threat ratings, so the player can tell that adding more warships makes them even less attractive.

How about changing the heading to "piracy threat factors," and then splitting it up into two line items, one for your fleet strength and one for your cargo space? For example:

| **piracy threat factors** | &nbsp;
| ---- | ----:
| cargo: minimal | (0)
| fleet: entirely defenseless | (0)

| **piracy threat factors** | &nbsp;
| ---- | ----:
| cargo: abundant | (4)
| fleet: heavily armed | (-6)

Cargo ratings could be something like this:
* minimal
* slim pickings
* fair amount
* profitable
* abundant
* bountiful
* quite valuable
* tons of plunder
* copious wealth
* untold riches

Strength ratings could be something like:
* entirely defenseless
* minimal weaponry
* puny defenses
* somewhat vulnerable
* well protected
* reasonably dangerous
* heavily armed
* serious threat
* deadly armament
* unassailable

Since your ratings are on an exponential curve anyway, it might make sense to just use an equation for them. Something like:

```c++
rating = (attraction <= 2. ? 0. : log(attraction - 1.) - 1.);
```

tehhowch:
That's a great idea

jafdy:
It strikes me that abundant and bountiful sound like a whole load more stuff than "quite valuable".

tehhowch:
One issue is that the difference between the two parameters is the important bit, so even if both values are in the same tier, the player can end up with an attraction > 200 (e.g. 10 raidfleets every time):
image

I used log scaling of the attraction value to select the category in https://github.com/endless-sky/endless-sky/pull/3087/commits/486eb74964679faad221d2e5d2bef375c73cef56 because the step from 2 to 3 attraction is more important than that from 3 to 4, and both are much more important than the step from 100 to 101 in terms of risk to the player.

I can resolve this loss of information by
 -  placing the current value from the OP into the "piracy threat factors" header, and using the row rating method (as in the combat ratings, valued 0 to size-1 of the label array).
image
This keeps the ordinal ranking of attraction and deterrence, and displays the average number of raid fleets spawned. However, it puts data in the header row
 - printing prettified attraction & deterrence values (raw, or transformed by log/sqrt to "obfuscate") instead of the row ratings (raw values here:)
image
Using obfuscation on the raw value kinda returns us to the first bit where the difference between two large numbers is much more important than the difference between two small values, (e.g. with a log2 scale, +10.5 & -10.2 spawns 10 fleets every time while 5.3 and -5.0 spawns only .3 on average). Not using obfuscation means you can get large numbers (above), but the difference is always of the same importance.
 - do both (data in header row and use "obfuscated" factor values).
image
this could use an easy-to-understand (e.g. not flavor-text) categorical filter instead of the `(2.05)`: 
none: 0
low: < 1 on average
medium: < 2 on average
high: < 5
insane: >= 5

I'm leaning towards the last one, with a categorical filter, and obfuscating via square root rather than log2

(fleet used for reference is 500 argosies with blasters swapped to meteor launchers and 1 cargo expansion, and a flagship with -38 attraction by itself, total cargo space ~67500)

endless-sky:
I'm worried that the "2.05" number will be meaningless and confusing to people. I'd say that should either be converted to a number between 0.0 and 10.0 representing the expected number of raid fleets, or a number between 0% and 100% representing the likelihood of having at least one raid.

Since the raid mechanics are of the most concern to a player in the mid-game (an attractiveness between 2 and 20, say), I think that showing the likelihood of at least one raid makes the most sense.

Along with that, perhaps the fleet probabilities should be tweaked. Right now, as soon as you go above 3 you go from a 0% chance of raids to a (1 - .99^10) = 10% chance of at least one raid. I think the probability should be changed to:

```c++
double attraction = (sum - 2.) / 200.; // Or maybe (sum - 3) to be kind.
if(attraction > 0.)
{
    for(int i = 0; i < 10; ++i)
        if(Random::Real() < attraction)
            raidFleet->Place(*system, ships);
}
```

The probabilities would rise at the same rate, but would ease in a bit more gradually:

attraction

The attractiveness ratings for the two sub-components could be shown with one decimal place. That would make it clear, for example, that if your cargo is +9.3 attractive and your weapons are -9.2 attractive, you should still expect pirates. That would be preferable to using a square root for the number and a log for the text.

tehhowch:
As of 756ddea:
- Stock Starbarge:
image

endless-sky:
Does anyone else have an opinion on whether the attractiveness numbers should be shown raw or scaled in some way?

As for the header row, the more I think about it the more I'm convinced that showing the "risk of pirate raids" (0 - 100%) will be more comprehensible than showing "expected number of pirate raids" (0 - 10).

tehhowch:
I updated the post above with more images of some stock ships' ratings

Were you thinking raw probability or scale the probability into the categorical labels?

endless-sky:
I guess it's a tradeoff between giving the player more information vs. avoiding breaking immersion. That is, there's no way in-game for the player to know their fleet is going to attract pirates exactly 14% of the time, but showing them that number is much more useful than just telling them "risk of pirate raids: moderate." I'm leaning in the direction of actually showing the risk percent, but I could be persuaded otherwise.

tehhowch:
Well, plenty of time before the next release to collect feedback and convert the percentage into a scale if desired :+1: 
d835917 drops the categorical label for a flat %
stock bulk freighter:
image

ghoulavenger:
I'm more in favor of a raw numbers approach, obfuscated or not. I don't need to know the odds of the pirates attacking, just how well defended my fleet is should I decide to travel in human space. The numbers, at least at present are nowhere near representative of my fleets ability to defend itself, since with 112 deep rivers, and 52 hurricanes I get constantly attacked by massive pirate fleets in human space but the only time trouble happens is if my flagship gets ahead of the rest of my fleet.

KiLEdEnNis:
I have no problem with having both in, just with the placement of the "15%" in Tehh's last picture
I think it would look better if it was more like

piracy threat___________________
cargo: fair amount           (+2.8)
fleet: minimal weaponry   (-1.6)
[risk factor]                         15%

Issue #3086, opened 2017-09-28
[Feature Request] Automatically replenish extra crew members on landing.


starscollide5:
Having to do it manually every time one lands after capture becomes infuriating experience after a few games, even with the help of keyboard shortcuts it's somewhat annoying. 

I propose to add an optional setting that will trigger effortless automatic extra crew member replenishing, the same way required crew members are replenished in the current release.

tehhowch:
Easy enough to do: requires setting up a private variable and a setter method on the PlayerInfo class (`desiredCrew`, `SetDesiredCrew`), whenever the flagship changes it gets zero'd, and in the relevant Land method `if(desiredCrew)` then `flagship->AddCrew( desiredCrew - flagship->Crew() )` 
Lastly, alter the HiringPanel's methods to call the setter method with the new total crew amount.

I'm sure an accessibility argument can be made for it as well.

jafdy:
This is going to come under "too much automation", I can feel it.
It would be nice though.

Amazinite:
Duplicate of #1504
Please close this issue.

jafdy:
This is a little more focused than 1504, that was asking for special crew to do this and other things as well where this could just be a button that says "refill to this many crew". MZ also responded in a general way about not wanting to detract from the experience and it is easy to argue this will enhance the experience.

tehhowch:
The solution I propose above would not be a button or really visible at all to the player. At most, I would use the Messages function to report `"You hired < added > crew to replace those lost during boarding operations."`, and only for excess crew additions (e.g. not the normal, automatic minimum crew restoration).

endlesssteve:
To reduce to a lower or normal number of crew would you them manually fire the ones you didn't want anymore?
Is this like cruise control?

tehhowch:
If done as mentioned above, yes. Boarding other ships is an incline that "slows you down". Changing your flagship is like braking (disables the functionality). Hiring and firing is using the Accel / Coast to fine tune the setpoint.

tehhowch:
Implemented my earlier comment in #3089. 
Doesn't quite "brake" when changing flagships per se, because the PlayerInfo's `flagship` pointer (which tells the game which ship in the list of ships the player owns is the flagship) is invalidated upon landing (since it can change a fair bit while landed, unlike while in-flight).

If this gets tagged as `unlikely` or `wontfix` I'll be closing it though.

starscollide5:
[...]

PR #3089, opened 2017-09-28
Store the flagship's desired crew level and try to recrew when landing

tehhowch:
Refs #3086 

If the player changes the crew level on their flagship, restore it if landing on a planet with a spaceport and available services.
Displays a message alerting the player that this has been done.
If the player changes the flagship while landed, the old flagship's crew is reset to the minimum in TakeOff so the player will not pay phantom salaries.

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -27,6 +27,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Outfit.h"
 #include "Person.h"
 #include "Planet.h"
+#include "Preferences.h"
 #include "Politics.h"
 #include "Random.h"
 #include "SavedGame.h"
@@ -1054,7 +1055,7 @@ void PlayerInfo::Land(UI *ui)
 	}
 	
 	// Hire crew for the flagship if the user changed it from the default.
-	if(hasSpaceport && flagshipCrew)
+	if(Preferences::Has("Rehire flagship crew on landing") && hasSpaceport && flagshipCrew)
 	{
 		int addedCrew = flagshipCrew - flagship->Crew();
 		if(addedCrew > 0)
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -291,8 +291,6 @@ void PlayerInfo::Load(const string &path)
 				break;
 			}
 	}
-	// Set the rehire crew target.
-	flagshipCrew = Flagship() ? Flagship()->Crew() : 0;
 	
 	// If no depreciation record was loaded, every item in the player's fleet
 	// will count as non-depreciated.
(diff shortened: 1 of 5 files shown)


tehhowch:
Defaults off, changed via preference in the preferences file

source/PlayerInfo.cpp
```diff
@@ -1043,6 +1054,22 @@ void PlayerInfo::Land(UI *ui)
 			ui->Push(new Dialog(message));
 	}
 	
+	// Hire crew for the flagship if the user changed it from the default.
+	if(Preferences::Has("Rehire flagship crew on landing") && hasSpaceport && flagshipCrew)
```
endless-sky: Maybe "Rehire extra crew when lost" would make more sense? (Just so it's clear that this is only talking about *extra* crew, and that normal crew are going to be replenished no matter what.)

source/Preferences.cpp
```diff
@@ -41,7 +41,8 @@ namespace {
 void Preferences::Load()
 {
 	// These settings should be on by default. There is no need to specify
-	// values for settings that are off by default.
+	// values for settings that are off by default unless they are not
+	// shown in PreferencesPanel.
```
endless-sky: I don't see any reason not to add this to the preferences panel, since there's plenty of space (in the "Other" section).

tehhowch: Should those be alphabetized?

source/PlayerInfo.cpp
```diff
@@ -1043,6 +1054,22 @@ void PlayerInfo::Land(UI *ui)
...
+	{
+		int addedCrew = flagshipCrew - flagship->Crew();
+		if(addedCrew > 0)
+		{
+			flagship->AddCrew(addedCrew);
+			Messages::Add("You hire " + Format::Number(addedCrew) + " extra crew member"
+					+ (addedCrew == 1 ? " to replace the one " : "s to replace the ones ")
+					+ "lost during boarding operations.");
+		}
+		// A new flagship was chosen with a higher crew requirement.
```
endless-sky: This would mean that if the player switches to a flagship with lower crew requirements, they will automatically hire extra crew even if they didn't intend to.

I think a better approach would be to save, not the desired flagship crew, but the desired number of extra crew members.

tehhowch: Won't that have the same issue?
Since the rehire action happens on landing, performing a check during PlayerInfo::TakeOff is perhaps the better solution to the player swapping from a capping ship to their "daily flier"

tehhowch:
With 7ff791e the crew level is set during PlayerInfo::TakeOff()
This ensures that the crew rehiring target is always relative to the player's selected flagship (as long as the player is unable to change flagships during flight, i.e. no escape pods), 
 -> if the player switches from a "capping" ship to a "daily flier," they will not hire up to its max crew on their next landing.

The caveat of this method is that if the player flies with only a single ship and hires to max crew, any mission passengers will reduce the rehired crew amount to the used amount. After completing the mission, the rehired number will not return to the flagship's maximum crew amount: it will stay at what the missions reduced it to. 
 -> If a player has other ships that can ferry the passengers, then this will not apply as the mission passengers will be added to those ships rather than the flagship (as expected).

PR #1971, opened 2017-01-03
Fighters now rearm on the carrier when launching

zwparchman:
This allows fighters to rearm missiles from the carriers cargo hold. This is one of the changes being considered for #1939.

The only real rough edge I've seen is that landing on a planet without an outfitter does not rearm your fighters, but I haven't tested much.

PR #3090, opened 2017-09-29
Prevent players from switching landing targets if they are already landing on a planet

Hadron1776:
In the `master` branch, I have noticed that if there is more than one planet in a system, you can land on one planet, then just before the planet panel pops up, you can click on the other planet. This makes your ship (which is _really_ small by now) fly off toward the second planet you clicked on - but it brings up the first planet's panel. 

This PR prevents this by not allowing other planets to be targeted (besides the current one) if the player is already in the process of landing.


tehhowch:
Strange... d61cb6a doesn't show any author or committer info for all contributors to the pull. For a small PR like this it's fine, but I think for something like #1971 or #3076 where two different contributors add a lot, a non-squash merge may be better for conveying authorship.

Issue #3093, opened 2017-09-30
Drag-selecting escorts when being offered an in-flight conversation


Amazinite:
Moments before attempting to drag-select an escort, I assisted a Hai ship and was offered the job for helping a Hai ship. After exiting the conversation, the selection box that had been created before entering the conversation stuck on my screen until I attempted to drag-select escorts again.
![ei1nzbn5t0giy9o3ky6nyq](https://user-images.githubusercontent.com/17688683/31041084-cf4b86cc-a55d-11e7-8b54-62cae779853f.png)

tehhowch:
Fixed in #3094

PR #3094, opened 2017-10-02
Pop-ups cancel drag actions

tehhowch:
- If a conversation, dialog, or other panel is generated while the player is dragging, stop drawing the drag box and reset the drag status.

Refs #3093

### DIFF
--- a/source/MainPanel.cpp
+++ b/source/MainPanel.cpp
@@ -151,17 +151,22 @@ void MainPanel::Draw()
 	
 	if(isDragging)
 	{
-		Color color(.2, 1., 0., 0.);
-		LineShader::Draw(dragSource, Point(dragSource.X(), dragPoint.Y()), .8, color);
-		LineShader::Draw(Point(dragSource.X(), dragPoint.Y()), dragPoint, .8, color);
-		LineShader::Draw(dragPoint, Point(dragPoint.X(), dragSource.Y()), .8, color);
-		LineShader::Draw(Point(dragPoint.X(), dragSource.Y()), dragSource, .8, color);
+		if(canDrag)
+		{
+			static const Color dragColor = *GameData::Colors().Get("drag select");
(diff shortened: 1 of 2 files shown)


source/MainPanel.cpp
```diff
@@ -151,17 +151,22 @@ void MainPanel::Draw()
...
+		}
+		else
+			isDragging = false;
 	}
 	
 	if(Preferences::Has("Show CPU / GPU load"))
 	{
 		string loadString = to_string(lround(load * 100.)) + "% GPU";
-		Color color = *GameData::Colors().Get("medium");
+		static const Color color = *GameData::Colors().Get("medium");
```
endless-sky: Should this maybe be a `const Color &` rather than a `static const`? That's what I use in most other places where I'm getting a reference to one of the colors defined in the interface.

source/MainPanel.cpp
```diff
@@ -151,17 +151,22 @@ void MainPanel::Draw()
...
 	if(isDragging)
 	{
-		Color color(.2, 1., 0., 0.);
-		LineShader::Draw(dragSource, Point(dragSource.X(), dragPoint.Y()), .8, color);
-		LineShader::Draw(Point(dragSource.X(), dragPoint.Y()), dragPoint, .8, color);
-		LineShader::Draw(dragPoint, Point(dragPoint.X(), dragSource.Y()), .8, color);
-		LineShader::Draw(Point(dragPoint.X(), dragSource.Y()), dragSource, .8, color);
+		if(canDrag)
+		{
+			static const Color dragColor = *GameData::Colors().Get("drag select");
```
endless-sky: Same thing here - maybe used `const Color &` instead.

In Escape Velocity, if you switched to an alien ship as your flagship it could completely alter the heads-up-display to look like an alien interface. You can't do that in ES yet, but it's worth writing the code in such a way that color and interface definitions could change while the game is running.

source/MainPanel.cpp
tehhowch: Addressed.

PR #3097, opened 2017-10-03
Also show minimap while preparing to jump.

flaviojs:
I can see when all jumps are done (minimap disappears).
It is easier to cancel a wrong jump (minimap is visible longer).

I also wanted to show the travel plan in the minimap but that is better done after some refactoring, which I'm not comfortable in doing right now.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -288,8 +288,7 @@ void Engine::Step(bool isActive)
 			doEnter = false;
 			events.emplace_back(flagship, flagship, ShipEvent::JUMP);
 		}
-		if(flagship->IsEnteringHyperspace() || flagship->Commands().Has(Command::JUMP)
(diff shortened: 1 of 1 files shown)


source/Engine.cpp
```diff
@@ -288,7 +288,7 @@ void Engine::Step(bool isActive)
 			doEnter = false;
 			events.emplace_back(flagship, flagship, ShipEvent::JUMP);
 		}
-		if(flagship->IsEnteringHyperspace()
+		if(flagship->IsEnteringHyperspace() || flagship->Commands().Has(Command::JUMP)
 				|| (flagship->Commands().Has(Command::WAIT) && !flagship->IsHyperspacing()))
```
endless-sky: This line can be taken out if you're adding the previous one, because Command::WAIT is never set in situations where Command::JUMP is not.

PR #3102, opened 2017-10-07
Enable turrets to aim at targeted asteroids

tehhowch:
Currently the player cannot target turrets at their asteroid of choice, and neither can AI miners.
This makes certain ships (especially turret-heavy ones) very poor choices for the game's mining fleets.

This commit would allow ships to aim their turrets at asteroids, and - if the player is ever given a method to target minables - also the player.

Please share feedback on this feature, per https://github.com/endless-sky/endless-sky/pull/3064#issuecomment-334801849

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -540,6 +540,8 @@ void AI::Step(const PlayerInfo &player)
 			if(it->CanBeCarried() && parent && miningTime[&*parent] < 3601 && minable
 					&& minable->Position().Distance(parent->Position()) < 600.)
 			{
+				// Set the target asteroid (to enable turret tracking)
+				it->SetTargetAsteroid(minable);
 				MoveToAttack(*it, command, *minable);
 				AutoFire(*it, command, *minable);
 				it->SetCommands(command);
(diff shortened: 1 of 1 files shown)


Rob59er:
Doesn't bother me that only the AI could do it because a player can sort of do it with a little skill, but I'm just one voice of many.

tehhowch:
My opinion is (unsurprisingly) that this wouldn't bother me, because I can respond to any minable on-screen, rather than only those in a set cone in front of me.

Lineth:
Wouldn't bother me either. The player can intuit good ways to mine even with turrets. The NPC miners would need it more. But it might raise issues with some players.

Amazinite:
Personally I wouldn't mind, but if the AI can mine with turrets then I see no reason as to why the player would be unable to.

Bladewood:
I'll just post this proverb:
> If the AI can do it, player fights would be ruined
If the player can do it, the experience would not be the same
Just leave the lone developer to his job and everything will turn out okay.

endless-sky:
I found a much cleaner way to do this (3899f38): just add the target asteroid to the "enemies" list.

tehhowch:
Awesome awesome awesome :+1:

PR #3112, opened 2017-10-11
Clarify "Landing on:" navigation mode.

flaviojs:
It can display "Landing on", "Can land", and "Cannot land".
It adds "(mission)" when it is the destination of a mission.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -483,17 +483,9 @@ void Engine::Step(bool isActive)
 	if(flagship && flagship->GetTargetStellar() && !isJumping)
 	{
 		const StellarObject *object = flagship->GetTargetStellar();
-		string navigationMode = flagship->Commands().Has(Command::LAND) ? "Landing on" :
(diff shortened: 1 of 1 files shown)


endless-sky:
I'm not sure about the "(mission") thing. It'd make more sense if it said something like "(mission target)", although I know that wouldn't fit.

flaviojs:
Hmm, then how about "(target)" instead?

source/Engine.cpp
```diff
@@ -483,7 +483,10 @@ void Engine::Step(bool isActive)
 	if(flagship && flagship->GetTargetStellar() && !isJumping)
 	{
 		const StellarObject *object = flagship->GetTargetStellar();
```
tehhowch: this could be `const Planet *planet = flagship->GetTargetStellar()->GetPlanet();` since this block is conditioned on the stellar object existing.
Then you'd be able to drop the extra `object->GetPlanet()` check below.

PR #3116, opened 2017-10-12
Extend reference-checking in GameData::CheckReferences

tehhowch:
- Add Name() to Minable & Phrase & GameEvent
 - Have MissionActions store GameEvent pointers rather than strings.
 - Maintains the current storage structure in MissionAction->fail, but generates a reference to the string in GameData so that its existence (or lack thereof) can be determined.

These changes will catch uses of systems as mission destinations (fixed the one I found with it :smile: ), planets as waypoints, systems as stopovers, variants defined in fleets that don't exist, phrases for governments that don't exist, etc.


source/Minable.cpp
```diff
@@ -33,6 +33,9 @@ using namespace std;
 // Load a definition of a minable object.
 void Minable::Load(const DataNode &node)
 {
+	// Set the name of this minable, so we know it has been loaded.
+	name = node.Size() >= 2 ? node.Token(1) : "Unnamed Minable";
```
endless-sky: The "Unnamed Minable" bit might not be needed, since minables always have a name. (Instead you could just say, `if(node.Size() >= 2) name = node.Token(1)` to deal with any future case where that's no longer guaranteed.)

source/MissionAction.cpp
```diff
@@ -164,10 +164,18 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			int maxDays = (child.Size() >= 4 ? child.Value(3) : minDays);
 			if(maxDays < minDays)
 				swap(minDays, maxDays);
-			events[child.Token(1)] = make_pair(minDays, maxDays);
+			string eventName = child.Token(1);
+			events[eventName] = make_pair(minDays, maxDays);
+			// Create a GameData reference to this mission name.
+			GameData::Events().Get(eventName);
```
endless-sky: A cleaner option might just be to add a Name() function to GameEvent. Then, that could be used to check for unloaded events *and* the events map here could use GameEvent pointers instead of strings.

source/Minable.cpp
tehhowch: I had drawn this from Mission::Load, where it gives "Unnamed Mission" if no name is given. But now that I think about how one would go about `using` minables, it is immediately apparent that minables require non-empty names if they are to be used by a system.

PR #3117, opened 2017-10-12
Short set of missions to unlock the Bactrian

Turtleroku:
I feel that to so easily acquire the Bactrian without any investment is robbing the player of possible value. The effort required to get the ship would be roughly a month of time, but not so much that it turns into a grind or achievement medal that everyone gets if they play long enough.
I modified the Bactrian to be less jack of all trades. The Hepatica and Crataegus make up for crew or cargo niches that the Bactrian previously filled.

Hopefully this is a valuable contribution. 

(I will make a blender of the Hepatica if this looks like the short story is going to be actualized, or I have spare time)
(also I have to remove the Bactrian from all fleets prior to completing the mission)

### DIFF
--- a/data/Bactrian_Search.txt
+++ b/data/Bactrian_Search.txt
@@ -796,34 +796,34 @@ mission "bactrian search 3a-1"
 			`	As you land the ship, you step out into the evergreen forest that covers the mountain. You can hear squaks and chittering of the forest as you make your way to the wreckage.`
 			`	On your way to the wreckage you see two medium sized predators, some sort of cross between a porcupine and a tiger, as well as few flocks of iridescent blue birds. The first crack in the hull you find looks to be somewhere near the engine room as one of the enormous exhaust nozzels is laying just off to your right.`
 			choice
-				`	Everyone stay in a group.`
-					goto slowsearch
-				`	Head off towards the bridge while your crew searches the rest of the ship.`
-					goto funsearch
-			label slowsearch
-			`	The ship had been throughly scavenged as you find only a few data hexes throughout the ship. Eventually you reach the bridge.`
-				goto slowsearch2
-			label slowsearch2
-			`	You are waiting for the rest of the search party to arrive when suddenly Leroy Jenkins says "Screw it" and runs onto the bridge. Out of the corner of your eye a large spike tiger flies into Leroy knocking him to the ground and taking a big bite out of this shoulder). immediately everyone starts opening fire and kills the spike tiger.`
-			`	Three of the crew attempt to stabilize Leroy and take him back to the ship. You search the bridge for any data hexes that might have been left over.`
-				goto foundit
-			label funsearch
+				`	Search the engines first.`
+					goto engines
+				`	Head off towards the bridge.`
+					goto bridge
+			label engines
+			`	The ship had been throughly scavenged as you find only a few data hexes throughout the ship.`
+				goto funsearch2
+			label bridge
 			`	After climbing all other the hull you find a hatch. The ship has long since run out of power but you are able to pry the hatch open.`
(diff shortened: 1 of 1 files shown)


tehhowch:
- There is still a fair amount of missing punctuation and grammatical errors.
 - Some of your `choice` nodes are missing gotos or actions - picking the conversation option that sounds like it should `decline` follows the same path as the affirmative option.
   - the missions otherwise are parsed without error by the game
 - I feel the mission text, especially how the player views or interacts with the characters, makes a lot of assumptions about the player. It's not common for this game to break the 4th wall.

Turtleroku:
Speaking of 4th wall, are you referring to line 590 as an example?
Can you give me an example of what made you say that the mission text makes lots of assumptions? What line so I can think about how to rewrite it?

tehhowch:
FWIW, you don't need to delete the entire file - `git` is capable of identifying what changed (e.g. just upload the file again)
4th wall:
 - https://github.com/endless-sky/endless-sky/pull/3117/files#diff-4bff46f7296e227401fd7ad16837fd4dR69
 - https://github.com/endless-sky/endless-sky/pull/3117/files#diff-4bff46f7296e227401fd7ad16837fd4dR695

data/Bactrian_Search.txt
```diff
@@ -0,0 +1,1326 @@
...
+		ship "Osprey" "Beard of Imsety" tag nocapture
+		ship "Osprey" "Cackle of Duamutef" tag nocapture
+		ship "Osprey" "Shriek of Hapi" tag nocapture
+		ship "Osprey" "Gaze of Qebehsenuef" tag nocapture
+		ship "Marauder Falcon" "Hound of Anubis" tag nocapture
+		ship "Marauder Falcon" "Song of Hathor" tag nocapture
+		dialog "Now you should be able to look through the base"
+	on offer
+		conversation
+			`	On your landing approach a pirate fleet launches from the surface. Leroy your co-pilot swears something about being ducks in a puddle while you change course to deal with the pirates.`
```
Turtleroku: Would it sound less like a 4th wall breach like this:
`	Suddenly, a pirate fleet launches from the surface. Leroy your co-pilot swears something about being ducks in a puddle and changes course to deal with the pirates.`

Turtleroku:
I was trying to add more character to the warning message. Would it be better if I had a fail message more like the FW campaign?
`on fail
		dialog `You have failed an essential Bactrian mission. If you want to acquire the Bactrian, revert to the autosave or another earlier snapshot of the game.`

data/Bactrian_Search.txt
KiLEdEnNis: (apart from a big Bactrian mission string PR that's been worked on for very long and that it's way more likely to be merged which makes working on this basically redundant)
remove that second line entirely and rewrite it to something without mentioning a co-pilot, ducks, or a puddle

jafdy: The player might have a ship that only takes one person to fly.

KiLEdEnNis: hence the "without mentioning a co-pilot", but good thing ya specified the reason

Turtleroku: Since this only occurs after an event requiring disabling a modified argosy, it would be highly unlikely (read impossible) that anyone would be in a single crew ship.

KiLEdEnNis: capturing MMK or using any other 1-man ship, and have big-ass escorts

jafdy: Never make any unneeded assumptions for the player. Ever.

Turtleroku: Should I remove the "bactrian search: 2b special" mission also so as to not assume pirate relationship?
I am truly wanting an honest answer.

Turtleroku:
I made a flow chart of the mission sequence.
[flow chart of missions.pptx](https://github.com/endless-sky/endless-sky/files/1384690/flow.chart.of.missions.pptx)

PR #3122, opened 2017-10-13
Add a combat rating check to "Hunted by State" missions.

jjhankins:
Under certain circumstances, or when using mods, it's possible to get a very negative rep while still too weak to handle the bounty hunters.

### DIFF
--- a/data/pirate jobs.txt
+++ b/data/pirate jobs.txt
@@ -2475,6 +2475,7 @@ mission "Hunted by State [0]"
 		government "Republic" "Syndicate" "Free Worlds" "Neutral" "Pirate"
 	to offer
 		random < 10
+		"combat rating" > 1100
 		or
 			not "chosen sides"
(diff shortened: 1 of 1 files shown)


data/pirate jobs.txt
```diff
@@ -2475,6 +2475,7 @@ mission "Hunted by State [0]"
 		government "Republic" "Syndicate" "Free Worlds" "Neutral" "Pirate"
 	to offer
 		random < 10
+		"combat rating" > 1100
```
tehhowch: These are pretty high - this is a combat rating (in Player Info) of 7.
Based on the deep jobs file, for which the same fleet is spawned with a rating of only 35, a much lower combat rating can be used.
perhaps 150 (CR 5) for this, and 1100 for the other.
Alternately, these missions could require the player to have performed certain quantities of various pirate jobs - that would help solve any unintended conflicts with mods.

PR #2990, opened 2017-09-01
Mission NPCs can be given individual travel instructions

tehhowch:
Kudos to @Elyssaen for the inspiration and collaborative work on this feature.

This PR introduces 4 keywords for the NPC node and 1 new NPC success criteria:
Keywords:
 - `waypoint`
 - `patrol`
 - `land`
 - `visit`

Succeed criteria:
 - `land`
______________

The `waypoint` and `patrol` keywords assign a system or set of systems to the ships in the NPC block. These systems will be traveled to, if they are accessible, in the order they are specified (though LocationFilter systems are created after all non-LocationFilter waypoints). If a system is inaccessible, it is removed in-flight. If no value is specified, the mission destination system is used. If `patrol` is used, passing no value creates a patrol between the mission's `origin` and `destination` systems. Both keywords can utilize LocationFilters to randomly select the waypoints. If the `distance` argument for LocationFilters is used, it is sequential, e.g. the distance is measured relative to the previous waypoint. All methods can be used together to specify the NPC's waypoints. 
Upon reaching the final system in the list, the NPCs will either behave as normal escorts would, or if `patrol` was used, begin traveling again with the first specified waypoint. The `staying` personality takes effect when the NPC no longer has waypoints to visit.
<details><summary>Usage:</summary>

```
waypoint/patrol (no value given)

waypoint/patrol <system 1> ... <system N>

waypoint <system 1> <system 2>
waypoint ...
waypoint <system N>

waypoint/patrol
    LocationFilter 1, line 1
    ...
    LocationFilter 1, line N
waypoint/patrol
    LocationFilter 2, line 1
    ...
    LocationFilter 2, line N
...
```
</details>
______________

The `land` and `visit` keywords assign a planet or a set of planets to the ships in the NPC block. These planets will be landed on, if they are accessible, if the NPC enters the system that contains the planet. If the keyword was `land`, the ship will not take off again from the planet - it has landed permanently. If the keyword was `visit`, the ship will not force itself to land on that planet again unless it also has `patrol`, and has finished its current loop.
As with the waypoint/patrol keywords, passing no value results in the mission's destination being used as the target. LocationFilters can similarly be used to randomly select the landing targets.
<details><summary>Usage:</summary>

```
land/visit (no value given)

land/visit <planet 1> ... <planet N>

land/visit <planet 1> <planet 2>
land/visit ...
land/visit <planet N>

land/visit
    LocationFilter 1, line 1
    ...
    LocationFilter 1, line N
land/visit
[...]

PR #3123, opened 2017-10-14
Unify refueling and routing code

tehhowch:
- Extends advanced refueling logic to ships moving independently & following orders
 - Extends advanced refueling logic to in-system escorts
 - Chooses the best route for a ship to reach its destination system

@flaviojs please check this pull for differences in escort behavior compared to your pull - I'm pretty sure it keeps all the behaviors you added while extending them to MoveIndependent (where they can serve `personality uninterested` and #2990 ).

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -150,14 +150,17 @@ namespace {
 		}
 		return false;
 	}
-	// Wrapper for ship-system uses.
+	// Wrapper for ship - target system uses.
 	bool ShouldRefuel(const Ship &ship, const System *to)
 	{
 		if(!to || ship.Fuel() == 1. || !ship.GetSystem()->HasFuelFor(ship))
 			return false;
+		double fuelCapacity = ship.Attributes().Get("fuel capacity");
+		if(!fuelCapacity)
+			return false;
 		double needed = ship.JumpFuel(to);
 		if(needed && to->HasFuelFor(ship))
-			return ship.Fuel() * ship.Attributes().Get("fuel capacity") < needed;
+			return ship.Fuel() * fuelCapacity < needed;
 		else
 		{
 			// If no direct jump route, or the target system has no
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -125,31 +125,33 @@ namespace {
 	}
 	
 	// Determine if the ship with the given travel plan should refuel in
-	// its current system, or if it should keep travelling.
-	bool ShouldRefuel(const Ship &ship, const DistanceMap &route)
+	// its current system, or if it should keep traveling.
+	bool ShouldRefuel(const Ship &ship, const DistanceMap &route, double fuelCapacity = 0.)
 	{
-		double fuelCapacity = ship.Attributes().Get("fuel capacity");
-		const bool hasFuel = fuelCapacity && ship.JumpFuel();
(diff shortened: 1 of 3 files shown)


source/AI.cpp
```diff
@@ -103,6 +103,109 @@ namespace {
...
+		return false;
+	}
+	// Wrapper for ship-system uses.
+	bool ShouldRefuel(const Ship &ship, const System *to)
+	{
+		if(!to || ship.Fuel() == 1. || !ship.GetSystem()->HasFuelFor(ship))
+			return false;
+		double needed = ship.JumpFuel(to);
+		if(needed && to->HasFuelFor(ship))
+			return ship.Fuel() * ship.Attributes().Get("fuel capacity") < needed;
```
flaviojs: When there is no fuel capacity, it tells the ship to refuel.

source/AI.cpp
```diff
@@ -1105,13 +1213,13 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
 		vector<int> systemWeights;
 		int totalWeight = 0;
 		const set<const System *> &links = ship.Attributes().Get("jump drive")
-			? ship.GetSystem()->Neighbors() : ship.GetSystem()->Links();
+			? origin->Neighbors() : origin->Links();
```
flaviojs: Hmm, looks like ships with a jump drive and a hyperdrive never go through links that are too far to be neighbors.

tehhowch: Neighbors() includes all Links()

flaviojs: Hmm, then ships that only have a jump drive can travel farther that the jump distance when a long hyperdrive link is available.

tehhowch: That is explicitly intended behavior

flaviojs: Ok.

source/AI.cpp
tehhowch: Should be addressed now.

flaviojs:
There is a case that can be improved further. When my escort is jumping with me, it could consider the next refuel stop of my travel plan.

I found this while playing as a merchant with the initial Star Barge as escort and the travel plan Wazn -> Alhena (fuel) -> Menkalinan (no fuel) -> Capella (restricted) -> Procyon (no fuel) -> Sirius.
It get's stuck in Capella and then Procyon. If it had refueled in Alhena it would only be stuck in Procyon.

tehhowch:
@flaviojs that happens on both `master` and with this PR?

Also, if you know the changes you'd make, I'm happy to merge them in. Just open the PR to my branch

flaviojs:
After rechecking, the example behavior is for ```master```.

In this PR it goes refuel when it doesn't really need to, which isn't that bad.
With a Mule as escort and the travel plan Epsilon Leonis -> Gomeisa (fuel) -> Zosma (fuel) -> Algieba (no fuel) -> Tania Australis (fuel), it made an unnecessary refuel at Zosma.

I'll go see what would need to change.

flaviojs:
I made a PR to your branch tehhowch/endless-sky#27.

A better approach would be to make a new class TravelPlan that supported multiple destinations and filled in the gaps with DistanceMap as needed. It would be able to handle link changes and unexpected system changes (by landing on a wormhole) with the same "fill in the gaps" approach.

Issue #1138, opened 2016-05-24
[bug] ships that set off their self-destruct don't count as killed (until landing)


LocalGod79:
for mission completion purposes. (just had a two out of three Marauders blow up under me, and the mission is incomplete)

edit; both ships registered as destroyed upon landing, giving me the 'npc kill' dialog, and then immediately the 'on complete' payment dialog.

Amazinite:
Wasnt this fixed in a commit for v0.9.1?

LocalGod79:
yep - as i said in the edit, the completion dialog came up upon landing, rather than the (self-)destruction of the final Marauder bounty.

it's perhaps the behaviour is less "not present", and more "not as expected".

LocalGod79:
yes, definitely "not as expected"...

if i land at the planet where the Marauder Bounty job was taken immediately after having had said ships s/d beneath me, the npc kill dialog appears *after* the government gratefully pays me one...

endless-sky:
Fixed in dbdc9f9.

PR #3136, opened 2017-10-18
Self-destructing NPCs should be marked as Destroyed

tehhowch:
Refs #1138 and https://github.com/endless-sky/endless-sky/commit/dbdc9f94768e12b000fbe2f415435770777227b0


endless-sky:
Addressed in f509198, since further refactoring made it so I couldn't merge this.

tehhowch:
I'm not sure if you meant to send the Destroy event for every ship that lands or is out-of-system long enough to be tagged for removal too. Pull #3142 addresses that.

Issue #3137, opened 2017-10-18
planet "Pug Wormhole" is referred to, but never defined


flaviojs:
When I run endless-sky from the console I get:
```Warning: planet "Pug Wormhole" is referred to, but never defined.```

Wrzlprnft:
#404

tehhowch:
I don't get this message on my testing save, but I do get it on a new save that hasn't yet done the story.

Amazinite:
It's because the wormhole exists in the Pug galaxy but nowhere else., so @Wrzlprnft, there's no way for the system to actually tell that it's a wormhole until it's actually linked to Deneb.

tehhowch:
We could add a pug wormhole planet entry, and then a suitable cutoff for link distance that prevents drawing wormholes if their links are more than, say, 2 galaxys long

tehhowch:
| Link | Distance |
|--|--:|
| Aescolanus - Cardea | 240.9 |
| Cardea - Terminus | 1746 |
| Cardea - Insitor | 142.5 |
| Coluber - Convector | 550.4 |
| Deneb - OtR | 27869 |
| Edusa - Peragenor | 421.2 |
| Insitor - Segesta | 395.1 |
| Sabriset - Sko-karak | 1529.8 |
| Segesta - Aescolanus | 303.5 |
| Ultima Thule - Waypoint | 764.7 |

I think the longest in-galaxy wormhole would be `Lone Cloud` to `Ka'pru`, at ~2180. So, 2500 would be a reasonable value (or smaller, if it is desired to not display some of these longer, in-galaxy wormholes.

tehhowch:
The reason the warning doesn't appear once it is linked is because:
https://github.com/endless-sky/endless-sky/blob/ef0a6043001928244214fcf89e51c745c4c2c7e8/source/Planet.cpp#L196-L202
(e.g. when it exists only in `Over the Rainbow`, it doesn't meet the criteria for a wormhole - more than one system - so its name is empty. Once it is a wormhole, the name check returns ??? so it will pass the reference check even without a `planet` definition.

tehhowch:
#3138 updates the check for planets - an undefined planet has no name and no system (e.g. typos in mission definitions or game events)
This keeps the ability to create a wormhole with no `planet` definition in order to keep it undrawn (e.g. #404)

Wrzlprnft:
i could now say that my statement included tehhowch's solution, but i disn't think of that.

[...]

PR #2601, opened 2017-06-05
Warn of invalid travel plans pre-takeoff, pulse invalid travel routes

tehhowch:
This PR introduces a warning upon attempting to take off from a planet when the player's selected "Travel Plan" or "Travel Destination" is no longer possible (much like if you don't have enough cargo space to leave without selling outfits in cargo). If the player continues launching, the travel plan / destination is cleared as appropriate.

The player can view any invalid portion of their travel plan in the standard map view that shows your intended route. The invalid segments are drawn [alternating between transparent and their normal color](https://youtu.be/0WSBg4FfSKk) to indicate that these segments are of special interest to the player. If the chosen planet in the final system to land upon is no longer accessible, the final system of the plan will be drawn with a red ring around it that also pulses.

This PR prevents the player from
1) Having a `player.TravelDestination` that corresponds to an inaccessible planet (which has caused application crashes in the past)
2) Having a travel plan that routes through restricted wormholes without the access key
3) Having a travel plan that needs a jump-capable drive, but only having a link-following drive

(It also makes sure the player is notified of any carried ships that will be sold even if a mission failure would also happen, though this can/should get cherry-picked to `master` anyway.)

PR #3138, opened 2017-10-18
Add "Pug Wormhole" planet definition

tehhowch:
- Wormhole "planets" are given a system, but unless also defined with a `planet` entry, would not pass GameData::CheckReferences until the wormhole was linked to another system

per #404 wormholes with no `planet` definition are not drawn on the map, so until #2601 merges, the link to the pug galaxy will be drawn (if one revisits it, as it will then be marked as visited).
refs #3137

### DIFF
--- a/data/pug.txt
+++ b/data/pug.txt
@@ -650,6 +650,8 @@ system "World's End"
 		period 360
 		offset 290
 
+planet "Pug Wormhole"
+
 planet Ruin
 	attributes uninhabited
 	landscape land/hills3
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -227,7 +227,7 @@ void GameData::CheckReferences()
 		if(it.second.Name().empty())
 			Files::LogError("Warning: phrase \"" + it.first + "\" is referred to, but never defined.");
 	for(const auto &it : planets)
-		if(it.second.Name().empty() && !it.second.GetSystem())
+		if(it.second.Name().empty())
 			Files::LogError("Warning: planet \"" + it.first + "\" is referred to, but never defined.");
 	for(const auto &it : ships)
 		if(it.second.ModelName().empty())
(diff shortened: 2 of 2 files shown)


endless-sky:
It seems likely that in the future there will be wormholes that are not created by events but that should not show their path on the map, so I'd prefer a more flexible solution. What I was thinking is just to say:

* If it has a description, it should show on the map. (Rather than, if it has a planet definition.)
* If it has a spaceport, NPCs should be able to land on it. (Current behavior.)

That would also allow wormholes that show no path but that NPCs can travel. The only downside would be that wormholes in plugins would stop showing up on the map until the plugins were updated, but once updated they would be backward-compatible.

tehhowch:
#2601  touches up the wormhole drawing considerably. I can add that behavior to it

jafdy:
That is quite a small downside, the wormholes still work and either way it's a minor enough change that I don't think anyone will really mind.

tehhowch:
@jafdy the change mentioned by MZ gives a lot more control over AI behavior with wormholes than there is currently. All a plugin writer will need to do extra is `description "."` if they want it to draw, rather than get forced to have it drawn if they want NPCs to land on it.

tehhowch:
@endless-sky should I take your comment to mean that the game shall warn about wormholes that do not have a planet definition, e.g. revert the proposed change here and add a planet definition for the Pug Wormhole?

endless-sky:
Yes, that's what I was thinking - requiring even wormholes to have a planet description is a lot simpler than having convoluted logic in the undefined object check to try to ignore wormholes.

endless-sky:
Note that as with "The Eye" does with the spaceport in wanderers.txt, just putting `description ""` would be sufficient to make the description non-empty, because newlines are appended in Planet after every description paragraph.

tehhowch:
Addressed

Issue #3139, opened 2017-10-19
Offering ship category condition for assisting/boarding missions


Amazinite:
Missions offered by assisting or boarding ships are currently only present in the form of the assisting missions that decline on offer. The amount of support for actually creating good assisting/boarding missions is lacking, #1663 and #2188 being examples of missing/broken features for such missions.

While brainstorming for possible assisting/boarding missions to create, an idea popped into my head: the ability to have assisting and boarding missions offer based off of the category of the ship you are assisting/boarding. For example, a boarding mission could exist in which a small pirate ship (interceptor or light warship) is used as a decoy, and upon boarding a large pirate fleet is triggered. On the opposite end of things, larger pirate ships (medium or heavy warships) may be carrying important cargo such as treasure maps or a large amount of credits which a smaller ship would likely not be carrying.

This, in addition to the system attributes mentioned in #526 and the above two mentioned issues, would allow for an interesting number of assisting/boarding missions to be created, and possibly even add variety to the current jobs by increasing the payout by the category of ship you saved (as the pilot of a Bulk Freighter will likely have more credits on hand to thank you with than that of a Star Barge).

tehhowch:
#3005 should resolve #1663 and #2188 (one of the test missions that prove it does involves triggering hostile fleets when you board one of their faction :+1: )

It sounds like you are looking to extend the `source` tag so it could be something like
```
source
    government "Pirate"
    class "Heavy Warship" "Medium Warship"
on offer
    payment -100000
    conversation
        `As soon as you dock with the <origin>, your life support systems fail. Even though you pay the ship's captain <payment>, he refuses to deactivate the virus.`
            die
```
sounds good to me :+1: 


(regarding payment scale, you can do that simply, by using a different mission, once you have the source filter enhancement)
What you'll probably also want is to be able to set characteristics of the npc ship on which the mission spawns... but that's for a different issue.

tehhowch:
#3140 should address this issue nicely, by adding support to filter on the ship's category, installed or carried outfits, and attributes.

Issue #2911, opened 2017-08-10
Allow missions to require an attribute to trigger


Amazinite:
In a similar fashion to how planets can require an attribute to land, it should be possible to have missions trigger based off of a ship's attributes. This would then allow us to create specific jobs for specific ships by giving those ships built-in attributes that trigger special jobs, solving old issues like #1751 and  #536 by allowing us to give certain transport ships a `"luxury"` attribute that would allow luxury transport missions to offer.

MCOfficer:
since you mentioned luxury transports as example, i'm just gonna leave #2661 here. either way, to me, this sounds like a useful addition with no big downsides.

Amazinite:
This would also allow those wealthy Hai jobs to offer regardless of which version of the stone you have on, instead of only offering when you have a stone from Hai space; I'm sure they won't be able to tell the difference. :)

Elyssaen:
Would PR https://github.com/endless-sky/endless-sky/pull/2616 satisfy this requirement?

Amazinite:
Effectively, yes.

lifeyouristhis:
```
mission "Escort (Tiny, Northern, Dangerous Origin)"
	to offer
		random < 40
		"combat rating" < 6
		or
			"combat rating" > 0
			has "ships: Interceptor"
```
I think it's already in-game.

Amazinite:
That's for a specific ship category, not for an attribute.

KyroFirehammer:
To go along with your trinkets idea the luxury tab could be added with a luxury outfit

warp-core:
[dfa9c8a](https://github.com/endless-sky/endless-sky/commit/dfa9c8a9d8d36d0d67560675f3eb06febe921f5d)

Amazinite:
Closing since #7797 allows the player's flagship to be checked, which is mainly what I wanted out of this issue.

PR #3140, opened 2017-10-19
Extend LocationFilter for ship sources

tehhowch:
Refs #3139 
Missions that trigger on boarding or assisting can now reference the ship's
  - category (e.g. "Heavy Warship")
  - outfits (e.g. "Electron Beam")
  - attributes (e.g. "cloaking fuel")

  Outfit and attribute references return true if the ship has a positive value for the provided token - a ship with a negative or zero-valued attribute/outfit count will not match that attribute or outfit.

  Outfits can also be used for planets to restrict by the items available in its outfitter.

These values are usable in the `source` (all), and  `destination` (outfits only) LocationFilters. A ship is not currently a supported mission destination, so specifying a ship attribute in the attributes node of a  `destination` filter will not work well (unless some planet in map.txt has that attribute).

Example:
```
assisting
source
    not category "Heavy Freighter"
    outfits "Heavy Laser" "Beam Laser"
    government "Merchant"
    attributes "scram drive"
destination
    outfits "Heavy Laser Turret"
on offer
    conversation
        `Please go find Joe on <planet> and inspect their supply of heavy laser turrets`
            accept
```
will match any merchant ship that is equipped or carrying at least one Heavy Laser or Beam Laser (gun versions), is not a Heavy Freighter, and has some outfit that provides a "scram drive" value (in base game, only the Scram Drive).
And will set the mission destination to be a random planet that sells Heavy Laser Turrets.

### DIFF
--- a/source/LocationFilter.cpp
+++ b/source/LocationFilter.cpp
@@ -46,6 +46,23 @@ namespace {
 		}
 		return false;
 	}
+	bool SetsIntersect(const set<const Outfit *> &a, const set<const Outfit *> &b)
+	{
+		auto ait = a.begin();
+		auto bit = b.begin();
+		// The stored values are pointers to the same GameData array:
+		// directly compare them.
+		while(ait != a.end() && bit != b.end())
+		{
+			if(*ait == *bit)
+				return true;
+			else if(*ait < *bit)
+				++ait;
+			else
+				++bit;
+		}
+		return false;
+	}
 	
 	// Check if the given system is within the given distance of the center.
 	int Distance(const System *center, const System *system, int maximum)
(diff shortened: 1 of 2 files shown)


Amazinite:
Looks good, but I fear that the difference between `outfits` and `outfit` (example:)
```
to offer
	outfits "Intrusion Countermeasure" # on the boarded/assisted NPC
on offer
	outfit "Cloaking Device" # on the player
```
will be confusing, and the use of `attributes` doesn't really tell someone viewing the mission data whether that attribute must be present on the player's ship (ref: #2911) or on the NPC's ship.

Great work, but different names would be appreciated.

tehhowch:
that `to offer` isn't valid - these are values for use in a LocationFilter, not a MissionAction

Amazinite:
```
source
	outfits "Intrusion Countermeasure"
```
Then. Fair enough, move along.

tehhowch:
My fault for not including an example in the OP :+1:

Issue #1360, opened 2016-07-21
Can "on enter" apply to any system with a specified attribute?


Bladewood:
For example, let's say we wanted an event to occur once we enter a Syndicate system, ANY Syndicate system. Could we use "on enter" to specify just any system part of Syndicate territory?

I made this issue to address the problem of arriving in a Syndicate system by jump drive, during the last parts of the Free Worlds war, where you're supposed to enter Syndicate systems by going through the "backdoor route". If this is implemented, we could fix that problem by adding a `on enter (government Syndicate)` line, or something like that, instead of a `event "fw syndicate welcoming" 11` line.

Wrzlprnft:
As far as i know you can specify the missiom to be offered on planets with the syndicate government, and, in a seperate line, set it to being offered on enter. It's basically two different conditions.

endless-sky:
At the moment, it needs to be either tied to a particular system, or left blank (in which case it applies as soon as you take off from the current planet).

jafdy:
#1082 would fix this. It could do with some rewriting though and I don't think I can bring it up to standard.

KiLEdEnNis:
jump drives ruin some mission mechanics, but adding an "on enter" tag that's tied to any sector governed by that government (unless unlinked) could help
and in this case having it as "on enter", tied to any Syndicate government, with a "near Foundry" limit should fix it
very confusing & complicated, but it's an idea

tehhowch:
With #3141 and e1c1a55c3a1bc2ec42ef0f9676a97c9a0c855e92, this is resolved.

PR #3141, opened 2017-10-19
Mission "on enter" can define a LocationFilter to pick the system(s)

tehhowch:
Refs #1360

Example use:
```
on enter
    system
        government "Syndicate"
    dialog "You're now in Syndicate space!"
```
 - Without a system filter, any `on enter` actions that do not name a system will apply as soon as the player takes off (current behavior).
 - ~~Each mission only supports a single `on enter` action without a named system (current behavior).~~
 - Multiple `on enter` actions that do not name a system are now supported, and will be checked to offer in the order they are defined
  - Only one `on enter` action will offer per jump/takeoff.


endless-sky:
I'm not convinced that that's the best syntax. For one thing, I usually try to avoid multi-word keywords - maybe just "system" would work? Also, it'd be a lot cleaner if there were a way for LocationFilter to just convert the "distance" filter into a "near" filter, rather than having to introduce a whole new object just to store what the distance is calculated from.

tehhowch:
I am open to alternate syntax - my first preference was `source` but since these are laid out in missions, and there are already folks with "indentation difficulty", having the same keyword permitted at multiple levels could be confusing.
That said, it would not be terribly confusing, since in both cases `source` is used to convey that the child nodes constitute a location filter that should match to a system.

I'll see about conversion from distance to near - I had thought about that briefly.

tehhowch:
One of the reasons i went with defining the mission origin was that there was no way to indicate to a mission during Mission::Load (and thus MissionAction::Load and the onEnter's systemFilter.Load) which system it will be called from during instantiation. For `near`, this is not an issue since the system reference is known at all times.
I suppose that a new method could be added to the LocationFilter class, which could be called during instantiation, and return a new LocationFilter (since ___::Instantiate() is const) with a converted distance.

tehhowch:
@endless-sky addressed. I went for utility with the function name, but it can probably be improved.

endless-sky:
I think it'd be really confusing to people that they can't define two different "on enter" actions with different filters in the same mission. It'd probably be better to store all the "on enter" actions with no specific system in a separate list and to check each of them against the system passed to Mission::Enter(). (Note: that would also remove the need for Mission::Enter() to be called twice, both with a system and with nullptr, in Mission::Do(ShipEvent...).)

tehhowch:
Multiple generic `on enter` are now supported

source/LocationFilter.cpp
```diff
@@ -215,6 +212,31 @@ bool LocationFilter::Matches(const Ship &ship) const
...
+LocationFilter LocationFilter::SetOrigin(const System *origin) const
+{
+	if(IsEmpty() || originMaxDistance < 0)
+		return *this;
+	
+	// Copy all parts of this instantiated filter into the result.
+	LocationFilter result = *this;
+	// If this LocationFilter has already defined a "near <system>" filter, then
+	// do not convert a "distance" filter into a "near" filter.
+	if(!center && originMaxDistance > -1)
```
endless-sky: Since `originMaxDistance` is an integer, `originMaxDistance > -1` is equivalent to `!(originMaxDistance < 0)`, so it will always be true if you get to this point in the function.

source/Mission.cpp
```diff
@@ -1068,14 +1080,27 @@ Mission Mission::Instantiate(const PlayerInfo &player) const
...
+	// consider matching a generic "on enter" to it. A generic "on enter" action
+	// may have a LocationFilter that governs which systems it can be performed in.
+	if(!genericOnEnter.empty() && !didEnter.count(system))
+		for(auto it = genericOnEnter.begin(); it != genericOnEnter.end(); ++it)
+			if((*it).CanBeDone(player))
+			{
+				(*it).Do(player, ui);
+				genericOnEnter.erase(it);
+				break;
+			}
```
endless-sky: The reason for "didEnter" is that if you erase the MissionAction immediately, and the MissionAction contains a conversation, that conversation is deleted while the panel that displays it is still active, resulting in a crash or other weird behavior.

A better option would be to modify didEnter to store a pointer to a MissionAction rather than to a system. That way, it could apply to both the sets of one enter actions, e.g.:

```
if(eit != onEnter.end() && !didEnter.count(&eit->second)...
```

Also, if multiple "on enters" apply to the same system, I feel like it's better to apply all of them (showing multiple dialogs, if necessary) rather than allowing all but one of them to silently be ignored.

source/MissionAction.cpp
```diff
@@ -176,6 +176,13 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			// Create a GameData reference to this mission name.
 			GameData::Missions().Get(toFail);
 		}
+		else if(key == "system")
+		{
+			if(system.empty() && child.HasChildren())
```
endless-sky: Generally unless there's a good reason not to, I allow data specifications to break a particular data entry into multiple pieces. So, I don't think there's a need to check `system.empty()` here.

source/MissionAction.cpp
```diff
@@ -384,6 +402,8 @@ MissionAction MissionAction::Instantiate(map<string, string> &subs, int jumps, i
 	MissionAction result;
 	result.trigger = trigger;
 	result.system = system;
+	// Convert any "distance" specifiers into "near <system>" specifiers.
+	result.systemFilter = systemFilter.SetOrigin(GameData::Planets().Get(subs["<origin>"])->GetSystem());
```
endless-sky: This won't work if the origin is a ship you're boarding (in which case `"<origin>"` is the name of the ship, not the name of a system). It might be necessary to just pass the current system to `MissionAction::Instantiate()`.

source/Mission.cpp
tehhowch: As written the generic `on enter`s wouldn't be ignored, just not activated (since some other O.E. activated instead). The next time they match a system they would be activated (since I was using the direct erasure and not checking for them in `didEnter`).
Without getting crazy specific in a LocationFilter, it's difficult to set a "sequential" scene, where a conversation triggers in one system and then a related conversation triggers in the next. With a single generic O.E. triggering per system entry (and only if a specific O.E. did not), this is a lot easier to do, and the order of display is controlled by the order of specification in the datafile.

I'll make the didEnter<MA*> change and we can invite further discussion on displaying all matched generics vs only one per jump.

source/MissionAction.cpp
tehhowch: `system` is the `on enter` MissionAction's named system, e.g. for `on enter Sol` system is `Sol`

This check is to prevent someone from explicitly naming a system at the root level of the MissionAction, and *also* defining a LocationFilter to pick the system.

tehhowch:
I've made the suggested changes, with the exception of allowing multiple on enter to activate at the same time.

endless-sky:
Well, now that I think about it, it'd be kind of cool to use a bunch of `on enter`s with no system specified to pop up a series of messages, one per system. (For example, the "wolf, cabbage, goat" mission would be able to make the goat start munching on the cabbage after one jump, and then have the goat's demise happen on the next.)

tehhowch:
Yep, it's now easier to tell a story as the player travels without requiring them to take a specific route (or defining a bunch of extra `on enter`s to account for slight deviations/jump drives).
:+1:

Issue #3143, opened 2017-10-20
System Attributes


tehhowch:
Like planet attributes.
Definable in the map

It makes sense to generate some automatic attributes though, based on the stellarobjects and planets it contains. I'm thinking
 - `"galactic phenomena"`
   - `>=4` stars, any wormhole sprite, or star has the `nova` or `wr` sprite
 - `"ternary star"`
 - `"binary star"`
 - `"no star"`
 - `"inhabited"` & `"uninhabited"`
 - `"high moon density"`
   - more than 4 moons per planet on average
 - `"high minable density"`
    - calculated by `total minable asteroids > 60`
 - `"nexus"`
   - System has 6 or more links to other systems
 - `"wormhole"`
   - the system has an accessible wormhole (which doesn't have to use the wormhole sprites)

What other automatic attributes make sense?

These automatic attributes, plus those of the planets it contains (so a system could be both wealthy and poor), plus those defined on the map, would be accessible to LocationFilters (to help make better missions without hard-coding destinations).
Refs #526

MCOfficer:
> "no star"

??


i like it a lot, will fix quite some problems content creators stumbled upon.

tehhowch:
Yeah, no reason to assume that a system has a star... Maybe it's a black hole :) that'd be interesting

Amazinite:
`"high asteroid density"` > some amount of asteroids and minables
`"high planet density"` > some amount of planets

Those are the only two automatic attributes I can think of that I would think should be simple to do given that you've already done. 

Change nexus to >= 4 or 5 links? How many systems in game right now have >= 6 links?

tehhowch:
Using map.txt systems only, and a regular expression `\d\n(\tlink [\"A-z '-\d]+\n){#}`:

| Minimum Links | Number of Systems |
|:---:|:---:|
| 1 | 344 |
| 2 | 262 |
| 3 | 186 | 
| 4 | 99 |
| 5 | 26 |
| 6 | 8 |
| 7 | 1 |

The base map's nexus systems, at 6+, would be
 - Sol Arach (7)
 - Alniyat
 - Gupta
 - Kaliptari
 - Lesath
 - Ruchbah
 - Similisti
 - Sol

After completing the main plot the numbers shift a little due to the addition of Deneb.

Amazinite:
`crossroads` == or >= 4
`hub`== or >= 5
`nexus` >= 6
?

tehhowch:
[...]

Issue #526, opened 2015-12-29
Let waypoints be random, just like stopovers


Pointedstick:
The docs indicate that waypoints have to be static and hardcoded in the mission description. I'd like to be able to specify random waypoints in much the same was as stopovers. For example, something like this:

> waypoint
>     attributes uninhabited

That would choose a random uninhabited system

> waypoint
>     attributes north

And that would choose a random system in the galactic north

etc.

endless-sky:
At the moment, systems do not have attributes, so this would require adding support for attribute tags on systems, and going through every system in the map and adding those attributes in.

Pointedstick:
It seems like systems could simply inherit attributes from a union of all of the attributes of populated planets in the system, or else just have "uninhabited", plus "weird phenomena" if the system has a wormhole or supernova-ing star or something in it.

endless-sky:
In commit d24fd18 I've added support for random waypoints based on location filters, but I'll still need to think about the best way to allow systems to have searchable "attributes" of their own.

tehhowch:
I support @Pointedstick 's idea of the system inheriting attributes from its contained planets, plus whatever attributes it might have on its own.
To begin with, perhaps only the "automatic" attributes are supported - those from the planets, plus "galactic phenomena" if it contains a wormhole or a specific star type.
As it is, I'm not sure what other attributes would exist or need specifying.

Amazinite:
The only attributes I can think of needing are the government, region, inhabited/uninhabited status, and galactic phenomena as mention above. Or at the very least that would be the baseline for generic attributes that I would expect each system to have.

There may always be special missions though which require a certain system attribute though. Say, for example. "visit these X system, which all contain a gas giant," or some other planet type (ocean, lava, forest, rocky, ice giant, etc), or "visit these X systems which all have Y amount of planets," or possibly even "visit this system with X amount of asteroids (minable, nonminable, or both)." In relation to something that @tehhowch is working on if I'm recalling correctly, a system's fleet activity could even be an attribute of interest.

tehhowch:
i could update #3140 to include pulling attributes from planets into the source system, but i'll open a new pull for that since 3140 is "done" and there are probably a few things to discuss on system attributes that i'd rather not condition 3140 with

tehhowch:
#3144 adds automatic system attribute generation, including sensible planet-system attribute pooling in LocationFilter. I feel it resolves this issue in full.

endless-sky:
[...]

PR #3144, opened 2017-10-20
Enable System Attributes and their use in LocationFilters

tehhowch:
Refs #3143 #526 

LocationFilter will now union a planet and its system, and a system and its planets, while checking for matches to a particular set of attributes.
A system's attributes are comprised of those specified via file, with the `attributes` node:
```
system "Sol Arach"
    attributes "nexus"
```
and those that are automatically generated from its StellarObjects and links.  The above "nexus" attribute would be unnecessary, as Sol Arach has 7 systems linked to it, and the "nexus" attribute is automatically assigned for those systems that have 6 or more linked systems.

Current automatic system attributes ( #3143 )
 - nexus
 - crossroads
 - galactic phenomena
 - ternary star
 - binary star
 - no star
 - high solar wind
 - high luminosity
 - high minable density
 - high asteroid density
 - many moons
 - wormhole (when unioned with `PlanetAttributes()` in LocationFilter::Matches, if the wormhole is accessible)
 - inhabited (when unioned with `PlanetAttributes()` in LocationFilter::Matches, if the inhabited planets are accessible)
 - uninhabited

The import of attributes from a system's accessible planets means the system will import any geographic or civilization based attributes (e.g. `pirate`, `north`, `dirt belt`, etc). It also means "contradictory" attributes can exist - for example, peacekeeping forces might need an escort to a system that is `urban rich poor` (note that because an individual planet won't have both attributes, contradictory attributes won't match if `and`ed in a source, destination, or stopover filter (but will work in a waypoint or NPC system filter).

An alternative (or perhaps complementary) approach would be to specify another embedded LocationFilter that is applied only during system match-checking.

### DIFF
--- a/data/stars.txt
+++ b/data/stars.txt
@@ -77,3 +77,13 @@ star "star/nova"
 star "star/wr"
 	power 0.4
 	wind 4
+
+
+
+# Sprites which classify as "galactic phenomena" - any system in which
+# these can be found will have the attribute "galactic phenomena".
+phenomena
+	"star/nova"
+	"star/wr"
+	"planet/wormhole"
+	"planet/wormhole-red"
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -58,11 +58,8 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include <algorithm>
 #include <iostream>
-#include <map>
 #include <utility>
-#include <vector>
 
-class Sprite;
 
 using namespace std;
 
@@ -101,6 +98,7 @@ namespace {
 	Trade trade;
 	map<const System *, map<string, int>> purchases;
 	
+	set<const Sprite *> galacticPhenomena;
 	map<const Sprite *, string> landingMessages;
 	map<const Sprite *, double> solarPower;
 	map<const Sprite *, double> solarWind;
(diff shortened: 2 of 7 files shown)


tehhowch:
Updated the moon criteria based on game data posted in #3143

endless-sky:
Having that many auto-generated attributes doesn't really make sense to me. The auto-attributes for planets (spaceport, shipyard, and outfitter) have really obvious meanings and are likely to be useful, but many of these attributes have unclear meanings or limited usefulness.

tehhowch:
I felt it better to err on the side of exposing more keywords for mission locations than fewer, as it allows a bit more specificity when writing random missions and wanting to have appropriately descriptive text. They are definitely on the arbitrary side, but make it possible to write missions to escort ships that want to go to a mining-heavy system/asteroid-heavy system, smuggle a passenger from/to a system having many links, lead researchers to systems with crazy/multiple stars, lots of moons, etc, without needing to write one mission for each possible destination.

tehhowch:
I've dropped two auto-attributes in favor of two new ones based on 317b495a7217784bb490637564e1ff57e22c0193.
I also removed the static definitions of what constitutes a "galactic phenomena" into the `data/stars.txt` file, allowing plugin creators to easily change which sprites generate the "galactic phenomena" auto-attribute.
There were also some legacy includes in GameData so I've pruned them away.

endless-sky:
This still seems like way too many auto-attributes and other new features - again, considering the fact that once they're added, the game needs to support them all in perpetuity. It makes more sense to me to add the bare minimum that's clearly useful, and only expand it over time if there's a definite need.

In commit e1c1a55 I've added support for system attributes, with a single auto-attribute (uninhabited). That's the one attribute that's clearly going to be useful.

Issue #3147, opened 2017-10-21
Outfitter is broken


Bladewood:
Click on any outfit.
You'll get weird information, and some of it is missing entirely.
![whatthehelloutfitter1](https://user-images.githubusercontent.com/16506683/31854496-53bf7b32-b668-11e7-974d-1959cef5cfb7.png)
![whatthehelloutfitter2](https://user-images.githubusercontent.com/16506683/31854497-54fd54f6-b668-11e7-85cd-d4b74392a193.png)
![whatthehelloutfitter3](https://user-images.githubusercontent.com/16506683/31854499-56c3a7ae-b668-11e7-9a87-5879785648eb.png)

Think this is due to https://github.com/endless-sky/endless-sky/commit/595d2e8557489ae4549e42e3eb42a5ac7eecd04d, since it was working the commit before, and that one seems to be the most significant change.

Edit: huh, it looks like the broken outfitter text randomizes on startup or something.
Edit^2: tooltips don't work.

tehhowch:
The first two columns are fine, but the 3rd panel is indeed goofed.

The problem occurs in OutfitInfoDisplay::UpdateAttributes
It is iterating over the result of `outfit.Attributes()` which is indeed a Dictionary object added in 595d2e8, and the iterator is `pair<const char *, double>`
For a Meteor Missile, one of the `it` values is:
![image](https://user-images.githubusercontent.com/20871346/31862800-9bba6090-b709-11e7-8130-f09538a41034.png)
The issue occurs when we reach line 195, and we combine this value with `':'` and then copy it into the attribute labels container (`vector<string>` type). This generally results in some very interesting text being accessed.

tehhowch:
This should be ready for a milestone as it was fixed in #3150

endless-sky:
Since the bug was introduced after the 0.9.8 release (and thus is only present in code straight from GitHub), I'll go ahead and close this.

Issue #3108, opened 2017-10-10
Best order for weapon attributes?


endless-sky:
Currently weapon attributes are listed in the following order:
* ammo
* range
&nbsp;
* shield damage / second
* hull damage / second
* heat damage / second
* ion damage / second
* slowing damage / second
* disruption damage / second
* firing energy / second
* firing heat / second
* firing fuel / second
&nbsp;
* shots / second
* turret turn rate
* homing
* tracking
* optical tracking
* infrared tracking
* radar tracking
* piercing
&nbsp;
* shield damage / shot
* hull damage / shot
* heat damage / shot
* ion damage / shot
* slowing damage / shot
* disruption damage / shot
* firing energy / shot
* firing heat / shot
* firing fuel / shot
&nbsp;
* inaccuracy
* blast radius
* missile strength
* anti-missile

That's not necessarily the most logical order for the attributes, as mentioned in PR #1654. The goal was to put the most important attributes (range and damage / second) first, but the order of the other attributes is somewhat random.

KiLEdEnNis:
in advance: not sure what to do with the /second and /shot, I put /shot below /second for now
and this could be "fine tuned" a bit, but I think this would be generally better

damage / second:
- shield damage / second
- hull damage / second
- ion damage / second
- heat damage / second
- disruption damage / second
- slowing damage / second
- anti-missile (only present if there is no other damage, unless a later update)

damage / shot
- shield damage / shot
- hull damage / shot
- heat damage / shot
- ion damage / shot
- slowing damage / shot
- disruption damage / shot
- firing energy / shot
- firing heat / shot
- firing fuel / shot

fire "requirement" / second:
- firing energy / second
- firing heat / second
- firing fuel / second

fire "requirement" / shot:
- firing energy / shot
- firing heat / shot
- firing fuel / shot

important non-damage stats:
- turret turn rate(?)
- range
- shots / second (not sure about this placement, it could be in the damage tab too)
- ammo (paired with shots / second on purpose)
- inaccuracy

secondary weapon stats:
- homing
- tracking
- optical tracking
- infrared tracking
- radar tracking
- piercing (not sure about this placement, it may be used on other weapons later)
- missile strength
- blast radius

tehhowch:
My thought is that, no matter the order, per shot and per second values need to be in the same row, as this cuts down on the vertical requirements, which means the outfitter does not have to scroll as much. Forced scrolling is jarring and IMO, bad UX.

[...]

PR #3150, opened 2017-10-22
Fix Outfitter Attribute display bug

tehhowch:
Refs #3147 

The fix is to not combine the `const char *` value `it.first` and `':'`. Previously, `it.first` was a `std::string`, but after 595d2e8557489ae4549e42e3eb42a5ac7eecd04d it became `const char *`. Due to use of `auto` for the iterator `it`, the pointer to char was kept, rather than converting to a string for display / use.

Summary:
 - Change `auto` to the type `pair<string, double>` when `it.first` is to be used as a string
 - `push_back` to `emplace_back`
 - Simplify display of per-second values
 - Omit per-second values if reload = 60 (i.e. per second == per shot)
 - Separate non per shot / per second values from values which should be printed as per shot / per second
    - Addition of new damage types no longer requires updating the `isContinuous` offset

### DIFF
--- a/source/OutfitInfoDisplay.cpp
+++ b/source/OutfitInfoDisplay.cpp
@@ -242,10 +242,9 @@ void OutfitInfoDisplay::UpdateAttributes(const Outfit &outfit)
 		outfit.FiringFuel()
 	};
 	
-	// Add any per-second values to the table. If the outfit reload is 60,
-	// do not print per-second values since they are the per-shot values.
+	// Add any per-second values to the table.
 	double reload = outfit.Reload();
-	if(reload && reload != 60.)
+	if(reload)
 	{
 		static const string PER_SECOND = " / second:";
 		for(unsigned i = 0; i < values.size(); ++i)
--- a/source/OutfitInfoDisplay.cpp
+++ b/source/OutfitInfoDisplay.cpp
@@ -153,8 +153,8 @@ void OutfitInfoDisplay::UpdateRequirements(const Outfit &outfit, const PlayerInf
 	for(unsigned i = 0; i + 1 < NAMES.size(); i += 2)
 		if(outfit.Get(NAMES[i + 1]))
 		{
-			requirementLabels.push_back(string());
-			requirementValues.push_back(string());
+			requirementLabels.emplace_back();
+			requirementValues.emplace_back();
 			requirementsHeight += 10;
 		
 			requirementLabels.push_back(NAMES[i]);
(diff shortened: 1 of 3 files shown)


tehhowch:
I dropped printing the per second version when per second == per shot because then it is apparent that these values are needed each time the outfit fires, and because continuous weapons will look different than discrete weapons.

endless-sky:
Given how few weapons fire exactly once per second, I feel like it would make sense to always show values per second, and only drop values per shot when the weapon fires continuously. (Especially because damage per second is what the player will use for comparing weapons, so it'll be really confusing to them if one weapon shows no DPS values.)

tehhowch:
I'm tempted to try my hand at reordering the attributes, but I'll leave that to you following the inputs of #3108

tehhowch:
Made a mockup - I have a stashed commit I can push if its OK:
```
{non-weapon attributes}

ammo
range
inaccuracy
blast radius
turret turn
anti-missile
missile strength
homing
tracking
piercing

shots / sec
{per sec damage & usage}

{per shot damage & usage}
```
Detainer:
image

Disruptor:
image

Finisher Pod:
image

Point Defense:
image

source/OutfitInfoDisplay.cpp
```diff
@@ -171,147 +171,117 @@ void OutfitInfoDisplay::UpdateAttributes(const Outfit &outfit)
 	attributeValues.clear();
 	attributesHeight = 20;
 	
-	map<string, map<string, int>> listing;
-	for(const auto &it : outfit.Attributes())
+	for(const pair<string, double> &it : outfit.Attributes())
```
endless-sky: When you do this, what's actually happening is that you're constructing a copy of each attributes node (via pair's templated constructor) and then getting a const reference to that temporary (which extends its lifetime) and then using that reference. That's a lot of behind-the-scenes stuff to hide behind a for loop. I think it'd make more sense to just iterate through the actual items in the attributes set (which are pair<const char *, double>s) and then typecast the name to a string if needed for concatenating strings.

tehhowch: That was how I had it originally, and then when I changed it to test if there would have been a warning when the type of `it` changed from <string, double> to <const char *, double> I found that no warning was generated (and no display bug from adding `':'` to the char * was present either), so I stuck with it.
I'll swap back

tehhowch:
@endless-sky addressed.
Thoughts on if I should push the reordering as well?

endless-sky:
I think I'm going to experiment with the idea of putting the values in a table, which would probably conflict with any reordering (or at least very quickly make it obsolete).

Issue #3146, opened 2017-10-21
Performance in fights with minelayer ships.


flaviojs:
While exploring outer edge systems and their planets with 2 Mules, I jumped (with Jump Drive) Sepetrosk -> Sayaiban -> Faronektu. As I waited for fuel to refill in Faronektu I noticed the battles were producing a constant stream of capturable ships. I decided to remain there capturing, aiming to get at least a ship of each kind. ^^

As I spent time there I noticed the game speed had big fluctuations, I even saw the cpu stat go over 1000%. After some time I narrowed down the cause to the mines that were being deployed by Kor Mereti ships. To confirm this I reloaded the game in Sepetrok to get a mine field deployed and see the difference.

A Model 64 and 2 other ships started chasing me. The Model 64 deployed a mine field, which caused the cpu stat to jump to 42%. After all the mines exploded the cpu stat went down to 15% (they were still chasing me).

**I know my cpu is weak but a jump from 42% to 15% over a single mine field is too much.**
Clearly something should be improved so I recompiled with ```scons mode=profile```, fired up the game with callgrind, reduced the size of the game window, and left the game running in Faronektu for a few hours (with caps lock on).

Callgrind made everything around 30 times slower, the biggest cpu stat I saw was 40099% (~1300% without callgrind?). It looked like a small battle followed by a big battle from the times I peaked into the window.

Within the +4h of run time ```Engine::CalculateStep()``` was called 13783 times. After inspecting with kcachegrind, these appear to be the critical calls:

```
mcount  [49.81% total, libc]
0x000000000004bf10  [19.46% total, libopenal1]
75.65%  Engine::ThreadEntryPoint()
  75.65%  Engine::CalculateStep()
     7.58%  [other code?]
     5.93%  AI::Step(PlayerInfo const&)
    62.14%  AsteroidField::Collide(Projectile const&, int, double, Point*)
       0.53%  [other code?]
       3.31%  Minable::Collide(Projectile const&, int)
      58.30%  AsteroidField::Asteroid::Collide(Projectile const&, int) const
        15.32%  [other code?]
         5.48%  drem  [libm]
         2.96%  Point::operator=(Point const&)  [3.96% total]
         3.01%  operator*(double, Point const &)  [3.06% total]
         5.00%  Point::operator-(Point const&) const  [6.69% total]
         5.12%  Body::Velocity() const  [5.49% total]
        10.75%  Body::GetMask(int) const  [11.86% total]
[...]

PR #3153, opened 2017-10-23
Add spatial container for asteroids

flaviojs:
Reference: #3146

This is a space-time tradeoff using http://spatial.sourceforge.net/.

Why did I use this particular library?

* header only library simplifies dependencies
* don't have to configure anything
* don't need to create or modify existing classes, standard ones are sufficient

I'm not sure how this kind of dependency should be handled, so I took the easy way out and just dumped the root files and the source (headers) into ```source/spatial/``` =~

----

Possible improvements that I didn't try:

1. If mask collisions are heavy, doing line-box overlap tests to decide when to do them should improve performance. This is not worth trying unless mask collisions are confirmed to have a significant performance impact. I would probably try the method described in the answer to https://math.stackexchange.com/questions/91362/fast-2d-line-triangle-intersection-test
2. When the movement crosses both vertical and horizontal wrap edges, there is 1 or 2 of the split boxes that can be discarded. This looks like a small performance impact, so probably not worth the trouble.


Amazinite:
Wow, that's a lot; I'll test this out when I get the chance.

jafdy:
This is a roughly 10% increase in the number of lines in the entire project (both C++ and .txt). OK, it is very well commented but even so I do wander if it is a bit obtuse.

tehhowch:
I'm not sure if you can submodule a repo hosted not on git, but I imagine that would be the preferred way to rope in an external dependency.

Going solely off your comments in da55965, the use of this library sounds a fair bit like the CollisionSet class, in which only those objects within a grid cell (containing the projectile) are checked for collisions rather than all objects

flaviojs:
CollisionSet is a spatial container too, but uses non-overlapping fixed-size cells instead of content-based regions. Looks like a matrix instead of a tree.
I don't see non-Body functions being used so Ship can probably be changed to Body.
Not sure what would need to change so that it works as an infinitely-repeating region.

It's definitely a possibility, I'll try adapting it. :)

tehhowch:
It was just recently made Ship-only :)
Definitely keep this pull open (maybe see if it can be done via git submodule for easily performing version updates first) for input on external dependencies in the future

endless-sky:
It doesn't make any sense to me to add a completely different set of functions for spatial lookup when CollisionSet already exists (and probably has very similar performance).

flaviojs:
This PR was superseeded.

PR #3158, opened 2017-10-24
Local Hai space merchants

jjhankins:
This creates a new government ("human Merchants") and merchants in Hai space are given that government. This is so the Hai don't get (directly) mad at you if you pirate Merchants in human space. Hai-space merchants will still get mad at you, and if you attack them that will damage your reputation with the Hai, but with this it's possible (though tricky) to be a pirate in human space and a cargo-hauler in Hai space.


Amazinite:
Capitalize Human.

jjhankins:
You think? I thought it'd emphasize the connection to regular merchants, make it read like a descriptor.

tehhowch:
As-is, they won't. The value added by my earlier PR is .01, below the threshold of .05 for permanent reputation change.

jjhankins:
Ah, didn't realize that. Hm.

jjhankins:
Nevermind, looks like this PR doesn't do anything useful.

PR #3160, opened 2017-10-24
Add spatial container for asteroids - take 2

flaviojs:
Reference: #3146

This is a space-time tradeoff using CollisionSet.

This PR superseeds PR #3153.


endless-sky:
After looking at this for a while, I decided that it makes more sense to implement the tiling behavior in AsteroidField (because it's a special behavior unique to that class) rather than in CollisionSet. I did that in commit 14d70ba.

Issue #3145, opened 2017-10-21
You can board and steal flamethrower from test dummy ship


JoshuaGrams:
I'm running version 0.9.8 on Windows.  Barmy Edwards tells you he'll "have an initial model [flamethrower] for sale in a month or so," but after disabling the flamethrower ship you can board and loot its four flamethrowers, sell them at a planet with an outfitter, and buy them back to get them immediately.

jafdy:
The easy solution would be to make the ship `self-destruct 1`, that way is the player where to board it they would fail the mission and not get anything to boot. It would however leave you almost certainly in need of a reload. I expect the best thing is just to leave it as the other options are worse.

tehhowch:
If players choose to board it after being told not to,, they clearly don't want to be a part of an organization that expects basic human decency. Let them flunk out FW.

(In the future, taking these outfits to a competing faction could be a side plot mission)

Amazinite:
If stealing the flamethrowers is an issue, then mechanics from #2496 could be used. Have this mission
```
on complete
	require "Flamethrower" 0
``` 
and have a different mission 
```
to offer 
	has "<the mission in question>: Active"
on offer
	require "Flamethrower 1"
	fail "<the mission in question>"
	fail
```

endless-sky:
Far easier would be to just give the test dummy ship 100% self-destruct probability.

tehhowch:
There aren't any direct tie-ins to having successfully completed or failed `FW Flamethrower 2`. The closest thing is `FW Southern Battle 1B`, in which you escort several flamethrower furies to Trinket.
The flamethrower event only adds the outfit to the Kraz outfitter, which is rather basic compared to the plasma turret event (which sends you a message saying "Hey, come buy this!"). The only real penalty of destroying the Doombat is that you cannot ever buy the Flamethrower, even though it becomes available for use by the rest of the Free Worlds military.

I considered adding an `on fail` node to the mission, in which a conversation is shown and the event time is delayed from 1 month to 6 months, but I lack the character context to convincingly write "mad Barmy.", and a simple time delay isn't much penalty at all.

tehhowch:
You can no longer obtain the Doombat's flamethrowers, but #3161 did not add any additional penalty (beyond losing all access to purchasing the flamethrower). So once this is milestones, I would close it.

PR #3161, opened 2017-10-24
Add self-destruct to Barmy's Doombat

tehhowch:
Refs #3145 

It's best to not attempt to board spacecraft that are described to be barely spaceworthy.


endless-sky:
Too bad it's not currently possible for a ship variant to also override how much damage a ship's explosion weapon does...

tehhowch:
That ellipsis implies it might be possible in the future :+1:

Hadron1776:
For the record, it's best not to blast said barely-spaceworthy ships with particle cannons, either.

Issue #1452, opened 2016-08-15
Have mission clearance stop security checks


Amazinite:
Make clearance stop security checks on the planet that you were given clearance to. You were just allowed to land on the planet for some special reason, so I doubt they are going to go through your cargo and fine you for some nerve gas grenades when you're there for some special reason.
This could also be used for missions where you supposedly land somewhere other than the spaceport, as it makes no sense for you to get fined in those missions.

Also possibly allow government clearances that would stop a government from fining you for the duration of the mission, effectively closing #1429.

KiLEdEnNis:
it's weird how a big gunboat can even find a small can of nerve gas with scanners, and if they'd check a ship you could "hide" the nerve gas, because it's a small box in a big ship

anyway maybe it'd help if the Republic wouldn't scan/fine you for illegal outfits during the war (first invasion to either end), because it's annoying if you have nerve gas knowing the enemy Republic won't scan you and at a ceasefire they do scan and fine you

Amazinite:
Yeah, it would make sense if the Republic couldn't find you for having illegals during cease fires, since technically at that time they have no jurisdiction over you. Maybe use the hostility attribute instead in that situation.

Bladewood:
This could work for missions with the `infiltration` tag (or something like that).

In the war, you would still be fined because nerve gas is not just banned by the Republic, it is banned in all of human space, by interstellar law. The Laws of War are the few things warring human factions will agree on.

KiLEdEnNis:
@Bladewood although nukes should be illegal as well, in the checkmate the FW allows them, although that would be a part of the Laws of War

Bladewood:
The Free Worlds allow them because you're a bloody council member. If you were any lower in rank, you'd be sentenced to death anywhere in human space.

KiLEdEnNis:
a democracy where only certain people may use nukes, bless the FW

Amazinite:
Looks to me that this is the case now.

https://github.com/endless-sky/endless-sky/blob/9b58c03f934c18b1d4e4a62acff8d28c2efc57e8/source/PlayerInfo.cpp#L2618-L2622

tehhowch:
The bit about stopping fines from government x is why this wasn't closed, I think

Amazinite:
I get the sense that that is already possible.

[...]

PR #3162, opened 2017-10-26
When you click a system in the map mission view, iterate through all missions that involve the system.

flaviojs:
This allows you to know which of the missions has the system as a waypoint or has a system planet to land on.

### DIFF
--- a/source/MissionPanel.cpp
+++ b/source/MissionPanel.cpp
@@ -43,6 +43,26 @@ using namespace std;
 
 namespace {
 	const int SIDE_WIDTH = 280;
+	
+	// Checks if a missions involves the provided system.
+	bool Involves(const Mission &mission, const System *system)
+	{
+		if(system == nullptr)
+			return false;
+		
+		if(mission.Destination()->IsInSystem(system))
+			return true;
+		
+		for(const System *waypoint : mission.Waypoints())
+			if(waypoint == system)
+				return true;
+		
+		for(const Planet *stopover : mission.Stopovers())
+			if(stopover->IsInSystem(system))
+				return true;
+		
+		return false;
+	}
 }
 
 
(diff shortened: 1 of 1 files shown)


jafdy:
Screen shot?

flaviojs:
There's nothing to screenshot? When you click a system:

* old behavior = select next accepted or available mission with that destination (skips missions that have a waypoint or stopover key there) 
* new behavior = old behavior missions + missions with a waypoint or stopover in the system

jafdy:
I read this as "highlight all missions that involve that system", as in multiple lines get highlighted. If that's not what it is then don't mind me.

source/MissionPanel.cpp
```diff
@@ -43,6 +43,26 @@ using namespace std;
 
 namespace {
 	const int SIDE_WIDTH = 280;
+	
+	// Checks if a missions involves the provided system.
```
tehhowch: `Determine if the given mission uses the given system`

or at least `a` -> `this` and drop the s on `missions`

source/MissionPanel.cpp
```diff
@@ -43,6 +43,26 @@ using namespace std;
...
+	bool Involves(const Mission &mission, const System *system)
+	{
+		if(system == nullptr)
```
tehhowch: !system

flaviojs:
Hmm, the function fits better as part of the Mission class but I can't think of another place that would use it.

tehhowch:
I was thinking similarly too. There may be a use for a Mission::LandingTarget(const System *) function that unions its stopovers and destination in the given system, then returns a(n accessible) stellarobject pointer for the first result (preferring an unvisited stopover to the destination).
This function would replace two of your steps here (you'd just test for a nullptr return), and could be used in AI::MovePlayer as well.

tehhowch:
The construct I mention above could be used to achieve #1452 - when landing, do not roll for illegal items if this planet is associated with a mission that gives clearance. Probably best done an anonymous namespace method that takes a missions list and a planet pointer, and checks to see if any mission gives clearance for that planet

flaviojs:
A waypoint only implies entering the system, no landing target/planet, so it wouldn't replace my function. Since refactoring other code is out of scope for this PR, I'll leave it as is unless MZ tells me to move my function to Mission.

tehhowch:
Indeed, it would not. It would only handle the destination and stopover portions.

Issue #2737, opened 2017-07-10
Message congratulating the player upon paying off their first mortgage


Pointedstick:
Such a thing would add a nice little touch.

tehhowch:
Could possibly be done with an invisible mission that is offered (once) when the player has a condition.
We could add to Account::Step and Account::PayExtra to set/update a simple condition "mortgages paid off" when a mortgage gets paid off.

Would be a neat touch indeed.

If someone else would write the mission and conversation/dialog to show, I (or others) can add the needed bits to Account.cpp.

I'd prefer a convo/dialog over a hail that shows up on the ticker, as that can be missed really easily, and for a "momentuous" occasion such as one's first starship loan... no one wants to miss it. We lose the guarantee that it shows up immediately upon payoff though, but that discrepancy can be handled in the text.

Pointedstick:
:notes: This is now possible
I think I'll do it :notes:

Pointedstick:
Another snag: you can't have a mission offered at the bank. https://github.com/endless-sky/endless-sky/issues/3297

PR #3164, opened 2017-10-27
Create player conditions for mortgages, fines, back pay, and credit score

tehhowch:
Creates a few new automatic conditions to allow missions based on the player's financial acumen.
Refs #2737 

I'll let more seasoned writers draft the message / conversation @Pointedstick was looking for :+1:

### DIFF
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -311,19 +311,6 @@ int Account::CreditScore() const
 
 
 
-// Find out how many mortgages and how many fines the player has.
-vector<int> Account::TypeCount() const
-{
-	static const string FINE = "Fine";
-	vector<int> result{0, 0};
-	for(const Mortgage &mortgage : mortgages)
-		++result[mortgage.Type() == FINE];
-	
-	return result;
-}
-
-
-
 vector<int64_t> Account::TypeOwed() const
 {
 	static const string FINE = "Fine";
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -311,6 +311,31 @@ int Account::CreditScore() const
 
 
 
+// Find out how many mortgages and how many fines the player has.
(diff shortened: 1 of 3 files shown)


endless-sky:
This would get incremented every time the player "refinances" their mortgage (i.e. takes out a new mortgage at a lower rate and uses it to pay off the original mortgage). For #2737, it'd be much better for the variable to instead store the current total number of mortgages, so you know when the player is actually completely debt-free.

Amazinite:
While on the topic of player conditions for mortgages, perhaps a `"credit score"` condition would go well along with the mentioned `"mortgages"` condition.

tehhowch:
I had gone with this approach to avoid needing to add a variable to the PlayerInfo class (and am sad to say I never considered refinancing my ES loans, so I hadn't considered that behavior).

Credit score autocondition shouldn't be too hard to add either.

tehhowch:
Updated to be more robust about the exposed autoconditions

endless-sky:
To me it would make a lot more sense to only have auto-conditions for total debt, credit score, and salaries owed. In general I try to avoid adding too many auto-conditions, because once they're added and plugin creators could theoretically have made use of them, they're impossible to remove even if the base game code never used them.

tehhowch:
How about a compromise to drop the TypeCount, and keep the separate tallies of fines vs mortgages? After all, one comes from being shady, and the other from being an aspiring entrepreneur - quite different situations (though sometimes not :wink: )

I'll push that change shortly

Pointedstick:
If you wrote a mission with a condition `unpaid mortgages == 0`, wouldn't it trigger every time the player paid off a new mortgage--not just the first time?

tehhowch:
Only if it is given repeat

Pointedstick:
Oh of course, that makes sense.

PR #3166, opened 2017-10-28
Update README.md

agassert:
Added image, contributing section, and license section.

See result [here](https://github.com/agassert/endless-sky/blob/6862e99cfd8ee395e0b85fa2b13f8c727e697a2e/README.md).


tehhowch:
Why reference an offsite image instead of a self-hosted one?

gunqqer:
Yeah I agree with tehhowch, there's no reason to have the image hosted offsite

agassert:
The README now references https://camo.githubusercontent.com/4f74d02549bd6f37404b1238088e450caba19451/68747470733a2f2f692e696d6775722e636f6d2f5a766e756958592e6a7067.

tehhowch:
Is it not possible to reference a file in the repository?

shadowfacts:
It is. If you stick the image inside the repo itself, you should be able to reference it through its raw.githubusercontent.com URL.

agassert:
Do we want to increase the size of the game for a file that is never used in-game? If we do take this route, should the image be in the root of /images or should a new directory be created in /images for this file?

shadowfacts:
Perhaps stick it in a `.github` folder or something similar that's not included with the build?

tehhowch:
Then PR the image to the website repo, and link to its location there: https://github.com/endless-sky/endless-sky.github.io

agassert:
Once the site PR is merged I'll change the image source here to reference the image in that repo directly.

endless-sky:
The "CONTRIBUTING.md" file ought to be shown automatically whenever anyone posts and issue or pull request, so I'm not sure if it's necessary to also link it here. I'm also not convinced that a "header image" just for decoration is necessary.

Issue #3156, opened 2017-10-23
Source or Destination: near <government> filter


Amazinite:
Similar to the `near <system>` source and destination filter, only instead of a specific system that the source or destination must be a distance from, it is a specific government.
Example use case: a job or repeating mission which is designed to have a destination in FW space which is at most 2 systems away from a Republic system, but does not need changed due to the increase in FW territory over story events.

Additionally, should #3144 be added, perhaps `near <system attribute>` would be a useful source and destination filter as well.

tehhowch:
As I mentioned on Discord, I don't think the `near` keyword is a good fit for this, as that will require complicating the LocationFilter::LoadChild method (requiring several new comparisons to the size of the DataNode, and if arguments are numeric)

Perhaps a better construct would be a keyword `neighbor`. One implementation could be:
 - only 1 argument is allowed
 - a match is any system for which at least one of the systems accessible by system->links has the argument as its government.
  - Note that this includes all non-isolated systems of the specified government, so to filter to only systems of different government that have a hyperspace border would require using a `not government <argument>` filter as well.

This construct could be abstracted to support
 - `neighbor government <government>`
 - `neighbor attribute <system attribute>`

Use of the keyword `not` with the above construct would match any system that has no links to a system with the specified government, and could include isolated systems of the named government (so combination with a `not government <argument>` filter is again recommended if the intent is to select only non-argument systems).

tehhowch:
Even better might be to make `neighbor` load a nested LocationFilter that must be true for (or `not` true for each) neighbor. Then support for additional properties with the neighbor criteria is automatic.

endless-sky:
This would make the location filters significantly more complicated, for very little benefit - you could get nearly the same effect by just making missions that have Republic sources, FW destinations, and a minimum and maximum distance between them.

Amazinite:
The problem is that the jobs are suppose to have FW sources and destinations, but the destination needs to be near the Republic-FW border, which changes throughout the story, and the source of a job can not shift based off of conditions.

tehhowch:
@endless-sky I don't particularly feel that neighbor matching is much more complex than the ability to specify what a system or planet must not be.
I applied the logic from the notFilter implementation in #3173, and if we are smart about the order of checking filters, we can avoid "expensive" checks bring done on many possible matches (i.e. first match based on qualities of the actual picked thing, like government or attributes or distance, then check the nested LocationFilters for `not`and `neighbor` compliance).

[...]

PR #3173, opened 2017-10-29
Allow LocationFilter to specify attributes of a neighboring system

tehhowch:
Refs #3156 

 - Neighboring systems are those that are linked, and not those that are reachable solely by Jump Drive, to prevent matching to possibly inaccessible systems. (Same behavior as `near` or `distance` for existing planet & system LocationFilters) 

Example use:
```
source
    neighbor government "Republic"
    not neighbor government "Quarg" "Independent" "Pirate" "Neutral"
```
The neighbor filter only applies system-level checking, which means without #3144 use of `attributes` (e.g. `neighbor attributes "dirt belt"`) does not make sense.


Amazinite:
So is a distance modifier as asked for in #3156 impossible or simply something you do not wish to do?

tehhowch:
You could certainly nest `neighbor`:
```
neighbor
    neighbor government "Republic"
    not government "Republic"
```
That says that this system's neighbor must neighbor a republic government system, and must itself not be a republic government system.

Amazinite:
I don't know why but that makes me laugh. So if i wanted to have a source/destination that was within 2 systems of a Republic system while not itself being a Republic system, I would need to do
```
neighbor government "Republic"
neighbor not government "Republic"
neighbor
    neighbor government "Republic"
    not government "Republic"
```
?

tehhowch:
If you want a source/destination that 1) is not a Republic system, 2) does not neighbor a Republic system, and 3) has a neighbor that does neighbor a Republic system:
```
source
    1) not government "Republic"
    neighbor
        2) not government "Republic"
        3) neighbor government "Republic"
```
is sufficient (if not redundant, as the source does not need to indicate it must not be in a Republic system since the source's neighbor is required to not neighbor a Republic system. For clarity though (and speed when matching, since neighbors are checked last), it is sensible to keep the 1st filter even though the 2nd will also enforce it.

tehhowch:
@Amazinite you are correct that it gets complicated to move outward to look at Nth nearest neighbors, and rightfully so. There are a lot of combinations that this has to evaluate.
That said, it isn't impossible to do with this feature, while it is impossible to do without.
for a system whose 3rd nearest neighbor is a republic system (and no closer ones are either):
```
destination
    not government "Republic"
    not
        neighbor government "Republic"
    not
        neighbor
            neighbor government "Republic"
    neighbor
        neighbor
            neighbor government "Republic"
```

tehhowch:
Here's 2 test missions that use the above 2nd and 3rd neighbor filters:
[NeighborFilterTesting.txt](https://github.com/endless-sky/endless-sky/files/1425516/NeighborFilterTesting.txt)

source/LocationFilter.cpp
```diff
@@ -180,6 +206,9 @@ bool LocationFilter::Matches(const Planet *planet, const System *origin) const
 	if(!governments.empty() && !governments.count(planet->GetGovernment()))
 		return false;
 	
+	if(!MatchesNeighborFilters(neighborFilters, planet->GetSystem(), origin))
+		return false;
```
endless-sky: Is this necessary here? It seems to me that neighbor filters will only ever apply to systems, not to planets, so it's sufficient for Matches(System...) to check the neighbors.

tehhowch: I'll investigate, but I think this is needed in case the neighbor filter is being checked a a part of a source, destination, or stopover filter.

endless-sky: You should be able to just modify Matches(System...) so that it checks the neighbor filter whether or not `didPlanet` is true. The only checks in that function that are gated behind `!didPlanet` are the ones that could apply either to a planet or to a system.

PR #3176, opened 2017-10-30
Turtleroku ijs conversation

Turtleroku:
I changed the sentence about mixes of isotopes slightly, I think it sounds a little better.


tehhowch:
PRs will submit all differences between the pulled branch and the destination that are not already added or rejected. This is why most contributors keep a branch that is a 1:1, identical copy of the endless-sky/master branch. Then, to make changes we first checkout that branch (`git checkout master`), pull in updates (`git pull endless-sky master`), create a new branch (now based on the most up-to-date version of this repo) via `git checkout -b <branch_name>`, and then make the desired changes in new commits.

Then when you want to go back to a personal build, you can just `git checkout <my_branch>`

PR #3180, opened 2017-10-30
Mission changes

Turtleroku:
line 922: unless the planet only had two continents; 'would drop' sounds like the temperatures were willingly dropping rather than passively dropping
line 930: last sentence removed 'That' to make the statement more clear.
line 1069: removed 'on' as it is makes the sentence sound off. An alternative, "For now our best bet would be to search the nearby mining planets"
line 1119: proper underground testing releases no radioactive material above ground. 'thrown up' gives the explosion personification. An alternative would be to replace 'thrown up' with 'resulted in'
line 1225: 'full of radiation' is like a lamp full of light, and Eyes would be measuring the radiation emitted by the radioactive isotopes anyways. An alternative would be to say 'full of radioactive material'

### DIFF
--- a/data/free worlds intro.txt
+++ b/data/free worlds intro.txt
@@ -1116,7 +1116,7 @@ mission "FW Katya 6B"
 		has "chosen sides"
 	on offer
 		conversation
-			`Mr. Eyes fires up his instruments and looks at the results for a few minutes. "Nothing," he says. "Even underground testing would've thrown up a debris plume large enough to measure traces of it. But I'm picking up nothing."`
+			`Mr. Eyes fires up his instruments and looks at the results for a few minutes. "Nothing," he says. "Even underground testing would've released enough trace elements to get a reading. But I'm picking up nothing."`
 			`	"No problem," says Katya, "let's continue to <planet>."`
 				accept
 
--- a/data/free worlds intro.txt
+++ b/data/free worlds intro.txt
@@ -919,7 +919,7 @@ mission "FW Katya 4"
 				`	"Was anyone killed?"`
 				`	"How did you escape?"`
 			
-			`	"The eruption was on the other continent," she explains, "so no one was killed in the blast. It was a slow-moving disaster. We had six months or so to move two million settlers off-world before the temperatures would drop too low to be livable. More than half of the Navy was mobilized to ferry our people to other planets."`
+			`	"The eruption was on another continent," she explains, "so no one was killed in the blast. It was a slow-moving disaster. We had six months or so to move two million settlers off-world before the temperatures dropped too low to be livable. More than half of the Navy was mobilized to ferry our people to other planets."`
 			`	"Here's the best part," says Eyes. "The Navy operates with a certain amount of autonomy. Turns out when they found out about the disaster, they mobilized without even waiting to clear it with Parliament. Which was technically within their mandate, this being a natural disaster. But the Navy is funded by taxes, and the Paradise Planets pay ten times as much in taxes as the Dirt Belt."`
 			`	"The Paradise Planets viewed the Navy as their own private security force," says Katya, "so they got really mad when half the Navy suddenly vanished to assist some dirty refugees. Because those admirals gave the order to assist us, millions of lives were saved. And then Parliament fired them."`
 			choice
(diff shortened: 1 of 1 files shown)


jafdy:
Not to sure about L1119, your change makes it worse for me. 
L1225 could be taken further, it reads as the crater is radioactive. As if the crater is more than a large hole in the ground (am I making sence?).

data/free worlds intro.txt
```diff
@@ -919,15 +919,15 @@ mission "FW Katya 4"
...
-			`	"I was a child on Vinci while this was happening," says Eyes. "A paradise world. I would listen to the news and could never understand why all the adults around me were so very, very angry that the Navy had gone off to rescue two million people from freezing to death. But they told me that we lived in one little island of safety in a dangerous galaxy, and that without the Navy protecting us the pirates or even maybe aliens would sweep in and destroy our homes and take our money. That it was us against the whole galaxy."`
+			`	"I was a child on Vinci while this was happening," says Eyes. "A paradise world. I would listen to the news and could never understand why all the adults around me were so very, very angry that the Navy had gone off to rescue two million people from freezing to death. But they told me that we lived in one little island of safety in a dangerous galaxy, and that without the Navy protecting us the pirates or even maybe aliens would sweep in and destroy our homes and take our money. It was us against the whole galaxy."`
```
tehhowch: He is recounting something his parents said to him, so use of `That` is IMO preferred. Removing it changes the meaning of the last sentence.

data/free worlds intro.txt
```diff
@@ -1066,7 +1066,7 @@ mission "FW Katya 5C"
 			`	"A few. I talked to one friend who had done research into nuclear development 'for purely academic reasons,' and he said that only a fool would perform atmospheric tests; Intelligence would notice that immediately, no matter what world you were on. He said the ideal development location would be a sparsely populated planet with some abandoned deep underground mines."`
 			`	"Anything else?" you ask.`
 			`	"Just one more thing," says Greg. "Several people mentioned that they think the Navy itself may have a nuclear arsenal, and that a renegade may have stolen the weapons from them. Much easier than developing them yourself."`
-			`	"Well, let's head back to <planet> and see what Eyes thinks," says Katya, "but I can't imagine the Navy would have such an arsenal. Searching on mining planets sounds like our best bet for now."`
+			`	"Well, let's head back to <planet> and see what Eyes thinks," says Katya, "but I can't imagine the Navy would have such an arsenal. Searching mining planets sounds like our best bet for now."`
```
tehhowch: It's character speech. Using `on` is perfectly fine.

data/free worlds intro.txt
```diff
@@ -1116,7 +1116,7 @@ mission "FW Katya 6B"
 		has "chosen sides"
 	on offer
 		conversation
-			`Mr. Eyes fires up his instruments and looks at the results for a few minutes. "Nothing," he says. "Even underground testing would've thrown up a debris plume large enough to measure traces of it. But I'm picking up nothing."`
+			`Mr. Eyes fires up his instruments and looks at the results for a few minutes. "Nothing," he says. "Even underground testing would've released enough trace elements to get a reading. But I'm picking up nothing."`
```
tehhowch: Character speech - Eyes' words and phrasing don't need to be clinical.

data/free worlds intro.txt
```diff
@@ -919,15 +919,15 @@ mission "FW Katya 4"
 				`	"Was anyone killed?"`
 				`	"How did you escape?"`
 			
-			`	"The eruption was on the other continent," she explains, "so no one was killed in the blast. It was a slow-moving disaster. We had six months or so to move two million settlers off-world before the temperatures would drop too low to be livable. More than half of the Navy was mobilized to ferry our people to other planets."`
+			`	"The eruption was on another continent," she explains, "so no one was killed in the blast. It was a slow-moving disaster. We had six months or so to move two million settlers off-world before the temperatures dropped too low to be livable. More than half of the Navy was mobilized to ferry our people to other planets."`
```
Amazinite: >unless the planet only had two continents

How do you know it doesn't? Earth has had two continents at times, even one. It's even hypothesized that when Earth had only one or two continents that were located in just the right position, this helped to cause a global ice age (Snowball Earth theory).

Turtleroku:
Changed the commits back to the original version. Line 1119 (mine or the original) just feels off for some reason.

endless-sky:
I'm not sure that either of these changes is really an improvement. Especially the first one - the original makes it much clearer that the settlers, in the past that they are describing, were looking forward at a future even that was going to happen, i.e. that the temperatures would drop.

Turtleroku:
It's up to you and depends on what you were wanting to convey. I would just say that Katya was in the present talking in past tense, i.e. using 'had' and 'mobilized' but 'would drop' is future tense not past. 
Also I am a scientist so my definitions of words are slightly different from the mainstream, and this game is designed for the majority (as it should be).
Just trying to be helpful and make the game better in any way I can. I think it is telling that the game is well written as in all of FW intro I only found 5 commits. 4 of which were subjective changes.

Issue #2333, opened 2017-03-29
Charge fighter shields in parallel?


lheckemann:
Currently, as far as I can tell, fighters that are being carried have their shields charged one by one, i.e. the shields of a second fighter don't start getting charged until the first is completely done. I think it would be more helpful to distribute the shield-charging capacity evenly across all the fighters once the carrier's shields are fully charged rather than charging them sequentially. Any thoughts?

gunqqer:
Would this go all get the same amount or lowest get regened firsted then when they're all even it is distributed?

KiLEdEnNis:
if ya have a lot of carriers regening fighters w/ different amounts of shields & all carriers have different amounts of fighter bays & shield regen, won't it be a lot harder for the CPU?

jafdy:
Most of the time it is so fast charging that it doesn't matter. This would only become useful if fighters that are low on shields auto dock to regenerate as this would cause them to re launch in groups.

endless-sky:
If fighters eventually get smart enough to stop at their motherships to repair and then launch again when ready, repairing the fighters one by one will presumably be a better strategy than repairing them in parallel.

Since it takes the same amount of time to repair all of them anyway, the only option that would make more sense would be to prioritize the most damaged fighter (which would automatically mean recharging them in parallel once they're at the same amount of damage).

EndrosG:
If talking about prioritizing [staying with the sequential recharge/repair] I would prefer to prioritize the least damaged fighter. The faster any (single) fighter can return to battle the better is. I rather recharge 3 fighters from 75% to 100% (3x 25% = 75% of charging) instead of recharging the most damaged fighter which is down to 10% (i.e. 90% charging).

tehhowch:
To address @endrosg 's comment, it could be a preference to choose the sort behavior of the BayOrder method in #3182, eg most vs least damaged

endless-sky:
Commit 24cab84 makes it so that fighters will retreat to the mothership if badly damaged, and only launch again if they are mostly repaired.

Given those new mechanics, the optimum strategy with fighters is to maximize how many are out fighting at a given time. That means that repairing the *least* damaged one first (in order to get one more fighter out on the field as quickly as possible) is the optimum strategy.

lheckemann:
[...]

PR #3182, opened 2017-10-30
Update recharging of carried ships

tehhowch:
- Prioritize recharging by absolute need (parallel strategy): capacity - current ( https://github.com/endless-sky/endless-sky/issues/2333#issuecomment-291005200 )
 - Implement recharging of carried ships' fuel
   - Carried ships do not utilize any ramscoop power they may have.
 - Use of the carried ship's own shield & hull repair is no longer free
   The parent must have enough energy and absorb the heat, and if not, the activated amount is scaled down to match the available energy.

Recharging of carried ships by the carrier was added in 08612b1ebe92a98e2b2b3bf53c3a8f8fb1121aa4 , and included the regen of the carried ship's own shield and hull repair, but did not account for the cost of using those functions.


tehhowch:
Added a preference so the player can choose the behavior they want:
image
Toggles between `most need` and `least need`, addressing @EndrosG 's https://github.com/endless-sky/endless-sky/issues/2333#issuecomment-291353212

endless-sky:
This is (mostly) superseded by commit 24cab84, which makes fighters return for repairs if badly damaged and prioritizes repairing the least damaged ones (in order to maximize the number of fighters that are in the battle at any given time).

Issue #3179, opened 2017-10-30
Espionage mission after "missing" the initial scan


flaviojs:
I picked up an espionage mission while doing other missions, probably because it was in the path to my final destination:

```
mission "Corporate espionage"
	name "Corporate espionage"
	description "The Serenity Empress is about to enter this system. Locate it and make detailed scans of its cargo, which are of considerable interest to a team of Syndicate researchers on Bluestone in the Bellatrix system; proceed there for 50,000 credits after retrieving the data."
	job
	repeat 0
	to offer
		random < 10
	destination Bluestone
	npc
		succeed 2
		government Merchant
		personality
			confusion 10
			entering
			fleeing
			target
			uninterested
		dialog
			"You manage to make the requested scans of the Serenity Empress's cargo. Time to head to Bluestone."
		ship Bounder
			name "Serenity Empress"
			sprite ship/bounder
			attributes
				category Transport
				cost 1140000
				bunks 17
				"cargo space" 40
				drag 3.7
				"engine capacity" 110
				"fuel capacity" 900
				"heat dissipation" 0.8
				hull 700
				mass 130
				"outfit space" 220
				"required crew" 1
				shields 2200
				"turret mounts" 2
				"weapon capacity" 50
			outfits
				Hyperdrive
				"LP036a Battery Pack"
				"D14-RN Shield Generator"
				"X3700 Ion Thruster"
				"X3200 Ion Steering"
				"nGVF-CC Fuel Cell"
				"Blaster Turret" 2
			cargo
				commodities
					"Heavy Metals" 4
					Industrial 2
					Medical 34
			crew 4
			fuel 900
			shields 2200
			hull 700
			position 1360.243 213.6807
			engine -12 44 1
			engine 12 44 1
			turret -37 4 "Blaster Turret"
			turret 37 4 "Blaster Turret"
			explode "tiny explosion" 10
			explode "small explosion" 20
			explode "medium explosion" 15
			"final explode" "final explosion small" 1
			system Vindemiatrix
	on complete
		dialog
			`You transmit your scans and receive a hearty congratulations, as well as 50,000 credits. "This will save us months!" you hear someone say in the background as you close the audio link.`
		payment 50000
```

I have no idea what "this system" was and there is no other indication in the map, so I went to Bellatrix system (destination) and waited for the ship there, eventually giving up.
I inspected the save file while doing other missions and the ship appears to be jumping and landing randomly.

**What should happen after the player "misses" the initial scan?** 

[...]

PR #3186, opened 2017-10-31
Fail mission when ship flees with unfulfilled actions

flaviojs:
Reference: option 3 of #3179

~~The event introduced in f5091982de0c9c882a7280295de05fd27851a240 was never used. It's was probably meant to reach the player mission npcs so I made sure all events of the event queue reach them, which includes the new event `ShipEvent::FLEE`.~~

### DIFF
--- a/data/syndicate world jobs.txt
+++ b/data/syndicate world jobs.txt
@@ -74,6 +74,8 @@ mission "Corporate espionage"
 	on complete
 		dialog `You transmit your scans and receive a hearty congratulations, as well as <payment>. "This will save us months!" you hear someone say in the background as you close the audio link.`
 		payment 50000
+	on fail
+		dialog "You failed to make the requested scans of the <npc>'s cargo."
 
 
 mission "Document delivery to <planet>"
--- a/data/syndicate world jobs.txt
+++ b/data/syndicate world jobs.txt
@@ -75,7 +75,7 @@ mission "Corporate espionage"
 		dialog `You transmit your scans and receive a hearty congratulations, as well as <payment>. "This will save us months!" you hear someone say in the background as you close the audio link.`
 		payment 50000
 	on fail
-		dialog "You failed to make the requested scans of the <npc>'s cargo."
+		dialog "You failed to make the requested scans of the <npc>'s cargo before it could deliver them."
 
 
 mission "Document delivery to <planet>"
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1255,6 +1255,9 @@ void Engine::MoveShip(const shared_ptr<Ship> &ship)
 		// self-destruct.
 		if(ship->IsDestroyed())
 			eventQueue.emplace_back(nullptr, ship, ShipEvent::DESTROY);
+		// Record that the ship was able to flee (by landing).
+		else
+			eventQueue.emplace_back(nullptr, ship, ShipEvent::FLEE);
 		return;
 	}
 	
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1256,9 +1256,14 @@ void Engine::MoveShip(const shared_ptr<Ship> &ship)
 		// self-destruct.
 		if(ship->IsDestroyed())
 			eventQueue.emplace_back(nullptr, ship, ShipEvent::DESTROY);
-		// Record that the ship was able to flee (by landing).
-		else
+		// The ship was able to flee. Carried ships only flee with the parent.
+		else if(!ship->CanBeCarried() || !ship->GetParent() || ship->GetParent()->ShouldBeRemoved())
+		{
 			eventQueue.emplace_back(nullptr, ship, ShipEvent::FLEE);
+			for(auto &bay : ship->Bays())
+				if(bay.ship)
+					eventQueue.emplace_back(nullptr, bay.ship, ShipEvent::FLEE);
+		}
 		return;
 	}
 	
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1248,6 +1248,10 @@ void Engine::MoveShip(const shared_ptr<Ship> &ship)
 	// Bail out if the ship just died.
 	if(ship->ShouldBeRemoved())
 	{
+		// Regular ships are silent about being removed.
+		if(!ship->IsSpecial())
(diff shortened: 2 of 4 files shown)


flaviojs:
NOTE: I can't think of any situation where the flee event should not be created. However, the probability of such a situation existing is very high.

source/Engine.cpp
```diff
@@ -1252,6 +1255,9 @@ void Engine::MoveShip(const shared_ptr<Ship> &ship)
 		// self-destruct.
 		if(ship->IsDestroyed())
 			eventQueue.emplace_back(nullptr, ship, ShipEvent::DESTROY);
+		// Record that the ship was able to flee (by landing).
+		else
```
tehhowch: This should not be an `else` unless you want
 - every random NPC that lands to throw a FLEE event
 - every random NPC that is out of system for 1000 frames to throw a FLEE event
 - every fighter that docks with its carrier to throw a FLEE event

source/Engine.cpp
```diff
@@ -352,6 +352,9 @@ void Engine::Step(bool isActive)
 		else if(jumpCount > 0)
 			--jumpCount;
 	}
+	// Player mission npcs can react to events.
+	for(const ShipEvent &event : events)
```
tehhowch: Is this not already done?
https://github.com/endless-sky/endless-sky/blob/9c1f20ade3dd8e4e3151fe8021def5a0ed8d11ae/source/MainPanel.cpp#L100-L106

data/syndicate world jobs.txt
```diff
@@ -74,6 +74,8 @@ mission "Corporate espionage"
 	on complete
 		dialog `You transmit your scans and receive a hearty congratulations, as well as <payment>. "This will save us months!" you hear someone say in the background as you close the audio link.`
 		payment 50000
+	on fail
+		dialog "You failed to make the requested scans of the <npc>'s cargo."
```
tehhowch: `"... before it could deliver them."`

source/Engine.cpp
flaviojs: I see... I guess at the time I was making this the FLEE events weren't arriving at NPC for another reason.

source/Engine.cpp
flaviojs: For drones/fighters that dock, they shouldn't throw a FLEE until the carrier throws a FLEE. Will change.

Other than that particular case I believe being removed from the game while still alive is a valid "flee" event since you won't be seeing the same ship again. But this will probably be a performance bottleneck at some point  `O(flees*missions*npcs)`. Ship should probably have a variable that indicates which events it can trigger.

data/syndicate world jobs.txt
flaviojs: Ok.

source/Engine.cpp
tehhowch: In #2990 I used two ship methods (1 new):
1) `HasLanded() - returns a property that gets set upon landing, which means fighters and disappearances can't throw the event
2) `IsSpecial()` - if the ship isn't special, a landing event is not of any importance, so don't throw it then either

flaviojs: Then if the ship isn't special, it should not throw a DESTROY either.

tehhowch: I believe there are rep hits associated with destruction events, so I wouldn't go that far

flaviojs:
This PR can fail other missions so I'll rename it.

Example: mission with a "scan outfits" npc that has a `fleeing` personality
A FLEE event means the ship was removed from the game alive. The `fleeing` personality removes the ship from the game when it lands (alive). When that happens, the mission fails if the player still had to perform the "scan outfits" action on that ship.

source/Engine.cpp
```diff
@@ -1248,10 +1248,22 @@ void Engine::MoveShip(const shared_ptr<Ship> &ship)
...
+		// Regular ships are silent about being removed.
+		if(!ship->IsSpecial())
+			return;
+		
 		// Make sure this ship's destruction was recorded, even if it died from
 		// self-destruct.
 		if(ship->IsDestroyed())
 			eventQueue.emplace_back(nullptr, ship, ShipEvent::DESTROY);
+		// The ship was able to flee. Carried ships only flee with the parent.
+		else if(!ship->CanBeCarried() || !ship->GetParent() || ship->GetParent()->ShouldBeRemoved())
```
tehhowch: `!ship->GetParent()` can only be true for:
 - uninterested NPCs
 - NPCs whose flagship was destroyed since the player last took off,

since otherwise the result of `GetParent()` is the player. A `personality fleeing` ship would thus never fire this event, even it landed on a planet, since the player is still alive. In a multi-ship NPC spec, the event would fire for each non-NPC flagship, but not the NPC flagship.

source/Engine.cpp
```diff
@@ -1248,10 +1248,22 @@ void Engine::MoveShip(const shared_ptr<Ship> &ship)
...
+		{
+			eventQueue.emplace_back(nullptr, ship, ShipEvent::FLEE);
+			for(auto &bay : ship->Bays())
```
tehhowch: `const Ship::Bay` isn't too long to write :)

tehhowch:
I can't really recommend this PR as it is a conflicting implementation of something done in #2990 with ShipEvent::LAND
That is, #2990 achieves everything in this PR as a natural part of giving mission NPCs travel instructions.

source/Engine.cpp
flaviojs: The full check is there to filters carried ships that have a live parent, the combined inverse might be clearer:
`!(ship->CanBeCarried() && ship->GetParent() && !ship->GetParent()->ShouldBeRemoved())`

source/Engine.cpp
flaviojs: My style preference,  I used auto because it's an inner class of another class.
Not idea about the official stance regarding this.

tehhowch: http://endless-sky.github.io/styleguide/styleguide.xml#C++11 has a bit about use of `auto`

source/Engine.cpp
tehhowch: Something about "missing the forest for the trees" would be appropriate for me here :grinning:

Issue #851, opened 2016-03-12
Scanning: Government Jurisdiction


daerogami:
Feature Request: Ships only scan on behalf of their Government when in their own territory.

I have been flying around the southern sector in the "Free Worlds" systems and noticed there is a strangely large number of Republic Gunboats flying around this expanse of the galaxy.
Given the fallout between the Free Worlds and the Republic, why is the Republic enforcing their contraband laws outside of their domain? Even in uninhabited areas?

Related Issues: #556 #576

endless-sky:
In their defense, from the Republic's perspective the Free Worlds systems *are* their territory.

tehhowch:
The simplest method of doing this is to add a check inside AI::DoSurveillance that the current system is of the same government as the ship.

This could be expanded by defining a new node to the government definition, that identifies the system governments in which this government can exercise authority:
```
government "Republic"
    enforces
        "Republic"
        "Free Worlds"
        "Syndicate"
        not "Pirate"
        ...
```
This structure is just a LocationFilter, making its use and customization very intuitive for modders and other content creators

jjhankins:
The AI already doesn't scan hostile ships, so the Republic isn't going to scan pirate ships no matter where they are - though perhaps this should be changed? In practice, playing as a pirate, being scanned (by Captain Parrot) is more of a threat than being attacked. You can run from an attack, but if you're fined they just reach into your account and take the money out.

Hm. If fines are simply another avenue of enforcement (which is more true to the story, imo) then they'd fine anyone they're hostile enough with to attack, whether or not they carry contraband. This would open up some interesting story possibilities. Pirates would want to keep their funds out of reach of the republic, so give them a new outfit (a ship safe) and the ability to purchase jewels and coins, and suddenly there's the possibility of treasure caches.

tehhowch:
That filter was meant as an example of possible syntax. Note that defining an inclusive filter for one property means any exclusive filters on that same property that mention new values are automatically satisfied.

KiLEdEnNis:
@jjhankins the only thing I'd do with pirates toward the Republic & other factions is being able to accumulate a large bounty & when ya pay it off they become friendly again, but doing that per government and everything doesn't seem worth it

jjhankins:
[...]

Issue #1958, opened 2016-12-31
Republic Navy Behavior


VonZorch:
It seems odd to me that the Republic Navy, after the actual shooting in the civil war stops, is patrolling deep in Free Worlds space acting like they are in charge when they are clearly out of their jurisdiction.
Especially the constant scanning, Scanning a military ship in their home space by a foreign navy could easily be taken as a hostile act, and after the war my ships are still in Free World colors and I'm still getting paid.

DingusShingleton:
After the war you are no longer a military officer, the payment you receive is a stipend to honor your huge contribution to the war. But you are a civilian.

Amazinite:
How deep into Free Worlds space? I've only ever seen them around the border systems, and even then it's super rare for them to spawn.

endless-sky:
It seems odd to me that the US routinely has some combination of drones, spy satellites, planes, and cyber-espionage going on in countries we're not technically at war with, but that doesn't stop them from doing it. The Republic is basically the same way: they're big and powerful enough that they don't even have to hide the fact that they're engaged in ongoing surveillance of the Free Worlds, and there's very little that the Free Worlds (even after the armistice) can do about that.

Besides, if you extend maritime law to space, the way it would work is that anything not within some distance of an inhabited planet is "international waters." That is, there could be reasonable arguments for not allowing a government to restrict what ships are allowed in a given star system, but instead only to restrict who can land on their planets.

VonZorch:
I've had it happen is the Pherkad system, and every time I go through Aldhibain or Dschubba there's either a capital ship with 2-3 escorts or several Gunboats and Rainmakers.

KiLEdEnNis:
that's no surveillance fleet, and is a bug then (I've had that issue with Lesath, the empty system right of it, which is fixed)

VonZorch:
I must need to download a new version.  I get them in Lesath.  I didn't mention it, as being empty and near the border, I considered it neutral space.

KiLEdEnNis:
depends on the part in the storyline, I got it after all those systems joined and Lesath was like 4 jumps off of Republic space

VonZorch:
In my game Rastaban and Girtan, just off Lesath, are still Republic

PR #3187, opened 2017-10-31
Curtailed surveillance

tehhowch:
Refs #851 

This PR makes it possible to limit the scope of a government's ability to assess fines and perform scanning.
 - `enforces` keyword introduces a `LocationFilter` that describes a set of systems and planets for which a government will scan ships
 - multiple `enforces` are supported - the government will perform scans if any of the filters is matched
 - if no `enforces` filters are used, the current behavior (universal jurisdiction) is kept.
 - **Mission NPCs are always given universal jurisdiction**

example:
```
government "Syndicate"
    ...
    enforces
        government "Syndicate" "Uninhabited"
    enforces
        neighbor government "Syndicate" "Uninhabited"
```
The above would make Syndicate ships with the surveillance personality only perform scans in systems they own (the 1st `enforces`), or in systems that neighbor their own (the 2nd).

Implementation detail: if the game ever supports special actions outside of the player's system, then the `canScan` helper variable (used to limit the number of calls to `LocationFilter::Matches`) would need updating.


tehhowch:
Also refs: #1958

Amazinite:
Must a new `enforces` token be made for every new location that a government enforces in (as shown in the example), or could it be written in a list?
```
    enforces
        government "Wanderer"
        neighbor government "Wanderer"
```

tehhowch:
>Must a new enforces token be made for every new location that a government enforces in (as shown in the example), or could it be written in a list?

You'd need a new `enforces` node for each set of locations, since all children of a node are criteria of the LocationFilter.
I gave a kinda-poor example since `neighbor government "Wanderer"` matches every system that matches `government "Wanderer"` so long as it isn't a lone outpost in the galaxy (i.e. most systems of a given government neighbor other systems of the same government).
Your spec makes it so that only Wanderer systems that aren't a lone outpost (i.e. have a neighboring Wanderer system and are Wanderer themselves) perform enforcement.

PR #3195, opened 2017-11-03
Suspend the boarding approach when too far away or too fast.

flaviojs:
The boarding approach changes the ship angle, velocity, and position artificially (no engines). It is activated when you are close enough and slow enough.

After activating it, you can become too far away by selecting another target and you can become too fast by getting hit. With this commit the boarding approach will be temporarily suspended until you get close enough and slow enough again.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -3000,12 +3000,10 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 		keyStuck.Clear(Command::LAND);
 	if(keyStuck.Has(Command::JUMP) && !(ship.GetTargetSystem() || isWormhole))
 		keyStuck.Clear(Command::JUMP);
-	if(keyStuck.Has(Command::BOARD) && !ship.GetTargetShip())
+	if(keyStuck.Has(Command::BOARD) && !(ship.GetTargetShip() && CanBoard(ship, *ship.GetTargetShip())))
 		keyStuck.Clear(Command::BOARD);
 	
-	if(ship.IsBoarding())
-		keyStuck.Clear();
-	else if(keyStuck.Has(Command::LAND) || (keyStuck.Has(Command::JUMP) && isWormhole))
+	if(keyStuck.Has(Command::LAND) || (keyStuck.Has(Command::JUMP) && isWormhole))
 	{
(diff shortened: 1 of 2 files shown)


endless-sky:
I think it would make sense for boarding to just be aborted in that situation. I think that could be accomplished by just changing the `|=` in line 1298 an `=`:

```C++
isBoarding = (distance < 50. && speed < 1. && commands.Has(Command::BOARD));
```

flaviojs:
I don't agree. We already have a clear way to abort the boarding: any manual movement.

Aborting because I'm getting hit (become too fast) would be annoying. If I want to engage the enemy I'll start moving (aborts boarding) and if my escorts can handle it I want to continue boarding.

Aborting because I changed target can be interpreted as "wanting to abort". For me it would be frustrating since I change target because another ship will get there first, so I wanna board another one quickly before someone else gets there.

flaviojs:
Besides manual movement, I wouldn't mind if manual shooting (Tab) aborted boarding too.

endless-sky:
The purpose of the `isBoarding` variable is to let the AI know that a ship has succeeded in "locking on" to its target. For example, once `isBoarding` is true, the [AI for assisting a ship](https://github.com/endless-sky/endless-sky/blob/master/source/AI.cpp#L469-L473) will not make any further efforts to move to the ship's location. So, your change as it stands would make a ship freeze and do nothing if knocked away while trying to assist another ship.

But, [elsewhere in the AI class](https://github.com/endless-sky/endless-sky/blob/master/source/AI.cpp#L3006-L3007) the code assumes that once a ship has begun that locked-on "boarding" process, it's always going to succeed (so the BOARD command can be cleared). I think the best way to address both those issues would be to remove the two lines that clear `keyStuck` and to make the change I suggested to isBoarding (having it be set only if the distance criterion is still met) rather than the change you've got here.

flaviojs:
Was able to reproduce the "freeze".
Reimplemented as you suggested (set `isBoarding` and don't clear `keyStuck` when boarding).

Everything appears to be fine. I'm not sure about the change of boarding target to the parent in `Ship::Move`, but it should be working the same as before.

Btw, your links are for the master branch (not the current commit) so the line contents will change.

PR #3205, opened 2017-11-06
Small tweaks to the particle cannon projectile and effect

AzureNyoki:
The projectile now looks less flat, and the impact effect now has a smooth fade out, and isn't left hanging in space.


tehhowch:
In the future note that you don't want to merge your feature branch into your master branch.. just set up the PR from your feature branch. Otherwise every PR you make is going to include possibly rejected, possibly duplicated commits, and likely unrelated changes.

AzureNyoki:
Noted, thank you. I'm still learning the ins and outs of this.

endless-sky:
The trouble with bright sprites moving very quickly is that rather than looking like a continuously moving object, they can look to your eye like a dashed line due to persistence of vision. For example, here is what several particle cannon frames look like when overlaid, using the original sprite (left) and yours (right):

particle

To my eyes, at least, the effect is really obvious with your new sprite.

As for the impact effect, I think the real issue is just that the lifetime is too long given the frame rate (it reaches the final frame of the animation after 15 frames, then stays around for another 20). But that could be fixed just be reducing the lifetime from 35 frames down to 20 or so.

KiLEdEnNis:
wouldn't that mean it'd get either a shorter range or more velocity to compensate?

Bladewood:
It's just a visual artifact, nothing more.

KiLEdEnNis:
well if the lifetime would go down the range would automatically go down, unless I misunderstood something

SirStigII:
They are talking about the impact frame life I believe, not the projectile.

KiLEdEnNis:
ah you're right ("As for the impact effect"), my bad

PR #3208, opened 2017-11-06
Refactor PlayerInfo::Land

tehhowch:
- Use StepMissions() to handle changes to active missions - recording NPC self-destructs, completions, failures, etc.
 - Planets for which mission clearance has been given (or if infiltrating) will not fine the player ( Refs #1452 )

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -171,28 +171,26 @@ void Mission::Load(const DataNode &node)
 		else if(child.Token(0) == "destination")
 			destinationFilter.Load(child);
 		else if(child.Token(0) == "waypoint" && child.Size() >= 2)
-			waypoints.insert(GameData::Systems().Get(child.Token(1)));
-		else if(child.Token(0) == "waypoint")
+		{
+			set<const System *> &set = (child.Size() >= 3 && child.Token(2) == "visited")
+					? visitedWaypoints : waypoints;
+			set.insert(GameData::Systems().Get(child.Token(1)));
+		}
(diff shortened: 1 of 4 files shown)


tehhowch:
bf64016 is needed since performing the stopover action check would clear the knowledge that clearance was given for the planet.
It also exposes functionality for reminding a player of the planets or systems that needed to be visited. I'll save any UI updates for a separate PR though. I feel conveying visited waypoints would certainly be helpful for the player, especially when given tasks that involve checking a number of systems for possibly hard-to-spot things (like disabled Kor Sestor drones that were flung far from the system center...).

endless-sky:
It seems odd to me for "stopover" to use a completely different syntax than "completed stopovers," i.e.:

```python
stopover "Earth"
stopover "Luna"
"completed stopovers"
    "Mars"
    "Prime"
```

It would make more sense for them to share the same format. And, if doing that anyway, the completion status could just be an optional third field in the `stopover` line, e.g.:

```python
stopover "Earth"
stopover "Luna"
stopover "Mars" visited
stopover "Prime" visited
```

(When loading, you'd just check for that third field before deciding which set to put the stopover / waypoint into.) This would avoid requiring a two-word keyword (`"completed stopover"`). Mission creators could also use this syntax to mark stopovers that are "hints" only, i.e. you don't actually have to go there to complete the mission.

tehhowch:
Rewrote the serialization method. I'll save the UI updates for visited stopovers/planets for a different PR, as that's more of a Mission/MapPanel change than PlayerInfo

PR #3209, opened 2017-11-06
Modified description of Sek Alarfarat

ph2000bis:
Appart from fixing a mistake, the description misleadingly lead to
believe that the mantle of a planet is liquid, which is not the case.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -26406,7 +26406,7 @@ planet "Secret Sky"
 planet "Sek Alarfarat"
 	attributes korath
 	landscape land/bwerner2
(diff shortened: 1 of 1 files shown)


endless-sky:
Huh. I actually didn't know that.

But, maybe something like "...so deep that pools of magma bled out..." would work better. The intention was not that the weapon was hot enough to liquefy that much rock, but just that it was powerful enough to "crack open" the world's crust.

ph2000bis:
My thought was that as the mantle is made bare, the sudden loss of pressure would allow for the mantle rocks to enter fusion.

> so deep that pools of magma bled out
works fine, and avoids giving unnecessary geology info

PR #2880, opened 2017-08-01
Updated the Hai ships' descriptions

Kryes-Omega:
The original single-sentence description for the Solifuge was a bit lackluster, so I expanded what it by adding a little bit more detail. 
Read the change and tell me what you think I guess
Edit: I added changes for some of the other Hai ships as well, on @Amazinite suggestion

PR #3215, opened 2017-11-09
Expanded the descriptions of the 4 main Hai ships

Amazinite:
Reference: #2880 

I tried to follow the advice given by @endless-sky on making ship descriptions:
>In general, I don't mind making ship descriptions wordier for the sake of adding new lore or funny or interesting information, but expanding the description just to tell the player things they can see easily from the stats (like how many turret mounts it has, or the fact that it has a lot of cargo space) doesn't make sense to me.

I feel that I did a fair job, but everyone else can be the judge of that.

### DIFF
--- a/data/hai ships.txt
+++ b/data/hai ships.txt
@@ -45,7 +45,7 @@ ship "Aphid"
 	explode "tiny explosion" 10
 	explode "small explosion" 20
 	explode "medium explosion" 15
-	description "The Aphid is the lightest freighter used by the Hai and was for many millennia the backbone of the Hai inter-planetary economy. Given that a captain does not have to fear being attacked in much of Hai space, Aphids are only lightly armed and mainly rely on larger Hai ships for protection."
+	description "The Aphid is the lightest freighter used by the Hai and was for many millennia the backbone of the Hai inter-planetary economy. Because of Hai space being peaceful for much of history, Aphids are only lightly armed and mainly rely on larger Hai ships for protection from piracy."
 
 ship "Aphid" "Aphid (Anti-Missile)"
 	outfits
@@ -137,7 +137,7 @@ ship "Lightning Bug"
 	explode "medium explosion" 24
 	explode "large explosion" 8
 	"final explode" "final explosion small"
-	description "The Lightning Bug is a small defensive ship, mostly used by the Hai as a convoy escort. The Lightning Bug, as with many insects, works best in a swarm; even a small number of Lightning Bugs can manage to leave the largest of ships helpless with their stock-fitted ion cannons."
+	description "The Lightning Bug is a small defensive ship, mostly used by the Hai as a convoy escort. The Lightning Bug, as with many insects, works best in a swarm: even a small number of Lightning Bugs can manage to leave the largest of ships helpless with their stock-fitted ion cannons."
 
 ship "Lightning Bug" "Lightning Bug (Pulse)"
 	outfits
@@ -273,7 +273,7 @@ ship "Shield Beetle"
 	explode "large explosion" 50
 	explode "huge explosion" 20
 	"final explode" "final explosion large"
-	description "This ship was first built millenia ago near the end of a more violent time in Hai history, and since then it has been the only heavy warship at the disposal of the Hai. The Shield Beetle is a versatile warship used by the Hai to guard their merchant convoys, or by the Unfettered to attack those very same convoys."
+	description "This ship was first built millenia ago, near the end of a more violent time in Hai history. Since then it has served as the only heavy warship at the disposal of the Hai. The Shield Beetle is a versatile warship used by the Hai to guard their merchant convoys, or by the Unfettered to attack those very same convoys."
 
 ship "Shield Beetle" "Shield Beetle (Tracker)"
 	outfits
@@ -426,7 +426,7 @@ ship "Solifuge"
 	explode "large explosion" 50
 	explode "huge explosion" 20
 	"final explode" "final explosion large"
-	description `Impressed by the Alpha's "Giftbringer", the Unfettered endeavoured to build their own fighter carrier to take more advantage of their supply of Jump Drives.`
+	description `Impressed by the Alpha's "Giftbringer", the Unfettered endeavoured to build their own fighter carrier to take more advantage of their supply of jump drives.`
 
 ship "Solifuge" "Solifuge (Pulse)"
 	outfits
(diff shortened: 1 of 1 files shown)


data/hai ships.txt
```diff
@@ -620,7 +620,7 @@ ship "Water Bug"
 	explode "large explosion" 20
 	explode "huge explosion" 5
 	"final explode" "final explosion medium"
-	description "The Water Bug is the primary freight ship used by the Hai."
+	description "The desgin of the Water Bug is almost as old as the Aphid, but until recently it never saw much use. A spike in aggression from the Unfettered Hai has lead the tougher and more well defended Water Bug to see an increase in popularity over its little brother, the Aphid. The Water Bug is now the most widely used freighter by the Hai, making up a bulk of all Hai merchant fleets."
```
tehhowch: Typos:
`design`
`well-defended`

Suggestion:
`comprises the bulk`

data/hai ships.txt
```diff
@@ -273,7 +273,7 @@ ship "Shield Beetle"
 	explode "large explosion" 50
 	explode "huge explosion" 20
 	"final explode" "final explosion large"
-	description "The Shield Beetle is a versatile warship used by the Hai to guard their merchant convoys."
+	description "This ship was first built millenia ago near the end of a more violent time in Hai history, and since then it has been the only heavy warship at the disposal of the Hai. The Shield Beetle is a versatile warship used by the Hai to guard their merchant convoys, or by the Unfettered to attack those very same convoys."
```
tehhowch: Suggestions:
`The Shield Beetle`
`ago, near`
`history. Since then it has served as the`

Not sure about the last sentence.

data/hai ships.txt
```diff
@@ -137,7 +137,7 @@ ship "Lightning Bug"
 	explode "medium explosion" 24
 	explode "large explosion" 8
 	"final explode" "final explosion small"
-	description "The Lightning Bug is a small defensive ship, mostly used by the Hai as a convoy escort. In its stock configuration, its only offensive weapon is a single ion cannon."
+	description "The Lightning Bug is a small defensive ship, mostly used by the Hai as a convoy escort. The Lightning Bug, as with many insects, works best in a swarm; even a small number of Lightning Bugs can manage to leave the largest of ships helpless with their stock-fitted ion cannons."
```
tehhowch: ~~as with many insects~~
; -> :

data/hai ships.txt
```diff
@@ -45,7 +45,7 @@ ship "Aphid"
 	explode "tiny explosion" 10
 	explode "small explosion" 20
 	explode "medium explosion" 15
-	description "The Aphid is a light freighter designed by the Hai. It has only minimal armament."
+	description "The Aphid is the lightest freighter used by the Hai and was for many millennia the backbone of the Hai inter-planetary economy. Given that a captain does not have to fear being attacked in much of Hai space, Aphids are only lightly armed and mainly rely on larger Hai ships for protection."
```
tehhowch: ~~Given that a captain does not have to deal being attacked in much of Hai space~~

Replace with something similar and less 4th-wall. Not quite sure what though, e.g `The low incidence of piracy...`

data/hai ships.txt
Amazinite: >`The Shield Beetle`

If you're talking about replacing `This ship`, I was trying to avoid using `The <ship>` for every single description.

tehhowch: Yeah I had been looking into a rephrase if the last sentence which changed that use of `the shield beetle` into `it`, but couldn't think of a suitable change.

data/hai ships.txt
```diff
@@ -45,7 +45,7 @@ ship "Aphid"
 	explode "tiny explosion" 10
 	explode "small explosion" 20
 	explode "medium explosion" 15
-	description "The Aphid is the lightest freighter used by the Hai and was for many millennia the backbone of the Hai inter-planetary economy. Given that a captain does not have to fear being attacked in much of Hai space, Aphids are only lightly armed and mainly rely on larger Hai ships for protection."
+	description "The Aphid is the lightest freighter used by the Hai and was for many millennia the backbone of the Hai inter-planetary economy. Because of Hai space being peaceful for much of history, Aphids are only lightly armed and mainly rely on larger Hai ships for protection from piracy."
```
Brick63: "its recent history" would sound better

Amazinite: 3/4ths of the 120,000 years of Hai history has been spent in the space that they have now, and from what we can tell, only recently has it not been peaceful (Korath Raids and Unfettered aggression).

Brick63: Okay was suggesting since it sounded weird

KiLEdEnNis: without "its", it sounds like history may imply any history even from when the universe started to exist
it's not a big deal but if ya add "its" it's more clear that it's about Hai space (even though it already kinda is)

Amazinite: **History**: History is the study of the past as it is described in written documents. Events occurring before written record are considered **prehistory**.

Abstracting the written documents part to simply recorded information, saying just `history` when speaking in relation to the Hai would 1) already imply Hai history and 2) not necessarily imply all of happening since the Big Bang.

Bladewood: Try this:
`Because of how peaceful Hai space has been for much of their history, [...]`

Also, I don't agree with the addition about piracy. The previous one more generally refers to any possible threat to the Hai.

hubbyist: "from piracy" seems two narrow.

data/hai ships.txt
hubbyist: "The Shield Beetle is a " -> "Shield Beetles are " may be better.

jafdy: The original sounds like a better sales pitch. Your version sounds more like something from a white paper.

data/hai ships.txt
```diff
@@ -273,7 +273,7 @@ ship "Shield Beetle"
 	explode "large explosion" 50
 	explode "huge explosion" 20
 	"final explode" "final explosion large"
-	description "The Shield Beetle is a versatile warship used by the Hai to guard their merchant convoys."
+	description "This ship was first built millenia ago, near the end of a more violent time in Hai history. Since then it has served as the only heavy warship at the disposal of the Hai. The Shield Beetle is a versatile warship used by the Hai to guard their merchant convoys, or by the Unfettered to attack those very same convoys."
```
MessyMix: Suggestion: Since then(,) it has...

it is an introductory clause.

data/hai ships.txt
```diff
@@ -620,7 +620,7 @@ ship "Water Bug"
 	explode "large explosion" 20
 	explode "huge explosion" 5
 	"final explode" "final explosion medium"
-	description "The Water Bug is the primary freight ship used by the Hai."
+	description "The design of the Water Bug is almost as old as the Aphid, but until recently it never saw much use. A spike in aggression from the Unfettered Hai has lead the tougher and more well-defended Water Bug to see an increase in popularity over its little brother, the Aphid. The Water Bug is now the most widely used freighter by the Hai, comprising the bulk of all Hai merchant fleets."
```
MessyMix: The second sentence is a little unwieldy. What about,

"However, renewed aggression from the Unfettered Hai has seen merchants favor the better-defended Water Bug over its little brother, the Aphid."

Brick63:
I've always felt that the Hai ships descriptions were dull so this is a nice addition.

PR #3222, opened 2017-11-11
Fix misspelling of deceleration

Janaszar:
Quick fix of a spelling error.  Mainly used as an example to figure out how to do Pull Request/Forks and the like for larger stuff.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1050,7 +1050,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 				// Exit hyperspace far enough from the planet to be able to land.
 				// This does not take drag into account, so it is always an over-
 				// estimate of how long it will take to stop.
-				// We start deceleration after rotating about 150 degrees (that
+				// We start decelerating after rotating about 150 degrees (that
 				// is, about acos(.8) from the proper angle). So:
 				// Stopping distance = .5*a*(v/a)^2 + (150/turn)*v.
 				// Exit distance = HYPER_D + .25 * v^2 = stopping distance.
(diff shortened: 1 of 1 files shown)


Janaszar:
Hmm... Was told it should be a PR from a branch on my fork, so going to redo...

Janaszar:
And now people are saying it's ok if it's a single PR change... So, open again?

Bladewood:
Erm... I think that was supposed to say "decelerating", not "deceleration"...

KiLEdEnNis:
both fit though as far as I know

Bladewood:
Yes, but it's usually better to preserve the original intent.

endless-sky:
I feel like "start decelerating" is clearer that "start deceleration," because the choice the AI is making is, "Should I decelerate? Yes, I should start decelerating."

Janaszar:
Technically it should be accelerating/acceleration since reducing your velocity is just applying acceleration counter to your existing velocity. (Something I still remember from my high-school physics with a slightly pedantic teacher)
As for the AI, it could also be thought of as "starting deceleration phase".

KiLEdEnNis:
why not just change it to "decelerating", it's (also) correct & saves pointless discussions

MessyMix:
@Janaszar and the term commonly used to describe that subset of acceleration—that is, acceleration in the opposite direction of your velocity—is deceleration.

Amazinite:
It's a code comment.
Don't think too deeply about it.

Janaszar:
Ok, so I know how to edit a PR now. (If I knew sooner, probably would have done it when I first noticed the difference.  But what an interesting conversation.)

Uses Decelerating instead of Deceleration.

PR #3226, opened 2017-11-12
Make Illegal and Stealth work for passengers, too

Pointedstick:
Now you can write passenger-only missions where being scanned will get you fined and/or fail the mission, like so:

it worked

### DIFF
--- a/source/CargoHold.cpp
+++ b/source/CargoHold.cpp
@@ -581,7 +581,7 @@ int CargoHold::IllegalCargoFine() const
 			return fine;
 		worst = max(worst, fine / 2);
 	}
-
+	
 	for(const auto &it : missionCargo)
 	{
 		int fine = it.first->IllegalCargoFine();
--- a/source/CargoHold.cpp
+++ b/source/CargoHold.cpp
@@ -581,8 +581,16 @@ int CargoHold::IllegalCargoFine() const
 			return fine;
 		worst = max(worst, fine / 2);
 	}
-	
+
 	for(const auto &it : missionCargo)
+	{
+		int fine = it.first->IllegalCargoFine();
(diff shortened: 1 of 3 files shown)


source/CargoHold.cpp
```diff
@@ -581,13 +581,21 @@ int CargoHold::IllegalCargoFine() const
 			return fine;
 		worst = max(worst, fine / 2);
 	}
-	
+
```
tehhowch: `\t`

source/Mission.cpp
```diff
@@ -123,6 +123,20 @@ void Mission::Load(const DataNode &node)
 				passengerLimit = child.Value(2);
 			if(child.Size() >= 4)
 				passengerProb = child.Value(3);
+			for(const DataNode &grand : child)
```
tehhowch: I would bundle this with the above, by defining a private method:
```
    ...
    if(child.Size() >= 4)
        passengerProb = child.Value(3);
    if(child.HasChildren())
        MarkIllegal(child);
}
...
void MarkIllegal(const DataNode &child)
{
    {
        if(grand.Token(0) == "illegal" ...
    ...
```
You'd then call this method for `cargo` as well, removing code duplication for illegal stuff, since the same values are set for cargo and passengers in your implementation.

endless-sky:
I've decided to implement this a simpler way, instead: making "illegal" and "stealth" be attributes of the mission (applying to both cargo and passengers) rather than specifically attributes of the cargo: 03b2421. The main advantage (aside from shorter code and less copy-paste) is that the contraband attributes won't appear twice if a mission has both cargo and passengers.

Pointedstick:
Fantastic, good enough for me!

Issue #2256, opened 2017-03-13
Too easy for unarmed player ships to board enemies in a battlefield


Bladewood:
I have a simple money-generating strategy.
I sit in Wah Ki with a boarding-optimized Lightning Bug, waiting for Hai and Unfettered Hai ships to jump in and start fighting each other.
I then board any disabled Unfettered ship that comes out of the battle, steal its reactors and other outfits, and beat a hasty escape, selling my goods at a world.

This strategy has become much too easy after the commit that made enemies prioritize armed ships over unarmed ships. (I believe it was https://github.com/endless-sky/endless-sky/commit/9b1014cb2386ff881ed98487acd9b9af696ecf7b.) I can just wait around in an unarmed Lightning Bug, praying that some stray ion bolt or tracker doesn't hit me, while happily watching as the two Hai factions murdered each other.

karkatic:
<praying that some stray ion bolt or tracker doesn't hit me>

at this point, you're save scumming, so... :-D

KiLEdEnNis:
no, if he gets hit he may have to jump away
don't assume one save scums, it's not like 1 ion bolt or tracker kills him

wambat1:
Similar issue in any conflict system. Having AI ships priority target ships that have targeted their faction's ships for hostile boarding should address this problem. It should be possible vulture-ise battlefields but not as easy as it is now.

tehhowch:
This should be accomplished pretty simply by referencing AI::Has() within FindTarget and setting an appropriate range reduction. Twice the deprioritization that a ship receives for being unarmed could work fairly well, without making it so that the second you board a ship, all of its allies beeline towards you.

Pointedstick:
This would also be largely solved by implementing some of the suggestions in https://github.com/endless-sky/endless-sky/issues/3346.

tehhowch:
With the merge of #3230, unarmed ships will be considered during targeting if they have boarded hostile vessels. That should pretty effectively require 1) the unarmed ship to have a large protective force to deal with hostile ships (i.e. the player is already very wealthy and powerful), or 2) the unarmed ship to jump away after boarding / capping to remove the targeting priority (increasing upkeep costs due to more elapsing days).
It could be further addressed with #3346, or as I mentioned in #3520 , having the opposing ship attack yours if they vastly outpower you when you board.

Amazinite:
Could this be considered done due to the merging of #3230?

Pointedstick:
I think so.

PR #3230, opened 2017-11-17
Boarding a faction increases the chance of being targeted by them

tehhowch:
Refs #2256 , 

There were two ways to perform the check in the new `Has()` method:
1) loop every ship in the `actions` map, and check if their government is the same as the ship performing the targeting check, and then if they have been the target of the event in question
2) Add a map which stores this information directly

I went with route 2 as it is logically simpler to understand, and requires no loops in the new method.

The specific value of the prioritization could be tweaked. I wanted it to offset the deprioritization caused by removing all weapons from one's ship, but it's possible this value will result in armed ships that plunder being excessively targeted by other members of the boarded factions. However, IMO this is a warranted side-effect of boarding and (and likely destroying) otherwise harmless entities.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -244,7 +244,10 @@ void AI::UpdateEvents(const list<ShipEvent> &events)
 	for(const ShipEvent &event : events)
 	{
 		if(event.Actor() && event.Target())
+		{
 			actions[event.Actor()][event.Target()] |= event.Type();
+			notoriety[event.Actor()][event.TargetGovernment()] |= event.Type();
+		}
 		if(event.ActorGovernment() && event.Target())
 			governmentActions[event.ActorGovernment()][event.Target()] |= event.Type();
 		if(event.ActorGovernment()->IsPlayer() && event.Target())
@@ -265,6 +268,7 @@ void AI::UpdateEvents(const list<ShipEvent> &events)
 void AI::Clean()
 {
 	actions.clear();
+	notoriety.clear();
 	governmentActions.clear();
 	playerActions.clear();
 	shipStrength.clear();
@@ -943,6 +947,8 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 			}
 			// Prefer to go after armed targets, especially if you're not a pirate.
 			range += 1000. * (!IsArmed(*it) * (1 + !person.Plunders()));
+			// Targets which have plundered this ship's faction earn extra scorn.
+			range -= 1000 * Has(*it, gov, ShipEvent::BOARD);
 			// Focus on nearly dead ships.
 			range += 500. * (it->Shields() + it->Hull());
 			if((isPotentialNemesis && !hasNemesis) || range < closest)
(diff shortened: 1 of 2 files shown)


tehhowch:
I don't believe this change will affect any mission battles, with the sole exception of when the player is attempting to obtain the Surveillance Pod and the Electron Gun/Turret. In those two missions, however, the player is generally the only hostile target in-system anyway, so all this would do is make it more likely that the AI ships will target the player's flagship rather than members of the player's fleet.

endless-sky:
This seems like a lot of complication to add to the AI just to avoid one minor exploit.

tehhowch:
I don't classify it as minor though - it's the de facto "hey guys, do this" recommendation given on Steam whenever someone posts about money.
Right now there is no penalty for cannibalizing ships beyond the rep hit. With a targeting priority, the boarding ship is now vulnerable no matter how many escorts or distracting AIs are around. If you're still far enough away, hostiles will keep firing at their current targets. But if you try to do this right under their noses, they'll put you down for it.

tehhowch:
It also opens the door to consideration of other events when targeting ships for actions.
 - `vindictive`/`heroic` ships prefer targeting those who have boarded or destroyed allies
 - `coward`/`timid` ships prefer fighting ships which have not destroyed any allies

etc

Amazinite:
>This seems like a lot of complication to add to the AI just to avoid one minor exploit.

>I don't classify it as minor though - it's the de facto "hey guys, do this" recommendation given on Steam whenever someone posts about money.

The way I see it, combat ships ignore ships without weapons because they pose no threat. But a ship that is going around and capturing their own ships, which then attack them, is a threat, and therefore it should be dealt with. 

If there was a pirate ship boarding and capturing my escorts to turn them against me, that would be a major target in my eyes, especially if it's capturing powerful ships. (Maybe have the increased attention toward your ship scale by the DPS of the ships you're capping?)

And as mentioned, I wouldn't consider the exploit minor; capturing ships is still a highly lucrative way to make money if you know what you're doing. I may be wrong in thinking this, but the best balance for capturing ships has always been the risk of dying in the process, but there is little to no risk in capturing ships if you don't even get attacked for doing it just because your flagship is unarmed.

source/AI.cpp
```diff
@@ -943,6 +947,9 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 			}
 			// Prefer to go after armed targets, especially if you're not a pirate.
 			range += 1000. * (!IsArmed(*it) * (1 + !person.Plunders()));
+			// Targets which are plundering allies earn extra scorn,
+			// especially to vengeful personalities.
+			range -= 1000 * Has(*it, gov, ShipEvent::BOARD) * (1 + person.IsVindictive());
```
endless-sky: Because the "vindictive" personality confers a significant disadvantage in combat (ships get distracted attacking enemies that are already dead), no one will ever want to use "vindictive" specifically to get this effect. So, I think it'd be better to leave "vindictive" with its current meaning and not to expand it to this.

endless-sky:
Given that this is only needed to prevent a player exploit, there's no reason for this much complexity. I would recommend, instead of a `map<weak_ptr<const Ship>, map<const Government *, int>, Comp>`, just having a `map<const Government *, double>` that tracks things the player has done to anger the given government on the current day. (For now, it would just get increased, up to some limit, when the player boards a ship of a given government. But eventually, there might be other things the player could do to become a special target.) That would make the code a whole lot simpler.

tehhowch:
With a player-exclusive `notoriety` counter as you suggest, it feels like we need a hackish check to determine if the ship being considered as a target is the player's flagship (without distinguishing between player flagship and player escort, the player's armed escorts will attract the attention and leave the player's flagship still free to plunder and capture freely).

I liked the idea of using the existing ShipEvent handling code and a new tracker object that strongly reflects the existing ShipEvent tracking objects, along with a natural, player- & AI-agnostic method of controlling ship target selection via the `Has` check. I'm happy to propose additional NPC personalities, or tweaks to the existing ones, to make use of the `notoriety` tracking that this PR adds, but I wanted to separate the exploit fix from general enhancements.

endless-sky:
I don't think that would be necessary, if the only goal is to avoid the exploit where the player is not getting attacked at all because their flagship is unarmed. If the player has armed ships that are close enough to the fight that they are getting targeted and potentially killed, then they're really not exploiting anything.

PR #3216, opened 2017-11-09
Terraforming Tundra

Amazinite:
The Terraforming Rand missions (Terraforming 1 to 5) left me with a couple of questions:
1. If the terraforming technique turned out so successfully, why was no follow-up done?
2. If you just crashed an asteroid into a planet on purpose, why does the Republic not care at all?

These mission seek to remedy that, as well as giving Amy's techniques another shot to prove themselves.

These new missions (Terraforming 6 to 12 + a follow-up mission) start off with the player having to answer questions about why they crashed an asteroid into a planet. When the case is dropped, Amy asks you to transport her and a new character, Nolan, to the Deep where they spend a few months researching Amy's ideas. When the researching phase is over and Amy gets grant money from the Deep as well as the permission of the Republic, she contacts you about transporting her to Tundra, where they attempt another round of terraforming.

I did some actual research to try and figure out how best to go about this. While the method is probably entirely impossible in real life, I think there's at least some level of believeability to it. (Volcanoes explode when the pressure of the magma exceed the pressure of the crust, volcano eruptions release greenhouse gases that can heat the atmosphere, and Earth got its initial atmopshere from the eruption of volcanoes that warmed the planet enough for liquid water to exist.) I also wanted to avoid using *just* an asteroid this time, so I included the aspect of a supervolcano, as well as spending some time having characters argue whether an asteroid impact is really the best way to terraform planets.

The missions end with a 1 to 2 year waiting period until the follow-up message, in which Amy informs you how Tundra and Rand have been progressing in the past months, and how the Republic has asked for a halt on all terraforming experiments from Amy while they monitor Tundra and Rand, putting a more formal close on the story than `Terraforming 5` did.
<hr/>

In addition to the new missions, I've moved the `terraforming Rand` event from `events.txt` to `tranport missions.txt`, as all other events used in the file were also located in the file. 
I've also implemented a fix that would resolve #2691, including a `blocked` dialog in `Terraforming 2` when the number of passengers gets bumped up from 2 to 3. This can be reverted if desired.

PR #3234, opened 2017-11-19
"Terraforming 2" should be 3 passengers?

luiges90:
Reading the text in the mission it seems there are 3 passengers (Eric, Alaric and Amy) but the quests actually require just 2 passengers.


jafdy:
If you are going to do that then you need to put a `blocked` in so players don't wander why they can't continue.

Amazinite:
This is handled in #3216

luiges90:
@Amazinite lets stick with your PR then 😄

Issue #3243, opened 2017-11-21
Mission ships don't all jump as group


SirStigII:
I am on the liberate Poisonwood mission, and the falcons and cruiser won't all jump at the same time, even when I hold down the J key. the cruiser and one of the falcons will prep for the jump, but one of the falcons always is drifting , causing both falcons to wait to jump while the cruiser jumps in immediately. This makes it super hard (for the cruiser) to successfully survive the waiting pirates when using a weak player ship. I have noticed this drifting behaviour at other times too, but usually it is a small inconvenience, but on missions like this it's a real pain.

tehhowch:
Are you on 0.9.8 (i.e. Steam), or a self-compile? Behavior is different between them.

SirStigII:
self compile. Just retested and it happens with jump and hyperdrive.

SirStigII:
it seems that the ships are using a different escort personality, as the cruiser looks for a fight, and the falcons like to stay a little closer to me in combat. Although that doesn't explain why one will prep for jump but not the other one...

tehhowch:
Post your save so we can easily reproduce

SirStigII:
[Flee Destiny~3018-07-09.txt](https://github.com/endless-sky/endless-sky/files/1489945/Flee.Destiny.3018-07-09.txt)
[Flee Destiny~3018-07-02.txt](https://github.com/endless-sky/endless-sky/files/1489946/Flee.Destiny.3018-07-02.txt)
With both drives.

SirStigII:
happens on the escort behemoth mission too, one behemoth with prep, other will go off and fight or drift near me.

tehhowch:
Yeah, this a a change you can thank me for ;)
The NPC flagships obey you (the cruiser and falcon that wait), while the NPCs' escorts obey their flagship (the falcon that is obeying you).
The fix is to propagate the JUMP && WAIT command pair through - though I thought I had done this already.

tehhowch:
It happens because the "middle" ship in the order - the NPC flagship - doesn't give itself the jump command until its parent (you) actually begin jumping. This means when its escort is deciding how to move, if everyone is in your system, its parent doesn't have the jump command set, and it instead chooses KeepStation behavior.

It works fine when you aren't in the system, because the NPC flagship executes a different command block, one in which it gives itself both the jump command (instructing escorts to prepare) and the wait command (until its escorts are ready).

tehhowch:
Should be addressed in #3246.

[...]

PR #3246, opened 2017-11-21
bugfix: Allied mission NPCs should prepare to jump with the player if player is holding J

tehhowch:
Refs #3243 , #3065 , #3052 

"Grandparent" knowledge is needed in AI::Step else the NPC's escorts will continually turn towards a hostile target and then back to the jump direction.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -72,17 +72,30 @@ namespace {
 		return min(a, 360. - a);
 	}
 	
-	// Determine if all able, non-carried escorts are ready to jump with this
-	// ship. Carried escorts are waited for in AI::Step.
-	bool EscortsReadyToJump(const Ship &ship)
+	// Determine if all able, non-carried escorts are ready to jump with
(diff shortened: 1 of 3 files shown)


SirStigII:
How would I add this to my build?

tehhowch:
From the git command line, you'd do something like `git cherry-pick 1b0a19c` 
If you don't want to use the command line, then you either wait for it or a different solution to merge, or manually PR this branch to your GitHub fork (careful here as it defaults to PRing to this repo) and merge it, then do whatever you normally do to get source.

endless-sky:
I think 69edbbe addresses all these issues, and in a simpler way (by propagating the "WAIT" command to the escorts, instead of by needing to check parents of parents). I've tested it with ordinary non-mission fleets, friendly NPC escorts, and hostile NPC escorts, and they all behave as expected.

tehhowch:
That's definitely simpler :+1:

Issue #3249, opened 2017-11-21
Request: add “medium freighter” category


Nescio0:
(Hopefully this is the correct place to post, I'm new to endless-sky.)

Currently warships are grouped under “light”, “medium”, or “heavy”, but freighters can only be either “light” or “heavy”, because “medium” does not exist. This leads to weird things, e.g. the Nest, Roost, and Skein are all medium warships, but for their freighter counterparts, the hauler is light and the hauler ii and hauler iii are heavy. Therefore I think a “medium frighter” category should be added.

After looking at the available ships in human space it seems the criterion is the empty mass (i.e. hull plus outfits):
light: <500
medium: 500-1000
heavy: >1000

So what I suggest is:
* `source/Ship.cpp`: 
  * insert at line 49: `"Medium Freighter",`
* `data/hai ships.txt`: 
  * change Water Bug category to "Medium Freighter"
* `data/ships.txt`:
  * change Argosy category to "Medium Freighter"
  * change Bastion mass to 540 and outfit space to 450
  * change Hauler category to "Medium Freighter"
  * change Hauler II category to "Medium Freighter"
  * change Hauler III category to "Medium Freighter"
  * change Modified Argosy category to "Medium Warship"

(If I had commit access I could push these changes myself.)

jafdy:
You can create a pull request.

Why are you changing the Bastion? 
The categories used are fairly arbitrary, especially for non-warships. Although this is an easy change, it feels unnecessary.
(Someone will request the freighter to be made a med is this is accepted.)

tehhowch:
Adding a category just to add a category seems a bit strange. What does this offer?
PS only MZ has commit access - everyone else who wants to submit changes goes through the Pull Request review process, no matter how seasoned of a contributor they are.

Nescio0:
Thank you for your quick replies!
How does the “Pull Request review process” work?

jafdy:
You make a fork of the repository, make a branch on that fork, make changes to that branch and click the pull request button. It then opens a pull request that looks a lot like an issue where we ~~shred~~ examine your ideas and MZ might accept them.

gunqqer:
With changes like this, the community discusses and when MK feels like giving his input, he does, and if he want to merge it, he does. There isn't really a formal process. With bug fixes sometimes he'll look over it and just merge it.

tehhowch:
[...]

PR #3251, opened 2017-11-21
Add “medium freighter” category

Nescio0:
Warships are subdivided into light, medium, and heavy; the criterion seems to be the empty mass (i.e. hull plus outfit capacity).
```
                    mass
light  warship:     <500
medium warship: 500-1000
heavy  warship:    >1000
```

Freighter categories are currently assigned rather arbitrarily. This pull request changes that to harmonize freighter categories to those of warships.
The Hauler, Hauler II, and Hauler III all fit within the medium category (cf. their warship counterparts Nest, Roost, and Skein), as do a few others.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -66,7 +66,7 @@ ship "Argosy"
 	plural "Argosies"
 	sprite "ship/argosy"
 	attributes
-		category "Light Freighter"
+		category "Medium Freighter"
 		"cost" 1960000
 		"shields" 4200
 		"hull" 2600
@@ -317,12 +317,12 @@ ship "Bastion"
 		"hull" 4200
 		"required crew" 32
 		"bunks" 47
-		"mass" 580
+		"mass" 540
 		"drag" 10.3
 		"heat dissipation" .5
 		"fuel capacity" 500
 		"cargo space" 110
-		"outfit space" 470
+		"outfit space" 450
 		"weapon capacity" 180
 		"engine capacity" 120
 		weapon
@@ -1483,7 +1483,7 @@ ship "Gunboat"
 ship "Hauler"
 	sprite "ship/hauler i"
 	attributes
-		category "Light Freighter"
+		category "Medium Freighter"
 		"cost" 1430000
 		"shields" 2500
 		"hull" 3700
@@ -1537,7 +1537,7 @@ ship "Hauler"
 ship "Hauler II"
 	sprite "ship/hauler ii"
 	attributes
-		category "Heavy Freighter"
+		category "Medium Freighter"
 		"cost" 2340000
 		"shields" 2900
 		"hull" 5200
@@ -1591,7 +1591,7 @@ ship "Hauler II"
 ship "Hauler III"
 	sprite "ship/hauler iii"
 	attributes
-		category "Heavy Freighter"
+		category "Medium Freighter"
 		"cost" 3260000
 		"shields" 3300
 		"hull" 6700
@@ -1923,7 +1923,7 @@ ship "Modified Argosy"
 	plural "Modified Argosies"
 	sprite "ship/modified argosy"
 	attributes
-		category "Light Warship"
+		category "Medium Warship"
 		"cost" 1960000
 		"shields" 4800
 		"hull" 1900
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -317,12 +317,12 @@ ship "Bastion"
 		"hull" 4200
 		"required crew" 32
 		"bunks" 47
-		"mass" 540
+		"mass" 580
 		"drag" 10.3
 		"heat dissipation" .5
 		"fuel capacity" 500
 		"cargo space" 110
-		"outfit space" 450
+		"outfit space" 470
 		"weapon capacity" 180
 		"engine capacity" 120
 		weapon
@@ -1923,7 +1923,7 @@ ship "Modified Argosy"
 	plural "Modified Argosies"
 	sprite "ship/modified argosy"
 	attributes
-		category "Medium Warship"
+		category "Light Warship"
 		"cost" 1960000
 		"shields" 4800
 		"hull" 1900
--- a/data/coalition ships.txt
+++ b/data/coalition ships.txt
@@ -55,7 +55,7 @@ ship "Arach Courier"
 ship "Arach Freighter"
 	sprite "ship/arach freighter"
 	attributes
-		category "Light Freighter"
+		category "Medium Freighter"
 		licenses
 			Coalition
 		"cost" 4580000
@@ -157,7 +157,7 @@ ship "Arach Hulk"
 ship "Arach Spindle"
 	sprite "ship/arach spindle"
 	attributes
-		category "Heavy Freighter"
+		category "Medium Freighter"
 		licenses
 			Coalition
 		"cost" 6652000
@@ -208,7 +208,7 @@ ship "Arach Spindle"
 ship "Arach Transport"
 	sprite "ship/arach transport"
 	attributes
-		category "Light Freighter"
+		category "Medium Freighter"
 		licenses
 			Coalition
 		"cost" 2772000
@@ -862,7 +862,7 @@ ship "Saryd Sojourner"
 ship "Saryd Traveler"
 	sprite "ship/saryd traveler"
 	attributes
-		category "Heavy Freighter"
+		category "Medium Freighter"
 		licenses
 			Coalition
 		"cost" 3385000
--- a/data/hai ships.txt
+++ b/data/hai ships.txt
@@ -575,7 +575,7 @@ ship "Violin Spider" "Violin Spider (Pulse)"
 ship "Water Bug"
 	sprite "ship/hai water bug"
 	attributes
-		category "Heavy Freighter"
+		category "Medium Freighter"
 		"cost" 6500000
 		"shields" 7900
 		"hull" 4500
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -45,7 +45,8 @@ namespace {
 
 const vector<string> Ship::CATEGORIES = {
 	"Transport",
-	"Light Freighter",
+	"Light Freighter",	
+	"Medium Freighter",
 	"Heavy Freighter",
 	"Interceptor",
 	"Light Warship",


data/ships.txt
```diff
@@ -317,12 +317,12 @@ ship "Bastion"
 		"hull" 4200
 		"required crew" 32
 		"bunks" 47
-		"mass" 580
+		"mass" 540
```
jafdy: Why are you changing an existing ship?

data/ships.txt
```diff
@@ -1923,7 +1923,7 @@ ship "Modified Argosy"
 	plural "Modified Argosies"
 	sprite "ship/modified argosy"
 	attributes
-		category "Light Warship"
+		category "Medium Warship"
```
jafdy: Why are you changing a ship to an existing category, if it was wanted in this category then it would have been put in this category in the first place.

jafdy:
#3249

data/ships.txt
tehhowch: I would agree that the Mod. Argosy is hardly a Medium Warship.

Bladewood: Notably, modified argosy is on par with a gunboat or a raven in terms of health, but it carries the armament of a firebird or a corvette.

Nescio0: Because I believe the modified argosy was only assigned the "light warship" category because the normal argosy was in the "light freighter" category because there used to be no "medium freighter" category.

Comparing the modified argosy with other human space warships, it's actually heavier than several medium warships, therefore I believe this change is justified.

data/ships.txt
Nescio0: Because the Bastion used to be heavier than e.g. the Vanguard warship, and I think it ought to remain a medium warship, therefore I believe slightly removing some unnecessary mass is justified here. (With this change the Bastion's free outfit space becomes 14/450, comparable to other medium warships.)

data/ships.txt
Bladewood: Mass *is not* the only thing that determines a ship's category. Otherwise, the Summer Leaf would be an interceptor, or Met Par Tek 53 and Model 32 would be medium warships, or some other ridiculous nonsense would occur. A Marauder Bounder or a Marauder Fury or even a stock Hawk would become a light warship.

data/ships.txt
Bladewood: Uh, mass is independent of outfit space...

Nescio0:
A comparison of human space medium warships:
```
                     shields, hull,    mass + outfit, weapons, engines, guns + turrets, bays
  Aerie            :    5700, 1900,  570 = 130 + 390,     150,      95,      5 = 2 + 3,    2
  Bastion          :    6700, 4200, 1050 = 580 + 470,     180,     120,      7 = 4 + 3,    0
  Corvette         :    6100, 1200,  570 = 150 + 420,     150,     100,      6 = 4 + 2,    0
  Firebird         :    6400, 2800,  690 = 290 + 400,     160,     100,      6 = 4 + 2,    0
  Manta            :    5900, 1500,  520 = 270 + 350,     140,     100,      6 = 6 + 0,    0
 *Modified Argosy* :    4800, 1900,  590 = 250 + 340,     140,      80,      6 = 4 + 2,    0 *
  Mule             :    5400, 4400,  770 = 320 + 450,     210,     110,      7 = 2 + 4,    1
  Nest             :    2500, 3700,  650 = 250 + 400,     140,      80,      6 = 2 + 4,    2
  Osprey           :    7200, 1600,  720 = 270 + 450,     180,     130,      6 = 4 + 2,    0
  Roost            :    2900, 5200,  810 = 360 + 450,     140,      80,      6 = 2 + 4,    4
  Skein            :    6700, 3300,  970 = 470 + 500,     140,      80,      6 = 2 + 4,    6
  Splinter         :    5200, 1700,  650 = 250 + 400,     150,     100,      5 = 2 + 3,    0
```

data/ships.txt
Nescio0: If not mass, then what? If outfits, weapons, engines, number of guns and turrets, the Modified Argosy clearly fits within the medium warship category, see below.

jafdy: In case you hadn't noticed, there is not set metric on how to define a ship.

Nescio0: No rules set in stone, perhaps, but there appears to be a pattern, at least for human space ships. I suggest we follow the warship category pattern and apply this to the freighter categories.

tehhowch:
There are some spreadsheets available that rank the various ships/outfits, made by @Amazinite : https://drive.google.com/drive/folders/0B635z_nU19WfQllrM2V2dWpFSFk

I will say that based on your table it does seem sensible to promote the Modified Argosy from Light to Medium Warship. I think it's a bit tough to argue for the creation of a new ship category that has only 5 ships total though, and I've no opinion on the changes to the Bastion.

Nescio0:
> I  will say that based on your table it does seem sensible to promote the  Modified Argosy from Light to Medium Warship. I think it's a bit tough  to argue for the creation of a new ship category that has only 5 ships  total though, and I've no opinion on the changes to the Bastion.
The "Heavy Freighter" category is even smaller; however, I don't think the number of ships should really matter. Why I think this category is necessary, is because "Medium Freighters" are actually statistically quite similar to "Medium Warships", and quite different from both light and heavy freighters.

Nescio0:
Number of freighters in each category with this diff:
```
            Light, Medium,  Heavy,  Total
Human    :      3,      4,      2,      9       
Hai      :      1,      1,      0,      2
Coalition:      3,      4,      2,      9
Wanderer :      1,      0,      1,      2
Korath   :      0,      0,      1,      1
Total    :      8,      9,      6,     23
```

Amazinite:
@ discussion on changing the Bastion and Mod Argosy.
This PR is for the purpose of creating a medium freighter category. The title and description make this clear. What this is not is a PR designed to standardize ship categories based off of certain criteria. Given this, I suggest you remove the Bastion and Mod Argosy changes as they are outside of the scope of this PR as is, and they make the PR less likely to be merged because of this.

As for my opinion on the actual changes:
The change to the Bastion seems very unnecessary to me. Yes, the Bastion is indeed heavier than the Vanguard. It is also heavier than the Behemoth, Falcon, Protector, and Leviathan. The change from 580 mass to 540 mass doesn't actually change any of that, except for making its mass equal that of the Behemoth. If looking at mass + outfit space, then the Vanguard is actually 10 tons heavier than the Bastion, and the ships below the Bastion are the Behemoth and Falcon (which would have the same mass as the Bastion of 1,050) and the Skein (another medium warship who would be at 970 tons). So whether the change in mass is needed or not, your argumentation for it just doesn't work out.
<hr/>

Ignoring the fact that I disagree with the mass change, the following is a graph showing the relationship between mass and outfit space for all ships in the game.
screenshot_271

While more mass does generally mean more outfit space, this is only because more mass is also tied to being a generally bigger ship, which is what is causing the higher outfit space, i.e. it isn't the mass. There's also a wide range of outfit space at any selection of mass. E.g. almost directly between 500 and 1,000 we can see 7 ships with very similar masses but a wide range of outfit spaces that go from what looks like 300 to almost 1000.

The design philosophy of MZ (@endless-sky) for ships also has to do with the size of a ship. He's stated in the past that he will create the ship model, size it, and then based off of the size he will create the stats. In addition to the size, the context of the ship matters as well. (In this case, the context is that the Bastion is a Southbound ship.) In short, the Bastion got lighter, not smaller, so the outfit space shouldn't change. That's not even mentioning the fact that you didn't even check the outfit space of the Bastion's two variants, the Bastion (Laser) and Bastion (Heavy), as both max out their outfit space in vanilla, meaning they'd have -20 outfit space from this PR, and any change to their outfits might warrant balance testing of Free World campaign missions to make sure the Bastion is still pulling its weight in Free Worlds fleets for battles against enemies.
<hr/>

**tl;dr, I disagree with the Bastion changes and believe that they should be reverted in this PR.**
<hr/>

Onto the Modified Argosy. Going back to the previously mentioned importance of context in balancing ships, the Modified Argosy is intended to be heavier set for its category. It's just an Argosy with a bunch of armor plates attached to it and some redesigned innards done by pirates.

Onto the question of stats. Here is a table of the top three light warships and bottom three medium warships in terms of hit points (hull + shields)

Ship | Category | Hit Points
------|---------|----------
Corvette | Medium War | 7,300
Gunboat | Light War | 7,200
Splinter | Medium War | 6,900
Argosy | Light War | 6,700
Nest | Medium War | 6,200
Raven | Light War | 6,100

When considering context, this makes sense to me:
* The Gunboat is top-notch military technology and therefore is expected to have higher hit points for its tier.
* The Nest is a Hauler I converted into a carrier, and therefore will have lower hit points for its category
* The Splinter is built by the Syndicate, who's ships are generally fragile (see Vanguard and Protector being the weakest heavy warships).
* The Corvette is a lightly-built ship designed for speed, and therefore it sacrifices strength.

The Argosy could have a place among the medium warships category, but then we'd have to argue over the placement of the Nest, Gunboat, and possibly the Splinter.
<hr/>
Looking at outfit, weapon, and engine space, here is how the Argosy stacks up to a few medium warships by the metric of outfit space:

Ship | Outfit | Weapon | Engine
-----|-------|-------|------|
Firebird | 400 | 160 | 100
Splinter | 400 | 150 | 100
Nest | 400 | 140 | 80
Aerie | 390 | 150 | 95
Manta | 350 | 140 | 100
Argosy | 340 | 140 | 80

In terms of these metrics, it's clear to see that should the Modified Argosy be a medium warship, it would be the weakest among them. The weapon space is similar to the Nest and Manta, but the Manta has higher outfit and engine space while the Nest has much higher outfit space. None of these medium warships fall lower than the Modified Argosy, and those that tie in certain categories exceed the Argosy in others.
<hr/>

So context in mind, the Modified Argosy can just barely be argued to be a medium warship in terms of hit points and should it be a medium warship would be the weakest in terms of outfit, weapon, and engine space.

Quickly making mention of turret and gun space, the Modified Argosy really only beats out (assuming guns > turrets) the Manta medium warship, which has 6 guns. The Splinter is comparable at 2 guns and 3 turrets, and the Firebird, Osprey, and Corvette match the Modified Argosy. Any other medium warships are better gunned.
<hr/>

**tl;dr The Modified Argosy is currently a powerful light warship, but changing its category only makes it the weakest medium warship, so a change in category is, in my eyes, unnecessary, but not incorrect.**

endless-sky:
There are currently 90 warships in the game, and only 29 freighters. That's why there needed to be 3 different size classes for warships. Given that there's only a third as many freighters, there's no good reason to split them up any further.

KiLEdEnNis:
eventually I think we should add a Medium freighter category, and maybe even split the Transports up in Light, Medium & Heavy
but for now and probably the next couple of years it shouldn't be needed (although I personally prefer Light/Medium/Heavy tabs for all 3)

Nescio0:
* Reverted Bastion tweaks, as it was clear that would not be accepted.
* Spun off  Modified Argosy class change into a new diff.


Let's focus the discussion on the actual proposal (adding a medium freighter category).
Arguments against:
* It does not add much
* There are fewer freighters than warships

Arguments in favour:
* It does not break anything
* Organizing freighters parallel to warships is less arbitrary than the current situation
* More freighters might be added
* Other categories are even smaller

If the number of freighters really is the only objection, then you might as well argue that light and heavy freighters ought to be merged into a single freighter category (cf. transports).

jafdy:
"does not brake anything" is not an argument in favor. 
When more freighters are added then an extra category can be added, once the category is added it can never be removed.
"other categories are even smaller" is a useless argument because those categories can't be merged in any coherent way. 

You appear to have confused "not against" with "in favor".

Nescio0:
> You appear to have confused "not against" with "in favor".

True, however, the arguments "against" are actually also "not in favour".

> once the category is added it can never be removed.

Nonsense. Simply remove a line of code from the `Ship.cpp` file and use `grep` to find all occurences or `sed` to automatically replace them. It could be done in a minute or so, if desired.

tehhowch:
Yes deleting on code is fine. However there is the issue that removing category support will result in a breaking change that can invalidate save files. Endless Sky has 100% backwards compatibility by design to prevent this, so adding support for an attribute (or equivalent item that is saved to or read from the save) means it must be supported forever.

KiLEdEnNis:
not directly relevant
"once the category is added it can never be removed"
not really good to have "Transport" then, when in the future it may be split in Light/Medium/Heavy
w/ Freighters ya just add a category, here ya'd add 3 & remove 1 if that can even be done safely

endless-sky:
I see no good reason to do this.

PR #3258, opened 2017-11-22
Removed common shuttle from luxury shipyard

Nescio0:
The Luxury Shipyard (at Shroud) already sells three other shuttles, so there is no real need for selling the cheap shuttle as well. Furthermore, the common shuttle is already sold at many other systems, including neighbouring Betelgeuse.

### DIFF
--- a/data/sales.txt
+++ b/data/sales.txt
@@ -103,7 +103,6 @@ shipyard "Northern Pirates"
 	"Berserker"
 
 shipyard "Luxury Ships"
-	"Shuttle"
 	"Blackbird"
 	"Falcon"
 	"Arrow"


Amazinite:
>The Luxury Shipyard (at Shroud) already sells three other ~~shuttles~~ [small transport ships], so there is no real need for selling the cheap shuttle as well.

Why not?

KiLEdEnNis:
the only argument I can think of is that a basic shuttle ain't a luxury ship, that's it

endless-sky:
The idea behind what "luxury ships" are for sale there is that rich people probably only need to transport people and small amounts of luggage; they're not hauling tons of commodities. But, there are plenty of situations there where a shuttlecraft could be a luxury item - e.g. to ferry guests to the top of your private mountain or to have a spontaneous picnic lunch on the moon.

KiLEdEnNis:
so basically the future version of a private yet

PR #3259, opened 2017-11-22
Tweaked gatling gun ammo price.

Nescio0:
Changed "Gatling Gun Ammo" value to 4 each, because 25% of 4 is 1 (and 25% of 2 is 0).

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -1021,7 +1021,7 @@ effect "nuke residue slow"
 outfit "Gatling Gun Ammo"
 	plural "Gatling Gun Ammo"
 	category "Ammunition"
-	cost 2
+	cost 4
 	thumbnail "outfit/bullet"
 	"mass" .002
 	"gatling round capacity" -1


jafdy:
This also doubles the cost of gatling ammo. Why?

Bladewood:
It makes me happy that a fully loaded gatling gun will at least be more expensive than a beam laser.

tehhowch:
An alternative solution is to make the minimum sale price 1 credit in the depreciation mechanic. If it's already 1, then I don't see a need for a change.

Nescio0:
True, but who would buy it anyway? And it still remains cheap anyway.

tehhowch:
Doesn't seem like you've used ammo based weaponry much.

They're quite strong . The balance is that they require restocking.

Nescio0:
True. So far I've only used Bactrians, which I equipped with Hai weapons.

Bladewood:
A fully loaded Meteor Missile Launcher costs 32.5k credits, with an overall damage output of 4.55k shield and 2.8k hull damage. A fully loaded Gatling Gun, right now, costs only 26k credits, but has an output of 9k shield and 18k hull damage. Combine that with the fact that it takes up 2 less outfit space, and especially with its devastating burst-fire nature, and it starts to become clear that the Gatling Gun is underpriced.
Even just doubling the price of a round at least pushes it up to 32k credits.

tehhowch:
The Gatling gun does come with that annoying side effect of being within weapons range of your opponent, and needing to not run away. A better comparison may be to the javelin, I think, as the usage scenario is much closer.

Nescio0:
For comparison:
```
                              range , shots/ ,  cost  , cost/
                                    , second , (each) , second
gatling gun (current)       ,   336 ,  20.0  ,      2 ,   40
gatling gun (proposed)      ,   336 ,  20.0  ,      4 ,   80
javelin pod                 ,   600 ,   3.0  ,     50 ,  150
heavy rocket launcher       ,   800 ,   0.3  ,   2000 ,  600
torpedo launcher            ,  4200 ,   0.5  ,   1400 ,  700
meteor missile launcher     ,  5400 ,   1.25 ,    500 ,  625
sidewinder missile launcher ,  6600 ,   1.5  ,    800 , 1200
hai tracker                 ,  8400 ,   1.0  ,   1000 , 1000
```

> An alternative solution is to make the minimum sale price 1 credit in the depreciation mechanic.

That would also be great.

Bladewood:
A loaded Javelin Pod inflicts 9.6k shield and 5.2k hull damage, and costs 30k.

Amazinite:
>A loaded Javelin Pod inflicts 9.6k shield and 5.2k hull damage, and costs 30k. So 4k less credits for 3x the hull damage and only 600 less shield damage.

The Javelin Pod can also be refilled everywhere but the Paradise Worlds, Deep, and Far North.
The Gatling Gun on the other hand can only be refilled on pirate worlds and Trinket in the Sargas system. Unless one stays near Trinket, they need to bribe their way on to pirate worlds, which is in itself an extra cost to purchasing ammo. They also need to be able to survive the pirates in the system, and given that the Gatling Gun is really only useful on small ships, the player's small ship doesn't have the best chances of landing, unless they also bribe the pirates which costs even more credits.

PR #3262, opened 2017-11-23
Even more ship names!

Pointedstick:
Lots of changes here:
- More words in all Civilian and Pirate categories
- Hugely simplified Civilian naming system, to make it less intimidating to others
- Extensively tested and tweaked to get rid of a lot of awkward combinations (like "Lucky Old")


data/names.txt
```diff
@@ -807,40 +922,129 @@ phrase "standalone civilian ship names"
...
+	phrase
+		"male names"
 		"male names"
+		"male titles"
+		"female names"
 		"female names"
+		"female titles"
+		"gender-neutral titles"
+	word
+		"'s "
```
Pointedstick: All the possessive titles and names were simplified to go here.

data/names.txt
```diff
@@ -2158,6 +2159,7 @@ phrase "pirate"
 		"Bony Fingers"
 		"Yo Ho Ho"
 		"Low Blow"
+		"bilgewater"
```
KiLEdEnNis: lacks a capitalized letter, should be "Bilgewater"

Pointedstick: Already fixed in the latest commit!

KiLEdEnNis: oops, my bad :P

Issue #1627, opened 2016-09-25
The Bastion and Splinter have crew complements that are too high given the resources of a player considering purchasing them


Pointedstick:
This is a follow-on to issue https://github.com/endless-sky/endless-sky/issues/1598, which made me realize that the Mule's superiority is exacerbated by issues with a number of other Medium Warships, which are not as good as they should be. In this issue, I will focus on an issue that affects both the the Bastion and Splinter, which makes them less than appealing for a player shopping for a Medium Warship: their high crew complements.

The splinter is an "entry-level" Medium Warship, the most affordable of all of them and with some favorable similarities to a freighter, such as a generous cargo bay and good hyperspace range. However, its affordability is undercut by its relatively high crew complement of 15, greater than all other Medium Warships save for the Bastion. This gives it a daily cash burn rate that quickly eliminates the benefit of a low purchase price, making the player realize that saving up a wee bit more for a different Medium Warship that's cheaper to operate would be a better choice.

The Bastion suffers from the same problem, but magnified: with 35 crew, a 3,400 credit/day burn rate is very high for a player who is looking to spend only about 5m for a ship. I recommend that each one be reduced by 40-50%.

Other than this, the Bastion and Splinter are reasonable ships.

PR #3263, opened 2017-11-23
Reduce Bastion and Splinter minimum crew complements

Pointedstick:
Right now their high minimum crew complement makes their daily cost is too high given their capabilities. This should make them more attractive to players in early-to-mid Tier 1 gameplay. This fixes https://github.com/endless-sky/endless-sky/issues/1627

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -315,8 +315,8 @@ ship "Bastion"
 		"cost" 3560000
 		"shields" 6700
 		"hull" 4200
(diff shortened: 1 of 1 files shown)


KiLEdEnNis:
I think these 2 ships are quite a bit underrated because of the high daily cost for what ya get
I'm not sure if it's too low now or fine, but either way I like this change

Pointedstick:
Yay, thanks!

PR #3269, opened 2017-11-24
Coalition "Tourists Home" appear to have attribute applied wrong way

luiges90:


jafdy:
Umm, no.
Tourists typically go from non-tourist destinations to tourist destinations on the way out and go from tourist destinations to non-tourist destinations on the way home.
So they where the correct way round.

Disiuze:
@jafdy, this PR makes it so that tourists who need to go home have tourist planets as source and planets meeting the destination criteria to be their destination as opposed to having to those tourist planets as the destination.

Janaszar:
Only thing I can think of to make this clearer/help out is to rename (name attribute) the missions themselves as "Return Tourists to <planet>", as the going to and return from missions both use "Vacationers to <planet>"

jafdy:
I am so sorry, for some reason I read the red lines as added code and the green lines as deleted code.

PR #3270, opened 2017-11-24
Adjusted the Meteor Missile's description 

Pointedstick:
Now it to provides some more hints about how best to use and defeat them.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -638,7 +638,7 @@ outfit "Meteor Missile Launcher"
 		"hull damage" 80
 		"hit force" 30
 		"missile strength" 4
-	description "The Meteor Missile is the cheapest and simplest homing weapon available in human space. The tracking system on these missiles is notoriously unreliable, and they are easily destroyed by anti-missile systems, but in large numbers or when fired at a slow-moving target they can be deadly."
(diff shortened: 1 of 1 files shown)


data/weapons.txt
```diff
@@ -638,7 +638,7 @@ outfit "Meteor Missile Launcher"
 		"hull damage" 80
 		"hit force" 30
 		"missile strength" 4
-	description "The Meteor Missile is the cheapest and simplest homing weapon available in human space. The tracking system on these missiles is notoriously unreliable, and they are easily destroyed by anti-missile systems, but in large numbers or when fired at a slow-moving target they can be deadly."
+	description "The Meteor Missile is the cheapest and simplest homing weapon available in human space. The missile's infrared tracking system can lose its lock on ships with a low heat signature, and it is easily destroyed by anti-missile systems. Nevertheless, Meteors can be deadly in large numbers and against poorly-defended targets."
```
endless-sky: The "it is" doesn't quite work here, because grammatically "it" would be referring to the tracking system, not the missile itself. Maybe something like, "Meteors are quickly destroyed by anti-missile systems, and their infrared tracking systems have trouble homing in on targets with low heat levels. Nevertheless, they can be deadly..."

Issue #2908, opened 2017-08-10
More Trinket Outfits


Amazinite:
By trinket outfits, I mean stuff like the Quantum Keystone. Right now it's really an oddity. It's the only outfit in the game that seems to serve no purpose once you first find it, but unlocks access to secret things such as new jobs and most importantly easy access into the Ember Waste. Since it sticks out like a sore thumb, maybe it would be worth adding other trinket outfits that you could install, mainly in human space so that the player doesn't find the Quantum Keystone to be an oddity. These trinkets would of course have a purpose... mostly. They could unlock jobs similar to the Quantum Keystone in Hai space or even spaceport missions on more superstitious worlds. 

Maybe one or two could be worthless just for the heck of it, but the addition of such trinkets would 1) make the Quantum Keystone less odd once it is found and 2) make it more interesting to those who actually figure out that the trinkets that they have been installing in human space allow them to do new things, and thus they would be interested in finding the purpose of the Quantum Keystone. And I think most importantly, 3) it would allow us to build lore around new topics that would previously be kind of difficult to explore without being very clunky at the start (e.g. maybe the player has a religous trinket on their ship and is approached by a group of preachers who wish to be transported in a "holy ship." I don't exactly see how they would find your ship holy otherwise 😄).

Some trinkets could have very explicit uses (e.g. someone literally points it out after you land, hinting to the player that such trinkets may trigger missions) and others implicit ones (strange jobs that you weren't getting before the trinket was installed but that don't neccessarily say that the trinket is the cause of them). Overall, I think it would make for a much more interesting universe.

solardawning:
Absolutely. This is simple enough content that it just needs someone to do the legwork making those missions. =)

Amazinite:
Yeah, I've placed it on my (long) to-do list of things to PR. Just wanting to get feedback on the general idea.

tehhowch:
This is a more general version of #2661 and is something that should be worked towards :+1:

jlnxr:
Idea for trinket: a set of scientific scanners that serve no obvious in game purpose, but would allow you to be offered specific scouting/research missions. But exactly a religious article trinket, but a way to have an odd no obvious purpose outfit be valuable with cool missions. 

Ex. "While walking through the spaceport you're stopped by a man you've never seen before. 'I noticed your ship has a set of <whatever> scanners. I was hoping to find someone with that equipment to head to <destination> and take some <whatever> readings. I'd be able to pay you <payment>'"

[...]

PR #3272, opened 2017-11-24
Lowered Star Queen minimum crew

Nescio0:
The Star Queen currently is not only the least efficient passenger transport, it's actually worse than almost all cargo freighters. Let's have a look at passenger to crew ratios (higher values are better):

```
Bounder       :  17/1  - 1 =  16    : 1
Blackbird     :  28/3  - 1 =   8.33 : 1
Heavy Shuttle :   8/1  - 1 =   7    : 1
Mule          :  43/6  - 1 =   6.2  : 1
Scout         :  12/2  - 1 =   5    : 1
Shuttle       :   6/1  - 1 =   5    : 1
Arrow         :   5/1  - 1 =   4    : 1
Hauler        :  12/3  - 1 =   3    : 1
Hauler II     :  12/3  - 1 =   3    : 1
Hauler III    :  12/3  - 1 =   3    : 1
Argosy        :  14/4  - 1 =   2.5  : 1
Bactrian      : 245/70 - 1 =   2.5  : 1
Freighter     :   7/2  - 1 =   2.5  : 1
Flivver       :   3/1  - 1 =   2    : 1
Clipper       :   9/3  - 1 =   2    : 1
Star Barge    :   3/1  - 1 =   2    : 1
Bulk Freighter:  18/6  - 1 =   2    : 1
*Star Queen*  : 112/43 - 1 =   1.6  : 1
Behemoth      :  30/12 - 1 =   1.5  : 1
```
As you can see currently only the Behemoth is worse than the Star Queen. By lowering the minimum crew to 27, the Star Queen becomes twice as efficient, with a ratio of 3.14:1, putting it at least above the freighters, although it remains the worst of all passenger transports.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -2695,7 +2695,7 @@ ship "Star Queen"
 		"cost" 5500000
 		"shields" 4100
 		"hull" 2200
-		"required crew" 43
+		"required crew" 27
 		"bunks" 112
 		"mass" 230
 		"drag" 5.5


Wrzlprnft:
The star queen is meant to be a luxury ship. which means at least one cook per....

the issue with the star queen is that there us no luxury attribute or outfit yet, which would allow for high paying luxury missions.

but for sure, it never meant to be effective ;)

Pointedstick:
> but for sure, it never meant to be effective ;)

And that's a problem. Why bother putting it in the game if there's no reason for a player to actually use it? I support this change.

endless-sky:
The Star Queen is not supposed to be the "best" passenger transport. It's supposed to be overpriced and useless.

Bladewood:
Try reasoning nodes or control transceivers. Their sole use is for mission ships.

Pointedstick:
> It's supposed to be overpriced and useless.

So why bother making it a player-purchasable ship? I've never been found of games with options that are just... bad. You've often mentioned that one of your goals is to reduce parts of the game that are just annoying. I think this counts as one. A player who buys a Star Queen for flavor of head-canon reasons with the expectation of maybe running lucrative high-dollar passenger jobs is going to be disappointed and frustrated. And for what gain? Just to taunt them for making a dumb choice? Clearly _someone_ buys and operates Star Queens in this universe? Are they all idiots?

Nescio0:
With this change the star queen remains the least cost-effective of all passenger transports. It just becomes slightly better than any freighter.

Bladewood:
Lore reasons. Same reason Navy ships are offered in shipyards even though they require licenses.

MCOfficer:
i'm tempted to agree, that sounds a bit like simulating variety by having useless options. however, it also reminds me of what valve tried with one cs:go update, which was was balancing weapons to get them back in the meta. If there's one truth in video games where you choose your equipment, it is that there is always a shitty option - in the case of ES, these options are star queen, roost, etc.. trying to change that will only put other ships in the bottom position, so variety is not an argument.

Janaszar:
I would expect that the extra crew are people like butlers and chefs who work on making the Star Queen a luxury ship.  If the game has the aspects added to it (new/different classes for passengers, separating crew bunks from passenger bunks...), I would think that you would end up using it as a transport for high-class individuals who are fine with paying a lot more than standard fare.

Wrzlprnft:
I personally hope it is still in there because MZ hopes that someone comes along with a sophisticated but easy sytem for luxury, not solely depending on a luxury outfit, together with jobs...

KiLEdEnNis:
that it has to be bad in general ok, but to me it wouldn't hurt to lower the required crew by, say, 10?
although special missions ya only get if ya have a Star Queen would make it nice

Amazinite:
The Star Queen has the worst bunks/required crew, but it has the highest number of spare bunks among the human transports. I'd say the real solution isn't to change the Star Queen, but instead to provide the Star Queen with special jobs that make it worth using. (#2908 comes to mind, but there are a bunch of issues that you can look up that ask for the Star Queen in specific to receive luxury jobs and not just any ship with luxury items. #2911 also comes into play, as a job could be triggered by an attribute on a luxury item like mentioned in #2908, but the Star Queen could also have that attribute built in, therefore making it a ship that you don't need to spend extra money on in order to take luxury jobs. Heck, the description of the Star Queen even says it has a bunch of luxury items preinstalled, but those have no effect in game.)

greatkek:
I agree 100% with this change. It is supposed to be the best passenger ship, not one of the worst. +1 for the change commit

MCOfficer:
it's supposed to be the most luxury passenger ship, not the most effective one. If you think about it, being a luxury liner almost automatically makes a ship ineffective for transporting everyday-passengers. Gotta give probs to @Nescio0 for detecting this weak point in the game (i never made a good PR myself), but i'm with @Amazinite here.

greatkek:
Why not alter the model to be a little longer or something, and make it just a bit more pricey, and sell it as a recent model of it? I'm sure you could photo shop that thinner portion to be a bit longer and increase the overall size by like, 1.1x. This would of course, be a seperate ship and not a replacement. Best of both worlds, no?

KiLEdEnNis:
what's the practical use of that? ya just add another hardly different useless ship for no reason

MCOfficer:
a ship being useless doesn't mean it shouldn't be added. if you have a variety of ships, you will always have some that are less useful. just look at the new cool breeze - no one will fly an interceptor in the wanderer story line.

still, just copying a ship seems pretty cheap to me. there are only 2 copied models in ES, as far as i'm aware (the roost series and the barbs).

ghoulavenger:
I thought the whole point of the Star Queen was to be a luxury liner that we were supposed to scoff at during the FW campaign (for being a gaudy piece of crap). I'm not against it getting a slight buff or two, but I never thought the point was to be a "good" ship. In fact, I'd argue there are a lot of "bad" ships, just look at most of the syndicate line (budget and not performance is pretty much their modus operandi).

Nescio0:
Again I'd like to emphasize that with this proposal the Star Queen remains the least efficient passenger transport. Nor am I opposed to Star Queen-specific missions. One improvement does not exclude another.

MessyMix:
I think the Star Queen-specific missions are a step in the right direction, because I feel like the Star Queen is to starships as a limo is to cars; that is, it's meant to be extravagant and serve really no practical purpose except luxury. I think giving it more functionality taints this image, whereas the SQ-specific missions from VIPs help reinforce it, which is why I'm leaning to the latter over the former.

I just don't see why the SQ needs to be _more_ efficient than before if it's still going to be the least efficient transport. I don't know, I feel like it should exude an aura of inefficiency, :P.

Nescio0:
> I just don't see why the SQ needs to be more efficient than before if it's still going to be the least efficient transport. I don't know, I feel like it should exude an aura of inefficiency. 😛

To make it at least more efficient than a freighter or hauler; in our world luxury cruise ships with swimming pools, tennis lawns, gilded furniture, marble bathrooms, etc. are also more cost effective than using a default container ship as a passenger transport.

KiLEdEnNis:
> To make it at least more efficient than a freighter or hauler; in our world luxury cruise ships with swimming pools, tennis lawns, gilded furniture, marble bathrooms, etc. are also more cost effective as a passenger transport than using a default, un-refitted container ship.

is it? how many people can you hide in just 1 container? and how many containers can a freighter hold?

on a more serious note; the Star Queen may have giant-ass rooms for every passenger, giant-ass luxury rooms etc. and a lot of staff to man all sorts of posts, which is a reasonable explanation to be this trash

MCOfficer:
we're talking about actual bunks here, not stuffing people in the cargo hold. he's got a point.

Issue #2130, opened 2017-02-11
Finisher torpedo explosion sound should be way more impressive


Pointedstick:
Given the weapon's stats and the visual impressiveness of the explosion it produces, I'm pretty underwhelmed by the explosion sound, which feels way too small.

Bladewood:
Along with that, let's have it make a cool sound when it become an active finisher!

Lineth:
I'm now picturing the torpedoes to make the Mortal Kombat "FINISH HIM." voice when launched.

Amazinite:
New sound for the Bubble Gun too?

KiLEdEnNis:
bubble gun description
"my bubbles don't only make ships wet ( ͡° ͜ʖ ͡°)"

Pointedstick:
Oops, I forgot to use the new Torpedo hit sound for Finishers with https://github.com/endless-sky/endless-sky/pull/3869.

Will fix.

Amazinite:
The Finisher creates both the torpedo sound effect (as of #7089) and the small explosion sound effect. @Pointedstick Would you consider this done, or does the Finisher explosion still need more oomph to it?

Pointedstick:
Yeah, this is fine now. Thanks for following up!

PR #3274, opened 2017-11-25
Better missile launch and explosion sounds

Pointedstick:
I was going to include this as part of the missile overhaul PR I'm putting together, but figured it would stand better on its own. IMHO these sounds drastically improve the audio experience of using and reacting to missiles.

- Meteor and Sidewinder launch sounds are a bit more raw and harsh, and not muted or cartooney: https://webmshare.com/BjjmO
- Most medium-sized missiles get a more impressive and distinctive-sounding hit sound: https://webmshare.com/nPPoP
- Large missiles (Torpedoes, Finishers) get a huge and impressive hit sound: https://webmshare.com/4BB94

Also resolves https://github.com/endless-sky/endless-sky/issues/2130


Wrzlprnft:
+1 for not stuffing too much in one pr that might get rejected because of one feature ;)

I haven't played with this, but the Sounds are good in their own.

MCOfficer:
i like those, but both the meteor and torpedo hit sound are cut off at the end - won't notice that in a battle, but it sounds odd when there's nothing to distract you.
audacity_2017-11-25_12-08-28

Pointedstick:
Thanks, fixed!

MCOfficer:
i still feel like both hit sounds are brickwalling, and you can ever-so-slightly hear a click at the end of the sidewinder sound. that's super-high level criticism, though. i doubt one could hear that ingame.

copyright
```diff
@@ -386,6 +386,21 @@ Copyright: Copyright Mike Koenig
 License: CC-BY-SA-3.0
 Comment: Taken from http://soundbible.com/1467-Grenade-Explosion.html
 
+Files: sounds/missile hit.wav
+Copyright: Copyright "Nbs Dark"
+License: Public Domain
+Comment: Taken from https://freesound.org/people/Nbs%20Dark/sounds/94185/
```
endless-sky: The credit for the torpedo hit is missing.

Also, because of how this file is parsed, you need to use a '?' in place of spaces in the file names, i.e. "sounds/missile?hit.wav".

copyright
```diff
@@ -386,6 +386,21 @@ Copyright: Copyright Mike Koenig
...
+
+Files: sounds/meteor.wav
+Copyright: Copyright "18hiltc"
+License: CC-BY-3.0
+Comment: Taken from https://freesound.org/people/18hiltc/sounds/202725/
+
+Files: sounds/sidewinder.wav
+Copyright: Copyright "Robinhood76"
+License: CC-BY-NC-3.0
```
endless-sky: Non-commercial licenses aren't compatible with the Debian software guidelines, so this sound can't be used in Endless Sky.

copyright
```diff
@@ -386,6 +386,21 @@ Copyright: Copyright Mike Koenig
...
 Comment: Taken from http://soundbible.com/1467-Grenade-Explosion.html
 
+Files: sounds/missile hit.wav
+Copyright: Copyright "Nbs Dark"
+License: Public Domain
+Comment: Taken from https://freesound.org/people/Nbs%20Dark/sounds/94185/
+
+Files: sounds/meteor.wav
+Copyright: Copyright "18hiltc"
+License: CC-BY-3.0
```
endless-sky: This isn't currently one of the licenses used in the game, which would require pasting the license's whole text in this file and in other places where license info is tracked. But, since you're creating a derivative work from that sound, you could also just say that the derivative version is CC-BY-SA-3.0, which *is* a license used for other game assets.

copyright
Pointedstick: My bad, the torpedo hit sound is one I've had on my computer for nearly 20 years, taken from a public domain sound website that's long since ceased to exist.

copyright
Pointedstick: Fixed.

copyright
Pointedstick: Fixed; I found another sound.

Pointedstick:
Hmm, that did not work.

Pointedstick:
See https://github.com/endless-sky/endless-sky/pull/3869 instead, sorry folks.

Issue #3268, opened 2017-11-24
Consider making missile weapons more dangerous and less numerous


Pointedstick:
## The issue ##
Endless Sky's missiles (particularly Human ones) have the following characteristics:
- Individual missiles don't do much damage relative to most ships' shields and hull (e.g. it takes three shots from what's called a **torpedo** to destroy a **shuttle**)
- Because of the above, missiles are spammed, which makes fights against missile-armed adversaries merely annoying rather than nerve-racking. Each individual missile doesn't really do much damage so it's not that big of a deal if you get hit with 10 or 20. The missile spam is particularly ridiculous in Hai-on-Hai fights, where there will be like 200 missiles in flight at any one point in time. It's just silly.
- Because of the above, missiles are extremely expensive damage-for-damage when compared to guns, but don't generally make up for it in effectiveness. Using them becomes a huge money drain, and after nearly every battle, you'll be nearly out of missiles and have to refill for tens or hundreds of thousands of credits.
- Missile Jammers are nearly useless when weighed against their tonnage requirement.

Overall, missiles weapons are annoying to fight against and annoying to use. I think this needs to be remedied.

## The proposal ##
- Make missiles do much, much more damage. Like, factor-of-4 more damage. Prices should also rise a bit.
- Make missile launchers fire much more slowly, have much smaller magazines, and weigh less (this would also increase the utility of the missile box outfits).
- Make the large heavy anti-capital-ship missiles [much more clumsy](https://github.com/endless-sky/endless-sky/issues/1599#issuecomment-256526764) so that they're nearly useless against small and medium sized ships. This would increase the utility of Sidewinders and even humble Meteors, because right now Torpedo-class weapons are practically the only worthwhile guided Human missile weapons.

## Why it would be a gameplay improvement ##
Primary reason: Missiles would actually be fun to use and defend against.

Secondary reasons: by increasing the danger of each individual missile, the player's missile defense strategy would be something to actually consider much of the time. Dodging them would become a more useful skill. Jammers, anti-missile defenses, and cool-running ships would become appropriately more important rather than afterthoughts. And with missiles' increased damage and dangerousness, they would become actually worth using against the AI. Also, with smaller launchers that don't have as much capacity, it would be worth putting missile launchers on fighters, since this would make them actually punch above their weight class like real fighters do--but they wouldn't be able to spam missiles with their low capacity.

Amazinite:
😆 
[...]

PR #3275, opened 2017-11-25
Improve the game's missile experience

Pointedstick:
All right everyone, let's give this a whirl...

# The changes #
In general, all secondary missile-type weapons received the following changes:
- Launcher reload time doubled; launchers fire half as fast
- Launcher capacity reduced to 10% of previous value
- Launchers reduced to roughly 50% of previous size and mass
- Missile storage outfits increased in size slightly to make it harder to fit them on fighter-sized ships, to limit them to using the launchers' internal magazines as much as possible (like, you know, real fighters)
- Missile damage and impact force quadrupled (yes you read that right, and no it's not crazy...)
- Missile cost increased by a factor of 1.6x
- Some missiles' lifespans reduced

Additional changes to specific missiles:
- Meteor missiles turn faster but have slightly dumber homing and tracking
- Sidewinder missiles turn much faster
- Torpedoes and Typhoon Torpedoes turn and move _much_ more slowly
- Torpedoes and Typhoon Torpedoes have higher missile strength to make up for their lower speed, which exposes them to more anti-missile fire now
- Torpedoes have a more impressive-sounding explosion sound
- Heavy Rocket damage and reload times changed by smaller amounts since Heavy Rockets were already better than other human missile options
- Nuclear Missile speed reduced to 33% of previous value to give ships and anti-missile systems more time to deal with them, to compensate for their now-outrageous 27,000 shield damage and 21,000 hull damage

All races' ships received the following changes:
- Many human fighter-sized ships replaced 1-2 guns with missile launchers and a small number of missiles (1-8).
- Larger ships gained missile box/storage racks to restore some of their lost missile capacity, but to a lower level (fewer missiles overall).
- Many (but not all) larger ships that previously lacked any anti-missile defenses replaced an offensive turret with an Anti-Missile turret of some sort.

# The in-game effects #
**Player effect:** Overall, missiles are now fun to use, fun to dodge, and fun to integrate into your ship builds. Players might rationally choose to use Radar Jammers, replace an offensive turret or two with more anti-missile, and carry multiple types of missiles themselves for dealing with different-sized ships. A Firebird-sized player ship might commonly fill the gun ports with two Heavy Lasers, a Torpedo Launcher, and a Sidewinder Missile Launcher, and that would be a fun and effective build.

**Game effect:** The net effect is that more ships fire missiles, but overall there are fewer missiles flying through the air at any given time. Each missile is much more dangerous, and ships with no thought to anti-missile defense are more vulnerable. The overall effect is much more muted than might be expected because anti-missile turrets now generally have fewer missiles to deal with than they did before, so they perform a bit better.

Carried fighters are actually worth using in their default configurations. Heavy interceptor craft carrying big missiles are actually dangerous.

Meteors are now stronger against large ships and slow heavy interceptors without anti-missile defenses.

Sidewinders are now useful against fast interceptor-sized craft and somewhat effective against ships with moderate anti-missile defenses.

Torpedoes are now very effective against large ships--especially those lacking anti-missile defenses--but virtually useless against anything fast.

Missiles cost more so players who use them will expend more money, but more players will actually want to use them because they're both fun and effective.

This fully resolves https://github.com/endless-sky/endless-sky/issues/3268. I'm sure it needs balance tweaks and changes, but I've been playing with this all day and it's an absolute blast. Please give it a try yourself!

### DIFF
--- a/data/coalition weapons.txt
+++ b/data/coalition weapons.txt
@@ -102,7 +102,7 @@ outfit "Finisher Pod"
 		"inaccuracy" 5
 		"velocity" 10
 		"lifetime" 60
-		"reload" 80
+		"reload" 160
 		"firing energy" 120
 		"firing heat" 20
 		"acceleration" .8
@@ -112,7 +112,7 @@ outfit "Finisher Pod"
 		"radar tracking" .8
 		"missile strength" 50
 		"submunition" "Active Finisher" 1
-	description `Finisher Torpedoes are one of the Heliarchs' most dreaded weapons. In fact, in one way or another most of their other weapons merely exist to hold a ship in place for long enough for a barrage of Finishers to destroy it.`
+	description `Finisher Torpedoes are one of the Heliarchs' most dreaded weapons. Made for destroying Quarg warships, Finishers deal an enormous amount of damage, though they are overkill against less powerful vessels. In one way or another, most of the Heliarchs' other weapons merely exist to hold a ship in place for long enough for a barrage of Finishers to destroy it.`
 
 outfit "Active Finisher"
 	weapon
@@ -130,14 +130,14 @@ outfit "Active Finisher"
 		"radar tracking" .8
 		"optical tracking" .5
 		"missile strength" 50
-		"shield damage" 1500
-		"hull damage" 1900
-		"hit force" 240
+		"shield damage" 6000
+		"hull damage" 7600
+		"hit force" 1000
 
 outfit "Finisher Torpedo"
 	plural "Finisher Torpedoes"
 	category "Ammunition"
-	cost 32000
+	cost 51200
 	thumbnail "outfit/finisher"
 	"mass" .5
 	"finisher capacity" -1
--- a/data/wanderer outfits.txt
+++ b/data/wanderer outfits.txt
@@ -220,11 +220,11 @@ outfit "Thunderhead Launcher"
 		"Wanderer Outfits"
 	cost 450000
 	thumbnail "outfit/thunderhead launcher"
-	"mass" 14
-	"outfit space" -26
-	"weapon capacity" -26
+	"mass" 7
+	"outfit space" -13
+	"weapon capacity" -13
 	"gun ports" -1
-	"thunderhead capacity" 40
+	"thunderhead capacity" 5
 	weapon
 		sprite "projectile/thunderhead"
 			"no repeat"
@@ -240,7 +240,7 @@ outfit "Thunderhead Launcher"
 		"inaccuracy" 10
 		"velocity" 4
 		"lifetime" 400
-		"reload" 75
+		"reload" 150
 		"firing energy" 18
 		"firing heat" 36
 		"acceleration" 1.1
@@ -250,15 +250,15 @@ outfit "Thunderhead Launcher"
 		"infrared tracking" .9
 		"radar tracking" .6
 		"missile strength" 12
-	description "A prime example of Wanderer ingenuity, Thunderhead missiles each carry five submunitions that are capable of independent tracking. Even if the target is able to shoot down or evade a few of them, the remaining missile fragments will still find their mark."
+	description "A prime example of Wanderer ingenuity, Thunderhead missiles each carry five submunitions that are capable of independent tracking. Even if the target is able to shoot down or evade a few of them, the remaining missile fragments will still find their mark. Each launcher's internal magazine can hold five missiles."
 
 outfit "Thunderhead Missile"
 	category "Ammunition"
-	cost 1800
+	cost 2880
 	thumbnail "outfit/thunderhead"
 	mass .3
 	"thunderhead capacity" -1
-	description "Thunderhead cluster munitions are long-range Wanderer missiles that split into five fast and agile submunitions when they are close to their target. Each launcher can hold up to 40 of these."
+	description "Thunderhead cluster munitions are long-range Wanderer missiles that split into five fast and agile submunitions when they are close to their target."
 
 outfit "Thunderhead"
 	weapon
@@ -269,16 +269,16 @@ outfit "Thunderhead"
 		"hit effect" "small explosion"
 		"inaccuracy" 90
 		"velocity" 2
-		"lifetime" 400
+		"lifetime" 300
 		"acceleration" 1.4
 		"drag" .1
 		"turn" 3
 		"homing" 4
 		"optical tracking" .9
 		"radar tracking" .6
-		"shield damage" 110
-		"hull damage" 80
-		"hit force" 20
+		"shield damage" 440
+		"hull damage" 320
+		"hit force" 80
 		"missile strength" 3
 
 outfit "Thunderhead Storage Array"
@@ -288,7 +288,7 @@ outfit "Thunderhead Storage Array"
 	cost 36000
 	thumbnail "outfit/thunderhead storage"
 	"mass" 2
-	"outfit space" -8
+	"outfit space" -14
 	"thunderhead capacity" 20
 	ammo "Thunderhead Missile"
 	description "The Thunderhead Storage Array is used to store extra ammunition for Thunderhead Launchers."
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -587,11 +587,11 @@ effect "flamethrower hit"
 
 outfit "Meteor Missile"
 	category "Ammunition"
-	cost 500
+	cost 800
 	thumbnail "outfit/meteor"
 	"mass" .2
 	"meteor capacity" -1
-	description "A Meteor Missile serves as ammunition for the Meteor Missile Launcher. Each missile launcher has the capacity for 35 missiles; you cannot install missiles if you have not first purchased a launcher to store them in."
+	description "A Meteor Missile serves as ammunition for the Meteor Missile Launcher. You cannot install missiles if you have not first purchased a launcher to fire them."
 
 outfit "Meteor Missile Box"
 	plural "Meteor Missile Boxes"
@@ -599,7 +599,7 @@ outfit "Meteor Missile Box"
 	cost 9000
 	thumbnail "outfit/meteor storage"
 	"mass" 1.4
-	"outfit space" -5
+	"outfit space" -8
 	"meteor capacity" 18
 	ammo "Meteor Missile"
 	description "The Meteor Missile Box is used to store extra ammunition for Meteor Missile Launchers."
@@ -609,10 +609,10 @@ outfit "Meteor Missile Launcher"
 	cost 15000
 	thumbnail "outfit/meteor launcher"
 	"mass" 3
-	"outfit space" -10
-	"weapon capacity" -10
+	"outfit space" -5
+	"weapon capacity" -5
 	"gun ports" -1
-	"meteor capacity" 35
+	"meteor capacity" 4
 	weapon
 		sprite "projectile/meteor"
 			"no repeat"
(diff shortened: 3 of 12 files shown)


luiges90:
When you are piloting a shuttle and ready to depart, and a torpedo-armed pirate is already waiting on top of you...  (I lost count how many times it happens)

And does I read it right? The nuke can apparently one-shot every human ship? 😱

Pointedstick:
Play it and see. Torpedoes are almost impossible to get hit by if you're not a bus. Shuttles were already made of paper. Nukes are now really slow. Some things are obviously gonna need changing of course, but give it a whirl with Meteors, Sidewinders, and Torpedoes. IMHO they feel so much more like real missiles should.

Amazinite:
Initial thoughts:
* Since missiles fire half as fast, AM turrets probably should as well.
* Currently, `Mass + Capacity * Mass of ammo = outfit space` for launchers and capacity upgrades. With this PR, that goes entirely out the window, and it probably shouldn't. (e.g. your Sidewinder takes up 7 outfit space but when fully loaded only weighs 5 tons.)
* The changes to the nuke are questionable. Yes, it's slower, but it still has the massive trigger radius and blast radius to it, and given that it's a one-shot weapon, ships will currently fire them all at once regardless of the reload of the weapon. (`stream` doesn't seem to work on nukes.) This'll likely make the campaign harder, especially the last mission since you won't be able to tank one or two nukes anymore.
* Speaking of the campaign, have you tested the campaign(s) with these changes? How does it affect the battles? Does the side that the player fights on still just barely lose without the player's help? The Navy uses a lot of missiles so this changes a lot.

tehhowch:
If we are boosting damage, then it's time for inverse square explosions :)

Amazinite:
Concerns and observations so far:
* The substantial reduction in size of the launchers means that you can fit more launchers on smaller ships, ruining the balance between small and large ships when missiles are at play. And since the damage per missile has been quadrupled, 4 heavy rockets on a Fury = instant death to pretty much anything smaller than a medium warship, and a second volley kills anything weaker than a heavy warship.
* Missiles can be serious overkill now. Possibly switch some missiles over to using `stream` instead of `cluster`
* Missiles are NOT fun when in smaller ships. Not every player is going to be able to dodge missiles while in a ship without a turret to defend them. In some cases the player can get killed or disabled in a single volley of missiles early on. I feel that overall this is going to increase the complaints about dying early game.
* NPCs don't try to actively dodge missiles, so a small player ship with missiles is able to punch way above its weight. (I had a Sparrow with 2 Hai Trackers and was having no problem taking down multiple medium warships in asteroid-sparse systems.)
* Some of the variants you made have negative outfit space. The ones I've found so far are the Manta at -5 and Frigate at -1.
* I disagree with the complete reliance on capacity upgraders to even make missiles a viable weapon. It just seems like an unnecessary hurdle to using missiles effectively, especially since you also increased the cost of missiles.

Suggestions:
* Revert outfit space and mass changes to both launchers and capacity upgraders.
  * Restores balance between categories, so two or three interceptors fitted with heavy rockets or torpedos won't be able to immediately gank a medium warship or small heavy warship.
* Revert addition of missiles to a bunch of ship variants.
  * 2 blasters, a meteor, and a heavy rocket on a stock Fury just doesn't look nice, and the increase in the number of ships with missiles just means an increase in the number of ships that have the ability to gank a novice pilot.
* Instead of 1/10th the capacity of current launchers, perhaps use 1/4th the capacity. Capacity upgraders generally provide an additional 50% of the current capacity, meaning that a single capacity upgrader would triple one's capacity under this PR.
  * This, I feel, strikes a more proper balance between the capacity of the launcher and the capacity of an upgrade, as an upgrader giving you about 5 times what a launcher holds is a little insane, and capacity upgraders are only necessary installs for people looking to use their missiles a lot.
 * I think the doubled reload times are great, but the quadrupled damage numbers are insane. A lot of changes are compounding though to make missiles so powerful under this PR, so I wouldn't change the damage any further as of right now.

data/coalition weapons.txt
```diff
@@ -112,7 +112,7 @@ outfit "Finisher Pod"
 		"radar tracking" .8
 		"missile strength" 50
 		"submunition" "Active Finisher" 1
-	description `Finisher Torpedoes are one of the Heliarchs' most dreaded weapons. In fact, in one way or another most of their other weapons merely exist to hold a ship in place for long enough for a barrage of Finishers to destroy it.`
+	description `Finisher Torpedoes are one of the Heliarchs' most dreaded weapons. Made for destroying Quarg warships, Finishers deal an enormous amount of damage, though they are overkill against less powerful vessels. In one way or another, most of the Heliarchs' other weapons merely exist to hold a ship in place for long enough for a barrage of Finishers to destroy it.`
```
Amazinite: In general, I try to avoid ship and outfit descriptions that do nothing but reiterate the information you can get from the weapon's stats. It's much better for the descriptions to provide some sort of flavor text (historical information, etc.) that is interesting or funny in its own right.

data/wanderer outfits.txt
```diff
@@ -250,15 +250,15 @@ outfit "Thunderhead Launcher"
 		"infrared tracking" .9
 		"radar tracking" .6
 		"missile strength" 12
-	description "A prime example of Wanderer ingenuity, Thunderhead missiles each carry five submunitions that are capable of independent tracking. Even if the target is able to shoot down or evade a few of them, the remaining missile fragments will still find their mark."
+	description "A prime example of Wanderer ingenuity, Thunderhead missiles each carry five submunitions that are capable of independent tracking. Even if the target is able to shoot down or evade a few of them, the remaining missile fragments will still find their mark. Each launcher's internal magazine can hold five missiles."
```
Amazinite: >In general, I try to avoid ship and outfit descriptions that do nothing but reiterate the information you can get from the weapon's stats. It's much better for the descriptions to provide some sort of flavor text (historical information, etc.) that is interesting or funny in its own right.

data/weapons.txt
```diff
@@ -688,20 +688,20 @@ outfit "Sidewinder Missile Launcher"
...
-		"shield damage" 100
-		"hull damage" 70
-		"hit force" 25
+		"shield damage" 400
+		"hull damage" 280
+		"hit force" 100
 		"missile strength" 12
 	description "The Sidewinder Missile is designed by Lovelace Labs, which is ironically also the designer of the Anti-Missile Turret. Rumors say that the two engineering teams at Lovelace who make the missiles and the anti-missiles are locked in furious competition with each other: the result has been generation after generation of smarter missiles and then smarter turrets to fend them off."
-	description "	Although not as powerful or cheap as the popular Meteor missiles, Sidewinders have several advantages, including their better tracking systems and protective casings that make them much harder for anti-missile systems to destroy."
+	description "	Although not as powerful or cheap as the popular Meteor missiles, Sidewinders have several advantages, including their better tracking systems and protective casings that make them much harder for anti-missile systems to destroy. Each of these launchers includes a five-round internal magazine."
```
Amazinite: You know what I'm going to post...

data/weapons.txt
```diff
@@ -858,56 +858,55 @@ outfit "Typhoon Launcher"
...
-		"shield damage" 610
-		"hull damage" 610
-		"hit force" 70
-		"missile strength" 30
-	description "Typhoon Torpedoes are an improved torpedo technology; they do considerably more damage than ordinary torpedoes. However, they are also much slower, because they are intended mostly for use against large and slow-moving capital ships. As a result, sufficiently skilled pilots can often dodge around them."
+		"shield damage" 2240
+		"hull damage" 2240
+		"hit force" 280
+		"missile strength" 40
+	description "Typhoon Torpedoes are an improved torpedo technology; they do considerably more damage than ordinary torpedoes. However, they are also much slower, because they are intended only for use against large and slow-moving capital ships. As a result, sufficiently skilled pilots can easily dodge around them. Each Typhoon Launcher can hold three torpedoes, but capital ships generally include enough supplementary storage capacity to last whole campaigns."
```
Amazinite: >... avoid... descriptions that... reiterate... the weapon's stats...

data/weapons.txt
```diff
@@ -919,18 +918,18 @@ outfit "Heavy Rocket Launcher"
...
 		"blast radius" 50
-		"shield damage" 700
-		"hull damage" 600
-		"hit force" 80
+		"shield damage" 2000
+		"hull damage" 2000
+		"hit force" 240
 		"missile strength" 16
-	description "Heavy Rockets are the most powerful missile weapon available, but at a price: instead of having a homing system, they simply fire straight forward and rely on a proximity trigger to set them off. Once triggered, however, they damage all ships within their blast radius, making them highly effective against clusters of fighters."
+	description "Heavy Rockets are the most powerful missile weapon available, but at a price: lacking a homing system, they simply fire straight forward and rely on a proximity trigger to set them off. Once triggered, however, they damage all ships within their blast radius, making them highly effective against clusters of fighters. Each launcher can only hold a single rocket; additional capacity requires storage racks."
```
Amazinite: >In...

Nescio0:
Actually I believe you've changed too much at once. I'd suggest to be conservative, and start by doing only the following:
* Halve the number of shots per second of anti-missile turrets
* Halve the number of shots per second of secondary weapons
* Double the damage per shot for all secondary weapons
* Keep everything else unchanged for now

This means that it will take twice as much time before you run out of missiles, and also that missiles become twice as cost effective. If the above works without breaking anything, you could consider increasing the damage per missile even further, and maybe later start tweaking their cost.

Pointedstick:
I considered a series of smaller PRs, but the changes all seemed to work best together. Only doubling damage and halving range and anti-missile speed makes the problem of 

The idea behind the launcher/storage capacity and size changes was to make missile _launching_ equipment more available to small and medium-sized ships, but with a much smaller number of missiles. Only larger ships should have high missile capacity, which is exactly in line with modern fighters and warships: fighters typically carry 4-10 missiles, while missile cruisers might have the capacity to fire 130+. This wasn't reflected in the game before and felt distinctly odd, particularly the Fury variant with 100 Meteors, the Hawk with 40 heavy rockets, etc. But reducing fighter-sized-ship missile capacity without making their weapons deal a lot more damage made them worse.

Fighter-sized ships shouldn't be able to waste multiple medium warships because they shouldn't be able to fit more than a maximum of 1 missile storage outfit, by design. I'll play with the numbers some more to cut that out.

Nescio0:
If you perceive missile launcher size and capacity as a serious problem (I don't; a tracker pod has mass 19, a ion cannon 47), then I'd simply suggest to reduce mass and halve capacity, thus allowing each purchased missile storage one complete refill.
Although related, I believe it's better to do this in a separate proposal.

Pointedstick:
> If you perceive missile launcher size and capacity as a serious problem

Even without any of my changes here, it's already a problem with interceptor-sized craft. A Sparrow with 2 meteors today can fire 70 crappy missiles, which is just all wrong. Missile furies have 140 missiles and aggravate the player to no end by spamming them while running away (infuriating). It makes much more sense, both tactically, and for semi-realism-reasons, for these kinds of ships to be able to fire a maximum of 10-ish much better missiles unless they seriously sacrifice their gun capacity.

Pointedstick:
I may indeed break this out into multiple smaller PRs eventually, but for now I want to do all the changes together to make sure they harmonize properly.

Nescio0:
To clarify, my point is that if you change too much at once, it becomes like comparing apples to pears: two completely different situations. I'm not against an overhaul, I just highly recommend to do it in smaller steps.

Pointedstick:
Yeah, I think you're right. Smaller PRs incoming.

Pointedstick:
Closing to focus on smaller individual PRs instead

Issue #2529, opened 2017-05-19
Should hitForce and blast damage scale with distance from the blast origin?


tehhowch:
This topic was brought up in Discord today among @EndrosG , @Elyssaen , and myself.
If implemented, only ships which are close to the source of the push / blast  will take the full amount of damage. Others will take a scaled amount of the damage. Scaling would apply to any blast, and any damage type (e.g. ion damage taken from an EMP weapon would decrease with distance from the blast).

Refs #1630 

I am seeking feedback on the multiplicative constant used to normalize the damage: it will use the scaling approach [MZ offered](https://github.com/endless-sky/endless-sky/issues/1630#issuecomment-250055568), but will include a multiplicative constant to alter where the post-scaling damage is equal to the current damage values. A constant k = 1 means the specified weapon damage is only true at the blast center, while k = 2 means the specified weapon damage is only true when you are 1 blast radius away. K = 1.0625 means the specified weapon damage is true at 25% of the blast radius, and k = 1.25 moves it to 50%. Currently, the received damage is the specified weapon damage for all positions within the blast radius.
Possible damage scaling profiles:
<details>
  <summary>dealt = k * damage / (1+r²)</summary>

  ![image](https://cloud.githubusercontent.com/assets/20871346/26264499/aadeacc0-3ca2-11e7-8bb3-19370b055e5a.png)
</details>

<details>
  <summary>dealt = k * damage / (1+r²)²</summary>

  ![image](https://cloud.githubusercontent.com/assets/20871346/26264904/28118234-3ca4-11e7-9582-7cdb4d28a64e.png)
</details>

It is also possible to increase the effective blast radius of explosions and the like by modifying [Engine.cpp](https://github.com/endless-sky/endless-sky/blob/d92fc7e506b4e66d33674f27bed49beb5ebd4c6d/source/Engine.cpp#L1342). I would prefer not to increase this value (as that rapidly increases the area spent checking for collisions). In other words, the implemented damage profiles would be as above, from 0..1, and not extending to 1.5. I show that extra bit to indicate what the damage profile could look like if we choose to give a bit of extra distance to the blast radius to compensate for making the explosions weaker.

Elyssaen:
Nice graphing. : ) I'm not sure how obvious it is what you're saying to someone just coming into this, especially if they haven't read #1630.

So to confirm:

1. The topic is having damage (and hit force) from a blast fall off with distance.
2. MZ has suggested a curve that would work, and you're looking to implement it.
[...]

Issue #1630, opened 2016-09-26
Inverse Square Explosions


Nechochwen-D:
Looking over the source as I have been working on various modifications to the game I noticed that there is no provision for an inverse square damage profile for explosions.

I was wondering of this was intentional to save on processing time or merely an oversight. My question is if this is something that could be of value in the game.

Bladewood:
I really want this to happen, but I can't help but feel that it was intentional because physics in space...

KiLEdEnNis:
what does that mean at all; ships imploding on death?

LocalGod79:
@KiLEdEnNis i believe he means that damage from explosions should be less, the further you are from the centre of said explosion.

e.g. at 100px from the explosion centre, take 10000 damage.  at 10000px from the explosion centre, take 100 damage...

KiLEdEnNis:
ah, should be a thing tbh

SolraBizna:
It wouldn't be that processor-intensive compared to what's already going on to detect collisions. If this is added, it'll need a cap, otherwise you'd take insane damage because you happened to be right on top of that Sparrow when it died...

Amazinite:
Well it would probably just use the current damages, just with altered blast radii to account for the tiny damage that would be taken if the blast radii weren't changed.

Nechochwen-D:
I suppose I should give a short explanation of what this is to remove some of the confusion.

As it stands right now when there is an explosion whether from a ship death, a missile, or some other effect if a ship is within the "blast radius" of the explosion it receives full damage from it.

What I propose is a damage profile along these lines, damageMod = ((blastRad / 4) / distance)^2, this would mean that at 1/4 of the "blast radius" you would receive the full damage levels escalating exponentially the nearer you get and deescalating the further you get from the blast. As the largest "blast radius" in the game is 450px from the center of the ship and while it is bad news if you are caught in the blast for these ships it already is. There would be a cutoff for the outer edge of the blast though set at   "blast radius" * 3 at which point the damage would only be 1/144 of the initial.

KiLEdEnNis:
having 2 "weapon" attributes or making the one better like this;
-set initial damage & max radius, then enabling editing the scaling (so damage barely drops at the max range or already gets low near) for editing possibility
that could be an idea too?

retrue:
When capturing robotic Kor ships most of them explode under your nose. If the explosion is bigger when you are closer, capturing ships would be much more difficult.
[...]

PR #3276, opened 2017-11-25
Weapon blast strength decreases with distance from blast

tehhowch:
- To mitigate the damage falloff for weapons with a trigger radius, a multiplier based on the ratio of trigger and blast radius is used.
 - For a weapon with 0 trigger radius, the damage will range from 100% at the center, to 25% at the edge of the blast radius.
 - "advanced" weaponry can define if it is not damage-scaled.

Refs #1630, #2529, #3275 

As of d400956, the denominator is chosen to vary with distance such that it does not immediately plummet, but rather gently plateaus before decreasing after ~40% of the blast radius:
image

This can be changed to a more aggressive profile (see discussion in #2529 ).

This change is best combined with an "Avoid" behavior in the AI that would seek away from danger (be it an exceptionally strong ship, or a potent explosive), but they are separate issues and should not be combined in a single PR. This PR would help balance some of the increased damage associated with #3275  (though only for Heavy Rocket and Nuclear Missile, as all other "missiles" are simply projectiles (just like the Beam Laser) that have no blast radius).

### DIFF
--- a/source/Mask.cpp
+++ b/source/Mask.cpp
@@ -199,7 +199,7 @@ namespace {
 	
 	
 	// Find the radius of the object.
-	double Radius(const vector<Point> &outline)
+	double ComputeRadius(const vector<Point> &outline)
 	{
 		double radius = 0.;
 		for(const Point &p : outline)
@@ -229,7 +229,7 @@ void Mask::Create(const ImageBuffer &image, int frame)
 	
 	Simplify(raw, &outline);
 	
-	radius = Radius(outline);
+	radius = ComputeRadius(outline);
 }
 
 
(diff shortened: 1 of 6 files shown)


tehhowch:
A partial "avoid" behavior is implemented in #3301 , for the ship that is firing the explosive weapon.
Better firing logic for explosive weapons is implemented in #3300

endless-sky:
Given how computationally costly mask range measurements are, it'd make more sense to just estimate the range based on the ship mask's "radius" and the distance between the projectile and the ship. (And, that might be more realistic anyway, because the damage done in real life would depend on how much of the ship is within the blast, not just on how close the closest part of the ship is.)

PR #3277, opened 2017-11-25
Make torpedoes worse at tracking fighters and fast ships

Pointedstick:
Torpedoes are now much slower, but with twice as much missile strength to compensate for the fact that they'll be receiving anti-missile fire for much longer. With this change, it becomes practically impossible to hit small fighters and interceptors with torpedo weapons, which is as it should be given their battle role and description. Large ships are still pretty easy to hit just because of their size, even if they're fast.

This should also have the side effect of making life easier for starting players--any pirate ship armed with torpedoes wont't be able to insta-kill the types of ships that the player starts with.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -796,14 +796,14 @@ outfit "Torpedo Launcher"
 		"hit effect" "large explosion"
 		"die effect" "tiny explosion"
 		"inaccuracy" 5
-		"velocity" 7
+		"velocity" 3.5
 		"lifetime" 600
 		"reload" 120
 		"firing energy" 2
 		"firing heat" 45
-		"acceleration" .7
+		"acceleration" .6
 		"drag" .1
-		"turn" 4
+		"turn" 1
 		"homing" 3
 		"optical tracking" .8
 		"shield damage" 450
@@ -858,14 +858,14 @@ outfit "Typhoon Launcher"
 		"hit effect" "large explosion"
 		"die effect" "tiny explosion"
 		"inaccuracy" 5
-		"velocity" 5
+		"velocity" 3
 		"lifetime" 600
 		"reload" 100
 		"firing energy" 4
 		"firing heat" 50
-		"acceleration" .5
+		"acceleration" .4
 		"drag" .1
-		"turn" 6
+		"turn" 1
 		"homing" 4
 		"optical tracking" .9
 		"shield damage" 610
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -809,7 +809,7 @@ outfit "Torpedo Launcher"
 		"shield damage" 450
 		"hull damage" 450
 		"hit force" 40
-		"missile strength" 20
+		"missile strength" 40
 	description "Torpedoes are the most powerful homing weapon, but they move slowly enough that fast ships can simply outrun them, making them most useful against large targets or against small ships at times when they are too distracted to dodge. Torpedo launchers are also much larger than other missile systems, so they are most often found in capital ships."
 
 effect "torpedo fire"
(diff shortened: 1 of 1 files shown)


Nescio0:
Doesn't this make torpedoes significantly less effective vs. large ships as well? If torpedoes are twice as slow, anti-missile turrets are effectively twice as efficient.

Pointedstick:
Oops, forgot to increase their missile strengths. I know I'm gonna make some mistakes splitting up the monster PR into pieces, so thanks for catching that. Will do that in a moment.

data/weapons.txt
```diff
@@ -858,14 +858,14 @@ outfit "Typhoon Launcher"
...
+		"velocity" 3
 		"lifetime" 600
 		"reload" 100
 		"firing energy" 4
 		"firing heat" 50
-		"acceleration" .5
+		"acceleration" .4
 		"drag" .1
-		"turn" 6
+		"turn" 1
```
tehhowch: Should still probably be better than the Torpedo, maybe 1.75 to 2.75 range.

Pointedstick: I dunno, the description says "However, they are also much slower, because they are intended mostly for use against large and slow-moving capital ships."

KiLEdEnNis: I'd say around 1.25-1.4 turn would be fine to make torpedoes at least decent against faster capital ships and slower regular ships

data/weapons.txt
```diff
@@ -796,20 +796,20 @@ outfit "Torpedo Launcher"
 		"hit effect" "large explosion"
 		"die effect" "tiny explosion"
 		"inaccuracy" 5
-		"velocity" 7
+		"velocity" 3.5
```
endless-sky: In general I prefer to have a projectile's starting velocity match its long-term velocity for the sake of the firing AI's range calculation. Is there a particular reason to make this lower?

data/weapons.txt
```diff
@@ -796,20 +796,20 @@ outfit "Torpedo Launcher"
...
 		"lifetime" 600
 		"reload" 120
 		"firing energy" 2
 		"firing heat" 45
-		"acceleration" .7
+		"acceleration" .6
```
endless-sky: I'm wondering if just the turn rate change would be enough here, without the velocity change. Torpedos are already pretty easy for any starting ship but the Star Barge to avoid.

endless-sky:
I've tested torpedoes in a star barge, and reducing their turn rate to 1 makes them incredibly easy to dodge, even without reducing their speed (because the low turn rate amplifies the effect of optical tracking having trouble with small ships). I think this needs a bit more testing to strike a balance between making things easier for small ships without nerfing torpedoes too much.

Pointedstick:
> reducing their turn rate to 1 makes them incredibly easy to dodge

Yeah, that's the point. Any ship capable of "dodging" should not be the target of something called a torpedo, and should not particularly fear getting hit by them, because they're anti-capital-ship weapons. Ideally, the damage would be commensurately increased, but I didn't want to overload this patch. If you're willing to entertain a damage increase, I can play around with that a bit.

data/weapons.txt
KiLEdEnNis: to back that up:
nerfing turn rates makes them less effective against fast/agile ships that can dodge them; intended
nerfing velocity makes outrunning them easier, and gives AM more time to shoot them down, which is more a nerf to the torpedoes in general

if a faster ship could dodge a torpedo but not outrun it it'd be perfectly fine to me

endless-sky:
With the turn rate of 1, 90% of torpedoes will miss a Star Barge even if I'm not trying to dodge. It makes it look like the torpedoes are just drifting aimlessly instead of tracking anything. I'd recommend a turn rate of 1.5 for normal torpedoes and 1.8 for typhoons. That still makes it possible to dodge, but you at least have to work at it.

endless-sky:
I've applied a slightly less aggressive nerf in 5bf534a. With balance changes, I prefer to under-do it than to make too big a shift and have to change back in the other direction later.

My change also increases how long it takes a torpedo to change direction, which is an alternative way to make it easier to evade them without making their tracking look completely broken.

PR #3278, opened 2017-11-25
Give Meteors and Sidewinders distinct battle roles

Pointedstick:
With this change, Meteor missiles are launched at a low speed but accelerate very quickly, have a shorter lifespan to compensate, and turn and track much worse (to match the in-game description). The net effect is to make them more useful against large ships--using high speed to bypass anti-missile defenses--but less useful against small fast ships. This should also have the side effect of making life easier for starting players, since most pirates in the Dirt Belt and galactic South are armed with Meteors which are now much worse against the types of ships that the player starts with.

Sidewinder missiles are now basically air-to-air missiles: they are slower than Meteors, but can maneuver to hit small fighters without 80% of them missing on the first pass and needing to turn around. This increases the importance and usefulness of radar jammers, which right now are pretty under-utilized.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -624,14 +624,14 @@ outfit "Meteor Missile Launcher"
 		"die effect" "missile death"
 		"hit effect" "medium explosion"
 		"inaccuracy" 10
-		"velocity" 9
-		"lifetime" 480
+		"velocity" 0.1
+		"lifetime" 150
 		"reload" 48
 		"firing energy" 1
 		"firing heat" 20
-		"acceleration" .9
+		"acceleration" 3
 		"drag" .1
-		"turn" 4
+		"turn" 2
 		"homing" 2
 		"infrared tracking" .7
 		"shield damage" 130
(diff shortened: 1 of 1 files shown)


endless-sky:
Given that missiles are one of the most common ways that new players end up unavoidably dying, I'm a bit hesitant to make them *more* effective against small ships.

Pointedstick:
Understandably so. Perhaps we should also equip all starting ships with Small Radar Jammers (and lower some prices) and make sure they run cool.

KiLEdEnNis:
making their tracking a bit bad & lowering their turn would allow for better maneuvering against them, but that'd make them bad again, kinda hard to make them effective without making them a guaranteed death for new players

endless-sky:
Won't that Meteor tweak make it so the AI never fires them, because it thinks their range is only 150 * .1 = 15 pixels?

Pointedstick:
Seems like a bug in the way the AI calculates range.

tehhowch:
Perhaps a rework of #1894 is needed. Performing a full integration of the motion equation over all submunitions is overkill, but certainly some combination of a creator-provided "optimal range" and fallback calculation from steady-state submunition velocities and initial velocities can provide range estimates suitable for AI usage.

PR #3279, opened 2017-11-25
Missile Overhaul part IV: Increase missile damage and durability, double reload times and price

Pointedstick:
With this change, Missile launchers fire half as fast, but their missiles cost double, inflict quadruple the damage, and are 50% stronger, to allow at least a few through light anti-missile defenses (which are now more dangerous with the missiles' reduced fire rates) Much more damage per missile impact feels both more realistic and more fun, and increases the utility of anti-missile preparations in player ship design.

Additional changes/exceptions:
- Heavy Rockets and Javelins now use the Stream mode, which makes them more vulnerable to anti-missile fire to compensate for their greatly increased effectiveness
- Nuclear missile damage values doubled, not quadrupled (that seemed a little excessive, and would let them one-shot all human ships). Even so, I halved their speed.
- Korath Mines received no changes. They seemed fine as-is.

### DIFF
--- a/data/coalition weapons.txt
+++ b/data/coalition weapons.txt
@@ -102,7 +102,7 @@ outfit "Finisher Pod"
 		"inaccuracy" 5
 		"velocity" 10
 		"lifetime" 60
-		"reload" 80
+		"reload" 160
 		"firing energy" 120
 		"firing heat" 20
 		"acceleration" .8
@@ -130,14 +130,14 @@ outfit "Active Finisher"
 		"radar tracking" .8
 		"optical tracking" .5
 		"missile strength" 50
-		"shield damage" 1500
-		"hull damage" 1900
-		"hit force" 240
+		"shield damage" 6000
+		"hull damage" 7600
+		"hit force" 960
 
 outfit "Finisher Torpedo"
 	plural "Finisher Torpedoes"
 	category "Ammunition"
-	cost 32000
+	cost 64000
 	thumbnail "outfit/finisher"
 	"mass" .5
 	"finisher capacity" -1
--- a/data/coalition weapons.txt
+++ b/data/coalition weapons.txt
@@ -110,7 +110,7 @@ outfit "Finisher Pod"
 		"turn" 0.1
 		"homing" 2
 		"radar tracking" .8
-		"missile strength" 50
+		"missile strength" 75
 		"submunition" "Active Finisher" 1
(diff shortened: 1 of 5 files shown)


tehhowch:
With blast radius weapons having varied damage, you may be able to keep the same increases for the Nuclear Missile and Heavy Rocket. See #3276

Nescio0:
This makes anti-missile turrets twice as effective; as a result missiles become more of a money drain than they already are.

Pointedstick:
> This makes anti-missile turrets twice as effective; as a result missiles become more of a money drain than they already are.

That's intentional, and balanced out by the fact that they're much more fun and useful compared to today (AM 2x as good + missiles do 4x as much damage = 2x effective missile DPS)

Nescio0:
In theory, perhaps. I fear that in practice ships without an anti-missile turret are almost instantly destroyed, and ships which have them become practically invulnerable. 
One basic anti-missile turret can already destroy meteor missiles six times as fast as they can be launched; with your proposal it can counter about a dozen launchers.

data/coalition weapons.txt
```diff
@@ -130,14 +130,14 @@ outfit "Active Finisher"
 		"radar tracking" .8
 		"optical tracking" .5
 		"missile strength" 50
-		"shield damage" 1500
-		"hull damage" 1900
-		"hit force" 240
+		"shield damage" 6000
+		"hull damage" 7600
+		"hit force" 960
```
KiLEdEnNis: this and the reload change makes Finishers twice as strong damage-wise, I really don't think they need this as they're very strong already
adding that you multiplied the missile strength by 1,5x while it's already very hard to shoot them down (due to their speed too), I think this would make Finishers overpowered

endless-sky:
This seems like way too large a shift to make at once. Something more like a 20-30% increase in reload times and a 30-50% increase in damage, applied only to homing missiles, would be more reasonable.

The real current limit on the usefulness of missiles is the bother of reloading them - in terms of damage per second they're already a pretty good deal.

tehhowch:
A problem with balancing to that limit (bothersome to reload) is that the AI ships have no such bother - a new fleet with a whole new salvo of those missiles is only a few minutes away.

Pointedstick:
>  in terms of damage per second they're already a pretty good deal.

Anti-missile turrets, jamming, and asteroids reduce the real-world DPS to vastly less than the on-paper DPS, the sole widely-available exceptions are human Torpedoes and Wanderer Thunderheads due to their greater durability and speed.

PR #3280, opened 2017-11-25
Add anti-missile defenses to most ships that lacked them

Pointedstick:
This change adds anti-missile turrets to most larger ships and variants that previously lacked them, and tries to cram Radar Jammers into smaller ships. Everything fits (at least the non-variant ships; I didn't check all the variants).

This is necessary if and when https://github.com/endless-sky/endless-sky/pull/3279 goes in. With that change, large ships that already have anti-missile defenses should not be that much more vulnerable to missiles on a DPS basis, but large ships without any anti-missile defenses get devastated. Anti-missile defenses become much more important for large ships.

Even if https://github.com/endless-sky/endless-sky/pull/3279 does not get merged, IMHO it still makes sense because missiles are only very dangerous to larger ships with zero anti-missile ability, so most ship designers and owners would want to build this in. And increasing anti-missile coverage for human ships makes changes to missile effectiveness more possible in the future since it wouldn't have such a disproportionate effect on all the currently-unprotected vessels.

### DIFF
--- a/data/variants.txt
+++ b/data/variants.txt
@@ -1412,7 +1412,7 @@ ship "Behemoth" "Behemoth (Hai)"
 	outfits
 		"Sidewinder Missile Launcher" 2
 		"Sidewinder Missile" 100
-		"Pulse Turret"
+		"Pulse Turret" 5
 		"Bullfrog Anti-Missile"
 		"Boulder Reactor"
 		"Hai Ravine Batteries"
--- a/data/variants.txt
+++ b/data/variants.txt
@@ -139,6 +139,7 @@ ship "Scout" "Scout (Speedy)"
 		"Outfits Expansion"
 		"A370 Atomic Thruster"
 		"A255 Atomic Steering"
+		"Small Radar Jammer"
 		"Hyperdrive"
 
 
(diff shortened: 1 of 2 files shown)


data/variants.txt
```diff
@@ -1362,7 +1412,8 @@ ship "Behemoth" "Behemoth (Hai)"
 	outfits
 		"Sidewinder Missile Launcher" 2
 		"Sidewinder Missile" 100
-		"Pulse Turret" 6
+		"Pulse Turret"
+		"Bullfrog Anti-Missile"
```
KiLEdEnNis: ya took out 6 turrets & placed only 2 back in

KiLEdEnNis:
apart from that 1 mistake I found & the fact that some ships like the Falcon & Protector will now be asymmetric, this PR is fine (didn't check for negative outfit space though)

data/variants.txt
Pointedstick: Oops.

endless-sky:
I think the game becomes less interesting if anti-missiles become so essential that a ship can't survive without them. Right now the game is aiming for the sweet spot where the added protection of an anti-missile system might or might not be worth it depending on whether you're willing to give up the damage that a turret can do, and I think it's struck that balance pretty well.

Pointedstick:
This was done in https://github.com/endless-sky/endless-sky/pull/3922

PR #3281, opened 2017-11-25
Missile Overhaul part VI: make starting shop options less vulnerable to missiles

Pointedstick:
Even without the proposed missile overhaul, the starting ships are often easy pickings for missile-armed pirates. This change gives them all at least _some_ form of anti-missile defense, in exchange for a 10,000 credit increase in purchase price. Also decreases the price of the radar jammers a bit, as

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -1147,7 +1147,7 @@ ship "Finch"
 	sprite "ship/finch"
 	attributes
 		category "Fighter"
-		"cost" 126000
+		"cost" 106000
 		"shields" 1100
 		"hull" 200
 		"required crew" 1
@@ -1169,6 +1169,7 @@ ship "Finch"
 		"nGVF-BB Fuel Cell"
 		"LP036a Battery Pack"
 		"D14-RN Shield Generator"
+		"Small Radar Jammer"
 		
 		"Chipmunk Plasma Thruster"
 		"Chipmunk Plasma Steering"
@@ -1737,7 +1738,7 @@ ship "Heavy Shuttle"
 	sprite "ship/heavy shuttle"
 	attributes
 		category "Transport"
-		"cost" 320000
+		"cost" 300000
 		"shields" 700
 		"hull" 700
 		"required crew" 1
@@ -1759,6 +1760,7 @@ ship "Heavy Shuttle"
 		"nGVF-AA Fuel Cell"
 		"LP036a Battery Pack"
 		"D14-RN Shield Generator"
+		"Small Radar Jammer"
 		
 		"X2700 Ion Thruster"
 		"X2200 Ion Steering"
@@ -2452,7 +2454,7 @@ ship "Shuttle"
 		"random start frame"
 	attributes
 		category "Transport"
-		"cost" 180000
+		"cost" 150000
 		"shields" 500
 		"hull" 600
 		"required crew" 1
@@ -2474,6 +2476,7 @@ ship "Shuttle"
 		"nGVF-AA Fuel Cell"
 		"LP036a Battery Pack"
 		"D14-RN Shield Generator"
+		"Small Radar Jammer"
 		
 		"X2700 Ion Thruster"
 		"X2200 Ion Steering"
@@ -2555,7 +2558,7 @@ ship "Sparrow"
 	sprite "ship/sparrow"
 	attributes
 		category "Interceptor"
-		"cost" 225000
+		"cost" 195000
 		"shields" 1200
 		"hull" 300
 		"required crew" 1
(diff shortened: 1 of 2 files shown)


Wrzlprnft:
I'd rather change the start money/mortgage than the prices

Pointedstick:
That would work too, but I'd leave that up to @endless-sky.

Nescio0:
Personally I'd also favour increasing the initial loan from 480,000 to 600,000 (then the story line messages also have to be updated), equipping the default Shuttle, Sparrow, and Star Barge with a small radar jammer (which costs an additional 70,000), and keeping the basic costs of all outfits and ships unchanged.
Note: there is still room for an anti-missile turret in the Star Barge, but I'd prefer not to install it by default. (New players might benefit more from two fuel pads).

KiLEdEnNis:
if we just add jammers to the starting ships & increase the mortgage, new players would spend MORE time paying off their debt, and some players already quit the game before paying off the current one

Pointedstick:
> PS Why do you change the configuration of the Finch (a fighter) and Heavy Shuttle in this proposal? Neither is a starting ship option.

The Heavy Shuttle is a clear upgrade path ship, so it seemed to quality.

The Finch was an opportunistic change since it fit and Finches face a lot of Republic Sidewinder missiles. Seems like Free Worlds techs would always fit radar jammers to offer their fighters more of a chance.

Nescio0:
> The Finch was an opportunistic change since it fit and Finches face a lot of Republic Sidewinder missiles.

“It fits” is not always a good reason. The Finch clearly isn't an option for new players, therefore it does not belong in this proposal.

> The Heavy Shuttle is a clear upgrade path ship, so it seemed to quality.

Realizing the Heavy Shuttle is indeed virtually identical to the Shuttle, equipping it with a jammer is acceptable. However, this also reduces their free outfit space from 8 to 4, meaning there is no longer space for a simple gun or a fuel pad. Therefore I'd recommend to improve their outfit size from 120 to 124 as well, to maintain a free space of 8 if they are equipped by default with a jammer.

Likewise the Sparrow needs an outfit increase, from 130 to 132, to allow installing a cargo scanner and an outfit scanner (see Sparrow (Patrol) variant).

The Star Barge does not need an outfit increase; installing a jammer by default leaves 16 free slots, enough for two fuel pads, an anti-missile turret, or a blaster turret.

Tadrix:
Radar jammers would be more attractive if they took up less outfit space, e. g. 3 or 2 instead of current 4. If player has a meager ship but tries to cram it full of useful things, outfit scanner and cargo scanner currently are more sound propositions (and take up less space than a jammer), since they add new mission opportunities, and jammer as such is ineffective enough that an encounter with a serious missile-spamming pirate will conclude with a game reload. _Some_ of them will get through, and that's enough for a starting ship.

Heavy shuttle indeed is an obvious upgrade at least for a player who chose transport as his starting ship. Also, its description says that it's an updated shuttle.

Finch essentially is a clipped Sparrow, so likewise, @Pointedstick 's reasoning makes perfect sense in this regard.

@Nescio0 Star Barge would need +4 os too. Some players like to buy cargo expansion or extra bunks instead of fuel pods. If make changes to starting ships at all, which currently, as I understand, are well balanced choices, then change all of them equally.

kikotheexile:
I dont see any reason any of this is needed. A simple text change adding built in jammers to starting ships (read: not an outfit) would work better for less work - assuming this is even a good idea, which frankly i am not sure if it is. Skill check jumping is already steep, and this would add a jump after the first ship - assuming people make it that far.

endless-sky:
Given that radar jammers only work on Sidewinders, and that nearly all pirates use Meteors, this wouldn't make starting ships any less vulnerable to pirates.

Pointedstick:
Would you accept changes to make the starting ships run cooler?

endless-sky:
The starting ships already do run pretty cool - idling at around 15% for a Shuttle or 20% for a Star Barge. But, it might make sense to boost their heat dissipation a bit, because it's weird that so many larger ships have better heat dissipation than the starting ships do.

data/ships.txt
```diff
@@ -2685,7 +2681,7 @@ ship "Star Barge"
 		
 	engine -9 38
 	engine 9 38
-	turret 0 -8 "Anti-Missile Turret"
+	turret 0 -8
```
KiLEdEnNis: no idea why ya did this, adding that it still has the AM turret installed, ya just removed it from the hardpoint

Pointedstick: Look at the diff. The change you're looking at is obsolete.

KiLEdEnNis: ah a lot of reverts, that explains it lol

Pointedstick:
Okey dokey.

PR #3282, opened 2017-11-25
Increased starting loan

Nescio0:
Increased starting loan from 480,000 to 600,000 to allow installing a small radar jammer in default Shuttle, Sparrow, and Star Barge. 
Edit: also updated the new player starting text to display new loan, daily payment, and total interest.


tehhowch:
I do not find it surprising that the file `intro missions` contains missions and not parameters that control the start of a pilot file.

Bladewood:
It's in `conversations.txt`.

KiLEdEnNis:
as I said in 3281

> if we just add jammers to the starting ships & increase the mortgage, new players would spend MORE time paying off their debt, and some players already quit the game before paying off the current one

jafdy:
I'm sure jammers don't cost 120,000 . . .

KiLEdEnNis:
that too
I don't think we should really touch the starting mortgage that much if at all, and if we do increase it, maybe it's worth lowering the interest a little bit so it's not straight-up harder to pay off your debt

Nescio0:
A Sparrow currently costs 448,000, a small radar jammer costs 70,000; add some additional funds for experimenting and fooling around, then 540,000 might work but 600,000 doesn't hurt, I think.
Lowering the interest rate would also be a good idea.

data/start.txt
```diff
@@ -17,6 +17,6 @@ start
 		score 400
 		mortgage Mortgage
 			principal 600000
-			interest 0.004
+			interest 0.003
```
KiLEdEnNis: do not forget to change the calculation results in the dialog as well

endless-sky:
I really don't think it's a good idea to increase the starting mortgage, especially by that much. Essentially you're increasing every player's starting debt in order to force them to purchase a radar jammer, and that just doesn't make sense.

ghoulavenger:
I agree that increasing the starting loan doesn't make much sense as a first solution to the problem being proposed. If missiles are the problem and radar jammers are necessary, they should just be part of the starting ships to begin with (which might require lowering the initial cost of the ships, the jammers, or both). Or more likely since the problem is pirates, tweaking the early pirate fleets that new players are likely to run into (outside of pirate infested systems naturally).

That being said, my only true argument against the proposed solution is that new players might struggle with even *more* debt without the basic understanding how of to make money in the game -- lowering the interest helps a little bit, but you're still talking about a larger debt.

Nescio0:
To clarify, in #3281 it was proposed to have starting ships equipped by default with a jammer. However, they would automatically become expensive as a consequence, and the starting loan would then be insufficient. Therefore I created this proposal.

By lowering the interest the larger debt becomes actually less of a burden. To summarize the difference:
```
Higher starting sum:              +120,000
Higher daily payments: -199*365 =  -72,635
Net benefit:                       +47,365
```

jafdy:
Jammers are only really useful for sidewinders. There aren't many pirates (if any) that use sidewinders so this won't help much.
More helpful would be making the starting ships run cooler so they are better for meteors.

Zireael07:
The proposal to have ships have jammers is due to the proposals to missiles...

KiLEdEnNis:
what exactly are ya trying to say? that the OP wants to add jammers to the starting ships to be more safe from missiles? if so; read the other comments on this post

endless-sky:
Given that this would only help against Sidewinders, and that forcing everyone buying a starting ship to pay for a specialized item (i.e. a radar jammer) instead of being able to just buy a bare-bones ship, this doesn't make sense to me.

PR #3283, opened 2017-11-25
Missile Overhaul part VII: Exchange launcher and storage capacities

Pointedstick:
This patch generally reverses the capacities for missile launchers and their storage outfits: now launchers are smaller and have a very limited internal capacity, and storage boxes/racks/tubes are much larger. This allows fighter-sized craft to be armed with missile weapons without giving them enormous missile-firing capacity that unbalances the game With this change:

- Military capital ships retain their high missile capacity
- Medium-sized ships with missile capacity lose quite a bit of their capacity
- Most fighter and interceptor craft can and do gain a very small amount of missile capacity

New launcher and storage sizes were balanced so that compared to the status quo, two launchers and one storage box/rack/tube are equal in total size, but result in fewer missiles. Regaining lost capacity requires buying more storage, which takes up more space now, and only larger ships will have that space available. No more Furies with 140 Meteor Missiles.

This change enables the creation and use of several new ship loadout options:
- Missile-armed fighters that quickly exhaust their missile supplies and then have to rely on guns
- Torpedo boats that exist entirely to fire torpedoes at capital ships (AI might need tweaking to make this actually work)
- Medium-sized ships that can mount multiple types of missile launchers for maximum flexibility
- Fighter-sized craft that can be armed with Heavy Rockets without turning into lone-capital-ship-destroying unholy terrors

Overall, this makes fighters feel more like real-world fighters, and capital ships like real-world capital ships. The differentiator in missile armament is not whether or not fighter-sized craft can carry missiles, but rather how many. Fighter-sized craft can now carry only a few missiles without sacrificing other armament--just like real-world fighters--while large capital ships maintain their huge missile stores--just like real-world missile cruisers and destroyers.

### DIFF
--- a/data/hai outfits.txt
+++ b/data/hai outfits.txt
@@ -191,9 +191,9 @@ outfit "Tracker Storage Pod"
 	category "Ammunition"
 	cost 28000
 	thumbnail "outfit/hai tracker storage"
-	"mass" 4.4
-	"outfit space" -10
-	"tracker capacity" 28
+	"mass" 8
+	"outfit space" -14
+	"tracker capacity" 30
 	ammo "Hai Tracker"
 	description "The Tracker Storage Pod is used to store extra ammunition for Tracker Pods."
 
@@ -201,11 +201,11 @@ outfit "Hai Tracker Pod"
 	category "Secondary Weapons"
 	cost 150000
 	thumbnail "outfit/hai tracker pod"
-	"mass" 8
-	"outfit space" -19
-	"weapon capacity" -19
+	"mass" 8.8
+	"outfit space" -10
+	"weapon capacity" -10
 	"gun ports" -1
-	"tracker capacity" 56
+	"tracker capacity" 6
 	weapon
 		sprite "projectile/tracker"
 			"frame rate" 10
--- a/data/hai ships.txt
+++ b/data/hai ships.txt
@@ -142,7 +142,7 @@ ship "Lightning Bug"
 ship "Lightning Bug" "Lightning Bug (Pulse)"
 	outfits
 		"Hai Tracker Pod"
-		"Hai Tracker" 56
+		"Hai Tracker" 6
 		"Pulse Turret" 2
 		"Geode Reactor"
 		"Hai Gorge Batteries"
@@ -241,7 +241,8 @@ ship "Shield Beetle"
 	outfits
 		"Ion Cannon" 4
 		"Hai Tracker Pod" 4
-		"Hai Tracker" 224
+		"Tracker Storage Pod" 2
+		"Hai Tracker" 84
 		"Chameleon Anti-Missile" 2
 		"Boulder Reactor"
 		"Geode Reactor"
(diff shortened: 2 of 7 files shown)


KiLEdEnNis:
I personally dislike missiles mainly because I can't store a large amount to make frequent restocking unnecessary, I don't think that increasing the outfit space of storages would be really smart

SirStigII:
I see the point of this pr, but I agree that unless restocking can be done  at any outfitter its going to make it even more annoying to be a missile boat outside your ammo supply area.

MessyMix:
I like the idea. 2 conditions will make this change viable:

1. Fighters can restock from motherships
2. Launchers can draw from ammo racks

No clue if either are implemented, but I think this change would work well if both conditions are satisfied.

KiLEdEnNis:
launchers can also use ammo from ammo storages
and I had an idea to make a bigger bulk ammo storage that the launchers can NOT access, so ya can hold a big amount of missiles to self-restock on planet, but to make this "realistic" all ammo would need lower mass, so say a 20 outfit 4 mass crate could store 80 .2-mass missiles for example
and at this point ya might as well just buy into your cargo hold

tehhowch:
Ammo racks are used to "install" the outfit, which means it can be fired normally.

Fighter reload is added in #1971 by @zwparchman and myself, (which does have some merge conflicts). Ideally #3182 merges here, then he resolves the merge conflicts, and then MZ merges #1971.

Tadrix:
The problem I see with this is size of missile storage should be highly customizable; it`s simply illogical to offer "100 missiles or nothing".

A working model could be something like this:
Missile _X_ launch rack: 8_y_ missiles, replenishable only by landing or docking with carrier, takes up 2_z_ space;
Missile _X_ launch system (a magazine with a launch rail): 10_y_ missiles, takes up 3_z_ space, extendable:
Missile _X_ launch system's additional magazine capacity: +5_y_ missiles, takes up _z_ space.

Smaller ships might favor launch racks, equipping one or two of them. Anything larger would have full-fledged launch systems installed.

tehhowch:
Cargo-based reloading would make it very difficult for player and AI alike to reload on the carrier, as for AI cargo is picked generally randomly (and does not include outfits), and the player's cargo is loaded in decreasing size order, so the ammo will generally be loaded in different ships every time you take off. In #1971 carrier-based reloading uses ammo loaded as outfits (which also enables one-click full restocking when you visit there appropriate Outfitter).

Tadrix:
Understood. But my other points still stand. In my opinion, though, reloading using missiles which are brought along shouldn't need an outfitter. These guys are just taking their missiles to the racks and putting them in, not buying or rebuilding anything.

tehhowch:
I'm going to handwave that moving missiles from safely packed cargo crates into ready-to-fire ammo racks requires specialized tools not available at your standard spaceport refueling facility.

Given that the player can refuel their launchers by plundering a boarded ship, you can probably drum up enough support for your cause though.

endless-sky:
This would essentially force anyone buying a launcher to also buy a storage outfit, which I consider a non-starter. The fact that the ammo, storage, and launcher are all separate items is already confusing enough for starting players.

Bladewood:
It could be a fun characteristic of a plugin species... :)

JannHL:
I personally love this idea, especially cause it goes hand in hand with @3372, giving em 2-4 ports for some missiles would make em more balanced, also making missile boats no longer be a way to destroy nearly anything :D

Issue #2147, opened 2017-02-16
Jumping with fleet ships with firing force weapons


Bladewood:
Was playing around with Heliarch ships: a stock Punisher, Interdictor, and a Neutralizer.
We'll focus on the first and second ones.

The Punisher was my flagship. The Interdictor was a fleet ship. I was playing in Shaula.
I noticed that every time I would attempt to order my fleet to prepare to jump (by holding [J]), fleet ships would fire their turrets at nearby hostiles.
Normally, that would be fine.
But the Interdictor's turrets had firing force.
And so every time I tried to prepare to jump, it would just push itself around by trying to fire on ships.

This probably happens to ships with particle cannons and proton guns and the like, too.

KiLEdEnNis:
I don't remember my escorts ever firing when I hold J

jafdy:
@KiLEdEnNis They do fire. You will notice it quite quickly unless you never jump with hostiles around.

KiLEdEnNis:
I do sometimes, I'll check today if I don't forget like usual

KiLEdEnNis:
just saw they did, seems new to me

PR #3286, opened 2017-11-25
AI does not autofire weapons with firing force if preparing to jump

tehhowch:
Refs #2147, fc785a18f37f3a1ac6a39bdd84d916d67268c4ba

 - Switched Outfit * to Weapon * and updated variable names
 - Updated the includes to reflect the used classes.

With this, while ships are being told to prepare to jump, they will not fire weaponry which imparts a firing force on their (e.g. could cause them to no longer be jump-ready). This behavior properly propagates to mission NPCs that would jump with you.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -12,12 +12,12 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include "AI.h"
 
-#include "Armament.h"
 #include "Audio.h"
 #include "Command.h"
 #include "DistanceMap.h"
 #include "Flotsam.h"
 #include "Government.h"
+#include "Hardpoint.h"
 #include "Mask.h"
 #include "Messages.h"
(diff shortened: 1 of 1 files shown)


jafdy:
If their own engines are enough to counter the force of forward guns then they would be quite safe to fire, simply by using the engines to hold station.

tehhowch:
Autofire is called for both turrets and forward guns. I could add a special case to handle checking the firing angle but it would be a much larger check given that it would have to handle forward and reverse thrust, and positive and negative firing forces. Additionally, only forward guns mounted at 0 X add a force directly in line with the ship's thrust - the others converge, and thus we'd half to check if the offsetting hardpoint exists, and is also set to fire, in order to counteract the X component of the firing force.

In short, it's a lot more in depth of a check compared to "will it move me when I don't want to be moved?"

PR #3289, opened 2017-11-25
mini change: fixed(?/improved) some grammar

KiLEdEnNis:
I've been told that the current form is fine, but my change is "more correct"

the change:
prying the lids off them --> prying the lids off of them
EDIT, it became:
prying their lids off

### DIFF
--- a/data/free worlds intro.txt
+++ b/data/free worlds intro.txt
@@ -654,7 +654,7 @@ mission "FW Katya 1C"
 				goto search
 			
 			label search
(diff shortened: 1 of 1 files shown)


Pointedstick:
"prying their lids off" would  be equally correct and sound better than the proposed change, thereby bypassing a potential argument entirely.

KiLEdEnNis:
you're right, thanks

Bladewood:
I personally think the original and the original proposal sound better.

endless-sky:
My general policy is, if it's not actually a grammar error and is just a matter of opinion, it's not worth tweaking it. There are probably just as many people who would want it switched back to the original.

PR #3294, opened 2017-11-26
Increase missile hit force to make them feel more powerful

Pointedstick:
This multiplies missiles' `hit force` by 15 to make them seem like they're actually doing something on impact. That seems like a very high number, but it's really not--the net effect is simply to make missiles visibly impact small light ships and push them back a little bit. It's a small but dramatic change and makes missiles _feel_ more dangerous without actually changing any of their actual danger-related game mechanics.

This change combines really well with https://github.com/endless-sky/endless-sky/pull/3274

### DIFF
--- a/data/coalition weapons.txt
+++ b/data/coalition weapons.txt
@@ -132,7 +132,7 @@ outfit "Active Finisher"
 		"missile strength" 50
 		"shield damage" 1500
 		"hull damage" 1900
-		"hit force" 240
+		"hit force" 3600
 
 outfit "Finisher Torpedo"
 	plural "Finisher Torpedoes"
--- a/data/hai outfits.txt
+++ b/data/hai outfits.txt
@@ -231,7 +231,7 @@ outfit "Hai Tracker Pod"
 		"infrared tracking" .4
 		"shield damage" 200
 		"hull damage" 160
-		"hit force" 50
+		"hit force" 750
 		"missile strength" 16
 	description "Trackers are fast, powerful, and accurate homing weapons. Their only weakness is their large turning radius: if a Tracker misses its target, it takes a long time to turn around."
 
--- a/data/korath weapons.txt
+++ b/data/korath weapons.txt
@@ -277,7 +277,7 @@ outfit "Korath Piercer Launcher"
 		"shield damage" 290
 		"hull damage" 440
 		"piercing" .2
-		"hit force" 30
+		"hit force" 450
 		"missile strength" 73
 		"stream"
 	description "Korath Piercer missiles carry a pair of warheads. The first, a smaller one in the very tip of the missile, explodes on impact to blast a hole in the ship's shields to allow some of the subsequent, larger explosion to pierce through."
@@ -449,7 +449,7 @@ outfit "Korath Mine Submunition"
 		"drag" .05
 		"shield damage" 350
 		"hull damage" 280
-		"hit force" 50
+		"hit force" 750
 		"missile strength" 22
 
 effect "minelayer fire"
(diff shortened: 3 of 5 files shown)


MessyMix:
Awesome. Awesome awesome awesome. A lot of times we neglect how it visually looks and feels to our senses in favor of balancing the stats—which, while the most important, doesn't necessarily correlate to how dangerous a weapon feels. Good change.

Lineth:
This + better, 'boomier' sounds + less spammy weak missiles in favor of fewer, nastier ones pls.

endless-sky:
That would make it a whole lot harder to jump if you're being hit by missiles. Right now, a medium-sized ship can take a few missile hits without accelerating it enough that it has to slow down all over again.

Pointedstick:
True--but only a bit. Like I said, the number seems high but in practice the impact isn't that much if you're in a Firebird-sized ship or larger, unless you're getting hit with enough missiles that you're gonna be dead before you even stop to make the jump. Also, more of a reason to use a Scram Drive!

KiLEdEnNis:
it would make Scram Drives a bit more interesting to use; do you get a cheap dangerous 100-fuel Hyperdrive or get a Scram Drive, which is safer at the cost of some fuel?

endless-sky:
I wonder if there would be an alternative option, like shaking the entire view or the heads-up display when you get hit by a powerful weapon. (That's been on my list of things to try for a while - I'm not sure if it'd look cool or just tacky.)

tehhowch:
I'm not sure the view should shake for something as common as missile hits. Some crazy one-of-a-kind weapon, sure... but missiles I don't think qualify.

endless-sky:
Another possibility would be to have a new "hit acceleration" weapon attribute, which would be like hit force except that it has the same effect on all ships regardless of their size. Otherwise I worry that increasing hit force this much will end up with tiny ships getting flung away at high speed every time they're hit. (That could also fix the problem of fighters getting flung away by huge ship explosions.)

Pointedstick:
If you merge some of my other patches, tiny ships won't be getting hit with torpedoes anymore. :slightly_smiling_face: 

Try it out: Meteors and Sidewinders don't fling small ships away at top speed with this patch. At most they bounce them a bit--and only for really tiny ships.

tehhowch:
"fighter fling" is alleviated immensely once the hit force is scaled with distance #3276

tehhowch:
After playing a bit as a new baby pilot (on master) with a berserker, Sparrow, hawk, and clipper, I have to agree that the missile hit force needs boosting.
While the increase will make it harder to jump, it also makes it easier to *survive*, because you are - in general - being pushed away from your attacker. Conversely, using these weapons also helps you survive, as you are pushing opponents away to help deal with them in reduced numbers.

Pointedstick:
> edit: no increase for Nuclear Missiles?

Their hit force was already pretty insane, but if you think it needs a boost, especially with https://github.com/endless-sky/endless-sky/pull/3276, then I can boost it.

tehhowch:
I would consider increasing these even more, perhaps in proportion to the danger a single hit poses, and then even more if it has a blast radius. Played with the Heavy Rockets a bit and they are able to push loaded Clippers around :)

Pointedstick:
I wouldn't object to that! 🙂

tehhowch:
They're OK with this PR as-is

Pointedstick:
Either way, I believe this is a significant improvement over what we have right now. I'll leave it be, and if it gets merged, we can investigate later whether we need to tweak the values some more.

endless-sky:
I've applied a similar set of changes in commit e61c25f, but with less extreme increases for the weapons that already had a lot of hit force and larger increases for the ones that had almost no hit force (like javelins). I also tweaked the hit force equation so that the acceleration is about half what it used to be for tiny drones and about 3 times what it used to be for giant ships, to eliminate some of the extremes in hit force responses.

PR #3295, opened 2017-11-26
Listed Flamethrower under Guns

Nescio0:
The Flamethrower does not require any ammunition (and fires continuously), therefore it would be more suitable to list under Guns instead of under Secondary Weapons. This would also make it easier to compare it with lasers, plasma cannons, Korath fire lances, etc.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -524,7 +524,7 @@ effect "plasma explosion"
 
 
 outfit "Flamethrower"
-	category "Secondary Weapons"
+	category "Guns"
 	cost 190000
 	thumbnail "outfit/flamethrower"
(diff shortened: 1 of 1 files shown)


warp-core:
But it uses hyperspace fuel.
If it's listed as a gun, then the player could quickly use up all their fuel, especially with autofire on.
While, as a secondary weapon, the player can control when it should be fired and when not to fire it.

Nescio0:
> But it uses hyperspace fuel.

Does it? Thanks for pointing this out! Then it would make sense to leave it as a Secondary Weapon.

Amazinite:
>Does it?

Please play the game you're making changes to. :)

Issue #3302, opened 2017-11-27
Scanner ranges


tehhowch:
Should scanners measure from the center of the player's flagship, or from the exterior?

This applies to:
 - Tactical Scans
 - Asteroid Scanning
 - Cargo Scans
 - Outfit Scans
 - future scans

Currently, the distance is measured from center to center, for both ships, while the effective distance (mask to mask) can be considerably shorter or even overlapping.
For comparison, weapons (which also have an associated range check) know where on the ship they are located, and check if they will hit the outline of the targeted ship, and not just the center.

Proposal: alter the scanning range checks from `center-to-center` to `mask-to-mask`, or `mask-to-center`, range checks. I think `mask-to-center` is the most sensible, as one probably needs to scan more than just the edge of a ship to learn what is in its cargo hold, etc.

Thoughts?

KiLEdEnNis:
when reading this, before I read your proposal I thought about mask-to-center for the exact same reason as you stated, and this is about as realistic as we can make it, so I agree

Amazinite:
So instead of the scan range being in a circle around your ship, it's in a ship-shaped mass around your ship? Sounds like that would be fun to look at if our scan range was visible.

Honestly, I think that center-to-center is fine. The scanner is only on one part of your ship, not on every part. Sure, that part might not be the exact center (and I'm not proposing that we make scanner slots for ships, or cause scanners to take up turret space), but the center is still a single point where the scan is emanating from, instead of the scanner somehow being omnipresent across the entire hull of your ship.

gunqqer:
Now if we want to get complicated, it could be from your center to greater than like 40-50% of the target ship so not the entire ship has to be in range, but some of it.

I feel like that'd be too complicated

tehhowch:
>So instead of the scan range being in a circle around your ship, it's in a ship-shaped mass around your ship?
Yep, basically anything within `X` of your ship's outline, rather than within `X` of the center of your ship. An overlay indicating scan range could be neat, but as you mention would probably necessitate defining a scanner point.

> if we want to get complicated, it could be from your center to greater than like 40-50% of the target ship so not the entire ship has to be in range, but some of it.
>
>I feel like that'd be too complicated

And I agree. For the most part, measuring to the target's center will mean that 40-50% of it is covered by scannable area.

jafdy:
[...]

PR #3303, opened 2017-11-27
Measure scanner distances from the ship's outline rather than its center

tehhowch:
Refs #3302 
 `Mask-to-center` distance:
 - Better agreement between displayed range and known weapon ranges than if measuring distances `center-to-center`.
 - Range to target is 0 if overlapping the target's center
 - Target "knowledge" requires measuring more than just its edge (i.e., `mask-to-center` range is more logical than `mask-to-mask` range.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2731,7 +2731,7 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 		{
 			for(const shared_ptr<Minable> &asteroid : minables)
 			{
-				double range = ship.Position().Distance(asteroid->Position());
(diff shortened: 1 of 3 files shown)


endless-sky:
It really doesn't make sense to me for a scanner to behave as if it is somehow located in every single part of the ship. And mask to point distance measurement is *much* more computationally intensive than point-to-point. I just don't think this change would provide enough benefit to justify the different behavior.

tehhowch:
Fair enough. What's your opinion on using a double-radius compensation? The mask radius is computed only once so it should be much faster than the Range method used here, but as I mention in #3302 it won't be quite right for skinny ships. I do think it will be "more" right than the current center-to-center range though.

PR #3304, opened 2017-11-27
Draw a pointer to the targeted asteroid

tehhowch:
~~Once an asteroid is no longer in range of the flagship's asteroid scanners, it should no longer be targeted (as it is no longer drawn with the targeting reticles).~~

While in range, a pointer is shown in the direction of the asteroid, as is done for ships.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2695,11 +2695,6 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 	}
 	wasHyperspacing = ship.IsEnteringHyperspace();
 	
-	// If the player's targeted asteroid leaves the scannable range, it is no longer targeted.
-	if(ship.GetTargetAsteroid() && ((ship.GetTargetAsteroid()->Position() - ship.Position()).Length()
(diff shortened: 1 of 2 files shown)


endless-sky:
The idea behind preserving the target was that you need to be in range for your sensors to detect that it is a minable asteroid, but once you've done that your ship can track it even if it's farther away. It could be really annoying if the asteroid that a player is trying to mine keeps getting un-selected just because their scanners have a short range (e.g. they're using a surveillance pod to scan them).

tehhowch:
Surveillance pods don't have asteroid scanning ability yet - are you planning to add that?

KiLEdEnNis:
it wouldn't make that much sense to add that, but it could be nice to make the Surveillance Pod some sort of "omniscanner"

tehhowch:
Reverted the removal of the targeted asteroid. The sole change in this PR is to now add a pointer to the targeted asteroid:
image

endless-sky:
I think it makes more sense to refactor the target pointer code (f8c1bf0) so that the minimum range check doesn't need to be done in two places.

tehhowch:
Yep that's better

PR #3306, opened 2017-11-27
New Energy Blaster and Modified Blaster sounds

Pointedstick:
In the spirit of improving under-appreciated weapons without actually changing their damage characteristics...

The Blasters have never sounded adequately weapon-like to me. This PR replaces the Energy Blaster and Modified Blaster sounds with ones that sound more like sci-fi space guns to my ears. More weapon-like--more _dangerous-sounding_ (especially the Modified Blaster IMHO). Here are some videos showing how they sound with the PR:

Blaster: https://webmshare.com/0BB9d

Modified Blaster: https://webmshare.com/dvvm7

These sounds were produced from the source file at https://freesound.org/people/jradcoolness/sounds/334224/, which is in the Public Domain.


tehhowch:
>sound now

Is that with this PR, or how they actually sound *right now*?

Pointedstick:
Hah, you (like most players, probably) use blasters so infrequently that you don't even know what they sound like! :rofl: The videos show the proposed new sounds.

KiLEdEnNis:
I'll have you know I cheated myself a Blaster Turret Archon

endless-sky:
Your new sound is about 0.7 seconds long, compared to the existing sound which is 0.3 seconds long. On some systems, OpenAL's limit for the number of concurrent audio sources may be much lower than others. That causes trouble with weapons like the blaster that can have a high rate of fire.

You could probably clip these sounds to the same length as the old ones - just cut off the end and then "fade out" so it fades to total silence at the point where you cut it off.

Issue #3264, opened 2017-11-24
Few Human ships have enough gun or turret mounts to take advantage of Energy Blasters' small size


Pointedstick:
I've [brought](https://github.com/endless-sky/endless-sky/issues/194) [this](https://github.com/endless-sky/endless-sky/issues/316) [up](https://github.com/endless-sky/endless-sky/issues/668) [before](https://github.com/endless-sky/endless-sky/issues/1533), but I wanted to take the opportunity to do so once more in an integrated manner, reflecting years of gameplay and observation of player trends.

## The issue ###
Energy blasters are reasonably efficient on a damage-per-ton basis, but most human ships don't have enough gun or turret mounts to actually take advantage of this, so their effective maximum DPS is capped. If you only have four gun or turret  mounts, then it makes more sense to fill them with better weapons because you don't get the option to achieve a similar or higher DPS by instead using a larger number of less efficient worse ones weapons.

## The proposal ##
Without increasing any ships' weapon or outfit capacities at all or changing their default loadouts, raise the number of gun ports and turret mounts on human ships across the board--and especially on fighter-sized craft. I mean, the Sparrow already _looks_ like it has four gun ports. Give the sparrow four gun ports, the Berserker 6, the Hawk and Fury 8, and give every turret-armed ship another turret or two (you can imagine that they're on the underside of the ship, so no sprite changes would be necessary).

## Why it would be a gameplay improvement ##
This would improve the humble Energy Blaster's competitiveness during Tier 1 play without actually needing to change the weapon at all. Instead, you'd have the option to fit more of them to achieve higher DPS compared to using Lasers. A reasonable player might actually decide to favor Energy Blasters over Laser weapons, whereas at present time nobody would make that decision; Heavy Lasers are just a straight-up better option if your gun and turret mounts are limited to 3 or 4 each (which is almost always).

This would also make gun-armed fighter craft more useful at low numbers (we all know that a swarm of 40 status-quo fighters with Gatling Guns or Pulse Cannons is deadly, but that's not actually seen in most play). This would also allow fighter-sized craft to feasibly be armed with missiles and also an effective number of guns which would vastly improve their missile-firing behavior: instead of annoyingly running away while spamming missiles, they would fire their missiles while they close to use their guns, which is more more fun to observe and fight against, and more reflective of how contemporary military aircraft operate.

This would not require any default ship loadout changes--and therefore no balance testing, either. You'd just have another practical option for how to arm your Tier 1 Human ships.

[...]

PR #3307, opened 2017-11-27
Faster blaster bolts

Pointedstick:
Something that makes Blaster weapons feel unimpressive is the slow speed at which their projectiles travel. They feel more like blaster _blobs_ than blaster _bolts_.

This change increases blaster bolt speed by 50%, keeping the overall range the same. Now the bolts look much more impressive and dangerous (especially when combined with https://github.com/endless-sky/endless-sky/pull/3306). This also helps Blaster turrets get a few more hits in against small fast-moving targets, which should improve their viability compared to Laser turrets.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -130,8 +130,8 @@ outfit "Modified Blaster Turret"
 		"hit effect" "blaster impact"
 		"inaccuracy" 4
 		"turret turn" 4
-		"velocity" 15
-		"lifetime" 30
+		"velocity" 11.25
+		"lifetime" 40
 		"reload" 6
 		"firing energy" 16
 		"firing heat" 55
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -21,8 +21,8 @@ outfit "Energy Blaster"
 		sound "blaster"
 		"hit effect" "blaster impact"
 		"inaccuracy" 3
-		"velocity" 7
-		"lifetime" 60
+		"velocity" 10.5
+		"lifetime" 40
 		"reload" 12
 		"firing energy" 10
 		"firing heat" 35
@@ -47,8 +47,8 @@ outfit "Blaster Turret"
 		"hit effect" "blaster impact"
 		"inaccuracy" 3
 		"turret turn" 5
-		"velocity" 7
-		"lifetime" 60
+		"velocity" 10.5
+		"lifetime" 40
 		"reload" 6
 		"firing energy" 10
 		"firing heat" 35
@@ -73,8 +73,8 @@ outfit "Quad Blaster Turret"
 		"hit effect" "blaster impact"
 		"inaccuracy" 3
 		"turret turn" 3
-		"velocity" 7
-		"lifetime" 60
+		"velocity" 10.5
+		"lifetime" 40
 		"reload" 3
 		"firing energy" 10
 		"firing heat" 35
@@ -104,8 +104,8 @@ outfit "Modified Blaster"
 		sound "mod blaster"
 		"hit effect" "blaster impact"
 		"inaccuracy" 4
-		"velocity" 7.5
-		"lifetime" 60
+		"velocity" 11.25
+		"lifetime" 40
 		"reload" 12
 		"firing energy" 16
 		"firing heat" 55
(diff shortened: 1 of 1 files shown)


data/weapons.txt
```diff
@@ -130,8 +130,8 @@ outfit "Modified Blaster Turret"
 		"hit effect" "blaster impact"
 		"inaccuracy" 4
 		"turret turn" 4
-		"velocity" 7.5
-		"lifetime" 60
+		"velocity" 15
+		"lifetime" 30
```
Amazinite: The modified blaster turret has a higher velocity than the modified blaster gun?

endless-sky:
Blaster bolts are *supposed* to be wimpy-looking. Its the smallest, cheapest weapon in the game.

Pointedstick:
I can't say I understand why you want the Blaster to both be and seem so awful. It's like you're telling players not to use it, that it's almost a joke weapon. In which case... why bother including it, and using as the default armament for more than half the human ships--including many Republic warships?

KiLEdEnNis:
I think making it a little bit faster is fine either way though, it won't become that much better but at least using it is a little less bad

jafdy:
While the basic blaster can look that way it seems a shame to resign an entire class of weaponry to uselessness because of this.

Amazinite:
>Blaster bolts are supposed to be wimpy-looking. Its the smallest, cheapest weapon in the game.

I can agree that Blasters should look wimpy and that their bolts should be slow, but I honestly still find the firing heat and energy way too taxing to even justify using blasters over lasers even early game, because the increased firing heat and energy of blasters practically makes you spend more money on other outfits in order to support it that it does to just buy a laser.

endless-sky:
A beam laser is 60% larger than a blaster, only does 33% more shield damage, and has a 30% shorter range. A heavy laser is 200% larger, does 140% more damage, and still has a slightly shorter range. I could certainly see an argument for reducing the heat and energy requirements of the blaster, but making it more powerful doesn't make sense - a 5 T weapon *ought* to be significantly wimpier than an 8 or 15 T one.

(Speaking of which, I also think it would be reasonable to buff the beam laser a bit. Or alternatively, to make the heavy laser and electron beam a bit less powerful.)

Pointedstick:
I don't object to the Blaster's low damage. Here are the ways I think we could make them more attractive:
- Heat and energy could stand to be reduced
- Ships could have a few more gun and turret mounts to fully take advantage of them (https://github.com/endless-sky/endless-sky/issues/3264)
- They could be look and sound more satisfying (this PR and https://github.com/endless-sky/endless-sky/pull/3306)
- Offer a Modified Quad Blaster turret (https://github.com/endless-sky/endless-sky/pull/3319)

Amazinite:
>(Speaking of which, I also think it would be reasonable to buff the beam laser a bit. Or alternatively, to make the heavy laser and electron beam a bit less powerful.)

In what way would they be changed? While I find the damage of the lasers a little high, where they really stand out is their high efficiency of firing heat and energy. Compared to the damage they produce, the human lasers produce among the lowest heat and require among the least energy, sitting up there with the Moonbeam, Bombardment Cannon, and Thrasher Cannon.

By that manner, I'd actually make all three human lasers less powerful by increasing their firing energies and heats by 45%. That might seem like a lot, but currently that only puts the Electron Beam at having just as much firing heat + energy as the modified blaster (except with more of that drain taken up by energy than heat compared to a modified blaster).

If that's too much for many ships that currently use those weapons, then a 20% increase in firing heat and energy but a 10% decrease in damage across the board looks like a viable change to me as well (and might actually be a better change).

Pointedstick:
I like the 45% idea. That would make heat management a much bigger part of the game for many ships and players, which seems appropriate.

KiLEdEnNis:
the problem with that would be having to modify all those variants, and some of them may not even "work" anymore
and if the Navy ships get affected too much we may even need to rebalance all the FW stories

Amazinite:
I've implemented the blaster change in #3350, but a change to the laser weapons would take more time to ensure that everything works smoothly.

Lineth:
I have to disagree with the blaster being just straight-up worse than most weapons. Have it do low damage, by all means, but I think it should have some redeeming quality to make it attractive. I see it's niche being a compact, efficient but low-damage weapon. As it is they seem to just be a waste of time. I did maker a personal mod of the mod quad blaster turret with stats based on the current mod blaster lione and it seems actually competitive with the Hai Pulse turret. This looks like it's a poor man's Plasma Turret, which IMO is not bad. Maybe a another smuggler's bay/custom pirate candidate?

endless-sky:
In beb7f03 I've increased blaster bolt speeds by about 20%.

Pointedstick:
Awesome!

Issue #1828, opened 2016-11-20
Pug systems can become friendly before completion of storyline


atyates:
Made a new account, wanted to capture as many pug ships as possible so I never finalized the end of Freeworlds campaign - for a long time I simply captured pug ships.  Then I figured I would try to conquer the Pug homeworld in Wanderer space, so I proceeded to start harvesting pirates and extremists - stopping infrequently to capture autonomous Korath ships.

Push comes to shove, I'm trying to capture some more Korath ships today (probably like 200 hours on this play through) and I notice that the Pug planets are friendly - but I still haven't finished the Pug storyline - something I don't believe should be possible.

KiLEdEnNis:
they have a negative attitude towards pirates, so that explains it
but I think there should be some option that permanently makes a government have a negative attitude towards the player until a mission changes that (to avoid this)

Amazinite:
Well the Pug start out friendly, become hostile during the campaign, then return to being friendly. Maybe during the length of the Pug part of the campaign they shouldn't have an attitude toward governments not involved in fighting.

KiLEdEnNis:
if you are at war with them, they should get a "tag" that makes them always hostile, until the war ends, & then the tag gets removed

Bladewood:
I believe that the Pug are neutral towards anyone that is not actively trying to kill them, or is not part of their plans in the war.

A repeating mission that triggers every time your reputation with a government falls below (or above) a certain threshold, changing your reputation with the government to a specified value, would work miracles.

Amazinite:
@KiLEdEnNis #1322

KiLEdEnNis:
@Amazinite yeah, that one (thanks)
& what @Bladewood said; having an event trigger when reputation = 0 or higher that sets government back to -1000 could be useful (but you can't "stop" it, so the government can't be friendly)

Bladewood:
I dunno... we could have a faction where you are capped out at 1 or something, and you instantly become their enemy if you slay one of their ships, no matter how much you've helped them in the past.

KiLEdEnNis:
basically, reputation borders
[government]
   reputation x
      boundaries -1000 < x < -1

so reputation will always be between -1000 & -1 (always hostile)
could also open some possibilities for designers

tehhowch:
[...]

Issue #264, opened 2015-11-26
[Spoil] bug? "hostility of the republic"


Oeleo:
During the Free world mission's "The Free World's fleet has been scattered by a Navy Invasion" and I had to go in the Tarazed system. I didn't go here, I went to Kor Nor'peli to farm. My combat rating went from 11 to 12 and pouf, all the Republic ships and planets become friendly (to me) but Free World's and Republic's ships continued to fight together. Sorry for the long story but I don't really know what did that.

endless-sky:
Your "reputation" with any government improves if you fight enemies of that government - e.g. pirates. So if you stop fighting for the Free Worlds and turn into a pirate hunter, sooner or later the Republic will decide you are no longer an enemy.

Oeleo:
Oh ok, that's really nice but I take tribute from half of the republic's planet, so maybe that it will be more coherent if the limit of the hostility toward the player could be bigger (is it understable ?). Like that he really need to do an intentional "redemption" to change their attitude toward him.

gchapyzh:
> Oh ok, that's really nice but I take tribute from half of the republic's planet.
Brother in arms! We are the same!

I do think that the factions forgive you a bit too easily. When I destroyed their gunboat and later a sensor drone, they all turned hostile, attacked me, but after I cleared the system of all ships and jumped, they became friendly again. AFTER I BLEW UP MULTIPLE CARRIERS THAT FIRED ON ME.

Perhaps have a special mission that will allow you to neutralize relationships with a faction.

PR #3311, opened 2017-11-28
Keep the Pug hostile if at war with them

tehhowch:
Refs #1828

### DIFF
--- a/data/events.txt
+++ b/data/events.txt
@@ -1771,13 +1771,13 @@ event "pug invasion 4"
 
 event "fw syndicate welcoming"
 	"reputation: Syndicate" = 1
-	government Syndicate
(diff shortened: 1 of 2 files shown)


Bladewood:
Would you not do the same for the Republic in the FW war?

tehhowch:
https://github.com/endless-sky/endless-sky/issues/264#issuecomment-159800618

The Pug are special change agents who shouldn't be friendly when they are at war with *humanity*

endless-sky:
I think it's much simpler to just make it so that your behavior toward other factions *never* influences the Pug's opinion of you (c3b9e15). Otherwise, this change would have to be made in both branches, not just the checkmate one.

Issue #558, opened 2015-12-30
Translations


ph2000:
It would be interesting to have different languages, maybe for players that don't speak English.
I would be glad to translate the missions and dialogues, but I don't know the game engine code enough to support language selection or translating text replacements (like <passengers>)

endless-sky:
I'm hoping to add full Unicode support eventually - right now only ASCII text is supported, which rules out most languages other than English. But given how text-heavy the game is - already about half the size of a novel - translating the entire thing would be a major undertaking.

Nightmareman:
Don't worry I am sure once the full game is out there will be people who will be happy to translate it even text heavy Japanese games get fan made translations if the game is popular enough. Translating from English to languages such as Spanish, Italian, Russian ect. is much easier then Japanese to English so once the game is finished translators will probably pop up for most languages.

ph2000:
to refer to me please use @ph2000bis

omagasohe:
If anyone wanted to translate the game, the data files would be the start. I'd imagine completing that effort would give motivation for adding support for unicode and localization of the game code. Generally though adding localization support after as much effort that has been give without an eye to it is painful. Every static bit of text that is display in the interface has to be changed into a variable and then a method of storing and selecting different languages has to be implemented. I don't know C++, though I now have a reason to learn,  but in most languages you either plan for localization or adding it in is very painful.

bobafetthotmail:
@omagasohe 99% of the game's text is in data files already, the source code already uses variables for more or less everything (just look at the data files, the only thing that is not there is maybe the names of the buttons in the user interface and the Preferences panel, not hard to move them to their own data file, the infrastructure to do so is already there)

As it is now, the easy and brutal solution is coding in the game "if the chosen language is X, go read the datafiles in the folder X", and the folder X contains the datafiles with the text in the right language.
Boom, you have localization.

To avoid the stats sync nightmare (keeping a zillion copies of datafiles means lots of pain to keep them in sync), keep all the stats in the english datafiles, while the localization datafiles have only the translated text entries.
This requires the game to be able to read stuff from 2 sources, the english one for the stats, and the localization data file for the translation (and if there is no translated entry it falls back to the english one), which is harder than the above, but should be doable. You know already what entries are text and what are numbers.

I'm suspecting that adding unicode support would be harder than this, but I haven't looked at those parts of the source so I don't really know.

omagasohe:
[...]

Issue #13, opened 2015-04-17
Improving numbers inputting


akien-mga:
There are two issues currently when trying to input numbers (e.g. to define how many credits one wants to borrow from the bank):
- Numpad keys do not seem to be recognised (on Linux at least)
- Normal number keys can't be used with a French azerty keyboard. The azerty keyboard [1] is always in bother in video games, as the number keys are accessed using the Shift key modifier. Pressing the number keys without Shift inputs a custom character: & é " ' ( - è _ ç à.
But Shift + one of the character keys listed above does not produce the numbers as it should in Endless Sky, instead I get:
  * 7 for Shift+1 (a.k.a. Shift+&)
  * Nothing for Shift+2 / Shift+é (which seems consistent with the fact that the "é" key does not input the "é" character either)
  * ' for Shift+3 / Shift+"
  * " for Shift+4 / Shift+'
  * 9 for Shift+5 / Shift+(
  * _ for Shift+6 / Shift+-
  * Nothing for Shift+7 / Shift+è (same issue as for 2)
  * - for Shift+8 / Shift+_
  * Nothing for Shift+9 / Shift+ç
  * Nothing for Shift+0 / Shift+à

Hope this helps :-)

[1] http://fr.wikipedia.org/wiki/AZERTY#mediaviewer/Fichier:Clavier-Azerty.svg

akien-mga:
Fixing the first point is most likely easier to do that fixing the second one, so feel free to split this issue if you want to add support for numpad first, and work on the funky French keyboard flavour later on :)

endless-sky:
Well, give the numeric keypad a try and see if it's working for you now.

To get the full AZERTY keyboard working right will probably require switching from plain ASCII text to full Unicode support and switching from bitmap fonts to using a font-rendering library. That's certainly doable, but it's a big enough change that I might put it off for a later milestone.

akien-mga:
With ec26e867a, the numeric keypad works fine (it works both with numlock active and inactive, but I think that's just fine).

With this change it's already way easier to input number with an AZERTY keyboard, so I guess you're right about putting off full Unicode support for a later milestone. Unicode support would most likely become relevant when/if you add support for internationalisation.

akien-mga:
Before looking into full Unicode support though, I would first have a look at how the Shift key modifier is handled. From what I could see, if I press Shift+& (which should input "1"), the current code interprets it as:
- SDLK_AMPERSAND is pressed
[...]

PR #3312, opened 2017-11-28
Unicode support - render with font files

flaviojs:
I looked at the [development roadmap](https://github.com/endless-sky/endless-sky/wiki/DevelopmentRoadmap), saw "Support for Unicode text." and decided to give it a shot.

This first step teaches the Font class about utf8, so now a multi-byte Unicode ~character~ code point shows up as 1 box instead of 1 ~box~ space per byte. Also, while redoing the truncate functions to support utf8, I made the ellipsis optional to simplify TruncateMiddle.

For the next step I need to know the Unicode code points of the last 3 glyphs. Can anyone tell me?

-----

This PR was highjacked for the rest of the unicode development.

#### Will implement this
- [x] Font can parse utf8
- [x] root data node for each font size in data/interfaces.txt
- [x] plugins can replace fonts
- [x] support font files as glyph sources (via [FreeType 2](https://www.freetype.org/))
- [x] allow glyph sources to compose multiple unicode code points into a single glyph (combining marks, variation sequences, positional forms)
- [x] multiple glyph sources in a single font (for full unicode support)
#### But not this
- arbitrary code points in image glyph sources
- plugins can append glyph sources to fonts
- unicode code point composition (via Pango? ICU? HarfBuzz? other?)
- different text directions (right to left, top to bottom)
- aggregate unicode stuff into a utility class


tehhowch:
For the box, curly quote, and double-curly?

flaviojs:
Yes.

flaviojs:
Likely candidates:
* [U+25AF](https://codepoints.net/U+25AF) &#x25AF;
* [U+2018](https://codepoints.net/U+2018) &#x2018;
* [U+201C](https://codepoints.net/U+201C) &#x201C;

flaviojs:
I also need to know how the font image was generated, I tried generating it in python with PIL (using Ubuntu-R.tff 0.83) but the edges look blurry in comparison.

tehhowch:
Builds fine - do you have some test text we can use to verify how this is working?

flaviojs:
Put this in credits.txt and start the game:
```
->á<- (C3 A1) U+00E1 LATIN SMALL LETTER A WITH ACUTE
->€<- (E2 82 AC) U+20AC EURO SIGN
->𐌀<- (F0 90 8C 80) U+10300 OLD ITALIC LETTER A
```
You will see a single box for each multi byte code point because they are not in the font image. In the current master you will see a space per byte instead.

To verify correctness, add this to `Font::Glyph`:
```
	if(c == 0x00E1)
		return Glyph('a', false);
	if(c == 0x20AC)
		return Glyph('b', false);
	if(c == 0x10300)
		return Glyph('c', false);
```
Recompile and open the game again. You will see a, b, and c where the boxes were.

tehhowch:
Can confirm.

endless-sky:
The font bitmaps are generated using [this tool](https://github.com/endless-sky/endless-sky-tools/blob/master/freetype.cpp) in the endless-sky-tools repository. But bitmapped fonts would probably not be a good solution for full Unicode support, because so many different glyphs would need to be rendered; instead, it would require using a library like freetype to render entire text strings and upload them to the GPU as textures.

flaviojs:
freetype2 uses code points (or UTF32) as the glyph index, so this will be needed in both cases.

With ttf fonts, it is currently impossible to have all unicode glyphs in a single font (65536 glyphs max), so it would need to allow a collection of fonts. Not sure if different font files would work well together.

With image fonts, the only problem I see is the advance calculation, which scales terribly.
I was able to fit all the 1139 standard visible (no "Other" category) unicode code points that are contained in `Ubuntu-R.ttf` in a 1240x1184 image for size 14 and 1600x1512 for size 18 (can be reduced further).

For full unicode support, initially I was envisioning Font able to hold multiple images. Each image would have their own code point glyphs. A root data node "font" would determine which images to load and what unicode code points each image holds, replaceable in plugins (maybe extendable too).
Maybe the extra images would be loaded on demand and/or individual advance calculations deferred until needed, depends on how it would perform.

But since fonts are preferable, I can try that first. It probably has slightly higher runtime cost than the current approach, and significantly lower load cost (advances already in the font).

-----

As for the tool, it produces sharper (thiner) letters than my initial size*2 guess (because fonts do not grow linearly). It looks like a standard grayscale filter is all that is needed to convert the generated bmp to png. I'll need to understand the "dpi black magic" in the tool before I can replicate it in python. =P

flaviojs:
Hijacking this PR for the rest of the unicode development.

flaviojs:
Apparently C++11 already has unicode stuff (std::u32string and std::codecvt), available since [Visual Studio 2015](https://blogs.msdn.microsoft.com/vcblog/2015/04/29/c111417-features-in-vs-2015-rc/), [Clang 3.0](https://clang.llvm.org/cxx_status.html), and [GCC 4.4](https://gcc.gnu.org/projects/cxx-status.html#cxx11).
Is this old enough to use?

flaviojs:
@endless-sky
About kerning, is it preferable to use our algorithm (which is inside advance calculation) or should we respect what the font wants?

endless-sky:
I think that correctly handling *all* unicode text will have to involve rendering and caching entire words, rather than individual glyphs, in order to handle combining diacriticals. So at that point, letter kerning would have to be handled by the typesetting library, whereas inter-word kerning and justification could still be handled by the game itself.

flaviojs:
Good idea, caching words instead of glyphs will take care of most subpixel/sharpness issues, and also matches the WrappedText use case.

flaviojs:
Text rendering with fonts is a big "can of worms"... I tried lots of approaches and finally got a satisfactory implementation for moving text.

Static text comparison:
(original)
with_image
(with Ubuntu-R.ttf 0.83)
with_font_file

oo13:
So far you don't intend to support non european languages, do you?

I rendered a japanese text with this PR as a test.

([IPA P Gothic](https://ipafont.ipa.go.jp/old/ipafont/download.html#en) font)
image
Excellent! The font rendering has no problem. &#x1F44D; 

By the way, WrappedText::Wrap() assumes char is unsigned.

flaviojs:
For now I'm focusing on Font internals with languages that I'm familiar with (english and european languages). After Font internals I'll check what needs to change outside and decide the next move. If it needs lots of changes, I might wait for this PR to be merged before continuing.

Languages with different layout directions, different word boundaries (not space) or tricky word compositions require a shaping library and some knowledge of the target language to verify correctness. I'll keep them in mind when I change the Font interface.

Bladewood:
> グレイズは変化に富んだ世界であり、晴れわたる砂浜から凍りついたツンドラまですべてがある。しかしおそらくいち

You can probably just repeatedly copypasta that to fill the box, assuming you fix the text going off-screen.

flaviojs:
The text going off screen is because the code doesn't know where japanese words end and thinks the text is a single word. If the shaping library can't tell when a word ends, the scripter will need to add some kind of space between each "word" (see http://jkorpela.fi/chars/spaces.html).

oo13:
I thought about the japanese layout. It can be independent of this PR because it needs only to change the WrappedText.

I'll be able to implement a method to support japanese layout. However, I'm afraid we'll need a kind of libpango when we want to support many languages.

I used this patch when I got that screenshot.
[japanese_test.patch.txt](https://github.com/endless-sky/endless-sky/files/1631554/japanese_test.patch.txt)

tehhowch:
This is looking really good :+1: 

Question: the new includes on the various *Panel.cpp ... future feature necessity?

flaviojs:
They need `#include "gl_header.h"` for glClear. They were getting it for free with Font.h but not anymore.

tehhowch:
Sounds like that should be done already in matter as part of iwyu

flaviojs:
Never used [iwyu](https://include-what-you-use.org/) before, but it seems like a good start for `scons test` :wink:

tehhowch:
This PR adds support for unicode display -- what else is need for supporting unicode input from the player?

I'm imagining a separate PR would be used to handle any text comparisons between game data and user input (e.g. `MapPanel::Search`)

flaviojs:
Search isn't critical, comparing current utf8 is probably good enough. To go beyond that you need [normalization](https://www.unicode.org/faq/normalization.html).

Unicode input requires proper handling all kinds of keyboard layouts (#13) and other input methods (https://wiki.libsdl.org/Tutorials/TextInput).

What I'm focused on is being able to draw all unicode characters at the same time (needs multiple font files), proper glyph layout of mixed text (includes reversing direction and composing glyphs), and boundary detection for formatting (ex: WrappedText).

Plugin support is not required by desired. :wink:

flaviojs:
Not sure how characters supported by multiple fonts should be handled, but this is good enough for now:
multiple_sources

flaviojs:
Ok, I'll stop here.

Modern fonts don't seem to need bitmap support from FreeType, even exotic ones.

I'm not sure if AtlasGlyphs should be kept, even with arbitrary code points. I was imagining images with alien alphabets mapped to private use areas, but it might be better to have text styling that allowed font selection. (No idea how hard it is to make a scalable font.)

What layout library to use:
 - [pango](http://www.pango.org/) recommends cairo and requires fontconfig, harfbuzz and glib. It includes support for [styling](https://developer.gnome.org/pango/stable/PangoMarkupFormat.html). Might be too much for this game.
 - [harfbuzz](https://harfbuzz.github.io/) is standalone. It also includes some unicode information (combining class, east-asian width, general category).
 - other alternatives didn't seem appropriate

@endless-sky
Your thoughts?

source/AtlasGlyphs.cpp
```diff
@@ -0,0 +1,403 @@
...
+	}
+	return str.length();
+}
+
+
+
+int AtlasGlyphs::Glyph(char32_t c)
+{
+	// ASCII
+	if(c >= 32 && c < 32 + GLYPHS - 3 && c != '\'' && c != '"')
```
tehhowch: Maybe add a `static const int ___ = 32` to help name this magic 32. Though anyone who is working with fonts likely knows its role without hints.

source/AtlasGlyphs.cpp
```diff
@@ -0,0 +1,403 @@
...
+		{
+			int maxD = 0;
+			int glyphWidth = 0;
+			uint32_t *begin = image.Pixels();
+			for(int y = 0; y < height; ++y)
+			{
+				// Find the last non-empty pixel in the previous glyph.
+				uint32_t *pend = begin + previous * width;
+				uint32_t *pit = pend + width;
+				while(pit != pend && (*--pit & mask) < half) {}
```
tehhowch: Probably unroll this advancement loop, i.e.
```
while(...)
{
    // Finding the last non-empty pixel in the previous glyph.
}
```

source/AtlasGlyphs.cpp
```diff
@@ -0,0 +1,403 @@
...
+	
+	GLfloat vertices[] = {
+		   0.f,    0.f, 0.f, 0.f,
+		   0.f, glyphH, 0.f, 1.f,
+		glyphW,    0.f, 1.f, 0.f,
+		glyphW, glyphH, 1.f, 1.f
+	};
+	glBufferData(GL_ARRAY_BUFFER, sizeof(vertices), vertices, GL_STATIC_DRAW);
+	
+	// connect the xy to the "vert" attribute of the vertex shader
```
tehhowch: Comment casing / sentence

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+	{
+		node.PrintTrace("was expecting font");
+		return false;
+	}
+	
+	// Get size.
+	size = 0;
+	for(const DataNode &child : node)
+	{
+		string key = (child.Size() >= 0 ? child.Token(0) : "");
```
tehhowch: I think you can do `const string &key = child.Token(0);` - this is done in other class's Load methods

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+		if(key != "size")
```
tehhowch: `if(key != "size" || child.Size() < 2)`

You access child.Value(1) without a bounds check, seems sensible to put it here.

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+			continue;
+		
+		if(size > 0)
```
tehhowch: `if(size)`

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+	{
+		node.PrintTrace("missing size");
+		return false;
+	}
+	
+	// Get glyph source.
+	sources.clear();
+	for(const DataNode &child : node)
+	{
+		if(child.Size() <= 0 || (child.Token(0) != "atlas" && child.Token(0) != "freetype"))
```
tehhowch:
```
static const set<string> allowed = {"atlas", "freetype"};
if(!allowed.count(child.Token(0))
    continue;
```

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+			continue;
+		string key = child.Token(0);
```
tehhowch: `const string &key`, as above. This can be done above the allowed check.

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+		
+		if(child.Size() <= 1 || child.Token(1).empty())
```
tehhowch: I think DataNode checks that the remainder of the line isn't just empty space, so this could just be
`if(child.Size() < 2)`

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+			return false;
+		}
+	}
+	if(sources.size() == 0)
+	{
+		node.PrintTrace("missing glyph sources (atlas or freetype)");
+		return false;
+	}
+	
+	// Ignore other children.
```
tehhowch: ?
(You've already stepped through all the children nodes at this point.)

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+			if(glyphs->Load(paths.back(), size))
+				sources.emplace_back(glyphs);
+		}
+		if(count == sources.size())
+		{
+			child.PrintTrace("load failed");
+			return false;
+		}
+	}
+	if(sources.size() == 0)
```
tehhowch: `sources.empty()`

source/Font.cpp
```diff
@@ -344,143 +360,177 @@ void Font::ShowUnderlines(bool show)
...
+	{
+		bool isUnsupported = true;
+		
+		// Supported data.
+		for(size_t i = 0; i < sources.size(); ++i)
+		{
+			size_t end = sources[i]->FindUnsupported(str, start);
+			if(end == start)
+				continue;
+			if(sections.size() > 0 && sections.back().first == i)
```
tehhowch: `if(!sections.empty() && ...)`

source/Font.cpp
```diff
@@ -344,143 +360,177 @@ void Font::ShowUnderlines(bool show)
...
+			isUnsupported = false;
+			start = end;
+			break;
+		}
+		
+		// Unsupported data.
+		if(isUnsupported)
+		{
+			size_t end = NextCodePoint(str, start);
+			if(sections.size() > 0 && sections.back().first == 0)
```
tehhowch: `if(!sections.empty() && ...)`

source/AtlasGlyphs.cpp
flaviojs: Inherited from original code, 32 is the ASCII code for space ' '. I can clarify it.

source/AtlasGlyphs.cpp
flaviojs: Inherited from the original code. I'll leave it as is.

source/AtlasGlyphs.cpp
flaviojs: Again, inherited. Ok.

source/Font.cpp
flaviojs: DataNode::Token does not check bounds.

source/Font.cpp
flaviojs: DataNode::Value checks bounds, not needed.

source/Font.cpp
flaviojs: The user can input negative sizes, which are invalid too.

source/Font.cpp
flaviojs: Just for 2? I rather not.

source/Font.cpp
flaviojs: DataNode::Token does not check bounds, a bounds check is needed.

source/Font.cpp
flaviojs: "" and `` are parsed as empty tokens.

source/Font.cpp
flaviojs: I stepped through all supported children. I'm not erroring on unrecognized children so that old code (stable version) can use data of future versions (unstable).

Example: a new type of glyph source is added to Font and a plugin that overrides the font data uses the new feature. The plugin can still provide a fallback source so it's still usable in the stable version.

source/Font.cpp
flaviojs: Ok.

source/Font.cpp
flaviojs: Ok.

source/Font.cpp
flaviojs: Ok.

source/Font.cpp
tehhowch: An empty line doesn't get parsed into a DataNode member though... DataFile skips empty / comment lines. So I read that as guaranteeing that if that `child` exists, `tokens[0]` exists.

flaviojs:
The only new dependency is FreeType 2.
This should work for windows users: http://gnuwin32.sourceforge.net/packages/freetype.htm

source/Font.cpp
tehhowch: True, and it also prints a trace in addition to returning `0.` So if `Token(1)` is out of bounds, you're double-printing a trace for the same error.

source/Font.cpp
tehhowch: So perhaps unify the error? Right now you have an error for positive size (i.e. size was defined more than once), then you read the value, and then have an error for non-positive integer.

Seems like you could do something like this:

```
size = 0;
for(const auto &child : node)
    if(child.Token(0) == "size" && child.Size() == 2)
    {
        if(child.Value(1) <= 0.)
            child.PrintTrace("Invalid 'size' specification:");
        else
            size = static_cast<int>child.Value(1);
    }
if(size <= 0)
    node.PrintTrace("No valid 'size' given:");
```

source/Font.cpp
tehhowch: Mostly I just hate seeing repeated `child.Token(0)...child.Token(0)...` etc. For 2, yes, using a `set.count()` check is perhaps a bit overzealous. It's certainly expressive :+1:

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+			child.PrintTrace("missing path");
+			return false;
+		}
+		
+		vector<string> paths;
+		paths.emplace_back(child.Token(1));
+		paths.emplace_back(Files::Resources() + child.Token(1));
+		for(const string &source : GameData::Sources())
+			paths.emplace_back(source + child.Token(1));
+		while(!paths.empty() && !Files::Exists(paths.back()))
```
tehhowch: Would it be better to check existence of the path before adding it to the path vector? If you have a non-existent path that gets sandwiched between valid paths, then it couldn't be removed using the method you've implemented.

(It would also be nice to convey some example values here.)

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+		return false;
+	
+	if(node.Size() < 1 || node.Token(0) != "font")
+	{
+		node.PrintTrace("was expecting font");
+		return false;
+	}
+	
+	// Get size.
+	size = 0;
```
tehhowch: Is this needed? The constructor initializes it to 0.

source/Font.cpp
```diff
@@ -12,91 +12,169 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+		while(!paths.empty() && !Files::Exists(paths.back()))
+			paths.pop_back();
+		if(paths.empty())
+		{
+			child.PrintTrace("path not found");
+			return false;
+		}
+		
+		size_t count = sources.size();
+		if(key == "atlas")
```
tehhowch: I wonder if you can use the implicit constructors to simplify this even further, i.e. do away with checking count & sources.size()
```
sources.emplace_back(paths.back())
if(!sources.back().IsLoaded())
    child.PrintTrace("Load failed for valid path '" + paths.back() + "':");
```
It probably can't simplify to that extent, but surely a check to sources.back().IsLoaded() could be implemented.

source/Font.cpp
flaviojs: I'm gathering all candidate paths and getting the last one that exists (so plugins can replace them).
Extra temporary memory is cheaper than extra disk access.

Example:
```
font
	size 14
	atlas "images/font/ubuntu14r.png"
	freetype "fonts/Ubuntu-R.ttf"
	freetype "fonts/ipagp.ttf"
	freetype "/full/path/to/some/font.ttf"
```

source/Font.cpp
flaviojs: Although it doesn't happen with the current use, a failed load can leave size as non-0.

source/Font.cpp
flaviojs: `sources` should only contain valid interfaces.

Concrete classes can have different load mechanics, like AtlasGlyphs not needing the size.
When new features are introduced to a type of glyph source, it will probably receive the child DataNode so it can read the grandchildren.

tehhowch: Yeah, i left off the `.pop_back()` and other error handling cleanup from the example, since the example wasn't likely to be implementable, but served well to convey the intent.

source/Font.cpp
flaviojs: Didn't realize index 0 always exists, I'll note it down in the function comment.
I can also see that DataFile suffers from the problem that WrappedText had, it assumes char is unsigned (treating unquoted non-ascii bytes as space).

source/Font.cpp
flaviojs: Since this is only executed at startup, I think it's better too have clear errors and as deep in the chain as possible.
I like your "Error message:" style, will use it.

source/DataNode.cpp
```diff
@@ -63,6 +63,7 @@ int DataNode::Size() const
 
 
 // Get the token with the given index. No bounds checking is done.
+// DataFile loading garantees index 0 always exists.
```
tehhowch: `guarantees`

source/DataNode.h
```diff
@@ -38,6 +38,7 @@ class DataNode {
 	// Get the number of tokens in this node.
 	int Size() const;
 	// Get the token at the given index. No bounds checking is done internally.
+	// DataFile loading garantees index 0 always exists.
```
tehhowch: `guarantees`

source/Font.cpp
```diff
@@ -12,91 +12,168 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+	{
+		node.PrintTrace("Must have one font size:");
+		return false;
+	}
+	
+	// Get glyph source.
+	sources.clear();
+	for(const DataNode &child : node)
+	{
+		string key = child.Token(0);
```
tehhowch: This should be `const string &key` to match the style used in other classes.

source/DataNode.cpp
flaviojs: Ok.

source/DataNode.h
flaviojs: Ok.

source/Font.cpp
flaviojs: Ok.

oo13:
This PR raises std::bad_alloc when loading a savefile has non ASCII characters.
Example: 
[savefile.zip](https://github.com/endless-sky/endless-sky/files/1649610/savefile.zip)

This patch fixed it.
```
diff --git a/source/DataFile.cpp b/source/DataFile.cpp
index e08fdcbd..6b108269 100644
--- a/source/DataFile.cpp
+++ b/source/DataFile.cpp
@@ -147,7 +147,7 @@ void DataFile::Load(const char *it, const char *end)
 			const char *start = it;
 			
 			// Find the end of this token.
-			while(*it != '\n' && (isQuoted ? (*it != endQuote) : (*it > ' ')))
+			while(*it != '\n' && (isQuoted ? (*it != endQuote) : (*it > ' ' || *it < 0)))
 				++it;
 			
 			// It ought to be legal to construct a string from an empty iterator
```

flaviojs:
Thanks, I missed that one.
When I add shaping I'll probably aggregate unicode stuff into it's own utility class and make a utf8 iterator.

source/Font.cpp
```diff
@@ -344,143 +359,177 @@ void Font::ShowUnderlines(bool show)
...
+				continue;
+			if(!sections.empty() && sections.back().first == i)
+				sections.back().second = end;
+			else
+				sections.emplace_back(i, end);
+			isUnsupported = false;
+			start = end;
+			break;
+		}
+
```
oo13: This longest-match logic can choose a glyph of a lower priority font.

For instance, when I have a font setting like this
```
font
	size 14
	atlas "images/font/ubuntu14r.png"
	freetype "images/fonts/ipagp.ttf"
```
and let it render a string "あA", then "A" is rendered by ipagp.ttf because ipagp.ttf contains both glyphs of "あ" and "A."

I expect that "A" is rendered by ubuntu14r.png.

tehhowch: I guess that is a preference behavior.. to me it seems sensible to use the same font to render the characters in a given string if possible, and only do piecemeal when that cannot be done.

However I can see that it would be weird if 99/100 sentences used an Ubuntu "A" and then the one sentence with a glyph has a different look for every other character too.

tehhowch:
Do classes like Sound need updating as well?
https://github.com/endless-sky/endless-sky/blob/d6b92f48810c4b36f109e8e49e1e29e5a9f34d6d/source/Sound.cpp#L42-47

source/Font.cpp
flaviojs: Before implementing it like that I debated myself over this particular issue and came to the conclusion that both cases are perfectly valid preferences for various circumstances.

Since the objective is to get full unicode support (which needs multiple font files), in the long term the most likely case is to have multiple fonts that are compatible with each other and cover all the ranges. The noto fonts from google are aiming for this.

Another way to overcome this is to have styling that allows font selection.

flaviojs:
That kind of case won't care about utf8 since it's only checking that the end is a specific ASCII sequence.

The utf8 standard specifies that the correct encoding of a code point use only the minimum number of bytes required, so ASCII is the only correct way to encode ASCII in utf8.

Fzzr:
@flaviojs would you mind bringing this up to date with `master`? It would be nice to get it into 9.9 potentially, and it's the big first step toward actual localization.

tehhowch:
Closing in preference of #4123

ThundySGI:
> I thought about the japanese layout. It can be independent of this PR because it needs only to change the WrappedText.

>I'll be able to implement a method to support japanese layout. However, I'm afraid we'll need a kind of libpango when we want to support many languages.

>I used this patch when I got that screenshot.
[japanese_test.patch.txt](https://github.com/endless-sky/endless-sky/files/1631554/japanese_test.patch.txt)

Can you help me with Japanese translation?
I managed to put all text data to keywords dictionaries.
I can make one more for Japanese.
[https://github.com/endless-sky/endless-sky/issues/558](url)

oo13:
I provided a japanese translation that is contained in [my lang-ja plugin](https://www.dropbox.com/s/ovzshpi67g3gkdq/endless-sky-plugin-lang-ja-5d4c59f-3.zip?dl=1). (cf. https://github.com/endless-sky/endless-sky/issues/558#issuecomment-582915541)

ThundySGI:
> I provided a japanese translation that is contained in [my lang-ja plugin](https://www.dropbox.com/s/ovzshpi67g3gkdq/endless-sky-plugin-lang-ja-5d4c59f-3.zip?dl=1). (cf. [#558 (comment)](https://github.com/endless-sky/endless-sky/issues/558#issuecomment-582915541))

I'm confused, You did so much good work and nothing was merged to game code!!! 
What a waste....
You had working utf-8 display, professional translation system in POT in 2019r.
Now its 2025 and game still stuck on only english, ASCII, and obsolete charmap,
with no modular text system,

My system is much simpler than that.
Just Keywords, and dictionaries - easy for automatic translate.
and backward compatibility, any text without keywords still work.

I'm devastated.
If your professional approach hasn't been merged into the code,
what chance do my sytem has?

Issue #1825, opened 2016-11-15
Really long hails go off the screen


Bladewood:
![screenshot 2016-11-15 17 28 10](https://cloud.githubusercontent.com/assets/16506683/20326937/0777f058-ab59-11e6-9f60-e792f5805917.png)
On an unrelated note, typo in the song name.

KiLEdEnNis:
I feel like these kinda hails become too long, anyone else?
in case people wonder about the typo: wont should be won't or will not

Pointedstick:
If MZ would accept my pull request to clean up the music-related hails (https://github.com/endless-sky/endless-sky/pull/1754), this would go away. 🙂

tehhowch:
While the OP image is no longer possible due to #1754, there's nothing preventing a future hail from going off-screen. Seems like maybe the messages should be drawn within the "message box" with the help of WrappedText rather than just simply drawn.
Relevant:
https://github.com/endless-sky/endless-sky/blob/c6c5d149262e74662c0c82f076af35f130dfdfbc/source/Engine.cpp#L782-L800

tehhowch:
Setting the wrap width to an absurdly small value just for demonstration:
![image](https://user-images.githubusercontent.com/20871346/33348573-81c2b712-d45c-11e7-9b44-5e825c66c34a.png)

Pointedstick:
Should be fixed with https://github.com/endless-sky/endless-sky/pull/3313 being merged.

PR #3313, opened 2017-11-28
Wrap messages when drawing them to prevent overflow

tehhowch:
Refs #1825 
For this screenshot I set the wrap width to be really short, in order to illustrate the wrapping.
image

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -401,7 +401,7 @@ interface "hud"
 		to 120 0 bottom left
 	box "messages"
 		from 120 0 bottom left
-		to -80 -200 bottom right
+		to -100 -200 bottom right
 	box "ammo"
 		from -80 460 top right
 		to 0 0 bottom right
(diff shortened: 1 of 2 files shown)


jafdy:
Is it  supposed to be right over to the side like that?

KiLEdEnNis:
could ya also include a screenshot with the actual feature implemented? (no absurdly small width)

tehhowch:
With wrap at the normal position (i.e. constraining messages to be within the box defined in the interfaces file) it looks like you expect because the normal hails are not so long that they wrap. Only really long hails / messages are going to wrap.
This PR adds support for longer hails/messages, and prevents bad UI for those who use a modified interface that might not allocate as much horizontal space for messages.
I'll try to get a message long enough to wrap

tehhowch:
By tweaking those planet names, I can get it to wrap at the normal position (which indicates that the normal position is not quite correct, as well):
image

tehhowch:
Adding a 20pt pad to the right edge of the messages box fixes that issue. This same pad could be achieved via adding margins in the WrappedText object, but changing the interface file means one less operation needs to be done to draw the text:
<details><summary> fixed box padding </summary>

image
</details>

Issue #3287, opened 2017-11-25
The Proton Gun needs a little something


Pointedstick:
It's just a fairly lackluster weapon. It doesn't have the Particle Cannon's extreme range or pushback, nor is it small enough that you can fit it on anything smaller than a warship or very large multipurpose ship and have much room left.

Ideas for improvement:
- Reduce the size to 20 tons
- Increase the damage
- Increase the range
- Inflict special damage(heat, ion, disruption, something else?)
- Something else?

KiLEdEnNis:
tbh just increasing the RoF and range would make it a fun long-range "bombardment" type weapon, and the humans don't really have that yet
what does the rest think about this?

Amazinite:
* Increase damage: Yes
   * The Particle Cannon should get this too
* Increased range: questionable. I think increasing the range on the Proton Gun warrants also changing the other cannon, blaster, and laser ranges (i.e. fully fleshing out #3266).
* Increased fire rate: the Proton Gun already fires super fast.
* Special damage type: No.
* Reduced size: unnecessary in my eyes.

>tbh just increasing the RoF and range would make it a fun long-range "bombardment" type weapon, and the humans don't really have that yet

Humans don't need everything.

Lineth:
T1 I near-exclusively use plasma weapons so anything to make particle and proton guns more attractive would be lovely.

KiLEdEnNis:
> Humans don't need everything.

true, but something as simple as this wouldn't hurt
if the range got buffed, it'd still be more effective up close due to the inaccuracy, but a damage+range buff would make it good

Amazinite:
>true, but something as simple as this wouldn't hurt
if the range got buffed, it'd still be more effective up close due to the inaccuracy, but a damage+range buff would make it good

By lore, the Proton Gun is suppose to be a crappier Particle Cannon, so pretty much anything we do to the Proton Gun means we also need to look over the Particle Cannon. i.e. a range buff to the Proton Gun means a range buff to the Particle Cannon, which I don't think the Particle Cannon needs.

KiLEdEnNis:
fair enough, as long as both options actually do some damage I'm fine with it

Pointedstick:
I'd be fine with it just being a crappier Particle Cannon if it weighed a lot less so I could stuff two of 'em onto a Hawk. :slightly_smiling_face:  KISS

Amazinite:
[...]

PR #3315, opened 2017-11-29
Slightly buffed the Proton Gun

Amazinite:
Reference: #3287 

* Submunition lifetime: 20 -> 24
  * Range: 528 -> 624 (96 pixel / 18% increase)
  * This makes the Proton Gun's range about 83% that of the Particle Cannon instead of the previous 70%.
* Submunition shield damage: 8 -> 8.4 (5% increase)
  * Shield DPS: 180 -> 189
* Submunition hull damage: 6.6 -> 7 (6% increase)
  * Hull DPS: 148.5 -> 157.5

An overall 5.5% increase in damage.

The increased range and slightly increased damage to the Proton Gun makes it a more viable alternative to the Plasma Cannon and its powerful heat damage or the Particle Cannon and it's massive range and substantial hit force, and the inaccuracy of the submunitions makes it an interesting weapon to use at range but a very powerful weapon up close.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -443,10 +443,10 @@ outfit "Proton Fragment"
 			"frame rate" 2
 		"hit effect" "proton impact"
 		"inaccuracy" 2
-		"lifetime" 20
+		"lifetime" 24
 		"hit force" 6
-		"shield damage" 8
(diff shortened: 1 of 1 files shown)


Amazinite:
👌

PR #3271, opened 2017-11-24
Adjusted (mostly reduced) turret turn rates

Pointedstick:
With this change, the largest turrets will seriously struggle against swarms of small ships, but still work fine against large slow-moving capital ships. Instead of using all Dual Sunbeam turrets and excelling against everything, you might want to swap out one or two for Electron Turrets to preserve your anti-fighter capabilities, or even employ dedicated anti-fighter picket ships armed with faster-turning anti-fighter turrets to protect your capital ships.

Partially addresses https://github.com/endless-sky/endless-sky/issues/3265

PR #3316, opened 2017-11-30
Proton Turret

Pointedstick:
Introducing the Syndicate Proton Turret.

proton turret - outfitter

proton turret - in game

A worse weapon in most ways than the Heavy Laser Turret, but a bit smaller and longer-ranged. Nearly a drop-in replacement. Since low range is one of the laser weapons' weaknesses, this should make the Proton Turret actually competitive in the turret market (especially if https://github.com/endless-sky/endless-sky/pull/3315 is merged)--at least until the Electron and Plasma turrets come along, but that's to be expected.

New ship variants and fleet options are included.

Turn rate is "post-https://github.com/endless-sky/endless-sky/pull/3271" and should not require modification when @endless-sky reduces turret turn rates across the board.

HiDPI assets/pull request: https://github.com/endless-sky/endless-sky-high-dpi/pull/31

Blender and GIMP assets: [Proton Turret assets.zip](https://github.com/endless-sky/endless-sky/files/1524345/Proton.Turret.assets.zip)


tehhowch:
Neat. Perhaps alphabetize the new ship variants, e.g. `"Argosy (Proton)"` comes after `"Argosy (Missile)"`

data/weapons.txt
```diff
@@ -457,6 +457,32 @@ effect "proton impact"
 	"velocity scale" -.005
 
 
+outfit "Proton Turret"
+	category "Turrets"
+	cost 260000
+	thumbnail "outfit/proton turret"
+	"mass" 36
```
Amazinite: I'd make it 32 tons. 36 (mass of turret) - 28 (mass of gun) = 8 additional mass on this turret in order to carry one gun. The average additional mass per gun for human turrets is 3.6, which rounds up to 4.

Pointedstick: I figured that the turret base itself was pretty heavy to support a weapon of this size. This is a fairly chunky Syndicate weapon, after all. Not that I wouldn't love 32-ton Proton Turrets, but that seems a little bit _too_ light. What do other folks think?

data/weapons.txt
```diff
@@ -457,6 +457,32 @@ effect "proton impact"
...
+		"inaccuracy" 4
+		"turret turn" 2.5
+		"submunition" "Proton Fragment" 3
+		"velocity" 24
+		"lifetime" 2
+		"reload" 8
+		"firing energy" 15
+		"firing force" 2
+		"firing heat" 60
+	description "The Syndicate Proton Turret is a relatively new product designed to compete with Deep Sky's perennially-popular laser turrets for the anti-fighter defense market. Though a cruder weapon than its competition, the Proton Turret maintains a loyal following among captains dissatisfied with laser weapons' short range.
```
tehhowch: `The Syndicate's Proton Turret is a relatively new product designed to compete with Deep Sky's perennially-popular laser turrets in the anti-fighter defense market. Though cruder than its competition, the Proton Turret maintains a loyal following among captains dissatisfied with laser weapons' short range.`

 - possessive `Syndicate`,
 - `for` -> `in`, 
 - ~~a~~ ~~weapon~~

data/weapons.txt
```diff
@@ -457,6 +457,32 @@ effect "proton impact"
...
+	"weapon capacity" -36
+	"turret mounts" -1
+	weapon
+		sprite "projectile/proton"
+		"hardpoint sprite" "hardpoint/proton turret"
+		"hardpoint offset" 14.
+		sound "proton"
+		"hit effect" "proton impact" 3
+		"inaccuracy" 4
+		"turret turn" 2.5
```
tehhowch: Is this turn rate considered to be pre- or post- #3271 ?

data/weapons.txt
Amazinite: Well the Plasma Turret at 60 tons and 25 tons per gun is using 10 tons to support both guns, which comes out to 5 tons each. The Proton Gun is only 3 tons heavier than the Plasma Cannon, so I'd say the Proton Turret should have at MOST 6 extra tons added on to it. But then we have the Electron Turret, which is 54 tons and 21 tons per gun, which means 12 tons of support/6 tons of support per gun. So clearly, the chunkiness of the gun isn't at play here, since the slim Electron Beam has more weight supporting it than the chunky Plasma Cannon. Since it's only a singular gun attached to a turret, I think that 4 tons is completely reasonable.

data/weapons.txt
Bladewood: Also, `perennially-popular` -> `perennially popular`; -ly adverbs should not have hyphens.

data/weapons.txt
Pointedstick: Post.

data/weapons.txt
Pointedstick: I'll do the last two. There's nothing wrong with "The Syndicate Proton Turret..."

tehhowch: Indeed. It was more to parallel the possessive Deep Sky usage.

data/weapons.txt
Amazinite: @ the 8 minute ago edit
>Not that I wouldn't love 32-ton Proton Turrets, but that seems a little bit too light. I could maybe go for 34 or 35 though.

I can agree on a 34 ton turret.

Pointedstick: I tend to consider the extra mass of the turret as embodied by the base and the support structure, not on a per-gun basis. Evidence for this would be the Quad Blaster Turret vs the regular Blaster Turret: in each case, the base and mount contribute only five tons. Those same five tons support two guns or four. Also, we have several other single-large-weapon turrets that also have gun versions to use as a point of comparison:
- Sunbeam turret is 12 mass for the base and mount
- Moonbeam turret is 16 mass for the base and mount

That said, I think going down to 35 tons would be reasonable. 34 is... well, I want this PR to be merged. :slightly_smiling_face:

Amazinite: I'd argue that the Sunbeam and Moonbeam are alien weapons so it's different than comparing to other human turrets. :)
And although the Moonbeam Turret is a single gun, the fire rate on that gun is increased when compared to the gun version of the Moonbeam.

tehhowch:
Are there any variant tweaks to be done using the smaller mass?

(The Protector variant actually might be good :grin: )

Pointedstick:
Should be all the same. The Proton Protector can stuff 6 of these in!

endless-sky:
For consistency, the hardpoint sprite should be rendered using the same light angle as the ship sprites (which will put its left side in shadow). It also needs a drop shadow (with a 10-pixel blur in the 2x version) under it - see the existing hardpoint sprite GIMP files for examples. And, the 2x version should have even dimensions so that the 1x version is exactly half the size of it.

Pointedstick:
Wooho!

Issue #3317, opened 2017-11-30
Republic Reputation 1 persists after "start of hostilities"


mootootwo:
There is an odd gap between "FW Prisoner Parole 1" and "FW Prisoner Parole 2" where Event "start of hostilities" triggers after 7 days (correctly) but you still have a positive Republic reputation until completing "FW Prisoner Parole 2"

I believe this is because the reputation changes that occur with "Start of Hostilities" is defined in the missions "Defend Sabik" and "FW Prisoner Parole 1".

I suggest moving that reputation change into the event, if possible, so that the republic will become properly hostile to you after the cease-fire expires.

If "Defend Sabik" and "FW Prisoner Parole 2" do need to have different levels of reputation change (current is -10 and -1000) then the -10 could be encoded in the event and the -1000 could remain at the end of "FW Prisoner Parole 2" -- or two events could be created.

mootootwo:
The upshot of this is that you can fly around republic space with impunity, and indefinitely, until you go back to Deep to complete "FW Prisoner Parole 2".

tehhowch:
>with impunity

They'll still punish you if you do something wrong, so 'with impunity' is not correct.

tehhowch:
The`start of hostilities` event refers only to the relationship between the Free Worlds and the Republic so bundling a player rep change into it should not be done - other storylines may use this same event and probably should not have the associated player rep change, even though the Free Worlds should be at war with the Republic.

Per https://github.com/endless-sky/endless-sky/issues/264#issuecomment-159800618 we can infer that if you are not being an enemy you get a bit of leeway. 

Currently your mission escorts are of the Escort government, so they won't incur the hostilities either (even though they probably should be Free Worlds). I believe this was changed and then reverted previously.

mootootwo:
I'm not sure why it is assumed to be different from player rep -- player rep was set to negative as soon as they join FW and they are still in the FW when hostilities resume.  

Rep is again set negative (even MORE negative) after the player lands, but they have not done anything to further anger the republic.

Part of the dramatic arc on the mission chain is the race against the clock -- or at least a tight deadline with no room for dawdling.  Instead I can meander up to the Deep and buy some atomic engines.  Seems unintended.

Based on the comment you linked, it is reasonable to assume that the -10 and -1000 rep changes are intentional and should be preserved, so that it is easier to regain your good graces before the hostage exchange, but I think the resumed hostilities deadline should still apply to you and your escorts.

[...]

Issue #2534, opened 2017-05-20
Safe passage does not apply to escorts with "Free Worlds" government


stinkray:
Related Commit: b5c6229838bd5d1bb769e6c034460b6d9639dac9 
By changing the government of the escorts in these missions from Escort to Free Worlds leads to the Free Worlds prisoner transport convoy and the Navy garrison over Kaus Borealis seeing each other as hostile parties, leading to the ultimate decimation of the fleet and failure of the mission. 
Either this commit should be reverted, or there needs to be some changes to how safe passage is managed.

gchapyzh:
Can confirm said issue. Landing in the target world does not complete the mission either.

endless-sky:
Reverted in 31d97c7.

gchapyzh:
That fixed it. Thanks!

PR #3318, opened 2017-11-30
Simplified FW Prisoner Parole mission set

tehhowch:
Refs #3317 
 - The player always knows the deadline (and is required to return before it ends to avoid failing the mission)
 - The Escort government can safely be swapped to be Free Worlds (as was originally done in b5c6229838bd5d1bb769e6c034460b6d9639dac9 ) since the `temporary ceasefire` event applies to FW and Republic)
   - https://github.com/endless-sky/endless-sky/issues/2534 and https://github.com/endless-sky/endless-sky/commit/31d97c7fe84e6dc7c7b8811aff921580a44f12fe occurred because the other set of changes in b5c6229838bd5d1bb769e6c034460b6d9639dac9  modified a mission in which only the player gets permission to travel. In other words, only half of b5c6229838bd5d1bb769e6c034460b6d9639dac9 needed to be reverted

### DIFF
--- a/data/free worlds start.txt
+++ b/data/free worlds start.txt
@@ -167,24 +167,24 @@ event "prepare for battle of Kornephoros"
 
 mission "FW Prisoner Parole"
 	name "Prisoner Parole"
-	description "Escort a fleet of transports carrying prisoners of war to <planet>. The temporary ceasefire will end on <date>."
+	description "Escort a fleet of transports carrying prisoners of war to <stopovers>. The temporary ceasefire will end on <date>, so return to <planet> before then."
 	deadline 7
 	source "Deep"
-	destination "New Tibet"
+	stopover "New Tibet"
 	to offer
 		has "Liberate Kornephoros: done"
 	on fail
 		dialog `You have failed an essential Free Worlds mission. If you want to complete the story line, revert to the autosave or another earlier snapshot of the game.`
 	
 	npc accompany save
-		government "Escort"
+		government "Free Worlds"
 		personality timid escort
 		ship "Falcon (Heavy)" "F.S. Starry Warrior"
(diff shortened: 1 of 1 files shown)


tehhowch:
Interesting factoid: stopovers weren't supported at the time Prisoner Parole 1 & 2 were written:
 - PP 1 & 2 added: 7f0d0c3b00bc3c4e7171e09a3fe4698929977173 (30 Nov, 2014)
 - stopovers added: d46d0991f18e9026b4229ab183419b798e4935f4 (28 Dec, 2015)

mootootwo:
This is an interesting solution, the "stopovers" is a great approach.

One side effect is that it would be impossible to successfully deliver the passengers, but then be caught out in Republic space without failing the missions.  

Possibly that is fine, but I do remember the first time seeing this mission, there was a strong temptation to exploit the ceasefire period to go run errands and shop for outfits in the Republic.  Would be nice to preserve that option.

I recognize I initially complained about consequence free gallivanting -- but I'm not sure it should be entirely impossible to gallivant, either.  Let me see if I can cobble together an alternate solution and you can compare.

tehhowch:
>I recognize I initially complained about consequence free gallivanting -- but I'm not sure it should be entirely impossible to gallivant, either. Let me see if I can cobble together an alternate solution and you can compare.

The player is generally always able to traipse about in Republic space. This PR makes it just not allowable when the player is dragging ships belonging to the Free Worlds military with them.

mootootwo:
compare to #3322

endless-sky:
In general, I prefer to use two missions rather than using stopovers, because it makes the player's current destination clearer in the UI. Stopovers are mainly useful for randomly generated missions that can't be split up into two.

PR #3322, opened 2017-11-30
enable "start of hostilities" to trigger Republic rep penalty

mootootwo:
This is a resolution to #3317 , as an alternate solution to #3318 

This change causes the player to become hostile with the Republic after "FW Prisoner Parole" is completed and while "FW Prisoner Parole 2" is still active.

The event timer in "FW Prisoner Parole" is extended to 8 days, as the mission can still be successful by the end of the 7th day (on the deadline) but the hostility event was triggering at the start of that day.

The -10 reputation setting in mission "Defend Sabik" is removed, as it is now redundant with the -10 setting in the event its self.

The -1000 reputation setting in mission "FW Prisoner Parole 2" still exists as it happens after and supersedes the event.

### DIFF
--- a/data/events.txt
+++ b/data/events.txt
@@ -798,7 +798,7 @@ event "start of hostilities"
 	government "Free Worlds"
 		"attitude toward"
 			"Republic" -.2
-	"reputation: Republic" = -10
+	"reputation: Republic" <?= -10
 
 
 
--- a/data/events.txt
+++ b/data/events.txt
@@ -798,6 +798,7 @@ event "start of hostilities"
 	government "Free Worlds"
 		"attitude toward"
 			"Republic" -.2
+	"reputation: Republic" = -10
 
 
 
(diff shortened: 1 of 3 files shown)


data/events.txt
```diff
@@ -798,6 +798,7 @@ event "start of hostilities"
 	government "Free Worlds"
 		"attitude toward"
 			"Republic" -.2
+	"reputation: Republic" = -10
```
tehhowch: This should be `<?=` rather than `=` as `=` provides a free rep boost if the player has done things that would otherwise decrease their reputation.

mootootwo: It copies the existing functionality.  I'm happy to change it though, this makes sense.

mootootwo:
Per your suggestion @tehhowch I have changed the `=` to `<?=` so instead of setting to 10 or 1000, it takes the minimum of current player reputation or the new number.  I have modified this in both the event as well as the "FW Prisoner Parole 2" for consistency.

I suspect there may be some other missions where this should be done as well, will have to search.

PR #3325, opened 2017-12-01
Added defense missions to the Core against Korath Attackers

czartrak:
It seems logical that there should be core missions to defend against the Korath, so I wrote a couple missions where you fight off some Raiders

### DIFF
--- a/data/jobs.txt
+++ b/data/jobs.txt
@@ -1503,15 +1503,15 @@ mission "Raider Attack 1"
 	on offer
 		conversation
 			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by alien raiders! We need every combat-worthy ship to join in the defenses!" The authorities will probably pay you quite well if you assist them, but this could also be an easy way to get yourself killed.`
-            choice
-                `    (Stay here until the fight is over.)`
+	choice
+	    	`    (Stay here until the fight is over.)`
                     decline
-                `    (Join the defense fleet.)`
-            `The local Syndicate defense forces are preparing to repel the alien attack. You join them, and take off together...`
+		`    (Join the defense fleet.)
+	`The local Syndicate defense forces are preparing to repel the alien attack. You join them, and take off together...`
 				launch
 	on complete
 		payment 550000
-		dialog "The government of <planet> pays you <payment> for helping to drive off the raiders."
+		dialog `The government of <planet> pays you <payment> for helping to drive off the raiders.`
 		
 mission "Raider Attack 2"
 	name "Defend <planet>"
@@ -1538,15 +1538,15 @@ mission "Raider Attack 2"
 	on offer
 		conversation
 			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by alien raiders! We need every combat-worthy ship to join in the defenses!" The authorities will probably pay you quite well if you assist them, but this could also be an easy way to get yourself killed.`
-            choice
-                `    (Stay here until the fight is over.)`
+	choice
+	    	`    (Stay here until the fight is over.)`
                     decline
-                `    (Join the defense fleet.)`
-            `The local Syndicate defense forces are preparing to repel the alien attack. You join them, and take off together...`
+		`    (Join the defense fleet.)
+	`The local Syndicate defense forces are preparing to repel the alien attack. You join them, and take off together...`
 				launch
 	on complete
 		payment 750000
-		dialog "The government of <planet> pays you <payment> for helping to drive off the raiders."
+		dialog `The government of <planet> pays you <payment> for helping to drive off the raiders.`
 
 mission "Cargo [0]"
 	name "Delivery to <planet>"
--- a/data/jobs.txt
+++ b/data/jobs.txt
@@ -1492,10 +1492,9 @@ mission "Raider Attack 1"
        		planet Shangri-la
        		planet Nimbus
 	npc evade
-		personality heroic staying
+		personality timid unconstrained
 		government Korath
-		fleet "Korath Raid"
-		fleet "Korath Raid"
+		fleet "Korath Raid" 2
 	npc
 		personality heroic staying
 		government Syndicate
@@ -1527,16 +1526,14 @@ mission "Raider Attack 2"
       		planet Shangri-la
         	planet Nimbus
 	npc evade
-		personality heroic staying
+		personality timid unconstrained
 		government Korath
-		fleet "Korath Raid"
-		fleet "Korath Raid"
-		fleet "Korath Raid"
+		fleet "Korath Raid" 3
+		
 	npc
 		personality heroic staying
 		government Syndicate
-		fleet "Large Syndicate"
-		fleet "Large Syndicate"
+		fleet "Large Syndicate" 2
 	on offer
 		conversation
 			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by alien raiders! We need every combat-worthy ship to join in the defenses!" The authorities will probably pay you quite well if you assist them, but this could also be an easy way to get yourself killed.`
--- a/data/jobs.txt
+++ b/data/jobs.txt
@@ -1486,14 +1486,14 @@ mission "Raider Attack 1"
 	to offer
 		"combat rating" > 550
 		random < 7
-		has "main plot: completed"
+		has "main plot completed"
 	source
 		planet "Sundive"
 		planet "Shangri-La"
 		planet "Nimbus"
 	npc evade
 		government "Korath"
-		personality timid unconstrained uninterested
+		personality target timid unconstrained uninterested harvests plunders
 		fleet "Korath Raid" 2
 	npc
 		government "Syndicate"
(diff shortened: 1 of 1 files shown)


Nescio0:
Don't you think such missions would make it too easy to acquire jump drives?

Amazinite:
>Don't you think such missions would make it too easy to acquire jump drives?

The random trigger values are low enough that farming jump drives with these missions isn't a very effective method of farming.

tehhowch:
Easily addressed as well... give the Syndicate allies `personality vindictive`

Brick63:
i like the idea as it adds another layer of why you should be careful in that part of space and alerts new players to the korath's existence

gunqqer:
@tehhowch yeah that would fix it, but then it would be almost impossible to get jump drives from the exiles, which personally I think I unneeded. If we were to impliment this I feel like a better way of balancing it would be needed

KiLEdEnNis:
I think Tehhowch meant to have the Syndicate ships that get spawned IN the raid defense should get the vindictive tag, so your chances to get Jump Drives outside of those missions are still the exact same
right, Tehh?

gunqqer:
That I would hesitently support. Don't like it a ton because I hate when NPCs kill my boarding target, but overall it makes it *slightly* easier to capture/board them and adds a good element to the game

tehhowch:
To reduce the "here, have a jump drive" nature of these missions, we've done several things:
 - The Korath having `timid` means they will clump up, making crossfire from the Syndicate NPCs more likely to hit non-targeted ships (i.e. a disabled target will get destroyed when the NPCs change to the other one).
 - The use of `heroic vindictive` on the Syndicate NPCs means they will overkill their last target (if they can't find a new target, they shoot their old target, even if it is disabled or already blowing up). This means more crossfire chances to hit a nearby disabled Korath Raider.
 - The lack of `staying` on the Raiders means they will abandon the fight if they start losing too much
   - The player has to choose to pursue it (though it won't despawn, so they could run into it again) or continue fighting in this system.
   - If `uninterested` is added, then the Raiders will not immediately follow or return to the player's system.


The last personality tweak I'd add is `uninterested` to the Korath fleets, to get proper "evade" behavior for any Raiders that decide to flee and live. Adding it will give them full control over where they can travel, rather than forcing them to follow the player around.

data/jobs.txt
```diff
@@ -1477,6 +1477,74 @@ mission "Core Pirate Attack"
...
+mission "Raider Attack 1"
+	name "Defend <planet>"
+	description "Defeat an alien raid on <destination>."
+	minor
+	repeat
+	to offer
+		"combat rating" > 550
+		random < 7
+	source
+		planet Sheratan
```
tehhowch: "Sheratan" is a system

data/jobs.txt
```diff
@@ -1477,6 +1477,74 @@ mission "Core Pirate Attack"
...
+		planet Shangri-la
```
tehhowch: Quote non-reserved keywords, e.g. "Shangri-La"
Also, the `l` needs to be `L` as `"Shangri-la"` is not the planet, `"Shangri-La"` is.

data/jobs.txt
```diff
@@ -1477,6 +1477,74 @@ mission "Core Pirate Attack"
...
+		planet Nimbus
+	npc evade
+		personality timid unconstrained uninterested
+		government Korath
```
tehhowch: "Korath" per style above

Personally I try to order the NPC specification
1. government
2. personality
3. system
4. fleet

but that is purely a personal style choice of mine.

data/jobs.txt
```diff
@@ -1477,6 +1477,74 @@ mission "Core Pirate Attack"
...
+		fleet "Korath Raid" 2
+	npc
+		personality heroic staying vindicitive
```
tehhowch: Incorrect spelling: `vindicitive` -> `vindictive`

(As personalities are all reserved keywords (the modder can't just drop in a random word here and have it do anything) they don't get quoted.)

endless-sky:
It's not fleshed out too much in the FW campaign, but my intention was that the Syndicate is trying to deal with the Korath invasions themselves, and to keep things as quiet as possible, so that the Syndicate gets exclusive access to any technology recovered from Korath ships. So, I'd rather save the Korath for a mid-level enemy to fight in the Syndicate campaign (as a variation from fighting pirates) rather than having ordinary non-plot defense missions against them.

That said, I think Korath defense missions would be fine *if* they were limited to players who have already completed the human story line.

czartrak:
What if i were to make it so you needed the main plot completed to offer

data/jobs.txt
```diff
@@ -1477,6 +1477,76 @@ mission "Core Pirate Attack"
...
+		
+mission "Raider Attack 1"
+	name "Defend <planet>"
+	description "Defeat an alien raid on <destination>."
+	minor
+	repeat
+	to offer
+		"combat rating" > 550
+		random < 7
+		has "main plot: completed"
```
tehhowch: `has "main plot completed"`

data/jobs.txt
```diff
@@ -1477,6 +1477,76 @@ mission "Core Pirate Attack"
...
+		
+mission "Raider Attack 2"
+	name "Defend <planet>"
+	description "Defeat an alien raid on <destination>."
+	minor
+	repeat
+	to offer
+		"combat rating" > 750
+		random < 5
+		has "main plot: completed"
```
tehhowch: `has "main plot completed"`

data/jobs.txt
```diff
@@ -1477,6 +1477,76 @@ mission "Core Pirate Attack"
...
+		"combat rating" > 550
+		random < 7
+		has "main plot: completed"
+	source
+		planet "Sundive"
+		planet "Shangri-La"
+		planet "Nimbus"
+	npc evade
+		government "Korath"
+		personality timid unconstrained uninterested
```
tehhowch: As with the above Core Pirate Attack missions, your opponent NPCs should also have `target harvets plunders`

data/jobs.txt
```diff
@@ -1477,6 +1477,76 @@ mission "Core Pirate Attack"
...
+		"combat rating" > 750
+		random < 5
+		has "main plot: completed"
+	source
+		planet "Sundive"
+		planet "Shangri-La"
+		planet "Nimbus"
+	npc evade
+		government "Korath"
+		personality timid unconstrained uninterested
```
tehhowch: same as above

czartrak:
@tehhowch thanks for your help. hopefully, it's all good and fixed now

data/jobs.txt
```diff
@@ -1477,6 +1477,76 @@ mission "Core Pirate Attack"
...
+		fleet "Large Syndicate"
+		fleet "Small Syndicate"
+	on offer
+		conversation
+			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by alien raiders! We need every combat-worthy ship to join in the defenses!" The authorities will probably pay you quite well if you assist them, but this could also be an easy way to get yourself killed.`
+			choice
+				"	(Stay here until the fight is over.)"
+					decline
+				"	(Join the defense fleet.)"
+			"The local Syndicate defense forces are preparing to repel the alien attack. You join them, and take off together..."
```
endless-sky: This paragraph should be indented (and the same below). And, for consistency, I usually put all lines of text in conversations in backticks (`` `...` ``) even if there are no quotation marks in a particular line.

tehhowch:
@endless-sky the conversations should be fixed now - they don't throw any warnings during load.

czartrak:
coolio

data/jobs.txt
```diff
@@ -1477,6 +1477,76 @@ mission "Core Pirate Attack"
...
+		personality target timid unconstrained uninterested harvests plunders
+		fleet "Korath Raid" 3
+		
+	npc
+		government "Syndicate"
+		personality heroic staying vindictive
+		fleet "Large Syndicate" 2
+	on offer
+		conversation
+			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by alien raiders! We need every combat-worthy ship to join in the defenses!" The authorities will probably pay you quite well if you assist them, but this could also be an easy way to get yourself killed.`
```
Fzzr: "Defenses" should be "defenses" here.

data/jobs.txt
```diff
@@ -1477,6 +1477,76 @@ mission "Core Pirate Attack"
...
+		personality target timid unconstrained uninterested harvests plunders
+		fleet "Korath Raid" 2
+	npc
+		government "Syndicate"
+		personality heroic staying vindictive
+		fleet "Large Syndicate"
+		fleet "Small Syndicate"
+	on offer
+		conversation
+			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by alien raiders! We need every combat-worthy ship to join in the defenses!" The authorities will probably pay you quite well if you assist them, but this could also be an easy way to get yourself killed.`
```
Fzzr: "Defenses" should be "defenses" here.

Bladewood: It would be, except that all the other planetary defense missions use "defenses".

Fzzr: That's why I recommended changing the rest in https://github.com/endless-sky/endless-sky/pull/3468.

czartrak:
k, now we're good?

czartrak:
noiceio

PR #3327, opened 2017-12-01
refer to the Plasma Turret properly

KiLEdEnNis:
so in the Plasma Turret testing, an option refers to it as a gun instead of a turret, which seems a little bit odd

### DIFF
--- a/data/free worlds start.txt
+++ b/data/free worlds start.txt
@@ -770,7 +770,7 @@ mission "FW Plasma Testing 1B"
 			
 			label improve
 			choice
-				`	"It seems like the only way to fit a gun this big into most ships is by leaving half the hardpoints empty. It's a bit impractical."`
+				`	"It seems like the only way to fit a turret this big into most ships is by leaving half the hardpoints empty. It's a bit impractical."`
 					goto hardpoints
 				`	"It worked well against a large ship, but I wonder how well the turrets could track fighters."`
 					goto fighters


Amazinite:
I'd say that in this case, gun is being used to mean weapon, not specifying a weapon that only fires in front of your ship.

tehhowch:
yeah, conversational usage, so gun and turret are interchangeable and this is a manner of preferred style (e.g. not likely to be changed per https://github.com/endless-sky/endless-sky/pull/3289#issuecomment-347363770 )

KiLEdEnNis:
I think this change makes a little more sense though, even though "gun" is correct in one way, "turret" is correct in both ways
so as usual: whatever MZ wants

Amazinite:
>\**I\** think this change makes a little more sense though,, even though "gun" is correct

>if it's not actually a grammar error and is just a matter of opinion, it's not worth tweaking it.

KiLEdEnNis:
> "turret" is correct in both ways

Amazinite:
>*I* think

>if it's... just a matter of opinion, it's not worth tweaking it.

endless-sky:
In naval terminology, it's referred to as a "gun" even if it's a turreted weapon. The shipyard uses the "gun" vs. "turret" labels to make it clear what slot things go in, but the general convention in the game is that people still use archaic naval technology when talking about ships.

KiLEdEnNis:
what if it's changed to "weapon" instead then?

PR #3331, opened 2017-12-02
Use Mk2 variants in Navy Surveillance fleets

tehhowch:


Bladewood:
Uh, then the Gunboats are useless for actual surveillance missions...

KiLEdEnNis:
I'm speaking from experience: FW borders get SPAMMED with Navy Surveillance, at least until the point where the 1st 3 worlds are joining the FW (which is where I am in my 1st playthrough after half a year+), so sitting back waiting for the tech isn't really an option
although that fix would be reducing the frequency a lot, which would make the weapon farming even easier like Bladewood said

still, I think that even the surveillance fleets should be updated to use the new tech, considering they're going into enemy territory, plundering Electron tech won't be that much easier and ya unlock it at the end of the story anyway

tehhowch:
>Uh, then the Gunboats are useless for actual surveillance missions....

Care to explain?

KiLEdEnNis:
Mark II Gunboats have no scanners

tehhowch:
Yep, just saw that.

Issue #3082, opened 2017-09-26
Feedback requested: compensation for "gifted" outfits


tehhowch:
If creating a mission that involves the transfer of game outfits, when the outfit is not known in advance or the quantity is generated procedurally in a manner to `cargo "random"`:
```
on complete
    outfit "random power outfit" -2 -5
    payment
    payment 50000
    dialog `The shipyard on <destination> thanks you for the delivery of <complete: gifts given> and pays you <payment>.`
```

1) Should the player be compensated for jumps with used cargo space?
Current jump-based payment value is `(# jumps + 1) * ( mission cargo size + 10 * mission passenger size ) * 150` and does not care about outfits being removed or added.
2) Should the player be compensated for the gift transfer `every time`, automatically?
What if the outfits are given to the player?
What if the outfits are removed from the player?
3) Should compensation for outfit value be at 100% value, 25% value, or some value in-between? Should that in-between value be constant? if not, how should it be conveyed? If it is random, should it be uniformly random (equally likely to be 25, 50, 75, or 100), or perturbed by some attribute of the outfit?
4) How should multiple gifting in a single mission be handled?
Say there is a stopover destination, and the player takes outfits from the origin to the stopover, where they are removed, and new outfits are given. These new outfits might be removed, or the player may get to keep them.
Should the player get "billed" for the new outfits?

What about for missions that do not involve random outfit generation - e.g. missions where both the outfit and the quantity are known in advance?

An example to consider:
The player is contracted to supply a distant outfitter with 3 Stack Cores. If the player buys the 3 stack cores, they spend 13.5M credits and gain 0 net worth. If the player plunders the 3 Stack Cores, they spend 0 credits and gain 3.375M credits in net worth. What is a fair price for the player to be paid when completing the mission?

<details><summary>my answers</summary>

1) I think probably only if the outfits are being removed. If outfits are being given to the player - unless they are being removed later - this is not really an inconvenience worth recompense.
2) Probably not every time - adding a new keyword at the Mission or MissionAction level will give more control to the creator, and without the keyword the existing payment code is used. This keeps backwards compatibility, but could be seen as keyword creep / cruft.
[...]

PR #3334, opened 2017-12-02
WIP: Allow missions to specify random outfits & random outfit quantities

tehhowch:
Refs #3082 
This PR extends support for random outfit picking, and random outfit quantity picking, in missions / jobs, to replicate the available features for mission cargo.

Also added to support the above
 - Category restrictions for the randomly selected outfit:
   - "random outfit" (e.g. purchased or looted from a ship)
   - "random minable" (located within a minable asteroid)
   - Use of an Outfit category, e.g. Power: "random power outfit", "random power minable"
 - Missions involving delivery of outfits to a destination (whether or not the outfits are given to the player) compensates the player for the used cargo space, just as though the outfits were "mission cargo"
 - Text substitutions in the form of `<trigger: _____>` where trigger is any of `offer`, `complete`, `accept`, etc.
   - These substitutions allow the mission writer to relay the requested outfits, cargo size, etc to the player in the appropriate place, e.g. tell the player about a stopover transfer in the stopover dialog and mention only the gifts transferred during the stopover, or tell the player on accepting which outfits will be picked up at the stopover and which outfits will be removed at the destination, without requiring them to be the same outfits.

A usage example:
```
on complete
    outfit "random power outfit" -1 -5
    payment 16000 150
    dialog `The store manager pays you <payment> and thanks you for delivering <complete: gifts given> to address the equipment shortage.`
```

TODO
 - [ ] Implement an optional value-based payment (#3082 ) that has the player either pay or be paid for gifted outfits
 - [ ] Check over random weights to ensure they introduce sensible bias for randomly picked outfits/minables
 - [ ] Perhaps implement space-checks to prevent jobs that are too large for the player to complete from being offered (non-jobs should probably be offered even if too large)
       - Always offering is a possibility because the player can upgrade their fleet in order to accommodate the requested job size
 - [ ] Clean up implementation (removing unused added functions / includes / development commentary)
 - [ ] Address community issues & feedback

### DIFF
--- a/source/Minable.cpp
+++ b/source/Minable.cpp
@@ -183,7 +183,7 @@ void Minable::TakeDamage(const Projectile &projectile)
 }
 
 
-	
+
 // Determine what flotsam this asteroid will create.
 const map<const Outfit *, int> &Minable::Payload() const
 {
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -1052,14 +1052,15 @@ Mission Mission::Instantiate(const PlayerInfo &player) const
 	for(const NPC &npc : npcs)
 		result.npcs.push_back(npc.Instantiate(subs, player.GetSystem(), result.destination->GetSystem()));
 	
-	// Instantiate the actions. The "complete" action is always first so that
-	// the "<payment>" substitution can be filled in.
+	// Instantiate the actions. The "complete" action is always first so that the "<payment>"
+	// substitution can be filled in. This will also instantiate any randomly chosen gifts or
+	// gift quantities.
 	for(const auto &it : actions)
-		result.actions[it.first] = it.second.Instantiate(subs, player.GetSystem(), jumps, payload);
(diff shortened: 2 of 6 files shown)


endless-sky:
I'm really confused about why this would be useful. Is this for missions that would require you to go out and acquire a random outfit and deliver it to a planet? That seems like a very limited use case.

tehhowch:
Indeed that is the intended use. It would have the most use in supporting mining jobs of variable quantity. Delivery jobs of outfits would also be possible.
The minimum addition I'd be happy with would be the random quantity generation for a specific outfit. A goal would then be random minables payload selection, with everything else as a bonus. To me these extras are a logical extrapolation of what is done for cargo also being supported for outfits.

endless-sky:
Except, for cargo there are never missions that say, "Go buy 10 tons of medical supplies and bring them here." So, they're really completely different: mission cargo can have random names, but it's always given to you by the mission itself.

Amazinite:

PR #3337, opened 2017-12-02
Added some assistance for "Wanderers: Sestor Attack"

warp-core:
Added 1 512, 2 128s and 4 32s 5 jumps away and a tough Derecho and 2 Autumn Leaves 6 jumps away to provide latent assistance during the mission for those who have more trouble with it.
Inspired by a thread on the Steam forums: http://steamcommunity.com/app/404410/discussions/0/1489992713700285274/

### DIFF
--- a/data/wanderers middle.txt
+++ b/data/wanderers middle.txt
@@ -1643,6 +1643,25 @@ mission "Wanderers: Sestor Attack"
 				"Hurricane (Tough)"
 				"Derecho"
 				"Tempest (Missile)" 2
+	npc
+		government "Kor Mereti"
+		personality heroic
+		system "Korsmanath"
+		fleet
+			names "kor mereti"
+			variant
+				"Model 512"
+				"Model 128" 2
+				"Model 32" 4
+	npc
+		government "Wanderer"
+		personality heroic
+		system "Aya'k'k"
+		fleet
+			names "wanderer"
+			variant
+				"Derecho (Tough)"
+				"Autumn Leaf" 2
 
 
 
--- a/data/wanderers middle.txt
+++ b/data/wanderers middle.txt
@@ -1650,7 +1650,7 @@ mission "Wanderers: Sestor Attack"
 		fleet
 			names "kor mereti"
 			variant
-				"Model 512"
+				"Model 256"
 				"Model 128" 2
 				"Model 32" 4
 	npc


tehhowch:
This seems like a rather large amount of assistance.

SirStigII:
A 512 is probably more than needed, a 256 would be better probably, as the player should be able to take out a lot of the fighters before they jump in.

SirStigII:
I'll try and test it tonight

SirStigII:
Well just tested it with a starling, the reinforments don't have much of a problem, I ran it twice, last time not interfering at all and the Derecho and one 34 were lost from the reinforcments, so I would say it is still too strong. Maybe just try without the 256? That would be a fairly small reinforment fleet that seems in line with a "wellness check".

SirStigII:
image
This is with adding a bit of flamethrower supprt. The armor DPS is a huge boost for the wanderers, I'm phsyched for the moonbeam!

SirStigII:
Tested without the 256 and it was about perfect.

tehhowch:
I'm interested in what happens (without reinforcements) if:
1) if some of the Korath ships that are `personality entering` are moved 2-3 systems away
2) if the player's allies are giving `timid` instead of `heroic` (at least the ones that are `staying`).
3) The mission fleets are reorganized

Option 1) should achieve a similar effect as giving the player reinforcements here does, by letting the initial combat group be a little smaller.
Option 2) should make the player's allies group up more closely and coordinate with the player more, rather than dashing off into the distance to challenge a group of automata solo.
Option 3) stems from #3052, which altered intricacies of how NPCs behave with respect to their immediate allies: within this mission I see "mini-fleets" defined when they do not need to be, which makes me think they were originally written this way to behave as mini-fleets (see https://github.com/endless-sky/endless-sky/issues/2995#issuecomment-326773539 ). Now that it is possible to make them behave as mini-fleets, it would not be out of the question to do so. To illustrate what I mean:
as written:
```
 	npc
 		government "Wanderer"
 		personality staying heroic
 		fleet
 			names "wanderer"
 			variant
 				"Hurricane (Turret)"
 				"Tempest"
 				"Derecho (Turret)"
 		fleet
 			names "wanderer"
 			variant
 				"Tempest (Heavy)"
 				"Strong Wind"
 		fleet
 			names "wanderer"
 			variant
 				"Autumn Leaf"
 				"Summer Leaf"
```
Post modification:
```
 	npc
 		government "Wanderer"
 		personality staying heroic
 		fleet
 			names "wanderer"
 			variant
 				"Hurricane (Turret)"
 				"Tempest"
 				"Derecho (Turret)"
 	npc
 		government "Wanderer"
 		personality staying heroic
 		fleet
 			names "wanderer"
 			variant
 				"Tempest (Heavy)"
 				"Strong Wind"
 	npc
 		government "Wanderer"
 		personality staying heroic
 		fleet
 			names "wanderer"
 			variant
 				"Autumn Leaf"
 				"Summer Leaf"
```

SirStigII:
Those modifications might work too, I'm glad to test if you want to do a pr

Issue #3326, opened 2017-12-01
Outfitter/Shipyard side-panel


Nescio0:
It would be nice if the ship icons in the outfitter and shipyard side panel on the right would become automatically smaller if you have a large number of ships. E.g. if you have more than fifty, display ten ships per row, instead of four.
Attached a screenshot; because of my 100 deep rivers I can no longer select a ship and simultaneously view its stats:
![screenshot from 2017-12-01 22-13-52](https://user-images.githubusercontent.com/21634324/33504085-5e9c87d6-d6e6-11e7-950e-6d9dadf5585d.png)

KiLEdEnNis:
I don't think putting 10 ships in 1 row is gonna make it better though; it'll get very cluttered & messy and a lot of people might not want this
my only problem is that the ship stats are in the same big column as all the ships, so ya HAVE to scroll to see the ship stats: they should be separated, so no matter how many ships ya have, ya can always easily see the stats of one
(maybe later on we could also add some button on the ship list that enlarges it (or even opens a small menu like the I screen), so ya can see a lot more ships at once, which basically gives ya what ya are asking for as well)

jafdy:
I suppose you could put 6 in a row, but any more than that and it will likely start to get on people nerves. Also, as the main focus is on single ship or small fleets, things for very large fleets are likely to be low priority.

KiLEdEnNis:
I wouldn't really go for 6 ships per row either, it just makes the problem start at a higher amount
I think my solutions should make bigger fleet players happy, and it doesn't really change anything for small fleet players, and it works no matter how big the fleet is

tehhowch:
Relevant: https://github.com/endless-sky/endless-sky/issues/500

There's no reason to consider only solutions which preserve the current layout. Those in the community with "art skills" can consider making single-image mockups of the layout to see which poll better, and when some good ideas are found, recruit someone who can implement the code.

Amazinite:
How about something like this? Your ships are listed on the right side in rows of 3. To the left of that is where the stats display.
![example](https://user-images.githubusercontent.com/17688683/33512837-a626ddf4-d704-11e7-9c41-f76f148ec6b8.png)

The Buy | Sell | Leave buttons are a little close together and create a substantial amount of empty space in that corner, but otherwise I think something like this could work out.

Nescio0:
[...]

Issue #500, opened 2015-12-23
A simple(ish?) fix to make the outfitter screen much easier to read


DingusShingleton:
Simply move the details of the item you are inspecting to the right side next to your ship description, and leave the outfit icons where they are, rather than having them fly around as your selection changes. This also makes it WAY easier to compare outfits by alternating between them, as the relevant numbers stay in the same place.

 Also, increase the brightness of the outfit you have selected to make it easier to pick out at a glance. This will also differentiate it visually from an outfit you only have installed.

You lose a column or two of outfits, but the gains in visual feng shui far outweigh the loss IMO

stealth edit: Also for what it's worth, listing thrust and steering in two different sections would be a huge improvement to alternating thrust and steering units.

 To wit: 

![outfitter](https://cloud.githubusercontent.com/assets/16414093/11977650/ebce65a2-a984-11e5-82fb-59afef168982.jpg)

brick96:
Yes please, and comparisons im the outfitter too

endless-sky:
The outfitter is designed to be usable even on a relatively small screen, e.g. 1024x768 or 1280x720, less than a third of the dimensions of your screen. At that resolution, you might not even be able to fit three columns of outfits, and in addition, the data for the selected outfit would not all fit on the screen at once.

DingusShingleton:
My screen is only 15", surely that's not that huge a screen. It is 4k, but it isn't some giant 27" behemoth... or bacrian for that matter.

sanjayatio:
Mine is 14" at 1366*768 and I put the windows taskbar on the left side (rather than the bottom default).
With this configuration, currently I can see 5 columns --> implementing your idea will bring it down to 3 columns.
Further:
- as someone who carries a laptop around on daily basis, I actually prefer smaller laptop.
- I don't like to scroll too much.

I agree with you though, the dynamic flow is kind of annoying.
My idea would be tabs - alternating description on the right side between ours & shop's.

==============================================

> ... as the relevant numbers stay in the same place.

Not true, with different length of description and different kinds of attributes an outfit has, the scrolling wheel will need some oiling; however
> ... the data for the selected outfit would not all fit on the screen at once.

ignoring aesthetic factor and putting the description at the bottom would minimize the issue, imho.

DingusShingleton:
[...]

PR #3342, opened 2017-12-03
Allow a minimum zoom of 50%

Nescio0:
Currently the minimum allowed main zoom factor is 100%; this patch allows to reduce it down to 50%, for better support of low resolution screens.

E.g. a 3840×2160 screen at 150% zoom is equivalent to 1280×720 at 50% zoom. At this setting outfitter and shipyard icons are already unnecessarily large. 1280×720 at 100% zoom would be equivalent to 3840×2160 screen at 300% zoom, which seems ridiculous.

### DIFF
--- a/source/PreferencesPanel.cpp
+++ b/source/PreferencesPanel.cpp
@@ -40,7 +40,7 @@ using namespace std;
 namespace {
 	// Settings that require special handling.
 	const string ZOOM_FACTOR = "Main zoom factor";
-	const int ZOOM_FACTOR_MIN = 100;
+	const int ZOOM_FACTOR_MIN = 50;
 	const int ZOOM_FACTOR_MAX = 200;
 	const int ZOOM_FACTOR_INCREMENT = 10;
 	const string VIEW_ZOOM_FACTOR = "View zoom factor";
--- a/source/Screen.cpp
+++ b/source/Screen.cpp
@@ -45,7 +45,7 @@ int Screen::Zoom()
 
 void Screen::SetZoom(int percent)
 {
-	ZOOM = max(100, min(200, percent));
+	ZOOM = max(50, min(200, percent));
 	WIDTH = RAW_WIDTH * 100 / ZOOM;
 	HEIGHT = RAW_HEIGHT * 100 / ZOOM;
 }


jafdy:
Has this been tested? I can't help but think that there might problems when trying to scale down UI sprites.

tehhowch:
I agree, you're going to need to offer at least some UI evidence that this does not make art assets look bad.

Nescio0:
Here two screenshots at 50% zoom (at 3840 screen, mind). The text is uncomfortably small, but other than that, the ship and outfit sprites do not look any worse than at higher zooms.

screenshot from 2017-12-03 16-47-07
screenshot from 2017-12-03 16-47-48

tehhowch:
You mentioned low resolution screens as the reasoning behind this PR: why show the change on a massive, high resolution screen?

Nescio0:
Because that's the only screen I have available at the moment. However, 50% zoom at 1280×720 would look like 150% at 3840×2160, which is what I comfortably use. Support for low resolution screens was mentioned at #3326 and #500, where it was pointed out only several outfitter columns could be displayed. At 50% at 1280×720 you can view twelve outfitter items per row, instead of five at 100%.

tehhowch:
You get what I'm saying though, right? It's impossible for us to make judgments based on your fullscreen use because no sane person would combine that real estate with that zoom level:

You can set your Endless Sky window size to the appropriate low resolution in the preferences.txt file, set the zoom to 50%, and then share the active window screenshot for a good approximation of the behavior you are changing.

Nescio0:
> You can set your Endless Sky window size to the appropriate low resolution in the preferences.txt file, set the zoom to 50%, and then share the active window screenshot for a good approximation of the behavior you are changing.

Thanks! Here you go:

screenshot from 2017-12-03 18-42-35
screenshot from 2017-12-03 18-42-46

The texts look poor, the images fine.

tehhowch:
The impact on text readability is going to be a sticking point.

Amazinite:
Yeah, I don't think this'll fly.
screenshot_1

Nescio0:
Ideally it would be nice if text could be zoomed independently from the main zoom, but that would be more work to implement.

Instead of allowing to zoom from 50% to 200%, an alternative would be to halve the sizes of UI images and allow zoom from 100% to 400%. This would preserve (and actually improve) text readability and simultaneously allow more items to be displayed in outfitter and shipyard.

tehhowch:
Close this pull and do one of those two.

endless-sky:
This really doesn't make sense to me. If I were on a small screen, I would still view being able to read the game's text as a much higher priority than being able to view a few extra rows of outfits in the shop panels.

PR #3344, opened 2017-12-03
prevented safe passage from ending before mission

mootootwo:
moved "fw safe passage ends" event out one day in two cases, to prevent it from ending on the same day as the mission deadline, before the mission has actually been failed.

### DIFF
--- a/data/free worlds middle.txt
+++ b/data/free worlds middle.txt
@@ -82,7 +82,7 @@ mission "FW Diplomacy 1B"
 	
 	on accept
 		event "fw safe passage starts"
-		event "fw safe passage ends" 24
+		event "fw safe passage ends" 25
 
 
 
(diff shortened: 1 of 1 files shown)


tehhowch:
What if the player takes one more day to jaunt about? This change won't have done anything then.

endless-sky:
I'm not sure if this is necessary. The "deadline" in this case is just an indicator of when you have to be safely back in Free space; it's not unreasonable for the Republic to turn hostile again on the deadline, instead of on the next day.

mootootwo:
Indeed.  The change really only prevents a single situation where the destination planet of the first mission in a two-mission chain turns hostile on the last possible day -- it is a pretty narrow case; This change only serves to make the free-passage sync up with the deadline of the mission.

The event can still trigger at any time during the second mission in the chain, before the player has exited republic space.

A related question could be:  If the player should be allowed to complete the first mission in the chain after free passage ends, does that mission need a deadline?

PR #3345, opened 2017-12-03
converted reputation adjustments to minimums

mootootwo:
per @tehhowch 's comments on #3322

### DIFF
--- a/data/events.txt
+++ b/data/events.txt
@@ -1076,7 +1076,7 @@ event "fw safe passage starts"
 
 
 event "fw safe passage ends"
-	"reputation: Republic" = -1000
+	"reputation: Republic" <?= -1000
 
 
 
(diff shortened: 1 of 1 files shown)


mootootwo:
Should the converse of this also be implemented?  

for example, when reputation gains are implemented (during end of war, ceasefire, and safe passage) should they similarly be made maximums to protect any reputation gains the player may have made while hunting pirates during the war?

Or, should the reputation changes be additive, like a flat + or -1000, so that if the player has been particularly bloody during the war their crimes may not be completely forgiven?  Such an implementation may actually prevent safe passage events from appearing to work, so extra exposition may be required.

endless-sky:
In general, using minimums and maximums makes the most sense, because usually the intention is to make sure the Republic or whoever is friendly or hostile, whichever the mission requires, without erasing the player's reputation if it's better or worse already. (The minimum and maximum operators didn't exist back when I started creating the FW campaign, which is why they're often not used.)

Of course, it might be better to just have a separate way to make a faction hostile or friendly, *overriding* the reputation rather than altering the reputation. Otherwise, a max followed by a min (e.g. a ceasefire that ends after some time) wipes out the player's previous reputation anyway.

mootootwo:
Completely agree, a different faction hostility control that was independent from reputation would be ideal.

PR #3347, opened 2017-12-03
Prisoner transport missions and the Brig

Pointedstick:
I've added a new style of job: in-person bounty hunting/prisoner transportation. These jobs are unlocked by owning a new outfit: the Brig, available on Pirate worlds and throughout the Galactic South.

brig

Pirate slavery jobs also now require a Brig.

Other planned additions in the same vein include the Luxury Accommodations outfit that unlock high-class tourist jobs--but only if you don't also have a Brig! The point is to make players choose what kind of captain they really are. Will you be a scoundrel who profits from human misery and lives dangerously? Will you cater to rich elitists instead? Or will you play it safe and stick to more normal jobs?

Why? @endless-sky [once said:](https://github.com/endless-sky/endless-sky/issues/1070#issuecomment-219245709)

> The only reason to add new features that make the game more complicated is if they would make the player's behavior more varied or interesting.

I think this counts. By purchasing a Brig, players are simultaneously forced to confront something about their own behavior, and also rewarded for it by unlocking new game content--at a potential cost to their humanity...

Errata: the boarding missions do work properly, but are affected by https://github.com/endless-sky/endless-sky/issues/3129.

I would have added much more bounty hunting content, but was stymied by the lack of https://github.com/endless-sky/endless-sky/issues/3340, https://github.com/endless-sky/endless-sky/issues/1763 and https://github.com/endless-sky/endless-sky/issues/3129.

PR #2496, opened 2017-05-07
require <outfit> can require a specific number, or specifically zero

Elyssaen:
**Update:** lot of great discussion below, but for newcomers: most of it's, I think, talking about other ways this could be extended (in a different PR). As of 10/05/17 I'm still suggesting this PR is merge-worthy and merge-ready, as a starting point if nothing else.

_____________________

With this PR, instead of only being able to write (in an mission's `on` clause)

`require <outfit>`

to always require at least one copy, you can now also write

`require <outfit> 3`
`require <outfit> 0`

respectively requiring at _least_ 3, or _exactly_ 0. (0 is the only one that will look for an exact match: any other number is happy with at least that many.) You can still use the existing syntax which will require at least one copy.

`outfit <outfit> [<number>]` is completely unchanged: it will still add/remove according to the number. `require` is for if you want to check the player has a certain number, but not change it.

PR #2221, opened 2017-03-05
Added human spaceport hint for Sestor fleet

MessyMix:
I, and apparently several others ( http://steamcommunity.com/home#scrollTop=0 ) have found difficulty in finding the Sestor when searching for them initially in human space. I just glossed over the actual system, because it is south of all the other pirate systems, which I checked. I thought that Farpoint would be the location of the Sestor fleet, so I was surprised when they did not turn up there. I actually had to resort to checking the save files to find the location of the fleet.

So I made a solution (took me a while) that involves talking to a dude who narrowly escaped the Sestor as two 53's jumped in while scouting human space. It's written in the style of the rest of the game, gives the player ample choices (I refrained from deciding actions _for_ the player; I gave him/her multiple choices), and nudges the player back in the direction of the pirate colonies. It will be super helpful for players like me who assumed that they had spread across the rest of human space beyond the pirate colonies of the northwest. This extra mission will just hint the player back in that direction.

What d'yall think? Feedback and everything welcome and appreciated. :D

PR #3359, opened 2017-12-04
Luxury Accommodations and high-class passenger jobs

Pointedstick:
To go along with [the Brig](https://github.com/endless-sky/endless-sky/pull/3347), here's the light-side equivalent: Luxury Accommodations.

luxury accommodations

This one requires a bigger investment than the Brig: 10 tons of space, 2 additional crewmen (a cook and a maid, presumably), and 250,000 credits. But the rewards are large: all of the game's passenger jobs that involve rich tourists, executives, wealthy students, and corporate professionals are now gated behind owning one of these outfits--and they now pay 33% more, too (the standard multiplier of 150 increases to 200, for most of them).

Luxury Accommodations are sold not just in Human space, but all throughout the galaxy--anywhere you can potentially take wealthy passenger transport jobs, you'll be able to buy Luxury Accommodations. And I was able to stuff one into the Star Queen by downgrading its battery, which was overkill anyway. I also added some Luxury variants of fancy or passenger-focused ships, and added them to fleets plying the North and the Core.

None of this should impact passengers just starting out since the wealthy passenger jobs were almost entirely available only to ships with larger capacities. Nor should it impact players who don't care much for role-playing; they're mostly making their money by tooling around with 100,000 tons of cargo space in a huge fleet, and don't bother with the job board anyway. But for people who do enjoy role-playing their captain, and heavily use the job board for flavor and role-playing as much as for money, this outfit, along with the Brig, opens up some interesting possibilities regarding what kind of captain you want to be.

This PR does not include the change to disallow these jobs if the player also has a Brig because the Brig isn't merged yet; that can be done later.

HiDPI PR: https://github.com/endless-sky/endless-sky-high-dpi/pull/36

Blender and GIMP assets: [Luxury Accommodations assets.zip](https://github.com/endless-sky/endless-sky/files/1525638/Luxury.Accommodations.assets.zip)

If these kinds of "trinket outfits" (see https://github.com/endless-sky/endless-sky/issues/2908) prove fun and popular, I can make a bunch more--including for aliens.


data/deep jobs.txt
```diff
@@ -150,18 +150,19 @@ mission "Transport scientists"
 	repeat
 	to offer
 		random < 15
+	on offer
```
tehhowch: This one I don't believe, for the sole reason of the scientists having government funding. Per diem limits exist now for reimbursable expenses, and I don't have any hope they'll disappear in the next 1000 years :rofl:

data/outfits.txt
```diff
@@ -301,6 +301,16 @@ outfit "Small Bunk Room"
...
 
+outfit "Luxury Accommodations"
+	category "Systems"
+	cost 250000
+	thumbnail "outfit/luxury accommodations"
+	"mass" 10
+	"outfit space" -10
+	"required crew" 2
+	"unplunderable" 1
+	description "Most deep space voyages involve a certain amount of passenger discomfort. With the addition of a few extra fresh water tanks, private bathrooms, wooden fixtures, and some hospitality staff, any ship can be made attractive to a significantly higher--and better paying--class of passenger."
```
tehhowch: `higher - and better paying - class`

data/paradise world jobs.txt
```diff
@@ -160,13 +163,15 @@ mission "Transport partiers [2]"
 	cargo "snacks and school supplies" 2 4
 	to offer
 		random < 10
+	on offer
```
tehhowch: Students probably don't care about the quality of their accommodations. Their party money is better off spent on... :beers: or newfangled space drugs. Even if the parents are bankrolling, that's just more money they could spend on party supplies.

Pointedstick: These are Paradise Planet students, though! I figure they're used to luxurious accommodations (I mean heck, their mothers are sending them a literal ton of brownie mix in school), and it also becomes a statement about the culture of the Paradise Planets.

data/outfits.txt
Pointedstick: I think the em-dash is grammatically correct here?

data/deep jobs.txt
Pointedstick: Yeah I wavered on this one too, and eventually decided that the Wanderer scientists wouldn't mind normal digs. Deep folks aren't supposed to be real hoity-toity, so I can revert this.

Pointedstick:
This doesn't meaningfully increase the amount of time it takes a player without Luxury Accommodations to get access to Coalition ships and outfits. There will still be a huge number of jobs in Coalition space that bump "coalition jobs" and don't require Luxury Accommodations.

data/outfits.txt
MessyMix: https://github.com/endless-sky/endless-sky/pull/2221#discussion_r109287047

It's happened before -- MZ isn't a fan of double dashes.

(I'm kind of hoping he reads that and chuckles to himself.)

Bladewood: Well, the difference here is that there aren't spaces on each end of the em dashes.

I can kind of understand not liking the double dashes version; there's just something very off-putting about it.

I feel like Michael himself used it somewhere, though. I don't remember where, and a search brings up nothing.

MessyMix: https://github.com/endless-sky/endless-sky/pull/2395/files

You're thinking of the Kestrel description, which I changed based on his comment and got merged.

Bladewood: Ah, I was just looking there. I thought something was wrong.

Amazinite:
Maybe not every tourist job should require luxury accommodations. Some people might be going on a budget vacation, you know. :)

tehhowch:
How will you be disallowing these missions from offering when the Brig is also installed on the player's flagship?

KiLEdEnNis:
isn't it just adding a random value in said missions & if the random value is met the missions do not get offered? or am I mistaken

Pointedstick:
> How will you be disallowing these missions from offering when the Brig is also installed on the player's flagship?

```
	on offer
		require "Luxury Accommodations"
		require not "Brig"
```

Which did seem to work in my testing.

tehhowch:
It gets fun because `to offer` is a ConditionSet, handling only things found in your save's `conditions` section, while outfit presence is handled in the `on offer` MissionAction.
My guess is the needed functionality is not currently available, or at least is pretty convoluted. As is seen in both this PR and #3347, the respective outfit is required to be installed or in cargo by `outfit "outfit name"` or `require "outfit"`.
The current implementation of require:
https://github.com/endless-sky/endless-sky/blob/224e62f563a55fc57e20b2e851a945836378bd0f/source/MissionAction.cpp#L150-L151

So to prevent these from offering, I believe they need
```
to fail
    outfit "Brig" -1
    outfit "Brig" 1
```
Which tells the game if it can take away a Brig from the player, then this mission should fail.

Pointedstick:
I'm okay with delaying the "Luxury Accommodations and Brig" quirk until support is enabled in the proper way; it's not a part of this PR anyway. It would be really great to get some of those PRs you mention merged.

Amazinite:
Idea (partially to stop the complaining about the Star Queen): 
Implement a mechanic where jobs can take up a special type of passenger or cargo space (likely in addition to the normal space). In this case, the Luxury Accommodations would have a luxury passenger space attribute on them (and when a job is taken for say 5 passengers, 5 luxury passenger space is taken up and 5 bunks are taken up; in total this isn't 10 bunks being taken up, it's just 5 bunks being taken up that were designated as luxury). What this would allow us to do is drastically lower the required crew of the Star Queen and instead fill it with multiple Luxury Accommodations. Since these outfits have a crew requirement to them, a stock Star Queen could have the same required crew as currently, but removing the Luxury Accommodations causes the Star Queen to have far less required crew at the cost of not being able to do luxury jobs.

Reasoning: it has been said that the reason for the high required crew of the Star Queen is because a lot of those required crew are servants, but why be forced to keep them if you aren't using the Star Queen for luxury jobs / when you are using the Star Queen, it doesn't give you any extra credits per passenger?

mootootwo:
I've had that on my mind for a while, @amazinite but the implementation is non trivial.

I took a pass through the code looking for all the places 'bunks' are mentioned, and considered either creating a duplicate 'luxbunks', or making bunks into a type with subtypes.  My coding skills are at least 20 years out of date, but I might be able to gin something up.

Then the interfaces need to be changed, and different classes of passengers and crew need to be given berth priorities such that lux passengers get first pass at lux bunks before they are given to standard passengers.

And of course ships like SQ and Fliver would get their inbuilt bunks converted to lux...

Wrzlprnft:
One thing i would have to add to amazinites ifea is that we need some sort of outfit expansion penalty for non-native luxuryships: it doesn't matter how much luxury accommodations one cramps into a hauler, it will never get the aesthetics of a Blackbird or a fliver, never thr sightseeing potential of a SQ. Besides maybe exklusiveness, that only would offer a luxury way of transport. which is part of we need i guess, but the star queen should have a certain... star factor.

KiLEdEnNis:
you're basically saying non-transports should have less benefits from these luxury outfits to prevent a big freighter w/ 20 luxury outfits? good idea
maybe it should be done through some certain multiplier a Transport class ship gets? or just the more ya get the less they will do

mootootwo:
If there was a separate bunk type, that inherent 'luxury ship' factor could be handled based on the cost of the required crew.  

Luxury Bunk outfits might require 1 crew per 2 or 4 luxury bunks, whereas SQ and Fliver would have better crew requirement ratios based upon their hull statistics and thus would operate at better profit margins.

Of course, you then open a can of worms where players will only want to have to hire the extra crew (stewards, chefs, etc) when they have the requisite number of luxury passengers onboard...

Wrzlprnft:
i was worried that regulating that via required crew can mess up other things the crew complement is connected to, but i haven'treally thought through. after all,the SQ is still as fast maneuverable as a broken toaster.

i think we can put that can of worms next to the can of worms containing engineers and soldiers, which again stands next to the can of squirrels and lizards and other aliens...

Pointedstick:
I would love to be able to differentiate capacities like that: 12 normal bunks, 5 luxury staterooms, 2 brig cells, etc. But in the absence of the necessary code changes, I opted to make the Brig and Luxury Accommodations apply to your whole passenger compartment for simplicity's sake.

endless-sky:
It really doesn't make sense to me to shift so many of the existing jobs to require this outfit. In particular, tourists are a major revenue source for a starting player in a shuttle.

Also, "luxury" accommodations probably shouldn't feature a metal-frame bunk bed.

Pointedstick:
Fair enough about the bed! I'll make a nicer one.

As for the jobs that require this, I'm just thinking for flavor reasons, does it really make sense that tourists (who are trying to enjoy their vacations or retirement) would tolerate the same kind of cramped and uncomfortable conditions as mill workers and farmers?

If I make new "super rich tourist" or "cheap dive tourist" jobs, that will further exacerbate the problem of the job board becoming oversaturated.

Pointedstick:
@endless-sky I'm happy to make changes regarding which jobs should be gated behind this, but I'd like some direction regarding what would be acceptable. We don't currently have rich tourist vs poor tourist jobs in human space, which is why I decided to apply it to all of them. Here are my worries regarding adding more:

1. There will be overall too many tourist jobs, and there are already quite a lot
2. It won't be as obvious that you're missing out on the tourist jobs when you don't have Luxury Accommodations since some will already be available
3. If getting Luxury Accommodations doesn't actually unlock that many jobs, it won't seem like as much of a reward for the sacrifice of space and extra crew

Pointedstick:
I've updated this to add (Luxury) variants for the Arrow, Bounder, and Flivver, gave the Blackbird Luxury Accommodations by default to match its description, and added (Luxury) variants to some of the Core and Northern merchant fleets.

Pointedstick:
I've altered the PR to only require Luxury Accommodations for jobs in Human and Hai space that were already previously marked in some way as transporting wealthy passengers. That should make the effect a bit less dramatic.

Tadrix:
How would this work for Star Queens? Their description says; "Every cabin has its own mini-bar, and the bathrooms have marble tile and showers that use actual hot water instead of chemicals or sonic vibrations."
So Luxury accommodations seem to be built in them by definition (and probably shouldn't be allowed to be taken out).

Also, I imagine such changes wouldn't take ship's outfit space, but rather decrease available bunks (because of less bunks per room and/or larger rooms), perhaps by half.

Pointedstick:
We don't have support for something like that yet, so in the meantime, I just gave the Star Queen the Luxury Accommodations outfit by default. You can always take them out (and perhaps replace with with a Brig) if you want; I don't see any problem with that.

tehhowch:
You can now prevent these from offering with `require "Brig" 0` in the on offer mission action, since #2496 has merged.

You cannot, however determine if the player obtains a brig after taking on these jobs, since `to fail` is a pure ConditionSet.
(Though you could prevent completion in that case, with a `require "Brig" 0` in the `on complete` mission action, but I imagine the player may be confused as to why the mission won't complete).

Pointedstick:
Right, and I'll do that if and when the Brig is merged.

Pointedstick:
Ping?

Pointedstick:
Ping!

data/hai jobs.txt
```diff
@@ -141,14 +143,15 @@ mission "Wealthy Hai [2]"
 	to offer
 		random < 30
 	on offer
+		require "Luxury Accommodations"
 		require "Quantum Keystone"
```
Amazinite: If the Wealthy Hai jobs would now require both a Quantum Keystone and luxury accommodations, perhaps also add a job or jobs that require only one or the other (e.g. a job for only keystones and a job for only accommodations in Hai space). Otherwise, it'll be more difficult for a player to make the connection between having a keystone and having these jobs offer.

data/ships.txt
```diff
@@ -493,6 +493,7 @@ ship "Blackbird"
 		"LP072a Battery Pack"
 		"D23-QP Shield Generator"
 		"Small Radar Jammer"
+		"Luxury Accommodations"
```
Amazinite: Glad to see that the Navy prisoners get a comfy transport back into Republic space during FW Prisoner Parole 1 and 2. :)

data/hai jobs.txt
Zitchas: "These superstitious Hai are looking for transport to <planet>, but believe that in order for their journey to be successful they must travel exclusively on a ship with the proper crystal on it."

data/ships.txt
Pointedstick: Haha yes, it had occurred to me on my last playthrough that maybe you should need a Brig for that mission. But then again the Navy officers are portrayed as so honorable that they're prisoners only in the loosest sense of the word. :)

data/hai jobs.txt
Pointedstick: Hmm, that's true. I'll see what I can do.

Pointedstick: I've adjusted the description of the Quantum Keystone instead, to hint at it being a piece of puzzle to unlocking the wealthy Hai passenger jobs.

Amazinite: I think that's good for now, but I would rather have Hai jobs that only require one of each trinket at some point.

Pointedstick:
Yay, glad you're enjoying it!

Have you found that there are enough of the jobs gated on this outfit to make it financially worth it? I'm doing one such playthrough myself and I find myself wanting wealthy passenger missions to be offered more frequently. I was toying with increasing the offer frequency for those missions, since you already need Luxury Accommodations to even have a chance of seeing them at all. Or maybe increase the payouts a little bit? Or no changes at all lol.

DingusShingleton:
I found that they offered enough that I appreciated them, but not often enough that I could rely on them each time I was doing a run. So if they weren't around, I would plan another block of drop offs and try again at my next destination. But it took my a while to realize that there were gated missions that I wasn't looking for, hence my earlier comment. 

Overall, I think the outfit has payed for itself many times over. But I find that the real issue is that the mission computer does a pretty woeful job of displaying information, and were it to more clearly indicate "this is a luxury transport job" it would be much easier to gauge and thus appreciate the value of the outfit.

DingusShingleton:
For example, if the mission list simply had a column that displayed payout, that alone would highlight the luxury jobs' lucrative nature

Issue #3004, opened 2017-09-05
Near death escort behavior


flaviojs:
When an escort is "near death" it follows my movements and ignores everything else. Which is the same behavior "healthy" escorts have when they don't have orders and there are no enemies around.

This behavior is paramount to suicide mid-battle since they are priority targets but just stand there waiting for damage, making it very hard to protect them.

I think they should actively try to stay near my ship while recovering (circle me?) and either run or face the enemy when they are hit (up to a certain distance from me).

I tried messing with the AI but failed to understand how to change their behavior when they are near death (too many behaviors are mixed in together).

MessyMix:
So, instead of behaving like escorts with no active orders, they behave
like they have the "gather" command, unless they are too far / already
engaged, in which case they attack? Do I "gather" your meaning? (God, that
was bad.)

flaviojs:
All I require is that they actively stay close to me (to defend them) so that's one possibility. I would have to test it mid-battle to see if it's appropriate. Gathering would probably be so-so since when I tell my escorts to gather they keep shooting past me. I imagine circling around me might be better, like when someone is scanning me.

As for when they are hit, the minimum I want is for them to face towards an enemy so the guns can shoot at it. Ideally the escort should run away from the enemy or hunt it while near me.

flaviojs:
Hmm, another alternative is for escorts to not to have any special "near death" behavior and be suicidal like the local police. When I board them they go right back into the fight, not even waiting for shields to recover.

At least escorts would be fighting back instead of just being a siting duck...

jafdy:
What if your ship was not a combat ship, perhaps they should gather round a combat ship instead. 
It sounds like you want escorts to behave like you would, unfortunately that is not likely to be a simple set of rules.

flaviojs:
My ship is not a combat ship, but it has better shields and shield regen so I can be a wall.

You hint at an interesting point there, the ideal behavior of the escorts depends on what kind of ship and strategy you have... maybe being able to set individual ship personalities is enough, I'll try it out.

ghoulavenger:
I have to admit, when I read this I thought for certain you were referring to captured ships. I don't recall any of my other ships ever behaving in that manner. If you are referring to captured ships, I thought this was a balancing metric, since the ships don't have captains but rather provisional crew.

flaviojs:
[...]

Issue #3092, opened 2017-09-29
Possible revised "health" calculation


tehhowch:
Current health system is very simplistic: `Health = Hull() + .5 * Shields()`
Hull() and Shields() are both 0 .. 1, meaning health is 0 .. 1.5.
The issue is that these functions and this representation do not convey how close a ship is to being destroyed, and if a ship has no shields, it starts with a health of 1, so the game thinks it is already rather damaged.

I propose the following:
1. a function which computes absolute pooled health:
```
 	// Calculate the absolute health of this ship, on a scale of 0 .. 1,
	// with the shield:hull importance ratio set by `shieldWeight`, default
	// 1:1. At 0, the ship is destroyed. At 1, the ship is undamaged.
	double Health(const Ship &ship, double shieldWeight = 1.)
	{
		double maxShields = shieldWeight * ship.Attributes().Get("shields");
		double maxHull = ship.Attributes().Get("hull");
		return (ship.Shields() * maxShields + ship.Hull() * maxHull)
				/ (maxShields + maxHull);
	}
```
This function is better - an undamaged ship always has health = 1, even if it doesn't have shields, but it still doesn't address awareness: if the ship has health = 0.5, is the next hit going to disable it? Or is it still quite far from the point at which it gets disabled?

2. a function which estimates the health at which a ship gets disabled:
```
	// Estimate the health at which this ship will become disabled, if it
	// takes damage in proportional with its current shield%-hull% ratio.
	// This accounts for direct hull damage effects rather than assuming
	// the shields will be ~0 before the hull takes damage.
	double DisabledHealth(const Ship &ship, double shieldWeight = 1.)
	{
		if(ship.IsDisabled())
			return Health(ship, shieldWeight);
		
		double maxShields = shieldWeight * ship.Attributes().Get("shields");
		double maxHull = ship.Attributes().Get("hull");
		// Estimated Health @ minHull =
		// (minHull + minHull * currentShields / currentHull) / (maxHull + maxShields)
		//  = minHull * (1 + currentShields / currentHull) / (maxHull + maxShields)
		double norm = 1. + (ship.Shields() * maxShields) / (ship.Hull() * maxHull);
		double minHealth = norm * ship.MinimumHull() / (maxHull + maxShields);
		
		return minHealth;
	}
```
[...]

PR #3369, opened 2017-12-05
Injured, player-owned escorts regroup with the player's flagship

tehhowch:
Refs #3008, #3004 
 - endless-sky#3008 fixed an issue with a loss of target combat
 - However if injured, the ships should not be lured away by enemy ships until they have repaired somewhat.

This makes the player's escorts behave with a bit more self-preservation instinct rather than - even while critically injured - chasing off into the distance after a hostile ship.


SirStigII:
Could this be made to work for allied ships as well?

tehhowch:
Allied ships are generally governed by the personality tags a content creator gave them, so I don't know that it is correct to override their chosen behavior.

It's certainly possible to do - the simplest way would be dropping the it->IsYours() check from the changes being added here. Note that this would work both ways - enemy ships will retreat to their parent to be a bit more protected during recovery, which will cause more disabled ships to be destroyed by crossfire (one of the most common 'complaints').

flaviojs:
Using `health < 1.` as a guard for the behavior is not good.

For small ships it might work fine (min hull 50%), but for big ships it definitely doesn't work well.
I let an escort Mule be disabled to see how it behaved. When it's shields are full it chases an enemy, but as soon as it gets a few hits it tries to run back to my side.
This would be acceptable for small fast/maneuverable ships that have little shield, but for big ships it is definitely bad. Another aggravating factor is that big ships have a lower minimum hull, and therefore trigger this behavior with more shield than small ships. A more consistent check would be `hull <= 50% && shield <= 50%` (assuming 0% shields when only hull is available).

From what I've seen during my play time with big fleets and small fleets, a good rule of thumb to avoid losses is to protect vulnerable ships (heavy hull damage) at least until their shields hit 50%. For some big ships, protecting until their shields reach the minimum hull is enough. Note that in really I'm just making a rough visual match of when they stop being priority targets, so a "smarter" way to guard the behavior would be to see if they are priority targets in comparison to other ships.

tehhowch:
I wanted to keep the health check the same as that which triggers alternate behavior in NPCs, in light of a possible rework to the health formulation. Given that such a rework will involve updating the values and target values anyway, I can tune for better behavior now.

tehhowch:
Swapped to use @flaviojs 's suggestion of `shields < .5 && hull <= .5`
Will need changing if we ever implement something that addresses #3092

endless-sky:
Wouldn't that make it so that small ships (i.e. ships that are disabled at 50% hull) never retreat?

tehhowch:
They would retreat while below 50% shields after being repaired, due to the `<=` compare for the hull.

endless-sky:
As with all AI behavior gated on shields or hull, this might run into the issue where a ship right on the threshold gets stuck oscillating between the two behaviors.

Also, I wonder if this could use the new Ship::Health() measure I added for fighter repairing, maybe somewhere in the 25-35% range. That measure is based on damage until disabled, so it avoid the issue where ships disabled at 50% think they've got more time left than they actually do.

And, note that this *would* make it so a ship that is disabled at 50% hull would never retreat (because its hull will never be below 50% if it's not disabled, and it only retreats if both shields and hull are below 50%). The likelihood of hull being *exactly* 50% without crossing below it is astronomically small, so in practice the `<=` is no different from a `<`.

tehhowch:
I saw the addition of that function in the big fighter commit and plan to address health based behavior with this pull this week.

tehhowch:
I originally included #3538 in this pull, but figured that should be a separate issue.
To avoid needing to rework the `appeasementThreshold` interactions, I've opted to keep it based on the ship's absolute health. Same for the cloak hysteresis.

LocalGod79:
i think more important to this would be to have damaged fighters and drones dock until repaired fully when the system is "safe" - the same conditions that will cause escorts to revive other escorts should trigger the "dock until repaired or hostiles are present" behaviour.  perhaps another issue/pr, but since preservation has been mentioned...

tehhowch:
@LocalGod79 when this pull was written, fighters did not dock with the carrier at all unless the deploy order was rescinded. I can certainly include that situation in the AI autobehavior - the only time it would be "inconvenient" would be if the fighters were assisting the parent ship in mining asteroids. Probably that will be a different PR, though.

PR #3370, opened 2017-12-05
Certain ships should not auto-assist their disabled parent

tehhowch:
While the parent won't ask its carried escort to help it, that escort would discover it has a disabled parent and attempt to assist it when combat ends, or when its own health is low enough that it has no combat target.

This PR removes that behavior, so that only non-carried ships will consider assisting their disabled parent.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -684,7 +684,7 @@ void AI::Step(const PlayerInfo &player)
 			else
 				MoveEscort(*it, command);
 		}
-		else if(parent->IsDisabled())
+		else if(parent->IsDisabled() && !it->CanBeCarried())
 		{
 			// Your parent is disabled, and is in this system. If you have enemy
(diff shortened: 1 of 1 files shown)


Pointedstick:
What's wrong with the current behavior, out of curiosity?

tehhowch:
Carried ships can't fix their parent at all, so they just park on top of it, which essentially guarantees the parent's destruction as the fighters get targeted

Pointedstick:
Ah, sounds good then.

tehhowch:
Derelict ships would also auto-repair their parent... Instead i've made them circle around the parent. 

(If no behavior is given here, they would fall through to KeepStation, and then it could appear that they are just doing nothing until the NPC flagship is repaired, as the only way to match position and velocity of a disabled ship is to act disabled yourself :rofl: )

Issue #3129, opened 2017-10-15
Conversations triggered upon boarding hostile mission NPCs


Amazinite:
I'm helping @TheUnfetteredOne write a spaceport mission that she hopes to get PRed, and while testing it we've run into a problem:
![screenshot_1](https://user-images.githubusercontent.com/17688683/31589522-fb905dc8-b1d0-11e7-8936-abb0f6bce89c.png)

A conversation starts when you board the hostile bounty ship spawned by the mission, but the boarding panel is displayed in front of the conversation panel.

Additionally, the following is a part from the wiki on the `assisting` and `boarding` tags for missions:
>A mission shown when "assisting" or "boarding" will be shown when you repair a friendly ship or plunder a hostile ship, respectively. These missions are never shown when boarding a ship that you have boarded before, that belongs to you, or that is an NPC in an active mission. In either case, if the "on offer" conversation results in "launch" or "flee," the ship in question will be destroyed.

Focusing on the last line there, `launch` or `flee` on a boarding mission conversation cause the boarded ship to explode. While the case above is on a mission NPC and not a random NPC where a mission has been triggered, the functionality of having being able to make the NPC in question explode upon the conversation's end would be useful.

tehhowch:
The display of the conversation before the BoardingPanel occurs because the mission & npc completion status is updated before the 'normal' result of a boarding event, via `player.HandleEvent()`:
https://github.com/endless-sky/endless-sky/blob/ef0a6043001928244214fcf89e51c745c4c2c7e8/source/MainPanel.cpp#L100-L134

Currently, all the events created by the `engine.Step(isActive)` call need to be handled on this run-through of `MainPanel::Step()`, so updating the value `isActive` to avoid the event handling (lines 107-133) is not a valid approach, unless the events are stored for later processing - which may be the smart way to go about this.
Reversing the order could be a possible solution, as that will place the dialog/conversation from completing the NPC on top of the BoardingPanel. This may have an (addressable) issue if the BoardingPanel's target is destroyed, and in any case the boarding panel would be visible while conducting the conversation.

For the launching/fleeing results, they *appear* to be already possible, but actually aren't. They require the boarded ship to be the source of a boarding mission since the value returned by `player.BoardingShip()` is currently only set by `player.BoardingMission` (called above).
https://github.com/endless-sky/endless-sky/blob/ef0a6043001928244214fcf89e51c745c4c2c7e8/source/ConversationPanel.cpp#L346-L357

[...]

PR #3371, opened 2017-12-05
Implement a ShipEvent queue to allow delayed ShipEvent handling

tehhowch:
Refs #3005 ,  #3129 
 - Handling ShipEvents can result in new UI elements.
 - Only one UI element should be created and displayed to the player at a time (i.e. do not display a dialog and the BoardingPanel simultaneously).
 - Callbacks from one UI element from a given ShipEvent may result in status changes that should influence other UI elements created from the same ShipEvent (e.g. if a mission conversation offered upon boarding an NPC results in that NPC's destruction, the player should not be able to plunder said NPC).

This PR does not attempt to fully address the existing issues with boarding missions (as that is done in #3005 ). It does add one tiny bit of #3005, to check that the player has space for a mission offered in space.
This PR does not implement the needed callbacks to the NPC completion step to enable player or NPC death (out of scope). This PR does, however, enable such a callback's effects to be properly handled by the UI: if the NPC dies, the BoardingPanel won't be shown.

### DIFF
--- a/source/ConversationPanel.cpp
+++ b/source/ConversationPanel.cpp
@@ -343,18 +343,21 @@ void ConversationPanel::Goto(int index, int choice)
 void ConversationPanel::Exit()
 {
 	GetUI()->Pop(this);
-	if(player.BoardingShip())
+	// If this is a conversation offered from boarding or assisting an NPC,
+	// being forced to leave via LAUNCH, FLEE, or DEPART destroys it. If it
+	// is a hostile NPC and not destroyed, and its mission (e.g. "leave me
+	// alone, here's X as payment") was not accepted, open a BoardingPanel
+	// to allow plundering it - unless you're dead.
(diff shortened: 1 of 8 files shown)


tehhowch:
Marking as WIP since testing showed that `GetUI()->IsTop(this)` doesn't return false after a ConversationPanel is pushed via `NPC::Do`

tehhowch:
`IsTop()` lies! (Well, doesn't tell the whole truth :grin: )

tehhowch:
Missed this bit in `ConversationPanel::Exit()`:
https://github.com/endless-sky/endless-sky/blob/224e62f563a55fc57e20b2e851a945836378bd0f/source/ConversationPanel.cpp#L350-L356
which already implemented the "when you exit a conversation with a hostile NPC,  board it to plunder unless it is being destroyed" bit.

tehhowch:
Awesome!

PR #3373, opened 2017-12-06
Tense Fix

Alkallid:


Bladewood:
The current form is still sort of correct, in a way: she is not immediately kicking you out of her office five seconds after you start talking to her.

Alkallid:
To me it sounds either like archaic speech that's mainly in present tense, or like you're currently deliberating with her about the meeting before it's occurred.

Issue #3228, opened 2017-11-13
"die" conversation endpoint does not work if conversation was started in space, from an "on enter" block


Pointedstick:
This doesn't work:

```
mission "Shady passenger transport - double cross"
	name "Shady passenger transport"
	description "Discreetly transport <bunks> shady passengers to <destination>."
	minor
	repeat
	source
		attributes frontier "dirt belt" rim north south pirate
	destination
		attributes "near earth" core urban
		distance 5 10
	passengers 6
	on enter
		conversation
			`As soon as you blast off, the cutthroat mercenaries you're transporting try to take control of the ship!`
			choice
				`(Shoot first!)`
					goto shoot
				`(Pay them off.)`
					goto "bribe"
			label shoot
			`Unfortunately, they're faster than you.`
				die
```

Instead of dying, the conversation simply ends.

The conversations wiki page suggests that `die` is an appropriate conversation endpoint and doesn't make any mention of it not working for conversations started while flying around, so this may simply be a bug. Either way, this should work, which would enable missions like the simplified stub above, where your passengers try to wrest control of your ship.

tehhowch:
This is the same issue behind #3206, #3167, and #3129 
There is no callback function that handles the conversation response unless the conversation is an `on offer`

tehhowch:
Possible with #3376

endless-sky:
Addressed by merging #3376.

Issue #3206, opened 2017-11-06
Can't die while assisting NPC


jjhankins:
```
mission "Test"
	name "Assist the derelict ship Test of Betrayal"
	landing

	npc assist
		personality derelict
		ship Bounder "Test of Betrayal"
		conversation
			`Something about the ship seems odd. Do go on board to assist with repairs?`
			choice
				`	(Yes.)`
				`	(no.)`
					goto end

			`	Pirates spring out of hiding and shoot you dead.`
				die

			label end
			`	You trust your instincts and flee the ship.`

	on offer
		conversation
			`Captain <last>! The ship 'Test of Betrayal' needs your help, please go and help it with its repairs.`
				accept
```

While if you enter the ship you do not see the "You trust your instincts" line, you also don't die. You're dropped back into the system, the derelict ship having been repaired.

EDIT: As a bonus, it would also be nice to destroy the ship and/or merely fail to repair it at the end of the conversation.

tehhowch:
This is because there is no callback function for NPC completion conversations, as mentioned in https://github.com/endless-sky/endless-sky/issues/3129#issuecomment-338011593

PlayerInfo::MissionCallback is what handles `launch`, `flee`, `die`, and other conversation actions:
https://github.com/endless-sky/endless-sky/blob/aad6f581e031e8eb71781180ff5a0e7a175b9e6d/source/PlayerInfo.cpp#L1524-L1558

As I mention in #3129, this is something I'll leave to MZ to address

jjhankins:
Ah, didn't realize that was the same issue.

tehhowch:
It wasn't explicitly stated there.
Even if destruction of that NPC ship from the conversation is not handled, `Die` certainly should be (so keep this open).

tehhowch:
Made possible with #3376

endless-sky:
Addressed by merging #3376.

PR #3376, opened 2017-12-07
NPC "completion" / MissionAction conversations can kill the player and/or NPC

tehhowch:
Refs #3206 , #3129, #3228 

 - Adds a new player-death keyword, `explode`, to resolve the ambiguity over how the player dies while in space
   - If the player's ship should be destroyed, use `explode`. The flagship will explode.
   - If the player's ship should be captured, use `die`. The player dies, and his ship is taken over by the ship it was boarding. (If `die` is used and there is no boarding ship, the flagship changes government, simulating a mutiny.)
 - If you're dying as a result of a conversation, you aren't going to be shown the BoardingPanel
 - `launch`, `depart`, and `flee` kill the NPC associated with a conversation, even if not boarded (say, if your outfit scan activated some embedded explosives' trigger mechanism...), unless you're dying too.

### DIFF
--- a/source/ConversationPanel.cpp
+++ b/source/ConversationPanel.cpp
@@ -48,8 +48,8 @@ namespace {
 
 
 // Constructor.
-ConversationPanel::ConversationPanel(PlayerInfo &player, const Conversation &conversation, const System *system, const Ship *ship)
-	: player(player), conversation(conversation), scroll(0.), system(system)
+ConversationPanel::ConversationPanel(PlayerInfo &player, const Conversation &conversation, const System *system, const shared_ptr<Ship> ship)
+	: player(player), conversation(conversation), scroll(0.), system(system), ship(ship)
 {
 	// These substitutions need to be applied on the fly as each paragraph of
 	// text is prepared for display.
@@ -346,18 +346,19 @@ void ConversationPanel::Goto(int index, int choice)
 void ConversationPanel::Exit()
 {
 	GetUI()->Pop(this);
-	// If this is a conversation offered from boarding or assisting an NPC,
-	// being forced to leave via LAUNCH, FLEE, or DEPART destroys it. If it
-	// is a hostile NPC and not destroyed, and its mission (e.g. "leave me
-	// alone, here's X as payment") was not accepted, open a BoardingPanel
-	// to allow plundering it - unless you're dead.
-	if(node != Conversation::DIE && player.BoardingShip())
+	// If this is a conversation offered from an NPC, ending it via LAUNCH,
+	// FLEE, or DEPART destroys the NPC. For hostile NPCs that are boarded
+	// and not being destroyed, show the BoardingPanel unless the player is
+	// being killed or the conversation is exiting via ACCEPT.
+	if(node != Conversation::DIE && ship)
 	{
 		if(Conversation::RequiresLaunch(node))
-			player.BoardingShip()->Destroy();
-		else if(player.BoardingShip()->GetGovernment()->IsEnemy()
+			ship->Destroy();
+		// Check that this ship is the player's boarding target, as NPC
+		// completion conversations can result from non-boarding events.
+		else if(ship == player.BoardingShip() && ship->GetGovernment()->IsEnemy()
 				&& node != Conversation::ACCEPT)
-			GetUI()->Push(new BoardingPanel(player, player.BoardingShip()));
+			GetUI()->Push(new BoardingPanel(player, ship));
 	}
 	// Call the exit response (e.g. ACCEPT, DIE) handler, which is usually
 	// PlayerInfo::MissionCallback.
--- a/source/ConversationPanel.cpp
+++ b/source/ConversationPanel.cpp
@@ -355,18 +355,17 @@ void ConversationPanel::Exit()
 	{
 		// A forced-launch ending (LAUNCH, FLEE, or DEPART) destroys any NPC.
 		if(Conversation::RequiresLaunch(node))
-			// TODO: This may need a ShipEvent, but not all instances
-			// are necessarily caused by the player.
 			ship->Destroy();
 		// Only show the BoardingPanel for a hostile NPC that is being boarded.
 		// (NPC completion conversations can result from non-boarding events.)
-		else if(ship->Position().Distance(player.Flagship()->Position()) <= 1.
-				&& ship->GetGovernment()->IsEnemy()
-				&& node != Conversation::ACCEPT)
+		// TODO: Is there a better / more robust boarding check than relative position?
+		else if(node != Conversation::ACCEPT && ship->GetGovernment()->IsEnemy()
+				&& !ship->IsDestroyed() && ship->IsDisabled()
+				&& ship->Position().Distance(player.Flagship()->Position()) <= 1.)
 			GetUI()->Push(new BoardingPanel(player, ship));
 	}
-	// Call the exit response handler (PlayerInfo::MissionCallback)
-	// to manage the conversation's effect on the player's missions.
+	// Call the exit response handler to manage the conversation's effect
+	// on the player's missions, or force takeoff from a planet.
 	if(callback)
 		callback(node);
 }
--- a/source/ConversationPanel.cpp
+++ b/source/ConversationPanel.cpp
@@ -48,7 +48,7 @@ namespace {
 
 
 // Constructor.
-ConversationPanel::ConversationPanel(PlayerInfo &player, const Conversation &conversation, const System *system, const shared_ptr<Ship> ship)
+ConversationPanel::ConversationPanel(PlayerInfo &player, const Conversation &conversation, const System *system, const shared_ptr<Ship> &ship)
 	: player(player), conversation(conversation), scroll(0.), system(system), ship(ship)
 {
 	// These substitutions need to be applied on the fly as each paragraph of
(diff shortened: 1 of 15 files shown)


tehhowch:
Examples:
```
	npc board
		government "Pirate"
		personality target entering unconstrained harvests plunders
		ship "Protector (Proton)" "Bastard of the Pond"
		conversation
			`"Good job, <first>! I like this ship!"`
			`	(You are killed.)`
				die
```
```
	on offer
		conversation
			`Your ship's dosimetry sensors begin crackling as you dock with <ship>, but before you can abort, a sharp clanking echoes from your hull. The momentary silence is followed by a chilling, robotic voice emanating from your ship's P.A. system:`
			``
			`	"Radiological threat detected."`
			``
			`A highly radioactive, shaped-charge explosive has attached to your hull. Uh oh.`
			`	(You disintegrate.)`
				explode
```

tehhowch:
Added support for #3228 as well

Pointedstick:
Wow, amazing! Almost there for my mutiny mission sketched out in https://github.com/endless-sky/endless-sky/issues/3228#issuecomment-343784930, but not quite... In the example mission above, I don't want the ship to blow up. The mercenaries capture it and fly off. Maybe `die` without `explode` could do that? In other words, just undo or don't implement the condition "(If die is used and there is no boarding ship, the flagship explodes.)"

tehhowch:
Right now that "explode if dead but not boarded" behavior is basically a guard against referencing a ship that doesn't exist. I haven't really checked what else can be done - it's possible just changing the "is yours" property would be sufficient.

tehhowch:
@Pointedstick should be all set now :+1: 
For now I've opted to have the mutineer government be the "Independent" government. A future PR could extend how this is selected.

(I also added some reparenting so that, should the player's flagship be captured in any manner, the rest of the player's fleet does not take orders from the captured ship. This means they won't follow it around, wait for it to jump, etc)

Pointedstick:
Outstanding. I have some fun missions queued up to submit once this is in...

tehhowch:
Confirmed these all still work after the change in ShipEvent handling :+1:

source/PlayerInfo.cpp
```diff
@@ -2579,3 +2621,46 @@ bool PlayerInfo::CanBeSaved() const
...
+			flagship->Destroy();
+		else
+		{
+			flagship->SetIsYours(false);
+			// If the flagship is boarding an NPC, the player's death results in its
+			// capture. Otherwise, consider the player to have died due to a mutiny.
+			// Change the flagship government to remove it from the EscortDisplay HUD.
+			if(boardingShip)
+			{
+				flagship->SetSwizzle(boardingShip->GetSwizzle());
```
endless-sky: Ships don't actually change swizzle when they're captured, since that requires a new paint job. (They don't change until you land on a planet.)

source/PlayerInfo.cpp
```diff
@@ -1634,6 +1645,37 @@ void PlayerInfo::HandleEvent(const ShipEvent &event, UI *ui)
...
+		// Nothing to do.
+	}
+	else if(response == Conversation::DECLINE || response == Conversation::FLEE)
+	{
+		// Nothing to do.
+	}
+	else if(response == Conversation::DEFER || response == Conversation::DEPART)
+	{
+		// Nothing to do.
+	}
```
endless-sky: I don't think all the above conditions need to be here. To my mind it would be clearer to only have an `if` for the responses that actually have a result.

source/Ship.cpp
```diff
@@ -1980,8 +1982,8 @@ void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
 	// Transfer some crew over. Only transfer the bare minimum unless even that
 	// is not possible, in which case, share evenly.
 	int totalRequired = capturer->RequiredCrew() + RequiredCrew();
-	int transfer = RequiredCrew();
-	if(transfer)
+	int transfer = RequiredCrew() - crew;
+	if(transfer > 0)
 	{
 		if(totalRequired > capturer->Crew())
 			transfer = max(1, (capturer->Crew() * RequiredCrew()) / totalRequired);
```
endless-sky: If you're adding support for capturing a ship that still has crew, I think this would need to be:

```C++
if(totalRequired > capturer->Crew() + crew)
    transfer = max(crew ? 0 : 1, (capturer->Crew() * transfer) / totalRequired);
```

That is, if there's not enough crew to go around, split the crew from both ships evenly. And, if the victim has no crew, make sure to give it at least one.

source/PlayerInfo.cpp
```diff
@@ -2579,3 +2621,46 @@ bool PlayerInfo::CanBeSaved() const
...
+		Die(true);
+	else
+	{
+		Die();
+		auto it = find(ships.begin(), ships.end(), flagship);
+		if(response == Conversation::EXPLODE)
+			flagship->Destroy();
+		else
+		{
+			flagship->SetIsYours(false);
```
endless-sky: Technically speaking, the flagship's crew should be reduced by 1 (since you're dead). But, that would create an issue if you're boarded by someone with only 1 crew, and you only have 1 crew too, because there would not be enough crew to control both ships.

source/PlayerInfo.cpp
```diff
@@ -2579,3 +2621,46 @@ bool PlayerInfo::CanBeSaved() const
...
+				it = ships.erase(it);
+		}
+		
+		// Let the player's fleet pick a new flagship to follow, if possible.
+		for( ; it != ships.end(); ++it)
+			if((*it)->CanBeFlagship())
+				break;
+		const shared_ptr<Ship> newFlagship = it != ships.end() ? *it : nullptr;
+		for(const shared_ptr<Ship> &escort : ships)
+			escort->SetParent(escort != newFlagship ? newFlagship : nullptr);
```
endless-sky: This seems unnecessary because the game is already over - it's no big deal if your escorts scatter instead of picking a new leader and departing together.

source/ConversationPanel.cpp
```diff
@@ -343,21 +343,22 @@ void ConversationPanel::Goto(int index, int choice)
...
-	// being forced to leave via LAUNCH, FLEE, or DEPART destroys it. If it
-	// is a hostile NPC and not destroyed, and its mission (e.g. "leave me
-	// alone, here's X as payment") was not accepted, open a BoardingPanel
-	// to allow plundering it - unless you're dead.
-	if(node != Conversation::DIE && player.BoardingShip())
+	// If this is a conversation offered from an NPC, ending it via LAUNCH,
+	// FLEE, or DEPART destroys the NPC. For hostile NPCs that are boarded
+	// and not being destroyed, show the BoardingPanel unless the player is
+	// being killed or the conversation is exiting via ACCEPT.
+	if(node != Conversation::DIE && node != Conversation::EXPLODE && ship)
```
endless-sky: Note: the `ship` pointer could be a ship you just scanned that triggered a conversation, in which case making it explode seems weird.

source/PlayerInfo.cpp
tehhowch: I agree, they don't need to be here. I added them because I thought it was possibly beneficial to explicitly convey that "yes, these responses are possible, here is where they are handled, and they do nothing w.r.t the player."

source/PlayerInfo.cpp
tehhowch: I thought of reducing it, but since the player is dead and can no longer inspect the ship that just boarded them, it was better to keep as-is than reduce it and perhaps cause an issue.

source/PlayerInfo.cpp
tehhowch: I added this so that the player's fleet doesn't simply obey whoever killed the player, and follow that the player's "old" ship around (since it isn't destroyed, they won't scatter).

source/Ship.cpp
tehhowch: You are correct

source/PlayerInfo.cpp
tehhowch: Indeed, here it was added solely as a visual indicator of "what happened," even if the mission-writer didn't really expound on the details.

source/ConversationPanel.cpp
tehhowch: `"We've sabotaged this <npc> with explosives, and they're set to go off when you scan their cargo"`

endless-sky:
Can you think of any reason why the original Die() function needed a version that did not clear the player's fleet? It seems like there's no reason for that boolean switch.

If the original Die() always cleared the player's whole fleet, cleared the flagship pointer, and cleared IsYours() on every one of the player's ships for good measure, just calling Die() would be enough to make the player's fleet fly off on its own. Then, instead of your whole extra Die() function, all you'd have to do is destroy the flagship if responding to an EXPLODE response (and not landed, and the flagship exists).

Also, your code goes halfway toward eliminating the need for PlayerInfo::BoardingShip() by passing a shared pointer to the ship in to ConversationPanel. If Mission::CanOffer() and Mission::Instantiate() had an optional ship pointer as a second argument, I think tracking `boardingShip` in PlayerInfo and clearing it at the proper time would become unnecessary.

In other words, I think there would be ways to implement this that would make the code a little bit simpler instead of making it more complex.

source/ConversationPanel.cpp
Pointedstick: > "We've sabotaged this \<npc\> with explosives, and they're set to go off when you scan their cargo"

[gets some ideas]

tehhowch:
I'll have to mull that change over in the next few days (feel free to take the initiative and do it for me if you're feeling inspired 👍). Otherwise, I can probably get to it Monday evening or so.

tehhowch:
Based on the commits involved, the boolean version was added because moving `ships.clear()` into `Die()` (from `MissionCallback()`) "broke the normal death" wherein the flagship is destroyed in flight (presumably because all the remaining player ships simply vanished). PI::HandleEvent (expected) and the fine-handling of `atrocity` while landed (almost certainly a bug, and now addressed in #3208) are currently the only non-`allShipsDie` caller on `master`.
I feel this non-`allShipsDie` behavior is what needs to be replicated when the player dies to capture / mutiny, and that clearing of the player's ship list only needs to be done if the player is on a planet (lest they simply press D to take off).

tehhowch:
With 6231420 (as well as #3208), the player being detained for life due to landing on a planet and getting caught with atrocity-level outfits actually sticks (instead of being able to simply press `D` and take off, or do anything else while landed, from memory.

tehhowch:
@endless-sky The only bits I'm not 100% satisfied with are the boarding checks that guard the player being captured, and showing the BoardingPanel after rejecting the NPC's mission. Other than that, I've successfully eliminated BoardingShip, and eliminated the need for the second callback.

source/ConversationPanel.cpp
```diff
@@ -343,21 +343,27 @@ void ConversationPanel::Goto(int index, int choice)
...
+	// block (e.g. scanning or boarding or killing all required targets).
+	if(node == Conversation::DIE || node == Conversation::EXPLODE)
+		player.Die(node, ship);
+	else if(ship)
 	{
+		// A forced-launch ending (LAUNCH, FLEE, or DEPART) destroys any NPC.
 		if(Conversation::RequiresLaunch(node))
-			player.BoardingShip()->Destroy();
-		else if(player.BoardingShip()->GetGovernment()->IsEnemy()
+			// TODO: This may need a ShipEvent, but not all instances
```
endless-sky: I think this is safe, because Engine::MoveShip() also creates a DESTROY event when a ship has finished exploding, to handle self-destructing ships.

source/ConversationPanel.cpp
```diff
@@ -343,21 +343,27 @@ void ConversationPanel::Goto(int index, int choice)
...
+			// are necessarily caused by the player.
+			ship->Destroy();
+		// Only show the BoardingPanel for a hostile NPC that is being boarded.
+		// (NPC completion conversations can result from non-boarding events.)
+		else if(ship->Position().Distance(player.Flagship()->Position()) <= 1.
```
endless-sky: Could this check `&ship == player.BoardingShip().get()` instead? Or is there a reason that wouldn't work? Relying on the distance to identify a ship you're boarding just seems like a bit of a hack.

source/ConversationPanel.h
```diff
@@ -124,6 +125,10 @@ template <class T>
 	// If specified, this is a star system to display with a special big pointer
 	// when the player brings up the map. (Typically a mission destination.)
 	const System *system;
+	// If specified, this is a pointer to the ship that this conversation
+	// acts upon (e.g. the ship failing a "flight check", or the NPC you
+	// have boarded).
+	std::shared_ptr<Ship> ship = nullptr;
```
endless-sky: shared_ptr<> doesn't need to be initialized, because its default constructor will set it to null.

source/ConversationPanel.cpp
endless-sky: Oops, never mind, now I remember that the point was to eliminate PlayerInfo::BoardingShip().

endless-sky: I wonder if checking `ship->IsYours()` would work. The only times a ConversationPanel has a ship is if 1. it's a flight check message (for one of your ships) or 2. it's a conversation being shown because you boarded a ship. So, if the ship is not yours, I think it must be a ship you're boarding.

tehhowch: NPC completion also packages a ship into ConversationPanel (hence the code comments), and the ship isn't necessarily being boarded - it might have been scanned, destroyed, disabled, etc.

PR #3379, opened 2017-12-07
Refactor MapPanel's Draw___ methods

tehhowch:
Splitting the MapPanel refactoring out of #2601 so that #2601 can deal only with adding functionality re: invalid travel plans and not also rewrite a good chunk of `MapPanel`
 
- Improve mod coloring support via expanded use of interfaces.txt:
   - Switch the minimap from a hardcoded position to an `Interface`
   - Use travel plan colors from interfaces.txt
   - Use System name, link colors from interfaces.txt
 - Move the increment of `step` from the `DrawMissions` method to the base `Draw` method
 - Add comments to `DrawMiniMap`, `DrawTravelPlan`, and `DrawWormholes`
 - Used and updated `MapPanel` constants
 - Draw the Travel Plan after wormholes, since wormholes by default have non-zero alpha.
   - This removes the small, dark purple centerline from drawn travel plan routes that use wormholes
 - Adds function to `Planet` to export all systems in which it can be found
 - Do not show the "planet description" box if the player selects a wormhole planet (since drawn wormholes have non-empty descriptions)
 - Change `DrawWormholes` iteration scheme from "every `StellarObject` in every `System`" to "check only `planets`"
   - Fixes bug in `DrawWormholes` caused when two wormholes reference the same system (i.e. Cardea), which cause the wormhole link to be drawn over a previous directional arrow

images from #2601:

|`master` | 993e617 |
|:-----:|:-----:|
|image | image|

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -64,6 +64,15 @@ color "radar blink" 1. 1. 1. 0.
 color "error back" .25 .1 .1 1.
 color "warning back" .21 .18 .08 1.
 
+color "map link" .6, .6, .6, .6
+color "map name" .6, .6, .6, .6
+color "map travel ok fleet" .2, .5, 0., 0.
+color "map travel ok flagship" .5, .4, 0., 0.
+color "map travel ok none" .55, .1, 0., 0.
+color "map used wormhole" .5, .2, .9, 1.
+color "map unused wormhole" .165, .066, .3, .333
+
+
 
 
 interface "menu background"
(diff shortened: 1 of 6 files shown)


data/interfaces.txt
```diff
@@ -261,6 +270,12 @@ interface "preferences"
 
 
 
+interface "mini-map" top
```
endless-sky: Would it make more sense for this to be part of the "hud" interface, so people could customize it along with the location for the messages, ammo, escorts, etc.?

tehhowch: It could be placed there, yes, though I think both locations are equally customizable. Placing it inside the "hud" interface does help imply to anyone doing ui tweaks when the minimap is used.

I'll make the changes shortly.

tehhowch:
Should be all set now. I moved it down another 50px as some systems could get cut off at the `(0, 100)` position.

PR #2456, opened 2017-04-28
Substitutions in conditions

EndrosG:
I think I got it done (#2351). At least it works now on my local build, after a few failed attempts. :-)

The `<...>` substitution now also applies to all conditions and not only to conversation texts. All = all: conditions in conversations (i.e. `branch` conditions and `apply` conditions), and conditions outside conversations, i.e. in each of the on-triggers as well as in the to-triggers (`toOffer`, `toFail`, `toComplete`).
For the toOffer trigger, the substitution mostly can only apply after the CanOffer has been checked, but at least `<origin>` and *new* `<originGov>` are available in the `toOffer` by now.

I added a few new replacements:
- `<originGov>`, `<planetGov>`, `<systemGov>` - all 3 yielding the GovernmentName of the origin, (dest.) planet, (dest.) system respectively.
- `<tonsNum>` and `<paymentNum>` since I wanted the first one already for a while now.

This patch allows for, but is not limited to:
- missions / jobs being offered depending on your reputation at the origin
```
	to offer
		"reputation: <originGov>" > 100
```
- modifying your reputation with either source or destination
- keeping track of certain things (I will provide examples for that, but may be I will wait until #2449 is merged because it will be another modification of the jobs.txt)
- up to certain limits, this even allows for job chains, I will provide an example later.

Attachments:
The testing mission I used during coding and one of the instances created by the latest code version (the one I hereby open/submit as pull)
[testing_mission_substitute_conditions.txt](https://github.com/endless-sky/endless-sky/files/965627/testing_mission_substitute_conditions.txt)
[Endros G~instance of TEST Substitute Condition.txt](https://github.com/endless-sky/endless-sky/files/965628/Endros.G.instance.of.TEST.Substitute.Condition.txt)

PR #3298, opened 2017-11-26
Blockade running missions

Amazinite:
Over a year later, but someone finally got around to doing the blockade running missions mentioned in [this blog post](https://endless-sky.github.io/2016/11/21/easy-ways-to-get-involved.html)! You have no designated escorts, but sometimes it isn't just "your own ship" (I just wrote down a few story ideas and went from there, and in some cases it seemed reasonable that FW or Navy forces would be involved), so while not 100% what was asked for in some cases (Northern and Southern blockade missions), I hope it still fits the bill.

The missions:
* **Northern Blockade**: The distant world of Featherweight is under attack by a sizable gang of pirates. The Navy has responded, but there are no available Navy ships that can transport relief supplies to help repair the world's crippled spaceport. 
* **Core Blockade**: The pirate world of Deadman's Cove is experiencing a famine, and the planet's warlords aren't too happy about it. They've threatened to attack Syndicate merchants and planets if they don't get food, and the Syndicate has decided that this would be the cheaper option... they just don't want to risk any of their ships taking it.
* **Southern Blockade**: Three frontier Free World planets close to the system of Men are under attack. The Free Worlds would like to deter any future raids, and have therefore asked you to transport spaceport defense turrets to the worlds... while the attacks are still occurring.
* **Pirate Blockade**: Not everyone on pirate worlds are bad people. Assist a woman on Buccaneer Bay who claims to be part of a nebulous organization who's goal it is to end slavery on all pirate worlds, something the Republic doesn't seem to want to spend the resources doing. Transport a small number of freed slaves to a nearby frontier world where another member of the organization will bring them far from any pirate world.

Issue #2351, opened 2017-04-03
request: additional text replacement in missions


EndrosG:
Teaser: Add a `<tons number>` placeholder.
**Edit**: And allow usage of placeholders in apply sections (conditions), i.e. on any parts of the `on complete` and not only in descriptive texts such as dialogs and conversations.

Relates to: `<tons>` and `<bunks>` placeholders.

Short reasoning:
`<bunks>` provides a pure number, while `<tons>` provides a number followed by a word.
Thus, `<bunks>` can be used for numeric operations while `<tons>` cant be used for numeric operations.

Long reasoning:
I was inspired by the current coalition jobs to create a set of jobs aiming to deliver a total of (xxx) tons of (supplies) to (some destination), similar to the supplies and volunteers for the ring worlds in the coalition, but not with a "Coalition Jobs"++ on complete but with some "ringworld workers" += `<bunks>` and "ringworld supplies" += `<tons number>`
And since cargo (tons) is usually flexible / random, one cant hardcode it in the += operation. If I would use fixed cargo sizes I could workaround, but fixed cargo sizes aint "nice". :-/

Of course I had some different usage in mind, different from the coalition plot. I want to create some deeper player interaction by events based upon sufficient supplies. But that doesnt matter ;-)

KiLEdEnNis:
I'm lost
what are you requesting?

EndrosG:
`<bunks>` is a pure number like `15`, but `<tons>` is a text like `15 tons`. I would like to have a pure number for the cargo as well. `<tons>` is more the equivalent to `<fare>` and there is no cargo equivalent to `<bunks>`.

Reference quote from wiki
```
# Text replacements #

Certain characteristics of a mission, such as the cargo or the destination planet, may be chosen at random. In order to refer to those randomly chosen elements in descriptive text, you can use the following placeholders:

    <commodity> = name of commodity being carried
    <tons> = "1 ton" or "N tons"
    <cargo> = "<tons> of <commodity>"
    <bunks> = the number of passengers
    <passengers> = "passenger" or "passengers"
    <fare> = "a passenger" or "N passengers"
```

KiLEdEnNis:
why would you need a pure number for cargo? it IS in tons

Wrzlprnft:
that's for sure a creative use of those replacements, but i assumed those are strings anyway? does that even work with the passenger number yet?

tehhowch:
the actual int values are in the mission's public CargoSize() property: 
[...]

PR #3384, opened 2017-12-08
Create a few <government: ___> substitutions for mission text

tehhowch:
Refs #2456, though I chose to use a more flexible / descriptive naming scheme than `originGov`

 - `<government: origin>`: The name of the government of where this mission was offered (e.g. the ship's government or the planet's government).
 - `<government: destination>`: The name of the government of the destination planet.
 - `<government: waypoints>`: The name(s) of the government(s) of all systems which are waypoints
   e.g. "Free Worlds and Republic"
 - `<government: stopovers>`: The name(s) of the government(s) of all planets which are stopovers
   e.g. "Free Worlds, Republic, and Syndicate"
 - `<government: npc>`: The name of the government of the first NPC in the mission's definiton.
 - `<government: npcs>`: The name(s) of all this mission's NPC's government(s).

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -61,6 +61,33 @@ namespace {
 		// Control will never reach here, but to satisfy the compiler:
 		return nullptr;
 	}
+	
+	// This returns the names of the governments within the set, joined as
+	// appropriate with "," and "and".
+	string GovernmentNames(vector<const Government *> &governments)
+	{
+		if(governments.empty())
+			return "";
+		else if(governments.size() == 1)
+			return governments[0]->GetName();
+		else
+		{
+			set<string> sortedNames;
+			for(const Government *gov : governments)
+				sortedNames.insert(gov->GetName());
+			
+			const string *last = &*--sortedNames.end();
+			string names;
+			int count = 0;
+			for(const string &name : sortedNames)
+			{
+				if(count++)
+					names += (&name != last) ? ", " : (count > 2 ? ", and " : " and ");
+				names += name;
+			}
+			return names;
+		}
+	}
 }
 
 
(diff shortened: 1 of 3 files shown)


Amazinite:
Just out of curiosity, could this (fairly) easily be done for conditions as well? https://github.com/endless-sky/endless-sky/pull/3298#discussion_r153672510 could be done easier if we could just put `"government: source" >= 0` on a lot of missions.

Also, maybe keep the pirate attack conversation in the jobs.txt next to the pirate attack jobs; the Wanderer events aren't even in their own file like the FW events are, probably because the FW events could also end up being used in the other two human campaigns, while the Wanderer events are only used in the Wanderer campaign and therefore are found by the missions that trigger them.

flaviojs:
I would prefer `<name[:formatter[:argument]*]>` for extendability. I suggested the same format in #2351.

In this case it would translate to `<origin:government>`, `<destination:government>`, `<waypoints:government>`, `<stopovers:government>`, `<npc:government>`, `<npcs:government>`.

tehhowch:
@amazinite I haven't attempted to extend the substitution behavior into conditions, which is indeed done in #2456. In #3128 I asked MZ if I should also perform that extension, or if it should be reserved for a new PR.

@flaviojs i am not opposed to your syntax. The current formatter is just map<string, string> and then text replacement, so there is likely a better way to go about this.
Your format is generally `thing->property` while mine is `property->thing`. When writing the text, is it easier for the layman (who doesn't know "code" beyond endless-sky keywords and indentation) if the first word is the desired substitution, and the rest is specificity, or if the last word is the desired substitution, and the start is the specificity?
That is I think a decision that needs to come down from MZ (or a better formatter helper that can handle both syntaxes natively).

source/Mission.cpp
```diff
@@ -1033,34 +1068,59 @@ Mission Mission::Instantiate(const PlayerInfo &player, const shared_ptr<Ship> &b
...
 	// Instantiate the NPCs. This also fills in the "<npc>" substitution.
 	for(const NPC &npc : npcs)
 		result.npcs.push_back(npc.Instantiate(subs, source, result.destination->GetSystem()));
+	// Fill in the NPC government substitutions, "<government: npc>" and
+	// "<government: npcs>". The singular always refers to the first NPC
+	// block, while the plural refers to all of the mission's NPCs.
+	if(!result.npcs.empty())
+	{
+		if(!result.npcs.front().Ships().empty())
+			subs["<government: npc>"] = result.npcs.front().Ships().front()->GetGovernment()->GetName();
```
tehhowch: I think i need to re-check this re: how `<npc>` is filled in. `<npc>` and `<government: npc>` should definitely refer to the same NPC to avoid confusion.

Amazinite:
What's the status on this?

Amazinite:
I repeat my previous question.

tehhowch:
The status on this is that it needs a rethink and a re-do. First item of order is to standardize how we should refer to properties of various things (i.e. the syntax that all mission writers must use to do something like refer to the 3rd NPC's government and the 1st NPC's system name). After that comes the implementation of the format system, and then we can swing back around to using the system.

PR #3386, opened 2017-12-09
Fancier ships in paradise

Pointedstick:
Added some more fancy and passenger-oriented ships to the fleets that hang around Paradise planets, to make that region of space feel busier and richer--full of wealthy tourists, busy killer suits, and spoiled teenagers racing Flivvers.

### DIFF
--- a/data/fleets.txt
+++ b/data/fleets.txt
@@ -343,7 +343,7 @@ fleet "Paradise Merchants"
 		"Wasp (Proton)" 3
 	variant 2
 		"Star Queen"
-		"Quicksilver 2
+		"Quicksilver" 2
 	variant 10
 		"Arrow"
 	variant 5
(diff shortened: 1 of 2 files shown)


tehhowch:
Will these mostly-stock ships be getting luxury accommodations if/when that merges?

Pointedstick:
That PR already adds Luxury Accommodations to the Star Queen. I could see adding (Luxury) variants for some of the other transport-class ships.

Pointedstick:
I've updated https://github.com/endless-sky/endless-sky/pull/3359 to add (Luxury) variants for the Arrow, Bounder, and Flivver, and gave the Blackbird Luxury Accommodations by default, to match its description. That takes off some of the pressure to merge this, but I think it's still a benefit. If https://github.com/endless-sky/endless-sky/pull/3359 is merged, this will need further adjustment to change the ships listed here to their (Luxury) variants.

PR #3350, opened 2017-12-03
Increased gun ports on some human interceptors to encourage blaster use

Amazinite:
* Increase the number of gun ports that some interceptors have in order for them to take better advantage of the small size but considerable power density of blasters
  * The increase to interceptor gun ports is outlined in #3264. The Sparrow and Wasp gain the ability to install 2 more blasters, and the Fury gains the ability to install 2 more modified blasters OR 1 more beam laser/gatling gun with this PR.
  * New Wasp and Sparrow graphics by @KiLEdEnNis. New Fury graphic by @Kryes-Omega. 2x images at the ready and assets linked below.
  * No additional variants have yet been added that would take advantage of these increased gun ports and show the player how one could manage to say fit 4 blasters on a Sparrow and still run smoothly.
  * I opted to exclude the Berserker change as the description implies that one of the reasons it is falling out of popularity is due to its lack of versatility, which a lack of gun ports in order to saturate its weapon space with blasters would support.
  * The Hawk has also received no changes due to the fact that it currently seems like an interceptor designed to carry anything BUT blasters, seeing as how it has the same weapon space as the Fury but half the gun ports.

PR #3393, opened 2017-12-10
Reduced damage and increased firing energy and heat of human lasers

Amazinite:
Reference: https://github.com/endless-sky/endless-sky/pull/3307#issuecomment-348807522

To note ahead of time, I have only changed lasers in this PR. I have not (yet) altered any ships or variants that use lasers in order to avoid any issues during the story. I plan to do that before considering this merge worthy, but I'm PRing now in order to get feedback on the changes and outline the effects that this PR has on ships.

Changes:
* All human lasers have seen... 
  * a roughly 10% decrease in damage
  * a roughly 20% increase in both firing energy and heat

Goal: make cannons and blasters more competitive with human lasers by reducing how efficient human lasers are.

I spreadsheeted out all 73 human ships (stock and variant) that use lasers, showing their firing energy/heat before and after the change and their time to run out of energy or overheat before and after the change assuming that all systems are firing (i.e. shooting weapons, thrusting, steering, and regenerating shields). That spreadsheet can be found [here](https://docs.google.com/spreadsheets/d/13JEoF8XAzcX2nZMB9cdZQKgFeLQdZOv1R-E7VUWDi5Y/edit?usp=sharing). All times are in seconds. Ships are not always running all their systems at once, I am simply looking at the worst case scenario change that would occur because of this PR. I expect that normal combat wouldn't be affected too much, only prolonged combat.

Notable results of this change:
* Ships that run out of energy with ALL systems firing constantly that didn't before
  * Clipper
  * Fury (Laser)
  * Marauder Raven
  * Falcon (Heavy)
  * Leviathan (Laser)
  * Mule (Hai Engines)
* Ships that will now overheat with ALL systems firing constantly that didn't before
  * Behemoth
  * Gunboat
  * Aerie
  * Mule
  * Leviathan (Laser)
  * Marauder Leviathan
* Ships that ran out of energy before now run out on average 26.6% faster
* Ships that overheat before now overheat on average 35.28% faster
* Notable ships/variants that changed (looking at ships that play a role in the campaign):
  * Gunboats now run out of energy roughly 26% faster, and Mark IIs run out roughly 41% faster.
  * All Navy Carriers now run out of energy roughly 20% faster. The stock Carrier overheats about 29% faster, but the Mark II and Jump variants overheat roughly 10% faster.
  * Mark II Cruisers now run out of energy roughly 83% faster, and normal Cruisers roughly 30% faster. No Cruisers that see combat (i.e. not the Cruiser (Heavy)) will overheat.
  * The Falcon (Heavy), without any batteries, may now struggle during combat.
  * The Leviathan (Heavy) now runs out of energy roughly 30% faster.

I strongly suggest merging #3350 alongside this, as there are a number of ships that use both lasers and blasters, and with the blaster change and despite the increased difficulty of using lasers, those ships actually end up overheating slower and running out of energy slower or even not at all due to how drastic the change is to blasters (or said differently, how drastically high firing energy and heat of blasters are).

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -157,8 +157,8 @@ outfit "Beam Laser"
 		"velocity" 300
 		"lifetime" 1
 		"reload" 1
-		"firing energy" .6
-		"firing heat" 1.45
+		"firing energy" .5
+		"firing heat" 1.2
 		"shield damage" 0.9
 		"hull damage" 1.15
 	description "In the early part of the space era, the settlements in the region of space known as the Deep developed in relative isolation from the rest of human space. One result of that isolation is that their weapons technology mostly uses beam weapons, instead of the energy projectile weapons that are more common elsewhere. A Beam Laser has a slightly shorter range than an Energy Blaster, but is also much more energy-efficient and does a comparable amount of damage."
@@ -184,8 +184,8 @@ outfit "Laser Turret"
 		"velocity" 300
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1.2
-		"firing heat" 2.9
+		"firing energy" 1
+		"firing heat" 2.4
 		"shield damage" 1.8
 		"hull damage" 2.3
 	description "Laser Turrets are hated by fighter pilots because it is nearly impossible to dodge them once you are within their reach. This turret carries a pair of lasers and can swivel almost instantly to fire on new targets as they approach. Laser Turrets are especially useful when mounted on slow-moving freighters to fend off packs of small pirate vessels."
@@ -217,8 +217,8 @@ outfit "Heavy Laser"
 		"velocity" 400
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1.2
-		"firing heat" 2.65
+		"firing energy" 1
+		"firing heat" 2.2
 		"shield damage" 1.6
 		"hull damage" 2.2
 	description "The Heavy Laser is an upgraded Beam Laser with a significantly longer range and higher power. It is mostly intended for larger ships, where energy and space are plentiful, but some pilots consider a single Heavy Laser to be a worthwhile alternative to a pair of Beam Lasers, because the longer range makes up for the fact that it does not quite deal twice as much damage."
@@ -244,8 +244,8 @@ outfit "Heavy Laser Turret"
 		"velocity" 400
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 2.4
-		"firing heat" 5.3
+		"firing energy" 2
+		"firing heat" 4.4
 		"shield damage" 3.2
 		"hull damage" 4.4
 	description "For ships with enough space to install one, the Heavy Laser Turret is a powerful weapon, equally useful for driving off fighters and for bombarding larger targets with continuous fire without having to worry about pointing your ship in the right direction."
@@ -278,8 +278,8 @@ outfit "Electron Beam"
 		"velocity" 450
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1.8
-		"firing heat" 3.1
+		"firing energy" 1.5
+		"firing heat" 2.6
 		"shield damage" 2.6
 		"hull damage" 3.5
 	description "The Electron Beam is a recent development by the Deep Sky labs, a more powerful weapon with a design similar to their perennially popular laser guns."
@@ -305,8 +305,8 @@ outfit "Electron Turret"
 		"velocity" 450
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 3.6
-		"firing heat" 6.2
+		"firing energy" 3
+		"firing heat" 5.2
 		"shield damage" 5.2
 		"hull damage" 7
 	description "This turret places two of Deep Sky's recently developed electron beam weapons onto a rotating turret, providing unsurpassed accuracy and power for shooting down fast-moving targets."
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -157,10 +157,10 @@ outfit "Beam Laser"
 		"velocity" 300
 		"lifetime" 1
 		"reload" 1
-		"firing energy" .5
-		"firing heat" 1.2
-		"shield damage" 1
-		"hull damage" 1.3
+		"firing energy" .6
+		"firing heat" 1.45
+		"shield damage" 0.9
+		"hull damage" 1.15
 	description "In the early part of the space era, the settlements in the region of space known as the Deep developed in relative isolation from the rest of human space. One result of that isolation is that their weapons technology mostly uses beam weapons, instead of the energy projectile weapons that are more common elsewhere. A Beam Laser has a slightly shorter range than an Energy Blaster, but is also much more energy-efficient and does a comparable amount of damage."
 
 outfit "Laser Turret"
@@ -184,10 +184,10 @@ outfit "Laser Turret"
 		"velocity" 300
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1
-		"firing heat" 2.4
-		"shield damage" 2
-		"hull damage" 2.6
+		"firing energy" 1.2
+		"firing heat" 2.9
+		"shield damage" 1.8
+		"hull damage" 2.3
 	description "Laser Turrets are hated by fighter pilots because it is nearly impossible to dodge them once you are within their reach. This turret carries a pair of lasers and can swivel almost instantly to fire on new targets as they approach. Laser Turrets are especially useful when mounted on slow-moving freighters to fend off packs of small pirate vessels."
 
 effect "beam laser impact"
@@ -217,10 +217,10 @@ outfit "Heavy Laser"
 		"velocity" 400
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1
-		"firing heat" 2.2
-		"shield damage" 1.8
-		"hull damage" 2.4
+		"firing energy" 1.2
+		"firing heat" 2.65
+		"shield damage" 1.6
+		"hull damage" 2.2
 	description "The Heavy Laser is an upgraded Beam Laser with a significantly longer range and higher power. It is mostly intended for larger ships, where energy and space are plentiful, but some pilots consider a single Heavy Laser to be a worthwhile alternative to a pair of Beam Lasers, because the longer range makes up for the fact that it does not quite deal twice as much damage."
 
 outfit "Heavy Laser Turret"
@@ -244,10 +244,10 @@ outfit "Heavy Laser Turret"
 		"velocity" 400
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 2
-		"firing heat" 4.4
-		"shield damage" 3.6
-		"hull damage" 4.8
+		"firing energy" 2.4
+		"firing heat" 5.3
+		"shield damage" 3.2
+		"hull damage" 4.4
 	description "For ships with enough space to install one, the Heavy Laser Turret is a powerful weapon, equally useful for driving off fighters and for bombarding larger targets with continuous fire without having to worry about pointing your ship in the right direction."
 
 effect "heavy laser impact"
@@ -278,10 +278,10 @@ outfit "Electron Beam"
 		"velocity" 450
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1.5
-		"firing heat" 2.6
-		"shield damage" 2.9
-		"hull damage" 3.9
+		"firing energy" 1.8
+		"firing heat" 3.1
+		"shield damage" 2.6
+		"hull damage" 3.5
 	description "The Electron Beam is a recent development by the Deep Sky labs, a more powerful weapon with a design similar to their perennially popular laser guns."
 
 outfit "Electron Turret"
@@ -305,10 +305,10 @@ outfit "Electron Turret"
 		"velocity" 450
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 3.0
-		"firing heat" 5.2
-		"shield damage" 5.8
-		"hull damage" 7.8
+		"firing energy" 3.6
+		"firing heat" 6.2
+		"shield damage" 5.2
+		"hull damage" 7
 	description "This turret places two of Deep Sky's recently developed electron beam weapons onto a rotating turret, providing unsurpassed accuracy and power for shooting down fast-moving targets."
 
 effect "electron impact"


Nescio0:
Yes, lasers are one of the best overall weapon choices. A damage reduction could improve balance.
However, a high energy efficiency and low heat production are fundamental characteristics of lasers in general, not just in Endless Sky. By increasing those values lasers become more similar to other weapons, which might not necessarily be a good idea. Have you considered e.g. a 20% damage reduction without energy and heat increases as an alternative?

Also, could you produce a table for human guns and anti-ship turrets to compare (x+y)/slot, (x+y)/mass, (x+y)/energy, (x+y)/heat, and (x+y)/credit, assuming both your laser and blaster proposals are implemented? [x+y shorthand for the shield and hull damage per second]

Amazinite:
My spreadsheets for everything: https://drive.google.com/open?id=0B635z_nU19WfQllrM2V2dWpFSFk
Go ahead and have a look around. If you go into the weapons folder and open the guns sheet, you'll find the stats on the changes I've proposed for both the blasters and the lasers (seen as black rows). [And just to note, I think I changed the Pug Zapper damage, firing energy, and firing heat on those spreadsheets to see how it compared to the Heavy Laser and Electron Beam without making a new row and never reverted that change.]

My argument for the firing energy and heat increase:
* Even with the 20% increase, the Beam Laser (the worst human laser) still has less firing energy and heat compared to its damage than the *best* human cannon (the Particle Cannon); the difference is about 46% with these changes as opposed to nearly 100%.
* Even if lasers are suppose to be efficient, I wouldn't think that human (tier 1) lasers would be as efficient as higher tier lasers. Currently, human lasers are actually *more* efficient in terms of firing energy and heat per damage than alien lasers. (Look at those spreadsheets and compare the current Electron Beam to the Sunbeam or Slicer; even with these changes, the Electron Beam is still more efficient by the mentioned metric, it just isn't massively better.)

gunqqer:
With balancing I prefer to instead of nerfing the good, buff the bad. That makes the transition easier, you can stay with laser and modify nothing, you try out the new and improved boasters and cannons and whatnot. This may just be me. Also only applies if there are multiple good. If one thing is OP AF, just nerfing it instead of buffing literally everything would be a better option.

Amazinite:
#3350 and #3315 buff(ed) the bad.
This PR nerfs the current no-duh choice of human lasers (which wield both massive power and efficiency).

jafdy:
Power creep is a thing to watch out for, you need to nerf some things and buff others to be able to get good balance.

ghoulavenger:
Considering the best reason to use lasers is that they fire continuously (and are thus better at disabling and not destroying) I'm not sure it matters if you nerf them or not. I'm also not too hot on the idea of a buff on one weapon and a nerf on the other at the same time -- since it is essentially a double nerf at that point. And yes, I do believe in nerfs, not just buffs in terms of balancing, because if things are all the same, then the choice doesn't matter. To make the choice matter there must be a specific reason to use the weapon over another, the obvious reason to use lasers is for the continuous fire rate of a beam weapon.

If the nerf has to happen though I'm okay with it, but I'm worried it actually just makes Hai weapons much more attractive than human ones (so rather than your intended increased usage of human cannons/blasters they just stick Hai guns on instead) -- they were already better than lasers, just not Electron Beams.

Pointedstick:
Right now there are _many_ reasons to choose lasers over blasters:
- 100% accuracy; higher _effective_ DPS
- Higher total DPS when saturating a limited number of weapon hardpoints
- Much lower energy and heat costs per DPS
- Radically higher hull DPS
- Easier to avoid accidentally destroying ships you want to plunder or capture

ghoulavenger:
@Pointedstick Sure, but is the point to be getting people to use blasters, and if so why is that? Weapons are allowed to have a progression curve, it doesn't need to be flat. So I will discount entirely anything related to the weapon being stronger, since it's entirely balanced by the fact that they are more expensive, and generally heavier.

The way I see it, blasters are meant to be an early game weapon, or used on small ships (which is why I like them on early merchantmen fleets, a lot (since it's easier to get a bigger bang on a smaller ship). They do not need to directly compete with lasers. Am I okay with a weapon in human space competing with lasers? Absolutely. But like I said before, they just need they're own niche otherwise it becomes more like what I don't want, which is that choice doesn't matter, all flavors are the same. It's bad enough that in large fleet combat, just large numbers trumps any weapons you pick for damage as it is (which is why you'd want the lasers so you kill less ships accidentally, ironically also why this nerf probably won't stop people from using lasers unless they want raw dps in which case they will go Hai instead), and you're better off picking weapons that debilitate enemies instead (sunbeams, ion cannons, plasma cannons, etc).

solardawning:
Bear in mind, a buff to blasters and nerf to lasers makes early game combat even more impossible for new players.
We've got an immense amount of foreknowledge about the game systems and mechanics, but, for a moment, pretend you've never heard of this game before, installed it, jump in and start with a Sparrow (because hunting pirates sounds fun!).

Now back in our shoes:
Do we want to make that new player's experience even more grueling than it already is?

jafdy:
If blasters are going to be a better starter weapon then we can change the stock sparrow to have blasters.

Amazinite:
>If the nerf has to happen though I'm okay with it, but I'm worried it actually just makes Hai weapons much more attractive than human ones

Yeah, I'm worried about that as well, but I'm unsure of what exactly do to about it, as nerfing pulse cannons (which I think are actually in a fairly good place balance wise) messes with the Wanderer story. Perhaps a price-hike would do? This would make it more difficult to obtain pulse cannons early on, leaving heavy lasers as a viable option, but in the long run it does not affect anything.

Nescio0:
Correct me if I'm wrong, but it seems you're looking at only two efficiencies, whereas there are actually five damage efficiencies, which all give different “best” weapons (top five):
* *slot efficiency*: sunbeam, electron beam, proton cannon, particle cannon, plasma cannon
* *price efficiency*: beam laser, energy blaster, heavy laser, modified blaster, electron beam
* *mass efficiency*: thrasher cannon, electron beam, pulse cannon, sunbeam, beam laser
* *energy efficiency*: beam laser, electron beam, heavy laser, pulse cannon, particle cannon
* *heat efficiency*: thrasher cannon, electron beam, beam laser, heavy laser, sunbeam
* *range*: inhibitor cannon, ion cannon, particle cannon, proton cannon, sunbeam
* *accuracy*: sunbeam, electron beam, heavy laser, inhibitor/particle/beam laser

Yes, the laser-like electron beams and sunbeams are overall excellent choices, however, those require lengthy campaigns before they're unlocked; the often overlooked thrasher cannon is also highly efficient.
(Note: for simplicity I omitted Korath and Heliarch weapons, because those can not be normally purchased; nor did I look at secondary weapons, although those can be highly efficient, e.g. Hai tracker; also, I didn't take special effects (e.g. heat damage) into account, which makes the Hai ion cannon such an excellent weapon.)

Although I'm not opposed to a price increase or minor damage reduction for lasers, I'm not sure your proposal is really necessary. What should be avoided is making different weapons statistically too similar. 

Besides, as others pointed out as well, the primary reason why many people (including myself) favour lasers is not really their efficiency, but rather their continuous fire rate. Changing laser statistics might not achieve a switch to other weapons.

Nescio0:
In case someone is interested, here are the efficiency values (without taking your proposed values into account):
```
  Efficiency per :  slot , credit ,   mass , energy ,  heat , range , inaccuracy
beam laser       : 138   , 4.7586 , 17.25  , 4.6    , 1.917 , 300   , 0.5
electron beam    : 408   , 2.4    , 19.43  , 4.5333 , 2.615 , 450   , 0.3
energy blaster   :  75   , 4.6875 , 15     , 1.5    , 0.429 , 420   , 3
heavy laser      : 252   , 4.2    , 16.8   , 4.2    , 1.909 , 400   , 0.4
inhibitor cannon : 210   , 0.4459 , 13.125 , 1.75   , 1     , 864   , 0.5
ion cannon       : 228   , 0.4653 ,  4.851 , 1.9    , 1.52  , 800   , 1
modified blaster : 100   , 2.8571 , 16.667 , 1.25   , 0.364 , 450   , 4
particle cannon  : 318   , 1.4455 , 10.6   , 3.0286 , 0.883 , 750   , 0.5
plasma cannon    : 293.3 , 1.5437 , 11.732 , 2.933  , 0.489 , 480   , 2
proton cannon    : 346.5 , 2.31   , 12.375 , 3.08   , 0.77  , 624   , 4
pulse cannon     : 237   , 1.8231 , 18.231 , 3.2917 , 1.129 , 450   , 1
sunbeam          : 612   , 0.4403 , 18     , 1.9615 , 1.645 , 540   , 0.2
thrasher cannon  : 252   , 1.1560 , 25.2   , 2.3333 , 3.111 , 360   , 9
```

Pointedstick:
Nobody is trying to make Blasters and Lasers similar to one another. What we're trying to do is create a genuine role for the humble blaster beyond "terrible weapon that you look down on once you've realized that lasers are just better 95% of the time."

Nescio0:
> What we're trying to do is create a genuine role for the humble blaster beyond "terrible weapon that you look down on once you've realized that lasers are just better 95% of the time."

No objections, nor do I oppose #3350 or #3393; however, even if these tweaks are implemented, lasers remain about always better than blasters, so why are these proposed changes imperative?

Personally I consider laser efficiency not really a serious problem, however, I think they're too cheap. A beam laser is cheaper than a modified blaster and an electron beam is cheaper than a plasma gun. Suppose lasers would become about 50% more expensive, e.g.:
* beam laser: from 29 to 42 thousand credits
* heavy laser: from 60 to 85 thousand credits
* electron beam: from 170 to 250 thousand credits

Then beam lasers would become slightly more expensive than modified blasters and electron beams would cost somewhat more than particle cannons. As a result blasters suddenly are most efficient in one aspect: they would have highest damage per credit.

Amazinite:
>Correct me if I'm wrong, but it seems you're looking at only two efficiencies,

I'm ignoring price, accuracy, and, to an extent, "slot" efficiencies. At the end of the day, the price of a weapon doesn't really matter once a player has large amounts of credits and they can buy anything they want. Accuracy only matters at large values and long ranges; on accuracy of 0 isn't very different from 1 or 2 at the ranges that weapons with those inaccuracies fire at. Inaccuracy only really becomes something to worry about beyond 3 degrees, at ranges beyond say 600 or 700, when fighting smaller ships, and very few weapons fit that bill, leaving alone the fact that the detriments of inaccuracy can be avoided by getting closer. 

Leaving those two alone, human lasers come out on top by your terms in terms of heat efficiency, energy efficiency, and mass efficiency, while cannons only come out on top in terms of range and slot efficiency. Once you take into account the fact that lasers also allow for easy disabling and therefore easily getting credits through capturing, there's really little reason to use cannons given the ease of use and return on investment of lasers. 

The reduction of firing energy and heat lowers the efficiency of lasers by those metrics, but they still come out on top. Reducing the damage of lasers also still leaves them with high mass efficiency, but if you look at weapons in terms of BOTH damage and range, the slight change in damage actually takes human lasers and puts them below human cannons (even if only slightly in some cases), giving cannons a better reason to be used.

>Yes, the laser-like electron beams and sunbeams are overall excellent choices, however, those require lengthy campaigns before they're unlocked; the often overlooked thrasher cannon is also highly efficient.

I'm not looking at alien lasers in this, so the Sunbeam is out of this. I've also mentioned the efficiency of the thrasher cannon (https://github.com/endless-sky/endless-sky/pull/3307#issuecomment-348807522). I also mentioned above how the Electron Beam is fine being powerful given that it's unlocked through the campaign. I mentioned how the reduction in damage puts lasers just below cannons in terms of both range and damage (per ton), and in the case of the Electron Beam this puts it about on par with the Particle Cannon, but the Electron Beam still has the slot, heat, and energy efficiency to beat out the Particle Cannon.

>Although I'm not opposed to a price increase or minor damage reduction for lasers, I'm not sure your proposal is really necessary. What should be avoided is making different weapons statistically too similar.

This PR is making human weapons statistically closer (which isn't necessarily a bad thing), but no where near statistically similar. Human lasers, cannons, and blasters are still within three distinct groups of statistics, it's simply that this change brings human lasers down a notch given how powerful they are compared to other human weapons at the moment.

Characteristics of blasters:
* Small size
* High power density
* Low range
* Extremely low heat and energy efficiency
* Moderate accuracy
* Moderate rate of fire

Characteristics of lasers:
* Medium size
* High power density
* Low range
* Extremely high heat and energy efficiency
* High accuracy
* Continuous rate of fire

Characteristics of cannons:
* Large size
* Low power density
* Medium to high range
* Moderate heat and energy efficiency
* Low to high accuracy
* Low to high rate of fire

<hr/>

>Nobody is trying to make Blasters and Lasers similar to one another. What we're trying to do is create a genuine role for the humble blaster beyond "terrible weapon that you look down on once you've realized that lasers are just better 95% of the time."

This PR is more about the role of cannons in relation to lasers than blasters in relation to lasers to me.

>however, even if these tweaks are implemented, lasers remain about always better than blasters, so why are these proposed changes imperative?

Again, I'm not here to make blasters more usable with this PR.

KiLEdEnNis:
> will players actually choose non-laser weapons? If not, then why these changes?
first of all, more players will because it'll be more evenly balanced, and "why these changes?" because when I was a new player I quickly found out Plasma Cannons & Heavy(/Electron) Lasers in general were the human no-brainers (their best weapons by quite a bit), and basically every other notices this as well and only gets these weapons, because compared to the rest they're simply not balanced

ghoulavenger:
I think one of the bigger problems with cannons is that most of them aren't turret mountable. Giving the proton gun a turret mount makes it significantly better, although this isn't exactly as effective as giving the particle cannon a mount (which actually has the range advantage over lasers). That would make the particle cannon the hit and run weapon for fast ships (granted, not nearly as much range as missiles, and ships with the particle cannon if fast enough are going to be a lot more annoying than a missile variant ship running away from you), rather than simply a range hike (guns only fire forward, but turrets fire in all directions).

And like I said earlier, I'm okay with this nerf if it has to happen I'm just worried that it'll make the Hai weapons the obvious choice instead. Or do nothing since continous fire is used primarily because it makes it easy to board ships.

Amazinite:
>I'm just worried that it'll make the Hai weapons the obvious choice instead.

I don't want this PR merged as-is, as I still need to change some of the variants that use lasers and look at how the change pans out across all of T1. As mentioned, I'll try to alleviate the issue with Hai Pulse Cannons becoming too much of a new "no-duh" choice.

>Or do nothing since continous fire is used primarily because it makes it easy to board ships.

Players that use lasers for this purpose will likely still be using lasers, simply because the return on investment aspect of lasers via disabling and capturing is still there. But even outside of this aspect, lasers are still extremely appealing weapons for other reasons as mentioned already (high power density and high heat/energy efficiency). These changes ultimately make human cannons more appealing by making it so that lasers aren't over *twice* as efficient as cannons and by causing their power in terms of both DPS/ton and range to become close to on-par with each other, making both types of weapons viable options given the space (which I personally often find that I have), but making it more of a question of whether you want more DPS or more range. So no, players who use lasers ONLY because of their continuous fire will likely not change their use of lasers, but players who use lasers for other reasons (their high damage and high efficiency) may find themselves looking more toward cannons, especially the newly buffed Proton Gun and its Proton Turret addition.

<hr/>
>assuming your proposals are implemented, will players actually choose non-laser weapons? 

Well really, we can never know given the fact that this is a single-player game. All we can do is ask around if we want to find out. Personally, I will find myself using cannons a lot more often after this change. Others can chime in with their own thoughts if they want.

>If not, then why these changes? 

Balance.

>And is it a serious problem lasers are popular?

The popularity of lasers isn't the problem, it's the reason they're popular, that reason being that they are very much a "no-duh" choice of weapon. There is currently little reason to choose cannons over lasers, except for perhaps the mentioned special effects of the Plasma and Particle Cannon, as well as the ability to fill slots (which on some ships, especially heavy warships, is not a problem at all with lasers to begin with). The ability to fill slots with cannons or lasers can be very much a ship-by-ship basis, but from asking around and just through general observations on Steam and the Discord over the nearly two years I've been playing the game, it seems that lasers are what occupy most people's weapon space anyways, and filling weapon space with lasers, especially when turret mounts come in to play, isn't too difficult on the ships that people use.

Amazinite:
>>I'm just worried that it'll make the Hai weapons the obvious choice instead.

>I don't want this PR merged as-is, as I still need to change some of the variants that use lasers and look at how the change pans out across all of T1. As mentioned, I'll try to alleviate the issue with Hai Pulse Cannons becoming too much of a new "no-duh" choice.

I'm unsure of just how much the Pulse Cannon would need changed with this PR. Looking mainly at the Pulse Turret and Heavy Laser Turret with the changes in this PR in order to see how much more viable Pulse Turrets becomes (as I'm going off of the assumption that the changes to the Heavy Laser Turret will be more consequential to the use of the Pulse Turret than the changes to the Heavy Laser will be to the use of the Pulse Cannon):
* The Heavy Laser Turret is 17.44% more efficiency in terms of both heat and energy efficiency
  * This breaks down into 3.95% *less* energy efficient *but* 27.06% *more* heat efficient.
* In terms of "slot" efficiency, the Pulse Turret is only 3.95% stronger
* In terms of DPS * Range (a combo of slot and range efficiency), the Pulse Turret is 16.94% *better* than the Heavy Laser Turret
  * This difference of the Pulse Turret being about 17% better by this metric interestingly coincides with the Heavy Laser Turret being about 17% better by a different metric, which is mainly why I question if the Pulse Cannon/Turret need looked at...

... This change might actually make the choice between Heavy Laser Turret and Pulse Turret more interesting, having the choice be between efficiency of firing energy and heat and efficiency of DPS over range, both metrics that alter the use of these turrets in combat; I certainly know from using Pulse Turrets that firing heat is their biggest drawback, and that remains to case now, being the major detriment I would find to using Pulse Cannons over Heavy Laser Turrets. This really breaks down across play style choices; I prefer to be able to run sustainable builds that can hold their own for a long amount of time (or even indefinitely if possible) while in combat, so I would choose a Heavy Laser Turret, but a player looking to kill their opponent as fast as possible with little regard for how sustainable their build is (in this case in terms of overheating) will go for the about 4% increased damage that the Pulse Cannon gives over the 12.5% increased range.

**tl;dr I'm unsure that the Pulse Cannon and Turret need any changes. Please voice any disagreement along with your reasoning though.**

KiLEdEnNis:
to me the pulse turrets seem a little bit too strong too, maybe make them require a bit more energy?
any rebalancing on them may mess up #3405, but that mission is more difficult than the last pre-Eye defense battle anyway so it shouldn't be that bad

Nescio0:
Personally I don't think the pulse weapon has to be changed. With or without this patch, the pulse already has slightly better range, mass efficiency, and higher shield damage, whereas the heavy laser has better hull damage and heat efficiency.

A massive switch from heavy laser to pulse seems unlikely; most people choose lasers for their continuous fire, which the pulse lacks; and if those who have different reasons now consider the pulse a more interesting alternative, I won't perceive that as a serious problem.

KiLEdEnNis:
I personally wouldn't mind a small Pulse Turret nerf, but I'll leave that up to others

Amazinite:
>Please voice any disagreement though with your reasoning as to why you disagree.
>with your reasoning as to why you disagree.
>with your reasoning

@KiLEdEnNis :)

KiLEdEnNis:
from the current Steam version, I'll take the Plasma Turret as a comparison
average DPS (shield+hull)/2
Plasma Turret: 293,33... (4,88... per ton)
Pulse Turret: 237 (6,97 per ton)
Pulse Turret has approximately 42,8% more DPS than the Plasma Turret

firing energy & heat per second
Plasma Turret: 200 (3,33.../t) & 1200 (20/t)
Pulse Turret: 144 (4,24/t) & 420 (12,35/t)
Pulse Turret uses ~27,2% more energy, but only 61,75% of the heat, meaning it's a bit more efficient

extra: the Plasma Turret does have 2.666,66... (44,44.../t) heat damage and 30 more range
Pulse Turret has 240 degree turning, Plasma Turret 120

basically, the Pulse Turret has a very high DPS/t(on), and is a little bit more efficient, it doesn't have the heat damage and has 30 less range, but also double the turning which is also a balance point now
not that the Plasma Turret is straight-up the best human turret, but it's basically top 3

I also didn't mean the Pulse Turret should get a big nerf, but it should either do a little bit less damage (maybe around 10%), or have more firing energy or heat (also about 10%) in my opinion

Amazinite:
I'd say that any comparisons with the Plasma Turret in terms of DPS are unfair given that most of the Plasma Turret's power comes from its heat damage and not its shield or hull damage. I did the comparisons between the Pulse and Heavy Laser Turrets already, and like I said, the Pulse Turret *with* these changes only does about 4% more DPS than the Heavy Laser Turret (23% more DPS/ton if we're looking at power density, but that's to be expected with the Hai's T1.1/1.2 status). The difference between the Pulse Turret and Heavy Laser Turret with this PR are already interesting enough (17% more combat power with the Pulse Turret and 17% more system drain efficiency with the Heavy Laser Turret; it becomes a choice of play style instead of a choice of true power, making the Pulse Turret an alternative instead of an upgrade), and any changes to the Pulse Turret would ruin this.

KiLEdEnNis:
I trust your facts, and
> it becomes a choice of play style instead of a choice of true power, making the Pulse Turret an alternative instead of an upgrade

is what I wanted (since Hai tech is easily accessable), so seems good to me

endless-sky:
When adjusting balance, I'd prefer not to make a ton of changes all at once. The changes that I'm leaning toward right now are:
1. Reducing the damage of the Heavy Laser and Electron Beam, without adjusting their energy use or heat.
2. Increasing the Blaster range (maybe by increasing speed rather than lifetime) and decreasing its heat production.

That would address the problem of electron beams being way too close to a T2 weapon right now, and would create a potential distinct use case for blasters, so it's not just a choice between a short-range weapon that shoots projectiles or a short-range weapon that shoots beams and does equivalent damage.

Amazinite:
Any specific numbers on those changes? A decreased in blaster firing heat could be all good, but if it's only firing heat being reduced and that reduction is any less than say 40% then I don't see it making much of a difference. 

(And just going to point out something that peeves me, the Beam Laser is currently more power dense than the Heavy Laser, 17.25 combined dps/ton vs 16.80, and a reduction in Heavy Laser damage would make that divide even larger and more odd. Literally the only advantage of the Heavy Laser over the Beam Laser is that it's bigger, which is generally odd for weapon progression where the larger weapon has more than its size that makes it better than the smaller weapon. E.g. the Modified Blaster is more power dense than the Energy Blaster, or the Particle Cannon has more, range, accuracy, and hit force than the Proton Gun.)

Nescio0:
> When adjusting balance, I'd prefer not to make a ton of changes all at once. The changes that I'm leaning toward right now are:
> Reducing the damage of the Heavy Laser and Electron Beam, without adjusting their energy use or heat.
> Increasing the Blaster range (maybe by increasing speed rather than lifetime) and decreasing its heat production.

Yes, I fully agree with all this.

> Any specific numbers on those changes? 

Maybe decrease (beam, heavy, electron) laser shield damage by 20%? A laser price increase would be nice as well.
As for blasters, my suggestions are in #3426

Pointedstick:
Blasters getting better range makes sense to me. Range is an under-appreciated weapon attribute, but it really starts to come into its own around range > 500.

Further biasing lasers toward hull damage rather than shield damage also makes sense, as it further distinguishes the weapons from one another.

Nescio0:
> Further biasing lasers toward hull damage rather than shield damage also makes sense, as it further distinguishes the weapons from one another.

And it encourages players to equip their ships with combinations of different weapons.

KiLEdEnNis:
lasers can get a ship to just the disabled threshold, while any projectiles have the risk of destroying it on autofire, making lasers more anti-hull & projectiles more anti-shield makes this difference a bit less, so people that capture ships mak consider non-lasers a bit more

Amazinite:
On the topic of shifting the damage profile of human lasers, they currently all have around 0.75 shield damage / hull damage. Since most ships have more shields than hull, shifting this number closer to 0 will reduce the time to kill (TTK) of laser weapons without reducing their combined DPS. 

Lets say we're using a single Electron Beam to kill a Cruiser with no shield generators (for the sake of simplicity). That single beam would take almost exactly 140 seconds to kill the Cruiser. Any % change to damage will change the TTK by that percentage (e.g. reducing Electron Beam damage across the board by 10% will cause the TTK to increase by 10%, from 140 to 154 seconds). But shifting more damage from hull to shields will have the same effect on TTK without compromising how powerful lasers actually are. Shifting the Electron Beam's shield/hull from 0.74 to 0.58 (shield damage from 2.9 -> 2.5 and hull damage from 3.9 -> 4.3) will cause the same (nearly) 10% increase in kill time, effectively nerfing the usefulness of having ONLY lasers, but opening up the door to installing weapons with more shield damage, as Electron Beams would still retain massive hull damage

Does anyone have an ideas one what the shield/hull damage profile of human lasers should become? Currently being around 0.75, should they all get closer to 0.6 or to 0.5 (the latter being double the hull damage as shield damage)?

endless-sky:
In beb7f03 I've increased the range of blasters and decreased their energy and heat requirements, which has a similar effect as this without the side effect of a bunch of stock ship models starting to overheat.

Issue #3394, opened 2017-12-10
Pluggins Names can overlap text box


mdsmestad:
Just put the T's there to test it out. Here's an example
![screenshot from 2017-12-10 17-12-02](https://user-images.githubusercontent.com/16338814/33810680-8130fdfc-ddcd-11e7-9d77-bc67d506d4e8.png)

tehhowch:
Should be an easy fix with Font::TruncateMiddle function.

Though, is the preferred behavior to truncate it, or scroll it?

mdsmestad:
@tehhowch Truncate would look better I think

Amazinite:
Essentially a duplicate of #2022

tehhowch:
I had done some of the needed truncation in #3070 but that's currently a ways from merging as I need to come up with a nice way to convey the per-system stellar wind and luminosity differences.

I won't add scroll to the plugin's about text but I can add the needed truncation to the plugin name.

KiLEdEnNis:
I remember MZ said one should just pick a shorter name (if ya get rid of the 2nd g in "pluggin" it'd be fine)
the only decent fix I can come up with (and I think that's what Tehh meant?) is to get the name cut off (end with a ...) and/or have a side scrolling bar at the bottom
example:
random-plugin-name-that's-unnec...
and if ya scroll ya may see ...ame that's-unnecesarily-long
that's not really a nice change though as it looks a bit... clunky

PR #3395, opened 2017-12-11
Truncate long player / player flagship names, and plugin names

tehhowch:
Refs #3394

### DIFF
--- a/source/PreferencesPanel.cpp
+++ b/source/PreferencesPanel.cpp
@@ -543,7 +543,7 @@ void PreferencesPanel::DrawPlugins()
 	table.Draw("Installed plugins:", bright);
 	table.DrawGap(5);
 	
-	const int MAX_TEXT_WIDTH = 235;
+	const int MAX_TEXT_WIDTH = 230;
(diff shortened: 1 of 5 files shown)


source/PreferencesPanel.cpp
```diff
@@ -561,10 +565,10 @@ void PreferencesPanel::DrawPlugins()
 				top.Y() += sprite->Height() + 10.;
 			}
 			
-			WrappedText wrap(FontSet::Get(14));
-			wrap.SetWrapWidth(230);
-			static const string empty = "(No description given.)";
-			wrap.Wrap(it.second.empty() ? empty : it.second);
+			WrappedText wrap(font);
+			wrap.SetWrapWidth(MAX_TEXT_WIDTH);
```
endless-sky: The panels in the settings screen are 250 pixels wide, designed to have a 10 pixel margin for the text on each side. This would reduce the right margin to 5 pixels, putting the description text off-center.

tehhowch: Oops, missed that 230->235 switch.
Addressed.

PR #3405, opened 2017-12-13
balance change to "Wanderers Defend Sich'ka'ara"

KiLEdEnNis:
so, due to the new turret turning angles we knew the combat missions would require some rebalancing, and I said I'd test them in my new playthrough (playing in 1 ship for fun & a challenge so might as well)

I've done this mission nearly 10 times in the old (current) state, and it was basically undoable in my Albatross (yes, the mission is balanced on having a Tempest, but apart from lower durability they're surprisingly even), so I tried adding a Tempest in to make it a little bit easier

the Albatross loadout I tested it in:
albatross loadout 2

I tried the mission a handful of times AFTER adding this Tempest, and cloaked the full battle to see how the fleets would do without the player; around half the Hai ships make it through

then I tried it a handful of times by focusing the same ships the Wanderers focused (teamwork), and maybe 2 Tempests and 1-2 Summer/Autumn leaves make it through at most, I was quite surprised how much of an impact the player has when using a proper strategy

I also considered maybe adding 1 or 2 Summer Leaves in for good measure, what do ya all think?

### DIFF
--- a/data/wanderers start.txt
+++ b/data/wanderers start.txt
@@ -2002,7 +2002,7 @@ mission "Wanderers Defend Sich'ka'ara"
 			names "hai"
 			variant
 				"Shield Beetle" 2
-				"Shield Beetle (Tracker)" 1
+				"Shield Beetle (Tracker)"
 				"Shield Beetle (Pulse)" 2
 				"Water Bug" 3
 				"Water Bug (Pulse)" 4
@@ -2012,7 +2012,7 @@ mission "Wanderers Defend Sich'ka'ara"
 		fleet
 			names "hai"
 			variant
-				"Shield Beetle" 1
+				"Shield Beetle"
 				"Shield Beetle (Tracker)" 3
 				"Shield Beetle (Pulse)" 2
 				"Water Bug" 7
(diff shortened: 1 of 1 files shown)


KiLEdEnNis:
anyone else that want to test it too btw? I can share my save file so all ya have to do is add that Tempest in the data files

tehhowch:
How's this mission play out if you use a Checkmate build and not a Reconciliation build?

KiLEdEnNis:
I haven't used my cloak in the battle itself, so the difference should be next to nothing

tehhowch:
I imagine unleashing a Nuclear Missile or 3 will heavily alter the outcome of the battle.

KiLEdEnNis:
unless ya get lucky ya usually hit some allies as well, and I don't think using nukes should really have an impact on mission balancing in the first place, considering they are extremely outbalanced (heavily altering the results of 1 battle)

SirStigII:
The problem with using nukes is that you fire them all at once, so instead of being able to load up a ship like the albatross and judiciously using them to weaken ships, you get all or nothing. Checkmate would be more compelling if there was a toggle on how missiles fired: all at once, or one at a time.

KiLEdEnNis:
either way I think nukes should not influence mission balance due to their outbalanced nature

endless-sky:
Given that the turret turn rates will probably be decreased across the board some time soon to better differentiate the anti-capital-ship ones from the anti-fighter ones, it might be premature to start re-balancing battles.

KiLEdEnNis:
in that case I'll make a backup save for the apparent 2 hardest battles (this one and the Sestor attacking the Wanderers), so I can easily balance test both of them after that's been done

I'll close this & make a new PR if needed after that

PR #3406, opened 2017-12-13
tweaked heavy laser and electron beam ranges

Nescio0:
This patch is related to #3350, #3393, and #3426, but can be implemented entirely independently from those. What this proposal changes:
* beam laser velocity unchanged, hence range at 300
* heavy laser velocity hence range from 400 to **360**
* electron beam velocity hence range from 450 to **420**

Effectively it results into laser ranges < blaster ranges < cannon ranges. This makes lasers slightly less attractive and other weapons more interesting. Furthermore, it increases weapon diversity without breaking any default or variant ship builds.

EDIT: for comparison other gun ranges:
* Remnant thrasher: 360
* blaster: 420
* modified blaster: 450
* Hai pulse: 450
* plasma: 480
* Wanderer moonbeam: 490
* Wanderer sunbeam: 540
* proton: 624
* particle: 750
* Hai ion: 800
* Remnant inhibitor: 864

Shortened images (`xcf`, `@2.png`, and `png`) are located at https://github.com/Nescio0/sprites


Pointedstick:
I don't think the Plasma Gun needs a range increase; it's already among the best of the Human big guns. Also 380 might be more reasonable for the Heavy laser; 360 seems a bit too short. Otherwise, I like it.

KiLEdEnNis:
keep in mind, altering the range of a laser-based weapon means altering the "projectile" itself in GIMP, as it's a 1-frame projectile every frame that cycles between the images to animate
so unless ya find someone to alter the laser sprites or do it yourself, this isn't merge ready

Amazinite:
I'm unsure on the increased blaster velocity given the reaction to #3307. And like @Pointedstick, I don't think the Plasma Cannon needs altered. I'll have to look more closely at the range changes later to really give an opinion on those though.

Nescio0:
> I don't think the Plasma Gun needs a range increase;

Without this patch, the plasma gun outranged modified blasters by 30, with this patch by 20. In all other situations, the range increase of 4% is hardly noticeable. I could revert it, but then plasma guns would get effectively the same range as blasters.

> so unless ya find someone to alter the laser sprites or do it yourself, this isn't merge ready

Thanks for pointing it out! I have GIMP and could shorten those images myself, but where do I upload the proposed edited sprites then to?

tehhowch:
Altered sprites have "1x" components that are done here, "2x" components that go to the High-DPI plugin, and you can store the updated asset files here or on google drive and link them here.

Nescio0:
> I'm unsure on the increased blaster velocity given the reaction to #3307. And like @Pointedstick, I don't think the Plasma Cannon needs altered. I'll have to look more closely at the range changes later to really give an opinion on those though.

Reverted minor plasma and blaster velocity increases, replaced them with a +2 blaster lifetime instead.

> you can store the updated asset files here

Thanks, I'll do that later this evening.

KiLEdEnNis:
a tip: edit the @2x version only, and once done make a copy that's scaled 50% and use that smaller one in your PR, PR the @2x version for the high-dpi plugin
and make a google drive folder (make sure that it's not private), with the new GIMP file and both the @2x & normal version in them, and put said link in this PR, so MZ has easy access to the new files should it get merged

Nescio0:
> a tip: edit the @2x version only, and once done make a copy that's scaled 50% and use that smaller one in your PR, PR the @2x version for the high-dpi plugin

Are you sure? The 2x versions seem to have two beams each instead of one, they are not really double resolution files:
* `laser`: 20×1200, `heavy laser`: 20×1600, `electron`: 20×1800
* `2x laser`: 30×1200, `2x heavy laser`: 34×1600, `2x electron`: 32×1800

KiLEdEnNis:
the 2x are the double beams, the @2x ones are the high dpi ones (although now I see it they're not in the folder I linked you), although if you export and laser beam in the GIMP file it should automatically be the @2x size anyway, so just including the GIMP & default size would do too

Nescio0:
> the 2x are the double beams, the @2x ones are the high dpi ones (although now I see it they're not in the folder I linked you), although if you export and laser beam in the GIMP file it should automatically be the @2x size anyway, so just including the GIMP & default size would do too

Now I'm confused. Do I only have to shorten the 2×2=4 GIMP files in the folder you linked, or also other files? And if so, where are those located?

jafdy:
You need to shorten the normal ones and the ones that end in @2x. Kile is saying that that it is easier to shorten the images that end in @2x and then reduce them to normal size.

Nescio0:
> You need to shorten the normal ones and the ones that end in @2x. Kile is saying that that it is easier to shorten the images that end in @2x and then reduce them to normal size.

Yeah, but there are no `@2x` files in the linked folder, only `2x` files

Nescio0:
Anyway, the shortened `xcf` files can be found at:
https://github.com/Nescio0/sprites
(Github only supports a few file types to be attached directly to this post, GIMP files are not allowed.)

tehhowch:
The high-dpi versions (e.g. `imagename{blendmode}@2x.png`) are located in MZ's high-dpi plugin as I mentioned earlier.
You'll need to include the normal versions of the modified sprites in this pull.

Nescio0:
Do you mean the `images/projectile/*.png` files? Are those simply the `xcf` layers exported at 50%?

KiLEdEnNis:
if ya open a .xcf file, it's probably the exact size a @2x export would be, so just change it & export, then size the exported .png down to 50%, so ya still have the ability to render the @2x version from the .xcf

Nescio0:
> if ya open a .xcf file, it's probably the exact size a @2x export would be, so just change it & export, then size the exported .png down to 50%, so ya still have the ability to render the @2x version from the .xcf

Yeah, I understand the scaling part, my question is about the export part.
E.g. in `images/projectile` there are *three* 2x electron files, all 16×900 (to be 16×840):
* `2x electon+0.png`
* `2x electon+1.png`
* `2x electon+2.png`
The *single* `2x electron.xcf` file contains *fifteen* layers, all 32×1800 (changed to 16×1680). 
Now which layers do I have to export exactly for each of the `png` files?

tehhowch:
You'll need to figure out which layers are part of which animation frame and show only those layers during the export.

Nescio0:
> You'll need to figure out which layers are part of which animation frame and show only those layers during the export.

That is straightforward for the `heavy laser.xcf` file, which has only three layers (background, frame 0, frame 1), but the electron files have numbered wave and beam layers, but also unnumbered background and blasts; where to include those?

Furthermore, upon exporting, GIMP asks for compression level etc.; are the default settings acceptable?
screenshot from 2017-12-13 21-20-05

tehhowch:
You'll have to do the legwork by comparing the current animation frames with the available layers.
I know the Sunbeam had a layer that was always on and then pairs of layers for each frame, and I imagine something similar for the others.

I use the defaults for png export.

Nescio0:
Thanks. Assuming the layers are in the correct order, which visually seems to match the existing `png`, then here are the resulting `@2` files:
heavy laser 0 2
heavy laser 1 2
2x heavy laser 1 2
2x heavy laser 0 2
electron 0 2
electron 1 2
electron 2 2
2x electron 1 2
2x electron 0 2
2x electron 2 2

KiLEdEnNis:
the heavy laser frames, especially the base ones, seem to become quite a bit brighter, did you miss a layer for export?

Nescio0:
It seems you're right, they are fainter indeed. Given that the heavy lasers have only two layers (frame 0 and frame 1), it is impossible to forget one.
To find out what's the problem, I downloaded the unchanged `xcf` from the drive you linked earlier, and exported the layer from there. Result: it's just as bright as the exported png from my edited file, and visibly different from `images/projectile/heavy laser.png`. Conclusion, there are only two explanations:
* the `xcf` files in the drive you linked are **not** the source files of the sprites in this repository
* the default GIMP export settings are incorrect

Amazinite:
Looking at this more closely, here are my thoughts.
* I'm indifferent on the blaster changes, but I see the point of them if it's for the sake of putting blaster range between laser and cannon range.
  * As an aside and in relation to the point I'm about to make, I think it's fine if the Electron Beam has more range than the blasters given the Electron Beam's campaign unlock status and the fact that it's essentially suppose to be the pinnacle of human weapons tech, and the current blaster ranges already beat out the Heavy Laser, hence why I'm generally indifferent to the change.
* I think that instead of reducing the range of lasers, you should increase the range of the Plasma Cannon (suggested just above 500 range) if you're trying to increase weapon diversity.
  * My reasoning for this stems from discussion on #3393. Ghoul has pointed out his concern with how that PR might cause Hai Pulse weaponry to become a no-duh choice that simply replaces the current role of the Heavy Laser Turret. Through further analysis, I found that with that PR, the Heavy Laser Turret and Pulse Turret have interesting trade offs between each other. Summarizing here, those trade offs are the fact that the Heavy Laser Turret is about 17% more efficient while the Pulse Turret has about 17% more DPS * range (and about 4% more DPS in general). These almost equal percentage differences in two very interesting categories leaves the choice between Heavy Laser Turrets and Pulse Turrets up to play style. A nerf to Heavy Laser range down to 360 would ruin this situation, as the Pulse Turret would go from being 17% better in terms of DPS * range to 29.93% better, a difference that would make the Pulse Turret a *very* clear choice over the Heavy Laser Turret given that both this PR and #3393 are merged, instead of the interesting choice created by the merging of only #3393.

Nescio0:
> if it's for the sake of putting blaster range between laser and cannon range. 

Yes, as I wrote earlier:
“Effectively it results into laser ranges < blaster ranges < cannon ranges. This makes lasers slightly less attractive and other weapons more interesting. Furthermore, it increases weapon diversity without breaking any default or variant ship builds.”

> you should increase the range of the Plasma Cannon (suggested just above 500 range)

Actually I did, which I subsequently reverted because of your earlier post (see above.

> reducing the range of lasers

Right now there is a huge jump from beam to heavy laser (300 to 400), and a smaller to electron beams (450). With this patch the heavy laser nicely fits in between. 
You also mentioned “DPS * range” values; I consider these rather meaningless, for several reasons:
* it assumes area damage, which contradicts how the game works
  * e.g. if there are three ships on a line within range, only the first takes damage
* damage is either full (1) or none (0), not something in between
* range matters in one ship vs one ship combat, but is largely irrelevant in many vs many fights

However, for the sake of argument, let's look at laser “DPS * range” values:
* beam laser (unchanged): (60+78)*300=41400
* heavy laser (currently): (108+144)*400=100800
* heavy laser (with this patch): (108+144)*360=90720
* electron beam (currently): (174+234)*450=183600
* electron beam (with this patch): (174+234)*420=171360

Or, to put it differently, beam:heavy:electron was about 4:10:18 and becomes 4:9:17. And if we divide the values by the mass (thus calculate DPS*range/mass) we get beam stays at ~5k, heavy from ~6.7k to ~6k, electron from ~5.1k to ~4.7k. With or without #3393, these ratios remain the same.
These range reductions might seem a lot on paper, but effectively they won't change much. The effect of this patch is that the heavy laser becomes about twice as good as the beam laser, instead of clearly more than twice as good.

About pulse cannons, I quote myself from #3393: 
> A  massive switch from heavy laser to pulse seems unlikely; most people  choose lasers for their continuous fire, which the pulse lacks; and if  those who have different reasons now consider the pulse a more  interesting alternative, I won't perceive that as a serious problem.

A larger range difference would make the pulse potentially more interesting, yes, however, I still believe many players would continue to use the shorter range lasers.

Nescio0:
> the heavy laser frames, especially the base ones, seem to become quite a bit brighter, did you miss a layer for export?

The problem is that it appears the xcf and high resolution sprites are different from those in this repository. Here is a comparison, from top to bottom: current default `images/projectile`, exported from `xcf`, shortened proposal; current high resolution, exported from `xcf`, shortened proposal:
heavy laser 0 comparison

Now, which files should I use as the source for the shortened sprites of this proposal? The GIMP files you linked earlier, in which case the colour changes? Or the current png files, in which case there remains a colour difference between the default distribution on the one hand, and the high resolution plug-in on the other?
Or could someone explain how to produce the current image in this repository from the high resolution GIMP file?

endless-sky:
In beb7f03 I decided it made more sense to increase the blaster ranges rather than decreasing laser ranges.

Issue #1626, opened 2016-09-25
The Corvette, Firebird, and Osprey are too similar to one another


Pointedstick:
This is a follow-on to issue https://github.com/endless-sky/endless-sky/issues/1598, which made me realize that the Mule's superiority is exacerbated by issues with a number of other Medium Warships, which are not as good as they should be. In this issue, I will focus on three in particular that seem insufficiently differentiated from one another and not good enough either: the Corvette, Firebird, and Osprey.

These ships are very similar. They have identical weapon hardpoints, similar overall durability (shields+hull), crew complement, space for cargo, outfits, weapons, and engines. Differences are minimal: the Firebird is a little less maneuverable than the other two and has lower range but more hull, and the Osprey has more shields and is grossly overpriced. The Corvette is also overpriced and doesn't really stand out in any way at all--not even matching its description, which includes the following:

> Its weak hull is more than compensated for by its advanced shielding matrix.

…But it has lower shields than the Firebird (an older design!), and it doesn't come with an unusually powerful shield generator like the D94-YV (It only has the D67-TM, which is carried by other Medium Warships).

These three ships need to be better differentiated from one another, and probably a bit better in general to make them more competitive with the Bastion, Manta, and especially the Mule.

LocalGod79:
i think it makes sense that the Syndicate, Betelgeuse and Tarazed shipyards would have competitors that were very similar to each other...

Pointedstick:
Yes, but it's still a missed opportunity to differentiate them gameplay-wise, and there's also no in-world reason why they still couldn't be broadly similar with each having a specific obvious advantage versus the others--think of the advertising potential! For example the Corvette could be even faster and have greater range; the Firebird could have even more hull durability and cargo space; and the Osprey could have even more shields.

MessyMix:
Agreed! I can try work up a solution. I think that the Bastion and the Mule are paragons of the Medium Warship class, so if we can buff the others to be close to—but not perfectly at—their level, we can have more warship to the two most obvious choices at the moment. However, considering the recent price hike to the Mule, we should keep it as a superior choice, reasonably justified by its higher cost.

LocalGod79:
are you guys going to play test the FW storyline to make sure that your Firebird/Osprey tweaks didn't mess with any of the battles, too?

MessyMix:
I will when I have time to do so, or I might get someone to do it for me. Do you think a steam thread would do any good?

Pointedstick:
[...]

PR #3411, opened 2017-12-14
Differentiated the Corvette and Firebird a bit more

Pointedstick:
This PR differentiates the Corvette and Firebird from one another, as right now they are very similar (see https://github.com/endless-sky/endless-sky/issues/1626). Changes:

Corvette
----------
- Increased shields, gave it a more powerful shield generator, and reduced hull, to better match its price and the description ("experimental lightweight materials", "its weak hull is more than compensated for by its advanced shielding matrix")

Firebird
----------
- Increased hull and reduced shields to better match the description ("long and storied past", "basic design remains almost the same", "rugged ship and easy to repair")
- Changed default armament to better match the galactic North region of space that its company Betelguese Shipyards comes from; now armed with Quad Blaster Turrets and Particle Cannons

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -872,8 +872,8 @@ ship "Corvette"
 	attributes
 		category "Medium Warship"
 		"cost" 4400000
-		"shields" 6100
-		"hull" 1200
+		"shields" 6800
+		"hull" 800
 		"required crew" 8
 		"bunks" 32
 		"mass" 150
@@ -895,7 +895,7 @@ ship "Corvette"
 		
 		"NT-200 Nucleovoltaic"
 		"LP072a Battery Pack"
-		"D67-TM Shield Generator"
+		"D94-YV Shield Generator"
 		"Large Radar Jammer"
 		
 		"X3700 Ion Thruster"
@@ -1188,8 +1188,8 @@ ship "Firebird"
 	attributes
 		category "Medium Warship"
 		"cost" 3700000
-		"shields" 6400
-		"hull" 2800
+		"shields" 5800
+		"hull" 3400
 		"required crew" 7
 		"bunks" 22
 		"mass" 290
@@ -1206,8 +1206,8 @@ ship "Firebird"
 			"hull damage" 500
 			"hit force" 1500
 	outfits
-		"Heavy Laser" 4
-		"Heavy Laser Turret" 2
+		"Particle Cannon" 2
+		"Quad Blaster Turret" 2
 		
 		"RT-I Radiothermal"
 		"nGVF-AA Fuel Cell"
@@ -1220,12 +1220,12 @@ ship "Firebird"
 		
 	engine -33 65
 	engine 33 65
-	gun -28 -27 "Heavy Laser"
-	gun 28 -27 "Heavy Laser"
-	gun -39 -13 "Heavy Laser"
-	gun 39 -13 "Heavy Laser"
-	turret -5 3 "Heavy Laser Turret"
-	turret 5 3 "Heavy Laser Turret"
+	gun -28 -27 "Particle Cannon"
+	gun 28 -27 "Particle Cannon"
+	gun -39 -13
+	gun 39 -13
+	turret -5 3 "Quad Blaster Turret"
+	turret 5 3 "Quad Blaster Turret"
 	explode "tiny explosion" 18
 	explode "small explosion" 36
 	explode "medium explosion" 24


Amazinite:
Looks good.

data/ships.txt
```diff
@@ -1188,8 +1188,8 @@ ship "Firebird"
 	attributes
 		category "Medium Warship"
 		"cost" 3700000
-		"shields" 6400
-		"hull" 2800
+		"shields" 5800
+		"hull" 3400
```
KiLEdEnNis: this does mean the Firebird will be disabled a little bit faster (total hp was 9200, stays 9200, previously disabled at 560, now 680), but shouldn't matter that much at all

Nescio0:
Concerning the shield and hull tweaks, no objections.

However, your proposed default armament changes would make the Firebird significantly weaker:
* shield damage per second would go down from 864 to 684 (i.e. -21%)
* hull damage per second would go down from 1152 to 552 (i.e. -52%)
* firing energy per second would go up from 480 to 610 (i.e. +27%)
* firing heat per second would go up from 1056 to 2120 (i.e. +101%)
Besides, Epsilon Leonis is one of three systems where heavy laser turrets are sold, but it also sells the Firebird. So why are these changes necessary?

Pointedstick:
Any change to replace laser weapons with something else will result in significant reduction in on-paper performance because lasers being too good--which is why there are multiple PRs to attempt to remedy this situation. That said, the move to Particle Cannons introduces a significant and powerful range advantage not accounted for in your calculations.

The proposed weaponry changes are not _necessary_, but I deemed them desirable to further differentiate the ships. The Corvette is manufactured by Lionheart, a Deep firm. So it would make sense for that ship to be armed entirely with Deep-style weaponry. The Firebird is a Betelguese ship, and Particle Cannons come from the galactic North. Quad Blaster turrets are because the Firebird has insufficient turret mounts to fit a bunch of regular Blaster Turrets.

Nescio0:
The Behemoth, another Betelgeuse ship, is equipped with Heavy Laser Turrets as well. And Quad Blaster Turrets are only sold and produced on major Syndicate planets. 

More important is that the proposed armaments will cause the Firebird to overheat after one second, while it inflicts significantly less damage. So I'd recommend to revert these weapon changes, or at least postpone them until blasters are made really useful. 
If you insist on changing something, replace the four heavy lasers with two particle cannons, but keep the heavy laser turrets.

Pointedstick:
> More important is that the proposed armaments will cause the Firebird to overheat after one second,

Actually it doesn't, I playtested it. Firing all guns and engines at once takes about 6 seconds to overheat IIRC.

But you're right that the new heat level is a bit crazy--and reiterates just how much improvement Blasters need to be competitive with lasers.

> If you insist on changing something, replace the four heavy lasers with two particle cannons, but keep the heavy laser turrets.

That seems reasonable. I do want to change _something_ about one of the ships' default armament in the service of differentiation, because without this patch they're armed identically.

ghoulavenger:
> But you're right that the new heat level is a bit crazy--and reiterates just how much improvement Blasters need to be competitive with lasers.

They don't need to be competitive with lasers. They just need their own niche.

KiLEdEnNis:
> They don't need to be competitive with lasers.

Pointed meant they need to be an actual choice to get as well; compete with lasers but be unique themselves, so they both have different functions but are about as good statistically

ghoulavenger:
@KiLEdEnNis They are an actual choice. Always have been. Try them on small ships. I was particularly fond of them on freighters (and they're pretty much the only choice for star barges).

The root of my disagreement is the idea that they need to be as good statistically. There is a progression curve, and I don't mean in tiers of weaponry (because that would be beam laser -> heavy laser -> electron beam), I mean in tiers of gameplay (what you start with and what you progress to next). People seem to want to ignore this progression curve though because the longer you play, the less money matters, and the more small ships get swept under the rug -- meaning it feels like theres no reason to use blasters, when in fact you're just ignoring their intended usage.

Pointedstick:
Please take the blasters vs lasers discussion to one of the relevant PRs or Issues.

endless-sky:
The real distinguishing mark is supposed to be their speed, not their armament. The Corvette has about 30% less drag and half the chassis mass. And, the Corvette already has less than half the hull strength of the Firebird. So, I'm not sure that further differentiation is needed.

Pointedstick:
What about the fact that the Corvette is described as having an advanced shielding system despite having lower shields than the Firebird?

Also don't you think it's a little boring for so many ships' stock armament to be nothing but heavy lasers?

Pointedstick:
Ping?

Amazinite:
I removed the loadout changes from this PR.
Ref: https://github.com/EndlessSkyCommunity/endless-sky/pull/26#issuecomment-432901874, and it conflicts with Pointed's changes in #3922 and is directly handled in #3982

Pointedstick:
Thanks!

I've put the "Lasers -> Particle Cannons" change in https://github.com/endless-sky/endless-sky/pull/3922, FWIW.

PR #3425, opened 2017-12-16
Define additional filled colors in interfaces.txt

tehhowch:
- Further support for modified interfaces / reskinning, such as @comnom 's "Pug HUD" ( https://github.com/comnom/es-huds )

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -17,6 +17,17 @@ color "medium" .5 .5 .5 0.
 color "dim" .25 .25 .25 0.
 color "faint" .1 .1 .1 0.
 
+color "panel background" .1 .1 .1 1.
+color "dialog backdrop" 0. 0. 0. .7
+color "conversation background" .125 .125 .125 1.
+color "map sidepanel background" .125 .125 .125 1.
+color "shop sidepanel background" .2 .2 .2 1.
+color "shop sidepanel footer" .3 .3 .3 1.
+color "tooltip background" .2 .2 .2 1.
+color "logbook sidebar" .09 .09 .09 1.
+color "logbook background" .125 .125 .125 1.
+color "logbook line" .2 .2 .2 1.
+
 color "shields" .43 .55 .70 .75
 color "hull" .70 .62 .43 .75
 color "heat" .70 .43 .43 .75
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -8,26 +8,31 @@
 # WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
 # PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
+# Colors used to reflect the active or hovered interface button.
 color "hover" 1. 1. 1. 0.
 color "active" .75 .75 .75 0.
 color "inactive" .25 .25 .25 0.
 
+# Colors generally used for drawing text or lines.
 color "bright" .75 .75 .75 0.
 color "medium" .5 .5 .5 0.
 color "dim" .25 .25 .25 0.
 color "faint" .1 .1 .1 0.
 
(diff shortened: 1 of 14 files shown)


data/interfaces.txt
```diff
@@ -17,6 +17,17 @@ color "medium" .5 .5 .5 0.
 color "dim" .25 .25 .25 0.
 color "faint" .1 .1 .1 0.
 
+color "panel background" .1 .1 .1 1.
+color "dialog backdrop" 0. 0. 0. .7
+color "conversation background" .125 .125 .125 1.
+color "map sidepanel background" .125 .125 .125 1.
+color "shop sidepanel background" .2 .2 .2 1.
+color "shop sidepanel footer" .3 .3 .3 1.
```
endless-sky: I think "side panel" should be two words. "Sidebar" is a well-established word, but "sidepanel" is not.

tehhowch: Perhaps these should be sidebar then instead of sidepanel? Name suggestions are always welcomed

tehhowch:
I've added some comments to the colors definition section to indicate what the colors are used for (so that the name alone doesn't need to convey what the color is for).

source/MissionPanel.cpp
```diff
@@ -506,9 +506,9 @@ void MissionPanel::DrawMissionSystem(const Mission &mission, const Color &color)
 Point MissionPanel::DrawPanel(Point pos, const string &label, int entries) const
 {
 	const Font &font = FontSet::Get(14);
-	Color back(.125, 1.);
-	Color unselected = *GameData::Colors().Get("medium");
-	Color selected = *GameData::Colors().Get("bright");
+	const Color &back = *GameData::Colors().Get("map sidepanel background");
```
endless-sky: This name needs to be updated ("side panel")

source/ShopPanel.cpp
```diff
@@ -148,19 +148,19 @@ void ShopPanel::Draw()
...
 	FillShader::Fill(
 		Point(Screen::Right() - SIDE_WIDTH / 2, 0.),
 		Point(SIDE_WIDTH, Screen::Height()),
-		Color(.1, 1.));
+		*GameData::Colors().Get("panel background"));
 	FillShader::Fill(
 		Point(Screen::Right() - SIDE_WIDTH, 0.),
 		Point(1, Screen::Height()),
-		Color(.2, 1.));
+		*GameData::Colors().Get("shop sidepanel background"));
```
endless-sky: Same here.

tehhowch: that's an oops. :laughing:

source/ShopPanel.cpp
```diff
@@ -262,14 +262,15 @@ void ShopPanel::DrawButtons()
 {
 	// The last 70 pixels on the end of the side panel are for the buttons:
 	Point buttonSize(SIDE_WIDTH, BUTTON_HEIGHT);
-	FillShader::Fill(Screen::BottomRight() - .5 * buttonSize, buttonSize, Color(.2, 1.));
+	FillShader::Fill(Screen::BottomRight() - .5 * buttonSize, buttonSize, *GameData::Colors().Get("shop sidepanel background"));
```
endless-sky: And, here and below.

source/MapSalesPanel.cpp
```diff
@@ -269,7 +269,7 @@ void MapSalesPanel::DrawInfo() const
 			width += box->Width() + compareInfo.PanelWidth();
 		}
 		
-		Color back(.125, 1.);
+		const Color &back = *GameData::Colors().Get("map sidepanel background");
```
endless-sky: And these two.

tehhowch:
@endless-sky got 'em all

PR #3426, opened 2017-12-16
Improved blasters

Nescio0:
Yet another counterproposal to make blasters somewhat more attractive. This one does not involve any changes to default or variant ship builds. As suggested in #3393:

> Increasing the Blaster range (maybe by increasing speed rather than lifetime) and decreasing its heat production.

This proposal does that:
* blaster velocity increased from 7 to 8.5, which means a range increase from 420 to 510
  * modified blaster from 7.5 to 9.0, thus range from 450 to 540
* blaster and modified blaster firing heat reduced by 3/7 (to make it twice the firing energy)
* blaster and modified blaster firing energy remains unchanged (i.e. highly inefficient)
* blaster shield + hull damage per second changed from 45+30 to 50+25
  * modified blaster from 60+40 to 70+35

With these changes blasters remain highly inefficient, but might become occasionally useful.

*not implemented* but to be considered, because of modified blaster's 40% higher damage:
  * modified blaster gun size from 6 to 7 (blaster has 5)
  * modified blaster turret size from 18 to 20 (blaster has 15, quad blaster 25); 
  * the "Modified Argosy (Blaster)" seems to be the only ship equipped with modified blasters (maybe replace the ordinary quad blaster turrets with modified blaster turrets to save space and heat?)

Effectively this patch results into:

Blasters:
* poor energy and heat efficiencies
* good credit and mass efficiencies
* mediocre range
* higher shield than hull damage

Lasers:
* great energy and heat efficiencies
* decent credit and mass efficiencies
* low range (especially with #3406)
* higher hull than shield damage
* continuous fire

So yes, lasers remain attractive, but blasters get their own niche.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -109,8 +109,8 @@ outfit "Modified Blaster"
 		"reload" 12
 		"firing energy" 16
 		"firing heat" 32
-		"shield damage" 16
-		"hull damage" 8
+		"shield damage" 14
+		"hull damage" 7
 	description "This is an Energy Blaster that has been tampered with to increase its power and range slightly, at the cost of a significant increase in its energy requirements. Modified Blasters also produce an extreme amount of heat when firing, increasing the risk that your ship will overheat."
 
 outfit "Modified Blaster Turret"
@@ -135,8 +135,8 @@ outfit "Modified Blaster Turret"
 		"reload" 6
 		"firing energy" 16
 		"firing heat" 32
-		"shield damage" 16
-		"hull damage" 8
+		"shield damage" 14
+		"hull damage" 7
 	description "This is a turreted version of the Modified Blaster, which is an Energy Blaster modified for greater power at the cost of higher energy consumption and heat. These turrets are popular with pirates and others who are trying to cram as much firepower into their ships as possible."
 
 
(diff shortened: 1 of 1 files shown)


tehhowch:
As an aside, "mb" is an absolutely terrible commit name. https://github.com/endless-sky/endless-sky/commit/845c304698d4442da64c81cebe21991c48742380

If I were to see that title and description in a git log I would have zero clue what that commit does or tries to do.

Nescio0:
> As an aside, "mb" is an absolutely terrible commit name.

You're right, I should have written "modified blasters" at the very least. Whenever I use command line git I typically use more precise commit messages. With web forms (such as github's) I have the bad habbit to use abbreviations or placeholders to fill in required fields quickly, which I do not always expand properly afterwards; a poor excuse, I know.

Amazinite:
Mostly a welcomed change. I'm just unsure about increasing the modified blaster's combined DPS (even though it is only a 5% buff). If we aren't afraid of damage values with decimals in them, then I'd suggest changing the modified blaster's shield damage to 13.3 and the hull damage to 6.7 in order to get its shield/hull as close to 2 as possible without using repeating decimals and have it retain its current combined DPS.

Nescio0:
https://github.com/endless-sky/endless-sky/commit/beb7f03836f37e173134e9e9f510494393924bb7 does implements something similar, so this proposal probably won't be implemented.

PR #3427, opened 2017-12-16
Equipped variants with rifles

Nescio0:
A frequent complaint is that seizing ships is too easy. This patch equips human variant ships with a number of hand-to-hand rifles equal to their total bunks (as is already the case with Marauders). 
Would this make seizing individual ships impossibly hard? Not really. However, the advantage is that variant ships are better able to defend themselves, inflicting larger crew losses on the attacker, thus making it harder to seize many ships in quick succession. 
Because ordinary rifles do not take up outfit capacity, this change does not break any builds.

Furthermore, space is a dangerous place. It always seemed a bit weird ships allow themselves to be plundered only to be blown up afterwards. With rifles they might be more willing to fight for their own lifes.

Why only variants and not default ships? Variant ships are not for sale; the AI does not purchase rifles; human players can choose for themselves whether or not they want any hand-to-hand weapons; adding rifles by default also increases purchase price unnecessarily; this is no issue for variants.

This simple patch has the potential to improve gameplay.

### DIFF
--- a/data/variants.txt
+++ b/data/variants.txt
@@ -17,6 +17,7 @@ ship "Star Barge" "Star Barge (Armed)"
 		"Chipmunk Plasma Thruster"
 		"Chipmunk Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 3
 
 
 
@@ -31,6 +32,7 @@ ship "Sparrow" "Sparrow (Patrol)"
 		"Chipmunk Plasma Thruster"
 		"Chipmunk Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 2
 
 
 
@@ -43,6 +45,7 @@ ship "Raven" "Raven (Afterburner)"
 		"X3200 Ion Steering"
 		"Hyperdrive"
 		"Afterburner"
+		"Laser Rifle" 13
 
 
 
@@ -55,6 +58,7 @@ ship "Wasp" "Wasp (Proton)"
 		"X2700 Ion Thruster"
 		"X2200 Ion Steering"
 		"Hyperdrive"
+		"Laser Rifle" 2
 
 
 
@@ -69,6 +73,7 @@ ship "Manta" "Manta (Proton)"
 		"X3700 Ion Thruster"
 		"X3200 Ion Steering"
 		"Hyperdrive"
+		"Laser Rifle" 10
 	gun "Meteor Missile Launcher"
 	gun "Meteor Missile Launcher"
 	gun "Proton Gun"
@@ -88,6 +93,7 @@ ship "Quicksilver" "Quicksilver (Proton)"
 		"Greyhound Plasma Thruster"
 		"Greyhound Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 6
 
 
 
@@ -112,6 +118,7 @@ ship "Cruiser" "Cruiser (Heavy)"
 		"A525 Atomic Steering"
 		"X2200 Ion Steering"
 		"Scram Drive"
+		"Laser Rifle" 136
 	gun "Particle Cannon"
 	gun "Particle Cannon"
 	gun "Particle Cannon"
@@ -140,6 +147,7 @@ ship "Scout" "Scout (Speedy)"
 		"A370 Atomic Thruster"
 		"A255 Atomic Steering"
 		"Hyperdrive"
+		"Laser Rifle" 12
 
 
 
@@ -155,6 +163,7 @@ ship "Clipper" "Clipper (Heavy)"
 		"Greyhound Plasma Thruster"
 		"Impala Plasma Steering"
 		"Scram Drive"
+		"Laser Rifle" 9
 
 
 
@@ -168,6 +177,7 @@ ship "Clipper" "Clipper (Speedy)"
 		"A370 Atomic Thruster"
 		"A255 Atomic Steering"
 		"Hyperdrive"
+		"Laser Rifle" 9
 
 
 
@@ -183,6 +193,7 @@ ship "Freighter" "Freighter (Fancy)"
 		"Greyhound Plasma Thruster"
 		"X3200 Ion Steering"
 		"Scram Drive"
+		"Laser Rifle" 7
 	turret "Laser Turret"
 	turret "Laser Turret"
 	turret "Heavy Anti-Missile Turret"
@@ -201,6 +212,7 @@ ship "Freighter" "Freighter (Proton)"
 		"Greyhound Plasma Thruster"
 		"X3200 Ion Steering"
 		"Scram Drive"
+		"Laser Rifle" 7
 
 
 
@@ -216,6 +228,7 @@ ship "Argosy" "Argosy (Laser)"
 		"Greyhound Plasma Thruster"
 		"Impala Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 14
 
 
 
@@ -232,6 +245,7 @@ ship "Argosy" "Argosy (Missile)"
 		"A370 Atomic Thruster"
 		"A255 Atomic Steering"
 		"Hyperdrive"
+		"Laser Rifle" 14
 	turret "Anti-Missile Turret"
 	turret "Heavy Anti-Missile Turret"
 
@@ -248,6 +262,7 @@ ship "Argosy" "Argosy (Proton)"
 		"Greyhound Plasma Thruster"
 		"Impala Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 14
 
 
 
@@ -262,6 +277,7 @@ ship "Argosy" "Argosy (Turret)"
 		"X3700 Ion Thruster"
 		"Greyhound Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 14
 
 
 
@@ -276,6 +292,7 @@ ship "Argosy" "Argosy (Blaster)"
 		"Greyhound Plasma Thruster"
 		"Impala Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 14
 
 
 
@@ -295,6 +312,7 @@ ship "Behemoth" "Behemoth (Proton)"
 		"A125 Atomic Steering"
 		"A375 Atomic Steering"
 		"Scram Drive"
+		"Laser Rifle" 30
 	turret "Proton Turret"
 	turret "Proton Turret"
 	turret "Proton Turret"
@@ -319,6 +337,7 @@ ship "Behemoth" "Behemoth (Speedy)"
 		"A125 Atomic Steering"
 		"A375 Atomic Steering"
 		"Scram Drive"
+		"Laser Rifle" 30
 
 
 
@@ -336,6 +355,7 @@ ship "Bulk Freighter" "Bulk Freighter (Heavy)"
 		"A250 Atomic Thruster"
 		"A375 Atomic Steering"
 		"Scram Drive"
+		"Laser Rifle" 18
 	turret "Laser Turret"
 	turret "Laser Turret"
 	turret "Heavy Anti-Missile Turret"
@@ -357,6 +377,7 @@ ship "Bulk Freighter" "Bulk Freighter (Blaster)"
 		"A250 Atomic Thruster"
 		"A375 Atomic Steering"
 		"Scram Drive"
+		"Laser Rifle" 18
 
 
 
@@ -373,6 +394,7 @@ ship "Bulk Freighter" "Bulk Freighter (Proton)"
 		"A250 Atomic Thruster"
 		"A375 Atomic Steering"
 		"Scram Drive"
+		"Laser Rifle" 18
 
 
 
@@ -386,6 +408,7 @@ ship "Fury" "Fury (Missile)"
 		"Greyhound Plasma Thruster"
 		"Chipmunk Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 3
 
 
 
@@ -399,6 +422,7 @@ ship "Fury" "Fury (Laser)"
 		"Chipmunk Plasma Thruster"
 		"A125 Atomic Steering"
 		"Hyperdrive"
+		"Laser Rifle" 3
 
 
 
@@ -411,6 +435,7 @@ ship "Hawk" "Hawk (Speedy)"
 		"A250 Atomic Thruster"
 		"A255 Atomic Steering"
 		"Hyperdrive"
+		"Laser Rifle" 4
 
 
 
@@ -424,6 +449,7 @@ ship "Hawk" "Hawk (Rocket)"
 		"Greyhound Plasma Thruster"
 		"Greyhound Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 4
 
 
 
@@ -439,6 +465,7 @@ ship "Bastion" "Bastion (Heavy)"
 		"X3700 Ion Thruster"
 		"Orca Plasma Steering"
 		"Hyperdrive"
+		"Laser Rifle" 40
 
 
 
@@ -456,6 +483,7 @@ ship "Bastion" "Bastion (Laser)"
 		"A370 Atomic Thruster"
 		"A525 Atomic Steering"
 		"Hyperdrive"
+		"Laser Rifle" 40
 	turret "Heavy Anti-Missile Turret"
 	turret "Heavy Laser Turret"
 	turret "Heavy Laser Turret"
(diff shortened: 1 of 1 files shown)


jafdy:
The AI also flies the default ships so it would be inconsistent.
I don't like this change as it feels like a very forced way of solving the problem.

tehhowch:
Rather than total bunks you could perhaps go to the average crew, which is required + half the remaining.
One reason to do this is that by equipping variants with items that increase their cost, you alter the game's strength calculation (the strength of a ship is measured in its cost). Reducing the number of rifles added reduces this impact, and the new disparity between base ships and variant definitions (both of which are widely used in instantiated fleets).

If you're going to modify variants you might as well add Security Stations to ships which make sense (i.e determine and report the impact on free outfit space).

Nescio0:
> The AI also flies the default ships so it would be inconsistent.

Well, I could also add hand-to-hand weapons to default ships, if you like, or perhaps only to navy ships? Also, Korath ships are equipped with rifles. 

> Rather than total bunks you could perhaps go to the average crew

Yes, I could. For most ships the difference is quite small (rifles are cheap and most small ships have few bunks), but for e.g. Bactrians it probably matters.

> If you're going to modify variants you might as well add Security Stations to ships which make sense

Actually I considered adding SS to heavy ships such as the Star Queen, Leviathan, and Behemoth, but rejected the idea because it would be inconsistent with other ships, because they decrease outfit capacity and speed, and because they might become too hard to seize.

jafdy:
No, you create a variant with rifles to put in fleets. (Or a variant without rifles to put in shipyards.)

tehhowch:
I'm fine with adding some h2h outfit to all variants.  Harder to capture doesn't mean impossible to capture. The starting capped ship is of course the suicidal pirate Sparrow, and that is a stock ship so this PR won't prevent that opening money scheme from being viable - just making capture slower at the nerve gas stage.

Nescio0:
> Harder to capture doesn't mean impossible to capture. 

Exactly. If you can now seize e.g. six medium ships before having to recruit additional crew, with this patch you have to stop after perhaps the third.

Something else: in our world missiles cost a few million each, hand guns a few thousand, approximately. In this game a rifle is 8,000, missiles range from 50 to 2,000. How are hand to hand weapon prices calculated?

Pointedstick:
Rifle prices are rather ridiculous when you compare them to missile prices, but that's somewhat outside of the scope of this patch (though still worth exploring elsewhere).

This PR makes sense to me--though please do follow @tehhowch's suggestion and reduce the number of rifles per ship.

ghoulavenger:
I don't know how much I like it. On the one hand, ships are pretty easy to capture as it is, on the other you're pretty much pigeonholing people into using nerve gas. Rifles and grenades as it stands are quite effective H2H weapons, but they stand to lose quite a bit of power if every ship is running rifles.

KiLEdEnNis:
I think some rifles in variants, especially non-warship ones makes sense, but loading them all to the max is a no for me

Nescio0:
It would make it somewhat more difficult to capture a ship, yes, but certainly not impossible, nor does it force “people into using nerve gas.”
Let's have a look at some attack/defence values:

unarmed crewman: 1.0/1.0
crewman with rifle: 1.6/1.8
crewman with grenades: 2.3/1.3
crewman with nerve gas: 3.8/1.8
crewman with security station: 1.0/4.4

If you have your crew equipped with grenades you'll still be able to seize a similar ship fully armed with rifles; the difference with this patch is that you'll suffer more casualties, thus making seizing many ships in a row less attractive.

And yes, I'm certainly willing to reduce the number of rifles (to average crew size? some different number?), but I'll allow some more discussion first.

Brick63:
If it's a republic or FW warship it makes sense to give them almost max laser rifles but not really the civilian ones

Pointedstick:
Really the larger warships with extra space available should have a couple of Security Stations _in addition_ to giving everybody a rifle.

endless-sky:
It wouldn't make any sense for so many ships to have more hand-to-hand weapons than they have crew members. In general, the number of rifles should never be more than the number of *required* crew, and in most cases it should be less by a random amount, so that ships are distributed across the whole range from no weapons at all to one weapon for every crew member.

Pointedstick:
Yeah. And this could be a world-realism touch, too. Military warships and pirate ships would probably arm every crew member. Star Queens might fit some Security Stations but _not_ arm anyone. Merchant ships could vary from being unarmed to somewhat well armed.

endless-sky:
It looks like all you did was decrease the numbers to match the required crew number. As I said in my previous comment, that really doesn't make sense. It means that when you board an Argosy, for example, it is always going to have either 0 rifles, or 4. It'd be much more realistic if there were some with 1, 2, or 3 rifles instead.

Also, to be consistent with other ship definitions, the hand-to-hand weapons should be listed at the end of the "systems" block of outfits, not after the engines and hyperdrive.

jafdy:
I feel like some code implementation could help here, Some way to equip an AI ship with some amount of an outfit between x and y. Perhaps `"extra outfit" "laser rifle" 0 12` would give a ship controlled by the AI some number of laser rifles between 0 and 12 (inclusive). There is bound to be a better syntax than mine.
This would stop all `Falcon (Heavy)`s having exactly 34 (or whatever other number) or rifles. That said it is not much benefit.

Nescio0:
> It looks like all you did was decrease the numbers to match the required crew number.

It now matches the minimum crew, yes, without taking the turret crew requirements into account. Assigning an arbitrary number x to a variant u is not a bit more random than matching the minimum, required, average, full, maximum, or any other crew number; I don't see how realism would be improved if all variants u always have x rifles. If you really want random numbers, implement a random number generator, as jafdy suggested.

By the way, the Marauder ships (see `data/marauders.txt`) typically have a number of rifles equal to the number of bunks (not minimum crew).

Bladewood:
> Thanks for pointing this out; I simply followed the outfitter category order. So not after Engines? Where then, above the Power outfits? And doesn't the hyperdrive also belong under Systems?

From https://github.com/endless-sky/endless-sky/wiki/CreatingShips#data:

> Outfits work by adding or subtracting to these same attributes. The outfits list can be in any order, but for the sake of consistency its preferred order is roughly from "front to back" of the ship:
> - guns
> - turrets
> - generators
> - batteries
> - shields
> - cooling
> - outfits that don't fit any other category (expansions, scanners, hand to hand weapons, etc.)
> - thruster
> - steering
> - hyperdrive

ghoulavenger:
> It would make it somewhat more difficult to capture a ship, yes, but certainly not impossible, nor does it force “people into using nerve gas.”
Let's have a look at some attack/defence values:

>unarmed crewman: 1.0/1.0
crewman with rifle: 1.6/1.8
crewman with grenades: 2.3/1.3
crewman with nerve gas: 3.8/1.8
crewman with security station: 1.0/4.4

Did you consider what I wrote and what you wrote? A crewman with a rifle is 1.6 attack, 1.8 defense (0.2 disadvantage). This means that if I'm using rifles to capture a ship, unless your ship has significantly more crew, you're at a serious disadvantage. Now look at Grenades 2.3 to 1.8 (0.5 advantage), basically makes grenades rifle tier for offense. 

I said they would take a significant loss, and they do. Using rifles would now be pointless, and using grenades would be like using rifles. The only weapon that maintains a significant advantage is nerve gas (at a full 2 point advantage, making it almost four times as effective as using grenades). Which is exactly what I said -- pigeonholing people into nerve gas. Honestly I think defenders actually get a bonus to boot iirc, but I don't recall what it is, which would only further push my point.

Does that matter? When people are swinging bactrians maybe not (might even be able to still go completely unarmed in a bactrian depending on what you're boarding). But in the early game, you're lucky to have a firebird or a mule where every single point counts. Heck I started capturing ships in a star barge (install a bunk room and some rifles, it works fine), then gradually upgraded to a bounder and firebird.

Does that mean I'm against it? Well I don't know. I can see the merits of both sides. On the plus side it adds a bit of realism, and increases the difficulty. On the minus side it pigeonholes people into nerve gas (I used nerve gas anyway but this makes it feel forced).

Bladewood:
> Honestly I think defenders actually get a bonus to boot iirc, but I don't recall what it is

For the record, it doubles the base crew defense (not including H2H weapons).

endless-sky:
What I'm saying is, that if there are 4 different variants and each one has a different number of rifles, and it's random which variant you're fighting at a given time, it will create the appearance of randomness even if a particular variant always has the same number.

Implementing a randomizer for the number of outfits just for this one use case would make no sense.

Nescio0:
> From https://github.com/endless-sky/endless-sky/wiki/CreatingShips#data:

Thanks, I'll correct it accordingly.

> On the plus side it adds a bit of realism, and increases the difficulty. On the minus side it pigeonholes people into nerve gas (I used nerve gas anyway but this makes it feel forced).

Before you started using nerve gas, were you able to seize any Marauder (Pirate) ships? If yes, then this proposal ought to be unproblematic, since Marauder ships already have a number of rifles equal to their total number of bunks, at least since 2016. 
I have no experience with early game capturing myself; I only discovered the boarding feature when I already had a fleet of Bactrians; nor did I ever use nerve gas (a larger crew fully armed with rifles typically is enough). However, looking at the numbers, I still believe this proposal would make seizing a single ships somewhat harder but certainly not impossible.

> What I'm saying is, that if there are 4 different variants and each one has a different number of rifles, and it's random which variant you're fighting at a given time, it will create the appearance of randomness even if a particular variant always has the same number.

Yeah, I understand what you're saying, I just happen to disagree. If there were dozens of variants for each ship, which isn't the case, then yes, defence strength might seem somewhat random. However, with only three or four variants, the occurence of which is determined by regional fleets, I believe players who pay some attention quickly see the repetitiveness (and players who don't won't notice the numbers are occassionally different at all).

> Implementing a randomizer for the number of outfits just for this one use case would make no sense.

Not just for this case. If the crew + passenger size of random ships would be a random number between minimum and maximum crew, individual ship strength wouldn't be repetitive, and realism would be greatly improved.

tehhowch:
Ship crew **already is** random between required and maximum.

As MZ mentions, implementing an equipped outfit quantity randomizer is not sensible. No one would want to have a ship with a random number of reactors, or engines, or guns, etc. so the use case is only hand to hand weapons. 

Between the variable crew strength of the same variant per instantiation, and the different type and/or quantity of h2h weaponry, suitable diversity can be achieved.

Nescio0:
> Ship crew already is random between required and maximum.

Is it? Great! I've seen ships with larger than required crews, but I wasn't aware the numbers were fully random, thanks for the confirmation.

> As MZ mentions, implementing an equipped outfit quantity randomizer is not sensible. No one would want to have a ship with a random number of reactors, or engines, or guns, etc. so the use case is only hand to hand weapons.

Yes, I agree, creating a random generator only for outfits is probably a waste of time.

> Between the variable crew strength of the same variant per instantiation, and the different type and/or quantity of h2h weaponry, suitable diversity can be achieved.

Default ships have 0 hand to hand weapons, marauders have maximum hand to hand weapons; which hardcoded not-random-but-still-different numbers would you suggest for different variants?

tehhowch:
Something between 0 and max. You can decide on some logic to help pick if you wish - perhaps a variant that has more firepower has more defense than one that is heavily armored but less able to deal damage. Or maybe vice versa. 

You could go a step further and check what other ships are generally used with a given variant, and make it so one ship is easy to cap, but the others are harder.

ghoulavenger:
>Before you started using nerve gas, were you able to seize any Marauder (Pirate) ships? If yes, then this proposal ought to be unproblematic, since Marauder ships already have a number of rifles equal to their total number of bunks, at least since 2016.
I have no experience with early game capturing myself; I only discovered the boarding feature when I already had a fleet of Bactrians; nor did I ever use nerve gas (a larger crew fully armed with rifles typically is enough). However, looking at the numbers, I still believe this proposal would make seizing a single ships somewhat harder but certainly not impossible.

I never said impossible, I said it would pigeonhole people into using nerve gas. I started using nerve gas long before I started messing around with Marauder ships, so I couldn't tell you. Not to mention for a very long time I ignored bounty hunting missions (where marauders spawn) since finding the ships was always so unpredictable (still is, but now I know they don't move around and they have special ship variants and other ships won't destroy them, so sometimes I do now).

Early game capturing I can be a LOT more insightful on. With a star barge with a bunk and 7 rifles, about the only thing you're going to capture is something with a crew of 3 or less. Which usually means an interceptor. But if you get really lucky you might be able to capture a raven (I usually lose which is why I end up buying a bounder). A bounder is a little bit better, and can nab you your raven pretty frequently, but at that point the next best thing is a firebird. It's a pretty close match depending on how many crew the firebird spawns, not to mention the difficulty in surviving getting close to a firebird (they frequently spawn with leviathans, so if they don't get blown up straight away good chance you get blown up trying to board it).

With a firebird, you're lucky to pull 30-40 crew before you start impacting your combat worthiness, and up to this point you're basically feeding off of scraps from the republic fleet. So this is my ship of choice until I start pulling in a bigger amount of coin. Usually nerve gas is still too expensive as well. So basically my capturing capacity is what I can nab in about 40 crew, reliably thats probably just more firebirds or equivalent since by this time I head down south where there is less republic interference. So now we're getting into clipper captures, which is pretty light, and if you're lucky an argosy. If you run into a falcon, you run, far less support than up north, and you can't capture it anyway, even with nerve gas. 

If you get supremely lucky you can nab a bastion. But I think at this point I just start trying to build up a few firebirds, corvettes, or even modified argosies and simply buy my next upgrade, the mule. If you're capable of decking the mule out in crew, you can start capturing larger ships like falcons. But doing so will be a one and done deal, but it's still actually far more profitable to capture clippers and argosies (where you have a significant crew advantage). This is pretty much your limitation unless you're willing to work with a subpar falcon which isn't significantly better than a mule for boarding purposes. Once I start getting ready to buy a bactrian I will capture falcons anyway (to use as escorts), but like I said, it's a one and done deal (which means I need to wait for a falcon to show up before I start boarding stuff).

How would each ship I board using rifles matter? Well it might make it impossible for me to capture that first firebird with a bounder, since I'd have to be using rifles and the crew compliments aren't that much better on a bounder. Nerve gas is too expensive at that point. Which would make the capturing early game much more difficult, I'd either be using a raven or a bounder for awhile until I could buy my next upgrade, instead of using the far superior firebird. By the time I go down south I can usually upgrade to grenades, but not nerve gas. This means at this point I'm probably okay, since you almost always capture ships that are smaller than yours as you prepare for your next upgrade. Basically, firebirds aren't capable of holding enough crew of capturing a heavy warship.

The stock mule can't take a heavy warship reliably either, even with nerve gas. You need to reoutfit the bugger to manage, and even then you'll only get one and it'll be undercrewed. It isn't until you get bactrian levels of crew that you can really capture the heaviest warships with ease, due to excessive crew (245 to 70 or so is over a 3:1 advantage on most heavy warships).

Nescio0:
Thank you for your elaborate post. Your concern is noted.

> Something between 0 and max.

Since nobody else posted a preference or suggestion afterwards, I simply used python to generate pseudo-random numbers, which are now implemented.

KiLEdEnNis:
I would personally make it between 0 and minimum crew, but this at least works

Nescio0:
> Security stations?
They should probably only be given to military and colony vessels, and even in the stock configuration. You'll have to ensure no outfit spaces are exceeded when adding them.

Bactrians and Carriers really don't need them, because they already have large crews, and are more likely to board than to be boarded; the same applies to a lesser degree to heavy warships. The Star Queen is possibly the most logical ship to have security stations (because you probably don't want to have rifles amongst your luxury guests), however, the SQ has no variants, and the default build has 0 free outfit space.
Maybe the Freighter could have one, and the Bulk Freighter two, to compensate for their vulnerability and small crew? Any other candidates?

tehhowch:
I would *expect* a colony ship like the Bactrian and Mule to have security stations. Surely there were some scuffles and both passenger and crew disagreements on a long colonization journey :)

Nescio0:
> I would expect a colony ship like the Bactrian and Mule to have security stations. Surely there were some scuffles and both passenger and crew disagreements on a long colonization journey :)

Well, I always assumed security stations were fixed and immobile, set up at the entrance of the ship, great for repelling invaders, but useless in internal scuffles (two tons and no wheels seems a lot to move around manually).

Hauler I, II, III, Nest, Roost, Skein could also benefit from two security stations each.

Brick63:
I always felt the security stations where put in hallways and other choke points and not near the entrances where they would be useless after being boarded

KiLEdEnNis:
I'll quote MZ's earlier comment:
> It wouldn't make any sense for so many ships to have more hand-to-hand weapons than they have crew members. In general, the number of rifles should never be more than the number of required crew, and...

so make the H2H weapon amount between 0 and required crew

Bladewood:
Note that the Heavy Cruiser variant has rifles and grenades.

tehhowch:
Nothing wrong with that, @Bladewood . I'd have it also employ Security Stations (within reason, of course).

Issue #1893, opened 2016-12-13
Usability: no cue for system search


lheckemann:
I had no idea I could search for star systems until I accidentally pressed F while on the map screen. It would be very useful to have some sort of visible cue that this is possible.

MessyMix:
Same condition as you! In fact, I was looking really hard for something, and was about to suggest it online, when the thought to press "Cmd+F" came to mind. And then _voila_!

Wrzlprnft:
2d77d1b

endless-sky:
My general philosophy has been that if something in the UI is nearly universal (like, using the scroll wheel on your mouse to scroll a list) I might mention it in the documentation, but there's no reason to mention it in the in-game help.

Granted, as a Linux user, once or twice a day I'm typing something like "^W" or "/" to do a find, but for 99% of users every program they use will respond to "^F".

Pointedstick:
It's still invisible, though; you need to have some conception that hitting the "f' key is going to have some effect before you even think to try it (I never did, and found out about the feature by reading this issue report).

If there's going to be a search feature, it really ought to have *some* kind of visual UI hint, or better yet: an actual search field somewhere, like most UI programs have when there's a search feature available.

lheckemann:
@Pointedstick exactly.

@endless-sky but the C-F shortcut isn't really universal. There's also F3 for instance, which I think is a lot more common in older Windows/DOS software (I may be wrong, I'm really not an expert); and the most widespread shortcut for searching in the software I use (which of course also doesn't make it universal) is / as you mentioned. I think a neat solution would be having an always-visible search bar in the middle at the top, which when you type in it changes the colour scheme so that matching planets have a white ring and others are grey (or something like that).

(slightly OT: the main use I have for the search feature is finding where my escorts are hanging about relative to me; what would also be nice is some way of displaying this directly on the map.)

endless-sky:
In my opinion it's a matter of giving each UI feature a degree of visibility that roughly corresponds to how useful it is. If every single feature had to have a visible UI widget pointing out its existence, it'd be hard for new players to figure out which widgets are useful to them and which aren't. That's the same reason that the pop-up help just focuses on the important stuff and lets the manual handle the rest: presenting too many options or too much information is as bad as not presenting enough.

And, be realistic: for 99.9% of users, they aren't running DOS or nano or vi, they're using modern GUI applications, in which ^F for find is nearly as universal as ^X for cut and ^V for paste.

[...]

Issue #1030, opened 2016-05-03
Find galaxy map function


Disiuze:
http://steamcommunity.com/app/404410/discussions/2/364040166692565759/

endless-sky:
It's already possible to search for a particular system within that galaxy to view it, e.g. "World" to center the map on World's End. Since the galaxy itself never has a name displayed, I don't know what you would expect the player to search for to view it.

Loke85:
Just a tip, I believe this doesn't work as if the route to the other galaxy is through a wormhole with "hidden connection"(the lack of a purple line, present in "puglaxy"), the game can't find a path to the destination and will center on (5000, 3500)(the middle).

Test it for yourself, I think it can be fixed by changing the check of the search function from "can find path" to "has visited".

Also, I presume the ability to prevent the purple line from appearing comes from the inability to get 'visited "Pug Wormhole"'in the save file. If so, how do you do that?, What causes the "visited" check?

jafdy:
You can't map a path through the Pug wormhole because the end points are in different galaxies. Simply there is no direction to draw the line in.

Loke85:
Um, actually that's wrong. Galaxies in the code are simply background pictures in a grid. The empty pug galaxy is really just some systems with background approximately 3 "screen diagonals" from the usual galaxy.
[Proof](http://imgur.com/RdyfGAy)
You could even edit in a hyperlink between the different "galaxies" and it would work.

jafdy:
I should just read the files more carefully. I was making my guess from never having scrolled to it (despite trying) and from not being able to use the find function between them.

endless-sky:
It's a little non-intuitive, but the purple wormhole lines will only be drawn if there is an explicit "planet" object corresponding to the wormhole:

```bash
planet "Wormhole Alpha"
```

Also, at least with the latest code, searching for "World" or "Incognita" (for example) in the map brings you to the Pug galaxy.

Loke85:
Yeah, i might have an outdated version.

PR #3431, opened 2017-12-17
User-defined colors & deprecated function

tehhowch:
- Swaps a few more colors from the map drawing system to be user-defined
 - Removes an unused function declaration (function removed in https://github.com/endless-sky/endless-sky/commit/cdb1cb881d04b246feee03dff4ec0028f34f6327#diff-e0a59618d76b767cb62fa83a9f007c59L295)

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -824,6 +824,12 @@ interface "map buttons" bottom right
 	button i "M_issions"
 		from -170 -50 to -80 0
 	
+	active
+	sprite "ui/dialog cancel"
+		from -170 -90 to -80 -40
+	button f "_Find..."
+		from -170 -90 to -80 -40
+	
 	active if "!is ports"
 	sprite "ui/dialog cancel"
 		from -90 -50 to 0 0
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -97,10 +97,6 @@ color "map travel ok none" .55, .1, 0., 0.
 color "map used wormhole" .5, .2, .9, 1.
 color "map unused wormhole" .165, .066, .3, .333
 
-# Colors used when drawing map details
-color "close" .6 .6 .6 .6
-color "far" .3 .3 .3 .3
-
 
 
 interface "menu background"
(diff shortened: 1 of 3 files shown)


endless-sky:
I think it'd just be better to use the "medium" and "dim" colors in place of "near" and "far." Note that when drawn on a background with a grey level of about 0.15 (32 to 48 in grey levels), the "medium" color (0.5 additive; 0.15 + 0.50 = 0.65) will look nearly the same as the "close" color (1 at 0.6 alpha; 0.15 * 0.4 + 1 * 0.6 = 0.66). Similarly, the "dim" color (0.25 additive; 0.15 + 0.25 = 0.400) has nearly the same effect as the "far" color (1 at 0.3 alpha; 0.15 * 0.7 + 1 * 0.3 = 0.405).

endless-sky:
(The map panel is one of the earliest parts of the game; I probably created it before the named colors in the interface even existed. That's why it's sometimes inconsistent about using those colors.)

tehhowch:
Added a button to visually convey that the `Find` dialog exists and has a hotkey (Refs #1893, #1030, perhaps more).

data/interfaces.txt
```diff
@@ -825,6 +824,12 @@ interface "map buttons" bottom right
 	button i "M_issions"
 		from -170 -50 to -80 0
 	
+	active
+	sprite "ui/dialog cancel"
+		from -170 -90 to -80 -40
+	button f "_Find..."
+		from -170 -90 to -80 -40
```
endless-sky: Having so many buttons in the bottom right corner that they have to be arranged in a grid two rows high is rather awful UI design. The +/- buttons are already bad enough, but adding this makes the UI look even more cluttered.

tehhowch: I wanted to go for the lower left corner, but it overlaps the pointer panel when viewing missions.
My second choice was above the +/- buttons, but we'd need to add a rotation keyword to the interface sprite specification (or add a a rotated version of ui/dialog cancel sprite, which seemed like too much) to get the sprite decorations between the +/- and the dialog sprite to play nicely.

Issue #3428, opened 2017-12-16
Secondary icons


comnom:
When doing an interface plugin, you can easily override the "ammo" box to suit the style, but the same isn't true of the secondary icons since the "background" is part of the icon sprite and not the interface element. 

I made an icon set with a transparent background and an ammo box that includes the icon background, but that had some draw order problems (the icon had to be drawn last, which puts the drop shadow on top of the ammo box).

I didn't want future interface plugins to have to do a full icon set, so whatever I do I'd like to PR.

My current idea is to do an icon set that is just the missile and drop shadow, and then add an interface element for the icon background so that I don't have to chage the draw order. So we'd end up with an extra element, and a draw order that's: "icon background" -> icon -> "ammo box".

Is there a better way to do this?

tehhowch:
Got some photos to help convey this info? That'd help us some :)

I think the ammo box should just always contain the icon, i.e. the icon should be scaled to for within the box. We'd define the "icon alignment" so you can set where in the ammo box sprite the icon goes, an "icon height" (and width?) to control the size, and a point for where the ammo count goes.

comnom:
> I think the ammo box should just always contain the icon

That lead to the first problem, where the icon drop shadow overlaps the middle of the ammo box.
![ammo](https://user-images.githubusercontent.com/22107649/34073705-a27d4f66-e26d-11e7-8288-998c1fe0cc1f.png)

So you'd still have to edit each icon to cut out that overlapping part (meaning if you change the ammo element, you have to do an icon set). With a separate element, you can draw the ammo box on top and (assuming full opacity) clip the drop shadow.

comnom:
So the proposal would be three parts:
The icon:
![finisher](https://user-images.githubusercontent.com/22107649/34073786-de2df5a4-e26f-11e7-91cc-fae8ae761440.png)

The "icon background" element:
![background](https://user-images.githubusercontent.com/22107649/34073790-e7e14394-e26f-11e7-829d-e7400b8a1b23.png)

And the "ammo box" which clips the drop shadow:
![ammo box](https://user-images.githubusercontent.com/22107649/34073800-f82de310-e26f-11e7-8ea3-1b6afe29f0dd.png)

This is how it currently works, only the first two are one sprite.

tehhowch:
And if we make the icons not have the middle element at all - relying on the ammo box to supply the background? Is there anything preventing us from doing that?

comnom:
[...]

PR #3434, opened 2017-12-18
draw icon background separate from icons

comnom:
refs: https://github.com/endless-sky/endless-sky/issues/3428

I took the simplest code change possible and went with icon + halo with a separate "icon background" sprite.

Ran into a couple snags with the gat icon and rail slug icon, which had an ugly border that was hidden by the background.

Also took the liberty of standardizing the alignment and making sure all icons fit within their "bounding box" (the background), as the typhoon 1x was aligned this way, but wasn't in the 2x, and the tracker had it's own 45 degree-ish alignment going on, so I hope that's ok.

Summary:
* Draw the background separate from the icon in Engine.cpp
* New icon set with transparent backgrounds
* Reworked gat and rail icons to get rid of the border
* Aligned several icons that were outliers
* Scaled several icons that broke their bounding box

So more than I wanted to do really, but the icons now have a standard orientation (within a pixel) and size (rough margin of 1-2px from the bounding box).

I moved the gat and rail into the xcf with the others which is here:
[icon.zip](https://github.com/endless-sky/endless-sky/files/3041189/icon.zip)


tehhowch:
@comnom what's your stance on updating the copyright files for the reworked icons?

tehhowch (dismissed):

comnom:
>what's your stance on updating the copyright files for the reworked icons?

Not necessary. All I did was re-render from existing models with a slight color touch-up to get a clean image to use.

I'd like to hold off on this being merged as is, as I was hoping to get some feedback from @endless-sky . Specifically, the goal of the PR is easier modding, and that's actually better served by icons without the "halo" we discussed in the [issue](https://github.com/endless-sky/endless-sky/issues/3428).

I didn't make that change because it alters the design (rather than splitting the existing design into two parts), and I wasn't comfortable doing so without feedback. Then I kinda totally forgot about it, so I never requested feedback ;) .

Pointedstick:
Is there a High DPI PR that goes with this?

comnom:
>Is there a High DPI PR that goes with this?

Not yet. I was still just adding the 2x to the main post for the dev to handle when I initially did this. When I'm satisfied, I'll be sure to put one up.

I'm still arguing with myself about doing a set without the halo/drop shadow, as the current implementation only allows changing the color without doing an icon set, but you have to do a whole set to change the shape (ex: https://github.com/comnom/es-huds/blob/master/pug-hud/images/icon/finisher%402x.png).

I've been meaning to look through the assets to see if there's a good case for writing a "drop shadow" shader (nice for highlighting buttons and such without needing a separate asset, as well as plain artsy stuff like we have here), but I just haven't gotten around to it.

comnom:
>I've been meaning to look through the assets to see if there's a good case for writing a "drop shadow" shader...

There isn't. Only three sprites are used solely for selection highlighting. I don't want to change the design either, so I'm satisfied.

High dpi: https://github.com/endless-sky/endless-sky-high-dpi/pull/74

petervdmeer:
Both Amazinite and Tehhowch already approved this PR, but it is no longer ready for merging due to a merge-conflict in the code and due to 3 weapons being added in the meantime (dragonflame, mhd and shard) for which the icons would likely also need to be changed to a transparent background (both in this PR as well as in the high-DPI PR). Adding "Waiting on OP" label.

petervdmeer:
Both Amazinite and Tehhowch already approved this PR, but it is no longer ready for merging due to a merge-conflict in the code and due to 3 weapons being added in the meantime (dragonflame, mhd and shard) for which the icons would likely also need to be changed to a transparent background (both in this PR as well as in the high-DPI PR). Adding "Waiting on OP" label.

Amazinite:

PR #3440, opened 2017-12-19
Fixed some undefined behavior at startup.

Rakete1111:
At the start of the game, `event.ActorGovernment()` is a null pointer, and dereferencing it is UB. Additionally, some ships do not provide active cooling nor heat dissipation or some other cooling properties, which results in a division by zero. Timer Ship suffers from this problem.

I'm not to sure what `Ship::IdleHeat` actually does, but it seems like the correct behavior is to return 0 in that case.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2146,7 +2146,8 @@ double Ship::IdleHeat() const
 	// heat * (1 - diss + activeCool / (100 * mass)) = (heatGen - cool)
 	double production = max(0., attributes.Get("heat generation") - cooling);
 	double dissipation = .001 * attributes.Get("heat dissipation") + activeCooling / (100. * Mass());
-	return production / dissipation;
+
+	return dissipation != 0. ? production / dissipation : 0.;
 }
 
 
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -31,6 +31,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include <algorithm>
 #include <cmath>
 #include <iostream>
+#include <limits>
 
 using namespace std;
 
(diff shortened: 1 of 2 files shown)


tehhowch:
Idle heat = the ship's temperature without using engines or weapons or regen equipment. Without dissipation, it should be +infinity as the ship cannot shed heat, and thus it can only gain or store heat.

If someone is making ships without "heat dissipation" they're doing it wrong and deserve their overheated ship. :grin:

Rakete1111:
@tehhowch Ah ok. Thanks for clarifying, fixed :)

source/Ship.cpp
```diff
@@ -2146,7 +2147,8 @@ double Ship::IdleHeat() const
 	// heat * (1 - diss + activeCool / (100 * mass)) = (heatGen - cool)
 	double production = max(0., attributes.Get("heat generation") - cooling);
 	double dissipation = .001 * attributes.Get("heat dissipation") + activeCooling / (100. * Mass());
-	return production / dissipation;
+
+	return dissipation != 0. ? production / dissipation : numeric_limits<double>::infinity();
```
tehhowch: If there is no dissipation and no production, then the ship's IdleHeat is its current heat, not infinity.
If there is no dissipation and any production, then the ship's IdleHeat is +infinity

Rakete1111: @tehhowch Why with both null, does it return the current heat? The heat at idle doesn't depend on the current heat, no?

tehhowch: the idle heat is whatever heat it has where its production is exactly offset by its dissipation. For 0 and 0, that is every possible value of heat.

Rakete1111: @tehhowch Ohh. Now it makes sense! :) Thank you

endless-sky:
Can you give an example of a case where NPC::Do() is called with an event that has no government defined?

Also, your second change is unnecessary, because 1. all ships must define a heat dissipation value, and 2. if they don't, that function will return infinity anyway (a nonzero value dividied by zero equals infinity).

Rakete1111:
@endless-sky But not on every platform. Most platform use IEEE754, which has that behavior. Not sure if it's worth it though to change it.

I don't really know how that is possible, I caught this by clang's UB analyser. I'll look into it.

Issue #3383, opened 2017-12-08
add a "frugal" personality for carriers


Lorantine:
an explanation on what this feature would be:
imagine "frugal", but then for carried ships (drones/fighters), because now a big Navy fleet will launch all their carried ships if there's just 1 hostile Interceptor, which is straight-up overkill, and they even spend a lot of time to redock

so basically fleets (ships) with this new personality would only launch their carried ships under the same circumstances as missiles would be used when the frugal personality is active (getting low on health or getting outgunned)

from my 1st reply
> ships don't fire missiles if they have the frugal personality unless they're nearing death or they're getting outnumbered
what I want is a personality that makes carriers not deploy their fighters unless they're nearing death or they're getting outnumbered as well, so for example a big Navy fleet doesn't deploy their fighters for 1 Interceptor

tehhowch:
From a carrier perspective, provided the fighters deployed are not using missile weaponry, they have a better chance of recovering all launched fighters if they launch all their fighters. So to be truly "frugal" and have the lowest cost per sortie they should launch everything they can (which is the current behavior).

If their fighters use missile technology, then it is possible (but not guaranteed) that launching some fighters rather than all will still result in a killed target without permanent injury to the carrier or fighter. So the preferred survival technique again is to launch all fighters

KiLEdEnNis:
it's a bit unclear what exactly ya mean and I think that's because I didn't make my point clear enough
my point was that carriers should not deploy any fighters the same way ships don't fire missiles when they have the frugal personality
ships don't fire missiles if they have the frugal personality unless they're nearing death or they're getting outnumbered
what I want is a personality that makes carriers not deploy their fighters unless they're nearing death or they're getting outnumbered as well, so for example a big Navy fleet doesn't deploy their fighters for 1 Interceptor

tehhowch:
If a carrier waits that long to launch, they are likely to lose that fight when they could have won it if they launched fighters earlier. 

Right now the decision to deploy is an automatic side effect of the ship having a target to attack (every ship issues the deploy command if it is not yours and has a target). We could gate this behind a strength check,  but I don't think it will get much use in the base game.

Any particular personality keyword you're imagining?

KiLEdEnNis:
isn't it as soon as they may be outnumbered they deploy them? the sending them out if their health is getting low is only to make them deploy their fighters even though they're not outnumbered
so
[...]

PR #3441, opened 2017-12-19
added a "prudent" personality for #3383

Lorantine:
so with a lot of blood, sweat, confusion, dumb errors and a lot of patience from @tehhowch & @Hadron1776 who helped me a LOT (teaching/reviewing/pointing out idiocy) I managed to add the "prudent" personality (from #3383)

what it does:
ships with the prudent personality will only deploy fighters/drones under the exact same circumstances as frugal ships would use missiles:
when the ship lost 50% hull or shield, or when the ship and its allies are outgunned

in other words:
ships with the prudent personality will only launch their fighters/drones if they are outgunned or they lost a significant portion of their health

I still don't exactly understand everything but I have some minor C# knowledge which isn't that different, so I'll be able to edit it based on feedback

& again big thanks to Tehhowch & Hadron* ~~(no idea why I thought it was Comnom, Hadron was the one that said "define that bool", oops)~~, 0 chance I could have done this without them


KiLEdEnNis:
already found it way too far indented, I'll get to it tomorrow, thanks

KiLEdEnNis:
got Code::Blocks on my laptop (couldn't wait)
fixed the feedback I got so merge-ready?
(after merging, we can look which fleets should have the personality, and of course have MZ add it to the Wiki)

source/AI.cpp
```diff
@@ -1737,7 +1737,21 @@ void AI::Attack(Ship &ship, Command &command, const Ship &target)
...
+		bool beConserving = false;
+		if(ship.GetPersonality().IsConserving())
+		{
+			beConserving = (ship.Hull() + ship.Shields() > 1.5);
+			auto ait = allyStrength.find(ship.GetGovernment());
+			auto eit = enemyStrength.find(ship.GetGovernment());
+			if(ait != allyStrength.end() && eit != enemyStrength.end() && ait->second < eit->second)
+				beConserving = false;
+		}
+	if(!beConserving)
```
tehhowch: This line and the next line below it were correctly indented before your latest commit. Now, they are incorrectly indented.

KiLEdEnNis: oh ye I just saw why, thanks again

KiLEdEnNis:
what the hell did it just do

tehhowch:
Now your entire addition to the AI file is indented one step too many.

KiLEdEnNis:
yeah just saw, no idea what was going on, it undid my fix

source/AI.cpp
```diff
@@ -1737,7 +1737,21 @@ void AI::Attack(Ship &ship, Command &command, const Ship &target)
...
+		bool beConserving = false;
+		if(ship.GetPersonality().IsConserving())
+		{
+			beConserving = (ship.Hull() + ship.Shields() > 1.5);
+			auto ait = allyStrength.find(ship.GetGovernment());
+			auto eit = enemyStrength.find(ship.GetGovernment());
+			if(ait != allyStrength.end() && eit != enemyStrength.end() && ait->second < eit->second)
+				beConserving = false;
+		}
+			if(!beConserving)
```
tehhowch: This and the line below it are indented one level too many.

KiLEdEnNis:
should be fine now? (5 commits of messing around for no reason but ok)

do not merge yet, because I have this to try & do:
> Given that this logic is an exact copy of the logic for frugal use of ammo, it should probably be refactored into a private method so changes to the logic need only be done in one area.

source/AI.cpp
Amazinite: This must be a real struggle.

KiLEdEnNis: didn't I fix that?

KiLEdEnNis:
I'll leave it at this tbh, unless someone that actually knows C++ wants to do Tehhowch's suggestion, this is merge ready

source/AI.cpp
```diff
@@ -1736,8 +1736,8 @@ void AI::Attack(Ship &ship, Command &command, const Ship &target)
 		shortestRange = 0.;
 	
 	// Deploy any fighters you are carrying.
-	if(!ship.IsYours())
-		command |= Command::DEPLOY;
+	if(!ship.IsYours() && (!(ship.GetPersonality().IsConserving() && ShouldActFrugally(ship))))
+			command |= Command::DEPLOY;
```
Hadron1776: There's an extra tab here

Hadron1776: There's an extra tab here

source/AI.cpp
```diff
@@ -2137,6 +2137,17 @@ void AI::DoScatter(Ship &ship, Command &command)
 }
 
 
+const bool AI::ShouldActFrugally(const Ship &ship) const
```
tehhowch: Should just be bool - no need to make it's return value unmodifiable.

source/AI.cpp
```diff
@@ -2137,6 +2137,17 @@ void AI::DoScatter(Ship &ship, Command &command)
...
+{
+	auto ait = allyStrength.find(ship.GetGovernment());
```
tehhowch: You can remove these and simplify the logic of this method by calling AI::AllyStrength and AI::EnemyStrength, as those handle the iterators and give you the strength value you are seeking.

i.e. `if(AllyStrength < EnemyStrength || health check)` -> don't act frugally, otherwise act frugally.

source/AI.cpp
```diff
@@ -1701,7 +1701,7 @@ void AI::KeepStation(Ship &ship, Command &command, const Ship &target)
 
 
 
-void AI::Attack(Ship &ship, Command &command, const Ship &target)
+const void AI::Attack(Ship &ship, Command &command, const Ship &target) const
```
tehhowch: const void? just use void.

source/AI.cpp
```diff
@@ -1736,7 +1736,7 @@ void AI::Attack(Ship &ship, Command &command, const Ship &target)
 		shortestRange = 0.;
 	
 	// Deploy any fighters you are carrying.
-	if(!ship.IsYours())
+	if(!ship.IsYours() && (!(ship.GetPersonality().IsConserving() && ShouldActFrugally(ship))))
```
tehhowch: You have an extra set of parentheses here.

source/AI.cpp
```diff
@@ -2374,13 +2385,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 	bool beFrugal = (ship.IsYours() && !escortsUseAmmo);
 	if(person.IsFrugal() || (ship.IsYours() && escortsAreFrugal && escortsUseAmmo))
 	{
```
tehhowch: You can probably modify the above lines, but at minimum you can drop the parenthesis here now since there is only one statement following the `if` branch.

source/AI.h
```diff
@@ -89,7 +89,7 @@ template <class Type>
 	static void CircleAround(Ship &ship, Command &command, const Ship &target);
 	static void Swarm(Ship &ship, Command &command, const Ship &target);
 	static void KeepStation(Ship &ship, Command &command, const Ship &target);
-	static void Attack(Ship &ship, Command &command, const Ship &target);
+	const void Attack(Ship &ship, Command &command, const Ship &target) const;
```
tehhowch: don't forget to remove the first `const` here

source/AI.h
```diff
@@ -98,6 +98,7 @@ template <class Type>
 	bool DoHarvesting(Ship &ship, Command &command);
 	void DoCloak(Ship &ship, Command &command);
 	void DoScatter(Ship &ship, Command &command);
+	const bool ShouldActFrugally(const Ship &ship) const;
```
tehhowch: or the first one here

source/AI.h
KiLEdEnNis: did the other one as ya suggested remove this or the other

source/AI.cpp
KiLEdEnNis: still have to do this

source/AI.h
Hadron1776: You'll still need to remove the first `const` here - otherwise, there will be compile errors as the implementation isn't `const`.

tehhowch: The first const here refers to the return type. Returning `const void` is nonsense. The second const indicates that this member function does not modify variables that belong to the AI class (which is true).

KiLEdEnNis: so remove the first const here
should I move this line above void DoSwarming? since it's just a void then

KiLEdEnNis:
unless I messed something up or someone has any feedback, this should be merge-ready? (I hope)

KiLEdEnNis:
> I'm still not set on "conserving" as the personality token though.

me neither, I considered other options for hours, but the only thing that made more sense was renaming the current frugal to mfrugal (as in missile frugal) and then add cfrugal (like carrier frugal)

KiLEdEnNis:
I renamed the personality to prudent, which is about the best personality name I could think of

merge-ready now, at last(?)

Tadrix:
A last minute question: Shouldn't carriers launch their craft when their hull is at a bit higher threshold, say, 75%? They become disabled at 20%+ hull, so launching their craft at 50% hull (meaning their shields aren't effectively protecting them) could be too late.

KiLEdEnNis:
it has the same requirements as the frugal personality, as for the health scenario
imagine hull between 0-100%, and shield between 0-100%, together they make 200%
if the ship has 150% or less, the requirement is met and thus frugal ships start using missiles and prudent ships would deploy fighters
this means that if the hull is 100% and the shield is 50% it deploys them, or if hull is 80% and shield is 70%
technically it's possible a carrier would deploy at 50% hull, but this means the shields have to be full, and this scenario is basically non-existent

and even if I'd want to create a better check (a 3rd if-statement that checks if the hull is 75% or more), this would affect the way frugal works as well, and that's not really part of this pull request
what this PR does is add the prudent personality; the "frugal" tag for deploying fighters, it's not about changing the way frugal (and thus prudent) works, although this 3rd check could be a good idea for another PR

KiLEdEnNis:
I just realized this still exist and I have no clue if it still works, plus I don't really see anyone but me wanting to merge this
so unless someone wants it open again, I'll leave it to rot in the cold void

PR #3443, opened 2017-12-20
Typos in Planet Descriptions

Alkallid:


data/map.txt
```diff
@@ -26938,7 +26938,7 @@ planet Twinstar
 	attributes south tourism frontier moon
 	landscape land/lava3
 	description `The only settlement on this small world is Twinstar Depot, a village created mostly just to service the spaceport, where food and equipment from the southern galactic arm is stored and sent out in all directions to supply the rest of human space. The gravity is far less than ideal for human development, but the atmosphere is breathable and some tourists come to the planet just for the experience of being able to run and jump in low gravity without the need for a suit or oxygen tanks.`
-	spaceport `The depot consists mostly of large storage racks, with automatic lifts constantly clanking up and down. Very few people are in sight, aside from a small cafe near the center of the port. It is not a hospitable place.`
+	spaceport `The depot consists mostly of large storage racks, with automatic lifts constantly clanking up and down. Very few people are in sight, aside from a small cafe near the center of the port. It is not a very hospitable place.`
```
Bladewood: `typo`

Apparently, "[hospitable](http://www.dictionary.com/browse/hospitable?s=t)" means only that it is favorable to others, which is not the same thing as "inhabitable", so technically the original is correct.

endless-sky:
The first change is definitely a typo fix, but the second (hospitable -> very hospitable) is just a stylistic change, and not an improvement in my opinion.

PR #3444, opened 2017-12-20
Replaced the Repeater sound with one that has a bit more presence

Pointedstick:
Current Repeater sound: https://webmshare.com/dvPaO

Proposed new Repeater sound: https://webmshare.com/VGV7Z

The proposed new one has a bit more battlefield presence, and sounds less clicky. It is Public Domain, from freesound.org, heavily processed in Audacity.


tehhowch:
The real test... when 30 drones are shooting at you?

Pointedstick:
It sounds pretty good in a battle. Not quite the same, but here's how it sounds with the player firing continuously at many adversaries: https://webmshare.com/mPR5J

tehhowch:
Given that Remnant weapons don't have their own sounds, could it perhaps be used there?
Could be a good Thrasher sound.

Pointedstick:
I'd be happy to do that too, but this is about the Repeater. To my ears at least, the existing Repeater sound isn't very impressive. Its clickiness is more annoying than weapon-like, and it is certainly a dangerous weapon. I thought it needed a more impressive sound. If nobody likes the proposed new sound, I can try again.

tehhowch:
The real problem with Sestor sounds is that the Warder drowns out everything else since it fires sooooo often in Sestor vs Mereti battles - it drowns out pretty much everything, and has a lot of "clicky" feel to it as well.
(Tried to listen to this sound in a fleet fight... heard only the Warders once the two sides get close together.)

tehhowch:
This new sound would at least differentiate the warder click from the repeater in a noticeable manner.

endless-sky:
As with the blaster, the repeater has such a high rate of fire that the sound needs to be short to avoid using up all the available audio channels in some OpenAL implementations. This sound is four times as long as the new one, meaning that you can have up to roughly 35-40 repeater sounds playing at once during a major battle, instead of about 8-12.

Pointedstick:
I've reduced the length to 0.42 seconds, but any shorter that that and it just doesn't sound very good anymore. If that's still not short enough, I may need to go back to the drawing board,

MessyMix:
I agree that the old sound was too clicky, but I'm not a fan of the new sound, either; personally, it sounds too mellow for a weapon that's meant to be staccato. I think that it would be a good fit for Remnant weapons, however.

Lineth:
I always thought a harsher, raspier staccato cracking sound would fit for the repeater.

Pointedstick:
I'll see what I can do. In the meantime, please feel free to propose your own, of course!

endless-sky:
Also, what's the source for the new sound, and what license is it under?

Pointedstick:
Freesound.org, public domain

PR #3451, opened 2017-12-21
More ship names!

Pointedstick:
You know the drill.


data/names.txt
```diff
@@ -1029,6 +1060,19 @@ phrase "standalone civilian ship names"
 		"Tiamat"
 		"Artemis"
 		"Annabel Lee"
+		"Old Ironsides"
+		"20 Chanri"
```
Bladewood: We sure putting digits in (human) ship names is okay?

Pointedstick: Why wouldn't it be?

Bladewood: Just that no other human ship name follows that pattern, as far as I know. They all use the full word forms, not digit forms.

KiLEdEnNis: it feels more [robot-like? non-personal?] to me, but a few of them shouldn't be a problem

PR #3455, opened 2017-12-22
Update AppStream file to standardize ID and add version, release date, and changelog information

### LINKED ISSUE #3453: Appstream IDs differ between appdata.xml file and Flatpak metadata
Endless Sky's Appstream ID on Flathub is `io.github.EndlessSky.endless-sky` (https://github.com/flathub/io.github.EndlessSky.endless-sky/blob/master/io.github.EndlessSky.endless-sky.json#L2)

But Endless Sky's AppStream file uses a different AppStream ID: `endless-sky.desktop` (https://github.com/endless-sky/endless-sky/blob/master/endless-sky.appdata.xml#L3)

This causes problems: if the same app has different AppStream IDs in different distribution channels, then it'll confusingly show up as different apps in clients that can display content from multiple sources, such as GNOME Software and KDE Discover. Here's what it looks like in KDE discover:

two endless skies

The top one is the Flathub version, the bottom one is provided by my distro's packages.

If you unify the AppStream IDs (ideally standardizing on a reverse-DNS style such as `io.github.EndlessSky.endless-sky`), then Discover and GNOME Software (and similar apps) can perform de-duplication correctly, marking both as the same app and offering the user the choice of which version they want to install.

(also, ideally the description text, icon, and screenshots can be standardized between the Flatpak metadata and the appstream file).

### LINKED ISSUE #3461: AppStream metadata file should include release, version, and changelog information
This will make life easier for downstream packagers. Ideally this information would be updated for every release. Relevant part of the AppStream spec: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#sect-Metadata-GenericComponent

What a lot of folks are doing is using the AppStream data as the canonical source of truth and using that to generate other metadata files as necessary (e.g. changelog.txt).

Pointedstick:
Resolves #3453
Resolves #3461

I've opted to not to rename the desktop file and instead add the `<launchable>` tag to associate the AppStream data with the desktop file. An alternative approach is to remove/not add the `<launchable>` tag, and instead rename the desktop file to match the AppStream ID (i.e. `io.github.EndlessSky.endless-sky.desktop`).

Relevant part of the AppStream spec: https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

### DIFF
--- a/endless-sky.appdata.xml
+++ b/endless-sky.appdata.xml
@@ -1,6 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component type="desktop">
   <id>io.github.EndlessSky.endless-sky</id>
+  <launchable type="desktop-id">endless-sky.desktop</launchable>
   <name>Endless Sky</name>
   <name xml:lang="de">Weltraumspiel</name>
   <name xml:lang="fr">Jeu spatial</name>
--- a/endless-sky.appdata.xml
+++ b/endless-sky.appdata.xml
@@ -18,7 +18,7 @@
   <p>
    Endless Sky is a sandbox-style space exploration game similar to Elite, Escape Velocity, or Star Control. You start out as the captain of a tiny space ship and can choose what to do from there. The game includes a major plot line and many minor missions, but you can choose whether you want to play through the plot or strike out on your own as a merchant or bounty hunter or explorer.
   </p>
-</description>
+  </description>
   <url type="bugtracker">https://github.com/endless-sky/endless-sky/issues</url>
   <url type="help">https://github.com/endless-sky/endless-sky/wiki/PlayersManual</url>
   <url type="homepage">https://endless-sky.github.io/</url>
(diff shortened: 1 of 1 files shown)


Pointedstick:
This is part of my push to make Linux software actually use the AppStream spec: https://github.com/flathub/flathub/issues/195#event-1400308483. If you're interested, @tehhowch, I would be enormously appreciative of a helping hand. It's a rather huge job since there is a *lot* of upstream software that does not make full or even any use of AppStream, which is a shame since it offers a lot of benefits for upstream software:
- Greater control over release communications
- Better presentation to users by making it easier for downstream packagers to add version numbers, changelogs, etc
- Information unification across all distribution channels

endless-sky:
As I said in #3453, the identifier in reverse-URL format should be `io.github.endless-sky.EndlessSky`, not `io.github.EndlessSky.endless-sky`.

Pointedstick:
And as I said in #3453 (:slightly_smiling_face:), dashes aren't allowed anywhere other than the last element of the string, so it has to be `io.github.EndlessSky.endless-sky`.

PR #3458, opened 2017-12-22
planet sprites

comnom:
Didn't manage to snag any other contributers despite having it up on steam for five months so it took a while, but here's the whole set.

I just replaced half of the occurences of a given sprite (rounded down) using a random sample.

2x here (updated):
[2x.zip](https://github.com/endless-sky/endless-sky/files/1592252/2x.zip)

xcf will be here once its up on google drive:
https://drive.google.com/file/d/1l9E9I12EEi0VZ0pxgRCZGBJBz0-bIDGs/view?usp=sharing


comnom:
>What's the usage count of each sprite now?

Doesn't count excluded sprites(stations etc).
Counts before:
[out.txt](https://github.com/endless-sky/endless-sky/files/1583528/out.txt)

Counts after:
[outNew.txt](https://github.com/endless-sky/endless-sky/files/1583529/outNew.txt)

endless-sky:
These look really good!

For a few of them, the "b" sprite is not the same size as the original, at least based on the sizes in the 2x zip file:

* desert1
* dust7
* gas1
* lava3
* rock1
* water1

Also, I'm a little worried about cases where the sprite has a very different color, and the landscape photo for a planet was based on the color of the planet sprite. Those would need to be reviewed individually to make sure they still work.

comnom:
> For a few of them, the "b" sprite is not the same size as the original, at least based on the sizes in the 2x zip file.

Fixed the size issues in https://github.com/endless-sky/endless-sky/pull/3458/commits/59ad726bc929ed9e28841579a33bb09d81fd3bdf . However, for some I didn't see a size difference, maybe you meant another sprite?

Comparing the zip against the current high-dpi master I get:

* desert1 at 400x400
* gas1 at 1600x960
* rock1 at 340x340

Just in case I've edited the first post with a new copy of the 2x.zip with the size changes.

> Those would need to be reviewed individually to make sure they still work.

I'll see about a script that'll pull any named objects that got a b variant and start reviewing those. Probably take a bit (maybe more than a bit depending on how many there are).

KiLEdEnNis:
what did ya change to water1-b? I can't see any difference

comnom:
> what did ya change to water1-b? I can't see any difference

Nothing. Only the 2x changed size. It's just habit to save in pairs. I'd have to look at it in image magick to know for sure why the hash changed.

KiLEdEnNis:
if ya didn't change anything then nothing changed at all; it just resaved as if it's new then

comnom:
I took the easy route and just reverted the changes to all named objects in https://github.com/endless-sky/endless-sky/pull/3458/commits/862d06c97768d14397196cafdcbd9182ff01f5f5 .

I was about fifty into the review when I realized some of the descriptions referred to the planet sprite, so going through 142 descriptions and checking the landscapes seemed like a waste of time. There's still around 1000 uses of the b variants, so it's a small reversion in the grand scheme.

PR #3468, opened 2017-12-26
Use "join the defenses" rather than "join in the defenses" in defense jobs

Fzzr:
Another minor grammar/spelling thing. "Defense" can refer to the _action_ of defending something or the _means_ of defending something. Usually the latter sense is pluralized and the former is not. For example: "In the _defense_ against their foe, the militia made great use of their static _defenses_." In the context of the various defense missions/jobs, we're using the singular "defense" sense, and so should not be using the plural. Another way to look at it is if the word "defenses" could be replaced with the word "battle" without changing the meaning, it should be changed to the singular "defense".


czartrak:
I think they refer it to defenses because it's multiple ships. kinda lie they say "flood waters" even though it's one flood

Fzzr:
In that case it would be "join the defenders". Which would actually be an equally valid change, and if that's preferable.

czartrak:
Each ship is considered a defense within itself, so thus, defenses

Fzzr:
Well, as written it's "aid _in_ the defenses", which doesn't still doesn't follow. Possible variants:

- join the defenders
- aid the defenses
- aid in the defense

Any of those work, it's only the current one that doesn't use the right pluraliation

KiLEdEnNis:
it's 1 defense with multiple defenders

imagine if ya were on the Pirate side: "join the attacks", attackS, while it's 1 attack, or even invasion; it's 1 invasion

endless-sky:
I'm not convinced that "defense" works better. If you're talking about the planet's ability to defend itself, "defenses" makes more sense. If you're talking about the defense force itself, "defense" makes more sense. So, it's just a matter of which way you conceptualize it.

Fzzr:
Yeah, that's what I meant by https://github.com/endless-sky/endless-sky/pull/3468#issuecomment-354240466. If the intent is the "means of defending something" meaning, "aid the defenses" makes more sense grammatically.

Fzzr:
I read through it all again, and decided that this version ("join the defenses") changes the meaning the least of all the options. Updated to include new defense jobs added in #3325.

data/pirate jobs.txt
```diff
@@ -2311,7 +2311,7 @@ mission "Core Pirate Defense"
 		fleet "Small Core Pirates" 2
 	on offer
 		conversation
-			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by the Syndicate! We need every combat-worthy ship to join in the defense!" This could also be an easy way to get yourself killed, and you can't even guarantee that you will be paid for your work.`
+			`Suddenly you hear raised voices and shouting outside: "We are under attack! <planet> is under attack by the Syndicate! We need every combat-worthy ship to join thes defenses!" This could also be an easy way to get yourself killed, and you can't even guarantee that you will be paid for your work.`
```
endless-sky: Typo.

Fzzr: Thanks, good catch. Gonna resolve conflict now as well.

Fzzr:
Typo was overridden by #3482. Fixed some inconsistencies across the two PRs in favor of that one.

PR #3478, opened 2017-12-30
reduced short term value depreciation

Nescio0:
Currently `FULL_DEPRECIATION = 0.25` (i.e. minimum value is 25%) and `DAILY_DEPRECIATION = 0.99`; this proposal changes these value to 0.2 and 0.995 respectively. The resulting values (see lines 331 to 343) are:

```
            current          proposed
one day   : 0.992  (-0.8%) ; 0.995  (-0.5%)
one week  : 0.944  (-5.6%) ; 0.967  (-3.3%)
one month : 0.788 (-21.2%) ; 0.868 (-13.2%)
one year  : 0.262 (-73.8%) ; 0.281 (-71.9%)
two years : 0.250 (-75.0%) ; 0.206 (-79.4%)
```

Effectively this proposal has the following effects:
* short term value loss is lessened, to make it less punitive to try out new things and to travel around to purchase outfits in other areas of space
* the difference after a year is hardly noticeable
* seizing ships is somewhat less profitable, because minimum value is lowered

PS To keep the minimum value of gatling gun bullets at 1, its cost is also increased.
(`5 * 0.2 = 4 * 0.25`)

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -1048,7 +1048,7 @@ effect "nuke residue slow"
 outfit "Gatling Gun Ammo"
 	plural "Gatling Gun Ammo"
 	category "Ammunition"
-	cost 4
+	cost 5
 	thumbnail "outfit/bullet"
 	"mass" .002
 	"gatling round capacity" -1
--- a/source/Depreciation.cpp
+++ b/source/Depreciation.cpp
@@ -27,8 +27,8 @@ namespace {
 	// Names for the two kinds of depreciation records.
 	string NAME[2] = {"fleet depreciation", "stock depreciation"};
 	// Depreciation parameters.
-	double FULL_DEPRECIATION = 0.25;
-	double DAILY_DEPRECIATION = 0.99;
+	double FULL_DEPRECIATION = 0.2;
+	double DAILY_DEPRECIATION = 0.995;
 	int MAX_AGE = 1000;
 }
 


jafdy:
This slows depreciation and lets it continue until the value is lower. So, compared to what it used to be, the new values give 0.5% less value loss per day with a 5% value loss after a long time. The loss in value is still the same curve so not really an improvement to the system, just a rejigging.

Nescio0:
>  The loss in value is still the same curve so not really an improvement to the system, just a rejigging.

Exactly. This proposal is a tweak, not an overhaul. The current behaviour of the curve is fine (a power law in combination with a linear function, steep start, slow tail), only the precise parameter values are changed.

KiLEdEnNis:
plundering/capping for money instead of the tech ya acquired isn't really that broken anymore tbh, I'm nearly as tempted to do cargo jobs instead of capping ships
I think it's still the fastest way to get money, but one needs to land on a shipyard planet & keep the ship alive (if capping), it's a bit of a hassle already, so I'm not sure if this should be "punished" even more
the other part this depreciation influences is players selling their ship or some outfits to upgrade, and this shouldn't be "punished" any harder than it is already

so a summary: for me it's a no

ghoulavenger:
Capturing and plundering is significantly worse right now to running cargo fleets. I see no reason to nerf capturing more. It's only strength really is on a per game day basis, and that can be overcome with a large enough cargo fleet. I averaged 15-20M a day on my best capturing runs, where my best cargo runs are 40M per jump (and I could make it bigger if I wanted to). The only weakness to running a cargo fleet is the very beginning of the game (cargo fleets need many more ships to make profit). Slowing the rate of decay on depreciation though isn't necessarily a bad thing, but also largely unnecessary.

endless-sky:
I think that capturing and plundering is already pretty well balanced right now compared to other ways of making money. Aside from cheats like going around a battlefield in an unarmed ship so you don't get attacked while plundering (which could be addressed by something like #3230), there doesn't seem to be a need for further nerfing.

Issue #3480, opened 2018-01-02
Mass in the latest commit is absolutely broken


Bladewood:
In the latest commit, owned ships' masses are initially treated normally.
But once you reload your save, the ships' base masses no longer count toward the total mass. That is, only their outfits count toward the mass.
This means that a Punisher, for example, moves twice as fast after reloading.

On that note, hit force is also very weird right now. A heavy rocket and even a meteor missile have far more knockback than a particle cannon, while something like a finisher barely pushes ships around at all.

Also, as @comnom noted, outfit masses aren't even displayed in the outfitter...

Hacklin:
Confirmed.

The ship's 'mass' attribute is not being saved.
Therefore, when you reload, a ship's hull mass is zero.

Note:
If you remove all outfits from a ship with a hull mass of zero,
the ship's 'acceleration' and 'turning' values in the outfitter are '0 / -nan'  ;-)

endless-sky:
Fixed by merging #3481.

PR #3481, opened 2018-01-02
Save a ship's base mass.

Hacklin:
Fix issue #3480

### DIFF
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -18,6 +18,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "GameData.h"
 #include "SpriteSet.h"
 
+#include <cstring>
 #include <cmath>
 
 using namespace std;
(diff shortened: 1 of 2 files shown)


Hacklin:
The OutfitInfoDisplay is a separate issue.
Not fixing it in this pull request.

endless-sky:
Given that the bug was only present for a day, I'll probably leave a week or two for people to use this code to mend any issues with their saved games, then revert the changes that were only work-arounds for the bug.

Hacklin:
Fine.
You should revert all changes to 'source/Outfit.cpp' and
the ship's chassis mass check in 'source/Ship.cpp' lines 373-387.

However, I'm not sure about the ships with a base class in 'source/Ship.cpp' lines 106-107.
These 2 lines shouldn't be needed (and can be reverted),
as a ship with any base attributes should also have a base/chassis mass.

Just leave the 'Save a ship's base mass.' commit.

Note:
I changed the `Outfit::Add()` and `Outfit::Reset()` methods,
but in hindsight I should just have added an `Outfit::ResetMass()` method.

Hacklin:
@endless-sky: What do you prefer, separate git branches for fixes and workarounds or separate commits for fixes and workarounds in the same git branch?

endless-sky:
In this case it made sense to put them together, for the sake of anyone who needed to patch a corrupted game.

PR #3482, opened 2018-01-02
Update Pirate planetary defense missions

tehhowch:
- Fix incorrect references to opposing forces (e.g. "militia" used where it should be "Navy")
 - Add unpaid variants of the defense missions
 - Use stock conversations to hold the `on offer`
 - Add reference checking for stock conversations


endless-sky:
I would vote for putting the conversation definitions in the `pirate jobs.txt` file itself, either right before or right after the missions themselves. Otherwise, it becomes way too hard to track down where the actual conversation definition is.

(Very early on in the game's development I was always only putting one type of data in each file, e.g. putting mission events in a separate `events.txt` file and effects in a separate `effects.txt` file, both of which now feel like bad ideas except for events and effects that would be shared across multiple data files. The same would apply here.)

PR #3483, opened 2018-01-02
Use add attributes

tehhowch:
- Simplify definition of the Kestrel and its variants
 - Simplify definitions of Marauder variants
 - Simplify definition of the cloaking Archon
 - Simplify definition of the Deep River variants

With these changes, plus the Doombat change from earlier, proper usage of `add attributes` is conveyed to any content creators that prefer to learn from existing examples, and not just from reading the wiki (where `add attributes` should get an entry, in the Variants header: https://github.com/endless-sky/endless-sky/wiki/CreatingShips#variants).

### DIFF
--- a/data/drak.txt
+++ b/data/drak.txt
@@ -164,7 +164,7 @@ effect "antimatter dust"
 	"random spin" 50
 
 
-ship Archon
+ship "Archon"
 	noun "entity"
 	sprite ship/archon
 		"frame rate" 0.87
@@ -223,31 +223,9 @@ ship Archon
 	explode "nuke explosion" 3
 	"final explode" "final explosion large"
 
-ship Archon "Archon (Cloaked)"
+ship "Archon" "Archon (Cloaked)"
 	"never disabled"
-	attributes
-		category "Heavy Warship"
-		"cost" 1000000000
-		"shields" 4000000
-		"hull" 1000000
-		"required crew" 1
-		"bunks" 1
-		"mass" 1000
-		"drag" 27
-		"heat dissipation" 40
-		"fuel capacity" 800
-		"cargo space" 0
-		"outfit space" 800
-		"weapon capacity" 600
-		"engine capacity" 200
-		"energy capacity" 10000
-		"energy generation" 500
-		"shield generation" 100
-		"shield energy" 100
-		"hull repair rate" 100
-		"hull energy" 100
-		"heat generation" 17
-		"ramscoop" 100
+	add attributes
 		"cloak" .04
 	turret "Drak Anti-Missile Field"
 	turret "Drak Anti-Missile Field"
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -212,7 +212,6 @@ event "kestrel available: more shields"
 
 ship "Unknown Ship Type"
 	sprite ship/kestrel
-	plural "Kestrels"
 	attributes
 		category "Heavy Warship"
 		"cost" 10300000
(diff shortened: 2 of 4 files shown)


tehhowch:
A side-effect from this change (to the Unknown Ship Type / Kestrel layout) is that a mod which sells only the Kestrel will now sell the "Unknown Ship Type". AFAIK that is only the all-content plugin, which should really sell the better Kestrels and not the base Kestrel anyway - I feel this is a moot point.

Nescio0:
Correct me if I'm wrong, but won't deleting the base “Kestrel” cause errors with all Kestrel variants?

tehhowch:
Testing revealed no such issues - explosion weapons exist and function properly, and the variant for sale on Wayfarer has the proper attributes. Loading from the player's save has no issue either (I destroyed both directly-purchased-then-launched Kestrels, and purchased-then-relaunch-game-then-launch Kestrels - the former uses the directly instantiated variant, while the latter uses a ship read from the save file).

Nescio0:
Good, then I'm no longer concerned; thank you for your elaborate reply.

I still think making the "Unknown Ship Type" a variant of the "Kestrel", instead of the other way around, would be neater, but it probably is “a moot point”.

tehhowch:
>I still think making the "Unknown Ship Type" a variant of the "Kestrel", instead of the other way around, would be neater, but it probably is “a moot point”.

Making that change would result in targeting the test ship and seeing it as a "Kestrel" in the target reticle, rather than as an "Unknown Ship Type".

Nescio0:
Not necessarily, e.g.
```
ship "Kestrel" "Unknown Ship Type"

ship "Unknown Ship Type" "Tarazed Test Type"
```

The Kestrel, UST, and TTT would all be identical; the Kestrel and UST would show up as a "Kestrel", but the TTT would show up as an "Unknown Ship Type" (line 56 probably has to be edited as well).

jafdy:
Isn't that the same solution on more lines?

endless-sky:
It seems a little risky to base a variant on a variant. If the game were ever changed so that FinishLoading() is called on ships in some order other than alphabetical, it would no longer be guaranteed that the "Kestrel" variant has copied over the attributes of the "Unknown Ship Type" before being used as a base model for "Kestrel (Engines)" etc.

tehhowch:
That seems like an important detail that should be mentioned in the wiki, as it is entirely likely others will do what I have done here, especially as the use of `add attributes` becomes known.

While we don't exactly need to multithread the FinishLoading steps, that would be one quick way to have dependency issues between ship objects. Until we have some organizer to defer loading B until its base A is finished (and catch B needing A which needs C which needs B, etc.), I agree it's going to be much safer to not let variants reference variants as their base model.

endless-sky:
The other option would be to always have FinishLoading() call FinishLoading() on the base model before using it. That would probably require having FinishLoading() do nothing if it has already been called (just in case someone creates two variants that are each based on the other), but otherwise it would be straightforward.

But, the other reason to not let there be variants of variants is that in general, you want all the variants to share the same base model name, which means they all need to "inherit" directly from the base model. The Kestrel is a very unusual corner case.

tehhowch:
I can live with +77/-591 :grin: 
No changes needed to the all-content plugin now that the Kestrel is kept as a root level base model (unless the ACP should sell the upgraded Kestrels too).

endless-sky:
That does kind of drive home how absurdly high some of the Marauder engine capacities are...

endless-sky:
Also, that last merge officially makes you one of the few contributors who has officially removed more lines of code than you've added, so good job helping to simplify things. :)

KiLEdEnNis:
(more like good job deleting our precious games)

tehhowch:
I've been net negative before :) Adding missions and other non-code content makes it pretty difficult though, so this could be one of the last times 😇

Amazinite:
>I've been net negative before

Most of the time before recently, that is. :)

PR #3489, opened 2018-01-04
integrated Korath ramscoop

Nescio0:
Because all Kor Sestor and Kor Mereti ships have an integrated ramscoop (of 3), it would make sense the ordinary Korath ships (Raider and Worldship) would have an integrated ramscoop as well. This patch assigns them a value of 2 (double the ordinary Ramscoop, half the Wanderer Ramscoop). 
This would also remove the need for them of using Deep Sky's ordinary Ramscoop.

### DIFF
--- a/data/korath ships.txt
+++ b/data/korath ships.txt
@@ -25,6 +25,7 @@ ship "Korath Raider"
 		"outfit space" 721
 		"weapon capacity" 284
 		"engine capacity" 159
+		"ramscoop" 2
 		weapon
 			"blast radius" 250
 			"shield damage" 3600
@@ -39,7 +40,6 @@ ship "Korath Raider"
 		"Systems Core (Medium)"
 		"Large Heat Shunt" 2
 		"Small Heat Shunt"
-		"Ramscoop"
 		"Korath Repeater Rifle" 196
 		
 		"Thruster (Planetary Class)"
@@ -75,7 +75,6 @@ ship "Korath Raider" "Korath Raider (Hyperdrive)"
 		"Systems Core (Medium)"
 		"Large Heat Shunt" 2
 		"Small Heat Shunt"
-		"Ramscoop"
 		"Korath Repeater Rifle" 196
 		
 		"Thruster (Planetary Class)"
@@ -146,6 +145,7 @@ ship "Korath World-Ship"
 		"outfit space" 778
 		"weapon capacity" 314
 		"engine capacity" 165
+		"ramscoop" 2
 		weapon
 			"blast radius" 400
 			"shield damage" 8000
@@ -159,7 +159,6 @@ ship "Korath World-Ship"
 		"Triple Plasma Core"
 		"Systems Core (Large)"
 		"Large Heat Shunt" 2
-		"Ramscoop"
 		"Korath Repeater Rifle" 150
 		
 		"Thruster (Planetary Class)"
@@ -216,7 +215,6 @@ ship "Korath World-Ship" "Korath World-Ship A (Jump)"
 		"Triple Plasma Core"
 		"Systems Core (Large)"
 		"Large Heat Shunt" 2
-		"Ramscoop"
 		"Korath Repeater Rifle" 150
 		
 		"Thruster (Planetary Class)"
@@ -242,7 +240,6 @@ ship "Korath World-Ship" "Korath World-Ship B (Jump)"
 		"Triple Plasma Core"
 		"Systems Core (Large)"
 		"Large Heat Shunt" 2
-		"Ramscoop"
 		"Korath Repeater Rifle" 150
 		
 		"Thruster (Planetary Class)"
@@ -268,7 +265,6 @@ ship "Korath World-Ship" "Korath World-Ship C (Jump)"
 		"Triple Plasma Core"
 		"Systems Core (Large)"
 		"Large Heat Shunt" 2
-		"Ramscoop"
 		"Korath Repeater Rifle" 150
 		
 		"Thruster (Planetary Class)"


tehhowch:
I'd make it 1. The Raiders are a much older design than the automata, and unlike the automata, do not stray far into uninhabited space. There's always fuel to be had when plundering it from your enemies, after all.

data/korath ships.txt
```diff
@@ -25,6 +25,7 @@ ship "Korath Raider"
 		"outfit space" 721
```
tehhowch: You should also adjust the outfit space, perhaps by half (if you go down to 1 integrated scoop) or in full if you keep the new scoop at 2.

data/korath ships.txt
```diff
@@ -146,6 +145,7 @@ ship "Korath World-Ship"
 		"outfit space" 778
```
tehhowch: Same as the Raider.

KiLEdEnNis:
actually I'm fine with making their ramscoops 2, as they need it for survival, while the Automatons have it more as a "just in case"
or at least 1.5, I find 1 a bit low

Nescio0:
The suggested value of 2 is a compromise between, on the one hand, that ordinary Korath technology is possibly not as sophisticated as Mereti or Sestor (which isn't brand new either), and, on the other hand, that all Korath technology is typically far more powerful (though highly inefficient) than any human counterparts.
Besides, the Korath Raider's jump drives consume fuel twice as quick, so they really need ramscoop. It's default fuel capacity (600) is just enough for three jumps, yet somehow they manage to raid Remnant territory, which is much further away, and presumably intend to return as well.

Although I think 2 is not unreasonable, 1.5 could be a compromise. And you're right the outfit capacity could probably be lowered somewhat.

data/korath ships.txt
Nescio0: +2 outfits, -2 mass, to get 20 free space? (See Raider reply)

data/korath ships.txt
Nescio0: How about -6 outfits, +6 mass? This would result in free space going to 20/715, enough to fit two additional fuel pods or a hyperdrive alongside a jump drive (or vice versa).

Tadrix:
Great change.

I happen to think that ramscoop 2 for Raiders is a perfect spot — less than drones, more than humans. 

Korath are Tier 2 after all. Remnant are less developed and they have ramscoops 1,8 times better than human basic ones (per t). Even Catalytic Ramscoops are almost as good (1,7 per t). So ramscoop 1,5 (1,2 times better than basic human model) for Korath sounds too low to me.

Startome:
I agree either 1.5 or 2 sound like solid choices.

Also agree some tradeoff needs to be done in the name of balance, taking a few outfit space, a small fee, sounds entirely reasonable.

Good PR.

tehhowch:
One issue not yet discussed is that the removal of the Ramscoop outfit from `Korath Basics` means that a player who makes it into Efreti space will be unable to purchase a ramscoop from the Quarg-protected worlds.
This means that a player could obtain a Jump Drive by plundering a Raider, but no longer be able to source a ramscoop (as it would be unavailable to be plundered from the Raider, and also unable to be locally bought).

My suggestion is to then create a "Korath Ramscoop" outfit for the Raider and the World-Ship at minimum. I think it should be a bit more mass-efficient than the Emergency Ramscoop (as the Korath have had much more time than the Remnant to develop their technology), but considerably worse than the Wanderer Ramscoop (since the Korath aren't exactly known for efficiency). Alternately, it could be a worse than the Emergency Ramscoop (say, the same size and performance of the Deep's Ramscoop), but bring in some uniquely Korath traits - such as serving as a small source of cooling. If it has cooling or some other Korathy attribute, then it should probably not just be called "Korath Ramscoop" - maybe "Thermalizing Collector," etc. 

This outfit would then be available in "Korath Basics," allowing the player to both plunder a ramscoop outfit, and purchase one when on the edge of a large uninhabited region.

MCOfficer:
given how ramscoops (supposedly) work, using them as heat dissipator is a reasonable idea. although it kinda feels like something the wanderers would do, not the korath.

Nescio0:
> One issue not yet discussed is that the removal of the Ramscoop outfit from Korath Basics means that a player who makes it into Efreti space will be unable to purchase a ramscoop from the Quarg-protected worlds.
This means that a player could obtain a Jump Drive by plundering a Raider, but no longer be able to source a ramscoop (as it would be unavailable to be plundered from the Raider, and also unable to be locally bought).

I'm not sure there are actually any Raiders in Efreti space. Besides, you can only reach Korath space if you already have a jump drive. Furthermore, all ships over there have integrated ramscoops.

KiLEdEnNis:
@Nescio0 I'm sure Tehhowch knows that, I'm just not sure what exactly he meant by what you quoted

tehhowch:
>I'm not sure there are actually any Raiders in Efreti space. Besides, you can only reach Korath space if you already have a jump drive. Furthermore, all ships over there have integrated ramscoops.

There are no Raiders in Efreti space. There are, however, outfitters with `Korath Basics` as the sole source of outfits. This PR removes the ability of the player to obtain a Ramscoop from plundering Korath Raiders (where the player is also obtaining their Jump Drive) or Korath World-Ships.
This removal means that a player may begin exploring the large uninhabited regions of automata space without any kind Ramscoop outfit - or even knowing one exists, as they may not have reached the Deep.
image

My point is that this PR makes it much harder for a player to begin exploring the map, and much more likely for the player to run out of fuel in an uninhabited system with no way to speed up fuel collection beyond plundering hostile ships (a newly exploring player is not going to be capable of surviving an automata encounter long enough to disable and plunder them) or getting lucky and having a Quarg patrol enter their system *and stay long enough to be hailed*.

So, I suggest you replace the Ramscoop outfit with a Korath version, rather than switching to the integrated ramscoop (which i think is fine for the automata) and removing the Ramscoop from the `Korath Basics` sales list.

kikotheexile:
Im on board with the change, but feel the efreti need to have, for sale, their own version of a ramscoop. I also feel that the raiders should still have a chance to carry human and remnant scoops as variants for *both* gameplay and lore.

Nescio0:
Thanks for the explanation. I'm not sure how likely it is to happen, but it is a valid concern.

Brick63:
I can get working on a ramscoop if you guys want me to

PR #3491, opened 2018-01-04
Fix nan heat values

tehhowch:
The megacommit (24cab842a9f3a66c1b5fb862e1139707c9dd2ee4) that altered mass storage and ship regen introduced a bug that causes a ship without a given repair attribute to cause the ship to have nan heat values (and also incorrectly computed the `IdleHeat` and amount of heat to add for each repair function).

This PR fixes the nan heat bug by only updating the heat for a repair attribute if that repair attribute is nonzero.
The incorrect computations (wrong thermal mass in `IdleHeat`) and heat to add (doubly divided by the repair rate) are also fixed.
Some additional guards against attribute reads (e.g. `hull heat` is unneeded if `hull repair rate` is 0) are included.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1394,13 +1394,15 @@ void Ship::DoGeneration()
 		// 4. Shields of carried fighters
 		// 5. Transfer of excess energy and fuel to carried fighters.
 		
-		double hullAvailable = attributes.Get("hull repair rate");
+		const double hullAvailable = attributes.Get("hull repair rate");
+		// "per hull point" heat and energy costs of hull repair.
 		double hullEnergy = attributes.Get("hull energy") / hullAvailable;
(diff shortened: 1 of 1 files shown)


endless-sky:
`DoRepair()` already bails out if the amount remaining is zero, so most of those extra checks are unnecessary - 7b4443c alone is enough to fix this.

tehhowch:
You're still dividing by the available amount twice when adding the heat. hullHeat is (total hull heat)/(max hull repair), making it heat per point. Then when you add the heat for the repair given, you multiple hull heat by the amount done (good), but then divide again by (max hull repair).

I added the extra bits to minimize attribute.Get() calls that we can know don't need to be done, and the const bit quickly signals intent regarding the quantities.

PR #3493, opened 2018-01-04
Consider target's heat level when picking targets

tehhowch:
1. If a target is extremely overheated, there are probably other ships around that are not overheated that can attack the ship - those targets should be attacked more preferably than a ship which is incapacitated for a significant period of time.

2. ~~Adds a visual conveyance of the degree ( :grinning: ) to which the player has been overheated, by segmenting the HUD's `heat` bar.~~ Adds a second overlay bar with the ship's excess heat (the heat which needs to be dissipated before the ship is no longer overheated). This information is not currently available, so if the player is extremely overheated but otherwise not disabled, they can get visual feedback of how long they have to wait before their ship will begin responding again.

<details><summary>After provoking 20 Doombats:</summary>

This image is pre-overlay bar:
image
</details>

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -343,6 +343,11 @@ interface "hud"
 		dimensions 0 -192
 		color "heat"
 		size 2
+	bar "overheat"
+		from -13.5 403
+		dimensions 0 -192
+		color "medium"
+		size 2
 
 	# Targets.
 	anchor top left
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -984,6 +984,9 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 			range -= 1000 * Has(*it, gov, ShipEvent::BOARD);
 			// Focus on nearly dead ships.
 			range += 500. * (it->Shields() + it->Hull());
+			// If a target is extremely overheated, focus on ships that can attack back.
+			if(it->IsOverheated())
+				range += 3000. * (it->Heat() - .9);
 			if((isPotentialNemesis && !hasNemesis) || range < closest)
 			{
 				closest = range;
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -551,7 +551,7 @@ void Engine::Step(bool isActive)
 		info.SetBar("fuel", flagship->Fuel(),
 			flagship->Attributes().Get("fuel capacity") * .01);
 		info.SetBar("energy", flagship->Energy());
-		info.SetBar("heat", flagship->Heat());
+		info.SetBar("heat", min(1., flagship->Heat()));
 		info.SetBar("shields", flagship->Shields());
 		info.SetBar("hull", flagship->Hull(), 20.);
 	}
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -551,7 +551,16 @@ void Engine::Step(bool isActive)
 		info.SetBar("fuel", flagship->Fuel(),
 			flagship->Attributes().Get("fuel capacity") * .01);
 		info.SetBar("energy", flagship->Energy());
-		info.SetBar("heat", min(1., flagship->Heat()));
+		double heat = flagship->Heat();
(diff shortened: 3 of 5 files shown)


endless-sky:
If you're reordering the Energy(), Heat(), and Fuel() functions, it'd make more sense to put them in the same order as the UI and the member variables (fuel, energy, heat).

For overheating, segmenting the bar really doesn't make sense to me as a way to show cooling progress. It'd make more sense to have a second "overheat" bar of a different color overlaid on top of it, or something like that.

And, I actually find it a really useful strategy in some battles, when I have heat weapons, to focus fire on one powerful ship in order to keep it permanently overheated. It'd take some experimentation to figure out whether stopping an attack on an overheated ship (thus allowing it to recover much more quickly) to go after other targets is actually more effective than continuing to heat it up.

tehhowch:
This change wouldn't make your escorts or your flagship change targets if ordered to attack a certain ship. It strongly improves combat effectiveness of high heat dealing ships like the Doombat as currently they neutralize opponents quickly, but can't finish them off and instead get killed by other ships that they could have safely targeted.

The reorder was to group the status functions into "clamped" and "not clamped" .

We could define an overlapped bar graph in the HUD interface, but without segmenting it too I'm not sure how we can again convey magnitude unless we just reference the points of the heat graph and manually draw a custom color from red to white to wow-so-bright-white as the heat level grows. That color changing approach isn't conducive to supporting interface mods though.

endless-sky:
My idea was that when the first bar fills up a second overlay bar begins filling to show how much your heat has exceeded 100%. Or, it could show up only if you're overheated, and show far you have to go to get down to 90%, e.g.:

```
max(0, min(1., (heat - .9) * 2.5))
```

That is, if a ship overheated normally, the overheat overlay will fill a quarter of the bar, leaving the remaining 3/4 of the bar's height to be filled only if you're overheating due to damage. Or, if it turns out that damage often results in heat levels above 1.3 (which would be mapped to 100% by that expression) it could just use a multiplier of 1 instead.

overheat

tehhowch:
If getting hit by ships with flamethrowers (which might happen to the player in Syndicate / Republic campaigns), you can exceed 130% heat pretty easily (my screenshot shows I hit over 1100% heat, albeit using an extreme case).
You're correct in that exceeding 130% heat is rather difficult to do to yourself with vanilla outfits, but its only a matter of time until a mod uses a rather powerful weapon with a long reload and excessive heat output.

I do like the idea of a second bar so that the original heat level display remains constant. I think segmenting the overlay is a simple approach that makes it very clear what is going on (especially since it won't just appear segmented - the player will get to watch it segment (because they sure can't fly while overheated :grinning: )).

tehhowch:
I've added the overlay bar and made it tally the heat needed to be shed before the flagship is no longer overheated.
<details><summary>Really overheated:</summary>


image
</details>
<details><summary>When almost not overheated:</summary>


image
</details>

endless-sky:
I still don't think it makes sense to have the number of segments in the bar indicate the number of times the bar has to clear before you're no longer overheated. That's not something I've ever seen in any other game UI. And, in regular play the chance of getting above 190% heat is probably pretty slim, especially with your change to make ships de-prioritize overheated targets.

tehhowch:
You'd prefer it if the player just knew "Hey, I'm really overheated" and not "Hey, I'm really overheated, [*but at least I'm cooling down now* **|** *and it'd be nice if these ships would stop heating me up even more*]."?

With the deprioritization, it would indeed be pretty rare, but for that rare occasion, shouldn't we still show the player what is going on?

endless-sky:
My concern is that there's no precedent that I'm aware of in UI design for indicating a bar is filled multiple times by segmenting it. (And, that's not the meaning of the other segmented bars in the UI.) That, plus the fact that the bar should rarely get that high anyway, makes me think it would be better for it to just max out in those rare instances.

tehhowch:
The segments fuel demarcate blocks of 100 fuel - the segments in the overheat demarcate units of 100% heat units

PR #3494, opened 2018-01-04
Always show tactical information if targeting a player-owned ship

tehhowch:
Seems a bit odd that we can't visualize the realtime data from our player-owned ships just because they're out of range.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -660,7 +660,8 @@ void Engine::Step(bool isActive)
 			}
 			// Actual tactical information requires a scrutable
 			// target that is within the tactical scanner range.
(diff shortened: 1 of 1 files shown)


endless-sky:
The tactical scanner was added for players who want that sort of information, but I think others would find the information uninteresting or confusing. So, I agree with the idea of not putting a range limit on the tactical scanner if it's your ship, but I think it would be better to only show the information if a tactical scanner is installed, meaning the player actually wants to clutter up their HUD with the extra information.

tehhowch:
Addressed. I hadn't considered that the always-on "is yours" check would affect the early game, when the player likely doesn't know the Tactical Scanner exists (and could get an escort before discovering it), and would then wonder why some ships showed information and others did not.

PR #3502, opened 2018-01-05
Move ship and outfit categories to data files

Hadron1776:
This PR will allow modders to rename, add or remove categories of ships or outfits. 

It also allows the categories to be rearranged in any order (However, the "Fighter" and "Drone" categories will always be present: if they aren't defined, they will be added to the bottom of the list)


source/GameData.cpp
```diff
@@ -106,6 +106,10 @@ namespace {
 	map<const Sprite *, double> solarWind;
 	Set<News> news;
 	map<string, vector<string>> ratings;
+	vector<string> outfitCategories;
+	vector<string> shipCategories = {
+	"Fighter",
```
tehhowch: I'd either single-line this, or at minimum indent, i.e.
`vector<string> shipCategories = { "Fighter", "Drone" };`
or
```
vector<string> shipCategories = {
    "Fighter", "Drone"
};
```

source/GameData.cpp
```diff
@@ -946,6 +958,19 @@ void GameData::LoadFile(const string &path, bool debugMode)
 			for(const DataNode &child : node)
 				list.push_back(child.Token(0));
 		}
+		else if(key == "categories" && node.Size() >= 2)
+		{
+			if(node.Token(1) == "outfit" || node.Token(1) == "ship")
```
tehhowch: Bind a reference to the correct list, i.e.
`container<string> &categoryList = node.Token(1) == "outfit" ? outfitCategories : shipCategories;`

and then you don't need to duplicate your push_back call, and can still perform uniqueness checks on the list (if using vector, `find(token) == end()`, or something like unsorted_set, `! .count(token)`

source/GameData.cpp
```diff
@@ -946,6 +958,19 @@ void GameData::LoadFile(const string &path, bool debugMode)
 			for(const DataNode &child : node)
 				list.push_back(child.Token(0));
 		}
+		else if(key == "categories" && node.Size() >= 2)
```
tehhowch: Move the token value check up here, so that if an unsupported category is passed, a warning is generated:
`else if(key == "categories" && node.Size() >= 2 && (node.Token(1) == "outfit" || node.Token(1) == "ship"))`

source/Ship.cpp
```diff
@@ -75,18 +75,6 @@ namespace {
...
-	"Light Freighter",
-	"Heavy Freighter",
-	"Interceptor",
-	"Light Warship",
-	"Medium Warship",
-	"Heavy Warship",
-	"Fighter",
-	"Drone"
-};
-
```
tehhowch: One line between the end of namespace and the start of class methods is fine (as it was when categories were here).

source/GameData.cpp
```diff
@@ -760,6 +765,14 @@ const string &GameData::Rating(const string &type, int level)
 
 
 
+// Store categories for outfits and ships
```
tehhowch: `Get player-defined categories for outfits and ships.`

source/GameData.cpp
```diff
@@ -946,6 +959,15 @@ void GameData::LoadFile(const string &path, bool debugMode)
 			for(const DataNode &child : node)
 				list.push_back(child.Token(0));
 		}
+		else if(key == "categories" && node.Size() >= 2 && (node.Token(1) == "outfit" || node.Token(1) == "ship"))
+		{
+			vector<string> &categoryList = (node.Token(1) == "outfit" ? outfitCategories : shipCategories);
+			for(const DataNode &child : node)
+			{
```
tehhowch: These `{` and `}` aren't needed any more.

Amazinite:
>The "Fighter" and "Drone" categories are the exception to this, as modifying them may cause issues with carriers.

If these two categories can't be modified, then why even include them in the list of categories?
Perhaps instead there should be a different section that fighters and drones go under, like `categories "carried ships"` instead of being under `categories "ships"`, which would allow the modification of fighter and drone categories and the addition of something like [this](https://github.com/endless-sky/endless-sky/pull/2200#issuecomment-283503934).

tehhowch:
One would include them so they can be placed in the order of the player's choosing in the Shipyard.

Allowing changing the name of the categories for carried ships gets wonky because of how one specifies the bay type, and how the game checks it in flight.

It can all be changed, yes, but let this pull not make the same mistake as #2200 did.

Hadron1776:
>Allowing changing the name of the categories for carried ships gets wonky because of how one specifies the bay type, and how the game checks it in flight.

I definitely noticed that while I was attempting to make the change
(I eventually undid that and decided it would be best as a follow-up PR)

>Move the token value check up here, so that if an unsupported category is passed, a warning is generated:
else if(key == "categories" && node.Size() >= 2 && (node.Token(1) == "outfit" || node.Token(1) == "ship"))

True. However, a root element that says "we don't know what this is" doesn't make sense if the game actually _does_ recognize the element and is actually complaining about something else. So I changed the warning:
```
			else
				node.PrintTrace("Skipping unsupported use of categories object:");
```

data/categories.txt
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2016 by Michael Zahniser
```
tehhowch: img

source/GameData.cpp
```diff
@@ -106,7 +106,8 @@ namespace {
 	map<const Sprite *, double> solarWind;
 	Set<News> news;
 	map<string, vector<string>> ratings;
-
```
tehhowch: Add a space between shipCategories and StarField to get this deletion to go away.

source/GameData.cpp
```diff
@@ -946,6 +959,18 @@ void GameData::LoadFile(const string &path, bool debugMode)
...
+		{
+			if(node.Token(1) == "outfit" || node.Token(1) == "ship")
+			{
+				vector<string> &categoryList = (node.Token(1) == "outfit" ? outfitCategories : shipCategories);
+				for(const DataNode &child : node)
+					if(find(categoryList.begin(), categoryList.end(), child.Token(0)) == categoryList.end())
+						categoryList.push_back(child.Token(0));
+			}
+			else
+				node.PrintTrace("Skipping unsupported use of categories object:");
```
tehhowch: Personally I would quote the special token in the warning message, i.e.
`...use of \"categories\":"`

data/categories.txt
Hadron1776: #copypasta

source/Outfit.cpp
```diff
@@ -38,7 +39,14 @@ void Outfit::Load(const DataNode &node)
 	for(const DataNode &child : node)
 	{
 		if(child.Token(0) == "category" && child.Size() >= 2)
+		{
 			category = child.Token(1);
+			const vector<string> &outfitCategoryList = GameData::Categories("outfit");
+			const vector<string> &shipCategoryList = GameData::Categories("ship");
+			if(find(outfitCategoryList.begin(), outfitCategoryList.end(), category) == outfitCategoryList.end()
```
tehhowch: You shouldn't do this check here (in Outfit::Load) as the state of GameData::Categories() is unknown while GameData::BeginLoad is still operating.

In GameData::FinishLoading, after you insert the Fighter/Drone categories, you know your category lists are complete. Given that GameData::CheckReferences is the major "hey is anything wonky?" call, in that method you should extend the check of outfits and ships for proper categories

Hadron1776:
Some pictures of this PR:

proof1

proof2
(In this second one, I add a new category, "Licenses" to the bottom.)

proof3
(And yes, they expand properly)

Nescio0:
This is great, hopefully it'll be implemented sooner rather than later!

Tadrix:
How outfits with unmatching categories are handled? Are they displayed?

I propose for the game to automatically create these missing categories and place them in alphabetical order between listed categories and Fighter and Drone categories (if these are unlisted and applicable, otherwise appends to the end).

tehhowch:
>How outfits with unmatching categories are handled? Are they displayed?

as of https://github.com/endless-sky/endless-sky/pull/3502/commits/6e568e2b3e3fe2346bedc3af67fbc1e787983052 they get warned about. They're still in the game, but you won't be able to buy or compare them (just like if they had a blank category) because they won't get displayed.

Not auto-creating the category means that the creator of the plugin is about 500 times more likely to realize their oops (especially if they don't know about errors.txt or launch from a terminal that can print error messages from ES).

Tadrix:
It's not oops, he may decide that auto-creation result works just fine and not bother. Also, great for testing. And the final one — this would make using various plugins together easier, for example, if one plugin overwrites category list of other plugins.

tehhowch:
If you review the code, you'll note that the category list is inclusive and does not get overwritten.

mdsmestad:
This would be an awesome feature!

Nescio0:
Perhaps drones and fighters could have a “Drone” respectively “Fighter” attribute, which allows them to be docked in drone/fighter bays of other ships and prevents them from independent hyperspace travel? Cf. `automaton 1` or the Gascraft's `gaslining 1` attribute.

If it's done with an attribute instead of a class, it won't be necessary to make exceptions for drone and fighter classes.

jafdy:
That is the first reasonable sounding implementation I have heard. How hard would it actually be too add?

On the subject of backwards compatibility I feel that braking fighters to be able to add this feature would be worthwhile, although I'm not overly comfortable.

tehhowch:
I'd say we address custom carried categories elsewhere

Nescio0:
Any update on this? I'd really like to have this feature in some form or another.

tehhowch:
MZ will review this when he gets to it. Asking for updates doesn't speed the process up.

Nescio0:
Although I'm aware your time is limited, it would be really nice if something like this could make it into the game, allowing plug-ins to easily introduce or reorder categories.

data/categories.txt
```diff
@@ -0,0 +1,31 @@
...
+# Endless Sky is free software: you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later version.
+#
+# Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+categories "ship"
+	"Interceptor"
```
KiLEdEnNis: this line should be below the Heavy Freighter line (3 lines below this)

source/LocationFilter.cpp
```diff
@@ -470,7 +470,7 @@ void LocationFilter::LoadChild(const DataNode &child)
 	else if(key == "category" && child.Size() >= 2 + isNot)
 	{
 		// Ship categories cannot be combined in an "and" condition.
-		static const set<string> allowed(Ship::CATEGORIES.begin(), Ship::CATEGORIES.end());
+		static const set<string> allowed(GameData::Categories("ship").begin(), GameData::Categories("ship").end());
```
tehhowch: I do have a concern here that file load order may cause issues. The first specification of a ship category (e.g. from a job or mission in the base game or earlier-loaded plugin) as a condition in a `LocationFilter` will result in the creation and finalization of this variable `allowed`, and it will persist while the game runs. Any further additions to the category list will not be reflected in the following code block.

I.e.
```
Base -> categories.txt
Base -> jobs.txt
...
Plugin 1 -> so many missions.txt
Plugin 2 -> boarding jobs.txt
Plugin 2 -> ship categories.txt
```

Any categories in `ship categories.txt` are not even known when `boarding jobs.txt` is parsed. And even if they were known, they would still not be in `allowed` if any preceding file reached this line.

Amazinite: Confirmed that this is a bug.

Amazinite:
Closing, as this PR has been forked and merged via #5971.

PR #3504, opened 2018-01-06
More faces and more opinions

Pointedstick:
This PR adds a number of new characters and 89 new faces. Each face is only used for a single character (no more seeing the same person be a Merchant Captain, Farmer, and Financial Analyst). I focused on adding more characters and more faces, but will add more of their opinions later.

I'm still not totally sold on this new system, but this is my best shot for now. In order to make it work, we're going to need **vastly** more images to avoid seeming awkward, to the point that I question whether it's feasible or even possible. It's pretty slim pickings on unsplash.com, unless we want the Endless Sky galaxy to be mostly populated by hat-wearing bearded hipsters and sexy young women in huge sunglasses.


data/news.txt
```diff
@@ -8,53 +8,158 @@
...
+		portrait/human58
+		portrait/human63
+		portrait/human72
+		portrait/human73
+		portrait/human76
+		portrait/human79
+		portrait/human98
+	message
+		word
+			"Can you believe how little they pay for used ships and outfits these days? It's just a scandal!"
```
Fzzr: Ha ha ha =P

jafdy:
I will need to look through this properly at some point. 101 looks a bit odd standing there with their mouth wide open.

Nescio0:
They all look like early 21st C faces to me; this game is situated in the early 31st C. Although we do not know how people will look like in thousand years, my guess is they will not look the same as now. After all, 21st C people are quite different from e.g. 11th C people as well.

Pointedstick:
> They all look like early 21st C faces to me; this game is situated in the early 31st C. Although we do not know how people will look like in thousand years, my guess is they will not look the same as now. After all, 21st C people are quite different from e.g. 11th C people as well.

That's kind of the point I'm making: this is going to take a lot of art assets to work--and they're all going to have to be of substantially higher quality than what we can get from unsplash.com.

After doing all of this, I have to say I'd be in favor of scrapping it and moving to a purely text-only interface, perhaps displaying news headlines rather than quips from fictional characters.

Amazinite:
>After doing all of this, I have to say I'd be in favor of scrapping it and moving to a purely text-only interface, perhaps displaying news headlines rather than quips from fictional characters.

👌 I was iffy on the visuals of this from the beginning. It just seems like such a change of pace from the rest of the game. I prefer the real images to the cartoony one that MZ showed at first (the real images make the game feel more real, in a way), but it's still seems out of place to have faces in game. (That includes the few instances during the FW campaign of the real life scenes being shown during conversations.) That might just be because of the fact that faces are seen so sparsely already, though. I can see the current news system having a place in the game, but it just feels to me that that would open questions of why we don't get to see the faces of the characters we're spending a majority of the time talking to through missions.

Nescio0:
There is one NPC with a face: that Kraz engineer who designed the plasma turret (if I recall correctly). But I never liked that image either and I hope it will be removed (or replaced) as well.

Amazinite:
>There is one NPC with a face: that Kraz engineer who designed the plasma turret (if I recall correctly). But I never liked that image either and I hope it will be removed (or replaced) as well.

Which I mentioned

>it's still seems out of place to have faces in game. (That includes the few instances during the FW campaign of the real life scenes being shown during conversations.)

Nescio0:
Yeah, you did :) They are located at `images/scene/loc*.jpg`

UnciaPrima:
> it's still seems out of place to have faces in game. (That includes the few instances during the FW campaign of the real life scenes being shown during conversations.)

This. I have the exact same feeling. I think there's a lot left to the imagination when reading the story (and that's definitely a good thing), and every time an image popped up, it was distracting, almost as if it was "forcing" me to imagine things in a certain way.

KiLEdEnNis:
> After doing all of this, I have to say I'd be in favor of scrapping it and moving to a purely text-only interface, perhaps displaying news headlines rather than quips from fictional characters.

I still don't like the actual face idea at all & don't think I will, so I'm for that

Brick63:
yeah adding faces may not be the best direction as it leaves less to the imagination

Nescio0:
Maybe all those unnecessary images (both the new portraits and those existing campaign scenes) could be moved to an optional plug-in for those who like them, but be removed completely from the default game? I'd like to see them go; every time I encounter such an image I felt a pang of disappointment.

Tadrix:
Then you'd need two separate plugins, because there are people who don't like faces who don't mind an image or two in story events. I'd say let's just leave things how they are now in this regard.

Image-less news announcement system would still be nice.

Pointedstick:
> every time I encounter such an image I felt a pang of disappointment.

I'm afraid I have to agree. I really tried to improve the experience by adding so many new faces, and separating them out into different characters so that most of them look the part, but IMHO even after that, the net effect is only so-so at best. It just doesn't shine, and probably won't without an enormous amount of the kind of professional artistic work that frankly we aren't likely to get anytime soon (though hope springs eternal...).

> Image-less news announcement system would still be nice.

+1

Brick63:
when this was first being discussed i thought the news would just be little things like you would see in an actual newspaper not something said by a specific person

KiLEdEnNis:
Amazinite made a voting poll for everyone on the Discord, as of now, 17 people are for the faces, 28 people are against it, so this already shows that a majority is probably against this feature in general

and my arguments against it were
- it'd make it harder to create alien species, thus raising the content creation barrier
- imagining what aliens look like is way more fun than having a set picture
- if ya only make it for humans it'd look like a clunky feature, as if the aliens are yet to be added

Fzzr:
Two cents: I wouldn't mind the portraits so much if they didn't look like 21st century humans. Even so, I'd prefer no portraits at all, mostly for the reasons in the first two bullets @KiLEdEnNis gave above.

As for the news itself, I could go either way on the "things heard from people" or "news announcements" type of content. Without the portraits it could even be both.

sivael:
I would not do it for random portaits.
Random people should be represented by their faction logo/flag. This way it's a random merchant or random carrier that talks to You.
I would include portaits for the main storyline characters only.

endless-sky:
I feel like the face is necessary to mark the "news" text as something distinct from the spaceport description. Otherwise there are just two separate blocks of text in two different places on the screen, which is visually uninteresting and will just look cluttered. (Another option would be to do away with spaceport descriptions entirely and just put "news" in their place, but I'd rather not do that.)

Plus, I want the spaceport news to parallel the hailing function while in flight - i.e. an optional flavor thing where you can keep hailing a ship just to see what sort of things they will say to you. So, in the same way as the hail panel showing the ship, I think having a picture of the person speaking is helpful.

endless-sky:
Also, I was shooting for as much consistency as possible in the size of the faces, in particular:
* Having the face take up half the width of the image (or slightly less, if there's a big hat or hair to fit in).
* Avoiding having anything beyond a person's head and shoulders in the image (e.g. no hands).
* Avoiding obviously 21st-century accouterments, like headphones or brand names.

Do you have high-enough resolution versions of these that you'd be able to change the crop size to be more consistent?

Pointedstick:
Not really. The ones with large heads were because that was the best I could do given the source material. Again, unsplash.com yields rather slim pickings. I can remove any faces that you don't like from the PR.

The inherent problem here is that our human brains have been evolutionarily calibrated to respond to human faces. Without literally hundreds of faces, the repetition will seem jarring. And making the faces look adequately "31st-century" is going to be flat-out impossible if we're limiting ourselves to public domain photographs.

That's to say nothing about aliens, which will have to be painted or 3D modeled.


> Another option would be to do away with spaceport descriptions entirely and just put "news" in their place, but I'd rather not do that

FWIW, I would strongly support this. Instead of faces, for visual flair we could have logos for different news companies. We'd only need a few, these would be hugely easier to create from scratch than it is to find photographs of human faces that fit in to the game's tole and mood.

endless-sky:
Do you have GIMP sources for these?

jafdy:
You would need to have somewhere over 300 different of faces to avoid the repetition being really obvious but random chance would still make it jarring not infrequently.

Pointedstick:
No, I didn't keep the GIMP files after processing, since these were pretty basic cutout-and-resize operations.

Brick63:
the easiest option which is also the best for keeping the bar low for new content creators is to use the faction icons for random people and having the news be shown with the logo of the news network which would mean not having to find or make the portraits for a new species

KiLEdEnNis:
I don't mind Brick's idea at all as I'm against the faces a lot
creating a couple logos isn't that hard for modders anyway, one could even use Blender for the shape, and use GIMP afterwards

Tadrix:
> Another option would be to do away with spaceport descriptions entirely and just put "news" in their place, but I'd rather not do that.

I'd suggest appending news under spaceport descriptions (kind of like sometimes there's new text saying that there's a new Navy base now). Just make a new paragraph about today's news.

- Spaceport descriptions usually are shorter than planet descriptions, so there is enough place for a short newscast (which should be brief anyway);
- Players would have an additional incentive to click on the "Spaceport" button;
- Clicking "Spaceport" suggests that protagonist put in time to mingle with people in the spaceport and actually hear news.

Lineth:
For portraits tbh I would just do a naev and use basic, generic drawn portraits for most civvies and more distinct portraits for special characters (if that ever becomes relevant). CGI faces are too uncanny-valley (though they don't bother me too much) and the real pictures are rather tacky and ill-fitting imo.

endless-sky:
I think this could easily turn into a case of "the perfect is the enemy of the good." Recruiting someone to do hand-drawn portrait artwork would be difficult, unless they're being paid, and doing cutouts from photos is easy. As with the landscape images (many of which were *really* tacky at first) the important thing is to get something usable into the game as a placeholder, and refine it later.

PR #3514, opened 2018-01-07
"more easy" -> "easier" in "Transport Workers A"

Hadron1776:
Using "more easy" in a sentence doesn't seem right.


Amazinite:
That's right both ways. https://github.com/endless-sky/endless-sky/pull/3289#issuecomment-347363770
>My general policy is, if it's not actually a grammar error and is just a matter of opinion, it's not worth tweaking it.

PR #3526, opened 2018-01-10
Modified Blaster description correction

solardawning:
Correcting now-inaccurate text which mentioned 'increased range'.
Changed phrasing about energy requirements to match phrasing in the Modified Blaster Turret, because I feel that the terminology 'significant increase' is no longer accurate, as the blaster rebalance has made the increased energy consumption of the Modified Blaster rather minor compared to its prior difference from the Blaster.
No changes needed to the Modified Blaster Turret text, as it didn't mention range.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -111,7 +111,7 @@ outfit "Modified Blaster"
 		"firing heat" 42
 		"shield damage" 12
 		"hull damage" 8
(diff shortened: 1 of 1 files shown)


solardawning:
Yeah, I'm just learning how to actually use GitHub. Apologies for any mistakes I make along the way- tonight involved a lot of Googling as-is, heh.

PR #3540, opened 2018-01-13
"war begins" removes Navy shipyard only

Nescio0:
The `"war begins"` event currently removes all outfitters and shipyards from Geminus. However, the description explicitly states only the *Navy* shipyard was targeted and destroyed. This proposal tweaks `"war begins"` to remove the Navy B&A shipyards only; `"geminus rebuilt"` is also changed accordingly.

### DIFF
--- a/data/events.txt
+++ b/data/events.txt
@@ -121,8 +121,8 @@ event "war begins"
 		description `Until recently, Geminus was home to the Republic Navy Yard, the largest shipyard in human space. Now, there is nothing left of the shipyard but a gaping crater in the ground, from the recent terrorist bombing.`
 		spaceport `Fortunately, the spaceport and residential areas survived the bombing, which was focused only on the Navy yard and happened early in the morning, when very few people were at work.`
 		spaceport `The Navy officers who are in charge here have no time for anyone who is not enlisted in the Navy and participating in the ongoing investigation.`
-		shipyard clear
-		outfitter clear
+		remove shipyard "Navy Advanced"
+		remove shipyard "Navy Basics"
 		"required reputation" 100
 	planet "Martini"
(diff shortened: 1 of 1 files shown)


Nescio0:
> This event was definitely added well in advance of the `remove` and `add` keyword support 😁

Yeah, I guessed so; it won't hurt to update the event though now those mechanics are supported. 
A side-effect is that if someone would want to add or remove any outfitter or shipyard to Geminus, he would currently have to edit both the `planet Geminus` entry in the `map.txt` file and the `event "geminus rebuilt"` entry in the `events.txt` file; the latter can be easily overlooked. With this change one can ignore the event and only edit the map entry.

MessyMix:
I agree with teh; the original event is a lot more impactful and
noticeable, which is fitting of an event that sparks a galactic war, imo.

endless-sky:
The "Navy shipyard" *is* the shipyard and outfitter on Geminus. The whole facility is gone, not just the manufacturing portions of it.

tehhowch:
On previous occasions it has been expressed that the "outfitter" and "shipyard" are not individual shops / locations, but rather represent everything available for sale on the planet. From this perspective, removal of only the Navy's catalogs makes sense.

MessyMix:
Then again, the bombing seems a lot less inconsequential if only a portion of the ships for offer at Geminus are removed. I personally would go with impact over realism in this instance.

Amazinite:
>On previous occasions it has been expressed that the "outfitter" and "shipyard" are not individual shops / locations, but rather represent everything available for sale on the planet. From this perspective, removal of only the Navy's catalogs makes sense.

And where everything is located may change from planet to planet. On Trinket, you might be able to get some Nerve Gas from a back alley, but on Geminus, everything you can buy might be located in a single district, the exact district that got nuked.

I will note that although this event was originally created before the ability to remove individual outfitters and shipyards, not only are all shipyards removed but also all outfitters, which already suggests that Geminus has all its shipyards and outfitters in one place.

PR #3542, opened 2018-01-13
Corrected X5200 description

Nescio0:
The current description of the "X5200 Ion Steering" describes it as “standard equipment for Republic cruisers and carriers”. However, none of the Republic ships nor any of their variants actually use a X5200, therefore the description is clearly wrong. This patch rewrites the text to something more generic.

### DIFF
--- a/data/engines.txt
+++ b/data/engines.txt
@@ -223,7 +223,7 @@ outfit "X5200 Ion Steering"
 	"turn" 2174
 	"turning energy" 3.5
 	"turning heat" 7.3
-	description ""The X5200 is a steering system so large and powerful that only a handful of ships can install it. This means that this steering system is rarely sold due to its limited market."
+	description "The X5200 is a steering system so large and powerful that only a handful of ships can install it. This means that this steering system is rarely sold due to its limited market."
 	description "	Ion engines consume less energy than plasma engines and produce less heat, but they also take up more space."
 
 
--- a/data/engines.txt
+++ b/data/engines.txt
@@ -223,7 +223,7 @@ outfit "X5200 Ion Steering"
 	"turn" 2174
 	"turning energy" 3.5
 	"turning heat" 7.3
-	description "The X5200 is a steering system so powerful that only a handful of ships use it. It is standard equipment for Republic cruisers and carriers, ships so large that they are practically a self-contained city."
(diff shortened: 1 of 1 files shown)


jafdy:
You're right. I just don't like it much though. It sounds off from useful onward.

Nescio0:
Feel free to make improvement suggestions or counter-proposals.

warp-core:
Make it: `useful for capital ships so large they are practically self-contained cities.`?

Fzzr:
The only ships that use the X5200 by default are the Kestrel and some Bactrian models. I don't know if the Bactrian is actually a "capital ship" by design, regardless of how players may use them, and the Kestrel isn't available right away. How about "standard equipment for the Bactrian City-Ship, the single largest ship built today."

jafdy:
We probably should avoid referencing a ship as we might end up doing this again down the line.

Fzzr:
Fair. How about just remove "capital" then?

data/engines.txt
```diff
@@ -223,7 +223,7 @@ outfit "X5200 Ion Steering"
 	"turn" 2174
 	"turning energy" 3.5
 	"turning heat" 7.3
-	description "The X5200 is a steering system so powerful that only a handful of ships use it. It is standard equipment for Republic cruisers and carriers, ships so large that they are practically a self-contained city."
+	description "The X5200 is a steering system so powerful and large that only a handful of ships can install it. It is useful for capital ships so large they are practically self-contained cities.
```
KiLEdEnNis: 1st ya forgot the " at the end

2nd, I did an attempt to improve the description: "The X5200 is a steering system so large and powerful that only a handful of ships can install it. This means that this steering system is rarely sold due to its limited market."

I entirely rewrote the second sentence since both sentences basically talked about how only giant ships can use one

data/engines.txt
```diff
@@ -223,7 +223,7 @@ outfit "X5200 Ion Steering"
 	"turn" 2174
 	"turning energy" 3.5
 	"turning heat" 7.3
-	description "The X5200 is a steering system so powerful and large that only a handful of ships can install it. It is useful for capital ships so large they are practically self-contained cities.
+	description ""The X5200 is a steering system so large and powerful that only a handful of ships can install it. This means that this steering system is rarely sold due to its limited market."
```
KiLEdEnNis: there's 2 " at the start now, remove 1

Issue #3544, opened 2018-01-14
shortcut key "i" for install will buy outfits too


whismerhill:
seems like an issue to me or at least not desirable
ALT+u will uninstall up to 500 outfits & put them in cargo just fine
ALT+i will first install outfits up until there is none at which point it'll buy the rest

doesn't sound desirable to me when you can use "b" or the button to buy instead for the exact same effect
example usage : 
-get into an outfitter that sells rifles
-transfer 150 rifles from previous capital ship to new capital ship
-using ALT+u & then ALT+i will put 500 rifles on your new capital ship, 150 of which are old ones.

tehhowch:
`i` only works if you don't have to buy the outfit (e.g. it is in cargo), while b always works (and just pulls from cargo first). There could be an issue where the loop called by `i` doesn't stop once you are out of cargo outfits - you are suggesting it not ever buy outfits, i.e. stop at 150/500 requested?

whismerhill:
yes exactly. since as you said pressing "i" will not buy outfits if none are in cargo

tehhowch:
Addressed in #3553

tehhowch:
Don't close this until it's been merged and milestoned.

tehhowch:
Resolved by efd11f1e37f6168ca43e5cd7b504e7bc68fc906f

PR #3553, opened 2018-01-15
Installing multiple only installs those that come from cargo

tehhowch:
Refs #3544 

If using the shortcut key `i` to install from cargo in conjunction with a modifier key to "move things along," only the number present in cargo will be added to ships (rather than buying the remaining), in accordance with the use of `install` rather than `buy`


endless-sky:
"Passing" a value to a function by setting a member variable doesn't seem like a good idea. I'd rather fix this by adding a function parameter (efd11f1) even though that parameter only applies to the Outfitter. (Plus, that way it parallels the "toCargo" parameter in the Sell functions.)

Issue #3522, opened 2018-01-09
Show discounted outfits checkbox


whismerhill:
pretty much the title
would go along with those other checkboxes  ("show outfits for sale"...)

Alternatively highlight item borders that have used outfits or something

edit : or both => a checkbox that highlights instead of filtering

the reason is that : 
 - currently if there's outfits I wanna keep from ships I will sell, I need to first strip them of outfits & put them on my ships in cargo & then sell the ships. 
 - if such a thing as showing discounted outfits existed, I could just directly sell a bunch of ships then check out whatever is discounted at a glance without having to check for EACH & every item sold by the outfitter. It would just be much faster.

there's plenty other example usage scenarios where you wanna keep discounted outfits and this could get handy.

tehhowch:
Is this not already possible? Simply uncheck the "show for sale" box.

whismerhill:
nope with "show outfits for sale" unchecked, it will show items on the selected ship & that's all

"show used outfits" would either filter out or highlight items which are not 100% price (that you sold)

Hadron1776:
(You can select multiple ships)

tehhowch:
Note that "used" has multiple meanings. I interpreted it as those outfits that are being used i.e. are installed on a ship.
It appears @whismerhill actually means "used" as "sold for less than full price" which is not its definition.

A better post title is "Button to show discounted outfits"

whismerhill:
hooooo facepalm

Brick63:
For what reason should this be implemented

ghoulavenger:
@Brick63 why not? When you sell ships or components to the outfitter, you have discounted outfits available for purchase that do not appear that way, they just show up as normal outfits (although it might show that it is discounted in its price section). It simplifies grabbing loose outfits from captured ships significantly (selling ships will strip the outfits and put them for sale in the outfitter). Not to mention if you're just swapping outfits around it could make it easier to just see the outfits you're looking for.

I don't see it as a high priority but I don't see a reason for it not to be implemented either.

Brick63:
okay my question was more inquisitive than anything and i was looking more for whismers reasoning but not to discount others

whismerhill:
the reason is that : 
 - currently if there's outfits I wanna keep from ships I will sell, I need to first strip them of outfits & put them on my ships in cargo & then sell the ships. 
[...]

PR #3556, opened 2018-01-15
Add outfitter checkbox to toggle showing only outfits with depreciation

tehhowch:
Refs #3522 

If checked, when buying only those that are below full price will be bought (i.e. you can buy the 4 that were sold at 25% by using the Shift modifier and get only those 4, rather than 4 and 1 full-price outfit).
 
<details><summary>2 cargo outfits, one is depreciated</summary>

Unchecked (current behavior):
image

Checked (only shows outfits purchasable at less-than-full price):
 image
</details>

High-DPI: https://github.com/endless-sky/endless-sky-high-dpi/pull/43
Assets: https://github.com/tehhowch/modding-assets/tree/master/outfitter%20key


endless-sky:
It seems to me that this would be a feature with very limited usefulness. A player who is selling a bunch of captured ships but wants to save a couple of outfits from them probably already has certain outfits in mind.

An alternative might be to add some sort of "25% off" label to used outfits so they're easy to spot. But again, the important question is: will this feature provide enough added benefit to the *average* player to compensate for the increased learning curve for *every* player of being faced with one more UI option that you don't understand when you first start playing the game?

tehhowch:
- I agree that an overlay would reduce the confusion over "what's this box actually do?", especially since the store often does not have depreciated outfits - they are only there if the player has sold them in some manner.
- The overlay also would make it unnecessary to select each item to determine if they are full or reduced price (as is currently necessary)
- The overlay would not allow limiting the purchased amount to that which is below full value, since there wouldn't be a way to flag that this is something that the player is interested in.

Issue #2852, opened 2017-07-29
0.9.7 Balance Issues: Are these as intended?


marphod:
Playing through the new content, I've noticed a few balance issues along the way.

- Wanderer Transport Missions (& other missions involving transport) seem to have a ridiculously high payout.  

Not that it is necessary excessive per sentient transported, but they are really large groups over a very short ranges, and often you will be offered multiple missions on the same path (since there are only so many options in the territory that small).  Comparable Bulk Cargo missions would need to be about an order of magnitude higher (300-2000 units rather than 30-200) to have similar payouts, but that just doesn't happen.  (For instance, I've got 4 cargo missions to the same planet that total approx 275 tonnes of goods and pays approx 230k credits; which is way better than average.  I've got got 4 transport missions for 185 wanderers to an equally distant system for 1.2m credits; and that's on the low side of average.)  I've even seen transport missions with a 5m+ payout.

This may be a side order of 'Transport missions scale better with character power than cargo missions'.

- The Remnant Crystal Capacitor seems very powerful

And/or the Coalition outfits are terrible. 

The Crystal Capacitor has a capacity of 283 units of energy per ton, and is under 20 tons in size.  The only other small footprint battery with similar capacity is the Large Battery Module at 250 units of energy per ton -- which was one of the only  useful items the coalition had.  (The only other batteries that compare are the Quarg Nanotech, which is Tier 4-ish, and possibly the Korath Large System Core and/or the Wanderer Sun generators; and those are much larger footprint at at least 50 tonnes and that depends on how you calculate the value of the multi-function outfits).

- Outfit Expansion Nerf may not be enough of a Nerf -- 

It doesn't hurt large-battery/small-generator builds

There are issues with that style of ship, but a Bactrian with a Large Biochem or Red Sun, 2-3 System Cores (Large), some batteries (Nanotech are best, but Crystal Capacitors or Large Battery Modules will do) and a defensive flagship load-out is a working build , doesn't care about the lack of cooling, and is fine in all but the longest firefight.  Throw a cloaking device in with enough ramscoops and it is even better.  You can do the same with most Missile-boat builds.

(I may find more)

Bladewood:
Wanderer missions are intentional.

The Crystal Capacitor is much more space efficient than the Large Battery Module, which is itself much more efficient than pretty much every other dedicated battery outfit in the game (ignoring nanotech batteries). I'm certain this is intentional.

[...]

PR #3557, opened 2018-01-15
Introduced Korath Fuel Station (fuel pod with ramscoop)

Nescio0:
In #3489 it was suggested to create an unique Korath ramscoop. Rather than combining it with some cooling or energy capacity, as suggested, I think it would be more interesting to have a combined ramscoop with fuel capacity, or actually a fuel pod with ramscoop.
<s> This proposal introduces a Korath “Fuel Collector” which is exactly as large and as effective as two fuel pods plus one emergency ramscoop:
```
                                            mass, fuel, ramscoop ; efficiency
updated Korath Fuel Collector proposal    :  20 , 200 , 2.0      ; 2.0/(20-16)=0.5
two fuel pods plus one Remnant ramscoop   :  20 , 200 , 0.5      ; 2.0/(20-16)=0.125
two fuel pods plus one ordinary ramscoop  :  26 , 200 , 1.0      ; 2.0/(26-16)=0.1
two fuel pods plus one Wanderer ramscoop  :  23 , 200 , 4.0      ; 2.0/(23-16)=0.57
two fuel pods plus one Catalytic ramscoop :  32 , 200 , 7.0      ; 2.0/(32-16)=0.4375
```

The Fuel Collector is available for sale at the Kor Efreti outfitters and replaces the human ramscoop and fuel pods used by the Korath. The Raider has two equipped, the World-Ship four. Both also have one integrated (because of hypothetical players who might plunder outfits before they capture the ship).
Their mass and outfit size are changed accordingly, their effective fuel capacity remains unchanged, and their total ramscoop increases from 1.0 to 6.0 (Raider) and from 1.0 to 10.0 (World-Ship).</s>

[EDIT] New proposal (masses are based upon universal fuel pods and Korath heat shunts):

* Fuel Station (Large): 0.69 m credits, 30 mass, 400 fuel capacity, 2.0 ramscoop
* Fuel Station (Small): 0.17 m credits, 9 mass, 100 fuel capacity, 0.5 ramscoop
* Korath Raider has internal 200 fuel and 1 ramscoop, plus one large fuel station equipped, replacing the ordinary ramscoop, bringing it to 600 fuel and 3 ramscoop (currently 600 and 1)
* Korath World-Ship has internal 200 fuel and 1 ramscoop, plus two large fuel stations equipped, replacing the ordinary ramscoop, bringing it to 1000 fuel and 5 ramscoop (currently 1000 and 1)
* Efreti outfitter sells small fuel station, replaces human fuel pod and ordinary ramscoop


data/korath outfits.txt
```diff
@@ -91,6 +91,15 @@ outfit "Large Heat Shunt"
...
 
+outfit "Fuel Collector"
+	category "Systems"
+	cost "101000"
+	thumbnail "outfit/unknown"
+	"mass" 20
+	"outfit space" -20
+	"fuel capacity" 200
+	"ramscoop" 0.5
+	description "The Korath Fuel Collector is a large system combining a weak ramscoop with enough fuel capacity for one jump drive jump or two hyperdrive jumps."
```
tehhowch: Description should not reference the amount of fuel needed for various jump modes - those value can be changed in mods.

I suggest developing the description from Korath lore, referencing their habitual tendency to plunder civilizations both near and far, etc.

data/korath.txt
```diff
@@ -576,8 +576,7 @@ outfitter "Korath Basics"
 	"Command Center"
 	"Small Bunk Room"
 	"Bunk Room"
-	"Fuel Pod"
```
tehhowch: I would not remove the Fuel Pod.

Brick63:
as i offered before i can do the sprite if you would like me to @Nescio0

data/korath ships.txt
```diff
@@ -17,12 +17,11 @@ ship "Korath Raider"
 		"hull" 9000
 		"required crew" 145
 		"bunks" 250
-		"mass" 720
+		"mass" 650
 		"drag" 12
 		"heat dissipation" .5
-		"fuel capacity" 600
```
tehhowch: The ships should have some base fuel capacity to support at least a single jump with their default drive.

Nescio0:
> as i offered before i can do the sprite if you would like me to @Nescio0

Yes, please! Make it large and Korath-like :)

data/korath.txt
Nescio0: Why not? This new system is the Korath fuel pod.

data/korath ships.txt
Nescio0: Each Fuel Collector allows a single jump drive jump or two hyperdrive jumps, and each Raider and World-Ship variant is equipped with several of them.

data/korath outfits.txt
Nescio0: Is the new description better?

data/korath.txt
tehhowch: I would consider the pod as the "small" version of the collector, used by Korath captains who just need a touch more fuel for their purposes (and don't want to wait to collect it).

data/korath outfits.txt
```diff
@@ -99,7 +99,7 @@ outfit "Fuel Collector"
 	"outfit space" -20
 	"fuel capacity" 200
 	"ramscoop" 0.5
-	description "The Korath Fuel Collector is a large system combining a weak ramscoop with enough fuel capacity for one jump drive jump or two hyperdrive jumps."
+	description "Because of their habit to raid systems both far and near, Korath ships need a lot of fuel for their jump drives. The Fuel Collector is a bulky system which has a ramscoop to collect new fuel and a large capacity to store it."
```
Bladewood: `Because of their habit of raiding systems near and far, Korath ships must carry a considerable amount of hyperspace fuel for their jump drives. The Fuel Collector is a bulky system that has both a ramscoop to harvest additional fuel, and a large capacity to store it.`

Also, "Fuel Collector" is a boring name.

data/korath ships.txt
tehhowch: And if the player plunders that outfit and then captures the ship, it is unable to jump *at all*
Note that the player will likely be unaware of the intricacies of the outfit prior to obtaining one. The name "Fuel Collector" does not imply that it also stores fuel, after all.

data/korath.txt
Nescio0: Perhaps we should have two versions then:
* Small Fuel Pod: 10 mass, 100 capacity, .2 ramscoop; for sale
* Large Fuel Pod: 20 mass, 200 capacity, .5 ramscoop; installed in Raiders and World-Ships, not for sale

jafdy:
It's going to be outfit sized, although it will be Korath like.

data/korath outfits.txt
```diff
@@ -91,6 +91,15 @@ outfit "Large Heat Shunt"
 	"cooling" 54
 	description "Without these powerful cooling systems, most Korath warships would overheat within seconds as soon as they began to fire their weapons."
 
+outfit "Fuel Collector"
+	category "Systems"
+	cost "101000"
```
Bladewood: um.

Nescio0: It is as effective as two fuel pods and one emergency ramscoop, which cost 2*20000+72000=114000 together; and because it's less flexible, it probably ought to cost somewhat less, hence the arbitrary value of 101000.

jafdy: Perhaps too arbitrary.

Bladewood: That's not the problem here...

(There are quotation marks on the price.)

jafdy: That ought to be fine.

data/korath.txt
tehhowch: If the LFC is only plunderable (or in "Korath Advanced") then it can actually be a small improvement - e.g. `"ramscoop" 1 / 1.25`.

Nescio0: The LFC would be twice the size of the SFC but with .5/.2=2.5× its ramscoop.
I'm not sure yet if introducing a SFC would be a good idea though.

data/korath outfits.txt
```diff
@@ -91,6 +91,15 @@ outfit "Large Heat Shunt"
...
 
+outfit "Fuel Collector"
+	category "Systems"
+	cost 101000
+	thumbnail "outfit/unknown"
+	"mass" 20
+	"outfit space" -20
+	"fuel capacity" 200
+	"ramscoop" 0.5
+	description "Because of their habit of raiding systems near and far, Korath ships need a considerable amount of hyperspace fuel for their jump drives. The Fuel Collector is a bulky system which has both a ramscoop to harvest additional fuel and a large capacity to store it."
```
Bladewood: As a grammar nazi, I will not accept [the use of "which" here instead of "that".](http://www.quickanddirtytips.com/education/grammar/which-versus-that-0)

data/korath.txt
Pointedstick: FWIW, I would support removing the Fuel Pod for Korath outfitters once this is in.

data/korath ships.txt
Nescio0: Similarily, if the player plunders the jump drive prior to capturing the ship, it won't be able to jump at all either. Anyway, I've now integrated one fuel collector (i.e. 200 capacity + .5 ramscoop) fixed in each Raider and World-Ship.

Brick63:
ive started on the sprite and i have a couple ideas that are not critical
1: make it generate some heat 
2: or make it have cooling in 
also if you would like to be more involved in it i would suggest joining the discord which will help make the end result better

Amazinite:
@Brick63 I don't think that this outfit needs any more quirks than it already has.

>This proposal introduces an integrated “Fuel Collector” which is **exactly as large and as effective as two fuel pods plus one emergency ramscoop.**

Then why use this instead of two fuel pods and an emergency ramscoop? Ton for ton it might cost less, but there's no *real* reason to use this instead of two fuel pods and an emergency ramscoop. And why base a Korath (tier 2) outfit off of human (tier 1) outfits?

Here's how I'd go about balancing this:
The Korath are a more advanced species, so let's assume that instead of fitting 200 fuel in 16 tons of outfit space, the Korath have fit 200 fuel in 14 tons of outfit space (a roughly 14% increase in efficiency). That leaves 6 tons for a ramscoop. Instead of basing it off of a ramscoop that is only 30% better than standard human ramscoops (but far worse than the later unlocked catalytic ramscoop), we should base the ramscoop efficiency of this Korath ramscoop off of the Wanderer ramscoop, a similarly tiered outfit. The Wanderer Ramscoop provides roughly 0.57 ramscoop per ton. Taking a page from the Korath System Cores, which are more efficient ton-for-ton than their equivalent units from other factions put together (e.g. the Large System Core is better than equipping Coalition batteries and hull repair and Wanderer shields, which are each individually the best outfits for their single purpose), the ramscoop ability of this Korath Ramscoop should be better than the Wanderer Ramscoop. Lets not go overboard and only give this Korath ramscoop 5% better efficiency than the Wanderer Ramscoop. That puts us a 0.6 ramscoop per ton. 0.6 r/t * 6 tons = 3.6 ramscoop for this outfit.

tl;dr greatly increase the ramscoop attribute from 0.5 to 3.6 so that this is an actual tier 2 outfit worth using, instead of the current tier 1 Frankenstein of an outfit that it seems to be. I also wouldn't sell this on Efreti worlds because of this, as the Efret currently don't sell any of the other outfits they equip on their ships anyways.

data/korath ships.txt
```diff
@@ -17,14 +17,15 @@ ship "Korath Raider"
...
+		"fuel capacity" 200
 		"cargo space" 175
-		"outfit space" 721
+		"outfit space" 751
 		"weapon capacity" 284
 		"engine capacity" 159
+		"ramscoop" .5
```
Amazinite: Why include an integrated ramscoop attribute on Raiders if they'll have 2 Korath ramscoops anyways?

data/korath ships.txt
```diff
@@ -138,14 +139,15 @@ ship "Korath World-Ship"
...
 		"drag" 21
 		"heat dissipation" .4
-		"fuel capacity" 1000
+		"fuel capacity" 200
 		"cargo space" 264
-		"outfit space" 778
+		"outfit space" 848
 		"weapon capacity" 314
 		"engine capacity" 165
+		"heat dissipation" .5
```
Amazinite: Looks like you accidentally added this.

Bladewood:
I agree that the Korath fuel thing should be superior to ordinary fuel pods and emergency ramscoops, but I think making the ramscoop portion better than a Wanderer ramscoop is going too far.

Amazinite:
>I agree that the Korath fuel thing should be far superior to ordinary fuel pods and ramscoops, but I think making the ramscoop portion better than a Wanderer ramscoop is going too far.

Who do you think has more of a reason to have advanced ramscoop technology? The race that only moves into unknown territory once every tens of thousands of hundreds of thousands of years, or the race that has been known to raid hostile territory where they are unable to refuel for hundreds of years now and for the past 100 years has been stuck in a region of space where normal means of acquiring fuel are likely sparse? From a lore perspective, I'd consider the Korath to have better ramscoops. And regardless, if someone is looking for ONLY ramscoop then this outfit wouldn't be the way to go, as ton-for-ton the entire outfit is less efficient for ramscoop than a Wanderer Ramscoop.

Nescio0:
Let's give it a ramscoop of 1 then, i.e. as good as two human fuel pods + one ordinary Deep Sky ramscoop, but more compact: 20 mass instead of 8+8+10=26. Also increased its arbitrary price.

Brick63:
korath ramscoop 2x
here's what i made

Amazinite:
>Let's give it a ramscoop of 1 then, i.e. as good as two human fuel pods + one ordinary Deep Sky ramscoop, but more compact

Again, we should make this
>an actual tier 2 outfit worth using, instead of the current tier 1 Frankenstein of an outfit that it seems to be.

Best case scenario assuming that 4 tons of this outfit consist of the actual ramscoop component, this ramscoop is still pitiful compared to the Catalytic and Wanderer ramscoops (0.25 ramscoop/ton vs about 0.44 and about 0.57 respectively). And going back to what I said to Bladewood,
>From a lore perspective, I'd consider the Korath to have better ramscoops,

meaning that having a ramscoop that is worse than the Catalytic and Wanderer ramscoop doesn't add up. Unless this outfit has at least 3.5 ramscoop, there is no reason to use it instead of just individually purchasing fuel pods and Wanderer or Catalytic ramscoops.

Nescio0:
You're looking at as a ramscoop with fuel capacity, I'm looking at it as a fuel pod with ramscoop. Korath outfits are not necessarily efficient (just look at their engines and energy generators) but they are compact, designed to save space. This also applies to the proposed Fuel Collector:

two fuel pods + catalytic ramscoop: 32 mass
two fuel pods + ordinary ramscoop: 26 mass
two fuel pods + two emergency ramscoops: 24 mass
two fuel pods + wanderer ramscoop: 23 mass
one fuel collector: 20 mass

Furthermore, you can plunder Korath outfits long before you've encountered the Coalition, Remnant, Wanderer, therefore I don't see why this outfits needs to have the highest ramscoop/ton.

PS Nice sprite! Perhaps you could make the fuel containers larger?

Brick63:
I was of similar mind as the others that this was gonna be a ramscoop that held fuel so I will fix the Sprite

data/korath.txt
Nescio0: One second thought, if a SMC is to be introduced, it could be even smaller, e.g.:
* LFC: 20 mass, 200 capacity, 1 ramscoop
* SFC: 6 mass, 50 capacity, 0.25 ramscoop

So one LFC would be four times as effective as the SFC but more compact (20 instead of 4*6=24). However, I don't think introducing a SFC is necessary (a jump drive requires 200 fuel per jump, so why would the Korath producing anything with a smaller capacity?).

Tadrix:
Well, if this is intended to be a better fuel pod instead of a Korath ramscoop, then please, don't edit out original ramscoop from Korath ships, they'll gonna need it regardless of how advanced their fuel pods are going to be.

Amazinite:
>two fuel pods + wanderer ramscoop: 23 mass
one fuel collector: 20 mass

For 3 extra tons of space, a 2 fuel pod + 1 Wanderer ramscoop set up is **3.48 times better** than a single Korath ramscoop.

>Again, we should make this
>>an actual tier 2 outfit worth using, instead of the current tier 1 Frankenstein of an outfit that it seems to be.

As for this:

>Korath outfits are not necessarily efficient (just look at their engines and energy generators) but they are compact, designed to save space.

Korath cooling and System Cores are both compact and highly efficient. The Korath *are* capable of efficiency, seemingly in all areas *except* when it comes to heat generation. Korath generators produce lots of heat, but they are also the best generators ton-for-ton in terms of energy generation. Korath engines produce a lot of heat, but they are also the best engines ton-for-ton in terms of thrust and steering, and also have very acceptable energy drain. Korath weapons produce lots of heat, but they're also *very* powerful and require energy on par with the weapons of other races. What this fuel collector is currently is a very weak and inefficient outfit in all aspects, whether you look at it as a ramscoop with fuel pods or fuel pods with a ramscoop (a distinction I'd consider pointless anyways).

Nescio0:
With the current values it is already much better than the Remnant's. If you insist I could double the Fuel Collector's ramscoop again, but then Korath ships would have a much higher total ramscoop than Wanderer's, so why would anybody buy a Wanderer ramscoop then?

> Well, if this is intended to be a better fuel pod instead of a Korath ramscoop, then please, don't edit out original ramscoop from Korath ships, they're gonna need it regardless of how advanced their fuel pods are going to be.

Please have a look at the proposal as is. Currently the Korath Raider and World-Ship have a total ramscoop of 1. With the proposed Fuel Collector they'll have 3*1 and 5*1 ramscoop, respectively.

Amazinite:
>With the current values it is already much better than the Remnant's. 

And the Remnant ramscoop is a tier 1 ramscoop. This is an outfit from a tier 2 race, so I don't see why it should be so bad. 

>If you insist I could double the Fuel Collector's ramscoop again, but then Korath ships would have a much higher total ramscoop than Wanderer's

Individual Korath *ships* would have a higher total ramscoop than Wanderer ships with a single ramscoop, but I'd say that's something to be expected from a nomadic alien race with little means to refuel otherwise. That is *not* the same thing as Korath ramscoops being better than Wanderer ramscoops.

>, so why would anybody buy a Wanderer ramscoop then?

Because like I mentioned, this should be a viable option:
>Do I choose the ramscoop with built in fuel pods that would be more efficient than installing separate fuel pods and ramscoops, or do I just install a bunch of ramscoops and deal with the fuel capacity that my ship currently has?

Even at a ramscoop of 2 for this outfit, this is still not an option.

Tadrix:
@Nescio0 I'd guesstimate that 200 fuel take up 13 or 14 tons by virtue of having two tanks together or one larger tank (and following extra bunks outfit progression). What follows is: What ramscoop per ton do we want for the Korath? I'd say it shouldn't be worse than what Catalytic ramscoop has, which is 0,4375.
8 tons of Catalytic ramscoop yield 3,5 ramscoop. If Korath are to be a little better, make it 7 tons. 13 + 7 = 20. So your oufit space is ok, but ramscoop is off. Make it 3,5, then it will be a tad better than what humans are selling.

Nescio0:
Actually I've never understood those tiers. The Remnant might be tier 1, but they have decent weapons, possibly the best anti-missile turret, very heat efficient energy generators and engines, and the most compact energy capacitor. I'd like to think the Korath are somewhere in between the Remnant and the Wanderers.

> Even at a ramscoop of 2 for this outfit, this is still not an option.

Some ramscoop is nice to have, but some players like to go as quickly as possible to a certain destination. I prefer 1 Wanderer ramscoop + 5 fuel pods to 1 fuel pod + 5 Wanderer ramscoops. With a ramscoop of 2 a single Fuel Collector will be less efficient than a Wanderer ramscoop, yes, but Korath ships will have a significantly higher fuel and ramscoop total.

@Nescio0 I'd guesstimate that 200 fuel take up 13 or 14 tons by virtue of having two tanks together or one larger tank (and following extra bunks outfit progression). What follows is: What ramscoop per ton do we want for the Korath? I'd say it shouldn't be worse than what Catalytic ramscoop has, which is 0,4375.
8 tons of Catalytic ramscoop yield 3,5 ramscoop. If Korath are to be a little better, make it 7 tons. 13 + 7 = 20. So your oufit space is ok, but ramscoop is off. Make it 3,5, then it will be a tad better than what humans are selling.

Or to put it differently: 20-2*8=4; 2/4=0.5>0.4375

Amazinite:
>Actually I've never understood those tiers. The Remnant might be tier 1, but...

The Remnant, having technology that comes from multiple sources, span tiers, but their ramscoop is clearly tier 1. 
Their ships, engines, and generators are more like the Hai, but their AM turret is indeed spectacular. 

>With a ramscoop of 2 a single Fuel Collector will be less efficient than a Wanderer ramscoop, yes, but Korath ships will have a significantly higher fuel and ramscoop total.

I don't see how that's an argument to keeping the ramscoop value of an individual fuel collector low. If anything, like I said, I think that Korath ships *should* have high ramscoop values, not only for lore reasons but also because it's actually a change that will make the Korath (exiles) stand out more, them being the only faction in the game that invests heavily into ramscoop ability.

Tadrix:
One more thing — Korath systems are anything but bulky. For example, a system core doesn't take up much space when you consider what it is capable of. Korath ramscoop systems probably should follow this example.

> I'd like to think the Korath are somewhere in between the Remnant and the Wanderers.

They are, but much closer to Wanderers than to Remnant.

> Or to put it differently

Humans are still better off. Just to make use of that small difference you'd need 4 Korath ramscoops (80 t), instead of Catalytic Ramscoop (16 t), and you might not need +800 fuel (that's the point of ramscoops).

Nescio0:
Again, I'd like to emphasize it's primarily a fuel pod, with secondarily some ramscoop. If you're looking for a ramscoop, go Wanderer; if you just want fuel and save space, go Korath. 
Also I think a ramscoop value of 2 each is already ridiculously high; World-Ships now have their ramscoop increased tenfold.

Amazinite:
How about this: make the ramscoop of the fuel collector 3. This means that 2 fuel pods + 1 Wanderer ramscoop is 15% heavy but 16% more efficient, instead of the current situation of 2 fuel pods + 1 Wanderer ramscoop being 15% heavier but a whopping 74% more efficient.

Amazinite:
>if you just want fuel and save space, go Korath.

If I want more fuel and to save space, I'll just install normal fuel pods at this point.

Nescio0:
> How about this: make the ramscoop of the fuel collector 3. This means that 2 fuel pods + 1 Wanderer ramscoop is 15% heavy but 16% more efficient, instead of the current situation of 2 fuel pods + 1 Wanderer ramscoop being 15% heavier but a whopping 74% more efficient.

Or to put it differently: a single fuel collector is 13% smaller and has 50% less ramscoop than two fuel pods plus one Wanderer ramscoop, which I think is fine. Why does it have to compete with Wanderer ramscoop efficiency? Keep in mind you can actually acquire Korath outfits *before* you've started any campaign.

Nescio0:
A Korath large heat shunt is 30% more efficient than a basic human liquid helium cooler. With a ramscoop value of 1 a Korath fuel collector is 30% more efficient than the basic human equivalent. And the ramscoop value has been raised to 2 already. Making it 3 would be an overkill.

Amazinite:
>Or to put it differently: a single fuel collector is 13% smaller and has 50% less ramscoop than two fuel pods plus one Wanderer ramscoop, which I think is fine. 

How does a number as big as being 50% less useful sound fine? (And if we look at ramscoop/ton, the fuel collector is currently 42.5% worse.) Framing the numbers differently doesn't help your case.

>Why does it have to compete with Wanderer ramscoop efficiency? 

Because both are tier 2 outfits. Wanderer engines (efficient) compete with Korath engines (powerful). Wanderer generators (efficient) compete with Korath generators (powerful). Wanderer shields (powerful) compete with Korath shields (efficient). Wanderer weapons (shield heavy) compete with Korath weapons (hull heavy/balanced damage). Why *shouldn't* a Korath ramscoop compete with a Wanderer ramscoop?

>Keep in mind you can actually acquire Korath outfits before you've started any campaign.

Yes, yes you can. That doesn't change the fact that the Korath are still T2.

tehhowch:
To reduce the aggregate ramscoop on the outfitted ships, the fuel capacity can be increased from 200 to 300, and the mass bumped by only 1-2 tons. That would
1. Reduce the number needed to restore the original ships' fuel value
2. Further establish it as a "fuel pod with a ramscoop" 
3. Allow keeping a smaller ramscoop value and still be t2

Amazinite:
>To reduce the aggregate ramscoop on the outfitted ships, the fuel capacity can be increased from 200 to 300, and the mass bumped by only 1-2 tons. That would

If this fuel collector is 22 tons with 300 fuel, then 2 ramscoop is still too low for it to compete with a Wanderer ramscoop + 3 fuel pods; it would need at least 2.8 or 2.9 ramscoop to compete.

Nescio0:
Tiers are a way to broadly compare species. That doesn't mean everything of a specific tier has to be better than everything with a lower tier. Despite being a tier one the Hai actually have the best engines. 

> To reduce the aggregate ramscoop on the outfitted ships, the fuel capacity can be increased from 200 to 300, and the mass bumped by only 1-2 tons.

Why would it contain 300 fuel? A jump drive jump consumes 200 fuel, so lore-wise it would make sense the Korath fuel pod has a capacity of 200. 

> Yes, yes you can. That doesn't change the fact that the Korath are still T2.

And they are much better than human T1. For the same amount of cooling, liquid helium is 30% larger than a large heat shunt. Two fuel pods plus two ordinary ramscoops are exactly as effective as one fuel collector, but 80% larger.

Tadrix:
How about 300 fuel, 24 t and 3 _ramscoop_? (For the record, I think this is bad, 200 fuel, 20 t and 3,5 _ramscoop_ would have been better.)

@Nescio0 Wanderer ramscoop is 31% more effective than best human ramscoop.

Nescio0:
> @Nescio0 Wanderer ramscoop is 31% more effective than best human ramscoop.

But the emergency ramscoop has to be unlocked by a lengthy campaign, which makes an unfair comparison. Korath outfits are immediately available, as is the ordinary ramscoop, which is currently being used by the Korath.

Amazinite:
>Two fuel pods plus two ordinary ramscoops are exactly as effective as one fuel collector, but 80% larger.

Again, you're making comparisons with the worst of outfits.

Let me frame this for you; what you just did is like observing that the Boulder Reactor generates 3% more energy per ton than the Armageddon Core and then using that to suggest that the Ravine Battery is overkill since it is 48% better than the LP036a battery. But if we make a fair comparison using the LP576a battery, you'll actually find that the Ravine battery is 3% better than it, an improvement on par with the improvement of the generators.

Tadrix:
@Nescio0 If we compare worst cooling systems, then Korath small heat shunt is 93% better than human cooling ducts. Most of Korath tech rules. Ramscoops you're proposing would be disproportionately worse, and, as Amazinite mentioned, would be unfitting for raiders.

To get Wanderer tech one needs to get a jump drive and do some 5 or 6 missions which involve mostly flying and are available from the start. It's "go to Hai, go to Wanderers, go to Hai, go to Wanderers, go to Hai, go to Wanderers, go to the Deep, go to Wanderers". Nothing that can't be done before the Remembrance Day.

Nescio0:
> Let me frame this for you; what you just did is like observing that the Boulder Reactor generates 3% more energy per ton than the Armageddon Core and then using that to suggest that the Ravine Battery is overkill since it is 48% better than the LP036a battery. But if we make a fair comparison using the LP576a battery, you'll actually find that the Ravine battery is 3% better than it, an improvement on par with the improvement of the generators.

Actually I would compare a Ravine with a 576 and a Boulder Reactor with a Fusion Reactor, because those are the nearest equivalents. I won't compare a Pebble Core (which is small) to a Stack Core (which has to be unlocked). Likewise I think the Emergency Ramscoop is not a fair comparison.

I think the fairest comparison for the Korath are the Remnant, because you can acquire their respective outfits at about the same time. Remnant engines and energy generators are very heat efficient but less powerful than their hot Korath counterparts. The Korath and Remnant both have a great anti-missile turret which can compete with the Wanderer's. The Remnant have the best pure capacitor, the Korath the best integrated capacitor. So if the Korath need a comparison, I'd choose the Remnant. If you look at the edit history of the first post you'll notice that was actually the first comparison I made before this debate started.

Two fuel pods plus one Remnant ramscoop has 20 mass, 200 fuel capacity, and 0.5 ramscoop.
The currently proposed Korath fuel collector has 20 mass, 200 fuel capacity, and 2.0 ramscoop.

Tadrix:
Emergency Ramscoop isn't human tech, it's Remnant tech. Closest equivalent of the Pebble Core is the Dwarf Core. Getting access to Remnant tech involves more work than getting access to Wanderer tech. So if you're comparing Korath to Remnant you're bound to make their tech useless.

Amazinite:
>Likewise I think the Emergency Ramscoop is not a fair comparison.

Catalytic ramscoop*. The Emergency Ramscoop is the Remnant one you're comparing with.

>I think the fairest comparison for the Korath are the Remnant, because you can acquire their respective outfits at about the same time.

Wanderer outfits can be acquired at around the same time as well. As Tadrix mentioned, it doesn't take much to unlock Wanderer technology. If anything, it takes *less* to unlock Wanderer tech than Remnant tech, as unlocking Remnant tech requires you to be strong enough to take down a World-Ship at one point and 2 Raiders at another point, both on your own. By comparison, all it takes to get to Wanderer space is to wait around in a system like Durax and wait to snag a jump drive, something that can be done in a Shuttle. And I wouldn't agree that the time it takes you to get Korath and Remnant tech means they're fair to compare when tier has a much larger sway on the power of technology.

> If you look at the edit history of the first post you'll notice that was actually the first comparison I made before this debate started.
Two fuel pods plus one Remnant ramscoop has 20 mass, 200 fuel capacity, and 0.5 ramscoop.

Exactly, which I responded to with
>Then why use this instead of two fuel pods and an emergency ramscoop? Ton for ton it might cost less, but there's no real reason to use this instead of two fuel pods and an emergency ramscoop. And why base a Korath (tier 2) outfit off of human (tier 1) outfits?

Let's not go full circle here.

Nescio0:
> Emergency Ramscoop isn't human tech, it's Remnant tech. 

Exactly.

> Closest equivalent of the Pebble Core is the Dwarf Core. 

Exactly.

>Getting access to Remnant tech involves more work than getting access to Wanderer tech.

You need a jump drive to get to the Wanderers, you don't need one to get to the Remnant. My first encounter with the Korath was in Remnant space, which I unlocked before I discovered the well hidden Free Worlds Campaign.

> So if you're comparing Korath to Remnant you're bound to make their tech useless.

I don't think either Korath or Remnant technology is useless.

>>Likewise I think the Emergency Ramscoop is not a fair comparison.
>Catalytic ramscoop*. The Emergency Ramscoop is the Remnant one you're comparing with.

Oops, typo, thanks for pointing it out; corrected.

> As Tadrix mentioned, it doesn't take much to unlock Wanderer technology.

Only if you know what you're doing, which doesn't apply to all players.

> Let's not go full circle here.

The Fuel Collector's ramscoop value has been quadrupled during this discussion. Can we agree this is a better value than the earlier proposals? If yes, shall wethen pause the perfect ramscoop discussion for now and focus on figuring out if there are any other issues with the proposed outfit?

Amazinite:
>Only if you know what you're doing, which doesn't apply to all players.

The same can be said for unlocking Remnant technology. Especially for a new player, unlocking Remnant tech is probably going to take much longer than unlocking Wanderer tech, as there is zero direction for the player to follow to know that they need to bring a Hai keystone to Remnant space. Looting a Raider and finding your way to Wanderer space is likely going to happen first for a new player before they find out how to unlock Remnant tech.

>The Fuel Collector's ramscoop value has been quadrupled during this discussion. Can we agree this is a better value than the earlier proposals? If yes, shall we then pause the perfect ramscoop discussion for now and focus on figuring out if there are any other issues with the proposed outfit?

Yes, 2 is far better than 0.5, but that doesn't mean that we should end the discussion on the ramscoop value of this outfit here. There are also no other issues that others seem to have with the outfit that haven't been addressed, so even if we pause discussion on the ramscoop value, then we'll just fall right back in to it. I could critique the current price, but if the ramscoop value changes then the price will need to change accordingly, so I'm holding off on arguing the price until the ramscoop value is finalized.

Tadrix:
> The Fuel Collector's ramscoop value has been quadrupled during this discussion. Can we agree this is a better value than the earlier proposals?

Is this haggling? In that case your opening bid was too low to stop now.

Nescio0:
> Yes, 2 is far better than 0.5, but that doesn't mean that we should end the discussion on the ramscoop value of this outfit here. There are also no other issues that others seem to have with the outfit that haven't been addressed, so even if we pause discussion on the ramscoop value, then we'll just fall right back in to it. I could critique the current price, but if the ramscoop value changes then the price will need to change accordingly, so I'm holding off on arguing the price until the ramscoop value is finalized.

No, certainly not end, just pause. I think 2.0 is already too high, you think it's still too low, we could continue repeating our positions, we could compromise on 3.0, or we could temporarily postpone the ramscoop value, to allow others to join in the discussion, and give some time to figure out if there is anything else which has to be done.

Brick63:
there is one way we can solve this
give it like around 3.5 ramscoop
BUT in korath form make it generate some heat
it is then efficient but at a cost

tehhowch:
My opinion is that, if this is truly to be a fuel pod first and then a ramscoop (as appears to be your intention, @Nescio0 , and possibly why you are resisting increasing its ramscoop value further), then the fuel capacity of the outfit should be increased.
Indeed, Raiders need only 200 fuel to jump, but they generally make at least 2 jumps while on a raid - out and back. So your same lore argument could be used to support 400 fuel per outfit.

The gist here is that 200 & 2 in 20 space isn't commensurate with the Korath's position as an advanced race. Very powerful T2 equipment is available in the same plundered craft, so I don't feel that arguments need to focus on Korath vs Remnant, or Korath vs Wanderer, or Korath vs Human. Instead, they should focus on "Is this T2 technology, like the rest of Korath tech." Given that this is 2x T1 Fuel Pod + 2x T1 Ramscoop and the only advantage is that it consumes 16t less space, I would say no, it is not.

Amazinite:
>@Brick63 I don't think that this outfit needs any more quirks than it already has.

>give some time to figure out if there is anything else which has to be done.

Alright, if we're to argue something else, I'd say we shouldn't sell this on Efret worlds. @tehhowch seems to be the one who originally suggested that it be sold though, so I'd like to hear from him.
My reasoning for not selling the fuel collector on Efret worlds: The Kor Efret currently don't sell any outfits to the player that they have installed. They sell storage outfits for piercers and mines, but that's only because they're sold nowhere else. They sell command centers, but that's so we're able to use Sestor and Mereti ships as our flagship. The rest of what they sell, aside from the human ramscoop, is just basic outfits. For them to sell this outfit among the rest of what they sell seems out of place to me. If the collector won't be sold though, I'd suggest that fuel pods remain for sale at Efret planets.

Amazinite:
>Lorewise, they should be proud of this outfit - it was developed to support their warring ways, sure, but it's just really nice for touring the galaxy.

I'd say lore wise, the better argument is that they use and sell it of necessity (instead of "lol it's cool, buy it"), as it's likely very difficult for them to harvest hyperspace fuel by normal manners (e.g. the deuterium factories in human space that harvest from gas giants), but they'd be ashamed of why it was built and the fact that they now rely on it.

Tadrix:
> Lorewise, they should be proud of this outfit

Especially because this by itself isn't a war technology. One could travel the galaxy to make friends.
If Kor Efret were selling basic ramscoop, there's nothing preventing them to sell Korath Fuel Tanks.

> 400 fuel per outfit

I'd start with the assumption that 400 _fuel_ Korath Fuel Tanks without ramscoops would take up around 26 tons and go from there. So it could, for example, be 27 t w/ 0,5 _ramscoop_, 28 t w/ 1 _ramscoop_, 29 t w/ 1,5 _ramscoop_ or 30 t w/ 2 _ramscoop_ (being a little better than Catalytic Ramscoop). YMMW.

Nescio0:
A bit of background:
* Currently the Raider and World-Ship are equipped with inefficient ordinary human ramscoops. Because of that, I proposed to give them internal ramscoops of 2.0 instead #3489 (some argued 2.0 was too high).
* If they have internal ramscoops, the Korath outfitter no longer has to sell the Deep Sky's ordinary ramscoop
* But then inexperienced players who don't know what they're doing and have never seen any ramscoop might get into trouble in Efret space, so it was proposed to create a Korath ramscoop instead to be sold there.
* Yet another pure ramscoop would be uninteresting (which would either be more efficient than the Wanderer's or worse); I didn't like the suggestion to have it produce heat (the Korath already have plenty of that), or cooling (that's what heat shunts are for), or alter the heat dissipation coefficient.
* However, I think a Korath fuel pod with some ramscoop would be interesting, hence this proposal.
* It is primarily a fuel pod; the ramscoop is a secondary bonus.

> My opinion is that, if this is truly to be a fuel pod first and then a ramscoop (as appears to be your intention, @Nescio0  , and possibly why you are resisting increasing its ramscoop value  further), then the fuel capacity of the outfit should be increased. Indeed, Raiders need only 200 fuel to jump, but they generally make at  least 2 jumps while on a raid - out and back. So your same lore argument  could be used to support 400 fuel per outfit.

Perhaps this is the way to go then. How about:
Fuel Station (Large): 32 mass, 400 fuel, 2.0 ramscoop, price to be determined, description to be rewritten.

Korath Raider would have internal 200 fuel and 1.0 ramscoop, plus one large fuel station equipped, bringing it to 600 fuel and 3.0 ramscoop (currently 600 and 1.0)

Korath World-Ship would have internal 200 fuel and 1.0 ramscoop, plus two large fuel stations equipped, bringing it to 1000 fuel and 5.0 ramscoop (currently 1000 and 1.0)


And perhaps also a Fuel Station (Small): 9 mass, 100 fuel, 0.5 ramscoop, price to be determined, description to be written, for sale at Efreti outfitters, replaces human fuel pod (8 mass, 100 fuel) and ordinary ramscoops (10 mass, 1.0 ramscoop) there.

Tadrix:
> ecause of that, I proposed to give them internal ramscoops of 2.0 instead #3489 (some argued 2.0 was too high).

Yes, I saw that. In my opinion internal ramscoop of 2 in Korath Raiders sounded good.

> If they have internal ramscoops, the Korath outfitter no longer has to sell the Deep Sky's ordinary ramscoop

Not necessarily. Kor Efret aren't using Korath Raiders, so they could be using basic ramscoop in their outfitters.

> Fuel Station (Large): 32 mass

I'd suggest 30 t, to make it better than human fuel tanks. And you could just as well give Korath Raiders internal ramscoops and equip World Ships with Fuel Stations.

Amazinite:
If these fuel stations are where we're bringing this, I'd suggest the following stats:
Large:
* 423k credits
* 29 tons
* 400 fuel
* 3.2 ramscoop

Small:
* 103k credits
* 9 tons
* 100 fuel
* 0.9 ramscoop

Those cost numbers are probably on the low end, though.

Nescio0:
Four universal fuel pods (which the Wanderers use, produce, and sell as well) have mass 32, so at the proposed mass of 32 the 2.0 ramscoop is free, making it far more efficient per mass than any ramscoop.

The small heat shunt has a mass of `(large heat shunt mass) / 4 + 1 = 24 / 4 + 1 = 7`, therefore `32 / 4 + 1 = 9` seems a fine mass for the small fuel station. Compared with the universal fuel pod, this means 0.5 ramscoop for just +1 mass, which is highly efficient.

Brick63:
korath ramscoop large 2x
changes to fit the large station
the main thing I've heard is that some think it looks a bit too much like it has a heat exchanger and not enough like it has a ramscoop

KiLEdEnNis:
pull request: `Introduced Korath Fuel Station (fuel pod with ramscoop)`

commit 5 mins ago: `improved Remnant ramscoop`

I'm sorry but that's for another PR

Brick63:
its a referance in a commit

KiLEdEnNis:
> its a referance in a commit

~~then edit the title to be clear~~

Nescio0:
What if you would put the fuel pods in the centre and attach a smaller but longer scoop in front?

Tadrix:
> at the proposed mass of 32 the 2.0 ramscoop is free, making it far more efficient per mass than any ramscoop

There is a cost: You have to buy 400 fuel as a unit. If your ship's hull holds, for example, 500 fuel, then with this outfit you could make 4 jumps, and 100 fuel capacity would be wasted. Whereas one could buy 300 fuel capacity, taking up 24 t, and add a Wanderer ramscoop on top of it (31 t total). Less waste, more ramscoop efficiency.

My or Amazinite's propositions would ensure this outfit stays relevant even in a such situation.

Nescio0:
> There is a cost: You have to buy 400 fuel capacity as a unit. If your ship's hull holds, for example, 500 fuel, then with this outfit you could make 4 jumps, and 100 fuel capacity would be wasted. Whereas one could buy 300 fuel capacity, taking up 24 t, and add a Wanderer ramscoop on top of it (31 t total). Less waste, more ramscoop efficiency.

Ship's 500 + station's 400 = 900, enough for 9 hyperdrive jumps, 6 scram drive jumps, or 4 jump drive jumps. In such a scenario I would probably just buy another pod to make it 1000. 32+8=40 mass with 2.0 ramscoop looks better to me than 5*8+7=47 mass with 4.0 ramscoop.

Also, I'd like to emphasize it's not a ramscoop; it's a fuel pod that has a ramscoop as a bonus. Besides, even if you insist on comparing it with ramscoops instead of fuel pods, 0.5 ramscoop/mass is highly efficient.

Amazinite:
>Also, I'd like to emphasize it's not a ramscoop; it's a fuel pod that has a ramscoop as a bonus. 

If that's the case, then these should provide more fuel for less mass than normal fuel pods, instead of being the exact or more mass as the equivalent fuel in fuel pods. That would at least give it some sort of advantage over the increased ramscoop efficiency of the fuel pod + ramscoop combo of other ramscoops that I've been stressing. 

And I'd also like to mention that whether you designed it as a fuel pod with extra ramscoop ability or not, I'd honestly expect most people to see it as a ramscoop with fuel pods attached to it, not only by the image that @Brick63 has made but also in its stats, and it will likely see little use because of that since it isn't inherently more efficient in any way, either as only a fuel pod, as only a ramscoop, or as a combo fuel pod-ramscoop.

solardawning:
Agreed with Amazinite and Tadrix here- if it's supposed to be a larger fuel pod, it should be providing fuel at a higher efficiency to outfit space than a regular fuel pod does.

You could make the argument that ~400 fuel chunk isn't enough of a commitment to fuel to justify much efficiency in outfit space, though.

You could try making it 600 fuel, 2 ramscoop, for 40 outfit space. 
That way the value is clear, but it's only a sure bet for someone who would want to install at least 5 fuel pods anyways.

Gwenio:
I have a quibble with the proposed art: a bunch of small fuel pods are space/mass inefficient compared to a single large one.

As to numbers, I recommend deciding on some guidelines for how much mass should be saved by having a single large outfit instead of a collection of smaller ones that give similar stats.

Amazinite:
Going with the idea of being mainly fuel pods that are more efficient than normal fuel pods, these would be the numbers I suggest:
Large:
* 896k credits
* 27 tons
* 450 fuel
* 2.7 ramscoop

Small:
* 288k credits
* 9 tons
* 125 fuel
* 0.9 ramscoop

Notes:
* Each fuel station has 0.1 ramscoop per ton, an efficiency equal to that of the basic Ramscoop. That is, it could replace a normal Ramscoop, but other ramscoops would be better to use if you're looking for just ramscoop ability.
* The small fuel station is about 11% more efficient than a normal fuel pod, and the large fuel station is about 33% better.
* The large fuel station is about 20% better than the small fuel station, a different that is near the roughly 17% increase in efficiency that the large heat shunt has over the small heat shunt.

Nescio0:
Would I purchase a small fuel station, if I have unlocked wanderer and all other technology? Certainly. Why? Because it saves space. A SFS has mass 9, a fuel pod + Remnant ramscoop has 8+4=12 mass.
Would I spend time on raiding Raiders to acquire a large fuel station. Certainly. Why? Because it saves space. A LFS has mass 32, four fuel pods + Remnant ramscoop has 4*8+4=36 mass.

> As to numbers, I recommend deciding on some guidelines for how much mass should be saved by having a single large outfit instead of a collection of smaller ones that give similar stats.

The proposed values follow the Korath heat shunt guideline. A large heat shunt is exactly four times as effective as a small heat shunt; the large mass is (small mass - 1) * 4 = (7 - 1) * 4 = 24.
Therefore a pure quadruple fuel pod (without ramscoop) would have mass (8 - 1) * 4 = 28.
The proposed small fuel station has mass 9, slightly larger than a fuel pod to justify its ramscoop. Therefore the large fuel station has mass (9 - 1) * 4 = 32. 
This works out at 0.5 ramscoop/mass, slightly below the expensive Wanderer's pure ramscoop, per:

> I agree that the Korath fuel thing should be far superior to ordinary fuel pods and ramscoops, but I think making the ramscoop portion better than a Wanderer ramscoop is going too far.

Which I agree with (if you can raid a Raider to get a jump drive to go Wanderer, you can also simply plunder the Raider's LFS directly, which is why I think the Wanderers deserve the most efficient ramscoop).

> You could try making it 600 fuel, 2 ramscoop, for 40 outfit space. That way the value is clear, but it's only a sure bet for someone who would want to install at least 5 fuel pods anyways.

I could make the LFS 30 mass, 400 fuel, 1 ramscoop, but then the SFS would be a problematic 8.5 mass.

Amazinite:
>Would I purchase a small fuel station, if I have unlocked wanderer and all other technology? Certainly. Why? Because it saves space. 

Yes, the fuel stations as they stand do save *a few tons* of space, but that is *all* they do right now. The thing is, that isn't very useful for a fuel pod/ramscoop. When it comes to fuel pods and ramscoops, you ideally want as much as you can get. This isn't like supercapacitors, where 150 energy can be all you need. If you're low on space and you want a ramscoop and fuel pods then yes, you *might* find yourself installing a fuel station, but a majority of the time people are looking to max out the power they get from their outfits, in which case a player will **still** want to look for fuel pods + Wanderer Ramscoops.

Nescio0:
> Yes, the fuel stations as they stand do save a few tons of space, but that is all they do right now. The thing is, that isn't very useful for a fuel pod/ramscoop. When it comes to fuel pods and ramscoops, you ideally want as much as you can get. This isn't like supercapacitors, where 150 energy can be all you need. If you're low on space and you want a ramscoop and fuel pods then yes, you might find yourself installing a fuel station, but a majority of the time people are looking to max out the power they get from their outfits, in which case a player will still want to look for fuel pods + Wanderer Ramscoops.

Nobody needs supercapacitors, Blue Sun reactors and Large Systems Cores already have plenty of energy capacity :)

Anyway, two LFS are exactly as effective as eight fuel pods and one Wanderer ramscoop, yet do save space (64 compared to 71). I would prefer the former. Saving space is my priority in re-outfitting my ships. Some ramscoop is good enough for me, I don't need five ramscoops per ship. I assume we simply have quite different playing styles.

solardawning:
I think for most people (and many ships, specifically), 8 fuel pods and a wanderer ramscoop is just too much to be worthwhile. You look at it as 'saving space', many other people would see it as not being a sensible choice, because we wouldn't be cramming that much fuel into most ships anyways. We'd be aiming for specific numbers that we could hit with a one, two, or three fuel pods plus a ramscoop.

Unless the efficiency gain on the proposed Korath outfit is enough where you can get more fuel for less space than using individual fuel pods, it's hard to see it being very useful.

Nescio0:
> 8 fuel pods and a wanderer ramscoop is just too much to be worthwhile.

Agreed. Most ships need just four or five fuel pods to get to 1000 fuel, having more is not really necessary if you're using hyperdrives. Nevertheless, if the choice is between four fuel pods plus a Wanderer ramscoop (39 mass, 400 fuel, 4 ramscoop) or one large fuel system (32 mass, 400 fuel, 2.0 ramscoop), I would still choose the latter, sacrificing some ramscoop to save space.

Tadrix:
> (32 mass, 400 fuel, 2.0 ramscoop)

As I said, make the mass 30 t at most. Leave the small storage as it is — it ought to be slightly less effective.
Or you could divide everything by two for the larger storage, for 200 fuel, 15 t and 1 _ramscoop_.

tehhowch:
I think the larger size of plundered outfit makes sense, as available outfit space is generally the limiting factor for an early game plundering style - you might be able to obtain these outfits, but using them in your cheap ship (e.g. Clipper, Fury) is tough because you don't have the space to meet the usage requirements. For things like weapons, engines, and shields, that's space for generation and cooling. For this fuel system, it's just the size that makes it difficult to use. Not unusable, just difficult - even when it's clearly superior, the player may not be able to effectively utilize it for a while, leaving them open to interesting gameplay decisions, which is an ultimate goal of the game - it should be fun to play, and different styles should be viable.

Summary: I like the 9t small available for purchase from a hard-to-reach outfitter, and the large "wowza" fuel system that can be plundered easily, but not immediately used easily - just like other T2 tech when the player is starting out.

Nescio0:
Forgive me my lack of flexibility, but my proposals are based on the following restricting principles:
* it is intended to be primarily a fuel pod; the ramscoop is a bonus
* a fuel pod with ramscoop should be larger than a pure fuel pod of the same capacity
* it should not exceed 0.5 ramscoop/mass to allow the Wanderer ramscoop (4/7) remain the best
* it should be much better than the ordinary ramscoop (1/10) the Korath currently use
* its fuel capacity should be a multiple of 200 (jump drive jump); otherwise problematic lore wise
* it should be Korath like (the heat shunts and systems cores are probably the outfits closest to it)
* Korath Raider and World-Ship should have a total ramscoop in the same order of magnitude as Kor Mereti and Sestor (3) or Wanderer (4)
* larger outfits should save some mass when compared to its smaller counterparts, but reasonably, e.g.:
  * small heat shunt 7, medium heat shunt 7*2-1=13, large heat shunt (7-1)*4=(13-1)*2=24
(yes, I'm aware the middle one doesn't exist)
  * fuel pod 8, double fuel pod 8*2-1=15, quadruple fuel pod (8-1)*4=(15-1)*2=28
  * small fuel station 9, medium fuel station 9*2-1=17, large fuel station (9-1)*4=(17-1)*2=32

> As I said, make the mass 30 t at most. Leave the small storage as it is — it ought to be slightly less effective.

Acceptable, but since I'm not willing to go beyond 0.5 ramscoop/mass, it'll be 30 mass, 400 fuel, 1.0 ramscoop then, and the small one 9 mass, 100 fuel, 0.25 ramscoop (which is low). The Korath Raider will then have a total ramscoop of 1.5, the World-Ship 2.5.

> Summary: I like the 9t small available for purchase from a hard-to-reach outfitter, and the large "wowza" fuel system that can be plundered easily, but not immediately used easily - just like other T2 tech when the player is starting out.

It seems there is no disagreement on the small fuel station's size, which is at least something. Determing its exact ramscoop is more difficult.

Amazinite:
>Forgive me my lack of flexibility, but my proposals are based on the following restricting principles:

Would have been helpful to get these right out the gate. Now I'd like to go through these and 

>it is intended to be primarily a fuel pod; the ramscoop is a bonus

Understandable.

>a fuel pod with ramscoop should be larger than a pure fuel pod of the same capacity

This seems to contradict the previous principle; this outfit is suppose to be primarily a fuel pod, BUT you want it to be larger (i.e. worse) than a normal fuel pod of the same capacity? This goes back to what we're saying: a player has little to no reason to use such an outfit because it is not inherently more efficient than the alternative, meaning that you're adding a relatively useless outfit that could otherwise make for a great alternative.

>it should not exceed 0.5 ramscoop/mass to allow the Wanderer ramscoop (4/7) remain the best

A threshold that has yet to be passed by any suggestions.

>it should be much better than the ordinary ramscoop (1/10) the Korath currently use

Much better how? Your current suggestions have it being much worse.

>its fuel capacity should be a multiple of 200 (jump drive jump); otherwise problematic lore wise

How is it problematic lore wise to have anything other than a multiple of 200? And this is a very restrictive principle, as it locks us into a position that leave very little freedom to make this outfit worth using.

>it should be Korath like (the heat shunts and systems cores are probably the outfits closest to it)

Expected of a Korath outfit, but I don't see how making it worse than the sum of its parts in other outfits is very Korath like. You might want to read over https://github.com/endless-sky/endless-sky/issues/2852 where we end up breaking down just how the Large System Core is essentially the best outfit in the game if you're looking for shield generation *and* hull repair *and* energy capacity, as it is more compact *and* more powerful than its individual competitors by each metric. The way you have currently designed the fuel stations, they are more compact but they are **not** more powerful, which is a point I have been arguing on since the beginning.

>Korath Raider and World-Ship should have a total ramscoop in the same order of magnitude as Kor Mereti and Sestor (3) or Wanderer (4)

For what reason aside from an arbitrary one? Raiders and World-Ships spend most of their time in space. The Wanderers and Kor Mereti/Sestor ships are at least able to land on planets and/or stations to refuel if they need to. The Korath Exiles on the other hand are left adrift around the Core and therefore it would make sense for them to have *considerably* more ramscoop than other factions, as they rely on it more. (There are Korath stations around the Core, but those are very literally just repurposed World-Ships that lore wise likely don't have all that much fuel to give).

>larger outfits should save some mass when compared to its smaller counterparts, but reasonably, >e.g.:
>* small heat shunt 7, medium heat shunt 7*2-1=13, large heat shunt (7-1)*4=(13-1)*2=24
>* (yes, I'm aware the middle one doesn't exist)
>* fuel pod 8, double fuel pod 8*2-1=15, quadruple fuel pod (8-1)*4=(15-1)*2=28
>* small fuel station 9, medium fuel station 9*2-1=17, large fuel station (9-1)*4=(17-1)*2=32

Size progression doesn't always have to be so neat, especially among alien races. This seems like an unnecessary and overly restrictive principle.

Bladewood:
I'll point out that the repair modules and battery modules are an unfair comparison with the system core, given that they're merely tier 1.5.

Amazinite:
>I'll point out that using repair modules and battery modules makes for an unfair comparison with the systems core, given that they're merely tier 1.5.

They're the best outfits for their sole purpose though, so there's no other available comparison to be made.

Nescio0:
@Brick63, I think one large scoop entry (cf. Remnant) would be more Korath-like than an array of smaller holes (cf. Wanderer). Perhaps you could design the small fuel station first? Basically take the fuel pod sprite, attach the emergency ramscoop in front, make it more compact, and repaint it in Korath style. (Easier said than done.)

@Amazinite, I have re-read each of your comments in this discussion a few times and I can follow your line of thought. I just happen to disagree with you in detail. Likewise, I don't expect you to fully subscribe to my principles, but I hope you could at least have a look at it from my perspective.

> a fuel pod with ramscoop should be larger than a pure fuel pod of the same capacity
it should not exceed 0.5 ramscoop/mass to allow the Wanderer ramscoop (4/7) remain the best
it should be much better than the ordinary ramscoop (1/10) the Korath currently use

Per @Bladewood: “I agree that the Korath fuel thing should be far superior to ordinary fuel pods and ramscoops, but I think making the ramscoop ***portion*** better than a Wanderer ramscoop is going too far.”

> Korath Raider and World-Ship should have a total ramscoop in the same order of magnitude as Kor Mereti and Sestor (3) or Wanderer (4)

Because in #3489 people argued the Raider and World-Ship ramscoop should remain well below the more sophisticated Mereti and Sestor.

It is hard to please everyone. 

Let's re-double the ramscoop values and have the Raider and World-Ship end up at 3 and 5, respectively.

Nescio0:
> I'd actually like to know why you wouldn't want to use the stats I provided. They'd provide everything that you want (a primarily fuel based Korath outfit that also has some ramscoop) while also hitting on what others want (an outfit that is a worthy alternative to normal fuel pods and ramscoops in more than just a size sense, making the use of it situational or preference based).

Keep in mind the fuel pod sold by the Wanderers has 100 fuel / 8 mass. Also, let's compare your proposed values with the current proposal:

Proposed Fuel Station (Small): 0.17m, 9 mass, 100 fuel, 0.5 ramscoop
Your suggestion: 0.103m, 9 mass, 100 fuel, 0.9 ramscoop
Proposed Fuel Station (Large): 0.69m, 30 mass, 100 fuel, 2.0 ramscoop
Your suggestion: 0.423m, 29 mass, 100 fuel, 3.2 ramscoop

Can't you agree the outfits are actually statistically close to each other? The cost can be tweaked to anything you think acceptable. The ramscoop portion is somewhat lower than you'd like, but higher than what was deemed acceptable in #3489.

Amazinite:
>Keep in mind the fuel pod sold by the Wanderers has 100 fuel / 8 mass.

Yes, 12.5 fuel/ton. Your proposals are either at or below that value, meaning that if the intention is for these outfits to be mainly fuel pods that they're no better or even worse than normal fuel pods.

>Also, let's compare your proposed values with the current proposal:

Those aren't my numbers. These are:

Large:
* 896k credits
* 27 tons
* 450 fuel
* 2.7 ramscoop

Small:
* 288k credits
* 9 tons
* 125 fuel
* 0.9 ramscoop

Nescio0:
400/30=13.333

> Those aren't my numbers. These are:

screenshot from 2018-01-17 14-31-17

Brick63:
yes but 17 hours ago he revised it

Amazinite:
screen shot 2018-01-17 at 8 33 12 am
screen shot 2018-01-17 at 8 33 00 am
screen shot 2018-01-17 at 8 33 07 am

Nescio0:
Yes, I've been constantly updating this proposal. Reading back starting at the top, the current version is significantly closer to your earlier suggestion than it is to my original idea. Also #3489.

TH1813254617:
Why not make it a Korah fuel compressor or Korath fuel refinery. A ramscoop that can collect fuel and compress it, so that the existing fuel capacity would increase, not just because the outfit itself holds fuel. The lore would be that the Korath, with their typical disregard for the environment, found a way to compress hyperspace fuel to dangerous pressures. Due to apparent safety concerns, only the compression capabilities of the one Korath fuel processor will be used.
Something like:
15 tons
increases fuel capacity by 30%
3.5 Ramscoop
50 compression energy (the ramscoop always consumes this amount of energy)

This way, this ramscoop will no longer have to directly compete with fuel pods or ramscoops, it is a ramscoop that compresses your fuel. Since it is more efficient on ships with larger fuel capacities, fuel pods are still desired. As a trade-off, it guzzles your energy and only one works at a time.

KiLEdEnNis:
increasing a stat by a percentage is not supported, and doing that for just this outfit is not worth it
adding that a player could freely buy several unless we'd restrict them with a special tag name: every ship in the game would have "tag name" 1 and this outfit "tag name" -1 which is a huge no
unless we'd add support for limiting the amount of certain outfits per ship, which is also not going to happen, for reasons I don't remember

in short; not gonna happen

TH1813254617:
Oh, OK... should this outfit be sold in the Korath outfitters? A fuel station is not an instrument of war, after all.

KiLEdEnNis:
I see no reason not to, and since Korath space lacks refueling places I think they really have to be sold by the Efreti

Lineth:
+1 for giving the Efreti more stuff to sell.

KiLEdEnNis:
~~"it's dangerous to go alone, take this!" where a lizard gives you the fuel station
you all know what I mean~~

endless-sky:
It looks like two people had the same idea to create a Korath fuel pod at the same time (#3570); I merged that one instead of this one because it had an outfit sprite to go along with it.

Issue #3534, opened 2018-01-12
Content request: hyperdrives and fuel pods


Nescio0:
Currently about all species use human hyperdrives, even though it is implied they developed independently of humans. It would be great if there were Coalition, Hai, Korath, Remnant, and Wanderer hyperdrives as well, so they no longer have to use the human version.

KiLEdEnNis:
that has been an idea for long, but the problem is there may be 1 that's straight-up better; the only stats to really balance it around is fuel usage, jump speed and the size
apart from that I'm not sure if it got a "no" or a "someday"

Nescio0:
> fuel usage, jump speed and the size

That should give plenty of possibilities, e.g. human as it is right now, Hai 10% smaller, Korath +100 fuel capacity, Remnant 10% faster, Wanderer 10% less fuel consumption. And perhaps one hyperdrive would consume some energy or produce heat.
And even if they would be statistitically identical, having different icons and descriptions for different species would add flavour to the game.

Amazinite:
#1455

KiLEdEnNis:
MZ, 28 August 2016

> I'm trying to avoid adding even more ways where, when outfitting a new ship, you have to go around and collect the best outfit of a particular kind from each species - e.g. Korath cooling plus Deep atomic thrusters, etc. So I'm not sure I like the idea of the hyperdrive becoming one more thing that is different for every species.

Brick63:
i would say that you could just have the hyperdrives be different sprites instead of different stats but that would just be extra fluff

endless-sky:
The basic idea is that all hyperdrives are interchangeable, in the same way that one species' cargo expansion behaves no differently from another's. Note that if there were many different hyperdrive variants, missions that are gated on what sort of drive you have would become a lot more complicated.

KiLEdEnNis:
> If they're all interchangeable it's still possible to have statistically identical hyperdrives with different sprites, names, and descriptions for different species.

"with different names"

I'll redirect ya to MZ:
> Note that if there were many different hyperdrive variants, missions that are gated on what sort of drive you have would become a lot more complicated.

to create different species' hyperdrive images for flavor would require new outfits which, like MZ said, make these missions more complicated

Nescio0:
> remember the Unfettered Jump Drive donations? 

Those do not *require* an hyperdrive, they *grant* one. Having multiple versions won't make this any bit more complex.

Anyway, how about giving each species its own fuel pod? E.g.:
[...]

PR #1922, opened 2016-12-24
Human Hai Fleets

Amazinite:
Addresses #1577 by making the Human fleets in Hai space sometimes use Hai outfits, very rarely using Hai ships.

PR #3559, opened 2018-01-15
Removed Supercapacitor from Hai outfitters

Nescio0:
None of the Hai ships or variants actually uses the human supercapacitor, therefore it doesn't make sense their outfitters would sell it nevertheless.


Bladewood:
Remember that human merchants live in Hai space, too.

KiLEdEnNis:
@Bladewood in that case the Hai should use it too at times
but the Hai don't really sell human tech apart from the Supercapacitor, so this PR makes sense

MCOfficer:
@Bladewood there are no actual human ships spawning, so merchant probably doesn't mean starship captain

jafdy:
The Hai sell projectile weapons so being able to buy a capacitor quickly is helpful.

Nescio0:
But why would anybody in Hai space want to *produce* an inefficient human capacitor? Hai batteries are significantly better (mass efficient and cost effective):
```
                          capacity per mass ;    capacity per credit
Human Supercapacitor  :   150 /   2 =  75.0 ;   150 /  25000 = 0.006
Hai Chasm Batteries   :   491 /   5 =  98.2 ;   491 /  18000 = 0.027
Hai Fissure Batteries :  1709 /  15 = 113.9 ;  1709 /  22000 = 0.078
Hai Gorge Batteries   :  3761 /  30 = 125.4 ;  3761 /  55000 = 0.068
Hai Ravine Batteries  :  8205 /  60 = 136.8 ;  8205 / 124000 = 0.066
Hai Valley Batteries  : 17777 / 120 = 148.1 ; 17777 / 300000 = 0.059
```

KiLEdEnNis:
in terms of usefulness I'm against this PR as well
but realistically the Hai shouldn't sell them; it's the only actual human tech they sell
and since the Hai have 5-ton batteries it's not that bad if the Supercapacitor was removed

Nescio0:
> in terms of usefulness I'm against this PR as well

Why? The Chasm is slightly larger but significantly more powerful and cheaper.

> also @Nescio0 now you made me realize the relative prices between the Hai batteries are a bit... messed up, especially of the Fissure ones
but that's not relevant to this PR

Yeah, the Chasm battery should be much cheaper in order to achieve increasing mass efficienct and decreasing cost effectiveness.

Bladewood:
@MCOfficer There are plenty of human ships in Hai space...

Although, now, I've renounced my decision, because I realize that a supercapacitor is completely useless in a space dominated by ion weaponry.

KiLEdEnNis:
> Why? The Chasm is slightly larger but significantly more powerful and cheaper.

the Supercapacitor being available in Hai space is useful

Amazinite:
>But why would anybody in Hai space want to produce an inefficient human capacitor? Hai batteries are significantly better (mass efficient and cost effective):
>Why? The Chasm is slightly larger but significantly more powerful and cheaper.

Ships that always run with max energy are actually better off spending as little space as possible on batteries, so the 2 ton supercapacitor remains helpful whether or not it costs more and gives less energy than the 5 ton Chasm battery, as long as your ship doesn't have any weapons that need more than 150 energy to fire. Whether supercapacitors should remain on sale in Hai space is a different story, though.

Nescio0:
This proposal is not about the usefulness of the human supercapacitor, but about whether or not the Hai should produce it and sell it at their outfitters.

ghoulavenger:
Personally I don't see a reason to remove the supercapacitor. The supercapacitor is just the smallest possible battery that remains useful. It isn't advanced tech, it's a battery which is pretty much universal. The only reason not to sell a supercapacitor imho is that none of your tech can take advantage of such a small battery (which would make it so obsolete that nobody would want it). But I leave that decision up to the creators of the content (if they don't want their race selling supercapacitors I'm okay with that too).

Amazinite:
>This proposal is not about the usefulness of the human supercapacitor, but about whether or not the Hai should produce it and sell it at their outfitters.

Then if we can agree that the supercapacitor is useful, then what is the gameplay reason to remove it from Hai outfitters? Perhaps the Hai saw it as being useful and produce them based off of human designs. While no Hai ships use the supercapacitor, the Arrow (Hai) variant that appears in Hai space uses two. (It's only one variant, I know, but it's at least something.)

Nescio0:
> what is the gameplay reason to remove it from Hai outfitters?

Lore. Ideally each species should have each own unique outfits and aliens should not sell any human things. (See also #3534 and #3535.) The Hai are the only aliens who sell the human Supercapacitor. If removing it is unacceptable, perhaps a Hai supercapacitor could be created (60 energy capacity, 1 mass), but I believe that's unnecessary, because the Hai do not use any supercapacitors nor really need them.

>  the Arrow (Hai) variant that appears in Hai space uses two.

The Arrow is not produced or sold in Hai space either; its use of supercapacitors can be easily explained by the fact it simply purchased them in human space when the Arrow was travelling towards Hai space.

Amazinite:
> Lore

Lore isn't gameplay.

>Ideally each species should have each own unique outfits and aliens should not sell any human things.

I'd consider #3534 and #3535 to be bad examples of this, since #3534 has its technical reasons and #3535 mentions the Remnant, who have connections to the other human factions. #3557 or #3570 are better examples, as the Kor Efret have 0 connection to humans and therefore shouldn't be selling human Ramscoops. The big difference here though is that the Hai do have connections to humans. Very strong, constant connections that stretch back over 200 years. They aren't like the Kor Efret with 0 contact to humanity or the Remnant who could reasonably have remnants of human technology still laying around (pun not intended). I would consider it within reason that the Hai could be selling some human technology, even only this single outfit.

>The Arrow is not produced or sold in Hai space either

Neither are any of the other human ships flying in Hai space, so I don't see how this is a point.

>its use of supercapacitors can be easily explained by the fact it simply purchased them in human space when the Arrow was travelling towards Hai space.

The general idea of the `(Hai)` variants [when I made them](https://github.com/endless-sky/endless-sky/pull/1922) was "what kind of changes could I do to my ship using only Hai outfitters if I was on a budget." So the intent of the `(Hai)` variants is that they bought all their outfits that differ from their normal variants while in Hai space. 

It's also notable that not all merchants in Hai space came from human space; some where born there, and perhaps inherited a human ship or bought it from someone else. The owner of the Arrows you see could have come from human space with an extra supercapacitor in tow, or the owner could have bought a stock Arrow (1 superacapcitor) from someone else and later upgraded it with technology they could buy from the Hai, such tech including supercapacitors at the time of the making of the `(Hai)` variants.

Nescio0:
Originally the Hai had human batteries, energy generators, and engines as well. Those were repainted and assigned different stats to create unique Hai outfits. I think that was a great change.
For me minor details such as the supercapacitor are important.

Nescio0:
> The general idea of the (Hai) variants when I made them was "what kind of changes could I do to my ship using only Hai outfitters if I was on a budget."

By the way, the Arrow (Hai) variant seems to have a free capacity of 7, which means that it could have a LP036a, resulting in 1000 energy capacty and unused outfit size of 1 (instead of 300 and 7), allowing it to outlive one ion hit (which drains 500 battery). Moreover, it would cost only 12000 instead of 2*25000=50000, thus saving the merchant 76% in battery cost.

That's not really relevant for this discussion though.

Issue #3554, opened 2018-01-15
Gas planet descriptions on galaxy map


Tadrix:
It would be nice if descriptions for Nasqueron and Slylandero in Nenia system after landing on them could be read even when player's flagship isn't a Gascraft. They still could be grey (indicating that no landing is possible), but would be selectable (the purpose of this being the same as for other planets with readable descriptions).

Zitchas:
This is kind of an old issue, but it is increasingly relevant with the larger number of missions in Remnant space using gas giant planets, and in particular having an inhabited gas giant.

Amazinite:
This could potentially be achieved using #11318. The knowledge of whether a planet is landable (and therefore whether you could see it on the map) could be controlled by a node that checks whether your flagship has the gaslining attribute, or if you'd visited the planet before.

PR #3563, opened 2018-01-16
Allow selection of inaccessible planets in the map panel if the player has visited them

Hadron1776:
Refs #3554 

This PR will allow a player to read the descriptions of Nenia's gas giants once he has visited them.

### DIFF
--- a/source/MapDetailPanel.cpp
+++ b/source/MapDetailPanel.cpp
@@ -442,7 +442,8 @@ void MapDetailPanel::DrawInfo()
 				// The same "planet" may appear multiple times in one system,
 				// providing multiple landing and departure points (e.g. ringworlds).
 				const Planet *planet = object.GetPlanet();
-				if(planet->IsWormhole() || shown.count(planet))
+				if(planet->IsWormhole() || shown.count(planet) 
+					|| !(planet->IsAccessible(player.Flagship()) || player.HasVisited(planet)))
 					continue;
 				shown.insert(planet);
 				
@@ -481,7 +482,7 @@ void MapDetailPanel::DrawInfo()
 				
 				bool hasVisited = player.HasVisited(planet);
 				bool isAccessible = planet->IsAccessible(player.Flagship());
-				font.Draw(!isAccessible ? "(Inaccessible)" : hasVisited ? "(has been visited)" : "(not yet visited)",
+				font.Draw(!isAccessible ? "(inaccessible)" : hasVisited ? "(has been visited)" : "(not yet visited)",
 					uiPoint + Point(-70., 28.),
 					isAccessible ? dim : medium);
 				if(commodity == SHOW_VISITED)
--- a/source/MapDetailPanel.cpp
+++ b/source/MapDetailPanel.cpp
@@ -255,7 +255,7 @@ bool MapDetailPanel::Click(int x, int y, int clicks)
 				selectedPlanet = it.first;
 			}
 		}
-		if(selectedPlanet && player.Flagship())
+		if(selectedPlanet && player.Flagship() && selectedPlanet->IsAccessible(player.Flagship()))
 			player.SetTravelDestination(selectedPlanet);
 		
 		return true;
--- a/source/MapDetailPanel.cpp
+++ b/source/MapDetailPanel.cpp
@@ -630,7 +630,7 @@ void MapDetailPanel::DrawOrbits()
 			continue;
 		
 		Point pos = orbitCenter + object.Position() * scale;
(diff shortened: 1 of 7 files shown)


Hadron1776:
Because being able to select inaccessible planets implies one has visited them, I just put the indication where it indicates whether a planet has been visited.

source/MapDetailPanel.cpp
```diff
@@ -442,7 +442,7 @@ void MapDetailPanel::DrawInfo()
 				// The same "planet" may appear multiple times in one system,
 				// providing multiple landing and departure points (e.g. ringworlds).
 				const Planet *planet = object.GetPlanet();
-				if(planet->IsWormhole() || !planet->IsAccessible(player.Flagship()) || shown.count(planet))
+				if(planet->IsWormhole() || shown.count(planet))
```
tehhowch: You probably need to gate accessibility on having visited it here - otherwise the player is told "hey, look, planet! (but you can't go there!)" before visiting it.

source/MapDetailPanel.cpp
```diff
@@ -480,9 +480,10 @@ void MapDetailPanel::DrawInfo()
 						10., 10., 0., medium);
 				
 				bool hasVisited = player.HasVisited(planet);
-				font.Draw(hasVisited ? "(has been visited)" : "(not yet visited)",
+				bool isAccessible = planet->IsAccessible(player.Flagship());
```
tehhowch: With the above change (only showing inaccessible planets in DrawInfo if you've visited them) this is fine, since you're guaranteed to have visited it.
If the above change is not done (DrawInfo for inaccessible planets without having visited) then this should probably be conditional on having visited the planet: ` hasVisited && planet->IsAccessible(player.Flagship());`

source/MapDetailPanel.cpp
```diff
@@ -480,9 +480,10 @@ void MapDetailPanel::DrawInfo()
...
+				font.Draw(!isAccessible ? "(Inaccessible)" : hasVisited ? "(has been visited)" : "(not yet visited)",
```
tehhowch: `I` -> `i` to match case with the other phrases.

source/MapDetailPanel.cpp
Hadron1776: I'm not sure I should draw info for inaccessible planets the player hasn't visited, considering you can't select them.

endless-sky:
The trouble with this is that the only way the map communicates planet accessibility is by whether the planet is shown as a planet in the orbits or not. So now, you're making a planet look accessible when it's not, just for the sake of letting the player view descriptions. That doesn't seem like a worthwhile tradeoff.

Tadrix:
@endless-sky AFAIK the map shows planets you can land on in blue, dominated planets in green, hostile planets in orange (they turn blue if bribed or if player has clearance) and inaccessible planets in grey.

Descriptions are shown for visited planets.

Therefore it seems consequential that currently inaccessible gas giants would have visible descriptions but be shown in grey color.

Amazinite:
>Therefore it seems consequential that currently inaccessible gas giants would have visible descriptions but be shown in grey color.

Then what is the distinction between a planet that can never be landed on and a planet that is only inaccessible because you don't have the proper attribute?

Tadrix:
Agreed. Then how about coloring visited planets which are currently inaccessible in bright grey, instead of medium grey (which indicates permanently inaccessible planets)?

source/StellarObject.h
```diff
@@ -14,6 +14,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #define STELLAR_OBJECT_H_
 
 #include "Body.h"
+#include "PlayerInfo.h"
```
tehhowch: This should be a forward declaration and not an include, since the class is simply "mentioned" in its interface, rather than kept in an object. The implementation, then, would need the `#include`, since it is accessing properties of the class.

(`#include`ing it in the header means whenever the PlayerInfo interface changes, any compilation unit that includes the StellarObject header will also need to be recompiled, while having it as a forward declaration would not, since the StellarObject header is basically telling units that include it, "I consume this PlayerInfo thing, but you don't need to worry about how I do that")

warp-core:
@Hadron1776 could you bring this up to date? I'd like to test it but it's pretty far behind, even using the old CodeBlocks project structure.

warp-core (commented):
I have no issues with this. Just needs to be brought up to date and the conflicts resolved.

Amazinite:

TomGoodIdea:
Superseded by #11318

PR #3564, opened 2018-01-16
improved Remnant ramscoop

Nescio0:
This proposal improves the Remnant's ramscoop by 60% <s>50% and increases its cost by 100%</s>. It deserves an improvement because currently it's hardly better than the ordinary ramscoop. Besides, in #3557 it was claimed it's easy to unlock Wanderer outfits long before getting a Remnant licence. In comparison:

``` 
                              credits, mass, ramscoop, efficiency
ordinary Ramscoop           : 0.060m ,  10 ,    1.0  , 0.1000
current Emergency Ramscoop  : 0.072m ,   4 ,    0.5  , 0.1250
proposed Emergency Ramscoop : 0.072m ,   4 ,    0.8  , 0.2000
campaign Catalytic Ramscoop : 0.320m ,  16 ,    7.0  , 0.4375
Wanderer Ramscoop           : 0.460m ,   7 ,    4.0  , 0.5714
```

### DIFF
--- a/data/remnant outfits.txt
+++ b/data/remnant outfits.txt
@@ -238,11 +238,11 @@ outfit "Emergency Ramscoop"
 	category "Systems"
 	licenses
 		Remnant
-	cost 72000
+	cost 144000
 	thumbnail "outfit/emergency ramscoop"
 	"mass" 4
 	"outfit space" -4
-	"ramscoop" 0.5
+	"ramscoop" 0.75
 	description "The Ember Waste is the sort of place where only a very foolish captain would travel without some sort of device for replenishing hyperspace fuel. This ramscoop is not particularly powerful, but it allows a ship to explore without fear of getting stranded."
 
 
--- a/data/remnant outfits.txt
+++ b/data/remnant outfits.txt
@@ -238,11 +238,11 @@ outfit "Emergency Ramscoop"
 	category "Systems"
 	licenses
 		Remnant
-	cost 144000
+	cost 72000
 	thumbnail "outfit/emergency ramscoop"
 	"mass" 4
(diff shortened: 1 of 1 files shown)


KiLEdEnNis:
the Remnant would need ramscoop technology way harder than normal humans, due to the nature of the Ember Waste, this improvement seems fine to me; still balanced and not particularly powerful, but at least a decent choice for a small ramscoop

Amazinite:
I'm indifferent to such a small increase to the ramscoop ability of this already pretty bad outfit (although I'd probably increase it to 0.8 instead of 0.75), but I'd say that the price increase is unnecessary. The Emergency Ramscoop is expensive as is, and doubling the price is far too much.

Outfit | Cost per Ramscoop
-----|------
Ramscoop | 60k
Catalytic Ramscoop | 45.7k
Wanderer Ramscoop | 115k
Current Emergency Ramscoop | 144k
Proposed Emergency Ramscoop | 192k
Proposed Emergency Ramscoop w/o the price increase | 96k

solardawning:
Eh. This doesn't seem very necessary to me.
The emergency ramscoop is already 25% better (tonnage to output) than the basic Ramscoop, and while it does get quickly outclassed for efficiency, it will always remain best at one specific thing- it's the smallest ramscoop you can install.

If you want to use the least outfit space possible to make sure you have at least some ramscoop attribute, this is the way to do it. It's best in class for that.
It doesn't really need to be competitive with the larger ones.

endless-sky:
I really don't see the point of this. The Remnant ramscoop is already the smallest existing ramscoop, and is useful for that reason.

PR #3567, opened 2018-01-16
Show visited mission systems

tehhowch:
If a mission defines a waypoint, this waypoint is no longer visible once it has been visited, even if a task in said waypoint is incomplete (e.g. "destroy the Sestor drones"). Missions began tracking visited waypoints / stopovers in #3208, but they were not used outside of granting clearance to land without being fined.

This PR creates a faint ring around any visited waypoint of an active mission, when that mission is selected. In this screenshot I am in a waypoint system, and have the mission selected, which highlights the missions waypoints (both visited and unvisited).
image

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -409,16 +409,30 @@ const Planet *Mission::Destination() const
 
 
 
-set<const System *> Mission::Waypoints() const
+const set<const System *> &Mission::Waypoints() const
 {
 	return waypoints;
 }
 
 
 
-set<const Planet *> Mission::Stopovers() const
+const set<const System *> &Mission::VisitedWaypoints() const
+{
+	return visitedWaypoints;
+}
+
+
+
+const set<const Planet *> &Mission::Stopovers() const
 {
 	return stopovers;
+}
+
+
+
+const set<const Planet *> &Mission::VisitedStopovers() const
+{
+	return visitedStopovers;
 }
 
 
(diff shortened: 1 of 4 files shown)


Tadrix:
Show "Has unfinished requirements" markers only where they hadn't been already visited.
Draw a circle around all waypoints for a selected mission. Lighter blue (like on screenshot, similar color to active job markers) for unvisited, darker color (similar to "waypoint you must visit" markers) for visited.

endless-sky:
Given that the usual meaning of a waypoint is "a system you have to visit to complete the mission," and therefore 95% of the time it would be preferable not to clutter the player's map with waypoints that have already been satisfied, I don't think this is useful. The Sestor missions are the only example I can think of of a case where you would want the marker to stick around, and in that case they still aren't helpful because their status shows which systems have been visited, not which ones still have drones in them.

tehhowch:
An alternate UI scheme with this information would be to draw darker grey (instead of blue) circles for any visited waypoints of the selected system. This would mean no need for an updated sprite, waypoint color, 6th pointer, etc.

I'll swap this to that scheme and provide a screenshot. If it's still no good then, I'll scrap it.

tehhowch:
<details><summary>Using "faint":</summary>


image
</details>
<details><summary>Using "dim":</summary>


image
</details>
<details><summary>Using "dim", with 50 segments:</summary>


image
</details>
Thoughts welcomed

Tadrix:
I'd say "faint" is the one. It's there, but not in the way of anything.

PR #3569, opened 2018-01-17
Don't try to match the top when scrolling down a big selected item.

flaviojs:
If the height of the selected item was a bit bigger than the main height, the auto-scroll would go up and down repeatedly and never stop.

### DIFF
--- a/source/ShopPanel.cpp
+++ b/source/ShopPanel.cpp
@@ -74,7 +74,7 @@ void ShopPanel::Step()
 		if(selectedBottomY > mainBottomY)
 			DoScroll(max(-30., mainBottomY - selectedBottomY));
 		// Scroll down until the bottoms or the tops match.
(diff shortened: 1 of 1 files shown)


tehhowch:
Can you give an example outfit for which this occurs?
(I've definitely had an issue where selecting an item caused the screen to rapidly pulse the sale catalog from top to bottom of the screen many times a second, but I can't seem to reproduce this on `master`)

flaviojs:
It's hard to reproduce since it depends on the initial position, the window size and the size of icon + details.

This will probably get you there:
 - use something that has an icon + details size a bit over window height - 40, ideally with a difference below 30 pixels (the max scroll amount)
 - force a scroll situation when you select the item by making sure the icon is partially hidden by the top of the window or in a position where the details will be partially hidden by the bottom of the window
 - fine tune it by adjusting the window height bit by bit

PR #3571, opened 2018-01-17
Adding inhibitor.wav

solardawning:
Adding an audio file for the Inhibitor Cannon, to replace its use of particle.wav

With the rapid rate of fire and its typical use in large numbers, the particle sound is jarring on the Inhibitor Cannon.

The intent of this replacement sound is not only to give it a unique audio, but also one that is less obtrusive when used rapid fire and in large numbers.

This audio is derived from the Korath Detainer (the supposed source of Remnant Inhibitor technology), but sped up and pitch altered to fit a smaller weapon.

For easy comparison: 
Here is the current Inhibitor Cannon, which shares audio with Particle Cannon:
https://webmshare.com/mP44X (I find this one very jarring when used by lots of ships)

And here is the proposed unique Inhibitor Cannon sound:
https://webmshare.com/q544g (Milder, based off Detainer audio)
https://webmshare.com/6j3nB (in Stream Fire mode, plus some engine noise for volume comparison in the last half)


tehhowch:
(you should post a comparison like @Pointedstick did for the blasters, so that it can be previewed easily)

solardawning:
Working on it. I've literally never recorded a video in anything before, but I'm figuring out how to as we speak.

tehhowch:
Can you also add a preview of the sound in streaming fire mode? (Just remove the `cluster` attribute from the cannon's definition to test this easily)
While the outfit is `cluster`, due to its range and speed you can get it into a stream fire equivalent if you rotate past your target under the right circumstances.

Also, do you have the same volume settings for both sounds? The particle is significantly louder than your proposed sound.

solardawning:
Yes, same volume for both. That's part of why I dislike the Particle Cannon sound. It's like nails on a chalkboard to me.

Here's a video of the proposed sound in stream fire mode, plus some engine noise for the latter half to give a volume scale comparison.
https://webmshare.com/6j3nB

tehhowch (approved):
sgtm

KiLEdEnNis:
in that 12-second video (stream fire) I notice some sort of mini pause after every second of firing, which makes it a little bit obvious the sound gets replayed
apart from that I like it

endless-sky:
What's the source and license information for the sound effect?

solardawning:
@endless-sky I mutated the Korath Detainer sound into it myself using Audacity (free, open source audio file editor).

PR #3574, opened 2018-01-17
Moved M. Fury to persons

Nescio0:
<s>Although I understand the Marauder Fury's need to defend itself, I think having it equipped with three Quarg Intrusion Countermeasures is hard to explain, lore-wise. This proposal replaces them with ordinary human laser rifles. There is no space (0/240) for security stations.</s>

Because it's a unique person ship, this proposal moves the M. Fury from the `marauders.txt` file to the `persons.txt` file to avoid unnecessary confusion.

### DIFF
--- a/data/marauders.txt
+++ b/data/marauders.txt
@@ -454,62 +454,6 @@ ship "Marauder Firebird" "Marauder Firebird (Weapons)"
 
 
 
-ship "Marauder Fury"
-	plural "Marauder Furies"
-	sprite "ship/mfury"
-	attributes
-		category "Interceptor"
-		"cost" 750000
-		"shields" 3000
-		"hull" 600
-		"required crew" 1
-		"bunks" 3
-		"mass" 100
-		"drag" 1.9
-		"heat dissipation" .85
-		"fuel capacity" 500
-		"cargo space" 15
-		"outfit space" 240
-		"weapon capacity" 80
-		"engine capacity" 100
-		"ramscoop" 2
-		weapon
-			"blast radius" 36
-			"shield damage" 360
-			"hull damage" 180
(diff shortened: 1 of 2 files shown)


Amazinite:
The Marauder Fury is a person ship, so I don't see the need to remove its intrusion countermeasures. This also makes the Marauder Fury considerably easier to capture, and it's intended to be a difficult capture.

kikotheexile:
The mfury in general, much like MZ's Kestrel, is a special ship that can freely break a few of the rules of the lore.

Nescio0:
It is the only Marauder which has Quarg technology. The M. Arrow (crew 2/5) has only laser rifles, the M. Raven has no hand-to-hand weapons at all.

Amazinite:
It is also the only Marauder that is a person ship.

Nescio0:
> It is also the only Marauder that is a person ship.

Why is it here and not in the `persons.txt` file then?

Amazinite:
That's a question for Local or Wrzl, who were the ones who made the Marauder ships.

Startome:
Truthfully being a person ship, I think it should be harder to capture or uncapturable. That said I don't think that idea will find much support

tehhowch:
I would place the ship definition below the person for whom it is used, rather than at the end of the file.

Wrzlprnft:
i think it's just a remnant of how local made that stuff: it was probably a normal marauder ship before he had the person idea.

Startome:
Yeah I really haven't had any problem right now with it set up as is. Any raider or baccy can easily capp it, it isn't too hard. My shield beetle could even capp it.

Nice to see you wrzl hope you are well.

PR #3575, opened 2018-01-17
Update needed declarations

tehhowch:
Double `using namespace std;`
Forward declaration of `class Sprite` in the .cpp when the .cpp already has `#include "Sprite.h"`


tehhowch:
Per https://github.com/endless-sky/endless-sky/pull/3312#issuecomment-357767874 I've added `#include "gl_header.h"` for uses of `glClear()`

PR #3576, opened 2018-01-17
Show location of player escorts on the map

tehhowch:
Another part of #3070 (and #2678 ) was that the player should have a way to know where their escorts are. This has also come up in other issues, where players have trouble locating systems (to go get stranded escorts, etc).

Rather than add a pointer or a thin ring around escort systems (which are generally used for missions), I've opted to fill in the central ring of a system with a small pip if it has at least one player ship. If the only escorts are parked, the pip will be dimmer than if there are unparked ships there.
This is an opt-out preference.

<details><summary>Preference panel</summary>

image
</details>
<details><summary>Escorts in Pherkad, Shaula, and Lesath</summary>

image
</details>

Possible alterations:
 - ~~Differentiate between systems with only parked escorts, and those with at least one unparked escort~~ Done
 - ~~Show the pip even for the player's system~~ Done
 - Choose a better preference name
 - ~~Make it opt-out or not optional~~ Done
 - Future Work: Tooltips

Feedback requested

### DIFF
--- a/source/MapPanel.cpp
+++ b/source/MapPanel.cpp
@@ -49,6 +49,26 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 using namespace std;
 
+namespace {
+	// Log how many player ships are in a given system, regardless if they are parked or carried.
+	void TallyEscorts(const vector<shared_ptr<Ship>> &escorts, map<const System *, int> &locations)
+	{
+		locations.clear();
+		for(const auto &ship : escorts)
+		{
+			if(ship->IsDestroyed())
+				continue;
+			if(ship->GetSystem())
+				++locations[ship->GetSystem()];
+			else if(ship->CanBeCarried() && ship->GetParent() && ship->GetParent()->GetSystem())
+				++locations[ship->GetParent()->GetSystem()];
+		}
+	}
+	
+	const Color black(0., 1.);
+	const Color red(1., 0., 0., 1.);
+}
+
 const double MapPanel::OUTER = 6.;
 const double MapPanel::INNER = 3.5;
 const double MapPanel::LINK_WIDTH = 1.2;
@@ -69,6 +89,9 @@ MapPanel::MapPanel(PlayerInfo &player, int commodity, const System *special)
 	// Recalculate the fog each time the map is opened, just in case the player
 	// bought a map since the last time they viewed the map.
 	FogShader::Redraw();
+	// Recalculate escort positions every time the map is opened, as they may
+	// be changing systems even if the player does not.
+	TallyEscorts(player.Ships(), escortSystems);
 	
 	if(selectedSystem)
 		center = Point(0., 0.) - selectedSystem->Position();
@@ -97,6 +120,7 @@ void MapPanel::Draw()
 	++step;
 	DrawWormholes();
 	DrawTravelPlan();
+	DrawEscorts();
 	DrawLinks();
 	DrawSystems();
 	DrawNames();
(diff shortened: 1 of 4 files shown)


Tadrix:
Looks good.

> - Differentiate between systems with only parked escorts, and those with at least one unparked escort

I'd opt not to mark systems with parked escorts. Or, if mark them, do it very subtly — probably not worth it. The player can have escorts parked in many places, showing them all would clutter the map.

> - Show the pip even for the player's system

If do this at all, it should be different than other pips. (Different color, perhaps.)

> - Change the drawn color based on the number of escorts in a given system

If do this at all, they shouldn't differ much. I'd like a tooltip or a side panel better (showing how much there are, and, if not too many, their names).

> - Choose a better preference name

`Show escort locations on map` (?)

> - Make it opt-out

Or give the ability to opt out in the `preferences.txt`.
This last one stems from my opinion that most of new UI changes should be opt-out in the `preferences.txt` if feasible to give people an opportunity to test them and decide if they like them better.
Admittedly, as it is now, if pips don't show for parked escorts, I don't see why someone would opt out. Seems like an upgrade to me.

tehhowch:
- I'd do a tooltip before I made a key for the escort pips. Having a key is basically requiring a coloring scheme, and given the general lack of diversity (see relevant images of #3070 as it developed - the map is nearly all uncolored), this would be a very boring color scheme for almost every system.
My first take used discs and made them larger based on how many escorts were in a system (hence the map system-int structure) so I've already got the count available, would just need to add hover support.

 - All preferences are controllable via preferences.txt, this is just opt-in currently. I'll wait for more feedback before choosing to default-on or always-on-with-toggleable-parked-inclusion.

Tadrix:
One more thought: If number of ships is so large it is better not to display their names in the side panel or a tooltip (say, >10), display ship types and count instead?

For example, tooltip or sidebar if there are 3 ships:
Ships: 3
Aurora
Galatea
Andromeda

Tooltip or sidebar if there are 38 ships:
Ships: 27
6 transports
2 heavy freighters
2 interceptors
12 light warhips
2 medium warships
11 heavy warships 
3 fighters

(no mention of light warships or drones because in this example there aren't any)

endless-sky:
That looks a lot better than the ideas I tried for showing where escorts are present!

Here's what I would recommend:
1. **Making this opt-out or always-on behavior rather than opt-in.** (Because it's subtle enough that it's unlikely to annoy people.)
2. **Showing a pip for the player's flagship, too.** (This would mean that the player gets used to seeing that pip in the "current system" when they have only one ship, so when they have escorts it will be immediately intuitive what the multiple pips mean. That wouldn't be the case if the pips never show up until you have multiple ships.)
3. **A dim pip for systems with parked escorts and no unparked ones.** (Dim is a universal UI indication of inactivity, so again I think it'd be pretty intuitive.)

I'd definitely vote against the pips being colored. There's already enough color on the map that one more colored item would just be distracting.

tehhowch:
I have to thank @MCOfficer for suggesting the little gap between the pip and the system ring (I had originally had them filled), among others who gave initial feedback on Discord.

Thoughts on a tooltip? Definitely best left for a separate PR, but I could imagine it turning into an onhover that says "you have W escorts in this system, X are holding position, Y are parked, and Z are routing to you"
That'd obviously require tie-ins to accessing the AI::Orders class member, so a work in progress for sure.

endless-sky:
It might be good to start out with tooltips that just say one of the following:

```
N escort(s) present
N active escort(s) and M parked escort(s) present
M parked escort(s) present
```

That'd get you the most important information without needing to introduce too much complexity.

tehhowch:
Parked escorts in Shaula and Pherkad:
image

Also added checks that the player knows about the systems the escorts are in, since they use the best route and not the best known route.

Tadrix:
Do you have an image how they would look if they were faint instead of dim?

tehhowch:
The uninhabited color is `.2 .2 .2 0.`, while unexplored is `.1 .1 .1 0.`
Default rgba for `"faint"` is the same as unexplored, and `"dim"` is slightly brighter (.25 .25 .25 0.) than uninhabited.

Tadrix:
> I think I prefer "dim", because it's basically impossible to see the pip in Pherkad for the escorts parked there.

Agreed, dim is better.

tehhowch:
If this is desired as non-optional, then the DrawEscorts method should be folded into DrawSystems (it could anyway, and would just require checking a boolean derived from the preferences state for each system). I'd probably switch the preference to then be `"Show pips for parked escorts"`
Barring that change, (reserving tooltip implementation for a different PR), I think this is set.

PR #3314, opened 2017-11-29
3 new Hai ships

Brick63:
adding 3 new Hai ships to give them a more diverse feel
Centipede: a luxury Hai transport 
Geocoris: a Heavy freighter to compete with human ones
Grasshopper: a Hai interceptor
Stats by @Amazinite  
thanks to @warp-core for fixing the hard-point order on the Geocrosis
thanks to @Pointedstick for the better flowing and sounding descriptions on the centipede and geocoris
Ship assets:
https://drive.google.com/drive/folders/1zG1I87cKfDcxG1xR4cTZBZTwdMe6Bvgh?usp=sharing

PR #3577, opened 2018-01-17
Alphabetized Pug ships

Nescio0:
Cf. `coalition ships.txt`, `hai ships.txt`, `ships.txt`.

### DIFF
--- a/data/pug.txt
+++ b/data/pug.txt
@@ -310,51 +310,63 @@ outfit "Pug Biodefenses"
 	"capture defense" 250
 	"unplunderable" 1
 	description "This defense system is designed to release a suite of toxins that are lethal to human beings (and probably to some other species, as well) but which have no effect at all on the Pug."
+	
 
-
-ship "Pug Zibruka"
-	sprite "ship/pug zibruka"
+ship "Pug Arfecta"
+	sprite "ship/pug arfecta"
 	attributes
-		category "Light Warship"
-		"cost" 1900000
-		"shields" 4000
-		"hull" 1000
-		"required crew" 4
-		"bunks" 6
-		"mass" 240
-		"drag" 3.1
+		category "Medium Warship"
+		"cost" 90000000
+		"shields" 98000
+		"hull" 80000
+		"required crew" 36
+		"bunks" 46
+		"mass" 670
+		"drag" 7.5
 		"heat dissipation" .8
-		"fuel capacity" 600
-		"cargo space" 20
-		"outfit space" 180
-		"weapon capacity" 68
-		"engine capacity" 76
-		"energy generation" 10
-		"shield generation" 6
-		"shield energy" 6
-		"heat generation" 20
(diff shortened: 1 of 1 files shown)


tehhowch:
Previously they were in strength / tier order

Tadrix:
No need to change this, IMO. "strength / tier order" is more helpful. Perhaps if we had so many Pug ships as humans have now…

KiLEdEnNis:
for the humans it's a big mess, and alphabetized because it's from various shipyards
as for aliens, they're mostly from small to large, split in groups if needed (like Wanderer, Korath (grouped in base, Sestor & Mereti), Coalition (grouped per species, & Heliarch))
if anything, human ships should be grouped per shipyard, but even then ya have warships & civilian ships
for my own mod I do warships, freighters & transports from small to large & I split them up in those groups as well, which makes more sense to work with than alphabetized

the only upside for an alphabetized lay-out is so one can quickly find a ship, but ya have Ctrl + F for that too

Nescio0:
So I've looked at the files again. I've seen no list sorted purely by size.
Humans, Marauders, Variants, Hai, Quarg, and Coalition ships are always sorted purely alphabetically.
Drak have just one ship.
Korath ships are sorted by faction. Korath by first encounter, then Sestor by decreasing strength, then Mereti alphabetically (i.e. increasing strength).
Pug ships are sorted by strength.
Remnant ships are unsorted.
Wanderer ships are sorted by introduction date.

I for one thinks this is internally inconsistent, so I'd prefer one way to sort ships. The alphabet works.

Tadrix:
With humans there's too many things you could sort them by: Shipyards; Civilian (Transport, Freighter) and Warships; sizes (Interceptor, Light, Medium, Heavy, Fighter, Drone). Alphabetized order at least ensures you'd know where to find the ship you're interested in.

This problem is absent with aliens, because they have only one shipyard, so to speak, and their ships can be ordered in whatever order seems logical. Those who maintain those lists are used to such order, and this is not something ordinary player would see, so, unless situation changes, if it ain't broke, don't fix it.

Nescio0:
This problem is absent with aliens, because they have only one shipyard, so to speak, and their ships can be ordered in whatever order seems logical. People who maintain ship lists ar most likely used to such an order (otherwise they would reorder them), and this is not something ordinary player would see, so, unless situation changes, if it ain't broke, don't fix it.

If only one person worked on a file and no new content was added, the order won't matter much. However, because the game is constantly being expanded, having one uniform standard is helpful. Have a look at the Hai. Originally they had four ships. Later someone else created four other ships. A third person has proposed another three: #3314 
Although it's not terribly important, I think it's wise to have a uniform sorting standard for all content files.

endless-sky:
This just feels like "churn" to me - changing things for the sake of changing them. I'd much prefer that you focus your (and my) time and energy on making changes that either fix definite bugs, or will make the game more fun to play.

PR #3585, opened 2018-01-19
Improvements to the Cloaking AI

tehhowch:
- Cloaked ships can decloak to repair their disabled / stranded allies (and be asked to do so by ships of their own government)
 - Extended the "don't cloak if you'll be stranded" check to ensure that the ship can actually even achieve cloak
 - Switched from absolute shield/hull health to using the "health until disabled" metric, using the same threshold as retreating fighters.
 - Cloaking for repairs does not occur unless you can actually repair
 - Added a rudimentary "retreat" behavior, so that ships that choose to cloak to repair will actually increase the distance from their target (rather than continue to move as though they are not cloaked)
 - Cloaked AI escorts will cloak when their friendly parent cloaks, and then rendezvous with them (thus making it more likely that they focus on the same hostile target when they decloak).

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -402,8 +402,12 @@ void AI::Step(const PlayerInfo &player)
 				command |= Command::CLOAK;
 		}
 		// Cloak if the AI considers it appropriate.
-		else
-			DoCloak(*it, command);
+		else if(DoCloak(*it, command))
+		{
+			// The ship chose to retreat from its target, e.g. to repair.
+			it->SetCommands(command);
+			continue;
+		}
 		
 		shared_ptr<Ship> parent = it->GetParent();
 		if(parent && parent->IsDestroyed())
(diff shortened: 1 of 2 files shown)


source/AI.cpp
```diff
@@ -811,7 +815,8 @@ void AI::AskForHelp(Ship &ship, bool &isStranded, const Ship *flagship)
 bool AI::CanHelp(const Ship &ship, const Ship &helper, const bool needsFuel)
 {
 	// Fighters, drones, and disabled / absent ships can't offer assistance.
-	if(helper.CanBeCarried() || helper.GetSystem() != ship.GetSystem() || helper.Cloaking() == 1.
+	if(helper.CanBeCarried() || helper.GetSystem() != ship.GetSystem()
+			|| (helper.Cloaking() == 1. && helper.GetGovernment() != ship.GetGovernment())
```
endless-sky: Why this change? It seems reasonable to let, for example, a Merchant ask for fuel from the Navy.

tehhowch: That should still be possible - rather than be unable to ask any cloaked ship for help (current behavior), I make it so that a cloaked ship can be asked for help if that cloaked ship is of the same government

PR #3587, opened 2018-01-19
Display a map tooltip for escorts in a given system

tehhowch:
Refs https://github.com/endless-sky/endless-sky/pull/3576#issuecomment-358497939

Current UI:
<details><summary>Active and parked</summary>

image
</details>
<details><summary>Flagship only</summary>

image
</details><details><summary>Active only</summary>

image
</details><details><summary>Parked only</summary>

image
</details>

~~To-do~~ Done
 - [x] Change escortSystems to track count and type, not just type (e.g. `pair<int, int>` instead of `bool`)
 - [x] Extended info tooltips per refernence
 - [x] Smarter fillshader / tooltip position (shifting near edge)
 - [x] Support "passive" hovering rather than only active hovering.
 - [x] Obey preference for pip display (no pips -> no pip tooltips)

### DIFF
--- a/source/MapPanel.cpp
+++ b/source/MapPanel.cpp
@@ -18,6 +18,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "FogShader.h"
 #include "Font.h"
 #include "FontSet.h"
+#include "Format.h"
 #include "Galaxy.h"
 #include "GameData.h"
 #include "Government.h"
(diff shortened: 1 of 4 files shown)


Tadrix:
It could probably be better if the tooltip didn't obscure its system (so it's upper left corner not right at system's center), and appeared below the mouse cursor (like in other programs). Any chances of this happening?

tehhowch:
The tooltip does appear under the mouse pointer, z-axis-wise. Screenshots don't capture my pointer, but it is within the system ring in all these.

Or are you meaning the text should offset more in the y direction? I think other tooltips in the game appear in the same manner as this one.

Tadrix:
I had in mind offsetting tooltip's block (rectangle encompassing it) downward, so that it would appear below the pointer, like in most other programs.
pointer
(The mouse pointer is approximately at the red marker.)

> I think other tooltips in the game appear in the same manner as this one.

That's right. And it looks weird when the pointer is drawn on the tooltip's block.

PR #3589, opened 2018-01-19
Largest Korath drones get more large heat shunts

Nescio0:
The Kar Ik Vot 349, Model 256, and Model 512 produce a lot of heat but also have some unused outfit capacity. This proposal gives them one more large heat shunt (size 24) to make them less likely to be overheated by heat-producing weapons. There is enough room for this, therefore mass and outfit size can be left unchanged.

The Kor Mereti Model 256 and Model 512 before:
screenshot from 2018-01-19 18-34-32

The Kor Mereti Model 256 and Model 512 after:
screenshot from 2018-01-19 18-43-42

The Kor Sestor Kar Ik Vot 349 currently has 4 large + 2 small heat shunts, resulting into:
screenshot from 2018-01-19 18-30-40

This proposal replaces those with 5 large + 0 small heat shunts, resulting into:
screenshot from 2018-01-19 18-31-55

### DIFF
--- a/data/korath ships.txt
+++ b/data/korath ships.txt
@@ -319,8 +319,7 @@ ship "Kar Ik Vot 349"
 		"Triple Plasma Core"
 		"Plasma Core"
 		"Systems Core (Large)"
-		"Large Heat Shunt" 4
-		"Small Heat Shunt" 2
+		"Large Heat Shunt" 5
 		"Control Transceiver"
 		
 		"Thruster (Planetary Class)"
@@ -1150,7 +1149,7 @@ ship "Model 256"
 		
 		"Triple Plasma Core"
 		"Systems Core (Large)"
-		"Large Heat Shunt"
+		"Large Heat Shunt" 2
 		"Reasoning Node" 3
 		
 		"Thruster (Planetary Class)"
@@ -1208,7 +1207,7 @@ ship "Model 512"
 		"Triple Plasma Core"
 		"Plasma Core"
 		"Systems Core (Large)"
-		"Large Heat Shunt"
+		"Large Heat Shunt" 2
 		"Small Heat Shunt" 2
 		"Reasoning Node" 3
 		


ghoulavenger:
The problem is that your best possible weapon at that point in the Wanderer campaign is a heat based weapon. You're not just using extra outfit capacity you're changing balance. I'm not sure that is such a good thing to do without testing some of the harder fights against the Kor.

Amazinite:
Plenty of ships have a considerable amount of left over outfit space;
* Blackbird (48)
* Bounder (24)
* Aphid (30)
* Argosy (43)
* Clipper (38)
* Hauler I, II, and III (22)
* Behemoth (40)
* Bulk Freighter (38)
* Deep River (41)
* Water Bug (33)

And that's just looking at non-variant transports and freighters with >20 unused outfit space, leaving out every single Coalition ship since they don't use weapons. (Glancing at heavy warships, we have the Carrier, Cruiser, Falcon, and more with lots of empty space.) For this reason, and what @ghoulavenger has mentioned, I see zero need to do this. You'll also be hard-pressed to find a stock ship that's outfitted in an optimal way (or else why would you outfit your ship in your own way if the default outfits are already optimal?), so changing the Kar Ik Vot's 4/2 large/small cooling to 5/0 is unnecessary as well.

tl;dr churn

Nescio0:
> Plenty of ships have a considerable amount of left over outfit space;

The difference is that Korath ships are not for sale and that their drones are supposed to be dedicated warmachines.

> The problem is that your best possible weapon at that point in the Wanderer campaign is a heat based weapon. You're not just using extra outfit capacity you're changing balance. I'm not sure that is such a good thing to do without testing some of the harder fights against the Kor.

That is a valid concern. I'm not sure how much difference one large heat shunt will make (a default Hurricane can inflict 7260 heat damage per second), but I haven't play-tested the campaign.

KiLEdEnNis:
I offered to retry the balance of the full Wanderer story, but from what I understood some changes had to be made first
I don't remember them by head but 1 of them was after bombing Zenith the Northern drone battles were straight-up impossible in a single ship, and it's easy mode with a cloaking device

once those balances are made & the missions adjusted to be do-able & hard in 1 ship, I could try it after these changes, but for now we shouldn't touch the balance imo

Pointedstick:
Cons:
- Increases the difficulty of the Wanderer storyline, which is already very challenging, especially in a single ship

Pros:
- ???

PR #2009, opened 2017-01-14
Expanded Swizzle Choices

kozbot:
More swizzle choices based on work done by @Amazinite in #1779

Issue #1779, opened 2016-10-26
Expanded Swizzle Choices


Amazinite:
It would be nice to have ~~custom~~ more swizzles beyond the current 7, expanding the number of different looking governments of the same race that we can create, or to just have some human governments like Neutral or Independent have a color other than that of the Republic or Pirates.
e.g. 
(rgr -> rgb)
![screenshot_158](https://cloud.githubusercontent.com/assets/17688683/19747411/8f90aaa8-9ba9-11e6-9736-b1d36d2b6287.png)
(rrb ->)
![screenshot_1](https://cloud.githubusercontent.com/assets/17688683/19747418/97312c24-9ba9-11e6-81d3-d4a940a375b2.png)
(gbg ->)
![screenshot_2](https://cloud.githubusercontent.com/assets/17688683/19747420/9e07dbd8-9ba9-11e6-9b62-361b967a902a.png)
I like this one.
(grr ->)
![screenshot_3](https://cloud.githubusercontent.com/assets/17688683/19747527/24733b2c-9baa-11e6-8567-5b15d52f9668.png)

The number of possible swizzles would go from 7 to 27 :)

Amazinite:
Swizzle choices would go from 7 to 27. :)

endless-sky:
Most ships in-game are designed to use only the range of hues between red and yellow, or slightly beyond that:

![tricolor](https://cloud.githubusercontent.com/assets/11450262/19748604/d97a5504-9bb0-11e6-8c88-03ea4c79140e.png)

If you apply a swizzle that uses the same input color channel for two of the outputs, you end up with something like this:

![bicolor](https://cloud.githubusercontent.com/assets/11450262/19748609/f191cbc2-9bb0-11e6-8b35-0408b317d67d.png)

Everything ends up as different shades of a single hue. And, that hue will always be a primary or secondary color, which in my opinion tends to look tacky because that's what graphics looked like back when we only had 16 colors:

![lzr23](https://cloud.githubusercontent.com/assets/11450262/19748703/7ab0b06c-9bb1-11e6-8331-7cff1f199c31.png)

Admiral-Danforth:
The blue matches independent perfectly.

Pointedstick:
All of this technical detail obscures the fact that it's extremely odd that you can't just have arbitrary colors for the trim bits in the first place.

Amazinite:
[...]

PR #3591, opened 2018-01-20
Add more swizzle choices

Disiuze:
Based on #1779 and #2009, this PR adds 18 new swizzle options.
 
`swizzle 27`
image

`swizzle 20`
image

### DIFF
--- a/source/SpriteShader.cpp
+++ b/source/SpriteShader.cpp
@@ -36,34 +36,34 @@ namespace {
 	GLuint vbo;
 
 	const vector<vector<GLint>> SWIZZLE = {
-		{GL_RED, GL_GREEN, GL_BLUE, GL_ALPHA},		 // red + yellow markings (republic)
-		{GL_RED, GL_BLUE, GL_GREEN, GL_ALPHA},		 // red + magenta markings
-		{GL_GREEN, GL_RED, GL_BLUE, GL_ALPHA},		 // green + yellow (freeholders)
-		{GL_BLUE, GL_RED, GL_GREEN, GL_ALPHA},		 // green + cyan
-		{GL_GREEN, GL_BLUE, GL_RED, GL_ALPHA},		 // blue + magenta (syndicate)
-		{GL_BLUE, GL_GREEN, GL_RED, GL_ALPHA},		 // blue + cyan (merchant)
-		{GL_GREEN, GL_BLUE, GL_BLUE, GL_ALPHA},		 // red and black (pirate)
-		{GL_BLUE, GL_ZERO, GL_ZERO, GL_ALPHA}, 		 // red only (cloaked)
-		{GL_ZERO, GL_ZERO, GL_ZERO, GL_ALPHA}, 		 // black only (outline)
-		{GL_RED, GL_RED, GL_GREEN, GL_ALPHA},		// 9 faded yellow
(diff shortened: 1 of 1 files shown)


ghoulavenger:
Not sure I care about the overlays, but I'd like to change regular colors, since doing the FW campaign pretty much locks you into green (I miss the blue merchant swizzle). The only question I have is did you include the ability to lock the swizzle during the FW campaign (because at that point you really need to be wearing their colors, it's a lore thing).

Amazinite:
https://github.com/endless-sky/endless-sky/issues/1779#issuecomment-266922948

>I don't think it makes sense to have 0-6 be normal swizzles, then 7 and 8 be strange, then it returns to normal at 9-28.

Same story here.

source/SpriteShader.cpp
```diff
@@ -31,20 +31,43 @@ namespace {
 	GLint blurI;
 	GLint clipI;
 	GLint alphaI;
-	
+
```
tehhowch: Your editor is eating whitespacing.

Disiuze:
> The only question I have is did you include the ability to lock the swizzle during the FW campaign (because at that point you really need to be wearing their colors, it's a lore thing).

What do you mean by this? This PR only adds new swizzles.
>Same story here. I also question the purpose of the overlays.

I'd like to keep it that way as I do not want to cause problems with cloaking or outlines.
What the overlays are used for is up to the content creators. I could see them being used for an alien government or unique NPC or ship.

ghoulavenger:
> What do you mean by this? This PR only adds new swizzles.

Ah, I was hoping for something that branches off of it then. I'd like to choose my own swizzle and read that as part of the PR (mistakenly).

Amazinite:
Given the fact that swizzles don't apply to a ship's turrets, I don't see the overlays being necessary in this PR until that happens. The picture you have for swizzle 31 shows what I mean perfectly; the turrets just pop out at you an look completely wrong on the ship.

czartrak:
I found myself wanting a black swizzle that doesnt make my entire ship black while writing a mission yesterday, so i'd like to have these

Lineth:
More swizzles would always be great. Quarg/Coalition would look good in black (probably the coalition more).

endless-sky:
As I've said before when this has come up: ships just don't look good if their color palette is limited to a 1D or 2D slice of the full 3D RGB color space. If it were possible to apply an arbitrary *hue shift* to the ship, so the colors changed while still occupying a full 3D color space, that would be a different story:

hue shift

That could be done by multiplying every pixel's RGB value with a 3x3 color transformation matrix, so it's doable and not too complicated. I'm not sure how much GPU overhead it would add, though.

Zitchas:
I'm not a programmer, but could you clarify in layman's terms how much overhead this would add? I'm running on a 6 year old desktop with a 2+ year old GPU, and ES currently barely registers as using any resources at all. I appreciate keeping the requirements low, but I think having a variety of new swizzles would be worth the GPU overhead (unless we're talking "becomes taxing for a modern GPU" level of workload, anyway)

Zitchas:
Is anything being done in terms of new swizzles or overlays?

I still really think we should have more than the current six, and overlays look like a good addition to me:
a) I'm getting sick of FW green
b) I suspect that I will get equally sick of Republic or Syndicate colours once getting locked into those.
c) As a content creator, having alternate swizzles for other alien races that aren't duplicating already-existing ones would be much appreciated.

PR #3599, opened 2018-01-23
Support "cloak disruption" as a weapon damage type

tehhowch:
Refs  d754ba6e2f7ccd105ab564c04445e69fb32fafc4, [comment](https://groups.google.com/d/msg/endless-sky/JlkMCWITkfU/8P7aw1jUBAAJ)
 - 1 point of actualized damage = 1 frame of cloak prevention
   - nominal is reduced by blast radius scaling and shields, just like other special damage types
 - Gave an example usage with Drak weaponry
 - Define a placeholder `effect` to visually convey that a ship cannot cloak
    - Not-great recording of it: https://youtu.be/V-Q0u1NUI-s
    - Better video: https://youtu.be/zlwDJxYjYTU


endless-sky:
To me it'd make more sense to have weapons (or sensors) that can target a cloaked ship, rather than just slowing down cloaking.

It might even make sense to apply an extreme nerf to cloaking by saying that weapons can still hit you when you're cloaked - all that cloaking does is prevent a ship from targeting you or a missile from tracking you.

tehhowch:
I didn't want to go too far into the lore of the cloak mechanism - so far all we "know" is that there are slight gravity distortions from cloaked ships.
I did consider going further and adding a "gravimetric scan" attribute along with the necessary targeting overhauls, but figured I shouldn't go that far right away.

KiLEdEnNis:
> It might even make sense to apply an extreme nerf to cloaking by saying that weapons can still hit you when you're cloaked - all that cloaking does is prevent a ship from targeting you or a missile from tracking you.

hmm, a new kinda missile tracking that CAN track cloaked ships
although we'd first need to figure out how we want cloaking to be; preventing a ship from targetting ya would be a bit... weird if ya were still visible
maybe some special effects that surrounds a "cloaked" ship could help; a ship could be "visible" but untargetable to most weaponry except the ones that have a special "anti-cloak" tag
the difficult part; what would those effects even be if we'd choose this

jafdy:
At some point we will need to have some kind of anti-cloak mechanic otherwise we will need to plan stories for players with and without a cloaking device. (Alternatively giving checkmate players a cloaking device at some point so the reconciliation reward is an earlier (or earlier then extra) device). 

Back on topic: Having a weapon that can hit cloaked ships and a sensor that can see cloaked ships might be good. With just the weapon you can try to hit a cloaked ship (blindly), with only sensors you can see the cloaked ship but can't do a thing about it, with both you can fight a cloaked ship normally.

tehhowch:
In a separate pull we could develop an active scan that effectively marks a ship, rendering it targetable even if cloaked. That could be achieved by tracking a new ShipEvent in the AI class, or by modifying the scanned ship such that IsTargetable() disregards its cloak check.

endless-sky:
What I meant was to make it so that cloaked ships are still invisible (and thus can't be targeted or tracked by missiles, and won't activate a projectile's `trigger radius`) but they can still collide with any hostile projectiles. Basically it would mean that a cloaked ship can't fly right through a minefield or a firefight, and if you cloak before a projectile hits you, you'll still need to dodge it.

That would still make cloaking useful for shedding missile locks, escaping combat, or flying through systems undetected, but it would be a bit less ridiculously overpowered.

tehhowch:
Indeed, it would make cloak much less overpowered. I think that change would simplify Engine a bit too, as cloakedCollisions could be dropped. I'll add that into this pull.

tehhowch:
The latest commit makes cloaked ships able to collide with any projectile. I considered moving Cloaking() to the Body class (which would allow for cloaked projectiles / asteroids) but that's a change for a different PR (specifically, one which adds logic to a Projectile to let it cloak :grin: )

endless-sky:
I'm still not convinced about how cloak disruption mechanics should work, so I've added the change to make projectiles hit cloaked ships separately (a4a4b3e).

tehhowch:
Reverted this PR to just allow weapons to indicate they increase a target's existing `cloakDisruption` counter

endless-sky:
Also, unless I'm missing something, right now in this commit cloaking disruption doesn't actually affect your cloaking speed in any way? It looks like the only effect it has is to show "sparks" around the ship.

tehhowch:
Based on your implementation of cloakDisruption last year, the ship can't cloak if it has any cloakDisruption, so it will just decloak if it gets hit - just like if it ran out of energy or fuel.

tehhowch:
@Kriegstofu, that comment seems unrelated to this PR. This PR lets the existing "cloak disruption" mechanic be activated by weapons. It also adds a visual effect that lets the player know cloak disruption is present.

Currently, the cloak disruption mechanic is activated only if the cloaking ship runs out of the needed resources to continue cloaking, and there is no visual conveyance of why the cloak has failed.

tehhowch:
The `antimatter spark` is actually really similar to the antimatter bolt (just with a shimmer or 3). Using the "nano spark" seems to go over well: https://youtu.be/jWlDIt6e7z4

Amazinite:
Looks good for now, but a proper unique effect is definitely in order.

Zitchas:
Just a question: How does cloak disruption interact with the variable cloaking speed. As in, will a ship with a fast cloak recover from the disruption faster than a ship with a slow cloak?

Likewise, is there any kind of resistance to this? Or any way to deal with the disruption other than simply waiting it out?

If not, I think there should be. Both, ideally. As in some form of resistance that increases how much disruption is needed to break a ship's cloak, as well as the ability to speed up recovery from the disruption).

tehhowch:
There is no relationship between cloaking speed and `cloakDisruption` - a ship with a non-zero `cloakDisruption` cannot cloak, at all, until it has returned to zero. 

It's possible to rename this existing behavior (e.g. `cloakMalfunction`) and continue supporting `"cloak disruption"` as a valid weapon damage type, one that affects the target in whichever new manner we decide is a better fit for affecting cloaks.

petervdmeer:
What about also automatically applying significant cloak disruption when the player gets near another ship (including friendly ships and disabled ships)?
The player is then still able to sneak up on 1 ship that the player wants to board, but it makes the player exposed for some time after such a boarding attempt. This also ensures that enemy ships searching/flying near a cloaked ship have a chance of finding the cloaked ship.
(In-game lore: the cloak will be disrupted by nearby objects just outside the cloaking field.)

And can we have cloaked ships trigger a "sensor echo" once in a while at some variable distance behind the player ship (without indicating the direction).

With those two changes, players (and the Arfectas) still can use the cloak to restore to full health, but they need to stay out of range from other ships (including disabled ones) in order to stay undetected. And just sitting around will not be sufficient, because the sensor echos will attract some attention.

Zitchas:
@petervdmeer Your suggestions are interesting possibilities, but I think are far too strong of a nerf to apply broadly to cloaking in general. If we had multiple cloak types, then I think it would quite interesting to have each of these things (Ship proximity disruption sensitivity and sensor echo as separate things) be separate attributes that could be applied to a particular cloaking effect (either to the outfit or to the ship, as appropriate) to create different trade offs between styles.

I'm adding these ideas to the discussion in #4119 topic which more broadly discusses cloaking types in general.

LocalGod79:
shouldn't all weapons collisions have a minor cloak disruption?

since the cloak ceased avoiding projectiles, i've often pondered - "wouldn't this ship shooting at someone and else having their projectiles disappear right in front of them notice?"

tehhowch:
> shouldn't all weapons collisions have a minor cloak disruption?
> 
> since the cloak ceased avoiding projectiles, i've often pondered - "wouldn't this ship shooting at someone and else having their projectiles disappear right in front of them notice?"

The `cloakDisruption` implemented by Michael wouldn't be a good fit for that, as it forces a full reset of the cloak state. I would imagine being hit to be only a minor disruption, e.g. reducing the current `cloak` value from 1 slightly, but otherwise not doing much.

LocalGod79:
i posit that if a hail of Gatling bullets (or the revised Railgun slugs) starts striking a cloaked ship, that the disappearance of that many projectiles would show up on everyone's sensors, effectively disrupting that ship's cloak continuously.

perhaps a (separate) coding change to cloak is in order, so that projectile impacts reduce the cloak value as you mentioned.  should this be a separate issue?

Amazinite:
> should this be a separate issue?

Yeah, this sounds like something separate entirely that could be more easily described as “cloak integrity”: how well the cloak remains active when under fire. Cloaking disruption is more for weapons *designed* to prevent cloaking.

Zitchas:
I think such weapons impacting a cloaked ship shouldn't affect the cloak - but it should create some kind of "ping" effect on sensors. Basically, your targeting system going "Hey, something's up there." but doesn't affect the cloak at all. (In terms of AI, this would be the sort of thing that might allow them to target a cloaked ship with guns/unguided projectiles, but they would not be able to actually "lock-on" with turrets or guided munitions.)

Amazinite:
This doesn't display in the outfitter right now. I'd suggest `cloak disruption`, as `cloak disruption damage` is too long.

Amazinite:

PR #3600, opened 2018-01-24
Replace most gov->IsPlayer with ship->IsYours

tehhowch:
- Swaps to using the less-complex yet still-correct check `Ship::IsYours()` rather than `Ship::GetGovernment()->IsPlayer()`
 - Every mission NPC declares its government
 - Swapped a number of unspecified (i.e. Player government) NPCs to using the more-appropriate Free Worlds government

### DIFF
--- a/data/free worlds checkmate.txt
+++ b/data/free worlds checkmate.txt
@@ -91,7 +91,8 @@ mission "FWC Attack Kaus Borealis"
 		event "fwc capture kaus borealis"
 	
 	npc
-		personality heroic
+		government "Free Worlds"
+		personality heroic escort
 		fleet
 			names "free worlds capital"
 			fighters "free worlds fighters"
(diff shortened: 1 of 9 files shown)


endless-sky:
There are situations in the Free Worlds story line where escorts have to use the Escort government, e.g. because the Republic is granting a ceasefire to the player but not the the Free Worlds as a whole.

tehhowch:
I believe I have kept those as the proper setting, and here just make it explicitly clear when there are escorts, by giving them `government "Escort"`

The missions concurrent with `fw safe passage` event (player rep only):
 - FWC Diplomacy 1B (given Escort)
 - FWC Diplomacy 1C (given Escort)
 - FW mid Diplomacy 1B (no escorts)
 - FW mid Diplomacy 1C (no escorts)
 - FW mid Diplomacy 1D (no escorts)
 - FW Southern Prisoners - Release (not altered here)
 - FW Southern Prisoners - Released (not altered here)

Issue #957, opened 2016-04-15
Enhancement: fighters/drones rally at their bay when being recalled


LocalGod79:
instead of the centre of their carriership.

the current effect is jarring for external bay ships like those in #939 (which is complete and ready to merge, btw)

endless-sky:
It's doable, but would take a bit of work, because right now a particular bay is not assigned to a particular fighter until the fighter docks. I could just have each fighter fly to the first open bay, but then you'd have multiple fighters trying to dock in the same bay.

LocalGod79:
perhaps assign bays at the moment that the player hits 'r'ecall?

LocalGod79:
the more that i think about this, the more i think that this might be the job of the 'retrieval effects' in #944

Saturnalia9:
So... soonTM?

Saugia:
https://github.com/endless-sky/endless-sky/pull/6130#issuecomment-2253121171

Issue #944, opened 2016-04-09
Feature Request: "launch"/"retrieve" effect for fighter/drone bays


LocalGod79:
pretty self explanatory, i think - decompressing bay would be neat, at the very least.

a two step effect would be nice (bay covers part of ship sprite, rotates to launch position, launches/retrieves fighter/drone, bay returns to "idle position"), but i think would get complicated, code wise.

endless-sky:
A decompression effect would be pretty easy to add - maybe just one animation that's shown for any drone bay launching a drone, and one for any fighter bay, not counting any external bays. Also technically, it should only be shown for non-automaton ships, since automatons would not have atmosphere.

LocalGod79:
people still work on the drones, though, don't they?  or drone maintenance is automated within their bay?  or they use EVA suits?

i guess there are enough possible explanations to omit an effect from the drone launch for the time being. (it appears you plan on automaton carriers launching automaton drones)

even external bays could use smaller omnidirectional decompression effects - i won't push it, because we may have differing ideas of what e'bays look like, and how they function...

obviously for more complicated effects (silos / elevators / rotating hangars / retractable pods / etc.) the effect would need to be defined as a ship attribute, tied to the bay (possibly adding another token to that line)...

mdsmestad:
I like this idea. Would add a nice little detail

toilethinges:
So when you press deploy, it shoots out an effect of choice?

LocalGod79:
i threw a few ideas out there...but yeah, making it user definable would make it easier for content creators to create a wide variety of deployment effects...i've outlined a few effects that would remain on the carrier, but "shooting" effects (beyond puffs of atmosphere) are always a possibility - i suppose one could even add a "single fire booster rocket" to deploying fighters (that even diverges from the fighter when it begins to track a target)...

tehhowch:
#3602 would add support (and a default "pressurized bay launch" effect) for launch effects.

PR #3602, opened 2018-01-24
Support launch effects for carried ships when departing from their carrier

tehhowch:
Refs #944

Until #957 is supported, supporting "retrieve effect" doesn't make much sense, since the effect location will be in a different place than the carried ship "gets absorbed" by the carrier. With #957 supported, adding support for a "retrieve effect" won't be difficult.

Example bay definition:
```
drone -66 24 over left
    "launch effect" "external launch"
```

Caveat: effects for an "under" bay will be drawn on top of the ship. I don't think this part is easily addressed.

I've also defined a "basic launch" effect that any non-automaton ship will use for internal bays (that do not define their own launch effects), to prevent needing to go through every existing ship definition and add it explicitly.
[Basic launch](https://youtu.be/IQVC30iEi78)
(I'm sure this effect could be greatly improved by an artist :grin: )

### DIFF
--- a/data/effects.txt
+++ b/data/effects.txt
@@ -143,6 +143,19 @@ effect "final explosion large"
 	sound "final explosion large"
 
 
+
+effect "basic launch"
+	sprite "effect/smoke"
+		"no repeat"
+		"random start frame"
+		"frame rate" 8
+	"lifetime" 30
+	"random angle" 15
+	"random spin" 360
+	"random velocity" 3
+
+
+
 effect "smoke"
 	sprite "effect/smoke"
 		"no repeat"
--- a/source/Effect.h
+++ b/source/Effect.h
@@ -13,9 +13,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #ifndef EFFECT_H_
 #define EFFECT_H_
 
-#include "Angle.h"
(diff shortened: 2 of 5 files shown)


tehhowch:
Sounds are implicitly handled in the creation of a `Visual` (a bit misleadingly named...):

https://github.com/endless-sky/endless-sky/blob/e6cc89a233a2fc200ae971b9ffade21c232a0c6b/source/Visual.cpp#L35-L36

So I'm not sure it makes sense for the game to also play a distinct deploy sound for each deploying bay, as each bay might also specify exactly which sound(s) it is supposed to play, according to the ordered list of `Effect`s it was given. It's also important to note that deploying is not a synchronous activity - by design there is some randomness to the deploy rate, so any sound associated with the deploy will be heard multiple times for almost every ship in the game.

I've amended the PR to allow shorthand list initialization, e.g. you can add 5 of the same effect by adding a count after the name of the effect:

```
drone -66 24 over left
    "launch effect" "external launch" 5
```
is equivalent to
```
drone -66 24 over left
    "launch effect" "external launch"
    "launch effect" "external launch"
    "launch effect" "external launch"
    "launch effect" "external launch"
    "launch effect" "external launch"
```

Issue #3605, opened 2018-01-24
Show the viewport on the radar


jostephd:
When I see a ship's blip on the radar, I'm never sure how far the ship is from me. I thought it would be easier to get a sense of how far things on the radar are if the boundary of the viewport were displayed subtly on the radar.

tehhowch:
I don't think adding the full box would be done, but corner markers could be possible. As of right now the radar scale is immutable, so the distance is always the same - just as learning what the proposed viewport markers represent, the player can learn the distance relative to their ship.

tehhowch:
This actually turned out kinda neat:
![zoomed out](https://user-images.githubusercontent.com/20871346/35361179-fc099874-0125-11e8-913f-a5dd200d270c.png)

jostephd:
That was fast, thanks!

endless-sky:
I'm really not convinced that this would be useful to most players. Because the radar scale is fixed, you learn pretty fast how to tell how close an object on the radar is. And in most cases what you're really interested in is how quickly the object is moving toward you or away from you (e.g. how long it will be until that missile hits you).

jafdy:
Up until recently I still thought the scale was logarithmic, and the distance to the edge was quite a bit further. So, I've been at this a while and never got the hang of the scale it's quite likely that there are others out there that haven't either.

jostephd:
The corner marks also help in getting a quick estimate of the distance in terms of screen-heights / screen-widths, which is a larger unit than "ship length" or "Heavy Laser range" or "three seconds at maximum forward speed" and so more amenable to estimates. Essentially, the corner marks are a point of reference that gives a sense of scale. The existing concentric circles don't serve that purpose well because there is no easy way for a player to learn what distance in the playfield they correspond to.

Janaszar:
Just a thought, but what if the viewport had the ranges of the equipped weapons?  Maybe just a general "can hit within this" circle.

tehhowch:
There are a lot of variables that need to be addressed if showing weapon ranges, to the point that I would think it to be useless in practice. Just install a Tactical Scanner and you can get a nice range estimate to your target.

Pointedstick:
I would really like this. I feel like it would essentially add a scale to the radar display, which is otherwise a bit hard to understand.

jostephd:
#3607 has been merged, this can be added to the `0.9.9 (done)` milestone.

PR #3607, opened 2018-01-24
Draw viewport boundaries on radar

tehhowch:
Refs #3605 

Scales up & down, ensures that the lines do not overrun the radar boundary. Even for a tiny screen (1024 W), the lines won't overlap and are still visible.
max zoom @ 1024x863
zoomed in

min zoom @ 1024x863
zoomed out

Uses an opt-out hidden preference `"Disable viewport on radar"`; I'm happy to alter that.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1834,6 +1834,15 @@ void Engine::FillRadar()
 				system->Position() - playerSystem->Position());
 	}
 	
+	// Add viewport brackets.
+	if(!Preferences::Has("Disable viewport on radar"))
+	{
+		radar[calcTickTock].AddViewportBoundary(Radar::INACTIVE, Screen::TopLeft() / zoom);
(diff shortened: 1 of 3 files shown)


endless-sky:
I'm not convinced that people would want this on by default. To me it just adds confusing clutter to the radar. What is the use case where this would be helpful?

Amazinite:
More easily judging distances from outside your field of view. 

I don't see this being confusing at all, but that might just be because I play RTS games where this is a very standard way of showing your field of view on the map in relation to the entire map.

KiLEdEnNis:
> but that might just be because I play RTS games where this is a very standard way of showing your field of view on the map in relation to the entire map.

zooming out & using the mouse to control ships individually gave the game a bit of an RTS touch, so this change would fit pretty well for players that do this

Issue #3611, opened 2018-01-26
Outfitter misaligns decimal places


jostephd:
The Hai Plasma Rifle is displayed as follows:

![decimal-point](https://user-images.githubusercontent.com/24784687/35447445-aa00f2bc-02af-11e8-8f2b-aae6605fa1c5.png)

This is easily misread as ".7 attack, .1 defense" or "7 attack, 1 defense".

Possible fixes:

- Always display a leading zero on fractions:

```
attack:  0.7
defense:   1
```

- Align all decimal points vertically:

```
attack:   .7
defense: 1.
```
(with leading/trailing zeroes according to MZ's taste)

- Scale all numbers by 10 to avoid decimals:

```
attack:   7
defense: 10
```

Brick63:
the only problem i can see is outfits like the bio-defenses at 250 causing issues but that's about it

KiLEdEnNis:
I'd like 0.7 and 1.0 but the unnecessary zeroes may not be appreciated by everyone

endless-sky:
This seems highly subjective - some people will prefer .7, and some will prefer 0.7.

Hacklin:
@endless-sky: Sorry no, most people use 0.7.

In natural languages and in C/C++ printed output  there is always at least one digit to the left of the decimal separator (= '.' | ',').
C/C++ allows zero digits to the left of the decimal point to be omitted in literal floating-point numbers.
Not all computer languages allow this.
Most natural languages disallow this or at least frown upon this, because the point (or comma) could be misinterpreted.

Guess you've been programming in C++ for too long ;-)

KiLEdEnNis:
it's partially just a matter of preference, but 0.7 does make it more clear than .7, so it should reduce possible confusion

Hacklin:
std::cout << .7 << std::endl;

Tadrix:
0.7 would certainly be better.

> When the integral part of a numeral is zero, it may occur, typically in computing, that the integer part is not written (for example .1234, instead of 0.1234). In normal writing, this is generally avoided because of the risk of confusion between the decimal mark and other punctuation.

jafdy:
0.7 does seem like a good idea. I and the people I work with would use .7 in our own day to day work so 0.7 is not entirely normal.

jostephd:
[...]

PR #3613, opened 2018-01-26
Always put a digit to left of the decimal separator.

Hacklin:
Show 0.7 instead of .7.
Natural languages always have at least 1 digit to the left
of the decimal separator; unlike some computer languages like C/C++.
Fix issue #3611

### DIFF
--- a/source/Format.cpp
+++ b/source/Format.cpp
@@ -74,7 +74,7 @@ string Format::Number(double value)
 	if(left > 3)
 		delimiterIndex = left - 3;
 	
-	while(rounded || right)
+	do
 	{
 		int digit = rounded % 10;
 		if(nonzero || digit)
@@ -90,7 +90,11 @@ string Format::Number(double value)
 			if(!right)
 			{
 				if(nonzero)
+				{
 					result += '.';
+					if(!rounded)
+						result += '0';
+				}
 				nonzero = true;
 			}
 		}
@@ -101,6 +105,7 @@ string Format::Number(double value)
 				result += ',';
 		}
 	}
+	while(rounded || right);
 	
 	// Add the negative sign if needed.
 	if(isNegative)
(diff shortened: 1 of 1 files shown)


jafdy:
I think this is a good change but you comment is a bit presumptuous, I don't say "zero point seven", I say "point seven", I also write .7 not 0.7 and typically type it too. So, good readability change but there isn't always a digit left of the point.

Hacklin:
Except for computer languages I've never seen it written as .7 always as 0.7.
And in Dutch and German you'd also say 0,7 and not ,7.

Could you point me to a publication in which they omit the 0 before the dot? (i.e. write it as .7)

tehhowch:
I believe jafdy is speaking in terms of intermediate work, and not formal printed publications, i.e. when conversing with coworkers, family, etc

Hacklin:
Ah.
I know it's done in english spoken text,
so it would make sense to write .7 in mission conversations.
But we're talking about descriptions of ships and outfits.

In other words it depends on where the number is being used:
- 0.7 in descriptions of ships, outfits, items,
-  .7 in spoken conversations,
- 0.7 in the HUD / main window ??
- always 0,7 for languages that use a comma as decimal separator
  (just in case ES gets translated).

I could add an extra parameter to the format function.

Tadrix:
0.7 is a good change in my opinion.

There's no really good way to cover _spoken_ conversations AFAIK. At least I heard (admittedly in SF movies and shows most of the time) both "point-seven" and "zero-point-seven", and exact wording should be left for mission writers to decide on.

However, where the game represents numbers numerically, they probably should be represented as in formal writing (at least where it is handy for them to be standardized).

source/Format.cpp
```diff
@@ -101,6 +105,7 @@ string Format::Number(double value)
 				result += ',';
 		}
 	}
+	while(rounded || right);
```
tehhowch: For `do {} while` loops in Endless Sky code, the `while` should be on the same line as the closing `do` brace.

Hacklin: Thanks.

PR #3615, opened 2018-01-27
Update erase usage

tehhowch:
- Variant toRemove only needs to be initialized once
 - set.erase first performs a count, and erase(<key>) returns 1 if <key> was erased
 - System.links & System.neighbor used to be vectors, but were made sets in 4fa19f67a4f4e559f08b769bf7596ff52005eff2, and set supports erase(<key>)

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -815,11 +815,9 @@ void Mission::Do(const ShipEvent &event, PlayerInfo &player, UI *ui)
 	{
 		const System *system = event.Actor()->GetSystem();
 		// If this was a waypoint, clear it.
-		if(waypoints.count(system))
-		{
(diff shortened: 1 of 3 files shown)


source/Mission.cpp
```diff
@@ -865,11 +863,8 @@ Mission Mission::Instantiate(const PlayerInfo &player) const
 		result.waypoints.insert(system);
 	}
 	// If one of the waypoints is the current system, it is already visited.
-	if(result.waypoints.count(source))
-	{
-		result.waypoints.erase(source);
+	if(result.waypoints.erase(source));
```
jostephd: Stray semicolon?

tehhowch: ah yep. :man_facepalming:

PR #3617, opened 2018-01-30
Set Deneb government to "None" when Pug leave

Nescio0:
When the Pug flee or are defeated in the Free Worlds campaign, Deneb joins Tarazed's Neutral government. This is problematic:
* gameplay: if you demand(ed) tribute from any or all Neutral (i.e. Tarazed) planets, you can no longer land in Deneb
* lore: no <s>humans live</s> human government controls Pugglemug or Pugglequat

<s>This patch changes the government to "Uninhabited" instead (cf. Kor Sestor planets), thus always allowing any player to land in Deneb.</s>
This patch assigns Deneb to a newly introduced "None" government, which is unrelated to other governments and unaffected by reputations.

PR #2541, opened 2017-05-22
Added an non-neutrality event to trigger at the end of the FW storyline

jafdy:
It feels odd to have these 5 systems continuing to be neutral after the war is over. So I added an event that would make them either FW (for near Tarazed) and Republic for Poisonwood.

Issue #1782, opened 2016-10-28
Why are Neutral planets still "Neutral" after the war has ended.


Admiral-Danforth:
Change them to independent or something. Another option would be to have more backstory surrounding this.

jafdy:
Deneb should stay neutral but Tarazed and surrounding systems should change to FW and Poisonwood should change to Republic.

Bladewood:
And Girtab stays neutral?

toilethinges:
Girtab independent?

jafdy:
I bundled Girtab with the near Tarazed systems. The assumption was that it was made neutral so that the player (and any NPCs) didn't have to be able to make 4 jumps.

tehhowch:
#2541 attempted to address this, I think.

Amazinite:
This has been covered in 9804d6c.
Poisonwood now shows as part of the Republic once the Free Worlds and Republic are friendly.
Harmony now joins the FW doing the eastern expansion instead of remaining neutral.
Tarazed either joins the FW or shows as part of the Republic again depending on the branch you take.

PR #3618, opened 2018-01-31
Non-neutrality events for "Neutral" systems

Amazinite:
References: #1782, #2541, #3617 

The point of neutral systems during the war is that they have not chosen a side. This allows the player to land on these planets when hostile to their in-universe parent governments, the suggestion there being that when Tarazed and its related systems become neutral, they are still technically under the jurisdiction of the Republic, the same being true with Poisonwood. These systems have changed the government they display in-game for gameplay purposes, but in-universe their government has not actually changed. (Edit, see also: ["Neutral" basically means, "Still part of the Republic but willing to do business with the Free Worlds."](https://github.com/endless-sky/endless-sky/pull/3619#issuecomment-362658324))

Instead of taking the approach of #2541 and changing these systems after the war, this PR changes the government of neutral systems when their neutrality is no longer necessary for the sake of gameplay.
* Poisonwood becomes no longer neutral after the event "fw armistice", in which the Republic is no longer hostile toward the Free Worlds, meaning the neutrality of Poisonwood is no longer necessary, reverting to its parent government of the Republic.
* During the event "fw southern expansion", the system Girtab becomes surrounded by FW systems, meaning it is no longer necessary for it to be neutral, causing it to join the Free Worlds.
  * This means that during events in which GIrtab is released from Navy control, it reverts back to a Free Worlds system instead of a Neutral system
* Tarazed is handled differently depending on the branch
  * Reconciliation: After the player goes to Parliament to give the evidence, it's clear that the FW are innocent, leading Tarazed to join the FW.
  * Checkmate: After the event "fwc capture kaus borealis", it's clear that the FW are using nukes. Tarazed severs ties with the FW.
* Deneb remains neutral as the reasons for its neutrality are separate from the reasons why these other systems are neutral, as it becomes a neutral system after the war has ended.


KiLEdEnNis:
is it really necessary to make Tarazed become Free Worlds when the Free Worlds surrounds it? the war is still going on and Tarazed should stay neutral until the war ends, and then become Free Worlds because they want to
at least that's my opinion, but it's fine either way

jafdy:
That's what I was going for as well, but mine was closed.

Amazinite:
>is it really necessary to make Tarazed become Free Worlds when the Free Worlds surrounds it?

It's not a question of necessity (or timing, really), but of reasoning; if Tarazed and the related systems are surrounded by the Free Worlds, then what is the reasoning for them to remain neutral, especially if they'll ultimately become part of the Free Worlds anyway? I pointed out in #3617 that when Tarazed and its related systems switch to the "Neutral" government, the Navy does not react. This backs up my idea stated in the PR description that the "Neutral" government is acting as a change for the sake of gameplay and does not represent the actual government of the Tarazed systems. Once surrounded, these systems then no longer have any connection to the Republic, and should therefore be part of the Free Worlds.

Nescio0:
Of the "fw expanded and cut" event, the "expanded" part means a string of five planets planets to the northeast of Tarazed join the Free Worlds, but the "cut" part means Tarazed remains connected to the Republic (Eltanin, Delta Sagitarii, Rastaban), therefore it would make sense Tarazed and its Neutral systems won't join the Free Worlds simultaneously. 
A later "Tarazed joins the Free Worlds" event would make more sense.

Amazinite:
Even with the loss of territory, the Tarazed systems are still surrounded by FW systems, even if that chunk of systems is separated from the main FW systems, as Eltanin is an uninhabited system, but Albaldah closes the Tarazed systems in with Kaus Australis. At least for the system of Tarazed at that time, there is 0 Navy fleet influence in the system.

KiLEdEnNis:
realistically Tarazed may be hesitant to join the Free Worlds after they are surrounded by them because the Navy could still strike back
but once the FW & Republic make peace (Reconciliation) or once the FW shows they can defend their space (Checkmate start), Tarazed would feel "safe enough" to join them (meaning they join a little later than in your PR yes?)

Amazinite:
So are you suggesting that all systems become neutral after fw armistice instead?

tehhowch:
I don't agree that the Free Worlds use of nuclear weapons would align worlds to their cause. I think the opposite would happen, resulting in an increase in time required for neutral planets to join the Free Worlds.

I'm also of the mindset that planets which have become neutral should remain neutral throughout the conflict duration. When the conflict changes direction, systems nearer to the newest threat are likely to search for safety through reaffirmed ties with existing powers, while those distant are not going to feel that pressure.

Perhaps what is truly needed are appeasement / lore missions wherein the  player leads or learns of actions that help influence a planet to align with `<faction>`
Simply being surrounded is IMO insufficient cause for a change in government.

Amazinite:
To note, the Tarazed Diplomacy mission talks about Tarazed being neutral *for the time being,* so the discussion then becomes when the best time for the Tarazed systems to join the Free Worlds would be. It looks like so far, the only complaint is on Tarazed and its two neighboring neutral systems joining the Free Worlds during "fw expanded and cut". If not during "fw expanded and cut", when else would you consider it reasonable for the systems in question to join the Free Worlds? During "fw armistice"? At the very end of the war? Exactly what was closed in #2541 and have Tarazed's government change be delayed?

Nescio0:
>  when else would you consider it reasonable for the systems in question to join the Free Worlds?

Instead of at the events you proposed, I'd prefer 30 days after the Kor Sestor are defeated. 
E.g. `data/wanderers middle.txt` line 2578:

> The way Danforth talked about the device he gave you made you assume that it was a nuclear bomb, but the explosion that rocks <origin> a few seconds after you depart is far too powerful for that. The light of the explosion is many times brighter than this system's sun. And unless your sensors are lying to you, there is now a crater near <origin>'s south pole that goes all the way down to the mantle. Lava is filling the hole, and the entire planet's crust is shifting and shivering in response.

The revelation the Navy has a secret arsenal of such weapons and the willingness to use it on human planets seems a good enough reason for a public outcry, resulting in Tarazed c.s. to reconsider their position and finally join the Free Worlds.

tehhowch:
And if the player does not drop the bomb? When then?

Nescio0:
> And if the player does not drop the bomb? When then?

Still at the same moment. If the bomb isn't dropped the case remains the Navy has such weapons and is willing to use them (even if the player decides otherwise). The principle ought to be enough for a public outcry.

tehhowch:
I'm unaware the player or anyone but Danforth and his regiment know of the device.

Anyhow, branching events off a still-in-development storyline is probably not a good idea.

KiLEdEnNis:
> So are you suggesting that all systems become neutral after fw armistice instead, or after the FW start flinging nukes?

my thoughts:
for the Reconciliation, I'm not sure at what time it'd be convenient, but after the Republic ceases hostility to the FW

for the Checkmate, either after ya win the first large battle against the Navy again (although I forgot if that's where the nukes are used already), or as soon as you decide to go for the Checkmate branch, Tarazed decides to join in an effort to help them
I don't really see a "cleaner" way to do it for this branch, but if ya do find one, go ahead

Amazinite:
>my thoughts:
for the Reconciliation, I'm not sure at what time it'd be convenient, but after the Republic ceases hostility to the FW
>for the Checkmate, either after ya win the first large battle against the Navy again (although I forgot if that's where the nukes are used already), or as soon as you decide to go for the Checkmate branch, Tarazed decides to join in an effort to help them
I don't really see a "cleaner" way to do it for this branch, but if ya do find one, go ahead

Just to note, "fw armistice" happens right when the player reaches the defector, and the first large battle against the Navy happens like right after choosing Checkmate, so this would just be delaying the joining of Tarazed 1 or 2 missions.

Nescio0:
"fw expanded and cut" is problematic because being surrounded is not a good reason to join another state.
"fw armistice" is problematic because an armistice is not a peace treaty; war could resume later; therefore Tarazed is more likely to wait and see what will happen (as was their policy during the war).
"fwc peace with the navy" is problematic because it also changes Vega to Pug, therefore Tarazed is likely to find out what's happening first.

How about a few years after completion of the storyline, to allow the Neutral government(s) time for careful deliberation and secret negotiations? E.g.:

```
event "Tarazed joins the Free Worlds"
  "reputation: Neutral" = 1
  system "Tarazed"
    government "Free Worlds"
  system "Umbral"
    government "Free Worlds"
  system "Dabih"
    government "Free Worlds"
  system "Lurata"
    government "Free Worlds"
  system "Albireo"
    government "Free Worlds"
  system "Sadr"
    government "Free Worlds"
  system "Girtab"
    government "Free Worlds"
  system "Alpha Arae"
    government "Free Worlds"
  system "Eber"
    government "Free Worlds"
  system "Alnasl"
    government "Free Worlds"
  system "Graffias"
    government "Republic"

mission "Tarazed switches to the Free Worlds"
  landing
  invisible
  to offer
    has "main plot completed"
  on offer
    event "Tarazed joins the Free Worlds" 777
    fail
```

KiLEdEnNis:
"hey why did that system suddenly become FW after 2 years?"
too confusing

Amazinite:
Going back to this:

>Of the "fw expanded and cut" event, the "expanded" part means a string of five planets planets to the northeast of Tarazed join the Free Worlds, but the "cut" part means Tarazed remains connected to the Republic (Eltanin, Delta Sagitarii, Rastaban), **therefore it would make sense Tarazed and its Neutral systems won't join the Free Worlds simultaneously.**

Tarazed can't see into the future. The story unfolds with the player getting the five mentioned systems northeast of Tarazed and its neighboring systems to join the Free Worlds after they sign the paperwork, followed by the Navy moving in to the FW's southern expansion area. Without this attack, the Tarazed would have been boxed in. Not knowing the attack was going to happen and using the standard that seems to be in agreement that Girtab should flip to FW when surrounded, it would not be unreasonable for Tarazed and its neighbors to also join the Free Worlds along with the other five systems in the "expanded" part of that event, and it would not just be the flip of a switch for them to go back to "Neutral" after seeing that the Navy has rolled in nearby.

Nescio0:
As long as they're neutral neither the Republic nor the Free Worlds will attack them. Once they choose sides there is no going back to their neutral position. Exactly because they can't see into the future and do not know how the war will unfold Tarazed would not and should not give up its valuable neutrality lightly or prematurely.

jschiav:
From the mission text, it seems heavily implied that Tarazed's decision to remain neutral is actually based on what's best for the Free Worlds, NOT what's best for Tarazed. The first thing the CEO says is that Tarazed is sympathetic to the FW cause. When Alondo suggests Tarazed act as a go-between with the Syndicate, the CEO even goes so far as to caution that not joining officially would allow the Republic to have free access to the system.

It seems pretty clear to me that Tarazed would join the FW as soon as the FW didn't need their neutrality for practical reasons.

Full text:

>  "I'm pleased to welcome you here," says Emily. "Many of us are sympathetic to your cause."
> 
> "Thank you," says Alondo, "although for the time being the best way to support us, may be for you to remain neutral in this conflict." He explains the idea of Tarazed being a go-between with the Syndicate.
> 
> "I don't like dealing with the Syndicate," says Emily, "but like yourselves, we could not build ships without the systems they manufacture. Of course, the materials we receive from Free space are equally valuable. But if we remain nominally a part of the Republic, they will have full access to this system, including searching incoming and outgoing freighters."
> 
> "True," says Alondo, "but a Navy fleet here would be vulnerable to being cut off from Republic space, so unless they occupy the whole Dirt Belt, I do not think they would risk it." They continue talking for a while, and you leave well convinced that Tarazed is willing to support you and that the best way they can do that for now is by remaining neutral.

Amazinite:
I've changed the dialog in the FW missions that mention the expansions to reflect the new changes in neutrality.

Amazinite:
So what are people's thoughts on Tarazed flipping from neutral back to Republic once the FW use nukes in the Checkmate route, but joining the FW during Reconciliation shortly after the FW-Navy conflict ends?

Zitchas:
I think that makes sense to me. Gives the branch a little bit more impact on the world at large, which is nice.

Amazinite:
I've put the events in place that lead to Tarazed flipping FW or Republic depending on which branch the player chooses. How should this change be relayed to the player during the missions?

After Tarazed joins the FW, should some dialog be added in "FW Reconciliation 2A" that goes something along the lines of `You receive a message from Alondo. "Great news! Tarazed has heard of the Parliament hearing and will now be joining the Free Worlds."`

After Tarazed reverts to Republic, should some dialog be added in "FWC Cebalrai 1" that goes something like `Alondo looks rather concerned before saying "I just received word that Tarazed has decided to officially sever ties with us." || "Bold move for a system we have completely surrounded," Tomek says. "They're none of our concern right now. Let's keep making our way to Sol."`

jafdy:
`...joining the Free Worlds officially."`? They were kind of going that way at the start when they became neutral.

Issue #3516, opened 2018-01-08
TDR crashes (possible solution, need user input)


whismerhill:
hello, continuing from this issue : https://github.com/endless-sky/endless-sky/issues/2067
who was wrongly named (because it happened over much more driver revisions than what's written in the title as seen in the comments) & wrongly closed (because it never got fixed whereever the issue comes from, be it an nvidia issue, an endless sky issue, both, or neither)

so here I am to reopen this issue and get people possible **feedback** from people who already experienced the issue.
Please do state : 
 -if you still have TDR crashes ? ( _a TDR crash happened when you see the message : "Display driver stopped responding and has recovered," and in event viewer you should see something like "A TDR has been detected. The application must close.  Error code: 7  (pid=5152 tid=5940 endlesssky.exe 64bit)"_ )
-which nvidia driver revision you're running  ?
-which endless sky version you last played with (github, 0.9.8 , 0.9.7 ......) ?

Because **I just built endless sky with Visual Studio Community, left it running overnight without any problem**
steam thread  : http://steamcommunity.com/app/404410/discussions/1/143388511231569856/?ctp=4
current github build : https://github.com/whismerhill/endless-sky/tree/VS2017

Thank you very much.

Thanks & Testers : 
 - M\*C\*O from discord left it running overnight in space without issues however he doesn't have TDR.
GTX  1060 6GB, DirectX12, OpenGL4.6, Drivers v.388.71, Windows 10 64bit

endless-sky:
Closing this because this appears to just be an announcement of the fact that you have no issues and have been unable to reproduce the named issue.

whismerhill:
what the fuck ?
@endless-sky

KiLEdEnNis:
> what the fuck

don't, please

whismerhill:
I don't care. I've spent HOURS diagnosing this, and he just closes this like right out of the blue without even trying to contact me first

whismerhill:
I have a lone tester beside MCO who is currently testing out the executable I compiled with VS
and no the problem is not gone and a bunch of people were affected from the steam forums and whatnot. I just can't believe this.
Alright I'm done seriously

whismerhill:
the only reason this isn't a PR instead of an issue is : 1 : I wanted more testers & feedback
2 : because you previously stated you didn't want to compile in yet another compiler which needs to be maintained afterward  : "VS", however the fix to this issue seems to be exactly that

[...]

PR #3622, opened 2018-02-01
Update includes & forward declarations

tehhowch:
- Removed some legacy includes
 - added includes where list/vector holds the complete type and not just a pointer
 - Swapped includes to forward declarations for shared_ptr<Type> if possible

I was testing some changes to Planet.h and noticed some unexpected files needed to be recompiled. Traced that to an `#include "Planet.h"` in PlayerInfo.h, and then found some missing includes, and decided to go through everything.


endless-sky:
A forward declaration in the header should be sufficient if a type is only used in a vector or a list. Both those classes are just a wrapper around dynamically allocated memory. The only time a forward declaration does *not* work in a header file is if the size of the class that is defined in that header depends on the size of the class that is being forward-declared.

To put it another way, a vector object is the same size regardless of what it's a vector of, and you don't need to know how big the things that the vector holds are until you actually make use of it (which happens in the .cpp file).

tehhowch:
Fair point. I'll revisit with that in mind. I had read a bit about that, but wasn't sure if we could count on that behavior across compilers (looking mostly at being compatible with MSVC, as for some players compiling with MSVC seems to resolve a host of stability issues, e.g. #3516, that are otherwise present in the steam / packaged binary.)

tehhowch:
I went through the repo again, and got a few more missing includes, and header includes that could be replaced with forward declarations. I had to basically fall back to compiling after every change, since my intuition for when `container<class B> b` influences the size of class `A` isn't up to snuff.

For example:
1. PlayerInfo holds `list<GameEvent>` and has `#include GameEvent.h` in its header. Changing to a forward declaration here results in allocator errors complaining about the incomplete `GameEvent` type when compiling `main.cpp` (which creates the PlayerInfo object).

2. Engine holds `vector<PlanetLabel>` and ha(d) `#include PlanetLabel.h` in its header. Changing to a forward declaration here does not result in allocator errors when compiling.


Given 1), I would have expected allocator errors to appear when compiling MainPanel, as it creates the Engine object and has no idea about the PlanetLabel class beyond the forward declaration, but no such issues occur.

That said, you're busy, so don't go out of your way to provide basic C++ knowledge :grin:

flaviojs:
`list` probably has a root node variable (instead of a pointer) because it simplifies logic, which would require knowing the size of the type.

MainPanel.cpp receives PlanetLabel.h from Engine.h "for free".

tehhowch:
With the PR, PlanetLabel's header is no longer in Engine's header, so only Engine.cpp knows about it - same as with GameEvent` and `main.cpp` if we use forward declares for all `list<GameEvent>` cases.

Mostly it'd be nice to be able to alter `MissionAction`, `Mission`, and `GameEvent` without recompiling everything that uses includes `PlayerInfo.h`.
Someday I might try setting up a managing class that holds a reference to PlayerInfo, and have it hold the lists of missions and mission-related logic

Issue #2524, opened 2017-05-14
Add support for multiple fleet types when specifying tribute fleets


Disiuze:
Currently, tribute only supports spawning one fleet type. I learned this the hard way when I had to [remove the merchant fleets from Greenwater](https://github.com/Disiuze/TributePlus/commit/a32a9b93969525396d89dacd8ba037e96c4307e3) in TributePlus, a plugin which allows the player to capture most planets. By allowing planets to have multiple types of defence fleets, planets such as Greenwater have could both Hai and Merchant defence fleets.

I know tribute is extra content for the bored player, but this seems like it wouldn't be too hard to add.

Elyssaen:
Could be kind of useful, though I guess it's only a nice-to-have. Code-wise, it'd take a little bit of reorganising, but it's achievable.

It's one of these things that I'd guess MZ won't personally place as high enough priority (he's got plenty of other more important stuff), but there are a few people around who'd happily code it up and PR it for fun. I'm one of them, but will only do it if the idea has some measure of support. Another forgotten PR won't do anyone any good. : )

KiLEdEnNis:
the only thing it'd add is multiple governments defending a planet, but it seems a good idea and a bit realistic
ya tribute a planet, so the local militia, maybe some republic ships, and a couple of merchants defend

endless-sky:
This seems like a really low priority, especially given that if you really wanted to you could just create custom fleets that combine the different fleets you want to include.

tehhowch:
I think the only not possible bit with a combined fleet is use of multiple governments, e.g. Efret and Quarg ships would defend their stations in automata space, but the tribute action can only apply `atrocity` to one of them.
To work around that, one could create a combined government solely for this purpose (with the `attitude toward` values), but that gets really unwieldy and ends up with weird government names when you target the defending ships as well.

Issue #2318, opened 2017-03-26
Using tribute to "cheat" battles


Amazinite:
So I'm once again coming close to the end of the Wanderer campaign and I'm at the point where the massive Kor Sestor fleet attacks Farpoint. Instead of bombing Zenith, I decided to take out the Sestor fleet. I immediately realized that my 7 automata ships, 10 Raiders, and single Derecho wasn't going to cut it, so I parked my fleet and decided to instead take a different route; what if I could simply tribute Farpoint and essentially give myself reinforcements? Well, it works, and given how tribute works, all I need to do is leave the system and come back to retribute and get an entire new batch of Navy ships to use as cannon fodder. After tributing about a dozen times, the Navy was finaly able to take down the Sestor fleet, and entirely without me at that fact since I needed to stay cloaked the entire time.

A situation like this is probably not the only situation where this could be done. I imagine that while fighting on the side of the Navy or the Syndicate in the future, fighting on home field would give you a really good advantage by just tributing your own planet then either bribing your own planetary fleet or skidding your way past the system boundry and letting the battle play out. This can't be done right now for the FW story though because FW worlds have militia defence fleets, which aren't enemies to the Navy or Syndicate.

I think a way to fix this would be to have an event trigger at the beginning of the battle that disables the planet's defence fleet, similar to how system fleets are changed for a battle to assist the player's side.

endless-sky:
I suppose this could be fixed by making demanding tribute always reduce your reputation with a government to below 0. Then you wouldn't do it for frivolous reasons.

Amazinite:
Have fun everyone who pads their Syndicate rep by killing Korath ships then dominates all of Syndicate space while keeping them friendly.

Bladewood:
Make demanding tribute an atrocity?

Amazinite:
Makes sense to me.

endless-sky:
Fixed by making it an atrocity to successfully dominate a planet (089bc1c). Simply demanding tribute still just provokes the government, because someone might do it by mistake or just to see what will happen.

KiLEdEnNis:
maybe make it that you can't bribe tribute fleets (and their entire government when doing a tribite)

Bladewood:
... So, wait, if I demand tribute from the world to attempt OP's scenario, I just have to never talk to the world to keep my reputation?

tehhowch:
[...]

PR #3623, opened 2018-02-01
Support multiple fleet types in planet tributes

tehhowch:
Refs #2524 , #2318, 

 - Use an vector of stock fleet definitions to allow customized tribute battles
 - Increase difficulty of all tributes by changing the personality from `staying heroic nemesis` to `staying heroic marked unconstrained`
   - Tribute fleets will now only target the player's ships, and will be ignored by all other ships even if they would be "natural" enemies.
 - Increase visibility of tribute fleets in busy systems by applying the "target" personality
 - Correct a bug in the bounds check for tribute fleet definition
   - `fleet "Small Republic"` is currently skipped while `fleet "Small Republic" 1` is accepted. Both are valid and acceptable with this PR.
 - Print a warning if there is a custom fleet definition (i.e. child nodes to a node that starts with `fleet`)
 - If multiple mods alter a given planet's defense fleets, only the last modification is kept. This reproduces the behavior w.r.t. the tribute value / `master`, as only the last value is kept. If a mod only changes the amount / threshold, the standard fleet definition is kept.

With this PR, a tribute definition could then look like this:
```
planet "Gauntlet of Doom"
    tribute 1e6
        threshold 1e7
        fleet "Paradise Merchants" 2
        fleet "Navy Surveillance" 5
        fleet "Large Republic" 6
        fleet "Large Syndicate" 7
        fleet "Large Free Worlds" 8
        fleet "Large Pug" 9
        fleet "Large Quarg" 100
```
which would require the player to defeat fleets with the Merchant, Republic, Syndicate, Free Worlds, Pug, and Quarg governments (137 instantiated fleets in total in this example). These fleets would spawn in the order defined - first the Merchants, then the navy surveillance, then Republic, Syndicate, ... Quarg.

### DIFF
--- a/source/Personality.cpp
+++ b/source/Personality.cpp
@@ -364,7 +364,7 @@ void Personality::UpdateConfusion(bool isFiring)
 Personality Personality::Defender()
 {
 	Personality defender;
-	defender.flags = STAYING | NEMESIS | HEROIC;
+	defender.flags = STAYING | MARKED | HEROIC | UNCONSTRAINED | TARGET;
 	return defender;
 }
 
--- a/source/Planet.cpp
+++ b/source/Planet.cpp
@@ -164,17 +164,31 @@ void Planet::Load(const DataNode &node, const Set<Sale<Ship>> &ships, const Set<
 		else if(key == "tribute")
 		{
 			tribute = child.Value(valueIndex);
+			bool resetFleets = !defenseFleets.empty();
 			for(const DataNode &grand : child)
 			{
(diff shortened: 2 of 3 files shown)


tehhowch:
I had considered choosing a random fleet from the supplied fleets N times, but felt that such an approach could still be simulated with this approach (by using more nodes and smaller node sizes), but that approach would be unable to provide a designed battle. We don't currently wait very long between spawning of each defense fleet (ie defeating the existing fleets is not a prereq to spawning the next ones) but should such a change be made (or the internal lengthened), then the approach taken here is already ready for modder's creativity.

Issue #3114, opened 2017-10-12
New Tibet braver than Earth


ldsgsm:
My combat rating is level 10. I've been systematically conquered all of Republican space, except for New Tibet, which tells me "Please don't joke about that sort of thing" when I demand tribute. I don't mind this, but it seems to me very bizarre that Earth, the capital of human space, is willing to take my forces seriously but New Tibet, a sparsely populated planet on the edge of Republican space, is not.

tehhowch:
it is not an error.

ldsgsm:
Is there a reason it is set up that way? What makes New Tibet special?

tehhowch:
You ever hear of Tibetan monks?
You should also search for duplicate issues prior to opening new ones - there are more than a handful about the lack of tribute fleets on New Tibet and Harmony.

ghoulavenger:
Normally when a planet does this, it means it lacks a defense fleet. It shouldn't be too hard to make a plugin to give it one, if you truly want to dominate it. 

On the lore side, I disagree very heavily with why aliens and certain human worlds resist domination. If theres someone threatening to kill you, you take them seriously unless you want to die -- in this case, we could institute razing worlds (which destroys the spaceport, shipyards and outfitters -- trading and hiring of crew may not be available either, but if you didn't kill everybody there might still be someone there for that). On the game mechanics side though, dominating worlds doesn't really do much other than make you a pirate, which means it doesn't matter overly much.

Wrzlprnft:
given that a fleet which could raze a planet is kind of immersion breaking anyway, i wouldn't even bother, but that's just my opinion

ghoulavenger:
> given that a fleet which could raze a planet is kind of immersion breaking anyway, i wouldn't even bother, but that's just my opinion

You think so? Nuclear weapons are available, and they can completely decimate a world many times over, what makes you think that even a small fleet isn't capable of the same? Actually destroying the world or the sun is quite another matter -- although the technology for it supposedly exists in the lore (check out the Korath systems, talk to the Quarg).

Amazinite:
Duplicate of #1406

Quantumshark:
From the planet description of New Tibet
"In the lowlands, large farms support sprawling cities full of factories"
So it's not that the only inhabitants of the planet are the monks, there just are some monasteries there as well.

MCOfficer:
to me it seems like the monks rule the planet, but of course allow outsiders to settle there.
also not the "religious" attribute.

[...]

Issue #1406, opened 2016-08-06
New Tibet no tribute, threshold, or fleet


kitsunemochi:
Don't know if it's intentional or not, but there is nothing defined for New Tibet under the tribute mechanics.

KiLEdEnNis:
oh yeah, I think I saw that once (never reported)
I wish the Hai had tribute fleets though D:

kitsunemochi:
Just noticed since i'm in the neighborhood conquering the galaxy, Harmony has no tribute stuff either.  Maybe since these are religious planets?

Amazinite:
They're a peaceful planet of monks.

kitsunemochi:
seems like an easy mark, i'm not playing the good guy over here

Bladewood:
We were all aware of the existence of worlds which cannot give tribute, but we were all too lazy to report it to Michael.

kitsunemochi:
yea, by no means complaining.  just noticed it and searched and there was nothing here.  seems super simple to fix, copy/paste 3 lines to a few places, and i think it'd soothe a lot of my fellow completionists who want to dominate _all_ of human space.

Amazinite:
I don't think that we should be able to dominate _all_ of human space for the sake of lore like the peaceful monk planet, but you could download this plugin: https://github.com/Disiuze/TributePlus

It allows you to dominate every single inhabited planet in the game. May or may not be up to date with the Korath, haven't checked.

KiLEdEnNis:
with tribute it's basically give money or I'll do orbital bombardment, so the planets should have some defense, even if it's just 1-2 "local fleets" & gives 100 credits

Isaacssv552:
Is it possible to change the tribute stats of a planet by event? If not it would make sense that Deneb can't be dominated.

KiLEdEnNis:
it'd make sense that there would be some Navy defense fleet, but if that happens a player could dominate a planet when the defense fleet is the weakest (like Large Navy: 5 vs Large Pug: 20 for example)
and it could break the game by accident by dominating a planet that later changes defense fleet

TELunus:
Did we ever hear whether this was intentional, or just not worth fixing, or what?  I'm seeing that it's been closed, but no rational for not fixing it (or saying that it's fixed), and it still seems to be an issue in my version of the game.

KiLEdEnNis:
[...]

PR #3624, opened 2018-02-01
Harmony and New Tibet can pay tribute

Nescio0:
Currently Harmony and New Tibet are the only two inhabited planets in human space unable to pay tribute. This patch assigns a tribute of 400 to Harmony, which has "millions of people", and a tribute of 600 to New Tibet, which has "sprawling cities"; both are given an appropiate large militia defence fleet, allowing a player who behaves as 'a galactic terrorist' to dominate and land here as well.

For comparison the tribute of a few nearby planets:

New Portland (200): "only a handful of towns"
Hopper (300): "only a few small settlements"
Sinter (400): "the population of the entire planet is only a few thousand people"
Lichen (500): "Planetary government, or indeed any government above the village level, is virtually nonexistent."
New Argentina (1000): "The spaceport is on the outskirts of one of New Argentina's largest towns... which has a population of barely ten thousand people."

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -25152,6 +25152,9 @@ planet Harmony
 	spaceport `The spaceport is on the outskirts of one of the larger cities, with a few medium-sized skyscrapers and many tall smokestacks for factories and oil refineries. The starship hangars here are shaped like oversized barns, and close to the hangars, grain is stored in silos. The spaceport police are on horseback.`
 	spaceport `	Inside the building at the center of the port is a large cafeteria, with food of the simple "meat and potatoes" variety. You notice a few children and teenagers in rough, homemade clothing staring out the windows at the ships taking off and landing.`
 	security 0
+	tribute 700
+		threshold 3000
+		fleet "Large Militia" 12
 
 planet Haven
 	attributes pirate mining "north pirate"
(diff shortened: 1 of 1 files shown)


Amazinite:
#1406 #3114

Nescio0:
If there was a mechanism to freely land unopposed on pacifist planets it would not be a problem if some are unable to pay tribute. Unfortunately there isn't, hence the need to pacify them. Even a tribute of 0 would be better than the current situation.

tehhowch:
https://github.com/endless-sky/endless-sky/commit/2e9979753a8d106877f26f6e3f1b9a3ce3deeb7b

This behavior (no tribute offering from these planets) is intentional.

If you want, use a negative tribute for these planets, of 100,000 credits each, and give them a tribute fleet that is only unarmed ships. Then the player can pay reparations for their war crimes.

jafdy:
I would be more inclined to remove tribute altogether than expand on it.

ghoulavenger:
If this community were more inclined to work on tribute in general I wouldn't be against this change, since I believe realistically if someone threatens to bomb your planet, you fight them off or you suffer heavy losses. But since I get the feeling that this community largely agrees with jafdy, I think you should just keep this as your own plugin for now.

PR #3625, opened 2018-02-02
Hyperspace code cleanup and minimum entry distances

accursedCursive:
**Code cleanup**
The hyperspace code was moved out of the Move() function, structured to better reflect the sequence of events (separated into intial, repeated, and transitional hyperspace events), and worded to be less vague.

**Functionality changes**
New features:
"border" attribute for systems, representing a region that a ship entering with hyperspace travel is unable to pass through. Hyperspace travel ends before this border is reached, and jumping ships are placed outside the border.
This border also affects the distribution of pre-existing fleets in a system, and adds to the maximum distance an AI can venture from the system centre.
To address numerous complaints of users entering a system only to immediately be in the firing range of more powerful ships, the border is set to 1000 by default, as most systems have occasional enemy ships and thus require a decent-sized border.

Alterations:
Hyperdrive travel ends when a ship is travelling below maximum speed, and slow enough that it could stop before reaching the target from its current location. This replaces the system where hyperspace travel would end when the ship was travelling slow enough to stop within 1000 units.
Jump Drive no longer scatters ships randomly: Positions in a fleet are retained, and a fleet will attempt to jump directly to a target, though the target will likely be within the border and thus the entry point will be moved.

Bugfixes:
Ships that are more than 1000 units away from yours in a hyperspace jump will no longer teleport closer, and instead smoothly glide closer. This effect might be a bit strong at long separation distances, feel free to tweak this part.

New issues:
As a "cruise" stage of hyperspace travel was necessary allow borders to affect each ship in a fleet individualy while still placing the entire fleet in one location, ships spend longer travelling at maximum hyperspace speed, which causes a framerate drop in asteroid-heavy systems (this was already an issue, now it's worse).


accursedCursive:
I'm not sure why it's saying that I've replaced identical code.
The changes total only two or three hunded lines, not three thousand.

accursedCursive:
Only style guide violation I could find was that some all-caps variables were not technically constants, due to a g++ bug where attempting to make them constants would cause the compiler itself to crash.

As for your method of restoring whitespace (which is missing due to a Notepad++ bug), it'd be much faster if you did that, since what you suggested is a rather obscure use of Github's functionality.
I can fix the other stuff afterwards, it's just a matter of replacing every quadruple space with a tab and naming the commits with the updates files.

jafdy:
I like the sound of most of this, however I can't vouch for the workability of the code. 

Changing the default entry distance will reduce the chance of this PR merging as it stops being a strictly under-the-hood change, PRing a change to the default distance after this is merged would be better.

A question: What does a border or 1000 mean?
A compliment: Moving ships closer instead of teleporting them is a nice touch.
A suggestion: During the "cruse", what about removing any asteroids (this is interstellar space). That would help the frame rates.

Finally, making changes by file upload has a habit of making trouble, most of us use GitHub Desktop for this reason. And fixing the whitespaces are your responsibility, @tehhowch is just telling you it needs to be done (if you don't do it then this won't get merged).

tehhowch:
>a g++ bug where attempting to make them constants would cause the compiler itself to crash

I am very confident you were doing something incorrect and were receiving compiler **errors** that explicitly stated what you were doing wrong, rather than you discovered a bug that crashed the GNU C++ compiler. Perhaps you were declaring them constant and then assigning to them later (which would mean the value is in fact ***not*** constant)?

>bug in Notepad++

No, you didn't turn off its *setting* that trims trailing whitespace. That's not a bug in the program, that's it following the protocol you said it needs to follow. Own up to your mistakes instead of blaming them on a program error.

As jafdy says, it is your responsibility to remove unwanted whitespace changes that you introduce, not ours.
Since you say there are only a few hundred lines of changes, making a new branch and applying only your changes should be very simple. If you continue to use the web interface, I suggest making liberal use of the "Preview Changes" tab next time so you are aware of the content of each commit and can name it appropriately. (If you had used that, you would have caught the whitespace errors before committing them.)

Once you fix the whitespace errors you made, we can review the actual content of your pull.

accursedCursive:
They were definitely g++ errors.
Plus, there appears to be no setting in Notepad++ that automatically trims trailing whitespace, just a bunch of tools to do it on command which I never used once let alone in every file I edited. Looking up anything to do with trimming trailing whitespace just comes up with people asking how to do it, which indicates that my copy of Notepad++ doing it by default is a bug as no one would be asking those questions if Notepad++ does not do it by default.

KiLEdEnNis:
did you press save in CodeBlocks somewhere when creating this... PR?

PaxIllyricum:
Disregarding whitespace, I've tried building your PR and [got a build error](https://pastebin.com/ivYZhJVk). I recommend you fix your whitespace so other people can review your changes and then use updated source files to work with. Until then, expect silence from others.

warp-core:
g++ gives you errors because your code doesn't adhere to the rules of the language. Meaning that your code is wrong and broken.
As for the trailing whitespace thing, the cause is not really relevant. What's important is that it has happened and it needs to be fixed before anyone is going to review the changes you've proposed because, as is, the changes are far more difficult to identify than they should be.

tehhowch:
>They were definitely g++ errors.

Indeed, `g++` very likely emitted a stream of errors, every one of which flagged a line in your code and then stepped back through all the underlying code to tell you exactly why your code was incorrect. Feel free to screenshot the errors here if you would like us to confirm that indeed, the code you wrote was incorrect and the compiler is simply doing its job.

As @PaxIllyricum indicates, you have managed to revert several commits with this PR, resulting in a build error in the main repository (you refer to methods in other files that do not exist). The source files which you are building are likely out-of-date files (i.e. you are not building from a `git` repository, but rather a static dump of the game files from some earlier date), which is why you don't observe this build error in your local copy.

accursedCursive:
None of the errors are to do with my own code:
capture

As for the issues you're having with building, this shouldn't revert any commits. After I was done, I manually merged my changes with the latest versions of all files I changed.

Turns out that it was Codeblocks deleting tabs in empty lines, by the way.
Annoyingly, text editor programmers hate tabs in empty lines so much that they provide a million ways to remove them and no ways to restore them.

warp-core:
Those errors are happening with your code and not the master branch. The only thing that's changed is the changes you're proposing.
Maybe just pay more attention to and be more careful with the whitespaces in the first place and you won't have this issue.
We can't really be certain about the issue of reverting other commits, though it does seem likely, until you fix the whitespace issue so we can actually identify your changes quickly.
(The fact that GitHub is giving a merge conflict suggests that an issue with uploading a file that isn't up to date with changes to the master branch is an issue though.)

warp-core:
That last change seems a little beyond the scope of this PR and hasn't addressed any of hte issues that have been raised so far.

KiLEdEnNis:
temp13
temp14
Now you've done it, there goes the ~fourth wall~ pull request.

accursedCursive:
That was a mistake, I'm trying to revert it but can't due to something that genuinely is a bug other users have reported.
Problem is, Github won't let me have some clientside commits that allow me to avoid the unclean setup the project currently has.

accursedCursive:
I'm just going to create a new fork then apply my changes to it, that'll solve all the issues I've had.

KiLEdEnNis:
~did you try turning it off and back on again~

fork the Endless Sky repo & download the GitHub desktop tool (think there were multiple ones, but from what I understand they are mostly for the same stuff)
in that desktop app, clone the repository ya forked
then edit the files in that cloned fork, if edits are saved ya see them in a list like this (example)
temp15
ya can select what files or even what lines to commit & choose the commit name (means ya can also review the changes while doing so), then ya press commit to master
once ya press sync in the right top ya upload the commits ya made to your fork, and you make a pull request with your fork

Wrzlprnft:
If you want to learn more about git, there are plenty of resources online. if you need more specific help, feel free to add me on steam or discord, i should be able to clear up versioning system problems.

Issue #3620, opened 2018-02-01
[Feature request] Auto visit spaceport and auto buy map options


clintonmead:
First of all, love the game, got a Bactrian and played through a good part of the Free Worlds war (didn't finish it though). Anyway now I've got the mechanics I'm thinking of doing a entire playthrough after the next release.

Enough about me, there's just two things I noticed I was doing upon landing everytime.

1. Checking the starport.
2. Buying the local map if I didn't have it complete already (as it's so cheap). 

I would like to add two options, namely "auto buy local map" and "auto check starport". So the first thing that will happen on landing is buying the local map if it's available in outfitting and your version is not complete, and then checking the starport. 

Of course these should be in preferences, and by default off, because there may be reasons why players don't want to do this.

I might get around with doing this myself at some point, by submitting a PR, but I'm not sure when I'll have the time to do that (I'd have to understand the code base a bit first) so I thought I'd drop it here for now incase you want to include it in your next release.

KiLEdEnNis:
it doesn't seem worth it to clutter the preferences with 2 options like this:
with the upcoming new spaceport system (the portraits which quite some people apparently dislike, including me)  it's not a good idea to do this at all
same reason for auto-buying a certain outfit upon landing; not worth to add new preferences for a tiny bit of automation

jafdy:
So this will come under the heading "too much automation" additionally any option that is disabled by default is unlikely to be added. In short, don't get your hopes up. Then again, you never know.

clintonmead:
In any case I’ll submit a PR sometime and if the devs don’t want to put it in mainline I’ll just offer it as a fork.

I was just inquiring because if the devs were interested I’d write it sooner but if they’re not I’ll just fork off the next release and anyone who is interested can use my fork.

KiLEdEnNis:
personally I'd wait for the *dev (singular, one man) to tell me if this feature would be worth it, but feel free to create it if ya want

clintonmead:
Indeed that’s what I meant KiLEdEnMis.

If the dev is interested I’ll try to get it in the main source otherwise sooner rather than later I’ll just write patch on the new version and host the fork on my own GitHub for whoever wants the feature to be able to use it.

tehhowch:
As I mention in your PR, moving the map info check into outfitter (to be like the refill ammo check) would be fine by me, and would let you only offer the option if the outfitter has a map, too (and reference the strength of the map as well).

[...]

PR #1292, opened 2016-07-03
add indicator for spaceport missions

MarcelineVQ:
Added an indicator to hint to the player when checking the spaceport will trigger a mission.
It's very easy to land on a planet and not check the spaceport. It's especially easy to not even know you need to check the spaceport. When you miss the opportunity to do this for a randomly given mission you have to wait until it hits the right random number again.
This seemed like a useful addition to alleviate some possible player frustration, I personally ran into the issue trying to start the freeworlds intros and get the story rolling because they rely on randomness and checking the spaceport panel.

alert

PR #3626, opened 2018-02-02
Added alerts for available starport missions and local maps

clintonmead:
As discussed in [issue 3620](https://github.com/endless-sky/endless-sky/issues/3620), here's a pull request implementing the following alerts (although the user must still take the actual action):

1. Alert when there is a mission waiting in the starport. (preference.txt option "Spaceport mission alert")
2. Alert when there is new local map information available in outfitting (preference.txt option "Local map available alert")

Note that the alerts to buy maps have the distance hard coded, instead of reading from the data file. As couldn't work out how to read it from the data file.

My pull request also has a bunch of whitespace changes, I couldn't entice my editor not to include them annoyingly.

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -18,6 +18,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "DataNode.h"
 #include "DataWriter.h"
 #include "Dialog.h"
+#include "DistanceMap.h"
 #include "Files.h"
 #include "Format.h"
 #include "GameData.h"
@@ -53,7 +54,7 @@ using namespace std;
 void PlayerInfo::Clear()
 {
 	*this = PlayerInfo();
-	
+
 	Random::Seed(time(nullptr));
 	GameData::Revert();
 	Messages::Reset();
@@ -74,7 +75,7 @@ void PlayerInfo::New()
 {
 	// Clear any previously loaded data.
 	Clear();
-	
+
 	const StartConditions &start = GameData::Start();
 	// Copy any ships in the start conditions.
 	for(const Ship &ship : start.Ships())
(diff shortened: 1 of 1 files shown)


jafdy:
The first (`Spaceport mission alert`) has been noed it the past.

Amazinite:
>The first (Spaceport mission alert) has been noed it the past.

Reference: #1292

source/PlayerInfo.cpp
```diff
@@ -991,22 +995,45 @@ void PlayerInfo::Land(UI *ui)
...
+
+  // Alert when there's spaceport missions available
+  if (Preferences::Has("Spaceport mission alert") && hasSpaceport && MissionToOffer(Mission::SPACEPORT))
+  {
+    Messages::Add("There is a spaceport mission available.");
+    ui->Push(new Dialog("There is a spaceport mission available."));
+  }
+
+  // Alert when you can buy a map
+  if (Preferences::Has("Local map available alert") && hasOutfitter)
```
tehhowch: `if(Preferences.....`

source/PlayerInfo.cpp
```diff
@@ -991,22 +995,45 @@ void PlayerInfo::Land(UI *ui)
...
+  {
+    // Hard coded distance, this is ugly.
```
tehhowch: You've used spaces and not tabs.

clintonmead: The whole point of this is to display on the main screen so you don't have to mindlessly click the outfitter. 

It's actually probably a good idea to do an ammo alert as well. 

I think however the approach pull request [#1292](https://github.com/endless-sky/endless-sky/pull/1292) takes is better though, with the red exclamation point on the starport.

I'd like to the same for the outfitter with local maps, and as you mentioned, it's probably worth adding ammo refills. But I need this info before entering the outfitter, that's the whole point.

Unfortunately, as far as I can tell, it seems the outfitting logic, like "CanBuy()", is tightly coupled with the outfitting GUI, and you can't use the logic from outside the GUI. Although perhaps I've missed something.

clintonmead:
I'm closing this pull request because it seems the spaceport alert is only working for mainline missions anyway, not random missions.

I think the approach used in [#1292](https://github.com/endless-sky/endless-sky/pull/1292) is better. I'll try that approach, with both the spaceport and map info, and perhaps even a reloading alert. But I'll make a new PR for that.

mootootwo:
I like that you have made it configurable in preferences.  If the spaceport alert defaults to off, it will be more likely to pass approval where 1292 failed.

PR #3628, opened 2018-02-02
Hyperspace code cleanup and minimum entry distances

accursedCursive:
Attempt 2: Electric Boogaloo
Now it should be much cleaner.

**Code cleanup**
The hyperspace code was moved out of the Move() function, structured to better reflect the sequence of events (separated into intial, repeated, and transitional hyperspace events), and worded to be less vague.

**Functionality changes**
New features:
"border" attribute for systems, representing a region that a ship entering with hyperspace travel is unable to pass through. Hyperspace travel ends before this border is reached, and jumping ships are placed outside the border.
This border also affects the distribution of pre-existing fleets in a system, and adds to the maximum distance an AI can venture from the system centre.
To address numerous complaints of users entering a system only to immediately be in the firing range of more powerful ships, the border is set to 1000 by default, as most systems have occasional enemy ships and thus require a decent-sized border.

Alterations:
Hyperdrive travel ends when a ship is travelling below maximum speed, and slow enough that it could stop before reaching the target from its current location. This replaces the system where hyperspace travel would end when the ship was travelling slow enough to stop within 1000 units.
Jump Drive no longer scatters ships randomly: Positions in a fleet are retained, and a fleet will attempt to jump directly to a target, though the target will likely be within the border and thus the entry point will be moved.

Bugfixes:
Ships that are more than 1000 units away from yours in a hyperspace jump will no longer teleport closer, and instead smoothly glide closer. This effect might be a bit strong at long separation distances, feel free to tweak this part.

New issues:
As a "cruise" stage of hyperspace travel was necessary allow borders to affect each ship in a fleet individualy while still placing the entire fleet in one location, ships spend longer travelling at maximum hyperspace speed, which causes a framerate drop in asteroid-heavy systems (this was already an issue, now it's worse).

**Purpose of change**
The purpose of the code cleanup is to improve the understandability and moddability of hyperspace code by making it follow a predictable structure instead of a random structure that breaks when changed.

The purpose of the border is to improve gameplay quality by adding a new parameter to systems.
A smaller border (like the game is now) makes non-combat gameplay better, by keeping everything in a convenient distance of the ship.
A larger border makes combat better by having new ships not arrive within attack range of already present ships, thus adding more emphasis to ship traits other than health and damage.
As such, the size of a border should reflect how significant a role combat plays in that system.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -322,11 +322,16 @@ void AI::Step(const PlayerInfo &player)
 			++it;
 	}
 	for(const auto &it : ships)
-		if(it->Position().Length() >= MAX_DISTANCE_FROM_CENTER)
+    {
+	    double border = 0;
+	    if(it->GetSystem())
+            border = it->GetSystem()->Border();
+		if(it->Position().Length() >= MAX_DISTANCE_FROM_CENTER+border)
 		{
 			int &value = fenceCount[&*it];
 			value = min(FENCE_MAX, value + FENCE_DECAY + 1);
 		}
+    }
 	
 	const Ship *flagship = player.Flagship();
 	step = (step + 1) & 31;
@@ -1097,7 +1102,7 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
 	if(target && !ship.IsYours() && !ship.GetPersonality().IsUnconstrained())
 	{
 		Point extrapolated = target->Position() + 120. * (target->Velocity() - ship.Velocity());
-		if(extrapolated.Length() >= MAX_DISTANCE_FROM_CENTER)
+		if(extrapolated.Length() >= MAX_DISTANCE_FROM_CENTER+ship.GetSystem()->Border())
 		{
 			MoveTo(ship, command, Point(), Point(), 40., .8);
 			if(ship.Velocity().Dot(ship.Position()) > 0.)
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -322,16 +322,16 @@ void AI::Step(const PlayerInfo &player)
 			++it;
 	}
 	for(const auto &it : ships)
-    {
-	    double border = 0;
-	    if(it->GetSystem())
-            border = it->GetSystem()->Border();
+	{
+		double border = 0;
+		if(it->GetSystem())
+			border = it->GetSystem()->Border();
 		if(it->Position().Length() >= MAX_DISTANCE_FROM_CENTER+border)
 		{
 			int &value = fenceCount[&*it];
 			value = min(FENCE_MAX, value + FENCE_DECAY + 1);
 		}
-    }
+	}
 	
 	const Ship *flagship = player.Flagship();
 	step = (step + 1) & 31;
@@ -2730,11 +2730,11 @@ double AI::RendezvousTime(const Point &p, const Point &v, double vp)
 	// to intersect the target?
 	// (p.x + v.x*t)^2 + (p.y + v.y*t)^2 = vp^2*t^2
 	// p.x^2 + 2*p.x*v.x*t + v.x^2*t^2
-	//    + p.y^2 + 2*p.y*v.y*t + v.y^2t^2
-	//    - vp^2*t^2 = 0
+	//	+ p.y^2 + 2*p.y*v.y*t + v.y^2t^2
+	//	- vp^2*t^2 = 0
 	// (v.x^2 + v.y^2 - vp^2) * t^2
-	//    + (2 * (p.x * v.x + p.y * v.y)) * t
-	//    + (p.x^2 + p.y^2) = 0
+	//	+ (2 * (p.x * v.x + p.y * v.y)) * t
+	//	+ (p.x^2 + p.y^2) = 0
 	double a = v.Dot(v) - vp * vp;
 	double b = 2. * p.Dot(v);
 	double c = p.Dot(p);
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -721,7 +721,7 @@ void Ship::Place(Point position, Point velocity, Angle angle)
 	slowness = 0.;
 	isInvisible = !HasSprite();
 	jettisoned.clear();
-	hyperspaceCount = 0;
+	hyperCount = 0;
 	forget = 1;
 	targetShip.reset();
 	shipToAssist.reset();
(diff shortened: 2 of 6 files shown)


source/AI.cpp
```diff
@@ -322,11 +322,16 @@ void AI::Step(const PlayerInfo &player)
 			++it;
 	}
 	for(const auto &it : ships)
-		if(it->Position().Length() >= MAX_DISTANCE_FROM_CENTER)
+    {
```
tehhowch: You still have spaces for indentations in many places.

source/AI.cpp
```diff
@@ -322,11 +322,16 @@ void AI::Step(const PlayerInfo &player)
...
+	    double border = 0;
+	    if(it->GetSystem())
+            border = it->GetSystem()->Border();
+		if(it->Position().Length() >= MAX_DISTANCE_FROM_CENTER+border)
```
tehhowch: You need spaces between binary operators like `+`

Additionally, I don't think it is correct to add these values. You likely want to consider only the maximum value.

source/Ship.cpp
```diff
@@ -721,7 +721,7 @@ void Ship::Place(Point position, Point velocity, Angle angle)
 	slowness = 0.;
 	isInvisible = !HasSprite();
 	jettisoned.clear();
-	hyperspaceCount = 0;
+	hyperCount = 0;
```
tehhowch: Why rename this variable? `hyperCount` is less clear than `hyperspaceCount`

source/Ship.cpp
```diff
@@ -1368,6 +1264,152 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 
 
 
+// Performs one step of the hyperspace animation. (This is called by Move().)
+void Ship::IterateHyperspace(vector<Visual> &visuals)
+{
+	// Don't apply external acceleration while jumping.
+	acceleration = Point();
+	
+	// Initial segment
```
tehhowch: Comments are to be full sentences.

source/Ship.cpp
```diff
@@ -1368,6 +1264,152 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
...
+				if(object.GetPlanet() && object.GetPlanet()->HasSpaceport())
+				{
+					hyperTarget = object.Position();
+					break;
+				}
+		}
+	}
+	
+	// Repeated segment
+	int hyperspaceRate = hyperspaceSystem ? 1 : -1; // Postive values move towards transition, negative values move away
```
tehhowch: Inline comments do not meet the style guide.

source/Ship.cpp
```diff
@@ -1368,6 +1264,152 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
...
+		fuel -= hyperspaceFuelCost / hyperSteps;
+	
+	if(GetParent() && GetParent()->currentSystem == currentSystem)
+	{
+		hyperOffset = position+velocity - GetParent()->position; // Allows formation to be held through hyperspace
+		double length = hyperOffset.Length();
+		if(length > fleetRadius) // Condense fleet around leader
+		{
+			Point offsetFromDesired = hyperOffset-hyperOffset*(fleetRadius/length);
+			position -= offsetFromDesired*((0.05+0.05*hyperCount)/hyperSteps); // This calculation is arbitrary, and chosen to look good in-game
```
tehhowch: Operator spacing

source/Ship.cpp
```diff
@@ -1368,6 +1264,152 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
...
+	{
+		if (hyperCount <= hyperSteps){ // If false, the ship is a cruising stage of hyperspace travel
+			velocity += (hyperAcceleration * hyperspaceRate) * angle.Unit();
+		}
+
+		if(velocity.Length() <= MaxVelocity() && !hyperspaceSystem)
+		{
+			double stoppingDistance = .5 * velocity.LengthSquared() / Acceleration(); //Distance ship travels while decelerating
+			double turnDistance = velocity.Length() * 150. / TurnRate(); //Distance ship travels before it can start decelerating
+			if (stoppingDistance+turnDistance <= (hyperTarget-position).Length())
```
tehhowch: `if(`, not `if (`
Operator spacing.

source/Ship.cpp
```diff
@@ -1368,6 +1264,152 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
...
+		// traveling in, so that when you decelerate there will not be a
+		// sudden shift in direction at the end.
+		velocity = velocity.Length() * angle.Unit();
+		
+		double borderRadius = currentSystem->Border();
+		
+		// Hyper Drive Transition
+		if(!isUsingJumpDrive)
+		{
+			Angle inverseAngle = Angle(-angle.Degrees());
```
tehhowch: ?

source/Ship.cpp
```diff
@@ -1368,6 +1264,152 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
...
+			double hyperStopDistance = (hyperSteps * hyperSteps) * .5 * hyperAcceleration;
+			double cruiseDistance = max(0.0, borderRadius-targetRotated.Y()+fleetRadius); // The furthest a ship could possibly have to be set back by to decelerate before the border
+			
+			position = (hyperTarget - (hyperStopDistance+cruiseDistance) * angle.Unit());
+			position += hyperOffset;
+			
+			Point offsetTargetRotated = inverseAngle.Rotate(hyperTarget+hyperOffset); // Like targetRotated, but taking into account the offset of the target.
+			if(abs(offsetTargetRotated.X()) < borderRadius)
+			{
+				cruiseDistance -= max(0.0, borderRadius*cos(asin(offsetTargetRotated.X()/borderRadius))-offsetTargetRotated.Y()); // How far into the border the ship has to go to reach its target
```
tehhowch: Why are you using trig functions instead of vector math?

source/Ship.h
```diff
@@ -469,11 +471,15 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	// hyperspacing, and exploding. Each one must track some special counters:
 	const Planet *landingPlanet = nullptr;
 	
-	int hyperspaceCount = 0;
-	const System *hyperspaceSystem = nullptr;
-	bool isUsingJumpDrive = false;
+	int hyperSteps;
+    double hyperAcceleration;
```
tehhowch: More spaces that should be tabs.

source/System.h
```diff
@@ -181,6 +183,7 @@ class System {
 	std::vector<FleetProbability> fleets;
 	double habitable = 1000.;
 	double asteroidBelt = 1500.;
+	double border = 1000.;
```
tehhowch: This should be `400.` not `1000.`

source/Ship.cpp
```diff
@@ -1368,6 +1264,152 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
...
+		{
+			Point offsetFromDesired = hyperOffset-hyperOffset*(fleetRadius/length);
+			position -= offsetFromDesired*((0.05+0.05*hyperCount)/hyperSteps); // This calculation is arbitrary, and chosen to look good in-game
+		}
+	}
+	
+	//Hyper Drive Repeat
+	if(!isUsingJumpDrive)
+	{
+		if (hyperCount <= hyperSteps){ // If false, the ship is a cruising stage of hyperspace travel
```
tehhowch: Inline comment
Braces must always be on a new line following parentheses.
Braces are also optional for single-statement blocks such as this one.

source/Ship.cpp
accursedCursive: Consistency: A lot of variables needed to be prefix with "hyperspace" to remove ambiguity, but formulas were becoming very long with the extensive variable names, so those variables were instead just prefixed with "hyper". `hyperspaceCount` was the only exception to the prefixing pattern, so to improve code consistency it was changed.

source/System.h
accursedCursive: 1000 is the sweet spot where it's not a nuisance in systems that have little combat, but point-blank engagements are still usually avoided.
It's surprisingly close-range, you can't really have the border get any smaller before the currently-existent gameplay issues return.

source/Ship.cpp
tehhowch: Use multiple lines.
General style used here is a double indent relative to the first line.
```
this = other
        + stuff
        + more stuff
```

source/System.h
tehhowch: I would save changing the default value for every person who plays the game for its own discussion.

source/Ship.cpp
accursedCursive: Multiple lines for what?
There's no formulas in this code long enough to justify the usage of multiple lines.

jafdy: You just said they where getting very long, and now you are saying that they are not long enough to use more than one line.

source/Ship.cpp
jafdy: Anytime you see a cos(arcsin()) or a sin(arcos()) there is almost always a better way to do it. In all my time doing maths there has never been a good reason to use either of those, the only time they might appear is while integrating.

source/System.h
jafdy: This is a balance change so should get it's own PR separate from this under-the-hood change. Changing the default will make this a hard PR to merge.

jafdy:
What does a border of 1000 mean?

source/System.h
accursedCursive: You can always edit and merge. But I strongly suggest you keep it at 1000, and then choose peaceful systems to have smaller borders. It's not an arbitrary number, it really is what's best.

accursedCursive:
A border of 1000 means that any ship entering the system will exit hyperspace before coming within 1000 units of the star.

source/Ship.cpp
mattsoulanille: [cos(arcsin(x)) = sqrt(1-x^2)](http://www.wolframalpha.com/input/?i=cos(arcsin(x)))

geojak:
thumbs up for a real attempt (not only talking about it as many including me do) in fixing a existing issue in the game (atleast an issue for me in my play experience of ES)

even if some don't like the idea of the boarders. this might still be worth a merge for the additional functionality for personal mods and game changes.

Issue #3149, opened 2017-10-22
[bug] turrets stop rotating when out of range(?)


KiLEdEnNis:
first of all, checked for duplicates and commits since I just run the latest Steam beta version, but I haven't seen any issue/commit that would have fixed this

basically how to "duplicate" it (and explains what happens):
- I have an escort hold position, and fire at it with turrets
- I basically fly right to it until I pass through it and it's behind me, then rotate back to the escort
- my turrets don't rotate, they just freeze and shoot at a certain direction, which I screenshotted

![temp5](https://user-images.githubusercontent.com/10348657/31862134-a247957e-b739-11e7-8e15-8ec536f40a2d.png)
I also recorded this bug on plays.tv (basically my only way to record something): http://plays.tv/video/59ec97a15cf63aeccb/endless-sky-turrets-freezing

as you can see, they basically shoot backwards, and I am 100% sure they aren't rotating at all
if I move closer, or basically stay close (not even sure)

again: latest Steam beta, I couldn't find anything that could have fixed it but I may have missed it

endless-sky:
Does this only apply to shooting at your escorts?

KiLEdEnNis:
no, it's a bit harder to test on AI ships, but when I shot a random Merchant Bulk Freighter and moved away a bit, the same issue happened

endless-sky:
Are you using opportunistic firing, or focused? I still can't seem to duplicate this.

KiLEdEnNis:
I'll look later today when I get home (in about... 8 hours), but I think focused (when I got swarmed by the Republic, my turrets still froze facing a certain direction)
when I know for sure I'll just edit it in this reply

KiLEdEnNis:
focused fire, but opportunistic gives this too
since it can't be duplicated, I'll leave this open until next Steam update (if only I knew how to build from source)
if the next release "fixes" this for me I'll close it then

tehhowch:
I've definitely seen this before, particularly when I was testing my ship list PR and had issued the board command for the current target, and was also holding the fire command - the ship would lock its turrets in a given direction (lasers too), even as the ship rotated to prepare for boarding. In that case it was with a hostile ship rather than an escort.

KiLEdEnNis:
I think that's not the same bug:
yours happens with ALL turrets when boarding a (hostile?) ship, which makes me think here that turrets don't listen to commands while the ship is doing so
[...]

PR #3630, opened 2018-02-04
Always track a target, even if out-of-range

tehhowch:
Refs #3149 

 - Ensure that at least one target will always be aimed at, by increasing the starting score to +Inf
 - Handle NaN returns from RendezvousTime

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1623,7 +1623,7 @@ void AI::Swarm(Ship &ship, Command &command, const Ship &target)
 	Point direction = target.Position() - ship.Position();
 	double maxSpeed = ship.MaxVelocity();
 	double rendezvousTime = RendezvousTime(direction, target.Velocity(), maxSpeed);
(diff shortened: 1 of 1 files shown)


endless-sky:
It might be necessary to use `std::isnan()` instead of just `isnan()`, because some compilers / runtime libraries have an `isnan()` function in the global namespace and will complain about an ambiguous function call.

For the tracking range, maybe 1000 pixels isn't a high enough limit, but if a ship's turrets are in opportunistic tracking mode to me it looks weird for them to all end up focused on a target that's on the other side of the star system.

tehhowch:
I've updated the `isnan` to include the `std::` prefix.
As far as an opportunistic turret aiming at a very far away hostile, that hostile would only make it into `enemies` if it is within 1.5x the weapon range:
https://github.com/endless-sky/endless-sky/blob/784d96f5ea91f240a868f356ad1235eb06f1516e/source/AI.cpp#L2358-L2384

Issue #2937, opened 2017-08-15
Fighters/drones do not follow through wormholes


tehhowch:
Any ship which does not have a hyperdrive / jump drive will not route through a wormhole to reach its parent's system.

Relevant bits from AI::MoveEscort
```
	bool hasFuelCapacity = ship.Attributes().Get("fuel capacity") && ship.JumpFuel();
	bool isStaying = ship.GetPersonality().IsStaying() || !hasFuelCapacity;
...
	else if(!parentIsHere && !isStaying)
	{
		...
		// Check whether the ship has a target system and is able to jump to it,
		// and that the targeted stellar object can be landed on.
		bool hasJump = (ship.GetTargetSystem() && ship.JumpFuel(ship.GetTargetSystem()));
		if(ship.GetTargetStellar()
				&& !(ship.GetTargetStellar()->GetPlanet() && ship.GetTargetStellar()->GetPlanet()->CanLand(ship)))
			ship.SetTargetStellar(nullptr);
		if(!hasJump && !ship.GetTargetStellar())
		{
			// If we're stranded and haven't decided where to go, figure out a
			// path to the parent ship's system.
			DistanceMap distance(ship, parent.GetSystem());
			const System *from = ship.GetSystem();
			const System *to = distance.Route(from);
			for(const StellarObject &object : from->Objects())
				if(object.GetPlanet() && object.GetPlanet()->WormholeDestination(from) == to)
				{
					ship.SetTargetStellar(&object);
					break;
				}
			// Only set a target system if there is no wormhole route to that system.
			if(!ship.GetTargetStellar())
			{
				ship.SetTargetSystem(to);
				// Check if we need to refuel. Wormhole travel does not require fuel.
				if(to && systemHasFuel && !to->HasFuelFor(ship) && ship.JumpsRemaining() == 1)
					Refuel(ship, command);
			}
		}
```
Player escorts will follow if and only if they are able to wait on the wormhole while the player is still approaching it, as they fall through the if-else if chain to reach
```
else if(parent.Commands().Has(Command::LAND) && parentIsHere && planetIsHere && parentPlanet->CanLand(ship))
	{
		ship.SetTargetStellar(parent.GetTargetStellar());
		MoveToPlanet(ship, command);
		if(parent.IsLanding() || parent.CanLand())
			command |= Command::LAND;
	}
```
[...]

PR #3635, opened 2018-02-06
Add a "no hyperdrive" flight check.

jostephd:


tehhowch:
That's not quite true though - a ship can use a wormhole just fine without a Hyperdrive or jump drive (barring some caveats (currently))
Related https://github.com/endless-sky/endless-sky/issues/2936, https://github.com/endless-sky/endless-sky/issues/2937

Any plugin which doesn't rely on hyperspace for its civilization would trigger this warning display for every ship with no conceivable way to disable the warning except by giving ships hyperdrives.

jafdy:
Isn't there a warning if you try to play a route without a drive that can handle it? (An interstellar route without wormholes in this case)

jostephd:
The in-flight warning doesn't help when I remove the hyperdrive from an escort, park it, and unpark it the next day when the flagship at another system.

Wormholes could be supported by rephrasing the warning or disabling it when the system has a wormhole.

No idea how to support plugins that don't use hyperspace.

mootootwo:
There are valid reasons to strip a hyperdrive out of a ship.  For example, to maximize space available when you have a job in the current system, or something to loot in the current system.  Or when stripping a hull for parking.

jostephd:
I agree and that's why it's only a warning that doesn't block takeoff.

PR #3636, opened 2018-02-07
Remnants now sing and chant more and talk less

Tadrix:


Darcman99:
In the discord certain humans have pointed out the fact that the prefect has not been singing in consistency with other Remnant dialog but whether or not this is because people have pointed it out in there, I don't know.

Bladewood:
I'm not going to explicitly express my position on this, but there's many perspectives on these kinds of issues: `he says`, `she replies`, `they respond`, `they state`. I'll copypasta what I wrote:

One such perspective is, it basically doesn't matter to vary the verbiage, because it'll only end up being a burden or annoyance to the writer and to readers - which is the opposite of what we want.
But another perspective is, if we keep using the same one over and over, `he says`, `she says`, it'll become repetitive and dull.

A user mentioned the idea that, "Just because they traditionally sing, doesn't mean they're incapable of simply talking." I don't know if they're totally incapable, but I think they'd sing one way or another.
The same user said that "advanced cultures" would adapt to whatever race they were communicating with. I disagree with this notion. Aliens will be aliens, and it would have been explicitly brought up if they spoke normally, as with Rek during the Molt.

Here's another way to look at the Remnant issue:
- "say" is used for normal, everyday communication (which typically involves singing or sign language);
- "chant" is used for very religious or [autonomous, droning] recitals;
- "sing" is used for particularly high-pitched, graceful, or cheerful situations.

In the case of the Remnant, the whole point is, it's implicit that they're singing.
`says`, `questions`, `asks`: they're basically all saying the same thing, just differently.
There have been discussions on it in the past, and not everyone has reached the same conclusion.
Some writers would just prefer to stick solely to "he says", except in very special situations.
Others would prefer to vary the prose, mixing in "responds" and "replies" with "says".

Tadrix:
Nothing wrong with `replies`, `respond`, `state`, `questions`, `asks`, one can do all these things in song (so there's no issue with these verbs). All this PR does is replace `say` with `sing` or `chant`.

endless-sky:
The nice thing about "says" as a dialog tag is that it's so bland that your brain doesn't really need to process it. Using any other verb draws attention away from the dialog itself and toward the tags. My opinion is that once it's been established that *everything* the Remnant people say is in song / chant, there's no reason to repeat that fact in every single line of dialog. That would be like having the Wanderers "twitter" and "chirp" whenever they talk.

Issue #3639, opened 2018-02-10
Cannot sell last grenade


jostephd:
1. Buy 50k Laser Rifles using Ctrl+Shift+Alt+B
2. Buy 150k grenades using Ctrl+Alt+Shift+B three times
3. Sell the rifles
4. Sell the grenades
5. `You cannot sell this outfit, because that would cause your ship's "capture attack" value to be reduced to less than zero`. I think it's a floating point rounding error.

I encountered this during normal play, with ~300 grenades and human/Hai/Korath rifles. At the time I was trying to change flagship.

Hacklin:
Confirmed.  I can get this error by buying 20000 Frag. Grenades and 20000 Laser Rifles and selling all Frag. Grenades followed by selling all Laser Rifles. [Note that selling the Laser Rifles before the Frag. Grenades does not cause the rounding error for the 20k / 20k case.]

It's a rounding error in Outfit.cpp.
Increasing the Outfit's epsilon to `0.000001` will fix this for up-to the 150k / 150k case.

```
diff --git git:source/Outfit.cpp source/Outfit.cpp
index 28479394..9453191a 100644
--- git:source/Outfit.cpp
+++ source/Outfit.cpp
@@ -21,11 +21,11 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include <cmath>
 
 using namespace std;
 
 namespace {
-       const double EPS = 0.0000000001;
+       const double EPS = 0.000001;
 }
 
 const vector<string> Outfit::CATEGORIES = {
        "Guns",
        "Turrets",
```

Hacklin:
Confirmed.  PR #3645 fixes this issue.

NCGThompson:
Is there any reason we cannot simply recalculate the attack/defense values rather than operating on the previous values?

IsocratesGrey:
This issue continues to happen in 0.10.0. If the number of frag grenades and laser rifles is above If it's above 2528, when the rifles are sold first the last grenade cannot be sold.

Possible fix to this: the player has a side arm as shown in the story, perhaps the capture attack and capture defense could have something where if the value would be less than 0 the value is instead 1. I'm not sure where that would need to be though.

toazd:
I just came here to open a bug report about this but it appears it is already well known. I'll leave my report here anyways.

I'm running the 0.11.1-alpha 2280489 AppImage using ESlauncher2 on CachyOS (Arch).

[...]

PR #3645, opened 2018-02-13
Minimize precision errors of outfit additions.

flaviojs:
Selling 50000 Fragmentation Grenades, 50000 Laser Rifle and 50000 Nerve Gas accumulates so many precision errors the you are left with 1 unsellable Nerve Gas.
You can get around the issue by loading the game again.

Precision errors are avoided by truncating to `EPS` precision whenever possible.
Note that it is still possible to get precision errors if `count` is big enough to propagate the `mul` precision error to `.5 * EPS`.

Fixes: #3639 (bulk selling processes 1 outfit at a time, so count is not an issue)

Test file provided by @KiLEdEnNis: [sell_issue.txt](https://github.com/endless-sky/endless-sky/files/1722054/sell_issue.txt)

### DIFF
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -31,6 +31,10 @@ namespace {
 	inline void PreciseAddMul(double &value, double mul, int count)
 	{
 		value = trunc(trunc(value * INV_EPS) + trunc(trunc(mul * INV_EPS) * count)) * EPS;
+		// Remove sign from negative zero.
+		// EPS does not have precision errors so a simple comparison is enough.
+		if(value == -0.)
+			value = 0.;
 	}
 }
 
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -24,6 +24,14 @@ using namespace std;
 
 namespace {
 	const double EPS = 0.0000000001;
+	const double INV_EPS = 1. / EPS;
+	
+	
+	// Minimize precision errors of "value += mul * count".
+	inline void PreciseAddMul(double &value, double mul, int count)
+	{
+		value = trunc(trunc(value * INV_EPS) + trunc(trunc(mul * INV_EPS) * count)) * EPS;
+	}
 }
 
 const vector<string> Outfit::CATEGORIES = {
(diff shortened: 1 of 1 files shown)


Hacklin:
While you're at it.
What about Outfit::CanAdd() ?
Shouldn't it at least be `<= -EPS` ?  Or use `PreciseAddMul()` ?


```
         // Allow for rounding errors:
-        if(value + at.second * count < -EPS)
+        if(value + at.second * count <= -EPS)
             count = value / -at.second + EPS;
```

flaviojs:
I rather not touch it unless the precision of big counts is handled. (for that I would need to fully understand the error propagation in the multiplication)

It is fine as is because it can't produce lasting effects.
Unitary counts have no problems. Big counts might produce an "off by x" result but you can just buy/sell what was left out after the bulk operation ends.

flaviojs:
@endless-sky 
After thinking about this, I'm pretty sure there are precision issues all over the place.
The usual double has enough precision to fully hold 15 decimal digits.
With a target `const double EPS = 0.0000000001;` (in Outfit.cpp) you effectively start losing the less significant data around `100000.0000000000` (100K).
With `out.precision(8);` in (DataWriter.cpp) there is also precision loss.

I want to know if you are open to using `mpf_class` (from `gmpxx.h`) in DataNode and Dictionary (maybe more). Is it a "definitely no" or "you can give it a try"?
It introduces dependencies to `gmpxx` and `gmp`, will probably bleed out to lots of other code and will probably impact CPU performance negatively (not sure by how much).

Hacklin:
How accurate does ES have to be?
Would it be enough to disable Subnormal/Denormal floats?
And round towards 0 instead of towards -Infinity?

I've never used Multiple Precision or BigNum Arithmetic.

tehhowch:
As far as incorporating GMP, it seems they report instabilities on a variety of platforms. 

https://gmplib.org/#STATUS

I don't know that it would be wise to include this library without knowing more about the issues encountered, and the user cross-section that would be affected by it.

Zitchas:
We have another case of this come up again, reported by matthew.najmon on the Discord.

petervdmeer:
This PR appears to reduce/minimize the precision errors, while PR #5498 appears to completely remove the precision errors in the calculations where the precision matters. If I can choose between a PR that reduces the problem and another PR that completely removes the problem, then I prefer the PR that completely removes the problem. (So #5498 is preferred by me over this PR if that PR works as advertised.)

petervdmeer:
Closing this PR (which minimizes the problem) in favor of another PR which removes the problem completely.

Zireael07:
You might want to mention which PR is that, as 5498 was closed too (and that one is mentioned above as fixing the problem completely... so I'm confused)

PR #3646, opened 2018-02-14
Allow modders to restrict physical access to a planet similarly to the way they can grant it.

Hadron1776:
In the `master` branch, one can limit a planet's accessibility with `requires: <attribute>`, which makes the planet inaccessible to any ship that does not have that attribute.

This PR adds a similar check for attributes of the form `prohibits: <attribute>`, which denies access to any ship that _does_ have that attribute.

### DIFF
--- a/source/Planet.cpp
+++ b/source/Planet.cpp
@@ -436,7 +436,7 @@ const vector<const System *> &Planet::WormholeSystems() const
 // land on this planet.
 bool Planet::IsAccessible(const Ship *ship) const
 {
-	// If the attribute is required yet not present, or prohibited yet present, landing is not allowed.
+	// If an attribute is required yet not present, or prohibited yet present, landing is not allowed.
 	static const string PREFIX[2] = {"requires: ", "prohibits: "};
 	static const string PREFIX_END[2] = {"requires:!", "prohibits:!"};
 	if(!ship)
--- a/source/Planet.cpp
+++ b/source/Planet.cpp
@@ -448,7 +448,7 @@ bool Planet::IsAccessible(const Ship *ship) const
 		auto end = attributes.lower_bound(PREFIX_END[i]);
 		for( ; it != end; ++it)
 		{
-			double val  = ship->Attributes().Get(it->substr(PREFIX[i].length()));
+			double val = ship->Attributes().Get(it->substr(PREFIX[i].length()));
 			if((!i && !val) || (i && val))
 				return false;
 		}
--- a/source/Planet.cpp
+++ b/source/Planet.cpp
@@ -438,27 +438,22 @@ bool Planet::IsAccessible(const Ship *ship) const
 {
 	// Check whether any of this planet's attributes are in the form of the
 	// string "requires: <attribute>"; if so the ship must have that attribute.
-	string PREFIX = "requires: ";
-	string PREFIX_END = "requires:!";
-	auto it = attributes.lower_bound(PREFIX);
-	auto end = attributes.lower_bound(PREFIX_END);
+	static const string PREFIX[2] = {"requires: ", "prohibits: "};
+	static const string PREFIX_END[2] = {"requires:!", "prohibits:!"};
 	if(!ship)
 		return false;
-	if(it != end)
+	for(int i = 0; i < 2; ++i)
+	{
+		auto it = attributes.lower_bound(PREFIX[i]);
+		auto end = attributes.lower_bound(PREFIX_END[i]);
 		for( ; it != end; ++it)
-			if(!ship->Attributes().Get(it->substr(PREFIX.length())))
+		{
+			double val  = ship->Attributes().Get(it->substr(PREFIX[i].length()));
+			if((!i && !val) || (i && val))
 				return false;
+		}
+	}
 	
-	// Check whether any of this planet's attributes are in the form of the
-	// string "prohibits: <attribute>"; if so the ship must not have that attribute.
-	PREFIX = "prohibits: ";
-	PREFIX_END = "prohibits:!";
-	it = attributes.lower_bound(PREFIX);
-	end = attributes.lower_bound(PREFIX_END);
-	if(it != end)
-		for( ; it != end; ++it)
-			if(ship->Attributes().Get(it->substr(PREFIX.length())))
-				return false;
 	
 	return true;
 }
--- a/source/Planet.cpp
+++ b/source/Planet.cpp
@@ -445,16 +445,14 @@ bool Planet::IsAccessible(const Ship *ship) const
 	if(!ship)
 		return false;
 	if(it != end)
-	{
 		for( ; it != end; ++it)
 			if(!ship->Attributes().Get(it->substr(PREFIX.length())))
 				return false;
-	}
 	
 	// Check whether any of this planet's attributes are in the form of the
-	// string "excludes: <attribute>"; if so the ship must not have that attribute.
-	PREFIX = "excludes: ";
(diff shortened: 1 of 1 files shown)


KiLEdEnNis:
wouldn't it just be very annoying if one couldn't access a landable object because of an outfit?

a use I could think of, though, is the Kor Mereti/Sestor planets/stations which would require one to have the [Reasoning Node/Control Transceiver] and not the other

source/Planet.cpp
```diff
@@ -438,18 +438,27 @@ bool Planet::IsAccessible(const Ship *ship) const
...
+		for( ; it != end; ++it)
+			if(!ship->Attributes().Get(it->substr(PREFIX.length())))
+				return false;
 	
-	for( ; it != end; ++it)
-		if(!ship->Attributes().Get(it->substr(PREFIX.length())))
-			return false;
+	// Check whether any of this planet's attributes are in the form of the
+	// string "prohibits: <attribute>"; if so the ship must not have that attribute.
+	PREFIX = "prohibits: ";
```
tehhowch: I think the duplication can be avoided by using a string array, and a for loop:
```
static const string PREFIX[2] = {"requires: ", "prohibits: "};
static const string PREFIX_END[2] = {"requires:!", "prohibits:!"};
if(!ship)
    return false;
for(int i = 0; i < 2; ++i)
{
    auto it = attributes.lower_bound(PREFIX[i]);
    auto end = attributes.lower_bound(PREFIX_END[i]);
    for( ; it != end; ++it)
    {
        double val = ship->Attributes().Get(it->substr(PREFIX.length()));
        // If the attribute is required yet not present, or prohibited yet present, landing is not allowed.
        if((!i && !val) || (i && val))
            return false;
    }
}
```

tehhowch:
From a design perspective, I wonder if Planet->IsAccessible() is the correct place to check for "prohibited" attributes. If a mod were to add e.g. `"prohibits: turn"` to all Republic planets, then the player would not even be able to target these planets - they would look like uninhabitable planetoids, and there would be no explanation for the behavior.

Perhaps this is a check better suited for Planet->CanLand()?  As it is, `requires` gates the fundamental discovery of habitability / utility behind an attribute. The converse (here implemented via `prohibits`) is not necessarily as sensible, or at least harder to make lore with. 
Or, perhaps both this and a more political `"banned: "` check are needed (with a corresponding politically-oriented "We won't let you land because you don't have this attribute" functionality).

source/Planet.cpp
```diff
@@ -436,20 +436,19 @@ const vector<const System *> &Planet::WormholeSystems() const
...
 {
-	// Check whether any of this planet's attributes are in the form of the
-	// string "requires: <attribute>"; if so the ship must have that attribute.
-	static const string PREFIX = "requires: ";
-	static const string PREFIX_END = "requires:!";
-	auto it = attributes.lower_bound(PREFIX);
-	auto end = attributes.lower_bound(PREFIX_END);
-	if(it == end)
-		return true;
+	// If the attribute is required yet not present, or prohibited yet present, landing is not allowed.
```
tehhowch: `// If the` -> `// If an`

Otherwise, the implementation looks good.

ghoulavenger:
The requires attribute I'm fine with, but I'm not a fan of the prohibits attribute. The difference being that requires assumes that the planet is already locked to you, and thus you are presenting the key (like submitting landing codes, presenting a warrant or proof of citizenship etc). If a world that is open to you denies you access because of something you have on board you're screwing with the concept of illegal attributes and fines; basically the assumption you need to be caught with something before you are penalized for it.

Amazinite:
That assumes that the reason you're prohibited from landing is because you have illegal outfits on board. I don't see a reason why this functionality shouldn't exist, and at this point all I see are arguments against certain uses of the mechanic and not the mechanic itself, although I do agree with @tehhowch's assessment that this should be changed so that planets don't appear uninhabited because you are prohibited from landing due to an attribute.

ghoulavenger:
@Amazinite While I do believe that it screws with illegal outfits, that went in line with what tehhowch was trying to do with political leanings no? It discounts smuggling pretty heavily if the check is an automatic failure. If you wanted to do something this way I'd much rather it be done by fleet combat or inspection of some sort than an automatic failure.

tehhowch:
@ghoulavenger you are jumping the gun and considering a poorly implemented mod. This PR adds a physical-basis restriction. It would be creator error to combine the tag this PR uses (as of https://github.com/endless-sky/endless-sky/pull/3646/commits/696d0d0d72c82e191ec29d3df527c2871a87bcd7) with an attribute that arises from "political" sensibilities (such as `illegal`).

Amazinite:
>Or do you have an example where prohibits wouldn't be used like that?

Not going into too much detail as to the intricacies of each idea, but here are three that I came up with in a few minutes.

* Wormholes that can not be entered if you have an attribute that makes the wormhole "unstable," just as the Quantum Keystone makes the Ember Waste wormholes "stable."
* An alien race that will not allow the player to land on their planets when a jump drive is installed, perhaps out of superstition or fear. This, though, relates to the previously mentioned possible `"banned: "` mechanic in which planets remain appearing inhabited.
* Major worlds in human space won't let you land if you have nukes installed out of fear of another Gemini/Martini-like bombing. (Again, would make more sense with the mentioned `"banned: "` mechanic, and would also require nukes to have a unique attribute, but the spirit of the idea of `"prohibits: "` is still present.)

tehhowch:
I've already indicated that I feel that a political check should be available. Such a check is more involved and requires changing a number of methods to provide the minimum support. This PR adds the converse of an existing feature, for which the game has full support already. While we can't see fantastic uses for it immediately, that does not mean there are none.
One such scenario is for an "Easter egg" planet located in a system otherwise only used for a `requires: ` wormhole entry. Unless you journeyed to its system solely to explore and not to use the wormhole, you wouldn't find this hidden planet.

ghoulavenger:
I still don't like it, but after consideration there are two good applications that have been presented to me. The concept of an easter egg planet, and the wormhole justification -- I don't know why either one of these would happen but they are well within the realm of possibility.

Amazinite's other two suggestions did not appeal to me as the blanket prohibits seems like it should be handled with proper events instead. Or like what I said before, space inspections and combat and not an automatic failure condition. 

@tehhowch Features are judged by their pros and cons are they not? When I see mostly cons and no pros, I see no reason to support a feature with every reason not to like it so your argument falls flat. Features for the sake of features just clutters the game without adding value. 

Naturally though I am a fallible person prone to errors in judgment. There may be value here that I cannot yet perceive, which is why I asked for viable applications. The fact that any were presented to me at all, gives this PR some value.

Amazinite:
>Features are judged by their pros and cons are they not? 

I wouldn't consider poor use of a feature or the inability to think of a good use of a feature on the spot to be a con, as that can be the case for any new feature.

elgeonmb:
other use cases:

- massive/poorly constructed ship can't land on high gravity worlds / worlds without sufficient facilities to receive them
- certain special craft (gascraft, experimental wormhole scouts, ships that are actually just a very large space monster you've stapled an armchair and a joystick to) can't land on most normal worlds
- Mysterious Space Phenomenon shifts your ship out of normal space, preventing you from interacting with normal planets until you fix the problem/ it's not actually a problem, you just have to visit certain "gateway" planets that exist in both dimensions to transition between states.

I'm also fond of "quantum keystones destabilize this wormhole because *mumble mumble* space"

RisingLeaf:
what exactly is needed in this PR?

Amazinite:

PR #3647, opened 2018-02-15
Workaround SDL generating extraneous keydown events when switching to fullscreen.

### LINKED ISSUE #3631: Switching from window to fullscreen fails when mouse not over the window
When the game is windowed and the mouse cursor is not over the window, pressing `"Toggle fullscreen"` (F11) doesn't switch to fullscreen. It works fine when the mouse cursor is over the window.

jostephd:
Add an ugly workaround for an SDL bug which is not fixed yet.

Fixes #3631.

Thanks to @Hacklin for debugging and reviewing earlier versions of this
patch.

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -274,14 +274,25 @@ int main(int argc, char *argv[])
 				{
 					// Toggle full-screen mode. This will generate a window size
 					// change event, so no need to adjust the viewport here.
-					isFullscreen = !isFullscreen;
-					if(!isFullscreen)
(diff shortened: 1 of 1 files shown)


jostephd:
It does not: I can reproduce the original problem both at 610088 and at its parent dafe0599.

endless-sky:
Give f385735 a try. (I'm trying to avoid a solution that depends on the granularity of time().)

jostephd:
It works now, thanks!

PR #3650, opened 2018-02-19
Make the Kuwaru Efreti Quarg more consistent with their language

Bladewood:
My main issue is specifically in the Wanderer mission when you are ordered to ask the Quarg for help, because the Quarg then have a very good handling of the modern English language, when in a previous encounter (if any) they clearly... had problems. I propose that we change that.


endless-sky:
It seems a *little* excessive to have the Quarg speak pseudo-old-English. I think just using unusual words and convoluted sentence structure gets the idea across well enough.

(Also, given how non-hierarchical Quarg culture seems to be, it's likely they would use "thee / thou" instead of "ye / you." Formal modes of address probably wouldn't be something they could easily wrap their heads around.)

PR #3655, opened 2018-03-02
Made Korath hostile towards Bounty Hunters

atoms118:
Considering Korath raiders will attack everyone else, it makes sense that they'd also attack bounty hunters; because why leave them out of the fun.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -148,7 +148,7 @@ government "Korath"
 	
 	"attitude toward"
 		"Kor Sestor" -.01
-		"Bounty Hunter" -0.01
+		"Bounty Hunter" -.01
 	
 	"player reputation" -10
 
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -148,6 +148,7 @@ government "Korath"
 	
 	"attitude toward"
 		"Kor Sestor" -.01
+		"Bounty Hunter" -0.01
 	
 	"player reputation" -10
 


data/governments.txt
```diff
@@ -148,6 +148,7 @@ government "Korath"
 	
 	"attitude toward"
 		"Kor Sestor" -.01
+		"Bounty Hunter" -0.01
```
tehhowch: You can match the existing style with `-.01`

atoms118:
@tehhowch I have changed it as you suggested

endless-sky:
Bounty Hunters are only a mission-related government; they never spawn on their own, and they're always supposed to just target the player. So, it doesn't make sense to me for the Korath to attack them, any more than it would make sense for the Korath to attack a "marked" mission target or Bounty ship.

atoms118:
*gets excited because the owner looked at my PR*

From a realism/gameplay perspective, it seems that the Korath attack pirates and merchants (presumably for resources like food, as noted by the lore). It doesn't make sense then that the Korath wouldn't attack the bounty hunters, except maybe if they were intelligent and decided to work together to destroy the player first. However, since pirates also attack the player and the Korath raiders will attack pirates as well, that logic doesn't hold up...

I really don't want to come across as arrogant or anything; if I'm missing anything please let me know

Amazinite:
I feel like the odds of bounty hunters ever crossing paths with the Korath is low enough that for the sake of the gameplay purpose of bounty hunters, the Korath don't need to attack them. Personally, I can't even recall ever having a situation where bounty hunters and Korath are in the same system. 

And why would the Korath attack bounty hunters but not bounties that you get from jobs or the Free Worlds/Republic ships that you can bring in to Syndicate space through missions? That is, if the justification for this PR is that Korath should attack any human ship in sight then this PR doesn't go far enough.

KiLEdEnNis:
it's true it's a bit unrealistic if the Korath don't attack Bounty Hunters, but those Bounty Hunters only spawn if the player has a bad reputation(?)
the Korath don't attack them so the player still gets punished for being a prick to humanity

short answer: balance over realism

atoms118:
@Amazinite you make a good point; the PR probably doesn't go far enough. However, I have actually had a fair number of bounty hunters show up while waiting in systems like Sheratan and Polaris for raiders, so it isn't so rare that it isn't worth considering (either that or I've just had really bad luck, which is a real possibility)

@KiLEdEnNis you also make another good point, although at the point the player is taking on Korath raiders, the bounty hunters are more of an annoyance than an actual challenge/balance thing (I might be wrong about that, but that was at least the case for my playthrough).

tehhowch:
Bounty Hunters spawn when you have a [bad reputation](https://github.com/endless-sky/endless-sky/blob/d02ac0f3be1976b3cf0cd92db0ec03ccb8c65c44/data/pirate%20jobs.txt#L2588-L2638), and when you have a [high combat rating](https://github.com/endless-sky/endless-sky/blob/d02ac0f3be1976b3cf0cd92db0ec03ccb8c65c44/data/jobs.txt#L2584-L2624).

atoms118:
It seems at this point that the census is that it's a gameplay over realism thing; if that's the case I'd be happy to close this PR

PR #3657, opened 2018-03-02
Made Kor Sestor attack Bounty Hunters after Zenith is bombed

atoms118:
After Zenith is bombed, the lore says literally says they `are attacking Navy and civilian ships indiscriminately.`, so it makes sense then that the Sestor drones would attack bounty hunters as well after that event

### DIFF
--- a/data/wanderers middle.txt
+++ b/data/wanderers middle.txt
@@ -2454,6 +2454,7 @@ event "bombed zenith"
 	government "Kor Sestor"
 		"attitude toward"
 			"Pirate" -.01
+			"Bounty Hunter" -.01
 	planet "Zenith"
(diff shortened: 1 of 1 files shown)


endless-sky:
Bounty hunters are all associated with missions and are only intended to attack the player, so it wouldn't make sense for the Korath to attack them.

atoms118:
I'm not sure I understand. I get that the bounty hunters are only supposed to attack the player, but the Korath automata are attacking Navy, merchant and pirate ships as well, so it doesn't really make sense from a realism perspective that they wouldn't attack pirates. If they're attacking everything and everyone indiscriminately, they should be attacking bounty hunters as well

jafdy:
It's for the same reason the ships you hunt in bounty missions are ignored by everyone else. The bounty hunters are meant to your problem to deal with.
Simply it's gameplay over realism.

atoms118:
OK, you make a good point. I still think that at the point where you've bombed Zenith, the bounty hunters are more of an annoyance than a challenge, but it's fair enough (plus, if they're not that big a deal, maybe they're not PR worthy)

PR #2574, opened 2017-05-29
Adding some mission log entries (non-Free Worlds, non-Wanderer)

Elyssaen:
Done Coalition, Drak, Hai, Kestrel and Quarg logbook entries. Transport missions is an early WIP, but I thought I'd post a PR and just add more commits. Feel free to chuck a PR into my branch to make changes / add new ones.

First Quarg one is too boring right now, and there's plenty more to do in transport missions. @Amazinite is going to work on FW ones and probably PR that separately. I'm not imagining touching Wanderer ones in this PR. The Pug in Wanderer space one I probably will.

PR #3658, opened 2018-03-05
Added log entries for various transport missions

Amazinite:
Ref: #2574 

Added log entries for the more important transport missions, especially those that have multiple parts to the storyline (Smuggler's Den, Deep/Rim Archaeology, TMBR, Terraforming) or may have future implications (Terminus Exploration).
Also added a log entry for the hallucination that is intentionally meta. This entry can be removed or modified if desired.

### DIFF
--- a/data/transport missions.txt
+++ b/data/transport missions.txt
@@ -198,6 +198,7 @@ mission "Smuggler's Den, Part 2"
 				accept
 	on complete
 		event "smuggler's den: payment" 365
+		log `Helped two teens named Joe and Maria and their daughter Jesse to escape their pirate captain. Dropped them off on Millrace where they'll start a new life.`
 		conversation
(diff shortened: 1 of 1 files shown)


mootootwo:
Nice!  Do Log entries support variables for destinations and such?

So, instead of "dropped them off on Millrace" it would be "dropped them off on <destination>" or "dropped them off on <planet>" ?

KiLEdEnNis:
I doubt it
those logs are basically saved strings, so they shouldn't remember what the original destination was, as only the mission itself does
the mission itself has a tag defining the destination, and upon completion, that mission gets removed from the save file (apart from the offered & deleted)

mootootwo:
I just tested, and it does take the string replacement variables.

It specifically does not record <first> <last> or <ship> in the log... so that is a little unexpected.  It also did not record <stopovers> for me, but my test mission did not have one.  All the other variables worked as expected.

Issue #1344, opened 2016-07-18
Arfecta cloaking device irremovable


OmegaZanicus:
I don't know if this was on purpose, but it is possible to capture the Arfecta, but since the ship has
`-.2 cloaking fuel`
It is impossible to remove the cloaking device.
1) If this was done on purpose, why not simply add cloaking device as a built-in.
2) If not, make it so the cloaking fuel can go negative.

KiLEdEnNis:
can you try and describe it a little bit better?
I saw in the files that the ship itself has the cloaking fuel stat, and I'm not sure why it has that, because in the main story, if you help the Syndicate defector, you later get a cloaking device in your flagship which you can remove, and it drains fuel, so I think that stat just has to get removed
in case it's on purpose (which I doubt), having it built-in may be very hard

OmegaZanicus:
This isn't really user error, more a feature request I guess but:

The Arfecta ship has a built in stat, like `cargo space`, or `shields`, `-.2 cloaking fuel` that balanced out the `.2 cloaking fuel` that the cloaking device uses. This allows it to stay cloaked forever. However, this stat also prevents the user from removing the cloaking device, because that would cause the `cloaking fuel` stat to go below zero, and the game doesn't allow that.

What I was asking, was if this was intentional for the Arfecta, or if MZ hadn't anticipated players capturing it, or if it was in fact a mistake.

KiLEdEnNis:
I don't think he intended it to be captured, but then he could add the "never disabled" feature
giving it a massive amount of fuel and ramscoop would nearly have the same effect

enabling it to go below 0 would require that stat to do absolutely nothing if no cloaking device was installed, so it's not really a bug, more a mechanics thing

KiLEdEnNis:
and in case MZ reads this; I am working on extra content (for myself), where can I put the idea and could it be released if I made an actual story?

OmegaZanicus:
the thing is, it _doesn't do anything when there is no cloak_, because it only has effect when the ship is cloaked.

P.S. please go make your own issue, don't just hijack my problem please. its a little rude.

KiLEdEnNis:
I try to help (but the last comment I understand, just don't know where to put it at all)

I know it doesn't do anything, but the game may not see it that way, maybe making another cloaking device for the Arfecta might help, like an upgraded version of the other one the player can get, although that'll enable the player to infinitely cloak

toilethinges:
#1066

[...]

PR #3669, opened 2018-03-30
Made the cloaking device unplunderable.

Bladewood:
This patches a particular exploit where the player is able to plunder the Cloaking Device from Pug Arfectas (yes, it can be done, I've tested it), even though they cannot uninstall it from within an outfitter without selling the ship. (This is [intentional](
https://github.com/endless-sky/endless-sky/issues/1344#issuecomment-234128902).)

- It doesn't seem logical to me that you would be able to plunder the cloaking device with on-ship tools, yet not be able to uninstall it with a full outfitter.
- Note that this patch would still allow a cloaking device in cargo to be plundered, according to [BoardingPanel.cpp#L88](https://github.com/endless-sky/endless-sky/blob/1fd6336f3194f5a2d2fc148f48cad8790a5e327b/source/BoardingPanel.cpp#L88). The cloaking device, as far as can be known to the player, has only ever been left completely deactivated and [therefore visible](https://github.com/endless-sky/endless-sky/blob/master/data/outfits.txt#L258) and removable (within reasonable ability) once [in a certain Korath ship's cargo hold](https://github.com/endless-sky/endless-sky/blob/master/data/free%20worlds%20reconciliation.txt#L2243).

### DIFF
--- a/data/outfits.txt
+++ b/data/outfits.txt
@@ -255,6 +255,7 @@ outfit "Cloaking Device"
 	"cloak" .01
 	"cloaking energy" 1
 	"cloaking fuel" .2
+	unplunderable 1
 	description "You're sure the cloaking device is here somewhere, but you can't see it."
 
 


Janaszar:
Considering that the idea is to have the Arfectas require cloaking, why not just remove the cloaking outfit from them (and change the outfit space to account for it) and give them integrated cloaking like the Starling?

Aside from that, giving Cloaking the "Unplunderable" aspect is a good idea, especially if we see it used on other ships.  But if all cloaking except that one from the Reconciliation path is going to be "mandatory" on the ships, I see no point as making the cloaking integrated would be simpler and gives the "reward" for doing Reconciliation a "boost".

jafdy:
I think the idea is that the cloaking device works better in the Arfectas. I can't think of a good way around this other than allowing some attributes go negative.

Bladewood:
The difference between the arfectas and the starlings is that the former, at least in the current game, are limited in quantity, and so making it possible to separate their cloaking devices is a not completely overpowered course of action.

Issue #1882, opened 2016-12-07
Build system


defanor:
It would be nice (at least for some users and systems) to introduce
the following build-related changes:

- Provide an `INSTALL` file, which would include instructions along
  with a list of dependencies.

- Use, at least as an alternative, something more common than `scons`:
  if not GNU build system, at least a `Makefile`, or a shell
  script. While the game itself doesn't seem to depend on Python,
  scons (which also requires Python 2, not 3) pulls dependencies that
  would be unnecessary otherwise.

vflyson:
I second this proposal. It seems to be a very good open-source game, yet it's a shame that it's not that easy to stay up-to-date with the latest version if you're using an open-source OS.

defanor:
Nice. I think it is an improvement, though there is a couple of small
issues with that, too:

- It requires cmake 3.1.0 or newer there, while Debian 8 (jessie,
  current stable) only has 3.0.2 in its repositories, and CentOS/RHEL 7
  has 2.8.12.

- Cmake itself is not that minimalistic, and IME it's not installed on a
  system more often than it is -- so, still additional dependencies just
  to build the game.

lheckemann:
Yeah, tbh I think it could probably be done with a plain makefile and I'll have a go at doing so tomorrow. Happy new year!

defanor:
Shouldn't be hard to compose a basic Makefile, indeed, since it can be
built even with just `g++ -std=c++11 -Wall source/*.cpp -lSDL2 -lpng
-ljpeg -lGL -lGLEW -lopenal -lmad`, though then there's a lot of special
cases to cover, like installation prefix, additional options, clean
uninstall, etc.

Meantime, I've tried to build it with autotools, but to make it work
nicely, it would require to adjust the sources as well -- to use correct
paths and such (looks like they are currently hardcoded in `Files.cpp`,
even though the `SConstruct` file appears to handle custom
prefixes). It's also quite cumbersome to install all the data files, so
not sure if I will continue; in case if somebody else would want to do
that, here is what I've got:
<https://gist.github.com/defanor/b9adc5fafbb6ee28765d0f3a978b424d>.

And happy new year to you too.

jraynal:
Hi, I do realize I am late to the party, I was curious about CMake support. I am using CMake quite extensively in my day-job, and I think it could bring some interesting aspects that a simple Makefile can't do. (I never designed build systems with autotools).

[...]

PR #3671, opened 2018-04-01
+Makefile

Muffindrake:
Scons is so evil, it has its own plane of existence in the deepest recesses of hell, where it tortures chained, naked and afraid distribution maintainers with implements of destruction and despair better known as `Python 2` and `doing everything that make does, but worse`. It has also been seen stealing cake from your fridge.

This adds a Makefile that is tested to work on _one_ GNU/Linux machine under GNU make. Changes are requested, if any are necessary to make it work under more systems.

I've just seen that Steam still doesn't ship libmad with its steam runtime boogaloo, so the static linkage shenanigans should be added later. There is also the matter of extending the developer readme or replacing it with a more pragmatic- and intimidating-sounding *INSTALL* file.

#1882 

Disclaimer: This PR has nothing to do with the date it is posted on.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,10 +1,10 @@
-######Building Endless Sky
+###### Building Endless Sky
 _NOTE: If you have strayed here by accident and are inexperienced with building
 software and the command line, it may be a better idea to grab the program from
 your package manager, the releases section in GitHub, or Steam, rather than
 fruitlessly wasting your time_
 
-###Linux
+### Linux
 You must install both headers and runtimes of the following libraries:
 - sdl2
 - libpng
@@ -36,7 +36,7 @@ dependencies using the following command, requiring root privileges:
 
     # apt install build-essential lib{sdl2,png12,jpeg62-turbo,glew,openal,mad0}-dev
 
-###Mac OS X
+### Mac OS X
 You must install [Homebrew](https://brew.sh), then you must install the
 following packages using `brew install`:
 
@@ -82,5 +82,5 @@ After sufficient cursing at Apple, you may proceed to build and run:
     $ make
     $ ./endless-sky
 
-###Windows
+### Windows
 TODO
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,86 @@
+######Building Endless Sky
+_NOTE: If you have strayed here by accident and are inexperienced with building
+software and the command line, it may be a better idea to grab the program from
+your package manager, the releases section in GitHub, or Steam, rather than
+fruitlessly wasting your time_
+
+###Linux
+You must install both headers and runtimes of the following libraries:
+- sdl2
+- libpng
+- libjpeg-turbo
+- glew
+- openal
+- mad
+
+The following tools are necessary to proceed:
+- C++ compiler that supports C++11 (GCC 4.8.1+, clang 3.3+)
+- GNU make
+- pkg-config
+
+You then run the following commands in the same directory as the `Makefile`:
+
+    $ make
+    $ ./endless-sky
+
+`-j9` may be passed to `make` to make the build go faster at the expense of
+memory and CPU time. Setting an environment variable `OPTFLAGS` will alter the
+default optimisation flags, while `CXXFLAGS` and `LDFLAGS` will let you change
+the compiler and linker flags respectively, if any changes are necessary.
+
+The program will infer the location of the `data` and `images` directories
+according to the rules outlined in the man page `endless-sky.6`.
+
+Debian-based distributions will probably want to install all necessary
+dependencies using the following command, requiring root privileges:
+
+    # apt install build-essential lib{sdl2,png12,jpeg62-turbo,glew,openal,mad0}-dev
+
+###Mac OS X
+You must install [Homebrew](https://brew.sh), then you must install the
+following packages using `brew install`:
+
+    $ brew install pkg-config sdl2 openal-soft jpeg-turbo lib{png,mad}
+
+You must also either set `PKG_CONFIG_PATH` or follow the instructions according
+to the output of homebrew after you install any of those packages in order for
+`pkg-config` to be able to find its library definition files necessary to
+proceed.
+
+Example output:
+    This formula is keg-only, which means it was not symlinked into /usr/local,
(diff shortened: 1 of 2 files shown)


jafdy:
(Without even thinking about the post date) this sounds like a joke PR. 
What is with that opening paragraph, it isn't saying much about anything other than you don't like Scons. (Also I don't keep cake in my fridge, I keep it in a cake tin.)

MCOfficer:
(despite the elaborate intro ;) ) from what i can tell with my limited knowledge this is a pretty good makefile. much better than my own one, which has ~300 lines in total.

Hacklin:
SCons requires:
- `Python >= 2.4 and < 3.0`  according to the [FAQ](http://scons.org/faq.html),
- `Python 2.7.*` or `Python >= 3.5`  according to the [Project Source](https://github.com/SCons/scons).

I think the FAQ is out-of-date.
Anyway, SCons is or will be working on `Python 3`.

Muffindrake:
@jafdy The opening paragraph is not exactly serious, if that is unclear.

Hacklin:
Sorry, no, I didn't get the irony.

Makefile
```diff
@@ -0,0 +1,74 @@
...
+$(info mkdir -p $(DIR_RELEASE) $(DIR_DEBUG) $(DIR_PROF))
+$(shell mkdir -p $(DIR_RELEASE) $(DIR_DEBUG) $(DIR_PROF))
+
+$(DIR_RELEASE)/%.o: $(DIR_SRC)/%.cpp
+	$(CXX) $(OPTLEVEL) $(CXXFLAGS) -o $@ -c $<
+
+release: $(OBJS)
+	$(CXX) $(OPTLEVEL) $(CXXFLAGS) -o $(PROG) $^ $(LDFLAGS)
+
+$(DIR_DEBUG)/%.o: $(DIR_SRC)/%.cpp
```
CyberShadow: This won't work right when just a header file changes. (Not a problem with Scons, which seems to do dependency discovery and rebuilds only the minimum set of objects as necessary).

It's possible to do this correctly in Make (see gcc's `-M` options), but it's probably simpler to stick to a language-aware build system (like Scons).

Muffindrake:
@CyberShadow Now includes proper tracking of header dependencies. Is there anything else?

Makefile
```diff
@@ -0,0 +1,80 @@
...
+OBJS_PROF = $(patsubst $(DIR_SRC)/%.cpp,$(DIR_PROF)/%.o,$(SRCS))
+DEPS_PROF = $(patsubst $(DIR_SRC)/%.cpp,$(DIR_PROF)/%.d,$(SRCS))
+
+$(info mkdir -p $(DIR_RELEASE) $(DIR_DEBUG) $(DIR_PROF))
+$(shell mkdir -p $(DIR_RELEASE) $(DIR_DEBUG) $(DIR_PROF))
+
+$(DIR_RELEASE)/%.o: $(DIR_SRC)/%.cpp
+	$(CXX) $(OPTLEVEL) $(CXXFLAGS) -o $@ -c -MMD $<
+
+release: $(OBJS)
```
CyberShadow: Currently this will cause `make` to link the game unconditionally on every invocation (i.e. the build is never clean). Because the binary for all build modes is the same, there is the problem that when building A, then B, then A again the build system would need to know to relink the game in mode A even though all timestamps are up to date.

I'm not sure there is a clean way to do this in Make. Scons seems to be aware of this and relinks the binary as needed, while still allowing clean rebuilds. I guess you could link the binary in the build directory, then just copy it over in the top-level target, which would save the effort of linking.

INSTALL.md
```diff
@@ -0,0 +1,87 @@
+###### Building Endless Sky
+_NOTE: If you have strayed here by accident and are inexperienced with building
+software and the command line, it may be a better idea to grab the program from
+your package manager, the releases section in GitHub, or Steam, rather than
+fruitlessly wasting your time_
```
CyberShadow: This file seems to be somewhat redundant with `readme-developer.txt`. Perhaps one should be merged into the other, or the data otherwise deduplicated and a reference to the other file added.

Muffindrake:
@CyberShadow The Makefile no longer links on every invocation, removing unnecessary work.

>Because the binary for all build modes is the same, there is the problem that when building A, then B, then A again the build system would need to know to relink the game in mode A even though all timestamps are up to date.

I'm not sure I understand - the build had (and still has) distinct binaries (`endless-sky{-debug,-profile}`) for each mode,

CyberShadow:
> I'm not sure I understand - the build had (and still has) distinct binaries (endless-sky{-debug,-profile}) for each mode,

Right, sorry, I misread the makefile - thought it was mimicking what Scons was doing and always building to ./endless-sky .

Muffindrake:
Given that the binaries are distinct anyway, I removed the copy, since the debug builds are quite sizable (`54M` vs `2.8M`).

>This file seems to be somewhat redundant with readme-developer.txt. Perhaps one should be merged into the other, or the data otherwise deduplicated and a reference to the other file added.

The `INSTALL.md` was meant as a replacement, but I have little information from MZ other than `want to build new files with minimal changes to build system`, so I'm unsure how to proceed.

There is still the standing issue of building on windows. I can see using cygwin and have the Makefile work without changes (at least I hope so), but I'm unsure whether that code blocks method is something that wants to be kept around.

CyberShadow:
> I can see using cygwin and have the Makefile work without changes (at least I hope so),

We probably don't want to support CygWin as the target platform on Windows, as programs built under Cygwin aren't really "native" Windows programs (they use UNIX APIs through cygwin's emulation layer). MSYS2 + MinGW is probably the way to go here.

OrdoFlammae:
If this is going to be a possibility, you would want to use something much more flexible than bare Makefile. At least, turn to Autotools (which I would also not recommend). There are build systems that have better than just a Makefile, although they might use it in the end. Also, I would recommend that you use something that is usable on Windows (so, Makefile and Autotools fly right out the window).

MCOfficer:
to dynamically create makefiles, one could look towards cmake, as suggested in #1882 .

Amazinite:
Closing as this is marked unlikely.

PR #3684, opened 2018-04-08
new map background and labels

Aurelite:
This PR add a new map background image remade from the [source](http://hubblesite.org/image/2092/category/37-spiral-galaxies) and new region labels.
I removed all the foreground stars and restored the blown-out parts.
I like it with a hint of color but it could also work monochrome.

new background with labels (40% size):
map-label

in-game:
ingame


MCOfficer:
(Probably 90% personal preference).
I like the new Galaxy, however, i don't like the labels. The old ones looked more like something a medieval ship captain would use on their map, which kinda fits the exploration aspect of ES.

Amazinite:
I'd say keep the old labels. These new ones seem too bland to me. The galaxy image is fine, though.

endless-sky:
Removing the foreground stars is definitely a _big_ improvement.

On the other hand, I'm worried that the star labels become a lot harder to read with this version, in certain parts of the map. It looks like you've added texture (lots of tiny spots for stars) in sections of the map that were previously a smooth, flat, cloudy texture. That texture makes it harder to read the overlying text. It also means that instead of some parts of the galaxy being hazy and others being speckled with lots of individual stars, the whole thing has a very similar texture. In visual design, it's important to balance out areas of tiny details and fine texture with other areas that aren't so detailed.

elgeonmb:
is pug iyik still hiding behind a star?

Amazinite:
@elgeonmb It's hiding in what looks like a star cluster or a nebula now instead of a single bright star. I worried that would have changed as well and made sure to check it before passing judgement on the new galaxy image.

elgeonmb:
good, it took me multiple days to find pug iyik and i expect all new players to have to do the same

Aurelite:
Thanks for the feedback.

> the whole thing has a very similar texture
Since the map has already a lot going on, i tried to dim the background a lot.
I see that i went a bit overboard there.

I reworked the galaxy image, especially around the core and coalition space.
Also brightened the labels a bit.

endless-sky:
I like that one a lot better, and it's a _huge_ improvement over the original. I especially like how much more distinctive the core of the galaxy looks now.

Would you be willing to create a separate PR with just the galaxy image, since the labels seem much more controversial?

Sample image for the sake of others reading this:

core

Amazinite:
>Would you be willing to create a separate PR with just the galaxy image, since the labels seem much more controversial?

Please do so.

PR #3693, opened 2018-04-10
Handle exception for std::stoi in Dialog.cpp

### LINKED ISSUE #3692: Overflow in dump cargo dialog
https://steamcommunity.com/app/404410/discussions/0/1696043806574246615/
User gets a crash at 1B cargo dump values. I don't think it make sense to use larger values but avoiding a crash seems like a good idea.

Muffindrake:
Fixes #3692

### DIFF
--- a/source/Dialog.cpp
+++ b/source/Dialog.cpp
@@ -28,7 +28,9 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "SpriteShader.h"
 #include "UI.h"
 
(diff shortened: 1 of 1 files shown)


endless-sky:
This doesn't follow the [style guide](http://endless-sky.github.io/styleguide/styleguide.xml), and it's also more complicated than it needs to be - treating malformed values as if "cancel" was clicked (2068740) makes more sense to me than truncating to 0 or MAX_INT.

Issue #3696, opened 2018-04-11
Pug Gridfire Damages Self after closing Boarding Panel


manitscold123:
After you board a disabled ship, you can exit the boarding screen, at which point you will fire upon the ship and destroy it.

Because the Pug Gridfire does damage based on position of sprites (or something like that), after you exit the boarding screen your ship will fire upon the ship you were just boarding, dealing damage to yourself, because you were positioned right upon the disabled ship, because you just boarded it. Sadly, most captains aren't actually suicidal, so doing so would seem counterproductive.

Firing upon yourself with the already broken-af gridfires will damage both the hull and the shields, and, for many ships, may even destroy yourself. For example, shooting yourself in a shield beetle, after exiting the board screen on a carrier will take out quite a bit of your health bar. Ouch.

All I'm saying is that for gameplay and non-stupidity, you probably shouldn't shoot yourself with your own tier 3 weapons, especially since the damage here actually goes through your shields lol.

**If someone ok's it I'll add a PR where I just add a conditional, so that you don't run the code to shoot the enemy ship if your weapon is like gridfire or something.** Sadly enough, I'm pretty inept when it comes to writing C++ (and also reading C++ for the most part), but I can probably handle *one tiny conditional, right?* 👍

manitscold123:
May I also add that it is mechanically simple to *accidently* disable yourself due to gridfire? Like, no joke, I just did it like three times in a row...

Twice against Korath, and then again against the Unfettered. Maybe it's just cuz I suck, but jeez this is just beyond.

KiLEdEnNis:
once one leaves the boarding panel, if they have autofire on, they automatically  destroy the ship they just boarded
your issue would apply to any non-secondary weapon with splash damage
I personally think autofire shouldn't destroy a ship once finished boarding, and that would fix this

Zitchas:
I agree. I don't use autofire most of the time, but even when I do, I rarely want to destroy the ship I just boarded. I'm interested in the loot, not blowing them up. If a player really wants to destroy a ship, it isn't that hard to shoot it.

KiLEdEnNis:
if one wants to destroy it after; they still have it selected, hold space bar
right now one has to quickly deselect it

manitscold123:
So is it okay to just remove the *auto-fire destroy a ship after cancelling boarding code?*

Though lol I'm not interested in the loot lol I'm more interested in the ship itself.

tehhowch:
[...]

PR #3697, opened 2018-04-11
Auto-Aim to Spare the Disabled after closing board panel

manitscold123:
As noted in #3696, auto aim, upon the closing of the board panel, fires upon the ship that was just boarded, destroying it.

Since most players aren't bloodthirsty hounds that kill for sport or just for the sensation of murder, it can be assumed that the auto-aim shouldn't carelessly wast lives either.

Also, splash damage non-secondary weapons (like Pug Gridfire Turret) firing upon an enemy ship will damage the player as well, making this decision doubly stupid. 

I'm pretty sure that this code will fix this problem, as now, if the ship belongs to the playerGovernment, it won't fire upon disabled ships, like it should.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2529,7 +2529,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 		currentTarget.reset();
 	
 	// Only fire on disabled targets if you don't want to plunder them.
-	bool spareDisabled = (person.Disables() || (person.Plunders() && ship.Cargo().Free()));
+	bool spareDisabled = (person.Disables() || (person.Plunders() && ship.Cargo().Free()) || ship.IsYours());
 	
 	// Don't use weapons with firing force if you are preparing to jump.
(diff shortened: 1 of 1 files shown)


jafdy:
I don't know much about the code but it looks like this would also prevent you from instructing your ships to kill a disabled ship (if you are bloodthirsty).
I would make more sense to block plundered ships from becoming targets.

tehhowch:
Player escorts already satisfy the `disables` check. You will also not autofire upon ships you own.

In general, if you are proposing code changes that alter behavior, you are required to verify that your changes, at minimum, have the intended effect. If you are unable to compile and test your changes, ensure you have a collaborator who can, and that they do.

In 0.9.9 there are now guards that consider splash damage when autofiring - please clarify in your issue and here which commit you are basing your observations on.

manitscold123:
I didn't know that there are guards in place in 0.9.9. In fact, I don't actually even know what's present in 0.9.9. Rip.

PR #3702, opened 2018-04-14
new map region labels

Aurelite:
second half of #3684, labels only

screenshot from 2018-04-14 20-40-45


jafdy:
I don't know, I like the cursive labels.

Amazinite:
The new galaxy image was a clear improvement over the old one, but these new map labels just seem like a matter of preference to me.

Zitchas:
The map itself looks good, but I was also partial to the cursive. I agree with Amazinite, though. Cursive vs block is just personal preference.

KiLEdEnNis:
that and it's harder to see now, the current labels look like you have some space pirate treasure map, which looks cool, while this doesn't really fit

PR #3703, opened 2018-04-14
Hotkey improvements for PlayerInfoPanel, ShipInfoPanel, TradingPanel

Aurelite:
**PlayerInfoPanel:**
- added key **i** to close the panel; it can now be toggled with **i**, analogous to the map with **m**  (before it opened the map)
- added key **p** to toggle park/unpark selected ships
- added key **a** to toggle park all/unpark all  
   (is there a reason these were not activated in the first place?)

**ShipInfoPanel:**
- added key **r** to rename the ship

**TradingPanel:**
- changed _Buy One_ to **Left-Arrow** from **Return**  (there's also **Space**)
- changed _Sell One_ to **Right-Arrow** from **Backspace**  (there's also **Delete**)
   (you could already scroll through the commodities with **Up** and **Down**)
- changed _Buy All_ to **Return** from **Shift-b**  (since **Shift** is already used as x5 modifier and **b** for Bank)
- changed _Sell All_ to **Backspace** from **Shift-s**  (... **s** used for Spaceport)

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -883,7 +883,7 @@ interface "info panel"
 		from 250 -290 to 500 280
 	
 	visible if "ship tab"
-	button R
+	button r
 		center -375 -270
 		dimensions 250 30
 	sprite "ui/ship tab"
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -934,7 +934,7 @@ interface "info panel"
 	active if "can park"
 	sprite "ui/dialog cancel"
 		center -55 305
-	button P "Park"
+	button p "_Park"
 		center -55 305
 		dimensions 70 30
 	active
@@ -942,7 +942,7 @@ interface "info panel"
 	visible if "show unpark"
 	sprite "ui/dialog cancel"
 		center -55 305
-	button P "Unpark"
+	button p "Un_park"
 		center -55 305
 		dimensions 70 30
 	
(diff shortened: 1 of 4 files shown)


endless-sky:
This would make it so that clicking the "next" and "previous" buttons in the player info panel no longer works. It would also make it so you can't click the ship name to rename it.

Even aside from the bugs, in general I'd prefer not to change keyboard shortcuts that many players will have gotten used to.

Aurelite:
I fixed the rename button and reverted the changes to the trading panel.
The "next" and "previous" buttons on the "ship info" tab are unaffected by these changes. Both clicking and hotkeys still work. 

These are only additional hotkeys now, no changes besides making i toggle the player info panel.
I just found i really annoying not being able to toggle the player info panel with i, while i can toggle the map with m.

I like to go through the ship list with the mouse and quickly park/unpark specific ships via hotkey, instead of ctrl-clicking them and then click park/unpark.

tehhowch:
When you are viewing the `ShipInfoPanel`, the **`n`** and **`p`** keys are bound to the `Next` and `Previous` buttons:
image

https://github.com/endless-sky/endless-sky/blob/d2e19f154a37c267562192847a720fd4a36e3c1b/data/interfaces.txt#L948-L970

Your PR changes the `SDL_Keycode` emitted by the UI's "Park" and "Unpark" buttons, from **`P`** to **`p`**. This means that ShipInfoPanel::KeyDown now receives **`p`** when you click Park/Unpark. Note that **`p`** is *also* emitted when you click the UI's  "Previous" button. Since the "Previous" case is handled first, there is no way for the player to park/unpark when viewing a specific ship - instead they navigate to the previous ship:

https://github.com/endless-sky/endless-sky/blob/d2e19f154a37c267562192847a720fd4a36e3c1b/source/ShipInfoPanel.cpp#L116-L127

As there is no preexisting key associated with **`p`** in the `PlayerInfoPanel::KeyDown` code, your statement re: next & previous UI buttons, and **`n`** and **`p`** still functioning as "Next" and "Previous" when viewing the `ShipInfoPanel`, is accurate.

PR #3704, opened 2018-04-14
reworked energy and heat balance table in ShipInfoDisplay

Aurelite:
The energy and heat balance in the outfitter's info panel is a bit inconsistent and hard to make use of without a calculator. And even then it is missing an important information: The heat capacity.

Lets take a look at the default Firebird.

**What we have now:**
old-balance
- _max_ lists two different quantities, a capacity for the energy and a dissipation rate for the heat
- What we really want to know is how long the energy lasts
   So we have to calculate 4800 / (336-180-594-30) = 10s, and do this every time we change the outfit
- for the heat we have the max rate so we calculate (1194 + 312 + 1920) - 2289 = 1137
   We produce more heat than we dissipate, so we'd like to know how long we have before we overheat
   Unfortunately there is no way of knowing without the formula for the heat capacity and looking into the ship definition for the dissipation factor

**This PR does the following changes:**
new-balance
- displays the heat dissipation as negative _idle_ heat rate
- adds a _sum_ row
- displays the _heat capacity_ (remaining after idle heat level) next to the _energy capacity_
- calculates the _time_ until the capacity is drained/reached under full load (in case of negative balance)

**Explanations in the tooltips:**
- sum: `Energy and heat balance with all the ships systems running. With a negative energy balance having some energy storage capacity is advised. A positive heat balance can eventually cause the ship to overheat.`

- capacity: `Energy and heat storage capacity. Energy capacity allows a ship to temporarily draw more energy than it produces. The heat capacity is the heat a ship can build up before overheating.`

- buffer time: `Time until the energy storage is depleted or the ship overheats when all systems are running.`

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -299,9 +299,6 @@ tip "sum:"
 tip "capacity:"
 	`Energy and heat storage capacity. Energy capacity allows a ship to temporarily draw more energy than it produces. The heat capacity is the heat a ship can build up before overheating.`
 
-tip "buffer time:"
-	`Time until the energy storage is depleted or the ship overheats when all systems are running.`
-
 
 
 # Outfit categories:
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -293,8 +293,14 @@ tip "repairing hull:"
 tip "charging shields:"
 	`Energy consumed and heat generated when this ship is recharging its shields. This is in addition to the "idle" amount.`
 
-tip "max:"
-	`Energy storage capacity and maximum safe level for heat generation. Energy capacity allows a ship to temporarily draw more energy than it produces. The heat maximum is the highest your heat generation can be without eventually causing your ship to overheat.`
+tip "sum:"
+	`Energy and heat balance with all the ships systems running. With a negative energy balance having some energy storage capacity is advised. A positive heat balance can eventually cause the ship to overheat.`
+
+tip "capacity:"
+	`Energy and heat storage capacity. Energy capacity allows a ship to temporarily draw more energy than it produces. The heat capacity is the heat a ship can build up before overheating.`
+
+tip "buffer time:"
+	`Time until the energy storage is depleted or the ship overheats when all systems are running.`
 
 
 
--- a/source/Format.cpp
+++ b/source/Format.cpp
@@ -124,7 +124,10 @@ string Format::Decimal(double value, int places)
 	double integer;
 	double fraction = fabs(modf(value, &integer));
 	
-	string result = to_string(static_cast<int>(integer)) + ".";
+	string result = to_string(static_cast<int>(integer));
+	if(places)
+		result += ".";
+	
 	while(places--)
 	{
 		fraction = modf(fraction * 10., &integer);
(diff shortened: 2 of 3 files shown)


Amazinite:
I'll be glad if I no longer have to do these calculations myself every time I'm outfitting highly efficient ships. What does buffer time show as if the ship can never run out of energy or overheat with its loadout?

Aurelite:
It shows "-" if it is stable. Unfortunately the is no infinity symbol. Any better ideas on what to show?

Amazinite:
I think just showing `-` is fine (maybe "inf" could work too), but perhaps the buffer time tooltip could explain what that means.

ghoulavenger:
I understood the energy portion well enough without this. But I do think this will help with outfitting ships anyway so I'm for it. On the heat front, I'm a little bit confused. It is quite possible that I have been overcooling my ships by a lot. I don't mind running a cool rig mind you but maybe this will help fix some errors in my loadouts (outfit space is precious after all).

Hacklin:
Two minor points:
- In the buffer time tool-tip: Instead of `Time` could you say `Seconds`.  I find it clearer.
- In the outfitter's info panel: Could you entertain the thought of displaying the `buffer time:` with 1 decimal after the decimal point?  Displaying smaller changes helps with showing the effect an outfit has.

endless-sky:
This would make it so that some ship stats don't fit in the ship info panel.

Also, I think that the number of people who would be confused by those added lines would be at least as many as those who would find them useful - especially for new players, who are already at risk of being overwhelmed with the complexity of the available information.

Also, I don't think your heat buffer time calculation is accurate. You would have to take into account the fact that heat dissipation rate is proportional to your ship's current heat level, as well as the fact that the heat level may not start out at 0.

Aurelite:
> This would make it so that some ship stats don't fit in the ship info panel.

Oh i see, its one line too many for ships with fighter bays. So removing one line should fix it or are there other ships that would break it?

> Also, I don't think your heat buffer time calculation is accurate. You would have to take into account the fact that heat dissipation rate is proportional to your ship's current heat level

It's indeed incorrect, i missed that one. The dissipation increases exponentially with the heat level, which makes calculating the time a [bit more difficult](https://en.wikipedia.org/wiki/Lambert_W_function). I'm removing the buffer time.

I find the current display not intuitive at all though with the max line showing to completely different measures. Also its not really clear what that heat value even means or how much of a positive heat balance is tolerable. By dividing the capacity by the sum you get at least a minimum time before overheating.

Amazinite:
What does `heat capacity` show now? Before, heat capacity was the maximum heat sum that one could have while still being able to launch. Now the number under this PR looks far different.

Aurelite:
The ship in the screen shot has different weapons from the one in the first post.
Otherwise its still the same minus the buffer time.

The values in the heat column until sum are `heat-units/second` that you produce.
The capacity is the amount of heat units you can accumulate until you overheat.
It is `capacity = maximum heat - idle heat` with `maximum heat = ship mass * 100` and the idle heat level is determined by your outfit.

A ship with an idle heat level above the maximum heat level has 0 capacity and is unable to launch.

It is analogous to the energy capacity except that you fill it up from 0, instead of draining it from full.

PR #3705, opened 2018-04-15
Adds extra response to the "Alondo is gay" revelation

elgeonmb:
As a huge lesbian I thought it kind of strange that, in revealing that Alondo is gay, the game implies that the PC _isn't_ (by sending you on the diplomacy mission that Alondo is too gay to go on). Since ES aggressively avoids implying anything about the PC other than their place of birth and vague age/backstory elements, I figured I'd add a response to Freyja that allows you to dispel that implication.

### DIFF
--- a/data/free worlds middle.txt
+++ b/data/free worlds middle.txt
@@ -1596,7 +1596,7 @@ mission "FW Albatross 1"
 			`	"Yes," she says. "Yes, it does. Now, let's go meet up with JJ."`
 				accept
 			label metoo
-			`	Freya raises an eyebrow. "I hadn't heard. Let's hope nobody on Albatross had heard either."`
+			`	Freya raises an eyebrow. "I hadn't heard. Try not to say that so loudly when you're on Albatross."`
 				accept
 
 
--- a/data/free worlds middle.txt
+++ b/data/free worlds middle.txt
@@ -1589,11 +1589,15 @@ mission "FW Albatross 1"
 			choice
 				`	"Okay, I feel silly for even asking. Let's go meet up with JJ!"`
 					accept
+				`	"So am I."`
+					goto metoo
 				`	"Ah. That certainly explains a lot."`
 				`	"It seems crazy that anyone would be bothered by that in this day and age."`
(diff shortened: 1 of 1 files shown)


Amazinite:
Good catch, but I think it would be more interesting to have the player be able to completely avoid this mission if they say that they're homosexual. Right now, I'd find this situation odd if I was told that Alondo couldn't go to Albatross because of the danger his sexuality could pose, but after saying "I'm gay as well," Freya just throws me into the situation without me being able to ask to do something else rather than put myself in danger.

elgeonmb:
Took Kiko's suggestion. While the idea of just not doing the Albatross mission because you're gay has some appeal, I'd rather keep the whole mission chain streamlined.

KiLEdEnNis:
to me, the game doesn't assume one isn't gay, just no one in-game would know you are, hence it's safe for you to go

kikotheexile:
If gender, identity, or sexuality is addressed you should have the right to respond to it. 

Most games dont assume anything of the player character, they just assume the player themself has nothing to say about it. Tbh i also think that the in a perfect world the player dialogue should be able to say "eww." 

But tbh there's already enough of that in the world and we dont want to reinforce that kind of behavior in the community.

elgeonmb:
Yeah I did think it was a LITTLE weird that the PC can't go "ACTUALLY I'M ALSO A HUGE HOMOPHOBE" but I feel like that would be a lot more disruptive than the converse.

Plus I mean, you _did_ throw in with a socialist revolution so assuming you're pretty socially progressive isn't too off base.

elgeonmb:
hey the "close and comment" button looks very similar to "cancel comment"

endless-sky:
I'm not opposed to a change like this, but the "So am I" dialog option doesn't really flow right because it's in response to a statement two sentences earlier. Maybe something else would flow better like, "Actually, I'm gay too," or, "If that's the case, I think I won't be welcome there any more than Alondo would be."

Alondo's sexuality is a remnant of my original plan for each of the three factions to be diverse in some ways but closed-minded in others: the Free Worlds being tolerant in that respect but pretty homogeneous racially and religiously, for example. Those themes don't yet get fleshed out much in the actual game, though.

PR #3710, opened 2018-04-21
Add hostile militia retorts

Ferociousfeind:
Add another "hostile militia" phrase, because the prisons in the Rim have some rather empty cells.


data/hails.txt
```diff
@@ -3112,6 +3112,62 @@ phrase "hostile militia"
...
+		"Flood"
+		"Longjump"
+		"Glaze"
+		"Deep"
+		"Clink"
+		"Winter"
+		"Solace"
+		"Twinstar"
+		"Sundrinker"
+		"Trinken"
```
Bladewood: hmm...

jafdy: Had you been drinken when you wrote that?

MCOfficer: skål

endless-sky:
"The prisons are rather empty" seems like a rather vague way of phrasing the threat - maybe something like "I hope you like prison food" would be clearer.

(Also, the fact that there isn't an actual game mechanic where you can be boarded and take prisoner could make this message confusing to players.)

Pointedstick:
I agree with @endless-sky. @Ferociousfeind can you re-work this PR according to review comments, or else close it?

data/hails.txt
```diff
@@ -3112,6 +3112,64 @@ phrase "hostile militia"
...
+		"We "
+	word
+		"hope "
+		"wonder if "
+	word
+		"you "
+		"criminals like you "
+		
+	word
+		"like"
```
Amazinite: -> `"like "`

Ferociousfeind: Oh, I see your point now. How didn't I catch that??

jafdy: Same way everyone else doesn't catch mistakes, too many lines for mistakes to slip into. This is why we review everything.

PR #3718, opened 2018-04-28
Expanded Friendly Disabled Hails

OverYrPaygrade:
The hails from disabled friendlies were quite lacking. Only a single hail could be used.

This change expands on this greatly and overall enriches what they have to say about their situation.


tehhowch:
You've got some capitalization mismatch that need resolving, e.g. `Those Artless` and `...!   <lowercase phrases>`

data/hails.txt
```diff
@@ -2030,7 +2030,232 @@ phrase "friendly author"
...
+		"could you lend a hand"
+		"can you get us back on our feet"
+		"would you help us"
+		"any chance you could get us up and running again"
+		"could you render assistance"
+		"can we get a little help here"
+	word
+		"?"
+		" please?"
+		"? I don't wanna get standed here!"
```
Darcman99: "standed" meant to be "stranded"?

data/hails.txt
```diff
@@ -2030,7 +2030,232 @@ phrase "friendly author"
...
+		"Hey! "
+		"S.O.S! S.O.S! "
+	word
+		"Please come board our ship and patch us up!"
+		"We would appreciate any help getting us running again!"
+		"We can't fix this our own!"
+		"Could you come over and help us please?"
+		"Requesting emergency repairs!"
+		
+phrase "friendly disabled"
```
KiLEdEnNis: the possibilities of this phrase can be like
"Hey, Our ship is disabled, "
^ capital letter after a comma, no
also, the first AND second word can be empty, creating "could you lend a hand?" which has no capital letter at all
go through these hails and make sure this stuff doesn't happen

CyberShadow:
If the current message didn't explicitly mention boarding, I don't think I would have figured out how to actually help them. I think that every message ought to mention boarding somehow, or a help text could be accompanied with the first time such a message is seen by the player.

CyberShadow:
@OverYrPaygrade Looks like your last commit there changed the file's line endings.

OverYrPaygrade:
...I don't get it.

Why does it say everything has been changed? It hasn't...
I don't know what to do.

endless-sky:
All the text files in the game use Unix line endings (line feed character alone). Your text editor modified the file to have Windows-style line endings (carriage return followed by a line feed). You should use a text editor (like Notepad++) that will avoid changing the line endings.

PR #3720, opened 2018-04-28
Fast boolean options

CyberShadow:
This PR optimizes help and setting code to use enums and static class fields instead of string lookups, thus cutting down on a LOT of unnecessary string comparisons.

This change makes follow-up PRs which introduce preferences that would be accessed many times per frame more palatable.

Please see the commits' individual commit messages for extended details.

An alternative version of the second commit, which uses an enum of all settings, is available here:
https://github.com/CyberShadow/endless-sky/commits/fast-boolean-options-v1
It has the benefit of deduplicating up the preferences panel and preferences class string literals, but the code is overall uglier.

### DIFF
--- a/EndlessSky.cbp
+++ b/EndlessSky.cbp
@@ -172,6 +172,8 @@
 		<Unit filename="source/HailPanel.h" />
 		<Unit filename="source/Hardpoint.cpp" />
 		<Unit filename="source/Hardpoint.h" />
+		<Unit filename="source/Help.cpp" />
+		<Unit filename="source/Help.h" />
 		<Unit filename="source/HiringPanel.cpp" />
 		<Unit filename="source/HiringPanel.h" />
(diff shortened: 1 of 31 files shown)


Hacklin:
ES uses short string literals everywhere as arguments.
Ideally, you want to convert all these literals to indexes at compile-time, like a `Perfect Hash`.
So you could keep using the string literals.
This would require either an external program that runs just before compilation or some template wizardry (maybe) or C++2y constexpr containers (hopefully).

Instead of changing string literals to enums, have you tried to use `Dictionary.{h,cpp}`?
This Dictionary was added to speed up the string literal look-ups.
Just make it a template, so you can use bools instead of doubles.

Hacklin:
The `Refactor help texts into an enum` does what it says on the tin.

Issue #3714, opened 2018-04-27
Shaders are too slow on integrated graphics


CyberShadow:
Hi and thanks for the amazing game!

I'm playing this game on a laptop with Intel integrated graphics, and noticed that my laptop gets pretty hot (and the fans get pretty loud) in certain situations while playing the game. Specifically, this seems to occur on the map screen, and when there is a lot of text on the screen (such as during dialog).

Building the game from source and profiling it reveals that the culprits are:
- Font::DrawAliased
- RingShader::Draw
- LineShader::Draw

All three seem to make use of shaders, so I'm inclined to believe that the shaders create a performance bottleneck on machines like mine.

As these primitives look like they could be implemented in other ways, what do you think about having an alternative implementation of them that doesn't make use of shaders?

KiLEdEnNis:
if you know how to do it, feel free to try, better game performance is greatly appreciated
and if you don't know how, thanks for bringing it up anyway, and hope someone finds a way to improve it

tehhowch:
Map code is the oldest code. If you are able to write GPU code or develop a less resource intensive equivalent of the existing map, please do

CyberShadow:
PRs: https://github.com/endless-sky/endless-sky/pull/3720 https://github.com/endless-sky/endless-sky/pull/3721 https://github.com/endless-sky/endless-sky/pull/3724

With these three, there is a significant reduction in heat and fan noise.

endless-sky:
I've checked in a change (7d74467) that takes a different approach than those PRs: rather than trying to optimize the map access functions, I just cache the map layout itself and only update the cache when the coloring changes.

With that change, on my computer, map drawing uses significantly less CPU. Rather than data lookups being the bottleneck, the biggest remaining bottleneck after the change is the text drawing code itself.

Optimizing text rendering is a separate issue. Once the game switches to full Unicode support, lines of text will need to pre-rendered and cached. In the meantime, there might be other ways to optimize (like using vertex buffers to draw a whole line of text with one draw call, instead of drawing one character at a time).

CyberShadow:
Thank you, that does seem to make a noticeable difference on my laptop!

PR #3721, opened 2018-04-29
Add naive shader-less paths for some graphics primitives

CyberShadow:
This is a start on #3714.

~~Requires (and includes) #3720.~~

Though the performance gains are measurable, my laptop is still hot and loud with these patches. Additional profiling revealed that there is a lot of map data being computed every frame that could be trivially cached, will follow up with that in a different PR.


tehhowch:
It is customary and expected to avoid mixing PR content when possible. For example, you can easily separate the contributions of #3720 from this separate feature addition (i.e write this with the existing `Preferences` code).
Why? Because if a different approach for revamping Preferences is taken, this PR has a massive conflict. This also may be more readily merged than a change to the Preference system.

Only if two features are so inextricably linked or otherwise interwoven should a single contributor PR add them both.

(PS your text editor removes whitespace. Disable that functionality for this repo.)

CyberShadow:
> For example, you can easily separate the contributions of #3720 from this separate feature addition (i.e write this with the existing Preferences code).

In theory; in practice, this PR pretty much depends on #3720, and I wouldn't trust any performance measurements taken without it. Though the setting's value can be cached in a global somewhere at the start of a frame, there is no obvious place to put it so it would need to be a temporary hack.

> Why? Because if a different approach for revamping Preferences is taken, this PR has a massive conflict.

That's fine, I'll be happy to rebase.

> PS your text editor removes whitespace.

No, any whitespace changes were done by hand. I don't see any in the diff though.

CyberShadow:
OK, removed the #3720 stuff. The original is here, please use it for benchmarks instead: https://github.com/CyberShadow/endless-sky/commits/fewer-shaders-v1

tehhowch:
For example, line 115 should have a tab indent level of 1. line 121 should have a level of 2, etc.

source/Font.h
```diff
@@ -71,6 +71,7 @@ class Font {
 	GLint aspectI;
 	GLint positionI;
 	
+	int width;
```
Hacklin: Forgot to initialize `width` in the member initializer list of the default constructor `Font::Font()` in `Font.cpp`.

Why use a constructor's member initializer list (in `Font::Font() :` define `width(0),`) instead of default member initialization (in the `class Font {` declare `int width = 0;`) and defaulting the default constructor (`Font::Font() = default;`)?
The latter is less error prone [and can be optimized in the STL].

CyberShadow: Thanks, fixed. I'll leave the initialization strategy for another time, though.

CyberShadow:
> For example, line 115 should have a tab indent level of 1. line 121 should have a level of 2, etc.

Re-added the whitespace :)

CyberShadow:
☹

The shader-less code is much shorter, though (the diff is big mostly due to the shader implementations having been reindented).

There's also the case of font rendering, which looks almost visually-identical with the GL_QUADS implementation: is a fragment shader really as efficient as using basic blending operations on most machines?

endless-sky:
`GL_QUADS` isn't even supported by newer OpenGL versions. I'd prefer not to have a feature that won't even work on some systems, and that is likely to be even less widely supported in the future.

CyberShadow:
Ah, no problem, I can redo it with GL_TRIANGLES. Would you be OK with this change then?

PR #3724, opened 2018-04-29
Precalculate Planet::IsInhabited

CyberShadow:
This + #3720 + #3721 finally get my laptop much quieter and cooler on the map screen.

### DIFF
--- a/source/Planet.cpp
+++ b/source/Planet.cpp
@@ -188,6 +188,8 @@ void Planet::Load(const DataNode &node, const Set<Sale<Ship>> &ships, const Set<
 		else
 			attributes.erase(AUTO_ATTRIBUTES[i]);
 	}
+
+	inhabited = (HasSpaceport() || requiredReputation || defenseFleet) && !attributes.count("uninhabited");
 }
 
 
@@ -279,7 +281,7 @@ const string &Planet::SpaceportDescription() const
 // have the "uninhabited" attribute).
 bool Planet::IsInhabited() const
 {
-	return (HasSpaceport() || requiredReputation || defenseFleet) && !attributes.count("uninhabited");
+	return inhabited;
 }
 
 
--- a/source/Planet.h
+++ b/source/Planet.h
@@ -142,6 +142,7 @@ private:
 	double requiredReputation = 0.;
 	double bribe = 0.01;
 	double security = .25;
+	bool inhabited;
 	
 	int tribute = 0;
 	const Fleet *defenseFleet = nullptr;


source/PlayerInfo.h
```diff
@@ -304,6 +311,7 @@ class PlayerInfo {
 	std::set<const System *> seen;
 	std::set<const System *> visitedSystems;
 	std::set<const Planet *> visitedPlanets;
+	mutable std::map<const System *, bool> seenCache, nameKnownCache;
```
Hacklin: Declare one name (only) per declaration.
Multiple names per declaration is considered bad form in C++ (and may even be against ES style guide).

Hacklin:
Do you really need to add a cache?
Isn't it enough to just change the current code?

For instance:
`HasSeen(System)` should check `seen.count(system)` first instead of last.
If the player has seen the system, then `HasSeen(System)` is just as fast uncached.

`HasVisited(System)` is a subset of `KnowsName(System)` is a subset of `HasSeen(System)`.
- This could be used to put the visited/knows/seen info in a single flags/bitset value.
This would be faster, IF**F** a function needs to check 2 or more bits of this value.
- The mission checks from `KnowsName(System)` could be added into `HasSeen(System)`.
This way the missions need not be visited twice and the `HasVisited(System)` check would be skipped.  However, this will add a **hidden** code duplication; i.e. if the implementation of `KnowsName(System)` changes, then `HasSeen(System)` must change too! (Document in comment!)

source/Planet.cpp
```diff
@@ -188,6 +188,8 @@ void Planet::Load(const DataNode &node, const Set<Sale<Ship>> &ships, const Set<
 		else
 			attributes.erase(AUTO_ATTRIBUTES[i]);
 	}
+
+	inhabited = (HasSpaceport() || requiredReputation || defenseFleet) && !attributes.count("uninhabited");
```
Hacklin: OK.
A planet's inhabited status is only changed in the `Planet::Load()`, oddly enough.

CyberShadow:
> MapPanel::DrawSystems() currently determines which systems need to be drawn on every step, when this is something that could be determined in the MapPanel constructor, and then just referenced again for the duration of that panel. I.e., a better implementation of map caching would be to add a method MapPanel::DetermineDrawnObjects() that computes all the wormholes links and systems / system links that get drawn, and then all the Draw___() methods simply determine the position and colors, rather than re-check the eligibility.

My first implementation was very much along those lines:

https://github.com/endless-sky/endless-sky/commit/015658948a64ca4bd40bbcb12269bf2ce5200091

I ended up redoing it as a general cache as it felt pretty silly to add caching for just one place in the game, whereas this allows the entire game code to benefit from it.

source/PlayerInfo.cpp
```diff
@@ -2090,6 +2111,28 @@ set<string> &PlayerInfo::Collapsed(const string &name)
 
 
 
+bool PlayerInfo::GetCached(const System *object, std::map<const System*, bool> &cache, bool (PlayerInfo::*func)(const System*) const) const
+{
+	auto it = cache.find(object);
+	if (it == cache.end())
+	{
+		bool value = (this->*func)(object);
+		cache.insert(make_pair(object, value));
```
Hacklin: Why do you use `cache.insert(make_pair(object, value))` instead of `cache.emplace(object, value)`?

Note that the generated code is identical, so it doesn't matter.
I'm just curious to why you choose to be more verbose.

CyberShadow: Wasn't aware of that C++11 addition. Fixed.

CyberShadow:
Rebased & fixed a bug with minimap not updating during a jump to a new system.

endless-sky:
With any optimization, there's a tradeoff between how much it speeds things up versus how much complexity (and potential for bugs) it adds to the code. That's especially true in the map code, because how efficiently the map draws is not as critical as how fast space battles draw.

The optimization of caching a planet's inhabited state makes a lot of sense, from that perspective - based on my own profiling, `Planet::IsInhabited()` is at least 10% of the execution time for map drawing, and it's only a change to a couple lines of code.

The rest of this, on the other hand, makes the code a bit more fragile (dependent on the cache being updated when appropriate). Also, I don't think it's the cleanest solution - it would be far better to just generate a cached representation of the map when the map panel is opened, rather than having it persist and therefore having to figure out all the situations where it needs updating.

CyberShadow:
OK, dropped the PlayerInfo cache commit, leaving only the `Planet::IsInhabited` optimization; will redo map prepcalculation in a new PR later.

Issue #1553, opened 2016-09-12
Improvements to Firefly Mode (dangerous space weather, etc.)


Pointedstick:
I saw Michael talking about how he envisions this game having essentially three modes, one of which is "Firefly Mode"--roughly corresponding to the beginning of the game, where you're small and weak and just trying to get by. However, it's far too easy right now, so Firefly mode doesn't last very long! The average player will quickly and easily overcome the basic limitations of the start of the game: your mortgage, small ship, and dangerous space pirates. You're really only in "harried star captain mode" for maybe an hour before your debts are paid off, you've got a better ship, and you're on an upwardly mobile trajectory. I think it could be made more fun and evocative in a number of ways:
- Make it harder or less lucrative to get rich quickly by selling captured outfits: #1535, #227 
- Make money truly more scarce. Other captains are all griping about money but the ease with which you transcend your initial poverty makes them seem like a bunch of whiners. Lots of ideas here: #275 #329 #486 #428
- More options to do things that seem rough or desperate: scavenging space debris, mining asteroids, listening to local gossip from old spacers, gambling, etc. #325 #1552
- Make the galaxy itself seem more hostile and threatening. Asteroid storms that can damage the player's ship; systems with high radiation levels that prevent passage; Navy quarantines that reinforce their awesome power; annoying bureaucrats and rules to make you feel more sympathy with the Free Worlds, etc.

toilethinges:
Mining asteroids is already in.

On the very last part, then we'd need more tales of horror about the Free Worlds at the same time, once the Republic and Syndicate arcs are out.

Otherwise, agreed.

endless-sky:
One difficulty is that already there are a lot of people who find the early game so challenging that they give up - which, I know, is hard for a more experienced player to imagine. Like, there are plenty of people who post on Steam who still have not paid off their mortgages after four hours.

The game originally had licenses that the Republic required you to pay for to get almost any type of ship. For example, buying a bulk freighter would require a capital ship license and a freight license in addition to the pilot's license. That gave the player a better sense of how annoying Republic bureaucracy is - but it was also a huge pain and made the game less fun.

Pointedstick:
[...]

Issue #1552, opened 2016-09-12
More reasons to go to the spaceport, and more to do there


Pointedstick:
Right now the only reason to go into the spaceport is the chance to get a mission. If you don't, then there's just a bit of text--nothing else to do.

In the old Escape Velocity games, the spaceport bar also gave you the opportunity to hire escorts, gamble, or read the news. Stuff like this would be great. Additional ideas:
* Buy some old spacers a drink to hear local gossip
* Get in a fight with pirates over something that causes you to gain some money, but your ship immediately takes off along with some weak pirate interceptors
* Look for opportunities to buy illegal outfits
* Look for opportunities to transport illegal goods

Amazinite:
A similar idea I had a while back: random events that can occur throughout human space that the player can only know about if they run into them, or if they decide to visit the spaceport to check out the news board.

Such events could be a sudden increase in piracy in a certain part of space (with bounties over the heads of a few pirates), trade changes between two parts of space that affect commodity prices (so pretty much commodity fluctuations to one or two commodities that don't require the player to be trading a lot to occur), and to go along with your Firefly idea of increasing how dangerous the universe is, have an event where a star may start having a massive spike in coronal activity that irradiates the system for a few days, lowering the fleet frequencies in the system and causing light damage when in the system.

jafdy:
Every thing you want can be accomplished with more missions that are taken in the space port.

Pointedstick:
Yes, but only very awkwardly, in much the same manner that you could (and people did) add asteroid mining by making a government whose ships looked like asteroids that dropped cargo when destroyed that looked like various minerals. The native implementation is much smoother and better.

jafdy:
Certainly points 2 and 4 can. eg
Drug Running missions
`mission "fight pirates"
on offer
   conversation
      Some pirates are gambling in the bar. Would you like to join in?
      choice
         yes
         no
            decline
      You win some money but the pirates now want to kill you.
         accept
on accept
   payment 1000
fleet "small pirate"`

For point 1 it would be an auto decline mission with a negative payment and some text.

Fair enough on point 3.

Hiring escorts is mostly a matter of how to do it. I set something up but the negative payments don't work properly in the job board.  

With gambling (not in above mission) and the things @Amazinite mentioned a new mechanic would probably be best.

Amazinite:
[...]

PR #3729, opened 2018-05-02
Add electromagnetic storm missions

Jacunn:
The Ember Waste is a unique region of space, but it's mostly just a different colour and has lots of special wormholes. What if there were missions that took place here, that involve electromagnetic storms? Essentially, the concept is that overtime, charged particles in the waste reach higher concentrations in certain regions, resulting in electromagnetic storms. It adds a new twist to deal with in a combat mission that I've included in the Ember Waste. This is probably far from ideal in it's current state, but it's a proof of concept.

### DIFF
--- a/emstorm
+++ b/emstorm
@@ -166,7 +166,6 @@ mission "Remnant: Stranded Korath"
 		fleet "Korath Raid"
 		dialog "All Korath in this system have been eliminated."
 	on complete
-		event "overload cannon available" 30
 		payment 3000000
 		conversation
 			`The woman you met earlier greets you as you land, with her usual calm voice`
--- a/emstorm
+++ b/emstorm
@@ -1,5 +1,4 @@
 # Ships
-
 ship "Electromagnetic Storm"
 	attributes
 		"hull" 100
@@ -16,14 +15,12 @@ ship "Electromagnetic Storm"
 		"thrust" 10
 		"thrusting energy" .3
 	outfits
-		"Storm"
-		
+		"Storm"	
 	engine -17 35
 	engine 17 35
 	turret 0 0 "Storm"
 	
 # Outfits
-
 outfit "Storm"
 	category "Turrets"
 	cost 0
@@ -45,21 +42,21 @@ outfit "Storm"
 		"ion damage" 0.1
 		
 # Fleet
-fleet "Electromagnetic(weak)"
+fleet "Electromagnetic (Weak)"
 	government "Korath Nanobots"
 	personality
 		heroic
 		vindictive
 	variant
-		"EMStorm" 8
+		"Electromagnetic Storm" 8
 		
-fleet "Electromagnetic(moderate)"
+fleet "Electromagnetic (Moderate)"
 	government "Korath Nanobots"
 	personality
 		heroic
 		vindictive
 	variant
-		"EMStorm" 15
+		"Electromagnetic Storm" 15
 		
 fleet "Electromagnetic(strong)"
 	government "Korath Nanobots"
@@ -67,7 +64,7 @@ fleet "Electromagnetic(strong)"
 		heroic
 		vindictive
 	variant
-		"EMStorm" 22
+		"Electromagnetic Storm" 22
 		
 # Mission
 mission "Remnant: Stranded Korath"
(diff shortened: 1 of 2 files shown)


Amazinite:
Related: https://github.com/endless-sky/endless-sky/issues/1553, https://github.com/endless-sky/endless-sky/issues/1552#issuecomment-246232974

emstorm
```diff
@@ -0,0 +1,183 @@
...
+	variant
+		"EMStorm" 22
+		
+# Mission
+mission "Remnant: Stranded Korath"
+	landing
+	source
+		government "Remnant"
+	name "Korath ships stranded in Remnant Territory"
+	description "There are Korath fleets located in the Caeculus, Stercutus and Segesta systems. Find and eliminate them, then return to <planet> for payment."
```
tehhowch: You can use #3336 here.

emstorm
```diff
@@ -0,0 +1,183 @@
...
+		has "event: remnant: void sprite research"
+		has "Remnant: Key Stones: done"
+		has "Remnant: Defense 3: done"
+		has "Remnant: Bounty: done" >= 3
+	waypoint Caeculus
+	waypoint Stercutus
+	waypoint Segesta
+	on offer
+		conversation
+			`As you land, you are greeted by a man and a woman, both dressed in quite a striking manner, perhaps as a sign of distinction. They collectively begin to chant in an almost gleeful tone`
```
tehhowch: You are missing a **lot** of punctuation in your conversation text, particularly before character speech.

emstorm
```diff
@@ -0,0 +1,183 @@
...
+					accept
+				`	"Sorry, but that sounds too dangerous."`
+					decline
+	npc evade
+		personality
+			heroic
+			vindictive
+		government "Korath Nanobots"
+		system Caeculus
+		fleet "EMStorm"
```
tehhowch: A `fleet` with this name is never defined.

Zitchas:
I like the idea of having electromagnetic storms in the Ember Wastes. It'd be neat to have something other than the background colour to change up the environmental factors.

Jacunn:
Oops, I refactored names poorly! I'll fix this up and resubmit it! :)

tehhowch:
Don't resubmit it, just add commits to the branch you are using that address the issues.

If you resubmit the PR, you separate the new one from all previous feedback (and also pollute the repo, since you can't delete an issue or PR, just close it).

robot-rover:
Perhaps similar to the way the swarm of nanobots from the Korath work?

emstorm
```diff
@@ -45,29 +42,29 @@ outfit "Storm"
...
+fleet "Electromagnetic (Moderate)"
 	government "Korath Nanobots"
 	personality
 		heroic
 		vindictive
 	variant
-		"EMStorm" 15
+		"Electromagnetic Storm" 15
 		
 fleet "Electromagnetic(strong)"
```
KiLEdEnNis: you made
> fleet "Electromagnetic (Strong)"

so put a space and capitalize the s in "strong" to make that work

Kriegstofu:
keep in mind that this might influence/interfere the gascraft mission

Jacunn:
As it stands the storms will not kick in until after the void sprite mission has been completed, among others. This is to ensure that there won't be any interference with other missions that are possibly future story-line critical.

Zitchas:
I'd like to dig into the content a bit more to see how this is actually working, but my thoughts are so far:
I really like the idea of having some kind of "natural hazard" in the Ember Wastes. Ideally, something that is fairly nasty but not necessarily lethal throughout the Ember Wastes, but weaker (or possibly almost non-existant) in the core Remnant systems. Which would make the Remnant both protected by it and penned in by it (and thus is becomes fodder for missions of them trying to figure out a way to reduce its effect on their ships). That being said, I'd prefer to have it in effect right from the beginning of the game - it would help explain why Hai explorers or adventurous humans with access to the crystals haven't bothered to spend more time in the Ember Wastes. After all, the Remnant got through that wormhole. If they can, someone else should have been able to too. That being said, having it start out at a low level of danger and ramp up at some point would be even better, as that provides for a different kind of challenge and mission set than the usual empire vs empire conflict.

As far as the gascraft and other missions go, those can be adapted or perhaps modified somehow. Maybe some way to give the gascraft a form of immunity to it?

Amazinite:
I imagine an EM storm should only do shield damage, not hull damage, so the Gascraft should be effected much given that you’re not surviving more than one shot from the Archon anyway.

Zitchas:
Yeah, or I was thinking of a combo of shield damage and ionization.

Jacunn:
I agree that the changes and considerations you all mentioned are
necessary. I more-so created this as a proof of concept for this idea, and
I'd be happier for someone with more experience and/or knowledge to take
this over as long as I'm given a mention for the concept. Ideally I'd like
this to be a system condition rather than a fleet to give space in general
a livelier feeling. Some ideas I have in light of what you have all
mentioned, however:
-As per the gas craft concerns, once this is implemented as a recurring
phenomenon, a flag can be set during this mission to prevent the storms
from occurring.
-As per the obscene levels of damage, I agree that the effects of these
storms must be lowered, otherwise they can just become crippling and
boring, rather than interesting and a new "mechanic".
-For graphical purposes, I do not have the talent to add any visual effects
towards the storm. If anyone would like to do this, that would be amazing.
-For a storm "dissipating" after a while, perhaps the storm has limited
energy capacity in the first place and slowly recharges, meaning that a
storm will have long recharge periods.
What do you all think? If nobody wishes to take this on, I will endeavour
to update this when I have the time do so.

<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
Virus-free.
www.avg.com
<http://www.avg.com/email-signature?utm_medium=email&utm_source=link&utm_campaign=sig-email&utm_content=webmail>
<#DAB4FAD8-2DD7-40BB-A1B8-4E2AA1F9FDF2>

Zitchas:
I'm interested in pursuing this, as it ties in nicely with my Remnant stuff. I think I'd rather keep it separate from my current PR as I think that's pretty much ready to merge, but this is something of a separate idea that will take some fine tuning before it is ready.

That being said, I have been testing some different damage values, and I've got some additional ideas for it, including at least one idea for making it a more "system-wide" type effect, even if I don't have the coding knowledge to make it into an actual game mechanic.

endless-sky:
It seems to me that random damage / disruption with no visible cause would just be annoying. At least with the nanobots there's dialog that mentions what's gnawing away at your ship.

If there's going to be environmental hazards, they should be visible in various ways. For example, an EM storm could cause occasional flashes of lightning in the "background" of the system, or it could cause your radar to flicker in and out.

Zitchas:
That would definitely be a great environmental effect. Part of the reason I'd like to see it attempted in the Waste is that it already has the ominous red colour, so I feel that players wouldn't be as surprised to have something happen. If we could get some kind of general system type damage, it would almost certainly be merged with a short mission along the same lines as the nanobot one. I would be happy to contribute that part if someone else can do the code for having a continuous damage effect.

I agree having lightning flashes or something else in the background would be nice, but between the omnious red background and a dialogue advisory, I don't think it is actually necessary.

Amazinite:
Closing this and moving the discussion on system/environmental hazards as a whole to #4116.

Issue #3733, opened 2018-05-06
Bug: 'Sells This Ship' shipyard colour not displaying


Zitchas:
Version: 0.9.9 (M*C*O's build from Jenkin's, May 5th build)
Not using plug-ins or modifications.

What is expected:
That when clicking on a ship in the shipyard section of the map, that planets with shipyards that sell that ship will change to the orange "Sells This Ship" colour.

What happens:
There is no change to the map. It simply displays which planets have shipyards and which do not.

Other notes:
This appears to be something that happened between the previous build (several months old) and the May 5th build, as in the previous one it appeared to work.

The attached picture shows what the map looks like while having the Scout selected. 

![shipyard sells this ship not working](https://user-images.githubusercontent.com/32169904/39674777-56ea3b6c-510e-11e8-9cca-db94bfb901cc.png)

Wrzlprnft:
https://github.com/endless-sky/endless-sky/commit/7d74467288d2e92df1aa48fe4678d7f19ea9acb6#commitcomment-28869567

MCOfficer:
for reference, may the 5th build was at commit 1f039f0c4db1d5d9a18a82f2d52b37ba539f1a4f.
for the future, please either include the build number (# 17) or the latest commit hash as found in the build title.

Zitchas:
Will do. Thanks!

Just to confirm that I'm doing this right:

This problem has been fixed as of build #19 ( 865b70dcad37dee40bb0e9bc96dcd0bfa0fb3836 ).

PR #3734, opened 2018-05-06
Fix stale map cache when changing ship/outfit selection

CyberShadow:
Fixes the map not being updated when another ship or outfit is selected in the shipyards/outfitters map modes.

Bug introduced in commit 7d74467288d2e92df1aa48fe4678d7f19ea9acb6.

Fixes issue #3733. An alternative way to fix it would be to translate the selection index (as seen in the `MapSalesPanel::Select(int index)` parameter to a fake commodity index, so that the map cache staleness check picks up the change automatically.

### DIFF
--- a/source/MapOutfitterPanel.cpp
+++ b/source/MapOutfitterPanel.cpp
@@ -102,6 +102,7 @@ void MapOutfitterPanel::Select(int index)
 		selected = list[index];
 		selectedInfo.Update(*selected, player);
 	}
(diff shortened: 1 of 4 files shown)


endless-sky:
Sorry, I uploaded 865b70d (which I think is identical to this) before noticing that this PR existed.

CyberShadow:
No problem. The diff indeed looks identical :) Thanks!

PR #3688, opened 2018-04-09
Add assassination missions to Pirate Worlds

czartrak:
It would seem logical that pirates would have some enemies that they would need taken care of. A few _"problems"_ that need _"solving"_. 
I have made 4 assassination jobs (2 political, 2 military), which vary in difficulty, and the target's faction changes based on where the job is offered.

Free Worlds targets in the south
Syndicate targets in the core
Republic targets in the north

Thank you @Amazinite for making the missions much better than they were

PR #3736, opened 2018-05-13
Removed "slave food" from pirate slavery jobs

Amazinite:
Given that illegal cargo is no longer required for a mission to fine the player upon being scanned, I've removed the "slave food" from pirate slavery jobs, since the only point of the slave food was so that the player could be properly punished for carrying slaves.

I'm keeping the extra cargo in the wanted passenger jobs because they have blurbs that pop up upon being scanned that relate to the cargo. (Although, I can see an argument for them being removed in the non-highly wanted passenger missions given the specificity of how the passengers are taken vs the relative commonality of the job, ref: https://github.com/endless-sky/endless-sky/pull/3688#issuecomment-385249464.)

### DIFF
--- a/data/pirate jobs.txt
+++ b/data/pirate jobs.txt
@@ -1378,10 +1378,9 @@ mission "Slave Transport [0]"
 	name "Transfer slaves to <planet>"
 	job
 	repeat
-	description "A pirate on <planet> has bought slaves from a local warlord. Transfer these <bunks> slaves and <cargo> to <destination> for <payment>. Slavery is punishable by death throughout the galaxy, so this may be a dangerous job to take."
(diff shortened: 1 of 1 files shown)


KiLEdEnNis:
not even feeding the slaves? kinda cruel

Amazinite:
Kinda the point. \:)

OverYrPaygrade:
I thought the slaves would be counted as cargo...

KiLEdEnNis:
imagine me posting an "it's time to stop" picture

jafdy:
You don't need to pack in food for your crew so unless a slave eats a heck of a lot more then the rest of us this makes perfect sense.

PR #3748, opened 2018-05-31
Make Hai not hostile to the Unfettered by default

temtemy:
In this PR, the regular Hai will only be hostile to the Unfettered when the latter attacks them. The Hai should only attack for self-defense. If the Hai attack the Unfettered before the latter do, then it is contradicting their claim that they are a peaceful race. And I don't think that the regular Hai would like to kill each other.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -86,7 +86,6 @@ government "Hai"
 	
 	"player reputation" 0
 	"attitude toward"
-		"Hai (Unfettered)" -.5
 		"Merchant" .01
 	"bribe" .02
 	"fine" 0


endless-sky:
It doesn't actually work that way - if either government is hostile toward the other, they'll fight if they encounter each other. The choice of what attitude each government has toward the other just affects how much your reputation changes for disabling or killing a government's allies or enemies. (And in this case, the negative attitude makes it so that if you disable an unfettered ship, you'll be able to land on Hai-home without bribing them.)

temtemy:
Oh, didn't knew about that (maybe it would be possible to code that?). But I don't think the regular Hai will like you for killing their brothers in the north though. It is also still possible to increase your reputation without killing some Unfettered, like repairing Hai ships in Wah Ki and finishing the "Wanderers: Hai Diplomat" mission.

KiLEdEnNis:
if a government would always be attacked by another one, naturally they're not gonna keep waiting to take the first hit
and the Hai don't "like" that you kill the Unfettered, but they don't really "mind" either as they know the Unfettered are the ones who get themselves killed

but in short, this PR isn't really necessary, the intended effects are already met

temtemy:
I don't see any chance for this PR to be merged, so closing.

Issue #2695, opened 2017-07-03
Can turn Syndicate friendly when at war with them


Amazinite:
Currently when doing FW Reconciliation, going to war with the Syndicate does not change their reputation toward the Korath or pirates. This means that during the 30 day break before kidnapping Alexander Soylent, the player has plenty of time to grind Korath or pirates until the Syndicate is friendly. This then breaks continuity when the mission text talks about the Syndicate using nukes on you, which it never did if you turned them friendly.

Suggested that when at war with the Syndicate, their reputation with pirates and Korath be -0.01. I imagine this should be done with the Navy as well if they'll still like you after killing a ton of pirates.

endless-sky:
Fixed in 382be55. Other than the assault on New Tibet, that's the only time you actually fight Syndicate NPCs directly, so I don't think there are any other missions that need to be altered.

PR #3749, opened 2018-05-31
Prevent messing up the FW storyline

temtemy:
This PR will make it very difficult to make the Republic friendly to the player after they have joined the Free Worlds (except when it is ceasefire time and when the war is over). This means killing pirates is not a viable option. Plus, because it is very difficult to make the Republic friendly again, this PR will also make the Republic hostile to the Korath exiles again.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -312,6 +312,7 @@ government "Republic"
 	"player reputation" 2
 	"attitude toward"
 		"Alpha" -.3
+		"Korath" -.3
 		"Merchant" .25
 		"Militia" .1
 		"Pirate" -.3
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -312,7 +312,6 @@ government "Republic"
 	"player reputation" 2
 	"attitude toward"
 		"Alpha" -.3
-		"Korath" -.3
 		"Merchant" .25
 		"Militia" .1
 		"Pirate" -.3
--- a/data/events.txt
+++ b/data/events.txt
@@ -791,7 +791,6 @@ event "start of hostilities"
 	government "Republic"
 		"attitude toward"
 			"Free Worlds" -.2
-			"Korath" -.01
 			"Merchant" .01
 			"Pirate" -.01
 	government "Free Worlds"
@@ -1482,7 +1481,6 @@ event "fw armistice"
 	government Republic
 		"attitude toward"
 			"Free Worlds" 0
-			"Korath" -.3
 			"Merchant" .25
 			"Pirate" -.3
 	government "Free Worlds"
@@ -2395,7 +2393,6 @@ event "fwc peace with the navy"
 	government "Republic"
 		"attitude toward"
 			"Free Worlds" 0
-			"Korath" -.3
 			"Merchant" .25
 			"Pirate" -.3
 	government "Free Worlds"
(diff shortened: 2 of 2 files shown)


KiLEdEnNis:
instead of this all, you could make it that when at war with the Republic, then its attitude towards all factions it has any attitude to could be (-).01 instead, so the reputation cannot go up

temtemy:
Done

temtemy:
Whoops, it should be (-).01, not .1. Fixing that

data/governments.txt
```diff
@@ -312,6 +312,7 @@ government "Republic"
 	"player reputation" 2
 	"attitude toward"
 		"Alpha" -.3
+		"Korath" -.3
```
KiLEdEnNis: any reason this is in, by the way?  if anything, this makes getting a positive rep with the Republic as easy as killing 1 Raider
I'd suggest just removing this change entirely, since it's not really necessary

Zitchas:
Is it really that big a problem for a player to manage to regain a positive relationship with the Republic? I can see making it really hard, but if they kill so many pirates that they are singlehandedly making the Navy obsolete, I think it is reasonable that the Republic might tolerate them at least enough to not shoot at them... (definitely no landing at restricted stations/planets, though...

data/governments.txt
temtemy: I actually made this PR to make it possible to increase your reputation by killing Korath exiles while preventing unintended side effects (like messing up the storyline). But I realized that this is against contributing guidelines (mixing two unrelated commits in one PR), so I am okay if it is needed to separate the two commits to two PRs. Yes, the Republic Navy barely ever meets the Korath exiles, but shouldn't the Republic have high respect for the player for killing some advanced alien pirates raiding human space?

temtemy:
If you live in a Republic planet during the civil war, while there was no ceasefire declared, wouldn't you wonder why the Navy is letting this Free Worlds ship wander in Republic space without attacking them?

tehhowch:
@temtemy barring players who complete Wolf Pact quests, the player has generally not committed any acts that could be construed as hostile to Merchant fleets, and Merchant fleets are what I would consider the de facto "normal fleet", to which 99.9% of humanity belongs. So the real question is, "Do the Free Worlds fleets attack Merchants?" to which the answer is "No."

A secondary question is "Do Merchant ships question the actions of Republic/Navy ships or simply ignore them?" to which the answer is "Ignore,". Really the only interaction they have is when the Merchant vessels are under attack, in the form of a request for assistance.

Given the above generic reasoning, and that the campaigns are largely meant to be played without the player taking an extended break to massacre pirates and/or alien species, I don't see this PR getting merged.

temtemy:
@tehhowch I can't really understand why you just brought up the merchants here, maybe you can make it more clear about that? And if it is decided that "messing up" the storyline is really not big deal, then maybe we can restore the ability to increase your reputation with the Republic by killing Korath exiles (which would be faster than killing pirates because of their world-ships)?

KiLEdEnNis:
I'd say if you wanna get this merged, take out any signs of Korath reputations, since it's irrelevant
you can try making the Korath part another PR, but I don't think it'd get merged

FireWolf74:
Is there a way after a certain time the war with the Republic is ended if your with the Free World.  Also with the Oathkeepers in the mix maybe there will be a way to have a groups mission to kill off Pirates, and other hostiles to the other members.

Amazinite:
I believe a similar change for the Syndicate may be in order, ref: #2695.

PR #3750, opened 2018-05-31
Make Independent hostile to pirates

temtemy:
This PR will make Independent ships (those in the Men system) hostile towards pirates, as a promise to the Free Worlds to denounce piracy in exchange for their independence. There is however a mission where a ship (Oxyrhynchus) whose government is "Independent" is following you. If somehow your reputation with the Independent is negative (because you dominated Men for example), currently this will make the ship hostile to you. It is supposed to just follow you. So this PR also creates an " Independent " for that ship, unaffected of whatever you did to Thule or Smuggler's Den. Another benefit, this also allows the player to increase their reputation with the Independent.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -123,6 +123,8 @@ government "Independent"
 	color .78 .36 .36
 	
 	"player reputation" 10
+	"attitude toward"
+		"Pirate" -.3
 	"bribe" .05
 	"fine" 0
 	"friendly hail" "friendly civilian"
--- a/data/transport missions.txt
+++ b/data/transport missions.txt
@@ -1105,7 +1105,7 @@ mission "Rim Archaeology 3B"
 				accept
 	
 	npc evade
-		government Independent
+		government " Independent "
 		personality waiting
 		ship "Marauder Falcon (Engines)" "Oxyrhynchus"
 	
@@ -1116,7 +1116,8 @@ mission "Rim Archaeology 3B"
 	on complete
 		payment 300000
 
-
+government " Independent "
+	"player reputation" 0
 
 mission "Rim Archaeology 4A"
 	landing


KiLEdEnNis:
I'm not entirely sure, but if I remember, the deal was that they just don't aid pirates anymore
IF this is true, this PR isn't needed, but if I'm wrong, it's fine

temtemy:
From the mission "FW Pirates 4B": "...they will try to put an end to piracy in this system." To really end piracy, the pirates should be driven off. You can't do that without killing the pirates (of course the pirates wouldn't accept that they lost one of their worlds). By killing the pirates, you are helping them guarantee their independence. Hence the reputation increase when you kill pirates.

KiLEdEnNis:
putting an end to piracy could also mean banning them from landing and arresting such people on the planets themselves, they wouldn't need to fight pirates in space to keep up the deal this way

MessyMix:
I personally think that it's really cool how the Independent get along with
both Pirates and the FW. It was an interesting little gameplay aspect that
made the universe seem more "alive". I think it's better kept as is.

jlnxr:
I agree with most of the above (that it's unneeded), except for the Oxyrhynchus part. I once killed Oxyrhynchus without realizing it would make the Men system hostile... Seems odd considering their "independent" nature would probably make them not care about the ship of some artifact thief dozens of systems away. I think putting Oxyrhynchus in a different " Independent " is an excellent idea.

KiLEdEnNis:
doing that seems fine, but making a new government for such one-time things shouldn't be taken lightly

temtemy:
How about making the Oxyrhynchus a "Merchant"? Better rather than risking one system being permanently hostile to you (it is easy to increase your reputation with merchants). Not all merchants don't do illegal things, so that should be okay. Look, why do some merchants risk their lives going to pirate systems, if there are safer and more profitable trade routes available? Who knows, they might be buying slaves.

temtemy:
And it seems there is very little chance to approve the Independent to be hostile to pirates, and the Oxyrhynchus is imo better to be "Merchant" rather than " Independent", so closing.

jlnxr:
Making merchants even temporarily hostile seems far worse to me than making a single system hostile. 

If @endless-sky is against creating new governments for single missions (despite doing that for the Archon in the ember waste) I guess that status quo makes sense, and the original purpose of the PR (make independent hostile to pirates) is unlikely to be merged, but I don't think closing this (or making Oxyrhynchus a merchant) resolves that problem at all, even if it is a minor one in the grand scheme of things.

PR #3753, opened 2018-06-01
Republic now hostile to Korath exiles

temtemy:
Separated from #3749. Same as the unrelated governments.txt commit on that PR: this will make the Republic hostile again to the Korath exiles.

(and ignore the may commits)

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -312,6 +312,7 @@ government "Republic"
 	"player reputation" 2
 	"attitude toward"
 		"Alpha" -.3
+		"Korath" -.3
 		"Merchant" .25
 		"Militia" .1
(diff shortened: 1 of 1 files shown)


Amazinite:
For what reason? Navy ships never naturally encounter Raiders, so why would the Republic care about you destroying the Korath?

KiLEdEnNis:
making it -.01 would just make them attack Exiles if they somehow encounter them, but not care about you killing them
it's that or don't change anything imo

endless-sky:
I don't see any point to this, since they never encounter each other. And, it doesn't make sense for killing Korath to increase your Republic reputation, because most Republic worlds don't know the Korath exist.

PR #3766, opened 2018-06-09
Fix typos

jostephd:
I've been fixing these as I was playing, but forgot to send them in. :flushed:

### DIFF
--- a/data/engines.txt
+++ b/data/engines.txt
@@ -223,7 +223,7 @@ outfit "X5200 Ion Steering"
 	"turn" 2174
 	"turning energy" 3.5
 	"turning heat" 7.3
-	description "The X5200 is a steering system so large and powerful that only a handful of ships can install it. This means that this steering system is rarely sold due to its limited market."
+	description "The X5200 is a steering system so large and powerful that only a handful of ships can install it. It is rarely sold due to its limited market."
 	description "	Ion engines consume less energy than plasma engines and produce less heat, but they also take up more space."
 
 
--- a/data/engines.txt
+++ b/data/engines.txt
@@ -223,7 +223,7 @@ outfit "X5200 Ion Steering"
 	"turn" 2174
 	"turning energy" 3.5
 	"turning heat" 7.3
-	description "The X5200 is a steering system so large and powerful that only a handful of ships can install it. It is rarely sold due to its limited market."
+	description "The X5200 is a steering system so large and powerful that only a handful of ships can install it. Due to its limited market, it is produced in small numbers."
 	description "	Ion engines consume less energy than plasma engines and produce less heat, but they also take up more space."
 
 
--- a/data/free worlds intro.txt
+++ b/data/free worlds intro.txt
@@ -513,7 +513,7 @@ mission "FW Katya 1"
 				`	"Okay, tell me more."`
 					goto `more`
 	
-			`	"Look kid," she says, "when I ask you to pick sides, when I offer you a mission that's gonna burn some bridges for you, I won't be sneaky about it. I'll tell you exactly what you're signing up for. What I'm offering you now is just a chance to get to know us, and let us get to know you, with no expectations. Heck, you can even run missions for the Reps at the same time."`
+			`	"Look, kid," she says, "when I ask you to pick sides, when I offer you a mission that's gonna burn some bridges for you, I won't be sneaky about it. I'll tell you exactly what you're signing up for. What I'm offering you now is just a chance to get to know us, and let us get to know you, with no expectations. Heck, you can even run missions for the Reps at the same time."`
 			choice
 				`	"That sounds fair. I'll hear you out."`
 				`	"Fair enough, but I don't like working with people who start conversations by pointing guns at my head. Even unloaded guns."`
--- a/data/map.txt
+++ b/data/map.txt
@@ -25390,7 +25390,7 @@ planet "Into White"
 planet "Karek Fornati"
 	attributes korath efret
 	landscape land/desert1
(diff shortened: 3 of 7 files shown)


data/engines.txt
```diff
@@ -223,7 +223,7 @@ outfit "X5200 Ion Steering"
 	"turn" 2174
 	"turning energy" 3.5
 	"turning heat" 7.3
-	description "The X5200 is a steering system so large and powerful that only a handful of ships can install it. This means that this steering system is rarely sold due to its limited market."
+	description "The X5200 is a steering system so large and powerful that only a handful of ships can install it. It is rarely sold due to its limited market."
```
tehhowch: `it. It is` is all really similar.

data/free worlds intro.txt
```diff
@@ -513,7 +513,7 @@ mission "FW Katya 1"
 				`	"Okay, tell me more."`
 					goto `more`
 	
-			`	"Look kid," she says, "when I ask you to pick sides, when I offer you a mission that's gonna burn some bridges for you, I won't be sneaky about it. I'll tell you exactly what you're signing up for. What I'm offering you now is just a chance to get to know us, and let us get to know you, with no expectations. Heck, you can even run missions for the Reps at the same time."`
+			`	"Look, kid," she says, "when I ask you to pick sides, when I offer you a mission that's gonna burn some bridges for you, I won't be sneaky about it. I'll tell you exactly what you're signing up for. What I'm offering you now is just a chance to get to know us, and let us get to know you, with no expectations. Heck, you can even run missions for the Reps at the same time."`
```
tehhowch: The original is easily allowable - this is a character's speech, and its punctuation requirements stem from the character and the context, not the formal language rules.

data/engines.txt
KiLEdEnNis: "The X5200 is a steering system so large and powerful that only a handful of ships can install it. Because of its size, it is rarely sold due to its limited market."

just adds 4 words in-between, but it does make the flow better

jostephd: How about "Due to its limited market, it is produced in small numbers"? Or delete the second sentence entirely.

data/free worlds intro.txt
jostephd: I get that characters don't always speak grammatically, but even so my two cents are that it's better with the comma.

data/map.txt
```diff
@@ -25390,7 +25390,7 @@ planet "Into White"
 planet "Karek Fornati"
 	attributes korath efret
 	landscape land/desert1
-	description `Karek Fornati was once a farming world, before global temperatures began to rise and the overworked farmland slowly deteriorated into deserts. The Korath who remain have dedicated their lives to finding a way to keep the "sand seas" from swallowing up the last of the fertile land, but without much success, and there are regions near the equator where sandstorms rage continually and no living things can survive.`
+	description `Karek Fornati was once a farming world, before global temperatures began to rise and the overworked farmland slowly deteriorated into deserts. The Korath who remain have dedicated their lives to finding a way to keep the "sand seas" from swallowing up the last of the fertile land, but without much success, and there are regions near the equator where sandstorms rage continually and no living thing can survive.`
```
endless-sky: I'd argue that either of these is grammatical, and they have slightly different implications. Saying "no living thing can survive" sounds more like it's saying that any living thing would be instantly annihilated, whereas, "No living things can survive" comes closer to my intended meaning: that the environment there is too hostile to sustain an ecosystem.

data/free worlds intro.txt
temtemy: I think too it is better with the comma, it really shows Katya thinking of what to say to the player.

data/map.txt
jostephd: The intended meaning sounds good to me.

I read "no living thing can survive" not as instant annihiliation but as "If you put a living thing in this environment, it would eventually die". That death would not necessarily occur in the blink of an eye, though; it might take some seconds or even days.

How about avoiding the word "thing", as in "...and neither animal nor plant can survive"? Or a slightly different take, "and bacterial colonies in sheltered caves are the only life form to be found"?

jostephd:
@KiLEdEnNis The change from "things" to "plants and animals" was intentional to add flavor. See previous discussion on https://github.com/endless-sky/endless-sky/pull/3766#pullrequestreview-127381767

KiLEdEnNis:
but it's not necessary, and MZ himself said both thing and things are correct, and that what we have now is intended

data/map.txt
```diff
@@ -25390,7 +25390,7 @@ planet "Into White"
 planet "Karek Fornati"
 	attributes korath efret
 	landscape land/desert1
-	description `Karek Fornati was once a farming world, before global temperatures began to rise and the overworked farmland slowly deteriorated into deserts. The Korath who remain have dedicated their lives to finding a way to keep the "sand seas" from swallowing up the last of the fertile land, but without much success, and there are regions near the equator where sandstorms rage continually and no living things can survive.`
+	description `Karek Fornati was once a farming world, before global temperatures began to rise and the overworked farmland slowly deteriorated into deserts. The Korath who remain have dedicated their lives to finding a way to keep the "sand seas" from swallowing up the last of the fertile land, but without much success, and there are regions near the equator where sandstorms rage continually and neither animal nor plant can survive.`
```
Amazinite: This just seems unnecessarily verbose, now. Reading this, I would simply think "why not shorten that as 'no living thing(s)'?" This description is probably just better off being left alone.

jostephd: Done. If MZ likes the "neither animals nor plants" language we can always put it back.

PR #3768, opened 2018-06-09
Show government name of a planet when different from system's government

jostephd:
When you select Tarazed in the map, the list of Planets will say "Echo (Quarg)" rather than "Echo".

### DIFF
--- a/source/MapDetailPanel.cpp
+++ b/source/MapDetailPanel.cpp
@@ -450,7 +450,10 @@ void MapDetailPanel::DrawInfo()
 				SpriteShader::Draw(planetSprite, uiPoint);
 				planetY[planet] = uiPoint.Y() - 60;
 			
-				font.Draw(object.Name(),
+				string name = object.Name();
+				if(planet->GetGovernment() != selectedSystem->GetGovernment())
+					name += " (" + planet->GetGovernment()->GetName() + ")";
+				font.Draw(name,
 					uiPoint + Point(-70., -52.),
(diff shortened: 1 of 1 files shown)


Zitchas:
That's a good suggestion. I like it. I don't know how often it will come up, but clarity is always good.

endless-sky:
I'm worried that the potential confusion of players thinking that "Echo (Quarg)" *is* the planet's name would more than balance out the usefulness of this.

jafdy:
I feel that if you know of the existence of the the faction in question then it will be quite obvious what is going on. On the other hand I'm not entirely sure how this actually help so I really don't see much difference between the two option here.

tehhowch:
Perhaps rework it to use tooltips? E.g. if hovering the planet name

jostephd:
Or add a noun inside the parentheses, for example: `Echo (Quarg outpost)`.

Pointedstick:
+1 for `Echo (Quarg outpost)` or `Echo (Owned by the Quarg)` or something like that.

jostephd:
Changed.

Amazinite:
I’d say we just not make this change because 1. MZ already marked the original change as unlikely and 2. This new change doesn’t fit the current UI.

jostephd:
Instead of making the line longer, the info could be placed on a new line and the box made accordingly taller, but I'm not going to work on that. If someone wants to make that happen, be my guest, otherwise let's close this.

Pointedstick:
Closing it, then.

Better luck next time!

PR #3775, opened 2018-06-14
Change version in credits.txt

temtemy:
This PR makes it very clear to nightly users that they are using the current nightly (0.9.9), not the current stable version 0.9.8. When I installed my built nightly and saw the version 0.9.8 in the credits, I thought there was a problem in installation and I was still in 0.9.8, so I have to check the map to be sure (and it is really nightly).

### DIFF
--- a/credits.txt
+++ b/credits.txt
@@ -1,5 +1,5 @@
 Welcome to Endless Sky!
-version 0.9.9
+version 0.9.9-nightly
 
 Programming
   Michael Zahniser
--- a/credits.txt
+++ b/credits.txt
@@ -1,5 +1,5 @@
 Welcome to Endless Sky!
-version 0.9.8
+version 0.9.9
 
 Programming
   Michael Zahniser


Zitchas:
This would definitely make things a bit clearer. 

Is there any way to specify that it is a nightly instead of a stable build?

temtemy:
@Zitchas Maybe "0.9.9-nightly" instead of "0.9.9"?

Wrzlprnft:
given that in the past, a version change always indicated that a new release would follow shortly after, i'd go with nightly included

tehhowch:
I don't see this as a meaningful change to the repo or the way things are done - effectively as soon as 1 commit is merged after the publication of v0.9.9, then the credits would need to be changed to `v0.9.10-nightly`. And then we're just making changes for changes sake, which wastes everyone's time.  
If you're making builds nightly, you know you're not on version 0.9.8. If you get builds from someone else, who builds nightly, perhaps they should include the latest commit hash as the version. It's really easy to write a shell script to make that an automated step in the build.

Wrzlprnft:
i actually preder tehh's solution even more.
could even be shared in https://github.com/endless-sky/endless-sky-tools?

Amazinite:
This change seems unnecessary to me. MZ only ever changes the version number once a new version is actually reached. If you're for some reason having trouble differentiating between a built version and a non-built version, then you should change the credits for your own version. Otherwise, I don't see why this should be changed for everyone, especially because the @MCOfficer nightly build on Steam that you appear to be referring to isn't an official build from MZ, it's just something that MCO made for the community given the unusually long time since the last update.

temtemy:
I agree with tehhowch's solution, and his solution is better than mine, so closing.

MCOfficer:
@tehhowch that's probably a good idea, yes.

PR #3779, opened 2018-06-15
Give atmosphere scan capabilities to the gascraft hull

Tadrix:
for possible use in exploration fleets once the Remnant begins to use gascraft again. They should be able to scan what they are about to plunge into.

### DIFF
--- a/data/remnant ships.txt
+++ b/data/remnant ships.txt
@@ -306,6 +306,7 @@ ship "Gascraft"
 		"shield energy" 0.4
 		"hull repair rate" 1.2
 		"hull energy" 0.9
+		"atmosphere scan" 200
 		"gaslining" 1
 	outfits
 		"Millennium Cell"


tehhowch:
Any mod which uses the Gascraft can easily add this with `add attributes` node, and if the base game direction results in the Remnant reviving the mothballed Gascraft, we can add it then.

Zitchas:
That makes a lot of sense. Also could be very useful in future mission arcs. I've got some ideas for putting the gascrafts to use, and this would add some options to that.

I didn't even know atmospheric scan was a thing. I was just going to do normal conversations to give messages about doing atmospheric research. I'll have to go add this in and see what sort of results and messages it gives.

jafdy:
I think atmospheric scan just tells the AI to fly between planets.

KiLEdEnNis:
yes, it's what makes Surveillance Drones fly around to scan instead of landing right away

endless-sky:
It doesn't make sense to add this when there are currently no NPC-controlled gascraft in the game. (And, given that the story says you're buying one from a limited supply of mothballed ships, not new ships that the Remnant has begun producing, there's no reason to expect NPCs to start flying them.)

Zitchas:
I've got missions and uses in mind for gascrafts with atmospheric scan, (specifically, triggers that start the Remnant using gascraft again) but there's no reason I can't add it in as a retrofit when I get to that point.

PR #3785, opened 2018-06-17
Add support for Eclipse to .gitignore

OrdoFlammae:
This commit makes git ignore the project files automatically created by
Eclipse.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@
 /obj/Release/
 /obj/Win32/
 /bin/
-
+.cproject
+.project


tehhowch:
Doing so would imply support for building with Eclipse. There is not support from this repo for building with Eclipse. The IDEs that are supported are [Code::Blocks and XCode](https://github.com/endless-sky/endless-sky/pull/2896#issuecomment-336547064). That other IDEs and toolchains (such as VS Code, Visual Studio, Eclipse, netbeans, etc) can be used to build the source code is outside the scope of the repo.

Issue #3780, opened 2018-06-16
"dogging"


temtemy:
From the mission "FW Rand 1":
``When you land on Dancer, Freya has just received some exciting news. "It turns out taking out Bartlett had some unexpected side effects," she says. "Bartlett has been dogging the systems on the northwestern fringe of the Dirt Belt for years, and had a lot of enemies.``

I just noticed that word right now. I thought "dogging" is a misspelling of "dodging", but after researching about "dogging", I understood now what that word means. I think it would be better if that word be changed into "following closely" or similar, since not all players are native English speakers (and I am one of those players). And probably to keep this game "family-friendly" (a minor might be surprised when they saw what that word means, and I don't think you would want a British parent would ban their child from playing this game because of that word).

Amazinite:
The best replacement would probably be harassing. 
And as for those wondering about the "family-friendly" part, apparently dogging is British slang for having sex in public. Exhibitionism, one could say.

MCOfficer:
+1 because i really had no idea what it meant.

jafdy:
If it is British slang then it is a rarer dialect. That is not one I have ever come across, my understanding was that it meant something along the lines of harassing.

jostephd:
https://en.wiktionary.org/wiki/dog#Verb gives both the author-intended meaning (2) and the non-family-friendly meaning (4). https://www.thefreedictionary.com/dog gives meanings such as "haunt", "harry", and "plauge".

It seems the family-friendly/intended sense is the established one.

Nescio0:
There is a difference between the *verb* to dog, dogging, dogged (follow (someone) closely and persistently) and the mass *noun* dogging (the practice of watching or engaging in exhibitionist sexual activity in a public place.) Both from the context and the grammar it's clear the former is used here, not the latter, so there is no need to change it.

MessyMix:
I too think there isn't really that much of a need to change it, although there's really no harm done in "harassing".

Lineth:
Eh, I prefer rarer/stranger word choices if it's dialogue. Characters shouldn't always speak like a game tutorial. The meaning in not hard at all to guess from context in this scenario anyway.

Vilhelm16:
I don’t have an opinion on whether it should be changed or not but if you
want to change it some of your options include:

-tormenting
-assailing
-ravaging
-beleaguering
-provoking
-raiding
-hassling

Take your pick. We could use lots of different words here.

PR #3787, opened 2018-06-19
Changes "dogging" to "hounding" (refs #3780)

MessyMix:
I personally did not see the absolute need to change this, but it's ultimately *safer* to implement it the way @Amazinite suggested (dogging --> harassing).

Because, after all, when someone not familiar with the word or it's usage goes to look it up, this is what they'll find:

image

And that might not be the best or most accurate thing for players to be seeing. :P

### DIFF
--- a/data/free worlds middle.txt
+++ b/data/free worlds middle.txt
@@ -1720,7 +1720,7 @@ mission "FW Rand 1"
 		log "Another two planets, Rand and Oblivion, have expressed an interest in joining the Free Worlds. They are rather far from the rest of Free Worlds territory, however, so JJ is worried that it would create an indefensible tactical position."
 		event "battle for zeta aquilae"
 		conversation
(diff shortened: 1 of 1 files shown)


endless-sky:
I feel like "harassing" has a completely different tone (formal / legal). "Hounding" would be a much better replacement.

Or, "bird-dogging"; that's the term used in [Chapter 6, verse 8](https://www.youtube.com/watch?v=JhkGuh_EFkM) of the Book of Whedon, which is one of the canonical scriptures of the Endless Sky universe:

> "Now think real hard. You been bird-dogging this township a while now. They wouldn't mind a corpse of you. Now you can luxuriate in a nice jail cell, but if your hand touches metal, I swear by my pretty floral bonnet, I will end you."

MessyMix:
Haha—haven't come across "bird-dogging" before!
What do others think? Personally, I'm a fan of "hounding", but I want to see what you guys have to say between `hounding` and `bird-dogging` before changing things.
React: 😄 for hounding, and 🎉 for bird-dogging!

gunqqer:
I say houding, people are more likely to know what that means. While I can figure out "bird-dogging" due to context here if I saw it in game it'd just confuse me.

Zitchas:
I think hounding is probably going to give a better flow to it, but bird-dogging gives it an interesting archaic twang.

MessyMix:
Alright, hounding it is. If it were the Quarg or an alien species, I probably would have gone with the other option, but this is Freya after all :P.

Issue #2336, opened 2017-03-29
[Feature] Fighters auto-dock to replenish ammo/shields/power, then relaunch


Tadrix:
What is says on the tin. In Endless Sky it would be very tedious to micromanage carrier fleet, thus I propose this:

If fighter has no more ammo (it can't shoot at all), it docks. If its carrier is carrying more ammo (in cargo), it takes some time to reload (recharging everything else during this time) and then relaunch; otherwise, remain docked.
If fighter is running on batteries, on low power it docks to recharge (recharging shields and reloading ammo, if possible, in the process), then relaunches.
If fighter has <25% shields left, it docks to recharge them (charging batteries and reloading ammo, if possible, in the mean time), then relaunches.
Carrier could also make repairs to fighter hulls?

This would make carriers much more useful, and also there would be more sense to use battery-powered ships.
Some sort of kamikaze order to fighters would be also nice, but it would probably warrant separate issue.

Wrzlprnft:
i think this came up before, but stayed unresolved.
Though, that would probably require fighters memorizing to which carrier they belong, and i recall that not being the case.

zwparchman:
One of the features of my #1971 pull requests hits most of these features. Hulls repair if the carrier has a hull regeneration module in the base game already. Battery management is not included in my PR, but it would be trivial to add.

Tadrix:
@Wrzlprnft If it's not the case, they could just retreat to a nearest carrier with free bays.

jafdy:
@Wrzlprnft it is the case so long as the carrier stays in-system. If you leave the system and come back you get the fighter mass moving from one carrier to the other for a while until they are all docked.

Amazinite:
Is the only request from this issue that isn't done the restocking ammo part?

warp-core:
Did that PR actually cover docking for power?

tehhowch:
> Did that PR actually cover docking for power?

No, docking to replenish power was not an intended use case to be satisfied by that PR.

warp-core:
This issue seems to be requesting docking for power as well. I think there is a PR that does that. So, unless we just don't want that functionality, I think this issue should be reopened?

EjoThims:
Reopened to link to https://github.com/endless-sky/endless-sky/pull/6726 covering the last requested piece.

samrocketman:
I updated `Related issues` in 6726 to note that it will close this issue when it is merged.

Issue #1939, opened 2016-12-29
Fighters and Missiles


gchapyzh:
_Over the past few months I have been compiling this assembly of thoughts, things I think need to be changed or added. Feel free to address these ideas individually. Or insult the whole list in general._

**1.1 [Missile Turrets](http://preview.turbosquid.com/Preview/2014/05/21__07_27_49/MissileTurret00b.jpg7db60eba-e707-4d3b-9c3d-c208933fdf7fOriginal.jpg):** This has been said previously, but as I can't quite understand why they are not in the game quite yet. As a Rocket or a Javelin is a direct firing weapon, it's only logical for them to have a rotating version for capital ships. A Quad-Javelin would go well with the Plasma Turret logic, where ships have a smaller number of larger sized weapons. A Quad-Javelin would also be fatal to any slow ship with insufficient PD.

**1.2 Torpedo Turret:** On a separate note, why not make a Torpedo turret - a Javelin pod on steroids, slower firing, with more survivability and damage. If you're not paying attention, this is something that will kill your big, slow ship, unless you have the PD to counter it.

**1.3 FF Missiles:** Fire and Forget missiles, for those of you who don't know. Something desperately needed in fleet battles I think. In a target rich environment, I would like to have ships that can plow through the battle at great speed, spewing FFs at all hostiles in the system. This way I can still dash out DPS without having to spam R. An upgraded version of the FF could be smart enough to optimize targets. Since your won't need a lock to hit a target, you could be dueling a ship with your turrets, while firing off FFs against other smaller ships or fighters, or you could simply spew munitions at a swarm of hostiles chasing you, without the need to turn and get distracted from maneuvering. 

_Missiles are somewhat of a human weapon in my mind, with most other races focusing on beams or projectile weapons. Hai and Wanderers may use missiles, but they have one type, and its radically different from anything humans have. As a clearly human weapon, I think it should be further expanded on. When a quarg fire on you with beams, humans return fire with overwhelming waves of Javelins, Torpedos, Missiles and Rockets._

**2.1 Fighters Suck:** They are far too easy to kill, and I think that's the number one issue with them. This is how I see this issue being fixed, and this is a mix of everything that has previously been said -

[...]

PR #3791, opened 2018-06-20
Improve carried ship docking logic

tehhowch:
This supersedes #1971, which we let languish, especially in the face of major upgrades to the game from 24cab842a9f3a66c1b5fb862e1139707c9dd2ee4
 - functionality for instructing fighters to return to their parent ship if they are damaged enough 
 - a combined shield & hull based health computation

My original addendum to #1971 featured some pathing logic before the fighter would plot a course back to its carrier (to avoid possible silliness such as being injured and running through a pack of enemies to reach the carrier on the other side, or drawing additional hostiles to the carrier). While this PR currently does not add that check, it may be worthwhile, pending feedback we collect during the next beta release.

Implementation details:
 - I considered treating the `Preferences::Has("Damaged fighters retreat")` setting as we do the `isLaunching` variable, where it is a private member of the `AI` class. Doing so would avoid needing to query it multiple times over the course of `AI::Step`, and we could possibly allow in-flight toggling of that setting, much like we do for deploy/recall and ammo frugality. However, this functionality change seems more appropriate for a separate pull.
 - The selection of `< 5` free cargo for a cargo-induced docking desire is due to the criteria of `>= 5` cargo space free in order to mine.
 - It is possible that adding a check for if this ship has previously decided to board would be worthwhile (such a check was present in [1971](https://github.com/endless-sky/endless-sky/pull/1971/files#diff-80825e2b1f90be6b565f227d9ddb18ccR1239). Since the checks here are less extensive than those there, I have omitted this check. Adding this check would mean we would not need to apply hysteresis when comparing the ship's health to the retreat health constant - once it decided to board for repairs, it would continue unless some other factor intervened.

As this PR contains commits from multiple authors, it should not be squash-merged in order to maintain proper attributions. (Unless GitHub has changed how they treat squash-merged commit authors since July 2017.)

I welcome comments and playtesting notes from anyone who can assist. There is an example mission which spawns NPCs that utilize ammo and fuel-based fighters [here](https://github.com/endless-sky/endless-sky/pull/1971#issuecomment-318568125).


source/Ship.cpp
```diff
@@ -1575,6 +1575,36 @@ void Ship::Launch(list<shared_ptr<Ship>> &ships)
...
+			for(const Outfit *outfit : toRefill)
+			{
+				int neededAmmo = bay.ship->attributes.CanAdd(*outfit, OutfitCount(outfit));
+				if(neededAmmo)
+				{
+					AddOutfit(outfit, -neededAmmo);
+					bay.ship->AddOutfit(outfit, neededAmmo);
+					carriedMass += outfit->Mass() * neededAmmo;
+				}
+			}
```
endless-sky: Does this mean that a fighter will "take" ammo that's installed in the carrier? I'm not sure that that's a desirable behavior.

tehhowch: Indeed it does. The other option is to pull from cargo, but then you cannot make sure the carrier has the ammo (vs some other ship in the fleet). Pulling from cargo also means you can resupply much more easily from flotsam.

tehhowch: We could augment the `Ship::Carry` routine to have the carried ship unload its ammo stores back into the carrier, as this PR does with fuel. That would maximize availability of ammo to the carrier (while fighters repair or are kept in bays), while also promoting maximal restocking of any launching fighters.

Once this restocking behavior is known to content creators, it can be pretty easy to build AI fleets that use it appropriately - stocking the AI ship with the ammo boxes, and not the launcher, is sufficient.

I admit, having a carried ship restock from the carrier could be initially confusing to players, but after they observe it happening (in either their fleet or in AI fleets), I think it will make more sense than not. This behavior has been requested before, on both Steam and here on GitHub.

https://github.com/endless-sky/endless-sky/issues/1939#issuecomment-269992593
https://github.com/endless-sky/endless-sky/issues/2336
https://steamcommunity.com/app/404410/discussions/2/492378265890026032/
https://steamcommunity.com/app/404410/discussions/0/1319962683445095725/#c1327844097131431281
(I'm sure there are other references as well)

Janaszar: In current warfare, carriers tend to be focused on sending out fighting units, and resupplying them, usually meaning they have a token defensive armament built into the ship, which tends to be something like CIWS/AMT.  The Carriers in the game are right now, more like battleships, especially given the rather pathetic strength of fighters (in my opinion and playstyle).  Having them at least be able to reload their fighters/drones seems like a commonsense addition to the game; perhaps allowing fighters/drones to be re-balanced as more durable, while removing weaponry from fighter/drone carrying ships.  
This also lends itself to the idea that the player could either focus on the fighters/drones dealing them damage, or focus on the carrier itself, to prevent them from rearming.

Tadrix: Personally I'd prefer fighters pull ammo from cargo, and, when assigning cargo, ships favour having ammo on a carrier that has fighters that can use it. It isn't logical having to pull missiles out of the ship's launch tube to resupply fighters, and if the player can use floatsam (assuming it even fits the fighters), then going to fish for them during the battle could be a fun playing experience.

Pointedstick: > Does this mean that a fighter will "take" ammo that's installed in the carrier? I'm not sure that that's a desirable behavior.

Conceptually, it's quite desirable. The entire purpose of carrier vessels is to arm, launch, and recover their carried fighters, generating what's known as a "sortie". Maximizing sortie rate by re-arming and re-fueling the fighters as quickly as possible is a key element of military aviation, particularly naval aviation. To be more realistic, Endless Sky carriers should be absolutely stuffed to the gills with ammo storage outfits (or ammunition in their cargo holds, depending on how the reload mechanic is implemented).

Tadrix: A thought: Divorce carrier own ammo supply from what is intended for the fighters. Storage facilities make it possible to launch the ammo; this is not the case for the ammo that is intended for fighters. We also have to ensure that the ammo is in that ship.

The solution: Introduce outfits that convert carrier's *cargo space* without losses for bulk storage of said ammo. It can only replenish fighters, not be launched from the carrier.

How to handle the situation if both carrier and its fighters use the same ammo? When landed, always assume carrier's own weapons are loaded first, and the rest goes to the fighter ammo storage.

Pointedstick: Hmm, having separate "fighter ammo storage" outfits seems like it would add a lot of complication for not a lot of real gain.

tehhowch: Given that this PR resolves #4112, I think I will pull the ammo-refilling changes from this PR and open those alone in a new PR.

tehhowch:
I've removed ammo-based return logic and resupplying, pending a decision on whether we should perform cargo-based restocking or installed-outfit-based restocking. This PR now primarily codifies where to insert new docking logic, and does so for parent fitness and fuel consumption (e.g. flamethrower fighters)

PR #3792, opened 2018-06-20
Add 3 hostile Heliarch hails

temtemy:
This PR will add 3 new hostile hails for the Heliarch. I hope the hails are true "Heliarch English" :) Suggestions are welcome of course.


data/coalition.txt
```diff
@@ -735,3 +735,98 @@ phrase "friendly heliarch"
...
+		"Betraying "
+		"Dishonoring "
+		"Ruining "
+		"Invading "
+		"Attacking "
+		"Conquering "
+		"Oppressing "
+		"Backstabbing "
+	word
+		"the Coalition "
```
tehhowch:
```
word
    " "
word
    "the"
    "our"
word
    " "
```

jafdy: No, `...out citizens of the coalition...` `...the just government of merit...` don't make complete sense. An argument can be made for both but they are a little odd. Having "the" instead of "our" on the last four would sound like they think that any other government is unjust (or something).

tehhowch: That's just an argument to keep going and diversify the hail though - it's not really an argument against this. You can make a set of phrases that work with with `the` and a set that work with `our`, and a set that works with both, etc.

jafdy: That would require 3 entire sets of hails that are almost the same apart from a couple of lines in the middle.

tehhowch: I mean to use a sub-`phase`, e.g. `phrase "heliarch possessive unions"`, which is all phrases that pair well with `our` as a lead-in, and then you call that within this actual hostile hail. At minimum we both can agree that the last bit's duplication is easily solved by moving "will" and "shall" to a `word` token:

```
  "will live on"
  "shall live on"
  "will not be destroyed"
  "shall not be destroyed"
```
should absolutely be
```
word
  "will"
  "shall"
word
  " "
word
  "live on"
  "not be destroyed"

Comment (temtemy) [reply data/coalition.txt]: Should be okay now

File: data/coalition.txt [hunk 2]
```diff
@@ -735,3 +735,151 @@ phrase "friendly heliarch"
...
+		"peacekeeping"
+		"powerful"
+		"everlasting"
+		"legendary"
+	word
+		" "
+	word
+		"military."
+		"law-enforcers."
+		"navy."
```
KiLEdEnNis: split the period to another word

data/coalition.txt
```diff
@@ -735,3 +735,151 @@ phrase "friendly heliarch"
...
+		"will"
+		"shall"
+	word
+		" "
+	word
+		"live on!"
+		"not be destroyed!"
+		"not be history!"
+		"not die!"
+		"not be extinct!"
```
KiLEdEnNis: split the exclamation mark to another word

data/coalition.txt
```diff
@@ -735,3 +735,151 @@ phrase "friendly heliarch"
 		"we have committed."
 		"we do serve."
 		"we do vow."
```
KiLEdEnNis: this isn't really part of your PR, but could you also split the period here to another word?

temtemy: I thought it is not needed to separate periods to another word because of the current hails. I think we need a standard for writing hails, for the sake of consistency.

KiLEdEnNis: the standard is basically splitting the interpunction at the end of a hail and spaces between the "words" in a hail
while your PR just adds some hails, you could standardize the whole file

temtemy:
Although commit 4fa2aafcb31d641ca73fb2682bc6e07a7408d15c seems to contain "unrelated" changes, I won't separate those changes to another PR to avoid merge conflict. And for consistency as well.

data/coalition.txt
temtemy: I am wondering, is the standard de facto, or said by MZ himself?

KiLEdEnNis: not sure if it was ever clearly said, but MZ and others have done this before, so it's at least implied to some extend
although while creating whole new files, one may forget some standards at times

PR #3794, opened 2018-06-22
"foosball" to "football"

temtemy:
Spelling fix in line 913 of `hails.txt`

(The Quarg play football in Luna? And the Quarg is controlled by the Hai? wtf!)


MessyMix:
Sorry, close issue please:

Foosball is not a typo.

image

temtemy:
Ah makes sense

Issue #3686, opened 2018-04-08
Grab-strike overcompensates for player velocity


SolraBizna:
I'm flying a souped-up Sojourner with Planetary-class engines and mainly Grab-Strikes for armament. If I am stationary, the Grab-Strike does a reasonable job of aiming at any ship I'm fighting, fast or slow. However, the faster I am moving, the worse its aim is. Unless the ship is directly behind me on a parallel course, the Grab-Strike effectively never hits its target, instead passing well in front. It seems as if the faster I am moving, the more the turret aiming code underestimates the Grab-Strikes' actual velocity. (Grab-Strikes do move "faster" when fired rearward than they do forward, because they have acceleration, drag, and homing.)

kyzentun:
Here's my analysis as someone just glancing through the code from the outside.

In AI.cpp, function AI::AimTurrets, there's the loop for finding the enemy each turret can aim at.  In there is this code:
```c++
				// Only take the ship's velocity into account if this weapon
				// does not have its own acceleration.
				if(!weapon->Acceleration())
					v -= ship.Velocity();
```
The grab-strike has acceleration and drag, so ship velocity is ignored.  The Projectile constructor in Projectile.cpp uses the ship velocity.  Thus, the aiming calculations are wrong about the velocity of the projectile.  Since the grab strike only has a turn of .2, lower than any missile, it can't compensate the way missiles do.

If the ship velocity were used in the calculations, they would be wrong in a different way.  AI::RendezvousTime calculates when the projectile would reach the target.  But it works with a constant velocity, it won't calculate the right time for a weapon that has drag and acceleration.

So turreted accelerating weapons like the grab strike need a separate function for calculating time to impact that factors in their drag and acceleration.

NCGThompson:
I am attempting to fix this using the method proposed by @kyzentun :

> So turreted accelerating weapons like the grab strike need a separate function for calculating time to impact that factors in their drag and acceleration.

 Where in game can a Grab-Strike or something else good for testing this be found?
Disclamer: I never looked at C++ before and have no idea what I'm doing so don't get your hopes to high.

tehhowch:
If you are testing code changes, it is wise to

1) avoid using a "real" save
2) give yourself whatever is needed to evaluate results

So just make a system in which AI ships with Grab-Strikes spawn, give your own ship Grab-Strikes, and spawn yourself there. Tweak, recompile, and relaunch as needed.

NCGThompson:
[...]

PR #3799, opened 2018-06-25
Improved Gunnery Algorithm for Projectiles with Drag

NCGThompson:
In response to issue #3686.
This causes turrets to correctly compensate for turret's velocity. It does so by finding the amount of drift caused by the velocity for the calculated rendezvous time. Since the correct rendezvous time is also dependent on offset of the turret velocity, they are calculated recursively with a small number of repetitions.
The math is explained [in this video](https://youtu.be/pOFl0BhUkDI?t=6m49s).
Balance compensation for the Korath Grab-Strike may be worth consideration.
The amount of repetitions can be changed by changing the initial value of 'iterations'.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2472,8 +2472,8 @@ void AI::AimTurrets(const Ship &ship, Command &command, bool opportunistic) cons
                         }
                         else
                         {
-                            //Calculate how much drift turret velocity will in that time.
-                            offset = ship.Velocity() * -2. * M_PI * expm1(-rendezvousTime * M_1_PI * 0.5);
+                            //Calculate how much drift turret velocity will cause in that time.
+                            offset = ship.Velocity() * -expm1(-rendezvousTime * weapon->Drag()) / weapon->Drag();
                         }
                         
                         iterations --;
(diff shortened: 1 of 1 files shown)


source/AI.cpp
```diff
@@ -2442,39 +2442,60 @@ void AI::AimTurrets(const Ship &ship, Command &command, bool opportunistic) cons
...
-				// weight. Outside that range, give them lower priority.
-				rendezvousTime = max(0., rendezvousTime - weapon->TotalLifetime());
-				// Always prefer ships that you are able to hit.
-				double score = turnTime + (180. / weapon->TurretTurn()) * rendezvousTime;
-				if(score < bestScore)
-				{
-					bestScore = score;
-					bestAngle = degrees;
-				}
+                Point p = target->Position() - start;
```
tehhowch: This is a `tabs` repo :)

For comments, they should be sentences and start with a space after the `//`

source/AI.cpp
```diff
@@ -2456,7 +2456,7 @@ void AI::AimTurrets(const Ship &ship, Command &command, bool opportunistic) cons
                 double rendezvousTime = 0.;
                 if(weapon->Acceleration())
                 {
-                    //Uses iterative approxamation for projectiles with drag.
+                    // For projectiles with drag, use iterative approxamation.
```
Bladewood: approximation

source/AI.cpp
Bladewood: Please use actual tabs, not space quadruplets or another form of indentation.

Amazinite:
Closing as this has been marked unlikely.

PR #3800, opened 2018-06-26
Add Heat Bomb Launcher

pscamman:
Hey everyone! I've just been getting started tinkering with this repo. Very exciting. I figured I'd throw a very simple pull request up just to start getting feedback.

### In-game description
Heat Bombs are made with a special material that absorbs heat phenomenally well. Unfortunately, for those interested in designing dedicated cooling systems, it also becomes severely unstable when superheated. Faced with this challenge, one clever engineer decided it would be best to simply eject the material from the ship. In the direction of an enemy.

### More details
This is basically a shorter range, extra strong Heavy Rocket with heat damage and the added twist of reducing your heat rather than adding to it. I added a simple check in the C++ code to prevent weapons from firing if they would drop your heat below 0 (I believe this is the first weapon to deduct heat).

I don't know anything about creating art assets, but if that would be expected of me then I'd be more than glad to learn how to do all that. For now this weapon just borrows assets from the Torpedo.

For now, I added the weapon to outfitters that carry Plasma Cannons. I did not add it to any ships.

Lastly, I added one line to the .gitignore to accommodate my IDE, and changed a few words in the developer readme to make a sentence more clear. If I should really separate those changes into a different pull request, I could easily do that.

### Weapon stat highlights
```
cost 200000
"outfit space" -35
"weapon capacity" -35
"heat bomb capacity" 20
"velocity" 10
"lifetime" 50
"reload" 120
"firing heat" -1000
"trigger radius" 30
"blast radius" 50
"shield damage" 900
"hull damage" 900
"heat damage" 500
"hit force" 800
"missile strength" 35
```

PR #3802, opened 2018-06-26
Heat as weapon resource

pscamman:
Simply prevents a weapon from firing if it would reduce your heat below 0.
To make use of this, create weapons with negative heat costs.

Isolated from other changes as recommended in #3800.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2756,6 +2756,8 @@ bool Ship::CanFire(const Weapon *weapon) const
 	
 	if(energy < weapon->FiringEnergy())
 		return false;
+	if(heat < -(weapon->FiringHeat()))
+		return false;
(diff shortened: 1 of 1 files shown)


DingusShingleton:
N.B. this could also be used to create an active heat vent outfit. Maybe a secondary weapon that simply vents excess heat

Kriegstofu:
a similar use of heat could be to heal ships, for example a magmatic/fire race that has living ships similar to void sprites but with a magma core, whenever a ship overheats, an active cooling drops its heat to 0 (and at the same time triggers a active healing/repair, which is only active when the active cooling is active.
This way one could make ships that are immune to heat (fire ships) or the opposite, active taking damage when heat is over a treshold, like melting ice ships

PR #3803, opened 2018-06-27
Add auto-hire check box to Hiring Panel

pscamman:
When checked, immediately maxes your crew and then will automatically max your crew every time you land, so long as it stays checked.

Because I always, always forget :)

I borrowed the check box code from the outfitter check boxes. Perhaps it should be abstracted out somewhere so that it doesn't have to be copy-pasted in this way?

### DIFF
--- a/source/HiringPanel.cpp
+++ b/source/HiringPanel.cpp
@@ -12,12 +12,18 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include "HiringPanel.h"
 
+#include "Color.h"
 #include "FillShader.h"
+#include "Font.h"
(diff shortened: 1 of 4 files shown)


tehhowch:
Can you elaborate on the differences between this and #3089?

pscamman:
@tehhowch Oh! I must have missed #3089 when searching the issues/pulls to see if this had been done before. I don't think this is much different, other than being only for maxing out crew and including the check box in the UI. To be honest, when I saw "desiredCrew" in the code I assumed it had to do with the old mechanic of simply rehiring up to the minimum amount. Feel free to close this pull if it's unnecessary.

PR #3805, opened 2018-06-28
EMP license

Nescio0:
All Remnant weapons and outfits require a Remnant license. The exception is the new EMP Torpedo Bay. My guess is this is not intentional, hence this patch, which applies the Remnant license to this weapon as well.

### DIFF
--- a/data/remnant outfits.txt
+++ b/data/remnant outfits.txt
@@ -170,6 +170,8 @@ effect "point defense die"
 outfit "EMP Torpedo"
 	plural "EMP Torpedoes"
 	category "Ammunition"
+	licenses
+		Remnant
 	cost 4400
 	thumbnail "outfit/emp torpedo"
 	"mass" 1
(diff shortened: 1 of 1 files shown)


Bladewood:
I'll mention that the Finisher Pod requires the Heliarch license, yet the Torpedo itself does not.

Nescio0:
Thank you for pointing that out. Also, the Thunderhead launcher requires a launcher, but the missile does not. Probably the same should apply to the Remnant.

Nescio0:
Thank you for pointing that out. Also, the Thunderhead launcher requires a license, but the missile does not. Probably the same should apply to the Remnant.

Bladewood:
I'm not sure whether that's a good thing, honestly: maybe a licensed launcher can be stolen from a (newly formed) faction within the topic species, or from a different species that stole from the topic species; and then we'd be able to reload our stolen launchers at their outfitters, which is wrong no matter how you look at it.

Amazinite:
>I'm not sure whether that's a good thing, honestly: ...

Maybe that's the point.

PR #3808, opened 2018-06-30
Fix XCode libraries

tiennou:
This fixes some linker issues I had trying to build against brewed dependencies.

- The path to SDL2.dylib was relative to the project, so it would only have worked if your ES checkout was made at the correct place.
- The strange header search path cause the OS-provided libjpeg (in `/usr/{include,lib}`) to be found first.

### DIFF
--- a/EndlessSky.xcodeproj/project.pbxproj
+++ b/EndlessSky.xcodeproj/project.pbxproj
@@ -169,7 +169,7 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
-		4C2DEF55201B8FAD0062315E /* libSDL2-2.0.0.dylib */ = {isa = PBXFileReference; lastKnownFileType = "compiled.mach-o.dylib"; name = "libSDL2-2.0.0.dylib"; path = "../../../usr/local/Cellar/sdl2/2.0.7/lib/libSDL2-2.0.0.dylib"; sourceTree = "<group>"; };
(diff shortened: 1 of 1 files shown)


EndlessSky.xcodeproj/project.pbxproj
```diff
@@ -169,7 +169,7 @@
 /* End PBXCopyFilesBuildPhase section */
 
 /* Begin PBXFileReference section */
-		4C2DEF55201B8FAD0062315E /* libSDL2-2.0.0.dylib */ = {isa = PBXFileReference; lastKnownFileType = "compiled.mach-o.dylib"; name = "libSDL2-2.0.0.dylib"; path = "../../../usr/local/Cellar/sdl2/2.0.7/lib/libSDL2-2.0.0.dylib"; sourceTree = "<group>"; };
+		4C2DEF55201B8FAD0062315E /* libSDL2-2.0.0.dylib */ = {isa = PBXFileReference; lastKnownFileType = "compiled.mach-o.dylib"; name = "libSDL2-2.0.0.dylib"; path = "/usr/local/Cellar/sdl2/2.0.8/lib/libSDL2-2.0.0.dylib"; sourceTree = "<absolute>"; };
```
tehhowch: Did you mean to bump the version as well?

tiennou: Not really. It just used the version that I have installed (which is up to date).

thomasballinger:
re version bump, in order to build I had to symlink /usr/local/Cellar/sdl2/2.0.8 to 2.0.7 (because I don't know how to do anything in Xcode) so this change would have made this easier for me.

Fzzr:
That will just happen again to the next person whenever 2.0.9 or whatever comes out due to the current installation instructions. Brew installs the most recent version, and point releases of libSDL aren't kept around. I had to manually install the older version from github to make it work with 2.0.7. See: https://danepowell.com/blog/homebrew-formula-versions

Fzzr:
I'm now finding that on Mojave and Xcode 10.1, the build works but simply displays a black screen.

Issue #3807, opened 2018-06-30
Uninhabited planets can fine you for illegal outfits


jonnybot0:
During the Remnant Gascraft mission, it's possible for the uninhabited gas giants to fine you if you have an escort possessing illegal outfits.
<img width="1552" alt="screen shot 2018-06-30 at 4 37 00 pm" src="https://user-images.githubusercontent.com/3723592/42129248-3b942434-7c85-11e8-9c43-608a59611c28.png">

Bladewood:
Relevant: #2059.

Amazinite:
Instead of continually fixing this every time MZ adds a new uninhabited planet and doesn’t give it `security 0` (e.g. #2821), perhaps uninhabited planets should be changed in the code to inherently be unable to fine the player.

PR #3810, opened 2018-07-01
Add security 0 to gas giants in Nenia

Tadrix:
… to prevent players from being fined.
Hopefully it's a temporary solution before either all planets without security attribute default to 0, or all uninhabited planets / planets in uninhabited systems default to security 0.


Zitchas:
Seems like a good solution for now.

MCOfficer:
ref #3807

PR #3813, opened 2018-07-03
Made pirates hostile to Pug

temtemy:
While doing my own solo Quicksilver challenge, I made the Pug jump to Orvala where the pirate ship "Dread" lurks. I expected that the pirates will attack the Pug, but it didn't happened. It seems it is no longer the case in master, unlike in 0.9.8. That doesn't make sense. This PR makes the pirates hostile to the Pug, while not affecting either faction's reputation.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -265,6 +265,7 @@ government "Pirate"
 		"Hai" -.01
 		"Korath" -.01
 		"Merchant" -.1
+		"Pug" -.01
 		"Syndicate" -.01
 	"bribe" .05
 	"fine" 0


KiLEdEnNis:
I'm not sure if this is worth it for the same reason the Navy & Korath aren't hostile
on the other hand, doesn't hurt

Zitchas:
As far as that goes, I think the Pirates should be innately hostile to everyone. That's basically their modus operandi: "If I can see it, I try to loot it. It doesn't matter how fancy it looks, o

Of course, there's possible plot exceptions (Alphas, Bounty hunters, etc?) but as a general rule, they should attack everyone.

KiLEdEnNis:
with the negative -.01 system we might as well use it for super rare cases tbh

tehhowch:
#3311

temtemy:
Pug crew member: Captain, those pirates are attacking those poor ships. Should we help?

Pug captain: Nah, just ignore those humans.

Not the Pug I expect.

Wrzlprnft:
to be honestly, that's exactly what i expect. given that the pug only interfere when things are about to escalate.

endless-sky:
In the Pug mode of thinking, nothing that you do to a small group of individuals qualifies as a crime. If you took the entire crew of a human starship and slaughtered them in a gruesome way or exiled them to a different galaxy or replaced their brains with sentient cabbages, the Pug might consider that wasteful or unnecessary or in poor taste, but they wouldn't actually consider it *immoral* per se. The only thing that bothers the Pug is an act that threatens the extinction of an entire species or a distinct subset of that species, because they don't actually think of individuals as discrete entities with their own rights.

KiLEdEnNis:
> or replaced their brains with sentient cabbages

is that a spoiler for something unimplemented yet? :P

endless-sky:
All I'm saying is, Arfectas are *far* from being the worst thing the Pug are capable of.

Although, that particular prank wasn't pulled by the Pug collective that is responsible for overseeing the humans, but by a different group that was visiting them... and those Pug are *no longer welcome* at parties on Pugglemug.

jostephd:
> Although, that particular prank wasn't pulled by the Pug collective that is responsible for overseeing the humans, but by a different group that was visiting them... and those Pug are no longer welcome at parties on Pugglemug.

Now I am curious as to what an interstellar conflict between two Pug factions might look like.

PR #3814, opened 2018-07-03
Added landing clearance for "Ambassadors from Earth"

temtemy:
If the player didn't disabled some pirate ships before "Ambassadors from Earth", they will have to bribe the spaceport authorities, which doesn't make sense, because you're going to Earth to escort their ambassadors. This PR gives landing clearance.

[Here's my save file that the above scenario can happen](https://github.com/endless-sky/endless-sky/files/2158627/Marcelo.Pilar.txt)


temtemy:
Also the "Ambassadors to Earth" mission.

PR #3819, opened 2018-07-05
Add a "no shield generation" flight check.

jostephd:


source/Ship.cpp
```diff
@@ -737,6 +739,8 @@ string Ship::FlightCheck() const
 		return "solar power?";
 	if(!hyperDrive && !jumpDrive && !canBeCarried)
 		return "no hyperdrive?";
+	if(!repair && !shield)
```
tehhowch: This checks both hull and shield regen, and absent both, shows a "no shield gen" message. What about ships that have no shields?

Brick63:
some players may choose not to have a shield generator so this seems to limit choices

source/Ship.cpp
jostephd: You mean, if a ship has hull repair but no shield generation? I'm not sure what to do in that case.

jostephd:
This doesn't prevent playing with no shield generator. It simply makes the outfitter show a yellow question mark if there's no shield generator, same as when there's no hyperdrive.

source/Ship.cpp
tehhowch: No, I mean literally, if a ship has no shields at all, then a shield regenerator is a waste of space for it,  Unless the available content for the player includes hull repair tech, then they will have an unremovable warning symbol.

Amazinite:
The thing is, having no shield generator is a far different scenario to having no hyperdrive. In one situation, you can fly around and simply have to be careful about combat. In the other situation, you can't even leave the system you're in. 

Having no shield generator, at least for me, is a common occurrence, and not something I would expect the outfitter to scream at me about. And as Tehhowch pointed out, what of having no hull repair or what if a future alien faction has ships with no shields and therefore no need for shield generators?

endless-sky:
When I'm starting a new game, one of the first things I always do is sell my ship's shield generator to make room for more cargo or bunks. The "flight checks" are really intended for situations that could leave a ship completely stranded, not for minor things like a lack of shield recharge.

Cerebrate-Daggoth:
jsut an idea, perhaps a question mark or something if there's no shield or
hull regeneration happening at all?

temtemy:
What about the ships which intentionally doesn't have a shield generator pre-installed? Like the boxwing. They will get question marks too?

Cerebrate-Daggoth:
do a check for shield capacity with an 'if/then' if shield capacity >0
display question mark. something like that. i'm not a coding master but
something along those lines should work.

KiLEdEnNis:
I'm not waiting for an annoying unremovable sign if I choose not to have shield regen
power is needed, a temperature below max is needed, engines are needed, a drive is needed, those things make a ship function
shield generation is a choice, like weapons, you're not gonna have a check saying you don't have weapons either

Cerebrate-Daggoth:
i know, i don't like to use shield generators either. perhaps a pop up or
something with an option to 'never show again for this ship' or something.

On Thu, Jul 5, 2018 at 8:46 PM, Dennis van Anrooij <notifications@github.com

jostephd:
Thanks for the feedback everyone.

PR #3820, opened 2018-07-06
Set Heliarch's hostile hail

temtemy:
I am wondering why the heck my hostile Heliarch hails doesn't appear, and I know now why. Silly me.


MCOfficer:
i love how we all approved of that. shame on us.

PR #3822, opened 2018-07-07
Use absolute rather than relative path for libSDL2

Fzzr:
This fixes a build failure on macOS when the `endless-sky` repo's directory isn't in the developer's homedir (I keep my code in `~/repos/`).


Fzzr:
Actually, #3808 is better.

PR #3824, opened 2018-07-07
Lower "Kestrel Testing" requirements

Tadrix:
… from `(9) force to be reckoned with` to `(8) respected foe` in order to prevent the *Kestrel* from being already outdated when it is unlocked.

### DIFF
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -14,7 +14,7 @@ mission "Kestrel Testing"
 	source "Wayfarer"
 	waypoint "Umbral"
 	to offer
-		"combat rating" > 8000
+		"combat rating" > 2980
 	
 	on offer
 		conversation


data/kestrel.txt
```diff
@@ -14,7 +14,7 @@ mission "Kestrel Testing"
 	source "Wayfarer"
 	waypoint "Umbral"
 	to offer
-		"combat rating" > 8000
+		"combat rating" > 2980
```
Bladewood: To be clear, combat rating 9 is actually 8103, not 8000, so evidently Michael was looking for a round number here.

jafdy: Having it linked to a number displayed in-game makes more sense as then we can point to that number. While the 8000 was likely because it is a round number that is not an important factor in choosing the number.

Amazinite:
Not sure how I feel about this. Dropping it from 9 to 8 might seem insignificant, but as you can see, that's a drop from 8000 to 2980, which makes the Kestrel about 2.7 times easier to get to. And I'm not too sure about doing this "in order to prevent the Kestrel from being already outdated when it is unlocked." I've had playthroughs where I get the Kestrel early and I've had playthroughs where I get it late. It seems to me that whether the Kestrel is outdated by the time they get it depends on the player. so I'm not sure that making it considerably easier to obtain is necessary.

Zitchas:
Personally, I like it with the lower accessibility. I usually don't hit 8 sometime late in the FW story, and I've usually long since moved on to using SB's or Remnant ships by the time the Kestrel is unlocked. It usually leaves me wishing that this was a human ship built over in Hai space using one of their shipyards (and thus their tech).

Janaszar:
How are you determining "outdated"?
You can easily trade for money and buy shield beetles right from the start, or you could go combat-heavy and fight/capture pirates and end up having the Kestrel be a good improvement.
Not to mention, going with FW path means you can eventually capture Carriers without (long-term) issues.

Zitchas:
I think that's the point. Reducing the combat requirements a bit means it will be more likely to be available when players are reaching the stage of "falcons and leviathans aren't cutting it anymore," and less likely to be opened up after they've expanded their horizons to alien races.

And a combat rating of  8 still isn't particularly wimpy, either.

data/kestrel.txt
Bladewood: I mean, sure, but then it just might feel a bit too convenient or unnatural (in-game, as opposed to someone just looking right at the code, because 8000 maps to 8.97 rather than 9). I agree with your statement, though, it might be better overall to change the scheme here.

Tadrix:
> How are you determining "outdated"?

> going with FW path means you can eventually capture Carriers

The moment you're buying Beetles or capturing Carriers Kestrels are outdated.

Bladewood:
Don't even need to buy beetles, just steal them from the Unfettered with a good transport.

ghoulavenger:
The Kestrel may be the best human available ship, but it isn't up to snuff with the Shield Beetle (which is available from the start, meaning the Kestrel is already obsolete when you get it no matter what). Since the FW campaign can be done at any time I don't really consider this necessary either, since you'd have to be playing the campaign early for this to be the real "problem" so to speak.

So I'm not for or against this change. I find combat rating 9 pretty easy to get when I start farming pirates for Leviathans or farming the Korath for jump drives. Which leaves pretty much players that aren't combat oriented getting it a little bit later? And I find that perfectly fine.

KiLEdEnNis:
> The Kestrel may be the best human available ship

1. *warship, not ship
2. top 5 sure, not best though imo

but the Kestrel is more of a half easter egg ship, lowering the requirements sure but you have to take 1 down to get it

Wrzlprnft:
>  I find combat rating 9 pretty easy to get when I start farming pirates for Leviathans or farming the Korath for jump drives. Which leaves pretty much players that aren't combat oriented getting it a little bit later? And I find that perfectly fine.

Given that you are an experienced player, this is not the norm. Or at least it shouldn't be, once the intro's are all over the place, and the non-experienced player picks up the story pretty much from the get go with the natural progression from a medium warship onwards.

Since most veteran players hardly even aim for the kestrel, cause shield beetle etc. and more unlock it as collateral damage, i don't think it would really matter for those anyway.

Maybe wait until other story intros and such help define the ship progression, and then adjust the kestrel's rating requirement to kick in the mission at the right time.

ghoulavenger:
@Wrzlprnft As a newbie player I didn't even discover how the quest system worked until I was combat rating 14.

ghoulavenger:
@Wrzlprnft You're trying to use my experience as a justification that it shouldn't be that way, when it is that very experience that makes your entire argument moot. Only an experienced player would know about the quest, let alone the rating requirement. So making it available to a new player is a seeming improbability if it takes an experienced one to know about it to begin with. The problem isn't in the rating requirement in this case. The problem is the quest system in general. But that has nothing to do with this PR.

Wrzlprnft:
that's the thing what i said with the intros and such: players will expect quests all over the place, so making it available via lower requirements so it fits in a new players progression do i prefer over keeping it a easter eggy trophy for experienced players. 
the pr is basically the former without the story context.

Nescio0:
Actually I had unlocked the Kestrel long before I discovered the Free Worlds campaign, not because I'm an experienced player who knows what he's doing, but because I was a new player who did not really know what to do. I'm neither opposed to this change nor in favour; I simply don't see the need.

Zitchas:
Experienced players who want the Kestrel will get the Kestrel regardless. I don't think it is much of an issue for them, although lowering the rating by one probably will increase the chance that they do use it before switching to Shield Beetles or other plot-locked ships. 

That being said, having a slightly lower rating requirement means that less experienced players are more likely to stumble across it at a point in their career where they are more likely to find it useful and worth getting. 

In my view, this comes down to a choice in purpose: Is there any desire for this ship to be a useful addition to the player's options where they are likely to at least consider using it for story? Or is this purely an Easter Egg that people admire for a few minutes, then never look at again?

If the point is the former, then I think the rating requirement should be dropped. If it is the latter, then leave as is. I doubt there really is an "average player," but given how easy the FW story is to get into, I suspect that there aren't too many who reach combat rating 8 without being at least a moderate ways into the campaign.

As a side note, every time I've unlocked the Kestrel I have been flying a Shield Beetle or better. (as in more shields and/or  hull). So the Kestrel's always been a bit of a disappointment for me.

Amazinite:
Closing this, as interest is more toward #3976 to help avoid the Kestrel from being outdated.

Issue #3771, opened 2018-06-11
Request: Projectile Velocity in Outfitter


ZBok:
First huge thanks for this incredible game.

I feel that projectile velocity is super relevant when choosing weapons and would be great to see in the outfitter for relevant guns/turrets, at least where it can be expressed simply.

Bladewood:
I think this will suffer problems when we consider multi-stage weapons like the Heliarch finisher pod, or the Korath minelayer.

Faptown:
For sure, I know some proposed weapons and even maybe some existing could be problematic, but if possible just for the basic weapons where it can be expressed simply it would be great.

EG: I at first completely dismissed Inhibitors as being rather weak but after re-visiting them much later, just messing around, I noticed they have ridiculously high proj velocity and are much more viable since they barely miss ever and make these weapons more viable for reverse kitey piloting.. Similarly other weapons are less viable at hitting small/fast targets with lower projectile velocity.

If of course it's not able to be done without much hassle, no probs :D Just a thought. Thanks for the amazing game

endless-sky:
When it comes to questions of how much information to present to the player, it's always a balance because too much information can be hard to sort through and therefore can be *less* informative than a smaller amount of key information.

With projectiles, in most cases the range is much more important than the velocity. And, in nearly all cases the player will have seen a weapon in action before they decide to buy it. So, they'll already have a reasonably good sense of how quickly it moves.

MessyMix:
Actually, this is really not a bad idea, but I think that yeah, we have to take information clutter into consideration.

jafdy:
Is there a nice way to define it in words. That might be a workable median.

Nescio0:
It could be displayed on one and the same line:
`projectile speed / range:`
Like we already do with `shields charge / max:` etc.

PR #3834, opened 2018-07-09
Display projectile velocity in outfitter panel

Nescio0:
Display projectile velocity in outfitter panel (#3771).

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -335,6 +335,9 @@ tip "ammo:"
 tip "range:"
 	`Total range of this weapon.`
 
+tip "velocity / range:"
+	`The distance the projectile can travel per second and in total (range is velocity times lifetime).`
+
 tip "shield damage / second:"
 	`Shield damage, converted to damage per second to allow easy comparisons between different weapons.`
 
--- a/source/OutfitInfoDisplay.cpp
+++ b/source/OutfitInfoDisplay.cpp
@@ -232,8 +232,9 @@ void OutfitInfoDisplay::UpdateAttributes(const Outfit &outfit)
 		attributesHeight += 20;
 	}
 	
-	attributeLabels.emplace_back("range:");
-	attributeValues.emplace_back(Format::Number(outfit.Range()));
+	attributeLabels.emplace_back("velocity / range:");
+	attributeValues.emplace_back(Format::Number(outfit.Velocity())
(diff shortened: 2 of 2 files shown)


tehhowch:
UI pull -> include screenshots of relevant things.

Nescio0:
Before:
screenshot from 2018-07-09 15-08-14
After:
screenshot from 2018-07-09 15-09-13
(particle cannon)

Nescio0:
Range = velocity * lifetime, therefore it makes sense to display it on the same line. 
Velocity is per frame, i.e. 60 times as high? I'll correct it.

Nescio0:
Here you go:
screenshot from 2018-07-09 15-52-29

Nescio0:
Here you go:
screenshot from 2018-07-09 15-58-50

KiLEdEnNis:
1800 velocity & 750 range means it'd travel 1800 pixels per frame, 1 frame being minimum, and dividing it by the 750 range a particle would only exist 5/12th of a frame

not sure what you're trying and honestly I didn't read it all but you messed up velocity

tehhowch:
> 800 velocity & 750 range means it'd travel 1800 pixels per frame, 1 frame being minimum, and dividing it by the 750 range a particle would only exist 5/12th of a frame

@KiLEdEnNis velocity units as used by the game are indeed distance per frame, (`30` for the Particle Cannon), but velocity units used by humans are generally distance per second. Note that velocity **displayed** for ships is given in distance/second, not distance/frame: so this representation (`30 * 60` -> `1800`) is consistent with existing displayed data:
https://github.com/endless-sky/endless-sky/blob/e574ba53f0b2d2ec4afac4f3685affd97eb60843/source/ShipInfoDisplay.cpp#L197-L198

KiLEdEnNis:
my mistake then, but isn't this gonna confuse a lot of people?

gunqqer:
Probably not unless they are aware of how ES does velocity, but then they *should* be able to figure it out. Or we could add a tooltip explaining how it's in dist/sec rather than dist/frame

jafdy:
People don't know that the "physics" simulation happens at 60fps so they won't automatically assume that the number refers to the distance travelled per frame.

Nescio0:
Added a tooltip, just in case anyone might be confused.

Amazinite:

PR #3835, opened 2018-07-09
Display cloaking energy and heat

Nescio0:
Display cloaking energy and heat in ship information table: 
screenshot from 2018-07-09 14-59-56

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -293,6 +293,9 @@ tip "repairing hull:"
 tip "charging shields:"
 	`Energy consumed and heat generated when this ship is recharging its shields. This is in addition to the "idle" amount.`
 
+tip "cloaking:"
+	`Energy consumed and heat generated when this ship is cloaked. This is in addition to the "idle" amount.`
+
 tip "max:"
 	`Energy storage capacity and maximum safe level for heat generation. Energy capacity allows a ship to temporarily draw more energy than it produces. The heat maximum is the highest your heat generation can be without eventually causing your ship to overheat.`
 
--- a/source/ShipInfoDisplay.cpp
+++ b/source/ShipInfoDisplay.cpp
@@ -301,6 +301,12 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 	double hullHeat = attributes.Get("hull heat");
 	heatTable.push_back(Format::Number(60. * (shieldHeat + hullHeat)));
 	attributesHeight += 20;
+	tableLabels.push_back("cloaking:");
+	double cloakingEnergy = attributes.Get("cloaking energy");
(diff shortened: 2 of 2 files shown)


Nescio0:
The reason I proposed this is because currently cloaking stats are displayed nowhere. Having to open the internal `data/` files to look up these values seems undesireable to me.
Where does it overflow?
As for spoiling, anyone who presses Esc to open the preferences can see "Toggle cloaking device: C".

Bladewood:
Quotes are taken from https://github.com/endless-sky/endless-sky/pull/3704#issuecomment-385250537.

> Where does it overflow?

I think he means this (I haven't tested it, so I don't know):

> This would make it so that some ship stats don't fit in the ship info panel.

> The reason I proposed this is because currently cloaking stats are displayed nowhere. Having to open the internal data/ files to look up these values seems undesireable to me.

You could say similar of Remnant ships' shield and hull regenerators, which are very energy efficient but merged into one, so you don't know the exact values of each. Or of Korath ships' integrated ramscoops. Or of any ship's drag and heat dissipation.

See also this, the note on new players:

> Also, I think that the number of people who would be confused by those added lines would be at least as many as those who would find them useful - especially for new players, who are already at risk of being overwhelmed with the complexity of the available information.

Not nearly as many players actually look at the preferences compared to the outfitter (sadly), for the simple reason that the outfitter system is probably a million times more important for building a ship.

Cloaking is generally considered very secondary to the other attributes displayed here.

Nescio0:
> Cloaking is generally considered very secondary to the other attributes displayed here.

It can be a significant drain, though. E.g. the Starling: 
`"shield energy" 1.2`
`"hull energy" 0.6`
`"cloaking energy" 5`
By not including this contribution the table is flawed.

How about including cloaking in the same line as shields and hull? Moving also includes different things (thrusters, steering, and afterburner).

Nescio0:
>This is going to overflow the available space, and also spoil the existence of this tech to every player immediately. At minimum you need to rectify both issues before this has a chance to merge

Update: this cloaking stats line is only displayed if a ship can actually cloak. Hopefully that solves one of your concerns.

tehhowch:
Shipyard view is not the view that overflows. Cloak your Carrier

Amazinite:
@pointedstick because that ship doesn’t have the max possible stats to show. As Tehhowch’s pic shows, the Carrier also has fighter and drone bays that push the stats out of the menu.

Nescio0:
Is there an easy way to hide something in the player info overview but display it in shop?

Pointedstick:
That wouldn't be the correct fix here.

Amazinite:
Closing as this steps into the realm of providing too much information, along with the UI issues currently present that have not been addressed. Any efforts to show more stats like this should probably go into creating an advanced stats page. Ref: #4215.

PR #3837, opened 2018-07-10
content: Adds 3 Coalition Outfit/Systems 

Kriegstofu:
Adds to both Coalition Sales (Basic and Avanced) the following 3 Outfits

	"Saryd Outfits Expansion" -  More Space Efficient, but heavier than regular Outfits Expansion
	"Arach Cargo Expansion" - As with Outfits Expansions, but reversed
	"Kimek Bunk Room" -  50% Bunk Increase compared to Bunk Room with 140% Mass Increase

Images are Placeholders and derived from the ship images of Saryd/Arach/Kimek Ships
Stats are not definitive but rather a suggestion.

Basically an outfit per race representing the coalition.


tehhowch:
So basically better in every way with no real penalty. Cost isn't a meaningful difference, nor is mass when that mass is realized as outfit space.

Kriegstofu:
Comparing the Wanderer Ramscoop with the regular Ramscoop, not only has it less outfit space, but also four times its ramscoop value. In the same logical pattern, i tried to make specialized additions to the Coalitions different races. More Mass means less velocity/turning. 
Would you suggest to add some bigger inefficiencies in cooling/heating? or reduce the amount of gained space?

Quantumshark:
The 100% conversion rate on these is definitely too high, and increased mass is not a significant downside. What might be more balanced would be a higher conversion rate in exchange for a less convenient size (for example, 40 space of one type for 50 of the other - conversion rate is 80% rather than 75%, but the larger size means you will likely have more space left over that cannot be converted with more expansions).

Pointedstick:
I'm afraid this is a no-go. MZ has said before that he doesn't want slightly modified alien versions of common outfits without a real reason (e.g. the Wanderer ramscoop reinforces their love of efficiency and recycling). And those sprites would need a total re-do.

PR #3838, opened 2018-07-11
Fixed the Albatross engine flare zoom factors

Amazinite:
I noticed that the right flare was slightly longer than the left flare on my Albatross. Did some digging and it looks like the engine zoom factors were swapped.

### DIFF
--- a/data/remnant ships.txt
+++ b/data/remnant ships.txt
@@ -192,8 +192,8 @@ ship "Albatross"
 		"Hyperdrive"
 	
 	engine -28 155 .8
-	engine 28 155 .9
-	engine 0 184 .8
+	engine 28 155 .8
+	engine 0 184 .9
 	gun 0 -190
 	gun -9 -177 "Inhibitor Cannon"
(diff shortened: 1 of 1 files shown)


KiLEdEnNis:
weren't engines supposed to be from left to right? in which case the central .9 engine flare should be in the center?

Bladewood:
That sounds right. Here are the Punisher's engines:
```
	engine -24 184 .6
	engine 0 187 1.1
	engine 24 184 .6
```

Nescio0:
And the Boxwing's:

	engine 23 -1
	engine -23 -1
	engine 19 29
	engine -19 29

Guns and turrets are first top-to-tail, then left-to-right, so I guess the same applies to engines.

warp-core:
All the Coalition ships follow the left-to-right system irrespective of y-coordinate.
The reason for the guns and turrets being the way they are is that some other orders would make the hardpoints appear in a strange order when rearranging weapon mounting positions in the ship info screen.

PR #3843, opened 2018-07-12
Introduce a "no fuel?" warning

Nescio0:
Currently all ships have some internal fuel capacity. However, it is not inconceivable ships with 0 fuel might be introduced, or outfits that lower the fuel capacity. (In fact I'm working on a plug-in that changes all ships to have 0 fuel.)
This patch introduces a "no fuel?" warning, analoguous to the "no hyperdrive?" warning (#3635).

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -468,4 +468,4 @@ tip "no hyperdrive?"
 	`This ship has no hyperdrive. If it took off, it would be unable to leave this star system.`
 
 tip "no fuel?"
-	`This ship has not enough fuel capacity to make a hyperspace jump. If it took off, it would be unable to leave this star system.`
+	`This ship has insufficient fuel capacity to make a hyperspace jump. If it took off, it would be unable to leave this star system.`
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -464,8 +464,8 @@ tip "limited turn?"
 tip "solar power?"
 	`This ship is powered entirely by solar energy. If you fly too far from the center of a star system, maneuvering may become difficult because your engines will only have a fraction of the power they require.`
 
-tip "no fuel?"
-	`This ship has no fuel. If it took off, it would be unable to leave this star system.`
-
 tip "no hyperdrive?"
 	`This ship has no hyperdrive. If it took off, it would be unable to leave this star system.`
+
+tip "no fuel?"
+	`This ship has not enough fuel. If it took off, it would be unable to leave this star system.`
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -464,5 +464,8 @@ tip "limited turn?"
 tip "solar power?"
 	`This ship is powered entirely by solar energy. If you fly too far from the center of a star system, maneuvering may become difficult because your engines will only have a fraction of the power they require.`
 
+tip "no fuel?"
+	`This ship has no fuel. If it took off, it would be unable to leave this star system.`
+
 tip "no hyperdrive?"
 	`This ship has no hyperdrive. If it took off, it would be unable to leave this star system.`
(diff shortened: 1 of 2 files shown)


source/Ship.cpp
```diff
@@ -735,6 +736,8 @@ string Ship::FlightCheck() const
 		return "limited turn?";
 	if(energy - .8 * solar < .2 * (turnEnergy + thrustEnergy))
 		return "solar power?";
+	if(fuel <= 0. && !canBeCarried)
```
tehhowch: And if the ship requires 0 fuel to jump? You currently would still tell the player they are stuck.

Nescio0:
> Needs at minimum a jump fuel check

Changed to `fuel < JumpFuel(hyperspaceSystem)`.

> possibly also a check for an accessible in -system wormhole.

The hyperdrive check doesn't have this either.

data/tooltips.txt
```diff
@@ -466,3 +466,6 @@ tip "solar power?"
 
 tip "no hyperdrive?"
 	`This ship has no hyperdrive. If it took off, it would be unable to leave this star system.`
+
+tip "no fuel?"
+	`This ship has not enough fuel. If it took off, it would be unable to leave this star system.`
```
Pointedstick: "has not enough fuel" -> "does not have the fuel capacity to make a hyperspace jump"

source/Ship.cpp
```diff
@@ -735,8 +736,10 @@ string Ship::FlightCheck() const
 		return "limited turn?";
 	if(energy - .8 * solar < .2 * (turnEnergy + thrustEnergy))
 		return "solar power?";
-	if(!hyperDrive && !jumpDrive && !canBeCarried)
+	if(!canBeCarried && !hyperDrive && !jumpDrive)
```
Pointedstick: Doesn't seem necessary to re-order these conditions.

tehhowch: Could "justify" it by joining the checks:
```
if(!canBeCarried)
{
    if(!hyperdrive && !jumpdrive)
        return "no hyperdrive?";
    if(fuel < JumpFuel())
        return "no fuel?"
}
```

data/tooltips.txt
```diff
@@ -466,3 +466,6 @@ tip "solar power?"
 
 tip "no hyperdrive?"
 	`This ship has no hyperdrive. If it took off, it would be unable to leave this star system.`
+
+tip "no fuel?"
+	`This ship has not enough fuel capacity to make a hyperspace jump. If it took off, it would be unable to leave this star system.`
```
Pointedstick: "has not enough fuel" isn't grammatically correct in English. It needs to be "doesn't have enough fuel" or "has insufficient fuel" or something like that instead.

Nescio0: Actually "has not enough" *is* grammatically correct in English. Maybe it's less common in American, but that's a matter of idiom, not grammar.

source/Ship.cpp
```diff
@@ -735,8 +736,13 @@ string Ship::FlightCheck() const
 		return "limited turn?";
 	if(energy - .8 * solar < .2 * (turnEnergy + thrustEnergy))
 		return "solar power?";
-	if(!hyperDrive && !jumpDrive && !canBeCarried)
-		return "no hyperdrive?";
+	if(!canBeCarried)
+	{
+		if(!hyperDrive && !jumpDrive)
+			return "no hyperdrive?";
+		if(fuel < JumpFuel(hyperspaceSystem))
```
tehhowch: `JumpFuel()`

Per my earlier comment:
> You also should not supply a destination system. Note that `hyperspaceSystem` is only populated while accelerating into a hyperspace jump, so in practice this will be like supplying no destination, but it certainly appears otherwise.

PR #3845, opened 2018-07-14
tweaked javelin reload time

Nescio0:
Currently the javelin pod (size 12) only has a better capacity than the meteor missile launcher (size 10), but in all other respects is significantly worse. This patch reduces the javelin's reload time from 20 to 15, which, although still not great, makes it at least a viable choice.

Before:
screenshot from 2018-07-14 13-37-44

After:
screenshot from 2018-07-14 13-39-14

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -774,7 +774,7 @@ outfit "Javelin Pod"
 		"inaccuracy" 1
 		"velocity" 10
 		"lifetime" 60
-		"reload" 20
+		"reload" 15
 		"firing energy" .2
 		"firing heat" 12
 		"shield damage" 48


Bladewood:
The point of the Javelin, as far as I can see, is that the massive ammo capacity and the rapid fire rate practically make it a blend of a gun and a missile launcher. (Currently, it takes twice as long as a meteor launcher to exhaust its ammunition.) From that perspective, this pull request makes it too powerful at close range.

Nescio0:
Yes, I know, and I fully agree with that purpose, hence this proposal. Currently the smaller Meteor is more effective in all aspects, including on short range, than the Javelin. By slightly improving the fire rate, the larger Javelin becomes more powerful (damage per second) at short range than the Meteor, as it's supposed to be.
With this tweak the Javelin Pod empties in 50 s, still significantly longer than the Meteor Launcher's 28 s or the Sidewinder's 33 s.

Lineth:
Anything that makes javelins even kinda-sorta dangerous is fine. As it is the things are a joke. Shame cause I have some fond memories of surprisingly-effective raven rocket spam from old EV Nova

PR #3849, opened 2018-07-15
Remove ramscoop from Sestor fighter and drone

Nescio0:
The Kor Sestor "Far Osk 27" fighter and "Far Lek 14" drone somehow have a ramscoop value of 3. They don't have fuel nor a hyperdrive, therefore this is rather useless and should be removed.


KiLEdEnNis:
rip Flamethrower drones

PR #3851, opened 2018-07-16
Center the turret on Pointedstick's Vanguard

czartrak:
While messing around with the hull config for Pointedstick's Vanguard for a ship in my mod, it was brought to my attention that the turret was off-center. Upon checking the actual ship, I found this was the case. This change should _hopefully_ center the turret.

### DIFF
--- a/data/persons.txt
+++ b/data/persons.txt
@@ -333,7 +333,7 @@ person "Captain Nate"
 		gun -31 -45 "Heavy Rocket Launcher"
 		gun 31 -45 "Heavy Rocket Launcher"
 		gun 31 -45 "Heavy Rocket Launcher"
-		turret 0 40 "Anti-Missile Turret"
+		turret -1 44 "Anti-Missile Turret"
 		explode "tiny explosion" 18
 		explode "small explosion" 36
 		explode "medium explosion" 24


Nescio0:
After having a look at it, I'd say `0 43` from the centre (i.e. at 79,177 in the sprite); the difference is not terribly important, though.

KiLEdEnNis:
if your fix is right, all other hardpoints are off by 1 pixel as well, which actually means that the ship image itself isn't centered properly in the game files (like I found in #3071 with the Marauder Leviathan)
I found out that it's true; the Pointedstick Vanguard has 1 more pixel on the right
image

so instead of fixing something slightly broken by making it slightly more broken, I'd say grab the Blend, re-render, re-paste it centered in the GIMP file, and PR those instead (with a link to the edited GIMP file and the @2x version)

jafdy:
Wouldn't it be easier to add a line of blank pixels on the left or remove a line of blank pixels on the right?

KiLEdEnNis:
could be nice to do that since I remember the Blender render image isn't even the correct size to paste in GIMP

KiLEdEnNis:
spent over an hour trying to figure out how to get a branch of endless sky that doesn't automatically have my current open PR commits and got sick of it

here's the updated Blend, GIMP file, @2x & base image files
turret should be changed to 0 42

czartrak:
i'll just close this PR and open up a new one for updating the images

PR #3852, opened 2018-07-17
Make Deep visible from Clink

DrBlight:
With this image, we can now see Deep from the surface of its own moon, Clink! High DPI version available here: https://github.com/endless-sky/endless-sky-high-dpi/pull/50

By the way, unlike the other times I've done this, I couldn't find any reference to the original image (desert3.jpg) in the copyright file. I'm not sure if this was an oversight or not, but I assumed it was a public domain image, as the others have been. I did note that my change to the image belongs in the public domain as well.


endless-sky:
I really like this idea, but your version doesn't quite look realistic. Physically speaking, a moon / planet in the sky can never be darker than the sky itself, because all it is doing is adding more light to that part of the sky. I played around with the source image a bit and found that using the "screen" blending mode gives a much more natural-looking image (d595604). (Additive blending would be more "correct" from a physical standpoint, but causes saturation in the brightest spots.)

PR #3861, opened 2018-07-28
doubled blaster velocity, halved lifetime

Nescio0:
Doubled blaster velocity from 8.5 to 17 (modified from 8 to 16) and halved lifetime from 60 to 30, to keep range unchanged. For comparison, particle cannon has velocity 30 and lifetime 25, sidewinder 11 and 600. 
The primary reason for this change is because some ships could currently outspeed blaster bolts, which is ridiculous. With this change they become harder to dodge and the blaster is somewhat less useless as a consequence.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -21,8 +21,8 @@ outfit "Energy Blaster"
 		sound "blaster"
 		"hit effect" "blaster impact"
 		"inaccuracy" 3
-		"velocity" 8.5
-		"lifetime" 60
+		"velocity" 17
+		"lifetime" 30
 		"reload" 12
 		"firing energy" 9.6
 		"firing heat" 30
@@ -47,8 +47,8 @@ outfit "Blaster Turret"
 		"hit effect" "blaster impact"
 		"inaccuracy" 3
 		"turret turn" 3.7
-		"velocity" 8.5
-		"lifetime" 60
+		"velocity" 17
+		"lifetime" 30
 		"reload" 6
 		"firing energy" 9.6
 		"firing heat" 30
@@ -73,8 +73,8 @@ outfit "Quad Blaster Turret"
 		"hit effect" "blaster impact"
 		"inaccuracy" 3
 		"turret turn" 3.2
-		"velocity" 8.5
-		"lifetime" 60
+		"velocity" 17
+		"lifetime" 30
 		"reload" 3
 		"firing energy" 9.6
 		"firing heat" 30
@@ -104,8 +104,8 @@ outfit "Modified Blaster"
 		sound "mod blaster"
 		"hit effect" "blaster impact"
 		"inaccuracy" 4
-		"velocity" 8
-		"lifetime" 60
+		"velocity" 16
+		"lifetime" 30
 		"reload" 12
 		"firing energy" 10.8
 		"firing heat" 42
@@ -130,8 +130,8 @@ outfit "Modified Blaster Turret"
 		"hit effect" "blaster impact"
 		"inaccuracy" 4
 		"turret turn" 3.5
-		"velocity" 8
-		"lifetime" 60
+		"velocity" 16
+		"lifetime" 30
 		"reload" 6
 		"firing energy" 10.8
 		"firing heat" 42


Pointedstick:
TBH I think double is too fast, but I'd support a 50% speed increase.

Nescio0:
Too fast? Compared to? The closest competitors of the energy blaster are the beam laser (velocity 300) and gatling gun (16).
Currently the blaster's velocity is 8.5/8, proposed 17/16, +50% would be 13/12. For comparison:
missile velocities range from 3 (mine) to 18 (piercer)
cannon velocities range from 10 (thrasher) to 36 (inhibitor)
beam weapon velocities range from 300 (laser) to 1440 (gridfire)
anti-missile velocities range from 150 (bullfrog) to 570 (tier 3)
The proposed values would place the blaster velocity above the average missile but below the average cannon.

Furthermore, the purpose of this proposal is to make it harder to outspeed blasters. The maximum speed of ships is in the same range. E.g. a Falcon equipped with a Bufaer thruster has a maximum speed of 1075.5 per second, i.e. 18 per frame. And there are faster ships and more powerful engines.

Pointedstick:
My assessment wasn't based on comparing it mathematically to alternatives; it was based on a visual impression. Blaster fire just doesn't look as good at double speed IMHO. But it's your PR.

Nescio0:
That's a quick reply! 
Aesthetics is important. Could you elaborate on the "just doesn't look good", though?

Pointedstick:
When the bolts are so much faster, there's more space between them and fewer of them on the screen at once. The speed definitely makes the weapon more useful and makes it feel more powerful, but you lose the "overwhelming weight of fire" property that Blasters currently have when mounted in large numbers.

Nescio0:
To preserve thecurrent "overwhelming" feeling I suppose it's possible to halve reload, damage, energy, and heat as well, but I'm not sure that'll be a good idea.

Nescio0:
To preserve the current "overwhelming" feeling I suppose it's possible to halve reload, damage, energy, and heat as well (to keep it unchanged per second), but I'm not sure that'll be a good idea.

Nescio0:
Update: decreased the blasters reload, firing energy, firing heat, hull damage, and shield damage by a third. Per second it's effectively unchanged. This tweak helps making blasters slightly harder to dodge and largely preserves the "overwhelming weight of fire" property.

Pointedstick:
No corresponding change for the mod blaster?

Nescio0:
Not yet, because 2/3 of 8 is an inconvenient number and because by not changing the modified blaster reload it allows it to be more differentiated from the energy blaster.

Pointedstick:
This is still doubling the velocity. It's still too fast.

Amazinite:
I don’t see any change to the blaster velocity as being necessary. When using blasters, I don’t generally have an issue hitting my target unless they’re at the very edge of my range or moving very fast, two scenarios that make complete sense to be missing in with such a basic weapon. And as mentioned, higher velocity with lower lifetime means less projectiles on screen at once which means there’s less of an aesthetic sense of blaster providing “overwhelming” firepower.

Nescio0:
Yes, *unrelated* changes belong in separate proposals, but it is not unusual to group *related* changes together.

This proposal is summarized by its title: it doubles blaster velocity and halves lifetime, thus keeping range unchanged. The reload, damage, etc. was later reduced by a third because someone mentioned the need to preserve the "overwhelming weight of fire" property; that edit has now been reverted.

As for the desirable velocity, let's first have a look at the velocities of other human weapons:
```
450 electron beam
400 heavy laser
300 beam laser
 30 particle cannon
 24 proton gun
 16 gatling gun
 14 sidewinder
 12 plasma cannon
 11 meteor
 10 javelin
  9 flamethrower
  9 heavy rocket
  6 nuclear missile
  6 torpedo
  5 typhoon
```

Now let's list the various suggested blaster values:
```
25 blaster +200%
24 modified +200%
17 blaster +100%
16 modified +100%
13 blaster +50%
12 modified +50%
11 blaster +25%
10 modified +25%
 8.5 current blaster
 8 current modified
```

Also, it's not impossible for human ships with human engines to have a maximum velocity beyond 15 (900 per second). Personally I think a +100% increase is justified, because that would at least make blasters slightly faster than missiles and ships.
Please take a second look at the numbers and let me know which value you think is best. I'll change it then accordingly.

Pointedstick:
Testing out +25% or +50%, I think the "overwhelming weight of fire" aspect is still preserved, but it isn't at 100%.

The problem that this PR solves is that Quad Blaster Turrets frequently miss with a great deal of their shots against fighter-sized craft, which are their principal targets. That makes the QBT unreasonably unattractive compared to a Laser Turret or a Proton Turret IMHO, and I think increasing the bolt speed a bit (but not too much) succeeds at solving that problem.

Nescio0:
So +50% velocity then?

Nescio0:
Actually I still think blasters need *at least* a 100% velocity increase, but I suppose even a little +25% bump is better than no increase at all.

Also, #3834 displays the projectile velocity in game, on the same line as weapon range, allowing players to easily compare weapons.

tehhowch:
What is wrong with 10.625 and 48? That would keep the range unchanged. 10.625 is also not a bad formatted number (637.5), though yes, it is not as "pretty" as 660.

Nescio0:
Well, I did a `grep` and it appeared all weapon lifetimes and velocities are integers, therefore I rounded 10.625 to 11. Besides, I'm not sure how to visualize 0.375 pixel.
I suppose the blaster's lifetime could be adjusted, e.g. 11*46=506 or 10*51=510.

Nescio0:
Well, I did a `grep` and it appeared all weapon lifetimes and velocities are integers, therefore I rounded 10.625 to 11. Besides, I'm not sure how to visualize 0.375 pixel.
I suppose the blaster's lifetime could be adjusted, e.g. 11×46=506 or 10×51=510.

tehhowch:
Lifetimes are coerced to integers. Velocity can happily be a floating point value (as it is for blasters on `master`).

Pointedstick:
Thanks. Please change the velocities back to being floating point numbers if necessary to preserve the original ranges.

Nescio0:
Honestly, I don't see why a range difference of 8.5×60-11×46=-4 matters that much. Also, #3866 changed meteor range by 9×600-11×490=10 and sidewinder range by 11*600-14*470=20 and nobody considered that problematic.

Anyway, blaster velocity has now been changed to 10.625, per request.

Nescio0:
Honestly, I don't see why a range difference of 8.5×60-11×46=-4 matters that much. Also, #3866 changed meteor range by 9×600-11×490=10 and sidewinder range by 11×600-14×470=20 and nobody considered that problematic.

Anyway, blaster velocity has now been changed to 10.625, per request.

PR #3866, opened 2018-07-28
Improve the playability and tactical flexibility of human missile weapons

Pointedstick:
Here's version 2 of my attempt to improve human missile playability, this time with more moderation and a lot more playtesting for balance.

All human missiles get +20% DPS to improve their combat usefulness compared to gun weapons. They get large increases in per-missile damage, accompanied (mostly) by significant reductions in missile launcher fire rate. Some missiles get a few additional tweaks to refine their battle roles.

The overall effect of these changes is to differentiate the human missiles from one another in terms of their ideal battle roles  in a way that makes the game world feel more realistic, and to make using them and facing them in battle more fun:
- Since missiles how have much larger per-missile payloads, getting a hit with individual missiles makes a difference: missiles feel powerful, not like little pinpricks.
- High missile damage coupled with low fire rate makes missile combat significantly cheaper and slower to drain missile magazines, which reduces the frustration of running out of ammo and needing to buy more missiles all the time.
- Being both faster and less agile, Meteors are now significantly easier to dodge in a small fast ship which makes the start of the game less frustrating for new players.
- Sidewinders are now much better at shooting down small fast ships, but their higher cost-per-DPS makes them not cost-effective to use against large ships. Against poorly-protected large ships, Meteors are a better option, and against well-defended capital ships, Torpedoes are more appropriate. Where Sidewinders shine is as defensive weapons against small incoming fighter-sized craft, much like modern ship-launched anti-air missiles. They can also be mounted to said fighter-sides craft to shoot down other fighters.
- Javelin pods' very high DPS/ton turns them into a real option. A fighter-sized craft with two Javelin pods is actually threatening.

I played through the early game as all three starting ships, and then most of the Free Worlds campaign in various transports, freighters, and light-to-medium warships. It was substantially easier to avoid being frustratingly killed by Meteor-armed pirates when flying a Shuttle or Sparrow. Major battles went more or less the same way. The Star Barge was very slightly harder without an anti-missile turret, but that was an easy fix: install an anti-missile turret! Overall I had _much_ more fun using missile weapons and nothing felt unbalanced.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -650,7 +650,7 @@ outfit "Meteor Missile Launcher"
 		"fire effect" "meteor fire"
 		"die effect" "missile death"
 		"hit effect" "missile hit"
-		"inaccuracy" 10
+		"inaccuracy" 5
 		"velocity" 13
 		"lifetime" 415
 		"reload" 75
@@ -658,7 +658,7 @@ outfit "Meteor Missile Launcher"
 		"firing heat" 20
 		"acceleration" 1
 		"drag" .1
-		"turn" 1.25
+		"turn" 1.3
 		"homing" 3
 		"infrared tracking" .8
 		"shield damage" 220
@@ -823,14 +823,14 @@ outfit "Torpedo Launcher"
 		"hit effect" "torpedo hit"
 		"die effect" "tiny explosion"
 		"inaccuracy" 5
-		"velocity" 5
+		"velocity" 5.5
 		"lifetime" 720
 		"reload" 150
 		"firing energy" 2
 		"firing heat" 45
 		"acceleration" .15
 		"drag" .025
-		"turn" 1.3
+		"turn" 1.5
 		"homing" 3
 		"optical tracking" .8
 		"shield damage" 650
@@ -885,14 +885,14 @@ outfit "Typhoon Launcher"
 		"hit effect" "torpedo hit"
 		"die effect" "tiny explosion"
 		"inaccuracy" 5
-		"velocity" 4
-		"lifetime" 750
+		"velocity" 4.5
+		"lifetime" 900
 		"reload" 140
 		"firing energy" 4
 		"firing heat" 50
 		"acceleration" .1
 		"drag" .02
-		"turn" 1.5
+		"turn" 1.8
 		"homing" 4
 		"optical tracking" .9
 		"shield damage" 900
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -824,7 +824,7 @@ outfit "Torpedo Launcher"
 		"die effect" "tiny explosion"
 		"inaccuracy" 5
 		"velocity" 5.5
-		"lifetime" 720
+		"lifetime" 920
 		"reload" 150
 		"firing energy" 2
 		"firing heat" 45
@@ -886,7 +886,7 @@ outfit "Typhoon Launcher"
 		"die effect" "tiny explosion"
 		"inaccuracy" 5
 		"velocity" 4.5
-		"lifetime" 900
+		"lifetime" 1200
 		"reload" 140
 		"firing energy" 4
 		"firing heat" 50
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -651,18 +651,18 @@ outfit "Meteor Missile Launcher"
 		"die effect" "missile death"
 		"hit effect" "missile hit"
 		"inaccuracy" 10
-		"velocity" 9
-		"lifetime" 600
-		"reload" 48
+		"velocity" 13
+		"lifetime" 415
+		"reload" 75
 		"firing energy" 1
 		"firing heat" 20
-		"acceleration" .9
+		"acceleration" 1
 		"drag" .1
-		"turn" 2
+		"turn" 1.25
 		"homing" 3
 		"infrared tracking" .8
-		"shield damage" 130
-		"hull damage" 80
+		"shield damage" 220
+		"hull damage" 150
 		"hit force" 220
 		"missile strength" 4
 	description "The Meteor Missile is the cheapest and simplest homing weapon available in human space. Meteors are quickly destroyed by anti-missile systems, and their infrared tracking systems have trouble homing in on targets with low heat levels. Nevertheless, they can be deadly in large numbers and against poorly-defended targets."
@@ -713,18 +713,18 @@ outfit "Sidewinder Missile Launcher"
 		"die effect" "missile death"
 		"hit effect" "missile hit"
 		"inaccuracy" 4
-		"velocity" 11
-		"lifetime" 600
-		"reload" 40
+		"velocity" 12
+		"lifetime" 550
+		"reload" 60
 		"firing energy" 1
 		"firing heat" 15
-		"acceleration" 1.1
+		"acceleration" 1.2
 		"drag" .1
-		"turn" 3
+		"turn" 5
 		"homing" 4
 		"radar tracking" .9
-		"shield damage" 100
-		"hull damage" 70
+		"shield damage" 165
+		"hull damage"115
 		"hit force" 170
 		"missile strength" 12
 	description "The Sidewinder Missile is designed by Lovelace Labs, which is ironically also the designer of the Anti-Missile Turret. Rumors say that the two engineering teams at Lovelace who make the missiles and the anti-missiles are locked in furious competition with each other: the result has been generation after generation of smarter missiles and then smarter turrets to fend them off."
@@ -777,8 +777,8 @@ outfit "Javelin Pod"
 		"reload" 15
 		"firing energy" .2
 		"firing heat" 12
-		"shield damage" 48
-		"hull damage" 26
+		"shield damage" 55
+		"hull damage" 30
 		"hit force" 50
 		"missile strength" 2
 	description "The Javelin Pod fires a rapid stream of unguided missiles. A Javelin Pod does far more damage than any gun of a comparable size, and has a longer range as well, but once a ship has fired off all its ammunition it must leave combat and find a planet where it can buy more, which makes Javelins less useful in protracted battles. As a result, they are very popular with pirates as a way to quickly disable a small target."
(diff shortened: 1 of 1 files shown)


DingusShingleton:
Ahh, keeping the old missile balance torch aflame. Love it.

Nescio0:
Out of curiosity, how did you settle down on the proposed values (especially things like accelaration and turn)? Also, don't you think you made torpedoes too slow? E.g. a velocity of 4 per frame means a maximum speed of 240 per second, and most large ships have no difficulty with that.

Pointedstick:
I approached this PR from a primary desire to solve the missile *fun* problem. A major reason why missiles aren't fun to use IMHO is because their low damage per missile encourages a spammy firing style that depletes magazines quickly and costs a lot of money. It's just not fun, and it's not better then gun combat by a large enough margin to make people put up with those annoyances anyway.

My idea was to solve this problem by boosting damage per missile by quite a lot, but reduce reload times to only result in a moderate boost to DPS. This has the effect of making each missile more dangerous and also makes missile combat slower to exhaust your ammunition stores and cheaper on a credit-per-damage/DPS basis, without radically altering their DPS values (which would change the feel of the game quite a bit). So that dictated the broad strokes of the damage and reload changes.

The turn and acceleration changes arose from the intention to differentiate the missiles better into distinct battle roles. I've spent the past 6 months researching how missile combat evolved during the 20th century and is currently conducted today on aircraft and surface ships. The analogues to Endless Sky aren't perfect, but the concepts map well enough. Here's the result of my attempt to make the human missiles better conform to what I see as realistic constraints and use cases:

- Sidewinder: essentially an anti-air missile suitable for being fired from either a small ship against other small ships, or by a large ship as a defense against small ships. Can be used as a light anti-capital-ship missile in a pinch when spammed, just like modern AAMs
- Meteor: cheap long range bombardment missile best used against large and poorly-protected targets, much like mid 20th-century anti-ship missiles. Most useful when fired from a small maneuverable ship or when spammed
- Heavy Rocket: swarm and small capital ship killer used to turn interceptor-sized craft into modern-day fighter-bombers/strike craft
- Torpedo: Large capital ship killer fired from other capital ships
- Javelin: Basically a fighter's short-range cannon. Should pair well as secondary armament for fighter-sized ships able to mount a Meteor or Sidewinder launcher as their primary armament.

In terms of how I arrived at the specific numbers beyond those broad goals, it was just a lot of playtesting and tweaking. I played a lot using ships that I equipped with various missile weapons, and also played against all the existing AI ships that have missile weapons.

Regarding the torpedo speed change, I encourage you to give it a shot (pun intended :smiley:). I'm willing to continue tweaking, but in my testing, the slow speed wasn't really a huge deal in practice for either 1-on-1 fights or in huge fleet battles. That said I could consider boosting the speed a bit and reducing the turn value. I really want to avoid making Torpedoes any good at all against very small maneuverable ships, which cuts into the Sidewinder's battle role. Capital ships should defend themselves against dangerous missile-armed strike craft using Sidewinders, defensive turrets, or their own carried interceptors.

Nescio0:
Thank you for your elaborate reply! I agree with you on the problem (missiles are ineffective and unattractive) and also largely on their supposed usage (sidewinder ship-vs-fighter, torpedo ship-vs-ship). 

Like you, I've been playing around with values in the past months; currently my (unpublished) plug-in includes the following changes:
* doubled hull and shield damage of all missiles (human and alien)
* kept hit force and special damages unchanged
* improved the velocity of some missiles, including the torpedo, because they're currently slower than many ships
* halved launcher capacity to keep total damage unchanged (some alien excepted)
* changed empty launcher mass to keep mass when full unchanged
* +50% reload time, thus increasing damage per second by a third (2/1.5=4/3)
* extra ammunition storage is kept unchanged, thus making them twice as effective (because they now offer a full reload, i.e. double the damage)
* anti-missile turrets are kept unchanged, thus making them more effective, because of the lower fire rate, and more important, because of the higher missile damage
Feel free to ignore these points; I don't want to hijack your proposal.

However, I'm probably biased, because:
* I don't use anything smaller than a shield beetle
* all my ships have two Wanderer anti-missile turrets each
* all my ships are equipped with Hai engines; as a consequence their maximum speed is typically above 420 per second (i.e. 7 per frame)

Pointedstick:
I previously submitted a set of PRs pretty similar to what you describe, but they were rejected because the changes were too extreme and fundamentally changed the way combat worked, and would have required far-ranging changes to the default load out of most ships and variants the the fleet compositions of scripted battles. My hope with this PR is to address playability issues of human missiles without changing the game's combat into something totally different.

Pointedstick:
You're right, the Torpedo changes were a bit much. I've toned them down a little bit after further playtesting.

Nescio0:
Effectively you reduced the torpedo velocity by -0.5; you might as well have kept it unchanged; or improved it.
Also, I noticed you increased the Typhoon range by more than a third (-10% velocity, +50% lifetime). Is that intentional?
Furthermore, alien missile weapons seem to be based on their human counterparts, so perhaps you ought to have a look at those as well.

Pointedstick:
Yes, that was intentional. I've now done the same thing with the torpedo too, because their short ranges were impairing their ability to be the kind of standoff anti-capital-ship weapons that they're supposed to be.

Nescio0:
Perhaps you could list how your tweaks effectively work out? E.g. missile X now has Y range instead of Z; damage per second changed from U to V.

Pointedstick:
You can get that information pretty easily by looking at the diff. If I duplicate it here, it'll be a pain in the next to update it in response to tweaks and feedback.

Only torpedoes have increased range right now, and the changes to Javelins and rockets are pretty small. It's only Meteors, Sidewinders, and Torpedoes that changed in overall behavior at all, which is in the service of differentiating them to make them each support a niche better.

Nescio0:
* heavy rocket and javelin tweaks: no objections
* meteor and sidewinder: the turn rate seem rather radical: sidewinder from 3 to 5 and meteor from 2 to 1.3 (below torpedo's 1.5 and typhoon's 1.8)
* torpedo and typhoon: your proposal changes torpedo range from 6*600=3600 to 5.5*920=5060 and typhoon range from 5*600=3000 to 4.5*1200=5400 (for comparison, the meteor's is 5395). To me this seems to be actually *less* range differentiation.
Furthermore, torpedoes are already slow and can be outrun by most ships; by increasing their lifetime and range you significantly increase the chance they hit an asteroid or are taken out by anti-missile from a different ship long before they reach their target.

I'd suggest the following:
* keep lifetime unchanged (all missiles)
* increase torpedo velocity to 7 (in between the nuke's 6 and rocket's 8), thus range to 4200 (in between typhoon's 3000 and meteor's 5400)
* reduce typhoon turn from 1.8 to 1.2 (torpedo's unchanged at 1.5)

Pointedstick:
One thing to consider with this change is that small fighter-sized craft become more vulnerable to Sidewinder missile fire. This isn't a big deal for the player in the beginning of the game because most pirates are armed with Meteors, but it does make the Free Worlds campaign more challenging in a solo small ship, and it does mean that Free Worlds fighters have less survivability (though not enough to alter the major battles, as mentioned earlier). It might be worth investigating giving some fighter-sized craft Small Radar Jammers by default. The Finch for example has enough outfit space. But I didn't want to overload this PR by making that change here.

Nescio0:
Interesting. I see you've completely replaced the previous set of changes with a new set of tweaks. At least the torpedo range increase is reverted, thanks. I'll have a critical look later.

Pointedstick:
Nah, it's mostly the same, but with the DPS increases harmonized to +20% (previously they were all close to this, but I decided to make it consistent for simplicity's sake).

endless-sky:
I'm worried that in the process of increasing missile damage to make them more attractive to the player as a weapon choice, you're also making missiles considerably more annoying when enemy ships use them. In particular, a sufficiently fast ship used to be able to just outrun meteors; this will make that harder. And, this change makes it so that a heavy rocket can one-shot disable a shuttle if that shuttle has taken any other damage whatsoever.

DingusShingleton:
Ironic that in the other thread you were poking fun at people who tried to outrun missiles, but here you are saying it's the preferred strategy XD

These changes would seem to incentivize actual dodging since, as pointedstick pointed out, meteors are faster but less able to juke if the player actual gets fancy with the footwork.

Pointedstick:
> I'm worried that in the process of increasing missile damage to make them more attractive to the player as a weapon choice, you're also making missiles considerably more annoying when enemy ships use them.

I specifically tried to address this and playtested the heck out of it. Meteors are now *easier* to dodge in a small ship, which I believe was most of the source of frustration for new players. Sidewinders are only going to be used against you in quantity (by the navy) once you're already expected to have a reasonable ship that can shoot them down or jam them. Javelins are still largely useless against a ship with even one Heavy Anti-Missile turret, and still dangerous against other ships. The torpedo is already crap against small ships, which doesn't change. It does become more dangerous against large ships, but I think this is important. Also, if you're flying a large ship, you're good enough at the game that the tactical challenge of figuring out how to handle enemy missiles is part of the fun.


> In particular, a sufficiently fast ship used to be able to just outrun meteors; this will make that harder.

The reason I did this was to make meteors more likely to miss in the first place, and less likely to be able to hit you "on the rebound"; higher velocity coupled with lower turn means that they're considerably less accurate against a small ship than they used to be. But if you'd like, I can revert that part (though I think it would require further reductions in the missile's `turn` value to compensate)


> And, this change makes it so that a heavy rocket can one-shot disable a shuttle if that shuttle has taken any other damage whatsoever.

To be fair, a Shuttle was already dead meat against anything firing Heavy Rockets. But if you're really worried about this, perhaps I could make the Heavy Rocket slower than it currently is, and increase its missile strength accordingly?

Nescio0:
And I've just separated some ideas into a plug-in for those who want to play around with more powerful missiles; it's available at https://github.com/Nescio0/endless-sky-missiles.git

Pointedstick:
I just got finished playing through the Free Worlds campaign again with the changes in this PR and It was hugely fun. I actually used missiles rather than doing all fixed guns and turrets. None of the scripted events had their outcomes meaningfully changed.

DingusShingleton:
I've been using the plug-in too and it's noticeably better. My only thought is if anti missile turrets need to be looked at? As is, anti missile strength and firing rate are balanced for the current missile landscape, but I'm not sure if these changes would affect that.

PR #3867, opened 2018-07-30
Corrected beam laser description

Nescio0:
Changed the beam laser description, because:
* 300 is not a "slightly" shorter range than 510 
* 60*(1+1.3)=138 is not "a comparable amount of damage" to 5*(9.6+6)=78.

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -161,7 +161,7 @@ outfit "Beam Laser"
 		"firing heat" 1.2
 		"shield damage" 1
 		"hull damage" 1.3
-	description "In the early part of the space era, the settlements in the region of space known as the Deep developed in relative isolation from the rest of human space. One result of that isolation is that their weapons technology mostly uses beam weapons, instead of the energy projectile weapons that are more common elsewhere. A Beam Laser has a slightly shorter range than an Energy Blaster, but is also much more energy-efficient and does a comparable amount of damage."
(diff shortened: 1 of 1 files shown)


Bladewood:
It isn't actually talking about raw output: it might be referring to damage / ton, or disable time.

And, you should compare the shield and hull numbers separately, because simple addition of shield and hull numbers runs into conflict with the fact that not all ships have the same shield:hull ratio.

In general, most human ships tend to have much stronger shields than hull, so the energy blaster actually performs a bit better (now).
You can do the math: 40 tons of beam lasers, or 5 lasers, has a damage output of 300 shield 390 hull damage / sec. 40 tons of energy blasters, or 8 blasters, deals 384 shield 240 hull damage / sec.
This doesn't look like an advantage at all until you actually calculate the time to disable a ship*: it takes 22.2 seconds for the laser array to disable a gunboat, versus 19.8 with the blasters; likewise, it takes 19.3 seconds with lasers, versus 19.6 with blasters. It's a not-insignificant difference in the former case, and it's close enough to at least be "comparable damage" in the latter case.


* assuming all ships are disabled at 20% hull, which is inaccurate for smaller ships, but it doesn't really make that much of a difference in this case.

Nescio0:
The description doesn't say "40 tons of", it currently compares "a beam laser" to "an energy blaster", i.e. singular, thus a 1:1 comparison. 
Perhaps the "comparable damage" part might be open to debate; this proposal changes it to "significantly more hull damage", though, which I believe is certainly true and ought to be uncontroversial. If you have a better suggestion, feel free to propose a different string.

Pointedstick:
Let's not get caught up in technical details. The descriptions aren't meant to be a perfect reflection of the stats, but they *are* meant to avoid obvious factual errors.

DingusShingleton:
"A Beam Laser has a shorter range than an Energy Blaster, but is also much more energy-efficient and does a comparable amount of damage, considering it's more than half again as large."

Nescio0:
> If anything, I'd prefer to just remove that last sentence entirely

Done.

PR #3869, opened 2018-07-31
Improve some missile launch and explosion sounds

Pointedstick:
I am bad at Github and messed up https://github.com/endless-sky/endless-sky/pull/3274. This is its new version.

Meteors and sidewinders get less cartooney launch sounds and more impressive-sounding hit explosions, and torpedoes get even more impressive sounding hit explosions.


copyright
```diff
@@ -407,12 +407,29 @@ Copyright: Various
...
 License: CC-BY-SA-3.0
 Comment: Taken from http://soundbible.com/1467-Grenade-Explosion.html
 
-Files: sounds/explosion huge.wav
+Files: sounds/missile?hit.wav
+Copyright: Copyright "Nbs Dark"
+License: public-domain
+Comment: Taken from https://freesound.org/people/Nbs%20Dark/sounds/94185/
+
+Files: sounds/torpedo?hit.wav
```
quyykk, line 420: @Pointedstick Whose copyright is this? I ask because as it turns out not having a Copyright field is a violation of the spec.

quyykk: Just saw on that other PR that you don't know. So `Copyright: Public Domain` should be enough I think. Sorry for the unnecessary ping.

PR #3870, opened 2018-07-31
Introduce new fuel attributes

Nescio0:
Currently fuel is used for hyperjumps, flamethrowers, cloaking, and afterburners. Fuel damage also exists but is not used yet. This proposal introduces fuel consumption, fuel generation, hull fuel, and shields fuel, analoguous to energy. 

I've been using it in my experimental plug-in (unpublished) and I figured others might want to try it out as well. Examples of things that will be possible:
* aliens that do not have ramscoop but instead have outfits or ships that directly produce fuel
* fuel cells that consume fuel and generate energy
* repair hull by using fuel


data/tooltips.txt
```diff
@@ -464,5 +485,8 @@ tip "limited turn?"
 tip "solar power?"
 	`This ship is powered entirely by solar energy. If you fly too far from the center of a star system, maneuvering may become difficult because your engines will only have a fraction of the power they require.`
 
+tip "fuel?"
+	`This ship has not enough fuel to generate additional energy.`
```
DingusShingleton: Huh? Fuel doesn't generate energy.

data/tooltips.txt
```diff
@@ -482,5 +482,8 @@ tip "limited turn?"
 tip "solar power?"
 	`This ship is powered entirely by solar energy. If you fly too far from the center of a star system, maneuvering may become difficult because your engines will only have a fraction of the power they require.`
 
+tip "fuel?"
+	`This ship has not enough fuel to generate energy.`
```
Bladewood: The expression `<verb> not` is seldom seen in modern grammar; `does not <verb>` is much more common. So `This ship does not have enough fuel` would be better.
Also, this tip should have an introduction like `"solar power?"`, such as `This ship relies entirely on its fuel for energy.` It really feels out of place without one.

Nescio0: Actually this "fuel?" warning checks only if fuel capacity + generation - consumption > 0, i.e. if it has enough fuel to depart. Anyway, I've rephrased it.

source/AI.cpp
```diff
@@ -2307,7 +2308,9 @@ bool AI::DoCloak(Ship &ship, Command &command)
 		// or 40% farther away before it begins decloaking again.
 		double hysteresis = ship.Commands().Has(Command::CLOAK) ? .4 : 0.;
 		// If cloaking costs nothing, and no one has asked you for help, cloak at will.
-		bool cloakFreely = !attributes.Get("cloaking fuel") && !ship.GetShipToAssist();
+		bool cloakFreely = 0;
```
tehhowch: Why did you change the structure of the original check? (And why initialize a `bool` with an `int` when the literals `true` and `false` exist and their use ensures your intent is explicitly clear?)

`bool cloakFreely = fuelCost <= 0. && !ship.GetShipToAssist()`

source/ShipInfoDisplay.cpp
```diff
@@ -264,10 +264,12 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 		60. * (attributes.Get("energy generation")
 			+ attributes.Get("solar collection")
 			- attributes.Get("energy consumption")
+			+ attributes.Get("fuel energy")
```
tehhowch: This should be with other generations, not in between two consumptions

source/Ship.cpp
```diff
@@ -1598,9 +1612,18 @@ void Ship::DoGeneration()
 		double coolingEfficiency = CoolingEfficiency();
 		energy += attributes.Get("energy generation") - attributes.Get("energy consumption");
 		energy -= ionization;
+		fuel += attributes.Get("fuel generation");
 		heat += attributes.Get("heat generation");
 		heat -= coolingEfficiency * attributes.Get("cooling");
 		
+		// Apply fuel to energy and heat conversion
```
tehhowch: `// Convert fuel into energy and heat only when the required amount is available.`

source/Ship.cpp
```diff
@@ -739,6 +747,10 @@ string Ship::FlightCheck() const
 		return "limited turn?";
 	if(energy - .8 * solar < .2 * (turnEnergy + thrustEnergy))
 		return "solar power?";
+	if(fuel < 0.)
+		return "fuel?";
+	if(!hyperDrive && !jumpDrive && !canBeCarried)
```
tehhowch: This appears to be a duplicate check (to the one immediately below it, which was added in #3635 and adjusted in #3843 ).

Nescio0:
Nice label! What is it you're waiting for exactly?

Nescio0:
Like the following? 
[Fuel Test.txt](https://github.com/endless-sky/endless-sky/files/2799397/Fuel.Test.txt)

screenshot from 2019-01-26 20-11-38
screenshot from 2019-01-26 20-05-19
screenshot from 2019-01-26 20-05-34

PR #3883, opened 2018-08-03
Add categories to endless sky appdata.

rrigby:
Following https://standards.freedesktop.org/menu-spec/latest/apas02.html

Did this mostly for better category filtering support for the flatpak appstream data.

### DIFF
--- a/endless-sky.appdata.xml
+++ b/endless-sky.appdata.xml
@@ -21,8 +21,6 @@
   </description>
   <categories>
     <category>Game</category>
-    <category>ActionGame</category>
-    <category>RolePlaying</category>
     <category>Simulation</category>
   </categories>
   <url type="bugtracker">https://github.com/endless-sky/endless-sky/issues</url>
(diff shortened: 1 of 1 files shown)


endless-sky.appdata.xml
```diff
@@ -19,6 +19,11 @@
    Endless Sky is a sandbox-style space exploration game similar to Elite, Escape Velocity, or Star Control. You start out as the captain of a tiny space ship and can choose what to do from there. The game includes a major plot line and many minor missions, but you can choose whether you want to play through the plot or strike out on your own as a merchant or bounty hunter or explorer.
   </p>
   </description>
+  <categories>
+    <category>Game</category>
+    <category>ActionGame</category>
+    <category>AdventureGame</category>
```
endless-sky: I'm not sure ES fits what most people would think of an an "adventure game." Maybe "role-playing game" would make more sense. Or "simulation game."

rrigby: Simulation and RolePlaying are both available categories in the spec.

I can replace AdventureGame with those.

PR #3889, opened 2018-08-10
tweaked javelin hull damage

Nescio0:
#3845 tweaked javelin reload time. This patch increases the javelin's hull damage by about a third, reduces hit force by 4%, and keeps the shield damage and other stats unchanged. This makes the javelin more useful for disabling or blowing up enemy ships. Although still not great, it becomes a somewhat more interesting choice.


Lineth:
I'm cool with this. My reaction to being shot at with javelins in a crappy early game human ship should definitely _not_ be a sigh of relief.

PR #3893, opened 2018-08-12
introduce Heliarch scanning module

Nescio0:
The Heliarchs have a policing function in Coalition space, therefore it makes sense they would be able to scan ships. Rather than having separate cargo, outfit, and tactical scanners, a single system with equal scan powers would be more appropiate. It does not have nor need asteroid or atmospheric scan.


Nescio0:
Free outfit space:
  "Heliarch Interdictor": from 13 to 6
  "Heliarch Interdictor (Bombardment)": from 5 to 9
  "Heliarch Neutralizer": from 23 to 16
  "Heliarch Neutralizer (Missile)": from 61 to 54
  "Heliarch Neutralizer (Nimble)": from 14 to 7
  "Heliarch Neutralizer (Scrappy)": from 0 to 4
  "Heliarch Punisher": from 19 to 12
  "Heliarch Punisher (Interdicting)": from 14 to 7
  "Heliarch Punisher (Missile)": from 3 to 4
  "Heliarch Punisher (Scrappy)": from 0 to 0

Nescio0:
Free outfit space:
"Heliarch Interdictor": from 13 to 6
"Heliarch Interdictor (Bombardment)": from 5 to 9
"Heliarch Neutralizer": from 23 to 16
"Heliarch Neutralizer (Missile)": from 61 to 54
"Heliarch Neutralizer (Nimble)": from 14 to 7
"Heliarch Neutralizer (Scrappy)": from 0 to 4
"Heliarch Punisher": from 19 to 12
"Heliarch Punisher (Interdicting)": from 14 to 7
"Heliarch Punisher (Missile)": from 3 to 11
"Heliarch Punisher (Scrappy)": from 0 to 0

Pointedstick:
Obviously this isn't merge-ready without art, but +1 conceptually. I like the idea of alien scanners with different characteristics from human scanners.

Also, shouldn't some Heliarch fleets gain the `surveillance` personality so they'll actually use these fancy scanners?

data/coalition.txt
```diff
@@ -919,4 +920,4 @@ phrase "hostile heliarch"
 		"not die"
 		"not be extinct"
 	word
-		"!"
```
Pointedstick: Unrelated change.

Nescio0: That's weird, I didn't touch the last line; I have no idea why github makes it show up.

Nescio0:
> Obviously this isn't merge-ready without art

It works with the placeholder thumbnail and the only way to acquire it is by boarding Heliarch ships; besides, other outfits (e.g. Pug gridfire turret, tier 3 anti-missile) don't have unique thumbnails either; so I suppose this is ready to be merged and art can follow later. But if someone can create an appropiate image now and include it here, great!

> how about following the general theme and make them half as good but take up only three tons?

Done.

> Also, shouldn't some Heliarch fleets gain the surveillance personality so they'll actually use these fancy scanners?

No, that's unnecessary. Ships with scanners will automatically scan nearby ships, regardless of their personality (e.g. Navy, Oathkeeper); I don't know if `surveillance` actually does anything.

Nescio0:
> Obviously this isn't merge-ready without art

It works with the placeholder thumbnail and the only way to acquire it is by boarding Heliarch ships; besides, other outfits (e.g. Pug gridfire turret, tier 3 anti-missile) don't have unique thumbnails either; so I suppose this is ready to be merged and art can follow later. But if someone can create an appropiate image now and include it here, great!

> how about following the general theme and make them half as good but take up only three tons?

Done. Neutralizers now have three, Interdictors two, Punishers one.

> Also, shouldn't some Heliarch fleets gain the surveillance personality so they'll actually use these fancy scanners?

No, that's unnecessary. Ships with scanners will automatically scan nearby ships, regardless of their personality (e.g. Navy, Oathkeeper); I don't know if `surveillance` actually does anything.

Bladewood:
> No, that's unnecessary. Ships with scanners will automatically scan nearby ships, regardless of their personality (e.g. Navy, Oathkeeper); I don't know if `surveillance` actually does anything.

Pretty sure that's wrong. Surveillance personality calls DoSurveillance.
https://github.com/endless-sky/endless-sky/blob/eefe75fb7a527037260927399cab3f847de91182/source/AI.cpp#L648-L652

From [the wiki:](https://github.com/endless-sky/endless-sky/wiki/ShipPersonalities)
> `surveillance`: scans random ships and visits random planets in system.

Amazinite:
>Also, shouldn't some Heliarch fleets gain the surveillance personality so they'll actually use these fancy scanners?

>On the other hand, it does control whether AI ships are willing to refuel/repair you.

Heliarch ships are a player's only source of fuel when they first enter Coalition space, so Heliarch ships shouldn't have the surveillance personality.

Nescio0:
Changed it again:
* Neutralizer: 2 scanning modules, thus range at 707
* Interdictor: 3 scanning modules, thus range at 866
* Punisher: 4 scanning modules, thus range at 1000

Brick63:
To breathe some life back into this i am currently working on a a sprite that could work for this outfit @Nescio0

data/coalition outfits.txt
```diff
@@ -143,6 +143,21 @@ outfit "Cooling Module"
 	"cooling energy" 1.4
 	description "This cooling module consumes energy. To conserve power, it does not ramp up to full strength unless a ship is very close to overheating."
 
+outfit "Scanning Module"
+	category "Systems"
+	licenses
+		Heliarch
+	cost 215000
```
Amazinite: Given that this is a three-in-one scanner, I'd up this cost to about 400k instead of nearly being the same cost as the same power's worth in human scanners (which is roughly 160k; the about 400k cost takes into account the fact that this scanner is about 2.6 times more compact).

Brick63:
scanning module 2x

Presenting the possible sprite for this outfit

data/coalition outfits.txt
Nescio0: Actually I don't remember how I decided on its price three months ago. Looking at it again, a single scanning module (mass 3) is about as effective as three cargo scanners, one outfit scanner, and one tactical scanner (mass 3*1+2+4=9), which cost 3*8+24+144=192 thousand. On the other hand, it is not for sale, you can only acquire it by plundering Heliarch ships, so the cost does not really matter. Coalition outfits are typically cheap, though.
Anyway, I've now raised its cost above 0.4 M, hopefully that addresses your concern.

Nescio0:
Actually that proposed sprite looks quite large; the idea is that the scanner is so small that all Heliarch ships have installed multiple of them. And because it's a module I'd expect it to consist of small parts. Perhaps you could design something in a style more similar to e.g. the small shield module?

Amazinite:
Here's an image for some side-by-side referencing with existing outfits.
coalition-scanner

Amazinite:
@Nescio0 Are there any specific changes you would suggest to the scanner module as is? What exactly from the small shield module would you model the scanner after?

Zitchas:
Personally, I think the scanner looks good. It seems fairly similar in style. 

One thing I notice in doing comparisons, though, is that almost all the Heliarch stuff that is circular is actually octagonal or hexagonal, and have fairly rough edges. So adjusting the smooth metal circular part to be either rougher (like the edges of the shield generators) or octagonal (like the engines or generators) would be more than enough to have it blend in with the current style. 

I'm not convinced it actually needs those changes, but if people want it to be "more like other Heliarch outfits," one of those two changes would be what I'd suggest. After all, it is a scanner and not an engine or generator, so it makes sense that it would have different shapes than other things.

data/coalition.txt
tehhowch: The file in `master` has no newline at the end of the file. Nescio's version has one. Moving forward, those who use editors that support EditorConfig rules will ensure that the EOF newline is present.

Nescio0:
The colour scheme of the proposed sprite is fine, but in detail it doesn't feel right.
To me it looks like a radio tower, whereas this module is supposed to be small. Also, when looking at it again, the antenna resembles a cannon and the dish looks flat. 
The shield modules consists of a couple of emitters directly on top of a foot consisting of several blocks with openings, emphasizing its modular aspect.

> Are there any specific changes you would suggest to the scanner module as is?
Perhaps the following?
* add more blocks, colour, connections, openings to the scanner's foot
* cut out the middle to make the sprite less tall
* add more curvature to the dish (cf. solar collectors and other scanners)
* replace the thick central antenna beam (which blocks the signal) with e.g. three diagonal thin beams to support the receiver; perhaps have a look at e.g. https://en.wikipedia.org/wiki/Radio_telescope#Gallery_of_big_dishes for inspiration

Zitchas:
Well, there is small, as in short, and there is small, as in it has a small footprint in terms of space used. Just because it doesn't take up much hull space doesn't mean it is short or light. 

Putting scanner equipment (radar, IR cameras, etc.) on posts to get them farther away from the vehicle or structure they are mounted on is long standing practice for fairly clear and simple reasons: The better the field of view they have, the more useful they are, and the less interference they get from their host. This makes even more sense in space than it does on land, as even moving the device a meter farther out can have a massive impact on field of view, especially in regards to the opposite side of the ship. (It could never see directly on the opposite side of the ship, but the farther out it is the smaller that blind cone will be. For ES, I assume that all ships rotate in space enough to cover those blind spots, or simply aim their scanner at whatever they are interested in scanning. In short, I don't think height should be changed.

Amazinite:
Brick roughed up the dish per @Zitchas's comment, but he isn't on board with the other suggestions.

Pointedstick:
I think the sprite is fine. Maybe a *little* bit large, but that's about it.

data/coalition outfits.txt
```diff
@@ -143,6 +143,21 @@ outfit "Cooling Module"
 	"cooling energy" 1.4
 	description "This cooling module consumes energy. To conserve power, it does not ramp up to full strength unless a ship is very close to overheating."
 
+outfit "Scanning Module"
+	category "Systems"
+	licenses
+		Heliarch
+	cost 453600
+	thumbnail "outfit/unknown"
```
tehhowch: Needs art, the art assets (.xcf), and associated copyright file updates

Nescio0:
Unfortunately I don't have any corresponding art assets.

Amazinite:
@Nescio0 I've still been keeping my eye on this PR and asking around for people to create images. So far I think this is the best suggested scanner image from @Kryes-Omega. I've asked for a few slight changes, but what are your thoughts on it now?
image

Amazinite:
Here's the updated version after my feedback.
image

Pointedstick:
That's fantastic!

Nescio0:
Looks great!

Amazinite:
Here's two scaled versions, sized to be as big as the small shield module.
small scanner module 2x
small scanner module

Nescio0:
Copyright? `xcf` file?

Amazinite:
Like I said, they were made by @Kryes-Omega, so just find him in the copyright file and add this to his list.
He can post the assets later.

Nescio0:
That's the problem: Kryes-Omega is not in the copyright file yet, therefore I don't know how he wants to be attributed and under which licence he released the art.

Amazinite:
Sorry, hadn't realized he wasn't in there under his GitHub name. Here you go.
```
Files: images/outfit/scan?module*
Copyright: Zachary Siple
License: CC-BY-SA-4.0
Comment: Derived from works by Michael Zahniser (under the same license).
```

Nescio0:
Done now. Anything else?

copyright
```diff
@@ -264,7 +264,9 @@ Comment: Screenshots of environments created with assets purchased by Tymoteusz
  Kapuściński from the Unity Asset Store. Post-processing applied by Michael
  Zahniser to make the images look less artificial.
  
-Files: images/outfit/tactical?scanner*
+Files: 
+  images/outfit/scan?module*
+  images/outfit/tactical?scanner*
 Copyright: Zachary Siple
 License: CC-BY-SA-4.0
```
Amazinite: The scan module should be its own entry with the proper comment that I mentioned, as the scan module was derived from the small reactor module, while the tactical scanner was done from scratch.

Amazinite:
I'll wait for a high DPI PR to be up first, but I saw we let @Kryes-Omega open that along with a link to the assets.

data/coalition outfits.txt
```diff
@@ -143,6 +143,21 @@ outfit "Cooling Module"
...
+	cost 453600
+	thumbnail "outfit/scan module"
+	"mass" 3
+	"outfit space" -3
+	"outfit scan power" 25
+	"outfit scan speed" 1
+	"cargo scan power" 25
+	"cargo scan speed" 1
+	"tactical scan power" 25
+	description "This system allows the Heliarchs to scan nearby ships."
```
Amazinite: This description could definitely be better. How about something like `All Heliarch ships come standard with an array of scanning modules used to ensure that Coalition citizens are not transporting any illegal goods in their ships, such as weapons and ammunition.`

Nescio0: The description ought to be about the scanning module, not about Heliarch ships.

Amazinite: The tactical scanner's description: `"A tactical scanner allows your ship to determine the fuel, energy, heat, and crew complement of nearby targets. For centuries tactical scanners were only employed by the Navy, but recently Kraz Cybernetics has begun manufacturing them for use by militia ships."`

So talking about the ships that the scanner is on is alright, as long as the abilities of the scanner are also mentioned. And to expand my description to also hit on the tactical scan aspect: `All Heliarch ships come standard with an array of scanning modules used to ensure that Coalition citizens are not transporting any illegal goods in their ships, such as weapons and ammunition, as well as find out the status of the ship being targeted.`

Nescio0: Mentioning ships is fine, but the emphasis ought to be on the outfit. Compare *TC [function]. TC user, new TC manufacturer.* with: *Heliarch ships' default equipment, Heliarch reason, Heliarch targetting.*

Amazinite: Whatever. A better description can be made when it's available for purchase or simply in outfitters.

PR #3894, opened 2018-08-12
tweaked Navy cargo space

Nescio0:
Larger Navy ships should have more cargo space.
* Carrier: 100 (unchanged)
* Cruiser: from 60 to 80
* Frigate: from 35 to 60
* Gunboat: from 50 to 40
* Rainmaker: from 25 to 20
* Lance, Combat Drone, Surveillance Drone: 0 (unchanged)

PR #3896, opened 2018-08-13
adjusted heavy rocket lifetime

Nescio0:
#3866 increased the velocity of the meteor and sidewinder and reduced lifetime to keep range about unchanged. The heavy rocket velocity was also increased, but it was omitted to decrease the lifetime. This proposal adjusts it.
Old: 8×100=800
Current: 9×100=900
Proposed: 9×90=810

(Also rounded meteor and sidewinder lifetime from 490 and 470 respectively to 480, i.e. 8 seconds; the effect on range is hardly noticeable. Meteors and sidewinders are similar weapons and used to have identical lifetime (600). Likewise, torpedoes and typhoons still have the same lifetime (unchanged).)

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -652,7 +652,7 @@ outfit "Meteor Missile Launcher"
 		"hit effect" "missile hit"
 		"inaccuracy" 5
 		"velocity" 11
-		"lifetime" 490
+		"lifetime" 480
 		"reload" 80
 		"firing energy" 1
 		"firing heat" 20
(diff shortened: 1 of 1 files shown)


Amazinite:
>(Also rounded meteor and sidewinder lifetime from 490 and 470 respectively to 480, i.e. 8 seconds; the effect on range is hardly noticeable.)

>["it looks nice" should not be an argument.](https://github.com/endless-sky/endless-sky/pull/3894#issuecomment-412349244)

Pointedstick:
Giving it a (slightly) slightly longer range was an intended change, FWIW.

PR #3905, opened 2018-08-21
moved variants from mission to ship files

Nescio0:
Human ship variants are typically listed together in the `variants.txt` file; non-human ship variants are listed directly after the default ship in the same ship file. This is a sensible approach that helps preventing variants from being overlooked. 
However, the `wanderers start.txt` file defined two variants (Deep River (Jump, Unarmed) and Carrier (Alpha)), even though other campaign specific variants (e.g. Carrier (Jump), Deep River 0, Shield Beetle (Jump)) are defined elsewhere. This patch moves those two exceptions to the appropiate files.

### DIFF
--- a/data/variants.txt
+++ b/data/variants.txt
@@ -411,6 +411,39 @@ ship "Bulk Freighter" "Bulk Freighter (Proton)"
 
 
 
+ship "Carrier" "Carrier (Alpha)"
+	outfits
+		"Ion Cannon" 2
+		"Hai Tracker Pod" 4
+		"Hai Tracker" 224
+		"Korath Disruptor" 2
+		"Korath Slicer Turret" 2
+		
+		"Triple Plasma Core"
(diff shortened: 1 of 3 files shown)


endless-sky:
The difference is that those two ships only appear in one mission, which makes them similar to the Asteroid or the Timer Ship. The others you named are either part of generic fleets or are used in multiple missions.

Nescio0:
The difference is the Asteroid and the Timer Ship are not variants.
The "Dreadnought (Jump)" variant is only used once (mission "FW Pug 4: Escorts"); nevertheless, it's defined in `variants.txt`. Likewise, the "Korath World-Ship A (Jump)", "Korath World-Ship B (Jump)", and "Korath World-Ship C (Jump)" variants are defined in `korath ships.txt`, not in `wanderers middle.txt`.

PR #3906, opened 2018-08-22
Replaced duplicate landscape images with unused images

Amazinite:
This PR replaces a few duplicated landscape images with some of the images that are currently not in use. I tried to swap images out with similar looking images as much as possible. In some cases, like replacing the planets with forest landscapes, the newly used image only matches the old image in color as to not conflict with the planet image. Landscape images should still reflect what the planet description says, assuming it says anything at all about the landscape.

With this PR, the only unused non-space non-station images are cities 15, 16, and 17 as well as industrial0. 55 planets still share landscapes among 27 images. 

Unused space and station images still include space 0, 2, 3, and 5 and station 3 and 4, as I didn't look at the space station images with this PR. 30 stations still share landscapes among 10 images.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -24252,7 +24252,7 @@ planet "Brass Second"
 
 planet "Bright Echo"
 	attributes saryd tourism urban
-	landscape land/mountain14-sfiera
+	landscape land/dmottl1
 	description `This is the first world that the Saryds colonized outside of their own solar system. Although somewhat colder than their native planet, it developed into a thriving metropolitan world. The original settlers planned out in exacting detail how every part of the planet's land mass should be used, in order to satisfy their aesthetic ideal of blending residential areas with parks and farms. Large areas have also been set aside as undeveloped wilderness.`
 	spaceport `The spaceport is in what, by Saryd standards, is a major city, with scattered skyscrapers surrounded by forests and grassy park land along with smaller dwellings and occasional factories or power plants. The city is built on either side of one of this continent's largest rivers, a few kilometers downstream from a dam and hydroelectric plant. Here, the river cuts through a deep gorge, and the graceful bridges that cross the gorge are often shrouded in mist.`
 	shipyard "Coalition Basics"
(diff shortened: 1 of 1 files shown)


data/map.txt
```diff
@@ -24252,7 +24252,7 @@ planet "Brass Second"
 
 planet "Bright Echo"
 	attributes saryd tourism urban
-	landscape land/dmottl1
+	landscape land/mountain14-sfiera
```
endless-sky: Given that the Saryd settlements are park-like, with buildings nestled among trees rather than filling large clear-cut areas, this particular image doesn't work very well for them.

data/map.txt
```diff
@@ -25396,13 +25396,13 @@ planet "Karek Fornati"
 
 planet "Kasichara Fet"
 	attributes korath
-	landscape land/sky8
+	landscape land/mountain25-spfld
```
endless-sky: You used this image twice (Firelode, above).

data/map.txt
```diff
@@ -25960,7 +25960,7 @@ planet "New Kansas"
 
 planet "New Portland"
 	attributes "dirt belt" farming
-	landscape land/sky3
+	landscape land/mountain19-harro
```
endless-sky: This replacement image doesn't have a reddish-looking sun, which is a key part of the planet's description.

data/map.txt
```diff
@@ -26879,7 +26879,7 @@ planet Thunder
 
 planet "Tik Klai"
 	attributes wanderer research
-	landscape land/badlands2
+	landscape land/desert9-sfiera
```
endless-sky: I usually try to avoid having images with obviously human architecture on non-human planets.

data/map.txt
```diff
@@ -27280,7 +27280,7 @@ planet "Wyvern Station"
 
 planet Zenith
 	attributes pirate "north pirate"
-	landscape land/water4
+	landscape land/sea14-sfiera
```
endless-sky: This image doesn't really match the description of a world with almost no sunlight and small hidden settlements.

Amazinite:
I've reverted the questionable landscapes for now. Will see if I can use the now-unused images to replace other duplicates instead of the ones I did before.

PR #3909, opened 2018-08-26
Fix disabled ship state

ksrb:
When a ship is disabled and recovered by boarding it will become re-disabled when the player lands on a planet without repair capabilities and reloads game.

To duplicate:
1. Go out into a system with a de-populated planet with a escort
2. Set a escort's hull to the calculated minimum hull value up to 8 digits
3. Recover escort by boarding
4. Land to save game
5. Reload
6. Escort should be disabled again

I'm pretty sure that using the `sqrt` function is creating trailing decimals that are causing the `MinimumHull` to always exceed the value stored in the save file for a disabled ship.

Using a `Model 8` as a example:
Value in save file: `576.19441`
Return value of call to `MinimumHull`: `max(.15, min(.45, 10. / sqrt(3320))) = 576.19441163...`

**Relevant files**
https://github.com/endless-sky/endless-sky/blob/f63e258e03287f38908e889a659e991fd2ba51f4/source/Ship.cpp#L3001-L3008

https://github.com/endless-sky/endless-sky/blob/f63e258e03287f38908e889a659e991fd2ba51f4/source/DataWriter.cpp#L29-L33

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1559,7 +1559,7 @@ void Ship::DoGeneration()
 	double maxHull = attributes.Get("hull");
 	hull = min(hull, maxHull);
 	
-	isDisabled = isOverheated || hull < MinimumHull() || (!crew && RequiredCrew());
+	isDisabled = isOverheated || (abs(hull - MinimumHull()) >= 1 && hull < MinimumHull()) || (!crew && RequiredCrew());
 	
 	// Whenever not actively scanning, the amount of scan information the ship
 	// has "decays" over time. For a scanner with a speed of 1, one second of
@@ -1922,7 +1922,7 @@ bool Ship::IsDisabled() const
 	
 	double minimumHull = MinimumHull();
 	bool needsCrew = RequiredCrew() != 0;
-	return (hull < minimumHull || (!crew && needsCrew));
+	return ((abs(hull - minimumHull) >= 1 && hull < minimumHull) || (!crew && needsCrew));
 }
 
 


source/Ship.cpp
```diff
@@ -1559,7 +1559,7 @@ void Ship::DoGeneration()
 	double maxHull = attributes.Get("hull");
 	hull = min(hull, maxHull);
 	
-	isDisabled = isOverheated || hull < MinimumHull() || (!crew && RequiredCrew());
+	isDisabled = isOverheated || (abs(hull - MinimumHull()) >= 1 && hull < MinimumHull()) || (!crew && RequiredCrew());
```
ksrb: Compensate for decimal precision errors when calculating ship disabled state by checking if the difference between the hull and minimum hull is at least 1.

source/Ship.cpp
```diff
@@ -1922,7 +1922,7 @@ bool Ship::IsDisabled() const
 	
 	double minimumHull = MinimumHull();
 	bool needsCrew = RequiredCrew() != 0;
-	return (hull < minimumHull || (!crew && needsCrew));
+	return ((abs(hull - minimumHull) >= 1 && hull < minimumHull) || (!crew && needsCrew));
```
ksrb: Not exactly DRY but it works...

tehhowch:
Why not just trunc/floor the minimum hull return value to a precision of 7? Then any place that uses it doesn't need to work about the check.

ksrb:
Oh I hadn't thought of that 😅, kind of new to the code base and the last time I did C was awhile ago, let me give it a shot and see how it works.

ksrb:
Actually looking into truncating floating point values the best way I've found to do it is:
```cpp
double Ship::MinimumHull() const
{
	if(neverDisabled)
		return 0.;
	
	double maximumHull = attributes.Get("hull");
	double minimumHull = maximumHull * max(.15, min(.45, 10. / sqrt(maximumHull)));
	unsigned int uiTemp = (minimumHull*pow(10, 7));
	minimumHull = ((double)uiTemp)/pow(10, 7);
	return minimumHull;
}
```

Which I totally stole from https://stackoverflow.com/a/304013, which doesn't feel very efficient (but works). Is there a better way?

ksrb:
Actually I lied, the code above is actually setting the minimum hull value to something really small... if the above approach is valid anyone mind taking a look?
https://www.onlinegdb.com/edit/BJx9ZwgvQ

tehhowch:
Would you perhaps summarize in the first post what the issue is and what it manifests as? I think you are saying that a mission or player ship which is just barely disabled when the game is saved, will no longer be disabled if the game is loaded from the save file, but I'd like you to confirm that understanding.

ksrb:
Updated description.

Yeah that's pretty much right but it happens to any ship that has been disabled and recovered by boarding if you're in a system without a populated planet (as the ship won't be repaired to full health).

endless-sky:
I think the best solution is just to make it so the disabled cutoff is always a whole number (cbdc163), because whole numbers will always be represented exactly by floating-point values.

Also note that floor() is a much better way to truncate floating point values than converting to int and back.

PR #3922, opened 2018-09-04
Tweak default ship and variant loadouts to reflect missile changes

Pointedstick:
# Changes
This PR changes the default loadout of many ships and variants to reflect the logical conclusions of  the recent missile tweak and improve in-game realism, fun, and tactical opportunities.

In a nutshell, here are the changes:
* Most ships that previously had no missile armament now have some according to their typical default battle roles:
  * Interceptor/fighter: meteor, sidewinder, or javelin launchers
  * Interceptor-killer/anti-fighter defense: sidewinder missile launchers
  * Capital ship bombardment: torpedo launchers
  * Multirole/pirate/anti-merchant: meteor missile launchers
* Nearly all fighters, interceptors, and light warships have a Small Radar Jammer. [In the event that these outfits ever become useful](https://github.com/endless-sky/endless-sky/issues/3912), they should allow these ships to have a modicum of a chance against sidewinder missiles. It seemed logical that countermeasures against anti-fighter missiles would be more or less standard equipment.
* All turreted ships that previously lacked an anti-missile turret got one. Even before my missile changes, it was always odd how some large slow ships like Modified Argosies let you pound them to death with Meteors at long range with no defense whatsoever. No more. Large capital ships generally mount two Heavy Anti-Missile Turrets--one of the ways you can distinguish them from smaller lighter warships.
* Added a number of new variants that are heavily armed with missiles. Check out the `Vanguard (Missile)` variant, which is basically an [arsenal ship](https://en.wikipedia.org/wiki/Arsenal_ship).
* More fleets include heavy missile-armed interceptor and light warship escorts.
* Gave the Hawk 10 more weapon space and reduced its Cargo space by 10. This was the only way I was able to make the Hawk actually excel at anything over a Berserker or a Fury. With 50 tons, they can carry a mix-and-match assortment of Plasma cannons and Torpedo Launchers, making them truly useful bomber craft.

# Testing
* Played through the entire Free Worlds campaign in various ships, some armed with missile weapons, and some armed exclusively with guns and turrets.
  * Early game was no more difficult than it was before with all three ships
  * No major change to the overall balance during the campaign
  * In some cases, short-range battles are a bit slower because of the replacement of guns and gun turrets with missile weaponry. I think this is a good thing overall, since fleet battles in Endless Sky are generally over very quickly.
* Made sure all changed ships and variants could fit their new loadouts, and that their energy and heat profiles still made sense. Made minor tweaks to preserve these where appropriate.

# Feelings
[...]

PR #3926, opened 2018-09-08
Added Wasp variants to Small Core Pirates fleet

Nescio0:
<s>Currently no Syndicate fleet deploys Wasps, which is odd, because they produce them. This proposal adds Wasps to small Syndicate fleets. It also reserves Splinters for large fleets.
With this proposal, of all small Syndicate fleet variants:
6/24=25% have a Manta (140 weapon capacity) – previously 2/11=18%
12/24=50% have three Quicksilvers (3×60=180 weapon capacity) – previously 5/11=55%
6/24=25% have five Wasps (5×28=140 weapon capacity) – 3/11=27% used Splinters previously</s>

[EDIT]: instead, introduced "Wasp (Gatling)", "Wasp (Javelin)", and "Wasp (Plasma)" variants and added them to the "Small Core Pirates" fleet. 
Furthermore, the "Wasp (Bomber)", recently introduced in 80f9f5967092b6f3590579bda1af796f37533480, has been renamed to "Wasp (Meteor)" for clarity and consistency.


Nescio0:
True, I ordered them alphabetically to make it easier for me to count them. Effectively the only rearrangement is that two Manta variants are moved from the bottom of the list to the top.

matkoniecz:
> I ordered them alphabetically to make it easier for me to count them

I strongly suggest to not introduce unneded changes in commits or make them as separate PR if truly useful

Amazinite:
I think that Syndicated Security ships not using Wasps makes sense. Wasps are interceptors, the weakest possible individual combat ship (i.e. not a fighter/drone), so I don't see why the Syndicate would use the Wasp for security purposes. The weakest individual combat ship that the Navy uses are light warships. Notably, among the FW, Syndicate, and Navy, only the FW uses interceptors right now. That makes sense though since the FW are also the least developed of the three and need to use whatever they can. Syndicated Security, being more developed, probably shouldn't be throwing interceptors into the mix of their defense fleets.

Made by the Syndicate doesn't necessarily need to mean used by Syndicated Security.

Nescio0:
On the other hand, the Syndicate's primary objective is maximizing profits, therefore it makes perfect sense for them to use their own Wasps, which are cheaper than Megaparsec's Quicksilvers. 
screenshot from 2018-09-10 18-09-12

Amazinite:
Yeah, it's less costly to run five Wasps versus three Quicksilvers, up until your Wasps start dropping like flies against pirate warships. Upkeep of Quicksilvers, replacing lost ships included, is probably cheaper than the upkeep for Wasps, which are more likely to die in any single battle.

Testing 5 Wasps or 3 Quicksilvers vs a single Firebird (a possible core pirate fleet), the Wasps win with 2 to 3 casualties (1.276 to 1.914 million credits lost) while the Quicksilvers win with 0 to 1 casualty (nothing to 2.018 million credits lost). While a loss of a Quicksilver hurts more when it does happen, the loss of a couple Wasps is practically guaranteed against any large pirate fleet.

Nescio0:
That's a quick reply! How many playtests did you actually run?
Besides, do small security fleets always have to easily defeat large pirate fleets?

Amazinite:
Five tests for each. I've found that 5 to 10 tests is typically enough to get an understanding of how fleet battles will generally end, but single fleet-on-fleet battles aren't exactly the norm anyways when merchants fleets are usually nearby.

I wouldn't say they should easily defeat large pirate fleets, I'm just taking a look at the actual efficiency of each fleet variant. And current, small syndicate fleets can more or less demolish any small core pirate fleet and hold their own against some large core pirate fleets (except for perhaps the multi-warship or the Falcon variants; the single Firebird admittedly has only a 4/34 chance of spawning, only 2/34 if we're looking at the exact variant I tested with, and only 10/34 large core pirate fleets spawn a single medium warship). If new fleet variants are added, this balance between small/large pirate fleets and small Syndicate fleets should remain as to not drastically alter the average outcome of a fight between said fleets.

Nescio0:
>  If new fleet variants are added, this balance between small/large pirate fleets and small Syndicate fleets should remain as to not drastically alter the average outcome of a fight between said fleets.

Which is why I added not only Wasp variants, but also Quicksilvers. Currently 6/12=50% of all small Syndicate fleets consisted of three Quicksilvers, with this proposal 12/24=50%.
Personally I don't really expect it to drastically change outcomes overall, because there are many other variables involved (other fleets and player ships), but if you believe otherwise, I suppose I could add a Wasp variant equipped with two meteors and a blaster.
The fun factor for me personally increases when I see some Syndicate Wasps flying around; they have a funny shape and colour.

Nescio0:
Now I think of it, actually Core pirates should occassionally use Wasps as well, they're cheap and readily available, and pirates are suicidal anyway, so it fits. Done.

Nescio0:
> It is really preferable to have single change in a single pull request.

Well, I suppose I could split this one into three, one that adds Wasps to the Small Syndicate fleet, one that adds Wasps to the Small Core Pirates fleet, and one that adds Wasps to the Human Miners fleets, if that's what people want; however, in my mind these are all related. Nor is it unusual; e.g. #3922 is an example with numerous changes in a single pull request.

Amazinite:
Speaking of #3922, I wouldn't merge any addition of Wasps into pirate fleets. #3922 specifically removed Wasps from a few areas due to them having Sidewinder missiles now, which are difficult for starter players to deal with. With #3922, only northern pirate fleets can spawn with Sidewinders, while all other pirate fleets will use Meteors and Javelins. 

Wasps to miner fleets doesn't seem necessary.

Wasps to Syndicate fleets is what this PR was created focusing on and is what it should've stayed as. I'm still not sold on the idea, though.

Nescio0:
So if I understand correctly, the Syndicate mass produces Wasps, but:
* the Syndicate doesn't use Wasps because they're too weak
* Pirates don't use Wasps because they're too strong
* Human Miners use Sparrows, Furies, Hawks, and Headhunters, but no Wasps
* players don't use Wasps because Berserkers and Furies are better interceptors

I'm starting to wonder why they are in game then.

data/variants.txt
```diff
@@ -1848,7 +1848,7 @@ ship "Vanguard" "Vanguard (Missile)"
 	gun "Sidewinder Missile Launcher"
 	gun "Sidewinder Missile Launcher"
 	gun "Sidewinder Missile Launcher"
-	turret "Heavy Anti-Missile Turret"
+
```
jafdy: Why was this line removed?

Amazinite:
It's not that pirates can't use the Wasp in general, it's just that I'd rather not them be using default Wasps. gatling, proton, or laser Wasps would be alright for use by pirates.

Wasps only have 10 cargo space, making them nearly useless for mining. The Sparrow and Fury at least have 15, so adding the Wasp to mining fleets for the sake of having it used more would just made it the worst mining ship. Someone who's trying to mine is probably just going to get a Sparrow, as it's cheaper and has more cargo.

Nescio0:
Replaced the Pirates' default Wasp with several variants and reverted the Human Miners' changes.

Amazinite:
I'm still not sold on the idea of Wasps in Syndicate fleets. They might be cheaper on upkeep than Quicksilvers, but they become more expensive as soon as any serious combat begins and Wasps start dropping quicker than Quicksilvers.

The Syndicate not having any Wasps also gives us some sort of differentiation between the military/security of the three major powers.
* Navy: Light warships, medium warships, and heavy warships (high tech, best in class ships)
* FW: Interceptors, medium warships, and heavy warships (middle of the road medium and heavies, need interceptors to fill ranks)
* Syndicate: light warships, medium warships, and barely-heavy warships (better off than the FW on the low end but worse than the Navy on the high end)

Nescio0:
Personally I think it's not too problematic if *small* Syndicate fleets with Wasps occassionally lose to e.g. a Falcon spawned by a *large* Core Pirates fleet. 

> * Navy: Light warships, medium warships, and heavy warships (high tech, best in class ships)
> * FW: Interceptors, medium warships, and heavy warships (middle of the road medium and heavies, need interceptors to fill ranks)
> * Syndicate: light warships, medium warships, and barely-heavy warships (better off than the FW on the low end but worse than the Navy on the high end)

Let's not compare apples with pears but look at their small fleets only.

* current fleet "Small Syndicate":
  * three Quicksilvers: 6/11=54.5%
  * one Manta: 2/11=18.2%
  * one Splinter: 1/11=9.1%
  * one Splinter plus one Quicksilver: 1/11=9.1%
  * one Splinter plus one Quicksilver: 1/11=9.1%
* proposed fleet "Small Syndicate":
  * three Quicksilvers: 12/24=50%
  * five Wasps: 6/24=25%
  * one Manta: 2/24=8.3%
  * one Manta plus one Quicksilver: 2/24=8.3%
  * one Manta plus two Wasps: 2/24=8.3%
* fleet "Small Free Worlds":
  * two interceptors (combinations of Furies, Hawks, and Sparrows): 34/58=58.6%
  * one Hawk plus two Sparrows: 12/58=20.7%
  * one Hawk: 6/58=10.3%
  * one Fury plus two Sparrows: 2/58=3.4%
  * one Sparrow: 2/58=3.4%
  * one Osprey: 2/58=3.4%
* fleet "Small Republic":
  * one Gunboat: 6/25=24%
  * two Rainmakers: 6/25%=24
  * one Gunboat plus one Rainmaker: 5/25=20%
  * one Frigate: 4/25=16%
  * two Gunboats: 3/25=12%
  * three Rainmakers: 1/25=4%

Basically I'd expect that even the five Wasps fleets are superior to the average small Free Worlds fleet of two interceptors. I'm not too concerned this proposal significantly alters the general status quo.

Nescio0:
> The Syndicate makes ships that they also just so happen to use for their own security forces

Actually the Manta, Quicksilver, and Splinter are all Megaparsec's. The only Syndicate ships the Syndicate currently uses are the Protector and Vanguard, in a minority of the large Syndicate fleets.

Amazinite:
Did I say that all the ships that the Syndicate uses are its own?

Nescio0:
Although I still think the Syndicate ought to use Wasps, the original purpose of this proposal, I've now reverted that, because it doesn't seem likely to be committed. 
Instead, this proposal introduces some Wasp variants and adds them to the Small Core Pirates fleet.

data/fleets.txt
```diff
@@ -1903,6 +1903,16 @@ fleet "Small Core Pirates"
...
+	variant 2
+		"Wasp (Gatling)"
+	variant 2
+		"Wasp (Javelin)"
+	variant 2
+		"Wasp (Meteor)"
+	variant 2
+		"Wasp (Plasma)"
+	variant 2
+		"Wasp (Proton)"
```
Amazinite: Fleet | Old chance | New chance
------|--------------|--------------
Single interceptor | 65.5% | 74.4%
Double interceptor | 6.9% | 5.1%
Light warship | 27.6% | 20.5%

Core pirate fleets might lean too much toward single interceptor fleets now. I'd suggest either reducing the frequency of all these wasp variants to 1, or adding new double interceptor and light warship variants to compensate and keep the chance of each fleet variant close to the original.

Nescio0: New changes:

    single Quicksilver from 4 to 6
    double Quicksilver from 2 to 3
    single Wasp variants from 0 to 10
    removed interceptor variants introduced in #3922 because some Wasps have missiles; from 5 to 0

That ought to largely preserve average fleet strength, I think.

PR #3927, opened 2018-09-08
add more zoom out levels

Nescio0:
Usually I play at a zoom level of 25% (minimum), but occassionally I want to be able to see where those annoying missiles are coming from, therefore an additional zoom level of 10% would be helpful.


source/Preferences.cpp
```diff
@@ -32,7 +32,7 @@ namespace {
 	const string EXPEND_AMMO = "Escorts expend ammo";
 	const string FRUGAL_ESCORTS = "Escorts use ammo frugally";
 	
-	const vector<double> ZOOMS = {.06, .09, .13, .18, .25, .35, .50, .70, 1.00, 1.40, 2.00};
+	const vector<double> ZOOMS = {.06, .09, .13, .18, .25, .35, .50, .71, 1.00, 1.41, 2.00};
```
Bladewood: I'm guessing Michael's original intention was to make them multiples of five. https://github.com/endless-sky/endless-sky/commit/436d345cb9a2b9a1a9990574ab16013d8155923d#commitcomment-19254321 might be relevant.

Amazinite:
These zoom levels are bigger than the mini-map is, which seems a bit odd when you're zooming out and seeing stuff further than your mini-map reaches. And as far as the reasoning behind this change goes, kinda hard to see where the missiles are coming from when everything's so small that you can't see anything smaller than a medium warship. Rainmakers become practically indistinguishable from stars at this zoom level. 
image

This much zoom out also puts massive stress on the GPU. Even in large battles I rarely even get close to like 80%, but zoomed out all the way with no battle going on and just the normal traffic flying through Vega, my GPU readout in game is over 100%, sometimes around 150%.

Janaszar:
Missiles appear on the minimap/radar in the top left of the screen, so finding where they come from isn't that much of an issue I find.  Usually a quick "which direction did the missile come from" check of the primary view, plus a quick check of the radar to see ships in that area, will tell me where the missiles are coming from.  Only odd situations like fighting the Unfettered in a fast ship, so their missiles constantly miss and have to turn around a lot, do I see missiles hitting me/getting close from a different angle than where the launching ship is.

And, as @Amazinite points out, zooming out too much just makes everything indistinguishable anyway.

Though, if such a zoom was added, I would request some way to "pause" the game without going to a menu (and maybe hiding the HUD), so you can take pictures of large battles/fleets.

comnom:
```int zoomIndex = 4;``` (line 36 in Preferences.cpp) should be modified as well or new players (and those without a preferences.txt) will end up with .25 as their view zoom. ```const vector<double> ZOOMS``` is a zero based index.

Amazinite:
Asking around, the consensus generally seems to be against adding more zoom out levels.

PR #3929, opened 2018-09-08
tweaked gatling

Nescio0:
b59a4ef25dd16c4ebe4fc5b4bb94cdda6c8835aa improved the damage per second of most human missiles by 20%. 
As a consequence the gap between javelin pods (size 12) and gatling guns (size 8) has widened. Yes, gatlings do not suffer from anti-missile systems, but bullets are also sold in fewer outfitters. Therefore I think they deserve some improvement. Currently:
screenshot from 2018-09-08 19-03-30

This proposal:
* increases gatling's shield damage by a third but leaves hull damage untouched;
* slightly tweaks lifetime to get average range between beam (300) and heavy (400) lasers;
* reduces bullet boxes size by one, because ammunition storage is typically half the size of the launcher (meteor 5/10, javelin 6/12, sidewinder 7/14) or less (larger launchers).

As a consequence:
screenshot from 2018-09-08 19-06-54

So the javelin pod is 50% larger and now inflicts about +200% shield and +20% hull damage compared to the gatling. (Shield damage per credit is 57/50 for the javelin and 4/4 for the gatling).

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -1059,8 +1059,8 @@ outfit "Bullet Boxes"
 	category "Ammunition"
 	cost 3000
 	thumbnail "outfit/bullet storage"
-	"mass" 2
-	"outfit space" -5
+	"mass" 1
+	"outfit space" -4
 	"gatling round capacity" 1500
 	ammo "Gatling Gun Ammo"
 	description "Bullet Boxes are used to store extra ammunition for Gatling Guns."
@@ -1082,7 +1082,7 @@ outfit "Gatling Gun"
 		icon "icon/gat"
 		"inaccuracy" 2
 		"velocity" 16
-		"lifetime" 1
+		"lifetime" 2
 		"reload" 3
 		"burst count" 180
 		"burst reload" 1
@@ -1102,7 +1102,7 @@ outfit "gbullet"
 		"random velocity" 8
 		"lifetime" 20
 		"random lifetime" 5
-		"shield damage" 3
+		"shield damage" 4
 		"hull damage" 6
 		"hit force" 2
 		"hit effect" "bullet impact"


Pointedstick:
Keep in mind that not being intercepted by anti-missile turrets is a *huge* advantage. Also keep in mind that the DPS is a bit misleading, since the burst-fire mechanic results in effectively triple the typical damage with the first volley, followed by a lengthy reload time. It's a gun meant for mounting four of them and decimating the target with the first shot.

DingusShingleton:
I really wish gatling anti missile turrets were possible. minigun point defense weaps are heckin sweet.

Nescio0:
> Keep in mind that not being intercepted by anti-missile turrets is a huge advantage.

Yes, it is, I'm aware of that, which is why I propose to increase the gatling's shield damage per second only to be about a third of the javelin pod. For comparison:
```
shield damage per shot, per second, per mass
energy blaster (5):        9.6 ,  48 ,  9.6
modified blaster (6):     10.8 ,  54 ,  9.0
beam laser (8):            1   ,  60 ,  7.5
current gatling gun (8):   3   ,  60 ,  7.5
proposed gatling gun (8):  4   ,  80 , 10.0
javelin pod (12):         57   , 228 , 19.0
```
This would put the gatling's shield damage per second per mass just above blasters, which do not require ammunition. It is important to note that gatling gun ammo is sold only in 10 systems, javelins in 53. 

> Also keep in mind that the DPS is a bit misleading, since the burst-fire mechanic results in effectively triple the typical damage with the first volley, followed by a lengthy reload time.

I'm not sure I quite understand. A rocket launcher fires once and then does nothing for four seconds, quite a lengthy time, relatively.

> It's a gun meant for mounting four of them and decimating the target with the first shot.

As is the case for several other secondary weapons.

> I really wish gatling anti missile turrets were possible. minigun point defense weaps are heckin sweet.

Gatling turrets would be nice to have, but it requires someone with the time and skills to make thumbnail and hardpoint sprites. Weapons that can target both missiles and ships would also be a great addition, but probably quite hard to implement. Anyway, discussing that is far beyond the scope of this proposal.

Pointedstick:
>> Also keep in mind that the DPS is a bit misleading, since the burst-fire mechanic results in effectively triple the typical damage with the first volley, followed by a lengthy reload time.
>
>I'm not sure I quite understand. A rocket launcher fires once and then does nothing for four seconds, quite a lengthy time, relatively.

Exactly. My point is that for these types of weapons, the DPS is of secondary consideration compared to the instantaneous burst of damage resulting from each hit. For example, the way you should think about the damage profile of a pair of rocket launchers is that they'll inflict 1700 damage per shot, not 424 DPS (which makes them look worse than javelin Pods on paper). Unguided weapons with high per-shot/volley damage and a long reload time make the most sense on small fast craft that make "attack runs" resulting on overshooting the target and turning around, which provide the requisite number of seconds for the weapon to reload. This minimizes the drawback of the reload time.

So for these types of weapons, it IMHO makes more sense to tweak and optimize their damage per shot, not their DPS. This is what I did with the missiles recently. The small DPS increases were of secondary importance compared to the large per-shot damage increases.

Nescio0:
Yes, b59a4ef25dd16c4ebe4fc5b4bb94cdda6c8835aa significantly increased per shot damage of human missiles, but the gatling gun wasn't changed, which is why I think they deserve a damage improvement, hence this proposal (shield+hull damage per bullet from 3+6 to 4+6).

Pointedstick:
> Yes, b59a4ef significantly increased per shot damage of human missiles, but the gatling gun wasn't changed

...Because the Gatling gun isn't a missile that can be jammed or intercepted by anti-missile turret fire. It's basically a gun weapon that happens to consume ammunition as a drawback.

Pointedstick:
For the Gatling Gun, neither "damage per shot" nor "damage per second" really displays the weapon's characteristics and utility. What's ideally needed is a display for "damage per burst", "bursts per second", etc.

Nescio0:
Perhaps you could explain and show me how to properly compare the gatling to other weapons?

> ...Because the Gatling gun isn't a missile that can be jammed or intercepted by anti-missile turret fire. It's basically a gun weapon that happens to consume ammunition as a drawback.

Currently the gatling has the same shield damage per second and per mass as the beam laser, which is the same size (8), but does not require ammunition, a serious advantage. 
Suppose you had a Fury (4 gun ports, 40 weapon capacity). How would you equip it? With gatlings? Personally I'd put them at the bottom of my list, even below blasters, which at least have range. If the gatling did not require any ammunition it might be an acceptable choice, however, that's not the case, therefore I believe it deserves some improvement, hence this proposal.
It is possible I'm overlooking something. Maybe you could enlighten me then?

Pointedstick:
I would equip my Furies stuffed to the gills with missile weapons, depending on their battle role. Two Sidewinder launchers and two Modified Blasters is a particularly lethal combination.

If I was choosing between guns, Beam Lasers would be on the bottom of my list, because of their extremely low range. I would prefer the Gatling Guns because they would do approximately three times as much damage per volley/attack run as the lasers--in addition to being slightly longer-ranged. A Fury equipped with four Gatling Guns would easily defeat a Fury armed with four lasers, no contest. Try it.

DingusShingleton:
The gatling gun as it currently is does 180/360 dps *while firing*, it just only fires 1/3 of the time. But for a ship that is doing strafing runs, it doesn't matter since ANY weapon they equip will be firing about the same amount.

Nescio0:
> A Fury equipped with four Gatling Guns would easily defeat a Fury armed with four lasers, no contest. Try it.

It seems you're right, I just tested it, and the Gatling Fury easily won (after changing "" from "never" to "always"). It doesn't really have a chance vs the Meteor Fury though.

Nescio0:
Hull damage is largely irrelevant in Fury vs Fury fights, because they have 2000 shields and only 400 hull, a ratio of 5:1. It is certainly important when fighting Kor Sestor ships, which have about a 1:1 ratio, but I don't expect anyone to be using gatlings at that stage.

PR #3935, opened 2018-09-11
Dialogue improvement

tmbutterworth:
The navy has a diverse selection of people with different religions: Christian, Islam and Buddhism are referenced in the game. Requiring a Buddhist or Muslim to swear on the Christian god is counter productive to diplomacy. Christianity declares Oath Swearing taboo as Jesus spoke against the practice so including God is taboo as well. The only commonality is their sense of honor which they consider sacred.


jafdy:
It could be reasonable to assume that most people will be Christian, or some other religion with a single god. Also it is worth noting that "God" doesn't necessarily refer to exclusively the Christian (and Jewish) god, it could just as easily refer to the Islamic god or one of the Sikh gods, it only really fails to apply to Buddhists and other religions without a god (plus atheists). And that they are swearing before god and not on god.
The result is a change that makes the sentence more clumsy without any tangible benefit.

There could be other things involved that make the situation more reasonable, like Muslims and Buddhists are in the minority, or the man in question could have been wearing a religious symbol like a cross and other men could swear a modified oath to the same effect that was acceptable to their own religion.

Nescio0:
Or just cut out "God and", simply "swear before these witnesses" ought to be enough.

tmbutterworth:
Nescio0

The game emphasizes The Honor of the navy. I was trying to emphasize that. It could be shortened to: "swear on your honor before these witnesses"

PR #3936, opened 2018-09-11
Dialogue Improvement

tmbutterworth:
Modified dialogue to emphasize honor.


jafdy:
You can change a previously opened PR as well, that does reduce the clutter some what.

tmbutterworth:
> You can change a previously opened PR as well, that does reduce the clutter some what.

Thanks just figured out how to do that.

endless-sky:
The current dialog is intentional; the Free Worlds planets have a higher prevalence of theistic religions than the rest of human space, and they're the ones deciding what the oath should be.

KiLEdEnNis:
(perhaps the handful that refuses follow some other religion fiercely enough not to take the oath)

PR #3937, opened 2018-09-11
Fixed grammer usage of parole

tmbutterworth:
Parole was being used in the wrong direction asking the navy to provide it instead of requiring them to accept the conditions to receive it.

### DIFF
--- a/data/free worlds start.txt
+++ b/data/free worlds start.txt
@@ -149,7 +149,7 @@ mission "Liberate Kornephoros"
 			
 			label speech
 			`	Tomek climbs up onto the hull of a nearby ship and addresses the prisoners. "Hello," he says. A few of the Navy officers respond, warily. "My name is Tomek Voigt," he says. "Our militia is very loosely organized, but I'm the one in charge of organizing it." That gets a few laughs from the Free Worlds guards. "As such, I am responsible for deciding your fate."`
-			`	At that, a hush falls over the crowd. "The Council has decided," he continues, "that any of you who gives us your parole, swearing to take no further part in fighting the Free Worlds, will be returned to Republic space. You will give us your word of honor that you will either seek reassignment in another part of the galaxy, or resign your commission."`
+			`	At that, a hush falls over the crowd. "The Council has decided," he continues, "that any of you who gives us your word, swearing to take no further part in fighting the Free Worlds, will receive parole and be returned to Republic space. You will give us your word of honor that you will either seek reassignment in another part of the galaxy, or resign your commission."`
 			`	A few of the Free Worlds guards shout, "What?" But only a surprising few. The rest are nodding, seemingly pleased with the Council's choice. Tomek climbs down from the hull and approaches the prisoners, and you notice Alondo and Freya have joined him.`
 			`	One of the Navy officers stands up. "I'll take that oath," he says, "if you'll honor your part of it."`
 			`	Tomek takes the man's hands, looks him in the eyes, and asks, "Do you swear before God and these witnesses to take no further part in military actions against the Free Worlds?"`


Nescio0:
There is no need to close requests and open new ones on the same issue. It's perfectly possible to edit files or add new changes.

As for "parole", there is a difference between the mass noun "the temporary or permanent release of a prisoner before the expiry of a sentence, on the promise of good behaviour" and the count noun "a promise or undertaking given by a prisoner of war to return to custody or act as a non-belligerent if released". (And there is also the verb "to parole": "release (a prisoner) on parole".)
In other words, prisoners can give parole (i.e. a formal promise) and can be paroled (be released on parole).

tmbutterworth:
> There is no need to close requests and open new ones on the same issue. It's perfectly possible to edit files or add new changes.
> 
> As for "parole", there is a difference between the mass noun "the temporary or permanent release of a prisoner before the expiry of a sentence, on the promise of good behaviour" and the count noun "a promise or undertaking given by a prisoner of war to return to custody or act as a non-belligerent if released". (And there is also the verb "to parole": "release (a prisoner) on parole".)
> In other words, prisoners can give parole (i.e. a formal promise) and can be paroled (be released on parole).

Prisoners or war, like prisoners in penal institutions receive parole from their captor upon meeting required conditions they do not give parole to the captor. Prisoners do not dictate parole to their captors as they have no power or authority. In the US military armed forces members are not allowed to accept parole from a captor because it constitutes serving a foreign government.

I know parole has lots of definitions but the current wording sounds like they are asking the navy prisoners to give them parole even though they are the captors. In military and prison only the group with authority to create the conditions can "give parole".

Mixing other rarely used definitions just creates confusion and most people just treat it as a typo.

tmbutterworth:
> There is no need to close requests and open new ones on the same issue. It's perfectly possible to edit files or add new changes.

Thanks I just figured out how to do that.

comnom:
>I know parole has lots of definitions but the current wording sounds like they are asking the navy prisoners to give them parole even though they are the captors.

Keep in mind that we're talking about the people who will become the Oathkeepers here.

So not only is the change not correcting any errors, it's weakening the statement from "a promise confirmed by a pledge of one's honor", to just any old joe giving his word.

FYI, the same usage is present in other missions concerning the soon to be Oathkeepers.

tmbutterworth:
> So not only is the change not correcting any errors, it's weakening the statement from "a promise confirmed by a pledge of one's honor", to just any old joe giving his word.

Pledging their honor and not to take up arms against the free worlds again is the conditional release to receive parole. Parole is being given by the free worlds to the navy. Parole of prisoners is given by the captor to the prisoner not the prisoner to the captor. It doesn't weaken it, it places it properly in common understanding. Not many people know parole as a "Secret Word" given to naval officers, which is the only way I see it could work as something the navy could give.

Nescio0:
Parole literally means a formal promise. The prisoners have to give their word to the authorities (as is clear from the current sentence), not the other way around (as your proposal suggests).

tmbutterworth:
> Parole literally means a formal promise. The prisoners have to give their word to the authorities (as is clear from the current sentence), not the other way around (as your proposal suggests).

If the navy won the battle and took free world captives parole would be theirs to give.

Any Navy members who refuse to swear to the conditions clearly would not receive parole as parole is a state of being/living condition. Prisoners or war in the dark ages that accepted parole then rejoined their unit to fight were clearly put to death if recaptured.

Only the side with the power to dictate the conditions can give parole the captive can only accept or decline it. A prisoner does no give parole to th e prison they receive it if the agree to the conditions being forced on them.

PR #3944, opened 2018-09-13
word change

tmbutterworth:


tmbutterworth:
Built is the past tense of build. As the church is ancient the construction would have completed a long time ago in the past making built a better choice.

PR #3945, opened 2018-09-13
Typo fix

warp-core:
Changed `"You're on council now,"` to `"You're on the council now,"` in FW Bloodsea 1.1

### DIFF
--- a/data/free worlds middle.txt
+++ b/data/free worlds middle.txt
@@ -1359,7 +1359,7 @@ mission "FW Bloodsea 1.1"
 				goto attack
 			label attack
 			`	"So, we attack them?" you ask.`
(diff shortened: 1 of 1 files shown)


Bladewood:
I don't explicitly disagree with this, although I wouldn't call it a typo. To me, it's more like an idiomatic expression, with a phrasing similar to "right on target/point" or "right on time".

endless-sky:
This wasn't actually a typo - if you google "serving on council" vs. "serving on the council" you'll find that the two phrases are used interchangeably by a lot of people. In my particular neck of the woods people are actually *much* more likely to say that someone is "on council" rather than "on the council" as long as it's clear which council they're referring to.

Issue #3498, opened 2018-01-05
Jump minimap, show all discovered connections


Tadrix:
This could be considered low priority, but it was bugging me for some (rather long) time.

When jumping, mini-map at the top shows origin and destination systems along with connected neighboring systems. It would be logical and more recognizable if connections between neighboring systems (if at least one of them is explored, so this connection is visible on the galaxy map) would be shown also.

tehhowch:
I think I hear what you're saying, but show a photo and remove the doubt for us

Tadrix:
![minimap](https://user-images.githubusercontent.com/17566547/34595328-c0a0beb6-f1df-11e7-8a55-a1d1a6eebace.png)

KiLEdEnNis:
so basically once a player made a Jump Drive only jump, they'll see a small "connection line"? partially nice but it could clutter the map a lot w/ even more lines, and those jump lines would cross normal hyperspace lines as well
if this happens, I want a preference to turn this off, and less obvious lines so it's clear yet doesn't stand out

endless-sky:
If all the neighbor lines were drawn, there would be a lot of them, and many of them would intersect. In my opinion, that would look cluttered and confusing.

Right now, the map shows your current system's "neighborhood" with a circle drawn around the system. That might be a less confusing option. But I still think it'd look cluttered. The mini-map is just intended to show you the current neighborhood you're in, and the shape of the hyperlink connections is part of what makes that neighborhood recognizable.

Tadrix:
Not jump lanes, only hyperspace lanes. I shouldn't use different color on the second image. The idea is to show all hyperspace connections between shown systems, if they are discovered and visible on the galaxy map. I've marked connections between Io Mann and Wah Oh systems, and connections between Heia Due and Io Lowe systems, because they are marked on the mini-map and connected via ordinary hyperspace lane.

Apparently I have trouble getting my point across. >< These yellow lines should be usual white lines. Gotta remember not to make issue posts when sleepy.

mdsmestad:
@Tadrix I have to agree with MZ on this one, this has the potential to make the map look cluttered

Tadrix:
@endless-sky @mdsmestad As I said one post earlier, I didn't propose to draw jump lines, only hyperspace lines — in other words, those that are currently visible on map.

I find it confusing that systems which are shown to be connected on the galaxy map aren't connected on the mini-map.

jafdy:
[...]

PR #3947, opened 2018-09-14
Added new outfit

tmbutterworth:
I added outfit for Tear Gas and made a Navy license that is required to purchase it.


jafdy:
A: You referenced an image that didn't exist. 
B: Why? What does it bring to the game? Why should the player care?

Yes, I know you closed this already but these are things that would have been pointed out anyway.

jostephd:
Duplicate of #3498. @tmbutterworth Next time you can just add commits to an existing PR instead of opening a new one.

jafdy:
#3948 jostephd is confused.

Issue #680, opened 2016-01-16
Use of the pilot license ?


ph2000bis:
Is the pilot license being available at outfitters bring anything to the game for either lore or gameplay purposes ?

endless-sky:
Originally there were a lot of different licenses depending on what sort of ship you wanted to fly. They were intended to reflect the fact that high licensing fees make it hard for small independent captains to compete with large corporations like the Syndicate. I took all the other licenses out because they were really annoying, but left the Pilot's license in place.

ph2000bis:
So it's not essential to have them available for sale in outfitters ?

Hyratel:
so by retaining the Pilot's License, you left the entire mechanic in place, just dummied out? That leaves an awesome hook for modders

endless-sky:
Yes, right now the only way that you acquire new licenses (e.g. the Wanderer ship license) is through missions. But it's still possible for them to be available as an outfit you purchase, instead.

Issue #275, opened 2015-11-27
Who pays all these space taxes that are repeatedly mentioned?


Pointedstick:
The whole human part of the galaxy is covered with governments, and taxation is prominently mentioned and discussed in various storylines. The Navy is funded with taxes; the Paradise planets pay ten times more taxes than the Dirt Belt planets; the Free Worlds' taxation model is weird and cool, etc. But we as the player never feel the impact of any taxes at all. There's no income tax, no sales tax, no personal property tax, no landing fees, no nothing! Say, who's paying for all those huge Republic ships and the sprawling Republic government, anyway? Are we really to believe that all starship-related activity is tax-free throughout the galaxy? The fact that taxes are repeatedly discussed but the player feels no impact from them weakens the concept and makes it feel slightly surreal. If you can get out of your tax burden 100% simply by living on a spaceship, well gosh, who wouldn't want that!?

Obviously taxes are no fun, but I think adding them in some form could strengthen the game's presentation of the storyline and provide interesting gameplay elements. For example, sales taxes applied to outfits could nudge players to shop on planets with lower tax rates, mimicking a real-world phenomenon. An income tax levied on missions completed in certain systems would do the same. Maybe the paradise planets levy a 15% income tax on any deliveries or commodity profits, but the Dirt Belt planets only levy a 5% tax. That would be interesting and really viscerally bring that element of the game world to life.

This could also provide a damper on the tendency of player net worth to snowball quickly. Taxes could be run through a multiplier determined by your net worth. Once you have 20 million credits, maybe you're paying like a 50% income tax. This could open the door to tax-related missions, where you smuggle untaxed goods onto high-tax planets, or work to lobby some government for lower tax rates.

Oeleo:
I totally agree, like that it will become harder to make money when you begin to become rich, and "it's getting harder and harder to avoid going into debt" will finally be true !

endless-sky:
In earlier versions there were a bunch of different ship licenses you 
had to buy (freighter license, transport license, capital ship license, 
etc.) but I took those out because people found them so annoying.

It's hard sometimes to balance realism with actually enjoyable game play.

Oeleo:
Sorry but are you sure that you posted in the good issue ? I don't understand. But yes for the last sentence that's one of the biggest challenge...

airtroop392:
[...]

PR #3953, opened 2018-09-17
Pilot's License Implementation

tmbutterworth:
The outfit "Pilot's License" states: "Without a pilot's license, you cannot purchase a ship on a Republic planet." This pull request implements that lore requirement in the game by adding "Pilot's License" as a license requirement to all ships that did not have some other license already applied to it.

The shipyard will now show that the Pilot's License is required for purchase.

The outfit "Pilot's License" also states: "And if you buy an unlicensed ship on another planet and then bring it into Republic space, you can be fined until you cough up the money for a license." This lore function could be implemented by creating a unique fleet of pirate ships that are only available through pirate shipyards which then have no license assigned to them. It doesn't make sense to have pirate shipyards require a republic license. The game could fine players for using these pirate ships.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -13,6 +13,8 @@ ship "Aerie"
 	thumbnail "thumbnail/aerie"
 	attributes
 		category "Medium Warship"
+		licenses
+			Pilot's
 		"cost" 3500000
 		"shields" 5700
 		"hull" 1900
@@ -72,6 +74,8 @@ ship "Argosy"
 	thumbnail "thumbnail/argosy"
 	attributes
 		category "Light Freighter"
+		licenses
+			Pilot's
 		"cost" 1960000
 		"shields" 4200
 		"hull" 2600
@@ -129,6 +133,8 @@ ship "Arrow"
 	thumbnail "thumbnail/arrow"
 	attributes
 		category "Transport"
+		licenses
+			Pilot's
 		"cost" 1200000
 		"shields" 2000
 		"hull" 400
@@ -180,6 +186,8 @@ ship "Bactrian"
 	thumbnail "thumbnail/bactrian"
 	attributes
 		category "Heavy Warship"
+		licenses
+			Pilot's
 		"cost" 17600000
 		"shields" 17500
 		"hull" 8600
@@ -248,6 +256,8 @@ ship "Barb"
 	thumbnail "thumbnail/barb"
 	attributes
 		category "Fighter"
+		licenses
+			Pilot's
 		"cost" 50000
 		"shields" 800
 		"hull" 200
@@ -290,6 +300,8 @@ ship "Barb" "Barb (Proton)"
 	thumbnail "thumbnail/barbp"
 	attributes
 		category "Fighter"
+		licenses
+			Pilot's
 		"cost" 50000
 		"shields" 800
 		"hull" 200
@@ -330,6 +342,8 @@ ship "Bastion"
 	thumbnail "thumbnail/bastion"
 	attributes
 		category "Medium Warship"
+		licenses
+			Pilot's
 		"cost" 3560000
 		"shields" 6700
 		"hull" 4200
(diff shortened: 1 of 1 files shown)


Amazinite:
>The game could fine players for using these pirate ships.

Such changes would just put us back at square 1: https://github.com/endless-sky/endless-sky/issues/275#issuecomment-160157480

NaruFGT:
This is not identical to the previous behavior. Simply requiring the pilot's license for piloting within the republic would be a reasonable and expected. If this is not the case, then we need to edit the misleading text that refers to the license being required.
@tmbutterworth clearly is not talking about the multi-tier license system that used to exist.
In any case, it would not encumber normal gameplay as obtaining the pilot's license is one of the first things to occur within the storyline.

Brick63:
What real difference would this make, to me this seems like change for change sake

NaruFGT:
It would be most useful to plugins and third party mods where someone may make a quest involving the pilot's license. 
It would be interesting if discarding the license resulted in daily fines by the republic.
Also I would be interested in adding a feature where negative standing with the republic results in revocation of the license.

NaruFGT:
There's already a framework for being fined if caught with illegal good, including checking for a pilot's license would be an easy addition.

NaruFGT:
The appropriate behavior would be when a user is on a republic controlled planet, access to the shipyard panel and the ability to depart would depend on the pilot having a license. The license would be presented as an option at the spaceport only if the pilot did not have a license in their inventory.

I think it would be fun to include a "forged" pilot's license available for purchase from the FW or pirate planets with a small chance of the forgery being discovered by a cargo scan.

MessyMix:
For the record, I think this isn't a bad idea. We have to be careful in the implementation as to not make gameplay "annoying" (like the licenses mentioned in #275), but implementing such a mechanic will only open new avenues in terms of game design. So not a bad shout.

DingusShingleton:
I agree. I think there is potential here for fun gameplay. The problems arise when mechanics add complication and drudgery but don't add a dimension of fun. As long as the license system is *optional*, that is, as long as you start with a license and can more or less do your thing without having to worry about it, I think adding ways to interact with it could be interesting.

Amazinite:
These ideas are all nice and good, but this PR doesn't actually get us any closer to those. This PR makes every human ship (that doesn't already require a license) require a pilot's license. That includes ships like the Modified Argosy that are only sold at southern pirate shipyards. The stated purpose of this PR is to add a "lore function," but I see this as actually moving us further away from the stated lore, as ships that shouldn't have a license like the mentioned Modified Argosy end up having one, and no changes happen as a result to the story when the FW splits from the Republic. 

Even ignoring these issues, this change by itself doesn't actually *do* anything. As @Brick63 said, this is really just a change for the sake of change right now.

MessyMix:
Of course for the moment this PR doesn't do anything for the moment, but it lays the groundwork for future use of this mechanic. Also, it can be adjusted so that ships like the Modified Agorsy _don't_ require the license, and we can implement the ramifications of this in a future PR (which, as far as maintaining a code base goes, is good practice with regards to separating commits with different functionality changes).

If this doesn't get merged at the moment, I would like to see it stay open until we find a way to make use of it. Still, I think embedding this change into the game will make it more likely that someone makes use of this mechanic to bring about an interesting change, so I'm all for this change right now.

Amazinite:
But again, #275 and #680. The mentioned uses of this change in the OP and subsequent posts all suggest returning to the mechanic of being fined if found without a proper license. Yes, it will not be on the same level as MZ has mentioned it being with various licenses for different classes of ships, but it still returns to the idea of "don't have the right license in the right region of space? Have a fine!"

Amazinite:
And as mentioned in https://github.com/endless-sky/endless-sky/pull/3867#issuecomment-410399949, a lot of descriptions of human ships and outfits are from early in the game's development, meaning that the best course of action may be not to change the outfit or how it is interacted with, but instead change the outfit's description to fit its current implementation.

NaruFGT:
The behavior of this implementation doesn't accurately reflect the lore, so I believe this PR should be closed, but I do not see a problem with the lore or the game mechanic as discussed. It would add depth without affecting normal gameplay.

PR #3960, opened 2018-10-15
Add a Remnant Hand to Hand

Brick63:
Repost of the original after it got messed up

Simply gives the Remnant a Hand to Hand weapon


Amazinite:
Are there any problems with this PR as is or any objections that anyone would have to merging it?

PR #3961, opened 2018-10-15
Add an Electron rifle

Brick63:
Repost of the original

Adds a new Human Hand to hand that makes it harder to capture the Mark II variants


Amazinite:
To be honest, I'm not really sure if I'm a fan of the electron rifle. Ref: the other criticism in #3923

PR #3436, opened 2017-12-19
Add color highlighting for best and worst deals

EmergencyTemporalShift:
This lets you see the best deals going to or from a system.
I'm up to discussion for making this feature better.

Issue #2671, opened 2017-06-24
Pos/Neg credit difference between comodities should be color coded


gunqqer:
Something to indicate which is + and which is -. I get exited when I see -410 then I notice it's *minus* 410, not plus. On this note maybe highlightning the higest positive difference.

edit: see MZs fix below (increasing text size for the plus)

Amazinite:
Similar to #1118

Festerdam:
I think that color highlighting is a bad idea. When you play this game, you feel like some futuristic stuff is going on and that you're touching steel when pressing buttons(it is used a lot of gray, white and black). But that look and feel would be destroyed if we put somewhere a color like red(which is a really aggressive color(blood pressure becomes higher and is great for warnings)).

But I think we could add an option in the settings called something like "dynamic commodity order", so that if this option is enabled, it would on the panel that we see on the map at the ports section display the commodities according to the absolute value of the profit(not the value of the profit, because people are normally looking not just for big positive changes, but also for big negative changes(useful if you are planning to go from one point to another and vice-versa.))

To make the distinction between negative values and positive values easier, maybe we could make the font-size bigger on the entire game, that way even color-blind people could still play. But I am not sure that would still look pretty, that would be a reason to only make the font-size bigger of the plus and minus that are displayed on the panels with the profit and not of the entire game.

tehhowch:
Numerical comparison may be better suited to using a font that has monospaced digits, as this would remove the left-right jumping one's eyes need to do when parsing the displayed relative prices. Alternately, removing or otherwise spacing the parentheses from the sign indicator could help as well.

Of these, the spacing can be done immediately, while the monospaced digit font would require obtaining a new font to use, and then using it only in tables of numbers (of which there are few).

current behavior:
![prespacing](https://user-images.githubusercontent.com/20871346/27557159-23f2a3b6-5a7e-11e7-94ce-81a3bee0c132.png)

space before + or -:
![post-spacing](https://user-images.githubusercontent.com/20871346/27557167-2e8b8f40-5a7e-11e7-9b79-f030e4a80af8.png)

space before + or -, and after number:
![double-spacing](https://user-images.githubusercontent.com/20871346/27557188-3ee2a702-5a7e-11e7-9df9-2632ebff2ab3.png)

removing parentheses entirely:
![no-parentheses](https://user-images.githubusercontent.com/20871346/27557202-4a8e0182-5a7e-11e7-9492-4bcd25faba74.png)

[...]

Issue #1539, opened 2016-09-10
The most profitable commodity to buy should be highlighted


Pointedstick:
Which commodity should I buy?

<img width="171" alt="screen shot 2016-09-10 at 4 18 31 pm" src="https://cloud.githubusercontent.com/assets/1097249/18413821/3c427f12-7772-11e6-8ba0-05cf41a02884.png">

Looks like food. Took me a couple seconds to parse because not only was it not highlighted, but my current selection was. It would be nice if the most profitable commodity could be somehow highlighted in a manner that was still distinct from your current selection.

endless-sky:
This is a duplicate of #1118. My worry is that if trading is made too "automatic," the player won't have any sense of agency - basically, the game would be telling the player, "this is the commodity you're supposed to buy," and that leaves even less for the player to actually do themselves.

Pointedstick:
Fair enough. I see your point.

KiLEdEnNis:
side note, buy the industrial

Issue #1118, opened 2016-05-22
[enhancement] Highlight Most Profitable Commodity in Map's List


Haravikk:
First up I want to say I love what's been done with trading in Endless Sky, it feels so much less like a chore than it does in EV thanks to the simplicity of finding good places to sell goods on your intended route etc.

However, at the moment I'm not really trading for its own sake but instead have some intended destination due to a story mission or other mission, and use the map to determine which commodity will be the most profitable in that system (or one on the way) so I can load up and head out to make some extra profit on the side.

This is fine, as it's usually fairly easy to pick out which commodity is the most valuable, but I think it would be useful if it were highlighted in some way, for example if the most valuable commodity (or its price difference) appeared in white, or even if they were all given a brightness based upon their relative value. I've added a super-simple mockup of the first of these (only the top commodity is brighter), but I think the graded option may be best.

![mockup](https://cloud.githubusercontent.com/assets/2184328/15453418/1d7a086e-200d-11e6-86c4-e7eb707d7af8.png)

After making this it occurs to me that hovering over or clicking on commodities also changes their colour, so it may make sense to apply this grading only to the price difference (the bracketed number) so that this is applied normally to the commodity name.

Anyway, the point is that it would be nice if the commodities were somehow highlighted based upon their relative value between the current and target systems, as this is usually what I need to see so I can quickly load up on the most profitable to cover my crew wages and make a bit of extra money while I go around saving the Free Worlds from disaster etc.

lshart:
That would be a nice feature.  I think it would be even more convenient if you could buy commodities from the map screen as well, perhaps with some kind of "fill up on most valuable" button.

jafdy:
I can't help but feel that it would detract from the interest of trading. If all the information is given to you color coded for your convenience you would never really look at what you are doing. More so with @lshart suggestion.

You can already click on commodities to see there relative price across the galaxy.

LocalGod79:
hypothetical situation; i'm on a planet, and i have some goods i bought at "very low", and they're "medium" here, but the next system over, they're "high".

under your proposal, when i click the next system over in the map, it will highlight the commodity with the highest profit margin compared to the system i'm in - not the commodity i had already selected to find a more profitable destination for.

wambat1:
[...]

PR #3963, opened 2018-10-16
Colour profit / price column in Trading Screen green / yellow / red.

ecraven:
To make it simpler to see how good prices are, colour the words "(very low)" etc.
maim
Also colour the profit (in the same column) green (if > 0) or orange (if < 0).

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -100,6 +100,12 @@ color "map travel ok none" .55 .1 0. 0.
 color "map used wormhole" .5 .2 .9 1.
 color "map unused wormhole" .165 .066 .3 .333
 
+# Colors used for drawing the trade good prices (green - yellow - red)
+color "trade level very low" 0.01 0.8 0.02 0.
+color "trade level low" 0.4 0.8 0.03 0.
+color "trade level medium" 0.8 0.75 0.01 0.
+color "trade level high" 0.8 0.4 0.02 0.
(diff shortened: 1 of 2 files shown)


Amazinite:
#1118
#1539
#2671
#3436

jostephd:
How about using a graphic representation of low/high instead of words? For example, the words "very high" could be replaced by a progress bar that's 80% empty, the words "very low" by a progress bar that's 80% full, the word "medium" by a half-full progress bar, and so on. The text "profit: 115" could be shown on top of the progress bar (it's only there right after the player lands, anyway). This would preserve the general grayscale UI design.

Zireael07:
@jostephd: Or arrow up/arrow down/dash.

Darcman99:
Could at least be a little softer, rather than full vibrancy.

ecraven:
> Could at least be a little softer, rather than full vibrancy.

The colours are defined in interface.txt, if this is acceptable in
theory, I'll try to find a much toned down palette.

jostephd:
In light of the :+1: reactions I opened https://github.com/endless-sky/endless-sky/issues/3964 for my suggestion. (If MZ prefers this approach we can close the other issue.)

Amazinite:
Closing this, as it seems like this isn't the implementation that would be used if color were added to the trading screen. Refer to #3964 for continued discussion on the topic.

Issue #2651, opened 2017-06-19
Trove isn't a moon


Amazinite:
[Trove](https://github.com/endless-sky/endless-sky/blob/master/data/map.txt#L26880) is a landable object described to be a moon and has a moon attribute, but [it isn't a moon of any planet](https://github.com/endless-sky/endless-sky/blob/master/data/map.txt#L11021). It's just a really small planet.

endless-sky:
Fixed in 0e2c590.

PR #3970, opened 2018-10-24
Fix Passenger Number

recneps:
As identified here, it should be 3 passengers. https://steamcommunity.com/app/404410/discussions/0/1736594593592744449/


jafdy:
Duplicate #2651

Amazinite:
This is also handled properly by #3216.

PR #3433, opened 2017-12-18
Increased Kestrel base cost

Nescio0:
Currently the Kestrel's base cost (10.3m) is actually lower than the Falcon's (10.9m), although it's stronger than the Bactrian (17.6m). This patch increases the Kestel's base cost to 14.7m, somewhere in between.

EDIT: d2e19f154a37c267562192847a720fd4a36e3c1b improved the Kestrel, justifying a price increase.

PR #3120, opened 2017-10-13
Bactrian Mini-Campaign

Amazinite:
This is a continuation of #2347 and addresses a number of my own critiques about my original attempt to license-lock the Bactrian as well as vastly expanding the amount of non-campaign story content in game, adding 26 new missions for the player to enjoy.
<hr/>

The following is an overview of the various mission strings added by this PR which will lead up to the unlocking of the Bactrian:

* `Deep: Convoy`: Transport convoys of multiple freighters across the galaxy to pick up unknown supplies, then return them to the Deep, all while on a deadline and under threat of pirate attacks.
   * Type: Escort
   * Mission count: 3 (each mission is separate and only one must be done)
   * Notes: Tested in a single Modified Argosy using only human technology, retested in a human tech Raven
* `Deep: Mystery Cubes`: Help the Deep attempt to apprehend a particularly meddlesome pirate warlord who goes by the name of Beelzebub.
   * Type: Combat
   * Mission count: 4
   * Notes: Tested in a single Corvette using only human technology, retested in a human tech Raven
* `Deep: TMBR`: An optional string that triggers if TMBR are in Hai space. Meet up with a group of Deep scientists who wish to see the band live, and make their dreams come true.
   * Type: Transport
   * Mission count: 2
* `Deep: Remnant`: This mission string acts as a continuation of `Terminus exploration` as the scientist that you met in that mission continues to research the wormhole and what lies beyond it. Help the Deep learn about the existence of the Remnant, the human faction which technologically rivals even the Deep, and if you wish to, bring some Remnant technology to the Deep for them to analyze.
   * Type: Exploration
   * Mission count: 11 (only 8 to 9 doable on a single pilot since 4 of the missions are involved in branches. The last 5 missions are optional.)
* `Deep: Project Hawking`: Help a group of Deep scientists (that you may or may not have met in Deep: TMBR) to travel to multiple locations in the Paradise Worlds and Far North to retrieve cargo that is crucial for a government project known only to the player as "Project Hawking."
   * Type: Cargo
   * Mission count: 1 (+ 2 preliminary missions)
* `Deep: Scientist Rescue`: A rewritten version of the original mission string from #2347. The premise is the same, but it has been rewritten to include information and characters from the above mission strings that all lead up to this one.
   * Type: Combat
   * Mission count: 3
   * Note: Tested in a single Falcon using human and Hai technology, retested in a Vanguard with mainly Hai tech

[...]

PR #3976, opened 2018-11-01
Tweaked Kestrel

Nescio0:
The Kestrel is supposed to be a superb, modern, state-of-the-art warship of alien design. In practice, it's hardly better than a Bactrian, Carrier, or Dreadnought, its nearest human competitors. This patch intends to improve the Kestrel's desirability by tweaking various stats:

* +2000 shields, +1000 hull, putting it in the same range as the Carrier
* reduced required crew to 64, putting its daily upkeep slightly below the Bactrian's 70
* lowered drag to 11, because the Kestrel is a highly streamlined design, therefore it is odd it is currently relatively slow; for comparison, Cruiser is at 10.1, Dreadnought at 10.1, and Falcon at 6.7 (maximum velocity = thrust / drag)
* improved heat dissipation to .6; for comparison, Dreadnought has .65 and Falcon .7
* +10 weapon capacity, to allow six plasma cannons plus four plasma turrets
* <s>increased the number of gun ports from six to twelve; for comparison, the Carrier and Shield Beetle have eight, the Vanguard and Albatross seven</s>

variants:
* (More Engines):
  * −1 drag, making it c. 10% faster with the same engines
  * +100 fuel capacity, useful if it has an afterburner
  * +30 engine capacity (instead of +20), to allow X5700 thruster plus X5200 steering, or A860 thruster plus A865 steering plus afterburner
* (More Shields):
  *  +3000 shields and +1500 hull (instead of +2700 and +1300)
* (More Weapons):
  * +40 weapon capacity (instead of +20), to allow six typhoon launchers plus four electron turrets
  * 3×6=18 gun ports instead of the ordinary 6

### DIFF
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -210,6 +210,7 @@ event "kestrel available: more shields"
 	shipyard "Tarazed Advanced"
 		"Kestrel (More Shields)"
 
+
 ship "Unknown Ship Type"
 	sprite "ship/kestrel"
 	attributes
@@ -217,15 +218,15 @@ ship "Unknown Ship Type"
 		"cost" 10300000
 		"shields" 17400
 		"hull" 6200
-		"required crew" 72
+		"required crew" 64
 		"bunks" 128
 		"mass" 740
-		"drag" 12.5
-		"heat dissipation" .45
+		"drag" 9.8
+		"heat dissipation" .6
 		"fuel capacity" 500
 		"cargo space" 120
 		"outfit space" 810
-		"weapon capacity" 380
+		"weapon capacity" 390
 		"engine capacity" 210
 		weapon
 			"blast radius" 260
@@ -233,16 +234,16 @@ ship "Unknown Ship Type"
 			"hull damage" 1300
 			"hit force" 3900
 	outfits
-		"Particle Cannon" 4
-		"Torpedo Launcher" 2
-		"Torpedo" 60
+		"Particle Cannon" 6
 		"Heavy Laser Turret" 4
+		
 		"Fusion Reactor"
 		"LP144a Battery Pack"
 		"D94-YV Shield Generator"
 		"Liquid Nitrogen Cooler"
+		
 		"Orca Plasma Thruster"
-		"X5200 Ion Steering"
+		"Orca Plasma Steering"
 		"Hyperdrive"
 	engine -14 177
 	engine 14 177
@@ -250,8 +251,8 @@ ship "Unknown Ship Type"
 	gun 31 66 "Particle Cannon"
 	gun -53 61 "Particle Cannon"
 	gun 53 61 "Particle Cannon"
-	gun -75 57 "Torpedo Launcher"
-	gun 75 57 "Torpedo Launcher"
+	gun -75 57 "Particle Cannon"
+	gun 75 57 "Particle Cannon"
 	turret -7 -4 "Heavy Laser Turret"
 	turret 7 -4 "Heavy Laser Turret"
 	turret -23 14 "Heavy Laser Turret"
@@ -262,6 +263,7 @@ ship "Unknown Ship Type"
 	explode "huge explosion" 30
 	"final explode" "final explosion large"
 
+
 ship "Kestrel"
 	sprite "ship/kestrel"
 	thumbnail "thumbnail/kestrel"
@@ -270,15 +272,15 @@ ship "Kestrel"
 		"cost" 10300000
 		"shields" 17400
 		"hull" 6200
-		"required crew" 72
+		"required crew" 64
 		"bunks" 128
 		"mass" 740
-		"drag" 12.5
-		"heat dissipation" .45
+		"drag" 9.8
+		"heat dissipation" .6
 		"fuel capacity" 500
 		"cargo space" 120
 		"outfit space" 810
-		"weapon capacity" 380
+		"weapon capacity" 390
 		"engine capacity" 210
 		weapon
 			"blast radius" 260
@@ -286,16 +288,16 @@ ship "Kestrel"
 			"hull damage" 1300
 			"hit force" 3900
 	outfits
-		"Particle Cannon" 4
-		"Torpedo Launcher" 2
-		"Torpedo" 60
+		"Particle Cannon" 6
 		"Heavy Laser Turret" 4
+		
 		"Fusion Reactor"
 		"LP144a Battery Pack"
 		"D94-YV Shield Generator"
 		"Liquid Nitrogen Cooler"
+		
 		"Orca Plasma Thruster"
-		"X5200 Ion Steering"
+		"Orca Plasma Steering"
 		"Hyperdrive"
 	engine -14 177
 	engine 14 177
@@ -303,8 +305,8 @@ ship "Kestrel"
 	gun 31 66 "Particle Cannon"
 	gun -53 61 "Particle Cannon"
 	gun 53 61 "Particle Cannon"
-	gun -75 57 "Torpedo Launcher"
-	gun 75 57 "Torpedo Launcher"
+	gun -75 57 "Particle Cannon"
+	gun 75 57 "Particle Cannon"
 	turret -7 -4 "Heavy Laser Turret"
 	turret 7 -4 "Heavy Laser Turret"
 	turret -23 14 "Heavy Laser Turret"
@@ -317,16 +319,15 @@ ship "Kestrel"
 	description "Several years ago, while visiting a sweat lodge in search of spiritual renewal, Tarazed's chief ship designer fell into a trance and journeyed to a reality separate from our own. There he saw visions of a strange and elegant starship. Returning to his work with renewed purpose, he spent the next seven years making the Kestrel a reality."
 
 
-
-ship "Kestrel" "Kestrel (More Weapons)"
-	add attributes
-		"weapon capacity" 20
-
 ship "Kestrel" "Kestrel (More Engines)"
 	add attributes
-		"engine capacity" 20
+		"engine capacity" 30
 
 ship "Kestrel" "Kestrel (More Shields)"
 	add attributes
-		"shields" 2700
-		"hull" 1300
+		"shields" 3000
+		"hull" 1500
+
+ship "Kestrel" "Kestrel (More Weapons)"
+	add attributes
+		"weapon capacity" 40
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -216,7 +216,7 @@ ship "Unknown Ship Type"
 	attributes
 		category "Heavy Warship"
 		"cost" 10300000
-		"shields" 18400
+		"shields" 19400
 		"hull" 7200
 		"required crew" 64
 		"bunks" 128
@@ -270,7 +270,7 @@ ship "Kestrel"
 	attributes
 		category "Heavy Warship"
 		"cost" 10300000
-		"shields" 18400
+		"shields" 19400
 		"hull" 7200
 		"required crew" 64
 		"bunks" 128
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -221,7 +221,7 @@ ship "Unknown Ship Type"
 		"required crew" 64
 		"bunks" 128
 		"mass" 740
-		"drag" 10.8
+		"drag" 11
 		"heat dissipation" .6
 		"fuel capacity" 500
 		"cargo space" 120
@@ -277,7 +277,7 @@ ship "Kestrel"
 		"required crew" 64
 		"bunks" 128
 		"mass" 740
-		"drag" 10.8
+		"drag" 11
 		"heat dissipation" .6
 		"fuel capacity" 500
 		"cargo space" 120
@@ -325,6 +325,7 @@ ship "Kestrel"
 
 ship "Kestrel" "Kestrel (More Engines)"
 	add attributes
+		"drag" -1.0
 		"fuel capacity" 100
 		"engine capacity" 30
 
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -249,18 +249,18 @@ ship "Unknown Ship Type"
 		"Hyperdrive"
 	engine -14 177
 	engine 14 177
-	gun -31 65 "Particle Cannon"
-	gun 31 65 "Particle Cannon"
-	gun -42 63
-	gun 42 63
-	gun -53 61 "Particle Cannon"
-	gun 53 61 "Particle Cannon"
-	gun -64 59
-	gun 64 59
-	gun -75 57 "Torpedo Launcher"
-	gun 75 57 "Torpedo Launcher"
 	gun -86 55
 	gun 86 55
+	gun -75 57 "Torpedo Launcher"
+	gun 75 57 "Torpedo Launcher"
+	gun -64 59
+	gun 64 59
+	gun -53 61 "Particle Cannon"
+	gun 53 61 "Particle Cannon"
+	gun -42 63
+	gun 42 63
+	gun -31 65 "Particle Cannon"
+	gun 31 65 "Particle Cannon"
 	turret -7 -4 "Heavy Laser Turret"
 	turret 7 -4 "Heavy Laser Turret"
 	turret -23 14 "Heavy Laser Turret"
@@ -311,18 +311,18 @@ ship "Kestrel"
 		"Hyperdrive"
 	engine -14 177
 	engine 14 177
-	gun -31 65 "Particle Cannon"
-	gun 31 65 "Particle Cannon"
-	gun -42 63
-	gun 42 63
-	gun -53 61 "Particle Cannon"
-	gun 53 61 "Particle Cannon"
-	gun -64 59
-	gun 64 59
-	gun -75 57 "Torpedo Launcher"
-	gun 75 57 "Torpedo Launcher"
 	gun -86 55
 	gun 86 55
+	gun -75 57 "Torpedo Launcher"
+	gun 75 57 "Torpedo Launcher"
+	gun -64 59
+	gun 64 59
+	gun -53 61 "Particle Cannon"
+	gun 53 61 "Particle Cannon"
+	gun -42 63
+	gun 42 63
+	gun -31 65 "Particle Cannon"
+	gun 31 65 "Particle Cannon"
 	turret -7 -4 "Heavy Laser Turret"
 	turret 7 -4 "Heavy Laser Turret"
 	turret -23 14 "Heavy Laser Turret"
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -216,8 +216,8 @@ ship "Unknown Ship Type"
 	attributes
 		category "Heavy Warship"
 		"cost" 10300000
-		"shields" 17400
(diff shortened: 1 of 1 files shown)


tehhowch:
I would have recommended you simply update your other pull, to avoid losing the relevancy of discussions. I think you are able to mark comments solely about now-discarded changes as "resolved" to keep only the relevant bits immediately visible.

From what I can see, the difference between here and #3433 is

 - you removed changes to weapon loadouts
 - you add 100 fuel capacity
 - you alter the hull/shields of the "More Shields" variant
 - you do not touch the Dreadnought

endless-sky:
Technically speaking it's not an alien ship; it's a human design from an [alternate universe](http://evn.wikia.com/wiki/Kestrel). So, as soon as the player has access to alien ships, it's going to be outdated.

I wonder if there would be some other way for the Kestrel to be unique or specially differentiated. The trouble is there are so many human ships already that even at the high end, there are not many niches that are not already filled. But maybe if it had something like an absurd number of hardpoints (for the many small guns approach) or a ton of crew (the ultimate ship-capturing flagship) or something like that, it would feel like more of a reward for the people who have no EV nostalgia.

Nescio0:
Increasing the number of bunks seems a bad idea; if you want to capture other ships, get a Bactrian. And increasing the number of gun slots would make it more similar to the Shield Beetle.
The Kestrel's uniqueness is its high weapon capacity (only the Hurricane, Kar Ik Vot 349, Model 256, and Model 512 have more) and its high engine capacity (higher than anything else). This patch increases the Kestrel's heat dissipation, allowing it to be able to support heavy weapons, and decreases the Kestrel's drag, allowing it to be faster than the Dreadnought or other large ships. 
Although it is no match for a Hurricane or Punisher, it can outspeed them; basically, the Kestrel's niche would be high maximum velocity and high firepower.

Human ships have low shields and hull, therefore the Kestrel is already outdated by the time it's introduced, which is unfortunate. Hai ships are immediately available and Remnant ships can be easily unlocked; all three times I played the storyline (starting with a Star Barge, Shuttle, and Sparrow, respectively), I owned multiple Shield Beetles and unlocked Remnant technology before I unlocked the Kestrel or started the Free Worlds campaign. The easiest solution would be to licence-lock the Shield Beetle and Bactrian, so you can no longer purchase them, only capture them. However, that's a different discussion and belongs elsewhere (see also #3120 and #3824).

Lineth:
A big part of my disappointment with the Kestrel is its flimsiness. Obviously, it should be significantly frailer than the Shield Beetle, but IMO it needs a modest buff in the survivability department. Making the shielded variant roughly equivocal to the Korath Raider and the regular versions roughly on par with the current shields variant would be a decent point to start with (would need fine-tuning, of course).

Nescio0:
Well, I suppose a slight shields and hull increase won't hurt. How about +2000, +1000?
For comparison:
```
                                shields + hull
Albatross:                        34200 + 12600
Shield Beetle:                    32000 +  9800
Raider:                           27000 +  9000
proposed Kestrel (More Shields):  22400 +  8700 *
Carrier:                          21400 +  8300
current Kestrel (More Shields):   20100 +  7500 *
proposed Kestrel:                 19400 +  7200 *
Dreadnought:                      18100 +  7300
current Kestrel:                  17400 +  6200 *
Bactrian:                         17500 +  8600
Spire:                            14200 +  4900
Hulk:                             13800 +  5700
Sojourner:                        13000 +  7700
```

Amazinite:
Just a heads up, people would certainly like to see the Kestrel buffed to be more relevant, it's just a matter of how it should be buffed.
image

I'm personally on board with this idea.
>an absurd number of hardpoints (for the many small guns approach)

The number of guns could be doubled up or even tripled. This wouldn't change how larger weapons are used, but it would allow for someone to equip 12 or 18 smaller missile launchers and just rain death from afar. That alone would be something to make the Kestrel a very interesting ship.

As for the changes done so far in this PR, I don't have any objections to them. I just wouldn't be changing the default loadout on the ship. Extra space doesn't mean it needs to be used by the default loadout.

Nescio0:
Maximum velocity = thrust / drag. Although the Kestrel (210 engines, 12.5 drag) has a lot of capacity, it is currently significantly slower than a Dreadnought (190 engines, 10.1 drag) or Cruiser (170 engines, 10.1 drag). I think the Kestrel ought to be the fastest heavy warship. By reducing its drag by about a fifth, its maximum velocity increases by about a quarter. It would still remain slower than smaller ships such as a Corvette (100 engines, 3.5 drag). 

As for the default loadout, the free space actually increases by +16.

Nescio0:
To put things into perspective, let's sort heavy warships by engine capacity / drag:
```
Falcon:           150 /  6.7 = 22.4
proposed Kestrel: 210 /  9.8 = 21.4
Derecho:          164 /  8.7 = 18.9
Dreadnought:      190 / 10.1 = 18.8
Shield Beetle:    150 /  8.8 = 17.0
Cruiser:          170 / 10.1 = 16.8
current Kestrel:  210 / 12.5 = 16.8
Albatross:        142 /  8.6 = 16.5
Leviathan:        120 /  7.6 = 15.8
Punisher:         187 / 12.2 = 15.3
Vanguard:         120 /  8.0 = 15.0
Raider:           159 / 12.0 = 13.3
Carrier:          210 / 17.7 = 11.9
Hurricane:        140 / 11.9 = 11.8
Bactrian:         180 / 16.1 = 11.2
Solifuge:         166 / 15.4 = 10.8
Kar Ik Vot 349:   173 / 16.8 = 10.3
Protector:        100 / 10.3 =  9.7
Model 256:        145 / 15.3 =  9.5
Model 512:        161 / 17.4 =  9.3
```

Nescio0:
In practice the limiting factor is the size of the engines for sale. The Dreadnought (190) can equip a Bufaer thruster (104) + Bufaer steering (76); the Kestrel has +20 capacity, enough for an additional Benga steering (20) to compensate for its higher mass. If the Kestrel's drag is reduced to 9.8, this set-up gives the Kestrel 3% more maximum velocity than the Dreadnought, but with a lower accelaration.

Nescio0:
Actually the set-up I was talking about was Bufaer thruster + Bufaer steering and Benga *steering*, because when I'm normally outfitting a ship, I prioritize steering, so I choose the largest steering I can get away with, then add the best thruster that fits, and if there is still some space left, I add extra steering. So effectively both my Dreadnought and my Kestrel would have "only" a Bufaer thruster; realistically, I doubt anyone would equip more thrusters than that.
And the 3% faster is the drag ratio: 10.1/9.8=1.03; currently the Kestrel is 10.1/12.5=0.81, i.e. 19% slower than the Dreadnought with the same engines.

Anyway, let's take a step back and look at the ship sprites only. Don't you agree the Kestrel looks much faster than the Dreadnought?

You consider a reduction to 9.8 as too severe, I view 11.5 as not enough; what do you think about the following compromise? Kestrel from 12.5 to 10.5 *and* Dreadnought from 10.1 to 11.1.

Amazinite:
Well now we'd just be arguing over a matter of preference as to what engines are used. Yes, using the same thrusters with variable steering will get you a 3% difference, but using the same steering with variable thrusters will give you similar steering (slight differences in mass change that) and vastly different max speeds. If we're going to look at some sort of middle ground though (balance thrust and steering instead of going all thrust or all steering), the 27.6% increase in current from the reduction in drag is still massive. I'd go for an increase of 15% at the highest, 10% at the lowest, which puts us between 10.9 and 11.3 drag. Being an already fast ship (or at least a ship capable of being fast), I'd still lean toward the higher number.

The Dreadnought doesn't need touched in this PR as that's outside its scope. The "looks" argument isn't always the best, because I think the Cruiser "looks" faster than the Dreadnought, but that isn't the case.

Let's not get caught up in the drag discussion though, as making a fast ship faster isn't exactly going to make it unique. Thoughts on tripling up on gun ports?

Nescio0:
Yeah, I agree the Cruiser looks faster, which is why a slight drag increase for the Dreadnought wouldn't be bad; however, that can be done in a separate request.

Trebling the gun ports is excessive. (Besides, isn't there a maximum of 16 slots?) What could be tried out is "only" doubling the number of gun ports, putting the Kestrel at twelve guns and four turrets. Right now the Shield Beetle has the most, eight guns and four turrets.
However, I think that's beyond the scope of this proposal and should be tried out in another request. The tweaks in this one would improve the Kestrel as it currently is, but also be helpful if the Kestrel is supposed to be a glass cannon. Besides, if the Kestrel has more gun ports, its sprite has to be changed as well.

Amazinite:
>Trebling the gun ports is excessive... Right now the Shield Beetle has the most, eight guns and four turrets.

Exactly. 

>Besides, isn't there a maximum of 16 slots?

[32](https://github.com/endless-sky/endless-sky/blob/ffd15a7453c7cc48b5c70ed7a8bbd3beb4f7e1cb/source/Command.h#L132)

>However, I think that's beyond the scope of this proposal and should be tried out in another request.

Giving the Kestrel an absurd number of guns (which 18 would be) is exactly what MZ mentioned as a possible change to the Kestrel in this very PR. I wouldn't consider such a change out of scope for that very reason.

>Besides, if the Kestrel has more gun ports, its sprite has to be changed as well.

Not necessarily. Check MZ's person ship. It has 12 guns, two for each port on the sprite (which is at the very least a good argument to double up on the existing gun ports, but I'd still go for 18). And as mentioned [elsewhere](https://github.com/endless-sky/endless-sky/issues/3264#issuecomment-346884228), ships already exist with more gun ports than are visible on the sprite because some gun ports are doubled up.

Nescio0:
Twelve guns would already be excessive. Its closest human "competitor" is the Vanguard, with seven. Anyway, let's try it out and see what others think.

Personally I highly dislike ship sprites with multiple slots on the same slot, therefore I've put the new slots in between the existing ones. And I've replaced the Kestrel's default weapons with a dozen heavy lasers, to make use of the new gun ports.

Here's a plug-in for those who want to play-test with these changes: 
[3976.zip](https://github.com/endless-sky/endless-sky/files/2569943/3976.zip)

KiLEdEnNis:
why add random floating heavy lasers just to "MaKe UsE oF tHe MoRe GuN pOrTs"? you can add more ports, but that doesn't mean you have to fill them for the default loadout

your PRs usually end up like some random mess because you do your own thing after getting and ignoring feedback, which doesn't get them merged; same for this one

so, like, listen more to what the rest says and cooperate if you want it merged, give and take

Nescio0:
Yes, I edit my proposals; if that makes them messy, so be it. Ideas evolve in conversations. Different people have different ideas.
* Lineth mentioned the Kestrel's flimsiness, therefore I've increased its hull and shields
* Amazinite considered a drag reduction to 9.8 too severe, thus I raised it
* He also suggested to increase, double or perhaps treble the number of gun ports, so I've added six more slots
* You objected to the use of heavy laser, and I've now reverted it

Nescio0:
And here's a screenshot at maximum zoom:
screenshot from 2018-11-12 10-24-16

Nescio0:
And here's a screenshot of the Kestrel with twelve lasers at maximum zoom:
d3976

Zitchas:
So many lazors.... 👍

Amazinite:
I ran a few more polls on the Discord server regarding the Kestrel.
image

And here's some of the feedback I got.
image
(From Kiko)
image

So TL;DR:
* There's a split between adding no more gun ports or tripling it, but more people would like to see the Kestrel gain gun ports than would not.
* Most people use the shield Kestrel, followed by the engines then the weapons Kestrel.
* People would like to see each variant of the Kestrel become more unique form one another.
  * Suggested unique changes to each variant include:
    * Weapons:
      * More gun ports (and possibly more turrets as well)
    * Engines:
      * Reduced drag (compared to other variants)
      * Built-in afterburners
    * Shields
      * Increased outfit/cargo space
      * Increased bunk space
      * Built-in shield regeneration

My suggestions based off of this feedback.
* Aside from the drag, all the other changes you made can stay
* Raise the base Kestrel drag to 11.5 and put the engines Kestrel drag to something like 10.4 (about 10% faster than other Kestrels with the same engines)
* Possibly give the engines Kestrel a built in afterburner (I'd just use the existing afterburner effects, no need to make new ones)
* Give the weapons Kestrel 18 gun ports (of those that want more guns, more want 18 than 12); I'd leave turrets alone and not change the outfit space
* Give the shields Kestrel rather efficient built in shield regen
  * I think we should think of something else to make the shields Kestrel unique as well, because 18 guns, a built in afterburner, and built in shield regen, that last one seems like the last I'd choose. The shields Kestrel is already the most popular though, so it may not be necessary.

DingusShingleton:
Shield piercing resistance comes to mind. I think these all sound like awesome ideas. Giving the variants something tangible rather than just boosted shield or outfit space will really make the Kestrel feel special imo

tehhowch:
Shields Kestrel is most popular when compared to Kestrel with +weapon space and +engine space. Shields Kestrel vs 18 (i vote 15 or 16, i.e. compromise between the 12 & 18 options) ports or blazing speed is a very different comparison.

The topic of damage resistances (ion, disruption, slowing) hasn't even been broached, and are not currently conveyed in the ship's info panel. (They are shown in the description of an outfit, if that outfit has those attributes.) While it will be hard for players to discern that these attributes are even present (text can be used to achieve this in the relevant mission conversation), the right value can provide appreciable worth. Perhaps each variant gets a resistance - `slowing` for Engines, `ion` for Weapon, and `disruption` for Shields.

DingusShingleton:
I think it can be conveyed pretty easily in the text when you are choosing which variant you want to unlock. And +1 to those resistances

Amazinite:
I'd be fine with 16 gun ports, seeing as how that's exactly double the current max.

Kiko also suggested that given how Tarazed are in such close proximity to the Quarg and are studying/emulating them (see the Wayfarer planet/spaceport descriptions), the Kestrel could perhaps have some advancement in ship making, some "brand new advanced technology," that was developed from observing the Quarg. Such tech *could* be damage resistances. (Although, in my head that would only be semi-canon given that 1. damage resistance hasn't been established in lore yet and 2. how could or why would Tarazed give a ship resistance to a damage they've never encountered?)

Nescio0:
More enlarged screen shots to compare:

6 heavy lasers, one per slot:

12 heavy lasers, two per slot:
2x6
18 heavy lasers, three per slot:
3x6
24 heavy lasers, four per slot:
4x6

6 lasers, at 22 pixel intervals:

12 lasers, 11 pixel intervals:

18 lasers, 7 pixel intervals:

24 lasers, 5 pixel intervals:


(Personally I think placing gun ports on top of each other looks bad. My first reactions are 'why do my laser guns have a different colour than my laser turrets?', 'why are their beams thicker?', and 'why do I only see six, I have twelve guns?'; afterwards I conclude 'oh, they're probably located on top of each other, what a pity.` 
Also, I usually play at 25% zoom, not 200%, and at minimum zoom you won't count individual beams, but you'll immediately see a colour difference.
I'd prefer each gun to be on a separate spot; those screenshots look visually more pleasing, especially the 12.)

tehhowch:
>2. how could or why would Tarazed give a ship resistance to a damage they've never encountered?

Sounds like we could add a follow-up mission to Tarazed, in which a specific piece of tech with appropriate characteristics is fetched, and which results in the associated resistance being added to the Kestrel. That development could be relegated to a separate PR (since it would be adding parity for each variant anyway).

Re overlapping ports - any increase in port positions requires a new thumbnail / sprite (no more "lasers from nowhere" after hardpoint sprites were added). So feel free to not overlap, and include the associated sprite updates in this PR. It won't merge without them.

Nescio0:
> (of those that want more guns, more want 18 than 12)

If the poll had included options for 24 or 30 gun ports, I'd expect a lot of those who now voted for 18 would have voted 30. However, 44% of those who voted think 6 is enough. Personally I'd prefer 6 too, but because there appears to be much enthousiasm for more guns, doubling the default to 12 seems a reasonable compromise.

> Possibly give the engines Kestrel a built in afterburner 
> Give the shields Kestrel rather efficient built in shield regen 

Pug and Remnant ships have a lot built-in stuff, however, the Kestrel is a human ship:
> Technically speaking it's not an alien ship; it's a human design from an alternate universe

Perhaps instead of integrated stuff, we could give the engines variant an ionic afterburner and the shields variant a S-970? 

> Perhaps each variant gets a resistance - slowing for Engines, ion for Weapon, and disruption for Shields.

This sounds really interesting. How do resistances work, though? Similar to scan interference?

Nescio0:
> Tarazed and the Syndicate do little more than simply tolerate each other.

Tarazed does purchase Syndicate systems without which they wouldn't be able to build any ships (`data/free worlds start.txt`, line 416), so in case of the Checkmate branch, it wouldn't be inconceivable Tarazed would purchase the most advanced Syndicate technology to produce their most advanced Starship; and in case of the Reconciliation branch, those would be Kestrel-only outfits: Tarazed is "a maker of alien-inspired ships *and* outfits" (`data/map.txt`, line 27217); so there are lore possibilities.
Anyway, I'm just wondering if we could do something else to make the engines variant more attractive.

Amazinite:
>so in case of the Checkmate branch

Having the loadout of the Kestrel change based off of the branch you chose (and assuming you've even finished the campaign yet) is more complexity than necessary. And if the Kestrel had these outfits regardless, why ever do the Checkmate branch when you have a supply of their outfits from Tarazed?

Amazinite:
As is, I think these engine and weapon Kestrel changes will do, but I'm iffy on keeping the shields Kestrel at this point just because it's the most chosen one. I'm up for keeping it like this though and seeing where the chips fall.

data/kestrel.txt
```diff
@@ -317,16 +323,35 @@ ship "Kestrel"
...
+	gun -53 61
+	gun 53 61
+	gun -53 61
+	gun 53 61
+	gun -31 66 "Particle Cannon"
+	gun 31 66 "Particle Cannon"
+	gun -31 66
+	gun 31 66
+	gun -31 66
+	gun 31 66
```
Amazinite: You also need to add the turret coordinates here, as changing any single weapon coordinate with a variant overwrites all previous weapon coordinates.
I'd also reorder these gun ports so that it goes 
```
-75
-75
-75
75
75
75
...
```
instead of 
```
-75
75
-75
75
...
```

so that it looks like this 
image
instead of this
image

tehhowch: its... beautiful 😍

tehhowch: it's... beautiful! 😍

Nescio0: Done.

Nescio0:
One more change, I've added two tactical scanners (effective range 800) to the default Kestrel, because it's a sophisticated warship. 

> As is, I think these engine and weapon Kestrel changes will do, but I'm iffy on keeping the shields Kestrel at this point just because it's the most chosen one. I'm up for keeping it like this though and seeing where the chips fall.

Something else I thought of is that the shields variant could have +1.0 scan interference, giving it a 50% chance to block a hostile scan; however, there is currently no way of displaying it (hence #3999).
Perhaps it's best if you implement the current proposal and organize a follow-up poll in a month or two, to see how these changes affect the popularity of each variant.

PR #3977, opened 2018-11-01
Bounder –5 weapon capacity

Nescio0:
The Bounder is not designed for combat; a weapon capacity at 45 differentiates it from the Arrow's 50 and the Scout's 40; it is large enough for the Bounder to equip two chameleon anti-missile or two laser turrets.


KiLEdEnNis:
> a weapon capacity at 45 differentiates it from the Arrow's 50 and the Scout's 40

...and? why nerf a ship that's not used that much just to make it more unique?

PR #3979, opened 2018-11-01
Headhunter +10 weapon capacity

Nescio0:
This differentiates the Headhunter from the Quicksilver's 60 and allows equipping e.g. two heavy laser cannons plus one heavy laser turret, or two typhoons and no turret.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -1803,7 +1803,7 @@ ship "Headhunter"
 		"drag" 2.6
 		"heat dissipation" .8
 		"fuel capacity" 400
-		"cargo space" 40
+		"cargo space" 50
 		"outfit space" 250
 		"weapon capacity" 70
 		"engine capacity" 80
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -1803,9 +1803,9 @@ ship "Headhunter"
 		"drag" 2.6
 		"heat dissipation" .8
 		"fuel capacity" 400
-		"cargo space" 50
+		"cargo space" 40
 		"outfit space" 250
-		"weapon capacity" 60
+		"weapon capacity" 70
 		"engine capacity" 80
 		weapon
 			"blast radius" 44


endless-sky:
The whole point of the Headhunter is that it's got a balance of cargo capacity and weapons, making it a good second ship. In my opinion reducing its cargo space just makes it *more* similar to the Quicksilver, which is meant to be a combat-only ship instead of a jack-of-all-trades.

Pointedstick:
If there's anything that would improve the Headhunter as a jack-of-all-trades second ship, it would be increasing the gun ports to three or four with no other changes.

Nescio0:
Even if the Headhunter's cargo space is reduced from 50 to 40, it's still four times as high as the Quicksilver's 10.

KiLEdEnNis:
> Even if the Headhunter's cargo space is reduced from 50 to 40, it's still four times as high as the Quicksilver's 10.

...and?
it still makes it less good all-round and more focused on 1 thing, this PR barely does what you want it to

Nescio0:
How about only +10 weapon capacity then? According to its description, one of the Headhunter's purposes is "for bounty hunting". With 70 capacity it can equip a heavy laser turret and two heavy laser cannons or sidewinders, making it more effective at hunting down (pirate) interceptors such as the Berserker.

KiLEdEnNis:
personally I'd like that, loosely compare it to a Gunboat; although that's a Navy ship, has more durability etc. it has the same hardpoints but 40 more weapon capacity, and only 20 more outfit space, even same cargo space
giving the Headhunter 70 instead of 60 space would make it more attractive
a reason not to would be comparing it to similarly-priced human ships (not that there are much of its category) like the Quicksilver, but if anything I'd give that ship some more cargo space too

so it depends if others'd deem it as "unnecessary since the ship is good enough", but it's nice to have

Pointedstick:
I could get behind this now. Might wanna adjust the variants a bit or add a few more heavily-armed versions.

Off-topic here, but what I'd still really like is an Endless Sky version of the EVO Arada. I was hoping that the Headhunter could be that, but it just doesn't have enough hardpoints. I think a really interesting ship would be something that's fairly inexpensive with two turret mounts and four gun ports, but a very limited weapon capacity: like 60 tons max. Basically it could be a super-heavy fighter or flexible light gunship. Now //that// would be the ideal bounty hinting ship IMHO.

Amazinite:
I'd be down for giving the Headhunter [the interceptor treatment.](https://github.com/endless-sky/endless-sky/pull/3350) 4 guns/1 turret instead of 2/1 would certainly open it to more interesting options, e.g. 4 Meteor missiles + an AM turret. You know what, I'm sold on the idea now.

Nescio0:
Wouldn't that make Lionheart's Headhunter more similar to Lionheart's Raven, which has four gun ports?

Amazinite:
The Raven has 40 more weapons space and no turret. The type of weapons you’re going to use on a 4/1 Headhunter will be different than on a 4/0 Raven.

Nescio0:
The default Raven currently has two heavy laser guns and two sidewinders, which means 32/90 weapon capacity free. The default Headhunter currently has two heavy laser guns and one heavy anti-missile turret, which means 0/60 weapon capacity free. Your suggested Headhunter would have four meteors and one light anti-missile turret, leaving 4/60 weapon capacity free. I think with such a configuration the cheaper Headhunter could defeat the larger Raven.

Nescio0:
If you think the Headhunter ought to have two additional gun ports, feel free to include that in #3350.
Personally I'd prefer to focus this one on the increase of the Headhunter's weapon capacity from 60 to 70.

Amazinite:
I think the changes done in #4014 are sufficient enough to differentiate the Quicksilver and Headhunter more.

Pointedstick:
Agreed.

PR #3981, opened 2018-11-01
Introduce "Rainmaker (Javelin)" variant

Nescio0:
The Rainmaker's purpose, to my understanding, is to overwhelm the opponent's anti-missile systems with a barrage of missiles. The weapon most suitable for this is obviously the Javelin, which has a far lower reload time than any other missile weapon. Javelins also tend to inflict slightly more damage than Meteors or Sidewinders, have no homing and thus no annoying missile boat behaviour, and are less likely to be intercepted because of its relatively short range.
Besides, the missile pods on the Rainmaker's wings visually resemble javelin pods.

### DIFF
--- a/data/fleets.txt
+++ b/data/fleets.txt
@@ -1074,18 +1074,25 @@ fleet "Small Republic"
 	cargo 0
 	personality
 		heroic opportunistic
-	variant 6
+	variant 3
 		"Rainmaker" 2
-	variant 5
+	variant 3
+		"Rainmaker (Javelin)" 2
+	variant 3
 		"Gunboat"
 		"Rainmaker"
+	variant 2
+		"Gunboat"
+		"Rainmaker (Javelin)"
 	variant 4
 		"Frigate"
 	variant 3
 		"Gunboat" 2
 	variant 1
 		"Rainmaker" 3
-	variant 6
+	variant 1
+		"Rainmaker (Javelin)" 3
+	variant 5
 		"Gunboat"
 
 fleet "Large Republic"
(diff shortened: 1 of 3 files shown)


tehhowch:
Note that the Mk I rainmaker's Energy Blasters mean that it also has no missile boat behavior.

I don't think wholly eliminating the use of Sidewinders from Rainmakers is appropriate.

Pointedstick:
This change would alter their battle role substantially, transforming them into short-range brawlers rather than essentially space artillery.

Amazinite:
I think a Javelin variant is fair to add, but to remove long range tracking missiles from all Rainmakers could very much change how battles against the Navy go. Javelins might be better at "overwhelm the opponent's anti-missile systems" on paper, but how good are they really when that means getting up close and personal with much larger ships, likely leading to the death of the Rainmaker and therefore the large ship no longer being distracted/overwhelmed by missiles.

Amazinite:
Why does the Javelin variant have atomic engines but is seen in Navy fleets at the same time as the Mark I variant, which uses ion engines?

Amazinite:
Why remove the radar jammer? And this thing has 50 tons of extra outfit space (without the radar jammer), so perhaps up the batteries and shields a bit so that this variant can survive better up close.
\+ Radar jammer
LP036a -> LP072a
D14-RN -> D41-HY

Nescio0:
Replaced the engines with ions, replaced the D14 with a D41, and added a large radar jammer. 
Why the batteries? Javelin Pods hardly require energy.

Pointedstick:
I don't think this is actively bad anymore, but I don't really see the point. As has been pointed out, the Navy already has a short-ranged brawler: the Gunboat. In a lot of ways it's better suited to that with its better shields. The Rainmaker seems to be designed in every way to be a flying howitzer, and I think it should stay that way.

Better luck next time. :)

PR #3982, opened 2018-11-01
Equip Firebird with particle cannons

Nescio0:
This patch equips the default Firebird with particle cannons, because:
* this makes the Firebird somewhat more effective
* Lovelace is quite close to Betelgeuse, whereas Deep Sky is much farther away
* the Firebird is basically a small Leviathan, which also has particle cannons (and quad blaster turrets)
* differentiates the Betelgeuse's Firebird from Lionheart's Corvette, which has four heavy laser cannons and two heavy laser turrets (cf. #3411)

### DIFF
--- a/data/fleets.txt
+++ b/data/fleets.txt
@@ -446,6 +446,9 @@ fleet "Large Northern Merchants"
 		"Berserker (Afterburner)" 2
 	variant 10
 		"Freighter" 2
+		"Firebird (Laser)"
+	variant 10
+		"Freighter" 2
 		"Firebird (Plasma)"
 	variant 10
 		"Freighter" 2
@@ -453,6 +456,9 @@ fleet "Large Northern Merchants"
 	variant 10
 		"Freighter" 3
 		"Firebird"
+	variant 2
+		"Freighter" 3
+		"Firebird (Laser)"
 	variant 2
 		"Freighter" 3
 		"Firebird (Plasma)"
@@ -574,6 +580,9 @@ fleet "Large Northern Merchants"
 	variant 1
 		"Star Queen"
 		"Firebird" 2
+	variant 1
+		"Star Queen"
+		"Firebird (Laser)" 2
 	variant 1
 		"Star Queen"
 		"Firebird (Plasma)" 2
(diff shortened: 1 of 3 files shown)


Lineth:
Less boring Heavy Laser spam is always a good thing.

endless-sky:
It seems like if you do this, it'd make sense to add the lasers Firebird back in as a variant, even if it's rarer. Or, maybe make the Plasma Firebird the stock model, and have both lasers and particle cannons be variants. (The Firebird is supposed to be an outdated ship model, so it'd make sense for the stock model not to have cutting edge weapons.)

Pointedstick:
This is also done in https://github.com/endless-sky/endless-sky/pull/3411, for exactly the same reasons given. :)

Nescio0:
The Leviathan is also considered an outdated design, yet it has particle cannons by default. Making the plasma firebird the default seems a bad idea, because plasma cannons originate in a different part of space (Kraz in the Rim). Actually the particle cannon is not really a "cutting edge" weapon; sure, it has nice range, but the smaller plasma cannon inflicts clearly more shield damage and has heat damage as a nice bonus:
screenshot from 2018-11-05 11-30-10

Anyway, I've added a "Firebird (Laser)" variant, as requested, and inserted it into some appropiate fleets.

Bladewood:
Particle cannons aren't quite "cutting edge", but the way they're presented in-game feels a lot more modern and sophisticated and clean to me (from the direction of the Paradise Worlds, who are supposed to be unimaginably rich; not nearly as much rust as on the plasma cannon's sprite).
You're right on the part where plasma cannons are from the opposite end of the galaxy, though.

Amazinite:
Seeing as how this would conflict with the changes made to the Firebird in #3922, I'd say that this could be closed, @Pointedstick could look at removing all heavy lasers from the default Firebird in #3922 (e.g. remove that one heavy laser turret and replace it with a quad blaster, perhaps), and then he could add the laser variant of the Firebird to his PR and place it into fleet accordingly.

Pointedstick:
Sounds good, I can do that.

Nescio0:
Or, alternatively, remove the Firebird changes from #3922 and commit this one (#3982) and #3411.

Amazinite:
Seeing as how #3922 predates this one and the changes to the Firebird their dealt with giving it missiles to fit in with the missile balance changes, I don’t think removing those changes is necessary.

Amazinite:
Handled by the merging of #3922

PR #3984, opened 2018-11-05
Assassination Missions

czartrak:
My new PR for assassination jobs, since I messed up the last time


data/pirate jobs.txt
```diff
@@ -2605,7 +3024,7 @@ mission "Hunted by State [0]"
 		system
 			distance 5 5
 		fleet "Bounty Hunters"
-
+
```
MCOfficer: in hindsight: is this supposed to be here?

PR #3986, opened 2018-11-08
Add issue templates

tehhowch:
Leave some feedback for MZ, Amazinite, and myself to help improve these documents. I'm not sure about the feature request template, but it sure seems common that we need to ask what OS and game version a bug report applies to.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,32 @@
+---
+name: Bug report
+about: Create a report to help us fix the game
+
+---
+
+**Describe the bug**
+[A clear and concise description of what the bug is.]
+
+**To Reproduce**
+Steps to reproduce the behavior:
+1. Go to '...'
+2. Click on '....'
+3. Scroll down to '....'
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**System (please complete the following information):**
+ - OS: [e.g. Windows]
+ - Game Source: [e.g. Steam beta]
+ - Version: [e.g. 0.9.8 or commit hash cab74fd3]
+
+**Link to save file**
+[e.g. pastebin .com]
+
+**Additional context**
+Add any other context about the problem here.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**System (please complete the following information):**
+ - OS: [e.g. Windows]
+ - Game Source: [e.g. Steam beta]
+ - Version: [e.g. 0.9.8 or commit hash cab74fd3]
```
MCOfficer: perhaps it would be useful to mention how to get commit hashes from nightly build numbers  (go to http://mcofficer.me:8080/job/endless-sky-win64/<build number>/ and look at the build name), as a lot of people simply write "latest nightly", which is pretty useless 2 weeks later

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Link to save file**
+[e.g. pastebin .com]
```
MCOfficer: just gonna suggest some non-cluttered services like hastebin or ghostbin and most importantly gists. pastebin isn't exactly well-suited for code.
also, is that space intentional?

.github/ISSUE_TEMPLATE/bug_report.md
gunqqer: If they compile it themselfs you can use `git log --pretty=format:'%h' -n 1` to get the short hash of the current build they have

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,32 @@
...
+4. See error
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**System (please complete the following information):**
+ - OS: [e.g. Windows]
```
Fzzr: Recommend asking for more detailed OS info, like Windows 10 64-bit or macos 10.14 etc.

.github/ISSUE_TEMPLATE/bug_report.md
tehhowch: @gunqqer To be fair, that indicates the current branch state, not the branch state when they built the binary. Hopefully they are the same, but one never knows. Someday perhaps we will modify the build system to read the latest hash, whether or not the source files are different from said hash, and then insert that into the appropriate 'version' response of the compiled binary, so that `./endless-sky -v` will be accurate

PR #3665, opened 2018-03-22
[On Hold] Expanding the Quarg as a faction

Amazinite:
A rebalancing and expanding of the Quarg has been on people's minds for a while now. After the work done in #3536, we thought "why not do it ourselves?" In order to not step on Brick's toes, this PR doesn't include the outfit images for those outfits which are added by #3536, meaning that that PR must be merged if this one is to be.

The main intent of this PR is to give a base to work off of and establish the lower boundaries of T3, a lower boundary that leaves lots of room between T2 and T3 in some cases (e.g. the Quarg weapons are both efficient and powerful, leaving room for T3 weapons that may be powerful but not efficient or vice versa). It leaves open the possibility of many more Quarg ships or weapons of larger sizes, especially as some of the larger outfits in the PR are not in use by the three Quarg ships. (The Quarg can always be rebalanced again as well if necessary, this is just balancing them as a T3 template based off of current T2 and T2.5 tech that is established in game already.)
yasbjfqwqoes9c305-9bdq
(From top to bottom, the new Skylark image, the new Oriole light warship scout, and the new Wardragon image.)

New stuff:
* Added two new weapons
  * Skydagger: a smaller Skylance with a quicker turn rate
  * Ranseur: a long range laser sniper
* Added different sized Quarg power generators, shields, and engines
* Added Quarg hull repair (#1551)
* Added a new Quarg ship: the Oriole, a scout warship

Outfit quirks:
* Quarg engines have reverse thrust that is 50% the power of their forward thrust (#3252)
* All Quarg outfits (minus engines) have built in active cooling that is equal to 15 cooling per ton and cooling energy at 1 energy per 5 cooling
  * Explanation: the crystals used in Quarg outfits have strange thermodynamic properties, causing them to absorb heat surrounding them, but at the cost of the outfits becoming less efficient as more heat is absorbed (i.e. they suck up more energy).
* Quarg power generators provide the highest energy/ton of any generator and also have high (but not the highest) heat efficiencies, but they also have among the highest heat/ton of any generator, meaning that Quarg power still requires massive cooling to use. (Think Korath Plasma Cores, but with less heat compared to the energy provided.)

Changes:
* Buffed Quarg AM by increasing the AM value from 25 to 30 and increasing velocity from 500 to 750 in order to make it compete better with the powerful but short ranged Point Defense Turret. (#1800, but Quarg AM is still far less powerful than the Pug T3 AM)
[...]

PR #3536, opened 2018-01-12
Skylance sprite and other quarg outfits

Brick63:
the skylance sprite is the work of several people and was received very well on the discord and was only recently finished by me 
~~while only the skylance is done if a general theme is decided upon the rest should follow quite easily~~
these remaining sprites are to give the quarg more than just question marks
to do
- [x] agree on a design
- [x] skylance
- [x] anti missile
- [x] shields
- [x] thrusters and turning
- [x] batteries
- [x] (possible) antimatter core
- [x] intrusion countermeasures

assets
[quarg outfits.zip](https://github.com/endless-sky/endless-sky/files/1679505/quarg.outfits.zip)

PR #3988, opened 2018-11-09
Introduce Quarg quantum hull repair

Nescio0:
Pug and Remnant ships have integrated hull repair and shield regenerators; Quarg ships have a separate shield generator, therefore it would make sense to have a separate hull repair as well. This proposal introduces a "Quantum Hull Repair", half the size of their shield generator, and two-thirds its cost, because hull repair is rarer than shield recharge. This proposal supersedes and reverts 9de051995dc7bb7f15a117460ee2c0243df483f2
I've also reordered the quantum shield generator near the other outfits, instead of behind the engines, and inserted a few white lines for consistency.
Because Quarg ships are practically uncapturable, Quarg outfits are not for sale, and Quarg outfits do not have sprites, this one doesn't need one either.


data/quarg outfits.txt
```diff
@@ -26,6 +26,24 @@ outfit "Antimatter Core"
...
+	"shield generation" 10
+	"shield energy" 10
+
+outfit "Quantum Hull Repair"
+	category "Systems"
+	cost 20000000
+	thumbnail "outfit/unknown"
+	"mass" 60
+	"outfit space" -60
+	"shield generation" 5
```
tehhowch:
```suggestion
	"hull repair rate" 5
```
surely you didn't mean to make the hull gen shield gen

data/quarg outfits.txt
```diff
@@ -26,6 +26,24 @@ outfit "Antimatter Core"
...
+	"shield energy" 5
```
tehhowch:
```suggestion
	"hull energy" 5
```

Darcman99:
Shouldn't it be hull repair...?

Nescio0:
Thank you for your quick replies! I've now corrected that shield typo.
This proposa can work in addition to and in combination with #3536. #3665, on the other hand, is more an overhaul that replaces the current Quarg content with similar named but different ships etc.

Brick63:
#3665 already adds in a hull repair outfit

KiLEdEnNis:
having various Quarg outfit & ship PR's open is a waste of time, since we have the one Brick listed
until that one's merged, it's better to list any idea on there to avoid a mess, and after that one's merged, people can look what to do next for the Quarg

Nescio0:
Then I'd recommend reviewing and possibly merging #3536 first; it basically adds sprites to the existing Quarg outfits. I'm not really in favour of #3665; it tries to do too many things at once, not all of them are improvements.

Anyway, I've included @Brick63's hull repair sprite into this proposal.

copyright
```diff
@@ -1,4 +1,4 @@
-﻿Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
```
jafdy: There seem to be quite a few unrelated changes in here, like this one.

tehhowch: Pretty sure this is another UTF-8 BOM getting chomped (or readded). A quick look at the format spec doesn't indicate that the BOM is required or disallowed.

Nescio0: Actually I have no idea what happened in this line.

Brick63: If you are using the desktop app, then you can deselect the top thing and it won't be in the update

Nescio0: No, I don't use desktop apps, I make edits directly in the browser. Anyone knows how to revert this line? (Or via a terminal command?)

tehhowch: Use something better than the browser. Then you can name your branches too

copyright
```diff
@@ -269,11 +269,12 @@ Copyright: Zachary Siple
 License: CC-BY-SA-4.0
 
 Files:
- images/outfit/korath?rifle*
  images/outfit/hai?rifle*
- images/outfit/scram?drive*
  images/outfit/korath?fuel?processor*
+ images/outfit/korath?rifle*
+ images/outfit/quarg?hull?repair*
```
Amazinite: This outfit wasn't derived from works by MZ.

Nescio0:
* #3536 has been committed as 87c22a9c5810fde83190b730f5310ad902dfd0e2
* a sprite in the same style has been added to this proposal
* copyright file has been updated (again)

tehhowch:
The copyright line you add is already matched by the more general one with the same license (`images/outfit/quarg*` matches `images/outfit/quarg hull repair`)

Nescio0:
You're right; reverted.

Amazinite:
If you have an issue with #3665 then comment on it, but note that I'm keeping that PR on hold for a while, as it isn't a necessary change at the moment given that the Quarg aren't the focus of any major storyline (e.g. the Coalition storyline). 

I don't see a dire need for the Quarg to have an outfit for their hull repair at this current moment since hull repair is already baked in more for gameplay purposes than completion purposes. The Quarg are an underdeveloped faction right now, hence why #3665 seeks to expand upon them, but we don't need to be expanding upon them piece by piece, PR by PR that could conflict when #3665 will eventually do it all at once. (Think of it like adding a whole new race to the game where everything is added either at once or in large chunks. That's essentially what we're doing given how little the Quarg have right now.)

PR #3990, opened 2018-11-11
Armageddon Core Balance

Lineth:
So I've long thought that the Armageddon core, far from being the best human reactor, is kinds rubbish. There's almost no reason to use it in favor of two Hai boulder reactors, which out perform-it in pretty much every way. While I understand that Hai tech is supposed to be more efficient and all-around better than their human counterparts, the extent to which the Armageddon is outstripped is a little sad. 

The problem is not its energy-to-weight ration, which IMO is fine. It's the excessive waste heat. I get the impression that it's humanities first attempt to create something like the (IMO overrated) Korath reactors, but to this extend the thing just doesn't have the stats to justify using it. I've decided a good way to make the navy's definitive heavy reactor a more attractive choice (even if just for flavor/making the free worlds campaign just a bit tougher) is to reduce the waste heat by a modest 10%. More would be overkill, less an insignificant difference.

### DIFF
--- a/data/hai outfits.txt
+++ b/data/hai outfits.txt
@@ -413,7 +413,7 @@ outfit "Boulder Reactor"
 	"mass" 90
 	"outfit space" -90
 	"energy generation" 17.3
-	"heat generation" 38
+	"heat generation" 34
 	description "The Hai designed these miniaturized fusion generators back when their territory was being raided by the Korath. In these relatively peaceful times, they have seen no need to update the design."
 
 outfit "Geode Reactor"
--- a/data/hai outfits.txt
+++ b/data/hai outfits.txt
@@ -413,7 +413,7 @@ outfit "Boulder Reactor"
 	"mass" 90
 	"outfit space" -90
 	"energy generation" 17.3
-	"heat generation" 34
+	"heat generation" 38
 	description "The Hai designed these miniaturized fusion generators back when their territory was being raided by the Korath. In these relatively peaceful times, they have seen no need to update the design."
 
 outfit "Geode Reactor"
--- a/data/power.txt
+++ b/data/power.txt
@@ -201,7 +201,7 @@ outfit "Fusion Reactor"
 	"mass" 100
 	"outfit space" -100
 	"energy generation" 18.2
-	"heat generation" 44
+	"heat generation" 40
 	description "The invention of practical fusion power near the end of the twenty-first century spurred a new era of prosperity and plenty. Centuries later, the generators have been reduced in size, but are still generally only used in capital warships."
 
 outfit "Armageddon Core"
@@ -211,7 +211,7 @@ outfit "Armageddon Core"
 	"mass" 130
 	"outfit space" -130
 	"energy generation" 24.2
-	"heat generation" 67.5
+	"heat generation" 69
 	description "The Armageddon Core was designed to meet the extreme power needs of the largest warships of the Republic Navy. One of these generators costs more than an entire fleet of smaller ships."
 
 outfit "Stack Core"
--- a/data/power.txt
+++ b/data/power.txt
@@ -210,7 +210,7 @@ outfit "Armageddon Core"
 	thumbnail "outfit/core"
 	"mass" 130
 	"outfit space" -130
-	"energy generation" 25
+	"energy generation" 24.2
 	"heat generation" 67.5
 	description "The Armageddon Core was designed to meet the extreme power needs of the largest warships of the Republic Navy. One of these generators costs more than an entire fleet of smaller ships."
 
--- a/data/power.txt
+++ b/data/power.txt
@@ -211,7 +211,7 @@ outfit "Armageddon Core"
 	"mass" 130
 	"outfit space" -130
 	"energy generation" 24.2
-	"heat generation" 69
+	"heat generation" 70
 	description "The Armageddon Core was designed to meet the extreme power needs of the largest warships of the Republic Navy. One of these generators costs more than an entire fleet of smaller ships."
 
 outfit "Stack Core"
--- a/data/power.txt
+++ b/data/power.txt
@@ -211,7 +211,7 @@ outfit "Armageddon Core"
 	"mass" 130
 	"outfit space" -130
 	"energy generation" 24.2
-	"heat generation" 75
+	"heat generation" 67.5
 	description "The Armageddon Core was designed to meet the extreme power needs of the largest warships of the Republic Navy. One of these generators costs more than an entire fleet of smaller ships."
 
 outfit "Stack Core"
--- a/data/power.txt
+++ b/data/power.txt
@@ -201,7 +201,7 @@ outfit "Fusion Reactor"
 	"mass" 100
 	"outfit space" -100
 	"energy generation" 18.2
-	"heat generation" 40
+	"heat generation" 44
 	description "The invention of practical fusion power near the end of the twenty-first century spurred a new era of prosperity and plenty. Centuries later, the generators have been reduced in size, but are still generally only used in capital warships."
 
 outfit "Armageddon Core"
@@ -210,7 +210,7 @@ outfit "Armageddon Core"
 	thumbnail "outfit/core"
 	"mass" 130
 	"outfit space" -130
-	"energy generation" 24.2
+	"energy generation" 25
 	"heat generation" 67.5
 	description "The Armageddon Core was designed to meet the extreme power needs of the largest warships of the Republic Navy. One of these generators costs more than an entire fleet of smaller ships."
 


Amazinite:
To be honest, I'd think that the Fusion Reactor needs to generate more heat more so than the Armageddon Core needs to generate less. 

Name | Energy/Mass | Heat/Energy | Energy^2 / (Heat*Mass)  | Diff from prev
-------|----------------|----------------|------------|------
Dwarf Core | 8.00 | 2.92 | 2.74 | 
Fission Reactor | 8.68 | 2.66 | 3.26 | +0.52
Breeder Reactor | 9.60 | 2.73 | 3.51 | +0.25
Fusion Reactor | 10.92 | 2.20 | 4.97 | +1.46
Armageddon | 11.17 | 3.10 | 3.60 | -1.37

* Energy/Mass: higher numbers are better, as you're getting more energy for less outfit space
* Heat/Energy: lower numbers are better, as you're then spending less extra mass on cooling
* Energy^2 / (Heat*Mass): The two above stats multiplied together (after Heat/Energy is inverted), essentially an absolute measure of how good a generator is given only energy, mass, and heat. Higher numbers are better

Tech progression in ES is partially based off of size, and the above reactors are all in a similar class (they use nuclear processes to generate energy) and in ascending order by mass, meaning that their efficiencies should also be in ascending order, preferably by similar values each time. But as you can see, the Fusion Reactor is a massive jump from its predecessor, followed by the Armageddon Core being worse than the Fusion Reactor (doesn't follow size progression) but better than the Breeder Reactor (follows size progression). A 10% decrease in heat from the Armageddon Core doesn't change this order. So either the Armageddon Core needs a massive buff, or the Fusion Reactor needs a considerable nerf at the same time that the Armageddon Core gets buffed if your goal is to make the Armageddon Core more viable.

Lineth:
I want to avoid buffing the crap out of the Armageddon because I do want the Hai Boulders to still be a viable option compared to it (or alternatively, introduce a Hai version of the Armageddon, locked behind a mission string possibly. Would help as Hai space is pretty barren content-wise), especially the comparative effort the player goes to to acquire Hai Boulders. Hopefully the heat changes help makes the difference.

Nescio0:
Here are the Hai energy generators for comparison (although it is unproblematic alien technology is more efficient than their human counterparts):
```
              cost, mass, energy, heat,   e/m ,  h/e , e^2/(h*m)
Sand        0.045 ,  24 ,  1.55,  2.4 , 0.065 , 1.55 , 0.042
Pebble:     1.05  ,  32 ,  4.5 , 11.9 , 0.141 , 2.64 , 0.053
Geode:      1.75  ,  58 ,  8.9 , 21.2 , 0.153 , 2.38 , 0.064
Boulder:    5.75  ,  90 , 17.3 , 34.0 , 0.192 , 1.97 , 0.098
```
And let's include the other human energy generators as well, sorted by size:
```
              cost, mass, energy, heat,   e/m ,  h/e , e^2/(h*m)
nGVF-AA:    0.040 ,  20 ,  1.2 ,  1.9 , 0.060 , 1.58 , 0.038
nGVF-BB:    0.047 ,  30 ,  1.9 ,  2.9 , 0.063 , 1.53 , 0.041
Dwarf:      0.8   ,  36 ,  4.8 , 14.0 , 0.133 , 2.92 , 0.046
nGVF-CC:    0.072 ,  40 ,  2.6 ,  3.8 , 0.065 , 1.46 , 0.044
RT-I:       0.3   ,  45 ,  4.4 , 18.0 , 0.098 , 4.09 , 0.024
S3:         0.36  ,  55 ,  5.0 , 12.0 , 0.091 , 2.40 , 0.038
nGVF-DD:    0.118 ,  60 ,  4.1 ,  5.7 , 0.068 , 1.39 , 0.049
NT-200:     0.57  ,  65 ,  7.2 , 15.0 , 0.111 , 2.08 , 0.053
Fission:    1.5   ,  65 ,  9.4 , 25.0 , 0.145 , 2.66 , 0.054
nGVF-EE:    0.177 ,  80 ,  5.9 ,  7.8 , 0.074 , 1.32 , 0.056
Breeder:    2.5   ,  80 , 12.8 , 35.0 , 0.160 , 2.73 , 0.059
Fusion:     5.0   , 100 , 18.2 , 40.0 , 0.182 , 2.20 , 0.083
Armageddon: 9.0   , 130 , 24.2 , 75.0 , 0.186 , 3.10 , 0.060
Stack:      4.5   , 140 , 22.9 , 55.0 , 0.164 , 2.40 , 0.068
```
If we look only at the last column (although I'm not sure that's meaningful), it is clear that the "bigger means better in all aspects" does not always apply. Nor does it have to; that would be quite boring. So if you want the most mass efficient outfit, get an Armageddon Core; if you want the most heat efficient outfit, get a nGVF-EE Fuel Cell; and if you want the most balanced one, get a Fusion Reactor. Or go Hai.

Amazinite:
e^2/(hm) is best used as a comparison stat between generators of the same class (e.g. looking at all fuel cells or all Hai reactors, not comparing fuel cells to nuclear reactors). It at the very least is useful to show that the Armageddon Core is practically useless partially due to just how efficient the Fusion Reactor is. If not looking at e^2/(hm), just looking at how energy/mass increases with mass and heat/energy generally decreases with mass but then suddenly spikes with the Armageddon Core tells you the same thing.

Bladewood:
I use an alternate metric for comparing reactor efficiencies. (I don't like Amazinite's because it often results in absurd statements like fusion reactor > plasma core > double plasma core.)

1. Choose a prototype cooler. In this case since we're mostly sticking to human reactors, we choose liquid helium cooling, the best human cooler. (The comparative results for liquid nitrogen and below should be pretty similar.)
2. Calculate the space taken up by cooling for a given reactor. (C = heat generation / cooling * cooling space)
3. Combine the cooling and reactor spaces. (M = reactor space + C)
4. Find energy / ton (E = energy generation / M).

Using this, we find that the breeder reactor has an efficiency of 7.658. The fusion reactor, 8.864. The armageddon core, 8.367. So I concur that either the armageddon core is too weak, or the fusion reactor is broken.

It doesn't fit the curve, but I think the reason for this has to do with its raw attributes more so than efficiency. Note that the armageddon core is solely used on Navy capital ships (heavy cruiser, mark II cruiser and carrier, and probably jump variants), which have a lot of space and are able to support the extreme heat with their good heat capacity. Raw power matters more for them than long, sustained combat potential. It would take about four fusion reactors to reveal a noticeable improvement over three armageddon cores, which is infeasible in-game.

Amazinite:
>I use an alternate metric for comparing reactor efficiencies. (I don't like Amazinite's because it often results in absurd statements like fusion reactor > plasma core > double plasma core.)

*cough*

>e^2/(hm) is best used as a comparison stat between generators of the same class (e.g. looking at all fuel cells or all Hai reactors, not comparing fuel cells to nuclear reactors [or human to Korath tech]).

Bladewood:
I also mentioned that it already failed when comparing plasma cores to double plasmas, for which I'm firmly convinced that the latter are superior.

Nescio0:
Let's check the Korath energy generators:
```
                  cost, mass, energy, heat,   e/m ,  h/e , e^2/(h*m)
Candle:         0.493 ,  23 ,  3.1 ,   7.0 , 0.135 , 2.26 , 0.060
Furnace:        0.925 ,  39 ,  5.5 ,  13.0 , 0.141 , 2.36 , 0.060
Inferno:        1.877 ,  71 , 10.4 ,  27.0 , 0.146 , 2.60 , 0.056
Plasma:         4.188 ,  54 , 11.9 ,  36.0 , 0.220 , 3.03 , 0.073
Double Plasma:  9.321 ,  93 , 22.0 ,  72.0 , 0.237 , 3.27 , 0.072
Triple Plasma: 20.460 , 159 , 43.6 , 154.0 , 0.274 , 3.53 , 0.078
```
So yes, Triple > Single > Double > Candle > Furnace > Inferno, apparently.

Remnant is also interesting:
```
                  cost, mass, energy, heat,   e/m ,  h/e , e^2/(h*m)
Millennium:     0.553 ,  29 ,  3.6 ,   6.0 , 0.124 , 1.67 , 0.074
Epoch:          2.123 ,  98 , 12.7 ,  19.8 , 0.130 , 1.56 , 0.083
Aeon:           4.384 , 160 , 21.9 ,  31.5 , 0.137 , 1.44 , 0.095
```
So if we only look at the last column, it is clear the three best energy generators are Boulder, Aeon, and Epoch, according to Amazinite's reasoning. 

Bladewood's algorithm, if I understand it correctly, starts with selecting cooling:
* Large Heat Shunt: 54 cooling / 24 mass = 2.25
* Small Heat Shunt: 13.5 cooling / 7 mass = 1.93
* Liquid Helium Cooler: 69 cooling / 40 mass = 1.725
* Wanderer Heat Sink: 19 cooling / 12 mass = 1.583
* Liquid Nitrogen Cooler: 29 cooling / 20 mass = 1.45
* Water Coolant System: 12 cooling / 10 mass = 1.2
* Hai Williwaw Cooling: 9.5 cooling / 8 mass = 1.1875
* Cooling Ducts: 5 cooling / 5 mass = 1.0

Let's take the Small Heat Shunt, because it's both efficient and small, so we can make a fair comparison:
* one Triple Plasma Core + eleven SHSs has an energy efficiency of 43.6/(159+11×7)=0.185
* one Double Plasma Core + five SHSs has an energy efficiency of 22.0/(93+5×7)=0.172
* one Boulder + two SHSs has an energy efficiency of 17.3/(90+2×7)=0.166
* one Plasma Core + three SHSs has an energy efficiency of 11.9/(54+3×7)=0.159
* one Fusion + three SHSs has an energy efficiency of 18.2/(100+3×7)=0.150
* one Armageddon + six SHSs has an energy efficiency of 24.2/(130+6×7)=0.141
* one Stack + four SHS has an energy efficiency of 22.9/(140+4×7)=0.136
* one Aeon + two SHSs has an energy efficiency of 21.9/(160+2×7)=0.126



Anyway, let's return to the actual changes of this proposal: a small heat increase for the Fusion or a heat reduction for the Armageddon wouldn't be bad, although I fail to see why the Armageddon's energy generation is raised; it is already very mass efficient.

Amazinite:
How about setting the Armageddon Core's heat generation per frame from 75 to 69, no other changes. If we want to use Blade's method, why not take it a step further and mix and match cooling to get the heat generation under zero without going massively over (i.e. without wasting space on extra cooling that isn't necessary for neutralizing the reactor's heat). I'd even accept near-zero positive values (say, within 5% of the reactor's heat gen) given the little difference it'd make. Example:

Generator | Cooling | Heat Gen | Efficiency
------------|---------|---------|-------
Fission | Nitrogen | -240 (-16%) | 6.64
Breeder | Nitrogen + Ducts | 60 (2.86%) | 7.31
Fusion | Nitrogen + Water | -60 (-2.5%) | 8.40
Armageddon (75 heat) | Helium + Ducts | 60 (1.33%) | 8.30
Armageddon (69 heat) | Helium | 0 | 8.54
Armageddon (71 heat) | Helium | 120 (2.82%) | 8.54

Something like 70 or 71 would also work as a more conservative change.

Amazinite:
@tehhowch @Pointedstick @Bladewood thoughts on the change as is?

tehhowch:
It's come a long way since [`27e6247`](https://github.com/endless-sky/endless-sky/commit/27e6247a4f914c919a8a7480620d14337f8e1ddd?short_path=cbbfb71#diff-cae3cf4ce9d5bc3eda3dd898ba8888b2R172) (when it was 95 and generated 2x as much energy)

Since my start in 0.9.5 I never bothered with Armageddon Cores before because of the heat output, so this is a step towards resolving that resistance. Whether or not the change is warranted or appropriate I'll leave to the others.

Janaszar:
I know, beating a dead horse...

I would shift it to <=4140 heat, and a price of ~8,000,000 credits.

<=4140 heat means a single helium cooler can handle it, which gives it a better efficiency rating compared to the Boulder when only using Helium coolers (8.54 energy/mass vs 7.98 energy/mass), but switches to be less efficient if lower cooling/mass modules are used (7.26 vs 8.30 for Cooling ducts, 7.48 vs 8.50 for Hai Willowa cooling)

I mention price, as while price will eventually become a non-issue for experienced/established players who can easily farm credits, early on there is a marked difference.  A single Boulder reactor is 5.75M, while a Armageddon Core is 9M, almost twice the cost.  Dropping the price down to 7M brings it in line with the Fusion Reactor's cost per energy, whereas right now, it stands closer to the cost efficiency of the KP-6 Solar Panel Array while far away from the star.

PR #3991, opened 2018-11-11
Added new sound for thrashers

Lineth:
Just fiddled in audacity to create a unique sound for the thrasher cannon & turret. Otherwise the rapid-fire 
PEWPEWPEWPEWPEW particle sound was obnoxiously annoying. Hopefully it passes muster. :3


Lineth:
Sadly  I was unable to add a more prominent fade out (the bombardment sound Im using is too short and simple for much, and my audacity skills are very, very rudimentary). I tried a slightly modified version to make it sound less like the bombardment weapons. Opinions?

.gitignore
```diff
@@ -38,3 +38,6 @@ ipch/
 *.vcxproj*
 *.userosscache
 *.sln.docstates
+*.depend
+*.dll
+*.layout
```
MCOfficer: those must've slipped in.

tehhowch:
Needs copyright attributions

Lineth:
Updooted copyright (hopefully correctly).
The only asset I used was the bombardment sound file that I modified in Audacity.

copyright
```diff
@@ -450,6 +450,12 @@ Copyright: Vilhelm (https://github.com/Vilhelm16)
 License: CC-BY-SA-4.0
 Comment: Derived from works by Michael Zahniser (under the same license) 
 
+Files: sounds/thrasher.wav*
```
tehhowch:
```suggestion
Files: sounds/thrasher.wav
```
The `*` glob matcher would be needed at the end if you had created e.g. `thrasher.wav1`, `thrasher.wav2`, etc. If you had made e.g. `thrasher~00.wav`, `thrasher~01.wav`, then you would attribute that as `sounds/thrasher*` or `sounds/thrasher*.wav` (since there are more than one and they match that pattern).

tehhowch:
Seems alright, perhaps a bit quiet? I'd like more feedback on that side re: current effect volume vs other common firing effects (it could be just that lasers are too loud, and there are lots of lasers in ES).

Zitchas:
It is a bit quieter, but that's one of the attractions. Whenever I'm in areas of human space with combat happening I usually crank down the volume to the point I almost can't hear hyperjumps because of how annoying continuous laser fire can get.

tehhowch:
I wonder if we can come up with something that will sound more like a minigun when fired from the thrasher turret while still having a nice Remnant-y sound when fired from the cannon

Lineth:
I made a sort of sawtooth-based pew noise for the bombard and thrasher but don't know of a good sound place to show them off. I also hopefully made a decently distorted version of the Coalition Death Bubbler AKA the bombardment cannon.

data/remnant outfits.txt
```diff
@@ -149,6 +149,9 @@ effect "point defense fire"
 	sprite "effect/point defense fire"
 		"frame rate" 17
 		"no repeat"
+	sound "remnant point defense"
```
Bladewood: Can we not just call it `point defense`?

Amazinite:
>but I tried to make it sound like a distorted version of the Coalition bombardment cannon it's all-but-explicitly based on

[You might want to see this.](https://github.com/endless-sky/endless-sky/commit/2e813777c01e36af5f6e5d35caa2056d7193432d#commitcomment-23572104)

image
image
image

>This weapon was originally developed in an attempt to reverse engineer a long-range energy projectile favored by an alien species that the Remnant encountered nearby.

The Seeker also fits the description of a long-rage energy projectile weapon favored by a certain alien species (at least favored by those we ran into).
The Pug are relatively closer to the Remnant than the Coalition as well, and there's always the possibility that the Pug could be even closer in a not yet created system.
While the Bombardment Cannon and Thrasher may be similar in size, the Inhibitor Cannon shows that the Remnant are no strangers to greatly reducing the size of weapons they find (the Inhibitor being 16 tons vs the 76 tons of the Detainer it's based off of).
And then if you look at the projectile images, this frame of the Thrasher looks very similar to a Seeker bolt, whereas the Bombardment Cannon projectile never coils like this.
thrasher comparison

Zitchas:
Pug Seeker, huh? Neat.

Zitchas:
That's an interesting sound. I like it much better than the current particle sound. I could definitely listened to it under heavy fire without cranking down the volume.

In regards to the source, there's also the fact that the Pug get around. They could show up anywhere, or have a system or two almost anywhere.

Bladewood:
I think it'd be better for it to be based on the bombardment cannon.

Thrasher's visual coiling can be justified by adding a description of the Remnant destabilizing the bombardment shot (which was likely refined by the Heliarchs over their millennium of peacetime) by adding more power, making it thrash about.
We know that the thrasher uses higher-energy particles, which means they'd be blueshifted from the BC's yellow. Being based on the seeker means that the sprite is redshifted, which isn't the case here.
And I don't think the Remnant would be willing to give up so useful a property as the seeker's perfect homing.

I also don't think there's a problem with the Remnant finding the Coalition, given that they're content to farm the Kor Sestor. Coalition technology is probably easier to understand than Pug magic, anyway.

Zitchas:
It would help explain how the coalition already know so much about humans, and presumably the Remnant could do the exact same thing as the Player - doing missions and getting a license and then just buying lots of the stuff and getting Coalition techs to install it in their ship. Having a working version of something installed into your ship by people who understand how it works probably provides a moderate advantage in reverse engineering something over just having an inactive "salvaged out of a ship in hopefully intact condition" copy of it.

Amazinite:
>It would help explain how the coalition already know so much about humans,

Already explained by the FW mission where you go to Poisonwood and talk to the governor about aliens.

Zitchas:
Fair enough. And it could be interesting to figure out how they might have gotten it. But I really want to keep "Pug magic" out of it. The Pug have too many fingers in human affairs as it is, and I don't think having them involved in every aspect of human advancement is very interesting, plot-wise.

Lineth:
Ah, sorry. Misread. Guess that's what happens when I try this after 20+ hours awake.

Lineth:
In my defense, I'm inexperienced with Audacity, and the bubbl- eh, bombardment cannon sounds just as silly. :p

Amazinite:
Well now the PDT sound seems too quite. :)

Amazinite:
@Lineth Could you make the PDT sound louder now? It should be about as loud or slightly louder than say the heavy AM sound. Before it was far too loud, but now it's far too quite.

Amazinite:
Alright, I think it's good enough now.

Lineth:
Part of me wants to take another crack at making a harsh/electronic version of the Pug/Coalition bombard cannon, but if I indulge that this'll never get merged. :p

PR #3992, opened 2018-11-11
reorder Carrier's HAM turret

Nescio0:
Placed the Carrier's first Heavy Anti-Missile turret on the first mount, to cover a greater area and protect it from full frontal missiles.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -745,8 +745,8 @@ ship "Carrier"
 	gun 25 -166 "Meteor Missile Launcher"
 	turret 0 -114 "Heavy Laser Turret"
 	turret 0 15 "Heavy Anti-Missile Turret"
-	turret 0 103 "Heavy Anti-Missile Turret"
-	turret 0 157 "Heavy Laser Turret"
+	turret 0 103 "Heavy Laser Turret"
+	turret 0 157 "Heavy Anti-Missile Turret"
 	fighter -37 -65
 	fighter 37 -65
 	fighter -46 40
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -743,10 +743,10 @@ ship "Carrier"
 	gun 25 -166 "Particle Cannon"
 	gun -25 -166 "Meteor Missile Launcher"
 	gun 25 -166 "Meteor Missile Launcher"
-	turret 0 -114 "Heavy Laser Turret"
-	turret 0 15 "Heavy Anti-Missile Turret"
-	turret 0 103 "Heavy Laser Turret"
-	turret 0 157 "Heavy Anti-Missile Turret"
+	turret 0 -114 "Heavy Anti-Missile Turret"
+	turret 0 15 "Heavy Laser Turret"
+	turret 0 103 "Heavy Anti-Missile Turret"
+	turret 0 157 "Heavy Laser Turret"
 	fighter -37 -65
 	fighter 37 -65
 	fighter -46 40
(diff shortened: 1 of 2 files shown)


tehhowch:
But now it can only inflict damage on a smaller area.

In other words, this is a preference change - both the current ordering and your proposed ordering are equally valid.

KiLEdEnNis:
either position still only covers a certain "aura" around it, i.e. you either cover all sides a bit, or 1 a lot and another not at all

comparing it to other ships which btw have several AMs doesn't validate this preference change, it's unnecessary

Nescio0:
Until 80f9f5967092b6f3590579bda1af796f37533480 a few days ago, the Carrier had only one AM turret, on the third slot. But because it now has two, it makes sense to spread them out, as is done with other large ships that have multiple AM turrets.
Because the HAM has a range of 200, the Carrier's sprite is 270×400 (which includes some empty space) and the Carrier is widest at the back, I'd recommend putting the AM turrets on the second and fourth slot.

Nescio0:
Sometimes it's helpful to visiualize things, therefore I've marked the Carrier's turret positions and drawn r=200 circles to show the area covert by a heavy anti-miisile turret:
carrierturrets

The Carrier used to have only one HAM, on the second slot (magenta), which provided imperfect although the best possible coverage. 80f9f5967092b6f3590579bda1af796f37533480 added a second HAM on the third slot (yellow). Initially I proposed to move that one to the fourth slot (cyan) and to preserve the other on the second (magenta). However, now I've made this image, I realize the Carrier's currently completely unprotected against full frontal missiles. Therefore the sensible thing to do is place its HAMs on the first (green) and third (yellow) slots.

tehhowch:
I'm pretty sure the range is 2x that, as the game draws at half size.

Pointedstick:
Cool, let's do it.

tehhowch:
Here we see the carrier firing its heavy laser turrets:
image

Heavy laser turret range is 400, 2x that of the Heavy Anti-Missile:
https://github.com/endless-sky/endless-sky/blob/ba784996fcb43aa6f1ffd7f7ff49ec1faf933a0d/data/weapons.txt#L244-L245
https://github.com/endless-sky/endless-sky/blob/ba784996fcb43aa6f1ffd7f7ff49ec1faf933a0d/data/weapons.txt#L366-L367

So even a HAM on the 3rd turret protects the bow (and the 4th just barely):
image

tehhowch:
Game draws sprites at half scale to maintain detail during affine transformations. So a 270x400 pixel image will be drawn in game at 135x200. This would have been apparent if you had launched the game and taken a screenshot at 100% Main (UI) zoom and 100% View zoom:
image
(145x206 there, and obviously could be cropped further)

This detail is mentioned in the wiki as well: https://github.com/endless-sky/endless-sky/wiki/CreatingShips#sprites

Pointedstick:
Revert if you like; I really have no strong opinion on the matter. So much ado over the placement of one turret. :)

Nescio0:
If the range is actually twice as large, then I don't really understand why the addition of the second heavy anti-missile turret was necessary.

Pointedstick:
Because it doubled the weight of anti-missile fire, which is important on large capital ships now that missiles are more powerful and numerous. I settled on the rough standard of mounting two HAMTs as a factor distinguishing capital ships from other ones.

Nescio0:
Yes, but more anti-missile turrets also means sacrificing ordinary firepower. Anyway, that's beyond the scope of this request.

PR #3996, opened 2018-11-13
Allow plugins to disable missions, events, and Person spawns

tehhowch:
This is a QoL feature for plugins, as other than instructing a player to remove the original data files from the game, there is no way to prevent missions or timed events from offering. Non-`repeat` missions can currently be disabled by flagging the mission as offered in the plugin's customized start, or in a `GameEvent` / `MissionAction` / `Conversation` that allows applying conditions. There is no other manner to disable or override a mission, and no existing workaround to disable timed events or prevent Person ships from spawning.

The `disable` keyword here becomes a root-level DataNode keyword, meaning it is used like `mission`:

```
mission "Some Mission [0]"
    ...

disable mission "Assisting Hai"
disable person
    "Power of the People"
    "Captain Nate"
disable event
    "war begins"
```

This also means any use of it applies to the game state as a whole - any pilot used while the data files with `disable` entries are loaded is subject to them (as is consistent with the converse - only data defined in loaded files is applied to any pilots used).

 - This isn't an end-all solution to plugin compatibility / plugin load support
 - There's no shorthand or wildcard syntax - you must specify the exact name of the items you want to influence, including matched capitalization
    - This could be changed to exact + prefix-matching, but would introduce a new requirement that no mission names end in `*`
 - Only Missions, GameEvents, and Persons are modified currently. For other classes (sale lists, fleets, governments, outfits, systems, planets) there are already mechanisms that support interacting with the previously defined version.

Please share thoughts, concerns, and other feedback. Otherwise, expect this to merge in ~2 weeks.


Zitchas:
That looks useful.

If a pilot is flown for a while with a plugin that disables certain elements, then the plug-in is removed, will they regain access to whatever was disabled? By that I mean: Does the disable get written to the save file at all, or is only in effect while the plug-in with it remains loaded?

tehhowch:
@Zitchas Just as with other game data, removing the definition of the item removes the game's knowledge about the item. Removing the lines that disable the data item will result in a game state in which that item is available for consumption by the player.

Note that purely time-based events (like "war begins") are added to a save only when a new pilot is created. If these events are disabled when the pilot is created, then they will not happen automatically if the `disable` lines are removed. (This is why patch missions are sometimes needed to add events that should have already happened.)
For things that are often checked during gameplay (i.e. mission offer and person spawn), they will be available immediately upon removal of the associated `disable` lines, as though `disable` had never been used to modify their availability.

In general, ES has two classes of data: data it manages (like mission templates, outfits, systems, and planets), and data the "player" manages (like instantiated missions, pending events, purchased ships, and a transactional log of changes). This keyword only affects the game-managed data, and does so by preventing them from being given to the player. Like removing a plugin, removing the disable keyword removes its effect on the managed data. (Unlike removing a plugin, removing the disable lines won't result in lots of gameplay instability or data replaced by `""`.)

In short:
 - if you play a pilot that started before you added `disable event "war begins"` to data files, "war begins" will happen because it was already in the player's save.
 - if you play a pilot that started after you added `disable event "war begins"` to data files, "war begins" will never happen unless you add it to the player's save with a patch mission after removing `disable event "war begins"` from the data files.
 - __While__ `disable event "A"` appears in the game data files, completing a `MissionAction` that gives the player event `"A"` will effectively not give that event to the player - if the event should occur while the game continues running, the event knows it is disabled (it came from the managed game state) and it will not do anything. If the game saves, that event will not write any data to the save (and thus will not exist in the save when the game next launches). 
 - __While__ `disable mission "Liberate Kornephoros"` appears in the game data files, you can never be *offered* `"Liberate Kornephoros"`. If the player already has `"Liberate Kornephoros"` active, it will remain active. Removing `disable mission "Liberate Kornephoros"` will result in regular condition checking for the offer of `"Liberate Kornephoros"`.

______

I recognize `disable` could indeed be confusing if used outside of total conversion plugins. It's possible this is not the best way to implement it either - maybe it should be something that is defined in the `start` block, and from then on only appears in the player's save. That has its own issues, largely in that it is very different from the existing plugin model.

I think perhaps ideally we would have an entirely different content model, in which the user selects the set of resources that should be agglomerated and used by the game. A set of resources (aka plugin folder) would identify the content it adds, the content it extends, and the content it cannot exist with (i.e. plugin dependencies). Until we have such a large refactoring of how ES works, I think being able to provide a TC with the install instruction set of "add folder to `plugins/`, then start a new pilot" and "remove the folder from `plugins/` before playing another pilot" is a lot better than also telling the player to remove certain text files from `data/`.

Nescio0:
This sounds like the most useful pull request since #3502, great! Hopefully both can be committed, to allow plug-in writers more freedom.

comnom:
Could this be simplified for total conversions somehow? If we think backwards, it seems simpler to _only_ load data from a specified plugin if we want a total conversion. For things other than that, I think the disable syntax makes a bit more sense.

tehhowch:
Currently there is no directory-based association for data. The load process is currently very simple (read base data then alphabetically read plugins, and load file content into game data when reading the file), so we would need to completely overhaul the loading system to support indexing the available content without loading it.

With that, we could effectively identify all the places that source a given content definition, whether that definition is "full" or "partial" (with additional changes), and create a method to let a directory say "only use my content" or "don't use these directories' contents"

I think both systems (specific entity and directory-wide disabling) are compatible, and directory content support is needed for better support in general (e.g. picking a start from the "New Pilot" loader, or choosing which plugin content to use from game UI).

comnom:
Yea, the idea was to identify a total conversion around GameData::LoadSources and just modify the the list we pass on to actually load.

source/Mission.cpp
```diff
@@ -351,6 +351,14 @@ void Mission::Save(DataWriter &out, const string &tag) const
 
 
 
+void Mission::NeverOffer()
+{
+	// Add the equivalent "never" condition, `"" != 0`.
+	toOffer.Add("has", "");
```
tehhowch: I think this needs to be re-checked now that #3128 is in

tehhowch:
>  Otherwise, expect this to merge in ~2 weeks.

lol

source/GameData.cpp
```diff
@@ -62,8 +62,6 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include <algorithm>
 #include <future>
 #include <iostream>
-#include <map>
-#include <set>
```
tehhowch, line 66: `<map>` is already included by GameData's own header.
`set<T>` is no longer used by this TU

petervdmeer (approved):
> Otherwise, expect this to merge in ~2 weeks.

Looks good to me, go for merge.

Amazinite:
Updated the CreatingPlugins wiki page with this: https://github.com/endless-sky/endless-sky/wiki/CreatingPlugins#disabling-preexisting-data

PR #3999, opened 2018-11-14
Display effective scan ranges and interference

Nescio0:
Effective scan range is 100×sqrt(scan power) and the chance to fail a scan is 1/(1+scan interference). I don't know about you, but I'm too lazy to calculate the result every time I'm considering installing scanners, so I'd prefer it if the result was displayed somewhere. That's what this proposal does.

If your ship has 1 interference plating (interference 0.5), 1 asteroid (power 40), 1 cargo (power 9), 1 cargo (power 25), and 1 tactical (power 32) scanner equipped, you'll see in outfit and shipyard just above the ship's energy and heat table:
screenshot from 2018-11-14 21-54-57
If you have no scanners, none of the above will be displayed, and if you only have tactical scanners, you'll only see the last line.

This also makes it possible (e.g. plug-in) for (alien?) ships to have integrated scanners (e.g. Albatross 4 scan interference, Skylark 100 asteroid scan power, and Wardragon 100 tactical scan power) and the player to be made aware of it.

### DIFF
--- a/source/ShipInfoDisplay.cpp
+++ b/source/ShipInfoDisplay.cpp
@@ -22,6 +22,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Table.h"
 
 #include <algorithm>
+#include <cmath>
 #include <map>
 #include <sstream>
 
--- a/source/ShipInfoDisplay.cpp
+++ b/source/ShipInfoDisplay.cpp
@@ -257,32 +257,31 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 	if(attributes.Get("scan interference"))
 	{
 		attributeLabels.push_back("chance to block scan:");
-		attributeValues.push_back(Format::Number(100.	
-			- 100. / (1. + attributes.Get("scan interference"))) + "%");
+		attributeValues.push_back(Format::Number(100. - 100. / (1. + attributes.Get("scan interference"))) + "%");
 		attributesHeight += 20;
 	}
 	if(attributes.Get("asteroid scan power"))
 	{
 		attributeLabels.push_back("asteroid scan range:");
-		attributeValues.push_back(Format::Round(100. * sqrt(attributes.Get("asteroid scan power"))));
+		attributeValues.push_back(Format::Number(100. * sqrt(attributes.Get("asteroid scan power"))));
 		attributesHeight += 20;
 	}
 	if(attributes.Get("cargo scan power"))
 	{
 		attributeLabels.push_back("cargo scan range:");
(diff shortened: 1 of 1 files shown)


jafdy:
This ~might~ will make the interface more cluttered with values. I'd like a neater way to implement this that doesn't get in the way of other (normally more important) information.

Nescio0:
Any suggestions? Personally I can't think of a better location to display this than in the shop's ship details panel.

jostephd:
Previous requests to add information to the ship stats panel have been rejected:

#3704 
#3835 

I realize that adding statistics to the panel will make it more overwhelming for new players, but _not_ adding those statistics is frustrating to veteran players, it reduces the outfitter for them to a continuous exercise in mental arithmetic. In the end, **I think we should find a way to show more stats in the interface.** It is true that more details in the UI will mean a learning curve for new players, but that learning curve is already there: new players _will_ have to eventually learn what's the range of a tactical scanner, whether that's shown in the UI or not. (Likewise for "time to overheating" from the linked issues.) I think the game will be more enjoyable if it doesn't force players to do quite so many calculations in their heads.

Amazinite:
An advanced stats page that can be accessed in the shipyard/outfitter and in the player's info panel for each ship would be viable. We already have multiple pages of preferences when it was once just one, so why not multiple pages of ship stats?

source/ShipInfoDisplay.cpp
```diff
@@ -253,6 +254,38 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 		attributesHeight += 20;
 	}
 	
+	// Miscellaneous stats
+	if(attributes.Get("scan interference"))
+	{
+		attributeLabels.push_back("chance to block scan:");
```
Pointedstick: "block scan" sounds awkward here. How about "Chance to evade scan"

DingusShingleton: Really what you are doing is spoofing the scan. You aren't blocking it or evading it. It happens to you, and returns false results. Spoof might be a bit to goofy though. Perhaps deceive?

Zitchas:
I really like Amazinite's suggestion of having an "Advanced Stats" page. 

In the shipyard, I would appreciate some kind of a button for each ship that pulls up a more complete info page (possibly including ship artwork or at least the 2x thumbnail).

And in the player info panel, having a complete info page that gets into all the gritty details about the player's current ship. All this scanning stuff, and finer details on everything else wherever possible.

source/ShipInfoDisplay.cpp
Ferociousfeind: If you don't mind my novice-level necromancy, perhaps consider "chance to fool scanners:"?

Amazinite:

Issue #2074, opened 2017-01-28
Wanderer Storyline Typos


Alkallid:
"You drop of" instead of 'off'.

Alkallid:
I might put any other typos I find here, so i I don't close that's why.

Alkallid:
"the Wanderer war machine is continuing to ramp up with remarkable speed"
Does this imply that "the Wander war machine" is itself a noun, like a concept of their technological advancement? It sounds a bit iffy.

endless-sky:
> **war machine**, _noun_: the military resources of a country organized for waging war.

endless-sky:
Fixed the first typo in 5474f75.

Alkallid:
What does "beach a breath" mean?

Alkallid:
Neeever mind. Breathing a sigh is a thing.

endless-sky:
No, that's a typo (should be "breathes a sigh of relief"). Good catch. Fixed in fe724a0.

PR #4004, opened 2018-11-15
New Navy Ships

Brick63:
Introducing the logistical marvels that are the Auxiliary and the Dropship.

 No longer do you have to take powerful warships away from the front lines to move around supplies or people, with its innovative modular construction, the Auxiliary can now handle all those tasks while the warships go and fight whatever people are afraid of currently.

 But lets not forget its little friend, the Dropship allows you to move soldiers into dangerous areas without risking the large, expensive transport in the process. 

Assets:
[auxiliary assets.zip](https://github.com/endless-sky/endless-sky/files/2804485/auxiliary.assets.zip)


data/ships.txt
```diff
@@ -176,6 +176,117 @@ ship "Arrow"
...
+	drone 0 31
+	drone 0 83
+
+
+
+ship "Auxiliary" "Auxiliary (Transport)"
+	sprite "ship/auxiliaryt"
+	thumbnail "thumbnail/auxiliaryt"
+	add attributes
+		category "Transport"
```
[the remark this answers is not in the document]
Brick63: fixed

MessyMix:
Nice model, but IMO the ships look too, er... new. Maybe more grunge is in order.

DingusShingleton:
This is awesome, glad the humble Navy Auxiliary Ship is still in the mix.

My only comment is that I maintain that the cargo variant should still have the ability to carry dropships. Just because it is not primarily a transport does not mean that having the versatility of a couple dropships would be extraneous.

To wit, I would opt for 2 fighter bays (dropships) and 2 drone bays (1 combat, 1 surveillance) on each variant, and I would bump the cargo capacity (and the bunks for that matter) of the dropship up a bit. Maybe to 20 tons and 20 bunks. That's utility right there. Space for a full squad and enough cargo for a useful payload of their gear.

Overall, really, just fluff-based tweaks, so take em or leave em, it really doesn't matter (although it is conceivable that they would make something broken in some way that I have overlooked). Either way this is really great work. Just thought I would toss out my 2 cents.

Brick63:
20 seems too high for it, and I felt like I was already pushing it as it was, I might increase it a little more. And I feel the cargo space is fine as is, since it's meant to be a personnel transport with room for a small amount of cargo that is likely the gear for the personnel.

I'll also go in and do some more work to make the ships look less new.

data/fleets.txt
```diff
@@ -1191,6 +1191,19 @@ fleet "Large Republic"
 		"Frigate" 2
 		"Rainmaker"
 		"Gunboat"
+	variant 4
```
tehhowch: Rather than modify the core `Large Republic` fleets with these variants, consider creating new fleets (`Navy Cargo / Navy Transport / Navy Support`) in which the Auxiliary and Dropship feature prominently. Then we can specifically use navy fleets that are not combat intensive when desired, and know that the (many) existing uses of `Small Republic` and `Large Republic` (and subsequent modifications by events or plugins) are not impacted.

e.g.
```
system Alnitak
    ....
    fleet "Large Navy Support" 8000
    fleet "Small Navy Support" 4000
    ....
```

Brick63: yeah, i had planned on doing that eventually, but put them there for now to get the fleets recorded and able to be adjusted

Amazinite:
I’m still of the opinion that this shouldn’t even appear regularly in Navy fleets across the galaxy. 

@Pointedstick the reason for three versions of the same ship is for added flexibility. It allows us to give the Navy a powerful freighter and a powerful transport ship without both being the same ship (and therefore being OP). If we only had one version of this ship, it would have to be the hybrid version, which doesn’t actually excel amazingly in either field (although it is better than the existing Navy ships). If anything, I’d say it would be reasonable to remove the hybrid variant.

Pointedstick:
Yeah, how about just two, then: a dedicated troop transport, and a dedicated cargo ship/tank landing ship type thingy.

I would agree that this thing should not show up all that often. I think it would be okay to be in some Navy fleets, but always with escorts, and it should definitely never appear in any of the scripted major battles.

Amazinite:
This is the most up-to-date version of the Auxiliary that I had on hand.
```
ship "Auxiliary"
	sprite "ship/auxiliary"
	thumbnail "thumbnail/auxiliary"
	attributes
		category "Transport"
		licenses
			Navy
#			Auxiliary
		"cost" 16720000
		"shields" 17100
		"hull" 5900
		"required crew" 87
		"bunks" 260
		"mass" 820
		"drag" 13.3
		"heat dissipation" .45
		"fuel capacity" 900
		"cargo space" 250
		"outfit space" 720
		"weapon capacity" 160
		"engine capacity" 200
		weapon
			"blast radius" 280
			"shield damage" 2800
			"hull damage" 1400
			"hit force" 4200
	outfits
		"Sidewinder Missile Launcher" 2
		"Sidewinder Missile" 100
		"Quad Blaster Turret" 2
		"Heavy Anti-Missile Turret" 2
		
		"Fusion Reactor"
		"LP288a Battery Pack"
		"D94-YV Shield Generator"
		"D67-TM Shield Generator"
		"Water Coolant System"
		"Large Radar Jammer"
		"Ramscoop"
		
		"X4700 Ion Thruster"
		"X1700 Ion Thruster"
		"X5200 Ion Steering"
		"Scram Drive"
	engine -41 181
	engine 41 181
	gun -12 -180
	gun 12 -180
	turret 0 -75 "Heavy Anti-Missile Turret"
	turret 0 -18
	turret 0 22.5
	turret 0 73.5 "Heavy Anti-Missile Turret"
	fighter 0 -75
	fighter 0 -18
	drone 0 22.5
	drone 0 73.5
	leak "leak" 30 50
	leak "flame" 30 80
	leak "big leak" 30 30
	explode "tiny explosion" 20
	explode "small explosion" 45
	explode "medium explosion" 50
	explode "large explosion" 40
	explode "huge explosion" 10

ship "Auxiliary" "Auxiliary (Cargo)"
	sprite "ship/auxiliary (cargo)"
	thumbnail "thumbnail/auxiliary (cargo)"
	add attributes
		"bunks" -112
		"cargo space" 200
	gun -12 -180
	gun 12 -180
	turret 0 -72 "Heavy Anti-Missile Turret"
	turret 0 -21
	turret 0 31
	turret 0 83 "Heavy Anti-Missile Turret"
	fighter 0 -72
	fighter 0 -21
	drone 0 31
	drone 0 83

ship "Auxiliary" "Auxiliary (Transport)"
	sprite "ship/auxiliary (transport)"
	thumbnail "thumbnail/auxiliary (transport)"
	add attributes
		"bunks" 112
		"cargo space" -200
	gun -12 -180
	gun 12 -180
	turret 0 -73.5 "Heavy Anti-Missile Turret"
	turret 0 -17
	turret 0 20
	turret 0 79 "Heavy Anti-Missile Turret"
	fighter 0 -73.5
	fighter 0 -17
	drone 0 20
	drone 0 79
```

Brick63:
i feel it is better to keep all 3 versions, as it means players will be able to choose to buy a version that better suits them, some players might want to transport, others haul, and some may want to do both

Zitchas:
I agree with that. It also fits with the current (RW) trends in ship building where several major navies are focused on building support ships designed to allow easy production of multiple ship types from the same hull. Makes sense that the same philosophy might make it into space navies as well.

Brick63:
Ill add a fleet for Navy Logistics or support and fix the sprites up next

Brick63:
the only thing left is adding the fleets to the map, feel free to offer changes for them if you think they are necessary

Brick63:
Yeah, I may just do that because I have no real idea what I'm doing in when I step away from making the pretty pictures

data/ships.txt
```diff
@@ -177,6 +177,109 @@ ship "Arrow"
 
 
 
+ship "Auxiliary"
+	sprite "ship/auxiliaryh"
+	thumbnail "thumbnail/auxiliaryh"
+	attributes
+		category "Transport"
```
Nescio0: "Transport"? Not "Heavy Warship"?

data/fleets.txt
```diff
@@ -1233,6 +1233,142 @@ fleet "Navy Surveillance"
...
+	personality
+		heroic opportunistic
+	variant 5
+		"Auxiliary (Cargo)"
+		"Combat Drone" 2
+		"Lance" 
+		"Dropship" 
+		"Frigate"
+		"Rainmaker"
+		"Gunboat" 2
```
Nescio0: So one auxiliary is escorted by four warships?

data/fleets.txt
```diff
@@ -1233,6 +1233,142 @@ fleet "Navy Surveillance"
...
+		"Auxiliary (Cargo)"
+		"Combat Drone" 2
+		"Dropship" 
+		"Lance" 
+		"Carrier"
+		"Lance" 4
+		"Combat Drone" 6
+		"Cruiser"
+		"Combat Drone" 4
+		"Frigate"
```
Nescio0: A carrier, a cruiser, a frigate, and three (heavy) auxiliaries; are you sure this is merely logistics and not a full armada?

data/fleets.txt
```diff
@@ -1233,6 +1233,142 @@ fleet "Navy Surveillance"
...
+		"Rainmaker"
+		"Gunboat"
+	variant 3
+		"Auxiliary (Transport)"
+		"Dropship" 2
+		"Surveillance Drone"
+		"Combat Drone"
+		"Frigate"
+		"Rainmaker" 2
+		"Gunboat"2
```
Nescio0: Or five warships per auxiliary?

data/fleets.txt
```diff
@@ -1233,6 +1233,142 @@ fleet "Navy Surveillance"
...
+		"Dropship" 2
+		"Combat Drone" 2
+		"Auxiliary (Transport)"
+		"Dropship" 2
+		"Combat Drone" 2
+		"Cruiser"
+		"Combat Drone" 4
+		"Frigate" 2
+		"Rainmaker"
+		"Gunboat"2
```
Nescio0: Two auxiliaries with one heavy, two medium, and three small warships?

data/ships.txt
```diff
@@ -1164,6 +1267,49 @@ ship "Dreadnought"
...
+		"outfit space" 100
+		"weapon capacity" 16
+		"engine capacity" 40
+		weapon
+			"blast radius" 12
+			"shield damage" 120
+			"hull damage" 60
+			"hit force" 180
+	outfits
+		"Beam Laser" 2
```
Nescio0: Wouldn't one turret (blaster, laser, or anti-missile) be more suitable than two laser guns?

Amazinite:
>Are they really auxiliaries, troop ships, or military freighters?

Yes.

Pointedstick:
> heck, do they even make AAS that large irl anymore?

https://en.wikipedia.org/wiki/San_Antonio-class_amphibious_transport_dock

They are very lightly armed though.

data/ships.txt
```diff
@@ -177,6 +177,109 @@ ship "Arrow"
...
 
+ship "Auxiliary"
+	sprite "ship/auxiliaryh"
+	thumbnail "thumbnail/auxiliaryh"
+	attributes
+		category "Transport"
+		licenses
+			Navy
+			Auxiliary
+		"cost" 16720000
```
Nescio0: 16.72 M? For comparison, the Bactrian's base cost is 17.6 M, the Carrier's 15.2 M, and the Cruiser's 11.2 M.

Nescio0:
> Heavy transport ships. In a sense, they are the largest ships that the Navy use to do one of three things, transport troopsTransport supplies, or a mix of both.

It has missiles, turrets, drones, and fighters, a lot of outfit capacity, plus more shield generators than any human ship. To me it doesn't look like merely a transport.

> So I beg to differ, but putting Auxiliaries as HW is both a waste of taxpayers money, and an idiotic move by whoever decided to put a transport ship on the front lines.

Could anyone check how expensive the default Auxiliary would work out? It has a high base cost plus expensive things such as a Fusion Reactor.

> (No, this ain't an amphibious assault ship, heck, do they even make AAS that large irl anymore?) 

https://en.wikipedia.org/wiki/Mistral-class_amphibious_assault_ship (very modern, about half a billion each)
https://en.wikipedia.org/wiki/List_of_countries_by_level_of_military_equipment (amphibious assault ships are ranked below aircraft carriers but above cruisers)
Of course, Wikipedia is not completely reliable, but still.

TheMarksman-ES:
>16.72 M? For comparison, the Bactrian's base cost is 17.6 M, the Carrier's 15.2 M, and the Cruiser's 11.2 M.

Gotta squeeze as much money off of the Republic.

And make the Navy reconsider having it near the frontlines.

DingusShingleton:
It has 2 defensive missiles, blaster turrets, dropships and one combat drone. The way you phrased that was pretty disingenuous

Nescio0:
> It has 2 defensive missiles, blaster turrets, dropships and one combat drone. The way you phrased that was pretty disingenuous

Let's phrase it differently then:
```
ship: guns, turrets, drones, fighters; crew / bunks ; (base cost)
Auxiliary : 2 , 4 , 2 , 2 ;  87 / 260 ; (16 720 000)
Carrier   : 8 , 4 , 6 , 4 ; 111 / 184 ; (15 200 000)
Cruiser   : 6 , 4 , 4 ,   ;  81 / 136 ; (11 200 000)
Frigate   : 4 , 3 ,   ,   ;  21 /  44 ;  (5 200 000)
Gunboat   : 2 , 1 ,   ,   ;   7 /  12 ;  (3 200 000)
Rainmaker : 6 ,   ,   ,   ;   7 /  14 ;  (1 580 000)
```

To me auxiliaries are large but vulnerable ships, relatively cheap, used to support other military vessels. The proposed Auxiliary looks rather heavy and expensive.

Nescio0:
Also, let's compare its health with other human heavy freighters:
```
ship:            shields + hull
Auxiliary:         17100 + 5900
Behemoth:           7600 + 6300
Bulk Freighter:     4000 + 7600
Hauler II:          2900 + 5200
Hauler III:         3300 + 6700
```
And now with human heavy warships:
```
ship:            shields + hull
Auxiliary:         17100 + 5900
Bactrian:          17500 + 8600
Carrier:           21400 + 8300
Class C Freighter: 13500 + 7600
Cruiser:           19600 + 6400
Dreadnought:       18100 + 7300
Falcon:            12800 + 3700
Leviathan:         14400 + 5000
Protector:          9500 + 6500
Vanguard:           9800 + 6000
```
Yes, I realize it's a military vessel, but still, don't you think it has a lot of shields for something that's supposed to be a transport?

Pointedstick:
You're being ridiculously pedantic. This isn't a Heavy Warship, end of story. The classification is far more about intended role than it is about actual stats; giving an Star Queen 25,000 shields wouldn't make it a Heavy Warship.

Nescio0:
>  The classification is far more about intended role than it is about actual stats;

Shouldn't its stats match its role? 

If I were in charge of Navy logistics, I'd order Behemoths for cargo and Bactrians instead of Auxiliaries for everything else: similar price, similar health, lower upkeep, and a lot more capacity.

MCOfficer:
the way i see it it's not that much above the freighter or behemoth in terms of durability and combat strength - and especially the former is a single-purpose ship, while the auxiliary is used by the navy, so you can expect it to be a bit stronger in its own right.
it would take far more to actually make it a warship.

Amazinite:
>It has missiles, turrets, drones, and fighters, a lot of outfit capacity, plus more shield generators than any human ship. To me it doesn't look like merely a transport.

That’s because it isn’t “merely a transport” in the same way that the Cruiser and Carrier aren’t “merely” heavy warships. The Cruiser and Carrier are the best warships humanity has ever made, and this ship is the best transport for the same reason; it can take more of a beating and transport more (at least in terms of bunks, not quite in cargo) than any other human ship. The number of shield generators is suppose to back up the fact that it’s not made for combat, though. If an Auxiliary is ever caught alone, it’s going to be a tough nut to crack. 

>To me auxiliaries are large but vulnerable ships, relatively cheap, used to support other military vessels. The proposed Auxiliary looks rather heavy and expensive.
 
I think you’re too caught up on the name. Here was the thought process of me, Kiko, Brick, and everyone else involved with making this ship. 
* Hey, the Navy needs some sort of dedicated transport ship for cargo and crew (partially stemming from #3894, partially ideas we’ve had ever since #2074)
* Let’s give it more bunks than any other human ship so it can tell the Bactrian to screw off
* How about we make different variants for different purposes so we don’t need to make half a dozen ships to fill different roles for the Navy. 
* Okay, cool. We have a ship and an idea. What should we call it?
* Idk, Auxiliary might do. It’s not exact but it’s close enough, and who knows what definitions could change over the next 1,000 years

Just because the ship doesn’t fit the modern definition of an auxiliary doesn’t mean there are issues with it. 

>Yes, I realize it's a military vessel, but still, don't you think it has a lot of shields for something that's supposed to be a transport?

Would you not what your transport vessel to be able to ensure that what is being transported is safe?

>Shouldn't its stats match its role?

They do.

Nescio0:
> I think you’re too caught up on the name.

Perhaps. As I wrote yesterday:

> Personally I'm still a bit confused what these ships are supposed to be.

To me the Auxiliary still seems to be a Navy Bactrian, but if that's its intended purpose, then I suppose it's fine.

TheMarksman-ES:
>Yes, I realize it's a military vessel, but still, don't you think it has a lot of shields for something that's supposed to be a transport? 

I'm gonna add on to this. IRL, you could have cargo ships transport your military stuff for you, no problems.

But are you sure that'll be the case for wartime?

Food for thought.

Nescio0:
> Would you not want your transport vessel to be able to ensure that what is being transported is safe?

> I'm gonna add on to this. IRL, you could have cargo ships transport your military stuff for you, no problems.
But are you sure that'll be the case for wartime?
Food for thought.

Of course, but the Auxiliary is not supposed to fight in the frontline, whereas the Cruiser is. So if those shields are deemed appropiate for the Auxiliary, I wonder why the Cruiser has only 13% more. Also, the proposed Auxiliary has a D94-YV and a D67-TM; the Cruiser only has a D67-TM. To put it differently,  the heavy warship looks rather flimsy compared to the transport.

What I would suggest for the Auxiliary would be something along the following lines:
* reduce its base cost to c. 10 M, putting it below the Cruiser
* reduce its required crew to c. 60, putting it below the Bactrian
* increase its hull to 9000 or more; it looks like a robust design and doesn't have to be agile, so it can afford to have a lot of armour plating to protect its cargo and passengers
* reduce its shields to 12000 or less; it's not supposed to fight in the frontline (the Frigate has 8000)
* remove either the D94-YV or the D67-TM, to make its shields recharge similar to the Carrier or Cruiser; surely the space of the second regenerator can be used for something more useful
* remove the drone bays or replace them with two additional fighter bays; Boxwings, Dropships, and possibly Lances all make sense for the different configurations; combat or surveillance drones not really.

data/ships.txt
MCOfficer: that's a valid concern, but also consider their use-case:  a shuttle is for civilian use while a dropship has to expect hostile conditions, so the comparison isn't quite fair.

Nescio0: The Dropship's purpose is to drop troops, not to engage in dog-fights, therefore I don't fully understand why it has to have two fixed forward pointing cannons, the same as the Lance fighter. A small defensive turret would make more sense, or no weapons at all; cf. Boxwing.
Speaking of which, the civilian Boxwing has 800 hull, therefore I think the military Dropship deserves more than merely 200.

Wrzlprnft: I'd put some long range missiles on it.

Brick63: that would lead to it having less space to move troops, and lowering its overall speed, in its current form it can get in, drop people off while providing support if necessary, and get out, the box wing is a cargo crate with engines, and is sluggish and slow, while also being cramped and not fit for rapid movement.

data/ships.txt
Brick63: it moves things not fight things

data/ships.txt
Nescio0: > that would lead to it having less space 

How? Two beam laser occupy 16 space, one anti-missile turret 16, a blaster turret 15. And no weapons at all would mean more room for passengers, cargo, and hull armour.

Brick63: i was talking of the thicker hull, weaponry is interchangeable in this situation, but the thicker hull would slow it down and take up space

Nescio0: Boxwing + 4 outfit expansions + 3 bunk rooms: 400 shields, 800 hull, 1 crew, 13 bunks.
Proposed Dropship: 600 shields, 200 hull, 1 crew, 9 bunks.
To me the sluggish Boxwing looks both safer and more cost effective.

DingusShingleton: I'm gonna have to side with Nescio0 on this one. Looking at the boxwing and dropship, such as it is, side by side... even when the boxwing is kitted as suggested with bunks, the speed differential is not tremendous. Though the dropship has more engine space, it doesn't do much with it out of the box, so I think it should rock chipmunks stock (losing the lasers does the trick. I also think it should indeed get an HP bump, it could just be shields, to like 200h/800s for example (same as the barb). Or more. I mean this thing is relatively a sitting duck anyway.

Also, a turret would probably prove much more useful on such a ship due to its propensity to be attacked from multiple sides while moving with purpose. IE, it's going to pickup or dropoff, it doesn't have time to be pointing fixed guns at anything. Note @Pointedstick that the V-22 has a *rear mounted* gun. Though, I think it would be fine without guns. It's not going to be doing much without support of some kind.

DingusShingleton: For the record, stock, the boxwing and dropship have almost exactly the same acceleration and turning and the boxwing actually has about 15% more top speed

Fzzr:
Small suggestion: Add hand-to-hand weapons, at least to the dropship. See #4125.

data/ships.txt
```diff
@@ -178,6 +178,109 @@ ship "Arrow"
...
+	leak "leak" 30 50
+	leak "flame" 30 80
+	leak "big leak" 30 30
+	explode "tiny explosion" 20
+	explode "small explosion" 45
+	explode "medium explosion" 50
+	explode "large explosion" 40
+	explode "huge explosion" 10
+	"final explode" "final explosion large"
+	description "The Auxiliary acts as the logistical backbone of the Navy. By being built in stages the Auxiliary has the ability to act as a transport, freighter, or a hybrid of the two."
```
Pointedstick: How about this instead?

"The Auxiliary is as the Republic Navy's logistical backbone. Built using flexible modules, the Auxiliary can serve as a transport, freighter, or a hybrid of the two."

Brick63: Yeah, that sounds much better

Amazinite:
>In order to be merge-worthy, this PR needs to add the new ships to some fleets and integrate them into some missions, or else they're not actually in the game at all. The commerce-raiding Wolf Pack missions seem particularly suitable.

I'd say it could be merge worthy without being in any missions. Heck, the Solifuge and the Pond Strider made it into the game without having been added to any missions, and the Pond Strider still can't even be found in game.

As for if it is added to anything, I'd say no to common fleets. The idea was that this is suppose to be a special ship that doesn't often appear. The Navy probably just uses merchant captains to transport less important stuff, but the Auxiliary is meant to be taken out when highly important stuff is being transported. 

In terms of integrating it into missions, I'd also have the Auxiliary be present wherever the Navy has a base, e.g. pre-Kornephoros when the player is scoping out the base on New Iceland.

>Shields and hull still seem too high. This thing is almost as well-protected as a Cruiser, despite being essentially a military freighter. I'd reduce the shields to the 8,000-10,000 range.

The very fact that it is a military freighter is why its shields are so high. 8k to 10k means that it's just a civilian freighter with a Navy design, becoming only slightly more durable than a Behemoth (which has 6.3k hull, more than the Auxiliary, and 7.6k shields).

>I don't see why we have three variants. I'm not a huge fan of the idea of variants in the first place, but if this is absolutely necessary, I'd like to see just two: a cargo variant and a passenger variant. Given those, I don't see what value a balanced variant adds. It just seems like unnecessary specificity.

The Navy needs a dedicated cargo ship and a dedicated troop transport ship right now. We couldn't make one ship do both because then we just have a Navy Bactrian, hence the variants of the same ship name. We could've made two ships, but we took the easy route. :)
As for dropping the hybrid variant, I'm indifferent on it.

Zitchas:
Having three variants makes sense to me: Sometimes a force will want to transport a huge military complement, sometimes they'll want to transport a lot of supplies, sometimes they'll want to transport a moderate military force along with essential supplies (a couple APCs, maybe some emplacement weaponry to establish a forward interdiction post, maybe some heavy breaching gear to deal with fortifications designed to handle starship caliber weaponry... There's a lot of usage scenarios where a force would want to mix troops and heavier-than-normal cargo. And they'll usually want it shipped together to avoid the whole "Well the troop ship got shot down, so we successfully droped a load of APCs without the troops to go in them" problem. They'd much rather split things up so if one gets shot down, they'll still have half the troops with half the gear.

Brick63:
updated the License image and the asset folder

Brick63:
@Pointedstick the barb has a special proton cannon variant so having 3 variants of a ship that each serve different roles should be fine, and i would rather not dive into the fleets again as that was a mess last time i tried

Pointedstick:
> @Pointedstick the barb has a special proton cannon variant so having 3 variants of a ship that each serve different roles should be fine, and i would rather not dive into the fleets again as that was a mess last time i tried

I'm not a fan of that either. I would have preferred a single ship with one gun mount, one turret mount, and 28 tons of weapon space. But that's just me. :) I'll accept the community's consensus, but do we have to expose all three in the shipyard? The thing that I just don't like is the fact that there will be three new ships in the shipyard. It just feels wrong somehow, like unnecessary specificity--especially considering that you can't even buy any of them. How about adding all three variants to the game, and integrating some of the variants into missions like you're already doing (+1 for that!), but only putting the balanced one in the shipyard?

Brick63:
None of the navy ships are purchasable yet, and all 3 are there because players who go through the republic/navy campaign, whenever that is added, may want to buy the other 2 more than the hybrid, why not allow them to get a good ship for cargo and transport. players would likely start asking to be able to buy the other 2 variants, plus why would you arbitrarily lock off the other options outside of a players choice of which they want.

Ferociousfeind:
I would add the specialized (cargo/passenger) variants in more important locations, while the balanced variant is more widespread in shipyards.

Zitchas:
@Pointedstick @Brick63 

I like all three variants, and I think they should all be available to the player. that being said, I definitely see the point about having three variations of the same ship all listed there. What about an alternate mechanic? Would it be possible to have the Auxiliary be the base model, and have two "refit" outfits that are only installable on the Auxiliary, and include all the changes necessary to have the desired stats of the other versions. 

Alternatively is it possible for missions to give/remove ships from the player? If it is, then purchasing the base Auxiliary could give the player access to a mission (perhaps a job-board mission that is only available from a specific navy shipyard planet) that allows them to refit their ship to the appropriate variant. Which would actually be removing the current ship from them and gives them the appropriate one.

Brick63:
okay, trying my hand at the fleets again.

Zitchas:
Would it be possible to have the Oathkeeper ship arrive after the player has already started the battle? This is a prime instance of where an auxiliary would be used - to drop a ground attack force in an assault.

That being said, I agree that it shouldn't be in the vanguard of an attack fleet. Or alternatively, be a pair with a cruiser + auxiliary.

Amazinite:
I've heavily reworked the fleet spawns for logistics fleets.
1) At game start, logistics fleets are most common in Castor (main Navy base), then snake out to Epsilon Leonis (Deep capital), Alnitak (northern Navy base), and Achernar (Syndicate capital, going through Sol for Luna and Markab for Hephaestus along the way), showing how the Navy is transporting materials too and from various key systems in the galaxy.
2) Once the war starts, logistics fleets cut back from the Syndicate and instead begin moving supplies to the Navy bases in Tania Australis, Algorel, and Menkent. The Menkent base is described as secret during the FW campaign, so the logistics fleets are much less common there.
3) At one point the Navy puts a base in Kaus Borealis, so one of the events causes the logistics fleets to stretch down there.
4) By the campaign's end, the logistics fleets now only travel between Castor, Sol, Epsilon Leonis, Alnitak, Tania Australis, and Algorel, as the Pug invasion causes the logistics fleets to pull away from Menkent and Kaus Borealis. 

Should the logistics fleets also pull back from the Navy bases inn Tania Australis and Algorel?

Amazinite:
Anyone? 
download

Zitchas:
I'd say let's merge and let people test how it fits the more direct way.

Brick63:
@Pointedstick @Tehhowch if you have the time it would be appreciated that you re-Review this

FixItYondu:
Would the Auxiliaries completely disappear after the end of the FW campaign? If not, where would they show up and why?

Amazinite:
>My concern here is that this PR adds new fleets too existing systems, but the existing fleet frequencies are not scaled back. We already have issues about too many ships appearing too frequently.

I'll get to tweaking fleet spawns of other fleets later.

>Would the Auxiliaries completely disappear after the end of the FW campaign? If not, where would they show up and why?

Auxiliaries are used to move important Navy goods and large amounts of troops. Just because the war is over doesn't mean that work doesn't still need done. As mentioned in [this comment](https://github.com/endless-sky/endless-sky/pull/4004#issuecomment-547101741), Auxiliaries will continue to travel between the (known) Navy bases (i.e. Castor, Sol, Alnitak, Tania Australis, and Algorel) as well as Epsilon Leonis to pick up Deep tech.

Zitchas:
As far as the fleets go, I'm not really concerned with some systems being a little busier. In some respects, I think it might actually be a good thing if some systems (especially inner human space) were a bit busier. 

IRL, looking around places that have large military bases, there tends to be a lot of military traffic. So I don't find it odd that there will be more such traffic in space either. And as far as that goes, I don't think the number of military transports flying through the system should have a bearing on how many offensive military ships are around. 

So I'll reiterate my support for this being merged for 0.9.11 so that people can actually see how it fits in and make better judgments about those frequencies.

Darcman99:
I agree, would be cool if the inner systems were busier. In saying that, I also think 'out of the way' systems should be less busy.

data/human/ships.txt
```diff
@@ -1202,6 +1307,51 @@ ship "Dreadnought"
...
+		"X1200 Ion Steering"
+	
+	engine 13 26
+	engine -13 26
+	gun 2 -27 "Beam Laser"
+	gun -2 -27 "Beam Laser"
+	leak "leak" 60 50
+	explode "tiny explosion" 15
+	explode "small explosion" 5
+	description "Dropships were designed with the Navy's Auxiliary in mind. Auxiliaries often act as a forward operating base when landed on planets, so moving the entire Auxiliary isn't much of an option when a large number of troops need moved to a different sector. Thats where the Dropship comes in, a small, fast, and cramped vessel able to move considerable numbers of troops at great speeds for its size."
```
Arrow2thekn33:
```suggestion
	description "Dropships were designed with the Navy's Auxiliary in mind. Auxiliaries often act as a forward operating base when landed on planets, so moving the entire Auxiliary isn't much of an option when a large number of troops need to be moved to a different sector. That's where the Dropship comes in, a small, fast, and cramped vessel able to move considerable numbers of troops at great speeds for its size."
```

Brick63:
ignore those last few, im just trying to figure out how to remove the test file change from this, cause im not sure how it happened

Brick63:
closing this since #5108 now exists

PR #4005, opened 2018-11-16
Connect the Coalition language with the Heliarch registration mission

Fzzr:
This fixes seeing the unrecognized language hail when trying to land on restricted Coalition worlds after registering with the Heliarchs.

Question: This makes Ablub's Invention and any other restricted Coalition planets bribable - if this is incorrect, which if any are specifically meant to be restricted (like the ringworlds are)?


Fzzr:
Alrighty

PR #4007, opened 2018-11-16
grouped and corrected @Brick63's copyright

Nescio0:
87c22a9c5810fde83190b730f5310ad902dfd0e2 added two more entries for @Brick63, which were placed at some distance from the already existing existing entry of Evan Fluharty (probably because #3536 was proposed nearly a year ago and in the intervening time several of his sprites were committed).
This patch simply groups them together, orders them properly, and makes a few corrections. It ought to be uncontroversial.

### DIFF
--- a/copyright
+++ b/copyright
@@ -14,7 +14,7 @@ License: CC-BY-SA-4.0
 Files:
  images/outfit/*battery?hai*
  images/outfit/anti-missile?hai*
- images/outfit/cooling ducts?hai*
+ images/outfit/cooling?ducts?hai*
  images/outfit/dwarf?core?hai*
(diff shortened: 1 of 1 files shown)


tehhowch:
Why are you reordering the attributions of others? The ` ` to `?` corrections i understand.

Nescio0:
That's a quick reply! For consistency: almost all entries have files listed alphabetically. If you open the `copyright` file and scroll slowly downwards you'll see what I mean.

tehhowch:
Revert the changes to how others have ordered their works. Just because other persons use an alphabetical order does not mean everyone does.

Nescio0:
Reverted.

PR #4012, opened 2018-11-18
Improve UX for PlayerInfo / ShipInfo panels

tehhowch:
Closes #3703 

Fixes a bug / UX detriment where the Player Info panel could not be toggled with `i` - instead, `i` would open the mission panel. This stems from `i` being the default key associated with the INFO command.

This PR regularizes the use of `Command::INFO` to always do what the ticker says:
https://github.com/endless-sky/endless-sky/blob/d2e19f154a37c267562192847a720fd4a36e3c1b/keys.txt#L20

Additionally, for certain buttons associated with the Player/Ship info panels, (rename, park/unpark, park/unpark all), allows the use of the Shift key to perform the action. (The game ships an interface definition which makes the buttons emit the uppercase key, which keyboards generally can't do.)
I skipped including the Caps Lock key in this, since it is fairly probable that someone will have it pressed without intending to emit "uppercase" keys.

Since `ShipInfoPanel::CanDump` requires `canEdit` (i.e, if the player is on a planet, they can edit their fleet composition) to be false (no littering!), I removed the check. The binding to uppercase `P` stems from [`519c5cd58`](https://github.com/endless-sky/endless-sky/commit/519c5cd580bc9862bd69c42964c98a89f7a78e6e#diff-573158c16b111d76168129cf5c0f29a2R213), and even then `P` appears to be associated with parking and unparking ships. Since `P` is otherwise handled already, it being here isn't doing anything but make us wonder why it's there.

### DIFF
--- a/source/PlayerInfoPanel.cpp
+++ b/source/PlayerInfoPanel.cpp
@@ -176,7 +176,9 @@ void PlayerInfoPanel::Draw()
 
 bool PlayerInfoPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command)
 {
-	if(key == 'd' || key == SDLK_ESCAPE || (key == 'w' && (mod & (KMOD_CTRL | KMOD_GUI)))
+	bool control = (mod & (KMOD_CTRL | KMOD_GUI));
+	bool shift = (mod & KMOD_SHIFT);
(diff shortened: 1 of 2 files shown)


Fzzr:
Nice, looks good to me.

PR #4014, opened 2018-11-19
Tweaked Headhunter and Raven

Amazinite:
Ref: #3979

Instead of giving the Headhunter more weapon space, this PR gives the Headhunter an additional two gun ports to differentiate it more from the Quicksilver. Giving the Headhunter extra gun ports increases the versatility of the Headhunter by allowing it to use a larger number of small weapons, helping it fit the "bounty hunter" role as mentioned in the description.

Also gave the Raven (and its Marauder variants) +10 weapon space so that it is able to carry more heavy weapons to compete with bigger ships as its description suggests.

Images courtesy of @mcofficer. Assets can be found [here.](https://github.com/endless-sky/endless-sky/files/2596845/Headhunter.zip)


Pointedstick:
In general I like the idea when considered in isolation. But I do feel that @Nescio0 brings up a good point that this makes the Headhunter much closer to the Raven in terms of armament potential. The Raven's lack of a turret mount is now a significant drawback in a world of missile combat, and this change might risk making the Headhunter a no-brainer vs the Raven. Thoughts?

Nescio0:
To clarify, I'm not opposed to the addition of gun ports per se, although I'm not convinced it's an improvement either. 
Ships do not exist in isolation. If one low-end ship gains more hardpoint slors, then others might have to be changed as well.

Nescio0:
For the Raven both the Hawk approach (+10 weapon capacity, i.e. 4 guns with 100 weapon capacity, allowing four torpedoes or plasmas) and the Headhunter approach (+2 gun ports, i.e. 6 guns with 90 weapon capacity, allowing six sidewinders or heavy lasers) could work.

Amazinite:
I gave the Raven +10 weapon space, as I'd rather the Raven carry 4 heavy weapons than give it more gun ports so it's carrying 6 light weapons instead (which in my eyes would make it more similar to the Headhunter since both ships would then be light-weapon saturated.)

Pointedstick:
Maybe we should also add some new variants for these ships too?

Amazinite:
I think editing the existing variants would be simpler, as making new ones means we need to change the fleet definitions to add them and make sure all the ships still spawn at similar rates as before.

Pointedstick:
You mean modifying some of the existing variants? I'd be okay with that too.

Pointedstick:
Hmm, that doesn't make the (Strike) Headhunter any different from the old super-annoying fury with four meteors that everyone hated, which I got rid of on purpose with my missile loadout tweak patch. How about two Javelins and two Sidewinders? Or two Torpedo Launchers and two Blasters?

Amazinite:
I went with 2 Javelins and 2 Meteors, as that fits perfectly without needing to remove the AM turret.

Pointedstick:
Great!

Amazinite:
Looking at human civilian light warships after this changes, we still have a progression of the Quicksilver being bottom of the barrel followed by the Headhunter, then the Raven, then the Modified Argosy. (I'd put the Navy's Rainmaker above the Quicksilver, and the Gunboat above the Raven.) This just pulls the Quicksilver and Headhunter apart a little more and closes the rather wide gap between the Raven and Modified Argosy, so I think this PR is good to go.

Issue #3912, opened 2018-08-28
Radar jamming is essentially worthless


Pointedstick:
Consider the following exercises:

### Large ship
1. Install two Large Radar Jammers on a large ship with no anti-missile turrets, giving it a `radar jamming` value of 20 in exchange for 28 tons of space.
2. Install Sidewinder Missile Launchers on your ship.
3. Point your ship away from that ship and fire missiles continuously. Then do it again while pointed at the ship.

Observation: When pointed away, 90% of the missiles turn around, head towards the ship, and impact it. When pointed at the ship, about 99% of the missiles hit it.

### Small ship
1. Install one (or even two) Small Radar Jammers on an Interceptor-class ship, giving it a `radar jamming` value of 2 or 4 in exchange for 4 or 8 tons of space (quite precious on a ship that size).
2. Install Sidewinder Missile Launchers on your ship.
3. Point your ship away from that ship and fire missiles continuously. Then do it again while pointed at the ship.

Observation: When pointed away, 95% of the missiles turn around, head towards the ship, and impact it. When pointed at the ship, about 99% of the missiles hit it. The other ship's high speed and maneuverability make essentially no difference.

### For fun
Increase the `radar jamming` values of both jammers by a factor of 5 and try again! You still get almost exactly the same results, especially for the case where you're pointing at the target ship.

## Conclusions
- For common values of `radar jamming` (2-20), radar jammers are virtually worthless because they are not effective: almost none of the missiles fired at them will miss--even when the firing ship is pointing away from the target ship! Even with absurdly high `radar jamming` values, most of them still find their mark. There seems to be either something wrong with how the jamming is calculated, or else radar jammers need their `radar jamming` values increased practically by a factor of 10.
- Radar jamming is almost **literally** worthless when the firing ship is already pointed at the target ship. It simply doesn't matter, because the missile is already going the right direction. This would be alleviated somewhat with #1865.
- On a large ship, It is not space-effective to install radar jammers instead of some type of anti-missile turret, if there is a free turret hardpoint. If there is no empty hardpoint, filling that space with more shield generators would probably be more useful if defense is the goal.
- On a small ship, it is not worth sacrificing anything to make room for radar jammers because low values such as 2 or 4 are is not effective _at all_.

DingusShingleton:
What happens in the small ship if you don't just fly in a straight line?

Zitchas:
[...]

PR #4017, opened 2018-11-20
Halved size and strength of radar jammers

Nescio0:
Although radar jamming is not very effective, it can offer some protection on ships that can't equip anti-missile turrets. The small radar jammer, although small, still takes up precious space and is thus not very useful for drones and fighters; therefore it would be helpful to have something smaller, e.g. by halving its jamming and size. Because this would leave a large gap between the small (1 jamming) and the large (10 jamming), <s>something in between would be helpful, a medium (3 jamming), similar to but a bit better than the old small jammer</s> it would make sense to halve the large radar jammer as well.

The current situation is:
* small radar jammer: 2 jamming / 4 mass = 0.5
* large radar jammer: 10 jamming / 14 mass = 0.71

<s>The first proposal was to change this to:
* small radar jammer: 1 jamming / 2 mass = 0.5
* medium radar jammer: 3 jamming / 5 mass = 0.6
* large radar jammer: 9 jamming / 12 mass = 0.75

Hopefully someone can create a third sprite in the same style as the existing two.</s>

The current proposal is:
* small radar jammer: 1 jamming / 2 mass = 0.5
* large radar jammer: 5 jamming / 7 mass = 0.71


Nescio0:
> especially considering that radar jamming is currently pretty worthless.

The problem is not necessarily radar jamming itself, but the way homing works (#3912). A radar jamming of 1 means a 50% chance of losing radar tracking, radar jamming of 3 means a 75% chance, radar jamming of 9 a 90% chance, and radar jamming of 1000 a 99.9% chance
However, chances are normalized to five seconds; each second there is a check with a chance^0.2.
The sidewinder has a turn of 240°/s and a velocity of 840/s, which means it can offen hit the target long before the homing checks take place. Likewise, meteors can hit cold targets and torpedoes small ships when fired from up close. 
Only when missiles are fired from a long distance at a moving target there is a serious chance to lose homing.

> I could conceivably accept the change to the Small Radar Jammer though.

Let's halve both the small and the large radar jammer then.

data/ships.txt
```diff
@@ -1522,7 +1522,7 @@ ship "Fury"
 		"nGVF-BB Fuel Cell"
 		"LP036a Battery Pack"
 		"D14-RN Shield Generator"
-		"Small Radar Jammer" 2
```
Pointedstick: This would have the effect of increasing the Fury's price by 110,000 credits, which seems a bit steep for an interceptor--even a heavy one. There may be room for a large jammer now, but I'm not sure the price increase could justify it.

Nescio0: Good point. The Hawk also has two small radar jammers, so I suppose that's more than enough for the Fury. Changed.

Amazinite:
/shrug
If there's an issue with radar jamming, I think the attribute itself is what we need to look at and not the outfits that apply it.

Amazinite:
I see now. Nescio0 didn’t update the PR description properly. Looked like he only halved the size of the large. Have at it then.

Pointedstick:
@Nescio0, let's perfectly halve the cost of the small jammer. I think going down to 250k for the large one is fine; 490k was always a bit odd anyway IMHO.

Pointedstick:
I'm good with this now. Any objections?

Amazinite:
I don't see it as necessary, really. The Lance and Dagger each already fit small radar jammers, so clearly reducing the size of the small jammer from 2 to 1 isn't necessary for making them fit onto the existing fighters. Boxwings have space for jammers right now but don't get deployed anyways. The Barb doesn't have a jammer and is expectedly packed already since it's Syndicate. The Finch right now is the only other human fighter in its default loadout without a jammer and without room for one, but if it's necessary that the Finch has one then the Javelin Pod can be swapped for a Meteor Missile and a jammer can be added.

DingusShingleton:
I think you are too bogged own in default loadouts. When people are sitting there setting up their fleet, having the ability to fit jamming into 2 tons instead of 4 will make the difference many times. Maybe it's not a game changer now, but that gets back to the effectiveness of jamming being a separate issue. I think change can only add to the game at this point.

Amazinite:
Making jamming outfits too small could make using it a trivial matter though. It shouldn’t be something you *always* have space to equip, it should be something you actively consider putting into your build or not. Do you want your fighters to dodge missiles better or do you want that extra space to perhaps upgrade your power just the slightest bit so that your fighter is less likely to run out of energy while in combat and therefore pack more of a punch while it is alive?

As an aside, making jammers smaller to make it easier to customize fighters also kinda goes against the prevailing opinion in #4034 to remove almost all fighter customization. @pointedstick :)

Pointedstick:
> Making jamming outfits too small could make using it a trivial matter though. It shouldn’t be something you _always_ have space to equip, it should be something you actively consider putting into your build or not. Do you want your fighters to dodge missiles better or do you want that extra space to perhaps upgrade your power just the slightest bit so that your fighter is less likely to run out of energy while in combat and therefore pack more of a punch while it is alive? The same can be said for interceptors; jamming should be an active decision, not a trivial matter.
> 
> As an aside, making jammers smaller to make it easier to customize fighters also kinda goes against the prevailing opinion in #4034 to remove almost all fighter customization. @Pointedstick :)

Well, that proposal doesn't seem to be going anywhere :)

But I agree, if we did that, then smaller radar jammers wouldn't be as useful--but it would still be important for small non-fighter ships. I disagree that jammers shouldn't be mandatory. It makes no sense that in a world with effective radar-guided anti-fighter/interceptor missiles, any small ship would forego jamming. It's suicide.

Amazinite:
You have to remember though that only Sidewinders use radar tracking in human space, so it really only makes sense to equip radar jammers if you're expecting to run into Sidewinders. Only northern pirates use Sidewinders, so it makes sense to use jammers up north, and the FW has to deal with the Navy, so it makes sense for them to use jammers, but elsewhere it isn't as suicidal as one might think.

Amazinite:
Closing this.
If radar jamming is too weak, then either the existing radar jamming outfits should have a higher jamming attribute, or the mechanic of radar jamming itself needs to be looked at (ref: #3912). Small ships can already comfortably fit the current small jammers, so reducing their size while keeping the jamming per ton the same isn't fixing anything.

PR #4023, opened 2018-11-21
Revert some .gitignore changes

Fzzr:
This reverts a change removing xcode project settings from `.gitignore` in c76aea7294f0f76a823e3009c6e7c96600a9eb98. That commit also changed ignoring `*.exe` to just `EndlessSky.exe` - Do we definitely want to do that?


tehhowch:
There's only one .exe for windows builds - and it's the one we make.

PR #4024, opened 2018-11-22
made Ember Wormhole require quantum keystone

Nescio0:
Because it improves consistency:
* other wormholes in the Ember Waste also require quantum keystones
* the "Ember Wormhole" is currently the only wormhole with a red sprite that doesn't require a quantum keystone; other reds do, purples don't


Zitchas:
I don't think that'll cause any problems, and it would improve consistency.

PR #4036, opened 2018-11-25
Quargs → Quarg

Nescio0:
Changed three instances of Quargs into Quarg, because it seems everywhere else Quarg can be both singular, plural, and adjective; cf. Hai, Pug, or Remnant.


Bladewood:
I can't believe I never noticed this.

PR #4037, opened 2018-11-25
Korathi → Korath

Nescio0:
Changed seven instances of Korathi into Korath, because it seems everywhere else Korath can be both singular, plural, and adjective; cf. Hai, Pug, Quarg, or Remnant.

### DIFF
--- a/data/quarg missions.txt
+++ b/data/quarg missions.txt
@@ -76,7 +76,7 @@ mission "First Contact: Kuwaru Efreti"
 	on offer
 		conversation
 			`In your past experience, you've never seen the Quarg move with anything approaching haste or excitement, but when you land on this station a group of them run up to you, moving faster than you thought they were able to. For a minute or two they just gawk at you and talk amongst themselves in their own language; you think you recognize the word "Humani," and perhaps, "Eartha."`
-			`	Finally a Quarg appears who speaks your language, or at least is willing to try. It says, "Salutations, astonishing sojourner. Whither came you, and wherefore visit you us here in the unquiet graveyard of the Korathi?"`
+			`	Finally a Quarg appears who speaks your language, or at least is willing to try. It says, "Salutations, astonishing sojourner. Whither came you, and wherefore visit you us here in the unquiet graveyard of the Korath?"`
 			choice
 				`	"I'm just here to explore. Can you tell me more about the Korath?"`
 				`	"I came here to learn what happened to the Korath."`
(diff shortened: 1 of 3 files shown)


Amazinite:
Noticeably these are *all* dialog from Hai and Quarg.

Nescio0:
Perhaps, but then it is at least confusing that descriptions and human and Wanderer dialogue consistently use Korath and not Korathi. Besides, Quarg dialogue uses Korathi twice, but also simply Korath, e.g. `quarg missions.txt` lines 79-83:

Finally a Quarg appears who speaks your language, or at least is willing to try. It says, "Salutations, astonishing sojourner. Whither came you, and wherefore visit you us here in the unquiet graveyard of the **Korathi**?"
  choice
    "I'm just here to explore. Can you tell me more about the Korath?"
    "I came here to learn what happened to the Korath."
  "Ah, the unwise **Korath**," it says. "Once a mighty empire, now sadly splintered and diminished. Their own thinking war machines, now breeding with abandon and inimical to all, hunt the survivors. And of them some, too, were exiled for their great crimes."

Bladewood:
I wasn't sure how to handle that second Korath reference from the Kuwaru Efreti when I was [rewriting their dialogue](https://github.com/endless-sky/endless-sky/pull/3650) (because I am not Michael), but yes, I think this is intentional.

I think what's important is that the Hai and the Quarg have known the Korath for much longer than the rest of us (which partly has to do with their extreme age); humans and Remnant alike have only interacted with them within the past century or so, and the Wanderers haven't even discovered them yet until the player's intervention.

Bladewood:
It looks like the issue, then, is intraspecies consistency, because the Quarg both use "Korath" and "Korathi". I was under the impression that "Korathi" was supposed to be the plural form of "Korath", although then I wondered what "Korati Efreti" (the name of the station opposite the ringworld) meant.

Nescio0:
But one and the same Quarg uses "Korath" as a plural and then "Korathi" as a plural in its next sentence. Surely that's not consistent?

Nescio0:
But one and the same Quarg uses "Korathi" as a plural and then "Korath" as a plural in its next sentence. Surely that's not consistent?

Zitchas:
*People* aren't consistent, there's no reason to assume that the Quarg or anyone else are.

Amazinite:
As mentioned, Korathi is intentional given the context. The instance of the Quarg saying Korathi then Korath comes after the player saying Korath, so I see that as the Quarg mimicking the player's pronunciation.

PR #4038, opened 2018-11-25
merchant fleet corrections

Nescio0:
Remove southern merchants from Algieba, because it's an uninhabited system and none of its neighbours has a southern fleet, so it's weird they appear here out of nowhere.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -2538,7 +2538,6 @@ system Algenib
 	fleet "Small Core Pirates" 300
 	fleet "Large Core Pirates" 500
 	fleet "Large Syndicate" 800
-	fleet "Small Core Merchants" 600
 	object
 		sprite star/k0-old
 		period 10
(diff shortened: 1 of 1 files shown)


Amazinite:
While you're at it, there are northern merchant fleets in these three southern systems for some reason.
Alioth, Alphecca, and Boral should have their large northern merchant fleets swapped to southern ones.
image

Nescio0:
Thanks for pointing those out! They've now been corrected.
I've also added northern merchants to Porrima, because its only inhabited neighbour has northern fleets.

Amazinite:
>I've also added northern merchants to Porrima, because its only inhabited neighbour has northern fleets.

Porrima already has southern merchants, so I don't think it needs northern merchants as well, and the "only inhabited neighbor" argument isn't the best when it could extend forever.

Merchants could easily be exploring (I know I didn't always have the credits to spare for a local map), but if we want to reduce the chance of merchants entering such systems then that should be dealt with alongside #4039 and not here.

Nescio0:
> Porrima already has southern merchants, so I don't think it needs northern merchants as well, and the "only inhabited neighbor" argument isn't the best when it could extend forever.

Also, its planet, New Switzerland, has the "near earth" attribute, not "dirt belt". Phecda has both Northern and Southern merchants, so it would make sense for Porrima as well.

PR #4040, opened 2018-11-27
pirate fleets corrections

Nescio0:
Removed pirates from five systems (Kugel, Eteron, Fingol, Nocte, Regulus), because they are uninhabited and each one's only neighbour has no pirate fleets at all, so it's weird they appear here out of nowhere. Cf. #4038. 

[EDIT]: Also changed the pirate fleets in Fala and Tania Australis from Northern to Southern, because:
* there are no pirate fleets in Algieba or Phecda
* Delta Velorum and Mora have Southern fleets
* the only inhabited object in these three systems is a Republic dirt belt planet
* `event "initial deployment 1"` removes all pirates from Tania Australis
  * because this isolates the uninhabited system of Lolami, pirates are removed from there as well.
* these systems are closer to Shaula than to Almaaz
  
In other words, Northern pirates make little sense here, Southern do.

This basically results in the following:
* Core Pirates in the Core
* Northern Pirates in the Deep and the North
* Southern Pirates in the Dirt Belt and the South

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -9262,8 +9262,8 @@ system Fala
 	fleet "Small Northern Merchants" 1600
 	fleet "Large Northern Merchants" 3000
 	fleet "Small Republic" 4000
-	fleet "Small Northern Pirates" 2000
-	fleet "Large Northern Pirates" 4000
+	fleet "Small Southern Pirates" 2000
+	fleet "Large Southern Pirates" 4000
 	fleet "Human Miners" 2000
 	object
 		sprite star/g5-old
@@ -14178,8 +14178,8 @@ system Lolami
 	fleet "Small Northern Merchants" 2000
 	fleet "Large Northern Merchants" 3000
 	fleet "Small Republic" 5000
-	fleet "Small Northern Pirates" 2000
-	fleet "Large Northern Pirates" 4000
+	fleet "Small Southern Pirates" 2000
+	fleet "Large Southern Pirates" 4000
 	fleet "Human Miners" 3000
 	object
 		sprite star/g0
@@ -21855,8 +21855,8 @@ system "Tania Australis"
 	fleet "Small Northern Merchants" 1000
 	fleet "Large Northern Merchants" 2000
 	fleet "Small Republic" 3000
-	fleet "Small Northern Pirates" 4000
-	fleet "Large Northern Pirates" 9000
+	fleet "Small Southern Pirates" 4000
+	fleet "Large Southern Pirates" 9000
 	object
 		sprite star/g0-old
 		period 10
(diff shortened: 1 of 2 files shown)


data/map.txt
```diff
@@ -21855,8 +21855,8 @@ system "Tania Australis"
 	fleet "Small Northern Merchants" 1000
 	fleet "Large Northern Merchants" 2000
 	fleet "Small Republic" 3000
-	fleet "Small Northern Pirates" 4000
-	fleet "Large Northern Pirates" 9000
+	fleet "Small Southern Pirates" 4000
+	fleet "Large Southern Pirates" 9000
```
Amazinite: Why? These systems have northern merchants, why would they then have southern pirates?

Amazinite:
For reference.
Northern pirates:
northern pirates
Southern pirates:
southern pirates
Core pirates:
core pirates

Nescio0:
Actually the `event "initial deployment 1"` removes all pirates from Tania Australis, which means the pirates in Fala and Lolami are near Southern Pirate systems but isolated from Northern Pirates.

Nescio0:
Thanks for the maps; those pirates in Eteron and Polaris also can't come from anywhere, therefore I've removed them too.

data/map.txt
Nescio0: Because they are in the Dirt Belt, connected to systems with Southern Pirates, but not to any systems with Northern Pirates. So N pirates would have nowhere to come from or go to, S pirates would. See also first post.

Nescio0:
Restored pirates to Polaris and removed them from Regulus, per request, and Lolami, because it's isolated by the `initial deployment 1` event.

data/map.txt
Amazinite: Porrima and Regulus are the only two systems where southern pirates currently overlap with northern merchants, one of those being thanks to #4038, and this PR would remove them from overlapping in Regulus. Due to this, I'd say that the addition of northern merchants in Porrima could be reverted, and these pirate fleets should be swapped back to northern. (Notably, northern pirates don't overlap with southern merchants anywhere, and core pirates only overlap with northern merchants in Cardax.)

I don't see disconnectedness as an issue here seeing as how northern pirates are generally disconnected in this region anyways. Core and northern pirates also appear around southern pirate systems, so clearly pirates are doing something to evade detection and make their way to these systems, making them *appear* disconnected.

endless-sky:
There's several times in the story line when someone mentions that pirates like to hang out in uninhabited systems where they can attack merchants with less risk of getting caught. And, having occasional "dangerous" uninhabited systems scattered throughout human space adds a bit of interest to the game. It feels to me like you're reducing the variety and story-related complexity of the map just in the name of consistency.

Nescio0:
Closing because it seems unlikely to be implemented.

Issue #4039, opened 2018-11-26
Reduce fleet frequencies in human space across the board?


Amazinite:
#4038 got me thinking; a critique I've seen a few times about the game is that systems are often *too* populated. [Here](https://docs.google.com/spreadsheets/d/1oZJLAlqxZwkJZtqya3egwe21U3yEkGt_VM8tCk8C6cI/edit?usp=sharing) is a spreadsheet showing the average fleet frequency for each human system (prior to any events) as well as the average number of seconds between fleet spawns. (Calculated by finding the inverse of the sum of the inverses of the fleet frequencies in a system, e.g. fleet 1 is 500, fleet 2 is 1000, 1/(1/500 + 1/1000) = 333.3, the average fleet frequency of the system).

As you can see, about 85% of all human systems see a fleet entering on average once every 10 seconds or less, while about 97% of all human systems see a fleet entering on average once every 20 seconds or less. Only five systems take longer for fleets to spawn than this, and other than Ultima Thule which has no fleet spawns, no human system takes longer than 48 seconds on average for a fleet to show up.

Arguments I can think of as to why we might want to reduce the fleet frequencies:
* There isn't much of a sense of exploration when every system you go to already has a steady stream of fleets entering and exiting. (The argument I've seen for this.)
* Big battles like during the campaign don't seem so special when hanging out near a pirate planet can result in a battle just as big rather often.
* A place like Coalition space has many times more inhabitants than human space in a much denser area, but the number of ships you see feels just like human space; reducing frequencies in human space could make alien region feel more special in this way.

And to end, here's a map of human space with systems color coded by how often fleets enter the system (prior to any changes to fleet spawns due to events). Sol is perfect blue and perfect black is for systems that take 10 seconds or longer on average for a fleet to spawn. (Ignore the boxed-in systems, that's stuff from plugins.) At least in terms of how the frequency of systems compare to one another, there are at least a couple of anomalies. (e.g. Algieba, an uninhabited system, being highly populated; Limen, an uninhabited system surrounded by uninhabited systems and near inhabited systems with very low frequencies, has a rather high fleet frequency; Sheratan neighbored on either side by rather high frequency systems, has a very low frequency; the system connecting Kraz to the rest of the galaxy has a lower frequency that Kraz itself and the other neighboring system, etc.)
![image](https://user-images.githubusercontent.com/17688683/49038754-d6207680-f18b-11e8-838e-254d8d18d2d3.png)

[...]

PR #4041, opened 2018-11-28
removed large merchant fleets from uninhabited dead ends

Nescio0:
This proposal removes large merchant fleets from Kugel, Nocte, Fingol, Regulus, Lolami, Limen, Ipsing, Boral, Hintar, Orvala, Naper, and Umbral, because these are uninhabited dead-end systems leading to nowhere, so merchants have no business doing here.
Small merchant fleets are kept in these systems, because people might be exploring or too poor to purchase a map, so these merchants could then refuel the player.
See also #4039.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -5502,7 +5502,6 @@ system Boral
 	trade Metal 511
 	trade Plastic 535
 	fleet "Small Southern Merchants" 3000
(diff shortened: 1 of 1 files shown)


Pointedstick:
+1. this makes little sense.

PR #4042, opened 2018-11-28
Add thumbnail definition to the "Unknown Ship Type"

Tadrix:
So that the player sees the (already existing) thumbnail in shipyard and outfitter screens. Will also fix it for the all-content-plugin.
sidebyside

### DIFF
--- a/data/kestrel.txt
+++ b/data/kestrel.txt
@@ -213,6 +213,7 @@ event "kestrel available: more shields"
 
 ship "Unknown Ship Type"
 	sprite "ship/kestrel"
+	thumbnail "thumbnail/kestrel"
 	attributes
 		category "Heavy Warship"
 		"cost" 10300000


Nescio0:
On the other hand, the Unknown Ship Type is not for sale, so it *might* have been intentional. Drak, Pug, and Quarg ships don't have thumbnails either. It doesn't hurt, though, so I'm not against it.

jafdy:
The `Unknown Ship Type` is simply saying that you don't know what the ship hull is called, this is not surprising as at that point the ship hull has not been named.

Tadrix:
The Unknown Ship Type is a Kestrel. Kestrels already have shipyard thumbnails. As the Unknown Ship Type is, essentially, a separate ship and it gets shown, for example, in all-content-plugin, why don't follow the example of Korath ships and assign it an (already existing) thumbnail? It certainly doesn't hurt, and is beneficial for consistency.

tehhowch:
I would argue that you can simply provide the minimal definition on the all-content plugin. This ship is not available for sale in the base game, nor will it be. If some other way of exposing the thumbnail photo in-game is added, then this change could be relevant.

```
ship "Unknown Ship Type"
	thumbnail "thumbnail/kestrel"
```

Tadrix:
For consistency's sake, why *shouldn't* we add thumbnails if available?

Tadrix:
For consistency's sake, why *shouldn't* we add thumbnails if available?

Moreover, own ship thumbnails can be seen in the shipyard and the outfitter view. The "Unknown Ship Type" Kestrel can be captured, however unwise that is. Therefore, it should have the thumbnail.

Amazinite:
*checks OP*
We really need to get that World Ship fixed.

PR #4046, opened 2018-12-01
Coalition Ramscoop

10010101001:
This adds a bit of diversity to the current ramscoops and is an alternative to the Wanderer ramscoop.


10010101001:
image

tehhowch:
You should absolutely be able to upload it to the branch you are PRing from - that branch is in your fork.

Please ensure you additionally update all copyright documents, and that you provide a link to the relevant assets as appropriate with the license under which you are releasing this content.

Also, I would never, ever pay 6 energy per frame for a ramscoop bump. Did you perhaps mean 0.1 per frame?

10010101001:
Sorry, I forgot that it was per frame instead of per second. Fixed

10010101001:
Done. Not sure about the license though.

Brick63:
I'll say this again, needs paint

Darcman99:
I don't see the need for a coalition ramscoop, they have the vast majority of their space inhabited with few uninhabited systems. Those systems still get a crazy amount of traffic so it seems like a non-issue, at least to me.

This outfit also seems like yet another better ramscoop, we already recently were given the korath ramscoop that although gives very little ramscoop for its' tonnage, it also gives two fuel pods worth of fuel capacity. The only lore-wise benefit I see from this is that the heliarchs have been hinted to have jump drives in storage, so they would prepare themselves with a way to replenish fuel if they got stranded outside of their space.

10010101001:
Is this still worth working on?

10010101001:
small ramscoop
What about this

PR #4051, opened 2018-12-04
Tweaked Raven base cost

Nescio0:
The Raven currently has a base cost (1.8 M) just below the Headhunter's (1.85 M). However, the Raven is a clearly more powerful ship (100 weapon capacity instead of 60), therefore it would make sense to make it more expensive. This proposal increases it by 0.8 M. For comparison the base costs of Lionheart ships:
* 0.129 M: Dagger
* 0.18 M: Flivver
* 0.85 M: Scout
* 1.8 M: current Raven
* 1.85 M: Headhunter
* 2.6 M proposed Raven
* 3.5 M: Aerie
* 4.4 M: Corvette
* 4.58 M: Mule
* 17.6 M: Bactrian

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -1825,7 +1825,7 @@ ship "Headhunter"
 	thumbnail "thumbnail/headhunter"
 	attributes
 		category "Light Warship"
-		"cost" 1800000
+		"cost" 1850000
 		"shields" 3800
 		"hull" 700
 		"required crew" 2
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -1825,7 +1825,7 @@ ship "Headhunter"
 	thumbnail "thumbnail/headhunter"
 	attributes
 		category "Light Warship"
-		"cost" 1850000
+		"cost" 1800000
 		"shields" 3800
 		"hull" 700
 		"required crew" 2
@@ -2476,7 +2476,7 @@ ship "Raven"
 	thumbnail "thumbnail/raven"
 	attributes
 		category "Light Warship"
-		"cost" 1800000
+		"cost" 2600000
 		"shields" 4700
 		"hull" 1400
 		"required crew" 6


Brick63:
headhunter is newer, raven is old, headhunter has a turret and 4 guns, raven has 4 guns. did you account for the cost including outfits?

Amazinite:
This change would make the Raven more expensive than the Modified Argosy (3.461 mil for the Mod Argosy vs 3.517 for the Raven). If the Raven is to be made more expensive, I think that 300k extra is enough. That would mean that the Raven would fall between the Mod Argosy and the Headhunter, instead of putting it between the Headhunter and Aerie.

1. Quicksilver (~2 mil)
2. Headhunter (~2.6 mil)
3. Raven (~3 mil)
4. Mod Argosy (~3.5 mil)

(And looking at that, I'm tempted to say drop the Quicksilver hull cost to something more like 1.6-1.8 mil so that it's the most direct upgrade to an interceptor, seeing as how the Hawk is currently 1.1 mil; that would also make the Quicksilver a viable cost over the Headhunter seeing as how it would become a significantly cheaper alternative to it.)

Nescio0:
The total price including outfits is unimportant. Outfits aren't fixed, they can be removed or replaced. Whenever I purchase a new ship, I first sell all installed outfits and afterwards start installing what I want. What matters is the base cost.

Nescio0:
Here is a comparison of the human light warships sorted by weapon capacity, plus the heaviest interceptor (Hawk) and lightest medium warship (Manta):
```
             base cost , shields , hull , weapon capacity
Hawk            , 0.67 ,    2500 ,  500 ,  50
Quicksilver     , 1.09 ,    3000 ,  800 ,  60
Rainmaker       , 1.58 ,    3500 , 1200 ,  60
Headhunter      , 1.85 ,    3800 ,  700 ,  60
Raven           , 1.8  ,    4700 , 1400 , 100
Gunboat         , 3.2  ,    5800 , 1400 , 100
Modified Argosy , 1.96 ,    4800 , 1900 , 140
Manta           , 3.4  ,    5900 , 1500 , 140
```

The Quicksilver's base cost seems reasonable, but the Raven (comparable to the Gunboat) and the Modified Argosy (comparable to the Manta) seem too cheap.

Amazinite:
>The total price including outfits is unimportant. Outfits aren't fixed, they can be removed or replaced. Whenever I purchase a new ship, I first sell all installed outfits and afterwards start installing what I want. What matters is the base cost.

You still need the amount of credits for the cost of the hull plus the outfits to get the ship in the first place. The minimum amount of credits it takes a player to get a Quicksilver is 2 mil and change, not 1.09 mil. The base cost is only important in the scenario you mention when you have credits to spare, but early game that isn’t always the case as one may often find themself buying a new ship with most of their funds and not changing much on it until they’re able to gain more credits. You have to realize that when most people look at the cost of a ship, they aren’t looking at the hull cost, they’re looking at the cost that the game shows them on the map and in the shipyard view, therefore that is what is most likely to affect their buying habits.

Nescio0:
Perhaps. However, the ship's base cost ought to depend on the ship's base stats, not on whatever outfits it happens to have installed by default. 
A Manta with a Blue Sun Reactor installed shows up as more expensive than a Kestrel, but that doesn't make it a better ship, nor is it an argument to alter the base cost of either.

Amazinite:
A Manta with a Blue Sun Reactor is something the player does, not something they can buy by default, so I don’t see that as a valid rebuttal. And while better hull = higher cost is valid to an extend (ignoring the lore of the shipyards that ships come from, e.g. how the Mod Argosy is a poorly put together ship made by pirates, or how the Syndicate make cheap stuff both in cost and ability), the cost of the ship with its outfits affects gameplay more than just the cost of the ship itself. Whatever the case may be, +800k on the cost of the Raven isn’t warranted. If the Raven hull is to get a price hike in comparison to the Headhunter, I think +300k is all that’s necessary.

jafdy:
The players who like changing outfits a lot should be able to get a better deal than the players who don't. I see no issue in balancing pricing on the default loadout so that there isn't really a standout or plain bad choice for people unwilling to mix and match parts.

Nescio0:
The Headhunter, Quicksilver, and Raven are statistically similar (shields, hull, 60 weapon capacity) to each other and have similar base costs. 
The Raven is similar to the Gunboat (both have 100 weapon capacity and 1400 hull); of course, the Gunboat is a Navy ship and therefore probably has a premium price, but still, the Raven's price ought to be closer to the similar Gunboat than to the weaker Headhunter. The average of G (3.2) and H (1.85) is 2.525, therefore I believe putting the Raven's base cost at 2.6 is not excessive.

The base cost of the M. Argosy probably ought to be increased too; it is statistically similar to the Manta, and Megaparsec ships are typically not expensive.

Pointedstick:
Hull cost cannot be considered in isolation. What are the actual as-purchased prices with this change? Can you show us a table with that?

Nescio0:
Here you go:

60 weapon capacity warships:
```
Quicksilver: 1.09 base, 2.018 total
Rainmaker:   1.58 base, 2.433 total
Headhunter:  1.85 base, 2.616 total
```

100 weapon capacity warships:
```
current Raven:  1.80 base, 2.713 total
proposed Raven: 2.60 base, 3.513 total
Gunboat:        3.20 base, 4.258 total
```

140 weapon capacity warships:
```
M. Argosy: 1.96 base, 3.461 total
Manta:     3.40 base, 4.886 total
Nest:      2.50 base, 4.509 total
Roost:     3.00 base, 5.139 total
Skein:     3.50 base, 5.859 total
```

Amazinite:
There's a lot more to a warship than just its weapon capacity, so saying the Modified Argosy should be similar in cost to the Manta just because they have similar weapon space is a fallacy. The Manta has better shields and hull, better heat dissipation, more outfit space, and lower drag and higher engine space that leads the Manta to being a much more nimble ship than the Modified Argosy. The Mod Argosy may be a heavy hitter in its weapon space, but it's sacrificing a lot to get to that point as a light warship.

Nescio0:
> There's a lot more to a warship than just its weapon capacity

Yes, there is, which is why I think the Raven ought to be more expensive. The Raven is in many respects statistically closer to the Manta than to the Headhunter:

```
        Headhunter , Raven  , Manta
base cost:  1.85 M ,  1.8 M ,  3.4 M
shields:      3800 ,   4700 ,   5900
hull:          700 ,   1400 ,   1500
crew/bunks:  2 / 4 , 6 / 13 , 6 / 10
cargo:          50 ,     30 ,     20
outfit:        250 ,    280 ,    350
weapon:         60 ,    100 ,    140
engine:         80 ,    100 ,    100
```

Nescio0:
> And I never said it shouldn’t be more expensive, I’m simply disagreeing on the amount of increase.

Yes, I know, however, I still believe increasing it only by +0.3 M is too little. 
* The Raven is statistically closer to the Manta than to the Headhunter
* The proposed Raven base cost (2.6) is the average of the Manta's (3.4) and Headhunter's (1.85).
* The total price including outfits of the proposed Raven (3.513) is actually closer to the Headhunter (2.616) than to the Manta (4.886)
* The Gunboat, a similar ship, is still clearly more expensive (3.2 M base, 4.258 M total)

Issue #4019, opened 2018-11-20
Pirates and other enemies should engage in combat when they board you


Pointedstick:
Right now ship combat is only initiated when you do it. It should be something that happens *to* you as well, when boarded! This might provide an incentive to buy any of the more defensively-oriented H2H outfits, like the Security Station.

Amazinite:
I don't see a large number of H2H weapons being necessary.

Pointedstick:
> i can see how this would be interesting, but it's not really an incentive for more defensive h2h. When a pirate disables you, you're basically screwed either way - either they overrun you in h2h or they eventually give up and blow up your ship.

Actually, if pirates boarded you and initiated combat, there's the possibility that your crew could not only fight them off, but counterattack, capture their ship, and use it to repair your own!

jafdy:
Let the pirates make some bad choices when attacking (unlike the perfect logic they currently display when defending), this makes it possible to counter attack and capture the pirate's ship.

DingusShingleton:
> if pirates boarded you and initiated combat, there's the possibility that your crew could not only fight them off, but counterattack

This is basically what I was proposing in the other thread, that each round of combat contained both attack and defense phases. It would make combat insanely more dynamic, without needing TOO too much of an overhaul.

Zitchas:
As part of the "have pirates make bad choices" thing, how about having some kind of faction (or fleet) variable (kind of like the turret inaccuracy) that controls their agressiveness (and a second variable, perhaps just use the turret inaccuracy one) that adds a random factor. 

A highly aggressive ship would attack even when the odds aren't in their favour (how much risk they'll take depends on how agressive they are. A value of 100 would attack all the time. I guess basically it would translate to "how low does the success rating for an attack have to be in order to do it?"

As far as that goes, does the AI take combat endurance into account? For instance, if by some weird chance the player has far less crew, but really good H2H outfits, such that the player is likely to win but take casualties every time the AI attacks - but the AI has enough crew that even if the player wins every round, the AI would end up winning simply by attrition. Will they AI do the attacks?

westhand:
If the flagship can be attacked that way, then certainly other ships in the fleet are vulnerable as well? Then you might want extra crew, which needs to be managed somewhere...

> Note: this is just a request for them to try to capture your ship and kill you, not plunder your outfits which could result in you confusingly and frustratingly getting stranded with no engines or hyperdrive.

[...]

PR #4055, opened 2018-12-08
Differentiate hand-to-hand weapons and remove the 'no-brainer' factor

Pointedstick:
@endless-sky has often mentioned that there shouldn't be any "no-brainer" outfits that you should just always want. But the hand-to-hand outfits all represent exactly that. They have no drawbacks, save for their financial cost which (let's face it) it pretty much irrelevant.

This attempts to remedy that situation by giving the hand-to-hand weapons drawbacks of some sort--principally energy consumption or heat generation, as appropriate. Attack and defense values are likewise tweaked to differentiate the weapons from one another to reduce the logical A->B->C progression factor. Costs are reduced because now there are other disadvantages; now the fairly basic Laser Rifle doesn't cost 22% of its wielder's annual salary! Finally, descriptions are tweaked accordingly.

It is hoped that these changes make hand-to-hand weapons more interesting and tactical rather than just something you thoughtlessly buy the most of the best of and forget about.


toilethinges:
Although I approve of differentiating them, defense is pretty much useless in H2H weapons. So this slightly nerfs three weapons I (and the people I have talked to) wouldn't use, does nothing to Frags, and adds a negligible amount of heat generation to Repeater Rifles.

You also say that financial cost is "pretty much irrelevant", but do nothing to Nerve Gas. Nerve Gas is the highest attack H2H weapon by a lot, and its only downside is... financial cost.

1. Make defense useful, then rebalance based on its use.
2. Also, since you can install and use multiple types of hand-to-hand weapons, it's basically finding whatever has the highest attack and whatever has the highest defense, and installing a full complement of both.

MCOfficer:
more diversity could be achieved by introducing more h2h attributes - for example:
- introduce armor of different sorts
- different types of rifles are effective against different types of armor
-> when boarding korath, choose remnant rifles

DingusShingleton:
The thing is, with boarding combat being as simple as it is, there isn't much wiggle room here. It's a simple X vs Y calculation. Get the outfits with the best X and call it a day, and since the player being boarded is basically game over anyway, Y doesn't really matter at all. The obvious way to change that up using the pieces we already have is to somehow incorporate Y into the equation in a new way.

If you think about the dynamics of combat, no matter how you slice it, you need to have some kind of defensive capabilities. A hundred guys with nothing but grenades are not going to beat fifty guys with rifles. Sure they can dish out some damage, but at the end of the day, if the defenders bum rush them, that's it. If I run at you with an M16 and all you have is frag grenades, you're going to have a bad time.

What if combat consisted of an attack, followed by a counter attack? When you board a ship and attack, it weighs your `attack` vs their `defense`, deals casualties to the defender, and then calculates their `attack` vs your `defense` and deals casualties to you. That way, rather than simply stockpiling whatever weapon has the best attack, you have to think about a balance of attack and defense. Maybe you can still go full offense and mount an attack so devastating that their counter attack is pitiful, but that's a risky strategy. Better to go for a more well rounded approach.

Thus, a weapon with really high `defense` would be a valuable addition to a crew--maybe add five or 10 just to round out your squad. You have a bunch of maniacs lobbing grenades, and then a handful of guys providing covering fire.

Obviously this would be a fundamental change requiring new code and some serious tweaking, but I think it would, using stats we already have, add quite a bit of depth to what is now an elegant, if rather simplistic, system.

Nescio0:
Actually I don't see why boarding weapons would consume energy or produce heat when they are *not* used.

tehhowch:
The simplest way to make defense matter is to enable boarded NPCs to initiate the attack. This has the side effect of making plundering a large ship from a small ship very risky

Pointedstick:
Yes, making them attack is pretty much required, since otherwise the `capture defense` trait is basically useless.

It would also help if they boarded you and tried to capture your ship (https://github.com/endless-sky/endless-sky/issues/4019)

jafdy:
It used to be the case that other ships could initiate attacks, but people found it annoying (despite the alternative being your ship getting blown up).
I forget where this was said and I don't care to find it.

Amazinite:
It was that ships could plunder your outfits, leaving you stranded if you did find help without any obvious reason as to why you might have been unable to move or jump to the next system.

Amazinite:
>Is the fact that defense is currently useless any reason not to make these changes anyway?

I'd say yes. If there's a problem at the core of how H2H weapons are used, I'd change the core issue instead of tweaking the stuff on the outside as this PR does. Perhaps fixing the core issue is all that's necessary and these tweaks wouldn't help with how the core issue is solved. To use an analogy to a past issue, this is in my eyes like equipping more ships with missiles and AM turrets before balancing missiles to be fun in the first place.

Pointedstick:
Fair enough. Should we start a new issue to discuss H2H improvements then, or do you think we already have enough?

PR #4060, opened 2018-12-10
Update copyright file for new images

DrBlight:
I forgot to update the copyright file for my new changes to the planet images on Norn and Muspel. As always, I assign them to the public domain. High DPI version change can be found here: https://github.com/endless-sky/endless-sky-high-dpi/pull/62

### DIFF
--- a/copyright
+++ b/copyright
@@ -98,12 +98,13 @@ Copyright: Nate Graham <pointedstick@zoho.com>
 License: CC-BY-SA-4.0
 
 Files:
- images/land/sky2.jpg
- images/land/sea3.jpg
- images/land/beach4.jpg
- images/land/canyon9.jpg
+ images/land/sky2*
+ images/land/sea3*
+ images/land/beach4*
+ images/land/canyon9*
 Copyright: Emily Mell <hasmidas@gmail.com>
 License: public domain
+Based on public domain images taken from unsplash.com
 
 Files:
  images/icon/gat*
@@ -320,7 +321,6 @@ Files:
  images/land/beach0*
  images/land/beach2*
  images/land/beach3*
- images/land/beach4*
  images/land/beach5*
  images/land/beach6*
  images/land/canyon0*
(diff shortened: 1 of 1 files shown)


copyright
```diff
@@ -100,6 +100,8 @@ License: CC-BY-SA-4.0
 Files:
  images/land/sky2.jpg
  images/land/sea3.jpg
+ images/land/beach4.jpg
```
MCOfficer: this conflicts with [L323](https://github.com/endless-sky/endless-sky/blob/dd5dfcacfa9dd484806f598ce65a49d4bbb2cfda/copyright#L323)

Amazinite: Notably sky2 and sea3 are also elsewhere in the copyright file, but canyon9 is now.

DrBlight: What would be a preferable arrangement?

copyright
```diff
@@ -100,6 +100,8 @@ License: CC-BY-SA-4.0
...
+ images/land/canyon9.jpg
```
Amazinite: As with everywhere else in the copyright file for images, you should drop the file extensions and instead have `*` at the end. And for the high DPI version, drop the `@2x` and the file extension for the `*`.

copyright
MCOfficer: (disclaimer: this is me thinking logically, not based on debian's specifications. in fact, i tried to find any guidelines regarding this case, but debian apparently didn't consider that somebody would want to license the same file twice)

it would be less confusing if the original source was under a wildcard (`images/land/beach*`), and the modification was explicitly stated. We actually have a similar case right in this repo:
```
Files: sounds/*
[...]

Files: sounds/heavy?rocket?hit.wav
```

this case is clear - the wildcard has less "weight" than the explicit statement below it. as it stands with this change, however, both have equal weight. without looking at this PR, it's impossible to tell for sure which file is the original and which one is modified. (There is a wildcard, butas Amazinite pointed out below, your change should have the same wildcard, so it's negligible)

imo if you want to mention the original source, do it like this:
```
Files: sounds/*
Copyright: Various
License: public-domain
Based on public domain sounds taken from freesound.org.
```

Issue #4050, opened 2018-12-03
Deep River in Wanderers: Unfettered Diplomacy 1A has no cargo


AlbertNewton:
This is pretty much what the name suggests. On a whim I scanned the Ayi'i Ki'ak'var, the Deep River in the aforementioned mission to see how much food there was.
I got this:
![image](https://user-images.githubusercontent.com/42004388/49398520-8d8e2d80-f736-11e8-8b09-3eb9f9a1ae2c.png)
The mission centres around delivering food to the unfettered, and to see that there's no food being delivered just killed the immersion.

Brick63:
Just because you can't see it doesn't mean it isn't there

AlbertNewton:
...I can cargo scan Void Sprites. Why not a deep river? I want the technology that lets them evade cargo scans.

Amazinite:
So right now the ship in that mission is defined like so.
```
	npc accompany save
		government "Wanderer"
		personality timid escort
		ship "Deep River (Jump, Unarmed)" "Ayi'i Ki'ak'var"
```
but I believe that changing it to this will mean that that ship will then carry food.

```
	npc accompany save
		government "Wanderer"
		personality timid escort
		fleet
			names "wanderer"
			commodities "Food"
			cargo 5
			variant
				"Deep River (Jump, Unarmed)"
```
Then on the return trip that would be changed to `cargo 0` so that it has no cargo on the way back.

Amazinite:
#4062 should fix this.

PR #4062, opened 2018-12-12
Gave the Deep River in Unfettered Diplomacy 1A food cargo

Amazinite:
Ref: #4050 

Scanning the Deep River in this mission will now show food cargo.
image

And on the return trip.
image

### DIFF
--- a/data/wanderers start.txt
+++ b/data/wanderers start.txt
@@ -565,10 +565,7 @@ mission "Wanderers: Unfettered Diplomacy 1B"
 	npc accompany save
 		government "Wanderer"
 		personality timid escort
-		fleet
-			names "deep river"
-			variant
-				"Deep River (Jump, Unarmed)"
+		ship "Deep River (Jump, Unarmed)" "Ayi'i Ki'ak'var"
 
 fleet "Unfettered Tribute"
 	government "Hai (Unfettered)"
--- a/data/wanderers start.txt
+++ b/data/wanderers start.txt
@@ -501,6 +501,9 @@ ship "Deep River" "Deep River (Jump, Unarmed)"
 		"Type 4 Radiant Steering"
 		"Jump Drive"
 
+phrase "deep river"
+	"Ayi'i Ki'ak'var"
+
 
 
 mission "Wanderers: Unfettered Diplomacy 1A"
@@ -525,7 +528,12 @@ mission "Wanderers: Unfettered Diplomacy 1A"
 	npc accompany save
 		government "Wanderer"
 		personality timid escort
-		ship "Deep River (Jump, Unarmed)" "Ayi'i Ki'ak'var"
+		fleet
+			names "deep river"
+			commodities "Food"
+			cargo 5
+			variant
+				"Deep River (Jump, Unarmed)"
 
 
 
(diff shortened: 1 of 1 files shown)


data/wanderers start.txt
```diff
@@ -501,6 +501,9 @@ ship "Deep River" "Deep River (Jump, Unarmed)"
 		"Type 4 Radiant Steering"
 		"Jump Drive"
 
+phrase "deep river"
```
jafdy: Would it be better to name this is a way that references the mission? Or just use the generic name generator?

Amazinite: The generic name generator can't be used because this ship is in two missions, and the ship needs to be altered between the missions (it loses its cargo). I thought about making the phrase name more specific, but it doesn't really matter right now unless we need to do this again.

tehhowch: In fairness, the return trip (empty) needs no special treatment. So really you are only jumping hoops on the should-have-cargo mission legs

Nescio0:
If this is to be done, then probably `mission "FWC Diplomacy 1B"` ought to be changed too (an Argosy loaded with heavy metals). And there are possibly a few other jobs or missions as well.

Amazinite:
Could you point out every mission where the escorts are said to have cargo?

Nescio0:
Every mission? It turns out there are more escort freighter missions than I initially thought. Looking only at `data/free worlds side plots.txt`, there are already six: 
* "FW Conservatory 2" (goods)
* "FW Conservatory 2B" (terraforming equipment)
* "FW Wolf Pack 2A" (food))
* "FW Wolf Pack 3" (food))
* "FW Wolf Pack 4" (food)
* "FW Stack Core 1B" (fissionable material)

And that's just one of the smaller mission files.

Also, are you sure it has to be the food *commodity*? Cf.:
```
mission "Unfettered Tribute 1"
	name "Hai Tribute to <planet>"
	job
	repeat
	description "Collect a tribute payment of <tons> of food from <stopovers>, and bring it to the Unfettered on <destination>. Your share of the tribute will be <payment>."
	to offer
		has "Unfettered Jump Drive 2: offered"
		random < 80
	cargo "food (tribute)" 25 2 .05
	on complete
		payment 5000 1600
		dialog "You drop off the tribute to the Unfettered, and they pay you <payment>."
	source
		attributes "unfettered"
	stopover
		distance 3 4
		attributes "hai"
	on stopover
		dialog
			`When you inform the Hai that you are here to receive a tribute payment for the Unfettered, they load the cargo onto your ship with a rather surprising swiftness and cheerfulness.`
```

Nescio0:
Basically, your approach (create an unique nameset and replace the one line ship with a six line fleet) works for a single mission. But if there are twenty or more missions, then I'm wondering if there isn't a more efficient approach. Perhaps a minor code change to allow specifying escort cargo?

Also, fully loaded freighters are a lot slower than empty ones, so these missions become harder (which is not necessarily bad).

Amazinite:
>are you sure it has to be the food commodity?

Commodities are what we’re able to make NPCs carry.

Amazinite:
The thing is, I wonder if it's even something worth "fixing" given how niche of a situation it is.

Pointedstick:
We do periodically get bug reports saying "I scanned this ship and its cargo didn't match what the mission said!" I thin it's worthwhile to fix. It's one of those things that's rare, but if you decide to do it and you get unexpected results, it'll break you out of your suspension of disbelief.

Pointedstick:
Fair enough.

Amazinite:
Closing this. Will open a separate PR later making this change in all places necessary, or would welcome someone else doing it.

PR #4070, opened 2018-12-15
Changed the NPCs in Terraforming 11 from bounty hunters to pirates

Amazinite:
For the entire terraforming string of missions, this is the only point in which there is any combat. Given the fact that a player can very easily reach this point without a very powerful ship or even any combat skills. I've swapped the NPCs that you have to fight from bounty hunters to pirates, meaning that the player must no longer take on these NPCs solo and will be supported by nearby merchants and any Navy fleets that enter the system.

### DIFF
--- a/data/transport missions.txt
+++ b/data/transport missions.txt
@@ -2343,7 +2343,7 @@ mission "Terraforming 11"
 			`Your ship is targeted by pirate fleets when you enter the system. Before you engage, you receive a message from Amy.`
 			`"Captain, these pirate ships have been occupying the system for days. You must take them out if we are to safely crash an asteroid into Tundra."`
 	npc evade
-		government "Bounty Hunter"
+		government "Pirate"
 		personality nemesis staying target
 		system "Cebalrai"
 		fleet "Marauder fleet I"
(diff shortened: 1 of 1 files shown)


tehhowch:
Do you happen to have those possible fleet tables already prepared? Base `map.txt` and also after any possible changes during the campaign?

Amazinite:
You have the right PR?

tehhowch:
You mentioned that the player could receive help, so I thought perhaps you had also quantified what help may be possibly received.

Amazinite:
I’m simply referring to the random fleets that enter the system, in the same way that a player taking an escort job can travel to a populated system and have the merchant and law enforcement fleets destroy the pirates for them. If a player reaches this point in an interceptor, it doesn’t become game over if the government of the NPCs is something others will attack.

tehhowch:
I too am referring to the random fleets that enter the system.
If the average credits per second of help is too high, then I think an alternate path is better: keep Bounty, lower the base enemy strength, and use a "reenforcements" mission if the player warrants it

Amazinite:
/shrug
Aside from this one instance, this entire string of missions is entirely non-combative. I'd rather it be too easy than to hard.

Fzzr:
Looking at Cebalrai in the data files, it has non-Pirate combat fleets ranging from "Small Militia" at the beginning of the game to "Large Republic" after the battle of Alioth.

PR #4071, opened 2018-12-15
Changed some dialog in the terraforming Tundra missions

Amazinite:
So right now there's somewhat of a contradiction between the new terraforming Tundra missions and the situation that happened on Hope. A supervolcano erupted on Hope and plunged the entire world into an ice age terrible enough to cause the mass evacuation of the entire planet. But a supervolcano eruption on Tundra, an already cold world, is going to be helpful?

The reason that volcanic eruptions cause global temperatures to drop (at least partly) is due to volcanic ash in the upper atmosphere soaking up the heat from the Sun, meaning that the Sun warms up the upper atmosphere instead of the surface. (There are also chemical reactions that occur with the sulfur in the ash, but really it's all a matter of the ash being in the atmosphere.) But volcanic eruptions are also exactly what is predicted to have given Earth its atmosphere, as the eruptions brought gasses trapped in the crust and mantle up above the surface, allowing an atmosphere to form in the first place, and more volcanic eruptions just meant a warmer planet due to the atmosphere. The current way the mission is written already tries to justify this by saying that the ash can be scooped out of the atmosphere by ships, leaving the beneficial greenhouse gases behind that cause the planet to warm while removing the harmful volcanic ash that would otherwise cool the planet down. This right now begs the question as to why this wasn't done on Hope to avoid the ice age there.

This PR seeks to try and handle this contradiction once and for all by stating how the eruption on Hope was sudden and no preparation was made; perhaps by the time the proper ships with the proper equipment could be in place, the ash had spread out too much for any reasonable number of ships to make an impact on the climate of the planet. In contrast, the eruption on Tundra is "controlled" in a way, meaning that the proper equipment can be on standby and ready to be used immediately, scooping up the ash from the eruption while it is still in a centralized area and before it is able to do damage to the atmosphere.


data/transport missions.txt
```diff
@@ -2311,8 +2311,8 @@ mission "Terraforming 10"
...
+			`	"Sir," Amy speaks up with a stern tone of voice, "we have already been authorized by the Republic to undergo this terraforming experiment, so unless you have half a teraton of cheap explosives, we're going to need to use an asteroid. The volcanic winter can be easily avoided by having a fleet of freighters sweep the volcanic ash out of the atmosphere, meaning we'll get the benefits of the geenhouse gases without the detriments of volcanic ash reflecting Cebalrai's heat back into space. No one knew the volcano on Hope was going to erupt, but we can prepare for this eruption and take steps to midigate the problems." Amy looks the representative straight in the eyes and smirks. "We're scientists, we have this handled."`
```
tehhowch: `mitigate`.

Last sentence may need a semicolon instead of a comma:
```
We're scientists; we have this handled.
```

Nescio0: Or perhaps "We're scientists, we know what we're doing." or something similar?

Also, "midigate" [sic] ought to be "mitigate" and "geenhouse" [sic] "greenhouse".

data/transport missions.txt
```diff
@@ -2176,7 +2176,7 @@ mission "Terraforming 6"
 				`	(Run away.)`
 					flee
 			
-			`	"The unlawful use of an asteroid, Captain. I'm not entirely sure what that means, but I've been told to send you to <planet>. Your compliance would be much appreciated."`
+			`	"The unlawful use of an asteroid, Captain. I'm not entirely sure what that means, but I've been told to send you to <planet> by <day>. Your compliance would be much appreciated."`
```
Zireael07: "I've been told to send you to <planet> " is perfect English. "Send you to <planet> by Tuesday" is not, though. I'd suggest making it a separate sentence, e.g. "I've been told to send you to <planet>. They will be expecting you by <day>".

jafdy: This will read (for the correct date): `. . . send you to Earth by the 22nd of January`

Zireael07: Either way, I'm not sure that's idiomatic English (but may be wrong, I am not a native, just an English major - but I've never heard "send you to X by the 1st of June")

data/transport missions.txt
```diff
@@ -2311,8 +2311,8 @@ mission "Terraforming 10"
...
+			`	"No, no, no!" the Republic representative exclaims, cutting Nolan off mid sentence. "The Republic will not gain a reputation for crashing asteroids into its own planets. The consequences of such an action could be disastrous should it go wrong! And in what way will this warm the planet when the resulting volcanic winter could last decades and only result in an even colder climate? Are you not even old enough to remember how a supereruption on Hope caused the evacuation of the entire planet? If the people of Tundra are this desperate to terraform their planet then normal techniques should be employed, none of this 'equilibrium mapping' nonsense."`
```
Pointedstick: mid sentence -> mid-sentence

data/transport missions.txt
Pointedstick: I think this is fine--both the original and the proposed change.

PR #4086, opened 2018-12-19
Add alien Brig jobs

Fzzr:
This adds a silly Hai job using the Brig, plus some Coalition jobs to and from Ring of Power.

### DIFF
--- a/data/coalition jobs.txt
+++ b/data/coalition jobs.txt
@@ -1154,3 +1154,125 @@ mission "Coalition: From Games 3"
 		payment
 		payment 25000
 		dialog `The passengers thank you for bringing them to <planet> and pay you <payment>.`
+
+
+mission "Coalition: Transport Prisoners 1"
+	job
+	repeat
+	name "Transport Kimek prisoners to <planet>"
+	description "Transport <bunks> Kimek prisoners from <source> for reeducation on <planet>."
+	passengers 20 200 .1
+	to offer
+		has "known to the heliarchs"
+		random < 10
+		"passenger space" > 20
+	source
+		government "Coalition"
+		attributes "kimek"
+	destination "Ring of Power"
+	on offer
+		has "Brig"
+	on complete
+		"coalition jobs" ++
+		payment 15000 150
+		dialog `The prisoners are led off of your ship by a single Kimek ensign. Curiously, none of them make any attempt to resist or escape. A Heliarch officer gives you your payment of <payment>.`
+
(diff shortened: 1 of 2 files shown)


data/coalition jobs.txt
```diff
@@ -1154,3 +1154,125 @@ mission "Coalition: From Games 3"
...
+	description "Transport <bunks> Kimek prisoners from <source> for reeducation on <planet>. Payment is <payment>."
+	passengers 20 200 .1
+	to offer
+		has "known to the heliarchs"
+		random < 10
+		"passenger space" > 20
+	source
+		government "Coalition"
+		attributes "kimek"
+	destination "Ring of Power"
```
tehhowch: I'm not sure that the `"Ring of Power"` is an appropriate destination - I view it as the Heliarch equivalent of the Pentagon, not a prison ringworld.

Naf5000: Ringworlds have more usable area than could ever be used by any of the species described in-universe. There's more than enough space to have military/governmental headquarters and prisons so far apart you'd basically need a spaceship to get between them anyway.

Fzzr: I figured it would have to be a Heliarch-controlled area, and Ring of Power makes the most sense to me as a place where more "forceful" applications of Heliarch will would happen.

Bladewood:
Oh, well, it's not quite approval, I'm still worried about what tehhowch wrote, but otherwise it looks good to me, if slightly repetitive.

data/coalition jobs.txt
```diff
@@ -1154,3 +1154,125 @@ mission "Coalition: From Games 3"
...
+	on complete
+		"coalition jobs" ++
+		payment 15000 150
+		dialog `The prisoners are led off of your ship by a single Arachi ensign. It's impossible to tell how they feel, but none of them step out of line or make any trouble. A Heliarch officer gives you your payment of <payment>.`
+
+
+mission "Coalition: Return Prisoners 1"
+	job
+	repeat
+	name "Transport released Kimek prisoners home to <planet>"
```
Amazinite: image
To where?

Fzzr:
Hmm. I'll withdraw this branch and make one with just the Hai job. As for there not being Brigs in Hai space, that's kinda the point - a ship with a brig is a novelty. Maybe I should make that a spaceport job instead? I'll come back with a more fully baked approach to the Coalition jobs if I can think of one later.

Issue #4081, opened 2018-12-18
Laser Turret sprite is incorrect


Nescio0:
Let's compare the laser turret and heavy laser turret images:

* projectiles:
![2x laser 0](https://user-images.githubusercontent.com/21634324/50153841-df52be00-02c7-11e9-9785-e0eab9d50698.png)
![2x heavy laser 0](https://user-images.githubusercontent.com/21634324/50153851-e2e64500-02c7-11e9-873c-316906c7f10d.png)
* hardpoints:
![laser turret](https://user-images.githubusercontent.com/21634324/50153871-f396bb00-02c7-11e9-92ee-7b1e61555cb8.png)
![heavy laser turret](https://user-images.githubusercontent.com/21634324/50153881-f8f40580-02c7-11e9-9698-d43325d43d93.png)
* outfit thumbnails:
![laser turret](https://user-images.githubusercontent.com/21634324/50153900-0c9f6c00-02c8-11e9-94fb-1ad4fa1a1003.png)
![heavy laser turret](https://user-images.githubusercontent.com/21634324/50153906-132de380-02c8-11e9-98d8-01f9cd046069.png)

As you can see the beam laser and heavy laser turret's projectiles are two parallel beams and the hardpoints of both clearly show the two cannons parallel to each other. The beam laser turret's outfit sprite, however, has the cannons mounted on top of each other, similar to the blaster turret. It ought to be parallel, similar to the heavy laser turret. It would be nice if someone could create a correct sprite.

Brick63:
should be fixed in #4090

PR #4090, opened 2018-12-21
fix laser turret sprite

Brick63:
fix related to #4081 makes the sprite match the hard-point


Brick63:
it has the PR, just took me a minute to open it

Nescio0:
Copyright file?

The proposed image looks nice and certainly fixes #4081. However, let's compare it with the heavy laser and electron turrets:

electron turret 2x
heavy laser turret 2x
laser turret 2x

As you can see the proposed laser turret has a small gap between the two halves; to me this suggest that the two cannons might rotate vertically independently from each other, which ought not to be the case. By contrast, the heavy laser and electron sprites have here some solid piece, emphasizing the two cannons are supported by a turret mount.
In short, could you raise or enlarge the piece between the two halves?

Pointedstick:
> As you can see the proposed laser turret has a small gap between the two halves; to me this suggest that the two cannons might rotate vertically independently from each other, which ought not to be the case. By contrast, the heavy laser and electron sprites have here some solid piece, emphasizing the two cannons are supported by a turret mount.
> In short, could you raise or enlarge the piece between the two halves?


I don't think this is actually a problem.

@Brick63 Please update the copyright file in this PR (as well as the HiDPI PR) and then I think we can merge it!

jafdy:
It looks to me like the laser turret is angled slightly differently to the heavy and electron turrets. The copyright file doesn't need updating as he only moved a bit off the turret not made a new piece (as stated by brick himself in the discord).

Brick63:
The original had that exact same dip there, only it was all on one side. I didn't design this, only altered it to fit something else and don't feel like that is deserving changing the copyright.

MCOfficer:
> It looks good, but I do agree with Amazinite: Why do we need all the turrets to be of the same "two barrels on either side" type design? Wouldn't it be easier to change the sprite instead?

looking at it from an engineer's perspective, there's no reason not to design it like this. There's no real upside of pairing guns vertically, the symmetry likely makes for easier construction, and turret mounts (or their parts) are probably standardized to some degree. If we want this turret to be the odd one out, i feel like we need some kind of solid reason.

Amazinite:
At that point I think you’re getting too detailed. Does every weapon description need to justify its design to you? No, it just needs to look good! Just because it looks different doesn’t mean it needs to explain why it looks different. Does the proton turret explain why it’s the only non-AM turret in human space with one gun on it? No, it just is.

Pointedstick:
So what are we doing with this? The discrepancy between the sprite and the beam is a real issue. I'm okay with this sprite change, but then we also need a new hardpoint sprite too.

The alternative of making the beam only depict a single brighter laser rather then two side-by-side lasers could work too. But we should do _something_.

Brick63:
That's the this, the hard point looks the same as what this sprite does, so it's in no need of change

Amazinite:
I still think that the hardpoint sprite and projectile should change and not the outfit sprite. It may be "more work" to change two things than to change one thing, but it's also a less noticeable change than changing the outfit sprite. Changing the projectile and hardpoint sprite will only go noticed when actually fighting against or with the laser turret, which isn't exactly all that common past a certain point, while changing the outfit image is highly noticeable given that you can see laser turrets in outfitters at any point.

Pointedstick:
That could work too. My only very minor concern with that approach is that it might be a bit weird having just this one turret fire one bright beam from the center when all the other beam-type turrets fire two side-by-side beams. But we can certainly try it out and see if it works.

Nescio0:
Actually I still don't see what's wrong with this.

PR #4091, opened 2018-12-22
Heavy laser fix (double beam to single beam)

Zitchas:
Changes the sprite for the Heavy Laser Turret from the double beam to the single beam, as the turret has both laser cannons are portrayed above each other, not side by side; so should just be a single beam when viewed from above.

edit: I was hoping this would just pull the branch itself, but no, it drags everything else along. If this is still useful in doing so, great, if not, I'll close it.


Zitchas:
Wow. 92 commits, just in my own stuff. lol.

PR #4093, opened 2018-12-22
Make player able to be more sympathetic to the Pug

10010101001:
Although some people despise the Pug, some don't have any special feeling towards them, some of us are more sympathetic to the Pug. On most dialogs you can pick which side you're on, but with the Pug it is mostly hateful.


10010101001:
Done all that

data/wanderers start.txt
```diff
@@ -364,6 +364,8 @@ mission "Tell Wanderers about the Pug"
 			choice
 				`	"But wouldn't you rather be free from their control, free to remain on these worlds you have worked so hard to restore?"`
 				`	"If your species is being secretly controlled by another, doesn't that bother you?"`
+				`	"I will try not to anger them."`
```
Amazinite: This choice doesn't sound like a response to what Rek just asked. Perhaps instead of a choice that simply ends the mission there, the player is able to respond with something along the lines of "I guess you're right," followed by a response from Rek that then ends the mission.

Pointedstick:
@10010101001, are you around to resolve @Amazinite's change request?

10010101001:
Sorry been away for a while

10010101001:
thx amazinite

PR #4095, opened 2018-12-23
Add a Gridfire sprite

10010101001:
Add a Gridfire sprite, which resembles a point defense system.


Nescio0:
On the one hand, I appreciate it someone took the effort to create a sprite for the Gridfire Turret. On the other hand, I don't think the proposal is good enough: too little detail, too little colour, too simple a shape, and too different from the current Pug content. Compare it with e.g. the Zapper Turret:

gridfire
pug zapper turret

Zitchas:
Looks like a pretty good start to me. I agree it looks a tad dark, but the overall concept looks good.

PR #4102, opened 2018-12-29
New wanderer transport

Brick63:
Finally bringing this bad boy back, it has limited implementation right now because i don't know what @endless-sky might want to do with it.
Stats by @Amazinite 

Assets:
[Riptide assets.zip](https://github.com/endless-sky/endless-sky/files/2724072/Riptide.assets.zip)


Darcman99:
Seems a little.. fast.

Brick63:
That's the idea

data/wanderer ships.txt
```diff
@@ -855,3 +855,55 @@ ship "Deep River Transport"
...
+		"required crew" 47
+		"bunks" 233
+		"mass" 350
+		"drag" 5.1
+		"heat dissipation" .7
+		"fuel capacity" 800
+		"cargo space" 164
+		"outfit space" 488
+		"weapon capacity" 98
+		"engine capacity" 186
```
Zitchas: I would recommend dropping the engine capacity down by 62 to 124, then remove one of the Type 4 steering and replace the Type 2 thruster with a type 1. 

This would drop the stock stats a notch to be a bit more reasonable:
Max speed: 856 down to 727
acceleration: 326 down to 300
Turning: 208 to 113

10010101001:
Seems like a competitor to the Bactrian with all its bunks

jafdy:
Not for the min-maxers out there. And I think a T2 stock competitor is fine.

solardawning:
> Seems like a competitor to the Bactrian with all its bunks

In terms of bunks and speed, yes- but it doesn't increase max bunk capacity, has far less outfit space, cargo space, and weapon load.

It seems like a nice addition to me, actually- fast, T2 durability, lightly armed, could fill in as an inpromptu but undergunned medium warship pre-Wanderer Warships if made available early enough in the Wanderer unlock chain. 
It'd be fun as a flagship for player control due to the speed and the bunks, too.

It's not objectively a 'generalist best in class in many ways' that the Bactrian is.

Darcman99:
Min-maxing the riptide gives it far less bunks than you can achieve with the bactrian but it still sports alot, the riptide with no additional bunks still puts it up there as a good capping ship. I wasn't arguing for it to lose all its' speed and become super slow, just reduced by 20-30%.

Zitchas:
What are its stock speeds with all these revisions?

Darcman99:
140 turn and 280 acceleration, for a top speed of 600.

Zitchas:
That sounds kind of fast on the turn, but those speeds look good. It is really fast, but not crazy fast. And I suspect that a bit of min-maxing can get it up to "crazy" levels. :)

10010101001:
Maybe change the looks. It looks kind of 'off' for Wanderer ships.

Zitchas:
@10010101001 Could you elaborate a bit on that? It is hard for people to make adjustments (or responses) to vague statements.

beccabunny:
maybe if you reduce the border size and a few more decorations it will be more similar to the vanilla Wanderer ships. Also, if you make the engines a bit larger it will follow more the first concept "speed depends just from engines"
riptide
this is an example of what i mean (not yet rendered)
it is yet a very good ship, btw

Brick63:
thank you, and i see what you mean. Ill work on it when i can.

beccabunny:
Nice decorations!

Zitchas:
Looks good.

Pointedstick:
@amazinite I'm not sure it's safe to assume that @endless-sky is coming back. We might want to consider merging it.

@Brick63 looks like an unintentional permissions change snuck in:

Screenshot_20190331_145446

Amazinite:
>@Amazinite I'm not sure it's safe to assume that @endless-sky is coming back. We might want to consider merging it.

Have you no faith?
The thing is, MZ is the one working on the Wanderer campaign. The intention of this ship is to be used during the Wanderer campaign, so unless you're suggesting we write our own end to the Wanderer campaign, I'm keen on just leaving this open.

Zitchas:
I'm sure he'll be back, I'm just not sure when.

That being said, the main story-based use for this ship, at least to me, is to be involved in the evacuation effort to get all the Wanderers out of their old space into new space. Given that them getting their first toehold established was done by simply having the player wait around for a month, I don't think there should be a problem introducing these ships at about that time and have them shown in-use as part of convoys from Wanderer worlds through the eye to their new worlds.

Not to mention, I think it'd be good to have some actual missions (or at least jobs) available during that time where the player could take part in said evacuation. Just because they're good at killing things doesn't mean they can't help transport people and supplies.

SirStigII:
I do think it would make sense to start seeing these just before or right after the eye opens, especially if the tracker hit force stays the same. The Tempests that you ecort through the last unfettered blockade get thrown around like crazy on that mission. Having better maneuverability on the escorts on that mission would maybe help.

Brick63:
No idea @MCOfficer, I'm not sure what happened

SirStigII:
Well I know what I'm going to be looking at tonight...

tehhowch:
@MCOfficer @Brick63 https://stackoverflow.com/questions/1257592/how-do-i-remove-files-saying-old-mode-100755-new-mode-100644-from-unstaged-cha

PR #4103, opened 2019-01-01
Add Hai jobs to bring commodities back from human space

Fzzr:
This adds some two-way transport missions for Hai space. They pay twice as much as the one-way delivery jobs, but have deadlines. This also nerfs the existing delivery jobs.

### DIFF
--- a/data/hai jobs.txt
+++ b/data/hai jobs.txt
@@ -350,7 +350,7 @@ mission "Delivery to Human Space [0]"
 	job
 	repeat
 	description "This human wishes for <cargo> to be delivered to <destination> in human space. Payment is <payment>."
-	cargo random 2 10
+	cargo random 5 10
 	to offer
 		random < 30
 	source
@@ -359,7 +359,7 @@ mission "Delivery to Human Space [0]"
 	destination
 		attributes "north"
 	on complete
-		payment 80000
+		payment 60000
 		dialog "You drop off your cargo of <commodity> and collect your payment of <payment>."
 
 mission "Delivery to Human Space [1]"
@@ -367,7 +367,7 @@ mission "Delivery to Human Space [1]"
 	job
 	repeat
 	description "This Hai wishes for <cargo> to be delivered to <destination> in human space. Payment is <payment>."
-	cargo random 2 10
+	cargo random 5 10
 	to offer
 		random < 25
 	source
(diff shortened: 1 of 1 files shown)


data/hai jobs.txt
```diff
@@ -396,6 +396,63 @@ mission "Delivery to Human Space [2]"
 		payment 80000
 		dialog "You drop off your cargo of <commodity> and collect your payment of <payment>."
 
+mission "Hai Retrieve Human Luxury Goods"
+	name "Special order to <planet>"
+	job
+	repeat
+	description "A Hai socialite on <destination> wants to show off their knowledge of human culture to their peers. They will pay <payment> for <cargo> from <stopovers>."
```
Fzzr: Question: Does `<destination>` or `<planet>` feel better here? I went with `<destination>` so that the return address would be in the description, but I want to hear what others think.

tehhowch: It's a good idea to include some UI examples when asking for feedback like this

Fzzr: Sure. 
Version with `<destination>`:
destination description
Version with `<planet>`:
planet description

Pointedstick: The `<planet>` version seems a bit more elegant to me.

Fzzr: Switched to that.

Amazinite:
I don't think adding these jobs would be all that great. They're essentially just "Special Delivery" jobs that pay you for returning as well. Those special delivery jobs already pay extremely well though; they're some of the highest paying jobs you can get very early on, and typically the way I pay off my starting loan. Creating missions that essentially do the same thing as special delivery jobs (I return to Hai space after making special deliveries regardless), but pay twice as much may be a bit overboard.

tehhowch:
One way to combat the accumulation is to give these all deadlines, which would then also justify the higher payout. I think automatic deadlines are not well equipped for wormhole traversal, so you'll want to use a fixed deadline. I think an occasional impossible deadline is fine (since the player may have a jump drive and be able to make some that would otherwise be impossible).

Amazinite:
And one could argue that being on the other side of the wormhole, the Hai asking you to do these jobs don't know how long it'll actually take. I still think that the payment is way too high and the cargo way too low, though. I'd increase the cargo req form `1 5` to `5 10`, and decrease the payment from 160k to 100k. While we're here, perhaps also nerf the special delivary jobs: `2 10` cargo to `5 10` and 80k payment to 50k.

PR #4069, opened 2018-12-15
Made FW Katya 1C only offer on inhabited worlds

Amazinite:
It's rather odd landing on the icy hellscape of Hope and finding a group of Navy troops waiting for me to check my cargo.

PR #4105, opened 2019-01-01
Do not trigger the remembrance day or war begins missions on uninhabited planets

### LINKED ISSUE #4104: Bug: Remembrance Day mission triggering on Hope
Hope is an uninhabited planet in Wei, a system in Republic space. If this is the first planet in the Republic the player lands on after date 19 2 3014, mission Remembrance Day will trigger despite a lack of a spaceport or other people mentioned in the dialogue.

Playing on Windows 10, nightly build, commit hash 9804d6c.

Save file: [https://hastebin.com/iriyahihap.pl](url)

remembrance day bug

Fzzr:
This excludes all `uninhabited` just to be sure.

Fixes #4104

### DIFF
--- a/data/events.txt
+++ b/data/events.txt
@@ -17,6 +17,7 @@ mission "remembrance day"
 	landing
 	source
 		government "Republic"
+		not attributes uninhabited
 	to offer
 		has "event: remembrance day"
 	on offer
--- a/data/events.txt
+++ b/data/events.txt
@@ -17,7 +17,7 @@ mission "remembrance day"
 	landing
 	source
 		government "Republic"
-		not attributes uninhabited
+		not attributes "uninhabited"
 	to offer
 		has "event: remembrance day"
 	on offer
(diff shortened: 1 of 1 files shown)


data/events.txt
```diff
@@ -17,6 +17,7 @@ mission "remembrance day"
 	landing
 	source
 		government "Republic"
+		not attributes uninhabited
```
Pointedstick: Let's quote it for consistency's sake.

Fzzr: Done in cccfaad.

Fzzr:
I realized that the war start mission works the same as the remembrance day one, so I made the same change. A quick scan of other uses of `landing` didn't reveal any other obvious cases where such unexpected behavior could occur.

Amazinite:
See #4069, using `attributes spaceport` may be better. Both approaches yield the same results (as in human space, no spaceport = uninhabited, at least for now), but using the spaceport attribute would be consistent with other missions.

Pointedstick:
> See #4069, using `attributes spaceport` may be better. Both approaches yield the same results (as in human space, spaceport = inhabited, at least for now), but using the spaceport attribute would be consistent with other missions.

+1, this seems like the semantically correct approach.

Issue #4115, opened 2019-01-07
Test Dummies should disable, not kill


Amazinite:
It seems odd to me that the guy you meet on Tarazed doesn't want you to blow up his Kestrel, but it will happily shoot you out of the sky. The same can be said for Syndicate dummy ships.

tehhowch:
I agree re: the Kestrel - but the Syndicate? Your death just proves that their ship is good enough -- and now they don't have to pay you.

Amazinite:
Well they might not need to pay you, but they’d then need to answer to the Republic why they’re killing law abiding citizens.

tehhowch:
Would they though? "Oops, another day, another industrial accident."

In terms of gameplay, adding `disables` to the personality (for both the Kestrel and the Syndicate test ships may result in something that is basically like death - you get disabled, then hail for repairs, and the second you get repaired, the test ship disables you again (or some pirates finish you off).

Fzzr:
I could imagine a case for being able to fail a mission due to being disabled. That would open up things like "duels" where you can bet and lose money due to being disabled, and allow for interesting dialogue branches depending on the results of testing missions.

jostephd:
> What captain is going to take such a job if they’ve heard of many other captains and crew dying at the hands of Syndicate drones?

Two kinds of people: those who love risks, and those who are desperate for money and are forced to take risks. I'm sure there are people who migrate to Syndicate worlds, run into debt, and when they have exhausted all easier options, rent a ship and start doing high-risk missions (freight to planets that neighbor pirate worlds, etc) because the payoffs are bigger. Some of them probably emigrate to the near-core pirate worlds in order to escape their Syndicate creditors (like Joe and Maria but in reverse).

Amazinite:
There are a lot more high paying jobs than the dummy tests, especially ones that are far less dangerous.

jostephd:
> There are a lot more high paying jobs than the dummy tests, especially ones that are far less dangerous.

That sounds like a balance problem. Risky jobs should pay more than non-risky jobs.

> And risk taking is why we have escort and bounty jobs.

[...]

PR #4118, opened 2019-01-08
Made Syndicate test dummies disable and gave them confusion

Amazinite:
Ref: #4115 

Syndicate test dummy ships will now disable instead of killing the player. A happy customer is an alive customer. No use in killing off captains in the Core that could otherwise be spending their money on Syndicate goods.

Also gave Syndicate test dummy ships a confusion value of 120. Confusion isn't used hardly enough in the game. Given that the technology that the Syndicate are working on with these ships is effectively the same thing as the Kor Automata, I thought it fitting to give these test dummies a behavior that makes it very clear that the technology of fully automating large vessels is only in the works for the Syndicate.

### DIFF
--- a/data/syndicate jobs.txt
+++ b/data/syndicate jobs.txt
@@ -174,7 +174,8 @@ mission "Syndicate target practice [0]"
 		"destroyed Syndicate target ship" < 3
 	npc save disable "scan cargo"
 		government "Test Dummy"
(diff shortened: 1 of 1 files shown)


tehhowch:
Should we update the confusion values for the Doombat and Kestrel in the relevant missions, too?

PR #4122, opened 2019-01-13
Add Rescue Jobs

czartrak:
These are a few jobs that I've worked on for a while. The objective is for the player to find a disabled ship and bring them back to the source world. The jobs scale from small rescuing Merchant interceptors to Republic or Militia light warships.

### DIFF
--- a/data/jobs.txt
+++ b/data/jobs.txt
@@ -1265,7 +1265,7 @@ mission "Escort (Extra Large, Southern, Dangerous Origin or Destination)"
 		
 mission "Rescue (Interceptor, Northern)"
 	name "Rescue Merchant Interceptor"
-	description "The captain of the <npc> bit off more than they could chew. Repair them and escort them safely back to <destination>."
+	description "The captain of <npc> bit off more than they could chew. Repair them and escort them safely back to <destination>."
 	repeat
 	job
 	source
@@ -1310,7 +1310,7 @@ mission "Rescue (Interceptor, Northern)"
 				
 mission "Rescue (Transport, Northern)"
 	name "Rescue Merchant Transport"
-	description "The captain of the <npc> was disabled by pirates while ferrying passengers through dangerous space. Repair them and escort them safely back to <destination>."
+	description "The captain of <npc> was disabled by pirates while ferrying passengers through dangerous space. Repair them and escort them safely back to <destination>."
 	repeat
 	job
 	source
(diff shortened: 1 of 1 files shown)


data/jobs.txt
```diff
@@ -1262,6 +1262,302 @@ mission "Escort (Extra Large, Southern, Dangerous Origin or Destination)"
...
+		government Republic "Free Worlds" Syndicate Neutral
+		attributes frontier north
+	to offer
+		random < 40
+		"combat rating" >= 20
+	
+	npc accompany save
+		system
+			distance 1 3
+		personality waiting derelict timid
```
tehhowch: `waiting` is inconsequential as a personality since these are disabled ships, and they don't spawn in your system. The purpose of `waiting` is that the ship is "waiting in orbit" instead of taking off alongside you.

czartrak:
I gave them "waiting" so that they would follow you back to the system. "escort" didn't work as you could then see what system they were in. No point in searching

tehhowch:
"follow the player" is the default behavior for mission NPCs. `uninterested`, `fleeing`, and `staying` are the only ones that do not have this behavior.

czartrak:
Oh I see. I was just looking at the Ravens from Deep Archeology to try to replicate this behavior. I'll make the change

Amazinite:
I'm not too sold on the idea of these jobs. Who is posting this on the job board? Won't random ships in the system repair the ship for you once you enter it? Why can't the ship just ask for assistance from nearby ships instead of somehow posting on a job board up to 3 systems away? Isn't this just the same thing as the assisting missions really, only with more steps?

tehhowch:
W.r.t. other ships repairing these mission NPCs, `derelict` ships never ask for help - the player alone can choose to repair them.

data/jobs.txt
```diff
@@ -1262,6 +1262,302 @@ mission "Escort (Extra Large, Southern, Dangerous Origin or Destination)"
 		dialog "The captain of the <npc> thanks you for escorting them safely, and pays you <payment>."
 	on visit
 		dialog "You have reached <planet>, but you left the <npc> behind! Better depart and wait for them to arrive in this star system."
+		
+mission "Rescue (Interceptor, Northern)"
+	name "Rescue Merchant Interceptor"
+	description "The captain of the <npc> bit off more than they could chew. Repair them and escort them safely back to <destination>."
```
Fzzr: Here it says `the captain of the <npc>` but below it says `The captain of <npc>`. I suggest going with whichever sounds better with ship names. This applies to all the merchant missions here. The military ones use the second version.

Amazinite:
Spoke with OP. I'd much rather see these as a non-repeatable spaceport mission.

PR #4125, opened 2019-01-15
Add hand-to-hand weapons to Republic Navy ships

Fzzr:
This adds 40 Laser Rifles and Frag grenades to all Carrier variants, 20 of same to all Cruisers, and 5 of same to all Frigates.

The effect of this will be making it harder for the player to capture what are nominally the toughest ships in human space. It also adds immersion - apparently right now Navy Marines are so badass they can defend against boarding parties with their bare hands.

I will do a separate PR to do similar for Dreadnoughts and maybe other Militia/FW ships, and a third to update Oathkeeper fleets with Mark II variants at the appropriate time.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -725,8 +725,6 @@ ship "Carrier"
 		"Heavy Laser Turret" 2
 		"Heavy Anti-Missile Turret" 2
 
-		"Laser Rifle" 40
-		"Fragmentation Grenades" 40
 		
 		"Fusion Reactor"
 		"LP288a Battery Pack"
@@ -734,6 +732,8 @@ ship "Carrier"
 		"Large Radar Jammer" 2
 		"Water Coolant System"
 		"Brig"
+		"Laser Rifle" 40
+		"Fragmentation Grenades" 40
 		
 		"X5700 Ion Thruster"
 		"X4200 Ion Steering"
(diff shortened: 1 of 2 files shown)


data/ships.txt
```diff
@@ -724,6 +724,9 @@ ship "Carrier"
 		"Meteor Missile" 140
 		"Heavy Laser Turret" 2
 		"Heavy Anti-Missile Turret" 2
+
+		"Laser Rifle" 40
+		"Fragmentation Grenades" 40
```
Amazinite: See the Marauder ships file. H2H weapons don't get their own "block" like this separated from the rest of the outfits. They go at the bottom of the center block.

Fzzr: Sure thing. Done in b087f88.

Fzzr:
I went for round numbers, around 1/2 the required crew for Frigate and Cruiser, more than that for the Carrier because I figured it should be harder to capture.

tehhowch:
Any thought to Security Stations? Currently they are only employed by marauder ships, and as I mentioned in #3427 I would imagine them being used at least occasionally by Navy / large ships.

Fzzr:
Looking at the variants: The default Carrier and Cruiser have space for security stations, but the Frigate and all the variants have little to no space to spare. I don't think it's worth the trouble of rebalancing all of them for this purpose.

Lineth:
The Cruiser and Carrier could use one or two then; it would gel with them being the Navy's most important assets and thus deserving more protection.

Amazinite:
The problem with security stations is that they take up outfit space that could otherwise be used for actual combat. I wouldn’t change the Carrier and Cruiser variants so much as to allow the installation of 8 or 4 stations. If you can make a trivial change and fit just one, that’d be alright, but making big changes to fit 8 or 4 stations won’t fly.

tehhowch:
hmm, guess that method of merging didn't close this. Welp, 940f89b

PR #4126, opened 2019-01-15
Add Mark II Oathkeeper fleets at the same time as the rest of the Navy

Fzzr:
This is a minor immersion fix - there are Mark II Oathkeeper ships later in the story, but no randomly spawned fleets. This way the Navy isn't neglecting the fleets responsible for defending the North.

I also considered adding the Oathkeeper variants later, so that they can't be used for the Electron Beam hunting missions. If that's preferable, I'll add them in a separate event sometime between then and the FW Alphas missions.

### DIFF
--- a/data/events.txt
+++ b/data/events.txt
@@ -1126,7 +1126,55 @@ event "navy using mark ii ships"
 			"Rainmaker (Mark II)" 2
 			"Gunboat (Mark II)"
 
+	fleet "Small Oathkeeper"
+		add variant 6
+			"Rainmaker (Mark II)" 2
+		add variant 5
+			"Gunboat (Mark II)"
+			"Rainmaker (Mark II)"
+		add variant 4
+			"Frigate (Mark II)"
+		add variant 3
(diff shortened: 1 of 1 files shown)


jafdy:
Flying all the way to the north just to get an electron beam off the oathkeepeers is unlikely to ever come up in normal play, or non normal play for that matter, I wouldn't worry about it.

data/events.txt
```diff
@@ -1126,7 +1126,55 @@ event "navy using mark ii ships"
 			"Rainmaker (Mark II)" 2
 			"Gunboat (Mark II)"
 
+	fleet "Small Oathkeeper"
+		add variant 6
```
Amazinite: Is there any reason for why the frequencies of these ships differs from those above? Why not just have the same additions to the Oathkeeper fleets as the other Navy fleets get?

Fzzr: I was indeed going for a lower frequency, but matching them is fine with me. Done in 8e4de9b.

PR #4127, opened 2019-01-17
Restock fighter ammo from carrier's cargo hold

tehhowch:
This is an implementation of #1971 with the ammo stockpile in the carrier's cargo - not its installed outfits
As with the installed-outfits implementation (#4128), a fighter that gets resupplied will launch even if it wasn't able to fill up on fuel.

 - Cargo refill allows both carrier and fighter to use the same ammo resource without impacting carrier damage output
    e.g. Carrier (Meteor) & Finch (Meteor)
 - Upon docking, carried ships pool ammo in cargo bay for other fighters to use

(Current) downsides:
 - Mission NPCs cannot be given explicit outfit cargo, thus scripted battles cannot guarantee use of ammo restocking - only that ships in the `fleet` will collectively have said outfits in cargo 
 - Player cargo is filled in reverse size order, beginning with your [first escort][firstescort] (see #2757) and your flagship is filled last and only if necessary
   - Biggest outfits are almost never ammo
   - Carriers would probably be geared to have large cargo holds (so they can hold lots of ammo for fighters)
   - **Ammo would likely end up in ships that fighters do not board without undue micromanagement** of carrier position in the escort list (fighter micromanagement is already a known detraction from their use)
   - No way to tell a fighter to load up on ammo without deploying it (you lose control of your ships' cargo space, e.g. reduced plunder capability)
 - If we resupply fighter ammo from cargo, why not carrier ammo from cargo too?

[firstescort]: https://user-images.githubusercontent.com/20871346/51285272-ed255d00-19b3-11e9-89ed-47df590a3d9f.png

Additional discussions:
https://github.com/zwparchman/endless-sky/pull/1#issuecomment-315591202
https://github.com/endless-sky/endless-sky/issues/1939#issuecomment-269992593
https://github.com/endless-sky/endless-sky/issues/2336
https://steamcommunity.com/app/404410/discussions/2/492378265890026032/
https://steamcommunity.com/app/404410/discussions/0/1319962683445095725/#c1327844097131431281
(I'm sure there are other references as well)

PR #4128, opened 2019-01-17
feat(fighters): Restock fighter ammo from carrier's installed outfits

tehhowch:
This is an implementation of #1971 with the ammo stockpile in its installed outfits - not its cargo hold.
As with the from-cargo implementation (#4127), a fighter that gets resupplied will launch even if it wasn't able to fill up on fuel.

 - Installed refill allows explicit control over which player ships have ammo for fighters, including quantity and type
 - Upon docking, carried ships refill the carrier's ammo storage, for other fighters or the carrier itself to use
 - NPC outfits are explicitly and easily customized by content creators, for both regular system fleets and for mission NPCs

(Current) downsides:
 - Requires either ammo storage outfits or the carrier to have the same weapon as the fighter
   - If both the carrier and the fighter have the same ammo weapon, the carrier will also consume the ammo. This may be confusing for players. Reference: https://github.com/endless-sky/endless-sky/pull/3791#discussion_r200775518

Additional discussions:
https://github.com/zwparchman/endless-sky/pull/1#issuecomment-315591202
https://github.com/endless-sky/endless-sky/issues/1939#issuecomment-269992593
https://github.com/endless-sky/endless-sky/issues/2336
https://steamcommunity.com/app/404410/discussions/2/492378265890026032/
https://steamcommunity.com/app/404410/discussions/0/1319962683445095725/#c1327844097131431281
(I'm sure there are other references as well)


Nescio0:
This seems both the natural and the practical option to me; I'd prefer it instead of #4127.

It might be an idea to make ammunition storage occupy cargo space instead of outfit space, to allow carriers (and other ships) to install a lot of missiles without sacrificing other outfits? I'm not completely sure it's a good idea, but it's worth considering.

DingusShingleton:
What if carriers had a certain amount of built in ammo storage which would only be used to restock fighters? Its capacity could be based on mass, since all the ammo types now have mass, and thus would be weapon independent. It would fill up only after you filled up the ship's usable weapon capacity, meaning that
A) you can use different weapons on your carrier and your fighters 

And B) fleet carriers can do it as easily as buying missiles for their launchers

Alternatively there could be an outfit that does the same thing to allow customization, but since bays are hard coded it makes sense to me that this would be too. Hard coding it also allows there to be more specialized carriers with a huge amount of space for carrying ordnance. Or both /shrug

tehhowch:
The number of ammo you can load up depends solely on the `attributes` of the ammo and the carrier. Since the game does not track which of its aggregated `attributes` came from each specific outfit, a carrier would have no idea that e.g. 300 of the "meteor missile capacity" it has is supposed to be used only for fighters and the remaining e.g. 70 comes from its own installed outfits.

The game also doesn't track which attribute is the one that limits how many of any item you can have installed - you can use any name you want for your non-reserved flavor attributes (e.g. `"fluxon bindings"`instead of `"meteor missile capacity"` is perfectly fine) - it only tells you which attribute is preventing you from installing *more* of a given outfit. There's no guarantee that this attribute is the one *intended* to prevent adding more of a given ammunition.

So yes, you could give carriers specific attributes that let them hold more of a given outfit, but without a large change to the attribute system the carrier would happily consume those outfits too. (Related: https://github.com/endless-sky/endless-sky/issues/4067#issuecomment-447505890)

Amazinite:
I'd still prefer ammo be refilled from cargo (ref: the last listed downside there, I don't see that as an issue given that docked fighters can be worked on from the outside and therefore be easily rearmed; just travel to the cargo bay, grab some missiles, and refill the fighter as if you were sitting in the outfitter. *waves hand in regard to externally docked fighters*), but out of simplicity since cargo can't be easily handled and allocated to the proper ship(s), I say we go with refilling from installed.

source/Ship.cpp
```diff
@@ -1751,7 +1751,28 @@ void Ship::Launch(list<shared_ptr<Ship>> &ships, vector<Visual> &visuals)
...
+				auto toRefill = set<const Outfit *>();
+				for(const auto &hardpoint : bay.ship->Weapons())
+				{
+					const Weapon *weapon = hardpoint.GetOutfit();
+					const Outfit *ammo = weapon ? weapon->Ammo() : nullptr;
+					if(ammo && OutfitCount(ammo))
+						toRefill.insert(ammo);
+				}
+				// Transfer as much ammo as the fighter needs.
+				bool tookAmmo = false;
```
ravenshining, line 1764: How about *both*? Pasting in some code from your cargo version, this

1. Draws ammo from cargo
2. If cargo is insufficient, draw remaining ammo from outfits

I agree with petervdmeer's [comments on the other PR](https://github.com/endless-sky/endless-sky/pull/4127/files#r362398112), so I haven't touched the restock.  This would, however, in a roundabout way, allow a carrier to slowly replenish its own launchers from cargo as long as its matchingly armed fighters survive.

```suggestion
				// Transfer as much ammo as the fighter needs from cargo.
				bool tookAmmo = false;
				for(const Outfit *ammo : toRefill)
				{
					int canTake = bay.ship->attributes.CanAdd(*ammo, cargo.Get(ammo));
					if(canTake > 0)
					{
						cargo.Add(ammo, -canTake);
						bay.ship->AddOutfit(ammo, canTake);
						tookAmmo = true;
						carriedMass += ammo->Mass() * canTake;
					}
				}
				// Recalculate ammo needs.
				for(const auto &hardpoint : bay.ship->Weapons())
				{
					const Weapon *weapon = hardpoint.GetOutfit();
					const Outfit *ammo = weapon ? weapon->Ammo() : nullptr;
					if(ammo && OutfitCount(ammo))
						toRefill.insert(ammo);
				}
				// Transfer as much ammo as the fighter needs from outfits.
```

ravenshining (commented):
If we don't do both as I suggest in the code edit below this comment, I much prefer this PR to the other one.

However, what I'd really like to see is for this process to take time, handed in the repair section of the code.  In my testing, I found it rather funny that a fighter would land empty and then immediately take off again full.

That being said, gameplay > realism tells me that fighters need all the help they can get, and the insta-reload can be handwaved by saying your crew anticipates the operation and has special future reloading mechanisms.

source/Ship.cpp
```diff
@@ -129,6 +130,28 @@ namespace {
 		else
 			stat = max(0., .99 * stat);
 	}
+
+	// Transfer the as many of the given outfits from the source ship to the target
```
petervdmeer, line 134:
```suggestion
	// Transfer as many of the given outfits from the source ship to the target
```

source/AI.cpp
```diff
@@ -1715,7 +1715,28 @@ bool AI::ShouldDock(const Ship &ship, const Ship &parent, const System *playerSy
 	if(ship.Health() < minHealth && (!ship.IsYours() || Preferences::Has("Damaged fighters retreat")))
 		return true;
 
-	// TODO: Reboard if in need of ammo.
```
petervdmeer, line 1718: I tested a version of this PR (with a recent master-branch merged into it) and noticed that the Outfitter feature "Do you want to reload all the ammunition for your ships?" didn't work for automatically reloading Ammo storage racks for ships that don't also have the proper weapon to fire such Ammo installed. (Which could be the case for carriers that want to resupply their fighters.)

EjoThims: That would certainly be a helpful addition with this

source/Armament.cpp
```diff
@@ -197,6 +198,26 @@ int Armament::TurretCount() const
...
+set<const Outfit *> Armament::RestockableAmmo() const
+{
+	auto restockable = set<const Outfit *>{};
+	for(auto &&hardpoint : Get())
+	{
+		const Weapon *weapon = hardpoint.GetOutfit();
+		if(weapon)
+		{
+			const Outfit *ammo = weapon->Ammo();
+			if(ammo && !ammo->IsWeapon())
```
tehhowch, line 212: @Amazinite ref 0c4947a1ff9c9ede413dba394f3495b0fc2a8738  -- was that addressing a particular issue?

source/AI.cpp
tehhowch: I believe that is added now

source/Armament.cpp
Amazinite: If a weapon has ammo usage, the outfitter attempts to refill as much of that ammo as possible. Normally, the amount of ammunition one can carry is capped by the capacity attribute for that ammo. But if the ammo usage is 0, it's expected that you're not dealing with normal ammunition (i.e. there is no ammo capacity attribute, and the ammo being used likely isn't without outfit space like normal ammo). What was happening was that people were using an ammo usage of 0 where the ammo was some normal outfits (e.g. a jump drive), and then whenever they landed. checked the outfitter, and refilled their ammo, the outfitter would fit as many of that outfit as it possibly could within the limits of the ship's outfit space, therefore cluttering the ship with ammo that wasn't actually needed, since an ammo usage of 0 only requires one instance of the outfit in order to fire.

source/Ship.cpp
tehhowch: I've basically done this, though I add a bit of flavor by restricting cargo restock to internally carried ships.

source/Ship.cpp
```diff
@@ -2305,7 +2328,40 @@ void Ship::Launch(list<shared_ptr<Ship>> &ships, vector<Visual> &visuals)
 			// Resupply any ships launching of their own accord.
 			if(!ejecting)
 			{
-				// TODO: Restock fighter weaponry that needs ammo.
+				// Determine which of the fighter's weapons we can restock.
+				auto restockable = bay.ship->GetArmament().RestockableAmmo();
+				bool tookAmmo = false;
+				// Ammunition found in the cargo hold can be transferred to internally carried fighters.
```
petervdmeer, line 2334: Nice addition, but why in this PR? Can we just keep this for a separate follow-up PR?
(This is not just a code-change, but this is also relevant for game-balancing and content-creators. I would really like to keep the relevant discussions on those topics for a follow-up PR.)

petervdmeer:
I don't consider the addition of tests blocking for this PR (because the integration tests I describe below are not very elegant and because unit testing is not yet possible for this code), but if you already want to add some automated testing, then it does seem possible to already add some tests through the integration testing framework.

Some examples:

**Test1 (basic reloading in outfitter):**
- Load a game (with ships with ammo not fully filled, and the mission from the later step present)
- Click to go to the outfitter.
- Click the confirm button to reload ammo on all ships.
- Fly to a planet where the "ammo" outfits are not for sale.
- Check with a mission with a [mission/location filter](https://github.com/endless-sky/endless-sky/wiki/CreatingMissions#conditions) (and without dialogs) that the "ammo" outfits are present.
- Check in the conditions that the mission action was performed.

**Test2 (racks reloading in outfitter):**
- Same as test1, but with ships with empty ammo-racks but no weapons for them)

**Test3 (fighter reloading in flight):**
- Load a game (with unarmed carriers that have ammo-storage, and armed fighters using ammo from such carriers, where the fighters have no non-ammo consuming outfits, and the mission from the later steps)
- Launch into an area with hostiles that are spawned by a mission
  - Destroying those hostiles should only be possible if the fighters reload from the carrier a number of times.
  - Killing the hostiles should have an effect on the mission (for example by having the npcs marked as "save", causing the mission to fail once the npcs are destroyed).
  - The carriers and fighters should be very durable (and/or the hostiles should be very lightly armed)
- Check for completion of the mission through the conditions (I believe the mission-statuses are updated real-time in the conditions)

source/AI.cpp
petervdmeer: Yes, the added code seems to resolve this.

petervdmeer (changes requested):

source/Ship.cpp
petervdmeer: Sorry, I should have made my comment about adding this in a separate PR here in this thread, instead of as a separate comment.

source/AI.cpp
```diff
@@ -1710,7 +1710,28 @@ bool AI::ShouldDock(const Ship &ship, const Ship &parent, const System *playerSy
...
+	// If a fighter is armed with only ammo-using weapons, but no longer has the ammunition
+	// needed to use them, it should dock if the parent can supply that ammo.
```
Amazinite, line 1714: Should this be preference based? Or should we just change it so that any ammo-using weapon that's out of ammo is enough for a fighter to restock? Although a fighter that has non-ammo-using weapons combined with ammo-using weapons doesn't need to restock in order to continue fighting, such a fighter with ammo for its ammo-using weapons will be more effective than one without, so it can be beneficial to restock even if it isn't necessary to continue fighting.

source/Ship.cpp
```diff
@@ -2305,7 +2328,40 @@ void Ship::Launch(list<shared_ptr<Ship>> &ships, vector<Visual> &visuals)
...
+					int count = OutfitCount(ammo);
+					if(count > 0)
+						toRestock.emplace(ammo, count);
+				}
+				auto takenAmmo = TransferAmmo(toRestock, *this, *bay.ship);
+				if(!takenAmmo.empty())
+				{
+					tookAmmo = true;
+					// Update the carried mass cache.
+					for (auto &&item : takenAmmo)
```
Amazinite, line 2362:
```suggestion
					for(auto &&item : takenAmmo)
```

source/Ship.cpp
tehhowch: This was a nod to the functionality request in the sister PR, which was as popular as, and perhaps even more so, than this approach to the functionality. I included it in this PR, as the purpose of this PR is to add the ability to restock fighter ammunition from the carrier. Doing so via ammo-from-cargo, or ammo-from-outfits, is an implementation detail.

I don't remember exactly which was first (these PRs or the functionality), but content creators can design AI fleets that will make use of the ammo-from-cargo, through the use of the fleet "outfitters" syntax:

```
fleet "Carriers w/Cargo Ammo"
  outfitters "Ammo North" "Ammo South"
  ...
```

source/AI.cpp
tehhowch: I don't know that it has a meaningful impact, and it is certainly something that can be added later if there is demand for it.

Amazinite: I'm fine with considering this for later.

Amazinite (approved):

source/Ship.cpp
petervdmeer: `if(bay.side == Bay::INSIDE` is not an implementation detail, but an explicit choice about some bays being different than others. Most content-creators will not review source-code-changes into the level of detail in which they would pick this up and I would really like to give content-creators the chance to respond to this.

Another thought is that I still like to explore some other options for "ammo loading from cargo" as was suggested in the sister-PR, but without the downsides as were listed in that sister-PR. Think of outfits that get installed in the cargo space (not in the outfit space) and that:
- Very slowly load their ammo into the main Outfits/weapons
- Or only load ammo into fighters
- Or can only be installed on some carriers due to some attributes.

Adding the "from cargo for internal bays" functionality here will prevent exploration of some of those other options.

source/Armament.cpp
warp-core: Shouldn't this not include the ammo if the launcher doesn't actually consume it, then?

warp-core:
Amazinite:
> I like the idea, but I'm not opposed to taking that bit out if it's in question.
quyykk
> You can't control where any of your cargo goes so unless your flagship is a carrier and you don't have any other escorts it's going to be hard to make it so that ammo goes into your carrier's cargo hold. 
> So I'm not a fan but I don't really mind either way 

Based on conversation on Discord, it sounds like @Amazinite and @quyykk are ambivalent on the cargo reloading for inside bays only. Myself, @petervdmeer, and @Saugia would prefer if either all bay locations were eligible for cargo reloading or none at all.

Would anyone else like to weight in?

EjoThims:
Should be all bays if done, but that option (and the changes to cargo storage to support it) should be a separate PR so we can get this one merged.

petervdmeer (changes requested):
My (blocking) review comment https://github.com/endless-sky/endless-sky/pull/4128#discussion_r889606210 is still open.

warp-core:
> Should be all bays if done

As in, either all bays should be able to reload from cargo or none of them?

EjoThims:
> > Should be all bays if done
> 
> 
> 
> As in, either all bays should be able to reload from cargo or none of them?

Yep.

quyykk:
I think this should be in 0.9.15, it's basically almost done already and is a long awaited feature for people that like to use carriers.

EjoThims:
Agreed. I think it's been waiting for considerations on also using cargo and preferences and such.

At this point I think we should go ahead and get this in ASAP though; we can add the rest later when how to do so is actually sorted out.

petervdmeer:
As suggested above: I just uploaded a commit that focuses this PR on just reloading from outfits and that takes out the reloading from cargo completely. As suggested above; we can handle the reloading from cargo in another PR.

petervdmeer (approved):

PR #4131, opened 2019-01-19
Fix "turn jitter" in escort AI when following a ship

tehhowch:
Current master: https://youtu.be/m9Pnv38m1fc
With this PR: https://youtu.be/Ij8IlIxCH60

I tried "damped" Turn commands of `.01`, `.05`, and went with `.025`.
- At `.01` the escorts will take a while to respond to changes in leader direction, and as such they spread out quite a bit
- At `.05` they are quite responsive, but one can more easily see the beginnings of the jitter behavior

In other words, if `.025` causes issues, we can change it.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1843,19 +1843,17 @@ void AI::KeepStation(Ship &ship, Command &command, const Ship &target)
 	if(abs(targetAngle) > 180.)
 		targetAngle += (targetAngle < 0. ? 360. : -360.);
 	// Avoid "turn jitter" when position & velocity are well-matched.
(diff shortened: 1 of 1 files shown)


Zitchas:
Would this affect mining too? I'm noticing the jitter on the Void Sprites pretty heavily.

tehhowch:
No, this change is purely to non-combat movement AI. Any targeting behaviors use a different AI method to set the desired turn/thrust commands. It's possible an improved consideration of relative position, velocity, and size could address jitter when using a very short range weapon with a larger ship. Probably should be its own issue though.

Issue #4018, opened 2018-11-20
Need a way to differentiate aborting a mission from failing it


Pointedstick:
The `Syndicate target practice` jobs fine you if you destroy the ship instead of disabling it, implemented with a negative payment in the `on fail` block.

But this also triggers if you abort the mission, which isn't appropriate. We need some way of allowing you to abort a mission without it technically counting as having failed.

jafdy:
An `on abort` block that would default to `on fail` block if absent?

Joeppie:
It should probably also be noted that the 'target practice' mission pays for example  20.000 credits for success, whereas capturing the ship and selling it nets me 110.000+ credits. (This was on the Steam version)

Amazinite:
The ship now self destructs when you try to capture it.

Joeppie:
@Amazinite Indeed! I just tried a similar mission, nasty self-destruct mechanisms!. Then again, of course the syndicate needs to protect its technological advancements from snooping privateers.

Fzzr:
This came up again, with another couple of related suggestions:
1. Add new flags to the save file based on failing or aborting missions, `: failed` and `: aborted`
2. Add the ability to `complete` one mission from another
3. Add the ability to `abort` a mission (that one or another) from within a conversation etc.

Basically, promote `abort` and `fail` to have the same capabilities as `complete`.

cc: @Zitchas

PR #4133, opened 2019-01-20
Add a flag to explicitly mark a mission as failed

Fzzr:
I decided to separate this out from the idea of an `": aborted"` flag because as far as I can tell it works without further intervention and is backward-compatible.

ref: https://github.com/endless-sky/endless-sky/issues/4018#issuecomment-455907094

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -796,7 +796,10 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 	else if(trigger == DECLINE)
 		++player.Conditions()[name + ": offered"];
 	else if(trigger == FAIL)
+	{
 		--player.Conditions()[name + ": active"];
+		++player.Conditions()[name + ": failed"];
+	}
 	else if(trigger == COMPLETE)
 	{
 		--player.Conditions()[name + ": active"];


Fzzr:
If accepted, the wiki change to document it is on branch `failure-flag` on https://github.com/Fzzr/endless-sky.wiki.git.

Amazinite:
There’s also the issue of how we would even retroactively label missions as failed in the first place.

jafdy:
It isn't, it can be guessed but no known.

Fzzr:
The other thing I want to do is add the `": aborted"` flag for missions that are dropped rather than failed. Retroactively labeling existing missions failed and then adding that would make the save file less correct, rather than more.

tehhowch:
The condition can be retroactively calculated in the same manner as existing missions can check for failure: query the number of offers, query the number of active, query the number of complete, and assume the discrepancy is due to failures. This would be a check performed upon loading a save. It would need to iterate all missions. Probably it would be smart to include a "we calculated this already" condition, and first check for that prior to the whole calculation bit. After a stable release or two, this bit could be removed.

Amazinite:
How does this distinguish between declining and failing though? That method of checking for offered but not done or active is going to give unnecessary false positives.

tehhowch:
It doesn't, and can't. So, add a declined condition and let's merge this.

Fzzr:
@tehhowch Wiki branch to document them is updated: https://github.com/Fzzr/endless-sky.wiki.git

PR #3475, opened 2017-12-30
Add a pirate mission to capture a rogue smuggler

Fzzr:
This is a simple boarding mission with some flavor. This is my first non-trivial contribution here, so please check carefully and give my any feedback or suggestions.

If this is well-liked, I may follow up with a short string of related missions.

PR #4137, opened 2019-01-23
Smuggler/Pirate mission string [WIP]

Fzzr:
This is the followup to the "Capture Smuggler" mission from #3475. It opens up a storyline about a mysterious organization that performs unusual smuggling (and other) jobs. I've added just the main storyline intro mission for now to get feedback on the tone/general suitability. The bounty hunter attack followup is an example of negative consequences other than just `die`.

TODO:
- [ ] Another mission between the Capture Smuggler mission and the employer intro to establish the player's reputation a bit more. Should be illegal. Move intro to Matthias to the end of this.
- [ ] Additional implications about the immoral nature of the missions. Demonstrate the reputation of the organization a bit more.
- [ ] One second chance mission, rather than skipping straight to bounty hunter attack/end of string. Only after established in org.
- [ ] Intro to opposing law enforcement/investigative organization.
- [ ] Missions with multiple resolutions (support/oppose smugglers, subtly interfere with goals, etc.).
- [ ] Multiple endings that both have appealing rewards.


Amazinite:
Seeing as how these missions require a positive pirate rep, it may be worth putting them into their own "pirate missions.txt" file instead of putting them alongside the other transport missions.

data/pirate missions.txt
```diff
@@ -0,0 +1,189 @@
...
+			`	The shark-toothed man takes another sip of tea and gives you his trademark grin. "<first> <last>," he says. "Today, we are friends. Friends use names, yes? I am Matthias. Now, I am your friend. To celebrate, we will drink tea."`
+			`	Matthias pours you a cup of tea from a ceramic teapot. You raise it to your lips and take a sip. As far as you can tell, it is of very high quality and prepared flawlessly. It is already the perfect temperature to drink.`
+			`	You and Matthias finish your tea in silence. When you put down your teacup, the light in the room dims. A viewscreen concealed in the wall turns on, displaying only the dark silhouette of a man lit from behind. Matthias inclines his head slightly.`
+			`	The man speaks. "<first> <last>, it is good to see you at last." His voice is clear and unaccented. "Pardon my rudeness in concealing my identity, but in my business it is always better to be too cautious rather than not cautious enough."`
+			choice
+				`	Thank you for the generous gift of your time. How may I be of assistance?"`
```
jafdy: Leading "

Pointedstick:
Very cool! One macro pet peeve: I don't like the pattern of conversation where each one of your conversational choices triggers a very minor modification of the conversation partner's next statement, like so:

"Would you like to do a thing?"
- "Yes please! -> "Ah, you are so polite! How nice. Now, onto business..."
- "What are we waiting for?" -> "Ah, a man of action! Indeed, onto business..."
- "Yeah whatever" -> "Now now, that attitude will not make you friends. Nevertheless I don't care so let's continue with our business..."

I think this is really awkward, and always breaks me out of the mood. If the player's choice of words should trigger the conversational partner to have a different response, it shouldn't just be a minor modification of the first sentence, with the conversation otherwise proceeding normally. It should actually make the conversation go differently. If the player's choice of words is not enough to reach this high bar, then all options should just lead to the same response.

data/pirate missions.txt
Fzzr: Fixed in 8c7541b.

data/pirate missions.txt
```diff
@@ -0,0 +1,189 @@
+# Copyright (c) 2019 by Michael Zahniser
```
temtemy: You are the one who created this file, so you should attribute the copyright to yourself, not MZ. Unless you want to transfer your copyright to MZ, but that's not needed

Fzzr: Done in 3a6a51e, thanks.

Fzzr:
Closing this until I have time to work on it more.

PR #4142, opened 2019-01-26
Buffing Coalition civilian modules

chadenr:
Moving coalition civilian from mostly tier 0.75 to proper tier 1.5 with unique offerings: shield modules buff shield HP, repair modules buff hull HP, solar panels offer small ion resistance.

### DIFF
--- a/data/coalition outfits.txt
+++ b/data/coalition outfits.txt
@@ -18,7 +18,7 @@ outfit "Small Collector Module"
 	"outfit space" -8
 	"solar collection" .8
 	"heat generation" .4
-	"ionization resistance" 1
+	"ion resistance" 1
 	description "Solar power is the mainstay of the Coalition's merchant marine. The design of these panels has been refined over the millennia, growing ever more efficient. Because of the distributed nature of the panels, they are harder to disrupt than other power systems."
 
 outfit "Large Collector Module"
@@ -31,7 +31,7 @@ outfit "Large Collector Module"
 	"outfit space" -28
 	"solar collection" 3.3
 	"heat generation" 1.5
-	"ionization resistance" 3
+	"ion resistance" 3
 	description "One of these solar panels is enough to meet the energy needs of a medium-sized Coalition civilian ship, especially given the energy efficiency of their engines and the fact that under their laws only Heliarch ships are allowed to have weapons. Because of the distributed nature of the panels, they are more resilient against weapons that target power systems."
 
 outfit "Small Reactor Module"
--- a/data/coalition outfits.txt
+++ b/data/coalition outfits.txt
@@ -18,7 +18,8 @@ outfit "Small Collector Module"
 	"outfit space" -8
 	"solar collection" .8
 	"heat generation" .4
-	description "Solar power is the mainstay of the Coalition's merchant marine. The design of these panels has been refined over the millennia, growing ever more efficient."
+	"ionization resistance" 1
+	description "Solar power is the mainstay of the Coalition's merchant marine. The design of these panels has been refined over the millennia, growing ever more efficient. Because of the distributed nature of the panels, they are harder to disrupt than other power systems."
 
 outfit "Large Collector Module"
 	category "Power"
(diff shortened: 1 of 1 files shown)


data/coalition outfits.txt
```diff
@@ -18,7 +18,8 @@ outfit "Small Collector Module"
 	"outfit space" -8
 	"solar collection" .8
 	"heat generation" .4
-	description "Solar power is the mainstay of the Coalition's merchant marine. The design of these panels has been refined over the millennia, growing ever more efficient."
+	"ionization resistance" 1
```
tehhowch: You meant `"ion resistance"` - ionization resistance is nothing special (note that there is no tooltip for this attribute of yours if you hover it in the Outfitter screen).

chadenr: Ion resistance fixed and committed.  Shield and hull buffs allow a stealth buff to coalition stock ships and allow greater flexibility in loadout on other ships as a reward for unlocking coalition license. This allows marginal ships like Leviathan and Falcon greater utility and makes human ships useful further into the late game. Lore-wise, I don't see a problem with aliens solving a problem that humans haven't seen the need to address.

solardawning: Hull + shield on outfits is a bit of a dead horse around here. It's something that MZ (the game's creator/primary author) has explicitly said several times he doesn't want used in the main game, as he wants to gradually create a progression of ship power, and use hull & shield as a way of obsoleting lower tier ships. It's probably best kept in a mod so that individual players can choose whether to use it or not, rather than submitted to the main branch.

Amazinite:
>from mostly tier 0.75 to proper tier 1.5

Where did you pull that .75 number from? These outfits are already superior to the human, T1 versions. Humans have no hull repair, the Coalition do. The best standard human shield generator has 1.31 shield gen per ton, were as the small shield module is at 1.53 shield/ton and the large is at 2 shield/ton. The collector modules, even with their added heat generation, are far superior to their human counterparts, with the best human solar collection being 2.75 energy/ton and the small collector module being 6 energy/ton (and 7.07 energy/ton for the large collector module).

data/coalition outfits.txt
```diff
@@ -90,7 +92,8 @@ outfit "Small Shield Module"
 	"shield generation" .28
 	"shield energy" .28
 	"shield heat" .1
-	description "Although combat is almost unheard of in Coalition space, almost all of their ships are able to recharge their shields when necessary."
+	"shields" 560
+	description "Although combat is almost unheard of in Coalition space, almost all of their ships are able to recharge their shields when necessary. Coalition shield rechargers make their shields more resilient to damage."
```
Pointedstick: Outfits that add shield capacity or hull strength are generally verboten because they would allow you to stuff any ship with enough of them to turn it into a heavy warship.

The wiki [explicitly recommends against this](https://github.com/endless-sky/endless-sky/wiki/CreatingOutfits):

> I recommend against providing outfits that give ships additional shield strength, because that could create balancing issues. But if you want to do it, this is the attribute to modify.

Any outfit that increased shield strength or hull capacity would need to incur its ship significant drawbacks to even be considered.

data/coalition outfits.txt
```diff
@@ -103,7 +106,8 @@ outfit "Large Shield Module"
 	"shield generation" 1.3
 	"shield energy" 1.3
 	"shield heat" .5
-	description "The Coalition manufactures heavy shield generators for their ships in case they ever face aggression from the Quarg or another species."
+	"shields" 2200
```
Pointedstick: Same ^^

data/coalition outfits.txt
```diff
@@ -116,6 +120,7 @@ outfit "Small Repair Module"
 	"hull repair rate" .12
 	"hull energy" .12
 	"hull heat" .1
+	"hull" 300
```
Pointedstick: Same ^^

data/coalition outfits.txt
```diff
@@ -129,6 +134,7 @@ outfit "Large Repair Module"
 	"hull repair rate" .57
 	"hull energy" .57
 	"hull heat" .5
+	"hull" 1200
```
Pointedstick: Same ^^

Amazinite:
I don't see any justification for these changes. Ignoring the fact that the proposed changes include to add something that is explicitly recommended against in the wiki (meaning there's no way it's making it into the base game), Coalition outfits are already sufficiently more powerful than human outfits.

Issue #3746, opened 2018-05-21
Reloading save during conversation duplicates mission action


Amazinite:
If the player exits the game while in the middle of a conversation within a mission that also gives new log entries alongside the conversation, the game will duplicate the log entry or entries for that mission upon reentering the game and continuing the conversation.

To reproduce: start a new pilot, begin your conversation with James, quit the game before finishing your conversation, reenter the game, finish the conversation, check your log book.

This was after initiating the conversation three times.
![download](https://user-images.githubusercontent.com/17688683/40286892-d975c58c-5c77-11e8-86e6-fea343c552bd.png)
(I also have three of the same entry for the first pirate, Syndicate, and Republic faction logs.)

Credit to @Tadrix for spotting this.

Edit: and as per @comnom's comment, this appears to work with any condition that is changed in an on offer as long as a conversation is included. I assume that this will also work with dialog boxes and in on completes.

comnom:
It's a consequence of the logs being added in the `on offer`. 

Out of suspicion, I checked to see if the same is true of modifying conditions, and it is. (I made a nonsensical condition in the above mission's on offer `+= 1`, and quitting/restarting mid conversation incremented it each time.)

tehhowch:
The reason for the repeats is that the mission's state is only changed by the conversation callback, while the  rest of the actions are performed in the same call to `MissionAction::Do` that created the conversation.
We could try to shoehorn a boolean in that prevents the rest of the MissionAction from occurring, but probably the best approach is to defer the actions until after the callback has completed.

tehhowch:
A third approach may be simpler yet - change when the game saves data: if a conversation or dialog is on top, don't save. This would be more extensible to non-offer actions, and work for dialogs too.

Amazinite:
Checked `on offer`, `on complete`, and for `branch`es and `apply`s in the following files for anything that might cause an issue.

- [x] coalition missions: no such case
- [x] deep missions: no such case (although I think I did try to use the same technique as in #4493 for `conversation "deep: remnant"`, but I was able to write that conversation without needing it since prior missions exist that can be checked during the branches. If the #4493 technique were possible though, the branching in this conversation could probably be simplified.)
- [x] drak missions: no such case
- [ ] FW checkmate:
- [ ] FW epilogue:
- [ ] FW intro:
- [ ] FW middle:
- [ ] FW reconciliation:
- [ ] FW side plots:
- [ ] FW start:
- [x] hai missions: no such case
[...]

PR #4144, opened 2019-01-26
Certain panels block game saves if still displayed

tehhowch:
Refs https://github.com/endless-sky/endless-sky/issues/3746

This needs to be playtested a fair bit before merging, to make sure there aren't unintended "why didn't it save there" situations. This change should only influence cases where the program is closed (x'd out, Alt-F4) or quit (menu -> quit), and not "normal" (auto)saving such as the pre-takeoff save.

### DIFF
--- a/source/ConversationPanel.cpp
+++ b/source/ConversationPanel.cpp
@@ -58,6 +58,9 @@ namespace {
 ConversationPanel::ConversationPanel(PlayerInfo &player, const Conversation &conversation, const System *system, const shared_ptr<Ship> &ship)
 	: player(player), conversation(conversation), scroll(0.), system(system), ship(ship)
 {
+	// All conversation panels prevent saving the game while displayed.
+	preventsSaving = true;
+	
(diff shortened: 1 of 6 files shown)


jostephd:
Why shouldn't it be possible to save the game while a choice is being offered? If I win a battle, land, and am given a choice, I would _want_ to save right there so I don't have to replay the whole battle in order to try the other choice.

MCOfficer:
that's no information the data format can hold as of now, so it falls under "technically possible, but needs a lot of work".

tehhowch:
It's possible my 2nd's solution in #3746 is the one that should be implemented (defer the remainder of the MissionAction until after the offer conversation has completed). It's more complex, but would preserve the "you did the previous thing" state recording that is currently done, even if a new conversation is displayed on landing. When reloading, the landing conversation would be shown in its entirety (there's no plans to allow serialization of partial progress through a `Conversation`).

PR #4145, opened 2019-01-27
Fixed the laser turret hardpoint and projectile

Amazinite:
Ref: #4090 

An alternative fix, changing the hardpoint sprite and the projectile instead of the outfit image.

The projectile was simply changed by bringing the right laser close to fully overlapping with the left laser. It's a pixel or two off from completely overlapping to give the turret's projectile more thickness when compared to the gun's projectile.


tehhowch:
Copyright update & link to final assets should be in this PR

Amazinite:
Assets linked in OP. The change is so small that I'd rather just ignore the copyright.

tehhowch:
I think that's @Janaszar's call as the creator of the new sprite?

I will note here that asymmetric turrets won't properly spawn firing effects that match the sprite, since the current implementation of the `Weapon::HardpointOffset` is a single number denoting the vertical offset. https://github.com/endless-sky/endless-sky/blob/232b54b8ad81153abef5e5ddf7d0321efeca5857/source/Weapon.h#L168

It's possible we should change this to a `Point`. It's used in 2 places, equivalently:
https://github.com/endless-sky/endless-sky/blob/232b54b8ad81153abef5e5ddf7d0321efeca5857/source/Hardpoint.cpp#L195-L203

With a `Point` specification, one would have a horizontally symmetric projectile instead of one positioned asymmetrically to match the hardpoint sprite.

Amazinite:
The projectile does come from a pixel or two to the right of the actual barrel. It's not too noticeable, but if it's something that you can fix then go ahead.

Janaszar:
All I did was render it from the Blender file that I believe @Amazinite got from the data repository.  As far as I can tell, he post-processed it himself.
As such, I don't really see it as needing a copyright change with regards to what I did.

Nescio0:
Personally I still prefer #4090 instead of this one. The laser turret belongs to the same family as the heavy laser and the electron turret, therefore I don't see why it has to follow a different design.

Pointedstick:
We now have two good options. Let's choose one of them and move on. There's no need to nitpick everything to death. :)

tehhowch:
With 23015a50 the projectile sprite can be made symmetric, and a horizontal offset supplied, using standard X-Y chart orientation (+y is up, +x is right).

The AI aiming code doesn't account for the hardpoint offset--be it symmetric or not--so that might be something we consider adding in the future. (The smaller the offset, the less significant the issue.)

PR #4151, opened 2019-01-28
Arfecta has fuel generation to allow cloaking

Nescio0:
Give the Arfecta fuel generation (introduced in #3870) instead of negative cloaking fuel. This means the Arfecta can still cloak indefinitely without spending any fuel and when it's not cloaked, it's fuel increases automatically; furthermore, players can safely remove the cloaking device from captured Arfectas.


Amazinite:
This conflicts with the intended purpose of the negative cloaking fuel in the first place. https://github.com/endless-sky/endless-sky/issues/1344#issuecomment-234128902

Issue #4149, opened 2019-01-28
Concerns over the changes to missiles from v0.9.8 to now


Amazinite:
Just an issue to discuss the balance changes made to missiles. I think we're in a far better place now than we were before, but there are some issues with the new balance system that need ironed out.

Ref: #3866 (missiles buffed), #3922 (missiles made more numerous), #3845 (buffed Javelin fire rate), #4162 (reduced frequency of missile usage in pirate fleets and made small tweaks to various missiles)
#4077 (similar issue)
<hr/>

<details><summary>Initial concerns</summary>
List of concerns:

* Missiles are now too numerous.
   * The solution to this may be to simply revert many of the changes done in #3922. (Not revert it fully, though.)
* Missiles are now too powerful.
   * #3866 gave missiles a roughly 20% DPS buff. Perhaps lower this to a 10% buff, of even down to 0% (slight changes only being due to wanting round numbers).
* Look at usage of Sidewinders vs effectiveness/cost of jamming.
   * Sidewinder missiles are very difficult for small ships without jamming to dodge. That's exactly the kind of ship a player flies early game. Either revert to pirates only having meteor missiles, or look at the effectiveness and/or cost of jamming such that it's easier to use early game.
* Heavy rockets are now too powerful in terms of single shot damage. Early game, interceptor-sized ships are too easily disabled or outright destroyed. 
   * Solutions that I can think of would be to either keep the DPS similar by reducuing the single shot damage while increasing the fire rate such that a player is able to survive a single shot, or reduce the trigger radius of heavy rockets to make them more precision weapons when fighting smaller ships, but still be very capable of attacking larger ships.
* Look at Javelin balance.
   * Not something I've looked into since the changes, but I've received reports that the Javelin may be a bit too powerful now.
</details>
<hr/>
Concerns post-#4162:

* Look at Javelin balance
  * Are Javelins too weak? Do they need a further DPS buff above 33% but below the 68% that they had?
  * Or do Meteors and Sidewinders need to be brought down to "make room" for the Javelin as a viable weapon?
* Are missiles in general, or at least just the small ones, too strong?
  * Strong missiles allow smaller ships to compete against larger ships, but is it too much? Should the smaller missiles (Javelin, Meteor, Sidewinder) be able to directly compete with the three P cannons (Particle, Proton, Plasma), or should they be just under them?

Reference any other issues below that have changes missile balance that I've missed, and add your own concerns as well.

Pointedstick:
[...]

PR #4155, opened 2019-01-30
give default Star Barge an anti-missile turret

Nescio0:
Star Barge has a basic anti-missile turret by default, offering it protection against javelins and meteors; also gave it X1 ion engines, making it slightly slower, but also somewhat cheaper and with a bit more free space

See also #4149


Pointedstick:
+1 on giving the Star Barge an Anti-Missile Turret by default. Makes sense since it's slow and lumbering.

-1 on removing the Meteor Missile Launcher from the Sparrow. This change would make it a largely worthless ship again with its default armament. Maybe create a mining version with two Beam Lasers instead.

Darcman99:
I would be in favour of giving the sparrow two beam lasers again rather than meteor missiles if the description classed it as a common asteroid mining ship, but the description clearly states that it is used for combat most times. If the sparrows description were rewritten to class it as an asteroid mining ship in general and also states that some ambitious pilots attempt real combat in it then I would see the beam lasers as a better stock choice. I would actually like to see the sparrow become an asteroid mining ship rather than just be an early way into combat and a starting ship that not many people choose because of the early game risk.

Amazinite:
Like I said in #4149, we can give the Sparrow beam lasers stock, but have James tell the player about buying missiles if they're more interested in combat. The player experience is more important than the experience NPCs have.

Pointedstick:
> I would be in favour of giving the sparrow two beam lasers again rather than meteor missiles if the description classed it as a common asteroid mining ship, but the description clearly states that it is used for combat most times. If the sparrows description were rewritten to class it as an asteroid mining ship in general and also states that some ambitious pilots attempt real combat in it then I would see the beam lasers as a better stock choice. I would actually like to see the sparrow become an asteroid mining ship rather than just be an early way into combat and a starting ship that not many people choose because of the early game risk.

That's quite a clever and sensible idea. I'd be up for it.

Nescio0:
The Sparrow description has been updated to mention asteroid mining.
As for James, he currently advises you to run rather than fight, because your Sparrow is no match for larger ships, therefore left unchanged.

data/ships.txt
```diff
@@ -2804,11 +2802,11 @@ ship "Sparrow"
 	engine -5 35
 	engine 5 35
 	gun -7 -10 "Beam Laser"
-	gun 7 -10 "Meteor Missile Launcher"
+	gun 7 -10 "Beam Laser"
 	leak "flame" 60 80
 	explode "tiny explosion" 15
 	explode "small explosion" 5
-	description "Classified as an interceptor rather than a fighter because it has its own hyperdrive instead of needing to be carried inside a larger ship, the Tarazed Sparrow is the smallest and cheapest combat ship available. Because of its limited cargo and passenger space, the primary way for a Sparrow pilot to pay the bills is by hunting pirates... or by becoming one. In either case, it is a perilous way to earn a living."
+	description "Classified as an interceptor rather than a fighter because it has its own hyperdrive instead of needing to be carried inside a larger ship, the Tarazed Sparrow is the smallest and cheapest combat ship available. Although its cargo space is limited, the Sparrow's maneuverability makes it a decent ship for asteroid mining. Another way for a Sparrow pilot to pay the bills is by hunting pirates... or by becoming one. In either case, it is a perilous way to earn a living."
```
Pointedstick: How about this instead?

"Despite limited cargo space, the Sparrow's promotional materials advertise it as an asteroid mining ship, though many Sparrow pilots pay the bills by hunting pirates... or becoming one. "

Nescio0:
Updated.

data/ships.txt
```diff
@@ -2804,11 +2802,11 @@ ship "Sparrow"
 	engine -5 35
 	engine 5 35
 	gun -7 -10 "Beam Laser"
-	gun 7 -10 "Meteor Missile Launcher"
+	gun 7 -10 "Beam Laser"
 	leak "flame" 60 80
 	explode "tiny explosion" 15
 	explode "small explosion" 5
-	description "Classified as an interceptor rather than a fighter because it has its own hyperdrive instead of needing to be carried inside a larger ship, the Tarazed Sparrow is the smallest and cheapest combat ship available. Because of its limited cargo and passenger space, the primary way for a Sparrow pilot to pay the bills is by hunting pirates... or by becoming one. In either case, it is a perilous way to earn a living."
+	description "Classified as an interceptor rather than a fighter because it has its own hyperdrive instead of needing to be carried inside a larger ship, the Tarazed Sparrow is the smallest and cheapest combat ship available. Despite having limited cargo space, the Sparrow is advertised as a maneuverable asteroid mining ship. However, many Sparrow pilots pay the bills by hunting pirates... or becoming one. In either case, it is a perilous way to earn a living."
```
Pointedstick: Sorry, one more proposed change to this sentence:

"However, many Sparrow pilots pay the bills by upgrading the weapons and hunting pirates... or becoming pirates themselves."

Nescio0:
Updated.

Pointedstick:
Looks good to me except for the Star Barge engine change. Not sure this is needed. The Star Barge is already slow enough. Making it even slower doesn't seem like it accomplishes anything.

Nescio0:
The purpose of the engine change is to make the Star Barge cheaper, consume less energy, and free up space. For comparison:
* Chipmunks, no turret: 375000 cost, 20/130 free space, moving+firing 90+0 energy
* Chipmunks, AM turret: 435000 cost, 4/130 free space, moving+firing 90+37.5 energy
* X1 ions, AM turret: 421000 cost, 11/130 free space, moving+firing 54+37.5 energy

Starter loan is 480000; players who want a faster Star Barge can install different engines if they like, but for new players saving a bit of cash and space by default is a good thing.

Pointedstick:
If þe default Star Barge price is now too high with the addition of an Anti-Missile Turret (which I think I agree with you about), then swapping out the engines like you've done only saves 14k, which isn't that much considering that the ship has still been made 46k more expensive than it used to be. Probably the base cost needs to be reduced instead of or in addition to the engine change.

Nescio0:
It's not too high: the Sparrow is still more expensive (with two beam laser it costs 448000 and has 4/130 free space), but I do think changing the engines to save some credits, space, and energy is helpful. It could be reverted, if you believe that's better.

jostephd:
If the default Star Barge has an AM turret then the player can sell it on New Boston before they launch, but if the default Star Barge has no AM turret then the player can't easily acquire one, since they aren't on sale at New Boston. So maybe add the AM turret but have James hint that you can sell it?

Amazinite:
#4162 handled the Sparrow, so the only other change from this PR to discuss is whether to give the Star Barge an AM turret by default.

Pointedstick:
+1 on the general idea of that, if we can work out all the details and avoid making the Star Barge too expensive (which also makes life harder for starting players and runs contrary to the goal of the patch).

Pointedstick:
Let's continue with these sorts of changes in https://github.com/endless-sky/endless-sky/pull/4218 and https://github.com/endless-sky/endless-sky/pull/4219.

PR #4156, opened 2019-01-30
tweaked javelin

Nescio0:
Currently javelins are rather useless against ships with anti-missile systems, because even a single basic anti-missile turret (size 16) can take out two javelin pods (size 2×12=24). This proposal:
* fires javelins in bursts of five: all five hit within 0.25 s (15 frames), after which no javelins are fired for 1.0 s (75-15=60 frames)
* <s>changes it from four shots per second to five, which means it'll empty in 40 seconds instead of 50</s>
* reduces per shot shield and hull damage to the old values prior to #3866

As a result, individual javelins inflict less damage per second against players without AM turrets, but more javelins could hit ships that have AM turrets

Before:
screenshot from 2019-01-30 18-02-37

First proposal:
screenshot from 2019-01-30 18-01-41

Current proposal:
screenshot from 2019-02-01 11-43-57

See also #4149

### DIFF
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -774,11 +774,13 @@ outfit "Javelin Pod"
 		"inaccuracy" 1
 		"velocity" 10
 		"lifetime" 60
-		"reload" 15
+		"reload" 12
+		"burst count" 5
+		"burst reload" 3
 		"firing energy" .2
 		"firing heat" 12
-		"shield damage" 57
-		"hull damage" 36
+		"shield damage" 48
+		"hull damage" 26
 		"hit force" 50
 		"missile strength" 2
 	description "The Javelin Pod fires a rapid stream of unguided missiles. A Javelin Pod does far more damage than any gun of a comparable size, and has a longer range as well, but once a ship has expended all of its ammunition it must leave combat and find a planet where it can buy more, which makes Javelins less useful in protracted battles."
--- a/data/weapons.txt
+++ b/data/weapons.txt
@@ -774,7 +774,7 @@ outfit "Javelin Pod"
 		"inaccuracy" 1
 		"velocity" 10
 		"lifetime" 60
-		"reload" 12
+		"reload" 15
 		"burst count" 5
 		"burst reload" 3
 		"firing energy" .2


jafdy:
There is a 0.7 chance of a javelin being destroyed by a basic anti-missile strike, so a basic anti-missile will on average shoot down 5.25 javelins per second (if the wiki is to be believed).

Zitchas:
I was under the impression that javelins were supposed to fill two roles:
a) supporting dumb fire projectiles to overwhelm defences so that heavier accompanying stuff can get through.
b) cheap-as-dirt dumb fire projectiles for use on fighters and other ships agile enough to aim decently and who can't afford the space and/or credits for anything heavier.

They already function in both capacities right now, I don't really see the need to change things until the whole missile issue is resolved as a whole.

Nescio0:
a) giving javelins burst behaviour actually makes them better at overwhelming AM turrets
b) space: javelin pods are larger than meteor launchers (which have homing and 9× range); credits: 50 per shot is a lot more than guns (which shots are free and unaffected by AM turrets)

Amazinite:
>a) giving javelins burst behaviour actually makes them better at overwhelming AM turrets

I'd actually argue that it makes it less effective at overwhelming AM turrets. A single burst will be able to overwhelm a turret and slip some damage in, but the lack of sustained fire means that between bursts AM turrets are free to fire on any other missiles, missiles that will be doing more damage than the Javelin per shot. (One lost Javelin isn't much, but one lost Torpedo or Meteor is.)

With the missile strength changes in #4162, I don't see this needing to be added.

Nescio0:
#4162 reduced the damage of the javelin, but not of the other missiles. As a result javelins are relatively useless compared to other missiles, therefore they deserve something in compensation.
This proposal lets the javelin pod fire in bursts of five. The time between two bursts is 60 frames, i.e. the same as the time between two sidewinder shots. 
(Also, I don't believe there is any variant that has both javelins and torpedoes.)

Amazinite:
>#4162 reduced the damage of the javelin, but not of the other missiles. As a result javelins are relatively useless compared to other missiles, therefore they deserve something in compensation.

Can you please respond to the points I'm making to counter you claiming that the Javelin is "relatively useless compared to other missiles" instead of continuing to repeat that point. 

>(Also, I don't believe there is any variant that has both javelins and torpedoes.)

There aren't any single ships with both Javelins and Torpedoes, but there are fleet variants both on different ships. Using Javelins to distract AM to allow Torpedoes to hit is also a strategy for the player to use.

<hr/>

Quotes from #4162 

>If you want an argument as to why one would use Javelins with this PR as is though, the capacity means that over time it will out damage the Meteor launcher and firing 5.33 times faster means it can more effectively distract AM turrets.

>> So the javelin pod can sustain 50 s by default and the meteor now lasts 38 s instead of 44 s? Honestly, I don't think capacity is a decisive factor for players.
>It's more so the fact that the Javelin can deal more damage before needing resupplied than the Meteor because of the fact that the Javelin can fire for longer.

>In regards to the Javelin, as I've pointed out with how the Sidewinder compares to the Meteor, DPS isn't everything. The Sidewinder has less DPS than the Meteor, but is so much better in so many other ways that in practice it is a more effective weapon. Looking at the changes to missiles overall, I think that the way that the Javelin can be seen as more effective is its ability to overwhelm AM in a way that no other missiles can do now given that they've practically all had their rates of fire decreased while AM turrets were untouched. There's also the longevity that I mentioned the Javelin having over the Sidewinder and Meteor, being able to stay in a fight for longer and at the end of the day dish out more damage before running out of ammo than those two launchers.

>To restate what I've said before, since the Javelin has had its rate of fire increased while other missiles have had their rate of fire decreased, but AM turrets have been untouched, the Javelin is now the most effective at distracting AM turrets. This also means that it takes less Javelin Pods to effectively slip past AM turret defenses, meaning that in practice under certain circumstances, Javelin Pods will be able to out DPS Meteor Launchers.

<hr/>

On paper DPS is not everything.

A basic AM turret can take down on average 4.5 Javelins per second. The Javelin fires 4 times per second. This means that two Javelins is guaranteed to be able to get past a single basic AM turret, while a single Javelin will occasionally be able to get past a single basic AM turret since its fire rate is close to the average take down rate of the basic AM.
A basic AM turret can take down on average 2.475 Meteors per second. The Meteor fires 0.75 times per seconds. That means it takes 4 Meteor launchers before you can overtake the average take down rate of a basic AM turret, and a single Meteor launcher isn't going to often slip past a single basic AM turret.

A Heavy AM turret can take down on average 7.5 Javelins per second, meaning 2 Javelin Pods can still overtake a single Heavy AM turret.
A Heavy AM turret can take down on average 5.625 Meteors per second, meaning you need 8 Meteors to overtake.

(@Jafdy for the math behind average take downs per second, but I trust his numbers.)

Those are just simulated numbers for the average case, and there are many other variables that one simply can not take into account in a simulated scenario that affect the hit rate of missiles, aside from the fact that not every case is the average case. But at the very least this shows that the Javelin is *far* superior at overwhelming AM than the Meteor, meaning that the Javelin does have use over at least the Meteor.

<hr />

TL;DR what makes the Javelin useful over other missiles is 1) it's longevity, being able to dish out more damage than the Meteor and Sidewinder before running out of ammo, and 2) it's FAR superior ability to overwhelm AM turrets due to its far higher fire rate compared to other missiles, not only in so far as allowing itself to damage a ship, but more so in allowing other, more powerful damage per shot missiles to damage a ship. If we want to list a third benefit of the Javelin I'd say 3) its lower range means that it's more difficult to dodge. You've been saying that the lesser range of the Javelin is a detriment to it compared to the Meteor, but it is in fact part of what can make the Javelin better, as firing up close is easier to hit than firing from far away, even with the Meteor's homing, which is notably poor.

The on paper combined DPS of the Meteor is only ~6.4% greater than the Javelin, but the in practice DPS when facing AM turrets is going to be greater for the Javelin on average. 

As a comparison, the Sidewinder is superior to the Meteor for the same reason, really. The Meteor does ~3.3% more combined DPS than the Sidewinder, but the Sidewinder is far more likely to land a hit.

Nescio0:
You're repeating yourself. Yes, damage per second is not everything. However, one question remains: why would anyone want to use javelins now?
If one wants to maximize potential damage in small ships, meteors are the obvious choice (long range and homing). If one wants cheapness, longevity, or reliably hitting ships with anti-missile turrets, then use beam weapons.

Amazinite:
>You're repeating yourself. 

I'm repeating myself because you're ignoring what I'm saying and repeating yourself.

>However, one question remains: why would anyone want to use javelins now?

Please see my above post.

Pointedstick:
> You're repeating yourself. Yes, damage per second is not everything. However, one question remains: why would anyone want to use javelins now?
> If one wants to maximize potential damage in small ships, meteors are the obvious choice (long range and homing).

Actually I think you're wrong about that. Meteors fired by the AI are hugely inaccurate and most will miss, so effective DPS is quite low. With the launcher's much shorter range, far more Javelins will hit their targets.

If you really want to maximize damage potential in small ships, I'd say you should cram as many Heavy Rocket launchers as you can into them. Sidewinders are probably second due to their high accuracy and anti-missile resilience, followed by Javelin Pods. Meteors are actually a pretty lousy choice for small craft piloted by the AI.

Tadrix:
> therefore they deserve something in compensation.

They already got some compensation though. In missile strength. (There are other things that can be done, like lowering the ammo price, since Javelins are clearly a missile spam weapon, but it's beside the point in this PR,)

Beam weapons aren't going to saturate AM defences for your fleet's heavy hitters to do their job. Javelins will.

Janaszar:
Just a thought...

If I was going to use Javelins, I would be using them in an Anti-Anti-Missile role, with many quick, small, and very cheap ships.  Say 20 sparrows just packed with Javelins.  I would be treating the whole system (ship, javelins, ammo) as essentially an expendable weapon, to allow stronger weapons, say Finisher pods or Torpedoes, fired from more expensive ships, to hit the targets.  In that case, the DPS from the Javelins is secondary to their goal of fouling the anti-missile systems of opposing ships.

Pointedstick:
I've had very good luck doing just that, in fact. I stumbled on the tactic almost by accident once and it's what impelled me to add that into the Javelin Pod's description.

It works great when you do it as a deliberate tactic. I once bought and outfitted a small escort fleet of 4 Clippers armed with nothing but Javelin pods to help my up-til-that-point-lone flagship win some big battle during the Free Worlds campaign. They saturated the anti-missile defenses of a bunch of Leviathans, allowing me to blast them to pieces with my flagship's Heavy Rocket launchers. It was fun as hell to see a plan come together like that. :smile:

Zitchas:
That reminds me of good'ol EV, where the clipper's main damage was from a rain of cheap javelins. Very effective at dealing with ships that used flares or jammers to nullify the guided weaponry.

PR #4157, opened 2019-01-30
give the Combat Drone a javelin pod

Nescio0:
Combat Drones are currently quite useless, because it is armed with only a single beam laser, which inflicts little damage and is short ranged. The CD is a dedicated combat craft, designed and used exclusively by the Navy. Because they don't have to accomodate a cockpit, living space, life support systems, fuel, or a hyperdrive, the CD should have plenty of space for a heavier weapon than a puny beam laser.

This proposal replaces the beam laser with a javelin pod and gives the CD +4 outfit and weapon space to accomodate it.
A single javelin pod is not very useful, however, CDs are deplayed in groups, allowing their javelins to overwhelm enemy AM turrets (especially in combination with #4156).

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -907,10 +907,10 @@ ship "Combat Drone"
 		"hull" 700
 		"automaton" 1
 		"mass" 20
-		"drag" .59
+		"drag" .6
 		"heat dissipation" .9
-		"outfit space" 58
-		"weapon capacity" 8
+		"outfit space" 62
+		"weapon capacity" 12
 		"engine capacity" 28
 		weapon
 			"blast radius" 5
@@ -918,7 +918,8 @@ ship "Combat Drone"
 			"hull damage" 25
 			"hit force" 75
 	outfits
-		"Beam Laser"
+		"Javelin Pod"
+		"Javelin" 200
 		
 		"nGVF-AA Fuel Cell"
 		"Supercapacitor"
@@ -928,7 +929,7 @@ ship "Combat Drone"
 		
 	engine -9 29
 	engine 9 29
-	gun 0 -29 "Beam Laser"
+	gun 0 -29 "Javelin Pod"
 	explode "tiny explosion" 15
 	description "Combat drones are pilotless attack ships used primarily by the Republic Navy. Although very weak and easy to destroy, they can be very effective in large numbers. Because drones do not need a cockpit, they can be filled entirely with equipment and solid metal, which makes their hulls stronger than other small ships."
 


Pointedstick:
Conceptually this makes much more sense.

However it probably makes combat drone swarms much more dangerous, especially to ships with no AM, and especially especially if we tweak Javelin Pods to be more effective against AM fire.

Probably needs substantial playtesting.

Amazinite:
Even if this change with Javelins is alright, this change allows for combat drones to also equip meteor missiles, which seems like too much of a buff to combat drone. That's aside from the fact that #4156 seeks to buff Javelins further than they've already been buffed. Combat drones are meant to be weak, so I don't see why we should be buffing them.

Pointedstick:
> Combat drones are meant to be weak, so I don't see why we should be buffing them.

Like all ships, Combat drones are designed to be effective within their original design parameters. They are deployed in a swarm of 4-6 and should be effective when deployed as such against their intended targets. I would posit that right now they are not. At the moment they are little more than a pointless, ineffective speed bump.

tehhowch:
Before anyone suggests limiting the installable capacity of ammunition for javelins / meteors by adding negative `"____ capacity"` attributes to the Combat Drone, note that this will effectively "bind" the launcher to the drone hull upon install. Any attempt to remove the launcher would then indicate that doing so would cause `<launcher ammo> capacity` attribute to become negative.
Maybe we still pursue that and rename the Combat Drone to `"Meteor Drone"` and `"Javelin Drone"` respectively, but in such a situation there should still be a laser-equipped `"Combat Drone"`.

Nescio0:
Right now Combat Drones are simply a joke; the only thing they can is blocking projectiles while they're exploding.

> Even if this change with Javelins is alright, this change allows for combat drones to also equip meteor missiles, which seems like too much of a buff to combat drone.

Actually I fail to see why that is a problem; for comparison, the Lance has a Sidewinder by default, which can be replaced with a Heavy Rocket - or *two* Meteors.
Besides, the Combat Drone requires a Navy licence (fighters don't), which means a player can't buy them; the only way to get a Meteor CDs is by first capturing a Carrier, then capturing CDs, and subsequently re-equip them in an outfitter; but that's a lot of work and at that point you've proven you're capable of handling swarms of CDs.

> It could drastically alter the balance of the FW campaign (and that TMBR mission with drones), without proper consideration this would be a major issue for new players.

Yes, the storyline might become slightly more difficult; however, the initial missions avoid a direct confrontation with the Navy; later on you're told to get a better ship.

Nescio0:
Actually I don't think adding negative capacities is a good idea.

Brick63:
barrage launcher 2x

By request of @Amazinite i'm posting this here, i saw this a while back and thought, why not give the navy a micro missile launcher to fit on the current version instead. This is the current sprite i have for it. @Amazinite is doing the stats.

Pointedstick:
That's pretty nice looking!

Amazinite:
Combat drones are designed to be weak and expendable, and buffing them only means more trouble for the player. Beam lasers deal 60 shield dps and 78 hull dps, while javelin pods deal 192 shield dps and 104 hull dps. That's a 33.33% buff to hull dps and a whopping 220% buff to shield dps! Looking just at the shield DPS, a single combat drone would need to have 68.75% of its missiles not land in order to match a beam laser's DPS (for hull, only 25% would have to miss), but when you're talking about deploying combat drones in bulk, most of those missiles are going to be landing. I don't really see this change adding anything positive to the game.

And while I oversaw the creation of this new micro-missile launcher, I'm more against adding it to the game now because, in addition to all the problems of giving a javelin to the combat drone, it would just become another headache of a weapon to balance among all the other human missiles that we're dealing with right now.

PR #4161, opened 2019-01-31
Add hand-to-hand to human Medium and Heavy Warships

Fzzr:
This was originally intended to give H2H to pirates, but per suggestion from @Amazinite it's easier to just add them to the defaults. I gave each ship laser rifles equal to 1/4 the required crew count, rounded up, floor 2. I also added H2H to the FW military at the same ratios as the civvie ones. Per discussion in #4125 the Free Worlds don't focus on H2H doctrine, so there's no reason for them to invest in H2H beyond generally accepted practices. Variants with Hai outfits received Pulse Rifles instead.

The Remnant and all non-human factions are already appropriately armed.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -1153,6 +1153,7 @@ ship "Dreadnought"
 		"D41-HY Shield Generator"
 		"Small Radar Jammer"
 		"Liquid Helium Cooler"
+		"Laser Rifle" 21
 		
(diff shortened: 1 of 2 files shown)


Fzzr:
> So why default ships and not variants as well?

Sure, done in 68653a8 with Pulse Rifles for variants that use Hai parts.

Issue #4158, opened 2019-01-31
Pirate fleet spawn strengths seem all over the place


Amazinite:
So here's a chart of the strength of various pirate fleets given as a percent chance that that fleet size can spawn. Although there is progression within each category, I'm only looking at by category as a shorthand for strength.

<details><summary>Big chart</summary>

  | Small South | Large South | Small Core | Large Core | Small North | Large North
-- | -- | -- | -- | -- | -- | --
1 Int | 57.14% |   | 70.27% |   | 59.46% |  
2 Int | 22.45% |   |   |   | 2.70% | 10.17%
3 Int | 10.20% | 4.69% |   |   |   | 5.08%
1 LW | 10.20% | 40.63% | 21.62% |   | 27.03% |  
4 Int |   | 9.38% |   |   |   |  
1 LW 1 Int |   | 17.19% |   |   |   |  
1 LW 2 Int |   | 6.25% |   |   |   |  
2 LW |   | 1.56% | 8.11% |   |   | 10.17%
2 LW 1 Int |   |   |   | 20.41% |   |  
1 MW |   | 12.50% |   | 24.49% | 10.81% | 18.64%
3 LW |   |   |   | 18.37% |   | 1.69%
1 MW 1 Int |   |   |   | 2.04% |   | 6.78%
1 MW 2 Int |   |   |   |   |   | 5.08%
1 MW 1 LW |   |   |   | 18.37% |   |  
2 MW |   |   |   | 8.16% |   | 13.56%
1 HW |   | 6.25% |   | 8.16% |   | 6.78%
1 HW 3 Int |   | 1.56% |   |   |   |  
1 HW 1 MW |   |   |   |   |   | 20.34%
Bactrian |   |   |   |   |   | 1.69%

</details>

And here's the ships that can spawn in each fleet.

<details><summary>Smol chart</summary>

  | Small South | Large South | Small Core | Large Core | Small North | Large North
-- | -- | -- | -- | -- | -- | --
[...]

PR #4162, opened 2019-01-31
Changes to missile balance and usage

Amazinite:
Ref: #4149 

I'm going to open this PR now to show what I'm working on and continue to update this PR in response to continued discussions over missile balance.

Current changes:
* Javelin Pod
  * Reverted damage buff from #3866 
     * The combination of the damage buff from #3866 and the fire rate buff from #3845 gave the Javelin Pod a 68% DPS buff, which was far too much. Reduces the DPS buff since v0.9.8 to 33%.
  * Increased missile strength from 2 to 3
     * Instead of changing the way the Javelin functions, as with #4156, increased the missile strength in order to make it more effective against AM. According to @jafdy, this change means that a basic AM turret can now shoot down on average 4.5 Javelins per seconds, nearly matching the 4 Javelins per second that are fired.
     * The chance for a basic AM turret to destroy a Javelin missile goes from 70% to 60%
     * The chance for a Heavy AM turret to destroy a Javelin missile goes from ~81% to 75%
* Meteor Missile Launcher
  * Increased missile strength from 4 to 6
     * Did this in order to keep the difference between Javelins and Meteors. This change also helps improve the usefulness of Meteors against larger ships, ref: https://github.com/endless-sky/endless-sky/issues/4149#issuecomment-457999729. Note that the Sidewinder, somewhat of a direct upgrade to the Meteor, has a missile strength of 12.
     * The chance for a basic AM turret to destroy a Meteor missile goes from 50% to 33%
     * The chance for a Heavy AM turret to destroy a Meteor missile goes from ~69% to ~56%
  * Decreased the missile capacity from 35 to 30
     * With #3866's changes to damage per shot and fire rate, the Meteor missile now outputs double the damage that it use to over the course of depleting its missile capacity. Reducing missile capacity by 5 reduced this buff from 100% to ~71%, making room for other weapons in terms of longevity, such as the Javelin. Ref: https://github.com/endless-sky/endless-sky/pull/4162#issuecomment-460744374
     * Also reduced the storage outfit capacity from 18 to 15.
* Sidewinder Missile Launcher
   * Decreased the missile capacity from 50 to 45
     * The reasoning behind this is the same as the reasoning for reducing the Meteor's missile capacity, listed above. The buff in damage output goes from ~79% to ~61%.
     * Also reduced the storage outfit capacity from 25 to 23.
* Heavy Rocket Launcher
  * Decreased trigger radius from 30 to 20
  * Reverted the velocity buff from #3866 
  * Increased the lifetime from 100 to 125
     * #3866 made Heavy Rockets too effective against smaller ships, a detriment to the early game. Reducing the trigger radius, reducing the velocity, and increasing the range makes Heavy Rockets easier to dodge while keeping them just as effective at what they were meant to do.
* Radar Jammers
  * Reduced the cost of Small Radar Jammers from 70k to 24k
  * Reduced the cost of Large Radar Jammers from 490k to 138k
     * Radar jammers are only an effective form of missile defense against Sidewinder missiles in human space. Sidewinders are also the missiles that give small ships the most trouble given their high speed and homing abilities. Due to the prohibitive cost of jammers for early game players, there is currently no defense for the small ships that players are expected to have against Sidewinders, leaving the player a sitting duck. Reducing the cost of jammers allows for early game players to actually use jammers and have a chance where it matters.
* Ship changes
  * Reverted the weapon changes made to the stock Sparrow, Fury, and Hawk in #3922
      * One of the problems caused by #3922 is that pirate fleets now have too many missiles, becoming a detriment to gameplay, especially early game. As highlighted [here](https://github.com/endless-sky/endless-sky/pull/4162#issuecomment-460097080), only changing these three ships greatly reduces the number of missiles that the player must encounter.
     * Also changed the end of the Sparrow's description slightly to suggest to players that their best chance of survival is to equip missile launchers, instead of changing the lore of the Sparrow as #4155 does to be mainly a mining ship that people for some reason just like to use as a pirate hunting ship that they have a high chance of dying in.
* Fleet changes
  * Small Core Pirates
    * Increased usage of missiles by increasing the spawn chance of the Wasp (Meteor) and (Javelin), as removing missiles from the Sparrow, Fury, and Hawk made missile usage in small Core pirate fleets too low.
    * Also removed the Fury (ref: #4158) to make room for the increase in Wasp spawns, keeping small Core pirate fleets the same strength relative to ship category.
  * Large Northern Pirates
    * Reduced usage of missiles by reducing the spawn chance of variants that contained missiles.
    * Also removed the only Beserker variants (ref: #4158). These changes have increased the chance of running into larger ships when a large Northern pirate fleet spawns.

Before the fleet and ship changes:

Fleet | Small South | Large South | Small Core | Large Core | Small North | Large North
-- | -- | -- | -- | -- | -- | --
% of missiles | 87.76% | 60.94% | 43.24% | 51.02% | 72.97% | 77.97%
Meteor | 67.35% | 39.06% | 24.32% | 32.65% | 37.84% | 59.32%
Sidewinder | 0.00% | 0.00% | 0.00% | 0.00% | 13.51% | 16.95%
Javelin | 14.29% | 10.94% | 8.11% | 6.12% | 21.62% | 22.03%
Heavy | 32.65% | 31.25% | 13.51% | 20.41% | 18.92% | 8.47%
Torpedo | 0.00% | 15.63% | 0.00% | 16.33% | 5.41% | 30.51%

After the fleet and ship changes:

Fleet | Small South | Large South | Small Core | Large Core | Small North | Large North
-- | -- | -- | -- | -- | -- | --
% of missiles | 32.65% | 54.69% | 35.14% | 51.02% | 45.95% | 54.72%
Meteor | 20.41% | 28.13% | 21.62% | 32.65% | 13.51% | 47.17%
Sidewinder | 0.00% | 0.00% | 0.00% | 0.00% | 13.51% | 9.43%
Javelin | 14.29% | 10.94% | 8.11% | 6.12% | 21.62% | 11.32%
Heavy | 18.37% | 23.44% | 5.41% | 20.41% | 13.51% | 3.77%
Torpedo | 0.00% | 15.63% | 0.00% | 16.33% | 5.41% | 26.42%

Considering changes:
 * Reduce the damage buff given to Meteors and Sidewinders from 20% to 10%, as outlined in #4149. Will do this if simply reducing missile frequency isn't enough.


Nescio0:
* javelin damage: no objections, also done in #4156 (which gives javelins burst behaviour)
* javelin missile strength: rather pointless, because all anti-missile turrets have strength of at least 5
* meteor missile strength: might help, although you possibly have to increase the Pug AM strength from 5 to 6 then
* rocket trigger radius: isn't that precisely the point? Quoting it's description: "highly effective against clusters of fighters."

Amazinite:
>javelin missile strength: rather pointless, because all anti-missile turrets have strength of at least 5

Missile strength vs anti-missile isn't "highest number always wins." This increases the chance of a Javelin surviving against a basic AM shot from 30%, as Jafdy mentioned in #4156, to 40%.

>meteor missile strength: might help, although you possibly have to increase the Pug AM strength from 5 to 6 then

Meteor missiles aren't really used to such an extent during the Pug invasion as to warrant that.

>rocket trigger radius: isn't that precisely the point? To quote its description: "highly effective against clusters of fighters."

It's still effective against clusters, as the blast radius wasn't changed. This is simply making the Heavy Rocket less effective against single ships, as a single ship can more easily dodge the rocket. Now that Heavy Rockets can one shot disable or kill certain starter ships, it's beneficial to make it easier to dodge the heavy rocket for the sake of the early game.

Zitchas:
I'm not sure why heavy rockets would make sense for the argossy, Bastion, and Dreadnought and not interceptors. dumb-fire weapons (especially those that are powerful but take up a lot of space and thus aren't very numerous) make a lot more sense on small agile things that have better aim and can get the rocket in closer before firing.

solardawning:
I'd suggest a range increase and lowered velocity on Heavy Rocket, too.

A big part of the 'sudden death' rocket danger is that you don't really see it coming. A hawk is flying at you, then you're dead. It's a communication issue- reloading doesn't feel as bad when you know why you died, what happened, and that it was avoidable.

A range increase of 200 or so and lowered velocity would help make it a more learnable experience.

Pointedstick:
@solardawning I think that makes a lot of sense.

Zitchas:
I agree with that. I like the idea that they *can* go for almost forever, and are going to inflict a lot of hurt on whatever they hit (up to and including one-hit-kills for smaller ships), but that they are slow enough that a decently agile ship should be able to dodge them in a one-on-one fight if they have any kind of warning.

DingusShingleton:
> fire and it goes forever until it hits something


Especially considering (and I know the connection to physics in this game is tenuous at best) that a dumb fire rocket in real life would literally just accelerate until it ran out of fuel and then coast forever.

Not that I think it should be like that in game, I'm just saying, it's a thing.

Zitchas:
That sounds like a good step for now. We can adjust more later if we decide it is needed - no sense making giant changes when that was part of the reason for this re-calibration.

Nescio0:
Because rockets are unguided, increasing their range may actually make them less effective; 8×150=1200 or 7×180=1240 could be fine too.

Zitchas:
It all depends on how they are used. The slower they are, the easier to dodge - and the AI doesn't really dodge, so that's just a bonus to human players.

Pointedstick:
https://github.com/endless-sky/endless-sky/pull/4155 removes the meteor launcher from the default Sparrow which I think is reasonable given the lore changes accompanying it. I don't like the idea of removing the missile weapons from all interceptors' default armaments because though, because that would take us back to where we were before the changes, when interceptor-sized ships were combat-ineffective because of their poor size-to-damage-output ratio.

Especially if https://github.com/endless-sky/endless-sky/pull/4155 lands, then the Star Barge will have an Anti-Missile turret by defaut, which is much more sensible than a Blaster Turret and further reduces the danger of missiles to the starting ship that is most vulnerable to them.

Amazinite:
If all that we did was remove Meteors from the stock Sparrow and Fury and rockets from the stock Hawk, missile usage in pirate fleets would look like this.

Fleet | Small South | Large South | Small Core | Large Core | Small North | Large North
-- | -- | -- | -- | -- | -- | --
% of missiles | 32.65% | 54.69% | 18.92% | 51.02% | 45.95% | 77.97%
Meteor | 20.41% | 28.13% | 8.11% | 32.65% | 13.51% | 59.32%
Sidewinder | 0.00% | 0.00% | 0.00% | 0.00% | 13.51% | 16.95%
Javelin | 14.29% | 10.94% | 8.11% | 6.12% | 21.62% | 22.03%
Heavy | 18.37% | 23.44% | 5.41% | 20.41% | 13.51% | 8.47%
Torpedo | 0.00% | 15.63% | 0.00% | 16.33% | 5.41% | 30.51%

Large North would be the only one that needs any further tweaking (likely just lower usage of `(missile)` variants) to get the missile usage around 50%. Small Core could also use some love to get missile usage closer to say 33% like small South.
That isn't even taking account of the fact that removing the missiles from the specified ships also reduces the severity of missiles when a missile fleet is spawned, as some fleet variants had say one light warship with missiles and two or three interceptors with missiles, so removing the missiles from the interceptors greatly reduces the number of missiles from such a fleet.

Nescio0:
> I’m trying to reduce missiles usage to at most 50% though

Aiming for 50% is sensible for large pirate fleets, although I recommend to bring small pirate fleets to 50% gatling, 25% missile; pirate fighters can be 100% gatling variants.

>  that would take us back to where we were before the changes

Which is not necessarily bad. #3922 has been implemented a few months ago and now a larger number have played with it; it was worth the try, but apparently it creates more issues than it solves, therefore partially reverting it makes sense. Basically:
* NPC fleets can use any variant and do not even have to use the default stock model
* players who want to do pirate hunting will probably change weapons to full homing missiles
* players who want to do asteroid mining will probably change weapons to only ordinary guns
* players who want to do transport jobs  will probably remove all weapons
* new players often haven't figured out how to change outfits and are therefore stuck with lop-sided half gun, half missile load-outs, which aren't really great at anything

So yes, restoring interceptor load-outs to what they used to be makes sense:
* Sparrow: beam lasers
* Wasp: energy blasters
* Berserker: beam lasers
* Fury: modified blasters
* Hawk: heavy lasers

Besides, it makes weapon configurations symmetrical again, which is aesthetically more pleasing.

Amazinite:
Well the Berserker and Wasp are already symmetrical with their current loadouts, and the stock Wasp isn't in any pirate fleets while the Berserker is more rarely used than those other interceptors, so I don't find it much of a problem if it keeps its Javelins.

Naminson:
I playtested it for a while, not quite enough to be 100% certain, but I have played around enough that I feel like it is much better than the current state of the missiles. I could feel that 50% drop very quickly, and it makes the game much more fun since you don't have get bumped around by interceptors every time you enter the south now.

The heavy rocket changes were what I was most interested in, since I found them annoying even before they were changed. So most of it was just playing around in a rocket hawk, and I think they've been made much more fair now. They still have that same high damage per shot that I hate, but the lowered velocity and trigger radius makes them feel much more fair, and should make players realize that it is possible to dodge them.

Nescio0:
* fleet changes: no objections
* radar jammers: large used to be exactly 7× as expensive as small, so perhaps change it to 140000 and 20000.
* #3866 increased the damage of all missiles by about 20%; this proposal reduces javelin damage, improves meteor missile strength, and keep others unchanged. Currently we have:
screenshot from 2019-02-05 10-44-38
With your proposal it becomes:
screenshot from 2019-02-05 10-47-58
In other words, javelin pod compared to meteor:
  * 20% larger
  * 33% more expensive
  * −88% range
  * less damage per shot and per second
  * no homing
Then why would anyone want to use javelins now?

data/ships.txt
```diff
@@ -2803,13 +2797,13 @@ ship "Sparrow"
...
-	gun -7 -10 "Beam Laser"
-	gun 7 -10 "Meteor Missile Launcher"
+	gun -7 -10
+	gun 7 -10
 	leak "flame" 60 80
 	explode "tiny explosion" 15
 	explode "small explosion" 5
 	description "Classified as an interceptor rather than a fighter because it has its own hyperdrive instead of needing to be carried inside a larger ship, the Tarazed Sparrow is the smallest and cheapest combat ship available. Because of its limited cargo and passenger space, the primary way for a Sparrow pilot to pay the bills is by hunting pirates... or by becoming one. In either case, it is a perilous way to earn a living."
-	description "	Insurance reports indicate that as many as two out of every three first-time ship buyers who choose to pilot a Sparrow lose their ship (and often, their life as well) within the first month of owning it."
+	description "	Insurance reports indicate that as many as two out of every three first-time ship buyers who choose to pilot a Sparrow lose their ship (and often, their life as well) within the first month of owning it. Pilots who have survived their first month tend to swear by their missile launchers."
```
Pointedstick: How about rewriting the new sentence as "Pilots who have survived longer than that tend to swear by replacing the stock Lasers with missile launchers"

Amazinite:
#3866 increased the DPS of the Meteor, Sidewinder, Torpedo, and Typhoon by almost exactly 20%. The Heavy Rocket only changed about 1%, and the Javelin changed about 26%. Then the Javelin got a 33% rate of fire buff, bringing the Javelin’s DPS buff to 68%. The 68% buff was FAR too strong, so clearly if reducing it to a 33% buff makes the Meteor stronger, the Javelin was too weak to begin with. 

The only action I would take to “fix” this would be to put into place the reduction in Meteor and Sidewinder damage to make their 20% buffs 10% instead. 

If you want an argument as to why one would use Javelins with this PR as is though, the capacity means that over time it will out damage the Meteor launcher and firing 5.33 times faster means it can more effectively distract AM turrets.

data/weapons.txt
```diff
@@ -777,10 +777,10 @@ outfit "Javelin Pod"
 		"reload" 15
 		"firing energy" .2
 		"firing heat" 12
-		"shield damage" 57
-		"hull damage" 36
+		"shield damage" 48
+		"hull damage" 26
```
Pointedstick: I have to agree with @Nescio0 here: with these changes, there's very little reason to use Javelins anymore. If the damage needs to be reduced, I would recommend doing it by not so much-- or not doing it at all.

Amazinite: As in my reply to him, if the only way to make the Javelin viable over the Meter is to give the Javelin it’s 68% DPS buff, then the Javelin was too weak to begin with. Given that the Javelin being buffed that much caused complaints (and honestly I think it’s too much given that it makes a 12 ton weapon the highest DPS human weapon outside of nukes), I’d rather go about reducing Meteor damage.

Pointedstick: >  then the Javelin was too weak to begin with

I think it certainly was.

> I’d rather go about reducing Meteor damage.

Especially with reduced frequency, I don't think Meteor damage will be a problem.

Amazinite:
The reason one use to want to choose the Javelin over the Meteor had to do with longevity. In v0.9.8 and prior, the Javelin could fire for 66.33 seconds before running out of ammo, in that time dishing out 9.6k shield damage or 5.2k hull damage. By comparison, the Meteor use to be able to fire for only 27.2 seconds, dealing 4.55k shield damage or 2.8k hull damage. The combined DPS (shield+hull) of the Javelin was 222 and the Meteor's was 262.5. So even then. the Meteor would out DPS the Javelin, but the Javelin had superior longevity.

Now though, given the changes, the Meteor now takes 45.33 seconds to empty, dealing 9.1k shield damage or 5.6k hull damage, and while doing 312 dombined DPS. Under this PR, the Javelin depletes in 49.75 seconds, dealing 9.6k shield damage or 5,2k hull damage with a combined DPS of 296. So while the Meteor still does more DPS, the Javelin is now closer to the Meteor than it use to be. But because of the rate of fire changes to the Meteor, the Javelin now does not have as much of an advantage in longevity.

Amazinite:
Given those facts, it may be that ammo capacity needs a look over on the Meteor and perhaps Sidewinder as well.

Amazinite:
I think that reducing the capacity on the Meteor and Sidewinder by 5 wouldn't hurt (and increasing the launcher mass by 1 to compensate).
Meteor: 45.33 seconds to 38.67, 32.67 before #3866; 9.1k shield to 7.8k, 4.55k before; 5.6k hull to 4.8k, 2.8k before
Sidewinder: 49 seconds to 44, 32.67 before; 9k shield to 8.1k, 5k before; 6.25k hull to 5.626k, 3.5k before

Amazinite:
Now = this PR as is
w/ change = with the suggested decrease of the ammo capacity
Before = before #3866 
TTD = time to deplete
Shield or Hull = total shield or hull damage dealt before running out of ammo

Weapon | TTD now | TTD w/ change | TTD before | Shield now | Shield w/ change | Shield before | Hull now | Hull w/ change | Hull before
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
Meteor | 45.33 seconds | 38.67 | 27.2 | 9100 | 7800 | 4550 | 5600 | 4800 | 2800
Sidewinder | 49 seconds | 44 | 32.67 | 9000 | 8100 | 5000 | 6250 | 5625 | 3500
Javelin | 49.75 seconds |   | 66.33 | 9600 |   | 9600 | 5200 |   | 5200

Nescio0:
> Then the Javelin got a 33% rate of fire buff, bringing the Javelin’s DPS buff to 68%.
Actually #3845 was merged *before* #3866 was proposed. To clarify, I'm not opposed to reverting the javelin damage increase; however, if that happens, then meteors should be made less effective too, or javelins should get something else in compensation (e.g. burst behaviour as in #4156).

Nescio0:
Then javelins become effectively useless, because the smaller meteor is more powerful per shot and per second (and that's disregarding range and homing).

Amazinite:
I'm all for a damage nerf to the Meteor, and by extension the Sidewinder since these two weapons are so closely tied. I'd just need confirmation from @Pointedstick that it's an okay change. In order to get the Javelin to have a higher DPS than the Meteor, then the 10% plan would need to be used as referenced at the bottom of the PR description.

Nescio0:
Actually I think the sidewinder is fine, because it's larger and more expensive. If only the meteor's reload was increased from 80 to 90, we would get the following damages per second:
* meteor (10): 173.3 shield, 106.7 hull (instead of 195, 120)
* javelin (12): 192 shield, 104 hull (instead of 228, 144)
* sidewinder (14): 180 shield, 125 hull (unchanged)

Amazinite:
Yes, the Sidewinder is 2 tons heavier than the Meteor, but it also has:
* Far better tracking
* A higher capacity that leads to more longevity and a higher damage output
* Much more difficult to destroy missiles (12 strength vs 6 or 4)
* A higher fire rate, which on top of the missile strength and tracking simply means that it's FAR easier to land the hit

I don't think the Sidewinder also needs a higher DPS given how superior it is in a number of other ways already.

Nescio0:
The sidewinder is 40% larger than the meteor. And yes, it's clearly a superior weapon. I'm not opposed to reducing the sidewinder's damage, but I think it's not as necessary as the meteor's. Alternatively, the javelin could get more damage.

Also, if the meteors capacity is reduced to 30, then the sidewinder's should probably become 40.

Zitchas:
Why should Javalines forcibly be superior in some situation? That is kind of like insisting that the standard blaster canon must be the best at something other "Cheapest". Javelins are the blasters of the missile world, and I think they should stay there. No need to increase their DPS or anything else, except downward as necessary to keep them as lowest dps when other missiles get adjusted.

Amazinite:
Alright. I'll leave this open a little longer to allow for any more thoughts and reviews over the next couple days as people playtest. I still have my eye on the damage changes to the Meteor and Sidewinder, though.

Nescio0:
> That is kind of like insisting that the standard blaster canon must be the best at something other "Cheapest"
Actually they are: blasters have the highest damage per mass of all human guns; they also have a longer range than most guns; only the proton and particle cannons reach further.

Pointedstick:
The reason why I don't favor reducing that is because it worsens the emotional **feel** of missiles.

We sometimes focus too much on stats and numbers and not enough on how something feels, and one of the major problems with missiles before was that they *just didn't feel right*. One of the major reasons for that negative feeling was because they were high fire-rate and low-damage--basically the opposite of most missiles today, whose inherent characteristics favor ever larger missiles that do more damage with lower numbers and fire rates. DPS is almost the wrong metric to be examining when you look at missiles (except for the Javelin, which has more in common with guns).

On another note, getting into Endless Sky has caused me to become a naval, aviation, and missile combat history nerd, and I've learned more about these topics in the last year than I ever imagined. Obviously we don't want this game to become too crunchy and lose its charm, but we also need for things to feel roughly how their inherent characteristics suggest they should. A heavy freighter whose default speed is faster than a capital ship would just feel *wrong*; it's the same thing with missiles that don't do enough damage on a per-missile basis.

Nescio0:
> they were high fire-rate and low-damage

Actually they still are: a single plasma shot inflicts 63 shield damage (javelin is reduced from 57 to 48) and an ion shot 168 (sidewinder 180). Personally I'd welcome reducing missile shots per second and increasing their damage per shot.

Pointedstick:
> > they were high fire-rate and low-damage
> 
> Actually they still are: a single plasma shot inflicts 63 shield damage (javelin is reduced from 57 to 48) and an ion shot 168 (sidewinder 180). Personally I'd welcome reducing missile shots per second and increasing their damage per shot.

You're comparing missiles to the largest guns, which *should* roughly be their equals. If a sidewinder does about as much damage as an ion cannon hit, that sounds about right to me, because historically speaking, missiles are the successor to large fixed guns.

In the days of yore, the largest 16" battleship guns could fire a 2,700 pound armor-piercing shell with a 158 lb bursting charge, which was utterly devastating, able to penetrate 10" of hardened steel armor or produce 50 foot craters on land when the fragmentation version was used. By contrast, modern anti-ship missiles are not armor-piercing and generally have warheads of between 100 and 1,000 pounds, but actually do less damage on a pound-for-pound basis due to the lack of a hardened metal casing to focus the pressure. In other words the damage dealt to modern targets is roughly comparable.

The advantage of missiles has never been that they do more damage compared to the largest guns, or even that they're more accurate (accuracy-wise, many to most missiles still miss their targets, even today). Rather, missiles' advantage is the lower weight of their firing platforms which allows lighter vessels to mount them, and their their greater range which allows the shooter to be safer from harm.You couldn't put a 16" battleship gun on a fighter or score hits on targets 150 miles away. But you can stuff a fighter or a patrol boat full of missiles that allow them to threaten much larger targets than themselves from very far away.

Amazinite:
>The reason why I don't favor reducing that is because it worsens the emotional **feel** of missiles.

There's the feel of using missiles, and there's the feel of having missiles used on you. I think that the change to fire rates did a lot to improve the feel of using missiles, but increasing the DPS was more detrimental to the feel of having missiles used on you.

Amazinite:
Any last words? I think I'll be merging this some time in the next 24 hours if there aren't any major objections.

In regards to the Javelin, as I've pointed out with how the Sidewinder compares to the Meteor, DPS isn't everything. The Sidewinder has less DPS than the Meteor, but is so much better in so many other ways that in practice it is a more effective weapon. Looking at the changes to missiles overall, I think that the way that the Javelin can be seen as more effective is its ability to overwhelm AM in a way that no other missiles can do now given that they've practically all had their rates of fire decreased while AM turrets were untouched. There's also the longevity that I mentioned the Javelin having over the Sidewinder and Meteor, being able to stay in a fight for longer and at the end of the day dish out more damage before running out of ammo than those two launchers. While the Javelin use to be able to deal more than twice as much damage before running out of ammo when compared to the Meteor, that difference is now only ~17.5%, which is at least still greater. If the suggested changes to the Meteor were put into place, that would become ~28% and the Javelin would be able to out-DPS the Meteor (and Sidewinder).

In so far as my suggestion to reduce the damage of the Meteor (and Sidewinder) instead of increasing the damage of the Javelin, I'm trying to avoid having the Meteor, Javelin, and Sidewinder be able to out-DPS the three P-cannons (Plasma, Proton, and Particle). I like @Pointedstick's analogy of missiles being comparable to the guns of larger ships, but from a gameplay standpoint, I think that these smaller missiles that can be easily fit on smaller ships should be under the power of the larger cannons. Reason being that if missiles directly compete with cannons, then from a pure DPS standpoint, you're practically giving smaller ships the ability to dish out just as much damage as larger ships, when those larger ships should be the ones dealing the most damage. 

(Percentages next to weapons are the buffs to missiles when compared to before #3866.)

Weapon | Combined DPS
---|---
Javelin (68%) | 372
Proton | 346.5
Particle | 318
Meteor (20%) | 315
Sidewinder (20%) | 305
Javelin (33%) | 296
Plasma | 293.33
Meteor (10%) | 288.75
Sidewinder (10%) | 280
Meteor (0%) | 262.5
Sidewinder (0%) | 255
Javelin (0%) | 222

Pointedstick:
> I like @Pointedstick's analogy of missiles being comparable to the guns of larger ships, but from a gameplay standpoint, I think that these smaller missiles that can be easily fit on smaller ships should be under the power of the larger cannons. Reason being that if missiles directly compete with cannons, then from a pure DPS standpoint, you're practically giving smaller ships the ability to dish out just as much damage as larger ships, when those larger ships should be the ones dealing the most damage.

That's actually the whole point of small ships though. If they didn't punch significantly above their weight, there would be no real reason for them to exist, which is exactly the problem small ships suffered from prior to improving missiles and giving them to small ships by default.

I'm really not sure that reducing Meteor and Sidewinder missile launcher capacities will make a difference for anything, but I won't block this on it.

Nescio0:
> radar jammers: large used to be exactly 7× as expensive as small, so perhaps change it to 140000 and 20000.

> if the meteors capacity is reduced to 30, then the sidewinder's should probably become 40.

> javelins become effectively useless, because the smaller meteor is more powerful per shot and per second (and that's disregarding range and homing).

Amazinite:
>radar jammers: large used to be exactly 7× as expensive as small, so perhaps change it to 140000 and 20000.

Not necessary. 

>if the meteors capacity is reduced to 30, then the sidewinder's should probably become 40.

Justification?

>javelins become effectively useless, because the smaller meteor is more powerful per shot and per second (and that's disregarding range and homing).

Ignoring that one because I already pointed out numerous times how the Javelin still has use, because DPS is not the only thing that makes a weapon powerful.

Amazinite:
>That's actually the whole point of small ships though. If they didn't punch significantly above their weight, there would be no real reason for them to exist, which is exactly the problem small ships suffered from prior to improving missiles and giving them to small ships by default.

I agree with you from a realism standpoint, but making small ships too powerful, enough to match larger ships, does seem detrimental to gameplay.
It's been mentioned numerous times before, but I think we need to also keep in mind how smaller ships are cheaper per gun than larger ships. Due to fleet sizes that might not come in to play much, but during large battles such as during missions, having smaller ships be even more powerful than their same cost in larger ships than they already were seems to be pushing it. So from that perspective in a way, smaller ships have already been able to punch above their own weight.

Amazinite:
I just checked the southern carrier events to see how that affects missile usage, and it isn't much. 
Small South goes from 32.65% to 35.29%.
Large South goes from 54.69% to 56.06%
Small North goes from 45.95% to 47.37%
Large North goes from 54.72% to 56.36%

Amazinite:
Continued discussion can return to #4149. I'll be leaving that issue open until v0.9.9 is released and we've received and handled feedback from a wider audience.

Issue #1856, opened 2016-11-27
Allow randomized dialogs/descriptions in mission definitions


Pointedstick:
To address some of the concerns in https://github.com/endless-sky/endless-sky/pull/1854 without writing a zillion mostly-boilerplate variants of the same missions, it would be great if I could do something like this:

```
on accept
	randomDialog
		"You pick up the package in an unmarked and uninhabited warehouse."
		"You pick up the package at a shipping facility near the spaceport."
		"You pick up the package at a dingy wooden shack several miles outside of the spaceport."
		"You pick up the package off the charred corpse of a pirate whom you best in a duel."
[...]

on complete
	payment 10000
	randomDialog
		"You hand the package to the first old beggar woman you see."
		"You hand the package to a street urchin who runs up and practically grabs it out of your hands."
		"You hand the package to an elderly gentleman who gives you a toothless grin."
		"You hand the package to the shopkeeper in a dingy restaurant."
		"You leave the package in a locker at the spaceport."
		"You drop the package into a trash can at the spaceport bar."
```

With this, you wouldn't need to duplicate the same mission dozens of times with only minor variations in one or two strings used in dialogs.

This would be super powerful and allow a lot more flavor and uniqueness in missions... just like hails and ship names.

kozbot:
Thoughts after looking through the code:

Dialog text (dialogText) is stored as a single std::string in MissionAction.

Would have to add another place to hold strings or maybe convert dialogText to a std::vector.
If changed to a vector rewrite some of the code to check first if size == 1 (treat like existing dialog) or size > 1 (treat as randomDialog) in the instantiate method.

If another place you could make it so it copies one of the dialog texts to dialogText and not have to modify the save functions since it's saving an already instantiated mission.  Just load into the action and then in instantiate check if dialogText.empty() then if randomDialogText.size()>0 pick one from there to copy.

I would be willing to make a PR if I could get an indication of the preferred method.

Pointedstick:
I was looking at it myself and thinking that it might be as easy as something like this:

```
		if(child.Token(0) == "randomDialog")
		{
			if(!child.HasChildren())
			{
				// print "it doesn't have any children, fool!
			}
			else
			{
				// randomly select a child and get its string representation
				dialogText = //the thing we randomly selected
			}	
		}
```

That would serialize the random string it chose into a static thing that gets written out into the mission description in your save file,. which is what you want.

[...]

PR #4164, opened 2019-02-01
Allow phrase-based randomized dialog

tehhowch:
Refs #1856 

This PR introduces reserved syntax for the combination of the `dialog` tag and the `phrase` tag when defining text for dialog pop-ups shown in missions' `on ____` actions and in NPCs' "completion".

Includes some examples in the existing datafiles, and otherwise generally extracts repeated text to a single location (so that it can be more easily diversified by future pulls).

Preferred syntax:
```
dialog phrase <stock phrase name>
```
a "stock phrase" is any phrase defined as a root level `DataNode`--i.e. the first thing on a new line:
```
phrase "hostile disabled"
    word
        ...
```
(see `hails.txt` and `names.txt` for lots more examples of root-level `Phrase`s and complex phrase construction.

Also supported: nested phrase specification
```
dialog
    phrase
        word
            ...
        word
            ...
```

**Any other usage of `phrase` with `dialog` is unsupported, including mixing regular directly-specified text and phrase-generated lines.** If you wish to do that, do it all in a custom phrase. You can define phrase parts as stock, in order to use them as such:
```
phrase "post-noun thanking"
    word
        ` `
        ` graciously `
        ` profusely `
    word
        `thanks you for your time.`

mission "do the thing with Jim-Bob"
    on complete
        dialog
            phrase
                word
                    `Jim Bob`
                phrase
                    `post-noun thanking`
```

Affected existing data: Any existing missions where the keyword `dialog` is followed by the word `phrase` possibly quoted (but not part of a larger token like "phrase you've heard before"), or where the `dialog` keyword is the end of its current line and the only token on the following indented child line is the possibly quoted word `phrase`, will be interpreted differently.

### DIFF
--- a/data/coalition jobs.txt
+++ b/data/coalition jobs.txt
@@ -8,6 +8,10 @@
 # WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
 # PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
+phrase "coalition ringworld workers"
+	word
+		`You drop off the workers on the <planet>. A few minutes later you receive a transfer of <payment>.`
+
 mission "Coalition: Ringworld Workers 1"
 	job
 	repeat
@@ -25,7 +29,7 @@ mission "Coalition: Ringworld Workers 1"
 		"coalition jobs" ++
 		payment
 		payment 10000
-		dialog `You drop off the volunteers on the <planet>. A few minutes later you receive a transfer of <payment>.`
+		dialog phrase "coalition ringworld workers"
 
 mission "Coalition: Ringworld Workers 2"
 	job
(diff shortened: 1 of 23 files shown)


jafdy:
Nice, now we just need a whole pile of new ways of saying the same thing.

Janaszar:
One thing this really needs is a way to "save" certain variables.
For instance, if you want to have random names for people you're dealing with, you want to have the same name on the start and end conversations;  "Adam wants you to deliver this to his friend Bob" -> "David thanks you for delivering the object from Charles." makes no sense.

Such a method for saving variables is also useful in other areas, such as eliminating "rep resets" ("We'll put you back to the standing you had before.") or temporarily removing all credits from the player ("You got robbed!")

tehhowch:
@janazar text substitution from condition variables is a different issue entirely.

(There is nothing stopping a content creator from saving the value of a reputation prior to changing it, either.)

Janaszar:
Huh.  Didn't know that about saving reputation.  Thanks.

tehhowch:
Yeah, there are some existing cross-file dialogs that are similar and could probably be sourced via phrase. I figured I would let them be and allow some other folks the chance to innovate :)

PR #4168, opened 2019-02-01
Increase crew attack and defense for Coalition species

Tadrix:
The Coalition is made up of species with increased affinity for close-quarters combat. They all move fast. Arachnids and beetles (animals the Arach and the Kimek are based upon) are carnivorous by design and have the benefit of their sturdy carapaces (natural armour) and natural weapons. Even Saryds would benefit from thick hides and strong kicks.

This proposed change reflects that.

Coalition civilians aren't taught to fight well, so they would have to rely on their natural instincts, There's also a possibility that a ship might be crewed by an all-Saryd crew (and they wouldn't be the ideal troops for fighting inside ships). Heliarch crews, however, are always mixed and are trained well to work in concert for their advantage.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -35,7 +35,7 @@ government "Bounty Hunter"
 government "Coalition"
 	swizzle 5
 	color 1 .6 .7
-	"crew attack" 2
+	"crew attack" .6
 	"crew defense" 3
 	language "Coalition"
 	"player reputation" 1
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -35,6 +35,8 @@ government "Bounty Hunter"
 government "Coalition"
 	swizzle 5
 	color 1 .6 .7
+	"crew attack" 2
+	"crew defense" 3
 	language "Coalition"
 	"player reputation" 1
 	"attitude toward"
@@ -112,6 +114,8 @@ government "Hai (Unfettered)"
 government "Heliarch"
 	swizzle 0
 	color 1 .8 .5
+	"crew attack" 3
+	"crew defense" 4
 	"player reputation" 1
 	"friendly hail" "friendly heliarch"
 	"hostile hail" "hostile heliarch"


Zitchas:
Sounds like a solid premise to me, although my first impression is that the values might be a tad high. I would be inclined to give the coalition 1.5/2 and the Heliarchs 2.5/3.

That being said, if this is accepted I would want to do more than just improve the Navy/Oathkeepers. Alphas should probably get a boost to their base stats too. Korath exiles, Remnant, Unfettered, and FW should probably get a much smaller boost. I'm thinking something along the lines of Navy/Oathkeepers to around 2, Korath Exiles to a similar 2, FW and Unfettered to 1.5, Remnant to 1.4.

Basically, all the primarily military organizations should have better H2H combat base stats than the average civilians, on the assumption that their crew probably have at least basic combat training and reasonably better standard gear.

Tadrix:
I consider a Heliarch worth two or three ordinary humans, considering what they are. They also shouldn't have many h2h outfits usable by the player (as only Saryds would use something even remotely fitting player's anatomy), something outfits cover for the other species.

I was planning to look at the Navy (giving them 1.2 and 2.2), the Remnant (2.2 defense), and the Korath (1.4 and 2.6, as they are warriors to the bone). The Alphas might warrant as high as 2.2 and 3.2, seeing how hard it was to deal with even one of them. But I wouldn't want to overdo it, and the majority of governments are fine as they are right now.

Amazinite:
I’d leave the Alphas alone, or at least not give them so large a buff. “Alpha” government ships likely aren’t fully staffed by Alphas, but actually mostly be pirates or perhaps slaves.

Nescio0:
Why do the Heliarch have different values? They're the same species and their ships already have larger crews.

Tadrix:
Because Heliarchs are armed to the teeth (their equipment just would't be usable by the player, hence higher values instead of outfits) and appropriately trained, while civilians don't have any access to weapons and aren't accustomed to combat (seeing as there's no piracy in the Coalition space).

Lineth:
> I’d leave the Alphas alone, or at least not give them so large a buff. “Alpha” government ships likely aren’t fully staffed by Alphas, but actually mostly be pirates or perhaps slaves.

I recall a discussion about the Alpha shit Giftbringer where it was determined that it had to be mostly human servants ass ~100+ alphas crammed in a flying tin can would tear each other to pieces, considering they hold each other in almost as much contempt as they do humans.

Tadrix:
I don't recall such discussion (link, please?), but I'd suppose that there are at least 50 Alphas on that carrier. However, here's your answer why numbers in the other PR are so low.

Amazinite:
Thinking on this a bit more, I think that instead of giving the Heliarch such a massive buff and explaining it away as them having high tech weapons, we actually give them weapons to use. So how about instead of +2/+2, they get say +0.7/+0.7. The Heliarch are the best of the best among Coalition society, so having higher base boosts than other trained groups like with #4170 seems sound. @Brick63 has already expressed interest in making such a weapon, discussed as being something that would be shoulder/back mounted as to be able to be used by all species of the Heliarch.

As for the boost to Coalition in general, I think the +1 to defense isn't congruent with the idea that trained, lightly armed (ref: laser rifle description) crew from #4170 are getting at most +0.6, when these are untrained, unarmed people that we're talking about. "Home advantage" doesn't account for such a high boost, as that's already accounted for by the fact that the base defense is twice the base attack. I also don't instincts is a good argument for such a massive boost, as any wounded animal that is cornered is dangerous, so the instincts argument would need to extend to all factions. At the very least, I think that the boost to the Coalition shouldn't be greater than the boost to trained crew, if they are to receive a boost at all.

Tadrix:
They get biological bonuses instead. Consider them being "fatter" and better armoured, basically having a racial on more hit points.

Meanwhile, I'm not aware of what the latest developments are, but as it was when you were writing your comment, I'm afraid I disagree with Brick's vision of how the Heliarch should be armed, so I'm not basing these numbers on what he might come up with. If the general consensus is in his favour, somebody will correct them anyway.

Amazinite:
I plan to expand the number of worlds that one can demand tribute from, the Coalition included, some time this update, so for a temporary measure it makes sense to give the Coalition and Heliarchs buffs higher than they otherwise would be if they had H2H weapons in game.

kittenpower800:
Well, I think the harliarchs should become exponencially more difficult the more individuals are on the ship because they coordinate with one another. If one is by themselves, vs two by themselves, vs ten in coordination. Fighting ten is going to be many times more difficult than fighting one at a time ten times over. So I think the square root of the ship's crew cap multiplied by 1.5 should be how much of a buff the ship gets. only 9? that's 3 times 1.5 for a 4.5% buff. with 100, that's 10, or times 1.5 that's 15% buff. Not only is there more crew, but each crew member is more difficult to fight. Or something of a scaling mechanic like this

PR #4170, opened 2019-02-01
Slightly improve h2h capabilities of trained crews

Tadrix:
Changes to attack and defence
Alpha: +0.4/+0.2
Korath: +0.4/+0.6
Navy Intelligence: +0.2/+0.2
Navy (Oathkeeper): +0.2/+0.2
Remnant: —/+0.2
Republic: +0.2/+0.2

Doesn't cover Coalition governments (there's a separate PR for them).

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -10,6 +10,8 @@
 
 government "Alpha"
 	swizzle 1
+	"crew attack" 1.4
+	"crew defense" 2.2
 	"player reputation" -1000
 	"bribe" 0
 
@@ -145,6 +147,8 @@ government "Indigenous Lifeform"
 government "Korath"
 	swizzle 0
 	color .8 .5 .1
+	"crew attack" 1.4
+	"crew defense" 2.6
 	language "Korath"
 	
 	"attitude toward"
@@ -218,6 +222,8 @@ government "Militia"
 
 government "Navy Intelligence"
 	swizzle 0
+	"crew attack" 1.2
+	"crew defense" 2.2
 	"attitude toward"
 		"Syndicate" -.1
 		"Pirate" -.3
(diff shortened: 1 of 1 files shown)


Lineth:
Shouldn't the Korath chances be lower than the Alphas? IIRC from an early first contact scenario now since dummied out it described the Korath as shorter than humans. Could be more muscular, but it made no references to that.

I can recognize it being equal or higher than the republic, with the implications of 'stocky lizard-dwarf warrior race' but they should not be stronger than outright super-soldiers. I'd put it +0.3/+0.3.

solardawning:
Questions: 
- How does this make the game more fun for the player?
- How is this communicated to the player to avoid confusion as to where the attack/defense values they see are derived from? 

That aside, I think you have an oversight in implementation: the base attack per crew is 0.5, but you're setting it to 1.x, thus more than doubling it in most cases. Though it's moot since you're not changing the player faction attack value, and the player is the only faction that can attack.

tehhowch:
@solardawning the default is 1 attack, 2 defense per crew:
https://github.com/endless-sky/endless-sky/blob/1fb48ecd1cf496885e7fd9e30b95e14ee0e328d9/source/Government.h#L128-L129

Tadrix:
> Shouldn't the Korath chances be lower than the Alphas?

Initially, Alphas had much higher numbers, lowered by the request of Amazinite (see the sister PR). This is just a tentative low number.

I'd give an all-Alpha crew +1.4/+1.2, but there's an opinion that many crew members are actually humans.

solardawning:
@tehhowch Thanks for the correction, for some reason I thought it was 0.5/1

Tadrix:
What tehhowch said. Also, the Kor Efret government doesn't get any bonuses, since all they have is a civilian population on what essentially are arkships. Defeating them is easier than Korath Exiles.

PR #4175, opened 2019-02-03
add newline to the end of coalition.txt

7even:
This adds a newline to the end of `data/coalition.txt` so that the parser of [my reference site for ES](https://github.com/7even/endless-ships) doesn't break; also it seems like a good practice anyway and is checked both by git and various programming language linters.


tehhowch:
Your parser should not be dependent on the presence--or lack thereof--of a final newline.

Pointedstick:
A harmless change, but @tehhowch is right: this is not and should not be mandatory, so you should fix your parser.

7even:
Sorry, I know that a proper parser should not fail because of a lack of newline but newlines are an important part of the data files grammar, and making a special rule for the last line of the file would really complicate things up; unfortunately I don't have the time to implement and maintain that.

Amazinite:
No harm.

7even:
Thanks.

PR #4177, opened 2019-02-05
Add a "battery too small to fire weapon" flight check warning

mattsoulanille:
New players may be confused if their projectile weapons don't fire even though their energy bar is full. This PR adds a flight check warning that warns the player if a ship has a non-continuously-firing weapon installed that requires more energy than the ship has energy storage.

Here's an example shuttle with no energy storage that has an energy blaster installed:
screen shot 2019-02-05 at 12 57 42 am

Here's the same shuttle with a beam laser, which produces no warning since it fires continuously:
screen shot 2019-02-05 at 12 57 20 am

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -494,5 +494,5 @@ tip "no hyperdrive?"
 tip "no fuel?"
 	`This ship has insufficient fuel capacity to make a hyperspace jump. If it took off, it would be unable to leave this star system.`
 
-tip "less battery than firing cost?"
+tip "insufficient energy to fire?"
 	`This ship has insufficient energy storage to fire an installed weapon.`
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -493,3 +493,6 @@ tip "no hyperdrive?"
 
 tip "no fuel?"
 	`This ship has insufficient fuel capacity to make a hyperspace jump. If it took off, it would be unable to leave this star system.`
+
+tip "less battery than firing cost?"
+	`This ship has insufficient energy storage to fire an installed weapon.`
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -760,7 +760,7 @@ double Ship::MaxFiringEnergy() const
 	for(const auto &it : outfits)
 	{
 		if (it.first->IsWeapon() && it.first->Reload() > 1) {
-			maxFiringEnergy = std::max(maxFiringEnergy, it.first->FiringEnergy());
+			maxFiringEnergy = max(maxFiringEnergy, it.first->FiringEnergy());
 		}
 	}
 	return maxFiringEnergy;
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -810,7 +810,7 @@ string Ship::FlightCheck() const
 	for(const auto &it : outfits)
 		if(it.first->IsWeapon() && it.first->FiringEnergy() > energy)
 			return "insufficient energy to fire?";
-
+	
 	return "";
 }
 
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -95,7 +95,7 @@ namespace {
 		double maxFiringEnergy = 0.0;
 		for(const auto &it : outfits)
 			{
-				if (it.first->IsWeapon() && it.first->Reload() > 1) {
+				if (it.first->IsWeapon()) {
 					maxFiringEnergy = max(maxFiringEnergy, it.first->FiringEnergy());
 				}
 			}
@@ -824,7 +824,7 @@ string Ship::FlightCheck() const
 		if(fuelCapacity < JumpFuel())
 			return "no fuel?";
 	}
-	if(battery < maxWeaponEnergy)
+	if((generation + battery + solar) < maxWeaponEnergy)
 		return "insufficient energy to fire?";
 
 	return "";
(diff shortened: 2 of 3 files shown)


source/Ship.cpp
```diff
@@ -752,6 +752,22 @@ int64_t Ship::ChassisCost() const
...
 
 
+// Get the maximum firing energy among all installed weapons that
+// don't fire continuously (like beams do).
+double Ship::MaxFiringEnergy() const
+{
+	double maxFiringEnergy = 0.0;
+	for(const auto &it : outfits)
+	{
+		if (it.first->IsWeapon() && it.first->Reload() > 1) {
```
tehhowch: This shouldn't care whether or not the energy drain is continuous or not - nothing is stopping a beam weapon from having a massive energy drain, after all.

source/Ship.cpp
```diff
@@ -752,6 +752,22 @@ int64_t Ship::ChassisCost() const
...
+			maxFiringEnergy = std::max(maxFiringEnergy, it.first->FiringEnergy());
```
tehhowch: We're `using namespace std` already.

data/tooltips.txt
```diff
@@ -493,3 +493,6 @@ tip "no hyperdrive?"
 
 tip "no fuel?"
 	`This ship has insufficient fuel capacity to make a hyperspace jump. If it took off, it would be unable to leave this star system.`
+
+tip "less battery than firing cost?"
```
tehhowch: "insufficient energy to fire?" is perhaps better phrased

source/Ship.cpp
mattsoulanille: Just tried a sunbeam on a firebird with only the smallest fuel cell installed, and I can confirm that it didn't fire.

I think a solution to this problem might be more complicated than just including continuous weapons in the max calculation. On the same firebird, a beam laser fires just fine even though it has a firing energy of 0.5 in weapons.txt, indicating that as long as a ship has enough `energy generation / second`, it will be able to fire a weapon. On the other hand, an energy blaster doesn't fire at all, even though its `firing energy / second` and `firing energy / shot` are well below the ship's `energy generation / second`. 

As far as I can tell, continuous weapons can fire as long as the ship has enough energy per second, but packeted weapons require a battery of size at least `firing energy / shot` to fire. I'll update the code in a few hours.

tehhowch: The key metric is per frame, not per second. All weapons consume their total firing energy each time they fire, and if that amount is available at the time of weapons checking, the weapon can fire--even if there is no energy storage and even if the weapon has a reload > 1. 

So if a ship has an energy generation of >= 5.2 per frame (>= 312 per second), it can possibly fire a Sunbeam (firing energy = 5.2 per shot) even without any energy storage. (If things like movement, shield repair, or other weapons have consumed energy this frame, then the ship may no longer have the needed energy in the current frame, in which case it would not be able to fire the Sunbeam.)

mattsoulanille: That makes sense. So the energy generation listed in the outfitter is `energy / second` but what really matters for firing weapons is `energy / frame`.  https://github.com/endless-sky/endless-sky/pull/4177/commits/a8a942f4a948fb85b07a8b86ddfb8c1ad3f30a9e takes this into account, checking all weapons (including those that have a reload of 1) against the total available `energy / frame` of the ship, which I think should be `solar collection + energy generation + battery`.

source/Ship.cpp
```diff
@@ -807,7 +824,9 @@ string Ship::FlightCheck() const
 		if(fuelCapacity < JumpFuel())
 			return "no fuel?";
 	}
-
```
tehhowch: Looks like your editor ate indentation

source/Ship.cpp
```diff
@@ -807,7 +824,9 @@ string Ship::FlightCheck() const
 		if(fuelCapacity < JumpFuel())
 			return "no fuel?";
 	}
-	
+	if((generation + battery + solar) < maxWeaponEnergy)
```
tehhowch: Note that the total energy available on a given frame is calculated already, as the local variable `energy`. (It comprises normal generation, storage capacity, solar collection, and fuel->energy conversion.)

So really you could distill the check to just:
```
...
for(const auto &it : outfits)
  if(it.first->FiringEnergy() > energy)
    return "insufficient energy to fire?";
```
(All `Outfit`s derive from the `Weapon` base class, so they all have [`FiringEnergy`](https://github.com/endless-sky/endless-sky/blob/6072a633dc87a0dfcf851825d3d3df6770e96785/source/Weapon.h#L180) available, with the default value of `0.0`.)
This isn't performance-critical code, so there isn't any penalty to keeping the expressiveness, i.e. `if(it.first->IsWeapon() && it.first->FiringEnergy() > energy)` (in case we ever create a non-`Weapon` use for `FiringEnergy()`).

mattsoulanille: My bad! I didn't know fuel->energy conversion was a thing. It just goes to show I should read more of the source code.

I've removed the MaxWeaponEnergy function in favor of your inline version, except I've kept the check for IsWeapon() for easier readability. If we ever need to know the max weapon energy elsewhere, we can refactor this.

source/Ship.cpp
mattsoulanille: The energy check is a warning condition, so I put it in the same paragraph as the other ones.

tehhowch: Yes, and the newline you added after it is missing indentation. Thus, GitHub shows you as also removing this indented line, instead of only adding the new lines. For a small pull it's easy to fix but as you edit more files you may want to find the IDE setting that governs this behavior and configure it for this repo.

Issue #4097, opened 2018-12-24
Art upgrade/sidegrade for (human) ships


8b8:
### tlwr: 

Created procedural cycles [texture](https://www.dropbox.com/s/706igoxckent3p5/_materials_human.blend?dl=0) for (human) ships + script to automatically [apply and render](https://www.dropbox.com/s/fxo59vwzkaa4rty/render.py?dl=0) most (human) [ships](https://www.dropbox.com/s/4nousr6ubb03cwb/render_all.py?dl=0). Please share your opinion on the [assets](https://www.dropbox.com/sh/a6k29u7wv3aea9j/AAC2_2FBFdI7Gt5u9Xn0hckSa?dl=0).

### Goals

The idea is to improve the visuals for ships in general, beginning with the human ships. They were presumably created first, and are a reminder how far the quality of the artwork has progressed. 

The two largest problems are imho, the hard edge, low-poly look they get from the lack of subdivisions as well as the lack of good materials on the 3d-models themselves.

### Work done so far

To improve this, I created a new procedural texture that adds a rivet/plating kind of surface detail without needing any new, real geometry. I also added a bevel and a subdivision surface modifier to make the edges "rounder". For the final touch I set up the compositing to add a little dirt to the crevices using the ambient occlusion pass.

On the bactrian this looks as follows (without manual postprocessing):
![bactrian](https://user-images.githubusercontent.com/27152901/50404343-079a5b00-07a7-11e9-8439-5da2737de7b9.png)
compared to the current in game model
![bactrian](https://user-images.githubusercontent.com/27152901/50404519-4af5c900-07a9-11e9-80e9-0b33289aecf5.png)

proposed 
![protector](https://user-images.githubusercontent.com/27152901/50405145-cdce5200-07b0-11e9-8177-7e5681afef56.png)

vs. in game
![protector](https://user-images.githubusercontent.com/27152901/50405146-d030ac00-07b0-11e9-9a34-ed7193e165e8.png)

These textures also look ok when rendered perspecitvely #3682
![berserkerpersp](https://user-images.githubusercontent.com/27152901/50404551-93ad8200-07a9-11e9-85c8-c29ffd5d3f08.png)

[...]

PR #4178, opened 2019-02-05
Adding marauder thumbnails

8b8:
Incorporating the changes from #4097 into a pull request to update the human ships art:

1. The images are rendered using cycles with a procedural texture instead of using the Krita post-processing
2. The base meshes have been rescaled and the edges have been softened with modifiers.
3. A light animation, like the one on the shuttle, has been added on all ships.
4. Compared to the current thumbnails the lighting direction has been changed to create shadows and make the thumbnail appear less "flat"

I created a repository containing python code to render/rerender the assets automatically https://github.com/8b8/endless-sky_render

I consider this an improvement mainly due to workflow reasons: 
Instead of having to manually rerender in blender, import into krita and export from krita, the images can be created automatically. This is especially usefull when creating animations.

Hint: When experimenting with an old savegame, the animation on the players (custom) ships will be way to slow. (glowing instead of blinking). All the .blend files can only be opened with blender2.8


Amazinite:
Not a fan of a lot of the running lights. You also seem to have quite a few ships with duplicate frames. For example, 3 of the 4 Barb frames look exactly the same. Instead of having duplicate frames to create a delay, you should be altering the attributes of the sprite, e.g. by increasing the delay.

Given that the addition of running lights is a separate issue from changing the graphics, I'd separate the two changes into different PRs.

Nescio0:
> All the .blend files can only be opened with blender2.8

Actually 2.79b is the latest stable.

Brick63:
I see no real need to do this, while some ships look a bit better, others look flat.

Amazinite:
You should really double check your work here. It looks like a good portion of the sprites have been screwed up in the process. It's most noticeable on the F-type Haulers (Nest, Roost, Skein), or on the Wasp, but in general it seems like most sprites have been altered in some detrimental way.

8b8:
Thanks for the feedback. 
Given the general direction, I think it might be best to work out which sprites are considered better and merge those. This will give me direction to work towards. (Pointing out the ones that are really bad also helps.)
 
@Amazinite I'm checking the difference after each rerender, but I seem to have become somewhat blinded... did I screw up any other ships on a geometry basis?

Amazinite:
Just look at the Roost, Skein, Nest, and Wasp as I mentioned. The Wasp has floating parts. The Roost, Skein, and Nest now have their fighter clamps floating in the air and at different locations than where they should be. The Star Barge has bigger fins now. And I haven't even looked at all the ships, just a few to get a quick sample of what's going on.

8b8:
I fixed the F-type Haulers (Nest, Roost, Skein) and the marauder mantas which got messed up. 

The Wasp and Star Barge files thumbnails seem to be based on newer blendfiles then the ones I get from the google drive here https://drive.google.com/drive/folders/0B9aK8dG39P29bmNQZlQxZ0pBZEU.

Is there a newer repository? Otherwise I'll try to update the blends to the best of my ability.

@tehhowch 
Is there a way to still get a better diff now, or does this only work with a clean PR?

tehhowch:
Now that filenames are equivalent for pre & post PR, GitHub's webpage (on desktop) shows the rich diff as appropriate. (Their mobile view doesn't render binary files.)

Amazinite:
I've reviewed all the ships alphabetically up to the Manta.

You swapped the Barb and Barb (Proton) sprites.
The Blackbird seems unacceptably dark now.
Almost all the sprites are bigger than they use to be, meaning that engine, gun, turret, drone, and fighter coordinates will be off on most ships. Smaller ships seem to have been effected worse, especially the Finch. The Firebird and Lance are the only ships that look like they haven't grown.
Not necessarily a bad thing, but this makes the colors on all ships very dull and washed out. But the Headhunter seems to have lost all its color due to this dulling.
Speaking of the Headhunter, it now has four guns due to #4014, but you replaced it with a sprite with only two guns.
Some ships, such as the Dreadnought or Gunboat, look far flatter with these changes.

Do you have 2x versions of all these sprites?

Amazinite:
On to the thumbnails.

The Aerie and Carrier appear to have lost their fighter and/or drone bays.
Starting with the Argosy, the same issue has popped up of ships becoming larger in size, although it doesn't seem to happen as often as with the ship images.
The change in sun direction is taking the color out of a lot of ships, as their upward and right side fins are now obscured in darkness, leaving only the less-seen left side fins in the light, which are typically obscured by the rest of the ship. See the difference in the Boxwing, Bactrian, Cruiser, or Freighter. This change in shadows on top of the details and dullness of the existing colors leads to the thumbnails in general just looking dull and depressing compared to now.
The Barb and Barbp thumbnails are also swapped.
The Bastion thumbnail actually appears to be physically flatter than before.
The Dreadnought looks like it's put on some weight; the major rear piece have become noticeably thicker.
Whatever happened to the Falcon is just... wrong.
All the Haulers, F-Types included, have gotten fatter.
The Heavy Shuttle has gotten fatter.
The Mule has gotten fatter.
The Rainmaker has lost its missiles.

8b8:
Thank you for your effort!

I fixed the barb and barbp in the blender files, the next rerender will have them

The blackbird with the unaltered material looked like that:
blackbird 0
This is the only model I manipulated the material color and specularity by hand to kind of get where the original artists might want it to be. I also think it is very black and can make it brighter.

The meshes in blender have been rescaled to have consistent scaling between all the ships (this wouldn't change the pngs though), additionally I adapted the scale of the camera to be exactly the size of the model +2 pixels buffer on each side. So the growth depends on the original buffer to the image border. I assumed, that the offset would be negligable and I didn't notice any wierd offsets during my playtests. I probably should recalculate all the engine, turret and gun positions to make it right.

For the coloring, I only have access the coloring of the models inside blender. Sprites that were colored in Krita will look very dull/different

I am currently rendering out high resolution images, they will upload as they finish.
 [high resolution perspective](https://www.dropbox.com/sh/b54kyztr9iajz56/AAC3wPyaZhj3ane7XPUbwqMoa?dl=0)
[double resolution top down](https://www.dropbox.com/sh/scxz4y3y2jfyoql/AADYewK4cgA80aoSnlWmEAk_a?dl=0)

I will try to work through your feedback tomorrow

Tadrix:
The ship renders I've seen from this PR were… hideous, to be honest. And while the shuttlecraft had neat running lights and I'd like to see them on all ships (but mind, for a civilian ship they were four, small, and blue; maybe military ships would incorporate a couple of red lights as well, but they'd be similarly small), what you're proposing turns the ships into gorram neon signboards. As if they needed to be better targets.

Worse, I don't see a need for this change, as current sprites look good.

That being said, you now started a discussion about this approach, so if you're really interested in it, stick around and see what happens.

Amazinite:
Whatever automated process you used to render the ships seems to have removed the beveling. (At least I believe that's the term for it. I'm not the guy who makes models.) It seemed most noticeably when looking at the Dreadnought, Firebird, or Falcon, but it's noticeably on all ships if you look close enough, which is itself a downgrade from what we have now. 

[Given history, I'm not a fan of automating stuff like this.](https://github.com/endless-sky/endless-sky/pull/3033#issuecomment-330630727) You gotta show the game some love and hand-make things, you know? Screwing up an automated process can make things take longer than if you hand-made it individually in the first place. :)

Pointedstick:
The beveled edges are still there, it's just that the lighting is so flat with this patch that they no longer pick up a specular highlight and stand out. Once the lighting regression is fixed, they should become more obvious.

Amazinite:
Whatever the case may be, there are a number of instances where the model has clearly been altered. E.g. the hallow areas on the Aerie or Carrier being removed, or whatever this is on the Falcon.
image

Given the noticeable changes like this, I wouldn't be surprised if there are a number of smaller errors across all the ships.

8b8:
Thank you for all the feedback, especially Amazinite for working through the differences.

I worked through most of the comments:
- Changed the lighting for the thumbnails to the wiki values (this increased the brightness in the thumbnails)
- Increased saturation
- Changed of the size of the sprites to match the originals more closely (max deviation ~ 1%)
- Fixed splinter + msplinter(e,s,w), mfirebirde, mbounderw, falcon +mfalcon(e,s,w), F-type haulers (clamp size) and headhunter 
- Reverted shuttle lights to lamps
- Updated Dreadnought, Aerie, Carrier, Rainmaker, Wasp and Star barge to match the current thumbnails
- Adapted  positions in ship.txt and marauder.txt to match new sprites
- As for ships getting fat, I hope the reverted lighting puts them on a diet.
- I added a bevel + subdiv modifier on all the models, ff edges seem sharper this should be a strong change in the normal maps overlapping with edges.

The current blends are [here](https://www.dropbox.com/s/iabhk9y6dbg9lyb/blends_18_2_8.tar.xz?dl=1)
(~20mb) use (blender2.8)[builder.blender.org] to open.

Here are the latest [sprites](https://www.dropbox.com/sh/c5whjgjvinrqduz/AABYqu40xdtIKXK1Xe-gpdc8a?dl=0) and the latest [thumbnails](https://www.dropbox.com/sh/c5whjgjvinrqduz/AABYqu40xdtIKXK1Xe-gpdc8a?dl=0). The thumbnails still need rescaling. These are wip for those interested in the progress, they will still change:

I'm note happy with the general saturation (red is very strong yellow and orange somehow weak) and the plating strength. The Bactrian, which was my initial motivation for this project,doesn't look very good atm. 
Also the buffer to the image border has become to small, I'll need to adjust this. Once this has happend I will update the PR.

I'd like to point out that the geometry on the models of the Dreadnought, Aerie, Carrier, Rainmaker, Wasp and Star barge are out of date on the google drive. I adapted to the current thumbnails to the best of my knowledge but it can still be off

Pointedstick:
Now that's what I'm talkin' about. Those new versions of yours are FANTASTIC:

aerie
berserker
carrier
frigate
marauder leviathan

8b8:
Updated sprites and thumbnails to the latest version, I tried to incorporate all suggestions. Compared to the lighting with the BI the sun light is a lot less hard (but can be adjusted). Also the background lighting is stronger reducing the hard strong shadows in the original sprites.

@Pointedstick  I really appreciate the enthusiasm, but the pictures kind of clutter up the comments :)

[blendfiles](https://www.dropbox.com/s/vymu7qtz50rffrx/blends_18_2_9.tar.xz?dl=1)

Nescio0:
Is it possible to generate sprites that are exactly the size of the existing ones, so the hardpoints do not have to be adjusted in the `data/*.txt` files?

Also, kestrel, lampyrid, mfury, mosprey, and pointedstick_vanguard?

Amazinite:
The resizing doesn't seem as big of a problem now, but I definitely agree with @Nescio0 that there shouldn't be any resizing going on at all.

Glancing through some of the ships now, I can't really say these changes are an improvement. In a good number of cases the change is barely noticeable, while in other cases like the Argosy or Combat Drone, the ship becomes very flat looking. Either that, or the current post-processing just looks better.
And there's definitely still something going on with models changing that has to be more than just the light catching parts differently.
<details><summary>Engines have become more blocky.</summary>

image

image

</details>
<details><summary>The tip is elongated.</summary>

image

</details>

jafdy:
I notice the Dreadnought's front most fins have also shrunk.

8b8:
@Nescio0 Yes I think I can remap the sprites to their original location. However I specifically changed the location to have the bounding box centered. Any rescaling is unintentional

@jafdy Yes... The size of the Dreadnought's front most fins is different between the thumbnails and the sprite. The only solution I see here, is to have different geometry for the sprite and the thumbnail

@Amazinite could you elaborate on the engines of the dreadnought? Or do you mean the "pointiness of the edges" The engines of the bactrian was an experiment that snuck into the commit. 

There generally two influences on the geometry: 
 1. I added a bevel and a subdiv modifier on all meshes, with the intention to remove all strong edges, as they make the models seem artificial. (Few real object have super sharp edges)
 2. The geometry inside blender has two independent scaling factors. The scale of the mesh itself and the object that contains the mesh. For the procedural textures to work well, I had to apply the scale on all the objects. In this operation, the scale gets transferred from the object space to the mesh. This doesn't change anything on the base geometry by itself. But the bevel modifiers that were used during modeling have,per default, an absolute offset depending on the scaling of the object. So when the scale is applied the value of the offset is changed and the geometry changes. An extreme example of this was the white hump on the back of the Falcon. 
The only way to avoid this, would have been to apply the modifier before the scale is applied. This however makes it hard to modify the object later on and increases the vertex count a lot.

Zitchas:
It is interesting to watch this project. I can't venture a guess as to whether or not this becomes a standard part of the workflow, but I could see an automated system like this that moves a lot of the post-processing into the render being quite useful for rapid prototyping of ship designs for creators (instead of the usual "here's a base ship - just imagine it with post-processing" stuff). It could also be useful for those who can make decent 3d objects but lack the skill for post-processing.

8b8:
Todo:

- [ ] The thumbnails are currently scaled down using a bad interpolation. 
- [ ] The Dreadnought fin should be scaled up and rotated (lower) to find a balance between sprite and thumbnail
- [ ] Ftypes don't have a material assigned to a mesh in the cockpit area.
- [ ] Rainmaker rockets need to be scaled
- [ ] scout engine rings


Generally I have to agree that most sprites (with 1x resolution) don't add enough to be added just for fun.

**Sprites** I think might merit a merge are:

- Base Leviathan (The Barrel of the gun is fixed)
- Bulk freighter (Color of the carrier is in line with the thumbnail)
- Rainmaker (Rockets are now hinted in the sprite, )
- Argosy (reduced "blockiness" of the base mesh)
- Frigate (more overall detail through the texture)
- f-types (more detail in the back center and soon in the cockpit area)

**Thumbnails** I think might merit a merge are:

- The marauder ones ( They look OK and the competition is the 2d sprite)
- barb (the current one looks really muddy)
- heavy shuttle
- Rainmaker
- frigate
- corvette
- quicksilver

For the highdpi ones there might be a few more candidates...



<details><summary>Click to view images for Comparison; new left, current right::</summary>
<p>

Sprites
comps0
comps1

Thumbnail

compt0

Sprites 2x

comps_d0
comps_d1
comps_d2
comps_d3
comps_d4

Thumbnail2x

compt_d0
compt_d1
compt_d2
compt_d3
</p>
</details>

8b8:
I won't have time in the near future to improve the stuff, so I reverted the pull request to the minimum. At this point it might be called "Adding marauder thumbnails" :)

For the future I will make other pull requests for fewer or single ships I'd like to see updated.

Somehow the ship.txt I download from master is different then the one I try to merge to...

Fzzr:
> I won't have time in the near future to improve the stuff, so I reverted the pull request to the minimum. At this point it might be called "Adding marauder thumbnails" :)

@8b8 You can change the PR name to make the new purpose more clear, both here and in the merge commit.

Pointedstick:
I really like the subtle "ship plating" texture and I think even doing just that without touching the meshes, lighting, or shadow sharpness would be a nice upgrade.

Amazinite:
Especially seeing as how touching the meshes is causing issues. :)

Wrzlprnft:
most of the earlier human ships are,due to scaling in object mode and different changes to bevels to get the same effect, a bit of a nightmare for a general script and basically every one of those needs/needed to be checked foe changes.

tehhowch:
Make sure you update your remotes before merging them:

```
git remote add upstream https://github.com/endless-sky/endless-sky
git fetch --all --prune
git merge upstream/master
```

I think the meshes and bevels were touched because the procedural texture has a scale associated with it, and if the ships are scaled differently then the texture of two different ships will have different scales, despite supposedly being the same type of construction materials. I could be wrong about that though.

8b8:
Given that Amazinite created marauder thumbnails the old fashioned way, I would like to close the PR.

Issue #2375, opened 2017-04-08
bug / mechanics: escorts departing from "wrong" planets


EndrosG:
As the title says, the game is currently unable to let escorts depart from different planets.

Scenario:
You once parked a ship on Mars.
Later, when you are on Earth, you unpark the ship.
When you launch from Earth, the other ship should launch from Mars, but it will be displayed as launching from Earth as well.

Applies to ships owned by the user and to escorts of all kind (?) from missions.

Found it while trying to work on question 3 of #2364. I wanted the Deep ships to launch from Mars and the Navy ship to launch from Luna, as the `on offer` conversation suggests.

Savegame for testing attached.
[Testing Tester~TEST FW Synd Cap 1 takeoff.txt](https://github.com/endless-sky/endless-sky/files/907428/Testing.Tester.TEST.FW.Synd.Cap.1.takeoff.txt)

Edit: Testing mission definition file attached too.
[testing_mission_fw_synd_cap_1.txt](https://github.com/endless-sky/endless-sky/files/907434/testing_mission_fw_synd_cap_1.txt)

Bladewood:
Relevant: You can access outfitters for a ship on a different planet, as long as the both of you are in the same system.

gunqqer:
Doesn't seem to unreasonable for the ships escorting you to do a quick planet hop to you

endless-sky:
I don't see much point in fixing this. Especially because if taken to its logical conclusion, you wouldn't be able to outfit those parked ships either, even though you can outfit ships that were holding position in system (and not disabled) when you landed. This is another case of game play trumping realism.

tehhowch:
I think addressing wrong-planet departure is addressable without impacting gameplay experience negatively. I'll look into that in the next week.

tehhowch:
A ship that had a planet when the player landed will now remember and use that planet when the player next takes off, thanks to #4111

PR #4179, opened 2019-02-05
Specify an initial planet for mission NPCs

tehhowch:
This PR would allow content creators to build off #4111 , and enable a mission's NPC to start out landed on a specific planet. If the specified planet is not in the instantiated NPC's system, then the planet specification is ignored.

e.g.
```
npc
    system "Sol"
    planet "Luna"
    government "Korath"
    fleet "Korath Raid"
```

will result in a Korath raid fleet taking off from Luna when the player next takes off, while
```
npc
    planet "Hevru Hai"
    government "Quarg"
    fleet "Large Quarg"
```
will only result in a large Quarg fleet departing from Hevru Hai if the mission is being offered from Hevru Hai (as the default system when unspecified is the mission's origin system).

The requirement to specify the proper system in addition to the planet ensures that the NPCs will be in the proper system even if the referenced planet is a wormhole (which has multiple systems). If we are ok with being unable to specify the appropriate wormhole planet properly, this requirement can be dropped and the PR amended to allow specifying just `planet` or both `system` and `planet`. 

Refs #2375 

----

test binary available from AppVeyor here: https://ci.appveyor.com/project/tehhowch/endless-sky/builds/22152377/artifacts
test missions: 
[depart-test-mission.txt](https://github.com/endless-sky/endless-sky/files/2834176/depart-test-mission.txt)

### DIFF
--- a/data/free worlds reconciliation.txt
+++ b/data/free worlds reconciliation.txt
@@ -676,11 +676,13 @@ mission "FW Syndicate Capture 1"
 		government "Free Worlds"
 	
 	npc accompany save
+		planet "Luna"
 		personality escort heroic waiting
 		government "Navy (Oathkeeper)"
 		ship "Cruiser (Mark II)" "N.S. Hand of Justice"
 	
 	npc accompany save
+		planet "Mars"
 		personality escort heroic waiting
 		government "Deep Security"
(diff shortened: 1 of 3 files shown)


Zitchas:
I can see this being useful.

PR #4180, opened 2019-02-06
Added shipyard thumbnails for Pug ships

Amazinite:
Any ship that can be captured is capable of being viewed in the shipyard, and thus all ships that can be captured should have a shipyard thumbnail. Here are the thumbnails for the current Pug ships. I have plans to get shipyard thumbnails for the Marauder ships as well, and perhaps also the Lampyrid seeing as how it can be captured.

@Kryes-Omega rendered the ships for me to post process, and he also put finishing touches on the Arfecta to make it look better.

Assets: [pug thumbnails.zip](https://github.com/endless-sky/endless-sky/files/2887347/pug.thumbnails.zip)


Amazinite:
So I realized I did the Zibruka differently than the other two ships and fixed that, but I'm wondering if the way I did the Zibruka was better. Check that commit for the difference.
Zibruka way: coloring layer is placed below the texture layers.
Other way: coloring layer is placed above the texture layers.

Nescio0:
A welcome addition!
For consistency and completeness, I think every ship should have a thumbnail. Archon, Skylark, Wardragon, and World-Ships B and C don't have them either.

Amazinite:
I think I'm gonna touch these up slightly before merging. They have the same textured layers as the ship sprites and a color layer for the red/orange/yellow, but I didn't do any extra shadowing or the like. I think the cockpits on the Maboro and Enfolta could use a touch up to make it more obvious that the cockpits are segmented, something that is very obvious on the ship sprites, but on the thumbnails the black of the various cockpit segments just blends together.

Amazinite:
Reduced the saturation on the Enfolta, made the segments of the Enfolta and Maboro more noticeable, and added extra shadows to the ships like other thumbnails have.

PR #4183, opened 2019-02-09
Prevent remnant technology missions from blocking each other when done

Fzzr:
The removed lines meant that only one of the technology missions would ever offer, blocking progression.

### DIFF
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -664,9 +664,7 @@ mission "Remnant: Heavy Laser"
 	to offer
 		has "Remnant: Technology Available: offered"
 		not "Remnant: Catalytic Ramscoop: active"
-		not "Remnant: Catalytic Ramscoop: done"
 		not "Remnant: Plasma Cannons: active"
-		not "Remnant: Plasma Cannons: done"
 		random < 50
 	on offer
 		require "Heavy Laser"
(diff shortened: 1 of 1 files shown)


Fzzr:
@Zitchas please clarify here. It doesn't look like this is intentional since `"Remnant: Salvage 1"` requires all three to be complete.

Fzzr:
Attached file exhibits the issue without this change and doesn't show the issue with this patch applied:
[J.D._Wolfe.txt](https://github.com/endless-sky/endless-sky/files/2848385/J.D._Wolfe.txt)

Zitchas:
The intent was that the player would only be able to start a single one of the tech retrieval missions, and the Salvage 1 mission is supposed to require the player to have learned Remnant sign and *one of* the three tech retrieval missions.

```	to offer
		has "Remnant: Learn Sign Follow Up: offered"
		random < 50
		or
			has "Remnant: Catalytic Ramscoop: done"
			has "Remnant: Plasma Cannons: done"
			has "Remnant: Heavy Laser: done"```

Based on the wiki's info, that or should mean that only one of those three needs to be done in order to progress.

That being said, I have fully intended to modify the dialogue in these three missions so that they do not block each other, however in order for that to happen there needs to be some branching dialogue so being re-introduced to Tali is handled as a subsequent meeting instead of a first meeting when this is the case. And the missions need to be differentiated a bit more. They *are* awfully similar.

Fzzr:
I see now. I think the issue must be due to missing a different condition. Closing, will re-open if signs point to this as an issue again.

Zitchas:
Thanks for keeping an eye on things!

PR #4190, opened 2019-02-17
Add an article "the" when selling a ship at a shipyard

DrBlight:
I missed this one when I was adding "the" to the beginning of all the ship names.

### DIFF
--- a/source/ShipyardPanel.cpp
+++ b/source/ShipyardPanel.cpp
@@ -243,7 +243,7 @@ void ShipyardPanel::Sell(bool toCargo)
 	
 	int count = playerShips.size();
 	int initialCount = count;
-	string message = "Sell ";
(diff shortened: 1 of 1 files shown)


Fzzr:
Nice, good catch. Since you're contributing, are you on the Discord? If not, it's at discord.gg slash ZeuASSx. We're always happy to have more contributors around to work with directly.

DrBlight:
I appreciate it, but for me this is a hobby that I only want to do at my own (slow) pace. I'm happy that people like some of my simple contributions enough to merge them!

Fzzr:
The discord isn't just for development, it's mostly a general discussion and help group for the game. It's just that people who contribute content also hang out there to get feedback. You're welcome anytime, thanks for contributing!

PR #4191, opened 2019-02-18
New Mereti and Sestor asteroid

beccabunny:
This is an upload of #4189: it contains uploaded images of sestor and mereti asteroids (that should look more damaged) as well as the other files


beccabunny:
What do you think about it?

Ferociousfeind:
Seems interesting to me, though I would like more varied bits floating around. And that would require a lot of work, just for this little area of space.

beccabunny:
I was thinking at something similar, but when i saw that just making these two debris took 86 more images i preferred to do just that

jafdy:
There will be a way to script it so you could have all the images produced automatically, I don't know how to do it but you could try asking around and see if anyone will help you.

beccabunny:
If you mean normal mapping, I don't think it can help in that case; the PR about the human ships upgrade may help about the details part, but I have absolutely no idea about how to use the python file in blender

Wrzlprnft:
One can render out an animation in blender in seperate frames rather easily (dabbled with it with the debris ships) without the need of a script.

beccabunny:
Can you give me more hints about that? For these I rotated the models manually

Wrzlprnft:
first, you make an animation by rotating it and setting the keyframes for rotation (shortcut i) for every quarter of the full 360° (depending on your frames per second (default 24) you'll need to do a bit of calculation)

since you want a animation loop, and blender defaults to smoothing it down, you either do the rotation twice and take the middle loop, or you go into an f-curve editor and remove the smooth to no rotation from the animation interpolation curves (sounds difficult, but it isn't)

now that you have a looping animation, you set start and endframe to match that, and then set the render output in the output panel below as still images and not a video
https://docs.blender.org/manual/en/latest/render/output/output.html#output-panel
you can already choose a fitting naming convention, as well as make them go into a subfolder.

for further questions i suggest using discord or so. The blender manual also can help a great deal. i do not have access to a pc with blender until at best tomorrow evening or tuesday, though, so i can't just make a template for you myself. there should be the debris ships assets somewhere though, you might need ask derpy/amazinite.

beccabunny:
Thanks, this really helps.
Since low people commented, what do you think of the PR?

Wrzlprnft:
I, too, think that a bit more variety would be good. But i like the base idea as it is based on flavor/story and makes sense.

data/map.txt
```diff
@@ -15447,6 +15450,8 @@ system Mesuket
 	link Eneremprukt
 	link Salipastart
 	asteroids "large metal" 1 1.6758
+	minables meretidebris 1 1.98864
+	minables sestordebris 1 2.11643
```
Amazinite: Should Mesuket not have the most debris since that's where the majority of the fighting occurs? It would certainly make for a good risk-reward scenario to have lots of debris in Mesuket.

The prevalence of debris so far seems to be tied to the prevalence of other minables in the system, when it should be tied to how much fighting goes on in that system.

data/harvesting.txt
```diff
@@ -8,6 +8,28 @@
 # WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
 # PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
+minable "meretidebris"
```
Amazinite: These names should be "mereti debris" or "sestor debris" with a space, not smashed together.

beccabunny: Thanks, I added a "_" between the names

data/harvesting.txt
```diff
@@ -8,6 +8,28 @@
 # WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
 # PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
+minable "mereti_debris"
```
Amazinite: Remove the underscore. The minable names in the maps can simply be in quotes, e.g. `minables "sestor debris" 2 7.17656`

data/map.txt
beccabunny: I increased them to 5 and 4, but i think that making them even higher might be strange since they would be almost the only tipe of asteroids in the system

Amazinite: >they would be almost the only tipe of asteroids in the system

That would be the point, would it not?

beccabunny: What values do you suggest?

data/harvesting.txt
beccabunny: Done

data/map.txt
tehhowch: If based on conflict, make sure the values are updated via GameEvent when the Sestor vacate their planets. It would be fine to have a gradual phaseout (i.e a few events over several in-game months / weeks).

beccabunny: I'm not sure about that, since the lore is based on the fact that the conflict has lasted for hundreds of years; probably even a hundred of years wouldn't be enough to remove all that debris from the space

tehhowch: If the payload is balanced to be rewarding while harvested under fire, once the player is not under fire the balance is gone, yes?

beccabunny: The asteroids are supposed to be balanced when the player is not under fire, so should I increase the reward to justify the risk pre-sestor shutdown?

beccabunny:
closing this to wait for when i have a time to make a more polished PR with the stuff corrected, i will also take this chance to delete my fork and do a new one

Azure3141:
Sources tell me this should totally be merged

TheGiraffe3:
Sources tell me the fork was deleted

PR #4192, opened 2019-02-19
Added sympathetic choices in Wanderer start (@10010101001)

Amazinite:
Ref: #4093

Seeing as how @10010101001 seems unresponsive with fixing up #4093, I've made the changes myself.
The second new choice is the same, just placed at the end of the list of choices instead of the beginning. The first new choice now has the player express their understanding, but still voice concern to Rek, meaning that the following dialog from Rek can remain the same.


10010101001:
Thanks, I've pretty much given up on my pr

Pointedstick:
+1 ship it!

Issue #3174, opened 2017-10-29
Should carried fighters & drones help recharge their carrier?


tehhowch:
Currently, only the parent helps the carried ship, and if the carried ship is fully repaired, any shield or hull repair ability it has is not used, even if the parent could benefit from it.

As-is, this innate repair functionality of the carried ship is cost-free - the carried ship gets to use it no matter how much energy it has and how much energy is needed to perform the repair. (This may warrant fixing). (edit: this cost is added to the parent in #3182 )

If the shared repairs are made into a two-way street, then the costs of the carried ship's repair functionality will need to be applied to the parent, lest players load their repair outfits into fighters and then simply never deploy fighters.

Amazinite:
This does sound like it could be an interesting mechanic to have and use, but it does beg a question of why not all functions of a ship (energy, cooling, and the mentioned hull and shield recharge) should be shared between carrier and carried. If a docked ship is fully charged, should the excess energy not go to the carrier, with the fighter and carrier sharing the heat generated? And for shield regeneration or hull repair, the generator which the fighter has equipped will likely already be sufficient to sustain these outfits, essentially giving the carried ship free repairs at maybe only the cost of heat, unless of course for some reason the fighter is able to provide repairs to the carrier but unable to provide energy as well.

To me, it makes sense that a carried ship would be able to acquire repairs from its mothership when docked but the fighter would be unable to assist the carrier. To have the carried ship provide shield regen and hull repair to the carrier also seems to me as if it would be a huge boost to the player (it is essentially increased outfit space for carrier ships) but a generally unused mechanic by the AI.

TL;DR, cool idea, I'd like it, but I question the balance and rationality of it.

Nechochwen-D:
The only thing I can see doing is allowing power generation to be shared.

ghoulavenger:
Considering that fighters have extremely limited outfit space, I don't see how the parent ship would benefit all that much from outfits installed on fighters. Even if you were to remove the weapon mounts from fighters that only adds a slight benefit for repair outfits or shield generators. The amount of energy produced and energy stored will not be significant enough to bother with on a carrier either. So while I understand the Amazinites objection, unless we see fighters with significantly more outfit space, I don't see a real problem. Conversely, I don't see a really large benefit either.

[...]

PR #4194, opened 2019-02-21
Implement heat transfer between motherships and drones/fighters

TurkeyMcMac:
I just implemented this because I saw the `TODO` comment, but I don't know if it is appropriate to implement. I haven't done very much testing with this feature, but drone carriers still appear to work fine.

The rate of heat transfer is proportional to the difference in heat per unit area between the mother ship and the drone and to the area of the drone (i.e. the area across which heat can flow.) There is also the proportionality constant `HEAT_TRANSFER_COEFFICIENT`, which is at the moment 0.04. The higher the constant, the faster the heat transfer rate.

I think that this feature adds an interesting mechanic to the game. If your drone carrier has very good cooling but your drones do not, then the carrier could act as a heat sink to keep the drones cool.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1544,11 +1544,13 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 void Ship::DoGeneration()
 {
 	// Transfer heat to or from ships in bays before they can do their own generation.
-	if (!bays.empty()) {
+	if (!bays.empty())
+	{
 		const double area = Width() * Height();
 		const double heatDensity = heat / area;
 		for(Bay &bay : bays)
-			if(bay.ship) {
+			if(bay.ship)
+			{
 				const double bayArea = bay.ship->Width() * bay.ship->Height();
 				const double bayHeatDensity = bay.ship->heat / bayArea;
 				// The amount of heat transferred is proportional to the difference in heat per unit
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -50,7 +50,7 @@ namespace {
 	const vector<Angle> BAY_ANGLE = {Angle(0.), Angle(-90.), Angle(90.), Angle(180.)};
 	
 	const double MAXIMUM_TEMPERATURE = 100.;
-	// The portion of the different in heat per unit area which is transferred per game tick:
+	// The portion of the different in heat per unit mass which is transferred per game tick:
 	const double HEAT_TRANSFER_COEFFICIENT = 0.04;
 	
 	const double SCAN_TIME = 60.;
@@ -1554,18 +1554,16 @@ void Ship::DoGeneration()
 	// Transfer heat to or from ships in bays before they can do their own generation.
 	if (!bays.empty())
 	{
-		const double area = Width() * Height();
-		const double heatDensity = heat / area;
+		const double heatDensity = heat / Mass();
 		for(Bay &bay : bays)
 			if(bay.ship)
 			{
-				const double bayArea = bay.ship->Width() * bay.ship->Height();
-				const double bayHeatDensity = bay.ship->heat / bayArea;
+				const double bayHeatDensity = bay.ship->heat / bay.ship->Mass();
 				// The amount of heat transferred is proportional to the difference in heat per unit
-				// area (the 2D heat density) and to the area through which the heat is transferred,
-				// which is the area of the ship in this bay.
+				// mass and to the mass through which the heat conducts, which is the mass of the ship
+				// in this bay.
 				const double transfer =
-					(heatDensity - bayHeatDensity) * HEAT_TRANSFER_COEFFICIENT * bayArea;
+					(heatDensity - bayHeatDensity) * HEAT_TRANSFER_COEFFICIENT * bay.ship->Mass();
 				heat -= transfer;
 				bay.ship->heat += transfer;
 				// The ships in bays must also be simulated alone.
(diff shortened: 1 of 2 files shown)


Ferociousfeind:
Question, does this actually take the fighter's/drone's sprite's area into account, or its mass? I ask because I myself can't quite discern what's happening in the added lines. Not exactly a code genie, you know?
Or, more general question, what does it take into account in all? Heat of mothership, heat of fighter, mass of mothership and fighter, etcetera, anything else?

TurkeyMcMac:
It takes into account the areas (width * height) and heat of each ship. Now that I think about it, mass might be a better measure of how big a ship is, although I doubt the effect will be great. It would be fairly easy to switch around in any case.

TurkeyMcMac:
Upon further thought, I'm not sure what effect this will have on gameplay. It will be beneficial to carriers with ample cooling, but it might be a detriment to ones with high idle heat. Hopeful someone more experienced with the game can have better insight. I wish I could have considered this before making the pull request, but it didn't occur to me.

Ferociousfeind:
In general, heat is a bit of a confusing topic to quantify. For example, what _would_ be the best way to determine how fast heat moves from one to another? How would one determine what "equilibrium" would be? And do the edge cases with these answers look nice, or unintuitive?

TurkeyMcMac:
I'm no expert in this area, but I have based my code on a few observations:
 1. In general, high pressure flows into low pressure. Therefore, the rate of heat transfer should depend on the difference between temperature per volume between the ships.
 2. Intuitively, a higher area of contact should cause a higher rate of transfer. If the surface area doubles, the transfer rate should double.
 3. Ship mass is a decent approximation of both volume and contact surface area.

That's why I used this formula:
```
carrierDensity = carrierHeat /  carrierMass
fighterDensity = fighterHeat / fighterMass
transferred = (carrierDensity - fighterDensity) * fighterMass
```

These creates a smooth transfer of heat between carriers and their fighters.

petervdmeer:
> Upon further thought, I'm not sure what effect this will have on gameplay.
Fighters currently share "positive" things with their carrier while docked. For example fighter shields get recharged when the carrier has fully charged shields and charge capacity left. The same for hull repair, the carrier repairs its own hull first and then repairs hulls of any docked fighters.

I would expect the same for heat and cooling. If a carrier has cooling capacity left, then I would expect the carrier to allocate some cooling capacity to cooling the docked fighter. But I don't expect heat problems to propagate between the two. Transferring heat would mean that the fighter or the carrier would act as some sort of heatsink for the other one and I think that such a mechanism would introduce more complexity and non-transparency than just keeping the heat in the ship that acquired it.

LocalGod79:
to be honest, i'd like to see fighters and carriers have a more symbiotic relationship - such that a fighter with excessive cooling/energy generation/heat dissipation (perhaps only on external carry bays?) could help their mothership in the way the mothership currently offers repair services to shield/hull of the drone/fighter.

carriers are typically the ship with a cooling problem within this relationship - small ships like fighters usually have such high heat dissipation characteristics that they seldom retain heat (that would potentially be dissipated by a mothership) like a larger ship will.

as it stands, carried ships already offer a mass boost to their mothership (which gives it higher max heat - if a mothership will overheat when it launches its carried ships, it won't launch them).

Ferociousfeind:
I'd like most a "oh, that's cool" moment, where a hot fighter docked with its mothership will make the mothership's heat shoot up, or vice versa, a cold fighter would act as a quick and temporary cooling mechanism.

However, perhaps what's here already is well enough on its own?

LocalGod79:
tbh, the thing that bothers me most is that a fighter with an albeit tiny fuel cell/battery (and shield generator) that's not contributing energy generation/capacity to their mothership while docked.

in fact, the main thing that can be said about the fighter/drone bay is that it basically adds a gun port or two (that can be destroyed separately) to the ship, and a liability in credit resources (as a result).

the added mass/heat capacity of a loaded carrier, in practice is of little value, since it usually carries a higher price in maneuverability without adding any other offensive capability, e.g. enough added power to operate guns that might overheat it if it were not carrying fighters (that were also helping to power them).

all of that said, fighters can be an impressive force multiplier in combat, cheaply absorbing more than their share of missile damage early in a battle, or adding savage missile damage until their ammo is expended...

Amazinite:
I'm going to decide against implementing this.

The purpose of fighters should be to deploy them, not hoard them, which this mechanics change would incentivize. Reason being is that it basically grants any carrier [free outfit space](https://github.com/endless-sky/endless-sky/issues/3174#issuecomment-340284067) in the form of being able to use its fighters as extra cooling outfits.

A change like this would mean that any carrier would be able to free up outfit space by having a dedicated fighter always docked for cooling purposes, as a fighter that can cool any amount of heat transferred to it by its carrier can just continually have heat transferred to it, resulting in no need for cooling on the carrier outside of whatever it needs to get past the spaceflight checks.

Through testing, a stock Korath Raider with two stock Chasers docked was able to lose >10% of its idle heat, dropping from a typical 50% idle heat to below 40%. With fighters dedicated to cooling, one could push that even lower. A Tek Far 109 on the other hand, which has 16 fighters and drones, dropped from 70% idle heat all the way down to 2% idle heat because of all the fighters it was able to push heat onto, which had sufficient cooling to just keep taking heat. I suspect the only reason it stopped at 2% heat is because that's how much heat the Tek Far 109 was producing at the end of each frame, which wasn't pushed onto the fighters until the beginning of the next frame.

PR #4199, opened 2019-02-25
Deep mission fixes

Amazinite:
Some of this was spotted by @oo13 

* Fixed the `to offer` of `Deep: Remnant 0` to refer to the proper mission name of `Deep: TMBR 1` and not 2.
* Used the new `declined` condition to make sure that the recruit escort/reinforcement missions during `Deep: Scientist Rescue 1` did not accept if the player declined `Deep: Scientist Rescue 1`.
* Simplified conditions and missions in order to ensure that missions offer or complete in the proper order and fail at the right times.
* Fixed an instance of a mission not being invisible when it should be.

### DIFF
--- a/data/deep missions.txt
+++ b/data/deep missions.txt
@@ -1971,6 +1971,8 @@ mission "Deep: Scientist Rescue 1"
 	description "Escort the <npc> to <destination> to retrieve the Star Queen. Escorts will be waiting for you on Memory in the Zosma system and Farpoint in the Alnitak system."
 	destination "Haven"
 	clearance
+	stopover "Memory"
+	stopover "Farpoint"
 	to offer
 		has "Deep: Scientist Rescue 0: done"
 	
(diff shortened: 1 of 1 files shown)


data/deep missions.txt
```diff
@@ -2257,14 +2260,21 @@ mission "Deep: Scientist Rescue 3A"
 			`	"Thank you, <first>." Lieutenant Paris starts walking over to the Star Queen, leaving you alone next to your ship.`
 				accept
 	
+	npc kill
+		government "Pirate"
+		personality staying nemesis
+		system "Arneb"
+		ship "Bactrian (Cosmic Devil)" "Cosmic Devil"
+		dialog 
+			`After the Bactrian explodes, you receive a transmission from an unknown source. It suddenly opens on its own andou receive a transmission from an unknown source. It suddenly opens on its own and an electronically distorted voice begins speaking to you.`
```
toilethinges: While we're here, I don't think "andou" is a word.

.."and you"?

PR #4204, opened 2019-03-01
feat: Overhaul jamming and tracking to make them more fun and realistic

### LINKED ISSUE #1865: Jamming balance: should missiles go haywire when jammed?
I've been playing about looking at jamming, and it bothers me that in a typical AI fight, no matter how high I set radar jamming the vast majority of Sidewinders hit. Realistic jamming numbers (4-20) make about 4-15% miss. Missile boat-on-missile boat is a bit better as they both keep distance, but they're a slim minority of encounters.

If the principle is intended to be something like **Projectile.cpp** says:

`	// Jamming of 1 is enough to increase your chance of dodging to 50%.`

then, really, in AI testing with 25 missiles shot I got:

```
                  AI vs AI misses
radar jamming     normal  missile boat
1                 0/25    1/25
20                4/25    11/25
2,000,000         7/25    18/25
```
I love the idea behind the implementation of jamming types, but I'd never actually use a jammer on a player ship or introduce a Tier 2/3 jammer in new content. Would you?

I feel the problem is that a jammed missile is still probably on target and with turn = 0 it's just spending a moment as a rocket. It _doesn't_ get a false lock or go haywire (**Projectile.cpp**, line 185-186) . And since at least half of an AI encounter involves them close and head-on, jammers really struggle. It's not much different for players.

My suggestion is to let jammed missiles temporarily go haywire. In code, I'd probably implement a `jammed `state, with `hasLock `lost when `jammed `and a block to do this:

```
	else if(target && homing && jammed)
		turn = [SOME RANDOM FUNCTION];
```

Obviously I'd need to test this carefully, and I wouldn't want to rush into anything that made jammers too good. But I think a jammed missile briefly obtaining a false target is more in line with what electronic countermeasures try to do, and could help make jammers more competitive.

If I'm on to something interesting that @endless-sky might approve of, I'm really happy to try to implement it myself, run any and all test cases suggested, and do a PR.

### LINKED ISSUE #3912: Radar jamming is essentially worthless

Pointedstick:
This PR aims to make missile tracking and jamming much more strongly affected by the characteristics of the target ship and the missile's tracking and homing methods, rather than being something that almost always works no matter what. This is more fun, rewards the player for equipping their ship appropriately, and makes missiles a bit more likely to miss their targets on average.

To that effect, it implements the following technical changes to jamming and tracking:
- Double the "resolution" of missile tracking by checking for lock twice as frequently
- When a missile has lost its lock, zero out its base chance to regain it
- Attenuate the effects of radar jamming proportional to the distance between the jammer and the missile; more jamming means that radar-guided missiles can become jammed at longer ranges
- Make jammed radar-guided missiles have a chance to "go haywire", proportional again to the strength of the jamming compared to the strength of the missile's radar tracking modified by the distance to the jammer
- Increase the reliability of infrared tracking when close to the target, mimicking the real-world phenomenon stemming from the wavelengths of infrared light being easier to distinguish with greater distance
- Slightly reduce the ship weight at which optical tracking becomes maximally effective to compensate for tracking being a bit less accurate now

Taken together, these changes mean that running a cool ship actually causes IR missiles to miss a lot of the time, and installing even a Small Radar Jammer is enough to make a fraction of Sidewinders miss against a maneuvering target. Torpedoes miss a little bit more against large ships, and are now almost useless against small maneuvering ships.

This PR also makes the combination of radar and IR tracking in a single missile make sense: radar tracking is used for long-range effectiveness, and the missile switches to IR when it's close to the target. Some of the best real-world missiles do this, too.

closes #1865
closes #3912

### DIFF
--- a/source/Projectile.cpp
+++ b/source/Projectile.cpp
@@ -139,7 +139,7 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 	double turn = weapon->Turn();
 	double accel = weapon->Acceleration();
 	int homing = weapon->Homing();
-	if(target && homing && !Random::Int(60))
+	if(target && homing && !Random::Int(30))
 		CheckLock(*target);
 	if(target && homing && hasLock)
 	{
@@ -150,7 +150,6 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 		double trueVelocity = drag ? accel / drag : velocity.Length();
 		double stepsToReach = d.Length() / trueVelocity;
 		bool isFacingAway = d.Dot(angle.Unit()) < 0.;
-		
 		// At the highest homing level, compensate for target motion.
 		if(homing >= 4)
 		{
@@ -198,6 +197,46 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 					accel = 0.;
 			}
 		}
+	}
+	// Homing weapons that have lost their lock have a chance to get confused
+	// and turn in a random direction ("go haywire"). Each tracking method has
+	// a different haywire condition. Weapons with multiple tracking methods
+	// only go haywire if all of the tracking methods have gotten confused.
+	else if(target && homing)
+	{
+		bool infraredConfused = true;
+		bool opticalConfused = true;
+		bool radarConfused = true;
+
+		// Infrared: proportional to tracking quality
+		if (weapon->InfraredTracking())
+			infraredConfused = Random::Real() > weapon->InfraredTracking();
+
+		// Optical: proportional tracking quality
+		if (weapon->OpticalTracking())
+			opticalConfused = Random::Real() > weapon->OpticalTracking();
+
+		// Radar: If the target has no jamming, then proportional to tracking
+		// quality. If the target does have jamming, then it's proportional to
+		// tracking quality, the strength of target's jamming, and the distance
+		// to the target (jamming power attenuates with distance).
+		if (weapon->RadarTracking())
+		{
+			if (target->Attributes().Get("radar jamming") == 0)
+				radarConfused = Random::Real() > weapon->RadarTracking();
+			else
+			{
+				// Closeness to target; 0 when just fired, 1 when impacting
+				double closenessTotarget = 0;
+				if (position.Distance(target->Position()) == 0.0)
+					closenessTotarget = 1;
+				else
+					closenessTotarget= 1 + weapon->RadarTracking() - (1 / (weapon->Range() / position.Distance(target->Position())));
+				radarConfused = Random::Real() < closenessTotarget * ( 1 - (weapon->RadarTracking() / target->Attributes().Get("radar jamming")));
+			}
+		}
+		if (infraredConfused && opticalConfused && radarConfused)
+			turn = Random::Real() - min(.5, turn);
 	}
 	// If a weapon is homing but has no target, do not turn it.
 	else if(homing)
@@ -286,15 +325,17 @@ shared_ptr<Ship> Projectile::TargetPtr() const
 }
 
 
-
+// TODO: add more conditions in the future. For example maybe proximity to stars
+// and their brightness could could cause IR missiles to lose their locks more
+// often, and dense asteroid fields could do the same for radar and optically
+// guided missiles.
 void Projectile::CheckLock(const Ship &target)
 {
-	double base = hasLock ? 1. : .5;
+	double base = hasLock ? 1. : .15;
 	hasLock = false;
 	
-	// For each tracking type, calculate the probability that a lock will be
-	// lost in a given five-second period. Then, since this check is done every
-	// second, test against the fifth root of that probability.
+	// For each tracking type, calculate the probability twice every second that a
+	// lock will be lost
 	if(weapon->Tracking())
 		hasLock |= Check(weapon->Tracking(), base);
 	
@@ -302,22 +343,42 @@ void Projectile::CheckLock(const Ship &target)
 	if(weapon->OpticalTracking())
 	{
 		double weight = target.Mass() * target.Mass();
-		double probability = weapon->OpticalTracking() * weight / (200000. + weight);
+		double probability = weapon->OpticalTracking() * weight / (150000. + weight);
 		hasLock |= Check(probability, base);
 	}
 	
-	// Infrared tracking is 10% when heat is zero and 100% when heat is full.
+	// Infrared tracking is 5% when heat is zero and 100% when heat is full.
+	// When the missile is at under 1/3 of its maximum range, tracking is
+	// linearly increased by up to a factor of 3, representing the fact that the
+	// wavelengths of IR radiation are easier to distinguish at closer distances
 	if(weapon->InfraredTracking())
 	{
-		double probability = weapon->InfraredTracking() * min(1., target.Heat() + .1);
+		double distance = position.Distance(target.Position());
+		double shortRange = weapon->Range() * 0.33;
+		double multiplier = 1.0;
+		if(distance <= shortRange)
+			multiplier = 2 - distance / shortRange;
+		double probability = weapon->InfraredTracking() * min(1., target.Heat() * multiplier + .05);
 		hasLock |= Check(probability, base);
 	}
 	
 	// Radar tracking depends on whether the target ship has jamming capabilities.
-	// Jamming of 1 is enough to increase your chance of dodging to 50%.
+	// The jamming effect attenuates with range, and that range is affected by
+	// the power of the jamming. Jamming of 2 will cause a sidewinder fired at
+	// least 1200 units away from a maneuvering target to miss about 25% of the
+	// time. Jamming of 10 will increase that to about 60%.
 	if(weapon->RadarTracking())
 	{
-		double probability = weapon->RadarTracking() / (1. + target.Attributes().Get("radar jamming"));
+		double baseRadarJamming = target.IsDisabled() ? 0 : target.Attributes().Get("radar jamming");
+		double currentRadarJamming = baseRadarJamming;
+		if(baseRadarJamming > 0.0)
+		{
+			double distance = position.Distance(target.Position());
+			double jammingRange = 500 + (sqrt(baseRadarJamming) * 500);
+			double rangeFraction = min(1.0, distance / jammingRange);
+			currentRadarJamming = (1 - rangeFraction) * baseRadarJamming;
+		}
+		double probability = weapon->RadarTracking() / (1. + currentRadarJamming);
 		hasLock |= Check(probability, base);
 	}
 }
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -258,7 +258,7 @@ tip "piercing protection:"
 	`Protection against weapons that pierce shields. If you add more than one outfit with this attribute, each additional one is less effective: a total value of 1 results in a 1 percent reduction in damage leak, while a total value of 11 only results in a 10 percent reduction.`
 
 tip "radar jamming:"
-	`Reduces the ability of missiles that use radar to track your ship.`
+	`Reduces the ability of radar-guided missiles to track your ship. The more radar jamming is installed, the greater the distance at which missiles can be jammed, and the greater the chance that a jammed missile will fly off in a random direction.`
 
 tip "ramscoop:"
 	`Replenishes fuel by harvesting the stellar wind. Ramscoops are more effective when closer to a star. If you add more than one ramscoop, each additional one is less effective: you need four ramscoops to achieve double the recharge rate that one ramscoop provides.`
@@ -564,13 +564,13 @@ tip "tracking:"
 	`How well this weapon maintains its target lock.`
 
 tip "optical tracking:"
-	`How well this weapon maintains an optical target lock, which depend on the size of the target.`
+	`How well this weapon maintains an optical target lock, which depends on the size of the target.`
 
 tip "infrared tracking:"
-	`How well this weapon maintains an infrared target lock, which depends on the target ship's heat level.`
+	`How well this weapon maintains an infrared target lock, which depends on the target ship's heat level and distance from the missile; IR-guided missiles become more accurate the closer they are to their targets.`
 
 tip "radar tracking:"
-	`How well this weapon maintains a radar target lock, which can be interrupted by a ship with radar jamming installed.`
+	`How well this weapon maintains a radar target lock; negatively affected by the target ship's radar jamming equipment.`
 
 tip "piercing:"
 	`This percentage of this weapon's damage always leaks through to the hull.`


data/weapons.txt
```diff
@@ -652,20 +652,20 @@ outfit "Meteor Missile Launcher"
 		"hit effect" "missile hit"
 		"inaccuracy" 5
 		"velocity" 11
-		"lifetime" 490
+		"lifetime" 240
```
Nescio0: Why?

data/weapons.txt
```diff
@@ -652,20 +652,20 @@ outfit "Meteor Missile Launcher"
...
-		"infrared tracking" .8
+		"turn" 1.8
+		"homing" 1
+		"infrared tracking" .65
 		"shield damage" 260
 		"hull damage" 160
 		"hit force" 220
 		"missile strength" 6
-	description "The Meteor Missile is the cheapest and simplest homing weapon available in human space. Meteors are quickly destroyed by anti-missile systems, and their infrared tracking systems have trouble homing in on targets with low heat levels. Nevertheless, their large warheads can spell doom for poorly-defended targets when the missiles are fired in great numbers."
+	description "The Meteor Missile is the cheapest and simplest homing weapon available in human space. Like all infrared-guided missiles, its seeker only reliably tracks targets with high heat levels--the closer the better. Though the relatively short-ranged Meteor is quickly destroyed by anti-missile systems, its large warhead can spell doom for poorly-defended targets when fired in great numbers or from an agile launch platform at close range."
```
Nescio0: Replace "--" with " - ".

Nescio0:
Your proposal halves Meteor lifetime and increases Torpedo and Typhoon velocity, thus getting the following ranges:
* Javelin: 600
* Rocket: 1000
* Meteor: 2640 (instead of 5390)
* Typhoon: 3600 (instead of 3000)
* Torpedo: 4200 (instead of 3600)
* Nuke: 4800
* Sidewinder: 6580

data/weapons.txt
Amazinite: >But the biggest changes are to the Meteor: it is now a short-range bombardment weapon that is optimally fired from fighters against medium-to-large-sized ships.

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
...
 	{
-		double probability = weapon->RadarTracking() / (1. + target.Attributes().Get("radar jamming"));
+		double baseRadarJamming = target.IsDisabled() ? 0 : target.Attributes().Get("radar jamming");
+		double currentRadarJamming = baseRadarJamming;
+		if(baseRadarJamming > 0.0)
+		{
+			double distance = position.Distance(target.Position());
+			double jammingRange = 1000 + (baseRadarJamming * 100);
+			double rangeFraction = min(1.0, distance / jammingRange);
+			currentRadarJamming = (1 - rangeFraction) * baseRadarJamming;
```
Amazinite: So radar tracking has "infinite" range that simply gets more effective the closer the missile is to the target? 
Should it perhaps instead have a range similar to scanners, that is the sqrt(scanning strength) * 100, so outside of that range there is no effect, with a weak effect just inside the range that still gets stronger the closer to the target?

source/Projectile.cpp
```diff
@@ -199,6 +198,15 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 			}
 		}
 	}
+	// If the weapon is homing, is radar-guided, has lost its lock, and the
+	// target has jamming, give a chance for the missile to turn in a random
+	// direction, in proportion to the strength of the jamming and the missile's
+	// tracking quality
+	else if(target && homing && weapon->RadarTracking() && target->Attributes().Get("radar jamming") > 0.0)
+	{
+		if(Random::Real() < (1 - (weapon->RadarTracking() * 2 / target->Attributes().Get("radar jamming"))) )
```
Amazinite: I'm guessing that `Random::Real()` is a number between 0 and 1.

Does this mean that a small radar jammer (2 jamming) has a 10% chance to cause a Sidewinder to go haywire while a large radar jammer (10 jamming) has an 82% chance?

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
 
 void Projectile::CheckLock(const Ship &target)
 {
-	double base = hasLock ? 1. : .5;
+	double base = hasLock ? 1. : .0;
```
Amazinite: So a missile that loses its lock will never have a chance to regain it?

source/Projectile.cpp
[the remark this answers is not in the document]
Pointedstick: I tried that, but it actually made jamming too powerful!

Amazinite:
I think a chart would be helpful that shows how many times a missile's lock will be checked over its lifetime before and after.

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
...
+	// Jamming of 10 will increase that to about 60%.
 	if(weapon->RadarTracking())
 	{
-		double probability = weapon->RadarTracking() / (1. + target.Attributes().Get("radar jamming"));
+		double baseRadarJamming = target.IsDisabled() ? 0 : target.Attributes().Get("radar jamming");
+		double currentRadarJamming = baseRadarJamming;
+		if(baseRadarJamming > 0.0)
+		{
+			double distance = position.Distance(target.Position());
+			double jammingRange = 1000 + (baseRadarJamming * 100);
```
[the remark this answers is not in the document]
Pointedstick: Oh, that's a good idea.

source/Projectile.cpp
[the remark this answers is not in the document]
Amazinite: Yeah, I was thinking that exponential might be more realistic, but too powerful. Was just a thought, though.

source/Projectile.cpp
Pointedstick: Yes. Note that this condition is only evaluated once the missile has already lost its lock. Maybe I should also make this only happen if the missile is within the jamming radius?

source/Projectile.cpp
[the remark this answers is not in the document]
Amazinite: Just fiddling around with some graphs, I think that `500 + 500 * sqrt(radarJamming)` fits pretty well compared to the current range scaling. From about 0 to 2 jamming, it's worse than the current equation. Then from about 2 to about 13 it's better, then beyond 13 it's worse, but the sqrt means that stacking outfits scales better.

source/Projectile.cpp
Amazinite: `Maybe I should also make this only happen if the missile is within the jamming radius?`

Can a missile lose its lock outside of the jamming radius?

Pointedstick:
> I think a chart would be helpful that shows how many times a missile's lock will be checked over its lifetime before and after.

That's simple enough, so I don't think there's any need for a chart: checking for lock simply now happens twice as often.

source/Projectile.cpp
Pointedstick: > `Maybe I should also make this only happen if the missile is within the jamming radius?`
> 
> Can a [radar-guided] missile lose its lock outside of the jamming radius?

Not from jamming. It can still lose its lock due to imperfect radar tracking, though.

source/Projectile.cpp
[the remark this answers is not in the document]
Pointedstick: I like that. Definitely better with the `sqrt()`.

Zitchas:
Sounds like a good change. I like the idea of having the different guidance systems being more differentiated (and range having more of an impact). The way that plays into the jammers and heat and what not 

I don't know if Nescio's numbers are right, but if they are the Sidewinder's huge range seems anomalous. I'd think that torpedos should probably have the longest range. I always picture torpedos as the long-range weapons typically used by capital warships (and the occasional destroyer rigged for anti-capital warfare) at extreme ranges before they get into missile range.

As far as heat goes, do heat-tracking things use the total heat present, or the "what percentage of max is the ship at" value? I can see reasons for either one being used.

Nescio0:
> I don't know if Nescio's numbers are right, but if they are the Sidewinder's huge range seems anomalous. 

Feel free to check yourself, `range = velocity × lifetime`; range is also displayed in game.
Hai Tracker range is 8400, by the way.

Amazinite:
You should probably update some of the tooltips to reflect these changes. E.g. add `The jamming range is proportional to the square root of this value.` to the jamming stat like scanners have.

Pointedstick:
I'm finding myself in agreement that that Sidewinder's very long range is a bit odd now (it kind of already is, really). I'm up for reducing that, but also starting to feel like the weapon changes should be in another patch, and this one should just be about jamming alone. Thoughts?

Zitchas:
I agree. There's been a lot of changes surrounding missiles lately, and I honestly haven't really got a feel for how they are now. Focusing on the jamming and guidance for now, and deal with mechanical adjustments to things like range later is probably easier to understand for everyone.

Pointedstick:
Moved the weapon characteristic changes to https://github.com/endless-sky/endless-sky/pull/4212

Amazinite:
These changes effect more than just human missiles, but #4212 only makes tweaks to human missiles. Have you tested these changes with alien missiles as well?

Pointedstick:
Yes. Most of the alien missiles use multiple tracking methods and so are in general are much less affected by jamming and cool-running targets than human missiles were, so the changes are smaller there. But they too will nonetheless miss a bit more often with this change. I didn't think any of them really needed any tweaks to adjust to this.

data/outfits.txt
```diff
@@ -137,7 +137,7 @@ outfit "Large Radar Jammer"
 	"mass" 14
 	"outfit space" -14
 	"radar jamming" 10
-	description "This high-power radar jammer will make it much more difficult for missiles that rely on radar to track you. It will not affect missiles that track you optically or that look for your ship's heat signature."
+	description "This high-powered jammer will make it much more difficult for radar-guided missiles to track you. It is even powerful enough to occasionally cause them to fly off in random directions. Like its smaller cousin, this jammer's effectiveness diminishes as the distance to the missile increases."
```
Amazinite: I don't think it's necessary for the jammer descriptions to basically just repeat what the `radar jamming` tooltip says.

data/weapons.txt
```diff
@@ -838,7 +838,7 @@ outfit "Torpedo Launcher"
 		"hull damage" 680
 		"hit force" 300
 		"missile strength" 30
-	description "Torpedoes are the most powerful homing weapon, but they move slowly enough that fast ships can simply outrun them, making them most useful against large targets or against small ships at times when they are too distracted to dodge. Torpedo launchers are also much larger than other missile systems, so they are most often found in capital ships."
+	description "Torpedo Launchers are the largest and most powerful homing weapons, often found on capital ships for fleet engagements. Fast ships can often simply outrun or dodge the Torpedoes they fire, making them generally effective only against large slow targets. However, very skillful pilots have been known to score hits on smaller ships, especially at close range."
```
Amazinite: I don't think that any of these weapon description changes are really necessary. They're largely just rephrasing what was already said.

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
 
 void Projectile::CheckLock(const Ship &target)
 {
-	double base = hasLock ? 1. : .5;
+	double base = hasLock ? 1. : .15;
```
Amazinite: With all the other changes made, I'm wondering if we should just leave this line alone. Missiles are already checking if they've broken lock twice as often, so reducing the change to regain a lock after that, especially by so much, seems overkill.

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
...
 	
-	// For each tracking type, calculate the probability that a lock will be
-	// lost in a given five-second period. Then, since this check is done every
-	// second, test against the fifth root of that probability.
+	// For each tracking type, calculate the probability twice every second that a
+	// lock will be lost
 	if(weapon->Tracking())
 		hasLock |= Check(weapon->Tracking(), base);
 	
 	// Optical tracking is about 15% for interceptors and 75% for medium warships.
```
Amazinite: Would these numbers not change because of the change in the probability equation?

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
...
+	// linearly increased by up to a factor of 3, representing the fact that the
+	// wavelengths of IR radiation are easier to distinguish at closer distances
 	if(weapon->InfraredTracking())
 	{
-		double probability = weapon->InfraredTracking() * min(1., target.Heat() + .1);
+		double distance = position.Distance(target.Position());
+		double shortRange = weapon->Range() * 0.33;
+		double multiplier = 1.0;
+		if(distance <= shortRange)
+			multiplier = (1 - (distance / shortRange)) * 2;
```
Amazinite: Once `distance == shortRange`, then this multiplier becomes 0, right? That doesn't seem like it should be the case; with this, outside of `shortRange` the multiplier is 1, then from `shortRange` to `shortRange`/2, the multiplier is between 0 and 1. Then between 0 and `shortRange`/2, the multiplier is between 1 and 2.

data/outfits.txt
Pointedstick: It doesn't just restate the tooltip; it now mentions the haywire behavior.

source/Projectile.cpp
Pointedstick: In my testing it was necessary. This was an integral part of making missiles not so accurate all the time.

source/Projectile.cpp
Pointedstick: I think these were always sort of a guesstimate. :man_shrugging:

Aetreus: These numbers are kind of right- the check is against probability^0.2, so if the check runs 5 times the chance of the missile passing all of them and tracking for 5 seconds straight(on average) is (probability^0.2)^5, which reduces to just the probability. Checking twice as often would make it 2.5 seconds for 5 checks on average.

Pointedstick:
Any interest in this?

Zitchas:
Yes. I really like the idea of differentiating the tracking systems more. I also like the idea of having infrared and radar being better at different targets.

Pointedstick:
Me too. :) I've merged master and would like to re-open it for consideration in 0.9.11.

Zitchas:
We currently have three tracking methods, right? Infrared, radar, and optical. 


Could we add a fourth, gravimetric (primarily for Pug, Quarg, and Drak, but possibly other high tier guided munitions. It'd be kind of like the proposed infrared, except targeting mass (basically, the smaller the target the poorer it fares. Almost never hits things light warship sized or smaller, but virtually impossible for it to lose tracking on big things like heavy warships). Anyway, not something for this PR, but an addition I'd like to keep in mind as changes are made to the current things.

Also, I still hope that we'll get flares someday. EV, EVO, and EVN all had them. They were the first secondary "weapon" I ever purchased in every pilot. They were very effective at giving players a shot at escaping from pirate missile boats. (so long as they could outrun the ships proper)

Pointedstick:
Man, I remember those Gravimetric missiles from EVO.

Optical tracking is already partially mass-seeking. Adding a Gravimetric tracking mode would seem to imply making Optical targeting purely visual, no?

And yes, I would like flares as well. Something that's interesting to note is that in modern naval missile combat, nearly all missiles that have been successfully defeated in battle were neutralized using flares, chaff, or jamming [1]. The incidence of CIWS (the modern version of Endless Sky's anti-missile turrets) physically destroying incoming missiles is actually very rare. Now presumably things have improved over the last thousand years, but still, the same maxim would seem to hold true: it's safer and more reliable to confuse the missile's primitive seeker than it is to try to blow it up while it's on a high-speed collision course with you. That's basically your last ditch defense, not the primary means!

[1] Reference: https://apps.dtic.mil/dtic/tr/fulltext/u2/b192139.pdf

jafdy:
Gravimetric could be differentiated by ignoring cloaking, it is mentioned that the cloak stops detection by everything other than gravimetric sensors. It would need to be high tier.

Zitchas:
I'm still hoping that we get multiple states of cloaking implemented at some point. It makes perfect sense that gravimetric would ignore certain types of cloaking but not others. 

(Again with the EVN references. Some cloaks were basically Star Trek "this just makes us invisible and that's it" cloaks, whereas others were actually "We fold reality around us so we're not there anymore" cloaks. The former act as cloaks we have now. The later act like the cloak we used to have (the one that rendered the ship untouchable)

Anyway, two ideas for optical tracking: 
1) It is based on the surface area of the sprite - bigger ships are easier to track. This often (but not always) correlates to the mass of the ship.
2) It is basically a "generic" tracking that follows whatever it is locked onto without much variation - except that it has the possibility of switching targets to other ships and asteroids that cross its path.

In terms of vulnerabilities:
- IR is vulnerable to ships cooling off and use of flares (very high hot points). Tends to be most effective at short ranges.
- Radar is vulnerable to jamming and interference. (EVN had interference as a general system trait. It affected both radar based stuff and would cause the map to flicker with static. Some systems the map was basically useless) Good for long ranges.
- Optical is vulnerable to breaks in the line of sight: Targets going behind cover or cloaking. Medium range.
- Gravimetric is vulnerable to targets hiding near objects of greater mass (perhaps allowing for hiding "behind" planets, stars, and black holes. It could be interesting to make those background features actually combat-significant). Most effective at long ranges, except near astronomical phenomenon. 

As far as jamming goes, what about the possibility of having indiscriminate jammers? Conventional jammers try to target the frequencies used by incoming ordinance while avoiding the frequencies used by "friendly" weapons. Which is ideal for combat craft. But I could see merchants and others who aren't really interested in sticking around and fighting opting to use broad-spectrum interference that sacrifices precision and combat-capability for sheer power output. Such an output would be similar sized, but moderatly more powerful than conventional jammers - but they'd affect *all* missiles of that guidance type, rather than just the ones that are hostile to them and/or their allies.

geojak:
Any progress on this?

MCOfficer:
I know you probably just want to kick up dust, but lemme give you the obvious answer: No, the last comment was 6 months ago.

Maybe i'm just grumpy because of other things, but know that for the last 2 days, everyone watching this repo has gotten hourly e-mail notifications about you asking what the status of X is.
I don't mind reviving old issues. But attempting to revive several dozen issues at once is pretty pointless, isn't it?

geojak:
Sorry, I just wanted to help find and close outdated issues and PRs. I did not intend to annoy you or others. Anyway good luck.

Amazinite:
If you're looking for outdated issues, then you should look for stuff that's actually been fixed, then when commenting on the issue, paste the commit hash that fixed the issue. I don't think a single issue you've posted on is actually outdated from being fixed; they're all just old.

Pointedstick:
The PR still merges cleanly, so technically there's nothing preventing people from reviewing it and suggesting changes.

Zitchas:
The proposed changes *sound* good to me, and the code looks fine to my limited ability to understand it. Perhaps it could be updated to be current with master - if only to trigger the error checking and artifact builds for people to use?

I definitely like the idea of heat (or lack thereof) being more relevant in closequarters and against certain missiles, while radar is better for long range. It is also very nice to make dodging when one has jamming to actually be useful, too.

I'm all for merging this.

Pointedstick:
Boom, rebased and all tests passed.

If folks want to merge this, I'll come out of hiding to tweak it in follow-up PRs as needed. :)

Zitchas:
@petervdmeer I know you've got more code experience than many of us. Care to take a look through this and see what you think?

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
...
 	if(weapon->Tracking())
 		hasLock |= Check(weapon->Tracking(), base);
 	
 	// Optical tracking is about 15% for interceptors and 75% for medium warships.
 	if(weapon->OpticalTracking())
 	{
 		double weight = target.Mass() * target.Mass();
-		double probability = weapon->OpticalTracking() * weight / (200000. + weight);
+		double probability = weapon->OpticalTracking() * weight / (150000. + weight);
 		hasLock |= Check(probability, base);
```
petervdmeer, line 311: I have the impression that weapons with more than 1 tracking type (for example radar-track and optical track) actually have a higher chance to lose a lock than weapons with only a single tracking type, because the (single) lock can be lost for each tracking type that is used.

I would normally expect missiles to use the best tracking type for each part of their trajectory and become more resilient against losing target locks if they track over more than 1 type.

This looks like something that already existed in the existing code (before this PR), so I don't consider this a blocker for this PR. But maybe it would be nice for a follow-up PR to perform the hasLock-Check one with the mminimum p



This looks like something that already existed in the existing code, so this PR doesn't make it worse.

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
...
+	// Infrared tracking is 5% when heat is zero and 100% when heat is full.
+	// When the missile is at under 1/3 of its maximum range, tracking is
+	// linearly increased by up to a factor of 3, representing the fact that the
+	// wavelengths of IR radiation are easier to distinguish at closer distances
 	if(weapon->InfraredTracking())
 	{
-		double probability = weapon->InfraredTracking() * min(1., target.Heat() + .1);
+		double distance = position.Distance(target.Position());
+		double shortRange = weapon->Range() * 0.33;
+		double multiplier = 1.0;
```
petervdmeer, line 322: The probability on line 325 is calculated with `1 + multiplier`, suggesting a factor 2 outside the 1/3rd range. Was that your intention?
Your comment above says that you want to increase by a factor of 3, but going from 2 to 3 is not a factor 3 increase.

source/Projectile.cpp
```diff
@@ -287,35 +295,54 @@ shared_ptr<Ship> Projectile::TargetPtr() const
...
+	// time. Jamming of 10 will increase that to about 60%.
 	if(weapon->RadarTracking())
 	{
-		double probability = weapon->RadarTracking() / (1. + target.Attributes().Get("radar jamming"));
+		double baseRadarJamming = target.IsDisabled() ? 0 : target.Attributes().Get("radar jamming");
+		double currentRadarJamming = baseRadarJamming;
+		if(baseRadarJamming > 0.0)
+		{
+			double distance = position.Distance(target.Position());
+			double jammingRange = 500 + (sqrt(baseRadarJamming) * 500);
```
petervdmeer, line 341: jammingRange get calculated on each CheckLock call for each RadarTracking projectile. Is this calculation expensive? (Is `sqrt` expensive?)

If this is expensive, then maybe we can calculate the jammingRange as a property of a target-ship when instantiating the target and just use the property instead of calculating this every CheckLock call?

Pointedstick: `sqrt` is typically implemented in hardware on the CPU to be nice and fast, so I don't think it's a bottleneck.

source/Projectile.cpp
Pointedstick: Good catch!

source/Projectile.cpp
```diff
@@ -199,6 +198,15 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 			}
 		}
 	}
+	// If the weapon is homing, is radar-guided, has lost its lock, and the
+	// target has jamming, and the missile is within the jammer's jamming radius,
+	// give a chance for the missile to turn in a random direction, in proportion
+	// to the strength of the jamming and the missile's tracking quality
+	else if(target && homing && weapon->RadarTracking() && target->Attributes().Get("radar jamming") > 0.0 && double(position.Distance(target->Position()) >= 500 + (sqrt(target->Attributes().Get("radar jamming")) * 500)))
```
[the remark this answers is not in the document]
Pointedstick: I can't remember why I did it this way--whether what you see is a bug or a difficult-to-understand implementation. Will re-check. Thanks for pointing it out, in any case.

Zitchas:
Regarding the effectiveness of optical against large ships, how does the optical guidance rating interact with the size and distance?

For instance, will something with optical guidance 1 track everything perfectly? Just under a certain distance? Or just over a certain size?

Pointedstick:
Right now Optical tracking is not affected by distance at all. It wasn't before, and I didn't add any. Overall the changes to optical tracking are very minimal.

Zitchas:
OK, so optical is purely based on size? Is there a "threshold" mass at which point a weapon with `optical tracking 1` will never lose its target?

Pointedstick:
No, but at a ship weight of 10,000,000, the probability is 98%. It's asymptotic though. It never reaches 100%.

Pointedstick:
However, as has been mentioned before, when a missile loses its lock, most of the time you don't notice because it simply stops turning but it was most likely already pointing in the right direction in the first place. This PR makes radar-guided missiles that have lost their lock have a chance to go haywire when the target has jamming. If people really want, I can apply the same haywire chance for all forms of tracking and also targets without jamming--in this case, having radar jamming and running cool (for IR missiles) would simply substantially increase the chance of the missile going haywire.

Thoughts?

Zitchas:
Hmmmm... I've got mixed feelings about missiles going haywire.

For radar? Definitely. Radar getting jammed is definitely something I'd consider a candidate for going haywire.
For optical? Likely. Jamming is likely to involve either blinding the sensor or feeding it erroneous data (or, for that matter, visual stealth patterns), all of which make some degree of haywire likely. That being said, I wouldn't expect it to be as wild as for Radar.

IR's counter is just being really cold; and that seems unlikely to result in going haywire at all. It does, however, seem highly likely that if the target ship is completely cold that the missile would have a high chance of locking onto something else in the area - either a hotter ship or the star of the system.

Basically, I'd expect each "mode" to fail differently: 
A falsified radar return could make the missile think its target is anywhere, resulting in a very chaotic haywire result.
A confused camera is limited to thinking that its target is somewhere else in front of it, resulting in a haywire behaviour that usually points away from its former target, but not wildly so. 
An IR system that has lost its target isn't really going to go haywire so much as lock onto the next best heat signature available roughly in front of it.

I'm not sure if all that is really feasible. A "cheap" aproximation might be to simply have lots of haywire chaos for the radar, and moderate-to-low for optical and IR. Which should also help keep the three tracking systems feeling different.

source/Projectile.cpp
```diff
@@ -199,6 +198,42 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
...
+	// and turn in a random direction ("go haywire"). Each tracking method has
+	// a different haywire condition. Weapons with multiple tracking methods
+	// only go haywire if all of the tracking methods have gotten confused.
+	else if(target && homing)
+	{
+		bool goHaywire = false;
+
+		// Infrared: proportional to tracking quality
+		if (weapon->InfraredTracking())
+			goHaywire = Random::Real() > weapon->InfraredTracking();
```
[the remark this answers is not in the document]
Pointedstick: Whoops, that was dumb.

source/Projectile.cpp
```diff
@@ -199,6 +198,43 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
...
+		// tracking quality, the strength of target's jamming, and the distance
+		// to the target (jamming power attenuates with distance).
+		if (weapon->RadarTracking())
+		{
+			if (target->Attributes().Get("radar jamming") == 0)
+				radarConfused = Random::Real() > weapon->RadarTracking();
+			else
+			{
+				// Closeness to target; 0 when just fired, 1 when impacting
+				double closenessTotarget = 1 + weapon->RadarTracking() - (1 / (weapon->Range() / position.Distance(target->Position())));
```
petervdmeer, line 230: Risk of division by zero if position and target->Position somehow become the same. (Unlikely, but possible when ships have open space in the middle.)

source/Projectile.cpp
```diff
@@ -199,6 +198,43 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
...
+
+				radarConfused = Random::Real() < closenessTotarget * ( 1 - (weapon->RadarTracking() / target->Attributes().Get("radar jamming")));
```
petervdmeer, line 232: Risk of division by 0 if radar jamming is zero?

source/Projectile.cpp
```diff
@@ -199,6 +198,43 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
...
+			}
+		}
+		if (infraredConfused && opticalConfused && radarConfused)
+			turn = Random::Real() - min(.5, turn);
```
petervdmeer, line 236: I'm not sure what was meant to be set as value here, but turn is used as input for angle, so I would expect the calculation here to use turn-values in the range of -180 to 180 or in the range 0 to 360 and I don't recognize those ranges.

petervdmeer (commented):
I agree with Zitchas that different haywire modes for different tracking systems would be appropriate (according to his description).

source/Projectile.cpp
petervdmeer: Are you sure that RadarTracking needs to be added to closenessToTarget? (And should closenessToTarget have 2 capital Ts?)

source/Projectile.cpp
Pointedstick: No because this code is only reachable if `if (weapon->RadarTracking())` evaluates to true on line 232

source/Projectile.cpp
Pointedstick: This gently modifies the existing missile's existing turn value to cause it to veer off of its current course without making it look like it's totally spazzing out.

source/Projectile.cpp
Pointedstick: > Risk of division by zero if position and target->Position somehow become the same

Fixed.

> Are you sure that RadarTracking needs to be added to closenessToTarget

Yes, because I want the closeness factor to be affected by the quality of radar tracking.

> (And should closenessToTarget have 2 capital Ts?)

Style thing, but it makes sense to me. :)

Zitchas:
@Pointedstick I'm not sure why this seems to have fallen by the wayside, but this and #4212 seem like good candidates for merging into the big Combat Revamp test. I'll try doing a test PR and see how it fares; but bringing these two up to par with master would be appreciated.

Zitchas:
Update: I successfully merged it into #5772 .

Pointedstick:
I've rebased on current master and squashed everything into one commit, which should make it easier for you.

Pointedstick:
I was waiting for some kind of formal approval from either @Amazinite or @tehhowch.

Or maybe we should just do it? This PR is over two years old at this point. :)

Pointedstick:
All right, let's just go for it. These are pretty tame changes compared to what you guys have been proposing recently! :) I really do think they improve missile weapons a lot. Reverting is easy if it turns out to be a huge disaster.

tehhowch:
Was the effect on non-human missiles evaluated in any manner? I see #4214 addressed human missiles, but no mention of anything else. Given that we're trying to ship 0.9.13 ASAP, introducing sweeping balance changes when MZ may release to Steam at any moment is not a great plan of action.

> Or maybe we should just do it? This PR is over two years old at this point. :)

The fact that a PR hasn't been reviewed or re-reviewed doesn't mean "merge it," it means the relative priority hasn't been high enough for the reviewers to carve out the time they need to adequately review the work.

Pointedstick:
> Was the effect on non-human missiles evaluated in any manner? I see #4214 addressed human missiles, but no mention of anything else. Given that we're trying to ship 0.9.13 ASAP, introducing sweeping balance changes when MZ may release to Steam at any moment is not a great plan of action.

That's fair. I didn't realize this. That information doesn't appear to be publicly posted anywhere, or did I just miss it?

Feel free to revert, or would you like me to? And can we un-revert after the release?

Amazinite:
>Feel free to revert, or would you like me to? And can we un-revert after the release?

You're fine to leave it.

>That information doesn't appear to be publicly posted anywhere, or did I just miss it?

It's been announced on the Discord server. As far as on GitHub goes, there's really no place to put such announcements, so the closest we can get to announcing an impending release here is updating the changelog, which was done in #5653 and merged the other week.

Pointedstick:
OK. Maybe it could be put in the README file which gets shown on github. Another option is to tag PRs that need to wait for the next release with a tag whose text indicates this, or milestone them to the release after the upcoming one.

Zitchas:
This might be slightly problematic, as up until now "0.9.15" has simply meant "Amazinite isn't going to touch this until we start merging stuff for "0.9.15". It really had nothing to do with release schedules.

PR #4209, opened 2019-03-03
introduce Hai tactical scanner

Nescio0:
The Hai highly value life, therefore it would make sense for them to have a tactical scanner to allow their heavy warships to determine the number of individuals on other ships. It is larger and more powerful than its human counterpart:
* human tactical scanner (144000 credits): 32 scan power / 4 mass = 8
* Hai scanner (420000 credits): 63 scan power / 7 mass = 9

Effectively its range is 794, about the same as the Ion cannon's 800.

Hopefully someone (@Kryes-Omega?) could design a decent outfit sprite.


Amazinite:
>The Hai highly value life

The Hai kill on sight and don't aim to disable. 
I'm not as sold on the idea of a Hai scanner as I am a Heliarch scanner.

Nescio0:
> The Hai highly value life

The lore suggest the Hai aren't bloodthirsty, e.g. `hai missions.txt` line 164:
`	He says, "I would not call it war. We launch no raids on them. We only defend when they attack. And, we supply them with food and other things needful for life. They are not our enemies, they are just those who do not fit in. You have surely seen what our society is like: safe, stable, predictable, deliberate. It suits most of us well. But there are those..."`

jafdy:
Even if your argument is sound I don't see why this would imply they need a scanner. What would they even be doing with that information? Presumably they would avoid killing any ship with a person on them, ie all of them.

Amazinite:
There's a difference between avoiding killing (as the Free Worlds do) and only being defensive (as the Hai are). Yes, the Hai aren't bloodthirsty, but that doesn't mean that they're angels who avoid killing. They still actively kill in defense of their territory.

PR #4211, opened 2019-03-03
Creation of the Salvage Scanner

Zitchas:
Adds:
 - The Salvage Scanner outfit (outfit scan 13, tactical scan 84)
 - Salvage Scanner image

Modifies:
 - Copyright for beccabunny

Note: Changes and balancing in progress for the following changes:
 - re-balanced Gascraft and Albatross built-in scan values
 - Added scanners to a variety of Remnant variants, in some cases along with an expansion as necessary.

[Link to zipped assets file](https://drive.google.com/file/d/1GGZwLbF77VJ1PkNZwPt8t36czeTE0JNW/view?usp=sharing)


data/remnant outfits.txt
```diff
@@ -341,6 +341,23 @@ outfit "Quantum Key Stone"
...
+	licenses
+		Remnant
+	cost 100000
+	thumbnail "outfit/salvage scanner"
+	"mass" 7
+	"outfit space" -7
+	"outfit scan power" 13
+	"outfit scan speed" 1
+	"tactical scan power" 84
+	description "The remnant are too few in number to support extensive ground-breaking research in all areas of technology, so they prefer to focus their efforts on a few areas of importance. To make up the deficiency, the Remnant rely on salvaging and reverse engineering technology from those around them. This scanner is a key part of this effort, as it provides the capacity to do detailed long-range scans to determine what they want to target. Its ability to do individual outfit scans is much shorter than that of its ship-level scanning equipment, but it means every Remnant ship can be on the lookout for new technology while being aware of the capabilities of the target ship."
```
toilethinges: Capitalize Remnant in the first sentence.

Zitchas: Thank you. Fixed.

data/remnant outfits.txt
```diff
@@ -341,6 +341,23 @@ outfit "Quantum Key Stone"
...
+	licenses
+		Remnant
+	cost 100000
+	thumbnail "outfit/salvage scanner"
+	"mass" 7
+	"outfit space" -7
+	"outfit scan power" 13
+	"outfit scan speed" 1
+	"tactical scan power" 84
+	description "The Remnant are too few in number to support extensive ground-breaking research in all areas of technology, so they prefer to focus their efforts on a few areas of importance. To make up the deficiency, the Remnant rely on salvaging and reverse engineering technology from those around them. This scanner is a key part of this effort, as it provides the capacity to do detailed long-range scans to determine what they want to target. Its ability to do individual outfit scans is much shorter than that of its ship-level scanning equipment, but it means every Remnant ship can be on the lookout for new technology while being aware of the capabilities of the target ship."
```
jostephd: The sentence about "Its ability is shorter" needs to be reworded since an _ability_ can't be _shorter_. You may have meant "more short-ranged" instead of "shorter", but that would sound somewhat awkward....

Zitchas: Thanks for pointing that out. That sentence has been reworded.

data/remnant outfits.txt
```diff
@@ -341,6 +341,23 @@ outfit "Quantum Key Stone"
...
+	licenses
+		Remnant
+	cost 100000
+	thumbnail "outfit/salvage scanner"
+	"mass" 7
+	"outfit space" -7
+	"outfit scan power" 13
+	"outfit scan speed" 1
+	"tactical scan power" 84
+	description "The Remnant are too few in number to support extensive ground-breaking research in all areas of technology, so they prefer to focus their efforts on a few areas of importance. To make up the deficiency, the Remnant rely on salvaging and reverse engineering technology from those around them. This scanner is a key part of this effort, as it provides the capacity to do detailed long-range scans to determine what they want to target. Its outfit scanning capability has a much shorter range than its ship-level scanning equipment, but it means every Remnant ship can be on the lookout for new technology while being focused on the capabilities of the target ship."
```
Pointedstick: `extensive ground-breaking` seems a bit excessive to me. I might remove "ground-breaking".

`This scanner is a key part of this effort` Too many instances of "this" in that sentence.

Zitchas: That does improve it. Removed `ground-breaking` and replaced `This` with `The salvage scanner`.

Zitchas:
- Reverted all changes to the Pelican & Albatross
- Added Salvage Scanner to the Albatross (Sniper)(this needed 5 tons as opposed to the old 7)
- Reduced the added outfit space on the Gull by 1 (Every variant couldn't fit the scanner as listed, with the Heavy neding 6 to fit it)

Amazinite:
The Remnant do have access to outfits expansions. Adding one expansion doesn't change much in terms of heat, and the Remnant don't run hot ships to begin with. If a ship needs more outfit space, then give it an outfits expansion, not change the base outfit space.

data/remnant outfits.txt
```diff
@@ -341,6 +341,23 @@ outfit "Quantum Key Stone"
 
 
 
+# Scanner:
+
+outfit "Salvage Scanner"
+	category "Systems"
+	licenses
+		Remnant
+	cost 100000
```
Amazinite: This is cheaper than a tactical scanner is. I think something around say 390k is more appropriate.

data/remnant outfits.txt
```diff
@@ -341,6 +341,23 @@ outfit "Quantum Key Stone"
...
+	thumbnail "outfit/salvage scanner"
+	"mass" 7
+	"outfit space" -7
+	"outfit scan power" 13
+	"outfit scan speed" 1
+	"tactical scan power" 84
```
tehhowch: This seems to be a straight upgrade over the Tactical Scanner, with basically no downsides.

|Outfits|Tactical Scan|Outfit Scan|Size|S/M|Range/M|
|:---:|:---:|:---:|:---:|---:|---:|
|2x Tactical Scanner & 1x Outfit Scanner | 2*32 = 64 (800px)| 25 (500px) | 2*4 + 2 = 10 | 6.4, 2.5 | 80, 50|
|1x Salvage Scanner| 84 (916px) | 13 (360px) | 7 | 12.0, 1.8 | 130, 51|

Why would anyone with access to the Salvage, Tactical, and Outfit scanners not choose the Salvage Scanner?

Nescio0: Perhaps increase OSP to 21 (i.e. 458 range) but reduce TSP to 49 (i.e. 700 range), to get per mass values of 3 and 7. (Human OS has 12.5, TS 8).

data/remnant outfits.txt
Zitchas: Sorry. That was a holdover from the Software Scanner. 390k would be good, but I think 410k would be better. That might help emphasize the fact that this a specialized piece of gear, not some generic scanner that is being mass produced for large swaths of the human market.

data/remnant outfits.txt
```diff
@@ -341,6 +341,23 @@ outfit "Quantum Key Stone"
...
+	description "The Remnant are too few in number to support extensive research in all areas of technology, so they prefer to focus their efforts on a few areas of importance. To make up the deficiency, the Remnant rely on salvaging and reverse engineering technology from those around them. The salvage scanner is a key part of this effort, as it provides the capacity to do detailed long-range scans to determine what they want to target. Its outfit scanning capability has a much shorter range than its ship-level scanning equipment, but it means every Remnant ship can be on the lookout for new technology while being focused on the capabilities of the target ship."
```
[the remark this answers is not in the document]
Zitchas: OK. I'll see what I can do in terms of a rewrite. Is there any particular part that doesn't read well?

data/remnant outfits.txt
Zitchas: Pretty much a straight upgrade, yes. If we had the capability to give alternate special abilities (like allowing the Remnant to target specific outfits with weapons or forcefully extract outfits without having to fully disable the ship, that'd be an interesting thing to consider.

That being said, I downgraded the outfit scanning down to 10. That should shift your values a bit.

Zitchas:
I'm not a fan of adding mass expansions to fit scanners on. These are pieces of equipment that military ships should have without sacrificing stuff. And this applies equally to the Heliarchs too. I don't see a problem with giving every military a few extra tons to allow them to fit their respective scanner whenever one is made for them.

data/remnant outfits.txt
Nescio0: For comparison, the tactical scanner has 32/4=8 TSP/mass, the proposed salvage scanner 84/7=12, i.e. its 50% better even if it would have no outfit scan at all.

Usually when outfit A has only stat x, outfit B has only stat y, and outfit C has both x and y, then:
* A is the best option for people who only want x
* B is the best option for people who only want y
* C is the best option for people who want both x and y

(E.g. systems cores.)

Amazinite:
>I'm not a fan of adding mass expansions to fit scanners on. These are pieces of equipment that military ships should have without sacrificing stuff.

Outfits expansions are already used on the Albatross (Heavy) and (Turret). A good number of Sestor ships use outfits expansions. Some Shield Beetles use outfits expansions. Prior to the scanning module, one of the Heliarch Punishers used outfits expansions. A Tempest, a Derecho, and two Hurricanes use outfits expansions. A number of Navy and FW ships use outfits expansions.

Adding one outfits expansion to allow for the installation of at least two of these salvage scanners while only sacrificing less than a percentage point of cooling is not going to be an issue.

data/remnant outfits.txt
Zitchas: This is the case when adding and comparing a comparable outfit in the same tier, but doesn't really apply when one starts looking across tiers.

The Salvage scanner is a full upgrade to the Tier 1 human tactical scanner for a faction that puts a premium on knowledge and survival. And a key part of survival is knowing your enemy's strengths *before* committing to a fight. 

A better comparison would be to include the Scanning Module in this list. It boasts 25 power for outfit, tactical & cargo for a total of 75 in a 3 ton module. Double that, and one has 50 in each of tactical, outfit and cargo and still have 1 ton left over. That's 707 range for all three versus 914 and 360 for two. Overall, that's a lot more scanning potential than the Salvage Scanner can provide.

In terms of per mass calculations for the totals, that's 25 per ton for the scanning module versus 13.9 per ton for the Salvage Scanner. That's almost double the effective power. And that seems reasonable, given that the Heliarchs are more advanced than the Remnant in most areas. It also fits, as the Heliarchs haven't had an active war in a very long time, and so are mostly interested in law enforcement (which presumably includes scanning for illegal weapons and cargo). 

Remnant, on the other hand, have been in active warfare for over a century. At least initially, they would have been significantly underpowered compared to the Korath, and forced to rely on superior tactics and long-range information gathering to pick and choose targets. Even in the current point in time, they don't have a significant edge other than the fact that they can quickly bring in support to swarm hostiles (aka fast fleet spawns in their home territory). This (and missions where the player is asked to launch surveillance satellites) implies that they continue to rely on knowledge based tactics and strategies to give them an edge. 

A better question would be "If the players can reach the Remnant, and thus can almost certainly reach the Coalition, why would they use the Salvage Scanner when they could get the Scanner Module for far less space?" And the answer would be "Only if they just want more long range Tactical Scan capabilities." Since the Scanner Module is clearly far superior in as an overall sensor package.

Zitchas:
OK, to clarify my statement a bit: I don't think it is appropriate to use expansions to accommodate key outfits on active-combat ships that date to the same era as the outfit in question. 

Expansions are, by their very description, inefficient temporary conversions of cargo space into outfit space. The Remnant are a highly adaptable and innovative faction involved in constant warfare. A mass expansion makes sense for the emergency inclusion of some new device that they deem essential to throw on their ships now. It does not make sense for an essential outfit that has been part of their repertoire for decades at the least, and more likely centuries.

Assuming that the Albatross does in fact predate the scanner (which it might not), I would expect that space adjustments would have been made in the next revision of the ship after it was decided to use it regularly in some configurations. For gameplay purposes we don't list all the iterative versions of ships, but that doesn't negate the fact that they almost certainly exist; simply that we are dealing with the most recent version.

data/remnant outfits.txt
Nescio0: The difference is the new Heliarch Scanning Module is not for sale; the only way to obtain it is by plundering Heliarch ships (which are clearly stronger than the Remnant).

Nescio0: Let's compare scan power / mass for all scanners:
* Asteroid Scanner: 40/6=6.667 (AS)
* Cargo Scanner: 9/1=9 (CS)
* Outfit Scanner: 25/2=12.5 (OS)
* Tactital Scanner: 32/4=8 (TS)
* *Navy Surveillance Pod*: 4/5=0.8 (CS), 16/5=3.2 (OS), 9/5=1.8 (TS)
* Remnant Salvage Scanner: 13/7=1.86 (OS), 84/7=12 (TS)
* *Heliarch Scanning Module*: 25/3=8.333 (CS, OS, TS)

Zitchas: Thanks, that's helpful! 

And yep, the Salvage Scanner looks right about where I was aiming it in terms of both outfit scanning and tactical scanning.

(Just for the record, I expect the Wanderers to "rediscover" an antique tactical scanner at some point to go with their rediscovered warships, and the Korath should probably have one too. Although in both cases the power and attributes of those should reflect their respective cultures and priorities.)

Zitchas:
@Amazinite There you go.

data/remnant outfits.txt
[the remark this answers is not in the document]
Zitchas: @tehhowch I changed it to:
"When the Remnant unraveled the secrets of the alien point defence turrets that guarded the vaults they found on Aventine, they also deciphered the mechanisms that guided the ancient weapons. With a little work they transformed those guiddance systems into the fundemental tools that support their aggressive pursuit of new technology."

Is that better?

data/remnant ships.txt
```diff
@@ -237,7 +239,7 @@ ship "Gull"
 		"fuel capacity" 600
 		"ramscoop" 1
 		"cargo space" 175
-		"outfit space" 310
+		"outfit space" 316
```
Amazinite: Ship | Extra outfit space 
------|-------
Gull | 55
Gull (Bunks) | 33
Gull (Cargo) | 33
Gull (Heavy) | 5
Gull (Sniper) | 3 
Gull (Support) | 1

Three of the six variants already have room for salvage scanners. The other three could very easily install an outfits expansion, as they don't have any cooling installed to begin with that would be harmed, and the loss in heat capacity due to the reduction in heat dissipation is in the low double digits compared to the four-digit heat capacity. If that's a concern, simply install a cooling outfit as well.

data/remnant ships.txt
```diff
@@ -438,7 +446,7 @@ ship "Starling"
 		"fuel capacity" 900
 		"ramscoop" 1
 		"cargo space" 36
-		"outfit space" 312
+		"outfit space" 319
```
Amazinite: Ship | Extra outfit space
-----|-----
Starling | 11
Starling (Heavy) | 3
Starling (Hunter) | 1
Starling (Sniper) | 1
Starling (Thrasher) | 6

Seeing as how the lowest extra space is 1, Starlings would only need +6 outfit space, not +7, but again, these ships could use outfits expansions. Starlings, like Gulls, lack cooling and only lose low double digits heat capacity from the installation of an outfits expansion, while the room made by installing an outfits expansion would allow for both a scanner and some cooling.

Amazinite:
>All remnant ships that get the scanner have their outfit scan power reduced from 12 to 3
>All remnant ships that get the scanner have their tactical scan power reduced from 26 to 6

I would just say reduce the built-in scan power on all Remnant ships, then change the description of the Albatross, as I believe it was, to describe Remnant ships as having weak, short-range scanners built in that are typically boosted by more powerful external scanners.

Zitchas:
You convinced me to drop installing the SS by default on their capital ships, now I rather like the idea of having three tiers of scanning present:
1. Ships built to have the Scanner installed, which have a  bare-minimum backup system.
2. Ships built without the expectation of having a Scanner, which have the current stats (outfitting at half the power of an Outfit Scanner, and Tactical at 81% of a tactical scanner.)
3. Ships from group 1 that have the Scanner installed, which have excellent long range tactical scanning and sufficient outfit scanning.

This also provides constructive lore as to why the one would have a little extra space for scanning equipment while the other doesn't.

data/remnant ships.txt
Amazinite: >The lack of cooling I believe is a failure on the part of the design of the original outfit set.

I'd disagree, given that they don't actually need the cooling (and some variants simply can't fit any given their current extra space).

Zitchas:
As per Derpy's Recommendations:
- Changed built-in scanning capabilities to be in line with  size. (Albatross & Pelican reverted to their original built-in values; Gull, Starling, & Gascraft decreased.)
- Removed the Gull "Heavy" variant (Now no gulls have Thrashers, which should encourage the "stay back and support or ninja salvage," which is their intended place in combat.
- Removed Scanner from Starling (Heavy) and Starling (Thrasher) variants.

Zitchas:
Changes in the last two updates:
- Starling outfit space reverted to their original space.
- Starling Sniper and  Hunter no longer have scanners.
- Gull Sniper and Support now have an outfit expansion
- Gull Sniper and Support scanners increased to 3.

(It didn't make sense for them to sacrifice that much cargo just to get a couple extra tons of space to fit the scanner. So they're fitting as much of the extra space as possible with scanners so at least that space is being put to use, and thus the Sniper and Support gulls can now take on a lore role as salvage coordinator ships that handle the ultra-range scanning for the ships active in their area.)

data/remnant ships.txt
Zitchas: Well, be that as it may, the change to their space has been reverted and scanners either fit without the changed space, or use expansions to do so. So this point is resolved.

data/remnant outfits.txt
```diff
@@ -343,6 +343,23 @@ outfit "Quantum Key Stone"
...
+	licenses
+		Remnant
+	cost 410000
+	thumbnail "outfit/salvage scanner"
+	"mass" 7
+	"outfit space" -7
+	"outfit scan power" 13
+	"outfit scan speed" 1
+	"tactical scan power" 84
+	description "When the Remnant unraveled the secrets of the alien point defence turrets that guarded the vaults they found on Aventine, they also deciphered the mechanisms that guided the ancient weapons. With a little work they transformed those guidance systems into the fundamental tools that supports their aggressive pursuit of new technology."
```
MCOfficer: This entire paragraph feels a bit clunky to me. Specifically:

1. the first half sentence (everything until `Aventine`) is basically a giant string of "which"s, and there's another one right after `mechanisms`
2. `With a little work` is... i can't think of a better wording, but i get the feeling there is one. With some tweaking? research?
3. A salvage scanner is technically a tool, but more of an outfit. and it doesn't really strike me as fundamental; the role it plays for the remnant is important, but it's not a fundamental tool or piece of equipment.
4. I stumbled over `agressive pursuit`, too. They're indeed pursuing new technologies, but they're not particularly agressive about it. The only context in can think of in which the remnant actively raided other alien species is the Inhibitor description (`"We aren't going home until we steal that weapon."`). So: While they are very invested in R&D, the efforts they make as well as the reason they want new tech are primarily defensive.

I'd probably ignore one such flaw, but they appear en masse in this description.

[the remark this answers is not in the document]
Zitchas: Good catch. In conjunction with point 3 from your other comment, how about changing it to "into an important tool that supports their pursuit of new technology." ?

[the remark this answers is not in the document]
MCOfficer: Either that, or `an important tool in their pursuit` or `for their pursuit` (pretty sure the latter isn't valid)

Zitchas: Thanks. I tried editing it and ended up with this:

"When the Remnant unraveled the alien point defence turrets that guarded the vaults on Aventine, they also deciphered the mechanisms that guided the ancient weapons. After much research they transformed those guidance systems into important tools for observing alien ships."

Zitchas: Upon reflection, how about this:
"When the Remnant unraveled the alien point defence turrets that guarded the vaults on Aventine, they also deciphered the mechanisms that guided the ancient weapons. After much research they transformed those guidance systems into long-range observation systems."

MCOfficer: that sounds a lot better. i assume you will combine this with the suggestion above (https://github.com/endless-sky/endless-sky/pull/4211#discussion_r266257960)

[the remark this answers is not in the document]
Zitchas: I'm just going to mark this "resolved" to keep the discussion in one spot.

Zitchas: Modified it a bit more. And yes, the previous comment and this one are more or less rolled together at this point. How does this sound?

"When the Remnant unraveled the alien point defence turrets that guarded the vaults on Aventine, they also deciphered the mechanisms that guided the ancient weapons. After much research they transformed those guidance systems into important tools for analyzing alien ships."

MCOfficer: i'm still stumbling over the repeated "that" in the first sentence. the second one could just be "their guidance systems". and "into important tools" also just feels... *off* to me.

@Amazinite @Janaszar @Pointedstick y'all muricans are better at this than i am - does this bother you?

Zitchas: I think I've fixed it. @MCOfficer How's this? Down to only a single "that" now.

"When the Remnant unraveled the alien point defence turrets guarding the vaults on Aventine, they also deciphered the mechanisms that guided the ancient weapons. After much research they transformed those guidance systems into important tools for analyzing alien ships."

Zitchas:
Thanks to both of you for all your work on this!

LatticeRegulator:
Looks Really good!

data/remnant outfits.txt
```diff
@@ -341,6 +341,20 @@ outfit "Quantum Key Stone"
...
+	licenses
+		Remnant
+	cost 410000
+	thumbnail "outfit/salvage scanner"
+	"mass" 7
+	"outfit space" -7
+	"outfit scan power" 13
+	"outfit scan speed" 1
+	"tactical scan power" 84
+	description "When the Remnant unraveled the alien point defence turrets guarding the vaults on Aventine, they also deciphered the mechanisms that guided the ancient weapons. After much research they transformed those guidance systems into important tools for analyzing alien ships."
```
tehhowch: `<...>. After significant investments in research and development, they have transformed those guidance systems into sophisticated scanning technology.`

data/remnant outfits.txt
```diff
@@ -341,6 +341,20 @@ outfit "Quantum Key Stone"
...
+	description "	While all Remnant ships have built in outfit and tactical scanners in order to determine whether an enemy ship is worth engaging and looting, some captains still prefer to install external scanners in order to have superior scanning ranges."
```
tehhowch: `"    While all Remnant ships are equipped with internal scanners, some captains still prefer to boost the range and power to more effectively select enemy ships worth targeting."`

Could also be `their range and power` instead of `the range and power`.

data/remnant outfits.txt
Zitchas: I like it. It flows better than what I had. Thanks!

data/remnant outfits.txt
Zitchas: I'm not convinced on this, but I think it is probably an improvement on mine, so adding it in.

data/remnant outfits.txt
```diff
@@ -341,6 +341,20 @@ outfit "Quantum Key Stone"
...
+	licenses
+		Remnant
+	cost 410000
+	thumbnail "outfit/salvage scanner"
+	"mass" 7
+	"outfit space" -7
+	"outfit scan power" 13
+	"outfit scan speed" 1
+	"tactical scan power" 84
+	description "When the Remnant unraveled the alien point defence turrets guarding the vaults on Aventine, they also deciphered the mechanisms that guided the ancient weapons. After significant investments in research and development, they have transformed those guidance systems into sophisticated scanning technology."
```
Amazinite: `defence` -> `defense`

Zitchas: Sorry about that. Fixed.

PR #4214, opened 2019-03-05
tweaked Surveillance Pod

Nescio0:
Currently the Navy Surveillance Pod has three different ranges (CS 200, OS 400, TS 300), but because it's a single device, it would make more sense to have only a single range. This proposal:
* standardizes scan powers to 9 (i.e. 300 range)
* adds an asteroid scan power (because the Surveillance Pod was introduced years before the Asteroid Scanner)
* alphabetizes scan attributes

### DIFF
--- a/data/outfits.txt
+++ b/data/outfits.txt
@@ -246,11 +246,12 @@ outfit "Surveillance Pod"
 	thumbnail "outfit/surveillance pod"
 	"mass" 5
 	"outfit space" -5
-	"outfit scan power" 16
-	"outfit scan speed" 2
-	"cargo scan power" 4
-	"cargo scan speed" 2
+	"asteroid scan power" 9
 	"atmosphere scan" 100
+	"cargo scan power" 9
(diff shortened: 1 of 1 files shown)


Zitchas:
Standardizing things doesn't really seem like the best reason to change these variables. Is there a lore reason why this change would be beneficial?

I disagree with this change for several reasons:
1. The Navy isn't in the business of mining. They have no interest in tracking asteroids.
2. The ability to scan cargo, outfits, and ships are not identical mechanisms, and thus different things may be stronger or weaker. This isn't a problem.
3. The Navy probably has varying priorities. For instance, they may want the most power for outfit scanning to catch pirates with illegal weapons as things like that are more dangerous than simple illegal cargo. 
4. The description mentions that "it was designed to fit perfectly inside one of their surveillance drones." This implies that it is not necessarily the best available scanning in all areas, but rather the best they could fit inside this particular drone. This also reinforces the fact that they built it to prioritize specific things.

PR #4219, opened 2019-03-06
Tweak and differentiate the starting ships

Pointedstick:
This PR tweaks the starting ships to differentiate them a bit better from one another and adjust to the recent changes to missiles.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -1259,7 +1259,7 @@ ship "Finch"
 		"hull" 200
 		"required crew" 1
 		"bunks" 1
-		"mass" 30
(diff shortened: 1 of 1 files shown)


Amazinite:
>Slightly increase drag so it no longer nonsensically turns faster than the Sparrow.

Drag doesn't effect turn rate, only max speed.

Speed = thrust / drag
Acceleration = thrust / mass
Turn rate = turn / mass

Amazinite:
I'd leave the drag of the Shuttle be, as there's really no point in touching it.
Otherwise, I'd say these are alright changes.

Issue #1605, opened 2016-09-21
[Mechanincs] Events Creating Systems


comnom:
So I was working with someone on steam who wanted a system to be created after a set of missions was complete. After working out a few errors, we got the event working, however the map would not show the system. So I modified the event to link it to an existing system, and this works but the system will not show on the map until you manually jump to it(the link just leads to empty space in the map).

So I'd like to know if this is a bug, intended feature, or just an unforseen use case.

Amazinite:
Makes it super secret. 
If the event-made-system is not linked, is it still reachable by jump drive if you use the jump key to toggle your destination?

KiLEdEnNis:
what I use; make sectors beyond jump-drive range (making them invisible), and have an event link a sector to a known sector

comnom:
>Makes it super secret.
If the event-made-system is not linked, is it still reachable by jump drive if you use the jump key to toggle your destination?

I didn't check to see if you could use a jump drive to hop to it manually, but when toilethinges and I were playing with it, it made toggling routes screwy. The only way to get to it was to be in the linked system, point your ship in the general direction and press J.

>what I use; make sectors beyond jump-drive range (making them invisible), and have an event link a sector to a known sector

His story requires it to be an unlinked system, discovered through the missions.

endless-sky:
The issue was that the game calculates what systems are visible from each system when the map is loaded, and wasn't updating those calculations when the map changed. Commit a266bb7 should fix that.

PR #4238, opened 2019-03-28
parse events for definitions while ref checking

comnom:
Both can be created entirely through events, so an undefined planet or system is acceptable.

### DIFF
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -213,37 +213,31 @@ bool GameData::BeginLoad(const char * const *argv)
 // Check for objects that are referred to but never defined.
 void GameData::CheckReferences()
 {
-	// Check events first, as they may contain additional definitions.
-	map<string, set<string>> deferred;
+	// Parse all GameEvents for object definitions & references.
+	auto deferred = map<string, set<string>>{};
+	const auto eventDefinitionNodes = set<string>{
+		"fleet",
+		"galaxy",
+		"government",
+		"outfitter",
+		"news",
+		"planet",
+		"shipyard",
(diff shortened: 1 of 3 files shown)


tehhowch:
While they can be, IMO they should not be. I consider the usefulness of detecting incorrect uses of a planet where a system was expected, or a system where a planet was expected, far more helpful to mission writers.

I believe MZ had a comment somewhere about using an event to finalize already-defined items, i.e. to set the position & links of a system, or assign a planet to a system, etc.

MCOfficer:
Why not print out exactly that, that it **may** cause issues but can also have no consequences at all? that's what warnings are for.

comnom:
>While they can be, IMO they should not be.

It's been supported for quite a while (https://github.com/endless-sky/endless-sky/issues/1605), and more than one plugin uses the feature. Akasha Chronicles and the ever popular Ursa Polaris to name a couple.

For new plugin authors, having warnings for supported behavior would be confusing.

comnom:
>Why not print out exactly that, that it may cause issues but can also have no consequences at all?

However you word it, the it still amounts to: "Warning: what you are doing is ok." Things like that lead people to ignore other, perfectly valid warnings.

tehhowch:
The warning will stay. However, what should happen instead, to address this use case, is to scan through `GameEvent` changes to check for additional definitions.

comnom:
>However, what should happen instead, to address this use case, is to scan through GameEvent changes to check for additional definitions.

Fair enough. I'll revert the removal and add that shortly.

PR #4249, opened 2019-04-11
windows console output

comnom:
A refined version of my earlier poor attempt. Attaches a console that has been detached due to the nature of building with subsystem windows and redirects stdout and stderr to it, or, failing that, allocates a new console.

<details>

consoleCapture

</details>

Redirection:
[out.txt](https://github.com/endless-sky/endless-sky/files/3069220/out.txt)

### DIFF
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -570,23 +570,23 @@ Conversation LoadConversation()
 #ifdef _WIN32
 void InitConsole()
 {
-	bool redirectStdout = (_fileno)(stdout) == -2;
-	bool redirectStderr = (_fileno)(stderr) == -2;
-	
-	if(!redirectStdout && !redirectStderr)
+	// If both stdout and stderr are already initialized (e.g. writing to a file), do nothing.
+	const int UNINITIALIZED = -2;
+	bool redirectStdout = _fileno(stdout) == UNINITIALIZED;
+	bool redirectStderr = _fileno(stderr) == UNINITIALIZED;
+	bool redirectStdin = _fileno(stdin) == UNINITIALIZED;
+
+	if(!redirectStdout && !redirectStderr && !redirectStdin)
 		return;
-		
-	if(!AttachConsole(ATTACH_PARENT_PROCESS))
-	{
-		AllocConsole();
-		return;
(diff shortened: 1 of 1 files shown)


MCOfficer:
Works on my machine (Win 10 1803), well done.

Refs #2146

tehhowch:
I'm curious if there's a nicer way to handle launching ES from a prompt on Windows, as it doesn't seem to play very nicely with output or blocking. Ex launching with `-d`, to print parse info, yields same-line overruns:
image

Unlike launching from a git terminal (e.g. the MSYS2 one that comes with git-for-windows), the command prompt doesn't block on other input either. You can freely use the terminal for input or other applications even though ES will be/is writing to it:
image

These issues are mentioned in this decade-old post: https://social.msdn.microsoft.com/Forums/windows/en-US/3273ff67-2f97-4df5-b5df-2216690a5d8c/console-and-gui-app-in-one?forum=winforms

(A bit more digging and I found this post: http://www.windowsinspired.com/how-to-add-a-debugging-console-to-a-windows-gui-application/ which seems like it might help with this behavior.)

comnom:
I'm just using the methods that Microsoft recommends for a GUI application. I don't alter the behavior of cmd.exe, and wouldn't since a console can have many processes attached to it at once.

I dont' have any knowledge about how msys2's terminal emulators work, so I can't say much about any features you've got there that you don't get in cmd.exe.

I haven't had any same-line problems on Win 7, so I'll have to look into that.

<details>

Capture


</details>

tehhowch:
A bit more testing showed that you need to redirect the streams even for the call to `AllocConsole`.

I'm tempted to move this to the `Files` class and invoke it when `LogError` is called, to prominently highlight when there's crap happening (instead of "oh, there's an errors.txt file?"), but I think that's something for a separate PR and also would require some way to achieve similar behavior on *nix. It would also probably be pretty annoying since almost every older save file will have references to something that has now been changed or removed or cleaned up, so almost everyone would have a terminal pop up alongside the game, and they'd have no real way to fix the issue except digging through a possibly large save and editing it manually.

Issue #2970, opened 2017-08-26
Turret mount hardpoint sprites - bases


Bladewood:
![heavy anti-missile 2x](https://user-images.githubusercontent.com/16506683/29738313-77bcd39c-89ed-11e7-8d37-4aa2e64c84fd.png)
![dual sunbeam turret 2x](https://user-images.githubusercontent.com/16506683/29738315-7ccc525e-89ed-11e7-9670-b4a553d55c84.png)
They tend to look weird with the turret bases.
![sillyturrets1](https://user-images.githubusercontent.com/16506683/29738326-df62fdfa-89ed-11e7-8d77-c65386475204.png)
![sillyturrets2](https://user-images.githubusercontent.com/16506683/29738327-e08bcb3a-89ed-11e7-9797-c956b64a91bd.png)

Also, @rob59er brought up the issue that the regular Anti-Missile Turret is basically invisible...

Rob59er:
https://drive.google.com/drive/folders/0B9khRJPmS75hWWpvOUItSG9qYTQ?usp=sharing collection of messed up/weird examples of turret sprites.

KiLEdEnNis:
to make it look way better
something to not let the turret base move along would require rendering the base independently and have that drawn below the rotating turret and not rotate itself
to cover turret mounts, something similar with size difference in engine flares could do, on bigger ships the same turrets look bigger yes but at least they cover the base

Bladewood:
independent base and turrets: yes, that was what I was hoping for.

tehhowch:
One thing to remember is that the bases will render by default pointing away from the center of the sprite

endless-sky:
In my opinion the only real issue is the anti-missiles - the base being way too light-colored on one, and nonexistent on the other. So, the solution might just be to rework those two sprites.

Pointedstick:
@endless-sky agreed. Smaller bases would help, as would tweaking the hardpoint locations on ships like the Kestrel that currently have the hardpoints on the outer edge of the hull.

Tadrix:
How about not including bases in hardpoint images at all? They would be embedded into the hull anyway. As it is now, they look "glued" on the hull, instead of being part of the ship.

KiLEdEnNis:
it's not unreasonable to think the turret base gets mounted on top of a hardpoint, meaning the base is visible too

Tadrix:
Looks weird to me (in which I concur with Bladewood), but it could be pertaining to taste.

[...]

PR #4250, opened 2019-04-11
edit human AM for better visibility

comnom:
refs: #2970

[assets.zip](https://github.com/endless-sky/endless-sky/files/3070915/assets.zip)


Amazinite:
Personally not too keen on the change to the basic AM. How about pulling a page from the Bullfrog AM's book and add some sort of plating around the top rim?
image

tehhowch:
Please either update the copyright file, or explicitly state that these works should remain copyright their existing holders.

Pointedstick:
If the regular AM turret's transparent base is part of the problem, than this latest version of the sprite doesn't fix that.

comnom:
>Please either update the copyright file, or explicitly state that these works should remain copyright their existing holders.

Copyright's fine as is.

>If the regular AM turret's transparent base is part of the problem, than this latest version of the sprite doesn't fix that.

The way I read it, not having a base was the problem. I don't think it needs to be solid. The base shows up poorly on a gray checkerboard anyhow. Here it is on white and black.

Old:
anti-missileOldonwhite

anti-missileOldonblack

New:
anti-missileNewonwhite

anti-missileNewonblack

Amazinite:
I think the considerably increase in base thickness is enough to help fix the issue,

PR #4254, opened 2019-04-13
Replacing Tali with Taely

Zitchas:
In order to enable this Easter Egg reference of sorts to Tali from Mass Effect to be more of an Easter Egg and less of a distraction, I would like to propose renaming her Taely. She still has a very similar name, but without the exact same spelling it is hoped that this character will favourably remind people of the ME Tali without people actually conflating the two.

This PR changes:
- All instances of `Tali` replaced with `Taely`
- All instances of `tali` replaced with 'taely' (primarily encounter counters and branches)
- All log entries in missions that introduce the player to Taely have been removed.
- Adds one invisible mission that triggers when the player has met Taely 1 or more time, which gives the introductory log entry.
- Adds one invisible mission that checks for the presence of "remnant tali" counter, and if present, gives the player one "remnant taely" counter and clears the "remnant tali" counter.

This change reduce the duplication in the code, and also simplifies future modifications for:
a) when the current tech retrieval missions have branching dialog added to handle meeting Taely to take into account it may not be the first time, and
b) to simplify the addition of other tech retrieval missions to resolve the problem of people getting stuck at that point, as said mission(s) wouldn't have to deal with the log entry, and
c) prevents the potential duplication of the initial introductory log entry if a player someone does more than one of these missions.


Zitchas:
Thank you for reviewing the changes!

tehhowch:
Any existing log entries that have already been entered into players' logbooks will not get updated, nor will existing, accepted missions, condition keys, etc. Thus changes like this should be reserved for exceptional circumstances. Is it really, truly, disruptive for the character's name to be Tali?

I agree with the deduplication changes.

Zitchas:
@tehhowch @Amazinite I'm not really sure how many people have done these missions, but I suspect it is fairly low, and there is at least one or two people already talking about the names being confusing. If that's the case with the comparatively small sample size that is active on here and discord, I suspect it will be quite a bit more numerous once it hits steam. Gentler on everyone to make the change now as opposed to later. 

That being said, the Wanderer Ramscoop mission is not restricted like the other Tech Retrieval missions are, and is triggered by tech that later-game players are more likely to have, so chance are they will receive the log entries and have the variables updated at that point. 

I might be able to make some kind of a "catch up" invisible mission that converts the tali counter into the taely counter, but honestly, I'm not sure adding a mission to do that, which will be part of the game forever, is worth the extra dozen lines of code.

Or could we add that in now, and then promptly remove that for 0.9.10, as hopefully everyone playing the nightly now gets caught up in the meantime?

Zitchas:
@jafdy @Pointedstick You two seem to be the ones who have the strongest opinions about why Tali's name should be changed. Would you care to make some of your points about it?

I still like keeping the Tali name, but I created this PR to change her name so that it wouldn't be a problem for those who can't separate the two characters. I'd rather have a subtle reference if it means players can enjoy both characters instead of amalgamating them into one.

Zitchas:
@Amazinite @tehhowch To keep things a bit cleaner and prevent potential loss of progress, I have added a short invisible mission that checks to see if the player has the `"remnant tali"` counter, and if so, increments the `"remnant taely"` counter by one (likely creating it), and then clears the `"remnant tali"` condition.

If it is possible to clear the log entry too, I'll add that in here as well.

data/remnant missions.txt
```diff
@@ -814,6 +814,17 @@ mission "Remnant: Taely Intro Log"
 
 
 
+mission "Remnant: Tali to Taely"
```
Amazinite: No instances of `"remnant tali"` even exist right now.

Zitchas: Ah, yeah. That was added during the de-duplification stuff.

Zitchas:
Change the invisible mission so that instead of switching a counter that was never actually there, it checks to see if any of the three tech retrieval missions are done, and if so, gives the player the `"remnant taely"` counter. Which should then trigger the Taely log entry.

I don't know how to get rid of the Tali log entry, but this will at least ensure that nothing that gets built on the Taely counter down the road will be missed by these players.

jafdy:
We are still between versions, and if we are too afraid to ever go back and change things we will dampen the rate of progress still further. Unless you expect every PR to be perfect changes like this will still need to be made occasionally.

MCOfficer:
I agree with jafdy, everyone who has played this already this is a factually a beta-tester, and such things happen between versions. Also, it's not a huge, breaking change, the name is still recognizable. Bottom line, the "damage" done by this PR is absolutely minimal.

So the question is only if this change is necessary, to which i say yes - there are likely to be more than a dozen ME fans in among ES's players.

comnom:
>Is it really, truly, disruptive for the character's name to be Tali?

I can think of a few things more irrelevant than the name of a fictitious character, but not very many. Any disruption for something so minor isn't something I could get behind.

If the reader can't separate two different characters in two different made up universes with the same name, I'd say that's an issue with the reader, and not the writer.

Zitchas:
Just to be clear, at this point the sum total "damage" that we are talking about is that a small number of nightly players will have an extra log entry for a person named "Tali" alongside the entry for "Taely."

jafdy:
A similar sound is a better tribute than identical spelling, people who know both games will spot it but it won't jump out as it would under the same spelling. When I read the first mission (with "Tali") the thing that really jumped out was that this was a mass effect character, and that isn't something you want if you are trying to develop some other character. A different similar name will give you some time to develop the character before the reader works it out.

Amazinite:
@Pointedstick Is the difference between Zitchas's Tali and my [Tom Paris](https://github.com/endless-sky/endless-sky/pull/3120#discussion_r240879254) simply the similarity between our character and the referenced character? (I gave Lt. Paris the first name of Tom simply because you asked if that was the case, so any similarities between the two are completely coincidental.)

jafdy:
I have the same view on that, I'm just not a star trek fan so am rather less bothered by it.

Zitchas:
As an additional note, replacing Tali with Taely shifts the pronunciation slightly to more closely resemble "Kaylee" from Firefly. Also being a ship mechanic of probably equal skill to Tali, this is desirable in my view, as it makes the tribute a reference to either story. Which makes it slightly more meaningful, since there are people who have watched the one and not played the other. 

Given that she's one of inspirations for the character, (albeit a secondary one), I think it is fitting. (and we already have a Firefly reference in-game, so that fits too).

data/remnant missions.txt
```diff
@@ -792,12 +792,39 @@ mission "Remnant: Plasma Cannons"
...
+		government "Remnant"
+	to offer
+		"remnant taely" >= 1
+	on offer
+		log "People" "Taely" `As an engineering prefect, Taely is the foremost authority among the Remnant on salvaged technology and a senior leader in their reverse engineering program.`
+		fail
+
+
+
+mission "Remnant: Tali to Taely"
```
Amazinite: Given that the old log can't be removed, I'd just get rid of this. Anyone with both logs is just a casualty of playing missions while they're in the middle of development.

data/remnant missions.txt
```diff
@@ -792,12 +792,39 @@ mission "Remnant: Plasma Cannons"
...
+
+
+
+mission "Remnant: Taely Intro Log"
+	description "You have just met Taely, an engineering prefect among the Remnant"
+	invisible
+	source
+		government "Remnant"
+	to offer
+		"remnant taely" >= 1
```
Amazinite: This condition should be more descriptive, e.g. `"remnant met taely"`. Same goes with any future conditions.

data/remnant missions.txt
```diff
@@ -792,12 +792,39 @@ mission "Remnant: Plasma Cannons"
...
-			`	Tali steps forward to face you. "Well met, newcomer among the Remnant," she trills. "Fresh blood amongst us sets songs in motion that have long been in caesura. I have a few things to finish up right now, but I hope to work with you soon." She gives you a polite nod and heads back the way she came.`
+			`	The weapon technician finishes strapping down the plasma cannons and looks up to see another Remnant approaching. He gives the new Remnant a respectful dip of his head that almost seems like a bow before turning to you and chanting, "Ah, Captain <last>, have you met Taely? She is the engineering prefect here on Viminal. She asked me to let her know when or if you managed to procure the new tech for me."`
+			`	Taely steps forward to face you. "Well met, newcomer among the Remnant," she trills. "Fresh blood amongst us sets songs in motion that have long been in caesura. I have a few things to finish up right now, but I hope to work with you soon." She gives you a polite nod and heads back the way she came.`
+
+
+
+mission "Remnant: Taely Intro Log"
```
Amazinite: I believe that invisible missions like this typically have lower cased names.

data/remnant missions.txt
```diff
@@ -792,12 +792,39 @@ mission "Remnant: Plasma Cannons"
...
+	description "You have just met Taely, an engineering prefect among the Remnant"
```
Amazinite: Again with the descriptions on invisible missions...

data/remnant missions.txt
Zitchas: Removed the ineffective "clear" line.

Added to offer condition so that this only triggers when the player has done one of three missions prior to this fix being implemented. This ensures that branches in conversations will happen properly.

data/remnant missions.txt
Zitchas: Changed to lower case.

data/remnant missions.txt
Zitchas: Removed this description.

data/remnant missions.txt
Zitchas: Changed.

Zitchas:
@Amazinite Thanks for catching those things!

Issue #3066, opened 2017-09-20
Gascraft description: inconsistency in pluralization of "Remnant"


Bladewood:
> Centuries ago, **the Remnant built** these ships in order to explore the upper atmosphere of nearby gas giants. The Archons took issue with **the Remnant's** research methods, and the studies were unable to be completed. The remaining gascraft have been gathering dust in storage for centuries while **the Remnant waits** to see if the anger of the Archons will subside.

It's apparent from all other appearances that "Remnant" is plural. 
[The part with an apostrophe is apparently insoluble, but just marking that here.]
If we're following the pattern in pluralization, `the Remnant waits` should be `the Remnant wait` instead.

endless-sky:
I'm actually not sure whether singular or plural makes more sense for the Remnant in general. I've been using a plural when using "the Remnant" as shorthand for "the people of the Remnant," and singular when just talking about "the Remnant" as a faction.

endlesssteve:
Just give my penny worth as native British English speaker.
I don't think there is real difference between "the people of the Remnant" and "the Remnant faction" and as such they should both be plural. Only a organization not directly linked to a specific group of people is singular.
 e.g. the Vatican. the Maffia. the Government  --- singular
the Romans. the Celts. the Toilet cleaners --- plural
However I've played this part several times and didn't even notice. It didn't spoil my game.

azerwyn:
tl;dr -- It doesn't matter.  The dominant, persistent voice(s) will define the inflection norm.

-------
Some words have irregular plurals that do not fit any of the defined grammar rules.  "Proper" use is largely steered by dialectic, cultural, social ear-training -- and thus is fungible.  

**person – people (also persons, in more formal contexts; people can also be a singular noun with plural peoples.)**

Thus Roman and Romans are acceptable with appropriate context -- singular people and plural peoples.
And Maffia and Maffias are also both acceptable with context -- The Formal, the group, group of groups.

You might dare to offend the Italian and Russian Maffias, but don't cross the Cheerleader-Mother Maffia.

The Vatican is the only one of the examples above that is truly singular and that is because it is unique.  It would be a linguistic contortion to refer to the many seats of religious power (Coptic, Orthodox, Catholic, Apostolic) as Christian Vatacans.  However, that is not to say that over time linguistic drift could adopt such an inflection.

[...]

PR #4255, opened 2019-04-13
Spelling & Grammar Fixes

Janaszar:
I ran through the various files in `/data` for spelling and grammar errors.
Most of the changes are obvious; spelling errors, subject-verb agreement, duplicate words, hyphenating compound words.

Frontier Jobs line 269: Re-positioned the quote, as you wouldn't say "As you call them out..."

Remnant Missions line 1200: You're talking about the group of Remnant students agreeing to meet back up.  I'm going to assume that "remnants" will be treated like "humans" in that it did not need to be capitalized.


Nescio0:
"This group of <bunks> university students" (two occurrences) is singular, therefore your change from "is" to "are" is wrong; same for "This family of" (eight occurrences).

Janaszar:
If the number of students/family members were removed, along with the portion that uses `of`, I would agree (`This group is...`, or `This family is...`).  But, the text is talking about a group of 20 students, or 20 miners, which turns `group` from the noun to an adjective/descriptor.

This - Singular, pointing at the Group/Family
Group/Family - Describing the people referenced compared to others; singular, as there is only one group
Of <number> - Tells you how many of them are in the group, turning it plural
Students/Miners - What they are.

If we just remove `group`/`family`, then `This` becomes `These`, as you are referring to a selection of 20 people, out of the entire set.

Nescio0:
You're mistaken. The number of people is irrelevant, what matters is the grammatical number (singular or plural) of the entire phrase. “20 students” is a plural, but “a group of 20 students” is a singular, just as “a sum of X dollars” is singular, regardless of the value of X. The word “of” does certainly **not** turn a singular into a plural, nor does it convert a noun into something that's not a noun.

data/frontier jobs.txt
```diff
@@ -266,7 +266,7 @@ mission "Small Pirate gambling"
 				decline
 
 			label call
-			`	"You confront them on their blatant cheating when one of them yells out, We're just playin' a game here, see? You'd better shut your big mouth if you know what's good for you," the meanest-looking one growls at you. "We can make this ugly real fast."`
+			`	You confront them on their blatant cheating when one of them yells out, "We're just playin' a game here, see? You'd better shut your big mouth if you know what's good for you!" The meanest-looking one growls at you. "We can make this ugly real fast."`
```
Amazinite: Perhaps instead of just comma to exclamation point, move`The meanest-looking one growls at you.` to the front of the quote.
```suggestion
			`	You confront them on their blatant cheating when the meanest-looking one growls at you, "We're just playin' a game here, see? You'd better shut your big mouth if you know what's good for you! We can make this ugly real fast."`
```

data/remnant missions.txt
```diff
@@ -1197,7 +1197,7 @@ mission "Remnant: Expanded Horizons Quarg 1"
 			`	Almost before you are in the main hall a sales manager is at your side, inquiring as to how he can help. You can almost see the credit signs in his eyes and you suspect that scanning your license may have done a quick check of your net worth too.`
 			`	You ask a few leading questions about new designs and play the part of a captain with more money than brains, and the manager eagerly directs you to the large hangars near the end that house their largest and fanciest ships. After it becomes clear that he doesn't have any experience flying them you politely disentangle yourselves from him to look at the ships in peace.`
 			label end
-			`	After an hour of browsing, the young remnant agree to meet you back at the spaceport in an hour and take off to the outfitters.`
+			`	After an hour of browsing, the young remnants agree to meet you back at the spaceport in an hour and take off to the outfitters.`
```
Amazinite: Ref: Remnant(s) being used like human(s) and not capitalized here: as far as I'm aware, human(s) isn't capitalized as a proper noun anywhere, while all other factions are capitalized as proper nouns. Unless there are examples to the contrary, I'd think that Remnant would be capitalized here.

Janaszar: I changed it, but what I was thinking was that you are referring to them in the general sense of being different than humans, not specifically because of their faction affiliations.  "Human Faction" vs "Those humans over there".

Amazinite: Ref: #3066
`Remnants` isn't used anywhere else, so I'd say drop the `s` with the only necessary change having been to capitalize the word.

Janaszar:
I think we're getting hung up on the differences between British English and American English for collective nouns...  <https://en.wikipedia.org/wiki/Collective_noun#Agreement_in_different_forms_of_English>
I see the phrases as becoming plural due to the inclusion of `of miners`, so the use of the plural form makes sense to me, whereas `of miners` is not doing so to you guys.

Give me a few minutes and I'll change it, but I still think it's wrong.

PR #4258, opened 2019-04-16
Remnant Tech Retrieval: Wanderer Ramscoop

Zitchas:
This mission is intended to reduce the chance of people getting "lost" at the Remnant Tech Retrieval stage of the Remnant Missions by adding one such mission that triggers with the Wanderer Ramscoop. 

This outfit was chosen because it is a fairly common outfit for people in the Wanderer campaign to have. This avoids the necessity of people in said campaign to have to go back to get old human outfits in order to continue the missions. 

Note:
- This mission *can* be done alongside or in addition to the other tech retrieval missions, as it has branching dialogue to account for whether or not the player has met Taely in the past. 
- This mission depends on PR #4254 

This PR adds a single mission, and makes no other changes (provided PR #4254 is already merged).

*UPDATE:*
As a general note to everyone who may be reviewing things: I am currently in the process of following Amazinite's suggestion regarding reworking this patch of missions, so there isn't really any point to reviewing this right now. I'll probably be using parts of this, but definitely not all of it.


Amazinite:
I think there's a better alternative than adding a *fourth* mission like this. 

* Have a mission where a Remnant approaches you, asking about possible human tech. (e.g. "With our confrontations with the Korath Raiders, we are interested in obtaining technology able to exploit the high heat of Raiders, laser technology able to compete with that of the Korath, and advanced ramscoops allowing us to chase them down if they flee.")
  * Specifically human tech would be asked for as it'd be the easiest to work with.
  * Avoids needing to accidentally have the desired tech.
* This mission then doesn't complete until one of the three existing tech retrieval missions are offered.
* Conversation of the tech retrieval missions would need to change to account for the fact that there's a prior mission that leads into them.
* Have additional missions to inform the player if they've brought the "wrong" tech (e.g. the player brings beam lasers or electron beams instead of heavy lasers [beam lasers too weak and electron beams too complex], or the player bring a flamethrower instead of a plasma cannon [flamethrowers lack range and the Remnant don't want to get stranded by accidentally using too much fuel], or a normal ramscoop instead of a catalytic ["This ramscoop is weaker than ours! Is there not something more advanced?"]).
* If the Remnant that approaches you is Taely and she introduces herself in the first mission, there'd be no need for branches in each mission.

Zitchas:
OK, I think I get where you are going with that. I'm not sure how quickly I'll be able to do it, but I can work that direction.

Zitchas:
As a general note to everyone who may be reviewing things: I am currently in the process of following Amazinite's suggestion regarding reworking this patch of missions, so there isn't really any point to reviewing this right now. I'll probably be using parts of this, but definitely not all of it.

Zitchas:
As per Amazinite's suggestion, I have rewritten the Tech Retrieval system, and it is found in the new PR: https://github.com/endless-sky/endless-sky/pull/4269

As such, I will now be tweaking this Wanderer Ramscoop tech retrieval mission to be an independent mission instead of being part of the "intro to Taely and recovering stuff" set.

data/remnant missions.txt
```diff
@@ -941,7 +941,59 @@ mission "Remnant: S-970 Regenerator"
...
+	source
+		government "Remnant"
+	destination "Viminal"
+	to offer
+		has "Remnant: Technology Available: offered"
+		random < 50
+	on offer
+		require "Wanderer Ramscoop"
+		conversation
+			`Strolling into the starport, you wonder if someone from the nosy gull would be approaching you. While you have made the approach to <planet> many times, this time was differentiated by a gull that seemed to be shadowing you closely. Just looking at it, you guessed that it was probably a reconnaissance ship, or possibly a science vessel. Few other ships would carry quite so much scanning equipment.`
```
Pointedstick: This paragraph reads very awkwardly to me. Suggest rewriting.

data/remnant missions.txt
```diff
@@ -941,7 +941,59 @@ mission "Remnant: S-970 Regenerator"
...
+			`	Looking around, you quickly notice a young Remnant trotting into the area from a different landing pad. He spots you and approaches, then clears his throat nervously "Greetings, traveler" he warbles, as if unused to singing. "We scanned you as you flew by, and noticed some unusual technology mounted to your hull. Can you tell us what they are?" With this, he pulls up his comlink and shows you an image of what appear to be your ramscoops.`
```
Pointedstick: `nervously` -> `nervously.`

Pointedstick: `what they are` refers to "unusual technology", which is not a plural noun, so it should be `what it is` instead

data/remnant missions.txt
```diff
@@ -941,7 +941,59 @@ mission "Remnant: S-970 Regenerator"
...
+			choice
+				`	"Those are wanderer ramscoops."`
```
Pointedstick: Capitalize "Wanderer"

data/remnant missions.txt
```diff
@@ -941,7 +941,59 @@ mission "Remnant: S-970 Regenerator"
...
+				`	"I'd rather not share that."`
+					decline
+			
+			`	"They are?" He looks disappointed. "I guess I'll be taking Mala's next watch duty, then. She guessed that they were ramscoops, I thought they were a scanning array." He looks slightly disappointed, but quickly regains his usual impassivity. "You said 'Wanderer' ramscoops? Who are these 'Wanderers?' Are they a group like us, hidden away somewhere?"`
+			choice
+				`	"Wanderers are an alien race that live in counter-spinward from humanity."`
```
Pointedstick: `live in counter-spinward` What does this mean? Sounds awkward

data/remnant missions.txt
```diff
@@ -941,7 +941,59 @@ mission "Remnant: S-970 Regenerator"
...
+				`	"I would prefer to not talk about them."`
```
Pointedstick: `to not` -> `not to`

data/remnant missions.txt
```diff
@@ -941,7 +941,59 @@ mission "Remnant: S-970 Regenerator"
...
+					goto notalk
+			`	He listens to your explanation calmly, but his tone takes on a note of eagerness as he chants "These Wanderers sound very interesting. Thank you for telling me about them. It seems that we might have much that we could learn from them." He looks thoughtful for a moment before continuing.`
+			label notalk
+			`	"Be that as it may, the opportunity to study the fuel generation systems from someone with a different technological path would be most advantageous. If you could procure two of these 'Wanderer Ramscoops' for us, we would reimburse you for the cost, plus 750000 credits for your efforts."`
+			choice
```
Pointedstick: This choice only makes sense after the `notalk` label text, but doesn't work as well when it's immediately after the other option, since in that case the guy hasn't actually asked you to do anything specific to which you could respond `Sure, I could do that`.

data/remnant missions.txt
```diff
@@ -941,7 +941,59 @@ mission "Remnant: S-970 Regenerator"
...
+				`	"Sure, I could do that."`
+					accept
+				`	"Sorry, that's a bit too far to travel."`
+					decline
+	on complete
+		outfit "Wanderer Ramscoop" -2
+		payment 1670000
+		"remnant taely" ++
+		conversation
+			`Following the directions of the starport director, you land on the designated pad. A few minutes later a team of Remnant jog up with a flatbed transport following along behind. "Captain <first>? We were told to expect a delivery for the research labs" chants the man in the lead.`
```
Pointedstick: `labs` -> `labs,`

Zitchas:
I appreciate the reviews and comments, and I'll get around to this eventually, but right now I'm much more focused on getting the other bits of Remnant content in. 

Since I originally created this the Tech Retrieval options have expanded from three to six, and there's a fair bit more hinting at what the player should be bringing. That being said, some people don't get to the Remnant until they've started on Wanderer missions, so this should still prove useful.

Zitchas:
This mission no longer really fits with the tech retrieval missions, as those are focused on human tech, and the lead-up to those actually hints to the player that what they are looking for is actually human tech.

Thanks to all those who commented and pointed out stuff for this. I'll be re-using a fair bit of it for a later mission, since the Remnant are obviously going to want to know about this stuff - just not at that early juncture.

Issue #4260, opened 2019-04-19
Afterburner effect is offset from engine flare point at certain times


Zitchas:
**Describe the bug**
When the ship is pointed in a direction perpendicular to the path of travel, the afterburner effects are created at a point that is ahead of the engine flare point in the path of travel.

Which is to say, when the ship is rotated relative to the path of travel, the afterburner flare appears to be coming from where the engine will be in the future, instead of where the engine is now.

**To Reproduce**
1. Outfit a ship to have an afterburner
2. Go fly in space.
3. Accelerate in a straight line
4. cease acceleration.
5. rotate 90 degrees to the path of travel. (either direction works)
6. accelerate and use afterburners.
7. Notice that the afterburner effect occurs to one side of the engine flare (the side that is ahead of the ship in the direction of travel)
8. Watch as the afterburner flare slowly comes into alignment with the engine flares. This seems to correlate directly with how close the ship is to being fully aligned with the path of travel. Fully aligned to path of travel = afterburner fully aligned with engine flares.

**Expected behavior**
I expect to have afterburner effects remain perfectly centered on the engine flare. They are, after all, a modification installed directly on the engine port.

The effect trail could be expected to *trail* the engine flare if it is a smoke effect that isn't moving as fast as the engine flare itself, but in no circumstance should it be *leading* the engine flare. In other words, the default behaviour for an afterburner flare should be that its point of origin should always be the engine point unless otherwise determined by an offset in the effect definition.

**Screenshots**
https://www.youtube.com/watch?v=kQMeGOVa9Tk
This video shows a starling with a standard afterburner installed. It starts out accelerating in a straight line, then rotates 90 degrees and accelerates while using the afterburner. One can see how the afterburner leads the engine flare, with the lead distance steadily decreasing until they match after the ship has fully changed direction.

**System (please complete the following information):**
 - OS: Windows 10 Pro, version 1803 build 17134.648]
 - Game Source: Github as of 2019-APR-18
 - Version: [MCO nightly build 144](https://ci.mcofficer.me/job/endless-sky-win64/144/)

Zitchas:
I have observed a second behaviour of the afterburner flare which I suspect is probably related. The afterburner effect position shifts forwards and backwards (relative to the ship orientation) based on how fast the ship is moving relative to its maximum speed: If the ship is pointed the opposite direction from its path of travel, the effect starts behind the engine flare's start point. That is to say, it starts farther along in the direction of the path of travel from the ship.

[...]

PR #4261, opened 2019-04-21
afterburner alignment quick fix

djlad:
quick fix for issue #4260

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1424,7 +1424,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 						for(const auto &it : attributes.AfterburnerEffects())
 							for(int i = 0; i < it.second; ++i)
 								visuals.emplace_back(*it.first,
-									pos + velocity, velocity - 6. * angle.Unit(), angle);
+									pos + .5 * velocity, velocity - 6. * angle.Unit(), angle);
 					}
 			}
 		}


Zitchas:
I can't read the code well enough to know if this is a fix, but hopefully this works. We just had someone post on the discord with a video of them showing the glitch even worse while flying a flivvr.

Fzzr:
Is it possible this is caused by display scaling or whether or not you're using the high-DPI mod?

Zitchas:
I'm running on "straight off the GitHub Nightly" without any plugins (although admittedly I'm alternating between a number of different branches writing stuff. All are up to date with the Nightly.)

Fzzr:
For what it's worth, there is no afterburner effect in the high-DPI mod. I can't test right now though.

Janaszar:
I've been looking at the history of `ship.cpp`, and it seems it broke with this commit: https://github.com/endless-sky/endless-sky/commit/547e43277528bbf391c1053c761bd579a7a2cf09

I can't find any "prior commit" to that that changes that line by adding a 0.5 multiplier.  Going all the way back to when the afterburner was originally added ( https://github.com/endless-sky/endless-sky/commit/d88e716f88a2ea7c43f02a834e25783c297341b7 ), there was always a 0.5 multiplier in the position function just above what is being suggested here.

Zitchas:
Anyone know why that modifier was removed?

djlad:
I think commit [d88e716](https://github.com/endless-sky/endless-sky/commit/d88e716f88a2ea7c43f02a834e25783c297341b7) addresses a different issue. It is adding `.5 * zoom`, but this issue is solved by adding `.5 * velocity`.

I found out what was changing the afterburner `Visual` position. It was this [line](https://github.com/endless-sky/endless-sky/blob/938f68dec84d8e9f0f9bbb97299ed7c0e2c3cba4/source/BatchDrawList.cpp#L58) in the draw function for visuals:

image

Here is where the additional `.5 * velocity` was coming from. It didn't affect game state, which is why I didn't see it before. It's just drawing the effect `.5 * body.Velocity()` away from where the visual actually is.

Removing this `.5 * body.Velocity()` term fixes the issue, but it would affect other projectiles and `Visual` effects.  I really don't think this term is supposed to be here though.

PR #4266, opened 2019-04-26
Remnant Fixes

MCOfficer:
Some small fixes for the Remnant Missions and/or Outfits.

So far, I've only played the first missions, so i'd prefer that this PR doesn't get merged yet (unless a mysterious release were to knock at the door).


MCOfficer:
I don't think I'll find more in the foreseeable future, so as far as I'm concerned this is merge-ready.

PR #4267, opened 2019-04-26
Changes "airport" to "spaceport" in "Wanderers: Sestor: Scan Drones"

Zitchas:
It seems rather odd that the Wanderers would already have an airport, and doubly curious as to why the player would be meeting Sobari Tele'ek out there. It makes a lot more sense to meet at the spaceport.

Thanks to Righthand on the ES Discord for spotting this.

This PR changes airport to spaceport.


Amazinite:
Given that this appears to be the only instance of `airport` in the game, meaning it isn't simply some Wanderer name for the spaceport, I'd say this is good.

PR #4273, opened 2019-05-03
Enabling the player to demand tribute from the Remnant

Zitchas:
This PR adds tribute amounts, thresholds, and fleets to the planet definitions for Aventine, Viminal, and Caelian.

All three use the "Large Remnant" fleet, 60 in the case of Viminal and Caelian, and 80 for Aventine. The tribute given is correspondingly 3000 for the former two, and 4000 for the latter.

The logic behind these values is thus: 
- The Remnant have no external trade routes, nor significant economy of any kind, thus it seems unreasonable to expect that they would have large amounts of cash flow.
- The Remnant are strongly oriented towards "survival at all costs" against perceived existential threats.
- They are routinely attacked by a strong alien faction that they are capable of meeting on sufficiently equal terms that they can afford to "manage" the battle to encourage the aliens to get enough loot to want to come back, while destroying enough to have valuable salvage.

As a result, this sounds like a situation where demanding tribute from them should be a fairly high risk/low reward situation. (admittedly, 3k and 4k per day is still significant, but the ratio of potential profit compared to the forces one has to overcome is significantly lower)

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -24130,6 +24130,9 @@ planet Aventine
 	bribe 0
 	shipyard Remnant
 	outfitter Remnant
+	tribute 4000
+		threshold 7500
+		fleet "Large Remnant" 60
 
 planet "Bank of Blugtad"
 	attributes arach wealthy urban
@@ -24306,6 +24309,9 @@ planet Caelian
 	bribe 0
 	shipyard Remnant
 	outfitter Remnant
+	tribute 3000
+		threshold 7500
+		fleet "Large Remnant" 40
 
 planet Calda
 	attributes paradise tourism rich
@@ -27187,6 +27193,9 @@ planet Viminal
 	bribe 0
 	shipyard Remnant
 	outfitter Remnant
+	tribute 3000
+		threshold 7500
+		fleet "Large Remnant" 40
 
 planet Vinci
 	attributes paradise factory
--- a/data/map.txt
+++ b/data/map.txt
@@ -24131,8 +24131,8 @@ planet Aventine
 	shipyard Remnant
 	outfitter Remnant
 	tribute 4000
-		threshold 7500
-		fleet "Large Remnant" 60
+		threshold 8500
+		fleet "Large Remnant" 80
 
 planet "Bank of Blugtad"
 	attributes arach wealthy urban
@@ -24310,8 +24310,8 @@ planet Caelian
 	shipyard Remnant
 	outfitter Remnant
 	tribute 3000
-		threshold 7500
-		fleet "Large Remnant" 40
+		threshold 8000
+		fleet "Large Remnant" 60
 
 planet Calda
 	attributes paradise tourism rich
@@ -27194,8 +27194,8 @@ planet Viminal
 	shipyard Remnant
 	outfitter Remnant
 	tribute 3000
-		threshold 7500
-		fleet "Large Remnant" 40
+		threshold 8000
+		fleet "Large Remnant" 60
 
 planet Vinci
 	attributes paradise factory


Zitchas:
Sure, I can definitely do that. Would it be possible to get that PR turned into an entry in the wiki instead? I went looking for a wiki page first, and couldn't find anything talking about tribute at all.

data/remnant.txt
```diff
@@ -105,6 +105,60 @@ fleet "Remnant Transport"
...
+		"Starling (Sniper)" 10
+	variant 10
+		"Starling (Heavy)" 5
+		"Starling (Hunter)" 5
+
+fleet "Heavy Remnant Defense"
+	government "Remnant"
+	names "remnant"
+	personality
+		staying heroic marked unconstrained
```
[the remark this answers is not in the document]
Zitchas: Thanks for the answer!

Zitchas:
@tehhowch The numbers are rather up in the air at this point, but this now adds an additional variant albatross and three fleets that are used in Remnant Defense. Probably overkill compared to most systems one can demand tribute from, but should be considerably more interesting than the old "one fleet" system.

Zitchas:
Just a note on the fleets: The defence fleets all have `staying heroic marked unconstrained`. While this is technically redundant since all defence fleets are assigned that personality, these fleets have been assigned it anyway so that these rather large fleets can be used as specific "hunter/killer" fleets in future missions.

Amazinite:
I'm not sure if it's wise to expand the number of planets that the player can demand tribute from if we have such an issue with the existing planets.

tehhowch:
No stock fleet in the game should use `marked`. If you want to use these fleets as part of a mission, you will be defining the desired personality there.

Zitchas:
OK, I'll remove the marked.

Zitchas:
OK, freshly updated and ready to go.

While tributing isn't a major gameplay element for most, I think we should be consistent about enabling it wherever possible so that players interested in the meta-challenge of "Dominate everything" can do so.

That being said, it occurred to me that tributing serves a second less obvious function: It gives players an easily trigerable way to challenge themselves against large fleets. Currently, those fleets are limited to human ones - This PR gives players the option of testing themselves against a Remnant fleet that will provide a much different challenge than the standard ones they currently have access to.

While I don't seek out large scale fleet combat much, I can see the appeal in being able to do so, and feel that this is a good additional reason to implement tributing universally. It might not typically add to the story, but it *does* add the opportunity for players to challenge themselves in new ways, which is a good thing.

data/remnant ships.txt
```diff
@@ -173,6 +173,38 @@ ship "Albatross" "Albatross (Heavy)"
...
+		"Forge-Class Steering"
+		"Hyperdrive"
+
+	gun 
+	gun 
+	gun 
+	gun 
+	gun 
+	gun 
+	gun
```
Amazinite: This isn't necessary.

data/remnant.txt
```diff
@@ -105,6 +105,71 @@ fleet "Remnant Transport"
...
+		"Starling (Heavy)" 5
+		"Starling (Hunter)" 5
+
+fleet "Heavy Remnant Defense"
+	government "Remnant"
+	names "remnant"
+	personality
+		staying heroic unconstrained
+	variant 6
+		"Albatross (Shield)" 1
```
Amazinite: Nitpicking, but if all you want is one ship then you don't even need to include a number.

data/remnant ships.txt
Zitchas: OK. Fixed.

data/remnant.txt
Zitchas: Ok. Removed several instances of ` 1`.

Issue #3361, opened 2017-12-04
Mission failure by ship loss is very hard to spot


bluss:
Imagine you are in a big space battle. Hundreds of ships, thousands of projectiles :boom: :smile:. During all that, it may happen that a mission failure is announced, without sound, in white text at the bottom of the screen. You are going to miss that and later wonder huh, why did the mission fail?

It would be great to improve the visibility of the mission failure message from losing ships; and it should be available after after the fact too.

It's really an awesome game, so thanks a lot for making it.

jafdy:
What about having the mission name in red (like disabled ships in the I menu) until the next takeoff.

tehhowch:
My suggestion from https://github.com/endless-sky/endless-sky/issues/3354#issuecomment-348837136
>Regarding the mission failure info getting "drowned out" - please open as a separate issue and request something like an "Alert panel" in which "important" messages such as mission failed, ship captured, ship destroyed, etc are kept, and possibly viewable from the Player Info Panel, much like the logbook can be viewed.

Coloring the message could be done by extending the Messages class to have a flag (e.g. `isHighPriority`) that can be referenced in the message-drawing code when setting the color (which right now is just an alpha variation based on message age).
I would say coloring is a step in the right direction, with the end goal of what I mentioned in my comment (which requires sprite / UI updates).

Laudrin:
Not only colouring the messages, I would further suggest specific **sounds** for events like ship is attacked, disabled or lost, mission failed etc.

Zitchas:
Yes, this would be very helpful. Personally, I'd also like it to stick around a bit longer, so when the text fades it, it just leaves the mission failed notification floating there for another minute or two.

temtemy:
I think all mission failures should be in dialogs instead (except the invisible "missions"), just like the FW missions. They definitely won't be missed out.

Pointedstick:
The main problem here is that the failure message disappears. Even if it were rendered in flashing red text and sounded a warning klaxon, it would still disappear in a matter of seconds in a large battle.

Very important messages like these should probably be persistent somehow. Using dialogs for all of them seems quite reasonable to me.

Zireael07:
redirecting from a duplicate:

Most games with many messages (RPGs, roguelikes) have a 'message/log history' screen where you can scroll back at your leisure. This often, but not always overlaps with storing the log in a file, as mentioned in #4794.

7eggert:
[...]

PR #4274, opened 2019-05-04
Add message if an escort is destroyed

SomeDer:
It can be hard to notice if an escort has been destroyed, but it is an important event which could make restarting the game the best option. Adding a message when an escort is destroyed could make it easier.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1742,14 +1742,14 @@ void Engine::DoCollisions(Projectile &projectile)
 				if(isSafe && projectile.Target() != ship && !gov->IsEnemy(ship->GetGovernment()))
 					continue;
 				
-				int eventType = ship->TakeDamage(projectile, ship != hit.get());
+				int eventType = ship->TakeDamage(player, projectile, ship != hit.get());
 				if(eventType)
 					eventQueue.emplace_back(gov, ship->shared_from_this(), eventType);
 			}
 		}
 		else if(hit)
 		{
-			int eventType = hit->TakeDamage(projectile);
+			int eventType = hit->TakeDamage(player, projectile);
 			if(eventType)
 				eventQueue.emplace_back(gov, hit, eventType);
 		}
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1742,14 +1742,14 @@ void Engine::DoCollisions(Projectile &projectile)
 				if(isSafe && projectile.Target() != ship && !gov->IsEnemy(ship->GetGovernment()))
 					continue;
 				
-				int eventType = ship->TakeDamage(player, projectile, ship != hit.get());
+				int eventType = ship->TakeDamage(projectile, ship != hit.get());
 				if(eventType)
 					eventQueue.emplace_back(gov, ship->shared_from_this(), eventType);
 			}
 		}
 		else if(hit)
 		{
-			int eventType = hit->TakeDamage(player, projectile);
+			int eventType = hit->TakeDamage(projectile);
 			if(eventType)
 				eventQueue.emplace_back(gov, hit, eventType);
 		}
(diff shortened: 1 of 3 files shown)


Zitchas:
Just to clarify (since I don't have compiling set up), is this a message that will show up in the bottom left log/message area?

SomeDer:
Yes, it appears in the bottom left area

SomeDer:
> Just to clarify (since I don't have compiling set up), is this a message that will show up in the bottom left log/message area?

Yes

hcatlin:
I'd love to have this feature... as it often throws me off. I kind of land in the next sector and see "okay, who made it?"

jostephd:
Maybe the message should be shown upon landing? Something like "You have lost 5 escorts, as a result, 2 story missions were aborted: <list of names of those missions>". This would only count story missions, not job board missions. (Should I open a separate issue for that? I don't want to hijack this.)

Nescio0:
At most three messages can be displayed simultaneously, right? So if more escorts are destroyed, you might miss the notification.

Pointedstick:
+1 to the idea of showing a notification for this. It may not be _immediately_ actionable, but it's actionable within a short period of time. It lets you know you'll need to re-buy it, and helps you understand where an escort went if it's disappeared.

However I'm generally concerned with putting too many messages in the message area on the bottom of the screen. The UX isn't great right now because the messages are really easy to miss, and this becomes worse the more messages are added because older ones time out even faster. Also, the message area intermixes messages that are ignorable (hails and flavor text) with messages that are critically important (you just failed a mission).

I think we generally need to overhaul the notification UX before we add a bunch of new notifications.

Amazinite:
Ref: #3361 in regards to the bottom left messages.

endless-sky:
I think the real question is whether a message should be shown for *every* ship that is lost, or only when a ship is lost that causes a mission to fail (the current behavior). If a message is shown for every ship, that would make the failed mission messages disappear more quickly. And as others have pointed out, the fleet info dialog already gives you an easy way to check if any escort has been destroyed.

SomeDer:
As mentioned by @jostephd, maybe important messages should be shown upon landing as well, so that they can be acted upon immediately if noticed, but can still be read even if missed. Also, maybe they should appear in red, to make them more noticeable.

>  that would make the failed mission messages disappear more quickly

I think the failed mission messages are harder to miss, because they are longer, so they are more noticeable. Also, the lost escort messages appear earlier than the failed mission messages, so the failed mission messages can only be pushed up by other escorts being destroyed. If so many escorts are destroyed, then the player will probably either want to restart anyway, or is so rich that wouldn't care about losing either ships or missions.

>  the fleet info dialog already gives you an easy way to check if any escort has been destroyed.

I didn't notice this. Maybe this should be somehow documented. Also, perhaps it makes this proposition obsolete, as it provides another way to notice if an escort is destroyed (though perhaps a slightly more difficult one).

jostephd:
> I didn't notice this. Maybe this should be somehow documented. Also, perhaps it makes this proposition obsolete, as it provides another way to notice if an escort is destroyed (though perhaps a slightly more difficult one).

Don't forget that this only works before you land. After you land, the info screen gives no indication of whether an escort had been destroyed before landing. It does show escorts that had been disabled, but again, an on-landing dialog saying that would be appreciated.

jafdy:
What about a on-landing dialog that tells you the ships you have lost since you last landed. Probably sorted by value giving name and hull (and possibly value).

jostephd:
@jafdy How about making that change in 0.9.11 (the next unstable release) and seeing how it's received by the wider community?

jafdy:
I don't make those decisions, I am simply offering a solution that may work.

SomeDer:
I'll close this for now, as it doesn't seem to be getting anywhere. If you need to check whether an escort has been destroyed, use the fleet info dialog.

PR #4282, opened 2019-05-15
fw epilogue edit

comnom:
When testing the Paradise Fortune mission chain, FW Epilogue: Alondo offered, leading to a situation where I sat down and had a beer with Alondo while the Republic was hot on my heels.

The mission Paradise Fortune 4 has a branch that I assume was intended to handle this, but it does nothing to stop the mission from being offered (since it cant), only changes the beginning dialog.

This PR handles this oddity by removing the ineffective branch from Paradise Fortune 4, and marks the all the epilogue missions minor to avoid the need for any special dialog for this case or any other.

### DIFF
--- a/data/free worlds epilogue.txt
+++ b/data/free worlds epilogue.txt
@@ -9,6 +9,7 @@
 # PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 mission "FW Epilogue: Ijs and Katya"
+	minor
 	landing
 	source "Winter"
 	to offer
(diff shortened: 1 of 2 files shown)


jafdy:
I don't see why you need to remove the branch, presumably it will trigger correctly if you have already done the epilogue and go through the other section otherwise, if you have yet to do the epilogue.

Amazinite:
The epilogue should have been made minor, but the branch has nothing to do with that.

comnom:
I've reverted it, but under protest. It uses the exact same wording as the epilogue and really doesn't seem to have a place there.

PR #4283, opened 2019-05-19
feat: Placement of mission cargo/passengers in a player's fleet

petervdmeer:
Something that bothered me a lot while playing Endless Sky was that my strong transport ships were empty while some of my weak front-line fighter ships ended up carrying mission critical cargo. Losing a front-line fighter then meant losing mission(s).

This commit adds a certificate that controls which ships carry mission cargo, passengers and outfits. Fleet captains can use this certificate to avoid carrying critical cargo on front-line battleships.

The loading mechanism introduced in this commit does not conflict with the plans that I read about to differentiate between regular bunks, luxury accommodations and jail cells.


Zitchas:
I haven't looked at your actual content yet, but your description sounds good. Having a way to control which ships mission cargo/passengers get assigned to would be great. (and having that method be something in-game is even better)

beccabunny:
Personally i really like that feature, but i am not happy of the role you gave to the outfit in its description. It sounds like: "You can't transport any passenger nor cargo if you don't carry this with you". How about something more like:
"This special equipment automatically sends to the local spaceport a signal that marks the ships it is installed in as the preferred location for the eventual transported goods and people. This way, it allows fleet captains to reduce the risk of losing passengers or valuable cargo putting these in their strongest or most-protected transport ships." and so changing this to a proper outfit?
In both cases, you need a new image in order to make this accepted into the game, and if you need i could make a proper one for this PR

petervdmeer:
@Zitchas: Thanks. I was considering to introduce the new outfit to control cargo-loading as part of a mission, but if I add a mission for that, then I will do that in a separate pull request after the mechanism itself is merged into master.

petervdmeer:
@beccabunny : Good idea. I changed the outfit to "Loading Priority Marker" (and updated the description) to match your suggestion. I'm not very good at graphics for Endless Sky. Can you create an appropriate image and add it?

beccabunny:
Sure i will

data/outfits.txt
```diff
@@ -397,6 +397,12 @@ outfit "Local Map"
 	"map" 12
 	description "This data chip contains complete information on the twelve star systems closest to this one: planets, ports, governments, trade prices, available services, etc. You can get all the same information just by exploring those systems yourself, but having a map can save you from making wrong turns if you are trying to travel through new territory to reach a certain system quickly."
 
+outfit "Loading Priority Marker"
+	category "Special"
+	cost 500
+	thumbnail "outfit/license"
+	"preferred transporter" 1
+	description "This transponder signals that the ship carrying it is a preferred ship for transporting passengers and cargo. Spaceport crews will board passengers and load customer cargo into preferred ships before any other ships. Priority markers allow fleet captains to reduce the risk of losing passengers or valuable cargo by only marking their strongest or most-protected ships as preferred transport ships."
```
beccabunny: I think that you used "ship" too many times in the description. How about:
"This transponder signals that the vessel it is installed in is the preference for transporting passengers and cargo. Spaceport crews will board passengers and load customer cargo into such ships before the other ones that belong to the same owner . Priority markers allow fleet captains to reduce the risk of losing passengers or valuable cargo by only marking their strongest or most-protected ships as preferred transports."

petervdmeer: Agreed. Used your text instead.

beccabunny:
I made this as the outfit image for the transponder, it isn't scaled yet
(way too big even for 2x version)

Il mar 21 mag 2019, 22:10 Peter van der Meer <notifications@github.com> ha
scritto:

data/outfits.txt
```diff
@@ -397,6 +397,12 @@ outfit "Local Map"
 	"map" 12
 	description "This data chip contains complete information on the twelve star systems closest to this one: planets, ports, governments, trade prices, available services, etc. You can get all the same information just by exploring those systems yourself, but having a map can save you from making wrong turns if you are trying to travel through new territory to reach a certain system quickly."
 
+outfit "Loading Priority Marker"
+	category "Special"
+	cost 500
+	thumbnail "outfit/license"
+	"preferred transporter" 1
+	description "This transponder signals that the vessel it is installed in is preferred for transporting passengers and cargo. Spaceport crews will board passengers and load cargo into such ships before the other ones that belong to the same owner. Priority markers allow fleet captains to reduce the risk of losing passengers or valuable cargo by only marking their strongest or most-protected ships as preferred transports."
```
Pointedstick: `the vessel it is installed in is preferred` -> `the vessel on which it is installed is preferred`

beccabunny:
transponder@2x
transponder
@petervdmeer I made this as the image, both normal DPI and high DPI. If you like it, in the copyright file it is enough to add it under the others that belong to Becca Tommaso

Pointedstick:
Very nice image, @beccabunny!

In principle this feels like the kind of thing that there should be a configuration-based user interface for rather than an outfit-based UI. However I rather like the way you've implemented this and I don't think it's a problem. I kind of like the way you need to buy this, in fact. It feels like the old Auto-Refueler in the EV games and serves as a way of actually exposing the feature to the player in a way they're going to see it, which is probably more visible than having it buried in the Ship panel somewhere.

So +1 initially. Need to test and perform code review (though on initial skim it looks fine). Please do use that nice new art asset if you like it too.

petervdmeer:
@beccabunny Very nice image indeed. Really looks like a piece of equipment for an Endless Sky spaceship.
But I'm thinking of the loading priority marker as a relatively cheap (500 credits) and small piece of equipment that receives a signal from the local spaceport-docks, then does some minor processing and then transmits a signal back.
The box in the middle is fine as computer/signal processor, but is it possible to make the signal sending and receiving more explicit by adding a stick-antennna (or two stick-antennas) somewhere?

The equipment also looks somewhat large (long-range transponders used in airplanes have the size of a car-radio, while short-range transponders as used in car tollway passes are about the size of a smartphone), but we are talking about a transponder in a huge spaceship that transports tons of cargo, so the equipment looking somewhat large should be okay.

petervdmeer:
@Pointedstick Just uploaded a commit with a small bugfix. I noticed that buying Outfits into cargo was showing the total cargo-space instead of the preferred cargo space. This is now fixed.

tehhowch:
The current UI *is* a configuration-based interface - things load into your escorts in the order they appear in the player info panel.

beccabunny:
transponder@2x
transponder
@petervdmeer Modified a bit and added the antenna
[transponder.zip](https://github.com/endless-sky/endless-sky/files/3212482/transponder.zip)
New files

petervdmeer:
@beccabunny Thanks, the antenna looks really industrial, durable. Very suitable for a spaceship.

I added your image and copyrights and I will do a separate pull-request for the high-DPI plugin.

beccabunny:
Great!

petervdmeer:
@tehhowch Good remark. I never considered the order of the ships in the player info panel to be an interface for loading passengers and cargo.
Is this already documented somewhere in the game or game manual? And should I change anything in this pull-request based on this interface?

Pointedstick:
> The current UI _is_ a configuration-based interface - things load into your escorts in the order they appear in the player info panel.

Unfortunately, it illustrates the pitfalls of such a user interface: it's 100% non-discoverable unless you have to be a developer. :)

endless-sky:
If I were a first time player with only a single ship, seeing this outfit for sale would be really confusing to me. Wouldn't it make more sense just to document the current behavior somewhere in the player's manual?

The trade-off here is between:
1. the few players who will find this outfit useful, and who don't want to control cargo by reordering their ships, vs.
2. the (probably many more) players who will just find this outfit confusing or unnecessary.

In my opinion, that's not worth it. Remember that even if you're a "power user" the game mechanics need to be designed with regular users in mind.

petervdmeer:
@endless-sky Documenting the current behavior in the player's manual is a good idea.

I have personally been using this plugin https://github.com/petervdmeer/endless-sky-cargo-controls-plugin to really avoid loading cargo in my fighters and front-line battleships. But that plugin has the downside that you buy "Bunk removal" and buying an outfit that removes something feels a bit counter-intuitive. The outfit in this pull-request made more sense to me (you buy something that actually does/controls something.)

I also considered new first time players; In a future pull-request I wanted to introduce the new outfit from this pull request with a specific unique passenger/cargo-mission (which would only be available to players with multiple transport ships). This way new first time players would not see it, while players that get a larger fleet would encounter it at a moment where the mission could properly introduce the new outfit and where the player might actually be interested in it.

There is a 3rd option for the trade-off: If somebody creates a pull-request that controls cargo loading through the player-info-screen (having options of "showing crew/parked/no-cargo" instead of the current "showing crew/parked", would you then be interested in that?

tehhowch:
With regards to customizing the player info panel, there are numerous functionality requests (such as sorting by name or type, grouping by parent / squadron) that would necessitate decoupling the order of ships from their display. A logical extension is to then use one of the display-orderings to control cargo and passenger loading, i.e. explicitly provide an interface with UI elements that convey the loading order functionality. This is my preferred solution.

petervdmeer:
@tehhowch This pull request is not only about passenger/cargo loading order. It is also about explicitly not loading passengers and cargo on some weak ships.
The screens that show how much cargo-space is available shows the space for the full fleet, including ships that I consider "too weak". As a captain I would not be interrested in the total cargo-space in such screens, but only in the cargo space that I want to use.

During implementation of this pull request I also noticed something else: If I have a boxwing (first in list, 80 tons cargo-space, filled with 10 tons mission cargo) and later 20 empty starbarges (with 50 cargo space each), then I still cannot carry a reactor of 73 tons as cargo (since outfits appear to be loaded after mission cargo, "undivided" and are sold if they are left). Just controlling loading order would not be sufficient to handle this case.

If somebody improves cargo loading, would you then also want to have this type of (corner) case nicely handled?
And how much flexibility can we allow ourselves here? Would you be okay with "overloading" one of the starbarges in this case (as long as the total cargo carried is not above local/departing fleet capacity)?

My preference for this type of corner-case is that we make it fit (by overloading a star barge if needed) or if that is not possible/allowed, that we then go for a solution where we give an error-message while buying outfits into cargo-space (instead of automatic selling at departure).

Pointedstick:
Documentation is not a solution for a poor UI, because in general users don't read documentation. If the UI is not as self-documenting as possible, then it needs improvement.

I think possible confusion regarding what the outfit does for players with only a single ship could be rectified fairly easily with changes to its description that make clear that it only has an effect when you have a fleet.

comnom:
>This pull request is not only about passenger/cargo loading order. It is also about explicitly not loading passengers and cargo on some weak ships.

Seems to me this may be better as an issue for now, as there's a bit of disagreement on features/implementation. It should always be an issue first, despite more frequent contributers ignoring that rule, so that things can be settled before you put more work in.

So far an issue should try to reconcile two problems before a PR:

* Undocumented/obtuse cargo/passenger load order
* The ability to control how much cargo/passenger space is available for missions

Amazinite:
I see that no one has referenced #2746 yet.
The heart of the issue of this PR and of #2746 are in part the same; frustration with failing a mission because cargo/passengers were on an escort. #2746 just also includes the problem of not being able to complete a mission for the same reason. So instead of a brand new issue, perhaps move discussion on the situation there, and I could rename it to something more general like "Placement of mission cargo/passengers in a player's fleet."

petervdmeer:
Closing: I personally also feel that a "load cargo in this ship/don't load cargo in this ship" checkbox (similar to parking ships) would on the long run be more intuitive for players than the specific outfit, so closing this PR to make space for a new PR that does just that.

Issue #4286, opened 2019-05-20
Feature Request: Crew Numbers as Condition


VonZorch:
Have the flagships required crew and current crew exposed to condition statements in missions.
This implies being able to modify current crew inside missions too.

Zitchas:
Definitely a good idea. As a mission writer, this would be a very useful variable to have for branching dialogue. Up until now it has basically been verbotten (forbidden) to reference the player's crew in descriptive text, because there is no way to know if they are flying a ship with 1 crew or a ship with a crew in the hundreds. This will probably still be the norm, but it would open up a lot of potential descriptions, scenes, and descriptive variety.

Fzzr:
I would say that checking flagship crew and modifying it from within a mission are distinct features. There's no upfront cost to hiring crew, so it's not a reward. All that can do is involuntarily change the player's recurring costs or damage their ship's handling without interactive combat. The only reason I can think of to modify player crew would be like the current firing of crew to fit passengers, rather than doing it manually during a mission. Checking for mission offers or conversation conditions makes sense, though.

Fzzr:
See https://github.com/endless-sky/endless-sky/pull/4294 for a quick implementation of just the condition check part.

tehhowch:
As of 7fe1e52 there are read-only conditions for the flagship crew. Allowing mission conversations or actions to explicitly modify the crew complement is not a priority.

PR #4294, opened 2019-05-26
Add flagship crew info to conditions. Partially implements #4286

Fzzr:
This allows checking crew-related conditions of the player flagship. Attached is a file to test.

This allows directly checking for `"crew"`, `"required crew"`, and `"bunks"` in the same way that `"cargo space"` and '"passenger space"` work now.

[test flagship conditions.txt](https://github.com/endless-sky/endless-sky/files/3220562/test.flagship.conditions.txt)

Note: recommend not merging until after the next stable release.

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2256,6 +2256,11 @@ void PlayerInfo::UpdateAutoConditions(bool isBoarding)
 		conditions["passenger space"] = flagship->Cargo().BunksFree();
 	}
 	
+	// Store conditions for flagship current crew, required crew, and bunks.
+	conditions["crew"] = flagship->Crew();
+	conditions["required crew"] = flagship->RequiredCrew();
+	conditions["bunks"] = flagship->Attributes().Get("bunks");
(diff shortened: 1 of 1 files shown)


Fzzr:
Documentation is at my fork of the wiki under the `flagship-crew-conditions` branch.

Fzzr:
Updated documentation to note that it takes effect in v.0.9.11.

tehhowch:
Perhaps it would be best as a follow-up to this, but I had previously considered exposing a datafile keyword that introduces a description of what conditions get computed, e.g.
```
<keyword>
  <class specificier>
    <method of class>...
    ...
  attribute
    <condition name><attribute to query> <location to query>
    ...
```
e.g.
```
autoconditions
  account
    credits, net worth
  attributes
    crew, outfit space, bunks
```
It would need a fair bit more work but could help avoid needing to modify source code to add things like this for missions or plugins.

Zitchas:
Awesome! I'm glad to see this merge. More attributes to work with!

PR #4295, opened 2019-05-26
Adds Beta-version text to Main Menu

Janaszar:
Adds a block of text on the main menu, just below the compass rose, mentioning that the version being played is a Beta version, and to submit issues to the community on the GitHub, Steam Forums, or Discord (including the Discord ending link, in case people need it).

I feel that this can help direct people playing the beta into reporting their issues so that we can fix them quickly for the Stable release.

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -144,8 +144,8 @@ interface "main menu"
 		center -420 155
 		dimensions 120 30
 
-	# For Beta Only
-	label "THIS IS THE BETA VERSION"
+	# For Unstable Only
+	label "THIS IS THE UNSTABLE VERSION"
 		center 0 100
 	label "PLEASE REPORT ALL ISSUES TO"
 		center 0 120
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -144,6 +144,18 @@ interface "main menu"
 		center -420 155
 		dimensions 120 30
 
+	# For Beta Only
+	label "THIS IS THE BETA VERSION"
+		center 0 100
+	label "PLEASE REPORT ALL ISSUES TO"
+		center 0 120
+	label "US ON GITHUB, STEAM FORUMS,"
+		center 0 140
+	label "OR DISCORD (/ZeuASSx)"
+		center 0 160
+	label "Thanks for playing!"
+		center 0 180
+
 
 
 interface "menu player info"


MCOfficer:
i like the idea, but are we supposed to remove this for every commit that marks a stable version? yikes.

Fzzr:
> i like the idea, but are we supposed to remove this for every commit that marks a stable version? yikes.

It's just a revert if it comes to that. Not ideal, but some way to get feedback fast would be a big help, I think. It also manages expectations that it's a non-stable version.

MCOfficer:
for that matter, we might as well maintain a stable branch, which doesn't has this commit (it's reverted there). upon release, merge master into stable, so they are even except for this commit.

Fzzr:
That's even better, and it allows for hotfixes to be added without new unstable features making into the released versions. Highly agree, but that's totally up to the maintainers.

Janaszar:
Well, we don't need to remove it, just comment it out.  Even then, it's so obvious, and releases take at least a few months between each other, it's relatively minor.  Think of it like setting the new Version number.

data/interfaces.txt
```diff
@@ -144,6 +144,18 @@ interface "main menu"
 		center -420 155
 		dimensions 120 30
 
+	# For Beta Only
+	label "THIS IS THE BETA VERSION"
```
Fzzr: Maybe make this "AN UNSTABLE VERSION" rather than "BETA VERSION" so as not to imply that the stable release is non-beta. It still is, since 1.0 is not out.

Janaszar: Just changed it over to Unstable.  Good catch; was thinking of the "Beta" branch on Steam I guess.

Amazinite:
If the purpose is to encourage bug reports, then perhaps just expand the general contribution notice to include the names of the places that bugs can be reported, and then place said notice near the top of the credits file just under the "Welcome to Endless Sky! / version X" part so that it's easier to notice instead of requiring someone to wait until the very end of the credits. That would certainly be less of a hassle than repeatedly adding and readding this notice with each release, and it would look better as well.

Amazinite:
So the following in credits
```
This game is open source.
To contribute to story writing,
artwork, or programming, visit:
https://endless-sky.github.io
```
would be moved to the top and expanded to something like
```
This game is open source.
To report bugs, give feedback,
or contribute to story writing,
artwork, or programming, visit:
https://endless-sky.github.io
The game's Steam forums
Discord (/<link>)
```

Amazinite:
I've implemented my idea in 044fd8b. Does that achieve what you intended?

PR #4298, opened 2019-05-28
Explicitly state the hit force of the Hai Tracker as being a feature

Amazinite:
Ref: [Feedback](https://steamcommunity.com/app/404410/discussions/0/1653297026046844771/#c1653297026047413327)

Explicitly state that the hit force of the Hai Trackers is a feature that compensates for their relatively low damage compared to human missiles.

<details><summary>Old description</summary>The Hai Tracker applies among the most hit force per shot and per second of the more agile missiles (i.e. Meteor, Sidewinder, Tracker, Piercer, Thunderhead). Given the volume with which Hai ships deploy Trackers, dealing with them can be difficult when they're constantly tossing you around.

This PR greatly reduced the hit force of the Tracker from 350 to 200, bringing its hit force more in line with something like the Meteor Missile, which actually does more damage per shot.</details>

### DIFF
--- a/data/hai outfits.txt
+++ b/data/hai outfits.txt
@@ -231,7 +231,7 @@ outfit "Hai Tracker Pod"
 		"infrared tracking" .4
 		"shield damage" 200
 		"hull damage" 160
-		"hit force" 350
+		"hit force" 200
 		"missile strength" 16
 	description "Trackers are fast, powerful, and accurate homing weapons. Their only weakness is their large turning radius: if a Tracker misses its target, it takes a long time to turn around."
 
--- a/data/hai outfits.txt
+++ b/data/hai outfits.txt
@@ -231,9 +231,9 @@ outfit "Hai Tracker Pod"
 		"infrared tracking" .4
 		"shield damage" 200
 		"hull damage" 160
-		"hit force" 200
+		"hit force" 350
 		"missile strength" 16
-	description "Trackers are fast, powerful, and accurate homing weapons. Their only weakness is their large turning radius: if a Tracker misses its target, it takes a long time to turn around."
+	description "Trackers are fast and accurate homing weapons. Although not as powerful as most human missiles, Trackers boast a significant hit force, helping to prevent the escape of whoever is unfortunate enough to be on the receiving end. Their only weakness is their large turning radius: if a Tracker misses its target, it takes a long time to turn around."
 
 effect "tracker fire"
 	sprite "effect/tracker fire"


ZBok:
I really enjoy the large pushback of Trackers. It gives a nice characteristic feel to Hai space, helps justify the constant reloading effort (going back to Hai space) if you're using them yourself, and introduces nice counterplay tactics when fighting against them. More specifically it indirectly buffs the use of AM turrets on the player's own ship and escorts, as they are very helpful in preventing any one ship from being quickly isolated and destroyed by several Trackers focusing them down.

I would humbly suggest the nerfbat was applied a little more gently, just my zero point zero two dollars.

Amazinite:
How about 250 hit force? That puts its per shot above the Meteor, while keeping its per second above all the human missiles, and tying it with the per second of the Korath Piercer.

Here's a table of the current numbers, sorted by hit force / second. Among these, the Tracker is probably deployed in the most bulk potentially against the player, which is where I think the feedback on Steam is coming from. At 250 or even 200, the player would still be knocked around by Trackers more than any other missile they go up against given that volume of fire.

Weapon | Hit Force / Shot | Hit Force / Second
----|----|-----
Thunderhead Launcher | 750 | 600
Finisher Pod | 750 | 562.5
Hai Tracker | 350 | 350
Korath Piercer Launcher | 300 | 250
Typhoon Launcher | 450 | 225
Javelin Pod | 50 | 200
Sidewinder Launcher | 170 | 170
Meteor Launcher | 220 | 165
Heavy Rocket | 600 | 150
Torpedo Launcher | 300 | 120

ZBok:
For sure, Trackers would definitely be the type of missile most players will see the most of, and in the highest concentrations.

I think it's also relevant to note that in terms of damage per second per outfit space, Hai trackers are *worse* vs both shields and hull than every single human weapon listed, by a pretty big margin in some cases (unless I am messing up real badly and if so I apologise). I know you can't just straight compare every missile system by just DPS, but Hai trackers aren't a particularly powerful weapon anyway, even compared to human tech.

I am of the opinion that the impact force is in a great place at the moment and helps characterise the the weapon and the Hai in general, and as a player that uses AM turrets regularly and doesn't enjoy using huge fleets, I don't find them oppressive. Keen to hear others opinions on this!

Amazinite:
>Hai trackers are worse vs both shields and hull than every single human weapon listed... Hai trackers aren't a particularly powerful weapon anyway, even compared to human tech.

Hence why they get deployed in bulk. But like I said, even with a substantial reduction in hit force, they're still going to be able to push ships around a lot simply because of that bulk deployment. This is helped even further by their high range, velocity, and homing compared to other missiles. If the player is using Trackers to push ships around, I'd expect them to be using them in bulk as well in order to get their full potential. 

>I am of the opinion that the impact force is in a great place at the moment and helps characterise the the weapon and the Hai in general, and as a player that uses AM turrets regularly and doesn't enjoy using huge fleets, I don't find them oppressive. Keen to hear others opinions on this!

I opened this PR so quickly after the feedback because it's an easy change, but I will wait on if we get more opinions here or if we receive feedback from others about the Trackers.

Pointedstick:
I also kind of like the high hit force. If it's doing to be toned down a bit, maybe 275 or even 300 would be more reasonable.

Alternatively we could update the description to explicitly mention and explain the unusually hard hit effect. Maybe it's deliberate, to prevent ships from being able to easily escape into hyperspace.

Amazinite:
To note, the Torpedo only does 300 hit force per shot. So at least reducing the Tracker to 300 might make sense.

ZBok:
I set up a cloak ship in Wah Ki and observed the eternal Hai vs Unfettered tickling contest for a while, just to get a feel for the 350 hit force, then started reducing hit force and seeing how it changes things.

Dropping hit force to 300 is super noticeable. The battlefield doesn't spread out as quickly or as far and ships being focused down don't quite get to the same high velocity under constant barrage, but it does seem to maintain that characteristic feel of "Wow, Hai space is... jolty" which I feel is good. I feel that it's fine at 350, but if nerf happens I would suggest that dropping to 300 would likely be enough or more than enough. Dropping to 250 seems to really remove that characteristic feel. Worth having a look for yourselves tho, Wah Ki seems a nice place to get solid insight on this.

MCOfficer:
> I also kind of like the high hit force. If it's going to be toned down a bit, maybe 275 or even 300 would be more reasonable.
> 
> Alternatively we could update the description to explicitly mention and explain the unusually hard hit effect. Maybe it's deliberate, to prevent ships from being able to easily escape into hyperspace.

+1 on both. One could also drop it to something like 275  and give the damage a one-pass nerf in return, so they don't feel too overwhelming.

Amazinite:
How's that sound for a change?

ZBok:
Love it.

Issue #3877, opened 2018-08-01
Round all numbers in the ship stats readout


DingusShingleton:
Currently my ship has a max speed of 501.05 and a idle heat rating of 293.2. In the interests of further streamlining the ever-controversial stat block, I suggest simply rounding all these numbers to the nearest integer. A fraction of a heat or energy or anything else is utterly meaningless, so leaving those digits in only adds noise to an already crowded bunch of data.

DingusShingleton:
Furthermore, and this might be slightly more controversial, but if the typeface used in the stat block had numbers that were all equally wide, the digit columns would line up perfectly and it would make mentally adding up the totals easier. 

Right now if you have let's say moving heat of 111 and firing heat of 600, the former is about half the width of the latter. This means that the hundreds digit of 111 roughly lines up with the tens digit of 600. Does anyone else find this mildly annoying or is it just me? I'm already doing enough math in my head without having to mentally account for kerning!

And this was a simple example to illustrate the point, there are 4 or 5 numbers that need to be added up sometimes.

Bladewood:
I was never able to reproduce this after it happened just the one time (a few weeks ago), but on the topic of rounding:
![coolingfail](https://user-images.githubusercontent.com/16506683/43603897-4f8893b6-9662-11e8-9ec0-26eaefe55054.png)

DingusShingleton:
Not to mention your turning. What would you do if you didn't know you had those extra three thousandths of a degree per second

Amazinite:
I’d personally like to see rounding to the nearest tenth instead of integer, but either would help to reduce stat clutter.

Nescio0:
* font: a different (monospace?) font would be helpful too; the hyphen-minus and decimal dot are very thin
* heat: rounding to the nearest integer makes sense
* energy: several weapons and shield generators have decimal energy usage; round to a single decimal is probably better
* speed, accelaration, turning: can vary greatly between different ships; if you have e.g. a bulk freighter or a deep river, a decimal might be significant if you want to know if one more level 1 engine makes a difference. (Actually it ought to be in degrees, then arcminutes, then arcsecs, instead of decimals.)

DingusShingleton:
Honestly if there are outfits with decimal resource usage... Just change them to not be. Like why should that level of precision ever exist in an outfit? To use the standard rubric: how does it enhance gameplay? Get rid of all decimal stats, I say. Nothing is lost, clarity is gained. Win win.

[...]

PR #4304, opened 2019-05-28
Better formatting for decimal numbers.

### LINKED ISSUE #4117: Truncation of decimals causes issues
Ref: https://github.com/endless-sky/endless-sky/commit/b219799a44cfd4f9c6f19d1bf0d5358c7a6ccbe6, #3877

The current UI behavior of truncating numbers off at tenths leads to stats displaying strangely. As mentioned in the comments to the commit that did this, outfits like gatling gun ammo or javelins show as having zero mass. Additionally, this truncation gives the appearance of inconsistency in the displaying of whole numbers, as some whole numbers may appear like `576`, while other times they appear like `576.0`.
(Image courtesy of @Tadrix. The left side is with decimal truncation, the right side is without.)
comparison

An alternative, as suggested previously, would be to use significant figures as opposed to truncation, but strictly using sig figs means that something like the Hai's huge battery would display its capacity at 77,780 instead of 77,777. What I propose is that any number equal to or larger than 1000 (four digits sans the decimal places) that also has a decimal part to it would be truncated to remove the decimals. Three digit whole numbers would only show to the tenths place. Two and one digit whole numbers would then only show their decimal part to the hundredths place. And to avoid inconsistencies as to how numbers show (e.g. 500 on the dot versus 500.04 showing as 500.0), all numbers could show their decimal places, even if it is only zero.

Table of examples of what I'm suggesting:

Full number | Displayed
--------------|------------
0.0540 | 0.05
0.1 | 0.10
0.101 | 0.10
1 | 1.00
1.502 | 1.50
12 | 12.0
12.04 | 12.0
632 | 632.0
632.58 | 632.6
1052 | 1052
1052.6734 | 1053
25862 | 25862
15235.6753 | 15236

Rakete1111:
This tries to fix #4117. Any numbers up to 1000 have two decimal places, up to 10000 only one and none afterwards. This is more or less arbitrarily chosen, but since if you are in the 1000's range you really don't need an extra digit of precision I believe.

It doesn't do any rounding though, just strips off the extra decimal places.

es-fmtnum


source/Format.cpp
```diff
@@ -96,27 +96,41 @@ string Format::Number(double value)
...
-			result += static_cast<char>('0' + static_cast<int>(round(decimal * 10.)));
-		
-		result += '.';
-	}
-	
-	// Convert the number to a string, adding commas if needed.
-	FormatInteger(value, isNegative, result);
+	} else
+		// Convert the number to a string, adding commas if needed.
+		FormatInteger(value, isNegative, result);
```
Rakete1111: This is basically dead code, since `to_string` will always add a `.`. Not sure how to handle it. Currently if `to_string` doesn't do that the number is displayed incorrectly. So:

1. leave it like this and in the pretty unlikely (impossible?) case get a badly formatted number
2. remove it and add an `assert` (or just assume that there is a dot).

gunqqer:
Pretty sure this is in relationship to #3877 and I still stand by we should round by sigfigs rather than just randomly truncating or rounding where we feel. At least with just doing 3 sigfigs we'll be consistant

tehhowch:
@Rakete1111 mind bringing this up-to-date and adding to the unit tests? I started some for you in 1ecd2d2a

tehhowch:
Thanks for bringing it up-to-date!

Based on the unit test benchmark section, it seems like this is an order of magnitude slower than the previous implementation.

Some previous results: [1](https://github.com/endless-sky/endless-sky/runs/3308796359?check_suite_focus=true#step:8:98), [2](https://github.com/endless-sky/endless-sky/runs/3308425269?check_suite_focus=true#step:8:98)

Is there a better way to achieve the same result?

Rakete1111:
The first two have equivalent run times now. Only the last one is ~20ns slower. Is that acceptable?

Issue #4303, opened 2019-05-28
FW Checkmate campaign - after Defend Silver mission Pug only attack player, ignore and ignored by FW fleet


starscollide5:
**Describe the bug**
In late FW Checkmate campaign, after Defend Silver mission, FW fleet accompanying the player ignores the Pug, and the Pug ignore FW fleet, attacking only the player. 

**To Reproduce**

1. Accept Defend Silver mission

2. Complete Defend Silver mission

3A. Attack Pug to steal a JD 

3B. Accept Defend Oblivion mission

**Expected behavior**
FW fleet and Pug fight each other, like in earlier missions. 

**Link to save file**
[Stars Collide~BuggyPug.txt](https://github.com/endless-sky/endless-sky/files/3229950/Stars.Collide.BuggyPug.txt)

**System (please complete the following information):**
 - OS: Windows 10 Pro, version 1809
 - Game Source: github release
 - Version: 0.9.9

jafdy:
This is because the pug are set to be peaceful after `FWC Pug 1B`. The fix would be to make the FW ships in `FWC Checkmate Escorts` of the escort government instead of free worlds.

Pointedstick:
> This is because the pug are set to be peaceful after `FWC Pug 1B`. The fix would be to make the FW ships in `FWC Checkmate Escorts` of the escort government instead of free worlds.

Nice, wanna submit a PR? :)

starscollide5:
Submitted PR #4306

jafdy:
I'm too busy to get my branch fixed up right now, so until next week I won't be able to do much more than offer solutions.

Amazinite:
This is entirely intentional, at least for the jump drives.
Once the FW fleet is trapped inside Pug space by "FWC Pug 1", the Pug only attack in order to destroy your nuke bearing ships. Your fleet then attacks the Pug in retaliation. But after that mission ends, the Pug reputation with the Free Worlds is set to neutral, and *your* reputation with the Pug is set to positive. This is because the Pug are no longer hostile with you since they took care of the nukes. So it's then the player that becomes hostile with the Pug by stealing the JD, not the FW as a whole. And as Jafdy said, making them escorts will just make your job more difficult if you don't already have a jump drive.

What should occur is that for the Defend Oblivion mission ("FW Pug 3B"), the FW/Pug attitude should be hostile again.

starscollide5:
Resolved.

PR #4306, opened 2019-05-29
Change FWC Checkmate Escorts government from "Free Worlds" to "Escort"

starscollide5:
To resolve issue #4303


Pointedstick:
Looks like we have a better solution here: https://github.com/endless-sky/endless-sky/pull/4309

Issue #4002, opened 2018-11-15
Amending behavior of `require` in Mission Actions (on ____)


tehhowch:
In #3005 I alter the behavior of `require <outfit> 0` to perform an exhaustive check of every ship in the players fleet that is not destroyed. To give more control to mission creators, it is probably useful to define an argument for this line and the regular `outfit <outfit> <#>` that specifies *where* the outfit should be looked for. Perhaps a mission being written should only check the flagship. Perhaps a mission should only check escorts, or cargo. Currently, there is no way to instruct the game where to search, and so installed outfits can only be taken from your flagship, and given outfits are only installed onto (or placed in the cargo of) your flagship (c.f. #3037 ). With a location specification, this default behavior could be overridden.

<details><summary>c.f. discord with @Amazinite  and @zitchas</summary>

Zitchas: Is there anyway to differentiate between the three?
There's basically three separate places an outfit could be:
 a) installed in flagship
 b) installed elsewhere in the fleet
 c) in cargo

For mission purposes, knowing how to differentiate between those would be very useful.
I've already got a few missions that use that mechanic. My descriptions assume that it is actually installed.

-------------

Amazinite: Yeah, I think the suggested changes to require make sense then. But as Zitchas noted, discriminating between a player's flagship, fleet, and cargo would be nice, although that could be a separate PR.
Such ability to discriminate could also mean offering if an outfit is on the flagship but not on any escorts. E.g. someone takes notice of your flagship having a keystone but your escorts not.

-------------

Me: to me i recognize the following definitive locations:
1. PLAYER
exhaustive check. Maybe it's installed on your flagship, or your combat escort. Maybe it's installed on a parked superfreighter. Maybe it's in your flagship cargo bay. Maybe your escort just scooped it up after you jettisoned it.
Wherever it is, you have it. (Your accountant knows about this item).
2. CARGO
only checks available cargo space. Always includes the flagship's cargo. If you're boarding, that's the only cargo - your escorts didn't board this ship, you did. If you're landed, then it includes anything that any escort that landed with you brought. Any out of system ships? Too bad, they're not around. (no ships have individual cargo space while landed)

There are also ambiguous locations:
3. FLAGSHIP
only checks the flagship. Always includes the flagship's outfits. If you're boarding, it also means your flagship's cargo. If you're landed, your flagship has no cargo (it was pooled with cargo from your other escorts that landed). Maybe this also includes the pooled cargo when landed. (This is very close to the existing behavior)
[...]

PR #4310, opened 2019-05-30
Give FW Pug 2C: Jump Drive priority so that you aren't given an extra JD

Amazinite:
For the FW Pug 2C missions where the player is given a Jump Drive by the Syndicate, having both a Hyperdrive and Jump Drive installed will result in the Syndicate removing your HD and giving you a second JD. 

Giving FW Pug 2C: Jump Drive priority means that this mission will prevent the Hyperdrive and Scram Drive versions of the mission from offering if the player has a jump drive already.

<details><summary>Old description</summary>This PR makes it so that the Hyperdrive and Scram Drive versions of this mission now check to make sure that the player doesn't have a JD already.

Note: This change will make it so that the player's entire fleet is checked for a JD, even though the mission states that only the flagship is being checked. Will need to extend the behavior of `require` to be able to only check the flagship as mentioned in #4002 in order to avoid that.</details>

### DIFF
--- a/data/free worlds reconciliation.txt
+++ b/data/free worlds reconciliation.txt
@@ -1229,8 +1229,7 @@ mission "FW Pug 2C: Scram Drive"
 
 
 
-mission "FW Pug 2C: Jump Drive"
-	priority
+mission "FW Pug 2C: A Jump Drive"
 	landing
 	name "Investigate the Pug"
 	description "Travel to the Deneb system to investigate the Pug; if possible, make contact with them and determine what their goals and desires are."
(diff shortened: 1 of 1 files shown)


jafdy:
Is there a way to force the order, that was the original intention. The JD mission was checked first and that blocked the others from triggering.

Amazinite:
I imagine the jump drive version of the mission could be given `priority`, and since these are the only missions that should be offering, I don't believe that would cause any issues.

Amazinite:
Nevermind, I forgot that `priority` and `minor` don't apply to `landing` missions.

PR #4313, opened 2019-05-30
Add nullptr check for 87981a4

comnom:


MCOfficer:
Does this ref anything or is it just something you forgot?

comnom:
Ha, sorry yea it refs discord:

>tehhowch: @comnom add a && ship->GetParent() term to the isHere if statement you PR'd 2 weeks ago. It's gonna be extremely rare that an NPC is parentless and systemless but I'd sleep better with the nullptr check ;)

PR #4315, opened 2019-05-31
Player will now receive on visit dialog for cargo and passenger jobs

Amazinite:
Currently, the player will only receive an `on visit` dialog in certain missions, mainly escort missions where they have left the escort behind, or those missions where an objective must be completed first like killing/evading/disabling the enemy. There are other situations though where a player may get to the destination but need an `on visit` dialog to tell them why it isn't complete, such as cargo or passenger missions where the player may have escorts lagging behind.

This PR adds generic `on visit` dialogs to all jobs that have cargo or passengers. This PR does not touch any spaceport missions. I also skipped over certain jobs that weren't strictly "take this cargo/these passengers here," such as a few jobs that had stopovers, like the Wanderer Invasive Species jobs or a job I spotted in the Paradise Worlds file.

Should we merge this, I'll go through the actual missions and give less generic `on visit` messages for them.

### DIFF
--- a/data/coalition jobs.txt
+++ b/data/coalition jobs.txt
@@ -25,6 +25,8 @@ mission "Coalition: Ringworld Workers 1"
 		government "Coalition"
 	destination
 		government "Heliarch"
+	on visit
+		dialog phrase "generic passenger on visit"
 	on complete
 		"coalition jobs" ++
 		payment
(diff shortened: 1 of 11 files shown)


Pointedstick:
+1 conceptually; makes sense for me. Maybe use less generic messages in this PR though?

Amazinite:
Well these *are* generic jobs. The on visit dialogs are just modified versions of the same dialog for escort missions, just changed to refer to cargo or passengers instead of escorts.

Pointedstick:
I was referring to this:

> Should we merge this, I'll go through the actual missions and give less generic on visit messages for them.

If you want to do it, why not in this PR?

Amazinite:
>If you want to do it, why not in this PR?

Mainly a matter of not having too much in a single PR.

PR #4316, opened 2019-05-31
Reduced the speed of the Centipede

Amazinite:
Even considering the fact that Hai engines are very good, the Centipede is rather fast for its size. This change brings the Centipede's speed in a bit.

Comparing with a Star Queen fitted with the same engines as the Centipede, as the Centipede was intended as a Hai version of the Star Queen.

Ship | Max Speed | Turn Rate (empty)
----|-----|----
Star Queen (5 spare outfit space) | 452.7 | 140
Centipede (current, 20 spare outfit space) | 830 | 169
Centipede (changed, 20 spare outfit space) | 622.5 | 156

### DIFF
--- a/data/hai ships.txt
+++ b/data/hai ships.txt
@@ -74,8 +74,8 @@ ship "Centipede"
 		"hull" 2400
 		"required crew" 35
 		"bunks" 131
-		"mass" 128
-		"drag" 3
+		"mass" 168
+		"drag" 4
 		"heat dissipation" .75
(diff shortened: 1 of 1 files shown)


ZBok:
I did thumbs up at the original proposed speed of mid 600's but it came down more? I thought 650ish was great

Amazinite:
At ~625, the Centipede was almost 40% faster than the Star Queen. I lowered that difference to only about 25%.

PR #4318, opened 2019-05-31
Add "Intro Epilogue" mission

jostephd:


data/intro missions.txt
```diff
@@ -451,3 +451,39 @@ conversation "end of intro missions"
...
+		not "main plot completed"
+	on offer
+		conversation
+			`The spaceport at <origin> is a completely different experience than you remember. When you first came here, you were a mere child gawking at the spaceships passing through to refuel on their way to bigger and richer worlds; when you left, you were a greenie captain, fighting tooth and nail to remain one step ahead of the bank's repossessors.`
+			branch "during main plot"
+				has "event: joined the free worlds"
+
+			`This time, however, you're an established merchant captain.`
+			`You receive a VIP's welcome in this small trading post. Merchants who shooed you away as a child are queueing up to buy your goods; young pilots are asking you for apprenticeships.`
+			`You ignore them all and walk over to the airport adjacent to the spaceport, where you rent a private jet and fly to your village. You spend a few days there, but soon you start feeling landsick. After resisting the symptoms for a few days, you finally give in, say your goodbyes and head back to your ship.`
```
jostephd: BeccaBunny on discord suggests to change "days" to "hours" because every planetside visit always takes a day.

AlbertNewton: I like the idea of landsick.

But what if the player was able to take on an apprentice?

data/intro missions.txt
```diff
@@ -451,3 +451,39 @@ conversation "end of intro missions"
...
+	landing
+	source "New Boston"
+	to offer
+		or
+			"net worth" > 100000
+			has "event: joined the free worlds"
+		not "main plot completed"
+	on offer
+		conversation
+			`The spaceport at <origin> is a completely different experience than you remember. When you first came here, you were a mere child gawking at the spaceships passing through to refuel on their way to bigger and richer worlds; when you left, you were a greenie captain, fighting tooth and nail to remain one step ahead of the bank's repossessors.`
```
Pointedstick: A spaceport isn't an experience, but visiting one is. So this could be `Visiting the spaceport at <origin> is a completely different experience from the one you remember all those years ago.`

data/intro missions.txt
Pointedstick: > `where you rent a private jet and fly to your village.`

I do? Just like that? With no choice or even having accepted a mission? And for free? Feels kinda weird.

Also if you're going to have the player visit their home village, there's gotta be more detail than just "You spend a few days there, but soon you start feeling landsick. After resisting the symptoms for a few days, you finally give in, say your goodbyes and head back to your ship."

I get that adding more detail would problematically define something that was previously very undefined (the player's childhood and backstory), but maybe that part _should_ stay undefined. Maybe the player doesn't need to rent a private jet and fly home for a few days to engage in unspecified activities.

data/intro missions.txt
```diff
@@ -451,3 +451,39 @@ conversation "end of intro missions"
...
+				has "event: joined the free worlds"
+
+			`This time, however, you're an established merchant captain.`
+			`You receive a VIP's welcome in this small trading post. Merchants who shooed you away as a child are queueing up to buy your goods; young pilots are asking you for apprenticeships.`
+			`You ignore them all and walk over to the airport adjacent to the spaceport, where you rent a private jet and fly to your village. You spend a few days there, but soon you start feeling landsick. After resisting the symptoms for a few days, you finally give in, say your goodbyes and head back to your ship.`
+				decline
+
+			label "during main plot"
+			`This time, however, you're a card-carrying warship captain, flying under the colors of the Free Worlds.`
+			`Everyone keeps their distance from you, some respectfully, some in fear of getting tangled in the hostilities. You're all too used to this experience by now, but it still feels jarring to receive this treatment on your homeworld, in a spaceport you know as well as the back of your hand.`
```
Pointedstick: `tangled in` -> `tangled up in`

data/intro missions.txt
```diff
@@ -451,3 +451,39 @@ conversation "end of intro missions"
...
+
+			branch "not surprised"
+				has "FW Diplomacy 1D: offered"
+
+			`Finally, you look up an old friend. When his shift is over, you meet and catch up. You welcome the breath of normal human interaction; the brotherhood of the sea has always felt a little shallow to you. Later in the evening two retired Navy officers join the two of you for a hand of contract bridge. As you walk back to your spaceship, you are surprised to find yourself thinking of the retired officers as friends.`
```
Pointedstick: > `you look up an old friend`

Who?

> `the brotherhood of the sea has always felt a little shallow to you`

It has?

I feel like this stuff would be railroading the player into a personality for their captain that they don't necessarily feel themselves.

data/intro missions.txt
jostephd: Adding a mission or game mechanic where the player takes on an apprentice is out of scope of this PR. If that's ever done we can easily change this dialog to not mention apprenticeship seekers.

data/intro missions.txt
jostephd: Thanks for the review. I added "Visiting" but I didn't add "all those years ago" because the player may have left New Boston less than a year ago.

data/intro missions.txt
jostephd: I assume the cost of a planetside jet is peanuts to an accomplished merchant captain. The mission conditions ensure that at this point the player is worth more than a quarter of the price of a brand new space ship, so they're rich by New Boston standards.

I'm open to adding a `choice` or changing the flavor story, but what could we say instead? It's hard to write this because we know next to nothing about the protagonist's history.

data/intro missions.txt
jostephd: Done

data/intro missions.txt
jostephd: > Who?

I wrote this as though the old friend works at the spaceport, but isn't themselves a captain. Probably someone the player met during spaceflight school.

> It has?

I suppose that doesn't say quite what I meant to say. I meant to say that the player longs for one normal evening with people they'd actually met once before, after the endless stream of passengers, crewmen, navy patrols, planetside merchants and pubgoers they meet in their travels. I think most people would feel this way (there is never any indication that the player has any family on friends on board their ship, for example, there are no bunks for them), but then again, this is just a flavor mission, it shouldn't paint the player into a corner, quite the opposite, it should make the player feel at home.

Amazinite:
>*No description provided.*

Can you provide one, please? Why should this be merged? What does it add?
@endless-sky can correct me, but I recall it being said somewhere that we don't want missions that refer to your "past life" or "people you use to know" on New Boston since it isn't mentioned anywhere else. 
I could say more, but it'd basically be repeated from what Pointed has already commented on. (e.g. An old friend? Really? Who?)

Amazinite:
If you want to write an "intro epilogue," you're much better off writing one for James, since he's *literally* the intro to the game after buying your first ship.

jostephd:
> Can you provide one, please? Why should this be merged? What does it add?

Just add a bit of atmosphere, a small break from the tedium on the first time you come back to New Boston after you've got a hang of how to play the game.

>  @endless-sky can correct me, but I recall it being said somewhere that we don't want missions that refer to your "past life" or "people you use to know" on New Boston since it isn't mentioned anywhere else.

Yeah, that's precisely why the trip to the village is written so generically.

I don't think it's possible to write this dialog without adding _some_ element of the player's past life, even if it's just that spaceport employee in the FW branch of the dialog. If you see a way to do this, let me know, otherwise feel free to close this (I might try writing a James mission along these lines, but if I do it'll be a new PR).

Amazinite:
>Yeah, that's precisely why the trip to the village is written so generically.

And writing so generically is exactly what we want to avoid. 

>I don't think it's possible to write this dialog without adding some element of the player's past life

So if this mission requires mention of the player's past life, but that past life must be written VERY generically, then it doesn't seem like this is the type of mission that should be added to the game.

>otherwise feel free to close this

Okay.

PR #4319, opened 2019-05-31
New Southbound ship sprites

beccabunny:
This PR, using the assets of the Southbound Shipyard ships, changes the original sprites into more detailed ones, which is more noticeble in the thumbnail view (but is different also on the top view). 
Some of the most different ones, that changed the structure a bit for a better look (on the left the new sprite, on the right the game one:
Dreadnought
comparison
Modified Argosy
comparison
Screenshot on Glaze
Senzanome
It also changes the hardpoints to fit the new images


AlbertNewton:
I personally love the new Dreadnought.
I always thought its nacelles would be giant, cuboid, thick blocks, until I saw the flimsy ship. This looks more realistic, and also a lot cooler, which are two things you rarely get together.

AlbertNewton:
After the edit that adds all of the new sprites, I've just picked my jaw up off the floor.

kikotheexile:
AND I HELPED (a little.)

solardawning:
That Argosy is beautiful. Incredible work on everything.

beccabunny:
@Pointedstick I should have corrected the fury, but the haulers didn't have that error. I think that the difference in them is only a visual effect caused by the subsurved bevel on the object itself

Wrzlprnft:
One thing that might be of interest, is that the vertex count is way higher on these models, to achieve a more realistic look with blender internal. If we add these, we might want to extend the wiki with a few hints on how to use a couple of modifoers to improve the shading with blender internal to these levels.

Pointedstick:
Could we get a High DPI PR too? I'd love to test this out.

beccabunny:
Yup, I forgot it. Tomorrow I will make the PR

tehhowch:
Links to the new assets are required in to merge this.

beccabunny:
https://drive.google.com/open?id=1S3uAj3z6zrGdCBQYh9vDRcrXTc-Xh_F4
The assets are here. @Pointedstick I also added the high DPI

UnciaPrima:
The Argosy and the Clipper are beautiful. Everything is less flat, it's much better! We need more venting grids on human ships :)

nylki:
This is a superb visual refresh, while still staying true to the original ship design! Great job @beccabunny, looking forward to see more of you!

Amazinite:
>I hope we can see a similar treatment for more of the human ships in the future. :)

That's the plan. This project to redesign the human ships/make them look generally better has been in the works for a while. The original attempt at it died off, but they were going for completely redesigning most of the ships. I much prefer this route of just adding more detail and maybe slightly redesigning certain features, though.

Pointedstick:
That was what I thought too. However in game I couldn't really tell the difference.

beccabunny:
@Pointedstick @MCOfficer I rescaled all the thumbnails from the original dimensions to be sure i didn't just half the dimension from the high DPI one, which probably caused them to be a bit more blurry; i didn't do the haulers and the type F series since they had alredy the 2x size when rendered. The dreadnought thumbnail, unfortunately, wasn't fixed by that method, but i noticed that by scaling it to a size 5 pixels wider it was much better. I also updated the high DPI of the dread to have the new dimension (no hardpoint needed further changes, since the modification was applied just to the thumbnail). In general, they should be less blurry now

Zitchas:
I agree. This is a very nice improvement.

beccabunny:
I forgot that the Bastion is Southbound apparently. I detailed it as well, and i added it to this PR. The high DPI PR has been updated as well. This image shows the old bastion on the left, and the new on the right
comparation
These are the new files of the bastion
[bastion.zip](https://github.com/endless-sky/endless-sky/files/3389670/bastion.zip)

Pointedstick:
I've been running with this since it was submitted and I just think the ships are gorgeous.

This has been open for a month and a half with lots of positive comments. I think it's time to merge it.

kikotheexile:
The right call imho for 9.10, this will let steam beta users see an actual, visual change some aspects of the game to reinforce the "hey things are still being refined, this isn't just good, but actively getting better." point.

Tadrix:
Love the new sprites.

The only problem, subjectively, I have with the new Bastion:
bast
The cockpit, to me, looks not like a canopy now, but a processor on a motherboard, with all those tiny legs. IMO, that's a potential automata accessory, not human.

As for the former pylons on the aft, formerly they looked like fuel tanks. Now… I have no idea what they are. And the new design making less sense than the old one is, in my opinion, bad. But the rest of the ship is great.

kikotheexile:
In what world would a combat centric ship have exposed fuel tank painted yellow on it's highest profile? Hyperspace fuel in endless sky is combustible, so one *not even lucky* shot on one of these brightly colored targets would instant kill the ship. 

They were never fuel tanks. There's no way in hell.

Zitchas:
Well, it is *possible* they were, it'd just be a *really bad* design decision. Which is a scarily common human trait.

kikotheexile:
I have a hard time believing that on a mechanical front as well, however. Considering each fuel tank that grants 1 jump worth of fuel takes up 8 tons of space. If these chunks of the sprite were indeed fuel tanks, the vast majority of the ship would be fuel storage, meaning its weight would be mostly fuel. 

Considering its cargo space and combat abilities, as well as default fuel storage amount? No way in hell.

Tadrix:
There's no way in hell warships would have half-exposed canopies in the fore, while relying on forward-shooting weapons. A lucky shot would instantly disable the ship. Yet, in ES, here we are, this is true even for the largest ships.

When talking aesthetics in ES, practicality often takes a back seat. Also, complaining about the lack of realism normally is *my job*, yet I'm not calling for redesigning *all* the ships in the game, since it wouldn't be Endless Sky any more.

kikotheexile:
look at it logically, using in universe examples. 

The closest analogue we have to fuel tanks are fuel pods, which are internal, and weigh 8 tons. 5 of them would be 40 tons. Each fuel pod is small enough to fit within a fighter without being seen. 

To imply that those exposed armor components are fuel pods would, from an in universe perspective, make zero sense. One of those pods alone could easily fit 500 jump fuel per, and there are 2 pods. This means that, by the logic of these being fuel pods, the bastions default fuel should be over 1,000. 

But with the stupid amount of internal cargo space open, why wouldn't you just expand fuel internally? Ya know, rather than putting a brightly colored explosive on the outside of your hull, on your ships most vulnerable profile.

I repeat, no way in hell those are fuel pods. There is zero logic to back it up in or out of universe unless southbound WANTED the bastion to die. They are armor components and, potentially, either cargo space or free outfit space, and to look cool to generate sales.

MCOfficer:
At which point did this stop being about the southbound redesign? i think becca did a great job, that's what my stomach tells me. i don't care if it looks like a bit like a fuel tank, we're a thousand years in the future so it could be anything.

Pointedstick:
Agreed. Let's move onto the next batch of ship redesigns! :)

kikotheexile:
I think Becca is currently still working on a mod, but after that I wanna say our next planned batch was BG, although it could have been the southern Birbs. Dunno what they decided on for sure, but I'm personally pushing for Betelgeuse

Tadrix:
Following the same logic, if it isn't a tank, in the Becca's design, we see an exposed chute where formerly an armoured shell was. For a warship it makes zero sense to have such opening (or remove the shell), inviting enemies to send a missile down that hole of a ship that isn't very manoeuvrable.

Anyway, as I said, it was a subjective opinion on one ship's looks and I didn't intend to have a longer argument. Overall, in the same subjective opinion, Becca did a great job.

LocalGod79:
heya, just thought i'd point out here that the boxwing was original art, and should not be moved into a section that denotes M.Z. derivative works within the copyright.

Issue #4320, opened 2019-06-01
Escorts do not replenish crew when refueling


piannucci:
**Describe the bug**
When a freshly-captured escort with insufficient crew lands to refuel while following the pilot, it still has insufficient crew when it takes off.

**Expected behavior**
Escorts never take off from inhabited planets with less than full crew.

**System (please complete the following information):**
 - OS: macOS Mojave 10.14.3 (18D109)
 - Game Source: Steam
 - Version: 0.9.9

tehhowch:
Ref https://github.com/endless-sky/endless-sky/issues/367#issuecomment-162529426

petervdmeer:
Closing issue; escorts only hire (new) crew when the player is landed and the escort is landed on the same planet. The linked issues give a good reason why escorts can refuel while the player is in flight, but I don't see any reason why we would add crew replenishment or repairing on top of refueling any time soon.

PR #4329, opened 2019-06-02
Landed ships hire required crew when possible.

piannucci:
Suggested fix for #4320.


tehhowch:
Per comments in the issues I linked in #4320, I'm not sure we want to add this. If we switched from the jump-based calendar to a frame-based calendar, sure. But there are a lot of preliminary steps to undertake first before we can do that.

petervdmeer:
Closing issue; escorts only hire (new) crew when the player is landed and the escort is landed on the same planet. The linked issues give a good reason why escorts can refuel while the player is in flight, but I don't see any reason why we would add crew replenishment or repairing on top of refueling any time soon.

Issue #3900, opened 2018-08-17
Install guns in turret ports


jostephd:
Should we allow guns to be installed in turret ports? A turret port is a place where a weapon can be mounted to shoot in all directions so mounting a weapon that can't turn is perhaps wasteful but not inconceivable. Players will have new strategic options to mount guns or secondary weapons in the turret hardpoints. A heavy warship will comfortably fit 10 Heavy Rocket Launchers with room for anti-missiles.

If I'm not talking complete nonsense, we can take this even further. There's no reason to require a gun mounted in a turret port to face forward, it could just as easily be mounted facing sideways or aft. The direction a gun faces should only be changeable in a shipyard, of course. For simplicity the player shouldn't be allowed to specify the direction in degrees but only to choose between some preset values, for example ±5, ±10, ±20, ±45, ±90, ±170, 180 degrees from forward.

Bladewood:
If you're asking for guns on turret mounts, do you also propose that turrets be installable in gun ports?

The gun-turret count is supposed to be one of the main distinguishing features of different ships, so I don't see this happening.
I can particularly see this becoming troublesome for balance should eventually some weapons that only have gun forms be added, because a ship would not be intended to run that many "guns".

jostephd:
> If you're asking for guns on turret mounts, do you also propose that turrets be installable in gun ports?

No.

> I can particularly see this becoming troublesome for balance should eventually some weapons that only have gun forms be added, because a ship would not be intended to run that many "guns".

Aren't guns generally _weaker_ than corresponding turrets? For example, Sunbeam has identical statistics to Sunbeam Turret, but is cheaper and 12T smaller. If I install a Sunbeam gun in a turret hardpoint, my ship is _weaker_ than it would be if I'd installed a turret in that hardpoint, because now I can only fire forward. Do the outfit space savings compensate for that?

There are already ships that have many gun ports (Vanguard, Carrier, Shield Beetle).

MCOfficer:
however, there are also some ships that specifically have many turret mounts, for example the behemoth. the ability to stack mass expansions (to some degree) and, stack particle cannons could quickly get out of control.

jafdy:
The idea of having fixed weapons that don't fire forward has been discussed before, while not a tricky problem for a human it is really hard for an AI. So that won't be happening, not soon at least.

Bladewood:
[...]

PR #4333, opened 2019-06-04
Stationary turret for missile hardpoint

RPBCACUEAIIBH:
With excelent tracking missiles it produces a nice curve shooting effect that was broken for me by the default turn rate. So I fixed it. :)
Add "turret notrurn" 1 to outfit attributes to make the turret stationary. (It overrides the specified or default turnrate!)

### DIFF
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -92,8 +92,13 @@ void Outfit::Load(const DataNode &node)
 			child.PrintTrace("Skipping unrecognized attribute:");
 	}
 	
+	// Stationary turret for missile hardpoint
+	if (IsWeapon() && attributes.Get("turret mounts") && attributes.Get("turret noturn"))
+	{
+		SetTurretTurn(0.);
+	}
 	// Legacy support for turrets that don't specify a turn rate:
-	if(IsWeapon() && attributes.Get("turret mounts") && !TurretTurn() && !AntiMissile())
+	else if(IsWeapon() && attributes.Get("turret mounts") && !TurretTurn() && !AntiMissile())
 	{
 		SetTurretTurn(4.);
 		node.PrintTrace("Warning: Deprecated use of a turret without specified \"turret turn\":");
--- a/source/OutfitInfoDisplay.cpp
+++ b/source/OutfitInfoDisplay.cpp
@@ -76,7 +76,8 @@ namespace {
 		{"unplunderable", "This outfit cannot be plundered."},
 		{"installable", "This is not an installable item."},
 		{"hyperdrive", "Allows you to make hyperjumps."},
-		{"jump drive", "Lets you jump to any nearby system."}
+		{"jump drive", "Lets you jump to any nearby system."},
+		{"turret noturn", "Stationary hardpoint."}
 	};
 }
 


tehhowch:
If it isn't a turret then it shouldn't use a turret mount IMO.

Amazinite:
`Add "turret notrurn" 1 to outfit attributes to make the turret stationary. (It overrides the specified or default turnrate!)`

Why not just set the turn rate to 0 then instead of adding another attribute which pretty much just means `"turn rate" 0`?

Tadrix:
Just use gun ports for non-turrets.

RPBCACUEAIIBH:
> If it isn't a turret then it shouldn't use a turret mount IMO.

Well it's easier to just turn off the motor of an existing turret then to invent a new hardpoint or something...


> `Add "turret notrurn" 1 to outfit attributes to make the turret stationary. (It overrides the specified or default turnrate!)`
> 
> Why not just set the turn rate to 0 then instead of adding another attribute which pretty much just means `"turn rate" 0`?


Because that's what I did initially, but someone added a default turn rate that specifies a turn rate of 4 if the turn rate is 0, and also gives an error... so "turn rate" 0 is not working anymore!

> Just use gun ports for non-turrets.

Gun ports fire forwards, you don't get the same visual effect, and not every ship has gun ports...

Tadrix:
`and not every ship has gun ports`
That's for a reason. The reason being that no ship should be able to do everything. The concept of an immobile turret is a wasteful one, and it's an attempt to circumvent the existing game balance.

Besides, the longer trajectory you're creating in this way is wasteful (presenting more opportunities for the missile to be shot down for no other apparent reason than the designer stupidity, and throwing the ship off-balance if we had the realistic inertia physics; there's a reason why IRL most aircraft launch their missiles forward). See also: [Artistic curving trajectories for no apparent reason](https://tvtropes.org/pmwiki/pmwiki.php/Main/Roboteching).

If you insist on having a weird firing angle, don't you think AI should be accounting for it? Which is the reason I believe that the point that you brought up is a specific case of *broadside firing mechanics* which were discussed, but never implemented. You could bring it up, if you wish, but first look up already existing issues on that topic.

jafdy:
I think you need to find a workaround to your problem, the AI currently isn't able to use fixed weapons off angle. Perhaps use a very small turn rate so the turret doesn't turn perceivably.

RPBCACUEAIIBH:
> That's for a reason. The reason being that no ship should be able to do everything. The concept of an immobile turret is a wasteful one, and it's an attempt to circumvent the existing game balance.

Look... I did not ask you to waste your time adding a feature for me, I simply did it. I did not add an actual turret that could possibly throw the game out of balance, I simply added the possibility to make the turret stationary. There's no need for complicating the AI it's a game not a real life simulator. A game where visuals count! (or at least they should...)

RPBCACUEAIIBH:
> I think you need to find a workaround to your problem, the AI currently isn't able to use fixed weapons off angle. Perhaps use a very small turn rate so the turret doesn't turn perceivably.

The AI doesn't need to use fixed weapons, excelent tracking will do... :)

jafdy:
If you add the ability to have an off angle turret then someone will try to put a fixed weapon on it. When they ask why it isn't working do you expect us to say "because some guy wanted this function for a mod he was doing". Adding a new feature to the game just for some mod should be a last resort when all workarounds have failed, and I have offered a workaround, using a number as large as 0.000004 will result it a turret that would take over an hour to make a full turn.

RPBCACUEAIIBH:
> If you add the ability to have an off angle turret then someone will try to put a fixed weapon on it. When they ask why it isn't working do you expect us to say "because some guy wanted this function for a mod he was doing". Adding a new feature to the game just for some mod should be a last resort when all workarounds have failed, and I have offered a workaround, using a number as large as 0.000004 will result it a turret that would take over an hour to make a full turn.

Ok, will thest that, if it works, fine. I'll close this... What's the smallest value btw?

MCOfficer:
> Ok, will thest that, if it works, fine. I'll close this... What's the smallest value btw?

It's defined as a `double`, so on most systems, the closest positive non-zero value should be 2^-1074.

That's 
`4,9406564584124654417656879286822e-324`
 or (it's hard to find calculators which don't break with these numbers, naturally)
 ```
1/202402253307310618352495346718917307049556649764142118356901358027430339567995346891960383701437124495187077864316811911389808737385793476867013399940738509921517424276566361364466907742093216341239767678472745068562007483424692698618103355649159556340810056512358769552333414615230502532186327508646006263307707741093494784
```

OR (brace yourself):

```
0.000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000004940656458412465441765687928682213723650598026143247644255856825006755072702087518652998363616359923797965646954457177309266567103559397963987747960107818781263007131903114045278458171678489821036887186360569987307230500063874091535649843873124733972731696151400317153853980741262385655911710266585566867681870395603106249319452715914924553293054565444011274801297099995419319894090804165633245247571478690147267801593552386115501348035264934720193790268107107491703332226844753335720832431936092382893458368060106011506169809753078342277318329247904982524730776375927247874656084778203734469699533647017972677717585125660551199131504891101451037862738167250955837389733598993664809941164205702637090279242767544565229087538682506419718265533447265625
```
(that's 323 zeroes after the comma)

Sorry for the digression. Credit goes to @0kku for explaining to me how `e` works.

jostephd:
Ref #3900

RPBCACUEAIIBH:
Right... Thanks! :D I think 0.00000001 is slow enough for me not to notice any change in the past few hours, and does't give an error... so I think this is solved!

PR #4340, opened 2019-06-06
Tribute Fix: Updating Defense Fleets during the Pug War

Zitchas:
This PR expands the events that switch control of the Pug occupied systems during the course of the war.

The earlier events now include changes so that pug fleets replace the human ones while the Pug are in control of those planets.

The pug defense fleet is:
			`fleet "Small Pug" 15`
			`fleet "Large Pug" 3`

The later events replace those fleets with human fleets again after the planets are liberated.

The defense fleet is the same fleet as was defending the planet before the Pug war, with the addition of 2 additional "Large" fleets of the same type. If the planet was only defended by small fleets (such as "Small militia") it is now defended by the same number of small fleets, plus two large fleets of the same type. If it was defended originally  by large militia, it is now defended by n+2 large militia, etc.

This increase in their defense strength is intended to reflect the increased military presence in the former occupied worlds in the wake of the Pug war.

In all instance the tribute amount and threshold are unchanged.

Thanks to Sasbot on the Endless Sky discord for pointing out the discrepancy of having human defense fleets defending Pug controlled worlds.

Given that having human (and in some cases republic or syndicate) defense fleets defend a planet while under the control of the pug is flat-out wrong, I've updated the title to label this as a fix.


tehhowch:
I would assume the Pug do not allow tribute if these planets to be a possibly more relevant response.

Zitchas:
Certainly possible. That being said, the mission descriptions mention them taking to the skies to destroy captains that dare to resist them. I'd interpret demanding tribute as the player going up to the Pug and telling them that they refuse to submit, and thus merit the same sort of response that the ships in the mission were described as receiving.

SirStigII:
Perhaps the planets that say "no thanks please" have impressive enough planetary defences that an orbital bombardment would not even be noticed? Although it doesn't make sense that a higher combat rating would induce them to launch a defence fleet.

Zitchas:
If that's the case, that works. But nowhere in-game is it described, with the exception of pugglemug. (and this PR doesn't touch it)

One of the instances of refusing is the pacifist monastery. Personally, I think that planet should just give a 100 credits tribute payment. After all, you are demanding tribute, not inviting them out to a duel for a cash prize. Planets that can't (or won't) mount a defense should just give the payment straight off - although admittedly it should be a very small amount (100 credits per day or less).

Of course, that is unless they have access to significant enough planetary defences (city-shields and the like) that even a T3 fleet doesn't phase them.

SirStigII:
For the sake of this pr I agree that the pug would likely be the ones fielding a defence fleet, but seeing as they are occupying the worlds I would imagine the response to be overwhelming (I'm imagining earth levels of ships)

Zitchas:
I have it set to 15 waves of the "Small Pug" fleet, followed by 3 waves of the "Large Pug" fleet. Which is pretty substantial, but not quite to Earth levels. Note that the game doesn't wait for one wave to be killed before sending out the rest. So the player is likely to get swarmed quite quickly.

Amazinite:
I'm wondering if this and #4273 should be rolled into a single PR that also extends the tribute system to every other planet, and causes the defense fleets to change with governmental changes; since the tribute system is largely an afterthought in this game, why not go big or go home?

Zitchas:
In terms of merge quality, I think it is preferable to do each region as an individual entity. That way they can be checked, approved, and merged as sets. This also means that parts that are already done (like the Remnant and this Pug fix) don't languish while people debate over the size and composition for fleets in places like, say, Coalition space. We don't have much to go on in terms of fleet strength there, so I suspect that those areas will take a lot of discussion to resolve.

Zitchas:
So, like this?
```
	remove tribute "Rand"
	remove tribute "Oblivion"
```

Zitchas:
Just to clarify, in regards to the behaviour I'd like to see.
1) When the pug take over a planet, the planet stops being dominated (if it ever was).
2) The planet is now defended by the pug, and can be tributed as normal - except the Pug have no compunction to actually pay the player. So they'll act as described in the mission and take off to destroy anyone who challenges them, but even if defeated they won't pay anything (except maybe a token 1 credit just to be insulting)
3) When the pug lose control of the planets, any domination that the player has achieved is removed.

Basically, this allows the player to fight a swarm of pug if they so choose to do so, which satisfies the mission description, but doesn't provide any real benefit to doing so.

Amazinite:
I'd imagine it would be
```
planet "Rand"
	remove tribute
```
but the creating events wiki page doesn't say that the remove/add keywords can be used on tribute. 
Even if you're not intending to do that, you're still missing `"tribute: <planet name>" = 0` in order to remove preexisting tribute from a planet, but I'm not sure if that actually relinquishes the planet or just sets the planet's tribute to 0 while it's still considered dominated.

Zitchas:
OK, I think I'm going to hold off on making further changes until I've got some confirmation as to how exactly we can remove the "currently tributed" status from a planet. I've asked in `The Hub` and hopefully @tehhowch can provide some insight into the code requirements.

Amazinite:
With the merging of #3128, an alternative could be to have the Pug give you the middle finger and reduce all your tribute payments to a single credit instead of completely removing it. That would be achieved by doing `"tribute: <planet>" /= <whatever the standard tribute for that planet is>`. Then once the planets return to their normal governments, you would do `"tribute: <planet>" *= <that planet's standard tribute>` to set it back to normal.

Zitchas:
That's an interesting possibility. The way I have it currently simply reduces it to 1.

Amazinite:
Well as of right now if a player had already had these planets dominated beforehand, they'll still be receiving the original tribute amount, hence why I suggested reducing the tribute amount after the Pug show up since you can't outright relinquish planets through events right now. So if anything changes, I'd say that.

Zitchas:
So the way I have it now with all those planets set to `tribute 1` isn't actually reducing the tribute those planets give?

Amazinite:
No. Changing that only changes what you're given when you dominate the planet. You're changing the attributes of the planet. It doesn't change existing tribute payments, as those are conditions in your save file.

Add to Pug Invasion:
```
"tribute: Rand" /= 500
"tribute: Oblivion" /= 500
```

Add to Pug Invasion 2:
```
"tribute: Silver" /= 700
"tribute: Shiver" /= 300
"tribute: Millrace" /= 1400
```

Add to Pug Invasion 3: 
```
"tribute: Maker" /= 1700
"tribute: Furnace" /= 1200
```

Add to Pug Invasion 4:
```
"tribute: Reunion" /= 2900
"tribute: Kraken Station" /= 500
```

Then when the planets flip back to their original government and tribute, you'd swap the `/=` for `*=`.
Also, it looks like you've changed the Shiver tribute twice, once in `"liberation of altair"` and again in `"pug territory liberated"`. Also, have you checked the FWC events?

Zitchas:
OK, thanks for spelling that out. I've got the division change implemented.

As far as restoration goes, do we want to restore it? As it stands, the tribute for the planet is reset to full value after the liberation, but the amounts aren't restored. Which kind of fits, given that the system has basically had a new government put in charge. Personally, if some captain had intimidated me into paying tribute, then I got conquered by aliens then the FW and Republic joined forces to free my planet again... Well, let's just say that restoring tribute payments to some pirate isn't going to be high on my list of priorities. If said pirate (player) really wants to get that tribute back, then can come back and ask for it the hard way. (aka release the planet then dominate it again.)

Amazinite:
Actually, this could be an elegant solution.
```
mission "check domination"
	invisible
	landing
	to offer
		or
			has "<whatever missions or event that occurs just before the pug invasion in reconciliation>"
			has "<whatever that mission or event is in the checkmate branch>"
	on offer
		fail
		"shiver dominated" = "tribute: Shiver" / 300	
		# Do this for every planet that changes governments all in this mission with the template
		"<planet> dominated" = "tribute: <planet>" / <original tribute>
```
So the condition `"<planet> dominated"` will either be 1 if the planet was dominated before the invasion and 0 if now.
Then when you're restoring tributes, you'd do
```
"tribute: Shiver" *= 299 * "shiver dominated" + 1
# The template
"tribute: <planet>" *= <original tribute - 1> * "<planet> dominated" + 1
```
This means that if the planet was previously dominated, then the tribute is multiplied by (299 * 1 + 1) = 300, that being the original tribute amount. But if the planet wasn't dominated, then you get (299 * 0 + 1) = 1, so the tribute remains the same (i.e. 1 if dominated during the invasion, 0 if never dominated).

Zitchas:
OK, so I currently have the restoration of tribute being multiplied back up to its full value as suggested previously.

I agree, that is a fairly elegant way to establish whether the player dominates a planet before or during the events of the Pug war; and then leave the tribute the player is receiving at 1 if they got the tribute during the war.

From a lore perspective, if planets are willing to restore the tribute they are paying when the pirate who dominated them also freed them from the pug; would they also be willing to pay the full tribute when the pirate went above and beyond to destroy part of the pug fleet that was occupying them?

The Pug fleet occupying those worlds is almost certainly a harder fight than the original fleets, so it isn't as if doing so is an easy path to dominating them. The only real difference is that one can dominate the planet under the pug without generating negative karma with any human faction. So the residents of said planet might be even happier to pay full tribute value than they would in the earlier scenario of paying the pirate that destroyed the human defence fleets first, then saved them from the Pug.

Amazinite:
I'd say that determination is up to you.

Zitchas:
In which case, I think I'll let the newly-freed planets honor the domination done under the pug, since they probably appreciated seeing a fleet of their occupier's ships blown out of the sky. 

(The pug don't seem particularly cruel. If they were the sort to, say, kill a hundred people for every one of them that die at the hands of a human as a revenge thing, then the residents would probably be much less happy about it having the fleet destroyed prior to actually being freed.)

Zitchas:
@Amazinite OK, I think that should do it. Now the oft-forgotten Reconciliation branch *also* has the fun of losing tribute payments from worlds the pug occupies for the duration of the occupation. And likewise, the fun of dealing with pug fleets if the player is brave enough to demand tribute from them. Not that they'll get anything worthwhile from it.

data/events.txt
```diff
@@ -2702,16 +2772,36 @@ event "fwc pug defeated"
 		security 0.1
```
Amazinite: The security changes to Silver, Rand, and Oblivion are now redundant here since you made the changes when the government flips.

Zitchas: I thought about that - I just wasn't sure if there might be some small branch I was overlooking that might skip one or the other. If you say they are redundant, though, I'll remove them.

Zitchas: resolved

data/events.txt
```diff
@@ -2587,6 +2740,12 @@ event "fwc reconnect vega"
 
 
 event "fwc pug defeated"
+	"tribute: Shiver" *= 300
+	"tribute: Milrace" *= 1400
```
Amazinite: `Milrace` -> `Millrace`

data/events.txt
```diff
@@ -2015,6 +2072,14 @@ event "pug territory liberated"
 		government Syndicate
 	system Caph
 		government Syndicate
+	"tribute: Rand" *= 500
+	"tribute: Oblivion" *= 500
+	"tribute: Silver" *= 700
+	"tribute: Milrace" *= 1400
```
Amazinite: Same typo here.

Amazinite:
Tinker gets taken over by the Pug in FWR but its fleet and tribute are never changed.

data/events.txt
Zitchas: Fixed. Thanks!

data/events.txt
Zitchas: Fixed

Amazinite:
>For some reason Shiver isn't giving me the tribute it should after FWR.

Well well well. For some reason the event `"liberation of altair"` is never actually used.

Amazinite:
In FWC, Silver isn't owned by the Pug until `"fwc peace with the navy"` should therefore not have its fleet and tribute changed until that event.
Millrace, Reunion, and Kraken Station are never involved in FWC and should therefore not have their tribute or security changed.

Zitchas:
OK, I think I got all of those.

Amazinite:
Looks like you're still missing this.
>Tinker in the Diphda system gets taken over by the Pug in FWR but its fleet and tribute are never changed.

Zitchas:
There. Now Tinker changes as well. Thanks for catching that.

Issue #4173, opened 2019-02-03
Text substitution from condition variables


Janaszar:
As https://github.com/endless-sky/endless-sky/pull/4164 will be allowing developers to use the Phrase system to select "random" choices of values for names, sentences, and the like in developing missions, there needs to be a way to have certain phrase values stored and recalled.  For example, having randomly picked names, cargos, and the like would only work if your initial and on complete messages used the same words; "Alice asks you to deliver Bon-Bons to Charlie" leading to "David accepts the Eclairs from Fiona" doesn't make sense.

As such, I would like it if the save/load method of storing and recalling numeric variables, such as Reputation, be extended to account for strings.
Alternatives, like resetting the RNG seed or "freezing" it somehow, may reduce the available combinations that could be encountered, depending on how "Random" the RNG used is.

tehhowch:
The current supported text replacements are detailed in the wiki: https://github.com/endless-sky/endless-sky/wiki/CreatingMissions#replacement

Note that the transported item is already replaceable. It's my opinion that missions involving named characters that the player should care about are probably supposed to be mission strings using conversations, and if the player isn't supposed to care about the name then generic nouns are preferred to avoid "noise."

Related: https://github.com/endless-sky/endless-sky/issues/2351
https://github.com/endless-sky/endless-sky/pull/2456
https://github.com/endless-sky/endless-sky/pull/3384#issuecomment-350551608
https://github.com/endless-sky/endless-sky/pull/2456#issuecomment-354668281

Janaszar:
Yup, know those replacements.

The main reason I would want such a change is to support "free floating" mini-campaigns.  For instance, if you get a certain rep condition (say <1000 Republic, >1000 FW), you could trigger a mission chain where some FW sympathizers in Republic space offer you a mini-chain of sending supplies.  But if your rep is reversed, then it could be some FW-based Republic spies sending info home, but all using the same mission base.  Seeing #2456 & #3384, I would say that that sort of functionality should also be included.
This way, we can code a baseline mission, write up a few dozen alternative phrases/names/options, and essentially add a ton of "new" missions everywhere.

PR #4342, opened 2019-06-06
Added on visit dialogs for nearly all missions that can trigger them

Amazinite:
Ref: #4315 

On visit dialogs for all remaining missions in the game. I split up the PRs to distinguish between jobs only using generic dialogs and proper missions with unique ones.

I've also made a new file named `dialog phrases.txt` used to store the generic dialogs used in various files.

This is taking longer than #4315 given that it isn't 100% copying and pasting `dialog phrase "generic..."` over and over again. Decided to PR what I have done so far.

- [x] Jobs missed in #4315 
- [x] Hai missions
- [x] Deep missions
- [x] Transport missions
- [x] Remnant missions
- [x] Wanderer missions
- [x] Free Worlds missions

Going forward, it should be standard to give a mission an on visit dialog if the mission can trigger it.
<hr/>

### Notes

I haven't added dialogs to a select few missions where it didn't seem like it would make sense, such as the April Fools job or @Pointedstick's shady passenger double cross mission; the passengers had to be in the player's flagship for the mission to make sense, but an on visit dialog would suggest that the passengers were in a different ship the entire time.

In instances where the on visit dialog could be triggered by multiple things and mentioning all the possible triggers didn't make sense, I went with that I considered to be the most likely cause of triggering it. For example, during the mission where you have to find the Alpha base, the on visit dialog could be triggered either due to Elias not being in the system, or the Alpha ship not being killed. I decided to have the on visit mention the Alpha ship not having been destroyed, as I'd find it more likely that the player would run back to Alta Hai after finding the Alpha base instead of the player not having Elias in the system.

### DIFF
--- a/data/deep jobs.txt
+++ b/data/deep jobs.txt
@@ -181,6 +181,8 @@ mission "Mystery retrieval"
 		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Hai" "Quarg"
 	on stopover
 		dialog phrase "deep mystery cube pickup"
+	on visit
+		dialog phrase "generic missing stopover or cargo"
 	on complete
 		"mystery cube" ++
 		payment 5000 1000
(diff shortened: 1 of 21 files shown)


tehhowch:
One issue is that visit messages are repeatedly displayed if the player lands again and still has not satisfied the criteria. Since we now have the ability to randomize dialog, it may be better to use some of that functionality.

What would be better yet would be to use some user-specified substitutions (e.g. #4173), so you could define something like `<dignitary>` and come up with several phrases that fit in various missions, and thus have dialog variation without much drudgery work.

Ex

```bash
mission "..."
    ...
    substitutions
        "<dignitary>" "Senor Frog"
    ....
    on visit
        dialog phrase "something that uses <dignitary> in its output"
```

Amazinite:
@tehhowch Do you want to add that feature and implement its usage with this PR, or should we merge this PR then go back and make changes accordingly once that feature is added?

PR #4346, opened 2019-06-07
Remant: Fixing the Gascraft to prevent another avenue of Cloak installation

Zitchas:
Added built-in thrust & turn equivalent to a set of crucible engines.

Removed the set of crucible engines & equivalent amount of space.

This change means that the player cannot swap these components out for the Coalition's small thrust module and small steering module, thus ensuring the player can't use those to outfit the gascraft with a cloaking device.

Note that a player could still technically replace the millennium cell with a Small Biochemical Cell and fit the Cloak in, but they would be unable to cloak for any real length of time, especially while maneuvering. I suspect that trying to use the cloak with said cell would probably actually be harder than the normal dodging. Since it is probably actually harder, I'm leaving it in, whereas with the smaller engines, it becomes easier.

Related thread: https://github.com/endless-sky/endless-sky/pull/3535 Wherein Endless-Sky describes how he removed the capacitor to force the player to not use the cloak.

### DIFF
--- a/data/remnant ships.txt
+++ b/data/remnant ships.txt
@@ -193,12 +193,18 @@ ship "Gascraft"
 		"energy capacity" 150
 		"ramscoop" 0.5
 		"cargo space" 8
-		"outfit space" 88
+		"outfit space" 54
 		"engine capacity" 34
 		"shield generation" 0.6
 		"shield energy" 0.4
 		"hull repair rate" 1.2
 		"hull energy" 0.9
+		"turn" 448.0
+		"turning energy" 1.32
+		"turning heat" 1.4
+		"thrust" 18.0
+		"thrusting energy" 2.65
+		"thrusting heat" 2.5
 		"gaslining" 1
 		"outfit scan power" 24
 		"outfit scan speed" 1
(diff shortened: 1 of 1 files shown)


Bladewood:
As @Lineth said:
>Although at that point, it's more effort than just playing the mission straight and dodging the antimatter cannon.

My impression is that since it already takes a fair amount of work to get Coalition gear, it just becomes less fun to be locked onto one path like this. The archon-dodging mission is a puzzle, and I'm interested in seeing puzzles with multiple solutions, not just one.

Zitchas:
That's fair. I personally like having multiple solutions, but MZ took the first step to blocking it out, this puts it out there to block the rest of the way.

Amazinite:
If the point of this change is to avoid having the player be able to install a cloak on their gascraft, but you admit that it's still possible to do regardless, then why make the change?

Zitchas:
Mainly because this change allows installing the cloak to be useful. With this change, they would have to strip out the reactor, and all the smaller options for reactors/batteries are low enough it'd probably be unable to sustain the cloak for anywhere even close to long enough to get to the planet.

jafdy:
I'm not sure we want to do this, while we don't want to make using the cloak too easy we shouldn't make it too hard either. This allows players who really can not fly to use the cloak instead. As long as the way to fit and use the cloak comes with it's own issues it will be fine.

Zitchas:
Sounds good, then. I think I'm good with that. 

Basically, it means that using cloak with the gascraft is impossible to do with human or tier 1 tech, but possible to do if one has access to alien or greater than tier 1.5 tech, although it'll still be problematic.

Issue #1258, opened 2016-06-22
[question/enhancement] Follow Target


NickGrease:
I may be remembering incorrectly, but did't EV/Nova have a keybinding for 'Follow Target' that would align your ship in the direction of your target (and full throttle, I think)? I've only been playing a few days, and auto-aim seems to partially work in pointing your ship the right way, as long as it's an enemy, and already in your general bow direction. Seems impossible to follow/align to friendlies off screen, not to mention planets/stations which, again, I seem to remember doing in Nova. The very non-obvious white dot around the target display reticle helps, but getting your ship going in the right direction seems clunkier  then it needs to be. 

Anyways, I want to thank everyone involved in making this great game.

Alkallid:
Pressing the land key can get you in the direction of any planets, and so usually the center of the system if you maybe happened to idle and drift off. Generally all ships will be around there, so I don't really see the need unless you're pirating merchants in uninhabited systems.

NickGrease:
This has more to do with keeping your forward guns pointing at the target than where ships are generally located in the system. It made switching targets and facing/moving toward them them a lot easier, especially if the targets were offscreen. The AI can do it, the original game could do it, and at this point, my muscle memory still has my finger searching for the follow key :) Maybe I'll get used to it, but I miss being able to do a fly-by, coast for a bit, hit a key to rotate to target direction and fly head on guns blazing.

Alkallid:
Ah, so moment strafing?

NickGrease:
Sorry, I'm not sure where the confusion is coming from. Either it wasn't in the original, which is explains your response, or it was, and you're suggesting I was the only person to use it. Either way, I would find it useful to have a key that would align your ship to the target.

LocalGod79:
duplicate:
#166

NickGrease:
Thanks @LocalGod79

Issue #166, opened 2015-11-07
Feature request: auto-aim button


SolraBizna:
My playstyle frequently has me in highly reflex-critical situations with relatively fragile ships. In those situations, precise control over my ship's attitude is more important than landing hits on my _selected_ target with my fixed weapons, and therefore I don't want auto-aim enabled. However, that's not always the case.

I propose adding a key to Endless Sky that performs "Automatic aiming" behavior while pressed, regardless of the current setting. The nearest equivalent is the EV "autopilot" key which simply points your ship toward the target.

Pointedstick:
Seconded. I often want to point my ship in the direction of a selected but non-visible ship and reflexively hit the 'a' key--which does this in EV.

dplepage:
A few times I've had auto-aiming kick in even when I wasn't holding fire, and on at least two occasions this has killed me - I hit the key to hyperjump but when enemy fighters got close my ship started tracking them instead of turning to make the jump. This is probably a bug in and of itself, but I haven't yet figured out how to reliably reproduce it.

Making automatic aiming an explicit button would certainly prevent that from happening.

endless-sky:
@dplepage, is it possible you had automatic firing active as well as automatic aiming?

dplepage:
@endless-sky No, I checked that - I have automatic aiming but not automatic firing. The bug seems to happen regardless of whether or not I'm holding the jump button.

SolraBizna:
For what it's worth, I disabled automatic aiming immediately after I first noticed its effects, and I am maybe 40% confident that I was not holding the fire button when it happened.

endless-sky:
Are you certain that the reason your ship turned toward approaching enemies
was not that you had been hit with a projectile that imparts momentum (like
a particle cannon or a missile) and your ship was having to turn around in
order to attempt to bring itself to a stop again?

dplepage:
@endless-sky Oh, that hadn't occurred to me. That probably is what was going on, and it only seemed like auto-aiming because "towards the target" is both the auto-aiming direction AND the direction opposite the momentum imparted by the target's missiles. Thanks!

dplepage:
I have a patch for this in #314; @gchapyzh suggested that maybe it should be a toggle so that you don't have to hold down the key while you're fighting. I'm not sure if I'd like that more - does anyone else have feelings about holding a key vs. a sticky key?

airtroop392:
[...]

PR #4349, opened 2019-06-08
Autosteer key for pointing the ship at the selected ship, asteroid or planet.

ChamEV:
Adds a keyboard key for auto steering your ship to point it at the targeted ship, asteroid, or selected planet giving the player more control of their movement separate from firing.

Named autosteer to avoid confusion with the more advanced autopilot functions.  Properly breaks autopilot jump/landing/boarding as other input controls do. 

Rearranges the preference panel and puts cloak in it's own category instead of under weapons.  Added z as the default key.

https://cdn.discordapp.com/attachments/266345072554016768/586514092643188760/Controlkeyspanel.png

(these are the default controls)

Example in action: https://streamable.com/b3fk5


Zitchas:
I don't have access to a system that can test this, so my question is thus: Does the autosteer act like a movement key, where the ship stops turning/tracking as soon as one lets go of the key? Or does it act like landing or boarding keys where one taps the key and it keeps being active until one taps something to break it?

I'm not sure which one I'd prefer, although I lean towards the former rather than the latter.

ChamEV:
It acts like a movement key, it only works while you're holding it down.

source/PreferencesPanel.cpp
```diff
@@ -332,7 +333,9 @@ void PreferencesPanel::DrawControls()
 		Command::FIGHT,
 		Command::GATHER,
 		Command::HOLD,
-		Command::AMMO
+		Command::AMMO,
+		Command::NONE,
```
jafdy: This line appears extraneous (although I may be wrong).

source/AI.cpp
```diff
@@ -1789,7 +1789,7 @@ void AI::PrepareForHyperspace(Ship &ship, Command &command)
 }
 
 
-
```
jafdy: Whitespace changes.

jafdy:
How is this different to the Automatic Aiming option, and how does it do a better job to justify the extra key it wants to use.

Zitchas:
I think the main difference between this and the automatic aiming option is the fact that it is much more fine tuned: If it only works when the player is holding down the relevant button, so it is only working when called upon instead of being an "always on / always off" toggle.

I know I, personally, would like to be able to use an auto-aiming function some of the time, but I find it annoying most of the time. Since most of the time trumps sometimes, I leave the option off and just make do without. So I'd personally use this a fair bit.

source/PreferencesPanel.cpp
ChamEV: NONE tells the interface not to draw a box for the category line, it corresponds with the CATEGORIES array just above.

source/AI.cpp
ChamEV: The three remaining white space changes align the tabs with the current indentation of the surrounding text.  In the case shown above, there was an extra tab when there was no indentation above or below (between functions).

By default code:blocks was stripping white space at the ends of lines from everything and I had to go back and fix all that so it wouldn't be distracting.  I left the 3 specific white space changes in as they aligned the tab of each line to the current indent used by the rest of the file.  (Code blocks also defaulted to windows line endings on everything, but that was easier to fix).

ChamEV:
> 
> 
> How is this different to the Automatic Aiming option, and how does it do a better job to justify the extra key it wants to use.

Automatic aiming only works while firing, forcing the player to turn towards a target at times when they may not want to as opposed to firing turrets and flying by.

In cases where there's a specific ship, perhaps disabled on the edge of the battle, you can select it and find which direction it's in easily.  If there's a ton of disabled other ships that you don't want (small fighters vs the big heavy cruiser that you want).

warp-core:
@ChamEV are you still interested in getting this merged? Could you bring it up to date and resolve the conflicts? I'd like to test it, but it doesn't build.

source/AI.cpp
```diff
@@ -3348,8 +3348,25 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 		if(keyHeld.Has(AutopilotCancelKeys()))
 			keyStuck = keyHeld;
 	}
+	
 	bool shouldAutoAim = false;
-	if(Preferences::Has("Automatic aiming") && !command.Turn() && !ship.IsBoarding()
+	if( keyHeld.Has(Command::AUTOSTEER) && !command.Turn() && !ship.IsBoarding()
```
warp-core, line 3353:
```suggestion
	if(keyHeld.Has(Command::AUTOSTEER) && !command.Turn() && !ship.IsBoarding()
```

NickGrease:
I posted about this back in 2016 (#1258) which was a dup from 2015 (#166). I really hope this gets merged, it's been a long time coming.

Amazinite:

UnorderedSigh:
I have resurrected this in #8508

PR #4356, opened 2019-06-11
Remnant: Pelican Rebalance

Zitchas:
The Pelican was a good start, but fell short of its intended role as a fast "blockade runner" style transport and had a number of other flaws:
- It was pointed out by Zarvox and Titanicguy that the Pelican actually seems to go against the Remnant description of devoting more power to flight than to weapons.
- It was pointed out that its purchase price is 8M more expensive than the Arach Hulk, whereas the hulk has 82 tons more outfit space, 19 more engine space, and 239 more cargo space, and has more health too. While the Pelican does have a bunch of built-in things, it doesn't really seem like 8M worth. 
- Compared to the Sojourner it also suffers: The Sojourner is faster, significantly tougher, and has 130 tons more outfit space, 27 more engine space, and 49 more bunks. But is 6M cheaper. And only slightly smaller cargo hold.

The following changes are intended to bring the Pelican in line with its intended role: To be able to run or snatch cargo through blockades and combat areas by relying primarily on speed and endurance.

Note: The Pelican retains its thrasher turret, mostly by virtue of the fact that the Remnant have no alternative turrets right now. If they had an inhibitor turret, that would be replacing the thrasher turret. Slowing down pursuers is much more appropriate for the Pelican than a close-quarters thrasher. On the same note, if the Point Defense Turrets actually targeted fighters/drones, the thrasher turret would be gone without needing to be replaced by anything.

Changes:
- Removed both Thrasher cannons
- Replaced the Forge thruster with a Smelter
- Adjusted engine capacity up to 144 to accommodate this (and to provide a small amount of room for customizing and a possible future afterburner)
- Added one crystal capacitor to accommodate engine draw
- Added one thermoelectric cooler to accommodate heat
- increased outfit space to 470 (that is +32 tons)
- Changed the description to include mention of sensory capabilities and mention of fast-logistics.


data/remnant ships.txt
```diff
@@ -410,7 +409,7 @@ ship "Pelican"
 	explode "large explosion" 20
 	explode "huge explosion" 5
 	"final explode" "final explosion large" 1
-	description "The Pelican is closely based on the original semi-organic materials the Remnant found in the Ember Waste. While they are not technically alive, their internal mechanisms and method of construction bear such a close resemblance to living organisms that they are often considered as such. These ships serve as the backbone of Remnant salvage efforts, as well as their resource distribution network."
+	description "The Pelican is closely based on the original semi-organic materials the Remnant found in the Ember Waste. While they are not technically alive, their internal mechanisms and method of construction bear such a close resemblance to living organisms that they are often considered as such. One key feature of this living resemblence is the innate sensory capabilities they posses; akin to sensor suites less advanced ships require for detailed scanning. These ships serve as the backbone of Remnant bulk salvage efforts, as well as their fast-logistics distribution network."
```
Pointedstick: Change the `;` to a `,` (semicolons separate distinct sentences but `akin to sensor suites less advanced ships require for detailed scanning` is a fragment)

tehhowch:
Even if this PR's content is clean and ready for merge, this change should not merge until we've released v0.9.10 to Steam.

There are still a few open bugs blocking that release, so the priority should be in deciding and making the appropriate code fixes.

Issue #4357, opened 2019-06-11
Missing single quote in remnant mission


jostephd:
In https://github.com/endless-sky/endless-sky/blob/a0322e99409ab4ee01b9364bad4d0cf29460b094/data/remnant%20missions.txt#L687 there are unbalanced single quotes.

Zitchas:
Thanks for pointing this out!

PR #4359, opened 2019-06-11
Remnant-Fix: A missing quotation mark

Zitchas:
Adds a missing `'` to line 687 in the remnant missions.txt file.

Thanks to @jostephd for spotting this and reporting it in issue #4357 

Thanks to the Writing Commune led by Moonbeam for the advice on correcting it.

### DIFF
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -684,7 +684,7 @@ mission "Remnant: Tech Retrieval"
 			choice
 				`	"Okay, I will see what I can find."`
 					goto find
-				`	"Could you elaborate on 'how fast they can travel in deep space?"`
+				`	"Could you elaborate on 'how fast they can travel in deep space?'"`
 				`	"No, I don't want to get involved in this."`
 					goto nothelp
 			`	She considers her thoughts briefly, then chants, "How fast a ship can move through an area of deep space basically depends on how fast they can collect fuel. Ramscoops were common at the time of the exodus, and if humanity has developed better ramscoops or other means of generating fuel, these would significantly improve their speed at exploring or crossing large volumes of space."`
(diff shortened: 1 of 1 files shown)


Zitchas:
OK, checked for an online source, and found: 
https://data.grammarbook.com/blog/commas/quoting-a-question-within-a-question/

According to which, the correct order is single quotation mark, question mark, double quotation mark.

UnciaPrima:
Why is there a quotation mark in the first place? 

"Could you elaborate on how fast they can travel in deep space?" without any single quotation mark seems better to me.

MCOfficer:
Because "how fast they can travel in deep space" is not just any topic, it's a direct quote from her previous statement.

Zitchas:
For greater context, here's the previous statement that the captain is replying to:

"To the point," her tone shifts to a more martial cadence. "We want to know how humanity has advanced in the past five and a half centuries. If you found us, I am sure it is only a matter of time until the Alphas, their servants, or others intent on galactic domination discover us. We need to know what sort of weapons they will be wielding, how their ships will be protected, and how fast they can travel in deep space."`

The response choices:
`	"Okay, I will see what I can find."`
`	"Could you elaborate on 'how fast they can travel in deep space'?"`
`	"No, I don't want to get involved in this."`

Zitchas:
I could expand that option to something like:

`"Could you elaborate on what you mean by 'how fast they can travel in deep space'?"`

It doesn't get around the issue of the stacked quotation marks and punctuation at the end, though.

jafdy:
`"When you say 'how fast they can travel in deep space,' what do you mean by that?"` Is the kind of statement I would use when I need to get information from someone, it is very precise way of saying it.

PR #4347, opened 2019-06-08
Turning the Bactrian into actual City Ship

Zitchas:
This PR is not intended for near-term merging, but is a proposal for shifting the stats of the Bactrian to more closely reflect the way it is described. I'm expecting a fair bit of discussion, and I suspect these stats aren't final. I'm basing this on the assumption that the description is cannon, and the stats are subject to debate.

> "The Lionheart Bactrian is the last of the great city-ships, a design hearkening back to the early days of space colonization when a long-distance vessel needed to be a self-contained world, able to survive for weeks at a time without encountering an inhabited planet. It is not only a freighter, but a carrier, and a capable warship either at short range or at a distance."

> Naturally, this versatility also makes it extremely expensive, and the Bactrian is normally only for sale to citizens who have been vetted by the local government of the Deep."

So, based on this description, it is a ship built for long haul traveling that is capable of defending itself and being self-supporting, as well as all-round versatility.

Breaking it down: 
1) "City ship from the early days of colonization" sounds like a transport to me.
2) Long distance travel. This implies the need to cover distances, not agility.
3) Both a freighter and a carrier. Good cargo capacity and fighter bays.
4) Capable warship. Not awesome warship, capable. 

This leads me to the following changes:
1) Changed the type from Heavy Warship to Transport.
~~2) Replaced the Steering with an X4200 to be in balance with the thrusters, which is more in line with a ship built to cover long distances instead of for agility.~~
3) The Bactrian is already a great hauler and can carry 3 fighters, so seems good as is for this.
4) IIt has more turrets than any other human ship except for the Protector. This fits, since turrets are quite appropriate for large non-agile things like cities and space stations. However, it *also* has 4 gun ports, which doesn't fit. Guns require agility and precision to aim, and that is something a city generally lacks. My first inclination is to remove all the gun ports entirely, but in ES missiles require gun ports, and having a pair of missile launchers for anti-fighter defense is reasonable. So, the torpedoes are removed, and the number of gun ports is decreased to 2.

As a result, they are still a very capable warship, but it reduces its clear combat superiority a touch so that dedicated warships like the Falcon and Leviathan at least outclass it in terms of guns, even if they can't surpass it in anything else. It also means that Dreadnoughts are a bit more significant, since the Bactrians still outclass them by a turret, so giving the dreadnought a 2 cannon advantage I think is important in setting the Dreadnought apart as a meaningful upgrade in the combat sense.

[...]

PR #4369, opened 2019-06-15
Correcting the Mule as a Heavy Freighter

Zitchas:
This update corrects the Mule:
- Changed classification to Heavy Freighter
- Changed description to indicate that the Republic officials were tempted to classify it as a warship instead of actually *being* classified as a warship.

Just to be clear, NO STATS ARE CHANGED, only the label and a slight wording change in the description

These changes bring the mule in line with the Behemoth (which has more hull, more shields, and more weapon space than the mule) and the Bulk Freighter (which also has significantly more hull and somewhat less shields (but still more combined health) and as much weapon capacity as some medium warships and more than most.). To a lesser extent, this also brings it in line with the Argosy, which has more hull, shields, and weapon space than many light warships.

One of the reasons for this change is to increase the number of options for players who wish to play as merchants who happen to be good at combat. It is hard to pretend to be a merchant when one is flying a heavy warship, no matter the justifications. Thus this change reclassifies the mule as a heavy freighter, and modifies the description to give it some ambiguity: It is more interesting to look at the mule and see how regulators might have classified it as a warship; rather than to look at it and see a subpar warship that only barely made it into the category.

Also, while what the regulators term a warship may have been relevant back when licenses for warships were a thing, now that they are not, keeping the mule as a warship simply muddies the waters and hides a ship that merchant players may wish to fly in a section where they might not look when comparing freighter capacity.


Zitchas:
This change also facilitates #4347 .

Nescio0:
Why not a transport? Both its number of free bunks (43-6=37) and the bunks to crew ratio (43/6) are superior to several other transports.

Zitchas:
Primarily because its description revolves around being used as a freighter. Nowhere does it mention that it was designed for or intended to be a transport. It just so happens to be decent at being one.

In terms of "what ships are classified as" I am concerned primarily with the lore of the game and what they are intended to be.

Nescio0:
To clarify, I'm not opposed to the change per se, I'm merely wondering as to why.
You state it's because of lore, however, the lore (description) makes it immediately clear the ship is classed as a warship, because of its shields and weapon capacity.
Classes are arbitrary, but to me this patch seems to change something slightly problematic into something else slightly problematic, i.e. no real improvement, so why change?

Zitchas:
The main reason is that the original intent behind the classification was so that a ship with a certain amount of combat capability would be locked behind a license that had to be purchased, like EVN and EVO did. Just to be clear, in EVN/EVO these were purchased licenses that basically amounted to "you can't buy a warship until you pay the license people 150k for a warship permit" (I don't know what it was actually like in ES, though.)

Subsequently, licenses were deemed too annoying, and were removed. Now that there is no reason to restrict ships based on combat capability within "warship" licenses, I feel that it would be more appropriate to sort ships based on their intended purpose, thus allowing the various warship categories to actually be just those ships that were specifically designed to be warships, and not clutter it up with ships that aren't really warships but are classified that way because they have a bit too much shield/hull/guns/whatever. 

This change also makes it easier for new players to understand the ships, as they can peruse ships based on what they are looking for (say, a freighter), and can then note that this ship, which is primarily intended to be a freighter, also happens to have very good transport and combat capabilities. I feel this is more intuitive than forcing players to focus on the warships and try to pick out which one makes the best freighter.

After all, ES is largely a space-combat game. Combat capabilities are relevant for *every* ship. As such, we should sort the main categories based on their intended purpose, realizing that every category is going to have at least a few and possibly many ships that have some degree of combat capabilities.

Pointedstick:
Frankly I would also support reducing the Mule's combat potential a bit, like halving the number of gun ports or making it a bit less maneuverable. That would make the classification change a bit more sensible.

Of course that's a larger more controversial change. :)

petervdmeer:
> Of course that's a larger more controversial change. :)

Small steps, one at a time; let's just first merge the current change here. It is just sufficiently non-controversial to have a chance to get merged somewhat quickly.

finite-galaxy:
In my version I removed one gun port from the Bactrian (from 4 to 3) and Mule (from 2 to 1).

(Also, because freighter/transport/warship (and what about carriers?) is rather arbitrary, I replaced ship categories and based them solely on total mall (hull mass + outfit size + cargo space). I'm not implying you should do the same, though.)

finite-galaxy:
In my version I removed one gun port from the Bactrian (from 4 to 3) and Mule (from 2 to 1).

(Also, because freighter/transport/warship (and what about carriers?) is rather arbitrary, I replaced ship categories and based them solely on total ship mass (hull mass + outfit size + cargo space). I'm not implying you should do the same, though.)

Tadrix:
The Mule is the best medium warship (installable firepower and durability-wise), yet faster than heavy freighters and with a small crew; its cargo space also isn't that big. It shouldn't be classed as a heavy. If anything, the Bastion's closer to that.

Zitchas:
I'm happy with the content of the discussion, and I continue to think it is a good change.

I personally like the idea of reducing both the Mule's and Bactrian's combat potential slightly, but that's for another merge & discussion. For now, I'm good.

Tadrix:
I'd say hold off this change till there's a need for Medium Freighters, and then class the Mule as them. It doesn't have enough of cargo capacity to be a heavy.

(For the purposes of Pointed being merge-happy, consider it a hard objection.)

Amazinite:
The Bastion also has the same justification in its description for why it's classed as a warship instead of a freighter. Given that both the Mule and Bastion mention this, it's clear that this is part of the lore of how ships are categorized, at least in human space, and I don't agree with retconning that just because. 

I can't really get behind the reasoning for this change.
Re: The Behemoth being bigger than the Mule. The Behemoth has less HP than the Vanguard, which is the weakest heavy warship. This harkens back to the idea of making the Mod Argosy a medium warship; if you make the change, then it just moves from becoming the biggest heavy freighter to the smallest heavy warship, just as moving the Mod Argosy would change it from the biggest light warship to the smallest medium. Ignoring the Bactrian, the Behemoth has nearly 5x more cargo than any human heavy warship, which fits with the current idea of ability over purpose in categorization.
Re: Number of options for merchant players. Your loss if you aren't checking the other categories for potential ships to use. I don't see the use of the Mule as immersion breaking in this regard given that its description even highlights this oddity about it. If I was someone who only wanted to play with freighters but bought myself a Mule, I'd realize that I was teetering on the edge of what it means to be in a freighter. Same goes with the Bactrian. A freighter wants lots of cargo but also wants to be well armed to defend itself, and the Mule and Bactrian get into the realm of "how well armed can a freighter be before it's no longer a freighter?" 
Re: Category licenses aren't a thing anymore so it should be changed. While category licenses were removed for gameplay purposes, it's important to note that they are still mentioned in game as a reason for the FW leaving the Republic, so they still have some bearing on the lore. But even if the reason that the Bastion, Mule, and Bactrian were originally classed as warships was to keep them behind more expensive licenses, we still run into the purpose versus ability argument as to if we should change the categories around. In terms of ability, the Bactrian, Mule, and Bastion are closer to the warships that they are categorized as, even if their purposes lean more toward being freighters/transports/what-have-you.

Zitchas:
Given the feelings and support that this issue has, I would like to propose that we (I) close this issue and at some point in the near future we have a broader discussion on how exactly ships should be classified, and then to a full pass through the whole set of ships. Humanity in particular, but there's a few oddities among the aliens too.

Issue #4308, opened 2019-05-30
Auto-testers / game-logic regression testers


petervdmeer:
**Is your feature request related to a problem? Please describe.**
I only found manual test-procedures for testing new Endless Sky features. Automatic testers for game-logic will probably help speed up feature development and help to avoid (regression) bugs.

**Related Issue Links**
Features/issues #861, #1624, #3639, #4101, #4305, #4307 and #4283 are examples that can benefit from automated testing.

**Describe the solution you'd like**
For basic test-data, I would like to re-use game-data loading and save-game loading functions
- Game-data loading can provide loading of a simple test-environment (for example a system with just 2 starts and 2 types of ships)
- Game-data loading can also provide "test-missions" with exactly the right content for a specific test.
- Save-game-data can provide a starting-state for tests (determines which ships a "test-player" has at start of the testing and which missions are active).

Besides this we would also need the ability to verify conditions (asserts) and we would need a way to move the test forward.

Moving the test forward is normally done by explicit test-steps and most test-systems use a level just below the GUI (so not a list of key-presses, but explicit calls to things like outfitter actions and commodity buy/sell actions). The idea there is that minor GUI changes should not affect game-logic-testcases.

**Describe alternatives you've considered**
An alternative for explicit test-steps that is possible within Endless Sky is to build some sort of "automatic mission resolver" and use mission data-files (with some small additions) to fully describe test-cases.

**Additional context**
Here is an example of how tests can look if we choose to use explicit test-steps in seperate test-files:

file: tests/auto-test/cargo_outfit_overload.txt

	test "Cargo: Simple commodoties transport"
		status "Active"
		game-data ./two_star_system
		starting-save ./two_star_two_barges
		test-sequence
			assert on-planet "First Star" "First Planet"
			assert in-cargo commodities "Electronics" 0
			commodities buy "Electronics" 100
			assert in-cargo commodities "Electronics" 100
			set-destination "Second Star" "Second Planet"
			depart
			assert on-planet "Second Star" "Second Planet"
			assert in-cargo commodoties "Electronics" 100
			commodities sell "Electronics" 100
			assert in-cargo commodoties "Electronics" 0

	test "Cargo: Outfit not fitting in single cargo-hold"
		status "Known Failure"
		bug-id "None yet"
		game-data ./two_star_system
		starting-save ./two_star_two_barges
		test-sequence
			assert on-planet "First Star" "First Planet"
			outfits buy-to-cargo "Large System Core" OK
[...]

PR #4370, opened 2019-06-15
feat: Add initial framework for game-logic auto-testing

petervdmeer:
This adds the first part of the framework for auto-testing as discussed in #4308.

The first upload has only 1 test (the empty testcase) succeeding.
If this is a direction that we want to go, then I can implement the test-steps for loading, launching and landing and get all 3 of the framework tests working.


tehhowch:
Haven't tried doing anything with your code, just some notes:

 - I don't think the `delete` is needed, as the tests are being managed by `GameData`
   - probably there is a solution that involves just iterating through a `const` test step vector
 - use the `tests/` directory for test data.
   - I'm fine with placing test-specific source in a subdirectory of `source` (this will involve changing some build scripts so I def understand not doing it right away)
 - we are a (sorry) `using namespace std` repo
 - why `auto-test` instead of `test`? Do you feel we'll be adding tests that run in some other manner? I don't foresee a conflict where we want to use `test` as a root level datanode that is doing something other than defining a test (although does one ever foresee these things?)
 - not something we need for a first pass, but eventually being able to run tests headlessly (i.e. no sdl window initialization) would be great. Parallel would be good as well, but that's more or less something we'd do externally rather than internally.

petervdmeer:
>     * I don't think the `delete` is needed, as the tests are being managed by `GameData`
>       * probably there is a solution that involves just iterating through a `const` test step vector

Yes using an iterator is nicer. I should probably add a 3rd class AutoTestRunner which contains the iterator and other dynamic data to really allow the AutoTester and the TestStep themselves to be read-only.

>     * use the `tests/` directory for test data.
>       * I'm fine with placing test-specific source in a subdirectory of `source` (this will involve changing some build scripts so I def understand not doing it right away)

I was considering to "embed" the test-savegames inside the test itself, but that will be quite some additional work and the added value of such embedding is not too high at this moment. So such embedding is then also not for now.

>     * we are a (sorry) `using namespace std` repo

No problem. This is an easy fix. Just uploaded a commit to change the cpp files to use namespace std.

>     * why `auto-test` instead of `test`? Do you feel we'll be adding tests that run in some other manner? I don't foresee a conflict where we want to use `test` as a root level datanode that is doing something other than defining a test (although does one ever foresee these things?)

I didn't really think about this one. If "test" is better, then I will just change the code and datafiles to use test. "test-sequence" is also a bit long. Would you also prefer to change that to just "sequence" (since it is already inside a test)?

>     * not something we need for a first pass, but eventually being able to run tests headlessly (i.e. no sdl window initialization) would be great. Parallel would be good as well, but that's more or less something we'd do externally rather than internally.

Agreed, those are very nice features to add somewhere in the future (when the test-mechanism is in active use). The only option I wanted to add for now is to only set delays when --debug is active (tests are too fast to follow without the delays).

petervdmeer:
The loading and the depart test-steps are now working! Code can still use some cleanup, but the mechanism is at least doing something.

petervdmeer:
@tehhowch The delete is replaced by an iterator and the keyword auto-test is replaced by test. I also performed some code-clean-up. Can you review if the code is good enough for merging?

petervdmeer:
Added a small change towards getting the Landing command to work.

petervdmeer:
@endless-sky Both Tehhowch and I think that it's a good idea to add auto-testers, but this change is getting quite big, so I think it's good to also hear your opinion about this. Do you want auto-testers for the game? And do you think that this branch is going in the right direction? (And if not, then what would you like to see changed?)

source/Test.cpp
```diff
@@ -0,0 +1,100 @@
...
+				status = STATUS_ACTIVE;
+			else if (child.Token(1) == "Known Failure")
+				status = STATUS_KNOWN_FAILURE;
+			else if (child.Token(1) == "Missing Feature")
+				status = STATUS_MISSING_FEATURE;
+		}
+		else if (child.Token(0) == "sequence")
+			for (const DataNode &seqChild : child)
+			{
+				testSteps.push_back(new TestStep());
```
tehhowch: Do we need to use `new`? For every `new` there must be a corresponding `delete` lest we leak memory. The std::vector destructor will not call delete on a raw pointer for you.

I would imagine there is a way we can use emplace_back with the node and a TestStep constructor that takes a DataNode to simplify this and the next line.

source/TestData.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "TestData.h"
+#include <string>
+
+using namespace std;
+
+
+
+TestData::TestData()
+{
+	dataSetName = "";
```
tehhowch: These are again the defaults for the datatypes.

source/Test.cpp
```diff
@@ -0,0 +1,100 @@
...
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+
+#include "DataNode.h"
+#include "PlayerInfo.h"
+#include "Test.h"
+#include "TestStep.h"
+#include "UI.h"
+#include <string>
```
tehhowch: You already have `<string>` in the header (and these includes are not ordered properly)

source/TestData.cpp
```diff
@@ -0,0 +1,110 @@
...
+bool TestData::Inject() const
+{
+	// We only know how to inject savegame data for now.
+	if ((dataSetType != TestData::SAVEGAME) || (dataSetName.empty()) || (sourceDataFile.empty()))
+		return false;
+
+	// Open the source-file and scan until we find the test-data
+	// Then scan for the contents keyword
+	// Then write out the complete contents to the target file
+	DataFile sourceData(sourceDataFile);
```
tehhowch: Use the `{ }` syntax where possible over `( )`

source/Test.h
```diff
@@ -0,0 +1,46 @@
...
+
+
+// Class representing a single test.
+class Test {
+public:
+	Test();
+	virtual ~Test();
+	
+	// Status indicators for the test that we selected (if any).
+	static const int STATUS_ACTIVE = 0;
```
tehhowch: Use a named enum

source/Test.h
```diff
@@ -0,0 +1,46 @@
...
+	static const int STATUS_KNOWN_FAILURE = 1;
+	static const int STATUS_MISSING_FEATURE = 2;
+	
+	virtual std::string Name() const;
```
tehhowch: Is this a base class? If so, where are the deriving classes? What would they look like (there should be at least two otherwise it doesn't make sense to have a base and a derived class).

Also, would these deriving classes really be expected to specialize things like the `Name` getter?

source/TestRunner.cpp
```diff
@@ -0,0 +1,90 @@
...
+TestRunner::~TestRunner()
+{
+}
+
+
+
+string TestRunner::ConditionsText(PlayerInfo &player)
+{
+	string conditions = "";
+	for (auto condition : player.Conditions())
```
tehhowch: Should be ref

source/TestRunner.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+	TestStep* testStep = testSteps[stepToRun];
+
+	int testResult = testStep->DoStep(stepAction, menuPanels, gamePanels, player);
+	switch (testResult)
+	{
+		case TestStep::RESULT_DONE:
+			// Test-step is done. Start with the first action of the next
+			// step next time this function gets called.
+			stepToRun++;
```
tehhowch: Prefer prefix ++ to postfix.

source/TestStep.cpp
```diff
@@ -0,0 +1,287 @@
...
+	// If we can cast the topPanel from the gamePanels to planetpanel,
+	// then we have landed and are on a planet.
+	return dynamic_cast<PlanetPanel*>(topPanel);
+}
+
+
+
+void TestStep::Load(const DataNode &node)
+{
+	if (node.Token(0) == "load" and node.Size() > 1)
```
tehhowch: Do not use `and`

source/TestStep.cpp
```diff
@@ -0,0 +1,287 @@
...
+	{
+		stepType = LOAD_GAME;
+		filePathOrName = node.Token(1);
+	}
+	else if (node.Token(0) == "assert")
```
tehhowch: Reorder the branches to operate with specific logic: first keys that require no second token, then those should have a second, then those that should have a third, etc. 
In this manner you can add a branch between the types to catch all uses that should have something but don't, and avoid needing to continually check the node size

source/TestStep.cpp
```diff
@@ -0,0 +1,287 @@
...
+		case TestStep::LAND:
+			// If the player/game menu is active, then we are not in a state
+			// where land makes sense.
+			if (PlayerMenuIsActive(menuPanels))
+				return RESULT_FAIL;
+			// If we are still flying around, then we are not on a planet.
+			if (PlayerIsFlyingAround(menuPanels, gamePanels, player))
+			{
+				Ship * playerFlagShip = player.Flagship();
+				if (playerFlagShip == nullptr)
```
tehhowch: `!playerFlagship`

source/TestStep.cpp
```diff
@@ -0,0 +1,287 @@
...
+				player.SetTestMode();
+				// Actual load succeeded. Allow game to adopt to new
+				// situation and then continue with enter/pilot step.
+				return RESULT_NEXTACTION;
+			}
+			else if (stepAction == 1)
+			{
+				// Clear the menu entries and go to main game screen
+				if (! menuPanels.IsEmpty())
+					menuPanels.Pop(menuPanels.Top().get());
```
tehhowch: We should send keystrokes / commands that the player actions perform where possible.

source/TestStep.h
```diff
@@ -0,0 +1,68 @@
...
+
+// Class representing a single step in a test
+class TestStep {
+public:
+	TestStep();
+	virtual ~TestStep();
+
+	// TODO: rename to LOAD_GAME_INLINE (and allow other loaders later?)
+	// Switch to game loading from inline in the test (from a child datanode)
+	static const int LOAD_GAME = 1;
```
tehhowch: Named enum

source/main.cpp
```diff
@@ -68,9 +70,12 @@ int main(int argc, char *argv[])
 	if(argc > 1)
 		InitConsole();
 #endif
+	TestRunner* testRunner = NULL;
```
tehhowch: nullptr

source/main.cpp
```diff
@@ -98,8 +105,22 @@ int main(int argc, char *argv[])
 		if(!GameData::BeginLoad(argv))
 			return 0;
 		
+		if (! testToRun.empty())
+		{
+			testRunner = new TestRunner((GameData::Tests()).Get(testToRun));
```
tehhowch: Where is the call to `delete`? It needs to be called in all cases when `new` is used, including when we are exiting due to an exception.

source/Test.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+using namespace std;
+
+
+
+Test::Test()
+{
+	name = "";
```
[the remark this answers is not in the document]
petervdmeer: Good point, I'll remove the initializer.

source/Test.cpp
```diff
@@ -0,0 +1,100 @@
...
+	{
+		node.PrintTrace("No name specified for test");
+		return;
+	}
+	if (node.Token(0) != "test")
+	{
+		node.PrintTrace("Non-test found in test parsing");
+		return;
+	}
+	name = node.Token(1);
```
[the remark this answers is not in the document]
petervdmeer: I don't see a reason why anybody would want to override a test (in a plugin or otherwise), so `name` should indeed be unique.

source/Test.cpp
petervdmeer: Yes, emplace_back is probably better here. I'm also considering to move TestStep to become a nested class in Test. It looks like the TestStep class doesn't really need to be a stand-alone class.

source/Test.cpp
```diff
@@ -0,0 +1,100 @@
...
+			{
+				testSteps.push_back(new TestStep());
+				(testSteps.back())->Load(seqChild);
+			}
+	}
+}
+
+
+
+string Test::Name() const
```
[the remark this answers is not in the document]
petervdmeer: Good point. I'll change this.

source/Test.h
petervdmeer: Yes, that is nicer. I'll change this.

source/Test.h
petervdmeer: Yes, initially I expected this to become a base class with multiple derived classes, but that was before I moved out some functionality to TestSteps. I will remove the virtual keywords, since I don't expect to derive any classes from this class anymore.

source/TestData.cpp
petervdmeer: I'll remove the unnecessary initialization.

source/TestRunner.cpp
petervdmeer: I'll add const.

source/TestRunner.cpp
petervdmeer: ok

source/TestStep.cpp
petervdmeer: Will change to `&&`.

source/TestStep.cpp
petervdmeer: Good idea. I think I have seen this done in other parts of the code as well.

source/TestStep.cpp
petervdmeer: Will fix.

source/TestStep.cpp
petervdmeer: Yes; I already tested #4715 in combination with this PR today to see that commands can be given from the TestRunner to the Engine/Fleet/flagShip using the same commands as that players would send.
A next step is to do the same for planetary actions (feature-request #4530 is related to that) and loading/saving or just generic keyboard input should also work this way where possible.

The in-flight and on-planet actions will take quite a bit of attention first, so I'll add a `//TODO` about improving this part of the code to make sure that I don't forget this.

source/TestStep.h
petervdmeer: ok

source/main.cpp
petervdmeer: ok

source/main.cpp
petervdmeer: Agreed, I still need to add a delete or solve this in a different way.

source/TestData.cpp
petervdmeer: I'm not sure what you mean here, but I'll check if there is a constructor type that uses an `{}` syntax.

petervdmeer:
@tehhowch Thanks for your extensive review.

> Why all the virtual methods?

I initially expected to need them, but they are not needed. I'll remove them.

> if you don't need to customize it, dont write the constructor or destructor. Otherwise you must also define or = default the remaining constructors.

Good point. I'll remove them.

> when you do, use default initialization ( define the non-default values in the header) and pass constructor args in the ): ... { section where possible.

Ok, I think I saw some examples in the ES code-base already.

> avoid using new unless it isn't possible to do so. If you must, use a smarter pointer e.g. std::unique_ptr so that ownership and delete responsibility is explicit.

Yes, emplace_back of full objects makes memory management a lot simpler.

source/Files.cpp
```diff
@@ -142,6 +149,11 @@ void Files::Init(const char * const *argv)
 	images = resources + "images/";
 	sounds = resources + "sounds/";
 	
+	// If we have no explicit config-dir given and we are using testers
```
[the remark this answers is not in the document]
petervdmeer: I agree, that is a much better solution for the config directory than that I implemented here. I'll update the code not to allow test-runs when no explicit config directory is given.

For resources I would like to keep using the default set of resources if no alternative set is given.

source/GameData.cpp
```diff
@@ -136,6 +141,8 @@ bool GameData::BeginLoad(const char * const *argv)
 				printShips = true;
 			if(arg == "-w" || arg == "--weapons")
 				printWeapons = true;
+			if(arg == "-l" || arg == "--list-tests")
```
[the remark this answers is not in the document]
petervdmeer: Changed to long parameters only.

source/PlayerInfo.cpp
```diff
@@ -590,6 +590,14 @@ void PlayerInfo::SetSystem(const System *system)
 
 
 
+// Set test-mode (which prevents creation of automatic savegames
```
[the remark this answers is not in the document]
petervdmeer: Removed test-mode

source/Test.cpp
petervdmeer: Changed to emplace_back

source/TestStep.cpp
petervdmeer: I handled this comment by adding a `//TODO` item for now.

source/main.cpp
petervdmeer: I fixed this missing delete by removing the TestRunner class completely. The test-status is now tracked through a TestContext class.

source/TestData.cpp
petervdmeer: Done

petervdmeer:
I added a workaround for #4736 by letting the launch test-step wait for zoom to become 1.0.

petervdmeer:
The launch/land test now also works (when combined with #4715 also merged in the testing branch).

petervdmeer:
Merged master to combine #4715 into this branch.

source/main.cpp
```diff
@@ -68,9 +69,12 @@ int main(int argc, char *argv[])
 	if(argc > 1)
 		InitConsole();
 #endif
+	Test::Context testContext;
```
tehhowch: Does this need to have the same scope as the variable that says which test is run?

tests/run_autotests.sh
```diff
@@ -0,0 +1,68 @@
...
+
+echo " Setting up environment and retrieving test-case list"
+mkdir -p "${ES_CONFIG_PATH}"
+mkdir -p "${ES_SAVES_PATH}"
+TESTS=$("${ES_EXEC_PATH}" --tests --resources "${RESOURCES}" --config "${ES_CONFIG_PATH}")
+TESTS_OK=$(echo "${TESTS}" | grep -e "ACTIVE$" | cut -d$'\t' -f1)
+TESTS_NOK=$(echo "${TESTS}" | grep -e "KNOWN FAILURE$" -e "MISSING FEATURE$" | cut -d$'\t' -f1)
+echo ""
+
+echo "Tests not to execute (known failure or missing feature):"
```
tehhowch: If a test is a known failure, we should still run it. If it doesn't fail, then that is useful information.

petervdmeer: Every failing test currently gives a pop-up, which stops automatic test execution.
I added a TODO-item for now to make sure this doesn't get forgotten.

source/TestData.h
```diff
@@ -0,0 +1,41 @@
...
+class TestData {
+public:
+	std::string Name();
+	void Load(const DataNode &node, const std::string sourceDataFilePath);
+	// Function to inject the test-data into the game or into the games
+	// environment. Savegames would be written as file.
+	bool Inject() const;
+
+	// Types of datafiles that can be stored.
+	static const int SAVEGAME = 1;
```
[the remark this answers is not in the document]
petervdmeer: Agreed, fixed

source/TestData.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+string TestData::Name()
+{
+	return dataSetName;
+}
+
+
+
+// Loader to load the generic test-data entry
+void TestData::Load(const DataNode &node, const string sourceDataFilePath)
```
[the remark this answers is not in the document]
petervdmeer: Done

source/TestData.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+#include "DataFile.h"
+#include "DataWriter.h"
+#include "Files.h"
+
+using namespace std;
+
+
+
+string TestData::Name()
```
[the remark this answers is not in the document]
petervdmeer: Done

source/Test.cpp
```diff
@@ -0,0 +1,370 @@
...
+	if(context.stepToRun >= testSteps.size())
+	{
+		//Done, no failures, exit the game with exitcode success.
+		menuPanels.Quit();
+		return;
+	}
+
+	const TestStep &testStep = testSteps[context.stepToRun];
+
+	int testResult = testStep.DoStep(context.stepAction, menuPanels, gamePanels, player);
```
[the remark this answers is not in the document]
petervdmeer: Good point. Changed to the enum type.

source/Command.h
```diff
@@ -13,6 +13,8 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #ifndef COMMAND_H_
 #define COMMAND_H_
 
+#include "DataNode.h"
```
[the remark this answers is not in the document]
petervdmeer: Yes, that's nice. Changed to forward declaration (also for some other includes).

source/Command.cpp
```diff
@@ -215,6 +214,71 @@ bool Command::HasConflict() const
 
 
 
+// Load this command from an input file (for testing or scripted missions)
+void Command::Load(const DataNode &node)
+{
+	int i = 1;
+	while(node.Size() > i)
```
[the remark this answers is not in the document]
petervdmeer: True, used the standard pattern.

source/main.cpp
petervdmeer: No, didn't need to. Moved one scope smaller.

petervdmeer:
Rebased to resolve merge-conflicts.

petervdmeer:
Closing this PR; split-off pre-requisite PR https://github.com/endless-sky/endless-sky/pull/4998 is not merged yet.
I plan to re-open this PR (or backport a new PR from ESCW if the version there is newer by then) when the pre-requisite PR gets merged.

petervdmeer:
Re-opening pull request; PR #4998 got merged.
Merged this PR with the latest updates from #4998. Also ported the latest auto-tester changes from ES-Stories back into this PR.

petervdmeer:
Converted to draft; one of the testcases is failing (probably due to a mistake in my last update). Needs to be fixed before merging.

source/Command.cpp
```diff
@@ -221,6 +221,70 @@ bool Command::HasConflict() const
...
 
 
+// Load this command from an input file (for testing or scripted missions).
+void Command::Load(const DataNode &node)
+{
+	for(int i = 1; i < node.Size(); ++i)
+	{
+		const string &input = node.Token(i);
+		
+		if(input == "menu")
```
tehhowch, line 231: Will there ever be an input that requires setting multiple commands, or requires disambiguation? If not, (and perhaps even if "maybe"), we should use a map instead of a large `if-else if` chain.

source/GameWindow.cpp
```diff
@@ -363,7 +363,8 @@ void GameWindow::ExitWithError(const string& message)
 	box.buttons = &button;
 	
 	int result = 0;
-	SDL_ShowMessageBox(&box, &result);
+	if(doPopUp)
```
tehhowch, line 366: A lot of the other code in this method should be skipped if we aren't displaying an error box to the user.

source/main.cpp
```diff
@@ -95,6 +96,12 @@ int main(int argc, char *argv[])
 		if(!GameData::BeginLoad(argv))
 			return 0;
 		
+		if(!testToRun.empty() && !GameData::Tests().Has(testToRun))
+		{
+			cout << "Test not found." << endl;
```
tehhowch, line 101: This should report the test that was asked for, and possibly also print the game's known tests.

source/main.cpp
```diff
@@ -130,7 +137,8 @@ int main(int argc, char *argv[])
 	catch(const runtime_error &error)
 	{
 		Audio::Quit();
-		GameWindow::ExitWithError(error.what());
+		bool doPopUp = testToRun.empty();
```
tehhowch, line 140: I think it's fine to show the user the error dialog if they are running a test manually

petervdmeer: There is no parameter that tells ES if the test was started from a bash-script or if it was started manually by an user directly from the command-line.

source/main.cpp
petervdmeer: I expect (or at least hope) the list of known tests to become quite long in the future, so I choose to only print the test-name to ensure that the error-message doesn't get lost by a long scrolling list.

source/Command.cpp
petervdmeer: Replaced here and in the split-off PR #4997 that just handles Command-loading.

petervdmeer:
The failing testcase is fixed; turned out that an additional hint in the Outfitter was not expected by the testcase (and the testcase failed on an unknown UI state).

source/Test.h
```diff
@@ -0,0 +1,161 @@
...
+			// Step that launches the players flagship.
+			LAUNCH,
+			// Step to perform loading of a savegame
+			LOAD_GAME,
+			// Instructs the game to set navigation / travel plan to a target system
+			NAVIGATE,
+			// Step that contains a set of test-steps below it. Repeats the
+			// steps inside it a number of times or until a break is given within
+			// the step.
+			REPEAT,
```
petervdmeer, line 68: I'm considering to replace the `repeat` and `wait for` (and `break if`) steps by `label`, `goto` and `branch` so that they match the syntax of conversation nodes. Question for reviewers: What do you think, would that be usefull?

tehhowch: If we plan to support scripted flight sequences / game control, it's possible some syntax should be shared between the two interfaces.

(I'm not sure if it makes the most sense for the two situations to share the same implementation / command injection method though, since during a scripted mission / behavior the player by definition should not be able to influence things, but in the case of an externally-driven test it is the "player" giving commands.)

petervdmeer: I agree that scripted flight sequences / game control should only have access to the in-game subset of controls. Those should not be able to access menus for things like preferences or loading savegames.

The question I was actually asking here was about sharing syntax with the _current_ conversations. Conversations use the `label`, `goto` and `branch` nodes. I wasn't planning on sharing implementation yet, but I can imagine that it would help if the syntax of the nodes in conversations and in tests matches a bit more.

petervdmeer: I added label and branch together with the relevant TODO items. This is now tracked in the code, so resolving this conversation.

petervdmeer:
@tehhowch
> Didn't look at the biggest files yet

I have reduced the biggest file Test.cpp to just the framework code and an overview of the keywords (and I removed the non-needed test-data). The actual Test-runner implementation and test-data can then go to next PRs.

petervdmeer:

source/Test.cpp
```diff
@@ -0,0 +1,124 @@
...
+	}
+	// If a test object is "loaded" twice, that is most likely an error (e.g.
+	// due to a plugin containing a test with the same name as the base game
+	// or another plugin). Tests should be globally unique.
+	if(!name.empty())
+	{
+		node.PrintTrace("Duplicate definition of test");
+		return;
+	}
+	name = node.Token(1);
```
tehhowch, line 51: Should we enforce that a tab character is not a part of the test name, since the tab character is reserved for use as a delimiter by the test runner?

tehhowch (commented):
1) Code::Blocks project files need to be updated with the new classes
2) I think it would be forward-thinking of us to use the TAP-format reporting style when running (optionally, only when run by CI, though the Test-Anything Protocol is quite readable for humans too).

👍 & thanks for the patience!

source/Test.cpp
petervdmeer: Sounds good, but I chose to validate against a list of allowed characters instead. This way we also take into account other characters that could lead to problems.

tests/run_tests.sh
```diff
@@ -0,0 +1,94 @@
...
+
+	TEST_RESULT="PASS"
+	echo "Running test \"${TEST}\" with ${ES_CONFIG_PATH}"
+	# Use pipefail and use sed to remove ALSA messages that appear due to missing soundcards in the CI environment
+	set -o pipefail
+	"$ES_EXEC_PATH" --resources "${RESOURCES}" --test "${TEST}" --config "${ES_CONFIG_PATH}" 2>&1 |\
+		sed -e "/^ALSA lib.*$/d" -e "/^AL lib.*$/d"
+	if [ $? -ne 0 ]
+	then
+		echo "errors.txt:"
```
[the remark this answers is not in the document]
petervdmeer: I expect that some tests will trigger error-messages because they explicitly test things like deprecated constructions (and trigger a deprecation error/warning), corner-cases or even test direct handling of invalid constructs.
If we want to check the error file after tests, then we should also have a mechanism from within the test to indicate which error-messages are allowed (or at least if errors are expected/allowed).

Shall we keep this for a follow-up PR?

tests/run_tests.sh
```diff
@@ -0,0 +1,94 @@
...
+HERE=$(cd `dirname $0` && pwd)
+RESOURCES=$(echo "${HERE}/.." | sed "s,/tests/..,,")
+
+# Determine paths to endless-sky executable and other relevant data
+ES_EXEC_PATH="${RESOURCES}/endless-sky"
+ES_CONFIG_TEMPLATE_PATH="${RESOURCES}/tests/config"
+
+if [ ! $? ]
+then
+	echo "Error: couldn't create temporary directory"
```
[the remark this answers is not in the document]
petervdmeer: This check was meant to check mktemp, but apparently I forgot to move it when I moved the mktemp call itself.

tests/run_tests.sh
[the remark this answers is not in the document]
tehhowch: In an ideal world, we would not have acceptance-level tests using the UI framework that check if a deprecated construct can still be loaded, and that doing so triggers the associated warning printout -- this type of test would be at the unit / module level. We're not in that world yet, so your plan to use UI tests for this is OK

tests/run_tests.sh
```diff
@@ -0,0 +1,105 @@
...
+TESTS=$("${ES_EXEC_PATH}" --tests --resources "${RESOURCES}")
+if [ $? -ne 0 ]
+then
+	echo "1..1"
+	echo "not ok 1 Could not retrieve testcases"
+	exit 1
+fi
+TESTS_OK=$(echo "${TESTS}" | grep -e "^ACTIVE" | cut -d$'\t' -f2)
+TESTS_NOK=$(echo "${TESTS}" | grep -e "^KNOWN FAILURE" -e "^MISSING FEATURE" | cut -d$'\t' -f2)
+TESTS_TOTAL=$(( 0 + $(echo "${TEST_OK}" | wc -l) ))
```
tehhowch, line 43: calculation isn't quite right:
image

tests/run_tests.sh
```diff
@@ -0,0 +1,105 @@
...
+	then
+		if [ -f "${ES_CONFIG_PATH}/errors.txt" ]
+		then
+			echo "# errors.txt:"
+			cat "${ES_CONFIG_PATH}/errors.txt" | sed "s/^/# /"
+		fi
+		TEST_RESULT="not ok"
+		NUM_FAILED=$((NUM_FAILED + 1))
+	fi
+	echo "${TEST_RESULT} ${TEST}"
```
tehhowch, line 94: We should include the test number here, e.g.
image

tests/run_tests.sh
```diff
@@ -0,0 +1,105 @@
...
+	fi
+
+	ES_SAVES_PATH="${ES_CONFIG_PATH}/saves"
+	mkdir -p "${ES_CONFIG_PATH}"
+	mkdir -p "${ES_SAVES_PATH}"
+	cp ${ES_CONFIG_TEMPLATE_PATH}/* ${ES_CONFIG_PATH}
+
+	TEST_NAME=$(echo ${TEST} | sed "s/\"//g")
+	TEST_RESULT="ok"
+	echo "# Running test ${TEST} with ${ES_CONFIG_PATH}"
```
tehhowch, line 79: Let's only print the temp directory that was used if the test fails, to avoid basically double-printing the test name

tests/run_tests.sh
```diff
@@ -0,0 +1,105 @@
...
+done
+
+IFS=${IFS_OLD}
+
+if [ ${NUM_FAILED} -ne 0 ]
+then
+	echo "# ${NUM_FAILED} tests failed"
+	exit 1
+fi
+echo "# All executed tests passed"
```
tehhowch, line 104: it's not in the TAP spec, but i like this summary style:
image
thoughts?

(this summary comes from a framework that doesn't know how many tests it will run until the end, i.e. it has conditional groups of tests that will be skipped if some other ones fails, so it prints the plan at the end)

petervdmeer: I also like the short summary style and I like the empty lines before the summary.
Just added those.

We know up-front how many tests will be run, so I prefer to keep printing the plan-line at the beginning.

source/main.cpp
```diff
@@ -95,6 +96,12 @@ int main(int argc, char *argv[])
 		if(!GameData::BeginLoad(argv))
 			return 0;
 		
+		if(!testToRun.empty() && !GameData::Tests().Has(testToRun))
+		{
+			cout << "Test " << testToRun << " not found." << endl;
```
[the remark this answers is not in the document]
petervdmeer: Files::LogError does a print to std-error, so I replaced this by Files::LogError without additional printing here..

source/main.cpp
```diff
@@ -262,15 +273,9 @@ void GameLoop(PlayerInfo &player, const Conversation &conversation, const string
...
-		//
-		// This testcase can catch issues related to startup/data-loading,
-		// and it helps to show if the CI testframework is working.
-		// This hardcoded testcase is expected to be replaced by the
-		// larger testframework that is planned for ES.
-		if(!testToRun.empty() && GameData::IsLoaded())
-			menuPanels.Quit();
+		// All manual events and processing done. Handle any test inputs and events if we have any.
+		if(!testToRun.empty())
+			(GameData::Tests().Get(testToRun))->Step(testContext, menuPanels, gamePanels, player);
```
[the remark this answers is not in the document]
petervdmeer: Sounds good, but I added a pointer to the test-to-run to the testContext instead.

source/TestData.cpp
```diff
@@ -0,0 +1,95 @@
...
+}
+
+
+
+// Inject the test-data to the proper location
+bool TestData::Inject() const
+{
+	// We only know how to inject savegame data for now.
+	if((dataSetType != TestData::SAVEGAME) || (dataSetName.empty()) || (sourceDataFile.empty()))
+		return false;
```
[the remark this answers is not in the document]
petervdmeer: I choose the first option with the switch statement. Derived classes would also be possible, but then we would need the Load-function to follow a factory pattern (providing us an object of the derived class after determining in the load which derived class to choose).

source/Test.cpp
```diff
@@ -0,0 +1,124 @@
...
+		menuPanels.Quit();
+		return;
+	}
+	
+
+	// Exit with error on a failing testStep.
+	// Throwing a runtime_error is kinda rude, but works for this version of
+	// the tester. Might want to add a menuPanels.QuitError() function in
+	// a later version (which can set a non-zero exitcode and exit properly).
+	throw runtime_error("Teststep " + to_string(context.stepToRun) + " failed");
```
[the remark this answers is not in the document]
petervdmeer: I added the type of the step. When we have more implementation of the steps, then we can also add test-specific info.

petervdmeer:
Thanks for the good review. I think I processed all comments. I'm also busy with life outside ES, so my response times are also a bit longer than before.

petervdmeer:
Merged master to resolve a minor merge-conflict with a recently merged jump-ranges PR.

source/TestData.cpp
```diff
@@ -0,0 +1,115 @@
...
+	
+	// Data-section was no longer found.
+	return false;
+}
+
+
+
+// Inject the test-data to the proper location.
+bool TestData::Inject() const
+{
```
tehhowch, line 106: style: this public method should be above the private method

source/main.cpp
```diff
@@ -87,14 +89,20 @@ int main(int argc, char *argv[])
...
 	}
 	
 	try {
 		// Begin loading the game data. Exit early if we are not using the UI.
 		if(!GameData::BeginLoad(argv))
 			return 0;
 		
+		if(!testToRunName.empty() && !GameData::Tests().Has(testToRunName))
+		{
+			Files::LogError("Test " + testToRunName + " not found.");
```
tehhowch, line 102:
```suggestion
			Files::LogError("Test \"" + testToRunName + "\" not found.");
```

source/Test.cpp
```diff
@@ -0,0 +1,186 @@
...
+		{
+			const string &statusText = child.Token(1);
+			auto it = find_if(STATUS_TO_TEXT.begin(), STATUS_TO_TEXT.end(),
+				[&statusText](const std::pair<Test::Status, string> &e) {
+					return e.second == statusText;
+				});
+			if(it != STATUS_TO_TEXT.end())
+				status = it->first;
+			else
+				child.PrintTrace("Unknown test-status " + statusText);
```
tehhowch, line 120: We may want to assign the status to something other than the default in this scenario, given that the default is "active"

(running `--tests` will display the "unknown status" message, then the table of tests will report it as "active")

source/Test.cpp
```diff
@@ -0,0 +1,186 @@
...
+	throw runtime_error(testFailMessage);
+}
+
+
+
+string Test::StatusText() const
+{
+	auto it = STATUS_TO_TEXT.find(status);
+	if(it != STATUS_TO_TEXT.end())
+		return it->second;
```
tehhowch, line 183: here too

tehhowch (commented):
via the git bash terminal, but still, on windows: 
image

Couple minor things then this is ready for 🚢

source/Test.cpp
```diff
@@ -0,0 +1,186 @@
...
+		menuPanels.Quit();
+		return;
+	}
+	
+	const TestStep &stepToRun = steps[context.stepToRun];
+	
+	// Exit with error on a failing testStep.
+	string stepTypeName = "unknown teststep type";
+	auto it = STEPTYPE_TO_TEXT.find(stepToRun.stepType);
+	if(it != STEPTYPE_TO_TEXT.end())
```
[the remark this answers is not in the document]
petervdmeer: Yes, those guards should be in Load.

source/Test.cpp
```diff
@@ -0,0 +1,186 @@
...
+		Test::TestStep::Type stepType;
+		auto it = find_if(STEPTYPE_TO_TEXT.begin(), STEPTYPE_TO_TEXT.end(),
+			[&typeName](const std::pair<Test::TestStep::Type, string> &e) {
+				return e.second == typeName;
+			});
+		if(it != STEPTYPE_TO_TEXT.end())
+			stepType = it->first;
+		else
+			child.PrintTrace("Unknown teststep type " + child.Token(0));
+		steps.emplace_back(stepType);
```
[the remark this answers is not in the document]
petervdmeer: I agree that we don't need to emplace invalid step types, but we should also not silently ignore them. I added a "Broken" test status to handle tests that contain invalid steps.

petervdmeer:
Processed review-comments and merged master to resolve merge-conflict with system hazards.

tehhowch (approved):
LGTM!

@petervdmeer take a look at the recent commits, and make any updates you'd like to see. Otherwise, this is ready to go

source/Test.cpp
```diff
@@ -0,0 +1,215 @@
...
+		{Test::TestStep::Type::NAVIGATE, "navigate"},
+		{Test::TestStep::Type::WATCHDOG, "watchdog"},
+	};
+	
+	template<class K, class... Args>
+	string ExpectedOptions(const map<K, const string, Args...> &m)
+	{
+		if(m.empty())
+			return "no options supported";
+
```
petervdmeer, line 64: I think this code also needs an exception for dealing with maps that only contain 1 element:
- Functional: I don't know if accumulate can handle the begin-range element being behind the end-range element.
- Textual: here should not be an `or` in the string when there is only 1 element.

petervdmeer:
@tehhowch Thanks for your updates, good improvements to the code!
I added a little bit of code for printing maps with 1 element and I also consider this ready to go now.

source/Test.cpp
petervdmeer: Handled in d7150a6

tehhowch: I didn't bother with those since this is file-local only and not truly "generic" -- we have the guarantee that any mapping passed to it is non-empty and also has more than 2 elements, so the beginning and end iterators are well-formed.
As a file-local method, anyone adding new calls to it is trivially able to view and/or modify the implementation.

I added the empty check mostly for sanity - we dereference `begin` - and because I considered it might result in better codegen since the compiler will know more about the map state after it. (I didn't verify that assumption, though.)

Issue #3682, opened 2018-04-07
Angle-view ship thumbnails in the shipyard


endless-sky:
I'm trying to brainstorm ideas for "experimental" additions to the game that could easily be added in unstable releases to get player feedback, then removed later on if people don't like them. (Like the news system.) One idea that has been floating around for a while, and that existed in Escape Velocity, is special thumbnail images of ships just for the shipyard. Here's a mock-up of what that might look like:

![shipyard mockup](https://user-images.githubusercontent.com/11450262/38460337-cb2ebe66-3a85-11e8-89c1-73b8be6dabba.jpg)

The disadvantages would be:
- More work for content creators.
- Some ship models are only designed to look good from the top down.
- More images to load and more memory usage.

On the other hand, for most models it'd only take 5-10 minutes to render and post-process a shipyard thumbnail. The shipyard would fall back to using the main sprite for any ship that doesn't have a separate thumbnail sprite.

jafdy:
It's a nice idea and I see no reason not to have the capability. 
Some ships (like the mule) might need a slight rework to make it look decent.

Brick63:
Personally I feel that this could work as an optional feature that can be turned on and off.

KiLEdEnNis:
as Kiko said before, the human ships are getting reworked by some of us (kryes omega, wrzl & me), so they're not really a problem as we aim for good models, not just top-down views
as for alien ships, I can't really imagine any alien species where this would look bad (maybe a couple Hai ships)
it could look bad on the current Quarg ships too, which aren't for sale anyway & if I remember, Brick and maybe some others were working on better Quarg ships

that plus the upsides it gives would make the shipyards look a lot better, and making it a toggle feature (making a toggle button both IN the shipyard and in preferences) would fix basically every downside apart from the "more memory to load"

kikotheexile:
Yup, this just basically just motivates the team to *want* to rework the ships so their flattering angles are used. 

Advertisements, actual shipyard images, and scenes to be included in story missions will go a long way to help player immersion and add polish to the game. People are going to love it, and it will be fun to work on. The only downside is that it will slightly raise the difficulty of ship modding/art.

KiLEdEnNis:
to be honest, if a modder can make a ship and post-process it, they can easily do this as well, it's just a matter of repositioning the camera, rendering, and post-processing a bit

jafdy:
Some people do make/import ships in 2D. So they can't just "reposition the camera".

[...]

PR #4377, opened 2019-06-17
Adding missing shipyard thumbnails

Amazinite:
Ref: #3682, #4178

Adding the remaining spaceport thumbnails that haven't been added yet.

- [x] Marauder ships
- [x] Archon
- [x] World-Ship B and C
- [x] Quarg ships
- [x] Lampyrid

You may notice that there is no `mfirebirde` thumbnail; that's because the extra engine that this variant gains isn't visible from the thumbnail perspective, so the `mfirebirde` and `mfirebirds` both use the same thumbnail.
Referencing #4178, these Marauder thumbnails don't change the style of post-processing. The high-DPI PR is also currently up and will be updated alongside this PR.
Renamed the `world-ship` thumbnail to `world-ship a` to follow the naming convention of the top-down images for the World-Ship.

GIMP file assets: https://drive.google.com/open?id=1lqO8OGcSUIUiI4DjWSQhOvdpOgQl7pv7


Amazinite:
Done all the images. I'll let this sit for a bit for comments before merging.

Nescio0:
This is great, many thanks!
However, wouldn't it be better to include the `mfirebirde` for consistency, even if it's identical to the `mfirebirds`?
Furthermore, I see you added the `lampyrid` and `mfury` author ships, but not the `mosprey` or `pointedstick_vanguard`? (Also, shouldn't the last one be renamed to `mvanguard` for consistency?

Amazinite:
>However, wouldn't it be better to include the mfirebirde for consistency, even if it's identical to the mfirebirds?

I don't see why it would make sense to have two of the same image, just with different names.

>Furthermore, I see you added the lampyrid and mfury author ships, but not the mosprey or pointedstick_vanguard? (Also, shouldn't the last one be renamed to mvanguard for consistency?

I excluded those two person ships because they aren't capturable and because they are not included in the [official all content plugin](https://github.com/endless-sky/all-content-plugin/blob/master/data/all%20ships.txt). The rest of the ships with thumbnails are capturable, with the exception of the Archon, but that is included in the all content plugin.

PR #4382, opened 2019-06-20
Add # of guns/turrets to --ships flag

fcfort:
This adds the maximum number of guns and turrets per ship to the `--ships` command line flag of the main executable. This is useful for creating data tables from the game's resource files.


source/GameData.cpp
```diff
@@ -1045,7 +1045,8 @@ void GameData::PrintShipTable()
 		<< "mass" << '\t' << "crew" << '\t' << "cargo" << '\t' << "bunks" << '\t'
 		<< "fuel" << '\t' << "outfit" << '\t' << "weapon" << '\t' << "engine" << '\t'
 		<< "speed" << '\t' << "accel" << '\t' << "turn" << '\t'
-		<< "e_gen" << '\t' << "e_use" << '\t' << "h_gen" << '\t' << "h_max" << '\n';
+		<< "e_gen" << '\t' << "e_use" << '\t' << "h_gen" << '\t' << "h_max" << '\t'
+		<< "guns" << '\t' << "turrets" << '\n';
```
petervdmeer: Overall change looks ok to me. One detail, I think "gun_ports" and "turrent_mounts" are more appropriate as header entries, since you are not printing the actual guns and turrets, but only the places to install them in.

fcfort: Done, went with `gun_mnts` / `turret_mnts` for brevity.

tehhowch:
> The headers are still too long for the default column width when printing to the console. I suggest just "guns" and "turrets". Those are clear enough, and this is developer info anyway.

Given that we don't align any columns based on their widths, printing this information to the console is gibberish anyway. Probably the table outputs are only useful when stdout is redirected to a file and opened by a spreadsheet editor, at which point it is more important for the column headers to be clear, rather than compact.

PR #4388, opened 2019-06-22
feat: If not carried, fighters and drones are launched instead of sold

petervdmeer:
Because they are no longer sold, the following issues get resolved:
 - Fixes endless-sky#3764
 - Fixes endless-sky#3706
 - Fixes endless-sky#4427

UI changes:
 - Updates Shipyard & Outfitter warning icons, to highlight exactly which ships will not be carried:
image
 - Updates departure message when over bay capacity:
image
 - Updates post-launch message when over bay capacity:
image


jafdy:
This will lead to people asking why their fighters can't jump.

petervdmeer:
> This will lead to people asking why their fighters can't jump.

Are you expecting a lot of such questions? Players that buy fighters did already make some progress through the game (at least they made it to a planet selling fighters), so I would expect such players to have seen and know about the Hyperdrive outfit by then.

If such questions are expected, then maybe it's better to add a warning at launch if the player launches ships that are not carried and that don't have hyper and/or jumpdrives (or jump fuel). That way you will warn players on all jump incapable ships instead of just the fighter and drones. (And you also don't warn players that equipped their fighters with a hyperdrive and a fuel tank.)

petervdmeer:
> This will lead to people asking why their fighters can't jump.

I added the warning back into the game, but now it warns about the fighters and drones not being able to jump instead of warning about selling the fighters and drones. (It just warns on all fighters and drones for now, including the ones with a hyperdrive and a fuel-tank installed.)

ChamEV:
Logic looks ok to me.  Not typically a fan of the pop up warnings, but this case may be acceptable so long as it only applies when the fighters without bays are in the current system, and not every system you land on.

jafdy:
It will happen, people try to buy fighters as extra defence ships for their small freighter. Even with the current warning in place. Many people are just none too bright, you get used to it after a while.

source/PlanetPanel.cpp
```diff
@@ -297,24 +298,20 @@ void PlanetPanel::TakeOffIfReady()
...
+		else if (fighterCount > 0 || droneCount > 0)
+		{
+			int nonJumpCount = fighterCount > 0 ? fighterCount : 0;
+			nonJumpCount += droneCount > 0 ? droneCount : 0;
+			out << "If you take off now you will launch with ";
+			if (nonJumpCount == 1)
+				out << "a ship";
+			else
+				out << nonJumpCount << " ships";
+			out << " that will not be able to leave the system.";
```
Amazinite: Why not just show this warning for all ships that can't jump instead of just counting uncarried fighters and drones?

petervdmeer: You're right. The reason I didn't do that yet was to keep this pull request simple.
I'll push a commit that also checks JumpsRemaining() of each ship in the loop.

petervdmeer:
> It will happen, people try to buy fighters as extra defence ships for their small freighter. Even with the current warning in place. Many people are just none too bright, you get used to it after a while.

The new warning is in place, people now get warned about some of their ships (fighters and ships without jump/hyperdrive) not being able to jump. Do you expect this warning to be sufficient?

Zitchas:
I would personally expect a simple warning that some ships will not be able to jump is sufficient.

petervdmeer:
> I would personally expect a simple warning that some ships will not be able to jump is sufficient.

Okay, then I consider this change ready for merge, or at least ready for re-review.

tehhowch:
In addition to being a behavior change, this is also a balance change -- it becomes considerably easier for the player to stash low-cost reinforcements / cannon fodder around the map. Consider especially the Wanderer storyline: without needing a carrier to hold excess drones, the player can launch with an absurd amount of Flycatchers, which significantly reduces the threat from Trackers to the player's flagship. 

This isn't to say the change is unacceptable, or that stashing reinforcements cannot be done currently (it can), just that this PR bears more consideration and is thus better suited to an unstable release.

petervdmeer:
The balance change is a good point; let's keep this for an unstable release.

tehhowch:
I did some research on the ES Google Group and couldn't really discern a design decision for selling the uncarried fighters and drones (vs simply launching them), so my guess is that they were sold off to simplify the player experience and game code: as @jafdy mentions, this will definitely increase the number of times that players are able to strand a ship in a system.

The warning does its job, but unfortunately fixing the issue is not entirely intuitive while landed, because the Shipyard sidebar does not display a "cannot jump" warning icon for fighters and drones (Ship::FlightCheck explicitly ignores performing this check when a ship can be carried). The warning is also not perhaps entirely correct - we can happily outfit some carried ships with jump / scram / hyperdrives and the fuel to use them, and the pre-launch warning will still be issued.

This PR also exposes a bug, in that if the player jumps out of the fighters' systems, they will stop following it around and instead will hyperspace randomly. (This bug is due to the AI reparenting code for orphaned fighters and drones, which removes the ship's parent under some now possibly faulty assumptions.)

Zitchas:
In regards to the "can't hyperspace" warning, is there a way to turn that into a three-way icon? As in have three possible states: 
- Blank (no warning) meaning that the ship can hyperjump
- Red, meaning that the ship cannot hyperjump and there are the player has no ships capable of carrying it.
- Yellow, meaning the ship cannot hyperjump, but the player has ships capable of carrying it.

The icon itself should be different too, not just the colour (there are colour blind players, I'm sure.)

A four-way variation on that might be:
- Blank: no warning, the ship can hyperjump.
- Green: the ship can't hyperjump, but the player has ships in the fleet that can carry it. (maybe even with an internal counter that causes any ships beyond the player's fleet's current limit to be yellow instead of green. So if the player has 1 ibis and they buy 5 terns, the first four are green and the fifth one is yellow)
- Yellow: the ship can't hyperjump, and ships exist that can carry it, but the player doesn't have any of them.
- Red: the ship can't hyperjump, and no ships exist that can carry it.

petervdmeer:
@tehhowch Thanks for looking up the design decissions and for the review.
I changed the code to launch carried ships outside carriers if they can jump by themselves and I fixed the exposed bug by reparenting jump-capable carried ships to the flagship if the flagship is in a different system.
I'll look at the "cannot jump" warning icon later and I will also test this a bit more when I have dealth with the warning icons. (The last 3 commits were uploaded to already allow some review.)

petervdmeer:
@Zitchas: Yes, I believe we can have different icons there, but I suggest to give no warning if the ship can hyperjump or the ship can be carried and there is sufficient carrier space for the carried category. (Why warn for a situation that needs no attention?)

Ships that cannot jump and cannot be carried already get a warning at this moment, we can probably just keep the current warning there.

A Yellow/Warning label indeed seems appropriate for ships that cannot jump by themselves but that can be carried and where there is insufficient space for the carried category. (And then apply the yellow/warning label to all ships in that category, not just the overflow.)

petervdmeer:
The warning on insufficient bays is now also added. It looks like this:
bays_warning
(My flagship has 0 bays, the carrier has 6 bays and I have 7 fighters in this example.)

source/PlayerInfo.cpp
```diff
@@ -1210,8 +1218,8 @@ bool PlayerInfo::TakeOff(UI *ui)
 		
 		bool fit = true;
 		const string &category = ship->Attributes().Category();
-		bool isFighter = (category == "Fighter");
-		if(isFighter || category == "Drone")
+		// Load only fighters and drones that cannot jump by themselves
+		if((ship->JumpsRemaining() < 1) && (category == "Fighter" || category == "Drone"))
```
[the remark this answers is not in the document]
petervdmeer: I added an extra loop to also load jump-capable carryable ships in any left-over bays.

petervdmeer:
> Feels kinda gross to be using an int array here (and doubly so to pass in a dummy one for the one invocation). Were there other options you considered?

I changed the flight-check function parameter to a single boolean and moved the logic of matching bays to before the flight-check.

tehhowch:
Re: wormhole pathing of carried ships: https://github.com/endless-sky/endless-sky/issues/2937

Amazinite:
I suspected that it was already an open issue.

PR #4391, opened 2019-06-23
refactor(mechanics): Change bay type identification to categories

petervdmeer:
Change the ways bay types are identified from boolean to the category of the ship that is to be carried in the bay.

This will make it easier to add new bay types in the future and/or to load bay types from data files instead of having them hardcoded.


petervdmeer:
I only did some static testing with this change (ship-info displays, loading and saving). This needs some testing during actual gameplay with deploying and recalling of fighters.

finite-galaxy:
`Fleet.cpp` and `PlayerInfo.cpp` use `isFighter` too, perhaps change that as well?

petervdmeer:
> `Fleet.cpp` and `PlayerInfo.cpp` use `isFighter` too, perhaps change that as well?

The variable in Fleet.cpp has more the meaning of "isCarriedShip" and already works with ship->CanBeCarried(), so logic-wise that code already works for the change in this PR.
I could change the variable name for consistency, but in this case I prefer to keep the change smaller and not touch this file. (Smaller changes reduce the risk of merge-conflicts with other PRs.)

The variables in PlayerInfo.cpp should indeed change, but I also have PR #4388 which completely changes that piece of code. If PR#4388 gets merged before this one, then I'll update that code with a new commit here. If this commit gets merged before #4388, then I'll update that code with a new commit there.

Amazinite:
If the new keyword for a carried ship is "bay", then this PR should probably make that change to all vanilla ships.

petervdmeer:
> If the new keyword for a carried ship is "bay", then this PR should probably make that change to all vanilla ships.

That is possible, but the change is backwards compatible (fighter and drone keywords also still work), so we can do the change to the vanilla ships also later in a separate PR.

New savegames will be automatically use the bay keyword after this PR, and I did test the savegame saving and loading (as well as the backwards compatible loading of old savegames and current datafiles).

Amazinite:
If this PR changes proper syntax though, even if there is backwards compatibility, I don't see why this PR wouldn't also replace the old syntax with the new.

petervdmeer:
The main reason for me to use separate PRs is merging-speed and avoidance of merge-conflicts. The datafiles related to ships got modified a number of times already since this PR was created. If new ships with bays get merged in while this PR is in flight, then I should update any such new ship while this PR is in flight.

Keeping the change to the datafiles for a later PR will reduce the time-window in which I can have merge-conflicts on the datafiles.

(This strategy is based on the assumption that the current PR is difficult/complex and has a long review and merging time, while my PR with the data-files update would be a no-brainer that gets merged within a couple of days after submitting.)

source/Ship.cpp
```diff
@@ -2702,33 +2724,33 @@ bool Ship::HasBays() const
...
 // not reserved for one of its existing escorts.
 bool Ship::CanCarry(const Ship &ship) const
 {
-	if(!ship.canBeCarried)
+	if(!ship.CanBeCarried())
 		return false;
-	// This carried ship is either a fighter or a drone.
-	bool isFighter = (ship.attributes.Category() == "Fighter");
+	// Check only for the category that we are interrested in.
+	string category = ship.attributes.Category();
```
tehhowch: This should probably also be a const ref to avoid string copies.

source/Ship.cpp
```diff
@@ -2738,21 +2760,21 @@ bool Ship::CanCarry(const Ship &ship) const
...
 
 bool Ship::Carry(const shared_ptr<Ship> &ship)
 {
-	if(!ship || !ship->canBeCarried)
+	if(!ship || !ship->CanBeCarried())
 		return false;
 	
-	// This carried ship is either a fighter or a drone.
-	bool isFighter = ship->attributes.Category() == "Fighter";
+	// Check only for the category that we are interrested in.
```
tehhowch: Typo here, const ref below

source/PlanetPanel.cpp
```diff
@@ -273,9 +273,8 @@ void PlanetPanel::TakeOffIfReady()
 	for(const auto &it : player.Ships())
 		if(!it->IsParked() && !it->IsDisabled() && it->GetSystem() == &system)
 		{
-			const string &category = it->Attributes().Category();
-			droneCount += (category == "Drone") - it->BaysFree(false);
-			fighterCount += (category == "Fighter") - it->BaysFree(true);
+			droneCount -= it->BaysFree("Drone");
```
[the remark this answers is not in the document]
petervdmeer: This looks a bit strange indeed. Restored the counting.

source/Ship.cpp
```diff
@@ -535,10 +561,6 @@ void Ship::FinishLoading(bool isNewInstance)
 		if(bay.side == Bay::INSIDE && bay.launchEffects.empty() && Crew())
 			bay.launchEffects.emplace_back(GameData::Effects().Get("basic launch"));
 	
-	// Figure out if this ship can be carried.
-	const string &category = attributes.Category();
-	canBeCarried = (category == "Fighter" || category == "Drone");
```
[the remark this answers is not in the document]
petervdmeer: Moved the calculation here and using the cached value again in Ship::CanBeCarried.

source/Ship.cpp
```diff
@@ -1720,15 +1742,15 @@ void Ship::DoGeneration()
...
+	// Allow carried ships to launch from a disabled ship, but not from a ship that
 	// is landing, jumping, or cloaked. If already destroyed (e.g. self-destructing),
 	// eject any ships still docked, possibly destroying them in the process.
 	bool ejecting = IsDestroyed();
 	if(!ejecting && (!commands.Has(Command::DEPLOY) || zoom != 1.f || hyperspaceCount || cloak))
 		return;
 	
 	for(Bay &bay : bays)
-		if(bay.ship && ((bay.ship->Commands().Has(Command::DEPLOY) && !Random::Int(40 + 20 * bay.isFighter))
+		if(bay.ship && ((bay.ship->Commands().Has(Command::DEPLOY) && !Random::Int(50))
```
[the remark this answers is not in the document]
petervdmeer: Used (crew > 0) as check to (almost reliably) verify that this code is about fighters. Fighters without crew now however use value 40 instead of 60.

source/Ship.cpp
```diff
@@ -2702,33 +2724,33 @@ bool Ship::HasBays() const
 
 
 
-int Ship::BaysFree(bool isFighter) const
+int Ship::BaysFree(string category) const
```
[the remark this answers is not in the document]
petervdmeer: Changed to const reference. For a future update we might want to switch to numerical identifiers to get even better performance (since number comparisons are faster than string compares). But I'll keep that out of this change.

source/Ship.cpp
petervdmeer: Agreed. Changed it to const ref.

source/Ship.cpp
petervdmeer: Fixed both.

source/Ship.cpp
```diff
@@ -100,6 +103,12 @@ const vector<string> Ship::CATEGORIES = {
 };
 
 
+const set<string> Ship::BAY_TYPES()
```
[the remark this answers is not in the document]
petervdmeer: Changed it to non-function call, similar to the categories.

petervdmeer:
This PR has been updated to the latest master to resolve a merge-conflict.

petervdmeer:
Merged master to handle merge conflict

source/ShipInfoDisplay.cpp
```diff
@@ -248,8 +248,8 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 			string bayLabel = bayType;
 			transform(bayLabel.begin(), bayLabel.end(), bayLabel.begin(), ::tolower);
 			
-			attributeLabels.push_back(bayLabel + " bays:");
-			attributeValues.push_back(to_string(ship.BaysFree(bayType)));
+			attributeLabels.emplace_back(bayLabel + " bays:");
+			attributeValues.emplace_back(to_string(ship.BaysFree(bayType)));
```
tehhowch, line 252: You should not use `BaysFree` here

source/ShipInfoDisplay.cpp
```diff
@@ -240,31 +240,18 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
...
-			if(bay.isFighter)
-				++fighter;
-			else
-				++drone;
-		}
-		
-		if(drone)
+	// Print the number of bays for each bay-type we have
+	for(auto bayType : Ship::BAY_TYPES)
+		if(ship.BaysFree(bayType))
```
tehhowch, line 245: Use of `BaysFree` won't properly display bay counts while in flight

source/Ship.cpp
```diff
@@ -1733,15 +1757,15 @@ void Ship::DoGeneration()
...
+	// Allow carried ships to launch from a disabled ship, but not from a ship that
 	// is landing, jumping, or cloaked. If already destroyed (e.g. self-destructing),
 	// eject any ships still docked, possibly destroying them in the process.
 	bool ejecting = IsDestroyed();
 	if(!ejecting && (!commands.Has(Command::DEPLOY) || zoom != 1.f || hyperspaceCount || cloak))
 		return;
 	
 	for(Bay &bay : bays)
-		if(bay.ship && ((bay.ship->Commands().Has(Command::DEPLOY) && !Random::Int(40 + 20 * bay.isFighter))
+		if(bay.ship && ((bay.ship->Commands().Has(Command::DEPLOY) && !Random::Int(40 + 20 * (crew > 0)))
```
tehhowch, line 1768: This makes carrier ships with a crew > 0 launch any of its carried ships more slowly than a carrier ship that is an automaton, even if the carried ships themselves are not automatons.

While that's a neat idea (i.e. an automaton carrier issues the launch order faster than a non-automaton carrier), the current behavior is that a drone "pilot" responds to the order faster than a live pilot.

I think you should be checking `bay.ship->attributes.Get("automaton")`

source/Engine.cpp
```diff
@@ -319,20 +319,19 @@ void Engine::Place(const list<NPC> &npcs, shared_ptr<Ship> flagship)
 {
 	for(const NPC &npc : npcs)
 	{
-		map<Ship *, int> droneCarriers;
-		map<Ship *, int> fighterCarriers;
+		set<Ship *> carriers;
```
tehhowch, line 322: This could be inside a map keyed by ship category to make handling of large numbers of NPC carried ships more intelligent.  (Could also keep the `map<Ship *, int>` signature too instead of using a set.)

source/Ship.cpp
```diff
@@ -209,16 +218,33 @@ void Ship::Load(const DataNode &node)
...
+			// Initialize as if the keyword was fighter. The values will be
+			// overwritten in case of drone or bay keywords.
+			// Newer format is bay "BayCategory".
+			string category = "Fighter";
+			int childOffset = 0;
+			if(key == "drone")
+				category = "Drone";			
+			else if(key == "bay")
+			{
+				category = child.Token(1);
```
tehhowch, line 234: Validation needed here - we don't support arbitrary carried types yet, so the bay category must be in BAY_TYPES.

When we do support custom bay types, the validation will have to be moved to Ship::FinishLoading as there won't be a guarantee that all bay types have been loaded yet.

source/Ship.h
```diff
@@ -54,16 +55,18 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
...
-		Bay(double x, double y, bool isFighter) : point(x * .5, y * .5), isFighter(isFighter) {}
+		Bay(double x, double y, std::string forCategory) : point(x * .5, y * .5), forCategory(forCategory) {}
 		// Copying a bay does not copy the ship inside it.
-		Bay(const Bay &b) : point(b.point), isFighter(b.isFighter), side(b.side), facing(b.facing), launchEffects(b.launchEffects) {}
+		Bay(const Bay &b) : point(b.point), forCategory(b.forCategory), side(b.side), facing(b.facing), launchEffects(b.launchEffects) {}
 		
 		Point point;
 		std::shared_ptr<Ship> ship;
-		bool isFighter = false;
+		std::string forCategory;
```
tehhowch, line 69: Let's use `category`

source/ShipInfoDisplay.cpp
petervdmeer: Agreed, I used the value cached earlier in this function.

source/Ship.cpp
petervdmeer: My original idea was to use bay.ship->crew, so this was actually a programming error (taking the crew from the carrier instead of the carried ship). But `bay.ship->attributes.Get("automaton")` also works nicely, so using that one.

source/Engine.cpp
petervdmeer: Yes, changed the code to use a map of maps (outer map with the category, inner map with the signature as given/original).

source/Ship.cpp
petervdmeer: I added validation if the category is known/valid and skip loading for invalid categories.

source/Ship.h
petervdmeer: Ok, done.

source/ShipInfoDisplay.cpp
petervdmeer: Agreed, using the new BaysTotal function.

petervdmeer:
Created a commit to process the first review comments, but there were no large changes required that needed their own commits, so I actually processed all review comments in one go.

source/Ship.cpp
```diff
@@ -2725,33 +2753,46 @@ bool Ship::HasBays() const
...
+// Check if this ship has a bay free for the given ship, and the bay is
 // not reserved for one of its existing escorts.
 bool Ship::CanCarry(const Ship &ship) const
 {
-	if(!ship.canBeCarried)
+	if(!ship.CanBeCarried())
 		return false;
-	// This carried ship is either a fighter or a drone.
-	bool isFighter = (ship.attributes.Category() == "Fighter");
+	// Check only for the category that we are interrested in.
```
tehhowch, line 2785: `interested`

source/Engine.cpp
```diff
@@ -319,20 +319,24 @@ void Engine::Place(const list<NPC> &npcs, shared_ptr<Ship> flagship)
...
 			// Redo the loading up of fighters.
-			ship->UnloadBays();
-			if(ship->BaysFree(false))
-				droneCarriers[&*ship] = ship->BaysFree(false);
-			if(ship->BaysFree(true))
-				fighterCarriers[&*ship] = ship->BaysFree(true);
+			if(ship->HasBays())
+			{
+				ship->UnloadBays();
+				for(string bayType : Ship::BAY_TYPES)
```
tehhowch, line 333: ref

source/Engine.cpp
```diff
@@ -351,9 +355,8 @@ void Engine::Place(const list<NPC> &npcs, shared_ptr<Ship> flagship)
 			if(ship->CanBeCarried())
 			{
 				bool docked = false;
-				map<Ship *, int> &carriers = (ship->Attributes().Category() == "Drone") ?
-					droneCarriers : fighterCarriers;
-				for(auto &it : carriers)
+				string bayType = ship->Attributes().Category();
```
tehhowch, line 358: ref

source/Ship.cpp
```diff
@@ -1733,15 +1761,15 @@ void Ship::DoGeneration()
...
+	// Allow carried ships to launch from a disabled ship, but not from a ship that
 	// is landing, jumping, or cloaked. If already destroyed (e.g. self-destructing),
 	// eject any ships still docked, possibly destroying them in the process.
 	bool ejecting = IsDestroyed();
 	if(!ejecting && (!commands.Has(Command::DEPLOY) || zoom != 1.f || hyperspaceCount || cloak))
 		return;
 	
 	for(Bay &bay : bays)
-		if(bay.ship && ((bay.ship->Commands().Has(Command::DEPLOY) && !Random::Int(40 + 20 * bay.isFighter))
+		if(bay.ship && ((bay.ship->Commands().Has(Command::DEPLOY) && !Random::Int(40 + 20 * bay.ship->attributes.Get("automaton")))
```
tehhowch, line 1772: Oops, I should have negated the suggestion. (Even thought it might be clearer to write `60 - 20 *` instead of `40 + 20 * !`, using the negation prevents a possible call to Random::Int with a negative argument.)

petervdmeer: I should have seen this as well. This caused automated drones to be faster than manual ships.
Added the negation.

source/Ship.cpp
```diff
@@ -209,37 +218,58 @@ void Ship::Load(const DataNode &node)
...
-				hasBays = true;
+				category = child.Token(1);
+				// The "bay" key has category as first child, so all
+				// other childs move 1 position.
+				childOffset += 1;
 			}
-			bays.emplace_back(child.Value(1), child.Value(2), key == "fighter");
-			Bay &bay = bays.back();
-			for(int i = 3; i < child.Size(); ++i)
+			if(BAY_TYPES.count(category) < 1)
```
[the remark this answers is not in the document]
petervdmeer: Yes, that looks nicer. Changed to use `continue`.

petervdmeer:
Processed the latest review comments as well.

PR #4392, opened 2019-06-23
Add option for built-in guns and turrets

petervdmeer:
Hyperdrives, hull-generation, cloaking and many other options can already be built-in into ship hulls. This change adds an option for guns and turrets to be built-in into the hull of the spaceship it is installed in.

Built-in weapons cannot be removed from the ship they are installed in.
And selling the ship that has built-in weapons does not put the built-in weapon in stock.

### DIFF
--- a/data/drak.txt
+++ b/data/drak.txt
@@ -202,12 +202,6 @@ ship "Archon"
 			"hull damage" 10000
 			"hit force" 30000
 	outfits
-		"Drak Antimatter Cannon"
-		"Drak Anti-Missile Field" 2
-		"Drak Draining Field"
-		"Drak Distancer" 2
-		"Drak Turret" 3
-		
 		"Medium Graviton Thruster"
 		"Medium Graviton Steering"
 		"Hyperdrive"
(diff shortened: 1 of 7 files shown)


Amazinite:
Perhaps make the Archon's weapons built-in so that there is an example of this being used in the game.

ChamEV:
Are built in weapons visible in the outfitter?  There probably should be someway of viewing their stats.

beccabunny:
If they were, what would be the sense of adding integrated weapons?

ChamEV:
Wouldn't a better way of doing this be to add specific flags to outfits?
* Never sell
* Invisible in outfiters
* Invisible when scanned (doesn't show up when scanned)
* Built in (takes and uses no space)

And allow a ship to set these flags as well when it specifies it has those outfits.  This will let any ship build in either regular outfits or special ones that can't be acquired.

```
ship "Archon"
    outfits
        "Drak Antimatter Cannon" builtin invisible nosell noscan
```

jafdy:
This can already be done (almost) with some attribute +1 and -1 on both the ship and weapon, this makes it impossible to separate the two without selling the ship and impossible to install the weapon on another ship anyway.

petervdmeer:
Amazinite: Good idea. I just made the change, tested it and fixed a bug. (The turrets and guns showed up on the ship, but they didn't fire because built-in weapons were not reloaded.)

ChamEV: Built-in weapons are currently not visible in the outfitter.
Placing flags on the outfits (and ship) is possible, but if we do that, then the guns or turrets can still be moved to other gun and turret positions within the ship. I would like built-in weapons to stay in the slot they are built in into.

jafdy: Yes, your idea would also work, but it would also allow moving the weapon between slots. Your idea would also require each built-in outfit to be defined separately. If I would want to make something like a Marauder Star Barge with a Quarg Skylance built-in, then I would like to re-use the outfit as-is.

petervdmeer:
Closing this PR; I still believe that the built-in weapons feature has value, but I see other far more valuable features being in development in parallel. The limited amount of time that the current devs have for code-reviews/integration/merging is better spent on those other features.

PR #3583, opened 2018-01-18
Arfecta update / Pug Farming

tehhowch:
<details><summary>old stuff</summary>The Arfecta is woefully outmatched by the player if boarded, and despite being warned about driving off the Pug, there aren't currently any consequences to doing so.

This PR addresses the capturability and combat effectiveness of the Arfecta (significantly increasing the former, and somewhat increasing the latter). A warning is added to foreshadow the permanent rep loss in the form of hostility when the Arfecta are attacked. If the player continues attacking the Arfectas, and ultimately kills or captures both of them, then they will not be able to resume working with the Wanderers.
With the right starting reputation and timings, the player would still be able to obtain 2 Arfectas and complete the Wanderer story (excepting the possibility that the final arc needs the mission Arfectas to be alive).</details>

This PR provides a small increase to the combat effectiveness of the Arfecta (ensuring it can maintain range and avoid its own splash damage, even as hostile ships encroach upon it), and makes fighting the Arfecta more challenging for the player, as the Wanderer ships will engage the player if they attack any in-system Pug ships.

So long as the player does not capture both Ap'arak Arfectas, or attack too many Wanderer ships in the process, there is no permanent damage to player-Wanderer reputation. Pug ships in any other systems remain free game for attack & plunder.

Issue #2131, opened 2017-02-11
Better bounty hunting payment?


ph2000bis:
By using the `on fail` and `npc save`, I could get the payment for bounty-hunting to happen whenever you land anywhere. The catch is, is there a way to display the dialogue when you kill the ship?

And is this a good change in the first place?

Bladewood:
Display dialogue? [Yes.](https://github.com/endless-sky/endless-sky/blob/master/data/wanderers%20middle.txt#L1040) Place `dialogue "[insert text here]"` within (as in, below with an indent) an npc line.
Or at least it should.

ph2000bis:
I tried it, payment happens on landing
From what I understood, the `dialogue` only happens when conditions are met.
The point is that conditions are failed, so that payment happens whatever planet you land on.

jafdy:
@ph2000bis have a look at my hired mecs pull. That way I ran it there may also work for you.

endless-sky:
The problem is that the "on fail" action is also triggered if you abort the mission.

I think the "right" solution would be a code change that allows missions to specify that they can complete on any planet once all other conditions are met. That would be really easy to do (adding about 6 new lines of code) but it's something I'll probably put off until 0.9.7 because 0.9.6 is focused on safe changes and bug fixes.

toilethinges:
destination any?

tehhowch:
Combining this "complete anywhere" mission behavior with the NPC behaviors in #2990 would be fantastic, especially if they can complete in flight (e.g. escort to planet job shows dialog as soon as the NPC lands on the target planet rather than when the player lands on a planet).

Pointedstick:
> The problem is that the "on fail" action is also triggered if you abort the mission.

See https://github.com/endless-sky/endless-sky/issues/4018

Amazinite:
Piggybacking off of this issue, perhaps instead of a simple and brute force "complete anywhere" mechanic, it could be expanded to instead have multiple valid destination planets at the same time. Using #4396 as an example, the Pug Arfecta mission could complete on any Wanderer planet instead of just a single planet and instead of on any planet (as it wouldn't make sense to get a message from the Wanderers if you jumped to human space).

PR #4396, opened 2019-06-25
Have the mission to not kill pug defenders use `save` instead of `kill`

Fzzr:
Right now the Wanderer reaction to killing the pug during "Wanderers Ap'arak 3: Pug" only activates on landing on "Varu Tev'kei". It also allows you to capture one of the Arfectas and continue the Wanderer campaign. This makes it so that the mission is resolved upon landing anywhere, and it fails the Wanderer campaign even if you capture or destroy just one of them.

Thanks to Harryyeo123 on Discord for pointing out this issue.

### DIFF
--- a/data/wanderers start.txt
+++ b/data/wanderers start.txt
@@ -2486,12 +2486,12 @@ mission "Wanderers Ap'arak 3: Pug"
 	source "Varu Tev'kei"
 	to offer
 		has "Wanderers Ap'arak 2: done"
-	npc kill
+	npc save
 		personality staying unconstrained uninterested
 		government "Pug"
 		ship "Pug Arfecta" "Hork Ekekek Ptui"
 		ship "Pug Arfecta" "Graak Sput Kchoo"
-	on complete
+	on fail
 		dialog `Shortly after defeating the Pug ships, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
 		"reputation: Wanderer" <?= -1
 


Fzzr:
If anyone knows how to make the failure popup *and* Wanderer reputation loss happen immediately upon destroying either ship instead of on landing, please point out how. It would be possibe to make the popup happen right away by keeping the `kill` version and putting it in the `dialog` for killing them, but that wouldn't trigger the reputation loss.

Amazinite:
Ref: https://github.com/endless-sky/endless-sky/pull/3583#issuecomment-402276500

The Pug are clearly capable of defending the Wanderers with only a single Arfecta, so I don't see a need to make the destruction/capture of only one of the Arfecta cause the player to become permanently hostile with the Wanderers. To note, #3583 also made it so that the Wanderers have a positive attitude toward the Pug, meaning that the player will also have to fight the Wanderers when fighting the Arfecta. If the player is capable of capturing one of the Arfecta while not killing any Wanderers, I say more power to them.

If the intention is to have the impact of destroying/capturing both Arfecta trigger upon any landing, I think the more sensible change would be to implement the "complete on any landing" mechanic that has been mentioned in the past. (i.e. multiple valid destinations, in this case the multiple destinations being any Wanderer world.)

Fzzr:
In-universe the reason the Wanderers would be offended by attacking the Pug are because they consider them almost like deities, not just because killing the Arfectas leaves them vulnerable. See #4397 for the alternate solution.

Amazinite:
Another issue with this method is that the destruction of one of the Arfecta means the disappearance of the other, as the mission will have been failed, which shouldn't be the case. We could get around that by having a mission for each ship, but then we have to deal with backwards compatibility.

Amazinite:
I'm going to say we just leave this as is for now. The only elegant solution would be to implement the idea from #2131.

PR #4400, opened 2019-06-29
Really minor code comment improvements

petervdmeer:
Some really minor improvements to some comments in the code.
Wanted to include those in other PRs, but didn't have another PR where those would fit in nicely.

### DIFF
--- a/source/UI.cpp
+++ b/source/UI.cpp
@@ -240,7 +240,10 @@ bool UI::IsDone() const
 
 
 
-// Check if there are no panels left.
+// Check if there are no panels left. No panels left on the gamePanels-
+// stack usually means that it is time for the game to quit, while no
+// panels left on the menuPanels-stack is a normal state for a running
+// game.
 bool UI::IsEmpty() const
 {
 	return stack.empty() && toPush.empty();
--- a/source/UI.cpp
+++ b/source/UI.cpp
@@ -240,7 +240,7 @@ bool UI::IsDone() const
 
 
 
-// Check if it is time to quit.
+// Check if there are no panels left.
 bool UI::IsEmpty() const
 {
 	return stack.empty() && toPush.empty();
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -214,12 +214,15 @@ int main(int argc, char *argv[])
 		if(!isFullscreen)
 			SDL_GetWindowSize(window, &windowWidth, &windowHeight);
 		
-		// menuPanels is used for the panels related to pilot creation, preferences,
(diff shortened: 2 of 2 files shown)


source/UI.cpp
```diff
@@ -240,7 +240,7 @@ bool UI::IsDone() const
 
 
 
-// Check if it is time to quit.
+// Check if there are no panels left.
```
MCOfficer: You might want to combine these 2 comments into one. No panels being left kinda implies that there's nothing left to do, but it might be better to explicitly state that.

source/main.cpp
```diff
@@ -214,7 +214,11 @@ int main(int argc, char *argv[])
 		if(!isFullscreen)
 			SDL_GetWindowSize(window, &windowWidth, &windowHeight);
 		
-		
+		// menuPanels is used for the panels related to pilot creation, preferences,
+		// game loading and game saving.
+		// gamePanels is used for the main-panel (flying your spaceship), the panet
```
MCOfficer: this could be split in two comments, one for each declaration.

source/main.cpp
```diff
@@ -214,7 +214,11 @@ int main(int argc, char *argv[])
...
+		// dialog and all other game-content related dialogs are placed on top of them.
+		// If there are menuPanels, then they are shown before gamePanels.
```
MCOfficer: before -> in front of?

source/UI.cpp
tehhowch: I don't feel this is necessary unless it is called in circumstances other than checking if we entered the quit state from UI interactions.

source/main.cpp
tehhowch: `panet`

source/UI.cpp
petervdmeer: The usage of 2 stacks (menuPanels and gamePanels) doesn't automatically make empty() always match quit(). Added some more text to make usage of this function more clear.

source/main.cpp
petervdmeer: Yes, it looks to the user as if the menuPanels are on top of the gamePanels. But the game internally switches between 2 stacks of panels. So technically they are not in front of the other panels. I added some text where I indicated that menuPanels take priority over gamePanels if both stacks exist.

source/UI.cpp
petervdmeer: The auto-test code from #4370 also checks if the menu-stack is empty to determine if the game is running or if we are in a menu.
I extended the text a little bit to make the usage of this code more clear.

PR #4402, opened 2019-06-29
Moved done button on map to far right, adjusted button sizes.

ChamEV:
* Moved done button to far right on map screen to be consistent with other interface elements such as shipyard, outfitters, planet landings, player info screen.
* All click areas for these buttons now match button visible areas.  Previously only the zoom +/- buttons did this.

~~Requires #4401 (removes old button code)~~ (merged)

Current version, 14 extra pixel separation.
image

Initial commit: 
image

Original:
image

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -841,27 +841,27 @@ interface "mission" bottom
 interface "map buttons" bottom right
 	active if "!is shipyards"
 	sprite "ui/wide button"
-		from -450 -50 to -340 0
+		from -464 -50 to -354 0
 	button s "_Shipyards"
-		from -440 -40 to -350 -10
+		from -454 -40 to -364 -10
 	
 	active if "!is outfitters"
 	sprite "ui/wide button"
-		from -350 -50 to -240 0
+		from -364 -50 to -254 0
 	button o "_Outfitters"
-		from -340 -40 to -250 -10
+		from -354 -40 to -264 -10
 	
 	active if "!is missions"
 	sprite "ui/dialog cancel"
-		from -250 -50 to -160 0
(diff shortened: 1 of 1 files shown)


tehhowch:
For UI-altering PRs, please include the relevant before/after screenshots/video so those who don't/can't compile can offer feedback.

ChamEV:
Added screenshots.

Pointedstick:
+1. I'd also like a little bit of horizontal spacing separating the done button from all the others ones on that row.

ChamEV:
Extra spacing for visual distinction, or to reduce accidental clicks?  For the latter, most of the time players enter the map using m which has the ports view open by default which itself is a barrier to accidental clicks.  For the former (visual distinction), I'm 100% indifferent.

Pointedstick:
The former. Not a blocker, but maybe a nice-to-have. :)

ChamEV:
Here's with an extra 20 pixel separation.
image

Here's with an extra 10 pixel separation.
image

I think 20 is too much.

Pointedstick:
To me, 10 is too little, and doesn't quite get the point across. If 20 is too high for you, maybe 15 or 16?

ChamEV:
Here's 14 extra pixels. (24 total).

image

Pointedstick:
Perfect, let's do it.

Amazinite:
Indifferent. I was gonna say that I'd rather not change the UI without actually improving it because I hate when other games do that (this change makes things more consistent but isn't necessary), but I always use key shortcuts anyways. If someone on Steam says something about this change though I'm pointing them in this direction. :)

Pointedstick:
(IMO this is an improvement :) )

PR #4403, opened 2019-07-01
WIP: Add squadrons (groups of identical ships)

petervdmeer:
Add support for squadrons, groups of identical ships.

Initial usage of the squadron would be to define larger NPC fleets where you want multiple identical ships in a single fleet. Another usage would be to have missions where the player is always escorted by a certain ship or group of ships, and when such a ship gets destroyed, then new instances of the ship will appear once the player jumps (starting from a planet where the ships hull is for sale).

Later usage would be to also allow the player to define and use squadrons. The player can command each ship in a squadron he/she owns just like any other ship, but the first ship in the squadron acts as squadron-leader. So any other squadron ship will move to the squadron leaders location if the other ship has no other commands.

A squadron could also consist of 0 ships if all ships in the squadron are destroyed. The squadron definition then still exists and the player can land on a planet where the squadron ships can be bought to get new ships in the squadron from the shipyard in the same way as that ammunition can be bought (buy ships to re-equip your squadrons? Question at the shipyard.).

Buying squadron ships doesn't need to be an exact match with ships in the shipyard. If the hulls and relevant outfits are available (either in the outfitter or as cargo in the players cargo-hold), then the squadron ships can automatically be bought and built for the player.

Squadrons owned by the player can only be started from a ship that the player owns and that is on the same planet as where the player has landed.

### DIFF
--- a/source/Squadron.cpp
+++ b/source/Squadron.cpp
@@ -1,0 +1,32 @@
+/* Squadron.cpp
+Copyright (c) 2019 by Peter van der Meer
+
+Endless Sky is free software: you can redistribute it and/or modify it under the
+terms of the GNU General Public License as published by the Free Software
+Foundation, either version 3 of the License, or (at your option) any later version.
+
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+using namespace std;
+
+#include "DataNode.h"
+#include "Ship.h"
+#include "Squadron.h"
+#include <string>
+
+void Squadron::Load(const DataNode &node)
+{
+	for(const DataNode &child : node)
+	{
+		const string &key = child.Token(0);
+		if(key == "template")
+			templateShip = new Ship(child);
+		else if (key == "desired" && child.Size() >= 2)
+			desired = child.Value(1);
+		else if (key == "actual" && child.Size() >= 2)
+			actual = child.Value(1);
+	}
+}
--- a/source/Squadron.cpp
+++ b/source/Squadron.cpp
@@ -10,12 +10,13 @@ WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
 PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 */
 
-using namespace std;
-
 #include "DataNode.h"
 #include "Ship.h"
 #include "Squadron.h"
+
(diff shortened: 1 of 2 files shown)


petervdmeer:
This is just a proposal at this moment. Implementation still needs to be added if we like this idea.

ChamEV:
The game supports a follow other ship behavior already, for example, click on one of your escorts and right click on another escort or neutral or enemy ship and it will follow/follow/attack that target.  So maybe this enhancement would be to implement a way of saving this behavior between landings, and some sort way to interact with your escorts on the player info screen (which could use some tlc) to create these squadrons.

Perhaps these squadrons also try to fly in formation?  Though formations themselves are hard without the AI cheating and nudging themselves a little in directions they don't have thrusters.

petervdmeer:
> The game supports a follow other ship behavior already, for example, click on one of your escorts and right click on another escort or neutral or enemy ship and it will follow/follow/attack that target. So maybe this enhancement would be to implement a way of saving this behavior between landings, and some sort way to interact with your escorts on the player info screen (which could use some tlc) to create these squadrons.

I think this is a good idea, but not for this PR. If we want to specify behaviors in the player info screen, then I want such functionality to be supported for all player ships, not just for squadrons.
 
> Perhaps these squadrons also try to fly in formation? Though formations themselves are hard without the AI cheating and nudging themselves a little in directions they don't have thrusters.

No, not for this PR. I have some ideas on how to easily implement formation flying (without the AI cheating), but that is something that I don't want to combine with this PR.

> Also, you may want to title this with a "WIP".

Done

source/Squadron.cpp
```diff
@@ -0,0 +1,32 @@
...
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+using namespace std;
+
+#include "DataNode.h"
+#include "Ship.h"
+#include "Squadron.h"
+#include <string>
+
```
ChamEV: There should be an emtpy line between "" includes and <> includes in both header and cpp files.

using namespace std; should come after the includes.

Tadrix:
Would these squadrons fly in some formation, or just swarm their leader?

petervdmeer:
> Would these squadrons fly in some formation, or just swarm their leader?

This PR does not address any formation flying. So if we are talking about just this PR, then the ships would just swarm their leader.

petervdmeer:
The discussion on Discord lead to some ideas with a much simpler implementation (just keep destroyed ships in the ships-list for easy re-buying). This simpler implementation provides most of the benefits here, with a simpler user-interface.

Closing this PR in favor of the simpler implementation with the simpler user-interface.

Issue #4394, opened 2019-06-25
Feature Request: Disabling weapon attribute


Zitchas:
**Is your feature request related to a problem? Please describe.**
- Weapons (especially tail end of missile volleys) often kill ships, especially smaller ships.
- Projectiles seem quite similar and not much benefit to alternate missile types.
- Opportunities for "dueling" weapons (useful for EVN in ES plug-ins

**Related Issue Links**
People having problems with small ships being destroyed by the tail ends of volleys: https://github.com/endless-sky/endless-sky/issues/4149

**Describe the solution you'd like**
An attribute that can be specified for weapons that cause it to do 25% of normal hull damage when the target is disabled. 

**Additional context**
There are three main reasons for implementing this attribute: Lore, Missions, and early-game boarding. A fourth is for plug-ins. 

1) Despite the name "hull" damage, it seems broadly reasonable to assume that this damage includes things such as damage that focuses on crippling shipboard systems such as power, life support, fuel, engines, etc. This is particularly true for weapons that deal ionization, slowing, fuel destruction, or other effects.

As such, it seems reasonable to thing that such systems would be significantly less effective when the target ship isn't functioning. Electrical systems aren't going to suffer as much damage from ionization when they aren't running, etc.

2) On the flip side, the opportunity to have dueling weapons (as in weapons that are blunted, or otherwise have their ultra lethal capabilities reduced or re-calibrated so that they excel at stripping armour but minimize the damage done to life support and other critical systems) would open up a variety of interesting missions where the goal is to disable the other combatant in "fight to disable" style fights. On the flip side, it could be quite interesting to have opponents similarly equipped and set to disable, not destroy, so that the player isn't killed when defeated, but instead has the opportunity to be repaired and try again. While such fights are possible without disabling weapons, it would be possible to send the player up against much more challenging fights (including on par combats) without inducing unnecessary deaths or hardship to newer players.

3) One imbalance in the game is that the boarding and salvaging part of the game (particularly for those using missiles) is significantly handicapped in the early game when the player may be seriously depending on looting fighters, interceptors, and other small craft. The opportunity to buy weapons that are designed to target external armour and avoid critical systems would be a great boon to this early strategy.

Conversely, at the later stages of the game, most players are focused on heavy warships that have sufficient health that a few hits more or less are unlikely to have a significant impact on their ability to salvage from said ships. As such, weapons with these attributes would have little to no impact on late game capturing/salvaging, but would improve the early game capturing/salvaging.

[...]

PR #4404, opened 2019-07-02
feat: Disabled damage type

ChamEV:
Adds a disabled damage type as requested in https://github.com/endless-sky/endless-sky/issues/4394 discussion thread.

Disabled damage is the amount of damage a weapon does to a ship that is disabled.  Disabled damage is proportional to how much damage left in the weapon on that shot.  Ex. if 60% of the damage killed the rest of the shields, and 20% of the damage lowered the hull enough to be disabled, then the final 20% of damage is then applied to the hull and multiplied by the disabled damage attribute.

`disabled damage` is a child of weapon and if not specified defaults to the `hull damage` value, if hull damage is not specified, then they both default to 0.

Weapons that have `disabled damage 0` will never kill ships unless their max hull hitpoints are less than 0.25.  There are no restrictions on negative disabled damage to repair ships, however the AI currently has no idea how to use such weapons.

Here is a video of it in action: https://streamable.com/y35rs

In the outfit shop it is displayed like this:

image

Note, some AI personalities may have an issue if they keep trying to attack with a weapon that cannot kill such as vindictive.


Zitchas:
I'm not enough of a coder to provide meaningful review of the code, but this definitely sounds useful and like something that I definitely want to have access to as a content creator.

Thanks for making this! I'm looking forward to testing it.

ChamEV:
One thing that was a little wordy to me is these tooltips.  There's probably a better way to word this.


```
tip "shield damage / second:"
	`Shield damage, converted to damage per second to allow easy comparisons between different weapons.`

tip "hull damage / second:"
	`Hull damage, converted to damage per second to allow easy comparisons between different weapons.`

tip "   if disabled / second:"
	`Hull damage if disabled, converted to damage per second to allow easy comparison between different weapons.`
```

source/Ship.cpp
```diff
@@ -2640,7 +2642,17 @@ int Ship::TakeDamage(const Projectile &projectile, bool isBlast)
...
 		shieldFraction = min(shieldFraction, shields / shieldDamage);
 	shields -= shieldDamage * shieldFraction;
-	hull -= hullDamage * (1. - shieldFraction);
+	
+	double hullFraction = (1. - shieldFraction);
+	double disabledFraction = 0;
+	double nonDisabledHull = hull - MinimumHull() + .25d;
+	if(hullDamage * hullFraction > nonDisabledHull)
+	{
+		hullFraction = max(nonDisabledHull/hullDamage,0.d);
```
petervdmeer: Division by zero error if hullDamage is 0. (Can happen when nonDisabledHull is below zero on disabled target.)

source/Ship.cpp
```diff
@@ -2640,7 +2642,17 @@ int Ship::TakeDamage(const Projectile &projectile, bool isBlast)
...
+		disabledFraction = 1 - shieldFraction - hullFraction;
+	}
+	hull -= hullDamage * hullFraction + disabledDamage * disabledFraction;
```
petervdmeer: Can we also get the special case when disabledDamage is negative?

I think that would be something like:

```
hull -= hullDamage * hullFraction
// disabled damage might actually restore hulls. The max ensures that restores don't re-enable ships.
if (hull < (MinimumHull() - 0.25d))
  hull = max(hull - disabledDamage * disabledFraction, MinimumHull() - .25d)
```

source/OutfitInfoDisplay.cpp
```diff
@@ -248,6 +248,7 @@ void OutfitInfoDisplay::UpdateAttributes(const Outfit &outfit)
 	static const vector<string> VALUE_NAMES = {
 		"shield damage",
 		"hull damage",
+		"disabled damage",
```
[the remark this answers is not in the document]
ChamEV: No, because it still needs to be a special case because the default value is hull damage, not 0.  See lines 265, 282, 361.

source/OutfitInfoDisplay.cpp
```diff
@@ -261,6 +262,7 @@ void OutfitInfoDisplay::UpdateAttributes(const Outfit &outfit)
 	vector<double> values = {
 		outfit.ShieldDamage(),
 		outfit.HullDamage(),
+		outfit.DisabledDamage() == outfit.HullDamage() ? 0. : 1., // only show if different from hull damage, handled below
```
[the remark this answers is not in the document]
ChamEV: oufit.HullDamage() can return 0 which is a valid value.  If it's 0, then it won't get displayed even though it should.

source/Ship.cpp
```diff
@@ -1648,7 +1648,8 @@ void Ship::DoGeneration()
 	double maxHull = attributes.Get("hull");
 	hull = min(hull, maxHull);
 	
-	isDisabled = isOverheated || hull < MinimumHull() || (!crew && RequiredCrew());
+	if(!isDisabled)
```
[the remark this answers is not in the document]
ChamEV: I think you're right.

source/Ship.cpp
ChamEV: So if a weapon has negative disabled damage (healing) it should only restore it to max disabled health, and would require negative hull damage to heal past that?  I will take a look at that then seems like a good idea.

data/tooltips.txt
```diff
@@ -359,6 +359,9 @@ tip "shield damage / second:"
 tip "hull damage / second:"
 	`Hull damage, converted to damage per second to allow easy comparisons between different weapons.`
 
+tip "   if disabled / second:"
+	`Hull damage if disabled, converted to damage per second to allow easy comparison between different weapons.`
```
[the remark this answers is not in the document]
ChamEV: ok

source/Ship.cpp
ChamEV: This has opened up a new can of worms in determining the order of damage to be dealt.  With SD, HD, DD being either positive or negative giving 8 branches, and also how healthy the ship is, are the shields up, hull > or < minimumHull.

source/Ship.cpp
ChamEV: In the fixed version, nonDisabledHull is always positive so this won't happen, same case with a new variable, healPart.

ChamEV:
Expanded takeDamage() function to handle cases of mismatched signs on weapons.  A weapon can have positive, 0 or negative values for shield, hull, and disabled damage types.  

If shield damage is positive, it will lose it's shields and then other damage types apply.  If negative it will heal the shields and only affect the hull if there is any piercing or disruption.

If hull damage is positive, the ship will eventually be reduced to minimumHull() - 0.25d and be disabled.  If disabled damage is negative, it will never rise above minimumHull() - 0.25d.

> ~Sorry~, I still have some review comments.

Thanks a bunch.

source/Ship.cpp
petervdmeer: Looks good!

ChamEV:
Addressing the AI personality potential issues, I was thinking of doing something like this...

When a weapon is loaded and all it's stats set, we determine these two booleans.
bool Weapon::isLethal() { return isLethal } // Set during loading, lethal if disabled damage > 10% of max(shield,hull) damage.
bool Weapon::isHealing() {} // true if healing outfit flag is set or all 3 damage types are >= 0.
These will also be able to be overidden through two flags, "healing" and "non lethal" allowing the designers to create "healing" weapons that don't actually heal.  Potentially allowing a group of malfunctioning robots to all shoot each other despite being of the same government, eye candy for the player.

When a ship is created/modified, go through each weapon and create a bitmask in the ship's armament of the gun/turret ports of lethality or not so we can do
unsigned_bitmask ship::hasLethalWeapon() { return bitmasklethality }
unsigned_bitmask ship::hasHealingWeapon { return bitmaskhealing }
and get which weapons by bit are each, and simply true if there is one.

This will be a separate project when I get around to it.

petervdmeer:
> Addressing the AI personality potential issues, I was thinking of doing something like this...

> This will be a separate project when I get around to it.

This PR is already quite complete without AI handling, so doing the AI for healing weapons in a separate PR sounds like a good idea. Maybe first file an issue to gather some feedback (and look at all the previous issues already filed with request for AI support for healing weapons)?

I don't think that overriding the flags is a good idea. We already have malfunctioning drones (Syndicate Red and Blue team) fighting each-other and we also already have malfunctioning robotic ships (Korath fractions) fighting each-other (which does give quite some eye candy).

ChamEV:
The flags just tell the AI how to use the weapons.  For non lethal, the designer of a new weapon may want to use a higher level of disabled damage but still have the AI treat it as non lethal.  For the healing flag, it would let ai use a weapon on their own government that drains shields and repairs their hull for example.

But all of this can be discussed in a new issue as you suggested.  For now there are no ships in the game with any of these weapon types so there is no current issue, just future potential ones for very specific cases.

LocalGod79:
i'm pretty sure to have weapons continue behaving as they are, the default value should be 1, not 0.

with 0 as the default, no matter the weapon, nothing ever destroys ships - until you specify disabled damage for all of them - content creators have been building/balancing implemented weapons based on projectiles doing damage to disabled ships for years - don't erase their hard work.

Zitchas:
I can't read the code, but @petervdmeer says that it is pretty much done? How playable is this now?

I also agree with LocalGod79- the default value should be 1 so that weapons (especially older plugins) don't suddenly become useless at destroying things. Other than that, looks good!

source/Weapon.cpp
```diff
@@ -155,6 +156,11 @@ void Weapon::LoadWeapon(const DataNode &node)
 				damage[SHIELD_DAMAGE] = value;
 			else if(key == "hull damage")
 				damage[HULL_DAMAGE] = value;
+			else if(key == "disabled damage")
+			{
+				damage[DISABLED_DAMAGE] = value;
+				disabledDamageSet = true;
+			}
```
Amazinite: @Zitchas @LocalGod79 
`"disabled damage"` is the value of the hull damage done when the target disabled, not a percentage of the standard hull damage that gets done when the target is disabled. Not setting a disabled damage value means that disabled damage defaults to being the same as the hull damage, which is the same behavior as right now. But because of `disabledDamageSet` seen here, one is able to specify that a weapon has a disabled damage of 0 and it will actually work, unlike most other attributes where a value of 0 is the same as if you never mentioned it (which is what you're assuming to be the case). (e.g. `"jump fuel" 0` causes hyperdrives/jump drives to consume the normal 100 or 200 fuel since no such boolean exists to identify if the 0 input is suppose to actually be used or not.)

Zitchas: Thanks for the explanation. 

So the default `disabled damage` is "whatever the hull damage is", not 0. That sounds good to me.

kikotheexile:
This has been suggested as a good fit for the Endless Sky Stories fork. With your permission I would love to add this to that alternate version of the game. If permission is given, I can grab this myself (and credit you for your work) or you could PR it yourself.

Zitchas:
This seems to have gotten lost by the wayside, but I think it would be a valuable addition to our weapon options.

Could we get this updated and conflicts resolved please?

petervdmeer (commented):
I merged master for this PR (and resolved the merge-conflicts). The merge was non-trivial (it even added the new relativeDisabledDamage), so some re-testing is required before I consider this merge-able.

source/DamageProfile.cpp
```diff
@@ -135,9 +135,51 @@ void DamageProfile::PopulateDamage(DamageDealt &damage, const Ship &ship) const
...
+
+			double disabledDamage = (weapon.DisabledDamage()
+			+ weapon.RelativeDisabledDamage() * attributes.Get("hull"))
+			* ScaleType(1., attributes.Get("hull protection")) * disabledFraction;
+
+			// disabledDamage only applies here if it damages.
+			if(disabledDamage >= 0)
+				damage.hullDamage += disabledDamage;
+		}
+	}
```
Amazinite, line 182: I think this is incorrect; you're adding disabledDamage to hullDamage after setting hullDamage to the normal amount, when we should be reducing hullDamage in some way.
```suggestion
	damage.hullDamage = (weapon.HullDamage()
		+ weapon.RelativeHullDamage() * attributes.Get("hull"))
		* ScaleType(1., attributes.Get("hull protection"));
	double hull = ship.HullLevel() - ship.MinimumHull();
	if(hull > 0.)
	{
		if(damage.hullDamage > hull)
		{
			damage.hullDamage *= hull / damage.hullDamage;
			damage.hullDamage += (weapon.DisabledDamage()
				+ weapon.RelativeDisabledDamage() * attributes.Get("hull"))
				* ScaleType(1., attributes.Get("hull protection"));
		}
	}
```
The above code should handle hull damage as shield damage is handled. We check if the hull damage would push us past the disabled threshold (as we check if shield damage would deplete the shields), then multiply the hull damage by the hull fraction (as we multiply shield damage by the shield fraction). *Then* we add the disabled damage to the hull damage.

This doesn't account for "healing" weapons, but neither does shield damage, so I don't see the point in accounting for that here as was done previously.

Amazinite: If we changed HullLevel to return the hull level remaining until disabled (as Hull returns the hull fraction until disabled) then we wouldn't need to make MinimumHull public and the `double hull = ship.HullLevel() - ship.MinimumHull();` line could be changed to just `double hull = ship.HullLevel();`.

Amazinite (changes requested):

source/DamageProfile.cpp
petervdmeer: > This doesn't account for "healing" disabled damage weapons, but healing weapons aren't otherwise handled differently, so I don't see the point in accounting for that here as was done previously.

Well, the "healing" disabled damage weapons were discussed in #4394 as a nice feature to add, but this PR is already valuable when they are not in yet, so I'm okay to keep that for a follow-up PR.

petervdmeer: > DisabledHullLevel()

I guess it should be called HullUntilDisabled(), but I agree with the rest of your comment.

source/Ship.h
```diff
@@ -287,6 +287,10 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	double Health() const;
 	// Get the hull fraction at which this ship is disabled.
 	double DisabledHull() const;
+	// Get the actual hull level of the ship.
+	double HullLevel() const;
+	// Get the hull amount at which this ship is disabled.
+	double MinimumHull() const;
```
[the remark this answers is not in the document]
petervdmeer: I restored the order (by adding the new function from your other review comment).

source/Ship.cpp
```diff
@@ -3020,6 +3020,18 @@ double Ship::DisabledHull() const
 
 
 
+// Get the (absolute) amount of hull that needs to be damaged until the
+// ship becomes disabled, returns 0 if the ships hull is already below the
+// disabled level.
+double Ship::HullUntilDisabled() const
+{
+	// Need to overshoot a little bit, since a ship only gets disabled when
+	// we get _below_ the MinimumHull.
```
Amazinite, line 3029: Using a slightly tweaked comment from what I posted on Discord.
```suggestion
	// Ships become disabled when they surpass their minimum hull threshold,
	// not when they are directly on it, so account for this by adding a small amount
	// of hull above the current hull level.
```

source/Ship.h
```diff
@@ -287,6 +287,10 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	double Health() const;
 	// Get the hull fraction at which this ship is disabled.
 	double DisabledHull() const;
+	// Get the (absolute) amount of hull that needs to be damaged until the
+	// ship becomes disabled, returns 0 if the ships hull is already below the
+	// disabled level.
+	double HullUntilDisabled() const;
```
Amazinite, line 293: Since this is returning a level instead of a fraction, I'd still move it below ShieldLevel so that the fractions (Shield, Hull, Health, DisabledHull) and the levels (ShieldLevel, HullUntilDisabled) are ordered the same way (Shield, then Hull, then Disabled).

source/Ship.h
```diff
@@ -287,6 +287,10 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	double Health() const;
 	// Get the hull fraction at which this ship is disabled.
 	double DisabledHull() const;
+	// Get the (absolute) amount of hull that needs to be damaged until the
+	// ship becomes disabled, returns 0 if the ships hull is already below the
+	// disabled level.
```
Amazinite, line 292:
```suggestion
	// Get the (absolute) amount of hull that needs to be damaged until the
	// ship becomes disabled. Returns 0 if the ships hull is already below the
	// disabled threshold.
```

source/Ship.cpp
```diff
@@ -3020,6 +3020,18 @@ double Ship::DisabledHull() const
 
 
 
+// Get the (absolute) amount of hull that needs to be damaged until the
+// ship becomes disabled, returns 0 if the ships hull is already below the
+// disabled level.
```
Amazinite, line 3025:
```suggestion
// Get the (absolute) amount of hull that needs to be damaged until the
// ship becomes disabled. Returns 0 if the ships hull is already below the
// disabled threshold.
```

petervdmeer:
The files I used for testing:
- Weapon definitions that include a disabling weapon [disabling_weapons.txt](https://github.com/endless-sky/endless-sky/files/8328024/disabling_weapons.txt)
- Savegame with targets and weapons equipped [Test Disabler.txt](https://github.com/endless-sky/endless-sky/files/8328021/Test.Disabler.txt)
All tests looked ok.

PR #4410, opened 2019-07-04
Removing obsolete fourth-wall dialog text

alextd:
It seems to me that the dialog option"I want to explore the rest of the galaxy, but apparently that part of the game hasn't been implemented yet" is probably obsolete dialog, since I am doing that in my game.

(I didn't check when this was added, don't know if there are huge plans for more exploring, didn't verify it works, but I'm pretty sure the format is still okay)

### DIFF
--- a/data/free worlds epilogue.txt
+++ b/data/free worlds epilogue.txt
@@ -30,8 +30,6 @@ mission "FW Epilogue: Ijs and Katya"
 					goto excitement
 				`	"It's kind of hard to adjust to civilian life."`
 					goto adjust
-				`	"I want to explore the rest of the galaxy, but apparently that part of the game hasn't been implemented yet."`
-					goto game
 			
 			label excitement
 			`	"I understand that feeling," says Katya. "But, just remind yourself, this is what we fought for. For peace. For a world where no one has to fear for their lives.`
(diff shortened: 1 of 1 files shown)


Amazinite:
Well there *is* still a lot of galaxy missing.
image

And the intention of that choice seems to largely be about the fact that the game is open source and that you can contribute, and nothing has changed about that, even if new systems have been added to the game since that choice was added.

ZBok:
I think this breaking of the 4th wall is ok and a cool way of letting the player know they can get involved without ruining any storyline vibe as it's already over by then.

I do think the Player dialogue ("I want to explore but that part of the game hasn't been implemented yet") could possibly be worded a little differently, like maybe "I want to explore but it seems that vast areas of the galaxy cannot be reached", while removing Ijs comment and have just Katya break the fourth wall with her response, but that's just my opinion.

petervdmeer:
> I think this breaking of the 4th wall is ok and a cool way of letting the player know they can get involved without ruining any storyline vibe as it's already over by then.

When I encountered this 4th wall dialog, then it did actually break the story-line vibe for me.
You are immersed in the story and you expect some in-game story-line closure, not references to outside the game by in-game characters.

A good thing about the 4th wall dialog is that it encourages players to contribute to the game. The dialog suggested that contributions from players are very welcome and it is one of the things that convinced me to send in some PRs.

Pointedstick:
I think this is something we want to keep. It could be worded with a bit more subtlety, maybe. But overall this should stay in. A little levity and fourth-wall-breaking can be used strategically, and as has been mentioned, it's a good invitation to contribute.

PR #4411, opened 2019-07-04
feat: Style checking script

petervdmeer:
Script that automatically performs a number of style checks from the ES Style Guide on the ES C++ code.

If we can get this script (and other tests) running as an automated check on every incoming new pull-request, then this can save some manual review effort on style.


tehhowch:
One issue with automated style checks is that they still don't replace style reviews (e.g. your recent PR has an issue with incorrect whitespace indentation in the header file). They can certainly help, sure, but there generally isn't an issue past the first PR or two with new contributors. A link to the style guide from a PR template would perhaps be more effective.

As far as running this on incoming pulls, only @endless-sky has repo hook admin access. You can modify the existing Travis and Appveyor configuration files to include this test in the tests that get executed, for those that use those services.

petervdmeer:
Thanks for your response. I never claimed that automated style checks would fully replace style reviews, I only claimed that some manual effort on style reviewing would be saved.

I noticed that I didn't put an example of the output of the tool here yet. Here is an example of the output:

```
peter@_:_/endless-sky$ ./tests/check_coding_style.sh 

******************************************************
Use tabs instead of spaces so that the code can be edited in any editor easily.
******************************************************
DataWriter.h:  template <class A, class ...B>
DataWriter.h:  template <class A>
Point.cpp:    return Point(_mm_andnot_pd(sign_mask, p.v));

******************************************************
Do not put a space before or after the parentheses for functions and control statements.
******************************************************
AI.cpp:	if (it != rosters.end() && !it->second.empty())
Engine.cpp:	for (const shared_ptr<Ship> &ship : ships)
PreferencesPanel.cpp:		if (Screen::Zoom() != zoom)
System.cpp:		else if (object.IsStation())
System.cpp:		else if (object.IsMoon())

******************************************************
Put a space after commas.
******************************************************
Font.cpp:			leftChars = min(prevChars,nextChars) / 2;
```

How do you suggest we proceed with this PR? Is there something I should still do?

petervdmeer:
> (e.g. your recent PR has an issue with incorrect whitespace indentation in the header file).
Btw, did I already correct this? I don't intentionally put style issues in header files, so if you noticed a style issue that I didn't fix yet, then there is a good chance that I haven't seen it yet and that I don't know about it yet.

tehhowch:
No one intentionally introduces style issues 🙂
Especially with whitespace, it's usually a case of IDEs "helping" a bit too much. Even with an .editorconfig file to help mitigate some behaviors, the respective plugins still don't always work correctly.

There's also the include order (own class header, then alphabetical first-party, then alphabetical third-party, then system headers). But adding this check only increases the level of nitpickery, and probably isn't a simple addition either.

petervdmeer:
> There's also the include order (own class header, then alphabetical first-party, then alphabetical third-party, then system headers). But adding this check only increases the level of nitpickery, and probably isn't a simple addition either.

Not a simple addition indeed (and Bash is not the best language for comparing large lists). I did add a check for it, but I need to do an update to tell the sorting mechanism the proper order for ShipEvent.h, Ship.h and ShipInfoPanel.h.

MCOfficer:
At this point, it might be viable to look towards a dedicated testing system rather than plain bash.

petervdmeer:
> At this point, it might be viable to look towards a dedicated testing system rather than plain bash.

Yes, Bash clearly has it's limitations. It is okay for the simple checks, but this last check was a bit much. I added comments in the script itself about future-work/extending to make the style-checker more professional.

I also just fixed the sorting issue. The fix itself was not really difficult, just something I didn't get to do yesterday. For now I would recommend using the style-checking script as-is until somebody feels a very strong need to rewrite (or parts of) it.

petervdmeer:
@tehhowch The message you get is actually great news. I never designed the script to run on Windows, but it:
- Apparently does run on Windows
- Finds the correct files in the correct directory (which means that the directory magic works)
- Actually gives mostly correct output (the style-error is about class definitions and it selects lines with classes on them).

The message you see means that either this part of line 226 `grep -v -e ";$"` is not working (this regular expression should filter out lines that end with a semicolon) or that you have an outdated version of the script. Can you post the full output when you run the script (so that I can check which of the checks are working as expected and which ones need an update)?

MCOfficer:
I cannot test the entire script atm, but the grep command is working as expected for me. my grep is version 2.5.4 from gnuwin32.

petervdmeer:
It looks like the grep command was working fine, also for Tehhowch, but the two grep-commands that were not working as expected were checking/matching for end-of-line characters. I updated both those commands to run tr before the grep to remove the Windows-specific carriage-return.

petervdmeer:
Tehhowch also reported that the script took 60 minutes to finish. I removed the function with the worst performance. The script should now finish a lot quicker.

petervdmeer:
Be aware that this script is still on an old baseline (with coding style issues that were fixed later). The issues you should find with the latest version of the script on it's current baseline are:

```
******************************************************
Each file should begin with a header which specifies the file name, copyright, and licensing information
******************************************************
BatchDrawList.cpp: mismatch in header
Body.cpp: mismatch in header
CaptureOdds.cpp: mismatch in header
CargoHold.cpp: mismatch in header
CollisionSet.cpp: mismatch in header
DataNode.cpp: mismatch in header
DataWriter.cpp: mismatch in header
Dictionary.cpp: mismatch in header
Format.cpp: mismatch in header
Messages.cpp: mismatch in header
Messages.h: mismatch in header
MissionAction.cpp: mismatch in header
Personality.cpp: mismatch in header
Politics.cpp: mismatch in header
SavedGame.cpp: mismatch in header
shift.h: mismatch in header
Table.cpp: mismatch in header

******************************************************
All header files should have #define guards, and they should be in the format "#ifndef MY_CLASS_H_".
******************************************************
ItemInfoDisplay.h: #ifndef INFO_DISPLAY_H_
ItemInfoDisplay.h: #define INFO_DISPLAY_H_

******************************************************
Use tabs instead of spaces so that the code can be edited in any editor easily.
******************************************************
Point.cpp:    return Point(_mm_andnot_pd(sign_mask, p.v));

******************************************************
Do not put a space before or after the parentheses for functions and control statements.
******************************************************
AI.cpp:	if (it != rosters.end() && !it->second.empty())
Engine.cpp:	for (const shared_ptr<Ship> &ship : ships)
PreferencesPanel.cpp:		if (Screen::Zoom() != zoom)
System.cpp:		else if (object.IsStation())
System.cpp:		else if (object.IsMoon())

******************************************************
Put a space after commas.
******************************************************
Font.cpp:			leftChars = min(prevChars,nextChars) / 2;
```

petervdmeer:
The check now seems to run in .github actions/workflows (and doesn't find any issues in the latest master).

.github/workflows/style_check.yml
```diff
@@ -0,0 +1,24 @@
...
+    - 'source/**'
+    - 'tests/**'
+  pull_request:
+    paths:
+    - 'source/**'
+    - 'tests/**'
+
+jobs:
+  check-style:
+    runs-on: ${{ matrix.os }}
```
tehhowch: Remove the strategy and matrix bits, just make this `ubuntu-latest`

tests/check_coding_style.sh
```diff
@@ -0,0 +1,192 @@
...
+# used to deal with such false positives and can also be used for allowed deviations.
+
+
+# Determine path of the current script
+HERE=$(cd `dirname $0` && pwd)
+SOURCES="${HERE}/../source"
+
+# Desired header, adapted for automatic parsing
+DESIRED_HEADER="/* FILENAME
+Copyright (c) 2xxx by Michael Zahniser
```
tehhowch: It is not a strict requirement that all files are solely copyright MZ. Is that going to be flagged by this?

tests/check_coding_style.sh
```diff
@@ -0,0 +1,192 @@
...
+		-e "#include <dirent.h>" \
+		-e "#include <unistd.h>" \
+		-e "#include <mutex>" \
+		-e "#include <jpeglib.h>" \
+		-e "#include <png.h>" \
+		-e "#include <mad.h>" \
+		-e "#include \"Files.h\"" \
+		-e "main.cpp: mismatch in header" \
+		-e "Color.cpp:#include \"Color.h\"" \
+		-e "Rectangle.cpp:#include \"Rectangle.h\"" \
```
tehhowch: How many of these exceptions should be kept vs just fixed?

tests/check_coding_style.sh
```diff
@@ -0,0 +1,192 @@
...
+	while read ISSUE; do
+		# Print the error-message once
+		if [ "${ERROR_MESSAGE}" != "" ]; then
+			echo ""
+			echo "******************************************************"
+			echo "${ERROR_MESSAGE}"
+			echo "******************************************************"
+			ERROR_MESSAGE=""
+		fi
+		echo "${ISSUE}"
```
tehhowch: Does it make sense to report the total number of issues?

petervdmeer: I don't think it makes a lot of sense; if the script reports more than zero issues (any number above zero), then some attention is required from the author of the affected PR. So the only relevant information is if the script reports zero issues or more-than-zero issues.

tests/check_coding_style.sh
petervdmeer: The struct ones really are an exception (the coding style recommends not using structs).
I moved WinApp.rc to the right location.
The #include checks were removed for performance reasons, so they no longer needed to be listed here (I removed them)
I fixed an issue for the using std which also removed an exception.
And I removed an exception regarding tabs/spaces but left the actual issue in the code to check/verify the style checking script.

tests/check_coding_style.sh
petervdmeer: The text now says:
`Copyright (c) 2xxx by NAME`
Do you think that is sufficiently clear?

tests/check_coding_style.sh
```diff
@@ -0,0 +1,169 @@
...
+find -type f | grep -v -e ".h$" -e ".cpp$" -e "WinApp.rc" | sed "s,^./,," |\
+	report_issue "Files: All C++ files should use \".h\" and \".cpp\" for their extensions."
+
+grep "^class" *.h | tr -d '\r' | grep -v -e ";$" | cut -d\: -f1,2 | sed "s,[{].*$,," | sed "s, *$,," | filter_non_matched_class |\
+	report_issue "Files: Each class \"MyClass\" should have its own .h and .cpp files, and they should be named \"MyClass.h\" and \"MyClass.cpp\""
+
+for FILE in *.{h,cpp}; do
+	HEADER=$(head -n 11 ${FILE})
+	HEADER_CHECK=$(echo "${HEADER}" | sed "s,^/\* ${FILE}$,/* FILENAME," |\
+		sed "s,^Copyright (c) 2[0-9]* by [A-Za-z0-9 ]*,Copyright (c) 2xxx by NAME,")
```
tehhowch: This regex seems overly restrictive, e.g. it will flag any names with non-space punctuation.

I would prefer matching all "copyright by ..." lines and stripping them out entirely.

petervdmeer: Fixed by allowing any valid character except newlines in copyright holder names.

oo13:
You mean we cannot use © as copyright symbol?
Technically speaking, "(c)" is completely meaningless.

Anyway it's no matter because this copyright notice includes "Copyright".

petervdmeer:
> You mean we cannot use © as copyright symbol?

Indeed, the copyright symbol is in not in the basic ASCII set and thus outside the range allowed for source files by the ES style guide.

> Technically speaking, "(c)" is completely meaningless.

It does have some meaning, just not the same legal meaning as the copyright symbol or the word copyright.

> Anyway it's no matter because this copyright notice includes "Copyright".

It doesn't really matter, but I just pushed an update to also allow headers with just the word copyright and without the "(c)" construction.

oo13:
> > You mean we cannot use © as copyright symbol?
> 
> Indeed, the copyright symbol is in not in the basic ASCII set and thus outside the range allowed for source files by the ES style guide.

Fair enough, but #4123 has this copyright notice.

> Copyright (c) 2017, 2018 by Flávio J. Saraiva

I'll change it to next one. Probably he doesn't care it.

> Copyright (c) 2017, 2018 by Flavio J. Saraiva

Oh, may I use ',' and '-' in the first publish years?

petervdmeer:
> Oh, may I use ',' and '-' in the first publish years?

I just pushed an update that should teach the style-checking script that such characters are okay for date ranges.

petervdmeer:
Closing: The effort still needed to get this to accepted/merged seems more than the benefits we will currently get from running this type of check automatically.

PR #4414, opened 2019-07-05
Always select the first available mission when clicking a new system.

alextd:
Without this change, it's unintuitive and feels inconsistent whether clicking a new system selects an available mission or an active mission. The way it works internally, the game holds the selected mission as an iterator (availableIt + acceptedIt basically acting as a single iterator). Clicking the same system again simply iterates, and that's fine. But when you click a new system, it finds the NEXT mission in the list that involves the new system - whether or not it's in the available list or accepted list. This makes what mission is selected when clicking a new system depend on the order of the mission list. That feels unintuitive. So, to make it consistent, the iterators are simply reset when selecting a new system.

For example: Say you have a long list of available and active missions for the new system, and you're selecting some mission for another system.
* If the previously selected mission is higher in the available list, it will go down the list and find a new available mission for the new system.
* If the previously selected mission is at the end of the available list, it will switch to the active mission list and find an existing active mission for the new system.

(Fair warning: I didn't compile this, couldn't get a project set up, but it seems to be a super simple change.)

### DIFF
--- a/source/MissionPanel.cpp
+++ b/source/MissionPanel.cpp
@@ -327,6 +327,13 @@ bool MissionPanel::Click(int x, int y, int clicks)
 		}
 	if(system)
 	{
+		// Always go to the first available mission when clicking a new system.
+		if(selectedSystem && selectedSystem != system)
+		{
+			availableIt == available.end();
+			acceptedIt == accepted.end();
+		}
+
 		Select(system);
 		int options = available.size() + accepted.size();
 		// If you just aborted your last mission, it is possible that neither


alextd:
The main reason for this, of course - when I am on the mission screen and I click a system, I want to select a new available mission, not a mission I'm already on. When I have 5 active missions to click through, it becomes infuriating that I have to click so much just to select the one new mission I want to accept.

alextd:
Now that I think of it - this will also make selecting a system with no mission change the selection to the top of the list. So that's something to think about.

tehhowch:
Selecting a system with no mission should not change the existing selection, so with that behavior this won't merge.

alextd:
I'm not going to do this myself so here is an issue for this change : https://github.com/endless-sky/endless-sky/issues/4464

PR #4417, opened 2019-07-06
Ship::JumpsRemaining() calculates the jump counts a bit more optimistically.

oo13:
This PR fixed 2 issues that tehhowch mentions in https://github.com/endless-sky/endless-sky/pull/4288#pullrequestreview-258137829.

If an escort has no destination, Ship::JumpsRemaining() try to use the parent's destination.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2407,11 +2407,21 @@ double Ship::DisabledHull() const
 
 
 
-int Ship::JumpsRemaining() const
+int Ship::JumpsRemaining(bool followParent) const
 {
 	// Make sure this ship has some sort of hyperdrive, and if so return how
 	// many jumps it can make.
-	double jumpFuel = JumpFuel(targetSystem);
+	double jumpFuel = 0.;
+	if(!targetSystem && followParent)
+	{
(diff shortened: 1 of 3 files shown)


source/Ship.cpp
```diff
@@ -2407,11 +2407,21 @@ double Ship::DisabledHull() const
...
 	// many jumps it can make.
-	double jumpFuel = JumpFuel(targetSystem);
+	double jumpFuel = 0.;
+	if(!targetSystem && followParent)
+	{
+		// If this ship has no destination, the parent's substitutes for it,
+		// but only if the location is reachable.
+		auto p = GetParent();
+		if(p)
+			jumpFuel = JumpFuel(p->GetTargetSystem());
```
ChamEV: GetTargetSystem() will return nullptr if the parent arrives at it's destination.  It would seem that you would use parent's current system in that case?

oo13: I'm not sure that escorts should go to the current parent's system and if it has meaningfulness in what kind of condition.

oo13:
The side effects were caused when escorts cannot reach directly the parent's target. This PR avoids its condition.

I didn't consider another implementation.

PR #4418, opened 2019-07-06
Guard positions for escorts around lead ships

petervdmeer:
Add guard (position) functionality to allow escort ships to do stationkeeping on a relative offset to their lead ship.

Here is a screenshot of 4 escorts on "guard positions":
0, 500
0, -500
-500, 0
500,0

guard_positions

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1830,9 +1830,18 @@ void AI::KeepStation(Ship &ship, Command &command, const Ship &target)
 	double mass = ship.Mass();
 	Point unit = ship.Facing().Unit();
 	double currentAngle = ship.Facing().Degrees();
+
+	// Target position, either the target-ship, or a relative position compared to
+	// the target ship if such a relative position is set
+	Point targetPosition = target.Position();
+	Point targetVelocity = target.Velocity();
+	Point guardPosition = ship.GetGuardPosition();
+	if (guardPosition)
+		targetPosition += (Angle(targetVelocity)).Rotate(guardPosition);
+
 	// This is where we want to be relative to where we are now:
-	Point velocityDelta = target.Velocity() - ship.Velocity();
-	Point positionDelta = target.Position() + LEAD_TIME * velocityDelta - ship.Position();
+	Point velocityDelta = targetVelocity - ship.Velocity();
+	Point positionDelta = targetPosition + LEAD_TIME * velocityDelta - ship.Position();
 	double positionSize = positionDelta.Length();
 	double positionWeight = positionSize / (positionSize + POSITION_DEADBAND);
 	// This is how fast we want to be going relative to how fast we're going now:
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -325,6 +325,13 @@ void Ship::Load(const DataNode &node)
 			description += child.Token(1);
 			description += '\n';
 		}
+		else if(key == "guard")
+		{
+			if(child.HasChildren())
+				for(const DataNode &grand : child)
+					if(grand.Token(0) == "position" && grand.Size() >= 3)
+						guardPosition = Point(grand.Value(1), grand.Value(2));
+		}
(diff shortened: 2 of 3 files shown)


petervdmeer:
Thanks for the good review feedback.

> I don't feel this is ready for merging without any manner of manipulation via the UI.

Manipulation via the UI is indeed required to allow the player to use this feature. I have some ideas on how to add that, but I first wanted to get the basic mechanism reviewed and in place. A next step that I had considered was to have some NPC fleets use this feature. This mechanism doesn't change any existing behavior, so I was actually hoping that I could get this merged without UI support and that I could handle the UI support in a separate PR.

> Nor do I agree that the Ship class is where "guard point" information should be stored (AI is responsible for in-game ship logic decisions, including the position a ship desires to travels toward).

My impression was that the AI object/class handles the "how" of what an in-game ship is doing, but that the "what" is stored in the Ship-objects, the Order-objects and the Command-objects. The Ship already stores some of the "what" in variables like targetShip, shipToAssist, targetPlanet.

We could argue that formations is a fleet property or even a property of the lead ship (communicating the formation to its team members). I took a different conceptual approach in that each ship/pilot should know its own place in a formation, so that the information about where to fly is then stored with the relevant ship/pilot.

Maybe the term "guard position" is not so suitable. I was initially implementing this option under the name "idle position"; a position (absolute or relative) that the ships moves to if it has nothing better to do. Guard position (a position where a ship stands guard) however seemed to sound better.

> Perhaps a more fitting implementation would be:
> * "formation" root-level datanode definition (held by GameData after being read from the .txt files)
>    * e.g. `formation <name>` with child nodes indicating `x` and `y` offsets in a standard Cartesian grid, 

This adds quite some levels of indirection. A big benefit of your proposal is that if a ship in a formation is destroyed, then other ships can move up to fill the gaps (from later/less important formation positions). But I see a lot of complexity as well and I'm not convinced that that is needed.

> whether the formation is held when attacked, etc

I was considering to let this be handled by a different PR for ships personalities. Bodyguard personality stays with the guard position as much as possible. Heroic personality just goes after any enemy it sees.

> "formation" token for NPC and/or Fleet, indicating the formation name(s) used by the relevant ships

I have another branch (not sent to Github yet, not as PR yet) where I was trying an approach where formations are set on NPC fleets. Setting a formation on an NPC fleet seems like a good idea. My approach there was however to calculate the formation on fleet-instantiate time and then handle the formation as guard positions of the individual ships while the fleet was active. A nice feature of NPC fleets is that they automatically get the next ship in line as lead ship if the previous one was destroyed.

> This then allows dynamic assignment of ships to formation positions within click handling & future interfaces. 

The implementation proposed here also allows assigning ships to formation positions by click handling. Any position around any other ship can be set as a position.

> It doesn't support guarding of fixed positional references in a system, but I would expect different behavior between ships in said orbital roles and those that fly in formation.

You are right for attack formations.
But I would expect the same behavior for protective type of formations. Protect a planet, protect a transport-ship, protect an NPC escort ship, protect a players flagship, protect newly captured ships. The difference is that planets don't rotate and don't move during a single day, but for the rest you just surround the thing that is protected with a group of your ships.

> For future PRs: please use the "Draft PR" dropdown option

Cool, I didn't notice this "Draft PR" feature before. I will use this next time for PRs that are for examples/discussions instead of for immediate merging.

> You have a very lengthy & detailed commit message here that seems much more appropriate for the PR text.

Sorry about that. I was doubting if I should put a lot of text in the PR text or keep it as simple and short as possible. Some of my other PRs seemed to contain a bit too much text in the PR.

petervdmeer:
Temporary closing this PR until I have some time to add some more details.

Pointedstick:
Instead, you could just mark it as a draft.

petervdmeer:
> Instead, you could just mark it as a draft.

Oh, is that also possible for PRs already submitted? Nice.

Issue #3487, opened 2018-01-03
Display ramscoop in ship info


Nescio0:
It would be nice if ramscoop values would be displayed in the ship panel and info screens, perhaps on the same line as fuel, e.g.: `ramscoop / fuel capacity : 4 / 600` (cf. hull repair). Some ships have ramscoop values of their own and currently the only way to figure this out is by looking directly into the data files (e.g. the Kar Ik Vot 349 I seized turned out to have a ramscoop of 3).

endless-sky:
The fuel line already shows `current / maximum` fuel if you're in flight, so I can't really add ramscoop as a third value.

Nescio0:
How about displaying ramscoop on a new line then?

Nescio0:
Although it's not the most important stat, it's not unimportant either. And currently there is no way to look up the actual ramscoop values in game, despite the fact that about fifteen ships have an integrated ramscoop. The Starling's integrated cloaking is at least mentioned in its description.

KiLEdEnNis:
I get that ya want to be able to see it in-game without looking in the data files, but like jafdy said it's not really an important stat and I've never had any desire to look at it

DingusShingleton:
Funny, I asked about this very thing almost exactly 2 years ago. #624. The only practical reason I had was to be able to compare afterburner usage to ramscoop generation. But yeah, not super important.

Zitchas:
I, too, would like to have this information available somewhere. 

Personally, the ideal would be in the player info box alongside the max fuel, but at least having an info line in the Shipyard listing its ramscoop value would be useful.

Given the choice between knowing what my current ramscoop value is and knowing how much fuel I currently have, I would pick the ramscoop. After all, my current fuel is very often identical to my max fuel. But especially with the ability to frequently change ships based on what I'm doing, knowing whether or not the ship I am currently in has any (significant) fuel generating abilities is quite important.

ChamEV:
I would like to add this, the only issue with this is space required and wording.  There are 4 possible variations required for the 3 attributes,

in flight:
fuel / gain / capacity: x/y/z
fuel / capacity: x / z
landed:
fuel gain / capacity: y/z
fuel capacity: z

With this wording, the first one runs out of space in some cases:
![image](https://user-images.githubusercontent.com/51398924/60752938-d504c280-9f80-11e9-9c5c-76721ebc556a.png)

So I'm looking for suggestions of better wording/synonyms that make sense and are compact.

[...]

Issue #4215, opened 2019-03-05
Advanced stats page


Amazinite:
It's been discussed in the past how we can't provide too much information as to overwhelm new players, but at the same time we can't provide too little information to those interested in having more information. I suggest that in the same way we have multiple pages of preferences now, we somehow have a way to have multiple pages of stats, with the most important stats being shown first (i.e. everything shown now), in addition to a second page of "advanced stats." 

Related issues: #3999, #3834, #3835, #3704, #1418, and I'm sure many more.

Ferociousfeind:
FIRST!!

Ahem, what I wish to suggest is actually a preference. In the menu, something like "advanced outfit stats" and "advanced ship stats" which are, by default, ticked off. If we want to provide the current stats for newer players, while also being able to give veterans the advanced information they want to know about, a simple toggle preference ought to do us good.

If not, the toggle can be with the other in-outfitter options, (like "show outfits in cargo" and "show outfits for sale") to allow players to swap between the two information modes much more easily.

If this has already been suggested before (as I believe it might've been) then feel free to ignore me, I think.

comnom:
I like the idea. I'd even go one step beyond and use it to cut out some extraneous info in the default display. Should probably be done in tandem with https://github.com/endless-sky/endless-sky/issues/3108 . A quick mockup:

A subtle clickable element is easy enough:<details>![ellipsis](https://user-images.githubusercontent.com/22107649/53821442-cdce9900-3f3b-11e9-9edf-4a6472e4adfc.png)</details>

Add in a tooltip:<details>![tooltip](https://user-images.githubusercontent.com/22107649/53821459-d32be380-3f3b-11e9-94ae-1e6c88b081d9.png)</details>

Then just pop up a new panel with the image and the full list of stats.

Amazinite:
What about for stuff like #3999 or #3835 where the stats are on your ship as the result of various installed outfits? Where would the button be for that?

comnom:
>What about for stuff like #3999 or #3835 where the stats are on your ship as the result of various installed outfits? Where would the button be for that?

The Shipyard and PlayerInfo panels I'd imagine. Same ship stats panel, but accessible from either.

Amazinite:
#4519 #4520 #4521 #4522 #4523 #4524  could all use this.

Zitchas:
[...]

PR #4421, opened 2019-07-07
Updated fuel attribute line to include fuel gain.  Updated associated tooltips.

ChamEV:
This updates adds fuel gain to the fuel attribute display line defined in ShipInfoDisplay.cpp and is used in the shop side panel and player info panel to describe the current fuel status.  #3487

The primary difficulty of this is simply getting the wording right and ensuring it fits all on one line as it would be inappropriate to use two lines for this.  The best I came up with is as follows.  Any better wording suggestions would be appreciated.

landed ships:
`"fuel gain / capacity:"`
`"fuel capacity:"` (not currently shown)
in flight:
`"fuel / gain / max:"`
`"fuel / capacity:"` (not currently shown)

"fuel" has a double use as the overall line description, and while in flight the individual fuel level.
"capacity" is preferred over "max" as it's more descriptive, but is substituted with "max" for the in flight version due to space constraints.

The fuel gain line exposes the player to the fact that every ship in the game has a default 0.4 fuel gain per second if they're close to a star.  Hence the (not currently shown) above.

image

Updated all 4 possible tooltips and defined fuel gain in them.  Notably this includes the new attributes `fuel generation` and `fuel consumption`.

`Fuel gain is the sum of fuel generation, consumption and the performance of any ramscoops installed while near an average star.`

image
image

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -95,11 +95,17 @@ tip "    engine capacity:"
 tip "fighter bays:"
 	`The number of ships from the "Fighter" category that this ship can carry.`
 
-tip "fuel:"
-	`This ship's current fuel amount. An ordinary hyperdrive jump consumes 100 fuel.`
+tip "fuel gain / capacity:"
+	`This ship's average fuel gain per second and total fuel capacity. An ordinary hyperdrive consumes 100 fuel per jump. Fuel gain is the sum of fuel generation, consumption and the performance of any ramscoops installed while near an average star.`
 
 tip "fuel capacity:"
-	`An ordinary hyperdrive jump consumes 100 fuel. Fuel is replenished for free when you land on an inhabited planet.`
+	`This ship's total fuel capacity. An ordinary hyperdrive consumes 100 fuel per jump.`
+
+tip "fuel / gain / max:"
+	`This ship's current fuel level, average fuel gain per second, and total fuel capacity. An ordinary hyperdrive jump consumes 100 fuel per jump. Fuel gain is the sum of fuel generation, consumption and the performance of any ramscoops installed while near an average star.`
+
+tip "fuel / capacity:"
+	`This ship's current fuel level and total fuel capacity. An ordinary hyperdrive jump consumes 100 fuel per jump.`
 
 tip "fuel consumption:"
 	`Fuel consumed per second while in flight (can generate energy or heat).`
--- a/source/ShipInfoDisplay.cpp
+++ b/source/ShipInfoDisplay.cpp
@@ -22,6 +22,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Table.h"
 
 #include <algorithm>
+#include <cmath>
 #include <map>
 #include <sstream>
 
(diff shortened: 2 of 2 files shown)


finite-galaxy:
Keep in mind that whilst fuel generation and consumption are constant, gain from ramscoop varies greatly from system to system; see `data/stars.txt`, therefore it's misleading to include it in the same line.

What I did in my version is display fuel values alongside energy and heat in the table, but display ramscoop on a separate line:
Screenshot from 2019-07-14 12-58-13
(Just something to look at; I'm not implying you should do the same.)

ChamEV:
> Keep in mind that whilst fuel generation and consumption are constant, gain from ramscoop varies greatly from system to system; see data/stars.txt, therefore it's misleading to include it in the same line.

The important part for players is to be able to compare ships and outfits.  The tooltip explains how ramscoops work, as well as any ramscoop outfit.

As for extra lines, there is a need for listing all the new attributes ships can have, but something hasn't been designed for it UI wise yet.  I was trying to avoid adding any extra lines in this.

Tadrix:
Gotta say, the way finite-galaxy did it is overwhelming. As a new player I'd feel frightened by all these numbers. Somehow, the experience of learning in Endless Sky was much more pleasant.

Ergo: Don't squeeze in too much information. If you absolutely must, there should be a toggle in preferences, "show advanced information", or something like that, that would be disabled by default.

Amazinite:
>Ergo: Don't squeeze in too much information.

And you can tell that you've squeezed in too much when the UI can't even contain it all...
Shield numbers clipping into the text to its left. "fuelenergy." The numbers in the rows of that table are also getting dangerously close to one another.

finite-galaxy:
Fundamentally there are three approaches:
* do not display a stat (endless-sky)
* incorrectly display a stat, e.g. ramscoop as “average” fuel gain (this draft)
* display a stat, using additional lines, if necessary (finite-galaxy)
None of these is perfect, I can see arguments in favour and against each approach.

> Gotta say, the way finite-galaxy did it is overwhelming. As a new player I'd feel frightened by all these numbers. Somehow, the experience of learning in Endless Sky was much more pleasant.

Stats are only shown if a ship actually has them, so starter ships have fewer lines, e.g. shuttles in finite-galaxy:
Screenshot from 2019-08-03 16-14-20
For comparison shuttles in endless-sky:
Screenshot from 2019-08-03 16-16-51
(As you can see both have the same number of lines in this example.)

> "fuelenergy." The numbers in the rows of that table are also getting dangerously close to one another.

Thanks for the feedback, I've now added 10 pixels to the energy column.

Zitchas:
Given the increased number of things that we'd like to see in the information panel, I'm increasingly of the opinion that the shipyard should have an "advanced" panel that pulls up an info screen similar to the player's that gives all this info on a much more spacious environment. Thus this small area can simply be the summary of major elements and the advanced panel can contain all the details. (along with perhaps a scene including the ship or other graphical elements)

Amazinite:
#4215

Amazinite:
While I like the idea of giving more information like this to the player about each of their ships, I think that we could have a more comprehensive solution that provides more information than just the amount of fuel gain that a ship has on this single line. I linked #4215 as an example of how we could go about providing this information without increasing the amount of clutter on the existing abbreviated list of ship stats.

#6587 takes a first step toward providing an advanced stats page, which could include this information about fuel gain. As such, I'd prefer to handle the underlying issue of this PR in that one.

PR #4423, opened 2019-07-07
Pre rendered rotation for Escape Velocity Nova total conversion

julbouln:
Hi,

It is a small patch to be able to render isometric sprites from original Escape Velocity Nova data with the total conversion script I'm working on.

The related total conversion script is available and working without this patch and it is playable even if some functionalities are still missing : https://github.com/julbouln/evn2es

Regards

### DIFF
--- a/source/Body.cpp
+++ b/source/Body.cpp
@@ -201,6 +201,10 @@ void Body::LoadSprite(const DataNode &node)
 			frameOffset += static_cast<float>(child.Value(1));
 			startAtZero = true;
 		}
+		else if(child.Token(0) == "pre rendered rotation" && child.Size() >= 2)
+		{
+			preRenderedRotation = static_cast<int>(child.Value(1));
+		}
 		else if(child.Token(0) == "random start frame")
 			randomize = true;
 		else if(child.Token(0) == "no repeat")
@@ -236,6 +240,8 @@ void Body::SaveSprite(DataWriter &out, const string &tag) const
 			out.Write("no repeat");
 		if(rewind)
 			out.Write("rewind");
+		if(preRenderedRotation);
+			out.Write("pre rendered rotation", preRenderedRotation);
 	}
 	out.EndChild();
 }
@@ -257,6 +263,10 @@ void Body::SetSwizzle(int swizzle)
 	this->swizzle = swizzle;
 }
 
+int Body::GetPreRenderedRotation() const
+{
+	return preRenderedRotation;
+}
 
 
 // Set the frame rate of the sprite. This is used for objects that just specify
@@ -298,21 +308,20 @@ void Body::UnmarkForRemoval()
 }
 
 
-
 // Set the current time step.
 void Body::SetStep(int step) const
 {
 	// If the animation is paused, reduce the step by however many frames it has
 	// been paused for.
 	step -= pause;
-	
+
 	// If the step is negative or there is no sprite, do nothing. This updates
 	// and caches the mask and the frame so that if further queries are made at
 	// this same time step, we don't need to redo the calculations.
 	if(step == currentStep || step < 0 || !sprite || !sprite->Frames())
 		return;
 	currentStep = step;
-	
+
 	// If the sprite only has one frame, no need to animate anything.
 	float frames = sprite->Frames();
 	if(frames <= 1.f)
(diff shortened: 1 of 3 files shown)


source/Body.cpp
```diff
@@ -201,6 +201,10 @@ void Body::LoadSprite(const DataNode &node)
 			frameOffset += static_cast<float>(child.Value(1));
 			startAtZero = true;
 		}
+		else if(child.Token(0) == "pre rendered rotation" && child.Size() >= 2)
```
Pointedstick: `pre rendered` -> `pre-rendered`

source/Body.cpp
```diff
@@ -236,6 +240,8 @@ void Body::SaveSprite(DataWriter &out, const string &tag) const
 			out.Write("no repeat");
 		if(rewind)
 			out.Write("rewind");
+		if(preRenderedRotation);
+			out.Write("pre rendered rotation", preRenderedRotation);
```
Pointedstick: `pre rendered` -> `pre-rendered`

source/Body.cpp
```diff
@@ -298,72 +308,81 @@ void Body::UnmarkForRemoval()
...
 
-
 // Set the current time step.
 void Body::SetStep(int step) const
 {
 	// If the animation is paused, reduce the step by however many frames it has
 	// been paused for.
 	step -= pause;
-	
+
```
Pointedstick: Unrelated whitespace change

source/Body.cpp
```diff
@@ -298,72 +308,81 @@ void Body::UnmarkForRemoval()
...
 	// If the step is negative or there is no sprite, do nothing. This updates
 	// and caches the mask and the frame so that if further queries are made at
 	// this same time step, we don't need to redo the calculations.
 	if(step == currentStep || step < 0 || !sprite || !sprite->Frames())
 		return;
 	currentStep = step;
-	
+
```
Pointedstick: Unrelated whitespace change

source/Body.cpp
```diff
@@ -298,72 +308,81 @@ void Body::UnmarkForRemoval()
...
-		if(!repeat)
-			frame = min(frame, lastFrame);
-		else if(frame >= frames)
+
+	if(preRenderedRotation) {
+		int curAngle = (int)(angle.Degrees() + 360.0) % 360;
+		frame = (curAngle / 360.f) * preRenderedRotation;
+	} else {
+
+
```
Pointedstick: These two newlines are unnecessary

source/DrawList.cpp
```diff
@@ -146,7 +146,6 @@ bool DrawList::Cull(const Body &body, const Point &position, const Point &blur)
 }
 
 
-
```
Pointedstick: Unrelated whitespace change

source/Body.cpp
```diff
@@ -298,72 +308,81 @@ void Body::UnmarkForRemoval()
...
-	
-	if(!rewind)
-	{
-		// If not repeating, frame should never go higher than the index of the
-		// final frame.
-		if(!repeat)
-			frame = min(frame, lastFrame);
-		else if(frame >= frames)
+
+	if(preRenderedRotation) {
```
Pointedstick: Might be nice to add a comment here explaining what it's used for

Pointedstick:
Ping @julbouln. You still interested in working on this?

julbouln:
Hi @Pointedstick 
Sorry I never answered you request ! Here the changes, but still need a testcase
Regards

tehhowch:
It sounds like this is not yet finished. Even were it finished, I'm not exactly keen on merging in an orthogonal way for the game to draw things that has to be maintained for perpetuity with no way of knowing if it is broken or not.

julbouln:
> It sounds like this is not yet finished. Even were it finished, I'm not exactly keen on merging in an orthogonal way for the game to draw things that has to be maintained for perpetuity with no way of knowing if it is broken or not.

I agree. It serves my proof of concept EVN total conversion, but I'm not sure it is a feature you want and I can't promise any maintenance.

Pointedstick:
In that case we will absolutely need a solid test case. :)

Zitchas:
Idea: One of the most distinctiveness applications of this in EVN was the freighters with solar arrays on them that stayed perpetually pointed a single direction that matched the shading on the ship. If this could somehow keep the shading pointed at the center of the system, and the solar panels, adding a similar sort of ship would make an excellent test case (as well as neat diversity)

beccabunny:
> Idea: One of the most distinctiveness applications of this in EVN was the freighters with solar arrays on them that stayed perpetually pointed a single direction that matched the shading on the ship. If this could somehow keep the shading pointed at the center of the system, and the solar panels, adding a similar sort of ship would make an excellent test case (as well as neat diversity)

That wasn't EVN though

Zitchas:
Ah, you are right. That was Escape Velocity Override's Freighter I was thinking of:
https://evn.fandom.com/wiki/Freighter and https://evn.fandom.com/wiki/Cargo_Transporter

Amazinite:
While I like the idea of adding more mechanics to the ES engine to allow more parity with EV's mechanics, I don't think this should be one of them. The entire game has been built around the top-down perspective, and as tehhowch has mentioned, if we do merge this it would become something that we need to maintain despite not making use of.

I can potentially see adding something like this very far down the line when we've done basically everything else where want to do with the base game, but that time is very far away at the moment, so unfortunately I think we should reject this feature for now.

Issue #4368, opened 2019-06-15
Increase DAILY_DEPRECIATION value to decrease speed of depreciation


Amazinite:
https://github.com/endless-sky/endless-sky/commit/1d0fe945249e518c1420af62cb9ee03e216d6815 increased the time it takes to reach 25% depreciation, but it didn't touch the daily depreciation value. This means that the time it takes to reach 50% depreciation went from 84 days to... 99. The impact of tripling the max age of outfits and ship hulls wasn't actually all that large. Suggested that we increase DAILY_DEPRECIATION from 0.99 to 0.997. This means that instead of taking 99 days to reach 50% depreciation, items take 264 days to reach this point, which is more of a tripling of the original 84 than tripling the max age was.

Depreciation model | Days to reach 90% | Days to reach 75% | Days to reach 50% | Days to reach 30% | Days to reach 25% from 30%
----|---|---|---|---|---
360 max age, .99 daily depreciation (old) | 12 | 32 | 84 | 194 | 166 (46.1% of the time)
1000 max, .99 daily (current) | 13 | 37 | 99 | 242 | 758 (75.8% of the time)
1000 max, .997 daily (the suggestion) | 36 | 100 | 264 | 599 | 401 (40.1% of the time)
1000 max, .998 daily (alternative) | 48 | 132 | 341 | 719 | 181 ( 18.1% of the time)
1000 max, .999 daily (alternative) | 71 | 193 | 468 | 845 | 155 (15.5% of the time)

.997 daily deprecation about triples the times from the old model.
.998 daily makes it take about a year to reach 50% depreciation.
.999 daily about doubles the time from the .997 suggestion, about sextuples (6x) it from the old model.

Zitchas:
That seems like a solid improvement. 

I think one downside to the depreciation is that it punishes players that play the game slowly and explore, and rewards players that are buying new ships in quick succession.  This change will improve that considerably. 

As a side benefit, it will make missions like the gascraft mission where it says the player will be able to sell it back to the shipyard promptly and get almost all their money back much more truthful.

Amazinite:
Captured ship depreciation is always set to FULL_DEPRECIATION, which under this suggestion is still 25%.

Zitchas:
Sounds very  beneficial to me, then.

Janaszar:
What if there was a period of "no questions asked returns"?  Such as, first week you see no depreciation, then it starts being counted?
[...]

Issue #2034, opened 2017-01-19
Extend depreciation time


Amazinite:
Rather unsurprisingly, depreciation hasn't been the most popular among the Steamies. Suggested increasing the time it takes for a brand new ship or outfit to fully depreciate as to make it less punishing when you want to try and experiment with new ships (maybe no depreciation for the first month too), as well as shifting depreciation from being a straight up economy nerf to something that more only effects capturing and plundering.

I feel like this would fix the main issue that some are having that it changes the once experimental/sandbox nature of being able to buy a new ship or outfit, fly it around for a bit to test it, and sell it back for full value if you don't like it.

Lineth:
I support this.

DingusShingleton:
Yeah I think a simple grace period would be the simplest and most effective way. Make it a few days, or even a week, but I think that would nip 95% of those complaints in the bud

KiLEdEnNis:
what if every 10 days items lose a portion of their value? would also fix it, but is more rough

zwparchman:
That feels rather arbitrary. Maybe depreciation should be tied directly to hull damage taken. Something like:

day 1: take 5% hull damage, your ship is now worth 95% of base cost.  
day 4: take 10% hull damage, your ship is now worth 85% of base cost. 

This would allow players that are peaceful to slowly climb the later into better ships without getting punished by depreciation  too hard. while still making combat oriented actions wear the ship out.

Maybe even give shield damage a smaller factor in depreciation so that a combat ship can never come out unscathed.

Amazinite:
Well even when you aren't a combat oriented player, you still occasionally take hull damage. And how would that even make sense to tie depreciation to hull damage when hull damage can be repaired but depreciation can't?

zwparchman:
You have the right idea though. Hull damage will happen to almost every ship so they will all depreciate, but combat ships are likely to take damage more often and thus depreciate faster. As to the hull damage remark, in real life cars lose value from damage, even if it is repaired. 

What does bothers me about the current system is how rough it makes trading. Trading makes far less per day than capturing ships until you have a large trading fleet. I don't have numbers to back it up, but capturing and trading seem to have comparable earnings for the player per real-world minute, but spending minutes in a system for a payday versus running a few trade routes cost vastly different amounts of credits, and I don't think they should.

[...]

PR #4426, opened 2019-07-10
Change deprecation function to a logistic function

gunqqer:
Oh boy, here we go again talking about changing deprecation
Personally I feel as if deprecation is a bit steep. I feel like a logistical function would better fit this. Yes this isn't very comparable to real life, but in my opinion what we have currently is still a bit sharp.
A graphical comparison: alt text
I tried my best to use the right styling but as this was done at 3:30 in the morning there will probably be problems

### DIFF
--- a/source/Depreciation.cpp
+++ b/source/Depreciation.cpp
@@ -27,10 +27,10 @@ namespace {
 	// Names for the two kinds of depreciation records.
 	string NAME[2] = {"fleet depreciation", "stock depreciation"};
 	// Depreciation parameters.
-	double SIGMOID_MIDPOINT = 500
-	double STEEPNESS = .0052
-	double CURVE_MAX = .87
-	double Y_INTERCEPT_CORRECTION = 1.06
+	double SIGMOID_MIDPOINT = 500;
+	double STEEPNESS = .0052;
+	double CURVE_MAX = .87;
+	double Y_INTERCEPT_CORRECTION = 1.06;
 	int MAX_AGE = 1000;
 	int GRACE_PERIOD = 7;
 }
@@ -339,8 +339,8 @@ double Depreciation::Depreciate(int age) const
 	if(age >= MAX_AGE + GRACE_PERIOD)
 		return FULL_DEPRECIATION;
 	
-	double exponent = exp((-STEEPNESS * ((age - GRACE_PERIOD) - SIGMOID_MIDPOINT))
-	return (CURVE_MAX / (1. + exponent)) + Y_INTERCEPT_CORRECTION
+	double exponent = exp((-STEEPNESS * ((age - GRACE_PERIOD) - SIGMOID_MIDPOINT));
+	return -(CURVE_MAX / (1. + exponent)) + Y_INTERCEPT_CORRECTION;
 }
 
 
--- a/source/Depreciation.cpp
+++ b/source/Depreciation.cpp
@@ -27,10 +27,11 @@ namespace {
 	// Names for the two kinds of depreciation records.
 	string NAME[2] = {"fleet depreciation", "stock depreciation"};
 	// Depreciation parameters.
+	double FULL_DEPRECIATION = 0.25;
 	double SIGMOID_MIDPOINT = 500;
-	double STEEPNESS = .0052;
-	double CURVE_MAX = .87;
-	double Y_INTERCEPT_CORRECTION = 1.06;
+	double STEEPNESS = .0053;
+	double CURVE_MAX = .8641;
+	double Y_INTERCEPT_CORRECTION = 1.0571;
 	int MAX_AGE = 1000;
 	int GRACE_PERIOD = 7;
 }
@@ -340,7 +341,13 @@ double Depreciation::Depreciate(int age) const
 		return FULL_DEPRECIATION;
 	
 	double exponent = exp((-STEEPNESS * ((age - GRACE_PERIOD) - SIGMOID_MIDPOINT));
-	return -(CURVE_MAX / (1. + exponent)) + Y_INTERCEPT_CORRECTION;
+	double percent -(CURVE_MAX / (1. + exponent)) + Y_INTERCEPT_CORRECTION;
+	if(percent > 1)
+	{
+		return 1.;
+	}
+	else return percent
+
 }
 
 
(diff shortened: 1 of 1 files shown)


source/Depreciation.cpp
```diff
@@ -337,9 +339,8 @@ double Depreciation::Depreciate(int age) const
 	if(age >= MAX_AGE + GRACE_PERIOD)
 		return FULL_DEPRECIATION;
 	
-	double daily = pow(DAILY_DEPRECIATION, age - GRACE_PERIOD);
-	double linear = static_cast<double>(MAX_AGE - (age - GRACE_PERIOD)) / MAX_AGE;
-	return FULL_DEPRECIATION + (1. - FULL_DEPRECIATION) * daily * linear;
+	double exponent = exp((-STEEPNESS * ((age - GRACE_PERIOD) - SIGMOID_MIDPOINT))
+	return (CURVE_MAX / (1. + exponent)) + Y_INTERCEPT_CORRECTION
```
Amazinite: Pretty sure that this equation doesn't match the one in the graph you showed.
```suggestion
	return -(CURVE_MAX / (1. + exponent)) + Y_INTERCEPT_CORRECTION
```

Amazinite:
Rel: #4372, #4368, #2034

The following table isn't factoring the grace period into the days it takes to reach each percentage.

Depreciation model | Days to reach 90% | Days to reach 75% | Days to reach 50% | Days to reach 30% | Days to reach 25% from 30%
----|---|---|---|---|---
360 max age, .99 daily depreciation (oldest) | 12 | 32 | 84 | 194 | 166 (46.1% of the time)
1000 max, .99 daily (v0.9.9 and prior) | 13 | 37 | 99 | 242 | 758 (75.8% of the time)
1000 max, .997 daily, 7 day grace (current) | 36 | 100 | 264 | 599 | 401 (40.1% of the time)
Logistic, 7 day grace (proposed) | 221 | 394 | 621 | 879 | 121 (12.1% of the time)

I'd dare say that this is *too* forgiving.

I'm pretty sure a better logistic equation could be used, because right now this equation is discontinuous with the grace period and max age. At 7 days, the output is 0.998, not 1 as it should be, and at 1007 days it is 0.25222, not .25.

gunqqer:
ok I updated the numbers, at 0 it wasn't perfectly 1 (1.0001) so I added a check to make sure it can't be above 1. The curve is mostly the same. (I also made it actually compile, sorry about that)

Personally, I'd make this even more slow as I dislike deprecation in general especially all the way to .25 of the initial value (but I'm not going to try and either get rid of it or raise the minimum, I know that will get me nowhere).  
Note: this doesn't change anything for captured ships that start with FULL_DEPRECIATION, only makes it softer for when the player buys a ship or outfit. 

also for those who want to themselves play with this here is the equation from desmos in LaTeX notation: `y=-\frac{.8641}{1+e^{-.0053\left(x-500\right)}}+1.0571\left\{0\le x\le1000\right\}` (Does not include 7 day grace)

gunqqer:
I don't like how you're punished for using something. I get why we have it. I know that we have to have it.  Giving more time for the player to figure out if they want to make the full comitment to that ship without also making it free is the goal of the proposed function. This will bring me satisfaction (I haven't had enough time to do a full playthrough with just this) not only because it will be better for me, but it will give new players more leeway into learning the game. 

If I knew more about C++ I'd really like to add teh functionality of multiple different curves and variables for those curve so each weapon or outfit could (keyword: could) be different. (Possibly having different difficulty settings, but that's for a different issue)

Pointedstick:
> Giving more time for the player to figure out if they want to make the full comitment to that ship without also making it free is the goal of the proposed function.

The grace period already does this, right? After a certain point, you can't return a product for your full money back anymore. Whether or not that period is long enough may be debatable, but I think having it at all significantly reduces the "get punished for experimenting with new builds" factor. Even so, the punishment isn't that much since you can still sell stuff back at 80-90% after even a few months.

I get where you're coming from though. For me it was the crew salaries mechanic that I *really* hated when I started playing Endless Sky. Losing money with every hyperjump forever made me feel super stressed out and I fought against it for a long time, making requests for crew-less ships, interest-bearing bank accounts that could be used to generate daily income for offsetting the cost, and even creating a ship (the Vanguard) with an intentionally low minimum crew complement for its size and class. Eventually I gave up and learned to live with it, eventually coming to appreciate it as both a balancing mechanism and also as an interesting gameplay mechanic.

gunqqer:
"learn to live with it"
Nice there

The grace period exists... for 7 days. After 12 more days it's already lost 10% of its value. This massive loss so quickly serves no good purpose. You can't learn much about a shit that fast. Hell, by the time you've gone to a pirate system you've probably already past a month.

This still has the mechanic. I am not advocating removing it entirely like you wanted. After 1000 days your ship will still be worth 25% of the original. It still has the same mechnics, just at different rate.

Amazinite:
>The grace period exists... for 7 days. After 12 more days it's already lost 10% of its value.

You're looking at the old numbers. It's a 7 day grace period, then 36 more days to lose 10% of its value right now.

Tadrix:
> Whether or not that period is long enough may be debatable, but I think having it at all significantly reduces the "get punished for experimenting with new builds" factor.

How about we just give them a grace period of 15 days? It's enough to finish the tutorial (so the player doesn't feel the urge to adhere to James' missions any more and might decide to get a different setup, the closest shipyard isn't far), and a fortnight is a reasonable return time for goods received by mail IRL.

Pointedstick:
Increasing the grace period by a bit doesn't seem totally unreasonable to me. If we make any changes here, I think that would be more appropriate than further lengthening the period of time required for things to depreciate, as it would directly aid the pointed-out use cases (people trying new things, new players).

gunqqer:
While increasing grace period will technically help, I still feel after it the function goes too fast. The biggest problem this is here to fix is that the things you own lose value too fast (in the short term, at the beginning) which then branches off to help fix other things like aiding new players. 

(I wouldn't be opposed to changing grace period in addition to the proposed change)

jafdy:
A fortnight is a fairly typical time period for cancellations of things like bills and insurance, not sure if it applies to houses . . .
Anyway, we should probably add a line somewhere that indicates to the player that the grace period exists, perhaps in the bit of dialog just as you buy your first ship.

Amazinite:
After playing around for a while with the grace period and slowed depreciation, I don't think that the depreciation function needs to be any more forgiving. Depreciation was suppose to serve two purposes: 1) make capturing ships less lucrative, and 2) intentionally slow down the early game. 
While changes to the way depreciation occurs don't impact the first purpose, this does effectively nullify the second purpose, as your assets will barely have depreciated by the time you're ready to upgrade to the next ship, making the mechanic effectively pointless at slowing down the early game.

Issue #4334, opened 2019-06-04
Request: Hyperjump arrival/departure distance variables


Zitchas:
**TLDR summary: Just to be clear: This request specifies that the default values should be ZERO, which would mean that they have no impact on the current gameplay. Then these variables would be available for use by content creators (particularly plug-in makers) who *do* want this behaviour.**

**Is your feature request related to a problem? Please describe.**
1) During almost all combat missions I arrive much too close to the system center, and thus am already in close-range combat immediately upon exiting hyperjumps. 
2) Some people would like a more "Classic Escape Velocity (EV/EVO/EVN) feel to the game with mandatory minimum distance from the system center before one can hyperjump (and arrive from hyperspace)

**Related Issue Links**
Some conversation related to this:
https://github.com/endless-sky/endless-sky/issues/1094
https://github.com/endless-sky/endless-sky/issues/1635
https://github.com/endless-sky/endless-sky/issues/3518
https://github.com/endless-sky/endless-sky/issues/3661

A PR that made some steps towards it, but was never merged:
https://github.com/endless-sky/endless-sky/pull/3628

**Describe the solution you'd like**
A) Implementation of `hyperjump arrival radius` and a `hyperjump departure radius` variable, ideally as attributes of systems. These values would default to 0, and in their default configuration would have no impact on current gameplay mechanics of Endless Sky.
B) Implementation of `hyperjump arrival distance` and `hyperjump departure distance` attributes for ships and outfits. This is intended to allow certain ships to have different arrival or departure radii, and to allow outfits to change these values. 

All of these variables would use the normal system measurements to describe distances.

**Impact and Usage**
In vanilla Endless Sky, one immediate usage that I can see would be using these variables to create an outfit (say, the "Long Range Tactical Computer") that adds an amount to the hyperspace arrival radius. This would be immediately useful for ships like the Albatross (Sniper) variants that are built for longer range engagements. Variations on this might be very useful for various other configurations of NPC ships that prioritize longer-range engagements.  That being said, it is important to note that this would *not* be another hyperdrive, but rather it would be a small outfit that modifies the behavior of whatever drive the ship has installed, and would apply equally to hyperdrives and jump drives.

This would expand the tactical flexibility of the game, and provide opportunities for smart players to arrive into battles with a bit more time to adjust. This could significantly decrease the "jump and pray" aka save-scumming methods of dealing with certain large battles in the storyline.

[...]

PR #4432, opened 2019-07-15
feat: Hyperjump arrival distance variable

petervdmeer:
Allow entering systems at additional distance from the target.

This implements the (arrival only) feature as requested in #4334 and detailed in the RFC [here](https://github.com/EndlessSkyCommunity/endless-sky/pull/51).


Amazinite:
And how about a better keyword like `"arrival distance"`? Then the other half of #4334 could be the `"departure distance"`. (And actually looking at #4334 again, those are exactly the terms that the issue uses for this mechanic.) If I saw `link-shortfall` I'd have no idea what it meant.

Tadrix:
How about just `arrival` and `departure`? Aren't two word keywords worse than one?

Amazinite:
That works too, as we don't have `"habitable zone"` or `"asteroid belt"`, we just have `habitable` and `belt`.

petervdmeer:
I changed they keyword to `arrival` as suggested.

Having a global keyword as suggested by Pointedstick would indeed be a nice service for plugin authors and for projects like EV conversion. I like to keep that for a separate PR.

The reason for me to implement this was this remark in #4334 :

> I was looking at ways of making combat more engaging for the player, and one of the issues is all the ships just thrown on top of each other.

With this change it is already possible for any player to experiment if the arrival distance will make combat more attractive.

For people that want to use this: @accursedCursive made a nice remark in #3628 that 1000 seems to be the sweet spot between more engaging combat and not being a nuisance in systems that have little combat and I think he has a good point there.

I also tried 10000 when I tested this feature, but that really gives one the feeling of traveling all alone in empty space.

petervdmeer:
@Pointedstick Is there also a way to change "Changes requested" to something like "Changes done, awaiting re-review?"

Tadrix:
Try 1250. That distance is notable as a "nominal point" for fuel collection values.

jafdy:
Later, this isn't a bug fix.

Amazinite:
Slapping that milestone on it and calling it a day.
If it works though, it's a +1 from me. All for giving plugin makers more tools.

source/System.cpp
```diff
@@ -268,6 +268,8 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 			trade[value].SetBase(child.Value(valueIndex + 1));
 		else if(key == "object")
 			LoadObject(child, planets);
+		else if(key == "arrival" && child.Size() >= 2)
+			arrivalDistance = child.Value(1);
```
Amazinite: It looks like this should be `arrivalDistance = child.Value(valueIndex);` so that events can properly add to and remove from the arrival distance value.

Zitchas:
Once this gets merged, my first order of business will be making a plugin that alters every system in the game so that all ships warp in a suitable distance out. Especially all those systems that have regular battles (and even more so, every system that has big mission battles).

Thanks to everyone working on this!

source/System.cpp
```diff
@@ -268,6 +268,8 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 			trade[value].SetBase(child.Value(valueIndex + 1));
 		else if(key == "object")
 			LoadObject(child, planets);
+		else if(key == "arrival" && child.Size() >= 2)
```
tehhowch: we're guaranteed that the key has a value by line 184
(note that the "trade" key above accesses additional indices, which is why it needs the extra `child.Size()` check)

Zitchas:
Actually, I like the idea that this change (if implemented widely) would actually make the JD more different from the current HD. Gives the two more tactical differences.

source/System.cpp
```diff
@@ -268,6 +268,8 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 			trade[value].SetBase(child.Value(valueIndex + 1));
 		else if(key == "object")
 			LoadObject(child, planets);
+		else if(key == "arrival")
+			arrivalDistance = child.Value(valueIndex);
```
tehhowch: Should we allow negative values of `arrivalDistance`?

source/Ship.cpp
```diff
@@ -1215,6 +1215,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 			// Have all ships exit hyperspace at the same distance so that
 			// your escorts always stay with you.
 			double distance = (HYPER_C * HYPER_C) * .5 * HYPER_A + HYPER_D;
+			distance += currentSystem->ArrivalDistance();
```
tehhowch: So one thing to mention is that this is usually not the correct usage of the **`System`** arrival distance, because the `target` position is generally that of a planetary body, and not the system center. This would, however, be (part of) the correct usage of a **`Planet`**'s "HyperspaceExclusionRadius". Only when the player jumps into a system that has no planets with spaceports and has also not explicitly selected the planet in that system as a landing target would `target` be that of the system center (aka `Point { 0, 0 }`).

I guess the question is, "how correct do we want to be?" (Or, "how much vector math would you like to do?")

source/System.cpp
Zitchas: What is the impact of negative values? Or I suppose the corollary question would be: Where does `arrivalDistance = 0` leave the player? Is it in the system center? or is is it the normal arrival spot around roughly the habitable zone?

Some speculations that come to mind:
If it is the former, and 0 is the center of the system, would a negative value catapult the player past the center of the system and drop them on the far side? If so, I'd say allow this value, because it make a neat change for certain specific systems or missions (If anyone were doing, say, a mission revolving around investigation of weird hyperspace phenomenon or something.)

What I'm getting at is this: So long as negative values have a noticeable and relatively consistent/predictable impact on ship arrival location, I would say we should allow them.

source/Ship.cpp
Zitchas: Given the choice, I would prefer the hypserspace exclusion radius to be from system center rather from the planet. How complicated is the vector math involved?

source/System.cpp
petervdmeer: `arrivalDistance = 0` is the default value, which leaves the player arriving at exactly the same spot as where he/she arrives without this PR.

A negative `arivalDistance` value would indeed catapult the player beyond his/her intended destination. I considered blocking negative values, but I actually came to the same conclusion as you; there are probably some content/plugin creators that can do something nice with negative values.

source/Ship.cpp
petervdmeer: A simple solution would be to always use the system center as reference point when `arrivalDistance` is not zero. That should be a one-liner of code.

I didn't check yet how complicated the complex solution would be.

source/System.cpp
Zitchas: Actually, if this gets merged, I'm going to include this in Remnant missions. :) A very long ways down the story tree, but still, fun ideas. : )

source/Ship.cpp
petervdmeer: I implemented the simple solution with my last commit.

petervdmeer:
Based on a good remark from @tehhowch on Discord I made a suggestion for multiple modifier keywords in #4334 .
The change here only affects hyperdrives, but if we accept the suggestion I made, then Zitchas can choose between 2 versions of the keywords in his plugin:
`arrival <distance>` for distances that should apply for both jump and hyperdrives in the future
and 
`arrival <distance> link` for distances that should only apply to hyperdrives in the future

Zitchas:
Ah, so that's the difference between the two. Thanks for clarifying.

geojak:
Help, sorry to bother you guys again. I would really love to use this feature in my plugins. Any chance of a merge for the next release?

Amazinite:
> Any chance of a merge for the next release?

I'd like to have this merged for the next release (the associated issue is marked for v0.9.13). As for when it gets merged, I've been focusing on reviewing content PRs, so after that is when I'll be reviewing any mechanics/enhancements PRs. Tehhowch may get to this before I do, though.

source/System.h
```diff
@@ -95,6 +95,8 @@ class System {
 	
 	// Get a list of systems you can travel to through hyperspace from here.
 	const std::set<const System *> &Links() const;
+	// Additional travel distance to target for ships entering through hyperspace.
+	double ArrivalDistance() const;
```
Pointedstick, line 99: To me, the function name does not match the description. I would expect that something called `ArrivalDistance` would be expressing an absolute distance, not an offset. A name like `AdditionalArrivalDistance` or `ArrivalOffset` might be clearer IMO.

petervdmeer: Changed to `ExtraArrivalDistance`.

Amazinite:
Bumping so that I can link.

petervdmeer:
Merged master to resolve minor merge conflict with jump distance PR.

Amazinite (commented):
This doesn't seem to work with jump drives, which I'd expect to be the case with such a mechanic. (Having this work for jump drive would also be highly beneficial for #5560 to avoid jumping into the system right on top of the hazard.) From the RFC there is a mention of having keywords “link” or “jump” after the distance to specify only one or the other travel method being affected, so I’d expect a lack of either keyword to mean it affects both. 

Also, what would it take for the added distance to be relative to the target planet instead of forcing the distance to be from the system center? It seems odd to me that should this mechanic be used, some systems I would exit my jump near a planet and in other systems I would always seem to exit some set distance from the system center instead of near a planet.

Otherwise, seems to be working as intended right now with hyperdrive travel.

petervdmeer:
This PR was just a small change to allow content creators to experiment with the more cinematic combat as mentioned in #4334.
I agree that we should also get this feature for jump drives and maybe I even create a follow-up PR for that. It just wasn't added into this PR.

The RFC mentions how to specify arrival distances from system-center and how to specify arrival distances from planet-centers (and also how to combine them). The basic distance comparisons are relatively easy (you just calculate all arrival distances from the center and from all planets that set distances and take the maximum one).
What could add some complexity is that a ship usually doesn't cross all objects that set arrival distance restrictions. If we want to go for a very precise calculation then we need to take into account the trajectory that a ship is on when entering the system relative to the objects that set arrival distance restrictions.

Amazinite:
Well what I’m looking for right now is at the very least a mechanic to change the distance of arrival for both hyperdrive and jump drive travel. From where that distance is measured doesn’t really matter right now as far as I’m concerned. So how difficult would it be to have the arrival distance from this PR also apply to jump drives? Then in the future we could deal with customizing it further (e.g. hyperdrive only, jump drive only, distance relative to planets, distance relative to center, whatever).

petervdmeer:
> <...> how difficult would it be to have the arrival distance from this PR also apply to jump drives?

The code change is very easy. But it is as easy to add here as it is to add in a separate PR.

I understand your need as content-creator, but this PR was created 1.5 years ago for a different need (more room for long range weapons in large space-battles) which I noticed Zitchas, geojak, ChamEV, Pointedstick, SirStigll and me are interested in.

If this PR is having bugs or issues, then I consider it very reasonable of you to ask me to fix those. But if this PR is working fine, and your only concern is that it is missing an additional feature that you really like to have, then I feel you should really review this PR for merging as-is (and hint during such review that you would really like to see another PR with the additional feature).

Issue #4420, opened 2019-07-06
Diversify space background in order to make it more interesting


Tadrix:
For starters, I'm apologising in advance for the text wall. I've yet to learn to express myself more briefly in English.

**Is your feature request related to a problem? Please describe.**
Honestly, flying for many years through the galaxy, it gets visually boring. The systems visually differ only with their stars, and the asteroid density. The eye learns to recognise tiling patterns on the fly, and it's a bad thing. Put bluntly, the space background is dull and boring. It might be somewhat realistic, as far as realism in this game goes, but after seeing [photographs](https://duckduckgo.com/?q=night+sky&iax=images&ia=images) (yes, I know they had a long exposition time), one would think the game can do better. We're playing for fun, after all, not for a realistic simulation of routine, repetitive tedium.

Even the game advertises itself a bit differently. Here's the endless-sky's avatar:

![Avatar 11450262](https://avatars2.githubusercontent.com/u/11450262)

Looking at it for a long time would probably be tiring (I've always thought that in SPAZ it was overdone, and seeing a black sky was a relief from being immersed in oversaturated brightness; in Homeworld, however, it might've been done just right), but it's a nice alternative to the usual blue haze. The interface features something akin to the red haze too:

![Hail Panel](https://raw.githubusercontent.com/endless-sky/endless-sky-high-dpi/master/images/ui/hail%20panel%402x.png)

Seeing a system with a [nova](https://github.com/endless-sky/endless-sky/blob/master/images/star/nova%2B.jpg), a [Wolf-Rayet star](https://github.com/endless-sky/endless-sky/blob/master/images/star/wr.png), or a [black hole](https://github.com/endless-sky/endless-sky/blob/master/images/star/black-hole%2B.jpg) in it is great. *Whoa!* The background is gorgeous there. It's a shame you as a player don't get to stay around for long, because such systems understandably are completely devoid of any business to be had, so people don't tend to linger there.

Technically speaking, every star has its own spectral characteristics (one of the reasons they are shown coloured differently). Everything would look a bit differently depending on what star is present. Instead, everything looks the same (I'm not sure that can be corrected, and if any filters are a good idea, but it certainly compounds the problem).

**Related Issue Links**
#2488, although I don't think having colours based on what faction inhabits what part of galaxy would be a good idea. It should depend on the part of the galaxy, and the local star.

**Possible solutions**
In short, making the galaxy a little more visually interesting place. I'm not a designer, so I don't have ready solutions.

[...]

Issue #1057, opened 2016-05-09
More interesting backrounds/backdrops.


mdsmestad:
I like black starscapes, but I'd love to see more diverse backrounds. Add some purple and blue haze textures to some systems, moving comets, etc. Would add a little extra something.

endless-sky:
I think it's really hard to do that without it looking tacky. I also like the fact that right now, systems with "backgrounds" (i.e. [huge star graphics](https://github.com/endless-sky/endless-sky/blob/master/images/star/nova%2B.jpg)) are rare enough that when you find something, you feel like you've discovered something special.

Note that in terms of real physics, even if you were in the center of a nebula it wouldn't appear to you to be anything more than a barely visible haze in the distance; the only way telescopes get those bright images is with very long exposure times (and often, false coloration, as well).

mdsmestad:
@endless-sky You have a point there..What about backround comets or other moving elements (other than asteroids) that are non-interactable (purely backdrop like stars)

jafdy:
Again it is prone to become tacky. Besides anything that you can see moving must be going very fast or be close to you or more likely both.
Could you make something to show what you would like?

LocalGod79:
comets could be added as planetary sprites...

jafdy:
Dose the game engine support elliptical orbits?

LocalGod79:
while i'm pretty sure they're all circular, i'm also pretty sure that the sprite would always be oriented correctly, and that an offset (if still round) orbit is possible.

mdsmestad:
@LocalGod79 I'm envisioning a bright, smallish glowing blue orb with a transulent blue tail slow moving in orbit around a system in the backround.

Gkramme:
A comet's tail always points away from a star, so the comet would need to be programmed to keep it's tail side away from the star

LocalGod79:
@mdsmestad that would be incongruous with the rest of the planetary movement within systems.  i'm not averse to the tails of the comets being animated, but their movement should not be drastically different than the rest of the planets...

@Gkramme i knew this - since planetary objects are already programmed to orient their sprites in relation to the star, having the comet's tails point in the correct direction should just be a matter of making sure that the sprites are oriented according to the standard...

jafdy:
The orientation could also be along the normal of there orbit which would look odd with offset orbits.

PR #4434, opened 2019-07-16
Increase background haze visibility

Pointedstick:
This PR slightly increases the visibility of the background haze effects, making it more prominent and prettier, especially when zoomed out. A picture's worth a thousand words, so I'll let some before-and-after screenshots do the talking:

Before:
Blue - zoomed in, before
Blue - zoomed out, before
Red - zoomed in, before
Red - zoomed out, before

After:
Blue - zoomed in, after
Blue - zoomed out, after
Red - zoomed in, after
Red - zoomed out, after

These haze files are only here in the main repo, not also in the high DPI repo, so no PR for it is necessary.


ChamEV:
Looks good,  The only issue with haze that comes up in games is often it can be too bright and make the foreground hard to see.  The red is close, but I see no issue here.

Pointedstick:
Yeah, I tried not to go overboard with it.

Tadrix:
That's a nice addition, but I request that you don't close #4420 yet (I'd have to rename it), as it touched not only on haze sprites, but also system sprites (and lighting), like having dust (or asteroid) rings around systems using the same principle as nova, black hole, or WR sprites which would change the background near the star.

Pointedstick:
Sure, no problem!

Tadrix:
Thanks.

Bladewood:
My impression was that background haze was meant to be a [very subtle effect](https://github.com/endless-sky/endless-sky/commit/ff5964b429bb95eac81d46117c9ffae7a486f53f). See also [this post](https://github.com/endless-sky/endless-sky/issues/1057#issuecomment-217769415).

These new hazes look very shrill in places where they used to smoothly fade into black.

Pointedstick:
> My impression was that background haze was meant to be a [very subtle effect](https://github.com/endless-sky/endless-sky/commit/ff5964b429bb95eac81d46117c9ffae7a486f53f). See also [this post](https://github.com/endless-sky/endless-sky/issues/1057#issuecomment-217769415).

Let's discuss whether or not we even want to do this in https://github.com/endless-sky/endless-sky/issues/4420

Pointedstick:
So, anyone up for this?

beccabunny:
I'm still very interested in this feature request.

Zitchas:
Would it be possible to add a opacity value to the haze? The haze itself is specified in the map.txt file. I would think it easy enough to have it look for a value to go with it that represents how bright and obtrusive it is. This would allow us to have a variety of haze without increasing the number of image files we need. 

Along the same lines, being able to stack the two haze on top of each other at varying levels would open up a lot of variety possibilities too.

Pointedstick:
Those are good enhancements for a separate patch IMO. :)

Amazinite:
@Zitchas Might be worth opening a PR on a haze opacity value.

MCOfficer:
@Pointedstick Are there any assets for these? Ref #4571

Pointedstick:
> @Pointedstick Are there any assets for these? Ref #4571

No, the assets are all here in the repo already. I just modified the files in GIMP.

MCOfficer:
Since the originals didn't have xcf's either, i suppose that's ok.

PR #4387, opened 2019-06-21
Change Bactrian Classification

Zitchas:
This PR is purely cosmetic, and is an effort to redirect the perception and fixation on the Bactrian by making one simple change:

- The Bactrian is reclassified as a Transport.

As @jafdy pointed out in the discussion for PR #4347 , a large part of the problem isn't the stats of the Bactrian, but rather the perception. As such, this PR seeks to rectify that perception in the most basic way: Change the label. After all, the label is the single most easily identifiable way to decide what something is. According to the discussion, the majority involved in the development of the game seems pretty clear that the Bactrian isn't a warship, it is simply a transport that is decent warship. And likewise, everyone seems pretty much in agreement that the Bactrian isn't a great warship when compared with the alternatives. As such, I propose that we pass on this knowledge of the development team to the player base by changing the label on the Bactrian so that first time players can see, right off the bat, that this is NOT a warship, but rather a combat capable transport. In other words, letting new players see what the veterans already know.

PR #4435, opened 2019-07-17
Reclass the Bactrian as a Heavy Freighter

Tadrix:
in line with how endless-sky classed the Korath WS (the Baccy T2 basically). Ships that perform well as both freighters and transports usually are classed as freighters in this game, and having the Bactrian as a transport would be too misleading for mission conditions (as a rule, frighters can be decently armed, but transports usually aren't) that would underestimate player fleet's capabilities.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -188,7 +188,7 @@ ship "Bactrian"
 	attributes
 		licenses
 			City-Ship
-		category "Transport"
+		category "Heavy Freighter"
 		"cost" 17600000
 		"shields" 17500
 		"hull" 8600


MCOfficer:
Relevant discussion at #4387.

Pointedstick:
I thought folks wanted to just revert the change?

Zitchas:
I think it'd be better to reclassify the World Ship as a Transport than the Bactrian as a Heavy Freighter.

Very few people look at the World Ship and go "There goes a freighter!" but rather "There goes the biggest troop transport in the galaxy!"

Tadrix:
> I thought folks wanted to just revert the change?

Either reverting or this is fine by me. I just don't think it should be a Transport, and I've explained why before. In either case I don't expect that the discussion will die completely, because it seems there are different points of view.

Pointedstick:
Let's do it in an issue. That's less confusing.

PR #4441, opened 2019-07-22
Write "cargo" keyword at the same level as it is read

petervdmeer:
The "cargo" keyword was written by CargoHold, but the keyword was read by PlayerInfo and by Ship.

This change moves the writing of the keyword one level up to PlayerInfo and Ship to allow reuse of the CargoHold for other things with other keywords (like planetary storage).

### DIFF
--- a/source/CargoHold.cpp
+++ b/source/CargoHold.cpp
@@ -97,11 +97,9 @@ void CargoHold::Save(DataWriter &out) const
 	for(const auto &it : commodities)
 		if(it.second)
 		{
-			// Only write a "cargo" block if it is not going to be empty.
+			// Only write a "commodities" block if it is not going to be empty.
 			if(first)
 			{
-				out.Write("cargo");
-				out.BeginChild();
 				out.Write("commodities");
 				out.BeginChild();
 			}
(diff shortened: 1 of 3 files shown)


petervdmeer:
Cancelling this pull request, this change is included in #4443.

tehhowch:
Why not just pass the desired keyword into the `Cargohold::Save` method? Then the burden of serialization is still the duty of Cargohold, and anything that has cargo only needs to know what keyword it uses for cargo (which is already a necessity in the respective `<class>::Load` method).

petervdmeer:
> Why not just pass the desired keyword into the `Cargohold::Save` method? Then the burden of serialization is still the duty of Cargohold, and anything that has cargo only needs to know what keyword it uses for cargo (which is already a necessity in the respective `<class>::Load` method).

Yes, that was my first idea too, but PR #4443 uses the following construction:
```
storage   <-- keyword for starting a list of cargo stores (one per planet)
	planet Mainsail <-- start of the cargo store for the planet
		outfits <-- contents of the cargo store for the planet
```
In this way, there is no single new keyword, the store starts directly after the planet identification. I could add another layer of indirection by re-using the keyword storage also below the planets identification, but this seemed to look a bit more logical in the save-game (the additional storage keyword looked a bit redundant).

Btw, we might want to continue this discussion in PR #4443, since this PR is included there. (I expected the implementation of #4443 to take longer, so that why I initially opened this PR as a standalone change).

petervdmeer:
@tehhowch I notice that we never finished this discussion (here or in #4443).

I see 2 options for #4443 (the planetary cargo storage):
- Add the cargo-keyword (or another keyword) as intermediate between the planet and cargo-contents and then just serialize as-is.
- Get this PR merged to allow cargo-storage that doesn't start with a single keyword.

Which option would you prefer? (Or do you even see a 3rd option?)

PR #4442, opened 2019-07-23
Completing a collection of Marauder ships shouldn't carry a death sentence.

LocalGod79:
Tier 1-2 Authors become hostile when the Mad Quest becomes disabled (as intended when MMK was created), as well as when higher tier " Authors " are destroyed.

Newer, higher tier author ships no longer bring their wrath when low tier authors are destroyed/disabled.

Other considered iterations included a mission that spawns and restricts higher tier author ships to isolated systems as "Boss Battles", or missions that spawn higher tier author ships only when the player's reputation with "Author" drops well below the consequences of capturing intended ships.

Open for discussion.


SirStigII:
Mad quest?

LocalGod79:
> Mad quest?
The Mad Quest is the Marauder Fury piloted by Marauding Max K (MMK for short).  It is:
* The only Marauder ship with only one variant.
* The only Marauder ship assigned to a person.
* The first contributor person ship added (after the initially created M.Z. and Cap'n Pester.)
* The (currently) only person ship the player can capture (to offer a challenge-game: collect all the Marauders).

The ship's name is derived from its description "...This ship appears to be one man's insane quest to make the most powerful single pilot warship ever..."

tehhowch:
Did you mean to prevent PotP from engaging pirates? If not, you will need to update the Pirate attitude toward the new government.

It was my intention that destroying the other Author ships would result in it's hostility, so in a later release we can work on exporting the destroyed persons to a place where missions could use them as conditions. An "is destroyed" check plus a rep check will account for author death via pirates (sorry, Nate).

The simpler solution is to move MMK to a different government, but that wouldn't help any existing savegames in which MMK is already captured by the player.

Tadrix:
> The simpler solution is to move MMK to a different government

I'd recommend doing exactly that. People playing existing saves probably either reloaded, or coped with it in some other way (by destroying other Authors, for example). Our main concern probably should be new pilots, as we're not breaking anything for the old ones, we're just not updating them half-way into the game.

LocalGod79:
It was not my intention to prevent PotP from engaging pirates.

As mentioned in the O.P., it was/is an option to have capturable author ships in their own government, and we should discuss them.  The immediate problems with this approach are:
1) MMK has no backup right now.
2) It perpetuates a farming style - piss off one of this type of ship, and they'll all come to get disabled in due time.

Having the original T1 author ship come after you for capturing the Mad Quest was intended, and I'm still okay with newer author ships expressing that displeasure - PotP is *very* strong - way stronger than all other author ships *combined*, and is a frequent death sentence for players until they've got billions of credits worth of fleet around them.

Perhaps we reduce the frequency that PotP appears, so that a player could still potentially just reload upon getting killed, just not quite so often.

another option is for a mission to be written to make high-end " Author " ships angry once sufficient damage has been done to the "Author" government (i.e. high value/crew warships have been killed)?

Tadrix:
Logically speaking, there is no reason for the other authors to bother coming after some marauder gets disabled. My original intention was to just put MMK into the Marauder government, but then I realised he'd be hostile right from the start, and that probably isn't what we want.

If many authors get destroyed, then it would make sense for them to gang up on the player. My recommended approach would be that we don't differentiate between high-end or low-end Authors, counting destroyed ships instead. Once the player defeats [capturableShips]+1, they get mad. This would ensure that, while the player can collect, senseless murder still gets punished. They could even band together and come after him as a fleet (except for Captain Pester and ships of comparable strength or stronger, who would still fly separately).

LocalGod79:
Cap'n Pester is of the Parrot government, and has no qualms with players that make Author hostile.

Collecting all the Marauder ships is a mini game accomplished by taking Marauder bounty missions, capping as many as possible, and selling duplicates - it's a competitive T1 credit farm, and a little side-quest, if one feels like.

The Mad Quest flown by MMK is the only Marauder Fury, and doesn't appear in any of the bounty hunting missions, so to complete the collection, you used to have to make M.Z. hostile.  Just having MMK spawn, and making him hostile is a challenge, let alone catching and disabling - but possible, having only farmed Marauders.

because the game was/is in a state of infancy, we hoped to show that content like this could span different features within the E.S. engine - jobs, invisible missions, and persons all tied together and scaled through progression.  The Remnants additions seem to be taking a similar direction.

In the scenario above, if you'd already farmed all the other Marauder ships before capping the Mad Quest, acquiring sufficient protection for M.Z.'s visits was affordable and not tremendously time consuming.  This continues to hold true for all other T1 and 2 author ships.

Without building a fleet of T2+ ships (which the player may not have even discovered while completing their Marauder collection), chances are that PotP will wipe them and their fleet every time he spawns.

I'm not super happy with this implementation, either - tbh, I'd prefer to leave everything the way it was, and nerf PotP's hull regeneration rate so that it's not quite a certainty for player death, but the figure is so high that I just guessed that Tehhowch intended to be indestructible in the current implementation, and who am I to argue with that?

PR #4455, opened 2019-08-05
Less blurry Southbound ship sprites

beccabunny:
This PR should fix all the blurriness issues that appeared in https://github.com/endless-sky/endless-sky/pull/4319, the only ones untouched are the top view haulers and haulers f as they were alredy as large as the 2x versions. The high dpi PR will arrive soon


beccabunny:
> The Bastion and Clipper top-downs look to me like they're blurrier with this change. Otherwise good.
I will remove them as soon as I get home, at the moment i'm on holiday.
> The Dreadnought thumbnail has shrunk. Intentional?
Yeah, #4319 increased its size a bit since it was way too blurry before, but now it isn't needed anymore so I reverted its size (I hope) to the original one.

PR #4458, opened 2019-08-07
New Betelgeuse ships

beccabunny:
This PR, the second of the detailments to human ships, is my attemp to redesign Betelgeuse Shipyards ships. It also changes the related Marauder ships and the copyright. There is only one Betelgeuse ship that hasn't been redesigned here, the Berserker, for teh following reasons:
-It is two of the 3 game icons (steam firebird, top down berserker and in-flight berserker) and Michael didn't share the 3d and xcf file of the in-flight icon, so it would have been difficult to recreate it; 
-The Berserker detailment that that icon has is alredy amazing, and if it was rendered as the top view and as the thumb it would be a perfect redesign as it is.
Here there are some comparations between old (left) and new (right) Betelgeuse ships:
Behemoth
comparation
Marauder Leviathan
comparation
Star Queen
comparation
screenshot on Chiron
Screenshot (85)
and the new Marauder Leviathan (Engines) and Behemoth, which had their number of engine ports (and the relative zoom factor because of the root2 rule)
Screenshot (84)
The new .blend and xcf files are here
https://drive.google.com/open?id=1V5xyoWgnmUq8cF_0JTPLZyafNimdaB1Y


gunqqer:
Hey these look really nice, I love what you did to the Behemoth's cockpit

finite-galaxy:
Separating the two turret mounts of the Firebird and putting them at a distance from each other is a great improvement! The better visibility of turret mounts on all ships is also nice. And bonus points for the extra engines on the Marauder Leviathan (Engines)!

On the other hand, I don't like everything you did to the Star Queen thumbnail (third image in your opening post):
* In your new one (right) the tail fins seem to slice through the engines, whereas in the old one (left) they are placed below, providing support.
* Those external cables extending to the engines look like a vulnerability and are not really aesthetically pleasing for a luxury liner.
* You've narrowed and flattened the “cockpit”; I think the more circular panorama dome of the old version would be more pleasing for the passengers.

Furthermore, I have a few suggestions for the Behemoth sprite:
* I see you've added two new guns (red) but the old ones are still visible (blue); perhaps you could remove both and add the guns to the frontmost compartiment (yellow)?
* Could you fill up those odd gaps (cyan)?
* The Behemoth has limited engine capacity, so I'm not sure having four engines is appropiate. I'd recommend removing the old and new engines (magenta)—but keep the tail fins—and converting the outermost compartments (green) into engines.
behemoth_requests

jafdy:
I don't think those things you circled in blue are guns, if you look at the side by side you will notice that they are not original anyway.
Moving the engines to the outer edge would be visually very distinct but is a major departure from the existing placement of engines in this game and most other si-fi entries where engines are either right at the back or on the inner wing aeroplane style.

finite-galaxy:
> I don't think those things you circled in blue are guns, if you look at the side by side you will notice that they are not original anyway.

Then what are they? Here cropped from current file (enlarged but otherwise unchanged):
blue

> Moving the engines to the outer edge would be visually very distinct but is a major departure from the existing placement of engines in this game and most other si-fi entries where engines are either right at the back or on the inner wing aeroplane style.

It would make the Behemoth more unique, yes, which is why I suggested it. It's not entirely unprecedented, see e.g. the Manta's engines.

beccabunny:
@finite-galaxy thank you for the suggestions, but i have a few concerns on them:
-holes: they aren't actual holes in the hull, they are separations between connectors from cargo pod to cargo pod; they would have been an issue in the old outline shader, but with the new one they shouldn't be.
-guns: it was considered to put them up in the cockpit, but they were worse than the current setup due to being closer in a ship so large. As for the parts where the guns where, they are just details, as you can see in the thumbnail
-engines: as with guns, it was alredy tried but the result was very bad because of the pods being almost in the middle of the ship; i agree that the manta is the closest example in game, but still it is a very different case since it has only a little of the tail under the engines, while behemoth would have 1/3 of the ship behind them.

kikotheexile:
We appreciate your feedback but I whole-heartedly disagree with you on the front of both the behemoth and the star queen.

On the behe:
We did actually consider making the outer pods engines in the modified variant of this ship, but it just wasn't feasible and was too large of a departure from MZ's design/vision. 
As for those front gun placements - it's just detailing in order to add interest to the eye. 
Putting the guns so close together on the nose of the ship is not in the style of Betelgeuse at all. That would be more like syndicate or republic. Southbound kinda went that direction with the dreadnaught, but that was just to compete with the republic. This freighter isn't designed to compete with military grade ships... its a freighter. 
Those holes aren't actually holes.

On the front of the star queen:
yea the engine tubes are absolutely required to make this ship look at all decent without going too far into the "aircraft IN SPACE" look of like, jet turbines or whatever. Adding additional armor there, for instance, would have ruined its cruise liner profile. The tubes are fine. 
the front dome is the same front dome just without all the ugly clipping. 
The fin placement makes more logical sense for a re-entry capable vehicle, especially one that likes putting on light-shows for its passengers. the burn up would be easy to see from the observation lounge added along the spine, and in general would just be hella neat.

finite-galaxy:
Thank you for your quick replies!
> -guns: it was considered to put them up in the cockpit, but they were worse than the current setup due to being closer in a ship so large. As for the parts where the guns where, they are just details, as you can see in the thumbnail
-engines: as with guns, it was alredy tried but the result was very bad because of the pods being almost in the middle of the ship;

There are no links to previous discussions in this pull request, so I don't know what has been tried already. If it didn't look good, fair enough.

>Putting the guns so close together on the nose of the ship is not in the style of Betelgeuse at all.

Sorry to be critical, but although the Firebird and Leviathan resemble each other, I fail to see any common style shared amongst the other Betelgeuse ships. If, for instance, the Berserker design would be followed, then perhaps the Behemoth's guns should be moved to the outermost compartiments.

Anyway, one question remains: why did you double the number of engines? Having four engines visible would make sense for agile warships such as the Osprey (130 engine capacity), or, indeed, the M Leviathan E (220), but not really for slow freighters with limited engine capacity such as the Behemoth (90) or the comparable Bulk Freighter (85).

beccabunny:
@finite-galaxy oh no the discussions didn't happen on github, sorry if it looked implied, and it isn't a problem at all that you didn't know about that, I was just explaining the reasons behind that.
About the 4 engines, the (main) reason behind them is for prettiness, like with the rest of the detailment. Also, the Argosy and the Fury have both 4 engines, despite having only 80 and 70 engine capacity

finite-galaxy:
> Also, the Argosy and the Fury have both 4 engines, despite having only 80 and 60 engine capacity

Yes, but a single Behemoth engine is much larger (sprite pixels) than all four Fury engines combined. My point is that the Behemoth has rather mediocre engine stats, especially for a ship its size; cf. Bactrian (180), Carrier (210), Cruiser (170), Dreadnought (190).
I like prettiness, but also some kind of correlation between sprite and stats. I wouldn't mind a *modified* Behemoth with four engines, but I think two is enough for the default.
It's a matter of taste, I suppose; I don't know how others feel about it.

beccabunny:
Yup, the ship itself has larger engines, but the total area of the flare will be the same because of the root2 rule applied to the engines' zoom. That said, it is indeed a matter of taste.

finite-galaxy:
To clarify, I mean the visible engines of the ship sprite, not the number of flares.

jafdy:
I note that you compare it to 3 warships and the Bactrian, hardly the best choice for a freighter comparison. 
For all you know, most of those rearmost pods are not engines, they could be space for cargo (that for some reason or other) needs to be kept warm.

kikotheexile:
To say that I, personally, generally avoid GitHub like the plague else I also would have that fancy contributor title would be an understatement. 

All behind the scenes discussion on this happened, quite literally, behind the scenes. Completely out of public eye. I am simply letting you know that some of your ideas were already brought up and discussed in more detail than is possible via GitHub or another forum setup, and were decided against. 

That is not a shot at you, nor stating that you are giving bad feedback. It is simply a statement of "Good idea, but we already decided against that, and these are now done."

data/ships.txt
```diff
@@ -437,16 +437,18 @@ ship "Behemoth"
...
+	engine 34 143 0.8
+	engine -34 143 0.8
+	engine 56 148.5 0.6
+	engine -56 148.5 0.6
+	turret 47.5 33.5 "Heavy Laser Turret"
+	turret -47.5 33.5 "Heavy Laser Turret"
+	turret 22 66.5 "Heavy Laser Turret"
+	turret -22 66.5 "Heavy Anti-Missile Turret"
+	turret 31 -50 "Heavy Anti-Missile Turret"
+	turret -31 -50 "Heavy Laser Turret"
```
Amazinite: Should be engine, gun, turret, not gun, engine, turret. The Heavy Shuttle hardpoints were also shuffled around.
Also, since hardpoint coordinates are saved in the save file, realize that drastically moving hardpoints will make the ships of current saves look weird.

beccabunny:
> Why is it that the Shuttle's animation was changed? The file names had `=`, but now its `-`. This makes it difficult to compare the old and new versions.

Damn. What happened was that when I was doing the Shuttle, I forgot about the mention of pre multiplication MZ does in the blending modes page, so when I swirched the images to the new ones they appeared very bad in game, and when I switched it to - it appeared better so I supposed that I changed it to = for some reasons while working on The Animated Ships plugin and forgot about it later. When I will get home I will do the pre multiply with MZ's tool to correct it.

For unknown reasons my phone refuses to select the rest of the comment's parts so I can't quote them directly, but to summarise it is mainly "Thanks for pointing them out, I will fix as soon as I get home"

Tadrix:
What are those rectangular holes on the side of the cockpit and on the front/starboard side of Behemoth's cargo containers? And why are they there? (At least these black things look awfully like holes.) @beccabunny

Lineth:
Cargo loading bays I presume.

beccabunny:
I should have fixed most of the issues, except the Shuttle one since i can't figure out how to premultiply the layers from the blending modes page, specifically how to use Michael's c++ program; i hope to figure it out tomorrow, otherwise it will have to wait another week since i'll go on holiday again for 5 days

beccabunny:
I tried to use MZ's program for premultiplying in many times, but I failed; if anybody that understands coding can try, the program is
[blend.zip](https://github.com/endless-sky/endless-sky/files/3511195/blend.zip)
and the sprites that need to be mixed are
[shuttle.zip](https://github.com/endless-sky/endless-sky/files/3511197/shuttle.zip)
All I could get was a bunch of errors on the command prompt, but hopefully somebody else will have more luck

Pointedstick:
I have to agree that the huge engine cables look a bit out of place on the Star Queen. Otherwise these look fantastic! I think the new Behemoth is fine (better than fine; much better, obvs).

beccabunny:
@Pointedstick i removed the star queen's cables and detailed that part instead. Now (i think) the last issue still unsolved is the Shuttle, but i am still unable to use MZ's program to mix the images with different blending modes

Pointedstick:
> @Pointedstick i removed the star queen's cables and detailed that part instead.

Much better!

> Now (i think) the last issue still unsolved is the Shuttle, but i am still unable to use MZ's program to mix the images with different blending modes.

What's the problem with the shuttle? It looks great to me.

> (p.s. the next shipyard will probably be Syndacate Shipyards, so when i will start it and do the Vanguard would you prefer to do the new Cap Nate Vanguard yourself or should i change it as well myself?)

Go right ahead! Someone else even made the original captain Nate vanguard anyway. I don't feel any possessiveness. :)

beccabunny:
@Pointedstick the shuttle's top down was premultiplied to have part of in in alpha blending and part in additive blending (I think mentioned at the bottom of the blending modes page) by MZ using a program in the Endless Sky Tools repo, but I was unable to use it so I just used = instead of - in the frames, unfortunately this way it has only one blending mode. A few comments above I attacked the program and the files that (I think) should be mixed if you want to give it a try

Pointedstick:
> @Pointedstick the shuttle's top down was premultiplied to have part of in in alpha blending and part in additive blending (I think mentioned at the bottom of the blending modes page) by MZ using a program in the Endless Sky Tools repo, but I was unable to use it so I just used = instead of - in the frames, unfortunately this way it has only one blending mode. A few comments above I attacked the program and the files that (I think) should be mixed if you want to give it a try

I mean, specifically, in terms of actual visuals... what's wrong with your new shuttle? What problem is being caused by the lack of the above?

jafdy:
You get white flashing boxes around the sprite and interesting collision meshes (namely the entire box).

beccabunny:
@jafdy yea that happens if I leave it as = since it is marked as pre multiplied but it actually isn't, using - instead fixes it
@Pointedstick the light is less visible on the ship I think

finite-galaxy:
The new Star Queen looks better indeed!
> p.s. the next shipyard will probably be Syndacate Shipyards
Sounds great! I have a few requests for a Protector, default or variant ( https://github.com/finite-galaxy/finite-galaxy/issues/35 ), perhaps you're willing to have a look at those?

beccabunny:
Oh, actually I alredy wanted many of those changes, first of all the artifacts!

beccabunny:
@Pointedstick @Amazinite I can't figure out how to use the program, so now the options are:
1) remove all the Shuttle changes from this PR;
2) don't change the top down of the Shuttle in this PR, but keep the new thumbnail;
3) use the shuttle with - instead of = between the frames (current state of the PR);
4) wait for somebody that understands coding better than me (and, admittedly, 2-3 people who tried as well) to build the blend.cpp and use it to put together the files a few comments above.

jafdy:
If no one can work out how to use that code the running lights is effectively dead in the water unless MZ does it, so I would say just move on with the updated models, it would be easier to add the lighting effect later then the model.

beccabunny:
https://github.com/endless-sky/endless-sky/pull/4458#issuecomment-522156352 both the source code and the images that it should mix are in this comment

beccabunny:
@Pointedstick Thanks! I updated the PR with the changes. Unfortunately i am on windows, so i can't do that

PR #4462, opened 2019-08-09
Extra ship in FW/Navy fleet attacking Pug if player has a Jump Drive

AlbertNewton:
If the player has a Jump Drive in the mission, FW Pug 2C, when they get a Jump Drive, normally nothing comes from it. With these changes, the player's Jump Drive will be noted by people in conversations a couple of times, and an extra Carrier with fighters and drones will accompany the player into occupied space, using the Jump Drive that the player would have taken but didn't need.
This is my first PR, so I might have got stuff wrong. If I have, then _please_ tell me, and I'll fix any issues.

### DIFF
--- a/data/free worlds reconciliation.txt
+++ b/data/free worlds reconciliation.txt
@@ -1282,7 +1282,7 @@ mission "FW Pug 2C: Jump Drive"
 	on accept
 		outfit "Jump Drive" 1
 		set "fw given jump drive"
-		set "already had jump drive by FW Pug 2C"
+		set "already had jump drive"
 	on visit
 		dialog `You land on <planet>, but realize that your escort carrying Alondo hasn't entered the system yet. Better depart and wait for it to arrive.`
 
@@ -1594,7 +1594,6 @@ mission "FW Pug 3C"
 	destination Hephaestus
 	to offer
 		has "FW Pug 3B: done"
-		not "already had jump drive by FW Pug 2C"
 	passengers 1
 	on fail
 		dialog `You have failed an essential Free Worlds mission. If you want to complete the story line, revert to the autosave or another earlier snapshot of the game.`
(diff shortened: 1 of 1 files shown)


data/free worlds reconciliation.txt
```diff
@@ -1633,14 +1635,65 @@ mission "FW Pug 3C"
 		dialog "You have reached <planet>, but part of your fleet is not here yet. Better depart and wait for them to arrive in this star system."
 
 
+mission "FW Pug 3C(Extra ship)"
```
Bladewood: There should be a space between the mission and the parenthesis. I'd also suggest that "extra ship" either be fully uppercase or lowercase, not a mix.
Or, we can change the mission name to something like "FW Pug 3.1C" or whatever. Same for the other missions.

data/free worlds reconciliation.txt
```diff
@@ -1282,6 +1282,7 @@ mission "FW Pug 2C: Jump Drive"
 	on accept
 		outfit "Jump Drive" 1
 		set "fw given jump drive"
+		set "already had jump drive by FW Pug 2C"
```
Bladewood: This seems pretty lengthy (and explicit, with the mission name reference) to me.

data/free worlds reconciliation.txt
AlbertNewton: Oh, the "Extra ship" capitalisation was a typo. I'm going to try to change that, and try to figure out how to change the PR accordingly

AlbertNewton: OK, done

data/free worlds reconciliation.txt
AlbertNewton: What would you suggest?

AlbertNewton:
How would the player find out that they have a extra Jump Drive? I don't think they get told until they actually get the fleet how many drives there are.

data/free worlds reconciliation.txt
Bladewood: Simply "already had jump drive" might suffice.

Bladewood:
It turns out you're actually told they have seven when you meet with Alastair Korban for the first time. I don't have a better solution here.

data/free worlds reconciliation.txt
Amazinite: You still didn't add a space before the parenthesis for any of the missions.

data/free worlds reconciliation.txt
```diff
@@ -1633,14 +1635,65 @@ mission "FW Pug 3C"
 		dialog "You have reached <planet>, but part of your fleet is not here yet. Better depart and wait for them to arrive in this star system."
 
 
+mission "FW Pug 3C(Extra Ship)"
```
Amazinite: Instead of duplicating the missions with the same conversations but different numbers of NPCs, you should have the escorts be a separate mission akin to FW Pug 4: Escorts.

data/free worlds reconciliation.txt
```diff
@@ -1732,6 +1809,52 @@ mission "FW Pug 4"
 	on complete
 		event "battle for delta capricorni"
 
+mission "FW Pug 4(Extra Ship)"
```
Amazinite: Instead of duplicating this mission, use a `branch` structure to change the necessary bit of the conversation based off of the player having had a JD before.

Amazinite:
Thoughts on this one now that I've patched it up? I got some requests to have the Pug send more ships after you if you have the extra escort, but I think I'll do that at a later date.

data/free worlds reconciliation.txt
```diff
@@ -1640,7 +1654,6 @@ mission "FW Pug 4: Escorts"
 	source Hephaestus
 	to offer
 		has "FW Pug 3C: done"
-
```
jafdy: Is there a reason for the whitespace change?

Amazinite: Either I was going to make changes to this or was reverting changes to this made by Albert and this stuck around. There's already a mismatch in the data in terms of if there should be whitespace between the `to offer` and the rest of the mission so it's no big deal.

Amazinite:
>Currently, having the extra ship killed on the way to Hephaestus doesn't fail the mission like losing any other ship and further, if it is lost then it will reappear for the next mission.

Fixed by making FW Pug 3C (Extra) a different mission entirely instead of something offered alongside FW Pug 3C, so now not only will killing the extra escort fail the mission, but it must also reach Hephaestus alongside the other escorts (whereas before you could actually duplicate it with a jump and non-jump variant if the extra ship didn't reach Hephaestus with the rest of the escorts).

Issue #351, opened 2015-12-02
Pause sound effects when bringing up certain UI panels


wkoot:
If you go to the map while jumping, the gfx are interrupted but sfx are not.
A possible solution could be to simply restrict changing view while jumping (as opposed to simply pausing the sfx).

Pointedstick:
I rather like being able to access the map while jumping.

wkoot:
So do I, but every time I get back to the normal view its still stuck jumping.
Otherwise, perhaps let the gfx/state continue until actually entering the next system and then pausing ?

endless-sky:
Basically, this would mean pausing the sound effects when certain dialogs are displayed, and then making sure they get unpaused when the dialog is closed. It's definitely doable, but it could get complicated if I ever want to add UI sound effects.

wkoot:
Is the other option (not interrupting jump gfx) viable? Main reason to pop up map when jumping is because it's a waiting moment.
Note that in EV Nova you can simply not open the map when jumping (sadly you cannot even toggle the map overlay); I assume they had a similar dilemma

AdamKauffman:
Hi, I'm new here. Both to this project and to C++. Please excuse any atrocities I have committed to the code. 

I have a patch that shows simple pause and resume of all game effects without limiting music and UI effect. Music and UI each have a dedicated ALuint source that is not a member of the main sources vector. In my brief testing I have seen some strange behaviors and it might just be due to the fact that I have no idea what is really going on here. I would guess that there is a race between my pausing and the Audio::Step thread. I believe much of the strangeness is due to fading sounds and looping sounds. Sometimes they pause and resume cleanly. Other times they don't. I'd like to get the code out here so that anyone else working on this issue can see that I've made good progress. 

I'm not sure this patch is "pull ready". What's the best way to get testing\opinions? Take a look at my fork. Let me know if I should create a PR, if that's easier.

PR #4467, opened 2019-08-12
Pause sounds when a MenuPanel displays over the GamePanel

ghost:
Refs #351 

Initial testing shows that the pause behavior works fairly well. You might find this behavior better than it was before.

KNOWN ISSUES:
Some fading sounds may still play after they should be paused. Some looping sounds might live longer than they should after a resume. Might have to do with timing of the pause coinciding with the step which contains a swap command that might befoul the vector of sources we are attempting to pause.


MCOfficer:
it's always nice to see new contributors, but you might wanna rename this to something more meaningful ;)

Zitchas:
Relevant link: #351 
It is nice to get new contributors, especially when they dig into a section of code that most others aren't touching.

AdamKauffman:
Rebased and fixed the title to be more descriptive.

Pointedstick:
Yes, thanks for the PR!

All the changes to code::blocks and .gitignore files are unrelated; please revert.

AdamKauffman:
Honestly, I don't know how. I tried ` git checkout -- EndlessSky.cbp ` and it does nothing. Those changes are required for building on the latest version of everything on 64-bit Windows. The ignored file is one automatically generated by codeblocks that only contains local paths. It should be ignored. The change to the CBP file may be undesired. As a new developer, that's the environment I ended up with after installing everything with default options. So other Windows developers starting from scratch may have the same setup.

source/MainPanel.cpp
```diff
@@ -132,10 +133,18 @@ void MainPanel::Step()
 	// and updating the isActive flag).
 	StepEvents(isActive);
 	
+	// Run the game, or pause the game
 	if(isActive)
-		engine.Go();
+	{	 
+		Audio::Resume();
```
tehhowch: Should Resume and Pause be called every step? or only when the `isActive` state toggles from true to false or vice versa?

source/Audio.cpp
```diff
@@ -256,10 +264,13 @@ void Audio::Play(const Sound *sound, const Point &position)
 	if(!isInitialized || !sound || !sound->Buffer() || !volume)
 		return;
 	
-	// Place sounds from the main thread directly into the queue. They are from
-	// the UI, and the Engine may not be running right now to call Update().
+	// Play sounds from the main thread directly on dedicated UI Source
 	if(this_thread::get_id() == mainThreadID)
-		queue[sound].Add(position - listener);
+	{
+		uiSoundFX.emplace_back(sound, uiSource);
```
tehhowch: Do elements actually ever accumulate in the `uiSoundFX` vector? it seems whenever any would be added, they are then removed. Does it make sense to instead simply construct the source here, e.g.
```c++
{
  auto toPlay = Source{sound, uiSource};
  alSourcePlay(uiSource);
}
```
as it will be automatically cleaned up when this block scope exits.

Or, is the vector needed, as the Source needs to exist for some time, in which case the call to `pop_back` needs to be elsewhere?

source/Audio.h
```diff
@@ -57,6 +57,15 @@ class Audio {
 	// "listener". This will make it softer and change the left / right balance.
 	static void Play(const Sound *sound, const Point &position);
 	
+	// Pause playing of all sources of audio.
+	static void Pause();
+	
+	// Pause one source by ALuint ID
+	static void Pause(unsigned int id);
```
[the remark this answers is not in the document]
AdamKauffman: Fixed

source/Audio.cpp
AdamKauffman: Good point. Further, do we need the toPlay at all? The return isn't used. We seem to be looking for the side effect of modifying the uiSource passed in to Source.

I made it like this:
`	{
		Source{sound, uiSource};
		alSourcePlay(uiSource);
	}`

and it still appears to work. However, I'm not sure if that is bad for some reason.

source/MainPanel.cpp
AdamKauffman: Yes, if `isActive` were a toggle. Unfortunately it was defined on the fly. I made a new version where `isActive` lives a private life modified by Pause and Resume procedures. This gives you the behavior you wanted. Let me know if this is what you were thinking of.

source/Audio.cpp
tehhowch: It is indeed bad to create an object that immediately goes out-of-scope.

We must guarantee that the `Source` remains valid while OpenAl needs it. Whether that is accomplished by storing it in a vector and then removing it elsewhere, or whether it living for this block scope only is sufficient, is something that you will need to determine.

source/MainPanel.cpp
tehhowch: I was thinking more of defining another  local variable that has the original value of `isActive`, e.g. `wasActive`, before going through the steps that could modify isActive. Then, if no longer active, we call audio::pause.

I think there is not a similarly easy way to determine the exact step in which MainPanel becomes the top panel without a new class member variable, so we either call resume every active step (which should be low cost as the paused sources list will be empty except on the first active frame), or use the new class member with the local `isActive` to determine which frame is the first active one.

AdamKauffman: Except then `isActive` and `wasActive` are reset on each step. There would be no tracking of `wasActive` across steps. Some variable needs to live outside the scope of `Step` in order to track what was active. That's my `paused` bool. Instead of comparing the two at the end I'm using switches to change the state. This allows you to compare and switch from anywhere.

source/Audio.cpp
AdamKauffman: Ok, I think I figured it out. I am now creating the source and using that ID to play.

Amazinite:
Since this is an orphaned PR from a literally dead account, and there's an associated issue already open for it (#351), I think this should be closed.

Issue #4307, opened 2019-05-30
segfault when leaving a planet, probably due to non-licensed ship


bmorel:
Using Debian 9 "stretch", currently the stable one, with a libsdl2 built from debian source but with disabled dbus support, I have a segfault when I leave a planet with a republic ship I just captured.
Using current last commit, aka 756b395165774ec1dce1ea967d7923d7338cabd1, but the bug is present since at least 1 week.

I attach the savegame that does the crash to this bug report. I would have loved to fix the problem myself, but can't find how to build debug version with scons... 

[Rake Skullgrab 2.txt](https://github.com/endless-sky/endless-sky/files/3236929/Rake.Skullgrab.2.txt)

Zitchas:
Tested using savefile provided via pastebin, and Endless Sky crashed when leaving the planet. No errors generated in errors.txt

Using up to date repo as of 756b395 and MCO's nightly [build #160](https://ci.mcofficer.me/job/endless-sky-win64/lastBuild/) which is for the same commit.

Running Windows 10.

tehhowch:
As mentioned in Discord, our SCons config allows specifying the `mode` parameter, with 3 values: `release` (default if omitted), `debug`, and `profile`, e.g.
```bash
scons mode=debug
```
The details of these modes (and other possible arguments to scons) are available in the SConstruct file. I'll update the relevant wiki and developer readme as needed this weekend.

bmorel:
Thanks, that will be very useful to people not knowing internal details of scons.

The bug is nastier than what I thought, here is what I can say about it:

Stack as shown by cgdb:
#0  0x000055555557c7d0 in std::vector<StellarObject, std::allocator<StellarObject> >::begin (this=0xb8) at /usr/include/c++/6/bits/stl_vector.h:558
#1  0x000055555564b809 in Fleet::ChooseCenter (system=...) at source/Fleet.cpp:536
#2  0x000055555564b2a2 in Fleet::Place (system=..., ship=...) at source/Fleet.cpp:473
#3  0x000055555561b299 in Engine::Place (this=0x555563b6a478) at source/Engine.cpp:292
#4  0x00005555556db22e in MainPanel::OnCallback (this=0x555563b6a440) at source/MainPanel.cpp:186
[...]

PR #4468, opened 2019-08-12
Restrict reparenting of mission-affiliated carried ships to the same missison

tehhowch:
Refs #4307 

Also reduces the frequency with which carried ships that are escorts of a ship that cannot carry them seek different parents.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -711,7 +711,7 @@ void AI::Step(const PlayerInfo &player)
 				
 				auto parentChoices = vector<shared_ptr<Ship>>{};
 				parentChoices.reserve(ships.size() * .1);
-				auto reparentWith = [&](const list<shared_ptr<Ship>> otherShips) -> bool
+				auto reparentWith = [&it, &gov, &parent, &parentChoices](const list<shared_ptr<Ship>> otherShips) -> bool
 				{
 					for(const auto &other : otherShips)
 						if(other->GetGovernment() == gov && other->GetSystem() == it->GetSystem() && !other->CanBeCarried())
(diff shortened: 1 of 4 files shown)


source/AI.cpp
```diff
@@ -702,29 +702,59 @@ void AI::Step(const PlayerInfo &player)
...
+					? find_if(player.Missions().begin(), player.Missions().end(), [&it](const Mission &m)
 					{
-						if(!other->IsDisabled() && other->CanCarry(*it.get()))
+						auto nend = m.NPCs().end();
+						return nend != find_if(m.NPCs().begin(), nend, [&it](const NPC &n)
 						{
-							parent = other;
-							it->SetParent(other);
+							auto send = n.Ships().end();
+							return send != find_if(n.Ships().begin(), send, [&it](const shared_ptr<Ship> &s)
```
petervdmeer: This part looks very magical to me. Could use some comments on what is happening here.

tehhowch: This set of lambdas is finding the associated mission, if any, in which this ship belongs to one of the NPC blocks, using `std::find_if`.

source/AI.cpp
```diff
@@ -702,29 +702,59 @@ void AI::Step(const PlayerInfo &player)
...
 				// Find a parent for orphaned fighters and drones.
 				parent.reset();
 				it->SetParent(parent);
-				vector<shared_ptr<Ship>> parentChoices;
+				
+				auto parentChoices = vector<shared_ptr<Ship>>{};
 				parentChoices.reserve(ships.size() * .1);
-				for(const auto &other : ships)
-					if(other->GetGovernment() == gov && other->GetSystem() == it->GetSystem() && !other->CanBeCarried())
+				auto reparentWith = [&](const list<shared_ptr<Ship>> otherShips) -> bool
```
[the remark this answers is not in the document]
tehhowch: The functionality in this lambda isn't needed anywhere else, even within the AI class, except within this block. For that reason, I chose to use a lambda over an anonymous namespace method or a class method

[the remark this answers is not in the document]
petervdmeer: Okay,  but could you then at least then put a comment just above the lambda function to mention that a lamba function is used here? I know a lot of programmers that are unfamiliar with lambda functions and if such a programmer sees this for the first time in his/her life, then I would like such programmer to at least have a pointer/indication of what is happening here.

tehhowch:
Still need to verify that this fixes the root cause of #4307 before it merges.
I might also add caching of the mission iterator; I need to look at how frequently this code actually gets executed to determine if that has merit.

tehhowch:
Added a guard against a special ship that has no system (which is what causes the segfault in #4307)
Also removed a couple of the nested lambdas in favor of a Mission class method, which should help readability.
It's possible that caching the mission iterator should also be done; adding all accepted missions from the "ManyMissions Tester" save file in #4424 to the save file provided in 4307 jumps the in-flight CPU counter for my laptop from 2% to 20-22% -- I will make a decision on that tomorrow.

Feel free to provide an additional review @petervdmeer

source/AI.cpp
```diff
@@ -702,29 +702,52 @@ void AI::Step(const PlayerInfo &player)
...
+					: player.Missions().end();
+				if(missionIt != player.Missions().end())
+				{
+					auto &npcs = missionIt->NPCs();
+					for(const auto &npc : npcs)
+						if(reparentWith(npc.Ships()))
+							break;
+				}
+				else
+					reparentWith(ships);
```
petervdmeer: Is it possible that this line of code causes a non-special/non-mission fighter to be reparented to a special/mission carrier? And if that happens, would that be bad? (For example due to the carrier no longer having space for a mission fighter?)

petervdmeer: As discussed on discord: not an issue, solved in CanCarry method

Issue #4438, opened 2019-07-18
Proposal for creating ship formations using lines


petervdmeer:
**Is your feature request related to a problem? Please describe.**
This is a proposal for specifying formations for ships (as discussed in #302 and others) by lines that are moved, rotated, extended and repeated.

My intention is that this proposal is sufficiently simple for:
- players to use (in an in-game graphical editor)
- scenario writers to use in missions (directly or through editor).
- developers to implement

The mechanism should also be sufficiently powerful to handle the most common formations that players and writers would want to define.

**Describe the solution you'd like**
The basic structure of a definition of a formation in the data-files would look like:
```
formation <name>:
	line <x> <y> <slots> <direction>
		spacing <nr>
		repeat <deltaX> <deltaY>
			increase <nr>
	line <x2> <y2> <slots2> <direction2>
		spacing <nr2>
```

Meaning of the keywords:
- `formation <name>`: Begins a new definition for a formation with name `<name>`.
- `line <x> <y> <slots> <direction>`: Begins a line at position x,y for a line with `<slots>` positions for ships. The very first ship in the line is at exactly x,y, the line stretches into the direction of `<direction>` with ships `<spacing>` distance apart. The last ship in the line is at `<slot> * <spacing>` distance from the first.
- `spacing <nr>`: The amount of space between ships on a line.
- `extend <nr>`: **(to be removed keyword)** The amount of slots to extend the line with if the formation needs to grow.
- `repeat <deltaX> <deltaY>`: The location where to repeat the line if the formation needs to grow (relative to x,y of previous repeated line).
- `increase <nr>`: The amount of extra slots to add to the line on each grow step.

Some examples:
```	
# Horizontal line formation, growing from left outwards.
formation "H-Line"
	line 0 0 1 90
		spacing 100
		extend 1

formation "Square"
	# Only specify a single top left position.
	# Expands to the right due to direction 90 and extend 1.
	# Expands to below with the repeat section. The repeat section also
	# gets the effect from extend 1, so the shape will remain a square.
	# To get a diamond formation, change the rotation to 135.
	line 0 0 1 90
		spacing 100
		extend 1
		repeat 100 0

formation "Triangle"
	# Specify only the top element and repeating lines 1 bigger every increment.
	# Each repeated line is half of the spacing to the left of the previous one.
	# The lines in the triangle all get filled from the left. If it is desired to
	# alternate the filling between left and right, then specify the first 2 lines
[...]

Issue #3381, opened 2017-12-08
"Linger" behaviors for NPC ships


tehhowch:
Figured I'd poll the community to see what, if any, consensus there is on what it should look like for an NPC to spend extra time in a system, outside of combat.

Currently, ships enter a system, decide if they want to land on a planet or jump, and then do that, immediately. There are a tiny handful of alternate behaviors:
 - surveillance (scan other ships, scan stellar objects, unload drones and jump away)
 - swarming (flit between any ships in the system)
 - mining (mine asteroids for a bit)

One issue brought up previously is that ES has a very high turnover on NPCs, and the lack of various behaviors is generally the root cause of this, as ships without one of those 3 behaviors have only one thing they can do: travel.

In #2990 @elyssaen and I introduce a "survey" behavior (which is exactly equivalent to the "scan stellar objects behavior" of surveillance ships), to increase the amount of time that our travelling NPCs spend in a system, but extending that specific behavior to all ships is probably not the greatest plan (it would look pretty weird to watch a bunch of Merchant Bulk Freighters traveling between the moons in a system).

If we reduce the turnover rate some, by requiring NPCs to do at least something in a system (i.e. we don't let them jump immediately), what kinds of things should NPCs do?
```
(please vote on suggestions you like)
```

tehhowch:
One possibility is to have ships first decide to rendezvous at a randomly chosen point in a system. Ships with no parent would enter a system, pick a point randomly (equally probable to be anywhere within that system's habitable radius), and then only jump out after having traveled to that point. Ships with a parent would naturally follow their parent, and would get ready to jump when the parent had picked its position and arrived there.

tehhowch:
Another possibility is to have some ships choose to spend some time circling a landable planet in the system (equally likely to circle counterclockwise as clockwise).

solardawning:
Circle a landable planet/station makes a lot of sense: you'd expect ships do that while waiting for spaceport control to give landing clearance, or they've sent a landing craft down and leave the ship in orbit. I like it!

Here's another one: Meeting. Two fleets meet up at a picked point in the system, and one ship from a fleet boards a ship from the other, then they split and leave. (Are they friends? Bartering cargo? Who knows! But it looks like something is happening.)

jafdy:
You could make landing ships slow down ahead of time and come in slowly instead of as fast as possible. If there was a fight going on then they would revert to as fast as possible.
[...]

PR #4471, opened 2019-08-12
feat: Formation flying functionality

petervdmeer:
This PR adds formation flying functionality for groups of ships.

Here is a picture to show the "Hexagon" and "Delta Trailing" formations in-game:
formation_flight

Formations are defined in the datafiles. Each ship can specify in which formation pattern it wants to participate. The actual position of a ship in a formations is calculated in the AI class. Different ships can have different formation patterns.

Contents of this PR:
- Part of the data-format as proposed in #4438, in [RFC](https://github.com/petervdmeer/endless-sky/blob/feature_formation_flying/docs/rfcs/text/0052-formation-flying.md) and in the [RFC-PR](https://github.com/EndlessSkyCommunity/endless-sky/pull/52).
  - Example formations in the formations.txt datafile.
- Formations can be formed around a formation leader
  - the carrier for fighters
  - the players flagship for player formations
- AI class will move ships into their formation patterns
- Multiple formation patterns can be formed around a single "formation leader".
  - But this is not preferred, fleets just get one formation to use.
- When a ship leaves a formation, then later ships automatically fill up the open position.
- Formation-patterns set on a ship are saved and loaded.
- Controls to set formations. For default controls the following:
  - shift-G to change the formation on selected ships (or the fleet)
  - ~~shift-F to move selected ships (or the fleet) to a higher ring~~ (ring functionality is removed)
  - ~~shift-H to move selected ships (or the fleet) to a lower ring~~ (ring functionality is removed)
- Formations turn at their own speed if the lead turns fast.
- Rotation and mirroring of formations during things like 180 degree turns
- ~~Formation rings to allow formations to start "more to the outside".~~ (ring functionality is removed)
- ~~Formations automatically resize based on the size of participating ships.~~ (scaling as was done in this pr is removed)

Examples/demonstration-movie:
- [4 formation patterns](https://youtu.be/W0RrTRSSqGs) (by petervdmeer)
- [Delta trailing formation](https://youtu.be/szszsXbuumU) (by Zitchas)
- [Wedge formation](https://youtu.be/iuwoZN2gaSI) (by Zitchas)
- [2 formations, hold/resume](https://youtu.be/T6dCA2hmOlk) (by Zitchas)
- [resizing](https://www.youtube.com/watch?v=IE_oZnYlEFE) based on ships sizes (by petervdmeer)
- [flipping and turning](https://youtu.be/xZsb7saNzRw) based on symmetry (by petervdmeer)


petervdmeer:
Also see this movie https://youtu.be/W0RrTRSSqGs to see formations (4 different patterns) in action.

Pointedstick:
Very cool WIP patch. I think we very much need for the UI to be added before we can merge it though.

petervdmeer:
> Very cool WIP patch. I think we very much need for the UI to be added before we can merge it though.

What if I add formations to some NPC fleets? That would make formations available for players to see (when they encounter those fleets) and for content creators to use (for any new fleets they create).
Would that be sufficient to get this PR merged?

Pointedstick:
Adding formations to certain NPC fleets with no UI might be an option, and would make battles much cooler. If this approach was taken, there should be a way to make an NPC fleet commanded by a plater have a formation too. The player leads quite a lot of fleets during various campaigns, so making the fleet ships fly in formation would be super cool.

If we did this, then maybe a UI to control it could be added later. Note that this is a community-developed game, so my opinion is not law. :) The decision would be made by consensus.

petervdmeer:
My end-goal is that we have all formation flying features available for ES:
- formations set by the player through the UI for player fleets
- a nice formation editor for the player (and for content creators)
- formations on NPC fleets
- automatic mirroring/flipping of formations
- automatic spacing between formation ships based on the size of ships in the formation
- etc

But I expect that trying to do everything all at once in a single PR will not work; the PR would be very large, would take forever to finish, would have multiple merge-conflicts and would be a horribly complex change to review.
So I'm not searching for ways to skip the UI options, but I am searching for ways to split the formation flying functionality into sensible smaller parts (which means multipe PRs).

petervdmeer:
Added formation-flying to NPC fleets and a small mission where the player escorts a group of NPC ships that is flying in formation.
earth_stunt_pilots

Amazinite:
Currently compiles and runs just fine on WIndows.

Zitchas:
This looks really nice. I'm looking forward to trying this out.

bosporos:
Diff for XCode support
```
diff --git a/EndlessSky.xcodeproj/project.pbxproj b/EndlessSky.xcodeproj/project.pbxproj
index ee075daa..40eeaaf5 100644
--- a/EndlessSky.xcodeproj/project.pbxproj
+++ b/EndlessSky.xcodeproj/project.pbxproj
@@ -16,6 +16,8 @@
                62A405BA1D47DA4D0054F6A0 /* FogShader.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 62A405B81D47DA4D0054F6A0 /* FogShader.cpp */; };
                62C3111A1CE172D000409D91 /* Flotsam.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 62C311181CE172D000409D91 /* Flotsam.cpp */; };
                6A5716331E25BE6F00585EB2 /* CollisionSet.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6A5716311E25BE6F00585EB2 /* CollisionSet.cpp */; };
+               938E07E4235970CE00DC2C2B /* FormationPattern.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 938E07E0235970CE00DC2C2B /* FormationPattern.cpp */; };
+               938E07E5235970CE00DC2C2B /* FormationPositioner.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 938E07E2235970CE00DC2C2B /* FormationPositioner.cpp */; };
                A90633FF1EE602FD000DA6C0 /* LogbookPanel.cpp in Sources */ = {isa = PBXBuildFile; fileRef = A90633FD1EE602FD000DA6C0 /* LogbookPanel.cpp */; };
                A90C15D91D5BD55700708F3A /* Minable.cpp in Sources */ = {isa = PBXBuildFile; fileRef = A90C15D71D5BD55700708F3A /* Minable.cpp */; };
                A90C15DC1D5BD56800708F3A /* Rectangle.cpp in Sources */ = {isa = PBXBuildFile; fileRef = A90C15DA1D5BD56800708F3A /* Rectangle.cpp */; };
@@ -184,6 +186,10 @@
                62C311191CE172D000409D91 /* Flotsam.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = Flotsam.h; path = source/Flotsam.h; sourceTree = "<group>"; };
                6A5716311E25BE6F00585EB2 /* CollisionSet.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = CollisionSet.cpp; path = source/CollisionSet.cpp; sourceTree = "<group>"; };
                6A5716321E25BE6F00585EB2 /* CollisionSet.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = CollisionSet.h; path = source/CollisionSet.h; sourceTree = "<group>"; };
+               938E07E0235970CE00DC2C2B /* FormationPattern.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = FormationPattern.cpp; path = source/FormationPattern.cpp; sourceTree = "<group>"; };
+               938E07E1235970CE00DC2C2B /* FormationPattern.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = FormationPattern.h; path = source/FormationPattern.h; sourceTree = "<group>"; };
+               938E07E2235970CE00DC2C2B /* FormationPositioner.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = FormationPositioner.cpp; path = source/FormationPositioner.cpp; sourceTree = "<group>"; };
+               938E07E3235970CE00DC2C2B /* FormationPositioner.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = FormationPositioner.h; path = source/FormationPositioner.h; sourceTree = "<group>"; };
                A90633FD1EE602FD000DA6C0 /* LogbookPanel.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = LogbookPanel.cpp; path = source/LogbookPanel.cpp; sourceTree = "<group>"; };
                A90633FE1EE602FD000DA6C0 /* LogbookPanel.h */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.h; name = LogbookPanel.h; path = source/LogbookPanel.h; sourceTree = "<group>"; };
                A90C15D71D5BD55700708F3A /* Minable.cpp */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; name = Minable.cpp; path = source/Minable.cpp; sourceTree = "<group>"; };
@@ -465,6 +471,10 @@
                654D33611BE92C9200D1E5AB /* source */ = {
                        isa = PBXGroup;
                        children = (
+                               938E07E0235970CE00DC2C2B /* FormationPattern.cpp */,
+                               938E07E1235970CE00DC2C2B /* FormationPattern.h */,
+                               938E07E2235970CE00DC2C2B /* FormationPositioner.cpp */,
+                               938E07E3235970CE00DC2C2B /* FormationPositioner.h */,
                                A96862CD1AE6FD0A004FE1FE /* Account.cpp */,
                                A96862CE1AE6FD0A004FE1FE /* Account.h */,
                                A96862CF1AE6FD0A004FE1FE /* AI.cpp */,
@@ -880,6 +890,7 @@
                                A96863E61AE6FD0E004FE1FE /* Point.cpp in Sources */,
                                A96863DE1AE6FD0E004FE1FE /* OutfitterPanel.cpp in Sources */,
                                62C3111A1CE172D000409D91 /* Flotsam.cpp in Sources */,
+                               938E07E4235970CE00DC2C2B /* FormationPattern.cpp in Sources */,
                                A96863B91AE6FD0E004FE1FE /* Effect.cpp in Sources */,
                                A96863AE1AE6FD0E004FE1FE /* ConditionSet.cpp in Sources */,
                                A96863DC1AE6FD0E004FE1FE /* Outfit.cpp in Sources */,
@@ -934,6 +945,7 @@
                                A96863F21AE6FD0E004FE1FE /* Ship.cpp in Sources */,
                                B5DDA6942001B7F600DBA76A /* News.cpp in Sources */,
                                A96863D01AE6FD0E004FE1FE /* main.cpp in Sources */,
+                               938E07E5235970CE00DC2C2B /* FormationPositioner.cpp in Sources */,
                                A96863BE1AE6FD0E004FE1FE /* Fleet.cpp in Sources */,
                                A98150821EA9634A00428AD6 /* ShipInfoPanel.cpp in Sources */,
                                A96864011AE6FD0E004FE1FE /* Table.cpp in Sources */,
```

petervdmeer:
Rebased to a recent master and added some requested formations.

petervdmeer:

Zitchas:
Having this merged before the player UI is merged makes sense to me. Even if we can't get the player fleet looking all nice and orqanized, it would be great to have (some) NPC fleets flying around in formation. Especially patrols. I'd love to have Remnant drone wings flying around with a group of drones in a hexagon formation around a fighter. 

As far as that goes, for testing purposes, you mentioned being able to put a formation into the ship definition so when you buy a ship it comes with the formation. Could you post an example?

petervdmeer:
@Zitchas Sure, here is a data-file [formation_ships.txt](https://github.com/endless-sky/endless-sky/files/4465401/formation_ships.txt) which you can put in the endless-sky data folder (or in any plugin data-folder) to get 2 new types of sparrows; one that flies in Wedge formation, another one that flies in Delta-Tail formation.

There are two ways to get an ES-executable and data-folder that can use this file:
- Either you clone/checkout and compile the branch that belongs to this PR by yourself.
- Or you take the option provided by MCO to download the executable as artefact directly from the latest CI-run here https://github.com/endless-sky/endless-sky/pull/4471/checks?check_run_id=579192312 . If you use this option, then you will also need to download the formation-definitions file from this PR (which can be done directly from the Github website).

Zitchas:
Well, I had fun testing these out. Using a fleet of 100 stock sparrows in Delta formation, and another 100 in Wedge formation.

Delta formation:
https://youtu.be/szszsXbuumU 

Wedge formation:
https://youtu.be/iuwoZN2gaSI 

Two formations in flight, with one being given the hold command, and then being released from it:
https://youtu.be/T6dCA2hmOlk 

Overall impression: 
Looks impressive, especially when doing slow turns. Fast manoeuvres tend to result in a short bit of chaos and then the formation appearing again. I could see this looking *really* good for patrol fleets, or missions where one is marshalling a fleet. I really enjoyed having my fleet follow me around in formation like this. I think this is a definite solid candidate for adding to the game. (I thought so before, but actually getting to fly around with it really cemented it.)

The possibility of having fleet formations like this really makes me miss hyperspace jump minimum system distances. If all ships had to head out to a certain radius before jumping, then having fleets like these (I'd go so far as to say the Delta formation should be a "default" behaviour for all ships in the absence of any other formation orders) would look awesome as they cruise in and out of a system. I could see a lot of merchants using something like that Hexagon shaped fleet, for instance. 

For making the galaxy seem more interesting, I'd be inclined to create, say, a half dozen formations plus a "no formation" option; and have an automatic behaviour if they don't have a formation assigned, they automatically pick one of the seven default options when they are created. So some fleets will still be random mash fleets, and the "swarming" tag might override this, as swarming is kind of a formation behaviour of sorts,  but most fleets would have some order to them.

beccabunny:
Personally I would prefer if it was possible to give them a list of possible formations which get randomly chosen on spawn, and keep the default as no formation

Zitchas:
It occurs to me that probably the one piece of functionality that we need in order for this to really be useful is being able to specify formation on fleets. While specifying it in the ship definitions is nice, we probably don't really want to be creating so many variants... But if we can put the formation on a fleet, that solves that problem. Even better if it can work with the variants of variants thing so the formation type can be a variant list to pick one of whatever is in the list.

petervdmeer:
The mission `To Earth Day celebration [2]` already specifies the formation in the fleet section:

```
npc
	government Merchant
	personality timid escort
	fleet
		names "civilian"
		formation "Wedge"
		variant
			"Flivver (Racing)" 9
```

What we do need for NPC fleets that don't follow the player is to add some behaviours on when and how long the ships stay in formation. Formations are currently formed when a fleet is idle and as noticed in #3381; fleets like the Navy fleets are always busy (scanning, fighting pirates or jumping away). they are not idle for a sufficiently long time to get into formation.

Another option is to create some "formation personalities" that specify how long ships stay in formation (stay in formation until you are within 2* weapons range of enemies for example, stay in formation during scanning or even form the formation during jump).

Zitchas:
Ah, excellent! for some reason I failed to notice that, and thought those were simply using ship attributes like the demonstration sparrows.

petervdmeer:
I noticed the discussion on Discord about how difficult it is to try out formations without an in-game control to set formations.
The last commit is a (temporary) change to allow setting formations by pressing shift and then pressing G. It switches to the next formation for the fleet (or for the selected ships if the player has ships selected).

petervdmeer:
I added some (temporary) functions to demonstrate the ring function for formations, rings can be controlled by shift-F(Fight) and shift-H(Hold). Here is a screenshot of sparrows flying in a protective `Delta (center)` outer ring to protect my valuable star barges:
formations_ring_example

The ring implementation here is a bit simple, whenever a ship from a high ring is encountered, then the specific formation switches to the higher ring (if it is not already above it). The official/later implementation should place ships on lower rings also on lower rings after a ship on a high ring is placed.

petervdmeer:
I created two movies of recently added functionality:
- [spacing](https://youtu.be/IE_oZnYlEFE) based on the ships in the formation.
- [flipping and turning](https://youtu.be/xZsb7saNzRw) (with rotational symmetry).

Zitchas:
I'm really enjoying playing around with it, and just discovered that shift-g will rotate between formations. Definitely fun, definitely worth having in-game.

1) formations default to starting out at ring 0, which is fine. But sometimes (especially with big ships) that ring 0 is a bit too far out for me, so I find myself wishing for some way to pull them in a bunch. Basically, wishing there was the equivalent of a -1 or -2. Maybe -3.

2) Just flying around with a bunch of escorts, I noted that the fighters & drones launched by one of my escorts seem to be spinning around more or less on top of their mothership instead of falling into the formation. I kind of wish that they'd fall into the full overall formation in the outermost layer.

3) the shift-F, shift-G, and shift-H are actually a pretty good combo. I would recommend those being included with this merge. Maybe take it out at some point down the road if we come up with something better; but for now I think this is a really good way of handling it and the replacement will have to be very good to warrant this being taken out.

4) Is there anything in particular standing in the way of this being merged?

petervdmeer:
1 The distance between ships is currently determined by the (max) radius of the biggest ship in the formation, not by the ring-number. Adding negative rings wouldn't help, but it is possible to create "cozy" variants of the current formation patterns. Open formations.txt, copy an existing definition, give it a new name, and multiply all coordinate-numbers by something like 0.75. This is quite a bit of work though and you shouldn't change the non-coordinate numbers.

For the long term we should probably think of something smarter (a manual override for scaling, or scaling based on a combination of px and ship-sizes). @tehhowch also made a remark on the RFC that he would like to see the option of using px for scaling.
Most large ships are much longer than that they are wide, so we might also want to use length and width for some formations instead of the radius.

2 Ships form formations around their parent. Regular escorts have the players flagship as parent, but drones and fighters have the ship they dock with as parent. You should see the drones and fighters fly in formation, just not around the flagship and the formations would be small due to the small sizes (radiusses) of the fighters/drones.

3 Fine for me to leave them in.

4 I think this PR (and it's RFC-PR) are waiting for (technical/code) reviews.

Zitchas:
OK, thanks for the answers.
1. I'll try changing some values and see what I get. In terms of negative numbers, I meant from a user point of view. I realize that from a coding perspective adding in negative numbers might be problematic. Having a way to scale things up or down in a similar manner to the current ring up or ring down sounds like it would be really nice way to implement this. (again, from a user perspective)
2. Regarding the fighters/drones forming up on their mothership - that makes sense. I like that. One suggestion, however: Would it be possible to have the fighters/drones use some sort of "radius of mothership + their own radius" as a starting point? I note that they appear to be using a formation based on just their own size, which means that they usually overlap with their mothership. Outside of the initial ring, using just their own size looks good. It's just that proximity to the mothership that is a bit on the close side.

petervdmeer:
> One suggestion, however: Would it be possible to have the fighters/drones use some sort of "radius of mothership + their own radius" as a starting point?

Yes, the ring setting already allows you to do this manually; just select the starting ring for all the ships at sufficient distance from the mothership.
The (starting) ring can also be set for mission NPCs in the datafiles, so there it is also possible to use this.

Automating this would also possible, but automatic setting of the starting ring it would work nicely for formations around a mothership, but it could look strange for a formation that is in front or behind the mothership.

petervdmeer:
> Having a way to scale things up or down in a similar manner to the current ring up or ring down sounds like it would be really nice way to implement this. (again, from a user perspective)

Implementing this through hotkeys is actually not very difficult; just add a scaling factor to the formations which defaults to 1 and gets changed by hotkeys. Most work will be in finding a suitable hotkey.

I  think we are also getting into the domain of the formation editor UI as discussed in #4606. Hotkeys are nice, but it might be easier for the user to change the scaling through something like sliders directly on the in-flight screen. (Such sliders will be quite a bit more work to implement though.)

Zitchas:
Would it be possible for you to post another example of the block formation that is much tighter? When flying with a mixed fleet (a couple of almost every size of ship) the formation ends up really spread out.

I tried to make a condensced version by halving all the spacing values, and it ends up with the formation non-centered.

Here's what I tried.
```
# A (hexagonal grid based) block of ships in front of the lead ship.
formation "Block"
	symmetry longitudinal
	line 0 -3.464 2 180
		spacing 1.732
	line 1 -2.598 2 180
		spacing 1.732
		repeat 2 0
	line -1 -5.196 2 180
		spacing 1.732
		repeat -2 0
	line -2 -2.598 2 180
		spacing 1.732
		repeat -2 0
	line 2 -3.464 2 180
		spacing 1.732
		repeat 2 0
```

petervdmeer:
@Zitchas Good try. I see that you halved most of the Y-coordinates, but had some copy/paste errors on the 3rd and 4th line. Here is your example with the 3rd and 4th line also on halved Y-coordinates:
```
# A (hexagonal grid based) block of ships in front of the lead ship.
formation "BlockYhalf"
	symmetry longitudinal
	line 0 -3.464 2 180
		spacing 1.732
	line 1 -2.598 2 180
		spacing 1.732
		repeat 2 0
	line -1 -2.598 2 180
		spacing 1.732
		repeat -2 0
	line -2 -3.464 2 180
		spacing 1.732
		repeat -2 0
	line 2 -3.464 2 180
		spacing 1.732
		repeat 2 0

```

What you probably wanted to do was to also halve the X-coordinates. Here is an example with the X-coordinates also halved:

```
formation "BlockHalf"
	symmetry longitudinal
	line 0 -3.464 2 180
		spacing 1.732
	line 0.5 -2.598 2 180
		spacing 1.732
		repeat 1 0
	line -0.5 -2.598 2 180
		spacing 1.732
		repeat -1 0
	line -1 -3.464 2 180
		spacing 1.732
		repeat -1 0
	line 1 -3.464 2 180
		spacing 1.732
		repeat 1 0
```

Sizing/scaling the formations based on ship sizes doesn't seem to work as well as I hoped.

For the future I think we should go to px (pixels) based formation positions only (and just select different formations for small and for large ships). Or we should go to a combination of px-based and ship-size-based.

Zitchas:
Just as an idea, what about having a combination thereof? For instance, having a purely px based formation that is - at the base level - tight enough to look good for small ships such as fighters, but then it has some kind of radius limiting on it so that if a ship is going to overlap another ship, it skips that spot and moves on to the next one. 

And now I'm going to go copy that BlockHalf formation in and see how it looks. Thanks!

Zitchas:
Some additional images of testing. All fleet images are using the same fleet in the same order, at the same zoom level, namely, max zoom.
Block formation:
Formation Testing - 1 Block
HalfBlock formation:
Formation Testing - 2 HalfBlock
Bowtie:
Formation Testing - 3 Bowtie
Hexagon
Formation Testing - 4 Hexagon

petervdmeer:
@Zitchas:

> Some additional images of testing. Block formation

Your block formation example uses the non-centered definition instead of the orginal definition.

> Overall, the formations look good - but are very spread out. 
> As such I think having "half" versions of each of the formations you have included here would be very helpful (and look really good, too.)

I updated the [RFC-PR](https://github.com/EndlessSkyCommunity/endless-sky/pull/52) with the option to specify formation positions in pixels, in ship-radiusses (what we have here now), in ship-widths and in ship-heights and most importantly; in a combination of those. The combination will allow us to bring the current formations close to the half-sized formations for large ships, but will keep small ships sufficiently separated. This is not implemented yet, but it is at least documented for review.

> For instance, having a purely px based formation that is - at the base level - tight enough to look good for small ships such as fighters, but then it has some kind of radius limiting on it so that if a ship is going to overlap another ship, it skips that spot and moves on to the next one.

I was also thinking about skipping spots when there is overlap, but the tricky part is that overlap applies to ships before and after the big ship, in the same, in earlier and in later rings. This can become quite complex quite fast and I'm not sure if it will look nice.
So what I did instead is to add an `instance` parameter to fleets/NPC formations in the [RFC-PR](https://github.com/EndlessSkyCommunity/endless-sky/pull/52). This allows you to create a Block formation for small ships and another "Block" formation for large ships without the small ships getting the spacing of the large ships. The `instance` parameter is not implemented yet, but at least the idea is documented for review.
You can already simulate this by copying the "Block" formation to "Block2" (or another name) and use one of them for small ships and the other one for large ships.

> One thing I've noticed is that there is a fair amount if "jitter" in fleets.

I also noticed this and I was thinking of adding a little bit of backwards/centering push force for ships ahead of their formation position. I feel it is okay if ships need to use their thrusters if they are behind their desired position, but the jitter from the >180 degree turns for ships ahead of their positions is indeed a bit distracting.

Zitchas:
Ah, yes, I did use my hacked up version of the Block definition. Sorry.

I don't think we should worry too much about overlap. As currently demonstrated, there is virtually no overlap in either your original or halfblock formations, with the sole exception being carried fighters sometimes overlap their carrier. Which I don't see as much of a problem. Adds to the 3d feel of things a bit anyway. 

Regarding the jitter, that sounds like a good method of handling it. As far as lore is concerned, it is ships using their steering and landing thrusters to maintain position. It makes sense to me, and would make things look even better.

Zitchas:
How's work on Formations going? Given how good it was when I tested, it looks like a lot of polish and improvements have happened since then.

petervdmeer:
> How's work on Formations going? Given how good it was when I tested, it looks like a lot of polish and improvements have happened since then.

Many features are indeed implemented and polished. Most formations should now have the ship-to-ship spacing that you recommended and there is some basic support for spheres, circles and arcs, but they aren't looking too great yet.

Thomas Ballinger also made a very nice start with an [online formations editor](https://observablehq.com/@ballingt-personal/wip-formations-from-formation-pr).

This PR and Thomas' editor are currently blocked on the syntax in the formation definitions; Tehhowch mentioned on [discord](https://discord.com/channels/251118043411775489/266345072554016768/746371546163183776) that he feels he needs to submit an alternative, more declarative, syntax for the formations.

Tehhowch asked for help a number of times, but I'm afraid that neither Thomas, nor me, nor anybody else in the community has sufficient information/insight to describe the syntax that Tehhowch is looking for. I hope that Tehhowch can post an example or a draft sometime soon of what he is looking for so that we can update the RFC with his inputs and proceed with the formations.

Zitchas:
Thanks for the update!

For what it is worth, I'm told this has been implemented over in Endless Sky Stories; Not sure if they've made any changes to it or not.

Could you explain a bit about what you mean by 
> This PR and Thomas' editor are currently blocked on the syntax in the formation definitions
?

Do the two not use the same terms or otherwise not line up properly?

petervdmeer:
> For what it is worth, I'm told this has been implemented over in Endless Sky Stories; Not sure if they've made any changes to it or not.

I know; I ported it myself and Kiko gave me permissions to also merge the code by myself.

> Could you explain a bit about what you mean by
> 
> > This PR and Thomas' editor are currently blocked on the syntax in the formation definitions
> > ?
> 
> Do the two not use the same terms or otherwise not line up properly?

The terms to be used are defined in the RFC, but the RFC got updated over time; for example the older version of the RFC (used by Thomas and by ES Stories) defined formation-lines using a start-point, a direction and a length, while a newer version of the RFC (used for this PR) defines lines using a start-point and an end-point (because review-feedback indicated that that is easier to work with).

We can solve this by updating all implementations to follow the latest PR, but such updates are quite a bit of work and it looks like the RFC will still need to change significantly, so it makes sense for Thomas and me to avoid double work by waiting until there is an accepted RFC before doing the updates.

Zitchas:
Ah. That makes sense. Thanks for the explanation! Hopefully things change so they can move forward soon. And here I thought we were into the "tweak and merge" stage.

petervdmeer:
Resolved merge-conflict on code-blocks file.

Zitchas:
Just some thoughts regarding putting effort into making it easy to create new patterns, reposted from discord:

> That being said, I know making it easy for content-creators to add new formations was listed as being a big stumbling block, but the more I think about it, the more I think that should be removed from the list of considerations. 
> 
> Honestly, most people have trouble mathematically specifying patterns. There's a reason why number sequences and basic patterns are common logic puzzles that people spend lots of time figuring out. 
> 
> Personally, I think 95% of all content-creators will almost certainly be more than happy with the current provided list of formations and never need to make a new one.

I think it falls into the same category as the map - most people should be using an editor or tool of some kind to do the math and heavy lifting for them. We don't need the intricacies to be understandable to everyone.

On another note, I am currently trying to add formations to generic fleets, and it doesn't seem to be working. (Player fleet formations work, as does the job taking formation flyers to Earth). Here's the start of a fleet definition showing what I am currently trying:

```
fleet "Small Core Merchants"
	government "Merchant"
	names "civilian"
	formation "Vee"
	cargo 3
	personality
		confusion 30
		timid frugal appeasing
	variant 40
		"Shuttle"
	variant 60
		"Star Barge (Armed)"
```

petervdmeer:
> On another note, I am currently trying to add formations to generic fleets, and it doesn't seem to be working. (Player fleet formations work, as does the job taking formation flyers to Earth).

The formations were working fine for ships that were idle. NPCs are never idle however (except when they follow the player). I just added a commit to also let ships move into formation when they are landing and it seems to work nicely with arrival distances.

Zitchas:
Yes, I transferred over your updates, and now I do see groups of ships forming up now and then. Which is probably ideal - Unless they were randomly picking fleets, I wouldn't want them to always be in a recognizeable formation. That'd get old too.

That being said, having ships taking off have a random 0-1 second "pause" might be interesting just so that they sometimes *do* get into formation as they leave. (so long as no hostiles are present, of course)

petervdmeer:
> Yes, I transferred over your updates, and now I do see groups of ships forming up now and then. Which is probably ideal - Unless they were randomly picking fleets, I wouldn't want them to always be in a recognizeable formation. That'd get old too.
> 
> That being said, having ships taking off have a random 0-1 second "pause" might be interesting just so that they sometimes _do_ get into formation as they leave. (so long as no hostiles are present, of course)

Good suggestion! Sounds very related to the linger-behaviors discussed in #3381 and the yes/no hostiles check from #5583 .

For the future we could also consider to have ship take formation positions just before jumping and we could also consider to move ships towards formation positions during a jump. But I feel that both of those should be kept for a follow-up PR for when formation flying itself gets merged.

ravenshining:
Just want to say great work, I'm looking forward to seeing it implemented!  Really been enjoying this in ESS.  It's a real blast, particularly how it makes merchants seem more human as they take the time to form up before leaving a system, or how terrifying it is when a group of potent enemies come at you in formation.

Zitchas:
Also, just for the record, this has been implemented in #5772 .(Combat Revamp) A fair number of people have enjoyed it there, although they may not necessarily have known it is from this particular PR.

Zitchas:
Given how big an improvement this PR adds to the game, I'm of the opinion that this should be merged as-is. If someone wants to refactor it or something later on they still can, right? In the meantime, this is solid and a massive improvement to the gameworld, and has been a major item that people have commented favorably on in both Revamp and ESS (Endless Sky Stories - an ES fork).

That being said, something in the recent game engine changes broke the code in the Combat Revamp PR. As I suspect I will be recreating the Revamp, would it be possible for this Formations PR to be brought up-to-date with Master to facilitate merging it into Revamp 2?

petervdmeer:
I brought this PR up-to-date with master, fixed some small issues, improved some documentation and added some TODOs with ideas how to make the code better.

@Zitchas: This code has been tested quite a bit (in Revamp and in ESS), the RFC also did go through quite some review rounds, but the code here didn't have a lot of reviews yet. We could ask @thomasballinger and/or @quyykk to do an extensive code-review of this PR to see if there are still issues in the code itself that need attention. (Thomas is very familiar with the syntax of formation patterns, Quyykk seems quite capable with c++ coding and proper algorithms and it seems unfair to expect tehhowch to do all the reviewing.)

Zitchas:
Thank you! And yes, that does sound good to have some others do reviews of the code.

source/FormationPattern.h
```diff
@@ -0,0 +1,204 @@
...
+		// Currently calculated Point.
+		Point currentPoint;
+		// Internal status variable;
+		bool atEnd = false;
+	};
+
+
+
+	// Returns the name of this pattern.
+	const std::string Name() const;
```
quyykk, line 97: `const &`?

quyykk (commented):
Just some comments on the code itself. I'll maybe do general architecture later.

tehhowch (commented):
Didn't look at the formation* classes, since that's where the syntax rfc had most sway.

source/FormationPositioner.h
```diff
@@ -0,0 +1,87 @@
...
+class Ship;
+
+
+
+// Represents an active formation for a set of spaceships. Assigns each ship
+// to a position (Point) in the formation.
+class FormationPositioner{
+public:
+	// Initializer based on the formation pattern to follow.
+	FormationPositioner(const Body *formationLead, const FormationPattern *pattern): formationLead(formationLead), pattern(pattern), direction(formationLead->Facing()) {}
```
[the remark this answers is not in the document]
petervdmeer: The inline constructors were also used in CollissionSet.h, EsUuid.h, MapPanel.h, Panel.h and Ship.h. I assumed that this was done for readability (to avoid constructors in the cpp files that are just empty), but I don't mind to move the empty body to a cpp file (which indeed seems to be a little bit more correct).

source/FormationPattern.h
```diff
@@ -0,0 +1,204 @@
...
+		// with some defaults for smaller ships.
+		unsigned int numberOfShips = 0;
+		double maxDiameter = 80;
+		double maxWidth = 80;
+		double maxHeight = 80;
+	};
+	
+	
+	// Iterator that provides sequential access to all formation positions.
+	// TODO: finish the implementation and start using this.
```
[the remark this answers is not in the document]
petervdmeer: I finished the implementation for the iterator yesterday evening (my timezone) and wrote some tests for it, but the iterator is not used in the game yet. This iterator should be used in `FormationPositioner::NextPosition(...)` (and replace a lot of code there).

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
 // Fleet commands from the player.
+void AI::IssueFormationChange(const PlayerInfo &player)
+{
+	// Figure out what ships we are giving orders to
+	vector<Ship *> ships;
+	bool fullFleet = player.SelectedShips().empty();
+	if(fullFleet)
+	{
+		for(const shared_ptr<Ship> &it : player.Ships())
+			if(it.get() != player.Flagship() && !it->IsParked()
```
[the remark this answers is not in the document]
petervdmeer: Formations are also applied to ships that are carried (and they have by default their carrier as parent, not the players flagship).

The "Get currently selected ships or whole fleet" functions are present in this code multiple times, but each selection seems to be a little bit different;
- the loop for `cloaking` has an additional check on "can cloak"
- the `AI::IssueOrders` loop doesn't seem to check for destroyed (to include destroyed ships for clearing orders)
- the `IssueDeploy` loop has an additional check to only select carried ships.

The only 3 functions that have the same selection criteria are the IssueFormation* commands ('PatternSelect', 'RingIncrease' and `RingDecrease`). I could write a common function for those 3 or I could try to create a more generic function that also gets a lambda function parameter to allow additional criteria to be set on such generic function.
If we go for the generic function, then I might want to keep that for a separate PR (because it has the potential to increase the likelyhood of merge conflicts).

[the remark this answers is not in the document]
quyykk: Ok makes sense, thanks for clarifying.

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+	// First check which and how many formations we have in the current set of selected ships.
+	const FormationPattern *toSet = nullptr;
+	bool moreFound = false;
+	for(Ship *ship : ships)
+	{
+		const FormationPattern *shipsPattern = ship->GetFormationPattern();
+		if(shipsPattern)
+		{
+			if(!toSet)
+				toSet = formationPatterns.Find(shipsPattern->Name());
```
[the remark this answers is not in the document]
petervdmeer: The trick is in the `else` part just below it; if all selected ships have exactly the same formationPattern, then what will happen is that the IssueFormationChange() will change all ships to use the next formationPattern that is available. If however the ships have different formationPatterns, then the command will change all ships to use the very first formationPattern that was found in the list.

I will add a comment explaining this behavior.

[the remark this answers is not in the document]
quyykk: I don't get it, sorry. I understood your comment, but I don't see why you can't use `toSet = shipsPattern;`. You're basically searching for a pattern with the name `shipsPattern->Name()`: But that pattern is exactly `shipsPattern`. Or is it not?

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+		// If no formation was set at all, then we set the first one from the set of formations.
+		auto it = formationPatterns.begin();
+		if(it != formationPatterns.end())
+			toSet = formationPatterns.Find(it->first);
+	}
+	else if(!moreFound)
+	{
+		// If only one formation was found, then select the next formation (or clear the formation if there is no next).
+		auto it = formationPatterns.begin();
+		while(it != formationPatterns.end() && it->first != toSet->Name())
```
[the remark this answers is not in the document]
petervdmeer: The `Find` from Set.h doesn't return an iterator (and I don't need a pointer to the current formation here, but a pointer to the one after it). The `find` from the map below Set.h is not exposed, but it is probably a good idea to change Set.h to expose the `find` from the map below it.

EndlessSky.xcodeproj/project.pbxproj
```diff
@@ -19,6 +19,9 @@
 		62A405BA1D47DA4D0054F6A0 /* FogShader.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 62A405B81D47DA4D0054F6A0 /* FogShader.cpp */; };
 		62C3111A1CE172D000409D91 /* Flotsam.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 62C311181CE172D000409D91 /* Flotsam.cpp */; };
 		6A5716331E25BE6F00585EB2 /* CollisionSet.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6A5716311E25BE6F00585EB2 /* CollisionSet.cpp */; };
+		938E07E4235970CE00DC2C2B /* FormationPattern.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 938E07E0235970CE00DC2C2B /* FormationPattern.cpp */; };
+		938E07E5235970CE00DC2C2B /* FormationPositioner.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 938E07E2235970CE00DC2C2B /* FormationPositioner.cpp */; };
+		9E1F4BF78F9E1FC4C96F76B5 /* Test.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 2E8047A8987DD8EC99FF8E2E /* Test.cpp */; };
```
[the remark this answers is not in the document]
petervdmeer: Already fixed in a follow-up commit; looks like this was fixed at the same time as when you were reviewing.

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
 
 // Fleet commands from the player.
+void AI::IssueFormationChange(const PlayerInfo &player)
+{
+	// Figure out what ships we are giving orders to
+	vector<Ship *> ships;
+	bool fullFleet = player.SelectedShips().empty();
+	if(fullFleet)
+	{
+		for(const shared_ptr<Ship> &it : player.Ships())
```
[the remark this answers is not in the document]
petervdmeer: I'll reserve `player.Ships().size() - 1` for the full fleet case and `selectedShips.size()` for the case when there are selected ships. This would potentially reserve more space than we finally need (due to destroyed ships and parked ships), but I expect that performance would still be better than without reserve.

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+	// Now determine what formation to set.
+	// If more than one formation was found in the set, then we set it to the first one found.
+	if(!toSet)
+	{
+		// If no formation was set at all, then we set the first one from the set of formations.
+		auto it = formationPatterns.begin();
+		if(it != formationPatterns.end())
+			toSet = formationPatterns.Find(it->first);
+	}
+	else if(!moreFound)
```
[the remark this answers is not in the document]
petervdmeer: I feel `else if` is correct here, there are actually 3 options:
- toSet is nullptr (no existing pattern was found at all on all the ships)
- moreFound is false (only one pattern was found on all ships)
- moreFound is true (more than one pattern was found on all ships)

If moreFound is true, then what should happen is that all ships get the same first-found pattern. This option doesn't need any code here, since all variables are already set to just apply the pattern change. I will add some comments to show this 3rd option more explicitly.

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+	
+	// Now determine what formation to set.
+	// If more than one formation was found in the set, then we set it to the first one found.
+	if(!toSet)
+	{
+		// If no formation was set at all, then we set the first one from the set of formations.
+		auto it = formationPatterns.begin();
+		if(it != formationPatterns.end())
+			toSet = formationPatterns.Find(it->first);
+	}
```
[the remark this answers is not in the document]
petervdmeer: Good point. I will add an earlier check that ensures bail out with a proper message.

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+	const FormationPattern *toSet = nullptr;
+	bool moreFound = false;
+	for(Ship *ship : ships)
+	{
+		const FormationPattern *shipsPattern = ship->GetFormationPattern();
+		if(shipsPattern)
+		{
+			if(!toSet)
+				toSet = formationPatterns.Find(shipsPattern->Name());
+			else if(toSet->Name() != shipsPattern->Name())
```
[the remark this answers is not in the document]
petervdmeer: Pointer equality should be fine here (there should be no duplicates of a specific FormationPattern anywhere).

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+			Messages::Add("No ships selected that can change formation.", Messages::Importance::High);
+		
+		return;
+	}
+	
+	const Set<FormationPattern> &formationPatterns = GameData::Formations();
+	
+	// First check which and how many formations we have in the current set of selected ships.
+	const FormationPattern *toSet = nullptr;
+	bool moreFound = false;
```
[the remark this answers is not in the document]
petervdmeer: Will change to `multiplePatternsSet`.

source/AI.cpp
```diff
@@ -1366,6 +1576,50 @@ bool AI::FollowOrders(Ship &ship, Command &command) const
...
+{
+	shared_ptr<Ship> parent = ship.GetParent();
+	if(!parent)
+		return;
+	
+	const Body *formationLead = parent.get();
+	const FormationPattern *pattern = ship.GetFormationPattern();
+	
+	// First we retrieve the patterns available for the parent.
+	std::map<const FormationPattern *, FormationPositioner> * patterns = &formations[formationLead];
```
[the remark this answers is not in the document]
petervdmeer: Will change to reference, that is indeed better fitting here.

source/Ship.cpp
```diff
@@ -951,6 +954,9 @@ void Ship::Save(DataWriter &out) const
 				out.Write("final explode", it.first->Name(), it.second);
 		});
 		
+		if(formationPattern)
+			out.Write("formation", formationPattern->Name());
```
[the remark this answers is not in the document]
petervdmeer: Are you referring to the situation where a ship refers to some formation-pattern which is not defined in the datafiles?

I still want to add some tests for this, but my impression is that this already gets handled well because an empty formation-pattern will be created upon load.
The only thing that would currently be missing is an warning/error-message about an empty formation being present.

source/AI.cpp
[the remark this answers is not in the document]
petervdmeer: I will add the common function for the 3 formation functions with name `GetShipsForFormationCommand` (in the anonymous namespace of AI).

source/AI.cpp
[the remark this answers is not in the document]
petervdmeer: Sorry, I misunderstood your comment. Yes, I'll change this to `toSet = shipsPattern`.

source/AI.cpp
```diff
@@ -309,6 +312,199 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+	}
+	else if(!moreFound)
+	{
+		// If only one formation was found, then select the next formation (or clear the formation if there is no next).
+		auto it = formationPatterns.begin();
+		while(it != formationPatterns.end() && it->first != toSet->Name())
+			++it;
+		if(it == formationPatterns.end())
+			toSet = nullptr;
+		else if(it->first == toSet->Name())
```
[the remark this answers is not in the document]
petervdmeer: I'll check if I can expose the map::find(...) through Set.h instead (and just use the map::find directly).

petervdmeer:
@tehhowch , @quyykk  Thanks for your good reviews. I uploaded a rework commit and briefly retested the formations interface as used by the player.

source/Ship.cpp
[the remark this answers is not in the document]
petervdmeer: I looked at this a bit further and added a check in GameData::CheckReferences for formations where the Name is empty (formations that are references, but not defined).

I also added a check in FormationPattern::Load() to check for formations that are defined multiple times. (The formation pattern could look quite weird if two patterns are loaded on top of eachother.)

source/FormationPattern.h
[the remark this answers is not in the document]
petervdmeer: I just changed the code to actively use this iterator and removed this TODO.

source/AI.cpp
```diff
@@ -1366,6 +1548,44 @@ bool AI::FollowOrders(Ship &ship, Command &command) const
...
+	const Body *formationLead = parent.get();
+	const FormationPattern *pattern = ship.GetFormationPattern();
+	
+	// First we retrieve the patterns available for the parent.
+	auto &patterns = formations[formationLead];
+	
+	// Find the existing FormationPositioner for the pattern, or add one if none exists yet.
+	auto insert = patterns.emplace(piecewise_construct,
+		forward_as_tuple(pattern),
+		forward_as_tuple(formationLead, pattern));
```
quyykk, line 1566: If I'm not missing something: Can this be moved into the formation change function? It seems like we can emplace all of the patterns there already.

This is the only reason this function isn't `const`, so it would be great if this is possible :)

source/FormationPattern.cpp
```diff
@@ -0,0 +1,508 @@
...
+		{
+			// If we checked all lines and repeat sections, then go for the next ring.
+			++(ring);
+			line = 0;
+			repeat = 0;
+			slot = 0;
+			lineRepeatSlots = pattern.get().Slots(ring, line, repeat);
+			
+			// If we scanned more than 5 rings without finding a slot, then we have an empty pattern.
+			++ringsScanned;
```
quyykk, line 132: Might as well be consistent: Missing parens here and in some other places too.

source/FormationPattern.cpp
```diff
@@ -0,0 +1,508 @@
...
+		name = node.Token(1);
+	else
+	{
+		node.PrintTrace("Skipping load of unnamed formation-pattern:");
+		return;
+	}
+	
+	for(const DataNode &child : node)
+		if(child.Token(0) == "flippable" && child.Size() >= 2)
+			for(int i = 1; i<child.Size(); ++i)
```
quyykk, line 190: `i < child.Size();`

source/FormationPattern.cpp
```diff
@@ -0,0 +1,508 @@
...
+		endOrAnchorPx = startPx;
+		startPx = tmpPx;
+	}
+	
+	// Calculate the step from each slot between start and end.
+	Point slotPx = endOrAnchorPx - startPx;
+	
+	// Divide by slots, but don't count the first (since it is at position 0, not at position 1).
+	if(slots > 1)
+		slotPx = slotPx / static_cast<double>(slots - 1);
```
quyykk, line 417: `slotPx /= slots - 1;`

source/FormationPattern.cpp
```diff
@@ -0,0 +1,508 @@
...
+			if(ring % 2 && repeat->alternating && slots > 0)
+			{
+				startAngle += endAngle;
+				endAngle = -endAngle;
+			}
+		}
+		
+		// Apply slots to get the correct slot-angle.
+		if(slots > 1)
+			endAngle = endAngle / static_cast<double>(slots - 1);
```
quyykk, line 388: `endAngle /= slots - 1;`

source/FormationPattern.h
```diff
@@ -0,0 +1,220 @@
...
+	};
+	
+	
+	// Iterator that provides sequential access to all formation positions.
+	class PositionIterator : public std::iterator<
+		std::input_iterator_tag, // iterator_category
+		Point,                   // iterator: value_type
+		std::ptrdiff_t,          // iterator: difference_type
+		const Point*,            // iterator: pointer
+		Point& >                 // iterator: reference
```
quyykk, line 66: `&>`

source/FormationPattern.h
```diff
@@ -0,0 +1,220 @@
...
+	{
+	public:
+		PositionIterator(const FormationPattern &pattern, const ActiveFormation &af,
+			unsigned int startRing = 0, unsigned int shipsToPlace = 0);
+		
+		// A subset of the default input_iterator operations. Limiting to
+		// only a subset, since not all operations are used in-game.
+		const Point *operator->();
+		const Point &operator*();
+		PositionIterator& operator++();
```
quyykk, line 76: ` &operator++`

source/FormationPattern.cpp
petervdmeer: The original code was a more complex and needed parens, I will remove them above since they don't add value here.

source/AI.cpp
petervdmeer: Line 1574 has `it->second.NextPosition(&ship)` which determines the position of the ship inside a formation. It  also is non-const.
The main difficulty is that the decission for a ship to fly in formation or to do something else is made inside `AI::MoveEscort()` (and that decission is made every Step()/frame). If a ship is flying in formation, then it needs to get a position in the pattern around the formation leader. If a ship then later decides to do something else (or becomes disabled), then the position in the pattern needs to be freed up for other ships to use.

If we can deduce the ships' decission outside the `AI::MoveEscort()` function, then we can handle the joining and leaving of formations there and handle the formation-positions as `const` inside the Move-functions. Another option would be to also give the formation-position-tracking information as parameter to MoveEsort, but that also doesn't look very nice.

petervdmeer: I just moved the iterators that calculate the positions to the 'Step()' function. This brings us closer to making the position retrieval a `const` method, since the iterators no longer need to be updated in the `MoveEscorts()`/`MoveInFormation()` functions.

petervdmeer:
Just did an update for the join/leave behavior; ships now take positions in the order in which they arrive. And if a ship gets destroyed, then the position gets backfilled from the end of the formation (instead of all later ships moving back a position).

Zitchas:
I don't know if it is still the case, but at one point the main obstacle to this being merged was a desire to have the formation specifications (as in, the code that defines the formation) being more "average user friendly."

If this is still the case, I'd just like to say that - as a content creator - I am more than happy with the array of formations we currently have, and I don't consider dumbing it down enough for me to make new formations important enough to delay merging. Actually, I don't really forsee the need to ever make a new formation, although I'm sure someone might.

As far as I'm concerned, having formations be just another part of the game engine that I can't change directly (as in, I need to talk to someone who has skills in this regard to do anything with) is perfectly fine by me. Not to mention someone made a great "proof of concept" formation simulator thingy to help with the process, which worked fairly well. That would put it into the same category as ship hardpoints more or less needing a webpage to help setup, which works well.

thomasballinger:
Re
> someone made a great "proof of concept" formation simulator thingy to help with the process, which worked fairly well.

If this lands I plan to make such a tool, this is just the kind of thing I'd hope to use https://github.com/thomasballinger/endless-sky-bindings for to make on web tool for creating formations. This formations PR was the original motivation for being able to write parser / formatter / simulation tools that can reuse game code, because that [original tool](https://observablehq.com/@ballingt-personal/review-of-formation-flying-rfc-for-endless-sky) reimplemented the logic in JavaScript instead of reusing the C++ code.

petervdmeer:
I just updated the Code::Blocks files since I noticed that that wasn't done yet for this PR. Hope that this also helps in getting this closer to merging.

Zitchas:
Well, as far as I'm concerned this is good to go. But I don't get to push that button. :)

Terin:
If 0.9.15 has a combination of the varied star sprites, parallax, and fleets in the wild flying around in formation? It'd be an _incredible_ jump in visual quality from 0.9.14. Here's hoping we can get it in!

MasterOfGrey:
My biggest question is what's the interface for getting your escorts to follow a formation?

shitwolfymakes:
I cannot wait to nae nae the Unfettered in a perfectly-spaced squadron of Hurricanes. Great work!

Zitchas:
The interface is fairly simple: You have a key to rotate through formations, and one key to tell them to make the formation spread out. Just like other escort commands, they apply to all escorts unless you have some selected, in which case it applies to just the escorts that are selected. I've got a few videos demonstrating this where I have multiple sets in delta formation passing over each other.

petervdmeer:
Merged master to resolve merge-conflict with data-move from GameData to UniverseObjects.

source/FormationPositioner.cpp
```diff
@@ -0,0 +1,216 @@
...
+		
+		while(fabs(deltaDir.Degrees() + symRot) < fabs(deltaDir.Degrees()))
+		{
+			desiredDir += Angle(symRot);
+			deltaDir = desiredDir - direction;
+		}
+	}
+	
+	// Angle at which to perform longitudinal or transverse mirror instead of turn.
+	static const double MIN_FLIP_TRIGGER = 135.;
```
quyykk, line 147: `constexpr`

source/FormationPositioner.cpp
```diff
@@ -0,0 +1,216 @@
...
+		if(pattern->FlippableY())
+			flippedY = !flippedY;
+		if(pattern->FlippableX())
+			flippedX = !flippedX;
+	}
+	else
+	{
+		// Turn max 1/4th degree per frame. The game runs at 60fps, so a turn of 180 degrees will take
+		// about 12 seconds.
+		static const double MAX_FORMATION_TURN = 0.25;
```
quyykk, line 164: `constexpr`

source/Set.h
```diff
@@ -36,6 +36,8 @@ class Set {
 	
 	typename std::map<std::string, Type>::iterator begin() { return data.begin(); }
 	typename std::map<std::string, Type>::const_iterator begin() const { return data.begin(); }
+	typename std::map<std::string, Type>::iterator find(std::string &key) { return data.find(key); }
```
quyykk, line 39: Not sure if should be done in this PR but since we're repeating the types here a lot we should create type alias. Either just for the map or also for the iterators.

Also the parameter should be taken as `const&`.

data/formations.txt
```diff
@@ -0,0 +1,649 @@
+# Copyright (c) 2019 by Peter van der Meer
```
quyykk, line 1: Shouldn't you update the date?

source/FormationPattern.h
```diff
@@ -0,0 +1,217 @@
...
+public:
+	// Struct that describes the properties of an active formation, like the maximum
+	// sizes of the ships participating in the formation and some data on the Body
+	// around which the formation is formed.
+	class ActiveFormation
+	{
+	public:
+		// Center radius of the formation that is to be kept clear. This
+		// is typically used to avoid positions of ships overlapping with
+		// the  body around which the formation is formed. The actual radius
```
quyykk, line 40: There's an extra space here.

source/UniverseObjects.cpp
```diff
@@ -276,6 +276,10 @@ void UniverseObjects::CheckReferences()
 	for(const auto &it : hazards)
 		if(!it.second.IsValid())
 			Warn("hazard", it.first);
+	// Formation patterns are not serialized, but their usage is.
+	for(auto &it : formations)
```
quyykk, line 280: `auto &&` since all the other loops use it to.

quyykk (commented):
Just a general code review. Feel free to ignore some review comments if you don't like them. Second part in the next few days :)

source/AI.cpp
```diff
@@ -469,6 +642,7 @@ void AI::Clean()
 {
 	actions.clear();
 	notoriety.clear();
+	formations.clear();
```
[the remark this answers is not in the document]
petervdmeer: I don't see the ordering that was used. How is the order determined?

data/formations.txt
petervdmeer: This file was indeed also changed in 2020.

source/AI.cpp
```diff
@@ -308,7 +349,126 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+{
+	// Figure out what ships we are giving orders to
+	vector<Ship *> targetShips = GetShipsForFormationCommand(player);
+	
+	// This should never happen, but just in case:
+	if(targetShips.empty())
+		return;
+	
+	const auto &formationPatterns = GameData::Formations();
+	if(formationPatterns.size() < 1)
```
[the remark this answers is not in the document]
petervdmeer: I'll add the `empty` function.

If a formation-pattern only contains invalid statements, then it would be loaded as an empty formation. The FormationPattern would then just return Point(0,0) for any ships requesting positions in such empty formation.
Is this the type of invalid pattern you were thinking of?

source/AI.cpp
```diff
@@ -308,7 +349,126 @@ AI::AI(const List<Ship> &ships, const List<Minable> &minables, const List<Flotsa
...
+	if(toSet)
+		Messages::Add(to_string(targetShips.size()) + " ships are now flying in " + toSet->Name() + " formation.", Messages::Importance::Low);
+	else
+		Messages::Add(to_string(targetShips.size()) + " ships are no longer flying in formation.", Messages::Importance::Low);
+}
+
+
+
+// Fleet commands from the player.
+void AI::IssueFormationRingDecrease(const PlayerInfo &player)
```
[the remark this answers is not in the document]
petervdmeer: Good idea. Those controls were based on some initial controls used for debugging and zeroing was more useful back then. But it looks like those controls are going to be official, and in that case we want the decrease function to just perform decrease.
I will also combine some code from the Decrease and Increase functions (since they will be almost identical).

source/AI.cpp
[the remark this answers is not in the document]
quyykk: Yeah.

source/AI.cpp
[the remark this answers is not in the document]
quyykk: It's the definition order inside the class and I think somewhere else too.

petervdmeer:
> Just a general code review. Feel free to ignore some review comments if you don't like them. Second part in the next few days :)

Thanks for the good review, looking forward for the second part. :)

EjoThims:
I also await the second part, in hopes this review gets this ready for merging.

petervdmeer:
I also briefly looked at the code myself and found some outdated comments and TODO items; some code now moved into separate functions and I did some performance optimization by only calculating the positions every 20 game-steps instead of every game-step when there are no urgent reasons to re-calculate the positions.

Zitchas:
@quyykk You mentioned a second part of a review?

Zitchas:
I found another really nice use case for formations: Comparing ship appearances while testing sprites for new content.

I'm currently comparing seven different ship appearances for upcoming content. In order to see how they look side by side in flight, I have to manually spread them out, and then summon them to me and hope I can get a good enough look at how they compare while in the few seconds before they clump up in a pile again. With formations, I just tell my fleet to form up, and then I can happily fly around comparing how they look as we go through turns and sweeps and manoeuvres together and they do their best to remain in position relative to my flagship. And as an added bonus, I don't have to keep stopping to manually reposition them every few seconds.

Zitchas:
Would it be possible to have some way to tighten up the formations some more? I just tried using 7 Bactrian sized ships in the "Leading Delta" formation, and the tightest ring I could get them into had something close to 1.5x their ship length between ships, so I had to zoom a long ways out to get them all on the screen. I know 1.5x separation is probably a good "real world" spacing, but it'd be nice (and combat relevant) to be able to pull that in to about .25x separation.

petervdmeer:
> Would it be possible to have some way to tighten up the formations some more? I just tried using 7 Bactrian sized ships in the "Leading Delta" formation, and the tightest ring I could get them into had something close to 1.5x their ship length between ships, so I had to zoom a long ways out to get them all on the screen. I know 1.5x separation is probably a good "real world" spacing, but it'd be nice (and combat relevant) to be able to pull that in to about .25x separation.

Yes, but I think it would be better to create a second set of formations instead; keep the current diameter-based formations and add an extra set width/height based formations where the ships are "ships-width + 10 pixels" apart. I also think I should put those in a separate PR. Would that work for you?

petervdmeer:
Closing this PR; it has served its purpose as test-environment and demonstrator of what we can do with formation flying.

There are still quite some improvements that can make the formations-flying feature even nicer than that it is in this PR, but I feel that it is appropriate to focus on getting the stable parts merged into the ES-main-branch first (with PRs like #6570 and #5583) and to spend some more attention to architecture and (auto)test improvements before adding even more functionality.

Thanks to all that contributed (in the discussions, feature requests and related activities like the online editor).

petervdmeer:
Re-opening as Draft;
- Todays master-branch is merged into this branch/PR
- This PR is briefly tested; it now compiles with CMake, and the game doesn't crash when I start a formation
- The ring functionaly is removed (because it wasn't working as well as I hoped)
  - Only the shift-G (gather) keybind now has meaning; the ring-control keybinds are out of the code.
  - I still need to remove some low-level ring-code
  - I expect the ring-functionality to be replaced by functions to specify which ships are allowed at which positions
- The code currently supports both "slots" and "positions"
  - "positions" is a much better keyword, so I plan to replace "slots" completely.
  - Also need to check what to do with the other keywords; the RFC also contains some other improvements.
- The formation-positions are still calculated every frame, they should be cached instead
  - The caching system should also efficiently handle "regular" and "flipped" positions.

Zitchas:
Awesome! Great to see this being worked on again. Looking forward to seeing it implemented.

Saugia:
@petervdmeer Any update on this PR? It's been a little while since it was re-opened as a draft in May. If there are no plans to work on it soon, we could close it for the time being.

petervdmeer:
> @petervdmeer Any update on this PR? It's been a little while since it was re-opened as a draft in May. If there are no plans to work on it soon, we could close it for the time being.

OK

petervdmeer:
Re-opened this PR to support @10039.

Zitchas:
Exciting to see action on this again! Thanks for working on it!

petervdmeer:
Removed automatic ship-size scaling support by merging #10039 into this PR. Ship-size scaling support is is quite a complex feature, and having formations in sizes "small" (for interceptors), "medium" (for warships) and "large" (for capital ships) probably works better, since it avoids the complex calculations in within the game. (At the costs of having those calculations done outside of ES when creating new formations.)

The formation-definitions in this pr still use automatic ship-size scaling support, so those definitions need to be rewritten to use pixel-sized coordinates only. (Polar coordinate support is also removed with this change.)

petervdmeer:
Merged master branch to resolve merge-conflicts with https://github.com/endless-sky/endless-sky/pull/10039

tibetiroka:
> Formations automatically resize based on the size of participating ships.

Is this still a part of this PR? I'm having a little trouble getting it to work.
Screenshot 2024-07-16_14:09:57
Screenshot 2024-07-16_14:10:16
Screenshot 2024-07-16_14:10:40
image


These ships are in the same (atomic) formation, at the same zoom level. As far as I can tell, their positions are the same, even if all ships are small fighters or KIVs.

Is there maybe so much space in the formation by default that scaling isn't necessary? If so, the current values might have to be adjusted.

data/formations_complex.txt
```diff
@@ -0,0 +1,235 @@
...
+		positions 3
+		repeat
+			start 150 0
+			end 0 259.8
+			positions 2
+
+
+# A sphere around the flagship
+# Built in six segment sections using the same layouting as the hexagon shape.
+formation "Sphere"
```
tibetiroka, line 179: Should this be Circle, considering the next one is Square?

data/formations_basic.txt
```diff
@@ -51,7 +51,6 @@ formation "Vic"
 
 
 # X shaped formation that rotates with the flagship.
-# For warship-sized escorts
```
tibetiroka, line 54: I couldn't find the scaling code in this PR. Maybe it got lost in the merge conflicts? If it's not planned for the PR, the comment should stay.

tibetiroka (commented):
The code looks good, except:
- The scaling code is missing
- There is something fishy going on with pattern rotation / mirroring (more details below).

As the PR is currently fairly small on the code side, I'm thinking maybe we could get the formation snapping functionality as well. Putting something like this at the end of AI::MoveInFormation (within the `inPosition` check) produced pretty nice results for me:
```
		// If the position and velocity matches, smoothly "snap" to position over multiple frames.
		Point positionDelta = it->second.Position(&ship) - ship.Position();
		Point snapMovement = positionDelta.LengthSquared() < 4. ? positionDelta : positionDelta.Unit() * 2;
		ship.SetPosition(ship.Position() + snapMovement);

		Point velocityDelta = formationLead->Velocity() - ship.Velocity();
		Point snapAcceleration = velocityDelta.LengthSquared() < 4. ? velocityDelta : velocityDelta.Unit() * 2;
		if((ship.Velocity() + snapAcceleration).LengthSquared() <= ship.MaxVelocity())
			ship.SetVelocity(ship.Velocity() + snapAcceleration);
```

-----
It seems to me that sometimes patterns are rotated/mirrored unnecessarily. I tried to capture it on video:
output2


Sometimes it rearranges the formation, and sometimes it doesn't. I don't really see the reason behind it.

(I understand that some of these suggestions are not directly related to the PR, and they could be separated if necessary. It's just really nice to be able to test everything at the same place.)

data/formations_complex.txt
petervdmeer, line 179:
```suggestion
formation "Circle"
```

data/formations_basic.txt
petervdmeer: The numbers below (spacing of 150) suggest to me that this formation is aimed at interceptor-sized escorts. So I removed this comment, since the spacing seems to small for warships.

petervdmeer: Scaling (using the MultiAxisPoint class) got taken out of this PR, because it was too complex.
I filed an issue (https://github.com/endless-sky/endless-sky/issues/10324) for a simpler, alternative, implementation.

tibetiroka:
I think Circle and Diamond also have the same flip issue.

petervdmeer:
>     * The scaling code is missing

It's not missing. It was taken out because the scaling in this PR was too complex. I created a feature-request for a simpler version of scaling; https://github.com/endless-sky/endless-sky/issues/10324

>     * There is something fishy going on with pattern rotation / mirroring (more details below).

This was a problem in the "Atomic" formation. It would never rotate, but it did have a keyword for flipping. I removed the flipping keyword.

> As the PR is currently fairly small on the code side, I'm thinking maybe we could get the formation snapping functionality as well.

I prefer not to add this to this PR, but please feel free to create a separate PR for this.

tibetiroka:
Doesn't the Atomic formation have 120 degress rotational symmetry? If so, I don't see why it shouldn't rotate.

data/formations_complex.txt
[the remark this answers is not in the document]
petervdmeer: Agreed, I merged them into 1 file.

petervdmeer:
> I think Circle and Diamond also have the same flip issue.

Should be fixed now.

data/formations.txt
```diff
@@ -0,0 +1,300 @@
...
+		repeat -150 150
+
+
+# Formation where ships fly in a diagonal line right behind the formation leader.
+formation "Echelon (right)"
+	position 150 150
+		repeat 150 150
+
+
+formation "Hexagon (wide)"
```
tibetiroka, line 126:
```suggestion
formation "Hexagon"
```

petervdmeer: I originally wanted to have 2 hexagon formations:
- A hexagon with points to the sides and flat sides front and after (relative to flagship)
- A hexagon with points to the front and back and flat sides on the sides (relative to flagship)

This hexagon is the first one, and thus post-fixed with "wide".
The second one would be post-fixed with "long".

Do you have suggestions for better names?

tibetiroka: It might just be easier to have a single stable formation. Having two formations that only differ in 60 degrees of rotation seems overkill.

petervdmeer: You're right, having two nearly identical formations is probably confusing.

tibetiroka (approved):

source/FormationPositioner.cpp
```diff
@@ -151,21 +155,25 @@ void FormationPositioner::CalculatePositions()
 
 void FormationPositioner::CalculateDirection()
 {
+	// Any direction is fine, just keep initial direction.
+	if(pattern->Rotatable() == 0.)
```
RisingLeaf, line 159:
```suggestion
	if(pattern->Rotatable() == 0)
```

tibetiroka: oh well

RisingLeaf: smh

RisingLeaf: ~~revert~~

PR #4472, opened 2019-08-13
NEW CLASS: GameWindow

AdamKauffman:
**NEW:** GameWindow Class. Contains all SDL window code that was in main.cpp
This resulted in changes to the order of operations to untangle SDL window code from other initialization tasks. The Window appears earlier in the loading process but otherwise appears the same to the user.

I modeled the class after the Audio class. Similar naming and use. One benefit is that initialization cruft that used to be in main has a smaller scope and can die when it's done, leaving only properties that we need to use later. 

_Curly's Law: Do One Thing_
To keep main short and sweet it initializes everything and then calls GameLoop, where the main loop lives. Reading main is now just reading a list of things. This makes 100% of real work done in modular bits. 

**CHANGE LOG:**
WAS: Normal VSYNC on.
NEW: Try to set "Adaptive VSync" and fail to normal VSync.

REMOVED: Manual tracking of fullscreen and maximize.
NEW: Return actual window state from SDL.

REMOVED: SDL_WINDOW_SHOWN flag 
BECAUSE: "SDL_WINDOW_SHOWN is ignored by SDL_CreateWindow(). The SDL_Window is implicitly shown if SDL_WINDOW_HIDDEN is not set. "

NEW: Explicitly ask for hardware acceleration. If it fails it will just use the software renderer.

DISABLED: Win32 exception for SDL_GL_DeleteContext. Everything suggests that this call needs to be left in. Let's see if this crash on quit has been resolved. I'm a Windows developer and I haven't seen the crash yet.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,3 +52,4 @@ ipch/
 *.userosscache
 *.sln.docstates
 *.xcodeproj/**
+EndlessSky.cscope_file_list
--- a/.gitignore
+++ b/.gitignore
@@ -52,4 +52,3 @@ ipch/
 *.userosscache
 *.sln.docstates
 *.xcodeproj/**
-EndlessSky.cscope_file_list
--- a/source/GameWindow.cpp
+++ b/source/GameWindow.cpp
@@ -27,6 +27,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
+	SDL_Window *mainWindow;
 	SDL_GLContext context;
 	int minWidth = 640;
 	int minHeight = 480;
@@ -34,8 +35,7 @@ namespace {
 	int height;
 	int monitorHz;
 	bool hasSwizzle;
-	SDL_Window *mainWindow;
-	
+		
 	// Logs SDL errors and returns 1 if found
 	int checkSDLerror()
 	{
@@ -101,26 +101,18 @@ bool GameWindow::HasSwizzle()
 
 int GameWindow::Init()
 {
+	// This needs to be called before any other SDL commands.
 	if (SDL_Init(SDL_INIT_VIDEO) != 0) 
 		return 1;
 	
 	// Start with a clear the error state
 	SDL_GetError();
 	
-	// Check how big the window can be.
+	// Get details about the current display.
 	SDL_DisplayMode mode;
 	if(SDL_GetCurrentDisplayMode(0, &mode))
 		return DoError("Unable to query monitor resolution!");
-	
-	monitorHz = mode.refresh_rate;
-	
-	Uint32 flags = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_SHOWN | SDL_WINDOW_ALLOW_HIGHDPI;
-
-	if(Preferences::Has("fullscreen"))
-		flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
-	else if(Preferences::Has("maximized"))
-		flags |= SDL_WINDOW_MAXIMIZED;
-	
+		
 	// Make the window just slightly smaller than the monitor resolution.
 	int maxWidth = mode.w;
 	int maxHeight = mode.h;
@@ -138,20 +130,36 @@ int GameWindow::Init()
 		windowHeight = min(windowHeight, Screen::RawHeight());
 	}
 	
-	// Create the window.
-	SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
-#ifdef _WIN32
-	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
-	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
-#endif
-	SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);
+	monitorHz = mode.refresh_rate;
 	
+	// Settings that must be declared before the window creation.
+	Uint32 flags = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI;
+
+	if(Preferences::Has("fullscreen"))
+		flags |= SDL_WINDOW_FULLSCREEN_DESKTOP;
+	else if(Preferences::Has("maximized"))
+		flags |= SDL_WINDOW_MAXIMIZED;
+	
+	// The main window spawns visibly at this point.
 	mainWindow = SDL_CreateWindow("Endless Sky", SDL_WINDOWPOS_UNDEFINED, 
 		SDL_WINDOWPOS_UNDEFINED, windowWidth, windowHeight, flags);
 		
 	if(!mainWindow)
 		return DoError("Unable to create window!");
 	
+	// Settings that must be declared before the context creation.
+	SDL_GL_SetAttribute(SDL_GL_DOUBLEBUFFER, 1);
+#ifdef _WIN32
+	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MAJOR_VERSION, 3);
+	SDL_GL_SetAttribute(SDL_GL_CONTEXT_MINOR_VERSION, 0);
+#endif
+	SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_CORE);	
+	SDL_GL_SetAttribute(SDL_GL_ACCELERATED_VISUAL, 1);
+	
+	// FSAA?
+	//SDL_GL_SetAttribute(SDL_GL_MULTISAMPLEBUFFERS, 1);
+	//SDL_GL_SetAttribute(SDL_GL_MULTISAMPLESAMPLES, 2);
+		
 	context = SDL_GL_CreateContext(mainWindow);
 	if(!context)
 		return DoError("Unable to create OpenGL context! Check if your system supports OpenGL 3.0.");
@@ -188,8 +196,6 @@ int GameWindow::Init()
 		return DoError(out.str());
 	}
 	
-	checkSDLerror();
-	
 	// OpenGL settings
 	glClearColor(0.f, 0.f, 0.0f, 1.f);
 	glEnable(GL_BLEND);
@@ -206,15 +212,14 @@ int GameWindow::Init()
 	// Make sure the screen size and view-port are set correctly.
 	AdjustViewport();
 	
+	string swizzleName = "_texture_swizzle";
+	
 #ifndef __APPLE__
 	// On OS X, setting the window icon will cause that same icon to be used
 	// in the dock and the application switcher. That's not something we
 	// want, because the ".icns" icon that is used automatically is prettier.
 	SetIcon();
-#endif
-
-string swizzleName = "_texture_swizzle";		
-#ifndef __APPLE__
+	
 	const char *extensions = reinterpret_cast<const char *>(glGetString(GL_EXTENSIONS));
 	hasSwizzle = strstr(extensions, swizzleName.c_str());
 #else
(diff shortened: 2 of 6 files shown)


Fzzr:
The build of this branch fails on macOS 10.14.6, Xcode 10.3. The errors are as follows:

```
Undefined symbols for architecture x86_64:
  "GameWindow::HasSwizzle()", referenced from:
      GameLoop(PlayerInfo&, Conversation&, bool&) in main.o
  "GameWindow::IsMaximized()", referenced from:
      _main in main.o
  "GameWindow::IsFullscreen()", referenced from:
      _main in main.o
      GameLoop(PlayerInfo&, Conversation&, bool&) in main.o
  "GameWindow::AdjustViewport()", referenced from:
      GameLoop(PlayerInfo&, Conversation&, bool&) in main.o
  "GameWindow::ToggleFullscreen()", referenced from:
      GameLoop(PlayerInfo&, Conversation&, bool&) in main.o
  "GameWindow::Init()", referenced from:
      _main in main.o
  "GameWindow::Quit()", referenced from:
      _main in main.o
  "GameWindow::Step()", referenced from:
      _main in main.o
      GameLoop(PlayerInfo&, Conversation&, bool&) in main.o
  "GameWindow::Width()", referenced from:
      _main in main.o
  "GameWindow::Height()", referenced from:
      _main in main.o
  "GameWindow::DoError(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)", referenced from:
      _main in main.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

AdamKauffman:
Drat, and me without a Mac. Thanks for testing @Fzzr . At first sight that appears to be the entire "GameWindow" class missing. I'm not familiar with XCode or clang. 

I can see that the project files are listed in a file found at: `endless-sky\EndlessSky.xcodeproj\project.pbxproj`

I cannot simply add a file to this project file because it includes some nasty identifiers that I'm not sure how to fake without xcode. Can you add the files "GameWindow.h" and "GameWindow.cpp" to your XCode project and then try again? If that works I need your EndlessSky.xcodeproj file to add to this PR.

Fzzr:
Go ahead and merge my branch https://github.com/Fzzr/endless-sky/tree/InTheBeginning to get a version with the project file updated. It now builds, and runs in a quick sanity test at least. I also recommend bringing your branch up to date with `master`.

AdamKauffman:
Thanks to @Fzzr I am now rebased on to current master with the new XCode project file in place.

.gitignore
```diff
@@ -51,3 +51,4 @@ ipch/
 *.userosscache
 *.sln.docstates
 *.xcodeproj/**
+EndlessSky.cscope_file_list
```
Pointedstick: Unrelated change

AdamKauffman: It's going to be in every push I make because I don't want to upload that file. It should be ignored.

Fzzr: Just don't include the changed `.gitignore` in the commit. You should only be commiting changes you want others to see. In this case what you probably want to do is add that path to your global `.gitignore`, which is in your homedir by default.

MCOfficer: (in fact, you can just gitignore the gitgnore)

realityforge: @AdamKauffman the easiest way is to just ignore them locally - i.e. patch your file `.git/info/exclude` to ignore them. For more details, both GitHub itself and stack overflow (https://stackoverflow.com/questions/1753070/how-do-i-configure-git-to-ignore-some-files-locally) have some good resources

source/GameWindow.cpp
```diff
@@ -0,0 +1,363 @@
...
+namespace {
+	SDL_Window *mainWindow;
+	SDL_GLContext context;
+	int width = 0;
+	int height = 0;
+	//int monitorHz;
+	bool hasSwizzle;
+		
+	// Logs SDL errors and returns 1 if found
+	int checkSDLerror()
```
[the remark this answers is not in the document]
AdamKauffman: Agreed. It is bool.

source/GameWindow.cpp
```diff
@@ -0,0 +1,363 @@
...
+	{
+		const char *sdlMessage = SDL_GetError();
+		if(sdlMessage && sdlMessage[0])
```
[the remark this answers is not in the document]
AdamKauffman: Thanks! I dropped this patch in.

source/GameWindow.cpp
```diff
@@ -0,0 +1,363 @@
...
+		{
+			string message;
+			message = " (SDL message: \"";
```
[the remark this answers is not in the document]
AdamKauffman: agreed

source/GameWindow.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+
+int GameWindow::Init()
+{
+	// This needs to be called before any other SDL commands.
+	if (SDL_Init(SDL_INIT_VIDEO) != 0) 
+		return 1;
+	
+	// Start with a clear error state
+	SDL_GetError();
```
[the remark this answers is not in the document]
AdamKauffman: Thanks. I added SDL_ClearError to my check so it calls get and clear just like the example shown at https://wiki.libsdl.org/SDL_ClearError

source/GameWindow.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+	if(mainWindow)
+		SDL_DestroyWindow(mainWindow);
+		
+	SDL_Quit();
+}	
+
+
+
+int GameWindow::DoError(const string& message)
```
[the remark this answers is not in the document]
AdamKauffman: I made DoError void. Then I made Init boolean because we are only using one exit status of 1.

source/GameWindow.cpp
```diff
@@ -0,0 +1,363 @@
...
+	int checkSDLerror()
+	{
+		const char *sdlMessage = SDL_GetError();
+		if(sdlMessage && sdlMessage[0])
+		{
+			string message;
+			message = " (SDL message: \"";
+			message += sdlMessage;
+			message += "\")";			
+			Files::LogError(message);
```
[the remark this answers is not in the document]
AdamKauffman: This is the function of GameWindow::DoError, which in turn calls checkSDLerror after reporting the error it was passed. So the SDL error is optionally added if there is one in the buffer. The checkSDLerror is only called by itself when there is no other message to pass. If you have a contextual message you pass it to DoError.

source/GameWindow.cpp
```diff
@@ -0,0 +1,375 @@
...
+	bool hasSwizzle;
+		
+	// Logs SDL errors and returns 1 if found
+	bool checkSDLerror(const string &context)
+	{
+		string message = SDL_GetError();
+		if(!message.empty())
+		{
+			Files::LogError("(SDL message: \"" + message + "\")");
+			SDL_ClearError
```
tehhowch: Did you mean to perform a function call here (and in several other places)?

AdamKauffman: Wow. I really borked it up good with that last patch. Somehow I thought I had tested this but I must have run the previous build. I resolved all errors and warnings. Double checked the build.

source/GameWindow.cpp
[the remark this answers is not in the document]
AdamKauffman: Does that make sense? I'm not sure what you want to do with the context.

source/main.cpp
```diff
@@ -91,313 +85,210 @@ int main(int argc, char *argv[])
...
+			// The caps lock key slows the game down (to make it easier to
+			// see and debug things that are happening quickly).
+			if(debugMode && event.type == SDL_KEYDOWN && event.key.keysym.sym == SDLK_BACKQUOTE)
+			{
+				isPaused = !isPaused;
+			}
+			else if(event.type == SDL_KEYDOWN && menuPanels.IsEmpty()
+					&& Command(event.key.keysym.sym).Has(Command::MENU)
+					&& !gamePanels.IsEmpty() && gamePanels.Top()->IsInterruptible())
+			{   // User pressed the Menu key
```
tehhowch: Comments should be on their own lines, and include full stops.

source/main.cpp
```diff
@@ -91,313 +85,210 @@ int main(int argc, char *argv[])
...
-			out << "Unable to query the GLSL version. OpenGL version is " << glVersion << ".";
-			return DoError(out.str(), window, context);
+			UI &activeUI = (menuPanels.IsEmpty() ? gamePanels : menuPanels);
+			
+			// If the mouse moves, reset the cursor movement timeout.
+			if(event.type == SDL_MOUSEMOTION)
+				cursorTime = 0;
+			
+			// The caps lock key slows the game down (to make it easier to
+			// see and debug things that are happening quickly).
```
tehhowch: I realize this was another comment that you didn't write, and just moved, but it no longer applies to the block of code it introduces.

source/main.cpp
```diff
@@ -91,313 +85,210 @@ int main(int argc, char *argv[])
...
 	}
 	
-	return 0;
+	// If player quit while landed on a planet, save the game if there are changes.
+	if(player.GetPlanet() && gamePanels.CanSave())
+		player.Save();
 }
 
 
-
```
tehhowch: Should be 3 lines between functions

source/main.cpp
```diff
@@ -573,18 +358,20 @@ Conversation LoadConversation()
 #ifdef _WIN32
 void InitConsole()
 {
-	// If both stdout and stderr are already initialized (e.g. writing to a file), do nothing.
 	const int UNINITIALIZED = -2;
 	bool redirectStdout = _fileno(stdout) == UNINITIALIZED;
 	bool redirectStderr = _fileno(stderr) == UNINITIALIZED;
 	bool redirectStdin = _fileno(stdin) == UNINITIALIZED;
-
+
```
tehhowch: Looks like some whitespace got eaten in this method.

source/main.cpp
AdamKauffman: FIXED

source/main.cpp
AdamKauffman: Removed

source/main.cpp
AdamKauffman: Added line

source/main.cpp
AdamKauffman: Normalized

AdamKauffman:
> This PR introduces a regression upon entering fullscreen - when exiting fullscreen, the display window has the "maximized" size, rather than the size it had before entering fullscreen.


I see what I did. I made it track the window size every time it was changed. I made it instead have a `if(!GameWindow::IsFullscreen())` like the original source did so that it doesn't save the window size while fullscreen. I tested the change and it appears to be working correctly now.

tehhowch:
The latest change to [main.cpp](https://github.com/endless-sky/endless-sky/commit/6c8e3e132a13d03b7997a808b0eb7171afa3a290#diff-bfd574051572855ffede390ec872fc68) generated conflicts you'll need to resolve.

What's the plan for the commented code? If it won't be used, remove it.

AdamKauffman:
I added the fast forward code to my branch and then rebased on upstream/master. 
Conflict is resolved. 
I tested that fast forward works.

tehhowch:
Your rebase was incorrectly done:

image


When in doubt, please use a standard merge.

AdamKauffman:
Can you explain what was incorrect and what the benefit of a merge is? I have read that I should always rebase. Is that not how it's done on this project? Or is there a way to rebase correctly?

MCOfficer:
> 
> 
> Can you explain what was incorrect and what the benefit of a merge is? I have read that I should always rebase. Is that not how it's done on this project? Or is there a way to rebase correctly?

It's hard to tell what exactly went wrong (git works in mysterious ways), but you have a lot of changes in datafiles that you likely didn't intend to be part of the PR.

In general, rebasing is much more cleaner, but can also give you harder-to-resolve conflicts. If you can, rebase, if you cannot or aren't sure, merge. It won't be easy to resolve this now. You can try an interactive rebase, dropping all but your own commits, to get your pre-rebase index back. Then perform a normal merge. If you have any uncommitted changes, make use of the stash to persist them across the operation. You also might want to create "backup branches" beforehand, so you have an easy hard-reset in case things go wrong.

PR #4474, opened 2019-08-13
Fix ordering of hardpoints

Darcman99:
@beccabunny noticed that they disordered many hardpoint whilst they had their new Southbound shipyard sprites merged. They asked whoever was available to help fix it, so here I am.

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -108,16 +108,16 @@ ship "Argosy"
 		"Greyhound Plasma Steering"
 		"Hyperdrive"
 		
+	engine -25 91 0.6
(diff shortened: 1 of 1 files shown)


Amazinite:
While you're at it, make sure that the hardpoints are ordered front to back, left to right. See: #2856. 
Same goes with @beccabunny in the Betelgeuse ship PR.

Darcman99:
Just updated the PR with your feedback @Amazinite.
I ordered each group (engines/guns/turrets) by lowest second value to highest second value, then between the hardpoints with identical secondary values, I ordered it lowest to highest values again.

PR #4476, opened 2019-08-13
Fix "Wanderers Evacuation" to occur in order

SolraBizna:
If you land on Vara K'chrai ahead of the transports in `Wanderers Evacuation 1B`, you get the message to wait for the transports and then `Wanderers Evacuation 1C` starts anyway. I think this prevents that, while still preserving the illusion of a single multi-stop mission.

Unfortunately, when trying to test it, I screwed up juggling my snapshots and now have no save from the correct part of the plot.


Amazinite:
Good catch.

PR #4479, opened 2019-08-15
add the "south pirate" attribute to Bloodsea

Kryes-Omega:
Working on missions involving specific regions, turns out Bloodsea is the only southern pirate world without the `"south pirate"` attribute for some reason. Now it is not. Small fix.


Kryes-Omega:
anything else I missed?

Amazinite:
That should be it assuming all the other pirate worlds have the proper attributes.

Issue #4436, opened 2019-07-17
Show the orange "available job" tag when active mission list is large


alextd:
**Is your feature request related to a problem? Please describe.**
There are a max of 12 mission tags per system on the map.

When you have 12+ active missions for a system, only 12 tags are shown.

When I am looking for new missions, I would prefer to add new missions in that system.

But that system doesn't show any orange tags because the 12/12 tags are blue and I don't see a mission to start, but there is one in the list to do.

**Describe the solution you'd like**
Simply show the available mission tags first, taking preference over active missions.

**Describe alternatives you've considered**
One blue tag + number of active missions, one orange tab + number of available missions. More consideration for blinking tags and dark blue tags needed for that.

**Additional context**
[Example image](https://user-images.githubusercontent.com/101399/61390571-e2b72380-a86f-11e9-8040-60f9e162addc.png)

^Note, the available mission is not shown in the map as an orange tag.

petervdmeer:
Attached is a savegame that can be used for testing this feature (it has a destination with more than 12 jobs, and a spaceport job that also goes to this destination):
[Busy Tester.txt](https://github.com/endless-sky/endless-sky/files/8403382/Busy.Tester.txt)

Note, the player/tester needs to buy more shuttles and/or starbarges in order to be able to accept all jobs to the crowded destination.

PR #4482, opened 2019-08-18
Draw "available job" pointers before "active job" pointers

Jamibaraki:
Ref #4436

This PR swaps draw code for the available and active missions, leaving the available missions visible when more than 12 missions are active in a system. You can see a hint of blue behind them as they are drawn over the top of them.

I've tested this with a save file with many missions available, though not with special missions.


MCOfficer:
please, give your PR's descriptive titles. nobody knows by heart what 4436 is.

Jamibaraki:
Sure, I'll try and change it to do it that way.

tehhowch:
I agree with Peter.

The special job system pointer should probably be drawn first, and then the remaining slots should be filled (possibly proportionally) by active and available jobs with the respective colorings.

We can consider a tooltip when hovering over the pointer ring that displays the job category count.

Jamibaraki:
OK, I've updated this, replacing the <map angles> with a map of integers to count the missions per system, and only drawing a max of 12.

As suggested, I've also moved the special system code to draw first. I'm not sure how to test the special systems in the game itself though.

The code at line 264 includes a temporary int variable to feed into the function, which is a little clunky, but it was doing the same thing with a temporary angle before.

Please let me know if there are any issues with it, then after that I'll look at adding a tooltip.

source/MapPanel.cpp
```diff
@@ -1143,16 +1145,21 @@ void MapPanel::DrawTooltips()
...
-void MapPanel::DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack, bool bigger)
+void MapPanel::DrawPointer(Point position, int &systemCount, const Color &color, bool drawBack, bool bigger)
 {
-	angle += Angle(30.);
+	Angle angle = Angle();
+	++systemCount;
+	angle += Angle(30.*systemCount);
+	if(systemCount>12) {
+        return;
+    }
```
petervdmeer: There is an alignment/indentation issue here with tabs and spaces.

source/MapPanel.cpp
```diff
@@ -1143,16 +1145,21 @@ void MapPanel::DrawTooltips()
...
 
 
 
-void MapPanel::DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack, bool bigger)
+void MapPanel::DrawPointer(Point position, int &systemCount, const Color &color, bool drawBack, bool bigger)
 {
-	angle += Angle(30.);
+	Angle angle = Angle();
+	++systemCount;
+	angle += Angle(30.*systemCount);
```
petervdmeer: Minor detail, but I think it's possible to construct the Angle in one go:
`Angle angle = Angle(30.*systemCount);`
Also for the special system above.

Jamibaraki:
OK - updated. Does that fix it correctly?

source/MapPanel.cpp
```diff
@@ -1143,16 +1144,20 @@ void MapPanel::DrawTooltips()
...
 
 
-void MapPanel::DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack, bool bigger)
+void MapPanel::DrawPointer(Point position, int &systemCount, const Color &color, bool drawBack, bool bigger)
 {
-	angle += Angle(30.);
+	++systemCount;
+	Angle angle = Angle(30.*systemCount);
+	if(systemCount>12) {
+    	return;
```
petervdmeer: There still is an issue with spaces instead of tabs here.

Jamibaraki: Thanks for spotting it! Is it OK now?

Jamibaraki:
Here are screenshots of the first two requested situations (1 available + 13 active, and 1 active and 13 available). Next I'll try to figure out how to get a save with the special mission active as well. If anyone would happen to have a save file representing that specific scenario, feel free to share :-)

13active1available
13available1active

Jamibaraki:
Here are two more screenshots taken after being assigned a special mission when landing and selecting spaceport, then, after quitting and adding all the extra missions at that location in the save file, reloading and viewing the map screen.

13active1availableWithSpecial
13available1activeWithSpecial

tehhowch:
Your last two screenshots don't show the special system pointer (view the map while being offered the spaceport mission, not after accepting it).

Your "many available + 1 active" screenshot doesn't show the blue pointer for an active mission

Jamibaraki:
Thanks. I was wondering how to actually view the special system pointer on the map. The trouble now is that I'm not sure how to get it to appear in a specific system for the purpose of generating a screenshot. I have a saved game where I can fly to a system that has a mission available, but it isn't always generated for the same location.

I'm creating the 13+1 active/available scenarios by modifying the save game file before or after flying to this system. I could alter the code to generate a special mission in the correct system, but then the screenshots wouldn't be from the actual code in the pull request.

Moving on to the second point about not showing the blue pointers - the goal was to fix the orange pointers not showing in the inverse situation, so this fix wouldn't be expected to show the blue pointer. 

I'm not sure if that situation will arise in game, but if so, we could switch it to draw whichever mission type has the lowest count first, perhaps? Or put a plus next to the systems with hidden missions?

What do you think?

petervdmeer:
I updated this PR with a merge from master and with the functionality to show both available jobs and accepted jobs when the total amount for some system is more than 12.

Tested with the savegame I provided on #4436:
- Nothing accepted:
nothing_accepted
- Special marker (and blinking destination): 
marker_and_blinking
- About half accepted:
about_half_accepted
- Most accepted:
most_accepted
- All accepted:
all_accepted

petervdmeer (approved):

source/MapPanel.cpp
petervdmeer: Yes, it is OK now.

source/MapPanel.cpp
```diff
@@ -1152,17 +1153,37 @@ void MapPanel::DrawMissions()
 	const Color &blockedColor = *colors.Get("blocked mission");
 	const Color &specialColor = *colors.Get("special mission");
 	const Color &waypointColor = *colors.Get("waypoint");
+	if(specialSystem)
+	{
+		// The special system pointer is larger than the others.
+		pointerCount[specialSystem] += 1;
+		Angle a = Angle(30.*pointerCount[specialSystem] );
```
quyykk, line 1160:
```suggestion
		Angle a = Angle(30. * pointerCount[specialSystem]);
```

source/MapPanel.cpp
```diff
@@ -1152,17 +1153,37 @@ void MapPanel::DrawMissions()
...
+		PointerShader::Draw(pos, a.Unit(), 20.f, 27.f, -4.f, black);
+		PointerShader::Draw(pos, a.Unit(), 11.5f, 21.5f, -6.f, specialColor);
+	}
+	// Calculate the available (and unavailable) jobs, but don't draw them yet.
 	for(const Mission &mission : player.AvailableJobs())
 	{
 		const System *system = mission.Destination()->GetSystem();
-		DrawPointer(system, angle[system], mission.CanAccept(player) ? availableColor : unavailableColor);
+		if(mission.CanAccept(player))
+			++(availableCount[system].first);
```
quyykk, line 1170:
```suggestion
			++availableCount[system].first;
```

? There seems to be examples with and without throughout the code. We should really decide which one. My vote is without.

source/MapPanel.cpp
```diff
@@ -1171,20 +1192,22 @@ void MapPanel::DrawMissions()
...
+	for(const auto &it : availableCount)
 	{
-		// The special system pointer is larger than the others.
-		Angle a = (angle[specialSystem] += Angle(30.));
-		Point pos = Zoom() * (specialSystem->Position() + center);
-		PointerShader::Draw(pos, a.Unit(), 20.f, 27.f, -4.f, black);
-		PointerShader::Draw(pos, a.Unit(), 11.5f, 21.5f, -6.f, specialColor);
+		auto &system = it.first;
+		auto &available = it.second;
+		for(int i=0; i < available.first; i++)
```
quyykk, line 1207:
```suggestion
		for(int i = 0; i < available.first; ++i)
```

source/MapPanel.cpp
```diff
@@ -1171,20 +1192,22 @@ void MapPanel::DrawMissions()
...
+			DrawPointer(system, pointerCount[system], availableColor);
+		for(int i=0; i < available.second; i++)
```
quyykk, line 1209:
```suggestion
		for(int i = 0; i < available.second; ++i)
```

source/MapPanel.cpp
```diff
@@ -1254,16 +1277,20 @@ void MapPanel::DrawTooltips()
...
 
-void MapPanel::DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack, bool bigger)
+void MapPanel::DrawPointer(Point position, int &systemCount, const Color &color, bool drawBack, bool bigger)
 {
-	angle += Angle(30.);
+	++systemCount;
+	Angle angle = Angle(30.*systemCount);
+	if(systemCount>12) {
+		return;
+	}
```
quyykk, line 1293:
```suggestion
	if(systemCount > 12)
		return;
```

source/MapPanel.cpp
```diff
@@ -1254,16 +1277,20 @@ void MapPanel::DrawTooltips()
...
 }
 
 
 
-void MapPanel::DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack, bool bigger)
+void MapPanel::DrawPointer(Point position, int &systemCount, const Color &color, bool drawBack, bool bigger)
 {
-	angle += Angle(30.);
+	++systemCount;
+	Angle angle = Angle(30.*systemCount);
```
quyykk, line 1290:
```suggestion
	Angle angle = Angle(30. * systemCount);
```

source/MapPanel.cpp
```diff
@@ -1171,20 +1192,22 @@ void MapPanel::DrawMissions()
...
-	if(specialSystem)
+	// Draw the available and unavailable jobs.
+	for(const auto &it : availableCount)
 	{
-		// The special system pointer is larger than the others.
-		Angle a = (angle[specialSystem] += Angle(30.));
-		Point pos = Zoom() * (specialSystem->Position() + center);
-		PointerShader::Draw(pos, a.Unit(), 20.f, 27.f, -4.f, black);
-		PointerShader::Draw(pos, a.Unit(), 11.5f, 21.5f, -6.f, specialColor);
+		auto &system = it.first;
```
quyykk, line 1205:
```suggestion
		const auto &system = it.first;
```

source/MapPanel.cpp
```diff
@@ -1171,20 +1192,22 @@ void MapPanel::DrawMissions()
...
+		auto &available = it.second;
```
quyykk, line 1206:
```suggestion
		const auto &available = it.second;
```

source/MapPanel.cpp
```diff
@@ -1152,17 +1153,37 @@ void MapPanel::DrawMissions()
 	const Color &blockedColor = *colors.Get("blocked mission");
 	const Color &specialColor = *colors.Get("special mission");
 	const Color &waypointColor = *colors.Get("waypoint");
+	if(specialSystem)
+	{
+		// The special system pointer is larger than the others.
+		pointerCount[specialSystem] += 1;
```
quyykk, line 1159:
```suggestion
		++pointerCount[specialSystem];
```

source/MapPanel.cpp
```diff
@@ -1171,20 +1192,22 @@ void MapPanel::DrawMissions()
...
-		Angle a = (angle[specialSystem] += Angle(30.));
-		Point pos = Zoom() * (specialSystem->Position() + center);
-		PointerShader::Draw(pos, a.Unit(), 20.f, 27.f, -4.f, black);
-		PointerShader::Draw(pos, a.Unit(), 11.5f, 21.5f, -6.f, specialColor);
+		const auto &system = it.first;
+		const auto &available = it.second;
+		for(int i = 0; i < available.first; ++i)
+			DrawPointer(system, pointerCount[system], availableColor);
+		for(int i = 0; i < available.second; ++i)
+			DrawPointer(system, pointerCount[system], unavailableColor);
```
tehhowch, line 1210: Hoist the map indexing out of the loop.

source/MapPanel.cpp
```diff
@@ -1143,7 +1143,8 @@ void MapPanel::DrawNames()
 void MapPanel::DrawMissions()
 {
 	// Draw a pointer for each active or available mission.
-	map<const System *, Angle> angle;
+	map<const System *, int> pointerCount;
+	map<const System *, pair<int, int>> availableCount;
```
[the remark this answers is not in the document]
petervdmeer: I added a file-local struct.

source/MapPanel.cpp
```diff
@@ -59,6 +59,15 @@ namespace {
 	const std::string SHOW_ESCORT_SYSTEMS = "Show escort systems on map";
 	const std::string SHOW_STORED_OUTFITS = "Show stored outfits on map";
 
+	// Struct to track per system how many pointers are drawn and still
+	// need to be drawn.
+	struct PointerDrawCount{
+		// Amount of systems already drawn.
```
quyykk, line 65:
```suggestion
	class PointerDrawCount {
	public:
		// Amount of systems already drawn.
```

MZ doesn't like structs.

source/MapPanel.cpp
```diff
@@ -1152,17 +1161,38 @@ void MapPanel::DrawMissions()
...
 	{
 		if(!mission.IsVisible())
 			continue;
 
 		const System *system = mission.Destination()->GetSystem();
+
+		// Reserve a maximum of 6 slots for available missions.
+		auto &&it = pointerCount[system];
+		int reserved = min(6, it.available + it.unavailable);
+		if(it.drawn >= 12 - reserved)
```
quyykk, line 1193: Make the 12 a global constant with a cool name (since it's used in two places). Maybe also for the `30.` (degrees), or we calculate the degree from the constant if that makes sense.

source/MapPanel.cpp
tehhowch: I'm fine with their use when it is to provide named fields to what would otherwise be a pair or a tuple.

quyykk: Yeah same. We'll need to update the style guide at some point.

source/MapPanel.cpp
```diff
@@ -1254,16 +1286,19 @@ void MapPanel::DrawTooltips()
...
 
 
-void MapPanel::DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack, bool bigger)
+void MapPanel::DrawPointer(Point position, int &systemCount, const Color &color, bool drawBack, bool bigger)
 {
-	angle += Angle(30.);
+	++systemCount;
+	Angle angle = Angle(30. * systemCount);
+	if(systemCount > 12)
+		return;
```
tehhowch, line 1301: Do less work:

```suggestion
	if(++systemCount > 12)
		return;
	Angle angle = Angle(30. * systemCount);
```

source/MapPanel.cpp
```diff
@@ -59,6 +59,15 @@ namespace {
 	const std::string SHOW_ESCORT_SYSTEMS = "Show escort systems on map";
 	const std::string SHOW_STORED_OUTFITS = "Show stored outfits on map";
 
+	// Struct to track per system how many pointers are drawn and still
+	// need to be drawn.
+	struct PointerDrawCount{
+		// Amount of systems already drawn.
+		int drawn = 0;
+		int available = 0;
+		int unavailable = 0;
```
tehhowch, line 68: Probably these should be `unsigned`

source/MapPanel.cpp
```diff
@@ -58,6 +58,17 @@ using namespace std;
 namespace {
 	const std::string SHOW_ESCORT_SYSTEMS = "Show escort systems on map";
 	const std::string SHOW_STORED_OUTFITS = "Show stored outfits on map";
+	const unsigned int MAX_MISSION_POINTERS_DRAWN = 12;
+	const double MISSION_POINTERS_ANGLE_DELTA = 30.;
+
+	// Struct to track per system how many pointers are drawn and still
+	// need to be drawn.
+	struct PointerDrawCount{
```
quyykk, line 66:
```suggestion
	struct PointerDrawCount {
```

source/MapPanel.cpp
```diff
@@ -320,9 +331,12 @@ void MapPanel::DrawMiniMap(const PlayerInfo &player, float alpha, const System *
 			RingShader::Draw(to, OUTER, INNER, color);
 		}
 
-		Angle angle;
+		unsigned int missionCounter = 0;
```
quyykk, line 334: I believe the style is to use plain `unsigned` over `unsigned int`.

source/MapPanel.cpp
```diff
@@ -58,6 +58,17 @@ using namespace std;
 namespace {
 	const std::string SHOW_ESCORT_SYSTEMS = "Show escort systems on map";
 	const std::string SHOW_STORED_OUTFITS = "Show stored outfits on map";
+	const unsigned int MAX_MISSION_POINTERS_DRAWN = 12;
```
petervdmeer, line 61:
```suggestion
	const unsigned MAX_MISSION_POINTERS_DRAWN = 12;
```

source/MapPanel.cpp
```diff
@@ -58,6 +58,17 @@ using namespace std;
...
+	const double MISSION_POINTERS_ANGLE_DELTA = 30.;
+
+	// Struct to track per system how many pointers are drawn and still
+	// need to be drawn.
+	struct PointerDrawCount{
+		// Amount of systems already drawn.
+		unsigned int drawn = 0;
+		unsigned int available = 0;
+		unsigned int unavailable = 0;
```
petervdmeer, line 70:
```suggestion
		unsigned drawn = 0;
		unsigned available = 0;
		unsigned unavailable = 0;
```

source/MapPanel.cpp
```diff
@@ -1171,20 +1216,22 @@ void MapPanel::DrawMissions()
...
+	for(auto &&it : missionCount)
 	{
-		// The special system pointer is larger than the others.
-		Angle a = (angle[specialSystem] += Angle(30.));
-		Point pos = Zoom() * (specialSystem->Position() + center);
-		PointerShader::Draw(pos, a.Unit(), 20.f, 27.f, -4.f, black);
-		PointerShader::Draw(pos, a.Unit(), 11.5f, 21.5f, -6.f, specialColor);
+		const auto &system = it.first;
+		auto &&counters = it.second;
+		for(unsigned int i = 0; i < counters.available; ++i)
```
petervdmeer, line 1231:
```suggestion
		for(unsigned i = 0; i < counters.available; ++i)
```

source/MapPanel.cpp
```diff
@@ -1254,16 +1301,18 @@ void MapPanel::DrawTooltips()
 
 
 
-void MapPanel::DrawPointer(const System *system, Angle &angle, const Color &color, bool bigger)
+void MapPanel::DrawPointer(const System *system, unsigned int &systemCount, const Color &color, bool bigger)
```
petervdmeer, line 1304:
```suggestion
void MapPanel::DrawPointer(const System *system, unsigned &systemCount, const Color &color, bool bigger)
```

source/MapPanel.cpp
```diff
@@ -1254,16 +1301,18 @@ void MapPanel::DrawTooltips()
...
 {
-	DrawPointer(Zoom() * (system->Position() + center), angle, color, true, bigger);
+	DrawPointer(Zoom() * (system->Position() + center), systemCount, color, true, bigger);
 }
 
 
 
-void MapPanel::DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack, bool bigger)
+void MapPanel::DrawPointer(Point position, unsigned int &systemCount, const Color &color, bool drawBack, bool bigger)
```
petervdmeer, line 1311:
```suggestion
void MapPanel::DrawPointer(Point position, unsigned &systemCount, const Color &color, bool drawBack, bool bigger)
```

source/MapPanel.h
```diff
@@ -163,8 +163,8 @@ class MapPanel : public Panel {
 	void DrawNames();
 	void DrawMissions();
 	void DrawTooltips();
-	void DrawPointer(const System *system, Angle &angle, const Color &color, bool bigger = false);
-	static void DrawPointer(Point position, Angle &angle, const Color &color, bool drawBack = true, bool bigger = false);
+	void DrawPointer(const System *system, unsigned int &systemCount, const Color &color, bool bigger = false);
+	static void DrawPointer(Point position, unsigned int &systemCount, const Color &color, bool drawBack = true, bool bigger = false);
```
petervdmeer, line 167:
```suggestion
	void DrawPointer(const System *system, unsigned &systemCount, const Color &color, bool bigger = false);
	static void DrawPointer(Point position, unsigned &systemCount, const Color &color, bool drawBack = true, bool bigger = false);
```

source/MapPanel.cpp
petervdmeer: Yes, that seems to be the common style; I changed them to `unsigned`.
Can we automatically check for this in the coding style checker? (Which I believe you were working on?)

quyykk: No unfortunately not.

petervdmeer:
All comments seem to have been processed; ready for merging.

PR #4486, opened 2019-08-25
Source code normalization

stechio:
I noticed that _frequently-used constants are expressed as literals throughout the source code_. For the sake of consistency, robustness and readability, this PR normalizes such instances into their respective constants:

- `60` to `FRAME_RATE`
- `100` to `PERCENT`
- recurrent angles (degrees): `0`, `90`, `180`, `360` to corresponding `DEG_*`
- `pi.h` constants file generalized to `global.h` (XCode update courtesy of @Fzzr)


source/global.h
```diff
@@ -0,0 +1,14 @@
+/*
```
petervdmeer: I would expect the default/standard ES copyright header here.

source/OutfitInfoDisplay.cpp
```diff
@@ -27,49 +28,49 @@ using namespace std;
...
+		{"thrusting energy", FRAME_RATE},
+		{"thrusting heat", FRAME_RATE},
+		{"turn", FRAME_RATE},
+		{"turning energy", FRAME_RATE},
+		{"turning heat", FRAME_RATE},
 		
-		{"thrust", 60. * 60.},
-		{"reverse thrust", 60. * 60.},
-		{"afterburner thrust", 60. * 60.},
+		{"thrust", FRAME_RATE * 60.},
```
finite-galaxy: square (because accelaration is per s²)

source/OutfitInfoDisplay.cpp
```diff
@@ -27,49 +28,49 @@ using namespace std;
...
+		{"reverse thrust", FRAME_RATE * 60.},
+		{"afterburner thrust", FRAME_RATE * 60.},
 		
-		{"ion resistance", 60. * 100.},
-		{"disruption resistance", 60. * 100.},
-		{"slowing resistance", 60. * 100.}
+		{"ion resistance", FRAME_RATE * 100.},
```
finite-galaxy: 100 because percentage

finite-galaxy:
Maybe merge `pi.h` into it? That's a constant too.
Also, shouldn't the XCode files be updated whenever a C++ source code file is added?

petervdmeer:
> Maybe merge `pi.h` into it? That's a constant too.
> Also, shouldn't the XCode files be updated whenever a C++ source code file is added?

That sounds like a good idea. Better than having many small files with just 1 or 2 lines.

source/OutfitInfoDisplay.cpp
stechio: ok

source/OutfitInfoDisplay.cpp
stechio: ok

source/global.h
stechio: ok

stechio:
> > Maybe merge `pi.h` into it? That's a constant too.
>
> That sounds like a good idea. Better than having many small files with just 1 or 2 lines.

ok

> Also, shouldn't the XCode files be updated whenever a C++ source code file is added?

IMHO, projects should be IDE-agnostic. Anyway, since I'm not an apple guy and cannot access that IDE, how should exactly update those files? thanks!

stechio:
[Amendment 1](https://github.com/endless-sky/endless-sky/pull/4486/commits/6b1bf0853af2e055ff12d8f335820f6cadb2df82) solves all the reported issues but the XCode files update (waiting for instructions here).

stechio:
[Amendment 2](https://github.com/endless-sky/endless-sky/pull/4486/commits/9142c06b4ebe322f86d4158e336e211ab598abf2) covers all orthogonal angle constants.

source/Ship.cpp
```diff
@@ -47,11 +48,11 @@ namespace {
 	const vector<string> BAY_TYPE = {"drone", "fighter"};
 	const vector<string> BAY_SIDE = {"inside", "over", "under"};
 	const vector<string> BAY_FACING = {"forward", "left", "right", "back"};
-	const vector<Angle> BAY_ANGLE = {Angle(0.), Angle(-90.), Angle(90.), Angle(180.)};
+	const vector<Angle> BAY_ANGLE = {Angle(DEG_0), Angle(-DEG_90), Angle(DEG_90), Angle(DEG_180)};
```
finite-galaxy: Shouldn't `-DEG_90` be `DEG_270`?

finite-galaxy:
It helps if you update your opening post whenever you commit more changes.
> IMHO, projects should be IDE-agnostic. Anyway, since I'm not an apple guy and cannot access that IDE, how should exactly update those files? thanks!

No idea, GNU/Linux here. All I know is that XCode is a reason why #4123 is still not accepted.

source/global.h
```diff
@@ -10,16 +10,25 @@ WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
...
+#ifndef GLOBAL_H_
+#define GLOBAL_H_
 
 
 
+const int FRAME_RATE = 60;
+
+const double DEG_0 = 0;
+const double DEG_90 = 90;
+const double DEG_180 = DEG_90 * 2;
```
finite-galaxy: Why not simply 180?

source/Ship.cpp
stechio: Yes, of course, they are geometrically equivalent but, considering they are relative to an oriented system (the ship vector), I think representing left as counterclockwise angle is more expressive -- for example, I saw the same notation in OpenCV:

```
enum RotateFlags {
    ROTATE_90_CLOCKWISE = 0, //!<Rotate 90 degrees clockwise
    ROTATE_180 = 1, //!<Rotate 180 degrees clockwise
    ROTATE_90_COUNTERCLOCKWISE = 2, //!<Rotate 270 degrees clockwise
};```

File: source/global.h (hunk 5 above)
Comment (stechio) [reply source/global.h]: I meant to emphasize the geometrical tie between those constants (oops: `DG_0 = DEG_90 * 0`).

**Comment by stechio** [comment]
> It helps if you update your opening post whenever you commit more changes.

You're right -- got it.

> > IMHO, projects should be IDE-agnostic. Anyway, since I'm not an apple guy and cannot access that IDE, how should exactly update those files? thanks!
> 
> No idea, GNU/Linux here. All I know is that XCode is a reason why #4123 is still not accepted.

It sounds like a conundrum... isn't there somebody designated to update those platform-specific files?

File: source/Ship.cpp (hunk 4 above)
Comment (finite-galaxy) [reply source/Ship.cpp]: Yeah, for rotations a [-180°, 180°) range makes sense, for positions a [0, 360°) range. I vaguely recall reading a comment to that effect somewhere.
Anyway, I don't really care either way, both options are clear enough for me (as is `-90.`).

File: source/global.h (hunk 5 above)
Comment (finite-galaxy) [reply source/global.h]: Then again, why not `DEG_90 = DEG_360 / 4`? Given that this is a file for constants, the multiplication seems a bit out of place. Anyway, I suppose it's simply a matter of personal taste.

**Comment by finite-galaxy** [comment]
> You're right -- got it.

`pi.h`

> It sounds like a conundrum... isn't there anybody designated to update those platform-specific files?

No idea, this is not my project; I merely comment on pull requests I think are useful.

**Comment by MCOfficer** [comment]
> It sounds like a conundrum... isn't there anybody designated to update those platform-specific files?

yeah, it's a mess. we have 4 build files for 3 platforms, and at least 2 need to be updated accordingly... i dream of a time where we have something universal like CMake or whatever, but until then your best bet is to ask someone with a mac to do it for you. The [Discord](https://discord.gg/ZeuASSx) might be a good place to start.

File: source/global.h (hunk 5 above)
Comment (stechio) [reply source/global.h]: Since they are resolved at compile time, I see no issue here, just a way to improve their consistency. Anyway, I agree with your criticism (see [Amendment 4](https://github.com/endless-sky/endless-sky/pull/4486/commits/c35cb7ec7a48ef55a385034307425337c3d426c9)).

**Comment by stechio** [comment]
> > It sounds like a conundrum... isn't there anybody designated to update those platform-specific files?
> 
> yeah, it's a mess. we have 4 build files for 3 platforms, and at least 2 need to be updated accordingly... i dream of a time where we have something universal like CMake or whatever, but until then your best bet is to ask someone with a mac to do it for you. The [Discord](https://discord.gg/ZeuASSx) might be a good place to start.

I mean: if an IDE-specific configuration is maintained, there should be someone who takes the responsibility to provide updates when needed. Otherwise, if there's not enough interest in its regular maintenance (without hunting for a helper), maybe it's time to ditch it... :)

**Comment by stechio** [comment]
> Of late I've done a lot of the macOS testing and support. Here's my branch with an updated Xcode project file for your changes: https://github.com/Fzzr/endless-sky/tree/feat/source-norm. 

Thank you, really appreciated!

> As for cross-platform build automation, @tehhowch suggested I investigate [scons-xcode](https://bitbucket.org/al45tair/scons-xcode/src/default/README.rst) which I have not yet had time to test in earnest. Someone else will have to look into options for Windows.

Sounds promising.

> Even with something like that automating build changes, ES will always need testers for all supported platforms. I don't know how many Windows or macOS users there are. (Maybe the stats are available on Steam, but only @endless-sky has access to that.) I know there are enough that bug reports come in regularly. As MCO said, I recommend joining the [Discord](https://discord.gg/ZeuASSx) for easier and faster communication.

That's fine, I'll consider in my next contributions.

File: source/Body.cpp [hunk 6]
```diff
@@ -263,15 +264,15 @@ void Body::SetSwizzle(int swizzle)
 // a sprite instead of a full animation data structure.
 void Body::SetFrameRate(float framesPerSecond)
 {
-	frameRate = framesPerSecond / 60.f;
+	frameRate = framesPerSecond / FRAME_RATE;
```
tehhowch: Note that here and several other places, you are changing a float to a int which means there is now int->float conversion whenever these lines gets executed. In most cases this will basically not be noticeable.

source/Planet.cpp
```diff
@@ -574,7 +575,7 @@ string Planet::DemandTribute(PlayerInfo &player) const
 // While being tributed, attempt to spawn the next specified defense fleet.
 void Planet::DeployDefense(list<shared_ptr<Ship>> &ships) const
 {
-	if(!isDefending || Random::Int(60) || defenseDeployed == defenseFleets.size())
+	if(!isDefending || Random::Int(FRAME_RATE) || defenseDeployed == defenseFleets.size())
```
tehhowch: Here again it is just a coincidence that we currently launch about 1 defense fleet per second.

source/Projectile.cpp
```diff
@@ -139,7 +139,7 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 	double turn = weapon->Turn();
 	double accel = weapon->Acceleration();
 	int homing = weapon->Homing();
-	if(target && homing && !Random::Int(60))
+	if(target && homing && !Random::Int(FRAME_RATE))
```
tehhowch: Here too

source/Ship.cpp
```diff
@@ -47,11 +48,11 @@ namespace {
...
 	
 	const double MAXIMUM_TEMPERATURE = 100.;
 	
-	const double SCAN_TIME = 60.;
+	const double SCAN_TIME = FRAME_RATE;
```
tehhowch: Another coincidence here.

source/global.h
```diff
@@ -10,16 +10,25 @@ WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
...
 */
 
-#ifndef PI_H_
-#define PI_H_
+#ifndef GLOBAL_H_
+#define GLOBAL_H_
 
 
 
+const int FRAME_RATE = 60;
```
tehhowch: Probably every constant you are adding should actually be `constexpr` instead of just `const`

tehhowch:
Closing this as it is very out-of-date and includes a number of changes that aren't sensible or desired.

Issue #4487, opened 2019-08-25
Plugins cannot override stock missions


SolraBizna:
**Describe the bug**
When a plugin and the builtin data folder each have a mission with the same identifier, the builtin version is used.

**To Reproduce**
1. Put the following in a new file in the Endless Sky data folder:
   ```
   mission "Override Test"
           name "Override Test"
           description "This is the STOCK version of the override test."
           job
   ```
2. Make a new folder in the plugins folder, a `data` folder inside that folder, and a new file in that `data` folder containing the following:
   ```
   mission "Override Test"
           name "Override Test"
           description "This is the PLUGIN version of the override test."
           job
   ```
3. Enter the game, take off, land, and check the Job Board.

**Expected behavior**
Seeing a mission named "Override Test" on the job board whose description says it is the "PLUGIN" version.

**Actual behavior**
Seeing a mission named "Override Test" on the job board whose description says it is the "STOCK" version.

**System:**
 - OS: Debian GNU/Linux
 - Game Source: self-compiled
 - Version: "commit hash e27dc281"

**Additional context**
I was attempting to make a plugin that replaces the stock "Hunt Down" missions with tweaked ones. New missions (with unique identifiers) that I created worked perfectly, but I wasn't able to override the stock missions in any way. Ideally, a plugin mission with the same identifier as a stock one will entirely override it, as it was in Escape Velocity; plugins could then neuter or replace troublesome missions as needed.

beccabunny:
The usual way to override a vanilla mission is to make a second mission (let's say, mission "Fail Override Test")  that fails that vanilla mission, then you give the plugin mission a different name and `to offer has "Fail Override Test: offered"`.

That said, I agree it would be great to have easier, and less exploitable, ways to do this

SolraBizna:
That lets you replace a mission that's already in progress, but it wouldn't let you prevent a stock mission from appearing in the Job Board or having "on offer" effects.

comnom:
This was added in https://github.com/endless-sky/endless-sky/commit/efba03c6c4e14064787a0fab438dfe3a99b74754 . Using a start node or an event to set a mission's offered condition allows you to bypass the mission and not need to modify it.

SolraBizna:
Bypassing a mission by "faking" its `offered` condition would leave that mission bypassed even if the plugin is subsequently removed, which isn't great.

[...]

PR #4489, opened 2019-08-25
Allow plugins to explicitly override already-defined missions and outfits

SolraBizna:
Related to #3996 and #4487.

I wanted to make a plugin that made a minor change to the way the stock "Hunt Down \<npc>" missions worked, but found that there was no way either to prevent the stock missions from appearing or to override any aspect of their behavior. Disabling the stock missions (e.g. with #3996) and making my own copies with unique names would be a workable solution, but would result in separate counts for e.g. `"Bounty Hunting (Small): offered"` and `"Fancy Plugin Name Bounty Hunting (Small): offered"`; in this particular case there would be no real harm, but I can easily imagine use cases where that would be bad. Manual and tedious condition-massaging would fix it, but this solution is far simpler...

Current behavior is that when a mission is loaded with the same name as an existing mission, the new data is completely ignored. This PR makes it so that a mission can explicitly tag itself as an `override`, in which case it will entirely replace any existing mission with the same name. The onus for dealing with the consequences is on the plugin developer, as expected when tampering with stock data.

Example:

```
mission "Bounty Hunting (Small)" override
    (mission data goes here)
```

Existing behavior in the absence of `override` tag is not changed. If a mission with an `override` tag does not share the same name as an existing mission, a warning is printed and the mission is loaded anyway.

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -107,7 +107,8 @@ void Mission::Load(const DataNode &node)
 		}
 		else
 		{
-			node.PrintTrace("Duplicate definition of mission:");
+			node.PrintTrace("Duplicate definition of mission "
+					"(add \"override\" if intentional):");
 			return;
 		}
 	}
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -92,14 +92,29 @@ void Mission::Load(const DataNode &node)
 		node.PrintTrace("No name specified for mission:");
 		return;
 	}
-	// If a mission object is "loaded" twice, that is most likely an error (e.g.
-	// due to a plugin containing a mission with the same name as the base game
-	// or another plugin). This class is not designed to allow merging or
-	// overriding of mission data from two different definitions.
+	bool isOverride = (node.Size() >= 3 && node.Token(2) == "override");
+	// If a mission object is "loaded" twice, that is most likely an error
+	// (e.g. due to a plugin containing a mission with the same name as the
+	// base game or another plugin). However, sometimes it's useful to
+	// intentionally override a specific mission from a plugin.
 	if(!name.empty())
 	{
-		node.PrintTrace("Duplicate definition of mission:");
-		return;
+		if(isOverride)
+		{
+			// If explicitly requested, start from a blank slate,
+			// so that all data is replaced.
+			*this = Mission();
+		}
+		else
+		{
+			node.PrintTrace("Duplicate definition of mission:");
+			return;
+		}
+	}
+	else if(isOverride)
+	{
+		node.PrintTrace("Warning: overriding a non-existent mission:");
+		// Proceed to load the mission as if override were not given.
 	}
 	name = node.Token(1);
 	
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -46,26 +46,10 @@ void Outfit::Load(const DataNode &node)
 	// Note: an outfit with no name wouldn't be global and therefore
 	// overriding wouldn't make sense (?).
 	bool isOverride = (node.Size() >= 3 && node.Token(2) == "override");
-	if(!name.empty())
-	{
-		if(isOverride)
-		{
-			// If explicitly requested, start from a blank slate,
-			// so that all data is replaced.
-			*this = Outfit();
-		}
-		else
-		{
-			node.PrintTrace("Duplicate definition of outfit "
-					"(add \"override\" if intentional):");
-			return;
-		}
-	}
+	if(!name.empty() && isOverride)
+		*this = Outfit();
 	else if(isOverride)
-	{
 		node.PrintTrace("Warning: overriding a non-existent outfit:");
-		// Proceed to load the outfit as if override were not given.
-	}
 	
 	if(node.Size() >= 2)
 	{
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -43,12 +43,36 @@ const vector<string> Outfit::CATEGORIES = {
 
 void Outfit::Load(const DataNode &node)
 {
+	// Note: an outfit with no name wouldn't be global and therefore
+	// overriding wouldn't make sense (?).
+	bool isOverride = (node.Size() >= 3 && node.Token(2) == "override");
+	if(!name.empty())
+	{
+		if(isOverride)
+		{
+			// If explicitly requested, start from a blank slate,
+			// so that all data is replaced.
+			*this = Outfit();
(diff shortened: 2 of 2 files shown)


Amazinite:
I believe that it's also the case that plugins can't override outfits or ships from vanilla. Perhaps a different PR should handle allowing those to be overrided.

source/Mission.cpp
```diff
@@ -92,14 +92,29 @@ void Mission::Load(const DataNode &node)
...
-		return;
+		if(isOverride)
+		{
+			// If explicitly requested, start from a blank slate,
+			// so that all data is replaced.
+			*this = Mission();
+		}
+		else
+		{
+			node.PrintTrace("Duplicate definition of mission:");
```
Amazinite: Perhaps this error should reference the existence of the `override` tag.
`Duplicate definition of mission, did you mean to override?:`

SolraBizna: That's a good idea, since it makes override a lot more discoverable.

Amazinite:
Do you want to extend this functionality to ships and outfits in this PR while we're at it? Overriding ships, outfits, and missions are the three things that people ask about the most for plugins.

SolraBizna:
The only computer I have access to at the moment can only run Endless Sky in software mode, but I might as well give it a shot.

SolraBizna:
This is about all I can do for now. I'm still a little fuzzy from a recent minor surgery.

I've now made it work with outfits the same way as with missions.

I found that ship replacement more or less already works, and is apparently intentionally partly supported. I couldn't get duplicate detection working reliably for ships—duplicates were detected, but then the presence of an `override` tag caused a spurious "overriding a ship that doesn't already exist" message. I'm not sure why, and don't have the tools to investigate further at the moment, so I gave up on ships.

Making this work "properly" for ships is made more difficult by the way the variant system works... if I modify "Sparrow" to a huge outfit space, "Sparrow (Gatling)" will still have the outfit space its parent had when *it* was loaded. This is probably proper behavior, but might be surprising to a plugin author. In addition, ships being stored in the pilot file may cause other obscure problems I can't test for now.

Amazinite:
I suspect that the problems with setting up overriding for ships in the same way is because of ships having variants. If the third token is typically the variant name, so `ship "Sparrow" override` will be seen as a Sparrow variant named "override", but the presence of `override` also sets isOverride to true. The name of the variant "override" isn't set until after checking if it's a duplicate, so it makes sense that `ship "Sparrow" override` would trigger that warning.

source/Outfit.cpp
```diff
@@ -43,12 +43,36 @@ const vector<string> Outfit::CATEGORIES = {
...
+	{
+		if(isOverride)
+		{
+			// If explicitly requested, start from a blank slate,
+			// so that all data is replaced.
+			*this = Outfit();
+		}
+		else
+		{
+			node.PrintTrace("Duplicate definition of outfit "
```
tehhowch: I don't agree that we should remove the current behavior of allowing plugins to extend the current definition of an Outfit, e.g. add a single attribute to the existing definition. With this PR, one would then need to fully reproduce the original definition, and additionally include the desired alteration.

SolraBizna:
> I suspect that the problems with setting up overriding for ships in the same way is because of ships having variants. If the third token is typically the variant name, so `ship "Sparrow" override` will be seen as a Sparrow variant named "override", but the presence of `override` also sets isOverride to true. The name of the variant "override" isn't set until after checking if it's a duplicate, so it makes sense that `ship "Sparrow" override` would trigger that warning.

I already handled that in the code I tried, by checking if the last element is "override" and treating the node as being one element smaller if it is, rather than just checking the third element (if it exists) like with outfits and missions.

> I don't agree that we should remove the current behavior of allowing plugins to extend the current definition of an Outfit, e.g. add a single attribute to the existing definition. With this PR, one would then need to fully reproduce the original definition, and additionally include the desired alteration.

This is a good point. "override" is still useful for outfits, but unlike missions it's not the only useful behavior. I will test your solution and add it to the PR.

SolraBizna:
I think there should be a warning in the duplicate outfit case, and another tag ("patch"?) to silence the warning. For now, I prefer not to push a change that might add warning spam to existing plugins.

source/Outfit.cpp
```diff
@@ -43,12 +43,20 @@ const vector<string> Outfit::CATEGORIES = {
...
+	else if(isOverride)
+		node.PrintTrace("Warning: overriding a non-existent outfit:");
+	
 	if(node.Size() >= 2)
 	{
 		name = node.Token(1);
 		pluralName = name + 's';
 	}
-	
+
```
tehhowch: Please avoid trimming whitespace in PRs. The indentation level of the neighboring blocks should be maintained.

SolraBizna: It was an accident. emacs' editorconfig support doesn't handle `trim_trailing_whitespace` completely. I'm not sure how to repair this without creating an entire new commit.

tehhowch: You can't. Don't worry about the number of commits in your PR; it will be squash-merged into a single commit unless you perform a collaborative effort with multiple commit authors.

tehhowch:
A question on implementation: would it be simpler to do this if `override` were a root-level token, rather than a token belonging to a node?

E.g. instead of
```
ship "Arrow" override # very strange interpretation by game engine
    <complete ship definition>
ship "Sparrow" "Worldeater" override # doesn't work
```
You would write
```
override ship "Sparrow" "Worldeater"
    <variant definition>
override ship "Arrow"
    <complete ship definition>
```

I think the implemention could be as simple as
```
<relevant GameData private member>.erase(<name of the thing to be overridden>);
<relevant GameData private member>[<name>].Load(<modified datanode without the "override" token>);
```
To avoid needing to update every Load method, you'd need to have a way to remove the"override" token before passing the node. Or, `override` could awkwardly be a token that has only a child of the new definition, rather than prefixing the other tokens:
```
override
    ship "Arrow"
      <complete ship definition at increased indent level>
```

SolraBizna:
That would move the responsibility of defining override behavior out of the overridable-thing and into the loading-thing, which is a loss for encapsulation, but this PR's override behavior working is predicated on the loading-thing's logic being written a certain way *anyway* (which I think is why it didn't work for ships)...

Your approach would certainly simplify implementation, since the same override support could be applied easily across every conceivable node type. Current type-specific duplicate handling would continue working as it already does. If I were well, I'd try implementing it myself.

Amazinite:

Issue #4409, opened 2019-07-04
[Spoilers] Remnant mission made artificially difficult by low fuel on gas craft


jpretori:
**Is your feature request related to a problem? Please describe.**
The "land on void sprite planets"  from Remnant space is needlessly frustrating.

A combination of factors cause it. The upshot of this combination of factors, is that it takes a very long time to finish the mission, and it's quite risky - I got ambushed by Korath raiders and blown apart.  The factors are:
 - Low fuel reserves on gas craft
 - Must have jumpdrive on gas craft, as there's no hyperspace path to the void sprite system - so it eats 200 fuel per jump, limiting the craft's range to 4 jumps
 - It has a ramscoop, but it is extremely slow, taking a really long time to fill the tanks back up to 200 fuel.

I could drag one of my fleet ships with me to refuel along the way, but it is difficult (maybe impossible?) to keep the Archon from destroying the companion ship, thus making the return journey equally hazardous and frustrating.

I guess I could take my whole fleet with, and try to destroy the archon, but I have little confidence of that succeeding - that thing is toting some serious guns, so I assume its shielding is equally impressive.  And anyway, I have no appetite for annoying the Drak any more than I appear to already be doing, by disturbing the void sprites.

**Related Issue Links**
None that I'm aware of

**Describe the solution you'd like**
I'd like a way (even a sneaky one) to have this mission be less frustrating.  Here's one idea: This craft is made to land on gas giants, right?  Well, maybe when I try to land on gas giants along the way to the void sprites, instead of just saying "can't land on gas giants", it could scoop up hyperspace fuel from the outer atmosphere, thus rapidly refiling the hyperspace fuel tanks to full.  Some players won't think of this, so maybe the mission description could hint at the possibility (or just state it outright).  Maybe even just flying around close to gas giants can have the same effect.
This could maybe be an outfit that I can install on other ships once I have a remnant license?  But it doesn't need to be, to be effective for making this mission less frustrating.

**Describe alternatives you've considered**
 - Change hyperspace pathing so I can get there and back in the gascraft without needing a jump drive. (Don't like this because it seems like such a dull solution, and besides it requires losing the "scattered clumps of systems" feel of the ember waste, which would be a shame)
 - Add a wormhole that cuts down the number of jumps so the gascraft can manage it (don't like this because it seems too close to cheating)
[...]

PR #4491, opened 2019-08-25
Remnant keystone no longer requries license

Naminson:
There has been at least one player who's had a save broken due to losing all keystones in remnant space before obtaining the license (https://steamcommunity.com/app/404410/discussions/0/1639794651173778949/), and therefore being trapped* in it. Although this PR doesn't prevent it from happening, it does make it much easier to escape remnant space. It also makes life easier for players who are outfitting in remnant space and don't want to go all the way up to hai territory.

Admittedly, it's a bit inconsistent with the lore, since the remnant probably wouldn't sell stones to an outsider, but I don't think anyone's going to uninstall the game over that.
*This is technically escapable by two methods, but neither are very intuitive.

### DIFF
--- a/data/remnant outfits.txt
+++ b/data/remnant outfits.txt
@@ -334,8 +334,6 @@ outfit "Emergency Ramscoop"
 
 outfit "Quantum Key Stone"
 	category "Systems"
-	licenses
-		Remnant
 	cost 240000
 	thumbnail "outfit/keystone"
 	"mass" 1


kikotheexile:
Thank you, a thousand times thank you. 

Even if this isn't the best long term solution, at least if it gets merged this will stop being an issue as of next update, removing the casual softlock

Amazinite:
Definitely a bandaid fix for now. If #2911 is finished then a more elegant solution could be done in the form of giving the player a keystone in Remnant space if they don't have the Remnant license yet and if they find themself without a keystone to use.

Do realize though that this fix will likely result in players who stumble upon the Ember Waste with a jump drive (which we have seen in #4409) never even getting to the Key Stones mission because they never care to bring a keystone from Hai space. While it isn't a soft lock on an entire save file, it is essentially a soft lock on the story.

kikotheexile:
> Definitely a bandaid fix for now. If #2911 is finished then a more elegant solution could be done in the form of giving the player a keystone in Remnant space if they don't have the Remnant license yet and if they find themself without a keystone to use.
> Do realize though that this fix will likely result in players who stumble upon the Ember Waste with a jump drive (which we have seen in #4409) never even getting to the Key Stones mission because they never care to bring a keystone from Hai space. While it isn't a soft lock on an entire save file, it is essentially a soft lock on the story.

Well we can have a bandaid fix and then revert this one when the elegant solution is finished, or we can continue to have a softlock. 

I mean its your call, but leaving the softlock is the most moronic idea I have ever heard.

Amazinite:
>I mean its your call, but leaving the softlock is the most moronic idea I have ever heard.

Wondering where you heard that idea from then.

kikotheexile:
>  Amazinite merged commit d9887b4 into endless-sky:master 1 minute ago 

A thousand thanks to you as well.

Zitchas:
I'm definitely not a fan of this change, but I accept the reason for it in the short term. This is something that should definitely be fixed and this change reverted for 0.9.11.

Issue #3341, opened 2017-12-03
"require <outfit>" needs to be in an "on" block; should be able to put it in "to offer" instead


Pointedstick:
This feels weird:

```
to offer
		random < 50
on offer
		require "Brig"
```

I feel like this would intuitively make more sense:

```
to offer
		random < 50
		require "Brig"
```

It would also let you do something like this, which currently isn't possible:
```
to offer
	or
		and
			require not "Brig"
			random < 75
		and
			require "Brig"
			random < 10
```

VonZorch:
Fully agree, it should be testable in a branch statement too.

Amazinite:
Looks like this is actually a duplicate of #2494.

PR #4493, opened 2019-08-26
Added additional Remnant: Key Stones missions 

Amazinite:
Ref: #4409 #4491
These additional missions account for the player not always reaching the Ember Waste with a Quantum Keystone. These missions also make it explicit to the player that the keystones are used to enter the wormholes in the region.

The possible scenarios are now as follows:
* The player arrives in the Ember Waste with a Hai Keystone and is approached by the outfitter manager. This is currently the only possible way to progress the Remnant story.
* The player arrives in the Ember Waste without a Hai Keystone, but does know about the Hai. After helping the Remnant with one set of missions, the outfitter manager may approach the player asking them to search for alternative sources of Key Stones, to which the player will answer that they know of one such location.
* The player arrives in the Ember Waste without a Hai Keystone, but does NOT know about the Hai. After helping the Remnant with one set of missions, the outfitter manager may approach the player asking them to search for alternative sources of Key Stones. The player will then accept the mission to search for new sources. After finding the Hai, the player may return to the manager and tell them about having found an alternative source.

If the player is offered any of the three missions and declines, they won't be able to be offered the other two missions.

### DIFF
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -398,6 +398,7 @@ mision "Remnant: Key Stones (Pre-Hai) 1"
 	to offer
 		not "Remnant: Key Stones (Hai): offered"
 		not "Remnant: Key Stones: offered"
+		not "First Contact: Hai: offered"
 		or
 			has "event: remnant: void sprite research"
 			has "Remnant: Defense 3: done"
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -317,6 +317,9 @@ event "remnant: surveillance end"
 
 
 conversation "remnant key stones"
+	branch found
+		has "Remnant: Found Keystones: offered"
+	
 	`As you are walking through the spaceport, a man comes up to you and introduces himself as the manager of the local outfitter. "Greetings, outsider," he says. "I have heard many good things about you and your assistance to our people. Would you be interested in going on a quest for me?"`
 	choice
 		`	"What kind of quest?"`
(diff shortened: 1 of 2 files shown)


Amazinite:
An alternative to my change to Mission.cpp would be to extend `branch` to be able to check stuff other than conditions, in this case being able to use `require` under a `branch`. If that was the case then I could just have the branches in the conversation check all the `to offer` conditions plus whether the player has a quantum keystone on them without needing to check on offer for a mission if that specific mission is offered.

tehhowch:
See https://github.com/endless-sky/endless-sky/issues/3341 for some related discussion.

One thought I've had about handling that wouldn't really help `branch`, as it would make the `to ____` nodes a combination of a ConditionSet (what branch uses) and some other functionality.

Amazinite:
I could change this to still work as intended without the source code change by having the `on offer` of the necessary missions add a condition to the player's save file that is then used in the conversation branches. Given that the outcome of every mission is either you accept it or you decline it, repeatedly incrementing this condition through reloading your save wouldn't cause any issues.

Amazinite:
This PR no longer makes changes to the source code, instead opting to increment the "offered" conditions of these specific missions through existing mechanics. This causes these missions to fall under the "missions which modify conditions and then use those same conditions in the conversation identified in that same `on offer` node" mentioned in #3746.

Amazinite:
Looks like that doesn't work.

Amazinite:
Decided to just take the simple route. Key Stones now has its own conversation, since that mission has no conditions outside of its own offering that can be used to determine which conversation should be offered. (Hai) and (Pre-Hai) 1 are still able to use a single conversation because they can be differentiated by "First Contact: Hai: offered", and (Pre-Hai) 2 can be differentiated by "Remnant: Found Keystones: offered".

PR #4507, opened 2019-09-01
Changed the names and descriptions of bounty missions

Naminson:
Something that's always annoyed me about the bounty hunting missions is that one that requires you to hunt down a leviathan and one that requires you to fight a firebird have the exact same name and descriptions, so the only way you can figure out which ones you should take early on is by looking at nothing but the payment.

This changes the descriptions of all bounty hunting jobs so they give you a rough description of what you're going to be fighting.

### DIFF
--- a/data/jobs.txt
+++ b/data/jobs.txt
@@ -2554,8 +2554,8 @@ mission "Recycle garbage"
 
 
 mission "Bounty Hunting (Small)"
-	name "Wanted ship near <system>"
-	description "A pirate vessel named the <npc> has been attacking merchants near the <system> system. Destroy it and return to <planet> for payment (<payment>)."
+	name "Wanted privateer near <system>"
+	description "A low-end pirate warship named the <npc> has been attacking merchants near the <system> system. Destroy it and return to <planet> for payment (<payment>)."
 	repeat
 	job
 	to offer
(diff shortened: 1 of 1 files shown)


Amazinite:
How about not using "medium" and just using small and large? So for bounty hunting, you'd see descriptions of either a "small pirate vessel," a "pirate vessel," or a "large pirate vessel," the intermediate size simply having no descriptor.

Naminson:
I've changed it to have just large and small, with size given for medium, like what Amazinite suggested.
I should mention I'm not too sure about the use of the word "small" to describe a medium warship. I think I might make it so instead of small > pirate > large, it goes pirate > large > very large, which would make the threat levels a little more clear.

Fzzr:
Maybe change the mission display names to indicate when it's a fleet as well?

Naminson:
Changing the names was a good idea.  I think it's in a good state now. For an example of how the bounty missions show up now:

[image](https://user-images.githubusercontent.com/37272086/64083673-2955c580-ccd8-11e9-921b-c7044a5b0045.png)

data/jobs.txt
```diff
@@ -2634,8 +2634,8 @@ mission "Bounty Hunting (Medium)"
 		dialog phrase "generic bounty hunting payment dialog"
 
 mission "Bounty Hunting (Big)"
-	name "Hunt down the <npc>"
-	description "A pirate vessel named the <npc> has been attacking merchants near the <system> system. Destroy it and return to <planet> for payment (<payment>)."
+	name "Very large wanted ship near <origin>"
```
Amazinite: This name looks dangerously long.

Naminson:
The long names should be resolved now. I tested at Gamma Corvi and Kornephoros (the longest system names I could find) and all of the job names and descriptions fit comfortably inside the window.

PR #4510, opened 2019-09-03
Change "require <outfit> 0" logic to check player cargo.

MageKing17:
Also, don't bother counting how many are available if we only need to know if one exists; just `return false` the moment one is found.

@beccabunny complained on Discord about `require <outfit> 0` not working right, and after much investigation, I'm reasonably certain it was caused by the lack of a `player.Cargo()` check, since (as the comment in the commit points out) landed ships all pool their cargo into the player's cargo.

(This is my first PR for the project; if I've overlooked something obvious, or missed something in the code style guide, let me know.)


tehhowch:
Ref #4002 for related discussion

MageKing17:
Closing in favour of the more-comprehensive #4561.

Issue #4326, opened 2019-06-01
Uninhabited planets should be unable to fine by default


Amazinite:
Ref: https://github.com/endless-sky/endless-sky/issues/3807#issuecomment-401572980

Fzzr:
Should this mean changing the default security from `.25` to `0.` or using the existing `IsInhabited` logic to check for fines?

PR #4512, opened 2019-09-05
Uninhabited planets are unable to fine

Amazinite:
Ref: #4326

Adds an additional check to PlayerInfo::Fine to see if the planet is inhabited or not. If the planet is uninhabited, it can't fine the player for illegal outfits.
Also removed `security 0` from all uninhabited planets since it is redundant. Notably, this should also make it so that colonized Wanderer worlds will now fine the player, since the events in the Wanderer campaign right now don't change the security of colonized planets.

### DIFF
--- a/source/Planet.h
+++ b/source/Planet.h
@@ -68,6 +68,10 @@ public:
 	// have the "uninhabited" attribute).
 	bool IsInhabited() const;
 	
+	// Check if the security of this planet has been changed from the default so
+	// that we can check if an uninhabited world should fine the player.
+	bool SecurityChanged() const;
+	
 	// Check if this planet has a shipyard.
 	bool HasShipyard() const;
 	// Get the list of ships in the shipyard.
@@ -146,6 +150,7 @@ private:
 	double bribe = 0.01;
 	double security = .25;
 	bool inhabited = false;
+	bool securityChanged = false;
 	// Any required attributes needed to land on this planet.
 	std::set<std::string> requiredAttributes;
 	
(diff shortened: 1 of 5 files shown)


Fzzr:
`planet Ruin` has both `uninhabited` and `security 0`.

tehhowch:
The referenced issue mentions "by default", but this PR makes it "at all"

Specifying a non-0 security value has historically meant that landing on the planet could result in a fine; should we preserve this current behavior for mod compatibility?

(The check for that situation requires first toggling a flag in `Planet::Load` to say that a user-supplied security value is being used, and then combining that flag check with the `IsInhabited` call)

Amazinite:
>The referenced issue mentions "by default", but this PR makes it "at all"

In what case would an uninhabited planet ever be able to fine the player?

tehhowch:
I'm sure a plugin can come up with a use case. Abandoned security systems, etc.

jafdy:
Locking out options is pointless, in many ways it makes more sense to just `security 0` any planet that shouldn't fine.

Amazinite:
That should would.

source/Planet.h
```diff
@@ -68,6 +68,10 @@ class Planet {
 	// have the "uninhabited" attribute).
 	bool IsInhabited() const;
 	
+	// Check if the security of this planet has been changed from the default so
+	// that we can check if an uninhabited world should fine the player.
+	bool SecurityChanged() const;
```
tehhowch: I think a better name is in order, e.g. `HasCustomSecurity` (default false) or `HasDefaultSecurity` (default true).

Amazinite: I'll go with the former.

Issue #4268, opened 2019-04-27
Update the status overlay to show the level at which a ship is disabled


Amazinite:
Ref: https://github.com/endless-sky/endless-sky/commit/e2e8898a4e154cf448c4ca0e6cc0eea69182c257

While the HUD shows this information, the status overlay does not.

![Screenshot_1](https://user-images.githubusercontent.com/17688683/56855073-91e9fb80-690e-11e9-8480-9de4f0b9d4cb.png)

tehhowch:
Feel free to propose different color options for the disabled hull for friendly and hostile. I can add the code to draw distinct colors for disabled hull vs regular hull tomorrow, but won't add a color change by default.

Amazinite:
How about something like this? Middle right and bottom right cells being the new colors. Could probably be better. 
Columns: shields, hull, disabled hull.
Rows: HUD, friendly overlay, hostile overlay.
![overlay colors](https://user-images.githubusercontent.com/17688683/56861325-bc1ed600-696d-11e9-9fb7-d5e139d1d983.png)

Friendly disabled: .6 .4 .2
Hostile disabled: .6 .4 .0
I noticed in the code though that the disabled hull color from the interface file is placed on top of the hull color; while the disabled hull color for the HUD is .3 .0 .0, the actual color seen is .75 .42 .29, as the .3 .0 .0 is being placed on top of the hull's .7 .62 .43 with an alpha of .3.

PR #4513, opened 2019-09-05
Status overlays now show the disabled threshold

Amazinite:
Ref: #4268

This is how it looks in game with the currently selected colors. 
image


jafdy:
Those colours are close to invisible for enemy ships, if those were moving I don't think I would ever be able to spot them.

tehhowch:
The overlay nature makes closely related colors hard to spot, especially on a busy background. Probably the disabled colors will need to be wildly different to be useful

Amazinite:
>Those colours are close to invisible for enemy ships

I guess that's why the colors can be customized. :)

>Probably the disabled colors will need to be wildly different to be useful

The disabled hostile color is already pretty bright compared to the normal hull color, but I'll up the brightness of the disabled friendly color. How does this look? (.5 red instead of .4.)

image

Issue #4223, opened 2019-03-10
Allow weapons to consume more than one unit of ammo per shot


Amazinite:
Currently, the `ammo` attribute is defined as
>if specified, an outfit which provides ammunition for this weapon. Each time it is fired, **one outfit of that type** is removed from your ship.

Suggested that a number following the name of the ammo type allow the weapon to consume that number of outfits of ammo upon being fired.

E.g. The following would consume four outfits of gatling gun ammo each time the weapon is fired.
```
ammo "Gatling Gun Ammo" 4
```

jafdy:
Out of curiosity, what kind of thing are you expecting this to be used for?

Amazinite:
I would typically expect this to be used when various weapons have a shared ammo type. Using the example already specified of Gatling gun ammo, a Gatling gun turret with two Gatling guns may fire two Gatling gun submunitions at once instead of having a doubled fire rate, meaning that the turret must consume two Gatling gun ammo at the same time. Given that Gatling gun ammo already exists, it makes more sense for such a weapon to consume two of it at once than to create a new ammo type just for the turret.

ghoulavenger:
Just to play devil's advocate, couldn't the same effect be had just by increasing the fire rate (twice as many guns, so fire twice as many shots)? I'm not saying that it's a bad idea, but I don't feel like it's overly necessary if all you want to do is use shared ammo types. The proposed change feels more to me like you'd want to be able to change the behavior of an individual weapon (like using more energy to overcharge a laser which could be done with variants, like the modified blaster).

Amazinite:
Fire rate can only be increased so much. Once you hit the cap, you can simulate higher fire rates through the use of submunitions, but that’s when you run into the scenario where you need to be consuming more than one ammo per shot, as each shot is actually firing more than one projectile.

Amazinite:
An example I used elsewhere is that the inclusion of this mechanic could allow for more novel weapons, such as a set of aliens weapons that all use the same ammunition, but they require different amounts of the ammunition per shot. Simply saying “just increase the fire rate” means that the more powerful weapons that use more of this ammunition can only increase in fire rate to increase in power and ammo usage, but allowing the use of more ammo per shot means that we could design a weapon that uses say 5 ammo per second by firing 5 times a second, while another weapon may consume 30 ammo per second by firing once every 2 seconds and consuming 60 ammo per shot. That is a more interesting scenario than simply upping the fire rate every time.

PR #4514, opened 2019-09-05
Extended the ammo attribute to specify how much ammo is used per shot

Amazinite:
Ref: #4223

Using the syntax from #4223: `ammo <outfit name> <number of outfits consumed>`. Putting no number is the same thing as putting one. 0 ammo usage is allowed, but the ammo outfit must be installed in order to fire.

This displays in game as a new line underneath ammo called "ammo usage". Ammo usage will only appear on weapons that consume something other than 1 ammo per shot.
image

### DIFF
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -71,7 +71,7 @@ void Outfit::Load(const DataNode &node)
 		else if(child.Token(0) == "weapon")
 			LoadWeapon(child);
 		else if(child.Token(0) == "ammo" && child.Size() >= 2)
-			ammo = GameData::Outfits().Get(child.Token(1));
+			ammoPair = make_pair(GameData::Outfits().Get(child.Token(1)),0);
 		else if(child.Token(0) == "description" && child.Size() >= 2)
 		{
 			description += child.Token(1);
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -71,7 +71,12 @@ void Outfit::Load(const DataNode &node)
 		else if(child.Token(0) == "weapon")
 			LoadWeapon(child);
 		else if(child.Token(0) == "ammo" && child.Size() >= 2)
+		{
+			// Non-weapon outfits can have ammo so that storage outfits
+			// properly remove excess ammo when the storage is sold, instead
+			// of blocking the sale of the outfit until the ammo is sold first.
 			ammo = make_pair(GameData::Outfits().Get(child.Token(1)), 0);
+		}
 		else if(child.Token(0) == "description" && child.Size() >= 2)
 		{
 			description += child.Token(1);
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -71,7 +71,7 @@ void Outfit::Load(const DataNode &node)
 		else if(child.Token(0) == "weapon")
 			LoadWeapon(child);
 		else if(child.Token(0) == "ammo" && child.Size() >= 2)
-			ammoPair = make_pair(GameData::Outfits().Get(child.Token(1)),0);
+			ammoPair = make_pair(GameData::Outfits().Get(child.Token(1)), 0);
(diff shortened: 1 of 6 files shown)


tehhowch:
One reservation here (vs the suggested syntax in the issue) is that if we allow using different ammo types for a given launcher, this syntax would mean we consume those ammo types equivalently, e.g. 2 "cheap ammo" would be as equal as 2 "rare expensive ammo" when it comes to firing.

The syntax in the issue (`ammo <outfit> <n>` would already be compatible with that world.

Amazinite:
I can make an attempt at `ammo <outfit> <n>`, but no promises.

Amazinite:
Done. 👍

source/Weapon.cpp
```diff
@@ -55,7 +55,12 @@ void Weapon::LoadWeapon(const DataNode &node)
 		else if(key == "sound")
 			sound = Audio::Get(child.Token(1));
 		else if(key == "ammo")
-			ammo = GameData::Outfits().Get(child.Token(1));
+		{
+			if(child.Size() <= 2)
```
tehhowch: Compute the count first with a conditional operator and then use just one `make_pair` call

source/Outfit.cpp
```diff
@@ -71,7 +71,7 @@ void Outfit::Load(const DataNode &node)
 		else if(child.Token(0) == "weapon")
 			LoadWeapon(child);
 		else if(child.Token(0) == "ammo" && child.Size() >= 2)
-			ammo = GameData::Outfits().Get(child.Token(1));
+			ammoPair = make_pair(GameData::Outfits().Get(child.Token(1)), 0);
```
tehhowch: Should this line also compute a count? Or be removed entirely?

Amazinite: This line is here because ammo storage outfits need to know which ammo they’ll remove if they’re sold and the ship would become overcapacity. I set the count to 0 because unless ammo is specified as a child of weapon, it doesn’t actually use the ammo.

tehhowch:
Should we handle the use case where a weapon is specified to consume 0 of an ammo? E.g. "You can only fire this weapon while you have this other outfit installed"

Example: a "Gamma Burst" weapon that requires you to have a Nuclear Missile installed, but once you fire that missile, the Gamma Burst can no longer fire either.

It's possible that fully supporting that case should be a separate PR, but I think in general the ammo consumption should be reported even if the weapon consumes 0 of its ammo

Amazinite:
>It's possible that fully supporting that case should be a separate PR, but I think in general the ammo consumption should be reported even if the weapon consumes 0 of its ammo

Shouldn't be difficult to add to this PR; just change the `max(1,usage)` to `max(0,usage)`.
As for showing the ammo usage line, my rationale behind hiding it for weapons with the default consumption is for the purpose of keeping new game mechanics hidden early on. It'll be more exciting to find weapons that consume more than one ammo per shot or even no ammo per shot if the game never made it clear in the first place that the ammo per shot can change. Think of it like how we don't show `disruption damage / shot 0` on every weapon; you don't know disruption exists until you find it.

Amazinite:
😈 
image

Amazinite:
For now an ammo usage of 0 still shows as `ammo usage: 0` in the outfitter.

source/Outfit.cpp
```diff
@@ -71,7 +71,7 @@ void Outfit::Load(const DataNode &node)
 		else if(child.Token(0) == "weapon")
 			LoadWeapon(child);
 		else if(child.Token(0) == "ammo" && child.Size() >= 2)
-			ammo = GameData::Outfits().Get(child.Token(1));
+			ammo = make_pair(GameData::Outfits().Get(child.Token(1)), 0);
```
tehhowch: Let's add a comment here just so no one else gets curious like I did.

Amazinite: Looks like this information isn't on the wiki either.

PR #4519, opened 2019-09-08
Protection attributes

Amazinite:
Ref: https://github.com/endless-sky/endless-sky/commit/2bf7c0f089d72aef14a301bbdd09b9639f17c1e0, #2719

These protection attributes do the same thing that the resistance attributes use to do. 
All forms of damage have been given a protection attribute that reduces their effect according to the equation `damage / (1 + protection)`. Forms of protection include:
* Shield
* Hull
* Heat
* Ion
* Disruption
* Slowing
* Hit Force
* Piercing

Why have an attribute for each damage type? Because we can. I was originally just going to do ion, disruption, and slowing since those also have resistances, but then I thought about how more than just those attributes have some form of resistance to them currently.

To note, the following attributes already have a resistance to them (something where small amounts of these damage do nothing but amounts past a certain threshold do yield results):
* Shield (shield generation)
* Hull (hull repair)
* Heat (cooling, kinda)
* Ion
* Disruption
* Slowing


Pointedstick:
Code is fine :+1:

Amazinite:
Do we really, though? :)

Zitchas:
Amazinite raises a good point: Couldn't we have ships that are deeply flawed in some way that actually have a vulnerability to certain types of damage.

For example, an experimental ship testing out a new energy system might be vulnerable to ionizing damage because it hasn't been hardened yet.

Or perhaps oddly specialized ships (say, a sun-skimming research vessel) that has extremely high heat protection, but is actually vulnerable to slowing because the engines are a lower priority than they would normally be.

Pointedstick:
My thoughts exactly! :)

Amazinite:
>Amazinite raises a good point

The reason behind preventing negative attributes is because a value of -1 will cause a divide by zero error. I could change the check to only catch values of -1 instead of <0, but if what you want is to be able to cripple a ship, I think that a different attribute is in order instead of using what is unintended behavior with these attributes.

When MZ originally committed this mechanic, [I asked why he didn't just use a multiplier](https://github.com/endless-sky/endless-sky/commit/2bf7c0f089d72aef14a301bbdd09b9639f17c1e0#commitcomment-22955495), as a multiplier would allow for 100% resistance, while this mechanic only allows for close to 100% resistance by just throwing a super high value at it. A multiplier would also allow you to more easily cripple a ship rather than having to deal with negative values between -1 and 0 in order to get the crippled effect. The thing with a multiplier is that it isn't as scaleable, since as Elyssaen mentioned, four outfits with a 25% multiplier and you have 100% resistance. Perhaps a multiplier is in order though as an attribute intended only for ships and not for outfits, although that sounds like a bit of a headache to me given that each damage type would then have a resistance, a protection, and a multiplier attribute. And this is exactly what I said I was afraid of with #4525, just opening the door to give every attribute a multiplier.

Zitchas:
I appreciate the explanation. I'm fine with not having negative values for now, but exploring ways of introducing such vulnerabilities would be welcome.

The downside to the "just use weaker engines" method is that a ship with bad engines is just a slow ship. A ship that is vulnerable to slowing could be a very fast ship that has *fragile* engines. (like, say, racecars. Tend to have very good engines. Often very, very finicky engines.) Goes great until it gets hit by slowing, and then does *terribly*

Amazinite:
In order to close the Pandora's box of multipliers, I've changed the check to disallow <= -1 values instead of < 0 values. Since < -1 values would heal the impacted ship, I've made those off limits due the the following:
* How do you kill a ship that heals from hull or shield damage? 
* Last time I checked, negative slowing damage values lead to very strange behavior that just looking at it feels like the game is about to crash.
* Negative disruption damage similarly leads to strange effects that I've seen heal ships.
* It looks like two years ago when resistance was changed, negative slowing, ionization, and disruption values were prevented from happening anyway.
* Negative piercing damage is akin to negative disruption.
* How would a ship generate fuel out of mid-air when being hit by fuel damage?

Zitchas:
Sounds like fun. So long as it is predictable and doesn't actually break the game, at least. :)

data/tooltips.txt
```diff
@@ -68,6 +68,9 @@ tip "cooling inefficiency:"
 tip "cost:"
 	`Cost, in credits. If you have sold any used (depreciated) items today, you can buy them back for the same price you sold them for.`
 
+tip "disruption protection:"
+	`Protection against weapons that have a disruption effect. A value of 1 cuts the effect in half, and 2 cuts it to a third of what it would otherwise be.`
```
tehhowch: Should these tips say `... total value of 1...`?

source/Outfit.cpp
```diff
@@ -87,7 +99,18 @@ void Outfit::Load(const DataNode &node)
 				licenses.push_back(grand.Token(0));
 		}
 		else if(child.Size() >= 2)
+		{	
 			attributes[child.Token(0)] = child.Value(1);
+			// These outfits are not allowed to have values <= -1. Catch any
+			// violations upon loading the game so that the corrections don't
+			// need to be repeatedly calculated later.
+			if(PROTECTION_TYPES.find(child.Token(0)) != PROTECTION_TYPES.end())
```
tehhowch: Just use `count`

source/Outfit.cpp
```diff
@@ -87,7 +99,18 @@ void Outfit::Load(const DataNode &node)
...
+				if(child.Value(1) <= -1.)
+				{
+					child.PrintTrace("Skipping illegal negative attribute:");
```
tehhowch: If this guard is before the attribute gets set, we avoid removing a base or plugin value if a later plugin sets an unsupported value.

data/tooltips.txt
Amazinite: It's the same language as the cargo and outfit scan power, so if these get changes then those should be changed too.

Amazinite:
> Should these have special formatting in the outfitter?

How would you suggest they possible be formatted differently? I could think of having it display the percentage of damage ignored, but that wouldn't exactly show the truth because of the diminishing returns. Ref: #4215.

source/Outfit.cpp
```diff
@@ -87,7 +99,20 @@ void Outfit::Load(const DataNode &node)
 				licenses.push_back(grand.Token(0));
 		}
 		else if(child.Size() >= 2)
-			attributes[child.Token(0)] = child.Value(1);
+		{
+			// These attributes are not allowed to have values <= -1. Catch any
+			// violations upon loading the game so that the corrections don't
+			// need to be repeatedly calculated later.
+			bool illegalAttribute = false;
+			if(PROTECTION_TYPES.count(child.Token(0)) != 0)
```
tehhowch: Sorry, I meant you can just use the implicit boolean conversion of `set::count`.

Amending my other suggestion, something like
```c++
const string &key = child.Token(0);
attributes[key] = PROTECTION_TYPES.count(key) ? max(-.99, child.Value(1)) : child.Value(1);
```
would do the job.

Amazinite:
Since a value of 1 results in such a massive change in incoming damage (just 1 results in a 50% reduction in damage), how does it sound if we change it from `damage / (1 + protection)` to `damage / (1 + protection / 100)`? This makes it so that a value of 1 is only about a 1% reduction in incoming damage. The reason for doing this would be to avoid any outfits with this attribute needing tiny decimal values (which would probably get rounded away) just to get a reasonable effect.

source/OutfitInfoDisplay.cpp
```diff
@@ -26,11 +26,12 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	const vector<pair<double, string>> UNIT_PAIRS = {
+	const vector<pair<double, string>> SCALE_LABELS = {
```
Amazinite: I changed the name to something more descriptive.

Amazinite:
Ref: 3e8e760
It looks like if we want to check for values <= -1 in Outfit::Load, we also need to check for them in Outfit::CanAdd, as Outfit::Load is only preventing individual outfits from having values <= -1, not for a ship's total attribute values from being <= -1. If anything, only Outfit::CanAdd should be what needs changed.

Amazinite:
With the latest commit it's notable that if you have an outfit with a negative protection value, you are prevented from going below a total of -.99, but for some reason the notice as to why you can't install another outfit past that point doesn't always pop up. For example, a value of -0.5 will cause the dialog box to appear, but -.3, -.25, -.2, -.1, and -.01 all don't show the notice.

Amazinite:
Got it working now. 👍

Issue #4517, opened 2019-09-08
Heat, Energy, and Fuel drain attributes for Resistance


Amazinite:
**Is your feature request related to a problem? Please describe.**
It was described in #2719 how the current resistance attributes are very similar to hull repair and shield generation in how they allow for a ship to shrug off a small amount of damage, but after the damage has reached a certain threshold then it will start having an effect. It's notable though how currently the resistance outfits are lacking in certain properties, such as the ability to alter the heat, energy, and fuel of the ship when active.

**Related Issue Links**
#2719

**Describe the solution you'd like**
Add `ion resistance heat`, `ion resistance energy`, `ion resistance fuel`, and similar attributes for disruption and slowing resistance.

**Additional context**
It's notable that the current implementation of the resistance attributes does NOT check to see if the ship is disabled before applying them. This seems to imply that resistance attributes are some sort of ability intrinsic to the hull or shields of the ship that doesn't require any energy to power. 

https://github.com/endless-sky/endless-sky/blob/e594c2e88b5586533af21f03086755d98a4f5dc2/source/Ship.cpp#L1626-L1631

Adding the suggested attributes would mean that a ship needs to not be disabled in order for them to be active, but I can see an argument for still having some way for resistance to apply to a disabled ship. 

In order to both keep the current mechanic and implement the suggested mechanic, what if resistance that doesn't drain energy, heat, or fuel will apply if disabled, but if the resistance comes with any of those attributes, it doesn't work when disabled? The one issue with this would be if a ship comes standard with resistance that has no drain, but then the player installs an outfit with resistance and drain, meaning the ship's stock resistance would no longer work when disabled, so it may just be better for resistance to never work when disabled.

Zitchas:
Would it be possible to have values as well as an attribute, so the values act as per normal, and the attribute is limited to a few states. 

Ex. 
0 = always disabled (this ship or any ship with this outfit can never have this type of resistance, 
1 = this resistance is always active regardless of whether the ship is active or disabled, 
2 = this resistance is only active when the ship is active, 
3 = this resistance is only active while the ship is disabled, etc. 

I could see having a second or third range of these variables that specify that said outfit/ship will/will never override said attribute when another source of said resistance is provided. Basically, a priority system for determining which has priority.

Amazinite:
>I could see having a second or third range of these variables that specify that said outfit/ship will/will never override said attribute when another source of said resistance is provided. Basically, a priority system for determining which has priority.

[...]

PR #4520, opened 2019-09-09
Added energy, fuel, and heat attributes for resistances

Amazinite:
Ref: #4517

I *think* this works. It compiles, at least. I implemented this by using a modified DoRepair function that I called DoStatusEffect. It first calculates how much resistance could possibly be needed given what the resistance attribute and current status effect are, then limits the needed resistance further by energy and fuel costs if necessary. Then, given how much resistance was done, it updates the energy, fuel, and heat of the ship. **Resistance attributes will not be applied if the ship is disabled,** just like shield generation and hull repair. This is the opposite of how resistance is currently applied, where it will affect the ship if disabled.

Given that I did this using a modified DoRepair function, and DoRepair is used to repair carried ships after the mothership is repaired, I left a TODO comment asking if the mothership should also impart its left over resistance attributes onto the carried ships. I'm pretty sure the DoResistance function would need to be altered slightly to allow that, though.


Amazinite:
Note that these attributes are not displayed anywhere on the ship stats because there's no room for them. Ref: #4215.

source/Ship.cpp
```diff
@@ -86,6 +86,39 @@ namespace {
...
+		if(fuelCost)
+			resistance = min(resistance, fuel / fuelCost);
+		
+		// Update the energy, fuel, and heat given how much resistance is being used,
+		// then update the stat.
+		if(resistance > 0.)
+		{
+			energy -= resistance * energyCost;
+			fuel -= resistance * fuelCost;
+			heat += resistance * heatCost;
```
tehhowch: We should account for negative heatCost, i.e. being required to supply heat to use the resistance.

source/Ship.cpp
```diff
@@ -1625,13 +1658,35 @@ void Ship::DoGeneration()
 		if(shieldsAvailable)
 			heat += (shieldsAvailable - shieldsRemaining) * shieldsHeat;
 	}
+	
 	// Handle ionization effects, etc.
+	// TODO: Mothership gives status resistance to carried ships?
 	if(ionization)
-		ionization = max(0., .99 * ionization - attributes.Get("ion resistance"));
+	{
+		const double ionResistance = attributes.Get("ion resistance");
```
tehhowch: Making these const here, but not in the method called with them, is meaningless. With the new reduced scope, you can discard the const qualifier (or update the called method's signature).

Zitchas:
If the other resistances apply even when a ship is disabled, why shouldn't these ones?

Ideally, I'd say there should be a companion attribute that determines whether the resistance does or does not apply. I think probably the ideal would be to have this companion value be a multiplier that is applied to the resistance value when disabled. The default value would be 0. So, when disabled, the resistance stops applying. Any passive thing that doesn't need a functioning ship to keep working would have that value set as 1. If a thing is some part passive, part active thing that loses some functionality but stays functional, it could have a value somewhere in between. Negative values could indicate a device that malfunctions (or a ship with an innate weakness that becomes vulnerable when disabled) could have a negative value. A ship that for whatever reason is less vulnerable to something when disabled could have a value greater than 1.

From a lore perspective, if something is an active system, (say, some kind of field the ship projects to reduce the effect of some kind of damage), then yes, it shouldn't apply when disabled. But if it is a passive system (or specifically designed to keep functioning without the support of the rest of ship), then it should keep functioning.

It seems rather arbitrary to decide in the mechanics that any particular resistance must or must not be an active or passive system, when realistically they could be either.

Amazinite:
> If the other resistances apply even when a ship is disabled, why shouldn't these ones?

The only other resistance-like attributes that exist outside of these here are are shield generation, hull repair, and cooling, and the former two of those don't work if disabled, so I'm not sure what you mean by "the other resistances."

Zitchas:
I think I misread something in your earlier post about some resistances working when disabled, but these ones won't work when disabled.

Amazinite:
It's the case that right now in the master, these three resistances are passive (i.e. work when disabled), but this PR makes them active (i.e. must be enabled).

Give me a few minutes though. A few extra lines and I can have attributes determine that. (Although, I believe you'll get bad results if you mix the costs with passive resistances, as I attempted to create passive hull repair/shield generation once and it didn't work.)

Amazinite:
Damn is this cutting it close.
image
Any wording that would make this shorter?

Zitchas:
Resist disruption while disabled is slightly shorter.

Amazinite:
I want to avoid that wording as it suggests that the effect is only resisted when disabled. "Even if" makes it clear that the resistance is applied while enabled, but while disabled as well.

Zitchas:
Yes, that does make sense. "Even if" is pretty clear. (Even if is also shorter, I think. Your picture shows "even when" though)

Amazinite:
For whatever reason, GitHub doesn't let you link issues and PRs sometimes. The search bar just doesn't show the issue or PR you want to select. Randomly commenting here to see if it's based on activity and only shows some number of recently active issues and PRs.

Amazinite:
That answers that question.

Issue #4406, opened 2019-07-03
Shield hardness / piercing mitigation proposal and request for comment (RFC), v0.9.11+


ChamEV:
## Shield Hardness ##
**Shield hardness** directly counteracts piercing and disruption.  If a ship has a shield hardness of 1.15, then korath piercers with 0.2 piercing will instead have 5% of damage skip through shields.  1.15 - 0.2 = .05  Typical values of shield hardness would be between 0 and 2, with 1 being default.

Two outfit attributes would need to be associated with this.  (Names need work.)
**Shield hardness increase.** how much this outfit increases the ship's shield hardness.
**Shield hardness cap.** the maximum this outfit can increase the ship's shield hardness.

Piercing is currently capped to between [0,1], uncapping this and shield hardness would enable specialized encounters where a very special gun with 1000 piercing is needed to kill a very special baddy with 1 million shields and regen and 999 shield hardness.  Hopefully if someone does this, they have a good story for it.

In vanilla, we might see some future or existing alien ships like Quarg, Drak, pug, with shield hardness values between 1.2 and 2.  Keep in mind shield hardness does not mitigate regular shield damage in any way, just prevents skipping it.

An example with disruption.  First recall that disruption makes any weapon damage pierce.

> `"disruption damage"`: works like ionization, adding a "shield disruption" effect that fades by 1% each frame. When disrupted, your shields only block 1 / (1 + .01 * disruption) of weapon damage, and the rest pierces through your shields and damages your hull. For example, if a ship has accumulated 10 disruption, about 9% of damage will leak through to the hull. (v. 0.9.0) 

With shield hardness, a ship will need to acquire enough disruption before any damage can pierce.  While the separate existing stat disruption resistance still removes disruption faster.

Required D disruption to achieve B blocking with H hardness.  Piercing is 1 - B.
D = 100*(H/B - 1)

Effective Piercing | Required Disrution 1.2 hardness | Required Disrution 1.0 hardness (default)
-------|----------|------------
0% | 20 | 0
10% | 33.3 | 11.11
20%|50 | 25
40%|100 | 66.66
50%|140 | 100
75%| 380 | 300
99%|11900 |9900

---

## Shield and Hull generation after delay ##
Moved to: https://github.com/endless-sky/endless-sky/issues/4408

## Shield and Hull Recharge multiplier ##
Moved to https://github.com/endless-sky/endless-sky/issues/4407

ghoulavenger:
[...]

PR #4521, opened 2019-09-09
Added piercing resistance

Amazinite:
Ref: #4406

As described in #4406, this attribute subtracts from the piercing of the projectile that is damaging the ship. I went with the name piercing resistance since it fits with the other resistances in game, and has a similar effect of requiring the incoming damage to reach a certain threshold before taking effect.
This PR removes the restriction on piercing needing to be below 1, but adds a check in Ship.cpp so that piercing values above 1 do not have any effect.


source/OutfitInfoDisplay.cpp
```diff
@@ -26,50 +26,52 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
-		{"shield fuel", 60.},
-		{"shield heat", 60.},
-		{"solar collection", 60.},
-		{"thrusting energy", 60.},
-		{"thrusting heat", 60.},
-		{"turn", 60.},
-		{"turning energy", 60.},
-		{"turning heat", 60.},
+	const map<string, pair<double, string>> SCALE = {
+		{"active cooling", make_pair(60., "")},
```
Amazinite: There's probably a better way of doing this that I don't know about instead of having a bunch of make_pairs.

tehhowch: Since there will only be a few scale types, and many uses of those, you could define those pairs once in a vector, and then associate the attribute string to an index into that vector.

Amazinite: Pictures, please.

Amazinite:
I realize that I didn't add the disruption mitigation part of #4406. Given that we already have disruption resistance, I'm not sure if we should have piercing resistance double-dip in what it can combat.

Amazinite:
Piercing resistance energy/fuel/heat? 🤔

Amazinite:
Pushed the outfit info display changes right to the master since multiple PRs that I've opened recently need it.

Issue #4478, opened 2019-08-14
Optimal range attribute that overrides calculated range


Amazinite:
**Is your feature request related to a problem? Please describe.**
As noted by #1826, the current range calculation isn't always perfect. It doesn't take into account drag, acceleration, random lifetime, and random velocity. As I noted as well, the range calculation ignores the inaccuracy of the weapon being used, meaning that the range that the weapon will be fired at by the AI may be too far for the weapon to reliably hit. In this way, the current way that range is calculated makes the creation of more novel weaponry difficult.

**Related Issue Links**
#1826 The issue where I first brought this idea up (but the issue itself was not asking for this change).
#1894 A PR that attempted to add this mechanic, but also attempted to change how the current range is calculated. I suggest that the current range calculations remain the same, as the optimal range attribute allows the creator to set whatever range they want.

**Describe the solution you'd like**
Suggested that we have an "optimal range" attribute that overrides the calculated range of a weapon. This optimal range is what would be used by the AI and by auto-fire when determining when to fire the weapon. In game, this optimal range would simply show where the current range stat is, as the player does not need to know if the range of the weapon is being overrided. 

**Describe alternatives you've considered**
As mentioned in #1826, it's somewhat possible to manually override the range calculation by using "fake" submunitions, but, akin to the creation of various timer ships mentioned in #4475, this is a rather messy solution that requires the creation of new "fake" submunitions with every new weapon.
The other solution would be to make the range calculation be more accurate by taking drag, acceleration, random velocity and lifetime, and inaccuracy into account, but as highlighted in #1894, the inaccuracy of the current calculation has its uses and is used in game already. The creation of an optimal range attribute would allow the creation of weapons that require a more accurate range number while not ruining the existing weapons that benefit from the range calculation being "wrong."

Amazinite:
Also, minimum range attribute, anyone?

Bladewood:
As an example, this is the kind of weapon I'm looking for:
![prismaticlaser](https://user-images.githubusercontent.com/16506683/63027142-6584cb80-be7a-11e9-9cf9-ee70cc84ddbe.png)
I've tried to implement this in the current game many times. However, no amount of submunition trickery will make this weapon work without it horribly breaking in some way. (With projectile weapons it can be much easier to disguise the effect, but the uniquely "instantaneous" quirks of beam weapons are insurmountable.)
[...]

Issue #1826, opened 2016-11-18
Weapon range problem (affecting AI behaviour)


Elyssaen:
I'm having some trouble with some 'clever' submunition chains: the AI can't work out the range of some of these weapons.

```
double Weapon::TotalLifetime() const
{
	if(totalLifetime < 0.)
	{
		totalLifetime = 0.;
		for(const auto &it : submunitions)
			totalLifetime = max(totalLifetime, it.first->TotalLifetime());
		totalLifetime += lifetime;
	}
	return totalLifetime;
}

double Weapon::Range() const
{
	return Velocity() * TotalLifetime();
}
```

I think there are two problems with what this does:

1. The "random lifetime" attribute isn't supported.
2. TotalLifetime()'s calculation is otherwise logical, but Velocity is only based on the first munition. If it's slow and a submunition is fast, the AI won't know how to fire the weapon.

I can see 2 probably doesn't matter for the Thunderhead: it shows 3200 instead of 4000, which isn't important, and getting to 4000 would require the submunition be aware of the parent velocity. It does matter for a few weapons I'm trying to create, as it gives them a very low range in the AI's (and outfitter's) eyes.

[**Edit:** I've improved on the below, and have a post a bit further down where I _think_ I've cracked this, the right way, and have pasted the code changes I think do it. Would appreciate some feedback! I've left the below for reference to my thought process but I've since worked out some of it is wrong. : )] 

What would the consequences be of switching to

```
		totalLifetime += lifetime + 0.5 * randomLifetime;
```
then making the submunitions aware of their true velocity (blind guess: not easy) and having a block like this? 

```
double Weapon::Range() const
{
	if(Range < 0.)
	{
		Range = 0.;
		for(const auto &it : submunitions)
			Range = max(Range, it.first->Range());
		Range += Velocity() * lifetime;
	}
	return Range;
}
```

Alternatively – probably easier and _perhaps_ even better – could we add a "range" attribute to a weapon so you can _tell_ the code to override the basic calculation with a custom range?

If I can get some feedback, I'm happy to try writing whatever change seems best and make a pull request. Some weapons I'm trying to make are broken right now, so if no code changes look like a good idea I'd appreciate some help with the weapons instead. : )

– Elyssaen _(not experienced with C++, but enthusiastic)_

Amazinite:
[...]

PR #4522, opened 2019-09-09
Added velocity and range overrides

Amazinite:
Ref: #4478, #1894

`"range override"`: A manual range override. This will be what shows up in the outfitter for a weapon's range. **Effects when the AI fires the weapon versus how far the target is.** Also used to calculate a weapon's lifetime for the AI, estimating it with `range override / velocity` if range override is present. Otherwise the normal lifetime is used.
`"velocity override"`: A manual velocity override. This will help to change where the AI points the ship when firing. The main and only use of this that I could think of and tested was in using weapons with submunitions that had their own velocity, as currently the game does not take in to account any changes in velocity caused by submunitions. **The higher the velocity override, the straighter the aim.** I assume that this could also be used for weapons that have drag and acceleration that drastically alter the true velocity of a projectile after it is fired; if the projectile increases in speed, then give it a higher optimal velocity than the normal velocity. If the speed decreases, then do the opposite. **Note that AI.cpp adds half of the random velocity to a weapon's velocity when calculating where to aim.** Just keep that in mind if you're changing velocity override because of a random velocity.

Note: **This PR does not improve the accuracy of the existing range/velocity equations. It only provides manual overrides to produce different AI behavior.**

@Bladewood I'd like to see if this works for your weapons that needed these, as these few changes alone were all I needed to get my weapons to work that I mentioned all the way back in #1826.


Amazinite:
Cases that I tested:

This energy shotgun. `"optimal range"` is used to make it so that the weapon will only fire once the target is close enough that most of the projectiles will hit, and `"optimal velocity"` is used because of how the submunition increases the velocity over the base projectile.
```
outfit "Energy Blaster"
	category "Guns"
	cost 16000
	thumbnail "outfit/blaster"
	"mass" 5
	"outfit space" -5
	"weapon capacity" -5
	"gun ports" -1
	weapon
		sprite "projectile/blaster"
		sound "blaster"
		"hit effect" "blaster impact"
		"inaccuracy" 1
		"reload" 12
		"optimal velocity" 10.625
		"optimal range" 250
		"firing energy" 9.6
		"firing heat" 30
		submunition "Energy Blast" 12
	description "Although not the most accurate or damaging of weapons, the Energy Blaster is popular because it is small enough to be installed on even the tiniest of ships. One blaster is not enough to do appreciable damage to anything larger than a fighter, but a ship equipped with several of them becomes a force to be reckoned with."

outfit "Energy Blast"
	weapon
		sprite "projectile/blaster"
		"hit effect" "blaster impact"
		"inaccuracy" 15
		"lifetime" 48
		"velocity" 10.625
		"shield damage" 9.6
		"hull damage" 6
```

I also tested an Electron Beam that I gave an optimal range of 100, firing alongside a Beam Laser with no changes. As expected, the Electron Beam would only auto fire when at 1/3rd the range of the Beam Laser.

Zitchas:
As the `range override` was suggested to be used for the #4977 PR, is this something that is likely to get merged in the near future? I note that it used to have the 0.9.11 milestone, but currently doesn't have one.

I don't think it is crucial - we can adapt to other ways of limiting range for now. But if it is going to get merged in the next week or two then we might as well just wait for it.

Amazinite:
@Zitchas Recently GitHub added the ability to turn PRs into drafts after they've already been open. I recently went through all my PRs and turned some of them into drafts if they still need worked on. Everything else is likely to merge shortly after v0.9.12 (this PR included).

Zitchas:
OK, thanks for the insight. That will be helpful.

Issue #3448, opened 2017-12-20
Custom disable threshold value


Amazinite:
The following is a graph of the current disable point for a ship's hull. The X axis is the amount of hull and the Y axis is the hull remaining at which the ship is disabled.
![screen shot 2017-12-20 at 2 03 15 pm](https://user-images.githubusercontent.com/17688683/34223778-8f32994c-e58e-11e7-9887-b35b284e713d.png)

The following graph has hull on the X axis and on the Y axis is the % of hull left at the point of being disabled.
![screen shot 2017-12-20 at 2 16 20 pm](https://user-images.githubusercontent.com/17688683/34224287-6266af5a-e590-11e7-9f19-53496e7b310f.png)

These values are true across almost all ships, the only exceptions being ships that are never disabled. Aside from the fact that I find it odd that all species have their ships disable at the same % hull left beyond a certain point (and up into T2, practically all ships have enough hull that the 20% disable point kicks in right away), I think that being able to fine tune when a ship disables would be beneficial to plugin makers.  (For example, one could technically make a ship more powerful by lowering the point at which it disables instead of increasing its hull, and it would especially help for fine tuning as currently a plugin maker must, for the most part, raise the hull of a ship by 5 points to raise the disable threshold by 1 point.)

One or both of the following attributes would help. The naming can change as it's the functionality that is desired.
* `"disabled at"` = hull left when disabled (doesn't take into account the ship's hull being increased or decreased by an outfit)
* `"disable threshold"` = % of hull left when disabled (takes into account any hull that is gained or lost through outfits)

Hadron1776:
If I had to pick, `"disable threshold"` would be the way to go. 

Getting disabled at 96% hull because of outfits would suck.

KiLEdEnNis:
not really sure what ya all said but I'm for a custom disabled value, more modding freedom

PR #4523, opened 2019-09-09
Added attributes for custom disabled thresholds

Amazinite:
Ref: #3448

`"hull threshold"` adds to the point at which a ship becomes disabled. E.g. a hull threshold of 100 will make a ship that is normally disabled at 500 hull become disabled at 600.
`"threshold percentage"` is a value between 0 and 1 that changes the default point at which a ship becomes disabled as a factor of the maximum hull. This means that hull threshold will add onto the point defined by threshold percentage.
`"absolute threshold"` is a hull value that determines when the ship becomes disabled and is unaltered by the above two attributes. E.g. an absolute threshold of 1000 will mean that even if you have a hull threshold of 100, the point that the ship is disabled at will remain at 1000.


Zitchas:
I'm not one for checking engine code, but I like the idea and the usefulness.

Is this something that could be applied both to ships as a category, as well as to NPCs and unique instances?

It comes to mind that it'd be an interesting way to differentiate similar versions of ships (say, the engine marauder from a defenses marauder. The engine one would disable at a higher value because it has so much sensitive stuff out in the open.)

Amazinite:
Right now, "disabled" dictates the hull value at which the ship becomes disabled. Should it instead be a value that gets added to the default disabled threshold? E.g. say you have a ship that gets disabled at 1000 hull left. If you had an outfit with a disabled attribute value of 200, then the ship would then become disabled at 1200, instead of at 200 as it works now.

Zitchas:
OK, just thinking about this a bit, there seems like room for a few things.
Just to clarify, when talking about `generic disable` this is the calculated disable point that the game currently does. 

`default disable` = this is overrides the `generic disable` to provide the base value a ship disables at.
`disable mod` = this is a value for outfits or other things that is applied to the the base value a ship disables at. Probably a percentage change, but just a hull value could work too.
`disable override` = this overrides the absolute end value of the disable calculation. Nothing is added to it, and it completely replaces the calculated value.  If this value is present, then the game just skips even considering any `disable mod` or other disable related calculations. This could be a percentage or just a hull value.

For clarity, both `disable mod` and `disable override` could be split in two to have one attribute that is a percentage, and the other is a straight hull value. Alternatively, they could be set up so that any value between a certain range is taken as a percentage, and all other values are set amounts of hull. 

For instance, for disable mod, a value between -1 and +1 would be a percentage, and anything outside that is a hull value. For disable override, 0 to +1 would be the percentage, and anything outside of that would be a hull value.

In terms of usage, I would expect the `default disable` would be fairly common as a ship attribute, and `disable mod` would be used main on outfits. (although it'd be neat to be able to specify that on a government attribute that some faction makes fragile versions of otherwise normal ships, or particularly tough ones, or something.). 'disable override' would be very rare, and would likely be used mainly to label particularly delicate ships that for some reason can't be modded. (not really fond of this, but it could happen). Might also be useful for outfits such as the Emerald Sword's "broken drive" or the "Broken Jump Drive" to put bit more meat behind the idea that the ship is damaged, and that those parts need to be fixed. I could see someone  using it to make, say, an infestation outfit that is applied to a player's ship via story event that has to be resolved to be removed, or something. But generally speaking, I think the hard "override everything" should be quite rare.

Amazinite:
Sounds like your "default" is what I've suggested for "percentage," and your "mod" is what I've suggested for "disable," so all that be missing is an "override".

Zitchas:
OK, that works. Then there's just the "clarity in naming" aspect.

Amazinite:
I think "hull integrity" would be a good name to use.

Zitchas:
"hull integrity" sounds... Normal. I'd expect that to mean the default disable point. "Critical integrity" would make a good override...

And that made me think of another possible value for all this: Minimum (and maximum) disable amounts. So one could make a ship that could have higher disable points, but couldn't be any lower. Or conversely, a ship that could get a lower disable point, but couldn't get any higher. That could probably be achieved using an attribute with three options: default(no limit), 1 (only higher) 2 (only lower) and would be used in conjunction with the default value.

Amazinite:
Changed the names of the attributes, changed how they work to be like how I mentioned the other day, and added an "override" attribute as @Zitchas has mentioned.

data/tooltips.txt
```diff
@@ -164,6 +164,15 @@ tip "hull heat multiplier:"
 tip "hull repair rate:"
 	`Amount of hull strength that can be repaired per second. Repairs usually consume energy and may generate heat as well.`
 
+tip "absolute threshold:"
+	`The remaining hull strength at which the ship becomes disabled, unaltered by any other attributes.`
+
+tip "hull threshold:"
+	`Alters the hull strength value at which the ship will become disabled.`
```
Zitchas, line 171: Given that the "threshold percentage" mentions that it is a percentage, perhaps clarifying that this is an amount of hull and *not* a percentage might be good.

Zitchas:
Thanks. These names do look clearer now. 

copied from the code to make it easier for others to see:
```
tip "absolute threshold:"
	`The remaining hull strength at which the ship becomes disabled, unaltered by any other attributes.`

tip "hull threshold:"
	`Alters the hull strength value at which the ship will become disabled.`

tip "threshold percentage:"
	`Alters the percentage of the remaining hull strength at which the ship will become disabled.`
```

data/tooltips.txt
```diff
@@ -164,6 +164,15 @@ tip "hull heat multiplier:"
 tip "hull repair rate:"
 	`Amount of hull strength that can be repaired per second. Repairs usually consume energy and may generate heat as well.`
 
+tip "absolute threshold:"
+	`The remaining hull strength at which the ship becomes disabled, unaltered by any other attributes.`
```
Zitchas, line 168: Might want to clarify the description for "absolute threshold." It could be interpreted as indicating the base value before it is altered.

`The remaining hull strength at which the ship becomes disabled. This value overrides any effects that change the disabled threshold.`

data/tooltips.txt
Amazinite: I feel like this fact would be apparent in its usage, given that "threshold percentage" will always be a number between 0 and 1 (shown in game as 0 to 100%), whereas hull threshold will also be a number without any units that is likely to be >100 from the start.

data/tooltips.txt
Amazinite: Is that not already clear in the "unaltered by any other attributes" part?

Zitchas: Not really. Saying that this is the value unaltered just means that *this* value hasn't been altered. It doesn't specify whether or not any other attributes will modify it afterwards.

Issue #4408, opened 2019-07-03
Shield and Hull generation after a delay proposal and RFC v0.9.11+


ChamEV:
## Shield and Hull generation after delay ##
**Shield recharge delay after taking shield damage.**
**Shield recharge delay after shields go down.**
**Hull repair delay after hull damage.**

Minimum and default: 0 (all values in frames)
Outfits can alter each directly.  positive value increases delay, negative decreases delay.  Stacks additively.

Allows for more diverse ship and outfit design.

SolraBizna:
I was just thinking about this the other day. This would, IMO, provide a really interesting way to break up the current strata of shield system desirability.

I would say that the delays should be an attribute of individual generators, so that different generators might kick in at different times. For example, if basic shield generators worked on zero delay, but shield regenerators had a three-second delay, then a ship that had one of each would have a low level of constant recharge that sharply increased after three seconds of not being hit.

Amazinite:
Do you have shorter names for these? :)

SolraBizna:
Could even be `"repair delay" x y` to allow delayed recovery from being disabled, like EVN's repair bots...

Amazinite:
Inb4 a super strong shield generator with a shield delay of 1 so only continuous damage will actually do anything.

SolraBizna:
You mean like the "Systems Core (Large)" we already have? :)

Amazinite:
Shield generation and hull repair don't have a delay right now. They'll still repair even when taking damage. That's why you can't take a beam laser to an Archon and do any damage; you first have to get over the 4000 shield generation/second threshold. 
A shield delay of 1 (frame) would mean that if damage happens on that frame, generation doesn't begin until the next frame, but if on the next frame damage happens again (such as with a laser), then the generation gets delayed again.

SolraBizna:
Aha. My mental simulation had an off-by-one.

PR #4524, opened 2019-09-10
Added attributes that delay shield generation and hull repair

Amazinite:
Ref: #4408 

These attributes delay when shield generation and hull repair occur after certain scenarios, either after taking damage to the shield or hull and after the shield are completely disabled or after the ship is disabled. 

These attributes apply to ALL shield generation or hull repair and are additive with additional outfits that have the attributes.


Amazinite:
I haven't made the change to have these attributes convert from frames to seconds in the outfitter, since I'd have to change it when #4521 is merged.

source/Ship.cpp
```diff
@@ -2640,7 +2653,11 @@ int Ship::TakeDamage(const Projectile &projectile, bool isBlast)
 	else if(shieldDamage > shields)
 		shieldFraction = min(shieldFraction, shields / shieldDamage);
 	shields -= shieldDamage * shieldFraction;
+	if(shieldDamage)
+		shieldDelay = (shields <= 0.) ? attributes.Get("disabled shield delay") : attributes.Get("shield delay");
```
tehhowch: `"disabled shield delay"` applies even if the ship is not actually disabled?
I would wait to set the delays until you know if the ship is disabled or not, and then you can make sure that the disabled hull delay is set for all hull damage received after being disabled and not just the initial hit that disabled the ship.

source/Ship.cpp
```diff
@@ -1621,6 +1627,11 @@ void Ship::DoGeneration()
 			heat += (hullAvailable - hullRemaining) * hullHeat;
 		if(shieldsAvailable)
 			heat += (shieldsAvailable - shieldsRemaining) * shieldsHeat;
+		
+		// Decrease the shield and hull delays by 1 now that shield generation
+		// and hull repair have been skipped over.
+		shieldDelay = max(0., shieldDelay - 1.);
```
tehhowch: Note that once a ship gets disabled, `DoGeneration` doesn't get executed. Is it intentional that these delays would then only kick in once the ship gets repaired?

Amazinite: Yes, that is intentional. This is actually necessary for the disabled hull delay to work given the current point that I’m setting that delay, as otherwise the disabled hull delay would begin decreasing while the ship is disabled instead of right when the ship becomes enabled again.

source/Ship.cpp
Amazinite: I've reworded the tooltips to clarify the conditions that each delay is suppose to trigger.

Amazinite:
If disabled shield/hull delay is not specified (i.e. it is 0), but the normal delay is, should the normal delay be used?

Zitchas:
Would it be possible to have whether or not the regen is active on disabled ships be controlled by an attribute?

It would be neat to be able to have ships (the automatons would be a good candidate) that - after they are disabled - will slowly regenerate their hull until they are active again. And by slow, I mean "30 secs to a minute of waiting". 

Likewise, having some kind of "repair droid" or something players could outfit on their ship that would gradually repair their ship up to the point of restoring functionality would be a great quality of life thing. I've lost count of the number of times I've been left floating adrift by pirates. It'd be nice if I could at least limp back to port eventually.

Amazinite:
I don't see it as being difficult to have that be the case, but that seems like a separate issue.

data/tooltips.txt
```diff
@@ -230,6 +236,12 @@ tip "shields charge / max:"
 tip "shields:"
 	`Total shield strength. Most weapons cannot damage your hull until your shields are depleted. Some weapon effects, such as heat damage, become more severe once shields are down.`
 
+tip "shield delay:"
+	`How many seconds it takes for shield generation to begin after taking shield damage.`
+
+tip "disabled shield delay:"
```
Amazinite: Rename to `"depleted shield delay"`?

petervdmeer:
The name `"disabled repair delay"` seems to match better with the feature suggested by Zitchas than with the feature that is inteded here for the keyword.

Amazinite:
What would you suggest otherwise?

petervdmeer:
If I search for the opposite of disabled, then I get a lot of suggestions; abled, enabled, unimpaired, mended, uncrippled, fixed, patched up, restored.

I would go for `mended repair delay` or for `enabled repair delay`.

Amazinite:
Those names don't really make sense to me. How does "disabled repair delay" suggest anything other than "how much delay before your repairs start after having been disabled"? Just as how "depleted shield delay" is "how much delay before your shield regeneration starts after having been depleted"?

petervdmeer:
The difference is in where to start the counter:
- "disabled repair delay" suggests to me "the delay after you have become disabled", the time-period since the start of the situation "ship is disabled".
- "mended/enabled repair delay" suggests to me "the delay after you have become no longer disabled", the time-period since the end of the situation "ship was disabled". You could also call it "post-disabled repair delay" if you like.

If "repair when disabled" was a thing, then I would also expect a "disabled repair delay" to apply for repairs while a ship is disabled. But I would expect a "mended/enabled repair delay" to be not relevant for repairs while a ship is disabled (since the counter starts outside/after the situation where the "repair when disabled" feature works).

"depleted shield delay" is also calculated since the start of the situation "ship has zero shields", since the shields will remain zero while the delay counter is counting down.

petervdmeer:
The tooltips now match the implementation; I'm still not happy with the fact that the post-hit delays don't elapse while disabled, as content-creator I expect those delays to work better if they only get affected by the moment of hit and nothing else.

Maybe we should ask Zitchas or tehhowch to give their opinion on the concepts behind those delays. If they agree with the implementation as given, then I'm okay with merging this.

Zitchas:
> How many seconds it takes for hull repairs to begin after taking hull damage while the ship is disabled. This delay does not elapse if the ship is disabled.

To me, this is a contradiction in terms. "Here's how many seconds it takes for repairs to start after taking damage while the ship is disabled. This amount of time will never elapse while the ship is disabled."

Amazinite:
The time is set when the ship is disabled, but does not elapse until the ship is no longer disabled. I don't see the contradiction in that.

Zitchas:
In which case, this really needs to be changed to something more along the lines of 
`How many seconds it takes for hull repairs to begin after recovering from being disabled.`

After all, ships are only disabled by taking damage, either physically in-game, or in the lore of how it came to be disabled. So the whole business about "This is how long it takes after being hit while being disabled to start repairing... but the timer doesn't start until you recovered from being disabled" is overly convoluted.

Amazinite:
@Zitchas @petervdmeer How's it look now? The behavior is the exact same, I've simply changed the attribute name and tooltip.

Amazinite:
> I'm also not fully happy with the text `This delay does not elapse if the ship is disabled.` This looks like too much info for regular players. I suggest to just remove this part of text in all the tooltips.

I explained on stream how I think that when these attributes become used for the first time, the description of whatever outfit that they're on should give a much more thorough explanation of the mechanic, which would include the idea that being disabled causes the delay to pause. (E.g. if the explanation for the delay is that the shield generator needs to warm up after taking damage, you can't exactly have it warm up while disabled.)

Amazinite:
Bumping so I can link.

Issue #4407, opened 2019-07-03
Shield and Hull Recharge multiplier proposal and RFC v0.9.11+


ChamEV:
## Shield and Hull Recharge multiplier ##
The primary purpose of **Shield recharge multiplier** and **Hull repair multiplier** is not to easily create overpowered regeneration, but to allow outfits to reduce the regeneration so that stacking multiple outfits will not work.

Consider a small shield regeneration outfit, 5 tons, 100 shield regeneration/sec.  Normally that'd be very overpowered.  But combine it with **Shield recharge multiplier** of -.3, that is recharging 70% of normal rate with 1, and 40% of normal rate with 2.  Now you have an outfit that can be put on a fighter that is not attractive in any number for larger ships.

Quantity | Shield multiplier | Shield Regen / second
---------| ---------| ------------
1 | 0.7 | 70
2 | 0.4 | 80
3 | 0.1 | 30

Pointedstick:
Makes sense, but this might be rather difficult to communicate to the player in the UI.

ChamEV:
I have a few ideas regarding that.

* First any time this is applied, the end result will instantly be apparent in on the `shields charge/max: x / y` line with x changing.
* Usage of "increased" and "decreased" as descriptors as these are stacked additively before being multiplied.  Borrowing from [path of exile's terminology](https://pathofexile.gamepedia.com/Stat#more_vs_increased)
* There is a general need for a complete stats page now that ships are coming equiped with built in ramscoops and cloaking among other things.  This would fit in there.

Amazinite:
So would that mean that 4 generators with a -.3 multiplier would yield 0 shield regen? And how would this interact when mixed with shield generators that don't have a multiplier? Does this multiplier effect ALL shield generation on a ship?

jafdy:
I should probably be a multiplicative penalty, so -.3 would become (-.3)^n

Amazinite:
I think a better vanilla use of such an attribute, instead of making shield generators supposedly only attractive to smaller ships, would be to give smaller ships an intrinsic positive multiplier.

jafdy:
So, I know you were worried about a can of worms, but if you want to make it work for every attribute, rather then give it a name you could use operator symbols. So for adding you use `<attribute> +<num>` or just `<attribute> <num>`, for multiply you use `<attribute> *<num>` (the same could then be used for `/` and `-` is there already in effect.
It just leaves the question of what order do you apply the operations.

[...]

PR #4525, opened 2019-09-11
Multipliers for shield generation and hull repair related stats

Amazinite:
Ref: #4407

Adds multipliers for shield generation, hull repair, and their respective energy/fuel/heat stats. 
"Multipliers" are actually additive, with a value of 1 being a 100% increase in the stat, and if you have two outfits with a multiplier of 10%, then the stat will end up being multiplied by 1.2. Negative values are possible, allowing for the reduction in a stat.


Amazinite:
This needs the ShipInfoDisplay updated to properly reflect the effects of a ship's multipliers on its stats.

Amazinite:
`reverse thrusting energy` is currently the longest attribute name, but 
`shield generation multiplier` would be longer than that, so maybe a different word than multiplier is necessary.
`shield generation booster`? Would still be the longest, but not by as much.

Amazinite:
It might be fine in game as is. Not quite as much space for larger numbers, but I wouldn't expect these attributes to have large numbers in practice. 
image

petervdmeer:
Most other stacked stats in the game use multiplication instead of addition. If one outfit adds 10% shield-boost and another outfit adds 10% shield-boost, then the total boost would be 21% (1.1 * 1.1).

Removing 20% of a stat would be multiplication by 0.8.

Is it critical for this PR that the combination of stats is additive instead of multiplicative? If it is not, then I suggest to switch to multiplicative.

Amazinite:
> Most other stacked stats in the game use multiplication instead of addition.

Really? What stats would those be?
Attributes from various outfits are all added together to get the final value of that attribute for the ship.

petervdmeer:
> > Most other stacked stats in the game use multiplication instead of addition.
> 
> Really? What stats would those be?
> If you're thinking of an attribute like ramscoop, ramscoop is still additive, it's just that the square root of the added value is used to get the diminishing returns effect.

Sorry, I was indeed thinking of stats like `ramscoop`, `outfit scan power` and `cooling inefficiency`, but you are right, they don't get multiplied initially.

> > Is it critical for this PR that the combination of stats is additive instead of multiplicative? If it is not, then I suggest to switch to multiplicative.
> 
> To make various outfits with these stats multiplicatively combine would require adding some way for the game to keep track of which stats are coming from which outfits when installed on a ship instead of just adding everything together.

Yes, it would be a variable that is initialized at 1 and instead of continuously adding the value for every outfit the value would be multiplied for every outfit. Would be a little bit more coding since until now all stats indeed just get added.

> Multiplicative outfits would also be far more difficult for the player to keep track of. (E.g. having 3 10% outfits and 4 5% outfits is easier to add to 50% than it is to figure out that that's actually a 155% boost in some stat, this example also showing the runaway effect of multiplicatively combining outfit attributes.)

Yes, adding is the correct operation if you don't want the effect of one multiplier outfit to also work on top of another multiplier outfit. I just wasn't sure if you indeed intended the multiplier to work that way.

Amazinite:
Ref: #4682
I've hard coded a whitelist for the new attributes added by this PR to allow them to be negative, but #4682 asks that this be something that is listed in the game data. For now the hard coded whitelist should do, though.

data/tooltips.txt
```diff
@@ -134,15 +134,27 @@ tip "hull:"
 tip "hull energy:"
 	`Energy consumed per second when repairing the hull.`
 
+tip "hull energy multiplier:"
+	`A multiplier that affects hull energy. A multiplier of 10% will increase the stat by 10%. Multipliers from multiple outfits are additive.`
```
petervdmeer, line 138: The tooltip texts here are a bit long and feel a bit "out of lore"; Maybe change to something like:
`Increases the energy consumed when reparing the hull by the given factor.`
This is a little bit less precise, but it seems to match the style of the `hull energy` text better and I think it should be fine for the average player.

Amazinite: Well these could potentially be negative (after #4942), so do we want to use the term "increases"?

petervdmeer: You could add the text `(Decreases if the factor is negative)` to be complete.

Amazinite: How do they look now?

petervdmeer: Looks good to me.

Issue #4488, opened 2019-08-25
Illegal Cargo fined and confiscated twice


Person-93:
**Describe the bug**
I was carrying illegal drugs, I got the popup about being fined and the drugs confiscated. I jumped to another system. I was scanned again. I got another popup about being fined and having the drugs confiscated.

**To Reproduce**
1. Accept a mission to carry illegal drugs
2. Get scanned.
3. Let a day pass (not sure if this is necessary, but I am including it because that is what happened.
4. Get scanned again.

**Expected behavior**
I expect the drugs to only be confiscated once. How can they be confiscated twice? That doesn't seem to make much sense.

**System (please complete the following information):**
 - OS: Linux
 - Game Source: self-compiled
 - Version: 0.9.9

jafdy:
Because you managed to keep some of them from being taken?
So what will be happening is that the mission cargo stays on your ship until you land, even if you have failed the mission for whatever reason. This allows you to be scanned and have the illegal item text come up more than once.

Amazinite:
Are you sure the illegal goods were confiscated? Not all illegal jobs result in the confiscation of your goods if you’re fined, only a fine.

Person-93:
@Amazinite @AdamKauffman the popup said that the goods were confiscated.

Amazinite:
Being scanned with a stealth mission is causing the mission to fail (the destination marker and mission name grey out to show that they can't be completed now), but since https://github.com/endless-sky/endless-sky/issues/1938 is the case and missions don't get removed until landing, the cargo can be scanned multiple times.

Amazinite:
Fixed in https://github.com/endless-sky/endless-sky/commit/c9b66ab7fcbdba7fe1cba00f99bfbd83f46a6e14.

Issue #2689, opened 2017-07-01
With multiple illegal jobs, you can only get fined for one at a time


Amazinite:
Say you have 10 pirate jobs, all with fines attached to them if you get scanned. Right now, if you do manage to get scanned, you will only get fined for one of the jobs and the other 9 jobs will not result in the player being fined.

It would make more sense that you get fined for all 10 at the same time, but seeing as how that could cause dialog box clutter (having to hit cancel or ok 10 times to get rid of all the boxes), the dialog box should probably combine all the fines and display what you've been fined in a single dialog box. Then there's a question of what should be done with illegal messages (e.g. for stealth missions with special messages when you get fined). Should those appear in multiple boxes, or should only one of them appear as what shows up in the dialog box?

tehhowch:
Given the time @Hardon1776 spent with the fine system, maybe he will make a PR to address this.

I certainly agree that you should be fined for each different illegal mission cargo /substance you are carrying. I also agree this dialog should only appear once.

The dialog box can have two sections, with the top mentioning the fines, and the bottom mentioning the number of missions that failed due to the scan.
Perhaps we should create a mission fail dialog snippet specifically designed for failed illegal mission scan dialog concatenation, and this snippet can be displayed to better convey which mission failed. (An `on scan` condition?)

tehhowch:
I did a little reviewing, and it seems this is already done?
[Politics.cpp:226](https://github.com/endless-sky/endless-sky/blob/c451d28a227163448ad30539284d1554660492e9/source/Politics.cpp#L226)
```
if(!scan || (scan & ShipEvent::SCAN_CARGO))
{
	int64_t fine = ship->Cargo().IllegalCargoFine();
	if((fine > maxFine && maxFine >= 0) || fine < 0)
	{
		maxFine = fine;
		reason = " for carrying illegal cargo.";

		for(const Mission &mission : player.Missions())
		{
			// Append the illegalCargoMessage from each applicable mission, if available
			string illegalCargoMessage = mission.IllegalCargoMessage();
			if(!illegalCargoMessage.empty())
			{
				reason = ".\n\t";
				reason.append(illegalCargoMessage);
			}
			// Fail any missions with illegal cargo and "Stealth" set
			if(mission.IllegalCargoFine() > 0 && mission.FailIfDiscovered())
				player.FailMission(mission);
		}
	}
}
```

Hadron1776:
Notice this:
[...]

PR #4526, opened 2019-09-12
Fixes and changes to fines for illegal jobs

Amazinite:
Ref: #4488, #2689 

Fines from multiple illegal missions will now stack. The behavior of fines for illegal outfits are not changed.
If the player failed multiple missions at once, only one of the reasons is shown. Now, if the player has failed multiple missions, the player is informed about how many other missions they failed.
The player will no longer be fined for missions that have failed.

Fined the first time while in flight
image

Without landing, fined again. Note that the fine has been reduced and the extra reason is gone, since failed missions no longer fine again 
image


Pointedstick:
The bugfix is good, but seeing "You failed n other missions" in the UI begs the question, "which ones?" If there's no way to see them, that's just going to be a bunch of frustration.

Amazinite:
You can tell which ones by going to your active missions and seeing which ones are now greyed out, since failed missions aren't removed from the mission's list until you land.

Pointedstick:
I know. But that's quite disconnected from the dialog. Maybe you could add that helpful pointer to it?

Amazinite:
That feels a little out of place to me. Outside of the intro missions and the help dialogs for the beginning of the game, nothing that I can recall says something like "Go to this menu to see X," so it would feel odd to me that whenever I fail a number of missions, it tells me "Go check your active missions to see what you failed."

Pointedstick:
It wouldn't have to be phrased like a tutorial message. Maybe something like "n other failed missions can be viewed on the map screen."

Amazinite:
That's still an instance of referring to a different menu as if it were a tutorial of some sort. If someone is taking a load of illegal jobs, some of which that fail on scan, I'd expect them to know where the mission list is and to check it to see what they still have left to finish.

Pointedstick:
Hmm, then how about rephrasing the dialog more like this:

```
You failed n missions after your illegal cargo was discovered.

<fine text>

<mission-specific flavor text for one of the failed missions>
```

Amazinite:
Ordering it that way gets a little more involved, but I'll take a shot at it.

Amazinite:
Might need to try to do that anyways, because I believe that this currently doesn't account for missions that fail on scan but that don't have special illegal messages, as illegal messages aren't required for stealth missions.

Amazinite:
@Pointedstick I've kept the order but changed the wording to match your suggestion.

Pointedstick:
No time for Endless Sky right now, and I trust your judgment on this.

source/Mission.cpp
```diff
@@ -679,6 +679,13 @@ bool Mission::HasFailed(const PlayerInfo &player) const
 
 
 
+bool Mission::HasFailed() const
```
tehhowch, line 682: We should name this and the above method differently, since this one is just a getter, while the above actually does things. `IsFailed`?

source/Politics.cpp
```diff
@@ -241,7 +245,15 @@ string Politics::Fine(PlayerInfo &player, const Government *gov, int scan, const
 					}
 					// Fail any missions with illegal cargo and "Stealth" set
 					if(mission.IllegalCargoFine() > 0 && mission.FailIfDiscovered())
+					{
 						player.FailMission(mission);
+						failedMissions++;
```
tehhowch, line 250: prefix operator

Amazinite:
> Seems like there is an edge case where, if the player is fined for having outfits installed, they won't be notified of any failed missions, provided the outfit fine was greater than the total mission fine

I believe the latest commit should change that. The number of missions that you've failed is now added after both cargo and outfits have been checked, meaning that even if your outfits were worse than your mission cargo, it still mentioned the number of missions failed. I think that if you've committed an atrocity it will overwrite the number of missions failed part, but at that point having failed a mission is probably the least of your worries.

PR #4528, opened 2019-09-13
Missions can now require attributes like they can require outfits

### LINKED ISSUE #2911: Allow missions to require an attribute to trigger

Amazinite:
Resolves #2911 

Adds a mission keyword `attribute` that works the exact same way as `require` except it checks for an attribute instead of an outfit.


Zitchas:
I can't speak to the code, but as a content creator I really, really like this and I think it should be very useful going forward. 

Also, if ships are given attributes identifying their race and/or shipyard, then this opens up all sorts of conditional missions opportunities, such as, say, what happens when you show up on a human world in a ship that is clearly completely alien?

tehhowch:
Unlike outfits, it is perfectly valid for an attribute to be negative and still have the player have it. 

>Also, if ships are given attributes identifying their race and/or shipyard, then this opens up all sorts of conditional missions opportunities, such as, say, what happens when you show up on a human world in a ship that is clearly completely alien?

The attributes referenced here are solely numeric in value, and not text like those associated with systems or planets. I believe @Elyssaen did some work on a tags system better suited to what you are proposing.

source/MissionAction.cpp
```diff
@@ -206,6 +206,11 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			else
 				child.PrintTrace("Skipping invalid \"require\" amount:");
 		}
+		else if(key == "attribute" && hasValue)
+		{
+			double count = (child.Size() < 3 ? 1. : static_cast<double>(child.Value(2)));
```
tehhowch: The static cast isn't needed for doubles

Amazinite:
> The attributes referenced here are solely numeric in value

No reason why one couldn't put an attribute `"quarg" 1` on an outfit/ship and use this mechanic to check for its existence. Would be out of place in the outfitter though, so the tags system is a more elegant solution for such a case.

source/MissionAction.cpp
```diff
@@ -385,6 +392,25 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
...
+		if(!it.second)
+		{
+			for(const auto &ship : player.Ships())
+				if(!ship->IsDestroyed() && ship->Attributes().Get(it.first))
+					return false;
+		}
+		else
+			attribute += flagship ? flagship->Attributes().Get(it.first) : 0;
+		
+		if(attribute < it.second)
```
tehhowch: This check won't work correctly for negative `requiredAttributes`

Amazinite: How so? Are you expecting a negative required attribute value to return true if the player has an even more negative attribute value?

Amazinite:
>What's the expected behavior when the mission has e g. `attribute "afterburner thrust" 1.5` and only the flagships escorts have afterburners?

Whatever the same expected behavior is for `require` under similar conditions.

source/MissionAction.cpp
tehhowch: This makes me wonder if instead of this handling like outfits, we instead treat the attributes as we do conditions like reputation, e.g.

```bash
attributes [<location specifier>]
    <attribute name | attribute value expression> <comparison operator> <value | value expression>
    ...
```
These could then be combined logically as is done with conditions.

Amazinite: That implementation is on you.

Amazinite: This PR does what I had in mind for the issue, so I think this is fine to merge as is, and the mechanic can be expanded upon later. If you think it should be the case though, I can change it so that negative attribute will compare for greater to return false than while positive values still compare for less than.

tehhowch:
>This PR does what I had in mind for the issue, so I think this is fine to merge as is, and the mechanic can be expanded upon later. 

I don't think introducing a keyword and then later changing the syntax of its use is a good idea. We must either invalidate the earlier syntax, or support both forms, which complicates the parsing of the tokens.

A syntax I envision for attribute checks handled by ConditionSets would be
```
attributes [<location specifier>]
   <value expression 1> <comparison operator> <value expression 2>
    ...
```
where the value expressions can be any nonzero number of tokens that will be parsed as attribute names from the ships identified by the optional location specification. Any number of child nodes could be specified for the `attributes` node, and multiple specifications would be allowed. Behavior for the same location specifier would be equivalent to that when defining multiple conditions (each parent `attributes` node would need to be satisfied).

The barrier to me working on this proposed implementation is your/Pointed's acceptance of #3128, which heavily refactors ConditionSet.

Amazinite:
>The barrier to me working on this proposed implementation is your/Pointed's acceptance of #3128, which heavily refactors ConditionSet.

If you think that #3128 is good to go then I'd say just merge it. I don't exactly have the expertise to say one way or another if #3128 does things as best as it could, but if it does what it claims to do then it seems like a great addition to me.

Amazinite:
In this PR we've mentioned the idea of using comparison operators for `attribute`, and in #5302 we talked about changing the syntax of `require` and `outfit` to fit a syntax along the lines of `<action> <object> <object name> <amount>` where `<action>` can be `own` (same behavior as `require`). If we want comparison operators for `attribute`, would we also want comparison operators for `own outfit <name>`? e.g. `own outfit "Jump Drive" <= 3`. And in that way, should this `own <object> <object name>` syntax be extended to here? That is, the syntax here would go from `attribute <name> <amount>` to `own attribute <name> <amount>`, later to be extended to allow a comparison operator between the name and the amount. And in that regard as well, should we simply add that syntax here then leave the mechanic without comparison operators similar to the proposition in #5302 to take just `give ship <model> <name>` for now?

Amazinite:
Converting back to a draft until we finalize syntax.

Amazinite:
Closing as I don't plan to get to this any time soon in terms of expanding the syntax, and as far as the mechanic itself goes I feel like the situations where it's absolutely necessary and a workaround can't be found are infrequent enough that it's not too much worth it to merge a syntax that we'll change later down the line.

PR #4531, opened 2019-09-13
Added a mission that helps the player if they become stuck in the Ember Waste

Amazinite:
Ref: #4491, #4493

This PR adds a mission that accounts for the scenario where the player enters the Ember Waste but loses their jump drive and/or quantum keystone for whatever reason and becomes stuck. The mission will only offer if the player doesn't have a quantum keystone/key stone and doesn't have a jump drive. Once the player is in this situation, they will trigger and event on a 14 day delay that triggers the mission. This delay is so that the player gets a sense of being stuck before help shows up. I just thought it'd be more interesting that way. 

This change reverts #4491 since that change will no longer be necessary to get the player out of this scenario.

Also, felt like pinning down the Caelian description with some story. Let it be known that the Remnant found out that the Alpha Wars had ended in the 2880s.

### DIFF
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -142,6 +142,8 @@ mission "Remnant Blood Test Patch"
 mission "lost in ember waste"
 	landing
 	invisible
+	source
+		governmant "Remnant"
 	to offer
 		has "remnant blood test pure"
 		not "license: Remnant"
@@ -158,6 +160,8 @@ event "lost in ember waste"
 mission "Remnant: Lost in Ember Waste"
 	landing
 	invisible
+	source
+		governmant "Remnant"
 	to offer
 		has "event: lost in ember waste"
 		outfit "Quantum Key Stone" 1
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -224,7 +224,7 @@ mission "Remnant: Lost in Ember Waste"
 			`	"Well I got lost when I was 17 and returned when I was 64, so perhaps some 130 years ago now that I returned." She chuckles at the surprised look on your face as you realize that she is nearly 200 years old. "Uncovering ancient alien technology is quite the blessing.`
 				goto end
 			
-			label daughtesr
+			label daughters
 			`	"Oh, I made sure to tell them I was returning home. They know where to find me if they ever need me, and I have made trips to see them every decade or so. I'm nearing the end of my days, though, so I may only have one more trip left in me.`
 			
 			label end
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -199,7 +199,7 @@ mission "Remnant: Lost in Ember Waste"
 				decline
 			
 			label republic
-			`	"I was an adventurous spirit in my youth," she says. "One day I found my way out of the Ember Waste and back to what I learned was the rest of humanity. For years my people had hidden here believing the Alphas had killed everyone else, but I now knew that we had little to fear.
+			`	"I was an adventurous spirit in my youth," she says. "One day I found my way out of the Ember Waste and back to what I learned was the rest of humanity. For years my people had hidden here believing the Alphas had killed everyone else, but I discovered that we had little to fear.
 			`	"I explored the Republic for many months, learning what had happened since the Alpha Wars, but when I attempted to return to the Ember Waste to tell my people, I found that my key stone had been stolen at some point during my journey. I had no way to return."`
 			`	She lets out a long sigh. "I don't want to bore you with all the details, but I resigned myself to my new life. I met a man, had two daughters, but never lost hope of returning to my true home. One day, after my daughters had grown and my husband had passed in an accident, I decided to go on another adventure. I dusted off my old ship, and soon discovered Hai space."`
 			branch hai
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -201,7 +201,7 @@ mission "Remnant: Lost in Ember Waste"
 			label republic
 			`	"I was an adventurous spirit in my youth," she says. "One day I found my way out of the Ember Waste and back to what I learned was the rest of humanity. For years my people had hidden here believing the Alphas had killed everyone else, but I discovered that we had little to fear.`
 			`	"I explored the Republic for many months, learning what had happened since the Alpha Wars, but when I attempted to return to the Ember Waste to tell my people, I found that my key stone had been stolen at some point during my journey. I had no way to return."`
-			`	She lets out a long sigh. "I don't want to bore you with all the details, but I resigned myself to my new life. I met a man, had two daughters, but never lost hope of returning to my true home. One day, after my daughters had grown and my husband had passed in an accident, I decided to go on another adventure. I dusted off my old ship, and soon discovered Hai space."`
+			`	She lets out a long sigh. "I don't want to bore you with all the details, but I resigned myself to my new life. I met a man, had two daughters, but never lost hope of returning to my true home. One day, after my daughters had grown and my husband had passed in an accident, I decided to go on another adventure. I dusted off my old ship, and soon discovered the Hai."`
 			branch hai
 				has "First Contact: Hai"
 			choice
@@ -225,7 +225,7 @@ mission "Remnant: Lost in Ember Waste"
 				goto end
 			
 			label daughters
-			`	"Oh, I made sure to tell them I was returning home. They know where to find me if they ever need me, and I have made trips to see them every decade or so. I'm nearing the end of my days, though, so I may only have one more trip left in me.`
+			`	"Oh, I only told them about deciding to travel the galaxy. Never told them about returning home. I made trips to see them every so often, but they both passed many decades ago now. It seems like the Republic can't quite get anyone to live past 130 years.`
 			
 			label end
 			`	"I really must get going, now," she says. The woman gets up to leave, but before walking away she turns to you and sings, "Safe travels, Captain," with a big grin on her face. You say goodbye to her in return and part ways.`
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -164,6 +164,7 @@ mission "Remnant: Lost in Ember Waste"
 		governmant "Remnant"
 	to offer
 		has "event: lost in ember waste"
+	on offer
 		outfit "Quantum Key Stone" 1
 		log `Became stuck in the Ember Waste, but was given a quantum key stone by an old woman who claims to have once been stuck in Republic space under similar circumstances.`
 		conversation
--- a/data/remnant missions.txt
+++ b/data/remnant missions.txt
@@ -161,6 +161,7 @@ mission "Remnant: Lost in Ember Waste"
 	to offer
 		has "event: lost in ember waste"
 		outfit "Quantum Key Stone" 1
+		log `Became stuck in the Ember Waste, but was given a quantum key stone by an old woman who claims to have once been stuck in Republic space under similar circumstances.`
 		conversation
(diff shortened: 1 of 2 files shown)


data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+			choice
+				`	"Thanks for the help. I'll go fit this on my ship."`
+				`	"How'd you get to Republic space."`
+					goto republic
+			
+			`	You say goodbye to the woman and begin walking to your ship. "Safe travels, Captain," she sings with a big grin on her face as you walk away.`
+				decline
+			
+			label republic
+			`	"I was an adventurous spirit in my youth," she says. "One day I found my way out of the Ember Waste and back to what I learned was the rest of humanity. For years my people had hidden here believing the Alphas had killed everyone else, but I now knew that we had little to fear.
```
LocalGod79: "...now knew..." is super awkward (mixing tenses) - ...found out...then knew...discovered...now know... are much better alternates

LocalGod79:
love the little story, btw - makes me want to lose an outfit, just to have the experience...

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+			`	"To my surprise, the Hai had a large number of quantum key stones." She pauses for a moment. "Might be a good idea for you to bring some of those Hai key stones here. Anyways, I was so thrilled that I didn't even ask any questions about how they had them. I bought one and flew as fast as I could back to the Ember Waste, and I've been here ever since."`
+			choice
+				`	"How long ago was that?"`
+				`	"What about your daughters?"`
+					goto daughters
+			
+			`	"Well I got lost when I was 17 and returned when I was 64, so perhaps some 130 years ago now that I returned." She chuckles at the surprised look on your face as you realize that she is nearly 200 years old. "Uncovering ancient alien technology is quite the blessing.`
+				goto end
+			
+			label daughtesr
```
petervdmeer: typo: daughters

petervdmeer:
Nice story indeed.

Amazinite:
Should this mission have an autosave? It isn't an important mission, but if the player *somehow* gets stuck again, then having an autosave here means they're guaranteed to have a place to jump back to where they can leave the Ember Waste.

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+	to offer
+		has "event: lost in ember waste"
+		outfit "Quantum Key Stone" 1
+		log `Became stuck in the Ember Waste, but was given a quantum key stone by an old woman who claims to have once been stuck in Republic space under similar circumstances.`
+		conversation
+			`You wander the <origin> spaceport feeling very lost in this strange region of space. Having flown around for a few weeks with no idea of how to get out, you wonder if you'll be stuck here forever.`
+			`	You sit down on a bench and think about how you're going to get out of this situation. While you think, an old woman approaches you and signs something.`
+			choice
+				`	"Sorry, I don't understand."`
+			`	The old woman chuckles to herself. "You seem lost." She says. You nod. To your surprise, she doesn't sing everything she says like the other Remnant.
```
Bladewood: Should be a comma after "lost," and lowercase "she."

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+				`	"You're just giving this to me?"`
+					goto thanks
+				`	"What's the catch?"`
+			`	She chuckles again. "No catch."`
+			
+			label thanks
+			`	The woman sits down on the bench next to you. "I was once lost in Republic space for many years after losing my key stone. I know what it can be like to be away from your people for that long, so I'd rather not see you share the same fate."`
+			choice
+				`	"Thanks for the help. I'll go fit this on my ship."`
+				`	"How'd you get to Republic space."`
```
Bladewood: Question mark.

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+					goto republic
+			
+			`	You say goodbye to the woman and begin walking to your ship. "Safe travels, Captain," she sings with a big grin on her face as you walk away.`
+				decline
+			
+			label republic
+			`	"I was an adventurous spirit in my youth," she says. "One day I found my way out of the Ember Waste and back to what I learned was the rest of humanity. For years my people had hidden here believing the Alphas had killed everyone else, but I discovered that we had little to fear.
```
Bladewood: Closing backtick.

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+			`	"I explored the Republic for many months, learning what had happened since the Alpha Wars, but when I attempted to return to the Ember Waste to tell my people, I found that my key stone had been stolen at some point during my journey. I had no way to return."`
+			`	She lets out a long sigh. "I don't want to bore you with all the details, but I resigned myself to my new life. I met a man, had two daughters, but never lost hope of returning to my true home. One day, after my daughters had grown and my husband had passed in an accident, I decided to go on another adventure. I dusted off my old ship, and soon discovered Hai space."`
+			branch hai
+				has "First Contact: Hai"
+			choice
+				`	"The who?"`
+			`	"How you've discovered and become lost here without having discovered the Hai first is beyond me." She chuckles. "The Hai are a peaceful alien people north of humanity. Go looking for a purple wormhole in the uninhabited systems of the Far North. You may find your way there.`
```
Bladewood: "How you've discovered and become lost here" sounds a little awkward.

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+				goto next
+			
+			label hai
+			choice
+				`	"You've met the Hai?"`
+			`	"Well of course. How else would I have been able to return home?`
+			
+			label next
+			`	"To my surprise, the Hai had a large number of quantum key stones." She pauses for a moment. "Might be a good idea for you to bring some of those Hai key stones here. Anyways, I was so thrilled that I didn't even ask any questions about how they had them. I bought one and flew as fast as I could back to the Ember Waste, and I've been here ever since."`
```
Bladewood: I don't know what to think about the use of "anyways" here.

Amazinite: She went a little off topic (told you to bring Hai Keystones to Remnant space) and the "anyways" is her getting back on topic.

petervdmeer:
Maybe we should add a new mission where the players jump-drives and key-stones get stolen in remnant space. This would make this mission available for more players.

data/remnant missions.txt
Bladewood: No, I meant as opposed to "anyway."

LocalGod79:
> Maybe we should add a new mission where the players jump-drives and key-stones get stolen in remnant space. This would make this mission available for more players.

a super low-chance random mission - on landing "When you return to your ship, you find several strangers milling about your landing pad.  When you approach them, they quickly sign apologetically, and swiftly walk away.  Their behaviour seemed odd, even for Remnants, but you are not very familiar with all aspects of their culture." (keystone removed)

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
...
+				`	"How long ago was that?"`
+				`	"What about your daughters?"`
+					goto daughters
+			
+			`	"Well I got lost when I was 17 and returned when I was 64, so perhaps some 130 years ago now that I returned." She chuckles at the surprised look on your face as you realize that she is nearly 200 years old. "Uncovering ancient alien technology is quite the blessing.`
+				goto end
+			
+			label daughters
+			`	"Oh, I made sure to tell them I was returning home. They know where to find me if they ever need me, and I have made trips to see them every decade or so. I'm nearing the end of my days, though, so I may only have one more trip left in me.`
+
```
Zitchas: Given how secretive the Remnant are, I don't think it is in character for them to casually let strangers (which her daughters would be) know where to find her. Might be better to change the sentence about "They know where to find me..." to something along the lines of "They think I am wandering around exploring the galaxy, and I left a QEC with them so they can send me messages when they need to."

For one, this change fits better with the Remnant culture of secrecy/security. For another, this opens up a potential venue for future missions, as the lady could come find the player to get help dealing with some problem that her daughters have let her know about.

Zitchas:
@LocalGod79 I'm not sure about having the Remnant steal stuff from the player.

That being said, having a parallel mission to this where the lady approaches you and has a similar conversation would be good. Possibly starting off with a good-natured warning to guard your Quantum Key Stone carefully, leading into her story about having hers stolen, then leading into her reminiscences that provide the player with the info.

data/remnant missions.txt
Amazinite: Ah, alright. Fixed another instance of "anyways" in this file, then.

Amazinite: Ah, alright. Fixed another instance of "anyways" in this file, then.

data/remnant missions.txt
Amazinite: >Given how secretive the Remnant are, I don't think it is in character for them to casually let strangers (which her daughters would be) know where to find her.

My intention with that was to show how she's rather different from the rest of the Remnant (not as secretive, didn't fear leaving the Ember Waste, doesn't sing (although that's just her not singing around you since you're from the Republic)), but I'm alright with changing this one part.

Amazinite:
After writing this mission, I thought of having such a mission where someone steals *just* your jump drive (the world really isn't as harsh to the player as it could be), but any such mission I'd think should be a voluntary thing. It shouldn't be "you show up to your ship and everything's stolen," it should be "you're approached by someone who asks if they can borrow your jump drive. Yes or no?"

data/remnant missions.txt
Zitchas: Maybe just change the wording a bit to imply it hasn't been very often. I think it might negatively impact the implications of the Remnant's isolation and lack of contact with humanity if they already have someone going back and forth regularly. But I don't see a problem with mentioning that she's gone back and forth a couple times. Especially if the way she says it implies that it is a long and/or difficult and/or dangerous trip.

Amazinite: Something I just realized too; the assumption here is that her daughters are still alive, which means that her daughters are at least say 150 years old.
As far as I'm aware, there's zero mention in game of the typical age that people live to in Republic space. But, this mission does make it seem that living for 200 years is remarkable and out of the ordinary, so it could be that no one lives near that long in the Republic. 

Something I mentioned on Discord is that while IRL life expectancy has increased with time, the ages of the oldest living people have not seemed to increase in the same way; we get more old people, but not people living older than anyone prior, that being a max of about 120 years. So, I think it'd make sense to say that her daughters passed decades ago, because "It seems like the Republic can't quite get anyone to live past 130 years," or something like that.

Zitchas: That makes sense, and would work for this.

data/remnant missions.txt
```diff
@@ -139,6 +139,99 @@ mission "Remnant Blood Test Patch"
 
 
 
+mission "lost in ember waste"
+	landing
+	invisible
+	to offer
+		has "remnant blood test pure"
+		not "license: Remnant"
```
MageKing17: I don't see a requirement for either this mission or its followup to actually be given *in* Remnant space, which means that visiting just long enough for the blood test, leaving, then selling your keystone could result in a very strange conversation about being "stuck in this strange region of space" when you're in Republic territory.

Amazinite: Always gotta be something missing, I guess.

data/remnant missions.txt
jafdy: You could go for comments about visiting grand or great-grand kids instead.

PR #4537, opened 2019-09-15
Add preference to toggle starfield

Disiuze:
Inspired by @warp-core in the Discord server.
This can be used for saving video bitrate when streaming.

~~I would make it so that disabling the starfield and haze are mutually exclusive, but I don't know how.~~


source/StarField.cpp
```diff
@@ -90,57 +90,61 @@ void StarField::SetHaze(const Sprite *sprite)
 
 void StarField::Draw(const Point &pos, const Point &vel, double zoom) const
 {
-	glUseProgram(shader.Object());
-	glBindVertexArray(vao);
+	// Draw the starfield unless it is disabled in the preferences.
+	if(Preferences::Has("Draw starfield"))
```
Amazinite: I think you should do
```
if(!Preferences::Has("Draw starfield"))
	return;
```
 as to not need to indent everything. See lower down the same thing done for the background haze preference.

SebastienBruckert: I think it would be better to write the opposite, so that you avoid adding an indentation level to the whole function and it reduces the git change. More readable too. It would be a good idea to cut this function in two though

```c++
if(not Preference::Has("Draw starfield"))
    return;
```

Amazinite: I thought the same thing, but the reason it's done this way is so that having the "Draw starfield" preference off doesn't also skip the "Draw background haze" section lower down in this function.

SebastienBruckert: So why not cutting this function in two ? Because the second part is not dependent of the first part, it does not make sens to keep them together. Maybe just add a `drawBackgroundHaze`, and call it at the end of this function

Issue #4533, opened 2019-09-14
String interpolation for phrases in hail messages


Thunderforge:
### Background

The `hails.txt` file is structured in such a way that one can combine words and phrases for randomly generated strings. For instance, we might have this:

```
phrase "planet"
 	word
 	 	"Earth"
 	 	"Luna"
 	 	"Mars"

phrase "friendly civilian"
 	word
 	 	"I'm going to "
 	 	"Looking forward to visiting "
 	 	"Heading over to "
 	phrase
 	 	"planet"
 	word
 	 	" after I finish this job."
 	 	" once I raise enough credits."
 	 	"!"
```

And this will generate hail messages like

> I'm going to Earth after I finish this job.
> Looking forward to visiting Luna once I raise enough credits.
> Heading over to Mars!

### Problem

Consider [this section](https://github.com/endless-sky/endless-sky/blob/eee43a1b54fba48d9777168d484b9a0732beebc1/data/hails.txt#L1910) of `hails.txt`:

 	word
		"buy drugs from"
		"sell drugs to"
		"sell slaves to"

Say I created a new phrase called `contraband` that included both "slaves" and "drugs". Because I need to say either "buy from" or "sell to", I can't use that new phrase in this situation. For instance, if I had

 	word
		"buy "
		"sell "
 	phrase
		"contraband"
	word
		" from"
		" to"

…then I might wind up with nonsensical phrases like

> buy drugs to
> sell slaves from

### Feature Request

I would like to be able to reference a phrase from within a word through [string interpolation](https://en.wikipedia.org/wiki/String_interpolation) (or something with equivalent functionality). The syntax might look like this:

 	word
		"buy $contraband from"
		"sell $contraband to"

…where `$contraband` is a reference to the phrase with the same name. This would ensure that we can provide more variety of hail messages for complex phrases like this, and avoid artificially limiting what can be said. 

This would also allow for phrases to be optional. Consider this section [this one](https://github.com/endless-sky/endless-sky/blob/eee43a1b54fba48d9777168d484b9a0732beebc1/data/hails.txt#L1796):

	word
		"selling drugs"
		"selling slaves"
		"plundering merchants"
		"fetching bounties"

With string interpolation, this could become:

 	word
		"buying $contraband"
		"selling $contraband"
		"plundering merchants"
		"fetching bounties"

Thus giving greater variety to the type of messages that can be generated.

### Considerations

[...]

PR #4538, opened 2019-09-16
Add phrase expansion and replacement functions to the phrase node.

oo13:
This PR adds 2 functions to the phrase node in order to avoid combinatorial explosion.
ref. #4533 and [this comment](https://github.com/endless-sky/endless-sky/issues/558#issuecomment-487903300).

(1) phrase expansion
A word in the phrase node can include phrase expansions:
```
phrase "buy and sell"
    word
        "buy ${contraband} from"
        "sell ${contraband} to"
```
"contraband" is a phrase name.

(2) child node "replace"
The "replace" node specifies the way of replacement some strings.

```
phrase "test phrase"
	word
		"I suspect that"
	word
		" "
	word
		"the CEO of the Syndicate%"
		"the leaders of the Syndicate"
	word
		" "
	word
		"don't have our best interests at heart."
		"are keeping secrets from the rest of the galaxy."
		"try to do right by people."
	replace
		"% do" " does"
		"% are" " is"
		"% try" " tries"
		"%" ""
```
I plan to create a translation support that needs these functions.

### DIFF
--- a/source/Phrase.cpp
+++ b/source/Phrase.cpp
@@ -139,7 +139,7 @@ Phrase::Choice::Choice(const DataNode &node, bool isPhraseName)
 	}
 	// Add the remaining text to the sequence.
 	if(entry.length() - start > 0)
-		emplace_back(string(entry, start, entry.length() - start), nullptr);
+		emplace_back(string{entry, start, entry.length() - start}, nullptr);
 }
 
 
--- a/source/Phrase.cpp
+++ b/source/Phrase.cpp
@@ -13,10 +13,30 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Phrase.h"
 
 #include "DataNode.h"
+#include "Format.h"
 #include "GameData.h"
 #include "Random.h"
 
 using namespace std;
+
+namespace {
+	// Replace oldStr with newStr in target.
+	string Replace(const string &target, const string &oldStr, const string &newStr)
+	{
+		const size_t oldSize = oldStr.size();
+		const size_t newSize = newStr.size();
+		string result(target);
+		for(size_t pos = 0;;)
+		{
+			pos = result.find(oldStr, pos);
+			if(pos == string::npos)
+				break;
+			result.replace(pos, oldSize, newStr);
+			pos += newSize;
+		}
+		return result;
+	}
+}
 
 
 
@@ -34,7 +54,22 @@ void Phrase::Load(const DataNode &node)
 		if(child.Token(0) == "word")
 		{
 			for(const DataNode &grand : child)
-				part.words.push_back(grand.Token(0));
+			{
+				const string &word = grand.Token(0);
+				bool error = false;
+				Format::Replace(word, [&](const string &s) -> string
+					{
+						const Phrase *subphrase = GameData::Phrases().Get(s);
+						if(subphrase->ReferencesPhrase(this))
+						{
+							child.PrintTrace("Found recursive phrase reference:");
+							error = true;
+						}
+						return "";
+					});
+				if(!error)
+					part.words.push_back(word);
+			}
 		}
 		else if(child.Token(0) == "phrase")
 		{
@@ -47,11 +82,21 @@ void Phrase::Load(const DataNode &node)
 					part.phrases.push_back(subphrase);
 			}
 		}
+		else if(child.Token(0) == "replace")
+		{
+			for(const DataNode &grand : child)
+			{
+				const string o(grand.Token(0));
+				const string n(grand.Size() >= 2 ? grand.Token(1) : "");
+				auto f = [o, n](const string &s) -> string { return Replace(s, o, n); };
+				part.replaceRules.emplace_back(f);
+			}
+		}
 		else
 			child.PrintTrace("Skipping unrecognized attribute:");
 		
-		// If no words or phrases were given, discard this part of the phrase.
-		if(part.words.empty() && part.phrases.empty())
+		// If no words, phrases, or replaces were given, discard this part of the phrase.
+		if(part.words.empty() && part.phrases.empty() && part.replaceRules.empty())
 			parts.back().pop_back();
 	}
 }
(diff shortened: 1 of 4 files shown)


source/Phrase.cpp
```diff
@@ -34,7 +54,22 @@ void Phrase::Load(const DataNode &node)
 		if(child.Token(0) == "word")
 		{
 			for(const DataNode &grand : child)
-				part.words.push_back(grand.Token(0));
+			{
+				const string &word = grand.Token(0);
+				bool error = false;
+				Format::Replace(word, [&](const string &s) -> string
```
tehhowch: Can we improve performance here by also calculating if `word` needs runtime replacement? Then we need not always call phrase expansion unless `word` actually needs it.

source/Phrase.cpp
```diff
@@ -13,11 +13,31 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
 
 using namespace std;
 
+namespace {
+	// Replace oldStr with newStr in target.
+	string Replace(const string &target, const string &oldStr, const string &newStr)
+	{
+		const size_t oldSize = oldStr.size();
+		const size_t newSize = newStr.size();
+		string result(target);
```
tehhowch: Use { and } for direct initialization, or use copy-initialization since we know a copy has to occur here anyway.

source/Phrase.cpp
```diff
@@ -47,11 +82,21 @@ void Phrase::Load(const DataNode &node)
 					part.phrases.push_back(subphrase);
 			}
 		}
+		else if(child.Token(0) == "replace")
+		{
+			for(const DataNode &grand : child)
+			{
+				const string o(grand.Token(0));
```
tehhowch: Use better variable names too

source/Phrase.cpp
```diff
@@ -76,7 +121,18 @@ string Phrase::Get() const
...
+			string word = part.words[Random::Int(part.words.size())];
+			word = Format::Replace(word, [&](const string &s) -> string
+				{
+					const Phrase *subphrase = GameData::Phrases().Get(s);
+					return subphrase->Get();
+				});
+			result += word;
+		}
+		else if(!part.replaceRules.empty())
+			for(auto f : part.replaceRules)
```
tehhowch: Should f be a reference?

source/Phrase.cpp
```diff
@@ -13,11 +13,31 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
 #include "DataNode.h"
+#include "Format.h"
 #include "GameData.h"
 #include "Random.h"
 
 using namespace std;
 
+namespace {
+	// Replace oldStr with newStr in target.
+	string Replace(const string &target, const string &oldStr, const string &newStr)
```
tehhowch: Should this be here, or in `Format`?

Also, you use a different algorithm for this method and for the your new Format::Replace method - any particular reason?

source/Phrase.cpp
```diff
@@ -76,7 +121,18 @@ string Phrase::Get() const
 		if(!part.phrases.empty())
 			result += part.phrases[Random::Int(part.phrases.size())]->Get();
 		else if(!part.words.empty())
-			result += part.words[Random::Int(part.words.size())];
+		{
+			string word = part.words[Random::Int(part.words.size())];
+			word = Format::Replace(word, [&](const string &s) -> string
```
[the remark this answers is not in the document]
oo13: I would only like to create the function in-place. It was no need to use [&].
Anyway, I will remove this function because it will be unnecessary.

source/Phrase.cpp
oo13: I'll change the class Part to this in order to avoid runtime replacement:
```
	class Part {
	public:
		std::vector<std::vector<std::pair<std::string, const Phrase *>>> words;
		std::vector<std::function<std::string(const std::string&)>> replaceRules;
	};
```
Phrase::Load() converts a word including phrase expansions to a vector *words* in advance.
Format::Replace I created will specialize in parsing word and be moved into Phrase.cpp.

source/Phrase.cpp
oo13: I'll fix it:
```
string result = target;
```

source/Phrase.cpp
oo13: I'll fix it:
'o' to 'oldStr'?

source/Phrase.cpp
oo13: Yes, I'll fix it.

source/Phrase.cpp
oo13: This function implements "replace" child node like  s///g including no regex, and treats the bracket ${ } as normal characters.
I would like to place here.

oo13:
A functional example for "replace" node:
```
news "test"
	location
		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Independent" "Hai"
	name
		word
			"Tester"
	portrait
		portrait/human02
	message
		word
			"I suspect that"
		word
			" "
		word
			"the CEO of the Syndicate%"
			"the leaders of the Syndicate"
		word
			" "
		word
			"don't have our best interests at heart."
			"have secrets that they aren't telling us."
			"are keeping secrets from the rest of the galaxy."
		replace
			"% do" " does"
			"% have" " has"
			"% are" " is"
			"%" ""
```

oo13:
A test data for (1) phrase expansion and (2) child node "replace":
[test_news.txt](https://github.com/endless-sky/endless-sky/files/3851554/test_news.txt)

tehhowch:
Thanks for the test data - we should also have some test phrases that will fail the self-reference check.

(There are also a couple more cases of using parentheses for direct initialization that can be replaced with braces.)

Do you think it is worthwhile to rename some of the class member variables, to be a little less confusing? We should also add comments. I was thinking the current `Phrase::parts` should actually be `Phrase::sentences`, and have the comment: `// Each time this phrase is defined, a new sentence is created.`
If it's helpful for logic and organization, we could also update the class structure, e.g. create a private class `Sentence` (a `Phrase` is one of several `Sentence`s), and each `Sentence` is composed from one or more `Part`s. The `Part` remains the most fundamental component, and each holds a list of one or more options that gets randomly picked from and converted into a string. (The options would be the `vector<pair<string, const Phrase *>>` you have implemented.)

oo13:
new test data including some phrases for the self-reference check:
[test_news_and_references_to_phrases.txt](https://github.com/endless-sky/endless-sky/files/3854129/test_news_and_references_to_phrases.txt)

> Do you think it is worthwhile to rename some of the class member variables, to be a little less confusing?

I agree.

I created three classes.
- The class Sentense represents a phrase node.
- The class Part represents a "word", "phrase", or "replace" in a phrase node.
- The type Option represents a child node in a "word" or "phrase" node.

I'm not very sure the type Option is named properly.

tehhowch:
I've opened a PR to @oo13 's branch for comment and review, if merged it will propagate here.

oo13:
Thank you for refactoring!

Issue #1185, opened 2016-06-03
[enhancement] "over" tag for engine mounts


LocalGod79:
that would cause the flare to be rendered above the ship sprite.

this would be most noticeable on the Ravens, but would allow more engine arrangements on future ships

endless-sky:
I think that would look really ugly for a lot of the existing engine flare sprites, just because I didn't make any effort to make the end of them that is hidden by the engine look good.

LocalGod79:
so, if someone gave the engine flares a touchup, you'd consider it?

endless-sky:
I'd consider it, but it doesn't seem like a high priority.

Amazinite:
https://github.com/endless-sky/endless-sky/commit/435ab5c32047d589b14bb73e380e3d1869410893

Issue #610, opened 2016-01-06
Add ship hardpoints for reverse engines to fire from


bassgoonist:
When coming out of hyperspace I've seen NPC ships use their reverse thrusters to slow down, as opposed to turning 180 and using the main thrusters, but if the flagship is equipped exactly the same it still turns 180 to slow down.

If the flagship used reverse thrusters when on 'autopilot' this could make landing on planets quicker, depending on how heavy the ship is/how fast it can turn

Gkramme:
In many ocasions, my ships have used reverse thrusters to land and to jump. They are not very powerful, however, and they are only used if it offers a quicker alternitive than spinning around to stop or to orient your ship for landing. If you have a large ship with powerful engines, it will use the quicker way instead of the slow reverse thrusters. Get a small flagship with reverse thrusters. They will be used.

bassgoonist:
Alright, I tested and it will use them to slow down if they're powerful enough, but if it's going for best efficiency you'd think it would fly around backwards all the time if you had 5 times as much reverse thrust as forward...but it doesn't.

bobafetthotmail:
we need capital-class reverse thrusters, 'nuff said.

guillochon:
The reverse thrusters really need some visual/aural indication that they are firing, that would I think make it obvious when they are being used.

LocalGod79:
@endless-sky did you just want the centre front tip mapped for each ship?  labeled 'rengine'?

Amazinite:
Implemented this in #4539 using a `"reverse engine"` keyword that works the same as `engine`, but for reverse thrusters.

Amazinite:
https://github.com/endless-sky/endless-sky/commit/435ab5c32047d589b14bb73e380e3d1869410893

PR #4539, opened 2019-09-17
Added reverse flares, steering flares, and over/under engine tags

Amazinite:
Ref: #610, #1185 

This PR adds two new ship hardpoints for different types of engines: `"steering engine"`, which fires when steering, and `"reverse engine"`, which fires when using reverse thrusters. Included along with this are `"reverse flare sprite"`, `"reverse flare sound"`, `"steering flare sprite"`, and `"steering flare sound"` attribute for outfits.
Along with this, I have also added over/under tags for engines, and allowed engines to have an angle that determines what angle they are emitted at with relation to the default angle; the default angle for engines and steering engines is 0 degrees, and the default angle for reverse engines is 180 degrees.

The following is the syntax for these changes.
```
"engine" <x> <y> [<zoom>]
	["zoom" <zoom>]
	["angle" <angle>]
	[over | under]
"reverse engine" <x> <y> [<zoom>]
	["zoom" <zoom>]
	["angle" <angle>]
	[over | under]
"steering engine" <x> <y> [<zoom>]
	["zoom" <zoom>]
	["angle" <angle>]
	[over | under]
	[left | right]
```
The left and right keywords for steering engines determines what direction the ship must be turning for those engines to fire. If these keywords are missing, then the steering engine will fire regardless of the direction of turning.

<hr/>

<details><summary>An image showing the reverse engines.</summary>

image
</details>

<details><summary>Images depicting steering right, steering left, and steering when the steering engines have been given no directional tag.</summary>

image

image

image
</details>

<details><summary>Images depicting the use of `over` engines next to the default `under` engines.</summary>

image
image
image
</details>

<hr/>

As of right now, this PR only adds the functionality and updates the reverse thrusters. It does not update the steering or add any of these new engines to the existing ships.

Where should reverse engines go? Should they fire from the guns? What about ships without guns?
Where should steering engines go? Should any existing ships use steering engines, or should we not since they weren't designed with them in mind? Should the steering engines be at the same location and angle as the existing engines, but with only having one of the "thrusters" fire depending on the direction that you turn.

<details><summary>Images depicting the steering engines being in the same location as the normal engines.</summary>

Normal thrusting.
image

Thrusting and turning in a direction.
image
image

Just turning.
image
</details>


source/Outfit.cpp
```diff
@@ -232,8 +239,22 @@ void Outfit::Add(const Outfit &other, int count)
 		else
 			oit->second += count * it.second;
 	}
+	for(const auto &it : other.reverseFlareSprites)
+	{
+		auto oit = reverseFlareSprites.begin();
```
tehhowch: use `std::find_if` to get an iterator to an existing flare sprite.

Probably you can refactor this & the regular flare sprite code to use a single method.

Amazinite: Can't seem to do any of that without breaking something.

Amazinite:
I've updated this PR to also include steering flares for feature completeness. 
I could probably also do over/under tags for engines (ref: #1185), but I'm wondering if that should be a separate PR.

source/Outfit.cpp
```diff
@@ -232,8 +246,36 @@ void Outfit::Add(const Outfit &other, int count)
...
+				break;
+		
+		if(oit == reverseFlareSprites.end())
+			reverseFlareSprites.emplace_back(it.first, count * it.second);
+		else
+			oit->second += count * it.second;
+	}
+	for(const auto &it : other.steeringFlareSprites)
+	{
+		auto oit = steeringFlareSprites.begin();
```
tehhowch: Now you *must* refactor this; it's being copy-pasted multiple times. I posted an example in discord.

source/Ship.cpp
```diff
@@ -2175,6 +2230,27 @@ bool Ship::IsThrusting() const
...
+
+
+bool Ship::IsSteering() const
+{
+	return isSteering;
+}
+
+
+
+bool Ship::SteeringDirection() const
```
tehhowch: We should use something more sensible than a boolean for direction.

source/Ship.h
```diff
@@ -225,8 +232,15 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	
 	// Check if the ship is thrusting. If so, the engine sound should be played.
 	bool IsThrusting() const;
+	bool IsReversing() const;
+	bool IsSteering() const;
+	// The direction that the ship is steering. If true, the ship is steering right. 
+	// If false, the ship is steering left.
+	bool SteeringDirection() const;
```
tehhowch: To me it would make sense if we just mirrored the current `Turn` command setting, where -1 is left and +1 is right, or use a scoped enum.

Mimicking `Turn` could let us differentiate fractional thrust, while the scoped enum would limit comparison logic by consumers.

source/Outfit.cpp
Amazinite: Created a helper method to reduce the duplicate code.

Amazinite:
>we can compute, based on the sprite, which direction any given steering engine is facing. Then, when displaying flares, we only display the appropriate flares for the direction, unless the engine was given "none" explicitly (in which case it is always displayed).

Is that really necessary, though? Seems to me that that's just another set of calculations that gets done that doesn't necessarily need to be when the directions can be explicitly stated already.

source/Ship.h
Amazinite: SteeringDirection is now a double equal to commands.Turn().

source/Engine.cpp
```diff
@@ -159,6 +159,26 @@ namespace {
 		Messages::Add(tag + message);
 	}
 	
+	void DrawFlares(Ship ship, DrawList *draw, bool calcTickTock, double cloak, vector<Ship::EnginePoint> enginePoints, vector<pair<Body, int>> flareSprites, double angle, uint8_t side)
```
Amazinite: If anyone has a better solution than sending 8 variables to a helper function, I'm all ears.

Amazinite:
>I feel steering and reverse should be separate PRs, as should over/under, even though they're all about engine specifications.

Well in looking to reduce some of the copy-pasting that had occurred in Engine.cpp, I was basically two steps away from having over/under be done, so I went ahead and added it. I can easily divide the features up at this point if desired, though.

Amazinite:
>Image 6 wouldn't be turning, it would be strafing.

For the fourth time (3 on Discord, 1 here), these images are a representation of the mechanic in action, not how the mechanic would be realistically used in game.

tehhowch:
I feel flare angling should be in its own PR, mostly because I don't agree that it's a sensible thing for a ship designer to implement - why should the ship's main engines fight each other, devoting a portion of their thrust to compressing or tensioning the ship's support structure?

As a separate PR, we can avoid delaying this one's acceptance, or even assign it to a currently neutral party for resolution.

tehhowch:
There's also an issue with steering flares: unlike forward and reverse thrust, turning is not a "hold to do more" effect: you turn at the max rate, immediately (provided you have the resources needed), for as long as you are issuing that turn command. Given that, presenting steering flares in the same manner as regular thrust flares visually suggests that the more you hold turn, the faster you turn. 

To match the game's turn implementation, I think steering flares need to be displayed for a handful of frames at the start of a turn (say, 2-5) and then the opposite flares need to be displayed for the same duration when the turn key is released.

The other alternative would be making turn work like thrust, in that it modifies a new `Body` class member, `rotation`, which is applied to the object every game step. This will make piloting considerably more difficult, as players will need to countersteer to stop turning.

Amazinite:
>To match the game's turn implementation, I think steering flares need to be displayed for a handful of frames at the start of a turn (say, 2-5) and then the opposite flares need to be displayed for the same duration when the turn key is released.

Displaying steering flares in such a way would suggest that energy is being used and heat is being produced by your steering only shortly after pressing the key and shortly after releasing it, when the truth is that energy is being drained and head is being produced for the entire duration of the steering key being pressed, not before or after, which I'd say is even more wrong than what the current implementation could incorrectly suggest. But given the drastic difference in how thrusting and steering work (thrusting ramps up, steering immediately hits max), I don't think that the flares need to visually depict the different when the movement of the player's ship already visually depicts the difference very well. A player is going to immediately notice that their ship is slow to get thrusting, but will start turning and stop turning immediately when pressing and releasing their steering button, regardless of what any flares might suggest.

tehhowch:
That's a fair rebuttal re: the steering implementation. I do think displaying steering thrust flares will highlight the physics disparity though, even if it is in line with the current side effects & requirements of turning.

jafdy:
Even if we take your suggestion it still wouldn't line up with the physics of the action, the acceleration takes place over a single frame.

Pointedstick:
This is super cool. Intuitively it seems like have them going sideways near the front of the ship would make the most sense. This would probably necessitating adding tags for every ship for where the steering flares should go. However I wonder if this would become too visually busy. If so maybe the size should just be reduced? If even after that it looks too visually busy, making one of the thrusting flares less intensive might communicate what's going on adequately without becoming visually overwhelming.

Regardless, this is a really nice addition.

ZanrG:
The over/under engine tags is a great idea, would the same tags be possible to placed on gun ports?

Zitchas:
For what it is worth, I think the default "no position specified" location for steering and reverse thrusters should be from just behind the nose of the ship, from underneath.

That being said, it would be interesting to have the option to override that in the outfit and have reverse thrusters that come from the engine nacels instead. For instance, while the generic "reverse thrusters" that are currently in-game state they are mini-engines installed in a gun hardpoint; most thrust-reverses in use today (in the Real-world) tap into the main engines and redirect some or all of the thrust forwards. For these types, having the plume wash forward over (or under) the ship from the normal engine nacels would look better.

Based on the images in the original post, and in light of PointedStick's comment about things getting too busy on the screen, I think that the steering flares in particular should be significantly smaller. I think we want steering to be more like a small detail that makes things feel more realistic, rather than a big "Look At My Flare!" sort of thing. Maybe aim at a quarter of the normal engine flare for a starting point.

source/Ship.cpp
```diff
@@ -173,7 +178,60 @@ void Ship::Load(const DataNode &node)
 				hasEngine = true;
 			}
 			enginePoints.emplace_back(.5 * child.Value(1), .5 * child.Value(2),
-				(child.Size() > 3 ? child.Value(3) : 1.));
+				(child.Size() > 3 ? child.Value(3) : 1.), (child.Size() > 4 ? child.Value(4) : 0.));
```
tehhowch, line 181: It might be easier to read these constructor calls if we compute the angle first.
Having more than one conditional operator per expression can get unwieldy fast

source/Ship.cpp
```diff
@@ -1396,6 +1504,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 				// If a reverse thrust is commanded and the capability does not
 				// exist, ignore it (do not even slow under drag).
 				isThrusting = (thrustCommand > 0.);
```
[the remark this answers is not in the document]
Amazinite: If there is no reverse thrust then there's also no reverse thrusters, and therefore no reverse thruster sprites. In practice, that is. 
But yes, having a reverse thrust sprite with no actual reverse thrust would (probably) cause the sprites to appear when hitting the back command key.

source/Ship.cpp
Amazinite: As per the comments in #4749, this syntax is probably changing anyways.

tehhowch: It makes some sense to create an intermediate class that is a bit more involved than just `Point`, that the Bay and EnginePoint derive from or contain. This class would handle the angle, facing, and z height child nodes, both read and write. (For compatibility Bay and EnginePoint will still need to parse their opening line.)
A complication here is the "launch effect" child node for Bays, since that can't be parsed by the superclass. It's also reasonable to expect that additional child nodes are added to either Bay or EP that are exclusive to their context. (Perhaps only "retrieve effect", but who knows.) In this sense, it might make sense to make the angle, facing, and z nodes be a child of their own tag, (e.g.`point`), which also serializes / deserializes the x, y, and zoom. This would make  engine specifications a little strange, since that's all they would be:
```
"[steering |reverse ]engine" [optional legacy keywords]
  point x y
    angle A
    drawn under|over
    zoom Z
```

Are there other features that could make sense on an engine specification that aren't discussed here? If so, that would be justification for being a little "wordy" like the above.

source/Ship.cpp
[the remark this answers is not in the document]
Amazinite: isReversing now looks for the presence of "reverse thrust".

source/Ship.cpp
```diff
@@ -632,7 +694,48 @@ void Ship::Save(DataWriter &out) const
...
+				out.EndChild();
+			}
+		}
+		for(const EnginePoint &point : steeringEnginePoints)
+		{
+			out.Write("steering engine", 2. * point.X(), 2. * point.Y(), point.Zoom(), point.Angle());
+			if(point.side && point.facing)
+			{
+				out.BeginChild();
+				out.Write(ENGINE_SIDE[point.side], STEERING_FACING[point.facing]);
```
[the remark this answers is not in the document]
Amazinite: Should a similar change be made to bays, then?

source/Ship.cpp
Amazinite: > It makes some sense to create an intermediate class that is a bit more involved than just `Point`, that the Bay and EnginePoint derive from or contain.

Should that be done in this PR, in #4749, or merge both and then make this change?

tehhowch: My comment above is something that should be addressed before merging either this or the other PR. I'm interested to hear thoughts from you, @petervdmeer, and other content creators

source/Ship.cpp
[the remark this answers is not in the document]
tehhowch: It can be done elsewhere

source/Ship.cpp
Amazinite: I'm leaning more away from this idea than toward it. The only things that engines and bays share between this PR and Peter's are the over/under and the angle. They also share their x/y coordinates, but that's also shared with guns and turrets, so I don't see why we'd want to make the syntax for defining bays and engines different from guns and turrets. Outside of that, only bays have their launch effect, and only engines have zoom (while only steering has its steering direction). So really we'd only be making this extra class to handle over/under and angles, which seems unnecessary to me given that we're only dealing with bays and engines.

source/Ship.cpp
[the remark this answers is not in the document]
Amazinite: I'll leave such changes for bays up to @petervdmeer in #4749.

source/Ship.cpp
```diff
@@ -165,15 +169,41 @@ void Ship::Load(const DataNode &node)
...
+			for(const DataNode &grand : child)
+			{
+				const string &grandKey = grand.Token(0);
+				if(grandKey == "zoom" && grand.Size() >= 2)
+					engine.zoom = grand.Value(1);
+				else if(grandKey == "angle" && grand.Size() >= 2)
+					engine.facing = Angle(grand.Value(1) + (reverse ? 180. : 0.));
+				else
+				{
+					for(unsigned j = 1; j < ENGINE_SIDE.size(); ++j)
```
tehhowch, line 198: Is it confusing to anyone else that we refer to "under" and "over" as a *side*? Working with this class, and reading about its usage, would probably be clearer if this was `ENGINE_Z` and `zLevel`

source/Ship.cpp
```diff
@@ -165,15 +169,41 @@ void Ship::Load(const DataNode &node)
...
+			editPoints.emplace_back(child.Value(1), child.Value(2),
 				(child.Size() > 3 ? child.Value(3) : 1.));
+			EnginePoint &engine = editPoints.back();
+			for(const DataNode &grand : child)
+			{
+				const string &grandKey = grand.Token(0);
+				if(grandKey == "zoom" && grand.Size() >= 2)
+					engine.zoom = grand.Value(1);
+				else if(grandKey == "angle" && grand.Size() >= 2)
+					engine.facing = Angle(grand.Value(1) + (reverse ? 180. : 0.));
```
tehhowch, line 195: You add 180 when it's a reverse engine and we're loading, but write it out as-is when saving. Is that right?

source/Engine.cpp
```diff
@@ -2053,18 +2096,11 @@ void Engine::AddSprites(const Ship &ship)
...
-			// three copies of the flare sprite.
-			for(const auto &it : ship.Attributes().FlareSprites())
-				for(int i = 0; i < it.second && i < 3; ++i)
-				{
-					Body sprite(it.first, pos, ship.Velocity(), ship.Facing(), point.Zoom());
-					draw[calcTickTock].Add(sprite, cloak);
-				}
-		}
+		DrawFlareSprites(ship, draw[calcTickTock], ship.EnginePoints(), ship.Attributes().FlareSprites(), Ship::EnginePoint::UNDER);
+	else if(ship.IsReversing())
```
tehhowch, line 2100: For these, we should help avoid some calls and property accesses: add a check requiring the respective enginepoint collection is not empty.

source/Ship.cpp
Amazinite: It's the same vocabulary as bays, so I'd expect that if one needs changed, both do.

Amazinite:
>One thing I'm not sure was mentioned yet: this PR adds a lot of noise to the game - currently, only when you are moving forwards does your ship play noise. With this PR, it'll play those same noises when turning or reversing, even if your ship doesn't have those enginepoints. We might want to consider an opt-out preference for 0.9.14 based on user feedback.

Yeah, that's something I was thinking about too. I'd be fine with removing the added sounds. Or, should it be the case that sounds only play if an engine point is present? In that way, no sounds would play still since none of the new engine points are implemented in this PR.

source/Ship.cpp
petervdmeer: Having the same syntax on engines, bays, guns and turrets will help content creators, so I'm happy that Amazinite took some action to get the syntaxes aligned.

Setting angles, under/over and zoom directly on the engine or bay feels more natural for content creation than adding those under a point, so changing the syntax to add the keywords under point is a NO for me.

Creating a common class is a good idea, but it also looks a bit too early for me; we are still in the process of getting to a more common syntax and let's first see how far we get with that (#4801 will also add angles to guns, making the gun also a little more common in syntax).

Zitchas:
For reverse engines, can I stack everything on one line like the engines?

ex.
`"reverse engine" -9 -177 .4 under`

Amazinite:
Only zoom is still supported on a single line. Everything else (over/under tags, angle) but be indented under the engine.

Zitchas:
OK. Right now it isn't displaying even with everything cut down to 
`"reverse engine" 0 -32`

Zitchas:
The game *is* recognizing it, though. While the above single line is all I added to the ship definition, after purchasing it, my save file shows: (this is just excerpting the engine, reverse engine, gun, and turret section for context)
```
	engine 0 30
		zoom 1
		angle 0
		under
	"reverse engine" 0 -32
		zoom 1
		angle -180
		under
	gun 0 -33
	turret 0 5.5
```

Zitchas:
OK, I currently have it working. I'm using the following:
```
	"reverse engine" 0 -32
		zoom 0.25
		angle 0
	"steering engine" -6.5 -26
		"zoom" 0.2
		"angle" 90
		over
		right
	"steering engine" 6.5 -26
		"zoom" 0.2
		"angle" 270
		over
		left
	"steering engine" -11 17
		"zoom" 0.2
		"angle" 90
		under
		left
	"steering engine" 11 17
		"zoom" 0.2
		"angle" 270
		under
		right
```

PR #4542, opened 2019-09-19
Updating Protector Description

Zitchas:
Removing the word "blaster" to reflect the fact that the Protector now carries two anti-missile turrets in addition to the blaster turrets.

Thanks to Hyugat on the ES Discord for pointing out this discrepancy.


Amazinite:
I'd call that a typo fix, so no need to update the changelog.

PR #4546, opened 2019-09-21
Creating the Tern, Ibis, and Petrel

Zitchas:
This PR adds the Tern, Ibis, and Petrel, three Remnant ships of a vintage similar to that of the Pelican. This is intended for a 0.9.11 merge.

The Tern is a durable drone initially designed for mining/harvesting from gas giants and frequently utilized to fill the "screening" role that most empires use fightercraft for. Given the Remnant don't have the population to spend lives in screening roles, they use these drones instead. At least, in their home systems. It will be unlocked with the basic license.

The Petrel is a heavy fighter intended to serve as leader for packs of drones or a light combat craft for specialty missions. They are not intended to serve as bulk wave fighters like larger factions tend to prefer. They are frequently used as low altitude craft for in-atmosphere combat. It will be unlocked with the basic license.

The Ibis is a medium carrier built to be a base of operation for a single wing of smaller craft. Depending on the circumstances, they operate as either remote mining operations or remote combat support roles.  It will be unlocked with the capital license.

link to asset zip file stored on google drive: https://drive.google.com/file/d/1n0wjXCUYvDrG45Ha6V_ZB5CfFrGPnric/view?usp=sharing

Graphics courtesy of BeccaBunny

Fully Loaded Ibis in flight:
Fully Loaded Ibis in flight

Ibis, Petrels, and Terns in flight:
IbisPetrelTernFlight


data/remnant ships.txt
```diff
@@ -560,3 +696,62 @@ ship "Starling" "Starling (Thrasher)"
...
+		"Crucible-Class Steering"
+		
+	gun -5 -20 "Inhibitor Cannon"
+	gun 5 -20 "Inhibitor Cannon"
+	turret 0 -8
+	engine -5 20.5
+	engine 5 20.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Tern is a small, hardy, drone developed to plunge into the atmosphere of nearby gas giants to collect valuable resources for the nearby Remnant worlds. As a result, these tough, thick shelled craft are surprisingly effective at providing distraction and covering fire when enemies intrude on Remnant space."
```
Darcman99: `small, hardy, drone` sounds off to me, I'm no grammer nanna but shouldn't it be:
`small, hardy drone`

Zitchas: I'm inclined to agree with you. It does look better without it.

Fixed.

Zitchas:
Created a PR to add the x2 images for these three to the High-DPI plugin:

https://github.com/endless-sky/endless-sky-high-dpi/pull/88

Zitchas:
Link to zip file with assets (.blend for each ship, and .xcf for both sprite and thumbnail for each ship, along with small and large png for sprite and thumbnail for each)

https://drive.google.com/file/d/1n0wjXCUYvDrG45Ha6V_ZB5CfFrGPnric/view?usp=sharing

Zitchas:
My apologies for confusing things. I tried to merge this PR into my long-running "Remnant holding place" PR, and ended up doing the reverse. I've reverted it, and there doesn't seem to any out-of-place data.... Except for having gained the entire commit history of my RemnantGraphics branch.

Zitchas:
OK, it appears to be fixed now.

data/remnant ships.txt
```diff
@@ -348,6 +348,86 @@ ship "Gull" "Gull (Bunks)"
...
+	
+	leak "leak" 50 50
+	leak "flame" 50 80
+	leak "big leak" 90 30
+	explode "tiny explosion" 10
+	explode "small explosion" 25
+	explode "medium explosion" 25
+	explode "large explosion" 10
+	"final explode" "final explosion medium"
+	description "This is the Ibis. A slightly carrier that serves as a mining platform for terns, as well as an escort ship for pelican fleets and a combat platform for terns used in screening roles. Stats are currently made up from whole cloth and need revision."
```
MageKing17: ...A slightly *what* carrier?

data/remnant ships.txt
```diff
@@ -413,6 +493,62 @@ ship "Pelican"
...
+		
+	gun 0 -38.5 "Inhibitor Cannon"
+	gun -10.5 -24 "Inhibitor Cannon"
+	gun 10.5 -24 "Inhibitor Cannon"
+	turret 0 2.5
+	engine 0 41.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Petrel is a multi-role mining/wing leader fightercraft designed to give remnant a forward control position for flights of Terns in higher risk or quickly changing environments. Their initial design was used as a urban mobile gun platform that flitted around structures, caverns, and natural terrain on their worlds while fighting on-planet threats."
+	description "Due to the comparatively small pool the Remnant draw on for pilots and crew they are loath to throw mass waves of people into the line of fire. As such, their fighters are not intended to be up-close glass cannons like most other factions of humanity tend to build. Instead, these ships are highly agile fighters equipped with medium range weaponry designed to snipe enemies from behind cover."
```
MageKing17: Multi-paragraph descriptions should have every paragraph after the first indented with a tab.

data/remnant ships.txt
```diff
@@ -559,3 +695,62 @@ ship "Starling" "Starling (Thrasher)"
...
+		
+	gun -5 -20 "Inhibitor Cannon"
+	gun 5 -20 "Inhibitor Cannon"
+	turret 0 -8
+	engine -5 20.5
+	engine 5 20.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Tern is a small, hardy drone developed to plunge into the atmosphere of nearby gas giants to collect valuable resources for the nearby Remnant worlds. As a result, these tough, thick shelled craft are surprisingly effective at providing distraction and covering fire when enemies intrude on Remnant space."
+	description "Unlike most empires, the Remnant do not have sufficient people to pilot fighter wings in waves likely to sustain significant casualties. As such, they frequently use these hardy drones to fulfill similar screening roles in their home systems, allowing manned craft to focus on more specialized tasks."
```
MageKing17: This one too.

data/remnant ships.txt

data/remnant ships.txt

data/remnant ships.txt
Zitchas: Thank you. Fixed.

LocalGod79:
i've only looked closely at the drone, but it seems excessively powerful.

even the wanderer Flycatcher - a T2 drone - has only 86 outfit space, and has 119 tons of mass (and the largest implemented).  the surveillance drone is 68 tons (and is the smallest).

this drone is over 200 tons: larger than any implemented fighter, let alone drone.  maybe it needs a special kind of drone bay (that only certain ships possess the attribute for)?

perhaps i'm wrong, and there's a lot of story being written that progresses the remnant technology?

Zitchas:
The Tern (drone) is supposed to be a tank. As in, its main use is getting in the way and taking fire. Outside of combat, it spends its time dipping in and out of the atmosphere of gas giants harvesting stuff. Hence the armor: Lots of re-entry plating. 

But I agree, I do need to tone down the space. It's stats read more like a fighter than a drone. To be fair, the Remnant use it like a fighter. And their fighter is more like a non-hyperspace capable light frigate. Something designed to provide forward leadership for the drones, not actually be a fighter. These ships are intended to fit into the Tier 1.5 to 2 range, albeit the most expensive segment thereof.

LocalGod79:
a fighter that looks like a T2 interceptor is the very definition of T3 - not close to 1.5-2.

if the drone is a tank, it should tank - not also have two guns and interceptor class weapon/engine space, in addition to hull attributes that would take dozens of tons of T1 and 2 outfits to achieve, with no appreciable increase in the hull mass.  i accept your multi-purpose explanation; i bet they'd put flamethrowers on them too, if they had that technology...(wink)

i think they need to be nerfed by like 50% unless you've got some story that pushes their ship tech along a millennia overnight...

even the ibis is actually a T1.5 heavy warship (if you look at it's capacities/capabilities) - not medium.

don't be discouraged - the Headhunter was originally added as a human Interceptor...

Zitchas:
I'm not discouraged. I'm happy to get some feedback. Would be happier with a straight up merge, but feedback is a close second. ;)

OK, I've done some revisions.

For the Ibis, I was originally looking at the Shield Beetle and Albatross and dropping from there. (It is currently 1600 less hull and 12400 less shields than the Albatross. In my view of things, the Ibis is a newer ship than the Albatross, but not by much. A few decades, maybe a century newer. But is still fairly old (several centuries) by the time the game starts.)

The tern (despite the label "drone" is basically the fighter for the Remnant. So falling into "interceptor" levels of power isn't really too far beyond what they're supposed to be. That being said, I agree that they were too good for their tier. Dropped the shields down to 500 and the hull down to 2500. Also reduced weapon capacity by 3. (the intention is that the player has a choice - they can either mount 2 inhibitors OR 1 PDT, but never both. And it won't fit the thrasher turret, either. Each version will be present in variants). As for the engines, though... That's the absolute smallest engine/steering pair that one can get from the Remnant. If there was a smaller set, I'd swap out the steering for it.

As for the petrel. A little bit of whittling, particularly to the top speed and acceleration side of things. Was definitely too fast.

LocalGod79:
except that a drone falling into the "power level" of a T1 interceptor *is* the definition of a T3 ship (having jumped not one, but two ship classes) - well above your intended tech level.

perhaps rather than starting with a T1.5 ship, and adding the remnant tech, start with a T2 ship, and make it less powerful.  for every improving hull attribute you add, something else needs to suffer (or you end up with power-creep, and overpowered content)

i'll probably take a closer look at the existing Remnant ships as well - cloaking as a hull attribute strikes me as odd for <T2, but i haven't looked that closely at the balance points...

Zitchas:
Well, not really. Unlike the Warship set, there isn't a direct progression between drones, fighters, and interceptors. The Republic's implementation of drones is that they are cheap sacrificial flying glass cannons, but nothing says that they *have* to be that. Likewise, strictly speaking an interceptor is just a fighter that comes with a hyperdrive. The typical human examples of them tend to be stronger than fighters, but again, there's nothing that says they have to be that way. As far as that goes, humans have interceptors that range from 1200/300 (shield/hull) up to 2500/500, with three steps in between.

This is one case where price *is* the balancing mechanic, and one area where the tier system really doesn't work: Namely, tiers don't take into account the difference between differences in quality versus differences in quantity. The republic seems more inclined to throw lots of drones at things, whereas the Remnant prefer small flights of high quality drones.  It's kind of like comparing a Chery QQ (Chinese car) with a Tesla S (American car), and then assuming that it means that all American vehicles are massively more advanced than all Chinese cars. 

On a side note, the Remnant are - by MZ's original design - the oddball faction that doesn't fit neatly into tiers. And honestly, I want them to stay that way. Neat predictable tiers are boring.

Zitchas:
Thanks for the expansion on your previous point. I look forward to seeing what you come up with. 

I'll talk to @beccabunny to see if they might be interested in making an even tinier thruster/steering unit.

I suppose worst comes to worst, I could just build the propulsion in and reduce the engine space to 0 and the outfit space down accordingly.

Zitchas:
@LocalGod79 OK, we now have a PR for a reduced-size combined thruster-steering outfit to use with the tern. I PR'd it separately to facilitate it being balanced and dealt with individually.

https://github.com/endless-sky/endless-sky/pull/4594

The summary is thus: It is roughly 75% of the size and power of the combined crucible class set, so using that, the tern could be reduced significantly in terms of mass.

Zitchas:
The drag I reduced just to keep it inline with its former values, might be upped again to slow it down a bit. 

Also removed one gun and one gun slot.

Just looking at it in the outfitter in comparison to other drones, one thing that stands out is its massive energy capacity. 3400, which massively exceeds its power needs.  Likewise, its power generation is also significantly higher than any of the other ships.
Tern:           3400 and 276
Flycatcher:    600 and 192
Flea:             491 and 93
Far Lek 14:   100 and 186
Combat:       150 and 72

Just going by random numbers, I think an appropriate place for the Tern would be around 500/175.

Alternatively, so long as they had at least a slight amount of power generation (<50) we could just leave them with the crystal capacitor and remove the millenium cell, which would cut 29 tons from the outfit space.

LocalGod79:
i feel like we're taking steps in the right direction...

i think that the number of additional hull attributes (over existing drones) should be limited to the ones available on other remnant ships - with even a few exceptions.

while hyperspace fuel capacity doesn't really get accounted for on star ships in general, <T2.1 fighters and drones need to have a 6-8ton mass allotment per jump of fuel (because they don't really have any need for the stuff, and the mothership/starships are perfectly capable of mining those resources with reasonable safety).  ramscoops in this scenario also need to be accounted for as well - in the remnants case, their ramscoop outfits are terrible, so a tiny ship should not have superior ramscoop tech for free (i suggest 10-tons base hull mass for T1 ramscooping)

any cargo space on these ships seems excessive to me - while i admire your attempt at flexibility, these ships can be *too* flexible to be balanced.  that said, having npc fighters/drones mining/collecting resources would be neat to see, but the player cannot take advantage of the gathering ai.  it's still possible to have cargo space, but the ship will need to suffer in ways i suspect you don't want it to in order to make it happen...

i kinda wish you'd added the new outfit to this pull instead of a separate one: X1050's were added with the Barb because they were part of the same ecosystem expansion...same with Baellie/Hai Carriers...

Zitchas:
Fuel & Cargo: The lore for the Tern is that it is an air-dipper. As in a ship that dips into the atmosphere of gasgiants to skim certain chemicals out of the atmosphere. The cargo hold is basically where it stores the raw material, and the fuel tank is where the refined stuff goes. That's the main reason why they're on there at all. That being said, while 5 and 50 are tiny in terms of our ships, they're still a lot of stuff. So, reducing the fuel tank from 100 down to 50, and the cargo from 10 down to 5. Thus they are still present for the lore, but reduced to the point that they are only really useful for automated fuel recovery in lore, not in-game use.

Ramscoop: This is a bit of an oversight. The Remnant ships have a scaling amount of ramscoop that basically comes down to how big a ship it is and how much wing space it is to hold ramscoop fields. The gascraft is at 0.5, so there's really no reason the Tern is at a 1. So that is reduced down to 0.5 as well. 

Sorry for splitting things up. After my initial PR that covered way too much ground, I've gotten into the habit of breaking things up into independent segments wherever possible. At least I merged the engine into this one, so we can see them and work with it - even if it is technically a separate PR.

So, summary for the moment: 
Fuel: 100 -> 50
Cargo: 10 -> 5
Regen: 1 -> 0.5

LocalGod79:
as i've said; without removing some of these features entirely, you're going to end up making sacrifices elsewhere to fit into the ecosystem properly.

i understand the lore, but it doesn't make sense to put these kinds of resources into ships that are otherwise intended to sacrifice themselves in combat.  if the Remnants had flamethrower technology (or really any other weapon that used hyperspace fuel that would fit on the small ships), it might make sense for the drones/fighters to have fuel capacity/scooping.  perhaps make a second drone that fills the non-combat roles?

this drone is still 128 tons (91o+32m+5c), and it wouldn't surprise me if M.Z.'s hard caps on drones are near 120t (~Flycatcher), and fighters near 165t (Boxwing, which *really* pushed the envelope).

also, the shield and hull regeneration rates seem a little high, for a ship this size - they're currently about 50% of their heaviest warship (while being ~5% of that ship's size)...perhaps numbers even closer to small systems core (potentially slightly lower)?

Zitchas:
Realistically, I don't see 50 fuel and 5 cargo having any real impact on anything. They add real flavour and lore, and have minimal-to-no impact on game balance and provide little-to-no mechanical advantage. (If someone can afford to buy 4 of these, the extra 20 tons of cargo space they add to the fleet is negligible, to say the least. Especially compared to the Boxwing's 80). In terms of determining "Whether or not they fit in a standard drone bay," I think these values should be ignored.  

I'm starting to wish I could just do something to restrict these to only being used by Remnant ships...

Zitchas:
OK. The Cassette-Class engine is now called the Anvil-Class engine.

I've also tweaked the Tern a bunch to be a more focused ship: It now only has a single gun hardpoint (no turret), and the outfit/weapon space has been reduced down to the minimum required to fit an inhibitor cannon. 

It is now much more focused on the "harass and screen" role and less suitable for offensive roles. In retrospect, I am not too concerned about removing the ability to mount a PDT on these, as currently the Remnant only face the Korath, which have a notable lack of missiles at the moment. 

These changes brings the Terns total max mass (including the 5 cargo) down to 112, and its normal operating mass to 107. So to me it looks like they now fit within the "universal drone bays" that all races seem to operate by. (I think there's a fundamental problem with assuming that all factions and races across the galaxy use the same principle of "drones must be tiny bare-bones ships," but that issue isn't resolvable until we have a way to limit bays to specific ships or attributes.)

@LocalGod79 In response to your question about the specific loadout, yes, there was a reason for that: The Terns aren't really designed to be supplementary firepower for destroying ships, but rather a sort of tackler ship that is designed to distract and pin down targets to allow the warships to deal with it appropriately. The combination of maneuverability and a slowing weapon means that a target ship tends to get stuck in a chasing pattern, constantly spinning around to try to track the drones around it. This, in combination with the reduced acceleration due to slowing, make the Terns quite effective in a tackling role.

data/remnant outfits.txt
```diff
@@ -359,11 +359,31 @@ outfit "Salvage Scanner"
...
+		"frame rate" 5
+	"flare sound" "plasma tiny"
+	description "This tiny set of engines were designed specifically for craft so small that they to combine both thrusters and steering into one unit."
+
 outfit "Crucible-Class Thruster"
 	category "Engines"
 	licenses
 		Remnant
-	"cost" 200000
+	"cost" 190000
```
LocalGod79: is there a reason for this change?

LocalGod79:
the tern looks much more current-ecosystem.

i agree, and have seen discussion about unique carry tokens, that would allow more flexibility when constructing alien carrier ecosystems...

on to the petrel?  currently 265 tons - nearly double any other current fighter...

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
 
 
+ship "Ibis"
+	sprite "ship/ibis"
+	thumbnail "thumbnail/ibis"
+	attributes
+		category "Medium Warship"
+		licenses
+			"Remnant Capital"
+		"cost" 30000000
```
Amazinite: This is 50% more expensive than the hull cost of the Albatross. Seems a bit much given that this is a medium warship.

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+		"shields" 22000
+		"hull" 11000
```
Amazinite: These shield and hull values seem pretty round.

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+		"required crew" 20
+		"bunks" 48
+		"mass" 300
+		"drag" 7.1
+		"heat dissipation" 0.7
+		"fuel capacity" 700
+		"ramscoop" 2
+		"outfit space" 568
+		"cargo space" 103
```
Amazinite: Outfit space and cargo space positions are flipped.

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+		"weapon capacity" 200
```
Amazinite: *Exactly* 200? As a medium warship and a carrier, I'd expect less weapon space (and maybe more engine space instead of the exact same as the Albatross).

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+		"engine capacity" 142
+		"shield generation" 6
+		"shield energy" 4.6
+		"hull repair rate" 2
+		"hull energy" 1.2
```
Amazinite: More shield and hull generation than the Albatross, but at the same energy drain?

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+		"outfit scan speed" 1
+		"tactical scan power" 39
+		weapon
+			"blast radius" 240
+			"shield damage" 2400
+			"hull damage" 1200
+			"hit force" 3000
+	outfits
+		"Inhibitor Cannon" 3
+		"Thrasher Turret" 1
```
Amazinite: You don't need to specify `1`.

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+	engine -16.5 120 0.7
+	engine 0 116.5 1
+	engine 16.5 120 0.7
+	gun 0 -135.5 "Inhibitor Cannon"
+	gun 15.5 -102 "Inhibitor Cannon"
+	gun -15.5 -102 "Inhibitor Cannon"
+	turret 0 -62 "Thrasher Turret"
+	turret 0 -20.5 "Point Defense Turret"
+	turret 0 58.5 "Point Defense Turret"
+
```
Amazinite: Extra whitespace.

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+	
+	leak "leak" 50 50
+	leak "flame" 50 80
+	leak "big leak" 90 30
+	explode "tiny explosion" 10
+	explode "small explosion" 25
+	explode "medium explosion" 25
+	explode "large explosion" 10
+	"final explode" "final explosion medium"
+	description "This is the Ibis, a multipurpose carrier that serves as a mining platform for terns, as an escort ship for pelican fleets, and a combat platform for terns used in screening roles. It is capable of fielding a single wing of support craft, consisting of two petrels and four terns."
```
Amazinite: `This is the <ship>...` isn't the way that any other ship descriptions are worded, instead using `The <ship> is...`
You need to capitalize the names of the other ships (Pelican, Petrel, Tern, Albatross).

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+	description "	The Ibis was designed shortly after the albatross as a means of extending the range of their resource collection efforts after the Remnant realized that an Alpha invsion might not be as imminent as original thought."
```
Amazinite: `invsion` -> `invasion`

data/remnant ships.txt
```diff
@@ -348,6 +347,87 @@ ship "Gull" "Gull (Bunks)"
...
+	attributes
+		category "Medium Warship"
+		licenses
+			"Remnant Capital"
+		"cost" 30000000
+		"shields" 22000
+		"hull" 11000
+		"required crew" 20
+		"bunks" 48
+		"mass" 300
```
Amazinite: This seems rather light, especially when you compare it to the heavier Pelican that is worse in HP by quite a bit (33k vs 17k). I'd say reduce the health of this carrier to be closer to that of the Pelican.

data/remnant ships.txt
```diff
@@ -413,6 +493,62 @@ ship "Pelican"
...
+		"shield generation" 1.9
+		"shield energy" 1.1
+		"hull repair rate" 0.9
+		"hull energy" 0.6
+		"outfit scan power" 12
+		"outfit scan speed" 1
+		"tactical scan power" 26
+		"asteroid scan power" 30
+		"weapon capacity" 80
+		"engine capacity" 55
```
Amazinite: Weapon capacity and engine capacity positions out of order.

data/remnant ships.txt
```diff
@@ -559,3 +695,57 @@ ship "Starling" "Starling (Thrasher)"
...
+		"shield energy" 0.4
+		"hull repair rate" 1.2
+		"hull energy" 0.9
+		"atmosphere scan" 100
+		"outfit scan power" 12
+		"outfit scan speed" 1
+		"tactical scan power" 26
+		"asteroid scan power" 30
+		"weapon capacity" 16
+		"engine capacity" 26
```
Amazinite: Weapon capacity and engine capacity positions out of order.

data/remnant outfits.txt
Zitchas: Nope, accidental. Thanks for spotting it, Fixed.

data/remnant ships.txt
Zitchas: Reduced to a bit lower than the hull price of an Albatross.

data/remnant ships.txt
Zitchas: Um. I like round numbers? Not sure if this actually needs fixing.

data/remnant ships.txt
Zitchas: Fixed.

data/remnant ships.txt
Zitchas: Weapon Space reduced to 180, and engine space increased to 160. How's that look?

data/remnant ships.txt
Zitchas: Thanks. Fixed.

data/remnant ships.txt
Zitchas: Fixed.

data/remnant ships.txt
Zitchas: Fixed.

data/remnant ships.txt
Zitchas: Fixed.

data/remnant ships.txt
Zitchas: Removed the 1

data/remnant ships.txt
Zitchas: Brought the numbers in much closer to the Albatrosses. Dropped all the values a little, since it isn't a heavy warship and doesn't have as much to regenerate. That being said, it is going to be a little more energy efficient. Not by much, though.

data/remnant ships.txt
Zitchas: Reduced both Shields and hull a bit, and upped the mass by 50 so its hull mass is now 30 higher than the Pelican, rather than 20 smaller.

data/remnant ships.txt
Amazinite: >I like round numbers?

Factions outside of Republic space don't tend to have round numbers. :)

data/remnant ships.txt
Amazinite: As with above, factions outside Republic space don't tend to have such "nice" numbers. Fudge them up or down a bit.

data/remnant ships.txt
Amazinite: No, the order was fine, but these should be up with outfit space.

data/remnant ships.txt
Amazinite: The order was fine, they're just in the completely wrong position (should be up below outfit space).

data/remnant ships.txt
Zitchas: OK, I'll break them up a bit.

data/remnant ships.txt
Zitchas: OK. Made them not nice.

data/remnant ships.txt
Zitchas: OK, changed.

data/remnant ships.txt
Zitchas: OK, moved them up with the outfit space.

data/remnant ships.txt
Zitchas: OK. Moved.

Amazinite:
I went ahead and pushed my own balance changes since I think there was so much that needed tweaked.

Summary:
* Weakened the Ibis rather substantially
  * The Ibis was looking more like a heavy warship than a medium warship, so I dropped its HP and other stats by quite a bit.
* Weakened the Petrel
  * The Petrel wasn't looking anything like a fighter, so I brought its stats down to better match other fighters.
* Altered the shield/hull ratios of the Petrel and Tern, as well as weakened their repair
  * Shifted these ships more toward the shields, and brought down their shield generation and hull repair by quite a bit given that they will be the only fighters/drones with built in repair, as well as the only fighters outside of Republic space that even have any form of repair to begin with.
* Increased the drag of the Tern and Petrel
  * These are powerful drones/fighters (high weapon space, high HP, built in repair), so I made their big downside their lower speed.
* Made the Tern and Petrel MUCH cheaper
  * With their stock outfits and the previous cost, these ships were more expensive than many human MEDIUM WARSHIPS. Brought the cost down by a LOT so that they're still the most expensive drones/fighters, but now so absurdly expensive as to cost more than medium warships.

Amazinite:
Also, I'm not sure if this ship should be used by the Remnant when you first arrive.

data/remnant ships.txt
```diff
@@ -437,6 +515,62 @@ ship "Pelican"
...
+		"Crucible-Class Steering"
+		
+	gun 0 -38.5 "Inhibitor Cannon"
+	gun -10.5 -24 "Inhibitor Cannon"
+	gun 10.5 -24 "Inhibitor Cannon"
+	turret 0 2.5
+	engine 0 41.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Petrel is a multi-role mining/wing leader fightercraft designed to give remnant a forward control position for flights of Terns in higher risk or quickly changing environments. Their initial design was used as a urban mobile gun platform that flitted around structures, caverns, and natural terrain on their worlds while fighting on-planet threats."
```
Pointedstick:
```suggestion
	description "The Petrel is a multi-role mining/wing leader fightercraft designed to provide support for flights of lighter Terns in high risk environments. The ship's design originated as an a urban mobile gun platform that flitted around structures and natural terrain, but proved easy to adapt into a spacecraft."
```

data/remnant ships.txt
```diff
@@ -437,6 +515,62 @@ ship "Pelican"
...
+	description "	Due to the comparatively small pool the Remnant draw on for pilots and crew they are loath to throw mass waves of people into the line of fire. As such, their fighters are not intended to be up-close glass cannons like most other factions of humanity tend to build. Instead, these ships are highly agile fighters equipped with medium range weaponry designed to snipe enemies from behind cover."
```
Pointedstick:
```suggestion
	description "	The Remnant are loath to sacrifice pilots and crew from the comparatively small pool they are drawn from. As such, their fighters are equipped with medium-range weaponry designed to strike enemies from afar."
```

data/remnant ships.txt
```diff
@@ -583,3 +717,57 @@ ship "Starling" "Starling (Thrasher)"
...
+		"Thermoelectric Cooler"
+		
+		"Anvil-Class Engine"
+		
+	gun 0 -22 "Inhibitor Cannon"
+	engine -5 20.5
+	engine 5 20.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Tern is a small, hardy drone developed to plunge into the atmosphere of nearby gas giants to collect valuable resources for the nearby Remnant worlds. As a result, these tough, thick shelled craft are surprisingly effective at providing distraction and covering fire when enemies intrude on Remnant space."
```
Pointedstick:
```suggestion
	description "The Tern is a small, hardy drone developed to plunge into the atmosphere of nearby gas giants for collecting valuable resources. These thick-shelled craft are also surprisingly effective when pressed into service providing distraction and covering fire when enemies intrude on Remnant space."
```

Amazinite:
>But the Tern had been hammered out into a pretty good form, so I'd rather have proper discussion of changes there.

I'm fine with swapping the HP distribution back toward the hull and having the hull repair be stronger than the shield regen, but I'd highly suggest swapping the shield regen and hull repair values that I had instead of just reverting the hull repair back to what it was. These repair values are FAR too high for a drone. Keep in mind that none of the fighters or drones in game right now have any form of shield or hull repair, outside of the Lance, Dagger, and Finch that have 12 shield repair. Yet you're giving the Tern 36 shield regen and 72 hull repair. That's 108 HP regen per second. For comparison, the 31 ton small systems core has 96 shield regen and 12 hull repair for a total of that same 108 HP regen. So comparing to the Far Lek 14, the Tern somehow manages to have 10 more outfit space, 5 more weapon space, 2 more engine space, HP regen on the scale of a 31 ton outfit, and all while only weighing 16 extra tons. I'd dare say that even the numbers I had were too high.

As for the drag change, I don't see any real downside to this drone right now, hence why I increased the drag. I like the idea of the Tern and Petrel being powerful yet slow drones/fighters. The base mass could also probably stand to be increased. For comparison, the Tern currently has 1012.5 max speed and 180/188.4 turn versus the Far Lek 14's 840 max speed and 184.6 turn.

Zitchas:
Seems reasonable. Drag has been increased, and both hull and shield regen have been dropped. 

As a form of explanation about the Tern's health stats: It is designed to withstand higher pressure environments and the stress+heat of dipping in and out of atmosphere on a regular basis. That, combined with the fact that current Remnant descriptions talk a lot about the differences in their architecture and materials, which is reflected in their hull appearances. Thus to me this is building on the fact that the shields aren't so remarkable for the amount of tech they have, but their hulls are special.

data/remnant ships.txt
```diff
@@ -372,6 +372,84 @@ ship "Gull" "Gull (Bunks)"
...
+	leak "leak" 50 50
+	leak "flame" 50 80
+	leak "big leak" 90 30
+	explode "tiny explosion" 10
+	explode "small explosion" 25
+	explode "medium explosion" 25
+	explode "large explosion" 10
+	"final explode" "final explosion medium"
+	description "The Ibis is a multipurpose carrier that serves as a mobile mining platform, an escort ship for Pelican fleets, and a combat platform using Terns for screening. It is capable of fielding a single wing of support craft, consisting of two Petrels and four Terns."
+	description "	The Ibis was designed shortly after the Albatross as a means of extending the range of resource collection efforts after the Remnant realized that an Alpha invasion might not be as imminent as originally believed."
```
MCOfficer: `as a means`, is this supposed to be plural?

data/remnant ships.txt
```diff
@@ -583,3 +718,57 @@ ship "Starling" "Starling (Thrasher)"
...
+		
+		"Anvil-Class Engine"
+		
+	gun 0 -22 "Inhibitor Cannon"
+	engine -5 20.5
+	engine 5 20.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Tern is a small, hardy drone developed to plunge into the atmosphere of nearby gas giants for collecting valuable resources. These thick-shelled craft are also surprisingly effective when pressed into service providing distraction and covering fire when enemies intrude on Remnant space."
+	description "	Unlike most empires, the Remnant do not have sufficient people to pilot fighter wings in waves likely to sustain significant casualties. As such, they frequently use these hardy drones to fulfill similar screening roles in their home systems, allowing manned craft to focus on more specialized tasks."
```
MCOfficer: sufficient people -> sufficient amounts of people ?

data/remnant ships.txt
Zireael07: Yes, this is what is called a stock expression, in 'as a means'  the word means is always plural.
(Source: am an English major)

data/remnant ships.txt
```diff
@@ -437,6 +515,63 @@ ship "Pelican"
...
+		"Crucible-Class Steering"
+		
+	gun 0 -38.5 "Inhibitor Cannon"
+	gun -10.5 -24 "Inhibitor Cannon"
+	gun 10.5 -24 "Inhibitor Cannon"
+	turret 0 2.5
+	engine 0 41.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Petrel is a multi-role mining/wing leader fightercraft designed to provide support for flights of lighter Terns in high risk environments. The ship's design originated as an a urban mobile gun platform that flitted around structures and natural terrain, but proved easy to adapt into a spacecraft."
```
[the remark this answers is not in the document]
Zitchas: I think "originated as" is a proper way of stating it.

data/remnant ships.txt
```diff
@@ -583,3 +718,57 @@ ship "Starling" "Starling (Thrasher)"
...
+		"Thermoelectric Cooler"
+		
+		"Anvil-Class Engine"
+		
+	gun 0 -22 "Inhibitor Cannon"
+	engine -5 20.5
+	engine 5 20.5
+	explode "tiny explosion" 30
+	explode "small explosion" 20
+	description "The Tern is a small, hardy drone developed to plunge into the atmosphere of nearby gas giants for collecting valuable resources. These thick-shelled craft are also surprisingly effective when pressed into service providing distraction and covering fire when enemies intrude on Remnant space."
```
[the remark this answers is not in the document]
Zitchas: Hmmm, `intrude in` would work better.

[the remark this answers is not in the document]
Zitchas: Replaced part of that sentence, so it is now:
`These thick-shelled craft are also surprisingly effective when pressed into service as screening ships when enemies intrude into Remnant space.`

I think it sounds better now.

data/remnant ships.txt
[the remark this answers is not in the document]
Zitchas: Heh. Empire seemed as good a term as anything. The Coalition feels like an empire, the Unfettered are trying to make one, the Korath *had* one and squandered it. Humanity always likes to make them..

Calling them "factions" feels like I'm almost breaking the fourth wall, but it is probably the next most reasonable term. Or "Nation."

Zitchas: changed "sufficient" to "enough," as that sounds a bit better.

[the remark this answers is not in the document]
MageKing17: I think "spacefaring nations" would be sufficiently generic without feeling meta.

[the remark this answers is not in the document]
Zitchas: I like that. Changing it.

data/remnant ships.txt
MCOfficer: yep, resolved

data/remnant ships.txt
MCOfficer:
```suggestion
	description "The Petrel is a multi-role mining/wing leader fightercraft designed to provide support for flights of lighter Terns in high-risk environments. The ship's design originated as an a urban mobile gun platform that flitted around structures and natural terrain, but proved easy to adapt into a spacecraft."
```

MCOfficer:
```suggestion
	description "The Petrel is a multi-role mining/wing leader fightercraft designed to provide support for flights of lighter Terns in high risk environments. The ship's design originated as an urban mobile gun platform that flitted around structures and natural terrain, but proved easy to adapt into a spacecraft."
```

MCOfficer:
```suggestion
	description "The Petrel is a multi-role mining/wing leader fighter craft designed to provide support for flights of lighter Terns in high risk environments. The ship's design originated as an a urban mobile gun platform that flitted around structures and natural terrain, but proved easy to adapt into a spacecraft."
```
i suppose both is valid but i felt inclined to mention it

Zitchas: Thanks for pointing it out. If both are valid, I think I will keep it as-is.

Zitchas:
@Amazinite OK, that looks good to me now.

Zitchas:
Looks good, @Amazinite

LocalGod79:
@Amazinite you're good with a 235t T1.99 fighter?

Zitchas:
Despite how it looks here, Amazinite did a direct edit to the Petrel after merging this which brings the shipyard listed mass down to 190 and the weapon capacity down to 33.

Amazinite:
@LocalGod79 It's basically a heavy fighter. I don't mind it being so much heavier than other fighters because the trade off is its power. As pointed out though, I lowered the Petrel's capabilities in https://github.com/endless-sky/endless-sky/commit/d89c69a34fe060eca65641cadea3d7f691290acd because even though it's a heavy fighter, a number of people pointed out that the gab between it and other fighters is so large that it basically makes the Petrel the only good choice as a fighter.

Amazinite:
Seems like a matter of #4610 if we want to limit bays by tonnage, which I don't see as entirely necessary; it seems rather limiting that no fighter ever can exceed some arbitrary mass limit. (Bay tonnage limits are also really just a roundabout way to add more carried ship categories, so we might as well do that instead.)

Zitchas:
Well, on the other hand, a lot of people consider most current fighters to be useless. Regardless of what they are *actually* ranked as, based on popular opinion, fighters in general seem to be considered about half a tier lower than they actually are.

PR #4555, opened 2019-09-26
Copy requiredOutfits in MissionAction::Instantiate().

### LINKED ISSUE #4554: Require outfit check always succeeds if in on complete
**Describe the bug**
If `require` is placed in `on complete`, the mission will always be able to be completed regardless of whether the player has the outfit required or now.

**To Reproduce**
1. Accept a mission where the `on complete` has a `require` 
2. Without the required outfit, note how you are still able to complete the mission.

**Expected behavior**
`require` should work `on complete` just as it does `on offer`, and just as `outfit` works in these scenarios.

 - OS: Windows 10 Home, Version 10.0.17763 Build 17763
 - Game Source: Self-compiled
 - Version: Lastest commit since posting (ac72fdb)

**Link to save file**
<details><summary>A test mission that I put together.</summary>

```
mission "Testing broken require"
	landing
	source
		attributes "quarg"
	destination "Humanika"
	on offer
		conversation
			`Go land on <destination> without a jump drive, even though you're required to have one.`
				accept
	on complete
		require "Jump Drive"
		conversation
			`It works.`
```
</details>

[Pilot just before picking up the mission](https://github.com/endless-sky/endless-sky/files/3655712/Test.Pilot.txt)
[Pilot after accepting the mission](https://github.com/endless-sky/endless-sky/files/3655723/Test.Pilot.previous-1.txt)
[Pilot after completing the mission](https://github.com/endless-sky/endless-sky/files/3655726/Test.Pilot.txt)

**Screenshots**

After picking up the job:
The mission shows on the map as being completeable, even though my ship has no JD.
image

Note how the `on complete` section of the save file doesn't properly reflect that an outfit is required.
image

Explicitly stating a number for `require` such as `require "Jump Drive" 2` yields no difference.
But if I were to change the `require` to an `outfit -1`, then that does properly write to the save file.
image

**Additional context**
It's notable that `outfit` is currently working perfectly fine on offer, accept, and complete (as seen in vanilla), and `require` is working on offer (again, it's used without issue in vanilla). It's only `require` in `on complete` that seems to be an issue.

MageKing17:
When `requiredOutfits` was implemented, it was included in `MissionAction::Save()` but not `MissionAction::Instantiate()`; since mission( action)s aren't saved until after they've been instantiated, this effectively meant that `requiredOutfits` could never be saved, and also couldn't be checked in any blocks except `on offer`, `on accept`, or `on decline`, all of which would have the same effect: prevent the mission from being offered if the `require` check failed.

This adds a line to copy `requiredOutfits` when instantiating the `MissionAction` object, allowing it to be saved in the save file and used for blocks not checked when the mission is offered.

Fixes #4554 (hopefully).

### DIFF
--- a/changelog
+++ b/changelog
@@ -17,6 +17,7 @@ Version 0.9.10:
     * Mission NPCs that have no saved system will not crash the game when the player departs a planet. (@tehhowch)
     * Ship thumbnails are now saved, just like sprites. (@MageKing17)
     * Hostile Mereti ships during the Wanderer Mind missions will no longer return understandable hails. (@Amazinite)
+    * The "require" mission tag is now evaluated in all actions, not just when the mission is offered. (@MageKing17)
   * Game content:
     * Black holes now have a landing message. (@Comnom)
     * Pirate occupation spaceport missions will no longer cause the player to launch if the occupation is in another system. (@ZBok)
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -483,6 +483,7 @@ MissionAction MissionAction::Instantiate(map<string, string> &subs, const System
 		result.events[it.first] = make_pair(day, day);
 	}
 	result.gifts = gifts;
+	result.requiredOutfits = requiredOutfits;
 	result.payment = payment + (jumps + 1) * payload * paymentMultiplier;
 	// Fill in the payment amount if this is the "complete" action.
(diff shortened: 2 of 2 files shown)


Amazinite:
I'll check if this works on my end later today.

Amazinite:
Can confirm that this fixes the issue.

PR #4561, opened 2019-10-01
Allow require outfit to search specific locations

Amazinite:
Ref: #4002, #4510

This PR allows the `require` mission action to contain children that specify where to look for the required outfit. The keywords are as follows:
* `flagship`: Check the player's flagship.
* `escorts`: Check player escorts.
* `installed`: Check the installed outfits of ships.
* `cargo`: Check the cargo of ships.
* `present`: Check ships in the player's system.
* `absent`: Check ships not in the player's system.
* `parked`: Check parked ships.
* `unparked`: Check unparked ships.

These keywords can be listed line by line or one after the other.
```
require "Jump Drive"
	flagship escorts

# That is the same as the above
require "Jump Drive"
	flagship
	escorts
```

Any combination of the above keywords can be provided, and all ships that match the provided keywords will be checked. 
If a keyword and it's opposite are *not* provided (e.g. neither `parked` and `unparked` or `installed` and `cargo` are provided), then both are checked. An exception to this is if `parked` or `absent` are specified without specifying `flagship` or `escorts`; in this case, only the escorts are checked, as the flagship may never fall under `parked` or `absent`.
If no keywords are provided, then the current behavior is used of only checking the flagship when `require` > 0 and checking everything if `require` == 0.

I believe that this implementation includes a fix for the same bug that #4510 fixes.
This PR does not do anything to the `outfit` mission action.


source/MissionAction.cpp
```diff
@@ -195,14 +222,41 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
...
 		}
 		else if(key == "require" && hasValue)
 		{
 			int count = (child.Size() < 3 ? 1 : static_cast<int>(child.Value(2)));
+			
+			OutfitLocations location;
+			for(const DataNode &it : child)
+			{
+				location.empty = false;
+				for(int i = 0; i < it.Size(); ++i)
```
MageKing17: This should be a range-based for loop.

source/MissionAction.cpp
```diff
@@ -356,32 +420,126 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
...
+		if(checkEscorts)
+		{
+			checkFlag = true;
+			for(const auto &ship : player.Ships())
+			{
+				// Ignore destroyed ships.
+				if(ship->IsDestroyed())
+					continue;
+				// The first ship in list is the flagship, so skip over it.
+				if(checkFlag)
```
MageKing17: It might be better to do an explicit comparison between `ship` and `flagship`, in case the flagship is ever not the first ship.

source/MissionAction.cpp
```diff
@@ -356,32 +420,126 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
...
+					continue;
+				}
+
+				if(checkCargo)
+				{
+					if(checkPresent && ship->GetSystem() == player.GetSystem())
+					{
+						// If the player is landed and not parked then check the player's pooled cargo
+						if(player.GetPlanet() && !ship->IsParked())
+							available += player.Cargo().Get(it.first);
```
MageKing17: This code is executed on a per-ship basis, meaning that being landed will erroneously result in the player's count of this outfit in cargo being multiplied by the number of escorts present.

source/MissionAction.cpp
Amazinite: How would I go about doing that? Everything I tried didn't seem to work.

MageKing17: `ship` is a `shared_ptr<Ship>` while `flagship` is a straight-up pointer to a `Ship`; therefore you either need to compare `ship` to `player.FlagshipPtr()` or `ship.get()` to `flagship`. So, most simply:
```cpp
if(checkFlag && ship.get() == flagship)
{
	checkFlag = false;
	continue;
}
```

source/MissionAction.cpp
Amazinite: How would I do that? Looking around, normal for loops are used for traversing the tokens of a DataNode in a similar fashion in [NPC.cpp](https://github.com/endless-sky/endless-sky/blob/07dd814209ed1e1ccf9547fd7fa86fa340097cf8/source/NPC.cpp#L49) and [Interface.cpp](https://github.com/endless-sky/endless-sky/blob/6dc743e3217756e2d031a231af05dc8a8e06076f/source/Interface.cpp#L41).

source/MissionAction.cpp
Amazinite: The latest commit should have fixed this by only checking the player's pooled cargo once before entering this for loop. Also fixed it by not counting the pooled cargo a second time if it was already counted for the flagship.

source/MissionAction.cpp
tehhowch: In #3128  I extend `DataNode` to expose the tokens vector. Without that, the begin/end iterators invoked by the range-based for loop would be accessing child `DataNode`s, rather than the string tokens.

source/MissionAction.h
```diff
@@ -40,6 +40,22 @@ class UI;
 // include showing a dialog or conversation, giving the player payment or a
 // special item, modifying condition flags, or queueing an event to occur.
 class MissionAction {
+public:
+	
+	class CheckLocations {
+	public:
+		bool empty = true;
+		
+		bool flag = false;
```
tehhowch: I think it could make more sense to use a bit flag like the NPC `actions` and an enum for these values. If you prefer having them "separate" still, you could use a `vector<bool>` with the enums being associated with valid vector indices.

source/MissionAction.cpp
MageKing17: If this is the currently-accepted paradigm, then it's fine until/unless the aforementioned PR gets merged.

Amazinite:
Not planning on wrapping this up any time soon, and similar to #4601, I want to do a fair amount of reworking anyways, so I'm closing this until I'm ready to pick it back up. (Instead of a vector of booleans, it makes more sense to me to use an enum like Missions do for triggers, or really any other system that is already used elsewhere in the code, because I haven't spotted anything like what I'm doing here. I'm sure that was mentioned back when I was first working on this, but we're nearly a year on from the posting of this, and at this point I have a better understanding of the code base to know now that this looks pretty clumsy.)

PR #4563, opened 2019-10-01
Pug staff

Brick63:
Add a new Hand to Hand weapon for the pug

[pug staff assets.zip](https://github.com/endless-sky/endless-sky/files/3678229/pug.staff.assets.zip)


data/pug.txt
```diff
@@ -319,6 +319,18 @@ outfit "Pug Biodefenses"
...
 
+outfit "Pug Peacekeeping Staff"
+	plural "Pug Peacekeeping Staves"
+	category "Hand to Hand"
+	cost 11000
+	thumbnail "outfit/pug staff"
+	"capture attack" 0.8
+	"capture defense" 1.5
+	"unplunderable" 1
+	description "These tools designed for the control of planetary populations are also good at protecting their ships. Using minor amounts of radiation and electric shocks it Non-Lethally removes the complete ability to oppose the pug."
```
petervdmeer: I don't think "Non-Lethally" needs to be capitalized.

Brick63: yeah, i just have a bad habit of accidentally capitalizing things sometimes, ill fix it

data/pug.txt
```diff
@@ -317,6 +317,16 @@ outfit "Pug Biodefenses"
 	"unplunderable" 1
 	description "This defense system is designed to release a suite of toxins that are lethal to human beings (and probably to some other species, as well) but which have no effect at all on the Pug."
 
+outfit "Pug Peacekeeping Staff"
+	plural "Pug Peacekeeping Staves"
```
Amazinite: "Staffs" is a more common plural form of staff, so no need for this.

data/pug.txt
```diff
@@ -317,6 +317,16 @@ outfit "Pug Biodefenses"
...
+	category "Hand to Hand"
+	cost 11000
+	thumbnail "outfit/pug staff"
+	"capture attack" 0.8
```
Amazinite: I'm somewhat tempted to say give this a higher capture attack since the defensive capability is pointless when the Pug have biodefenses.

data/pug.txt
```diff
@@ -317,6 +317,16 @@ outfit "Pug Biodefenses"
...
+	"capture defense" 1.5
+	"unplunderable" 1
+	description "These tools designed for the control of planetary populations are also good at protecting their ships. Using minor amounts of radiation and electric shocks it non-lethally removes the complete ability to oppose the pug."
```
Amazinite: Capitalize Pug.
Also, I'd say this description could be more interesting when the way it's described in the story is as follows.

>With a somewhat contemptuous gesture, the Pug points the staff toward your face and a wave of wonderful rainbow light washes over you and bathes the whole world in iridescent sparkles that leap nimbly from the whiskers of things and nuzzle into your contented consciousness like flashing minnows of pure joy. You caper off to chase a trail of floating soapy baubles that burst with the sound of wind chimes and distant trumpets as you touch each one and the smell of ripe peaches fills the air until you stagger through your airlock and throw up in a corner.

data/pug.txt
tehhowch: Is it though? Staffs is also a verb, so comparing word usage in general isn't straightforward. When talking about a beat-you-on-the-head stick, "staves" sounds better to me than "staffs." Plus, I think "staves" is better-suited to the Pug.

Zitchas: I agree. Whenever there are multiples in books (fantasy and/or medieval genre), they generally use staves in my experience.

data/pug.txt
Amazinite: How about "Despite being designed for planetary population control, peacekeeping staves also work exceptionally well in ship defense. The staff emits a short-ranged beam that immediately incapacitates anyone it touches by causing them to enter a euphoric and hallucinogenic state. wandering off to somewhere where they won't cause any trouble. Exiting this euphoric state is never pleasant, though, often resulting in vomiting."

data/pug.txt
Amazinite: How about something >1.6 but <2.8 so that it falls between the Korath rifle and nerve gas?

Brick63: bumped both to 2.3 to give it a unique setup

data/pug.txt
```diff
@@ -317,6 +317,16 @@ outfit "Pug Biodefenses"
...
 
+outfit "Pug Peacekeeping Staff"
+	plural "Pug Peacekeeping Staves"
+	category "Hand to Hand"
+	cost 11000
+	thumbnail "outfit/pug staff"
+	"capture attack" 2.3
+	"capture defense" 2.3
+	"unplunderable" 1
+	description "Despite being designed for planetary population control, peacekeeping staves also work exceptionally well in ship defense. The staff emits a short-ranged beam that immediately incapacitates anyone it touches by causing them to enter a euphoric and hallucinogenic state. wandering off to somewhere where they won't cause any trouble. Exiting this euphoric state is never pleasant, though, often resulting in vomiting."
```
Amazinite: `hallucinogenic state. wandering` -> `hallucinogenic state, wandering`

data/pug.txt
```diff
@@ -317,6 +317,16 @@ outfit "Pug Biodefenses"
 	"unplunderable" 1
 	description "This defense system is designed to release a suite of toxins that are lethal to human beings (and probably to some other species, as well) but which have no effect at all on the Pug."
 
+outfit "Pug Peacekeeping Staff"
+	plural "Pug Peacekeeping Staves"
+	category "Hand to Hand"
+	cost 11000
```
Amazinite: Laser Rifle: 8k @ .6 attack
Pulse Rifle: 11k @ .7 attack
Tuning Rifle: 65k @ 1.2 attack
Frags: 17k @ 1.3 attack
Korath Rifle: 73k @ 1.6 attack

1) This should probably be more expensive
2) I think we should have a look at the cost of some of the new H2H vs its attack.

Amazinite: I'd say 102k for cost.

data/pug.txt
```diff
@@ -317,6 +317,16 @@ outfit "Pug Biodefenses"
...
 
+outfit "Pug Peacekeeping Staff"
+	plural "Pug Peacekeeping Staves"
+	category "Hand to Hand"
+	cost 123456
+	thumbnail "outfit/pug staff"
+	"capture attack" 2.3
+	"capture defense" 2.3
+	"unplunderable" 1
+	description "Despite being designed for planetary population control, peacekeeping staves also work exceptionally well in ship defense. The staff emits a short-ranged beam that immediately incapacitates anyone it touches by causing them to enter a euphoric and hallucinogenic state, wandering off to somewhere where they won't cause any trouble. Exiting this euphoric state is never pleasant, though, often resulting in vomiting."
```
Pointedstick: `Despite being` could be shortened to `Though with no loss of meaning

`peacekeeping staves also work` -> `the peacekeeping staff also works` (otherwise it's referred to in the plural in one sentence, and in the singular in the next)

`well in` -> `well for`

`immediately` is probably not needed

`Exiting this euphoric state is never pleasant, though, often resulting in vomiting.`
->
`Coming down from the state of euphoria is never pleasant.` (or something like that; I don't think mentioning the vomiting is really necessary)

PR #4566, opened 2019-10-02
Add a new Hand to Hand weapon for humans

Brick63:
adds a laser pistol, wont really change much but seems like a harmless addition

[laser pistol assets.zip](https://github.com/endless-sky/endless-sky/files/3743700/laser.pistol.assets.zip)
laser pistol@2x


data/outfits.txt
```diff
@@ -389,6 +389,15 @@ outfit "Nerve Gas"
...
 	description "This is a highly illegal weapon for hand to hand combat when boarding an enemy ship. Because chemical weapons are banned by interstellar law, you can be severely fined for getting caught in possession of nerve gas canisters. But, they do clear out an enemy ship quicker than just about anything else..."
 
+outfit "Laser Pistol"
+	category "Hand to Hand"
+	cost 4000
+	thumbnail "outfit/sidearm"
+	"capture attack" .3
+	"capture defense" .4
+	"unplunderable" 1
+	description "A high quality sidearm that offers increased protection for less than a rifle, it is often times used to dual wield it and a standard sidearm. The result of this action varies, but generally don't end up well."
```
MageKing17: The last sentence is ungrammatical: "The result of this action varies, but generally don't end up well."

You either need to change it to be all plural ("The **results** of this action **vary**, but generally don't end ~~up~~ well.") or all singular ("The result of this action varies, but generally **doesn't** end ~~up~~ well.").

Brick63: woops. missed that, thanks

data/outfits.txt
```diff
@@ -389,6 +389,15 @@ outfit "Nerve Gas"
 	"unplunderable" 1
 	description "This is a highly illegal weapon for hand to hand combat when boarding an enemy ship. Because chemical weapons are banned by interstellar law, you can be severely fined for getting caught in possession of nerve gas canisters. But, they do clear out an enemy ship quicker than just about anything else..."
 
+outfit "Laser Pistol"
+	category "Hand to Hand"
+	cost 4000
+	thumbnail "outfit/sidearm"
```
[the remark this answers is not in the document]
Brick63: its just what i had it named up till now, ill change it

Brick63:
they are about as blocky as the original laser rifle, which looked more uncomfortable to hold, and just because few outfits have glowing bits doesn't mean more shouldn't have it.

and why is it not a good thing, The M16 was called a toy when it was first introduced, now it defines what we think of when we imagine a service rifle. Flashy will sell more than bland. 

Modern guns were not held to the standards of Muskets, so why would lasers be held to the standards of an M4. Change brings with it new looks.

Bladewood:
I think your new rifle is much more uncomfortable than the old one, which would have been wielded differently because of how much more room there is around the trigger.
The old rifle is blocky, but it's better done because it's spaced out enough and it's not around *the parts that people would be holding on to.* Their blocks are also much smoother and significantly less repetitive, and clearly have some direction to them - there is a clear sense that something is *changing* as you move along the barrel because it's smoothing out, while the growing yellow bits at the front create good contrast.
Here, on the other hand, there is no direction: it's perfectly straight and the "rail" bits aren't changing along the barrel. There is a change in the middle, but it's very abrupt, disturbing the pattern unnaturally.

Lights aren't bad. Poorly done, repetitive lights without any direction are.

It's bad because in my eyes that kind of Sci-Fi art has become a cliché. I'd perceive it more as a joke than as a serious, dark portrayal of the future, which is how I view Endless Sky.
"New looks" do not preclude usability.

Brick63:
the old ones edges look sharp, the stock is a mess, it has less quality than other Hand to hand weapons. it leaves the humans with a second rate design. Space has been addressed and added. and the blocks are repetitive on the old one too.

there is a direction, Forward.

Cliche is there, but it exists because there is a general agreement that this will probably look like this, so we will do it as such. Why would we go to the effort of making an assault rifle looking weapon if its simpler to make a new looking weapon.

Brick63:
updated the image to one that looks better

Pointedstick:
+1 on the new Laser Rifle image. It looks much cooler than the old one, which is quite chunky and crude IMO. However that should be changed in a separate commit so we can debate the merits of adding the laser pistol separately.

On that subject, I am also not really sold on it. It's just a cheaper and worse version of the laser rifle ("cheaper" being relative; the fact that a handheld Laser Rifle sells for 16 times the price of a meteor missile strikes me as somewhat absurd).

If you want to provide some interesting balancing here, make the pistol have a higher defense than the rifle, which would be somewhat true to life: pistols are defensive weapons useful for close-quarters combat, while rifles are offensive weapons, used for attacking faraway defended positions.

Brick63:
Yeah, I'll work on a second PR.

Amazinite:
I don't get the sense that this adds much of value to the game. The laser rifle is intentionally the cheapest and weakest H2H weapon. If the rationale behind adding a laser pistol is to add another step down that's easier to purchase (not that the laser rifle is already super easy to get), then where does that rationale end? And there's also the fact that the player and crew are already equipped with pistols to begin with, although they're likely normal weapons instead of laser weapons, but I don't see a laser pistol being all that much better than a normal weapon.

I'm interested in updating the current laser rifle image (the trigger on the current laser rifle is like half way between both ends of the weapon...), but I don't see the need to add a laser pistol.

Issue #4567, opened 2019-10-03
Making the Arrow actually fast.


Hyugat:
Currently, the Arrow serves hardly any purpose within the game. I propose that we reduce the Arrow's drag (and possibly increase engine capacity as well) in order to give it some sort of purpose, at least a fun factor for the player using it.   Currently, the arrow's description states the following: "The Arrow is a light transport, designed for bringing a small number of very rich passengers - Syndicate CEOs, for example - from one planet to another with speed and style. It makes up for its light armament by **being able to outrun nearly every other ship in the sky."**

Despite its description specifically noting that it is a speedy ship, the arrow can be outrun by the following: any interceptor with similar engines (some of which exist already in the game as variants), a Clipper with atomic engines, a falcon with atomic engines, and a **stock** Sparrow (a very common ship for pirates to use). I know it says "nearly every other ship", but this is a little too slow for my liking, and many other players would likely agree.

Does anyone in the es universe really think that important Syndicate CEOs should be transported in a ship that is outrun by common pirates? Should a ship engineered specifically for speed be slower than a common interceptor by default? Should the Syndicate's most important leaders really be so unsafe?

I propose that we decrease the value for the Arrow's drag (along with all the marauder variants) down to 2.0, from 2.7, making it faster than interceptors with the same engines, but still slower than certain ships: Some of the marauder engines variants, a Bounder with atomic engines, and the Flivver. This way its description would still be perfectly accurate, and it would be at least a little more useful. This change buffs the arrow's stock max speed from around 606 to 819, making it quite a speedy ship, but not unreasonably so. The Marauder engines Arrow's speed goes from around 1057, to a whopping 1428. (The min-maxed ship is still slower than a min-maxed Marauder Fury, the ship with the highest potential for speed in the game.) This makes these ships more fun to use, but doesn't in my opinion detract at all from the game in any way (it's not like this ship is OP, and it fits within the lore).

 If we want to further increase the Arrow's fun factor, we can buff the engine space from 60 to 70, or even 80 (buffing the Marauder Arrows as well, making them even more fun to zoom around in), but this change comes with a few issues (creating the new fastest ship in the game and such, making the engines variant be able to reach even more ridiculous speeds, etc.).

TLDR: Reduce the Arrow's drag (and maybe buff engine space) in order to to make it have at least some sort of niche or fun factor.

Zitchas:
[...]

PR #4569, opened 2019-10-03
Tweaks to some human transport ship stats to make them more viable choices

Zitchas:
In response to Issue #4567 "Making the Arrow Actually Fast", a proposed solution. 

<details><summary>Old PR description.</summary>Note: I'm not saying that this is definitely needed, but issue #4567 requested this change, so I'm PRing this as a possible solution that people can try out. This is probably a "minimal change" version that doesn't come anywhere near allowing the sorts of permanent speed boosts requested in that issue, but it would definitely make the Arrow a bit faster and give it better outfitting potential to be very zippy.

This PR changes:
- the Arrow's drag from 2.7 to 2.6 (increases base max speed from 606 to 630)
- the Arrow's engine space from 60 to 74 (which would give it the space to install an afterburner)

In stock configuration, this has a slight change, most notably it increases its speed to just barely be faster than a combat drone, and only slightly slower than a stock sparrow; thus improving its "run away" chances.

It does open things up for people interested in making a ship that relies on running away by allowing a little more flexibility in installing alternate engine configurations or adding an afterburner.</details>

<hr/>

@Amazinite here, editing the OP to explain the changes since Zitchas implemented stuff that I asked.

* Arrow and Star Queen fuel increases: A sign of a good transport ship should be its ability to travel long distances without needing to refuel. These two ships were on the lower side in terms of fuel capacity when compared to the other human transport ships, so they've had their fuel capacity increased.
* Arrow drag reduction: Despite the claim in the ship's description that it's one of the fastest ships around, the potential for the Arrow to go fast was actually very low. Reducing the drag makes the Arrow a truly fast ship, with its base max speed going from 607 to 819, as compared to the Flivver's base max of 924,
* Arrow engine space increase: One of the key attributes of stock ships is that they leave room to upgrade. The Arrow though already comes with atomic engines, so there's really no upgrade room. Increasing the engine space gives the Arrow some upgrade room, and allows it to reach even greater speeds.
* Star Queen drag reduction: The Star Queen has been a ship that many have considered useless; a transport ship with a high required crew that cuts into your profits. The ship isn't helped by its low speed either, although would could argue that it's intentionally low since the only people who use the ship are in a pirate free region of space. Still, the ship's drag has been reduced by a decent amount to make it more appealing.
* Centipede drag and mass changes: This ship's drag was reduced slightly to compensate for the Star Queen's drop in drag, as the Centipede was intentionally designed as an upgrade to the Star Queen. The mass was changed because when I was creating the Centipede, my spreadsheet had a typo for the Star Queen mass that resulted in my making the Centipede's mass lower than it should have been.

### DIFF
--- a/data/hai ships.txt
+++ b/data/hai ships.txt
@@ -75,7 +75,7 @@ ship "Centipede"
 		"required crew" 35
 		"bunks" 131
 		"mass" 187
-		"drag" 4
+		"drag" 3.9
 		"heat dissipation" .75
 		"fuel capacity" 600
 		"cargo space" 38
--- a/data/hai ships.txt
+++ b/data/hai ships.txt
@@ -74,7 +74,7 @@ ship "Centipede"
 		"hull" 2400
 		"required crew" 35
 		"bunks" 131
-		"mass" 168
+		"mass" 187
 		"drag" 4
 		"heat dissipation" .75
 		"fuel capacity" 600
--- a/data/marauders.txt
+++ b/data/marauders.txt
@@ -19,9 +19,9 @@ ship "Marauder Arrow"
 		"required crew" 2
 		"bunks" 5
 		"mass" 145
-		"drag" 2.5
+		"drag" 2.0
 		"heat dissipation" .85
-		"fuel capacity" 400
+		"fuel capacity" 600
 		"cargo space" 5
 		"outfit space" 210
 		"weapon capacity" 55
--- a/data/marauders.txt
+++ b/data/marauders.txt
@@ -19,13 +19,13 @@ ship "Marauder Arrow"
 		"required crew" 2
 		"bunks" 5
 		"mass" 145
-		"drag" 2.7
+		"drag" 2.5
 		"heat dissipation" .85
 		"fuel capacity" 400
 		"cargo space" 5
 		"outfit space" 210
 		"weapon capacity" 55
-		"engine capacity" 80
+		"engine capacity" 95
 		"self destruct" .25
 		weapon
 			"blast radius" 27
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -2965,7 +2965,7 @@ ship "Star Queen"
 		"required crew" 41
 		"bunks" 112
 		"mass" 230
-		"drag" 5.5
+		"drag" 4.5
 		"heat dissipation" .65
 		"fuel capacity" 700
 		"cargo space" 60
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -141,13 +141,13 @@ ship "Arrow"
 		"required crew" 1
 		"bunks" 5
 		"mass" 130
-		"drag" 2.6
+		"drag" 2.5
 		"heat dissipation" .85
 		"fuel capacity" 400
 		"cargo space" 10
 		"outfit space" 180
 		"weapon capacity" 50
-		"engine capacity" 74
+		"engine capacity" 75
 		weapon
 			"blast radius" 24
 			"shield damage" 240
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -141,13 +141,13 @@ ship "Arrow"
 		"required crew" 1
 		"bunks" 5
 		"mass" 130
-		"drag" 2.7
+		"drag" 2.6
 		"heat dissipation" .85
 		"fuel capacity" 400
 		"cargo space" 10
 		"outfit space" 180
 		"weapon capacity" 50
-		"engine capacity" 60
+		"engine capacity" 74
 		weapon
 			"blast radius" 24
 			"shield damage" 240
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -141,9 +141,9 @@ ship "Arrow"
 		"required crew" 1
 		"bunks" 5
 		"mass" 130
-		"drag" 2.5
+		"drag" 2.0
 		"heat dissipation" .85
-		"fuel capacity" 400
+		"fuel capacity" 600
 		"cargo space" 10
 		"outfit space" 180
 		"weapon capacity" 50
@@ -2967,7 +2967,7 @@ ship "Star Queen"
 		"mass" 230
 		"drag" 5.5
 		"heat dissipation" .65
-		"fuel capacity" 500
+		"fuel capacity" 700
 		"cargo space" 60
 		"outfit space" 360
 		"weapon capacity" 120


Amazinite:
Humans like multiples of 5. Outside of the Combat Drone and Surveillance Drone (designed to just fit their outfits), Barb and Boxwing (thanks Local), and the Wasp's weapon space (intentionally designed to just fit a Proton Gun), all human ships have an outfit space, weapon space, and engine space that's a multiple of 5. 
Bump that 74 up to a 75.

Nescio0:
Let's have a look at the engine capacity / drag of human transports (higher ratio means higher maximum velocity):
```
Arrow         :  60 / 2.7 = 22.2
Blackbird     : 110 / 5.0 = 22.0
Bounder       : 110 / 3.7 = 29.7
Flivver       :  45 / 1.0 = 45.0
Heavy Shuttle :  60 / 2.1 = 28.6
Scout         : 100 / 3.4 = 29.4
Shuttle       :  60 / 1.7 = 35.3
Star Queen    : 100 / 5.5 = 18.2
```
As you can see the Arrow is actually not that fast and could use an improvement. Although this proposal makes it a faster ship (74 / 2.6 = 28.5), it is still below most other transports.
A more elegant approach would be to keep the engine capacity unchanged, but cycle the drag of Arrow, Heavy Shuttle, and Shuttle:
* Arrow: 2.7→1,7
* Heavy Shuttle: 2.1→2.7
* Shuttle: 1.7→2.1
This would put the Arrow above the Bounder and Scout but below the Flivver.

Hyugat:
Here's the thing about the Arrow. People on the Discord have been making comparisons to cars and such, so I'll say this here. I don't think of the Arrow as a limo. I think of the arrow as something akin to a Ferrari or Bugatti, just like the Flivver. Keep in mind that ships in ES are on a larger scale than cars. By default, the Arrow already has super powerful engines. It's faster than most stock ships by default, but that's because it already has those expensive A250-series Atomic engines. The Arrow still being this slow by default, makes it serve basically no purpose within the game.

 It is specifically created to be fast, yet has a disproportionately high drag value for its engine space. I propose again that we give this ship 2.0 drag, to match its appearance (just look at the thing), along with this engine space buff, giving players an actual reason to want to use the ship (that being insane blistering speed). If I'm a Syndicate CEO spending 3 million credits on a tiny transport ship, fast enough just isn't fast enough. I would want this ship to move at breakneck pace to outrun almost anything imaginable, and keep me as safe as possible from pirates. 

Even with such a drastic change, it's still not very viable for transport or courier missions (not really having much of a practical purpose), but at least it has SOMETHING. Extremely fast speed being the Arrow's main schtick makes sense within the game, and it's not like it's gamebreakingly good (the Arrow is still bad, but again, has something). I see no reason for it not to be even more fast.

Also, you probably want to adjust the marauder Arrows as well.

Brick63:
i see the arrow as a spaceship

Zitchas:
When I think of the arrow, I prefer not to think of cars. Aircraft are much more applicable. In this case, I think of something like the [Airbus A318 ELite](https://www.aircharterservice.ca/aircraft-guide/private/airbus-europe/airbusa318elite) or the [Bombardier Learjet 315](https://www.aircharterservice.ca/aircraft-guide/private/bombardier-canada/bombardierlearjet3536). These are both jets the fulfill roughly the same parameters in our realworld as the Arrow does in Endless Sky. Both have a listed cruise speed of 540mph/879km/h which is *massively* slower than almost *all* fighter aircraft currently in service (in the so-called "Western world" anyway)

Anyway, all that aside, I reduced the drag on all the arrows down to 2.5, which means the stock speed the common arrow is now 655, which is a bit faster than the sparrow. Nothing short of the Flivver can catch one now. Which makes sense, given that the Flivver is even more focused on speed for the rich traveler than the arrow is.

I really don't think the Arrow should get much (if any) faster than that. It is now capable of outrunning anything human except for things that are built specifically for chasing down fast things like this, and have the potential to get a lot faster in the player's hands.

Nescio0:
You can install atomic engines in about any ship. Looking only at the maximum velocity is not a fair comparison. What matters is the engine capacity to drag ratio. Your proposal (75/2.5=30.0) makes the Arrow faster, though it is still below the much cheaper Shuttle (60/1.7=35.3), Sparrow (40/0.9=44.4), or Wasp (60/1.2=50.0).

Zitchas:
OK, here's an example of where I have a problem with adjusting stats just based on preference and not based on *what the ship actually is.*

Arrow is (base, without outfits) 130 tons.
Shuttle is 70.
Heavy shuttle is 110.
Wasp is 60.
Sparrow is 50.

To me, it makes perfect sense that in our weird physics world of ships having top speeds, that their top speed should be connected in some way to the size of a ship. Stick a rocket on a smaller ship, it makes sense that it will go faster than if it is attached to a bigger ship. As a general rule, the bigger the ship, the higher the drag. As far as I can tell, the only reason the two aren't directly tied together is so that there was some leeway for making some ships "better built for speed" than others at the same weight point. 

No matter how well built a ship is, no matter how "optimized for speed" it might be, the hard physics dictate that if you have the same thrust on a ship that has half the mass, it is going to go faster, roughly twice as fast.. This is made up for in the fact that the Arrow has (now) 75 tons of engine space and 180 tons of outfit space versus the shuttle's 120 and 60. 

This is the problem with looking at absolute capacities: When one gets right down to it, we're in space. The only real thing that should determine how fast a ship goes is its mass and how much thrust it can deliver. The shape of the metal really doesn't have much bearing on anything, nor does the material its made of. Could be a ten million nuyen ship or a half million starter; space and physics don't really care. 

This is why the arrow *starts* with atomics, and the shuttle starts with ionics. Because the absolute capacities of both ships fall under the same physics, and are extremely similar, with the advantage going to the ship that is smaller. So the more expensive ship has more expensive engines that can produce more thrust. 

Also, installing atomics on a shuttle doubles the cost of the ship. The fact that this enables it to go faster than the arrow is - to me - the physics and drag working as intended. This is also why this PR adjusts the *engine capacity* of the ship rather than just cranking down the drag. This gives the player the option to install more stuff in it. 

More engine = more thrust = a good solid reason for why the ship should be going faster. 

That being said, if the balancing-powers-that-be think the arrow needs more speed, I think increasing its outfit space and increasing its engine space some more would be a much better route.

Zitchas:
On Amazinite's suggestion, reduced the drag to 2 and increased the fuel to 600.
Also as per Amazinite's request, made some tweaks to the Star Queen and Centipede:
- Centipede mass up to 187 (fix for Amazinite)
- Star Queen fuel up to 700.
- Centipede drag 4 to 3.9 (minor correction for Amazinite)
- Star Queen drag 5.5 to 4.5

Amazinite:
Updated the PR description explaining the new changes.

PR #4580, opened 2019-10-06
Remnant Tweak: Giving Albatrosses and Pelicans Scram Drives

Zitchas:
This PR replaces the Hyperdrives on Albatrosses and Pelicans with Scram Drives; and adjusts the outfit space to accommodate the change.

It also adds the Scram Drive to the Remnant outfitter.

Reasoning:
1) The Remnant mostly operate in a tight set of systems where they rarely have to go more than a couple jumps in a comparatively high ramscoop environment. Therefore, they are less likely to be concerned about fuel use for short range ships.
2) They have limited manufacturing and fleet capacity compared to other races, and thus are likely to prioritize their heavy ships getting back and forth faster.
3) This is especially the case when chasing and/or running from Korath Raiders with jump drives that can move quickly between systems.

Personally, I think there is a solid case for these two ships to have a Scram Drive. While some of the same logic could apply to the Gull and Starling, it isn't as strong due to their already fast turn speed.


data/remnant ships.txt
```diff
@@ -366,7 +366,7 @@ ship "Pelican"
 		"heat dissipation" 0.7
 		"fuel capacity" 600
 		"ramscoop" 2
-		"outfit space" 470
+		"outfit space" 480
```
Amazinite: You just increased the Pelican outfit space substantially and you need to increase it again?

data/remnant ships.txt
```diff
@@ -29,7 +29,7 @@ ship "Albatross"
 		"heat dissipation" 0.6
 		"fuel capacity" 600
 		"ramscoop" 2
-		"outfit space" 658
+		"outfit space" 668
```
Amazinite: It shouldn't be an issue to install an outfits expansion where necessary instead of buffing the base stats, except for perhaps the heavy variant, which I'd say could go without a scram drive.

data/remnant ships.txt
Zitchas: Other changes don't really have much bearing on tweaking ships. It was pointed out that having scram drives as default instead of hyperdrives would fit the lore and utilization scenario for the ship. 

More to the point, does the change in space create a potential imbalance (allow for too many bunks re: capture potential, or allow it to become too good of a warship?), or does it simply make it a more interesting variation on the ship type?

data/remnant ships.txt
Zitchas: The ship already something of a hot cannon, in stock configuration it launches with over half heat and shuts down due to overheat within a couple minutes of intense use. I'd say that yes, an outfit expansion is an issue for it. Even after this change, it still has 16% less space than the similarly price Shield Beetle. (that'd be 130 tons of space)

I'm not really arguing for a major buff to the Albatross, just pointing out that it is more expensive than the SB and also license locked, and has 130 tons less space. Sure, it has built in stuff, but that's not much of an advantage compared to "game-provided means of easily acquiring dozens or hundreds of SBs for free anytime you want." Not to mention that the SB has higher engine, weapon, and cargo space, too. Given how the two stack up, I don't think that 10 tons for a scramdrive is exactly throwing the balance between the two of these off. 

That being said, adjusted the price up slightly on both ships.

PR #4583, opened 2019-10-07
Respect CXX

pkubaj:
Fixes build on GCC architectures on FreeBSD.


tehhowch:
What exactly is broken in them, currently?

pkubaj:
> What exactly is broken in them, currently?

Endless Sky needs C++11 compiler:
http://pylon.nyi.freebsd.org/data/head-powerpc64-default/p513859_s353130/logs/errors/endless-sky-0.9.10.log

PR #4584, opened 2019-10-07
Providing more variety with hails related to media from friendly civilians

Thunderforge:
This PR provides new hails for friendly civilians. Namely:

 - Locations related to songs now have venues separated from planets. So "at a concert hall in New Sydney" can now be a concert hall on various worlds.
   - E.g. "You should watch the video of "There Might Be Riots" at a concert hall on Splashdown!"
 - The "song" names are renamed "media names"
 - New hails are added talking positively and negatively about a variety of visual media, using the new "media name" phrase (i.e. it uses the same names as songs)
    - E.g. "I gotta admit, I'm hooked on "Sad Archie", definitely the most original holovid I've seen in ages!"
    - E.g. "On boring trips, I waste time watching "So Much For The Showroom". It has to be the cheesiest soap opera I've ever seen!"


data/hails.txt
```diff
@@ -317,22 +327,122 @@ phrase "friendly civilian"
...
-		"at Lionheart Arena!"
-		"at the Grange on Heartland!"
-		"at a resort on Vail!"
-		"at Canyonland Park!"
-		"in a stylin' party-hall on New China!"
+		"out there!"
+		"I've seen in ages!"
+		"this year!"
+		"yet!"
+		"since Seinfeld."
```
Amazinite: Might want to stay away from direct modern references like this.

Thunderforge: It was intended to be a joke that Seinfeld was still considered relevant in the far future, but I see how it could be distracting. I've removed it.

data/hails.txt
```diff
@@ -431,11 +540,11 @@ phrase "friendly civilian"
 	word
 		" "
 	word
-		"song"
+		"band"
 	word
 		" "
 	phrase
-		"song"
+		"media name"
```
Pointedstick: If you're going to change this to be about bands rather than songs/media names, then this needs to be `band`, not `media name`.

Also some of the phrases that come next might need a bit of adjustment to refer to bands rather than songs

Thunderforge: @Pointedstick Finally getting a chance to address this. It looks like I got mixed up with the word `"song"` and the phrase `"song"`, and wound up making something nonsensical. I think also I got mixed up because some of the suffixes to the phrase were things like `"Check 'em out!"`, which sounded to me more like a band. A couple of ways I could address this:

1. Keep the word `"song"` and add quotes around the `"media name"`
  * E.g. "That song "Sad Archie" is so great! Have you heard it yet?"
  * E.g. "The new song "Nobody Gets out of Here Alive" is my kind of music! Check 'em out!"
2. Keep the word `"band"` and change the `"media name"` to the phrase `"band"`
  * E.g. "That band There Might Be Riots is so great! Have you heard it yet?"
  * E.g. "The new band "Mediocre Generica" is my kind of music! Check 'em out!"
3. Remove the word `"song"`/`"band"` and add quotes around the media name
  * E.g. "That "Sad Archie" is so great! Have you heard it yet?"
  * E.g. "The new "Nobody Gets out of Here Alive" is my kind of music! Check 'em out!"

Is there a preference?

FishermanHMeM:
> I'm not 100% sold on changing the song names to be generic media titles, since a lot of them are quite long and silly in a way that makes at least a little bit of sense for song titles but is (IMO) too over-the-top for to be the names of TV shows (or the 31st century equivalent of course). I won't block on that, though I wouldn't mind discussing it a bit more.

"Is It Wrong to Try to Pick Up Girls in a Dungeon?" and thousands more asian names for animation and live-action films.
So, long and silly names may live. At least from my perspective.

data/hails.txt
Amazinite: I'd say go with 1, since you've renamed the "song" phrase to "media name", so line 434/543 should really just stay as "song" instead of being changed to "band".

PR #4588, opened 2019-10-10
Add a new "scale" attribute to sprites

### LINKED ISSUE #4585: Feature request: flare sprite scaling in engine outfit attributes
some time ago, the ability was added for ships engine definitions to scale flare sprites to accommodate multiple engine hardpoints with different sized flares by applying a multiplier to the hardpoint definition within a ship's attributes.

perhaps a similar technique could be applied to the attributes of engine outfits such that only the largest flare is an actual sprite, while smaller variants of the engine scale the large sprite down (which may then be scaled again for the reference feature).

specifically in reference to #4551, this feature would eliminate the need for the creation of two more engine flares (for Basrem and Baellie - that's 4 more files, ~1kb each).

the main benefit of this feature would be the elimination of roughly half of all current and future engine flare graphics assets - cutting distribution bloat, as well as launch time (especially for hdd driven storage - small files have a performance toll).

at the moment, this only amounts to ~200kb of data in some 48 files, but stands to improve over time as more engine flare styles are created.

this technique could potentially be applied to other effects as well, potentially offering more depth to explosions - mostly just adding the appearance of more sprites without actually adding those assets.  this may be a different feature request, so this is really just a footnote.

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4585.
Resolves #4585

## Feature Details
Adds a new attribute to all sprites called `scale` which scales the sprite by a given factor. Uses for such a feature are outlined in #4585; although `scale` can be used on any sprite, be it a ship, projectile, flare, or effect, its intended use is on flares and maybe effects in order to reduce the number of sprite images required for flares (and effects) of various sizes.

TODO:
- [x] Scale the collision mask alongside the sprite.

## UI Screenshots
A half-scaled Shield Beetle next to a normal scaled Shield Beetle; turret sprites and hardpoint locations scale with the ship.
image
Four Shield Beetles (one still half-scaled) using four different Hai thrusters, but the thrusters are all using the same sprite (atomic/large), just scaled to different sizes.
image
Double scaled huge explosion effects.
image
Double scaled energy blaster projectiles next to normal scaled modified blaster projectiles (fired from a normal sized Shield Beetle).
image

## Usage Examples
For the images shown above.
On a ship:
```
ship "Shield Beetle"
	sprite "ship/hai shield beetle"
		"scale" 0.5
	...
```
Used on engine flare sprites:
```
outfit `"Basrem" Atomic Thruster`
	...
	"flare sprite" "effect/atomic flare/large"
		"frame rate" 14
		"scale" 0.2

outfit `"Benga" Atomic Thruster`
	...
	"flare sprite" "effect/atomic flare/large"
		"frame rate" 14
		"scale" 0.4

outfit `"Biroo" Atomic Thruster`
	...
	"flare sprite" "effect/atomic flare/large"
		"frame rate" 13
		"scale" 0.6

outfit `"Bondir" Atomic Thruster`
	...
	"flare sprite" "effect/atomic flare/large"
		"frame rate" 12
		"scale" 0.8

# Normal scaled, no scale attribute needed
outfit `"Bufaer" Atomic Thruster`
	...
	"flare sprite" "effect/atomic flare/large"
		"frame rate" 11
```
On a weapon's projectile (and one could also use this on a hardpoint sprite in the same way):
```
outfit "Energy Blaster"
	...
	weapon
		sprite "projectile/blaster"
			"scale" 2
	...
```
On an effect:
```
effect "huge explosion"
	sprite "effect/explosion/huge"
		"no repeat"
		"frame rate" 15
		"scale" 2
	...
```


Amazinite:
I just noticed in that image with the two Shield Beetles fighting the Gunboat that the collision mask on the scaled ship is still the normal size. I'd assume the same is true with projectiles. Looks like there's work still to be done.

source/Body.h
```diff
@@ -93,6 +94,7 @@ class Body {
 	// A zoom of 1 means the sprite should be drawn at half size. For objects
 	// whose sprites should be full size, use zoom = 2.
 	float zoom = 1.f;
+	double scale = 1.;
```
tehhowch: Should be a float, like zoom.

source/Hardpoint.cpp
```diff
@@ -190,11 +190,11 @@ void Hardpoint::Fire(Ship &ship, vector<Projectile> &projectiles, vector<Visual>
 	// Get projectiles to start at the right position. They are drawn at an
 	// offset of (.5 * velocity) and that velocity includes the velocity of the
 	// ship that fired them.
-	Point start = ship.Position() + aim.Rotate(point) - .5 * ship.Velocity();
+	Point start = ship.Position() + aim.Rotate(point * ship.Scale()) - .5 * ship.Velocity();
```
tehhowch: Should we even scale hardpoint positions?

jafdy: Probably not at the moment, the current syntax suggests its a property of the sprite that it effects while it actually effects the sprite *and* other parts of the ship.

jafdy:
I have a concern that images will be blurrier when scaled by small amounts. So a 0.8 scale would not look as crisp as a 1 or a 0.5 scale image. It's less of a issue for things like projectiles and flares but for ships it could look quite bad. 
Then again, it could just be left to content creators to fix that problem if it comes up.

source/Hardpoint.cpp
Amazinite: Why wouldn't we? Scaling the ship but not the hardpoints would mean that 1) the positions given by the ship builder would be wrong and 2) in game you'd see the engine flares, weapon projectiles, and turret sprites all at the wrong spots.

petervdmeer: I agree with tehhowch and jafdy; scale does look like a property of the sprite.
The ship builder can be updated to also support scaling if users need that to calculate the correct positions.

Amazinite:
This is half working. When you first load the game, a ship with a `scale` will be scaled properly, but the collision mask won't exist.
image
Then, if you *reload* the save file, the collision mask appears. (Landing on a planet and departing does not fix it.)
image
Any save file reloads after the first one cause no issues.

Zitchas:
It occurred to me that using ship scaling might be a good way to facilitate experimentation with fighter and drone sizes. This would allow us to trial shrinking them without having to manually resize graphics or lose resolution.

Amazinite:
Good thinking.

Zitchas:
Not sure if it'd make a good story or anything, but could also make for interesting innovations. Something to the effect of "group makes drone. It's a decent drone with good stats, but gets shot down easily. mission stuff helping test and find new parts. They figure out how to streamline and condense the drone into a smaller package. Basically the same design, but now they  know what they are doing and optimize the internals a bit. One graphic, and basically two drones - the prototype big one and the production model smaller one. Happens a fair bit in real-life, too, so not unrealistic.

Zitchas:
Graey has requested that I merge this into the Combat Revamp. Is it in sufficient working order for general use/testing?

Amazinite:
Yes. Simply be aware that scaling a sprite does not scale its hardpoints. You’ll need to manually scale and hardpoint coordinates by the same scale factor that you want the sprite scales by. Same goes for hardpoint offsets if you’re using scale on a hardpoint sprite.

Amazinite:
So ships are properly scaling, but projectile sprites appear to be scaling twice. The following image shows two Barbs, one double scaled. One of the barbs is firing a normal blaster and a modified blaster, the projectiles of which are typically the same size, but with the normal blaster's projectiles double scaled. While the double scaled ship is clearly twice the height and width, the double scaled projectile is more like four times the original size. This can also be seen in the screenshots of the original post. I believe the double scaled explosion sprite is also actually more than double the dimensions.
image

Terin:
Huh, I wonder if that might be a culprit for #5711. I've been poking around in StarField but can't see anything that'd cause it in there, but hadn't thought to look in Body.

source/Mask.cpp
```diff
@@ -414,16 +414,17 @@ const vector<vector<Point>> &Mask::Outlines() const
...
 	// Keep track of the closest intersection point found.
 	double closest = 1.;
 	
 	for(auto &&outline : outlines)
 	{
-		Point prev = outline.back();
+		Point prev = outline.back() * scale;
 		for(auto &&next : outline)
 		{
+      next *= scale;
```
Amazinite, line 427:
```suggestion
			next *= scale;
```

source/Mask.cpp
```diff
@@ -466,9 +467,10 @@ bool Mask::Contains(Point point) const
 	int intersections = 0;
 	for(auto &&outline : outlines)
 	{
-		Point prev = outline.back();
+		Point prev = outline.back() * scale;
 		for(auto &&next : outline)
 		{
+      next *= scale;
```
Amazinite, line 473:
```suggestion
			next *= scale;
```

tehhowch:
This approach needs to be profiled due to the changes to collision detection. 

An alternate approach that will have different impacts (also needing profiling) is to update the Body class to manage the specific sprite masks it uses, rather than defer to the Sprite class to get the mask.  
So each time a sprite / sprite scale is set on a body, its `masks` collection is replaced with the appropriate scaled mask.
This approach reduces the opportunity for bugs with targeting / collision, since consuming classes don't ever have to know about a body's scale (just the body and draw code).
However, it does have a downside in that we then need to use more memory per Body, even though the overwhelming majority of them would use the same mask, and some bodies don't even need masks. The latter can be mitigated somewhat with refactoring (e.g. decompose `Body` into `Drawable` (has-a sprite) and a `Collidable` (is-a drawable, has-a mask)), but the former could stress older systems in a different way (due to the larger object).

A memory optimization to this other approach would be to instead use a mask manager, which maps a sprite-scale combination to a particular set of scaled masks. Then we would have only as many variations of a mask as the game actually attempts to use.  
In this managed approach I think we would again be going through the Sprite class to get the mask, e.g.
```c++
const Mask &Body::GetMask() const
{
  //...
  return sprite ? sprite->GetMask(frame, scale) : EMPTY;
}

const Mask &Sprite::GetMask(int frame, float scale) const
{
  // mask manager should know if this Sprite has collision masks, return empty if none exist.
  // ask manager for ref to set of masks for given scale. If not yet existing, manager scales the default masks and caches them.
  // return a ref to the correct frame of the manager's masks.
}
```
The tradeoff in this optimization would be complexity, and we would still want to profile since there is an additional pointer that needs to be chased when fulfilling each call to Body::GetMask.

Amazinite:
> So each time a sprite / sprite scale is set on a body, its `masks` collection is replaced with the appropriate scaled mask.

This is more or less what I originally tried to do, but as described [above](https://github.com/endless-sky/endless-sky/pull/4588#issuecomment-686750834), I had issues with it and therefore switched to the method I'm using now as it seemed more straight forward to me.

tehhowch:
The issue there was that a Sprite instance is not guaranteed to have been processed by the image-loading background threads when the definition for a ship, effect, etc is being loaded by the main thread. Deferring the scaling until the mask is requested will remove that race condition.

Amazinite:
So would this MaskManager be a statically accessed class akin to GameData, having some method `vector<Mask> &GetMasks(Sprite *sprite, double scale)` that any sprite can access via `MaskManager::GetMasks(this, scale)` within Sprite::GetMask? Or would it be some object stored somewhere that all sprites have a reference to?

Given that Sprite::GetMask is a const function, wouldn't MaskManager be unable to create new masks on the fly (irrespective of the form that MaskManager takes)? I'm imagining that perhaps MaskManager should create all the necessary masks during loading akin to how Systems calculate all their neighbors during loading; i.e. just as the game keeps track of all possible jump distances while it loads the game data then creates all necessary neighbor sets, it would also keep track of all possible sprite, scale pairs while loading the game data then create all the necessary masks before the game starts.

tehhowch:
> So would this MaskManager be a statically accessed class akin to GameData, having some method `vector<Mask> &GetMasks(Sprite *sprite, double scale)` that any sprite can access via `MaskManager::GetMasks(this, scale)` within Sprite::GetMask? Or would it be some object stored somewhere that all sprites have a reference to?

Without object factories to help configure instances, a "sensible global"--some class instance that is accessed via a static global--is an ok intermediate. The most fitting approach (if adopting lazy mask initialization) would be for the Sprite class to access a const instance provided by GameData.  

> I'm imagining that perhaps MaskManager should create all the necessary masks during loading akin to how Systems calculate all their neighbors during loading; i.e. just as the game keeps track of all possible jump distances while it loads the game data then creates all necessary neighbor sets, it would also keep track of all possible sprite, scale pairs while loading the game data then create all the necessary masks before the game starts.

If we can determine all collision masks that need to be generated before gameplay starts, in terms of both Sprite and Scale, then the manager could indeed precompute all masks and set them on the Sprite. (It's important to remember that a new scale could be parsed both before and after the Sprite's image has been converted into an outline.)
 
> Given that `Sprite::GetMask` is a const function, wouldn't MaskManager be unable to create new masks on the fly (irrespective of the form that MaskManager takes)? 

Const methods suggest to callers that the object's observable state isn't changed by the call. It's really a promise, rather than an inviolable rule. Consider `Set<T>::Get`, which offers both const and non-const overloads: both mutate the Set.

Amazinite:
That was easier than expected to get something working.

tehhowch (changes requested):
IMO looking _way_ better than the previous implementation.

Couple things to address:

source/GameData.cpp
```diff
@@ -456,6 +459,7 @@ void GameData::Preload(const Sprite *sprite)
 void GameData::FinishLoading()
 {
 	spriteQueue.Finish();
+	maskManager.ScaleMasks();
 }
```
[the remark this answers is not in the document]
Amazinite: Preload/FinishLoading also seems to be strictly used for landscape images, so really it's only a coincidence that scaling the masks here is done when it's needed for scaled collision masks to work, and there's probably some more appropriate place elsewhere that this could be done.

source/Body.cpp
```diff
@@ -213,6 +224,9 @@ void Body::LoadSprite(const DataNode &node)
 		else
 			child.PrintTrace("Skipping unrecognized attribute:");
 	}
+	
+	if(scale != 1.f)
+		GameData::GetMaskManager().AddScale(sprite, Scale());
```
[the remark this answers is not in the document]
Amazinite: Technically a body could have multiple "scale" lines and only the last line is what gets used, but if we kept track of the scales as they were added then we'd be creating masks at a number of scales that are never used.

source/Sprite.h
```diff
@@ -60,14 +58,14 @@ class Sprite {
...
-	const Mask &GetMask(int frame = 0) const;
+	const Mask &GetMask(int frame = 0, double scale = 1.) const;
 	
 	
 private:
 	std::string name;
 	
 	uint32_t texture[2] = {0, 0};
-	std::vector<Mask> masks;
+	MaskManager *maskManager = nullptr;
```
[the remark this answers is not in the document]
Amazinite: I can make this const by moving the mask clearing from Sprite::Unload to SpriteQueue::DoLoad, which is what calls Sprite::Unload, but as soon as I change this from storing a pointer to storing a reference it no longer seems to work.

If instead of having the Sprite store information about the MaskManager I change it so that the Body passes the MaskManager as a const reference when it asks for the mask (which already has access to the MaskManager via GameData for the use of adding scales) then it works, but at that point it seems like we could cut out the Sprite::GetMask method entirely and just have the Body in Body::GetMask ask the MaskManager for the relevant Mask.

source/GameData.cpp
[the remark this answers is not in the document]
Amazinite: I've moved the scaling of masks to after SpriteQueue is done loading at the start of the game (when GameData::Progress is finished). The only sprites that don't get loaded at the start of the game are the landscape images, which don't have masks, so we shouldn't need to scale any new masks after the game starts.

source/Sprite.cpp
```diff
@@ -82,23 +82,14 @@ void Sprite::AddFrames(ImageBuffer &buffer, bool is2x)
...
-
 // Free up all textures loaded for this sprite.
 void Sprite::Unload()
 {
 	glDeleteTextures(2, texture);
 	texture[0] = texture[1] = 0;
 	
-	masks.clear();
+	if(maskManager)
+		maskManager->Clear(this);
```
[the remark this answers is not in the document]
Amazinite: Looks like clearing of masks is entirely unnecessary, as Sprite::Unload only gets called on landscape sprites, which lack masks.

source/MaskManager.h
```diff
@@ -0,0 +1,49 @@
...
+	// Create the scaled versions of all masks from the 1x versions. 
+	void ScaleMasks();
+	
+	// Get the masks for the given sprite at the given scale. If a
+	// sprite has no masks, an empty mask is returned.
+	const std::vector<Mask> &GetMasks(const Sprite *sprite, double scale) const;
+	
+	
+private:
+	std::map<const Sprite *, std::map<double, std::vector<Mask>>> spriteMasks;
```
tehhowch, line 44: One possible point of failure of this approach is that `float` and `double` can be terrible map keys, especially when they are derived from computations. In our case, we don't perform any computations (beyond `static_cast<double>`) on the `Body::scale` field, so we should effectively always be searching for the original value whenever we are looking for a scaled body's collision masks.

If collision mask retrieval does end up being an issue, we could enforce an arbitrary set of allowed scaling factors based on what enables good scaling (i.e. multiples of 2). We'd then convert the users' input "scale" into a power of 2 and then use that integer power as the map key.

source/MaskManager.cpp
```diff
@@ -0,0 +1,92 @@
...
+
+using namespace std;
+
+
+
+// Move the given masks at 1x scale into the manager's storage.
+// The given vector will be cleared.
+void MaskManager::AddMasks(const Sprite *sprite, std::vector<Mask> &masks)
+{
+	spriteMasks[sprite].emplace(1., move(masks));
```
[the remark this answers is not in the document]
Amazinite: So should we look to see if a mask has already been added and erase the 1. scale entry if it has?
```suggestion
	auto &spriteMask = spriteMasks[sprite];
	if(spriteMask.find(1.) != spriteMask.end())
		spriteMask.erase(1.);
	spriteMasks[sprite].emplace(1., move(masks));
```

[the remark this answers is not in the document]
tehhowch: It'd make sense to store the iterator, so if `.find(1.)` isn't end, you can just do e.g. `it->second.swap(masks)`, but yes, that's the idea.

tehhowch (approved):

Issue #4565, opened 2019-10-02
Add preference to disable "on visit" for repeated missions


tehhowch:
**Is your feature request related to a problem? Please describe.**
Seems #4315 and #4342 may have overlooked some UX behaviors: https://steamcommunity.com/app/404410/discussions/0/1631916887483823523/

**Related Issue Links**
Haven't searched for the comment itself but I believe MZ had mentioned that many missions did not have on visit dialogs to avoid being spammy.

**Describe the solution you'd like**
We should add a preference that prevents displaying the "on visit" dialog for a mission you have completed before. As an implementation note, we can reference the mission name in the player's `Conditions` map, and look at the number done, the number active, or even write a new condition, e.g. `"<name>: visit message shown"`

**Describe alternatives you've considered**
 - We could revert these two PRs but that's overkill 😉

These would reduce the annoyance but not eliminate it:
 - We could define a "stacked visit message" and show that when more than one mission having the same identifier has a visit message to display, rather than each mission's message
 - We could implement a "job offer filter" that reduces the offer chances of missions having the same identifier and destination when a given number of that mission are already active (e.g. if you have 5+ of the same mission with the same destination, your offer chance for another one exactly like it decreases 10% for each extra active mission. So at 5 of the same, you have `regular*.9` chance of getting another, then at 6 of the same kind it becomes `regular*.9*.9`, and so on).

Amazinite:
I'd say do something akin to the `You failed X other mission(s)` from #4526, having a message that reads  something like `You have X other incomplete mission(s) at this destination.`

tehhowch:
Closing as the "stacked visit message" was implemented in #4592, which should effectively eliminate the bad UX. Reducing the offer rate for missions the player has a lot of is not a bad idea, but is fundamentally one that can be done in its own issue + PR.

PR #4592, opened 2019-10-12
Combine multiple on visit dialogs into a single dialog to avoid being spammed

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4565.

## Feature Details
Implements a similar system as #4526 for on visit dialogs by telling the player how many other incomplete missions they have at this location instead of spamming them with a new dialog for each mission. The on visit dialog that gets shown is currently simply from the first mission in the list. Unique missions (i.e. one time missions) will show their dialogs separately from non-unique (i.e. repeatable) missions. If you have multiple non-unique on visits and any of them have deadlines, then you will be reminded of that.

## UI Screenshots
image
image
image
See https://github.com/endless-sky/endless-sky/pull/4592#discussion_r334246030 for more screenshots of various cases.

### DIFF
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -327,6 +327,13 @@ int MissionAction::Payment() const
 
 
 
+string MissionAction::DialogText() const
+{
+	return dialogText;
+}
+
+
+
 // Check if this action can be completed right now. It cannot be completed
 // if it takes away money or outfits that the player does not have.
 bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &boardingShip) const
@@ -419,9 +426,13 @@ void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, co
 			subs["<ship>"] = player.Flagship()->Name();
 		string text = Format::Replace(dialogText, subs);
 		
+		// Don't push the dialog text if this is a visit action; on visit dialogs 
+		// are handled by PlayerInfo as to avoid the player being spammed by 
+		// dialogs if they have multiple missions active with the same destination
+		// (e.g. in the case of stacking bounty jobs).
 		if(isOffer)
 			ui->Push(new Dialog(text, player, destination));
-		else
+		else if(trigger != "visit")
 			ui->Push(new Dialog(text));
 	}
 	else if(isOffer && ui)
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -327,7 +327,7 @@ int MissionAction::Payment() const
 
 
 
-string MissionAction::DialogText() const
+const string &MissionAction::DialogText() const
 {
 	return dialogText;
 }
@@ -433,7 +433,7 @@ void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, co
 		// stacking bounty jobs).
 		if(isOffer)
 			ui->Push(new Dialog(text, player, destination));
-		else if(!(trigger == "visit" && !isUnique))
+		else if(isUnique || trigger != "visit")
 			ui->Push(new Dialog(text));
 	}
 	else if(isOffer && ui)
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -401,7 +401,7 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
 
 
 
-void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, const shared_ptr<Ship> &ship) const
+void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, const shared_ptr<Ship> &ship, const bool isUnique) const
 {
 	bool isOffer = (trigger == "offer");
 	if(!conversation.IsEmpty() && ui)
@@ -426,13 +426,14 @@ void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, co
 			subs["<ship>"] = player.Flagship()->Name();
 		string text = Format::Replace(dialogText, subs);
 		
-		// Don't push the dialog text if this is a visit action; on visit dialogs 
-		// are handled by PlayerInfo as to avoid the player being spammed by 
-		// dialogs if they have multiple missions active with the same destination
-		// (e.g. in the case of stacking bounty jobs).
+		// Don't push the dialog text if this is a visit action on a nonunique
+		// mission; on visit, nonunique dialogs are handled by PlayerInfo as to
+		// avoid the player being spammed by dialogs if they have multiple
+		// missions active with the same destination (e.g. in the case of
+		// stacking bounty jobs).
 		if(isOffer)
 			ui->Push(new Dialog(text, player, destination));
-		else if(trigger != "visit")
+		else if(!(trigger == "visit" && !isUnique))
 			ui->Push(new Dialog(text));
 	}
 	else if(isOffer && ui)
--- a/source/MissionAction.h
+++ b/source/MissionAction.h
@@ -52,6 +52,8 @@ public:
 	
 	int Payment() const;
 	
+	std::string DialogText() const;
+	
 	// Check if this action can be completed right now. It cannot be completed
 	// if it takes away money or outfits that the player does not have, or should
 	// take place in a system that does not match the specified LocationFilter.
--- a/source/MissionAction.h
+++ b/source/MissionAction.h
@@ -52,7 +52,7 @@ public:
 	
 	int Payment() const;
 	
-	std::string DialogText() const;
+	const std::string &DialogText() const;
 	
 	// Check if this action can be completed right now. It cannot be completed
 	// if it takes away money or outfits that the player does not have, or should
--- a/source/MissionAction.h
+++ b/source/MissionAction.h
@@ -60,7 +60,7 @@ public:
 	bool CanBeDone(const PlayerInfo &player, const std::shared_ptr<Ship> &boardingShip = nullptr) const;
 	// Perform this action. If a conversation is shown, the given destination
 	// will be highlighted in the map if you bring it up.
-	void Do(PlayerInfo &player, UI *ui = nullptr, const System *destination = nullptr, const std::shared_ptr<Ship> &ship = nullptr) const;
+	void Do(PlayerInfo &player, UI *ui = nullptr, const System *destination = nullptr, const std::shared_ptr<Ship> &ship = nullptr, const bool isUnqiue = true) const;
 	
 	// "Instantiate" this action by filling in the wildcard text for the actual
 	// destination, payment, cargo, etc.
--- a/source/MissionAction.h
+++ b/source/MissionAction.h
@@ -60,7 +60,7 @@ public:
 	bool CanBeDone(const PlayerInfo &player, const std::shared_ptr<Ship> &boardingShip = nullptr) const;
 	// Perform this action. If a conversation is shown, the given destination
 	// will be highlighted in the map if you bring it up.
-	void Do(PlayerInfo &player, UI *ui = nullptr, const System *destination = nullptr, const std::shared_ptr<Ship> &ship = nullptr, const bool isUnqiue = true) const;
+	void Do(PlayerInfo &player, UI *ui = nullptr, const System *destination = nullptr, const std::shared_ptr<Ship> &ship = nullptr, const bool isUnique = true) const;
 	
 	// "Instantiate" this action by filling in the wildcard text for the actual
 	// destination, payment, cargo, etc.
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2372,7 +2372,11 @@ void PlayerInfo::StepMissions(UI *ui)
 			// getting spammed by on visit dialogs if they are stacking jobs
 			// from the same destination.
 			if(visitText.empty())
-				visitText = Format::Replace(mission.GetAction(Mission::VISIT).DialogText(), substitutions);
+			{
+				const auto &text = mission.GetAction(Mission::VISIT).DialogText();
+				if(!text.empty())
+					visitText = Format::Replace(text, substitutions);
+			}
 			++missionVisits;
 		}
 	}
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2342,7 +2342,6 @@ void PlayerInfo::StepMissions(UI *ui)
 	
 	string visitText;
 	int missionVisits = 0;
-	int deadlineMissions = 0;
 	auto substitutions = map<string, string>{
 		{"<first>", firstName},
 		{"<last>", lastName}
@@ -2366,31 +2365,21 @@ void PlayerInfo::StepMissions(UI *ui)
 		else if(mission.Destination() == GetPlanet() && !freshlyLoaded)
 		{
 			mission.Do(Mission::VISIT, *this, ui);
-			if(mission.IsUnique())
+			if(mission.IsUnique() || !mission.IsVisible())
 				continue;
 			
 			// On visit dialogs are handled separately as to avoid a player
 			// getting spammed by on visit dialogs if they are stacking jobs
 			// from the same destination.
-			auto action = mission.GetAction(Mission::VISIT);
-			if(!action.DialogText().empty() && visitText.empty())
-				visitText = Format::Replace(action.DialogText(), substitutions);
+			if(visitText.empty()
+				visitText = Format::Replace(mission.GetAction(Mission::VISIT).DialogText(), substitutions);
 			++missionVisits;
-			if(mission.Deadline())
-				++deadlineMissions;
 		}
 	}
-	if(!visitText.empty())
+	if(!visitText.empty() && missionVisits > 1)
 	{
-		visitText += "\n\tYou have " + Format::Number(missionVisits) + " unfinished " 
-			+ ((missionVisits > 1) ? "missions" : "mission") + " at this location";
-		if(deadlineMissions)
-		{
-			visitText += "(" + Format::Number(deadlineMissions) + " of which "
-					+ ((deadlineMissions > 1) ? "have approaching deadlines" : "has an approaching deadline)");
-		}
-		visitText += ".";
-		
+		visitText += "\n\t(You have " + Format::Number(missionVisits - 1) + " other unfinished " 
+			+ ((missionVisits > 2) ? "missions" : "mission") + " at this location.)";
 		ui->Push(new Dialog(visitText));
 	}
 	// One mission's actions may influence another mission, so loop through one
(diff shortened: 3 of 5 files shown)


source/PlayerInfo.cpp
```diff
@@ -2354,7 +2356,35 @@ void PlayerInfo::StepMissions(UI *ui)
...
+				}
+				else
+					++visitDialogs;
+			}
+		}
+	}
+	if(!visitText.empty())
+	{
+		if(visitDialogs)
+			visitText += "\n\tYou have " + Format::Number(visitDialogs) + " other incomplete " + ((visitDialogs > 1) ? "missions" : "mission") + " at this location.";
```
petervdmeer: I'm not sure about the word `incomplete`. Maybe a word like `unfinished` would fit better?

Similar for `mission/missions`, maybe use the word `job/jobs` there?

source/MissionAction.cpp
```diff
@@ -419,9 +426,13 @@ void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, co
...
 		string text = Format::Replace(dialogText, subs);
 		
+		// Don't push the dialog text if this is a visit action; on visit dialogs 
+		// are handled by PlayerInfo as to avoid the player being spammed by 
+		// dialogs if they have multiple missions active with the same destination
+		// (e.g. in the case of stacking bounty jobs).
 		if(isOffer)
 			ui->Push(new Dialog(text, player, destination));
-		else
+		else if(trigger != "visit")
```
petervdmeer: I think this feature is great for avoidance of spamming from minor jobs. But if a major storyline job (or a job with a deadline) has an "on visit", then I would like to see that one regardless of how many other jobs are relevant for the current location.

source/PlayerInfo.cpp
```diff
@@ -2340,6 +2340,8 @@ void PlayerInfo::StepMissions(UI *ui)
 				if(ship->IsDestroyed())
 					mission.Do(ShipEvent(nullptr, ship, ShipEvent::DESTROY), *this, ui);
 	
+	string visitText;
+	int visitDialogs = 0;
```
petervdmeer: The first visitDialog is not counted in the code below. Maybe call the variable `extraVisitDialogs` or `otherVisitDialogs`.

source/PlayerInfo.cpp
Amazinite: Not too keen on distinguishing jobs from other missions. Jobs and spaceport missions both go under the active missions tab, so it isn't out of the ordinary to refer to jobs as missions. Plus, I get the sense that using the term `job(s)`, there'd also need to be a case for if you had a spaceport mission *and* a job on visit.

source/MissionAction.cpp
Amazinite: How about only condensing the on visit dialogs from missions that return false for `IsUnique()`, otherwise have it shown here? That would distinguish between jobs (plus repeatable spaceport missions) and proper missions. Then in terms of deadlines, since the player might have multiple deadline based jobs, how about just adding another line that says something along the lines of "X of your unfinished missions here are time-sensitive/have an approaching deadline," instead of just showing all missions with deadlines?

tehhowch: I prefer my suggestion of stacking messages with the same mission `Identifier` (e.g. all `"Bounty Hunting (Small)"`), since it makes sure the player still gets reminded of the various types of conditions they haven't satisfied, even though it still results in showing a handful of visit dialogs. At minimum, I agree that non-repeated missions should not be treated as though they are spammy since they are, by definition, not.

Another recommendation from #4565 would be to add a mechanism that prevents displaying a mission's "on visit" dialog more than once per session (that could be done by adding a non-serialized boolean to the MissionAction class).

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2356,35 @@ void PlayerInfo::StepMissions(UI *ui)
...
+			if(ait != actions.end())
+			{
+				if(visitText.empty())
+				{
+					map<string, string> subs;
+					subs["<first>"] = firstName;
+					subs["<last>"] = lastName;
+					if(Flagship())
+						subs["<ship>"] = Flagship()->Name();
+					visitText = Format::Replace(ait->second.DialogText(), subs);
```
tehhowch: Since `Format::Replace` takes the string by const ref, `DialogText` can return const ref and avoid a string copy.

source/MissionAction.cpp
Amazinite: image

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2356,35 @@ void PlayerInfo::StepMissions(UI *ui)
 		else if(mission.CanComplete(*this))
 			RemoveMission(Mission::COMPLETE, mission, ui);
 		else if(mission.Destination() == GetPlanet() && !freshlyLoaded)
+		{
 			mission.Do(Mission::VISIT, *this, ui);
+			
+			// On visit dialogs are handled separately as to avoid a player
+			// getting spammed by on visit dialogs if they are stacking jobs
+			// from the same destination.
+			map<Mission::Trigger, MissionAction> actions = mission.GetActions();
```
tehhowch: I'd use `auto` here.

Perhaps `GetActions` should be `GetAction` and take an argument for the action type to retrieve. It would need to return a pointer to the MissionAction, which you would null-check in this method.

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2356,35 @@ void PlayerInfo::StepMissions(UI *ui)
...
+			auto ait = actions.find(Mission::VISIT);
+			if(ait != actions.end())
+			{
+				if(visitText.empty())
+				{
+					map<string, string> subs;
+					subs["<first>"] = firstName;
```
tehhowch: Since we know we want to add these, might as well directly initialize the `subs` map with them rather than do it on the next two lines. Also, rather than reinitialize the same substitution map for each mission, we should probably do it beforehand.

source/MissionAction.cpp
```diff
@@ -419,9 +426,14 @@ void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, co
...
 		
+		// Don't push the dialog text if this is a visit action on a nonunique
+		// mission; on visit, nonunique dialogs are handled by PlayerInfo as to
+		// avoid the player being spammed by dialogs if they have multiple
+		// missions active with the same destination (e.g. in the case of
+		// stacking bounty jobs).
 		if(isOffer)
 			ui->Push(new Dialog(text, player, destination));
-		else
+		else if(!(trigger == "visit" && !isUnique))
```
tehhowch: Is `else if(isUnique || trigger != "visit")` a clearer way to state the condition?

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2357,46 @@ void PlayerInfo::StepMissions(UI *ui)
...
 			mission.Do(Mission::VISIT, *this, ui);
+			
+			// On visit dialogs are handled separately as to avoid a player
+			// getting spammed by on visit dialogs if they are stacking jobs
+			// from the same destination.
+			map<Mission::Trigger, MissionAction> actions = mission.GetActions();
+			auto ait = actions.find(Mission::VISIT);
+			if(ait != actions.end())
+			{
+				if(!mission.IsUnique())
```
tehhowch: I'd move this to be an exit condition after performing the visit action, e.g.
```c++
mission.Do(Mission::VISIT, *this, ui);
if(mission.IsUnique())
  continue;

// Repeatable missions' "on visit" dialogs are handled separately....
```

source/MissionAction.cpp
Amazinite: Idk, would it be to you? I was thinking of simplifying the Boolean algebras here but wondered if this way was easier to read.

petervdmeer: The double negative makes the statement relatively hard to read; Tehhowchs version is easier to read for somebody seeing the code for the first time.

Amazinite:
>I hope we can also do something about the amount of spamming from the "on complete" minor jobs.

That should be easy to do after this, but I'll leave that to a separate PR.

>I prefer my suggestion of stacking messages with the same mission `Identifier` (e.g. all `"Bounty Hunting (Small)"`), since it makes sure the player still gets reminded of the various types of conditions they haven't satisfied, even though it still results in showing a handful of visit dialogs.

I feel that the fact that you have unfinished missions is more important than telling the player every single type of unfinished mission they have (especially because the objective of various jobs of different identifiers can be the same); if a player sees that they have X unfinished missions at *this* location, I'd expect them to just check their active missions for everything at the destination that they're at to find out what they're missing.

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2363,37 @@ void PlayerInfo::StepMissions(UI *ui)
...
 		else if(mission.Destination() == GetPlanet() && !freshlyLoaded)
+		{
 			mission.Do(Mission::VISIT, *this, ui);
+			if(mission.IsUnique())
+				continue;
+			
+			// On visit dialogs are handled separately as to avoid a player
+			// getting spammed by on visit dialogs if they are stacking jobs
+			// from the same destination.
+			auto ait = mission.GetAction(Mission::VISIT);
```
tehhowch: This variable needs to be renamed as it is not an iterator anymore.

source/MissionAction.h
```diff
@@ -52,13 +52,15 @@ class MissionAction {
...
+	const std::string &DialogText() const;
+	
 	// Check if this action can be completed right now. It cannot be completed
 	// if it takes away money or outfits that the player does not have, or should
 	// take place in a system that does not match the specified LocationFilter.
 	bool CanBeDone(const PlayerInfo &player, const std::shared_ptr<Ship> &boardingShip = nullptr) const;
 	// Perform this action. If a conversation is shown, the given destination
 	// will be highlighted in the map if you bring it up.
-	void Do(PlayerInfo &player, UI *ui = nullptr, const System *destination = nullptr, const std::shared_ptr<Ship> &ship = nullptr) const;
+	void Do(PlayerInfo &player, UI *ui = nullptr, const System *destination = nullptr, const std::shared_ptr<Ship> &ship = nullptr, const bool isUnqiue = true) const;
```
tehhowch: Typo

source/PlayerInfo.cpp
```diff
@@ -2340,6 +2340,15 @@ void PlayerInfo::StepMissions(UI *ui)
 				if(ship->IsDestroyed())
 					mission.Do(ShipEvent(nullptr, ship, ShipEvent::DESTROY), *this, ui);
 	
+	string visitText;
+	int extraVisitDialogs = 0;
+	int deadlineMissions = 0;
+	map<string, string> subs;
```
tehhowch: Initialize this map with the name substitutions

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2363,37 @@ void PlayerInfo::StepMissions(UI *ui)
...
+		}
+	}
+	if(!visitText.empty())
+	{
+		if(extraVisitDialogs)
+		{
+			visitText += "\n\tYou have " + Format::Number(extraVisitDialogs) + " other unfinished " + ((extraVisitDialogs > 1) ? "missions" : "mission") + " at this location.";
+			if(deadlineMissions)
+			{
+				visitText += "\n\t" + Format::Number(deadlineMissions) + " of your unfinished missions here " + ((deadlineMissions > 1) ? "have approaching deadlines." : "has an approaching deadline.");
```
tehhowch: Is there a better construct that doesn't start a sentence with a digit?

Amazinite: I would like one.
"Of your unfinished missions here, X have approaching deadlines."?

tehhowch: I'd be open to having the deadline missions be just an parenthesized addendum, e.g.

"You have 8 other unfinished missions here (4 of which have deadlines)."

source/PlayerInfo.cpp
tehhowch: sorry, i meant initialize it with the `<first>` and `<last>` substitutions, not to rename it 😅 
e.g.
```c++
auto subs = map<string, string>{
  {"<first>", firstName},
  {"<last>", lastName}
};
```

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2364,40 @@ void PlayerInfo::StepMissions(UI *ui)
...
+				continue;
+			
+			// On visit dialogs are handled separately as to avoid a player
+			// getting spammed by on visit dialogs if they are stacking jobs
+			// from the same destination.
+			auto action = mission.GetAction(Mission::VISIT);
+			if(!action.DialogText().empty())
+			{
+				if(visitText.empty())
+					visitText = Format::Replace(ait.DialogText(), substitutions);
```
tehhowch: compile error

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2364,40 @@ void PlayerInfo::StepMissions(UI *ui)
...
+		if(extraVisitDialogs)
+		{
+			visitText += "\n\tYou have " + Format::Number(extraVisitDialogs) + " other unfinished " 
+				+ ((extraVisitDialogs > 1) ? "missions" : "mission") + " at this location";
+			if(deadlineMissions)
+			{
+				visitText += "(" + Format::Number(deadlineMissions) + " of which "
+					+ ((deadlineMissions > 1) ? "have approaching deadlines" : "has an approaching deadline)");
+			}
+			visitTest += ".";
```
tehhowch: `visitText`

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2364,40 @@ void PlayerInfo::StepMissions(UI *ui)
...
+				else
+					++extraVisitDialogs;
+				if(mission.Deadline())
+					++deadlineMissions;
+			}
+		}
+	}
+	if(!visitText.empty())
+	{
+		if(extraVisitDialogs)
```
tehhowch: This PR currently only counts "other unfinished missions" as those that attempted to display a visit dialog. However, it is not a strict requirement that a mission must have a visit dialog, so it is possible for the player to have a great number of pips for a given planet, and yet be told "you have only 1 other unfinished mission". Thoughts?

Amazinite: If none of the unfinished missions have visit dialogs, would something still pop up? Or would it only be when there is a visit dialog that it also tells you however many other missions you have unfinished?

tehhowch: I think, if we are telling the player a number of unfinished missions for their port, that the number needs to be accurate, regardless of whether the missions have a visit message. We could change the phrasing too, and mention something like "you received N other, similar messages"

Amazinite: So make it a default feature that you'll be told about unfinished missions?

tehhowch: Or change the message text. I don't want the pop-up to say something like "you have 1 other unfinished mission here" when the player can see on the map that they have at least 12 unfinished missions there.

tehhowch: Or change the message text. I don't want the pop-up to say something like "you have 1 other unfinished mission here" when the player can see on the map that they have at least 12 unfinished missions there.

Amazinite: I've changed it so that all unfinished missions are counted, but it only tells you that if any one of those unfinished missions has an on visit dialog.

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2364,34 @@ void PlayerInfo::StepMissions(UI *ui)
...
 		else if(mission.Destination() == GetPlanet() && !freshlyLoaded)
+		{
 			mission.Do(Mission::VISIT, *this, ui);
+			if(mission.IsUnique())
+				continue;
+			
+			// On visit dialogs are handled separately as to avoid a player
+			// getting spammed by on visit dialogs if they are stacking jobs
+			// from the same destination.
+			auto action = mission.GetAction(Mission::VISIT);
```
tehhowch: Use `&` to avoid a copy. Probably we can avoid even needing to do this call at this block scope, since we only need the action's text if `visitText` is currently empty. We also don't need anything else about the action other than its text.

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2364,34 @@ void PlayerInfo::StepMissions(UI *ui)
...
+			if(!action.DialogText().empty() && visitText.empty())
+				visitText = Format::Replace(action.DialogText(), substitutions);
+			++missionVisits;
+			if(mission.Deadline())
+				++deadlineMissions;
+		}
+	}
+	if(!visitText.empty())
+	{
+		visitText += "\n\tYou have " + Format::Number(missionVisits) + " unfinished "
```
tehhowch: If we have only 1 repeating mission with a deadline and a visit dialog, (and no others at this location) we see something like:
```
<that mission's visit text>

You have 1 unfinished mission at this location (1 of which has an approaching deadline).
```

Should we even show the counts for a single mission? I feel like it should be an "other" snippet, e.g. `(You have N other unfinished missions at this location.)` and we shouldn't include any information about the job whose text we showed, even if it has a deadline. Probably that information, if relevant for the player, should have already been included in the mission's visit text.

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2364,34 @@ void PlayerInfo::StepMissions(UI *ui)
...
+			if(mission.IsUnique())
+				continue;
+			
+			// On visit dialogs are handled separately as to avoid a player
+			// getting spammed by on visit dialogs if they are stacking jobs
+			// from the same destination.
+			auto action = mission.GetAction(Mission::VISIT);
+			if(!action.DialogText().empty() && visitText.empty())
+				visitText = Format::Replace(action.DialogText(), substitutions);
+			++missionVisits;
```
tehhowch: I think we should avoid telling the player about any invisible missions that involve the current location.

source/PlayerInfo.cpp
Amazinite: Yeah, I caught the fact that this shows up regardless, but hadn't implemented anything to avoid showing this extra bit if you only have one on visit.

source/PlayerInfo.cpp
```diff
@@ -2354,7 +2363,24 @@ void PlayerInfo::StepMissions(UI *ui)
...
+			
+			// On visit dialogs are handled separately as to avoid a player
+			// getting spammed by on visit dialogs if they are stacking jobs
+			// from the same destination.
+			if(visitText.empty())
+				visitText = Format::Replace(mission.GetAction(Mission::VISIT).DialogText(), substitutions);
+			++missionVisits;
+		}
+	}
+	if(!visitText.empty() && missionVisits > 1)
```
tehhowch: Won't this mean a single visit dialog never gets shown? I think the guard needs to be around adding to the text.

Amazinite: You're right.

Issue #4591, opened 2019-10-11
Operating costs for active automated ships


petervdmeer:
**Is your feature request related to a problem? Please describe.**
Ships with crew have crew costs every day that the ship is active.
Automated ships currently have no such daily costs for being active.

Capturing automated ships is near the end of the current game, so this is not much of an issue at this moment. But if we want to add more storyline, then we might want to create some form of daily operating costs for automated ships as well.

**Related Issue Links**
Cloaking is another issue which needs to be addressed when we want more storyline in the main game (and also for more storyline from plugins), but that is handled in other issues like:
- https://github.com/endless-sky/endless-sky/issues/3825
- https://github.com/endless-sky/endless-sky/issues/4119
- https://github.com/endless-sky/endless-sky/issues/4121

There is also an issue called "A fix to the problem of (too much) money" (https://github.com/endless-sky/endless-sky/issues/486) which sounds kinda relevant.

**Describe the solution you'd like**
Some ideas that I have:
- Require "remote operators" for automated ships.
  - This can be very basic, like some persons somewhere on some unspecified planet.
  - This can also be explicit, like command centers (or automation support) on active non-automated ships.
    - Possibly with the requirement that the command center is within 1 jump of the automated ship.
- Require "local operators" for automated ships (basically not allowing the player to have ships fully automated).
  - Downside of this option is that automated ships owned by the player will be weaker than their full automatic counterparts, since the player will need to add commandcenters and bunks.

**Describe alternatives you've considered**
I didn't really consider other alternatives yet, but I hope that this feature request triggers a discussion which results in more alternatives (and better options than that I considered).

Amazinite:
Seems like something that could be solved by having #4284, allowing ships/outfits to cost you without actually needing crew.

petervdmeer:
> Seems like something that could be solved by having #4284, allowing ships/outfits to cost you without actually needing crew.

That is already a better solution than my proposals above.

Amazinite:
#4596 has added the functionality of operating costs, but I don't think that it should be used on the automaton ships, or at least not right now. Capturing automata is basically the end game right now. Maybe after more alien factions are added and there's more choice around T2 outside of buying Wanderer ships or capturing automata, then it might make sense to give operating costs to these ships as a balancing factor. The whole point of the automata is that they don't cost you anything daily, though, so if ever operating costs are added, the cost should be less than the crew salaries of ships of similar size.

petervdmeer:
[...]

Issue #4284, opened 2019-05-20
Feature Request: Maintenance Cost


VonZorch:
Would it be too difficult to add an optional maintenance cost to outfits?
This would be a daily cost to have an outfit installed.
For instance an outfit I have already done.
outfit "Palatial Captains Cabin"
	category "Systems"
	cost 5000000
	thumbnail "outfit/PalatialCabin"
	"mass" 10
	"bunks" 5
	"required crew" 4
	"outfit space" -20
	"cargo space" -80
	"art capacity" 40
	"unplunderable" 1
	description "The Captain of a larger ship needs a place for privacy and doing administrative tasks. The cabin provides this space."
	description "	This version provides a private galley with a five star chef and a steward with two assistants, to attend to the Captains needs."
	description "	The furnishings put a Star Queen to shame. Brightwork is solid gold, platinum or some other exotic, expensive, and shiny metal. The furniture can be moved to create a dance floor and is of the highest quality. Fine art adds to the decorations."

The cabin staff would not actually be involved in boarding actions or needed to operate the ship.
so
 	"bunks" 5
	"required crew" 4
would become
	"bunks" 1
	"maintenance cost" 1000
the break down being 400 for the chef, 150 for the steward and 100 each for the assistants with another 250 for high quality foods and supplies.

Zitchas:
Hmmm. That's an interesting idea. You could probably get something similar by using negative salary, but I don't think that can be assigned to an outfit.

As sort of corollary to this, it would be interesting to be able to assign both negative and positive expense values. It could be used as you describe for maintenance or non-boarding-crew-salaries, or it could be used with a positive value for, say, GICs (Guaranteed Investment Certificates) (basic financial instruments that one buys for a certain amount, say, 1000, then gets a specific interest for a specific length of time, at which point the purchaser gets their original money back plus the interest. Extremely low risk in financial terms, so has an extremely low interest return.

Pointedstick:
This has been repeatedly vetoed, but I have to admit I'd like it too. We can basically implement this by giving an outfit crew requirements, so I don't see any reason why you shouldn't be able to do it without requiring extra crew.

For that matter, I still think you should have to pay a small amount to refuel, repair, and recharge your ship in port...

finite-galaxy:
> pay a small amount to refuel

This is implemented in my game over three months ago, feel free to have a look at https://github.com/finite-galaxy/finite-galaxy/pull/15 by @IntegratedQuantum.

[...]

Issue #3659, opened 2018-03-05
[suggestion] Add maintenance requirement to outfits


starcatter:
Purpose: add the option to balance different outfits and ship builds against each other through varying maintenance costs.

Mechanics:
- Basic ship crew requirement should be a minimal viable crew (pilot,navigator,loadmaster,captain) depending on ship size/type
- Outfits get a maintenance requirement, expressed in man-hours/day
- Maintenance requirement will generally scale with outfit efficiency and size
- Crew members can commit a number of man-hours of maintenance per day, depending on their role
- By default all ships will be fully crewed, with a possible surplus maintenance capacity (does not carry over to other ships)
- Should the crew become insufficient to maintain all ship outfits, the ship's effective stats will deteriorate down to a level reflecting the currently met maintenance needs

Optional - alien tech:
- Outfits originating from different alien species will require different maintenance types
- Players can hire engineers specializing in alien tech, provided the aliens trust them
- All engineers can maintain all outfit types but maintaining foreign tech results in a penalty

Optional - self-maintaining ships:
- Certain outfits can be installed which provide maintenance for ship systems
- For balance sake, all such systems come at a cost, either energy use, crew capacity or large amounts of outfit space

A few examples of how I imagine this might work out:
1 - The player gets a small fighter which only requires the pilot onboard. Outfits require a total of 1.4h of maintenance per day, while the pilot can spare 1.5h. Replacing the radiothermal reactor with a lighter and more powerful pebble core would require an engineer to keep the ship flying in perfect condition. An alternative might be using a Hai or Remnant radiothermal generator, along with a pilot hired at one of their planets.

2 - The player boards their first Korath Raider. Having stolen the outfits, they learn that running the engines and reactor would require adding extra bunks to their ship, unless they can find engineers who can work with Korath outfits.

3 - Korath drone ships may depend on heavy and energy-hungry systems to keep them maintenance free. Removing them allows the player to fit more useful outfits, but requires them to carry a small crew of engineers.

4 - Self-maintenance systems may include Korath systems-cores, Coalition repair modules and some new outfits that don't provide hull repair. Another possibility would be installing maintenance robot stations, either "standalone" costing much outfit space or "bunk-conversion" which take up some outfit and bunk space. These might provide an alternative to hiring engineers who can work with specific tech.

[...]

Issue #2219, opened 2017-03-04
Propositions: Changes in crew salaries and maintenance costs


Tadrix:
**Salaries:** Consider that each ship which isn't piloted by player needs its captain. Now take into account that said captain needs his own Pilot's License (somehow I can't imagine unparked or left behind ships successfully convincing, for example, Navy authorities: "But, but… our employer has a Pilot's License. It's the same for us, right? Oh, (s)he's several systems away by now." Consider also that this license costs 50K. 50000 for ship captain's license, while player pays them 100/d? If this license would need to be refreshed annually, it wouldn't be worth it (a crew member earns 36500/y).
**Suggestion: One crew member per player's escort ship (excluding flagship) earns 200/d.**
**More elaborate version:** interceptor's (or transport's) captain earns 600/d, light warship's, light freighter's and Tarazed Blackbird's captain earns 700/d, medium warship's captain earns 800/d, heavy warship's, heavy freighter's and Star Queen's captain earns 1000/d. If we want to favor traders (because plundering pays better) then we can set (any) freighter captain's salary to 300/d, since his job is to just get cargo from A to B.
The point is, personal Pilot's Licenses should pay themselves off in year's time for pilots employed by the player, otherwise they wouldn't be motivated to obtain said license and join his/her crew (because only way to actually pay off that license would be being an independent merchant).

**Maintenance:** Yes, I'm aware that this has been brought up before, and that M.Z. had unfavorable opinion of it in the past (but absence of which, I may add, made androids so unbalanced they needed to be removed from the game). This is the same way Kor Sestor and Kor Mereti warships are unbalanced. With any luck, capture them and — yay! — no more upkeep.
**Suggestion: Maintenance costs 0,025% ship's total full (undepreciated) value per day.** Not especially hurting new players (should be about 100/d), but will scale nicely with big fleets with fancy outfits (e.g. using Quarg stuff).
**More elaborate version:** Give the player choice to not pay for maintenance. Downside? Greatly accelerated 
depreciation. And if already at bottom, then affected ship starts to malfunction (like it's understaffed), its speed gradually drops, weapons inaccuracy gradually increases, and so on, until repairs are maid (meaning, pay what's unpaid and then some more for neglect). Of course, pirates maintain their ships poorly, so depreciation could even be set to longer time period without hurting balance IMO.

MessyMix:
[...]

PR #4596, opened 2019-10-14
Added daily maintenance attributes

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4284. 
Maintenance costs have also been requested or discussed in #4591, #3659, #2219, and more.

## Feature Details
This mechanic is not intended to be used in vanilla, but seeing as how it has been requested many times, I see this as a worthwhile addition. 

Adds `"maintenance costs"` and `"operating costs"` outfit attributes that work similarly to how crew salaries work. Every day, the total maintenance cost is deducted from the player's credits. `"maintenance costs"` are always active so long as the player owns the outfit/ship while `"operating costs"` only apply to unparked ships. Both will show as `Maintenance` in the bank and with each day that passes.

Overdue maintenance costs have zero effect on the efficiency or sale price of the outfit or ship they are attached to, and having maintenance costs on an outfit or ship in no way effects how it depreciates.

If the player has overdue maintenance costs and sells the outfit or ship that has the maintenance cost, then they will still need to pay the overdue costs, just as they do with overdue crew salaries after firing the crew. This is to make it so that the player can't simply ignore overdue maintenance costs by selling the offending outfit or ship. If the purpose of maintenance costs is like the usage outlined in #4284 where it's a roundabout way to have crew that cost extra, then this makes sense, since you're technically paying overdue crew salaries. But in the case where maintenance costs are actually meant as maintenance costs, one could explain this away as the player paying for the repair of the outfit/ship even after they sold it off because of how overdue maintenance costs don't impact sale price.

This PR also fixes an oversight with crew salaries; overdue crew salaries will continue to be paid even if you fire all of your crew, but if you have 0 crew aside from yourself then the crew line will not appear in the bank panel, meaning you can't tell how much overdue crew salaries you still have to pay, but you could still click on the `Pay All` button, even though it was greyed out, and pay off the overdue salaries. This fixed behavior extends to overdue maintenance costs without any current maintenance costs.

## UI Screenshots
Crew salaries will now display in the bank if you have no crew but have overdue salaries, and the `Pay All` button highlights accordingly.
image
A display of two types of payments being paid at once.
image
Three types of payments being paid.
image
All four payment types being paid. Payment types are displaying in alphabetical order because I guess that's how a `map` stores keys.
image
The bank panel will size the number of mortgages properly depending on if you have crew or maintenance costs.
image
image

## Usage Examples
`"maintenance costs"` and `"operating costs"` are attributes of an outfit or ship, used as follows.
```
outfit "Jump Drive"
	category "Systems"
	cost 1000000
	thumbnail "outfit/jump drive"
	"mass" 20
	"outfit space" -20
	"jump speed" .3
	"jump fuel" 200
	"jump drive" 1
	"maintenance costs" 1000

ship "Kar Ik Vot 349"
	sprite "ship/kar ik vot 349"
	thumbnail "thumbnail/kar ik vot 349"
	attributes
		category "Heavy Warship"
		"cost" 41280000
		"operating costs" 10000
		...
```

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -155,8 +155,8 @@ tip "jump speed:"
 tip "jump fuel:"
 	`How much fuel is consumed when you jump between systems using this drive.`
 
-tip "maintenance cost:"
-	`How many credits per day you need to spend to maintain this outfit.`
+tip "maintenance costs:"
+	`How many credits per day you need to spend to maintain this outfit. Even if a ship is parked, you still need to pay for its maintenance.`
 
 tip "map:"
 	`How many star systems are included in this map.`
@@ -166,6 +166,9 @@ tip "mass with no cargo:"
 
 tip "mass:"
 	`Mass, in tons. A ship's mass determines how fast it turns and accelerates.`
+
+tip "operating costs:"
+	`How many credits per day you need to spend to maintain this outfit while it is in active use.`
 
 tip "outfit scan:"
 	`Lets you scan your target ship's outfits from up to this distance away.`
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -156,7 +156,7 @@ tip "jump fuel:"
 	`How much fuel is consumed when you jump between systems using this drive.`
 
 tip "maintenance costs:"
-	`How many credits per day you need to spend to maintain this outfit. Even if a ship is parked, you still need to pay for its maintenance.`
+	`How many credits per day you need to spend to maintain this outfit. Even if a ship is parked or an outfit is in cargo, you still need to pay for its maintenance.`
 
 tip "map:"
 	`How many star systems are included in this map.`
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -155,6 +155,9 @@ tip "jump speed:"
 tip "jump fuel:"
 	`How much fuel is consumed when you jump between systems using this drive.`
 
+tip "maintenance cost:"
+	`How many credits per day you need to spend to maintain this outfit.`
+
 tip "map:"
 	`How many star systems are included in this map.`
 
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -162,6 +162,28 @@ string Account::Step(int64_t assets, int64_t salaries, int64_t maintenance)
 		}
 	}
 	
+	// Maintenance costs are dealt with after crew salaries given that they act similarly.
+	int64_t maintenancePaid = maintenanceDue;
+	if(maintenanceDue)
+	{
+		if(maintenanceDue > credits)
+		{
+			// Like with crew salaries, maintenance costs can be paid in part with
+			// the unpaid costs being paid later.
+			maintenancePaid = max<int64_t>(credits, 0);
+			maintenanceDue -= maintenancePaid;
+			credits -= maintenancePaid;
+			if(!missedPayment)
+				out << "You could not pay all your maintenance costs.";
+			missedPayment = true;
+		}
+		else
+		{
+			credits -= maintenanceDue;
+			maintenanceDue = 0;
+		}
+	}
+	
 	// Unlike salaries, each mortgage payment must either be made in its entirety,
 	// or skipped completely (accruing interest and reducing your credit score).
 	int64_t mortgagesPaid = 0;
@@ -197,26 +219,6 @@ string Account::Step(int64_t assets, int64_t salaries, int64_t maintenance)
 			++it;
 	}
 	
-	// Maintenance costs are dealt with last.
-	int64_t maintenancePaid = maintenanceDue;
-	if(maintenanceDue)
-	{
-		if(maintenanceDue > credits)
-		{
-			maintenancePaid = max<int64_t>(credits, 0);
-			maintenanceDue -= maintenancePaid;
-			credits -= maintenancePaid;
-			if(!missedPayment)
-				out << "You could not pay all your maintenance costs.";
-			missedPayment = true;
-		}
-		else
-		{
-			credits -= maintenanceDue;
-			maintenanceDue = 0;
-		}
-	}
-	
 	// Keep track of your net worth over the last HISTORY days.
 	if(history.size() > HISTORY)
 		history.erase(history.begin());
@@ -228,7 +230,7 @@ string Account::Step(int64_t assets, int64_t salaries, int64_t maintenance)
 	creditScore = max(200, min(800, creditScore + (missedPayment ? -5 : 1)));
 	
 	// If you didn't make any payments, no need to continue further.
-	if(!(salariesPaid + mortgagesPaid + maintenancePaid + finesPaid))
+	if(!(salariesPaid + maintenancePaid + mortgagesPaid + finesPaid))
 		return out.str();
 	else if(missedPayment)
 		out << " ";
@@ -276,7 +278,6 @@ string Account::Step(int64_t assets, int64_t salaries, int64_t maintenance)
 		if(finesPaid)
 			out << creditString(finesPaid) << " in fines.";
 	}
-	
 	return out.str();
 }
 
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -254,15 +254,13 @@ string Account::Step(int64_t assets, int64_t salaries, int64_t maintenance)
 	// include commas, so just handle that separately here.
 	if(typesPaid.size() >= 3)
 	{
-		string output;
 		auto it = typesPaid.begin();
 		for(unsigned int i = 0; i < typesPaid.size() - 1; ++i)
 		{
-			output += creditString(it->second) + " in " + it->first + ", ";
+			out << creditString(it->second) << " in " << it->first << ", ";
 			++it;
 		}
-		output += "and " + creditString(it->second) + " in " + it->first + ".";
-		out << output;
+		out << "and " << creditString(it->second) << " in " << it->first + ".";
 	}
 	else
 	{
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -31,7 +31,7 @@ namespace {
 
 // Default constructor.
 Account::Account()
-	: credits(0), salariesOwed(0), creditScore(400)
+	: credits(0), salariesOwed(0), maintenanceDue(0), creditScore(400)
 {
 }
 
@@ -42,6 +42,7 @@ void Account::Load(const DataNode &node)
 {
 	credits = 0;
 	salariesOwed = 0;
+	maintenanceDue = 0;
 	creditScore = 400;
 	history.clear();
 	mortgages.clear();
@@ -52,6 +53,8 @@ void Account::Load(const DataNode &node)
 			credits = child.Value(1);
 		else if(child.Token(0) == "salaries" && child.Size() >= 2)
 			salariesOwed = child.Value(1);
+		else if(child.Token(0) == "maintenance" && child.Size() >= 2)
+			maintenanceDue = child.Value(1);
 		else if(child.Token(0) == "score" && child.Size() >= 2)
 			creditScore = child.Value(1);
 		else if(child.Token(0) == "mortgage")
@@ -73,6 +76,8 @@ void Account::Save(DataWriter &out) const
 		out.Write("credits", credits);
 		if(salariesOwed)
 			out.Write("salaries", salariesOwed);
+		if(maintenanceDue)
+			out.Write("maintenance", maintenanceDue);
 		out.Write("score", creditScore);
 		
 		out.Write("history");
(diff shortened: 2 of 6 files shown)


source/PlayerInfo.cpp
```diff
@@ -687,6 +687,19 @@ int64_t PlayerInfo::Salaries() const
...
 
 
+
+// Calculate the daily maintenance cost for all unparked ships.
+int64_t PlayerInfo::Maintenance() const
+{
+	int64_t maintenance = 0;
+	for(const shared_ptr<Ship> &ship : ships)
+		if(!ship->IsParked() && !ship->IsDestroyed())
+			maintenance += ship->Attributes().Get("maintenance cost");
```
petervdmeer: Does this work properly if I have 3 outfits with maintenance costs on a ship (does it automatically add all up)?

data/tooltips.txt
```diff
@@ -155,6 +155,9 @@ tip "jump speed:"
 tip "jump fuel:"
 	`How much fuel is consumed when you jump between systems using this drive.`
 
+tip "maintenance cost:"
+	`How many credits per day you need to spend to maintain this outfit.`
```
petervdmeer: "maintenance costs" suggests costs that you pay for just owning something, regardless of if it is in use. The implementation here only looks at non-parked ships, so I think the implementation below better fits the description "operating costs".

I would like to have both btw, maintenance costs for just owning something, and operating costs for having something active on a ship. (And they can show up on a single item in the accounts overview, no need to separate them there.)

data/tooltips.txt
```diff
@@ -155,6 +155,9 @@ tip "jump speed:"
 tip "jump fuel:"
 	`How much fuel is consumed when you jump between systems using this drive.`
 
+tip "maintenance cost:"
```
petervdmeer: My impression is that "maintenance costs" is usually written with an s on the end, but I'm not a native speaker, so please verify if there should be an additional s.

source/Account.cpp
```diff
@@ -191,20 +197,40 @@ string Account::Step(int64_t assets, int64_t salaries)
...
+	int64_t maintenancePaid = maintenanceDue;
+	if(maintenanceDue)
+	{
+		if(maintenanceDue > credits)
+		{
+			maintenancePaid = max<int64_t>(credits, 0);
+			maintenanceDue -= maintenancePaid;
+			credits -= maintenancePaid;
+			if(!missedPayment)
+				out << "You could not pay all your maintenance costs.";
```
petervdmeer: Add `missedPayment = true;` here, similar to salaries and morgage above.

source/PlayerInfo.cpp
Amazinite: `Ship::Attributes` returns all the attributes on a ship, including those from outfits, as `Ship::AddOutfit` adds the attributes of the outfits to the attributes of the ship.

source/Account.cpp
Amazinite: I've omitted that because overdue maintenance costs don't impact credit score, and if missedPayment being true is what causes your credit score to decrease.

data/tooltips.txt
Amazinite: Well this is a cost for a single outfit, so just "cost" sounds alright to me, but if others think it should be changed that I'll do it.

data/tooltips.txt
Amazinite: >I would like to have both btw

That shouldn't be a difficult change.

source/Account.cpp
petervdmeer: I think it's okay if overdue maintenance costs impact the credit score. You pay something too late (in this case maintenance) -> you get a lower credit score.

Amazinite: With the idea that [you're borrowing to pay for maintenance](https://github.com/endless-sky/endless-sky/issues/4284#issuecomment-541769571) if you don't have the credits on hand, or that operating costs are [to account for more expensive crew](https://github.com/endless-sky/endless-sky/issues/4284), I guess it isn't out of the question that overdue maintenance would impact your credit score.

source/Account.cpp
```diff
@@ -191,20 +197,40 @@ string Account::Step(int64_t assets, int64_t salaries)
 			++it;
 	}
```
petervdmeer: Just something to consider: Does it make sense to pay the maintenance (and operating) costs before the mortgages?

data/tooltips.txt
```diff
@@ -155,6 +155,9 @@ tip "jump speed:"
 tip "jump fuel:"
 	`How much fuel is consumed when you jump between systems using this drive.`
 
+tip "maintenance costs:"
+	`How many credits per day you need to spend to maintain this outfit. Even if a ship is parked, you still need to pay for its maintenance.`
```
petervdmeer: Sorry, but I have another exception/request; can we have those maintenance costs also for outfits in cargo? (Who-ever implements planetary storage should also add it there, but planetary storage is currently not merged, so you don't need to consider that now.)

source/PlayerInfo.cpp
petervdmeer: Okay, then this works as intended.

data/tooltips.txt
Amazinite: Yeah, I was just thinking that after adding operating costs that stored outfits would eventually also need to have their maintenance accounted for, and by extension in cargo outfits should now as well.

source/Account.cpp
Amazinite: Well now that overdue maintenance impact your credit score, maintenance doesn't need to be last and might as well occur alongside crew costs since they act in the same way.

data/tooltips.txt
Amazinite: Alright, should have this done. Given that the player is only charged for crew salaries/maintenance costs while in space, I think that only checking the ship cargo is necessary, and the player's pooled cargo never needs checked.

data/tooltips.txt
```diff
@@ -155,6 +155,9 @@ tip "jump speed:"
 tip "jump fuel:"
 	`How much fuel is consumed when you jump between systems using this drive.`
 
+tip "maintenance costs:"
+	`How many credits per day you need to spend to maintain this outfit. Even if a ship is parked or an outfit is in cargo, you still need to pay for its maintenance.`
```
petervdmeer: Not just outfits, maintenance costs can also apply to ship hulls.

data/tooltips.txt
```diff
@@ -164,6 +167,9 @@ tip "mass with no cargo:"
 tip "mass:"
 	`Mass, in tons. A ship's mass determines how fast it turns and accelerates.`
 
+tip "operating costs:"
+	`How many credits per day you need to spend to maintain this outfit while it is in active use.`
```
petervdmeer: Again, not just outfits. This attribute can also apply to ship hulls.

source/BankPanel.cpp
```diff
@@ -227,9 +248,12 @@ void BankPanel::Draw()
 	// Draw the "Pay All" button.
 	const Interface *interface = GameData::Interfaces().Get("bank");
 	Information info;
-	for(const Mortgage &mortgage : player.Accounts().Mortgages())
-		if(mortgage.Principal() <= player.Accounts().Credits())
-			info.SetCondition("can pay");
+	if((salariesOwed || maintenanceDue) && player.Accounts().Credits() != 0)
```
petervdmeer: What if player.Accounts().Credits() is negative?

source/PlayerInfo.cpp
```diff
@@ -687,6 +687,28 @@ int64_t PlayerInfo::Salaries() const
...
+// Calculate the daily maintenance cost for all ships and in cargo outfits.
+int64_t PlayerInfo::Maintenance() const
+{
+	int64_t maintenance = 0;
+	for(const shared_ptr<Ship> &ship : ships)
+		if(!ship->IsDestroyed())
+		{
+			maintenance += ship->Attributes().Get("maintenance costs");
+			auto outfit = ship->Cargo().Outfits().begin();
+			for(unsigned int i = 0; i < ship->Cargo().Outfits().size(); ++i)
```
petervdmeer: I would expect the loop to loop over the actual iterator `while(outfit != ship->Cargo()->Outfits().end()){...}`.

Or alternatively (but the Endless Sky coding guidelines prefer the use of the iterator), that the outfit within the loop is retrieved using variable `i`.

tehhowch: This should be a range-based for loop anyway, e.g. 
```c++
for( const auto &outfit : ...) {
  ...
```

data/tooltips.txt
Amazinite: One will only ever be able to see this tooltip while looking at an outfit, though.

data/tooltips.txt
Amazinite: As with above, this tooltip is only visible on outfits, hence why the tooltip mentions outfits.

source/BankPanel.cpp
Amazinite: Can it be?

data/tooltips.txt
petervdmeer: The "Ship Info" panel also has some tooltips.
But you are right, the tooltip here is not visible in the "Ship Info" panel, so the text here is correct.

data/tooltips.txt
petervdmeer: Ok

source/BankPanel.cpp
petervdmeer: Credits is a signed 64 bit integer value, so yes, in theory the number can be negative.
I expect that the number can go negative when a mission has a negative payment (but I didn't check this).
Loading a savegame with a negative number is probably also possible (but again I didn't check this).

My suggestion is to use some defensive programming here `player.Accounts().Credits() > 0`and thus handle the situation when the number is negative, even if we are not sure if the number ever actually goes negative.

Amazinite: Negative payment missions can’t offer/complete if you don’t have enough credits for the negative payment. 
As with many situations in game, I wouldn’t be surprised if forcing something to be negative via save editing or text file editing caused strange results. (See how negative slowing damage causes ships to spin rapidly then fling off into space.) Save editing to have negative credits does indeed give you negative credits in game, though.

Cat-Lady:
What is missing for merging this? I can't wait to use it in plugin(s).

Amazinite:
@tehhowch’s blessings.

tehhowch:
Color me curious, but why do outfits in cargo need maintenance? They cannot be used at all. At some time in the future, ammo in the cargo hold *may* be eligible as resupply for fighters, (and so I could understand performing maintenance on ammunition), but currently the only use of cargo is to either jettison as a distraction, or transport for sale/use.

I could see getting billed maintenance for just plundering an outfit from some ship being obnoxious (currently there's no way to tell if plundering something is going to cost you money, be it in fines or with this PR, maintenance, until you actually take the outfit).

Amazinite:
>Color me curious, but why do outfits in cargo need maintenance? They cannot be used at all.

Well if an outfit only needs maintained when in use, then give it `"operating costs"` instead of `"maintenance costs"`. An example of something that needs maintained even when not in use could be Zitchas' void rifle in #4222, which says `While durable, they rarely last more than a decade, so even the best cared for of the old Void Rifles degraded into uselessness centuries ago.`, implying that they need consistent maintenance in order to stay functional (although I'm not suggesting that this mechanic be used on the void rifle since it's intended to be vanilla content), as the missions also imply that the Remnant hadn't used any void rifles they may have had for some time, i.e. it isn't through use alone that the void rifle degrades.

Amazinite:
Right now one can't install a negative maintenance cost outfit because of the negative attribute check, but if one already has maintenance costs and they install something with a negative maintenance cost then it will reduce the total payment. If one uses save editing to force having negative maintenance cost outfits or the negative maintenance cost is put on the hull of the ship, then you'll gain that many credits daily.
image

Amazinite:
Fixed a bug where the bank panel wouldn't show the correct maintenance costs if you have outfits in cargo with maintenance costs, and made it so that total maintenance costs must be positive.

source/PlayerInfo.cpp
```diff
@@ -687,6 +687,32 @@ int64_t PlayerInfo::Salaries() const
...
+int64_t PlayerInfo::Maintenance() const
+{
+	int64_t maintenance = 0;
+	// If the player is landed, then all cargo will be in the player's 
+	// pooled cargo. Check there so that the bank panel can display the
+	// correct total maintenance costs. When launched all cargo will be
+	// in the player's ships instead of in the pooled cargo, so no outfit 
+	// will be counted twice.
+	for(const auto &outfit : Cargo().Outfits())
+		maintenance += outfit.first->Get("maintenance costs") * outfit.second;
```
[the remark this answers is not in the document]
Amazinite: I thought about it, and the same with checking per ship, but decided against that. I'm not married to the idea though, so if you think it's better to check per outfit and per ship then I'll do that.

Issue #4598, opened 2019-10-15
Warp Inhibitor/Jammer/Disruptor Mechanic


Amazinite:
**Is your feature request related to a problem? Please describe.**
A common trope in sci-fi is that of the warp inhibitor (or jammer, or disruptor, or what-have-you), but this is something that ES currently lacks. Whatever form that a warp inhibitor mechanic would take, its purpose would be to completely prevent one or more ships from jumping from the system.
Gameplay wise, this could create higher-stakes combat, as right now leaving combat is as simple as flying far enough (or just cloaking) to give you enough of a breather to jump, but if you were inhibited from jumping then you would be forced to either fight or die instead of being able to just leave.
Warp inhibiting would differ from current forms of preventing a ship from jumping (hit force, slowing damage) in that preventing jumping would be the *only* effect, and it would be a guaranteed effect as well (as one can still escape while being impacted with hit force or slowing damage under certain circumstances).

**Related Issue Links**
My second issue brought this idea up (#855), and even developed a somewhat hacky solution, but I'd like to retackle this issue now that I have more knowledge of the game.
This idea would also benefit from field/aura weapons (mentioned in #2847).

**Questions that need answered**
Warp inhibiting could take many forms, so we need to figure out either what form of warp inhibiting we want or what various forms we want to support.
* Should warp jamming be targeted at a single ship, a set number of multiple ships (i.e. allowing one weapon to have multiple targets), or be system wide?
* If warp jamming is an AOE effect, should it only affect hostile ships or all ships?
* Should warp jamming have a range to it? (Ranged warp jamming could be similar to #4334.)
* Should warp jamming be able to discriminate between what drive the ship has? e.g. only affect hyperdrives or only affect jump drives.
* Should warp jamming be able to affect ships of different masses differently? e.g. have smaller ships be less affected as to have a counter to being jammed. Or maybe heavier ships are less affected. Do we choose one to favor or allow the weapon to specify a mass range of what's affected?
* Should warp jamming affect cloaked ships?

**Describe the solution you'd like**
My current thought is that warp jamming would come in the form of `"jump disruption"` on weapons. `"jump disruption"` would mechanically work the same as #3599's cloak disruption. That is, a ship will gain jump disruption when impacted by a weapon with that damage type. So long as the ship has jump disruption, it will be unable to jump, but jump disruption is decreased with every frame until it reaches 0 and the ship is able to jump again.
[...]

PR #1664, opened 2016-10-02
Hyperspace boosters

jafdy:
In effect this adds the "Hyperspace Boosters" mod to the main game.
I almost feel obliged after winning the contest.
~~The outfit sprites are different from Wrzl's original. With the @2x images Ill need to get in touch with @FallenCat for the purple one but I can PR the orange one quickly enough.~~

PR #4601, opened 2019-10-16
Hyperdrive and Jump Disruption damage types

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4598. 
I merged #3599 into this branch to build off of since both that PR and this one are adding new damage types, and #3599 made some cleaning changes: basically just avoiding conflicts. I'd expect #3599 to be merged before this, though. Just mind those changes in this PR.

## Feature Details
Adds warp jamming attributes as weapon damage types. As with cloaking disruption's effect on cloaking (see #3599), having hyperdrive/jump disruption on your ship means that you can not jump until the disruption has completely dissipated. I designed this to have the highest degree of flexibility given the questions laid out in #4598.

If one of a ship's drives are disabled, then it will still be able to use the other drive. But, if an escort's hyperdrive is disabled and you use a jump drive to get to another system, the escort won't use its hyperdrive to reach you and will instead wait for its jump drive to reactivate and then follow you. This wasn't intentional, but I'd say is preferable, as otherwise the escort could potentially begin following you by HD, then have its JD reactivate and need to backtrack, wasting fuel.

Disruption is bled at a rate of `1 + jump speed / default jump speed` per frame, where default jump speed = 0.2 for hyperdrive disruption and 0.3 for jump drive disruption. This means that for each additional hyperdrive and/or jump drive that a ship has installed, the disruption will dissipate quicker. The default expected bleed value of most ships then is 2. Should these attributes become prevalent in a certain region of space, I'd suggest looking into reviving hyperdrive boosters (#1664) as an outfit to counter drive disruption.

The four new attributes are as follows.
`"hyperdrive disruption" <x>`
`"hyperdrive disruption effect" <min> <max> <deviation>`
`"jump disruption" <x>`
`"jump disruption effect" <min> <max> <deviation>`

The disruption attributes determine the base value of disruption that is applied to the impacted ship. The disruption effect attributes are used to modify the base value based off of the mass of the ship being impacted. The purpose of this is to allow for warp inhibitors that may only impact ships of certain sizes, or may simply impact some ships better than others. 
The min value is the minimum mass and the max value is the maximum mass, while the deviation impacts how ships outside of the min-max range are affected. The disruption effect does not need to be specified, which results in the default behavior of affecting all ships to the same degree, but if it is, the max and deviation values are optional. Not providing a max value could be used to create a disruptor that only impacts a specific mass (and therefore perhaps a specific mission ship), and not providing a deviation value means that only those ships inside the given range are affected. If a deviation value is given, then ships outside of the min-max range are able to be affected, but to a lesser degree. Negative deviation values are allowed, and this flips the effect such that those ships inside the range will now not be impacted, while those ships outside the range are to a greater degree the further from the range that they are.
[This](https://www.desmos.com/calculator/c2yag7wolv) Desmos graph shows how the deviation affects the disruption for ships outside the given mass range. The equation itself is based off of a normal curve, with the deviation simply being the standard deviation of the curve.

## UI Screenshots
Instead of the usual effect created around the ship like for ionization and similar, I decided to go for a UI effect that tells you when your drive is disabled. This doesn't tell you when other ships are disabled though, so a visual effect on the ship may still be in order.

If only your hyperdrive is disrupted, your fuel gauge blinks red.
If only your jump drive is disrupted, your fuel gauge blinks blue.
If both drives are disabled, your fuel gauge blinks purple.
These colors are subject to change (and probably should in some way, as the blue and purple can look similar for certain types of color blindness).
fuel gauges

## Usage Examples
```
# This weapon would disrupt both drives for ships between 400 and 800 tons.
outfit "Warp Inhibitor"
	...
	weapon
		...
		"hyperdrive disruption" 10
		"hyperdrive disruption effect" 400 800
		"jump disruption" 10
		"jump disruption effect" 400 800

# This weapon would disrupt hyperdrives on all ships, but ships between 400 and 800 
# would be hit the most.
outfit "Hyperdrive Disruptor"
	...
	weapon
		...
		"hyperdrive disruption" 10
		"hyperdrive disruption effect" 400 800 10

# This weapon would disrupt all jump drives.
outfit "Jump Jammer"
	...
	weapon
		...
		"jump disruption" 10

# This weapon would disrupt jump drives on all ships outside 
# of 1000 to 2000 tons, but ships closer to the range are less affected
outfit "Mosquito Catcher"
	...
	weapon
		...
		"jump disruption" 10
		"jump disruption effect" 1000 2000 -50
```

## Questions

* Should these disruption effects create visual effects on ships?
* Does anyone have a better name suggestion for the `disruption effect` attributes?
* Currently the disruption effect attributes don't display anywhere in the outfitter. I have no idea how one would show that information. Should these attributes just be left out of the outfitter with the description of the weapon describing what it is that the weapon affects?
* Currently these disruption effects aren't impacted by leakage (i.e. having the shields up doesn't decrease the incoming disruption). Should this be changed? I'd personally argue against this, as unlike the other attributes, jump disruption could effectively leave you completely stuck for a fair amount of time if the disruption you took was high, and the difference between shields up and shields down is rather drastic and could therefore lead to difficult to balance jump disruption.
* Given that a ship without a hyperdrive or jump drive doesn't need to bleed that disruption type, should the additional +1 be removed from the disruption bleed per frame? This would mean that the expected bleed value of most ships would be 1. The only thing this would really change is the value that drive disruption weapons would have.
* Should there be a maximum possible disruption on a ship as to avoid a scenario where the player is forced to stick around an uninhabited system for a while because they got hit with high amounts of disruption? Say 60 seconds.

### DIFF
--- a/data/drak.txt
+++ b/data/drak.txt
@@ -145,6 +145,7 @@ outfit "Drak Antimatter Fragment"
 		"hit force" 1000
 		"shield damage" 5610
 		"hull damage" 5610
+		"cloak disruption" 150
 		"blast radius" 100
 
 effect "antimatter spark"
--- a/data/effects.txt
+++ b/data/effects.txt
@@ -43,6 +43,15 @@ effect "slowing spark"
 	"random spin" 30
 	"random velocity" .1
 
+# TODO: Use a unique effect and sprite
+effect "cloak disruption spark"
+	sprite "effect/nano spark"
+		"random start frame"
+		"frame rate" 16
+	"lifetime" 1
+	"random angle" 360
+	"random spin" 4
+
 
 effect "flame"
 	sprite "effect/flame"
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -40,6 +40,9 @@ color "heat" .70 .43 .43 .75
 color "overheat" .70 .61 .43 .75
 color "energy" .6 .6 .6 .75
 color "fuel" .70 .62 .43 .75
+color "hyper disrupted" .5 .0 .0 .75
+color "jump disrupted" .0 .0 .5 .75
+color "double disrupted" .35 .0 .35 .75
 
 color "flagship highlight" .5 .8 .2 0.
 
@@ -340,6 +343,21 @@ interface "hud"
 		from -53.5 425
 		dimensions 0 -192
 		color "fuel"
+		size 2
+	bar "hyper disrupted"
+		from -53.5 425
+		dimensions 0 -192
+		color "hyper disrupted"
+		size 2
+	bar "jump disrupted"
+		from -53.5 425
+		dimensions 0 -192
+		color "jump disrupted"
+		size 2
+	bar "double disrupted"
+		from -53.5 425
+		dimensions 0 -192
+		color "double disrupted"
 		size 2
 	bar "energy"
 		from -33.5 415
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -374,6 +374,12 @@ tip "slowing damage / second:"
 tip "disruption damage / second:"
 	`Disrupts the target's shields, allowing weapon damage to "leak through" to the hull even if shields are up. Shield disruption wears off over time.`
 
+tip "drive disruption / second:"
+	`Disrupts the target's hyperdrive, preventing them from using it. Hyperdrive disruption wears off over time.`
+
+tip "jump disruption / second:"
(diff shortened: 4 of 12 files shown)


Amazinite:
*sighs*. If you have one drive and the other is disrupted, the flagship can't jump, but other ships can. I'll look into it later, because I just spent 6 hours on this.

Amazinite:
I think that this is unnecessarily complex. I'm just going to close it and come back to this idea later.

PR #4605, opened 2019-10-18
New Laser Rifle image

Brick63:
update the Laser Rifle with a new sprite
[laser rifle assets.zip](https://github.com/endless-sky/endless-sky/files/3763405/laser.rifle.assets.zip)

laser rifle@2x


Brick63:
@Pointedstick that better?

Pointedstick:
Actually here's one more thing for maximum realism: this gun has a rear sight, but no front sight. Since your rear sight is two posts, the front sight should be a single post at the front of the gun that sits between them when viewed from the firer's eyeball. A  person wielding the rifle should see a sight picture that looks like this:

ProperSightPicture-627x381

That's a sight picture for a pistol of course, because you've chosen to use pistol sights for this rifle. Rifle icon sights are typically more like this: https://www.ecosia.org/images?q=rifle+iron+sight. You don't need to change to a more typical rifle icon sight if you don't want to, but a front sight of some sort would be nice. :)

Brick63:
that may not work with how the rifle barrel is lower than the body, meaning a front sight like normal would be a protrusion that looks out of place imo.

Amazinite:
It's not unusual IRL for front sights to be a fair bit above the barrel.
image

Pointedstick:
Yeah when the top of the stock is on the same plane as the top of the receiver, the sights need to be elevated another inch or so due to the positioning of the human eye. This is why in most long guns, the top of the stock is lower down--and then the sights can be lower down, too. AR-15 style rifles (and derivatives) have tall sights because the stock is on the same plane as the receiver, because the stock is actually hollow so that the bolt can slide into it. Presumably there's no such restriction here, but the top of the receiver on your rifle is higher up than the end of the barrel anyway, so the on;y way to avoid having a tall front sight would be to lower the top of the receiver and stock as well. Not sure it's really necessary though. A tall front sight is fine IMO.

Cat-Lady:
Actually, I believe that adding such sight-elevating element to the barrel would fix some issues with the rifle - namely, that the barrel part looks kind of bland. At least, that seemed to be universal consensus in the Discord discussions that took place after presenting the latest iteration.

Brick63:
@Pointedstick @Cat-Lady added the raised sight, and some more decorations for the barrel

Amazinite:
Maybe shrink the image to be closer to the original 120x90. (Right now yours is 150x105, and there's a fair amount of whitespace below the weapon, so that odd 105 height isn't necessary.)

Cat-Lady:
What do you people think of that:
laser rifle

...? I was trying to make the gun look less bland, at the same time, making more strong resemblance to the tech behind ship's beam laser - the generator/capacitors part being yellow(ish), and only the barrel part red:
laser@2x

Thoughts? Better? Worse? Indifferent?

PR #4607, opened 2019-10-18
Heliarch shield generation and hull repair

Arachi-Lover:
this PR includes military-grade shield generators/hull repair outfits exclusive to the Heliarch.
most* variants also now utilize these new outfits by default.

*the Nimble variant of the Neutralizer could not fit both the Overcharged Shield Module and the Overclocked Repair Module, so instead it kept it's Large Repair Module, only replacing the Shields.

thanks @beccabunny for the images


MCOfficer:
Are there any assets for this? @beccabunny

Amazinite:
We also need a high DPI PR.

MCOfficer:
> 
> 
> We also need a high DPI PR and a copyright update.

that, too? why was this merged, then?

PR #4611, opened 2019-10-20
Correction to the behemoth

beccabunny:
Anarchist on Discord noticed that the Behemoth doesn't have a free pixel on the upper side; this PR shortens a little the cockpit, not changing any hardpoint. New files: 
[behemoth.zip](https://github.com/endless-sky/endless-sky/files/3748057/behemoth.zip)


Pointedstick:
It is a problem for it to not have line of transparent pixels on the very top?

beccabunny:
I knew that the ships need to have one free pixel on every side for the outline to work correctly, if that's not correct anymore then the "creating ships" page could need updating

Amazinite:
I know that it use to ruin the collision mask, causing shots to go straight through the ship (and at one point selecting a ship without a collision mask would cause a crash), but testing the current Behemoth, it seems to be fine. I think it's still a good idea to leave a gap of white space around ships, though. 
@beccabunny Have you double check all the other ships you've edited?

beccabunny:
@Amazinite yep

Pointedstick:
No objections, I was just curious. :)

PR #4634, opened 2019-11-02
Fixing problematic changes to the Petrel and Tern

Zitchas:
## Fix Details:
The Petrels were erroneously converted into being close-quarter fighters. This fixes that change.

## Changes:
*Petrel:*
- restores their fuel capacity to 50, instead of the 100 it was pre-Derp
- ramscoop restored to 0.5, instead of the 1 it was before.
- engine capacity back up to 34, exactly the same as before (and the minimum to fit the crucible engines)
- scanning restored (10 from 12, and 20 from 26. So both ranges are shorter than they were)
- 1 inhibitor cannons again instead of thrashers (they originally had inhibitors. This is a long-range support fighter, NOT close quarters. Remnant aren't suicidal)
- restored the description that explains how the Remnant use these fighters.
*Tern:*
- scanning restored (10 from 12, and 20 from 26. So both ranges are shorter than they were)

### DIFF
--- a/data/remnant ships.txt
+++ b/data/remnant ships.txt
@@ -592,7 +592,7 @@ ship "Petrel"
 		"required crew" 2
 		"bunks" 3
 		"mass" 50
-		"drag" 1.8
+		"drag" 1.3
 		"heat dissipation" .8
 		"outfit space" 133
 		"weapon capacity" 33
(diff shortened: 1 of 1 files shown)


Zitchas:
With the Petrel in its current state (drag 1.4), its speed and offense balances out thus:

- 2 inhibitors with the anvil, top speed 578
- 1 point defense turret with the anvil, top speed 578
- 1 inhibitor with crucible set, top speed 771

I think this is a good balance point for it. The default configuration (2 inhibitors) is effective against raiders, and sacrificing offensive power for speed in the 1 inhibitor with crucible set are fast enough to chase down Chasers, but only barely. Allowing players to make a offense/speed trade-off like this I think is valuable from a gameplay side of things

I am fully aware this isn't an exhaustive list of weapon/engine combinations, but that's an exercise in futility.

Zitchas:
Changes:
- Reduced Petrel cargo to 15 (from 19)
- Changed drag to 1.44

Results:
- Petrel with 2 inhibitors and Anvil has a max of 562
- Petrel with 1 inhibitor and Crucible has a max of 750

PR #4635, opened 2019-11-02
More weapon hardpoints

painted-sky:
-----------------------
## Feature Details
Added the capability to use more than 32 weapon hardpoints on a ship.

Switches from using a 32-bit lower word of Command.state to a set of 8 64-bit flags for the firing weapons, and increases the size of the turret aiming data array from 32 to 512.

## UI Screenshots
Screen Shot 2019-11-02 at 10 23 47
A ship with 400 laser turrets.


Zitchas:
That is impressive. I'll be testing this just as soon as I have time to re-learn how to compile ES again. (I've done it before... It'll just take a bit)

jafdy:
Not to be that guy: But Why?
Also, is there any performance change from the switch?

beccabunny:
It gives plugin creators much more freedom on ship design

Zitchas:
The quarg were probably thinking the same thing: "Come on, we only limited you to an area a quarter the size of the galaxy and stopped a few wars. Why are you suddenly all 'you're sticking us in cages!'?"

On a more serious note, while the numbers are kind of ridiculous (500 turrets?!) I think this helps open things up for supporting bigger ships, especially proper heavy cruisers, battleships, and true large scale carriers.  Such ships, I think, could probably justify getting up to the 30-50 range, maybe a bit more. Probably mostly in the realm of the plug-in creators, but I could see some vanilla content getting to that eventually. If we keep this up for another few years, I could see us eventually getting into tier 3 or tier 4 stuff (as player purchasable ships) going up against potentially tier 5 enemies...

painted-sky:
There doesn't seem to be a major performance change, but I'd appreciate it if someone else could test that.

painted-sky:
Realized that I was testing with the wrong version of code; will reopen when working.

Zitchas:
OK. Let me know when you have it working.

painted-sky:
Alright, working again

Performance doesn't honestly seem to be too much of an issue—firing 400 turrets in the middle of a battle only increased CPU usage by 2~3% (late-2014 mac mini).

source/Ship.cpp
```diff
@@ -201,9 +201,6 @@ void Ship::Load(const DataNode &node)
 				armament.AddGunPort(hardpoint, outfit);
 			else
 				armament.AddTurret(hardpoint, outfit);
-			// Print a warning for the first hardpoint after 32, i.e. only 1 warning per ship.
-			if(armament.Get().size() == 33)
-				child.PrintTrace("Warning: ship has more than 32 weapon hardpoints. Some weapons may not fire:");
```
Amazinite: Why remove this warning entirely if a limit still exists?

painted-sky: Ah, yeah, good point.

painted-sky: Fixed with 918d1ad

source/Command.h
```diff
@@ -120,23 +120,20 @@ class Command {
...
 	
 private:
-	// The key commands and weapons to fire are stored in a single bitmask, with
-	// 32 bits for key commands and 32 bits for individual weapons.
-	// Ship::Load gives a soft warning for ships with more than 32 weapons.
-	uint64_t state = 0;
+	// Turret turn rates, reduced to 8 bits to save space.
+	signed char aim[512];
+	// 512 bits for weapons firing.
+	uint64_t firing_weps[8];
```
tehhowch: Did you consider using std::bitset?

source/Command.h
```diff
@@ -68,7 +68,7 @@ class Command {
 	static std::string ReplaceNamesWithKeys(const std::string &text);
 	
 public:
-	Command() = default;
+	Command();
```
tehhowch: ?

source/Command.cpp
```diff
@@ -235,21 +245,34 @@ void Command::Clear(Command command)
...
-	return (state & command.state);
+	if(state & command.state)
+		return true;
+	for(size_t i = 0; i < 8; i++)
+	{
+		if(firing_weps[i] & command.firing_weps[i])
+			return true;
+	}
+	return false;
+	// return (state & command.state);
```
tehhowch: If this is dead code, it should be removed, not just commented out.

source/Command.cpp
```diff
@@ -235,21 +245,34 @@ void Command::Clear(Command command)
...
 }
 
 
 
 // Get the commands that are set in this and not in the given command.
 Command Command::AndNot(Command command) const
 {
+	// At the moment, this is only used for key commands (AI.cpp:288)
```
tehhowch: A specific line reference is meaningless.

source/Command.h
```diff
@@ -120,23 +120,20 @@ class Command {
...
+	explicit Command(uint32_t state);
+	Command(uint32_t state, const std::string &text);
 	
 private:
-	// The key commands and weapons to fire are stored in a single bitmask, with
-	// 32 bits for key commands and 32 bits for individual weapons.
-	// Ship::Load gives a soft warning for ships with more than 32 weapons.
-	uint64_t state = 0;
+	// Turret turn rates, reduced to 8 bits to save space.
+	signed char aim[512];
```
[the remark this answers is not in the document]
painted-sky: Reverted.

source/Command.h
```diff
@@ -120,23 +120,20 @@ class Command {
...
+	// 512 bits for weapons firing.
+	uint64_t firing_weps[8];
 	// Turning amount is stored as a separate double to allow fractional values.
 	double turn = 0.;
-	// Turret turn rates, reduced to 8 bits to save space.
-	signed char aim[32] = {};
+	// The key commands are stored in a single bitmask, with
+	// 32 bits for key commands
+	uint32_t state = 0;
```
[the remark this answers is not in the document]
painted-sky: Reordering

source/Command.h
```diff
@@ -120,23 +120,20 @@ class Command {
 	Command operator|(const Command &command) const;
 	Command &operator|=(const Command &command);
 	
-	
 private:
-	explicit Command(uint64_t state);
-	Command(uint64_t state, const std::string &text);
-
```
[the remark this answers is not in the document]
painted-sky: Reverting

source/Command.cpp
```diff
@@ -364,23 +388,24 @@ Command &Command::operator|=(const Command &command)
...
 
 
 
 // Private constructor.
-Command::Command(uint64_t state)
-	: state(state)
-{
-}
+Command::Command(uint32_t state)
+		: state(state), firing_weps{0, 0, 0, 0, 0, 0, 0, 0}
```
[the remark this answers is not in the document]
painted-sky: You're right; never actually realised that, switching to in-decl initialisation.

source/Command.cpp
```diff
@@ -326,22 +349,23 @@ void Command::SetAim(int index, double amount)
 // Check if any bits are set in this command (including a nonzero turn).
 Command::operator bool() const
 {
-	return !!*this;
+	return !!*this || IsFiring() || turn;
```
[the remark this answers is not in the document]
painted-sky: Fixing

painted-sky:
Did some profiling of tribute battles in Sol system.

512 weapons seems to take Command::*() functions from about 0.2% of CPU time to 0.5%.
This seems mainly to be due to Command::Has() and Command::IsFiring().

Using std::bitset<N> seems to get about the same performance as my implementation.

source/Command.cpp
```diff
@@ -274,29 +291,30 @@ double Command::Turn() const
...
+	firing_weps[index >> 6] |= (1ull << (index % 64));
 }
 
 
 
 // Check if any weapons are firing.
 bool Command::IsFiring() const
 {
-	return (state & 0xFFFFFFFF00000000ull);
+	return (bool) (firing_weps[0] | firing_weps[1] | firing_weps[2] | firing_weps[3] |
```
tehhowch: C-style casts should be avoided whenever possible. If you must cast explicitly, use `static_cast`

source/Command.cpp
```diff
@@ -164,7 +164,8 @@ void Command::SaveSettings(const string &path)
 void Command::SetKey(Command command, int keycode)
 {
 	// Always reset *all* the mappings when one is set. That way, if two commands
-	// are mapped to the same key and you change one of them, the other stays mapped.
+	// are mapped to the same key and you change one of them, the other stays
```
tehhowch: Revert this wrapping change

source/Command.cpp
```diff
@@ -364,23 +382,24 @@ Command &Command::operator|=(const Command &command)
...
 
 
 // Private constructor.
-Command::Command(uint64_t state)
-	: state(state)
-{
-}
+Command::Command(uint32_t state)
+		: state(state)
+		{}
```
tehhowch: Revert unnecessary changes

source/Command.h
```diff
@@ -122,21 +122,19 @@ class Command {
...
 	
 private:
-	explicit Command(uint64_t state);
-	Command(uint64_t state, const std::string &text);
+	explicit Command(uint32_t state);
+	Command(uint32_t state, const std::string &text);
 	
 	
 private:
-	// The key commands and weapons to fire are stored in a single bitmask, with
```
tehhowch: Is the reordering intentional?

painted-sky: Was trying to get a bit of speed from trying to keep the big data block in cache, however, after profiling, it looks like firing_weps might actually need that more; by putting `aim` last, it hopefully makes firing_weps is faster to access.

Additionally, making `aim` the 'dangling' variable is a matter of risk mitigation: at the moment, ships generally only inhabit the lower reaches of the `aim` array, making high-address cache misses less likely.

source/Command.cpp
painted-sky: Done.

source/Command.cpp
painted-sky: Done.

source/Command.cpp
painted-sky: Switched to using logical or instead of bitwise or to reap possible performance gains from short-circuit logic that bitwise or ignores.

FishermanHMeM:
> 512 turrets

_Blasters overwhelming._

At least it isn't 65536 hardpoints. While I can't see a ship with more than 30 hardpoints (unless it's maradeur's capital ship with overwhelming blaster fire from _both_ guns and turrets), i find this PR useful.

Also, T3 and T4 fighting vs T5 won't use many weapons, they'll use heavier weapons because even now heavier warships while have more weapon space, have equal or less amount of hardpoints.

TheronEpic:
no the 32 limit was put there for balance reasons you shouldn't remove it

haha 512 gatling guns go _**BRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRRT**_

InfiniteDonuts:
Good idea, but... why?

Zitchas:
@painted-sky  Could you bring this up to par with the ES master? This could be useful to add to the Combat Revamp test PR #5772 . Updating it would make the merging much easier on my side. Thanks! This could also provide valuable support for the contemplated addition of default emission points and the possibility of allowing things to *change* how many hardpoints a ship has.

To my non-coding perspective, this looks like a fairly straightforward improvement to the engine with no real problems on the front-end. Chances are vanilla ships will never have more than 32 hardpoints, but there's no reason why plug-in ships shouldn't (Star Wars recreation plug-ins come to mind for breaking this).

Also, I see no problem why a T3/T4 race might not want to use mass numbers instead of higher power. There's a lot of philosophies in the galaxy, and some much prefer "drown them in numbers" rather "punch harder."

FIM43-Redeye:
Original dev has disappeared off Github. Anyone willing to pick up this request and replicate it, or just close it up somehow?

Pointedstick:
I guess we should just close it. :/

Zitchas:
I'd have picked this up if I had even the slightest idea what I was doing, but ah well. Maybe someone else will, some day.

Issue #4172, opened 2019-02-02
FW Southern Recon 1B: Alnasl is too unintuitive


Bladewood:
In FW Southern Recon 1B, the player is tasked with stealing an electron beam/turret off of a gunboat. Freya recommends either Wei or Alnasl.
On my first several playthroughs, it *never* occurred to me that trying Alnasl would have any improvement. I had always painstakingly farmed Wei, hoping that RNG would spawn a small fleet that I could sweep through and steal a beam from, or perhaps that I could lead away into Kornephoros for the Free Worlds to pick them off. (Yes, I know fleets don't actually follow the player.)

This problem has appeared on the Steam forums and on the server too many times for me to count. If Alnasl is intended to be the real solution, I think that it should be made more obvious somehow.

Lineth:
This mission is one of the few I've unabashedly cheated past. I've near lost count of the times I actually disabled a gunboat only to have half the Republic jump in on me.

endless-sky:
It doesn't seem unreasonable to me to expect a player to be able to say, "Hey, this system is awfully high-traffic, maybe I should look elsewhere." Freya isn't psychic; like everyone else in the game, she's working with incomplete knowledge, and you need to take what she says with a grain of salt. To me, that feels better than spoon-feeding the player directions about how exactly they should complete the mission.

Amazinite:
One of the problems isn't just that Alnasl is where players should be going when they're given two possible systems, it's that you have to pass through Wei to get to Alnasl in the first place, and Wei can have large Navy fleets that are pretty much impossible to deal with on your own in the type of ship you're expected to have at the time. Perhaps the `"gunboats in Alnasl"` event should also reduce the frequency of larger Navy fleets in Wei to make it easier to not only get to Alnasl, but also brute-force it and get your Electron Beam in Wei.

pyorokun7:
Having completed this mission last night, I wouldn't mind this addressed somehow.

After failing some 10-15 times to kill a lone Gunboat in Wei and survive long enough to jump back, I figured that I could try in Alnasl... then I needed at least 10 attempts to go PAST Wei / that no fleet followed me in Alnasl / there wasn't a too big fleet in Alnasl, so I could finally loot my Electron Beam.

I didn't even try to get more than one, I just wanted to complete the mission.

[...]

PR #4636, opened 2019-11-03
Make stealing an electron beam more consistent and reasonable

Amazinite:
**Content (Missions)**

## Summary
Ref: #4172
The difficulty of this mission has been of concern for a while now. This change should make stealing an electron beam more consistent and reasonable now.

1. Wei no longer spawns large Republic fleets doing the duration of the mission. Given that the player is only expected to be in a single ship, it's unreasonable to expect the player to be able to either take on a large Republic fleet solo or be capable of always evading a large Republic fleet when jumping to Alnasl.
2. Greatly reduced the spawn rate of Mark II Gunboats in Alnasl, and added an NPC Gunboat spawned by the mission. This means that the player is guaranteed to run into at least one Gunboat in Alnasl upon entering, and is less likely to have to deal with multiple Gunboats at once.

### DIFF
--- a/data/events.txt
+++ b/data/events.txt
@@ -1175,6 +1175,7 @@ event "navy using mark ii ships"
 			"Frigate (Mark II)"
 			"Rainmaker (Mark II)" 2
 			"Gunboat (Mark II)"
+
 fleet "gunboat only"
 	government "Republic"
 	names "republic small"
@@ -1188,7 +1189,7 @@ event "gunboats in Alnasl"
 	system "Alnasl"
 		fleet "Small Southern Merchants" 1300
 		fleet "Large Southern Merchants" 2500
-		fleet "gunboat only" 3600
+		fleet "gunboat only" 2400
 	system "Wei"
 		remove fleet "Large Republic"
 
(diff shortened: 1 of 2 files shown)


jafdy:
You could also have a repeating mission that spawns one there. It is unlikely to make much difference but if a player does land then they are back where they started at least.

data/events.txt
```diff
@@ -1176,29 +1176,23 @@ event "navy using mark ii ships"
...
-		heroic
-	variant
-		"Gunboat (Mark II)"
-
 event "gunboats in Alnasl"
-	system Alnasl
+	system "Alnasl"
 		fleet "Small Southern Merchants" 1300
 		fleet "Large Southern Merchants" 2500
-		fleet "gunboat only" 800
```
Zitchas: Are you intending to remove this "gunboat only" fleet? This change will mean that there won't be any ongoing republic fleet spawns in this system. Might it be good to leave the gunboat spawn in, although potentially lowering the spawn rate a touch to avoid stacking too many on the player might be good.

Zitchas: Looks good now.

PR #4646, opened 2019-11-06
Create short Hai mission string to unlock the Shield Beetle

Darcman99:
1 Hai planetary defense mission
2 follow-up missions on dealing with the pirate threat

## Summary
This PR addresses the blatant issue of the Shield Beetle, an immensely strong early/mid-game heavy warship, being available so early and so easily to players. The changes address the issue by locking the Shield Beetle behind a license and creating a short mission string that leads the player to being provided with the license, thus allowing the player to purchase the ship.
image

## Plugin File
This plugin file can be used to play through the new mission content:
[Hai Enlisted Missions.zip](https://github.com/endless-sky/endless-sky/files/3812723/Hai.Enlisted.Missions.zip)

## PR Checklist
#### Irrelevant:
 - ~~[ ] I uploaded the necessary image, blend, and texture assets here:~~
 - ~~[ ] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets:~~

### DIFF
--- a/data/hai missions.txt
+++ b/data/hai missions.txt
@@ -1069,16 +1069,16 @@ mission "Enlisted by Hai [1]"
 	description "Defeat a pirate raid on <destination>."
 	to offer
 		"combat rating" > 400
-		"random" < 35
+		"random" < 25
 	source "Allhome"
 	npc
 		government "Hai"
-		personality "heroic" staying"
+		personality heroic staying
 		fleet "Large Hai" 2
 		fleet "Large Militia"
 	npc evade
 		government "Pirate"
-		personality "heroic" "staying"
+		personality coward waiting
 		fleet "Large Northern Pirates" 10
 	on offer
 		conversation
@@ -1096,7 +1096,7 @@ mission "Enlisted by Hai [1]"
 
 mission "Enlisted by Hai [2]"
 	name "Enlist with the Hai"
-	description "You have been recommended for your attentiveness for keeping the peace throughout Hai space. Hai officials would like to consult you about possible enlistment."
+	description "You have been recommended following your demonstration of peacekeeping. Hai officials would like to consult you about possible enlistment."
 	priority
 	to offer
 		has "Enlisted by Hai [1]: done"
@@ -1109,11 +1109,11 @@ mission "Enlisted by Hai [2]"
 	destination "Hai-home"
 	on offer
 		conversation
-			`As soon as you depart your ship you discover an envoy of sorts awaiting you. "Few human visitors demonstrate care for the lasting peace we have established so long ago," the Hai begins hastily. "You are a prominent example of your kind, and we request audience on <destination>. The unfettered humans have revealed a tentativeness to prey on those who follow a path unlike their own."`
+			`As soon as you depart your ship you discover an envoy of sorts awaiting you. "Not all human visitors demonstrate care for the lasting peace we have established so long ago," the Hai begins hastily. "You are a prominent example of your kind, and we request audience on <destination>. The unfettered humans have revealed a tentativeness to prey on those who follow a path unlike their own."`
 			choice
 				`	"I suppose I can hear you out."`
 				`	"I would prefer to not get caught up in any trouble."`
-					defer
+					decline
 			`	"That is most pleasing," it says. "the consuls will be notified of this news and will be expecting attendance."`
 				accept
 
@@ -1124,11 +1124,11 @@ mission "Enlisted by Hai [3]"
 	to offer
 		has "Enlisted by Hai [2]: offered"
 	source "Hai-home"
-	npc "kill"
+	npc kill
 		government "Pirate"
 		dialog
 			`You have eliminated the pirate leader's ships. Hopefully this sends a warning to any other pirate warlords who meddles in the affairs of the Hai.`
-		personality "disables" "heroic" "nemesis" "plunders" "staying" "target"
+		personality disables heroic nemesis plunders staying target
 		system "Oblate"
 		ship "Falcon (Heavy)" "Gash"
 		ship "Splinter (Laser)" "Gnar"
--- a/data/hai missions.txt
+++ b/data/hai missions.txt
@@ -1067,19 +1067,10 @@ mission "Hai Honeymoon"
 mission "Enlisted by Hai [1]"
 	name "Defend <planet>"
 	description "Defeat a pirate raid on <destination>."
+	source "Allhome"
 	to offer
 		"combat rating" > 400
 		"random" < 25
-	source "Allhome"
-	npc
-		government "Hai"
-		personality heroic staying
-		fleet "Large Hai" 2
-		fleet "Large Militia"
-	npc evade
-		government "Pirate"
-		personality coward waiting
-		fleet "Large Northern Pirates" 10
 	on offer
 		conversation
 			`Suddenly you notice a massive rush in the spaceport: "We are under attack! <planet> is under attack by the unfettered humans! We need every combat-worthy ship to join in the defenses!" The authorities will probably be very gracious, but this could also be an easy way to get yourself killed.`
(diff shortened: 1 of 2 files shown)


data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
+		"random" < 35
+	source "Allhome"
+	npc
+		government "Hai"
+		personality "heroic" staying"
+		fleet "Large Hai" 2
+		fleet "Large Militia"
+	npc evade
+		government "Pirate"
+		personality "heroic" "staying"
```
tehhowch: It might make sense to have part of the pirate fleet already present, and the rest starting in the neighboring system. That would give the impression of a sustained attack / reinforcements, and make it slightly less reload-y if the player has bad launch RNG.

Is it desired that these pirates do not retreat when the battle is lost?

data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
 		dialog `Instead of landing in the spaceport where they can be seen, you drop Anaya and Touhar off as close as you can to where they will be staying. The house is a few kilometers away, but they assure you that they can make it on their own. They pay you <payment> and thank you for bringing them this far.`
+
+
+
+mission "Enlisted by Hai [1]"
+	name "Defend <planet>"
+	description "Defeat a pirate raid on <destination>."
+	to offer
+		"combat rating" > 400
+		"random" < 35
```
tehhowch: This is a pretty high offer rate

data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
+			`A mix of Hai and Human pilots have amassed to help repel the unfettered attack. You join them, and take off together...`
+				launch
+	on complete
+		dialog
+			`The government of <planet> pays you <payment> for helping to drive off the attack.`
+		payment 280000
+
+mission "Enlisted by Hai [2]"
+	name "Enlist with the Hai"
+	description "You have been recommended for your attentiveness for keeping the peace throughout Hai space. Hai officials would like to consult you about possible enlistment."
```
tehhowch: Is it really "attentiveness" if the player only does a single defense mission?

data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
+				accept
+
+mission "Enlisted by Hai [3]"
+	name "Hai Bounty"
+	description "Eliminate the coordinator of the attacks on Hai space."
+	landing
+	to offer
+		has "Enlisted by Hai [2]: offered"
+	source "Hai-home"
+	npc "kill"
```
tehhowch: Drop the quotes here as kill is a reserved token

data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
+	source
+		government "Hai"
+	destination "Hai-home"
+	on offer
+		conversation
+			`As soon as you depart your ship you discover an envoy of sorts awaiting you. "Few human visitors demonstrate care for the lasting peace we have established so long ago," the Hai begins hastily. "You are a prominent example of your kind, and we request audience on <destination>. The unfettered humans have revealed a tentativeness to prey on those who follow a path unlike their own."`
+			choice
+				`	"I suppose I can hear you out."`
+				`	"I would prefer to not get caught up in any trouble."`
+					defer
```
tehhowch: Should probably be decline rather than defer, since the displayed conversation isn't quite "bland" (like the unfettered JD sale convo). If you intend to keep defer, the text should probably reflect that this is a temporary dismissal, e.g. something like, "can it wait? I'm a little busy at the moment."

Edit: I see that the intention is for decline to happen in the next mission, so I think this conversation needs to be made more agreeable for being displayed a handful of times.
Perhaps it also makes sense to include a timer of sorts on defer, so that it is not offered every day. You can use an event, or do some algebra with the day, month, and year conditions.

data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
+	to complete
+		never
+	to fail
+		has "enlisted"
+	source
+		government "Hai"
+	destination "Hai-home"
+	on offer
+		conversation
+			`As soon as you depart your ship you discover an envoy of sorts awaiting you. "Few human visitors demonstrate care for the lasting peace we have established so long ago," the Hai begins hastily. "You are a prominent example of your kind, and we request audience on <destination>. The unfettered humans have revealed a tentativeness to prey on those who follow a path unlike their own."`
```
tehhowch: Any merchant ships will also engage the pirate fleets, so idk about "few human visitors"

data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
+				launch
+	on complete
+		dialog
+			`The government of <planet> pays you <payment> for helping to drive off the attack.`
+		payment 280000
+
+mission "Enlisted by Hai [2]"
+	name "Enlist with the Hai"
+	description "You have been recommended for your attentiveness for keeping the peace throughout Hai space. Hai officials would like to consult you about possible enlistment."
+	priority
```
tehhowch: This is a spaceport only mission, and also marked priority. Off the top of my head, it sounds like this will block any minor missions in hai space until it is accepted. Is that correct?

data/hai missions.txt
```diff
@@ -1061,3 +1061,93 @@ mission "Hai Honeymoon"
...
+	description "Eliminate the coordinator of the attacks on Hai space."
+	landing
+	to offer
+		has "Enlisted by Hai [2]: offered"
+	source "Hai-home"
+	npc "kill"
+		government "Pirate"
+		dialog
+			`You have eliminated the pirate leader's ships. Hopefully this sends a warning to any other pirate warlords who meddles in the affairs of the Hai.`
+		personality "disables" "heroic" "nemesis" "plunders" "staying" "target"
```
tehhowch: Same here, no quotes on single-word tokens that are reserved words (i.e. you can't just throw any ol' text here and have it be used as a personality).

data/hai missions.txt
Darcman99: Having some of the pirates enter the system results in them flying in from other hai systems, which looks weird as why would they not be taken care of in those neighbouring systems?

Darcman99: Having the pirates retreat when they see the battle is lost wouldn't be a bad idea, however what's to stop them from jumping to other hai systems, which doesn't actually seem like so much of a problem so I might just go ahead and remove `staying` anyway.

data/hai missions.txt
Darcman99: I'll lower it to 25% rather than 35%, I don't want to lower it so much so that it becomes easy to miss however.

data/hai missions.txt
beccabunny: You could make the pirates spawn in a different system and make them chase the player, this way they wouldn't come from a random direction but from the right system

data/hai missions.txt
Darcman99: Yeah..  never really digged using `attentiveness` there anyway, could change `You have been recommended for your attentiveness for keeping the peace throughout Hai space.` -> `You have been recommended following your demonstration of peace.`

data/hai missions.txt
Darcman99: Could easily change it back to `decline`, `defer` was used originally just so that the player doesn't get locked out of the shield beetle unlock forever, but I think the mission is noteworthy enough as a *don't throw it away, it sounds important* mission.

data/hai missions.txt
Darcman99: It is, because often-times when `Enlisted by Hai [2]` should show up, damn Elliot shows up midway through and ruins it. That's so that the missions are able to flow easily without the player getting thrown into another mission midway through.

Amazinite: Elliot is a minor mission, so it shouldn't offer over these missions if they're non-minor (and non-priority).

Darcman99: While testing, 2 times in a row I had Elliot offer right in between `[1]` and `[2]`, it's infuriating to have the string just be cut in by another mission like that.

Amazinite:
Closing in favor of #4664.

PR #4652, opened 2019-11-09
Remnant: Bellows-Class Afterburner

Zitchas:
**Content (Artwork / Outfit)**
## Summary
This PR adds the Bellows-Class Afterburner. This is a Remnant afterburner that is currently unavailable, but will be revealed and made available by an upcoming mission string that will available after completing Remnant: Salvage. Due to the size of said string and complexity of certain aspects of it (system and map additions, events, etc), I would prefer to have the afterburner dealt with separately so that the mission PR can focus on the missions proper and the unusual elements therein.

The Bellows-Class Afterburner is intended to be a tier 2 afterburner that is developed over the course of the Remnant missions and is finalized about halfway (~8 missions in) through a larger plotline. It is somewhat smaller and a fair bit more powerful than the two currently in-game. It uses three times as much fuel as the normal afterburner, uses slightly more energy than the ionic afterburner, and provides roughly six times as much thrust. 

## Save File
This save file can be used to experiment with the new afterburner. The pilot has a fleet of four ships: Two Albatrosses and two Ibises (and full complement of drones & fighters). Of each pair, one ship has the afterburner, and one does not, thereby allowing one to compare.
[Jimmy Afterburner.txt](https://github.com/endless-sky/endless-sky/files/3827420/Jimmy.Afterburner.txt)

## PR Checklist
 - [X] I uploaded the necessary image, blend, and texture assets here: [google_drive_zip](https://drive.google.com/open?id=1ytKsYfq3Ngb1naRFV27W48dgNsxbTmwd)
 - [X] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: [Remnant: Bellows-Class Afterburners high-dpi](https://github.com/endless-sky/endless-sky-high-dpi/pull/99)
  
 
remnant afterburner@2x


Zitchas:
One idea I am toying with: Instead of having the Remnant afterburner simply be a "better" afterburner, actually differentiating it by giving it triple or maybe even quadruple the fuel drain, and accompany that with double or triple the thrust. The idea being to create an afterburner that effectively can only be used for short duration; but gives a very solid kick when it is used. So, more like "quick rocket boost" rather than the current afterburners that are more "extra cruise thrust with fuel cost".

petervdmeer:
I like the idea, but I'm also afraid that it might lead to balance problems:
- Wouldn't the triple or quadruple thrust lead to higher sustained speeds?
- If I install 4 "normal" afterburners, would that have the same effect as the Remnant one? (But with 4 times the install space of the Remnant afterburner)?

Zitchas:
I'm not sure of the exact mechanics on that. If we use any afterburner to get to a higher speed (or, for that matter, force effects from explosions or weapons), do we slow down after the force ceases to be applied?

Zitchas:
As a side note: This currently uses the standard afterburner sound effect. If someone wants to make a new one, that'd be great, although not required.

Amazinite:
>Wouldn't the triple or quadruple thrust lead to higher sustained speeds?

Afterburner thrust doesn't affect normal thrusting max speed. Install an afterburner with normal thrusters and you'll see that your max speed doesn't change, but remove all your normal thrusters and your max speed will read as the max speed when using the afterburner. (Sounds like something #4215 could help in by showing both your max speed and max speed + afterburner.)

Zitchas:
Alright, let's try out making them more of a "on-the-spot boost" style afterburner instead of "faster-cruise" style burners. These now guzzle fuel at a prodigious rate.

Changes:
 x3 the thrust
 x4 the fuel consumption
 x2 the energy consumption

Zitchas:
Has anyone tried it out? Anyone like it, or suggest any changes? I'm rather enjoying using it, although it does burn through fuel fast enough to make using it problematic...

data/remnant/remnant outfits.txt
```diff
@@ -361,6 +361,34 @@ outfit "Salvage Scanner"
...
+	"mass" 13
+	"outfit space" -11
+	"engine capacity" -11
+	"afterburner thrust" 153.0
+	"afterburner fuel" 2.0
+	"afterburner energy" 6.0
+	"afterburner heat" 8.0
+	"afterburner effect" "remnant afterburner"
+	description "The Remnant live in an environment where the raw materials for fuel are relatively common on the winds of the nebula, and with ships that incorporate significant built-in ramscoop technology. As such, while they are reluctant to waste fuel, neither are they afraid of using it when needed. Far more important is the ability for their ships to quickly and efficiently enter or exit combat as needed. At the end of the day, it is continued life that is paramount."
+	description "	The bellows-class afterburner embodies this philosophy by providing captains with a tool that allows them to make great leaps of acceleration, typically during evasive manouevers or to catch up with a fleeing target. Due to their enormous fuel usage, it is rare for any ship to use these for more than a few seconds at a time."
```
Arrow2thekn33:
```suggestion
	description "	The bellows-class afterburner embodies this philosophy by providing captains with a tool that allows them to make great leaps of acceleration, typically during evasive maneuvers or to catch up with a fleeing target. Due to their enormous fuel usage, it is rare for any ship to use these for more than a few seconds at a time."
```

data/remnant/remnant outfits.txt
```diff
@@ -361,6 +361,34 @@ outfit "Salvage Scanner"
...
+	thumbnail "outfit/remnant afterburner"
+	"mass" 13
+	"outfit space" -11
+	"engine capacity" -11
+	"afterburner thrust" 153.0
+	"afterburner fuel" 2.0
+	"afterburner energy" 6.0
+	"afterburner heat" 8.0
+	"afterburner effect" "remnant afterburner"
+	description "The Remnant live in an environment where the raw materials for fuel are relatively common on the winds of the nebula, and with ships that incorporate significant built-in ramscoop technology. As such, while they are reluctant to waste fuel, neither are they afraid of using it when needed. Far more important is the ability for their ships to quickly and efficiently enter or exit combat as needed. At the end of the day, it is continued life that is paramount."
```
Amazinite:
```suggestion
	description "The Remnant live in an environment where the raw materials for fuel are relatively common on the winds of the nebula. As such, their ships incorporate significant built-in ramscoop technology, and while they are reluctant to waste fuel, they aren't afraid of using it when needed. Far more important is the ability for their ships to quickly and efficiently enter or exit combat as needed. At the end of the day, it is continued life that is paramount."
```

data/remnant/remnant outfits.txt
```diff
@@ -361,6 +361,34 @@ outfit "Salvage Scanner"
...
+	description "	The bellows-class afterburner embodies this philosophy by providing captains with a tool that allows them to make great leaps of acceleration, typically during evasive maneuvers or to catch up with a fleeing target. Due to their enormous fuel usage, it is rare for any ship to use these for more than a few seconds at a time."
```
Amazinite: Given that this thing guzzles 120 fuel per second of usage, I'd hardly even say that they are rarely used for more than a few seconds; after a few seconds, I'm already out of fuel!
```suggestion
	description "	The Bellows-class afterburner embodies this philosophy by providing captains with a tool that allows them to make great leaps of acceleration, typically during evasive maneuvers or to catch up with a fleeing target. Due to their enormous fuel usage, all but the most specialized of ships have enough fuel to use these for more than a few seconds at a time before completely running out of fuel."
```

tehhowch: "all but the most specialized of ships have enough fuel"? I think you meant "only the most specialized of ships have enough fuel". And then the extra bit is not needed.

Zitchas: Thanks, I actually caught that at the time, and it should already read "none but the most specialized of ships..."

Amazinite: Oops. I had like three different things typed out as to how to change this up and I guess I accidentally smashed them together. 

“All but the most specialized ships run out of fuel in a few seconds” vs “only the most specialized ships can use this for more than a few seconds”.

data/remnant/remnant outfits.txt
```diff
@@ -361,6 +361,34 @@ outfit "Salvage Scanner"
...
+	"mass" 13
+	"outfit space" -11
+	"engine capacity" -11
+	"afterburner thrust" 153.0
+	"afterburner fuel" 2.0
+	"afterburner energy" 6.0
+	"afterburner heat" 8.0
+	"afterburner effect" "remnant afterburner"
+	description "The Remnant live in an environment where the raw materials for fuel are relatively common on the winds of the nebula. As such, their ships incorporate significant built-in ramscoop technology, and while they are reluctant to waste fuel, they aren't afraid of using it when needed. Far more important is the ability for their ships to quickly and efficiently enter or exit combat as needed. At the end of the day, it is continued life that is paramount."
+	description "	The Bellows-class afterburner embodies this philosophy by providing captains with a tool that allows them to make great leaps of acceleration, typically during evasive maneuvers or to catch up with a fleeing target. Due to their enormous fuel usage, none but the most specialized of ships have enough fuel to use these for more than a few seconds at a time before completely running out of fuel."
```
Amazinite: "None but" is an odd wording. Just go with "only".
```suggestion
	description "	The Bellows-class afterburner embodies this philosophy by providing captains with a tool that allows them to make great leaps of acceleration, typically during evasive maneuvers or to catch up with a fleeing target. Due to their enormous fuel usage, only the most specialized of ships have enough fuel to use these for more than a few seconds at a time before completely running out of fuel."
```

Issue #4640, opened 2019-11-04
AI (auto land) gets confused with reverse thrusters


fizzfaldt:
**Describe the bug**
I installed 3 sets of reverse thrusters, such that it's stronger than my forward thrusters.
When I try to land on a planet, it does this weird hopping around.  Instead of just slowing down when reaching the planet (or the old turn around and use forward thrust to slow down) it does some weird hybrid and takes a lot longer to actually land (it starts orbiting the planet almost and firing the forward thrust intermittently).  It seems like it's confused.

**To Reproduce**
Note: I heavy modified the ship I'm flying in, so it's possible what I put in it generated some bugs.  Will attach a save file.
Load this save, leave the planet, go far away.  Then try to land again.

**Expected behavior**
I expect it to choose between reverse thrust or turn+forward thrust and land perfectly.  It shouldn't start orbiting around the planet.

**System (please complete the following information):**
 - OS:
```
$ cat /etc/lsb-release: 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=18.04
DISTRIB_CODENAME=bionic
DISTRIB_DESCRIPTION="Ubuntu 18.04.3 LTS"
$ uname -a:
Linux buk 4.15.0-64-generic #73-Ubuntu SMP Thu Sep 12 13:16:13 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```
 - Game Source: Steam beta
 - Version: 0.9.10 beta

**Link to save file**

**Additional context**
It's possible that the heavily-faked sparrow ship is causing the problem.
[Fighty McFighterson.txt](https://github.com/endless-sky/endless-sky/files/3802658/Fighty.McFighterson.txt)

Amazinite:
Yeah. Reverse engines are undersupported right now. #1583 #1838 #2716

tatami4:
I cannot reproduce this with this save and the ship I modified. I think this should be closed, because it appears to be fixed.
@Amazinite

PR #4655, opened 2019-11-10
Teach AI ships to afterburner during CircleAround

lukearndt:
## Feature Details

I've noticed that escorts use their afterburners when attacking, but not
when escaping from danger.

This commit adds an afterburner condition to the `AI::CircleAround`
routine. As a reminder, `AI::CircleAround` is used in these scenarios:

- An escort has been told to gather around the flagship
- An AI ship has reason to pursue a particular target (eg. cargo scan)

In these situations, the ship will use its afterburner (if it has one)
to approach its target more quickly. To prevent wasteful usage, the ship
will only fire its afterburner if its target is further away than 600.

## UI Screenshots

image

## Usage Examples

- Afterburners can now help escorts escape from dangerous situations
- Government ships can now use afterburners to pursue ships for scanning

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1829,6 +1829,10 @@ void AI::CircleAround(Ship &ship, Command &command, const Ship &target)
 	command.SetTurn(TurnToward(ship, direction));
 	if(ship.Facing().Unit().Dot(direction) >= 0. && direction.Length() > 200.)
 		command |= Command::FORWARD;
+
+	// Use an equipped afterburner if possible
+	if (command.Has(Command::FORWARD) && direction.Length() > 600 && ShouldUseAfterburner(ship))
+		command |= Command::AFTERBURNER;
 }
 
 


Amazinite:
Afterburners and reverse thrusters are both underutilized by the AI, so if you're tackling afterburner usage, do you think you could take a stab at tackling the usage of reverse thrusters as well in a separate PR? (See #4640 and all the issues mentioned there.)

lukearndt:
Sure, I could have a go at that sometime. I recall buying my escorts a bunch of reverse thrusters a while back, and it bugged me that they didn't use them. If they worked properly, would that motivate us to add some bigger ones for larger ships? 🤔 

Anyway - what do you think of this PR?

Zitchas:
Looks like this should be good. I'll see if I can figure out compiling again to test this. :)

Anyway, the four who can merge are: Endless-Sky, Amazinite, Tehhowch, and PointedStick. Amazinite and PointedStick tend to merge story and game content, and Tehhowch is usually the final word on game engine stuff. Endless-Sky is the ultimate authority on everything, but unfortunately isn't available most of the time. 

The general process is: 
1) People go over your PR with a fine-tooth comb and comment on it
2) People test your PR and provide feedback
3) Once one of those 4 with merge authority decide that a) people want this in the game, and b) it works properly and is implemented nicely.
4) Then it gets merged. 

Generally, PRs take anywhere from a day or two (for really simple things, fixes, etc) to a few months (for complicated things). Stuff that is controversial, or has a really big impact on the game can take significantly longer. One of my major mission additions took close to a year to merge. My guess is that this PR could probably be one of the faster ones, since it is more of a fix that fixes a lack that many of us would like to see fixed. But ultimately this will probably depend on how much time Tehhowch has to devote to it.

lukearndt:
Thanks for explaining the process @Zitchas ! I appreciate you giving me a run-down of how things work around here.

I agree that control-scheme option for explicitly telling escorts to afterburner (or not) seems like a nice addition for the future.

source/AI.cpp
```diff
@@ -1829,6 +1829,10 @@ void AI::CircleAround(Ship &ship, Command &command, const Ship &target)
 	command.SetTurn(TurnToward(ship, direction));
 	if(ship.Facing().Unit().Dot(direction) >= 0. && direction.Length() > 200.)
 		command |= Command::FORWARD;
+
+	// Use an equipped afterburner if possible
+	if (command.Has(Command::FORWARD) && direction.Length() > 600 && ShouldUseAfterburner(ship))
```
tehhowch: Are there cases where the ship will have `FORWARD` set but not from the block immediately above? If not, we should probably bundle this check inside that conditional block.

I also think that a more stringent facing alignment with the desired direction of travel could be warranted.

Pointedstick: Ping @lukearndt

lukearndt: My passion for the project ran dry a while ago. You're very welcome to push additional commits to the branch if you'd like to get it to a state where you're happy to merge it. 😄

tehhowch:
Closing in favor of #5352

Issue #4570, opened 2019-10-03
Phrase generation algorithm favors some results over others


Thunderforge:
## Background

A phrase in Endless Sky is a combination of phrases and words. For instance consider the following set of phrases:

```
phrase "civilian name"
	phrase
		"male names"
		"female names"
	word
		" the Great"

phrase "male names"
	word
		"Alexander"
		"Bob"

phrase "female names"
	word
		"Alice"
		"Beatrice"
		"Charlotte"
		"Denise"
```

Thus the phrase "civilian name" will generate `Alexander the Great`, `Bob the Great`, `Alice the Great`, etc.

The [algorithm to generate certain phrases](https://github.com/endless-sky/endless-sky/blob/94346d8905ec516b9a5fad2683f9ca0e5ab5f11f/source/Phrase.cpp#L68) works like this:

1. Take an input phrase (which will contain subsequent phrases or words)
1. Create a new string `result`
1. For each phrase or word within the phrase:
    - If it's a phrase, pick a random value, go to 1. with that phrase value as input, and append the result returned to `result`
    - If it's a word, pick a random value and append it to `result`
1. Return `result`

## Problem

The problem is that by having each step be a discrete random event, the outcomes are uneven. Consider that in the above example, there is a 1/2 chance for either "male names" or "female names" to be selected. Also, there is a 1/2 change for either "Alexander" or "Bob" to be selected. But there is a 1/4 chance for any of "Alice", "Beatrice", "Charlotte", or "Denise" to be selected. Thus the odds of a particular name to be chosen would be:

- Alexander the Great: 1/2 * 1/2 = 1/4
- Bob the Great: 1/2 * 1/2 = 1/4
- Alice the Great: 1/2 * 1/4 = 1/8
- Beatrice the Great: 1/2 * 1/4 = 1/8
- Charlotte the Great: 1/2 * 1/4 = 1/8
- Denise the Great: 1/2 * 1/4 = 1/8

Or to put it another way, if you parsed the phrase "civilian" eight times, the expected probability outcome would be some combination of:

- Alexander the Great
- Alexander the Great
- Bob the Great
- Bob the Great
- Alice the Great
- Beatrice the Great
- Charlotte the Great
- Denise the Great

Thus `Alexander the Great` and `Bob the Great` appear more often than any other names.

[...]

PR #4657, opened 2019-11-11
More pirate names

Thunderforge:
A major goal was to add more female and non-western pirates, as well as make it easier to determine (for the interest of anybody looking at the list) whether the pirate names were real or fictional. I realized a lot of the ship names previously listed were ghost ships, so I moved them into their own category (perhaps if derelict ships are implemented, those names can be repurposed for that).

I decided not to use any historical names newer than the last hundred years or so out of sensitivity to those affected by modern piracy.


data/names.txt
```diff
@@ -3518,6 +3448,165 @@ phrase "pirate"
...
+		"Sultana"
+		"Ranger"
+		"Lark"
+		"Postillion"
+		"Good Fortune"
+		"Royal Fortune"
+		"Marianne"
+		"Amity"
+		"Pembroke"
+		"Margaret"
```
Thunderforge: Some of these single word pirate ships were on the old list of pirate names that I moved down to this list. I would be okay with removing some of the less memorable ship names if we like.

data/names.txt
```diff
@@ -3069,80 +3069,16 @@ phrase "bad outcomes"
...
-		"Ranger"
-		"Lark"
-		"Flying Dragon"
-		"Golden Fleece"
-		"Trompeuse"
-		"La Nouvelle Trompeuse"
-		"Sea Nymph"
-		"Rising Sun"
-		"Postillion"
 		"Indian Queen"
```
Thunderforge: It's about this point that my Google-fu failed me and I couldn't tell if these were historical pirate ships or names just made up that sounded "piratey". I've left it as-is.

data/names.txt
```diff
@@ -3518,6 +3448,165 @@ phrase "pirate"
...
+		"Lark"
+		"Postillion"
+		"Good Fortune"
+		"Royal Fortune"
+		"Marianne"
+		"Amity"
+		"Pembroke"
+		"Margaret"
+
+phrase "historical ghost ships"
```
Thunderforge: I know the phrase "historical" is a bit strange, especially for stuff like the Flying Dutchman, but I guess I was referring to real-world.

data/names.txt
```diff
@@ -3069,80 +3069,16 @@ phrase "bad outcomes"
...
-		"Resolute"
-		"Mary Celeste"
-		"Resolven"
-		"Zebrina"
-		"Deering"
-		"Bavchimo"
-		"Governor Parr"
-		"Joyita"
-		"Teignmouth Electron"
-		"Ocean Wave"
```
Thunderforge: The ghost ship names appear to have been taken from [this Wikipedia page](https://en.wikipedia.org/wiki/Ghost_ship#Historically_attested). The *Ocean Wave* was from 1975, which was a bit too modern for my tastes, so it and subsequent ships were not moved to the new "historical ghost ships" category.

data/names.txt
```diff
@@ -3518,6 +3448,166 @@ phrase "pirate"
...
+		"Philip Lyne"
+		"Kanhoji Angre"
+
+phrase "historical female pirates"
+	word
+		"Lady Killigrew"
+		"Ching Shih"
+		"Anne Bonny"
+		"Mary Read"
+		"Johanna Hård"
```
Amazinite: The `å` character and any other accented/non-standard to English characters (I've also spotted a use of `ø`) won't display in game without #4123.

Thunderforge: The non-ASCII names in this PR are:

 - [Johanna Hård](https://en.wikipedia.org/wiki/Johanna_Hård)
 - [København](https://en.wikipedia.org/wiki/København_(ship))
 - [Bêlit](https://en.wikipedia.org/wiki/Bêlit_(Robert_E._Howard))

My preference is to not merge this PR until #4123 is resolved (rather than ASCII-izing the names to "Johanna Hard:, etc), since I think the 30th Century in the game should reflect a variety of cultures, and that we've finally converted from ASCII to Unicode.

tehhowch: Perhaps remove them and put them in a follow-up PR, then?

jafdy:
This feels like splitting one list into lots of lists, and does beg the question why.

tehhowch:
@jafdy my thought is that it is perhaps intended to improve the phrase diversity (i.e. https://github.com/endless-sky/endless-sky/issues/4570 ).

jafdy:
If so that is not made clear in the OP.

Thunderforge:
My thought for adding more lists was to make it easier to maintain. For instance, as I was going through the current pirate names, and I had no idea why "Demeter" was on the list. I knew that she was [the Greek goddess of agriculture](https://en.wikipedia.org/wiki/Demeter), but how is that a pirate name? Only after further research did I realize that it was a reference to Bram Stoker's _Dracula_ as the ill-fated ship he used to cross to England (and I'd read that book, shame on me for forgetting that).

After some thinking, I think that I can instead add some comments to add these things and make it easier to maintain (i.e. save a bunch of Wikipedia lookups), instead of breaking apart into lists. That way, we get the benefits I was hoping for while also preventing separate lists.

Reluctantly, I'll remove the non-ASCII names so that this PR won't be held up any longer.

Thunderforge:
I'd appreciate some clarification: is the problem that I'm creating new `phrase`s when they might instead just be `phrase "pirate"`, or the problem that the lists are too short (they are currently each 12-20 names)? 

Would it be okay to do this?

```
#Historical male pirates
phrase "pirate"
	word
		"Blackbeard"
                "William Kidd"
                […]

#Historical female pirates
phrase "pirate"
	word
		"Lady Killigrew"
		"Ching Shih"
                […]
```

Or would it be better to do this?

```
phrase "pirate"
	word
                #Historical male pirates
		"Blackbeard"
                "William Kidd"
                […]
                #Historical female pirates
		"Lady Killigrew"
		"Ching Shih"
                […]
```

Thunderforge:
@Pointedstick Please review my changes to this. Since I didn't get any feedback on which variation I should, use, I picked the one that seemed better to me.

Amazinite:
I think this is fine as is with multiple lists. Functionally it'd be no different than having a single list with comments denoting the different categories, outside of the issue raised in #4570, but that's not too bad, and since the lists are all right next to each other anyways it isn't as if it becomes more difficult for someone to read the file to understand where names are being pulled from.

data/names.txt
```diff
@@ -3518,6 +3448,161 @@ phrase "pirate"
 		"Fate"
 		"Boys"
 
+#Historical male pirates
```
Amazinite: I'd just suggest adding a space after the comment mark for easier reading, here and everywhere else.
```suggestion
# Historical male pirates
```

PR #4662, opened 2019-11-11
Remnant Peregrine

Zitchas:
**Content (Artwork / Ship)**
## Summary
This PR adds the peregrine, as part of the foundation for the next mission series for the Remnant. Like the afterburner (which will be revealed about halfway through the mission set), the Peregrine will be revealed near the end of the mission series. This mission set will require the Salvage and Void Sprite sample return missions to be complete. The Peregrine is being PR'd in advance of the actual mission set so that it can be worked on separately from the mission itself. The mission set is going to be a longer set, so I'm breaking the PR into pieces in order to be more manageable and keep conversations focused.

The peregrine is a tier 2 fast attack cruiser. This is the Remnant making good on their attraction to always being able to run taken to its logical extreme: A powerful medium warship that - like its namesake - is capable of outrunning virtually anything in the game. In metagaming terms, this is the Remnant equivalent to the engine variants of the Marauders, providing a tier 2 ship for those players who preferred those engine variants over any of the other ships available in human space.

Compared to the Albatross, which is the ship the player is most likely to be flying when the Peregrine is unlocked, the Peregrine is slightly weaker in terms of shields and hull, but regenerates faster. It has vastly more speed and maneuverability, even more so with the afterburners taken into account, but has fewer weapons slots. In short, it is much better for chasing down enemies, but worse for having ship-to-ship slugging matches. The Peregrine also has less cargo space. As such, the Peregrine is unlikely to be a "This is absolutely a better ship" than the Albatross, but is going to be a very tempting upgrade for a variety of reasons.

In terms of larger scale balancing, this is intended to sit in between the Strong Wind and the Tempest. Stronger than the Strong Wind, but weaker than the Tempest.

Note: This ship will not be visible in the Remnant shipyard nor available for purchase until after appropriate point in the missions. This is a *new* ship, developed almost from the ground up by Taely and her team, and incorporates at several centuries of advancement since the Starling and Albatross were originally designed. And yes, it does have a small Korath heat shunt in its outfits. They have access to them, might as well use them. 

## Save File
This save file can be used to play the new ship content:
[Prima Peregrine.txt](https://github.com/endless-sky/endless-sky/files/3831964/Prima.Peregrine.txt)

## PR Checklist
 - [X] I uploaded the necessary image, blend, and texture assets here: [Google drive zip file](https://drive.google.com/open?id=1YBttI-2b9DMSDBoZYs2hqCZR7WPUy171)
 - [X] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the [`@2x` versions of these art assets.](https://github.com/endless-sky/endless-sky-high-dpi/pull/130)


Arachi-Lover:
description mentions still that "the cloak is built into the Gull's hull"

that aside, I feel like the HP is a bit excessive for a medium warship

data/remnant ships.txt
```diff
@@ -576,6 +576,90 @@ ship "Penguin"
...
+	leak "leak" 50 50
+	leak "flame" 50 80
+	leak "big leak" 90 30
+	explode "tiny explosion" 10
+	explode "small explosion" 25
+	explode "medium explosion" 25
+	explode "large explosion" 10
+	"final explode" "final explosion medium"
+	description "The Peregrine is the first warship designed by Taely using fresh insights garnered from recent experimentation as well as renewed studies of the void sprites. Encompassing over a century of development since the last generation of warships was built, the Peregrine is the forerunner of a new breed of Remnant ships."
+	description "	The peregrine was designed along the lines of old-earth's fast-attack cruisers. Drawing on inspiration from its namesake, famed for being the fastest self-propelled lifeform on Earth, the Peregrine is designed to strike hard and fast. It closes quickly with enemies to unleash devastating barrages in the opening seconds of combat."
```
MCOfficer: 4th `The peregrine` in 3 sentences

Zitchas:
OK. Making some changes. As part of the context, this is supposed to be a jump in tech level. Something kind of along the same lines as the Wanderers' step up from the "Strong Wind" to the "Tempest."

Also, I am revising my story a bit, and the Peregrine is going to get pushed a little farther. Or at least, it is going to feel like it. It'll actually be roughly the same point in the primary storyline, but it is going to require some side-story tie-ins to unlock. Side-stories that require doing some additional research on the Quarg. (a follow up story to Dawn's research on them) as well as the Hai.

data/remnant ships.txt

MCOfficer:
@Zitchas are there assets for the peregrine thumbnail?

PR #4663, opened 2019-11-11
Boosting the Strong Wind

Zitchas:
The Strong Wind is described as a science vessel that has been retrofitted for combat. Given the Wanderer's focus on terraforming activities, it makes sense to me that their science vessel would have very powerful sensors related to that.

This PR adds the following to the Strong Wind:
- "asteroid scan power" 100
- "atmosphere scan" 100
- "tactical scan power" 20

This could result in the Strong Wind being an interesting candidate for a late-game mining ship for players so inclined. This change shouldn't have much impact on combat, but will help the Strong Wind feel a bit more unique than "just another warship."

### DIFF
--- a/data/wanderer ships.txt
+++ b/data/wanderer ships.txt
@@ -213,7 +213,7 @@ ship "Strong Wind"
 		"heat dissipation" .7
 		"fuel capacity" 600
 		"cargo space" 68
-		"outfit space" 496
+		"outfit space" 493
 		"weapon capacity" 198
 		"engine capacity" 114
 		"asteroid scan power" 100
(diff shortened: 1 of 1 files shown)


data/wanderer ships.txt
```diff
@@ -211,11 +211,14 @@ ship "Strong Wind"
 		"mass" 260
 		"drag" 4.7
 		"heat dissipation" .7
-		"fuel capacity" 400
+		"fuel capacity" 600
 		"cargo space" 68
-		"outfit space" 493
+		"outfit space" 496
```
Amazinite: Remove a thunderhead storage array to add another ramscoop instead of increasing the outfit space.

Zitchas: Done.

PR #4670, opened 2019-11-13
Consider more factors when calculating crew salary

lukearndt:
## Feature Details

We can now calculate crew salaries using more factors than just how many
crew members the player has in their active ships. This will allow us to
create various different kinds of crew, each paid different amounts.

This goal of this system is to allow for interesting trade-offs in fleet
design. It also lets us create anti-snowballing mechanics, which
help prevent the player from suddenly catapulting upward in power.

Previously, salaries were constrained linearly: 100 credits per crew
member across the entire fleet. Since income potential grows at a
non-linear rate, linearly-scaling salaries soon become irrelevant.

Using this tool, we can make salaries increase in proportion to actual
power gain. This should hopefully add interesting decisions for the
player - do they really need all those heavy warships, or will they cost
too much to maintain?

### Context:

Fleet management is currently very simple from an economic perspective.
Crew members cost a flat 100 credits per day, regardless of what they
actually do on their ship. Salary costs can be mostly ignored because
the ships' benefits to the player far outweigh them.

### Changes:

This commit makes crew salaries more complex by introducing the ability
for us to define various kinds of crew members in the data files. The
system currently supports the following attributes, with default values:

- name
- "avoids escorts" (false)
- "avoids flagship" (false)
- "parked salary" (0)
- "place at" (empty list)
- salary (100)
- "ship population per member" (0)

The commit includes a `crew.txt` data file with a few commented-out
examples, plus an active data node for the default crew members.

Since we also display the salaries on the Hiring panel, this commit
changes that code to call the appropriate functions on the `Crew` class.

Merging this PR will not meaningfully affect gameplay. It just gives us
some tools that we can use for later changes.

### Considerations:

I considered adding this logic to the `PlayerInfo` class, since that is
where crew salaries used to be calculated. However, `PlayerInfo` is a
really huge class with a ton of responsibilities. When editing it, I
often got lost if I moved my editor window too far. Instead, I added a
new class called `Crew`.

Ideally it would be nice to update the fleet panel so that players can
see how many of each crew member on a given ship and how much salary
they are paid each day. This would help with decision making. We don't
need to do that yet, though.

### Review Note:

It's worth noting that the first version of this pull request was a lot
simpler. It introduced a bunch of new kinds of crew members, but it was
not configurable using data files. Knowing that will help you make sense
of the discussion.


Zitchas:
Parked ships are assumed to get crew from the same place that the player does when they hit the "hire" button: Apparently the economics of the galaxy are such that on any inhabited world, being paid 100/day is a good enough wage that there's *always* people qualified for the job just waiting to leap at it. Although it is probably better to think of them like independent truckers rather than employees: They only get paid so long as there is work for them to do. When there isn't, they're no longer working for you.

I like the idea for this, but the values are much too big. A thousand credits per day for the captain means that if the player happens upon a disable pirate interceptor and captures it (fairly easy to do for an early game player) they suddenly have a significant spike in their daily costs. Given that the early game is pretty much the *only* time when crew costs are actually a limiting factor, I think the the numbers need to be re-balanced so that they don't impact the early game as much. 

In a paradigm like this one, I'd suggest something more like: Commanders/captains get 300, and the officers get 200; and the ratio of officers to crew is at least 1:10, but probably closer to 1:20. (with commanders counting as officers so basically ships under 21 are 1 commander/player and up to 20 crew)

Given the way crew scales, this would reduce most of the impact that this would have on the early game and players with very small fleets, but would add significantly to the costs of flying large ships with large numbers of crew and/or large fleets.

For an example, the numbers for having a stock Bactrian in your fleet: (70 min crew. I'm assuming that the commander and officers are part of that total, and don't count themselves)
Currently, having one in your fleet costs: 7000/day
Under your proposal: 1 commander + 10 officers + 59 crew = 1000+2500+5900=9400/day
My suggested numbers: 1 commander + 5 officers + 64 crew = 300+1000+6400=7700/day

For the bactrian, a significant change, but not insurmountable.

An example with a marauder arrow, probably a very likely ship for the player to capture early in the game: (2 crew)
Currently, having one in your fleet costs: 200/day
Under your proposal: 1200/day
My suggested numbers: 400/day

I suspect that most players - especially ones for whom capturing interceptors is a new and/or challenging thing - will be of the opinion that having a marauder arrow join your fleet is *not* worth 1200/day. Even more to the point, is having a shuttle or a star barge following you around worth 1000/day? Given their limited capacity to add to one's profit margin, that's *really* going to eat into the profits. This is fine if one wants to extend the "desperate for money and scratching the bottom of the barrel for credits" feel of the game, but I don't think it is particularly fun.

As such, I suspect that the net result of this PR will be that interceptor and other small ships will go from "ships early players get to cut their teeth on expanding their fleets with" to "just blow these out of the sky because they are too expensive to keep" and will result in players simply avoiding using any of the small ships in their fleets. 

It will also make the current usability of fighters even worse. Most people find fighters too expensive to use given their short lifespan. Tack an additional 900/day onto the cost of having them, and people may not even want to have boxwings anymore.

I also think that the long-term dynamic of this will be even more emphasis on condensing fleets into a few big ships and less small ships. Given that this is already the metadynamic for the game, I don't think adding additional costs to penalize the use of small ships and give preference to larger ones benefits anything, and in fact may actually make the game less varied and less fun.

Zitchas:
I would suggest that in addition to the specific wages, the ratio of officers to crew itself should also be a datapoint accessible to plug-in creators.

I would like to suggest that another variable that might have more of a beneficial impact might be "Admiral." Which would be a position that replaces one of your fleet's captains, and your fleet has 1 admiral for every X number of ships. (I would envision something to the effect of 1 per 10 HD/JD capable ships).

I like the idea of the admiral because it adds extra costs for large fleets without adding any costs to small fleets. As defined by the "only counts HD/JD capable ships," it also wouldn't negatively impact the use of fighters.

An alternative might be "only counts ships above a certain mass. Say, ships with a hull-mass greater than 100. (this number would also be a variable pulled from the datafile)

So, this would make for a "Crew Economics" datafile that would have something along the following:
- "Crew Wages"
- "Officer Wages"
- "Captain Wages"
- "Admiral Wages"
- "Officer Ratio"
- "Admiral Ratio"
- "Hull Mass Minimum"
- "HD/JD capable" (binary 0/1)

Listing both the hull mass and the HD/JD capable as options means that either could be used, and plugin creators could experiment using either or both of those as limiting factors. And setting both to 0 would result in *all* ships counting.

Defaults would be (as per Tehhowch's suggestion) 100 for all postions, and the ratios could be set to 0 as signifiying that they don't actively apply.

source/Crew.cpp
```diff
@@ -0,0 +1,223 @@
...
+	if(ship->IsDestroyed()) return 0;
+	
+	const Set<Crew> crews = GameData::Crews();
+	const Crew *defaultCrew = crews.Find("default");
+
+	int64_t salariesForShip = 0;
+	int64_t specialCrewMembers = 0;
+
+	// Add up the salaries for all of the special crew members
+	for(const pair<const string, Crew> crewPair : crews)
```
lukearndt: XCode gives me a warning here:

> Loop variable 'crewPair' of type 'const pair<const std::__1::string, Crew>' (aka 'const pair<const basic_string<char, char_traits<char>, allocator<char> >, Crew>') creates a copy from type 'const pair<const std::__1::string, Crew>'

It suggests the following fix:

> 1. Use reference type 'const pair<const std::__1::string, Crew> &' (aka 'const pair<const basic_string<char, char_traits<char>, allocator<char> >, Crew> &') to prevent copying

Implementing the suggestion completely breaks the feature, though. With the change, the game doesn't count any non-default crew members.

Any ideas?

Zitchas:
I can't vouch for the code, but having a class like that which can be used to setup any number crew ranks with differing pay structures could work very well.

Two questions for you:
1) How exactly does this determine how many of a particular crew are on each ship?

2) How easily extensible is this? For example, based on past suggestions, I suspect that some people are immediately going to want to add "Space Marine" or "Security Officer" as a crew type, with some kind of benefit in regards to boarding actions. (not necessarily as vanilla content, but definitely as plug-in stuff)

lukearndt:
@zitchas

1. Crew members are calculated on a per-ship basis.

The system first checks if the crew member can appear on the ship at all, based on whether it's the flagship or an escort. Then it applies the “minimum per ship” rule. Finally, it divides the ship’s crew count by “population per occurrence”.

For example, say we have a crew member called “junior officer” with a “population per occurrence” of 5. The 5th crew member would be a junior officer, then the 10th, 15th, 20th, etc.

If we add supply a “minimum per ship” of 1, the 1st crew member would be a junior officer, then the 10th, 15th, 20th, etc.

If we don’t supply a “population per occurrence”, it will only add crew members equal to “minimum per ship”.


2. The code should be quite extensible. I didn’t make crew members affect the performance of the ships, since that goes beyond the scope of this particular change, but it should be possible to do at a later stage.

Adding attributes that affect ship or boarding performance will require additional code changes, of course.

jafdy:
This feels a lot like increasing crew costs to 180 (excluding the pilot thing).

source/Crew.cpp
```diff
@@ -0,0 +1,232 @@
...
+	const Crew *defaultCrew = crews.Find("default");
+	
+	int64_t salariesForShip = 0;
+	int64_t specialCrewMembers = 0;
+	
+	// Add up the salaries for all of the special crew members
+	for(const pair<const string, Crew>& crewPair : crews)
+	{
+		// Skip the default crew members.
+		if(crewPair.first == "default")
```
tehhowch: Is there a way to avoid needing to treat one member of the collection differently? Perhaps GameData holds a crew class that has the default type's values, and then that class has a collection of the other crewTypes?

Then you'd be able to have each member of the collection be handled in the same way, and the remaining needed crew are handled by the default type config.

source/Crew.h
```diff
@@ -0,0 +1,68 @@
...
+	static int64_t SalariesForShip(
+		const std::shared_ptr<Ship> ship,
+		const bool isFlagship,
+		const bool includeExtras = true
+	);
+
+	// Load a definition for a crew economics setting.
+	void Load(const DataNode &node);
+	
+	const bool &AvoidsEscorts() const;
```
tehhowch: These const getters should just return bool / int by value, rather than const ref.

source/Crew.h
```diff
@@ -0,0 +1,68 @@
...
+#ifndef CREW_H_
+#define CREW_H_
+
+#include "Ship.h"
+
+class Crew
+{
+public:
+	// Calculate one day's salaries for the Player's fleet
+	static int64_t CalculateSalaries(
```
tehhowch: Almost all of these class methods' arguments should be by const ref to avoid copies. Primitives can be just bool by value.
Generally the declarations are single-line, as well.

source/Crew.cpp
```diff
@@ -0,0 +1,223 @@
...
+	const bool includeExtras
+) {
+	int64_t totalSalaries = 0;
+
+	for(const shared_ptr<Ship> &ship : ships)
+	{
+		totalSalaries += Crew::SalariesForShip(
+			ship,
+			// Determine whether if the current ship is the flagship
+			ship->Name() == ships.front()->Name(),
```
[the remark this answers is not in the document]
lukearndt: What's the correct way to do it?

source/Crew.cpp
```diff
@@ -0,0 +1,223 @@
...
+			else if(child.Token(0) == "on flagship?")
+				isOnFlagship = child.Value(1);
+			else if(child.Token(0) == "pay salary while parked?")
+				isPaidSalaryWhileParked = child.Value(1);
+			else if(child.Token(0) == "daily salary")
+				dailySalary = child.Value(1);
+			else if(child.Token(0) == "minimum per ship")
+				minimumPerShip = child.Value(1);
+			else if(child.Token(0) == "population per occurrence")
+				populationPerOccurrence = child.Value(1);
```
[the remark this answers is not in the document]
lukearndt: I don't think they would make any sense. Good question; it might be worth validating that it's positive.

source/Crew.cpp
```diff
@@ -0,0 +1,223 @@
...
+
+int64_t Crew::SalariesForShip(
+	const shared_ptr<Ship> ship,
+	const bool isFlagship,
+	const bool includeExtras
+) {
+	// We don't need to pay dead people.
+	if(ship->IsDestroyed()) return 0;
+	
+	const Set<Crew> crews = GameData::Crews();
```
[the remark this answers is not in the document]
lukearndt: This was for convenience so that the code didn't have to call `GameData::Crews()` twice, but I see now that it's unnecessary.

source/Crew.cpp
lukearndt: The reason they're handled differently is that there isn't a straight-forward way to figure out how many there are on the ship. They're the crew type that fills in the blanks after we've figured out all the special roles, so we have to find all the special roles first.

I wanted us to be able to use the data files to specify the behaviour for the default crew as well. Would your suggestion get us that feature, or would they be defined at the code level?

data/crew.txt
```diff
@@ -0,0 +1,29 @@
...
+# crew "pilot"
+#   name "Pilot"
+#   "daily salary" 250
+#   "minimum per ship" 1
+#   "on flagship?" 0
+# 
+# crew "junior officer"
+#   name "Junior Officers"
+#   "daily salary" 500
+#   "population per occurrence" 5
```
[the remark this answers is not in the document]
lukearndt: I agree that the name isn't great, but I don't find `fraction` an improvement. There's got to be a better term lurking somewhere at the edge of our minds. Any other ideas?

data/crew.txt
```diff
@@ -0,0 +1,29 @@
...
+# terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later version.
+#
+# Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+crew "default"
+  name "Regulars"
+  "daily salary" 100
```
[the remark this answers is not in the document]
lukearndt: The rest of the game calls them salaries, so I opted to be consistent. We could potentially drop the `daily`, though?

source/Crew.cpp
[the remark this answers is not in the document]
tehhowch: `PlayerInfo::Flagship()`, for the player's ships. For a group of NPC ships, walk up the parent chain until a ship either has no parent, or that parent is a player ship.

source/Crew.cpp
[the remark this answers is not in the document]
tehhowch: I meant, why create a copy of `Set<Crew>` instead of using a reference?

source/Crew.cpp
tehhowch: My suggestion is a bit different design - only one `CrewEconomy` class for GameData to manage, and the datafile would look something like
```
"crew economy"
    salary 100
    type "Pilot"
        salary 125
        "escort only"
        minimum 1
    type "Officer"
        salary 110
        "crew fraction" .1
    type "XO"
        salary 175
        "always paid"
    ...
```

Note that one could easily enter a scenario where a ship cannot satisfy the minimum values of all crewTypes due to lack of bunks.

Zitchas:
@jafdy If I'm understanding things correctly, this will establishes the framework wherein we could implement varying pay structures for different ranks of crew members, however it doesn't change the default arrangement. So, even after this is merged there'll still just be the player and X standard crew at 100/day. So this being implemented will largely be an invisible change that the plugin creators can use to experiment with various pay scales. 

Once we have the mechanics sorted out, I imagine we'll probably have a new discussion about what the default values should be; but this particular change is supposed to be flexible enough to handle whatever sort of crew ranking scheme we come up with.

jafdy:
I should have clarified, this is in reference to your example. Paying every 5th crew member an extra 400 credits per day is a lot like just paying the whole lot at 180 credits per day, but with extra stuff that now needs explaining. Any other complexity that you then add will largely speaking have the same effect, the only exception I can think of is a one per ship kind of cost, but that has it's own issues as already discussed.

Zitchas:
Thanks for clarifying. Yes, all the 1 per so many does mathematically break down to being the same thing as just increasing the wages for everyone a certain amount. 

Where it would get more interesting is if we could break it down more. For instance, if warships had a 1:5 ratio of officers to crew while a freighters have 1:10. Perhaps Korath ships, with their focus on automation need less direct oversight, and thus have 1:15 on warships and 1:25 on freighters.

source/Crew.cpp
lukearndt: Interesting idea. How well would this play with plugins overriding the crew definitions? Consider the following scenario:

1. Base game data defines a set of crews.

2. Plugin wants to:
	a. Define one additional type of crew
	b. Override one of the base game's crews

The plugin doesn't want to have to redefine the whole tree because they want to allow the base game to keep updating it.

What would the plugin put in their `plugin/crew.txt`?

lukearndt:
@jafdy I see what you're saying. I disagree on the basis that different ships end up having entirely different cost scales. Consider that not every ship will have enough crew members to reach certain breakpoints.

To use the current command structure example, there are Junior Officers every 5 crew members and Senior Officers every 20 crew members.

This means Fighters and Interceptors will never need officers.
It also means that Senior Officers are largely just for big ships.

The player might decide to avoid these costly crew members by buying a lot of smaller ships instead of a few big ones. Or they might decide to avoid the replacement costs (small ships die more often) and pay crew instead.

In this manner, the system introduces decisions for the player.

data/crew.txt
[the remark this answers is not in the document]
lukearndt: I went with "population per member".

I don't want to define this variable as a decimal; having it as an integer makes the underlying mathematics a lot simpler, and not susceptible to floating point rounding issues.

source/Crew.h
lukearndt: Could you give me an example of how to do this? GitHub suggestions are really great for this. You can make a suggestion using `ctrl+g` (or command+g if you're on MacOS).

source/Crew.cpp
```diff
@@ -0,0 +1,240 @@
...
+	populationPerMember = 0;
+	
+	for(const DataNode &child : node)
+	{
+		if(child.Size() >= 2)
+		{
+			if(child.Token(0) == "name")
+				name = child.Token(1);
+			else if(child.Token(0) == "salary")
+				if(child.Value(1) >= 0)
```
tehhowch: Just use `max` instead of branching. You can add a code comment above the declaration of `salary` e.g. `// The number of credits paid daily (minimum 0).`

source/Crew.cpp
tehhowch: You would implement `add / remove` syntax for collection variables.

source/Crew.h
tehhowch: `foo(const T t)` takes variable t of type `const T` by value, and must copy it.
`foo(T &t)` takes variable t of type `T` by non-const reference.
`foo(const T *t)` takes a pointer t by value that dereferences to an object of type `const T`

source/Crew.h
```diff
@@ -0,0 +1,70 @@
...
+	bool AvoidsEscorts() const;
+	bool AvoidsFlagship() const;
+	bool IsPaidSalaryWhileParked() const;
+	int64_t MinimumPerShip() const;
+	int64_t PopulationPerMember() const;
+	int64_t Salary() const;
+	std::string Name() const;
+
+private:
+	bool avoidsEscorts;
```
tehhowch: These should all be initialized with the desired defaults, e.g. `int64_t salary = 100`

source/PlayerInfo.cpp
```diff
@@ -729,6 +716,10 @@ Ship *PlayerInfo::Flagship()
 }
 
 
+const shared_ptr<Ship> &PlayerInfo::FlagshipPtr() const
```
tehhowch: Nope, remove this. Use `PlayerInfo::Flagship` for const access to the player's flagship.

source/Crew.h
lukearndt: Oh, good point. Then I don't need to do it in the `.cpp` file!

source/Crew.h
lukearndt: Okay, I did some reading and I think I'm passing by const ref now.

Is `foo(const T &t)` what you're after?

source/Crew.cpp
lukearndt: I gave this some thought and had a look at the `add / remove` examples throughout the codebase.

Honestly, the single-node solution seems needlessly complex for what we're trying to achieve here. I'd prefer to stick with one node per crew type.

source/PlayerInfo.cpp
```diff
@@ -668,21 +669,8 @@ Account &PlayerInfo::Accounts()
...
-	for(const shared_ptr<Ship> &ship : ships)
-		if(!ship->IsParked() && !ship->IsDestroyed())
-			crew += ship->RequiredCrew();
-	if(!crew)
-		return 0;
-	
-	// Every crew member except the player receives 100 credits per day.
-	return 100 * (crew - 1);
+	const Ship * playerFlagship = Flagship(); 
+	return Crew::CalculateSalaries(ships, playerFlagship);
```
tehhowch: Just use `Flagship()` here

source/Crew.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "Ship.h"
+
+class Crew
+{
+public:
+	// Calculate one day's salaries for the Player's fleet
+	static int64_t CalculateSalaries(
+		const std::vector<std::shared_ptr<Ship>> &ships,
+		const Ship * &flagship,
```
tehhowch: This should not be a reference.

Also, you need to match the style of other headers and have a single line declaration:
```c++
static int64_t CalculateSalaries(const std::vector<std::shared_ptr<Ship>> &ships, const Ship *flagship, bool includeExtras = true);
```
The style guide's suggestion of 80 columns is rarely adhered to, especially when it comes to header declarations.

source/Crew.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+	// Set the id of this crew member so we know that we have loaded it.
+	if(node.Size() >= 2)
+		id = node.Token(1);
+	
+	for(const DataNode &child : node)
+	{
+		if(child.Size() >= 2)
+		{
+			// The crew members' display name (plural, Title Case)
```
tehhowch: These comments should go above the variables' declarations (i.e. in the header file).

source/Crew.cpp
```diff
@@ -0,0 +1,242 @@
...
+	const bool includeExtras
+)
+{
+	int64_t totalSalaries = 0;
+	
+	for(const shared_ptr<Ship> &ship : ships)
+	{
+		totalSalaries += Crew::SalariesForShip(
+			ship,
+			ship.get() == flagship,
```
tehhowch: It's a bit micro-optimization, but you could set a bool flag when you find the flagship so that you can skip this pointer dereference + comparison check afterwards.

source/Crew.cpp
```diff
@@ -0,0 +1,242 @@
...
+	// If this is an escort, check if this crew avoids escorts.
+	if(!isFlagship && crew.AvoidsEscorts())
+		return count;
+	
+	const int64_t countableCrewMembers = includeExtras
+		? ship->Crew()
+		: ship->RequiredCrew();
+	
+	// Apply the per-ship minimum.
+	 count = min(crew.MinimumPerShip(), countableCrewMembers);
```
tehhowch: Is this aligned right?

source/Crew.cpp
```diff
@@ -0,0 +1,242 @@
...
+{
+	// We don't need to pay dead people.
+	if(ship->IsDestroyed())
+		return 0;
+	
+	int64_t salariesForShip = 0;
+	int64_t specialCrewMembers = 0;
+	
+	// Add up the salaries for all of the special crew members
+	for(const pair<const string, Crew>& crewPair : GameData::Crews())
```
tehhowch: It is this repo's style to attach ointer or reference signifier to the variable and not the type, e.g. `int *i` and not `int* i`

source/Crew.cpp
```diff
@@ -0,0 +1,242 @@
...
+	// Figure out how many regular crew members are left over
+	int64_t defaultCrewMembers = (
+		includeExtras 
+			? ship->Crew()
+			: ship->RequiredCrew()
+		) - specialCrewMembers
+		// If this is the flagship, one of the crew members is the Captain
+		- isFlagship;
+
+	const Crew *defaultCrew = GameData::Crews().Find("default");
```
tehhowch: You need to nullcheck this variable.

source/HiringPanel.cpp
```diff
@@ -12,12 +12,14 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
 #include "HiringPanel.h"
 
+#include "Crew.h"
 #include "FillShader.h"
 #include "GameData.h"
 #include "Information.h"
 #include "Interface.h"
 #include "PlayerInfo.h"
 #include "Ship.h"
+#include "System.h"
```
tehhowch: Is this still needed?

source/PlayerInfo.cpp
lukearndt: We can't because the compiler complains that the argument has to be an `lvalue`.

tehhowch: That's because your function is still passing it as a reference instead of by value.

source/Crew.cpp
lukearndt: Yes, that would help performance, good idea.

source/Crew.cpp
lukearndt: Yes. If it looks weird, it's because of the ternary above it.

source/Crew.cpp
lukearndt: Good catch!

source/Crew.cpp
petervdmeer: No, it looks weird because there is a space-character too much before the c of count.

source/Crew.cpp
petervdmeer: I agree with tehhowch about not having an element in the collection treated differently.
But I like the original implementation for the datafiles from lukearndt better.

Can we change the code to automatically fill up open ship positions with the cheapest crew member available? So no special and regular crewmembers, but threat all crew as special, then fill all the minimum numbers and then just fill any left-over positions with the cheapest member possible.

source/Crew.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+int64_t Crew::CalculateSalaries(const vector<shared_ptr<Ship>> &ships, const Ship * flagship, const bool includeExtras)
+{
+	bool isFlagship = false;
+	bool checkIfFlagship = true;
+	int64_t totalSalaries = 0;
+	
+	for(const shared_ptr<Ship> &ship : ships)
+	{
+		if(checkIfFlagship)
```
petervdmeer: For maintainability of the code I would suggest to define the boolean isFlagship here within the loop and remove the "checkIfFlagship" variable completely.

data/crew.txt
```diff
@@ -0,0 +1,29 @@
...
+
+crew "default"
+  name "Regulars"
+  salary 100
+
+# crew "pilot"
+#   name "Pilots"
+#   salary 250
+#   "avoids flagship"
+#   "minimum per ship" 1
```
petervdmeer: How does the "minimum per ship" work for automated/automaton ships?

source/Crew.h
```diff
@@ -0,0 +1,64 @@
...
+	int64_t Salary() const;
+	const std::string &Id() const;
+	const std::string &Name() const;
+
+private:
+	// If true, the crew member will not appear on escorts
+	bool avoidsEscorts = false;
+	// If true, the crew member will not appear on the flagship
+	bool avoidsFlagship = false;
+	// Each valid ship has at least this many of the crew member
```
petervdmeer: What about automated ships?

source/Crew.h
```diff
@@ -0,0 +1,64 @@
...
+	int64_t minimumPerShip = 0;
```
petervdmeer: What if a ship has less crew that the minimum required? Suppose each ship needs a Captain, a Cook and a Pilot. How would that work for ships with a crew of 2?

data/crew.txt
```diff
@@ -0,0 +1,29 @@
...
+# 
+# crew "junior officer"
+#   name "Junior Officers"
+#   salary 500
+#   "population per member" 5
```
petervdmeer: If we want to incorporate Zitchas remarks, then we need to differentiate between `ship-population per member` and `fleet-population per member`.

petervdmeer:
@Zitchas: The original crew implementation and the crew implementation here by lukearndt allows for full automatic calculation of crew costs. I don't think that allowing the addition of unique beneficial crew members like "Space Marines" or "Security Officers" is a good idea.

If we get such different classes of crew members, then we also need to add UI and micromanagement to indicate which type of beneficial crew member got hired for which ships.
The effects of such beneficial crew member types can also already be gotten by adding outfits like Security stations and H2H weapons (and theses days they can also have maintenance and operating costs if desired).

Zitchas:
Fair enough. It sounded like an interesting possibility, but if it is going to take too much to get it to work...

An alternative thought occurred to me: Nothing's stopping anyone from creating a "Space Marine" outfit that has a 200/day maintnance cost, and requires 1 crew (similar to how turrets require 1 crew), and list them as a H2H outfit. Then they'd be used as weapons in combat, providing the appropriate bonus.

source/Crew.cpp
```diff
@@ -0,0 +1,232 @@
...
+			: ship->RequiredCrew()
+		) - specialCrewMembers
+		// If this is the flagship, one of the crew members is the Captain
+		- isFlagship;
+
+	const Crew *defaultCrew = GameData::Crews().Find("default");
+	
+	if(!defaultCrew)
+	{
+		defaultCrew = new Crew();
```
tehhowch: This use of [`new`](https://en.cppreference.com/w/cpp/language/new) will create a memory leak. You should revise your approach.

source/Crew.cpp
lukearndt: Oh well spotted!

source/Crew.cpp
lukearndt: This was a micro performance optimisation; it does add a fair amount of clutter though. Probably isn't worth it.

source/Crew.cpp
lukearndt: That's a neat idea.

source/Crew.h
lukearndt: Good point. I've addressed this by replacing "minimum per ship" with a "place at" node. This will allow us to hand-place crew members in specific positions.

This solves the problem because we can have our "minimum per ship" idea with finer-grained control. It also means that we can have non-scaling crew members on bigger ships, such as your Cook.

data/crew.txt
lukearndt: Good point. The variable is now called "ship population per member". I like the new name better, it's clearer.

source/Crew.h
lukearndt: Perhaps in the future we can add a flag that prevents a crew member from showing up on automated ships?

Sounds like a great idea for a follow-up PR.

source/Crew.cpp
```diff
@@ -0,0 +1,246 @@
...
+		// Add the crew members to the manifest
+		crewManifest[crew.Id()] = numberOnShip;
+		
+		// Add the crew members to the total so far
+		crewAccountedFor += numberOnShip;
+		
+		// If this is the cheapest crew type so far, keep track of it
+		// Use non-parked salaries so that crew are consistent
+		if(get<0>(cheapestCrew).empty() || crew.Salary() < get<2>(cheapestCrew))
+			cheapestCrew = make_tuple(crew.Id(), numberOnShip, crew.Salary());
```
petervdmeer: The cheapest crew member for a parked ship could be different from the cheapest crew-member on a non-parked ship.

lukearndt: Yes, I considered that and decided it was better for the ship to have consistent parked / non-parked crew than for it to be the cheapest possible result in each state. Hence the comment.

source/Crew.cpp
```diff
@@ -0,0 +1,246 @@
...
+#include "Crew.h"
+#include "Files.h"
+#include "GameData.h"
+
+using namespace std;
+
+void Crew::Load(const DataNode &node)
+{
+	if(node.Size() >= 2)
+		id = node.Token(1);
```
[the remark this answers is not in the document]
lukearndt: 🤔 I think that would be an invalid crew entry and we should reject it with an error.

data/crew.txt
```diff
@@ -0,0 +1,29 @@
...
+
+crew "regular"
+  name "Regular Crew"
+  salary 100
+
+# crew "pilot"
+#   name "Pilots"
+#   salary 250
+#   "avoids flagship"
+#   "place at" 1
```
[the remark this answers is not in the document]
lukearndt: But then we need a strategy for defining crew priority relative to each other, which adds just as much micro-management if not more.

source/Crew.cpp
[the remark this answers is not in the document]
lukearndt: Ah - yes, and that's exactly what happens already. If I remove the crew's ID string, I get this when I start the game:

```
Skipping unrecognized root object:
file "<folder path>/Endless Sky.app/Contents/Resources/../Resources/data/crew.txt"
  crew
```

source/Crew.cpp
petervdmeer: Okay, clear decision.

data/crew.txt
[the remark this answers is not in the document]
petervdmeer: Sorry, my last comment was not fully clear. The strategy that I wanted to suggest for (automatically) filling the manifest is:
- 1 First fill ships manifest with "minimum on ship" crew-members; most expensive to least expensive
   - If there are more "minimum on ship" than crew openings, then only the most expensive ones get filled.
- 2 Fill remaining ships manifest with "ship population per member"; most expensive to least expensive.
- 3 Fill remaining ships manifest with "cheapest crew-member possible"; for all remaining crew openings.

This strategy deterministically resolves the crew composition without getting more crew than that the ship has.

My impression is that the crew salaries mechanism would work great to set larger recurring costs for fleets (when the "fleet population per member" option would also be added). There are however already other mechanisms to get additional recurring costs (maintenance costs, operating costs) for certain (larger) ship types, so having some method to trigger a specific member above a certain amount of crewmembers per ship doesn't add too much value in my opinion.

Regarding `place at`; I really don't like numbering the "seats" on a ship. In the end we want to know how much of each crewmember the ship has, but I don't really care where exactly in the ship each crewmember is placed.

Pointedstick:
Alternatively, it occurs to me that going even more abstract could work too:, each ship would simply define a "daily upkeep cost" which would reflect the combination of crew salaries, refueling, shield recharging, hull repair, etc. Hiring additional crew would just increase it by 100 credits per extra crewman, and you can assume that these are all NCOs, with the salaries of the officers and captain baked into the basic upkeep cost.

ThrosturX:
> Alternatively, it occurs to me that going even more abstract could work too:, each ship would simply define a "daily upkeep cost" which would reflect the combination of crew salaries, refueling, shield recharging, hull repair, etc. Hiring additional crew would just increase it by 100 credits per extra crewman, and you can assume that these are all NCOs, with the salaries of the officers and captain baked into the basic upkeep cost.

I personally prefer the tiered approach (captain - officer - crewman) as this would open up the possibility to fill "officer seats" with specialized officers in the future, e.g. security officer for increased capture strength and decreased reload time or whatever. This also paves the way to improving the "boarding mechanic minigame" by upgrading or training your crew or whatever. It also allows for the interesting idea of "slave crew" which will hopefully be something most captains only try once or twice due to some discouraging mechanics.

That being said, I tend to think of crew salaries as upkeep costs anyway. Maybe the lowest paid crew member gets paid 100 credits, but uses 98 on most days to refuel the ship, poor fellow.

lukearndt:
The goal of this system was to be flexible enough to enable all sorts of different salary models based on the config. I think it's worth noting that many of the ideas that have been presented in the past few days are already achievable with the code that I developed. The captain / officer / crewman model that @pointedstick proposed would just be a set of config; it and I don't think it would require any code changes at all.

I gave up trying to get this merged quite some time ago, and I don't expect that I'll be pushing any further commits. I've enabled the 'allow edits from maintainers' option; use the code as you wish, as long as you preserve the copyright the files that I created.

PR #4674, opened 2019-11-14
Reparent jump-capable carried ships to out-of-system flagship

petervdmeer:
**Bugfix:** This PR addresses an issue where jump-capable carried ships were not jumping to the flagship

## Fix Details
This fix sets the flagship as parent for player-owned jump-capable carried ships without a parent.
[Bobby Bugfix.txt](https://github.com/endless-sky/endless-sky/files/3848557/Bobby.Bugfix.txt)

I believe the loss of the parent can happen if the flagship is in a different system and has no or insufficient bays and when this condition in AI.cpp "(!hasSpace && !Random::Int(1200))" is true.

## Save File
This save file [Jumping Bug.txt](https://github.com/endless-sky/endless-sky/files/3848561/Jumping.Bug.txt) can be used to verify the bugfix. The bug will occur when using 7558528209a338fcdd1d1a254fa835ef15bdbc03, and will not occur when using this branch's build.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -707,9 +707,19 @@ void AI::Step(const PlayerInfo &player)
 			bool hasParent = parent && parent->GetGovernment() == gov;
 			bool hasSpace = hasParent && parent->BaysFree(it->Attributes().Category() == "Fighter");
 			bool inParentSystem = hasParent && parent->GetSystem() == it->GetSystem();
-			if(!hasParent || parent->IsDestroyed() || (!hasSpace && !Random::Int(1200))
+			bool canJump = it->JumpFuel();
+			// If the flagship is in a different system, then set it as parent for orphaned jump-capable
+			// carried ships to ensure that the orphaned ships jump towards the flagship.
+			if(it->IsYours() && canJump && !hasParent && flagship && flagship->GetSystem() != it->GetSystem())
+			{
+				// Set flagship as parent for orphaned fighters and drones.
+				parent.reset();
+				parent = player.FlagshipPtrReadOnly();
+				it->SetParent(parent);
+			}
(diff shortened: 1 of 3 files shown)


tehhowch:
I think 2f34685c6915e224c32074612e4d0d384b464347 should address this.

(Adding `FlagshipPtrReadOnly` feels really hacky to me)

petervdmeer:
> I think [2f34685](https://github.com/endless-sky/endless-sky/commit/2f34685c6915e224c32074612e4d0d384b464347) should address this.

Your change lacks a little bit of documentation/explanation, but by reading the code I get the impression that the code now makes sure that the flagship is always set as parent for all jump-capable ships and that the parent is never removed for player-owned ships, so that carried ships always have a parent.

That looks like a good solution. We might still have an issue if:
- A player-owned carrier gets captured (and it's escorts become orphaned)
- Some jump-capable fighter had that carrier as parent
- The player was just busy jumping out of the system while this happens.

But this situation is quite unlikely, a fighter would re-parent the next time the AI::Step() function is run so this issue can only happen on the last AI::Step() call that the player still is in a system. 

The situation would also fix itself when the jump-capable fighter would enter a system with a non-carryable player-owned ship and it would also fix itself when the player lands and takes off again.

> (Adding `FlagshipPtrReadOnly` feels really hacky to me)

Agreed, it didn't look very nice to me either, but I didn't have a better idea.
I was actually expecting a review-comment with some suggestion on how to improve that part, but a full fix for the issue also works for me. :-)

petervdmeer:
Closing PR, fixed by other commit.

tehhowch:
That situation is even more unlikely since no other ships capture other ships currently 😅

PR #4677, opened 2019-11-17
Update MFirebird Hull

Arachi-Lover:
the description of the Marauder Firebird at one point mentions, like in other Marauder ships, how it has more shields and hull:
"With extra ..., hull plating"
but the MFirebird actually has *less* hull than it's normal counterpart, having 3000 for the Marauder version vs the 3400 for the standard.
In the same fashion as for other Medium Warships such as the Splinter and the Manta, and how their Marauder versions have 250 more hull than the Standard ones, this increases the Hull of the MFirebird to 3650.

### DIFF
--- a/data/marauders.txt
+++ b/data/marauders.txt
@@ -357,7 +357,7 @@ ship "Marauder Firebird"
 		category "Medium Warship"
 		"cost" 4100000
 		"shields" 7000
-		"hull" 3000
+		"hull" 3650
 		"required crew" 10
 		"bunks" 25
 		"mass" 350
--- a/data/marauders.txt
+++ b/data/marauders.txt
@@ -356,8 +356,8 @@ ship "Marauder Firebird"
 	attributes
 		category "Medium Warship"
 		"cost" 4100000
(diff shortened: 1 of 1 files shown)


tehhowch:
The Firebird hull was previously 2800, but was modified here: https://github.com/endless-sky/endless-sky/pull/3411

Amazinite:
Good catch. I’d just give the MFirebird the same +600 hull, -600 shields treatment that the Firebird got, so that the difference between the Firebird and MFirebird will match what they were before #3411.

Amazinite:
Guess the MFirebird just got a +400 hp buff, then. 😎

tehhowch:
Ah, yep, should go to 3600 not 4000.

tehhowch:
And done

Arachi-Lover:
yah, for some reason couldn't understand wtf Derp meant, and I had like 7 git tabs open and ended up getting confused with all the stats and shit, sorry about that, whoops.

Amazinite:
All good.

PR #4687, opened 2019-11-27
Differentiated Archons' appearance

beccabunny:

## Summary
This PR adds two more variants to the base Archon, to make the biological part of it a little more explicit. Following the example of Michael's animation, subtle and noticeable only if looked very closely, these changes are mostly restricted to the tendrils and are supposed to be almost as hard to notice as the animation.
The base Archon is used two times, same for the Third Archon
**Base Archon**
Screenshot (541)
**Second Archon**
Screenshot (540)
**Third Archon**
Screenshot (537)

## PR Checklist
[archons.zip](https://github.com/endless-sky/endless-sky/files/3898030/archons.zip)
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/102


Zitchas:
Looks good. Rather small differences. I'm not sure they are different enough to really be worth having several different ones. It isn't like anyone is going to see them next to each other, right? But on the same token, I like them.

Amazinite:
Don’t rename the “archon” image to the “archon0”, otherwise I believe you’ll turn the archons on all old pilots invisible since the ships in the player’s save will be looking for “archon”, not “archon0”.

beccabunny:
Whoops, I didn't think of it

finite-galaxy:
For comparison, the World-Ship variants are labelled B and C, rather than (Second) and (Third), and the image file names end in ` a`, ` b`, ` c` (with space), not `0`, `1`, `2`. Wouldn't that be better here as well, for consistency?

Zireael07:
Either way, renaming the current archon gfx  will turn it invisible, right?

Bladewood:
I think @finite-galaxy is right about the lettering. It might also be better to reserve numbers for animation frames, rather than to distinguish ships.

data/drak.txt
```diff
@@ -168,9 +168,9 @@ effect "antimatter dust"
 
 
 
-ship "Archon"
+ship "Archon A"
```
Amazinite: The ship name shouldn't have the variant letter in it, otherwise this actually shows up in game. (The world-ship a image doesn't go to the "World-Ship A" ship, it's just "World-Ship".)

copyright
```diff
@@ -542,6 +542,8 @@ Files:
  images/planet/station1c*
  images/planet/station2c*
  images/planet/station3c*
+ images/ship/archon1*
+ images/ship/archon2*
```
Amazinite: The file names are no longer 1 and 2.

PR #4692, opened 2019-11-29
Add Local God person

LocalGod79:
**Content (Person)**
## Summary
add a graphically updated old-style Korath World Ship (Ursa Polaris) person named Local God, who promotes the plugin of his ship's name (which revolves around building a station that uses the same base model).

the intention is; use the persons feature of the engine to showcase how plugin content can be created/developed with or without the intention to be added to the main game via external connections to the community.

also, to appease many of the opinion that there should be more person ships, and that I should be one of them (because of my contributions in the past).

## PR Checklist
 - [✔] I updated the copyright attributions, or decline to claim copyright of any assets produced or modified.
 - [✔] I uploaded the necessary image, blend, and texture assets here https://drive.google.com/open?id=1n_Zk3SRjF5pVWKaQvIi1V6aP132nMCE0
 - [✔] I created a PR to the [endless-sky-high-dpi repo] with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/104


Arachi-Lover:
yes

Zitchas:
That's quite the ship. I like it!

data/persons.txt
```diff
@@ -611,3 +611,161 @@ ship "Modified Osprey"
...
+		category "Heavy Warship"
+		cost 142857
+		shields 39000
+		hull 27000
+		"required crew" 559
+		"bunks" 972
+		"mass" 1839
+		"drag" 12
+		"heat dissipation" .9
+
```
Amazinite: Looks like you have some extra whitespace here.

data/persons.txt
```diff
@@ -611,3 +611,161 @@ ship "Modified Osprey"
...
+		hull 27000
+		"required crew" 559
+		"bunks" 972
+		"mass" 1839
+		"drag" 12
+		"heat dissipation" .9
+		
+		"fuel capacity" 600
+		"solar collection" 15
+		"afterburner thrust" 200
```
Amazinite: If you have built in afterburner thrust, then I believe you should also be able to have a built in afterburner sound and effect.

data/persons.txt
LocalGod79: from where i removed the cloak attribute...lol

data/persons.txt
LocalGod79: i probably could, but i actually wanted the extra thrust to be subtle

Amazinite: The afterburner thrust of a normal afterburner is only 25, but here you have 200. Since there's no afterburner heat, energy, or fuel, the afterburner will always be in use when in combat. And with only a drag of 12, I believe that means that this ship has a max speed of 1217 while out of combat and 2217 while in combat.

That seems a bit more than subtle to me. :)

LocalGod79: all intentional - heat damage is very difficult to balance against, and pretty much all other ships fare poorly against it - this ship was designed to be resistant to it, and yet is not indestructible (or "heat-proof" - since there's no such thing with flamethrowers).

in practice, because of the mining personality, afterburners will nearly always be in use - i usually play with the sound off, but Power of the People's spawn is always obvious because of the ionic afterburner plume (which is nearly always present when they are)

if you think the ship is too fast, i'm open to slowing it down, but the ship is supposed to attract attention (and perhaps point people at a long standing and popular plugin which may or may not ever get pulled for the main branch - which is part of the point), and further PotP's "Boss Ship" motif for [Author] government ships.

Amazinite: It's fine by me, I'm just saying that it isn't something I'd consider subtle.

LocalGod79: fair enough - to me, having a ship operate with a max speed of 1217 in very rare instances rather than 2217 most of the time seemed pretty subtle in my play testing - almost unnoticeable, actually.

Amazinite: >i'll also note that the multiple thrusters have more to do with the appearance of the flare(s)

Yeah, I guess as much when flying this around in game. Looks great!

PR #4671, opened 2019-11-13
FW campaign changes

Amazinite:
## Summary
Addressing a number of small issues pertaining to the FW Campaign all at once.

## Todo
- [x] In FWC, the player now doesn't magically know the name of the Pug before being told what they are (#2693).
- [x] In FW Middle, characters will now react if the player was suspended from the FW at the end of FW Start (#2687).
- [x] Rastaban and the surrounding systems now revert to FW ownership for bad karma Reconciliation runs, just much later than during good karma (#2696).
- [x] Flamethrowers now unlock for FW runs regardless of whether you helped test them or not.
- [x] Fixed some minor reputation issues that could occur in FWR (#2695).
- [x] Normalized capitalization of Navy ships (#2698).
- [x] Epilogue for Danforth changes based off of the branch (#2663).
- [x] Stacks Cores now unlock for FW runs regardless of whether you helped escort the freighters or not. 
- [x] Allow the player to decline going to the meeting on Albatross if they say they're gay (mentioned in #3705).
- [x] Because I removed Rastaban and the surrounding systems from being completely tied to karma in FWR, I've made it so that Tarazed and the surrounding systems now join the FW based off of karma. 
- Normalize Ijs vs Eyes, at least within the same mission (#2585)? **Doesn't seem necessary.**
- Epilogue for Tomek (mentioned in #2663). **Will change in or post #4672.**

PR #4693, opened 2019-11-30
Moved the flamethrower missions to side plots instead of start.

Bladewood:
The "FW Flamethrower" missions are listed under `free worlds start.txt`. It might sound logical, since flamethrowers are subsequently used in "FW Southern Battle" in `free worlds middle.txt`, but their use there is completely independent of whether or not you actually choose to help Barmy.
Compared to the stack core and catalytic ramscoop series, the flamethrower really isn't that much more special to warrant its special place under `free worlds start.txt`.

It's a pretty minor issue that doesn't actually affect gameplay, but for organizational reasons, this pull moves FW Flamethrower into `free worlds side plots.txt` to match.

There might be value in making an alternate mission line where you fight FW Southern Battle without flamethrowers, but that's probably an unnecessary amount of work compared to just moving the mission into another file.


Amazinite:
>There might be value in making an alternate mission line where you fight FW Southern Battle without flamethrowers, but that's probably an unnecessary amount of work compared to just moving the mission into another file.

I recently (#4671) made it such that the flamethrower and the stack core (another sideplot reward) are unlocked regardless of if the player does those missions, they just show up much later.

PR #4672, opened 2019-11-13
Improve FW Checkmate

Amazinite:
## Background
Ref: #4121

The Checkmate route is currently flawed, made evident by the poll mentioned in #4121 where players rarely choose it. I believe that the following plot points are what make the route flawed:
* The choice between Checkmate and Reconciliation is basically made for you. The Syndicate, who are suppose to be your ally, make no claim as to why they are pursuing this individual who claims to be a defector, and in their pursuit of him they ruthlessly murder your people. This gives you no good reason *not* to rescue the defector, as doing the opposite is an illogical choice that goes against everything you stood for leading up to that point (i.e. proving the FW innocent and gaining independence).
* Even within Checkmate, the ending isn't as satisfying as Reconciliation. This is because at the very beginning of the story when the bombing occurs, two questions are asked: 
   * Will the FW become independent? Both routes currently answer this question, the answer being "yes, the FW does become independent."
   * Who's behind the bombing? Currently, only Reconciliation answers this question, the answer being the Alphas and Syndicate extremists. Checkmate gives no answer to this question; everyone seems to just not care who it was after the Pug show up. 

So the story goals are the following:
1. Don't have the choice between Reconciliation and Checkmate lean so heavily toward the former.
2. Have Checkmate provide an answer to the question of who was behind the bombing.

A non-plot related criticism (not flaw) of Checkmate is the weaker rewards, so it would also be beneficial to provide Checkmate with a better reward.

## Summary
Improved the FW Checkmate route by doing the following:
* FWC has been given an additional reward that improved upon the nuclear missile: a warhead loading system that allows 6 nukes to be held at the same time instead of having a one-and-done system.
* Changed the end of FW Middle to make the choice between rescuing and handing over the defector more ambiguous.
  * Characters now sometimes refer to the defector as the terrorist.
  * The Syndicate now don't ruthlessly murder the monks, instead just making a mess of their monasteries in search of the defector.
  * The Syndicate make their own claims about the defector, saying that he is in fact the leader of the extremist group who developed the nukes.
  * The Syndicate promise to use defector to track down the "true terrorists." 
* In FWC, the player can now ask about the progress that the Syndicate are making toward proving the FW innocent, and added some extra dialog to make the FW more sympathetic for using nukes.
* In FWC, instead of the story ending after the Pug have been defeated, there are now missions that mirror the Syndicate Extremist missions of FWR where you defeat the "terrorists" responsible for the bombings.
[...]

Issue #3772, opened 2018-06-11
Capturing Bloodsea- some Free Worlds plot inconsistantices. 


jlnxr:
While this game is well polished in general, there's a few rough points in some of the less-traveled routes. One is dominating Bloodsea for Tomek vs for the Council. 

In "FW Pirates 3.1" You can ignore the Senate's orders and attack Bloodsea instead of going to Thule, the Senate apparently having authority here. If you succeed in completing the mission, Tomek is placed under house arrest and you are sent to the Senate, where you are practically laughed at for trying to occupy a world. And you are suspended and loss salary for a whole before getting another mission for no follow up on your disobedience. 

In "FW Bloodsea 1.2" and "FW Bloodsea 1.2A" you can instead chose, without consulting the Senate, to do _the exact same thing_ and yet... this time it works. You can now occupy Bloodsea.

This is not logically possible. To make matters worse, a player who might follow Tomek into attacking Bloodsea in the beginning is likely to chose occupying them later on vs the spaceport based on playing style/brutuality. 

Suggestions: 

-Remove the option to occupy Bloodsea later on 
Or

-Have an explaination for why the Free Worlds will now occupy the planet, such as, before Poisonwood they opposed threatening obital bombardment, but they are now no longer above that.

jafdy:
At that point it is seen as "The only option", it also doesn't work so well, rather than occupying it you are simply not giving any other option but to stop with pirates. In the end bloodsea goes independent.

jafdy:
At that point it is seen as "The only option", it also doesn't work so well, rather than occupying it you are simply not giving any other option but to stop with pirates. In the end bloodsea goes independent.

Amazinite:
From FW Bloodsea 1
> "The Senate has authorized a diplomatic mission to Bloodsea," he says. "It's the smallest of the pirate worlds, and we're going to see what we need to offer them to get them to join the Free Worlds. The official settlements there only amount to a few thousand people, so it shouldn't be too hard to convince them that joining us is better than resisting us."
>
> (Player choice) "Wait, didn't things go south rather quickly the last time we attempted this?"
>
> Alondo grins. "Yes, but the **Free Worlds is in a much stronger position now,** and the pirate worlds know it's only a matter of time before they are forced to become civilized. And, the incident on Poisonwood proved that we can't afford to fight a war on two fronts. We need to eliminate the pirate threat, and a diplomatic solution with them is the only sort that will last. So, let's get going!"

[...]

PR #4694, opened 2019-12-01
FW Middle: (better) explain why attacking Bloodsea is fine now when it wasn't before

Amazinite:

## Summary
Ref: #4671 (Forgot about this issue for the first PR.), #3772, #4672
A plot hole that I've seen brought up before concerns how the player gets punished in FW start for invading Bloodsea, but is given the option to invade Bloodsea in FW middle without consequences. This is brought up in #3772, where I highlighted how it is somewhat addressed after a single choice, but I feel that it isn't explained well enough currently, especially given that it's only explained by a single line that you only see if you choose the correct dialog.

* The leaders of Bloodsea now recognize you if you invaded them previously.
* Alondo now references the Tomek situation and explains why it's different now.
* Should you choose to invade, Alondo contacts the Senate for permission, which is granted.
* JJ is now confused about you deciding to invade, instead of immediately giving into the idea. This plays into his character trait expressed during the FW end choice, where he wants to take the defector for diplomacy instead of continuing to fight the Navy. (JJ seems to be more of a reluctant warrior, but currently it doesn't seem like that in these missions.)
* Changed a sentence in FWC that is wrong if the player attack Bloodsea or Greenrock previously, qualifying the claim that the FW hadn't invaded any unwilling system prior by saying that it's only an unwilling Republic system that they hadn't invaded.

### DIFF
--- a/data/free worlds middle.txt
+++ b/data/free worlds middle.txt
@@ -1413,18 +1413,16 @@ mission "FW Bloodsea 1.1"
 				`	"Well, building a spaceport is probably cheaper than fighting a major battle."`
 					goto port
 				`	"They're asking a lot. Why don't we just invade them?"`
-					goto attack
+					goto invade
 			label port
 			`	"Yes," he says, "except that I can almost guarantee that when we show up on their doorstep with everything they asked for, some local warlord will attack our ships, steal our supplies, enslave our workers, and build himself a nice cozy island fortress."`
 			`	"So, we attack them?" you ask.`
-			label attack
+			label invade
 			branch hostile
 				not "event: fw suppressed Bloodsea"
 			`	"They didn't seem very happy about you doing that the first time. But..." Alondo pauses for a moment.`
 			label hostile
-			`	"That's certainly the simpler option," he says, "especially because I can almost guarantee that when we show up on their doorstep with everything they asked for, some local warlord will attack our ships, steal our supplies, enslave our workers, and build himself a nice cozy island fortress.`
-			label invade
-			`	"Tomek may have been punished for disobeying the Senate's orders, but we're in a different situation now. Our numbers have swelled because of our victories and new territory, and we now have the Dreadnoughts at our disposal, so occupying Bloodsea is now a viable option.`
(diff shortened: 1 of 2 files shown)


data/free worlds middle.txt
```diff
@@ -1411,13 +1416,14 @@ mission "FW Bloodsea 1.1"
 					goto invade
 			label port
 			`	"Yes," he says, "except that I can almost guarantee that when we show up on their doorstep with everything they asked for, some local warlord will attack our ships, steal our supplies, enslave our workers, and build himself a nice cozy island fortress."`
-				goto attack
-			label invade
-			`	"That's certainly the simpler option," he says, "especially because I can almost guarantee that when we show up on their doorstep with everything they asked for, some local warlord will attack our ships, steal our supplies, enslave our workers, and build himself a nice cozy island fortress."`
-				goto attack
-			label attack
 			`	"So, we attack them?" you ask.`
```
AmkG: Should this text, as well as `label attack` be moved as well?

Amazinite: What do you mean? I've removed `label attack` from this conversation.

AmkG: I think you should *retain* `label attack` as well, and move this and its label to before the new position of "You're on the Council now".

As-is, I believe what will happen, if you select "port" option here, because of fallthrough, if the player never suppressed Bloodsea:

> "Well, building a spaceport is probably cheaper than fighting a major battle."`
>
> "Yes," he says, "except that I can almost guarantee that when we show up on their doorstep with everything they asked for, some local warlord will attack our ships, steal our supplies, enslave our workers, and build himself a nice cozy island fortress."
>
> "So, we attack them?" you ask.
>
> 	"That's certainly the simpler option," he says, "especially because I can almost guarantee that when we show up on their doorstep with everything they asked for, some local warlord will attack our ships, steal our supplies, enslave our workers, and build himself a nice cozy island fortress. Tomek may have been punished for disobeying the Senate's orders, but we're in a different situation now. Our numbers have swelled because of our victories and new territory, and we now have the Dreadnoughts at our disposal, so occupying Bloodsea is now a viable option."`

Mentioning the "cozy island fortress" twice with almost the same wording each time after the player asks about attacking them, seems a bit too repetitive of Alondo.

Amazinite: Ah, hadn't noticed the double mention of the fortresses. Good catch.

Issue #4698, opened 2019-12-03
Damage dropoff/decay with range mechanic


Amazinite:
**Is your feature request related to a problem? Please describe.**
In relation to #4697, inaccuracy can't be increased on beams as a way to make them worse at their maximum range, as high inaccuracy on lasers looks very weird. Lasers should still have some way to be less effective at range, though.

**Related Issue Links**
#4697

**Describe the solution you'd like**
Add some sort of mechanic that causes a weapon's damage to decay with range, similar to how damage scaling works with blast radius. This could make it so that lasers for example are at their most powerful up close, but at the very edge of their range they do reduced damage, say 75% of their maximum.

jafdy:
Could we make damage formulaic, so that we can make the damage the weapon does a function of the distance travelled, and possibly some other stuff. Or just make a new damage type/modifier to the existing type that decays to 0 at maximum range (with a second damage line adding a constant to that).
This would need to be conveyed to the player some how as well.

Amazinite:
>Could we make damage formulaic

How do you imagine the syntax for that would look?

Amazinite:
Syntax wise for this, I'm thinking having some attribute `"damage dropoff" <a> <b>` where `a` is when damage dropoff starts and `b` is when it ends. If `b` isn't specified, then the end is just when the projectile dies. Then `"damage decay" <x>` where `x` is a number from 0 to 1 representing the % damage at `b`, and with the damage decaying linearly from 100% at `a` to `x/100`% at `b`.

Mechanically, I imagine that a projectile would be given the knowledge of the position it was fired from by its armament and the position it impacted by Engine::DoCollisions, then just figure out the distance to know the range the projectile was at when it impacted to be used for the damage decay.

jafdy:
`damage 10/(distance+1)` would be 10 damage next to the weapon 1 at distance 9 ect.

Amazinite:
Sounds like a situation begging for someone to create a divide-by-zero.
That would also require weapon to be able to read a conditionSet, or something like it.

jafdy:
Yes, but we don't need to protect content creators from themselves.

NomadicVolcano:
It would also be useful to increase with distance, or have an optimal distance (increase and then decrease).  I see a straightforward way of implementing this:

    velocity 25
    lifetime 20
    "center point distance"  300
    "ranged damage percent"  50 100 80

[...]

PR #4703, opened 2019-12-05
Added a mechanic for projectiles to lose damage over range

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4698.

## Feature Details
Adds two new weapon attributes.
`"damage dropoff" <min> [<max>]`: The starting and ending range for when damage modification takes place. Before `<min>`, the weapon does 100% damage. After `<max>` the weapon does `"dropoff modifier"`% damage. Between `<min>` and `<max>` the damage output changes linearly from 100% to `"dropoff modifier"`%. If `<max>` is not specified, then it is set equal to the weapon's range.
`"dropoff modifier" <value>`: The % damage output after the max dropoff has been reached. Typically 0 to 1, but values greater than 1 are allowed (allowing the creating of a weapon that increases its damage with range). Omission of this means that dropoff modifier is 0. Including this without specifying a damage dropoff yields no effect. 

These values do not currently display in the outfitter. I imagine that the description of the weapon would describe whatever behavior the damage dropoff has. e.g. "The laser loses energy over range, meaning that it is more powerful up close than from afar."

## Usage Examples
This would make the Electron Beam's damage start decaying past half its range, and at the very tip of the laser the beam would be doing 75% normal damage.
```
outfit "Electron Beam"
	...
	weapon
		...
		"damage dropoff" 275
		"dropoff modifier" .75
```

This would create a Particle Cannon that between 400 and 600 range increases its damage up to 125%. Everything past 600 range would deal 125%, while everything before 400 range would deal normal damage, with the in between values being where the damage ramps up.
```
outfit "Particle Cannon"
	...
	weapon
		...
		"damage dropoff" 400 600
		"dropoff modifier" 1.25
```

This would create proton fragments with a sharp dip in damage past 200 range, stopping at 250 range. The range is from the creation of the projectile, meaning that the damage dropoff for the proton fragment starts 200 range after the original Proton Cannon projectile fragmented.
```
outfit "Proton Fragment"
	weapon
		...
		"damage dropoff" 200 250
		"dropoff modifier" .5
```


source/Weapon.cpp
```diff
@@ -318,6 +328,20 @@ double Weapon::Range() const
...
+double Weapon::DamageDropoff(double distance) const
+{
+	double minDropoff = damageDropoff.first;
+	double maxDropoff = damageDropoff.second;
+	
+	if(distance <= minDropoff)
+		return 1.;
+	if(distance >= maxDropoff)
+		return damageDecay;
+	return (1 - damageDecay) / (minDropoff - maxDropoff) * (distance - minDropoff) + 1;
```
petervdmeer: This calculation looks correct, but is a bit difficult to read. Maybe splitting the calculation would improve readability? For example:
```
double dropoffFactor = (distance - minDropOff) / (maxDropoff - minDropOff);
return 1. - (1. - damageDecay) * dropoffFactor;
```

source/Weapon.h
```diff
@@ -204,6 +207,10 @@ class Weapon {
 	
 	double piercing = 0.;
 	
+	bool hasDamageDropoff = false;
+	std::pair<double, double> damageDropoff;
```
petervdmeer: I suggest calling this variable(s) `damageDropoffRange`.

source/Weapon.cpp
Amazinite: I've put a comment here saying that the calculation is linear and split it up by defining the slope of the line separately.

Zitchas:
Could this be used to make a "safety" missile? Something like a nuke or a heavy rocket that has a very high amount of damage, but it starts at a very low damage amount and scales up to full strength over the course of the first few seconds of travel? So instead of having projectiles with area effects that are actually "smart" and don't damage allies, they merely don't do as much damage to things that are closer to the launching ship. Which helps keep the user safe, although it wouldn't help keep other allies safe.

Might make for an interesting set of weapons in that mid-ground between "indiscriminate explosions" and "precision detonation."

Amazinite:
I believe that a far more appropriate mechanic for doing that would be submunitions. (The parent projectile would have negative damage such as Korath Mines do.) With this though, you could do something like have a weapon of range 2000, have a base damage of 100, then have `"damage dropoff" 500 1000 "damage decay" 10`, creating a weapon that does only 100 damage within a range of 500, but then it ramps up and does 1000 damage beyond 1000 range up until its max range.

Zitchas:
I was thinking of  having it scale, as if whatever the damaging element had to "warm up" before reaching full strength, but either way, I'm liking the versatility of this mechanic. Both fading at range and intensifying up to specific ranges sounds both fun, interesting, and useful.

Amazinite:
While the attribute itself is called damage decay, damage growth is allowed. Should the attribute name be changed, and if so, to what?

jafdy:
What happens if someone sets `damageDropoffRange.first` to be `Range()` and `damageDropoffRange.second` is less than (or equal to) that? Do we then get a div0 error on L343.

petervdmeer:
@jafdy No I don't think that would happen, because distance is either equal to minDropoff (and line 338 handles the issue), or the sanity check on line 190 ensures that second is not less.

source/Weapon.cpp
```diff
@@ -172,13 +172,23 @@ void Weapon::LoadWeapon(const DataNode &node)
...
 			else
 				child.PrintTrace("Unrecognized weapon attribute: \"" + key + "\":");
 		}
 	}
-	// Sanity check:
+	// Sanity checks:
 	if(burstReload > reload)
 		burstReload = reload;
+	if(damageDropoffRange.first > damageDropoffRange.second)
+		damageDropoffRange.second = Range();
```
petervdmeer: To make the code more robust/resistant to corner cases I would suggest to use max(Range(), damageDropoffRange.first) here.

petervdmeer:
@jafdy I don't think there is a problem, but I did just post a comment which should make the code even safer for corner-case bugs.

petervdmeer:
> While the attribute itself is called damage decay, damage growth is allowed. Should the attribute name be changed, and if so, to what?

How about `damage mutation` or `damage transformation`? (I just looked up some synonyms for change in an online dictionary. Such synonym lists usually work nicely to get inspiration for alternative words.)

source/Weapon.cpp
Amazinite: That doesn't really seem necessary to me since if damageDropoffRange.first > Range(), then damageDropoffRange.second never gets used, as the `distance` of Weapon::DamageDropoff can only be at most equal to the weapon's range, so `if(distance <= minDropoff)` will always be true. (Which is just to say that a minDropoff >= Range is the same as not using this mechanic at all.)

Amazinite:
>What happens if someone sets damageDropoffRange.first to be Range() and damageDropoffRange.second is less than (or equal to) that? Do we then get a div0 error on L343.

The slope would never get calculated in such a scenario, as `if(distance <= minDropoff)` will always be true.

Zitchas:
I know "mutation" sounds terribly organic, but it how often do people read the code for the feels? Calling it a mutation instead of a modifier might make it clearer, since I would expect there may be many modifiers in the game. Is someone talking about The Damage Modifier, or just talking about applying modifiers to damage in general? vs if they talk about a damage mutation, well, that's pretty clear what they are talking about.

Amazinite:
To be more explicit, since this is only used for the damage dropoff, it could be `"damage dropoff modifier"`.

Amazinite:
I just realized that this mechanic will *not work as intended* when combined with missiles. Since this mechanic compares the distance between the point of firing and the point of impact to calculate the dropoff, using that as an analog for how far the projectile has traveled, any projectiles which turns will not result in an accurate calculation for how this mechanic is intended to work. 

The only solution I can think of is to have a separate form of dropoff that uses the remaining lifetime of the projectile to figure out how far it's traveled. This would then make missiles work as intended, but this way of calculating damage can't replace the current way, as otherwise lasers (with a lifetime of 1) wouldn't be able to have damage dropoff.

Amazinite:
This should work now for all weapon types. The projectile now tracks how far it's traveled, and that information is what it used to compute the damage scaling.

source/Weapon.cpp
```diff
@@ -176,13 +176,23 @@ void Weapon::LoadWeapon(const DataNode &node)
 				rangeOverride = max(0., value);
 			else if(key == "velocity override")
 				velocityOverride = max(0., value);
+			else if(key == "damage dropoff")
+			{
+				hasDamageDropoff = true;
+				double maxDropoff = (child.Size() >= 3) ? child.Value(2) : 0.;
+				damageDropoffRange = make_pair(max(0., value), maxDropoff);
```
petervdmeer, line 183: should we also do a max(0. maxDropoff) for maxDropoff, to ensure that that one is also always positive.

source/Weapon.cpp
```diff
@@ -324,6 +334,24 @@ double Weapon::Range() const
 
 
 
+// Calculate the percent damage that this weapon deals given the distance
```
petervdmeer, line 337: Not percent, but fraction.

Amazinite: Same difference.

source/Weapon.cpp
Amazinite: Well technically maxDropoff already can't be negative because minDropoff most be positive, and if maxDropoff is less than minDropoff then maxDropoff gets set to Range().

Issue #4668, opened 2019-11-13
Variants within variants: an easier way to implement new ship variants into fleets


Amazinite:
**Is your feature request related to a problem? Please describe.**
This is basically resurrecting #2143, but with a new suggested syntax. The core problem is that creating new ship variants and adding them to fleets is someone cumbersome, especially if you want a high variety of possible fleet combinations. This was brought up in #2143 in that if you have 3 ships with 3 variants each, there are 27 possible combinations of these three ships. It would be unreasonable to create 27 different fleet variants, so we end up losing out in fleet variant diversity for the sake of our sanity.

**Related Issue Links**
#2143

**Describe the solution you'd like**
Here is an example syntax of my idea of how we could make it easier to add variants.
```
variants "Leviathan"
	"Leviathan" 5
	"Leviathan (Laser)" 3
	"Leviathan (Hai Weapons)"
	"Leviathan (Hai Engines)"

fleet "Large Northern Pirates"
	...
	variant 4
		variants "Leviathan" 2
		"Headhunter" 2
	...
```

`variants "<name>"` is a list of ships similar to how `fleet` is a list of variants. The numbers following the ship names are their weight, functioning the same way as a fleet's variants. WIthin the fleet variant, the line `variants "Leviathan" 2` means to pull a ship from the `variants "Leviathan"` list twice (not the same ship twice, but roll for a ship twice).
The name of the `variants` does not need to match the name of the ship, and the ships in the `variants` don't even need to be the same ship. For example, one could create the following `variants` list.
```
variants "Republic Warships"
	"Cruiser" 2
	"Cruiser (Heavy)" 2
	"Cruiser (Mark II)"
	"Carrier" 4
	"Carrier (Mark II)"
```
`variants` could be modified by events in the same way as fleets, either being completely overwritten or simply added to or removed from. This means that whenever we create a new ship variant, instead of having to come up with a bunch of fleet variants for it to fit into, we could simply add it to a `variants` list that the `fleets` pull from.

Amazinite:
Slapping the milestone on this because #2143 was milestoned, and this really seems to me like a necessary feature to have at some point now that I've requested it.

Zitchas:
I think this would go a long way to encouraging a larger variety of ships for the placer to encounter without massively magnifying the size of our fleets file. I would like to see this implemented.

tehhowch:
[...]

PR #4704, opened 2019-12-06
Variants within variants

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4668

## Feature Details and Usage Examples
These changes allow variants to be defined by name just like conversations or fleet. Named variants can then be listed as children of variants in fleets.
```
variant "Large Navy Alpha"
	"Carrier"
	"Cruiser"
	"Frigate" 3

fleet "Large Republic"
	variant
		variant "Large Navy Alpha" 2
		"Combat Drone" 4
		"Gunboat" 2
```
Variants behave differently depending on if they are a child of a fleet or a child of another variant (i.e. they are nested variants). If they are a child of a fleet, then every ship in the variant is used according to the number provided. This is the exact same behavior as currently.
```
fleet
	variant
		"Cruiser" 2
		"Carrier" 3
```
The above would spawn 2 Cruisers and 3 Carriers if that variant was chosen. Similarly in the first syntax example (`variant "Large Navy Alpha" 3`), the "Large Navy Alpha" variant has a 75% change of being chosen and would spawn a Carrier, a Cruiser, and 3 Frigates.
But a variant that is a child of a variant instead chooses one of the ships in its list, using the numbers as weights similarly to how a fleet chooses a single variant according to their weights.
(e.g. the following would spawn a Gunboat and the nested variant would choose either a Cruiser 40% of the time or a Carrier 60% of the time.)
```
fleet
	variant
		variant
			"Cruiser" 2
			"Carrier" 3
		"Gunboat"
```
Similarly in the first example (`variant "Large Navy Alpha" 2` inside the variant), that variant would spawn 4 Combat Drones, 2 Gunboats, then choose 2 ships from the "Large Navy Alpha" variant, choosing a Frigate 60% of the time, a Carrier 20% of the time, and a Cruiser 20% of the time. (The same ship is not chosen twice. Instead, the variant rolls for two ships using the same weights each time, meaning there's a 36% chance of 2 Frigates, a 4% chance of 2 Cruisers, a 4% chance of 2 Carriers, and so on for any mix.)


Zitchas:
I like the idea of having nested variants like that, but I'm dubious about using the term "formation," mainly because it might add some confusion with the [formations that are being worked on](https://github.com/endless-sky/endless-sky/pull/4471).

Amazinite:
If I am to implement the idea, I'll go with "division" or "squadron" or "flotilla" (single words used to describe a smaller formation of a naval fleet) to avoid confusion with flying formations.
(Although I don't think using the keyword formation for both would conflict data-wise, as a flying `formation` is a root keyword and presumably a child of `fleet` whereas this `formation` is a child of `variant`).

Amazinite:
An alternative or addition to the `formation` idea might be to allow ships/variants in a variant to specify their weight and their size. e.g. This would send 2 Cruisers 25% of the time and 4 Frigates 75% of the time, the first number being the weight and the second number the size. 
Or perhaps reverse that so that the first number is the size and the second the weight. 
```
fleet
	variant
		variant
			"Cruiser" 1 2
			"Frigate" 3 4
```

Amazinite:
Looks like as of right now, this is creating some errors.
There are a lot of out of bounds traces. Notably, every out of bounds trace is for a fleet in a mission where the fleet has a number.
```
Requested token index (1) is out of bounds:
"deep missions.txt"
  mission "Deep: Scientist Rescue 1: Escorts"
    npc kill
      fleet 3
        variant
```
Every fleet defined by name is also creating the following warning.
```
Warning: variant "Bounty Hunters" is referred to, but never defined.
```
Help diagnosing these and finishing this PR would be appreciated.

source/Fleet.cpp
```diff
@@ -205,19 +206,38 @@ void Fleet::Load(const DataNode &node)
...
+				name = node.Token(1 + add);
+				named = true;
+			}
+			if(node.Size() >= 2 + add + named)
+				weight = child.Value(1 + add + named);
+
+			if(named)
+				variants.emplace_back(make_pair(GameData::Variants().Get(name), weight));
+			else
+				variants.emplace_back(make_pair(new Variant(child), weight));
```
tehhowch: Memory leak - you cannot call `new` without a corresponding `delete`

source/Fleet.cpp
```diff
@@ -205,19 +206,38 @@ void Fleet::Load(const DataNode &node)
...
+			if(node.Size() >= 2 + add && !node.IsNumber(1 + add))
+			{
+				name = node.Token(1 + add);
+				named = true;
+			}
+			if(node.Size() >= 2 + add + named)
+				weight = child.Value(1 + add + named);
+
+			if(named)
+				variants.emplace_back(make_pair(GameData::Variants().Get(name), weight));
```
tehhowch: Use the argument forwarding feature of `emplace_____` to select the appropriate constructor and create the emplaced type in-place

petervdmeer:
@Amazinite:

> (Although I don't think using the keyword formation for both would conflict data-wise, as a flying `formation` is a root keyword and presumably a child of `fleet` whereas this `formation` is a child of `variant`).

Formations as implemented in #4471 apply to individual ships and to variants. I even wrote an RFC to describe the keywords and locations ( https://github.com/EndlessSkyCommunity/endless-sky/pull/52).

Could you review the RFC for conflicts with your proposal? (And maybe also just review as a generic review to provide me with review-feedback?)

Amazinite:
Looks like an error where I used `node` instead of `child` was causing all the issues, so this is mostly working as intended now. It looks like the number of ships spawned by nested variants is incorrect, though.

I created this mission with a named nested variant and a nested variant defined within the fleet.  
```
variant "Warship"
	"Cruiser" 3
	"Carrier"

mission testing
	repeat
	job
	npc
		government "Republic"
		personality staying
		fleet 100
			variant
				"Frigate"
				variant "Warship"
				variant
					"Gunboat" 2
					"Rainmaker"
```
Ship | Expected Value | Actual Value (Expected given actual) |
----|----|----
Frigate | 100 | 100 (100)
Rainmaker | ~33 | 143 (133)
Gunboat | ~67 | 257 (267)
Cruiser | ~75 | 378 (375)
Carrier | ~25 | 122 (125)

So it looks like right now the named variant is giving five times the expected number of ships (500 combined between the Cruiser and Carrier instead of 100) and the nested variant defined inside the fleet is giving four times the expected number of ships (400 combined between the Rainmaker and Gunboat instead of 100).

EndlessSky.cbp
```diff
@@ -320,6 +320,8 @@
 		<Unit filename="source/TradingPanel.h" />
 		<Unit filename="source/UI.cpp" />
 		<Unit filename="source/UI.h" />
+		<Unit filename="source/Variant.cpp" />
+		<Unit filename="source/Variant.h" />
```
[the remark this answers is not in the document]
Amazinite: >(I don't have a Mac, so I cannot give you the exact change that is needed.)

Well I don't have a Mac either, so we'll need someone with the knowledge of updating that to do it.

source/Variant.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "Ship.h"
+
+#include <algorithm>
+#include <cmath>
+#include <iterator>
+
+using namespace std;
+
+namespace {
+}
```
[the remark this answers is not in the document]
Amazinite: Yeah, that's just a holdover from copy-pasting the file structure from another file and removing what wasn't needed. This will either be used or removed by the time I mark this PR ready for review.

Amazinite:
Right now I've prevented a variant from containing itself directly, but I haven't prevented a loop of variants from being a thing. e.g.
```
variant "Testing 1"
	variant "Testing 2"

variant "Testing 2"
	variant "Testing 1"
```
How about just preventing infinite nesting of variants? (i.e. changing Variant::NestedChooseShips to not look for variants.)

source/Fleet.cpp
Amazinite: Where would that go, or what would be another way to do this without using `new`?

tehhowch: Review the lifetime of the parent container to see the lifetime of a created object. It is undefined behavior or at best implementation defined to refer to/use objects that have gone "out of scope" and that have temporary lifetime - those are the situations when you would need to explicitly manage the object's lifetime via `new`+`delete`.

Most game objects remain in scope (ergo valid) because they are held within some container that is (eventually) a member variable of GameData, so as long as GameData is in scope, they are also in scope.

Amazinite:
Looks like the `1 + total weight` error was because I accidentally initialized the chosenShips of NestedChooseShips with all of the variant's ships, so
```
variant
	"Gunboat"
	"Rainmaker"
```
would choose a Gunboat, a Rainmaker, then randomly choose one more of those ships to add to the list.

source/Fleet.cpp
Amazinite: Would it be a bad idea to have `variants` be a `vector<pair<Variant, int>>` instead of having `const Variant *` and dereference the `const Variant *` variants from GameData so that `new/delete` doesn't need to be used here?

Amazinite: The better solution is probably to have two different vectors, one with created variants and another with variants pulled from the GameData, as MissionAction does something similar with conversations. (`Conversation conversation` for conversations created in the mission action and `const Conversation *stockConverstion` for conversations created by name and referenced by the MissionAction.)

Amazinite:
Looks like certain fleet variants are crashing the game. 😎 
Sit in Sol with the latest changes and you'll eventually crash.

Amazinite:
All the necessary code changes should now be done. If anything's broken, it's going to be the latest commit.

data/events.txt
```diff
@@ -1074,107 +1054,20 @@ event "fw safe passage ends"
...
+	variant "Rainmaker"
+		add "Rainmaker (Mark II)" 3
+	variant "Gunboat"
+		add "Gunboat (Mark II)" 3
+	variant "Frigate"
+		add "Frigate (Mark II)" 3
+	variant "Cruiser"
+		add "Cruiser (Mark II)" 3
+	variant "Carrier"
+		add "Carrier (Mark II)" 3
```
Amazinite: Something that I think is important pointing out is how the inclusion of this new variant functionality differs from the old system of adding new fleet variants. Whereas before 3/4ths of all Navy fleets became Mark II fleets from this event, this change instead makes it so that 3/4ths of all Navy ships within a single fleet variant are Mark II. That is, you no longer get fleets of all Mark II ships and all Mark I ships, but fleets of mostly Mark II ships and some Mark I ships.

Is this a desirable outcome for Navy fleets? Is there anywhere else that comes to mind where this mixing of ship variants could be an issue?

data/fleets.txt
```diff
@@ -17,50 +17,27 @@ fleet "Small Southern Merchants"
 		timid frugal appeasing
 	variant 60
 		"Shuttle"
```
Amazinite: Another question is, should we create a `variant` for every ship, even if that ship doesn't have any variants? For example, a "Shuttle" variant consisting of only "Shuttle". This would mean that all vanilla fleet variants would consist entirely of nested variants. The reason for doing this would be so that any future ship variant additions would require only to add to the `variant` for that ship in order to include it in any fleets, instead of perhaps having to figure out if said ship is already in a variant first, then creating the variant and placing it wherever needed if necessary. It's basically a question of doing the work now, even if nothing comes of it in the future (e.g. there may never be a Shuttle ship variant aside from the base model), or doing the work later.

data/fleet variants.txt
```diff
@@ -0,0 +1,340 @@
...
+	"Wasp (Meteor)"
+	"Wasp (Plasma)"
+	"Wasp (Proton)"
+
+# Multi-ship variants
+
+variant "Core Pirate Interceptors"
+	variant "Wasp (Pirate)" 5
+	variant "Hawk (Pirate)"
+	variant "Sparrow (Pirate)"
```
Amazinite: These multi-ship variants were more so made as examples of what can be done, but should I go through and create such variants wherever possible?

One thing to note about variants like these is that they're probably best used by pulling a single ship from them, at most two (as a stretch), but not more. Reason being that if I were for to example create a "Northern Pirate Medium Warships" variant containing the Firebird, Mule, and Corvette variants, then use that multi-ship variant wherever possible, you'd certainly see a higher variety in what type of fleets you see, but that'd come at the cost of a loss of perceived structure. Right now ships of the same model will often appear in groups (e.g. two or three of the same ship), but if you were to instead do `variant "Northern Pirate Medium Warships" 3`, you'd see any mix of those three ships, which outside of them all being medium warships could look rather unorganized, as if the pirates are just throwing together whatever they can. Now that perception might be exactly what we want from pirates, but such multi-fleet variants likely won't be desirable for organized factions like the FW/Militia, Syndicate, Navy, and perhaps even merchants. (Some multi-fleet variants could be used in some merchant fleet variants, making it look like some merchants are just using whatever they can get their hands on, but on the whole it wouldn't be a good appearance for merchants.)

As I say that, though, I've used the "Human Fighters" mutli-ship variant in greater quantities. Since it consists of only two ships, this doesn't lead to as much of a disorganized looking outcome.

EndlessSky.cbp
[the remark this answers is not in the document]
petervdmeer: This would be blocking before merging this change. You could ask somebody on Discord that has a Mac to provide you the contents. I also did that for formation flying and got a response on the required change.

data/events.txt
petervdmeer: The mixed fleet looks fine to me, but I guess a remove would also be in order. For example to remove "Mark I"s after some time.

petervdmeer: And we should test the game how it handles empty variants (for if some content creator makes a mistake and removes all variants).

data/fleet variants.txt
petervdmeer: Good considerations. For me it is fine to just leave the data-file changes here as-is and do further data-file changes in follow-up PRs if you feel that such updates are beneficial.

data/fleets.txt
petervdmeer: No, I don't think it makes sense to do this now. We might also just want to store top-level fleet variants and passed events in savegames so that such updates can also be done in the future for existing savegames.
If we go that direction, then defining such variants at this moment would be a lot of work with no return-on-investment.

data/fleets.txt
```diff
@@ -2784,38 +1864,25 @@ fleet "Marauder fleet IX"
...
 		"Marauder Leviathan (Weapons)"
-		"Firebird"
-		"Firebird (Plasma)"
-		"Firebird (Missile)"
+		variant "Firebird" 3
 	variant 1
 		"Marauder Leviathan (Weapons)"
-		"Firebird (Laser)" 3
-		"Firebird (Missile)"
+		variant "Firebird" 4
```
petervdmeer: The reviewers for data-files and code-changes are different;
This remark is a bit too late for this PR (since you already changed a very large amount of datafiles), but for future PRs please consider to keep the code-change together with a small example data-file change in one PR and to file the data-file changes in a separate PR.
(You can mark this remark with "Conversation Handled" after you read this comment.)

source/Variant.h
```diff
@@ -0,0 +1,63 @@
+/* Variant.h
+Copyright (c) 2019 by Michael Zahniser
```
petervdmeer: I would expect you to have created this file.

source/Variant.h
```diff
@@ -0,0 +1,63 @@
...
+	std::vector<const Ship *> NestedChooseShips() const;
+	int64_t NestedStrength() const;
+	
+private:
+	std::string name;
+	int total = 0;
+	int variantTotal = 0;
+	int stockTotal = 0;
+	std::vector<const Ship *> ships;
+	std::vector<std::pair<const Variant *, int>> stockVariants;
```
petervdmeer: Can you add some comments on what is exactly stored in stockVariants?

source/Fleet.cpp
```diff
@@ -203,25 +204,57 @@ void Fleet::Load(const DataNode &node)
...
-				if(toRemove.ships.size() == it->ships.size() &&
-					is_permutation(it->ships.begin(), it->ships.end(), toRemove.ships.begin()))
-				{
-					total -= it->weight;
-					variants.erase(it);
-					didRemove = true;
-					break;
-				}
+			// If toRemove is named, check if a stockVariant shares that name.
+			// Else, check if toRemove is equal to any of the variants.
```
petervdmeer: Shouldn't the remove remove both the stockvariants and the variant?

source/Fleet.cpp
```diff
@@ -496,45 +531,33 @@ void Fleet::Place(const System &system, Ship &ship)
...
 	unsigned index = 0;
-	for(int choice = Random::Int(total); choice >= variants[index].weight; ++index)
-		choice -= variants[index].weight;
-	
-	return variants[index];
+	// Choose between stockVariants and variants.
+	int chosen = Random::Int(total);
+	if(chosen < stockTotal)
+	{
+		for(int choice = Random::Int(stockTotal); choice >= stockVariants[index].second; ++index)
```
petervdmeer: Why Random again? Is it not possible to re-use chosen here? (It is in the same range as this random number.) Same for the for-loop below.

source/Fleet.cpp
```diff
@@ -496,45 +531,33 @@ void Fleet::Place(const System &system, Ship &ship)
...
+			choice -= stockVariants[index].second;
+		return *stockVariants[index].first;
+	}
+	else
+	{
+		for(int choice = Random::Int(total - stockTotal); choice >= variants[index].second; ++index)
```
petervdmeer: What happens with empty variants? Does this code still work if variants has zero elements (same for stockVariants above)?

source/Fleet.h
```diff
@@ -87,9 +78,11 @@ class Fleet {
 	const Government *government = nullptr;
 	const Phrase *names = nullptr;
 	const Phrase *fighterNames = nullptr;
-	std::vector<Variant> variants;
+	std::vector<std::pair<const Variant *, int>> stockVariants;
```
petervdmeer: Can you add a bit of comments on the stockVariants and on the variants variable to explain the difference?

source/Variant.cpp
```diff
@@ -0,0 +1,303 @@
...
+	
+	// If Load() has already been called once on this variant, any subsequent
+	// calls will replace the contents instead of adding to them.
+	bool reset = !variants.empty() || !stockVariants.empty() || !ships.empty();
+	
+	for(const DataNode &child : node)
+	{
+		bool add = (child.Token(0) == "add");
+		bool remove = (child.Token(0) == "remove");
+		bool variant = (child.Token(add || remove) == "variant");
```
petervdmeer: Should this first have a check if Token(1) exists if add or remove was given?

source/Variant.cpp
```diff
@@ -0,0 +1,303 @@
...
+							break;
+						}
+				
+				if(!didRemove)
+					child.PrintTrace("Did not find matching variant for specified operation:");
+			}
+			else
+			{
+				// If given the name of a ship, remove all ships by that name from this variant.
+				string shipName = child.Token(1);
```
petervdmeer: Also here, should there first be a check if Token(1) was actually given?

source/Variant.cpp
```diff
@@ -0,0 +1,303 @@
...
+	{
+		chosen = Random::Int(variantTotal);
+		unsigned variantIndex = 0;
+		vector<const Ship *> variantShips;
+		// Randomly choose between the stock variants and the non-stock variants.
+		if(chosen < stockTotal)
+		{
+			// Choose a variant according to the weights of the variants.
+			for(int choice = Random::Int(stockTotal); choice >= stockVariants[variantIndex].second; ++variantIndex)
+				choice -= stockVariants[variantIndex].second;
```
petervdmeer: Some of the code here looks duplicated from the code in Fleet.cpp. Is it possible to use a single/common set of functions instead of 2 sets?

data/remnant.txt
```diff
@@ -8,102 +8,77 @@
...
+		variant "Starling" 2
 	variant 2
-		"Albatross"
-		"Albatross (Sniper)"
-	variant 2
-		"Albatross (Heavy)"
-		"Albatross (Turret)"
+		variant "Albatross" 2
+
+variant "Gull"
```
petervdmeer: I also still feel that a different keyword (like versions) would be more appropriate here, since it still seems to have a different meaning than variant within fleets.

data/fleets.txt
Amazinite: Yeah, I was actually considering earlier removing the data changes and adding those separately after the code changes are merged, which I'll likely do so that the comments I made on the data changes can be discussed more, and since the code changes alone retain the current functionality, meaning no data change is actually necessary.

source/Fleet.cpp
Amazinite: The first random int call is from 0 to `total`, whereas this one is from 0 to `stockTotal`. `stockTotal` is the size of `stockVariants`, whereas `total` is the size of `stockVariants` + `variants`, so reusing the same chosen random number could result in an out of bounds error.

EndlessSky.cbp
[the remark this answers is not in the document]
Amazinite: I was asking around but got no responses.

source/Variant.cpp
Amazinite: I don't believe it would be possible to even reach this point if Token(1) didn't exist. Variant::Load isn't going to be called by a line that just reads `add` or `remove`.

source/Fleet.cpp
Amazinite: What do you mean by that? `remove variant` can only match either a stockVariant or a normal variant. There won't be both a stockVariant and normal variant that match.

source/Fleet.cpp
Amazinite: If `variants` is empty, then `total == stockTotal`, so `if(chosen < stockTotal)` will always be true. If `stockVariants` is empty, then the reverse is true where `total == total - stockTotal` and `if(chosen < stockTotal)` will always return false. If both `variants` and `stockVariants` are empty then we don't even have a fleet.

source/Variant.cpp
Amazinite: There is some duplication, so I'll look in to figuring out if that can be addressed.

source/Variant.h
Amazinite: Just because I created the file doesn't mean the copyright needs to be to me.

EndlessSky.cbp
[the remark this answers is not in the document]
petervdmeer: Sounds like we need something more structurally for this. I have the same issue with the MacOS change for #4370. Maybe we can get some CI-job to automatically generate this for us.

source/Fleet.cpp
petervdmeer: Yes, line 548 gives chosen a range from 0 to `total`.
But the if-statement line 549 makes sure that line 551 only gets executed if the actually chosen number on 548 was in the range of 0 to `stockTotal`

source/Fleet.cpp
petervdmeer: I wasn't just talking about the normal cases. What if somebody makes a mistake in mission files and uses multiple remove of variants to get to the situation where we do have a `fleet` where both `variants` and `stockVariants` are empty?

Are we really sure that such situation cannot happen? And if we are not sure and it does happen, would the code then just result in placing an empty fleet or could we expect some sort of crash then? (The empty fleet placement in such case would be okay for me, the game crashing would not be okay for me.)

source/Variant.h
petervdmeer: Okay, good to hear this is not a copy-paste issue but a deliberate decision.

EndlessSky.cbp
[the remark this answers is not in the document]
petervdmeer: I now have a structural solution, #4817 implements a CI-job that automatically generates the diff to apply to EndlessSky.xcodeproj/project.pbxproj . I used the script in the CI-job to generate the Xcode changes for #4370 (and that seems to build fine).

source/Fleet.cpp
Amazinite: Looks like Fleet::Enter and Fleet::Place already handle the case of having no variants or having variants with no ships by returning if there are no variants or if the variant that is chosen returns no ships.

data/remnant.txt
Amazinite: I'm not against the idea of renaming the GameData (since it's just a matter of changing 3 lines of code, and the functionality is what I desire, so I'm not going to quibble over the name), but I'd be interested to hear if others think that a name change is appropriate.

Zitchas:
So, hopefully not crashing the game anymore? Is this likely to get a 0.9.13 tag?

Amazinite:
It hasn't been crashing for a while. As far as merging goes, I'm actually less confident about this than #4931. At least with that PR it's an entirely new mechanic to the game, so if the mechanic itself is broken is some way then that doesn't seem like too big an issue to me, but this PR massively reworks an existing mechanic that the game has been built on expecting to work, meaning if something is broken here then that's worse, so I'd really appreciate reviews on this one from @petervdmeer and @tehhowch before merging.

Zitchas:
Fair enough.

source/Variant.cpp
```diff
@@ -0,0 +1,356 @@
...
+	// If this variant is global (i.e. a root node variant stored in GameData),
+	// it must be named.
+	if(global && node.Size() < 2)
+	{
+		node.PrintTrace("No name specified for variant:");
+		return;
+	}
+	if(node.Size() >= 2 && !node.IsNumber(1))
+		name = node.Token(1);
+	else if(global && node.IsNumber(1))
```
petervdmeer, line 47: If this check (IsNumber) and the check on line 40 are moved to before the variant-load in GameData, then there would no longer be a need to give the global-variable as parameter and it would be possible to avoid the load completely if the name is missing.

source/Fleet.h
petervdmeer: The comment adds quite some clarity, but it still sounded a little bit confusing on what stockVariants really is.
I think something like the following text is even more clear:
```
// StockVariants contains references to the named root-node variants (as defined
// in GameData) that are referenced from this fleet definition. The variants vector
// contains the local variants that were defined by this fleet definition.
```

source/Fleet.cpp
```diff
@@ -244,12 +280,13 @@ const Government *Fleet::GetGovernment() const
 // Choose a fleet to be created during flight, and have it enter the system via jump or planetary departure.
 void Fleet::Enter(const System &system, list<shared_ptr<Ship>> &ships, const Planet *planet) const
 {
-	if(!total || variants.empty())
+	if(!total || (variants.empty() && stockVariants.empty()))
```
petervdmeer, line 283: should this be `if((!total || variants.empty()) && (!stockTotal || stockVariants.empty()))`?
Total and variants non-empty seems needed to select from local variants.
stockTotal and stockVariants non-empty seems needed to select from stock variants.

source/Fleet.cpp
```diff
@@ -558,10 +562,10 @@ pair<Point, double> Fleet::ChooseCenter(const System &system)
 
 
 
-vector<shared_ptr<Ship>> Fleet::Instantiate(const Variant &variant) const
+vector<shared_ptr<Ship>> Fleet::Instantiate(vector<const Ship *> &variantShips) const
```
petervdmeer, line 565: Can we change the variable name variantShips here to just ships? (This function does not seem to have any variant-specific code anymore.)

source/Fleet.cpp
```diff
@@ -204,21 +205,28 @@ void Fleet::Load(const DataNode &node)
...
+			string variantName;
+			if(child.Size() >= 2 + add && !child.IsNumber(1 + add))
+				variantName = child.Token(1 + add);
+			if(child.Size() >= 2 + add + !variantName.empty())
+				weight = child.Value(1 + add + !variantName.empty());
+			
+			if(!variantName.empty())
+				variants.emplace_back(GameData::Variants().Get(variantName), weight);
+			else
+				variants.emplace_back(child, weight);
```
quyykk, line 220: Would it be a good idea to store a `Variant` inside a fleet and then use that to load a `variant` node. Would avoid duplicating logic between a `Fleet` and a `Variant`.

source/WeightedList.h
```diff
@@ -86,4 +92,14 @@ const Type &WeightedList<Type>::Get() const
 
 
 
+template<class Type>
+void WeightedList<Type>::RecalculateWeight()
+{
+	total = 0;
+	for(const auto &choice : choices)
+		total += choice.Weight();
```
quyykk, line 100: Since this function is only used when erasing, you can simplify the implementation by reducing `total` by the weight of the element to remove. That way you avoid iterating through the whole list.

source/Variant.cpp
```diff
@@ -0,0 +1,249 @@
...
+			if(variant)
+			{
+				string variantName;
+				if(child.Size() >= 2 + add && !child.IsNumber(1 + add))
+				{
+					variantName = child.Token(1 + add);
+					if(variantName == name)
+					{
+						node.PrintTrace("A variant cannot reference itself:");
+						return;
```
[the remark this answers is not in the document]
Amazinite: What gave you that idea?
image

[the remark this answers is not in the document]
quyykk: Sorry. I mean the `return`. I never saw a `PrinttTrace` and then a `return` the parsing always continues on (ignoring the error) if that makes sense.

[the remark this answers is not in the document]
Amazinite: Oh, I see what you mean. Yeah, this should be `continue;`.

source/Variant.cpp
```diff
@@ -0,0 +1,255 @@
...
+				// If given a full definition of one of this fleet's variant members, remove the variant.
+				Variant toRemove(child);
+				auto VariantToRemove = [&](const WeightedVariant &v) noexcept -> bool
+				{
+					return v.Get() == toRemove;
+				};
+				
+				auto removeIt = find_if(variants.begin(), variants.end(), VariantToRemove);
+				if(removeIt != variants.end())
+					variants.erase(remove_if(removeIt, variants.end(), VariantToRemove), variants.end());
```
quyykk, line 72: You don't need the `find_if`. You can store the iterator returned by `remove_if` to check whether any elements were removed. Same thing below.

source/WeightedVariant.cpp
```diff
@@ -0,0 +1,59 @@
...
+bool WeightedVariant::operator==(const WeightedVariant &other) const
+{
+	return other.weight == weight && other.Get() == Get();
+}
+
+
+
+bool WeightedVariant::operator!=(const WeightedVariant &other) const
+{
+	return other.weight != weight || other.Get() != Get();
```
quyykk, line 58: `!(*this == other)` is more maintainable.

Zitchas:
I finally came up with a situation where I really want to use these variants. Adding fleets that use the Broadsider, of all things. I have three different variants of the Carrier that I want to show up, and the fleets are basically identical except for which Carrier is selected.

Which raises a question: Can events add additional options within the nested variant?

For example, if one has:
```
	fleet "Large Oathkeeper"
		variant 2
			variant
				"Carrier (Mark IIIa)"
				"Carrier (Mark IIIb)"
			"Lance" 4
			"Combat Drone" 6
			```
How would one specify an event to add an additional "carrier (mark IIIc) into that inner variant?

**Comment by Amazinite** [comment]
Events can only add or remove variants from fleets. They can not directly modify existing variants. (This is already the case.) But named variants can be modified by events. So if you wanted to go from 
```
fleet "Large Oathkeeper"
	variant 2
		variant
			"Carrier (Mark IIIa)"
			"Carrier (Mark IIIb)"
		"Lance" 4
		"Combat Drone" 6
```
to
```
fleet "Large Oathkeeper"
	variant 2
		variant
			"Carrier (Mark IIIa)"
			"Carrier (Mark IIIb)"
			"Carrier (Mark IIIc)"
		"Lance" 4
		"Combat Drone" 6
```
You have two options:
1. Remove the existing variant and add the new one.
```
event "update fleets"
	fleet "Large Oathkeeper"
		remove variant
			variant
				"Carrier (Mark IIIa)"
				"Carrier (Mark IIIb)"
			"Lance" 4
			"Combat Drone" 6
		add variant 2
			variant
				"Carrier (Mark IIIa)"
				"Carrier (Mark IIIb)"
				"Carrier (Mark IIIc)"
			"Lance" 4
			"Combat Drone" 6
```
2. Define a named variant that you can modify through an event.
```
variant "Carriers"
	"Carrier (Mark IIIa)"
	"Carrier (Mark IIIb)"
	
fleet "Large Oathkeeper"
	variant 2
		variant "Carriers"
		"Lance" 4
		"Combat Drone" 6

event "new carrier"
	variant "Carriers"
		add "Carrier (MarkIIIc)"
```

**Comment by Zitchas** [comment]
Cool. Thanks for the explanation and examples!

File: source/Variant.cpp [hunk 27]
```diff
@@ -0,0 +1,265 @@
...
+			if(variant)
+			{
+				// If given a full definition of one of this fleet's variant members, remove the variant.
+				Variant toRemove(child);
+				auto VariantToRemove = [&](const WeightedVariant &v) noexcept -> bool
+				{
+					return v.Get() == toRemove;
+				};
+				
+				auto removeIt = remove_if(variants.begin(), variants.end(), VariantToRemove);
```
quyykk, line 71: `remove_if` removes every variant that matches, while the old code only removes one instance. That's my bad for recommending `remove_if` without checking what the code did, sorry. While it doesn't make sense to have duplicate variants (since we have weights), this should probably be changed so that a single element is removed  to avoid breaking backwards compatibility.

source/Variant.cpp
```diff
@@ -0,0 +1,265 @@
...
+					}
+				}
+				
+				if(child.Size() >= index + 1 && child.Value(index) >= 1.)
+					n = child.Value(index);
+				
+				// If this variant is named, then look for it in GameData.
+				// Otherwise this is a new variant definition only for this variant.
+				if(!variantName.empty())
+					variants.emplace_back(GameData::Variants().Get(variantName), n);
```
quyykk, line 125: It would be useful if we get a warning for the following:

```
fleet "Large Oathkeeper"
	variant
		variant "test"
			"Carrier (Mark IIIa)"
			"Carrier (Mark IIIb)"
                        # ^ the children are ignored since the variant named "test" is used
```

source/Variant.cpp
Amazinite: Honestly, since having the exact same variant defined twice in a fleet is likely a content error to begin with, I'm not really worried about slightly changing how removal of variants from a fleet functions to remove all matching variants. From a content creator perspective, I'd expect my attempt to remove a variant to remove it entirely, not just the first instance of it if I accidentally have two.

(There's a similar scenario with adding/removing fleet spawns from a system where you can have the same fleet spawn with two difference periods, and using `remove fleet <name>` only removes the first instance, where I'd prefer that it removes all instances. That's another content error situation, though, having the same fleet twice, so I'm not worried about changing that behavior. Changing it here was more of a happy accident.)

If others think that the old behavior should be kept, though, then I'll revert to it.

EjoThims:
This is just for fleets, right?

Not ship variants?

Amazinite:
Correct.

quyykk:
Is there anything left to do derpy? At least IMO this PR looks good to go in its current state.

Amazinite:
Nothing I can think of, although I often go over my own code every so often hunting for any potential errors or improvements and it's been a while since I've looked this over, so maybe there's something that could be patched up. In any case, I mainly closed this because there doesn't seem to be an active push to get it merged and I have no particular need to get it merged immediately or even any time soon (it is over two years old as is already, what's another few months), although this is definitely closer to completion than my other PR that I closed (#5977), so perhaps I could just leave this open. I'll handle the merge conflict later.

EndlessSkyLib.cbp
```diff
@@ -330,13 +330,18 @@
...
 		<Unit filename="source/Visual.cpp" />
 		<Unit filename="source/Visual.h" />
 		<Unit filename="source/Weapon.cpp" />
 		<Unit filename="source/Weapon.h" />
 		<Unit filename="source/Weather.cpp" />
 		<Unit filename="source/Weather.h" />
 		<Unit filename="source/WeightedList.h" />
+		<Unit filename="source/WeightedVariant.cpp" />
+		<Unit filename="source/WeightedVariant.h" />
+		<Unit filename="source/gl_header.h" />
```
Amazinite, line 344: Where'd this come from?

quyykk: You made a mistake while merging I think. `gl_header.h` was the old name for `opengl.h`.

Amazinite:
<details><summary>Ran the following tests:</summary>

Spawned this fleet 100 times per test and checked which ships spawned and how many to see if it was working as intended. Only testing a normal fleet right now without any nested variants.
```
fleet "test fleet"
	government "Merchant"
	names "civilian"
	variant 1
		"Shuttle"
	variant 1
		"Sparrow"
	variant 1
		"Star Barge"
```

Changes tested individually. After each change was made, rolled back to the original fleet.
The following tests look good:
```
# Removing a variant.
fleet "test fleet"
	remove variant
		"Shuttle"

# Adding a variant.
fleet "test fleet"
	add variant 1
		"Flivver"

# Removing and adding at the same time.
fleet "test fleet"
	remove variant
		"Shuttle"
	add variant 1
		"Flivver"

# Redefining.
fleet "test fleet"
	variant 1
		"Flivver"

# Adding a duplicate variant.
fleet "test fleet"
	add variant 1
		"Shuttle"

# Removing a variant doesn't care about the variant weight.
fleet "test fleet"
	remove variant 2
		"Shuttle"
```

The following test didn't work:
```
# Adding a duplicate variant then removing both with a single remove call.
# ISSUE: Still had Shuttles spawning. This occurred with adding and removing in the same change and having a separate change for the removal after the addition. Adding a second `remove variant` only threw an error about not being able to find a matching variant yet Shuttles continued to be spawned.
fleet "test fleet"
	add variant 1
		"Shuttle"
	remove variant
		"Shuttle"
```

</details>

Amazinite:
Looks like the issue was that the line `add variant 1` was being sent to Variant::Load, which was looking for a name at index 1, setting the name of the variant to "variant." Then when attempting to remove the variant with the Shuttle, the equality operator saw that one of the variants was named and one was not, meaning the added variant didn't match what was being removed.

Accounting for the shift in index when adding a variant fixed the issue, so all the tests described above now work as intended.

Amazinite:
Testing this one now.
```
variant "test variant"
	"Shuttle"
	"Sparrow"
	"Star Barge"

fleet "test fleet"
	government "Merchant"
	names "civilian"
	variant 3
		variant "test variant"
	variant 1
		Flivver
```
The tests that work:
```
# Adding a ship to named variant.
variant "test variant"
	add Scout

# Removing a ship from a named variant.
variant "test variant"
	remove Shuttle

# Adding then removing a duplicate.
variant "test variant"
	add Shuttle
	remove Shuttle

# Redefining.
variant "test variant"
	Flivver
```
The one that doesn't:
```
# Removing a named variant from a fleet.
# ISSUE: The game crashes when the mission loads.
fleet "test fleet"
	remove variant
		variant "test variant"
```

source/Variant.cpp
```diff
@@ -0,0 +1,279 @@
...
+}
+
+
+
+// Choose a list of ships from this variant. All ships from the ships
+// vector are chosen, as well as a random selection of ships from any
+// nested variants in the WeightedList.
+vector<const Ship *> Variant::ChooseShips() const
+{
+	vector<const Ship *> chosenShips = ships;
```
Amazinite, line 191: So somehow things are crashing on this line, or about here. If I put a Files::LogError above and below this, sometimes I only get the line above this printing.
```
# No ships in this variant?
Fleet::Place test fleet
1 # Number of variants to choose from
	Variant::ChooseShips
	0 # Number of chosenShips
	0 # Number of nested variants
	0 # Number of ships returned
0 # Number of ships returned
# Then, with only one variant to choose from, there's somehow a ship now.
Fleet::Place test fleet
1
	Variant::ChooseShips
	1
	0
	1
1
# Then back to no ships?
Fleet::Place test fleet
1
	Variant::ChooseShips
	0
	0
	0
0
Fleet::Place test fleet
1
	Variant::ChooseShips
# Crashes
```

Amazinite: It's accessing trash data. The variant being removed has a weight of 3 and the remaining variant has a weight of 1, yet after the removal of the variant with a nested variant within it, the number of variants left correctly drops to 1 but the total weight is only changing to 3. No idea why, though.

Amazinite: So either there's something wrong with the range erase of WeightedList or it has something to do with using remove_if, because I switched back to using the older syntax (see below) and it works fine and properly adjusts the weight.
```c++
bool didRemove = false;
Variant toRemove(child, true);
for(auto it = variants.begin(); it != variants.end(); ++it)
	if(it->Get() == toRemove)
	{
		it = variants.eraseAt(it);
		didRemove = true;
	}

if(!didRemove)
	child.PrintTrace("Warning: Did not find matching variant for specified operation:");
```

Amazinite: Alright, it'd definitely the use of the remove_if at fault somehow (and likely something that WeightedList should be doing that it isn't).
If I order the variants like this it crashes:
```
fleet "test fleet"
	government "Merchant"
	names "civilian"
	variant 3
		variant "test variant"
	variant 1
		Flivver
```
But if I order them like this it's fine and works as expected:
```
fleet "test fleet"
	government "Merchant"
	names "civilian"
	variant 1
		Flivver
	variant 3
		variant "test variant"
```
To my understanding remove_if is suppose to push the objects to be removed to the end of the list, but the list doesn't seem like it's actually getting modified, yet it's still erasing the last object in the list since it knows that one object is suppose to be erased.

Amazinite:
Alright then. Anything else we could test that I might have missed?

source/UniverseObjects.cpp
```diff
@@ -217,6 +223,10 @@ void UniverseObjects::CheckReferences()
 	for(auto &&it : effects)
 		if(it.second.Name().empty())
 			NameAndWarn("effect", it);
+	// Variants are not serialized. Any changes via events are written as DataNodes and thus self-define.
+	for(auto &&it : variants)
+		if(!it.second.IsValid() && !deferred["variant"].count(it.first))
```
quyykk, line 228: This should check whether the variant is unnamed, not whether it is invalid. Or else the warning doesn't make sense.

Amazinite: This is copied from Fleet's behavior just below this. If this needed changed then would not Fleet also need changed?

quyykk: Oh yeah I can see why Fleet (and Variant) use IsValid now. I guess it does make sense for them. sorry

Amazinite:
@quyykk Want to give this a final review pass? I think I've squeezed everything out of this that I can at this point.

quyykk (commented):
Just a small thing I noticed. I think it's all good now, but I'm not feeling all too confident right now :sweat_smile:

tehhowch (changes requested):
The difference of behavior based on the containing class suggests that there is a missing part in the class design. Conceptually, we want to wrap the data in one class when the parent is a `Fleet`, and a different class when the parent is a `Fleet::Variant` instance. The wrinkle arises due to the addition of "named variants," which can appear in either or both places. The solution here is a step towards the separation of load-time data models and play-time "game models," wherein the root-level variant definitions are actually treated as "variant templates." These templates would then be used to instantiate the appropriate version of the "variant" when the parent class is finalized by a `FinishLoading()` call.


Some of these review comments won't be applicable with the fixed class relationship, but they're useful to share anyway.

tehhowch:
As a side note, I think if you wanted to ship a subset of this PR (i.e. no root-level named variants that could be either a fleet::variant or a "variant variant"), I think that could be much more straightforward.

source/Variant.cpp
```diff
@@ -0,0 +1,306 @@
...
+}
+
+
+
+// An operator for checking the equality of two variants.
+bool Variant::operator==(const Variant &other) const
+{
+	// Is either variant named? Do they share the same name?
+	if(!other.name.empty() || !name.empty())
+		return other.name == name;
```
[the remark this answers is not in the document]
Amazinite: Variants of the same name aren't always the same object, at least how I'm doing things now. If a variant is being removed then a new variant is created, and this new variant could share a name with an existing root level variant yet not be the same object.

```
# Root level variant.
variant "remove me"
	"Shuttle"

fleet "remove from me"
	...
	variant "remove me" 1 # Gets stored as a pointer to the root level.

event "remove from fleet"
	fleet
		remove variant "remove me" # Creates a new variant with the name "remove me".
```
This simplifies variant removal by not needing the class that is being removed from to check whether the variant being removed is named or not, as is required to be done when adding a variant.

source/Variant.cpp
```diff
@@ -0,0 +1,306 @@
...
+	if(name.empty())
+		return;
+	
+	// Prevent a named variant from containing itself.
+	for(auto it = variants.begin(); it != variants.end(); )
+	{
+		if(it->Get().NestedInSelf(name))
+		{
+			it = variants.eraseAt(it);
+			Files::LogError("Error: Infinite loop detected and removed in variant \"" + name + "\".");
```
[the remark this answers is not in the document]
Amazinite: > What's stopping this from being done in load?

I was probably spooked by potential load order scenarios, but I believe that no matter the load order, any loop will eventually get caught.

Suppose variant A contains variant B, which contains variant C, which contains variant A.
If variant A loads first it would be checking if variant B contains itself somewhere, yet neither variant B nor variant C have been loaded. That's the scenario that had me making this method. But then once variant C loads, it will see that it contains itself through variant A and remove variant A, so the loop does eventually get broken.

source/Variant.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+
+// Choose a ship from this variant given that it is a nested variant.
+// Nested variants only choose a single ship from among their list
+// of ships and variants.
+const Ship *Variant::NestedChooseShip() const
+{
+	int total = ships.size() + variants.TotalWeight();
+	// Randomly choose between the ships and the variants.
+	if(static_cast<int>(Random::Int(total)) < static_cast<int>(variants.TotalWeight()))
```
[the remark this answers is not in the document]
Amazinite: Random::Int returns a uint32_t and WeightedVariant::TotalWeight returns a size_t. If I recall correctly my compiler threw an error on comparing those two, hence why I casted them both to ints.

Amazinite:
> As a side note, I think if you wanted to ship a subset of this PR

Yeah, I've thought of doing that. So an appropriate subset would be making Variant its own class that Fleet holds onto using a WeightedList, and variant would remain just a list of ships?

source/WeightedVariant.h
```diff
@@ -0,0 +1,43 @@
...
+// A weighted variant consists of a variant and a weight to be used by Fleet
+// or Variant in a WeightedList.
+class WeightedVariant {
+public:
+	WeightedVariant() = default;
+	WeightedVariant(Variant variant, int weight);
+	WeightedVariant(const Variant *stockVariant, int weight);
+
+	const Variant &Get() const;
+	int Weight() const;
```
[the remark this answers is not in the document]
Amazinite: How would we erase from a two-vector WeightedList?

source/Variant.cpp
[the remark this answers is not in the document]
Amazinite: I spotted that the only other place a Random::Int call is directly compared against anything that isn't another Random::Int call, at least so far as I could find, was a comparison between Random::Int and explosionCount in Ship.cpp. Since explosionCount is an `unsigned`, I've changed this to store variants.TotalWeight() in an `unsigned` variable.

Amazinite:
* Got to a number of the review comments with the latest commits, including replacing WeightedVariant with WeightedUnionItem.
* Removed the ability for a named variant to be defined as a first-level variant under a fleet. Named variant can now only ever be utilized as nested variants. This simplifies Fleet, and we don't really use anything since having a named variant as a first-level variant of a fleet would admittedly be a strange place to use such a variant, since the main intention of named variant is for them to be nested.

Amazinite:
Going back to a draft again since I broke another piece of this PR out into its own: #6835.

Amazinite:
> The difference of behavior based on the containing class suggests that there is a missing part in the class design.

How about this?
* Have a Variant superclass that defines the shared methods of all Variants (Load, NestedInSelf, equality operators, and virtual versions of ChooseShips, IsValid, and Strength.
* Have a FleetVariant subclass that extends Variant and contains a list of NestedVariants.
* Have a NestedVariant subclass that extends Variant and can be either named (i.e. a root object) or defined as a child of a FleetVariant. Also contains a list of NestedVariants.
* Each of subclasses would define ChooseShips, IsValid, and Strength.
* The only thing that would have a list of FleetVariants would be Fleet itself.

source/NestedVariant.cpp
```diff
@@ -0,0 +1,242 @@
...
+		if((add || remove) && child.Size() == 1)
+		{
+			child.PrintTrace("Skipping invalid \"" + child.Token(0) + "\" tag:");
+			continue;
+		}
+		bool variant = (child.Token(add || remove) == "variant");
+
+		if(remove)
+		{
+			if(variant)
```
tehhowch, line 70:
```suggestion
			if(variant && !variants.empty())
```

source/NestedVariant.cpp
```diff
@@ -0,0 +1,242 @@
...
+			{
+				// If given a full definition of a nested variant, remove all instances of that
+				// nested variant from this variant.
+				auto removeIt = std::remove(variants.begin(), variants.end(), UnionItem<NestedVariant>(child));
+				if(removeIt != variants.end())
+					variants.erase(removeIt, variants.end());
+				else
+					child.PrintTrace("Warning: Did not find matching variant for specified operation:");
+			}
+			else
```
tehhowch, line 80:
```suggestion
			else if(!ships.empty())
```

tehhowch (changes requested):
Didn't dive into the FleetVariant/NestedVariant too much.

(Consider moving the "UnionItem" out to its own PR; it is useful beyond just this functionality)

source/UnionItem.h
```diff
@@ -0,0 +1,38 @@
...
+
+// A UnionItem is an object stored as either a pointer or a stock
+// object, but not both.
+template <class Type>
+class UnionItem {
+public:
+	UnionItem() = default;
+	UnionItem(Type &&item) : item(item) {}
+	UnionItem(const Type *item) : stockItem(item) {}
+	const Type &GetItem() const noexcept { return stockItem ? *stockItem : item; }
```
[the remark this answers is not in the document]
Amazinite: I like it.

source/UniverseObjects.cpp
```diff
@@ -386,6 +392,16 @@ void UniverseObjects::LoadFile(const string &path, bool debugMode)
 			testDataSets.Get(node.Token(1))->Load(node, path);
 		else if(key == "trade")
 			trade.Load(node);
+		else if(key == "variant" && node.Size() >= 2)
+		{
+			// Variant names cannot be numbers, as Variants need
+			// to be able to distinguish between a given Variant's
+			// name or weight, which is a number.
```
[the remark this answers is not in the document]
Amazinite: Yes.
```
fleet
	variant 1 # Fleet variant, never named
		"Cruiser"
		variant 2 # Nested variant, unnamed, number in place of where name would be found.
			"Gunboat"
			"Rainmaker"
		variant "Navy Drones" 2 # Nested variant, named, name can't be a number.
```

source/Fleet.h
```diff
@@ -18,6 +18,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Variant.h"
 #include "WeightedList.h"
 
+#include <cstdint>
```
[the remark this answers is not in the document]
Amazinite: I believe you commented that Variant needed this since Strength returns int64_t. I spotted that Fleet also has a Strength method that returns int64_t, but it lacked this include.

warp-core:
Is this actually ready for review now? Are all the things split off from this merged?

source/NestedVariant.h
```diff
@@ -0,0 +1,69 @@
+/* NestedVariant.h
+Copyright (c) 2022 by Amazinite
+
+Endless Sky is free software: you can redistribute it and/or modify it under the
+terms of the GNU General Public License as published by the Free Software
+Foundation, either version 3 of the License, or (at your option) any later version.
+
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
```
tibetiroka, line 10:
```suggestion
PARTICULAR PURPOSE. See the GNU General Public License for more details.

You should have received a copy of the GNU General Public License along with
this program. If not, see <https://www.gnu.org/licenses/>.
```

Amazinite:
Not orphaning this, just not working on it at the moment and it isn't too important a feature.

1010todd:
I know this isn't being worked on at the moment but I'll comment now so I don't forget to do it later.

I think it should be called something else because people already seems to be confused by fleet variants and ship variants. Adding another variant especially with similar syntax to ship variant but acts differently sounds like it'll just cause more confusion. 

For new name for this I don't have any good idea right now.

Issue #4695, opened 2019-12-02
Log entry for "Remnant: Technology Available" has different point-of-view


AmkG:
**Describe the bug**
The log entry for "Remnant: Technology Available" has a different point-of-view compared to other log entries:

https://github.com/endless-sky/endless-sky/blob/deda255028f20fabb6ae814cf55f0eb1c2b0c3e1/data/remnant%20missions.txt#L579

For example:

https://github.com/endless-sky/endless-sky/blob/deda255028f20fabb6ae814cf55f0eb1c2b0c3e1/data/remnant%20missions.txt#L648

**To Reproduce**

1. Complete Remnant missions until you have been offered Remnant license.
2.  Look at logbook for the entry.

**Expected behavior**
The log entry should be from point-of-view of player, starting with a verb phrase with the player/writer as the implied subject.

> Considered a valuable friend of the Remnant after prefects have reached a consensus.  Was given permission to purchase their ships and technology.

**Screenshots**

**System (please complete the following information):**
 - OS: Debian 9
 - Game Source: self-compiled
 - Version: deda255028f20fabb6ae814cf55f0eb1c2b0c3e1

**Link to save file**
inapplicable

**Additional context**
inapplicable

Amazinite:
Fixed by merging #4706.

PR #4706, opened 2019-12-07
Fix remnant tech available log entry

Zitchas:
**Bugfix:** This PR addresses issue #4695 

## Fix Details
This change brings this log entry inline with the prevailing style for <first> <last>'s personal thoughts and entries; namely speaking in first person while removing all the references to oneself, including "I" or "me". It is an odd style, but that is what we've got.

## Specifics:
Replaces the current log entry with:
`"The prefects have reached a consensus. Shall be considered a valuable friend of the Remnant, and given permission to purchase their ships and technology."`


tehhowch:
This doesn't yet remove the original log...

Zitchas:
Fixed. Thanks for spotting that.

data/remnant missions.txt
```diff
@@ -576,7 +576,8 @@ mission "Remnant: Technology Available"
 	on offer
 		payment 2000000
 		set "license: Remnant"
-		log "The prefects have reached a consensus that you should be considered a valuable friend of the Remnant, and shall be given permission to purchase their ships and technology."
+		log "The prefects have reached a consensus: Shall be considered a valuable friend of the Remnant, and given permission to purchase their ships and technology."
```
Amazinite:
```suggestion
		log "The prefects have reached a consensus: have been considered a valuable friend of the Remnant, and was given permission to purchase their ships and technology."
```

Zitchas: That sounds good, but "have been considered" sounds like <first> was a valuable friend, past tense. As in, no longer is a one.

Amazinite: "Shall be" is future tense though, when the truth is that you're considered a valuable friend now. And I don't necessarily see "have been" implying no longer. Putting the pronouns in there, it would read "I have been considered a valuable friend," which is different from "I was considered a valuable friend." Both are past tense, but one implying no longer ("was") while the other does not.

jafdy: `Am now`?

Zitchas: I prefer `Am now considered a valuable friend" as that clearly indicates the log is talking about a present state of affairs.

Amazinite: Simply "now considered" (drop the "am") would flow better.

Zitchas: Agreed. I'll make the change as soon as I get Git setup on my machine again.

Zitchas: Changed

Zitchas:
Made the correction. "Shall be" has been switched to "Now"

Issue #4707, opened 2019-12-07
Gun mentioned in Intro mission for shuttle craft


diggorygray:
In the Intro mission "Intro [2 Transport]", removing the gun on your ship is mentioned.
But the stock shuttle has now gun.

Text from "intro missions.txt" :

"Great," he says. "This is Chuck, and Sarah, and their son Carl. Let's help them carry their luggage to your ship." As you begin walking back, slightly ahead of them, James whispers to you, "Tourists are always a good way to make money. They're on vacation, so they don't mind paying a bit extra. But no point in your cargo space going to waste, so take a look at the job board and see if there are any delivery missions we can run at the same time. **And if you want to be a little safer should we come across any pirates, sell off your gun.** It might seem counterintuitive, but being unarmed makes you less of a target to pirates if there are others in the system that they could be defending themselves against."`

**To Reproduce**
1. Start a new pilot, choosing a shuttle craft
2. Accept the intro missions.
3. When reaching the point of the transport family from 'New China' you'll see the gun refance.

**Expected behavior**
Either no mention of a gun, or mention not arming a small ships with guns or the stock armament of a shuttle to include a blaster.
 
**Screenshots**

**System (please complete the following information):**
 - OS: Mint Linux 19.2
 - Game Source: self-compiled
 - Version: commit hash 409c9af

Zitchas:
Just to clarify, you are pointing out that the stock shuttle currently does not have a gun, and thus it does not make sense that James tells you to sell your gun, right?

So we should either:
a) give the shuttle a blaster again,
b) change the dialogue so that James is recommending that you do not buy a gun, or
c) remove that part of the dialogue entirely.

Personally, A or B would be much better than C.

diggorygray:
That is the issue in a nutshell :).

diggorygray:
Yes option C removes some of the richness of the game.

Zitchas:
OK, we now have two PRs for fixing this. One for option A, one for option B. I'll close whichever of the two isn't accepted.

diggorygray:
Great, thanks.

Zitchas:
Your welcome. Thanks for filing an issue for it.

PR #4708, opened 2019-12-07
Fix: James's intro mission - A: Give the Shuttle an Energy Blaster

Zitchas:
**Bugfix:** This PR addresses issue #4707 

## Fix Details
Jame's transport intro mission includes James giving the player instructions that "if they want to be safer, they should sell their gun." There are three options for resolving this:
A) give the shuttle a blaster
B) change Jame's dialogue to specify that the player should *avoid* buying a gun, rather than telling them to sell the one they have.
C) remove that section of dialogue entirely.

This PR is option A.

Note: Only one of option A or B should be implemented. When one is merged, the other should be closed. Option [B can be found here.](https://github.com/endless-sky/endless-sky/pull/4709)

### DIFF
--- a/data/ships.txt
+++ b/data/ships.txt
@@ -2718,6 +2718,7 @@ ship "Shuttle"
 			"hull damage" 60
 			"hit force" 180
 	outfits
+		"Energy Blaster"
 		"nGVF-AA Fuel Cell"
 		"LP036a Battery Pack"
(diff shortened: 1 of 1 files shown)


Amazinite:
Given that having a gun is exactly what James is warning against, I don't know if we want to give the Shuttle a gun by default to make yourself more of a target before James tells you to sell it.

PR #4709, opened 2019-12-07
Fix: James's intro mission - B: Change James' dialogue

Zitchas:
**Bugfix:** This PR addresses issue #4707 

## Fix Details
Jame's transport intro mission includes James giving the player instructions that "if they want to be safer, they should sell their gun." There are three options for resolving this:
A) give the shuttle a blaster
B) change Jame's dialogue to specify that the player should *avoid* buying a gun, rather than telling them to sell the one they have.
C) remove that section of dialogue entirely.

This PR is option B.

Note: Only one of option A or B should be implemented. When one is merged, the other should be closed.  Option [A can be found here.](https://github.com/endless-sky/endless-sky/pull/4708)

### DIFF
--- a/data/intro missions.txt
+++ b/data/intro missions.txt
@@ -135,7 +135,7 @@ mission "Intro [2 Transport]"
 				`	"Sure!"`
 				`	"No thanks, I think I'll strike off on my own now."`
 					decline
-			`	"Great," he says. "This is Chuck, and Sarah, and their son Carl. Let's help them carry their luggage to your ship." As you begin walking back, slightly ahead of them, James whispers to you, "Tourists are always a good way to make money. They're on vacation, so they don't mind paying a bit extra. But no point in your cargo space going to waste, so take a look at the job board and see if there are any delivery missions we can run at the same time. And if you want to be a little safer should we come across any pirates, keep your weapon hardpoints empty. It might seem counterintuitive, but being unarmed makes you less of a target to pirates if there are others in the system that they could be defending themselves against."`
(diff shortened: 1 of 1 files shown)


data/intro missions.txt
```diff
@@ -135,7 +135,7 @@ mission "Intro [2 Transport]"
...
+			`	"Great," he says. "This is Chuck, and Sarah, and their son Carl. Let's help them carry their luggage to your ship." As you begin walking back, slightly ahead of them, James whispers to you, "Tourists are always a good way to make money. They're on vacation, so they don't mind paying a bit extra. But no point in your cargo space going to waste, so take a look at the job board and see if there are any delivery missions we can run at the same time. And if you want to be a little safer should we come across any pirates, keep your weapon hardpoints empty. It might seem counterintuitive, but being unarmed makes you less of a target to pirates if there are others in the system that they could be defending themselves against."`
```
Amazinite: When I wrote this, I was more thinking of James speaking in the abstract than referring to your ship in this very moment, so I don't exactly see an issue with the text as-is, but I'm not against changing it for clarity.
Instead of referring to hardpoints, maybe just stick with the "sell your guns" phrasing.
```suggestion
			`	"Great," he says. "This is Chuck, and Sarah, and their son Carl. Let's help them carry their luggage to your ship." As you begin walking back, slightly ahead of them, James whispers to you, "Tourists are always a good way to make money. They're on vacation, so they don't mind paying a bit extra. But no point in your cargo space going to waste, so take a look at the job board and see if there are any delivery missions we can run at the same time. And if you want to be a little safer should we come across any pirates, sell off any guns you might have. It might seem counterintuitive, but being unarmed makes you less of a target to pirates if there are others in the system that they could be defending themselves against."`
```

PR #4710, opened 2019-12-08
fix: Make gather command work for out-of-system fleet

### LINKED ISSUE #3774: Cannot "gather" when whole fleet is in other system
I used to use the gather command to get my fleet to follow me when they were holding position in another system. Now, pressing G does nothing unless I already have at least one escort in the system, and I have to right-click to send them to a specific location in the current system, then wait for them to arrive before I can issue the gather command.

petervdmeer:
**Bugfix:** This PR addresses issue #3774

## Fix Details
This PR fixes #3774 by checking if a target is valid based on the flagships location instead of the escorts location.
The check for order cancellation is also extended by the flagship location, but the escorts own location is also kept there to allow for situations where the player jumps out of a system and back in.

This PR also keeps the intended fix from #2984 by by ensuring that the cancel order is always given, also (or especially) when the target-ship is no longer in the (flaghips)system.

This PR also allows players to give gather, attack and keep-station orders to escorts in other systems as long as the target is in the same system as the flagship (which is in line with tehhowchs comments in #2984).

The situation where a target is no longer valid gives a different status-message which should help to avoid confusion for players on why a certain order is not followed up.

## Save File
Needs a game where all escorts are in a different system (and holding position) and then get the order to gather. This is a very common situation, but I can provide a savegame if required.


source/AI.cpp
```diff
@@ -3686,17 +3687,27 @@ void AI::IssueOrders(const PlayerInfo &player, const Orders &newOrders, const st
...
 		if(ship == newTarget)
 			continue;
 		
-		// Never issue orders that target a ship in another system.
-		if(newTarget && ship->GetSystem() != newTarget->GetSystem())
-			continue;
-		
 		gaveOrder = true;
+		
+		// If we don't have a valid target, then we cancel the orders for
```
tehhowch, line 3704: Refactor the loop to avoid this. Probably we can do this with a `find_if` lambda.

source/AI.cpp
```diff
@@ -358,7 +358,8 @@ void AI::UpdateKeys(PlayerInfo &player, Command &clickCommands, bool isActive)
 		if(it->second.type & Orders::REQUIRES_TARGET)
 		{
 			shared_ptr<Ship> ship = it->second.target.lock();
-			if(!ship || !ship->IsTargetable() || (it->first->GetSystem() && ship->GetSystem() != it->first->GetSystem())
+			if(!ship || !ship->IsTargetable() || (it->first->GetSystem() && ship->GetSystem() != it->first->GetSystem()
```
tehhowch, line 361: Feel free to make this two checks (ship exists and is targetable; ship is in a targetable system).

petervdmeer: Split into two checks.

source/AI.cpp
petervdmeer: Refactored this by taking it out of the loop.

source/AI.cpp
```diff
@@ -3679,57 +3685,73 @@ void AI::IssueOrders(const PlayerInfo &player, const Orders &newOrders, const st
...
 		{
-			bool shouldReverse = false;
-			// Set the point this ship will "guard," so it can return
-			// to it if knocked away by projectiles / explosions.
-			existing.point = StoppingPoint(*ship, Point(), shouldReverse);
+			// Never issue orders to a ship to target itself.
+			if(ship == newTarget)
+				continue;
+			
+			gaveOrder = true;
```
tehhowch, line 3708: This wasn't quite what I had in mind when I mentioned refactoring the loop. I was expecting that the loop modification would entirely remove the need for the below comment about not having a valid target.

petervdmeer: And that is actually true, since the loop itself is now guarded by if(validTarget). The break statement would never have been executed. (Looks like I forgot to remove some code.)
I added a commit that removes the comment and the conditional break.

petervdmeer:

petervdmeer:
@tehhowch  Could you have a look at what still needs to be changed to this PR to get this merge-able?

source/AI.cpp
```diff
@@ -3737,63 +3743,72 @@ void AI::IssueOrders(const PlayerInfo &player, const Orders &newOrders, const st
...
+
```
Amazinite, line 3767: Extra whitespace that wasn't there before.
```suggestion
```

source/AI.cpp
```diff
@@ -3737,63 +3743,72 @@ void AI::IssueOrders(const PlayerInfo &player, const Orders &newOrders, const st
...
 		Messages::Add(who + description);
 	else
 	{
 		// Clear all the orders for these ships.
-		Messages::Add(who + "no longer " + description);
+		if (!isValidTarget)
+			Messages::Add(who + "unable to and no longer " + description);
+		else
+			Messages::Add(who + "no longer " + description);
+
```
Amazinite, line 3811:
```suggestion
		if(!isValidTarget)
			Messages::Add(who + "unable to and no longer " + description);
		else
			Messages::Add(who + "no longer " + description);
		
```
or
```suggestion
		Messages::Add(who + (!isValidTarget ? "unable to and no longer " : "no longer ") + description);
```

Amazinite (changes requested):
Looks good to me. Just some tiny style changes.

Amazinite (approved):

Issue #3972, opened 2018-10-25
Controller integration


crabctrl:
So I know this has been brought up before, and people didn't really like the idea of it, but I (think I) actually have the skill to write the code, so it's a slightly different ball game now.

That said, I'd love to contribute to the actual game, so it would be great if we could discuss. Currently I have some minimal code [here](http://beancode.org/aldenp/endless-sky/src/branch/joystick-integration) if anybody wants to take a look. I've already talked with some folks on Discord, and we've come up with some controls suggestions, which I'll list off here as well. Any opinions are greatly appreciated.

```
-[back] main menu
[ls] forward
[ls] left
[ls] right
[ls] back
[lt] primary
[rt] secondary
-[dpad] select
[b] land
-board
[x] hail
-scan
[a] jump
-[dpad] nearest
-deploy
-afterburner
-cloak
[start] map
-info


-- Fleet Commands --
fight
gather
hold
ammo
wait
stop
target


Remaining controller inputs:
rs
lb
rb
y
dpad
dpad
```
^^^ This document lists of all the commands I think need to be implemented, as well as available controller inputs. The things with hyphens haven't actually been implemented into the branch I linked yet.

Yet again thanks for your time

MCOfficer:
somewhat related, #3846 (offers a makeshift solution)

atoms118:
Something brought up in #3846 was analog joystick support, which I think would be an interesting venture. On the other hand, I don't really know what kind of refactoring that would take. Simple binary inputs were surprisingly easy to integrate into the code, but analog is going to be a lot more complicated.

Something else I'd like to focus on more than analog is having controls menus. If you look at the above document, it's clear there's nowhere near enough controller inputs to play the game. One solution is having a button which opens an additional menu. The menu would give you more options, eg fleet commands, which you would navigate through using the dpad or right stick.

One question I'm not certain of is should the game be slowed down when the controls menu is open? That could result in cheating, depending on how it's implemented (One solution to cheating is to disable all other controls while the menu is open, but that could cause a poor user experience)

gunqqer:
Slowing time doesn't really mean cheating. You can always hit escape to pause and slowing time wont give a signifigant advantage. Even if it did this is a singleplayer game. You can easilly cheat already (see save file editing in easy to read plaintext) so it's more of up to the player if they want to take advantage of something enabling controller support. (Personally I dont reallly see why you'd want to use a controller but that doesn't mean it's a bad idea)

[...]

Issue #3846, opened 2018-07-14
Joystick Support


VonZorch:
Is there any chance of adding ship control by joystick?

KiLEdEnNis:
I'm nearly sure this will be a no due to the various difficulties to add this feature for the tiny amount of people who would even use it
it's simply not worth it

warp-core:
If Steam supports Joysticks, then you could configure one to be usable through Steam but I don't think it does. It is possible to use a controller via Steam though, if you're interested.

MCOfficer:
there are lots of ways to bind joystick inputs to keyboard inputs, but that will only ever be binary (either on or off), which kind of defeats the purpose of the joystick's main feature (axes with gradial input)

PR #2432, opened 2017-04-20
xcodeproj fixes

mashers:
Some filenames had been changed in the source folder which had not been updated in the xcodeproj, causing linker errors when building the Mac OS X version. This pull request resolves this issue.

Issue #955, opened 2016-04-14
Complete mouse piloting


Amazinite:
Pretty much the exact opposite of #945, allow the player to completely control their stop with only the mouse. 
This was something brought up on the Steam forums. Someone wanted to be able to move, shoot, and target enemy ships entirely with their mouse instead of using the keyboard. 
Proposed controls:
LMB: Fire
RMB: Move to position
Mouse Wheel: swap targets

Obviously you couldn't do everything on the mouse, but this would at least fill the basic controls and likely be what the person on the steam forum wanted.

toilethinges:
As long as it'd be togglable, this'd be fine. As there's no way that I know of to make your cursor simply disappear, you'd always be going in one direction, and you'd constantly have to be piloting away from where your mouse points.

LocalGod79:
so, if i just wanted to select a ship with the mouse, without moving to it?

Amazinite:
That's why I suggested the RMB version, as the move to cursor version seems like it could be difficult at times.  I was just stating that the move to cursor was what they requested. Possibly a toggle on/off button for times like when you want to click to select.

lehgogh:
@DingusShingleton good idea, but we would need to make all of these bindable (laptops don't have middle click).

DingusShingleton:
Mine does ;) But you are right of course.

LocalGod79:
being an accessibility issue, this would certainly be worth a look at.

the steam user, who is finding co-ordinated button pushing difficult due to nerve damage, would probably be happy with a "make ship try to point at mouse, with auto aim" toggle in the preferences, and perhaps the ability to create mouse-button key bindings.

endless-sky:
I can think of a couple of different ways this could be done. It would take some experimentation to figure out which would be the easiest to control:

* Rely on auto-fire for primary trigger control. Have the ship always try to point toward the mouse cursor, and use left mouse for thrust and right mouse to fire secondary weapons. Maybe have auto-aim take precedence over the mouse direction unless you're actively applying thrust?

* "Tactical control" mode: rather than directly steering your ship, the mouse is just used to give the AI commands. This is already partly implemented: you can click to target a ship or planet, and double-click to board or land. Full mouse control would require at least two other commands: click to move to a certain location, and double-click a live ship to have the AI fight it.

* Mouse toggle: Allow some of the same commands as "tactical" mode above, but also let the player right click to toggle a mode where your ship is following the cursor and left mouse button controls the thrusters.

[...]

PR #4715, opened 2019-12-10
Combined single (keyboard and mouse) input from Engine towards AI

petervdmeer:
**Feature:** This PR combines all input-device signals in the Engine class into a single input towards the AI class

## Feature Details
This PR combines the inputs from the keyboard and the mouse into a single activeCommands Command variable and sends the inputs through this variable into the relevant AI functions.
Keyboard inputs were moved into the Engine class for this.

## UI Screenshots
N/A

## Testing
I tested all the relevant input scenarios that I could come up with;

- MENU
- MAP
- INFO
- FORWARD, LEFT, RIGHT, BACK, PRIMARY, AFTERBURNER
- SCAN
- DEPLOY
- HAIL
- BOARD
- CLOAK
- LAND (autopilot)
- LAND (switch landing targets)
- LAND (landing based on mouseclick)
- JUMP (autopilot)
- JUMP (insufficient fuel warning)
- JUMP (WAIT, fleet waiting for jump)
- TARGET (cycle through enemy ships)
- NEAREST (cycle through friendly ships)
- SELECT (switch secondary weapon)
- SECONDARY (fire secondary weapon)
- DEPLOY
- FIGHT
- GATHER (looks the same as before the change)
- HOLD (looks the same as before the change)
- AMMO

The inputs that I tested still respond the same as before this change (which means that bugfixes like #4710 still need to be merged).

##  Context
Allowing the auto-testing-framework #4370 to control the flagship and the fleet was the main reason for implementing this PR. The changes in this PR should also benefit the addition of other new input-devices (like touchscreen-support #2432, mouse-only flight #955, joysticks #3846, controllers #3972).

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -3432,7 +3432,7 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player, Command &activeCommand
 	if(autoPilot.Has(Command::LAND) || (autoPilot.Has(Command::JUMP) && isWormhole))
 	{
 		if(ship.GetPlanet())
-			activeCommands.Clear(Command::LAND | Command::JUMP);
+			autoPilot.Clear(Command::LAND | Command::JUMP);
 		else
 		{
 			MoveToPlanet(ship, command);
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -3441,28 +3441,23 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player, Command &activeCommand
 	}
 	else if(autoPilot.Has(Command::JUMP))
 	{
-		bool isNewPress = !isRequestingJump && activeCommands.Has(Command::JUMP);
-		isRequestingJump = activeCommands.Has(Command::JUMP);
 		if(!ship.Attributes().Get("hyperdrive") && !ship.Attributes().Get("jump drive"))
 		{
 			Messages::Add("You do not have a hyperdrive installed.");
 			autoPilot.Clear();
-			if(isNewPress)
-				Audio::Play(Audio::Get("fail"));
+			Audio::Play(Audio::Get("fail"));
 		}
 		else if(!ship.JumpFuel(ship.GetTargetSystem()))
 		{
 			Messages::Add("You cannot jump to the selected system.");
 			autoPilot.Clear();
-			if(isNewPress)
-				Audio::Play(Audio::Get("fail"));
+			Audio::Play(Audio::Get("fail"));
 		}
 		else if(!ship.JumpsRemaining() && !ship.IsEnteringHyperspace())
 		{
 			Messages::Add("You do not have enough fuel to make a hyperspace jump.");
 			autoPilot.Clear();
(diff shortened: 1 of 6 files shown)


source/AI.cpp
```diff
@@ -3439,7 +3432,7 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 	if(autoPilot.Has(Command::LAND) || (autoPilot.Has(Command::JUMP) && isWormhole))
 	{
 		if(ship.GetPlanet())
-			autoPilot.Clear();
+			activeCommands.Clear(Command::LAND | Command::JUMP);
```
tehhowch: Is this right?

source/AI.cpp
```diff
@@ -3448,7 +3441,8 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player)
 	}
 	else if(autoPilot.Has(Command::JUMP))
 	{
-		bool isNewPress = keyDown.Has(Command::JUMP) || !keyHeld.Has(Command::JUMP);
+		bool isNewPress = !isRequestingJump && activeCommands.Has(Command::JUMP);
+		isRequestingJump = activeCommands.Has(Command::JUMP);
```
tehhowch: It'd be nice if we could avoid needing a new class member variable just to avoid repeatedly playing the fail sound.

source/Engine.cpp
```diff
@@ -1578,6 +1582,42 @@ void Engine::SendHails()
...
+	Command oldHeld = keyHeld;
+	keyHeld.ReadKeyboard();
+	Command keyDown = keyHeld.AndNot(oldHeld);
+	
+	++landKeyInterval;
+	if(oldHeld.Has(Command::LAND))
+		landKeyInterval = 0;
+	
+	// Wait with actual jumping until the jump command is released
+	// Or if pressing land quicky in succession, then use WAIT to switch landing targets
```
tehhowch: Perhaps we should document the interpretations of Command::WAIT in Command.h/.cpp

source/AI.cpp
petervdmeer: No, should have been autoPilot. Changed to autoPilot.

source/AI.cpp
petervdmeer: Agreed, and I apparently also solved this by using WAIT and sending the JUMP command only once as active command. So I could remove this code and still have the audio and error-message played/shown only once.

source/Engine.cpp
petervdmeer: Good idea. I added this documentation.

pouar:
Seems with this change, pushing the arrow keys while holding down J seems to cancel the jump and requires you to push J again. Should I open a new issue?

MCOfficer:
> 
> 
> Seems with this change, pushing the arrow keys while holding down J seems to cancel the jump and requires you to push J again. Should I open a new issue?

i think whether or not that's a bug is debatable - so imo you should open an issue where it can be discussed if necessary.

Issue #868, opened 2016-03-21
Choosing which drones/fighters deploy


Amazinite:
A useful mechanic for the future would be that when you press the deploy key, you get a menu that pops up that allow you to choose which drone/fighters deploy and how many. Either that, or be able to map different keys to different fighters and drones (maybe 1-0). This would mainly be useful for when there are more drone and fighter types and you might want to deploy different drones or fighters for different situations.

Wrzlprnft:
I like the idea with 1-0. but since you would need a extra screen where you can groupd fighters to keys, sit may take a while to work that out. maybe acessable via the pilot info screen? However this develops, i think the D-key should remain as it is.

LocalGod79:
perhaps clicking on the ship icon in your fleet in the bottom left of the h.u.d. to deploy/recall would work?

Amazinite:
I'd open a new issue but there's really no point since this is exactly the issue for it: currently, clicking on a group of fighters or drones on the left-side ship list and hitting deploy will deploy all your fighters and drones, but the fighters that you selected will deploy highlighted with the green markers and only they will follow the orders you give them. As such, I think it would make sense if when you hit deploy, only the selected fighters or drones would deploy instead of all of them.

PR #4719, opened 2019-12-16
feat: Deploy and recall only selected fighters instead of all.

petervdmeer:
**Feature:** This PR implements the feature request detailed and discussed in issue #868.

## Feature Details
The global deploy-variable is replaced by a deploy variable per (carried) ship. The command to deploy goes to individual carried ships (not to carriers), since a carrier might carry ships that should be deployed and other ships that shouldn't be deployed.

If the player selects fighters where part of the group is deployed and part not and then presses deploy, then the undeployed fighters from the group will also be deployed. Pressing deploy again will then recall the whole group.

## UI Screenshots
deploy_and_recall

## Usage Examples
- Just hit deploy with no ships selected, and then all ships deploy
- Select fighters and drones and then press deploy to just deploy (or recall) the selected fighters and drones.

## Testing
The following scenarios were tested by me (no issues found):
- Deploy and recall for flagship
- Deploy of only subset of ships
- Deploy and recall for escort
- Deploy of only subset of ships (together with escort)
- Deploy and recall of AI controlled ships
- Return (and not deploying) damaged fighters
- Recall of fighters (to escort) if escort is holding position and player is out of system


petervdmeer:
Merged master to resolve merge-conflict

source/AI.h
```diff
@@ -168,6 +168,7 @@ template <class Type>
 
 
 private:
+	void IssueDeploy(const PlayerInfo &player);
```
tehhowch, line 171: If this method doesn't use any member variables, it should be either a static method or be implemented in the anonymous namespace.

petervdmeer: Moved to anonymous namespace; the class didn't have any static methods, so moving to the anonymous namespace looked nicer.

source/AI.cpp
```diff
@@ -121,10 +121,20 @@ namespace {
 	void Deploy(const Ship &ship, bool includingDamaged)
 	{
 		for(const Ship::Bay &bay : ship.Bays())
-			if(bay.ship && (includingDamaged || bay.ship->Health() > .75))
+			if(bay.ship && (includingDamaged || bay.ship->Health() > .75) &&
+					(!bay.ship->IsYours() || bay.ship->IsDeploying()))
 				bay.ship->SetCommands(Command::DEPLOY);
 	}
 	
+	// Check if the ship contains a carried ship that needs to launch.
```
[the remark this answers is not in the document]
petervdmeer: A glossary of commonly used words in the "Contributing" section on the Wiki could be nice and helpful to align on the terms we use.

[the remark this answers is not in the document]
petervdmeer: I created an initial Glossary page and started a discussion on Discord ( https://discord.com/channels/251118043411775489/285540320823869441/727235840467796140 ) on how to get the new page on the Wiki.

petervdmeer:
Updated to latest master and briefly re-tested.

source/AI.cpp
```diff
@@ -1609,7 +1667,9 @@ bool AI::ShouldDock(const Ship &ship, const Ship &parent, bool playerShipsLaunch
 	
 	// A fighter should retreat if its parent is calling it back, or it is
 	// a player's ship and is not in the current system.
-	if(!parent.Commands().Has(Command::DEPLOY) || (ship.IsYours() && !playerShipsLaunch))
+	if(!parent.Commands().Has(Command::DEPLOY) && !ship.IsYours())
+		return true;
+	if (ship.IsYours() && (!ship.IsDeploying() || ship.GetSystem() != playerSystem))
 		return true;
```
tehhowch, line 1673: Can this logic be combined into an if-else if branch?

Maybe
```c++
if(ship.IsYours())
{
	if(!ship.IsDeploying() || ship.GetSystem() != playerSystem)
		return true;
}
else if(!parent.Commands().Has(Command::DEPLOY))
	return true;
```

At minimum, i think moving the `IsYours` check first for both lines helps clarify the logic that applies to the player ship vs NPC ship cases. Feel free to update the comment above this block as appropriate.

(Also, an `if (` snuck in.)

source/AI.cpp
```diff
@@ -3489,7 +3549,7 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player, Command &activeCommand
 		}
 	}
 	
-	if(ship.HasBays() && isLaunching)
+	if(ship.HasBays() && IsLaunching(ship))
```
tehhowch, line 3552: Perhaps to reduce some ambiguity, we name this method `HasDeployingFighter` or similar.

source/AI.cpp
```diff
@@ -119,10 +119,74 @@ namespace {
...
+			for(const shared_ptr<Ship> &it : player.Ships())
+				if(it.get() != player.Flagship() && !it->IsParked() && it->CanBeCarried())
+					ships.push_back(it.get());
+		}
+		else
+			for(const weak_ptr<Ship> &it : player.SelectedShips())
+			{
+				shared_ptr<Ship> ship = it.lock();
+				if(ship)
+					ships.push_back(ship.get());
```
tehhowch, line 144: Is this missing a "CanBeCarried" or "HasBays" condition?

Also note that player ships are not [deleted until the player lands](https://github.com/endless-sky/endless-sky/blob/d3ccbdd72f71b59b6539dc5658b8d201ae2cdf0a/source/PlayerInfo.cpp#L1088-L1105), so if a selected ship gets destroyed, it will still be present in `SelectedShips()` until the player re-selects other ships. Note that this means that `SelectedShips()` can possibly contain only destroyed ships, in which case we probably want to treat pressing the deploy/recall key as the "full fleet" scenario.

source/AI.cpp
```diff
@@ -119,10 +119,74 @@ namespace {
...
+			}
+		
+		// This should never happen, but just in case:
+		if(ships.empty())
+		{
+			if(fullFleet)
+				Messages::Add("No carried ships in the fleet to deploy.");
```
tehhowch, line 151: This message can show up a lot, if the player has 2 or more ships and no fighters or drones to deploy. Past behavior was that the deploy/recall message was only shown if you had carried ships in your active fleet.

source/AI.cpp
```diff
@@ -119,10 +119,74 @@ namespace {
...
+			else
+				Messages::Add("No carried ships selected that can be deployed.");
```
tehhowch, line 153: This is currently unreachable, due to the above missing condition:
image

source/AI.cpp
```diff
@@ -472,7 +530,7 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 		double healthRemaining = it->Health();
 		bool isPresent = (it->GetSystem() == playerSystem);
 		bool isStranded = IsStranded(*it);
-		bool thisIsLaunching = (isLaunching && isPresent);
+		bool thisIsLaunching = (isPresent && IsLaunching(*it));
```
tehhowch, line 533: Are there other uses of `thisIsLaunching` that need to be revisited, now that launching is a selective process for player escorts?

source/AI.cpp
petervdmeer: Agreed, it was missing CanBeCarried, and it was also missing IsDestroyed. I added both and changed the order to first go through selected ships.

source/AI.cpp
petervdmeer: It now only shows up when the player hits the deploy button, and only when the player has no deployable ships in the fleet.
Do you prefer the deploy command to just remain silent in this situation? I have no strong feelings about this, but my first impression was that this message is better than just silently ignoring the command.

source/AI.cpp
petervdmeer: Agreed, removed the unreachable message.

source/AI.cpp
petervdmeer: It is used 3 times in the code after this point, and all 3 locations seem to make sense in the new situation;
- line 540: Disabled ships under the players command only deploy when the player commands the carried ships to deploy.
- line 567: Intented deploy behavior, only ships with carried ships that deploy will deploy.
- line 835: Recall behavior, player ships only recall when the carrier and carried ship are in a different system or when the player explicitly orders a recall.

source/AI.cpp
petervdmeer: Yes, that looks quite readable; implemented as suggested and updated the comment block(s).

source/AI.cpp
petervdmeer: Was indeed a bit ambiguous, renamed to `HasDeployments`.

PR #4724, opened 2019-12-28
Nitpicks

perey:

## Summary
Fixed a number of grammatical and spelling errors/typos.

## PR Checklist
 - [✓] I decline to claim copyright of any assets produced or modified


data/free worlds reconciliation.txt
```diff
@@ -940,7 +940,7 @@ mission "FW Syndicate Capture 1C"
 mission "FW Pug 1"
 	landing
 	name "What happened to Rand?"
-	description "Try to determine why communications with the Rand system were cut off. If you cannot reach Rand, at least look for anything odd in the neighboring systems."
+	description "Try to determine why communications with Rand, in the Zeta Aquilae system, were cut off. If you cannot reach Rand, at least look for anything odd in the neighboring systems."
```
MCOfficer: not sure if these commata are necessary

data/remnant missions.txt
```diff
@@ -576,7 +576,7 @@ mission "Remnant: Technology Available"
 	on offer
 		payment 2000000
 		set "license: Remnant"
-		log "The prefects have reached a consensus that you should be considered a valuable friend of the Remnant, and shall be given permission to purchase their ships and technology."
```
MCOfficer: iirc this line has been debated before. @Zitchas ?

data/remnant missions.txt
```diff
@@ -1031,8 +1031,8 @@ mission "Remnant: Return the Samples 2"
...
-			`	The lab itself is a microcosm of Remnant development, the structure appears to be some strange semi-organic material, and you can see an eclectic mix of familiar and unfamiliar shapes cluttering the tables. If you had to guess, you would think that this lab was the result of several different families of technology, only one of which was human.`
+			`Following Plume's directions, you find a lab tucked behind the main area of the spaceport, where a short walkway leads into an underground structure. Stepping inside the door you note a rack of camouflage netting positioned easily at hand. Given what you have learned of their history, it is probably a last-ditch preparation in case they need to hide the entrance from a raid.`
+			`	The lab itself is a microcosm of Remnant development. The structure appears to be some strange semi-organic material, and you can see an eclectic mix of familiar and unfamiliar shapes cluttering the tables. If you had to guess, you would say that this lab was the result of several different families of technology, only one of which was human.`
```
MCOfficer: I'm fine with think tbh, but no hard feelings either way

data/remnant missions.txt
Amazinite: This line is being fixed in #4706.

data/free worlds reconciliation.txt
Amazinite: This is structured the same way as the `<destination>` replacement, which reads `<planet> in the <system> system`, so the commas can be removed.

data/remnant missions.txt
Zitchas: The dash between last and ditch makes sense.

The comma after development, though... I think that might actually be better as a colon.

As to changing the think to say, I disagree with. I don't think that's really a necessary change.

data/free worlds reconciliation.txt
perey: My main concern was to name the system properly. If no-commas is more in line with the established style, no-commas it is.
```suggestion
	description "Try to determine why communications with Rand in the Zeta Aquilae system were cut off. If you cannot reach Rand, at least look for anything odd in the neighboring systems."
```

data/remnant missions.txt
perey: Full stop or colon, either way... just not a comma!

say → think: To elaborate on the curt commit message, my reasoning was that this text is describing your internal monologue. So, of course you _think_ that. Also, this may just be my idiom, but "if [I] had to guess, [I] would think" seemed awkward to me. It may be the implication of "I will make a guess, but I'll only think it, not say it" that strikes me as odd.

Zitchas: Agreed on that. The comma needs to go. Switch that to a colon and that will be good.

Sounds like the say/think part will need more input from the community. Maybe leave it off this PR and/or get more feedback on it before including it.

Amazinite: I'd say that both say and think work here, and one doesn't necessarily sound "more correct" to me than the other. When a situation like this pops up (i.e. the change isn't agreed upon as being an improvement, but it isn't wrong either), we usually air on the side of just leaving it as it was.

Amazinite:
#4706 changed one of the lines you were looking to fix here, so you'll need to resolve the conflict caused by that.

PR #4685, opened 2019-11-23
Nanachi missions and variants

Janaszar:
Corrected/updated upload of @Arachi-Lover 's Nanachi missions and required ship variants.  I have permission from him to do so.

Everything looks clear on my end.

@Pointedstick Per your requests, the grammar and spelling have been double checked.  Everything looks ok.
Per your talk of her being a miner and making the gift, she gained the hobby while mining, but made the gift after originally meeting you.

Anything else, comment here and I'll take care of it.

PR #4738, opened 2020-01-05
Adding a spaceport to Wormhole Alpha and reworking containing systems

Gods-Righthand:
## Summary
This PR adds a spaceport to Wormhole Alpha. Enabling NPCs and fleets that spawn in Waypoint to land on the wormhole.

Thanks to @Zitchas for suggesting this, as I only thought of doing it for the Remnant wormholes in #4737

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -417,7 +417,6 @@ government "Wanderer"
 
 # A dummy government used to prevent Hai from entering the wormhole in Waypoint/Ultima Thule, but allowing human merchants and other governments to enter.
 government "Wormhole Alpha"
-	color .4 .4 .4
 	"attitude toward"
 		"Hai" -.01
 		"Hai (Unfettered)" -.01
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -419,5 +419,5 @@ government "Wanderer"
 government "Wormhole Alpha"
 	color .4 .4 .4
 	"attitude toward"
-		"Hai" -.3
-		"Hai (Unfettered)" -.3
+		"Hai" -.01
+		"Hai (Unfettered)" -.01
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -90,6 +90,7 @@ government "Hai"
 	
 	"player reputation" 0
 	"attitude toward"
+		" Hai " 1
 		"Hai (Unfettered)" -.1
 		"Merchant" .01
 	"bribe" .02
@@ -104,6 +105,7 @@ government " Hai "
 	
 	"player reputation" 0
 	"attitude toward"
+		"Hai" 1
 		"Hai (Unfettered)" -.1
 		"Merchant" .01
 	"bribe" .02
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -403,4 +403,4 @@ government "Wormhole Alpha"
 	color .4 .4 .4
 	"attitude toward"
 		"Hai" -.3
-		"Hai (Unfettered" -.3
+		"Hai (Unfettered)" -.3
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -98,6 +98,20 @@ government "Hai"
 	"hostile hail" "hostile hai"
 	"hostile disabled hail" "hostile disabled hai"
 
+government " Hai "
+	swizzle 0
+	color .84 .42 .81
+	
+	"player reputation" 0
+	"attitude toward"
+		"Hai (Unfettered)" -.1
+		"Merchant" .01
+	"bribe" .02
+	"friendly hail" "friendly hai"
+	"friendly disabled hail" "friendly disabled hai"
+	"hostile hail" "hostile hai"
+	"hostile disabled hail" "hostile disabled hai"
+
 government "Hai (Unfettered)"
 	swizzle 4
 	color .69 .33 .82
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -399,6 +399,7 @@ government "Wanderer"
 	"hostile hail" "wanderer untranslated"
 	"hostile disabled hail" "wanderer untranslated"
 
+# A dummy government used to prevent Hai from entering the wormhole in Waypoint/Ultima Thule, but allowing human merchants and other governments to enter.
 government "Wormhole Alpha"
 	color .4 .4 .4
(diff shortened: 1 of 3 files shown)


Amazinite:
The Eye is currently unique in that it's the only wormhole that NPCs will actually enter. With this PR and #4737, do we really want to make it so that all wormholes will be used by NPCs?

beccabunny:
I'd say that at least #4737 should be added, but I'm not too convinced about this

MCOfficer:
> The Eye is currently unique in that it's the only wormhole that NPCs will actually enter. With this PR and #4737, do we really want to make it so that all wormholes will be used by NPCs?

Technical difficulties aside - can we make them *rarely* enter wormholes? in the case of alpha, that feels like the most realistic scenario to me.

Zitchas:
Perhaps the Alpha wormhole usage should be added via event triggered by a mission, so that ships don't use it up until the point when Hai/Human interactions are a little bit more public knowledge?

That being said, Waypoint seems to have really heavy traffic for an edge system. It feels like the amount of traffic one would see in an uninhabited system that lies between two fairly major systems and has a lot of transit traffic. Except it isn't. There's nowhere for ships to go except back where they came from or into the wormhole. If they *aren't* using the Wormhole, then what are they doing there? Just spectating?

Gods-Righthand:
>do we really want to make it so that all wormholes will be used by NPCs?

The player does, why wouldn't others?
That said I don't currently see any point to adding this, unless we could do what M*C*O suggested.

>Technical difficulties aside - can we make them rarely enter wormholes?

@Zitchas idea of creating a mission.
Great idea, I would assume that someday the secrecy around the Hai will end, and trade will blossom. There's already small interactions between them.

Zitchas other point of Waypoint being busy. 
For now, should we leave out wormhole traffic and cut down spawn rates and fleet sizes for that system?

Amazinite:
> If they aren't using the Wormhole, then what are they doing there? Just spectating?

Refueling merchants after their long journey. It would sure suck to make it that far then have no fuel to make the final jump and have to wait for a long time for a help.

MCOfficer:
> For now, should we leave out wormhole traffic and cut down spawn rates and fleet sizes for that system?

and yet another suggestion enters the ring - how about both? that would make wormhole crossings *actually* rare, without too much effort.

Amazinite:
Could we add a dummy government that's hostile to the Hai and give it to the wormhole, causing the Hai to not be able to land on it, but merchants would be able to?

Gods-Righthand:
Added a government to the wormhole that prevents Hai and Unfettered Hai from going through.

Even with that, I think it would be good to cut down on the frequency of fleets in Waypoint. I don't see the purpose for large Hai fleets to go  "Refuel merchants after their long journey."
When I just cut out "Large Hai", the small Hai still come not longafter the last fleet leaves.

data/governments.txt
```diff
@@ -398,3 +398,9 @@ government "Wanderer"
 	"friendly disabled hail" "wanderer untranslated"
 	"hostile hail" "wanderer untranslated"
 	"hostile disabled hail" "wanderer untranslated"
+
+government "Wormhole Alpha"
+	color .4 .4 .4
+	"attitude toward"
+		"Hai" -.3
+		"Hai (Unfettered" -.3
```
jafdy: Closing bracket is required here.

Gods-Righthand: Good catch, thanks

data/governments.txt
```diff
@@ -398,3 +398,9 @@ government "Wanderer"
 	"friendly disabled hail" "wanderer untranslated"
 	"hostile hail" "wanderer untranslated"
 	"hostile disabled hail" "wanderer untranslated"
+
+government "Wormhole Alpha"
```
Amazinite: Perhaps add a comment here?
```suggestion
# A dummy government used to prevent Hai from entering the wormhole in Ultima Thule, but allowing human merchants to enter.
government "Wormhole Alpha"
```

I just realized though that this is going to cause an issue with "Pirate Troubles [4]" where the player needs to escort a number of Hai Geocoris out of Hai space and back.

Gods-Righthand:
Done. I fixed the Pirate Troubles [4] as well. I'm assuming that this will get merged before #4685 and will leave that to @Janaszar

data/governments.txt
```diff
@@ -98,6 +98,20 @@ government "Hai"
 	"hostile hail" "hostile hai"
 	"hostile disabled hail" "hostile disabled hai"
 
+government " Hai "
+	swizzle 0
+	color .84 .42 .81
+	
+	"player reputation" 0
+	"attitude toward"
```
Amazinite: This needs attitude toward `"Hai" 1` and `"Hai"` needs `" Hai " 1` so that their reputations will match and attacking one angers the other.

data/governments.txt
```diff
@@ -398,3 +414,10 @@ government "Wanderer"
 	"friendly disabled hail" "wanderer untranslated"
 	"hostile hail" "wanderer untranslated"
 	"hostile disabled hail" "wanderer untranslated"
+
+# A dummy government used to prevent Hai from entering the wormhole in Waypoint/Ultima Thule, but allowing human merchants and other governments to enter.
+government "Wormhole Alpha"
+	color .4 .4 .4
+	"attitude toward"
+		"Hai" -.3
+		"Hai (Unfettered)" -.3
```
Amazinite: Set these values to -.01 so that the player's rep with Wormhole Alpha doesn't change; repairing Hai could lead to the player being unable to access the wormhole under this implementation.

Gods-Righthand: So many small details I don't know, anything else?

data/governments.txt
```diff
@@ -398,3 +414,10 @@ government "Wanderer"
 	"friendly disabled hail" "wanderer untranslated"
 	"hostile hail" "wanderer untranslated"
 	"hostile disabled hail" "wanderer untranslated"
+
+# A dummy government used to prevent Hai from entering the wormhole in Waypoint/Ultima Thule, but allowing human merchants and other governments to enter.
+government "Wormhole Alpha"
+	color .4 .4 .4
```
Amazinite: I'd remove this line before it's useless.

PR #4739, opened 2020-01-11
Change Relic's description

Naminson:
**Content**

## Summary
This PR replaces the description of Relic with one that is much less underwritten. The only change in lore is that there is now only one settlement, unlike the previous version, which mentioned multiple. However, I think the point of the planet is still intact. As in, it was once expected to be a major planet, but has since become just a reminder of the past.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -26322,7 +26322,7 @@ planet Relic
 	attributes "near earth"
 	landscape land/valley4
 	description `Relic was colonized over seven hundred years ago, back when the construction of the first hyperdrive was still in living memory. At this point, most starships had a very limited range, so the stars adjacent to earth where the only viable opinion for any colonization attempts. Because of this, Relic seemed like it was going to become a major colony, despite it's low gravity, harsh environment, and limited resources.`
-	description `However, this would not last. In less than fifty years, the colonies had spread to much more favourable planets, and Relic was left behind by most human development.`
+	description `	However, this would not last. In less than fifty years, the colonies had spread to much more favorable planets, and Relic was left behind by most human development.`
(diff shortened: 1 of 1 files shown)


Zitchas:
All in all, I'd say that's an improvement.

Just to be pedantic, I'll note that "only one recognized city" doesn't necessarily mean "only one settlement." No reason to change that, no. It's good the way it is.

The last paragraph seems a bit off to me. I'm not sure exactly what it is, although I do note that it uses "here" a lot.

data/map.txt
```diff
@@ -26321,8 +26321,9 @@ planet "Rekat Moraski"
...
+		description `Relic was colonized over seven hundred years ago, back when the construction of the first hyperdrive was still in living memory. At this point, most starships had a very limited range, so the stars adjacent to earth where the only viable opinion for any colonization attempts. Because of this, Relic seemed like it was going to become a major colony, despite it's low gravity, harsh environment, and limited resources.`
+	description `However, this would not last. In less than fifty years, the colonies had spread to much more favourable planets, and Relic was left behind by most human development.`
+	spaceport `The spaceport here is located in the only recognized city on Relic, and it's mostly designed to handle the occasional merchant or navy ship. Almost all the ships that do arrive are here to refuel, since the industry here is made up of a few factories and a trading post. This leads to a running joke among the locals that the chief industry here is the gas station.`
```
Amazinite: I'd keep the old description's characterization of the spaceport as being just a bunch of concrete slabs for ships to land on instead of having proper landing pads. It helps to better show just how poor and crappy this spaceport is.
Also, Navy.
```suggestion
	spaceport `The spaceport here is located in the only recognized city on Relic, and it's mostly designed to handle the occasional merchant or Navy ship. Almost all the ships that do arrive are here to refuel, since the industry here is made up of a few factories and a trading post. This leads to a running joke among the locals that the chief industry here is the gas station.`
```

Issue #2738, opened 2017-07-10
Make timelapse key (default capslock) configurable


nylki:
Currently the 3x speed key is bound to caps-lock. It would be useful to have it configurable.

Pointedstick:
Indeed. People sometimes re-map their caps lock keys to something more useful.

lelutin:
I for one would like to have this since I configure caps-lock to be a second Escape key (which means I can't use the time lapse feature of the game. I'd like to for example remap that function to Tab

PR #4741, opened 2020-01-12
Make fast-forward key configurable

bojidar-bg:
**Feature:** This PR implements the feature request detailed and discussed in issue #2738

## Feature Details
This PR makes the fast-forward key configurable. It does not make using Caps Lock for low framerate in debug mode configurable, however.

This would be helpful for people using Caps Lock as a compose, layout switch, or second escape key.
It also makes the fast-forward feature easier to discover.

The fact that Caps Lock still toggles low framerate in debug mode means that one can use fast-forward in conjunction with it, for example to reproduce a bug which occurs only after some time.

(Note: It would be nice if the keybinding is listed in [the manual](https://github.com/endless-sky/endless-sky/wiki/PlayersManual#key-bindings-and-shortcuts), even if this PR is not merged)

## UI Screenshots

| Before | After |
| --- | --- |
| image | image |

(Setting below "Toggle Fullscreen")

## Usage Examples
N/A

### DIFF
--- a/source/Command.cpp
+++ b/source/Command.cpp
@@ -61,12 +61,13 @@ const Command Command::CLOAK(1uL << 17, "Toggle cloaking device");
 const Command Command::MAP(1uL << 18, "View star map");
 const Command Command::INFO(1uL << 19, "View player info");
 const Command Command::FULLSCREEN(1uL << 20, "Toggle fullscreen");
-const Command Command::FIGHT(1uL << 21, "Fleet: Fight my target");
-const Command Command::GATHER(1uL << 22, "Fleet: Gather around me");
-const Command Command::HOLD(1uL << 23, "Fleet: Hold position");
-const Command Command::AMMO(1uL << 24, "Fleet: Toggle ammo usage");
-const Command Command::WAIT(1uL << 25, "");
-const Command Command::STOP(1ul << 26, "");
+const Command Command::FASTFORWARD(1uL << 21, "Toggle fast-forward");
+const Command Command::FIGHT(1uL << 22, "Fleet: Fight my target");
+const Command Command::GATHER(1uL << 23, "Fleet: Gather around me");
+const Command Command::HOLD(1uL << 24, "Fleet: Hold position");
+const Command Command::AMMO(1uL << 25, "Fleet: Toggle ammo usage");
+const Command Command::WAIT(1uL << 26, "");
+const Command Command::STOP(1ul << 27, "");
 
 
 
--- a/source/main.cpp
+++ b/source/main.cpp
@@ -314,7 +314,7 @@ int main(int argc, char *argv[])
 					else
 						SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN_DESKTOP);
 				}
-				else if(event.type == SDL_KEYDOWN
+				else if(event.type == SDL_KEYDOWN && !event.key.repeat
 					&& (Command(event.key.keysym.sym).Has(Command::FASTFORWARD)))
 				{
 					isFastForward = !isFastForward;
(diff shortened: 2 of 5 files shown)


source/Command.cpp
```diff
@@ -61,12 +61,13 @@ const Command Command::CLOAK(1uL << 17, "Toggle cloaking device");
 const Command Command::MAP(1uL << 18, "View star map");
 const Command Command::INFO(1uL << 19, "View player info");
 const Command Command::FULLSCREEN(1uL << 20, "Toggle fullscreen");
-const Command Command::FIGHT(1uL << 21, "Fleet: Fight my target");
-const Command Command::GATHER(1uL << 22, "Fleet: Gather around me");
-const Command Command::HOLD(1uL << 23, "Fleet: Hold position");
-const Command Command::AMMO(1uL << 24, "Fleet: Toggle ammo usage");
-const Command Command::WAIT(1uL << 25, "");
-const Command Command::STOP(1ul << 26, "");
+const Command Command::FASTFORWARD(1uL << 21, "Toggle fast-forward");
```
tehhowch: Any reason you went for 21 instead of 27? Going for 27 would minimize the diff

source/main.cpp
```diff
@@ -313,6 +314,11 @@ int main(int argc, char *argv[])
 					else
 						SDL_SetWindowFullscreen(window, SDL_WINDOW_FULLSCREEN_DESKTOP);
 				}
+				else if(event.type == SDL_KEYDOWN
```
tehhowch: I think we should also check if this is a new press, as if one has updated the keybinding, holding the new FF key will result in a continuous toggling of the FF state

source/Command.cpp
bojidar-bg: From what I saw (didn't look too carefully, so I might have overlooked something), the constants have the same order in the command class, in the preferences panel, and here. So, I decided to keep the order, and then kept the numbers consecutive.

bojidar-bg:
> […] but is something to note for a follow-up (none of the other keybinding defaults overflow here):

I could fix it in this PR, if you say so; the code using `Table` looks a bit messy, but should be manageable :smiley: .

Amazinite:
I'm not sure that "menus" is an apt header for the fast forward control to go under; perhaps add an "other"/"misc" section?

bojidar-bg:
> (This would also minimize the diff in Command.cpp by having fast forward be the last in the list.)

Actually it would not, since "Toggle fullscreen" should probably go there as well.. and this would again shift all numbers, this time to one less :smiley:

bojidar-bg:
About the column size, it would require something more than just a slight layout change:
image
(Note how "Select nearest hostile ship" and "CapsLock" are perfectly aligned with the highlight, yet "PrintScreen" manages to overflow it)

I decided against moving fullscreen and fast-forward in a seperate section, instead opting for renaming the section to "Interface" (as you might see in the above image).

Amazinite:
Are there any other key binds that would be longer than caps lock? PrintScreen could be shortened to something like PrntScr.

bojidar-bg:
There are some, yeah. From [the SDL wiki](https://wiki.libsdl.org/SDL_Keycode):
* "Application" (meta / windows key)
* "Backspace"
* "Keypad Enter" (and some others starting with "Keypad ")

Also a few obscure ones:
* "AC Bookmarks", perhaps others starting with "AC " as well
* "BrightnessUp" / "BrightnessDown" / "KBDIllumUp" / "KBDIllumDown" / "KBDIllumToggle"
* "Calculator" / "Computer"
* "CurrencySubUnit" / "CurrencyUnit"
* "DecimalSeparator" / "Separator" / "ThousandsSeparator"
* "DisplaySwitch"
* "VolumeDown", but probably not "VolumeUp"

MCOfficer:
> 
> 
> About the column size, it would require something more than just a slight layout change (the red highlight is not in the PR, but was added just for debugging purposes):
> image
> (Note how "Select nearest hostile ship" and "CapsLock" are perfectly aligned with the highlight, yet "PrintScreen" manages to overflow it)
> 
> I decided against moving fullscreen and fast-forward in a seperate section, instead opting for renaming the section to "Interface" (as you might see in the above image).

i feel like this is a bug, but none that should prevent this PR from merging. It already exists for all other key settings.

bojidar-bg:
I agree. The main reason it was a problem for the PR is that "CapsLock" was the first _default_ keybind which was longer than the column size.

PR #4748, opened 2020-01-18
New Williwaw sprite

Brick63:
**Content: Artwork**

Summary:
A Replacement of the williwaws sprite to make the hai more unique

Assets:
[Hai Williwaw Assets.zip](https://github.com/endless-sky/endless-sky/files/4081803/Hai.Williwaw.Assets.zip)

hai williwaw@2x


Lineth:
I'd say as an alternative is to make this a Hai version of the Water cooling system, available from Hai-home or a mission string. It would make up for how later in the story the Williwaw by itself is a little inadequate for the ships it's cooling.

But as a straight replacement it looks great too.

Issue #4700, opened 2019-12-04
Request: Default emission point for ships


Zitchas:
**Is your feature request related to a problem? Please describe.**
There have been a number of requests for features of various kinds that run into the problems related to "where does it come from?" The default answer is usually "the center of the ship," but we have some ships for which something coming from the center doesn't look good. For most of these, taking up a turret slot doesn't really feel appropriate, and the thing may not really function without having some kind of defined point.

The things running into these include such things as:
* being able to add fighter, drone, or other ship bays as outfits (where does the ship come out of?)
* being able to add gun or turret slots via outfits (where does it go? It isn't that hard to weld a hardpoint onto a ship)
* adding a new mechanic to the game (such as a flotsam tractor beam) that needs a point of emission but doesn't necessarily need to take up a turret slot.

**Related Issue Links**
[Amazinite's flotsam tractor beam](https://github.com/endless-sky/endless-sky/issues/4699) would be one instance of this being useful. 
[My request for the ability to add bays as outfit types](https://github.com/endless-sky/endless-sky/issues/4610) would also benefit from this.
[Finite Sky's request for more hardpoint types](https://github.com/endless-sky/endless-sky/issues/4684) might be relevant as well.

**Describe the solution you'd like**
I would like a point to be added to the ship definition, just like guns, turrets, engines, fighters, and drones currently are. This "default" point would be listed in the ship data file, and would be used as a catch-all point whenever the game needs a point for something to come off a ship but doesn't have a more appropriate point available. This point would by default be on the belly of the ship, so anything coming off it would always come from underneath. 

In the data file it would look something like: `default 0 0`

For an example in context:
```
    outfits
        X2700 Ion Thruster        
        X2200 Ion Steering        
        Hyperdrive        
    engine -6 30        
    engine 6 30
    gun 0 -31        
    default 0 15
    leak leak 60 50```

**Additional context**
Given that this is an ambiguous point, I don't think it should be listed on the ship UI, although I suppose it could be identified as "Misc. Hardpoint" or something of the sort. 

It might be useful to allow multiple points to be specified, with the engine simply cycling through points whenever they are used. Some ships might benefit from this, although it may also look kind of odd in some circumstances, depending on how it's done.

beccabunny:
I think that it would be nice that if no "default" is specified, then it is read as having "default 0 0", since:
-the features this is needed for don't seem like it will limit the quantity of an item that can be installed like the weapon hardpoints do
[...]

PR #4749, opened 2020-01-19
feat: Allow more bay orientations / facing-angles

petervdmeer:
**Feature:** This PR implements a feature where bays can get a custom facing-angle, besides the current/previous facing-angles forward, left, right and back.

## Feature Details
Bay is now internally using an angle instead of a facing (index) variable.
Standard angles like 90 degrees and 180 degrees are still written out as right and back.
Non-standard angles are read and written using the angle child keyword.

Added comparison operator to Angle-class for ship/bay-savegame writing of standard angles.

## Testing Done
Tested with attached savegame:
- load and save
- checked angles both in save-file (after re-saving) and in flight

## Performance impact
None expected, the angles used after this PR are the same angles as for left, right, back angles.

## UI Screenshots
Here is a screenshot of the "Twister" ship (as also provided in attached savegame) that uses the new angle functionality for two of its bays.
six_directions

## Usage Examples
Here is an example of a bay with a current/previous angle and another bay with a custom facing angle:
```
fighter 47 -61 under
	angle 45
fighter -47 14 under left
```
Attached savegame can be used for testing this feature.
[six directions.txt](https://github.com/endless-sky/endless-sky/files/4083399/six.directions.txt)


Zitchas:
I like the sounds of it. Probably not an urgent issue, but useful nontheless.

How does this interact with the other projects that interact with fighters? Thinking mainly of default launch location one.

Amazinite:
Given that having engine sprites be able to come out at arbitrary angles is a part of #4539, I'd think that the syntax for bay angles and engine angles should be the same, or at least as similar as possible. The current syntax in #4539 is:
```
"engine" <x> <y> <zoom> <angle>
	[over | under]
```
Whereas the suggested syntax for bays is:
```
"fighter" <x> <y> [over | under] [left | right]
	"angle" <angle>
```
For reverse compatibility's sake, engines can't move their zoom and bays can't move their over/under and directional tags. The rationale behind having the over/under tags be a child of `"engine"` is so that you don't always need to specify the zoom and angle just to specify the over/under as well. Swapping the angle and the over/under to match the suggested syntax here (i.e.
```
"engine" <x> <y> <zoom> [over | under]
	"angle" <angle>
```
) would mean that specifying over/under for an engine still requires you to also specify the zoom, but it would closer match the suggestion here (and I don't think it'd be impossible to parse `"engine" <x> <y> [over | under]` without the zoom being listed).

petervdmeer:
The most generic way to make those keywords similar is to change the syntaxes to:

```
"engine" <x> <y>
	"angle" <angle>
 	[over | under]
	"zoom" <zoom>

"fighter" <x> <y>
	"angle" <angle>
	[left | right]
	[over | under]
	"zoom" <zoom>
 ```

The "bay" keywords would need to support over, under and left/right on the same line as well for backwards compatibility, and the "engine" keywords would need to support the zoom on the same line as well for backwards compatibility. But this seems to be the most explicit way to make the keywords look similar. (And as a bonus you also get the format on how to have carried ships drawn at a zoom.)

petervdmeer:
> How does this interact with the other projects that interact with fighters? Thinking mainly of default launch location one.

I don't see any negative interaction with #4700 (which I think is the feature request you are referring to). If the default emission point would specify angles, then it can just use the same keywords as discussed here, but I would expect the default emission point not to use angles at all (and only allow forward-facing bays from the default emission point).

petervdmeer:
@Amazinite This PR now also contains an update to parse the generic format as discussed above.

petervdmeer:
I processed all review comments and did a short re-test to verify that the angles still get saved and loaded.

PR #4757, opened 2020-01-24
fix: CI: Don't run apt update on ubuntu runners

MCOfficer:
**Bugfix:** This PR addresses an issue raised by [this action run](https://github.com/endless-sky/endless-sky/pull/4411/checks?check_run_id=408031429).

## Fix Details
Specifically, the run failed because `sudo apt-get update` was called, while Microsoft's repository was apparently taking a nap or something. This appears to be a [frequent issue](https://github.com/dotnet/core-setup/issues/7778#issuecomment-523518122), which isn't nice for test runners that should be as reliable as possible.

Subsequently, @petervdmeer rightfully questioned running `apt update` in the first place:
> I also wonder why you would need to run apt-get update inside the job. At the size that Github is operating I would expect images to be automatically generated daily or hourly with an up-to-date apt-cache.

Can't argue with that. Here's to more stable Ubuntu builds.

### DIFF
--- a/.github/workflows/build_parse.yml
+++ b/.github/workflows/build_parse.yml
@@ -49,9 +49,7 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Install dependencies
-      run: |
-        sudo apt-get update
-        sudo apt-get install libsdl2-dev libpng-dev libjpeg-turbo8-dev libopenal-dev libmad0-dev libglew-dev libgl1-mesa-dev libegl1-mesa-dev libgles2-mesa-dev scons ccache
+      run: sudo apt-get install libsdl2-dev libpng-dev libjpeg-turbo8-dev libopenal-dev libmad0-dev libglew-dev libgl1-mesa-dev libegl1-mesa-dev libgles2-mesa-dev scons ccache
     - name: Cache ccache
       uses: actions/cache@v1.1.0
       with:


MCOfficer:
Welp, seems like the test run proved the two of us wrong 😆 

```
 Err:8 http://azure.archive.ubuntu.com/ubuntu bionic-updates/main amd64 libdrm-dev amd64 2.4.97-1ubuntu1~18.04.1
  404  Not Found [IP: 52.177.174.250 80]
```

I'll see if i can put together a PR that enforces the use of the official ubuntu mirror tomorrow.

tehhowch:
I originally was using [`apt`](https://github.com/tehhowch/endless-sky/pull/30/files#diff-898e737ee4e56ba042bda31764bed49eR27) but went for apt-get since the build logs were complaining that `apt` is more suited to a user and not a script, and also because `apt update` spewed 404's.

MCOfficer:
> 
> 
> I originally was using [`apt`](https://github.com/tehhowch/endless-sky/pull/30/files#diff-898e737ee4e56ba042bda31764bed49eR27) but went for apt-get since the build logs were complaining that `apt` is more suited to a user and not a script, and also because `apt update` spewed 404's.

I was talking exclusively about `apt-get` in this PR, apologies that i used them synonimously.
Using `apt` shouldn't change behavior (as far as this PR is concerned), since it uses the same backend.

PR #4763, opened 2020-01-26
feat: CI: Run windows builds in a matrix of MinGW versions

MCOfficer:
**Feature:** This PR implements the feature request detailed and discussed in discord:
> Mco, thoughts on having the windows builds also matrixed out over the various g++ versions, going back to the first with support for c++11? Seems like something we would want to do, but maybe not for every commit. Mostly because these various versions of MinGW also mean various versions of the c++ libs

## Feature Details
With this PR, instead of one Windows build, there are now 2, for MingW latest and v4.8.5 (as found on [chocolatey](https://chocolatey.org/packages/mingw)).

It also reformts the YAML file to align with - what appears to be - the YAML "standard", e.g. most formatters will produce this or something very similar.

### DIFF
--- a/.github/workflows/build_parse.yml
+++ b/.github/workflows/build_parse.yml
@@ -18,142 +18,142 @@ name: Build & Parse
 on:
   push:
     paths:
-    - 'source/**'
-    - 'data/**'
-    - 'tests/**'
-    - 'EndlessSky.xcodeproj/**'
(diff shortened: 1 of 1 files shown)


MCOfficer:
Closing since the workflows have changed a fair bit since this was opened. Please comment if you're still interested in this @tehhowch

Issue #4765, opened 2020-01-26
map panel does not consider completion conditions when coloring active missions


NomadicVolcano:
The map panel only runs basic checks in `Mission::IsSatisfied` and never checks the actual completion conditions.  That causes some missions to be displayed as being able to complete, even though they can't.

NomadicVolcano:
This bug is fixed by pull request #4766

PR #4766, opened 2020-01-26
consider completion conditions when coloring missions in map panel

NomadicVolcano:
As described in #4765, the map panel only runs basic checks in `Mission::IsSatisfied` and never checks the actual completion conditions. That causes some missions to be displayed as being able to complete, even though they can't.

This fix moves the test of completion conditions from `Mission::CanComplete` into `Mission::IsSatisfied` (which is called by `Mission::CanComplete`).  

A consequence of this is that we cannot lie to the player about the completion conditions quite so easily.

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -636,10 +636,11 @@ bool Mission::CanComplete(const PlayerInfo &player) const
 // bright or dim text colors.
 bool Mission::IsSatisfied(const PlayerInfo &player) const
 {
-	if(!toComplete.Test(player.Conditions()))
+	if(!waypoints.empty() || !stopovers.empty())
 		return false;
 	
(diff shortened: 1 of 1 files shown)


source/Mission.cpp
```diff
@@ -639,6 +636,9 @@ bool Mission::CanComplete(const PlayerInfo &player) const
 // bright or dim text colors.
 bool Mission::IsSatisfied(const PlayerInfo &player) const
 {
+	if(!toComplete.Test(player.Conditions()))
```
tehhowch: Move this check after the waypoint & stopover checks, as it is more expensive.

NomadicVolcano: The outfits and fines checks may be cheaper too.  NPCs may be expensive though.  Maybe it should go between outfits/fines and NPCs?

PR #3619, opened 2018-01-31
Empty systems follow the surrounding government

Nescio0:
Related: #2541, #3524, #3618

Currently uninhabited systems do not always change government when the only system(s) they're connected to change hands, which leads to odd, isolated "Republic" systems with no "Republic" presence whatsoever. This patch changes empty system governments accordingly. Feel free to point out anything that is overlooked.

PR #4785, opened 2020-02-01
Adding the Subsidurial

Zitchas:
**Content (Artwork / Alien Life)**
## Summary
This adds the Subsidurial, an indigenous space creature. It can be found in all uninhabited systems across the galaxy, but has a very, very low spawn rate. (60000). It is more common in Nenia (5000).

It is intended to add a little interest to these otherwise lonesome places, in much the same way that the hyperion did in EVN. It is not currently attached to any jobs or missions, although it is likely to be the subject of some Remnant jobs in the future.

Images courtesy of BeccaBunny

## PR Checklist
 - [X] I uploaded the necessary image, blend, and texture assets here: [zip file on google drive](https://drive.google.com/file/d/1zO8Y1I63ONFjSEtCtDU1smZLaTf4eX2-/view?usp=sharing)
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi/pull/111)


Amazinite:
I feel like these shouldn’t be so common in Nenia, as otherwise that would suggest some connection to the void sprites. The player may even get confused as to which lifeform they’re suppose to be worrying about. With so few lifeforms in space, I’d also imagine that if these were in Nenia, they’d be predatory against the void sprites, or vice versa. 

I’d limit these to less than just all uninhabited systems. Perhaps only all uninhabited systems in human/Hai/Coalition/Wanderer space. Or maybe take the Coalition out of that. We know that at one point human, Hai, and Wanderer space were all accessible via hyperdrive, so perhaps these lifeforms spread during that time period.

Zitchas:
1) Sure, I'll bump their rarity in Nenia up a whole bunch. Actually, I was thinking of adjusting the whole Ember Waste area to around 45000 (which puts them more common there than the rest of the galaxy, but still very rare.)

2) I was thinking - along the lines of the Hyperions in EVN - of something that is either sufficiently ancient or sufficiently mysterious to have somehow spread across the galaxy. In a way, kind of the opposite of the Void Sprites: Instead of being something that come up from planets into space to eat, these drift in from deep space to eat. Million-year-old giant space amoebas. While I don't have a specific "Here's where they came from, and here's their story" in mind right now, the main point is to introduce some mystery for players to wonder about. Well, for observant ones, anyway. Many players will just be curious as to what this "space cow" is doing out here at all, nevermind logisitics of how they came to be spread around. 

As a side issue, with the 60000 rarity, I'm not sure how many players will see enough of them to figure out that they *are* scattered across the galaxy.

data/indigenous.txt
```diff
@@ -8,6 +8,39 @@
...
+		Mouthparts? 3
+	gun 0 -97 Mouthparts?
+	gun -23 -95 Mouthparts?
+	gun 23 -95 Mouthparts?
+	turret -28 0
+	turret 28 0
+	leak "blood" 30 50
+	explode "blood" 45
+	"final explode" "void sprite adult death"
+	description "It is unclear what these creatures are, where they come from, or how they evolved. Scientists have dissected several, but so far the results have simply been that these creatures contain no materials or resources of value; are not even slightly edible, and are only barely sentient. To date, non have demonstrated any sign of sapience."
```
Anarchist2:
```suggestion
	description "It is unclear what these creatures are, where they come from, or how they evolved. Scientists have dissected several, but so far the results have simply been that these creatures contain no materials or resources of value; are not even slightly edible, and are only barely sentient. To date, none have demonstrated any sign of sapience."
```

Amazinite:
> It can be found in all uninhabited systems across the galaxy

You've added these to 58 systems, but there are more than 58 uninhabited systems in the game. Were there some systems that you didn't add them to for some reason?

Zitchas:
For a starter point, I added to to every system that had the "uninhabited" government (as well as all three Remnant systems), as opposed to every system that was actually uninhabited. 

As for the name, it is a purely created name. Its linguistic roots come from "subsist" (as in "to survive on" or "to barely survive") and "dura" (as in hard), so, something that subsists on hard things.

Back to the distribution side, is it possible to make person ships that will continue to respawn even after they've been killed? Because this ship would actually fit such a mechanic better. It is intended to be a more or less universally occurring entity, on the line of thought that they evolved hundreds of millions of years ago and have simply been drifting around the galaxy ever since. (the slow way - long bouts of millenias in hybernation as they drift between stars, although perhaps they use their innate propulsion as a kind of "long burn" that might get up to a significant portion of the speed of light, although they'd still end up in hibernation for decades or centuries at a time)

Short of such universal spawning, however, is the current form, where they have a very low rate in most systems that are formally uninhabited - not because they can sense whether a system is claimed or not, but because they are more likely to get harassed by researchers and curiosity seekers and people who just want to go poke the thing when they venture into inhabited/claimed systems. So in such systems, they'd tend to leave much faster than in the others.

If you have suggestions for other systems that are nominally claimed that would be good candidates, I would be happy to add them.

Amazinite:
I don't believe many if any systems in human space strictly have the uninhabited government, instead having a government that matches the surrounding inhabited systems, so you're missing a lot of systems by only adding them to "uninhabited" government systems.

Amazinite:
It might be worth changing all uninhabited systems to the uninhabited government, at least in human space since the government of uninhabited systems isn't important. Ref: #3619.
And given the characterization of:
> ... they have a very low rate in most systems that are formally uninhabited - not because they can sense whether a system is claimed or not, but because they are more likely to get harassed by researchers and curiosity seekers and people who just want to go poke the thing when they venture into inhabited/claimed systems. So in such systems, they'd tend to leave much faster than in the others.

It might be worth reassessing the criteria for these to spawn. What about uninhabited systems around the deep Core or in Kor space? Wouldn't the Korath and automata attack them? Same story for the Graveyard with the Ka'het. And what of more peaceful regions like Wanderer space? I imagine the Wanderers would leave the Subsidurials to do their own thing, and given the length of time the Wanderers have been there, they may spawn in all Wanderer systems. How about the same for Hai space?

So what I'm thinking is the following:
* None in Exile or Kor space.
* None in the Ember Waste as a whole. 
* Only uninhabited systems of human and Coalition space.
* In all Wanderer systems. (Start as the Unfettered move in, the Subsidurials disappear and start appearing in Kor space again.)
* Either only uninhabited Hai systems or in all systems. (Perhaps the spawns in Wanderer space are higher than Hai space.)

Arachi-Lover:
actually, there could be some on  Kor Space, just in Far En Lai's system, there they could work

data/indigenous.txt
```diff
@@ -8,6 +8,49 @@
...
+ship "Subsidurial"
+	noun "creature"
+	sprite "ship/subsidurial"
+	thumbnail "thumbnail/subsidurial"
+	"never disabled"
+	attributes
+		category "Light Freighter"
+		"hull" 11600
+		"required crew" 3
+		"bunks" 3
```
Amazinite: These need zero crew, otherwise killing one will eternally piss off the Archons, which could happen before the player even gets anywhere near meeting an Archon.

Naminson:
I don't mind these becoming a thing that you could see in really sparsely populated areas like the Ember Waste or Korath Ruins, and these actually could be a nice touch to make those regions a little less dry, but I think that putting them in human space would make them really out of place. 

Human space is supposed to be relatively grounded compared to the rest of the game. Yes, there's sci-fi stuff, but people act like people and there's nothing that's too weird, aside from some suggestion of future intrigue with the Quarg, but even that doesn't feel too exotic. When you introduce giant flying space cows, I can't imagine any way that wouldn't be very out of place.

I don't think *human space* really needs anything like these, anyways. Pirates spawn more frequently in uninhabited systems, so they already are different. In fact, they're different in a way that fits way better, since pirates are an established part of human space, and their existence fits the tone of the rest of human space.

data/indigenous.txt
Zitchas: Maybe the Archons are secretly watching over all natively space-faring life?

Zitchas:
@Naminson I agree, to a certain extend. I'd like these to be still found in human space, but I'm thinking of adjusting it so that it is around the periphery rather than throughout, and not very frequent along the civilized-pirate paths, either.

Zitchas:
As an alternative to the current method of implementation, I have considered switching this to using a Person. The only reason I don't, actually, is the fact that a person once killed never returns. If there could be some kind of attribute/variable attached to the person entry that specifies that said person doesn't follow the usual "once killed it never returns" rule, I would switch to using that.

Zitchas:
@Amazinite How is this as an alternative to adding it system by system?

It will be rare, but will occasionally show up in most places I'd like it to be. So long as the player doesn't kill it, it will continue to do so, too. 

I'm not entirely confident in the system filtering code, though.

data/persons.txt
```diff
@@ -768,3 +768,16 @@ ship "Ursa Polaris"
...
+	government "Indigenous Lifeform"
+	frequency 100
+	personality
+		timid unconstrained coward appeasing uninterested mining harvests mute
+	ship "Subsidurial"
+		"never disabled"
+	system
+		neighbor government "uninhabited"
+		neighbor government "Wanderer"
+		neighbor government "Remnant"
```
Amazinite: `system` is a LocationFilter, which means it will work the same way as a `source` or `destination`, which also means that this is looking for a system that is a neighbor to Uninhabited, Wanderer, AND Remnant. 
`neighbor government "uninhabited" "Wanderer" "Remnant` is what you're looking for.

Zitchas: Fixed. Thanks!

data/indigenous.txt
Zitchas: I think this now counts as resolved, as this is now a person that players will occasionally see in uninhabited systems rather than a "frequently" occurring ship. (also, it has been dropped to crew 1 / bunks 1, so it'll have a lower impact on ratings when killed, too.)

data/persons.txt
```diff
@@ -768,3 +768,14 @@ ship "Ursa Polaris"
 	explode "big explosion" 40
 	explode "huge explosion" 50
 
+
+person "Subsidurial"
+	government "Indigenous Lifeform"
```
Amazinite: I still don't believe that this should be part of the indigenous lifeform government, as killing this person ship will anger the Drak. Either this needs a separate government, or it needs 0 crew.

Zitchas: OK. Crew reduced to 0.

data/indigenous.txt
```diff
@@ -8,6 +8,49 @@
...
+ship "Subsidurial"
+	noun "creature"
+	sprite "ship/subsidurial"
+	thumbnail "thumbnail/subsidurial"
+	"never disabled"
+	attributes
+		category "Light Freighter"
+		"hull" 11600
+		"required crew" 0
+		"bunks" 0
```
Amazinite: Zero crew ships *must* have the automaton attribute, as otherwise the game forces at least one crew onto the ship.

Zitchas: Fixed.

PR #4796, opened 2020-02-08
Add script that checks if the Xcode project is complete.

petervdmeer:
**Feature:** This PR adds some scripting to help deal with missing source-files in the XCode-project.

## Feature Details
Intended usage is that if the MacOS compilation fails, then the script as provided in this PR would run to add the missing XCode files and print the fix in the output-logs for the PR owner to apply.

## UI Screenshots
N/A

## Usage Examples
Add some source-code files, then notice them missing for MacOS and copy the entries as printed in the MacOS Github actions to the proper sections in the XCode project to make sure the code also compiles for MacOS.

### DIFF
--- a/.github/workflows/build_parse.yml
+++ b/.github/workflows/build_parse.yml
@@ -120,6 +120,9 @@ jobs:
       run: |
         brew update
         brew install libmad libpng jpeg-turbo sdl2
+    - name: Try fixing XCode entries if required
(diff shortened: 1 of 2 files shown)


petervdmeer:
Testing here doesn't add much value, will test first on PR to auto_testers branch and then re-open here.

PR #4797, opened 2020-02-08
Only handle Engine keyboard inputs when the game is not paused

petervdmeer:
**Bugfix:** This PR addresses an issue regarding Engine keyboard input reading while the game is paused (as reported on Discord)

## Fix Details
The engine is continuously running (at 60 fps), even when the game is paused, to ensure that functions like drawing of the main flight screen are continuously active.

Some functions like movement of ships, movement of projectiles and player commands to his/her flagship and fleet should however be suspended when the engine is not active.

This PR changes the reading of keyboard inputs to only register those inputs when the game is not paused.

## Testing
I didn't have time yet to extensively test this PR; it needs some more testing before we can merge this.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -276,7 +276,7 @@ void AI::IssueMoveTarget(const PlayerInfo &player, const Point &target, const Sy
 
 
 // Commands issued via the keyboard (mostly, to the flagship).
-void AI::UpdateKeys(PlayerInfo &player, Command &activeCommands, bool isActive)
+void AI::UpdateKeys(PlayerInfo &player, Command &activeCommands)
 {
 	shift = (SDL_GetModState() & KMOD_SHIFT);
 	escortsUseAmmo = Preferences::Has("Escorts expend ammo");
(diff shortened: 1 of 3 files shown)


petervdmeer:
I've checked the original code (from before #4715). It looks like the keyboard reading from AI/Engine was continuously active to detect if the player presses a key that would cancel the autopilot, even if the key in the paused context would have a completely different meaning.
If you would have changed the flight-plan in the map using keyboard-inputs, then the autopilot would be cancelled, but if you would have changed the flight-plan using mouse-inputs, then I'm not sure if the autopilot would have been cancelled.

For this PR I changed the code to ignore all flight inputs while the map-screen or any other pop-up is active. We might want to do an update in the future to cancel the auto-pilot when a MAP command is given or to cancel the auto-pilot only when a new route gets set, but I wasn't sure which behavior is most desired here so I choose the most simple option (both in implementation and in explaining the feature).

I performed the following tests:
- Changed target in Map (using keyboard) while jumping
  - OK: The jump continues, but with the new target set
- Changed target in Map (using mouse) while jumping
  - OK: The jump continues, but with the new target set
- Changed target in Map to current system (using mouse) while jumping
  - OK: The jump gets cancelled silently (due to no target available)
- Pressed L (for landing) on map screen
  - OK: Landing doesn't get started
- Pressed L (for landing) on flight screen
  - OK: Auto-pilot starts landing.

PR #4801, opened 2020-02-09
feat: Add gun-port angles and gun convergence setting

petervdmeer:
**Feature:** This PR implements a feature to have gunports facing in different directions than forward and to control convergence of guns.

## Related features
This feature uses the syntax as proposed for #4539 (bay facing) and #4749 (engines facing) as discussed with Amazinite.

## Feature Details
This PR adds two new keywords to gunports (hardpoints); `angle` to indicate the direction in which the gun faces and `parallel` to indicate that weapons installed in the gunport should not get a converged focus aiming point (along the centerline of the ship).

This PR also adds the `parallel` keyword to (weapon) outfits to indicate that the weapon should not get a converged focus aiming point.

Examples of the keywords on the gunport from attached savegame:
```
gun 86 122 "Beam Laser"
	angle 90
	parallel
gun -86 -106 "Beam Laser"
	angle -90
```
Example of the parallel keyword used in the Welder outfit:
```
outfit "Welder"
	category "Guns"
	weapon
		parallel
```

Guns fire non-converged when the outfit weapon itself has the `parallel` keyword or when the port has the `parallel` keyword (or when both have the keyword). This also ensures that convergence remains the default for existing gunports and outfits.

## UI Screenshots
This is an example ship using gunport angles and gunport parallelism:
gun_orientations

This is a Kestrel firing a short-range weapon without parallel enabled:
converged_short_range

This is the same Kestrel firing the same short-range weapon with parallel in its outfit definition:
parallel_short_range

## Usage Examples
There is currently no AI-support for non-forward-facing guns, so I expect the gun angles feature to be mostly useful for plugin creators (at least until we teach auto-fire to work with non-forward facing guns).

The `parallel` keyword on outfits can be useful for weapons with a relatively short range and for homing missile type weapons (which don't need to converge, but which seek out their targets by themselves). See above screenshots with the Kestrel for the short range example.

## Testing Done
Tested gunport angles and convergence:
- with beam-lasers, proton guns and Sidewinder launcher.
- in forward, 45 degrees, sideways and backwards direction.
- with converging weapons and non-converging(parallel) weapons.
- loading and saving (usage in savegames).
Here is the savefile [Flying Dutchman.txt](https://github.com/endless-sky/endless-sky/files/4177539/Flying.Dutchman.txt that was used for testing gunports.

Tested outfits parallel keyword:
- with Welder in a Kestrel without parallel keyword.
- with Welder in a Kestrel with parallel keyword on outfit.
Here is the datafile [welder_weapon.txt](https://github.com/endless-sky/endless-sky/files/4206435/welder_weapon.txt that was used for testing the parallel keyword.

Also tested the combination (using the datafile and the welder_weapon file).

Screenshots from the tests are also in this PR.

## Performance Impact
None expected, the angle calculation for weapons is done when a weapon is installed. Not when a weapon is fired.


downRightCoder256:
it could be used on missile only ships. i love it

petervdmeer:
Merged parallel weapons(outfits) into this branch. Both features are quite related.

source/Armament.cpp
```diff
@@ -26,17 +26,18 @@ using namespace std;
...
 }
 
 
 
 // Add a turret hardpoint (omnidirectional weapon).
 void Armament::AddTurret(const Point &point, const Outfit *outfit)
 {
-	hardpoints.emplace_back(point, true, outfit);
+	bool isParallel = false;
+	hardpoints.emplace_back(point, Angle(0.), true, isParallel, outfit);
```
Amazinite, line 40: I'd just do this, as why would you create an isParallel variable so that we know what is being set as false but not have a variable for whatever is set as true? If someone wants to know what these booleans are for, I'd say to just look at Hardpoint's constructor(s).
```suggestion
	hardpoints.emplace_back(point, Angle(0.), true, false, outfit);
```

source/Weapon.h
```diff
@@ -37,6 +37,7 @@ class Weapon {
 	// Load from a "weapon" node, either in an outfit or in a ship (explosion).
 	void LoadWeapon(const DataNode &node);
 	bool IsWeapon() const;
+	bool IsParallel() const;
```
Amazinite, line 40: I feel like IsParallel would make more sense to be organized under IsStreamed,

Amazinite (changes requested):

tehhowch (commented):
I don't recall how far I got in this review, but I saw I had staged comments here 😅

source/Ship.cpp
```diff
@@ -641,6 +651,16 @@ void Ship::Save(DataWriter &out) const
 					hardpoint.GetOutfit()->Name());
 			else
 				out.Write(type, 2. * hardpoint.GetPoint().X(), 2. * hardpoint.GetPoint().Y());
+			double hardpointAngle = hardpoint.GetBaseAngle().Degrees();
+			if(hardpoint.IsParallel() || hardpointAngle)
+			{
+				out.BeginChild();
+				if(hardpointAngle)
```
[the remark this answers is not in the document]
Amazinite: Ref: https://github.com/endless-sky/endless-sky/pull/5161#discussion_r454525652

source/Armament.cpp
```diff
@@ -26,17 +26,18 @@ using namespace std;
...
+	hardpoints.emplace_back(point, angle, false, isParallel, outfit);
 }
 
 
 
 // Add a turret hardpoint (omnidirectional weapon).
 void Armament::AddTurret(const Point &point, const Outfit *outfit)
 {
-	hardpoints.emplace_back(point, true, outfit);
+	bool isParallel = false;
```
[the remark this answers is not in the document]
petervdmeer: My original reason was readability of the code, but I fixed this using Amazinites suggestion.

petervdmeer:
Thanks for your reviews. All review-comments so far have been processed.

source/Weapon.h
```diff
@@ -156,6 +157,7 @@ class Weapon {
 	bool isSafe = false;
 	bool isPhasing = false;
 	bool isDamageScaled = true;
+	bool isParallel = false;
```
tehhowch, line 160: Let's add a line comment above `isParallel` that indicates what effect this flag has on a given weapon. If someone is reviewing the game's class headers and wondering how they can customize the game's weapons, they shouldn't need to dive all the way into Hardpoint/Armament to see what this flag does.

source/Hardpoint.h
```diff
@@ -95,6 +100,8 @@ class Hardpoint {
 	bool isTurret = false;
 	bool isFiring = false;
 	bool wasFiring = false;
+	// Indicates if this hardpoint disallows converging (guns only).
+	bool isParallel = false;
```
tehhowch, line 104: Let's move this (and `isTurret`) up in order here, so that the members we are initializing in the constructor are adjacent in the class's memory layout. (This will also creates two logical groups of hardpoint fields - the static as-initialized ones, and the dynamic ones that are frequently modified.)

tehhowch (approved):
Two quibbles then 👍

Amazinite (approved):

petervdmeer:
I added some comments to isParallel to explain what the variable does.

Zitchas:
Just to give people a fun demo of something that can be done using these gun-port angles:

https://www.youtube.com/watch?v=GVnxRp1_nEk

Engines that will actually burn stuff! This was done by replacing the Heron's engines with gun ports like this:

```
	gun -111 307.5
		angle 180
		parallel
	gun 112 307.5
		angle 180
		parallel
	gun -86 289 1.5
		angle 180
		parallel
	gun 86 289 1.5
		angle 180
		parallel
	gun -52 274 1.5
		angle 180
		parallel
	gun 52 274 1.5
		angle 180
		parallel
	gun 0 328 1.5
		angle 180
		parallel
```

And then giving the standard flamethrower `"firing force" 10`

And then equipping the heron with enough flamethrowers to fill all those ports.

PR #4813, opened 2020-02-14
feat: Allow some weapons to be parallel / non-converged.

petervdmeer:
**Feature:** This PR implements a feature to allow weapons to be parallel, non-converged

## Feature Details
This PR adds a new keyword `parallel` that can be used in outfit weapon definitions to indicate that the weapon should not get a converged focus aiming point.

The `parallel` keyword is also used for gunport angles in https://github.com/endless-sky/endless-sky/pull/4801. Intended combination of those features is that guns fire non-converged when the outfit weapon itself has the `parallel` keyword or when the port has the `parallel` keyword (or when both have the keyword).

## UI Screenshots
(The definition of the weapon used is attached)
This is a Kestrel firing a short-range weapon without parallel enabled:
converged_short_range

This is the same Kestrel firing the same short-range weapon with parallel in its definition:
parallel_short_range

## Usage Examples
This feature/keyword can be useful for weapons with a relatively short range and for homing missile type weapons (which don't need to converge, but which seek out their targets by themselves). See above for the short range example.

## Testing Done
Screenshots above are from two short tests of this feature. Here is the weapons data-file used for testing:
[welder_weapon.txt](https://github.com/endless-sky/endless-sky/files/4206435/welder_weapon.txt)

## Performance Impact
None expected, the angle calculation for weapons is done when a weapon is installed. Not when a weapon is fired.

### DIFF
--- a/source/Hardpoint.cpp
+++ b/source/Hardpoint.cpp
@@ -77,6 +77,10 @@ Angle Hardpoint::HarmonizedAngle() const
 	if(!outfit)
 		return Angle();
 	
+	// Weapons that fire in parallel beams don't get a harmonized angle.
+	if(outfit->IsParallel())
+		return Angle();
+	
 	// Find the point of convergence of shots fired from this gun. That is,
 	// find the angle where the projectile's X offset will be zero when it
 	// reaches the very end of its range.
--- a/source/Weapon.cpp
+++ b/source/Weapon.cpp
@@ -46,6 +46,8 @@ void Weapon::LoadWeapon(const DataNode &node)
 			isPhasing = true;
 		else if(key == "no damage scaling")
 			isDamageScaled = false;
+		else if(key == "parallel")
+			isParallel = true;
 		else if(child.Size() < 2)
 			child.PrintTrace("Skipping weapon attribute with no value specified:");
 		else if(key == "sprite")
@@ -214,6 +216,13 @@ void Weapon::LoadWeapon(const DataNode &node)
 bool Weapon::IsWeapon() const
 {
 	return isWeapon;
+}
+
+
+
+bool Weapon::IsParallel() const
+{
+	return isParallel;
 }
 
 
--- a/source/Weapon.h
+++ b/source/Weapon.h
@@ -37,6 +37,7 @@ public:
 	// Load from a "weapon" node, either in an outfit or in a ship (explosion).
 	void LoadWeapon(const DataNode &node);
 	bool IsWeapon() const;
+	bool IsParallel() const;
 	
 	// Get assets used by this weapon.
 	const Body &WeaponSprite() const;
@@ -155,6 +156,7 @@ private:
 	bool isSafe = false;
 	bool isPhasing = false;
 	bool isDamageScaled = true;
+	bool isParallel = false;
 	
 	// Attributes.
 	int lifetime = 0;


Amazinite:
Sounds similar to the stream vs cluster fire behavior. Great idea. 👍

Zitchas:
I can't verify the code for this, but it definitely looks like a solid addition to the repertoire of weapon attributes. That being said, it *looks* like a simple change...

tehhowch:
Default weapon behavior is convergent fire; this PR adds parallel. Do we care to add a way to specify a weapon's fire is always anticonverged? (diverged, divergent, spreading, etc.)

Is it possibly confusing to content creators, or to players, how one can determine the angle at which a weapon will fire, in a world where hardpoints and guns each can have angles? What impact is there on AI ship aiming when mixtures of these weapons are present?

petervdmeer:
> Default weapon behavior is convergent fire; this PR adds parallel. Do we care to add a way to specify a weapon's fire is always anticonverged? (diverged, divergent, spreading, etc.)

I can think of the `inaccuracy` property as a specific case of spreading, but such spreading would be fine both for convergent weapons (spreading around the focus point) and non-convergent weapons (spreading around the fire-axis of the gunport).

> Is it possibly confusing to content creators, or to players, how one can determine the angle at which a weapon will fire, in a world where hardpoints and guns each can have angles?

I only see value in the gunports getting angles; I don't see value of putting angles on guns themselves. The way to determine the angle would then be:
- Angle as set by the gunport (straight forward if not set)
- Corrected for the convergence (based on weapon-range and off-centerness of gunport) if both the weapon and the gunport support converging.

The keyword `parallel` might be a little bit confusing, I earlier considered `non focussed` but that sounded even more confusing. Maybe I should rename the keyword to `"no convergence"`?

> What impact is there on AI ship aiming when mixtures of these weapons are present?

I didn't check yet. My expectation is that the AI works best with converged weapons, but that it will also work fine with homing weapons that are non-converged.

Amazinite:
Whether a weapon converges its firing angle or fires in parallel, the ship is still going to want to fire at center mass of the target for maximum impact, so I don't see `parallel` weapons requiring a change in AI.

Zitchas:
I agree with Petervdmeer: I don't think it makes sense to put angles on the guns themselves. Guns go in the port they are assigned to, done. I can see having the possibility of setting them to "non-convergent" that allows them to override whatever the setting for the gun port is and just fire straight forward.

Although that being said, which ones takes priority? A gun port that has a few degrees of angle can be easily corrected by adjusting the installation a bit. Having a gun or missile override that is totally understandable. If a ship has something more unique, though. (say, a gunboat that has ports angled 45 degrees off center to launch a large spread of missiles, or maybe a port that points backwards for shooting at things chasing it)... Those cases don't make sense for being overridden. 

I think what I'm getting at here is that maybe we need three steps:
1) The port itself specifies the default angle the gun will point.
2) The gun will normally just use the default angle, but has two attributes that will either adjust the angle it is pointing to either a) actual convergence, or b) parallel. These two keywords will override the specified angle.
3) the port can have a "forced angle" attribute. This overrides the convergence/parallel attributes and dictates that the weapon will fire at whatever the specified angle is.

So, probably all the ports that are currently in the game will just fall into category 1 where they have either no angle or a very small angle - probably something that provides convergence or a little bit beyond convergence for their typical loadout.

Most weapons would probably not have any attributes, but a few might have forced convergence, and some (like missiles, and possibly mass-area effects like flamethrowers) might have forced parallel. 

No current slots would need the "forced angle" attribute; but I would anticipate a few ships might have odd port alignments that would utilize this. (for instance, Escape Velocity had an interceptor that had a rear turret. We don't have the ability to restrict turret fields of fire yet, but a rear-facing gun could work similarly)

petervdmeer:
@Zitchas The priorities as I planned them for the code (in combination with #4801) are:
1. The gun points at the angle as set in the gun-port. (No discussion here.)
2. If `parallel` is set on the gun-port, then don't apply convergence.
3. If `parallel` is set on the gun itself, then don't apply convergence.
4. If no non-convergence is set on port or gun itself, then apply convergence.

The only type of overriding is to disable converging (on port or on gun). I didn't see a need to actively override something towards enable; if converging isn't supported for some port/gun combination then we just don't use it there.

Here is the screenshot from #4801 that shows why convergence also makes sense for gun-port-angles other than facing forward (left-facing guns converge, right-facing guns and blasters have the parallel keyword):
gun_orientations

tehhowch:
Re: anticonverged, I had basically thought of it as applying the convergence angle in the other direction. It would mostly be useful for launching homing projectiles, such that they would approach a "forward" target from more distinct angles.
In the current ES antimissile implementation, there is no "A/M turn rate" that makes such projectiles harder to acquire and defeat, but if A/M turn rate were a thing, then missile boats would prefer diverse projectile trajectories to increase hit rate.

Zitchas:
OK, that sounds good to me.

And I'm still hoping that we'll get true AM behaviour someday. :)

petervdmeer:
> Re: anticonverged, I had basically thought of it as applying the convergence angle in the other direction. It would mostly be useful for launching homing projectiles, such that they would approach a "forward" target from more distinct angles.

I like the idea of projectiles arriving from more distinct angles, but this doesn't look like the right mechanism to accomplish that.
Many projectile weapons have a long range, causing the convergence angle to be small anyway.
I also have a concern that firing outwards from forward facing guns could look a little bit unprofessional.

If projectiles approaching from more angles is desired, then I would prefer to handle that not during launch, but by adding some steering/circling in the path from launch to target.

Zitchas:
I agree that it might look weird for, say, interceptors or gunboats to launch missiles in a spread, but for big ships, I think it makes perfect sense. Most especially for things like the Bactrian. Not every hardpoint can face forwards, after all. This video, for instance, shows a mass missile volley being launched off of a variety of ships, which I think looks really cool. [EVE-Online Inferno trailer](https://www.youtube.com/watch?v=kED11aGobUk&list=PLF614A7A6461E61E1&index=21) (also noteworthy, vertical launch systems are common on larger naval vessels, whereas direct fire missiles are typically only on aircraft and small missile boats. ([Lockheed martin vertical launch promotional video](https://www.youtube.com/watch?v=z2KSqSvKz-Y). Jump forward to about the minute mark for the best footage)

That being said, I think that debate is probably getting ahead of ourselves. This PR is about getting the capacity to make these adjustments into the game. Implementing them on ships & weapons should be a separate thing. That being said, the issue of where gunports are located comes back to design philosophy as well: Regardless of what ships in ES are classified as, everything is built like and handles like a fighter. I'd like to see us move away from that, and this PR opens this up as a possibility.

Zitchas:
I don't know why it never occured to me to link this, but for examples of non-forward facing guns, check out the AC-130 Spectre: [wikipedia page](https://en.wikipedia.org/wiki/Lockheed_AC-130)

That's a 105 mm (4.13 in) M102 howitzer sticking out the side (amongst all the other weaponry...)

800px-AC-130H_Spectre_(2152981898)

Anyway, looking forward to seeing this merged.

petervdmeer:
Merged into #4801 since the features are very similar.

PR #4817, opened 2020-02-15
Add script that checks if the Xcode project is complete (and that suggests a fix)

petervdmeer:
**Feature:** This PR adds a script to get an automatically generated fix for incomplete Xcode projects

## Feature Details
The script automatically runs as a GitHub action. When it finds source-files that are not present in the Xcode project, then it determines the required delta to fix the project and prints the delta as output in the GitHub action so that the creator of the PR can apply the delta as fix.

The procedure to fix incomplete XCode projects without this GitHub action is to ask on Discord if a MacOS user would be willing to check-out the relevant branch and send the diff for the Xcode project to the PR owner.

## Usage Examples
 - Add new files in a PR (for example in #4704 or #4370) and then see the GitHub action fail. For most cases, a patch can then be downloaded from the action, and used in CLI with `git apply`
 - Run the test locally via CLI (after installing mod-pbxproj too!), and commit the changes.
When the update failed, the script will indicate manual action is necessary.

## Testing Done
Tested with #4370 and got the relevant output for that PR.
tehhowch did a more elaborate test with moving files and adding a file into a subdirectory

### DIFF
--- a/.github/workflows/xcode_check.yml
+++ b/.github/workflows/xcode_check.yml
@@ -20,7 +20,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get install python3-setuptools python3-wheel
-        sudo pip3 install docopt openstep_parser
+        pip3 install docopt openstep_parser --user
         pip3 install pbxproj --user
     - name: Execute xcode completeness check
       run: |
--- a/.github/workflows/xcode_check.yml
+++ b/.github/workflows/xcode_check.yml
@@ -24,6 +24,12 @@ jobs:
         sudo apt-get update
         sudo apt-get upgrade
         pip3 install docopt openstep_parser --user
+    - name: Clone and test pbxproj
+      run: |
+        git clone https://github.com/kronenthaler/mod-pbxproj.git
+        cd mod-pbxproj && python3 setup.py test
+    - name: Install mod-pbxproj
+      run: |
         pip3 install pbxproj --user
     - name: Execute xcode completeness check
       run: |
--- a/.github/workflows/xcode_check.yml
+++ b/.github/workflows/xcode_check.yml
@@ -18,7 +18,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get install python3-setuptools
-        pip3 install pbxproj
+        sudo pip3 install pbxproj
     - name: Execute xcode completeness check
       run: |
         ./tests/check_xcode.sh
--- a/.github/workflows/xcode_check.yml
+++ b/.github/workflows/xcode_check.yml
@@ -30,7 +30,7 @@ jobs:
         cd mod-pbxproj && python3 setup.py test
     - name: Install mod-pbxproj
       run: |
-        pip3 install pbxproj --user
+        cd mod-pbxproj && python3 setup.py install --user
     - name: Execute xcode completeness check
       run: |
         ./tests/check_xcode.sh
--- a/.github/workflows/xcode_check.yml
+++ b/.github/workflows/xcode_check.yml
@@ -1,0 +1,24 @@
+name: Check xcode completeness
+
+on:
+  push:
+    paths:
+    - 'source/**'
+    - 'tests/**'
+  pull_request:
+    paths:
+    - 'source/**'
+    - 'tests/**'
+
+jobs:
+  check-xcode-completeness:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install dependencies
+      run: |
+        sudo apt-get install python3-setuptools
+        pip3 install pbxproj
+    - name: Execute xcode completeness check
+      run: |
+        ./tests/check_xcode.sh
--- a/.github/workflows/xcode_check.yml
+++ b/.github/workflows/xcode_check.yml
@@ -20,17 +20,10 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get install python3-setuptools python3-wheel
-        sudo rm /etc/apt/sources.list.d/* && sudo dpkg --clear-avail # Speed up installation and get rid of unwanted lists
-        sudo apt-get update
-        sudo apt-get upgrade
         pip3 install docopt openstep_parser --user
-    - name: Clone and test pbxproj
-      run: |
-        git clone https://github.com/kronenthaler/mod-pbxproj.git
-        cd mod-pbxproj && python3 setup.py test
     - name: Install mod-pbxproj
       run: |
-        cd mod-pbxproj && python3 setup.py install --user
+        pip3 install pbxproj --user
     - name: Execute xcode completeness check
       run: |
         ./tests/check_xcode.sh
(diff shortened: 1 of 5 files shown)


petervdmeer:
It looks like this script needs feature https://github.com/kronenthaler/mod-pbxproj/issues/199 as requested for mod-pbxproj to place the new files under the right parent.

tehhowch:
If we can do that using their API, i.e. from within a Python script, then we should do that instead of a shell script.

petervdmeer:
> If we can do that using their API, i.e. from within a Python script, then we should do that instead of a shell script.

Yes, or we send a PR to that project to get the API option also available on the command-line. 
 
> We should also make sure that subfolders are properly handled. We don't have any now, true, but with `walk` it should be easy enough to have the proper paths.

When it comes to Xcode projects, then sources is already a subfolder/subproject of project-root.

MCOfficer:
kronenthaler/mod-pbxproj#262 got merged.

tests/check_xcode.sh
```diff
@@ -0,0 +1,67 @@
...
+cd ..
+
+ESTOP=$(pwd)
+XPROJECT=${ESTOP}/EndlessSky.xcodeproj
+
+echo "script-dir: ${HERE}"
+echo "es-top-dir: ${ESTOP}"
+
+NUM_ADDED=0
+for FILE in $(ls -1 source)
```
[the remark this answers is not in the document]
MCOfficer: not files in subdirectories, but it will loop over the subdirectory names itself.
a better alternative would be `find source -type f -printf "%f\n"` (`type` excludes directories, `printf` ensures only filenames are printed)

tests/check_xcode.sh
```diff
@@ -0,0 +1,67 @@
...
+do
+	# Check if the file is already in the XCode project
+	cat EndlessSky.xcodeproj/project.pbxproj | grep "$FILE" > /dev/null
```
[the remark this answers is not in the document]
MCOfficer: it's more convenient *in general*, since you could also cat multiple files, or pipe the output multiple times. in this case, it should make no difference.

tests/check_xcode.sh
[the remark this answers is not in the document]
petervdmeer: I used a different implementation that preserves the paths under source (for files in subdirectories), since I have the impression that Xcode requires the paths to be complete (not just the filenames).

tests/check_xcode.sh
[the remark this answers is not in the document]
petervdmeer: Yes, it is indeed more a habit than than that it makes a difference. I changed the code to use grep directly.

.github/workflows/xcode_check.yml
```diff
@@ -0,0 +1,24 @@
+name: Check xcode completeness
+
+on:
+  push:
+    paths:
+    - 'source/**'
+    - 'tests/**'
```
tehhowch: Should this check really run on changes to `tests/`? Wouldn't changes to the Xcode project itself be more apt?

.github/workflows/xcode_check.yml
```diff
@@ -0,0 +1,24 @@
...
+  pull_request:
+    paths:
+    - 'source/**'
+    - 'tests/**'
+
+jobs:
+  check-xcode-completeness:
```
tehhowch: What do you think about adding one of these checks for Code::Blocks, too? It's more accessible, but I'm not the only one who uses something else.

It'd be nice to have both project file checks in the same action, but running both when either one fails and having the steps named well seems like it might not be fun

tests/check_xcode.sh
```diff
@@ -0,0 +1,67 @@
...
+		NUM_ADDED=$(( NUM_ADDED + 1 ))
+	fi
+done
+
+if [ ${NUM_ADDED} -gt 0 ]
+then
+	echo "You should add ${NUM_ADDED} files"
+	echo "Example of the diff to apply:"
+	echo ""
+	git diff EndlessSky.xcodeproj/project.pbxproj
```
tehhowch: Should this use the variable too?

tests/check_xcode.sh
```diff
@@ -0,0 +1,67 @@
...
+#
+# mod-pbxproj should already have been installed (through PIP) before
+# running this script.
+
+
+# Determine path of the current script
+HERE=$(cd `dirname $0` && pwd)
+
+# Go to toplevel ES directory
+cd ${HERE}
```
tehhowch: `cd "${HERE}/.."`?

I'm assuming this is so that we can run the test from a shell that is in any working directory, e.g. `$ ../../path/to/testfile.sh`

petervdmeer: Yes, that is indeed what this is for.

tests/check_xcode.sh
petervdmeer: Not critical, but yes it is nicer to use the variable here.

.github/workflows/xcode_check.yml
petervdmeer: Good point, changed to check only updates to the script itself.

.github/workflows/xcode_check.yml
petervdmeer: Sounds like a good idea, I'm in favor of checking such project files automatically.

But can we keep that for a separate PR? I currently don't know what exactly needs to be done for Code::Blocks, while I do see issues with XCode projects in combination with PRs that add new files by developers that don't have a Mac.

tests/check_xcode.sh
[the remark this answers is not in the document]
tehhowch: Can you verify that somehow? E.g. open a PR to your fork, based from this branch, that creates and uses `source/foo/bar.*`, and then commit the diff that this check suggests to ensure the project builds?

Also, should we provide instructions for how one applies the generated diff?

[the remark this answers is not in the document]
petervdmeer: > Can you verify that somehow? E.g. open a PR to your fork, based from this branch, that creates and uses `source/foo/bar.*`, and then commit the diff that this check suggests to ensure the project builds?

I already did some basic verification in a Virtual Machine on my laptop (added a subdirectory called subdir with in it SubFile.cpp and SubFile.h). The SubFile files get added into the XCode project file at the locations where I expect them and the relative paths looked good.

Getting this to build on XCode would be more work, since I would also need to put some sensible content in the SubFile files and trigger the PRs and CI jobs. Let's keep that for when we actually start moving things to sub-directories.

> Also, should we provide instructions for how one applies the generated diff?

I don't know; this is not an end-user feature, but a feature for a developer that is making a non-trivial change (since he/she needs to add new files). Can we expect such developer to know (or at least be able to lookup) how to apply a diff?

If we would have hundreds of new files added every week, then we could also consider to make applying the diff very user-friendly by putting the diff in an automatically generated PR and thus change the interface to apply the diff to just a simple merge-button. But this looks quite a bit overkill to me with the current volume of incoming changes on ES.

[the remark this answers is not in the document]
MCOfficer: > Also, should we provide instructions for how one applies the generated diff?

I'd argue that a person who modifies the XCode project and presumably adds new libraries should be able to apply diffs.

PR #4822, opened 2020-02-16
Fill the default value in the save and ship rename dialogs

jostephd:
**Feature:** This PR implements the feature request detailed and discussed ~~in issue #{{insert number}}~~ below

## Feature Details
The "rename ship" dialog will start prefilled with the name of the ship, rather than empty.

The save dialog will start prefilled with the date, rather than empty. The behavior is unchanged, I think: currently if the user clicks "OK" in the save dialog when the textbox is empty, the date is used.

## UI Screenshots
The following two dialogs are affected. In master they both start with the textbox empty, the screenshot shows the new value:

2020-02-16-184741_619x380_scrot
2020-02-16-184728_260x163_scrot

## Usage Examples
N/A

## Testing Done
I opened the dialogs and verified the textbox starts prefilled with the correct value, even when the day-of-month is single digit.

## Performance Impact
N/A

### DIFF
--- a/source/Dialog.h
+++ b/source/Dialog.h
@@ -50,8 +50,7 @@ template <class T>
 	Dialog(T *t, void (T::*fun)(int), const std::string &text, int initialValue);
 	
 template <class T>
-	Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text,
-			const std::string& initialValue = "");
+	Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text, std::string initialValue = "");
 	
 template <class T>
 	Dialog(T *t, void (T::*fun)(), const std::string &text);
@@ -119,12 +118,10 @@ Dialog::Dialog(T *t, void (T::*fun)(int), const std::string &text, int initialVa
 
 
 template <class T>
-Dialog::Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text,
-		const std::string& initialValue)
-	: stringFun(std::bind(fun, t, std::placeholders::_1))
+Dialog::Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text, std::string initialValue)
+	: stringFun(std::bind(fun, t, std::placeholders::_1)), input(initialValue)
 {
 	Init(text);
-	input = initialValue;
 }
 
 
--- a/source/Dialog.h
+++ b/source/Dialog.h
@@ -50,7 +50,8 @@ template <class T>
 	Dialog(T *t, void (T::*fun)(int), const std::string &text, int initialValue);
 	
 template <class T>
-	Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text);
+	Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text,
+			const std::string& initialValue = "");
 	
 template <class T>
 	Dialog(T *t, void (T::*fun)(), const std::string &text);
@@ -118,10 +119,12 @@ Dialog::Dialog(T *t, void (T::*fun)(int), const std::string &text, int initialVa
 
 
 template <class T>
-Dialog::Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text)
+Dialog::Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text,
+		const std::string& initialValue)
 	: stringFun(std::bind(fun, t, std::placeholders::_1))
 {
 	Init(text);
+	input = initialValue;
 }
 
 
--- a/source/LoadPanel.cpp
+++ b/source/LoadPanel.cpp
@@ -198,8 +198,15 @@ bool LoadPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, boo
 		if(it == files.end() || it->second.empty() || it->second.front().first.size() < 4)
 			return false;
 		
+		const Date &today = player.GetDate();
+		const int year = today.Year(), month = today.Month(), day = today.Day();
+		const string &defaultFilename =
+			to_string(year) + "-" +
+			(month < 10 ? "0" : "") + to_string(month) + "-" +
+			(day < 10 ? "0" : "") + to_string(day);
 		GetUI()->Push(new Dialog(this, &LoadPanel::SnapshotCallback,
-			"Enter a name for this snapshot, or leave the name empty to use the current date:"));
+			"Enter a name for this snapshot, or leave the name empty to use the current date:",
+			defaultFilename));
 	}
 	else if(key == 'R' && !selectedFile.empty())
 	{
--- a/source/LoadPanel.cpp
+++ b/source/LoadPanel.cpp
@@ -199,8 +199,10 @@ bool LoadPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, boo
 			return false;
 		
 		const Date &today = player.GetDate();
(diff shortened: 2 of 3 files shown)


petervdmeer:
> The "rename ship" dialog will start prefilled with the name of the ship, rather than empty.

Nice to see a fix for this, this is something that bothered me in the past when I had a typo of 1 character in my ships name.

jostephd:
Force pushed to remove the unused constructor's declaration as well as its definition.

source/LoadPanel.cpp
```diff
@@ -198,8 +198,15 @@ bool LoadPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, boo
 		if(it == files.end() || it->second.empty() || it->second.front().first.size() < 4)
 			return false;
 		
+		const Date &today = player.GetDate();
+		const int year = today.Year(), month = today.Month(), day = today.Day();
+		const string &defaultFilename =
```
tehhowch, line 203: This variable declaration creates a constant reference to a temporary value. Just make it a regular string.

source/LoadPanel.cpp
```diff
@@ -198,8 +198,15 @@ bool LoadPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, boo
 		if(it == files.end() || it->second.empty() || it->second.front().first.size() < 4)
 			return false;
 		
+		const Date &today = player.GetDate();
+		const int year = today.Year(), month = today.Month(), day = today.Day();
```
tehhowch, line 202: This pattern of multiple declarations in a single statement should be avoided.

source/Dialog.h
```diff
@@ -44,13 +44,12 @@ class Dialog : public Panel {
...
 	// case, the callback is called only if the user selects "ok", not "cancel."
-template <class T>
-	Dialog(T *t, void (T::*fun)(int), const std::string &text);
 template <class T>
 	Dialog(T *t, void (T::*fun)(int), const std::string &text, int initialValue);
 	
 template <class T>
-	Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text);
+	Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text,
+			const std::string& initialValue = "");
```
tehhowch, line 52: Lose the `const &` from the initial value here, so that the compiler can omit copying in some cases. This can also be on the same line as above.

Also, is a default value necessary (are there dialogs with a callback that takes a string but don't provide an initial value)?

source/Dialog.h
```diff
@@ -44,13 +44,12 @@ class Dialog : public Panel {
 	// Three different kinds of dialogs can be constructed: requesting numerical
 	// input, requesting text input, or not requesting any input at all. In any
 	// case, the callback is called only if the user selects "ok", not "cancel."
-template <class T>
-	Dialog(T *t, void (T::*fun)(int), const std::string &text);
```
tehhowch, line 48: Keep this template. While it isn't used right now, template code that isn't used is not included in the binary.

(This method was originally used by BankPanel before [support for shorthand numbers like "800k" was added](https://github.com/endless-sky/endless-sky/commit/8fb7692023ac4b0ca6f55de7a3c5ddbbc8c8a134).)

source/Dialog.h
```diff
@@ -118,10 +108,12 @@ Dialog::Dialog(T *t, void (T::*fun)(int), const std::string &text, int initialVa
 
 
 template <class T>
-Dialog::Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text)
+Dialog::Dialog(T *t, void (T::*fun)(const std::string &), const std::string &text,
+		const std::string& initialValue)
 	: stringFun(std::bind(fun, t, std::placeholders::_1))
 {
 	Init(text);
+	input = initialValue;
```
tehhowch, line 116: as is done for `stringFun`, initialize input directly with the initial value, rather than assign.

source/ShipInfoPanel.cpp
```diff
@@ -138,7 +138,8 @@ bool ShipInfoPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command,
 		GetUI()->Push(new PlayerInfoPanel(player));
 	}
 	else if(key == 'R' || (key == 'r' && shift))
-		GetUI()->Push(new Dialog(this, &ShipInfoPanel::Rename, "Change this ship's name?"));
+		GetUI()->Push(new Dialog(this, &ShipInfoPanel::Rename, "Change this ship's name?",
+					(*shipIt)->Name()));
```
tehhowch, line 142: This can also be on the line above.

source/LoadPanel.cpp
jostephd: Done.

source/LoadPanel.cpp
jostephd: Done

source/Dialog.h
jostephd: Done. Thanks for the reference.

source/ShipInfoPanel.cpp
jostephd: Done.

source/Dialog.h
jostephd: Done.

source/Dialog.h
jostephd: > Lose the `const &` from the initial value here, so that the compiler can omit copying in some cases. This can also be on the same line as above.

Done.

> Also, is a default value necessary (are there dialogs with a callback that takes a string but don't provide an initial value)?

Yes, there are such dialogs:

https://github.com/endless-sky/endless-sky/blob/82487024212a7a1f93fad982048a933fe6d83631/source/BankPanel.cpp#L269-L275

Should the default value be kept then? Or do you see another approach?

jostephd:
Quick question, why is `string foo` preferable to `const string& foo`? In both cases (`defaultFilename` and `initialValue`), I used a const reference because I thought that would _save_ copying, not _add_ copying...

tehhowch:
> Quick question, why is `string foo` preferable to `const string& foo`? In both cases (`defaultFilename` and `initialValue`), I used a const reference because I thought that would _save_ copying, not _add_ copying...

It's not easy to tell if it will help or hurt unless you can see both sides of the equation. Here we know that -- no matter what `initialValue` is -- initialValue is used to set the value of Dialog's class member field `input`.
If `initialValue` comes from a `const &` lvalue like `(*shipIt->Name())`, we have a mismatch -- a non-const thing (`input`) is given a `const &` thing (`initialValue`), and the only way to convert a `const &` into something that can be modified is to copy it.
If `initialValue` comes from a temporary, the compiler may be able to use return value optimizations, i.e. determine if the thing is going straight from A to B, so it can _make the thing in B_. We can coax this behavior out a bit more as well, by using e.g. `std::move` to signal that some would-be-lvalue can be treated as though it won't exist there anymore, and the receiving function can do whatever it wants with it.

I can't speak to any additional optimizations that would have been applied to the previous version of this where initialValue was `const & = ""`, but if we `std::move(defaultFilename)`, the compiler should be most free to optimize it. For the other uses of dialog that match with this string callback, but don't provide an initial value, `initialValue` will be similarly free to be optimized away.

And as always, YCMV, and we shouldn't stress about minuscule performance improvements like this that aren't in a tight runloop. The biggest issue was with the const refs to temporaries.

jostephd:
@tehhowch Thanks for explaining that. It all makes sense. (And yeah, I wasn't too worried about the cost copying two short strings. I was asking more for future reference.)

Issue #4723, opened 2019-12-26
Ships with scanners will build up over time in busy systems


Amazinite:
**Describe the bug**
If a ship has a scanner, it will attempt to scan any ship that has not yet been scanned by its government before it does anything else. In high traffic systems where new ships are always entering, this causes ships with scanners to always be trying to scan a ship. As time goes on and more ships with scanners enter the system, they built up to large numbers.

**To Reproduce**
1. Enter Sol and wait for a while.
2. Observe how the number of Gunboats increases over time and they rarely leave.

**Expected behavior**
Ships with scanners should either not attempt to scan a ship that an ally is also attempting to scan, or ships with scanners should only scan a limited number of ships, at which point they then no longer scan any ships in the system.

**Screenshots**
Here's an image of nearly 20 Gunboats in Sol with less than 15 other ships, after having waited in Sol for 5 to 10 minutes.
![image](https://media.discordapp.net/attachments/536900466655887360/659438947092398130/unknown.png)

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Self-compiled
 - Version: Up to date as of this issue (6d775e5).

warp-core:
#7229 might be able to help with this.

Amazinite:
#9714 adds a time and scan count limit to ships with scanners so that they eventually quit scanning and leave the system.

PR #4823, opened 2020-02-17
Hai Security: varying planet security and Hai surveillance fleets

NomadicVolcano:
This is a revision to the PR #4649 which improved Hai security due to, in part, to their dislike of nerve gas. It does not make sense that all Hai planets would have equal security; slipping nerve gas into Cloudfire should be more difficult than a nearly uninhabited planet like Heartvalley. Also, the Hai seem wise enough (or cowardly enough) to scan for dangerous neurotoxins *before* they reach the planet surface.

Changes:

1. Hai planet security is about 0.1-0.3 on most planets, but a bit higher on Hai-home and planets with Humans. Cloudfire, a military base, has a security of 0.9

2. Waypoint and all Hai systems except Wah Ki have surveillance fleets.  You'll see one or two lightning bugs, or a squadron of unconstrained grasshoppers.

*Note:* This commit adds no surveillance ships in Wah Ki. The fights there are non-stop, so any surveillance ships would fight instead of scan.


NomadicVolcano:
The way I've coded this, you're scanned maybe once every three or four journeys through Hai space.  That is because only the surveillance fleets scan you and they're relatively infrequent. I'm not sure how high security should be turned up in Hai space.

In `data/map.txt`, The "Hai Surveillance (Slow)" fleets are the Lightning Bugs and the "Hai Surveillance (Unconstrained)" fleets are the grasshopper squadrons.  The grasshoppers are more likely to succeed in their scans since they come in groups of 5 to 7, and they are hard to evade. I made them less frequent.

NomadicVolcano:
Another odd thing I've seen is that the Lightning Bugs will leave a system immediately in a few seconds, sometimes before they even finish their scans. The Grasshoppers will stay forever.

NomadicVolcano:
If you leave the game running in one system for a few hours, you'll have a dozens and dozens of stationary Grasshoppers scanning you while spouting hippie love phrases.

Amazinite:
Ref: #4723

beccabunny:
Might be worth adding the outfit and cargo scanner to the hai outfitters

NomadicVolcano:
> Might be worth adding the outfit and cargo scanner to the hai outfitters

Thank you for catching that. I just added it to the intermediate and advanced Hai outfitters.

NomadicVolcano:
> It seems unnecessary to me to have two separate surveillance fleets. I'd suggest just combining both into one, or completely removing one or the other. (Perhaps only Grasshoppers do surveillance.)
> I'd say that 5 or 7 Grasshoppers is a bit much, but given that that's an uncommon fleet size (the most you'll see otherwise is 3), it'd make it obvious that Hai with scanners just showed up, so I think it's fine as a way to signal to the player that they're likely about to be scanned.

The Lightning Bugs are ineffective at scanning, but serve to warn the player that the Hai do scan. This is especially important for experienced players who do not expect the Hai to scan them. The Grasshoppers are the ones that will actually scan you; the bite that comes after the bark. 

Often the Grasshoppers won't try to scan anyone; they'll just sit around. Seven seems to be enough so at least two attempt to scan you. This is likely due to issue #4723 and is a necessary workaround. An alternative is for two or three to appear at a time, but appear far more often.

NomadicVolcano:
I have merged the fleets into one, reduced fleet sizes, made them more frequent, and added Shield Beetles. The beetles have seven each of outfit and cargo scanners, so they're quite skilled at scanning you. They also have a brig to put you in, should you misbehave.

NomadicVolcano:
I've also removed the "Small Hai" fleets from Waypoint, because there is no reason for Hai to fly there and back unless they're policing. Instead, there is a non-stop flow of surveillance ships.

NomadicVolcano:
> > It seems unnecessary to me to have two separate surveillance fleets. I'd suggest just combining both into one, or completely removing one or the other. (Perhaps only Grasshoppers do surveillance.)
> > I'd say that 5 or 7 Grasshoppers is a bit much, but given that that's an uncommon fleet size (the most you'll see otherwise is 3), it'd make it obvious that Hai with scanners just showed up, so I think it's fine as a way to signal to the player that they're likely about to be scanned.
> 
> The Lightning Bugs are ineffective at scanning, but serve to warn the player that the Hai do scan. This is especially important for experienced players who do not expect the Hai to scan them. The Grasshoppers are the ones that will actually scan you; the bite that comes after the bark.
> 
> Often the Grasshoppers won't try to scan anyone; they'll just sit around. Seven seems to be enough so at least two attempt to scan you. This is likely due to issue #4723 and is a necessary workaround. An alternative is for two or three to appear at a time, but appear far more often. That would negate the effectiveness of the Lightning Bugs as a warming that the Hai will scan.
> 
> An interesting alternative to the Lightning Bugs would be a Shield Beetle with a brig and extra Pulse Rifles.

I suppose a changelog entry is enough to warn experienced players.

PR #4827, opened 2020-02-18
Mystery cube jobs for stations

Anarchist2:
**Bugfix:**

## Fix Details
The wording used in the phrase "deep mystery cube dropoff payment", which is in the on complete dialog of Deep Mystery Cube jobs, assumes features specific to a planet. However, Deep Mystery Cube jobs can still have a destination of a space station, making the phrase seem out of place. This fix stops the main mystery cube jobs from choosing stations as a destination, and adds a new job and phrase specifically for Mystery Deliveries to stations. #4831 is a version of this pr without the replacement jobs.

## Testing Done
The job was tested for whether it could offer and whether the on complete dialog would be correct. Omnis was used for a faster set up, and the ship was chosen for no particular reason.

## Save File
This save file can be used to verify the bugfix. The bug will occur when using 732c68849fac62e7f13a6d71c5784552d32e05a9, and will not occur when using this branch's build.
[gg g~G.txt](https://github.com/endless-sky/endless-sky/files/4219251/gg.g.G.txt)
To see the job, simply go to the job board and accept the mission, then complete it.

### DIFF
--- a/data/deep jobs.txt
+++ b/data/deep jobs.txt
@@ -60,6 +60,7 @@ mission "Deep Mystery Cube [0]"
 	destination
 		distance 7 16
 		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Hai" "Quarg"
+		not attributes "station"
 	on accept
 		dialog phrase "deep mystery cube pickup"
 	on visit
@@ -90,6 +91,7 @@ mission "Deep Mystery Cube [1]"
 	destination
 		distance 9 20
 		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Hai" "Quarg"
+		not attributes "station"
 	on accept
 		dialog phrase "deep mystery cube pickup"
 	on visit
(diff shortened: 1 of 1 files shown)


Amazinite:
Given the stated purpose of the cubes in the Deep missions, I think it doesn't make too much sense to put them on stations, so perhaps just restrict the destination from being a station without adding a station variant of the job. 
The point of the cubes is that they're (somehow) highly advanced surveillance equipment that is meant to be hidden. A station is going to be a far harder place to hide a cube; compared to the kilometers away from the spaceport that one of the dialogs mentions while on a planet, there isn't exactly much space on a station.

Anarchist2:
Created #4831, which is a version of this pr without the replacement job.

Anarchist2:
Even though the alternate version of this pr was merged, I'm still keeping this up to get more opinions on whether having mystery cubes jobs to stations would be good.

Issue #4836, opened 2020-02-20
Friendly ships manually targeted with Heavy Rockets should become hostile


jostephd:
**Describe the bug**
Targeting friendly ships (not escorts) with Heavy Rockets doesn't make them hostile

**To Reproduce**
1. Take off
2. Left-click a friendly AI ship (not an escort)
3. Fire Heavy Rockets at it and hit it
4. Observe that it doesn't become hostile

**Expected behavior**
The ship should become hostile, same as if you shot it with any other weapon

**System (please complete the following information):**
 - Game Source: self-compiled
 - Version: current master

**Additional context**
I see https://github.com/endless-sky/endless-sky/issues/1341#issuecomment-234122992 but I think that issue is not the same as this one. #1341 and #821 are about splash damage. I'm not asking to change _that_, only to change what happens when I specifically target (left click) the friendly AI ship in order to be able to hit it in the first place.

Discussed on discord, @Zitchas said:
> So basically, have the AI do a check: If hit by AoE, check to see if targeted. If being targeted, treat as hit by normal directed fire weapon.

And by the way: Try doing this in a Vanguard with HRLs mounted in all gunports...

downRightCoder256:
> And by the way: Try doing this in a Vanguard with HRLs mounted in all gunports...

try doing this in a kestrel with extra gun ports (spoilers: 18 gun ports !). tears through quarg ships like a hungry child through pancakes

downRightCoder256:
i would suggest fixing this issue by adding something like "target" tag to munition? im not quite familiar with the core game code yet

McloughlinGuy:
Hitting ships with nukes doesn’t make them hostile, either.

Amazinite:
This occurs with any blast radius weapon.

Issue #1341, opened 2016-07-18
governments don't care about dying from splash damage


KiLEdEnNis:
so I was bored, played around with some Drak weaponry (like 2 antimatter cannons), and I found out something:
I fought for tribute from the Republic, and the splash damage killed merchants: they stayed friendly
later I fired at pirates, killed merchants and Oathkeepers doing so: they stayed friendly
I even deliberately killed some by shooting at asteroids next to them and they didn't turn hostile at all
there should be a mechanics change that makes a government hostile if killed by splash damage or after receiving significant damage (because an accident shot can happen)

jafdy:
This has been brought up before. The risk was that your "Ally" could turn hostile if you had too many heavy rockets. In this case realism was sacrificed for playability.

KiLEdEnNis:
well the reputation still lowers and can make friendlies hostile (tested w/ Hai, had 19.25 reputation, killed a water bug w/ an asteroid, they became hostile due to reputation loss)

toilethinges:
#821

endless-sky:
@KiLEdEnNis, is it possible you had the Water Bug selected as a target when you killed it? That would count as a hostile action even if it was killed with splash damage.

If allies couldn't take splash damage that would drastically alter the balance of certain weapons (like nukes). I am planning at some point to add "field" weapons, however, that damage all hostile ships within some range but not friendly ones. (For example, to allow the Drak ion field to properly damage all ships in range, not just one.)

KiLEdEnNis:
@endless-sky no, but killing ships w/o targetting does lower reputation, and mine with the Hai was maybe 20, killing a few through splash damage lowers reputation to under 0
like if you have -0,1 reputation w/ merchants that are in the sector, you can kill a pirate & they stop firing (had that too w/ Republic once, were some in sector, killed a pirate Bastion, they liked me again & stopped firing)

endless-sky:
Yes, you're right, killing people with splash damage affects your reputation, even though it doesn't "provoke" an angry response. But, in my opinion that's a pretty good balance - if you keep killing friendly ships as collateral damage, people *should* get mad at you.

KiLEdEnNis:
yeah, only weird things is that people die & they don't get mad, adding a "rescue" part to ships after they are destroyed (they don't disappear but they have holes & everything) could help a bit
so adding a limit of ships or crew, and when you destroy/kill more than that limit they get angry, could maybe help?

Alkallid:
[...]

Issue #821, opened 2016-02-27
Blast radius missiles used to kill without angering


toilethinges:
I don't know if it's really an issue or not (hehehe,) but:

If you use a missile with a blast radius (nuke, torpedo, heavy missile, etc) and shoot it into an asteroid near an enemy, it can damage them, even kill them, without angering anybody. It only angers enemies if you select them and they're caught in the blast, but if they aren't selected, they aren't mad that you just fired a nuke at them...

Gkramme:
Considering this could also unintentionally hurt your allies and turn them against you, I don't think that this is a bug and it shouldn't be changed.

jafdy:
although this is a bit of a problem I can't think of a way to change it with out friendly ships turning on you if they get clipped by the blast. But I'm no programmer so it may be doable.

masterpierround:
You could keep track of the damage you've dealt to a ship. If it's greater than a certain amount (a nuke, probably, or many, many torpedoes) and you manage to disable that ship, then other ships of that government get angry. This could cause problems with using nukes though, as a single nuke that clips an ally ship would turn them hostile.

endless-sky:
It's intentional, just because it's less problematic than having allies get mad at you right in the middle of a major fight just because one of them got caught in your heavy rocket's blast area.

PR #4842, opened 2020-02-23
fix: Ships hit by blast damage while targeted will be provoked

jostephd:
**Bugfix:** This PR addresses issue #4836 

## Fix Details
Ignore the `!isBlast` check when the ship being damaged is targeted

Should I uncomment the assert? I don't see any other asserts in the code (only `static_assert`) so I wasn't sure.

## Testing Done
I shot heavy rockets at an asteroid that a merchant ship was passing next to. The merchant exploded but didn't become hostile (as expected, see #1341 #821).

Then I selected a merchant, hit it directly with heavy rockets, and merchants became hostile.


jostephd:
Just realized that this doesn't fix the problem when the player left clicks a merchant and gives the `Fight my target` order to escorts armed with HRLs.

Issue #4829, opened 2020-02-18
Provide automated official nightly builds


MCOfficer:
**Is your feature request related to a problem? Please describe.**
During "the Hiatus" (0.9.8 - 0.9.9, see #3637) there was a growing need of nightly builds to test out the bleeding edge, especially from windows users, where compiling is rather difficult for the uninitiated. I somewhat filled that gap with unofficial nightlies, but that's not quite the same as official ones.

**Describe the solution you'd like**
With Github Actions being public, we can now run CIs & CDs without being dependant on a third party platform (other than github, which we depend on anyways). I would like to create a CD similar to [cd.yaml](https://github.com/endless-sky/endless-sky/blob/master/.github/workflows/cd.yaml), except that it runs on every new commit on `master`.

Those builds would then be released to a pre-release tag `continuous` (see for example [linuxdeploy](https://github.com/linuxdeploy/linuxdeploy/releases)).

**Describe alternatives you've considered**
Should this be rejected, i will continue to serve nightly builds myself - but it will never be as clean and efficient as with Github Actions, since i only have one platform at my disposal, whereas GitHub has all 3 supported ones.

**Additional context**
In case it isn't clear enough, i'd like to do this myself, since I'm already doing the CI changes and this is my best chance to contribute. Furthermore, i've already layed the foundation - a Linux CD has been merged this week and will run with 0.9.11, and some months ago i managed to build a [couple other nightlies](https://github.com/MCOfficer/endless-sky-actions/actions/runs/26740484) on an experimental fork.

Zitchas:
This would be awesome. Both from a usage point of view and reputational. MCO's nightly builds have been extremely useful, but are likewise slightly dubious for new users: Downloading unsigned executables from unknown people on the web is kind of discouraged for good reason. Having these served from the main endless-sky github location would add a layer of respectability to the process, which could help get more people onto the bleeding edge of development of material for the game.

tehhowch:
As long as the stable release remains visible/first, I'm fine with CD to the same tag on commits to master.

How will this work for forks? Can we make this action "opt-in"?

MCOfficer:
There may be a sofisticated way that prevents the action from running in the first place, the same way we only make it trigger on certain branches:
```
on:
  push:
    branches:
      - master
```

If that doesn't work, we just check the repo name in the first build step, and abort with a nice message if it doesn't match `endless-sky/endless-sky`.

MCOfficer:
[...]

PR #4861, opened 2020-02-29
feat: Add Continuous Builds

MCOfficer:
**Feature:** This PR **partially** implements the feature request detailed and discussed in issue #4829

## Feature Details
- runs on each commit on `master`
- Builds artifacts for all 3 platforms (64bit only)
- Artifacts can be downloaded like so (currently requires a GitHub account, see https://github.com/actions/upload-artifact/issues/51):
Peek 2020-02-29 15-12 GIF Gfycat

## Small changes & notes
- Version numbers are being adjusted on build-time, so Nightly builds always display the commit rather than the last version before them.
- Renamed `cd.yaml` to `cd_release.yaml`, as opposed to the new `cd_nightly.yaml`. Also renamed `build_parse.yaml` to `ci.yaml`.
- No AppImages are being built, i plan to add that once we have [ways to avoid duplicate code](https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336)


tehhowch:
oof: 
image

You could use the 7-char short hash instead of the full SHA to avoid the overflow.

This should be a bit more informative about what it is:
image

What do you think about changing it to something like
"Commit {{SHA}}: {{commit title}}"?
You could probably just prefix the output of `git log --oneline -n 1` with "Commit", or use your own format string for git log.

Because the downloaded artifacts aren't git repos, it could also be nice to add commits since the most recent actual release to the changelog file, so those who obtain the zip don't need to try to convert a git SHA into a commit and then figure out how to look at the commit history for it.

(It would be nice if the downloads could be single files instead of zipped zips, but it sounds like that change is on the horizon and I know you'll update this workflow when it's possible.)

MCOfficer:
> oof:
> image
> 
> You could use the 7-char short hash instead of the full SHA to avoid the overflow.
> 
> This should be a bit more informative about what it is:
> image

Seems like the regex was a bit *too* greedy there. It should read `Endless Sky [sha]`

> What do you think about changing it to something like
> "Commit {{SHA}}: {{commit title}}"?
> You could probably just prefix the output of `git log --oneline -n 1` with "Commit", or use your own format string for git log.
> 

I'll look into it, there are certainly ways to make the credits file more verbose.


> Because the downloaded artifacts aren't git repos, it could also be nice to add commits since the most recent actual release to the changelog file, so those who obtain the zip don't need to try to convert a git SHA into a commit and then figure out how to look at the commit history for it.

That's a potentially a tough one. I'll look into it eventually, but since this is under construction anyways i wouldn't put it into the scope of this PR.

MCOfficer:
Point 1 & 2 are done. 3 Will be postponed for a later PR.
image

.github/workflows/cd.yaml
```diff
@@ -1,43 +1,99 @@
-name: CD
+name: Nightly CD
```
tehhowch: This should probably be "Latest Commit" or similar, and not "nightly" since its continuous

.github/workflows/cd.yaml
```diff
@@ -1,43 +1,99 @@
...
 
 jobs:
-  appimage_amd64:
+  ubuntu_x86_64:
     runs-on: ubuntu-latest
     env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      ARCH: x86_64
-      OUTPUT: endless-sky-x86_64.AppImage
+      OUTPUT: endless-sky-amd64.tar.gz
```
tehhowch: Is it intended that the downloaded filename never changes?

MCOfficer: It's a relic from when i tried to make this pipeline run on release and commit. I'll put the commit message in there (and give the AppImage a version while i'm at it).

MCOfficer:
image
The credits.txt formatting works in mysterious ways... Should i drop the indentation for broken lines?

Amazinite:
Nice.

PR #4862, opened 2020-02-29
Fix: Normalizing remnant salvage available capitalization

Zitchas:
This fixes the "remnant salvage available" event, which currently has capitals and the standard for ES is for events to be entirely lowercase.

There are no missions that currently depend on this event, but probably will in the future.


Amazinite:
Just know that in the future if you do use this event as a trigger, you need to check for both the capitalized and the all lowercase versions for backwards compatibility (unless we want to require a new pilot).

Zitchas:
Yes, I'll do that.

PR #4884, opened 2020-03-05
Removed bad advice

Naminson:
One of the hails you can get from merchants says that one way to deal with pirates is to "get you rich you can buy them off". However, the way bribes work means that trying to do this is highly impractical. This PR removes that line.


Amazinite:
> the way bribes work means that trying to do this is highly impractical. 

Bribes are a factor of your fleet's hull cost. 

https://github.com/endless-sky/endless-sky/blob/94346d8905ec516b9a5fad2683f9ca0e5ab5f11f/source/HailPanel.cpp#L61-L71

https://github.com/endless-sky/endless-sky/blob/94346d8905ec516b9a5fad2683f9ca0e5ab5f11f/source/HailPanel.cpp#L312-L322

Getting rich will still help you pay off pirates.

tehhowch:
Bribing should always count your full fleet net worth. If you're worth that much, you're probably well known. If you're well known, people know you own those things, even if they aren't flying with you at the moment. All the more reason, then, to ask for more money before relenting...

And I disagree with removing flavor text that also indicates what bribing does.

Zitchas:
I'm kind of two minds on that.

On the one hand, if you're rich enough to have dozens of ships parked around, people are going to know about it, and thus the higher bribe cost. On the other hand, the fact that I have a dozen Albatrosses parked in Aventine that have never left the Ember Wastes should have absolutely no bearing on what a pirate is demanding for a bribe. 

I guess what I'm getting at is that if the ships are in human space, they should be fair game for pirates to use in determining how much they charge in bribes. But ships that are outside of human space, shouldn't. (Possible inclusion of Hai space)

Amazinite:
> Now, if this was James telling you in mission dialog that building up a huge networth was the best way to deal with pirates... Then there'd be a problem.

Well the thing is, the calculation *is* based off of your fleet's hull worth, not your net worth, so saying to get rich regardless isn't wrong. And on top of that, it's the square root of your fleet's hull worth that it used, meaning that the you're actually paying a smaller percentage of your fleet's worth in the bribe the bigger your fleet gets. So if it was the case that you paid based off of your net worth, it would still stand that you'd want to be richer given that square root scaling of the bribe amount.

Amazinite:
Final verdict: the advice isn't actually bad, but we do need to make it clear how bribe work. :)

Issue #4885, opened 2020-03-05
underused solar panels


downRightCoder256:
**Is your feature request related to a problem? Please describe.**
at current point in game, there are two human outfits: KP-6 Photovoltaic panel and KP-6 Photovoltaic array, that seem to be more or less obsolete. Main issue for that would be terrible energy generation per ton and energy generation per credit features
**Related Issue Links**
     #4065
     #4034

**Describe the solution you'd like**
-make solar panels a bit better:
     make solar panels produce around 50% more energy. so far probably the simplest solution
-have more non combat ships equipped with it by default
     have some ships that aren't designed for direct combat utilize such panels.good examples
     i would like to point out are star barge and space shuttle: the only real big energy consumer
     on 'barge is AM turret. having it equipped with 2 solar panels right off the bat would
     greatly reduce its purchase cost, and make for a more energy managing experience
-make them more available
     knowing that they have been around for a while, and that their original company is now
     defunct, it would be expected to find such an outfit on almost any station with an outfitter,
     and since it can be easily manufactured out of cheap components (some plastic, silicium,
     and some conductor). it would be expected for such a piece of electronics to be dirt cheap;
     around 2000 credits on average and probably no more that 5000 credits in worst case
     scenario
-make "subships"(drones & fighters) able to recharge
     as of 0.9.11, fighters and drones can be repaired by its mothership, but their batteries cannot
     be recharged. Adding such an ability would greatly increase the demand for solar panels,
     since they are cheaper than fuel cells (and could be cheaper if points above are considered),
     and most fighters and drones aren't designed to last long enough in combat to even deplete
     a LP036a battery, the smallest of them all, such cheap power source would probably fix the
     issue of demand for "subships".

edit: since I've only now noticed it may not make sense at first; each of these points is its own proposal, not a single unified solution.

downRightCoder256:
my proposal of the solution on #4034 problem is to make fighters either dirt cheap (<200,000 credits per unit), or more combat effective. last point tackles the former

Zitchas:
My PR doesn't affect the usage of solar panels, but it does give them a bit of a boost in that they now act much more like their description states: Things that use only minimal space in the ship and are bolted on the outside - but still provide a fairly hefty mass, thus impacting mobility. 

It is a rather minor change, but it could be enough to make them more desirable for at least poor merchants that are trying to maximize cargo space and don't need to power weaponry.

downRightCoder256:
[...]

PR #4886, opened 2020-03-05
Reducing the Outfit Size of Solar

Zitchas:
**Balance Fix:** This PR addresses issue #4885 

## Fix Details
- Size of KP-6 panel reduce to 1 ton (from 3), but mass is left at 3.
- Size of KP 6 array reduced to 8 ton (from 24), but mass is left at 24.
- Size of small collector module reduced to 4 (from 8), but mass is left at 8.
- Size of large collector module reduced to 14 (from 28), but mass is left at 28.

The description and functionality of solar panels is specifically something that can be bolted to the outside of a ship. As such, the amount of space it takes up in our outfitter should be minimal. However, just because it doesn't take up space internally, doesn't mean that it doesn't add mass to the ship. 

This presents an interesting trade off: 
- For people who value energy per unit of outfit space (for instance, merchants who don't care about speed and just want the most space possible), then these outfits will be a little bit more useful. 
- For those who care about energy per unit of mass added to their ship (people who care about speed and mobility, or have high power demands), then these outfits will be just useless as they always have been. 

This, I feel, elevates solar panels from the realm of "completely useless except for maxed out cargo/bunk builds" to "a poor merchant might actually consider using these." Which I think makes them a more interesting part of the game.

That being said, I would like to hear from those who are focusing on Coalition space, like @Arachi-Lover . How does this change affect the balance over there? Is it too much? Not enough?

Also, I am aware of MZ's general dislike of 1 ton outfits, because "no one would have any reason to *not* put one on their ship to fill the last bit of space." Firstly, having a tiny backup solar panel as an extremely common outfit on a spaceship makes a huge amount of sense, so it actually sounds like a really good lore thing. Secondly, it's 6 energy. That's tiny. That's 1/8th the power output of the next smallest power generator, and an even tinier fraction of the power output of anything used for combat purposes. And thirdly, it is solar. Which means that it has to be near a star to give full power. So even as a "thing to fill that last ton of space" sort of outfit, it's a rather pathetic value. I suspect that for most people, the extra 3 tons of mass is a bigger detriment than the small energy gain.

### DIFF
--- a/data/coalition outfits.txt
+++ b/data/coalition outfits.txt
@@ -15,7 +15,7 @@ outfit "Small Collector Module"
 	cost 60000
 	thumbnail "outfit/small collector module"
 	"mass" 8
-	"outfit space" -8
+	"outfit space" -4
 	"solar collection" .8
 	"heat generation" .4
 	description "Solar power is the mainstay of the Coalition's merchant marine. The design of these panels has been refined over the millennia, growing ever more efficient."
@@ -27,7 +27,7 @@ outfit "Large Collector Module"
 	cost 300000
 	thumbnail "outfit/large collector module"
 	"mass" 28
-	"outfit space" -28
+	"outfit space" -14
 	"solar collection" 3.3
 	"heat generation" 1.5
 	description "One of these solar panels is enough to meet the energy needs of a medium-sized Coalition civilian ship, especially given the energy efficiency of their engines and the fact that under their laws only Heliarch ships are allowed to have weapons."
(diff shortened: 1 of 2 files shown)


Arachi-Lover:
I haven't meddled much with the balance of the civilian Coalition stuff, such as their solar collection modules, so I don't know enough about that to speak about it (though I'll say I doubt it would ever pose a serious balancing issue), but I'm all for this buff to Solar in general, makes sense to me.

PR #2618, opened 2017-06-11
Scaling heat generation for solar collectors

TheUnfetteredOne:
Solar collectors generate heat in proportion to their solar collection.
Coalition collector modules are tweaked to keep their maximum heat
generation the same, but they'll generate less heat away from star.
<hr/>
Literally all done by @Elyssaen but he's letting me PR it for some reason.

Issue #2508, opened 2017-05-10
Does the heat generation of solar panels scale with how much energy they're providing?


TheUnfetteredOne:
That is, will a solar panel produce more heat closer to the system center than when farther away?
If this isn't the case right now, should it be?

Elyssaen:
Doesn't look like it. They just define `heat generation` like any other outfit, so it's static.

Maybe it should work as you say. Would have to define a new attribute, `solar heat generation`, that was added to your ship's heat in the same place, but scaled by the `scale` variable that's used to scale `solar collection`.

If there's some consensus that it should work that way, I'll code and PR it.

tehhowch:
It is kinda silly that we are told a planet is too hot to land on, but not too hot to orbit indefinitely :)

:+1: for heat generation from based on your distance to the system star. 
:-1: for possibly needing to rebalance all fleets because of extra "heat damage" to all ships involved.

`solar heat generation` should consider the star sprite, btw - those old red stars aren't nearly as hot.
One possibility would be to scale based on the system star and then on your ships mass - as in general most ships are made of similar materials with similar densities and the densities of typical things such as weapons, bunks, reactors, etc does not vary from ship to ship, so the mass of your ship scales with its volume, and the solar fluence scales with its surface area.

so something like
`solar heat gen = star_energy_output (J/s) * distance scaling (1/m^2) * exposed_surface_area (m^2)`
If we assume ships to be cylinders with aspect ratio H = 4R, then R is determined from the ship's average density and its baseAttributes.Mass value (since storing outfits/cargo inside doesn't alter the volume):
`R= ( (Mass/avg_density )/(4 * pi) )^1/3`
which lets you determine the exposed surface area (assuming 1/2 is shaded from the star):
` esa = 0.5 * 10 * pi * R^2 ` -> ` esa ~ 2.91 * (Mass/ avg_dens)^(2/3)`
Given the average density of most metals is 2k-10k  kg/m^3  and the spaceship needs at least some non-metal sections, you could probably peg the average ship density to somewhere in the range of 1-4k  kg/m^3, and introduce a density scaling factor as a function of total outfit space, ship type, etc. 
This leaves you with solar heat generation dropping off with 1/r2 (or the ramscoop efficiency profile, probably), and increasing with Mass^(2/3). 

[...]

PR #4894, opened 2020-03-09
feat: Solar heat mechanic for increased heat generation the closer you are to the system center

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #2508. Also, turns out #2618 existed, so this is really just that PR redone.

## Feature Details
Adds an attribute `"solar heat"` that would allow an outfit/ship to produce heat given its distance to the system center. Meant as a counter to solar collection's energy production given the distance to the center.
This PR does not add this attribute to any outfit, only implementing the mechanic to be used.

I'll note that I didn't use the same `"solar heat generation"` name as #2618; solar collection isn't `"solar energy generation"`, after all. 

## Usage Examples
```
outfit "Small Collector Module"
	...
	"solar collection" .8
	"solar heat" .4
```

## Testing Done
It builds and the CI tests pass. That's about all the testing I've done.

## Performance Impact
N/A

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1674,9 +1674,10 @@ void Ship::DoGeneration()
 		{
 			double scale = .2 + 1.8 / (.001 * position.Length() + 1);
 			fuel += currentSystem->SolarWind() * .03 * scale * (sqrt(attributes.Get("ramscoop")) + .05 * scale);
-		
-			energy += currentSystem->SolarPower() * scale * attributes.Get("solar collection");
-			heat += currentSystem->SolarPower() * scale * attributes.Get("solar heat");
+			
+			double solarScaling = currentSystem->SolarPower() * scale;
+			energy += solarScaling * attributes.Get("solar collection");
+			heat += solarScaling * attributes.Get("solar heat");
 		}
 		
 		double coolingEfficiency = CoolingEfficiency();
@@ -2505,8 +2506,7 @@ double Ship::IdleHeat() const
 	// heat = heat * diss + heatGen - cool - activeCool * heat / (100 * mass)
 	// heat = heat * (diss - activeCool / (100 * mass)) + (heatGen - cool)
 	// heat * (1 - diss + activeCool / (100 * mass)) = (heatGen - cool)
-	double production = max(0., attributes.Get("heat generation") 
-		+ attributes.Get("solar heat") - cooling);
+	double production = max(0., attributes.Get("heat generation") - cooling);
 	double dissipation = HeatDissipation() + activeCooling / MaximumHeat();
 	return production / dissipation;
 }
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1676,6 +1676,7 @@ void Ship::DoGeneration()
 			fuel += currentSystem->SolarWind() * .03 * scale * (sqrt(attributes.Get("ramscoop")) + .05 * scale);
 		
 			energy += currentSystem->SolarPower() * scale * attributes.Get("solar collection");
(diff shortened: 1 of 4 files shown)


source/Ship.cpp
```diff
@@ -2504,7 +2505,8 @@ double Ship::IdleHeat() const
 	// heat = heat * diss + heatGen - cool - activeCool * heat / (100 * mass)
 	// heat = heat * (diss - activeCool / (100 * mass)) + (heatGen - cool)
 	// heat * (1 - diss + activeCool / (100 * mass)) = (heatGen - cool)
-	double production = max(0., attributes.Get("heat generation") - cooling);
+	double production = max(0., attributes.Get("heat generation") 
+		+ attributes.Get("solar heat") - cooling);
```
Amazinite: Looks like fuel heat isn't added here despite the fact that it is factored into the heat table in ShipInfoDisplay. Add it here as well, or remove solar heat?

downRightCoder256:
maybe have something like “working heat” instead, which the game would interpret as a factor of energy generation used to calculate heat generation? for eg. a nGF-AA fuel cell (energ. 72,heat 114) would have a “working heat”= 1.583

Amazinite:
Not sure how that has any relation to this PR or what that would add.

source/Ship.cpp
```diff
@@ -1676,6 +1676,7 @@ void Ship::DoGeneration()
 			fuel += currentSystem->SolarWind() * .03 * scale * (sqrt(attributes.Get("ramscoop")) + .05 * scale);
 		
 			energy += currentSystem->SolarPower() * scale * attributes.Get("solar collection");
+			heat += currentSystem->SolarPower() * scale * attributes.Get("solar heat");
```
tehhowch: Compute the scaled solar power once.

source/Ship.cpp
tehhowch: I don't know that we want to modify IdleHeat like this, since this is ignoring all components that otherwise influence the per frame added heat (system solar power and range from center). Probably it is OK to add the maximal solar heat value, rather than actually compute the scaled power factor.

As far as adding "fuel heat", that should require a conditional check since fuel heat is only added to the ship's heat when the ship also has both "fuel consumption" and enough fuel to be consumed that frame. The computation is already stateful--a ship can report a different IdleHeat at different game frames due to changes in mass--so updating it with dependencies on fuel, fuel consumption, and the energy needed for active cooling is fine. The function is only used when a ship is instantiated in-flight, lands at a spaceport, or when performing an overheating flightcheck, so this calculation can be a little more correct without negatively affecting performance.

Comparing the values of IdleHeat before and after adding in those values, for ships with active cooling, could be interesting from an academic standpoint, along with comparing the value of `heat - IdleHeat()` for one of these ships that launches (and does nothing).

downRightCoder256:
by adding “solar heat” you are tying heat output to the proximity to the sun. by using my proposed solution, you are tying heat output to the energy output itself, which means that you can use it in the same way as “solar heat” awhile giving increased functionality for mod makers and existing outfits. i can see this particularly useful when regulating ion damage

Zitchas:
Hmmm, actually yes, I can see what you are getting at. Solar heat could be done that way, with some extra flexibility for other potential energy generators that provide heat per unit of energy instead of heat at a fixed rate.

That being said, it would probably be better if solar heat functions as described by Amazinite (this could also be used to implement other mechanics, such as ships heating up when close to the sun, which your suggestion can't do) and your "heat per energy unit" is implemented as a separate attribute. If you haven't already, I suggest filing an issue or request for it.

source/Ship.cpp
Amazinite: I'll just remove this change.

PR #4898, opened 2020-03-09
Added a necessary requirement to a Ka'het mission

beccabunny:
## Summary
This PR adds the requirement for the Remnant first contact to be completed, this way it won't be offered when the player met the Ka'het but is landing on a Remnant planet for the first time


data/ka'het/ka'het missions.txt
```diff
@@ -41,6 +41,7 @@ mission "First Contact: Ka'het: Remnant 1A"
 		government "Remnant"
 	to offer
 		has "Graveyard Label: offered"
+		has "First Contact: Remnant: declined"
```
Amazinite: What you should actually do is actually have the correct condition for the map label. :)
It should be `has "event: ember waste label"`. Using `has "remnant blood test pure"` would work as well and be more "correct" in terms of the intention of that condition.

This change also needs made for the other Remnant mission here.

beccabunny: ..but then, how do you know that the player visited the graveyard?

beccabunny: Done

PR #4906, opened 2020-03-10
Reduce kestrel's gun ports from 18 to 12

Naminson:
Currently, the weapons Kestrel better than every other variant of the Kestrel, if not every ship in human space. The reason for this is that it has 18 gun ports, and can dedicate over 15 tons to each. This means that it can use 18 heavy lasers (and power them, as long as you use a few mass expansions) and 18 sidewinders, allowing it to easily take out any human ship without much effort. (It also means it can use a lot of heavy rocket launchers, although they do tend to cause self-damage.)

This PR removes six of the Kestrel's gun ports. Ideally, this will mean that the ship has less versatility and a much lower potential DPS, while still keeping the weapons Kestrel novel.


Zitchas:
This is a "meh" from me. I don't really care either way. 18 gun seems silly and excessive, but it is different, and I like having more different stuff. I usually go with either the shield or the engine variants, depending on my mood.

downRightCoder256:
well i mean, 12 is still more than any other ship in game afaik

starscollide5:
I think it should be limited to 8.

PR #4895, opened 2020-03-09
Let there be subfolders

Amazinite:
## Summary
#2579 take 2.
Ref: #2050 #2125

* Creates a data subfolder for the following:
  * Coalition
  * Drak (contains Indigenous)
  * Hai
  * Human
  * Ka'het
  * Korath
  * Pug
  * Quarg
  * Remnant
  * Wanderer
* Files that don't fit in a subfolder have simply remained in data.
* This PR does not rename or reorganize anything else as to avoid creating headaches. 
  * Any other organization like splitting up the Pug file into outfits, ships, etc. can come after this.

## Screenshot
image

Issue #1519, opened 2016-09-02
Choice start position


Amazinite:
The ability to choose your starting planet by being given a choice at the beginning. 
So be given a choice like:

```
Where do you wish your starting location to be?
    Dirt Belt
    Paradise Worlds
    The Deep
    Core
    Pirate World
```

or even:

```
What are you?
    Human
    Hai
    Korath
```

and each choice would direct you to a different start with different start conditions and start conversations.

```
choice start
    `What are you?`
    choice
        `   Human`
            goto human

    label human
    `Where do you want to start?`
    choice
        `   The Deep`
            start deep

start deep
    date 16 11 3013
    system "Naos"
    planet "Asgard"
    set "Deep Born"
    ...

conversation deep
    `You step off the front steps of the university in your city, a new graduate ready to leave your mark on the galaxy...`
    ...
```

Would definitely help for mods that modify the start, not having to remove those plugins if you wish to have a normal start, as well as opening up new possibilities.

KiLEdEnNis:
not only would it help mods, something like this would eventually raise ES to a whole new level, there isn't a huge team working on it but being a different species or starting somewhere else with other missions will basically make it a triple A game in my opinion

endless-sky:
If the player decided to start as a different species, pretty much every mission in the game would need to be rewritten, So, at that point you're probably talking about a total conversion mod.

As for starting on different planets, a more general solution might just be to allow mission actions to "warp" the player to an arbitrary other location. That's something I'd already considered for other reasons (adding missions where it turns out the Quarg have some sort of instantaneous travel system between their ringworlds).

Bladewood:
wait, what was that last part?
they have a super jump drive built into their ringworlds?

Amazinite:
> If the player decided to start as a different species, pretty much every mission in the game would need to be rewritten, So, at that point you're probably talking about a total conversion mod.

That was the point of the little `set "Deep Born"` line in the start section; you could have certain missions that only offer if you chose a certain start position, or are restricted due to your starting position. So you wouldn't necessarily need to rewrite all the missions to accommodate for being a different species, you could just have them all require you to have a `"Human"` condition.

endless-sky:
[...]

PR #4910, opened 2020-03-11
Move human related data into human sub-folder

Darcman99:
Relocate two data files into human sub-folder because all the data in the files are relevant to human content only. There are no non-human starts in the game yet and the current intro missions don't even scratch aliens, so it would be assumed that they belong to the human sub-folder.

One could easily argue that the start/intro of the game isn't strictly human and could be alien if that direction was taken as alternate starts to the game, however as of right now it remains only relevant to humans.

Obviously this is related to #4895 where sub-folders were created in the first place.


Amazinite:
I went back and forth on whether to put these files in the human folder. I think there's a good argument for putting the intro mission in the human folder, as they do relate entirely to human space (human characters, human sources and destinations, etc.), but I think the start file should stay outside of the subfolders. 

Whereas the intro missions file may expand in the future but always contain intro missions pertaining to humans, the start file may expand in the future to include starts that aren't for humans. (Ref: #1519) The start file is also not really "human" outside of the fact that the only start right now is for humans. It feels like a more general file to me akin to governments (and given #1519, that's exactly what it'd become, as mentioned).

Darcman99:
Then could `start.txt` not simply be moved back to the general area in the same eventual commit that adds alternative starts? It seems pointless to prepare for a feature that we don't even have an estimate for when it might be created. (this is purely from what I know, as far as I know there's no estimate yet)

MCOfficer:
Judging by topics: both files are about the player, not about humans.

Amazinite:
> Then could `start.txt` not simply be moved back to the general area in the same eventual commit that adds alternative starts?

In the same vain, if it’s going to be moved back, why move it in the first place?

Darcman99:
Going off MCOs comment, I understood that the text files were ordered by what they contained. If they are ordered by their topics and what they are meant for then I agree on how it was.

Issue #4912, opened 2020-03-11
Depreciated outfit cost rounding error.


ErdrickXLII:
**Describe the bug**
Fully depreciated outfits cost more to buy back than what they sold for.

**To Reproduce**
1. Obtain a fully depreciated outfit (either by capture or the outfits given to you by the story.)
2. Sell it to an outfitter (e.g. the cloaking device sells for 300 000.)
3. The price to buy it back has increased slightly (e.g. that cloaking device now costs 300 318.)

**Expected behavior**
Outfits should be able to be bought back for the price you just sold them for...right?

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Github download page
 - Version: 0.9.11

**Additional context**
Not sure if it happens with outfits that have been allowed to depreciate naturally.

Amazinite:
I'm gonna take a wild guess and say that #4372 is to blame. That final 318 credits is probably the last 7 days of depreciation. The age of plundered outfits is probably 1000 when it should be 1007 as to account for the grace period.

PR #4914, opened 2020-03-12
fix: Account for the depreciation grace period in the max age of an outfit or ship

Amazinite:
**Bugfix:** This PR addresses issue #4912.

## Fix Details
MAX_AGE is now 1000 + GRACE_PERIOD (i.e. 1007). 
This makes it so that Depreciation::Save and Depreciation::Buy will use 1007 as the age for fully depreciated items instead of 1000 as is used currently, which meant that you'd actually gain back 7 days of depreciation (as seen in #4912).
Simplified the linear equation in Depreciation::Depreciate as a result of this change. The outcome of the Depreciate function has remained unchanged, though.

### DIFF
--- a/source/Depreciation.cpp
+++ b/source/Depreciation.cpp
@@ -29,8 +29,8 @@ namespace {
 	// Depreciation parameters.
 	double FULL_DEPRECIATION = 0.25;
 	double DAILY_DEPRECIATION = 0.997;
-	int MAX_AGE = 1000;
 	int GRACE_PERIOD = 7;
+	int MAX_AGE = 1000 + GRACE_PERIOD;
 }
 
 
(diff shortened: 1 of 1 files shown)


source/Depreciation.cpp
```diff
@@ -29,8 +29,8 @@ namespace {
 	// Depreciation parameters.
 	double FULL_DEPRECIATION = 0.25;
 	double DAILY_DEPRECIATION = 0.997;
-	int MAX_AGE = 1000;
 	int GRACE_PERIOD = 7;
+	int MAX_AGE = 1000 + GRACE_PERIOD;
```
tehhowch: These depreciation parameters should be made `constexpr ___`.

source/Depreciation.cpp
```diff
@@ -334,11 +334,11 @@ double Depreciation::Depreciate(int age) const
 {
 	if(age <= GRACE_PERIOD)
 		return 1.;
-	if(age >= MAX_AGE + GRACE_PERIOD)
```
tehhowch: Was there a reason we didn't reduce `age` by `GRACE_PERIOD`, check if it was <= 0, and exit if so? Could we have then used the original calculations?

Amazinite: I don't believe so, as the true age of fully depreciated outfits and ships would still be set to 1000, meaning that the depreciation calculation would use 1000 - 7 = 993 days, which isn't max depreciation.

PR #4915, opened 2020-03-12
Add flagship System and Planet to player conditions

petervdmeer:
**Feature:** This PR implements a feature to publish the flagships location (system and planet) in the player conditions.

## Feature Details
This adds two new conditions:
```
"flagship planet: <planet-name>"
"flagship system: <system-name>" 
```
Those conditions can be used during automatic tests to put asserts on the location of the flagship.

## UI Screenshots
N/A

## Usage Examples
To be used during automatic testing as a condition to wait for or to assert on.

## Testing Done
Verified that the conditions get set (on Shangri-La in the Polaris system) and that the conditions get updated/rewritten (on Amazon in the Schedar system).

## Performance Impact
No significant impact expected, those conditions are only calculated on major events like departing or landing.

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2332,12 +2332,26 @@ void PlayerInfo::UpdateAutoConditions(bool isBoarding)
 		conditions["passenger space"] = flagship->Cargo().BunksFree();
 	}
 	
+	// Clear any existing flagship system: and planet: conditions. (Note: '!' = ' ' + 1.)
+	auto firstS = conditions.lower_bound("flagship system: ");
+	auto lastS = conditions.lower_bound("flagship system:!");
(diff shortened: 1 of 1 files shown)


source/PlayerInfo.cpp
```diff
@@ -2332,12 +2332,26 @@ void PlayerInfo::UpdateAutoConditions(bool isBoarding)
 		conditions["passenger space"] = flagship->Cargo().BunksFree();
 	}
 	
+	// Clear any existing flagship system: and planet: conditions. (Note: '!' = ' ' + 1.)
+	auto firstS = conditions.lower_bound("flagship system: ");
```
tehhowch: We do a similar thing above - write a lambda to do this and pass the condition prefix to it in each case.

PR #4918, opened 2020-03-13
feat: CD: Also build AppImage

MCOfficer:
**Feature:** This PR adds an AppImage to the CD.

## Feature Details
Thus far AppImages were only build for releases. Since tarballs can link against libraries that the target system doesn't have, AppImages are a better choice for distributing portable linux packages.
This PR extracts the build steps into a script `build_appimage.sh`, which both the CD and Release CD use. It also fixes an issue with the Release CD's version number.

Please advise on where to put `build_appimage.sh`. Do we want a folder specifically for CI/CD scripts, similar to `test`?

## Testing Done
https://github.com/MCOfficer/endless-sky/releases/tag/appimage-test-2
https://github.com/MCOfficer/endless-sky/actions/runs/55112999

### DIFF
--- a/build_appimage.sh
+++ b/build_appimage.sh
@@ -1,0 +1,19 @@
+# Builds Endless Sky and packages it as AppImage
+# Control the output filename with the OUTPUT environment variable
+
+# We need an icon file with a name matching the executable
+cp icons/icon_512x512.png endless-sky.png
+
+# Build
+scons -j $(nproc) install DESTDIR=AppDir 
+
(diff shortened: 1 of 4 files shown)


build_appimage.sh
```diff
@@ -0,0 +1,19 @@
+# Builds Endless Sky and packages it as AppImage
```
tehhowch: No shebang?

MCOfficer: oops.

PR #4920, opened 2020-03-14
Remnant: Adding the Inhibitor Turret

Zitchas:
**Content (Outfit)**

## Summary
This PR adds the Inhibitor turret. 
- Stats are equivalent to two Inhibitor canons
- Noteworthy difference: The turret takes up less space than two inhibitor canons, but it still adds as much mass to the ship as two canons. The description points out that this is a recent development in which the Remnant have figured out how to put a significant amount of the weaponry on the turret itself (which is projecting out of the ship into space) and thus does not occupy as much space within the ship. 

The amount of outfit space it takes up is still nowhere close to the size I *think* should be be on the outside, but this is a break from the "traditional" method, so I opted to take a gentler approach to see how it works and how it affects game balance. (Note: I'm not advocating that all turrets do this. In fact, this description introduces a lore 'suggestion' explaining why most turrets don't take up less space: Namely, that they have so much machinery within the ship that just isn't visible. Thus this Remnant turret is actually special/unique in that it manages to extract that machinery and put it up top instead.)

## PR Checklist
 - [X] I uploaded the necessary image, blend, and texture assets here: [link to zip file on google drive](https://drive.google.com/open?id=1uPOIolf-nonc_amMMeNtXFn67bfLiwSG)
 - [X] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the [`@2x` versions of these art assets](https://github.com/endless-sky/endless-sky-high-dpi/pull/118)


downRightCoder256:
since it says all the machinery is ON the turret, why not make it look more like so ?

Zitchas:
Well, I'm looking at as an aspect of miniaturization. Whereas the older tech (like the gun proper) is basically just the nozzle for a bunch of machinery inside the ship, this newer tech has sufficiently miniaturized and/or re-arrange the components such that much more of it fits inside the guns and the turret mount (for instance, look at the extended area between the two guns. That's added space not present on the original guns). So, aside from that area, the majority of the improvement is within the existing shell or contained within the turret mount. 

Said improvements probably haven't been retroactively moved back to the base guns due to expense and lack of available resources. That being said, like most recent technologies, this improvement will probably migrate back down the supply chain and extend to other things over time. (or missions)

downRightCoder256:
hmm, I see. Tough I still find it unlikely for remnant, who up until now have only really been reverse engineering, to have made such an major breakthrough in turret technology. maybe add a higher crew competent to make up ?

data/remnant/remnant outfits.txt
```diff
@@ -51,6 +51,42 @@ effect "inhibitor impact"
...
+outfit "Inhibitor Turret"
+	category "Turrets"
+	licenses
+		Remnant
+	cost 1371000
+	thumbnail "outfit/inhibitor turret"
+	"mass" 32
+	"outfit space" -25
+	"weapon capacity" -25
+	"required crew" 1
```
downRightCoder256:
```suggestion
	"required crew" 3
```

Zitchas: No, I do not think that a higher crew count is appropriate here. Higher price is definitely worth considering, but figuring out a way to miniaturize stuff so that more of it is in the external turret isn't something that would lend itself to having triple the crew count to utilize it.

Zitchas:
The Remnant do a lot more than "just" reverse engineer things. Virtually all their resources for the past five centuries has been poured into military and survival pursuits. And people in those sorts of situations tend to learn very quickly to simplify and reduce the amount of resources needed. 

Other balancing mechanisms (such as cost) are a possibility, although increased crew count is not one of them. Remnant do not have access to a large recruit-able population like most other factions do, so they are unlikely to create weapons or ships that require putting an unusually high amount of their population in the line of fire.

It is also worth noting that inhibitors are not particularly strong weapons; this turret is not going to be a game-changer for most players (and it will be laughable for players who have already reached the Wanderers by the time they discover the Remnant)

downRightCoder256:
>Higher price is definitely worth considering

indeed: but maybe we could increase its mass slightly too? Currently, its mass is equivalent to two to the inhibitor cannons, although that is not the case for any other ship. The mounting hinge would still weight quite a bit, and even if the machinery has been placed in the turrets itself, as you said, it would still add to its gravimetric density.

> It is also worth noting that inhibitors are not particularly strong weapons; this turret is not going to be a game-changer for most players (and it will be laughable for players who have already reached the Wanderers by the time they discover the Remnant)

I highly doubt that; although the Sunbeam packs better shield damage per ton, its hull damage is surprisingly low, it has shorter range and, in my opinion most importantly, it draws so much more energy, making cross-species builds unviable. Inhibitor cannon, on the other hand, takes up much less space, has longer range, features push-back force, preventing enemies from just jumping away and, most notably, packs slowing damage, which is much more noticeable than the Sunbeam's heat damage.

data/remnant/remnant outfits.txt
```diff
@@ -51,6 +51,42 @@ effect "inhibitor impact"
 	"random velocity" 6
 	"velocity scale" -.05
 
+outfit "Inhibitor Turret"
+	category "Turrets"
+	licenses
+		Remnant
+	cost 1371000
+	thumbnail "outfit/inhibitor turret"
+	"mass" 32
```
downRightCoder256:
```suggestion
	"mass" 36
```

Zitchas: Yes, increased mass makes sense.

Zitchas:
Increased the mass slightly, so it more accurately reflects the idea of "two guns + a platform".

data/remnant/remnant.txt
```diff
@@ -292,6 +292,7 @@ shipyard "Remnant Gascraft"
 
 outfitter "Remnant"
 	"Inhibitor Cannon"
+	"Inhibitor Turret"
```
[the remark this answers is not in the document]
Zitchas: OK, I'm good with doing that.

If we're going to do that, could we continue towards merging this anyway so the asset is in place for when the mission calls for it?

data/remnant/remnant outfits.txt
```diff
@@ -51,6 +51,42 @@ effect "inhibitor impact"
...
 
+outfit "Inhibitor Turret"
+	category "Turrets"
+	licenses
+		Remnant
+	cost 1371000
+	thumbnail "outfit/inhibitor turret"
+	"mass" 42
+	"outfit space" -22
+	"weapon capacity" -22
```
Amazinite: I'd say make this take up 28 tons of space instead of 22. That way, a third of the space is saved due to whatever tech the Remnant use, instead of the current nearly-half of the space being missing, which seems a bit excessive. At 28 tons, this becomes comparable to a Dual Sunbeam; similar damage per ton, but the Inhibitor Turret has more range, while the Dual Sunbeam requires less energy and heat for the damage it's putting out.

Beyond that, this looks good to me.

Zitchas: OK, that sounds reasonable. I'll make the change.

PR #4924, opened 2020-03-16
fix: Fully ignore the BACK command if FORWARD is also pressed.

petervdmeer:
**Bugfix:** This PR addresses an issue regarding BACK with reverse thrusters and FORWARD behaving strangely.

## Fix Details
Zogdigi reported on Discord that the behaviour was confusing.
Behavior after this change is quite predictable: pressing FORWARD and BACK at the same time causes BACK to be completely ignored.
And just BACK with reverse thrusters causes reverse thrusters to fire.
While just BACK without reverse thrusters causes the ship to turn against its movement direction.

Discord is currently down for my location, so I cannot respond on the Discordapp.

## Testing Done
Briefly tested FORWARD+BACK, FORWARD-only and BACK-only with ships with and without reverse thrusters. The ships appeared to behave as expected.

## Save File
No save-file provided. But buying reverse thrusters on any ship allows testing of this feature.
I could create a save-file if desired.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -3357,9 +3357,9 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player, Command &activeCommand
 			command |= Command::FORWARD;
 		if(activeCommands.Has(Command::RIGHT | Command::LEFT))
 			command.SetTurn(activeCommands.Has(Command::RIGHT) - activeCommands.Has(Command::LEFT));
-		if(activeCommands.Has(Command::BACK))
+		if(activeCommands.Has(Command::BACK) && !activeCommands.Has(Command::FORWARD))
 		{
-			if(!activeCommands.Has(Command::FORWARD) && ship.Attributes().Get("reverse thrust"))
+			if(ship.Attributes().Get("reverse thrust"))
 				command |= Command::BACK;
 			else if(!activeCommands.Has(Command::RIGHT | Command::LEFT))
 				command.SetTurn(TurnBackward(ship));


Amazinite:
Not gonna lie, I actually use the existing behavior quite a bit.

tehhowch:
>Not gonna lie, I actually use the existing behavior quite a bit.

Is there anything it gets you that forward + turn does not?

Amazinite:
There are a few things that I’ll use W+S for as opposed to W+A or D so I can more easily reach another key at the same time, such as toggling and firing my secondary weapons (Q and E), just given the fingers I use for each key.

tehhowch:
Are those just current shortcomings of your own remapped keybinding though? Seems like you could update your bindings to make `forward+turn + {{other key}}` equally possible as `forward+back + {{other key}}`

Amazinite:
> Seems like you could update your bindings...

Many years of muscle memory would disagree. :)
Seeing as how OP has a thumbs down, looks like I'm not the only one. Even ignoring that though, I don't necessarily see a point in making this change. The forward key tells your ship to thrust forward, whereas the backwards key tells your ship to turn to the opposite direction of your velocity, so I don't even necessarily consider the current behavior to be "behaving strangely" as the OP states. Why restrict both of those commands from being ordered at once?

tehhowch:
Also of note: https://github.com/endless-sky/endless-sky/pull/2982

petervdmeer:
The main issue that Zogdigi was reporting on Discord is that pressing Forward and Back at the same time while reverse thrusters are installed causes the ship to turn as if Back was pressed without reverse thrusters installed.

tehhowch:
Per the linked issue, it is intended behavior that Forward + Back engages the "turn around" instruction, with or without reverse thrust available.

petervdmeer:
> Per the linked issue, it is intended behavior that Forward + Back engages the "turn around" instruction, with or without reverse thrust available.

Ok, if this is the intended behaviour, then let's close this bugfix with the remark: "not a bug, but a feature."

Issue #4116, opened 2019-01-08
System Hazards


Amazinite:
Ref: #3729, https://github.com/endless-sky/endless-sky/pull/3729#issuecomment-439544894, #1553, https://github.com/endless-sky/endless-sky/issues/1552#issuecomment-246232974, http://prntscr.com/m4l25g

System hazards would be akin to the Korath nanobots, applying some sort of damage to ships within the system over time, only this would be applied to all ships equally instead of certain ships taking the brunt of the force and some ships taking no damage at all while the nanobots/any invisible ships are not attacking them.

Hazards would be able to apply any form and combination of damage to ships within the system. This damage could be constant, or it could fluctuate to give the effect of the hazard being like weather, where sometimes things are calm and sometimes they are rough.

Any hazard should ideally have some sort of visual effect within the system and/or to affected ships to convey to the player that the hazard is present. If hazards are able to fluctuate like weather, this effect should change to reflect the severity of the hazard at that time.

Pointedstick:
Even non-destructive hazards like haze would be great. EV Nova had that and it always added a really spooky atmosphere to certain systems.

Zitchas:
Does that haze apply overtop of ships, or just over the background? Also, does it increase over distance?

Amazinite:
Haze right now is just the blue gas (or red gas for the Ember Waste) that you see in the background.

Pointedstick:
Right, I'm thinking about about EVN-style haze that actually obscures things rather than just being visual differentiation for different areas of space.

Zitchas:
I'm not the one writing code, but from my point view the ideal would be having seven attributes that could be assigned to a system:

1) Damage type: (shield, hull, heat, ionization, slow... basically everything that we have)

2) Frequency of hit: none, continuous, or a hit every so many frames.

3) Damage amount: how much damage per "hit" (or damage per frame if frequency is set to continuous)

4) Haze intensity (and this would be haze over top of ships) how much it obscures

5) View distance (how far one can see in through it. Basically, creating a "fade out" distance

6) Tracking interference type(s)

7) Tracking interference intensity.

All that being said, the first three are the original "system hazards" idea, the remainder are things that we could use to make systems much more interesting and varied.

Zitchas:
Yes, exactly! There were a number of systems that had a "fog of war" effect with varying levels of obscurity. Some of them one could only see another ship up to about halfway to the edge of the screen (and that game couldn't even zoom out, either)

[...]

PR #4931, opened 2020-03-19
System hazards

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4116.

## Feature Details
This PR adds a potential system hazards mechanic. System hazards are essentially "weather" events in systems. These weather events are capable of creating visual effects in the system, as well as damaging any ships that are in the system. These weather events will randomly appear (according to how they are defined by the system), and their strength will increase then decrease over the length of the event.

## UI Screenshots
N/A

## Usage Examples
The syntax of `hazard` is as follows:
```
hazard "<hazard name>"
	"constant strength"
	"period" <number>
	"duration" <min> [<max>]
	"strength" <min> [<max>]
	"range" [<min>] <max>
	"environmental effect" "<effect name>" [<amount>]
	weapon
		...

system "<system name>"
	...
	hazard "<hazard name>" [<period>]
	...
```

For the hazard's definition:
* `"constant strength"`: This hazard will not deviate its strength over the course of it being active. Without this, the strength of a hazard changes according to a normal curve starting at about 10%, peaking at 100%, then falling back down to 10% over the course of the event. (I mainly see this being used for hazards with a short duration.)
* `"period"`: This is the chance for the hazard to damage any ships in the system while active. Without this, this hazard will deal damage every frame it is active.
* `"duration"`: This is the minimum and maximum possible duration of this hazard, counted in frames. When a hazard is triggered, a random integer between these two values will be chosen. Without this, the duration of the hazard is always a single frame.
* `"strength"`: This is the minimum and maximum possible strength of this hazard. When a hazard is triggered, a random number between these two values will be chosen. This chosen value then dictates the maximum strength of this hazard event (the 100% that is referred to in the `"constant strength"` description). Without this, the strength of the hazard will always be 1x.
   * Strength is able to change two things: how fast the hazard impacts you (decreasing the period) and how hard the hazard hits (increasing the damage). If a hazard has “constant strength,” the period is not affected, only the damage is increased. 
   * The period is divided by a factor of the square root of the strength, while the damage is multiplied by the same factor. In this way, a strength of 2 is double the DPS of a strength of 1. With constant strength, the damage is multiplied by the strength, not its square root. 
* `"range"`: This is the range from the system center within which ships will be impacted. Without this, a max range is set to a default of 10k (the same distance as the invisible fence past which NPCs avoid). If two numbers are given, then only ships between the two ranges will be damaged.
* `"environmental effect"`: This is the effect that will be created randomly throughout within the min and max range of the hazard while this hazard is active. The amount dictates the number of effects per frame that are created, multiplied by the current strength of the hazard as to provide visual feedback on the strength of the hazard. Without this, no effects are created.
* `weapon`: This is a weapon definition that defines how this hazard impacts ships in the system. The only relevant attributes are the damage attributes (hit force included), any hit effects (which will be created as a spark on the impacted ship, similar to ionization), and blast radius (which will cause the damage to weaken with range). If this hazard has hit force, the hit force comes from the system center. (i.e. positive values will push ships from the system center, while negative values will push ships toward the system center.) This also applies to where blast radius is at its strongest.

For the system's usage of a hazard:
* The hazard name simply refers to any hazard defined within the game, while the period defines how likely it is for this hazard to create a weather event. This acts identically to how fleets spawn in a system. (This also means that if a period is not given, the default period is 200 frames). 
* One system can have multiple hazards, which will trigger independently of one another.

## Testing Done
<details><summary>Tested all of the following without error and with the expected behavior occurring.</summary>

```
hazard "ember waste"
	"duration" 1 100
	"strength" 0.1 1.1
	weapon
		"ion damage" 0.02
		"heat damage" 25
		"slowing damage" .02
```
```
hazard "Ion Storm"
	"duration" 3600
	"strength" 1 2
	"environmental effect" "blaster impact" 10
	"range" 1000
	weapon
		"ion damage" 1
```
```
# Have fun bouncing around with these two.
hazard "Gravity"
	"constant strength"
	"environmental effect" "blaster impact" 50
	range 500 1000
	weapon
		"hit force" -1000

hazard "Repulsor"
	"constant strength"
	"environmental effect" "blaster impact" 6
	range 300
	weapon
		"hit force" 1000
```
</details>

## Performance Impact
Given a large number of hazards in a system for some reason, it might make things run slower (largely due to any visual effects), but for the vast majority of the systems in the game not having any sort of hazard (and I imagine those systems that do eventually get hazards only having one or a few), it shouldn't be an issue.


beccabunny:
As for the effects, I think a good option could be making them work like the leaks, e.g. appearing on a random part of the border of a ship

ravenshining:
Hit Force might be useful to have as an option, toward or away from the system barycentre.  For example, Saggitarius A* could have a system hazard with a hit force, drawing ships towards the black hole,

tehhowch:
As mentioned previously, the proc for hazards should not be done in the Ship class. It should be done in Engine, either before or after checking for projectile collisions. This will also make it simple to do effects, sounds, compute directionality for hit force, and other logic.

Rather than a nested item in a system definition, these should be root level nodes to facilitate reuse and customization, and added to a system via hazard name. These standalone defs should have at minimum a weapon node, but could naturally define effects, sprites, origin positions, radius of effect, etc. (Perhaps one should still be able to override the period and strength in a system definition, but that could be a follow-up feature if highly desired.)

Zitchas:
I've been playing around with the current implementation - and while it takes some fiddling to get some good values - it definitely does what I was hoping and is fun to deal with in gameplay.

Zitchas:
Given that you worked on this during the live-stream, any changes or progress? I'm not sure how "solid" this is, but I'm considering basing a branch of your branch so I can start working on setting up a PR to make use of it.

Amazinite:
> Given that you worked on this during the live-stream, any changes or progress?

The livestream involved basically completely redoing this from the ground up. What is possible in this PR as of right now would still be possible under what I'm attempting to do, but it would also allow for much more interesting results.

Zitchas:
Looks good! I just tried using the artifact exe from here with the demo black hole code you posted on discord, and it works. Isn't nearly strong enough for my permacloak starling to get "stuck" in the gravity well, but it is still neat to see it having an effect there.

One question that just came to me: Is it possible to give random numbers to an effect? Specifically, thinking of how to emulate the buffeting of some kind of wind type weather that knocks a ship around. Hit force as I understand it here is great for pulling things towards the system center or pushing them away from it, but doesn't look like it would be able to knock a ship sideways at all.

Being able to give it a random number between +2 and -2 (so it would randomly pull or push a ship would approximate some of that effect; but being able to give it some kind of variable direction would be much better.

That being said, I don't see that as a necessary feature for system hazards. It would be nice, but it is something that can wait for future tweaks.

Zitchas:
Follow up: I boosted the hit force for Sagittarius A up to -10 (I previously tried -2 and -5), and then it really starts acting how I'd expect a black hole to feel like. I can still escape, even just flying straight outwards against the "pull", but it is enough that I feel like I'm speeding up the process by going back and forth doing a sort of slingshot effect where I use the pull to speed myself up flying towards it to go faster.

It does make me wish that there was some gravitational fall off / acceleration to go with it, but that's secondary, because it feels slightly silly to feel like the ship is being accelerated up to max speed by the hit force and then not accelerate any more beyond that point. Not sure if that's what is actually happening, or if that's just a factor of the lack of references. 

Either way, the gravity definitely works and feels neat. Now I'm going to go play with some Ember Waste effects.

Zitchas:
I "discovered" a repeatable crash:
Here's what I'm using:
```
hazard "ember waste"
	"length" 1 100
	"strength" 0.1 1.1
	weapon
		"ion damage" 0.02
		"heat damage" 25
		"slowing damage" .02

system Pantica
	...
	hazard "ember waste" 50
        ...
```
I'm in pantica, flying my normal perma-cloak starling. I take off, fly around a little, and get a crash. (a full Windows "EndlessSky.exe has stopped working" error message, too. No codes or anything useful, though. The exact time it takes to get the crash varies from about 3secs to 20secs:
- If I'm actively flying, with x2 speed on, then crashes happens in 1-5 secs.
- If I'm actively flying, normal speed, then crash happens around 3 to 15 secs.
- If I take off and just let me ship coast with no inputs at all in normal speed, it takes up to 20secs.

The errors.txt file gives:
```
Skipping hazard attribute with no value specified:
file F:/ZPersonal/ZDev/GitHub/ZEndlessSky/endless-sky/data/map.txt
L11:   hazard "ember waste"
L14:     weapon
```

It is odd that it says "skipping hazard attribute" as everything is specified and the game does appear to be appying heat, ionization, and slowing to my ship. Very slowly, though. My starling is keeping up fairly well.

Amazinite:
Fixed the error with weapon "not being recognized" (even though it was). Investigating the crash. It's notable that adding "no deviation" seems to be stopping the crash.

source/Hazard.cpp
```diff
@@ -0,0 +1,120 @@
...
+#include "GameData.h"
+#include "Random.h"
+
+using namespace std;
+
+
+void Hazard::Load(const DataNode &node)
+{
+	if(node.Size() < 2)
+		return;
```
Amazinite, line 26: Just as an aside, I've noticed that there's a lack of consistency among a lot of Load functions across the various classes. Some have an error if there is no name, others just return. Some use a "key" that is `child.Token(0)` then check against that, or use a `hasValue` boolean and a `value` double, others don't. Might be worth organizing all that at some point.

ravenshining:
I had an idea in my sleep about this - Could it be possible to attach hazards to star definitions?

Say one adds a bunch of systems and several of them have black holes or one of them has several black holes, boom, they've got some stock black hole hazard without having to define it for each black hole.

Or, lets say one wants a star type to give off some heat in their immediate vicinity, all one would have to do is add the hazard to the star in stars.txt and all systems with that star would gain that hazard without having to go in and find each one.

Amazinite:
I mean, potentially I could do that, but I don't see that as much of a need. There are already other system attributes that relate to the star(s) in the system that must be defined by the system (e.g. habitable), so I don't see why we'd make `hazard` be a potential aspect of the star(s). What's the issue with just copy/pasting a single line to the various systems that need it?

(Although, that could be said to be an argument for starting to make more things an aspect of the star, in which case I'd make that something we tackle in a separate PR.)

Zitchas:
The capability of including hazard data in the star definition would be potentially useful. But at the same time, the hazard data shouldn't be restricted to it. Some hazards could definitely be attached to specific star types (say, blackholes?), but most are more a feature of the system rather than the star itself.

I could see that being particularly useful for someone wanting to make a procedural generation system for Endless Sky. (I think there already is one, actually)

ravenshining:
Zitchas, I was thinking of having it defined *either* as part of a star *or* as part of a system, not instead of.

> What's the issue with just copy/pasting a single line to the various systems that need it?

Because for how I'd want to use it, that would mean finding and pasting the appropriate definition to *every single system on the map*.

Zitchas:
@ravenshining I think I'd prefer if it was *in addition to.* As such, a particular type of star could have a particular hazard, and systems can have their own hazards. If a system with hazards happens to also have the star, then the system would have both hazards. Which sounds good to me. Potentially very challenging, but that's not really a problem either. And if it makes it easier to do something like giving all stars varying amounts of heat damage or gravitational pull or something... That sounds good.

ravenshining:
Isn't that what I said?  I think you're just using different phrasing to describe the same idea.

Amazinite:
I'll look into making that a new feature, among others, after this is merged (instead of bloating this PR).

source/Engine.cpp
```diff
@@ -1156,12 +1158,27 @@ void Engine::EnterSystem()
...
 	for(int i = 0; i < 5; ++i)
+	{
 		for(const System::FleetProbability &fleet : system->Fleets())
 			if(fleet.Get()->GetGovernment() && Random::Int(fleet.Period()) < 60)
 				fleet.Get()->Place(*system, newShips);
+		for(const System::HazardProbability &hazard : system->Hazards())
+			if(Random::Int(hazard.Period()) < 60)
+			{
+				const Hazard *weather = hazard.Get();
+				int hazardLifetime = weather->Length();
```
Pointedstick, line 1175:
```suggestion
				const int hazardLifetime = weather->Length();
```

source/Hazard.cpp
```diff
@@ -0,0 +1,120 @@
+/* Hazard.cpp
+Copyright (c) 2020 by Michael Zahniser
```
Pointedstick, line 2: Should be your copyright

source/Hazard.h
```diff
@@ -0,0 +1,57 @@
+/* Hazard.h
+Copyright (c) 2020 by Michael Zahniser
```
Pointedstick, line 2: Ditto

source/Hazard.h
```diff
@@ -0,0 +1,57 @@
...
+class Hazard : public Weapon {
+public:
+	void Load(const DataNode &node);
+	
+	const std::string &Name() const;
+	// Does the strength of this hazard deviate over time?
+	bool Deviates() const;
+	// How often this hazard deals its damage while active.
+	int Period() const;
+	// Generates a random integer between the minimum and maximum length of this hazard.
```
Pointedstick, line 31: Maybe rename this `randomLength()` for greater accuracy

source/Hazard.h
```diff
@@ -0,0 +1,57 @@
...
+	int Length() const;
+	// Generates a random double between the minimum and maximum strength of this hazard.
+	double Strength() const;
```
Pointedstick, line 34: Similarly, `randomStrength()` might be clearer

source/Ship.cpp
```diff
@@ -2729,14 +2730,47 @@ double Ship::MaxReverseVelocity() const
 
 
 
+// Take environmental hazard damage from the current system.
+void Ship::DoHazard(vector<Visual> &visuals, const Hazard *hazard, double strength)
+{
+	double distance = position.Length();
```
Pointedstick, line 2736:
```suggestion
	const double distance = position.Length();
```

source/Ship.cpp
```diff
@@ -2729,14 +2730,47 @@ double Ship::MaxReverseVelocity() const
...
+	double maxRange = hazard->MaxRange();
```
Pointedstick, line 2737:
```suggestion
	const double maxRange = hazard->MaxRange();
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
+/* Weather.cpp
+Copyright (c) 2020 by Michael Zahniser
```
Pointedstick, line 2: Your copyright

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+	if(hazard->Deviates())
+	{
+		// If the square root of the strength is greater than the period, then Period()
+		// will return 1. Given this, we need to multiply the amount of strength
+		// going toward the damage by some corrective factor. Figure out what the "true
+		// period" is (without it bottoming out at 1) and divide that with the current
+		// period in order to correctly scale the damage so that the DPS of the hazard
+		// will always scale properly with the strength.
+		// This also fixes some precision lost by the fact that the period is an integer.
+		double truePeriod = static_cast<double>(hazard->Period()) / sqrt(Strength());
```
Pointedstick, line 70:
```suggestion
		const double truePeriod = static_cast<double>(hazard->Period()) / sqrt(Strength());
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+		double multiplier = static_cast<double>(Period()) / truePeriod;
```
Pointedstick, line 71:
```suggestion
		const double multiplier = static_cast<double>(Period()) / truePeriod;
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+}
+
+
+
+int Weather::Step(vector<Visual> &visuals)
+{
+	// Environmental effects are created by choosing a random angle and distance from
+	// the system center, then creating the effect there. If a hazard has no range,
+	// effects are only created out to the invisible fence.
+	double minRange = hazard->MinRange();
```
Pointedstick, line 85:
```suggestion
	const double minRange = hazard->MinRange();
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+	double maxRange = hazard->MaxRange();
```
Pointedstick, line 86:
```suggestion
	const double maxRange = hazard->MaxRange();
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+	if(!maxRange)
+		maxRange = 10000.;
+	for(const auto &effect : hazard->EnvironmentalEffects())
+		for(int i = 0; i < static_cast<int>(effect.second * Strength()); ++i)
+		{
+			Point angle = Angle::Random().Unit();
+			double magnitude = (maxRange - minRange) * sqrt(Random::Real());
```
Pointedstick, line 93:
```suggestion
			const double magnitude = (maxRange - minRange) * sqrt(Random::Real());
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+			Point pos = (minRange + magnitude) * angle;
```
Pointedstick, line 94:
```suggestion
			const Point pos = (minRange + magnitude) * angle;
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+	return --lifetime;
+}
+
+
+
+double Weather::Strength() const
+{
+	if(hazard->Deviates())
+	{
+		double castTotal = static_cast<double>(totalLifetime);
```
Pointedstick, line 106:
```suggestion
		const double castTotal = static_cast<double>(totalLifetime);
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+		double castCurrent = static_cast<double>(lifetime);
```
Pointedstick, line 107:
```suggestion
		const double castCurrent = static_cast<double>(lifetime);
```

source/Weather.cpp
```diff
@@ -0,0 +1,115 @@
...
+		// Using a deviation of totalLifetime / 4.3 causes the strength of the weather to start and end at about 10% the maximum.
+		double DEVIATION = castTotal / 4.3;
```
Pointedstick, line 109:
```suggestion
		const double DEVIATION = castTotal / 4.3;
```

Also, why is the variable name all uppercase?

Amazinite: This is a constant that should really only need to be calculated when the Weather is created, not every time this function is called, hence why I had it all caps.

source/Engine.cpp
Amazinite: While this variable (and many others that you've commented about) is "constant," the scope of the variable doesn't leave this if statement (or a single for loop iteration), so labeling it as const seem unnecessary. (See #4520 where a similar scenario occurs where I labeled various variables as const but tehhowch said it was unnecessary.)

Zitchas:
Alright, things still look good to me.

Zitchas:
I see the 0.9.13 milestone tag was removed from this? I was under the impression that it was ready to go (it certainly works and accomplishes what I was hoping for it).

Amazinite:
@Zitchas I removed the milestone from those PRs of mine that have an open issue, and updated the milestones on those issues from 1.0.0 to 0.9.13.

Zitchas:
Ah, OK. Thanks for the explanation. I was worried that meant it was being pushed off to 0.9.15 or something.

source/Hazard.h
```diff
@@ -0,0 +1,57 @@
...
+
+// Hazards are environmental effects created within systems. They are able to create
+// visual effects and damage or apply status effects to any ships within the system
+// while active.
+class Hazard : public Weapon {
+public:
+	void Load(const DataNode &node);
+	
+	const std::string &Name() const;
+	// Does the strength of this hazard deviate over time?
```
tehhowch, line 27: Should this constant vs. variable strength attribute be a property of the `Hazard` itself, or of a particular system "weather" effect?

You may be able to simplify this PR by re-evaluating the responsibilities of the Hazard class and of the Weather class.

source/Ship.cpp
```diff
@@ -2729,14 +2730,47 @@ double Ship::MaxReverseVelocity() const
 
 
 
+// Take environmental hazard damage from the current system.
+void Ship::DoHazard(vector<Visual> &visuals, const Hazard *hazard, double strength)
```
tehhowch, line 2734: I don't think this method should be in the Ship class; it belongs in Engine alongside the collision checks. Note that the CollisionSet of ships can be used to determine the subset of ships affected by a given hazard.

source/Engine.cpp
```diff
@@ -1331,6 +1348,9 @@ void Engine::CalculateStep()
 	// missile systems ready to fire.
 	hasAntiMissile.clear();
 	
+	// See if any new system weather has started and perform any active weather events.
+	DoWeather();
+
```
tehhowch, line 1353: Is it intended that this occurs after the other "generate new things that happen this step" actions?

source/Engine.cpp
```diff
@@ -1912,6 +1932,41 @@ void Engine::DoCollisions(Projectile &projectile)
...
+	// environmental effects.
+	for(auto it = activeWeather.begin(); it != activeWeather.end(); ++it)
+	{
+		if(it->HasWeapon() && !Random::Int(it->Period()))
+			for(const shared_ptr<Ship> &ship : ships)
+				if(ship->GetSystem() == playerSystem)
+					ship->DoHazard(visuals, it->GetHazard(), it->DamageMultiplier());
+		
+		// If this weather event has ended, remove it from the list.
+		if(!it->Step(visuals))
```
tehhowch, line 1960: Probably this should be `newVisuals` if you update when this is called.

source/Engine.cpp
```diff
@@ -1912,6 +1932,41 @@ void Engine::DoCollisions(Projectile &projectile)
...
+	}
+	
+	// Go through each active weather event to deal damage to ships and generate
+	// environmental effects.
+	for(auto it = activeWeather.begin(); it != activeWeather.end(); ++it)
+	{
+		if(it->HasWeapon() && !Random::Int(it->Period()))
+			for(const shared_ptr<Ship> &ship : ships)
+				if(ship->GetSystem() == playerSystem)
+					ship->DoHazard(visuals, it->GetHazard(), it->DamageMultiplier());
```
tehhowch, line 1957: This should probably be done as a part of the collisions method

source/Weather.cpp
```diff
@@ -0,0 +1,113 @@
...
+		// If the square root of the strength is greater than the period, then Period()
+		// will return 1. Given this, we need to multiply the amount of strength
+		// going toward the damage by some corrective factor. Figure out what the "true
+		// period" is (without it bottoming out at 1) and divide that with the current
+		// period in order to correctly scale the damage so that the DPS of the hazard
+		// will always scale properly with the strength.
+		// This also fixes some precision lost by the fact that the period is an integer.
+		double truePeriod = hazard->Period() / sqrt(Strength());
+		double multiplier = Period() / truePeriod;
+		return sqrt(Strength()) * multiplier;
```
tehhowch, line 74: I feel like this calculation (or a similar one) is done somewhere in the standard library for us already, or can instead be tabulated at startup in a manner similar to the Angle cache.

At a minimum, you're using the term `sqrt(Strength())` twice, so cache it.

source/Weather.h
```diff
@@ -0,0 +1,55 @@
...
+
+class Hazard;
+class Visual;
+
+
+// Weather is used to contain an active system hazard, keeping track of the hazard's lifetime, its strength, and if it should cause any damage.
+class Weather {
+public:
+	Weather() = default;
+	Weather(const Hazard *harzard, int lifetime, double strength, int totalLifetime = -1);
```
tehhowch, line 30: `harzard`

Also, this should be `explicit`

source/Weather.h
```diff
@@ -0,0 +1,55 @@
...
+	
+	const Hazard *GetHazard() const;
+	// Whether the hazard of this weather deals damage or not.
+	bool HasWeapon() const;
+	// The probability of this weather dealing damage while active.
+	int Period() const;
```
tehhowch, line 36: Probability or Period?

source/Weather.h
```diff
@@ -0,0 +1,55 @@
...
+#define WEATHER_H_
+
+#include "Hazard.h"
+#include "Visual.h"
+
+#include <string>
+#include <vector>
+
+class Hazard;
+class Visual;
```
tehhowch, line 23: You both include and forward declare these classes.

Zitchas:
One thing that occurred to me today: A while back we were discussing whether system hazards should be attached to the system, or to a specific object in the system, or both. This was in the context of having specific stars (such as black holes) have specific hazards attached to them. 

I'm not sure if there was a decision on that, but both myself and RavenShining seem to be in accord that we like the idea of having them being attachable to both systems and specific objects in the system. (some hazards are specific to the system, irrespective of the star in it; some hazards are general to specific kinds of stars).

The idea I had today was this: Could this be used to give planets or stations ground-based defences/artillery/etc? The mechanics as I understand them are basically treating the system (firing from 0,0) or the star (firing from its position) as shooting an invisible projectile at everything in the system that has the specified effects, as per the weapon definition. It should be fairly straightforward to use the exact same process (albeit much slower) to fire regular looking projectiles at things that their government is hostile to.

This idea, obviously, doesn't need to be included in this PR, but it is an interesting idea/application for it to enable something else that some people have expressed an interest in.

source/Hazard.cpp
```diff
@@ -0,0 +1,120 @@
...
+			maxStrength = max(0., (child.Size() >= 3) ? child.Value(2) : 0.);
+		}
+		else if(key == "range")
+		{
+			minRange = max(0., (child.Size() >= 3) ? child.Value(1) : 0.);
+			maxRange = max(0., (child.Size() >= 3) ? child.Value(2) : child.Value(1));
+		}
+		else if(key == "environmental effect")
+		{
+			int count = (child.Size() >= 3) ? child.Value(2) : 1;
```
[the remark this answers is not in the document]
Amazinite: This is copied from a place like Weapon.cpp, where no such check for negative effect counts is done. If we do something here, we should reevaluate all other effects as well. I imagine though that a negative value is treated like zero, because somewhere along the line this count is just being fed into a for loop with the counter initialized at 0.

source/System.h
```diff
@@ -70,6 +71,18 @@ class System {
...
 	
+	class HazardProbability {
+	public:
+		HazardProbability(const Hazard *hazard, int period);
+		
+		const Hazard *Get() const;
+		int Period() const;
+		
+	private:
+		const Hazard *hazard;
```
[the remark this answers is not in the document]
Amazinite: This is copied from FleetProbability, so if we change this then we need to reevaluate FleetProbability as well.

source/Hazard.h
Amazinite: Currently, the only thing the system controls is the probability of a weather event activating. Beyond that, all aspects of a particular hazard are handled by the `Hazard` class.

source/Ship.cpp
Amazinite: CreateSparks is a private method of Ship, so either we need to make CreateSparks public, or Ship needs a method that handles taking Hazard damage so that it can create the proper effects.

source/System.h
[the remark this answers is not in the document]
tehhowch: A different PR can address that tech debt refactor.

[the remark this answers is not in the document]
tehhowch: A different PR can address that tech debt refactor.

source/Ship.cpp
tehhowch: Use CollisionSet::Circle?

source/Hazard.cpp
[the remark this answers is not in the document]
tehhowch: We can evaluate other effects in a different PR

source/Ship.cpp
Amazinite: That requires a radius, meaning that would require that all hazards have a range, whereas right now the range is optional and not having it means it doesn't matter where the ship is in the system. That would also make it someone clunky to deal with the minimum and maximum range, as we'd need to get all ships within the max range then all the ships within the min range and remove the ships in the min range from the max range before figuring out which ships should actually get affected.

tehhowch: Using collisionset::circle should actually offer improved performance, due to the grid lookup. To support a minimum range, you would call Circle with the max range and then exclude those ships closer than the minimum range. If that isn't performant enough, we would add a new method to CollisionSet for retrieving ships within an annular slice.

I also note that a no-range-given hazard still has a range, of the fence, so you can use that. 

Another benefit of using CollisionSet is that we could (eventually) support line-of-sight hazards that can be blocked by other ships or asteroids.

Amazinite: Damage still goes beyond the fence. The fence is only used as the limit of where environmental effects are created since we aren't tiling them.

tehhowch: For an infinite range hazard, iterating in-system ships directly will be fine (hard to beat having explicit knowledge about the solution space). For the others, circle will be best

Amazinite: I've actually gone ahead and removed infinite range hazards. A hazard with no specified range now has a max range of 10k (the system fence). This means that the environmental effects of a hazard will always indicate exactly where damage is dealt, instead of effects ending after a certain range but damage still occurring.

Amazinite: Implemented a CollisionMask::Ring method so that we don't need to run CollisionMask::Circle twice and see what's in one but not the other (which, as mentioned on Discord, actually had edge case issues).

vkr-13:
> we like the idea of having them being attachable to both systems and specific objects in the system

I agree, it allows more flexibility in system creation and in the cases where there may be two or more separate effect sources of interest, this would cause the least hassles by far.

Zitchas:
Question: How close to "finalized" are the definitions in the original post? Can I start building content with them with a reasonable expectation they aren't likely to change?

Amazinite:
You should be good to use this as is described. Any changes at this point will just be tweaks to organize and optimize things.

Zitchas:
Just having fun nailing down some reasonable values for the Ember Waste hazards, and just had a thought: We currently need to set a range, or have it default to the invisible fence radius of 10k. For some things (especially in the Ember Waste where the hazards are "universal" sorts of things), I'd prefer to have some sort of "infinite" range. Aka, it doesn't matter where they are, they are getting hit with it. 

Granted, I know I can just put in an arbitrarily large value as the range (say, 100k, or maybe 1000k just be thorough), but there is still the possibility that players find the edge of it, and then get to wonder why there is an edge to a seemingly infinite effect, whereas I would prefer to have the lore be that it actually is ember-waste-wide.

PR #4936, opened 2020-03-21
Fix: Removed "Navy" license from Carrier and Cruiser

Zitchas:
These two ships have a higher license restriction - that of Carrier and Cruiser respectively. Given that these two licenses are both only given after a player has already received the Navy one, and are themselves Navy restricted licenses, there is no point in having two license restrictions instead of just listing the more restrictive of the two.

The reason for this change are:
- Consistency with all other license configurations where ships are restricted based on the highest license required.
- Clarity: Requiring two different licenses implies situations where one might have a carrier license but not a Navy one. Given that the Carrier and Cruiser licenses are at this point specifically higher Navy licenses, I don't see this happening.

Additional Changes:
- The "Carrier" license has been updated to "Navy Carrier".
- The "Cruiser" license has been updated to "Navy Cruiser".
- The Carrier and Cruiser ship entries have had their licenses updated to reflect these name changes.

These changes clarify that these licenses are for their specific category of ships within the Republic Navy; and are not general category licenses.


Amazinite:
> Side issue: Given that we have all sorts of licenses in the game, we may want to consider renaming those two licenses "Navy Cruiser" and "Navy Carrier" instead. Because nothing in their current name actually connects them to the Navy, and they are irrelevant to all the various other factions in the game.

If you are going to be removing the Navy license requirement from these ships, it does make sense to rename the other license requirements like this; the Wanderer warships do have a "Wanderer Military", not just "Military", license requirement after all.

Zitchas:
I made the change, so the licenses are now updated to reflect their factional alignment.

tehhowch:
Don't the license outfits need to be updated too?

Amazinite:
The outfits are updated, it looks.

tehhowch:
Sorry, the sprites and copyrights

Amazinite:
The sprite name doesn't always match the outfit name (see: all the Hai ships), so I don't see the need.

Zitchas:
Yes, I'm just updating the outfit images. I'm not seeing any mention of carrier or cruiser in the copyright file, so I think they may be just subsumed under the broad copyright section:
```
Files: images/*
Copyright: Michael Zahniser
License: CC-BY-SA-4.0
```

Zitchas:
Given that there is no in-game way to acquire these licenses, we don't need to worry about making entries in `depreciated`, do we? I know there are some plug-ins that grant these, but I don't think we need to clutter the game files with depreciated stuff in order to maintain plug-ins.

Amazinite:
Sounds like it'd be the nice thing to do, but you'd actually need a mission to check for the old licenses and add the new ones, as having depreciated versions of these outfits would do nothing given the change in the license requirement for the ships.

data/human/navy license compatability.txt
```diff
@@ -0,0 +1,34 @@
...
+# Compatibility patch for those who use plug-ins that give all the licenses.
+# Each one checks for the presence of one of the two licenses, then replaces it.
+mission "Navy Cruiser License Compatibility"
+	landing
+	invisible
+	to offer
+		has "license: Cruiser"
+	on offer
+		set "license: Navy Cruiser"
+		clear "license: Cruiser"
```
Amazinite: Don't clear the old licenses; plugin ships could be using this version. I'd also simply put these missions in the depreciated outfits file instead of creating a whole new file for just these.

data/human/outfits.txt
```diff
@@ -409,14 +409,14 @@ outfit "Navy License"
...
 	category "Special"
-	thumbnail "outfit/cruiser license"
+	thumbnail "outfit/navy cruiser license"
 	description "It is a very rare privilege to be permitted to purchase a Navy capital ship for private use - not just because of a Cruiser's firepower, but because it is a potent symbol of the Republic's strength."
 
-outfit "Carrier License"
+outfit "Navy Carrier License"
 	category "Special"
-	thumbnail "outfit/carrier license"
+	thumbnail "outfit/navy carrier license"
```
Amazinite: If we didn't change the image name then we wouldn't need a high-DPI plugin as well, but given that it's already opened, little point in going back now I guess.

data/human/navy license compatability.txt
Zitchas: *Are* there any Republic Navy plug-ins?

That being said, I'll make the change.

data/human/outfits.txt
Zitchas: OK.

data/human/navy license compatability.txt
Zitchas: Changed.

PR #4940, opened 2020-03-22
Tweak: Remnant-Ka'het, Taely's dialogue

Zitchas:
Adjusting a sentence where Taely talks about the Ka'het and their tech. This adjustment just tweaks the wording so that Taely states that the Ka'het are a source of inspiration instead of they are their *main* inspiration.

While the Remnant certainly draw much inspiration from the Ka'het, I'd like to keep it a touch more vague about where they draw their inspiration from. Stating that the Ka'het are a source of inspiration is good. Stating that the Ka'het are their main source of inspiration is just a bit too absolute.

### DIFF
--- a/data/ka'het/ka'het missions.txt
+++ b/data/ka'het/ka'het missions.txt
@@ -69,7 +69,7 @@ mission "First Contact: Ka'het: Remnant 1B"
 					defer
 			`	You find Taely inside one of the outer hangars of the shipyard, apparently teaching a small group of people how to make basic repairs on an Albatross. You sit on a bench waiting for her to finish her lesson, after which she walks over to you. "Hello, <first>!" she says. "Is there something I can help you with?" `
 			`	"I'm here because I need to ask you a few questions," you reply.`
-			`	As she sits near you, you start explaining to her the area you reached, including the weird alien ships that attacked you with weapons similar to the ones the Remnant possess, although this doesn't seem to be news for her. "We know of the existence of the Ka'het, the aliens you encountered, since we first reached the Ember Wastes. We found many of their 'shells' within our planets, crashed there thousands, perhaps even tens of thousands of years before our arrival, but some of their outfits survived long enough for us to reverse engineer these advanced technonologies."`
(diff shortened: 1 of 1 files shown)


data/ka'het/ka'het missions.txt
```diff
@@ -82,7 +82,7 @@ mission "First Contact: Ka'het: Remnant 1B"
 				`	"Wait, why did you call them 'shells' and not ships?"`
 					goto shells
 			label tech
-			`	"Back then we found large numbers of their EMP torpedoes, most of which in almost pristine conditions; their internal mechanisms were also simple enough that we were able to reproduce them perfectly. We didn't find any other suriving systems in the 'shells' to do the same, but their tech have been our main inspiration point for centuries."`
+			`	"Back then we found large numbers of their EMP torpedoes, most of which in almost pristine conditions; their internal mechanisms were also simple enough that we were able to reproduce them perfectly. We didn't find any other suriving systems in the 'shells' to do the same, but their tech has been a source of inspiration for centuries."`
```
tehhowch: `most of which in` sounds odd, maybe `most in`?

`suriving` -> `surviving`

Why is `shells` quoted in this line?

Zitchas: I think it is because they are actually ships, not shells. But the ships serve the same purpose that a shell would serve for something like a snail.

Zitchas: Also, fixed:
suriving -> surviving
most of which in -> most in

tehhowch: I bring up the quoting of shells by Taely since it isn't something that she is stressing or quoting - it's normal. When the player responds to the usage of shells, then the emphasis on the word is correct.

Zitchas: OK, removed the quotes, as I agree with your comment. Also reworked the sentence a bit.

tehhowch:
Was that the only use of quotes around shells by Taely?

Zitchas:
Nope, there were a couple other spots. They have been changed. Now only the player uses quotes, Taely just uses the term as if it were a normal thing.

Issue #4682, opened 2019-11-22
Whitelist for negative attributes


Amazinite:
**Is your feature request related to a problem? Please describe.**
As of right now, the game prevents the player from installing/selling outfits if ANY attribute will become negative. It may be the case though that we want certain attributes to be able to go negative. Examples of such attributes would be those from #4525, which have been intentionally designed to allow negative values with differing effects.
![image](https://user-images.githubusercontent.com/17688683/69455153-c5222b00-0d35-11ea-8047-430aa72b9175.png)

**Related Issue Links**
#1128 The negative attribute warning has plagued me for some time now.
#4528 The ability for `attribute` to check negative attributes was mentioned here, but negative attributes are actually rare because of this issue, leading to the only examples of negative attributes being those set on ship hulls (e.g. the Arfecta).

**Describe the solution you'd like**
Create some sort of whitelist for attributes that dictates whether they can be negative or not. This whitelist would preferably be in the game data so that plugin makers can add to it if they have some attribute that they want to allow the player to have a negative value of.

**Describe alternatives you've considered**
We could also do the reverse and create a blacklist of attributes that we don't want to be negative. This blacklist would include cargo, outfit, weapon, engine space, gun ports, turrets, etc. All the stuff on ships that we don't want to have be negative. I could more reasonably see this being only in the source code, but a blacklist would likely have issues dealing with ammo capacity values.

petervdmeer:
I'm not convinced by the `shield generation multiplier` example you link to. If the `shield generation multiplier` would be negative, then you would actually lose shield every second, and you would lose more shields when you have more shield generators installed. Technically this would work, but it is not really intuitively. For hulls it might even be more problematic (ships automatically dying).

I do like the current implementation that no single value ever goes below zero. Anybody creating new attributes can consider this and do whatever is needed to make the attributes work (in case of shield generation, do something like split it into two keywords, `shield generation boost` and `shield generation limiting`).

If we do really need to implement this, then I would prefer a whitelist over a blacklist, the default option should be that values don't go below zero.

Amazinite:
Negative shield generation values have zero effect on a ship (as the DoRepair function in Ship.cpp does nothing if the generation is at or below zero). And in the example of multipliers, multipliers affect stats by `stat * (1 + multiplier)`, meaning that values between -1 and 0 don’t cause negative values of the stat, but reduced values.

petervdmeer:
Okay, I wasn't aware that negative values where automatically ignored.

PR #4942, opened 2020-03-23
Attribute whitelist that allows negative attributes

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4682 

## Feature Details
Adds a hardcoded whitelist of attributes that should be allowed to be negative on a ship. To start, I've added the repair cost attributes that as of #4932 have functionality when negative. (Ref: #4525 from where I copied over some of this code.)

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
I put `"shield energy" -1` on some cooling ducts and installed a bunch.  The recharging shields energy cell of the energy/heat table became positive, showing that shield energy was now at a net negative; previously, this would not be possible, as the outfitter would block you from having negative attributes. And thanks to #4932, being attacked meant my energy production shot up.

## Performance Impact
N/A


Zitchas:
I know the initially impression of taking hits *generating* energy sounds odd initially, but for, say, a water cooling system, just imagine having a "steam turbine" outfit that could be added on to it. When the system produces a lot of heat, it runs the turbine and generates power! Theoretically, anything that produces heat could be hooked up to a turbine system.

tehhowch:
I'm not a fan of a user-configured whitelist, as it's extremely easy to "break" the game. Consider what happens when the whitelist include things like "outfit space", any of the ammo capacity attributes, etc. 

While offering customization of the game experience is something that plugins should be able to do, fundamentally violating the core constructs of the game is not.

Amazinite:
If somebody really wants to break the game, I don't see a reason why we should stop them. I can hard code the whitelist in Outfit.cpp though; any use case I can imagine for an easily customizable whitelist all involved being combined with #4528, but with the suggestion of that being expanded to allow basically using a ConditionSet with attributes, anything that would be done with a single attribute that can be negative could also be done with a comparison of two attributes.

e.g. say we have an alien civilization with outfits that have a "taboo" attribute; some outfits are taboo to have (positive values), while others are seen as good to have (negative values), then missions could be offered based off of your net taboo value. Without allowing negative attributes though, the good outfits could instead have a different attribute (let's say "not taboo" because I can't think of anything else), then to do the same mission offering you'd just do `attribute "not taboo" >= "taboo"` or what have you.

Amazinite:
@tehhowch Everything look good here now?
(Note to self: #4519 needs minimum values of -0.99, #4525 needs minimum values of -1.)

source/Outfit.cpp
```diff
@@ -201,10 +216,21 @@ int Outfit::CanAdd(const Outfit &other, int count) const
 {
 	for(const auto &at : other.attributes)
 	{
+		// The minimum allowed value of most attributes is 0. Some attributes
+		// have special functionality when negative, though, and are therefore
+		// allowed to have values less than 0.
+		double minimum = 0.;
+		if(WHITELIST.count(at.first))
```
tehhowch, line 223: Don't double-scan the map, by using count and then find.

Amazinite: So instead 
```
auto it = WHITELIST.find(at.first);
if(it != WHITELIST.end()) {
	minimum = it->second;
	...
}
```
?

PR #4946, opened 2020-03-24
Bug fix: ships at Waypoint to refuel you

NomadicVolcano:
Now that the Hai ships in Waypoint scan you, they refuse to refuel you which does not match the expected behavior of Hai when they see a pilot in trouble. ~~This update adds a permanent greeter at Waypoint with extra fuel. There is only one so it is plausible for fast pirates to slip by it. The greeter is well-armed though, so it is unlikely to be disabled by accident.~~

This adds infrequent Hai merchants at Waypoint, and increases the frequency of Human merchants there. It also adds some rare merchant ships at Ultima Thule if you get stuck there.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -23587,8 +23587,6 @@ system "Ultima Thule"
 	trade Medical 732
 	trade Metal 417
 	trade Plastic 360
-	fleet "Small Human Merchants (Hai)" 15000
-	fleet "Large Human Merchants (Hai)" 20000
 	object
 		sprite star/f0
 		period 10
--- a/data/map.txt
+++ b/data/map.txt
@@ -23587,6 +23587,8 @@ system "Ultima Thule"
 	trade Medical 732
 	trade Metal 417
 	trade Plastic 360
+	fleet "Small Human Merchants (Hai)" 30000
+	fleet "Large Human Merchants (Hai)" 40000
 	object
 		sprite star/f0
 		period 10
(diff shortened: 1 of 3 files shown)


Amazinite:
Might it make more sense to just increase the frequency of small human fleets in Waypoint? Or readd small Hai fleets/swap the surveillance fleets back with small Hai fleets?

NomadicVolcano:
> Might it make more sense to just increase the frequency of small human fleets in Waypoint? Or readd small Hai fleets/swap the surveillance fleets back with small Hai fleets?

Yes. Please check the latest changes.

data/map.txt
```diff
@@ -23587,6 +23587,8 @@ system "Ultima Thule"
 	trade Medical 732
 	trade Metal 417
 	trade Plastic 360
+	fleet "Small Human Merchants (Hai)" 15000
+	fleet "Large Human Merchants (Hai)" 20000
```
Amazinite: I'd keep Ultima Thule as an empty system. Therefore if the player runs out of fuel in Ultima Thule, there only option is to head through the wormhole. (Helps to make Hai space a little more discoverable.)

NomadicVolcano: Okay. Please just make sure you fix refueling at Waypoint before 0.9.12. The fact that it takes only 400 fuel to get safely to Hai space is pretty important.

Amazinite: > Please just make sure you fix refueling at Waypoint before 0.9.12. 

Well isn't that what this PR is for? So you should, you know, address the review? >insert James emote<

NomadicVolcano: Somehow I didn't notice there was a review.

Amazinite: All good. I got it.

PR #4956, opened 2020-03-28
More Ka'het names

beccabunny:

## Summary
This PR expands the possible names of the Ka'het and the Ka'mar, by adding different options (and splitting into phrases when needed). The total of the Ka'het names is now 20,988 if I did my calculations correctly, but it's just syllabes together
Screenshot (1507)
Screenshot (1510)


tehhowch:
Please update the spellings from syllabe to `syllable`

beccabunny:
Should  be fixed now, thank you

data/ka'het/ka'het.txt
```diff
@@ -10,21 +10,34 @@
...
+		"Aerite"
+		"Letis"
+		"Maiar"
+		"Lateria"
+		"Gertima"
+		"Santari"
+		"Antamesa"
+		"Delia"
+	word
+		" "
```
Amazinite, line 38: I don't think that this should be a space, as "ka'het syllable" words don't start capitalized. An apostrophe would be the obvious, boring replacement, so perhaps a hyphen `-` instead, or no spacer character at all.

data/ka'het/ka'het.txt
```diff
@@ -57,6 +95,21 @@ phrase "ka'mar"
...
+		"Gas"
+		"Tie"
+		"Esi"
+		"Ora"
+		"Tana"
+		"Fae"
+		"Gar"
+		"Fasi"
+	word
+		" "
```
Amazinite, line 108: Same here.

data/ka'het/ka'het.txt
beccabunny: done

data/ka'het/ka'het.txt
beccabunny: done

PR #4871, opened 2020-03-01
Remnant: Cognizance 1

Zitchas:
**Content (Missions)**
## Summary
A new mission set continuing the story of the Remnant and void sprites, in which the Remnant conduct more research, discover a hostile presence hunting the void sprites, and the Remnant embark on a means of preserving them. Meanwhile, something strange appears in the Ember Waste.

Adds:
- 14 visible missions (and 1 invisible mission)
- 1 new ship (the ember waste node - not acquirable)
- 6 new systems (5 are present from the start, 1 appears via event)
- 1 new visit-able station (found in the system that appears via event)
- 3 new wormholes (1 are present from the start, 2 appear via event)
- Makes all the remaining planets in the Remnant trio of systems visitable gas-giants.

## Lore implications:
- The Remnant know how to lure and capture void sprites
- The ability to create quasi-cloaking effects on a station-sized scale is fairly old Remnant tech. (quasi-cloak as in making the subject ghost-like, insubstantial, and virtually invisible to sensors)
- The Remnant know how to build space stations in gas-giants (at least, in the upper atmosphere, but deep enough to require gaslining)
- There is another entity that seems particularly interested in the Void Sprites, but is unknown, aside from demonstration hints at telepathy (nowhere near as coherent or refined as the Archons, though).
- There was a hidden system in the Ember Waste.

To-do:
 - [X] New descriptions for Remnant gas-giants ~~or remove them from this PR.~~
 - [X] New images for Remnant gas-giants ~~or remove them from this PR.~~
- [X] Move missions from Remnant Cognizance.txt to Remnant Missions.txt. Planning on doing that once the bugs are worked out. I've been working on it in a separate file for ease of navigation.

## Save File
This save file can be used to play through the new mission content:
[google drive link.](https://drive.google.com/file/d/1qcJmDTS_3GdAAgo7B887bu_Q0Xh-1nYV/view?usp=sharing)

## PR Checklist
 - [X] I updated the copyright attributions, or decline to claim copyright of any assets produced or modified
 - [x] I uploaded the necessary image, blend, and texture [assets here.](https://drive.google.com/file/d/1Z2tMSV9so3sh-R6pfyRHY-PQ5pnmlvTn/view?usp=sharing)
 - [X] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the [`@2x` versions of these art assets.](https://github.com/endless-sky/endless-sky-high-dpi/pull/129)
  
[...]

PR #1524, opened 2016-09-05
Animated Stars

Amazinite:
I created this and got some good feedback from a couple of people who tried it out, so I thought I might as well just PR it to try.

As stated, it animates all the stars, giving them the look of their upper layers moving around. As to keep with MZ's wishes in #1517, the frame rates are low so that the animations aren't too drastic/visually distracting.

@2x images at the ready.

PR #4960, opened 2020-03-29
Brown dwarves

ravenshining:
**Content (Artwork)**

## Summary

This PR adds some brown dwarves.  IRL, brown dwarves are a sort of grey area between planets and stars, often found either orbiting stars or sometimes at the centre of their own planetary systems.  They tend to be just a tad larger than Jupiter, and are very dim or emit no light at all.

While working on a larger star-related project, I saw #4943 , which included a single brown dwarf sprite.  After discussion there I decided to take on making high-resolution brown dwarf sprites.

Because brown dwarves may potentially be visually interesting, and because they would be very tiny as compared to stars, I have rendered these at planet-scale and placed the sprites in the planet folder.  

I have included three different styles of brown dwarf, each rendered both in a self-illuminated form for use as the centre of a planetary system, and in a directionally-illuminated form for use in orbit around a proper star.

Also included are landing messages and solar power/wind stats.  I have adhered to the general trend of declining solar power and increasing solar wind, but:

- Should brown dwarves buck this trend, producing less fuel AND less power?
- Should power and/or fuel be removed from the "orbital" versions of the sprites?

Not yet included are map placements.

## PR Checklist
 - [x] I decline to claim copyright of any assets produced or modified.  Type L and Y were created from scratch, Type T is a modification of a public domain image from NASA.
 - [x] I uploaded the necessary GIMP assets here: https://drive.google.com/open?id=1ZaO9TgDoU84Dg4tiKfLZg24of3y-M1Jn
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/122
  
 -----------------------
**Feature:** This PR implements a feature request detailed and discussed in PR #4943 

## UI Screenshots

dwarfvesdone
dwarfvesdone2

## Testing Done

I replaced various planets and/or stars in existing systems, flew around them, attempted to land on them to trigger landing messages.

## Performance Impact
N/A


Zitchas:
These look really good. I'm looking forward to using them. Considering switching one or two planets in the Remnant systems out for some of these; and possibly using them elsewhere as well.

MasterOfGrey:
They should definitely buck the trend and provide both less fuel and less solar, but I don't think there's any need to remove the contribution they grant as an orbital version.

Amazinite:
Definitely an interesting idea, and the artwork looks great.

I'm wondering if they should be this size, though. Red dwarves in game are 55x55 (m8) to 65x65 (m0). The largest star is 150x150 (b5). Then planets range from 60x60 (moons) to 450x450 (gas giants, excluding those with rings that have bigger dimensions). In so far as brown dwarves aren't stars, it makes sense to me that their size would be in the 400x400 range, but in so far as they're used as system centers, I feel like having them be closer to the size of red dwarves makes more sense. (I don't think that having small and large versions for system centers and orbital objects respectively is the right approach, though.)

(Sidenote: 419x419 irks me. Not 420x420?)
(Sidenote^2: Can you animate these? Resurrecting #1524 is on my mind.)

Lineth:
Honestly, an art upgrade for the stars in the game would would be cool, and this would be a good starting point, art-wise.

ravenshining:
> I'm wondering if they should be this size, though. Red dwarves in game are 55x55 (m8) to 65x65 (m0). The largest star is 150x150 (b5). Then planets range from 60x60 (moons) to 450x450 (gas giants, excluding those with rings that have bigger dimensions). In so far as brown dwarves aren't stars, it makes sense to me that their size would be in the 400x400 range, but in so far as they're used as system centers, I feel like having them be closer to the size of red dwarves makes more sense. (I don't think that having small and large versions for system centers and orbital objects respectively is the right approach, though.)

That's the trouble with having different scales for different things :-)  What do you do with something that can be both or neither?

What it came down to, in my opinion, is what looks good and isn't confusing.  Proper stars have a distinctive appearance, and these look like glowing planets more than they look like dim stars.  If I make them small like the M8, they look more like tiny moons than tiny stars.

I've seen others demoing some artwork for adding rouge planets to ES.  In those systems, you'd have a planet at the system centre with no star, which really isn't all that different from having a big, softly-glowing gas giant with some moons as a system centre.

Then, there's the whole spaceflight aspect.  Flying a spaceship close to a star is a dangerous proposition!  ES drawing stars small makes them look far away.  Now, while brown dwarves do have high gravity, the cooler classes (that is, the ones I have here) would be relatively safe to fly up to, or possibly even enter the upper atmosphere of.  Scientists have found brown dwarves as cool as the human body, although those ones do not emit light at all.

The radar subtly reinforces this sense of scale, drawing far-away stars as yellow circles, and planets as grey circles. Because I placed the sprites in the planets folder, they show up on radar as grey circles instead of yellow circles, to give that nudge that these are "closer" to you than stars.

> > Should brown dwarves buck this trend, producing less fuel AND less power?
> 
> Yes.
> 
> > Should power and/or fuel be removed from the "orbital" versions of the sprites?
> 
> No.
> 
> (Sidenote: 419x419 irks me. Not 420x420?)

Okay, sure.  I'll make those tweaks.

> (Sidenote^2: Can you animate these? Resurrecting #1524 is on my mind.)

I could, ironically, animate the purple one - the one that's coldest and least likely to be animated - because I have full-size separate layers for the burning core and upper clouds.  Unfortunately those layers got cropped for the red one, and the brown one doesn't have separate cloud layers at all because it's a NASA image that I took, de-shaded, and applied ES standard shading to.  I'd have to reconstruct those two from scratch in order to animate them.

TLDR: yes. it'd be a lot more work than I want to get into right now, but potentially they could be animated in the future.

ravenshining:
Okay, sprites are more easily machine digestible sizes now, and I've updated the solar wind so they all produce less fuel than the largest (least windy) star.

Following up on Zitchas's suggestion, I took a tour of the Ember Wastes and took down some notes on candidate systems for brown dwarves.  I'll wait for some feedback before pushing changes to the map:

**Systems where a brown dwarf orbiting the main star might be appropriate:**
- Delia:  Perhaps the presence of a brown dwarf on the periphery of the system made the location ideal for a Dyson sphere for some reason?  Or conversely, maybe some planets got knocked together in... whatever happened to the Ember Wastes, creating the brown dwarf, whose gravity destabilised the sphere?
- Lire, Segesta, Arculus: The inhabited world is outside the star's habitable zone, maybe a brown dwarf helps make it possible
- Parca:  Near the core, has three gas giants.  Core systems should have lots of stars.

**Systems where a brown dwarf system centre might be appropriate:**
- Perfica, Paeli:  Otherwise boring systems.  The Ember Wastes shouldn't have boring systems.

Amazinite:
> Okay, sprites are more easily machine digestible sizes now, and I've updated the solar wind so they all produce less fuel than the largest (least windy) star.

It's not a machine digestible thing, it's a "literally every other star and planet has a size ending in 0 or 5" thing.

ravenshining:
Okay, sure, the sprites are all 420x420 now - and 840x840 for the high dpi versions.

Taking that thumbs-up as an approval for now I made the noted changes to the map, as well as replacing Feraticus's star with a brown dwarf.  Feraticus is another boring system, and with it now all classes are represented.

A couple notes while gathering the below screenshots
- The first time I jumped into a system with a brown dwarf as the system centre, I did a bit of a double take, wondering where the system's star was.  That I surprised myself I think is a good thing.
- Confirmed that solar wind still works in systems with only a brown dwarf, and that escorts that are out of fuel will still gather around the system centre.

Delia, zoomed out:
delia0

Delia, zoomed in on its Y-type brown dwarf:
delia1

Lire, zoomed out:
lire0

Lire, zoomed in on its L-type brown dwarf:
lire1

Segesta, zoomed out:
segesta0

Segesta, zoomed in on its T-type brown dwarf:
segesta1

Arculus, zoomed out:
arculus0

Parca, zoomed out and in portrait mode because the RNG wanted to put the planets north-south:
pacra0

Perfica, zoomed out:
perfica0

Pefica, zoomed in on its lone Y-type brown dwarf:
perfica1

Paeli, zoomed out:
paeli0

Paeli, zoomed in on its lone T-type brown dwarf:
paeli1

Feraticus, zoomed out:
feraticus0

Feraticus, zoomed in on its lone L-type brown dwarf:
feraticus1

ravenshining:
I *really* like the way it looks when a moon is in front of a gas giant, but okay :-(  I'll have time to fool with orbits tomorrow.

Zitchas:
I, too, really like the way they look when they overlap. A lot of really good imagery out there (especially artist's renditions of moons and extrasolar planets) superimposes moons over/above planets. 

For instance, the images above of Segesta and Paeli I think in particular look very good. 

Perhaps overlapping shouldn't be a super-common thing; but having a few systems (especially in the Ember Waste, where one can argue things might not be following the planar norm) where such overlaps would be very good to have.

As far as that goes, it is very common to have moons who don't orbit their planets in the same plane as their planet's orbit around the star. ES is kind of an aberration in that every orbital system is perfectly planar (and perfectly circular too, but that's not really an issue here)

Amazinite:
>As far as that goes, it is very common to have moons who don't orbit their planets in the same plane as their planet's orbit around the star.

Looking at the major moons in the solar system, their inclinations are actually all relatively close to the orbital plane; the "Moons of \<planet>" Wikipedia pages seem to use inconsistent points of reference (some use the equator of the planet being orbited, some use the orbital plane, and I'm not doing that math right now), but basically all the major moons in the solar system orbit at low angles, so you wouldn't get "overlapping" if looking at everything from above. The only exception is Uranus with its freak 97 degree axial tilt, meaning that its major moons that all have inclinations less than a degree from Uranus' equator would overlap with the planet if looking from above relative to the orbital plane.

So the case is that it's common for the tilt from the orbital plane for planets to be low, for their axial tilts to be low, and for their moons to roughly stick to the equator of the planet, therefore making it uncommon for moons to be orbiting nearly perpendicular to the orbital plane, i.e. making it uncommon for moons to have an overlapping appearance in game.

And after looking through all those Wikipedia pages to find that out, it comes to mind that even if it was the case that moons can occasionally orbit perpendicular to the orbital plane, that's not what the moons would be doing visually in ES, as the moon is never going to go behind the planet, as it just goes in circles around the center of the planet.

ravenshining:
Okay, as requested, the moons and/or planets in Lire, Perfica, Paeli, and Segesta no longer touch or overlap the new sprites.

Also I had a look at #4871 to make sure there's no conflict; both our PRs modify the Arculus system but it should be fine as I just added a dwarf to the system outskirts and did not replace any of the planets Zitchas added descriptions for.

Zitchas:
That's kind of the amusing factor: In ES the moons don't actually move; and given that the pilot may want to access them, it can be assumed that they are changing their angle of approach to pick favourable angles that always have the moon in a convenient visibility. Thus changing positions from day to day is simply seeing different angles. The only time that it would be obvious that the moon is circling above the planet is when viewing it in the map editor, which I don't think is really relevant for in-game perception.

PR #4961, opened 2020-03-29
More Shields Kestrel +10% heat dissipation

finite-galaxy:
Prior to #3976, the More Shields Kestrel was by far the most popular option. <s>According to #4906, it is now by far the least chosen. That means it deserves a little extra (though nothing too major).</s>
This patch gives the More Shields Kestrel +10% heat dissipation, making it somewhat less vulnerable to heat damage weapons, consistent with it being the defensive variant.

### DIFF
--- a/data/human/kestrel.txt
+++ b/data/human/kestrel.txt
@@ -336,6 +336,7 @@ ship "Kestrel" "Kestrel (More Shields)"
 	add attributes
 		"shields" 3000
 		"hull" 1500
+		"heat dissipation" .06
 
 ship "Kestrel" "Kestrel (More Weapons)"
 	add attributes


beccabunny:
Where does #4906 say that it is by far the least chosen exactly?

finite-galaxy:
It does not, my mistake, I skimmed the pull request without reading carefully, I apologize.
Nonetheless, I do believe this tiny tweak will make the More Shields variant a more interesting choice without disrupting balance.

Zitchas:
I'm not opposed to it, although I suspect that the recent nerf to the gun ports might balance things out a bit anyway.

Amazinite:
Is better heat dissipation how we want to buff the shield Kestrel?

The main idea behind the weapons Kestrel is dealing more damage; it has more weapon space and more weapon ports.
The main idea behind the engines Kestrel is going faster; it has more engine space and less drag.
The main idea behind the shields Kestrel is survivability; it has more HP and... nothing else right now. Heat dissipation though doesn't really seem to me to be something that boost the survivability of the Kestrel. It does make you a little more resilient against heat damage, but the main purpose behind heat dissipation is dictating the max idle heat of a ship. It's more of an outfitting stat than a combat stat. 
If any extra bonus to the Kestrel makes sense to me, it'd either be in-built shield and hull repair, or to make use of the resistance attributes.

finite-galaxy:
> It's more of an outfitting stat than a combat stat. If any extra bonus to the Kestrel makes sense to me, it'd either be in-built shield and hull repair, or to make use of the resistance attributes.

If I recall correctly, there is no lore indication Tarazed encountered and studied alien species other than the Quarg. In-built shield regenerators are a characteristic of the Remnant. Hull repair does not exist in human space. The three resistance stats (disrupting, ion, slowing) counter specific alien weapons. 
However, Plasma Cannons are not uncommon in human space, and the stat to resist heat damage *is* heat dissipation.

Amazinite:
The Kestrel is very much an easter egg ship though, so I see little issue in "breaking lore" to show the ship some love. The bonus doesn't even need to be extra shield *and* hull repair, though. Could drop hull repair, and only give in-built shield generation, described as quite literally being an in-built shield generator, difference from the Remnant's in-built repair being a characteristic of the material they make their ships out of.

Zitchas:
Well, as far as that goes, how hot a ship runs (in theory) has an impact on how easily it is targeted by thermal guiddance missiles. So giving it a little better head dissipation might have actual combat consequences.

Beyond what Finite mentions about Tarazed not having contact with any aliens beyond the Quarg; and if they did, the most likely is the Hai due to already existing (albeit secretive) trade existing there. And they don't have built-in regen either.

Beyond that, though, the Kestrel is based on a hallucination of the Escape Velocity universe. What change could we make that renders it more in-line with that world?
- Increased heat dissipation: In EV, EVO, and EVN the flare was a commonly equipped outfit that was quite effective at dealing with thermally guided missiles, albeit only when you were running away. (They dropped like bombs and lost all their momentum really quickly, and only distracted thermal missiles when they were between the missile and its target. Awesome for running away, almost useless when attacking). Not to mention, that heat wasn't actually a thing that was tracked, either. So increasing the Shield variant's thermal dissipation would theoretically reduce the Kestrel's vulnerability to a missile type that wasn't very effective in the universe it comes from; and reduces its vulnerability to weapons that deal heat damage as well.  Both are good impacts in line with its defensive focus.
- Fighter bays: In both EV and EVO the Kestrel had a fighter bay that carried two of the heaviest fighters available to civilians in the game, the Lightning. (which itself is tough enough that it is classed in-game as an interceptor). Now, I didn't use fighters much in EV, and I don't use them a lot in ES either; but I did enjoy show-boating with them and launching them to fly alongside me when I felt like showing off; As such, the Kestrel has always been - in my mind - a ship that has at least two fighters onboard. Given the current state of fighters, giving the Kestrel a pair of bays isn't going to be a hightly sought after feature, but it could be a nice perk for distracting enemies while the Kestrel gets away, which does fit the survival theme.

Amazinite:
> Beyond that, though, the Kestrel is based on a hallucination of the Escape Velocity universe. What change could we make that renders it more in-line with that world?

I feel like any change to make the Kestrel more in-line with its appearances in EV should be done to all three variants, not just one of the three.

finite-galaxy:
> (And I will say, the Lightning definitely looks like a Tarazed ship.)

(As does Endless Sky's Raven.)

waterhouse:
I want to correct some EV lore here:
- I think decoy flares only existed in EV, not Override or Nova.
- The Kestrel didn't exist in EV Override; reasonably close analogues would be the Crescent Warship and the Zidara.
- The EVN Kestrel had a pirate viper bay (I had to look this up)—I don't think Lightning bays existed in the base game.

IMO, the EVN Kestrel was overpriced and underwhelming—when you come to it in a Vell-os Javelin, Thunderforge, Unrelenting, or Polaris Raven, its shields/armor are _somewhat_ better than your current ship (unless it's a Nil'kimas Raven), its outfits are mediocre, and it costs 50M (possibly discounted) compared to the 10-15M for each human faction's biggest capital ships; and if you can't afford it when it's offered, it won't be offered again, without a plugin or starting a new game; plus you get it after the storyline is over and all that's left is dominating planets.  I think I bought it only once.

By contrast, the EV Kestrel was the most powerful civilian warship (but weaker than the Rebel and Confed capital ships, which you could only access after finishing their storyline), and, at 10M, cost 4x as much as the second-best Corvette; you would fight pirate Kestrels now and then, and you'd probably buy your own Kestrel sometime during the main mission string, likely using it to defeat the Alien Cruiser at the end (though I did do that in a Corvette sometimes).  It was teased, it was cool, it was tantalizingly expensive, and it was worth it.

I think the EV Kestrel is the one we're primarily imitating thus far: Endless Sky's Kestrel is available before entering any major mission strings, and is arguably the strongest civilian warship, but weaker than the Carrier (but the margin is much smaller than EV Kestrel vs Confed Cruiser).  Frankly, in terms of being the strongest pirate ship and what you might aspire to, the Leviathan occupies much of the niche that the EV Kestrel did (and bears some physical resemblance), and the Bactrian to a certain extent as well.  Still, given what we have...

I definitely like the fighters idea.  Two fighters, to match the original.  Unfortunately, I don't think Endless Sky ships are sold with fighters, but if there are any NPC Kestrels, they should have fighters resembling Lightnings: their armament was heavy rockets, missiles, javelin rockets, and laser cannons; the Berserker and Hawk would be good candidates, and are shaped similar to Lightnings too.  But at the very least, giving it two fighter slots seems like a no-brainer.

The EVN Kestrel had better armor than everything except Auroran capital ships, and better shields than everything except the Polaris Raven and Vell-os Javelin.  It also regenerated shields faster than everything (though Polaris Organic Shield outfits give a bonus larger than its innate regen), _and_ regenerated _armor_, faster than even Polaris organic ships (though the Zephyr came close).  And it was more maneuverable than the capital ships (comparable to the Javelin).  I think Endless Sky's Kestrel is already maneuverable, but this might give justification for the built-in regen stuff—though if we did go for that, I would put it in the "More Shields" variant.  If we _did_ that, then, to balance the other choices, I'd be inclined to add extra space (not just extra weapon/engine space) to them.

Some references:
http://web.archive.org/web/20080921060304/http://www.evula.com/survival_guide/ev-ships/warships.html
http://web.archive.org/web/20081003031816/http://www.evula.com/survival_guide/evn-ships/fighter_warship.html
http://web.archive.org/web/20081002031455/http://www.evula.com/survival_guide/evn-ships/polaris.html

Zitchas:
1) I'll disagree on that, but short of someone firing up a copy of the games, that'll just have to be a point we disagree on. None of the wikis I've seen look anything close to being complete, so the absence of something from any of them isn't really conclusive. Even the archive of evula's just lists a "coming soon" in place of EVN's outfit section.

2) I am certain that the Kestrel was in EVO, but it may have simply been an author ship. Definitely not a ship seen in-game with any frequency.

3) Yes, I think it is fair to say that the original Escape Velocity is the version of Kestrel we are aiming at. It is the original one, and it is - in terms of how it fit into the game world - the best one. 

Anyway, flares remain on my wish list for many reasons - not the least of which is missile balance and merchant survivability. But that's not the topic of this thread.

Back to the topic at hand, I would certainly be in favour of giving the Kestrel a pair of fighter bays as a general boost; and giving the shield variant the heat dissipation boost. It isn't a huge thing, but it does give a little more flexibility in outfitting, at least. Which could be a nice perk for the shields variant.

Zitchas:
After the live-stream discussion  yesterday, I'd just like to propose an additional change for this particular PR:

That the Shield Kestrel get the extra heat dissipation as proposed here, but also get some scan resistance. Something along the lines of `"scan interference" 1` (which is the equivalent of two "Interference Plating". That's a boost of 8 outfit space, which doesn't seem unreasonable.

My justification is that while scan interference isn't directly a "make this ship tougher" stat, it does fit in nicely with the idea that picking the shield variants leads Tarazed to put more work into the hull & shield system. Integrating some more visionary inspiration into tweaks of these two systems could very reasonably lead to some interference on the scans.

finite-galaxy:
While I'm not opposed to the idea per se, I do have a few concerns:
* How to convey that information? Scan interference is one of those ship stats that isn't displayed. This is also why I *didn't* propose giving the Shields variant radar jamming (which would make sense for a defensive variant).
* Why only the Shields variant, and not the other Kestrels? If any, I'd expect the super-secret Unknown Ship Type to have scan interference.
* If the Kestrel would have scan interference, then why not other Tarazed ships as well? E.g. the Blackbird has similar colours, hence probably made from the same materials.

PR #4446, opened 2019-07-29
Add mission about Southbound ship designer.

petervdmeer:
This is a first mission in a string of missions about ship designers that I had come up with. The mission can be played stand-alone, so sending this as a separate PR to see if this is suitable for inclusion in the main game.

Signed-off-by: Peter van der Meer <peter.vd.meer@gmail.com>

PR #4968, opened 2020-04-01
Referenced the Hauler IV in the "Hauler VI cargo prototype" missions

Anarchist2:
## Fix Details
~~In the "Hauler VI cargo prototype" missions, the Hauler VI is referred to. This is an odd choice in my opinion, as VI means 6, not 4, and the Hauler IV (4) would be the logical step up. This PR changes all mentions of "Hauler VI" to "Hauler IV". An alternative solution could be to keep it as the Hauler VI, but add a bit of dialogue referring to this oddity ("Why VI and not IV?").~~

In response to the feedback from Zitchas, Derpy and petervdmeer, I've changed the PR to add a new choice that covers the apparent oddness of moving from the Hauler III straight to the Hauler VI.

### DIFF
--- a/data/human/transport missions.txt
+++ b/data/human/transport missions.txt
@@ -3472,7 +3472,7 @@ mission "Inheritance Redirection"
 
 
 
-mission "Hauler VI cargo prototype 1"
+mission "Hauler IV cargo prototype 1"
 	name "Escort Hauler prototype to formal dinner"
 	description "Escort the <npc> to <destination> to deliver sushi for a formal dinner."
 	minor
@@ -3494,7 +3494,7 @@ mission "Hauler VI cargo prototype 1"
 					defer
 			`	She smiles and replies, "Well, I actually don't know you, but I just read your name on one of the cargo manifests. And my experience is that this greeting works best if you need something from a spaceship captain."`
 			label "hello too"
-			`	She says, "My name is Veronica Oxygenpocket, and I'm the lead for a team that's designing the new Hauler VI transport ship series at Southbound Shipyards."`
+			`	She says, "My name is Veronica Oxygenpocket, and I'm the lead for a team that's designing the new Hauler IV transport ship series at Southbound Shipyards."`
 			choice
 				`	"'Oxygenpocket' is an uncommon name."`
 				`	"And what help do you need from me?"`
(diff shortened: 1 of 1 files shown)


Zitchas:
Well, it's a prototype for a vehicle. Given how these days every manufacturer seems to have a different numbering pattern with some going up by 2s or by 10s or by 100s, or some sticking with the *same* number and going up by some undecipherable letter somewhere in the name instead, I just figured that for whatever reason, the manufacturer didn't like 4 & 5. Or maybe they were limited edition models for some reason. Test beds, prototypes, or special production runs for government or a private buyer.

As such, I think it might be better to create a hauler 4 & 5 variant and add very rare missions that ask the player to escort them somewhere. "A rich collector asks you to escort their one of a kind hauler from X to Y for a spaceship show" or something. Extra conversation options in such a mission could explain why they aren't generally for sale. (and adding new missions to the game to explain this - especially rare one-off missions - would be much more beneficial than simply tweaking the text on this mission)

Amazinite:
Seems to me like this is something @petervdmeer should decide since these are his missions. (#4446)

petervdmeer:
Here is the picture I had photoshopped before writing the Hauler VI prototype missions. I indeed used Hauler VI as name because of the 6 cargo-pods used.

hauler vi

Adding some text to explain how the Hauler VI looks and that the name is based on the amount of pods in the ship sounds like a good solution.

Adding a proper Hauler VI (and maybe also a Hauler IV) ship model could also be nice. Then we could do the mission with a proper Hauler VI. (It should probably be single cockpit, at least for the production version. The prototype might be dual cockpit since it is a prototype.)

data/human/transport missions.txt
```diff
@@ -3505,12 +3505,22 @@ mission "Hauler VI cargo prototype 1"
...
 				`	"Isn't it a bit risky to transport food using a refrigeration system that hasn't been tested yet?"`
+					goto refrigerate
 				`	"Am I also invited for the dinner?"`
 					goto invited
 				`	"Why do you need me to escort the ship?"`
 					goto internal
+				`	"Shouldn't you be making the Hauler IV first?"`
 				`	"Sure, I'll help you."`
 					accept
+			`	Veronica softly chuckles. "That would be the obvious choice, wouldn't it be? But this isn't just a case of slapping another cargo pod between the cockpit and engines, we've modified the design to have the optimal amount of cargo pods without comprimising the integrity of the vessel. And besides, our marketing department says that the people who are willing to by a Hauler IV are just as willing to buy larger Haulers."`
```
Amazinite, line 3516:
```suggestion
			`	Veronica softly chuckles. "That would be the obvious choice, wouldn't it be? But this isn't just a case of slapping another cargo pod between the cockpit and engines. We've modified the design to have the optimal amount of cargo pods without compromising the integrity of the vessel, or making it difficult to build in our existing shipyards. And besides, our marketing department has found that captains willing to buy a Hauler IV are just as willing to buy larger Haulers."`
```

PR #4975, opened 2020-04-04
feat: Publish CD artifacts as pre-release

MCOfficer:
**Feature:** This PR implements the feature request detailed and discussed in issue #4829 

## Feature Details
This PR expands the CD job that runs on every release. It will now
- Create a `continuos` tag at, or move the existing one to, the current commit
- Create a new pre-release at the `continuous` tag, if it doesn't exist already
- Upload the artifacts produced during the CD job to said pre-release, replacing any earlier artifacts

Example Release:
https://github.com/MCOfficer/endless-sky/releases/tag/continuous

Note that Pre-Releases are never marked or linked to as "latest" release, though they do appear first in the list if they're newer.

## Testing Done
Tested on my fork, see above. The build process itself is the same as in the current CD workflow.


.github/workflows/cd.yaml
```diff
@@ -32,35 +69,72 @@ jobs:
...
+          echo "::set-env name=GOPATH::$(go env GOPATH)"
+          echo "::add-path::$(go env GOPATH)/bin"
       - name: Install dependencies
         run: |
           sudo rm /etc/apt/sources.list.d/* && sudo dpkg --clear-avail # Speed up installation and get rid of unwanted lists
           sudo apt-get update
           sudo apt-get install -y --no-install-recommends libsdl2-dev libpng-dev libjpeg-turbo8-dev libopenal-dev libmad0-dev libglew-dev libgl1-mesa-dev libegl1-mesa-dev libgles2-mesa-dev scons
+      - name: Adjust version strings
+        run: |
+          perl -p -i -e 's/(Endless Sky) [\d.]+/$1 ${{ github.sha }}/ig' source/main.cpp
```
tehhowch, line 103: Does it make sense to create a utility script for this?

MCOfficer: yes, will do

utils/cd_version_strings.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+# Adjusts the version strings in main.cpp, credits.txt and EndlessSky-Info.plist for a nightly build.
+
+perl -p -i -e 's/(Endless Sky) [\d.]+/$1 ${{ github.sha }}/ig' source/main.cpp
```
tehhowch, line 5: Can we make this script just "set-version" rather than "cd-version-strings" and take an argument, rather than requiring it to be run only from the actions environment (where `${{ }}` does something)?

Then if called with the current version, running `git diff` would result in no changes.

(The line which updates the credits string would probably not be something included, as that's very specific to the CD workflows. A script which just sets versions, and a script which updates credits strings and calls the set version would be viable, too.)

My reasoning here is that - since we will export utils, they should be generally useful too

MCOfficer: The simplest solution imo would be to just call git instead of the convenient action variables:
```bash
perl -p -i -e "s/(Endless Sky) [\d.]+/\$1 $(git rev-parse --verify HEAD)/ig" source/main.cpp
```

> My reasoning here is that - since we will export utils, they should be generally useful too

I don't disagree, but we shouldn't be making up usecases. I can't think of a case where this script is used other than some kind of bleeding-edge build. The main reason that this is a utils script is because it's used 4 times in the actions job.

.github/workflows/cd.yaml
```diff
@@ -32,45 +66,76 @@ jobs:
...
 
   windows_win64:
     runs-on: windows-latest
+    needs: setup
     env:
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       DIR_MINGW64: C:\ProgramData\chocolatey\lib\mingw\tools\install\mingw64\x86_64-w64-mingw32
       DIR_ESLIB: .\dev64
-      OUTPUT: EndlessSky-win64-${{ github.sha }}.zip
+      OUTPUT: EndlessSky-win64-continuos.zip
```
Zitchas, line 124: Shouldn't this be `      OUTPUT: EndlessSky-win64-continuous.zip` ? (One missing "u" in "continuous")

MCOfficer: whoops. good spot.

Zitchas: You've also got the same spelling mistake on your first post. I was just going to comment on that, then decided to take a look through the rest just in case.

utils/cd_version_strings.sh
tehhowch: Don't consider it "making up" usecases, more "enabling". Any fork or contributor or collaborator (or non-GitHub CI) that pulls in this script would then be able to do something as simple as `./utils/set-version "my cool version"` (provided of course that `perl` is available in the executing context).

MCOfficer: I see what you're going for. I'll also remove the change to the .plist file, since i just noticed no sha256 can conform to apple's [`CFBundleVersion` spec](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102364).

tehhowch: Can we just append "a" or "d" to the CFBundleShortVersion and set that as the CFBundleVersion, as a part of the cd's script?

There's one other change that i think the set-version script should do, and that's update the man file with the date and version:
```
.TH endless\-sky 6 "24 Feb 2020" "ver. 0.9.11" "Endless Sky"
```

MCOfficer: The man file shouldn't be a problem. 

About XCode, some research: [`CFBundleShortVersionString`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) is the actual version, e.g `0.9.11`, and needs to be in `x.y.z` format, no exceptions (not even letters). [`CFBundleVersion`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) also only allows integers, but is intended for build numbers, so `1234` is valid.

We can abuse that by not modifying `CDBundleShortVersionString`, but incrementing `CDBundleVersion` (e.g. by setting it to the actions run ID, or the amount of commits).

MCOfficer:
I reenabled builds to trigger on my branch to test any changes, so please don't merge this until it's changed back (`cd.yaml`, L6).

tehhowch:
I'm guessing the "7 days ago" is because we're just moving the release rather than re-creating it: 
image

MCOfficer:
> I'm guessing the "7 days ago" is because we're just moving the release rather than re-creating it:
> image

correct. If you insist, i can put the actual build date in the description. Otherwise this lgtm.

PR #4977, opened 2020-04-06
Fix: Reduce "take-off and reload" instances in the gascraft mission while making the Archon Deadlier

Zitchas:
Update: This PR has expanded a bit, as it became increasingly obvious that one of the main problems with increasing the Archon's potential lethality was its impact on the gascraft missions. As such, this PR present a two-fold update to the Archon's weaponry. 

In the process of increasing the Archon's combat capabilities, I was able to fix a long-standing problem I have had with the gascraft missions: While the approach part I have always considered to be a fun challenge, the departure was miserable. Regardless of player skill level, the departure has always boiled down to a russian-roulette situation revolving around where exactly the Archon will be relative to the player when they take off. If they are too close (which happens frequently), the player is dead almost before they can realize what is happening, and it is time to reload. In some instances, I have had to reload multiple times in a row just to leave the system. These changes to the Drak Draining field fix this.

##Changes:
- The Drak Anti  Matter Canon now has 300 disruption damage. This means the first hit will be normal damage, and the subsequent hits will punch significant amounts of damage straight through to the hull. 
- The Drak Draining Field now acts as both a drain and a forcefield. It creates an "aura" around the Archon that provides a light amount of shield damage, disruption, and ionization, as well as a significant kick. This kick is sufficient that a stock dreadnought or Albatross has to take a run at it in order to close with it. 

##Gameplay effects:
- Firstly, as a result of the aura effect, if the gascraft is within the Archon's turret range, it will be kicked by the Archon's aura just before the AMC bolt can connect with it. In testing, this has resulted in the "death" side of the russian-roulette spin being replaced with a "just barely managed to escape" situation. It still requires dodging and whatnot to actually escape, but it has virtually eliminated the "insta-kill" that often takes place before the player can actually take action.
- Secondly, the Archon is capable of taking down significantly more powerful enemies than before, and putting a heavier strain on their regenerative systems as they will have to contend with both shield and hull damage simultaneously.
- Thirdly, it has rendered it much, much harder to effectively swarm the Archon. The range on the aura keeps all ships that lack the thrust to counter the aura outside of most gun ranges, which means that the Archon can pick them off at its leisure. It also looks much more impressive and worthy of the "untouchable" allure of what a T4+ ship should be like. 
- Fourthly, it *increases* the length of combat against low tier ships. As befitting a species that is reluctant to kill, the aura keeps ships that aren't a threat to the Archon beyond the range of its turret. As such, against fleets of small ships it will leisurely swat them with its AMC, presumably to allow them the chance to run away. It is only against ships designed to close with the Archon that it will bring its turret to bear. This opens the door for the turret to be improved in a variety of ways (so long as such changes do not affect its velocity or range) as needed in the future. It also means that the Archon could gain other weapons, so long as their velocity does not exceed that of the turret.


data/drak/drak.txt
```diff
@@ -143,6 +143,8 @@ outfit "Drak Antimatter Fragment"
 		"inaccuracy" 1
 		"lifetime" 499
 		"hit force" 1000
+		"disruption damage" 50
+		"heat damage" 250
```
Amazinite, line 147: This heat damage is hardly anything. A plasma cannon has nearly five times this DPS.

Arachi-Lover:
if you're only hit by one of the fragments, the Gascraft is not 1HKO'd. the damage is pretty high, could see some buffs still, but it's pretty odd how it takes about the same time for a Mereti Fleet to disable a sitting duck 512 as it'd take an Archon, all because the AMC has to go through shields first. If it's "the Drak's ultimate weapon." , just a bunch of shots that deal high damage and nothing else is kinda boring

Amazinite:
I checked the code after seeing this first posted, and it's important to note that disruption is applied after shield and hull damage, meaning that it's only subsequent hits that will benefit from the disruption damage. The first shot will do the same damage as always.

data/drak/drak.txt
Zitchas: OK. I'll increase it.

Zitchas:
So the 50 disruption won't do anything on a first hit... But it significantly increase the chance that the next shot is a knock out?

beccabunny:
Why not use piercing instead?

Arachi-Lover:
piercing would make the gascraft be oneshot even by a single fragment

Amazinite:
I prefer the idea of using disruption. Explanation: antimatter mumbo jumbo reacting with the matter of your shields. Result: the more hits you take, the more the antimatter cannon is going to be damaging you.

Arachi-Lover:
yah I think the heat damage is out of place tbh

Zitchas:
My idea here is to not significantly impact the gascraft/archon balance; but increase the lethality of the archon versus other threats. So giving it disruption works really well for that; as it will increase the damage of the subsequent shots without affecting the first shot. And, well, if a gascraft gets hit with two shots, it's toast anyway, regardless.

Re: The heat. I'll have to check numbers (specifically, what the running heat of a gascraft is), but the intent is to be high enough that it'll make a gascraft pilot sweat but shouldn't actually fry them. Versus larger and more long-term threats, it is more of an after effect of that much stuff reacting nearby - this isn't intended to make the AMC into a plasma/heat weapon; it is more intended to put a little extra strain on ship systems - especially those who run with active cooling and high heat = high power draw.

data/drak/drak.txt
```diff
@@ -143,6 +143,8 @@ outfit "Drak Antimatter Fragment"
 		"inaccuracy" 1
 		"lifetime" 499
 		"hit force" 1000
+		"disruption damage" 300
```
Amazinite, line 146: *Nervous sweating.*

50 disruption damage per fragment meant that if you took a full blast of the AMC, then you'd take about 18.6% piercing next time you get hit (assuming no protection/resistances). At 300 disruption damage per fragment, the second blast will have about 57.9% piercing.

Zitchas: Sounds like I guestimated my math at about the right spot, then. Archons are supposed to be a *threat*, right?

Zitchas: I'm also assuming that anything down the road that is actually supposed to threaten these things probably has a) too much hull to care, and/or b) disruption resistance.

But as far as that goes, nothing short of the T3+ Pug ships should really be having any kind of chance against it. In other words, nothing else currently in-game.

Amazinite: Just some more math:
* 50 disruption/fragment = about 21.3% piercing at most for a prolonged fight where you're taking every hit.
* 300 disruption/fragment = about 61.9% piercing under the same circumstances.

Kinda wish we could have more of a ramp up over time; 50 disruption is only ramping from 18.6% piercing on the second shot to 21.3% piercing after a few shots, whereas 300 disruption is ramping from 57.9% on the second shot to 61.9% after a few shots.

Zitchas: That'd be neat. It would also be interesting if the Archon could channel disruption received into disruption dealt. The Pug make things "fair" by balancing their ships to only slightly higher than their enemies. The Drak make things "fair" by only ramping up their disruption when someone's using it on them...

NomadicVolcano:
I have three alternative suggestions:

1. Give the Archon another weapon that is deadly at close range and has multiple debuffs.
2. The Archon is ridiculously overpowered even for a Tier 4 heavy warship. Don't make that worse.
3. Add a second Archon-like creature the Drak use for "real threats." This one would be much bigger and have a lot more firepower.

If you want to go with option 3, I would love to make the creature. I'm about to do a bunch of indigenous creatures for the Whispering Void, and I would be honored to add the uber-archon to my "to do" list. It'll be about the size of a Korath Haven (see #4987) and designed after a jumping spider rather than a cockroach. I was going to use that artwork as one of the apex predators in the Whispering Void, but I could use it for an uber-archon instead.

Arachi-Lover:
no, the Archon is in fact underpowered for a Tier 4 heavy. sure it is tanky, but its only real weapon is the AMC, which is easy to dodge on any ship that isn't moving at the speed of a sailboat, and if we scale Quarg ships up, a Quarg Heavy would pretty much be taking fire of the equivalent of a pair of Quarg Torpedoes. giving the Archon more weapons removes the whole niche of the ship - though I'll agree stuff like the Drak Turret and Drak Draining Field need a buff, just making something like them that just does everything-debuffs ehhh nah, that's just downright annoying and OP for the sake of it, that is it excels at the entire alphabet when it comes to its arsenal, rather than it being OP in XYZ ways. I'm also against doing more Drak/Archon-ish ships until we finally decide the time to expand on them (which is not the time now since we really shouldn't even be thinking of doing anything Drak when the human storylines are still not complete, Wanderers still not done, Coalition not even started yet, Quarg still not reworked). for Archon excels at sniping from a distance for long periods of time, and having an arsenal that helps it stay away from ships that try and come closer, buffing its main weapon to actually feel like a main weapon is only proper. Giving the Drak new ships/weapons should have more thought than just "bigger and stronger" too, when we decide to do that.

Quantumshark:
It makes sense for the Archon to be weak offensively for its tier, because the Drak aren't a species that would devise weapons of war on par with their other advances - their goal is to promote peace across the galaxy, not to conquer it. The Archon's shield/hull total of 5 million is about twice what would be expected for an upper-end tier 4 based on extrapolating the jump from tier 1 to tier 2 (my figures ended up as 1.5 to 2.5 million), so again being offensively weak for a tier 4 heavy seems reasonable.

Arachi-Lover:
they're never mentioned to be intentionally weak in their offense, and I doubt that'd be the case. it is a balance oversight, if anything they really shouldn't be so "weak" now, since even the Pug mention them being there as one of the reasons they can't go defend the Wanderers in Korath space

Amazinite:
> [As more alien species are added, it's likely the Quarg and Drak will need to be made stronger to make room for more tiers in between.](https://github.com/endless-sky/endless-sky/wiki/Story-Ideas#advanced-alien-civilizations)

NomadicVolcano:
I'm planning on having a few Tier 3 (Quarg level) species in the SMC with a strict non-intervention policy. I need to know what that Tier will be. I was planning on balancing based on Arachi's thoughts (can't find the issue number right now, sorry).

NomadicVolcano:
The flaw in the Tier system is this, "A heavy warship of a civilization on one tier is a match for a light warship of the next tier up." It doesn't leave room for:

1. "Half Tiers:" A medium warship in one civilization is a match for a light warship in another. It might make sense to have "Half Tiers" be cannon.

2. Significant variation in the size and capabilities of a heavy warship. The idea of a "heavy warship" is like a "category 5 hurricane:" an open-ended category limited only by the laws of reality. A civilization could build something significantly larger than an ordinary heavy warship that will match medium warships one tier up.

3. Fleet sizes, as the Story Ideas page already points out. There are probably a lot more Quarg than Drak. Even if Drak have Tier 4 technology, the Quarg have something like Tier 5 firepower in comparison.

4. Advanced ship technology isn't the entire measure of a civilization's achievements.

Amazinite:
I think you're misunderstanding the idea of the tier system.
1. We already have half-tiers. See: [Coalition](https://github.com/endless-sky/endless-sky/issues/1519#issuecomment-244568916) and Remnant.
2. The measurement of "TX light = TX-1 heavy" doesn't preclude the idea of superheavies.
3. Tier isn't measuring total strength, but just the general strength of individual ships and outfits.
4. The tier system is in no way concrete, and is more-so an out-of-game explanation of the general strength of factions.

NomadicVolcano:
Regardless, I do need to know how strong the Quarg really are before I can make the other Tier 3 races.

If the definition of an Archon is, "it can defeat a fleet of any size," then I'll need to plan around that too.

Arachi-Lover:
though to be fair, I'll say that for a half-tier, Heliarchs' T2.5 feels more like a full tier under Quarg's T3 at times. 

Rebalancing the Quarg will be paramount to further the Coalition campaigns too, so yeah we do need to have them as the T3 baseline established in the somewhat-kinda-near future

NomadicVolcano:
> Rebalancing the Quarg will be paramount to further the Coalition campaigns too, so yeah we do need to have them as the T3 baseline established in the somewhat-kinda-near future

I fear I may need that to happen before I can do a proper job of the intergalactic Tier 3 faction. Otherwise we'll have two horribly unbalanced Tier 3 factions. 

How close are you to having the revised Quarg planned? I'd be happy to do some artwork and txt files based on your plans.

NomadicVolcano:
It'll be hard to judge whether the Tier 3 races are balanced with the Drak Tier 4 if the Drak have no light warships to test against.

Arachi-Lover:
the PR I'm referencing is 3665, and I've done stats for ships and outfits to try meet the needs of the checklist there for when it's time to do it, but I think it's more Derpy's thing, I've just done some stats and concept for weapons. the PR has artwork from 1010todd, and Becca also tried to do some Quarg ship redesigns a while ago, so I dunno which one would be used, like what style of ship would be used I mean, but I imagine we'd follow one of theirs.

in the most recent stream I asked Derpy if the rebalancing should be something we had as a priority for 0.9.15, but he mentioned how he's gonna be focusing heavy on the Syndicate campaign for that, so I'd guess 0.9.17 is the earliest we could get it

NomadicVolcano:
If he's okay with others working under his guidance, the work could be done much sooner. Otherwise I'll just balance the SMC Tier 3 races against each other based on the #3665.

Zitchas:
Definition of Archon's tier: "Tier Drak" : Can solo defeat a fleet a fleet of ships any size fielded by any race* currently in the game."**

*Special caveat regarding the Pug: Only applies regarding ships they have fielded during the human and wanderer campaigns

**When using everything at their disposal. It is assumed that the Drak have access to interdimensional tech that may allow them all sorts of defensive and offensive abilities that they have yet to demonstrate, including the ability to shift into dimensional space as a form of retreat.

That actually gives me an idea: Why don't we replace the Archon's death effect by some almost instantaneous sparkly teleporter effect, and lore it that in the odd cirumstance that they get "killed" by someone leading a thousand punishers against them or something, they are actually simply retreating to another dimension? Thus the "best" the mortal races could achieve is a "we manage to push the Archon to the point where they had to either use their really big stuff or retreat, and they picked retreat."

NomadicVolcano:
> "Tier Drak" : Can solo defeat a fleet a fleet of ships any size fielded by any race* **currently in the game**."

The Quarg are being updated, and there will be new races in intergalactic space and the SMC. The statement may no longer be applicable. We need to know define the intended limits of the Drak Archon in reference to other races that may be on par with them technologically. If that means "Tier 5 heavy warship" then fine. We need a specific definition though.

Also, from Free Worlds Reconciliation in one of the log messages, "The Quarg claim that the Pug are powerful enough that they could defeat the Quarg, *or even the Drak*, if they chose to."

> interdimensional tech that may allow them all sorts of defensive and offensive abilities

That may not necessarily apply to the Archon though. The Navy has a planet-killer bomb (almost) but their ships do not reflect that level of technology.

> replace the Archon's death effect by some almost instantaneous sparkly teleporter effect

Two suggested alternatives. It is supposed to be a living creature. Maybe we could give it an animation of dying before it goes "boom!"  Or, we could have it die and then its soul floats away (like ascending in the Stargate series).

NomadicVolcano:
To put it another way, the Archon's purpose is, "we'll spank you if you misbehave," while the Drak probably have other things that they bring out when it is time to destroy the galaxy.

Arachi-Lover:
I like that idea of the Archon death thingy being a warp-ish effect of sort.

>Also, from Free Worlds Reconciliation in one of the log messages, "The Quarg claim that the Pug are powerful enough that they could defeat the Quarg, or even the Drak, if they chose to."

I should pay more attention to logs... cause that raises many eyebrows, feels like a mistake or like it was worded wrong. How would the Quarg know the Pug have something capable of taking on the Drak like that if they are a tier under the Drak? Pug would not use T4 ships against them. and if the Drak told them they would be defeated... how is it that there is any stalemate at all? like if the Pug were capable of taking on the Drak, why would they worry about the Archons and mention them when talking about why they won't help guard the Wanderers in Kor space? the Pug have the technology to do wormholes (and I could've sworn somewhere it's said something like "the Pug broke one of the Quarg rings, something the Quarg thought even the Drak couldn't do"), while the Drak have the Archons as some eldritch higher-dimensional beings, the Drak themselves are extremely secretive so far, and both them and Pug can do hyperspace link shizzles. I always thought for sure it was intended to be portrayed as a type of stalemate.

I'd expect such a certain statement like "the Drak would lose too" from maybe a faction full of themselves or that don't know any better like the Unfettered or some crazy pirate, but the Quarg?
matter of fact the Quarg doesn't even say that actually, the response to the playing asking if the Drak would help is: `	"No," it says. "The Drak do not interfere in a war unless their rules are broken. And, it would not help you anyway; the Pug always adjust the strength of their fleet to give their opponent what they consider to be fair odds."`

that'd suggest they *can* match the Drak in combat, but I always got that as "the Drak won't help cause the space laws aren't being broken, and even if they did the Pug would just break out the strong ships, so humans would then become cannon fodder, and fleets of powerful Pug ships would be very close to Earth" , so like saying that the Drak joining wouldn't help not because the Drak would lose, but because humanity would not benefit of being caught in the middle of having the USA and USSR bring all their nuclear arsenal and duke it out (in which case humanity is some lone tribe in an isolated island or something like that), but rather would suffer with the ensuing fallout

NomadicVolcano:
> feels like a mistake or like it was worded wrong

This detail is important to balancing Tier 4. As it is in the game now, I'll assume the Pug can beat the Drak. If the endless sky master is updated to state otherwise, I'll assume otherwise.

Arachi-Lover:
>I'll assume the Pug can beat the Drak

I don't think you should, again I really think this is a mistake with the wording, the Quarg being so certain of the Pug's capabilities isn't that good a source since the Quarg was certainly biased due to them having broken one of their ringworlds, and MZ has always warned about how people may well be lying to you in the game. a lot. the Quarg are a notorious case of "oh but it's the Quarg, so they are probably lying or hiding the full truth etc. etc.". Aside from taking into account everything else I said, do pay mind that this is the word of one singular Quarg that is trying to warn human visitors of what the Pug do, maybe even doing propaganda against them one could argue, and even then, *that Quarg does not say "the Pug could beat the Drak"*.
that log entry is flawed or misworded, is what I'm saying.

Zitchas:
From the point of view of the pilot, about all it really says is "Drak and Pug are both so far above humans (and everyone else we know much about) that it doesn't really matter."

Also, I was basing my statement on what is currently in the game right now. There's no way I can take every race (or rebalancing) that may happen into account. Although I'd say it *should* remain applicable for rebalancing.

NomadicVolcano:
These points I like, and I must admit that with the current in-game lore, there is no way to know the tech levels of the Drak and Pug:

> people may well be lying to you in the game
> oh but it's the Quarg, so they are probably lying or hiding the full truth etc. etc
> Drak and Pug are both so far above humans (and everyone else we know much about) that it doesn't really matter

Thus any race that can beat the Drak or Pug must also have an ambiguous "it's somewhere up there" tier.

NomadicVolcano:
In this case, I come back to my suggestion that we give the Archon another weapon, rather than boosting the existing one. The Archon's biggest weakness is that it has only one real weapon.

I suggest an aft-launched torpedo that has enormous hit force and heat damage with a large blast radius. It would push you far away from the Archon and shut down your ships, while pushing the Archon far away from you. The Archon should not care about the heat damage. I've tried hitting the Archon with 80k heat damage per second sustained for ten minutes and it doesn't overheat.

The torpedo won't be a problem for the Gascraft if we either give that Archon no torpedoes (launcher with no ammo) or make it slow enough that the Gascraft can dodge it.

Arachi-Lover:
I've done some tinkering and did the data for 2 ideas of weapons I had for the Drak, one of which is a "secondary weapon" of sorts that lasts a long time, is slow but follows you for that long time, and hits you with about a trillion slowing damage once it hits, but I'd want to save that for something else, the Archon's build is kind of iconic now so if the AMC is the only weapon it has in a sense, we should buff the other weapons along with the AMC

NomadicVolcano:
The Archon relies too much on the AMC. That's the reason why it is easy to beat. If you want it to be harder to kill, then the other weapons need to be stronger. In particular, the turret is next to useless. I suggest adding a lot of shield disruption and some heat damage to the turret. That way, when people do get within range with their skylances, their shields will become useless.

Quantumshark:
The two anti-missile fields feel redundant; one of them is enough to make the Archon essentially immune to any weapon with missile strength. Maybe one of them could be replaced with a second Draining Field?

NomadicVolcano:
> two anti-missile fields feel redundant

The fields each are limited to killing 60 missiles per second. (One per game step.) Putting in two doubles that.

NomadicVolcano:
Someone said this:

> Kinda wish we could have more of a ramp up over time; 50 disruption is only ramping from 18.6% piercing on the second shot to 21.3% piercing after a few shots, whereas 300 disruption is ramping from 57.9% on the second shot to 61.9% after a few shots

You'd have something like that if the turret has shield disruption. The longer you stay within range of the turret, the more your shields will be disrupted.

Amazinite:
Disruption on the draining field makes the most sense to me.

Zitchas:
One thing that occurs to me: The Drak were basically just trying to stop the Remnant from hurting the void sprites. Obviously, they weren't trying to wipe them out because if that was the case, well, the Remnant would be dead. 

So I see no problem with creating two AMC: The current version that is what the Archon in the gascraft mission uses, expanded with some disruption; and a "full strength" AMC that does significantly improved damage as well as decent disruption. Maybe even a slightly increased rate of fire? Basically, for the gascraft missions the Archon is pulling its punches - it is trying to scare off the Remnant, not annihilate them. But anyone who engages an Archon outside of that mission is going to get the full brunt of it.

Amazinite:
Well the Gascraft already gets one shot by the AMC (even by a single fragment of the shot, so I'm unsure what Arachi was talking about earlier), so no point in having a separate version just for those missions unless we're making it even weaker.

Arachi-Lover:
I am 100% positive my Gascraft was nearly out of shields in more than one occasion when doing the missions, but did not die, and I never got close to the Archon in those missions either

Zitchas:
Likewise, I'm sure I've had instances where I got glancing blows from the AMC and survived. Maybe some kind of explosion from it hitting a nearby asteroid instead? I'm not sure what exactly happened, but I know I've been nicked before and survived. Really rare, though. Normally if I connect with any part of it I'm toast.

Amazinite:
Are you sure you weren't taking shots from the turrets? They do have some pretty substantial range. I'm flying around Nenia right now and a single fragment is in fact taking me out from a direct hit, and I'm not even surviving any blast radius hits.

NomadicVolcano:
Could we up the confusion on that Archon to make it less accurate? It would make the Archon-dodging missions less of a death ride.

NomadicVolcano:
Based on Lia's test with the Quarg-outfitted Haven, I suggest:

1. Massive ion and heat damage in the turret
2. Massive shield disruption in the field, and expand the field to 480 so it can hit ships with skylances.

NomadicVolcano:
@Zitchas I did a PR to your branch in your fork with my suggested changes. If you merge my PR into your branch, it will update this PR.

NomadicVolcano:
The Archon's turret is a major threat to the Gascraft after this, but I found 30 confusion is enough so the mission is no longer an "instant death by orbit camping" mission. It's actually easier than before the Archon's debuff boosts.

Zitchas:
Thank you for the PR. That being said, there were too many things changing all at once. For this particular project, I'd rather implement changes to the weapons one at a time so they can be tested to see if they get the desired effect. 

As for updating my branches... I've got a system worked out to update almost everything fairly quickly. The only reason it wasn't already up to date was that Amazinite has expressed the opinion in the past that updating *too* often has unintended consequences. Namely spamming notifications, but also cluttering the graphical trace display thingy. But as you illustrated, a bunch of stuff has been merged recently, so... Updated.

As everyone can see, I just implemented Amazinite's suggestion for turning the draining field into more of an "instant-aura." I'd appreciate feedback as to how that changes the balance of things.

NomadicVolcano:
> Thank you for the PR. That being said, there were too many things changing all at once. For this particular project, I'd rather implement changes to the weapons one at a time so they can be tested to see if they get the desired effect.

I'll do a separate PR for the turret then.

Zitchas:
@Amazinite I implemented your "range override" attribute on the Draining field, but the error checker isn't liking it. Was that just a "hoped for future attribute," or did I make a mistake inputing it?

NomadicVolcano:
Your changes are insufficient to deal with a fleet of super-heavies armed with Quarg weapons, but they will definitely help with fleets of smaller ships

PR #4993 adds heat and ion damage to the turret, which is enough to shut down super-heavies before they can get in range. The turret shuts down the gascraft for several seconds, spelling instant death, so I gave that Archon 30 confusion. Those missions are doable again.

I think the two of these PRs together should be enough to make the Archon invulnerable to super-heavies, even from the Quarg.

Amazinite:
> I implemented your "range override" attribute on the Draining field, but the error checker isn't liking it. Was that just a "hoped for future attribute," or did I make a mistake inputing it?

Well I did reference a currently open PR. :)

data/drak/drak.txt
```diff
@@ -35,23 +35,28 @@ effect "distancer flare"
...
 	"outfit space" -84
 	"weapon capacity" -84
 	"turret mounts" -1
 	weapon
-		"turret turn" 4.
-		"velocity" 400
+		"turret turn" 180 # instant turret tracking
+		"velocity" 1 # reduced to one as to get an aura effect
 		"lifetime" 1
+		"range override" 400 # from #4522 so that the weapon still fires as intended
```
Amazinite, line 48: For the sake of removing the error, you can comment this line out until the functionality is actually added. Anyone who wants to test this needs to uncomment this line and compile #4522.
```suggestion
#		"range override" 400 # from #4522 so that the weapon still fires as intended 
```

ravenshining:
The Haven build I used to solo the Archon involved - on top of the Haven's innate stats - 12 Quantum Shield Generators, 6 Quarg Skylances, 5 Antimatter Cores, and 4 Heliarch Attractors.  Minimal engines and no cooling.  I had room to spare, and while I didn't need all 6 Skylances to overcome the Drak's regen, it took a *while* to take it down.

I think adding disruption to the Drak would effectively nerf that build, as it relied on using the Quantum Sheild Generators in order to overcome the Draks DPS without ever scratching the hull.

Zitchas:
Thanks for the feedback! Would you care to test this revised version?

The AMC has the fairly strong disruption, and the Drak Draining Field now has a very strong "force field" effect that will make it quite difficult for it to be swarmed. Something like the Haven probably still has enough thrust to overcome it, but I'm interested in how it changes things anyway.  (This is in addition to the fairly small disruption and ionization effect that were already added to the Draining field by Amazinite's suggestion).

I've tested this under two scenarios:
- The gascraft mission. The current range and force of the Draining field is enough that if the Archon happens to be fairly close to the gascraft when the player takes off, the gascraft gets booted *just* fast enough to stay ahead of the AMC bolt. Sometimes the turret gets a hit in, but not always. I tried it quite a few times, and everytime the archon is too close to me it gives me a "escaped by the skin of my teeth" sort of feeling. Which is reinforced by the fact that I took damage as well as a touch of disruption and ionization. Neither is enough to seriously hamper the gascraft,

- Using a "mega fleet" of 500 dreadnoughts and 30 Albatrosses. I positioned the ships in a loose ring around the system so that when I attack a void sprite and then told them to attack the archon when it appeared, they could attack it from all sides. The results were hilarious. The Drak Draining field kept them all well back from it. Occasionally one ship would get a run at it and some good timing, and would break through enough to get some shots off; but by and large it felt exactly like I would expect a fight between a Tier 4 ship and a tier 1.x fleet should feel like: flies trying to swarm a fan, The Archon probably wasn't killing dreadnoughts quite as fast as I would have expected, but they were going down at a reasonably fast clip. 

With these values, I think it is fairly safe to say that the time of the "let's swarm the Archon with sparrows" strategy would be well and truly ended.

Zitchas:
At this point, I'm confident that these modified values are now actually a positive change to the Archon's weaponry.

More importantly, I feel that these changes are a massive improvement to the feel of the gascraft missions. Given that the gascraft missions are the main time the player is going to be on the receiving end of the Archon's weapons, this element weighs strongly in my opinion that this is now merge-worthy. 

In fact, given that it fixes the fact that the return portion of the gascraft missions are basically a russian-roulette style "Let's see if you die on take-off today" where even veteran players expect that they'll probably have to reload a significant fraction of the time; I would consider this a "Fix" category change to the game.

NomadicVolcano:
@ravenshining Could you please test your Haven build against Zitchas's modified field? If it is sufficient to stop the Haven, I'd like to close my turret change PR (#4993) to put off the impossibility of the gascraft missions to a later date.

MasterOfGrey:
Just for my 2 cents worth up there. I feel like a lot of the difficulty of definitions could be resolved by simply redefining the Archon as a Light Warship.

Then the Drak are just whatever tier that they turn out to be when more content finally fills in the scale all the way up.

Zitchas:
As far as definitions go, that's not necessarily a bad idea.

Arachi-Lover:
meh, the Archon is never defined as a HW outside of its data in the code, so if one doesn't got digging for information, they'd never come across that. to play along though, I doubt Drak ships will just be "Archon but X times larger/smaller", and even then the Archon is the ship with the highest weapon cap in the game currently, has 8 turrets, and 1 very heavy forward facing gun. I'd classify that as Heavy, even if it ends up not being the most powerful ship the Drak have at their disposal

Zitchas:
I agree, it clearly is "heavy" as far as everything else is concerned. 
That being said, I'd just as soon have a "Special" category for it. It's unique (so far). It probably does both it and everyone else a diservice by having a label like HW which leads everyone to make assumptions about where it may fit into the Drak lineup.

MasterOfGrey:
In any event, divorcing it from the 'heavy warship' definition probably has value for the ongoing discussions around balance.

Zitchas:
@Amazinite I just tried switching this to the 0.9.13 base so I could test the range override, and it pulled changes from one of the source code files along with it which also caused errors in the checks, so I switched it back. So now it is back to just not liking the "range override" attribute.

Should I just leave this until 0.9.13 is merged? Or close and re-create this as a 0.9.13 based branch so it can be tested and merged there? It's a very simple change, so it would be easy enough to do.

For that matter, is there anything else that needs changing? It looks pretty much ready to go to me.

Amazinite:
> Should I just leave this until 0.9.13 is merged? 

Yes.

Zitchas:
@Amazinite As you suggested on Discord, I merged the master into this so it is up to date and, as predicted, it now successfully passes its checks. So it should be good to go now.

Issue #2578, opened 2017-05-30
ability to change transparancy value on haze


KiLEdEnNis:
to explain the title: add the ability to define how transparent a haze is

example: say you have 3 "corners" of space, connected through hyperlinks, and you want to give all 3 corners a different color of haze, ya want a smooth transition without making like 200 more haze images
instead ya can give the haze a lower transparancy the closer the system gets to the center, so ya'd see that every corner has a different haze color, but there's no clear border

since there are only 2 actual hazes in the base game, it may not be that useful outside of plugins, so it's not really high priority, but any plans of adding it at some point?
and is it even possible since the haze pictures themselves are colored over a black background?

endless-sky:
Given the limits of how small a difference from system to system would be noticeable, I think just having a dozen or so different colors would be a fine work-around to this.

PR #2488, opened 2017-05-06
Lots of haze

TheUnfetteredOne:
Looking at a black and blue background with a bunch of white dots all over the place gets boring.
Looking at that awesome red background in Remnant space is cool. 
I thought it would be cool to make haze for other areas of space, so I did it. :3

Here are some previews.
Wanderer space:
wanderer haze
Unfettered space:
unfettered haze
Coalition space:
coalition haze
Korath space (both around the core and in their home territory):
korath haze
Hai space:
hai haze

PR #4986, opened 2020-04-12
Haze for sparse and dense areas of the galaxy

ravenshining:
**Content (Artwork)**

## Summary
This PR adds ~~5~~ ~~8~~ 7 new types of haze, including 3 with wholly new made-from-scratch textures.
- Brighter and darker hazes for use anywhere the galaxy is brighter or darker
- A hazeless haze for the fringes of the galaxy
- ~~A smattering of larger stars instead of haze, in Wanderer space and other dark areas, like near Earth~~
- ~~A starry haze for weird dark places beyond human space, such as Wanderer territory~~
- Yellow haze ~~surrounding the Core~~ ~~in Drak nova systems~~ in Host
- ~~Orange haze~~ ~~transition between the Core and the Ember Wastes~~ ~~for a few other strange systems in Coalition and Wanderer space~~
- Multicolour haze ~~in Core systems~~ in Korath worlds very close to the core
- Double-size, sky-filling multicolour haze for Sagittarius A* and Kor Ak'Mari

## Save File
This save file can be used to play through the new art content. It zeros most rep, has the whole galaxy visited, gives you a Falcon with tons of engines and fuel processors, keystone, hyper & jump drives, etc.  Pug wormhole added by hand, All Content plugin start used but not required.

[Max Bunks.txt](https://github.com/endless-sky/endless-sky/files/4469384/Max.Bunks.txt)

## PR Checklist
 - [x] I updated the copyright attributions
[haze.xcf.zip](https://github.com/endless-sky/endless-sky/files/4466400/haze.xcf.zip)
 - [N/A] ~~I created a PR to the endless-sky-high-dpi repo with the `@2x` versions of these art assets:~~

-----------------------
**Feature:** This PR partially implements a feature request discussed in issue #4420 and MZ's comment on #2578 , while hopefully avoiding the problems with #2488 by using different haze sparingly, in non-arbitrary colours, and in ways appropriate to the galaxy background.

## UI Screenshots

Fullscreen haze for the innermost Core:
haze-full
haze-full-menu

All screenshots below are mildly out of date

Variable density haze for human/hai/quarg space:
haze-double
haze-double-menu
haze-half
haze-half-menu
haze-none
haze-none-menu

These hazes are included but no longer appear in these particular star systems, see below:
hazeorange
hazeorange-menu
haze-blackbody
haze-blackbody-menu
haze-stars
haze-stars-menu
haze-yellow
haze-yellow-menu

These two hazes were adjusted, see below, but here they are as I had them originally:
hazefull
hazefull-menu
hazeyellow
hazeyellow-menu

## Performance Impact
Possible negative impact on Sagittarius A* and Kor Ak'Mari. However, there is presently no combat in the former and little in the latter, so it should be fine.


beccabunny:
Some of these seem way too colourful imo

ravenshining:
Sure, I'm totally open to toning it down.  Can you be more specific?  I did try to avoid greens and magentas in the multicolour ones, but they may all be too saturated.

ravenshining:
Darkened, shrunk, & desaturated the large haze:
darker
haze-fill-menu

Desaturated the yellow haze:
darkery
haze-yellow-menu

Zitchas:
I really like the looks of that large haze.

Regarding the edges of the Ember Waste - I'm not sure we want to soften the edges of that too much. It was created by sudden cataclysmic forces - I'm not sure that gentle fade ins / fade outs is entirely appropropriate for the edge of something so sudden. (supernova style events, for instance)

Amazinite:
I'd suggest not changing any of the haze in human space as per the ideas expressed in https://github.com/endless-sky/endless-sky/pull/2488#issuecomment-299669662.
Outside of that, I think that haze-stars makes sense to use in the Wanderer region and the outer Coalition regions given how sparse the background looks in those region. And the more cluttered haze makes sense for the inner core (i.e. the core regions no in human space). 

How about a unique haze for WR stars and for nova remnant systems? Instead of adding cool visual to places where they (justifiably) don't exist right now, bolster the visuals of the places that already look cool.

NomadicVolcano:
Amazine - remember we're expanding beyond the Milky Way now into intergalactic space and the Small Magellanic Cloud.

I need alternative hazes for the dark intergalactic space where the Whispering Void will reside. The Philosopher's Paradise (#4943) has a "black haze" that has hints of purple and green so it doesn't look too black. It's used in intergalactic space in regions where no stars are visible. The stars in the systems are weird types that would *not* be visible from thousands of lightyears away. A wider variety of "black haze" would be nice as you get farther and farther into deep space.

Also, the SMC is a lighter galaxy so it'll need a different haze, especially near the center. We probably don't need six types of haze, but two would be good: one for the bright region in the middle and one for the dimmer regions on the outside.

To put it simply: although we may not need new haze in the Milky Way, we will need some in the near future as we expand beyond the galactic rim.

Amazinite:
>Amazine - remember we're expanding beyond the Milky Way now into intergalactic space and the Small Magellanic Cloud.
>
>I need alternative hazes for the dark intergalactic space where the Whispering Void will reside...
>
>To put it simply: although we may not need new haze in the Milky Way, we will need some in the near future as we expand beyond the galactic rim.

I never said these hazes were bad and wouldn't be merged. I just said they shouldn't be used where they're being used right now.

ravenshining:
Confining haze based on species also runs afoul of another criticism of #2488 , that political boundaries shouldn't be defined by space colour.  MZ certainly had a point about keeping the spectacular things out of most of human space, but I don't think that rules out subtle variations or an exception here and there.  Sundive in the Sheratan system, for example, is further into the bright bulk of the core than some Korath exile systems, and its description notes that:

> Sundive is also home to a large astronomical laboratory with radio telescopes and other devices for studying the galactic core and the massive black hole that resides there.

A few human systems are far into the black, and such systems tend to be sparsely populated or pirate systems, I think having the haze go away would be a nice way to tell the player they've reached the edge of civilisation as they know it.

Furthermore, with the exception of Answer, Coalition space is actually quite bright, located in a dense spiral arm.  The Rim, the Deep, Near Earth, and the Far North are all much darker, and, ignoring for a moment how horribly wrong all our star positions are, IRL Sol is in a relatively dark section of the Milky Way.  I think it'd be really weird do go from a dark part of the galaxy into a brighter part. 

So, how about this:
- Keep the full-sky red-yellow-white-blue haze for Saggitarius A* and Kor Ak'Mari
- Systems surrounding the Core, possibly including Sheratan, get the darker multicolour haze
- I'll create a darker and lighter/bigger variant of the standard blue haze for use in most of the galaxy, including human space.
- Use the yellow and orange hazes only for the occasional system with a nova, WR, or something crazy on the galaxy map
- Distribution of haze something like so:

galaxy-haze

ravenshining:
Okay, made the blue variants and made changes as above.  Normally I like making fresh artwork rather than just making minor adjustments to existing things, but in this case, I think it will allow us to add some visual cues as to how far rimward or coreward one is, while still preserving the surprise factor when players jump out of human space and see a new backdrop for the first time.

- ALL human, Hai, and Quarg worlds have one of three darknesses of blue haze, or in a few rare fringe-of-the-galaxy cases, no haze.
- Reverted northern Ember Waste systems to red
- "Star" and Orange haze only appears in Coalition and Wanderer space
- Orange and Yellow haze only appear in about 2-4 systems each
- Korath exile space is now multicolour instead of yellow

haze-double
haze-double-menu
haze-half
haze-half-menu
haze-none
haze-none-menu

ravenshining:
Just competed a tour of the galaxy, and caught up with Amazanite's twitch comments.

- It is impossible now to see anything other than blue or black without a Jump Drive or Quantum Keystone

- The thinking behind adding stars to the "hazeless" hazes, is that having less dust and less light overall allows you to see more stars - like how light pollution hides the night sky on Earth.

- I'm thinking of nixing the orange haze entirely since it's not unique and looks too similar to the Ember Wastes, and change the two systems that use it to yellow.  Unless people think it would look good for some select EW systems, like ones with WR stars.

- I'm leaning towards making one or two more Coalition systems yellow or starry

NomadicVolcano:
In intergalactic space, you'd expect to see nearby galaxies. Do you think you could make some haze with dimmed images of distant galaxies? (It would save me the trouble later.)

MCOfficer:
> In intergalactic space, you'd expect to see nearby galaxies. Do you think you could make some haze with dimmed images of distant galaxies? (It would save me the trouble later.)

i'd advocate against that's because distinct objects (unless they're spread like the stars here) are a very easy-to-spot pattern. the only alternatives, given the haze image's size, are to plaster it with uniform galaxies, which is probably gonna look horrible.

NomadicVolcano:
> > In intergalactic space, you'd expect to see nearby galaxies. Do you think you could make some haze with dimmed images of distant galaxies? (It would save me the trouble later.)
> 
> i'd advocate against that's because distinct objects (unless they're spread like the stars here) are a very easy-to-spot pattern. the only alternatives, given the haze image's size, are to plaster it with uniform galaxies, which is probably gonna look horrible.

A fine point. I forgot that aspect. The galaxies would look silly and repetitive (and sometimes overlapping).

ravenshining:
Yes, and as little sense as it makes for us to see stars going by while flying inside a star system, having *galaxies* go past is even worse.

Unfortunately for making games fun, from what I can tell intergalactic space is probably just black for the most part.  I tried flying to a point midway between the Milky Way and the LMC, and looking out into the void in Celestia, and beheld... this...

intergalactic

Unsure if it really would be that black, I immediately started to search for what scientists think intergalactic space would look like, and they tell pretty much the same story:  space is big and empty, except where it's not like where I took the shot above, and what its full of there is dark matter.

NomadicVolcano:
I tried a purely black image for the haze, but it doesn't look good. That's why I went with an "almost black" one with hints of purple and green. Try the Philosopher's Paradise (#4943) and go to the regions of the map that are all black. You will see a nearly-black haze. The stars that are in a nebula have the normal haze.

Also, to respond to your scientific point. When you're just outside Milky Way, you're not in deep space, a million lightyears from the nearest galaxy. In reality, you'd look out and see a vast swath of stars in the direction of Coallition space. A vast swath known as the Milky Way. That's why I used an image with the Milky Way in the background for the Devil-Eye planet's landing image.

ravenshining:
The camera in that image is not pointed towards the Milky Way, it's pointed out into intergalactic space. Tilt up an you'll see the Milky Way, tilt down and you'll see the LMC.

ravenshining:
> I tried a purely black image for the haze, but it doesn't look good. That's why I went with an "almost black" one with hints of purple and green. Try the Philosopher's Paradise (#4943) and go to the regions of the map that are all black. You will see a nearly-black haze. The stars that are in a nebula have the normal haze.

Okay, I've had a chance to test that now.  The "almost black" haze might fill the gap between haze-half and haze-none, but it doesn't do anything about the fact that you're no longer amongst the stars.

I've tried adding a non-additive haze, and while that *works* as far as reducing the number of stars goes, having the engine draw stars only to cover them up with black images is a bit redundant.

A better solution, particularly for intergalactic space but possibly for the galaxy in general, would be to change the number of stars.  Currently, this is a fixed number, the first one set here in GameData.cpp line 305:
`	background.Init(16384, 4096);`
which is referencing StarField.cpp:
`void StarField::Init(int stars, int width)`
So, adding an attribute "stars" to a system description, which defaults to 16384 and is read in StarField::Init, is where I'd like to go next, if I take this PR into the realm of intergalactic space.

ravenshining:
Changes in response to this morning's stream:

(systems that were --> are now)

- orange --> normal
- double --> 133%
- half --> 67%
- none --> 33%
- starry --> none
- yellow --> blackbody

- Starry is deleted.
- Orange is deleted.
- Yellow is darkened 50% and used only in Host.
- Blackbody has been desaturated 50% to reduce it's "uniformity"
- None in human space restricted to the truly remote and black systems of Arneb and Gamma Corvi
- Some coreward Wanderer systems have 33% haze, since fewer human systems have no haze

PR #4992, opened 2020-04-15
Update World-Ship description so the world-ship is used for things other than storing population.

NomadicVolcano:
Presently the World-Ship is used for combat roles despite the description saying it is a population center. This updates the description so the ship is also used for defensive or offensive purposes.

### DIFF
--- a/data/korath/korath ships.txt
+++ b/data/korath/korath ships.txt
@@ -238,7 +238,7 @@ ship "Korath World-Ship"
 	explode "large explosion" 40
 	explode "huge explosion" 50
 	"final explode" "final explosion large"
-	description "Korath World-Ships are massive nomadic habitats Korath use to transport troops, pillaged goods, or house their population."
+	description "Korath World-Ships are massive nomadic habitats Korath use to transport troops and pillaged goods, or house their population."
 
 ship "Korath World-Ship" "Korath World-Ship B"
 	sprite "ship/world-ship b"
--- a/data/korath/korath ships.txt
+++ b/data/korath/korath ships.txt
@@ -238,7 +238,7 @@ ship "Korath World-Ship"
 	explode "large explosion" 40
 	explode "huge explosion" 50
 	"final explode" "final explosion large"
-	description "Korath World-Ships are massive nomadic habitats that hold all that is left of their people."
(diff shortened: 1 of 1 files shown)


ravenshining:
This description is never shown anywhere in the vanilla game, so I'm not sure what the point of this PR is?

NomadicVolcano:
> This description is never shown anywhere in the vanilla game, so I'm not sure what the point of this PR is?

The lore in game and out of game is that the Korath live entirely in these world-ships. This updates the lore so that the world-ships are used for other purposes.

When you're wandering the stars slaughtering the world-ships, you can plausibly believe that there were no children inside.

The Korath Exiles are no longer sending world-ships filled with children to attack the Remnant.

Zitchas:
That's actually a valid point. Is there anything in-game that a player can see without using any plug-ins that tell them the World-Ship actually houses large chunks of their population? Or is it only implied by the name?

NomadicVolcano:
We see almost nothing in-game from the Korath. What little lore we have comes from planet and outfit descriptions, and hearsay on what MZ has spoken.

NomadicVolcano:
I will clarify: this PR isn't just to update the text, it is to update the lore. In-game, out-of-game, wherever. The lore shall say that some Korath world-ships are not packed with children.

NomadicVolcano:
As Amazinite has given the thumbs down to this, I am concluding we *do* need some additional ships. It is not sufficient to repurpose or rebalance the existing ones and say there's no kids inside.

NomadicVolcano:
I'll wait to close this PR until people other than me have noticed Amazinite's thumbs down.

Zitchas:
I don't think it was ever in doubt that we needed new ships. Hence why there's the entire PR full of them.... :)

NomadicVolcano:
If we could simply say that the World-Ship doesn't always have kids inside, then we could use it for both combat and kids. It would fill the big plot hole without adding new ships. Having new ships would just be nice to have.

As we can't simply say the ships are kid-free, either you have to stop shooting them, or we need a different ship for boss fights.

NomadicVolcano:
What I mean is, adding new ships is a necessity now, not a "nice to have feature"

NomadicVolcano:
Closing PR.

PR #4993, opened 2020-04-16
Drak turret updates to shut down super-heavies

NomadicVolcano:
The Drak Archon is unable to handle ships whose damage output and repair are equal to its own. As Tier 2 and 3 super-heavies are added, this makes the Archon vulnerable, especially if one mixes outfits from many races. 

This update adds ion and heat damage to the Drak Turret sufficient to shut down super-heavies with Korath cooling, Quarg batteries, Quarg generators, and enough thrust to outrun the Archon. It was tested with three min-maxed Korath Havens who, as a result of these changes, could no longer get in range.

Unfortunately, this makes the turret essentially instant death to the Gascraft since that ship is shut down for a few seconds after one hit. This update also adds 30 confusion to the Nenia Drak during the gascraft missions, sufficient to dodge the shots with good reliability.

### DIFF
--- a/data/drak/drak.txt
+++ b/data/drak/drak.txt
@@ -94,6 +94,10 @@ outfit "Drak Turret"
 		"shield damage" 256
 		"hull damage" 256
 		"hit force" 120
+		# Updates to shut down super-heavies that can handle
+		# the Drak's damage output:
+		"ion damage" 20000
+		"heat damage" 40000
 	description "This is a turret based cannon with significant range and accuracy."
 
 effect "drak bolt impact"
--- a/data/remnant/remnant missions.txt
+++ b/data/remnant/remnant missions.txt
@@ -768,7 +768,9 @@ mission "Remnant: Void Sprites 3"
 	npc
 		government " Drak "
 		system "Nenia"
-		personality staying heroic nemesis uninterested
+		personality
+			staying heroic nemesis uninterested
+			confusion 30
 		ship "Archon (Cloaked)" "Lifted Lorax"
 	on stopover
 		conversation
@@ -1256,7 +1258,9 @@ mission "Remnant: Return the Samples"
 	npc
 		government " Drak "
 		system "Nenia"
-		personality staying heroic nemesis uninterested
+		personality
+			staying heroic nemesis uninterested
+			confusion 30
 		ship "Archon (Cloaked)" "Lifted Lorax"
 	on accept
 		log "People" "Plume" `A Remnant researcher specializing in xenobiology, Plume is at the forefront of recent efforts to restart research on the void sprites and a leading expert in non-carbon based lifeforms.`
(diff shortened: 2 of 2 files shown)


Zitchas:
I think part of the fun of the Gascraft missions is being able to deal with Archon on one's own merits, not just because it has an artificial added confusion. 

As far as that goes, that confusion affects all shots, and that will actually make the AMC less predictable, which in my view, makes it harder.

NomadicVolcano:
An alternative is to make an Archon variant that won't use its turret at all. As to the other points:

I've found that when the Archon starts right next to you (as it often does) then no matter what you do, you're dead. Confusion seems to get rid of that. 

If you stay far away from the Archon, you can see the shot coming on the HUD and avoid it regardless of confusion.

It changes the mission from "often impossible" to "requires watching the HUD" while also adapting to the strength of the turret.

NomadicVolcano:
The problem with the gascraft missions is that they rely on the Drak turret being useless. If it ceases to be useless, then that breaks the gascraft missions.

Zitchas:
I made some more changes to the Drak Draining field, which has pretty much solved the gascraft issue. With those changes, the need for the Confusion is gone; and we can safely ramp up the turret with less worries about how it will affect the gascraft mission. 

The main caveat is that any improvement to the turret has to be kept low enough that a single hit won't crippled the gascraft. Push it right to the edge, sure, but not actually kill it. With the way the Draining field works in my PR, the turret shouldn't be able to land more than a single hit most of the time. 

Even so, adjusting the turret so that it'll push the gascraft up to 90% of its heat and energy shouldn't be a problem for the gascraft mission, and does up its threat somewhat to everything else (although, admittedly, not nearly to the extent of your current values...)

NomadicVolcano:
I'd prefer not to put in unnecessary changes to the game. If the disruption field is enough to shut down the Haven based on @ravenshining 's next round of tests, then these are unnecessary at the moment. We can revisit the problem later when more super-heavies are added.

Also, pushing to 90% of its heat and energy isn't sufficient to stop a 5000 ton craft with good heat dissipation and a ton of batteries. Even with Korath outfits alone, I'm sure the Haven can handle that.

Zitchas:
Fair enough. It'll be interesting to see what people manage to do once the Emerald Sword and the Haven are in the game.

NomadicVolcano:
It will definitely be impossible to capture the Haven. It has about 35k required crew with 35k rifles. Nothing else in the game comes anywhere close to that. The Refuge is probably out of reach too with ~8k required crew and rifles. The biggest you can get is the Refabricator, which is nowhere near as durable as the Haven and only has 13 turrets.

NomadicVolcano:
With the current incarnation of the Drak Archon, one Korath Haven is unable to get within range of any weapons that can overcome the Archon's shield regeneration. I did not try a large fleet of Havens, but I'm satisfied that it is no longer possible for one to do it.

PR #4994, opened 2020-04-16
feat: Log shader link errors

janisozaur:
## Summary
Log shader link errors just like compilation ones

### DIFF
--- a/source/Shader.cpp
+++ b/source/Shader.cpp
@@ -43,10 +43,11 @@ Shader::Shader(const char *vertex, const char *fragment)
 	
 	GLint status;
 	glGetProgramiv(program, GL_LINK_STATUS, &status);
-	if(status == GL_FALSE) {
+	if(status == GL_FALSE)
+	{
 		GLint maxLength = 0;
 		glGetProgramiv(program, GL_INFO_LOG_LENGTH, &maxLength);
-		std::vector<GLchar> infoLog(maxLength);
+		vector<GLchar> infoLog(maxLength);
 		glGetProgramInfoLog(program, maxLength, &maxLength, &infoLog[0]);
 		string error(infoLog.data());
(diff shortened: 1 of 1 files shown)


source/Shader.cpp
```diff
@@ -43,8 +43,16 @@ Shader::Shader(const char *vertex, const char *fragment)
 	
 	GLint status;
 	glGetProgramiv(program, GL_LINK_STATUS, &status);
-	if(status == GL_FALSE)
+	if(status == GL_FALSE) {
```
tehhowch, line 46: Please match the style of the existing code

(braces on newline when following parentheses, no `std::` prefix when not required)

janisozaur: What tool do I use to apply your formatting style? I haven't seen any clang-format rules and the reply hardly looks like the default one.

tehhowch: There are no automated tool rules for the conventions used by this project. We do have editorconfig support for the whitespace requirements.

Other than that, there is the style guide:  https://endless-sky.github.io/styleguide/styleguide.xml

janisozaur: Well, that's quite surprising. How about clang-format then?

janisozaur:
```suggestion
	if(status == GL_FALSE)
	{
```

source/Shader.cpp
```diff
@@ -43,8 +43,16 @@ Shader::Shader(const char *vertex, const char *fragment)
...
+		GLint maxLength = 0;
+		glGetProgramiv(program, GL_INFO_LOG_LENGTH, &maxLength);
+		std::vector<GLchar> infoLog(maxLength);
```
janisozaur, line 49:
```suggestion
		vector<GLchar> infoLog(maxLength);
```

Issue #4720, opened 2019-12-20
Allow missions to override navigation and in-flight controls.


petervdmeer:
**Is your feature request related to a problem? Please describe.**
This is a request to allow missions to control navigation and to control the players spaceship in flight.

Examples of missions that could benefit from this:
- "Your hyperdrive suddenly went out of control" type of stories could use this to make the players ship makes some hyperspace jumps before returning (in-flight) controls to the player
- A mission like "Shady passenger transport 3 - double cross" could use this to let the hijackers fly the players ship to their desired destination
- The intro missions might use this to let James fly the ship instead of the player (if the player wants to have a demo on how to fly).

**Related Issue Links**
PR #4715 allows for more easy adding of new types of flight-control-inputs; one possible new control-input would be to have missions provide control inputs.
PR #4370 requires this type of in-flight controls override for automatic tests; it would probably use the same syntax as the feature described in this feature request.

**Describe the solution you'd like**
A mission can already trigger a launch of the players ship. We could add a section under launch (or under a different keyword) that contains instructions for:
- navigation (set destination system, set destination planet)
- setting (AI) behaviours on the entity controlling the players spaceship
- directly controlling the player spaceship using basic command inputs (turn left/right/jump/land)
- directly controlling the player spaceship using higher-level commands (turn to ... angle, kill enemies, ...)

Controls inputs could only be given until the next landing, on landing a next mission could trigger though (with possibly new automated controls). If the inputs result in the ship just flying around in some system, then controls are returned to the player at that point.

An On-Enter mission trigger would not trigger until the automated controls are done (and might not trigger at all if the end-result of the flight is landing).

**Describe alternatives you've considered**
Just implementing a subset of this feature ("only navigation and AI behaviour controls" or "only basic direct controls") could already be valuable for some game-content.

**Additional context**
Automatic control of the players ship can be tricky; I would expect mission writers to only use this type of override when the starting, intermediate and end-systems are well known. A mission writer wouldn't want to accidentally send a player through unknown hostile territory.

Please do provide feedback if you have any relevant feedback. Is this a feature which you would like to use in stories that you write? If so, then which controls would you like to have?

Zitchas:
Hmmm... As content creators we'd have to be very careful with this. I could see players getting very frustrated with this.

[...]

PR #4997, opened 2020-04-17
feat: Add Command loading and sending it to the Engine

petervdmeer:
**Feature:** This PR implements the command loading and sending feature from #4370

## Feature Details
This PR adds a function to the Command class to load commands from datafiles and adds a function in the Engine to use a given command in the active frame.

This is part of the automatic-testing PR #4370 and can also be used for mission scripted commands as suggested in PR #4720.
The automatic testing PR was getting quite large, so splitting this off so that it can be reviewed more easily.

## UI Screenshots
N/A

## Usage Examples
See #4370 for usage of this feature.

## Testing Done
Tested multiple times as part of #4370.

## Performance Impact
No performance impact expected. This functionality would only be used during automated testing and/or during scripted missions and would typically be used only once per frame during such usage.


petervdmeer:
Closing, to be handled as part of #4370.

petervdmeer:
Re-opened; #4370 is not ready for merging yet, this PR might be merged ahead (to make #4370 smaller).

source/Command.cpp
```diff
@@ -221,6 +221,48 @@ bool Command::HasConflict() const
...
+			{"gather", Command::GATHER},
+			{"hold", Command::HOLD},
+			{"ammo", Command::AMMO}
+		};
+		
+		auto it = lookup.find(node.Token(i));
+		if(it != lookup.end())
+			Set(it->second);
+		else
+			node.PrintTrace("Skipping unrecognized command:");
```
tehhowch, line 260: Seems like the input node can be e.g. `<somekey> "primary" "secondary" "tertiary" "ammo"`, so we should also flag the particular skipped item
```suggestion
			node.PrintTrace("Skipping unrecognized command \"" + node.Token(i) + "\":");
```

petervdmeer:
> Looks much better without the huge if-else if chain

Yes, it indeed looks better.

PR #4998, opened 2020-04-17
feat: Extend CI environment with graphical testframework

petervdmeer:
**Feature:** This PR implements the graphical test framework for the CI-environment from #4370

## Feature Details
This PR adds a simulator of the graphical X-server (xvfb) and its dependencies to the github-workflows
This PR adds a shell-script that calls the ES-executable with a parameter to run the (load-only) empty testcase. This testcase tests if ES can start successfully in the CI-environment.
This PR adds the minimum amount of code to main.cpp to run the empty testcase (hardcoded).

## UI Screenshots
N/A

## Usage Examples
Upload this PR to Github and check if the CI-test that is added succeeds.
Other usage instructions (for local testing) are in run_tests_headless.sh included in this PR.

## Testing Done
The CI-test for this PR will test this change.

## Performance Impact
No impact expected; this adds a couple of instructions for every frame that verifies if a test is running. Should be insignificant compared to all the other calculations that are done every frame.


.github/workflows/ci.yml
```diff
@@ -221,6 +221,31 @@ jobs:
         chmod +x endless-sky
         ./tests/test_parse.sh ./endless-sky
 
+
+  # This section is for tests on ubuntu 16 (and on ubuntu 20).
+  # The tests with Xvfb in this section work fine on ubuntu 16 and on Ubuntu 20, but they don't work on Ubuntu 18.
+  test_ubuntu16:
```
MCOfficer, line 227: perhaps we should rename this to `autotest_ubuntu16`, and the others to `parse_*`

petervdmeer: Good idea, done.
I also re-ordered them so that all the parse-tests are together again.

petervdmeer:

source/main.cpp
```diff
@@ -251,6 +255,16 @@ void GameLoop(PlayerInfo &player, Conversation &conversation, bool &debugMode)
...
 		((!isPaused && menuPanels.IsEmpty()) ? gamePanels : menuPanels).StepAll();
 		
+		// Currently running the hardcoded "empty" testcase;
+		//   Wait for the game to be fully loaded and then quit.
+		//
+		// This testcase can catch issues related to startup/data-loading,
+		// and it helps to show if the CI testframework is working.
+		// This hardcoded testcase is expected to be replaced by the
+		// larger testframework that is planned for ES.
+		if(!testToRun.empty() && GameData::Progress() == 1)
```
tehhowch, line 265: [GameData::Progress](https://github.com/endless-sky/endless-sky/blob/b274eae897e2aa4c70f187f8f016dd56a47da73a/source/GameData.cpp#L312) is not a simple accessor, unfortunately, despite its name implying it is.

petervdmeer: Would it be okay if I move the isReady variable from https://github.com/endless-sky/endless-sky/blob/f3edfc4cd4cfdfcf37b1f90b81d511cdd555868a/source/MenuPanel.cpp#L46 to the UI class, set (and check) it from MenuPanel and also check it from here?

tehhowch: That sounds reasonable for now. `IsReady` might be a little ambiguous of a name, though.

(Ideally, `GameData` would be able to tell you that it had finished loading the game data it was asked to load)

source/main.cpp
```diff
@@ -258,6 +262,16 @@ void GameLoop(PlayerInfo &player, Conversation &conversation, bool &debugMode)
...
+		// This testcase can catch issues related to startup/data-loading,
+		// and it helps to show if the CI testframework is working.
+		// This hardcoded testcase is expected to be replaced by the
+		// larger testframework that is planned for ES.
+		if(!testToRun.empty() && GameData::IsLoaded())
+			menuPanels.Quit();
+		
 		// Caps lock slows the frame rate in debug mode.
 		// Slowing eases in and out over a couple of frames.
 		if((mod & KMOD_CAPS) && inFlight && debugMode)
```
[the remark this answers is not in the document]
petervdmeer: The idea of the current location is that the test-code is safely placed after all processing for the step (events, local logic, stepping of the panels) is done and the game is in a consistent state.

MenuPanels.StepAll() performs processing like pushing or popping menu-panels from the UI-stack, so if we run the testcode before the MenuPanels.StepAll, then it is possible that some actions got triggered that will add or remove menus on the stack, but that the menus themselves have not been pushed or popped yet. Having the test-code interact with the menus just before MenuPanels.StepAll() sounds risky to me, especially since the game itself does not support giving inputs at that moment.

I prefer not to place the test-code after DrawAll, because skipFrame (on line 281 before this change, line 295 in this change) causes DrawAll to be called less often when fast-forward is active. Calling the test-functions with variable timing can make it more difficult to use timeout-counters in the test.

Another location that I would consider consistent and safe for calling the test-code would be the beginning of this loop (before the `while(SDL_PollEvent)`).

petervdmeer:
Updates from tehhowch also look good to me.

PR #5000, opened 2020-04-18
feat(graphics): Add user preference for VSync setting

tehhowch:
**Feature:** This PR implements a feature cut from #4472 by https://github.com/endless-sky/endless-sky/commit/eb2e1b0881757ea5471cbd3a6fcfba5c11730eb8

## Feature Details
 - Game defaults to VSync ON (current behavior, and behavior prior to #4472 
 - Preferences -> Settings has new option under display, "VSync", with 3 possible options: "off", "on", and "adaptive". Only users' whose systems can successfully enable adaptive VSync will see the "adaptive" option (unless the user manually edits their preference file to the associated vsync option value).

An additional GL extension query is performed to check if the GL driver actually supports adaptive VSync. I don't have any systems that actually support adaptive vsync, so I can't verify that the queried extension is the correct one, but based on readings of OpenGL documentation this appears to be correct (e.g. matching "GLX_EXT_swap_control_tear")

## UI Screenshots
Before:
image

After:
image

## Usage Examples
 - Upon clicking the current VSync option, the next VSync state will be attempted. If it is unsupported or setting it fails, then the next option after that is tried. (e.g. on -> adaptive -> off, when adaptive fails to be set properly).

## Testing Done
I loaded up the game and toggled the setting. Prior to adding the extension check, because my linux gfx drivers don't support adaptive vsync, when clicking "on" I receive the "SDL message: Negative swap interval unsupported in this GL" error and the state changes to "off". Clicking off -> "on", as expected.
After the extension check was added, the preference toggles only between "off" and "on" without generating the SDL message.

## Performance Impact
Users who want to disable VSync, or use adaptive VSync, can now do so without recompiling the game locally, and can even do so without exiting the game.


Amazinite (commented):
Copying from Discord.

Vertical Sync setting | Outcome
----------------------|--------------------------------------------
Application setting   | Switches between on and off
Fast                  | Ditto
Adaptive              | Can't switch from whatever it started on
On                    | Started on, can't switch; if started off, moves to on then can't switch
Off                   | Started on, moved to off, then can't switch; if started off, then just can't switch

tehhowch:
Seems like nvidia cards may not expose adaptive VSync as an application-controlled setting via their OpenGL interface (no `*_swap_control_tear` extension support is reported). Given that the other behavior is testing correctly (e.g. nvidia forced on -> stuck on, nvidia forced off, stuck off), this shouldn't have any negative impacts, and will allow those with slower systems to determine if turning VSync off improves their game experience without needing to exit & restart.

PR #5006, opened 2020-04-22
Add "Antibiotics to South"

Naminson:
**Content (Mission)**

## Summary
Adds a new mission that's offered on planets in non-southern worlds with the "medical" attribute. It's also supposed to only offer if the player has a reasonable armament, since it involves a fight against an engine Manta that might as well be forced.

This mission was tested on Nifel, which is in the Deep, but it should work on a number of planets.


data/human/transport missions.txt
```diff
@@ -3582,3 +3582,48 @@ mission "Hauler VI cargo prototype 3"
...
+		"armament deterrence" > 2.1
+		random < 5
+	cargo "antibiotics" 25 2 .05
+	
+	on offer
+		conversation
+			`You spot a well dressed man eyeing your ship. When you approach him, he turns to you and asks "Is this your ship? I'm looking for someone with a heavily armed fleet."`
+			choice
+				`	"It is. Why?"`
+			`The stranger begins to explain himself. "I'm arranging a shipment from a pharmaceutical plant here on <planet> to <destination>. Specifically, <cargo>. I'm sure you're aware of the fact that they're in very high demand in the south, so we have a trade deal with that planet. Unfortunately, they're in even higher demand in pirate territory, so the last couple convoys that have gone down there have all been attacked by pirate ships. I don't have the budget for an escort, so anyone I send down there has to be alone. If you can take one of these shipments to <destination>, I can arrange a payment of <payment>."`
```
Arachi-Lover, line 3605: <planet> displays the destination planet, I assume you meant the source planet? if so it needs to be <origin>

data/human/transport missions.txt
```diff
@@ -3582,3 +3582,48 @@ mission "Hauler VI cargo prototype 3"
 			`The Testament enters the planet's atmosphere at high speed and lands somewhat uncontrolled near a waste processing plant. The cargo hold is glowing red hot and some areas in the cargo hold appear to be on fire.`
 			`	Veronica disembarks, walks over to you, and apologizes for the difficult trip. She explains that the cooling systems failed, causing the sushi calamity you experienced during the multiple legs of your journey.`
 			`	You are paid <payment> and she asks you not to talk about this escort job to anybody. You leave the docks quickly in order to escape the horrible smell of overheated fish that is coming from Veronica and her ship.`
+
+mission "Antibiotics To South"
+	name "Antibiotics shipment to <destination>"
+	description `Deliver <cargo> to <destination>. Payment is <payment>. Expect attacks from a pirate ship known as the 'Australis'`
```
Anarchist2, line 3588:
```suggestion
	description `Deliver <cargo> to <destination>. Payment is <payment>. Expect attacks from a pirate ship known as the 'Australis'.`
```

Zitchas:
So, this is currently a one-off mission. Is that intentional? Is this the start of a small set of missions, or possibly leading to a repeating job or something?

Naminson:
At the moment, there isn't anything big planned to come out of this. There was a job form of this mission in an old mod I made, and I found out that it's pretty easy to accept one of them by accident when skimming over descriptions. I'd also rather not make a chain out of this mission (at least directly), since it doesn't really have anywhere to go, and I think there's too few one-offs already.

Zitchas:
Ideally we should have a lot of one offs, but they should all be a bit rarer. As a result, each playthrough people will get a different selection of one-offs; which in turn help playthroughs feel different and and thus the "world" feels bigger. It's really hard for a world to feel big if *you* are *always* the one to help the lady deliver her dog to somewhere; and *you* are always the one to help the young family move worlds. The larger the number of mini-strings and one-offs we have, the lower we can push the rng on each of them while still having people getting roughly the same number of unique missions. 

Of course, this doesn't work for strings that give unique and special rewards, but it works very well for unique missions that just have generic payments and/or no unlocks.

Naminson:
(This isn't fully ready for merge yet because theres still a few things I still want to change)

Naminson:
This should now be about ready to merge unless something else comes up, but I don't think it will.

PR #5014, opened 2020-04-24
Increase Quarg attitude towards Kor Efret

NomadicVolcano:
The Quarg claim to protect the Kor Efret, but presently the penalty for attacking the Kor Efret is quite low. This updates the Quarg attitude to 0.25, so they really will protect the Kor Efret.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -426,7 +426,7 @@ government "Quarg"
 	"player reputation" 1
 	"attitude toward"
 		"Merchant" .01
-		"Kor Efret" .01
+		"Kor Efret" .25
 		"Kor Mereti" -.01
 		"Kor Sestor" -.01
 		"Hai" .01


tehhowch:
I'm not sure the intent here - a positive value means the Quarg will defend the Efreti. Your PR goes further and makes it so that the reputation with Quarg can be increased by aiding the Efreti.

Zitchas:
It is worth noting that the size of the value doesn't impact how much effort a group will put into defending another one. A group with a value of 0.01 will put just as much effort into defending the other government as a group with a value of 1 will. The only different is how fast reputation will change in response to attacking or aiding the other group. 

Governments with a really low value are basically a way of indicating "they care about this other group and will defend them, but not so much that it has a major bearing on your relationship with them."

beccabunny:
That's not true, values of 0.01 or lower mean that the reputation will not change at all, no matter what

tehhowch:
Reputation magnitudes in (0, 0.05), exclusive, do not result in reputation changes.

Zitchas:
So a 0.05 *will* change reputation, just that's the bare minimum? Does size of the ships involved matter below this point? Ex. If one has a 0.01 relationship, and kills a ship with 10000 crew, that still won't affect the reputation?

NomadicVolcano:
This belongs in an issue, not a PR.

Issue #5013, opened 2020-04-24
Continuous Build Crashing When Opened (Mac)


Rob59er:
MacOS 10.14.16

Downloaded the continuous build off of GitHub, upon opening the .app it promptly crashed without even showing the loading screen. Downloaded the game again and met the same errors. Error is as followed: 
> Dyld Error Message:
  Symbol not found: _objc_opt_respondsToSelector
  Referenced from: /Users/USER/Desktop/Endless Sky.app/Contents/MacOS/../Frameworks/libSDL2-2.0.0.dylib (which was built for Mac OS X 10.15)
  Expected in: /usr/lib/libobjc.A.dylib

Here is the details of the thread that crashed:
> Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   dyld                          	0x00000001117183ae __abort_with_payload + 10
1   dyld                          	0x0000000111717ba0 abort_with_payload_wrapper_internal + 82
2   dyld                          	0x0000000111717bd2 abort_with_payload + 9
3   dyld                          	0x00000001116d7a9d dyld::halt(char const*) + 343
4   dyld                          	0x00000001116d7bc7 dyld::fastBindLazySymbol(ImageLoader**, unsigned long) + 167
5   libdyld.dylib                 	0x00007fff74c4a32e dyld_stub_binder + 282
6   ???                           	0x0000000107704140 0 + 4419764544
7   libSDL2-2.0.0.dylib           	0x00000001076a23ad SDL_CreateWindow_REAL + 981
8   Endless-Sky                   	0x00000001072e07c7 GameWindow::Init() + 343
9   Endless-Sky                   	0x000000010735a710 main + 1312
10  Endless-Sky                   	0x000000010729c934 start + 52

Reinstalled SDL2 with no success.
As for reproduction I have no advice.
If you need further information or details from the error file feel free to ask.

starmlerp:
..Yeah, the continuous builds are... continuous builds. Even the disclaimer states that it may not work or malfunction in any way. Unless this keeps happening when the official release comes out, I would recommend you to close this issue.

tehhowch:
Without a Mac to work from, this will probably stay open a while.

MCOfficer:
Can you give more information regarding your system? macOS version, stuff  installed via homebrew (if any), you know the drill.

MCOfficer:
I actually did get a similar report from CallMeShmae some time ago:
```
Dyld Error Message:
[...]

PR #5021, opened 2020-04-26
feat: Also build CD on PRs

MCOfficer:
**Feature:** This PR implements enables CD builds on Pull Requests, so people can download and playtest PRs without having to compile.

## Feature Details
With this change, the CD workflow will run,
1. when a new commit is pushed to the master branch
2. when a PR is opened
3. when new commit are added to a PR

In the case of 1), artifacts will be published to the `continuous` release. For PRs, artifacts can be downloaded from the "Checks" tab:
https://gfycat.com/annualtintedcopepod

The download is still only available for logged-in users: https://github.com/actions/upload-artifact/issues/51

## Testing Done
Tested on a multitude of PRs. As far as i can tell, this change needs to be in the head branch, so this will not work for old PRs until they merge master.

## Follow-up features:
Post a comment on the PR with links to the assets. Unfortunately, Github's Actions API is still somewhat immature, so this would require querying and parsing the Github API. There is a nice action for "Sticky" comments, though: https://github.com/marocchino/sticky-pull-request-comment

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -4,6 +4,10 @@ on:
   push:
     branches:
       - master
+  pull_request:
+    types: 
+      - synchronize # New commit has been pushed
+      - opened
 
 jobs:
 
@@ -114,7 +118,8 @@ jobs:
           name: ${{ env.OUTPUT }}
           path: ${{ env.OUTPUT }}
 
-  process_artifacts:
+  upload_artifacts_to_release:
+    if: github.event_name == 'push'
     runs-on: ubuntu-latest
     needs:
       - ubuntu_x86_64
@@ -209,4 +214,4 @@ jobs:
             --replace \
             --name ${{ env.OUTPUT_MACOS }} \
             --file ${{ env.OUTPUT_MACOS }}
-      
+    
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -11,7 +11,7 @@ on:
 
 jobs:
 
-  ubuntu_x86_64:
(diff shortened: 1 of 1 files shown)


MCOfficer:
Here's some more incentive: (If a PR includes this change,) everyone can playtest it with minimal effort.

 
I already implemented this into the [ESLauncher2](https://github.com/EndlessSkyCommunity/ESLauncher2) prototype. It uses a [workaround](https://endlesssky.mcofficer.me/artifacts.html) to download artifacts despite not having a Github account:
ESLauncher2-PR-Demo

tehhowch:
Sounds reasonable. Since we'll be doing more on each commit there are some things we should also do:
1. shared prefix (e.g. `cd` or `pkg`) for CD jobs
2. indicate compatibility (e.g. MacOS 10.15) where needed.
3. Cancel pending / running workflows whenever a new commit is added to a PR, to avoid redundant work. I think this will require using the REST API: https://developer.github.com/v3/actions

MCOfficer:
I don't get 1). The workflow is named "CD", and all jobs reference their workflow.
image

2) i agree with, but i'm not sure where to put in a PR. It's easy once actions can post to the PR, but as i said, actions/upload-artifact#51. We can at least mark it on the releases page, but that has little to do with this PR (rather #5013).

3) Is reasonable.

tehhowch:
I suggested 1 due to the mobile page view: 
<details><summary>screen cap</summary>


Screenshot_20200427-140010
</details>

For 2 I was thinking to just name the output accordingly.

tehhowch:
We can add run cancelling as a follow-up.

PR #5027, opened 2020-04-30
Missile Tweaks

ScruffyKitty:
This is a tweaking of how both early game and late game missiles work, making early game missiles more practical, and giving underperforming late game missiles boosts where they need them. 

Changes:

All missile racks and box given double storage capacity. (I'm counting gatling gun as a missile launcher.)
All human missiles, barring the typhoon and the torpedo, have 1/2 price.
Typhoon damage increased from 880 to 920 and missile strength increased form 40 to 60.
Heavy rocket hull damage reduced from 720 to 460 and description tweaked.
Korath piercers piercing upgraded form 0.2 to 0.6.
Korath mines given a slow value of 2.
Thunderhead missiles missile strength greatly increased (12-40) before the casing breaks off.
Korath missiles available at Korath outfitters.

### DIFF
--- a/data/hai/hai outfits.txt
+++ b/data/hai/hai outfits.txt
@@ -193,7 +193,7 @@ outfit "Tracker Storage Pod"
 	thumbnail "outfit/hai tracker storage"
 	"mass" 4.4
 	"outfit space" -10
-	"tracker capacity" 28
+	"tracker capacity" 56
 	ammo "Hai Tracker"
 	description "The Tracker Storage Pod is used to store extra ammunition for Tracker Pods."
 
--- a/data/human/weapons.txt
+++ b/data/human/weapons.txt
@@ -614,7 +614,7 @@ effect "flamethrower hit"
 
 outfit "Meteor Missile"
 	category "Ammunition"
-	cost 500
+	cost 250
 	thumbnail "outfit/meteor"
 	"mass" .2
 	"meteor capacity" -1
@@ -627,7 +627,7 @@ outfit "Meteor Missile Box"
 	thumbnail "outfit/meteor storage"
 	"mass" 2
 	"outfit space" -5
-	"meteor capacity" 15
+	"meteor capacity" 30
 	ammo "Meteor Missile"
 	description "The Meteor Missile Box is used to store extra ammunition for Meteor Missile Launchers."
 
@@ -677,7 +677,7 @@ effect "meteor fire"
 
 outfit "Sidewinder Missile"
 	category "Ammunition"
-	cost 900
+	cost 450
 	thumbnail "outfit/sidewinder"
 	"mass" .2
 	"sidewinder capacity" -1
@@ -689,7 +689,7 @@ outfit "Sidewinder Missile Rack"
 	thumbnail "outfit/sidewinder storage"
 	"mass" 2.4
 	"outfit space" -7
-	"sidewinder capacity" 23
+	"sidewinder capacity" 46
 	ammo "Sidewinder Missile"
 	description "The Sidewinder Missile Rack is used to store extra ammunition for Sidewinder Missile Launchers."
 
@@ -740,7 +740,7 @@ effect "sidewinder fire"
 
 outfit "Javelin"
 	category "Ammunition"
-	cost 50
+	cost 25
 	thumbnail "outfit/javelin"
 	"mass" .04
 	"javelin capacity" -1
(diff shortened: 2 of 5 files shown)


Zitchas:
I like the decrease in price, but I'm not sure about giving everything double capacity. They are already numerous enough to be hard to outlast - twice as many will make them feel almost impossible to outlast.

That is an interesting effect for the mines, though. That's an interesting application.

ScruffyKitty:
> I like the decrease in price, but I'm not sure about giving everything double capacity. They are already numerous enough to be hard to outlast - twice as many will make them feel almost impossible to outlast.

If it is too much of a problem, switch it to decreasing the outfit space taken up by the racks might be better. The issue is that carrying a significant stock of missiles takes up too much outfit space for the player, compared to other weapons.

Amazinite:
> The issue is that carrying a significant stock of missiles takes up too much outfit space for the player, compared to other weapons.

That's literally the whole point of secondary weapons; high DPS for their size at the cost of limited firepower. While reducing the cost of ammunition and increasing the amount of ammo that can be held by a factor of 2 for each doesn't entirely remove that downside, it does severely weaken it.

ScruffyKitty:
> > The issue is that carrying a significant stock of missiles takes up too much outfit space for the player, compared to other weapons.
> 
> That's literally the whole point of secondary weapons; high DPS for their size at the cost of limited firepower. While reducing the cost of ammunition and increasing the amount of ammo that can be held by a factor of 2 for each doesn't entirely remove that downside, it does severely weaken it.

It makes them impractical and obnoxious to use compared to other early game weapons.

Arachi-Lover:
on discord I've already voiced concerns about the missile strength, but just to also have them here, those are very substantial buffs to the Typhoon and Thunderhead, and could very well unbalance chunks of storylines. if the intent is to make the missiles better for early game, the reduced cost  for human ones already accomplishes that pretty well I'd say, anyone seeing alien missiles most likely is either a veteran with ways of getting money very easily, or just already has amassed a large wealth by that poin; also I agree that the increased capacity added by the racks is too much

ScruffyKitty:
Testing the changes: the missile rack capacity was too much of an increase. The heavy rocket nerf works as intended (two heavy rockets leave a sparrow almost dead, rather than dead).

Zitchas:
I'm dubious about testing the heavy rocket against the sparrow. Heavy rockets are not intended to be fighter hunting weapons; but rather heavy munitions for use against capital ships. Hitting a sparrow with them should be rather like hitting [a torpedo boat](https://en.wikipedia.org/wiki/File:USS_PT-105.jpg) with a [Fritz-X](https://en.wikipedia.org/wiki/Fritz_X) (anti-capital ship missile). As in, no contest. The ship would be obliterated. 

Javelins, on the other hand, are the dumb-fire weapon designed to launch en-mass for dealing with fighters. Javelines are the ones that small ships should be able to take a few hits from; and for that matter, should be able to carry a lot, too.

Amazinite:
> It makes them impractical and obnoxious to use compared to other early game weapons.

Again, that's the point that they can be more difficult to use than normal weapons with no ammunition requirement. You're trading MUCH stronger firepower for the downside of not being able to use it for extended periods of time, and needing to pay to reload. Just compare the missiles to similarly sized, or even larger, guns. A Meteor Missile Launcher, for example, is only 10 tons, while a Heavy Laser is 15 tons. But despite the Heavy Laser being 50% larger than the Meteor Missile Launcher, the Meteor only deals about 17% less hull DPS, while having **over 90% more shield DPS**. That's on top of the fact that secondary weapons have minuscule energy and heat costs compared to other weapons, meaning that using secondary weapons frees up your loadout to perhaps be able to spare better engines or shield generators. 

By drastically reducing the barrier to entry for secondary weapons without doing anything else, we'd be making them *even more* attractive than they already are, making normal guns even less attractive than they can be. The only way these changes would make sense to me is if we also reduced the damage that these missiles do, but at that point we might as well just revert back to before #3866.

ScruffyKitty:
> I'm dubious about testing the heavy rocket against the sparrow. Heavy rockets are not intended to be fighter hunting weapons; but rather heavy munitions for use against capital ships.

The description itself explicitly says it's good against swarms of fighters.

ScruffyKitty:
The price changes aren't having a significant impact on the game.

Zitchas:
Well, the current heavy rocket (which, as you point out, obliterates a sparrow) is considerably more effective than the revised version that takes two hits to disable it... To go back to my example, the Fritz-X is *very* effective against torpedo boats.... when it connects. They generally could only carry a few per plane (and often just one), and thus couldn't litter the sea with them, which is why they were primarily used for killing capital ships. Not because they "aren't effective" against the smaller ships, but because they can't be fielded in big enough quantities to be useful in saturating an area to hit small and agile targets. 

It's the same case with the Heavy Rocket - It is lethal if it connects, and has an area effect so it is very effective against groups if it manages to connect with anything. And hitting a *group* is much easier than hitting an individual.

Issue #1998, opened 2017-01-11
Outfit Variants


Amazinite:
Same thing as ship variants.

I can't find this having been brought up before, but if I somehow missed something I'll just close this.

toilethinges:
#927

Amazinite:
K

PR #5042, opened 2020-05-06
New jump drive effect and sound

Kryes-Omega:
## Summary
I changed the jump drive effect and sound. I hated the white noise and didn't like the boring purple fuzz. Now the effect is _**d y n a m i c**_ and moves in the direction the ship is moving. I did not include the absolutely wild change of increasing jump speed for jump drives, but would like to. Reason: its basically a scram drive after the first stop anyway, why not

[Anyway, behold](https://streamable.com/zfahvo)

## PR Checklist
 - [x] messed up with github a few times
 - [ ] convince people to increase jump speed for the jump drive
 - [x] noticed by papa derpu
  
## Performance Impact
idk a few more mb in the file size.

## Closing Statement
add support for jump drive and hyperdrive variants please i beg you


Zitchas:
1) The sounds effect is great. I really like the lower pitch and tone. That being said, It probably needs to be a little louder. It doesn't need to be as loud as the current hyperdrive or jump drive effects, but it should be close. Right now I need to boost my volume enough to make current sound effects unpleasant to listen to in order to really listen to this new sound.

2) The red visual effect is a good start, but I think it needs some improvement. One change that I think improves it is if the red aura comes closer to actually obscuring the ship at the height of the "jump", like the current JD effect.

ravenshining:
I agree with Zitchas on both points, but also:

3. It looks really weird to me to have the effects going in the direction you have them. I'd expect them to be going in the opposite direction, or some direction correlated with the direction of the jump.

4. While this effect is awesome, I like the old effect, too.  Can we, say, keep the old effect for Quarg & Pug, and have this effect for Korath & Hai?  That would necessitate creating a new outfit, but offers an opportunity for differentiation - perhaps the Korath jump drive is less efficient but allows for more movement?

5. If possible, the B5 fan / physics nerd in me would love to see the effects be red on jumping out like you have it, and switching to blue when jumping in - that is, redshift for things leaving the system, blueshift for things arriving in system.

Kryes-Omega:
From @Zitchas 
>     1. The sounds effect is great. I really like the lower pitch and tone. That being said, It probably needs to be a little louder. It doesn't need to be as loud as the current hyperdrive or jump drive effects, but it should be close. Right now I need to boost my volume enough to make current sound effects unpleasant to listen to in order to really listen to this new sound.

>     2. The red visual effect is a good start, but I think it needs some improvement. One change that I think improves it is if the red aura comes closer to actually obscuring the ship at the height of the "jump", like the current JD effect.

Volume should be easy to change. As for the visual, I can scale up the frames to cover more of the ship. 

From @ravenshining:

>     1. It looks really weird to me to have the effects going in the direction you have them. I'd expect them to be going in the opposite direction, or some direction correlated with the direction of the jump.
> 
>     2. While this effect is awesome, I like the old effect, too.  Can we, say, keep the old effect for Quarg & Pug, and have this effect for Korath & Hai?  That would necessitate creating a new outfit, but offers an opportunity for differentiation - perhaps the Korath jump drive is less efficient but allows for more movement?
> 
>     3. If possible, the B5 fan / physics nerd in me would love to see the effects be red on jumping out like you have it, and switching to blue when jumping in - that is, redshift for things leaving the system, blueshift for things arriving in system.

1. Just having the effects trail behind the ship seems a bit boring to me. From the front feels more like, oh the jump drive is doing something to compensate for the ship's velocity. Its a teleportation device not a go very fast device like the hyperdrive. I don't plan on changing it unless everyone screams at me at once that its awful. As for correlating with direction of jump, I would like that too, but sadly... not possible.

2. #5044. 

3. Not possible. Also, not really physically accurate to my knowledge.

Not sure what changes @Amazinite is looking for exactly in this PR, nobody has said anything about balance or usage of the new jump drive since we're shifting to plan B, aka "use them both."

Considering the Korath just steal jump drives (unless lore changed at some point and I wasn't paying attention) and neither the Quarg or Pug have a use case for a scram jump drive, dunno. Maybe just swap the Pug's jump drives for scram jump drives? That would probably involve changing the amount of fuel their ships carry or something.

beccabunny:
The korath (exiles) manufacture their own jump drives, so it would make sense to use this effect there

Amazinite:
Even with #5044, I don't see myself merging this. The visuals look great to me, but the sound is meh. Far too quiet and subtle I think compared to the current sound, which is very noticeable. In so far as the idea of multiple drives go, we'd still need #1998/#927 after #5044, lest we run into the issues outlined in #3534/#1455.

ravenshining:
I *really*, *really* like this sound.  And the way it kinds sounds like something is spooling up, rather than accelerating, more befits a jump drive than a hyperdrive to me.  I get where the "boom" sounds like a hyperdrive though.

Maybe if that twangy cyclic sound was amplified, and the boom effect dulled, it would work?  Like here's it after just running it through a compressor, if you have a mutil-track source to play with the different effects that would be a lot easier to deal with:
[jump drive.wav.zip](https://github.com/endless-sky/endless-sky/files/4607168/jump.drive.wav.zip)

Darcman99:
This hits me as more of an alternative jump drive, probably in part because of the dramatic change in visual and sound from the original. It would also be a shame to lose the original that we've had for so long..

PR #5044, opened 2020-05-07
Customizable jump drive and hyperdrive sounds and effects

Amazinite:
**Feature:** This PR implements a feature request discussed on Discord in reference to #5042

## Feature Details
Adds the ability to customize the sound and effect that a jump drive or hyperdrive has, instead of all jump drives and hyperdrive needing to use the same sounds and effect.

Adds the following new attributes:
* `"jump effect" <effect name>`: The effect that is created over a ship when it is jumping. This is an effect, not simply a sprite, so the effect must be defined elsewhere.
* `"jump sound" <sound name>`: The sound made when the player's flagship jumps. 
* `"jump in sound" <sound name>`: The sound made when other ships jump into the player's system.
* `"jump out sound" <sound name>`: The sound made when other ships jump out of the player's system.
* `"hyperdrive sound" <sound name>`: The sound made when the player's flagship hyperjumps. 
* `"hyperdrive in sound" <sound name>`: The sound made when other ships hyperjump into the player's system.
* `"hyperdrive out sound" <sound name>`: The sound made when other ships hyperjump out of the player's system.

If none of the above attributes are defined, the game reverts to the default sounds and effect. (This is why the jump drive, hyperdrive, and scram drive are unchanged in this PR. Although come to think of it, the scram drive could use a unique sound.)
If multiple drives are installed, each with their own custom sounds or effect, then the sounds and effects of each drive are created. 

## UI Screenshots
N/A

## Usage Examples
The following could be an example of the syntax for the jump drive in #5042.
```
outfit "Scram Jump Drive"
	"jump speed" 1
	"jump fuel" 200
	"jump drive" 1
	"jump effect" "scram jump drive"
	"jump sound" "scram jump drive"
	"jump in sound" "scram jump drive in"
	"jump out sound" "scram jump drive out"
```

## Testing Done
The following video depicts three Albatrosses, one with a normal JD, one with a "scram JD" using the sounds and effects from #5042, and one with both.
https://drive.google.com/open?id=1X-dBfNPUeZNFO6nBi5WrTLF31qq5YWg1

## Performance Impact
Each ship when jumping now needs to check if they have any custom sounds to be played or effects to be created. If not, revert to the default, as opposed to previously where it was a simple check of if the jump or hyperdrive sound should be used.
Given that ships are not jumping constantly though, I expect no noticeable impact on performance from these extra computations.
The amount of particle effects created when you have multiple jump drive effects is the same as if you were to only have one, as the amount of each effect produced is divided by the number of effects. This means that having multiple jump drive effects shouldn't cause extra stress on the game when compared to only one jump effect.

### DIFF
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -127,6 +127,20 @@ void Outfit::Load(const DataNode &node)
 			++steeringFlareSounds[Audio::Get(child.Token(1))];
 		else if(child.Token(0) == "afterburner effect" && child.Size() >= 2)
 			++afterburnerEffects[GameData::Effects().Get(child.Token(1))];
+		else if(child.Token(0) == "jump effect" && child.Size() >= 2)
+			++jumpEffects[GameData::Effects().Get(child.Token(1))];
+		else if(child.Token(0) == "hyperdrive sound" && child.Size() >= 2)
+			++hyperSounds[Audio::Get(child.Token(1))];
+		else if(child.Token(0) == "hyperdrive in sound" && child.Size() >= 2)
+			++hyperInSounds[Audio::Get(child.Token(1))];
+		else if(child.Token(0) == "hyperdrive out sound" && child.Size() >= 2)
+			++hyperOutSounds[Audio::Get(child.Token(1))];
+		else if(child.Token(0) == "jump sound" && child.Size() >= 2)
+			++jumpSounds[Audio::Get(child.Token(1))];
+		else if(child.Token(0) == "jump in sound" && child.Size() >= 2)
+			++jumpInSounds[Audio::Get(child.Token(1))];
+		else if(child.Token(0) == "jump out sound" && child.Size() >= 2)
+			++jumpOutSounds[Audio::Get(child.Token(1))];
 		else if(child.Token(0) == "flotsam sprite" && child.Size() >= 2)
 			flotsamSprite = SpriteSet::Get(child.Token(1));
 		else if(child.Token(0) == "thumbnail" && child.Size() >= 2)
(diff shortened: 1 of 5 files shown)


kestrel1110:
Nice! Hope it gets in!

WarlordMike:
The current sounds get old after a while. Won't lie, this would be excellent to have.

Would it be possible to also have a "sound plays twice as fast in 2x speed mode?" Having the warp sound not synced to how fast the game is actually playing annoys me a bit.

Amazinite:
> Having the warp sound not synced to how fast the game is actually playing annoys me a bit.

I guess my mind has completely ignored this fact about the current sounds, but I did notice that when testing the sound from #5042. That's a whole issue of its own (that someone should probably post an issue for).

source/Outfit.cpp
```diff
@@ -127,6 +140,20 @@ void Outfit::Load(const DataNode &node)
 			++steeringFlareSounds[Audio::Get(child.Token(1))];
 		else if(child.Token(0) == "afterburner effect" && child.Size() >= 2)
 			++afterburnerEffects[GameData::Effects().Get(child.Token(1))];
+		else if(child.Token(0) == "jump effect" && child.Size() >= 2)
+			++jumpEffects[GameData::Effects().Get(child.Token(1))];
```
petervdmeer, line 144: Did you also test what happens if the referenced effect or sound is not present? (I don't expect anything bad to happen; just an empty effect to be played, but let's test this just to be sure.)
Also for the effects below.

Amazinite: Tested giving the JD `"jump effect" "this effect does not exist"`. 
When jumping, the game didn't crash or anything, there was just no visible effect. An error was also logged in errors.txt:
`Warning: effect "this effect does not exist" is referred to, but never defined.`

source/Outfit.h
```diff
@@ -81,6 +81,14 @@ class Outfit : public Weapon {
 	const std::map<const Sound *, int> &SteeringFlareSounds() const;
 	// Get the afterburner effect, if any.
 	const std::map<const Effect *, int> &AfterburnerEffects() const;
+	// Get this oufit's jump effects and sounds, if any.
```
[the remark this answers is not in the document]
Amazinite: Added a comment above the variables that the integer is the number of sprites/effects/sounds.

PR #5048, opened 2020-05-08
Fix spelling in Intro Misions

Zitchas:
This is simply a spelling correction, found in the `intro [2 interceptor]`

`recgonize` -> `recognize` (line 385, near the end of the paragraph)


tehhowch:
Only features should be pointed at feature branches, btw.

Zitchas:
So fixes like these should point to the master, then?

PR #5054, opened 2020-05-10
Random lifetime for effects

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed ~~in issue #{{insert number}}~~ no where.

## Feature Details
Allows effects to have a random lifetime akin to the random lifetime on projectiles, and to match the fact that they already have random velocity. 

## UI Screenshots
N/A

## Usage Examples
```
effect "jump drive"
	sprite "effect/jump drive"
		"no repeat"
		"frame rate" 12
	"lifetime" 10
+	"random lifetime" 200
	"random angle" 180
	"velocity scale" 0.
```

## Testing Done
Give a random lifetime to the jump drive effect, causing the jump drive particles to decay at random intervals. 
The jump drive effect after moving away from where I just jumped in. A high random lifetime leaves many particles remaining from where I just was.
image
A short time later, only a few particles are left.
image

## Performance Impact
N/A

### DIFF
--- a/source/Visual.cpp
+++ b/source/Visual.cpp
@@ -24,6 +24,9 @@ using namespace std;
 Visual::Visual(const Effect &effect, Point pos, Point vel, Angle facing, Point hitVelocity)
 	: Body(effect, pos, vel, facing), lifetime(effect.lifetime)
 {
+	if(effect.randomLifetime > 0)
+		lifetime += Random::Int(effect.randomLifetime);
+	
 	angle += Angle::Random(effect.randomAngle) - Angle::Random(effect.randomAngle);
 	spin = Angle::Random(effect.randomSpin) - Angle::Random(effect.randomSpin);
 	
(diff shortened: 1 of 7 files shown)


source/Visual.cpp
```diff
@@ -24,6 +24,9 @@ using namespace std;
 Visual::Visual(const Effect &effect, Point pos, Point vel, Angle facing, Point hitVelocity)
 	: Body(effect, pos, vel, facing), lifetime(effect.lifetime)
 {
+	if(effect.randomLifetime > 0)
+		lifetime += Random::Int(effect.randomLifetime);
```
tehhowch, line 28: If the random lifetime is 1, then the lifetime is never increased (since random::int in [0, 1) is always 0). This is why the random lifetime for a projectiles draws from random::int in [0, lifetime + 1).

PR #5056, opened 2020-05-12
Polished "Wanderers: Defend Vara Ke'sok Hint"

Anarchist2:
This issue arose during a discussion on the discord server. I decided to fix up the mission so that players who rush to Wanderer space after getting a JD don't get called a hero even if they haven't done anything. Plus a bunch of other changes


Zitchas:
This looks like a good adjustment, although modifying it with another branch so that the player *does* get the original statement if they have completed the main human campaign would make it even better.

MasterOfGrey:
I am in favour of this change, and disagree that the original text should be brought into any branch. The player's reuptation in human space is ultimately irrelevant to the Wanderer's and has little to no bearing so soon after meeting them.

Issue #4791, opened 2020-02-06
Kor Sestor, Kor Merety and other automata capture.


WildLynxDev:
Problem: I feel that there is a plot hole here.

What exactly happens when you capture a robotic ship? There is zero crew capacity and no human controls. That's why "Command Center" outfit exists. 
But why captured ships follow your orders before it is installed and captain hired?
I think repairing a disabled robotic ship should cause it to become angry again if you don't looted remote control or reasoning node, or leave them permanently crippled if you did.

Possible solutions:
1) Make robotic and automated ships uncapturable. Simple, but not fun.

2) Add ability to plant outfits. Can lead to unforeseen consequences. Planting Command Center or crew bunks also transfers some of your crew to the disabled ship once it captured and not blown up.
Or, better, planting Command Centre and may be also Crew Bunks happens in separate menu available when you successfully captured the automated ship.

3) Add ability to tow disabled ships, and dock them to the stations. Landing on the planet with hunk of inert metal on your tow is kinda unrealistic. Problem: some stations have no outfitters. Also new physics - ships on the tow behave like inert mass on the tow and it's probably difficult to implement.

4) Some as 2, but add Temporary Remote Control outfit. Ship with TRC can only follow player and ignore all commands and can not use any active abilities, like weapons. Additionally, if player uses Jump Drive and ship with TRC cannot directly follow, because of no hyperspace link here, hyperspace unit looted or no fuel, ship becomes disabled again, but with no ca-boom on attempt to revive it. 
May be TRC is automatically planted into captured robotic ship if you have it in the cargo. Probably works by removing automaton flag and adding dumb flag. Also drones and fighters with TRC can dock to player's ship if there is slots available.

Zitchas:
Options 1, 2, and 3 all sound good to me. 

I also dislike the fact that it is so easy to capture automata ships without any kind of prep. Even we simply made automata ships capture into the "dumb" mode automatically, then require either the control node or command center in order to get them to work normally, that'd be acceptable (and take a little bit more work)

petervdmeer:
The current solution is that robotic ships have a high chance of self-destructing when you try to capture them. (The more valuable the ship, the higher the chance on self-destruct.)
Solution 1 mentioned here is also possible in the current game and some ships do use the uncapturable flag.

petervdmeer:
[...]

Issue #3638, opened 2018-02-09
self destruction is not fun and frustrating. My solution/idea to ease the problem 


meiwud334:
I think the self destruction in the game to balance capping the drones is introducing grind and no fun gameplay. it feels arbitraty and unfun being fully reliand on RNG.

Solution I would like to discuss: Give the player some way to influence the destrcution rates and reduce exposure to extreme RNG.

Effreti could sell a special equipment: (something like Anti Selfdestruction Hacking Drone) which reduces the propability of self destrcution by C/(target mass) when used on boarding.
Example Numbers: outfitt space 3 Tons. Prize 5 million
Description: the bigger the target the less effect
This equippment could be unlocked by story, eg wanderer or just be sold from start by them (which i prefer as they currently have only the command center as unique item of interest).

when boarding a korath drone (or any ship with self destrcution rate <1) a new window will pop up: do you want to spend your Hacking Drones. yes or no
if yes how many:  chosse between 0-min (#installed,#minimum number needed to achieve 0 self destruct rate)

n=#Drone will be used up (removed from outfits installed) and the self destruction rate will be set to
new self destruction rate = old safe destruction rate -C x n/(target mass) >=0

Using Mass is a dirty quick fix to approxiamte how rewarding a ships is to cap and therefore how expensive it should be in terms of needed drones to be balanced (a 512 would need more drones than a model 8 to get perfect 0% blow up chances, this balances the holw drone thing).

Better: Alternatively to mass one could alter the self desctrution attribute to allow a second argument
"self destruct" .65 .1
the second number would tell how much one single drone would reduce the self destruct rate when used.
if left blank it is set to 0 and drones can not be used (no pop up window on boarding)

lets take a 512 as example with the new value set to .05 and the orginal self destruct rate
"self destruct" .80 .05
the player would need to use 20 drones (worth 20x5mil=100 mil) (20x3=60 outfit space) to achieve a self destruct rate of 0 and after that he needs to go and get new drones as these 20 are used up.

Lets discuss the andavtages and disadvantages of this idea:
pro: 
1.less frustrating RNG,
2. players than increase their chances by using proper equippment (buy with money), 
3. should be easy to programm, the players must not grind capping drones but has to grind money instead.
4. if the player feels like playing risky he can use fewer drones to reduce the explosion risk with some risk remaing with the consequeence that if it still blows up he lost some drones.
5. players that dont need to buy drones at all and can continue using RNG as is

con: 
[...]

PR #5057, opened 2020-05-12
Automata costs

Zitchas:
## Summary
One of the main problems with the automata is that they break the game economy. Namely, they are free. Once you get a 512, you've got it for free for life. Every other ship in the either requires an up front purchase (which can be around a hundred million credits and a long mission chain to unlock for the hurricane, which is the only comparable purchasable ship), or requires a dedicated combat boarding ship with an associated fleet and/or convoluted means of disabling the target ship. In either case, every other ship in the game has both a large amount of effort required to acquire AND requires continuous ongoing costs to keep it. 

In contrast, the automata ships are easy to get (simply wait for a ship to be disabled, which they obligingly do regularly, and hope you get lucky); have no involvement in any story, and require minimal special equipment (only a JD is actually required; although a cloak and better-than-generic ramscoops are very helpful); and then result in a ship that has no ongoing costs ever again while simultaneously being one of the toughest ships in its category in the game and a tier 2 ship at that.

## Changes:
This gives all automata ship a "maintenance cost" roughly equivalent to their comparative power:
- KIV 349: 1100
- TF 71:  700
- TF 78: 600
- TF 109: 500
- MPT 52: 300
- FL 14: 10
- FO 27: 15
- Model 8: 50
- Model 16: 200
- MOdel 32: 400
- Model 64: 600
- Model 128: 700
- Model 256: 900
- Model 512: 1000

Note that this is incredibly cheap compared to other faction's ships.  A fighter (typically 1 pilot, so 100/day is instead 25 a day. Heavy Warships range from 18 to 250 crew, which means 1800 to 25000 credits per day. Currently, the only purchasable ship that is in their same ballpark is the Hurricane, which sits at a minimum of 7600/day. As such, 1100/day is downright cheap, especially considering that it is acquired for free.

My logic is that - being an automated ship - it will have need for additional resources to maintain itself as well as potentially additional maintenance by port facilities. After all, it does not have a large crew maintaining it. Given how they act in space, they appear to be used as disposable swarms to pound enemy ships into oblivion, which suggests a utilitarian and likely disposable attitude. You don't waste resources on long-term maintenance automation if you don't expect the ship to last long enough to need it, after all. Much more efficient to have ground based repair facilities that can specialize in it and that are likewise less likely to be lost in combat. (internal hull repair systems are not the same thing as ongoing general maintenance. Note that hull repair systems *cannot* repair a ship once it has taken sufficient damage to impact internal systems - only the hull) 

In terms of lore, automata already have advanced shipyards that are pretty close to where the combat happens. Such shipyards would be far more efficient at doing repairs than any onboard system could be.

This doesn't completely solve the automata problem, but it at least injects a small dose of realism and means that automata are no longer "acquire and forget" type ships, while still retaining their strong economic advantage.

For contrast, the old "Androids" that could be used to replace crew were deemed to be OP in part because they allowed the player to convert a ship that they had purchased into an automated ship by buying large numbers of them. If gaining a no-daily-cost ship (among other things) was ruled OP when it could be done through the expenditure of of hard-earned resources (Androids cost 80000), how much more so is it when the entire thing - the ship AND the automation can both be acquired at the cost of just a little patience?

## Alternatives:
While this is a move in the right direction, I think a better (and more permanent) solution lies in future mechanics changes that render the automata capture process a bit more involved. The main problem I have with it is that it seems really unreasonable that an average human captain and (maybe) a few supporting crew with zero training or experience are capable of reverse engineering the AI for an advanced alien warship to the degree necessary to establish the Captain as the ship's commanding authority, and override any previous commands, loyalties, and programming. 

**A)** Ideally, I'd like to see some kind of storyline that involves bringing numerous "Control Tranceivers" and/or "Reasoning Nodes" back to some group, leading to a bunch of missions lead to the unlocking of a device that - so long as the flagship has it outfitted - enables the flagship to take control of automata ships. Without the device, they can board and loot automata ships but not capture them. The Wanderers would be a logical candidate, as they already have some missions surrounding understanding the automata. These could also be good missions for the Kor Efret, as they have the most experience at dealing with them. And of course, the Remnant are the legends of reverse engineering and salvaging things, so there's an obvious possibility there. 

**B)** Alternatively, the addition of some kind of "in-space outfitting" where certain types of  outfits could be installed into ships that the player is currently boarding, which would allow players to forcefully install command centers into a ship. Requiring the player to carry a command center with them and install it into an automata ship in order to capture it would also be a suitable lore (and challenge) for the player to overcome. Alternatively, missions could lead to the creation of some kind of modified Reasoning nodes or control tranceivers; and the player would need to bring some of these alternate versions along to use to replace the default ones in the ship. Tearing out the old control system and replacing it with a new one would definitely be appropriate for the lore and technical abilities of <first> and their crew (if any).

That being said, even if such mechanic(s) is(are) added, I don't think that the costs listed here are a problem, as nothing changes the base logic behind them and they remain sufficiently low (and the ships sufficiently powerful) that automata ships remain a very attractive addition to fleets.

## Related Issues/PRs:
https://github.com/endless-sky/endless-sky/issues/4591  (this one is the most relevant)
https://github.com/endless-sky/endless-sky/issues/4791
https://github.com/endless-sky/endless-sky/issues/3638


Arachi-Lover:
I don't object to the maintenance costs on their own, but imo its best to have them different for each ship, rather than grouping up classes together. IE, the 512 and 349 are "matching" Heavy Warships, but like that one issue brought up, the 512 is much much harder to come by, it is a rarer ship, and more often than not the 349 is considered superior. 512 having less maintenance costs than the 349 could help make it more desirable. likewise, the Sestor Carrier, on its own, is complete garbage, while the other 2 Medium Warships they have are pretty darn good on their own, so would also have different maintenance, and so on.

lore-wise, I'd be asking who the hell I'm paying for all this maintenance, especially when I just captured the ship and am jumping to the Efreti systems, but its not big a problem.

that's where I'd draw the line though, adding maintenance costs, and nothing else to alter Automata ships capping ordeal. with the exception of veteran players that know what they're doing, and deliberately want to get a 349 to steamroll FW in for fun, who would even manage to get to Korath space early? you need to know what you're looking for, otherwise you're the extremely rare oddball that happened to get a JD on accident, and managed to also on accident find Korath space and capture an Automata ship. so basically, I disagree with the Alternatives, I don't think they're needed. if the point is to make it so the player can't get a powerful ship before a certain point, then we also need to remove the ability to: use the Wardragon exploit to capture a Punisher, use the Wardragon exploit to farm Quarg rep and steal a Skylark (doable before even FW due to Rim archeology), steal the Arfecta. at that point, we might as well just make JDs unplunderable. if the point is to try and make sense of how the player can hack an automata, should we also do that for the cases of where the player steals any alien ship, without having previously purchased a ship from the same faction to "get the manual"? how would we know how to fly Hai, Korath, Pug, Wanderers, Quarg, all 3 Coal species + Heliarch ones, since we would not know the controls?

both options to me seem like they just restrict the freedom of playing through the game, and balance is a joke of an excuse when any one of us can just farm six billion sparrows and steamroll everything within a year of leaving New Boston, so for something that is already a pain to many people, since they have to rely on RNG, I don't see a need to make the capturing process any more tedious, or artificially lock it behind an outfitsomethingsomething.

Quantumshark:
I would say these should be "operating costs" rather than "maintenance costs, so the cost does not apply if the ships are parked. There's no obvious reason why these would need a maintenance cost while parked in a spaceport any more than a ship that usually uses crew, because parked ships do not have crew assigned to them.

Zitchas:
Thanks for the feedback. The costs are definitely adjustable. I'll make some tweaks; but suggestions would be appreciated. 

As for the "Alternatives" section: Those were alternate ideas that I'd considered. Personally, my preference would be for "All of the above." As in multiple different ways of doing it.
1) Right from day 0 the player could go buy a command center, find an automata, rip out the node/transceiver, stick in the command center, and go on their merry way (with the exception that it's not a pure automata ship - it has at least 1 crew). 
2) Then have the ability to get them as full automatons through some kind of hacked node/transceiver that one can acquire (probably from a Kor Efreet mission set that is *not* storyline dependent)
3) Then have the ability to hack them and gain their loyalty through some kind of specialized hacking outfit that the flagship carries. (probably a Remnant thing)
4) After the relevant Wanderer mission where they gain an insight into the Automata AI, the player can do direct hacks the way they currently do. 

Thus it is a scale with lots of alternatives and pros/cons: 
Option 1 requires zero story and can be done right from the start of the game; but the ship won't actually be an automata anymore, but needs an in-flight outfit installed on the target.
Option 2 requires minimal story and could be done without any campaign involvement, and simply swaps out an outfit they already have for an equivalent one provided by the Kor Efreet. Little bit slower, but the ship is actually an automata and doesn't lose space to a command center.
Option 3: takes longer since there'd be a fair number of Remnant missions; but still does *not* require campaign missions. Plus side, no outfits need to be installed on the target. Downside, have to carry a special outfit to do the hacking. Results in a full automata.
Option 4: Takes a lot of campaign missions, but gives the player full ability to take control of automata at a point in the story where it makes sense that they'd be able to do so. (ex. Wanderer Chapter 2).

The point is that it turns the automata from being purely a spoiler ship that veterans use to rofl-stomp the early game into a mysterious ship that they can kind of get initially still, but as story unfolds they get better access to it until eventually they've got full access/control as part of the story.

That basically turns the whole automata situation from a plot hole into a strength. But that being said, none of that is being implemented in this PR. This is purely adding some small costs so that they aren't complete 100% freebies.

As a side note, if the simple fact that they are complete freebies in terms of ongoing fleet maintenance/crew cost is actually the major "selling feature" or benefit; then we should expect that other T2+ ships that still have to be bought and crewed will be comparatively more powerful as a result, since those other ships *won't* have the benefit of low/no crew costs.

Arachi-Lover:
I never really saw it as a plot hole, just another instance of gameplay > realism. Tier 2+ sure, but for Tier 2 ships that aren't automata, like the Derecho and Hurricane, I think they're at a good spot, and don't need buffed, nor stuff like 512/349 nerfed. those ships get beaten by the Hurricane mainly due to matchup, which is the whole point I'd argue. on top of what I said about the alternatives, those 4 steps imo just add too much complexity for ships that won't see any use since everyone will just find it easier and more convenient to buy Hurricanes, especially since money is a nonfactor by the current Wanderers endgame.

on the costs, I have no idea really, but the Sestors Fighter/Drones are so similar (Fighter bieng literally just a larger Drone with one more gun port) that I found it odd the Fighters have a maintenance cost and Drones don't; I'd either have both at a very low value or both without maintenance costs altogether.

Zitchas:
OK. I had them split because human drones have no ongoing costs but fighters do; but putting them both at something low is fine by me.

Those 4 ideas basically presume that the automata are going to figure prominently in at least another chapter worth of story down the road. I suspect that they'll figure fairly highly in Wanderer Chapter III, and I know the Exiles using some form of automata tech will figure in the Remnant storyline. If this really is it for the automata and Chapter 3 doesn't touch them much and they just end up being the "barbarians" of central/Northern Automata space for evermore as more (and equal or higher tier) fill in the galaxy around them - then yeah, all those are probably overkill for not much benefit. 

In some respects, this is probably highlighted a bit by the Ka'het where the automata ships there are just simply "not capturable." Begs the question "why not these ones when we can the other ones?"

jvpoz:
Would there be a way that the cost could be in commodities, not money? You'd loose some ammount of metal, heavy metal, and/or equipment per day per ship instead of a monetary upkeep.

Zitchas:
That would be a good idea, and it would make more sense, too. Unfortunately, we don't have the mechanics for that.

vitalchip:
This and #5059 are. addressing similar issues, namely that it's actually fairly easy to capture these ships. It can be very easily done with a Starling but even a fast interceptor (and some luck) can capture a 349..  What about having the automata crewed by robots?  Then you'd need to fight as usual and it would make more sense that the captured ship follows you as now it has your people on board.  Also addresses ongoing costs as now it needs crew. Bonus is that it is then possible to use a captured ship properly as a flagship and use it to capture others, depending on crew size, obviously. We can keep the reasoning/tranceiver nodes in place (for the missions), just that there would be physical robots on board to carry out the instructions.

Zitchas:
Thanks for the idea, but I don't think that reversing the distinction and returning to a standard combat model (even if it is using robots that don't count as crew for economic purposes) is a good solution. That would make very much like every other ship and more or less eliminate the ability to get them at all in the early game. While I think there need to be some kind of limiting factor (other than accessibility), removing one of their unique attributes isn't a step in the right direction.

vitalchip:
Wasn't an ideal solution but it's easily executed, I did it last time I played the game through via a plugin (I just crewed the ships) and it made it a lot harder but not at all impossible, even early game. Just means you need a decent size ship rather than being able to capture a massive warship with a sparrow.  Right now the only thing stopping any player getting a huge (free) fleet is a jump drive and they're too easy to get as well!  Just for kicks I bought a bounder, put one of. my jump drives in it and went to the right place..  captured a 349 in under 4 minutes in the system ( was lucky, only 5th ship tried ) That's not right, a 1 person low end ship captures one of the most powerful ships in the game and now it does my bidding for some reason!  Bung in a command centre and back I go to capture more, in relative safety now.

Arachi-Lover:
so I didn't know there was a difference between maintenance costs and operating costs (rather, didn't know the latter existed, just figured maintenance costs did what it did). now I'd say these should change to operating costs

Zitchas:
Thanks for the suggestion. I have now swapped out all the `"maintenance costs"` for `"operating costs"`.

MasterOfGrey:
"The main problem I have with it is that it seems really unreasonable that an average human captain and (maybe) a few supporting crew with zero training or experience are capable of reverse engineering the AI for an advanced alien warship to the degree necessary to establish the Captain as the ship's commanding authority, and override any previous commands, loyalties, and programming."

I shall say, I never saw this as the actual process that's happening. These ships are intelligent and at least somewhat autonomous. Even the Sestor have some autonomy, their source of control is just external instead of internal programming. They're all coordinated to work towards the destruction of their stated enemies at all costs, but you aren't strictly their stated enemy, you're just an interloper. You're also a hostile interloper to their stated enemy. Making you the enemy of their enemy.

Their command/control/programming demands that they not fall into enemy hands, but are you strictly their enemy? I see the self-destruct chances as being representative of how much their own intellect has, through their experiences, come to value their own continued existence. Are they willing to overcome the instinct/command to self-destruct to continue existing? The bigger ones are under stricter control/programmed reinforcement.

Brick63:
Perhaps add mechanics where the the addition of the Command center where it removes the ships operating cost. Then balance the costs around that mechanic.

Zitchas:
Yeah, that makes sense. Have the cost associated with keeping highly advanced automation equipment functioning.

Amazinite:
Given that I don't plan to merge this any time soon (and in discussing this there seems to have been better ideas that popped up that just aren't possible or reasonable right now), I'm just going to close it. We should return to the idea behind this once there are more T2 factions in the game with obtainable ships outside of just the Korath and Wanderers.

Zitchas:
OK. I still think they are reasonable changes to make the automata "not-unquestionably-the-best-for-all-situations," and I think this is a good solution until the mechanics for one of the other ideas is developed, but this isn't something I feel strongly about either.

Issue #5058, opened 2020-05-12
Ships can jump when disabled if they started jumping before they were disabled


Rakete1111:
As the title says, any ship (including your ship) that starts the jump sequence before they are disabled can still jump out. This is especially noticeable if the ship has a jump drive. I don't know whether this is intended or not, but it's annoying when almost disabled Korath raiders manage to jump away like that.

**Steps to reproduce the behavior:**
1. Get a ship with a jump drive
2. Stay still in a pirate system
3. Wait for them to attack you
4. When you're almost disabled jump

**Expected behavior**

IMO if the jump drive animation is still playing, then the jump can't actually complete if the jump drive becomes inoperable (due to being disabled) in mid jump.

 - OS: Linux
 - Game Source: self-compiled
 - Version: 0.9.12 / 7623fae5

Amazinite:
This just seems like part of the mechanics behind jumping (in-lore) to me. That is, we've started the jump, now there's no turning back, even if we get hit in the middle of it. Why exactly would being disabled pull you out of a jump? Is it really just because "it's annoying" that we should change this?

Rakete1111:
> Is it really just because "it's annoying" that we should change this?

No of course not.

> This just seems like part of the mechanics behind jumping (in-lore) to me.

I can't agree. During the jump animation the jump drive is using up the 200 fuel. When it reaches 200 the actual jump happens. So, if the jump drive is disabled it can't actually use any more fuel and the jump should fail.

Also, IMO the jump drive animation is just the preparation for the jump. The actual change in systems is instantaneous. If the preparation for the jump is stopped for whatever reason, then you can't actually jump. At least that's how I see it.

starmlerp:
keep in mind that the same happens with the hyperdrive, although it is harder to recreate

Amazinite:
As mentioned on the PR, I think that the current functionality is preferable to the suggest change. There also seems to be a lack of strong feeling for any change in the current system. As such, I've closed the PR, so I think the issue should be closed as well.

PR #5060, opened 2020-05-13
Abort the jump if a ship is disabled mid-jump.

Rakete1111:
**Bugfix:** This PR addresses issue #5058 

## Fix Details

Stop the jump sequence when the ship is disabled while the jump animation is playing. This is mainly done for jump drives, but is also applicable for hyperdrives.

For hyperdrives, this can only happen in the very beginning of the jump, while the ship is accelerating in the origin system. While I haven't tested this for hyperdrives (mainly because of how rare/hard to do it is), the result would be that the ship stops the jump with an increase in velocity.

## Testing Done

Tested this on my main save on Korath raiders. Works as intended, the jump drive animation stops and the Korath raider doesn't jump.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1168,6 +1168,15 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 		// Don't apply external acceleration while jumping.
 		acceleration = Point();
 		
+		// If at some point during the jump process the ship is disabled, abort
+		// the jump sequence.
+		if(isDisabled)
+		{
+			hyperspaceSystem = nullptr;
+			hyperspaceCount = 0;
+			return;
+		}
+
 		// Enter hyperspace.
 		int direction = hyperspaceSystem ? 1 : -1;
 		hyperspaceCount += direction;
@@ -1181,15 +1190,6 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 		// or more particles per ship per turn at the peak of the jump.
 		if(isUsingJumpDrive && !forget)
 			CreateSparks(visuals, "jump drive", hyperspaceCount * Width() * Height() * .000006);
-
-		// If at some point during the jump process the ship is disabled, abort
-		// the jump sequence.
-		if (isDisabled)
-		{
-			hyperspaceSystem = nullptr;
-			hyperspaceCount = 0;
-			return;
-		}
 		
 		if(hyperspaceCount == HYPER_C)
 		{
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1165,9 +1165,6 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 	}
 	else if(hyperspaceSystem || hyperspaceCount)
 	{
-		// Don't apply external acceleration while jumping.
-		acceleration = Point();
-		
 		// If at some point during the jump process the ship is disabled, abort
 		// the jump sequence.
 		if(isDisabled)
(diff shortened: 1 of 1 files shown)


Rakete1111:
Done. FYI I put the check after the acceleration gets nulled since that is exactly what happens a bit further down in the code when landing.

tehhowch:
> Done. FYI I put the check after the acceleration gets nulled since that is exactly what happens a bit further down in the code when landing.

Yep, I figured that! That behavior makes sense for landing, since we need the sprite to always shrink into the planet.

(I accidentally hit edit post on the mobile view and was thinking I hit quote in reply... 🤦‍♂️)

source/Ship.cpp
```diff
@@ -1165,9 +1165,18 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
...
+		if(isDisabled)
+		{
+			hyperspaceSystem = nullptr;
+			hyperspaceCount = 0;
+			return;
+		}
+
 		// Don't apply external acceleration while jumping.
 		acceleration = Point();
-
```
tehhowch, line 1170: Looks like your change ate some whitespace

Rakete1111: oops, fixed :)

Amazinite:
I will say that I mentioned this issue/pr on stream the other day, and the general idea seemed to be to either leave this mechanic as is, or only have the ship be pulled out of jump for jump drives (but not hyperdrives/scram drives).

petervdmeer:
When I open the Twitch link, then I get the error-message; "Sorry. Unless you’ve got a time machine, that content is unavailable.".

My first impression is also that this change is fine. Can you summarize the arguments against this change? (And/Or the arguments to only do this for jump drives?)

Amazinite:
> When I open the Twitch link, then I get the error-message

Yeah, Twitch deletes videos after two weeks. I have the download for it, though.

TLDR:
* This would make capturing a Raider (or other JD-enabled ships) easier, as right now it's not an uncommon occurrence for you to disable such a ship while it's jumping. If that happens, you need to go on an interesting chase to see where the disabled ship went, while with this PR, as mentioned, it'd just make it easier by giving you a window to disable a ship while it can't fight back. In some sense, this would cheapen the usefulness of hit force weapons, as you wouldn't need to prevent a ship from jumping when you can just disable it while it jumps and keep it in-system at the same time.
* Hyperdrives work by creating a "wave" in spacetime that your ship "rides." If you're somehow disabled while jumping, that wave still exists, so could still push you along to your destination.

MCOfficer:
> 
>     * This would make capturing a Raider (or other JD-enabled ships) easier, as right now it's not an uncommon occurrence for you to disable such a ship while it's jumping. If that happens, you need to go on an interesting chase to see where the disabled ship went, while with this PR, as mentioned, it'd just make it easier by giving you a window to disable a ship while it can't fight back. In some sense, this would cheapen the usefulness of hit force weapons, as you wouldn't need to prevent a ship from jumping when you can just disable it while it jumps and keep it in-system at the same time.

Either i suck, or the chances for this happening are very slim.

>     * Hyperdrives work by creating a "wave" in spacetime that your ship "rides." If you're somehow disabled while jumping, that wave still exists, so could still push you along to your destination.

In lore, the explanation still works. Both hyper- and jumpdrive might be charging up before releasing the energy & jumping.

petervdmeer:
I have seen ships getting disabled while they were jumping (and they still jumped away). Somehow I never considered that the ship would still be present and disabled in one of the adjacent systems.
It is indeed a nice gameplay feature if you can still go and search for the disabled ship in neighbouring systems.

The in-lore explanation also works for me.

Amazinite:
> Either i suck, or the chances for this happening are very slim.

It's a fairly common occurrence for me when I'm attempting to board a Raider early game. I also have this happen often when messing around and trying to board Quarg ships.

Amazinite:
I'm not really interested in this change. (See my past explanations for why.) As such, I'm going to close it.

Issue #4325, opened 2019-06-01
Spaceport news feedback


Amazinite:
This issue is just a collection of the feedback I've seen and gotten about spaceport news.

[Here](https://github.com/endless-sky/endless-sky/issues/3682#issuecomment-379503390) is the link to the comment where I mentioned having done a poll on this feature. The following is an image of a new set of polls that I did after the release of the update (posted 5/30/19 and stayed open for about 50 hours), with the first question being the same as the previous poll. At least from the Discord server among those that responded to the poll, the opinion seems to be rather split, with a sizable number of people not being sure on the images.
![image](https://user-images.githubusercontent.com/17688683/58753010-77172380-8486-11e9-99eb-28fd43b27b70.png)

The second question shows that while opinions on the images are mixed, the news system itself is popular. 

With the third question about alien images, although the poll showed positive, I got a sizable number of people giving written feedback about how they feel that the inclusion of alien images would ruin the immersion of imagining what the aliens look like for yourself. Some said that the inclusion of images of people in general was somewhat immersion breaking when every other interaction with someone else in the game, aside from the small number of images we get for Barmy Edward, is done through text alone.

I think the I worded the fourth question poorly. Not really sure what to make of it in hindsight.

The last question gets at the idea of #3742; if we don't have images of people, we could have images of inanimate objects. I received some feedback saying that it would probably be best to mix the two, though.

<hr/>

I opened a [Steam thread asking for feedback](https://steamcommunity.com/app/404410/discussions/0/1653297026046844771/), including on the portraits. Here are some quotes from that thread.

<details><summary>Quotes from Steam</summary>

>The spaceport news system adds a lot of immersion, but I'm not sure how many messages there are and if they will get stale over time. [[source](https://steamcommunity.com/app/404410/discussions/0/1653297026046844771/#c165329702604)]

>I personally feel mixed on the news tab addition, mostly the faces, but I could not articulate why well enough. I didn't mind the inclusion of a picture for the Kraz Engineer. [[source](https://steamcommunity.com/app/404410/discussions/0/1653297026046844771/#c1653297026048427754)]

>j͏ust don't add pi͏ctures for the aliens. it͏ is better left to th͏e imagination. [[source](https://steamcommunity.com/app/404410/discussions/0/1653297026046844771/#c1653297026050051225)]

>>Should the images be removed, but the news system be kept in place?
>
>Honestly this is one of the changes that I hate, but I didn't expect to hate.
[...]

PR #5072, opened 2020-05-16
bug fix to: spaceport news mixing up roles and messages

### LINKED ISSUE #4463: spaceport news mixing up roles and messages
people with different professions in spaceport (e.g. scientist, merchant, loan breaker etc.) may sometimes say something that is more sensible for other profession 

To Reproduce:
1. land on a planet/station/whatever
2. press letter p on keyboard repeatedly
3. pay close attention to the profession and message: the bug seems to happen at random

the person with a given profession would present a corresponding message

Screenshot 2019-08-10 at 17 36 18
in a screenshot attached, a scientist is presenting a message that would be more corresponding of a loan breaker

 - OS: macOSX 10.14.3 (Mojave)
 - Game Source: GitHub .dmg image download
 - Version: 0.9.9(unstable)

here is a link to my savefile, in case it is case-specific: https://pastebin.com/3QiaqWdV

NomadicVolcano:
Fixes  #4463

This update ensures a news story will always use a portrait+name combination that is valid for that news story, while avoiding duplicate occupations if possible. The new logic for picking is:

1. Pick a random portrait and use it if it hasn't been used. Record the news, portrait, and name.
2. Reuse a portrait+name, chosen at random, from the possibilities already used by that news.
3. Pick a random portrait and name that are correct for that news, allowing somebody to have multiple occupations.

Thus far, in my tests, the fallback option 3 has never been used. However, there is nothing wrong with somebody having two professions. That is common in real life, and there is no reason why, for example, a biologist can't be a tourist or a photographer can't be a merchant. Still, for the sake of variety, that should be avoided if possible, and this PR still avoids that if possible.

### DIFF
--- a/source/News.cpp
+++ b/source/News.cpp
@@ -17,6 +17,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "SpriteSet.h"
 
 #include <algorithm>
+#include <iostream>
 
 using namespace std;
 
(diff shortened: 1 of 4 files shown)


Amazinite:
We're most likely removing the portraits this update (ref: #4325), so no need to waste any energy on this.

Issue #1216, opened 2016-06-09
Allow missions to deduct more money than the player has by creating a mortgage


jafdy:
If the negative payment can't be made the the money is not deducted. The alternative is to be able to set a mortgage but I don't know how to do that ether.

MarcelineVQ:
I can't get a mission to trigger with an on accept payment costing more than I have.
Is this a feature you're requesting? If so give this branch a try:
https://github.com/MarcelineVQ/endless-sky/tree/mission-payment

LocalGod79:
@jafdy i've commented on your pull with the possible explanation, and a check that you need to do.

jafdy:
That branch should work but it is no use to me if it isn't merged. (not to mention that I haven't worked out how to compile for windows yet)

endless-sky:
It seems to me that having a mission give you a new mortgage but only if its payment is negative *and* you can't afford the payment is a somewhat confusing behavior. And, it means that it's no longer possible to create missions that are only offered if you have a certain amount of cash on hand (e.g. the "drop bear" attacks in EV Nova). Wouldn't a better option be to allow missions to specify that they apply a new debt to the player, instead? That would have the added advantage that the debt could be at a considerably higher interest rate than a typical mortgage.

jafdy:
That was my preference.

NomadicVolcano:
> Wouldn't a better option be to allow missions to specify that they apply a new debt to the player, instead? That would have the added advantage that the debt could be at a considerably higher interest rate than a typical mortgage.

PR #5073 implements this solution

Hurleveur:
I have the same issue of not knowing if the player has the money or not and not wanting to branch too much, but I just use fines all the time, and try to justify them. Granted its not as flexible as a fully fledged out mortage.

PR #5073, opened 2020-05-16
allow missions to fine the player or give a mortgage

NomadicVolcano:
Adds a new "debt" action to give a fine or mortgage. This resolves issue #1216 using the solution suggested by @endless-sky 

```
mission ...
  ...
  on ...
    debt 100000
      "interest rate" .3
      type fine
      term 300
```

Syntax. Defaults are the same as the game internals:

- `debt 100000` = amount of the fine or mortgage
- `"interest rate" .3` = 0.3% interest rate; default is based on credit rating, with a credit rating of 0 for fines
- `type fine` = fine or mortgage. Default is mortgage if an interest rate is specified, otherwise the default is fine.
- `term 300` = term to calculate daily payments. Default is 60 for fines and 365 for mortgages.

You can put `<debt>` in the description, name, etc. to get the fine amount followed by "credit" or "credits" just like `<payment>`.

The code in Mortgage.cpp that makes the string for the interest rate is updated to handle interest of 1% and higher correctly.

### DIFF
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -333,10 +333,19 @@ void Account::AddMortgage(int64_t principal)
 
 
 
-// Add a "fine" with a high, fixed interest rate and a short term.
-void Account::AddFine(int64_t amount)
+// Add a "fine" with a high, fixed interest rate
+void Account::AddFine(int64_t amount, int term)
 {
-	mortgages.emplace_back(amount, 0, 60);
+	mortgages.emplace_back(amount, 0, term);
+}
+
+
+
+// Add a "fine" with a high, fixed interest rate and a short term.
+void Account::AddFine(int64_t amount, double interest, int term)
+{
+	double creditScore = 2*(600-interest/.00001);
+	mortgages.emplace_back(amount, int(creditScore), term);
 }
 
 
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -344,8 +344,8 @@ void Account::AddFine(int64_t amount, int term)
 // Add a "fine" with a high, fixed interest rate and a short term.
 void Account::AddFine(int64_t amount, double interest, int term)
 {
-	double creditScore = 2*(600-interest/.00001);
-	mortgages.emplace_back(amount, int(creditScore), term);
+	double creditScore = 2 * (600 - interest * 100000);
+	mortgages.emplace_back(amount, creditScore, term);
 }
 
 
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -333,19 +333,30 @@ void Account::AddMortgage(int64_t principal)
 
 
 
-// Add a "fine" with a high, fixed interest rate
-void Account::AddFine(int64_t amount, int term)
+// Add a "fine" with a high, fixed interest rate and a short term.
+void Account::AddFine(int64_t amount)
 {
-	mortgages.emplace_back(amount, 0, term);
+       mortgages.emplace_back(amount, 0, 60);
 }
 
-
-
-// Add a "fine" with a high, fixed interest rate and a short term.
-void Account::AddFine(int64_t amount, double interest, int term)
+// General debt addition, specifying all parameters
+void Account::AddDebt(int64_t amount, int term, double interest, const string &type)
 {
-	double creditScore = 2 * (600 - interest * 100000);
-	mortgages.emplace_back(amount, creditScore, term);
+	double localCreditScore = creditScore;
+	if(term<=0)
+	{
+		if(type == "Mortgage")
+			term = 365;
+		else if(type == "Fine")
+			term = 60;
+		else
+			term = localCreditScore <= 0 ? 60 : 365;
+	}
+	if(interest > 0)
+		localCreditScore = 2 * (600 - interest * 100000);
+	else if(type != "Mortgage")
+		localCreditScore = 0;
+	mortgages.emplace_back(amount, localCreditScore, term, type);
 }
 
 
--- a/source/Account.cpp
+++ b/source/Account.cpp
@@ -336,14 +336,14 @@ void Account::AddMortgage(int64_t principal)
 // Add a "fine" with a high, fixed interest rate and a short term.
 void Account::AddFine(int64_t amount)
 {
-       mortgages.emplace_back(amount, 0, 60);
+	mortgages.emplace_back(amount, 0, 60);
 }
 
 // General debt addition, specifying all parameters
 void Account::AddDebt(int64_t amount, int term, double interest, const string &type)
 {
 	double localCreditScore = creditScore;
-	if(term<=0)
+	if(term <= 0)
 	{
 		if(type == "Mortgage")
 			term = 365;
--- a/source/Account.h
+++ b/source/Account.h
@@ -53,8 +53,8 @@ public:
 	// Liabilities:
 	const std::vector<Mortgage> &Mortgages() const;
 	void AddMortgage(int64_t principal);
-	void AddFine(int64_t amount,int term=60);
-	void AddFine(int64_t amount,double interest,int term);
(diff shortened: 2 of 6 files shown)


source/MissionAction.h
```diff
@@ -88,6 +88,10 @@ class MissionAction {
 	int64_t payment = 0;
 	int64_t paymentMultiplier = 0;
 	
+	int64_t fine = 0;
+	int64_t fineTerm = -1;
```
tehhowch, line 92: I think it's okay for us to define the default term and interest rates here. (Probably the term should just be an `int`.)

Would doing so reduce the branching needed for calling AddFine?

source/MissionAction.cpp
```diff
@@ -497,6 +542,14 @@ MissionAction MissionAction::Instantiate(map<string, string> &subs, const System
 	result.gifts = gifts;
 	result.requiredOutfits = requiredOutfits;
 	result.payment = payment + (jumps + 1) * payload * paymentMultiplier;
+	if(fine)
```
tehhowch, line 545: This should be `if(fine > 0)`

source/MissionAction.cpp
```diff
@@ -497,6 +542,14 @@ MissionAction MissionAction::Instantiate(map<string, string> &subs, const System
...
+	{
+		result.fine = fine;
+		subs["<fine>"] = Format::Number(abs(result.fine))
```
tehhowch, line 548: The code in ::Load, and use of "is positive" instead of just non-zero means we don't need to call `abs` here.

This should also use Format::Credits, not Format::Number.

source/MissionAction.cpp
```diff
@@ -223,6 +223,42 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
...
+			fineTerm = -1;
+			fineInterest = -1;
+			if(fine < 0)
+			{
+				child.PrintTrace("Fines must be larger than 0.");
+				continue;
+			}
+			if(child.Size() > 2)
+			{
+				fineTerm = child.Value(2);
```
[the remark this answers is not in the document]
NomadicVolcano: I prefer to move the term and interest to child nodes.

source/Account.cpp
```diff
@@ -333,10 +333,19 @@ void Account::AddMortgage(int64_t principal)
...
 
+// Add a "fine" with a high, fixed interest rate
+void Account::AddFine(int64_t amount, int term)
+{
+	mortgages.emplace_back(amount, 0, term);
+}
+
+
+
 // Add a "fine" with a high, fixed interest rate and a short term.
```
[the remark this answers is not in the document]
NomadicVolcano: What is outdated?

source/MissionAction.cpp
```diff
@@ -223,6 +223,42 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 				swap(minDays, maxDays);
 			events[GameData::Events().Get(child.Token(1))] = make_pair(minDays, maxDays);
 		}
+		else if(key == "fine")
+		{
+			if(child.Size() < 2)
+			{
+				child.PrintTrace("Invalid fine block; must have 1 to 3 arguments.");
```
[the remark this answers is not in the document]
NomadicVolcano: If it has more than 3 arguments, it is complete, but still invalid.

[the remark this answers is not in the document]
NomadicVolcano: ... which won't be relevant once I move the term and interest to child nodes.

NomadicVolcano:
0.9.13 has been merged to this branch

source/MissionAction.h
NomadicVolcano: The AddFine with an interest rate is a different function than the one without since the interest rate one has to calculate a credit score.

I'll merge the two AddFines functions.

source/Account.cpp
```diff
@@ -333,10 +333,19 @@ void Account::AddMortgage(int64_t principal)
...
-void Account::AddFine(int64_t amount)
+void Account::AddFine(int64_t amount, double interest, int term)
 {
-	mortgages.emplace_back(amount, 0, 60);
+	double creditScore = 2*(600-interest/.00001);
```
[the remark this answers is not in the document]
NomadicVolcano: It turns out the multiplication by int is less accurate. I end up with values that are a bit off. That is because `*100000` is not the same as `/.00001`. The later one gives you exactly the inverse of the calculation in Mortgage.cpp, which is what you need to get the correct interest rate.

NomadicVolcano:
> 0.9.13 has been merged to this branch

It's too late to undo this, but I can close the PR and make a new one if you need me to.

tehhowch:
> 
> 
> > 0.9.13 has been merged to this branch
> 
> It's too late to undo this, but I can close the PR and make a new one if you need me to.

You should be able to interactively rebase onto the proper base branch and keep the correct commits.

NomadicVolcano:
> You should be able to interactively rebase onto the proper base branch and keep the correct commits.

I got lucky. Compared to 0.9.13, only my changes are present.

source/Account.cpp
```diff
@@ -336,7 +336,27 @@ void Account::AddMortgage(int64_t principal)
 // Add a "fine" with a high, fixed interest rate and a short term.
 void Account::AddFine(int64_t amount)
 {
-	mortgages.emplace_back(amount, 0, 60);
+       mortgages.emplace_back(amount, 0, 60);
```
tehhowch, line 339: This is indented by spaces.

source/Account.cpp
```diff
@@ -336,7 +336,27 @@ void Account::AddMortgage(int64_t principal)
...
+}
+
+// General debt addition, specifying all parameters
+void Account::AddDebt(int64_t amount, int term, double interest, const string &type)
+{
+	double localCreditScore = creditScore;
+	if(term<=0)
```
tehhowch, line 346: Operators must be spaced.

source/Account.h
```diff
@@ -54,6 +54,7 @@ class Account {
 	const std::vector<Mortgage> &Mortgages() const;
 	void AddMortgage(int64_t principal);
 	void AddFine(int64_t amount);
+	void AddDebt(int64_t amount,int term=-1,double interest=-1,const std::string &type="");
```
tehhowch, line 57: Spacing around commas and operators. Also, will this function ever be called without a type specified? If not, then that parameter should not be defaulted.

source/Mortgage.cpp
```diff
@@ -40,13 +41,17 @@ int64_t Mortgage::Maximum(int64_t annualRevenue, int creditScore, int64_t curren
 
 
 // Create a new mortgage of the given amount.
-Mortgage::Mortgage(int64_t principal, int creditScore, int term)
-	: type(creditScore <= 0 ? "Fine" : "Mortgage"),
+Mortgage::Mortgage(int64_t principal, int creditScore, int term, const string &requestedType)
+	:type(requestedType),
```
tehhowch, line 45: Space between the : and member initializer.

source/Mortgage.h
```diff
@@ -36,7 +36,7 @@ class Mortgage {
 	Mortgage() = default;
 	// Create a new mortgage of the given amount. If this is a fine, set the
 	// credit score to zero for a higher interest rate.
-	Mortgage(int64_t principal, int creditScore, int term = 365);
+	Mortgage(int64_t principal, int creditScore, int term = 365, const std::string &type="");
```
tehhowch, line 39: Will this ever be called without a type? If not, then do not allow defaulting the type.

NomadicVolcano: Yes. The existing code needs a default for term and type.

NomadicVolcano:
I've rebased manually, and I think I've fixed all of the whitespace issues.

NomadicVolcano:

Issue #3201, opened 2017-11-05
Make the fleet listing sortable by ship type


muglyObserver:
When your list of ships is several pages long on the Player Info screen, it can be difficult to keep things organized.  It would be nice if you could click on the headings (i.e. "ship", "model", "system", etc) to sort the list by those values.  I would be Okay if the list reverts to the "natural" order if you exit and come back.  That might actually be better than permanently sorting the list.

Amazinite:
FYI: You can edit your issues instead of needing to make a new one because you sent it too soon.

muglyObserver:
Thank you.  I figured that out right after it closed it.  :-)

endless-sky:
Note that if the list were sorted in any way, it wouldn't be possible to change your flagship by dragging a ship to the top. (In fact, it wouldn't be possible to drag ships around at all.) I'm worried that the confusion that would add for the player would cancel out any benefit.

hellminister:
we could have also a "custom" sort that keeps the order of the as set by the player that could also be used for the flagship

tehhowch:
Care to elaborate more clearly, @hellminister ? 
This phrase:
> keeps the order of the as set by the player

does not make any sense to me.

hellminister:
just a setting to keep the ordering in the same way as it is at the moment in the game (without the sorting)

(the missing word is ships)

in other words, have an unsorted setting that lets us place the ships as we like, like at the moment

the only change for this needed in the save file would be a line that gives the sorting type currently used, and the actual sorting is done programatically, (so no need to save the new order based on the sorting choice)

tehhowch:
What MZ had said above is that such a method (programmatic sort) inherently means the player can no longer drag to reorder if they sort the list. The confusion caused by such a change in behavior (I can't change my flagship!!) needs to be mitigated somehow, and the only benefit is finding a ship faster. Once you've found a ship in this sorted mode, you can only park /disown it, or open its detail view (which would also need modification to use a different list order when clicking next and previous).

jafdy:
Always putting the flag ship at the top would let you always find it, giving it an extra box for it's self would also help.
It's won't let the player sort by dragging though, unless it was only a one-time sort that only puts ships in the requested order.

KiLEdEnNis:
what if the I menu got an extra button in Ship Info: "Set Flagship", and a * or something before its name on the Player Info tab?
[...]

PR #5077, opened 2020-05-16
Holding shift toggles the park buttons to only affect in-system ships

Amazinite:
**Feature:** This PR implements the feature request ~~detailed and~~ discussed ~~in issue #{{insert number}}~~ on Discord.

## Feature Details
This PR adds the ability to only toggle the park status of in-system ships by holding the shift key before either clicking the (un)park or (un)park all buttons, or pressing the hotkeys for these buttons (P and A).
This allows easier management of larger fleets if for example you park different fleet compositions on different planets. (E.g. war fleet on one planet, trader fleet on another, backup fighters/drones on another, etc.)

## UI Screenshots
Here's a selection of in ship and out of ship systems. Normally when activating the (un)park command, all these ships would be toggled.
image
With this, holding shift then activating park only toggles the in-system ships.
image

## Usage Examples
N/A

## Testing Done
I parked a number of ships on various different planets, then proceeded to make different selections of ships and toggle the (un)park and (un)park all keys with and without holding shift. In all instances, the expected behavior occurred. 

## Performance Impact
N/A

### DIFF
--- a/source/Panel.cpp
+++ b/source/Panel.cpp
@@ -93,7 +93,7 @@ void Panel::AddZone(const Rectangle &rect, const function<void()> &fun)
 
 void Panel::AddZone(const Rectangle &rect, SDL_Keycode key)
 {
-	AddZone(rect, [this, key](){ this->KeyDown(key, 0, Command(), true); });
+	AddZone(rect, [this, key](){ this->KeyDown(key, SDL_GetModState(), Command(), true); });
 }
 
 
--- a/source/PlayerInfoPanel.cpp
+++ b/source/PlayerInfoPanel.cpp
@@ -287,7 +287,7 @@ bool PlayerInfoPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &comman
 		{
 			const Ship &ship = *player.Ships()[i];
 			if(!ship.IsDisabled() && &ship != flagship 
-				&& ((shift && ship.GetSystem() == flagship->GetSystem()) || !shift))
+				&& (!shift || (shift && ship.GetSystem() == flagship->GetSystem())))
 				allParked &= ship.IsParked();
 		}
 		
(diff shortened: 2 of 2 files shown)


source/Panel.cpp
```diff
@@ -93,7 +93,7 @@ void Panel::AddZone(const Rectangle &rect, const function<void()> &fun)
 
 void Panel::AddZone(const Rectangle &rect, SDL_Keycode key)
 {
-	AddZone(rect, [this, key](){ this->KeyDown(key, 0, Command(), true); });
+	AddZone(rect, [this, key](){ this->KeyDown(key, SDL_GetModState(), Command(), true); });
```
Amazinite, line 96: Without this change, holding shift and clicking on the park buttons would not yield the proper behavior; it was only doing what was expected if I used the hotkeys.
I'm unsure of if this will cause any issues, as this is now the case for all panels it seems. I can't off the top of my head see any places where this might be an issue, but I'm sure there was a reason this ignored the mod state in the first place.

petervdmeer:
This is a nice feature, but can we provide this feature by providing sort-buttons in the ship list?

If we would have sort-buttons, then we could sort ships by name (first column), model (second column), system (third column), etc.

Parking all ships in a certain system would then mean: sort by system and then select the first until the last ship in that system (in the sorted list) and then press park.

Having sort-buttons also allows me to quickly park ships not only on location, but also on model and name (and all the other columns we can sort on).

Amazinite:
>can we provide this feature by providing sort-buttons in the ship list?

Ref: #3201
I don't see these features as mutually exclusive, so I don't see why we couldn't just add the ship sorting later down the road. Shift+Park would then become a shortcut for first sorting all your ships, scrolling to where you want, selecting all the local ships, then parking.

Amazinite:
Closing this since there doesn't seem to be much of a desire for it.

PR #5078, opened 2020-05-16
mission-defined string substitutions (phrases in mission text)

NomadicVolcano:
**Feature:** Allow mission-defined `<substitutions>` which allows (among other things) phrases in mission text.

## Usage Examples
There's a great multitude of uses for this, but I'll give a simple, stupid, example.

```
mission...
  description `Hunt a pirate that said "<rudeness>"`
  name `Hunt a rude pirate.`
  "text substitutions"
    rudeness1 phrase "hostile disabled"
    rudeness2 phrase "hostile disabled"
    rudeness = "<rudeness1> <rudeness2>"
  on offer
    dialog `An angry pilot wants you to find a pirate that said, "<rudeness>"`
  on complete
    dialog `The pilot thanks you for decimating the pirate that said, "<rudeness>"`
    payment 100000
  npc kill
    ...
```

This will bring up a dialog saying something like:

> An angry pilot wants you to find a pirate that said, "Leave me alone, you impertinent milk-livered ratsbane. I can't believe I'm at the mercy of such a qualling spur-galled codpiece."

You'll get a consistent completion message:

> The pilot thanks you for decimating the pirate that said, "Leave me alone, you impertinent milk-livered ratsbane. I can't believe I'm at the mercy of such a qualling spur-galled codpiece."

and a consistent description.

You can reference standard keys too:

```
    "text substitutions"
      key1 = "I want <payment> right now!"
      key2 = "My favorite place is <destination>."
```

## Testing Done

A variety of subs combinations were attempted, including invalid ones. 

Three special cases are tried:

1. A sub key that contains a `>` is discarded with a warning
2. Substitutions are processed in order so each one can see the earlier ones
3. References to invalid phrases result in ""
4. Text substitutions are seen in both dialogs and conversations.

## Performance Impact
I'd expect each mission to take a fraction of a microsecond longer to instantiate

### DIFF
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -187,6 +187,8 @@ bool GameData::BeginLoad(const char * const *argv)
 		it.second.FinishLoading(true);
 	for(auto &it : persons)
 		it.second.FinishLoading();
+	for(auto &it : missions)
+		it.second.FinishLoading();
 	startConditions.FinishLoading();
 	
 	// Store the current state, to revert back to later.
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -187,8 +187,6 @@ bool GameData::BeginLoad(const char * const *argv)
 		it.second.FinishLoading(true);
 	for(auto &it : persons)
 		it.second.FinishLoading();
-	for(auto &it : missions)
-		it.second.FinishLoading();
 	startConditions.FinishLoading();
 	
 	// Store the current state, to revert back to later.
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -30,7 +30,6 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include <cmath>
 #include <sstream>
-#include <iostream>
 
 using namespace std;
 
@@ -255,18 +254,18 @@ void Mission::Load(const DataNode &node)
 			else
 				child.PrintTrace("Skipping unrecognized attribute:");
 		}
-		else if(child.Token(0) == "subs")
+		else if(child.Token(0) == "text substitutions")
 		{
 			for(const DataNode &grand : child)
 			{
 				if(grand.Size() > 1 && grand.Token(0).find(">") != string::npos)
-					child.PrintTrace("No \">\" allowed in subs keys (\""+grand.Token(0)+"\"):");
+					grand.PrintTrace("Skipping invalid use of angle brackets in substitution key \""+grand.Token(0)+"\":");
 				else if(grand.Size() == 3 && grand.Token(1) == "phrase")
 					missionSubs.emplace_back(grand.Token(0), "phrase:"+grand.Token(2));
 				else if(grand.Size() == 3 && grand.Token(1) == "=")
 					missionSubs.emplace_back(grand.Token(0), "assign:"+grand.Token(2));
 				else
-					child.PrintTrace("Skipping unrecognized subs definition \""+grand.Token(0)+"\":");
+					grand.PrintTrace("Skipping unrecognized text substitutions definition \""+grand.Token(0)+"\":");
 			}
 		}
 		else
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -30,6 +30,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include <cmath>
 #include <sstream>
+#include <iostream>
 
 using namespace std;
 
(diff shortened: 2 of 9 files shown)


source/GameData.cpp
```diff
@@ -187,6 +187,8 @@ bool GameData::BeginLoad(const char * const *argv)
 		it.second.FinishLoading(true);
 	for(auto &it : persons)
 		it.second.FinishLoading();
+	for(auto &it : missions)
```
tehhowch, line 190: Should be after `startConditions.FinishLoading()` and introduced by a different comment. The other calls are all finalizing ship definitions.

source/Mission.cpp
```diff
@@ -30,6 +30,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include <cmath>
 #include <sstream>
+#include <iostream>
```
tehhowch, line 33: ?

source/Mission.cpp
```diff
@@ -254,6 +255,20 @@ void Mission::Load(const DataNode &node)
 			else
 				child.PrintTrace("Skipping unrecognized attribute:");
 		}
+		else if(child.Token(0) == "subs")
```
tehhowch, line 258: "text substitutions"

source/Mission.cpp
```diff
@@ -254,6 +255,20 @@ void Mission::Load(const DataNode &node)
...
+		{
+			for(const DataNode &grand : child)
+			{
+				if(grand.Size() > 1 && grand.Token(0).find(">") != string::npos)
+					child.PrintTrace("No \">\" allowed in subs keys (\""+grand.Token(0)+"\"):");
```
tehhowch, line 263: If you print the trace from `grand` instead of `child`, you don't need to include any of its tokens in the trace string.

I would also prefer `"Skipping invalid use of angle brackets in substitution key:"` as the warning.

source/Mission.cpp
```diff
@@ -264,6 +279,22 @@ void Mission::Load(const DataNode &node)
...
+void Mission::FinishLoading()
+{
+	for(auto &it : missionSubs)
+	{
+		if(it.second.find("assign:") != string::npos)
+			it.second=it.second.substr(7);
+		else if(it.second.find("phrase:") != string::npos)
+		{
+			const Phrase *phrase = GameData::Phrases().Get(it.second.substr(7));
+			it.second = phrase ? phrase->Get() : "";
```
tehhowch, line 291: This isn't quite a flexible as you'd like. If a job uses a phrase substitution, then the phrase used will only ever differ when the game is loaded: any jobs created from the template will have the same phrase substitution during the same player session.

To fix this you need to maintain the (always non-null) Phrase pointer and invoke it in each use of Mission::Instantiate. (Which then means you do not need Mission::FinishLoading at all, since you will be storing string replacements and Phrase replacements in different maps, and don't need to prefix things with `assign:` or `phrase:`)

source/GameData.cpp
NomadicVolcano: I removed this because it is no longer needed after your other suggested change.

source/Mission.cpp
NomadicVolcano: I did something similar to what you said. I have to retain the order of the non-phrase substitutions, so I'm still using a vector of pairs. However, this does let me get rid of FinishLoading, and reduce the code by a dozen lines or so.

NomadicVolcano:
> Features should be pointed at 0.9.13 as well

Are you asking me to PR to 0.9.13 instead? I don't think I can change the destination branch of a merge on github.

tehhowch:
> 
> 
> > Features should be pointed at 0.9.13 as well
> 
> Are you asking me to PR to 0.9.13 instead? I don't think I can change the destination branch of a merge on github.

Click "edit" on the title. Note that you should ensure that your branch will only have your commits.

NomadicVolcano:
I've rebased the branch and I think I've fixed all the style errors.

source/Mission.cpp
```diff
@@ -952,6 +966,14 @@ const MissionAction &Mission::GetAction(Trigger trigger) const
 Mission Mission::Instantiate(const PlayerInfo &player, const shared_ptr<Ship> &boardingShip) const
 {
 	Mission result;
+	std::vector<std::pair<std::string, std::string>> missionSubs;
+	for(auto &it : missionSubPhrases)
+	{
+		const Phrase *phrase = GameData::Phrases().Get(it.second);
+		missionSubs.emplace_back(it.first, phrase ? phrase->Get() : "");
```
tehhowch, line 973: `phrase` is never nullptr when obtained via Set::Get. Combined with the previous change to missionSubPhrases, you can just insert the result of `Get()` (which will be `""` if the phrase was never actually loaded or defined.)

source/Mission.cpp
```diff
@@ -1122,6 +1144,7 @@ Mission Mission::Instantiate(const PlayerInfo &player, const shared_ptr<Ship> &b
 	subs["<destination>"] = subs["<planet>"] + " in the " + subs["<system>"] + " system";
 	subs["<date>"] = result.deadline.ToString();
 	subs["<day>"] = result.deadline.LongString();
+	Format::MergeReplacements(subs,missionSubs);
```
tehhowch, line 1147: Comma, here and elsewhere.

source/Mission.h
```diff
@@ -49,6 +51,8 @@ class Mission {
 	
 	// Load a mission, either from the game data or from a saved game.
 	void Load(const DataNode &node);
+	// Expand phrases in subs
```
tehhowch, line 54: Match the style for comments, please.

NomadicVolcano:
I've corrected the style issues I could find, and I fixed a bug caused by style issue corrections.

NomadicVolcano:

PR #5080, opened 2020-05-16
Add the ability for plugins to override builtin missions

Rakete1111:
**Feature:** This PR implements the feature request detailed and discussed in issue #4487

## Feature Details

This adds the ability for missions to be overridden by plugins for example.

It works by basically copy constructing a new empty object into the mission to be overridden and then load the new mission normally.

As far as I can understand, `GameData` is responsible for loading any missions in the game. That is also the only point where missions are loaded (where their `Load` function is called). The builtin missions are loaded first, so any plugin that provides a mission with the same name will have the builtin plugin correctly overridden by this PR.

## Testing Done

Using the example mission provided in the issue. Works.

## Performance Impact

N/A

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -92,15 +92,11 @@ void Mission::Load(const DataNode &node)
 		node.PrintTrace("No name specified for mission:");
 		return;
 	}
-	// If a mission object is "loaded" twice, that is most likely an error (e.g.
+	// If a mission object is "loaded" twice, override the current mission (e.g.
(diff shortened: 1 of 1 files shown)


tehhowch:
What are your thoughts on the suggestions made in the related PR #4489?

I also had a comment that may be worth investigating: https://github.com/endless-sky/endless-sky/pull/4489#issuecomment-546349791

I'm not sure we should completely nuke an existing mission definition without warning, just because the name appeared again, which is why I would prefer plugins need to use a keyword to enable this behavior.

PR #5070, opened 2020-05-16
fixes ancient bug: Clicking "depart" will fail a passenger transport mission if escorts carrying passengers are out of system

NomadicVolcano:
This will fix #1624 (Clicking "depart" will fail a passenger transport mission if escorts carrying passengers are out of system)

Changes:

1. Cargo holds in save files include mission cargo and mission passengers, tracked by UUID.
2. Cargo holds are not cleared of mission cargo on load/save.

This needs #5081 which adds UUIDs to missions, ships, and NPCs.

The title of bug #1624 misses an important detail: the failure only happens after loading the game. This is because cargo holds were discarded both when the game was saved and when it was loaded.

### Testing

Four tests so far:

#### The Bug

1. Get missions that put mission cargo or passengers into your escorts.
2. Fly somewhere and land without the escorts.
3. Save game.
4. Load game.
5. Depart
6. You should not fail missions with off-planet escorts.

#### No Free Reward

1. Accept a job that goes into the cargo hold of an escort
2. Go to the destination system without your escort
3. Land and get the "wait for your escort" message
4. Save game.
5. Load game.
6. You should not get rewards for incomplete missions.

#### No Vanishing Cargo

1. Accept a job that goes into the cargo hold of an escort
2. Save game.
3. Load game.
4. Space available for missions should be the same as before loading.

#### New UUIDs

1. Load a game that has not been accessed by this branch.
2. Save the game.
3. The new save file should have UUIDs for all missions, available jobs, and available missions.

PR #5081, opened 2020-05-17
Assign UUIDs to ships, missions, and npcs

NomadicVolcano:
**Feature:** Adds a unique identifier to ships, missions, and NPCs. Also adds a Random::UUID() function to generate random UUIDs. This aids in recreating pointers to missions, ships, and npcs after loading a game.

## Feature Details

UUIDs are these: https://en.wikipedia.org/wiki/Universally_unique_identifier
Standard definition: https://tools.ietf.org/html/rfc4122

In your save files, you should see:

```
ship
  ...
  uuid 2c8ab67c-e7cc-471a-a415-08a63ba52527

mission
  ...
  uuid 27ecb32b-e8a2-4350-aa5b-6b4b773520c7
  ...
  npc
    ...
    uuid b45551e8-324c-4b35-96ba-cb02357694e2
```

## Usage Examples

In and of itself, this PR is useless, but it will be needed by other PRs that have to uniquely identify ships, missions, or NPCs when loading data back from a file. This allows pointers to ships, NPCs, and missions to be restored.

Used by #5070 to load pointers to mission cargo and mission passengers, and #5082 to load the mission whose conversation was being displayed.

## Testing Done

1. Load a game with no UUIDs. Save it. Make sure all ships, missions, available missions, available jobs, and NPCs have UUIDs.

2. Buy a ship. Save. Make sure it has a UUID.

3. Go to a planet with available jobs and available missions. Save. Make sure they have UUIDs.

4. Go to a planet with a job or mission that has an NPC. Save. Make sure the NPC and its ships have UUIDs.

5. Run with this branch. Save. Load mission from earlier version of the code. It should load correctly; the uuids should be discarded.

## Performance Impact

It'll take a fraction of a microsecond longer to instantiate a mission or make a ship.

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -432,7 +432,7 @@ bool Mission::IsMinor() const
 
 void Mission::EnsureUUID()
 {
-	if(uuid.size() < 1)
+	if(uuid.empty())
 		uuid = Random::UUID();
 }
 
@@ -1130,7 +1130,6 @@ Mission Mission::Instantiate(const PlayerInfo &player, const shared_ptr<Ship> &b
 	
 	// Generate the substitutions map.
 	map<string, string> subs;
-	subs["<uuid>"] = result.uuid;
 	subs["<commodity>"] = result.cargo;
 	subs["<tons>"] = to_string(result.cargoSize) + (result.cargoSize == 1 ? " ton" : " tons");
 	subs["<cargo>"] = subs["<tons>"] + " of " + subs["<commodity>"];
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -107,6 +107,8 @@ void Mission::Load(const DataNode &node)
 	{
 		if(child.Token(0) == "name" && child.Size() >= 2)
 			displayName = child.Token(1);
+		else if(child.Token(0) == "uuid" && child.Size() >= 2)
+			uuid = child.Token(1);
 		else if(child.Token(0) == "description" && child.Size() >= 2)
 			description = child.Token(1);
 		else if(child.Token(0) == "blocked" && child.Size() >= 2)
@@ -223,7 +225,10 @@ void Mission::Load(const DataNode &node)
 		else if(child.Token(0) == "stopover" && child.HasChildren())
 			stopoverFilters.emplace_back(child);
 		else if(child.Token(0) == "npc")
+		{
 			npcs.emplace_back(child);
+			npcs.back().EnsureUUID();
+		}
 		else if(child.Token(0) == "on" && child.Size() >= 2 && child.Token(1) == "enter")
 		{
 			// "on enter" nodes may either name a specific system or use a LocationFilter
@@ -272,6 +277,8 @@ void Mission::Save(DataWriter &out, const string &tag) const
 	out.BeginChild();
 	{
 		out.Write("name", displayName);
+		if(uuid.size())
+			out.Write("uuid", uuid);
 		if(!description.empty())
 			out.Write("description", description);
 		if(!blocked.empty())
@@ -374,6 +381,13 @@ void Mission::Save(DataWriter &out, const string &tag) const
 
 
 // Basic mission information.
+const string &Mission::UUID() const
+{
+	return uuid;
+}
+
+
+
 const string &Mission::Name() const
 {
 	return displayName;
@@ -412,6 +426,14 @@ bool Mission::HasPriority() const
 bool Mission::IsMinor() const
 {
 	return isMinor;
+}
+
+
+
+void Mission::EnsureUUID()
+{
+	if(uuid.size() < 1)
+		uuid = Random::UUID();
 }
 
 
@@ -953,6 +975,7 @@ Mission Mission::Instantiate(const PlayerInfo &player, const shared_ptr<Ship> &b
 {
 	Mission result;
 	// If anything goes wrong below, this mission should not be offered.
+	result.uuid = Random::UUID();
 	result.hasFailed = true;
 	result.isVisible = isVisible;
 	result.hasPriority = hasPriority;
@@ -1107,6 +1130,7 @@ Mission Mission::Instantiate(const PlayerInfo &player, const shared_ptr<Ship> &b
 	
 	// Generate the substitutions map.
 	map<string, string> subs;
+	subs["<uuid>"] = result.uuid;
 	subs["<commodity>"] = result.cargo;
 	subs["<tons>"] = to_string(result.cargoSize) + (result.cargoSize == 1 ? " ton" : " tons");
 	subs["<cargo>"] = subs["<tons>"] + " of " + subs["<commodity>"];
--- a/source/Mission.h
+++ b/source/Mission.h
@@ -54,6 +54,7 @@ public:
 	void Save(DataWriter &out, const std::string &tag = "mission") const;
 	
 	// Basic mission information.
+	const std::string &UUID() const;
 	const std::string &Name() const;
 	const std::string &Description() const;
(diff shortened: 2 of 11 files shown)


NomadicVolcano:
**Someone with a windows or mac machine needs to test this.**

The code in the `Random` class that generates these has two lines that differ between Linux and non-Linux. I only have Linux, so I need someone else to test this on other platforms.

Zitchas:
1. Will this compare the generated value with already existing values to ensure that the new uid is unique?
2. Will this include a way for missions to store the resulting uid in a way that can be referred to in subsequent missions?
ex. Mission 1 gives the player a ship and generates a uid.
Mission 2 requires the player to be flying that ship in order to trigger.
The `on complete` of mission 3 removes the ship from the player.

Basically, we need some way to either communicate the generated uid's between missions; or have some kind of a system to ensure that missions can assign specific uid's to ships so that content creators can refer to that specific ship in future missions. The former sounds like more work on the engine side; the later is more work on the creator side. As a creator, the former sounds awesome, but I'm fine with having to manually assign uid's for ships that I'm going to be referring to in future missions. If we go the later route, there should probably be some format for defining ranges of values that are restricted to manually assigned values. For instance, perhaps values where the first `2c8ab67c` block of the uid is equal to the first 8 letters of a government. `remnant0` for Remnant missions, `syndicat` for the syndicate, etc.

NomadicVolcano:
> Will this compare the generated value with already existing values to ensure that the new uid is unique?

The idea behind UUID is that it is sufficiently large that you can't have collisions, so no such checks are needed. If we run into problems, there are smarter UUID-generation algorithms, but I'm using a 128-bit random number, which should be good enough. (Actually, 121 random bits and 7 bits of format definition.)

> Will this include a way for missions to store the resulting uid in a way that can be referred to in subsequent missions?

I suspect tags would be a better option for the use cases you describe.

The UUIDs are mainly to set up pointers correctly when saving and loading. For example, #5070 uses it to keep track of mission cargo in ship holds. 

We need string variables and a reflexive API for the UUIDs to be usable in-game. Once we have that though, there are a great many things that could be done.

source/Mission.cpp
```diff
@@ -1107,6 +1130,7 @@ Mission Mission::Instantiate(const PlayerInfo &player, const shared_ptr<Ship> &b
 	
 	// Generate the substitutions map.
 	map<string, string> subs;
+	subs["<uuid>"] = result.uuid;
```
petervdmeer, line 1133: Do we ever want to display the UUID in-game?
I would expect the UUID not to show up on the UI anywhere.

source/Mission.h
```diff
@@ -67,6 +68,9 @@ class Mission {
 	// offered if no other missions (minor or otherwise) are being offered.
 	bool IsMinor() const;
 	
+	// For backward-compatibility to old saves, generate a UUID if none is present
```
petervdmeer, line 71: The current ES style rules requires comment to end with a dot.

source/Random.cpp
```diff
@@ -112,3 +116,36 @@ double Random::Normal()
 #endif
 	return normal(gen);
 }
+
+
+
+// Get a random UUID
```
petervdmeer, line 122: Maybe add some text to the comment here how an UUID looks?
Also, comments in ES need to be full sentences and end with a dot.

source/Mission.cpp
```diff
@@ -416,6 +430,14 @@ bool Mission::IsMinor() const
 
 
 
+void Mission::EnsureUUID()
+{
+	if(uuid.size() < 1)
```
petervdmeer, line 435: Most other sections (for ships and NPCs) use `if(uuid.empty())`. Is there a reason the check here is different?

source/Random.cpp
NomadicVolcano: I added an example, the IETF RFC number, and an explanation to both Random.cpp and Random.h.

source/Mission.cpp
NomadicVolcano: Removed.

It was for debugging. I suppose if I need to debug again, I can put that line back in manually.

NomadicVolcano:
I've fixed the problems @petervdmeer found.

NomadicVolcano:
This PR is backward-compatible to old saves. The mission, npc, and ship Load() functions are smart enough to ignore the uuid.

source/Random.cpp
```diff
@@ -27,10 +27,14 @@ namespace {
 	mt19937_64 gen;
 	uniform_int_distribution<uint32_t> uniform;
 	uniform_real_distribution<double> real;
+	uniform_real_distribution<> zeroToFifteen(0, 15);
+	uniform_real_distribution<> eightToEleven(8, 11);
 #else
 	thread_local mt19937_64 gen;
 	thread_local uniform_int_distribution<uint32_t> uniform;
 	thread_local uniform_real_distribution<double> real;
+	thread_local uniform_real_distribution<> zeroToFifteen(0, 15);
```
tehhowch, line 36: Specify the type

source/Random.cpp
```diff
@@ -112,3 +116,38 @@ double Random::Normal()
...
+// Get a version 4 (random) Universally Unique Identifier (see IETF RFC 4122)
+// which are hexidecimal strings like "2c8ab67c-e7cc-471a-a415-08a63ba52527"
+// and contain 121 bits of random data.
+string Random::UUID()
+{
+#ifndef __linux__
+	lock_guard<mutex> lock(workaroundMutex);
+#endif
+	char uuid[37];
+	static const char * const hex="0123456789abcdef";
```
tehhowch, line 131: Space your operators.

source/Random.cpp
```diff
@@ -112,3 +116,38 @@ double Random::Normal()
...
+	int i=0;
+	
+	for(int j = 0; j < 8; j++)
```
tehhowch, line 134: prefix over postfix.

NomadicVolcano:
> UUID class that manages the generating the UUID on first use

Yes, that thought occurred to me as well in recent days. I'll implement it when I catch up with my other PRs. I want to finish the Korath Exiles expansion first.

NomadicVolcano:

tatami4:
> create a UUID class that manages the generating the UUID on first use.

@tehhowch  What do you mean by this? Could you explain it more? Right now I don't see how this "UUID" class would work. Who would call it to generate the UUIDs?

Zitchas:
I can't speak for the coding side, but conceptually, when a mission gives the player a ship, it would do one of two things:
a) assign the ship a specific preset UID (this would be good for unique ships), or
b) assign the ship a freshly generated UID and record it somewhere. (this would be good for times when a future mission wants to know that the player is using the ship it gave earlier, but that ship isn't necessarily unique in and of itself. For instance, if someone hired the player to pilot ship XYZ ship from point A to B, the mission would want to be able to check that the player is still flying ship XYZ. This could be a generic job with specific ships for that instance, but not important in the larger scheme of things)
c) When the player purchases a ship, it would be assigned a randomly generated UID. useful for when a mission wants to make sure the player is using a specific ship. Say, if they won some famous battle flying the Crimson Flee, and someone wants to hire them to fly the ship they won the battle with in a parade somewhere. This could be used for that.

tehhowch:
I've implemented UUIDs in the sequence of commits:
954f6eb7ff414349b0417822e530352988060b60 (this PR's foundational work)
eff16d4c03c99dc45663d29bba9d866fd457e7da (my request for lazy initialization)
497722c9a23709b7b87884c3802a9fb58d37bb4b (use of platform-specific APIs instead of an oversized `std::string`)

Zitchas:
Awesome!

PR #5082, opened 2020-05-18
save the user interface state (avoids duplicate log entries after save/load)

### LINKED ISSUE #1938: Setting mission to fail delays the failure until next landing
I believe this has been brought up before and it was intentional to stop a game crash, but maybe the mission should become invisible when it has been set to fail so that it actually looks as though it has failed, even if it still exists without your save file.

### LINKED ISSUE #3746: Reloading save during conversation duplicates mission action

NomadicVolcano:
**Feature:** When the game is saved mid-conversation, it will be loaded back at that point in the conversation without running mission actions or conversation apply blocks a second time. It will also load back to the screen you were on (bank, shipyard, etc.)

This will fix #3746 and fix #1938.  The fix for #1938 comes from merging #4764.

This PR requires #5081 and #4764 which have both been merged to this branch.

## Feature Details

Save files now contain this:

```
"resume ui"
        panel Spaceport
        "mission uuid" d51de53d-d793-4616-817a-7ab605d0c4bb
        "mission trigger" offer
        "conversation index" 0
```

- `panel Spaceport` = which panel you were on when saving
- `"mission uuid" d51de53d-d793-4616-817a-7ab605d0c4bb` = the mission whose dialog or conversation you were viewing
- `"mission trigger" offer` = which mission trigger was being viewed
- `"conversation index" 0` = which conversation node you were in (zero-based index), or 0 for a dialog.

If you're not in a conversation when saving, then the uuid and trigger will be empty ("") and the index will be -1.

### Confirmed Working

- [x] loading/saving from spaceport, bank, hiring, trading, planet
- [x] special cases for new player creation
- [x] on accept - job missions (resume to job board with dialog/convo)
- [x] on accept - non-job missions (no ui for accept)
- [x] on offer
- [x] on decline (no ui for decline)
- [x] on defer (no ui for defer)
- [x] on visit
- [x] on fail
- [x] on complete
- [x] on stopover
- [x] blocked mission dialogs (game already handles that without code changes needed)
- [x] if the convo/dialog is for a done mission, save it as a "done mission" and load it into doneMissions
- [x] this PR can load older versions' saves (UI is not resumed)
- [x] make sure older code isn't confused by "resume ui" or "done mission" (uuid lines generate warnings)
- [x] make sure the "done mission" loads correctly when it has NPCs
- [x] retest in Valgrind to find memory leaks or other memory errors
- [x] replay significant portions of the game, saving and loading frequently
- [x] save/load while naming a new ship results in the ship not being purchased

## Performance Impact

A fraction of a microsecond longer to load a game, save a game, or run steps of a mission conversation.

### DIFF
--- a/data/coalition/coalition outfits.txt
+++ b/data/coalition/coalition outfits.txt
@@ -233,6 +233,9 @@ outfit "Small Steering Module"
 	"turn" 178.8
 	"turning energy" 0.26
 	"turning heat" 2.3
+	"steering flare sprite" "effect/coalition flare/small"
+		"frame rate" 4.8
+	"steering flare sound" "atomic tiny"
 	description "Despite its high efficiency, this module is not sufficient for any but the tiniest of ships. Instead, the modules are designed with the expectation that ships may use two or three of them in tandem."
 
 outfit "Large Steering Module"
@@ -247,6 +250,9 @@ outfit "Large Steering Module"
 	"turn" 711.9
 	"turning energy" 1.0
 	"turning heat" 8.2
+	"steering flare sprite" "effect/coalition flare/large"
+		"frame rate" 3.5
+	"steering flare sound" "atomic small"
 	description "By focusing on small, modular engines rather than the massive capital ship drives that most other species develop, the Coalition loses some of the potential efficiency for very large ships, but gains a considerable amount of flexibility."
 
 outfit "Enforcer Confrontation Gear"
--- a/data/hai/hai outfits.txt
+++ b/data/hai/hai outfits.txt
@@ -464,6 +464,9 @@ outfit `"Baellie" Atomic Engines`
 	"flare sprite" "effect/atomic flare/tiny"
 		"frame rate" 14
 	"flare sound" "atomic tiny"
+	"steering flare sprite" "effect/atomic flare/tiny"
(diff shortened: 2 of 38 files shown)


tehhowch:
I don't feel it is common for games to remember the exact state within a conversation / event / etc that the user is experiencing, and thus don't think this is an approach we should implement.

I acknowledge that the current conversation implementation can result in multiple calls to `apply` if the player quits mid-conversation, but think the more appropriate solution is to collect these changes during the course of the conversation, and then apply the net changes upon conclusion of the conversation.

NomadicVolcano:

Issue #4880, opened 2020-03-04
Reconfigured fast-forward key is not interpreted as character input for dialogs


tehhowch:
**Describe the bug**
When rebound to a character allowed in names / dialogs, the fast forward command key toggles the game's state, rather than being used as a character literal.

**To Reproduce**
1. Rebind FF to `t` 
2. Open the ship info panel
3. Click the ship. Name field
4. Try to enter the character in question, e.g. attempt to set the name "Tilted Terror"

**Expected behavior**
The FF command key should only be intercepted while in-flight

**Screenshots**
n/a

**System (please complete the following information):**
 - Version: 0.9.11

**Link to save file**
n/a

**Additional context**
Reported on discord: 
https://discordapp.com/channels/251118043411775489/536900466655887360/684680299401052226

Caused by #4741

Issue #5038, opened 2020-05-04
Fast Forward Inputs dropped in the Boarding Panel


Anarchist2:
**Describe the bug**
When in the boarding panel, fast forward cannot be toggled. This is contrary to every other situation in the game, where if the fast forward button is pressed, it toggles. Additionally, if Caps-Lock is the fast-forward key, then the Caps-Lock is received but not the fast forward, meaning that they can become de-synced, which, while a small issue, can be frustrating if one uses the fast-forward icon to see whether Caps-Lock is on in their game or not.

**To Reproduce**
1. Disable a ship
2. Board it
3. Try to activate/deactivate fast forward

**Expected behavior**
Fast forward should be togglable in the boarding panel, to match with every other situation.

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Self-Compiled
 - Version: b998c77eda52f89579a913aa4e100f52ad9526ef

tehhowch:
Do not use the FF symbol to determine whether Caps Lock is enabled on your system. The FF symbol only tells you if FF mode is enabled or not.

It is indeed intentional that you cannot toggle FF state when not in active flight.

NomadicVolcano:
It is important to be able to toggle FF at critical times, like if you just captured a ship. You may need to react rapidly to keep the ship alive. Also, the claim that you cannot toggle it in active flight is false; you can toggle it anywhere except text fields, boarding and hailing panels.

The boarding and hailing panels should both allow fast-forward, so long as it does not map to a key used by those panels.

PR #5084 fixes this bug by explicitly allowing fast-forward in boarding and hailing panels.

NomadicVolcano:
The new fix, in #5088, is to turn off fast-forward automatically any time another window comes up on top of the main panel.

NomadicVolcano:
> Do not use the FF symbol to determine whether Caps Lock is enabled on your system. The FF symbol only tells you if FF mode is enabled or not.

It would be possible to sync these two by handling caps-lock specially. It is a modifier, so you can check the on-off state just like you would shift and alt. That would require handling caps-lock specially though.

Amazinite:
> **Expected behavior**
> Fast forward should be togglable in the boarding panel, to match with every other situation.

[...]

PR #5084, opened 2020-05-19
allow fast-forward to be toggled in boarding and hail panels

NomadicVolcano:
A recent UI change blocked fast-forward in panels that handled the keypress. This PR explicitly allows the fast-forward key in boarding and hail panels, unless the key is mapped to one specific to that panel (like "a").

The bug was introduced by #4882 to fix issue #4880 which reported that you cannot type the fast-forward key in text fields. The fix #4882 blocked fast-forward everywhere, except when the UI didn't handle the keystroke. For some reason, the boarding and hailing panels block all keystrokes by default, including fast-forward.

Issue #5038 reported the bug in which fast-forward breaks in boarding and hailing panels. @tehhowch's explanation is:

> It is indeed intentional that you cannot toggle FF state when not in active flight.

which is definitely false. You can toggle it everywhere except boarding panels, hailing panels, and text fields.

It is dangerous to block the fast-forward key outside active flight since you may need to quickly react to something than happened, like capturing a valuable ship.

The fix is to have the BoardingPanel and HailPanel ignore the fast-forward key, unless it mapped to a key needed by that panel (like "c").

### DIFF
--- a/source/BoardingPanel.cpp
+++ b/source/BoardingPanel.cpp
@@ -404,6 +404,8 @@ bool BoardingPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command,
 	}
 	else if(command.Has(Command::INFO))
 		GetUI()->Push(new ShipInfoPanel(player));
+	else if(command.Has(Command::FASTFORWARD))
+		return false;
 	
 	// Trim the list of status messages.
 	while(messages.size() > 5)
--- a/source/HailPanel.cpp
+++ b/source/HailPanel.cpp
@@ -303,7 +303,9 @@ bool HailPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, boo
 		else
 			message = "I do not want your money.";
 	}
-	
+	else if(command.Has(Command::FASTFORWARD))
+		return false;
+
 	return true;
 }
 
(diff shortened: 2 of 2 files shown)


petervdmeer:
I would go the other way around; any panel that pops-up should disable fast-forwarding (and no toggling while the pop-up is active).

NomadicVolcano:
> I would go the other way around; any panel that pops-up should disable fast-forwarding (and no toggling while the pop-up is active).

I had considered implementing a feature where fast-forward would automatically turn off when certain things happened, like an escort being disabled or a dialog popping up. Then you have a control panel where you can choose when fast-forward automatically turns on or off.

petervdmeer:
I suggest to start with a simple PR that just turns fast-forward off when a dialog pops up (without any control panels and other triggers).

Turning off fast-forward when an escort gets disabled will probably stir up quite a bit more discussion and there might be good reasons to make that setting configurable, so I would keep that for a separate PR.

NomadicVolcano:
Closed in favor of #5088

NomadicVolcano:
> I would go the other way around; any panel that pops-up should disable fast-forwarding (and no toggling while the pop-up is active).

I implemented this in #5088 a few minutes ago.

ZombieChibiXD:
Any updates on this? I've lost 3 hours of gameplay efficiency because I couldn't react fast enough to toggle off fast forward 😂

PR #5085, opened 2020-05-20
flag to let missions use wormholes in payment, deadline, and location filters + new hai jobs that use this feature

### LINKED ISSUE #3163: cargo mission payment calc across a wormhole
The payment calc for a cargo mission across a wormhole seems (based on my experiments) to be:

abs(base + multiplier * -1 * tons)

(or maybe it's only shown as positive via <payment>, and the actual payment is sometimes negative?)

I don't claim to know what the best solution for calculating jumps across a wormhole is, but -1 is pretty hard to work with from a mission-designing standpoint. The actual number of jumps in the path would be nice, but even 1 would be easier to work with than -1.

NomadicVolcano:
**Feature:** This will fix #3163 ("cargo mission payment calc across a wormhole") by adding a `"use wormholes"` attribute. When enabled, the mission will use wormholes when calculating payments, or when deciding stopovers, waypoints, npc locations, deadline, and the destination.

Hai jobs in human space are updated to use this feature, and I added a few more Hai jobs with longer distances or more cargo.

## Feature Details

Adds a new attribute to missions, `"use wormholes"`, which will cause them to use wormholes when calculating system distances. By default, wormholes are not used, so this change will have no effect on missions unless they use the new `"use wormholes"` attribute.

```
mission
   ...
   "use wormholes"
```

## Usage Examples

Look at the human-space Hai missions in `hai/hai jobs.txt`

## Testing Done

All six parts of mission instantiation with distance calculations have been tested.

- [x] Waypoint filter
- [x] Destination filter
- [x] Payment calculation
- [x] NPC spawn locations
- [x] Stopover filter
- [x] deadline calculation

## Performance Impact
Should not be discernible.


NomadicVolcano:

PR #5088, opened 2020-05-20
option to turn off fast forward everywhere except main screen

NomadicVolcano:
**Feature:** By default, turn off fast forward when the player leaves the main screen. Bring up a dialog warning fast-forward is disabled when the fast-forward key is pressed anywhere else. This feature can be disabled in the settings.

## Feature Details

This PR adds an AllowFastForward() virtual function to the Panel class, which lets Panels decide whether fast-forward is allowed when that panel is visible. The logic is rigged so that fast-forward can only be on in the main screen and will turn off if you leave the screen or another panel pops up (boarding, hail, etc.) The entire screen is redrawn when fast-forward is toggled so the icon (or its absence) will always be correct.

The first time you try to turn on fast-forward somewhere you can't, the game will warn you fast-forward is only available in space. This warning only shows up once per game execution.

You can turn off this feature in the settings under "Toggle fast-forward anywhere."  Personally, I find this feature extremely annoying, so I just keep "Toggle fast-forward anywhere" turned on at all times. However, other people want this feature, so I've implemented it anyway.

## Testing Done

- [x] "Toggle fast-forward anywhere" option can be toggled in settings
- [x] fast forward button works anywhere if the "Toggle fast forward anywhere" option is enabled
- [x] fast-forward icon is properly synced with the state of fast-forward
- [x] player can toggle fast-forward in main screen
- [x] "Toggle fast-forward anywhere" defaults to false when it isn't set.

#### When "Toggle fast forward anywhere" is turned off

- [x] landing on a planet turns off fast-forward
- [x] boarding panel turns off fast-forward
- [x] hailing panel turns off fast-forward
- [x] player info panel turns off fast-forward
- [x] load panel turns off fast-forward
- [x] preferences panel turns off fast-forward
- [x] menu turns off fast-forward
- [x] dialog shows up once per program execution to warn fast-forward is only available in space

## Performance Impact
One extra full-screen draw when fast-forward is toggled.

### DIFF
--- a/source/LoadPanel.cpp
+++ b/source/LoadPanel.cpp
@@ -366,6 +366,13 @@ bool LoadPanel::Scroll(double dx, double dy)
 
 
 
+bool LoadPanel::AllowFastForward() const
+{
+	return true;
+}
+
+
+
 void LoadPanel::UpdateLists()
 {
 	files.clear();
--- a/source/LoadPanel.cpp
+++ b/source/LoadPanel.cpp
@@ -366,13 +366,6 @@ bool LoadPanel::Scroll(double dx, double dy)
 
 
 
-bool LoadPanel::AllowFastForward() const
-{
-	return true;
-}
-
-
-
 void LoadPanel::UpdateLists()
 {
 	files.clear();
--- a/source/MenuPanel.cpp
+++ b/source/MenuPanel.cpp
@@ -177,13 +177,6 @@ void MenuPanel::OnCallback(int)
 
 
 
-bool MenuPanel::AllowFastForward() const
-{
-	return true;
-}
-
-
-
 bool MenuPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, bool isNewPress)
 {
 	if(!isReady)
--- a/source/MenuPanel.cpp
+++ b/source/MenuPanel.cpp
@@ -177,6 +177,13 @@ void MenuPanel::OnCallback(int)
 
 
 
+bool MenuPanel::AllowFastForward() const
+{
+	return true;
+}
+
+
+
 bool MenuPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, bool isNewPress)
 {
 	if(!isReady)
--- a/source/UI.cpp
+++ b/source/UI.cpp
@@ -32,6 +32,17 @@ UI::UI()
 
 
 
+bool UI::AllowFastForward() const
+{
+	int count=0;
+	for(const auto &it : stack)
+		if(!it->AllowFastForward())
+			count++;
+	return count==0;
+}
+
+
+
 // Handle an event. The event is handed to each panel on the stack until one
 // of them handles it. If none do, this returns false.
 bool UI::Handle(const SDL_Event &event)
--- a/source/UI.cpp
+++ b/source/UI.cpp
@@ -34,11 +34,10 @@ UI::UI()
 
 bool UI::AllowFastForward() const
 {
-	int count=0;
 	for(const auto &it : stack)
 		if(!it->AllowFastForward())
-			count++;
-	return count==0;
+			return false;
+	return true;
 }
 
 
--- a/source/LoadPanel.h
+++ b/source/LoadPanel.h
@@ -47,9 +47,6 @@ protected:
 	virtual bool Drag(double dx, double dy) override;
 	virtual bool Scroll(double dx, double dy) override;
 	
-	// Allow fast-forward.
-	virtual bool AllowFastForward() const;
-	
 	
 private:
 	void UpdateLists();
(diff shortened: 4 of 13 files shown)


NomadicVolcano:
Replaces #5084 using solution suggested by @petervdmeer

source/UI.cpp
```diff
@@ -32,6 +32,17 @@ UI::UI()
 
 
 
+bool UI::AllowFastForward() const
+{
+	int count=0;
+	for(const auto &it : stack)
+		if(!it->AllowFastForward())
+			count++;
```
petervdmeer, line 40: Any non-zero count returns in returning false, why not just return false immediately instead of incrementing count?

source/LoadPanel.cpp
```diff
@@ -366,6 +366,13 @@ bool LoadPanel::Scroll(double dx, double dy)
 
 
 
+bool LoadPanel::AllowFastForward() const
+{
+	return true;
```
petervdmeer, line 371: Why would anyone want to toggle fast-forward while being in the LoadPanel?

source/MenuPanel.cpp
```diff
@@ -177,6 +177,13 @@ void MenuPanel::OnCallback(int)
 
 
 
+bool MenuPanel::AllowFastForward() const
+{
+	return true;
```
petervdmeer, line 182: Why would anyone want to toggle fast-forward while being in the MenuPanel?

NomadicVolcano: You start the game in the MenuPanel. If we disable fast-forward in the menu panel, players new to this will get confused and think fast-forward is broken.

source/UI.cpp
NomadicVolcano: This was leftover from debugging. I've removed it.

source/LoadPanel.cpp
NomadicVolcano: They wouldn't, and it'll catch the keys when you're trying to type save file names. I've disabled it in the loading screen. Good catch.

source/MenuPanel.cpp
NomadicVolcano: I found a better solution to that particular problem. Now, fast-forward is only available in space, and you'll get a warning dialog if you try to enable it elsewhere. The dialog only shows up once per program execution.

NomadicVolcano:
I've tried this a few times and I find it extremely annoying. I'm changing this to a draft PR until someone can come up with an idea of how to make it less annoying.

NomadicVolcano:
I've found a compromise. The feature can be toggled. The default is to forbid fast-forward outside active flight, but you can revert to the previous behavior by turning on the "Toggle fast-forward anywhere" option.

petervdmeer:
What I understood is that the goal of this PR (and of #5084) is to solve the following problem:

> It is dangerous to block the fast-forward key outside active flight since you may need to quickly react to something than happened, like capturing a valuable ship.

This dangerous situation is solved by turning fast forward off when leaving the main screen and by only responding to the fast-forward key while in active flight in space.

Players would notice if fast-forward is active only during active flight, and the fast-forward button works during active flight, so the button works where it is needed. I don't understand why we need the "Fast-Forward is only available in space."-warning-dialog. We don't have warning-dialogs that the fire-button or the forward/thrust-button only works while flying in space.  What makes fast-forward so special that it should have the warning-dialog during "paused" states?

I also don't understand why we need the "Toggle fast-forward anywhere" setting. If toggling fast-forward everywhere is a good idea, then let's just allow that. But https://github.com/endless-sky/endless-sky/issues/4880 gives very good reasons on why to only allow fast-forward toggling during active flight. What are the reasons to allow setting fast-forward to "on" while a panel is active?

NomadicVolcano:
> What I understood is that the goal of this PR (and of #5084) is to solve the following problem... What are the reasons to allow setting fast-forward to "on" while a panel is active?

No. The point of this PR is to allow fast-forward to be toggled during dialogs. In other words, I want to fix the bug tehhowch's introduced: you cannot toggle fast-forward during the boarding dialog. However, some people want fast-forward to be disabled every time a dialog comes up, and to stay disabled even if the player tries to turn it back on. 

Forcing fast-forward to be disabled is extremely annoying when repeatedly doing the same short trade route, or testing new content. In those situations, you want fast-forward to stay on until you turn it off.

petervdmeer:
I did some testing to see if I could understand what was annoying with fast-forward turning off when dialogs pop up, but I still don't see the issue. PR #5148 is the result of my tests and it implements just the feature "By default, turn off fast forward when the player leaves the main screen." (without warnings, without additional preferences and without toggling during dialogs) and it just seems to work.

The dialogs already slow down the game and requires me to click and press buttons. Just pressing the button to re-enable fast-forward after the dialog closes didn't feel annoying to me.

The only situation which I could think of where this new behaviour could be annoying is for long-time regular ES players that got accustomed to the old behaviour of persistent fast-forward. I really like to have such regular players involved in the game and if we can support them without much downsides, then I would love to also have the option of keeping the old behaviour. But I feel that the additional preference and the warning from this PR have too much downsides for the majority of the player-base just for the benefit of only a very few regular players.

Please feel free to review the PR I just created for if I missed something.

NomadicVolcano:

Issue #5093, opened 2020-05-22
Circular conversation when naming the Kestrel


pega3:
**Describe the bug**
In the 'On Complete' part of the 'Kestrel Testing' mission, the player is supposed to help find a name for the new ship. Various choices are given, which may result in a circular conversation.

**To Reproduce**
1. Initiate the Kestrel mission, disable the prototype and return to Wayfarer
2. Go through the conversation up to the point that the question about the name is brought up
3. Keep selecting the first suggestion. This leads to an endless loop of 
  The Wraith -> The Dire Wolf -> The Grey Goose -> The Reaper -> The Wraith -> ....

**Expected behavior**
A name which was suggested once should not be offered again as a choice in the dialog. 

**System (please complete the following information):**
 - OS: Win10 and MacOS
 - Game Source: GitHub
 - Version: 0.9.12

**Link to save file**
n/a

**Additional context**
This is certainly a minor bug, since the circular conversation can be left at any time by selecting the 'Kestrel' option. However, it was mentioned on Discord that circular conversations should not exist, and this is something easy to fix.

jvpoz:
I don't know. I like the fact you can circle the conversation endlessly.

tehhowch:
I'm not in favor of this fix either. A possibly useful engine change could be to remember which choice options have been visited in the conversation instance, and present them differently the next time they appear (e.g. darker).

Note that other games (such as Mass Effect 1) don't prevent the player from asking the same questions of NPCs ad nauseum, either, so it isn't as though we are the only ones that do this.

Amazinite:
There's also more than one instance of looping conversations in game already. Another one that comes to mind is from Rescue Katya 2, where you can endlessly decide to wait for Katya to show up. Therefore I wouldn't strictly consider this a bug (hence why I marked it content).

I'd say that looping conversations are better off avoided, though. When it comes to comparing this case to Rescue Katya, at least it makes some sense in Rescue Katya to be able to continue waiting, but in this case you're dealing with a person who I'd expect to get annoyed with you continually giving the same answer.

pega3:
Since no one thinks this is something which changing would improve the game, I am closing it for the sake of not having useless issues / CRs lying around.

PR #5094, opened 2020-05-22
Fix circular conversation when naming the kestrel

pega3:
fix: Prevent circular conversation by setting a condition to eliminate choices which had already been seleted.

-----------------------
**Bugfix:** This PR addresses issue #5093 

## Fix Details
The player is prevented from selecting the same naming option more than once.
This is achieved by introducing a condition which is set at the two points in the conversation which circle back to a previously offered selection and preventing that back-circle if the option has already been chose before. 

## Testing Done
Prepared a save-file before the Kestrel mission. Reproduced the issue with the unmodified 0.9.12 version. Then introduced the fix and performed a whitebox test of the various dialog options. 

## Save File
n/a


pega3:
The issue that motivated this change was not deemed as something which requires fixing, therefore, I am closing this PR.

PR #5104, opened 2020-05-24
Fix - Hyperspace capabilities of the Subsidurial

Zitchas:
As best I can tell, persons have to have hyperspace capabilities to be spawned. 

This gives the Subsidurial hyperdrive capabilities. Also re-arranges a few of the attributes so they are closer to the standard order.

As a side change, replaced the word "cow" in the description with "bison." Bison are, in general terms, cow-like creatures with the key difference that they are much more prone to be found wild and wandering; whereas cows are almost 100% domesticated.

### DIFF
--- a/data/drak/indigenous.txt
+++ b/data/drak/indigenous.txt
@@ -22,16 +22,15 @@ ship "Subsidurial"
 		"drag" 8.2
 		"heat dissipation" .8
 		"fuel capacity" 100
-		"outfit space" 20
 		"thrust" 10
 		"turn" 800
 		"hull repair rate" 3
 		"energy capacity" 1
 		"inscrutable" 1
 		"gaslining" 1
+		"hyperdrive" 1
 	outfits
 		Mouthparts? 3
-		Hyperdrive
 	gun 0 -97 Mouthparts?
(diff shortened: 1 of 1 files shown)


Amazinite:
Well technically person ships either need to be able to jump into the system (hyperdrive or jump drive), or they need a planet to depart from. But given that the Subsidurial is mainly suppose to spawn in uninhabited system, that clearly causes an issue. :)

data/drak/indigenous.txt
```diff
@@ -18,18 +18,20 @@ ship "Subsidurial"
...
 		"thrust" 10
 		"turn" 800
 		"hull repair rate" 3
 		"energy capacity" 1
-		"fuel capacity" 100
 		"inscrutable" 1
 		"gaslining" 1
 	outfits
 		Mouthparts? 3
+		Hyperdrive
```
Amazinite, line 34: You could just give the ship a `"hyperdrive" 1` attribute instead of needing to give it the outfit. (Would still need the fuel capacity, though.)

Zitchas: Good suggestion. Made the change.

PR #5106, opened 2020-05-25
repeatable jobs give +1 reputation to fitting goverment.

geojak:
repetable jobs give +1 reputation. 

Jobs in hai space give hai rep, 
unfettered tributes give rep for these. 
coalition jobs give coalition rep.
 remnant bonuties give remnant rep. 
syndicate jobs give rep for them. 
deep scientist and cubes give deep security rep. 
smuggling, drug and pirate mission give pirate rep. 
normal tranport, escrot, bounty and bulk delivery jobs in human space give merchant rep.

This feature allows for a immersive reputation buildup by doing mission and people getting to know you in the are you are working in. I originally just added this for pirates to make this part of the game for accesible by being less punishing. I think it is fun when there are more ways then chessing repairs of disabled ships to gain repuation. It might be unrealistic, it might unabalcned but it is fun in my opinion.

The reputation towards republic in a few mission might need to be changed or removed. Was unsure about these.


Arachi-Lover:
don't think that's a good idea for the Coalition and Remnant, since you can't farm reputation for those two the usual way as you do with say Republic by killing pirates, since for the Coalition you can only do it via the Wardragon exploit, and the Remnant iirc you can farm Korath for this but it takes a lot longer, yes? jobs for the Remnant would take long still I suppose but I still see it as pretty out of place, pretty sure it was intentional to make it a grind to get friendly with them again if you ever tribute their worlds. for the Coalition, it renders the Wardragon exploit obsolete, since now you can farm rep with them at any point, rather than just when certain missions are active, so anyone with a Jump Drive can easily farm Heliarch ships and Outfits without a care in the world (and while yes Heliarch ships would be hostile after you kill one of theirs, that doesn't really matter when their government only owns 3 systems, which don't even have outfitters, so if anything it's just mildly inconveniencing people that don't have a massive fleet, and providing constant new ships that get insta-disabled for those that do).

also it seems the Coalition jobs are giving Wanderer reputation for some reason

Zitchas:
I like the *idea* of jobs giving reputation for their respective faction, but I am also concerned about its impact on balance and restrictions.

geojak:
I am not too consercend about balance because 1 reputation is very little. It would be quite the grind. I think this change would make this game more sandboxy and forgiving opening new this to players without permanently hampering the self (so you never do it)

Also the quarg Escort exploit is not much fun to use.

geojak:
In the past demanding tribute was also much easier possible. Yes the new more punishing system is more realistic... But it is also less fun.

Arachi-Lover:
with each job giving +1, you can easily farm reputation to get away with stealing a Heliarch Punisher and still being on chill grounds with Coalition planets, likely all under 30 minutes, and that probably shouldn't happen since those jobs are accessible as soon as you get a Jump Drive, and considering there have been concerns and talk about how it's easy to get Automata ships early on, I'm not sure we should facilitate getting even stronger ships

geojak:
Every playtroigh before I even start the free world missions I get myself the arfecta from the pug early on so that later my pig reputation can be resette. 

This pull doesn't allow heliarch reputation to be farmed, just coalition.

Nobody ever I a serous playthourhg captures quarg or heliarch because you ruin your reputation. This is a fun thing to do that woukd be great if it was made more accessible.

Arachi-Lover:
you don't need Heliarch reputation, only the 3 rings are under the Heliarch government, and those don't even have outfitters. everything else in the Coalition space is under the Coalition government, so if you have a means to easily farm Coalition reputation, it doesn't matter if you have Heliarch reputation since you don't need them to steal the ships, just yoink one of their ships, land on a Coalition world and put a Jump Drive in it.

also, unlike the Quarg, who have a more continuous impact in missions, mainly in the Wanderer campaign, after you do the Coalition missions (literally just first contact and the license one) that's it, there's nothing else you need reputation there for for now, and even with the stuff in my PRs, you would only need reputation with the Heliarch if you specifically wanted to do the Heliarch side of the Coalition campaigns, and iirc for just 1 part of a miscellaneous mission set

PR #5126, opened 2020-06-19
Updates Help with outfitter/cargo

Zitchas:

## Summary
This adds a section to the help file that talks about buying outfits to cargo and uninstalling to cargo. It also changes the outfitter panel code so that it will display each of the three messages once per visit to the outfitter, in sequential order.

Can be tested by creating a new pilot and visiting the outfitter three times in succession.


Zitchas:
Hmmm, alright. It compiles and passes the checks cleanly now...
But it still doesn't display the help messages in the outfitter, though.

Zitchas:
OK, thanks to Tehhowch for all the hints and nudges to help me figure out what I was doing wrong with this.

Now it displays a single help message per visit to the outfitter, and it will display them in sequential order.

source/OutfitterPanel.cpp
```diff
@@ -73,8 +73,18 @@ OutfitterPanel::OutfitterPanel(PlayerInfo &player)
...
+		if(checkedhelp == false) {
+			if(DoHelp("outfitter 1") == true) {
+				checkedhelp = true;
+				} else if(DoHelp("outfitter 2") == true) {
+					checkedhelp = true;
+					} else if(DoHelp("outfitter 3") == true) {
+						checkedhelp = true;
+						}
+		}
+	}
```
tehhowch, line 87: At minimum, fix the indentations and remove the verbose boolean comparisons.

```suggestion
	if(GetUI()->IsTop(this) && !checkedHelp) {
		if(!DoHelp("outfitter 1") && !DoHelp("outfitter 2") && !DoHelp("outfitter 3")) {
			// All help messages have now been displayed.
			checkedHelp = true;
		}
	}
```
This suggestion will make the dialogs show in succession for a single visit. I'm not sure it makes sense for the dialogs to appear on separate visits, given their current phrasing.

source/OutfitterPanel.h
```diff
@@ -88,6 +88,10 @@ class OutfitterPanel : public ShopPanel {
 	std::set<Ship *> previousShips;
 	
 	Sale<Outfit> outfitter;
+	
+	// Keep track of how many of the outfitter help screens have been shown
+	int outfithelp = 0;
```
tehhowch, line 93: Unused

source/OutfitterPanel.h
```diff
@@ -88,6 +88,10 @@ class OutfitterPanel : public ShopPanel {
...
+	bool checkedhelp = false;
```
tehhowch, line 94: Should be `checkedHelp`, to match variable naming style.

data/help.txt
```diff
@@ -74,12 +74,22 @@ help "navigation"
 	`When you reach a new system, you can press <Land on planet / station> to land on any inhabited planets that are there.`
 	`Also, don't worry about crashing into asteroids or other ships; your ship will fly safely below or above them.`
 
-help "outfitter"
+help "outfitter 1"
```
tehhowch, line 77: This change will make the this particular dialog display again, even for players that have seen this text before.

source/OutfitterPanel.cpp
Zitchas: I was thinking that players probably don't need to be bombarded by three successive boxes all at once - I thought it better to break it up over a few successive visits. But I suppose it could be right together too.

source/OutfitterPanel.h
Zitchas: Removed. Thanks!

source/OutfitterPanel.h
Zitchas: Changed. Thanks!

data/help.txt
Zitchas: I personally don't see a problem with that and find it more standardized going forward, but I can take it off if that's what people prefer.

data/help.txt
```diff
@@ -80,6 +80,16 @@ help "outfitter"
 	`Use your scroll wheel, or click and drag, to scroll the view.`
 	`As in the trading panel, you can hold down Shift or Control to buy 5 or 20 of an outfit at once, or both keys to buy 100.`
 
+help "outfitter 2"
+	`Sometimes you may want to buy some outfits directly into your cargo hold instead of having them installed. Perhaps a mission has asked you to retrieve some outfits or you want to carry extra ammunition to refill your ships. There are several ways you can do so:`
+	`1) The outfitter interface has a small checkbox labeled 'in-cargo' that, when checked, will instruct the dockworkers to load any outfits you purchase into your cargo bay instead of installing it.`
+	`2) If you do not have any ships selected in the outfitter control panel when you make a purchase, the outfitter will assign it for delivery to your cargo hold instead of scheduling an installation.`
```
Amazinite, line 86: Instead of basically copy-pasting from the Remnant outfitter, I'd say reword these so that they fit the style of the other help dialogs. That is, they aren't written "in-universe" but instead from the perspective of someone playing a game.

Zitchas: @Amazinite OK, I changed a  bunch of these two texts. I didn't really think that either one of them was overly in-character, just something that wasn't explicitly fourth-wall breaking. But either way, they're somewhat different now.

PR #5137, opened 2020-06-28
Remove Navy License from the Auxiliary to bring it in line with other large navy vessels

Anarchist2:
Navy ships that have their own specialised license don't have a Navy License requirement, but the Auxiliary does.


Amazinite:
While we're here, mind adding the ships (or at least the main ship) to the Navy shipyard? (Ref: https://github.com/endless-sky/endless-sky/commit/07175ec981370e973fc009ab45c3e7925fd56409#commitcomment-40225581)

Darcman99:
*While we're here,* mind putting quotes around the `Navy Auxiliary` license requirement? I seem to have borked my last PR.

PR #5141, opened 2020-07-02
Parrot and Author governments no longer give fines.

Eloraam:
## Fix Details
The "Parrot" and "Author" governments could give fines.  "Cap'n Pester" often shows up and scans the user.  Normally this is fine, but if the player has any outfits with the "atrocity" property set, then suddenly they're being attacked by a Quarg Wardragon that just sorta shows up out of nowhere.  That's a bit much even for such a terrible outfit.

This just makes the author vehicles not shoot players for being terrible people.

## Testing Done
Very small change.  The game still starts and nothing else is different.

## Save File
Nothing in the default content uses the "atrocity" property, so a save game testing it is not possible without a plugin.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -18,6 +18,7 @@ government "Alpha"
 government "Author"
 	"player reputation" 1
 	"bribe" 0
+	fine 0
 
 government "Builder"
 	# While when you attack the drones they will try to defend themselves, when you come back 
@@ -373,6 +374,7 @@ government "Parrot"
 	swizzle 2
 	"player reputation" 1
 	"bribe" 0
+	fine 0
 
 government "Pirate"
 	swizzle 6


tehhowch:
This is intended behavior.

Eloraam:
The fine is fine, but he goes full hostile if you use the "atrocity" property.  Is that also intended?

tehhowch:
Yes, the atrocity property invokes a death sentence for the player. If the player is landed and caught, they will promptly die. If in flight (e.g. scanned with "atrocity" goods/outfits or attempting planetary domination), then the player's relationship with that government becomes immediately hostile.

Zitchas:
"Atrocity" is intended to only be put on things that are so terrible that authors and responsible governments have included it on "destroy all instances upon discovery" lists. Rather like the "Omega Particle" in Star Trek lore.

Eloraam:
My question is whether, specifically, "Cap'n Pester" should attack the player on sight.  He tends to be a bit out of scale with the locals in most parts of the galaxy.

tehhowch:
> My question is whether, specifically, "Cap'n Pester" should attack the player on sight. He tends to be a bit out of scale with the locals in most parts of the galaxy.

That's certainly true, but is [consistent with the reference](https://evn.fandom.com/wiki/Hector).

What could be done, and I wouldn't be opposed to, would be to make the Parrot government only enforce within non-human systems, via [an `enforces` node](https://github.com/endless-sky/endless-sky/wiki/CreatingGovernments#enforcement-zones). That will unfortunately also prevent normal fines from being levied for e.g. Nerve Gas.

Eloraam:
That was difficult to test.  Had to temporarily patch the source code to make them appear a lot more often to get it working.

I figured that they should probably appear in Hai space too.

tehhowch:
An alternate solution would be a bit more expansive, but preserve the current fining behavior.
It would need
1) custom atrocity penalties for Author & Parrot of 0 by default
2) code changes
   1) scanning methods now return a fine amount, and also whether or not an atrocity item was found, e.g. `pair<int64, bool>`
   2) if a government doesn't care about atrocities (`PenaltyFor(scan) == 0`), then scanning the player could only levy a fine rather than returning a message sentencing the player to death. (I'm not sure whether we should still call Offend to notify other governments of the atrocity outfit discovery, discussion would be needed.)
3) a follow-up mission-triggered event that restores the current atrocity penalties after some condition (like completing the main plot or discovering some region of space or making some other government hostile).

The bits from 2) are probably viable in their own pull, as the atrocity mechanic is not exactly used often and the current behavior seems not-quite-right, for governments that scan the player but don't care about atrocities.

Eloraam:
So if I understand you, the idea is that after the Free Worlds questline, the current state should be restored, and you should be attacked by Pester in say... Korath space?

If that's the case, I'll have to rethink the content I'm developing (plugin for now, but trying to keep it stock-like).

Eloraam:
Pure mechanism now, no policy at all.  No change to existing code or plugins that don't use it.  And a nice new feature.

And it still lets me solve the problem I'm having in my plugin.

BaddestApple:
I like the idea of being able to set legality and fines differently depending on the government, but would you be able to provide some clarification of the syntax?

In the example legality block, the named legality would be an atrocity in the Republic and a fine of 20,000 in the Free Worlds, but what does the `default illegal xxx` and `default atrocity` line do? If it's to set a different legality level for all other governments apart from the named ones, wouldn't `default illegal xxx` and `default atrocity` be mutually exclusive?

And how I do integrate the legality block into an outfit? "Illegal name" and "Atrocity name" is a little unclear; do I replace `illegal xxx` with `illegal "name"` in the outfit, with the "name" being the legality block?

For example, is this syntax proper or am I butchering it?

```
outfit "Smuggling Compartments"
	category "Tactical"
	cost 75000
	thumbnail "outfit/smuggling compartments"
	"mass" 3
	"cargo space" 5
	"weapon capacity" -2
	"outfit space" -2
	"engine capacity" -2
	"cooling inefficiency" .5
	"heat dissipation" -.02
	"scan interference" 4
	illegal "smuggling"
	description "	The classic Smuggling Compartments find hidden nooks and crannies anywhere they can between your outfit, weapon, and engine spaces, giving additional smuggling capacity and scan interference with its specially designed scan interference lining. Hiding cargo in your bulkheads requires sacrificing some cooling efficiency and heat dissipation, however."
	description "It's illegal to outfit your ship in this way... that is, if scanners or inspectors are able to discover your hidden stash."

legality "smuggling"
	default illegal 1000
	illegal 500000
		government "Republic"
	illegal 20000
		government "Free Worlds"
```

Eloraam:
@BaddestApple Yes that's correct.

Zitchas:
Anyway, I really like the idea of being able to establish different legalities for outfits. 

Could this be used to establish variable legality, not just between something being an atrocity versus being merely illegal,  but also allow something to be legal?

For example, the Remnant overall have an ambivalent attitude towards nerve gas. In their view, dead is dead: Killing someone by shooting them is just as bad as killing them with poison or blowing up their ship. There's nothing "good" or "noble" about it, it is just something unpleasant that sometimes needs to happen. Soldiers are expected to use the most appropriate tool for the job, whatever that happens to be. 

So they wouldn't care if someone was carrying nerve-gas... Unless the person wasn't friendly with the Remnant, in which case they'd be likely to "kill on sight" anyone with Nerve Gas.

Eloraam:
@Zitchas I'm pretty sure that if you set the fine to 0, by saying "default illegal 0" or "illegal 0", that the outfit or cargo is considered legal either by default, or by that government.  I used it that way in my own test case.

It wouldn't be very difficult to add the name "legal" to those cases, if anyone's still interested in this patch.

tehhowch (changes requested):
Needs conflict resolution at a bare minimum; style fixes would be appreciated too.

w.r.t completeness:

 - I'd like to see ideas / mockups for displaying variable legal status for outfits which are only illegal to certain governments, which would be obscured from the player based on the current PR.
 - Should legality definitions be extendable via events?
 - How should the game handle plugins defining the same legality node? As I read it, due to use of `map::emplace`, the first value always wins, which is the opposite behavior of most of the game's data models.

Zitchas:
I'd say "Yes, definitely" to the "should they be extensible via events?" question.

Honestly, the more that events can change, the better. Especially for something like legalities. Heh. It'd actually be kind of interesting if "what is legal" in the Republic change a bit depending on how campaigns end. If you end up on the checkmate ending, then, say, certain illegal outfits aren't fined as heavily. (or maybe certain atrocities aren't are merely heavily fined)

Galaucus:
There's been some discussion on the Discord channel regarding whether to assign illegality at the government side of things or in the outfit itself.

General consensus - and the approach I support - is behind doing it on the government panel. This prevents the problem of spoiling the existence of governments that the player hasn't met, and we can move any player-facing information to the logbook by creating an entry (and corresponding popup) any time they encounter an illegal outfit being sold.

We'll also be instituting a new hailing panel that pops up when ships scan you, and it's cleanest to just do that all from the same place; the government definition.

It's my hope that we can implement such a change as part of a more comprehensive smuggling overhaul as outlined in #6314 .

Hurleveur:
I just noticed that one of the PRs I was asked to do is a duplicate of this (https://github.com/endless-sky/endless-sky/pull/6407)
I'll try to see if mine matches what tehhowch asked to change here soon, since the owner of this PR seems away.

Eloraam:
Vaguely paying attention but really not having the time to update the conflicts.  Have at it.

Hurleveur:
We really did it in different ways. Too different for me to be able to implement any of this really.
For instance, mine it's only definable from the government's perspective.
But if someone disagrees I could make some changes to mine.

I don't see any change tehhowch asked that I can actually apply to my PR. 
Also mine allows for making an ex illegal outfit be ignored by a given government, and for all of it to change by events.

> Vaguely paying attention but really not having the time to update the conflicts. Have at it.

As far as I can see its been a year this is open and you haven't found the time to address tehhowch's requests.
Do you think you can find the time, or should this PR be closed?

Eloraam:
I could probably find the time to clear the conflicts, but some of the requests looked a lot more involved than that.  I'm not currently developing a plugin anymore, so a deep dive isn't likely at the moment.

Your PR would have solved my problem anyway.

Hurleveur:
In that case I'll be closing this PR since we prefer to only have active PRs open.
It is not permanent by any mean, feel free to reopen it when/if you feel like working on it more! 
The thing is that, on top of that, this is a duplicate.

PR #5142, opened 2020-07-02
Fixed pathing through undefined systems.

Eloraam:
## Fix Details
The pathfinder could still choose to path through a system that has not
yet been defined (for example, when it is created by a mission).  The
rest of the game ignores these systems with similar checks to these.

## Testing Done
I played a fair bit of the game with these fixes applied.  Testing that it actually fixes the bug was achieved by playing with PR 4871 applied, which triggers it.  The bug causes a hidden system at 0,0 (near Kor Men) to be pathable, but not visible.  This just takes it out of the pathfinder as well.

## Save File
Bug only occurs with plugins or PR 4871 applied.

### DIFF
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -553,8 +553,12 @@ void GameData::Change(const DataNode &node)
 // that a change creates or moves a system.
 void GameData::UpdateNeighbors()
 {
-	for(auto &it : systems)
+	for(auto &it : systems) {
+		// Skip systems that have no name.
+		if(it.first.empty() || it.second.Name().empty())
+			continue;
 		it.second.UpdateNeighbors(systems);
+	}
 }
 
 
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -553,7 +553,8 @@ void GameData::Change(const DataNode &node)
 // that a change creates or moves a system.
 void GameData::UpdateNeighbors()
 {
-	for(auto &it : systems) {
+	for(auto &it : systems)
+	{
 		// Skip systems that have no name.
 		if(it.first.empty() || it.second.Name().empty())
 			continue;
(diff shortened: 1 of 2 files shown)


source/GameData.cpp
```diff
@@ -553,8 +553,12 @@ void GameData::Change(const DataNode &node)
 // that a change creates or moves a system.
 void GameData::UpdateNeighbors()
 {
-	for(auto &it : systems)
+	for(auto &it : systems) {
+		// Skip systems that have no name.
+		if(it.first.empty() || it.second.Name().empty())
```
tehhowch, line 558: I'm not sure `it.first` can ever be empty, as that would imply that the game called `GameData::Systems().Get("")`

Eloraam: I agree, it shouldn't be.  I copied the condition from another place in the source that also used it, and it felt like the extra paranoia wouldn't hurt anything.

source/System.cpp
```diff
@@ -333,9 +333,14 @@ void System::UpdateNeighbors(const Set<System> &systems)
 	
 	// Any other star system that is within the neighbor distance is also a
 	// neighbor. This will include any nearby linked systems.
-	for(const auto &it : systems)
+	for(const auto &it : systems) {
```
tehhowch, line 336:
```suggestion
	for(const auto &it : systems)
	{
```

source/GameData.cpp
```diff
@@ -553,8 +553,12 @@ void GameData::Change(const DataNode &node)
 // that a change creates or moves a system.
 void GameData::UpdateNeighbors()
 {
-	for(auto &it : systems)
+	for(auto &it : systems) {
```
tehhowch, line 556:
```suggestion
	for(auto &it : systems)
	{
```

Eloraam:
Sorry about that, missed the minor style point.  Thought I had them all the first time.

Amazinite:
Death to the 0,0 system.

PR #5146, opened 2020-07-03
Add heliarch patrols that watch armed players

Quantumshark:
## Summary
The Heliarchs state that they do not trust civilians with weapons, but currently they do not act on this when a player carries weapons.
This PR adds Heliarch patrols that follow and monitor the activity of any player travelling around Coalition space with weapons.
This was briefly discussed on the discord server, so I have opened this PR to allow further discussion.

### DIFF
--- a/data/coalition/coalition missions.txt
+++ b/data/coalition/coalition missions.txt
@@ -109,6 +109,8 @@ mission "Heliarch Watchers"
 		not "Heliarch Watchers: active"
 	to fail
 		"armament deterrence" == 0
+	to complete
+		never
 	npc
 		personality heroic opportunistic
 		government "Heliarch"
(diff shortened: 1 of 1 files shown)


Zitchas:
Instead of it just failing invisibly, what about having an actual statement when the mission fails? I'm sure Arachi could write the dialog better than me, but something to the effect of "Wisely chosen to disarm your ships you have, no longer supervise you must we. Thank you for following our laws we do."

Either way, looks like a good start.

Amazinite:
Seeing as how I am giving Arachi decision making power over what happens with the Coalition, and he isn't fond of this idea, I'm going to be closing this. I've just made a change to the Heliarch conversation where they explicitly state that your weapon ownership is permissible (with the implication that this is only the case so long as you are respectful of the Heliarchs and Coalition). 16f8b39

Zitchas:
OK, sounds good to me.

PR #5148, opened 2020-07-05
feat: Switch off fast-forward if the player is not in flight

petervdmeer:
**Feature:** This PR implements the switch off fast-forward on dialogs feature as discussed in PRs #5088 and in #5084

## Feature Details
When fast-forward is on and the player is not in flight, then the game automatically switches fast-forward off. Fast-forward can be re-enabled once the player is in flight again.

## UI Screenshots
N/A

## Usage Examples
Just press fast-forward and notice that it switches off when boarding and a boarding dialog opens or when you open the map or land on a planet. Then switch fast-forward back on later when in flight again.

## Testing Done
- Switched on fast-forward while hanging around in Shauna and boarded some ships. Fast forward turned off as expected.
- Navigated from Shauna to some systems higher up. Fast forward switched off after opening the map, but could be switched on again during the flight as expected.

## Performance Impact
One additional if-statement in the main game-loop (executed 60 times per second). I don't expect any noticeable performance impact.


MCOfficer:
Maybe i'm missing the point. What is the benefit for the player? FF has no effect in anything but flight, so why not leave it turned on at all times?

petervdmeer:
@NomadicVolcano made a very good point in #5084 and in #5088 that resuming at fast-forward speed directly after a dialog (especially after the boarding dialog) can be undesirable; You just captured a new ship and both the new ship and your flagship are waiting for new orders, or you spot something during the boarding dialog for which you want to give commands immediately after the dialog (like a next nice ship to capture).

In older versions of ES it was possible to switch fast-forward off during the dialog itself, but this option was removed in #4882 because the fast-forward keys could conflict with normal dialog-input-keys.

NomadicVolcano, Amazinite and I discussed quite a bit on the possible ways to avoid annoyance about fast-forward just after resuming from a dialog and switching fast-forward off anytime a dialog starts looked like an appropriate way to handle this.

pega3:
I have wished for this feature every time I start using FF. I prefer having the takeoff with normal speed, and only want to trigger FF at a certain point. If this behavior was standard or at least configurable, I don't need to turn it off shortly before I land, but can let it run. 
This would also hopefully solve an issues I have occasionally experienced, where I pressed "F" while landed and suddenly was in FF mode when taking off.

source/main.cpp
```diff
@@ -247,6 +249,13 @@ void GameLoop(PlayerInfo &player, Conversation &conversation, bool &debugMode)
 			showCursor = shouldShowCursor;
 			SDL_ShowCursor(showCursor);
 		}
+		bool inMap = !gamePanels.IsEmpty() && (dynamic_cast<MapPanel*>(gamePanels.Top().get()) != nullptr);
+		bool inPlayerInfo = !gamePanels.IsEmpty() && (dynamic_cast<PlayerInfoPanel*>(gamePanels.Top().get()) != nullptr);
```
Amazinite, line 253: From the [style guide](https://endless-sky.github.io/styleguide/styleguide.xml):
image
Can these dynamic_casts be avoided?
(I haven't dealt with the panels before, and I haven't looked too deeply into them so this suggestion might be useless, but could you perhaps add a function to Panel.cpp that is implemented by the other panels that returns a string or enum telling us what panel it is? Then instead of using dynamic_cast, we would just call this function from the top panel and set a boolean to true if the top panel is the Map or PlayerInfo.)

source/main.cpp
```diff
@@ -247,10 +249,12 @@ void GameLoop(PlayerInfo &player, Conversation &conversation, bool &debugMode)
...
 		}
-		// Switch off fast-forward if the player is not in flight (for example when the
-		// boarding dialog shows up or when the player lands or opens the map). The player
-		// can switch fast-forward on again when flight is resumed.
-		if(!inFlight && isFastForward)
+		bool inMap = !gamePanels.IsEmpty() && (dynamic_cast<MapPanel*>(gamePanels.Top().get()) != nullptr);
+		bool inPlayerInfo = !gamePanels.IsEmpty() && (dynamic_cast<PlayerInfoPanel*>(gamePanels.Top().get()) != nullptr);
+		// Switch off fast-forward if the player is not in flight or flight-related screen
+		// (for example when the boarding dialog shows up or when the player lands or opens
+		// the map). The player can switch fast-forward on again when flight is resumed.
```
Amazinite, line 256: Opening the map is now an exception, so should be removed from this list of examples of where FFW is switched off.

source/main.cpp
petervdmeer: > Can these dynamic_casts be avoided?
> could you perhaps add a function to Panel.cpp that is implemented by the other panels that returns ...

Yes, I used the solution from @NomadicVolcano from #5088 and added an "AllowFastForward" virtual function to the panels. The function returns true for the MainPanel, MapPanel, PlayerInfoPanel and returns false for all other panels.

source/main.cpp
petervdmeer: Agreed, updated the comment.

Issue #5149, opened 2020-07-06
Tinker tribute after Pug


Pete-2004:
**Describe the bug**
After Pug war, Tinker tribute remains at 1. The line
`"tribute:` Tinker" *= `1100`
 is missing in event "pug territory liberated".

Zitchas:
Thank you for reporting the omission. 

Just a pointer for future issues: Please remember to delete whatever stuff from the template that you aren't filling out to avoid confusion. (or ideally, fill it out).

Pete-2004:
OK. I edited my post.

tehhowch:
Note that if you delete most of it, you will be asked to fill it out. Especially relevant are things like the game version you are playing, where you obtained it, and so on.

Closing as fixed in 0.9.13; you will need to release and re-tribute the planet since you already received the event.

PR #5150, opened 2020-07-06
Fix: Restores Tinker's tribute after Pug War

Zitchas:
**Bugfix:** This PR addresses issue #5149 

## Fix Details
In the event "pug territory liberated", Tinker's tribute is not restored - unlike all the other planets and stations in the affected territories.

Thanks to Pete-2004 for spotting this bug.


tehhowch:
Note that this will fix the tribute condition for anyone who has not yet received the event, but it will not fix the tribute condition for any saves that already received the event from the game engine. I'm not particularly troubled by that, since affected players can just release and re-tribute the planet to obtain the post-war planet-defined value.

PR #5152, opened 2020-07-06
feat: Move swizzling into shader

janisozaur:
## Summary
Use swizzling in shader instead of extension. This is effectively the same but removes the requirement for extension.

## Testing Done
I forced swizzle to a constant value at compile time and verified the images look the same before and after my change.

### DIFF
--- a/source/SpriteShader.cpp
+++ b/source/SpriteShader.cpp
@@ -82,7 +82,7 @@ void SpriteShader::Init(bool useShaderSwizzle)
 		"uniform float frame;\n"
 		"uniform float frameCount;\n"
 		"uniform vec2 blur;\n";
-	if (useShaderSwizzle) fragmentCodeStream <<
+	if(useShaderSwizzle) fragmentCodeStream <<
 		"uniform int swizzler;\n";
 	fragmentCodeStream <<
 		"uniform float alpha;\n"
@@ -124,7 +124,8 @@ void SpriteShader::Init(bool useShaderSwizzle)
 		"  }\n";
 	
 	// Only included when hardware swizzle not supported, GL <3.3 and GLES
-	if (useShaderSwizzle) {
+	if(useShaderSwizzle)
+	{
 		fragmentCodeStream <<
 		"  switch (swizzler) {\n"
 		"    case 0:\n"
@@ -172,7 +173,7 @@ void SpriteShader::Init(bool useShaderSwizzle)
 	blurI = shader.Uniform("blur");
 	clipI = shader.Uniform("clip");
 	alphaI = shader.Uniform("alpha");
-	if (useShaderSwizzle)
+	if(useShaderSwizzle)
 		swizzlerI = shader.Uniform("swizzler");
 	
 	glUseProgram(shader.Object());
@@ -258,7 +259,7 @@ void SpriteShader::Add(const Item &item, bool withBlur)
 	// Bounds check for the swizzle value:
 	int swizzle = (static_cast<size_t>(item.swizzle) >= SWIZZLE.size() ? 0 : item.swizzle);
 	// Set the color swizzle.
-	if (SpriteShader::useShaderSwizzle)
+	if(SpriteShader::useShaderSwizzle)
 		glUniform1i(swizzlerI, swizzle);
 	else
 		glTexParameteriv(GL_TEXTURE_2D_ARRAY, GL_TEXTURE_SWIZZLE_RGBA, SWIZZLE[swizzle].data());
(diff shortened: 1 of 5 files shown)


Zitchas:
This sounds interesting, but I'll admit I don't really know what this is changing; or why. Does doing it with shaders instead of extension lead to fewer dependencies and thus broader compatibility?

janisozaur:
Yes

Eloraam:
GL_ARB_texture_swizzle was promoted to core in GL 3.3, so all hardware made since then has support for this as required by the standard.  Is there a motivation for not using it?

MCOfficer:
> 
> 
> GL_ARB_texture_swizzle was promoted to core in GL 3.3, so all hardware made since then has support for this as required by the standard. Is there a motivation for not using it?

We do still support all OpenGL 3+ drivers, including devices that may not have 3.3.

vkr-13:
This should probably be revisited when GL 3.2 and under are dropped from the game's supported list of libraries. I think this is a good idea as long as it does not impact current players with legacy machines.

tehhowch (commented):
@thomasballinger

What's the performance impact of the move?

I think it makes more sense to provide shader program swizzling as a fallback for hardware that does not have swizzle extension support, rather than remove use of the extension, especially if the extension is more performant.

thomasballinger:
@tehhowch Hm good question, I played with this for a bit and couldn't get my GPU utilization to budge even with more dramatic shader changes so I don't know. Instead I've made this a fallback as you suggest.

source/SpriteShader.cpp
```diff
@@ -113,10 +121,48 @@ void SpriteShader::Init()
 		"      else\n"
 		"        color += scale * texture(tex, vec3(coord, first));\n"
 		"    }\n"
-		"  }\n"
+		"  }\n";
+	
+	// Only included when hardware swizzle not supported, GL <3.3 and GLES
+	if (useShaderSwizzle) {
+		fragmentCodeStream <<
```
tehhowch, line 128:
```suggestion
	if(useShaderSwizzle)
	{
		fragmentCodeStream <<
```

source/SpriteShader.cpp
```diff
@@ -210,7 +258,10 @@ void SpriteShader::Add(const Item &item, bool withBlur)
 	// Bounds check for the swizzle value:
 	int swizzle = (static_cast<size_t>(item.swizzle) >= SWIZZLE.size() ? 0 : item.swizzle);
 	// Set the color swizzle.
-	glTexParameteriv(GL_TEXTURE_2D_ARRAY, GL_TEXTURE_SWIZZLE_RGBA, SWIZZLE[swizzle].data());
+	if (SpriteShader::useShaderSwizzle)
```
tehhowch, line 261:
```suggestion
	if(SpriteShader::useShaderSwizzle)
```

source/SpriteShader.cpp
```diff
@@ -223,5 +274,8 @@ void SpriteShader::Unbind()
 	glUseProgram(0);
 	
 	// Reset the swizzle.
-	glTexParameteriv(GL_TEXTURE_2D_ARRAY, GL_TEXTURE_SWIZZLE_RGBA, SWIZZLE[0].data());
+	if (SpriteShader::useShaderSwizzle)
```
tehhowch, line 277:
```suggestion
	if(SpriteShader::useShaderSwizzle)
```

source/SpriteShader.h
```diff
@@ -43,14 +43,17 @@ class SpriteShader {
...
+	static void Init(bool useShaderSwizzle);
 	
 	// Draw a sprite.
 	static void Draw(const Sprite *sprite, const Point &position, float zoom = 1.f, int swizzle = 0, float frame = 0.f);
 	
 	static void Bind();
 	static void Add(const Item &item, bool withBlur = false);
 	static void Unbind();
+
+private:
```
tehhowch, line 55:
```suggestion
	static void Unbind();
	
	
private:
```

tehhowch (approved):
Only style nitpicks. Thanks!

source/SpriteShader.cpp
```diff
@@ -71,12 +75,16 @@ void SpriteShader::Init()
...
-	static const char *fragmentCode =
+	ostringstream fragmentCodeStream;
+	fragmentCodeStream <<
 		"// fragment sprite shader\n"
 		"uniform sampler2DArray tex;\n"
 		"uniform float frame;\n"
 		"uniform float frameCount;\n"
-		"uniform vec2 blur;\n"
+		"uniform vec2 blur;\n";
+	if (useShaderSwizzle) fragmentCodeStream <<
```
tehhowch, line 85:
```suggestion
	if(useShaderSwizzle) fragmentCodeStream <<
```

source/SpriteShader.cpp
```diff
@@ -126,6 +172,8 @@ void SpriteShader::Init()
 	blurI = shader.Uniform("blur");
 	clipI = shader.Uniform("clip");
 	alphaI = shader.Uniform("alpha");
+	if (useShaderSwizzle)
```
tehhowch, line 175:
```suggestion
	if(useShaderSwizzle)
```

thomasballinger: Thanks + sorry about these, I'll set a formatted up!

PR #5153, opened 2020-07-06
feat: Use RGBA texures instead of BGRA for better portability

janisozaur:
## Summary
Use RGBA texures instead of BGRA for better portability, as only some variants of libjpeg support BGRA, use the more common variant.

## Testing Done
Visually checked if the textures look the same before and after my change

### DIFF
--- a/source/ImageBuffer.cpp
+++ b/source/ImageBuffer.cpp
@@ -316,7 +316,7 @@ namespace {
 		
 		jpeg_finish_decompress(&cinfo);
 		jpeg_destroy_decompress(&cinfo);
-
+		
 		// Expand RGB to RGBA
 		for (unsigned int y = 0; y < cinfo.image_height; y++)
 		{
@@ -329,7 +329,7 @@ namespace {
 				pptr[idx * 4 + 3] = static_cast<unsigned char>(0xFF);
 			}
 		}
-
+		
 		return true;
 	}
 	
--- a/source/ImageBuffer.cpp
+++ b/source/ImageBuffer.cpp
@@ -292,7 +292,7 @@ namespace {
 		int height = cinfo.image_height;
 		// If the buffer is not yet allocated, allocate it.
 		try {
-			buffer.Allocate(width, height);
+			buffer.Allocate(width, height); // ImageBuffer always allocates 32 bits per pixel
 		} catch (const bad_alloc &) {
 			jpeg_destroy_decompress(&cinfo);
 			const string message = "Failed to allocate contiguous memory for \"" + path + "\"";
@@ -312,6 +312,8 @@ namespace {
 			rows[y] = reinterpret_cast<JSAMPLE *>(buffer.Begin(y, frame));
 		
 		while(height)
+			// starts each line at corresponding rows[y] location, each line using
+			// only 3/4 of the space allocated for the line by ImageBuffer
 			height -= jpeg_read_scanlines(&cinfo, &rows.front() + cinfo.output_scanline, height);
 		
 		jpeg_finish_decompress(&cinfo);
@@ -321,6 +323,8 @@ namespace {
 		for (unsigned int y = 0; y < cinfo.image_height; y++)
 		{
 			auto pptr = reinterpret_cast<uint8_t*>(buffer.Begin(y, frame));
+			// expands the 3 byte RBG values in a line into 4-byte RGBA values
+			// using up what was empty space at the end of each line
 			for(int idx = width - 1; idx >= 0; --idx)
 			{
 				pptr[idx * 4 + 0] = pptr[idx * 3 + 0];
--- a/source/ImageBuffer.cpp
+++ b/source/ImageBuffer.cpp
@@ -248,8 +248,6 @@ namespace {
 			png_set_expand_gray_1_2_4_to_8(png);
 		if(colorType == PNG_COLOR_TYPE_GRAY || colorType == PNG_COLOR_TYPE_GRAY_ALPHA)
 			png_set_gray_to_rgb(png);
-		if(colorType & PNG_COLOR_MASK_COLOR)
-			png_set_bgr(png);
 		// Let libpng handle any interlaced image decoding.
 		png_set_interlace_handling(png);
 		png_read_update_info(png, info);
@@ -285,7 +283,7 @@ namespace {
 		
 		jpeg_stdio_src(&cinfo, file);
 		jpeg_read_header(&cinfo, true);
-		cinfo.out_color_space = JCS_EXT_BGRA;
+		cinfo.out_color_space = JCS_RGB;
 		
 		// MAYBE: Reading in lots of images in a 32-bit process gets really hairy using the standard approach due to
 		// contiguous memory layout requirements. Investigate using an iterative loading scheme for large images.
@@ -318,7 +316,20 @@ namespace {
 		
 		jpeg_finish_decompress(&cinfo);
 		jpeg_destroy_decompress(&cinfo);
-		
+
+		// Expand RGB to RGBA
+		for (unsigned int y = 0; y < cinfo.image_height; y++)
+		{
+			auto pptr = reinterpret_cast<uint8_t*>(buffer.Begin(y, frame));
+			for(int idx = width - 1; idx >= 0; --idx)
+			{
+				pptr[idx * 4 + 0] = pptr[idx * 3 + 0];
+				pptr[idx * 4 + 1] = pptr[idx * 3 + 1];
+				pptr[idx * 4 + 2] = pptr[idx * 3 + 2];
+				pptr[idx * 4 + 3] = static_cast<unsigned char>(0xFF);
+			}
+		}
+
 		return true;
 	}
 	
--- a/source/Sprite.cpp
+++ b/source/Sprite.cpp
@@ -71,7 +71,7 @@ void Sprite::AddFrames(ImageBuffer &buffer, bool is2x)
 	// Upload the image data.
 	glTexImage3D(GL_TEXTURE_2D_ARRAY, 0, GL_RGBA8, // target, mipmap level, internal format,
 		buffer.Width(), buffer.Height(), buffer.Frames(), // width, height, depth,
-		0, GL_BGRA, GL_UNSIGNED_BYTE, buffer.Pixels()); // border, input format, data type, data.
+		0, GL_RGBA, GL_UNSIGNED_BYTE, buffer.Pixels()); // border, input format, data type, data.
 	
 	// Unbind the texture.
 	glBindTexture(GL_TEXTURE_2D_ARRAY, 0);
--- a/source/text/Font.cpp
+++ b/source/text/Font.cpp
@@ -261,7 +261,7 @@ void Font::LoadTexture(ImageBuffer &image)
 	glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_WRAP_T, GL_CLAMP_TO_EDGE);
 	
 	glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, image.Width(), image.Height(), 0,
-		GL_BGRA, GL_UNSIGNED_BYTE, image.Pixels());
+		GL_RGBA, GL_UNSIGNED_BYTE, image.Pixels());
 }
 
 


janisozaur:
I also cross-checked against #5152 and it works fine.

petervdmeer:
This PR also lacks some background information.  @janisozaur Can you add the following information to the PR:
- Which versions of libjpeg lack support for BGRA? (Which operating systems, which versions of the libraries, is the libjpeg library upgradable on those systems?)
- Is there a performance impact when switching from BGRA to RGBA? (And if so, then which video-cards would have which impact? Is this operating system specific? How much would the impact be? How often would we get a performance hit, only at loading or also while the game is running?)

If changing from BGRA to RGBA causes ES to run properly on more platforms, then I will be very happy with this change. But if this change also causes a performance degradation on current platforms, then I would like to know the actual performance impact before I decide if I'm happy with it.

tux2603:
In my (somewhat limited) experience with computer graphics, BGRA is used just as much if not more than RGBA. There shouldn't be any major performance impact of changing it though.

I checked some libjpeg, and all the ones that I've found have supported BGRA since 2015 or before. I wouldn't say it's an overly large portability issue, but I'll let @janisozaur chime in with the config that it doesn't

janisozaur:
@tux2603 I have no idea what you checked, as libjpeg has no support for BGRA and can only RGB. 

@petervdmeer as mentioned above, all libjpeg versions don't support BGRA.

MCOfficer:
> as only some variants of libjpeg support BGRA

> all libjpeg versions don't support BGRA.

I'm confused.

janisozaur:
I can understand your confusion. Some [variants](https://dictionary.cambridge.org/dictionary/english/variant) of libjpeg support BGRA, but libjpeg doesn't.

MCOfficer:
> I can understand your confusion. Some [variants](https://dictionary.cambridge.org/dictionary/english/variant) of libjpeg support BGRA, but libjpeg doesn't.

The link wasn't necessary. Anyways, are we talking about niche cases or are these variants (whatever they may be) commonplace? What about libjpeg-turbo specifically?

janisozaur:
libjpeg-turbo specifically supports BGRA, you use this custom extension already.

Does "[being shipped with Debian](https://packages.debian.org/sid/libjpeg9)" fulfil your definition of "commonplace"? Does "[the first result in Google](https://user-images.githubusercontent.com/550290/88775479-0bf9d780-d185-11ea-8f93-1ff25fa812c9.png)" do? [Having its own Wikipedia entry](https://en.wikipedia.org/wiki/Libjpeg)?

MCOfficer:
Sorry if that was ambiguous. I was interested in what variants *do* support BGRA, since you already stated that libjpeg does not. I guess that collided with peter's question earlier:

> Which versions of libjpeg lack support for BGRA? (Which operating systems, which versions of the libraries, is the libjpeg library upgradable on those systems?)

janisozaur:
I do not know what variants of libjpeg, other than the (most?) popular libjpeg-turbo, support BGRA. I can't speak for other users, but I can imagine replacing such common library would be a daunting process, especially if there are custom extensions in use. The common denominator presented here should work with more variants -- especially with libjpeg.

petervdmeer:
I looked into this a bit more; the GL-format that is being changed here is the format for transferring pixel data to textures, not the internal texture format. So any performance hit (or performance improvement) would only be at texture load, not at texture use.

GL_RGBA is also the default for OpenGLES 2 (where GL_BGRA is only available as an extention), so this change also improves portability from the GL-side of things.

This link provides some nice background information; https://community.khronos.org/t/why-opengles-2-spec-doesnt-support-bgra-texture-format/72853

I'm now more in favor of this change than I was before, since the positive impact on portability looks bigger than the the negative impact of the (load-time-only) hit on performance.

Thanks @thomasballinger for allowing me to efficiently review without whitespace issues. @janisozaur please merge Thomas' whitespace fix or fix whitespace issues directly here so that we can properly review this PR without seeing hundres of unrelated changed lines.

thomasballinger:
Looks like I have some testing to do with ubuntu-20.04 and libglew2.1. Any feedback would be welcome in the meantime, especially any pointers to what might be different about that environment, I'm guessing it's the libglew version.

source/ImageBuffer.cpp
```diff
@@ -300,6 +298,19 @@ namespace {
 		jpeg_finish_decompress(&cinfo);
 		jpeg_destroy_decompress(&cinfo);
 		
+		// Expand RGB to RGBA
+		for (unsigned int y = 0; y < cinfo.image_height; y++)
+		{
+			auto pptr = reinterpret_cast<uint8_t*>(buffer.Begin(y, frame));
```
petervdmeer, line 304: When going from RGB (three 8-bit values) to RGBA (four 8-bit values), then I would expect different starting points for all lines after the first line. (And I would expect the for-loop above to loop from height back to 0 to avoid the risk of overwriting existing data).
Also, isn't there some library function available in libJPEG to handle this conversion (since I can imagine that different platforms actually store things differently)?

janisozaur: it already does strides in reverse order - see the loop below.

What do you mean by "different starting points for all lines after the first line"? I request the pointer to a specific line in the buffer already, but I feel you meant something else?

> some library function available in libJPEG to handle this conversion

Honestly haven't thought of that, but I can check - my reasoning was that if such a format was supported, user would be able to request it explicitly when decoding (like libjpeg-turbo does) to avoid duplicate work, so it makes little sense to add such a trivial function to the library.

janisozaur: Haven't seen anything interesting in https://github.com/emscripten-ports/SDL2_image/blob/master/external/jpeg-9/jpeglib.h

petervdmeer: I was surprised that the strides all start at the same location. (Both the source and target strides seem to start from the `auto pptr`.) This suggests that the RGB image is having much larger strides than it would normally need.

petervdmeer: > > some library function available in libJPEG to handle this conversion
> 
> if such a format was supported, user would be able to request it explicitly when decoding (like libjpeg-turbo does) to avoid duplicate work, so it makes little sense to add such a trivial function to the library.

Such a function would be trivial, until we get to a point where we need to compensate for differences in platforms (big-endian vs little-endian, 8-bit aligned vs 32-bit aligned vs 64-bit aligned memory layouts). The least that needs to be done when writing those functions ourselves is to test them for all supported platforms (to verify that we didn't miss anything platform specific).

thomasballinger: > I was surprised that the strides all start at the same location. (Both the source and target strides seem to start from the auto pptr.) This suggests that the RGB image is having much larger strides than it would normally need.

@petervdmeer, that's right, the code above here uses `cinfo.out_color_space = JCS_RGB;` so 24-bit pixels, but in the ImageBuffer class the space for image data is always 4 bytes per pixel:
https://github.com/endless-sky/endless-sky/blob/ffd6053b9c475a01ed48b4758489ca17f079e3f2/source/ImageBuffer.cpp#L68

so `buffer.Begin(y, frame))` will be the same for both encodings, but the location of each pixel in the line will be different. Would you suggest changing this, maybe by using an out_color_space of JCS_RGBX if that exists? Or maybe just a comment describing this?

Regarding tests, I could use a pointer on how to write tests for this; is it just a new unit test that decodes an image and tests whether a couple pixels match what we expect?

petervdmeer: A comment to describe why the start positions are the same would be sufficient for me.

A unit test that decodes an image and checks a couple of pixels is indeed the test-size I was thinking about. The nice thing of such unit-test is that we can automatically run it on all platforms where ES is supported (even on the web-assembly version if you like).
Here is a nice [starting point for an unit testing framework](https://github.com/endless-sky/endless-sky/blob/master/tests/src/README.md) that is already in use for ES.

thomasballinger: I'm looking into this now, it seems to me I'll need to change the way the tests are run to link in libjpeg.

tehhowch: You won't be able to use third party code such as libpng or jpg-turbo from the tests without linking in their code, which is intentionally not done for a unit test. Instead one should provide a mock with the appropriate endpoints to be consumed by the class, which could then be inspected to ensure the code under test used the library correctly.

I haven't delved deeply into how exactly one would provide a mock library implementation in the tests that would be used by the game's class with our setup, but I imagine it would not be trivial. Isolating each test's code from other tests naturally means you also isolate it from the game classes' code, so you'd need to write an actual shared library and link it, which means updating scons and codeblocks projects so that it works on CI and also for windows devs using CB instead of scons.

I'm totally open to implementing our unit tests in some other framework or manner that makes it easier to test, but a core requirement I have is that the tests have to be manageable and runnable from CodeBlocks.

thomasballinger: Since it sounds like this will be some work, @petervdmeer could I convince you that endianness won't be a problem because these are all single-byte values and that regardless of alignment, RGB output colorspace will always be 24 bits? I'm not sure about that second one but reading around it seems to be true.

petervdmeer: As I said before: A number of good comments in the code on why this works is sufficient for me. The tests would be a nice extra, but not required.

source/ImageBuffer.cpp
```diff
@@ -314,11 +312,28 @@ namespace {
...
+			// expands the 3 byte RBG values in a line into 4-byte RGBA values
+			// using up what was empty space at the end of each line
+			for(int idx = width - 1; idx >= 0; --idx)
```
petervdmeer, line 328: Do we also have an idea of how much performance costs (at loading-time) we have because of the additional loop that is now needed? (Or performance gain, but that seems a bit unlikely.)

janisozaur:
I looked into the tests and they don't compile.
```
$ scons mode=debug -j 8 test
scons: Reading SConscript files ...
scons: done reading SConscript files.
scons: Building targets ...
g++ -o tests/endless-sky-tests tests/build/helpers/datanode-factory.o tests/build/text/test_alignment.o tests/build/text/test_displaytext.o tests/build/text/test_format.o tests/build/text/test_layout.o tests/build/text/test_truncate.o tests/build/test_account.o tests/build/test_conditionSet.o tests/build/test_datanode.o tests/build/test_esuuid.o tests/build/test_main.o tests/build/test_point.o tests/build/test_random.o tests/build/test_set.o tests/build/test_ship.o tests/build/test_weightedList.o lib/debug/libendless-sky.a -luuid
/usr/bin/ld: lib/debug/libendless-sky.a(Files.o): in function `Files::Init(char const* const*)':
/home/xx/workspace/endless-sky/source/Files.cpp:80: undefined reference to `SDL_GetBasePath'
/usr/bin/ld: /home/xx/workspace/endless-sky/source/Files.cpp:85: undefined reference to `SDL_free'
/usr/bin/ld: /home/xx/workspace/endless-sky/source/Files.cpp:125: undefined reference to `SDL_GetPrefPath'
/usr/bin/ld: /home/xx/workspace/endless-sky/source/Files.cpp:133: undefined reference to `SDL_free'
/usr/bin/ld: /home/xx/workspace/endless-sky/source/Files.cpp:151: undefined reference to `SDL_GetPrefPath'
/usr/bin/ld: /home/xx/workspace/endless-sky/source/Files.cpp:153: undefined reference to `SDL_free'
collect2: error: ld returned 1 exit status
scons: *** [tests/endless-sky-tests] Error 1
scons: building terminated because of errors.
```

It seems all your tests in CI are executed in release mode, but I'm not interested in that.

The patch to make it work looks like this:
```diff
diff --git a/SConstruct b/SConstruct
index ac7e571e..af085ddf 100644
--- a/SConstruct
+++ b/SConstruct
@@ -145,7 +145,7 @@ test = env.Program(
         # Add Catch header & additional test includes to the existing search paths
        CPPPATH=(env.get('CPPPATH', []) + [pathjoin('tests', 'include')]),
        # Do not link against the actual implementations of SDL, OpenGL, etc.
-       LIBS=sys_libs,
+       LIBS=sys_libs + game_libs,
        # Pass the necessary link flags for a console program.
        LINKFLAGS=[x for x in env.get('LINKFLAGS', []) if x not in ('-mwindows',)]
 )
```

Should I submit it?

Having looked at the tests you have so far I see following constraints:
1. There are no tests that read files
2. Hence there is no canonical `testdata` path to use
3. The APIs in setting up images all operate on files as sources, there is no way to create a memfd and pass it to buffer, leaving the only viable way forward with setting up `testdata` and having it read files. Refactoring the code to accept memory buffers falls outside of scope of this PR

@petervdmeer what do you suggest?

janisozaur:
Submitted https://github.com/endless-sky/endless-sky/pull/6147

petervdmeer:
> I looked into the tests and they don't compile.
> 
> @petervdmeer what do you suggest?

The tests you are referring to are part of the unit-testing framework, I have more experience with the integration testing framework. But having said that; I've posted a review-comment on the fix you submitted with some information which I expect to be relevant.

thomasballinger:
@janisozaur I deleted the tests (rebased and removed the commit where I added them) here because it didn't look feasible to actually test the behavior. I'm now looking into compiling libjpeg-turbo with emscripten, if I get can get that working I won't personally be as interested in moving from BGRA to RGBA.

I was also confused by the tests at first, I'm also more familiar less isolated tests.

thomasballinger:
I gave compiling a libjpeg-turbo to WebAssembly a shot, and if I understand right we still need to `GL_BGRA` -> `GL_RGBA` because that's [a GL ES thing](https://community.khronos.org/t/why-opengles-2-spec-doesnt-support-bgra-texture-format/72853), not a turbo-jpeg thing.

Would load time be the right thing to run here to see how much it affects things?

thomasballinger:
I did a rough benchmark of first SpiteQueue::Add call to finishing loading and did not observe any slowdown.

image

data and plot: https://observablehq.com/d/1ed6d4c43d2369d0

source/ImageBuffer.cpp
thomasballinger: Performance gain seems possible! average 5% faster, not significantly different with 9 observations each. https://observablehq.com/@ballingt/timing-data-for-endlesssky-5153

tehhowch:
Jpeg loading is largely deferred to runtime, rather than being processed in startup and kept in memory for the duration of the game session. Changes to loading PNG files will have a much more pronounced effect on game startup, and can be stressed by also installing the high DPI plugin.

If jpg timings are desired, you'll be better off creating a custom game entry point and ensuring the images are not deferred.

source/ImageBuffer.cpp
```diff
@@ -285,7 +283,7 @@ namespace {
 		
 		jpeg_stdio_src(&cinfo, file);
 		jpeg_read_header(&cinfo, true);
-		cinfo.out_color_space = JCS_EXT_BGRA;
+		cinfo.out_color_space = JCS_RGB;
```
petervdmeer, line 286: As discussed on discord; if libjpeg-turbo is working for all platforms, then what is preventing us from using JCS_EXT_RGBA directly (and avoiding the extra loop below)?

thomasballinger: Seems like that should work, I've just been looking at trying to measure load time instead. I make sure libjpeg-turbo works for the emscripten build.

petervdmeer: Ok

thomasballinger: Here's that much smaller change: https://github.com/endless-sky/endless-sky/pull/6155

thomasballinger:
This can be close in favor of https://github.com/endless-sky/endless-sky/pull/6155.

PR #5155, opened 2020-07-09
Trivial update in credits.txt

comnom:
I prefer a real name rather than an account name. Been following the project sporadically, impressed at the work. Don't really think I should get a programming credit for minor edits, but you folks made the call. Keep on man, and good luck.

### DIFF
--- a/credits.txt
+++ b/credits.txt
@@ -20,12 +20,12 @@ Developers
 Major Programming
   AdamKauffman
   Amazinite
-  Frederick Goy IV
   CyberShadow
   db47h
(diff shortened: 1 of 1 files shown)


credits.txt
```diff
@@ -20,7 +20,7 @@ Developers
 Major Programming
   AdamKauffman
   Amazinite
-  comnom
+  Frederick Goy IV
```
northeast-eph, line 23: Should this also be moved down to beneath flavious to preserve alphabetical order?

Issue #1564, opened 2016-09-13
Allow mission directives inside conversations (payment, event, npc, outfit, reputation, etc)


Pointedstick:
I'm trying to write a gambling mission and feel like I'm fighting the system (#1563, #1562). As an alternative to addressing either of those concerns, it to occurs to me that all I really want to do is be able to change the player's money within a conversation. Essentially I'd like to be able to do this:

```
mission "Pirate gambling"
    description "Gamble with pirates."
    minor
    repeat
    to offer
        random < 5
    source
        attributes "dirt belt" south rim

on offer
    conversation
            `Do you want to gamble with pirates?`
            choice
                `No way!`
                    defer
                `Absolutely!`
                    goto gamble

        label gamble
        branch "good luck" "bad luck"
            random < 50

        label "good luck"
        `You win some money!`
            payment 5000
            defer

        label "bad luck"
        `The pirates are cheating! Do you call them out on it?`
         choice
             `(Sure, what could possibly go wrong?)`
                 goto "bad idea"
             `(Keep your mouth shut.)`
                 goto "lose"

        label lose
        `You lose some of money to the pirates!`
            payment -10000
            defer

        label "bad idea"
        `The pirates pull out their guns and demand all the money you're carrying!`
        choice
            `(Comply.)
                 goto "lose a lot"
             `(Grab the pot and run for it!)
                 goto "really really bad idea"

        label "lose a lot"
        `The pirates take a lot of your money!`
             payment -50000
             defer

        label "really really bad idea"
        `The pirates shoot you dead!`
             die
```

You can see how it would be useful to be able to add or subtract credits at multiple points, not just in the "on complete" or "on fail" sections.

Same things with NPCs; I'd like to have some outcomes spawn NPC ships, and some not. The current mechanism of "on accept" and "on decline" isn't nearly flexible enough to accommodate this (see #1563).

KiLEdEnNis:
"The pirates shoot you dead!"
not sure if that grammar is correct

toilethinges:
Being shot dead is, remarkably, correct.

Amazinite:
Not just payments, but any condition really. Events, rep, etc.

KiLEdEnNis:
@toilethinges thanks
in Dutch (my language) it would be "de piraten schieten je dood", literally translated word by word you would get "the pirates shoot you dead", and I'm not used to Dutch grammar in English

EndrosG:
I would like to add something to this enhancement issue.

[...]

PR #5007, opened 2020-04-22
Add "Earth Retirement"

Naminson:
**Content (Mission)**

## Summary
This adds a short mission where you bring an aging couple to Earth for a low payment that you have the option to refuse. This is something that wiki says is very common, but so far is never mentioned in game. Hopefully, this PR can change that. 
This mission offers anywhere in the Paradise region.

PR #5157, opened 2020-07-11
Allow conversations to grant payments, trigger events, create log entries, and more

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #1564.

## Feature Details
Greatly extends the capabilities of the `apply` conversation node using a new GameAction class. `apply` can not only alter conditions, but also give payments, fines, events, logs, fail missions, gift ships, and gift/remove outfits. 
```
conversation
	...
	apply
		payment [<#payment> [<#multiplier]]
		fine <#fine>
		event "<event name>" [<#min delay> [<#max delay>]]
		log [<category> <header>] <text>
		fail [<mission name>]
		give ship <ship> [<#number>]
		outfit <outfit> [<#number>]
		<condition> <comp> <value>
		...
	...
```
All of these functionalities have been moved from MissionAction to the GameAction class, and MissionAction now inherits GameAction. This also means that any new functions added to GameAction in the future are automatically able to be done in `apply` nodes, but such functions should be safe for use in conversations.

If a payment is negative and the player doesn't have enough credits in their account for the payment, then instead of blocking the mission (as `payment` in a MissionAction currently does) the player's account is reduced to 0. This can be accounted for in a conversation by using a branch to see if the player's credits are above a certain point. (See the example.)

Similarly, an `apply` node that removes outfits that the player doesn't have won't stop the mission from triggering. Mission writers should make sure that the player has any outfits a conversation could remove should they reach such a node.

## UI Screenshots
N/A

## Usage Examples
See testing done.

## Testing Done
<details><summary>This mission works as intended.</summary>

Using the same conversation as used in #1564. This conversation would either grant you credits or remove them based off of your luck, but if you get lucky, you'll trigger a delayed event that does something for the pirates to get back at you for winning. Also threw some logs and conditions in there just to see that they work.

```
event "angered pirate gamblers"

mission "angered pirates"
	landing
	to offer
		has "event: angered pirate gamblers"
	on offer
		conversation
			`The event trigger worked.`
				decline

mission "conversation test"
	repeat
	source
		near "Sol" 100
	destination Luna
	passengers 2
	on offer
		conversation
			`Do you want to gamble with pirates?`
			apply
				set "meme magic"
				log "Conversation" "Actions"
					"Hey, look at me."
					"Hey, listen."
			choice
				`	No way!`
					goto end
				`	Absolutely!`
					goto gamble

			label gamble
			branch "good luck" "bad luck"
				random < 50

			label "good luck"
			apply
				payment 5000
				log "You had good luck on this day."
				event "angered pirate gamblers" 1 100
			`	You win some money!`
				goto end

			label "bad luck"
			apply
				log "You had bad luck on this day."
			`	The pirates are cheating! Do you call them out on it?`
			choice
				`	(Sure, what could possibly go wrong?)`
					goto "bad idea"
				`	(Keep your mouth shut.)`
					goto "lose"

			label lose
			branch robbed
				credits < 10000
			apply
				payment -10000
			`	You lose some of money to the pirates!`
				goto end

			label "bad idea"
			`	The pirates pull out their guns and demand all the money you're carrying!`
			choice
				`	(Comply.)`
					goto "lose a lot"
				`	(Grab the pot and run for it!)`
					goto "really really bad idea"

			label "lose a lot"
			branch robbed
				credits < 50000
			apply
				payment -50000
			`	The pirates take a lot of your money!`
				goto end

			label robbed
			apply
				payment -50000
			`	The pirates rob you for all you have!`
				goto end

			label "really really bad idea"
			`	The pirates shoot you dead!`
				die
			
			label end
			`	Now head to <destination>.`
				accept
	
	on complete
		conversation
			`Testing, testing, one, two.`
			apply
				payment
				payment 10000
			`	Thank you for testing..`
```
</details>
Existing jobs and missions appear to work as they did before, meaning that the creation of GameAction has had no negative effect elsewhere.

## Performance Impact
N/A


source/ConversationPanel.cpp
```diff
@@ -339,6 +340,33 @@ void ConversationPanel::Goto(int index, int choice)
 			// Update any altered government reputations.
 			player.CheckReputationConditions();
 		}
+		else if(conversation.IsPayment(node))
+		{
+			int64_t payment = conversation.Payment(node);
+			int64_t account = player.Accounts().Credits();
+			// If the payment is negative and the player doesn't have enough
+			// in their account, then the player's credits are reduced to 0.
```
petervdmeer, line 348: If the player has a large negative account, and the payment is small positive (or even negative), then the player gains money in this action.

Amazinite: If the player has a large negative account, and the payment is small positive (or even negative), then the player gains money in this action. The player's account is never allowed to go negative at any point in the game. From the comment on Account::AddCredits:
> // Give the player credits (or pass  negative number to subtract). **If subtracting,
> // the calling function needs to check that this will not result in negative credits.**

Amazinite:
> A generic remark is that some actions from MissionAction appear to be re-implemented here, resulting in some code/functionality duplication. Would it be possible to make the shared actions (and applying them) generic and shared between Conversations and MissionActions?

I thought of that while working on this, but I'm unsure of a simple way to do that. What's being done by MissionAction and Conversation for payment, events, and logs isn't terrible complex, so is a removal of the duplication all that necessary?

source/ConversationPanel.cpp
petervdmeer: Ok, but if we change the else-statement to `else if(account > 0)` then we are also safe if some bug somewhere else in the code does causes the account to go negative. (Or if we do allow negative numbers in the future).

petervdmeer:
I was thinking about a superclass for MissionAction and Node that contains the common parts, including the common load, save and apply code or about embedding a MissionAction in the Node.

If we go for the superclass, then it could have a common Load function that would consume keywords that it recognizes and return whenever it encounters a specific keyword. If we have a common superclass for MissionAction and Node, then we could check if we can also use it for events (but that can be for a different PR).

petervdmeer:
If we do combine Events in the discussion, then we can consider to also set visited and unvisited Planets and Systems in MissionActions and in Nodes. (And create Log-entries and Payments in Events.) But this might also be better kept for a different PR.

Amazinite:
If we do put a MissionAction inside of a conversation node, then perhaps we should change the syntax to something like this:
```
conversation
    action
        payment...
        event...
```
And since a MissionAction can also apply conditions, this could effectively replace the `apply` node.

source/Conversation.cpp
```diff
@@ -256,6 +265,16 @@ void Conversation::Save(DataWriter &out) const
 				out.EndChild();
 				continue;
 			}
+			if(!node.actions.IsEmpty())
+			{
+				out.Write("action", TokenName(node.data[0].second));
```
petervdmeer, line 270: Does the action keyword need a TokenName? It doesn't seem to be read during loading.

source/Conversation.cpp
```diff
@@ -298,6 +317,25 @@ Conversation Conversation::Substitute(const map<string, string> &subs) const
...
 			choice.first = Format::Replace(choice.first, subs);
+	
+	return result;
+}
+
+
+
+// Do text replacement throughout this conversation, and instantiate any
+// potential actions.
+Conversation Conversation::Instantiate(map<string, string> &subs, int jumps, int payload) const
```
petervdmeer, line 328: Do we need both the Instantiate and the Substitute function?
If Instantiate is the better/more-complete version that can completely replace Substitute, then please just remove Substitute.

source/GameAction.cpp
```diff
@@ -0,0 +1,228 @@
...
+GameAction::GameAction(const DataNode &node, const string &missionName)
+{
+	Load(node, missionName);
+}
+
+
+
+void GameAction::Load(const DataNode &node, const string &missionName)
+{
+	empty = false;
```
petervdmeer, line 41: What if we have just an action keyword, but no child entries below it. Shouldn't that be considered empty? (Set empty to false in LoadSingle?)

And do we even need the empty variable? Cannot we check the individual data-items to see if they contain data in the IsEmpty() check?

source/GameAction.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+
+
+// Load a single child at a time, used for streamlining MissionAction::Load.
+void GameAction::LoadSingle(const DataNode &child, const string &missionName)
+{
+	const string &key = child.Token(0);
+	bool hasValue = (child.Size() >= 2);
+	
+	if(key == "log")
```
petervdmeer, line 54: Needs to also check hasValue (since Token(1) is accessed).

source/GameAction.cpp
```diff
@@ -0,0 +1,228 @@
...
+			out.BeginChild();
+			{
+				// Break the text up into paragraphs.
+				for(const string &line : Format::Split(eit.second, "\n\t"))
+					out.Write(line);
+			}
+			out.EndChild();
+		}
+	if(payment)
+		out.Write("payment", payment);
```
petervdmeer, line 117: Should this also write out the paymentMultiplier if there is one? (Probably not, since it gets handled in instantiate, but maybe add a comment here that the paymentMultiplier is never used in dynamic data.)

source/GameAction.cpp
```diff
@@ -0,0 +1,228 @@
...
+	for(const auto &it : specialLogText)
+		for(const auto &eit : it.second)
+			player.AddSpecialLog(it.first, eit.first, eit.second);
+	
+	if(payment)
+	{
+		// Conversation actions don't block a mission from offering if a
+		// negative payment would drop the player's account balance below
+		// zero, so payment is handled differently within conversations.
+		if(conversation)
```
petervdmeer, line 162: You could use the Conversations code both for the conversations and the non-conversations case (since you are sure in the non-conversations case that `account + payment >= 0` is always true).

source/MissionAction.h
```diff
@@ -72,26 +74,30 @@ class MissionAction {
...
 	Phrase dialogPhrase;
 	
 	const Conversation *stockConversation = nullptr;
 	Conversation conversation;
 	
-	std::map<const GameEvent *, std::pair<int, int>> events;
 	std::map<const Outfit *, int> gifts;
 	std::map<const Outfit *, int> requiredOutfits;
+	
+	GameAction actions;
```
petervdmeer, line 87: The MissionAction now has a GameAction inside it. By itself this is fine, but I was expecting that the MissionAction would actually inherit from the GameAction class instead of embedding a GameAction class inside it (Similar to how a Ship inherits from the Body class).

If you have good reasons for choosing this pattern, then that is fine for me, but maybe add a comment on why the "contain pattern" instead of the "inherit pattern" was choosen.

source/MissionAction.h
```diff
@@ -72,26 +74,30 @@ class MissionAction {
...
+	/* Now handled by GameAction
```
petervdmeer, line 88: This out-commented code makes sense during the PR, but will just be out-commented code after this gets merged. I suggest removing all the out-commented code (and maybe put some comment above GameAction on what the GameAction handles).

source/Conversation.cpp
Amazinite: This is the same behavior as branch and apply. Instead of needing a `goto <label>` (which is something I pointed out about the prior implementation), the conversation understands the second argument here as the goto location

source/Conversation.cpp
Amazinite: Instantiate is only called by MissionAction, but Substitute gets called at other points in the code. (I forget exactly where, but I did run into them when I tried to only have one function. Substitute also takes in a const map, while Instantiate is just a map, which was important for those places where Substitute is still used now.)

source/GameAction.cpp
Amazinite: This is handled by the IsSpecial bool.

source/MissionAction.h
Amazinite: The difference between Ship/Body and MissionAction/GameAction is that Body gives new functions to Ship, but MissionAction has functions that go by the same name as those in GameAction while doing the same thing as them and more. I did actually have MissionAction inherit GameAction at first, but this seemed to result in duplicating code because of the functions of the same name. (Perhaps there’s simply some process to avoid that that I don’t know about, though.)

source/GameAction.cpp
Amazinite: GameAction is basically just ripped out of MissionAction, so if it works in the game right now then it’ll work in this PR. :)

source/GameAction.cpp
Amazinite: Checked every potential thing in a GameAction to see if it has changed seemed clunkier to me than just seeing if the GameAction has ever been loaded.

source/GameAction.cpp
petervdmeer: You're right. I didn't read the code properly.

source/GameAction.cpp
petervdmeer: Yes, it is clunkier, but it also feels more correct to base the check on actual data instead of on metadata like having passed the load-event.

It's not a big deal to me though. Please feel free to keep the boolean as-is.

source/MissionAction.h
petervdmeer: Functions that have the same name but do something more sound like something for which the "super()" functionality in Java was invented. Super() allows a function to call the ancestors version of itself and then add something more around it (before or after calling super()). C++ also has something similar, but you need to specify the name of the base class due to multiple-inheritance.

Amazinite: Yeah, I recall Super() in Java. So how might that be done here?

source/Conversation.cpp
petervdmeer: I'm not too familiar with this part of the Conversations code and the code looked a bit strange, but I just looked at `branch` and `apply` and they look equally strange so for now I will just believe you that this code here is correct.

source/Conversation.cpp
petervdmeer: I would expect that `subs` could also be given as const-map in Instantiate, because I would expect Instantiate to never change data in the map itself.

Other uses of Substitute that I just found are in main.cpp, MissionAction.cpp, NPC.cpp and I don't really see a reason for them to perform text-substitutes but not action-initialisation (with some sensible defaults).

source/GameAction.cpp
```diff
@@ -0,0 +1,228 @@
...
+	
+	// Check if applying the conditions changes the player's reputations.
+	player.SetReputationConditions();
+	conditions.Apply(player.Conditions());
+	player.CheckReputationConditions();
+}
+
+
+
+GameAction GameAction::Instantiate(map<string, string> &subs, int jumps, int payload) const
```
petervdmeer, line 197: I would expect subs to be const here as well.

petervdmeer: Sorry, just noticed that subs is changed here due to payment being set. That makes the code a little bit more complex indeed.

source/Conversation.cpp
Amazinite: I’ll look into those other instances to see if Substitute can be entirely replaced.

source/MissionAction.h
petervdmeer: My first thoughts on this are:
- make the class MissionAction inherit from GameAction.
- make LoadSingle() in GameAction.h virtual and override it in MissionAction.
   - process MissionAction-specifics in MissionAction::LoadSingle()
   - call GameAction::LoadSingle() from MissionAction::LoadSingle() for every keyword not handled in MissionAction.
   - Keep Load() only in GameAction (and not virtual).
- make Save() in GameAction.h virtual and override it in MissionAction.
   - call GameAction::Save() from MissionAction::Save()
- make Do() in GameAction.h virtual and override it in MissionAction.
   - Call GameAction::Do() from MissionAction::Do() after or before Doing the MissionAction specific things.
- Instantiate() cannot be virtual and overriden since they return different objects. There you could look at using a constructor for the returned object that calls baseclass constructors.

source/Conversation.cpp
Amazinite: It looks like the ones in NPC should be Instantiate, but I'm not sure what's going on with the LoadConversation function in main. Looks like it's some sort of launch option that gets you there. That could be Instantiate as well, since that doesn't actually affect the game.

source/GameAction.cpp
```diff
@@ -0,0 +1,223 @@
...
+	{
+		// Allow randomization of event times. The second value in the pair is
+		// always greater than or equal to the first, so Random::Int() will
+		// never be called with a value less than 1.
+		int day = it.second.first + Random::Int(it.second.second - it.second.first + 1);
+		result.events[it.first] = make_pair(day, day);
+	}
+	result.payment = payment + (jumps + 1) * payload * paymentMultiplier;
+	// Fill in the payment amount if this is the "complete" action.
+	if(result.payment)
```
petervdmeer, line 208: How do we ensure that filling out the payment is only done for the complete action? We should not overwrite payment with an in-conversation payment.

petervdmeer: Btw, if we take this out of the Instantiate and just handle this outside this function for the "on complete" section only, then we can make the subs map const again.

Amazinite: This is another instance of this simply being something that was a part of the MissionAction, but I think we might want to keep this here. I mentioned how you can't use a \<payment> text replacement to get the payment from an action, but if subs is non-const, then an action node can change the payment replacement for the text nodes after it.

petervdmeer: The thing is, if I'm in the middle of a conversation, then I might want to refer both to the payment on the end (on complete) and to the payment in a conversation node; for example in a sentence like:
"Listen, I have a job for you, you get 5000 credits upfront to buy some weapons, and you will get <payment> if you deliver the weapons to the shady figure waiting for you at <destination>."

Amazinite:
Missions to update using a payment `action`:
* Earth Retirement (#5007)
* Saryd Couple (#4934)
* Lost Boy

source/MissionAction.h
Amazinite: It still needs cleaned up some, but how does the recent commit (e13190b) look?

source/GameAction.cpp
Amazinite: As I understand it, the current behavior is as follows:
* `subs` is created in Mission::Instantiate and filled in with various substitutions that don't get changed.
* NPCs and their conversations/dialogs get instantiated.
* The `on complete` conversation gets instantiated. This fills in the `<payment>` and `<fine>` subs.
  * NPC conversations can't use the `<payment>` and `<fine>` subs because of this? (This could be fixed by cherry picking the payment and fine out of the `on complete` action before doing NPC or MissionAction instantiations, as just reversing the order of instantiations would mean that MissionActions couldn't use the `<npc>` sub.)
* The rest of the `on *` actions are instantiated. Each action, if it has its own payment or fine, replaces the `<payment>` or `<fine>` substitutions with its own values before instantiating its conversation/dialog, then replaces the subs with the old values. This means that any `on *` action with a payment can't refer to the `on complete` mission payment, but other `on *` actions that don't have a payment could.
* NEW BEHAVIOR: if a conversation action has a payment or fine, it will replace the `<payment>` or `<fine>` subs. This means that you can not refer to any `on *` action payment after a conversation action has replaced the payment. The `on complete` subs for `<payment>` and `<fine>` are still restored after the conversation ends.

So we have three types of payments here that we might want to refer to: the `on complete` payment, the current `on *` payment, and the most recent `action` payment.

Amazinite: I've opened issue #6258 for tracking this discussion. This issue would also apply to the new payment method added by conversation actions.

Amazinite:
Since `apply` can be replaced with `action`, should I just make it so that the keyword for using a GameAction is `apply`? All old `apply` nodes would function as usual since they're just conditions that `action` can understand.

source/MissionAction.h
petervdmeer: I briefly reviewed the latest version of the PR (at commit fc8478a86cf7fcb1720fe5193dd3ac2a5386a0e5) and my first impression is that it looks quite good.
I didn't look into all the details yet; that requires me to review a bit more in-depth, but you can consider this conversation done/handled.

source/GameAction.cpp
petervdmeer: Fine for me to keep this for a follow-up PR and track this in #6258.

petervdmeer:
> Since `apply` can be replaced with `action`, should I just make it so that the keyword for using a GameAction is `apply`? All old `apply` nodes would function as usual since they're just conditions that `action` can understand, so on the mission creating side of things this would just look like expanding the functionality of the `apply` node.

I just checked your spec, and it looks like every keyword that is not recognized automatically gets interpreted as a condition-name. This can become tricky if we add new keywords to actions later and it can be problematic if I have a condition that has the same name as a keyword.

I feel it is more correct to have the `apply` node only work with conditions and to have the `action` node not support conditions.

Amazinite:
> every keyword that is not recognized automatically gets interpreted as a condition-name.

This was the previous behavior with MissionActions, so the same potential issue has existed with MissionActions for some time, and splitting conditions from a conversation action wouldn’t change that. 

I see this as a non-issue though since adding new keywords for MissionAction/GameAction isn’t common and such keywords are often very particular, but conditions can be absolutely anything, so the odds of us choosing a new keyword that is the exact same as a condition that someone previously used is uncommon, and even if such a situation were to occur then the condition can simply be changed since there’s effectively no limit on what the condition could be.

petervdmeer (approved):
The code-change looks good.

Amazinite:
I'm going to retest this and post a new test mission that makes sure that `apply` still changes conditions as it did before, tests all the new features, and demonstrates the `<payment>` text replacement behavior. And given that I don't find any issues, I'd say this is good to go.

Amazinite:
Looking good. Normal mission actions are still working and all the new conversation mechanics are working as expected.
```
mission "GameAction testing"
	landing
	on offer
		give ship "Shuttle"
		conversation
			`Testing the new GameAction and apply node mechanics!`
			`	The following payment replacement should say 100 credits: <payment>`
			`	That is the payment for this mission's on complete.`
			apply
				payment 1000
			`	The following payment replacement should say 1,000 credits: <payment>`
			apply
				payment 10000
			`	The following payment replacement should say 10,000 credits: <payment>`
			`	You should now have 11,000 more credits!`
			`	You should also have been given a new Shuttle by the on offer action.`
			apply
				set "branch test"
			`	Depart and land to complete this mission.`
				accept
	
	on complete
		payment 100
		conversation
			`Congrats on the 100 credits: <payment>`
			`	Now for more conversation actions.`
			apply
				fine 11100
				outfit "Jump Drive"
				give ship "Shuttle"
			`	Have a <fine> fine, a jump drive, and another Shuttle from a conversation action.`
			choice
				`	"I want a log entry."`
					goto log
				`	"I want an event trigger."`
			
			apply
				event "you watched the ring tape" 7
			`	Here you go. You will die in 7 days.`
				goto conditions
			
			label log
			apply
				log "You chose to get a log entry."
			`	Here you go. Check your log book.`
			
			label conditions
			branch success
				has "branch test"
			`	You shouldn't be here.`
				goto end
			
			label success
			`	You successfully recieved a condition from the on offer conversation!`
			
			label end
			`	Thank you for testing the new conversation mechanics. Now pay your fine.`

event "you watched the ring tape"
```
image
image
image

PR #5158, opened 2020-07-11
Add plural gender-neutral titles to human names

SubstandardZeal:


Amazinite:
I'll just handle using the new syntax where possible for the entire file on my own, or someone else could tackle that feat.

PR #5160, opened 2020-07-14
feat: Use RGBA tetxures instead of BGRA for better portability

thomasballinger:
This is https://github.com/endless-sky/endless-sky/pull/5153 with whitespace changes removed.

## Summary
Use RGBA texures instead of BGRA for better portability, as only some variants of libjpeg support BGRA, use the more common variant.

## Testing Done
Visually checked if the textures look the same before and after my change


tehhowch:
Please open the PR to the upstream branch, rather than duplicate the PR.

Issue #5159, opened 2020-07-13
New NPC specification: toggle NPC activation based on conditions


Amazinite:
**Is your feature request related to a problem? Please describe.**
The heart of the issue of #1564 is that causing things to occur based off of conversation choices can be difficult right now, as often times you need to delay the effect to the next landing. (See #4786 or #5007 where missions are forced to delay payment because the player's conversation choices can not immediately affect payment.) While #5157 allows conversations to trigger events and yield payments, it does not allow for an `npc` block to be triggered based off of conversation choices (another request of #1564). Taking the route of adding an `npc` block into a conversation that triggers if reached seems like it would be a difficult undertaking and a messy solution to the problem of needing to delay certain events, but some sort of way to immediately activate an `npc` based off of conversation choices is still desired.

**Related Issue Links**
#1564

**Describe the solution you'd like**
Create a new `conditions` keyword for the `npc` block. This block is a ConditionSet (similar to `to offer`) that if evaluated to true will cause the associated ships to spawn for the mission. If the ConditionSet evaluates to false and the `npc` does not spawn its ships, then the objective associated with the `npc` (e.g. `kill`, `save`, etc.) is ignored.

```
npc kill
	conditions # only spawns if the following conditions are met
		has "made a bad choice"
		random < 50
	government "Pirate"
	ship "Falcon" "Bad Choice"
```

These conditions should be evaluated *after* the `on offer` has finished, meaning that a `conversation` can use the `apply` node to change the conditions that dictate whether the `npc` activates.

**Describe alternatives you've considered**
The only other alternative would be to do a similar thing as #5157 and allow `npc` blocks to be embedded within a conversation (#1564), but as mentioned, I imagine that would be messy and is not the best solution to the problem.

**Additional context**
N/A

Amazinite:
This could also be used to have missions spawn more ships based off of the player's combat rating or net worth, leading to missions that scale with the player (without requiring duplicate or extra missions to handle the scaling).

tehhowch:
Shouldn't be too difficult for someone to implement. I think as far as the keyword goes, `"to spawn"` is the logical choice.

[...]

PR #5161, opened 2020-07-14
Allow mission NPCs to spawn or despawn based off of ConditionSets

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #5159. Related to #1564.

## Feature Details
This PR adds the ability to spawn or despawn `npc`s based off of the player's conditions. 
The `to spawn` conditions are first evaluated after the player has accepted the mission, meaning that conditions changed during an `on offer` conversation can trigger whether `npc`s spawn. Should the `to spawn` conditions fail on accept, then they are reevaluated on every landing, departure, and jump until succeeding. If the conditions succeed, the `npc` will spawn on the next departure.
The `to despawn` conditions are evaluated at the same times as the `to spawn` conditions (on accept and after every landing, departure, or jump until succeeding), but only if the `to spawn` conditions have succeeded. (That is, if the NPC hasn't even spawned yet, it won't test its `to despawn` conditions.) If the conditions succeed, the `npc` will despawn on the next landing.

If an `npc` block either isn't spawned or has been despawned, then the objective of the `npc` is ignored. This means that you can create alternative objectives for a mission. E.g. you may have an `npc` that must be directly killed in order to complete the mission, or you may be able to do something that triggers the despawn conditions on the `npc`, therefore ignoring the kill objective and allowing you to complete the mission. (Or you can create additional objectives by having an `npc` with a specific objective spawn in the middle of the mission.)

The syntax for spawn and despawn conditions is as follows:
```
npc
	...
	to spawn
		<conditions>
	to despawn
		<conditions>
	...
```

## UI Screenshots
N/A

## Usage Examples
The following `npc` block would only spawn its ships if you have the "angered warlord" condition, perhaps gained by making a wrong choice during a conversation. Then should the player obtain the "appeased warlord" condition, these pirates will despawn on the next landing.
```
npc kill
	to spawn
		has "angered warlord"
	to despawn
		has "appeased warlord"
	government "Pirate"
	fleet "Large Southern Pirates" 2
```

## Testing Done
<details><summary>Test mission.</summary>

```
mission "NPC Conditions Test"
	repeat
	source "Shiver"
	destination "Earth"
	on offer
		conversation
			`Do you want to spawn an NPC?`
			choice
				`	Yes.`
					goto yes
				`	No.`
			`	Okay, there will be no NPC. If you enter Fomalhaut and land, the NPC should spawn.`
					accept
			
			label yes
			apply
				set "spawn NPCs"
			`	Okay, spawning the NPC. If you enter Vega and land, the NPC should despawn.`
				accept
	
	on enter "Vega"
		set "despawn NPCs"

	on enter "Fomalhaut"
		set "spawn NPCs"
	
	npc kill
		to spawn
			has "spawn NPCs"
		to despawn
			has "despawn NPCs"
		government "Drak"
		ship "Archon" "Shoulders of Giants"
	
	on fail
		clear "spawn NPCs"
		clear "despawn NPCs"
	on complete
		clear "spawn NPCs"
		clear "despawn NPCs"
		dialog "Did it work?"
```
</details>

## Performance Impact
N/A

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -354,7 +354,8 @@ void Mission::Save(DataWriter &out, const string &tag) const
 			out.Write("stopover", planet->Name(), "visited");
 		
 		for(const NPC &npc : npcs)
-			npc.Save(out);
+			if(npc.PassedSpawn() && !npc.PassedDespawn())
+				npc.Save(out);
 		
 		// Save all the actions, because this might be an "available mission" that
 		// has not been received yet but must still be included in the saved game.
@@ -811,6 +812,9 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 	{
 		++player.Conditions()[name + ": offered"];
 		++player.Conditions()[name + ": active"];
+		// Any potential on offer conversation has been finished, so update
+		// the active NPCs for the first time.
+		UpdateNPCs(player);
 	}
 	else if(trigger == DECLINE)
 	{
@@ -851,6 +855,18 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 const list<NPC> &Mission::NPCs() const
 {
 	return npcs;
+}
+
+
+
+// Update which NPCs are active based on their spawn and despawn conditions.
+void Mission::UpdateNPCs(const PlayerInfo &player)
+{
+	for(auto &npc : npcs)
+	{
+		npc.CanSpawn(player);
+		npc.CanDespawn(player);
+	}
 }
 
 
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -353,7 +353,9 @@ void Mission::Save(DataWriter &out, const string &tag) const
 		for(const Planet *planet : visitedStopovers)
 			out.Write("stopover", planet->Name(), "visited");
 		
-		// Save all NPCs, except those that have despawned.
+		// Save all NPCs, except those that have despawned. This is so that despawned
+		// NPCs will not reappear should the player quit the game and return, and the
+		// NPCs no lonager pass the despawn conditions.
 		for(const NPC &npc : npcs)
 			if(!npc.PassedDespawn())
 				npc.Save(out);
@@ -864,10 +866,7 @@ const list<NPC> &Mission::NPCs() const
 void Mission::UpdateNPCs(const PlayerInfo &player)
 {
 	for(auto &npc : npcs)
-	{
-		npc.CanSpawn(player);
-		npc.CanDespawn(player);
-	}
+		npc.UpdateSpawning(player);
 }
 
 
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -353,8 +353,9 @@ void Mission::Save(DataWriter &out, const string &tag) const
 		for(const Planet *planet : visitedStopovers)
 			out.Write("stopover", planet->Name(), "visited");
 		
+		// Save all NPCs, except those that have despawned.
 		for(const NPC &npc : npcs)
-			if(npc.PassedSpawn() && !npc.PassedDespawn())
+			if(!npc.PassedDespawn())
 				npc.Save(out);
 		
 		// Save all the actions, because this might be an "available mission" that
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -931,6 +931,10 @@ void Mission::Do(const ShipEvent &event, PlayerInfo &player, UI *ui)
 		
 		// Perform an "on enter" action for this system, if possible.
 		Enter(system, player, ui);
+		
+		// Update any potential NPCs for this mission, as an "on enter" action may have
+		// changed the player's conditions.
+		UpdateNPCs(player);
 	}
 	
 	for(NPC &npc : npcs)
--- a/source/NPC.cpp
+++ b/source/NPC.cpp
@@ -276,14 +276,14 @@ void NPC::Save(DataWriter &out) const
 // Update spawning and despawning for this NPC.
 void NPC::UpdateSpawning(const PlayerInfo &player)
 {
-	// The spawn and despawn conditions are tested every time this function is called
-	// until they pass. This is so that a change in a player's conditions don't cause
-	// an NPC to "un-spawn" or "un-despawn."
+	// The conditions are tested every time this function is called until
+	// they pass. This is so that a change in a player's conditions don't
+	// cause an NPC to "un-spawn" or "un-despawn." Despawn conditions are
+	// only checked after the spawn conditions have passed so that an NPC
+	// doesn't "despawn" before spawning in the first place.
 	if(!passedSpawnConditions)
 		passedSpawnConditions = toSpawn.Test(player.Conditions());
-	
-	// Only test despawn conditions if the spawn conditions have passed.
-	if(passedSpawnConditions && !toDespawn.IsEmpty() && !passedDespawnConditions)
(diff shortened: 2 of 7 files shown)


Amazinite:
Should the despawn conditions of `npc`s be evaluated more often, such as on landing and on each jump? (The ships would still only despawn after you've landed, but evaluating the despawn conditions while in flight means that the objective of an `npc` can be ignored while in flight.)

source/NPC.cpp
```diff
@@ -203,6 +212,29 @@ void NPC::Save(DataWriter &out) const
...
+		// were spawned do not then become despawned because they no longer
+		// pass the spawn conditions.
+		if(!toSpawn.IsEmpty() && !checkedSpawnConditions)
+		{
+			out.Write("to", "spawn");
+			out.BeginChild();
+			{
+				toSpawn.Save(out);
+			}
+			out.EndChild();
```
Amazinite, line 226: This syntax is copy-pasted from MissionAction.cpp, but I've noticed that not all files use this additional indentation + curly braces for `out.BeginChild(); ... out.EndChild();`. Might be worth standardizing.

source/NPC.h
```diff
@@ -86,6 +93,13 @@ class NPC {
 	Conversation conversation;
 	const Conversation *stockConversation = nullptr;
 	
+	// Conditions that must be met in order for this NPC to be placed or despawned:
+	bool checkedSpawnConditions = false;
```
petervdmeer, line 97: Why not initialize both passed conditions with false and skip the checked variable? Then you can set passedSpawnConditions to true if the conditions-set is empty or if the test passes.

source/NPC.cpp
```diff
@@ -241,6 +273,45 @@ void NPC::Save(DataWriter &out) const
 
 
 
+// Update to spawn conditions for if this NPC can be placed.
+void NPC::CanSpawn(const PlayerInfo &player)
+{
+	// Only check the toSpawn conditions the first time this function is called.
```
petervdmeer, line 279: Does this mean that an NPC only has 1 single chance to spawn?
I would expect the CanSpawn to be checked multiple times during the mission (as long as the NPC didn't spawn yet).

source/NPC.cpp
```diff
@@ -241,6 +273,45 @@ void NPC::Save(DataWriter &out) const
...
+	if(!toSpawn.IsEmpty() && !checkedSpawnConditions)
```
petervdmeer, line 280: if toSpawn is empty, then I also expect passedSpawnConditions to go to true.

source/NPC.cpp
petervdmeer: Yes, sounds appropriate for the style guide.

source/Mission.cpp
```diff
@@ -354,7 +354,8 @@ void Mission::Save(DataWriter &out, const string &tag) const
 			out.Write("stopover", planet->Name(), "visited");
 		
 		for(const NPC &npc : npcs)
-			npc.Save(out);
+			if(npc.PassedSpawn() && !npc.PassedDespawn())
```
petervdmeer, line 357: This again suggests to me that the check on passing spawn only gets done once. Is that really the intention?

Amazinite: Yes. `to spawn` is only a matter of the NPCs being created once the mission is accepted. It is not a matter of "the mission is accepted, so now if at any point when the mission is active these conditions are met, the NPCs will spawn."

source/Mission.cpp
```diff
@@ -855,6 +860,18 @@ const list<NPC> &Mission::NPCs() const
 
 
 
+// Update which NPCs are active based on their spawn and despawn conditions.
+void Mission::UpdateNPCs(const PlayerInfo &player)
+{
+	for(auto &npc : npcs)
+	{
+		npc.CanSpawn(player);
+		npc.CanDespawn(player);
```
Amazinite, line 869: Should despawn conditions only be checked if the spawn conditions have succeeded? Right now an npc can fail its spawn conditions but succeed its despawn conditions. Then if the player's conditions later change such that the spawn conditions pass but the despawn conditions would no longer pass, the fleet can't spawn, as the despawn conditions has already succeeded before.

source/NPC.cpp
```diff
@@ -417,6 +495,11 @@ NPC NPC::Instantiate(map<string, string> &subs, const System *origin, const Syst
 	result.mustEvade = mustEvade;
 	result.mustAccompany = mustAccompany;
 	
+	result.passedSpawnConditions = passedSpawnConditions;
+	result.passedDespawnConditions = passedDespawnConditions;
```
petervdmeer, line 499: I don't have a lot of experience with this part of the ES-codebase, but I would expect those two variables to only be relevant for Instantiated NPCs, not for the template NPCs. Is it useful to copy the (default) values from the template when those variables also get initialized with the default values?

source/NPC.cpp
```diff
@@ -241,6 +273,42 @@ void NPC::Save(DataWriter &out) const
...
+{
+	// Check the toSpawn conditions each time this function is called until they pass.
+	if(!passedSpawnConditions)
+		passedSpawnConditions = toSpawn.Test(player.Conditions());
+}
+
+
+
+// Return if spawned conditions have passed, without updating.
+bool NPC::PassedSpawn() const
```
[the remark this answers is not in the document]
Amazinite: Not when I'm not saving out despawned NPCs.

source/NPC.cpp
Amazinite: These lines seemed necessary for this to work as I initially didn't have them and NPCs would always spawn and never despawn, ignoring their conditions.

source/NPC.cpp
```diff
@@ -241,6 +273,42 @@ void NPC::Save(DataWriter &out) const
 
 
 
+// Update to spawn conditions for if this NPC can be placed.
+void NPC::CanSpawn(const PlayerInfo &player)
+{
+	// Check the toSpawn conditions each time this function is called until they pass.
+	if(!passedSpawnConditions)
+		passedSpawnConditions = toSpawn.Test(player.Conditions());
```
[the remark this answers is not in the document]
Amazinite: > Do empty conditions always test-evaluate to true? If not, then you would want to add toSpawn.IsEmpty() as a reason to set the variable to true.

~~Grumble grumble I had that and you made a comment about it.~~ Yes. Empty conditions are always true. In Mission::CanOffer, toOffer is always tested, even if empty, whereas toFail must not be empty (as otherwise an empty toFail would evaluate to true). Same with how I'm always testing toSpawn but only testing toDespawn if it's not empty.

https://github.com/endless-sky/endless-sky/blob/3d4a0b233dbd6a73b46285be64bd3052be49efed/source/Mission.cpp#L576-L580

I also skimmed through ConditionSet::Test, which seemed to back this up.

source/NPC.cpp
```diff
@@ -241,6 +273,42 @@ void NPC::Save(DataWriter &out) const
 
 
 
+// Update to spawn conditions for if this NPC can be placed.
+void NPC::CanSpawn(const PlayerInfo &player)
```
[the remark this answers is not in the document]
Amazinite: Perhaps.

source/NPC.cpp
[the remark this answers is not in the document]
petervdmeer: Okay, then this code here is fine.

source/NPC.cpp
Amazinite: Through testing of removing these, they are necessary.

petervdmeer: The variables are also initialized as false in the template, so I don't expect any issues when just all variables including those are copied from the template.

PR #5164, opened 2020-07-18
System attributes for the Ember Waste and select systems

Zitchas:
**Content (Map Data)**

## Summary
This PR adds attributes to a variety of systems in order to label a few key features, as per the specification included in the [MapData ](https://github.com/endless-sky/endless-sky/wiki/MapData) wiki page. 

## Details
The attributes that it adds are:
- "ember waste" (applied to all systems in the Ember Waste)
- "graveyard" (applied to all systems in the Graveyard)
- "void sprites" (applied to Nenia)
- "archon" (applied to all systems that have resident Archons.)
- "wormhole" (applied to all systems that have wormholes)

## Reason
I am planning a variety of generic jobs for the Remnant because, well, they don't really have anything other than the Korath Bounties. In order to target these jobs properly, they need to be able to have identifiers that allow the game to randomly pick from sets. For instance, picking a random system that has a wormhole. Or a random system in the Ember Waste, etc. 

## Importance
I would just like to note that a number of small upcoming mission sets that give the player some alternative repeating opportunities (that is to say, jobs) depend directly on this PR. And by "small mission sets," I mean "1 mission to introduce the idea, and then 1-3 different repeatable jobs based around the idea" sorts of things. As such, I would like to see it merged for 0.9.13, and consider it a higher priority for merging as it simply provides foundational data for other missions and jobs to be built on.

## Future development possibilities (not intended for this PR)
- Adding an attribute tag that specifies its region to all systems.
- Adding a region item to the system properties. (this would be tidier than using an attribute, but would require an engine change)

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -2062,7 +2062,7 @@ system Adhara
 system Aescolanus
 	pos -15 340
 	government Uninhabited
-	attributes "The Ember Waste"
+	attributes "The Ember Waste" "Wormhole"
 	habitable 233.28
 	belt 1323
 	haze _menu/haze-red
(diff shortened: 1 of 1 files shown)


Zitchas:
OK. Changed!

PR #5170, opened 2020-07-24
Mouthparts sprite

beccabunny:

## Summary
This PR adds a sprite for the "Mouthparts?" outfit used by the Void Sprites (which is, currently, unobtainable) for flavour. It can be seen only through all contents plugins, and has the Secondaries angle instead of the normal one since otherwise nothing on it could be recognisable.

Screenshot (1939)
*The outfit in game.*

## PR Checklist
 - [x] I uploaded the necessary image, blend, and texture assets here: [mouthparts.zip](https://github.com/endless-sky/endless-sky/files/4971810/mouthparts.zip)
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/131


Arachi-Lover:
can also be used for scenes

Zitchas:
Yes, it could be used for a scene. Given that it is apparently problematic to even get a thumbnail for the void sprites, I'm not sure how likely/reasonable it would be to get a scene of a void sprite charging at you looking like it wants to either grab or eat you...

Issue #2442, opened 2017-04-22
Shipyard/Outfitter GPU usage


lheckemann:
I have some 150-odd ships. This is fine most of the time (e.g. in space, including in big battles), but causes a huge amount of GPU usage and heat when I enter the shipyard or outfitter panels with the ships present, to the extent that it crashed my whole system the first time I encountered it.

Bladewood:
what the hell kind of computer are you using

lheckemann:
Intel integrated graphics, on a "Intel(R) Core(TM) i7-2720QM CPU @ 2.20GHz"

The crash is presumably a driver issue, but the performance problems shouldn't be.

jafdy:
Try updating the drivers. I know it seems obvious but it's a good idea to start with the obvious.

EndrosG:
I would think it's the driver or so...
I have some 400 ships parked on a planet and I have no problems in neither shipyard nor outitter.
Memory consumption of the game is low too (no changes at all).
And my system is rather "weak". A low budget Asrock Q1900M mainboard ("Intel(R) Celeron(R) CPU  J1900  @ 1.99GHz").
On my netbook I dont have any issues with the shipyard or outiftter either.
But I can't really bother to fly with that many ships since all those engine sprites require a lot of computation on

lheckemann:
>Try updating the drivers

I'm on linux and already running the latest kernel available in my distro.

lheckemann:
Tested on linux 4.9.22 and linux 4.10.10, both with ES 0.9.4 and latest master, could reproduce issue on all 4 combinations. Interestingly, it only occurs when the panel with the ship outlines is visible: if I make the window smaller vertically and scroll to the bottom so only the selected ship's stats are visible, GPU usage drops to normal levels.

gunqqer:
400 ships? Wtf guys do you never sell any of your old ships?

EndrosG:
> cat /proc/version 
Linux version 3.16.0-4-amd64 (debian-kernel@lists.debian.org) (gcc version 4.8.4 (Debian 4.8.4-1) ) #1 SMP Debian 3.16.36-1+deb8u1 (2016-09-03)

No issues here. Using Steam version 0.9.6 as well as latest master.
[...]

PR #5179, opened 2020-07-27
Allow the "fancy" sobel filters in ShopPanels to be replaced with "fast" sprite images

Amazinite:
**Feature:** This PR implements a feature to fix the issue #2442.

## Feature Details
Adds a new preference that toggles the behavior of the ship outlines in the ShopPanel. When set to "fancy," the game renders the sobel filters of the ship sprites, as it does now. When set to "fast," the game instead uses the top-down sprite of the ship instead of applying the shader to it.

NOTE: This preference ONLY affects the ShopPanel. The sobel filter shader is still used for the flight UI, the main menu, and for the ShipInfoPanel. In each of these cases you only have one or two ships that need the shader rendered, and therefore I saw the need to change these cases unnecessary, as the issue was with a large number of ships rendering this shader at once, which is only found in the ShopPanel.

## UI Screenshots
The location of this setting in preferences.
image

A side-by-side of the two settings, fancy on the left and fast on the right.
unknown

## Usage Examples
N/A

## Testing Done
Tested that the preferences itself properly toggles the outlines in the ShopPanel.
Tested that the preference saves when the game is closed.
Tested GPU usage with 50 Shuttles on fast vs fancy.
* On the main landing screen (our control): ~19% GPU usage
* In the shipyard (fancy): ~50% GPU usage
* In the shipyard (fast): ~19% GPU usage (basically the same as the control)

## Performance Impact
Given that this is allowing for a less GPU intensive option, switching to the "fast" outlines option should improve performance in ShopPanels for those with large fleets.

### DIFF
--- a/source/Preferences.cpp
+++ b/source/Preferences.cpp
@@ -57,6 +57,7 @@ void Preferences::Load()
 	settings["Draw starfield"] = true;
 	settings["Hide unexplored map regions"] = true;
 	settings["Turrets focus fire"] = true;
+	settings["Ship outlines in shops"] = true;
 	
 	DataFile prefs(Files::Config() + "preferences.txt");
 	for(const DataNode &node : prefs)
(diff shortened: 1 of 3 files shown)


Amazinite:
>One small note: In your testing you didn't specify if you test if the setting stays on when a game is closed and re-started and if the setting stays off over a close/re-start.
I expect this to work, but I would like to see this explicitly mentioned as tested. (It is two short simple tests.)

Good catch. I did test this, but hadn't mentioned it in the PR.

PR #5180, opened 2020-07-27
Replace modulo reduction with faster, better integer multiplication

SubstandardZeal:
## Feature Details
Replaces the modulo reduction of `Random::Int` with the integer multiplication reduction proposed here:
[Fast Random Integer Generation in an Interval](https://arxiv.org/abs/1805.10941)

It uses only a single multiplication rather than a division and is much faster. The inherent bias is also more evenly distributed (It is not *less* biased but it is perhaps *better* biased). The modulo reduction causes lower numbers to be more common while the integer multiplication reduction is more even (I have visualizations of this if you want).

## Testing Done
Tested in both Debug and Release builds on my machine.

## Performance Impact
The random bounded integer generation is likely never a bottleneck but this is still a free performance gain. In my testing it had no discernible effect on CPU/GPU load (even with ~1000 Pug Seekers). On some very old 32-bit machines without decent multiplication instructions it might be slow but in that case modulo might be even slower.


SubstandardZeal:
We could also consider replacing the dated Mersenne Twister PRNG with something faster & stronger. That would probably be determined by what architecture we want to target (32/64 bit, fast multiplication?).

petervdmeer:
Nice to see different algorithms being proposed for ES.

The modulo reduction indeed seems to have a small bias for lower numbers, but maxInt32 is 2147483647 and the ES missions typically use random up to value 10000 or less. Effect from the bias would then be in the range of 1/214748 ~= 0.000005.

I don't know which platforms we officially target, but I recently saw a discussion from somebody running ES on a Raspberry Pi 4 (Arm processor, OpenGL-embedded) and I also saw efforts to get parts of ES to run in a Javascript based C++ emulator framework https://observablehq.com/@ballingt-personal/wip-formations-from-formation-pr.

I'm no experts on RNG algorithms, so it is difficult for me to review which algorithms would work best. Faster/stronger PRNGs sound nice, but I would really like an efficient PRNG that works on as many platforms as possible. It is okay for me if the performance is somewhat reduced on uncommonly used platforms, but I would like to prevent issues as seen in #4457 (ES running fine on PPC, but some parts not working due to a small detail).

MCOfficer:
Free performance without significant downsides? sign me up.

> The modulo reduction causes lower numbers to be more common while the integer multiplication reduction is more even (I have visualizations of this if you want).

I'd be interested in how much the two deviate.

SubstandardZeal:
Here's a demonstration of the bias patterns for `upper_bound = 100`
image

And an interactive graph
https://www.desmos.com/calculator/ftkta2bkak

And a histogram for `upper_bound = 30`
image

SubstandardZeal:
Regarding PRNGs, there are many modern algorithms designed for very minimal hardware. [Xoshiro/Xoroshiro](http://prng.di.unimi.it/) and [Small Fast Counting RNG](http://pracrand.sourceforge.net/RNG_engines.txt) use only addition and bitwise operations. 

Melissa E. O’Neill's [Permuted Congruential Generators](https://www.pcg-random.org/) are nice but we may want to avoid multiplication.

SubstandardZeal:
A recent paper on the issues with Mersenne Twister https://arxiv.org/abs/1910.06437

peteroupc:
Unbiased integer generation is possible, even without multiplication.  The following JavaScript code illustrates this.

function randomInt(minInclusive, maxExclusive) {
  var maxInclusive = (maxExclusive - minInclusive) - 1
  var x = 1
  var y = 0
  while(true) {
    x = x << 1
    var randomBit = nextBit()
    y = y << 1 + randomBit
    if(x > maxInclusive) {
      if (y <= maxInclusive) { return y + minInclusive }
      // Rejection
      x = x - maxInclusive - 1
      y = y - maxInclusive - 1
    }
  }
}

This is the Fast Dice Roller by @jlumbroso (https://arxiv.org/abs/1304.1916).

Discussions of PRNGs, though, are perhaps out of scope for this pull request.

peteroupc:
@SubstandardZeal: On an unrelated GitHub issue, @theironborn made charts that are highly related to this topic: https://github.com/rust-lang/rust/issues/74928

source/Random.cpp
```diff
@@ -58,12 +58,13 @@ uint32_t Random::Int()
...
 
-uint32_t Random::Int(uint32_t modulus)
+uint32_t Random::Int(uint32_t upper_bound)
 {
 #ifndef __linux__
 	lock_guard<mutex> lock(workaroundMutex);
 #endif
-	return uniform(gen) % modulus;
+	const uint32_t x = uniform(gen);
+	return ((uint64_t) x * (uint64_t) upper_bound) >> 32;
```
tehhowch, line 67: C style casts are forbidden by the style guide.

tehhowch (changes requested):

PR #5185, opened 2020-07-29
Move the Archaeology missions into the emerald sword.txt, rename emerald sword.txt to archaeology missions

Anarchist2:
Due to the new Emerald Sword missions, combined with the length and general interplay of the two mission strings, I see it fit that those missions be move to outside `transport missions.txt` and into their own file, `archaeology.txt`. This both declutters transport missions and makes it easier to figure out how to get to the Emerald Sword missions, making them better defined.


data/human/archaeology missions.txt
```diff
@@ -0,0 +1,858 @@
...
+			`	Albert continues, "What I think is that an unknown alien species has been involved in events in the Deep since the beginning, and is perhaps still involved there today. I think this is known to at least some of the Deep's inhabitants."`
+			choice
+				`	"So, what do we do now?"`
+				`	"How can we find out more information?"`
+			`	"Well," he says, "unfortunately, Captain, they're going to be watching you. I'll have to seek out someone else for the next stages of my research. But I will keep you notified if I discover anything further."`
+			`	It is an unsatisfactory conclusion to your work, with so many questions unanswered. But to make up for that, Albert gives you <payment> as payment for your services. "You will hear from me again," he promises.`
+
+
+
+mission "There Might Be Riots 1"
```
Amazinite, line 235: While TMBR is required to continue through the archaeology missions, it isn't itself an archaeology missions, so I don't think it makes much sense to put it here.

Zitchas:
I agree, it makes sense to have all the archaeology missions in the same file.

Amazinite:
Should we also rename the file at this point to just “archaeology missions.txt”?

Zitchas:
Maybe just `archeology.txt` ? I don't think it really needs the "missions" on the end, but I suppose it doesn't hurt, either.

Amazinite:
> That being said, the main reason to keep them where they are (and move the rest of the Emerald Sword stuff into the transport missions file) would be the connection with T.M.B.R.

The Wanderer campaign requires that the main plot be finished, but we don't have all that in the same folder or file. I don't see an issue with having TMBR in a different file than the archaeology missions just because they require TMBR to be done at one point.

Zitchas:
Sure. I didn't mean that it was a big issue, just that it was the only reason I could see for keeping it all in "transport missions."

Anyway, I'm good with moving them over, although I'd prefer `archeology.txt` rather than `emerald sword.txt`

Issue #4165, opened 2019-02-01
Should the news shown in spaceports only refresh on landing?


Amazinite:
Currently, clicking the spaceport button repeatedly will show you various news messages. Given the limited number of portraits and messages though, it's not uncommon to have the same image pop up with the same message and title, or even more oddly to have the same image pop up for supposedly two totally different people with different titles giving you different messages.

Suggested changing this so that the news that you get is generated on landing and only that one message will ever show up whenever you click on the spaceport until you depart. This removes the potential scenario of oddly having the same message twice or the same image twice in a row, let alone being able to flip through all the possible messages on one landing by just sitting there and reading. This would not only remove those scenarios, but also incentivize traveling and hitting the spaceport button every time you land to see what message you get on that landing.

tehhowch:
a9feb0e0 moves the news selection to when the planet panel is created (i.e. the player has just landed or reloaded their save).

We could also put a timer on it, e.g. it only changes every 5 minutes, by using a `step` counter much like the MissionPanel uses for blinking the deadline pointers.

MCOfficer:
Do you mean 5 minutes during which the player is on the planet, or 5 minutes regardless of how often they take off and land?
I don't really see a use-case for the former, and... actually i don't see a use-case for either of them.

Wrzlprnft:
Spontanous, not that well thought out idea: how about an arbitrary yet to be found number of clicks that's not the spaceport button? Premise: Players who spend thought time on a planet spend inlore time, going through the outfitter, going through the commodity prices on the map and such could encounter new news due to the character spending time and walking around.

Amazinite:
I'd rather not have the news update at any point for any reason after landing. If the news only updates on landing, and spaceport missions are only offered the first time you click the spaceport button, then having some way to have news update after say a certain amount of time or a certain number of actions is taken may improperly imply that the chance to get offered a spaceport mission has also refreshed.,

endless-sky:
Arguably, if you think it makes sense that every time the player "returns" to the spaceport they run into the same person saying the same thing, the current behavior (they run into multiple people, but occasionally the same one twice) isn't any less logical.

And, being able to cycle through messages if you feel like it just feels more fun than always getting the same one, especially if there are funny ones. It's the equivalent of how in Warcraft II, every time you clicked on a unit they'd say something slightly different, and get more annoyed the more you clicked - it was fun to pester them just to hear all those messages.

Ferociousfeind:
[...]

PR #5187, opened 2020-07-29
Remove portrait images from spaceport news

### LINKED ISSUE #4463: spaceport news mixing up roles and messages

Amazinite:

## Summary
This PR addresses #4325 and removes all portraits from news. This PR does *not* remove the functionality of portraits, only removing the current portrait images from being used. This PR also does *not* remove the portrait images from the game, as plugins may be using them. (If it's decided though that we should also remove the images from the game, we can easily do so.)

This PR also updates the interface so that the name and message move locations depending on if the news does have a portrait or not. Given that the heart of #4165 was the repetition of portrait images, and this PR removes said images from the base game, I've reverted dea4cf7 as well. This also fixes #4463 by removing the portraits and the commit that caused the issue when portraits are present (so if a plugin decides to use portraits, #4463 won't be an issue, and #4165 should really be solved by smart use of which news gets which portraits, and by having a greater variety of portraits).

## Screenshots

Simply removing `portrait` from every news entry resulted in the following image, hence the code changes:
image

Now the spaceport news will look like the following with no portrait:
image

Or like this with a portrait:
image
Portraits now appear on the right side of the UI so that the name and start of the message are always in the same spot regardless of the presence of a portrait.

## Save File
N/A

## PR Checklist
N/A
 - [x] ~~I updated the copyright attributions, or decline to claim copyright of any assets produced or modified~~
 - [x] ~~I uploaded the necessary image, blend, and texture assets here: {{insert link to assets}}~~
 - [x] ~~I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: {{insert PR link}}~~

### DIFF
--- a/data/hai/hai news.txt
+++ b/data/hai/hai news.txt
@@ -9,7 +9,6 @@
 # WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
 # PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
-# Visual suggestion: a bit quirky or weird; expression is content, confident, or stoic
 news "food in Hai space"
 	location
 		government "Hai"
@@ -19,11 +18,6 @@ news "food in Hai space"
 			"Food connoisseur"
 			"Tourist"
 			"Tour guide"
-	portrait
-		portrait/human26
-		portrait/human31
-		portrait/human57
-		portrait/human68
 	message
 		word
 			"This may sound odd, but the acorn cookies they sell here are the best cookies I've ever eaten."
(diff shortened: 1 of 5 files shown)


Amazinite:
Something mentioned in https://github.com/endless-sky/endless-sky/issues/4325#issuecomment-592526024 that I did not do here is move the portrait to the right side. Doing this would mean that the name and start of the message are always in the same spot regardless of if there is a portrait. Should this be done?

MCOfficer:
> Something mentioned in [#4325 (comment)](https://github.com/endless-sky/endless-sky/issues/4325#issuecomment-592526024) that I did not do here (but can easily do with a simple interface change) is move the portrait to the right side. Doing this would mean that the name and start of the message are always in the same spot regardless of if there is a portrait. Should this be done?
> image

If we make portraits optional, moving them to an optional position makes sense.

Zitchas:
This change makes sense to me, on multiple grounds:
1) As stated, it prepares things for the potential complete removal of portraits in the news, while not removing the *capability* of having them.
2) Given the general left-to-right reading pattern for English and similar languages, this change means that instead of generally "reading" things in the picture-title-text order, we get it in something closer to a title-text-picture order. Of course, the picture still stands out, being a picture, but I feel like this subtly de-emphasizes the pictures, which might help reduce their focus and impact, which in turn helps shift them from being an obvious and potential obstacle into simply an aid.

Issue #5167, opened 2020-07-21
Request: Have Jump Drive range be an Attribute for Systems.


Zitchas:
**Problem:**
Jump Drives currently have a fixed and fairly large range, which means that parts of the galaxy (such as the Ember Waste) that are set up to be confusing and kind of like a maze must have reduced density because otherwise ships with JD can just skip right across it without worry. This also means that some areas (like the galactic core) are almost forcibly very sparse in systems despite being very dense in stars and thus potential systems.

**Related Issue Links**
https://github.com/endless-sky/endless-sky/issues/788

**Describe the solution you'd like**
I would like the addition of a "Jump Range" attribute that could be placed on systems. This would be an optional value, and if not set it would default to the current jump distance. If set, it would override the Jump Drive range when jumping out from that system.

**Impact**
- This would enable select systems and regions to be denser than the rest of the galaxy without overly promoting JD usage. 
- This could also be used to create some interesting variations on mazes, such as having systems that could be jumped into, but not out of, resulting in systems that the player needs to work their way out of the maze even if they jump in elsewhere. 
- Conversely, it could create "jumping off systems" that have significantly increased range where they could cover large distances but be hard to actually reach.

**Describe alternatives you've considered**
- Having JD distance be tied to a region of space. (which would be nice, but requires defining a new characteristic of thing in the game and more variable confusion.
- Spore had a variable system where the closer one was to the galactic core, the shorter the jump range was. This would solve some of the problems, but doesn't do anything for creating interesting sub-regions. It also introduces a new set of calculations and variables to the map as a whole.

Janaszar:
So, as you may know, I've often toyed with the idea of changing jump-drive distances...

For this, I'd like to know three things regarding the mechanics:
1) Would the `Jump Range` attribute be a multiplier, an addition/subtraction to the value, or perhaps set it to a static value? Or allow for both (Multiplier + Addition/Subtraction or a Static value; if you're going to have a multiplier or addition, I'm going to advocate to include the other just for possible uses down the line)?

2) How would you handle things such as having an "Extended Jump Drive", which gives the player say, 200 jump distance, instead of the current 100?

3) How would you handle the map pathfinding algorithm?  Right now, it's a basic search checking only the distance between neighbors, and the hard-coded jump drive distance, to find a path.  Would you be adjusting the reported distance to the "neighbors", or would you leave it assuming standard distances?  If you adjust it, would it be by changing the reported distances, or would you be changing the jump distance just for that node?

[...]

Issue #788, opened 2016-02-15
A jump drive with changeable range : Feature Request


Wrzlprnft:
Even if i do not support the idea of a galaxy-wide-jump-drive-range (http://steamcommunity.com/app/404410/discussions/2/412446890563414360/?tscn=1455448887) i think a adjustable jump drive range would be a part of the aliens technology ladder, and would also open up space for modders.

I looked in the source code but figured out nothing, so if i've overseen something just tell me :)

Oherwise maybe intialise the jump drive as normal, but maybe add a range attribute, and make another if-clause in the method which calculates the jump.

I think the fuel cost should be directly connected to the range though.

LocalGod79:
fuel:range ratio for a single jump should perhaps grow more rapidly than 'directly' - exponentially (to the power of 1.5, perhaps)?

this would ensure that galaxy wide jumps would be an extreme rarity (if even possible)...

DingusShingleton:
Enter : Bactrian with 100 fuel cells

jmhorjus:
I think the source code value you're looking for is the hard-coded constant "static const double NEIGHBOR_DISTANCE = 100.;" in System.cpp.  It determines which systems are considered "adjacent" for purposes of jump drives, but is currently universal - i.e. all jump drives and map code use the same constant distance and the "neighbors" of a system are calculated only once when the map is loaded. 

If you're able to build the source code, you could start by modifying NEIGHBOR_DISTANCE and observing the effects.  To make that distance variable during run-time you'd probably want to replace the system's "std::vector<const System *> neighbors;" (a vector of neighboring systems) with a vector of vectors or a map of vectors indexed by the neighbor_distance value used to calculate them.  This would work well if you had only a few different ranges of jump drive, which seem likely how you would want to have it.  
The map could possibly be populated based on the different ranges of drives found in the outfits data so it would still only happen once when the map was loaded. 

Then you'd have to find all the places where the neighbors list is used and update them.  Right now "system.neighbors" is already only used if you have a jump drive (otherwise "system.links" is used), so you should be able to pass in a parameter indicating your range so it knows what list of neighbors to use.  

Happy hacking!

Wrzlprnft:
@jmhorjus thx man! i just was to fixed on the idea of a range-attribute somewhere

[...]

PR #5193, opened 2020-07-30
Added a "jump range" attribute for changing how far a jump drive can travel

Amazinite:
**Feature:** This PR implements the feature requests detailed and discussed in issue #788 and #5167.

## Feature Details
This PR implements the following new attribute:
* `"jump range"`: When put on an outfit with the `"jump drive"` attribute, or when put on a system, alters the distance between systems that the player is capable of jumping. If not present then the default range of 100 is used. 

At game start, the game keeps track of how many different jump ranges have been seen on outfits. With this PR, the only jump range seen would be 100, but say you have two more jump drives with ranges 150 and 200, then the three known jump ranges would be 100, 150, and 200. Each system in the game then creates a set of neighbors for each of these ranges. When a ship has a jump drive of a certain range, the set of neighbors of the current system that the ship can jump to is then the set of neighbors that correspond to the ship's jump range.

If a system has its own jump range, then all potential jump ranges are ignored, and it only calculates the neighbors of its own jump range. Then, no matter the range of a ship's jump drive, the same set of neighbors is returned.

If the player's flagship has a non-standard jump range, then all unexplored systems within the player's current range are considered seen. This means that you can potentially have a system that is just out of range of the default jump drive range, but the player can be given a jump drive with a longer range that then brings the system within the player's range.
Alternatively, you can hide systems by shrinking the jump range of the surrounding systems, and later increasing their ranges to reveal the hidden systems.

## UI Screenshots
<details><summary>If the player has a jump range different than the default, then the ring drawn on the map changes accordingly. All systems within the player's jump range (or within the default jump range if the player has no jump drive) are considered seen.</summary>

image
image
</details>

<details><summary>Having a longer jump range means that the number of jumps needed to reach the destination system is decreased.</summary>

image
</details>

<details><summary>But swapping back to a normal jump drive reverts your jump range back to that.</summary>

image
</details>

## TODO

- [x] As can be seen in the screenshots, the jumps of the travel plan don't update accordingly to what your escorts are capable of doing. It assumes that your escorts can make the same long jumps as you, even if they have weaker drives. Preferably, just as the travel plan can recognize when some escorts have hyperdrives and some have jump drives, it should recognize when some escorts have short jump drives and some have long jump drives.
- [x] Should a ship have multiple jump drives of different ranges and fuel values, the lowest fuel and highest range is always chosen, even if the higher range one is what is actually being used. Preferably the range that a ship is jumping would exclude JDs without that range from being used in the fuel calculation.
- [x] Also implement #5167.
- [x] Optimization/simplifying code.

## Usage Examples
The following outfit would allow a ship to jump to systems that are up to 150 units of distance on the map away.
```
outfit "Long Jump Drive"
	...
	"jump drive" 1
	"jump range" 150
	...
```
The following system would override the jump range of all ships contained within it, giving them all the same jump range of 50.
```
system Aescolanus
	pos -15 340
	government Uninhabited
	attributes "ember waste" "wormhole"
	habitable 233.28
	belt 1323
	"jump range" 50
	haze _menu/haze-red
	...
```

## Testing Done
Equipped drives with varying jump ranges and turned the entirety of the Ember Waste into a 50 jump range region. Pathfinding still works, choosing the shortest path to the destination system, and the jump range of all systems along the path are respected. The jump range updates accordingly with the drive you have equipped and the system you are in. Escorts and other ships still obey their own jump ranges instead of magically inheriting the flagship's jump distance. Those systems within jump range become seen. 

## Performance Impact
The game calculates all neighbors for all possible jump ranges at game start, so if there is an excessive number of jump ranges then the game might load slower. I imagine that any performance impact during normal gameplay from this PR will be negligible, though.


tux2603:
This added a bit of stuttering for me with unrealistically large jump ranges (eg, a developer drive that can reach the entire galaxy), but besides that all seems to work as expected

Amazinite:
When does the stuttering occur? Just generally? When jumping? Whatever the case, that's one of those "play stupid games, win stupid prizes" type of things; throw preposterously large numbers at it and expect it to chug.

tux2603:
Yup, exactly that. If you have a stupid game of a super powered jump drive, you win the stupid prize of stuttering when jumping. I'd say it's a non issue, since I don't see any actual game mechanic that would ever use it

Zitchas:
This sounds very good. A question about implementation, though.

1) how do multiple outfits with a jump range value interact? For instance, if I have a standard jump drive with a range of 100, would it be possible to have a "jump field projector" with a range of 25 that adds to the standard drive's range, bringing it up to 125?

For that matter, if I have two jump drives with a value of 100 each, will those stack and give a JD range of 200?

Amazinite:
Jump range, like jump fuel, will be one of the few attributes in the game that does not stack with multiple copies of the same outfit. When jumping, the jump drive with the best range and the best fuel for the range being jumped are chosen.

Zitchas:
Hmmm... That's problematic for some idea I'd had, although I could probably work around it. So what you are saying is basically that - in order to create a Jump Drive range booster type outfit, what it *actually* would be is a brand new Jump Drive with the appropriately increased range that happens to have a requirement for a normal Jump Drive to be present?

Likewise, another idea I wanted to toy with is having a ship "perk" where the thing that makes the ship special is that it has a slightly bigger Jump Radius. I didn't want to have a JD built-in, just that *if*  the player chooses to use a JD, then they'd get the benefit of the slightly increased range.

Amazinite:
Unfortunately having a built in range upgrade isn’t possible either without also having a built in jump drive. (This could be changed, but I’m not too interested in doing so, as I think that jump fuel and jump range should behave similarly, and we are restricted by the fact that we really don’t want jump range or fuel to stack with more drives.) 

Whatever the case, I wouldn’t see myself merging anything that changes jump ranges on outfits/ships any time soon, so you should probably play around with putting jump ranges on systems instead.

Zitchas:
Oh, definitely. :)

source/DistanceMap.cpp
```diff
@@ -309,7 +310,8 @@ bool DistanceMap::CheckLink(const System *from, const System *to, bool useJump)
 	// the two systems that you can jump between them, you can plot a course
 	// between them even if neither system is explored. Otherwise, you need to
 	// know if a link exists, so you must have explored at least one of them.
-	if(useJump && from->Position().Distance(to->Position()) <= System::NEIGHBOR_DISTANCE)
+	double distance = from->JumpRange() ? from->JumpRange() : jumpRange;
```
petervdmeer, line 313: The code here seems fine, but add some comment here to explain that if the from system has a JumpRange, then we use the source-systems jump-range, otherwise we use the ships jump-range.

source/GameData.cpp
```diff
@@ -180,8 +182,11 @@ bool GameData::BeginLoad(const char * const *argv)
 			LoadFile(path, debugMode);
 	}
 	
-	// Now that all the stars are loaded, update the neighbor lists.
-	UpdateNeighbors();
+	// Now that all the stars are loaded, update the neighbor lists and other
```
petervdmeer, line 185: The ships and outfits also need to be loaded at this point (since we get the neighbor distances from the jumpdrive outfits and jumpdrives built-in into ships).
I would expect they are.

source/GameData.cpp
```diff
@@ -549,21 +554,29 @@ void GameData::Change(const DataNode &node)
...
+		it.second.UpdateSystem(systems, neighborDistances);
 	}
 }
 
 
 
+void GameData::NeighborDistance(double neighborDistance)
+{
+	if(!neighborDistances.count(neighborDistance))
+		neighborDistances.insert(neighborDistance);
```
petervdmeer, line 575: This code is needed when using a vector, but when using the datatype set, then it is overkill since the check if the element is present is already done within the set itself.

source/MapPanel.cpp
```diff
@@ -865,14 +872,15 @@ void MapPanel::DrawTravelPlan()
 		if(!isHyper && !isJump && !isWormhole)
 			break;
 		
+		double systemDistance = previous->Position().Distance(next->Position());
```
petervdmeer, line 875: The name systemDistance is a little confusing. Maybe use jumpDistance?

source/GameData.cpp
Amazinite: They are (everything is loaded at this point), but I’ll update the comment.

source/System.h
```diff
@@ -181,6 +186,7 @@ class System {
 	std::vector<FleetProbability> fleets;
 	double habitable = 1000.;
 	double asteroidBelt = 1500.;
+	double jumpRange = 0.;
```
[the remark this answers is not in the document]
Amazinite: Such systems could just be given a distance of 1 or something similarly small; a range so tiny that no other systems can be jumped to (unless for some reason a plugin is overlapping systems, which is bad practice).

source/System.cpp
```diff
@@ -362,14 +354,43 @@ void System::UpdateNeighbors(const Set<System> &systems)
...
+
 // Modify a system's links.
 void System::Link(System *other)
 {
 	links.insert(other);
 	other->links.insert(this);
 	
-	neighbors.insert(other);
-	other->neighbors.insert(this);
+	for(auto &neighborSet : neighbors)
```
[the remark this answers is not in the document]
Amazinite: >When there is only one set of neighbors, then it looks okay to add the hyperspace links and the jump links into one set, but when we have neighbor-sets per distance, then it looks a bit unnatural to add (and remove) hyperspace links to all the possible neighbor sets.

Given that jump drives use neighbor sets (they don’t actually ever use the links set), it’s important to update every set of neighbors, as the set of neighbors used depends on the jump drive.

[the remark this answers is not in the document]
petervdmeer: The jump drives can use the link sets without the hyperspace links being visible in them.

The code that does care is the code that determines if other systems are visible from some system. The loops there currently only use the neighbors and will need to be changed to use neighbors and links if the links are not contained in neighbors.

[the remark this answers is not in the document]
Amazinite: I think that simply putting all linked systems within the neighbors as has been the case is simpler than updating every other instance in the code where jump drives are used to check both neighbors and links.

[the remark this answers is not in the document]
petervdmeer: Okay, the code by itself is not wrong and the neighbors code indeed already worked this way before this PR, so I can understand that you just want to keep it as-is for this PR.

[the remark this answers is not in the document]
Amazinite: While we're here though, I think that updating neighbors in Link and Unlink is actually 100% unnecessary.
1. PlayerInfo::AddChange is where events are fed.
2. That function passes the changes to GameData::Change.
3. If an event changes links, then GameData calls these Link and Unlock function.
4. Returning to PlayerInfo::AddChange, if an event altered a system or changes links, then it calls GameData::UpdateSystems, which calls System::UpdateSystem.
5. UpdateSystem clears out all neighbors and then recalculates them.
Therefore, Link and Unlink don't need to update neighbors, since it's eventually done anyways.

source/System.cpp
```diff
@@ -757,6 +714,34 @@ void System::LoadObject(const DataNode &node, Set<Planet> &planets, int parent)
...
+// or links, figure out which stars are "neighbors" of this one, i.e.
+// close enough to see or to reach via jump drive.
+void System::UpdateNeighbors(const Set<System> &systems, double distance)
+{
+	set<const System *> &neighborSet = neighbors[distance];
+	
+	// Every star system that is linked to this one is automatically a neighbor,
+	// even if it is farther away than the maximum distance.
+	for(const System *system : links)
+		if(!(system->Position().Distance(position) <= distance))
```
petervdmeer, line 727: This if-statement is not needed; an element will only be in a set once, even if it gets inserted multiple times.

Amazinite: Looks like this was a holdover from when links and neighbors were vectors. Removed the if statement.

source/PlayerInfo.cpp
```diff
@@ -385,7 +385,11 @@ void PlayerInfo::AddChanges(list<DataNode> &changes)
 		// Recalculate what systems have been seen.
 		GameData::UpdateSystems();
 		seen.clear();
-		double jumpRange = (flagship && flagship->JumpRange()) ? flagship->JumpRange() : System::DEFAULT_NEIGHBOR_DISTANCE;
+		double jumpRange;
```
petervdmeer, line 388: Doesn't this give an uninitialized variable warning?

source/PlayerInfo.cpp
```diff
@@ -385,7 +385,11 @@ void PlayerInfo::AddChanges(list<DataNode> &changes)
...
+		if(flagship)
+			jumpRange = flagship->JumpRange();
```
petervdmeer, line 390: What if the JumpRange is lower than System::DEFAULT_NEIGHBOR_DISTANCE? Don't we want the planets within default_neighbor_distance to be also seen in that case? (Same as when the player had no jumpdrive.)

source/PlayerInfo.cpp
```diff
@@ -1841,7 +1845,11 @@ void PlayerInfo::Visit(const System *system)
 	
 	visitedSystems.insert(system);
 	seen.insert(system);
-	double jumpRange = (flagship && flagship->JumpRange()) ? flagship->JumpRange() : System::DEFAULT_NEIGHBOR_DISTANCE;
+	double jumpRange;
+	if(flagship)
+		jumpRange = flagship->JumpRange();
+	if(!jumpRange)
+		jumpRange = System::DEFAULT_NEIGHBOR_DISTANCE;
```
petervdmeer, line 1852: This code looks similar as the code above in AddChanges(changedSystems==true), can't we just call Visit again in the code above to avoid code duplication?

source/PlayerInfo.cpp
Amazinite: I get 0 warnings on compiling this.

source/PlayerInfo.cpp
Amazinite: The way I'm looking at it is that if the player can jump to it, the player can see it. If the player can't jump to it, then the player can't see it.

source/PlayerInfo.cpp
Amazinite: I don't see how Visit could be called from AddChanges. Doing so would just keep recalculating the player's jump range (which is what I wanted to avoid in one of the latest commits). If we instead hand the jump range off to Visit from AddChanges, then wherever Visit gets called elsewhere then needs to calculate the jump range, so the code will just be duplicated elsewhere.

source/PlayerInfo.cpp
petervdmeer: Some compilers (probably including yours) perform default (zero) initialization for local variables, but I understood that the C++ specification does not require default initialization for local variables.
If other compilers don't perform default initialization, then this can lead to hard to find and hard to fix bugs when the code gets compiled with a different compiler, so please do perform default initialization of this local variable.

source/PlayerInfo.cpp
petervdmeer: It feels strange to me that a player without a jumpdrive can see systems further away than a player with a "smaller-than-normal-range" jumpdrive.
Somehow I would expect a minimum vision distance to always apply (even if you do have a jump drive and cannot jump to it). And if you then want to hide systems, then hide them at a "just-beyond-normal-range" location.

source/PlayerInfo.cpp
petervdmeer: Okay, this is a trade-off between performance and some code-duplication. The systems-changed/Visit code is not highly performance critical, but I'm fine with you making this choice here.

source/PlayerInfo.cpp
Amazinite: What if the system itself has a jump range smaller than the default? Should that also be ignored for the view range?

Amazinite: Right now the ring drawn on the map is the jump range, but the comments in that section of the code describe it as the visible systems range. Should a second ring be drawn for the visible range if the jump and visible range differ (i.e. if the player's jump range is less than 100 due to their jump drive)?

petervdmeer: > What if the system itself has a jump range smaller than the default? Should that also be ignored for the view range? 

Yes, I feel this should also use the default view-range.

> Should a second ring be drawn for the visible range if the jump and visible range differ (i.e. if the player's jump range is less than 100 due to their jump drive)?

Sounds appropriate, for the system that the player is in. But I can imagine that having 2 rings can be confusing, so it would also be okay for me if in that case only the jump-range ring is drawn for the system that the player is in. (Because the player will be more interested in where he/she can jump to. The player will see the systems anyway.)

source/PlayerInfo.cpp
```diff
@@ -383,12 +383,24 @@ void PlayerInfo::AddChanges(list<DataNode> &changes)
...
+			double flagshipRange = flagship->JumpRange();
+			if(flagshipRange > jumpRange)
+				jumpRange = flagshipRange;
+		}
 		for(const System *system : visitedSystems)
 		{
 			seen.insert(system);
-			for(const System *neighbor : system->Neighbors())
+			for(const System *neighbor : system->Neighbors(jumpRange))
 				seen.insert(neighbor);
```
petervdmeer, line 404: Some things that just wondered about:
- Is seen also updated/extended at the moment I install a new Jump-drive with a higher range?
- Is this updated for all previously visited systems, or only for the current one? (And what would we prefer?)

source/PlayerInfo.cpp
Amazinite: > Yes, I feel this should also use the default view-range.

I’m not so sure on that, as the argument for having the view range be 100 if the player’s jump drive has a lesser range than that doesn’t also apply to if the system’s jump range is lesser. That is, in the situation where the player’s JD range is less than 100, they could remove the JD to increase their view range (which admittedly doesn’t make much sense), but in the case of the system’s jump range being lesser, whether the player does or does not have a JD doesn’t affect the view range, as it is always static and set by the system.

source/PlayerInfo.cpp
Amazinite: In the case of PlayerInfo::AddChanges, this is only a one time thing after an event changes systems. It updates the seen systems using your jump range at the time of the change for all visited systems. Changing your jump range after this has occurred doesn't change which systems are seen.
PlayerInfo::Visit on the other hand updates the seen systems from your current jump range upon entering a system (not when departing a planet), buying a map, having an event set a system as seen, or loading a pilot.

This means that if you visited a system when your jump range was 100, then later gain a jump range of 200, the seen systems from the system that you only ever visited with 100 updates to show the systems within 200 whenever AddChanges detects a system change, or when the player reloads the pilot. This also means that changing your jump drive does not update the seen systems for the current system until you reenter it (although the map panel will update with the proper jump range each time you change your JD).

I believe I could easily change the seen systems to also update on planet departure so that changing your JD and then departing shows the proper seen systems (although I'd have to look somewhere else to make it so that seen systems update as soon as your jump range changes), but to only update seen systems with the jump range that the player had when they last entered the system would be more involved (and it may be the case that we don't even want to do that so that an event change can immediately reveal systems across the galaxy instead of requiring the player to travel there with an updated range).

source/PlayerInfo.cpp
petervdmeer: Ok, when the system has a jump-range smaller than the view-range, then I would also be okay with the visual range being reduced to the smaller jump-range.

source/PlayerInfo.cpp
tehhowch: Isn't it lored that seen systems are observed via parallax measurements? If so, then seen should always be based solely on the default neighbor distance and hyperspace links, no matter what the jump drive range is.

Amazinite: Where is that “lored”? Searching for “parallax” returns no results, and I can’t even find any references to “hyperspace” that mentions how systems become seen.

tehhowch: It comes from the player's manual: https://github.com/endless-sky/endless-sky/wiki/PlayersManual#interstellar-travel

Having a larger jump range than visible range would make it possible to gate discovering a system behind having a mission active to that system (since active mission destination systems appear in the map screen even when you have no known route to it). After visiting the system, you would then be able to return to it without an active mission.

Amazinite: Well using parallax to get the location of systems would imply that you need to visit two nearby systems to actually "see" a system, not just one as is the case right now, so that explanation doesn't make much sense as is. And even if we stick with the parallax explanation, what's stopping a longer range jump drive from including better measurement tools for locating systems?

tehhowch: Even though we have only visited one system, we have used parallax to compute the positions of a great number of other systems. (That the game does not instantly reveal the whole map to everyone who connects their navigation computer to the galactic web, vs just those nearby systems, I don't think we will ever really explain in lore.)

I don't think we should drive jump range and system discovery from the same attribute, because it limits creative possibilities. Exactly as you suggest, what's to say some different drive doesn't have better sensors? So rather than use "jump range" for that too, we should use a different attribute, e.g. "parallax sensitivity" or "jump sensor range," whose value is used for computing seen neighbors from the current system.

Even if we don't allow modifying the discovery range, I believe everything we could do with a larger jump+discovery range can be done with just the jump range boost + clever use of missions, but not vice-versa.

Amazinite: Well it's not too difficult a change, so I'm fine with leaving the view range as is.

source/MapPanel.cpp
```diff
@@ -116,6 +116,12 @@ MapPanel::MapPanel(PlayerInfo &player, int commodity, const System *special)
 	hoverText.SetWrapWidth(150);
 	hoverText.SetAlignment(WrappedText::LEFT);
 	
+	// Find out how far the player is able to jump. The range of the system
+	// takes priority over the range of the player's flagship.
+	double systemRange = playerSystem ? playerSystem->JumpRange() : 0.;
+	double playerRange = player.Flagship() ? player.Flagship()->JumpRange() : 0.;
+	playerJumpDistance = systemRange ? systemRange : (playerRange ? playerRange : System::DEFAULT_NEIGHBOR_DISTANCE);
```
tehhowch, line 123: You can get rid of a ternary level by initializing to the constant in the member initialization section (i.e. `) : <in here> {` ) or in the header.

source/Ship.cpp
```diff
@@ -2630,15 +2630,45 @@ double Ship::JumpFuel(const System *destination) const
...
+	for(const auto &it : outfits)
+		if(it.first->Get("jump drive"))
+		{
+			double range = it.first->Get("jump range");
+			if(!range)
+				range = System::DEFAULT_NEIGHBOR_DISTANCE;
+			if(!best || range > best)
+				best = range;
+		}
+	return best;
```
tehhowch, line 2667: How frequently is this method called? We might want to consider caching the best value and updating it only when the ship's outfit list changes, if it gets called a lot.

source/Ship.cpp
```diff
@@ -3356,26 +3386,46 @@ double Ship::MinimumHull() const
 
 
 // Find out how much fuel is consumed by the hyperdrive of the given type.
-double Ship::BestFuel(const string &type, const string &subtype, double defaultFuel) const
+double Ship::BestFuel(const string &type, const string &subtype, double defaultFuel, double distance) const
```
tehhowch, line 3389: It's not great to have multiple consecutive parameters of the same type, but I don't know that we can easily change that here. Based on your comment below though, the name should be e.g. `jumpDistance`

source/Ship.cpp
```diff
@@ -3356,26 +3386,46 @@ double Ship::MinimumHull() const
...
 	// Search through all the outfits.
 	for(const auto &it : outfits)
 		if(it.first->Get(type) && (subtype.empty() || it.first->Get(subtype)))
 		{
-			double fuel = it.first->Get("jump fuel");
-			if(!fuel)
-				fuel = defaultFuel;
-			if(!best || fuel < best)
-				best = fuel;
+			double jumpRange = it.first->Get("jump range");
```
tehhowch, line 3418: Thoughts on writing a local lambda that gets called with an Outfit& and the current best fuel double&? We use nearly identical logic to inspect the base attribute outfit and each of the installed outfits.

source/GameData.cpp
```diff
@@ -549,21 +554,28 @@ void GameData::Change(const DataNode &node)
...
+// This must be done any time that a change creates or moves a system.
+void GameData::UpdateSystems()
 {
 	for(auto &it : systems)
 	{
 		// Skip systems that have no name.
 		if(it.first.empty() || it.second.Name().empty())
 			continue;
-		it.second.UpdateNeighbors(systems);
+		it.second.UpdateSystem(systems, neighborDistances);
```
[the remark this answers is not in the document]
Amazinite: This method actually does do more than just update the neighbor sets. It does that, but it also updates the solar power and wind of the system as well as any auto attributes (currently only "uninhabited").

source/Ship.cpp
Amazinite: The most frequent instance in which this will be called is when the ship wants to jump, so right now on every jump it is figuring out its jump range.

source/Ship.cpp
Amazinite: ¯\_(ツ)_/¯

source/Fleet.cpp
```diff
@@ -270,11 +270,13 @@ void Fleet::Enter(const System &system, list<shared_ptr<Ship>> &ships, const Pla
 		// drives and hyperdrives.
 		bool hasJump = false;
 		bool hasHyper = false;
+		double jumpDistance = System::DEFAULT_NEIGHBOR_DISTANCE;
 		for(const Ship *ship : variant.ships)
 		{
 			if(ship->Attributes().Get("jump drive"))
 			{
 				hasJump = true;
+				jumpDistance = ship->JumpRange();
```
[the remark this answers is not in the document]
Amazinite: ¯\\\_(ツ)\_/¯

Zatishko:
Downloaded on the 21st. Has the behavior I wanted for jump drive range of 52 and a fraction, which is won't jump most places, only select close pairs, which is around four pairs that aren't linked by hyperlane or, without a long path around.

Zitchas:
Awesome! I'm excited to start working on the possibilities of this. Thanks for working on it, Amazinite!

Issue #5204, opened 2020-08-03
Allow Jump Drives to have the "scram drive" attribute and behave like Scram Drives


jchamlin:
**Is your feature request related to a problem? Please describe.**

When jumping into a system with a Jump Drive, all your ships are motionless, pointed in all different directions (whatever direction they needed to be in to come to a full stop in the origin system), and randomly distributed in an area that is very close to the landing spot calculated for the system. This is also where enemies in the system spawn, so you usually jump right on top of them and come under fire instantly. Since you are motionless and appear on top of the enemy fleet, you are a sitting duck and often get pummeled, even while the jump animation is finishing before your ships can even start moving or shooting back. The solution is to try to press G to gather your fleet around your flagship, and then run like wind to try to get out of the middle. The problem is that ships that are not aligned with the flagship take time to rotate to the flagship's heading first, and then start accelerating to move towards the flagship. For the ships that are way off in their orientation, they often die before they can escape.

I thought it might be nice to make it so that holding the J key for longer than necessary to just simply prep for jump with jump drives should possibly have all your ships align towards the destination system (like they do for a Hyperspace jump), even though they don't need to. It would just be nice if they did anyway. They have time to continue to rotate,  but they choose not to (and then often die because of it). Of course, you can always release J before all ships are aligned but after all ships have stopped to get the whole fleet to jump at the same time before all ships are rotated to the proper orientation. But a conversation with @MasterOfGrey  on the Discord server revealed that this might not be the best solution since he uses the fact that all the ships have stopped moving as the key to know that they are all ready to jump. He suggested using a different keyboard key instead. Or maybe it could be a UI preference to control what holding down J does. MasterOfGrey posted the idea of using a different key in #5191.

[...]

PR #5205, opened 2020-08-03
feat: Allow the scram drive attribute to be applied to Jump Drives

jchamlin:
Enables a scram-enabled jump drive outfit.

**Feature:** This PR implements the feature request detailed and discussed in issue #5204 

## Feature Details

Allows the "scram drive" attribute to be applied to both hyperdrives (which makes a scram drive) and jump drives (which makes a scram-enabled jump drive). When applied to a jump drive, it makes the ship align to the target system (like a scram drive) but jump like a jump drive. The changes have zero impact on existing drives, Hyperdrives, Scram Drives, and Jump Drives (and ships with those bulit-in) all behave the same way the did before the code change. The code changes only enable new behavior if you apply the "scram drive" attribute to a jump drive outfit (or apply both attributes to a ship to have a built-in scram-enabled jump capability).

## UI Screenshots

N/A

## Usage Examples

It is not currently used in the data files. I plan on adding a new outfit, the Slip Drive (Quantum Slipstream Drive) which is a scram-enabled jump drive to plugins. Here's the outfit I used for testing:

```
outfit "Slip Drive"
	category "Systems"
	cost 5000000
	thumbnail "outfit/slip drive"
	"mass" 20
	"outfit space" -20
	"jump fuel" 175
	"scram drive" 0.2
	"jump drive" 1
	description `The Quantum Slipstream Drive, nicknamed the "Slip Drive", allows ships to travel between systems without a hyperlink connection. A slipstream is a narrowly focused, directed field that is initiated by manipulating the fabric of the space-time continuum at the quantum level. This creates a subspace tunnel, which is projected ahead of the vessel. Once a ship has entered the slipstream, the forces inside propel it at incredible speed. Like a Scram Drive, ships must be oriented towards the destination system before jumping and can be in motion during the jump. A sequence of jumps with a Slip Drive takes more time than a Jump Drive and the ships are vulnerable to attack between jumps because ships must reorient towards the next target system before each jump.`
```

## Testing Done

Tested setting targets on the map, and jumping across multiple systems (mixed hyperspace and jump), with multiple drive configurations, including ships with multiple drives and fleets of ships with mixed drive configurations, including all 16 possible combinations:
* Ship with no drives
* Hyperdrive
* Scram Drive
* Jump Drive
* HyperDrive + Scram Drive
* HyperDrive + Jump Drive
* Scram Drive + Jump Drive
* HyperDrive + Scram Drive + Jump Drive
* Slip Drive
* Hyperdrive + Slip Drive
* Scram Drive + Slip Drive
* Jump Drive + Slip Drive
* HyperDrive + Scram Drive + Slip Drive
* HyperDrive + Jump Drive + Slip Drive
* Scram Drive + Jump Drive + Slip Drive
* HyperDrive + Scram Drive + Jump Drive + Slip Drive

## Performance Impact

There are a few places in the code where the CPU will have to search the outfits list of a ship instead of grabbing the "scram drive" attribute right off the ship's composite attribute list, but the outfits lists are fairly small and bounded so iterating the list is pretty quick, and the PrepForJump, IsReadyForJump, and JumpFuel calculations are only done when plotting courses and actually jumping.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2280,14 +2280,20 @@ bool Ship::IsReadyToJump(bool waitingIsReady) const
 	if(!fuelCost || fuel < fuelCost)
 		return false;
 	
-	Point direction = targetSystem->Position() - currentSystem->Position();
-	bool isJump = !attributes.Get("hyperdrive") || !currentSystem->Links().count(targetSystem);
-	double scramThreshold = attributes.Get("scram drive");
+	bool hasHyperdrive = attributes.Get("hyperdrive");
+	bool isJumping = !hasHyperdrive || !currentSystem->Links().count(targetSystem);
 	
-	// The ship can only enter hyperspace if it is traveling slowly enough
-	// and pointed in the right direction.
-	if(!isJump && scramThreshold)
+	bool hasHyperScramDrive = HasOutfit("hyperdrive", "scram drive");
+	bool hasJumpScramDrive = HasOutfit("jump drive", "scram drive");
+	bool isScramming = (isJumping && hasJumpScramDrive) || (!isJumping && hasHyperScramDrive);
+	
+	Point direction = targetSystem->Position() - currentSystem->Position();
+	
+	// If scramming, the ship can only enter hyperspace or jump if it is traveling 
+	// slowly enough and pointed in the right direction.
+	if(isScramming)
 	{
+		double scramThreshold = attributes.Get("scram drive");
 		double deviation = fabs(direction.Unit().Cross(velocity));
 		if(deviation > scramThreshold)
 			return false;
@@ -2295,7 +2301,7 @@ bool Ship::IsReadyToJump(bool waitingIsReady) const
 	else if(velocity.Length() > attributes.Get("jump speed"))
 		return false;
 	
-	if(!isJump)
+	if(isScramming || !isJumping)
 	{
 		// Figure out if we're within one turn step of facing this system.
 		bool left = direction.Cross(angle.Unit()) < 0.;
@@ -2646,7 +2652,7 @@ double Ship::HyperdriveFuel() const
 	if(!attributes.Get("hyperdrive"))
 		return JumpDriveFuel();
 	
-	if(attributes.Get("scram drive"))
+	if(HasOutfit("hyperdrive", "scram drive"))
 		return BestFuel("hyperdrive", "scram drive", 150.);
 	
 	return BestFuel("hyperdrive", "", 100.);
@@ -2659,6 +2665,9 @@ double Ship::JumpDriveFuel() const
 	// Don't bother searching through the outfits if there is no jump drive.
 	if(!attributes.Get("jump drive"))
 		return 0.;
+	
+	if(HasOutfit("jump drive", "scram drive"))
+		return BestFuel("jump drive", "scram drive", 250.);
 	
 	return BestFuel("jump drive", "", 200.);
 }
@@ -3139,7 +3148,19 @@ void Ship::AddOutfit(const Outfit *outfit, int count)
 	}
 }
 
-
+// Determine if the ship has an outfit with the specified type and optionally subtype
+bool Ship::HasOutfit(const string &type, const string &subtype) const
+{
+	// Make it possible for the outfit to be integrated directly into the ship
+	if(baseAttributes.Get(type) && (subtype.empty() || baseAttributes.Get(subtype)))
+		return true;
+	// Search through all the outfits.
+	for(const auto &it : outfits)
+		if(it.first->Get(type) && (subtype.empty() || it.first->Get(subtype)))
+			return true;
+	// The ship does not have the outfit with the type(s) requested
+	return false;
+}
 
 // Get the list of weapons.
 Armament &Ship::GetArmament()
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1841,20 +1841,27 @@ bool AI::Stop(Ship &ship, Command &command, double maxSpeed, const Point directi
 void AI::PrepareForHyperspace(Ship &ship, Command &command)
 {
 	bool hasHyperdrive = ship.Attributes().Get("hyperdrive");
-	double scramThreshold = ship.Attributes().Get("scram drive");
 	bool hasJumpDrive = ship.Attributes().Get("jump drive");
+	// Can't jump without a drive
 	if(!hasHyperdrive && !hasJumpDrive)
 		return;
 	
-	bool isJump = !hasHyperdrive || !ship.GetSystem()->Links().count(ship.GetTargetSystem());
+	bool hasHyperScramDrive = ship.HasOutfit("hyperdrive", "scram drive");
+	bool hasJumpScramDrive = ship.HasOutfit("jump drive", "scram drive");
+	bool isJumping = !hasHyperdrive || !ship.GetSystem()->Links().count(ship.GetTargetSystem());
+	bool isScramming = (isJumping && hasJumpScramDrive) || (!isJumping && hasHyperScramDrive);
 	
 	Point direction = ship.GetTargetSystem()->Position() - ship.GetSystem()->Position();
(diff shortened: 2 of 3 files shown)


source/Ship.cpp
```diff
@@ -2660,6 +2670,9 @@ double Ship::JumpDriveFuel() const
 	if(!attributes.Get("jump drive"))
 		return 0.;
 	
+	if(HasOutfit("jump drive", "scram drive"))
+		return BestFuel("jump drive", "scram drive", 175.);
```
Amazinite, line 2674: Why would a jump drive with more abilities than a normal one have a lower default fuel? 250 makes more sense to me than 175.

source/Ship.cpp
```diff
@@ -3446,3 +3459,17 @@ void Ship::CreateSparks(vector<Visual> &visuals, const Effect *effect, double am
 			visuals.emplace_back(*effect, angle.Rotate(point) + position, velocity, angle);
 	}
 }
+
+// Determine if the ship has an outfit with the specified type and optionally subtype
+bool Ship::HasOutfit(const string &type, const string &subtype) const
```
Amazinite, line 3464: New functions should be placed in the same spot as in the header files. i.e. don't put this at the bottom of the file, but it below AddOutfit.

source/Ship.cpp
```diff
@@ -2280,22 +2280,32 @@ bool Ship::IsReadyToJump(bool waitingIsReady) const
 	if(!fuelCost || fuel < fuelCost)
 		return false;
 	
+	bool hasHyperdrive = attributes.Get("hyperdrive");
+	bool hasJumpDrive = attributes.Get("jump drive");
+	// Can't jump without a drive
+	if(!hasHyperdrive && !hasJumpDrive)
+		return false;
```
Amazinite, line 2287: This is redundant with the fuelCost check above; if the player doesn't have a jump drive or a hyperdrive, JumpFuel will return 0. This also makes the hasHyperdrive and hasJumpDrive bools unnecessary.

source/Ship.cpp
Amazinite, line 2674: It seems to me like a scram jump drive would be better than a normal jump drive, so I'd expect the default cost to be more than a normal JD, not less. Perhaps 250 instead of 175.

jchamlin: I would have thought that too. But in reality it is somewhat better in specific circumstances (jumping into combat you intend to fight) and somewhat worse otherwise: having to align to the destination to jump means that multiple jumps actually takes longer, sometimes a lot longer for slow turning ships, and if you jump into combat you want to escape from, you are more vulnerable because you spend longer in the system aligning to the new destination and take more fire before you jump out. Thus, I think it has some advantages, and some disadvantages. Since the normal case is that you're just jumping a bunch of hops to get somewhere and usually avoiding combat, the scram-enabled jump drive is actually worse than a regular jump drive, this reducing the fuel costs seemed to be a way to balance out the negatives. You take a little longer each jump prep, but use a little less fuel per jump, so people will have to make a choice on which drive they actually prefer. Thoughts?

source/Ship.cpp
jchamlin: I will make the correction and post a commit with the update.

source/Ship.cpp
jchamlin: I will make the correction and push an update with the fix.

jchamlin:
Merged master from from upstream and reapplied commit, including code review changes requested by Amazinite.

petervdmeer:
The use of `scram drive` doesn't feel appropriate for jump-drives here, because a scram drive for hyperspace allows hyperjumping at higher speeds, it does not seem to be meant to restrict facing-alignment.

Adding attributes like the following feels more correct to me:
- Put a facing-alignment with destination-direction requirement (`align <angle>`?) on drives;
  - This is none/zero for jumpdrives (your ship doesn't need to face in the direction you are jumping).
  - This is exact match for hyperdrives and scram-drives (your ship does need to face exactly in the direction you are jumping).
- Put a maximum speed for jumping (`max speed <nr>`) on drives;
  - This is zero for hyperdrives.
  - This is zero for jump-drives.
  - This is between zero and full-ship-speed for scram-drives.

As a bonus, having a minimum speed (`min speed <nr>`, where nr is an absolute number, not relative to the ships max speed) for jumping would also be nice to build drives where you cannot jump from stand-still.

The "fast move" animation for hyperdrives would also match better with the focussed tunnelling as described for the slip-drive described here (instead of the fade-out/fade-in attribute of jump-drives), so maybe it would be nice to also make that a property/attribute `"system change animation" [moving] [fading]`(but I would recommend to keep that for a separate PR).

jchamlin:
@petervdmeer, yep, all great suggestions and all things I've thought of. I think genericizing all the attributes of drives so multiple drive types and animations can be added is the right call, but that's a much more invasive change since there are only two drive types and they are hard-coded all over the ES code base. Also, being able to specify per-drive effects has already been done, but adding new transitions (like a glowing blue web hyperspace-like transition for a Slipstream Drive) us my ultimate goal. However, baby steps. This is my first contribution to ES and I wanted to submit something that was the minimum code changes necessary to implement something which provides the remediation for the "jump into a ball of death" problem, and also make something logical but that didn't work in the code (scram-enabled jump drive) actually work properly in the code. I've got one more idea to try which might make the scram-enabled jump drive fully superior to the regular jump drive, and that's aligning to the flagship's direction for scramming instead of aligning to the destination for scramming. That would make multi-jumps as fast as a regular jump drive, getting rid of the scram-enabled jump drive's biggest weakness. I'll try to implement it and see how it works out. I'll update the PR with the code if it does.

tehhowch:
As I mentioned in the backing issue, I don't think there is consensus on the behavior for combining the "jump drive" and "scram drive" behaviors. Please use the issue (or even better, submit an RFC in the EndlessSkyCommunity repo) to refine the topic and identify the actual implementation that is desired by content creators.

jchamlin:
Can we just start with the simplest change possible, just making it so the code actually jumps the ship if it encounters a Jump Drive with the "scram drive" attribute and no Hyperdrive on the ship (i.e. this PR)? Right now, if that happens, when you press J the ship doesn't even jump. With this PR, at least the code does something, even if it isn't the final behavior that the community determines is the desired behavior. I will open an RFC as you requested, but it would be nice if that behavior and whatever comes from the discussion becomes a new Issue and a new PR, and doesn't hold up merging this PR.

tehhowch:
The smallest change is to only care about the "scram drive" attribute when performing a jump via hyperdrive, which is _not_ what this PR currently does. Currently, a new behavior for the combination of "jump drive" + "scram drive" is implemented.

vitalchip:
I'll just throw in an idea I've implemented.  I have added "Jump Offset" as a possible attribute for Jump Drives.  There's no impact at all for any existing drives but should a new drive be created with "Jump Offset" 500 ( for example ) then 3 new things happen. The first is that the jumping ship arrives at 500px further from the planet than a regular jump. The second is that all jumping ships arrive pointing in their departure direction but facing the planet, they will still be scattered round the clock face, depending on which direction they jumped from. The third is that if the player holds the jump key down, all ships ( having drives with the attribute) will align to face the destination system. All ships will thus arrive on the same vector, pointing at the planet.  I feel this solves the often complained about "jumping too close to the planet" and also allows a sort of "formation arrival" to be implemented using the familiar "hold to align" jump procedure.

petervdmeer:
RFCs and discussions on features and their exact keywords can be frustrating, but they also bring a lot of value in getting us very stable keywords for very stable features that content-creators can depend upon when creating new content.
This PR in it's current state is introducing a temporary and ambiguous use of the `"scram drive"` keyword, which would not benefit content-creators. The PR is also idle since 2020, so I feel it makes sense to close this PR as-is.

@jchamlin I like the ideas you are proposing about more drive configurability, so please feel welcome to continue participating in the development (discussions, RFCs, etc) of those features and feel welcome to create another PR once those features are sufficiently stable.

PR #5206, opened 2020-08-04
Content: Breaking out Chilia logs (adjustment, not new content)

Zitchas:
**Content**

## Summary
This pulls the two Chilia log entries out of their respective missions, and creates two log entry missions so that the player will always get the two entries in the specific order. It also centralizes introduction logs to avoid risk of duplicates logs.

## Visible Impact

This does not change the first log entry for Chilia, but it does remove the reference to him being the Defence Prefect from the second. As such, a the visible impact of this break out will be that the log entries no longer duplicate each other.

## Behind the scenes impact

As there are multiple ways and orders the player can meet Chilia, this break-out means that regardless of order, the player will always get the first log entry (that identifies him as being the defense prefect) first, and the second one afterwards. Thus the second one doesn't need to mention him being the defense prefect. It also means that other places that introduce Chilia don't need new log entries, just increment the `"remnant chilia"` variable. 

This also follows the example used by Taely.

This change will be used by the Cognizance PR.

### DIFF
--- a/data/remnant/remnant missions.txt
+++ b/data/remnant/remnant missions.txt
@@ -56,7 +56,6 @@ mission "First Contact: Remnant"
 				decline
 			
 	on enter
-		log "People" "Chilia" `Despite being quite young, Chilia is the Remnant Prefect in charge of Remnant military actions. He is highly respected among Remnant society.`
 		"remnant chilia" ++
 		conversation
 			`As you break orbit and get ready to escape, you receive a message from <planet>. A chant comes from your speakers.`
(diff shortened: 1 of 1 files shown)


data/remnant/remnant missions.txt
```diff
@@ -1274,6 +1273,30 @@ mission "Remnant: taely intro log"
 
 
 
+mission "Remnant: Chilia Intro Log"
+	invisible
+	source
+		government "Remnant"
```
Amazinite, line 1279: I'd remove the sources. They're pretty redundant with how you're only ever given the "remnant chilia" attribute while in Remnant space.

PR #5207, opened 2020-08-04
Kor Efreti spaceport news

W1zrad:
## Summary
Adds 17 spaceport news dialogues for Kor Efret planets/stations (none of which is actual dialogue, as the player does not know Korath language). I tried to elaborate on the existing planet/spaceport descriptions of the dichotomy of market stalls and plazas in the middle of cities. Added a new minor aspect to existing lore:
Cities are partly abandoned/dilapidated due to the civil war: buildings range from brand new to structurally unstable. I thought this was appropriate considering the Korath's role as a declining species.

Below are a couple screenshots of the news.
korefretinews1
korefretinews2

## Save File
[Test Pilot~korathnews.txt](https://github.com/endless-sky/endless-sky/files/5023874/Test.Pilot.korathnews.txt)

### DIFF
--- a/data/korath/korath news.txt
+++ b/data/korath/korath news.txt
@@ -11,6 +11,7 @@
 news "kor efret merchant"
 	location
 		government "Kor Efret"
+		not attributes "station"
 	name
 		word
 			"Korath merchant"
@@ -28,6 +29,7 @@ news "kor efret merchant"
 news "kor efret worker"
 	location
 		government "Kor Efret"
+		not attributes "station"
 	name
 		word
 			"Korath worker"
@@ -42,6 +44,7 @@ news "kor efret worker"
 news "kor efret local"
 	location
 		government "Kor Efret"
+		not attributes "station"
 	name
 		word
 			"Korath local"
@@ -51,15 +54,13 @@ news "kor efret local"
 	message
 		word
 			"A ball chased by a young Korath rolls up to your feet. You pick it up, and the youth pauses and stares at you for a moment before gesturing in thanks and accepting the ball. It runs back to a grown Korath who holds both its hands up with palms toward you."
-			"There's a Korath nearby precariously balancing several boxes above its head. You can't tell if it's a street performer or just transporting the crates."
 			"A local Korath gestures toward a building you've never been inside before. Out of curiousity, you poke your head inside to see the reptilian aliens conversing and sharing drinks: it's a bar. Some things never change, no matter where you are in the galaxy."
-			"A Kor Efret near your ship is speaking with a worker, gradually increasing its volume until it's doing some strange mix of shouting and hissing. Suddenly, it turns and stalks away with its chest puffed out."
-			"You try asking one of the Korath about the abandoned buildings littered throughout the spaceport, but it's no good. You're unable to convey the strange dichotomy the mixture of old and new makes; in fact, it seems to be the norm here."
 			"You pass by one of the abandoned buildings and hear a door open and close rapidly behind you. You turn back and see a Korath dressed in rags quickly walking the other direction."
 
 news "kor efret scientist"
 	location
 		government "Kor Efret"
+		not attributes "station"
 	name
 		word
 			"Korath scientist"
@@ -70,3 +71,19 @@ news "kor efret scientist"
 		word
 			"A Kor Efret pushes a cart filled with strange chemicals and bits of technology past you."
 			"A Korath walks through the spaceport reading numbers off a device and inputting them onto another one. It looks like it's taking samples of the air quality."
+			
+news "kor efret station friendly"		
+	location
+		government "Kor Efret"
+	name
+		word
+			"Korath local"
+			"Kor Efret local"
+			"Korath"
+			"Kor Efret"
+	message
+		word
+			"There's a Korath nearby precariously balancing several boxes above its head. You can't tell if it's a street performer or just transporting the crates."
+			"A Kor Efret near your ship is speaking with a worker, gradually increasing its volume until it's doing some strange mix of shouting and hissing. Suddenly, it turns and stalks away with its chest puffed out."
+			"A Korath is studying you intently from a distance. You guess it's never seen a human face to face before."
+			"A local greets you by holding its hands up with its palms facing toward you."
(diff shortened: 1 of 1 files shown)


Arachi-Lover:
`2 - Cities are partly abandoned/dilapidated due to the civil war: buildings range from brand new to structurally unstable. I thought this was appropriate considering the Korath's role as a declining species.`

no Kor Efreti world shows any signs of having been a site of fighting during their Civil War, rather as they tell Rek, and she translates `The Kor Efret were those who sought to remain neutral and not join in the fighting.`; and before the inevitable "take everything with a grain of salt" argument comes up, the Efreti have zero reason to lie to the player, and I don't see how they could possibly not know what happened during their civil war. point being, I don't see that as a very natural conclusion to become part of news.

on these `"A Korath wearing an insignia - probably to designate it as some sort of government official - strides through the spaceport with a strange device. It looks like it's taking samples of the air quality."` , `"One Korath wearing a toga-like outfit walks directly toward you. You barely manage to get out of the way before it bumps into you, and it makes a hissing sound. As you watch it continue through the spaceport, every other Korath gives it plenty of space. You hope you didn't just cause a diplomatic crisis on behalf of your entire species."` , `"Nearby, a Korath with a toga-like robe speaks to a small group of noticably worse-dressed Kor Efreti. You don't know what they're talking about, but it's obvious they're being told what to do."` , as far as I remember there's no information at all on how the Kor Efreti government works, or how it is structured, and I wouldn't say spaceport news are the best place to first put that information. the other two, the entire Korath population is 1 Billion, most of them being Efreti. divided over 3 very poor worlds. how am I supposed to believe there's some sort of noble class in this that's defined by clothing and are so distinctly respected - for no reason other than what I assume would be wealth - in a society that's struggling to survive? why would the Efreti care about someone with better clothes being supposedly some big shot? how come these big shots even exist in Efreti society when all their worlds are in harsh conditions, and them having little to no trade to make a profit, profit which they could only spend on the limited goods their three poor worlds can provide? it doesn't make sense to me for the Efreti to exhibit such behavior over something like a dress when the status it could give to the individual or the individual wearing it would have isn't something I can realistically see exhisting in Efreti society due to their extremely limited resources and poor living conditions

W1zrad:
@Arachi-Lover 
To your first point, I think you're right about the Kor Efret worlds shouldn't have suffered in the civil war, but current planet/spaceport descriptions do still point to these planet. Take the description of Setar Fort, for example:

> There are several large Korath cities on this planet, but the ones farther from the equator have been covered by encroaching glaciers as the planet slowly enters a new ice age. A small number of crops are still grown in the open air, but many of the fields have now been covered with row upon row of gleaming glass as the Korath resort to greenhouse farming to lengthen the growing season and feed their remaining population more efficiently.

My bit about the cities being a mix of new and run-down was based on this line from the spaceport at Karek Fornati:

> The spaceport city is built around an enormous public square, with stalls around its periphery for produce or grain or cattle to be sold and shipped offworld. But only one corner of the square is still active as a market; the rest has fallen into disrepair.

So while they weren't directly affected by war, it seems to me that their worlds are - to some degree - damaged. The dialogue was intended to be interpreted that these worlds weren't necessarily war-damaged, but in decline - similar to the Rust Belt in the U.S. to a wider degree. Does that make sense? I'll go back and try to clarify that the disrepair isn't exactly from war.

To your second point, you're right that we haven't seen much about the Kor Efret government. This line can easily be reworked into just some random local going around testing air quality, so I'll do that until a larger discussion about fleshing out their government takes place (some of the stuff I'm currently working end up doing that, so I definitely need to talk that over with others).
Social hierarchy regardless of actual conditions is a given in almost every society. For example, Dark-Age Europe is partly characterized by the wealthy living in (relative) luxury with elaborate clothing while the poor starved. That said, the Korath are, in fact, not human, so whether human examples can be used as a basis for an alien species is largely up to in-game precedent - of which there hasn't been much. These lines are fluff, so I can remove them if you'd prefer.

Based on your overall response it seems like a lot of the ways I'm envisioning the Kor Efreti from a worldbuilding perspective don't line up with what you've set out here, and it's not like there's a huge amount to go off of in-game. Has there been any discussion about Kor Efreti government/culture/society that I'm unaware of that might help me understand how I should characterize their stuff that's not in-game yet, or is that a conversation that still needs to happen/be established?

Arachi-Lover:
overall I didn't have an issue with most of the stuff, that's why I only pointed to those things. on their society again, the Efreti are less than a Billion, spread across 3 planets, a station, and those that live in the ringworld. their one world that was dedicated to agriculture is now in a worse state than medieval europe as fertile lands are being overtaken by deserts, Setar Fort is very cold, which I don't believe is very comfortable for a species of reptile aliens, and their one remaining planet is polluted to no end. given this bit `Every once in a while, louder footsteps announce the presence of a group of Korath visiting the station.` from Kuwaru Efreti's spaceport, then this `This small station is still home to well over a thousand Korath, and they are constantly on the move from one section to another, their boots ringing loudly on the metal floors.` from Korati Efreti's spaceport, since we know the Korath in Korati Efreti wear some boots which make a great deal of noise, and that those in Kuwaru Efreti are also the most noticeable and recognizeable noise in the otherweise silent ringworld, and that to be visiting a ringworld those Korath must be fairly wealthy (in comparison to the rest), I don't think some fancy togas or dresses match that image very well. (don't quote me on this, but *I believe* in one of the Korath Exile stations it's stated they wear metal boots, which would also cause the loud ringing sound, and would suggest that Efreti and Exile fashion isn't very different, which isn't unbelievable since unlike with the humanity and the Remnant, the Exiles have only been out of their main space for ~1/40th of the total Korath spacefaring time); another thing that made me find odd the reverence the other Korath treated the well-dressed one in that news is that in all instances of interacting with the Efreti they don't exhibit some pretentiousness or want to show off their status, feel like it'd be almost contradictory that people belonging to the group closest to extinction in the game, which relies on another alien species to not be killed by the Automata, and has very few resources, would do something mundane like wanting to show off by wearing some fancy dress, and that those that saw said person would be gullible enough to think they really are in some sort of higher  social level

data/korath/korath news.txt
```diff
@@ -0,0 +1,72 @@
...
+			"Korath"
+			"Kor Efret"
+	message
+		word
+			"A ball chased by a young Korath rolls up to your feet. You pick it up, and the youth pauses and stares at you for a moment before gesturing in thanks and accepting the ball. It runs back to a grown Korath who holds both its hands up with palms toward you."
+			"There's a Korath nearby precariously balancing several boxes above its head. You can't tell if it's a street performer or just transporting the crates."
+			"A local Korath gestures toward a building you've never been inside before. Out of curiousity, you poke your head inside to see the reptilian aliens conversing and sharing drinks: it's a bar. Some things never change, no matter where you are in the galaxy."
+			"A Kor Efret near your ship is speaking with a worker, gradually increasing its volume until it's doing some strange mix of shouting and hissing. Suddenly, it turns and stalks away with its chest puffed out."
+			"You try asking one of the Korath about the abandoned buildings littered throughout the spaceport, but it's no good. You're unable to convey the strange dichotomy the mixture of old and new makes; in fact, it seems to be the norm here."
```
Amazinite, line 57: Not sure what this one is trying to say. Given that the player can't communicate with the Korath, I'd just remove it.

PR #5213, opened 2020-08-09
new merchant hail

Naminson:

## Summary
Very small change for very small people. Essentially just adds a hail phrase from the Merchants where they complain about how people aren't paying them enough these days.


Naminson:
ok

PR #5214, opened 2020-08-09
Fix typoes in "Asteroid Course"

Anarchist2:
Change `damages` to `damage`, change `my ship` to `your ship` to preserve second-person, switch the randoms the right way round (where the number is equal to the chance (out of 100) to get the outcome, rather than the number being equal to 100 minus the percentage chance).


Anarchist2:
frigg

PR #5217, opened 2020-08-11
fix Unable to initialize GLEW on wayland

eternal-sorrow:
-----------------------
**Bugfix:** This PR addresses issue flathub/io.github.endless_sky.endless_sky#1
## Fix Details
Glew fails to initialize on wayland because there is no GLX on wayland and distro-provided GLEW is build to forse use of GLX over EGL. This is upsteram GLEW issue nigels-com/glew#172. This PR ignores the GLEW_ERROR_NO_GLX_DISPLAY error. This is the temporary fix until this is fixed by GLEW upstream. @Pointedstick asked me to open PR here.

## Testing Done
Tried playing game on my linux laptop with sway. No crashes or other problems. 

## Save File
No need for save file, bug occurs just when you try to start the game on wayland. It occurs on both v0.9.12 which is in the flatpak and on v0.9.6 which is in my distro repository (Gentoo), and does not occur when using this branch's build.

### DIFF
--- a/source/GameWindow.cpp
+++ b/source/GameWindow.cpp
@@ -126,7 +126,7 @@ bool GameWindow::Init()
 	// Initialize GLEW.
 #ifndef __APPLE__
 	glewExperimental = GL_TRUE;
-	if(glewInit() != GLEW_OK){
+	if(GLenum err = glewInit(); GLEW_OK != err && err != GLEW_ERROR_NO_GLX_DISPLAY){
 		ExitWithError("Unable to initialize GLEW!");
 		return false;
 	}


eternal-sorrow:
CI builds failed due to missing GLEW_ERROR_NO_GLX_DISPLAY macro, but I don't know which GLEW version is used on these machines and why this macro isn't defined there.

MCOfficer:
> CI builds failed due to missing GLEW_ERROR_NO_GLX_DISPLAY macro, but I don't know which GLEW version is used on these machines and why this macro isn't defined there.

The CI uses [libglew-dev from ubuntu 18.04](https://packages.ubuntu.com/bionic/libglew-dev), and the development libraries from [here](https://github.com/endless-sky/endless-sky/wiki/BuildInstructions#development-libraries) for windows.

oo13:
C++11 does not permit an init-statement in if statement.

eternal-sorrow:
I don't see how

Pointedstick:
Why is this closed? You could just refactor it a tiny bit to be like this:

```
GLenum result = glewInit();
if(result != GLEW_OK && result != GLEW_ERROR_NO_GLX_DISPLAY){
    // blah
}
```

eternal-sorrow:
I have no Idea how to make this work with old versions of GLEW which apparently are supported because they are used on ubuntu and windows CI. There is no GLEW_ERROR_NO_GLX_DISPLAY macro and (as I could see in GLEW source code) the situation where no GLX display is detected is not handled there at all.

Pointedstick:
Add a GLEW version check then, maybe?

eternal-sorrow:
As far as I can tell, there is no API to get the GLEW version neither in form of macro nor some kind of function. Maybe I'm wrong. The only way I see is direct check `#ifdef GLEW_ERROR_NO_GLX_DISPLAY`.

PR #5223, opened 2020-08-13
Support selecting systems without planning route

Terin:
## Feature Details
This feature lets you use a modify key (control in this case) to prevent automatic route planning on the map panel. While routes are very easy to plan in Endless Sky, having a trip with 3-4 stops make re-planning it tedious. With this change, holding down control lets you directly select a system, cycling you through available missions on the Missions tab, or letting you preview commodity prices for a stop on the way on the Ports tab.

Future development could allow for an option to enable/disable automatic route planning, where this modifier would be the inverse of that (plot route if enabled, directly select if disabled), but in the meantime, this is a quality-of-life change that benefits players who take on many jobs at once.

## Testing Done
Tested for clicking on a variety of explored/unexplored systems, and using multiple modifiers. Control will overwrite shift's behavior due to the conditional block, and it's not planning a route in the first place.

### DIFF
--- a/source/MapPanel.cpp
+++ b/source/MapPanel.cpp
@@ -547,7 +547,8 @@ void MapPanel::Select(const System *system)
 	
 	bool shift = (SDL_GetModState() & KMOD_SHIFT) && !plan.empty();
 	bool ctrl = (SDL_GetModState() & KMOD_CTRL);
-	if (ctrl) {
+	if (ctrl) 
+	{
 		return;
 	}
 	else if(system == source && !shift)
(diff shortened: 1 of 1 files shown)


tehhowch:
Any particular reason for choosing Ctrl instead of another key, e.g. Alt or Super/Windows?

Please match the brace styling of the project (which in this case would be "omitted," due to the single statement in the block scope).

MCOfficer:
> Any particular reason for choosing Ctrl instead of another key, e.g. Alt or Super/Windows?


Super/Windows would collide with global start menu keybinds. Alt seems fine.

Terin:
I choose control, since my pinky tends to rest on the left edge of the keyboard when playing the game, and Super would cause lots of issues on Windows and potentially Mac, yeah. If there's a good case for Alt, though, I'd be more than happy to change it, as long as the functionality is available.

MCOfficer:
If it's just a personal preference thing, i'd lean towards ctrl.

PR #5229, opened 2020-08-14
Hot jupiters

beccabunny:

## Summary
This PR adds 11 different sprites of Hot Jupiters, a special type of planet that is supposed to be very close to the star, tidally locked to it and resembles a gas giant while having a much, much higher temperature https://en.wikipedia.org/wiki/Hot_Jupiter . Some planets all around the universe are so changed to be Hot Jupiters, 15 totally.

## Screenshots:
Screenshot (83)
Screenshot (85)
Screenshot (87)
*Some Hot Jupiters in-game.*

## PR Checklist
 ~~[ ] I uploaded the necessary image, blend, and texture assets here: {{insert link to assets}}~~
 - [ ] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: {{insert PR link}}


data/map.txt
```diff
@@ -265,7 +265,7 @@ system "11 Autumn Above"
 		period 14.9791
 		offset 180
 	object
-		sprite planet/gas1-b
+		sprite "planet/hot jupiter10"
```
Amazinite, line 268: To be in keeping with the naming scheme of existing planets, perhaps use `gas#-hot` as the naming scheme instead of `hot jupiter#`

beccabunny: Done, both here and in high DPI

data/map.txt
```diff
@@ -1973,7 +1973,7 @@ system Acrux
 		period 11.4474
 		offset 180
 	object
-		sprite planet/dust5
+		sprite "planet/gas2"
```
Amazinite, line 1976: Should this have `-hot`?
You also don't need quotes around any of these anymore since the sprite name has no spaces.

PR #5230, opened 2020-08-14
fix Unable to initialize GLEW on wayland

eternal-sorrow:
-----------------------
**Bugfix:** This PR addresses issue flathub/io.github.endless_sky.endless_sky#1
## Fix Details
Glew fails to initialize on wayland because there is no GLX on wayland and distro-provided GLEW is build to force use of GLX over EGL. This is upsteram GLEW issue nigels-com/glew#172. This PR ignores the GLEW_ERROR_NO_GLX_DISPLAY error. This is the temporary fix until this is fixed by GLEW upstream. @Pointedstick asked me to open PR here.
This is updated version of #5217 with fix of CI failures and C++11 compatibility.

## Testing Done
Tried playing game on my linux laptop with sway. No crashes or other problems. 

## Save File
No need for save file, bug occurs just when you try to start the game on wayland. It occurs on both v0.9.12 which is in the flatpak and on v0.9.6 which is in my distro repository (Gentoo), and does not occur when using this branch's build.

### DIFF
--- a/source/GameWindow.cpp
+++ b/source/GameWindow.cpp
@@ -128,10 +128,11 @@ bool GameWindow::Init()
 	glewExperimental = GL_TRUE;
 	GLenum err = glewInit(); 
 #ifdef GLEW_ERROR_NO_GLX_DISPLAY
-	if(GLEW_OK != err && err != GLEW_ERROR_NO_GLX_DISPLAY){
+	if(err != GLEW_OK && err != GLEW_ERROR_NO_GLX_DISPLAY)
 #else
-	if(GLEW_OK != err){
+	if(err != GLEW_OK)
 #endif
+	{
 		ExitWithError("Unable to initialize GLEW!");
 		return false;
 	}
@@ -372,5 +373,4 @@ void GameWindow::ExitWithError(const string& message)
 	
 	GameWindow::Quit();	
 }
-
 
--- a/source/GameWindow.cpp
+++ b/source/GameWindow.cpp
@@ -126,7 +126,12 @@ bool GameWindow::Init()
 	// Initialize GLEW.
 #ifndef __APPLE__
 	glewExperimental = GL_TRUE;
-	if(glewInit() != GLEW_OK){
+	GLenum err = glewInit(); 
+#ifdef GLEW_ERROR_NO_GLX_DISPLAY
+	if(GLEW_OK != err && err != GLEW_ERROR_NO_GLX_DISPLAY){
+#else
+	if(GLEW_OK != err){
+#endif
 		ExitWithError("Unable to initialize GLEW!");
 		return false;
 	}


source/GameWindow.cpp
```diff
@@ -126,7 +126,12 @@ bool GameWindow::Init()
 	// Initialize GLEW.
 #ifndef __APPLE__
 	glewExperimental = GL_TRUE;
-	if(glewInit() != GLEW_OK){
+	GLenum err = glewInit(); 
+#ifdef GLEW_ERROR_NO_GLX_DISPLAY
+	if(GLEW_OK != err && err != GLEW_ERROR_NO_GLX_DISPLAY){
+#else
+	if(GLEW_OK != err){
+#endif
```
Amazinite, line 134:
```suggestion
#ifdef GLEW_ERROR_NO_GLX_DISPLAY
	if(err != GLEW_OK && err != GLEW_ERROR_NO_GLX_DISPLAY)
#else
	if(err != GLEW_OK)
#endif
	{
```

tehhowch:
What's the timeframe for this issue to be resolved upstream? This seems like possibly a distro-specific patch that should be applied by the maintainers of the uploaded package, rather than committed here.

Furthermore, is there anything that `glewInit()` doesn't do in the cases where it returns this new error int, that it would have done if it had returned successfully? Proceeding with an unknown initialization state seems not great.

eternal-sorrow:
@tehhowch I don't have the answers to these questions.

MCOfficer:
> What's the timeframe for this issue to be resolved upstream? This seems like possibly a distro-specific patch that should be applied by the maintainers of the uploaded package, rather than committed here.

The PR in question (https://github.com/nigels-com/glew/issues/172) has seen zero activity for over a year, so it seems unlikely that it'll be merged soon. And even if it's merged tomorrow, it may take months for this change to arrive on all affected systems.

lanodan:
Hi, I managed to get endless-sky working on `libX11`-less wayland (`master` branch, commit 398e91fed27ab6b00f6bd68c357f2d23af3c0f44) with just this (ugly) change on your side:
```diff
--- a/SConstruct
+++ b/SConstruct
@@ -77,7 +77,7 @@ game_libs = [
        "SDL2",
        "png",
        "jpeg",
-       "GL",
+       "OpenGL",
        "GLEW",
        "openal",
        "pthread",
```

As for GLEW I managed to get it to compile with:
- Applying https://github.com/nigels-com/glew/pull/216/commits/584a266e71c3c3f7ec4a0e4e61c31b67f80a2419
- Running <code><kbd>sed -i 's/-lGL/-lOpenGL/' config/Makefile.linux-egl</kbd></code>
- Passing <code>POPT="${CFLAGS} <kbd>-D__native_client__</kbd>"</code> and <code><kbd>SYSTEM=linux-egl</kbd></code> to make options, for both build and install

(And I have ``-DEGL_NO_X11`` in my cflags, see https://hacktivis.me/notes/pure-wayland for related informations)

Pointedstick:
@tehhowch in answer to your original question, the open upstream PR is milestoned to 2.2.1, but it's been there for over a year. I don't think it's reasonable to expect the upstream fix to be merged anytime soon, so IMO a workaround in Endless Sky is appropriate.

BTW I'm unable to run Endless Sky on Fedora 36's Wayland session without this patch.

quyykk:
The workaround is not ideal (ignoring errors never is), but I think it is safe to merge this. If there is no GLX display then some other part of our initialization code (most likely the SDL initialization) will have already failed. I don't really see how we would reach this code except when on Wayland.

quyykk:
> BTW I'm unable to run Endless Sky on Fedora 36's Wayland session without this patch.

Wait so does that mean that ES doesn't run at all under Wayland? That would be unfortunate.

Pointedstick:
That is what it means, yes. Which is why I'm interested in seeing this approved and merged.

quyykk:
Running the game through the AppImage seems to work on my Fedora 36 VM (which if I understand correctly uses Wayland by default). So that's interesting. I checked and it uses XWayland of course.

This seems to be the more or less official workaround for this issue and from what I can tell it is also safe to do. Having proper Wayland support would definitely be interesting though, although we'd have to still use GLX for Steam, because it doesn't look like Steam will support Wayland anytime soon (neither for the client nor for the games).

Amazinite:
This has been replaced by #7027.

PR #5239, opened 2020-08-19
Rename Defend to Retreat in boarding panel

Terin:
## Feature Details
This PR renames Defend and associated labels, to Retreat for the boarding. After dozens of hours for my first playthrough, I never once clicked the "Defend" button when boarding ships, and the name perplexed me. It gave the impression that boarding was more of a symmetrical mini game, and that, somehow, this panel might appear if an enemy boards me. However, "Defending" is only something done when you are very in over your head, or are familiar with the game and are trying to get lucky. Normal players are given the odds, and would rarely attempt an attack where a retreat would be unsuccessful, as a situation where defending/retreating is unlikely would have attacking be very unlikely.

Renaming it to "Retreat" would make it much more in-line with the experience of most players, and make it less confusing. If also accepted, I'd be willing to update variable/function/comment names, but for now, I just went with the player-side text.

## UI Screenshots
retreat

### DIFF
--- a/data/interfaces.txt
+++ b/data/interfaces.txt
@@ -645,7 +645,7 @@ interface "boarding"
 		from 330 -22
 		align right
 	
-	label "survival odds (defending):"
+	label "survival odds (retreating):"
 		from 50 8
 		align left
 	string "defense odds"
@@ -679,7 +679,7 @@ interface "boarding"
 		dimensions 80 30
 	
 	active if "can defend"
-	button d "_Defend"
+	button d "_Retreat"
 		center 210 185
 		dimensions 80 30
 
--- a/source/BoardingPanel.cpp
+++ b/source/BoardingPanel.cpp
@@ -304,7 +304,7 @@ bool BoardingPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command,
 		}
 		isCapturing = true;
 		messages.push_back("The airlock blasts open. Combat has begun!");
-		messages.push_back("(It will end if you both choose to \"defend.\")");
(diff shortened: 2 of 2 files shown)


Zitchas:
I know there has been talk about re-doing the boarding combat that would make better use of "defend" (including having pirates board you, and such like).

But until that happens, yes, I think having it as "retreat" makes a lot more sense and would be more understandable for new players.

tehhowch:
Retreating is not the same as defending. If you are attacked while retreating, I would expect you to take heavier losses than if you are defending or attacking.

I don't think this is a change that makes sense to implement for that reason alone. Then as mentioned, future changes to the game will make having defense significantly more important for looting.

Terin:
If defending was a thing that actually happened more often, I might agree with that. But as it stands, it's such a rare occurrence, only coming up if you try to capture a warship with a sparrow, or a world ship on the few times they're a mission target, that the scale of the numbers aren't the critical part, it's the effect on the player. Pressing the retreat button and seeing it suddenly not go through feels more right with the current system than the "Defend" button that, 99% of the time, is not about defense. 

And even then, combat on spaceships feels like corridors and rooms crew would get pinned down in, not open fields that people would get gunned down at while retreating, so the scale still works for me.

As for future updates, this is the first I've heard of it, and I'm not sure how far those are out. But even then, wouldn't that likely come with a UI design for the mini game? All the current files would get shaken up as is, so string adjustments now shouldn't affect overhauls that are months/years off.

tux2603:
Having the button labeled "Retreat" makes it sound more like it'll cancel the attack/boarding procedure to me, so it could possibly be confusing if you press retreat and you don't actually exit the panel

Issue #5017, opened 2020-04-25
Add on-demand automatically generated conditions


petervdmeer:
**Is your feature request related to a problem? Please describe.**
All player-conditions are currently stored in the conditions-map in PlayerInfo and they are updated on pre-defined times.
This works fine for conditions that are explicitly set from missions (since they are set directly in the map the moment they are updated), but it causes some lag and limitations for conditions that are automatically generated (since they need to be generated on planet-landings and on entering-systems before they can be used):
- The lag is because automated conditions only get updated on pre-defined times (for example on entering a system or on landing on a planet).
- The limitation is because all conditions need to be generated up-front (and get saved to active savegames), so having a lot of automatic conditions would cause quite some overhead, especially for conditions that are rarely used and compute-intensive to generate.

**Related Issue Links**
- This comment seems very related: https://github.com/endless-sky/endless-sky/issues/3297#issuecomment-348696133
- The frequency/locations where mission checks are done also affects the lag until some mission starts or gets triggered, but that is not discussed or addressed in this request.
- The auto-testers #4308 and #4370 rely heavily on conditions for asserts and wait-loops (and rely on those being up-to-date without lag).
- #3735 (and #5412) is a specific example of conditions that can be made available through the mechanism discussed here.
- #4915 implemented an example of conditions that are automatically generated.
- #4917 added an additional conditions-update moment to reduce some lag.
- #4798 also needs conditions updated on-the-fly (and it will also need a notion/reference of "local ship" for conditions that are based on the ship for which the attribute applies)
- Mission writers might want to use new automatically generated conditions once they become possible/available.

**Describe the solution you'd like**
This feature-request is about changing automatically generated conditions from generated up-front to generated on-demand.
Effects on the ES source code is that all locations where conditions are used need to retrieve conditions in a different way; they need to retrieve required conditions by name instead of requesting the full map of conditions.

Some conditions will be read-only (like the in-game-day, in-game-month and in-game-year), but other conditions like `"unpaid maintenance"` might also be provided readable and writeable. A class like Accounts might just provide conditions for many strings that start with `"unpaid "`.

Conditions could be provided also for information currently contained in classes like Armament, CargoHold, Engine, Government, Planet, Preferences and Ship by choosing a proper prefix for each piece of information and then implementing the proper  interface on those classes.

**Describe alternatives you've considered**
Performing the conditions-generation more often; this would reduce the lag, but could increase the computational overhead (especially if we also add more auto-generated conditions).
[...]

Issue #4798, opened 2020-02-08
Conditional Attributes


Pathless-Visions:
**Is your feature request related to a problem? Please describe.**
Not problem related but it would be neat to have the ability to make attributes conditional or non-conditional through the data files.

Currently attributes work in a way that makes it so they trigger only if a related attribute is active or a condition is met and they can't be overridden in the data files.
For example:
`"shield generation"` and `"shield energy"`
`"active cooling"` and `"cooling energy"`

**Related Issue Links**

https://github.com/endless-sky/endless-sky/pull/4520
https://github.com/endless-sky/endless-sky/pull/4522
https://github.com/endless-sky/endless-sky/pull/4523
https://github.com/endless-sky/endless-sky/pull/4524

These are some examples of PRs and issues that bring up creating **explicit** attributes that only trigger under specific conditions, **BUT** this is not what this issue is meant to address. Instead this issue addresses **allowing conditions to be created in the data (text) files**.

**Describe the solution you'd like**
Allow attributes to have or not have conditions tweaked in data files.
For example:
```
"shield generation" 10
    "active"
        "hull repair" < 1
    "never active"
        "heat generation" > 20
"shield energy" 10
"shield heat" 10
```
would make it so `"shield generation"` only activates if `"hull repair"` is greater than `1` and never activate if the ship's current `"heat generation"` exceeds `20`. This would allow creating outfits with more strategic uses. 

I would also encourage the creation of new condition names explicitly for certain conditions.
For example:
```
"hull repair rate" 10
    "active"
        and
            "current energy" > 1000
            "ship category" "Heavy Warship"
    "never active"
        "current hull" > 2500
"hull energy" 10
"hull heat" 10
```
would make it so that the `"hull repair rate"` attribute only activates if the ship has `1000` energy at its disposal and is in the `"Heavy Warship"` category, and never activate if the ship's current hull exceeds `2500`.

Some more examples for condition names:
  * `"target category" <category>` applies an outcome if a target meets a certain category
  * `"target distance" <category>` applies an outcome if a target is at a certain distance
  * `"current heat" <amount>` applies an outcome if the ship's heat has met a certain threshold
  * `"current thrust" <amount>` applies an outcome if the ship's thrust has met a certain threshold
  * there are probably more that I haven't thought of but could be useful.

You could also make it so conditionals could apply to more than one attribute as a whole as well
For example:
```
outfit "Stabilizer"
    category "Systems"
    "conditional"
        "mass" + 100
[...]

PR #5244, opened 2020-08-20
feat(conditions): Implement on-demand auto-generated conditions.

petervdmeer:
**Feature:** This PR implements the feature request detailed and discussed in issue #5017

## Feature Details
This PR changes the game-conditions (as stored under PlayerInfo, and as used in mission logic) from a `(string)key->(int64_t)value` map to an interface where key->value pairs can still be set, stored and queried, but where it is also possible to register "derived" conditions for querying (and setting) data-values that are stored outside of the PlayerInfo conditions.

A major benefit of this change is that information currently unavailable to content creators (mission writers) can be made available (as int64_t values) by just registering the desired information as derived conditions. Another benefit (as mentioned in issue #5017) is that generated conditions are always "up-to-date", in contrast to being generated and synchronized at predefined moments.

Derived conditions come in 2 flavors; exactly matching single derived conditions (like `flagship crew` or `armament deterrence`) and prefix-matching derived conditions (like `ships: <category>` and `reputation: <government>`).

Derived conditions are not iteratable at the conditions-level, expected usage is that if iteration is needed for data in such derived conditions, that the iteration itself happens on the actual class that contains the data (the players ships-list for `ships: <category>` conditions, and the set of governments for `reputation: <government>` conditions.

All generated conditions are changed to be derived conditions in this PR.

## UI Screenshots
N/A

## Usage Examples
Conditions like `flagship crew`, `armament deterrence`, `ships: Fighter`, as described on the [Wiki](https://github.com/endless-sky/endless-sky/wiki/CreatingMissions#conditions), can just be used by content creators (mission writers) as they were used before, but those conditions are internally no longer stored as conditions, but will be derived from the active flagship and the active players fleet.

## Testing Done
Added relevant automatic unit tests for the following:
- Accounts (test moved to https://github.com/endless-sky/endless-sky/pull/6139)
- ConditionSet
- ConditionsStore

Performed the following tests/checks (and found and fixed 2 bugs with them):
- BankPanel: Checked "Crew Salaries" before and after this change.
- BankPanel: Checked "Salary and Tribute Income" before and after this change.
- OutfitterPanel: Checked in outfitter that licenses were visible.
- Mission: Tested takeoff and land in an active game. Checked job-board.
- Mission: Tested taking a mission and update of offered and done counters.
- Planet: Demanded tribute and checked in savegame that it was set.
- PlayerInfo: Tested by the other tests done.
- PlayerInfoPanel: Checked if the salary/tribute/licenses fields were ok.
- ShipyardPanel: Tested buying a Bactrian (with license: OK).
- ShipyardPanel: Tested buying a Bactrian (without license: OK, blocked).

## Performance Impact
Some processing time will be saved by not always generating all possible automatic conditions at predefined moments. Requesting the same derived conditions multiple times could however be less efficient, since it would be calculated each time it is requested (instead of at the predefined generation moments). We might want to add some caching outside the conditions code to compensate for this if this becomes a problem.


source/PlayerInfo.cpp
```diff
@@ -1721,6 +1721,56 @@ int64_t PlayerInfo::GetCondition(const string &name) const
...
+bool PlayerInfo::AddCondition(const string &name, int64_t value)
+{
+	conditions[name] += value;
+	return true;
+}
+
+
+
+// Remove the value for some condition.
+bool PlayerInfo::EraseCondition(const std::string &name)
```
Amazinite, line 1766: This file is already `using namespace std`.
```suggestion
bool PlayerInfo::EraseCondition(const string &name)
```

source/PlayerInfo.cpp
```diff
@@ -1721,6 +1721,56 @@ int64_t PlayerInfo::GetCondition(const string &name) const
...
+{
+	auto it = conditions.lower_bound(prefix);
+	for( ; it != conditions.end() && !it->first.compare(0, prefix.length(), prefix); ++it)
+		targetMap[it->first] = it->second;
+}
+
+
+
+// Get the sum (addition) of all conditions starting with the given prefix.
+int64_t PlayerInfo::GetConditionSum(const std::string &prefix) const
```
Amazinite, line 1735:
```suggestion
int64_t PlayerInfo::GetConditionSum(const string &prefix) const
```

source/PlayerInfo.cpp
```diff
@@ -1721,6 +1721,56 @@ int64_t PlayerInfo::GetCondition(const string &name) const
 
 
 
+// Copy conditions matching the prefix into the given map.
+void PlayerInfo::GetConditions(std::map<std::string, int64_t> &targetMap, const std::string &prefix) const
```
Amazinite, line 1725:
```suggestion
void PlayerInfo::GetConditions(map<string, int64_t> &targetMap, const string &prefix) const
```

source/PlayerInfo.h
```diff
@@ -180,6 +180,17 @@ class PlayerInfo {
...
+	// Copy conditions matching the prefix into the given map.
+	void GetConditions(std::map<std::string, int64_t> &targetMap, const std::string &prefix) const;
+	// Retrieve a sum (addition) of all "condition" flags starting with the
+	// given prefix.
+	int64_t GetConditionSum(const std::string &prefix) const;
+	// Set a "condition" flag to the given value. Returns true on success,
+	// false on failure.
+	bool SetCondition(const std::string &name, int64_t value);
+	bool AddCondition(const std::string &name, int64_t value);
+	bool EraseCondition(const std::string &name);
```
Amazinite, line 192: I'm guessing having a return type is setting up for later.

source/PlayerInfo.h
```diff
@@ -180,6 +180,17 @@ class PlayerInfo {
 	
 	// Access the "condition" flags for this player.
 	int64_t GetCondition(const std::string &name) const;
+	// Copy conditions matching the prefix into the given map.
+	void GetConditions(std::map<std::string, int64_t> &targetMap, const std::string &prefix) const;
```
Amazinite, line 184: Why modify a map handed to the function instead of returning a map? Are we expecting to have a situation where we want to add to an existing map in a function calling GetConditions?

source/PlayerInfo.h
petervdmeer: Yes, that is correct, this is for when we get read-only conditions.

source/PlayerInfo.h
petervdmeer: Yes, I expect some situations in the future where the map gets filled by recursive calls. Those calls can be implemented by a map that is returned, which will be more readable, or with the current implementation, which should have a somewhat better performance.

The code here is not performance critical, so if we prefer readability, then I can change this to a function that returns a map.

Amazinite: I'll leave it up to tehhowch. I think it's readable enough both ways, but having the function return a map is probably more readable than the current method of modifying a map.

tehhowch:
Mind putting together a document / some diagrams that show where this is heading? There may be better ways to go about this, such as defining an interface that gets returned by the call to `Conditions()` and having several classes implement that interface, depending on what is supposed to be happening.

petervdmeer:
> Mind putting together a document / some diagrams that show where this is heading? There may be better ways to go about this, such as defining an interface that gets returned by the call to `Conditions()` and having several classes implement that interface, depending on what is supposed to be happening.

I just uploaded branch https://github.com/petervdmeer/endless-sky/tree/conditions_provider. That branch is not ready for review/merging yet (it even contains 2 bugs that I already solved in the PR here), but it shows the next step that I want to take by having PlayerInfo implement the new ConditionsProvider interface.

The step after this is to have the automatically generated information (from PlayerInfo::UpdateAutoconditions()) calculated during a call to GetCondition() instead of generated at fixed moments using the UpdateAutoconditions() call. Some of those variables will be read-only, but I can imagine that we make a variable like "unpaid maintenance" also automatically updatable from missions.

If we want to keep the logic for such "unpaid maintenance" in the Accounts class, then we could implement the ConditionsProvider interface on accounts and have PlayerInfo call accounts for all condition-accesses that start with the prefix `"unpaid "`.

The steps after that is to also unlock information in classes like Armament, CargoHold, Engine, Government, Planet, Preferences and Ship for missions by choosing a proper prefix for each piece of information and then implementing the ConditionsProvider interface on those classes.

Amazinite (dismissed):

petervdmeer:
> There may be better ways to go about this, such as defining an interface that gets returned by the call to `Conditions()` and having several classes implement that interface, depending on what is supposed to be happening.

When we consider issue #4798 related (as discussed on #5017), then having an object returned that implements the interface seems a better way to go about this. Since returning an object allows for context-specific interfaces.

petervdmeer:
Converted to draft; needs Iterator implementation as being added in https://github.com/petervdmeer/endless-sky/tree/conditions_provider before this will be "ready for merge" again.

source/PlayerInfo.h
petervdmeer: Will be replaced by iterator (maybe filtered iterator) as per discussion with tehhowch.

petervdmeer:
Returned to "ready for review" state. PlayerInfo::GetCondition already existed, this PR now only adds PlayerInfo::SetCondition, PlayerInfo::EraseCondition and PlayerInfo::GetConditions. I can take PlayerInfo::GetConditions also out of this PR if it is not good.

petervdmeer:
Rebased to resolve merge-conflict.

petervdmeer:
I updated this PR with the work to make all accesses indirect and I also added the interfaces for the on-demand conditions to show where this PR is heading to.
Review comments are welcome. Some re-testing needs to be done before we should consider merging this. I hope that the test-framework (#5620 and a few follow-ups) will be done before this PR is ready so that I can create some auto-testers instead of doing manual testing again.

source/PlayerInfo.h
```diff
@@ -189,8 +190,16 @@ class PlayerInfo {
...
+	// Set a "condition" flag to the given value. Returns true on success,
+	// false on failure.
+	bool SetCondition(const std::string &name, int64_t value);
+	bool AddCondition(const std::string &name, int64_t value);
+	bool EraseCondition(const std::string &name);
+	void EraseManualByPrefix(const std::string &prefix);
+	// Direct access to "condition" flags data.
+	ConditionsStore &Conditions();
+	const ConditionsStore &Conditions() const;
+	const std::map<std::string, int64_t> &GetManualConditions() const;
```
Amazinite, line 202: Just by reading the names, I'm not exactly sure what "manual" is suppose to mean in these method names. Would "raw" be a better term to use? "GetRawConditions" and "EraseRawByPrefix"? Or maybe "direct" for "GetDirectConditions" etc.

source/PlayerInfo.cpp
```diff
@@ -1782,36 +1783,88 @@ void PlayerInfo::HandleEvent(const ShipEvent &event, UI *ui)
...
+
+
+// Temporary function to erase conditions by prefix. This function should
+// no longer be needed once automatically generated conditions are on-demand
+// automatically generated.
+// This function is performance-wise also not very efficient, but that
+// might be okay for a temporary function.
+void PlayerInfo::EraseManualByPrefix(const string &prefix)
+{
+	set<string> toErase;
```
Amazinite, line 1822: Why can't we erase as we go? Wouldn't you just need to walk back the iterator when something matches and gets erased.

source/ConditionsStore.h
```diff
@@ -0,0 +1,60 @@
...
+// it's internal storage format).
+class ConditionsStore {
+public:
+	// Constructors to initialize this class.
+	ConditionsStore();
+	ConditionsStore(std::initializer_list<std::pair<std::string, int64_t>> initialConditions);
+	ConditionsStore(const std::map<std::string, int64_t> initialConditions);
+
+	// Retrieve a "condition" flag from this provider.
+	int64_t operator [] (const std::string &name) const;
```
Amazinite, line 39: Other operator functions don't have spaces around the operator.
```suggestion
	int64_t operator[](const std::string &name) const;
```

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,131 @@
...
+// Constructor where a number of initial manually-set values are set.
+ConditionsStore::ConditionsStore(const map<string, int64_t> initialConditions)
+{
+	for(auto it : initialConditions)
+		SetCondition(it.first, it.second);
+}
+
+
+
+int64_t ConditionsStore::operator [] (const std::string &name) const
```
Amazinite, line 44:
```suggestion
int64_t ConditionsStore::operator[](const std::string &name) const
```

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,131 @@
...
+	return GetCondition(name);
+}
+
+
+
+bool ConditionsStore::HasCondition(const std::string &name) const
+{
+	// When we add support for on-demand conditions, then we can
+	// lookup the on-demand condition here before searching the
+	// manually set conditions.
```
Amazinite, line 55: Looking forward to this. :)

Amazinite (commented):
Looking good so far. I can see the direction that this is headed. Looking forward to what that yields.

source/ConditionsStore.cpp
petervdmeer: Yes, the end-goal of what this PR is preparing for. :-)

source/PlayerInfo.cpp
petervdmeer: The list of manual conditions is only provided as `const` (meaning read-only). I could make it writeable as well, but I'm afraid that that could lead to bugs in the future if developers don't notice that they are working with a subset of the conditions.

source/PlayerInfo.h
petervdmeer: This is the set of conditions that is not On-Demand. I guess `GetNotOnDemandConditions` would also be an appropriate name for this function, but it sounds a bit odd as well.
`Raw` doesn't sound any more clear than `Manual` to me. `Direct` however also okay for me.

petervdmeer: We could also consider if `On Demand` is the best name for conditions that are based on other data, maybe we should rename those to `Derived Conditions`, in which case we can use `GetNonDerivedConditions` as name here (and add a comment that non-derived conditions are stored in a savegame)

Amazinite: Okay, I see where the term comes from. Some conditions are “automatic” (maintained by the source code and writing to them just results in that being overwritten), meaning those conditions which aren’t automatic are manual. 

Are you expecting to have GetAutomaticConditions or similar? If not then maybe it’d be better to just have this be GetConditionsSubset. That way the method name makes it clear that these aren’t all the conditions (or rather, won’t be all the conditions), and then a comment can explain exactly what that subset consists of.

petervdmeer: Let's go for `Derived Conditions` for conditions that are based on other parts of the code and let's use `Primary Conditions` for conditions that are set directly not based on any data in any other class. I'll update the code for this.

The function-name would then be `GetPrimaryConditionsSubset()`.

I don't see a lot of value in getting a list of Derived conditions. The classes that those conditions derive from should already have their own functions to get the relevant data out of those classes.

petervdmeer:
Just updated the ConditionsStore to lookup direct condtions and condition providers in a single to. (This should provide better performance for conditions itself, but it also makes the code is a little bit more complex and it makes accesses to the direct conditions map slower because that map needs to be built on-demand.)

Needs a little bit of re-testing.

petervdmeer:
Merged master branch and resolved merge conflicts.

quyykk (commented):
I don't know what you exactly have in mind for more complex conditions, but I'd like to propose another interface for `ConditionsProvider`.

It seems like every class deriving from `ConditionsProvider` is gonna overwrite the virtual functions in the same exact way but with different values. So, I'd propose having `ConditionsProvider` have 2 virtual functions, maybe named `GetConditionCallbacks` and `SetConditionCallbacks` or something. And what they do is return a map from strings (the condition strings) to a callback. So for `Account` the map for the getter would look like this:

- `"credit"` -> `[this] { return Credits(); }`
- `"net worth"` -> `[this] { return min(limit, max(-limit, NetWorth())); }`

And then `ConditionsProvider` defines all of the functions that `Account` currently overrides.

source/ConditionsStore.h
```diff
@@ -0,0 +1,93 @@
...
+	int64_t GetCondition(const std::string &name) const;
+	bool HasCondition(const std::string &name) const;
+	bool AddCondition(const std::string &name, int64_t value);
+	// Add a value to a condition, set a value for a condition or erase a
+	// condition completely. Returns true on success, false on failure.
+	bool SetCondition(const std::string &name, int64_t value);
+	bool EraseCondition(const std::string &name);
+	
+	// Direct (read-only) access to the stored primary conditions.
+	const std::map<std::string, int64_t> GetPrimaryConditions() const;
```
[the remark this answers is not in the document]
quyykk: Good idea! Yeah that will work.

source/Account.cpp
```diff
@@ -25,6 +25,19 @@ namespace {
 	// For tracking the player's average income, store daily net worth over this
 	// number of days.
 	const unsigned HISTORY = 100;
+	
+	// Enumaration to minimize the risk of typos compared to using string identifiers.
+	enum class AccountCondition {NET_WORTH, CREDITS, UNPAID_MORTGAGES, UNPAID_FINES, UNPAID_SALARIES, UNPAID_MAINTENANCE, CREDIT_SCORE};
+	
+	const map<AccountCondition, const string> CONDITION_TO_TEXT = {
```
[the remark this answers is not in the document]
petervdmeer: Good ideas, I also considered using static strings in an earlier commit to be able to identify the strings (the enum is not really needed, but I didn't want to copy-paste the strings around either).

source/ConditionsProvider.cpp
```diff
@@ -0,0 +1,53 @@
...
+bool ConditionsProvider::HasCondition(const std::string &name) const
+{
+	return HasConditionImpl(name);
+}
+
+
+
+// Set a value for a condition. Calls the implementation as provided
+// by the class implementing this interface. Implementing classes are
+// expected to return false and change nothing for read-only conditions.
```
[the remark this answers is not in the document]
petervdmeer: Yes, that is possible, but during an earlier review we considered such defaults vs the benefits of having the need for providers to explicitly specify the desired behavior. If I remember correctly then in that review-session we considered the explicit specification to be more valuable than the reduction in functions to implement. (Providers choose explicitly for read-only, not by accidentally forgetting.)

petervdmeer:
> I don't know what you exactly have in mind for more complex conditions, but I'd like to propose another interface for `ConditionsProvider`.
> 
> It seems like every class deriving from `ConditionsProvider` is gonna overwrite the virtual functions in the same exact way but with different values. So, I'd propose having `ConditionsProvider` have 2 virtual functions, maybe named `GetConditionCallbacks` and `SetConditionCallbacks` or something. And what they do is return a map from strings (the condition strings) to a callback. So for `Account` the map for the getter would look like this:
> 
>     * `"credit"` -> `[this] { return Credits(); }`
>     * `"net worth"` -> `[this] { return min(limit, max(-limit, NetWorth())); }`
> 
> 
> And then `ConditionsProvider` defines all of the functions that `Account` currently overrides.

Using lambda functions/callback functions would indeed be very elegant. I did some tests with them in branch https://github.com/petervdmeer/endless-sky/commits/conditions_provider. Storing the lambdas/callbacks is however a bit tricky in C++11 and I didn't get it fully working yet.

quyykk:
> Storing the lambdas/callbacks is however a bit tricky in C++11 and I didn't get it fully working yet.

How so? `std::function`? Since it the lambdas would only capture the `this` pointer, it should be well within the real of SOO thus incurring minimal impact on performance compared to a regular function pointer :)

petervdmeer:
> > Storing the lambdas/callbacks is however a bit tricky in C++11 and I didn't get it fully working yet.
> 
> How so? `std::function`? Since it the lambdas would only capture the `this` pointer, it should be well within the real of SOO thus incurring minimal impact on performance compared to a regular function pointer :)

Ah, looks like I missed the difference between `std::function` and function pointers in my other branch.
If it is indeed possible to convert lambdas (with capture sections) to `std::function` and then store them in a map or vector, then that could also lead to better performance (since the ConditionsStore could lookup a relevant lambda directly in the internal map and execute it without further lockups, the current implementation has a second lookup at Account.h which would disappear).
This is probably worth looking into (but I'm quite busy this weekend, so it might be for later in August).

petervdmeer:
Converted to draft; branch https://github.com/petervdmeer/endless-sky/tree/conditions_provider now contains a (mostly working) implementation where derived-conditions are retrieved using lambda functions (I finally got that working thanks to quyykk s remark on the use of `std::function`). I would like to transfer that implementation into this PR since I expect that implementation to work much better than the inheritance currently done with ConditionsProvider.

source/Account.cpp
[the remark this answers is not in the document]
petervdmeer: The enum and the map have been removed fully due to the switch to lambda based condition providers, so this review comment is also handled.

source/ConditionsStore.h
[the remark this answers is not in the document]
petervdmeer: I uploaded a first version of the iterator. It still needs some work, but the basics seem to be working.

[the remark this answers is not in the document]
petervdmeer: This method has been removed; all code now uses the iterator instead.

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,276 @@
...
+	if(!(name.compare(0, it->first.length(), it->first)) &&
+			(it->second.type == PREFIX_PROVIDER || it->first.length() == name.length()))
+		return &(it->second);
+	
+	return nullptr;
+}
+
+
+
+const ConditionsStore::ConditionEntry *ConditionsStore::GetEntryConst(const string &name) const
```
quyykk, line 258: Why not just `GetEntry`? Why the extra `Const` in the name?

source/ConditionsStore.h
```diff
@@ -0,0 +1,152 @@
...
+	// Retrieve a condition entry based on a condition name, doCreate indicate if the entry
+	// should be created if it wasn't there yet.
+	ConditionEntry *GetEntry(const std::string &name);
+	const ConditionEntry *GetEntryConst(const std::string &name) const;
+
+
+
+private:
+	// Storage for both the primary conditions as well as the providers.
+	std::map<std::string, ConditionEntry> storage;
```
quyykk, line 147: Quick question: Have you considered having 3 maps instead of one? It might simplify lots of code (like the iterator not being needed anymore). We'll need to do 3 map lookups in the worst case to get a condition though but map lookups aren't expensive. I haven't fully thought it out so just a thought.

source/ConditionsStore.h
```diff
@@ -0,0 +1,152 @@
...
+	void SetProviderPrefixed(const std::string &prefix, DerivedProvider conditionsProvider);
+	void SetProviderNamed(const std::string &name, DerivedProvider conditionProvider);
+	
+	// Helper to completely remove all data and linked condition-providers from the store.
+	void Clear();
+
+
+private:
+	// Retrieve a condition entry based on a condition name, doCreate indicate if the entry
+	// should be created if it wasn't there yet.
```
quyykk, line 139: Wahts `doCreate` that this comment alludes to?

source/ConditionsStore.h
```diff
@@ -0,0 +1,152 @@
...
+	// Structure that describes the interface for DerivedProviders. When
+	// registering a new provider, then the (lambda)functions for accessing
+	// the conditions in the derived provider are communicated using this
+	// structure.
+	struct DerivedProvider {
+		std::function<int64_t(const std::string&)> getFun;
+		std::function<bool(const std::string&)> hasFun;
+		std::function<bool(const std::string&, int64_t)> setFun;
+		std::function<bool(const std::string&)> eraseFun;
+	};
```
quyykk, line 52: What would be a condition where `hasFun` and `eraseFun` would do something different than the default implementation? Especially `eraseFun`. Can't really delete a auto generated condition right?

source/ConditionsStore.h
petervdmeer: Yes, earlier implementations had 2 or 3 separate maps, but tehhowch was concerned about the performance when we would need to do all those separate lookups, so I switched to this (more complex) implementation that gets the relevant condition(provider) in a single lookup.

source/ConditionsStore.h
petervdmeer: An artifact of some old code (where the has-function should never create new entries, while the setFun should). I think I can completely remove this comment now.

source/ConditionsStore.cpp
petervdmeer: Agreed, the extra Const in the name is probably not needed, I'll check if I can remove it.

source/ConditionsStore.h
petervdmeer: I expect nearly all auto-generated conditions to be read-only, but suppose that we make mortgages a read-write auto-generated condition, then erasing the condition that refers to a specific mortgage could remove that specific mortgage. (Not a very good example, because I expect the mortgages to be provided read-only if we make them available through conditions.)

source/ConditionsStore.cpp
quyykk: Also, since the 2 functions are the same I'd consider using the `const_cast` hack if it doesn't violate the style guide :)

source/ConditionsStore.h
petervdmeer: The [tributes](https://github.com/endless-sky/endless-sky/issues/5448) might be a better example.

source/ConditionsStore.h
quyykk: I guess he has a point. Although IMO: The majority of condition access are for value conditions right? Sooo you'd only have multiple lookups for anything other than a value condition....... :)

petervdmeer: Derived conditions take precedence over value conditions, so you would have 2 look-ups for value conditions:
-1- Check if a derived condition exists with the given name or prefix.
-2- Get the relevant value condition.

I guess we could do something smart to actively prune derived conditions out of the value conditions, but even then we would need 2 lookups for an unset value condition:
-1- Check if the value exists in the list of value conditions.
-2- If it is not set, then check if a derived condition exists.

petervdmeer:
All automatic conditions are now implemented as derived conditions, but not all of them were tested yet. (Some are covered by automatic tests, but not all of them yet.)

Amazinite (changes requested):
I will admit that some of the stuff in ConditionStore is definitely outside of my knowledge base, but from what I can tell everything looks good and checks out. My only comments are on a few minor errors.

I've also opened a PR to this branch with some style enforcement changes: https://github.com/petervdmeer/endless-sky/pull/8

source/PlayerInfo.cpp
```diff
@@ -2520,76 +2480,157 @@ void PlayerInfo::ValidateLoad()
...
-		if(flagship->GetSystem())
-			conditions["flagship system: " + flagship->GetSystem()->Name()] = 1;
-		if(flagship->GetPlanet())
-			conditions["flagship planet: " + flagship->GetPlanet()->TrueName()] = 1;
-	}
-	else
+		int64_t retVal = 0;
+		for(const shared_ptr<Ship> &ship : ships)
+			if(!ship->IsParked() && !ship->IsDisabled() && ship->GetSystem() == system
+					&& name == "ship: " + ship->Attributes().Category())
```
petervdmeer, line 2584:
```suggestion
					&& name == "ships: " + ship->Attributes().Category())
```

source/PlayerInfo.cpp
```diff
@@ -2520,76 +2480,149 @@ void PlayerInfo::ValidateLoad()
...
+	// (e.g. Heavy Warships).
+	conditionsProvider.getFun = [this] (const string &name)->int64_t {
+		int64_t retVal = 0;
+		for(const shared_ptr<Ship> &ship : ships)
+			if(!ship->IsParked() && !ship->IsDisabled() && ship->GetSystem() == system
+					&& name == "ship: " + ship->Attributes().Category())
+				++retVal;
+		return retVal;
+	};
+	conditions.SetProviderPrefixed("ships: ", conditionsProvider);
```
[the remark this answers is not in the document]
petervdmeer: No, that was a bug. Should be `"ships: "` on both values => changed this.

source/PlayerInfo.cpp
```diff
@@ -2520,76 +2480,149 @@ void PlayerInfo::ValidateLoad()
...
 	// in the player's entire fleet. The only fleet parameter offered to a
 	// boarding mission is the fleet composition (e.g. 4 Heavy Warships).
-	if(isBoarding && flagship)
-	{
-		conditions["cargo space"] = flagship->Cargo().Free();
-		conditions["passenger space"] = flagship->Cargo().BunksFree();
-	}
+	conditionsProvider.getFun = [this] (const string &name)->int64_t {
+		if(flagship && !boardingMissions.empty())
+			flagship->Cargo().Free();
```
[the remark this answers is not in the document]
petervdmeer: Yes, the return was missing.

Amazinite (approved):

petervdmeer:
Rebased to solve merge-conflict.

petervdmeer:

petervdmeer:
Converted to draft, want to try to remove the PrimariesInterator completely:
- Moving the saving (serialization) to ConditionsStore will remove the iterator for saving.
- Allowing providers to save through ConditionsStore (through a callback) will also allow splitting off iteraterable items (removing the need to centrally iterate).

Amazinite:
@petervdmeer Is there anything that I or anyone else could do to help move this along, or is it mainly just waiting on you refactoring it? Like #5440, this is one of those PRs that repeatedly gets referenced as something that could greatly improve current mission writing capabilities, so I'd like to get this finalized and merged soon.

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,456 @@
...
+	{
+		storage.erase(name);
+		return true;
+	}
+	return ce->provider->eraseFun(name);
+}
+
+
+
+ConditionsStore::ConditionEntry &ConditionsStore::operator[](const std::string &name)
```
Amazinite, line 349: Style change: the file is already using namespace std. There are various other instances of `std::` in this file, but those are on standard function calls, in which case we typically do prepend `std::` to those.
```suggestion
ConditionsStore::ConditionEntry &ConditionsStore::operator[](const string &name)
```

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,456 @@
...
+
+
+// Default constructor
+ConditionsStore::DerivedProvider::DerivedProvider(const string &name, bool isPrefixProvider): name(name), isPrefixProvider(isPrefixProvider)
+{
+}
+
+
+
+void ConditionsStore::DerivedProvider::SetGetFun(std::function<int64_t(const std::string&)> newGetFun)
```
Amazinite, line 29: Judging by other files, the style is to drop the `std::` from here too.
https://github.com/endless-sky/endless-sky/blob/9124602fe322f4ef92ec920a3ec89ca7bd49022a/source/Panel.cpp#L72
```suggestion
void ConditionsStore::DerivedProvider::SetGetFun(function<int64_t(const string&)> newGetFun)
```
There are three more functions below this one that would also need this change.

source/ConditionsStore.h
```diff
@@ -0,0 +1,197 @@
...
+		using pointer = const std::pair<const std::string, int64_t>*;
+		using reference = std::pair<const std::string, int64_t>;
+
+		// Default input_iterator operations.
+		std::pair<std::string, int64_t> operator*() const;
+		const std::pair<std::string, int64_t>* operator->();
+		PrimariesIterator& operator++();
+		PrimariesIterator operator++(int);
+		bool operator== (const PrimariesIterator& rhs) const;
+		bool operator!= (const PrimariesIterator& rhs) const;
```
Amazinite, line 122: Style change:
```suggestion
		const std::pair<std::string, int64_t> *operator->();
		PrimariesIterator &operator++();
		PrimariesIterator operator++(int);
		bool operator==(const PrimariesIterator &rhs) const;
		bool operator!=(const PrimariesIterator &rhs) const;
```

Amazinite (dismissed):
Reasserting my approval since it's been a while since my last review. I only have a few style-related comments, but outside of that—from what I can understand of the ConditionsStore—this looks like some great work.

source/ConditionSet.cpp
```diff
@@ -162,12 +163,12 @@ namespace {
...
-				if(temp != created.end())
-					value = temp->second;
-				else if(perm != conditions.end())
-					value = perm->second;
+				bool tempHas = created.Has(str);
+				bool permHas = conditions.Has(str);
+				if(tempHas)
+					value = created.Get(str);
+				else if(permHas)
+					value = conditions.Get(str);
```
Hurleveur, line 171: Is this an optimization? Idk how exactly has and find work, but we could be looking for it twice when it's there?

source/PlayerInfo.cpp
```diff
@@ -2564,76 +2519,151 @@ void PlayerInfo::ValidateLoad()
...
+	});
+	reputationProvider.SetGetFun([](const string &name) -> int64_t
+	{
+		string govName = name.substr(strlen("reputation: "));
+		auto gov = GameData::Governments().Get(govName);
+		if(!gov)
+			return 0;
+		return gov->Reputation();
+	});
+	reputationProvider.SetSetFun([](const string &name, int64_t value) -> bool
```
Hurleveur, line 2658: Just wondering, why is this the only one that has a setter?
reputation is the only primary condition, that is not generated by code and is written in the save file, right?

maybe it could be explained a comment

source/Test.cpp
```diff
@@ -560,13 +560,13 @@ void Test::Fail(const TestContext &context, const PlayerInfo &player, const stri
...
-	auto it = player.Conditions().lower_bound(TEST_PREFIX);
-	for( ; it != player.Conditions().end() && !it->first.compare(0, TEST_PREFIX.length(), TEST_PREFIX); ++it)
+	auto it = player.Conditions().PrimariesLowerBound(TEST_PREFIX);
+	for( ; it != player.Conditions().PrimariesEnd() && !it->first.compare(0, TEST_PREFIX.length(), TEST_PREFIX); ++it)
 		conditions += "Condition: \"" + it->first + "\" = " + to_string(it->second) + "\n";
 
 	if(!conditions.empty())
 		Files::LogError(conditions);
-	else if(player.Conditions().empty())
+	else if(player.Conditions().PrimariesBegin() == player.Conditions().PrimariesEnd())
```
Hurleveur, line 569: You already have an empty() function defined, does this not check for the same thing?
Btw does it only checks for primaries? Probably should be clear in the function's name.

Hurleveur (commented):
Is it a convention to call the functions Fun? (~~we having fun coding I guess~~)
If not maybe it would be clearer replacing that by Function, we are defining custom getters, setters etc after all.

Looks pretty solid, finally we no longer need to stock conditions as separate from the actual thing.
Indeed the tests working means its all good. Nice that you've added some more for this.
There wasn't much freedom of action in this; what I mean is that most things had to be changed this way. Still I tried to comment on the parts I think can be changed.
Code wise it's much more elegant this way.

samrocketman (commented):
I was able to do a style review.  I would like to do a "part 2" where I exercise the game logic.  However, I think there's value in submitting the style review first in case it is desirable for this to be merged before I get a chance to more thoroughly review the code.  As well as play test the code.

# What was tested

Ran unit tests.
Ran integration tests.
Started a new game with sounds of music and the music started playing after leaving the planet.

From this initial play test the conditions haven't drastically changed in a negative way.  I look forward to testing out the dynamic nature of the conditions this feature provides.

# Code style fixes

I have summarized all of your code style issues into a convenient patch that you can apply as opposed to pointing everything out in this PR.

How to apply:

1. [Download the patch](https://gist.github.com/samrocketman/5d94bb1f6d2ff06edaa4b65d3e274e17).
2. Run the command `git apply style-changes.patch`.

The style fixes are based on commit dc7f347a0f00495183f394cbd964c9a8cdd29b26

tests/unit/src/test_conditionsStore.cpp
```diff
@@ -0,0 +1,579 @@
...
+#include <map>
+#include <string>
+
+
+
+namespace { // test namespace
+
+// #region mock data
+
+void verifyName(const std::string &name1, const std::string &name2)
```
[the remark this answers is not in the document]
petervdmeer: I compared the indentation of those tests with the other tests; all other tests also didn't have the indentation as suggested. If we want to fix this, then we should fix it for all tests at once. I applied your patch, but removed this additional indentation for this specific test.

Maybe we should also fix this by some specific entries in the style guide about the indentation specific for the catch-based unit-tests; I feel that avoiding the initial indentation in the catch-specific files actually increases readability.

source/ConditionsStore.h
```diff
@@ -0,0 +1,197 @@
...
+class ConditionsStore {
+public:
+	// Forward declaration, needed to make ConditionEntry a friend of the
+	// DerivedProvider.
+	class ConditionEntry;
+
+	// Class for DerivedProviders, the (lambda) functions that provide access
+	// to the derived conditions are registered in this class.
+	class DerivedProvider
+	{
```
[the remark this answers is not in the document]
petervdmeer: Thanks for the patch, I applied it.

source/ConditionsStore.h
petervdmeer: Style fix was done by patch from samrocketman.

source/PlayerInfo.cpp
petervdmeer: Reputations with governments are stored in `GameData::Governments()` and they typically get copied between the conditions and the governments. This PR removes all the 2-directional copying and just performs reads and writes of those conditions directly in the `GameData::Governments()` reputation data.
This also means that reputations are not primary conditions (according to terminology in this PR they are derived conditions), and the reputation conditions will no longer be stored in savefiles, but the Government reputations still are.

source/ConditionSet.cpp
petervdmeer: You're right, this code is indeed looking for the conditions twice when they are there. I'll have a look if I can improve on this.

petervdmeer: I added an `HasGet` function which performs the Has and Get functions with a single lookup (and returns a pair with the results).

An alternative approach would be to implement a Find which returns a `ConditionEntry`, but the ConditionEntry will not always exist for derived conditions, and I believe it is performance-wise better not to create it when we are only performing read-access.

source/Test.cpp
petervdmeer: The `ConditionsStore::empty()` function does not have a well-defined meaning when we are working with derived conditions. I removed this function.

source/ConditionSet.cpp
Hurleveur: Looks good to me this way

Hurleveur (approved):
Seems good to me now, this is a very good improvement to the quality of the code in my opinion.

tests/unit/src/test_conditionsStore.cpp
```diff
@@ -0,0 +1,702 @@
...
+}
+
+
+SCENARIO( "Providing multiple derived conditions", "[ConditionStore][DerivedMultiple]" )
+{
+	GIVEN( "A conditionsStore and a prefixed provider" )
+	{
+		auto store = ConditionsStore();
+		auto mockProvPrefixShips = MockConditionsProvider();
+		mockProvPrefixShips.SetRWPrefixProvider(store, "ships: " );
```
samrocketman, line 673:
```suggestion
		mockProvPrefixShips.SetRWPrefixProvider(store, "ships: ");
```

My mistake on this

tests/unit/src/test_conditionSet.cpp
```diff
@@ -28,6 +31,19 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+int primarySize(const ConditionsStore &store)
+{
+	int size = 0;
+	auto it = store.PrimariesBegin();
+	while(it != store.PrimariesEnd())
+	{
+		++it;
+		++size;
+	}
+	return size;
```
quyykk, line 44: You can use `std::distance` here.

quyykk (commented):
Nice, I think this looks good now. If you can the primary iterators in a later refactor that would be great yeah.

source/ConditionsStore.h
```diff
@@ -0,0 +1,187 @@
...
+		using pointer = const std::pair<const std::string, int64_t>*;
+		using reference = std::pair<const std::string, int64_t>;
+		
+		// Default input_iterator operations.
+		std::pair<std::string, int64_t> operator*() const;
+		const std::pair<std::string, int64_t>* operator->();
+		PrimariesIterator& operator++();
+		PrimariesIterator operator++(int);
+		bool operator== (const PrimariesIterator& rhs) const;
+		bool operator!= (const PrimariesIterator& rhs) const;
```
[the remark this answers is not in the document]
petervdmeer: This review-comment is not on the latest commit; this was already fixed.

source/ConditionsStore.cpp
petervdmeer: This review-comment is not on the latest commit; this was already fixed.

source/PlayerInfo.cpp
```diff
@@ -2564,76 +2519,151 @@ void PlayerInfo::ValidateLoad()
...
+
+	auto &&unpaidMaintenanceProvider = conditions.GetProviderNamed("unpaid maintenance");
+	unpaidMaintenanceProvider.SetGetFun([this](const string &name) { return min(limit, accounts.MaintenanceDue()); });
+
+	auto &&creditScoreProvider = conditions.GetProviderNamed("credit score");
+	creditScoreProvider.SetGetFun([this](const string &name) { return accounts.CreditScore(); });
+
+	// Read-only flagship conditions.
+	auto &&flagshipCrewProvider = conditions.GetProviderNamed("flagship crew");
+	flagshipCrewProvider.SetGetFun([this](const string &name) -> int64_t { if(flagship) { return flagship->Crew(); } return 0; });
```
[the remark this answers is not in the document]
petervdmeer: Yes, that's nice. I updated the code as suggested.

tests/unit/src/test_conditionSet.cpp
petervdmeer: I added a TODO item for this. (As you noticed before; the iterator that is needed for this is not yet correctly implemented.)

source/ConditionsStore.h
```diff
@@ -0,0 +1,187 @@
...
+	
+	public:
+		PrimariesIterator(CondMapItType it, CondMapItType endIt);
+		
+		// Iterator traits
+		using iterator_category = std::input_iterator_tag;
+		using value_type = std::pair<const std::string, int64_t>;
+		using difference_type = void;
+		using pointer = const std::pair<const std::string, int64_t>*;
+		using reference = std::pair<const std::string, int64_t>;
```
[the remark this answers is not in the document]
petervdmeer: This code is indeed not correct as it is currently implemented. I expect that we can fix this by returning pointers to ConditionEntries instead of the pairs that are used now, but for now I added some TODO texts to the code here to at least signal the issues. I also started [a new PR](https://github.com/endless-sky/endless-sky/pull/7075) to remove the last condition-iterators, after which the PrimariesIterator can also be fully removed.

I am a bit affraid that tehhowch might care. He has some ideas on conditions of which I'm not fully aware yet, and he did remark in the past that he saw a need to be able to iterate over all possible conditions that exist. I'm not sure if this remark still applies after this PR and after #7075.

Amazinite (approved):

Hurleveur (approved):
Do I have to click viewed on every single file for it to not dismiss my review?
In general I don't do that because I like to look back and forth between the classes.

tehhowch (commented):
Sharing whatever comments I had stashed for when I was able to review.

(App doesn't want to make seeing which ones are relevant still easy to see)

source/PlayerInfo.cpp
```diff
@@ -1721,6 +1721,43 @@ int64_t PlayerInfo::GetCondition(const string &name) const
 
 
 
+// Copy conditions matching the prefix into the given map.
+void PlayerInfo::GetConditions(map<string, int64_t> &targetMap, const string &prefix) const
```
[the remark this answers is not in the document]
petervdmeer: This code is no longer in use. The latest version of the code returns a ConditionsStore, which seems to be more elegant than this code.

source/ConditionsProvider.h
```diff
@@ -0,0 +1,54 @@
...
+
+
+
+// Class that describes the interface to retrieve and store condition-variables.
+// This class is to be inherited from by any class that provides condition-
+// variables.
+class ConditionsProvider {
+public:
+	// Retrieve a "condition" flag from this provider.
+	virtual int64_t GetCondition(const std::string &name) const = 0;
```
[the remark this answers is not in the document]
petervdmeer: Agreed, but this code is also no longer in use; the Provider interfaces are now defined as a subclass within ConditionsStore, and the abstraction is done using lambda functions..

source/ConditionsStore.h
```diff
@@ -0,0 +1,67 @@
...
+#include <string>
+
+
+
+// Class that describes and implements the interface to retrieve and store
+// condition-variables. This class can be used as stand-alone conditions-
+// provider or as basis for other classes that need the same implementation
+// as is done in this class.
+// Classes that want to store conditions differently than that is done in
+// this class should inherit from ConditionsProvider.
```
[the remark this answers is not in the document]
petervdmeer: No classes derive from the providers anymore in the latest version of the code.

source/SpaceportPanel.cpp
```diff
@@ -112,7 +112,7 @@ const News *SpaceportPanel::PickNews() const
 {
 	vector<const News *> matches;
 	const Planet *planet = player.GetPlanet();
-	const map<string, int64_t> &conditions = player.Conditions();
+	auto &conditions = player.Conditions();
```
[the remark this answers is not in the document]
petervdmeer: Good comment, [quyykk also noticed this](https://github.com/endless-sky/endless-sky/pull/6373#discussion_r776866023) and the issue was already fixed.

source/ConditionsStore.h
```diff
@@ -0,0 +1,152 @@
...
+// formulae).
+class ConditionsStore {
+public:
+	// Storage types for conditions;
+	// - VALUE means that the condition-value is stored directly in this store.
+	// - PREFIX_PROVIDER means that conditions starting with the given prefix
+	//         are all handled by the given provider.
+	// - EXACT_PROVIDER means that the condition with the exact name is handled
+	//         by the given provider.
+	enum StorageType {VALUE, PREFIX_PROVIDER, EXACT_PROVIDER};
```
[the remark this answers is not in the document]
petervdmeer: Good comment, but the enum no longer exists in the latest version of the code.

source/ConditionsStore.h
```diff
@@ -0,0 +1,152 @@
...
+
+	// Typedefs for DerivedProvider functions
+	typedef std::function<int64_t(const std::string&)> GetFun;
+	typedef std::function<bool(const std::string&)> HasFun;
+	typedef std::function<bool(const std::string&, int64_t)> SetFun;
+	typedef std::function<bool(const std::string&)> EraseFun;
```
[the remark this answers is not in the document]
petervdmeer: The latest version of the code uses private `std::function` variables instead of the typedefs.

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,276 @@
...
+	
+	--it;
+	// The entry is valid if we have an exact stringmatch, but also when we have a
+	// prefix entry and the prefix part matches.
+	if(!(name.compare(0, it->first.length(), it->first)) &&
+			(it->second.type == PREFIX_PROVIDER || it->first.length() == name.length()))
+		return &(it->second);
+	
+	return nullptr;
+}
```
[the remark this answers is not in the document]
petervdmeer: Already fixed based on [the same comment from quyykk](https://github.com/endless-sky/endless-sky/pull/5244#discussion_r687091461).

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+
+
+// Helper function to ensure that the primary-conditions iterator points
+// to a primary (value) condition or to the end-iterator value.
+void ConditionsStore::PrimariesIterator::MoveToValueCondition()
+{
+	while((condMapIt != condMapEnd) && (condMapIt->second).type != StorageType::VALUE)
+		condMapIt++;
```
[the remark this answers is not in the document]
petervdmeer: > If you're incrementing `condMapIt` every time before calling this

We don't increment before the call if/when we create the initial iterator. (the very first position could already contain a valid value.)

source/ConditionsStore.cpp
```diff
@@ -0,0 +1,276 @@
...
+	
+	if(condMapIt != condMapEnd)
+		itVal = make_pair(condMapIt->first, (condMapIt->second).value);
+}
```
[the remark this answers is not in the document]
petervdmeer: I'll add a comment (in https://github.com/endless-sky/endless-sky/pull/7226)

source/ConditionSet.cpp
```diff
@@ -162,12 +163,12 @@ namespace {
...
-				if(temp != created.end())
-					value = temp->second;
-				else if(perm != conditions.end())
-					value = perm->second;
+				bool tempHas = created.HasCondition(str);
+				bool permHas = conditions.HasCondition(str);
+				if(tempHas)
+					value = created.GetCondition(str);
+				else if(permHas)
+					value = conditions.GetCondition(str);
```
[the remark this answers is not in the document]
petervdmeer: The permHas is indeed moved within the else part (since we don't need to look up anything for it if the tempHas matches).

source/ConditionsStore.h
```diff
@@ -0,0 +1,151 @@
...
+	
+	// Helper to completely remove all data and linked condition-providers from the store.
+	void Clear();
+
+
+private:
+	// Retrieve a condition entry based on a condition name, doCreate indicate if the entry
+	// should be created if it wasn't there yet.
+	ConditionEntry *GetEntry(const std::string &name);
+	const ConditionEntry *GetEntryConst(const std::string &name) const;
```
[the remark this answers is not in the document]
petervdmeer: I feel it's more elegant to explicitly handle this at the functions themselves; having the casts at only a single location.

tests/src/test_conditionsStore.cpp
```diff
@@ -0,0 +1,498 @@
...
+			const auto store = ConditionsStore{{"hello world", 100}, {"goodbye world", 404}};
+			THEN( "given primary conditions are in the Store" ) {
+				REQUIRE( store.GetCondition("hello world") == 100 );
+				REQUIRE( store.GetCondition("goodbye world") == 404 );
+				REQUIRE( primarySize(store) == 2 );
+			}
+			THEN( "not given conditions return the default value" ) {
+				REQUIRE( 0 == store.GetCondition("ungreeted world") );
+				REQUIRE( primarySize(store) == 2 );
+				REQUIRE( store.GetCondition("ungreeted world") == 0 );
```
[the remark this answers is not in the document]
petervdmeer: Agreed, this is a bit overkill for the Get function. I'll improve this code in https://github.com/endless-sky/endless-sky/pull/7226.

tests/src/test_conditionsStore.cpp
```diff
@@ -0,0 +1,498 @@
...
+		}
+	}
+}
+
+SCENARIO( "Setting and erasing conditions", "[ConditionStore][ConditionSetting]" ) {
+	GIVEN( "An empty conditionsStore" ) {
+		auto store = ConditionsStore();
+		REQUIRE( primarySize(store) == 0 );
+		WHEN( "a condition is set" ) {
+			REQUIRE( store.SetCondition("myFirstVar", 10) == true );
```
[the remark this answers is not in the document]
petervdmeer: Good point; I will improve this in https://github.com/endless-sky/endless-sky/pull/7226

source/ConditionSet.cpp
```diff
@@ -546,21 +547,21 @@ bool ConditionSet::Expression::Test(const Conditions &conditions, const Conditio
...
 
 // Assign the computed value to the desired temporary condition.
-void ConditionSet::Expression::TestApply(const Conditions &conditions, Conditions &created) const
+void ConditionSet::Expression::TestApply(const ConditionsStore &conditions, ConditionsStore &created) const
 {
-	int64_t &c = created[Name()];
+	int64_t c = created.GetCondition(Name());
 	int64_t value = right.Evaluate(conditions, created);
-	c = fun(c, value);
+	created.SetCondition(Name(), fun(c, value));
```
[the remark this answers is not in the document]
petervdmeer: Was fixed in a different way by adding `operator[]`.

PR #5259, opened 2020-08-25
Wanderer Campaign--Sestor Jump Drives

kestrel1110:
## Summary
Adds Jump Drives to the Sestor fleets in human space during the Alpha incursion. No interference with any of the other Sestor fleets at any other point in-game.
The alphas must have gotten the korath ships over somehow, so this adds them to the ships.

## PR Checklist
 - [N/A] I uploaded the necessary image, blend, and texture assets here: {none}
 - [N/A] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: {none}


Amazinite:
The Alphas did the same thing as the Unfettered. That is, with a small number of jump drives, they moved a group of ships across the gap, then had one ship return with all the jump drives to fit more ships, and repeated. Then, given the difficult of obtaining JDs in the first place, they kept the JDs somewhere else while the ships were sent into combat. (Notice how after a certain point, no Unfettered ships in Wanderer space during battles are equipped with JDs.) This is why the Sestor ships don't have JDs when you fight them, and why it doesn't make sense to change it so that they do.

Zitchas:
It is also worth pointing out that it can be problematically hard tracking down and destroying all those ships to begin with. Problematic to the point we have people asking for help on reddit and discord thinking that the mission not completing is a bug. If that fleet has a bunch of JD to potentially run around with, it'll be even harder to kill them all.

Amazinite:
(The NPCs have `staying` so the drive they have doesn't matter for gameplay purposes.)

PR #5270, opened 2020-08-31
Fix: Changing Alpha Combat Values to be better than the Navy's

Zitchas:
**Bugfix:** This PR addresses the fact that Alpha Hand to Hand combat values are equal to that of a Navy soldier, despite being described as worth several of them.

## Fix Details
Attack from 1.4 up to 3
Defense from 2.2 up to 3

Biologically, Alphas are supposed to be super-soldiers, engineered to be perfect warriors; Most of the description revolves around the physical, but there are hints here and there that it is mental as well.

Socially, the current Alphas are portrayed as veteran survivors who are aggressive guerrilla warfare specialists and stealth operators. They are described in the lore as one Alpha as being worth several Navy soldiers. This implies that they are experienced, well trained, and combat-hardened. It is worth noting that even when the Navy is forwarned and can bring their elite forces to bear, they still expect to need ten soldiers for every one alpha. 

The stats are intended to represent a combination of being biologically a "perfect warrior" as well as considerable combat experience and training. 

The values are arbitrary. They look good enough to be impressive, but not so high as to be insurmountable. They should make capturing Alpha ships significantly harder without making them impossible.

edit: Values changed to 3 for each. This is intended to represent the fact that not all Alpha crew are actually Alpha's. They may have a number of pirates and/or slaves too. (especially for cannon-fodder roles) I personally think it is likely that all the *required* crew is Alphas, and that extra is canon-fodder; but we really have no insight into Alpha society.

On the one hand, if they are super-egotistical then an Alpha wouldn't want to be subordinate to another, and thus they would need to have lots of slaves to fill the roles under a small set of Alphas. 
On the other hand, they were engineered to be perfect soldiers. Perfect soldiers implies being perfect members of an army, which involves a large numbers of people working together for optimum combat effectiveness. If this is the case, then they would see relying on slaves as a weakness, since no slave would be able to perform at the level of an Alpha, and thus would be a detriment to the ship.

## Impact
Giftbringer is going to be harder to capture, as is any other Alpha ships we happen to run into during future stories; but otherwise the impact is minimal. The largest impact will simply be people no longer looking at the combat stats and going "But aren't Alpha's supposed to be *better* than Navy Soldiers?"

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -10,8 +10,8 @@
 
 government "Alpha"
 	swizzle 1
-	"crew attack" 8
-	"crew defense" 6
+	"crew attack" 5
+	"crew defense" 4
 	"player reputation" -1000
 	"bribe" 0
 
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -10,8 +10,8 @@
 
 government "Alpha"
 	swizzle 1
-	"crew attack" 1.4
-	"crew defense" 2.2
+	"crew attack" 8
+	"crew defense" 6
 	"player reputation" -1000
 	"bribe" 0
 
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -10,8 +10,8 @@
 
 government "Alpha"
 	swizzle 1
-	"crew attack" 3
-	"crew defense" 3
+	"crew attack" 1.5
+	"crew defense" 2.5
 	"player reputation" -1000
 	"bribe" 0
 
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -10,8 +10,8 @@
 
 government "Alpha"
 	swizzle 1
-	"crew attack" 5
-	"crew defense" 4
+	"crew attack" 3
+	"crew defense" 3
 	"player reputation" -1000
 	"bribe" 0
 


Amazinite:
>[I’d leave the Alphas alone, or at least not give them so large a buff. “Alpha” government ships likely aren’t fully staffed by Alphas, but actually mostly be pirates or perhaps slaves.](https://github.com/endless-sky/endless-sky/pull/4168#issuecomment-459770898)

> [Initially, Alphas had much higher numbers, lowered by the request of Amazinite (see the sister PR). This is just a tentative low number.](https://github.com/endless-sky/endless-sky/pull/4170#issuecomment-459870056)

Zitchas:
I feel that they should at least be higher than the Navy, if only marginally.

Zitchas:
On the one hand, if they are super-egotistical then an Alpha wouldn't want to be subordinate to another, and thus they would need to have lots of slaves to fill the roles under a small set of Alphas. 

On the other hand, they were engineered to be perfect soldiers. Perfect soldiers implies being perfect members of an army, which involves a large numbers of people working together for optimum combat effectiveness. This by default means that they would not be continually clashing when they work together. If this is the case, then they would see relying on slaves as a weakness, since no slave would be able to perform at the level of an Alpha, and thus would be a detriment to the ship.

Also, using slaves as crew on a ship works (and might even be necessary) if one has a large empire and need to spread out trusted elites to run lots of ships. When one only has a *single* ship (that we know of), it makes more sense to fill it with the best of the best to ensure it has the best possible chance of survival and maximum impact.

Amazinite:
The problem is moreso with numbers than with the personality of the Alphas.
> I personally think it is likely that all the required crew is Alphas, and that extra is canon-fodder; but we really have no insight into Alpha society.

There simple can't be enough Alphas to fill out that Carrier, otherwise that would have been like half the Alphas in that group.
If you take the don't bomb Zenith route:
> As the Navy troops begin pouring the cement, the massive gate to the tunnel bursts open and more Alphas than you have ever seen in one place come flooding out, several dozen of them. But even these superhuman soldiers cannot survive the concentrated firepower of hundreds of Navy troops. The air fills with explosions and smoke and powdered snow...

There were only "several dozen," which I take to be say 36 to 60 Alphas. A single Carrier requires 111 crew just to fly. There is no way that there were enough Alphas to crew an entire Carrier then only a fraction of that number on Zenith. Not only are those numbers unreasonable, but if we assume that the Carrier did have 111 Alphas then it would be stupid for them to have put that many of themselves on a single ship. My guess is that there were only a handful of Alphas tops on that Carrier.

Zitchas:
It's kind of hard to get into their mindset. What is their motivation? Are they still acting like super-soldiers out to complete their mission? Or are they self-identifying as a race/species/group where survival of the group takes priority over their mission?

If the former, then they totally should have the carrier filled with as many Alphas as they can. Perhaps that is why there are so few Alphas at the base on Zenith - Because they suffered a serious loss when their carrier (which appears to be their main mobile stronghold) was taken out. Being a key part of their mission (and, let's face it, one where they need every advantage they can get - raiding into Korath space for JD on a regular basis), it makes sense for them to concentrate their strengths there. Only having a handful of Alphas on their Carrier and staffing it with slaves and pirates puts their (possibly only) method of reliably obtaining JD and transportation at risk, and with it, potentially seriously jeopardizing everything they are doing. This means significantly increasing the risk of failure in order to reduce the chance of casualties. In a military point of view, this sort of behaviour only makes sense if those soldiers are being redirected into other equally critical missions. And so far, we've seen very little indication that they had anything else on the go.

Of course, if the Alphas have morphed from being super-soldiers with a mission to being a race of quasi-humans who have developed their own identity as superhumans who value their species more than they value their mission... Then yeah, it makes sense to spread things out.

As a side note, was there any indication of how long they had been on Avalon? It could well have been multiple decades or longer. The planet could be littered with covert bases and all sorts of breeding programs. Having enough Alphas to staff one carrier doesn't exactly seem out of the realm of possibility.

Amazinite:
We know from the invasion of Poisonwood that the Alphas do work with pirates. It was only a few Alphas that were actually seen then. We also know from the [v0.9.3 blogpost](https://endless-sky.github.io/2016/09/13/unstable-release-0.9.3.html) that it's intended to eventually be able to side with the Alphas. So I don't find it unreasonable for "Alpha" government ships to be mainly crewed by pirates, and I don't see that being a risk to the Alphas when it's likely that the pirate crew are simply loyal to them. The Syndicate decided to make deals with the Alphas after all despite the fear of the Alphas among human space. There's no telling what the Alphas could be offering their human crew.

Zitchas:
I meant risk as in capabilities. If you can have elite operatives running all critical systems, why would you settle for having a slave (or former slave turned pirate/mercenary) who has significantly inferior physical and mental abilities, not to mention training, run those critical systems? Sure, the janitor doesn't have to be an Alpha. Nor the cook. But all the security, piloting, navigation, engineering, and control positions I'd sure want to be my elites. Basically, any position where snap judgement, reflexes, or superior tactical ability would be an asset.

I agree, I'm sure the Alphas could provide sufficient incentive for them to stay loyal, one way or another. It isn't just about loyalty, though. It is about capability.

Amazinite:
Can non-Alphas not sufficiently run a ship normally? A ship is a ship is a ship. I don't get how having it crewed by Alphas versus only have a few Alphas at the head with the rest being pirates changes anything in terms of capabilities.

Zitchas:
I gameplay terms, nada. In real life terms, well, there's a reason why there are ace pilots and newbies. And virtually every profession on a ship has similar sorts of things. There are people with strong affinities and innate skill and abilities to do things, and lots of people that don't. The Alphas are supposed to be all-round better. And not just a little bit, either.

Anarchist2:
Would they _want_ to use as many Alphas as possible in their ships? It's an awful amount of potential loss for very little potential gain, especially on a vessel that can already defeat any other human vessel around.

Zitchas:
Flipping that around, if it is on JD ferrying duty, that makes it a prime opportunity for training lesser experienced Alphas? In theory, the most powerful ship in the fleet (and, for that matter, basically the most powerful ship in human space) should be a pretty safe place to concentrate valuable assets.

Amazinite:
I'll give you +.1 on attack and +.3 on defense. Take it or leave it.

Zitchas:
OK, I'll take it. Not at my system with convenient access right now, though, so might be a bit until I can make the change.

Zitchas:
Done. The change has been made.

Issue #2978, opened 2017-08-29
Balance: Plundering and Capturing overpowered


flaviojs:
I'm playing version 0.9.8. First I started by mining and capturing/plundering disabled pirates with my initial Shuttle. Then, with a Mule, I continued mining and capturing/plundering pirates and bounties, and built a permanent fleet of Headhunters (half for defense, half for offense). Now I'm plundering/capturing pirates and plundering Korath ships in Almach, have 7 pirate systems dominated, and have enough ships ready to go explore Jump Drive systems and kick ass.

Before exploring those systems I decided to advance in the Free Worlds storyline and that's when it hit me... the Free Worlds haven't come out yet!!! Need to wait another month of game time... I have a net worth of 200M credits, plenty of good outfits and hulls, and can go to Jump Drive systems. ~~This is mostly **thanks to capturing and plundering**, which means they are **overpowered**.~~ _(EDIT)_  Since the game is balanced based on real-word time then I guess it's balanced.

### Analysis

Mining was a good way to pay for crew salaries, pay back my loan and a nice supplement to buy my next ship. It was also very game-time effective since (after finding a good system) I could fill my whole cargo and return with any captured ships in 1 day of game-time. (I wouldn't mind being able to buy an outfit to scan asteroids that showed me the location and/or orbit of the asteroid in the minimap, but that's too much work for little gain)

Capturing ships was very profitable. The crew salaries to capture them is covered just by selling their cargo, but the most useful result is getting outfits to improve your ships for "free". The rest can be sold off when you need the money to buy your next ship or it can be used to build your fleet (reuse hulls or temporarily store outfits). I only had to buy a Mule ship, bunk rooms, laser rifles, fragmentation grenades and local maps. The rest was obtained from capturing ships (or plundering the Korath).

Plundering is also very profitable. The advantage over capturing is that it can be sold on more systems. The disadvantage is the limited cargo space, so I only use it to hunt for specific outfits. (I would prefer to have some way of boarding my fleet to transfer cargo around, but positioning myself well enough and jettisoning cargo is a good enough)

### Proposals

Messing with the value of things would do nothing about the hulls and outfits you obtain, which is the main benefit, so instead I propose other costs:

1. After the boarding panel closes, both ships will **stay "paused" an amount of time proportional to the amount of tons plundered and the amount of rounds the crew fought**, which introduces a cost of vulnerable play time.
   Lore: your crew is busy transferring stuff and fighting.
[...]

Issue #871, opened 2016-03-22
Have plundering ships take time


wargwargwarg:
Looting disabled cruisers is very unbalanced (in about 1 hour one can gather about 15Milion looting the small to medium cruisers, in a "good system with many pirates and many lawful ships to do the hard work). If you have a cavernous Mule it's much too easy to earn a ton of money at no risk.

This has been touched by several posts. Some say include wear for forcefully removing the components, others drastically drop the prices of used components.

 I propose a procedural approach, perhaps easier to do (it only needs the file with ship components)

Say you disabled a ship with 10 systems installed and 5 crew onboard. It is drifting, with 40% hull (either powerplant or engines or both are damaged). The 5 crew defend the critical systems, like powerplant and thrusters, in the hope of repairing them and getting out of there.

40% hull would mean 6 out of 10 systems are damaged in some way, on a random basis and may or may not be worth removing. At least one system should be the powerplant or thrusters.
The 5 crew defend the 2 systems ( incidentally, the most lucrative) .

In boarding panel, the list details what is damaged (perhaps degree, estimated value), what is intact and ready for looting (again value) and what is defended (by how many each, value, damage if any). And you choose if you loot, drive away defenders and loot that system or attempt capture.

More, all this looting is being done on the run, with hostiles onboard and hostile ships in the area. There should be a time counter ("you have 1 hour to grab what you can"), the limitating factor being that the more mass the device has, the more time it takes to bring it to your hold). There is no time to "loot all their shit" (Firefly, anyone?).

PS. Much simpler (and very unpopular, I guess) downgrade the Mule

nylki:
Additionally to what @wargwargwarg said, adding more danger to plundering may help, eg. self destruct (random self destruct is already implemented afaik) that damages your ship, or danger to your crew for looting instead of only when trying to capture.

jmhorjus:
The exact amount of hull damage on a ship probably shouldn't factor in, since it is very difficult to control this and it could easily go either up or down if the ship is assisted and disabled again.  Other than that, interesting ideas - I encourage you to try implementing some of them and ask questions if you run into trouble.  

[...]

Issue #2392, opened 2017-04-10
Consumable H2H outfits on attack/defend


Bladewood:
http://steamcommunity.com/app/404410/discussions/0/135512625245082896/?ctp=2#c135513901708595037
The idea is that certain (or all, depending on your opinion on the issue) hand-to-hand outfits should be consumed after attacking or defending in boarding.

SirStigII:
I'd just like to say I like this idea and I also like the cut of this "Sir_Stig" fella's jib!

tehhowch:
It would make sense that Frag Grenades and Nerve Gas are consumed no matter what the outcome, 1 per crew per cycle. After all, it blew up, or it didn't get used (yet).

Laser Rifles: you should only lose 1 per dead crew per attack action (oops, it got lost on the other ship). If you successfully capture that vessel, those laser rifles should now be on that ship. It wouldn't be unreasonable to say that the boarding crew handed all the previously dropped rifles back to the flagship after clearing, so here there would be no net change (maybe a roll for broken-in-service, at 0.1% per rifle in service).  

I like.

This would need some modifications and checks in BoardingPanel.cpp
https://github.com/endless-sky/endless-sky/blob/83506ed9ade1eebc2e2fc112068edc0d7599d7c0/source/BoardingPanel.cpp#L328, and maybe a designation on the H2H outfits that they are indeed consumable.

SirStigII:
I like those suggestions.

endless-sky:
A change like this would probably also require giving the player a choice in each "turn" of combat of whether they want to expend their grenades or not. And in general, for a long time, there's been a to-do item to redo the boarding panel to give more control over things (like, how much fuel you transfer, and whether you attack or repair the ship). But, that's been a rather low priority.

Of course, it'd be possible to implement this *without* giving the player a choice of what weapon they use, but I suspect people would immediately clamor for a choice to be added.

KiLEdEnNis:
the stronger outfits like the grenades would be "nerfed" due to getting consumed, while weaker stuff like rifles would become relevant
maybe lower the prices a bit on consumed stuff though, and maybe cooldown (in turns) on stuff like biodefenses or whatever can be a balance point as well? (infinite uses but not used every turn)

Lineth:
[...]

PR #5283, opened 2020-09-04
Feat: Realtime hailing

petervdmeer:
**Feature:** This PR implements a first step towards the feature discussed in issue #871

## Feature Details
This is the first step towards allowing the game-engine to continue running when certain panels are active. The panels get an IsModal function which will ensure that the game-engine is not paused when the function returns true.

The first panel that I'm testing this with is the HailPanel. With this PR the game-engine keeps running while the HailPanel is open. (And the planetary defence force launches while the panel still is open if you demand tribute. Note to self: don't ask tribute when being too close to the hailed planet.)

If this works well, then one of the next steps would be to update the BoardingPanel to run in real-time as well (which is the feature requested in #871).

The IsModal functionname was inspired by Tehhowchs remark in #2978.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
I need to review the HailPanel code a bit more and some more testing is required before this can be merged.

The main reviewing/testing that still is required is to check if the panel uses cached/static game-data, for example what happens if a ship gets hailed that then flies away, or if the player dies while the panel is open? Is the code in the HailPanel able to handle such situations?

The HailPanel is now also active while the event-loop is active. I also need to check if this gives undesirable interactions between those pieces of code.

## Performance Impact
No performance impact expected.

### DIFF
--- a/source/HailPanel.cpp
+++ b/source/HailPanel.cpp
@@ -309,6 +309,13 @@ bool HailPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, boo
 
 
 
+bool HailPanel::IsModal() const
+{
+	return true;
+}
+
+
+
 void HailPanel::SetBribe(double scale)
 {
 	// Find the total value of your fleet.
--- a/source/HailPanel.h
+++ b/source/HailPanel.h
@@ -40,6 +40,9 @@ public:
 	
 	virtual void Draw() override;
 	
+	// The hailpanel on the stack doesn't pause the game.
+	virtual bool IsModal() const override;
+	
 	
 protected:
 	// Only override the ones you need; the default action is to return false.
--- a/source/MainPanel.cpp
+++ b/source/MainPanel.cpp
@@ -64,7 +64,12 @@ void MainPanel::Step()
 	
 	// Depending on what UI element is on top, the game is "paused." This
 	// checks only already-drawn panels.
-	bool isActive = GetUI()->IsTop(this);
+	bool isActive = false;
+	{
+		shared_ptr<Panel> topPanel = GetUI()->Top();
+		if(topPanel)
+			isActive = topPanel->IsModal();
+	}
 	
 	// Display any requested panels.
 	if(show.Has(Command::MAP))
@@ -191,6 +196,13 @@ void MainPanel::OnCallback()
 	// Start the next step of the simulation because Step() above still
 	// thinks the planet panel is up and therefore will not start it.
 	engine.Go();
+}
+
+
+
+bool MainPanel::IsModal() const
+{
+	return true;
 }
 
 
--- a/source/MainPanel.h
+++ b/source/MainPanel.h
@@ -42,6 +42,9 @@ public:
 	// Send a command to the engine (on behalf of the player).
 	void GiveCommand(const Command &command);
 
+	// The mainpanel on the stack doesn't pause the game.
+	virtual bool IsModal() const override;
+
 	// The main panel allows fast-forward.
 	virtual bool AllowFastForward() const override;
 	
--- a/source/Panel.cpp
+++ b/source/Panel.cpp
@@ -65,6 +65,16 @@ bool Panel::TrapAllEvents()
 
 
 
+// Check if this panel is modal, which means that the gameloop in the main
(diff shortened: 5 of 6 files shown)


Amazinite:
Would it be possible to have the game not just continue running, but instead slow down when you have a modal panel open? I don't see that as being necessary right now, but I can imagine it being something we might want in the future. E.g. when you board a ship, the game outside of the panel still runs, but only at a third the speed. This means that if you're boarding a ship, you have extra time to react should you start getting fired upon.

tehhowch:
I'm of the opinion that the game should not continue running in the background when the UI is up. Rather, the boarding panel would close / minimize when you engage in some action that should take time, and would reopen automatically when said time has elapsed. While the action is underway, you would either be unable to do anything (attacking / defending), or (if plundering) any such action would result in disengaging and canceling the action (after a disengagement time, rather than immediately). In both cases, you would be unable to use offensive weapons (limitation of the engine at this point, since your projectiles may hit the boarded ship) or crewed defensive weapons (since your crew is presumably busy at the moment).

petervdmeer:
@Amazinite Sure that is possible. The discussion/alignment on the preferred slowdown/timing will likely take more time than the implementation, so doing that in a separate PR sounds very appropriate.

petervdmeer:
@tehhowch
For plundering I was thinking of the following:
- give the player the option to show and to hide the panels during the plundering (for example by pressing B again).
- allow the player to queue the plunder actions by selecting the outfits and pressing take, after which those actions are executed in the order the player selected them.
- execute fuel transfers (from the hostile ship to the player) in the background while the ships are boarded/connected and in parallel to other actions.
- execute ammo transfers (from the hostile ship to the player) also the background while the ships are boarded/connected and in parallel to other actions.
- automatically stop the plundering when the player flies away, presses weapons or performs any other ship-control-action with the panel minimized.
  - on such disengage; have a chance (depending on how far the plundering process for the current item is) that the current item being plundered flies away as flotsam.
  - we could also go for a fast disengage (just flying away) that has a risk of some crew loss (due to crew still being outside plundering stuff) and a slower disengage that takes longer but doesn't trigger crew loss.

For capturing I was thinking of the following:
- as soon as the player presses capture, then the UI closes.
- the "primary fire" button is then used to indicate that the player wants his/her crew to attack.
- the "secondary fire" button can then be used to have the crew attack with consumable H2H outfits (#2392)
- when both the target and the player don't attack for 3 seconds, then both crews retreat to their own ships and capture ends.
- any ship-movement button forces a dis-engage, possibly causing some crew to be lost because they are in the airlocks inbetween ships.
  - the loss would be bigger if the player just attacked (defending crew retreat to the players ship)
  - we could extend this later with counters that count how many crew from the player are on the hostile ship and vice-versa
    - capture could then still succeed after disengage (due to player crews still fighting)
    - and in a similar sense, the player could still die due to hostile crew still fighting after disengage.
- adding a soft-disengage option (pressing B?) that takes longer, but that does not cause crew to be lost in space.

I like the idea that time also continues while the panels are up, but those proposals could also work if the panels do pause the game.

Having said that; if we really want every panel (like the hail-panel in this PR) to always pause the game, then we should close this PR, since this PR only adds the code to continue the game while UI panels are up.

tehhowch:
My objection to running the game in the background is that the player is presented with text to read and then penalized for being unable to read it instantly. For a text-based story engine, I believe this goes against the desired user experience.

petervdmeer:
I agree with pausing the game for story engine panels that pop-up unexpectedly with relevant/unique mission-texts. The player however has full control on when the hail- and boarding-panels pop-up, so I consider them closer to the "real-time spaceship control engine" than to the "story engine".
But I'm okay with having the game paused during the hail- and boarding-panels and move all time-consuming capture and plunder actions to after closing of the boarding-panel, so I'll focus my efforts on that when I continue with this.

tehhowch:
It should be noted that the player currently is the only one that can initiate the boarding or hail panel, but this is not going to always be true. NPCs will eventually hail the player (be it to bribe or to e.g. offer a special mission), and it's possible we could come up with a way for the player to respond to being boarded (though such a "defensive" boarding panel would likely be a different class than the current one).

Issue #5247, opened 2020-08-21
NPCs that have passed their despawn conditions but have yet to despawn won't trigger objective complete conversations/dialogs


Amazinite:
**Describe the bug / To Reproduce**
Take the following NPC.
```
npc kill
	to despawn
		has "some condition"
	...
	(conversation | dialog)
		`	"You've killed the <npc>."`
```
If the despawn conditions have passed but the NPC has yet to be despawned (i.e. the player has yet to land after the `to despawn` conditions have been met), then killing the NPC will not trigger the conversation or dialog.

**Expected behavior**
The conversation or dialog of an NPC should always trigger once its objective (e.g. kill, disable, scan, board, assist) has been passed while the NPC is still active (i.e. has been spawned or has yet to despawn).

**Screenshots**
N/A

**System:**
 - Version: Latest commit (c30080a).

**Link to save file**
N/A

**Additional context**
This section of NPC.cpp is the culprit. NPC::HasSucceeded will return true if the NPC's despawn conditions have passed without checking the objectives of the NPC. This means that when the objective does pass, even though it is being ignored for the sake of the mission, the conversation or dialog that would normally pop up does not. 
```C++
void NPC::Do(const ShipEvent &event, PlayerInfo &player, UI *ui, bool isVisible)
{
    ...
    // Check if this NPC is already in the succeeded state.
    bool hasSucceeded = HasSucceeded(player.GetSystem());
    bool hasFailed = HasFailed();
    ...
    // Check if the success status has changed. If so, display a message.
    if(HasFailed() && !hasFailed && isVisible)
        Messages::Add("Mission failed.");
    else if(ui && HasSucceeded(player.GetSystem()) && !hasSucceeded)
    {
        // If "completing" this NPC displays a conversation, reference
        // it, to allow the completing event's target to be destroyed.
        if(!conversation.IsEmpty())
            ui->Push(new ConversationPanel(player, conversation, nullptr, ship));
        else if(!dialogText.empty())
            ui->Push(new Dialog(dialogText));
    }
}
```
Solutions that I can think of:
* Allow npc conversations to trigger regardless of the state of the despawn flag. This would require splitting HasSucceeded into two functions. I imagine a function named HasSucceeded would remain that is the exact same as the function right now, but there'd be a new function named SucceededObjectives that only checks the objectives without checking the spawn/despawn conditions. This function would be called by HasSucceeded and replace HasSucceeded in this instance in order for the conversation to show up.
[...]

PR #5286, opened 2020-09-04
refactor(mechanics): Update conditional spawning & despawning

tehhowch:
**Bugfix:** This PR addresses issue #5247 

## Fix Details
 - Upon landing, when the NPCs would despawn, set a flag that indicates they were actually despawned
 - Don't perform spawn condition updates when reloading the game.
 - Populate the initial state of `passedSpawnConditions` based on load-time data
 - Only short-circuit the NPC completion check via spawn conditions when the spawn conditions have been evaluated at least once (lest reloading a save complete a just-offered mission).

## Testing Done
Add this job file to a plugin / game data, launch & land, and then begin your experiments.
[npc conditional spawn missions.txt](https://github.com/endless-sky/endless-sky/files/5176290/npc.conditional.spawn.missions.txt)


source/NPC.cpp
```diff
@@ -457,9 +466,12 @@ bool NPC::IsLeftBehind(const System *playerSystem) const
...
 bool NPC::HasFailed() const
 {
-	// If this NPC has been despawned or was never spawned in the first place
-	// then ignore its objectives.
-	if(!passedSpawnConditions || passedDespawnConditions)
+	// An unspawned NPC, one which will despawn on landing, or that has
+	// already despawned, is not considered "failed."
+	// TODO: Should we re-evaluate this? Should a currently-spawned NPC
+	// influence mission state even if it would despawn on the next landing?
+	// (Removing `passedDespawnConditions` here will enable that behavior.)
```
tehhowch, line 473: Ref: https://github.com/endless-sky/endless-sky/issues/5247#issuecomment-687294133

Amazinite (changes requested):
Seems to be working as intended, but causes an issue that I expected would occur.
Could HasSucceeded be updated to ask it to ignore the passedDespawnCondition bool for NPC:Do, but not for wherever else it is called to update the mission marker?

```C++
// Default this bool to true in the header file, then we check HasSucceeded(playerSystem, false)
// in NPC::Do and everywhere else things remain the same.
bool NPC::HasSucceeded(const System *playerSystem, bool checkDespawnConditions) 
{
	if(checkedSpawnConditions && (!passedSpawnConditions || (checkDespawnConditions && passedDespawnConditions) || despawned)
		return true;
	...
}
```
Basically, in the case of checking if a mission can be completed, we *want* to check if the despawn conditions have passed, but in the case of showing conversations for completed objectives, we don't want to check the despawn conditions and instead only check if the NPC has actually been despawned. (Could this perhaps also completely remove the need for the `despawned` bool?)

source/NPC.h
```diff
@@ -60,9 +60,13 @@ class NPC {
 	
 	// Handle the given ShipEvent.
 	void Do(const ShipEvent &event, PlayerInfo &player, UI *ui = nullptr, bool isVisible = true);
-	bool HasSucceeded(const System *playerSystem, bool checkDespawnConditions = true) const;
+	// Determine if the NPC is in a successful state, assuming the player is in the given system.
+	// (By default, a despawnable NPC has succeeded and is not actually checked.)
+	bool HasSucceeded(const System *playerSystem, bool ignoreIfDespawnable = true) const;
```
Amazinite, line 65: Aren't we *not* ignoring if this is true? Variable name seems to imply the opposite of what the boolean does.

Amazinite (approved):
Just have a question about a variable name. Otherwise, looks good to me.

source/NPC.h
tehhowch: I picked the name because of what it means for HasSucceeded:

- IF {{condition}} AND the NPC is despawnable, return TRUE, no matter what.
- IF {{condition}} but not despawnable, return based on full criteria evaluation
- IF NOT {{condition}}, return based on full criteria evaluation

I thought "we can ignore success criteria when despawnable" was a suitable substitution for {{condition}}, so I went with "ignoreIfDespawnable". "skipAllChecksIfDespawnable" would also work, but is rather lengthy.

Amazinite: I don’t see why what I had didn’t work, as it describes exactly what is being done if the bool is true. (I.e. if checkDespawnConditions is true, we are checking passedDespawnConditions, but if it’s false then we’re not.) Could simply include a comment as to why we might want to not check if the despawn conditions have passed if the variable name wasn’t clear.

tehhowch: The issue I had with `checkDespawnConditions` is that you can't infer from the name of the parameter what the semantics are. "Ok, if I send `checkDespawnConditions` as `true` to HasSucceeded, it will check the despawn conditions... but ... what does that mean about the result?"

Compare to "if I tell HasSucceeded to `ignoreIfDespawnable`, it's going to ignore {{things}} if the NPC is despawnable.

(An alternative name for this boolean parameter could have been `itsOkToLieToMeAboutTheConditions` -- if we really & truly evaluate the NPC's success & failure criteria, we might find that a despawnable NPC is in a failed state or hasn't otherwise succeeded. But, we're saying it's OK to lie about that for this call.)

PR #5292, opened 2020-09-06
feat(mechanics): Support 'G' for billions.

jostephd:
**Feature:** This PR implements the feature request detailed and discussed ~in issue #{{insert number}}~ in this issue

## Feature Details
In the loan take/pay extra dialogs, allow specifying an amount in billions with `G`. (Format::Parse isn't used by anything else)

`G` denotes billions both in computers (e.g., GB of RAM) and in [SI prefixes](https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes), and doesn't have any conflicting meaning.

## UI Screenshots
2020-09-06-004135_255x129_scrot

## Usage Examples
N/A

## Testing Done
Took a loan of `0.002G`, paid back `0.001G`.

## Performance Impact
N/A

### DIFF
--- a/source/Format.cpp
+++ b/source/Format.cpp
@@ -180,7 +180,7 @@ double Format::Parse(const string &str)
 			value *= 1e3;
 		else if(*it == 'm' || *it == 'M')
 			value *= 1e6;
-		else if(*it == 'b' || *it == 'B')
+		else if(*it == 'b' || *it == 'B' || *it == 'g' || *it == 'G')
 			value *= 1e9;
 		else if(*it == 't' || *it == 'T')
 			value *= 1e12;


tehhowch:
In terms of money, though, a G is not a billion dollars.

Amazinite:
"G" can also denote thousands, though.

jostephd:
What, you think someone will actually expect the loan dialog to use G in the slang sense? And, anyway, what alternative do you propose?

Amazinite:
> And, anyway, what alternative do you propose?

No change.

> What, you think someone will actually expect the loan dialog to use G in the slang sense?

Well given that both @tehhowch and I instantly thought of Gs as denoting thousands, I guess the answer is yes. Why introduce ambiguity?

Zitchas:
In all the financial news, economic papers, and conversations I've had; I've seen lots of instances of people using K for thousands, M for millions, and B for Billions. I have never once seen anyone talking about "Giga-dollars." Even in Europe, which has whole heartedly embraced SI units. Of course, this is all anecdotal, so hardly conclusive.

Terin:
I wouldn't be adverse to some shortcuts when it comes to the bank panel. Right now, if I'm trying to take out a maximum loan, or put all my credits on hand into paying one off, I just hold down the "9" key until it's big enough, and it's pretty easy for it to overflow into an invalid input. I've been looking at adding a "Maximum" button for that, but being able to quickly type "1T" for a trillion credits or such would work as well, or even be faster.

Using metric prefixes for money doesn't quite fit though, as said. I'd make good use of it if it was swapped to B, with M/T added to millions/trillions. Once I'm taking out heavy late-game loans, being able to see I need to pay 12 million credits per day, and being able to type 80M instead of carefully counting zeroes, would also be very useful.

MCOfficer:
As long as this is just used for money, i'd say stick with the current abbreviations. Once it's also used for something else, there's nothing stopping us from supporting both `B` and `G`.

jostephd:
Closing per feedback.

@Zitchas Check out tarsnap's pricing :)

PR #5299, opened 2020-09-07
Lose frames instead of slowing down.

chrisfahr:
The following code is not production ready, as it modifies the main code path.
It is meant to test the proposed option.

------

This enables to play the game at normal speed with slow computers
or with software rendering.
There will be lagging instead of a slowed down gameplay.
Maybe this should be an option.

You could use this environment for software rendering with Mesa and llvmpipe:

export LIBGL_ALWAYS_SOFTWARE=1
export MESA_GL_VERSION_OVERRIDE=3.0
export LP_NUM_THREADS=<number of hyperthreads>

### DIFF
--- a/source/FrameTimer.cpp
+++ b/source/FrameTimer.cpp
@@ -41,14 +41,17 @@ FrameTimer::FrameTimer(int fps, int maxLagMsec)
 
 
 // Wait until the next frame should begin.
-void FrameTimer::Wait()
+double FrameTimer::Wait()
 {
+	double lagCount = 0;
 	// Note: in theory this could get interrupted by a signal handler, although
 	// it's unlikely the program will receive any signals that do not terminate
 	// it and that it does not ignore. But, the worst that would happen in that
 	// case is that this particular frame will end too quickly, and then it will
 	// go back to normal for the next one.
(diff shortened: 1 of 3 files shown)


Terin:
This is a great improvement on my older laptop, letting me use the autopilot at full speed even when I'm dropping frames while fast-forwarding. I notice I'm able to open the map at any time, even while frozen for a few seconds while stress-testing it, so I'm not sure if that's an avenue for bugs. But still, would love to see this implemented in the game fully.

InfiniteDonuts:
I can't play ES on my Raspberry Pi because the game slows down instead of losing frames. This ought to fix that problem. Looking forward to seeing it implemented!

petervdmeer:
Good to read in #5300 that this test-implementation is showing the expected results, but are any activities expected in the near-future to bring this PR to "production ready"?
If not, then I would suggest to close this PR for now; anybody interested in the test-implementation can still test by checking out the relevant branch, but this Pull Request is currently not really requesting a Pull (for production) at this moment.

petervdmeer:
Closing this Pull Request; it doesn't contain production ready code and the branch will still work fine for anybody that wants to try this code.

Issue #4399, opened 2019-06-26
Request: Give Ship command for missions


Zitchas:
**Is your feature request related to a problem? Please describe.**
There is currently no way for a mission to simply give the player a ship. This is particularly important for situations where the idea of the mission is to give the player control of a unique ship that is not available for purchase.

**Describe the solution you'd like**
The creation of a tag that could be added to missions that acts similar to the `outfit` tag that can be used to give the player the ship. One possible example could look like this:
```ship technical_name custom_name #```
Where the term `ship` is the command, technical_name is the formal name for the ship in the data files, and custom_name is what the ship is named when it is added to the player's fleet. The # is for situations where more than one ship should be given, in which case the custom name is numbered accordingly.

**Describe alternatives you've considered**
- Having all instances where a custom ship is supposed to be transferred to the player requiring the ship to be dumped in orbit as `disabled` and have the player board and capture it. 
- Using a very large series of missions to determine the player's wealth, then set the price of the ship to a very high percentage of their networth using events keyed to those missions, and then make the ship available for purchase for only a few days so they don't have time to make any more money. Thus they will only be able to afford one. A follow-up mission could restore sufficient funds to bring the the amount lost back down to the desired price.

The former doesn't always fit the story (You're giving me a prototype? Why are you ditching it in orbit and forcing me to board it to get it?) And the latter would require potentially hundreds or thousands of missions and events to manage.

**Additional context**
Add any other context or screenshots about the feature request here. If you're proposing new UI features, creating a mock-up can really go a long way in conveying your intent.

jafdy:
`custom_name` should probably be able to use the random name selection, not sure how you would code for that.

Fzzr:
> `custom_name` should probably be able to use the random name selection, not sure how you would code for that.

I'm doing a test implementation that will just use the random civilian name list if the second token is empty. A named gift ship would be specified `ship "Cruiser (Mark II)" "Welcome to the Oathkeepers"` and one with a random name would be specified `ship "Carrier (Mark II)"`. I'm leaving out `count` for now. If both random name and `count` were to be supported the code would check if the second entry is a number and use random names in that case.

jafdy:
I think making count non-optional might be better, you can always leave it at 1.

Fzzr:
Here's a WIP bare-minimum implementation. It's not working at this time - for reasons I can't figure out, when it comes time to write the `ships` to the mission the vector of ships is empty, even though it's definitely being populated while reading.

[...]

PR #5302, opened 2020-09-08
Feat: Give ship

Gods-Righthand:
## Summary
Gives creators the ability to give ships to the player through missions

-----------------------

## Feature Details
This is a partial implementation of  #4399 it allows the giving of ships through missions.

## Usage Examples
It can be used in a mission under `on offer`, `on accept`, etc.. 
```
on <action>
    give ship <ship model> [<name>]
```
If the second name is left out, a civilian ship name will be generated for the gifted ship.
e.g.
```
mission "some mission"
    ...
    on complete
        give ship Sparrow "Mote of Might"
        give ship "Star Barge"
```
indicates that completing the mission will result in the player being given a Sparrow that is named "Mote of Might," and being given a Star Barge with a randomly picked civilian name

## Testing Done
Using the example mission gives two ships, one with a name, and one without a predetermined name, it gets generated later.
#### Example mission
[test.txt](https://github.com/endless-sky/endless-sky/files/5188133/testing.txt)

## Performance Impact
n/a

### DIFF
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -39,7 +39,7 @@ namespace {
 			return;
 		
 		player.BuyShip(model, name, true);
-		Messages::Add("The " + model->ModelName() + " \"" + name + ",\" was added to your fleet.");
+		Messages::Add("The " + model->ModelName() + " \"" + name + "\" was added to your fleet.");
 	}
 	
 	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
@@ -200,7 +200,7 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			if(child.Token(1) == "ship" && child.Size() >= 3)
 				giftShips[GameData::Ships().Get(child.Token(2))] = child.Size() >= 4 ? child.Token(3) : "";
 			else
-				child.PrintTrace("Skipping unsupported give syntax:");
+				child.PrintTrace("Skipping unsupported \"give\" syntax:");
 		}
 		else if(key == "outfit" && hasValue)
 		{
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -39,8 +39,15 @@ namespace {
 			return;
 		
 		player.BuyShip(model, name, true);
+		
+		string noun = model->Noun();
+		string message;
+		char c = tolower(noun.front());
+		bool isVowel = (c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u');
+		message = (isVowel ? "An " : "A ");
+		message += noun;
 		if(ui)
-			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
+			ui->Push(new Dialog(message + ", the " + model->ModelName() + " \"" + name + ",\" was added to your fleet!"));
 	}
 	
 	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -39,14 +39,7 @@ namespace {
 			return;
 		
 		player.BuyShip(model, name, true);
-		
-		string noun = model->Noun();
-		string message;
-		char c = tolower(noun.front());
-		bool isVowel = (c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u');
-		message = (isVowel ? "An " : "A ");
-		message += noun;
-		Messages::Add(message + ", the " + model->ModelName() + " \"" + name + ",\" was added to your fleet.");
+		Messages::Add("The " + model->ModelName() + " \"" + name + ",\" was added to your fleet.");
 	}
 	
 	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -196,7 +196,7 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
-		else if(key ==  "ship" && hasValue)
+		else if(key == "ship" && hasValue)
 			giftShips[GameData::Ships().Get(child.Token(1))] = child.Size() >= 3 ? child.Token(2) : "";
 		else if(key == "outfit" && hasValue)
 		{
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -46,8 +46,7 @@ namespace {
 		bool isVowel = (c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u');
 		message = (isVowel ? "An " : "A ");
 		message += noun;
-		if(ui)
-			ui->Push(new Dialog(message + ", the " + model->ModelName() + " \"" + name + ",\" was added to your fleet!"));
+		Messages::Add(message + ", the " + model->ModelName() + " \"" + name + ",\" was added to your fleet.");
 	}
 	
 	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -33,14 +33,17 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	void DoGiftShip(PlayerInfo &player, const Ship *model, const string &name, UI *ui)
+	void DoGift(PlayerInfo &player, const Ship *model, const string &name, UI *ui)
 	{
+		if(model->ModelName().empty())
+			return;
+		
 		player.BuyShip(model, name, true);
 		if(ui)
 			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
 	}
 	
-	void DoGiftOutfit(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
+	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
 	{
 		Ship *flagship = player.Flagship();
 		bool isSingle = (abs(count) == 1);
@@ -307,7 +310,7 @@ void MissionAction::Save(DataWriter &out) const
 			conversation.Save(out);
 		
 		for(const auto &it : giftShips)
-			out.Write("ship", it.first->Name(), it.second);
+			out.Write("ship", it.first->ModelName(), it.second);
 		for(const auto &it : giftOutfits)
 			out.Write("outfit", it.first->Name(), it.second);
 		for(const auto &it : requiredOutfits)
@@ -457,15 +460,15 @@ void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, co
 			player.AddSpecialLog(it.first, eit.first, eit.second);
 	
 	for(const auto &it : giftShips)
-		DoGiftShip(player, it.first, it.second, ui);
+		DoGift(player, it.first, it.second, ui);
 	// If multiple outfits are being transferred, first remove them before
 	// adding any new ones.
 	for(const auto &it : giftOutfits)
 		if(it.second < 0)
-			DoGiftOutfit(player, it.first, it.second, ui);
+			DoGift(player, it.first, it.second, ui);
 	for(const auto &it : giftOutfits)
 		if(it.second > 0)
-			DoGiftOutfit(player, it.first, it.second, ui);
+			DoGift(player, it.first, it.second, ui);
 	
 	if(payment)
 		player.Accounts().AddCredits(payment);
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -35,7 +35,7 @@ using namespace std;
 namespace {
 	void DoGiftShip(PlayerInfo &player, const Ship *model, const std::string &name, UI *ui)
 	{
-		player.GiftShip(model, name);
+		player.BuyShip(model, name, true);
 		
 		if(ui)
 			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -36,7 +36,6 @@ namespace {
 	void DoGiftShip(PlayerInfo &player, const Ship *model, const std::string &name, UI *ui)
 	{
 		player.BuyShip(model, name, true);
-		
 		if(ui)
 			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
 	}
@@ -306,6 +305,7 @@ void MissionAction::Save(DataWriter &out) const
 		}
 		if(!conversation.IsEmpty())
 			conversation.Save(out);
+		
 		for(const auto &it : giftShips)
 			out.Write("ship", it.first->Name(), it.second);
 		for(const auto &it : giftOutfits)
@@ -507,7 +507,6 @@ MissionAction MissionAction::Instantiate(map<string, string> &subs, const System
 		int day = it.second.first + Random::Int(it.second.second - it.second.first + 1);
 		result.events[it.first] = make_pair(day, day);
 	}
-	
 	for(const auto &it : giftShips)
 		result.giftShips[it.first] = !it.second.empty() ? it.second : GameData::Phrases().Get("civilian")->Get();
 	result.giftOutfits = giftOutfits;
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -33,7 +33,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	void DoGiftShip(PlayerInfo &player, const Ship *modelName, const std::string &name, UI *ui)
+	void DoGiftShip(PlayerInfo &player, const Ship *model, const std::string &name, UI *ui)
 	{
 		player.GiftShip(modelName, name);
 		
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -33,7 +33,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	void DoGiftShip(PlayerInfo &player, const Ship *model, const std::string &name, UI *ui)
+	void DoGiftShip(PlayerInfo &player, const Ship *model, const string &name, UI *ui)
 	{
 		player.BuyShip(model, name, true);
 		if(ui)
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -195,7 +195,7 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
 		else if(key ==  "ship" && hasValue)
-			giftShips[GameData::Ships().Get(child.Token(1))] = (child.Size() > 2 ? child.Token(2) : "");
+			giftShips[GameData::Ships().Get(child.Token(1))] = child.Size() >= 3 ? child.Token(2) : "";
 		else if(key == "outfit" && hasValue)
 		{
 			int count = (child.Size() < 3 ? 1 : static_cast<int>(child.Value(2)));
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -196,8 +196,13 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
-		else if(key == "ship" && hasValue)
-			giftShips[GameData::Ships().Get(child.Token(1))] = child.Size() >= 3 ? child.Token(2) : "";
+		else if(key == "give" && hasValue)
+		{
+			if(child.Token(1) == "ship" && child.Size() >= 3)
+				giftShips[GameData::Ships().Get(child.Token(2))] = child.Size() >= 4 ? child.Token(3) : "";
+			else
+				child.PrintTrace("Skipping unsupported give syntax:")
+		}
 		else if(key == "outfit" && hasValue)
 		{
 			int count = (child.Size() < 3 ? 1 : static_cast<int>(child.Value(2)));
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -201,7 +201,7 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			if(child.Token(1) == "ship" && child.Size() >= 3)
 				giftShips[GameData::Ships().Get(child.Token(2))] = child.Size() >= 4 ? child.Token(3) : "";
 			else
-				child.PrintTrace("Skipping unsupported give syntax:")
+				child.PrintTrace("Skipping unsupported give syntax:");
 		}
 		else if(key == "outfit" && hasValue)
 		{
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -35,7 +35,7 @@ using namespace std;
 namespace {
 	void DoGiftShip(PlayerInfo &player, const Ship *model, const std::string &name, UI *ui)
 	{
-		player.GiftShip(modelName, name);
+		player.GiftShip(model, name);
 		
 		if(ui)
 			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -33,7 +33,15 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
+	void DoGiftShip(PlayerInfo &player, const Ship *modelName, const std::string &name, UI *ui)
+	{
+		player.GiftShip(modelName, name);
+		
+		if(ui)
+			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
+	}
+	
+	void DoGiftOutfit(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
 	{
 		Ship *flagship = player.Flagship();
 		bool isSingle = (abs(count) == 1);
@@ -186,11 +194,13 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
+		else if(key ==  "ship" && hasValue)
+			giftShips[GameData::Ships().Get(child.Token(1))] = (child.Size() > 2 ? child.Token(2) : "");
 		else if(key == "outfit" && hasValue)
 		{
 			int count = (child.Size() < 3 ? 1 : static_cast<int>(child.Value(2)));
 			if(count)
-				gifts[GameData::Outfits().Get(child.Token(1))] = count;
+				giftOutfits[GameData::Outfits().Get(child.Token(1))] = count;
 			else
 			{
 				// outfit <outfit> 0 means the player must have this outfit.
@@ -296,8 +306,9 @@ void MissionAction::Save(DataWriter &out) const
 		}
 		if(!conversation.IsEmpty())
 			conversation.Save(out);
-		
-		for(const auto &it : gifts)
+		for(const auto &it : giftShips)
+			out.Write("ship", it.first->Name(), it.second);
+		for(const auto &it : giftOutfits)
 			out.Write("outfit", it.first->Name(), it.second);
 		for(const auto &it : requiredOutfits)
 			out.Write("require", it.first->Name(), it.second);
@@ -342,7 +353,7 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
 		return false;
 	
 	const Ship *flagship = player.Flagship();
-	for(const auto &it : gifts)
+	for(const auto &it : giftOutfits)
 	{
 		// If this outfit is being given, the player doesn't need to have it.
 		if(it.second > 0)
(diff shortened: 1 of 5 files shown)


Zitchas:
I'm happy to see this appear. Thanks for working on it!

Gods-Righthand:
Yes, I picked it up from Fzzr worked on it some then asked NomadicVolcano for help. He did most of it and didn't walk me through what he did, so with my knowledge of C++ I wasn't able to pick it back up and keep going. Since then it has just been sitting.

Amazinite (changes requested):
Needs to conform to the [style guide](https://endless-sky.github.io/styleguide/styleguide.xml).
* Use tabs, not spaces.
* You do not need to use braces if a loop or conditional body is a single expression. Some of your new code uses extra braces and you've also added them to existing code. (Example:)
```
	if(ship == nullptr) // This can also just be if(!ship)
--	{ // Delete
		return;
--	} // Delete
```
* Variable names should be in camelCase, starting with a lower-case letter.
  * recordDepreciation, not record_depreciation.

Gods-Righthand:
Oh, right, the style guide, thanks for reminding me.

source/PlayerInfo.cpp
```diff
@@ -841,15 +841,32 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
+// Receive a ship as a gift
+shared_ptr<Ship> PlayerInfo::GiftShip(const Ship *model, const std::string &name, bool record_depriciation)
+{
+	return ReceiveShip(model, name, 0, record_depriciation);
+}
+
+
+
+// Purchase or receive a gift of a ship of the given model and name
+shared_ptr<Ship> PlayerInfo::ReceiveShip(const Ship *model, const string &name, int64_t cost, bool record_depriciation)
```
Amazinite, line 861: Why does ReceiveShip return a pointer to the ship? Under what circumstances would this ever return anything aside from the ship that was just given to it? The only instance in which a nullptr would be returned is if the cost was greater than the player had, but gifted ships have no cost. (A nullptr could also be returned if the model doesn't exist, but checking if the model exists is already done by DoGiftShip in MissionAction.)

source/MissionAction.h
```diff
@@ -83,7 +84,8 @@ class MissionAction {
 	Conversation conversation;
 	
 	std::map<const GameEvent *, std::pair<int, int>> events;
-	std::map<const Outfit *, int> gifts;
+	std::vector<std::pair<std::string, std::string>> giftShips;
```
Amazinite, line 87: giftShips should be a map like giftOutfits. I'd suggest `std::map<const Ship *, std::string>` where the string is the name. Reason being that as this code is written right now, a mission could be written to gift a nonexistent ship (e.g. the mission writer had a typo in the ship name), but this won't throw an error until the mission actually tried to gift the ship, because that is the only point where the string for the ship model is actually converted into the actual ship model.

source/PlayerInfo.cpp
```diff
@@ -841,15 +841,32 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
 {
 	if(!model)
-		return;
+		return nullptr;
 	
 	int day = date.DaysSinceEpoch();
-	int64_t cost = stockDepreciation.Value(*model, day);
-	if(accounts.Credits() >= cost)
+	if(cost == DEFAULT_COST)
```
Amazinite, line 867: This DEFAULT_COST seems clumsy to me. The cost parameter can ONLY ever be one of two things; either it is 0, or it is DEFAULT_COST. We should just be using a boolean. This same boolean can replace recordDepreciation, as the cost is always tied to that boolean right now; if cost is 0, then we know that recordDepreciation is false. If cost is DEFAULT_COST, then we know recordDepreciation is true. So just create a single `gifting` bool that is true for GiftShip and false for BuyShip.

Amazinite (changes requested):

source/PlayerInfo.cpp
Gods-Righthand: Yeah, that makes sense. Made the changes

source/MissionAction.cpp
```diff
@@ -33,7 +33,16 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
 
 namespace {
-	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
+	void DoGiftShip(PlayerInfo &player, const Ship *modelName, const std::string &name, UI *ui)
+	{
+		player.GiftShip(modelName, name);
+		
+		string message = "The " + name + " was added to your fleet!";
+		if(ui)
+			ui->Push(new Dialog(message));
```
Amazinite, line 42: This dialog can only ever be one thing, so I wouldn't bother with storing it in a variable. Just pass it directly to the dialog.
```suggestion
		if(ui)
			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
```

source/MissionAction.cpp
```diff
@@ -186,11 +195,15 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
+		else if(key ==  "ship" && hasValue)
+		{
+			giftShips[GameData::Ships().Get(child.Token(1))] = (child.Size() > 2 ? child.Token(2) : "");
+		}
```
Amazinite, line 201: Style: unnecessary braces.
```suggestion
			giftShips[GameData::Ships().Get(child.Token(1))] = (child.Size() > 2 ? child.Token(2) : "");
```

source/MissionAction.cpp
```diff
@@ -202,7 +215,9 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 		{
 			int count = (child.Size() < 3 ? 1 : static_cast<int>(child.Value(2)));
 			if(count >= 0)
+            {
 				requiredOutfits[GameData::Outfits().Get(child.Token(1))] = count;
+            }
```
Amazinite, line 220: Style: unnecessary braces.
```suggestion
				requiredOutfits[GameData::Outfits().Get(child.Token(1))] = count;
```

source/MissionAction.cpp
```diff
@@ -296,9 +311,14 @@ void MissionAction::Save(DataWriter &out) const
...
-		
-		for(const auto &it : gifts)
+		for(const auto &it : giftShips)
+		{
+			out.Write("ship", it.first->Name(), it.second);
+		}
+		for(const auto &it : giftOutfits)
+		{
 			out.Write("outfit", it.first->Name(), it.second);
+		}
```
Amazinite, line 321: Style: unnecessary braces.
```suggestion
			out.Write("ship", it.first->Name(), it.second);
		for(const auto &it : giftOutfits)
			out.Write("outfit", it.first->Name(), it.second);
```

source/MissionAction.cpp
```diff
@@ -445,14 +465,18 @@ void MissionAction::Do(PlayerInfo &player, UI *ui, const System *destination, co
 		for(const auto &eit : it.second)
 			player.AddSpecialLog(it.first, eit.first, eit.second);
 	
+	for(const auto &it : giftShips)
+	{
+		DoGiftShip(player, it.first, it.second, ui);
+	}
```
Amazinite, line 471: Style: unnecessary braces.
```suggestion
		DoGiftShip(player, it.first, it.second, ui);
```

source/MissionAction.cpp
```diff
@@ -494,7 +518,10 @@ MissionAction MissionAction::Instantiate(map<string, string> &subs, const System
 		int day = it.second.first + Random::Int(it.second.second - it.second.first + 1);
 		result.events[it.first] = make_pair(day, day);
 	}
-	result.gifts = gifts;
+	
+	for(const auto &it : giftShips)
+	result.giftShips[it.first] = !it.second.empty() ? it.second : GameData::Phrases().Get("civilian")->Get();
```
Amazinite, line 523: Format: missing tab.
```suggestion
		result.giftShips[it.first] = !it.second.empty() ? it.second : GameData::Phrases().Get("civilian")->Get();
```

source/PlayerInfo.cpp
```diff
@@ -841,15 +841,32 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
+// Receive a ship as a gift
+void PlayerInfo::GiftShip(const Ship *model, const std::string &name)
+{
+	return ReceiveShip(model, name, 0, true);
+}
+
+
+
+// Purchase or receive a gift of a ship of the given model and name
+void PlayerInfo::ReceiveShip(const Ship *model, const string &name, int64_t cost, bool gifting)
```
Amazinite, line 861: Someone still didn't double check what I posted.
```suggestion
void PlayerInfo::BuyShip(const Ship *model, const std::string &name)
{
	ReceiveShip(model, name, false);
}



// Receive a ship as a gift
void PlayerInfo::GiftShip(const Ship *model, const std::string &name)
{
	ReceiveShip(model, name, true);
}



// Purchase or receive a gift of a ship of the given model and name
void PlayerInfo::ReceiveShip(const Ship *model, const string &name, bool gifting)
```

source/PlayerInfo.cpp
```diff
@@ -841,15 +841,32 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
 {
 	if(!model)
 		return;
 	
 	int day = date.DaysSinceEpoch();
-	int64_t cost = stockDepreciation.Value(*model, day);
-	if(accounts.Credits() >= cost)
+	if(cost == !gifting)
```
Amazinite, line 867: Cost is not a boolean.
```suggestion
	int64_t cost = 0;
	if(!gifting)
```

source/PlayerInfo.cpp
```diff
@@ -841,15 +841,32 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
+		cost = stockDepreciation.Value(*model, day);
+	if(cost <= 0 || accounts.Credits() >= cost)
```
Amazinite, line 869: Unnecessary change.
```suggestion
	if(accounts.Credits() >= cost)
```

source/PlayerInfo.cpp
```diff
@@ -859,13 +876,17 @@ void PlayerInfo::BuyShip(const Ship *model, const string &name)
 		ships.back()->SetIsYours();
 		ships.back()->SetGovernment(GameData::PlayerGovernment());
 		
-		accounts.AddCredits(-cost);
+		if(cost)
+			accounts.AddCredits(-cost);
```
Amazinite, line 880: Unnecessary change.
```suggestion
		accounts.AddCredits(-cost);
```

source/PlayerInfo.h
```diff
@@ -28,6 +28,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include <string>
 #include <utility>
 #include <vector>
+#include <limits>
```
Amazinite, line 31: No longer needed.
```suggestion
```

source/PlayerInfo.h
```diff
@@ -247,6 +249,9 @@ class PlayerInfo {
 	
 	
 private:
+	// Internal implementation of BuyShip and GiftShip
+	void ReceiveShip(const Ship *model, const std::string &name, int64_t cost, bool gifting);
```
Amazinite, line 253:
```suggestion
	void ReceiveShip(const Ship *model, const std::string &name, bool gifting);
```

source/PlayerInfo.h
```diff
@@ -325,6 +330,8 @@ class PlayerInfo {
 	Depreciation stockDepreciation;
 	std::set<std::pair<const System *, const Outfit *>> harvested;
 	
+	bool gifting = true;
+
```
Amazinite, line 334: Unnecessary change.
```suggestion
```

Amazinite (changes requested):

source/PlayerInfo.cpp
```diff
@@ -841,14 +841,32 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
+void PlayerInfo::ReceiveShip(const Ship *model, const string &name, bool gifting)
 {
 	if(!model)
 		return;
 	
 	int day = date.DaysSinceEpoch();
-	int64_t cost = stockDepreciation.Value(*model, day);
+	int64_t cost = 0;
+	if(!gifting)
+		cost = stockDepreciation.Value(*model, day);
```
Amazinite, line 869: Just one more quick suggestion.
```suggestion
	int64_t cost = gifting ? 0 : stockDepreciation.Value(*model, day);
```

Amazinite (changes requested):

Amazinite:
Somehow. a commit on this branch had changed files in the util and test folders, but it was one of those strange instances where somethings is changed but at the same time it says file not changed. Couldn't just replace them with "unchanged" files because it didn't recognize that a change had occurred. I opted to just reset the branch to before the commit was made that changes the util and test folder files then update it from there, but turns out that ate every single commit that was here previously.

Need to credit Fzzr and Nomadic once this is merged.

source/MissionAction.cpp
```diff
@@ -33,7 +33,15 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
+	void DoGiftShip(PlayerInfo &player, const Ship *modelName, const std::string &name, UI *ui)
```
petervdmeer, line 36: The const Ship pointer points to a complete model, not just to a modelname, so I suggest to call the variable model.

source/MissionAction.cpp
```diff
@@ -186,11 +194,13 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
+		else if(key ==  "ship" && hasValue)
+			giftShips[GameData::Ships().Get(child.Token(1))] = (child.Size() > 2 ? child.Token(2) : "");
```
Amazinite, line 198: `>=` is typically used over just `>` for checking the size of a DataNode.
```suggestion
			giftShips[GameData::Ships().Get(child.Token(1))] = child.Size() >= 3 ? child.Token(2) : "";
```

Amazinite (changes requested):
The new comments in PlayerInfo could be improved, and ReceiveShip should be moved in the .h file to somewhere more sensible than above the constructor and in the .cpp file to match its location in the .h file.

source/MissionAction.cpp
petervdmeer: Should we print a warning when trying to give a ship that doesn't exist in the GameData?

source/MissionAction.cpp
```diff
@@ -186,11 +194,13 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
...
 		else if(key == "outfit" && hasValue)
 		{
 			int count = (child.Size() < 3 ? 1 : static_cast<int>(child.Value(2)));
 			if(count)
-				gifts[GameData::Outfits().Get(child.Token(1))] = count;
+				giftOutfits[GameData::Outfits().Get(child.Token(1))] = count;
```
petervdmeer, line 203: Should we print a warning when trying to give an outfit that doesn't exist in the GameData?

source/PlayerInfo.cpp
```diff
@@ -841,14 +841,30 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
 
 
 
-// Buy a ship of the given model, and give it the given name.
-void PlayerInfo::BuyShip(const Ship *model, const string &name)
+// Charge the player the cost of a ship and give it the specified name:
+void PlayerInfo::BuyShip(const Ship *model, const std::string &name)
```
petervdmeer, line 845: Why adding "std::" here (and on line 853)? std should be covered by an using statement at the top of the class.

source/PlayerInfo.cpp
```diff
@@ -841,14 +841,30 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
+
+// Receive a ship as a gift
+void PlayerInfo::GiftShip(const Ship *model, const std::string &name)
+{
+	ReceiveShip(model, name, true);
+}
+
+
+
+// Purchase or receive a gift of a ship of the given model and name
```
Amazinite, line 860:
```suggestion
// Adds a ship of the given model with the given name to the player's fleet. If this ship is being gifted, it costs nothing and starts fully depreciated.
```

source/MissionAction.cpp
Gods-Righthand: Changed it in pursuit of a different problem, didn't know it was used more commonly this way.

source/PlayerInfo.cpp
```diff
@@ -841,14 +841,30 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
-void PlayerInfo::BuyShip(const Ship *model, const string &name)
+// Charge the player the cost of a ship and give it the specified name:
+void PlayerInfo::BuyShip(const Ship *model, const std::string &name)
+{
+	ReceiveShip(model, name, false);
+}
+
+
+
+// Receive a ship as a gift
```
Amazinite, line 852:
```suggestion
// Receive a ship of the give model with the given name as a gift.
```

source/PlayerInfo.cpp
```diff
@@ -841,14 +841,30 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
 
 
 
-// Buy a ship of the given model, and give it the given name.
-void PlayerInfo::BuyShip(const Ship *model, const string &name)
+// Charge the player the cost of a ship and give it the specified name:
```
Amazinite, line 844: This comment doesn't need to change.
```suggestion
// Buy a ship of the given model, and give it the given name.
```

source/PlayerInfo.cpp
```diff
@@ -841,14 +841,30 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
+// Purchase or receive a gift of a ship of the given model and name
+void PlayerInfo::ReceiveShip(const Ship *model, const string &name, bool gifting)
 {
 	if(!model)
 		return;
 	
 	int day = date.DaysSinceEpoch();
-	int64_t cost = stockDepreciation.Value(*model, day);
+	int64_t cost = gifting ? 0 : stockDepreciation.Value(*model, day);
 	if(accounts.Credits() >= cost)
```
petervdmeer, line 868: I believe giving a ship should still work if the player has negative credits. (Change the if-statement to what is is now, but add an or-statement that makes this true when gifting is true.)

source/PlayerInfo.h
```diff
@@ -247,6 +248,9 @@ class PlayerInfo {
 	
 	
 private:
+	// Internal implementation of BuyShip and GiftShip
+	void ReceiveShip(const Ship *model, const std::string &name, bool gifting);
+
```
Amazinite, line 253: Perhaps move this to above or below SelectShip since both functions relate to ships. I'd also change the comment to match the one in the .cpp file.

source/PlayerInfo.cpp
```diff
@@ -863,9 +879,12 @@ void PlayerInfo::BuyShip(const Ship *model, const string &name)
...
 		// Record the transfer of this ship in the depreciation and stock info.
-		depreciation.Buy(*model, day, &stockDepreciation);
-		for(const auto &it : model->Outfits())
-			stock[it.first] -= it.second;
+		if(!gifting)
+		{
+			depreciation.Buy(*model, day, &stockDepreciation);
+			for(const auto &it : model->Outfits())
+				stock[it.first] -= it.second;
+		}
```
petervdmeer, line 887: Are ships that are gifts valued at 100% of the deprecation value, or at 25% of the deprecation value? (Or at some other value.)

Amazinite (changes requested):

source/PlayerInfo.h
```diff
@@ -121,8 +121,9 @@ class PlayerInfo {
 	std::map<const std::shared_ptr<Ship>, std::vector<std::string>> FlightCheck() const;
 	// Add a captured ship to your fleet.
 	void AddShip(const std::shared_ptr<Ship> &ship);
-	// Buy or sell a ship.
+	// Buy a ship, receive a gifted ship, or sell a ship.
```
petervdmeer, line 124: Receive a ship given. (a gifted ship is a bit different than a ship that is given to the player).

source/PlayerInfo.cpp
```diff
@@ -841,14 +841,30 @@ void PlayerInfo::AddShip(const shared_ptr<Ship> &ship)
...
+// Charge the player the cost of a ship and give it the specified name:
+void PlayerInfo::BuyShip(const Ship *model, const std::string &name)
+{
+	ReceiveShip(model, name, false);
+}
+
+
+
+// Receive a ship as a gift
+void PlayerInfo::GiftShip(const Ship *model, const std::string &name)
```
petervdmeer, line 853: Do we need the 2 functions with the oneliners inside them? I think it looks nicer if we just keep using buyShip as function with a third parameter (isGift) which by default is false. You can give such default parameter in the header.

source/PlayerInfo.cpp
Gods-Righthand:
```suggestion
// Receive a ship of the given model with the given name as a gift.
```
the give model or given model?

source/MissionAction.cpp
Amazinite: Don't we already get an "(outfit | ship) X is referred to but never defined" warning?

source/PlayerInfo.cpp
Amazinite: If a ship or outfit how no recorded depreciation, it is fully depreciated.
https://github.com/endless-sky/endless-sky/blob/8f29a0b3d1401d4fd0f4de6e7433867c9dbe7710/source/Depreciation.cpp#L86-L87

source/MissionAction.cpp
petervdmeer: Good point! Yes that is probably handled by a generic check on empty ship definitions after loading.

source/MissionAction.cpp
petervdmeer: Same as above, probably already handled by a generic check.

source/PlayerInfo.cpp
Gods-Righthand:
```suggestion
void PlayerInfo::BuyShip(const Ship *model, const string &name)
```
Not sure, will remove it.

source/MissionAction.cpp
Gods-Righthand:
```suggestion
	void DoGiftShip(PlayerInfo &player, const Ship *model, const std::string &name, UI *ui)
```
Ok, will change it then.

source/PlayerInfo.cpp
Gods-Righthand: That adds a lot of clarity, so yes.

source/MissionAction.cpp
Gods-Righthand: So I don't need to worry about trying to add a checks here? Nice!

source/MissionAction.cpp
```diff
@@ -33,7 +33,15 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	void DoGift(PlayerInfo &player, const Outfit *outfit, int count, UI *ui)
+	void DoGiftShip(PlayerInfo &player, const Ship *model, const std::string &name, UI *ui)
+	{
+		player.GiftShip(modelName, name);
```
Amazinite, line 38: You also need to rename the variable here.
```suggestion
		player.GiftShip(model, name);
```

Amazinite (changes requested):

Amazinite (approved):

source/MissionAction.cpp
```diff
@@ -186,11 +196,13 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
+		else if(key ==  "ship" && hasValue)
```
tehhowch, line 199: `ship "some model"` is not a sufficient interface for this functionality.

Unlike outfits, there is not a simple way to reflect the direction of the transfer (e.g. by passing a negative count).

Please refactor this to account for  all of the desired ship transference mechanisms (e.g. give a ship, take a ship, has a ship).

I'm certain several content creators can work with you to determine a better, more flexible syntax.

tehhowch (changes requested):

source/MissionAction.cpp
Amazinite: `ship <action> "<model>" [<name>]`?

Gods-Righthand: That sounds like a resonable way to put it.

tehhowch: A drawback to that syntax is the inability to change the implementation to store instantiated ship objects without breaking compatibility.

I think using the action first allows us more flexibility for future improvements.

Amazinite: What?

Zitchas: So, something like this? (Based on Tehhowch's suggestion on discord
```
(give|take|require)
  ship <model name> [<name>] [<InvisibleIdentifier>] [<LockShip>] [<LockOutfits>]
  outfit <outfit name> [#count]
(give|take|require) [....inline specifier]
```
InvisibleIdentifier = identifier that is stored invisibly in the ship's attributes so that the player can change the ship's name and still be identified by the mission
LockShip = variable the specifies that the ship's name is locked. (recorded in the ship's attributes)
LockOutfit = variable that specifies that the ship's outfits cannot be modified (recorded in the ship's attributes)

Amazinite:
The syntax is now `give ship "<model>" ["<name>"]`, leaving it open to expansion to include `outfit`, uniquely defined ships, and more actions.

tehhowch (commented):
We could optionally add a comment in `CanBeDone` about parsing the gifted ship collection, but it's likely we will be changing things up anyway for the require/take scenarios anyway, so it isn't necessary to do so.

source/MissionAction.cpp
```diff
@@ -33,6 +33,16 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
 
 namespace {
+	void DoGift(PlayerInfo &player, const Ship *model, const string &name, UI *ui)
+	{
+		if(model->ModelName().empty())
+			return;
+		
+		player.BuyShip(model, name, true);
+		if(ui)
+			ui->Push(new Dialog("The " + name + " was added to your fleet!"));
```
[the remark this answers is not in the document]
Amazinite: Perhaps `“A <model>, the <name>, was added to your fleet.”`?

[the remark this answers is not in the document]
tehhowch: I think `The <model> <noun> "<name>" was added to your fleet.` or `You received the <model> <noun> "<name>."` would work.

To match outfit messaging, it could be `A/An <noun>, the <model> "<name>," was added to your fleet.`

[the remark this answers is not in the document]
Gods-Righthand: Based on community feedback on Discord as well, I went with `A/An <noun>, the <model>"<name>," was added to the fleet.`

source/MissionAction.cpp
```diff
@@ -33,6 +33,23 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+		player.BuyShip(model, name, true);
+		
+		string noun = model->Noun();
+		string message;
+		char c = tolower(noun.front());
+		bool isVowel = (c == 'a' || c == 'e' || c == 'i' || c == 'o' || c == 'u');
+		message = (isVowel ? "An " : "A ");
+		message += noun;
+		if(ui)
+			ui->Push(new Dialog(message + ", the " + model->ModelName() + " \"" + name + ",\" was added to your fleet!"));
```
Amazinite, line 50:
```suggestion
			ui->Push(new Dialog(message + ", the " + model->ModelName() + " \"" + name + ",\" was added to your fleet."));
```

tehhowch (commented):
For outfits, a dialog is only displayed for an outfit that could not be installed, and in all cases a message is printed in the "chat" log.

In contrast, ship gifting as proposed here will always trigger a dialog per ship. Is giving the player a ship of such importance that a dialog should be displayed for each ship given, **_in addition to_** whatever existing conversation or dialog text is already associated with the given action? Or should the "give" case be only recorded in the message log, and the mission writer expected to have used normal expository methods to let the player know that they got something?

Amazinite:
I'd say the latter.

Zitchas:
I agree with the later. I'd say using this falls into the same category as the payment mechanism: We *can* give people money without mentioning it in conversation, but we generally should tell them about it.

(I"m just imagining a situation where the player goes into the starport, then leaves and they've got this new ship sitting in the shipyard. "Who are you?" "I'm the pilot of your new ship, Sir!" "Where did you come from?" "I just got hired this morning to fly this new ship over to the landing pad here for the new owner, which according to these records, is you, as of today." Player: ???? )

That could be a funny way to start a mission sequence.

source/MissionAction.cpp
```diff
@@ -33,6 +33,15 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
+	void DoGift(PlayerInfo &player, const Ship *model, const string &name, UI *ui)
+	{
+		if(model->ModelName().empty())
+			return;
+		
+		player.BuyShip(model, name, true);
+		Messages::Add("The " + model->ModelName() + " \"" + name + ",\" was added to your fleet.");
```
tehhowch, line 42: Either a comma both before & after the name, or no comma at all

source/MissionAction.cpp
```diff
@@ -186,11 +195,18 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
+		else if(key == "give" && hasValue)
+		{
+			if(child.Token(1) == "ship" && child.Size() >= 3)
+				giftShips[GameData::Ships().Get(child.Token(2))] = child.Size() >= 4 ? child.Token(3) : "";
+			else
+				child.PrintTrace("Skipping unsupported give syntax:");
```
tehhowch, line 203:
```suggestion
				child.PrintTrace("Skipping unsupported \"give\" syntax:");
```

tehhowch (commented):
Fix these and it'll be in good shape

source/MissionAction.cpp
Gods-Righthand:
```suggestion
		Messages::Add("The " + model->ModelName() + " \"" + name + "\" was added to your fleet.");
```
Okay, whatever.

tehhowch (approved):

Zitchas:
Question: Is the example file and code in the initial post accurate as to the final formatting we should be using?

Amazinite:
The usage example in the PR description is accurate.

Zitchas:
````
        give ship Sparrow "Mote of Might"
        give ship "Star Barge"
````

Just to clarify, the two options are "give ship with a specified name" and "give a ship with a random name", correct? Neither option allows the player to pick a name on the spot?

Amazinite:
Correct.

Zitchas:
OK. It would be nice to have that functionality, but that's a luxury, not a necessity. The player can manually change the name later. 

Thanks to everyone for working on this and getting it implemented. 

Just for reference, putting it into use: https://github.com/endless-sky/endless-sky/pull/5388

Issue #5171, opened 2020-07-24
Easier Save Game Comparisons


jchamlin:
**Is your feature request related to a problem? Please describe.**

When I am attempting to see what has changed in a save game file, it is hard to compare them in a compare tool like Beyond Compare, because some of the sections are written in non-deterministic order. For example, ship outfits, explosions, the visited entries, the visited planet entries, and the harvested entries. Some sections are deterministic (sorted, or in some order that is the same every time), like the order of the ships in the fleet, the missions, the conditions, the log, etc and it makes those sections easier to compare.

Also, the missions change quite frequently and there can be large changes to those, and it is hard sometimes for a diff tool to resync after such a large amount of change, so it would be better to put the jobs/missions at the end of the file.

**Related Issue Links**

I'm not sure there is a request for this anywhere else.

**Describe the solution you'd like**

Sort or write out in a deterministic manner all entries of all sections of the save game file, so when you bring it up in a comparison tool and compare to a different save game file, the tool shows you only exactly what is changed. There will still be some noise (like ship positions), but that can be ignored by a rule.

**Describe alternatives you've considered**

Just comparing them and trying hard to ignore the noise caused by the somewhat random order some sections are written out in.

**Additional context**

Here is a screenshot of Beyond Compare showing an example of the conditions (nicely sorted and easy to compare) and ship outfits (which are identical but written out in a different ourder).

![Endless Sky Save File Compare Issue](https://user-images.githubusercontent.com/1820516/88397184-357fc100-cd89-11ea-9453-59a2df358c1a.png)

Zitchas:
I can't say that I've ever compared the Archon outfits, but it does seem weird that they would be saved in different orders in different games. It is the same ship created by the same mission code, after all.

jchamlin:
Internally the Ship.cpp is using a map object, and the default version of that map doesn't have any deterministic ordering. I'm going to see if there's any way to either use a map with deterministic iteration order (so the order the outfits are read in / added, is the same they are iterated in when writing them back out), or at least make sure when the map is iterated for output, the items are written out sorted by name.

tehhowch:
@jchamlin is this resolved now?

petervdmeer:
jchamlin mentioned outfits, explosion effects, visited systems, visited planets and harvested items as examples that showed up in various orders, so I expect that the issue is not solved for jchamlin yet.

PR #5319, opened 2020-09-11
Save ship outfits in alphabetic order to savegame.

petervdmeer:
**Feature:** This PR implements part of the feature request detailed and discussed in issue #5171

## Feature Details
This PR adds a temporary map in the function that saves the ship-outfits to a save-game file to ensure that the outfits are saved in alphabetical order.

## UI Screenshots
N/A

## Usage Examples
N/A, the game automatically applies the ordering when saving.

## Testing Done
I started the game with an existing save-game, pressed depart and then closed the game. The outfits in the save-game before and after starting where the same (but the outfits after starting the game were sorted in alphabetical order).

## Performance Impact
The effect of this change is that the function that saves the game takes a little bit longer time and temporary uses a little bit more memory (since it needs to build up the temporary map). But ships have small lists of outfits, so the impact should be very minimal.
Saving the game is not performance critical code.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -715,13 +715,13 @@ void Ship::Save(DataWriter &out) const
 			for(const auto &it : outfits)
 				if(it.first && it.second)
 					orderedOutfits[it.first->Name()] = it.second;
-			for(const auto &ito : orderedOutfits)
-				{
-					if(ito.second == 1)
-						out.Write(ito.first);
-					else
-						out.Write(ito.first, ito.second);
-				}
+			for(const auto &it : orderedOutfits)
+			{
+				if(it.second == 1)
+					out.Write(it.first);
+				else
+					out.Write(it.first, it.second);
+			}
 		}
 		out.EndChild();
 		
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -704,13 +704,23 @@ void Ship::Save(DataWriter &out) const
 		out.Write("outfits");
 		out.BeginChild();
 		{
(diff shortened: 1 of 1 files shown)


source/Ship.cpp
```diff
@@ -704,13 +704,23 @@ void Ship::Save(DataWriter &out) const
...
+			// The game-engine doesn't care in which order outfits are
+			// saved or loaded, the sorting here is done as a service for
+			// content creators that use savegames for working on their
+			// new content.
+			map<string, int> orderedOutfits;
 			for(const auto &it : outfits)
 				if(it.first && it.second)
+					orderedOutfits[it.first->Name()] = it.second;
+			for(const auto &ito : orderedOutfits)
 				{
```
tehhowch, line 719: Braces are over-indented

source/Ship.cpp
```diff
@@ -704,13 +704,23 @@ void Ship::Save(DataWriter &out) const
...
-					if(it.second == 1)
-						out.Write(it.first->Name());
+					if(ito.second == 1)
```
tehhowch, line 720: No need to change the local variable name from the common `it` to `ito`

source/Ship.cpp
petervdmeer: Sorry, was an editting issue. Fixed now.

source/Ship.cpp
petervdmeer: I thought it was more clear for somebody reading the code, but it indeed doesn't really make a difference. Changed back to it.

PR #5321, opened 2020-09-11
Fix: Tweaking Puffin-Drak conflict

Zitchas:

Changes:
- Drak draining field force 50 -> 100
- Drak Turret velocity and lifetime swapped. So same effective range, but a bit slower.
- Puffin mass dropped from 44 to 40.

## Testing Done
Repeatedly running the Nenia puffin dodging mission.

## Results
Things are back to the Puffin getting shoved, although not as dramatically as it was when the force was 500 and working properly. The puffin usually gets hit enough times to build up some graphical indicators, but it is enough to outpace most of the turret shots. I've only tested it a dozen times or so, but it seems to work fairly well at  being exciting and stressful, but not dying. So, this is what I'm aiming for.

## Alternative

Alternatively, we could leave it at force 50, and drop the puffin mass some more. I don't know for sure, but dropping it to somewhere close to 30 would probably work.

### DIFF
--- a/data/drak/drak.txt
+++ b/data/drak/drak.txt
@@ -53,7 +53,7 @@ outfit "Drak Draining Field"
 		"firing energy" 20
 		"firing heat" 5
 		"firing force" 30
-		"hit force" 50
+		"hit force" 100
 		"shield damage" 50
 		"ion damage" .5
 		"disruption damage" 1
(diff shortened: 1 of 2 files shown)


Zitchas:
Doing a whole bunch more testing with the puffin, and this change seems to bring it back in line with what the goal is. I'm finding that most take-offs from Nasqueron (which is closer to system center and the one that seems to have the Archon start off closer to it, and thus be more problematic) now tend to have 2-4 hits from the turret before my ship is pushed away. The push away gives me enough time to get my bearings, identify where the Archon is, and start dodging. At the same time, the damage from the turrets does consistently give me a "escaped by the skin of my teeth" sort of feeling.

PR #5326, opened 2020-09-12
Arkhne's Unfettered First Contact Update

arkhne:
First Contact with Unfettered Update

In regards to endless-sky#2664 , I have modified the Hai missions to provide two different alternative versions for the First Contact mission to the Unfettered. This causes minimal changes and disruption. Existing saves that have already made contact should be unaffected, and existing files referencing the old mission tag will be unaffected, as I have designed this to use the existing tag.

I have mostly just copy pasted and rearranged the conversation in question to attempt to maintain the writing style in question, and not have the mission feel out of place to those who have done it previously.

This should be relatively easy to test, but if needed, I can provide a save landed on the Wanderer home world without contact with the Hai. The dual-mission structure was also chosen to make testing it easier, the renamed version of the original mission has no changes at all.

If you want me to take a closer look at the Hai side of First Contact, and followup mission (should not be needed), I can do so, I just wanted to get the ball rolling by addressing the issue directly, but further changes are definitely possible.

Please feel free to tell me what I have done wrong, not just in the work, but in making PRs and stuff, I am a complete and utter GitHub Newbie.

### DIFF
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -146,7 +146,7 @@ mission "First Contact: Unfettered 1"
 				`	"I'm sorry, I don't know the secret of the jump drive. No human being does."`
 				`	"If I ever have an extra jump drive to sell, I will bring it to you."`
 					goto sell
-			`	"Then leave us alone," it says, and it walks off.`
+			`	"Then leave us alone," it says, and it walks off."`
 				decline
 			
 			label sell
@@ -230,7 +230,7 @@ mission "First Contact: Unfettered 2"
 				`	"I'm sorry, I don't know the secret of the jump drive. No human being does."`
 				`	"If I ever have an extra jump drive to sell, I will bring it to you."`
 					goto sell
-			`	"Then leave us alone," it says, and it walks off.`
+			`	"Then leave us alone," it says, and it walks off."`
 				decline
 			
 			label sell
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -75,10 +75,96 @@ mission "Assisting Hai"
 
 
 
-mission "First Contact: Unfettered"
+mission "First Contact: Unfettered 1"
 	landing
 	source
 		attributes "unfettered"
+	to offer
+		not "First Contact: Hai: offered"
+	on offer
+		conversation
+			`This planet is populated by an alien species that resembles giant, intelligent squirrels. Do you want to approach one of them?`
+			choice
+				`	(Sure.)`
+				`	(Not now, it's too risky.)`
+					defer
+			`	You catch the eye of one of the Hai dock workers and it walks over. "Ah," it says, "the monkey is curious. The monkey has never been face to face with a true Hai before?"`
+			choice
+				`	"Hai? Is that what you call yourselves?"`
+					goto name
+				`	"What do you mean, 'true Hai'?"`
+					goto true
+				`	"Why did you shoot at me? Are you at war with everyone?"`
+					goto everyone
+			
+			label name
+			`	We are Hai. Hai the conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."
+				goto masters
+			
+			label true
+			`	"It means that we are unaltered, Hai as Hai were first born to be. Hai the conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."`
+				goto masters
+			
+			label everyone
+			`	"We fight to defend ourselves from extinction," it says.`
+			`	"'Extinction?'" you ask. "Do you mean that there are no other Hai?"`
+			`	"We are all that is left of the original Hai." it says, "Unaltered. The conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."`
+				goto masters
+			
+			label masters
+			choice
+				`	"Where was your territory? Why have I never heard of you before?"`
+				`	"What do you mean, the Hai were altered?"`
+	
+			`	"This is our story," it says. "Once, many worlds beyond wormhole were ours. Once, hundred thousand years now gone. We needed more worlds. A species that does not expand becomes like brackish water, like stunted tree. But dragonfolk in the south had created hyperdrive, and human monkeys had begun banging rocks together in what the Drak said was very intelligent fashion. So we were forbidden from taking more worlds."`
+			`	"'Dragonfolk?'" you ask. "I've never heard of them."`
+			`	"Sheragi," it says, "they are extinct. Only their dumb ancestors survive, just as monkeys will outlast the thinking humans."`
+			`	"How were you altered?" you ask.`
+			`	It says, "We Hai took worlds that Drak said belonged to Sheragi, although Sheragi had not yet even discovered them. Drak retaliated. On all Hai worlds, Drak ships appeared in orbit. Sickness swept each planet: quick fever, strange feeling of frailness, like you are brittle bones with mouth full of dirt. Victims left alive, but all ambition gone, no desire but to die of comfortable old age."`
+			choice
+				`	"Did you try to resist the Drak?"`
+					goto resist
+				`	"How do you know this, if it happened a hundred thousand years ago?"`
+					goto know
+			
+			label resist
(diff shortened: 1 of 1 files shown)


arkhne:
Ok, what did I do wrong? GitHub is showing unpleasant red X and apparently some checks failed?

data/hai/hai missions.txt
```diff
@@ -144,7 +230,7 @@ mission "First Contact: Unfettered"
 				`	"I'm sorry, I don't know the secret of the jump drive. No human being does."`
 				`	"If I ever have an extra jump drive to sell, I will bring it to you."`
 					goto sell
-			`	"Then leave us alone," it says, and it walks off.`
+			`	"Then leave us alone," it says, and it walks off."`
```
Amazinite, line 233: Not sure why you've added this. (This same error exists in the new conversation.)
```suggestion
			`	"Then leave us alone," it says, and it walks off.`
```

data/hai/hai missions.txt
```diff
@@ -157,6 +243,17 @@ mission "First Contact: Unfettered"
 
 
 
+mission "First Contact: Unfettered"
```
Amazinite, line 246: This mission will become unnecessary when using a branch, but in the future know that you shouldn't create dummy missions that do absolutely nothing but fail. Instead, you would do the following for the previous missions:
```
on decline
	set "First Contact: Unfettered: offered"
```

This avoids the need of a dummy missions.

Amazinite (changes requested):
This should use a conversation branch instead of two missions in order to avoid a large amount of duplicate text.

data/hai/hai missions.txt
arkhne: That was one of the first things I tried, but it failed. I'm sure I did something wrong, though.

data/hai/hai missions.txt
arkhne: Copy paste error, I've just been working through these, since they weren't flagged until I tried to PR them here. Sorry.

arkhne:
Sorry for making a huge mess of things like this. I have to cook dinner right now, but I'll try restructuring it into a branch after that, the quick and easy testing served its purpose though, considering how many times I had to test this (the rust was real).

arkhne:
Ok, I managed to fix it up beforehand, but I really need to go now.

arkhne:
Ok, changed to branch structure, and fixed the failed checks (again).

data/hai/hai missions.txt
```diff
@@ -81,6 +81,24 @@ mission "First Contact: Unfettered"
...
 	on offer
 		conversation
+			branch hai
+				has "First Contact: Hai: offered"
+				
+			`This planet is populated by an alien species that resembles giant, intelligent squirrels. Do you want to approach one of them?`
+			choice
+				`	(Sure.)`
+				`	(Not now, it's too risky.)`
+					defer
```
Amazinite, line 91: This "too risky" choice doesn't really make sense in this context. The player knows nothing of the Hai at this point. Why would it be risky to talk to these Hai and not the other Hai (as the choice for the Hai first contact just has a "not now" option).

Since the player at least knows that the Unfettered are hostile, perhaps mention that in the introduction.
```suggestion
			`This planet is populated by a hostile alien species that resembles giant, intelligent squirrels. Do you want to approach one of them?`
			choice
				`	(Sure.)`
				`	(Not now, it's too risky.)`
					defer
```

data/hai/hai missions.txt
```diff
@@ -81,6 +81,24 @@ mission "First Contact: Unfettered"
...
+			`	You catch the eye of one of the Hai dock workers and it walks over. "Ah," it says, "the monkey is curious. The monkey has never been face to face with a true Hai before?"`
```
Amazinite, line 92: The player does not know at this point that these aliens are named the Hai.
```suggestion
			`	You catch the eye of one of the alien dock workers and it walks over. "Ah," it says, "the monkey is curious. The monkey has never been face to face with a true Hai before?"`
```

data/hai/hai missions.txt
```diff
@@ -93,7 +111,7 @@ mission "First Contact: Unfettered"
 				`	"What do you mean, 'true Hai'?"`
 					goto true
 				`	"Why are you at war with everyone else?"`
-					goto everyone
+					goto evelse
```
Amazinite, line 114: Labels should still be words. To avoid duplicate labels, I'd do `everyone1` and `everyone2` instead of `everyone` and `evelse`, or perhaps `everyone` for one of the labels and `war` for another.

data/hai/hai missions.txt
```diff
@@ -103,8 +121,18 @@ mission "First Contact: Unfettered"
...
 				goto masters
 			
+			label name
+			`	"We are Hai. Hai the conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."`
+				goto masters
+			
 			label everyone
 			`	"We fight to defend ourselves from extinction," it says.`
+			`	"'Extinction?'" you ask. "Do you mean that there are no other Hai?"`
+			`	"We are all that is left of the original Hai." it says, "Unaltered. The conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."`
```
Amazinite, line 131:
```suggestion
			`	"We are all that is left of the original Hai," it says. "Unaltered. The conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."`
```

data/hai/hai missions.txt
```diff
@@ -103,8 +121,18 @@ mission "First Contact: Unfettered"
 			`	"It means that we are unaltered, Hai as Hai were first born to be. Hai the conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."`
 				goto masters
 			
+			label name
+			`	"We are Hai. Hai the conquerors. We were masters of a hundred worlds before human monkeys had learned the use of thumbs."`
+				goto masters
+			
 			label everyone
 			`	"We fight to defend ourselves from extinction," it says.`
+			`	"'Extinction?'" you ask. "Do you mean that there are no other Hai?"`
```
Amazinite, line 130: While we're here, `"'Extinction?'"` should be replaced with either `"Extinction?"` (dropping the single quotes) or `"'Extinction'?"` (moving the question mark outside of the single quotes) in both this new section and where it had previously been.

Amazinite (changes requested):
"First Contact: Hai" should also change in some way according to which first contact you've done first. (I can take care of that if you'd like, though.)

data/hai/hai missions.txt
arkhne: I like that change, lets go with that.

data/hai/hai missions.txt
arkhne: Yeah, I agree. I realized afterwards that this (and one of the Unfettered lines) should be changed a little, I figured it was best to wait until a second pass, instead of keep spamming changes.

arkhne: Actually, it was the Unfettered first version of this that I intended to change, so the original will revert to Everyone, and the new version will just be the player asking if they are at war.

data/hai/hai missions.txt
arkhne: I think we should remove the single quotes.

arkhne:
Oh man, I feel so dumb. I totally should've caught most of those myself when I was testing.

I'm totally willing to go over the Hai themselves (and would've done so more closely than this), since I agree with you that we should probably go over that. However, I'm not a big Hai guy, so I'd mostly try to stick to the originals as closely as possible, and not flex my creativity, Lore of anything but the Remnant isn't really my strong suit, so it wouldn't be a total re-write.

 Two things come to mind though.

If we go over First Contact with the Hai AFTER the Unfettered, should we merge the Asking the Hai about the Unfettered into the First Contact, then auto-fail or exclude the asking the Hai about the Unfettered, because it would feel pretty silly to NOT handle the topics of the asking the Hai mission when bringing that up.

Secondly, the point of stuff like this was to have you delegate, whoever does it, it shouldn't be you, this has already taken enough of your time.

Amazinite (approved):

PR #5327, opened 2020-09-12
Mebulmem fix

arkhne:
As per Arachi's request, all instances of Membulem have been changed to match the planet name of 'Corral of Mebulmem'.

A quick search of other files did not turn up any more results, and as such, this should be the only file that needs changing.


arkhne:
Alright, Can somebody please teach me how to do this properly? I feel like everytime I try to get it to go nicely adds unneccesary things.

PR #5332, opened 2020-09-13
Fix typo

tatami4:


Amazinite:
I get the sense that this is intentional. "Can't hardly" taken literally is a double negative, meaning that they'd be saying that they can make enough money, but taken as a figure of speech, the double negative doesn't cancel itself out but instead emphasizes the negative nature of the situation. (Such use of double negatives to emphasize the negative instead of canceling it out is fairly common in my experience, at least around me.)

So basically, while grammatically incorrect, it does make sense as something that someone could say, which is what is happening here.

tatami4:
Ok

PR #5324, opened 2020-09-12
feat(mechanics): Save ship explosion effects in alphabetic order to savegame.

petervdmeer:
**Feature:** This PR implements part of the feature request detailed and discussed in issue #5171

## Feature Details
This PR adds a temporary map in the function that saves the ship-explosion-effects to a save-game file to ensure that the explosion-effectss are saved in alphabetical order.

## UI Screenshots
N/A

## Usage Examples
N/A, the game automatically applies the ordering when saving.

## Testing Done
I started the game with an existing save-game, pressed depart and then closed the game. The explosion effects in the save-game before and after starting where the same (but the effects after starting were sorted in alphabetical order).

## Performance Impact
The function that saves the game will take a little bit more time and temporary uses a little bit more memory (since it needs to build up the temporary maps). But ships have small lists of explosion effects, so the impact should be very minimal.
Saving the game is not performance critical code.

PR #5333, opened 2020-09-13
feat(mechanics): Save collections in order by element name

tehhowch:
**Feature:** This PR implements the feature request detailed and discussed in issue #5171 
closes #5324 

## Feature Details
Where parts of the player's save were not deterministically ordered, they now are. A template method, and a specialization for uses of std::map for a pointer type key, are used to create a vector of pointers to the collection elements, which is then sorted and iterated.

To avoid future code from manipulating DataWriters incorrectly, deletes the copy & move constructors / operators for it.

## UI Screenshots
n/a''

## Usage Examples
n/a

## Testing Done
Saved & loaded & saved again a couple times

## Performance Impact
n/a

### DIFF
--- a/source/Depreciation.cpp
+++ b/source/Depreciation.cpp
@@ -79,7 +79,8 @@ void Depreciation::Save(DataWriter &out, int day) const
 	{
 		using ShipElement = pair<const Ship *const, map<int, int>>;
 		WriteSorted(ships,
-			[](const ShipElement *lhs, const ShipElement *rhs) { return lhs->first->ModelName() < rhs->first->ModelName(); },
+			[](const ShipElement *lhs, const ShipElement *rhs)
+				{ return lhs->first->ModelName() < rhs->first->ModelName(); },
 			[=, &out](const ShipElement &sit)
 			{
 				out.Write("ship", sit.first->ModelName());
@@ -97,7 +98,8 @@ void Depreciation::Save(DataWriter &out, int day) const
 			});
 		using OutfitElement = pair<const Outfit *const, map<int, int>>;
 		WriteSorted(outfits,
-			[](const OutfitElement *lhs, const OutfitElement *rhs) { return lhs->first->Name() < rhs->first->Name(); },
+			[](const OutfitElement *lhs, const OutfitElement *rhs)
+				{ return lhs->first->Name() < rhs->first->Name(); },
 			[=, &out](const OutfitElement &oit)
 			{
 				out.Write("outfit", oit.first->Name());
--- a/source/Depreciation.cpp
+++ b/source/Depreciation.cpp
@@ -77,32 +77,38 @@ void Depreciation::Save(DataWriter &out, int day) const
 	out.Write(NAME[isStock]);
 	out.BeginChild();
 	{
-		for(const auto &sit : ships)
-		{
-			out.Write("ship", sit.first->ModelName());
-			out.BeginChild();
+		using ShipElement = pair<const Ship *const, map<int, int>>;
+		WriteSorted(out, ships,
+			[](const ShipElement *lhs, const ShipElement *rhs) { return lhs->first->ModelName() < rhs->first->ModelName(); },
+			[=](DataWriter &dw, const ShipElement &sit)
 			{
-				// If this is a planet's stock, remember how many outfits in
-				// stock are fully depreciated. If it's the player's stock,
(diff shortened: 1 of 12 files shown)


source/Ship.cpp
```diff
@@ -704,24 +704,15 @@ void Ship::Save(DataWriter &out) const
...
-			{
-				if(it.second == 1)
-					out.Write(it.first);
-				else
-					out.Write(it.first, it.second);
-			}
+			using OutfitElement = pair<const Outfit *const, int>;
+			WriteSorted(outfits,
+				[](const OutfitElement *lhs, const OutfitElement *rhs){ return lhs->first->Name() < rhs->first->Name(); },
+				[&out](const pair<const Outfit *, int> &it){
```
petervdmeer, line 710: Is it possible to use OutfitElement here as well instead of the pair(Outfit, int>?

source/Ship.cpp
```diff
@@ -805,12 +796,22 @@ void Ship::Save(DataWriter &out) const
 		}
 		for(const Leak &leak : leaks)
 			out.Write("leak", leak.effect->Name(), leak.openPeriod, leak.closePeriod);
-		for(const auto &it : explosionEffects)
+		
+		using EffectElement = pair<const Effect *const, int>;
+		auto effectSort = [](const EffectElement *lhs, const EffectElement *rhs)
+		{
```
petervdmeer, line 802: The code `if(it.first && it.second)` on line 809/807 suggests that `it.first` might be a null-pointer, but it looks like the check on `it.first` on line 809/807 itself is not actually needed).

petervdmeer (approved):
Wow, this looks quite generic indeed. Good change!
(I have 2 minor remarks, but they are not blocking for merging this.)

source/GameData.cpp
```diff
@@ -458,9 +458,14 @@ void GameData::WriteEconomy(DataWriter &out)
 		{
 			out.Write("purchases");
 			out.BeginChild();
-			for(const auto &pit : purchases)
-				for(const auto &cit : pit.second)
-					out.Write(pit.first->Name(), cit.first, cit.second);
+			using Purchase = pair<const System * const, map<string, int>>;
```
Amazinite, line 461: Other places where you've done this don't have the space here.
```suggestion
			using Purchase = pair<const System *const, map<string, int>>;
```

source/GameData.cpp
```diff
@@ -458,9 +458,14 @@ void GameData::WriteEconomy(DataWriter &out)
...
+			WriteSorted(purchases,
+				[](const Purchase *lhs, const Purchase *rhs) { return lhs->first->Name() < rhs->first->Name();},
```
Amazinite, line 463:
```suggestion
				[](const Purchase *lhs, const Purchase *rhs) { return lhs->first->Name() < rhs->first->Name(); },
```

source/Depreciation.cpp
```diff
@@ -77,32 +77,38 @@ void Depreciation::Save(DataWriter &out, int day) const
 	out.Write(NAME[isStock]);
 	out.BeginChild();
 	{
-		for(const auto &sit : ships)
-		{
-			out.Write("ship", sit.first->ModelName());
-			out.BeginChild();
+		using ShipElement = pair<const Ship *const, map<int, int>>;
+		WriteSorted(ships,
+			[](const ShipElement *lhs, const ShipElement *rhs) { return lhs->first->ModelName() < rhs->first->ModelName(); },
```
Amazinite, line 82: Should we spread these functions out on multiple lines for readability? It's not as if we only have so many lines to go around.
```suggestion
			[](const ShipElement *lhs, const ShipElement *rhs)
			{
				return lhs->first->ModelName() < rhs->first->ModelName(); 
			},
```

Amazinite (changes requested):
Still trying to wrap my head around this. I vaguely understand what is going on (we have a generic WriteSorted function that takes in a container of some type as well as two functions that define how to sort the container then write the sorted container, then each point where we need to write something sorted we create the functions that correspond to how we want to sort and write), but I still want to nail down my understanding on the syntax used. (I know these... things... have popped up in the past, so it seems like something worth learning about when basically this entire PR is built around using lambda expressions.) I'm sure everything works as is, though. The only comments I can really give right now are on formatting.

Amazinite (approved):
Thank you cd_macos_x86_64, very cool.

Issue #4945, opened 2020-03-24
conditional spaceport news


starmlerp:
**Is your feature request related to a problem? Please describe.**
Currently, spaceport news will only display depending on the government of the planet and the attributes of the speaker. 

This feature request is about adding the possibility of checking for other conditional attributes, such as the players combat reputation, cargo capacity, completed missions etc..

**Related Issue Links**
brought up in #4943
could possibly aid in #4907

**Describe the solution you'd like**
Syntax I would suggest would be similar to the way missions are currently offered. For example:

```
news "human merchant anywhere warning"
	location
		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Independent" "Hai"
	name
		word
			"Merchant captain"
        to show
                "combat rating" > 60
	portrait
		portrait/human02
		portrait/...
	message
		word
			"You might want to watch your back, pal. With such a reputation, you are sure to have quite a few bounty hunters chasing you around"
```

the news dialog above would only display if the players combat rating is above 60 (i am not sure if that is the point at which bounty hunters will spawn, but it is irrelevant at this point)

downRightCoder256:
Does anyone know how code highlighting works in github?

tehhowch:
This should be a pretty simple feature we can add in 0.9.13.

MCOfficer:
> Does anyone know how code highlighting works in github?

https://help.github.com/en/github/writing-on-github/creating-and-highlighting-code-blocks

AlbertNewton:
I used to think that events could change News, but then I checked and as of five months ago they don't seem to.

tehhowch:
@AlbertNewton

PR #5334, opened 2020-09-13
Conditional spaceport news

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4945.

## Feature Details
Adds a new node to news, `to show`. This is a condition set like `to offer` that must be passed in order for the news item to show, in addition to the location filter matching.

## UI Screenshots
N/A

## Usage Examples
```
news "reminiscing about the war"
	...
	to show
		has "main plot completed"
	...
```

## Testing Done
Created three news entries. One had no show conditions. The second was set to only show before the main plot was completed. The third was set to only show after the main plot was completed. All three news entries showed under the proper conditions.

## Performance Impact
N/A

### DIFF
--- a/source/News.cpp
+++ b/source/News.cpp
@@ -13,7 +13,6 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "News.h"
 
 #include "DataNode.h"
-#include "PlayerInfo.h"
 #include "Random.h"
 #include "SpriteSet.h"
 
@@ -86,7 +85,7 @@ void News::Load(const DataNode &node)
 				messages.Load(child);
 		}
 		else if(tag == "to" && hasValue)
-		{		
+		{
 			if(child.Token(valueIndex) == "show")
 			{
 				if(remove)
--- a/source/News.cpp
+++ b/source/News.cpp
@@ -84,6 +84,19 @@ void News::Load(const DataNode &node)
 			else
 				messages.Load(child);
 		}
+		// Handle the attributes which cannot be "removed" or "added."
+		else if(add || remove)
+		{
+			child.PrintTrace("Cannot \"add\" or \"remove\" a specific value from the given key:");
+			continue;
+		}
+		else if(tag == "to" && hasValue)
+		{
+			if(child.Token(valueIndex) == "show")
+				toShow.Load(child);
+			else
+				child.PrintTrace("Unrecognized news attribute:");
+		}
 		else
 			child.PrintTrace("Unrecognized news attribute:");
 	}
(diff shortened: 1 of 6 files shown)


source/News.cpp
```diff
@@ -98,14 +111,14 @@ bool News::IsEmpty() const
...
-bool News::Matches(const Planet *planet) const
+// Check if this news item is available given the player's planet and conditions.
+bool News::Matches(const Planet *planet, const PlayerInfo &player) const
 {
 	// If no location filter is specified, it should never match. This can be
 	// used to create news items that are never shown until an event "activates"
 	// them by specifying their location.
 	// Similarly, by updating a news item with "remove location", it can be deactivated.
-	return location.IsEmpty() ? false : location.Matches(planet);
+	return location.IsEmpty() ? false : (location.Matches(planet) && toShow.Test(player.Conditions()));
```
Amazinite, line 121: Should we just pass the `player` all the way from SpaceportPanel to here, then use `player.GetPlanet()` instead of passing the player reference and the planet pointer at the same time?

source/News.h
```diff
@@ -13,8 +13,10 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #ifndef NEWS_H_
 #define NEWS_H_
 
+#include "ConditionSet.h"
 #include "LocationFilter.h"
 #include "Phrase.h"
+#include "PlayerInfo.h"
```
tehhowch, line 19: This should be a forward declaration.

source/SpaceportPanel.cpp
```diff
@@ -48,7 +48,7 @@ SpaceportPanel::SpaceportPanel(PlayerInfo &player)
 
 void SpaceportPanel::UpdateNews()
 {
-	const News *news = GameData::PickNews(player.GetPlanet());
+	const News *news = GameData::PickNews(player.GetPlanet(), player);
```
tehhowch, line 51: I'm up for moving the PickNews method to SpaceportPanel, and making GameData return the collection of known News. Then we don't need to pass the player object around as much.

tehhowch (commented):
I'm fine with allowing the news to not always be present, i.e. the picked news still always is relevant to the planet, but might not satisfy the display conditions.

Perhaps that is not a good plan, though, as more conditional news is added, the probability of "valid" news being picked will decrease. (Or, we make it such that when no news is currently displayed, the next keypress for news considers only news that matches and can be displayed.)

source/News.h
```diff
@@ -33,8 +35,8 @@ class News {
 	
 	// Check whether this news item has anything to say.
 	bool IsEmpty() const;
-	// Check if this news item is available on the given planet.
-	bool Matches(const Planet *planet) const;
+	// Check if this news item is available given the player's planet and conditions.
+	bool Matches(const PlayerInfo &player) const;
```
tehhowch, line 39: Let's actually go one better and pass only the planet and conditions, rather than the player.

tehhowch (commented):
I would expect a plugin author to be able to declare that their conditions for displaying a particular news item are the only such conditions (i.e. support removing them).

source/News.h
```diff
@@ -21,6 +22,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 class DataNode;
 class Planet;
+class PlayerInfo;
```
tehhowch, line 25:
```suggestion
```

tehhowch (approved):
Cleanup tweaks and then :shipit:

PR #5335, opened 2020-09-14
Support data files encoded by UTF-8.

oo13:
**Feature:** This is a part of #4123 (for now) and needs no extra libraries.

## Feature Details
This PR allows Endless Sky to read a data file contains a character encoded by UTF-8.

The code originates from #3312 and it is written by flaviojs.

## UI Screenshots
N/A

## Usage Examples
[test-data.zip](https://github.com/endless-sky/endless-sky/files/5217965/test-data.zip)
This plugin introduces the government β that has the display name "Beta."

```
government β
	"display name" "Beta"
```

## Testing Done
The current master detects this when reading the plugin:

```
Skipping unrecognized root object:
file ${HOME}/.local/share/endless-sky/plugins/test-for-datafile-encoded-utf-8/data/governments.txt
L11:   government
```

This PR version raises no error.

## Performance Impact
N/A

### DIFF
--- a/copyright
+++ b/copyright
@@ -814,10 +814,6 @@ Copyright: Tommaso Becca
 License: CC-BY-SA-4.0
 Comment: Derived from works by ESA/Hubble & NASA (under the same license)
 
-Files: source/Utf8.*
-Copyright: 2017, 2018 by Flávio J. Saraiva
-License: GPL-3+
-
 License: GPL-3+
  This program is free software: you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
--- a/copyright
+++ b/copyright
@@ -814,6 +814,10 @@ Copyright: Tommaso Becca
 License: CC-BY-SA-4.0
 Comment: Derived from works by ESA/Hubble & NASA (under the same license)
 
+Files: source/Utf8.*
+Copyright: 2017, 2018 by Flávio J. Saraiva
+License: GPL-3+
+
 License: GPL-3+
  This program is free software: you can redistribute it and/or modify
  it under the terms of the GNU General Public License as published by
--- a/source/DataFile.cpp
+++ b/source/DataFile.cpp
@@ -106,9 +106,10 @@ void DataFile::LoadData(const string &data)
 	size_t lineNumber = 0;
 	
 	size_t end = data.length();
-	for(size_t pos = 0; pos < end; pos = Utf8::NextCodePoint(data, pos))
+	for(size_t pos = 0; pos < end; )
 	{
 		++lineNumber;
+		size_t tokenPos = pos;
 		char32_t c = Utf8::DecodeCodePoint(data, pos);
 		
 		// Find the first non-white character in this line.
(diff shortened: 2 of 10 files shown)


tehhowch (commented):
Thanks for splitting this out!

Should we add a (temporary) warning when the game loads a character that cannot be displayed?

E.g. "Warning: unicode glyphs are not yet rendered in-game:"

If there would be many of these warnings, we could instead warn for the first instance per file and then maybe report the total count per file if multiple were observed.

copyright
```diff
@@ -814,6 +814,10 @@ Copyright: Tommaso Becca
 License: CC-BY-SA-4.0
 Comment: Derived from works by ESA/Hubble & NASA (under the same license)
 
+Files: source/Utf8.*
+Copyright: 2017, 2018 by Flávio J. Saraiva
+License: GPL-3+
+
```
Amazinite, line 820: Source files aren't shipped with the game, and therefore don't go in the copyright file. Source file copyright is handled by the block comment at the top.
```suggestion
```

source/DataFile.cpp
```diff
@@ -123,23 +127,28 @@ void DataFile::Load(const char *it, const char *end)
...
 			}
-			else if(fileIsSpaces && !warned && *it != ' ')
+			else if(fileIsSpaces && !warned && c != ' ')
 			{
 				warned = true;
 				stack.back()->PrintTrace("Mixed whitespace usage in file");
 			}
 			
 			++white;
+			pos = Utf8::NextCodePoint(data, pos);
```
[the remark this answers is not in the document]
oo13: I'll change 'pos' to the referenece to size_t.

source/Utf8.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+#include "Utf8.h"
+
+using namespace std;
+
+namespace Utf8 {
+	size_t NextCodePoint(const string &str, size_t pos)
+	{
+		if(pos >= str.length())
+			return str.length();
```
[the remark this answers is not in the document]
oo13: Yes. This case is a programming error so it should return stirng::n_pos as a error state.

copyright
oo13: Uhm, I have no idea how to cancel the thumb-down..., it's just a misclick.

I think this states the copyright for the git repository but it's not very important because it shows only a first author and we can confirm all copyright holders by the each files and the changelog.

oo13:
> Should we add a (temporary) warning when the game loads a character that cannot be displayed?

To be honest, I doubt that someone will use this function pragmatically. In fact, current version has accepted any character codes if it's quoted in a pair of quotations, almost no one confuses so far.

It's relatively easy to implement if we have no need to distinguish from a text that will be eventually displayed or not. What do you think?

copyright
MCOfficer: That would be conflicting information, since source files are in fact handled by the copyright file:

https://github.com/endless-sky/endless-sky/blob/7de6f95e0d55d6f712e5be744240f6f232dd8701/copyright#L6-L8

I see no reason not to include it.

oo13: Yes, but I had a idea and submitted #5346.

tehhowch:
> > Should we add a (temporary) warning when the game loads a character that cannot be displayed?
> 
> To be honest, I doubt that someone will use this function pragmatically. In fact, current version has accepted any character codes if it's quoted in a pair of quotations, almost no one confuses so far.
> 
> It's relatively easy to implement if we have no need to distinguish from a text that will be eventually displayed or not. What do you think?
> 

I think we can skip it, then.

tehhowch:
I don't understand what GitHub just did there, but that commit should have been your and @flaviojs 's authorship >_>

Amazinite:
Looks like it commit merged.

tehhowch:
Ah, yep. merge commit + the two commits from the branch. I feel better now, but had wanted to add a Co-Authored trailer to the squash commit for Flavio.

PR #5348, opened 2020-09-17
Phrasing fix for merchant hail

Terin:
image

The above hail uses the word "I don't hate them, but..." which is a slightly positive disclaimer which would normally be used to soften a negative follow-up. That does not match with the other words in that group, though, which are all slightly negative disclaimers followed by a positive statement. This PR adjusts it to the match the others.

### DIFF
--- a/data/human/hails.txt
+++ b/data/human/hails.txt
@@ -1477,7 +1477,7 @@ phrase "friendly civilian"
 		"I just don't get it"
 		"I'm not that into music"
 		"I prefer classical works"
-		"I don't hate them"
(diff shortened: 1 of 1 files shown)


Zitchas (approved):
This change makes sense as it improves the consistency of the message set and sounds a lot better.

McloughlinGuy:
I would suggest changing it to I don't like them, but there latest song wasn't too bad! Here is the file with the added correction: 
[hails.txt](https://github.com/endless-sky/endless-sky/files/5240526/hails.txt)

PR #5350, opened 2020-09-17
Remove unused member variable.

Rakete1111:
Nothing more to add. `ShipInfoPanel::selectedIndex` was unused and clang generated a warning which failed the build due to `-Werror`.


tehhowch:
I'm surprised this fails your build but not the repo's actions. What're the configuration differences between your XCode project and the one here?

Rakete1111:
I'm not using XCode :) Seems like [clang warns about this since clang 3.2](https://godbolt.org/z/sYcs5n). Don't know why the warning doesn't appear on XCode.

PR #5352, opened 2020-09-18
AI improvements for afterburners and reverse thrusters.

Rakete1111:
This PR improves the AI a bit when the ship has afterburners or reverse thrusters.

1. `CircleAround` now uses the afterburner for speed - basically PR #4655 which the author has abandoned with the review incorporated.

2. `MoveTo` now better uses the reverse thruster if possible. For example, when landing on a planet while flying straight to it will now use the reverse thruster to stop instead of turning around<sup>1</sup>.

3. `MoveToAttack` uses reverse thrusters to reach the enemy quicker.

More details:

`CircleAround` will, if the ship is 750 away from the target fire its afterburners. This gives it enough time to gather without shooting past the target.

`MoveToAttack` circles the enemy, but if the enemy flies through the ship then currently the ship turns around. But if it has reverse thrusters it can reverse to reach the enemy, all while still turning to face it. The enemy has a tiny bit of a harder time evading.

<sup>1</sup> (`MoveTo`) The ship will only use the reverse thrusters if doing Shift+Down in that frame results in the ship being on the target. But keep in mind that if turning + stopping is faster than the reverse thruster then the reverse thruster won't be used.

## Testing Done

I tested each feature; here are the save files. [1](https://gist.github.com/Rakete1111/5cb0a9627dff7c9295cc31ffcdf72843) [2](https://gist.github.com/Rakete1111/1d86e4a82b92ea0be30a53ed25e60de5) [3](https://gist.github.com/Rakete1111/2fd6dfd93db408b90e4f5a1264d209e4)

## Performance Impact
N/A

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1755,8 +1755,15 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	
 	bool shouldReverse = false;
 	dp = targetPosition - StoppingPoint(ship, targetVelocity, shouldReverse);
+	if(dp.Length() < radius && shouldReverse)
+	{
+		// We can directly use the reverse thrusters to stop at the target.
+		command |= Command::BACK;
+		return false;
+	}
+
 	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .8);
-	if(!isClose || (!isFacing && !shouldReverse))
+	if(!isClose || !isFacing)
 		command.SetTurn(TurnToward(ship, dp));
 	if(isFacing)
 		command |= Command::FORWARD;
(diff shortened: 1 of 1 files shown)


Zitchas:
This sounds interesting. I don't have the code knowledge to properly review it, but I do like the sounds of making changes to the AI to improve its use of afterburners and reverse thrust.

MasterOfGrey:
If it works as it says on the tin I am very excited about this.

Sandwichs-del:
Amazing timing. I just got a ship with weak steering + reverse thrusters.
Landing manually using the reverse thrusters is faster than letting the current AI circle the planet a few times, but it is SO annoying.

source/AI.cpp
```diff
@@ -1755,8 +1755,15 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	
 	bool shouldReverse = false;
 	dp = targetPosition - StoppingPoint(ship, targetVelocity, shouldReverse);
+	if(dp.Length() < radius && shouldReverse)
```
tehhowch, line 1758: Flip the check order to avoid the length calculation when not necessary.

PR #5356, opened 2020-09-20
Allow use of multiple ammunition types per weapon

vitalchip:
Currently an ammunition based weapon can only use one type of ammunition, this PR allows use or more than one type and also allows negative ammunition use to 'create' new ammunition.

Useage is such that if a number of ammo types are specified, all are required to fire the weapon.

If a negative value is specified, firing the weapon will add the ammunition outfit to your ship, providing there is enough ammo storage space to accomodate.

If the 'ammunition' is a specified as different type of outfit, it will be added to cargo, again provided there is space.

If no storage/cargo space is available the weapon will not fire.

Screenshots attached to show how multiple ammunition is specified in outfit files and displayed in Outfitter.

Screenshot 2020-09-20 at 09 38 10
Screenshot 2020-09-20 at 09 39 10

Ammunition counters on screen now show number of shots remaining for each weapon, to account for multiple ammunition types and quantities.

Performance impact is negligible, ammunition checks will naturally take longer if a large number of ammunition types are specified.


tehhowch:
- Restore the the PR template headers Particularly, you should include related issues and how this PR addresses / solves their requirements.

 - Reusing the same attribute label "ammo" but having an entirely different meaning is a no-go, especially when both meanings are intended to appear at the same time.

 - You mention this "consume-all" behavior can be modified in the future to support "consume-one-of" -- do you have an example proposed syntax for this?

 - Please ensure your PR follows the style guidelines, including whitespace levels and the role of code comments.

vitalchip:
Can you expand on this one?  

Reusing the same attribute label "ammo" but having an entirely different meaning is a no-go, especially when both meanings are intended to appear at the same time.

My original idea was to have a second ammo attribute "subammo" was my working name but could have been anything, is this what you're meaning?

tehhowch:
```
ammo A
ammo B
ammo usage -1
ammo C
```

No player, without coming here to read the text of this PR, is going to understand what is happening when they see the above bits.

vitalchip:
I understand, I've very little experience coding and was simply following the suggestion given. Seems like your request would be for a number of ammo pairs with different names, rather than one ammo map?   Can I refer you to Derpy at this point for discussion on the merits?

If it's decided that this isn't the way forward, go ahead and close the PR and it can get looked at another time. Thanks.

Screenshot 2020-09-21 at 13 04 44

Amazinite:
This is a display issue in the outfitter, not an issue of how the information is stored. The map doesn’t need to change.

vitalchip:
I don't know how to do the formatting but the list idea seems best.

Something like "Ammo Required" <list>. and then later possibly "Ammo Options:" <list>

Longer term I'm thinking something like three <for space> small icons in a row under the main icon/counter which can be clicked to select optional ammo types, so there would always be at least one required ammo for every shot and then a selection of options to change some weapon properties.

Everything "Ammo Options" related is for another time, just putting the idea out.

Right now, I'll need to have a look at the Outfitter Display and play around a bit, I've a lot on at the moment, would it be preferable to close this PR for now and I'll open a new one when I've addressed the points mentioned, rather than this one laying around for a while?

vitalchip:
How does this sit with you guys?  I think it's best to keep the header and list format even if it's just 1 ammo, keeps things consistent.  I've also attached a second screenshot of a possible future, just to demonstrate how it might be used.

Screenshot 2020-09-22 at 17 48 01
Screenshot 2020-09-22 at 20 00 36

vitalchip:
I've found two issues.

1. Ammo is not properly used in the case of Nuclear Missiles.
2. Required ammo for a given outfit doubles if the outfit is defined twice ( base game and plugin )

These will need to be worked on and resolved. I will resubmit PR when this has been completed.

tehhowch:
No need to close the PR and/or delete the branch -- just convert it to draft and work on it as you are able.

PR #5359, opened 2020-09-22
feat: Text truncating more widely

oo13:
**Feature:** This PR implements the feature request discussed in the comment. (https://github.com/endless-sky/endless-sky/pull/4123#issuecomment-683469962)

## Feature Details
This PR introduces ES to text truncating more widely and the class Font has an interface that is similar to #4123.

The class Font is responsible for text truncating now.

## UI Screenshots
<details><summary>The differences between current master and this PR (and #4123)</summary><div>

|master|this PR|Pango|
|---|---|---|
|master_load|pr_load|pango_load|
|master_land|pr_land|pango_land|
|master_map|pr_map|pango_map|
|master_job|pr_job|pango_job|
|master_playerinfo|pr_playerinfo|pango_playerinfo|
|master_shipinfo|pr_shipinfo|pango_shipinfo|
|master_outfitter|pr_outfitter|pango_outfitter|
|master_shipyard|pr_shipyard|pango_shipyard|

Combination with truncating and wrapping is complicated, and each version has each way.
- The master truncate the ship names other than first and last. I guess the creator thought it's hard to truncate them flawlessly.
- This PR version truncates a word in a line which have a single word.
- The Pango version truncate the text only if it has more than two ship names because Pango doesn't wrap automatically a text if it is specified a truncation and may truncate the number of credit.

|number of ships to sell|master|this PR|Pango|
|---|---|---|---|
|1|master_sell1|pr_sell1|pango_sell1|
|2|master_sell2|pr_sell2|pango_sell2|
|3|master_sell3|pr_sell3|pango_sell3|

</div></details>

## Usage Examples
[text-truncating-data.zip](https://github.com/endless-sky/endless-sky/files/5261390/text-truncating-data.zip)
This plugin and save file introduce very long names.

## Testing Done
I have took some screen shots.

## Performance Impact
This makes ES slightly slow down due to additional parameters and calculation for widths of texts.


source/Font.h
```diff
@@ -36,14 +36,29 @@ class Font {
...
+			int width = -1;
+			// Set the alignment mode.
+			Align align = LEFT;
+			// Set the truncate mode.
+			Truncate truncate = TRUNC_NONE;
+			
+			Layout() noexcept;
+			Layout(int w, Align a) noexcept;
+			Layout(int w, Truncate t) noexcept;
+			Layout(int w, Truncate t, Align a) noexcept;
```
tehhowch, line 53: question: Should this be width - Align - Truncate? Or is width - Truncate - Align more "natural," in call-site usage?

source/Information.h
```diff
@@ -14,11 +14,13 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #define INFORMATION_H_
 
 #include "Color.h"
+#include "Font.h"
```
tehhowch, line 17: Would it make sense to break out the `Layout` class from `Font`? Doing so would mean we have a "lightweight" include for when all a class's interface needs to know is how we represent a layout, rather than how we draw fonts. The penalty is then that we need to have more explicit includes in more files, but at the same time, we only need to include the bits that are used.

If you choose to do that, I'm fine with grouping the font-related helper classes in a `fonts` subfolder. (If you also move the Font.cpp & Font.h files in this PR and they are shown as "deleted & added" instead of renamed, I would  recommend waiting for a follow-up PR to move those 2 files.)

source/Table.cpp
```diff
@@ -253,13 +281,13 @@ Rectangle Table::GetRowBounds() const
 
 
 Table::Column::Column()
-	: offset(0.), align(0.)
+	: offset(0.), layout()
 {
 }
```
tehhowch, line 286: Remove & ` = default`

source/WrappedText.cpp
```diff
@@ -148,8 +163,26 @@ int WrappedText::Height() const
...
-		font->Draw(text.c_str() + w.Index(), w.Pos() + topLeft, color);
+	if(words.empty())
+		return;
+	
+	if(truncate == Font::TRUNC_NONE)
+		for(const Word &w : words)
+			font->Draw(text.c_str() + w.Index(), w.Pos() + topLeft, color);
+	else
+	{
+		// Apply the truncation to a word only if a line has a single word.
```
tehhowch, line 174: A single word? or more than one?

source/WrappedText.h
```diff
@@ -69,7 +76,6 @@ class WrappedText {
 	
 private:
 	void SetText(const char *it, size_t length);
-	void Wrap();
```
tehhowch, line 72: Why do we need to change the visibility of the method?

source/WrappedText.cpp
```diff
@@ -170,10 +203,6 @@ Point WrappedText::Word::Pos() const
 
 void WrappedText::SetText(const char *it, size_t length)
 {
-	// Clear any previous word-wrapping data. It becomes invalid as soon as the
-	// underlying text buffer changes.
-	words.clear();
```
tehhowch, line 175: We should probably still clear the words here.

tehhowch (commented):
Thanks for the PR, it will definitely improve the state of displayed long text :+1: 

Most of the changes make sense based on the changes to the Font class, but I have some comments & questions that may inspire alternatives / refactoring

source/Font.h
oo13: I feel Align is more important, so width - Align - Truncate.

source/WrappedText.cpp
oo13: A single word.
If there are multiple word in a line, it does not try to truncate the line.
Too complicated.

source/WrappedText.h
oo13: Probably it was needed at some time and I forgot to restore it.

source/WrappedText.cpp
oo13: Same as the previous comment. I'll revert it.

source/Dialog.h
```diff
@@ -36,24 +37,28 @@ class Dialog : public Panel {
 public:
 	// Dialog that has no callback (information only). In this form, there is
 	// only an "ok" button, not a "cancel" button.
-	explicit Dialog(const std::string &text);
+	// width and align in layout will be ignored.
+	explicit Dialog(const std::string &text, const Font::Layout &layout = Font::Layout{});
```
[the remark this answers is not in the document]
oo13: I plan to create the class ```DisplayText``` and move the class ```Layout``` and some enums into ```DisplayText```.

source/BankPanel.cpp
```diff
@@ -35,7 +35,20 @@ namespace {
 	// Column headings.
 	const string HEADING[6] = {"Type", "Principal", "Interest", "Term", "Payment", ""};
 	// X coordinates of the columns of the table.
-	const int COLUMN[5] = {-290, -180, -100, -30, 20};
+	struct Columns {
```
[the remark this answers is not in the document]
oo13: I would like to add a method AddColumn(int x, DisplayText::Align align) as a simple version. This will be equivalent to AddColumn(x, {-1, align, Truncate::NONE}).

source/Dialog.cpp
```diff
@@ -285,6 +285,7 @@ void Dialog::Init(const string &message, bool canCancel, bool isMission)
 	text.SetAlignment(WrappedText::JUSTIFIED);
 	text.SetWrapWidth(WIDTH - 20);
 	text.SetFont(FontSet::Get(14));
+	text.SetTruncate(layout.truncate);
```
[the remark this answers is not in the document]
oo13: I'll change ```Layout``` to ```Truncate``` of the parameter of the constructors ```Dialog```.

source/Information.h
oo13: I plan to create ```DisplayText.h``` as a lightweight include.

oo13:
Thank you for the suggestions. I updated it.

tehhowch (approved):
@oo13 please review the changes I've made, especially to the copyright headers. Make any changes that make sense, & we should be able to wrap this up shortly.

As far as extending `Interface::Element::TextElement` to also support setting the text alignment strategy (via a keyword such as `"text-align"`), I think we should defer that to a follow-up PR, given the size of this one.

oo13:
LGTM. Thank you for refactoring.
I posted a commit to change the expression of my name because I prefer another expression of my name (LAST, First). I have not found other problems.

PR #5361, opened 2020-09-23
feat(interface): add hire-max and fire-max hotkeys

kborer:

-----------------------
**Feature:** This PR implements the feature with no related issue.

## Feature Details
Add two hotkeys to the hiring panel. 'H' for hire max and 'F' for fire max.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
N/A

## Performance Impact
N/A

### DIFF
--- a/source/HiringPanel.cpp
+++ b/source/HiringPanel.cpp
@@ -117,6 +117,16 @@ bool HiringPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, b
 		player.Flagship()->AddCrew(-min(maxFire, Modifier()));
 		player.UpdateCargoCapacities();
 	}
+	else if(key == 'H')
+	{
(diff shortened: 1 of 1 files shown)


Amazinite:
Ctrl, shift, and alt already act as modifiers on the number of crew hired and fired. Shift = 5x, ctrl = 25x, and alt = 500x, and holding any combination of these keys multiplies those modifiers together (e.g. shift + ctrl = 100x). (Holding alt alone is enough for any ship in the game outside of a Worldship in terms of hiring or firing max possible crew at once.) Given then that shift is already used as a modifier, but would need to be used to get from "h" to "H", you're changing the existing behavior of the shift key. 

(This same modifier behavior exists for buying/selling commodities and outfits and buying ships as well.)

kborer:
my bad

PR #5363, opened 2020-09-23
Skip capacity check on Uninstall

Sandwichs-del:
**Feature:** This PR implements a trivial quality-of-life feature

## Feature Details
Having 'u' sometimes uninstall the outfits and sometimes sell them is very confusing.
Always place uninstalled outfits into the cargo hold, even if it exceeds its capacity.
This change will make swapping active and inactive outfits much more comfortable.

## UI Screenshots
N/A

## Testing Done
None, sorry.
If it passes CI, it should be good to go.

## Performance Impact
N/A

Thank you!

### DIFF
--- a/source/OutfitterPanel.cpp
+++ b/source/OutfitterPanel.cpp
@@ -582,9 +582,10 @@ void OutfitterPanel::Sell(bool toCargo)
 			if(selectedOutfit->Get("required crew"))
 				ship->AddCrew(-selectedOutfit->Get("required crew"));
 			ship->Recharge();
-			if(toCargo && player.Cargo().Add(selectedOutfit))
+			if(toCargo)
 			{
-				// Transfer to cargo completed.
+				// Transfer to cargo even if it would exceed the capacity
+				player.Cargo().Add(selectedOutfit, 1, true)
 			}
 			else
 			{
--- a/source/OutfitterPanel.cpp
+++ b/source/OutfitterPanel.cpp
@@ -585,7 +585,7 @@ void OutfitterPanel::Sell(bool toCargo)
 			if(toCargo)
 			{
 				// Transfer to cargo even if it would exceed the capacity
-				player.Cargo().Add(selectedOutfit, 1, true)
+				player.Cargo().Add(selectedOutfit, 1, true);
 			}
 			else
 			{
--- a/source/CargoHold.cpp
+++ b/source/CargoHold.cpp
@@ -458,13 +458,13 @@ void CargoHold::TransferAll(CargoHold &to, bool transferPassengers)
 
 
 // Add the given amount of the given commodity.
-int CargoHold::Add(const string &commodity, int amount)
+int CargoHold::Add(const string &commodity, int amount, bool force)
 {
 	if(amount < 0)
 		return -Remove(commodity, -amount);
 	
-	// If this cargo hold has a size limit, apply it.
-	if(size >= 0)
+	// If this cargo hold has a size limit, apply it (unless forced).
+	if(!force && size >= 0)
 		amount = max(0, min(amount, Free()));
 	commodities[commodity] += amount;
 	return amount;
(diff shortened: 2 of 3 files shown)


Zitchas:
This makes sense - While we are on-planet, it is reasonable to assume that we can uninstall tings and have them sitting on the tarmac next to our ship. Having a warning when we depart that we have stuff left on the tarmac that will be auto-sold would be good, but isn't absolutely necessary.

Sandwichs-del:
Actually, we'll get negative "cargo space free" and trigger the existing auto-sell warning.
You can trigger this situation already - fill up the cargo space and then sell CargoExpansion outfit.

source/OutfitterPanel.cpp
```diff
@@ -582,9 +582,10 @@ void OutfitterPanel::Sell(bool toCargo)
 			if(selectedOutfit->Get("required crew"))
 				ship->AddCrew(-selectedOutfit->Get("required crew"));
 			ship->Recharge();
-			if(toCargo && player.Cargo().Add(selectedOutfit))
+			if(toCargo)
 			{
-				// Transfer to cargo completed.
+				// Transfer to cargo even if it would exceed the capacity
```
petervdmeer, line 587: Minor remark: missing dot at the end of the sentence.

petervdmeer (approved):
Change looks good to me. (One really minor remark.)

tehhowch (changes requested):
Rather than modify the function behavior to have two ways to ignore the size, use the established size manipulation method: modify, transfer, then restore the size.

(This alteration to the mechanic was suggested in https://github.com/endless-sky/endless-sky/pull/2806#issuecomment-334766770 but not implemented)

Sandwichs-del:
> Rather than modify the function behavior to have two ways to ignore the size, use the established size manipulation method: modify, transfer, then restore the size.
@tehhowch 
Sounds weird and complicated, but OK.
I find it hard to grep for "size manipulation". Could you point me to an example?

I like install/uninstall, because it makes "check the tarmac for leftover outfits" very easy - just switch to the cargo hold view. With buy/sell I have to go through the shop checking for "in stock: 1" labels and even then risk "losing" outfits that have unlimited stock.

petervdmeer:
> > Rather than modify the function behavior to have two ways to ignore the size, use the established size manipulation method: modify, transfer, then restore the size.
> 
> @tehhowch
> Sounds weird and complicated, but OK.
> I find it hard to grep for "size manipulation". Could you point me to an example?

I believe this is an example of the method tehhowch was talking about:
https://github.com/endless-sky/endless-sky/blob/ec7e642f78d5a77bf9fe4239500e85f8a9bcff3a/source/MissionAction.cpp#L96

source/OutfitterPanel.cpp
Sandwichs-del: Thanks. Will fix in v2

Sandwichs-del:
I found no good way to edit this PR, so I opened a v2 here: https://github.com/endless-sky/endless-sky/pull/5507

Sandwichs-del:
@tehhowch Yes, but implementing the requested change effectively reverts all commits. I found it easier to start from scratch. If you prefer I revert all commits and push the v2 one, say so :)

Sandwichs-del:
@tehhowch Yes, but implementing the requested change effectively reverts all commits. I found it easier to start from scratch. If you prefer I revert all commits and push the v2 one, say so :)

MCOfficer:
If need be, you can still push all changes to your feature branch

Sandwichs-del:
OK. I tried updating https://github.com/Sandwichs-del/endless-sky/tree/ignore_capacity_on_uninstall without a force push, but it looks like a mess, right?

MCOfficer:
Uhhh... it should've updated this PR as well :raised_eyebrow:

petervdmeer:
> Uhhh... it should've updated this PR as well raised_eyebrow

Yes, but I have the impression that the update only happens when the PR is not closed. (So if you re-open this PR, then I expect that you would see the new commits in this PR.)

PR #5365, opened 2020-09-24
Fixed Mission.cpp default init for EMPTY

thebigh2014:
**Build fix:** This PR addresses inability to successfully build master/continuous for macOS (10.11 using Xcode 8.2.1)

## File Name(s) / Line Number(s)
Mission.cpp - Line 961

## Error Encountered:
Default initialization of an object of const type "const MissionAction" without a user-provided default constructor.

## Fix Details
Add empty object initializer {} - this is the recommended fix by Xcode.

## Testing Done
Builds successfully and game operates normally.

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -958,7 +958,7 @@ const string &Mission::Identifier() const
 const MissionAction &Mission::GetAction(Trigger trigger) const
 {
 	auto ait = actions.find(trigger);
-	static const MissionAction EMPTY;
+	static const MissionAction EMPTY{};
 	if(ait != actions.end())
 		return ait->second;
 	else


tehhowch:
Seems like this may be an issue with your compiler that was addressed in the standard and corrected in a later version of clang. Otherwise I would expect the Linux and windows builds, and the official macOS build, to fail.

What version is reported on your system? Are you able to upgrade it?

thebigh2014:
My system reports LLVM 8.0.0, clang-800.0.42.1, which is based on 3.9.0svn.  Sadly I cannot upgrade without upgrading my OS as this is the latest my current OS (10.11) supports.  Guess I'll just have to keep making the change manually for my builds until I'm able to upgrade my OS.

tehhowch:
The proposed change here is not incompatible, but shouldn't be necessary. I'm surprised clang cannot be installed independently of XCode.

thebigh2014:
I think it can be but getting it to work with Xcode isn't the easiest.  It's because Apple's version in Xcode is not straight vanilla clang, hence the "based on 3.9.0svn".  Apparently they have a need to backport some things and make other adjustments which is why the version doesn't line up with mainline clang.

It did strike me as odd that not having this change resulted in an outright error instead of a warning.  At least it was the only thing in the code base that prompted manual editing on my part so it still has compatibility with 10.11 otherwise.

PR #5367, opened 2020-09-25
Content: Remnant Expanded Horizons Astral

Zitchas:
**Content (Artwork & Missions & Jobs)**

## Summary
The Remnant are a faction of explorers, driven to explore the various corners of the galaxy, document what they find, and add it to the collected mass of data that they use to help understand their place in the universe.

This PR adds:
- 2 missions wherein Dawn embarks on her first major data gathering expedition
- 1 repeating job where the player can go collect updated data for various systems around the galaxy.
- 1 outfit: The Research Laboratory (Acts kind of like luxury quarters in serving as the prerequisite for new job types, one of which is included in this PR.

Lore implications:
- This PR pretty much nails down that the player isn't special for having explored. Long story short: If the player can find it within a year or two of flying around with a JD, then the Remnant can (and have) probably found it too. 
- Remnant Rites: In one mission we see a small departure ritual that the Remnant have for those who take part in long range exploration before their first trip.

Research Laboratory outfit:
- This is basically the companion outfit to the Salvage Scanner, and provides the Remnant with Asteroid and Cargo scanning capabilities.
- Requires 1 crew to operate, thus establishes that the player now has a scientist, research tech, science officer, or similar sort of crew on their ship. (the remnant refer to this person as either a scientist or researcher, but the player might be more familiar with "science officer" from Star Trek)

Note: The second half of the Cognizance mission set requires the Research Laboratory, so depends on this PR.

## Save File
This save file can be used to play through the new mission content:
{{attach a save file that allows people to easily test your added mission content or see your new in-game art}}

## PR Checklist
 - [x] I have tested my content and know that it works as intended
 - [x] I updated the copyright attributions, or decline to claim copyright of any assets produced or modified
 - [X] I uploaded the necessary image, blend, and texture assets here: [zip file on google drive](https://drive.google.com/file/d/1kjnQ2N84-WYtrMWLKYt66-eQFlU2yMgl/view?usp=sharing)(Note: Assets updated on 2020-Oct-17)
 - [X] The files for Sagittarius A* can be found in [this zip file on a google drive](https://drive.google.com/file/d/1sCZTkf99VPZC-Km1scn4oCOs6UST5_xx/view?usp=sharing)(uploaded 2020-Nov-25)
[...]

PR #5374, opened 2020-09-26
Add regional/noteworthy features attributes to systems

Galaucus:
## Feature Details
This adds attribute tags to every system in the map file. Tags are based on the dominant species/government, with additional tags for sub-regions (such as the Deep), noteworthy points in galactic topography - such as dead-ends, hyperspace "islands", and system-defining phenomena such as the presence of supernova remnants, ringworlds, or wormholes.

New tags are prefixed with a ! to make find-and-replace easier. This isn't necessarily intended to be their final name; just something to facilitate replacement when that's decided on.

Topographical tags:
**!fringe** - applied to uninhabited systems which don't lead to any other inhabited systems. Only explorers, miners, or those with something to hide would have a reason to visit these systems.
**!isolated** - applied to systems which have no hyperspace links. 

System phenomena:
**!ringworld** - This system hosts a ringworld, intact or otherwise.
**!nova** - This system's star has gone nova.
**!wormhole** - This system hosts a wormhole
**!black hole** - This system has a black hole. At the moment this is only applied to Sag. A*.
**!hazard** - This system can destroy your ship if you linger. At the moment this is only applied to Sayaiban (nanobots).
**!warzone** - This system is the front-line of a conflict. At the moment this is applied to Mesuket and Wah Ki. It could, however, find use in the campaigns to generate job board missions based on your faction - relief shipments, rescue/exfiltration, reinforcements, EOD, etc.
**!war-torn** - Not currently applied to any systems. I recommend applying it to ex-warzone systems. Could be used to generate missions regarding reconstruction, EOD, refugee aid/transport, salvage, etc.

**!human** - applied to all systems in greater human space. Includes uninhabited systems, Deneb, Quarg near Tarazed, does not include Remnant or human-inhabited Hai space.

Tags for sub-regions of human space:
**!north
!deep
!paradise
!core
!near earth 
!rim
!dirt belt
!south**

**!coalition** - Greater coalition space, including Heliarch-governed systems.

Tags for sub-regions of Coalition space:
**!arach**
**!saryd**
**!kimek**

**!korath** - This is applied to the full historical extent of Korath civilization, including now-abandoned systems, supernova'd stars, and the current extent of Korath spread.

Tags for sub-regions of Korath space:
**!automata** - Includes uninhabited systems where automata spawn, or have left traces.
**!efreti** - Includes the Quarg ringworld in Korath space.
**!exiles** - Includes the uninhabited star between the two main exile "islands".

**!wanderer** - Describes historical Wanderer space, including territory ceded to the Unfettered. Does not include Wanderer colonies/bases in Korath space.

**!hai** - Applied to all of Hai space, includes the Unfettered

**!ember waste** - Most systems in the ember waste already have the "ember waste" attribute added by Zitchas. My additions are prefixed with the ! to disambiguate. If the change is acceptable these should have the ! taken out.

**!pleiades** - Systems in the galaxy found through the Deneb wormhole.

## UI Screenshots
N/A

## Usage Examples
```
system Peresedersi
	pos 81.4305 -154.812
	government Uninhabited
	attributes "archon" "!korath" "!isolated" "!nova"
```
The system Peresedersi has had three new tags appended to it: "Korath" shows that it's within the historical extent of Korath space (including the current diaspora and their former stomping grounds), "isolated" is appended due to there being no hyperlanes linking to this system, and "nova" on account of the system hosting supernova remnants. Archon was already there.
## Testing Done
None so far (9/26/20)

## Performance Impact
N/A

## Lake's notes and concerns
The main intention behind the addition of these tags is to make it easier to generate missions that target - or exclude - specific regions of space. Along the way I decided adding in noteworthy system-wide features could be conducive to that aim. 

Mixed feelings about tagging the Pleiades system "World's End" with **!ringworld**. There's one there, but generating missions that target it seem like an unintended result. As such, I'm leaving it out unless y'all think otherwise.

### DIFF
--- a/data/human/news.txt
+++ b/data/human/news.txt
@@ -1192,5 +1192,7 @@ news "graffiti on spaceport wall"
 			"Haikus are easy, / but sometimes they don't make sense. / Refrigerator."
 			"Haikus are easy. / I can't think of a good line. / It's snowing on Mount Fuji."
 			"A long time ago / I was respected by all, / but now I sit here."
+			"A handsome young cyborg named Ace / Wooed women at every base / But once ladies glanced at / His special enhancement / They vanished with nary a trace"
+			"Mary had a little lamb / Little lamb little lamb / Mary had a little lamb / Whose fleece was white as snow"
 		word
 			`"`
--- a/data/human/news.txt
+++ b/data/human/news.txt
@@ -1174,8 +1174,6 @@ news "graffiti on spaceport wall"
 			"Graffiti painted on a wall"
 	message
 		word
-			`"`
-		word
 			"A lone, soulless husk. / Appointed immortal king. / To sulk forever. - fgnt"
 			"The galaxy's child / Cursed with madness of time / Behind bloody chains - fgnt"
 			"Too proud for farming / Too poor to buy Paradise / I'll hoist the black flag - Lake"
@@ -1194,5 +1192,3 @@ news "graffiti on spaceport wall"
 			"A long time ago / I was respected by all, / but now I sit here."
 			"A handsome young cyborg named Ace / Wooed women at every base / But once ladies glanced at / His special enhancement / They vanished with nary a trace"
 			"Mary had a little lamb / Little lamb little lamb / Mary had a little lamb / Whose fleece was white as snow"
-		word
-			`"`
--- a/data/map.txt
+++ b/data/map.txt
@@ -1829,7 +1829,7 @@ system Ablub
 system Acamar
 	pos -284 -3
 	government Syndicate
-	attributes "!human" "!syndicate" "!core"
+	attributes "!human" "!core"
 	habitable 1080
 	belt 1013
 	link Diphda
@@ -1888,7 +1888,7 @@ system Acamar
 system Achernar
 	pos -93 154
 	government Syndicate
-	attributes "!human" "!syndicate" "!core"
+	attributes "!human" "!core"
 	habitable 625
 	belt 1963
 	haze _menu/haze-133
@@ -1960,7 +1960,7 @@ system Achernar
 system Acrux
 	pos -808 192
 	government Republic
-	attributes "!human" "!free worlds" "!rim"
+	attributes "!human" "!rim"
 	habitable 590.625
 	belt 1407
 	haze _menu/haze-67
@@ -2040,7 +2040,7 @@ system Acrux
 system Adhara
 	pos -669 -200
 	government Republic
-	attributes "!human" "!republic" "!deep"
+	attributes "!human" "!deep"
 	habitable 670
 	belt 1014
 	link "Epsilon Leonis"
@@ -2202,7 +2202,7 @@ system "Aki'il"
 system "Al Dhanab"
 	pos -177 207
 	government Syndicate
-	attributes "!human" "!syndicate" "!core"
+	attributes "!human" "!core"
 	habitable 1080
 	belt 1432
 	haze _menu/haze-133
@@ -2272,7 +2272,7 @@ system "Al Dhanab"
 system Albaldah
 	pos -350 484
 	government Republic
-	attributes "!human" "!free worlds" "!south"
+	attributes "!human" "!south"
 	habitable 534.375
 	belt 1622
 	haze _menu/haze-133
@@ -2347,7 +2347,7 @@ system Albaldah
 system Albireo
 	pos -270 503
 	government Republic
-	attributes "!human" "!free worlds" "!south"
+	attributes "!human" "!south"
 	habitable 760
 	belt 1030
 	link Tarazed
@@ -2472,7 +2472,7 @@ system Alcyone
 system Aldebaran
 	pos -356 -26
 	government Republic
-	attributes "!human" "!republic" "!paradise"
+	attributes "!human" "!paradise"
 	habitable 625
 	belt 1296
 	link Capella
@@ -2518,7 +2518,7 @@ system Aldebaran
 system Alderamin
 	pos -272 258
 	government Syndicate
-	attributes "!human" "!syndicate" "!core"
+	attributes "!human" "!core"
 	habitable 1705
 	belt 1908
 	haze _menu/haze-133
@@ -2583,7 +2583,7 @@ system Alderamin
 system Aldhibain
 	pos -659 451
 	government Republic
-	attributes "!human" "!free worlds" "!south"
+	attributes "!human" "!south"
 	habitable 455.625
 	belt 1840
 	link Kornephoros
@@ -2698,7 +2698,7 @@ system Algenib
 			period 42.6838
 
 system Algieba
-	attributes "!human" "!republic" "!deep"
+	attributes "!human" "!deep"
 	pos -691 -52
 	government Republic
 	habitable 2695
@@ -2770,7 +2770,7 @@ system Algieba
 system Algol
 	pos -210 17
 	government Syndicate
-	attributes "!human" "!syndicate" "!core"
+	attributes "!human" "!core"
 	habitable 534.375
 	belt 1839
 	haze _menu/haze-133
@@ -2832,7 +2832,7 @@ system Algol
 system Algorel
 	pos -631 145
 	government Republic
-	attributes "!human" "!republic" "!dirt belt"
+	attributes "!human" "!dirt belt"
 	habitable 320
 	belt 1993
 	haze _menu/haze-67
@@ -2897,7 +2897,7 @@ system Algorel
 system Alheka
 	pos -350 -271
 	government Republic
-	attributes "!human" "!republic" "!north"
+	attributes "!human" "!north"
 	habitable 1080
 	belt 1380
 	haze _menu/haze-67
@@ -2956,7 +2956,7 @@ system Alheka
 system Alhena
 	pos -430 -80
 	government Republic
-	attributes "!human" "!republic" "!paradise"
+	attributes "!human" "!paradise"
 	habitable 945
 	belt 1078
 	haze _menu/haze-67
@@ -3026,7 +3026,7 @@ system Alhena
 system Alioth
 	pos -620 315
 	government Republic
-	attributes "!human" "!republic" "!dirt belt"
+	attributes "!human" "!dirt belt"
 	habitable 455.625
 	belt 1889
 	link Wei
@@ -3079,7 +3079,7 @@ system Alioth
 system Alkaid
 	pos -825 318
 	government Republic
-	attributes "!human" "!republic" "!rim"
+	attributes "!human" "!rim"
 	habitable 1080
 	belt 1142
 	haze _menu/haze-67
@@ -3257,7 +3257,7 @@ system Almach
 system Alnair
 	pos -272 314
 	government Syndicate
-	attributes "!human" "!syndicate" "!core"
+	attributes "!human" "!core"
 	habitable 135
 	belt 1969
 	haze _menu/haze-133
@@ -3326,7 +3326,7 @@ system Alnair
 system Alnasl
 	pos -553 380
 	government Republic
-	attributes "!human" "!republic" "!dirt belt"
+	attributes "!human" "!dirt belt"
 	habitable 640
 	belt 1412
 	haze _menu/haze-133
@@ -3467,6 +3467,7 @@ system Alnilam
 system Alnitak
 	pos -331 -412
 	government Republic
+	attributes "!human" "!north"
 	habitable 349.375
 	belt 1907
 	haze _menu/haze-67
@@ -3523,6 +3524,7 @@ system Alnitak
 system Alniyat
 	pos -691 501
 	government Republic
+	attributes "!human" "!south"
 	habitable 534.375
 	belt 1760
 	link Pherkad
@@ -3595,6 +3597,7 @@ system Alniyat
 system "Alpha Arae"
 	pos -481 394
 	government Republic
+	attributes "!human" "!dirt belt"
 	habitable 945
 	belt 1405
 	link "Delta Sagittarii"
@@ -3665,6 +3668,7 @@ system "Alpha Arae"
 system "Alpha Centauri"
 	pos -429 125
 	government Republic
+	attributes "!human" "!near earth"
 	habitable 320
 	belt 1116
 	haze _menu/haze-67
@@ -3727,6 +3731,7 @@ system "Alpha Centauri"
 system "Alpha Hydri"
 	pos -97 98
 	government Syndicate
+	attributes "!human" "!core"
 	habitable 1400
 	belt 1839
 	haze _menu/haze-133
@@ -3780,6 +3785,7 @@ system "Alpha Hydri"
 system Alphard
 	pos -588 -47
 	government Republic
+	attributes "!human" "!paradise"
 	habitable 1080.62
 	belt 1101
 	link Algieba
@@ -3846,6 +3852,7 @@ system Alphard
 system Alphecca
 	pos -546 308
 	government Republic
+	attributes "!human" "!dirt belt"
 	habitable 1400
 	belt 1101
 	link Alioth
@@ -3895,6 +3902,7 @@ system Alphecca
 system Alpheratz
 	pos -176 125
 	government Syndicate
+	attributes "!human" "!core"
 	habitable 1080.62
 	belt 1645
 	haze _menu/haze-133
@@ -3962,6 +3970,7 @@ system Alpheratz
 system Altair
 	pos -357 161
 	government Republic
+	attributes "!human" "!near earth"
 	habitable 945
 	belt 1830
 	link Vega
@@ -4024,6 +4033,7 @@ system Altair
 system Aludra
 	pos -606 -374
 	government Republic
+	attributes "!human" "!deep"
 	habitable 455.625
 	belt 1511
 	haze _menu/haze-33
@@ -4087,6 +4097,7 @@ system Aludra
 system "Ancient Hope"
 	pos -1159.59 702.051
 	government Coalition
+	attributes "!coalition" "!saryd"
 	habitable 455
 	belt 1439
 	haze _menu/haze-67
@@ -4156,6 +4167,7 @@ system "Ancient Hope"
 system Ankaa
 	pos -230 100
 	government Syndicate
+	attributes "!human" "!core"
 	habitable 625
 	belt 1003
 	haze _menu/haze-133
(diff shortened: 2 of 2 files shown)


Zitchas:
Thanks for doing all this! This will make a variety of Remnant jobs much more customizable.

General note: Once this is merged, it will be used for scientific research jobs by the Remnant. See #5367 's job as an example. That job currently picks a random system in the galaxy (it only omits systems with the Pleiades attribute), but ideally that job would be split into one lower-probability job for each "region" of the galaxy so that the pay can be varied depending on how far away it is and how risky the Remnant consider it.

petervdmeer:
I like the `warzone` and `war-torn` tags since they seem to add new unique information to the game that is useful for mission writers.

For the other tags I feel that they add some redundancy or are not sufficiently specific:
- The game should be able to determine the information encoded in the topographical tags automatically and make this available for mission writers; it doesn't do this at the moment, but I feel the game should just to that automatically without needing attributes for it. Benefit is that map changes are then also automatically applied without a lot of manual work.
- The game should also be able to automatically determine wormhole locations and make them available for mission writers. And not all wormholes are the same (some need a keystone, others can just be travelled through, some are 2-directional, others are 1-directional).
- The governments already tell me where Humans, Hai, Korath and Coalition species live. Adding tags for those feels somewhat redundant.
- I don't see how the hazard tag would be used. It seems too generic for most missions; `thrill seekers like to see the dangerous phenomenon in ...` sounds too generic. While it also seems too specific for missions that want to avoid all dangerous places. And also here, not all hazards are the same, some might attract tourists while others are really things to avoid at all costs (thrill seekers usually have a desire to have a good chance of surviving the thrilling experience).

Using the exclamation marks in front of the tags is also a no-go for me; especially since an exclamation mark is used in many programming languages to express `not`, which would give the tag exactly the opposite meaning of what you want to convey.

beccabunny:
> others are 1-directional

This is not currently possible, as you can give a wormhole no sprite to try simulating it but the landing autopilot will still try (and fail) to land on spriteless planets and the minimap will show it as well. Maybe usable in a plugin, but way too unpolished to make it into vanilla as it is currently.
If you meant the wormhole sequence in the middle of the EW, I don't think they are different enough imo from the behaviour of normal wormholes to deserve being differentiated, as I don't really see possible missions making use of this

> Using the exclamation marks in front of the tags is also a no-go for me; especially since an exclamation mark is used in many programming languages to express not, which would give the tag exactly the opposite meaning of what you want to convey.

Agreed on this point, but I believe they were left there only for ease of making changes to the PR. They can go now that it's not a draft imo

Zitchas:
You weren't around for some of the discussion, petervdmeer, but none of the ! tags are intended to be permanent, or even merged. Right now Lake's just tagging everything with those, because nothing in-game uses it; thus makes it much easier to find-and-replace the tag with whatever we decide on should be the *actual* attribute tag.

Galaucus:
Aye - while the PR is being discussed and edited I'm leaving the ! in to facilitate editing. Once there are no further concerns with the PR I'll go in and replace them with more appropriate text strings.

Amazinite:
Attributes should have an actual purpose for use in location filters, not just be there for the sake of being there. Many of these are redundant with existing location filters. For example, `coalition` is redundant with `government "Coalition" "Heliarch"` or `near "Quaru" 100`. Same deal with `human`, you can just do `near "Sol" 100` and `not government "Quarg"`. The regional attributes like `kimek` in Coalition space or `dirt belt` in human space are redundant with the planets within those systems that have those attributes, so it seems to me like it'd make more sense to just have system inherit the attribute of their planets.

MasterOfGrey:
> * The governments already tell me where Humans, Hai, Korath and Coalition species live. Adding tags for those feels somewhat redundant.

>The regional attributes like kimek in Coalition space or dirt belt in human space are redundant with the planets within those systems that have those attributes, so it seems to me like it'd make more sense to just have systems inherit the attribute of their planets.

@petervdmeer @Amazinite Uninhabited systems do not necessarily possess the government, despite arguably being valid targets for many concepts. Additionally, perhaps it doesn't matter which faction, merely the area. Post FW for example there are pirate, independent, uninhabited, one single republic and FW systems all in the same area. Yet this whole 'southern' area can be conceptually defined distinctly in a way that can't be safely done with governments, or `near <anything>`.

It is certainly an easy concept to inherit planet tags, but it leaves a lot to be desired compared to an intentional delineation.

Zitchas:
For Remnant missions, I predominantly want specifically labelled things, such as geographic areas that *includes* uninhabited systems, systems that change government (or lose governments entirely), etc.

While values that are inherited makes sense for governments and other things that are person-centric, I think having things concretely nailed down with manual labels makes more sense for stellar, interstellar, and non-person-centric descriptors.

Galaucus:
Any next steps to take on this?

Zitchas:
1. Automatic attributes makes sense for government. (How will it handle having multiple government in one system, though?)
2. Automatic attributes do not makes sense for uninhabited systems, because there is nothing for it to automatically parse. What would it do? Try to get a regional attribute from the nearest inhabited planet? Average together the values of surrounding systems? Given that regional names do not necessarily make sense from a logical perspective (nor should they - borders are legendary for not being logical.)

As such, to me it seems like - at best - a method for automatically assigning certain attributes should be limited to governments and/or species, and for regional things it would be at best a tool to help ease the workload as opposed to being a built-in law.

A different thing that comes to mind is that it would be useful if Coalition space could have some identifiers. @Arachi-Lover I don't really know how the Coalition would want to divide up its territory. If races are found in blocks, then that could work; but it seems just as likely that they are at least somewhat intermixed; thus having named regions/territories could be useful. From my perspective, three territories would be good. Basically, an East, West, and Central district. That's fairly boring, though.

Arachi-Lover:
they are intermixed, yes. also, the planets themselves already have attributes regarding the species, so I'm not sure how the system attributes could add something there (granted, I haven't messed around with attributes a lot, so might be missing something)

Zitchas:
The main thing is that for some Remnant jobs I will be sending the player to random systems in general areas of the galaxy. Ideally, the pay would be based on jump distance, but our pay system as best I can tell just handles hyperjump distances, not JD jump distances.

So, instead of having one generic job that could theoretically go *anywhere*, which might lead to the player being paid 500k for a 1 jump job, and then the next one they get 500k for a job where they have to go all the way to 4 pillars and back.

And I specifically want these random jobs to be able to go to uninhabited systems, and they do *not* involve landing on planets.

Right now, I could get something kind of aproximate to it by doing things like "is within 0 1 jump of a Free World system" and "is within 0 2 jumps of a Coalition system", etc. to get rough areas based on the main government, but it isn't very exact.

So for the job in the Astral set, just fixing the pay scaling so it counts absolute jumps instead of hyperdrive ones would solve the problem.

Beyond this, though. Later on I will want to be able to send missions to random specific sectors for plot purposes, and at that point, having each region of the galaxy labeled in the system files would be very useful.

Amazinite:
> Uninhabited systems do not necessarily possess the government

Most uninhabited systems in the game do have a government. Just click on any uninhabited system in human, Hai, Coalition, or Wanderer space. Even Kor Ak'mari is under the Korath government as an uninhabited system, and in automata space there are uninhabited systems with Kor Sestor or Kor Mereti as their government. The only uninhabited systems with the uninhabited government are in the Ember Waste, Sag A*, the nova systems, Host, and some systems in automata space that are under the influence of none of the governments in the region, including Mesuket which is just a warzone.

> Automatic attributes makes sense for government. (How will it handle having multiple government in one system, though?)

Well there's no need to create an automatic attribute for a system's government because the government can already be used as part of the filter. "System X is owned by Y government" is different from "system X contains planets from Y/Z governments," though. Right now, the government of the system is the government that would be used for location filters. So Tarazed for example won't trip a `government "Quarg"` system filter.

<hr/>

I'd say ditch the following:
* Coalition (redundant with `near <any Coalition system> 100` or `government "Coalition" "Heliarch")
* Human (redundant with `near <any human system> 100` and `not government "Quarg"` or `government <all major human governments>` like jobs do)
* Hai (redundant with `near <any Hai system> 100` and `not government "Quarg"` or `government "Hai" "Unfettered")
* Pleiades (there's no real reason for this one to exist, as any mission that for some reason wanted to return to the Pug galaxy could just name the systems)
* Efreti (redundant with `government "Kor Efreti" since their influence doesn't extend to any uninhabited systems)
* Wanderer (redundant with `near <any Wanderer system> 100`, at least up until the Unfettered show up and they go to Korath space, but you haven't changed this attribute according to events either)
* Korath (it doesn't make sense to me to only have only attribute which is "historical" while all other attributes are for "right now")

Zitchas:
I suppose I wasn't really clear - What would be more accurate is - for the purposes of the missions I want to do - I don't care about governments. I only care about the geographic region. Ideally, I'd like to see geographic labels Hai, Wanderer, Korath, and Coalition space. Something that - like the areas in human space - are actually names of regions, not just "Korath X" or "Hai Y". I'm sure the Coalition has some names for there larger areas.

And beyond that, I want geographic labels so that my missions are independent of whatever plot or stories come in the future. If, somehow, the Korath become extinct, these missions keep running. Or if the Coalition gets forced into exile and all move into human space, these missions to the now former coalition areas keep running without a hitch. The point is, I don't know what the future holds for missions, and the engine has the capacity to radically change things (including making planets and governments vanish entirely), and I want to future-proof these missions so that whatever happens with those governments and areas, the authors can write them without any concern over how they might impact my jobs and missions.

Having manually assigned regional attributes that identify the name of the region (which is the baseline of what I want - any other descriptors are a bonus) enables that to function, specifically because these basic regional identifiers will never, ever change. Most of the other ones (including governmental or species ones) probably should be automatic. Things that can change makes sense to have automatic tagging.

Amazinite:
> I want geographic labels so that my missions are independent of whatever plot or stories come in the future.

Than any such missions that require geographic labels should also introduce those labels. That's why I said in my first comment that attributes should exist for the purpose of actually being used instead of existing just to exist, which it seems to me is the case right now.

Zitchas:
I can't attribute much motivation to others, but in this case at least one motivation on the part of Galaucus was to do the manual labor of labeling the regions for me so that I could focus on writing the actual jobs & missions that will make use of them. They went above and beyond my initial requirements, but I see plenty of ways to make use of them, so I don't consider it either wasted effort or pointless data.

Galaucus:
Sorry folks, I really should have clarified that these labels are intended to be strictly geographic/historical - it's why I mentioned that the Korath tag extends to the limit of everywhere they've settled in the past, even if they're not there now.

The main question I asked myself when deciding how to tag a system was "If an explorer from society X got lost, would you be able to find their derelict drifting in this system?" 

The fact that every species is organized into an island of sorts made that easier to answer.

arkhne:
I think some automatic attribute sharing between planet and system is a good idea, and should probably be looked into by somebody way smarter than I am.

As for other attributes, I'm not sure whether they exist or not, existing lore distinctions such as "Dirt Belt", "Paradise Worlds", "The Rim", and the "warzone" and such that was discussed are too useful not to have. If I have to write some generic Jobs specifically using these regional distinctions to justify their addition, so be it. 

A little effort for something that "should be in place" or is "too useful to not support" is a small price to pay for some of us. You can *sort of* achieve this sort of effect with distance from X, a near Y, but "sort of" is not what mission writters want. I'm also the sort who believes a clean-cut and efficient system is much better to work with than wrestling with the existing stuff to make it work, especially when somebody has either offered to do the heavy lifting, or even already done it as is the case here.

I can understand wanting to skip redundant tags, and those that are too niche, however, I think this PR in general is a step in the right direction, and merely needs serious discussion on which tags should and shouldn't be used. I feel that the overall impact (such as file bloating) is low enough that it needs to be at least taken seriously and discussed to find the most efficient addition. I'm most definitely NOT saying "Yes, lets go and merge the whole thing as is", I'm just saying this needs more debate.

Amazinite:
North, deep, paradise, core, near earth, rim, dirt belt, south, arach, saryd, kimek, and the additional ember waste attribute I'm all fine with since they give information that can't easily be filtered already and they can easily be used for jobs.

The ones I say we should definitely ditch are the redundant ones, which I've already mentioned. (Hai, wanderer, exiles, efreti, coalition, and human are redundant with existing government filters.)

Automata is fairly redundant as well with `neighbor government "Kor Sestor" "Kor Mereti"`, as that substitutes all systems where Automata travel save for a single system (Korsmanath). The benefit of using this is that it also changes with the campaign as the Sestor disappear and their systems switch to the Uninhabited government.

Beyond redundancies, I think the other attributes are more useless, or at least I don't think they should be added unless a mission or job actually has a use for them.
* Pleiades? Why return there? Any return to the Pug galaxy (which is what the attribute should be, as the Pleiades is only the name of the image but the name of the galaxy is the Pug galaxy) should be mission based, not job based, in which case you'd be filtering the systems by name, not by attribute.
* Hazard? Seeing as how it's currently only one system that has it, it seems pointless to me. Even with system hazards becoming a thing, I don't see how Sayaiban and wherever else hazards may be used would have enough in common to all fall under a single job/mission waypoint filter. (e.g. you could have two systems on opposite sides of the galaxy that both have the hazard attribute. Why would a job ever use this attribute with a range that massive?)
* Black hole is similarly only one system. Why label it?
* The critique for warzone is the same as the one for hazard; separate "warzone" systems are different enough that it doesn't make sense for them to both fall under a single job/mission.
* For warzone, black hole, hazard, wormhole, and nova, you also have to think about how any mission using these attributes for a location filter means that those missions could *reveal* warzones, black holes, hazards, wormholes, and novae unknown to the player. Such a situation shouldn't be the case, as these features should be found through exploration, not random chance by a location filter.

Amazinite (changes requested):
My latest comment constitutes my review.

MasterOfGrey:
@Amazinite here are some answers to those questions/counter-comments:

`The ones I say we should definitely ditch are the redundant ones, which I've already mentioned. (Hai, wanderer, exiles, efreti, coalition, and human are redundant with existing government filters.)`
It's heavily implied that when Wanderer's act 3 comes along, the wanderers will no longer occupy much/any of their current space. 
- Jobs to that region of space may not care per se which faction is actually in charge of it, so a tag is of value.
- You have brought up the possibility of a human-hai war in a future plot, government shifts there again have the potential to mess with things without a tag.
- There are 8 different egion tags for human space now. Maybe, just maybe, there's value to having just one tag for some things instead of having to itemise all the tags in a mission/job?

`Hazard? Seeing as how it's currently only one system that has it, it seems pointless to me. Even with system hazards becoming a thing, I don't see how Sayaiban and wherever else hazards may be used would have enough in common to all fall under a single job/mission waypoint filter. (e.g. you could have two systems on opposite sides of the galaxy that both have the hazard attribute. Why would a job ever use this attribute with a range that massive?)`
You are missing the potential for crossover. Especially as hazards become more common in other spaces, jobs could target <has government> and <has hazard>

`Black hole is similarly only one system. Why label it?`
See prev.

`The critique for warzone is the same as the one for hazard; separate "warzone" systems are different enough that it doesn't make sense for them to both fall under a single job/mission.`
See Prev.

`For warzone, black hole, hazard, wormhole, and nova, you also have to think about how any mission using these attributes for a location filter means that those missions could reveal warzones, black holes, hazards, wormholes, and novae unknown to the player. Such a situation shouldn't be the case, as these features should be found through exploration, not random chance by a location filter.`
#5395 should cover this.

Amazinite:
> #5395 should cover this.

>**Additional context**
>Lol no this isn't me saying merge those attributes after this is done. We still need missions to actually use them.

:james:

MasterOfGrey:
It doesn't *require* changing the design philosophy of jobs, it facilitates a parallel philosophy for different types of jobs.

`That is to say, there are no jobs that are solely region based right now that would still make sense should a different species occupy the area. So that again wraps back to the active need for attributes instead of putting them there for the sake of putting them there.`

Yes, there are none right now, there's also no ability to make them without doing this first. Z has already stated his intent to make a series of jobs using several of these tags. Actually applying the tags is a large and time-consuming process, which acts as a barrier to the development time to be invested into those jobs.
I understand the desire not to add things that aren't immediately needed, but this particular example is fundamentally a case of "build it and they will come".

Essentially the statement of yours that I quoted is a self-fulfilling prophecy: there are no jobs that need the attributes, so you don't put them in, so their absence serves as a barrier to putting in jobs that need the attributes, so there are no jobs that need the attributes.

If I use myself for an example; I might think "Oh, I wanna make 3-4 jobs that do X".
With tags already in place that is a matter of an hour or two of main effort. Without tags already in place, that time easily doubles, and *additionally* the entirety of that extra time is also boring and tedious. This is a significant hurdle for creators who are time-poor. The lack of an immediate use for a given tag just doesn't negate its value.

Even if a tag is perhaps a poor choice for use case ultimately; they add search utility if different tags are needed to be added, or if someone is trying to find possibly suitable locales for some cause or other, because they add additional searchability to the map file so that creators can narrow down what they need to look at without missing things, and without requiring an encyclopedic knowledge of every system in the galaxy.

Zitchas:
From my perspective, adding attributes to systems serves the same purposes as attributes on planets: They allow missions and jobs to use a variety of individual or combinations of attributes to create on the spot sets to pick from. For instance, other traits that I'm planning on making jobs for in the future would be looking for things like:
- Binary star system
- Rich asteroids (as in plentiful. Not necessarily valuable)
- Specific star types
- Stellar Phenomenon (Pulsar, Quasar, black hole, etc. One broad category for all of them would be sufficient)

While the Remnant will have a variety of jobs aimed at research purposes, I could see them being useful for things like:
- Prospecting companies asking the player to check out a potential system.
- Various universities looking to fund expeditions to various interesting phenomenon.
- Tourists wanting a cruise past a binary star. (luxury quarters special job)
- Interrogation tour (criminal mafia wants to interrogate a prisoner with a red dwarf background for effect. Brig special job)
- Movie filming (various stellar phenomenon with both large number of bunks and available cargo hold for sets)
- Movie filming (asteroid filled systems with large number of bunks and cargo)

OK, not so confident about that interrogation tour. But you get the idea. Lots of possibilities.

Galaucus:
I'm unclear how to proceed aside from re-labeling the tags to a permanent version without the !. Could anyone advise on if there are other changes that need to be made?

Amazinite (changes requested):
Copying over from GitHub in changing what I'll take:
> So we'd be cutting it down to `"dirt belt"`, `core`, `paradise`, `deep`, `south`, `north`, `rim`, `"near earth"`, `arach`, `kimek`, `saryd`, `korath`, `"ember waste"`, `nova`, `wormhole`, and `"wolf-rayet"`. Leave out the wormhole in the Pug galaxy from getting that attribute.

Issue #5346, opened 2020-09-15
Not synchronized between copyright file and some text files


oo13:
**Is your feature request related to a problem?
Some data files shows someone else is the copyright holder but copyright file shows only MZ.

- hai/hai news.txt
- human/marauders.txt
- human/frontier jobs.txt
- ka'het/ka'het.txt
- ka'het/ka'het ships.txt
- ka'het/ka'het outfits.txt
- ka'het/ka'het missions.txt
- korath/korath news.txt
- remnant/remnant missions.txt
- remnant/remnant news.txt
- sheragi/sheragi ships.txt
- sheragi/archaeology missions.txt
- sheragi/sheragi outfits.txt

**Related Issue Links**
I can't find it.

**Describe the solution you'd like**
Modify the catch-all element to this:

Files: *
Copyright: MZ **, devs, and contributers.**
License: GPL-3+

**Describe alternatives you've considered**
Append those copyright holders to copyright file.

**Additional context**
Generally a text file has many copyright holders. IMO, the first author is not very important.

tehhowch:
If both solutions are compatible with the machine readable format, then either is fine for me.

Zitchas:
In terms of the first solution, would that not turn into basically a list of almost everyone who has contributed to ES? modifications to text files are the simplest contributions, and I suspect that most of the 150+ contributors have changed at least one. Example:
```
Files: data/*
Copyright: 
 Michael Zahniser <mzahniser@gmail.com>
 Amazinite ...
 Tehhowch ...
 Pointedstick ...
 contributor 1 ...
 contributer 2 ...
 etc.
License: CC-BY-SA-4.0
```
Having a copyright entry that states all that has two problems. The first being that it effectively assigns every contributor a more or less equal copyright claim on every text file. Most of us (Devs excluded) have not even touched every data file, let alone made noteworthy additions thereto. 

The second being it will make for a very long entry, one that will probably require us to contact every contributor to find out a) what their preferred name for the entry is, and b) what contact info (if any) they want attached to it. This isn't as important as the attribution element, but it could still be quite significant.


[...]

PR #5380, opened 2020-09-27
fix(doc) Update copyright file.

oo13:
**Bugfix:** This PR addresses issue #5346.

## Fix Details
The file "copyright" doesn't indicate some authors.
This PR adds "endless-sky contributors" to the default item of copyright.

You may still claim your copyright if you want because this PR modifies only the default line.

## Testing Done
N/A

## Save File
N/A

### DIFF
--- a/copyright
+++ b/copyright
@@ -5,6 +5,7 @@ Source: https://github.com/endless-sky/endless-sky
 
 Files: *
 Copyright: Michael Zahniser <mzahniser@gmail.com>
+           endless-sky contributors (see credits.txt and changelog)
 License: GPL-3+
 
 Files: images/*


copyright
```diff
@@ -5,6 +5,7 @@ Source: https://github.com/endless-sky/endless-sky
 
 Files: *
 Copyright: Michael Zahniser <mzahniser@gmail.com>
+           endless-sky contributors (see credits.txt and changelog)
```
Amazinite, line 8: This should probably be a comment like with other entries.
```
Files:
 images/outfit/*battery?hai*
 ...
Copyright: Maximilian Korber
License: CC-BY-SA-4.0
Comment: Derived from works by Michael Zahniser (under the same license).
```
Perhaps also mention the copyright headers in the text and source files.

oo13: Do you recommend removing "endless-sky contributors"?

```
Files: *
Copyright: Michael Zahniser <mzahniser@gmail.com>
           endless-sky contributors
Comment: For more details, see credits.txt, changelog, and the copyright headers in the text and source files.
```
or
```
Files: *
Copyright: Michael Zahniser <mzahniser@gmail.com>
Comment: For more details, see credits.txt, changelog, and the copyright headers in the text and source files.
```

tehhowch: This is an appropriate format. It should not be a comment. Ref [the copyright format spec](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/):

> Formatted text, no synopsis: one or more free-form copyright statements. Any formatting is permitted; 

Ref "Formatted text":

> Formatted text fields use the same rules as the long description in a package's `Description` field in Debian control files. 

[Package `Description`](https://www.debian.org/doc/debian-policy/ch-controlfields#description):

> The lines in the extended description can have these formats:
-     Those starting with a single space are part of a paragraph. Successive lines of this form will be word-wrapped when displayed. The leading space will usually be stripped off. The line must contain at least one non-whitespace character.
-     Those starting with two or more spaces. These will be displayed verbatim. If the display cannot be panned horizontally, the displaying program will line wrap them “hard” (i.e., without taking account of word breaks). If it can they will be allowed to trail off to the right. None, one or two initial spaces may be deleted, but the number of spaces deleted from each line will be the same (so that you can have indenting work correctly, for example). The line must contain at least one non-whitespace character.
-     Those containing a single space followed by a single full stop character. These are rendered as blank lines. This is the only way to get a blank line.
-     Those containing a space, a full stop and some more characters. These are for future expansion. Do not use them.

> Do not use tab characters. Their effect is not predictable.


Also the last example in the copyright spec shows a similar use:
```
Files: *
Copyright: 2008, John Doe <jdoe@example.com>
           2007, Jane Smith <jsmith@example.org>
           2007, Joe Average <joe@example.org>
           2007, J. Random User <jr@users.example.com>
License: PSF-2
```

Issue #5379, opened 2020-09-27
Typo and bug in 'Investigating Kor Efreti Worlds'


roadrunner56:
**Describe the bug**
When performing the job 'Investigating Kor Efreti Worlds', the dialog upon landing on a Kor Efreti planet/station states that the Wanderers are there to investigate environmental problems on a Efreti world. However, this mission can have the abandoned station Tefkar Ret as its destination, which does not make sense in the context of the dialog and the larger story. 
In addition, the dialog that pops up when the player lands on a Kor Efreti planet has a typo: 'The Wanderers tour the planet to investigate the _envoronmental_ problems...'.  _envoronmental_ is spelt wrong.

**To Reproduce**
1. Accept a  'Investigating Kor Efreti Worlds' job on a Wanderer planet that has Tefkar Ret as its destination
2. Land on Tefkar Ret
3. Read description, see errors

**Expected behavior**
I expected the mission to only target planets, and not stations, as you cannot terraform a station. I also expected the dialog to be typo-free.

**Screenshots**
![Kor Efret](https://user-images.githubusercontent.com/65418682/94355169-27774680-003f-11eb-98ef-be60007d1487.png)

**System (please complete the following information):**
 - OS: Kubuntu 20.04
 - Game Source: Continuous
 - Version: Commit f67b5de

**Additional context**
I really like the Kor Efreti missions, they flesh out an often-forgotten faction nicely.

tehhowch:
Good eye! Looks like a (very minor) miss in these (very new) missions from @W1zrad -- dynamic stopover filters in 0.9.13 are a small bit more lenient in the landing locations they can select, so pure government filters may not be sufficiently specific.

W1zrad:
Thanks for catching these! This should be easy fixes; I can get to it in an hour or two. Happy to hear you like the missions!

PR #4867, opened 2020-03-01
feat(mechanics): Update random planet selection criteria

tehhowch:
**Feature:** This PR implements a missing feature I identified after a post on Discord by @Zitchas  

## Feature Details
A mission will not offer if a stopover is requested and none could be found. The procedure for picking a random stopover was the same as for picking a random destination, and uses `LocationFilter#PickPlanet`. This method's ancestor was originally implemented in Mission.cpp and featured sensible things like [this restriction](https://github.com/endless-sky/endless-sky/commit/af58e48f7324a650cc2ba7380a9dc142eb51a3a3). In endless-sky#3501 I moved these methods to LocationFilter, but did nothing about the mission-specific restrictions.

This PR
 - allows stopovers picked from a location filter to include non-spaceport planets
 - allows explicitly providing planet names to enable randomly picked planets to be wormholes, planets without spaceports, or planets for which the player does not have clearance
   - This applies even for mission/job destinations

Previously, to get a random planet to be a stopover or destination, the mission writer was required to provide `clearance`. With this PR, situations can arise where the player is forced to obtain landing permission on their own, either by bribing or raising their reputation, or obtaining clearance from some other mission. **These "get-your-own-clearance" situations require an explicit opt-in by the mission creator.**

## UI Screenshots
n/a

## Usage Examples
```
destination
    system "Civil War"
    planet "Friendly With Player" "Hostile With Player"
```
can now pick "Hostile With Player" and not just "Friendly With Player", while
```
destination
    system "Civil War"
```
can only pick "Friendly With Player", because there was not an explicit override.
(Previously, a mission had to provide `clearance` in order to pick a hostile destination randomly.)

## Testing Done
 - [x] Random missions still offer when landing on planets.
 - [x] Thorough inspection of existing missions to determine which are affected 

## Performance Impact
 - Trivial performance impact, as this method is only used when instantiating missions (e.g. on land or board) and not continuously in-flight.

PR #5381, opened 2020-09-27
(fix) New Wanderer jobs corrections

W1zrad:
Fixes issue #5379 opened by @ roadrunner56.

-Corrected a typo
-Ensured station could not be a stopover for the job by adding `not attributes "station"` to stopover

To the best of my knowledge and some quick testing, this fixes the issue, but @tehhowch said something that made me a little hesitant:
> dynamic stopover filters in 0.9.13 are a small bit more lenient in the landing locations they can select, so pure government filters may not be sufficiently specific

So it sounds like there could be some things going on under the hood that I'm not fully aware of. If someone with a bit more knowledge of the code could check this in regards to @ tehhowch 's comment it would be much appreciated.

This save can be used to test this:
[Firstname Lastname~test start.txt](https://github.com/endless-sky/endless-sky/files/5288233/Firstname.Lastname.test.start.txt)

### DIFF
--- a/data/wanderer/wanderer jobs.txt
+++ b/data/wanderer/wanderer jobs.txt
@@ -831,10 +831,11 @@ mission "Wanderers to Kor Efreti"
 	stopover
 		distance 2 8
 		government "Kor Efret"
+		not attributes "station"
 	on visit
 		dialog phrase "generic missing stopover or passengers"
 	on stopover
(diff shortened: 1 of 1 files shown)


tehhowch:
My comment was in regards to the functionality added by this PR: https://github.com/endless-sky/endless-sky/pull/4867

W1zrad:
Based on what I've read, it looks like there won't be any problem with fixing this issue the way I've done here (please correct me if I'm wrong). Thanks for the clarification!

PR #5384, opened 2020-09-27
feat: Use lambda expression for adding incomes in BankPanel.

petervdmeer:
**Feature:** This PR implements a small step towards the **on-demand conditions** as proposed in issue #5017

## Feature Details
PR #5244 already added some functions to abstract access to the conditions, but the GetConditionSum function from that PR looked a bit too specialized in a generic environment.
This PR makes a first step towards a more generic CollapseConditions function that should provide the same functionality of GetConditionSum from #5244, but in a more generic (and better performing) way by having the collapsing function (sum) provided as lambda expression parameter.

The CollapseConditions function should become a member of the class that contains the conditions in or after PR #5244 (and should get the conditions from the class instead of as parameter), but it doesn't look out of place in its current shape in BankPanel until the discussion there is done on how the new class will look.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
Checked if the BankPanel for one of my savegames gave the correct number for Salary and Tribute income.

## Performance Impact
The lamba-expression can convert directly from the input-conditions to the desired target-format, so performance should be better than for the GetConditionSum function (and should be near the original performance of the direct map-accesses).

The affected code is not performance critical, so I don't expect many people to care a lot.

### DIFF
--- a/source/BankPanel.cpp
+++ b/source/BankPanel.cpp
@@ -47,7 +47,7 @@ namespace {
 	// Helper function to collapse (summarize, multiply or otherwise combine) a number of
 	// conditions (that start with the same prefix) into a single value.
 	// The caller of this function should provide the lamba expression that performs the
-	// actual collapsing. This function just calls the collapse-funtion once for every
+	// actual collapsing. This function just calls the collapse-function once for every
 	// condition that matches the prefix.
 	template <typename A>
 	void CollapseConditions(const map<string, int64_t> &conditions, const string &prefix, A collapseFun)
--- a/source/BankPanel.cpp
+++ b/source/BankPanel.cpp
@@ -43,6 +43,19 @@ namespace {
 	
 	// Maximum number of rows of mortages, etc. to draw.
 	const int MAX_ROWS = 8;
+	
+	// Helper function to collapse (summarize, multiply or otherwise combine) a number of
+	// conditions (that start with the same prefix) into a single value.
+	// The caller of this function should provide the lamba expression that performs the
+	// actual collapsing. This function just calls the collapse-funtion once for every
+	// condition that matches the prefix.
+	template <typename A>
+	void CollapseConditions(const map<string, int64_t> &conditions, const string &prefix, A collapseFun)
+	{
+		auto it = conditions.lower_bound(prefix);
+		for( ; it != conditions.end() && !it->first.compare(0, prefix.length(), prefix); ++it)
+			collapseFun(it->second);
+	}
 }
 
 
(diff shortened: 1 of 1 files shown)


source/BankPanel.cpp
```diff
@@ -108,9 +121,10 @@ void BankPanel::Draw()
 	static const string prefix[2] = {"salary: ", "tribute: "};
 	for(int i = 0; i < 2; ++i)
 	{
-		auto it = player.Conditions().lower_bound(prefix[i]);
-		for( ; it != player.Conditions().end() && !it->first.compare(0, prefix[i].length(), prefix[i]); ++it)
-			income[i] += it->second;
+		CollapseConditions(player.Conditions(), prefix[i],
```
tehhowch, line 124: If generalizing this sum, I would expect us to use the standard library rather than reinvent the wheel. i.e. use `find_if`/`find_if_not` to determine the end of the range, and then pass the iterators to `std::accumulate`.

tehhowch (changes requested):

source/BankPanel.cpp
petervdmeer: The goal of this PR was not to make the iterators nicer, it was to check if we can get rid of the iterators completely for code that uses player-conditions.

Do we still want player.Conditions() to be iteratable after #5017? And would the iterator keys still need to be string (and sorted by string) or would a more complex type like pair<list-identifier, string> also be okay?

tehhowch: I would indeed expect that an iteration interface is exposed. Whatever solution we go with should be able to provide a `begin` and an `end` ForwardIterator; they don't have to point to the same container (but obviously would then need to implement the increment operator correctly).

petervdmeer:
Closing this PR; I processed the review-comments from this PR in https://github.com/petervdmeer/endless-sky/tree/conditions_provider and will merge that into #5244 when it is more finished.

source/BankPanel.cpp
petervdmeer: If we implement this iterator, should it then be an iterator over just the manually set conditions, or should it also iterate over all the automatically on-demand generated conditions?
I feel that generating conditions on-demand to just iterate over them and then discard the value if the key doesn't match is not a good use of CPU power.

tehhowch: That's a good point. Can the key and value be generated separately? That is, will we be able to identify the available condition keys without computing their corresponding values?

Dereferencing the iterator would then yield a function that returns the condition value, rather than the condition value itself.

petervdmeer: Yes, I think that would technically be possible, but what does it then mean to iterate over on-demand conditions?

Suppose that we at some point support get on-demand conditions like `named ship: [Ship Name]: Outfit: Energy Blaster` which tells us how many Blasters are installed on a ship with a certain name. Or even conditions like `tagged ships: [tag id]: Outfit: Energy Blaster` which tell us how many Blasters there are installed in total on a set of ships that share the same tag.

If we would want to iterate over this type of data in the C++ code, then I would expect us to iterate directly over the ships and not to iterate over the condition sets.

So far I only found 2 locations where there was an iteration over the conditions set:
- When the conditions are saved to a savegame (which only requires the manual conditions, not the on-demand ones)
- When the salaries and tributes are retrieved for display (which currently uses manual conditions, but might move to their own interatable class in the future)

When would we want to iterate over all possible (manual and on-demand) conditions?

PR #5390, opened 2020-09-30
Fix: Conversation branch condition in "Remnant: Face to Maw 2"

Ornok:
Changed branch condition `"remnant met taely" > 1` to `"remnant met taely" >= 1` so that Taely doesn't introduce herself again in "Face to Maw 2" if the player met her once before (e.g. by accepting "Tech Retrieval" without completing it).

### DIFF
--- a/data/remnant/remnant missions.txt
+++ b/data/remnant/remnant missions.txt
@@ -2264,7 +2264,7 @@ mission "Remnant: Face to Maw 2"
 		payment 1000000
 		conversation
 			branch taely
-				"remnant met taely" > 1
+				"remnant met taely" >= 1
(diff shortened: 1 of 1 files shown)


Amazinite:
Really, any branches like these should be using `has "remnant met taely"`. And also `has "remnant chilia"` for lines 2323, 2905

Amazinite (changes requested):
Use `has <condition>` instead of `<condition> >= 1` and fix the other similar errors in this file.

Ornok:
Such conditions are also used in intro log missions ("Remnant: taely intro log" line 1362 and "Remnant: Chilia Intro Log" line 1374). Should we do the same with those?

Amazinite:
Those ones are fine since there are multiple log missions, and the use of `>= x` makes it explicit that "this one requires X encounters, this one requires X+1 encounters" and so on.

PR #5394, opened 2020-10-01
Spacing fix on hail

Terin:
Missing space in one of the new merchant hails


Amazinite (changes requested):
Move the space to the beginning of the following words so that this hail doesn't potentially end with a trailing space.

PR #5398, opened 2020-10-02
Change Picture of Kestrel in "Kestrel Available" from top-down to thumbnail

Anarchist2:
This matches the precedent of using thumbnail images to show ships in conversations, such as the Tempest, Penguin and Emerald Sword.


Zitchas (approved):
I'd say that is a good change. The thumbnail is much more evocative than the sprite.

Issue #5399, opened 2020-10-03
"You have visited" map key should recognize unvisited planets in uninhabited systems


Amazinite:
**Is your feature request related to a problem? Please describe.**
The "You have visited" map key (switched to by clicking `(not yet visited)` or `(has been visited)` under a planet menu on the left side when in port mode) currently prioritizes showing a system as uninhabited over showing whether the landable planets in the system have been visited.

![image](https://user-images.githubusercontent.com/17688683/94984552-2f415a00-051b-11eb-9b1f-a4f682ea8eb5.png)
![image](https://user-images.githubusercontent.com/17688683/94984554-35373b00-051b-11eb-904b-d9322857cca7.png)

**Related Issue Links**
N/A

**Describe the solution you'd like**
Only highlight a system as uninhabited if it contains no landable planets. Otherwise, show if you've visited all the landable planets in the system.
Planets that are unlandable due to an attribute requirement should still show as uninhabited.

**Describe alternatives you've considered**
N/A

**Additional context**
Relevant code:
https://github.com/endless-sky/endless-sky/blob/c77348a08ed60db698227c4d14aadf7856059651/source/MapPanel.cpp#L702-L744

TheDeathStar3:
Hi, I am using the 0.9.13 beta version of Endless sky and uninhabited planets are still not marked as visited or not visited for me. I have visited every single planet and system in the game. I tried to download the continuous version on GitHub and use that but that still did not work. Is anyone else having this issue or is it just me? I did use new pilots and just transferred the visiting planets and system statuses across the save files.

tehhowch:
> Hi, I am using the 0.9.13 beta version of Endless sky and uninhabited planets are still not marked as visited or not visited for me. I have visited every single planet and system in the game. I tried to download the continuous version on GitHub and use that but that still did not work. Is anyone else having this issue or is it just me? **I did use new pilots and just transferred the visiting planets and system statuses across the save files.**

Save-editing is not a supported use case. Please verify on a new pilot with no installed plugins and no save file edits.

TheDeathStar3:
I confirm that on a new pilot, with no plugins and no save file edits the uninhabited planets still do not show if they have been visited or not.
![Screen Shot 2021-04-20 at 6 59 48 am](https://user-images.githubusercontent.com/54761256/115305503-c5805900-a1a9-11eb-8421-ee760df38630.png)

PR #5402, opened 2020-10-04
"You have visited" map key highlights uninhabited systems with landable planets

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #5399.

## Feature Details
<insert the title here\>

## UI Screenshots
Map system coloring using:
* commodity key (unchanged, notice Algieba)
image
* visited key (now highlights uninhabited systems with landable planets, notice Algieba)
image
* visited key (looking at a restricted planet w/o access)
image
* visited key (looking at a restricted planet w/ access)
image

## Usage Examples
N/A

## Testing Done
See UI screenshots section.

## Performance Impact
N/A

### DIFF
--- a/source/MapPanel.cpp
+++ b/source/MapPanel.cpp
@@ -699,11 +699,12 @@ void MapPanel::UpdateCache()
 		Color color = UninhabitedColor();
 		if(!player.HasVisited(&system))
 			color = UnexploredColor();
-		else if(system.IsInhabited(player.Flagship()) || commodity == SHOW_SPECIAL)
+		else if(system.IsInhabited(player.Flagship()) || commodity == SHOW_SPECIAL || commodity == SHOW_VISITED)
 		{
 			if(commodity >= SHOW_SPECIAL)
 			{
 				double value = 0.;
+				bool colorSystem = true;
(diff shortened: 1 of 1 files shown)


MasterOfGrey:
`visited key (looking at a restricted planet w/o access)`
I feel like this should look different to the systems with no planets at all still.

Amazinite:
Then that would reveal the existence of attribute locked planets before knowing how to access them.

MasterOfGrey:
Right, but if you've bought a map to be able to see the system and know roughly what's in it without having visted (which is the only way I can imagine you'd know about the visitable planets), surely that map would also tell you you probably can't access a planet in the same way?

To be honest I'm not exactly sure, but I don't think hiding the locked planet when unlocked planets are visible makes a lot of sense when your information is enough to know about the system but limited by not having visted. The planet is there either way.

Amazinite:
The behavior I've used here matches the behavior of the orbit map. Planets you do not have access to do not show regardless of whether you've visited.
image
image

MasterOfGrey:
Huh, ok, I'll chalk that up to 'behaviour I find weird in ES' and leave that be then.

PR #5404, opened 2020-10-04
Allow gifted ships to be defined in place

### LINKED ISSUE #5400: Gifting a ship variant instead gifts the base model if a save is reloaded
**Describe the bug**
It is currently possible for a mission to gift a variant of a ship, but should the player reload their save file while an active mission gifts a ship variant on complete or on fail, that variant will be converted to the base model on the reload.

**To Reproduce**
Take this job from the spaceport. Launch and land. Observe how both gifted Kestrels have 12 gun ports. Then take the job again, launch, reload the save (this will give you the on complete Kestrel right away, see #4108), and take note of how the third Kestrel you were given (the on offer of the second mission) has 12 gun ports, but the final Kestrel given only has 6, as this one was converted to the base model when the save was reloaded.
```
mission "gift test"
	repeat
	job
	on offer
		give ship "Kestrel (More Weapons)"
	on complete
		give ship "Kestrel (More Weapons)"
```

**Expected behavior**
MissionActions should remember that they are suppose to be gifting a variant instead of the base model.

**Screenshots**
N/A

 - Game Source: self-compiled
 - Version: latest commit as of posting (7c85825)

**Link to save file**
N/A

**Additional context**
The culprit here is that MissionAction only saves out the ship model. The ship model refers to the base model though, not the name of the variant. The simple solution would be to have ship variants remember their variant name and have MissionAction save out that (e.g. Ship::VariantName() that returns modelName if the ship isn't a variant). The more "complete" solution though would perhaps be to save out the entire ship definition, which was mentioned in #5302 as being something we potentially do in the future anyway so that a MissionAction can define a ship to give.

Amazinite:
**Feature:** This PR implements the feature request discussed in issue #5400 and #5302.

## Feature Details
This PR solves #5400 by allowing gifted ships to be defined within the mission. Any gifted ships that are only given by name get saved to the save file as a full ship definition, meaning that any named variants get saved in their variant configuration, and therefore gifted in their variant configuration even after reloading a save without the need to save the variant name.

Note that much of the distinction between giftShips and giftStockShips and how they both get merged into giftShips upon instantiation has been lifted from NPC.cpp.

## UI Screenshots
N/A

## Usage Examples/Syntax
```
on *
	give
		ship "<model name>" ["<name>"]
			<ship definition here>
```

## Testing Done
Same testing as #5400.

## Performance Impact
N/A

## Additional Context
Near the end of finishing this PR, I've decided that I think that #5403 is the better solution.

One of the potential reasons why we would allow a gifted ship to be defined in place and therefore have the ship's entire definition be saved to the save file is to allow on-the-fly changes to a ship definition. For example, perhaps you want a gifted ship that is never disabled, you can reference the base model and only change that tag. This is akin to how this is possible to do with `npc` blocks.

After looking through the game though, I can't recall or even find any instance of a ship actually being defined within an `npc` block. All instances of unique ships used for `npc` blocks are defined outside of the mission itself, often just above or below where the mission is defined, but still outside of it. Then looking at the wiki, while it is possible to define a ship entirely within an `npc` block, it actually looks to me that this is undefined behavior, as it is not listed on the wiki. The relevant section of code in NPC.cpp also mentions how having an entire ship definition in an `npc` block is intended for loading from a save file. With gifted ships though, this is entirely unnecessary to do, as a gifted ship from a mission doesn't need to be kept track of by the mission in the same way as an NPC does. NPC ships are written in their entirety to the save file because we need to keep track of their health and other such important stats. For gifted ships though, all we care about is the model and name of the ship, so that's really all we should be saving. And seeing as how the convention is already to define the variant outside of the mission that uses it, I don't see an issue in requiring that a gifted variant be defined outside of the mission as well.

An important thing to note as well is that if you're defining a ship within a `give` or `npc` block, you actually either HAVE to name it on the same line or HAVE to give an ENTIRE ship definition. This is because Ship.cpp only recognizes that a ship is a variant if there is a token after the model name. This informs the game that the given ship has a base model that it should use to fill in any missing information. Otherwise, it thinks it has a whole ship definition, which may not be the case. The following will crash the game if you try to view it anywhere, as you're creating a ship with no attributes and the game doesn't recognize that this ship has a base model since the line with `ship` doesn't have three tokens.
```
give
	ship "Kestrel"
		"never disabled"
```
What this means is that you can't gift a ship that is defined in place with a randomized name unless you're giving an entire ship definition.

Overall, this just seems like a less refined way of doing this as compared to the status quo of defining variants outside of the mission that uses them and simply using #5403 to fix #5400. I'll leave it up to @tehhowch though as to which approach we use, because there's really no harm in using this approach outside of being aware of its pitfalls.

### DIFF
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -33,7 +33,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	void DoGift(PlayerInfo &player, const Ship *model, const string &name, UI *ui)
+	void DoGift(PlayerInfo &player, const Ship *model, const string &name)
 	{
 		if(model->ModelName().empty())
 			return;
@@ -195,10 +195,23 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			conversation.Load(child);
 		else if(key == "conversation" && hasValue)
 			stockConversation = GameData::Conversations().Get(child.Token(1));
-		else if(key == "give" && hasValue)
+		else if(key == "give")
 		{
-			if(child.Token(1) == "ship" && child.Size() >= 3)
-				giftShips[GameData::Ships().Get(child.Token(2))] = child.Size() >= 4 ? child.Token(3) : "";
+			if(child.Size() >= 3 && child.Token(1) == "ship")
+				giftStockShips[GameData::Ships().Get(child.Token(2))] = child.Size() >= 4 ? child.Token(3) : "";
+			else if(!hasValue && child.HasChildren())
+			{
+				for(const DataNode &grand : child)
+					if(grand.Token(0) == "ship" && grand.Size() >= 2)
+					{
+						shared_ptr<Ship> ship = make_shared<Ship>(grand);
+						if(grand.Size() >= 3)
+							ship->SetName(grand.Token(2));
+						giftShips.emplace_back(ship);
+					}
+					else
+						grand.PrintTrace("Skipping unsupported \"give\" syntax:");
+			}
 			else
 				child.PrintTrace("Skipping unsupported \"give\" syntax:");
 		}
@@ -314,7 +327,14 @@ void MissionAction::Save(DataWriter &out) const
 			conversation.Save(out);
 		
 		for(const auto &it : giftShips)
-			out.Write("give", "ship", it.first->ModelName(), it.second);
+		{
+			out.Write("give");
+			out.BeginChild();
+			{
+				it->Save(out);
+			}
+			out.EndChild();
+		}
 		for(const auto &it : giftOutfits)
 			out.Write("outfit", it.first->Name(), it.second);
 		for(const auto &it : requiredOutfits)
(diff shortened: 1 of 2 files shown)


petervdmeer:
I prefer the simpler "save the model" approach from #5403; each mission that you can write with the "save the full ship" features here can also be written with the simpler approach. Let's not add complexity when it doesn't also add significant value.

The difference between this PR and #5403 does matter when the definition of a given ship changes between saving with an active ship-giving-mission and re-loading. This approach will by default give the ship according to the "old definition at to mission instantiate-time", while #5403 will give the ship according to the "definition at the time of actual giving".
But we don't change ship definitions often and I would argue that if we update a ship definition, then it would be fine to give the updated ship. (And if some mission really needs a very specific never-changing variant, then the mission writer can choose to define that exact variant just for the mission and put a comment in the datafiles to never change that definition.)

Amazinite:
Closing this in favor of #5403. After thinking these two PRs over more, I just don't see this PR as being the solution to the issue given its own issues that it has.

PR #5409, opened 2020-10-05
Add new human mission

AlphaBeta1906:
----------------------
**Content (Missions)**

## Summary
This mission is about a group of archaeologist that found a strange artifact in Asgard of Naos system.The arthifact is so strange and looks very very old. That group of researcher don't have enough money left to pay transport so players come and offer a help to bring them to earth for study the artifact even furhter.

What di you think about this mission?

### DIFF
--- a/data/human/transport missions.txt
+++ b/data/human/transport missions.txt
@@ -3239,15 +3239,15 @@ mission "The Artifact"
 		random < 20
 	on offer
 		conversation
-			`As you leave your ship and walking to the spaceport canteen, you see a group of researchers talking, they seem so confused because doesn't have any money to transport them back to Earth`
+			`As you leave your ship and walking to the spaceport canteen, you see a group of researchers talking, they seem so confused because doesn't have enough money to transport them and the Artifact to Earth`
 			choice
 				`	(offer them transport..)`
 				`	(ignore.)`
 					decline
 			`"	Hey, can I help you?" you ask them` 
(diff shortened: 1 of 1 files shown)


W1zrad:
Don't have time to fully review now, but from a quick glance I'd strongly recommend checking your spelling, grammar, and punctuation. Once those are a little more clean, it will be easier for people to discuss the actual content.

Amazinite (changes requested):
You still have quite a few errors with spelling, grammar, formatting, and style. (If I pointed them all out I'd practically be rewriting this for you.)

Beyond those issues though, the story itself here is lacking. With something as suspicious as "we found this strange artifact and need it transported," I'd expect more than just a single mission. Given though that the artifact was found in the Deep. I doubt it would be transported to Earth, given that the Deep are rather secretive with these types of things. (See Deep Archaeology for such an example where you're tracked by the Deep just for grabbing a splinter of wood from an old church building.)

I'd suggest simply closing this.

AlphaBeta1906:
> You still have quite a few errors with spelling, grammar, formatting, and style. (If I pointed them all out I'd practically be rewriting this for you.)
> 
> Beyond those issues though, the story itself here is lacking. With something as suspicious as "we found this strange artifact and need it transported," I'd expect more than just a single mission. Given though that the artifact was found in the Deep. I doubt it would be transported to Earth, given that the Deep are rather secretive with these types of things. (See Deep Archaeology for such an example where you're tracked by the Deep just for grabbing a splinter of wood from an old church building.)
> 
> I'd suggest simply closing this.
Well, I knew this was going to happen so I was just thinking about the other parts, but it's okay I just realized that I could create a mission just by changing the mission file. about grammar problems, I'm not really a native speaker, and yes I will fix it 
thanks for your review,it Will be lesson for me:)

PR #5410, opened 2020-10-05
Add Fuel Pods and Bunks to Coalition outfitters

Terin:
This PR adds Fuel Pods and Bunk Rooms to Coalition outfitters. For a space so large (15 jumps from end-to-end), and with the Coalition being a largely peaceful, commerce focused society, additional fuel capacity would be a valuable option for Coalition captains. This would be consistent with the Hai as well, an alien space where these outfits are also available.

Bunks may be more tenuous fictionally, but the Kimek Spire description points out how generic bunks are viable, just not comfortable for all three species.


Zitchas (approved):
These are very generic outfits. Makes sense to me.

PR #5412, opened 2020-10-06
New auto conditions: visited locations, installed outfits, and total ships

### LINKED ISSUE #2494: enhancement for missions: allow `require` in `to *`
As of now, the `require <outfit>` special condition can only be used in any `on` mission action.
During a longer conversation outside of this Git, it became clear that:
- `require <outfit>` should be available in the `to` sections of a mission. `to offer` and `to fail` are the most relevant ones.
  - `to offer` is the place where it would usually (logic wise) belong to instead of `on offer`.
  - `to fail` will allow a mission to fail if the player acquiers an outfit after the mission was accepted (i.e. while it is active)

Furthermore, it would be desirable, to allow the `require` construct to accept a number. This is not possible with the current implementation though. For numbers greater than zero, this would mean, that the player should have *at least* `<n>` units. For `<n> <= 0` it would require the player to *not have* any units (ideally accross all ships) - thinking of something "illegal" or "not liked on planet xy" here.

Furthermore, it might be desirable, to extend this from outfits to cargo in general.

### LINKED ISSUE #3735: Add visited planets and systems to player conditions
**Current situation:**
The game does record which systems and planets have been visited, but it does not do so in the conditions section. 

Thus if a mission creator wants to query whether or not a player has visited a current system or planet, they need to set an invisible mission that triggers upon visiting that planet, and then use the presence of that invisible mission to indicate that someone has visited the planet.

**Desired behaviour:**
Having the game record systems and planets visited in the conditions section so that it can be checked directly.

**Benefit:**
 - Allowing content creators to customize conversations using branches or conditions that relate to planets or systems that the player has (or has not) visited, which makes the game feel more responsive to the player's individual journey.
 - As mission creators add new content, they will have a reasonable degree of reverse compatibility since pilots will already have the systems and planets logged in their pilot file, so no need to worry about pilots having to return to visit obscure worlds after the new content was added in order to trigger invisible missions.
 - Reduce the need for invisible missions

Amazinite:
**Feature:** This PR implements the feature request discussed by https://github.com/endless-sky/endless-sky/issues/2494#issuecomment-704409776.
Effectively resolves #2494.
Resolves #3735.

## Feature Details
Adds the following auto conditions:
* `"total ships"`: How many ships the player has in their fleet, irrespective of their location or parked status.
* `"visited planet: <planet name>"`: Exists for each planet the player has visited.
* `"visited system: <system name>"`: Exists for each system the player has visited.
* `"outfit: <name>"`: Exists for each outfit the player has installed on any ship that is not disabled or parked in system. (Same behavior as `"ships: <category>"`.)
* `"flagship outfit: <name>"`: Exists for each outfit installed on the player's flagship. These outfits are also counted in the previous condition.

## UI Screenshots
N/A

## Usage Examples
They're conditions. You use them like any other condition.

## Testing Done
Loaded the game with these changes. The proper conditions were written to the save file. Conditions were updated after changing my fleet composition. Unvisiting systems/planets removed those conditions.

## Performance Impact
N/A

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2356,7 +2356,8 @@ void PlayerInfo::UpdateAutoConditions(bool isBoarding)
 			for(const auto &outfit : ship->Outfits())
 				conditions["outfit: " + outfit.first->Name()] += outfit.second;
 		}
-		++conditions["total ships"];
+		if(!ship->IsDestroyed())
+			++conditions["total ships"];
 	}
 	// If boarding a ship, missions should not consider the space available
 	// in the player's entire fleet. The only fleet parameter offered to a
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2339,18 +2339,25 @@ void PlayerInfo::UpdateAutoConditions(bool isBoarding)
 		if(first != last)
 			conditionsMap.erase(first, last);
 	};
-	// Clear any existing ships: conditions. (Note: '!' = ' ' + 1.)
+	// Clear any existing ships: or outfit: conditions. (Note: '!' = ' ' + 1.)
 	clearRange(conditions, "ships: ", "ships:!");
+	clearRange(conditions, "outfit: ", "outfits:!");
 	// Store special conditions for cargo and passenger space.
 	conditions["cargo space"] = 0;
 	conditions["passenger space"] = 0;
+	conditions["total ships"] = 0;
 	for(const shared_ptr<Ship> &ship : ships)
+	{
 		if(!ship->IsParked() && !ship->IsDisabled() && ship->GetSystem() == system)
 		{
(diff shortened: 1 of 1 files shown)


MasterOfGrey:
Heya, with the `"outfit: <name>"`, if you capture a ship with the relevant outfit, will it correctly detect that you have it on that ship the first time you land? Or will it have to depart once first so it's fully coloured as 'yours' before it registers upon landing?

Amazinite:
It registers whenever UpdateAutoConditions is called:
1) Game start.
2) On each jump/planet departure.
3) On landing.
4) On offer of a boarding mission.

MasterOfGrey:
Right, and I assume "On Landing" condition check precedes the bit that applies mission code. Cool. 👍

Amazinite:
Closing this because although it gets the job done, it makes a mess of things in the save file conditions, so I'd rather we work toward #5017.

Issue #1763, opened 2016-10-23
New mission success condition: capture npc ship


Pointedstick:
I'm writing a mission where you retrieve a merchant captain's ship that's been captured by pirates but I can't find a way to make capturing a ship a part of a mission's success conditions.

Safari-Is-Amazing:
Look at the escort mission where you have to get the freighter from Leslath maybe?

Pointedstick:
That's different; in that mission; you board and repair a disabled ship. It works because the mission's success condition is just to bring the NPC ship back to a chosen planet, and as soon as you board it, it's repaired and starts following you.

Bladewood:
That one involves **repairing** a **derelict merchant** freighter. @Pointedstick is talking about **capturing** a **pirated* ship.

aand fuck ninjas

Johny-Cash:
Why do you hate ninjas? Also I think @Safari-Is-Amazing made a point because possibly the code is similar and could be helpful.

Bladewood:
But we're not coders!

A ninja posted what I wanted to post, seconds before I did!

Johny-Cash:
Or reverse engineer the missions where you have to fight the flamethrower ship (can't remember faction) and if you capture that ship you fail the mission, so go off that maybe?

Bladewood:
> npc save disable
		personality staying heroic
		government "Test Dummy"
		ship "Fury (Flamethrower)" "Doombat"

Pointedstick:
Technically there's already something *close*: the "disable" NPC requirement. However simply disabling a ship doesn't require fighting the pirates, and potentially paying death benefits. That part is important since I'd like it to be a really high-stakes mission, where if you lose crew and have to pay death benefits, that could offset the reward for recapturing the ship, making you think twice about taking the mission or making you ensure that your crew is armed to the teeth first.

Safari-Is-Amazing:
This could be used in the corporate espionage missions. The Syndicate wants you to capture a ship with new technology. As well as being in the future Syndicate storyline where you may have to disable and board the Korath for jump drives.

Elyssaen:
There's a way to do something close, which I'm intending to use in a mission I have in mind. I haven't implemented it yet, though, so I can't guarantee it will work.

1. Create an unplunderable outfit.
2. Set up a variant of your target ship class that carries the outfit, and use that in the NPC definition.
3. Then write:

```
on complete
  outfit "My Outfit" -1
```

Three issues though:

[...]

PR #4952, opened 2020-03-26
Hai secret no more

MasterOfGrey:
Note: This is a DRAFT pull request, much work remains.

## Summary
It struck me on the road from the Free Worlds to the Wanderer story line, as odd, that at no point in the immediate wake of an alien invasion did anyone think to try and capitalise off of the alien hysteria. Sure everyone important who've been to Hai space obviously appreciate the Hai for the peaceful and genteel beings that they are, but many of those people are still profit-motivated traders, with crew who are significantly worse paid.

Obviously the Republic would have major media outlets in its pocket on things which are _secret_, as would the Syndicate. It's plausible that something could slip through the cracks, but unlikely. The Free Worlds on the other hand, have not existed long enough to develop that sort of arrangement, and there's been a big upset generally.

The story goes that a trader's crewman, tired of earning merely pennies, gets to FW space as a crewman and then drops off. He makes his way around until he manages to find a talk show that will listen to his evidence, and consequently gets picked up by a medium sized outlet as a surprise guest. The people who would stop such a thing simply don't know, or are too far away and disconnected to take action before it happens. One Republican outlet realises they can't not report on this stir at all, that would be too suspicious, so they assign one of their experience but less competent journos to the story intending for them to be set up to fail. However, that journo knows someone, who knows someone, who accidentally asks the right person at the wrong time, and takes her content to air without consulting with her supervisors first.

Before long it seems like the secret is 'out', but evidence is still thin on the ground, so maybe it's not completely lost just yet....

-----------------------
## Feature Details
For the moment this just the intro missions seeking feedback. Essentially building up to the first actual decision point.

From here I'm thinking the secret gets out, and a diplomatic station is built in Ultima Thule as both embassy and border control. Alternatively, maybe they try and cover it up by building a station inhabited by a "Dedicated Furry RPG Community". Tbd.

##Feature Intent
The idea is that this leads a player north, encourages them to explore and learn more, and to nudge them in the direction of finding the Wanderers (especially if they've finished the FW without meeting the Hai). In the eventual future of Navy & Syndicate campaigns, its also intended to be a tie-point to link things back together. The other idea, is to set up a foundation of characters, stations, and player investment, for any other future Hai stories (someone has mentioned a Human-Hai war with the Unfettered). It seems unreasonable that in the wake of the Pug, it would be possible for a known alien race to remain "secret" all the way up until a declaration of war, so this is intended to either break the secret first and let everyone calm down again so they can be surprised later, or to at least soften the blow when it comes in the future.

[...]

PR #5414, opened 2020-10-06
Allow capturing an NPC to be an objective

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #1763.

## Feature Details
Creates a new NPC mission objective: capture.
Yes, it really is just these two lines.

## UI Screenshots
N/A

## Usage Examples
See testing done.

## Testing Done
Tested on the following mission.
```
mission "capture test"
	job
	repeat
	npc capture
		government "Bounty"
		personality staying
		ship "Shuttle" "Cap Me"
		dialog "Oh no, you captured me."
	on complete
		dialog "<npc> was successfully captured."
```

## Performance Impact
N/A

### DIFF
--- a/source/NPC.cpp
+++ b/source/NPC.cpp
@@ -62,6 +62,8 @@ void NPC::Load(const DataNode &node)
 			succeedIf |= ShipEvent::SCAN_CARGO;
 		else if(node.Token(i) == "scan outfits")
 			succeedIf |= ShipEvent::SCAN_OUTFITS;
+		else if(node.Token(i) == "capture")
+			succeedIf |= ShipEvent::CAPTURE;
 		else if(node.Token(i) == "evade")
 			mustEvade = true;
 		else if(node.Token(i) == "accompany")


Zitchas:
Very nice!

MasterOfGrey:
I can rework a whole mission to work muuuuch better in #4952 if this is merged. So, watching this. 👍

Zitchas:
Can this be combined with `save` to make a situation where the player has to provoke a target, but will fail if they kill them? I wouldn't want that to be the default behaviour, since in some circumstances whether the target lives or dies doesn't matter; but for some sorts of missions one would want to make sure that the target is still active.

tehhowch (commented):
I think we need a larger change to make `capture` more useful for mission writing, since a captured ship is currently always destroyed first, in the eyes of a mission. Nothing stops the player from capturing and then losing said ship.

One possible way would be through an explicit ShipEvent that maps to `npc kill`, e.g. `ShipEvent::KILL_CREW`, and is emitted when the ship's original "crew" has died (i.e. on explode or capture). The above "captured-is-destroyed" behavior could then be replaced, enabling mixing `save` or `accompany` with `capture`. 
(A complication with this approach occurs if we later allow a ship being boarded to surrender its crew rather than force them to be killed.  
Additionally, this change could easily place us into some murky waters: `kill` would mean "kill the crew" but `save` would mean "don't destroy the ship chassis" and `accompany` would mean "have the ship chassis with you" - how would one say "don't kill the crew or destroy the ship?" Currently, we use `save`.  
Getting the success and failure criteria correct internally, maintaining backwards compatibility, and ensuring mission writers can easily determine which criteria they should use will not be trivial.)

Amazinite:
I've updated it so that captured NPCs aren't counted as destroyed if capturing was the objective. Tested using the following mission:
```
mission "capture test"
	job
	repeat
	npc capture save
		government "Bounty"
		personality staying
		ship "Shuttle" "Cap Then Save Me"
		dialog "Successfully captured. (Save objectives don't count toward this dialog showing, apparently.)"
	npc capture kill
		government "Bounty"
		personality staying
		ship "Shuttle" "Cap Then Kill Me"
		dialog "Successfully captured then killed"
	on complete
		dialog "You did it!"
```
1) Capturing both then killing the correct one completes the mission.
2) Killing either one before capturing fails the mission.
3) Capturing the save one then killing it fails the mission.
But...
4) Capturing the NPCs then landing and reloading the save without having completed the mission respawns the captured NPCs.

Amazinite:
> I think we need a larger change to make capture more useful for mission writing, since a captured ship is currently always destroyed first, in the eyes of a mission. Nothing stops the player from capturing and then losing said ship.

The syntax isn't going to change, so I'd prefer a half-measure now where `capture` can't be mixed with `save` or `accompany` but can still be used to do things that are currently impossible. We can tackle mixing those objectives later.
Can `capture` be made *more* useful than this? Yes. But is this useful enough as is? I'd say yes as well.

Zitchas:
The Remnant are going to *love* having a "capture" mission objective, once we reach a point in the story where the player has access to ships that can reasonably capture current ships without requiring massive casualties.

tehhowch:
> The syntax isn't going to change, so I'd prefer a half-measure now where `capture` can't be mixed with `save` or `accompany` but can still be used to do things that are currently impossible. We can tackle mixing those objectives later.
> Can `capture` be made *more* useful than this? Yes. But is this useful enough as is? I'd say yes as well.

Add an explicit error message for combining capture with `save` or `accompany`, lest we get folks attempting to use them together before it is supported.
We should also report warnings for superfluous combinations as well, such as `kill` or `evade` together or paired with `capture`.

Issue #5395, opened 2020-10-01
Allow location filter to only include visited planets/systems


Amazinite:
**Is your feature request related to a problem? Please describe.**
Location filters will currently pick a random location that matches the filter regardless of whether the player has visited the selected location. This means that location filters can effectively be used to discover systems. While this is fine for the purposes of existing missions, it may be beneficial to have a mission that only requires traveling to locations which the player has already discovered, either so that we can guarantee that the player can reach them or so that we can keep hidden a system that might otherwise match the location filter.

**Related Issue Links**
https://github.com/endless-sky/endless-sky/pull/5374#issuecomment-701694272
> [Y]ou also have to think about how any mission using these attributes for a location filter means that those missions could reveal warzones, black holes, hazards, wormholes, and novae unknown to the player.

**Describe the solution you'd like**
Create a `visited` keyword for location filters that requires that the source/destination/stopover/waypoint system have been visited by the player.

Example: the following destination would be 5 to 10 jumps from the source, but it also has to have been visited.
```
destination
	distance 5 10
	visited
```

This keyword would obvious also be capable of using the `not` modifier (i.e. `not visited` for choosing locations which the player hasn't been to), and the `neighbor` modifier (i.e. `neighbor visited` for choosing locations which are linked to locations which are visited).

**Describe alternatives you've considered**
Create two keywords, `visited planet` and `visited system`, instead of a single `visited` keyword that only checks the system.

**Additional context**
Lol no this isn't me saying merge those attributes after this is done. We still need missions to actually use them.

PR #5415, opened 2020-10-06
Allow location filters to test if a planet/system has been visited

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #5395.

## Feature Details
Creates two new keywords for location filters:
* `visited`: The system must have been visited before.
* `visited planet`: The system and planet must have been visited before.

## UI Screenshots
N/A

## Usage Examples
See testing done.

## Testing Done
Tested the following mission on a fresh pilot after visiting a few systems and worlds. One waypoint and stopover was already from a visited location and the other waypoint and stopover were from unvisited locations.
Checking the jobs board, all other jobs appeared to act normally.
```
mission "visited test"
	job
	repeat
	description "Stopovers: <stopovers>. Waypoints: <waypoints>."
	waypoint
		distance 1 100
		visited
	waypoint
		distance 1 100
		not visited
	stopover
		distance 1 100
		visited planet
	stopover
		distance 1 100
		not visited planet
```

## Performance Impact
N/A

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -478,7 +478,7 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 	// First, figure out the comparative strengths of the present governments.
 	const System *playerSystem = player.GetSystem();
 	map<const Government *, int64_t> strength;
-	UpdateStrengths(strength, playerSystem);
+	UpdateStrengths(player, strength, playerSystem);
 	CacheShipLists();
 	
 	// Update the counts of how long ships have been outside the "invisible fence."
@@ -3624,7 +3624,7 @@ bool AI::Has(const Ship &ship, const Government *government, int type) const
 
 
 
-void AI::UpdateStrengths(map<const Government *, int64_t> &strength, const System *playerSystem)
+void AI::UpdateStrengths(const PlayerInfo &player, map<const Government *, int64_t> &strength, const System *playerSystem)
 {
 	// Tally the strength of a government by the cost of its present and able ships.
 	governmentRosters.clear();
@@ -3664,7 +3664,7 @@ void AI::UpdateStrengths(map<const Government *, int64_t> &strength, const Syste
 	
 		// Check if this ship's government has the authority to enforce scans & fines in this system.
 		if(!scanPermissions.count(gov))
-			scanPermissions.emplace(gov, gov && gov->CanEnforce(playerSystem));
+			scanPermissions.emplace(gov, gov && gov->CanEnforce(player, playerSystem));
 
 		// Only have ships update their strength estimate once per second on average.
 		if(!gov || it->GetSystem() != playerSystem || it->IsDisabled() || Random::Int(60))
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -476,9 +476,8 @@ void AI::ClearOrders()
 void AI::Step(const PlayerInfo &player, Command &activeCommands)
 {
 	// First, figure out the comparative strengths of the present governments.
-	const System *playerSystem = player.GetSystem();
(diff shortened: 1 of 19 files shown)


MasterOfGrey:
I can rework a whole mission to work muuuuch better in #4952 if this is merged. So, watching this. 👍

source/AI.cpp
```diff
@@ -3624,7 +3624,7 @@ bool AI::Has(const Ship &ship, const Government *government, int type) const
 
 
 
-void AI::UpdateStrengths(map<const Government *, int64_t> &strength, const System *playerSystem)
+void AI::UpdateStrengths(const PlayerInfo &player, map<const Government *, int64_t> &strength, const System *playerSystem)
```
petervdmeer, line 3627: Functions that get the full playerInfo class as parameter don't seem to need the playerSystem as a separate parameter (since they can get the playerSystem from playerInfo).

petervdmeer (changes requested):
I'm not too happy with all the additional classes that now need to include `PlayerInfo.h` and all the additional passing of the PlayerInfo object throughout the code, but I also don't really see a better alternative to it.

I do have 1 concrete review-comment that needs some attention.

source/AI.cpp
Amazinite: I've changed this function and Person::Frequency to pass just the player instead of the player and the player's system. All other functions that are changed though need to remain the same, either because they get a planet or system that isn't the player's current planet or system, or there are multiple functions of the same name that are only differentiated by being given a planet or system.

petervdmeer: > because they get a planet or system that isn't the player's current planet or system

This is a good reason I agree with.

> or there are multiple functions of the same name that are only differentiated by being given a planet or system.

This I disagree with; function overloading is not a goal in itself and if we need to give duplicate data in parameters to make it work, then we need to fix this in a different way (for example by not overloading using a single name, but by using different function names, or by adding a new function in `LocationFilter` that has as signature `bool Matches(const PlayerInfo &player, const System *origin = nullptr) const;)` besides the existing Matching functions.

I also wonder now if we really always need to pass PlayerInfo all the way to LocationFilter. some examples:
- If we want to know if some government can enforce fines in some system, then can we ignore the "visited system" and "visited planet" checks, since the check only happens when the player is already present in the relevant system or on the relevant planet?
- If we want to know if some government can scan the player, then isn't the player also already in the system when the check is relevant?

tehhowch (commented):
That we need to pass the player around so much to enable the ~16 lines of real change is a great sign that our classes do too much and our architecture needs refactoring.

One way to avoid passing the player object through all the calls is to move where we store information about visited planets and systems, e.g. to provide `GameData::VisitedSystems()` and `GameData::VisitedPlanets()` getters. I'm not necessarily advocating we do that, though, because in general GameData is a view to the universe without consideration of the player (excepting events that have occurred in the player's timeline).

Another approach is to introduce a service class whose responsibility is to _evaluate_ filters, which is separate from the _filter data_. This evaluating class would be instantiated with a const& to the player, and would have methods that take a given filter and indicate whether the given system, planet, or ship satisfies the constraint. It would also host the `PickPlanet` and `PickSystem` methods. To work with the service class, it's possible we would want classes to publicly expose their filters (which would then also reduce their responsibilities).

Amazinite:
> Another approach is to introduce a service class whose responsibility is to _evaluate_ filters, which is separate from the *filter data*.

Sounds like something that could be done with conditions as well. 
I'm down with the idea.

Amazinite:
Closing as it seems this specific implementation isn’t desired, and I’m not looking to work on an alternative implementation any time soon. Others are free to take their own crack at this issue.

PR #5416, opened 2020-10-07
Make LG's gridfire friendly-safe

ravenshining:
## Summary

This marks LG's person ship's gridfire turrets as "safe" so they don't blow up unsuspecting players.

Gridfire is invisible and deadly to early game ships, and uses damage mechanics not present in human space.  Should Local God appear in the early game, and happen to attack a ship near the player, the player and their escorts may be easily destroyed or disabled.  Unlike, for example, Pointedstick's heavy rocket attacks, the Gridfire cannot be seen or dodged, may come from far off-screen, and may leave the player in the confusing (to new players) state of being disabled with their shields up.

For me, losing a valuable escort and having my flagship disabled out of the blue despite being nowhere near the local furball, it was a surprising annoyance and cause for a scum-save, for a new player I imagine this would be cause for great confusion and frustration with the game.

## Fix Details

A duplicate, pared-down version of the Gridfire turret, named "LG Gridfire Turret," with the "safe" attribute applied, has been added to persons.txt, and Ursa Polaris has been armed with them instead of standard Pug gridfires.

## Save File

This save gives you an Archon equipped with the modified gridfire turret, and 100 void sprites to test it on.

[Local Cod.txt](https://github.com/endless-sky/endless-sky/files/5338056/Local.Cod.txt)

## Testing Done

I armed a ship with the modified gridfires to ensure they worked.

### DIFF
--- a/data/persons.txt
+++ b/data/persons.txt
@@ -760,7 +760,7 @@ ship "Ursa Polaris"
 			"hit force" 6000
 	outfits
 		"Quarg Skylance"
-		"Pug Gridfire Turret" 3
+		"LG Gridfire Turret" 3
 		"Plasma Turret"
 		"Korath Disruptor" 5
 		"Quarg Anti-Missile" 5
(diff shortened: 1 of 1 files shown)


Zitchas (approved):
Yeah, I fully agree with this one. While I don't object to the author's being unpredictable and deadly, it is definitely problematic that a non-hostile author might accidentally destroy the player by mining near by; especially without having a visible effect or being dodgable.

PR #5418, opened 2020-10-07
Fix: Remnant Cognizance 1

### LINKED ISSUE #5309: Wrong text shown in Remnant: Cognizance 1
**Describe the bug**
In Remnant: Cognizance 1 the text that should be only shown when landed on the first of the gas giants is instead shown immediately after entering the system.

**To Reproduce**
1. Accept Remnant: Cognizance 1
2. Go to Nenia
3. Read the message that is shown

**Expected behavior**
This line
https://github.com/endless-sky/endless-sky/blob/390d6a044a7db541082757d95deb56e179569eee/data/remnant/remnant%20missions.txt#L2802
should be shown only once you land on the first gas giant, not when entering the system.

**System:**
 - OS: Manjaro
 - Game Source: nightly
 - Version: 390d6a0

Zitchas:
**Bugfix:** This PR addresses issue #5309 

## Fix Details
This changes the on enter and stopover dialog for Remnant: Cognizance 1.
Fixes #5309 

**Note:** This fix provides `on stopover "Nasqueron"` and `on stopover "Slylandro"`, however due to current engine limitations only one of them (usually Nasqueron's) will trigger, and it triggers on the second planet. As such, both stopover texts are not planet or order specific, and the story will continue just fine if only one of them triggers or if both of them do. I am leaving them both in on the hope that someday soon the engine will support having both. In the meantime, having both doesn't cause any problems.

## Testing Done
Flew the mission.

## Save File
This save file can be used to verify the bugfix.  (note: New test file: Old one had some old data in it causing the korath to be hostile to the void sprites right from Cog 1, which is not supposed to be the case.)
[CogTester07Oct2020.txt](https://github.com/endless-sky/endless-sky/files/5341951/CogTester07Oct2020.txt)


Amazinite:
You need to use separate missions which only offer on the desired planets if the main mission is active and then immediately decline in order to achieve the desired effect. See the Project Hawking or Remnant Surveillance missions in the Deep missions file.

Zitchas:
I don't want to drag multiple missions into this. I'm happier just commenting out the second stopover (which I have done) and leaving it with a single stopover until such time as we have expanded the engine capabilities.

data/remnant/remnant missions.txt
```diff
@@ -2798,12 +2798,14 @@ mission "Remnant: Cognizance 1"
...
 		dialog
-			`As you scan the system, the first thing you notice is that the void sprites' behavior seems different: There are fewer of them flying around, and the few that you see seem to be almost timid - floating quietly, and disappearing into the clouds whenever your ship moves towards them.`
-			`	After several minutes of recording the elusive void sprites, you decide to visit the next gas giant, recalling that Plume will want footage from both planets.`
-	on stopover
-		conversation
-			`Here you find the same behavior as before: void sprites have suddenly become timid and prone to diving into the shrouded depths of their world at the slightest quick movement.`
-			`	Despite the new challenge, you manage to record a variety of different void sprites before deciding to head back to <destination>.`
+			`As you scan the system, the first thing you notice is that the void sprites' behavior seems different: There are fewer of them flying around, and the few that you see seem to be almost timid - floating quietly, and dropping down into the cloud deck whenever your ship moves towards them. You recall that Plume asked you to visit both planets.`
+	on stopover # "Slylandro"
```
W1zrad, line 2802: Just dropping something here since this is the first time it's used.
Do the hashtags belong here? From my own (basic) scripting knowledge, hashtags _usually_ invalidate the things on the line that come after it. And if the intent here is simply to invalidate, what would be the point of adding hashtags instead of entirely removing the lines?
If this is something related to scripting that makes sense and this comment is wrong/irrelevant feel free to ignore. Just wanted to note it.

data/remnant/remnant missions.txt
```diff
@@ -2798,12 +2798,14 @@ mission "Remnant: Cognizance 1"
...
+		conversation
+			`	The void sprites here prove to be elusive. You spend quite a while chasing radar signatures before you finally manage to get into visual range of a small flock of void sprites. You record them from a distance for several minutes, until they suddenly disperse into a storm and vanish.`
```
W1zrad, line 2804:
```suggestion
			`	The void sprites here prove to be elusive. You spend quite a while chasing radar signatures before you finally manage to get a small flock of them into visual range. You record them from a distance for several minutes, until they suddenly disperse into a storm and vanish.`
```
Personal preference. Thought the flow was a little better this way.

data/remnant/remnant missions.txt
Zitchas: I commented them out with the idea that - when support for multiple `on stopover` fields are added - these can be removed and the writing thus re-added to the game.

And yes, the intent is that the lines remain present, but not in use.

W1zrad (approved):

data/remnant/remnant missions.txt
Zitchas: Granted, I'm not sure if the Devs will be OK with doing that, in which case I'll delete all the commented stuff, but that's the idea anyway.

Amazinite (changes requested):
Either go the multiple missions route if you want each planet to have a conversation (we don't have a limited supply of missions, nothing bad will happen if you do), or only change what is necessary (i.e. don't leave commented lines in).

Issue #5420, opened 2020-10-07
Make jobs fail starting from the less paying one


beccabunny:
**Describe the bug**
I had accepted two jobs, one to transport 5 passengers and one to transport 1 tourist to the same system, while having only 5 bunks free. When I departed, the message correctly told me I would have failed a mission due to not having enough free space, but when I still tried to depart I found out the one that failed was the 5 passengers, much higher paying job.

**To Reproduce**
1. Depart
2. See the message about failing a mission due to not having enough space for one passenger
3. Accept
4. See that sometimes the job that fails is the 5 passengers one and some the others

**Expected behavior**
If any jobs would fail with the current capacity, make it so that the less paying one fails first

**Screenshots**
*before departing*
![Screenshot (883)](https://user-images.githubusercontent.com/45330165/95359970-075f4700-08cb-11eb-9765-977f6c3eccf3.png)
*after departing*
![Screenshot (885)](https://user-images.githubusercontent.com/45330165/95359994-134b0900-08cb-11eb-849f-6bee6869484c.png)

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Github
 - Version: Lastest continous build

**Link to save file**
[tommy b.txt](https://github.com/endless-sky/endless-sky/files/5342237/tommy.b.txt)

Amazinite:
Seems to me like job sorting would need to come first.

beccabunny:
Given that it would not cause save files issues (changing syntax or similar) and that sorting is still very far away, it doesn't seem unreasonable to me to have this change sooner rather than later

Amazinite:
How do we remove the least paying job if we can’t sort by pay yet?

Zireael07:
@Amazinite: I think we can sort by number of passengers required?

PR #5426, opened 2020-10-09
Differentiate aborting from failing

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4018.

## Feature Details
Adds a new `on abort` action.
* Only triggers if the mission is aborted by the player.
* Creates a new `"<mission>: aborted"` condition.
* Also creates a `"<mission>: failed"` condition when aborting. (BACKWARDS COMPATIBILITY)
* If a mission is aborted and no `on abort` action exists, any `on fail` action is triggered instead. (BACKWARDS COMPATIBILITY)

## UI Screenshots
N/A

## Usage Examples
See testing done.

Given that aborting a mission creates both the aborted and failed conditions, the following condition sets can be used to test certain outcomes.
Mission failed **OR** aborted:
```
has "<mission>: failed"
```
Mission aborted:
```
has "<mission>: aborted"
```
Mission objective failed:
```
has "<mission>: failed"
not "<mission>: aborted"
```

## Testing Done
Tested the following two missions. For the first mission, aborting properly triggers the on abort action and failing properly triggers the on fail action. For the second mission, since no on abort action exists, aborting properly triggers the on fail action. In both missions, the proper failed and aborted conditions are set given the outcome of the missions. (i.e. both failed and aborted if the mission is aborted, only failed if the mission is failed by objective.)
```
mission "on abort test"
	job
	repeat
	npc save
		government "Hai"
		personality staying
		ship "Shuttle" "If you kill me, I swear."
	on abort
		dialog "You aborted the mission!"
	on fail
		dialog "You failed the objective!"

mission "abort fallback test"
	job
	repeat
	npc save
		government "Hai"
		personality staying
		ship "Shuttle" "If you kill me, I swear."
	on fail
		dialog "You failed the objective or you aborted the mission!"
```

## Performance Impact
N/A

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -811,7 +811,7 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 	// If this mission was aborted but no ABORT action exists, look for a FAIL
 	// action instead. This is done for backwards compatibility purposes from
 	// when aborting a mission activated the FAIL trigger.
-	if(it == actions.end() && trigger == ABORT)
+	if(trigger == ABORT && it == actions.end())
 		it = actions.find(FAIL);
 	// Don't update any conditions if this action exists and can't be completed.
 	if(it != actions.end() && !it->second.CanBeDone(player, boardingShip))
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -244,6 +244,7 @@ void Mission::Load(const DataNode &node)
 				{"accept", ACCEPT},
 				{"decline", DECLINE},
 				{"fail", FAIL},
+				{"abort", ABORT},
 				{"defer", DEFER},
 				{"visit", VISIT},
 				{"stopover", STOPOVER}
@@ -808,7 +809,8 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 	}
 	// Don't update any conditions if this action exists and can't be completed.
 	auto it = actions.find(trigger);
-	if(it != actions.end() && !it->second.CanBeDone(player, boardingShip))
+	bool hasAction = it != actions.end();
+	if(hasAction && !it->second.CanBeDone(player, boardingShip))
 		return false;
 	
 	if(trigger == ACCEPT)
@@ -829,6 +831,20 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 		--player.Conditions()[name + ": active"];
 		++player.Conditions()[name + ": failed"];
 	}
+	else if(trigger == ABORT)
+	{
+		// Only decrement the active mission condition if an ABORT action
+		// exists. Otherwise, this condition will be decremented when the
+		// FAIL trigger is done.
+		if(hasAction)
+		{
+			--player.Conditions()[name + ": active"];
+			// Set the failed mission condition here as well for
+			// backwards compatibility.
+			++player.Conditions()[name + ": failed"];
+		}
+		++player.Conditions()[name + ": aborted"];
+	}
 	else if(trigger == COMPLETE)
 	{
 		--player.Conditions()[name + ": active"];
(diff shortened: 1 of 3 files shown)


Zitchas:
Looks like a very useful addition to the mission writing toolset.

The abort/fail crossover that you have described there also makes sense. For a lot of things, abort = fail, and in the absense of specific abort consequences, more general fail consequences would be appropriate. So long as we can use the tests to differentiate between whether an on abort should trigger or an on fail should trigger, this works for me.

source/Mission.cpp
```diff
@@ -808,7 +809,8 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 	}
 	// Don't update any conditions if this action exists and can't be completed.
 	auto it = actions.find(trigger);
-	if(it != actions.end() && !it->second.CanBeDone(player, boardingShip))
+	bool hasAction = it != actions.end();
+	if(hasAction && !it->second.CanBeDone(player, boardingShip))
```
petervdmeer, line 813: I would prefer to handle the alternative action lookup here, because I'm afraid that you could get double counts for aborted if `return Do(FAIL, player, ui, boardingShip);` on line 870 returns false for some reason when the `++player.Conditions()[name + ": aborted"];` on line 846 was already executed.

(Note that this suggestion is incomplete, the hasAction boolean also needs to be updated after the second find or needs to be removed and reverted to the original code of comparing `it` with `actions.end()`.)

```suggestion
	if(it == actions.end() && trigger == ABORT)
		it = actions.find(FAIL)
	if(it != actions.end() && !it->second.CanBeDone(player, boardingShip))
```

source/Mission.cpp
```diff
@@ -829,6 +831,20 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 		--player.Conditions()[name + ": active"];
 		++player.Conditions()[name + ": failed"];
 	}
+	else if(trigger == ABORT)
+	{
+		// Only decrement the active mission condition if an ABORT action
+		// exists. Otherwise, this condition will be decremented when the
+		// FAIL trigger is done.
+		if(hasAction)
```
petervdmeer, line 839: If you accept my change above, then the update of the player.Conditions has to be done always.

source/Mission.cpp
```diff
@@ -843,10 +859,15 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
...
-	if(it != actions.end())
+	if(hasAction)
 		it->second.Do(player, ui, destination ? destination->GetSystem() : nullptr, boardingShip, IsUnique());
 	else if(trigger == OFFER && location != JOB)
 		player.MissionCallback(Conversation::ACCEPT);
+	// If this mission was aborted but no abort action exists, look for a fail
+	// action instead. This is done for backwards compatibility purposes from
+	// when aborting a mission activated the FAIL trigger.
+	else if(!hasAction && trigger == ABORT)
+		return Do(FAIL, player, ui, boardingShip);
```
petervdmeer, line 870: If you accept my suggestion above, then this code is no longer needed.

petervdmeer (changes requested):
I have one concern about the possibility of counting aborts double. (And created 3 review-comments for it.)

The change itself looks quite okay to me.

source/Mission.cpp
Amazinite: Good catch.

petervdmeer (approved):
Looks good. Sorry for the wrong trigger-name in my earlier suggestion.

Amazinite:
I'm actually the one who screwed up the trigger name. Your suggestion was missing a semicolon, though. :)

tehhowch (commented):
Changes so far are good. We should also update any jobs that should differentiate failure from abort, minimally the syndicate target practice ones.

Are there other situations where `fail` is sent, but `abort` should be instead? On first thought, if we're auto-failing missions on takeoff, then any mission that was accepted today should get the aborted condition instead of failed. (To support that, I believe we will need either a new mission field that is set upon acceptance, or a new data structure in the player class that maps a mission ID + name to the accepted date.)

I'm fine with deferring a mission aborting another via action keyword to another PR.

petervdmeer:
> Changes so far are good. We should also update any jobs that should differentiate failure from abort, minimally the syndicate target practice ones.

I think it is fine to merge the mechanism in this PR and perform the content update in a separate PR. (Just merging the mechanism doesn't make the game worse in any way, it just brings us closer to being able to handle aborts better.)

And to bring theory to practice I also did create this separate PR #5430 to handle the minimal content update.

petervdmeer:
> Are there other situations where `fail` is sent, but `abort` should be instead? On first thought, if we're auto-failing missions on takeoff, then any mission that was accepted today should get the aborted condition instead of failed.

I'm not sure about this, abort is an action by the player. If the player needs to transport 300 people and only has space for 280, then I would expect the player to either arrange for sufficient bunks (leading to success), to manually abort the mission (which is a proper abort) or to automatically fail the mission (by accepting the job and then not arranging the required amount of bunks before take-off).

Amazinite:
Agreed with Peter. Abort should be "the player explicitly went to the mission list and chose to abort this exact mission." When it comes to the "some missions may fail" dialog, the game is choosing which missions to fail (#5420), so I wouldn't consider that aborting.

petervdmeer (approved):

ravenshining:
Two things I've noticed in testing this out under various scenarios:

- If a conversation is set to trigger on abort, it happens immediately after confirming the abort.  I really like this behaviour, although it is different from how a conversation triggered on fail doesn't happen until you land.

- Aborting a mission *after* either failing or completing the objective still produces identical conditions to aborting the mission *before* failing the objective.

PR #5428, opened 2020-10-09
Fix: Remnant: moving the eligibility for Learning Sign to be later.

### LINKED ISSUE #5427: Remnant: Void Sprites 1 does not take into account the player knowing the sign language
**Describe the bug**
In this playthrough I completed the Remnant: sign language mission chain first, including the follow up mission. When I started the Void Sprites missions, I noticed that although <first> <last> can already fluently speak it, the mission treats them as if they don't, specifically in the last part.

**To Reproduce**
1. Finish the sign language mission chain
2. Start the Void Sprite missions

**Expected behavior**
The mission should take that into account and change the dialog accordingly.

**Screenshots**
Screenshot (891)

 - OS: Windows 10
 - Game Source: Github
 - Version: Lastest continous build

Zitchas:
**Bugfix:** This PR addresses issue #5427 

## Fix Details
Moved the requirement for the language from "have met the Remnant" to "they've demonstrated a willingness to work with the Remnant.

fixes #5427

Specifically, this mirrors the requirement for the technology available mission, except uses the Void Sprite 3: done criteria instead of the void sprite research event. This later has a 20 day timer on it. The idea that it will tend to mean that the learning sign mission should happen during this gap period that at least some players find distracting and cause to leave the area.

This also means that the learning sign missions cannot be done before the player has done the basic set of missions, thus resolving #5427

### DIFF
--- a/data/remnant/remnant missions.txt
+++ b/data/remnant/remnant missions.txt
@@ -983,10 +983,16 @@ mission "Remnant: Learn Sign 1"
 	destination "Aventine"
 	repeat 3
 	to offer
-		has "Remnant: Technology Available: done"
+		has "Remnant: Void Sprites 3: done"
+		has "Remnant: Defense 3: done"
 		not "Remnant: Learn Sign 1: done"
 		not "Remnant: Learn Sign 1: active"
 		random < 30
+		or
+			has "Remnant: Key Stones: done"
+			has "Remnant: Key Stones (Hai): done"
+			has "Remnant: Key Stones (Pre-Hai) 2: done"
+			has "Remnant: Broken Jump Drive 2: done"
 	on offer
 		conversation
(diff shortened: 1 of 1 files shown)


Amazinite (changes requested):
These new conditions are effectively synonymous with `has "license: Remnant"`, so I'd simply use that.

Zitchas:
They are very close, but they aren't quite. Swapping the "event void sprite research" for "void sprite 3" moves it from being guaranteed to be after a 20 day delay, to possibly being within that otherwise potentially very empty stretch of time. I think potentially adding something of interest to this time.

In short, I think having it the way I have it listed (not using the event or the license) not just resolves Becca's issue with the sign language, but does so to help fill what otherwise might be an empty gap in the flow of story. Thus a two for one solution.

PR #5429, opened 2020-10-10
Fix: Erroneous Heron image & flares

Zitchas:
**Bugfix:**

## Fix Details
- Updating the Heron image with the correct one.
- Fixing some flares.

Thanks to Anarchist2 for spotting the discrepancy between the regular and x2 image.


Anarchist2:
doesn't that mean you also have the update the hardpoint positions?

Zitchas:
Not really. The Heron doesn't really have designated "here's a round spot to serve as a hardpoint" locations. The only hardpoint that is affected is the rear-most central hardpoint (which has a point defense turret in it), and the image change moves it from being nestled among misc equipment above fore of the engines to being halfway up a shipyard dome; and that's probably a good improvement for it - better field of you for targeting incoming projectiles.

PR #5430, opened 2020-10-10
Add empty "on abort" sections to Syndicate target practice.

petervdmeer:
**Content (Jobs)** Don't fine players for ships they didn't destroy in target practice.

## Summary
This PR uses the feature from #5426 to ensure that aborted Syndicate target practice jobs don't fine the player for destroying the dummy ship by providing an empty on abort section.

## Testing Done
I checked both the fail situation and the abort situation (using the compiled code from #5426). The abort situation doesn't give any output, the fail situation still gives the proper message.

## Save File
I can provide a save file if desired, but the change is quite trivial, so I'm not sure if this is needed.

### DIFF
--- a/data/human/free worlds intro.txt
+++ b/data/human/free worlds intro.txt
@@ -1329,6 +1329,9 @@ mission "FW Escort 1"
 	on accept
 		log "Serving as an escort for a Free Worlds supply convoy to Tarazed. It sounds like pirates have been specifically targeting their convoys, recently."
 	
+	on abort
+		# No action, the player just silently walks away from this mission(chain)
+	
 	on fail
 		set "fw intro escort failed"
 	
@@ -1397,6 +1400,9 @@ mission "FW Escort 1B"
 		conversation
 			`You are met at the spaceport by Tomek Voigt, one of the members of the Free Worlds Council. He thanks you and the other escort captains for your service, and pays you each <payment>. "The convoy is heading out again immediately," he says, "so meet me in the spaceport if you are willing to volunteer for escort duty again."`
 	
+	on abort
+		# No action, the player just silently walks away from this mission(chain)
+	
 	on fail
 		set "fw intro escort failed"
 	
@@ -1459,6 +1465,9 @@ mission "FW Escort 2"
 			
 			`	"Great!" says Tomek. "Good luck to all of you, and I hope to see you all safely back here in a couple of weeks."`
 				accept
+	
+	on abort
+		# No action, the player just silently walks away from this mission(chain)
 	
 	on fail
 		set "fw intro escort failed"
@@ -1526,6 +1535,9 @@ mission "FW Escort 2B"
 		conversation
 			`Again, Tomek Voigt meets you at the spaceport, thanks you all for your courage and dedication, and pays you <payment>. Then, he pulls you aside and says, "Captain <last>, if you're interested in helping us further, please meet me in the spaceport bar."`
 	
+	on abort
+		# No action, the player just silently walks away from this mission(chain)
+	
 	on fail
 		set "fw intro escort failed"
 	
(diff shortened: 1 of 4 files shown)


tehhowch:
Probably we will want an explicit comment added in the node, as is done in situations where an "empty block" is suspicious in some way:
```
on abort
    # No penalty
on fail
    ...
```

(As currently written, the changes could look like a switch case fallthrough to someone not well-versed in reading ES datafiles.)

petervdmeer:
Agreed. Added comment as suggested.

Amazinite:
While we’re changing missions, I believe the FW second chance escort mission from FW Intro should require an explicit objective failure of the associated missions, as right now aborting those missions gets you the second chance, which is written as if you failed the objective.

Amazinite:
Any pirate job that was changed in #4287 also needs an empty `on abort`.

petervdmeer:
> Any pirate job that was changed in #4287 also needs an empty `on abort`.
I used a non-empty `on abort` here since that looked more appropriate. Please review. If you prefer empty, then I will replace those by empty aborts.

Amazinite:
Well if there was a reputation hit on abort then I would imagine there to be one on fail as well. The dialog wouldn't be a part of the on fail though since it's redundant with the scan message.

petervdmeer:
> Well if there was a reputation hit on abort then I would imagine there to be one on fail as well.

The pirates might be more forgiving of somebody that gets caught (which is a risk they understand) than of somebody that just runs away from the mission silently (which looks cowardly).
But I'm okay with also adding the pirate reputation hit in the on fail if that is appropriate.

> The dialog wouldn't be a part of the on fail though since it's redundant with the scan message.

Yes, I noticed.

petervdmeer:
> While we’re changing missions, I believe the FW second chance escort mission from FW Intro should require an explicit objective failure of the associated missions, as right now aborting those missions gets you the second chance, which is written as if you failed the objective.
> 
> FW Escort 1, 1B, 2, 2B, and 3 need empty `on abort`s.

I feel that empty `on abort`s are not really appropriate if you leave freighters to travel alone without defensive escorts through an area crowding with pirates. An abort dialog that says "you abandoned your mission and the freighters got captured without their defensive escorts." might work better to directly connect those missions to second chances.

Btw, those are main storyline content missions, I feel they require more testing than the changes in this PR. Does it make sense to keep those for a separate PR?

Amazinite:
In the case of the Remnant missions, the on fail exists because the missions are treated as completed whether or not the Korath Raiders survive, just with different outcomes based on the player's actions. If the player aborts the mission though, they never fought the Raider, and therefore neither ending makes sense to continue from.
I know that at least in the case of Salvage 4, [I pointed this out to Zitchas](https://github.com/endless-sky/endless-sky/pull/3761#discussion_r245404419), but highlighted how there's nothing we could do about it at the time. Now is the time to fix that, though.

Salvage 5 needs `not "Remnant: Salvage 4: aborted"` and Cognizance 5 needs `not "Remnant: Cognizance 4: aborted"` in their to offers should we give the previous missions on aborts.

In the case of the FW missions, if the player is purposefully aborting the missions, then I think we should take that as them completely quitting instead of dragging them back into the second chance as if they had simply failed the objective.

petervdmeer:
> In the case of the FW missions, if the player is purposefully aborting the missions, then I think we should take that as them completely quitting instead of dragging them back into the second chance as if they had simply failed the objective.

Yes, that makes a lot of sense. I added the empty 'on abort's for the FW missions to this PR.

petervdmeer:
> Any pirate job that was changed in #4287 also needs an empty `on abort`.
> Remnant: Salvage 4 and Cognizance 4 as well.



> Salvage 5 needs `not "Remnant: Salvage 4: aborted"` and Cognizance 5 needs `not "Remnant: Cognizance 4: aborted"` in their to offers should we give the previous missions on aborts.

Yes, sounds plausible. Added this as well.

Amazinite (approved):
This now covers all the instances that I can recall needing updated for when abort and fail can be differentiated.

Issue #4078, opened 2018-12-17
Korath Worldship swears at me in the mission Remnant: Defense 2


AlbertNewton:
**Describe the bug**
During the mission Remnant: Defense 2 (spelling mistake intentional as that is what the mission is called - another issue here?) the ship can send hails reminiscent of 'Pirate' ships when disabled to the main screen. I have an example here(look at the bottom)
![image](https://user-images.githubusercontent.com/42004388/50111341-7a9a5380-0234-11e9-9741-909ab88fc699.png)

**To Reproduce**
1. Take the mission in a pilot
2. Disable the ship
3. Wait a little while
4. You should see insults thrown at you from the Worldship

**Expected behavior**
I expected to not receive English hails from a creature speaking an alien language.

**System (please complete the following information):**
 - OS: Windows 10 Home. I know no more except that I always do all updates.
 - Game Source: nightly, last updated yesterday
 - Version: 16 December's nightly

Zitchas:
That is problematic. I'll test that later today.

Zitchas:
Sorry, took me longer than expected to test this. Along the way I noticed that disabled Remnant ships also have the same problem.

In regards to the spelling of Defense, yeah, Fzzr nailed it: Due to the fact that the founding/primary developer for ES lives and works in the USA, he used American English spelling for everything, so that's the standard for ES.

tehhowch:
For a government that does not define friendly or hostile disabled hails, they have default values:
https://github.com/endless-sky/endless-sky/blob/46713549172774bf5777f0319678365eed93b2c4/source/Government.cpp#L124-L128

While the Korath have a defined `language`, language checks are pretty sparse in the game currently, and one is not present in the standard `Engine::SendHails` logic:
https://github.com/endless-sky/endless-sky/blob/46713549172774bf5777f0319678365eed93b2c4/source/Engine.cpp#L1537-L1541

but one is present in the `Engine::DoGrudge` logic:
https://github.com/endless-sky/endless-sky/blob/46713549172774bf5777f0319678365eed93b2c4/source/Engine.cpp#L2023-L2030

(The only other use of a `Government::Language` check is performed when you open the hail panel with a ship or planet)

I am not quite sure why the disabled ship is able to send the hail at all, to be honest, since the `CanSendHail` function explicitly states that disabled ships cannot send a hail. I'd need to walk through the call stack to see exactly how / when these messages are popping up.

tehhowch:
[...]

PR #5431, opened 2020-10-10
Make the language check a part of CanSendHail

Amazinite:
**Bugfix:** This PR might address issue #4078.

## Fix Details
Adds the check to see if the hailing ship shares a language with the player to the CanSendHail function. This should prevent ships from hailing the player when a language is not shared under any circumstance, as CanSendHail is always checked before sending a hail message.

Still not sure how the World Ship in #4078 sent a disabled hail, though, as disabled ships do not send random hails, and the only way to view disabled hails is to hail a disabled ship, as best I can tell.

## Testing Done
Flew around Ka'het space with and without this change. Without this change, Ka'het ships would occasionally hail me, but hailing them returned the expected alien language response. With this change, I was never hailed by a Ka'het ship. The same should extend to Korath ships.

## Save File
N/A

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -132,7 +132,8 @@ namespace {
 			return false;
 		
 		// Player ships shouldn't send hails.
-		if(!ship->GetGovernment() || ship->IsYours())
+		const Government *gov = ship->GetGovernment();
+		if(!gov || ship->IsYours())
 			return false;
 		
 		// Make sure this ship is able to send a hail.
@@ -141,7 +142,6 @@ namespace {
 			return false;
 		
 		// Ships that don't share a language with the player shouldn't send hails.
-		const Government *gov = ship->GetGovernment();
 		if(!gov->Language().empty() && !player.GetCondition("language: " + gov->Language()))
 			return false;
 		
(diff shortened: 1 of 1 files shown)


source/Engine.cpp
```diff
@@ -139,6 +140,11 @@ namespace {
 				|| ship->Cloaking() >= 1. || ship->GetPersonality().IsMute())
 			return false;
 		
+		// Ships that don't share a language with the player shouldn't send hails.
+		const Government *gov = ship->GetGovernment();
```
petervdmeer, line 144: Is it possible that a ship has no government (i.e. you getting a null-pointer here)?

petervdmeer (approved):
One remark (not sure if it is relevant though).
Change looks good otherwise.

source/Engine.cpp
Amazinite: A nullptr check is done further up in this function. (Moving this line up above that one as I hadn't noticed it earlier.)

Sandwichs-del:
I think it would be better to clearly distinguish the technological and language barrier.
I should receive nothing if it is due to a technical barrier, but I should receive "Ship XXX tries to hail you, but you do not understand what it says" when it is due to a language barrier.

Since I can hail them I see no reason why they shouldn't be able to hail me.
Would you accept a patch that overrides the message for such cases?

PR #5432, opened 2020-10-10
Added fine 0 and bribe 0 for Governments where those don't make sense

petervdmeer:
**Content (Governments)** Disabled fines for Governments that shouldn't fine

## Summary
Added `fine 0` and `bribe 0` for some governments (like Indigenous Lifeform) where it makes no sense if those governments give fines.

## Save File
N/A

## Follow Up
I would expect most future governments not to levy fines and to accept bribes, so we might want to do a change in the future to change the defaults for fine and bribe to 0 and only explicitly set them for governments that do levy fines and that do accept bribes.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -14,6 +14,7 @@ government "Alpha"
 	"crew defense" 2.5
 	"player reputation" -1000
 	"bribe" 0
+	"fine" 0
 
 government "Author"
 	"player reputation" 1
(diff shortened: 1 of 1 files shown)


Zitchas:
Not accepting bribes seems like a good default to have; but on the other hand levying fines seems a very civilized sort of thing that governments would do. Much more civilized than the alternatives of ignoring crime or blasting people out of the sky.

That being said, the indigenous having both fine and bribe 0 makes sense.

Zitchas (approved):
Changes look good.

Issue #5439, opened 2020-10-12
Conditional conversation nodes


SolraBizna:
**Is your feature request related to a problem? Please describe.**

I'm working on a plugin with a very non-linear, highly interdependent mission tree, to the point that it's more like a mission web. Very often, I'll have a conversation node where I want to offer six or more choices to the player, except that many of the choices require and/or are exclusive with things that may or may not have happened in previous missions. The usual solution for conditional `choice` nodes in Endless Sky is ... inadequate at this level.

**Related Issue Links**

(A quick search didn't find anybody else requesting this.)

**Describe the solution you'd like**

Synthesizing a very silly example:

```sh
conversation
    `"Hello!" the man sitting next to you at the bar says. "You're Captain <last>, right?"`
    branch "don't recognize him"
        not "got told about the renegade deputy"
    `    You recognize his face from the wanted bulletin that Captain Snafu sent you. This is the deputy that let the criminals who stole the Boundless Frontiers source code escape.`
        to display
            not "learned the truth about Boundless Frontiers LLC"
    `    You recognize his face from the wanted bulletin that Captain Snafu sent you. This is the deputy that helped those Remnant scientists steal the Boundless Frontiers source code so they could work on a cure.`
        to display
            has "learned the truth about Boundless Frontiers LLC"
    label "don't recognize him"
    `    You have to be careful what you say. Potato Pete warned you that this bar has listening devices under the stools.`
        to display
            has "know about the listening device"
    choice
        `    "Uh... hello."`
            goto "hello" # (clearer than just falling through)
        `    (Ignore him.)`
            defer
        `    (Try miming to him that there are listening devices under the stools.)`
            to display
                has "know about the listening device"
            goto "get arrested for public miming"
        `    (Try answering him in Remnant sign language.)`
            to display
                has "event: remnant: sign studies complete"
                or
                    has "know about the listening device"
                    has "learned the truth about Boundless Frontiers LLC"
            goto "get away with helping the deputy"
        `    (Shoot him.)`
            to display
                has "got told about the renegade deputy"
            goto "maybe shoot the deputy" # ...but NOT the sheriff?
        `    "You're under arrest!"`
            to display
                has "got told about the renegade deputy"
            goto "deputy runs away"
        `    "Hey, keep your voice down. You're a wanted man."`
            to display
                has "got told about the renegade deputy"
            goto "get caught helping the deputy"
```

[...]

PR #5440, opened 2020-10-12
Feature: Add support for conditional hiding of text or choices in conversations

SolraBizna:
**Feature:** This PR implements the feature request detailed and discussed in issue #5439

## Feature Details

This PR allows a `condition` child to be applied to text in `conversation` and `choice` nodes. If this child is present, it contains a set of condition test expressions, all of which must evaluate to true in order for the paragraph / choice to be displayed and/or selectable.

## UI Screenshots

N/A (the user experience is exactly the same as before, except that behind the scenes there's another way to hide text or choices from the player)

## Usage Examples

Copy-pasting the very silly example from #5439 here:

```sh
conversation
    `"Hello!" the man sitting next to you at the bar says. "You're Captain <last>, right?"`
    branch "don't recognize him"
        not "got told about the renegade deputy"
    `    You recognize his face from the wanted bulletin that Captain Snafu sent you. This is the deputy that let the criminals who stole the Boundless Frontiers source code escape.`
        to display
            not "learned the truth about Boundless Frontiers LLC"
    `    You recognize his face from the wanted bulletin that Captain Snafu sent you. This is the deputy that helped those Remnant scientists steal the Boundless Frontiers source code so they could work on a cure.`
        to display
            has "learned the truth about Boundless Frontiers LLC"
    label "don't recognize him"
    `    You have to be careful what you say. Potato Pete warned you that this bar has listening devices under the stools.`
        to display
            has "know about the listening device"
    choice
        `    "Uh... hello."`
            goto "hello" # (clearer than just falling through)
        `    (Ignore him.)`
            defer
        `    (Try miming to him that there are listening devices under the stools.)`
            to display
                has "know about the listening device"
            goto "get arrested for public miming"
        `    (Try answering him in Remnant sign language.)`
            to display
                has "event: remnant: sign studies complete"
                or
                    has "know about the listening device"
                    has "learned the truth about Boundless Frontiers LLC"
            goto "get away with helping the deputy"
        `    (Shoot him.)`
            to display
                has "got told about the renegade deputy"
            goto "maybe shoot the deputy" # ...but NOT the sheriff?
        `    "You're under arrest!"`
            to display
                has "got told about the renegade deputy"
            goto "deputy runs away"
        `    "Hey, keep your voice down. You're a wanted man."`
            to display
                has "got told about the renegade deputy"
            goto "get caught helping the deputy"
```

## Testing Done

I made a simple test mission in the vein of the above, and attempted to hit all the edge cases regarding paragraphs before or after paragraphs with endpoints, conditions, applies, gotos, and the same with choices. 

## Performance Impact

Conversations use a little more memory now (one `ConditionSet` instance per possibly-merged paragraph). There should never be any noticeable slowdown; the vast majority of nodes will have empty `ConditionSet`s that trivially pass through all text.

Edit 2022-04-07: altered syntax in example, now has `to show` instead of `condition`
Edit 2022-11-17: altered syntax again, now has `to display` instead of `to show` :)


MasterOfGrey:
Ok this looks really useful.

SolraBizna:
I've made a few passes through the vanilla content, and noticed a few places where this feature would make the implementation simpler, eliminate redundant choices, or make it harder to miss content by asking the wrong question first. Should I be adding those content changes to this PR, or making a separate, dependent one, or...?

Amazinite:
Let's do those in a separate PR, but have them at the ready for when this is finalized.

Galaucus:
This would be absolutely fantastic for the Spaceport Culture project, among other things - having it serve as a way to just have casual conversations at spaceport hubs would add so much flavor, or open up roleplaying possibilities as you build up reliable contacts you can return to for information.

SolraBizna:
Shortly after my last comment, I suddenly found most of my previously free time being taken up by an actual paying job. (Some of the fruits of this are on GitHub, actually...) This is great, except I haven't had much time and energy to actually go through Endless Sky again and work out specifics of how the existing vanilla content can be streamlined by this feature. Is it important that I do that? Will that affect whether (and how soon) the feature gets included?

And, apart from keeping an eye on this PR and periodically rebasing it & resolving conflicts (like... the one it has now, uh oh), is there anything else expected of me, as the submitter, that I should be keeping up on?

Amazinite:
>Is it important that I do that? Will that affect whether (and how soon) the feature gets included?

Not at all. We typically keep the addition of a mechanic and its implementation separate anyways for ease of reviewing.

> And, apart from keeping an eye on this PR and periodically rebasing it & resolving conflicts (like... the one it has now, uh oh), is there anything else expected of me, as the submitter, that I should be keeping up on?

As of right now, no. At some point we'll eventually get around to reviewing this (it's definitely a feature that I would like to have and already have a few spots in mind where it would really help), and when that happens, should there be any changes that need made, that's when we'll need more of your attention. When that point is though, couldn't tell you. \:)

MasterOfGrey:
Hi there, this presently says it has a failed check.

Could you please resolve that and update it to confirm compatibility with Master? I'd like to push for this being included sooner rather than later.

SolraBizna:
I looked at that failure when it happened, and it appeared to have nothing to do with any code I touched. I'm not sure how to rerun it to confirm that theory.

MCOfficer:
> I looked at that failure when it happened, and it appeared to have nothing to do with any code I touched. I'm not sure how to rerun it to confirm that theory.

Can't really tell since the logs expired. There might be a button in the CI interface somewhere, but the most reliable way is to push a new ([empty](https://stackoverflow.com/a/61789697/7653274)) commit.

SolraBizna:
All checks are passing after an empty commit, and no new conflicts. (Thanks for the link, without it I wouldn't have known how to make an empty commit.)

tehhowch:
Much better than an empty commit is _merging the target branch in_, or if you're familiar, rebasing. Otherwise your branch can be extremely outdated and not include checks or tests it might fail.

SolraBizna:
I assumed that these CI checks are running on a merged version of my PR, not on my raw branch. Is this not the case? (It seems like more than a few people would be unhappily surprised if, when they merge a PR that's passing CI checks, they find that CI fails horribly post-merge.)

MCOfficer:
> 
> 
> I assumed that these CI checks are running on a merged version of my PR, not on my raw branch. Is this not the case? (It seems like more than a few people would be unhappily surprised if, when they merge a PR that's passing CI checks, they find that CI fails horribly post-merge.)

They run on your branch, it's on you to keep it at least somewhat uptodate.

SolraBizna:
> They run on your branch, it's on you to keep it at least somewhat uptodate.

Well, color me surprised. I'll try to find time tonight to merge my branch to the latest master just to be safe, then.

I'm confused as to why the CI went from failing to passing with no actual changes in my branch, if that's the case. Is that because the CI configuration changed?

quyykk:
> I'm confused as to why the CI went from failing to passing with no actual changes in my branch, if that's the case. Is that because the CI configuration changed?

From personal experience sometimes CI fails because some job fails to download a file for some reason. That was probably the case.

SolraBizna:
Rebased without incident, checks still pass. 🎊

source/Conversation.cpp
```diff
@@ -127,8 +127,9 @@ void Conversation::Load(const DataNode &node)
 				
 				// Store the text of this choice. By default, the choice will
 				// just bring you to the next node in the script.
-				nodes.back().data.emplace_back(grand.Token(0), nodes.size());
-				nodes.back().data.back().first += '\n';
+				nodes.back().data.emplace_back(grand.Token(0), nodes.size(),
+											   ConditionSet());
```
Amazinite, line 131:
```suggestion
				nodes.back().data.emplace_back(grand.Token(0), nodes.size(), ConditionSet());
```

source/Conversation.cpp
```diff
@@ -155,14 +156,15 @@ void Conversation::Load(const DataNode &node)
 			for(int i = 1; i <= 2; ++i)
 			{
 				// If no link is provided, just go to the next node.
-				nodes.back().data.emplace_back("", nodes.size());
+				nodes.back().data.emplace_back("", nodes.size(),
+											   ConditionSet());
```
Amazinite, line 160:
```suggestion
				nodes.back().data.emplace_back("", nodes.size(), ConditionSet());
```

source/Conversation.cpp
```diff
@@ -180,14 +182,16 @@ void Conversation::Load(const DataNode &node)
...
 			// This is just an ordinary text node.
 			// If the previous node is a choice, or if the previous node ended
-			// in a goto, create a new node. Otherwise, just merge this new
-			// paragraph into the previous node.
-			if(nodes.empty() || !nodes.back().canMergeOnto)
+			// in a goto, or if the new node has a condition, then create a new
+			// node. Otherwise, just merge this new paragraph into the previous
+			// node.
+			if(nodes.empty() || !nodes.back().canMergeOnto
+					|| HasCondition(child))
```
Amazinite, line 189:
```suggestion
			if(nodes.empty() || !nodes.back().canMergeOnto || HasCondition(child))
```

source/Conversation.cpp
```diff
@@ -402,40 +414,84 @@ int Conversation::NextNode(int node, int choice) const
...
 	
-	return nodes[node].data[choice].second;
+	return nodes[node].data[choice].next;
+}
+
+
+
+// Return whether the text has failed conditions and should be skipped
+bool Conversation::ShouldSkipText(const map<string, int64_t> &vars,
+								  int node, int choice) const
```
Amazinite, line 424:
```suggestion
bool Conversation::ShouldSkipText(const map<string, int64_t> &vars, int node, int choice) const
```

source/Conversation.cpp
```diff
@@ -402,40 +414,84 @@ int Conversation::NextNode(int node, int choice) const
...
+{
+	if(static_cast<unsigned>(node) >= nodes.size()
+			|| static_cast<unsigned>(choice) >= nodes[node].data.size())
```
Amazinite, line 427:
```suggestion
	if(static_cast<unsigned>(node) >= nodes.size() || static_cast<unsigned>(choice) >= nodes[node].data.size())
```

source/Conversation.cpp
```diff
@@ -402,40 +414,84 @@ int Conversation::NextNode(int node, int choice) const
...
+		return false;
+	const auto& data = nodes[node].data[choice];
```
Amazinite, line 429:
```suggestion
	const auto &data = nodes[node].data[choice];
```

source/Conversation.cpp
```diff
@@ -402,40 +414,84 @@ int Conversation::NextNode(int node, int choice) const
...
+	else
+		return !data.conditions.Test(vars);
 }
 
 
 
-// Parse the children of the given node to see if then contain any "gotos."
-// If so, link them up properly. Return true if gotos were found.
+// Parse the children of the given node to see if then contain any "gotos," or
+// "conditions."  If so, link them up properly. Return true if gotos or
```
Amazinite, line 439:
```suggestion
// "conditions." If so, link them up properly. Return true if gotos or
```

source/Conversation.cpp
```diff
@@ -402,40 +414,84 @@ int Conversation::NextNode(int node, int choice) const
...
 			child.PrintTrace("Ignoring extra text in conversation choice:");
 		else if(child.Size() == 2 && child.Token(0) == "goto")
 		{
 			// Each choice can only have one goto
-			Goto(child.Token(1), nodes.size() - 1, nodes.back().data.size() - 1);
-			hasGoto = true;
+			if(hasGoto)
+			{
+				child.PrintTrace("Ignoring extra endpoint in conversation choice:");
+			}
```
Amazinite, line 454:
```suggestion
				child.PrintTrace("Warning: Ignoring extra endpoint in conversation choice:");
```

source/Conversation.cpp
```diff
@@ -402,40 +414,84 @@ int Conversation::NextNode(int node, int choice) const
...
+				hasGoto = true;
+			}
+		}
+		else if(child.Size() == 1 && child.Token(0) == "condition")
+		{
+			// Each choice can only have one condition
+			if(hasCondition)
+			{
+				child.PrintTrace("Ignoring extra condition in conversation choice:");
+			}
```
Amazinite, line 467:
```suggestion
				child.PrintTrace("Warning: Ignoring extra condition in conversation choice:");
```

source/Conversation.cpp
```diff
@@ -402,40 +414,84 @@ int Conversation::NextNode(int node, int choice) const
...
 			// Check if this is a recognized endpoint name.
 			int index = TokenIndex(child.Token(0));
 			if(child.Size() == 1 && index < 0)
 			{
-				nodes.back().data.back().second = index;
-				hasGoto = true;
+				if(hasGoto)
+				{
+					child.PrintTrace("Ignoring extra endpoint in conversation choice:");
+				}
```
Amazinite, line 483:
```suggestion
					child.PrintTrace("Warning: Ignoring extra endpoint in conversation choice:");
```

source/Conversation.cpp
```diff
@@ -483,5 +539,17 @@ void Conversation::Goto(const string &label, int node, int choice)
...
+
+
+
+bool Conversation::HasCondition(const DataNode &node)
+{
+	for(const DataNode &child : node)
+	{
+		if(child.Size() == 1 && child.Token(0) == "condition")
+			return true;
+	}
```
Amazinite, line 553:
```suggestion
		if(child.Size() == 1 && child.Token(0) == "condition")
			return true;

```

source/Conversation.h
```diff
@@ -24,14 +24,20 @@ class DataNode;
 class DataWriter;
 class Sprite;
```
Amazinite, line 26:
```suggestion



```

source/Conversation.h
```diff
@@ -24,14 +24,20 @@ class DataNode;
...
 // accept or decline a mission. The conversation can take different paths depending
 // on what responses you choose, leading you to accept, decline, or (rarely) to be
 // killed. A conversation can also branch based on various condition flags that
 // are set for the player, and can also modify those flags.
 class Conversation {
+	class Data {
+	public:
+		std::string text; // the text to display
+		int next; // the next node to visit
+		ConditionSet conditions; // conditions for displaying the text
```
Amazinite, line 37:
```suggestion
		// The text to display:
		std::string text; 
		// The next node to visit:
		int next;
		// Conditions for displaying the text:
		ConditionSet conditions; 
```

source/Conversation.h
```diff
@@ -76,8 +82,8 @@ class Conversation {
 	const std::string &Text(int node, int choice = 0) const;
 	const Sprite *Scene(int node) const;
 	int NextNode(int node, int choice = 0) const;
-	
-	
+	bool ShouldSkipText(const std::map<std::string, int64_t> &vars,
+						int node, int choice = 0) const;
```
Amazinite, line 86:
```suggestion
	bool ShouldSkipText(const std::map<std::string, int64_t> &vars, int node, int choice = 0) const;


```

source/ConversationPanel.cpp
```diff
@@ -176,20 +176,22 @@ void ConversationPanel::Draw()
 	{
 		string label = "0:";
 		int index = 0;
-		for(const Paragraph &it : choices)
+		for(const auto &it : choices)
 		{
 			++label[0];
+			
+			const auto& paragraph = it.first;
```
Amazinite, line 183:
```suggestion
			const auto &paragraph = it.first;
```

source/ConversationPanel.cpp
```diff
@@ -322,40 +324,68 @@ void ConversationPanel::Goto(int index, int selectedChoice)
...
+				// This is an ordinary conversation node. Perform any necessary text
+				// replacement, then add the text to the display.
+				if(!conversation.ShouldSkipText(player.Conditions(), node))
+				{
+					string altered = Format::Replace(conversation.Text(node), subs);
+					text.emplace_back(altered, conversation.Scene(node), text.empty());
+				}
+				else
+				{
+					// if the text was skipped, don't follow its goto
```
Amazinite, line 360:
```suggestion
					// If the text was skipped, don't follow its goto.
```

Amazinite (changes requested):
Finally getting around to setting my sights on this one. We'll be putting this one to use as soon as it's merged. Sorry about the wait.

This review is just looking over the code style. I'll get to the substance of it later. The only other thing I'll mention at this point is that when dealing with condition sets, the keyword in the data is typically `to *` where `*` is some word. For example, `to offer` or `to complete` for missions, or `to spawn` or `to despawn` for NPCs, or `to show` for news. So instead of just `conditions`, it'd probably be better to have a `to *` keyword.

SolraBizna:
I'm too sick to get to it myself today, but a cursory glance suggests that all it needs to get working again are a few replacements of `first` with `text` and `second` with `next`.

Amazinite:
All the merge conflicts are probably whitespace related. We use to have trailing whitespace on empty lines that matched the indentation of the non-empty line above it, but that's no longer the case.

source/ConversationPanel.cpp
```diff
@@ -454,3 +484,13 @@ Point ConversationPanel::Paragraph::Draw(Point point, const Color &color) const
...
+
+
+
+int ConversationPanel::MapChoice(int n) const
+{
+	if(n < 0 || n >= static_cast<int>(choices.size()))
+		return 0;
+	else
+		return next(choices.cbegin(), n)->second;
+}
```
tehhowch, line 496: If `Conversation::NextNode` should _always_ be called with the output of `MapChoice`, make `MapChoice` return a custom class that wraps an int (aka a "strong type").

That will help prevent new calls from not integrating with the conditional display system.

source/Conversation.h
```diff
@@ -32,6 +32,17 @@ class Sprite;
...
+	class Data {
+	public:
+		// The text to display:
+		std::string text; 
+		// The next node to visit:
+		int next;
+		// Conditions for displaying the text:
+		ConditionSet conditions; 
+		explicit Data(std::string text, int next, ConditionSet conditions)
+			: text(text), next(next), conditions(conditions) {}
```
tehhowch, line 44: If all calls to the explicit `Conversation::Data` constructor are passing in an empty, default-constructed `ConditionSet`, then you don't need that parameter in the explicit constructor.


```suggestion
		explicit Data(std::string text, int next)
			: text(std::move(text)), next(next) {}
```
It's also common for our codebase to place constructors before fields, even when the fields are public.

source/Conversation.cpp
```diff
@@ -402,40 +411,76 @@ int Conversation::NextNode(int node, int choice) const
...
-	return nodes[node].data[choice].second;
+	return nodes[node].data[choice].next;
+}
+
+
+
+// Return whether the text has failed conditions and should be skipped
+bool Conversation::ShouldSkipText(const map<string, int64_t> &vars, int node, int choice) const
+{
+	if(static_cast<unsigned>(node) >= nodes.size() || static_cast<unsigned>(choice) >= nodes[node].data.size())
```
tehhowch, line 422: This bounds check should probably be pulled out into a private method

tehhowch (commented):
Is there a good reason to only make the text nodes conditionally displayable?

SolraBizna:
> All the merge conflicts are probably whitespace related. We use to have trailing whitespace on empty lines that matched the indentation of the non-empty line above it, but that's no longer the case. That change caused a merge conflict for basically every open source code change, though.

This would explain all the places where git merge gives up on what should be a trivial patch, followed (or preceded) by an "empty" line.

> Is there a good reason to only make the text nodes conditionally displayable?

I don't think I understand this question. Text nodes and choices are both conditionally displayable, or I've made a mistake.

source/ConversationPanel.cpp
SolraBizna: Good idea, though now I'll have to remember how to make a "newtype" wrapper in C++. (Rust has really spoiled me.)

`MappedChoice` seem like an appropriate name? Or, maybe, `ChoiceResult`?

source/Conversation.h
SolraBizna: > If all calls to the explicit `Conversation::Data` constructor are passing in an empty, default-constructed `ConditionSet`, then you don't need that parameter in the explicit constructor.

This change makes me feel weird, since I really wanted this to just be a 3-tuple, but... you raise a very good point.

> It's also common for our codebase to place constructors before fields, even when the fields are public.

I'll move that before I do anything else (so I don't forget).

source/ConversationPanel.cpp
SolraBizna: On deeper inspection, I like it better as a bare integer. The index given to `Conversation::NextNode` is among all choices present in the data, while the one given to `ConversationPanel::MapChoice` is only among choices *currently* visible in the panel. Other consumers of `NextNode` (of which there are currently only one) don't necessarily want to work with it in a post-conditioned state, and if they do, what shape that state takes should be up to them.

SolraBizna:
I'm most of the way through a rebase but I've hit a snag. I think c07485b80f39fab054c83e8f3fac2f038a5021c8 subtly changed the way the conversation code works, and more effort than a simple rebase is required. I'll keep working at it.

tehhowch:
> > Is there a good reason to only make the text nodes conditionally displayable?
> 
> I don't think I understand this question. Text nodes and choices are both conditionally displayable, or I've made a mistake.

I'm referring to the other kinds of "root nodes", such as scenes.

(I agree that making a `branch` conditional on other conditions is nonsense, and "directives" such as `accept` or `apply`/`action` should follow the conditions of the node they adjoin with.)

Amazinite:
I don't see conditional showing of scenes as necessary as conditional hiding of text. Scenes aren't as prevalent as text or choices, and it isn't too much of a hassle to skip a scene with a branch should you want to; building the condition set into the `scene` node wouldn't be that much of a boon, but building them into choices MASSIVELY cuts down on the work needed to create choices that you return to without them allowing for a loop.

SolraBizna:
It's rebased, and it builds, though it remains to be seen whether it works. Pushing what I have, and embarking on a test playthrough and some test mission creation. (I've long since lost my original test missions.)

SolraBizna:
That last commit is some defensiveness I added in when I thought I had a bug in my code, but it turned out just to be a bug in my test mission. The basic functionality check has now passed. I'm now proceeding with the "play the game for a while and see if any missions act strange" part of the testing. :)

source/Conversation.cpp
SolraBizna: There are lots of bounds checks that are duplicated inline like that. Should I fix all of them?

SolraBizna:
I haven't had as much time to test as I'd like, but I've gotten about halfway through the Remnant string and done some miscellaneous human missions without any weird mission behavior. Except... I managed to acquire a mission with empty title and description at some point. I thought it was the Raid Warning mission sticking around, so I didn't keep or analyze a save, I just aborted the mission and made a mental note to come back to that. When I went back later and forced Raid Warning to re-trigger, there was no ghost mission. I'm not sure what it was.

Apart from currently unresolved conversations, is there anything else needed from me?

source/ConversationPanel.h
```diff
@@ -110,9 +110,10 @@ template <class T>
 
 	// The "history" of the conversation up to this point:
 	std::list<Paragraph> text;
-	// The current choices being presented to you:
-	std::list<Paragraph> choices;
+	// The current choices being presented to you, and their indices:
+	std::list<std::pair<Paragraph, int>> choices;
 	int choice;
+	int MapChoice(int n) const;
```
Amazinite, line 116: Style fix: This line needs moved. This is in the private fields section of the header. The private methods section is further up the file at line 59. Could also use a comment that says what exactly this method does.

source/ConversationPanel.cpp
```diff
@@ -470,3 +504,13 @@ Point ConversationPanel::Paragraph::Draw(Point point, const Color &color) const
 	point.Y() += wrap.Height();
 	return point;
 }
+
+
+
+int ConversationPanel::MapChoice(int n) const
```
Amazinite, line 510: Style fix: Functions in the .cpp file should match their order in the header file. So this should go below ConversationPanel::ClickChoice (assuming that's also where you move it in the header file). The comment I mentioned that should be added in the header should also appear here.

source/ConversationPanel.cpp
```diff
@@ -340,38 +342,70 @@ void ConversationPanel::Goto(int index, int selectedChoice)
...
+			if(!conversation.NodeInRange(node))
+				node = Conversation::DECLINE;
 		}
 		else
 		{
-			// This is an ordinary conversation node. Perform any necessary text
-			// replacement, then add the text to the display.
-			string altered = Format::Replace(conversation.Text(node), subs);
-			text.emplace_back(altered, conversation.Scene(node), text.empty());
+			break;
```
Amazinite, line 407: Style fix: This `else` is just a single line now, so the curly braces can go.

source/Conversation.h
```diff
@@ -75,11 +86,15 @@ class Conversation {
...
-	const ConditionSet &Branch(int node) const;
+	const ConditionSet &Conditions(int node) const;
 	const GameAction &GetAction(int node) const;
 	const std::string &Text(int node, int choice = 0) const;
 	const Sprite *Scene(int node) const;
 	int NextNode(int node, int choice = 0) const;
+	bool ShouldSkipText(const std::map<std::string, int64_t> &vars, int node, int choice = 0) const;
+	bool NodeInRange(int node) const {
+		return node >= 0 && static_cast<unsigned>(node) < nodes.size();
+	}
```
Amazinite, line 97: We typically don't define functions in header files. It seems to me like this would be better as a namespace function in Conversation and ConversationPanel. (It's also only used once, in ConversationPanel, right now, so should probably check where else it can be put to use.)

source/Conversation.h
```diff
@@ -33,6 +33,17 @@ class Sprite;
 // killed. A conversation can also branch based on various condition flags that
 // are set for the player, or even trigger various changes to the game's state.
 class Conversation {
+	class Data {
```
Amazinite, line 36:
```suggestion
public:
	class Data {
```

source/Conversation.h
```diff
@@ -33,6 +33,17 @@ class Sprite;
...
+	public:
+		explicit Data(std::string text, int next)
+			: text(std::move(text)), next(next) {}
+		// The text to display:
+		std::string text;
+		// The next node to visit:
+		int next;
+		// Conditions for displaying the text:
+		ConditionSet conditions;
+	};
```
Amazinite, line 46:
```suggestion
	};


```

source/Conversation.cpp
```diff
@@ -523,3 +577,14 @@ void Conversation::AddNode()
 	nodes.emplace_back();
 	nodes.back().data.emplace_back("", nodes.size());
 }
+
+
+
+bool Conversation::HasCondition(const DataNode &node)
```
Amazinite, line 583: As mentioned, functions should appear in the same order as in the header, so this should be moved to below LoadGotos.

source/Conversation.h
```diff
@@ -75,11 +86,15 @@ class Conversation {
 	int Choices(int node) const;
 	bool IsBranch(int node) const;
 	bool IsAction(int node) const;
-	const ConditionSet &Branch(int node) const;
+	const ConditionSet &Conditions(int node) const;
 	const GameAction &GetAction(int node) const;
 	const std::string &Text(int node, int choice = 0) const;
 	const Sprite *Scene(int node) const;
 	int NextNode(int node, int choice = 0) const;
+	bool ShouldSkipText(const std::map<std::string, int64_t> &vars, int node, int choice = 0) const;
```
Amazinite, line 94: As with below, this is only used in ConversationPanel right now, so might it be better to define it there?

source/Conversation.cpp
```diff
@@ -440,40 +458,76 @@ int Conversation::NextNode(int node, int choice) const
...
+
+
+
+// Parse the children of the given node to see if then contain any "gotos," or
+// "to shows." If so, link them up properly. Return true if gotos or
+// conditions were found.
 bool Conversation::LoadGotos(const DataNode &node)
 {
-	bool hasGoto = false;
+	bool hasGoto = false, hasCondition = false;
```
Amazinite, line 485:
```suggestion
	bool hasGoto = false;
	bool hasCondition = false;
```

source/Conversation.cpp
```diff
@@ -269,10 +270,26 @@ void Conversation::Save(DataWriter &out) const
...
 				// Break the text up into paragraphs.
-				for(const string &line : Format::Split(it.first, "\n"))
+				for(const string &line : Format::Split(it.text, "\n"))
+				{
 					out.Write(line);
+					// If the conditions are the same, output them for each
+					// paragraph.
+					//
+					// (We currently don't merge paragraphs with identical
+					// ConditionSets, but some day we might.
```
Amazinite, line 280:
```suggestion
					// If the conditions are the same, output them for each
					// paragraph. (We currently don't merge paragraphs with
					// identical ConditionSets, but some day we might.)
```

Amazinite (changes requested):
Another style pass.

source/Conversation.h
tehhowch: `Data` doesn't need to be a public class that all other classes can see and use.

Instead, it should be moved below the public methods and be explicitly declared private.

source/Conversation.h
tehhowch: I disagree; the conversation is in charge of the node chain. The panel renders the node chain and handles player input.

Moving this to the panel means the panel has to know more about what makes a node belong in the displayed chain, rather than trusting the conversation to do that work.

source/ConversationPanel.cpp
```diff
@@ -340,38 +342,70 @@ void ConversationPanel::Goto(int index, int selectedChoice)
...
-	// Not every conversation node allows a choice. Move forward through the
-	// nodes until we encounter one that does, or the conversation ends.
-	while(node >= 0 && !conversation.IsChoice(node))
+	while(node >= 0)
 	{
-		int choice = 0;
-		if(conversation.IsBranch(node))
+		// Not every conversation node allows a choice. Move forward through the
+		// nodes until we encounter one that does, or the conversation ends.
+		while(node >= 0 && !conversation.IsChoice(node))
```
tehhowch, line 349: I feel like we should be able to avoid indenting this whole block, but I haven't truly checked that.

source/Conversation.h
SolraBizna: Apparently I was in the middle of following @tehhowch 's suggestion to make repeated bounds checks into private methods, and I got distracted.

I define methods in headers when they are trivial, like this one is. When compiling without LTO, doing otherwise prevents them from being inlined. Endless Sky gets compiled with LTO (in release builds), so this performance concern is irrelevant.

It can't be an anonymous-namespace function because it accesses a field. Shall I, then, simply move its implementation into `ConversationPanel.cpp`?

SolraBizna:
In the process of making the requested changes. I'm also thinking of changing `Conversation::Data` to `Conversation::ChoiceData` or even `Conversation::Choice`, so it's clearer what it actually is from the name of the type.

source/Conversation.cpp
```diff
@@ -180,14 +180,15 @@ void Conversation::Load(const DataNode &node, const string &missionName)
...
+			// node. Otherwise, just merge this new paragraph into the previous
+			// node.
+			if(nodes.empty() || !nodes.back().canMergeOnto || HasCondition(child))
 				AddNode();
 
 			// Always append a newline to the end of the text.
-			nodes.back().data.back().first += child.Token(0);
-			nodes.back().data.back().first += '\n';
+			nodes.back().data.back().text += child.Token(0);
+			nodes.back().data.back().text += '\n';
```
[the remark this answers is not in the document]
SolraBizna: I'm assuming you meant `'\n'` and not `"\""`

source/ConversationPanel.cpp
SolraBizna: The inner loop scans for the first choice node. Then, conditions are checked, and if every choice node was skipped because of conditions, the outer loop iterates again.

I'm not sure if there's a clearer way to represent the desired code flow here. It's a bit of a weird case.

source/Conversation.h
SolraBizna: I subscribe to the same reasoning as @tehhowch, but I could be convinced.

source/Conversation.cpp
SolraBizna: I hope 33b87e3a is satisfactory. I'll leave it to you to hit "resolve" if so.

source/Conversation.cpp
```diff
@@ -440,40 +458,76 @@ int Conversation::NextNode(int node, int choice) const
...
+		if(hasGoto && hasCondition)
 			child.PrintTrace("Warning: Ignoring extra text in conversation choice:");
 		else if(child.Size() == 2 && child.Token(0) == "goto")
 		{
 			// Each choice can only have one goto
-			Goto(child.Token(1), nodes.size() - 1, nodes.back().data.size() - 1);
-			hasGoto = true;
+			if(hasGoto)
+				child.PrintTrace("Warning: Ignoring extra endpoint in conversation choice:");
+			else
```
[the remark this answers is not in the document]
SolraBizna: I'm not sure I follow the requested change. Would this mean making an:
```c++
else if(child.Size() == 2 && child.Token(0) == "goto" && hasGoto) {
```
branch for the "ignoring extra endpoint" diagnostic? If so, I don't think it's clearer that way.

SolraBizna:
I did another round of changes, and am currently in the process of figuring out why I've broken conversations entirely (while also increasingly doubting whether I'm well enough to work on this today after all).

SolraBizna:
Conversations work again, and I've taken the opportunity to clarify documentation and nomenclature a bit. (My confusion over just what `Conversation::data` was used for, and the fact that `choice` parameters can be applicable even for nodes that aren't `"choice"` nodes, led directly to the bug I introduced in 33b87e3 and fixed in 6a67558. Hopefully now nobody else will get confused the same way.)

source/Conversation.cpp
```diff
@@ -434,46 +452,122 @@ const Sprite *Conversation::Scene(int node) const
...
+			{
+				Goto(child.Token(1), nodes.size() - 1, nodes.back().elements.size() - 1);
+				hasGoto = true;
+			}
+		}
+		else if(child.Size() == 2 && child.Token(0) == "to" && child.Token(1) == "show")
```
Amazinite, line 530: How about "display" instead of "show."

SolraBizna: That may be better. I'll think about it some more.

SolraBizna: I was leaning towards `to display`, but then I found that there is currently one other use of `to show`: News. Should we mirror that usage, or is it better to have a different phrasing here?

Amazinite: Well there's no other `to *` phrase that's duplicated anywhere else that I can recall, hence why I suggested the alternative of `display`.

SolraBizna: Just ran a quick search:

- `to show` in News
- `to spawn`/`despawn` in NPC
- `to` in Interface (doesn't count)
- `to offer`/`complete`/`fail` in Mission

There are indeed no existing pairs. Since every `to *` phrase is currently unique, I'll go ahead and make this one `to display`.

source/Conversation.cpp
```diff
@@ -434,46 +452,122 @@ const Sprite *Conversation::Scene(int node) const
...
+{
+	if(!NodeIsValid(node))
+		return false;
+	else if(IsChoice(node) ? !ElementIsValid(node, element) : element != 0)
+		return false;
+	const auto &data = nodes[node].elements[element];
+	if(data.conditions.IsEmpty())
+		return true;
+	else
+		return data.conditions.Test(vars);
```
Amazinite, line 476: Seems like we can ditch the `else`s here, since it's going to fall through to the next check regardless.
```suggestion
	if(IsChoice(node) ? !ElementIsValid(node, element) : element != 0)
		return false;

	const auto &data = nodes[node].elements[element];
	if(data.conditions.IsEmpty())
		return true;
	return data.conditions.Test(vars);
```

source/Conversation.cpp
```diff
@@ -434,46 +452,122 @@ const Sprite *Conversation::Scene(int node) const
...
-// Parse the children of the given node to see if then contain any "gotos."
-// If so, link them up properly. Return true if gotos were found.
+// Returns true if the given node index is in the range of valid nodes for this
+// Conversation.
+bool Conversation::NodeIsValid(int node) const
+{
+	if(node < 0)
+		return false;
+	else
+		return static_cast<unsigned>(node) < nodes.size();
```
Amazinite, line 488: Ditto.
```suggestion

	return static_cast<unsigned>(node) < nodes.size();
```

source/Conversation.cpp
```diff
@@ -434,46 +452,122 @@ const Sprite *Conversation::Scene(int node) const
...
+// Conversation *and* the given element index is in the range of valid elements
+// for the given node.
+bool Conversation::ElementIsValid(int node, int element) const
+{
+	if(!NodeIsValid(node))
+		return false;
+	else if(element < 0)
+		return false;
+	else
+		return static_cast<unsigned>(element) < nodes[node].elements.size();
```
Amazinite, line 503: Ditto.
```suggestion
	if(element < 0)
		return false;

	return static_cast<unsigned>(element) < nodes[node].elements.size();
```

Amazinite (approved):
The Conversation code is admittedly not one of my areas of expertise, but this looks pretty good to me. I'll leave the merging to @tehhowch for if there's any lingering comments he still has.

source/Conversation.cpp
SolraBizna: I like to leave `else`s in in cases like this, to make it more explicit to the reader. But that's purely a matter of style.

I'll snip all three `else`s next time I'm in a position to do so.

SolraBizna:
> I'll snip all three `else`s next time I'm in a position to do so.

I'd actually forgotten that I was going to do this. I'll try to get to it soon.

Is there anything else I need to do? Should I rebase it again?

tehhowch:
> Is there anything else I need to do? Should I rebase it again?

If you wanted to implement conditional `scene` nodes you could, but otherwise this is waiting on me to pull it down and give it a pass. 

(Side note: buying a house is rather time consuming 😅)

source/Conversation.h
```diff
@@ -72,17 +72,61 @@ class Conversation {
 	// the user to select; others just automatically continue to another node.
 	// Nodes may also display images or include conditional branches.
 	bool IsChoice(int node) const;
+	// If the given node is a choice node, check how many choices it offers.
 	int Choices(int node) const;
```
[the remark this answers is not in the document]
SolraBizna: It returns the length of the vector in question, which has other semantics for different kinds of nodes. I remember parsing through the logic to figure out the details, but I don't remember the result. I think I added a comment to the vector itself explaining everything it's used for.

source/Conversation.h
```diff
@@ -72,17 +72,61 @@ class Conversation {
...
+	// Returns false if:
+	// - The node (or element) is out of range.
+	// - The node is not a choice node, and element is non-zero.
+	// - The node (or element) has conditions and those conditions are not met.
+	// and true otherwise.
+	bool ShouldShowText(const std::map<std::string, int64_t> &vars, int node, int element = 0) const;
+	// Returns true if the given node index is in the range of valid nodes for
+	// this Conversation.
+	// Note: The "outcomes" listed above do *not* refer to nodes *within* the
+	// Conversation, and are therefore considered invalid by this function!
```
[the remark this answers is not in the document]
SolraBizna: What about:

> Note: This function only considers actual Conversation nodes to be valid choices. The negative "outcome" values (listed above) are special sentinel values, rather than indices of Conversation nodes, and are therefore considered invalid by this function.

source/Conversation.cpp
```diff
@@ -127,12 +127,12 @@ void Conversation::Load(const DataNode &node, const string &missionName)
 
 				// Store the text of this choice. By default, the choice will
 				// just bring you to the next node in the script.
-				nodes.back().data.emplace_back(grand.Token(0), nodes.size());
-				nodes.back().data.back().first += '\n';
+				nodes.back().elements.emplace_back(grand.Token(0), nodes.size());
+				nodes.back().elements.back().text += '\n';
```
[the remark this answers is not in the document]
SolraBizna: I wrote it that way because of old habits. I honestly misread the code slightly until just now, in a way that made me think there was a (very small) performance penalty to doing it the other way.

source/Conversation.h
```diff
@@ -112,14 +157,17 @@ class Conversation {
 
 
 private:
-	// Parse the children of the given node to see if they contain any "gotos."
-	// If so, link them up properly. Return true if gotos were found.
+	// Parse the children of the given node to see if they contain any "gotos"
+	// or "conditions." If so, link them up properly. Return true if gotos or
+	// conditions were found.
 	bool LoadGotos(const DataNode &node);
```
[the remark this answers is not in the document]
SolraBizna: Maybe `LoadDestinations` or `LinkDestinations` or something? IIRC, it used to be used only for `goto`s, to set up the "next node" indices properly for that, and the additional work that I shoehorned in was setting up the "next node" indices properly for the skipped and not-skipped outcomes. So it doesn't load the condition *itself*, but it does act differently depending on whether a condition is present.

source/Conversation.h
```diff
@@ -72,17 +72,61 @@ class Conversation {
...
 	const Sprite *Scene(int node) const;
-	int NextNode(int node, int choice = 0) const;
+	int NextNode(int node, int element = 0) const;
+	// Returns whether the given node should be displayed.
+	// Returns false if:
+	// - The node (or element) is out of range.
+	// - The node is not a choice node, and element is non-zero.
+	// - The node (or element) has conditions and those conditions are not met.
+	// and true otherwise.
+	bool ShouldShowText(const std::map<std::string, int64_t> &vars, int node, int element = 0) const;
```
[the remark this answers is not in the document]
SolraBizna: That *would* match up with the (new) corresponding syntax from the data file. (It was originally `to show`, and is now `to display`.)

Should it be `ShouldDisplay` or `ShouldDisplayText`? The former would be more general, and if somebody adds `to display` support for `scene` nodes like @tehhowch wants then it's more correct. I'm unsure, though, because I tend to shy away from misinterpretably generic function names, and `ShouldDisplay` gives me a bad vibe in that respect.

SolraBizna:
I see I've got a little maintenance work to do here. I'll hopefully get to it tomorrow, along with the remaining suggested changes.

Terin:
Thanks for continuing to stick with this. As someone helping with building some complex missions, this will go SUCH a long way to improving that process. Hopefully, we'll be able to get this all wrapped up soon.

SolraBizna:
Shoot. I definitely wiped out a bunch of the smaller review changes with that force-push, and I'm not sure how to get them back.

warp-core:
Regarding my previous review, going further through the Conversation class just seems confusing to me. I can't really do a good review on this. I don't think this PR is responsible, and if the others can review it and are happy with it, I don't think this is a reason to not merge this. I will probably attempt to refactor the entire class at some point in the nearish future.

source/Conversation.cpp
```diff
@@ -271,17 +270,35 @@ void Conversation::Save(DataWriter &out) const
...
+				{
 					out.Write(line);
+					// If the conditions are the same, output them for each
+					// paragraph. (We currently don't merge paragraphs with
+					// identical ConditionSets, but some day we might.
+					if(!it.conditions.IsEmpty())
+					{
+						out.BeginChild();
+						{
+							out.Write("to", "display");
```
RisingLeaf, line 289:
```suggestion
							out.Write("to display");
```
easier and a tiny bit better performance

RisingLeaf (changes requested):
I have not looked into it in depth, but this seems to be a bit too complex for this kind of change?
I will have an in depth review soon.

source/Conversation.h
```diff
@@ -75,17 +75,63 @@ class Conversation {
...
+	//   optionally, a condition set which, if not met, prevents the text from
+	//   being displayed (without affecting which node is processed next).
+	// - In a choice node, there's one of these for each possible choice,
+	//   containing the text to display, the node the choice leads to, and,
+	//   optionally, the conditions under which to offer the choice.
+	// - In a branch node, there's two of these. The first one contains the
+	//   condition for the branch. If the condition is met, the "next" member
+	//   of the first element is followed. If it's not met, it's the second
+	//   element whose "next" member is followed.
+	class Element {
```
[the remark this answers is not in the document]
warp-core: I don't think it is unreasonable for this class to be nested (it is done elsewhere in the game).

I'm not sure why it shouldn't be private? This is the only class that needs it in this implementation.

source/Conversation.cpp
warp-core: Doing this means needing to change the parsing code further down. That expects `to` and `display` to each be their own token, with this change, it would become `"to display"`, a single token. I'm not entirely sure it should be a single token. While we may not have other things after the `to` now, we might in the future, and other places where something similar is present (missions, for example), two separate tokens are used.

RisingLeaf: If you want it like this I am ok with it

source/Conversation.h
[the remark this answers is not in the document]
RisingLeaf: We always make classes public I think?

Also while it may not be used yet, there is still the possibility that it may be needed somewhere else later on.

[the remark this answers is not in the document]
SolraBizna: It used to be a `std::pair<std::string, int>`. I made it a class so that it would have informative field names (instead of `first` and `second`), and so I could add a third field. It's still basically just a tuple that exists as an implementation detail of `Conversation`.

Both versions of this file have `class Node` nested inside `Conversation`.

Since `Conversation::Element` and `Conversation::Node` both contain references to specific node indices within a specific `Conversation`, it wouldn't really make sense to have either independent of the context of a containing `Conversation`.

SolraBizna:
I see that there's some activity here. Fortunately, no new merge conflicts have arisen that I can see. Unfortunately, I will *definitely* not have the ability to do any significant maintenance on this PR in the next few months. That means that, if further polish or changes are needed, it will likely have to be done without me. I'm very sorry about this.

RisingLeaf:
> I see that there's some activity here. Fortunately, no new merge conflicts have arisen that I can see. Unfortunately, I will _definitely_ not have the ability to do any significant maintenance on this PR in the next few months. That means that, if further polish or changes are needed, it will likely have to be done without me. I'm very sorry about this.

I dont want to destroy your hopes and I really want this PR, but without tehhowch there is not much to be done here. Sadly they are very busy sometimes so it can take a long time for this to get merged.

source/Conversation.h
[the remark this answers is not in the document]
RisingLeaf: Anyways, class definitions to the top. You can have it private, but in the top

source/Conversation.cpp
```diff
@@ -188,14 +187,14 @@ void Conversation::Load(const DataNode &node, const string &missionName)
 		{
 			// This is just an ordinary text node.
 			// If the previous node is a choice, or if the previous node ended
-			// in a goto, create a new node. Otherwise, just merge this new
-			// paragraph into the previous node.
-			if(nodes.empty() || !nodes.back().canMergeOnto)
+			// in a goto, or if the new node has a condition, then create a new
+			// node. Otherwise, just merge this new paragraph into the previous
+			// node.
+			if(nodes.empty() || !nodes.back().canMergeOnto || HasCondition(child))
```
RisingLeaf, line 193: I dont really know why HasCondition has to be a function, as it is only called once.
I am not particular against it, but I want to know why.

SolraBizna: It makes the code easier to read, especially since actually answering the question "does the given `DataNode` have a condition?" requires a for loop.

Hurleveur (approved):
Just looked at it, this is definitely something I want for missions
The code looks good, there are a few things that could be moved around or renamed but I dont think any of it matters enough to delay merging. There is a good amount of comments so it should not be hard to maintain.

source/Conversation.h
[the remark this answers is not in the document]
petervdmeer: `ShouldDisplayNodeContent` seems to match well with the comments above, and content can both be text or scenes. (`ShouldDisplayNode` is a bit shorter and I guess also works here.)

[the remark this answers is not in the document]
SolraBizna: I really like that name. I'll apply it as soon as I am able.

source/Conversation.cpp
petervdmeer: What if we just remove the need for a loop by using `child.HasChildren()`? I would expect any node that has children of it's own not to be mergeable onto the previous node.

source/Conversation.cpp
```diff
@@ -442,46 +459,119 @@ const Sprite *Conversation::Scene(int node) const
...
+				child.PrintTrace("Warning: Expected goto, to show, or endpoint in conversation, found this:");
 		}
 	}
-	return hasGoto;
+	return hasGoto || hasCondition;
+}
+
+
+
+bool Conversation::HasCondition(const DataNode &node)
```
petervdmeer, line 568: The function name is not great; because this function only checks if the node contains a `to display`. (But we might want to remove the function anyway...

source/ConversationPanel.cpp
petervdmeer: Good to check such corner-cases; we don't want the conversation to get stuck because the player has to choose between zero choices and cannot select one.
A possibly cleaner way to handle this loop is to update the `IsChoice()` function to check if there is at least one valid choice, then we might not need the double loop.

source/Conversation.h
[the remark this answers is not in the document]
petervdmeer: `LoadDestinations` seems fine as name (it only links half of the time, the rest of the time it just stores unresolved labels for later linking).

petervdmeer:
Please also merge the master branch into this branch; this PR is based on a commit that doesn't even have the ConditionsStore in the codebase yet. (And not having a ConditionsStore is quite a bit of a thing for stuff that works with conditions.)

source/ConversationPanel.cpp
SolraBizna: The reason to have a double loop is for conversations structured like:

- Say thing 1
- Iff condition A, offer choice 1A
- Iff condition B, offer choice 1B
- Say thing 2
- Iff condition C, offer choice 2C
- Iff condition D, offer choice 2D
- Say thing 3

so that, if none of A, B, C, or D, then it acts just like:

- Say thing 1
- Say thing 2
- Say thing 3

source/Conversation.cpp
SolraBizna: I can't off the top of my head think of a reason why that wouldn't work, but something is tickling the back of my mind, saying there's some vanilla case that will be screwed up by that.

SolraBizna:
Do you think that will require any major changes to my code? (The answer affects how soon I can get to it.)

source/Conversation.cpp
```diff
@@ -442,46 +459,119 @@ const Sprite *Conversation::Scene(int node) const
...
+// "to shows." If so, link them up properly. Return true if gotos or
+// conditions were found.
 bool Conversation::LoadGotos(const DataNode &node)
 {
 	bool hasGoto = false;
+	bool hasCondition = false;
 	for(const DataNode &child : node)
 	{
-		if(hasGoto)
+		if(hasGoto && hasCondition)
```
petervdmeer, line 521: Do we still need this check here if we already check for double goto's and double to-display's? The last check in the else-part would also seem to catch the issues that are otherwise caught here.

source/Conversation.cpp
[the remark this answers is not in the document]
petervdmeer: Yes, I believe that is what tehhowch means, and having a single chain of if/else is likely easier to review. Single chains of if/else is also a pattern that is used often in other parts of the code. (And yes, it does result in some duplication, but as mentioned before; the first if statement might be taken out, and the check for multiple goto's doesn't need to check the amount of children).

UnorderedSigh:
If you need any help debugging or testing, please let me know. I'm in dire need of this for the Hai Reveal cleanup (#7594, #7570, and others eventually).

source/Conversation.cpp
petervdmeer: Right, it messes up the reload on the last goto-label in a savegame (that text-line would not be merged onto the previous text node while this code would merge it)...

last_goto_label

source/Conversation.cpp
petervdmeer: Maybe rename it to `HasDisplayRestriction` ?

SolraBizna:
> If you need any help debugging or testing, please let me know.

Any and all testing is always welcome. I'm just not able to do much serious work for the next 45 days.

petervdmeer:
> Do you think that will require any major changes to my code? (The answer affects how soon I can get to it.)

Just resolving some compile errors like this:
compile_error
The Conditions changed from being a map to being a ConditionsStore, so I guess it is just changing a variable type at a few places.

UnorderedSigh:
> Any and all testing is always welcome. I'm just not able to do much serious work for the next 45 days.

Are there any known bugs yet? For example, are there any conversation types that would fail on reload, or other oddities?

source/ConversationPanel.cpp
petervdmeer: Yes, if none of A or B, then the amount of choices after "Say thing 1" are zero. And if none of C or D, then the amount of choices after "Say thing 2" are zero. So I expect my proposal to work just as you specify here, because IsChoice() would return zero if no valid choice is found.

source/Conversation.cpp
SolraBizna: That name is much better. (You seem to be on a naming roll.)

source/ConversationPanel.cpp
petervdmeer: Maybe call the new function `HasAnyChoice()`?

SolraBizna:
> The Conditions changed from being a map to being a ConditionsStore, so I guess it is just changing a variable type at a few places.

That sounds like something I'd be able to get to soon. 🤞

> Are there any known bugs yet? For example, are there any conversation types that would fail on reload, or other oddities?

I tested all the edge cases I could think of, and I've been playing the game with my changes in place (when I play) and not had issues. My testing included testing through save/load. There was one weird anomaly once, but I couldn't reproduce it after a few hours of trying.

My attempt at automating the testing didn't go very well; some information that would be required to reconstruct the original data is lost on load, so I couldn't just use diff. Unit testing support was added to the engine after I made my changes; it might be worthwhile to write some unit tests.

source/ConversationPanel.cpp
SolraBizna: I like that name too.

If I'm able to revisit this in detail, I'll try it out with the suggested change. It seems like it should work.

UnorderedSigh:
It sounds like the remaining tasks are:

1. Update to master
2. Tidy up names and other reviewer stuff
3. Unit tests (which probably requires updating to master)
4. Test with real missions (Hai Reveal has good ones for this)

Do you want help with items 1-3?

UnorderedSigh:
Your test case doesn't work because you switched from `on show` to `on display` in the code. Also, I'm getting *huge* delays during the conversation. It just stalls for a while after my mouse click.

Here's a test case that works:
[test.txt](https://github.com/endless-sky/endless-sky/files/10026512/test.txt)

UnorderedSigh:
The slowness was because you were doing `++node` without checking if the node was valid, so it kept looping until `node` went negative. I've fixed that, and I've also updated to master. Constructing a better test case now...

UnorderedSigh:
Here's a test case based on a real mission. The conversation is not updating as variables are set though.

[test2.txt](https://github.com/endless-sky/endless-sky/files/10026645/test2.txt)

source/Conversation.h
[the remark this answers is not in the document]
UnorderedSigh: This is placed next to the inner `Node` class, so moving it to the top doesn't make sense in this file. (Unless you also want to move `Node` to the top.)

source/Conversation.cpp
UnorderedSigh: No, the check shouldn't be there because it will mask the more meaningful `"Warning: Ignoring extra condition in conversation choice:"` error.

UnorderedSigh:
I got this feature working and my PR is here: #7629. I think I resolved all reviewer comments, except one request I didn't understand, to add a function. That PR is still in draft status because:

1. I haven't added unit tests.
2. I don't know what Endless Sky development community social norms I'm breaking by hijacking a 2-year-old PR that was almost working and implements a feature I need.

UnorderedSigh (changes requested):
The fixes to get this working on top of master and address most (all?) reviewer comments are here:

https://github.com/endless-sky/endless-sky/pull/7629

I don't know how the test system works, so I haven't added unit or integration tests.

UnorderedSigh:
I can do a PR from my branch to your fork if you prefer.

UnorderedSigh:
By request of @Amazinite I have closed my draft PR and opened a PR to this branch here: https://github.com/SolraBizna/endless-sky/pull/1

UnorderedSigh:
I've been asking around for help with the integration tests as I don't know how to implement one.

SolraBizna:
>  Do you want help with items 1-3?

As much as I hate to give up control of my baby, that's the best way to make sure it actually gets done. :)

> The slowness was because you were doing `++node` without checking if the node was valid, so it kept looping until `node` went negative. I've fixed that, and I've also updated to master. Constructing a better test case now...

Yikes! I thought I fixed that forever ago. It might have been lost in a rebase.

> By request of @Amazinite I have closed my draft PR and opened a PR to this branch here: [SolraBizna#1](https://github.com/SolraBizna/endless-sky/pull/1)

Merged. Thank you so much!

source/ConversationPanel.cpp
UnorderedSigh: When I was implementing all the requested changes in the reviewer comments, I couldn't figure out what you were requesting here. The only effects I see of moving the loop out to a function is to increase the code length and reduce indentation. In my mind, one level of indentation doesn't matter, while longer code makes it harder to maintain. I don't have a strong feeling on these matters, I'm just trying to parse all of your comments.

I'm 93% certain I don't understand what you want, but if you can explain it to me, I can implement it.

source/Conversation.h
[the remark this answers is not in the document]
UnorderedSigh: I moved both to the top in my PR to this PR.

source/Conversation.h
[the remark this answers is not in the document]
UnorderedSigh: Did this for you.

source/Conversation.h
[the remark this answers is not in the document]
UnorderedSigh: Did this for you.

source/Conversation.cpp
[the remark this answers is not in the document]
UnorderedSigh: I turned this into a sequence of else ifs.

source/Conversation.cpp
UnorderedSigh: Did this for you.

Hurleveur:
I'll open a PR to your branch to add integration tests today

source/Conversation.cpp
UnorderedSigh: I removed that "if" entirely, because it can hide the more specific error messages. I'm only 98% certain of that, so if people could skim over that code again, I'd appreciate it.

It'll break if you get rid of `&& hasCondition` and retain `hasGoto` though. You need either both or neither for the code to work.

Hecter94 (approved):
I'd like to see this merged sooner rather than later. It would open up many possibilities and make many things easier to do when it comes to mission writing.

warp-core:
As I understand it, `to display` and `goto` would both be child nodes of any of the conversation nodes, so the ordering shouldn't matter. (Though, we ought to decide on a convention for which order we use just for consistency, but that should not affect the code support for it.)

source/ConversationPanel.cpp
petervdmeer: SolraBizna introduced an additional `while(node >= 0)` loop at line 356. This additional loop was needed to deal with the situation where a player has choices, but all choices are hidden (due to all the choices having conditions, and the conditions hiding all possible choices). I was suggesting to not add the additional loop (so remove the loop now at line 356), but to call a function that checks if at least one choice can be chosen in the other loop (currently at line 360).

UnorderedSigh: Are you saying we should have a single loop with something like this?

```
while(node >= 0 && (!conversation.IsChoice(node) || !conversation.HasAnyChoice(node)))
```

UnorderedSigh: That change will simplify the code, so I see why you want it.

I guess HasAnyChoice() can call IsChoice(), letting you only have two conditions in the while().

UnorderedSigh: I implemented it, and put it in a PR to this branch here: https://github.com/SolraBizna/endless-sky/pull/3

UnorderedSigh: I had to handle a corner case. A choice with zero options is how the game represents "set the player's name." I had to differentiate between a choice with no options, and a choice with options, none of which are displayed.

UnorderedSigh:
I think the remaining problems will be resolved by these two PRs to @SolraBizna's branch:

1. https://github.com/SolraBizna/endless-sky/pull/2 - from @Hurleveur to add the tests
2. https://github.com/SolraBizna/endless-sky/pull/3 - from @UnorderedSigh to clean up the Goto function.

SolraBizna:
We could use a lot more tests. Writing these things is normally very much up my alley but I have so little time... 😿 Nevertheless, we have some tests now, including one to make sure that the "enter your name" feature doesn't accidentally go away again.

Hurleveur:
I think that 3 integration tests is already good enough (it makes this one of the most tested feature), but we could always make more in the future for sure.

SolraBizna:
More tests! All the tests! TEST ALL THE THINGS!

...Ahem.

tests/integration/config/plugins/integration-tests/data/tests/tests_conditional_choice.txt
```diff
@@ -0,0 +1,188 @@
...
+			key "Return"
+		call "Depart"
+		# use a test-data temporary variable for debugging in case of errors
+		apply
+			"test: %TEST%: NAME IN CONVERSATION: offered" = "%TEST%: NAME IN CONVERSATION: offered"
+			"test: <first>" = "<first>"
+			"test: <last>" = "<last>"
+		assert
+			"%TEST%: NAME IN CONVERSATION: offered" == 1
+			"<first>" == "F"
```
petervdmeer, line 187: Variable `"<first>"` is not set, so it will have a default value of 0. And conditions don't support strings, so I expect the `"F"` to be converted to a value of 0 as well. (The comparison returns true then, but doesn't really test anything.)
Same for the `"<last>"` variable; those are available as text-substitutions, but not as conditions.

petervdmeer: Btw, Derpy recently added an option to check the players name, then conditions should be something like `"first name: F" == 1` and similar for last-name.

Hurleveur: Well the goal wasnt to test that command I just thought lets go ahead and do it. 
I do have another PR that does just that though, so even if this one test does not work we'll figure it out in that other PR.

The main point of it was that you get a `name` thing happening and not a blank screen, I just thought may as well.

Now, in order to check for that still, we can just do it in an on accept branch, somehow?

Hurleveur: > Btw, Derpy recently added an option to check the players name, then conditions should be something like `"first name: F" == 1` and similar for last-name.

Yeah we should totally use that.
I know "name: F L" would work in this case, not sure about first name but it should work as well if it exists.

petervdmeer: [It exists](https://github.com/endless-sky/endless-sky/blob/d925975124dcf3ae67760b161b5e4459056ace5d/source/PlayerInfo.cpp#L2759)

petervdmeer (approved):
I noticed an issue in one of the new tests, but the code itself seems okay (and I tested name input manually; it seems to work).

SolraBizna:
Hooray! It went in!

Thank you everybody who reviewed and/or contributed, especially @Hurleveur and @UnorderedSigh for getting it over the final hurdle.

Issue #1643, opened 2016-09-28
Drone class ships cannot be piloted by the player


LocalGod79:
as it turns out, the Flycatcher can be outfitted with a Hyperdrive, Fuel Pod, Command Center, and enough other outfits to pass flight check, but cannot be made the Flagship (ridiculous as it seems).

i almost feel foolish raising it, but...(this issue is linked in the Steam thread that raised the point)

jamespglaze:
Hey look I made a problem!

KiLEdEnNis:
I remember when I could make a fighter my flagship
not sure what commit changed that

Zitchas:
Opening more possibilities for people's gameplay options seems like a good thing to me.

PR #5449, opened 2020-10-15
Allow carried ships to be the player's flagship

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #1643.

## Feature Details
Title says it all. Fighters, drones, or any carried ship category in the future can be piloted by the player as a flagship, if they so choose.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
<details><summary>Tested by changing my flagship to a Finch. Was able to successfully depart solo, as well as with carries as escorts. Carried ships as escorts work normally.</summary>

image
image
image
image
</details>

## Performance Impact
N/A

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2795,7 +2795,7 @@ void Ship::AddCrew(int count)
 // Check if this is a ship that can be used as a flagship.
 bool Ship::CanBeFlagship() const
 {
-	return !CanBeCarried() && RequiredCrew() && Crew() && !IsDisabled();
+	return RequiredCrew() && Crew() && !IsDisabled();
 }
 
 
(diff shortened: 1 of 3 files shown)


WarlordMike:
Do it, man. DO IT!

source/Ship.cpp
```diff
@@ -2990,9 +2990,16 @@ bool Ship::CanCarry(const Ship &ship) const
 
 
 
+void Ship::IsFlagship(bool isFlagshipChange)
```
petervdmeer, line 2993: Can we change this to something like:
```
void Ship AllowBeCarried(bool allowShipToBeCarried)
{
	canBeCarried = allowShipToBeCarried && BAY_TYPES.count(attributes.Category()) > 0;
}
```

Duplicating data makes the code more complex, so information like "Which ship is the players flagship?" should be stored only on 1 location (in PlayerInfo), not in every ship.

But the mechanism you are proposing could work. We just store if a ship is allowed to be carried (instead of storing if it is a flagship).

petervdmeer (changes requested):
I disagree with the flagship references in Ship, but those could be fixed by using a different name.

petervdmeer (approved):

starmlerp:
Escape Pods are becoming a thing, I see, too bad they will have little use until we implement hardcore mode.

Zitchas:
Not necessarily - Plenty of people like playing "realistic" games without being tied down by hardcore mode.

PR #5451, opened 2020-10-16
Feat: Add 'D' keybind to finish conversation

Terin:
## Feature Details
This PR adds the ability to press 'D' to finish conversations. Currently, you can use the number keys to quickly and seamlessly make dialog choices with the left side of the keyboard, but have to move to the right side or the mouse to finish them. Using the 'D' key would make finishing conversations consistent with various other dialogs through Endless Sky, which can be confirmed with the same binding.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
Finished the conversations at the beginning of the game, with the banker and James, by pressing the D key and return, to verify they both continue to work. Continues to work with both lower and uppercase D, and doesn't cause conflicts on the naming panel.

## Performance Impact
N/A

### DIFF
--- a/source/ConversationPanel.cpp
+++ b/source/ConversationPanel.cpp
@@ -206,7 +206,7 @@ bool ConversationPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &comm
 	if(node < 0)
 	{
 		// If the conversation has ended, the only possible action is to exit.
-		if(isNewPress && (key == SDLK_RETURN || key == SDLK_KP_ENTER))
+		if(isNewPress && (key == SDLK_RETURN || key == SDLK_KP_ENTER || key == 'd'))
 		{
 			Exit();
 			return true;


Amazinite:
I'd imagine that if a certain key can be used during conversations then it could also be used during dialogs. For example, enter will end both a conversation and a dialog. Problem is, D is already used for Decline in dialogs.

Terin:
I'm not sure about that. By that logic, you shouldn't be able to use D to exit hails, because those are also about communication and dialog. We already trust players to adjust to context-sensitive binds as well, and the conversation and dialog UI are fairly distinct on the user end. Accept/decline dialogs are also kind of the odd one out compared to 9 other uses of the D bind as a general purpose exit (some that don't even have a word with D on screen, like Leave in the shipyard/outfitter), and they're pretty rare in general, that I can't see them suddenly gaining primacy for such low-stakes operations.

PR #5459, opened 2020-10-18
feat: Introduce a capability to limit the rotation of a turret.

### LINKED ISSUE #3148: Feature Request: Limited turret turning angle
Hello, I would like to ask for a feature: limited turret turning angle, from the game a turret can turn and fire at any angle it wants but turrets shouldn't be firing at an angle like this V.(The part I circled in red)

That part supposes to be higher than the mount as I see here.
ice_screenshot_ -

My main problem isn't just the ones above, but I also want to make special turret mount that the turret shouldn't be facing this way
weird_turret
I wanted that gun to stay on the curved parts but don't go inside the hull.
This feature allows more variety of ship designs so that we could have ships with turret mount something like this!
default
This will also allow players to use more pilot skills as they can attack enemy ship on a blind spot where less of the ship's turret can fire at!
I can already imagine more ship design that could use this feature to make them looks better we could even make ships that need to face their starboard against the enemy to be able to use their weapons more effectively similar to these warships!
default
ijn_kitakami_torpedo_cruiser-49317

From my idea the turret that are limited in firing angle would be like
`turret [x] [y] [first angle] [second angle]`
So typing in like this
`turret 2 8 0 90`
would result in turret that can only fire at this degree
90degree
Here some example what it would be like using my idea of angle limiting.
degree
So in the picture above I've also demonstrated what if only one number is defined after turret position numbers, it could be either any angle other than 0 to that number or from that number to 360/359.
So each ship would have their own firing angle defined, the ones that aren't limited don't have to define those numbers.

From this, firing angle would be limited if the specific turret hardpoint has the angle defined and if you move the turret to another mount it won't be limited anymore, so with this idea I think it won't need to change the engine as much as what people said in issue #3053 right?

Thanks!

oo13:
**Feature:** This PR implements the feature request detailed and discussed in issue #3148.

MZ closed #3148 but, as far as I know, MZ and devs have not refused it.

## Feature Details
This PR introduces a capability to limit the rotation of a turret, and it makes us possible to place the turrets in the arrangement of [superfiring](https://en.wikipedia.org/wiki/Superfiring).

New syntax is similar to angled gun-port (#4801), and I owe this code to #4801, of course.
```
	# The turret is omnidirectional and points to the front when idling.
	turret x y outfit_name
	
	# The turret is omnidirectional and points to 'baseAngle' when idling.
	turret x y outfit_name
		angle baseAngle
	
	# The turret can rotate in the range from 'minAngle' to 'maxAngle' (clock orientation) and points to the center of the range when idling.
	turret x y outfit_name
		arc minAngle maxAngle
	
	# The turret can rotate in the range from 'minAngle' to 'maxAngle' (clock orientation) and points to 'baseAngle' when idling.
	turret x y outfit_name
		angle baseAngle
		arc minAngle maxAngle
```
And the turrets calculate the idle position considering the "parallel" parameter.

For Example:
```
	turret 0 -40 "Test Turret"
		angle -60
		arc -120 120
	turret 0 40 "Test Turret"
		angle 40
	turret 40 80 "Pug Anti-Missile"
		angle 180
		arc 90 180
```
A weapon also has an attribute "arc", which restricts the range where the turret can traverse. If the weapon is installed on a hardpoint with a limited arc, the narrower arc is used.

## UI Screenshots
Current Version:
current
Noooo.

This PR:
thisPR

Point to the front (same results in both versions):
forward

## Usage Examples
I provide this plugin [test-for-turret-v6.zip](https://github.com/endless-sky/endless-sky/files/5854996/test-for-turret-v6.zip) and a savefile [savefile-v6.zip](https://github.com/endless-sky/endless-sky/files/5854999/savefile-v6.zip).

## Testing Done
- Lock a ship and rotate my ship.
- Deselect any ships and rotate my ship when there are some enemy ships in the system.
- Select Opportunistic Tracking and fly a ship for a while.
- Save and reload.
- Set 'baseAngle' out of range from 'minAngle' to 'maxAngle' and ES outputs a warning message.
- Install/uninstall/save/load weapons that has an attribute "arc".
- Defend missiles with an anti-missile turret that is restricted arc.

## Performance Impact
- Slightly slow down because this PR adds some condition checks for all omnidirectional turrets.
- And for the directivity turret AI.cpp checks the angle every expected target.
- I think nearly all humans cannot feel the delay.


Zitchas:
Thank you for making this PR. I haven't tested it yet, but I had been hoping for this and similar features to be implemented. 

Potential uses:
- Enables the creation of "partial use" turret slots that aren't as useful as full turret slots because they are limited to certain firing arcs (for instance, nose or tail gunners)
- Enables more realistic directional restrictions for large ships such as battleships where turrets on one side may not be able to fire over the bulk of the ship at targets on the opposite side.
- Moves us closer to fully supporting Escape Velocity, Escape Velocity Override, and Escape Velocity Nova ports; all of which had directionally limited turrets (and swivel canons)

From what I can see here, the restrictions are based on the turret slot. Would it be possible to have the restrictions placed on the outfit as well?

The reason I ask is because the Escape Velocity series had a few outfits that occupied an interesting space in terms of weaponry, such as the rear turret. This was effectively a blaster turret that was restricted to firing in a 90 degree arc centered on the rear of the ship. While its stats were basically identical to a regular turret of its type, it took up significantly less space as a trade-off for the fact that it didn't have the full traversal capabilities of a regular turret. This allowed it to be used on a few ships (such as the rapier, a heavy interceptor style ship) where mounting a full turret wouldn't have made much sense. As a general rule, the rear-turret was not a super effective weapon, but it did help the rapier deal with pursuing fighters, making it situationally quite valuable. 

As such, I think it would make for an interesting pair of abilities to have both the ability to specify and limit firing arcs on the turret slot, as you have done here; as well as on outfits themselves, enabling specialized turrets designed for these restricted slots. Thus we would have a situation where any turret could be installed in the restricted turret slot, and work fine (just with the restricted arc); and we could also have turrets that don't have a full traversal that save some space or are cheaper or have other trade-offs that can be installed anywhere, but are intended for these restricted slots. 

I don't think this additional ask should delay this particular PR; I would be more than happy to have these restricted turret slots merged now and the later version that enables a parallel outfit version to come later.

Thanks again!

oo13:
I would like to implement a weapon limited rotation, and the weapon specified a limited angle will be able to rotate right and left from the base angle as a center.

I don't realize any problem for the hardpoint that specifies an angles in this feature, but the default behavior is problematic.

Do you think what appropriate base angle is by default? If we install the turret specified a limitation on the existing ship, what direction the turret should point to? The front or outside?

In this PR now, the base angle will be equal to the hold position, that is the front of the ship. In the current ES, the turret point to the front of the ship when holding, but the base angle points outside from the center of the ship (although we can notice it only when departing a planet right after loading).

In my opinion, the turret should point to the front of the ship when holding because it uses for mining, but it's ridiculous that a hardpoint in the rear directs forward, or that a turret struggles to face the front of the ship when we install the turret specified a rotation limit to a turret mount that directs to backward.

Should the default base angle be outside only when a turret limiting rotation installs on it?

~~Ummm, coding in C++ is far easier than writing in English.~~

Zitchas:
I think turrets should default to pointing forwards unless the restrictions on its arc do not allow that.

Thinking about what I said before, the weapon/outfit definition probably shouldn't generally specify a direction, but rather just an arc, which then combines with the turret hardpoint's direction. 

Swivel canon example situations and how I think it should be handled: ("Forwards" and "backwards" mean "straight forward parallel to the ship's longitudinal axis" and "backwards" means straight backwards parallel to the ship's longitudinal axis")
- Generic (unspecified direction) hardpoint: center of arc facing straight forward.
- Hardpoint with specified direction: The weapon's arc should be centered on the specified direction. 
    - If that arc includes straight forward, then its resting position should be forward. 
    - If it does not include straight forward, but does include straight backwards, then it should point backwards. 
    - If it includes neither straight forward nor straight backward, then it should default to being centered on its arc.

This means that we don't need to have "nose turrets" or "tail turrets", but instead can simply have a turret pod, and whether they point forwards or backwards (or, for that matter, sideways as broadside turrets) is purely determined by the slot that they are installed in. And I know there have been a bunch of people interested in having broadside weaponry.

oo13:
OK, then, if we install a weapon with restricted arc on a default hardpoint, it is directed to the front.

But for hardpoint with specified direction, it should be what it specifies because this PR allows us to specify three parameters, the minimum and maximum angle of the arc, and the resting position. We also have a free to omit some parameters.

source/Hardpoint.h
```diff
@@ -89,12 +99,16 @@ class Hardpoint {
 	const Outfit *outfit = nullptr;
 	// Hardpoint location, in world coordinates relative to the ship's center.
 	Point point;
-	// Angle of firing direction (guns only).
+	// Angle of firing direction (guns) or holding position (turret).
 	Angle baseAngle;
+	// Range of the capable angle to point to the target (turret only).
+	std::pair<Angle, Angle> movableAngle;
```
petervdmeer, line 105: I think a name like `turnRange` will reflect the meaning of this variable better than movableAngle.

Similar for the comment:
`// Range over which the turret can turn, from leftmost position to rightmost position (turret only)`

source/Hardpoint.h
```diff
@@ -30,7 +31,13 @@ class Visual;
...
 public:
 	// Constructor. Hardpoints may or may not specify what weapon is in them.
-	Hardpoint(const Point &point, const Angle &baseAngle, bool isTurret, bool isParallel, const Outfit *outfit = nullptr);
+	// If isTurret is true, angles have various parameters according to the size:
+	//  0: The turret is omnidirectional and points to the front when holding.
+	//  1: The turret is omnidirectional and points to the angles[0] when holding.
+	//  2: The turret can rotate in the range from angles[0] to angles[1] and points to the center of the range when holding.
+	//  >=3: The turret can rotate in the range from angles[0] to angles[1] and points to the angles[2] when holding.
+	// If isTurret is false, angles[0] specify the angle of firing direction.
+	Hardpoint(const Point &point, const std::vector<Angle> &angles, bool isTurret, bool isParallel, const Outfit *outfit = nullptr);
```
petervdmeer, line 40: I'm not too happy with using a vector here, especially since all 3 angles have a different meaning.
What if we just add a new variable pair<Angle> turnRange?

source/Hardpoint.h
```diff
@@ -42,11 +49,14 @@ class Hardpoint {
 	const Angle &GetAngle() const;
 	// Get the base angle that this weapon is aimed at (without harmonization/convergence), relative to the ship.
 	const Angle &GetBaseAngle() const;
+	// Get the movable angle. Return value is invalid if this is omnidirectional.
+	std::pair<Angle, Angle> GetMovableAngle() const;
```
petervdmeer, line 53: See my comment below about renaming to turnRange.
I would also expect omnidirectional turrets to get a sensible default return-value like pair(Angle(0.), Angle(0.)).

source/Hardpoint.h
```diff
@@ -89,12 +99,16 @@ class Hardpoint {
 	const Outfit *outfit = nullptr;
 	// Hardpoint location, in world coordinates relative to the ship's center.
 	Point point;
-	// Angle of firing direction (guns only).
+	// Angle of firing direction (guns) or holding position (turret).
```
petervdmeer, line 102: default/idle position sounds better to me than holding position.

source/Ship.cpp
```diff
@@ -239,13 +239,20 @@ void Ship::Load(const DataNode &node)
...
 			bool gunPortParallel = false;
 			if(child.HasChildren())
 			{
 				for(const DataNode &grand : child)
 					if(grand.Token(0) == "angle" && grand.Size() >= 2)
-						gunPortAngle = grand.Value(1);
+					{
+						const int end = min(4, grand.Size());
+						for(int i = 1; i < end; ++i)
+							angles.push_back(grand.Value(i));
```
petervdmeer, line 251: I feel that a separate keyword for the turn-range will be more clear;
`turnRange <min> <max>` => for specifying the allowed turnrange for turrets (and maybe also for guns?)

source/Ship.cpp
```diff
@@ -239,13 +239,20 @@ void Ship::Load(const DataNode &node)
 				if(child.Size() >= 2)
 					outfit = GameData::Outfits().Get(child.Token(1));
 			}
-			Angle gunPortAngle = Angle(0.);
+			vector<Angle> angles{};
```
petervdmeer, line 242: I prefer 3 separate variables or 2 separate variables where the turnRange is a pair.

turnRange.first is more readable than angle[1] in the code below.

source/Ship.cpp
```diff
@@ -765,17 +786,35 @@ void Ship::Save(DataWriter &out) const
 			else
 				out.Write(type, 2. * hardpoint.GetPoint().X(), 2. * hardpoint.GetPoint().Y());
 			double hardpointAngle = hardpoint.GetBaseAngle().Degrees();
-			if(hardpoint.IsParallel() || hardpointAngle)
+			if(hardpoint.IsTurret())
```
petervdmeer, line 789: I'm not to happy with the code for turrets and guns drifting further apart. I hope that we get to more generalized hardpoint types in the future (where some gun slots can also turn somewhat and where some guns might also have a hardpoint sprite and where we also get launchers as a separate hardpoint type).

source/Angle.cpp
```diff
@@ -171,6 +179,17 @@ Point Angle::Rotate(const Point &point) const
 
 
 
+// Judge whether this is inside from "base" to "limit."
+// The range from "base" to "limit" is expressed by "clock" orientation.
+bool Angle::isInRange(const Angle& base, const Angle& limit) const
+{
+	Angle normalizedLimit = limit - base;
+	Angle normalizedTarget = *this - base;
+	return normalizedTarget.angle <= normalizedLimit.angle;
```
petervdmeer, line 188: I'm not sure I understand this code.
And isn't it easier to just perform 2 comparisons `this.angle >= base.angle && this.angle <= limit.angle`?

source/AI.cpp
```diff
@@ -2720,7 +2720,8 @@ void AI::AimTurrets(const Ship &ship, Command &command, bool opportunistic) cons
 			{
 				// Get the index of this weapon.
 				int index = &hardpoint - &ship.Weapons().front();
-				double offset = (hardpoint.HarmonizedAngle() - hardpoint.GetAngle()).Degrees();
+				double offset = (hardpoint.GetBaseAngle() + hardpoint.HarmonizedAngle()
```
petervdmeer, line 2723: Do we really need to give a default where the turret aims at (and provide that as keyword)? (I would expect harmonized-forward if harmonized-forward is in range, otherwise normal forward if it is in range, and otherwise the center of the range).

source/Hardpoint.cpp
```diff
@@ -39,9 +39,30 @@ namespace {
 
 
 // Constructor.
-Hardpoint::Hardpoint(const Point &point, const Angle &baseAngle, bool isTurret, bool isParallel, const Outfit *outfit)
-	: outfit(outfit), point(point * .5), baseAngle(baseAngle), isTurret(isTurret), isParallel(isParallel)
+Hardpoint::Hardpoint(const Point &point, const vector<Angle> &angles, bool isTurret, bool isParallel, const Outfit *outfit)
+	: outfit(outfit), point(point * .5), isTurret(isTurret), isParallel(isParallel)
 {
+	if(isTurret)
```
petervdmeer, line 45: I prefer a solution where we don't need to add a body to the constructor (and I think that is possible).

petervdmeer (changes requested):
Nice to see this feature being added. I do have some review comments about code readability/maintainability.
And if we want to have new hardpoint types in the future, then we should drive for commonality between guns and turrets also in PRs like this.

source/Hardpoint.h
oo13: I changed the name in local before reading your review. It's "angleOfTraverse" now.

Do you think which is name better?

source/Hardpoint.h
oo13: I agree for not using a vector. I hesitated to add a dependence on Hardpoint.h to Armament.h at that time.
It's need to provide a holding position.

source/Hardpoint.h
oo13: I prefer to return a value pair(Angle(baseAngle+180), Angle(baseAngle+180)), but anyway I'm not sure the use-case.

Technically speaking, the 360 degree arc is different from omnidirectional in the hardpoint. The latter can rotate infinitely. I know it's trivial...

source/Hardpoint.h
oo13: Thanks.

source/Ship.cpp
oo13: "turnRange", "turn range" or "angle of traverse". I'll use what you like.
And I'll create "idle angle", "default angle", or something.  I would like to know your suggestion.

source/Ship.cpp
oo13: I'll try to generalize the code, because ES doesn't warn redundant parameters...

source/Angle.cpp
oo13: It chooses an edge of the arc as the reference angle (base) and compares relative angles to decide whether this is in the range. It assumes all angles are in 360 degrees from the reference angle.

source/AI.cpp
oo13: An appearance of a ship depends on the idle position of turrets.
It can create a turret like the second and the third turret of [HMS Dreadnought](https://en.wikipedia.org/wiki/File:HMS_Dreadnought_(1911)_profile_drawing.png) and to shoot an asteroid when mining as same as a current behavior. It points to the front, but possibly someone wants it points the back.

source/Hardpoint.h
petervdmeer: Both turnRange as well as angleOfTraverse sound better to me than movableAngle. I'm not a native English speaker, so I leave it to other reviewers to comment on the most proper name.

source/AI.cpp
petervdmeer: This can be a nice feature to have, but I'm not sure this is required now. We can also limit the rotation of a turret without providing a default holding/idle position.

source/Angle.cpp
petervdmeer: Those calculations are in the range from 0x0 to 0x10000, not in the range from 0 to 360. But your explanation sounds good and this probably also works in the range to 0x10000.

source/Hardpoint.h
petervdmeer: I'm not sure we need to provide a holding position, we don't do that today either.

source/Hardpoint.h
petervdmeer: Such differences can be quite relevant if a turret indeed can rotate 360 degrees but needs to cross the 1 point where it cannot turn (meaning that the turret needs a lot more in-game time to perform the turn).
Nice to differentiate between the 2 situations.

source/Ship.cpp
petervdmeer: My preference for the keyword is `turn range`, but I'm not a native English speaker, so you might want to get some input from other reviewers for this choice.

source/Hardpoint.h
oo13: Sorry, I'm wrong. 360 degree means 0 degree.
Anyway, it returns pair(Angle(baseAngle+180), Angle(baseAngle+180)) now.

ravenshining:
_"angle of traverse"_ seems a bit wordy.  Just _traverse_ would be kinder on those typing up content.

oo13:
Thanks.
So, the candidates are "turn range" and "traverse" now.

oo13:
I committed a "turn range" version for now.

Zitchas:
Sounding good. How testable is this? Or, for that matter, how close is this to being usable?

oo13:
I expect this works fine if you can accept the name "turn range".

Zitchas:
"turn range" is fine by me. "Angle of Traverse" is a technical surveying term for almost exactly what we have here, so that term would be great, but I think "turn range" is sufficient and descriptive enough to convey the intended meaning.

tehhowch:
I've not reviewed the code yet, but i think "swept angle" is probably a more accurate and expressive name for the keyword

Zitchas:
Could you explain that term for me? All the examples I've been able to find online refer to wing shapes.

Since we're dealing with turrets, it seems appropriate to stick with already used terminology. That way if someone googles the term they'll be able to get information sources that describe what a traverse means for a weapon. (and guns in particular.)

tehhowch:
> 
> 
> Could you explain that term for me? All the examples I've been able to find online refer to wing shapes.
> 
> Since we're dealing with turrets, it seems appropriate to stick with already used terminology. That way if someone googles the term they'll be able to get information sources that describe what a traverse means for a weapon. (and guns in particular.)

The tool-tip I imagine for "swept angle" is "The angular range of protection provided by this weapon, in degrees."

If we worked in 3D, this quantity would be measured in steradians as we'd be working with solid angles, but we're in a simpler 2D world. As far as the etymology of the term, consider the behavior of the game's opportunistic turrets with no selected/hostile target in-system: they repeatedly turn back & forth over the same angular range relative to the ship's facing: https://github.com/endless-sky/endless-sky/blob/c5588f9d0002f1765fd193d2daaa56c185a38e30/source/AI.cpp#L2736-L2738

With this PR then, we allow content creators to implement custom "swept angles"

Zitchas:
Thanks for the explanation. 

Given that, I still think that "Angle of Traverse" makes more sense than "Swept Angle" because the traverse terminology already exists for shipboard weaponry, cannons, and turrets; whereas the "Swept Angle", is a more generic / mathematical term that is not generally used in the real-life equivalent fields. 

That being said, this isn't a major issue for me. The former is a preference, in part based on the fact that I personally like it when my games teach me real-life terms and vocabulary where possible. If you are set on the alternative, though, I'm fine with it.

oo13:
I push a "swept angle" version now.

source/AI.cpp
```diff
@@ -2786,7 +2800,31 @@ void AI::AimTurrets(const Ship &ship, Command &command, bool opportunistic) cons
...
+				double degrees = 0.0;
+				Angle angleToPoint = Angle(p);
+				if(hardpoint.IsOmnidirectional())
+					degrees = (angleToPoint - aim).Degrees();
+				else
+				{
+					auto range = hardpoint.GetSweptAngle();
+					const Angle facing = ship.Facing();
+					range.first += facing;
+					range.second += facing;
```
tehhowch, line 2812: Seems like we should enable this automatically, i.e. pass the reference angle into the method, if this is a common usage of the method call. Otherwise the caller might forget to do this.
But, if this is a special scenario, then we should not.

(Right now this method is used 3 times, and 2 adjust the return value, while 1 does not.)

source/Hardpoint.cpp
```diff
@@ -241,14 +270,25 @@ bool Hardpoint::FireAntiMissile(Ship &ship, const Projectile &projectile, vector
 	double range = outfit->Velocity();
 	
 	// Check if the missile is within range of this hardpoint.
-	Point start = ship.Position() + ship.Facing().Rotate(point);
+	const Angle facing = ship.Facing();
```
tehhowch, line 273:
```suggestion
	const Angle &facing = ship.Facing();
```

source/Hardpoint.cpp
```diff
@@ -193,7 +210,19 @@ void Hardpoint::Aim(double amount)
...
-	angle += outfit->TurretTurn() * amount;
+	const double add = outfit->TurretTurn() * amount;
+	if(isOmnidirectional)
+		angle += add;
+	else
+	{
+		const Angle newAngle = angle + add;
+		if(add < 0. && sweptAngle.first.IsInRange(newAngle, angle))
+			angle = sweptAngle.first;
+		else if (add > 0. && sweptAngle.second.IsInRange(angle, newAngle))
```
tehhowch, line 221:
```suggestion
		else if(add > 0. && sweptAngle.second.IsInRange(angle, newAngle))
```

source/Hardpoint.cpp
```diff
@@ -282,6 +322,9 @@ void Hardpoint::Install(const Outfit *outfit)
 		this->outfit = outfit;
 		Reload();
 		
+		// Update swept angle.
+		UpdateSweptAngle();
+		
 		// For fixed weapons, apply "gun harmonization," pointing them slightly
 		// inward so the projectiles will converge. For turrets, start them out
 		// pointing outward from the center of the ship.
```
tehhowch, line 330: This comment should be updated

tehhowch (commented):
Another possible user-facing name for the functionality could be "angular coverage"

source/AI.cpp
```diff
@@ -2786,7 +2800,31 @@ void AI::AimTurrets(const Ship &ship, Command &command, bool opportunistic) cons
...
+					if(!angleToPoint.IsInRange(range.first, range.second))
+					{
+						// Decrease the priority of the target.
+						rendezvousTime += 2. * weapon->TotalLifetime();
+						
+						// Point to the nearer edge of the arc.
+						const double degree1 = (range.first - angleToPoint).Degrees();
+						const double degree2 = (range.second - angleToPoint).Degrees();
+						if(fabs(degree1) < fabs(degree2))
```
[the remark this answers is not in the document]
oo13: It is needed because Degree() returns the nearer angle.

source/AI.cpp
oo13: I would not change it because any methods of Hardpoint except firing don't consider the direction of the ship.

oo13:
I applied suggestions.

Also the idle position of the turrets is calculated when installing, instead of in-flight, and the turrets calculate the idle position considering the "parallel" parameter.

Amazinite:
I know the keyword has gone through a number of changes already, but why not just "arc"? Turrets would then have a (facing) angle and arc (of fire). "Swept angle" seems wordy/clunky to me. "Arc" makes more sense to me, since we are quite literally defining an arc of a circle within which the turret can fire.

oo13:
I post the "arc" version now.

Zitchas:
Unclear. There's conflict(s) to resolve, and debate about the terminology to use, and a few comments from Tehhowch that need to be resolved. I'm not enough of a coder to be able to estimate how quickly this could be brought up to mergeable state.

For the terminology: three options suggested:
- "Angular Coverage" : Descriptive term that basically explains itself clearly.
- "Arc" : Simple and to the point, and commonly used English term for a portion of a circle.
- "swept angle" : This is the proper terminology used in artillery and shipboard weaponry to describe this exact thing we are discussing.

My personal preference is for the "swept angle." If there is a real-life clearly defined term for what we are trying to describe, I feel that we should use it. It's relatively clear what it means, and has a completely unambiguous and precise real-world definition for those that know it, easily searchable for more information, and even encourages people to learn more proper real-world vocabulary, which I also feel is a desirable thing.

oo13:
I'm sorry for unclear state of my PRs.
You can take over any PRs initiated by me. I would answer your questions, but I would neither update nor test the PRs. Should I close them?

EjoThims:
> I'm sorry for unclear state of my PRs. You can take over any PRs initiated by me. I would answer your questions, but I would neither update nor test the PRs. Should I close them?

Sad to hear. :(

That decision should come down from others like Amazinite and Tehhowch, though, and will probably be made based on what they can readily make mergeable and what would continue to sit.

Probably best to go ahead and toss the same comment on any other outstanding PRs you have though, so we know which are in that state and to get them that attention.

oo13:
Thank you, 1010todd!

Issue #5220, opened 2020-08-11
Dying in motion preserves motion blur


Terin:
**Describe the bug**
When your flagship explodes, the motion blur effect on screen will stay stuck on whatever momentum you had as the ship exploded, even as action continues to play out in the system before you reload a save. This is most noticeable on stars, but appears to affect everything else as well.

**To Reproduce**
1. Enable motion blur
2. Load a save or create a new pilot, buying a fast, light ship like a Sparrow
3. Get your ship destroyed
4. Observe motion blur on stars or other objects

**Expected behavior**
Motion blur should be applied as if your ship was stationary.

**Screenshots**
Motion blur on stars after being dead for 20 seconds

![blur1](https://user-images.githubusercontent.com/4471575/89957696-b0f0c780-dbfd-11ea-9c38-ffb3efc894bd.png)

Comparison screenshot of nearly stationary ship
![blur2](https://user-images.githubusercontent.com/4471575/89957728-c49c2e00-dbfd-11ea-8791-d4e1a2c11b35.png)

**System (please complete the following information):**
 - OS: Windows 10 Pro version 1903 / Arch Linux
 - Game Source: Steam / Self-compiled
 - Version: 0.9.12 / `https://github.com/endless-sky/endless-sky/commit/990834a74d0429f920f7b2e5bb4a53b37f2af2ca`

**Link to save file**
None needed

PR #5464, opened 2020-10-22
Stop blurshader thinking there is centervelocity movement after player death

bobrobbow:
**Bugfix:** This PR addresses issue #5220 

## Fix Details
This being my first ever PR and my first hour of scanning through the code - I chose a very simple seeming issue, with little chance to sprawl and create more problems with the fix, that said however, I am obviously not completely au fait with the codebase, so apologies for any mistakes in my understanding.

From what I can see, the `centrevelocity` is what determines the blur shader strength, and the centrevelocity is set in Engine::step() based on flagship velocity, after the flagship "dies" I'm guessing that its velocity is never zeroed, hence the centrevelocity keeps accessing the "dead" ships velocity which is now frozen forever at its last updated value.

I added a conditional in Engine::Step() which asks if the flagship is destroyed, if so, set centrevelocity to Point(0,0) ( no movement ) and if not, uses the ships velocity as normal.

I dont know if this is the best place for this conditional, but it seems better than adding checks elsewhere further down the line in the drawing/shader code ( or changing the ship code to do more stuff when destroyed )

## Testing Done
Itd be hard to see how this change could introduce other errors, but I did the original test case from the linked issue of getting my ship destroyed while in motion, and the motion blur was gone, no other oddities observed ( i.e the screen still moves and stars still blur appropiately before the ship is destroyed )

## Save File
N/A

before: 

Webp net-resizeimage

after:

Webp net-resizeimage(1)

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -447,7 +447,7 @@ void Engine::Step(bool isActive)
 	else if(flagship)
 	{
 		center = flagship->Position();
-		centerVelocity = flagship->Velocity();
+		centerVelocity = flagship->IsDestroyed() ? Point(0,0) : flagship->Velocity();
 		if(doEnter && flagship->Zoom() == 1. && !flagship->IsHyperspacing())
 		{
 			doEnter = false;
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -447,7 +447,7 @@ void Engine::Step(bool isActive)
 	else if(flagship)
 	{
 		center = flagship->Position();
-		centerVelocity = flagship->IsDestroyed() ? Point(0,0) : flagship->Velocity();
+		centerVelocity = flagship->Velocity();
 		if(doEnter && flagship->Zoom() == 1. && !flagship->IsHyperspacing())
 		{
 			doEnter = false;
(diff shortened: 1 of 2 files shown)


source/Engine.cpp
```diff
@@ -447,7 +447,7 @@ void Engine::Step(bool isActive)
 	else if(flagship)
 	{
 		center = flagship->Position();
-		centerVelocity = flagship->Velocity();
+		centerVelocity = flagship->IsDestroyed() ? Point(0,0) : flagship->Velocity();
```
Amazinite, line 450: The default value of a point is already (0,0). (See a few lines above, `centerVelocity = Point();`.)
```suggestion
		centerVelocity = flagship->IsDestroyed() ? Point() : flagship->Velocity();
```

I think it might make more sense though to set the velocity of a destroyed ship to 0, instead of always checking if the flagship is destroyed. That is, add `velocity = Point();` here, say after L1294.
https://github.com/endless-sky/endless-sky/blob/29caeb623f574eeab2ad1e46bb654e6749c7f623/source/Ship.cpp#L1291-L1296

Amazinite (changes requested):

source/Engine.cpp
bobrobbow: Makes sense, moved to ship:: so it dosnt need to check every engine::step() tested and has same end effect.

Issue #5061, opened 2020-05-13
Autopilot tries (and fails) to land on spriteless planets


beccabunny:
**Describe the bug**
Currently, the easiest way by far to make one-way wormholes is not to give the exit wormhole a sprite. However, the autopilot still tries to land on it, simply going towards it over and over trying to get as close as possible, but never actually landing.

**To Reproduce**
1. Create a planet without a sprite
2. Depart and go to its system
3. Press "L"

**Expected behavior**
The autopilot ignores planets without a sprite, given that there is no reason I can see to do so which should end up with the player actually landing there.

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Github
 - Version: 0.9.10

PR #5466, opened 2020-10-23
Add check for planet having a sprite before returning valid planet to land on

bobrobbow:
**Bugfix:** This PR addresses issue #5061 

## Fix Details
As per linked issue, autopilot fails to land on spriteless planet, as it tries to calculate radius /distance from the sprite to start landing process, and constantly circles around a zero point.

Even though the unintended behaviour comes about as trying to use spriteless planets as a form of one-way wormholes, which probably isnt the ideal future way to make these, the behaviour itself is problematic, and I cannot think of any situation where you would want the game to land on a spriteless planet.

Added a conditional in get_planet() to return nullptr if there is no sprite on the object, get_planet() is checked for truthiness in all checks for landable planets in that system, so player and AI and everything else will just not consider it a planet at all, and it can function as a one-way wormhole exit point.

The map still draws a two-way link, but thats probably outside the scope of this little "fix"

## Testing Done
Edited map.txt to make a wormhole link, and on one part of it, deleted the sprite entry.

Appeared in that system, pressed L to land, got the message about no landable planets, AI ships didnt try and go towards it to land.

Tested with a random planet, removing its sprite, and it didnt show up as a landable planet. 

All other planets and landings seemed to work normally.

## Save File
N/A

### DIFF
--- a/source/StellarObject.cpp
+++ b/source/StellarObject.cpp
@@ -57,9 +57,10 @@ double StellarObject::Radius() const
 // If it is possible to land on this planet, this returns the Planet
 // objects that gives more information about it. Otherwise, this
 // function will just return nullptr.
+// Planets without sprites are not landable
 const Planet *StellarObject::GetPlanet() const
 {
-	return planet;
+	return HasSprite() ? planet : nullptr;
 }
 
 


source/StellarObject.cpp
```diff
@@ -57,9 +57,10 @@ double StellarObject::Radius() const
 // If it is possible to land on this planet, this returns the Planet
 // objects that gives more information about it. Otherwise, this
 // function will just return nullptr.
+// Planets without sprites are not landable
 const Planet *StellarObject::GetPlanet() const
 {
-	return planet;
+	return HasSprite() ? planet : nullptr;
```
tehhowch, line 63: Revert this.

If a ship is trying to land on a stellar object, that process should check for a sprite.

This method is a simple getter, and with this change _all_ of the code doesnt know that a spriteless StellarObject is attached to a planet.

tehhowch (changes requested):
This is the wrong place for such a check.

source/StellarObject.cpp
bobrobbow: The comments above this function do specifically note that the function returns a planet *only* if it is landable, else it returns nullptr, so the overall logic is unchanged, but I see your point about cluttering up a simple getter.

The alternative is a lot more checks in various other places, from what I can see, logic in player ship and AI ship logic for running away, refuelling etc, I tried in the Planet::canland() function, but I couldnt see a way to get the body.hassprite() from Planet class.

PR #5467, opened 2020-10-23
Add checks for AI so it does not try and land on Spriteless planets

bobrobbow:
**Bugfix:** This PR addresses issue #5061 

## Fix Details
As per #5466 being the wrong way to address this, have discussed and added relevant checks in AI.cpp instead
( and then one check in Engine.cpp so no labels surround the spriteless planet.

## Testing Done
I have gone into a system with a spriteless planet and tried to land on it, it was not visible, had no label, no AI tried to circle endlessly around it, the player autopilot didnt find any planet to land on.

Set up the use case of a one-way wormhole, exiting into a system using the spriteless planet, the wormhole worked fine.

I am less confident in this PR, because more changes were required than I originally intended. But seeing as the first one wasnt suitable I feel the obligation to do it properly! hopefully the changes to the AI are in the right place. 

But the testing didnt seem to throw up any oddities.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -267,7 +267,7 @@ namespace {
 			for(const StellarObject &object : from->Objects())
 			{
 				const Planet *planet = object.GetPlanet();
-				if(planet && planet->IsWormhole() && planet->IsAccessible(&ship)
+				if(object.HasSprite() && planet && planet->IsWormhole() && planet->IsAccessible(&ship)
 						&& planet->WormholeDestination(from) == to)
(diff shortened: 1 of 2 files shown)


Amazinite (changes requested):
I think line 240 should also have a sprite check to avoid a nonlandable object from being considered for refueling. 2315 could use one as well, although that's not critical, as that section just chooses objects for NPCs to fly over, not try to land on. Otherwise, looks good and seems to work.

PR #5475, opened 2020-10-27
Update CodeBlocks 64-bit paths to use modern defaults

Terin:
Currently, the CodeBlocks project file has three paths built into it: The precompiled libraries paths (which the wiki mentions you should extract them to that location), the 32-bit MinGW paths (which use a proper path, although for a version from years ago), then the 64-bit MinGW paths. These are not only inconsistent with the 32-bit paths, but completely nonsensical, as a user would only end up with them if they installed MinGW directly into C:\Program Files\ (which would leave other files lose in there, as well).

This PR updates the CodeBlocks project file to use the default install location from the latest version of MinGW (from 3-4 years ago), 8.1.0. It doesn't seem like a single developer working on ES uses the provided defaults, so changing them up will only help a user getting the game compiled with as little configuration as possible.

### DIFF
--- a/EndlessSky.cbp
+++ b/EndlessSky.cbp
@@ -66,7 +66,7 @@
 		<Linker>
 			<Add option="-Wl,--subsystem,windows" />
 			<Add option="-lwinmm" />
-			<Add library="C:\Program Files\mingw64\x86_64-w64-mingw32\lib\libmingw32.a" />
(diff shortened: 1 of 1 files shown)


MCOfficer:
Please do ignore the failing MacOS tests. Some screwup on homebrew's part.

PR #5483, opened 2020-11-02
Change Syndicate fleets near Korath space to focus on disable+board

Galaucus:
**Content (Revised fleet personality)**

## Summary
This changes the behavior of Syndicate security fleets in systems where Korath spawn. Rather than trying to destroy the Raiders, they will instead focus on disabling and boarding the intruders.

This should result in a closer match between gameplay behavior and the lore we see in the campaigns.

## Save File
Just head on over to Sheratan, Polaris, or Durax.

## PR Checklist
 - [x] I uploaded the necessary image, blend, and texture assets here: {{insert link to assets}}
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: {{insert PR link}}


Galaucus:
Related conversation:
https://discordapp.com/channels/251118043411775489/266345072554016768/772637635369238539

Anarchist2:
You defined the fleet definitions twice.

Zitchas:
If syndicate ships in the edge systems where Korath sometimes hunt are going to disable instead of destroy, how does this account for the lore that the syndicate has so few jump drives? 

- Did they only just recently switch tactics? 
- Are the syndicate fleets balanced such that they will often lose to a single Raider + two chasers? 
- Or (perhaps more realistically), perhaps the Korath fleets should always win fleet against fleet, but they sometimes lose when they end up getting swarmed by more than one fleet (and Syndicate reinforcements arrive faster than Korath ones)?
- Did the Korath only recently start raiding into Syndicate space? 
- Or are they more prone to self-destruction to prevent the Syndicate from getting their technology?
- I assume that the stuff that these appeasing Syndicate ships dump must be much less appealing than the stuff the Remnant drop, otherwise the Korath would definitely shift focus.

ravenshining:
Since the AI only takes cargo, this could leave too much high tech ripe for the player's picking.  We really need a personality that isn't just plunder, but plunder-then-kill, for this and other fleets.

Galaucus:
> Since the AI only takes cargo, this could leave too much high tech ripe for the player's picking. We really need a personality that isn't just plunder, but plunder-then-kill, for this and other fleets.

How about swapping the personality to vindictive? That would have them plunder then destroy.

data/human/fleets.txt
```diff
@@ -1637,6 +1637,103 @@ fleet "Large Syndicate"
 		"Vanguard (Missile)"
 		"Quicksilver (Proton)" 2
 
+fleet "Small Syndicate Core"
+	government "Syndicate"
+	names "syndicate small"
+	cargo 2
+	personality
+		appeasing disables plunders
```
Amazinite, line 1645: I believe that removing `disables` but keeping `plunder` results in the behavior "disables to plunder, then destroys after plundering," while having `disables` here is making it so that these fleets try to leave ships disabled. That to me seems like much more sensible behavior; these Syndicate fleets would be plundering what they can, then destroying the ship so no one else can get their hands on it. And in fact, it might make sense to just add `plunders` to the existing Syndicate fleets instead of adding new ones just for the Korath.

Amazinite (changes requested):

data/human/fleets.txt
Zitchas: Agreed. Plunder then destroy fits the Syndicate mentality much better than simply "disable and leave them".

Amazinite:
Test it out and tell me if `plunders` without `disables` has the desired behavior. If so, I'd say this is good to go, unless we want to further discuss changing all Syndicate fleets with this behavior instead of just those that come into contact with Korath fleets.

Galaucus:
Tested these personality combinations:
`appeasing disables plunders vindictive`
`appeasing disables plunders`
`appeasing plunders`
`plunders`
`heroic plunders`
`heroic plunders vindictive`

None produced the desired result of "disable enemy ship, board and loot, then destroy it". If the vindictive personality is set, the AI will proceed immediately to destroying disabled ships instead of plundering them. As far as I can tell the presence of "disables" doesn't change anything while "plunders" is also present.

Amazinite:
After some more thought (and discussion on stream today), I think we should just leave the Syndicate fleets as is, at least for now. The explanation as to why one sees Syndicate fleets destroying Raiders despite knowing that the Syndicate have boarded Raiders is that not everyone in the Syndicate knows about the Raiders, so as far as most of the Syndicate is concerned Raiders are just another hostile target to be eliminated. It's only a small section of the Syndicate that knows about the Raiders, what they are, and is involved in plundering them. 

Whatever the case, the issues with the plunders personality is something I've taken note of in #5540. Until that's done, we can't really get the behavior we want without making it much easier to plunder/capture ships in Syndicate space. As such, I think it makes sense to close this PR and perhaps revisit it if/when #5540 is resolved.

Issue #5462, opened 2020-10-19
Data race when loading audio on startup


ceeac:
**To Reproduce**
1. Compile with `CXX=g++-10 CXXFLAGS="-fsanitize=thread" LDFLAGS="-fsanitize=thread" scons mode=debug`
2. Start the game

**Expected behavior**
No data race

**Actual behavior**
```
==================
WARNING: ThreadSanitizer: data race (pid=23659)
  Read of size 8 at 0x5590f58fa768 by main thread (mutexes: write M2087):
    #0 std::_Rb_tree<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, Sound>, std::_Select1st<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, Sound> >, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, Sound> > >::size() const <null> (endless-sky+0x960e4)
    #1 std::map<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, Sound, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const, Sound> > >::size() const <null> (endless-sky+0x945ae)
    #2 Audio::GetProgress() source/Audio.cpp:192 (endless-sky+0x8cb12)
    #3 GameData::Progress() source/GameData.cpp:314 (endless-sky+0x1d1e2a)
    #4 MenuPanel::Step() source/MenuPanel.cpp:70 (endless-sky+0x2d2d9a)
    #5 UI::StepAll() source/UI.cpp:101 (endless-sky+0x4a4f24)
    #6 GameLoop(PlayerInfo&, Conversation const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, bool) source/main.cpp:263 (endless-sky+0x4b1003)
    #7 main source/main.cpp:128 (endless-sky+0x4b0370)

  Previous write of size 8 at 0x5590f58fa768 by thread T14:
[...]

PR #5486, opened 2020-11-03
Fix race condition while loading the audio.

Rakete1111:
**Bugfix:** This PR addresses issue #5462 

## Fix Details

I fixed this by creating the `sounds` entry inside the mutex lock instead of outside, so that loading the sound file and accessing `sounds`'s size do not clash.

## Testing Done

I couldn't get thread sanitizer to error out, but I verified using OP's error message that I correctly identified the two functions that race (`map::operator[]` and `map::size`).

However, I reduced the code into a small example to test my fix. If you remove the comment (my fix), then the race condition disappears. The example can be found here: https://godbolt.org/z/fP5r9s

## Save File
N/A

### DIFF
--- a/source/Audio.cpp
+++ b/source/Audio.cpp
@@ -576,6 +576,10 @@ namespace {
 					return;
 				name = loadQueue.begin()->first;
 				path = loadQueue.begin()->second;
+
+				// Since we need to unlock the mutex below, create the map entry to
+				// avoid a race condition when accessing sounds' size.
+				sounds[name];
 			}
 			
 			// Unlock the mutex for the time-intensive part of the loop.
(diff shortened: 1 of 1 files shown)


source/Audio.cpp
```diff
@@ -576,6 +576,10 @@ namespace {
 					return;
 				name = loadQueue.begin()->first;
 				path = loadQueue.begin()->second;
+
+				// Since we need to unlock the mutex below, create the map entry to
+				// avoid a race condition when accessing sounds' size.
+				sounds[name];
```
petervdmeer, line 582: It seems appropriate to also store the created/looked-up entry in a variable and use it on line 586 (line 582 in the original code) to avoid the need to accessing the map again.

Rakete1111: I guess. Didn't think it was necessary since I doubt that it will have an impact on performance (since it's already race free). thx :)

tehhowch (approved):
Thanks for the PR!

Issue #5479, opened 2020-10-30
Bug: A die effect jumps back when it's borne by a projectile


oo13:
**Describe the bug**
A die effect jumps back when it's borne by a projectile, that is, a projectile causes a die effect before the location where it dies.
Please see a video in the Screenshot section.

**To Reproduce**
1. Install [this plugin](https://github.com/endless-sky/endless-sky/files/5464783/test-die-effect-plugin.zip) and use [this savefile](https://github.com/endless-sky/endless-sky/files/5464781/die_effect_savefile.zip) "pom pom.txt".
2. Start "endless-sky" with "-d" option.
3. Load the pilot "pom pom".
4. Depart.
5. Press CapsLock and slow the game speed down.
6. Fire the gun.

**Expected behavior**
A projectile and the die effect should move continuously.

**Screenshot**
![Peek 2020-10-30 20-39](https://user-images.githubusercontent.com/22392249/97701079-35adfc00-1af0-11eb-8835-b114bce2f56d.gif)
A red circle is a projectile, a white is a die effect.

**System (please complete the following information):**
 - OS: Debian GNU/Linux Testing
 - Game Source: Github
 - Version:  current master 70eb5967f20dfd7cc4aaa697cc62d46575e24173

**Link to save file**
[die_effect_savefile.zip](https://github.com/endless-sky/endless-sky/files/5464781/die_effect_savefile.zip)
[test-die-effect-plugin.zip](https://github.com/endless-sky/endless-sky/files/5464783/test-die-effect-plugin.zip)

```
outfit "Pom Pom Gun"
	category "Guns"
	cost 100
	"mass" 5
	"outfit space" -5
	"weapon capacity" -5
	"gun ports" -1
	weapon
		sprite "projectile/red_circle"
		sound "blaster"
		"hit effect" "blaster impact"
		"die effect" "pom pom dead"
		"velocity" 60
		"lifetime" 5
		"reload" 5
		"firing energy" 1
		"firing heat" 1
		"shield damage" 1
		"hull damage" 1

effect "pom pom dead"
	sprite "projectile/white_circle"
	"lifetime" 60
```

PR #5489, opened 2020-11-03
Add an offset when drawing certain sprites

Rakete1111:
**Bugfix:** This PR addresses issue #5479 

## Fix Details
This fix implements the solution recommended in the TODO in BatchDrawList: Add a draw offset to sprites.

Under `"sprite"` this PR adds a `"draw offset" <num>`, which represents the offset of the sprite to draw. The offset is added in direction of the velocity of the projectile.

## Testing Done
I tested this PR with the save file on the linked issue. However, I didn't test any weapon except Beam Laser. I'll do the others soonish.

- [ ] Make sure that I didn't miss a weapon to adjust

## Save File

In the linked issue, the plugin and the save file.

### DIFF
--- a/data/coalition/coalition weapons.txt
+++ b/data/coalition/coalition weapons.txt
@@ -207,6 +207,7 @@ outfit "Heliarch Attractor"
 		sprite "projectile/attractor"
 			"frame rate" 5.2
 			"random start frame"
+			"draw offset" 320
 		"hardpoint sprite" "hardpoint/heliarch attractor"
 		"hardpoint offset" 9.
 		sound "heliarch attractor"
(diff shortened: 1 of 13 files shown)


oo13:
I tried this PR and it worked fine. But I afraid it introduces an incompatibility with existing plug-ins, although I expect we can notice a significant difference for only beam weapons.

Peek 2020-11-06 19-58
The die effect appears at in the same place where the projectile disappears. The effect seems to pause at the moment.

"draw offset" can solve the problem of the pause.
Peek 2020-11-06 20-07
It's the result that the die effect has "draw offset" 60 that is the projectile velocity.

tehhowch (commented):
I believe we can minimize the plugin compatibility hit by providing an automatic draw offsets for all outfits' effect/weapon sprites that don't have one, based on a custom heuristic. E.g. any sprite with a given aspect ratio for weapons (or submunitions) having particular lifetime/velocity characteristics would trigger the computation (and log a warning).

Amazinite:

PR #5493, opened 2020-11-06
Clean up borders of haze sprites

Terin:
image

The original haze sprites added by MZ had slight wisps on the border of the image, and was exacerbated by the images being made brighter. These result in small, but noticeable wisps that have hard cutoffs in space, as seen above. This PR removes these on the main haze sprites, while the rest of the haze sprites are unaffected, as they are more continuous blobs.


Amazinite:
Being honest, I don't see anything off in that image. It wasn't until I messed with some things in GIMP until I got something to show up.

The hard-edges of the haze.
image

Old and new.
image

So it is a fix, but I'm not even sure how many people will notice. Clearly some will if you did. :)

MCOfficer:
> Being honest, I don't see anything off in that image. It wasn't until I messed with some things in GIMP until I got something to show up.

I played around with my screen settings and noticed that this depends entirely on your gamma mode or gamma value (depending on what settings your monitor offers). Once you use a decently high gamma value, it becomes very obvious.

for reference, here's an exaggerated example with gamma 1.6:
image

Amazinite:
Yeah, I can just barely notice it there, but if I wasn't looking for it I'd probably just glance over it even with that level of gamma.

PR #5497, opened 2020-11-10
Added ability to scale galaxy sprites

FranchuFranchu:
**Feature:** This PR implements the ability to scale galaxy sprites using the "scale" attribute

## Feature Details
The main reason why I'm doing this is because I'm working on a plugin with a really large galaxy. I could make a 12800x128000 PNG/JPG image, but that turns out to be prohibitively memory-inefficient. So this PR allows scaling galaxy sprites.

## UI Screenshots
N/A

## Usage Examples
    
    galaxy "Milky Way"
         pos 0 0
         sprite ui/galaxy
         scale 4 # Will make the sprite 4x larger. May result in inevitable blurring. Default value is 1
         # scale 0.1 # Will make the sprite 10x smaller.

## Testing Done
Tested various scaling values in many situations where the map is displayed.

## Performance Impact
N/A


Darcman99:
Won't the galaxy image be blurred rather dramatically?

ravenshining:
Have you tried splitting your image up into 2048x2028 chunks?  I'm curious as to whether ES would handle that better.

tehhowch (changes requested):
The proposed changes need some small updates. As far as the goal of the proposal, I'm fine with it, but I don't see a wide applicability of the feature.

I believe there was a request for generalized image transformations, which could in theory provide similar functionality to all sprites if implemented. Such an approach would avoid needing to implement keyword support for a variety of different classes, where the desired transformation specification keywords may have more applicable meanings to the object with the sprite.

MCOfficer:
> I believe there was a request for generalized image transformations, which could in theory provide similar functionality to all sprites if implemented. Such an approach would avoid needing to implement keyword support for a variety of different classes, where the desired transformation specification keywords may have more applicable meanings to the object with the sprite.

Refs #4588

FranchuFranchu:
@tehhowch: I also think that maybe waiting for #4588 to be merged is a good idea



> Won't the galaxy image be blurred rather dramatically?

@Darcman99: Yes, it will be blurred if you scale it too much. However, if you choose low-contrast, low-brightness galaxy sprites the difference isn't as noticeably



> Have you tried splitting your image up into 2048x2028 chunks? I'm curious as to whether ES would handle that better.

@ravenshining: I thought of that, but having many large images still takes up too much space on the disk. And it's also impossible to edit.

Amazinite:
So how is this not just a less general form of #4588?

source/Galaxy.cpp
```diff
@@ -45,3 +47,8 @@ const Sprite *Galaxy::GetSprite() const
 {
 	return sprite;
 }
+
+float Galaxy::Scale() const
+{
+	return scale;
+}
```
tehhowch, line 54: Nit: Missing newline. Your text editor may not support EditorConfig definitions, as it should have ensured a newline at EOF.

tehhowch (approved):
Lgtm! If we proceed with sprite-level scaling we can deprecate setting scale as a token of a `galaxy` node, and then remove the functionality in 1.0

Amazinite:
I'll note that #4588 works by creating a `"scale"` attribute of Body. A Galaxy is not a Body, though, meaning that the scale keyword of #4588 doesn't work on galaxy images.

petervdmeer (approved):

quyykk:
Galaxy should probably be a Body so that it can be animated as well as scaled.

petervdmeer:
> Galaxy should probably be a Body so that it can be animated as well as scaled.

Ok. And I just noticed that the `scale` keyword from #4588 is placed on the sprite, not on the level where it is placed in this PR. We probably want to keep things consistent and also have the scale keyword on the sprite here in this PR.

petervdmeer (changes requested):
scale is placed on a different level than in #4588 .

warp-core:
Is this still desired given #6275? I think I prefer this to the option of only rendering a single galaxy at a time.

Amazinite:
We already have a class `Body` that is capable of scaling sprites, among other things. Instead of just adding a scale attribute to galaxies, I'd prefer if Galaxy extended Body or was refactors in some other way to include many of the sprite-modification mechanics that Body does.

PR #5498, opened 2020-11-11
Fix: Loss of trailing digits when adding outfit

oo13:
**Bugfix:** This PR addresses issue #3639 and #3645.

## Fix Details
This PR addresses an issue that is caused by an accumulation of a loss of trailing digits when adding an outfit.

#3645 could solve the issue by another way; this PR needs more calculation amount, but is more proper and removes a mysterious parameter EPS.

## Testing Done
1. Load a pilot Sell Issue.
2. Sell all Laser Rifles.
3. Sell all Grenades.
4. Try to sell all Nerve Gases.

Current master cannot sell a last Nerve Gas, but this PR can sell all Nerve Gases.

## Save File
Test file provided by @KiLEdEnNis (same as #3645): [sell_issue.txt](https://github.com/endless-sky/endless-sky/files/1722054/sell_issue.txt)

## Performance Impact
This PR affects a performance only when Adding/Removing outfits and the impact is not noticeable.

### DIFF
--- a/source/Outfit.cpp
+++ b/source/Outfit.cpp
@@ -21,12 +21,11 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include <algorithm>
 #include <cmath>
+#include <numeric>
 
 using namespace std;
 
 namespace {
-	const double EPS = 0.0000000001;
-	
 	// A mapping of attribute names to specifically-allowed minimum values. Based on the
 	// specific usage of the attribute, the allowed minimum value is chosen to avoid
 	// disallowed or undesirable behaviors (such as dividing by zero).
@@ -87,6 +86,31 @@ namespace {
 			if(thisMap[it.first] == 0)
 				thisMap.erase(it.first);
 		}
+	}
+	
+	// Summation that avoids a loss of trailing digits as possible.
+	double Sum(const map<double, int> &partialTerms)
+	{
+		vector<double> partialSums;
+		partialSums.reserve(partialTerms.size());
+		for(const auto &it : partialTerms)
+			partialSums.push_back(it.first * it.second);
+		sort(partialSums.begin(), partialSums.end(), [&](double a, double b){ return fabs(a) < fabs(b); });
+		return accumulate(partialSums.begin(), partialSums.end(), 0.);
+	}
+	
+	// Add value * count to an attribute.
+	// This function doesn't modify attribute if valud * count is zero.
+	void AddToAttribute(double &attribute, map<double, int> &partialTerms, double value, int count)
+	{
+		if(value == 0. || count == 0)
+			return;
+		
+		auto &countOfValue = partialTerms[value];
+		countOfValue += count;
+		if(!countOfValue)
+			partialTerms.erase(value);
+		attribute = Sum(partialTerms);
 	}
 }
 
@@ -216,6 +240,10 @@ void Outfit::Load(const DataNode &node)
 	};
 	convertScan("outfit");
 	convertScan("cargo");
+	
+	partialTerms["mass"][mass] = 1;
+	for(const auto &it : attributes)
+		partialTerms[it.first][it.second] = 1;
 }
 
 
@@ -305,9 +333,8 @@ int Outfit::CanAdd(const Outfit &other, int count) const
 				continue;
 		}
 		double value = Get(at.first);
-		// Allow for rounding errors:
-		if(value + at.second * count < minimum - EPS)
-			count = (value - minimum) / -at.second + EPS;
+		if(value + at.second * count < minimum)
+			count = (value - minimum) / -at.second;
 	}
 	
 	return count;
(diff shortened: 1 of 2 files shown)


source/Outfit.cpp
```diff
@@ -88,6 +87,31 @@ namespace {
 				thisMap.erase(it.first);
 		}
 	}
+	
+	// Summation that avoids a loss of trailing digits as possible.
```
tatami4, line 91: Shouldn't this be `// Summation that avoids loss of trailing digits as much as possible.`

source/Outfit.cpp
```diff
@@ -88,6 +87,31 @@ namespace {
...
+		vector<double> partialSums;
+		partialSums.reserve(partialTerms.size());
+		for(const auto &it : partialTerms)
+			partialSums.push_back(it.first * it.second);
+		sort(partialSums.begin(), partialSums.end(), [&](double a, double b){ return fabs(a) < fabs(b); });
+		return accumulate(partialSums.begin(), partialSums.end(), 0.);
+	}
+	
+	// Add value * count to an attribute.
+	// This function doesn't modify attribute if valud * count is zero.
```
tatami4, line 103: typo: `if valud * count`

oo13:
@tatami4 Thank you for suggestions. I posted the PR applied your comments.

source/Outfit.cpp
```diff
@@ -88,6 +87,31 @@ namespace {
...
 	}
+	
+	// Summation that avoids loss of trailing digits as much as possible.
+	double Sum(const map<double, int> &partialTerms)
+	{
+		vector<double> partialSums;
+		partialSums.reserve(partialTerms.size());
+		for(const auto &it : partialTerms)
+			partialSums.push_back(it.first * it.second);
+		sort(partialSums.begin(), partialSums.end(), [&](double a, double b){ return fabs(a) < fabs(b); });
```
petervdmeer, line 98: If we want to avoid loss of trailing digits as much as possible, then I would expect the smallest numbers to be accumulated first before the large numbers. Because the loss mostly happens when small numbers are added to large numbers.

source/Outfit.h
```diff
@@ -105,6 +105,7 @@ class Outfit : public Weapon {
 	std::vector<std::string> licenses;
 	
 	Dictionary attributes;
+	std::map<std::string, std::map<double, int>> partialTerms;
```
petervdmeer, line 108: The code here looks quite elegant, but please add a few comments here on how the partialTerms help to avoid the rounding errors (because I expect that it is a bit hard to follow for people that just encounter this code later out of the context of this PR).

source/Outfit.cpp
```diff
@@ -88,6 +87,31 @@ namespace {
...
+		return accumulate(partialSums.begin(), partialSums.end(), 0.);
+	}
+	
+	// Add value * count to an attribute.
+	// This function doesn't modify attribute if value * count is zero.
+	void AddToAttribute(double &attribute, map<double, int> &partialTerms, double value, int count)
```
petervdmeer, line 104: Do all additions and removals always use the proper `base value`, e.g. if we have situations where an attribute has `+0.5` and `2* -0.25`, then the magic added here wouldn't work well.

source/Outfit.cpp
```diff
@@ -88,6 +87,31 @@ namespace {
 				thisMap.erase(it.first);
 		}
 	}
+	
+	// Summation that avoids loss of trailing digits as much as possible.
+	double Sum(const map<double, int> &partialTerms)
```
petervdmeer, line 92: The memory usage for attributes increases quite a bit using this solution; we now get an entry in the partialTerms map for every fraction that was added to the attribute. This will happen for every ship, including ships where the outfits are typically not changed.
If we go this route, then there might be another solution that uses an equal amount of memory, but that computes faster and that is to use very high precision numbers instead of the partial terms.

Another consideration is that we only need this very high precision when we are actually changing ship outfits, and the raw data to generate the high precision already seems to be stored in the outfit-counts on a ship. We should be able to generate the high precision data on-the-fly when we need them during Outfitting (Thus reducing the overal memory usage at the costs of additional CPU time when we are starting to change Outfits.)

source/Outfit.cpp
oo13: I cannot remember it exactly, but I guess the only intention is to avoid infinitely increasing the size of "partialTerms."

oo13:
If you want to fix the root cause of the precision problem, I recommend that you design it with the fixed-point and implement it with the double precision. It is very easy and less modification than you probably expect.

petervdmeer:
> If you want to fix the root cause of the precision problem, I recommend that you design it with the fixed-point and implement it with the double precision. It is very easy and less modification than you probably expect.

Yes, doing the actual calculations in base-2 fixed-point would likely resolve the problem completely. (And keeping the doubles "at flight time" should result in zero/no performance impact.) Would you like to give that a try and (create a new) PR for) this?

oo13:
I have no plan of a PR.
You need to determine the dynamic ranges and the LSB positions for the maximum number of the outfits and all attributes before implementing it, anyway.

petervdmeer:
@oo13 Good to know! We will take your advise and this PR as inspiration to address the issue.

PR #5499, opened 2020-11-14
Added scrolling in the logbook category list

FranchuFranchu:
**Feature:** This PR implements scrolling in the list of categories in the logbook

## Feature Details
Right now, it is possible to scroll the list of dates and sub-categories in the logbook. However, it is not possible to scroll the list of years and categories ("Factions", "People", "3034"). This is an issue, because it means that on low screen resolutions, on large save files it is not possible to access the latest years of the logbook.

## UI Screenshots
image

## Usage Examples
No changes to data files are necessary

## Testing Done
I have a large save file which has ~29 years of logbook history. I opened that save's logbook and tried scrolling

## Performance Impact
N/A

### DIFF
--- a/source/LogbookPanel.cpp
+++ b/source/LogbookPanel.cpp
@@ -245,15 +245,10 @@ bool LogbookPanel::Click(int x, int y, int clicks)
 
 bool LogbookPanel::Drag(double dx, double dy)
 {
-	if ((hoverPoint.X() - Screen::Left()) > SIDEBAR_WIDTH)
-	{
+	if((hoverPoint.X() - Screen::Left()) > SIDEBAR_WIDTH)
 		scroll = max(0., min(maxScroll, scroll - dy));
-	}
 	else
-	{
-		// Scroll sidebar
 		categoryScroll = max(0., min(maxCategoryScroll, categoryScroll - dy));
-	}
 	
 	return true;
 }
@@ -270,7 +265,6 @@ bool LogbookPanel::Scroll(double dx, double dy)
 bool LogbookPanel::Hover(int x, int y)
 {
 	hoverPoint = Point(x, y);
-	
 	return true;
 }
 
--- a/source/LogbookPanel.cpp
+++ b/source/LogbookPanel.cpp
@@ -227,11 +227,10 @@ bool LogbookPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, 
 			int position = i * LINE_HEIGHT - categoryScroll;
 						
 			// If it's out of bounds, recenter it
-			if (position < MINIMUM_SELECTION_DISTANCE || position > (Screen::Height() - MINIMUM_SELECTION_DISTANCE))
+			if(position < MINIMUM_SELECTION_DISTANCE || position > (Screen::Height() - MINIMUM_SELECTION_DISTANCE))
 				categoryScroll = position - (Screen::Height() / 2);
 
 			categoryScroll = max(categoryScroll, 0.);
-
 		}
 	}
 	
--- a/source/LogbookPanel.cpp
+++ b/source/LogbookPanel.cpp
@@ -107,7 +107,7 @@ void LogbookPanel::Draw()
 	Point highlightOffset = Point(4. - PAD, 0.) + .5 * highlightSize;
 	Point textOffset(0., .5 * (LINE_HEIGHT - font.Height()));
 	// Start at this point on the screen:
-	Point pos = Screen::TopLeft() + Point(PAD, PAD);
+	Point pos = Screen::TopLeft() + Point(PAD, PAD - categoryScroll);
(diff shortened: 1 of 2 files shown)


source/LogbookPanel.h
```diff
@@ -60,9 +61,13 @@ class LogbookPanel : public Panel {
...
 	std::vector<std::string> contents;
 	std::vector<Date> dates;
+
+	Point hoverPoint;
 	
 	// Current scroll:
 	double scroll = 0.;
+	double categoryScroll = 0.; 
+	mutable double maxCategoryScroll = 0.;
 	mutable double maxScroll = 0.;
```
Amazinite, line 71: Just organizing these variables to match the organization of ShopPanel. (Scroll variables in some order then max scroll variables in that same order.)
```suggestion
	double categoryScroll = 0.; 
	double scroll = 0.;
	mutable double maxCategoryScroll = 0.;
	mutable double maxScroll = 0.;
```

source/LogbookPanel.cpp
```diff
@@ -243,7 +245,15 @@ bool LogbookPanel::Click(int x, int y, int clicks)
...
-	scroll = max(0., min(maxScroll, scroll - dy));
+	if ((hoverPoint.X() - Screen::Left()) > SIDEBAR_WIDTH)
+	{
+		scroll = max(0., min(maxScroll, scroll - dy));
+	}
+	else
+	{
+		// Scroll sidebar
+		categoryScroll = max(0., min(maxCategoryScroll, categoryScroll - dy));
+	}
```
Amazinite, line 256: [Style guide](https://endless-sky.github.io/styleguide/styleguide.xml):
> Single-line loop bodies: You do not need to use braces if a loop or conditional body is a single expression.
> Parentheses: Do not put a space before or after the parentheses for functions and control statements.
```suggestion
	if((hoverPoint.X() - Screen::Left()) > SIDEBAR_WIDTH)
		scroll = max(0., min(maxScroll, scroll - dy));
	else
		categoryScroll = max(0., min(maxCategoryScroll, categoryScroll - dy));
```

source/LogbookPanel.cpp
```diff
@@ -257,6 +267,15 @@ bool LogbookPanel::Scroll(double dx, double dy)
 
 
 
+bool LogbookPanel::Hover(int x, int y)
+{
+	hoverPoint = Point(x, y);
+
```
Amazinite, line 273: I'd just drop this empty line.
```suggestion
```

Amazinite (changes requested):
Something I've noticed is that in order to click to drag the category column, you also end up clicking on one of the categories. This causes the selected category to change to the one you incidentally clicked on in order to drag the column. If you had previously selected a year and click on any category below that year, it collapses the months of that category and causes the column to jump.

<details><summary>Large images</summary>

Selected 3014 with an entry in every month:
image

Clicked on 3025 in order to drag, collapsing 3014. My cursor is now on 3036 because of this.
image

</details>

This is actually also the case in the outfitter and shipyard where clicking on an outfit or ship to drag also selects the outfit or ship. This isn't really an issue now because clicking on an outfit or ship doesn't expand or collapse anything, and you can actually drag these panels without clicking on an outfit or ship given the empty space in the panel, but it is notable that this is the case with all panels that can scroll and also have things that can be selected. Given that, I wouldn't consider this issue blocking, but it is something we should track.

I think the sensible solution would be to only select something on the LMB release, but not select on LMB release if the panel had been dragged (perhaps dragged beyond a certain point so that tiny drags don't prevent you from selecting something). We don't need to deal with that now, though.

Amazinite:
Wait, found an issue. Using up and down arrow keys doesn't scroll the panel for you if you end up selecting a category past the screen. The outfitter/shipyard map has the behavior that I'd expect the logbook to have.

source/LogbookPanel.cpp
```diff
@@ -208,6 +214,24 @@ bool LogbookPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command,
...
+			// Check if it's too far down or up
+			int position = i * LINE_HEIGHT - categoryScroll;
+						
+			// If it's out of bounds, recenter it
+			if (position < MINIMUM_SELECTION_DISTANCE || position > (Screen::Height() - MINIMUM_SELECTION_DISTANCE))
+				categoryScroll = position - (Screen::Height() / 2);
+
+			categoryScroll = max(categoryScroll, 0.);
+
```
Amazinite, line 234:
```suggestion
			if(position < MINIMUM_SELECTION_DISTANCE || position > (Screen::Height() - MINIMUM_SELECTION_DISTANCE))
				categoryScroll = position - (Screen::Height() / 2);

			categoryScroll = max(categoryScroll, 0.);
```

Amazinite (approved):

PR #5512, opened 2020-11-21
Added ability for plugin creators to define multiple starts, and for players to choose between them.

FranchuFranchu:
**Feature:** This PR adds the ability for plugin creators to define multiple start scenarios, and for players to choose between them when creating a new pilot

## Feature Details
The default behaviour for start scenarios is that the last defined one overrides previous ones. This is fine for plugins that only want one scenario, but it forces plugins that want to provide multiple start dates on the same universe to split up. This fixes that, and lets the player choose the start scenario. It also warns the player if there are no start scenarios defined (which can be an installation issue).

## UI Screenshots
image
Image that appears when there is more than one start scenario available

## Usage Examples
```
start
	name "Other Scenario"
	description "	Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum. "
	sprite "scene/lobby"
	date 16 11 3013
	system "Sol"
	planet "Luna"
	account
		credits 480000
		score 400
		mortgage Mortgage
			principal 480000
			interest 0.004
			term 365
	set "license: Pilot's"
```

## Testing Done
I created a new pilot with the sample start shown above and played until the Pact Recon 2 mission. Everything seemed fine.

## Performance Impact
N/A


Zitchas:
Excellent! Being able to define multiple start points is *definitely* something that would be good to have.

source/StartConditionsPanel.h
```diff
@@ -0,0 +1,49 @@
...
+    virtual bool Scroll(double dx, double dy) override;
+    
+private:
+private:
+    PlayerInfo &player;
+    UI &gamePanels;
+    StartConditions *chosenStart;
+
+    // Stored here  that we can remove it if the player chooses a scenario
+    LoadPanel *loadPanel;
```
tehhowch, line 42: Always default pointer fields to `nullptr`, even if the class has no default constructor.

source/StartConditionsPanel.h
```diff
@@ -0,0 +1,49 @@
...
+    PlayerInfo &player;
+    UI &gamePanels;
+    StartConditions *chosenStart;
+
+    // Stored here  that we can remove it if the player chooses a scenario
+    LoadPanel *loadPanel; 
+    Point hoverPoint;
+
+    double listScroll;
+    double descriptionScroll;
```
tehhowch, line 46: These should also be initialized

source/StartConditionsPanel.h
```diff
@@ -0,0 +1,49 @@
...
+protected:
+    // Only override the ones you need; the default action is to return false.
+    virtual bool KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, bool isNewPress) override;
+    virtual bool Click(int x, int y, int clicks) override;
+    virtual bool Hover(int x, int y) override;
+    virtual bool Drag(double dx, double dy) override;
+    virtual bool Scroll(double dx, double dy) override;
+    
+private:
+private:
```
tehhowch, line 36: Two `private:` visibility specifiers; & there should be a double (& indented) newline separation between sections to match the rest of the game headers

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,253 @@
...
+    text->Wrap(descriptionText);
+
+    // Only allow draws inside the description box
+    glScissor(
+        -105 + GameWindow::TrueWidth() / 2,
+        -160 + GameWindow::TrueHeight() / 2,
+        210,
+        320);
+    glEnable(GL_SCISSOR_TEST);
+    text->Draw(Point(-105, -160+descriptionScroll), Color(1.,1.,1.));
```
tehhowch, line 119: Didn't we already draw the interface info? That's where all the bits about the chosen start should be handled, not in an imperative & hard-coded position.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,253 @@
...
+
+void StartConditionsPanel::Draw()
+{
+    glClear(GL_COLOR_BUFFER_BIT);
+
+    Information info;
+
+    string descriptionText; // String that will be shown in the description panel
+
+    if(chosenStart){
```
tehhowch, line 77: I'd very much prefer this to be an unconditional path, i.e. _always_ display the details of the selected start. When there isn't a user-selected start, the selected start is e.g. a customized start condition that has all the default strings & info.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,253 @@
...
+        210,
+        320);
+    glEnable(GL_SCISSOR_TEST);
+    text->Draw(Point(-105, -160+descriptionScroll), Color(1.,1.,1.));
+    glDisable(GL_SCISSOR_TEST);
+
+    Point point(-470., -157. + listScroll);
+    for(const auto &it : GameData::Start())
+    {
+        Rectangle zone(point + Point(110., 7.), Point(230., LIST_ITEM_SIZE));
```
tehhowch, line 125: Don't use an imperative draw code. The user should be able to supply their interface definition and sprite, and this panel's draw code needs to still work correctly. The rest of the game isn't quite at that point, but I don't want to add _any_ new panels that will add to the existing tech debt.

tehhowch (changes requested):
Thanks for the PR!

There's a fair number of style guidelines that need to be adhered to, and a lot of memory leaks that need to be addressed.

source/StartConditionsPanel.cpp
FranchuFranchu: The last start in the list is automatically selected in the constructor anyways. The only way chosenStart can be null is when there are no `start`s defined

source/StartConditionsPanel.cpp
FranchuFranchu: The reasons why I decided not to draw the text with interfaces was that the text needs to be translated and trimmed (which is what glScissor does) to let scrolling work properly. I don't think interfaces support these kinds of text operations.

Amazinite:
Starts can have varying starting locations, amounts of credits, mortgages, conditions, and even ships, as a start can grant a ship or multiple ships to the player. Would we want to display more of this information in the right panel when looking at a start or should that be handled in the description of the start? I imagine we could at least look at having the right panel show information about whatever mortgages you might have just below where the credits are shown. Perhaps a "credit score" and "outstanding loans" line?

Zitchas:
I think that'd be useful, but not required. As you say, it can all be included in the descriptions.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,253 @@
...
+    text->SetWrapWidth(210);
+    text->Wrap(descriptionText);
+
+    // Only allow draws inside the description box
+    glScissor(
+        -105 + GameWindow::TrueWidth() / 2,
+        -160 + GameWindow::TrueHeight() / 2,
+        210,
+        320);
+    glEnable(GL_SCISSOR_TEST);
```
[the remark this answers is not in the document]
FranchuFranchu: That's true, but it should in some places. PlanetPanel, for example, allows the description to overflow into space if it's too long.

glScissor is there since GL 2, so it shouldn't cause any issues with older hardware.

source/GameWindow.cpp
```diff
@@ -305,6 +305,27 @@ int GameWindow::Height()
 
 
 
+// These functions also work for fullscreen mode
+int GameWindow::TrueWidth()
```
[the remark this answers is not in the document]
FranchuFranchu: Width() doesn't work on fullscreen mode, so I made a function for getting the screen size on fullscreen (which I had to use in glScissor to convert panel coordinates to gl window coordinates). Now that I noticed that `Screen::Left` exists, I'll remove this function and use it instead.

source/StartConditions.cpp
```diff
@@ -32,13 +33,49 @@ void StartConditions::Load(const DataNode &node)
...
+
+
+void StartConditions::Save(DataWriter &out) const
+{
+	out.Write("start");
+	out.BeginChild();
+	{
+		out.Write("name", name);
+			
+		istringstream iss(description);
```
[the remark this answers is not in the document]
FranchuFranchu: Start condition descriptions are meant to be lengthy, so they're likely to have newlines in them.

source/StartConditions.h
```diff
@@ -16,6 +16,9 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Account.h"
 #include "ConditionSet.h"
 #include "Date.h"
+#include "Ship.h"
+#include "Sprite.h"
+#include "SpriteSet.h"
```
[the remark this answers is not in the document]
FranchuFranchu: Ship needs to be included because it's not a pointer in `std::list<Ship> ships;`. The other ones can be safely removed.

source/OutfitInfoDisplay.cpp
```diff
@@ -175,7 +175,7 @@ void OutfitInfoDisplay::UpdateRequirements(const Outfit &outfit, const PlayerInf
 		else
 		{
 			ostringstream out;
-			out << "sells for (" << (100 * sellValue) / cost << "%):";
+			out << "sells for(" << (100 * sellValue) / cost << "%):";
```
tehhowch, line 178: ?

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,275 @@
...
+	text->Wrap(descriptionText);
+
+	// Only allow draws inside the description box
+	glScissor(
+		descriptionBox.Left() + GameWindow::TrueWidth() / 2,
+		descriptionBox.Top() + GameWindow::TrueHeight() / 2,
+		descriptionBox.Width(),
+		descriptionBox.Height());
+
+	glEnable(GL_SCISSOR_TEST);
```
tehhowch, line 116: Remove use of glScissor _entirely_. I disagree that plugin authors should be writing novels to describe a custom start. As with planet descriptions and plugins' "about" text, it's the content author's _job_ to make sure their content looks good.

In a future PR, I'm open to refactoring how we get the game to draw text within specific user-definable-at-runtime bounds -- e.g. it would make a lot more sense for glScissor to be invoked directly by `WrappedText::Draw(const Rectangle &bounds, const Color &color)` or e.g. `Interface::TextBoxElement::Draw`

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,275 @@
...
+	gamePanels.CanSave(true);
+	gamePanels.Push(new MainPanel(player));
+	// Tell the main panel to re-draw itself (and pop up the planet panel).
+	gamePanels.StepAll();
+	// If the starting conditions don't specify any ships, let the player buy one.
+	if(player.Ships().empty())
+	{
+		gamePanels.Push(new ShipyardPanel(player));
+		gamePanels.StepAll();
+	}
```
tehhowch, line 266: This specific code is now repeated in 3 places. We should avoid that. I am OK with that being done in a separate PR, though.

tehhowch (changes requested):
Still a number of style issues to be fixed up, in addition to code changes and a memory leak to fix.

source/Account.cpp
```diff
@@ -402,3 +402,13 @@ int64_t Account::YearlyRevenue() const
 	// played for long enough to accumulate a full income history.
 	return ((history.back() - history.front()) * 365) / HISTORY;
 }
+
+
+bool Account::operator==(Account other) const
```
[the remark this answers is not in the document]
FranchuFranchu: 1. We need it to compare a start condition to another, in order to highlight the selected one in StartConditionsPanel.cpp
2. It shouldn't, I think passing an `Account&` instead should fix it

source/StartConditions.cpp
```diff
@@ -91,3 +129,32 @@ const list<Ship> &StartConditions::Ships() const
...
+{
+	return description;
+}
+
+const Sprite *StartConditions::GetSprite() const 
+{
+	return sprite;	
+}
+
+bool StartConditions::operator==(StartConditions other) const
```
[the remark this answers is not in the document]
FranchuFranchu: 1. It's needed to test whether the selected start conditions is equal to the displayed start conditions, in order to highlight them on StartConditionsPanel.cpp::136 (there might be a better way to do this)
2. It shouldn't copy it, I'll change the argument to a `StartConditions&`

source/GameData.cpp
```diff
@@ -974,7 +975,11 @@ void GameData::LoadFile(const string &path, bool debugMode)
 		else if(key == "shipyard" && node.Size() >= 2)
 			shipSales.Get(node.Token(1))->Load(node, ships);
 		else if(key == "start")
-			startConditions.Load(node);
+		{
+			StartConditions thisConditions = StartConditions();
+			thisConditions.Load(node);
+			startConditions.emplace_back(thisConditions);
+		}
```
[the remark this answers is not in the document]
FranchuFranchu: This won't work because `StartConditions` has no constructor with a `DataNode&`. Do you think I should add one @tehhowch ?

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,275 @@
...
+StartConditionsPanel::StartConditionsPanel(PlayerInfo &player, UI &gamePanels, LoadPanel *loadPanel)
+	: player(player), gamePanels(gamePanels), loadPanel(loadPanel)
+{
+	if(GameData::Start().size())
+	{
+		chosenStart = GameData::Start()[GameData::Start().size()-1];	
+		hasChosenStart = true;
+	}
+	
+	descriptionBox = GameData::Interfaces().Get("start conditions menu")->GetBox("start description");
```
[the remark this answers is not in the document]
FranchuFranchu: GetBox returns a (0, 0, 0, 0) box when it's not found, so it doesn't matter (since it will make the box not display in the panel, which is what we want).

FranchuFranchu:
There you go, @tehhowch ! I've changed most of the things you've suggested, including removing the glScissor call there. However, I overloaded a WrappedText::Draw function that uses glScissor to limit the bounds, which is what you said we could include in a future PR. I hope you don't mind it being inserted in this one. Eventually we can do this in PlanetPanel and other panels that have issues with overflowing text.

source/SavedGame.h
```diff
@@ -40,6 +42,8 @@ class SavedGame {
 	
 	const std::string &GetSystem() const;
 	const std::string &GetPlanet() const;
+	
+	const StartConditions &GetChosenStart() const;
```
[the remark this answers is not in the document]
FranchuFranchu: Not now, but maybe in the future someone will need it.

source/text/WrappedText.h
```diff
@@ -71,6 +72,9 @@ class WrappedText {
 	// Draw the text.
 	void Draw(const Point &topLeft, const Color &color) const;
 	
+	// Additionally, this function also limits the area where it has to be drawn
+	void Draw(const Point &topLeft, const Rectangle &bounds, const Color &color) const;
+
```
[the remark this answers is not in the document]
FranchuFranchu: @tehhowch Is it okay if I just let the text overflow like in PlanetPanel? After #4123 gets merged, I can make another PR which prevents the text from overflowing.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,280 @@
...
+#include <SDL2/SDL.h>
+
+#include "gl_header.h"
+
+using namespace std;
+
+StartConditionsPanel::StartConditionsPanel(PlayerInfo &player, UI &gamePanels, LoadPanel *loadPanel)
+	: player(player), gamePanels(gamePanels), loadPanel(loadPanel)
+{
+	if(!GameData::Start().empty())
```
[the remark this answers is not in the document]
FranchuFranchu: @tehhowch I personally think that it isn't necessary, but if you prefer I'll change it.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+	}
+	const Interface *startConditionsMenu = GameData::Interfaces().Find("start conditions menu");
+	
+	if(startConditionsMenu)	
+	{
+		descriptionBox =   startConditionsMenu->GetBox("start description");
+		entryBox =         startConditionsMenu->GetBox("start entry");
+		entryListBox =     startConditionsMenu->GetBox("start entry list");
+		entryInternalBox = startConditionsMenu->GetBox("start entry internal");	
+	}
```
tehhowch, line 67: Generally the idea of the Interface class is to pass data to it via `Information` and let it render & create buttons, rather than just use it to extract positioning anchors. I'm guessing we need additional functionality in the interface elements for that to be a valid approach for the scrollable list. If not, use that pattern here, and if so, please comment this usage with the "ideal" API that would be exposed.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+	for(const StartConditions &it : GameData::Start())
+	{
+		// emplace_back will implicitly instantiate a ClickZone
+		startConditionsClickZones.emplace_back(
+			Rectangle::FromCorner(
+				Point(entryListBox.Left(), entryListBox.Top() + i * entryBox.Height()),
+				entryBox.Dimensions()
+			),
+			it);
+		i++;
```
tehhowch, line 80: Inconsistent styling here ("flag" and postfix increment). There are other occurrences in this file that should also be addressed.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+		// This looks inefficient but it probably gets optimized by the compiler
+		listScroll -= dy;
+		listScroll = min(entryBox.Height() * (GameData::Start().size()-1), listScroll); // Avoid people going too low
+		listScroll = max(0., listScroll); // Snap the list to avoid people scrolling too far up
+	}
+	else
+	{
+		// Scroll the description
+		descriptionScroll += dy;
+		descriptionScroll = min(0., descriptionScroll); // Snap the list to avoid people scrolling too far up
```
tehhowch, line 191: Inline / non-sentence comments don't meet the style guide. It's generally sufficient to annotate the block with the intent, rather than the play-by-play. If the code itself is not clear, then more verbose comments are warranted.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+}
+
+
+
+bool StartConditionsPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, bool isNewPress)
+{
+	if(key == 'b' || key == SDLK_ESCAPE || command.Has(Command::MENU) || (key == 'w' && (mod & (KMOD_CTRL | KMOD_GUI))))
+		GetUI()->Pop(this);
+	else if(key == 's' || key == 'n' || key == '\n')
+	{
```
tehhowch, line 219: Should up/down keys change the selected start? Probably also Page Up/Down

source/StartConditionsPanel.h
```diff
@@ -0,0 +1,68 @@
...
+	
+	
+private:
+	PlayerInfo &player;
+	UI &gamePanels;
+	
+	WrappedText descriptionWrappedText;
+	
+	// Workaround around the absence of std::optional<>
+	StartConditions chosenStart;
```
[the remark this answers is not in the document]
FranchuFranchu: Should I also turn StartConditions into a pointer in other classes such as PlayerInfo?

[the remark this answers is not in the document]
tehhowch: Yeah, a new player should copy the start summary, and loading a save should read that in. It should be separate from those in GameData since the player might remove the source plugin, or it may get changed after the player begins their game.

This does mean we won't be able to update existing player's start data after they begin (e.g. setting some conditions based on future changes) but also means players don't lose start dates, etc. just because their "upstream" start condition was modified

source/StartConditionsPanel.cpp
FranchuFranchu: What do you mean by "flag" and postfix increment? The postfix increment and the `i` variables are there to move the clickzones down

Amazinite (commented):
Just taking a glance at this right now. Not looking at the code, just commenting on how it looks in game.

I mentioned already how there seems to be spare space in the UI for more information about each start. While of course each start will likely have unique differences that should be described in the, well, description, I think that information that all starts will have in common should be listed on the right panel.

Right now we have an image, the start name, the planet and system, the starting credits, and the date. I think you should add a "debt" line underneath "credits," as the debt you start with can be different from your starting credits. I'd also like a "credit score" line, but that doesn't look like it'll fit, and just "score" isn't descriptive enough, and "credit" isn't distinct enough from "credits," unless we rename "credits" to something else, which I'm not fond of.

The current image seems to be a tight fit. I'd suggest shrinking it a bit to give some space on the right and left side of the image from the edge of the UI.
image


When on the screen to select your start, the arrow keys don't currently cycle you through the existing start conditions. I'd expect them to do so, just as the arrow keys change what save you're looking at on the load panel.

How about renaming `start.txt` to `starts.txt` now that we can have more than one?

And how about having the New Pilot button bring up the starts panel even if only one start is defined? That way we can write out a proper description for the existing start to read even if there is only the single start right now. (Could include information from the [player's manual](https://github.com/endless-sky/endless-sky/wiki/PlayersManual) that's currently only seen there.) And in the future if we define more starts but don't have them accessible immediately (#5517), then I imagine the start panel would still show up with only a single choosable start.

source/StartConditions.cpp
```diff
@@ -32,13 +41,57 @@ void StartConditions::Load(const DataNode &node)
...
+			description += child.Token(1) + "\n";
+		else if(child.Token(0) == "sprite" && child.Size() >= 2)
+			sprite = SpriteSet::Get(child.Token(1));
 		else if(child.Token(0) == "account")
 			accounts.Load(child);
 		else if(child.Token(0) == "ship" && child.Size() >= 2)
 			ships.emplace_back(child);
+		else if(child.Token(0) == "conversation" && child.HasChildren())
+			conversation.Load(child);
+		else if(child.Token(0) == "conversation" && child.Size() >= 1)
```
Amazinite, line 56:
```suggestion
		else if(child.Token(0) == "conversation" && child.Size() >= 2)
```

source/StartConditions.cpp
```diff
@@ -91,3 +152,37 @@ const list<Ship> &StartConditions::Ships() const
...
+
+
+const Sprite *StartConditions::GetSprite() const 
+{
+	return sprite;	
+}
+
+
+
+bool StartConditions::operator==(const StartConditions &other) const
```
Amazinite, line 179: Why not define starts like we defined missions? That is,
```
start "<name>"
	...
```
That way starts could be identified by a unique name instead of needing to do this. Doing that would also allow plugins to alter existing starts by redefining them, instead of needing to add a new start.

source/StartConditions.cpp
Amazinite, line 56:
```suggestion
		else if(child.Token(0) == "conversation" && child.Size() >= 2)
```

source/StartConditions.cpp
```diff
@@ -32,13 +41,57 @@ void StartConditions::Load(const DataNode &node)
...
+			conversation.Load(child);
+		else if(child.Token(0) == "conversation" && child.Size() >= 1)
+			stockConversation = GameData::Conversations().Get(child.Token(1));
 		else
 			conditions.Add(child);
 	}
+	if(description.empty())
+	{
+		description = "No description provided";
+	}
```
Amazinite, line 64:
```suggestion
		description = "No description provided.";
```

Amazinite (changes requested):

tehhowch (commented):
> Why not define starts like we define missions? That way starts could be identified by a unique name instead of needing to use `operator==`. Doing that would also allow plugins to alter existing starts by redefining them, instead of needing to add a new start.


I'm ambivalent on this. I think it's good that the current approach enables guaranteed-unique starts for each plugin, with no conflicts due to name collision. But, it does mean that we're doing things a little differently than other customization points and a full declaration of a `start` is needed if a plugin wants to modify any piece of the game's current starting scenario, and that the user must willingly opt-in to the plugin's starting conditions.

A large argument against this, though, is that requiring a named `start` node is a compatibility break -- every single plugin out there, if they're following the ES "spec," is not in conformance with `start <name>`. Changing our strategy here means a plugin written for 0.9.12 will not work as intended without additional updates from the plugin author, unless we intentionally allow a StartCondition with a named key of `""` in our mapping of names to object, where `""` has always previously indicated bad data.

And of course, the current PR is also a compatibility break -- any plugins that intentionally override just a small part of the game's default start will now incur C++ code defaults instead of the shipped `start`'s defaults. For example, a plugin whose `start` node is just `set "my plugin is awesome"` will now be starting them with basically nothing but a credit score of 400.

I think we can propose a handling scheme that will support this "extend an existing start with name `<name>` or create it if not given" use case, but we don't need to add it in this PR. Basically, my thought is that we would handle `start` as an always-new start condition, and `start <name>` as a request to create it if not present, and edit it if present. Plugins are already required to have loaded after the content they wish to modify, so the target start condition will likely have been loaded already. Note that this also has a drawback if we ever allow customizing the load order, as it will break that implicit ordering dependence unless we first/also introduce a way to explicitly state an ordering dependence.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,340 @@
...
+		info.SetString("system", chosenStart->GetSystem()->Name());
+		info.SetString("date", chosenStart->GetDate().ToString());
+		info.SetString("credits", Format::Credits(chosenStart->GetAccounts().Credits()));
+
+		descriptionText = chosenStart->GetDescription();
+	}
+	else if(!GameData::StartOptions().size())
+	{
+		descriptionText = "No start scenarios were defined!\n\nMake sure that you installed Endless Sky and all of your plugins properly";
+	}
```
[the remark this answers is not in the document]
FranchuFranchu: Not really, deleting it on the next commit.

FranchuFranchu:
> How about renaming start.txt to starts.txt now that we can have more than one?

I think we should do this on the commit that adds #1519 (since right now, no other starts are added in that file)

> And how about having the New Pilot button bring up the starts panel even if only one start is defined?

The reason why I chose to make it skip the start panel is to cause the least amount of interference to players who don't have any other plugins installed (because I originally planned this to be for plugin developers). However, since by 0.9.13 we're going to have more than 1 scenario in the vanilla game, then a situation where there is only one start scenario defined will never happen. 

If a total replacement plugin decides to only have 1 start scenario and removes the rest, then I would expect them to not need any context specific to that start scenario, because there would be no other start scenarios to compare it to.

Amazinite:
> If a total replacement plugin decides to only have 1 start scenario and removes the rest, then I would expect them to not need any context specific to that start scenario, because there would be no other start scenarios to compare it to.

It's not a matter of comparing the details of start scenarios, it's a matter of providing details that aren't provided right now. The current start drops you off in the bank already getting your loan. It says nothing of your time in the textile mills mentioned in the [player's manual](https://github.com/endless-sky/endless-sky/wiki/PlayersManual). Having an extra description to provide for your started scenario allows us to say more than we otherwise would be able to without breaking the flow of the intro.

There's also the selfish reason that I sometimes accidentally click the new pilot button, which shoves you into the start of a new pilot without any confirmation, and you need to finish the intro conversation before returning to the main menu. Going to the starts panel regardless of the number of starts basically works as a confirmation dialog about creating a new pilot.

> I'll do this on the next commit. We could log any invalid start scenarios to errors.txt, and list the rest. If there are no valid start scenarios, then exit with an error.

I wouldn't exit from the game if there are no valid starts, as there may be an existing pilot that can be played without needing a valid start. "No valid starts" is basically a worst scenario from a total conversion plugin having an invalid start.

FranchuFranchu:
There it is, @Amazinite and @tehhowch. Now it is possible to name start scenarios, and then partially override them later, like with systems. Unnamed start scenarios are anonymous and can't be overridden.

Zitchas:
This is sounding really good. I was just about to ask if you could update & fix the merge conflicts so I could test, but it looks like you just took care of that. ;) Looking forward to trying it out.

Zitchas (approved):
I can't speak to the code - only that I can test this and it works. I used the example alternate start to create an alternate start that placed me in on Viminal with a Remnant basic license and an alternate set of funds and mortgage; and it all worked. 

Question: Using "set" can I set any condition (or completed missions, etcs) that I could normally set in missions?

Can I use the ship giving commands as well? (I'm going to go try it out, so I'll find out fairly quickly, but a formal answer as to whether or not it will would be appreciated.)

Anyway, thanks for all your work on this! Definitely appreciate and like this functionality.

Probably one of the first things I'll be doing is creating some 'testing' pilots that start me off in various places to facilitate testing Remnant content. Such as starting in Tania with a stock medium warship with a keystone; or starting on Viminal at various points in the Remnant campaign.

FranchuFranchu:
@tehhowch 

> Question: Using "set" can I set any condition (or completed missions, etcs) that I could normally set in missions?

Yup, looks like it. I haven't touched that part of the code, but it seems like when there's an unrecognized statement is in a start condition, it is interpreted as a statement for the ConditionSet.

source/GameData.cpp
```diff
@@ -218,6 +221,13 @@ bool GameData::BeginLoad(const char * const *argv)
 			LoadFile(path, debugMode);
 	}
 	
+	// Add the named start conditions to the start conditions vector
+	startConditions.reserve(namedStartConditions.size() + startConditions.size());
+	for(auto &it : namedStartConditions)
+	{
+		startConditions.push_back(it.second);
+	}
```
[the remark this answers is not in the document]
FranchuFranchu: @tehhowch  What do you mean with `startConditions.insert`? `namedStartConditions` is a Set, so its `begin()` and `end()` methods return an iterator over a `pair<string, StartConditions>`, which means I can't insert it directly on a `vector<StartConditions>`

source/GameData.cpp
```diff
@@ -1068,8 +1080,30 @@ void GameData::LoadFile(const string &path, bool debugMode)
...
-		else if(key == "start")
-			startConditions.Load(node);
+		else if(key == "start" && node.Size() >= 2)
+		{
+			// This copies the StartConditions. We're avoiding using the pointer
+			// returned by Get in case the newly updated start conditions turn out to be invalid
+			StartConditions thisStart = (*namedStartConditions.Get(node.Token(1)));
+			thisStart.Load(node);
+			
+			if (!thisStart.Valid())
```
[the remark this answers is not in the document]
FranchuFranchu: @tehhowch How should I do `node.PrintTrace` (in case the start conditions are invalid) after the node and its parents go out of scope?

source/StartConditions.cpp
```diff
@@ -63,27 +86,85 @@ void StartConditions::FinishLoading()
...
 }
 
 
 
-const Planet &StartConditions::GetPlanet() const
+const Planet *StartConditions::GetPlanet() const
 {
-	return planet ? *planet : *GameData::Planets().Get("New Boston");
+	return planet ? planet : GameData::Planets().Get("New Boston");
 }
```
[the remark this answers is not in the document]
FranchuFranchu: Seems like an absent-minded change while merging from upstream.

source/StartConditions.cpp
```diff
@@ -13,15 +13,24 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+StartConditions::StartConditions(const DataNode &node)
+{
+	Load(node);
+}
+
+
+
 void StartConditions::Load(const DataNode &node)
 {
 	for(const DataNode &child : node)
```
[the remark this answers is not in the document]
FranchuFranchu: @tehhowch I think we shouldn't do this because it would cause inconsistencies between named and anonymous start scenarios, in the former the `name` statement wouldn't be necessary and in the latter one it would. Unless you mean after loading, in case `name` was never set?

source/GameData.cpp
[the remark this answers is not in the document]
Amazinite: > How should I do `node.PrintTrace` (in case the start conditions are invalid) after the node and its parents go out of scope?

You don't. As mentioned, we can't actually check if something is valid until everything is loaded. Otherwise you might have a scenario where a start references a system or planet that hasn't been defined yet, erroneously marking the start as invalid.

Looking to #3462, not everything that is invalid necessarily gets printed to the errors file. In our case here, we only want to use the IsValid function to exclude showing those starts which aren't valid, not print out a list of all invalid starts.

source/StartConditions.cpp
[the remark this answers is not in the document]
Amazinite: Think of it like missions.
```
mission <name>
	name <name>
```
Missions have a unique name and a display name. If no display name is given, the unique name is displayed in game. So while having a separate display name for a start that is already named isn't necessary, that doesn't mean it should be disallowed. One might want to uniquely name starts separately from the display names. (And I am such one.)

source/GameData.cpp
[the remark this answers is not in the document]
FranchuFranchu: @Amazinite I've decided to log the error using Files::LogError (which doesn't need us to store the DataNode) instead not printing anything, because I think it would be useful for a plugin developer in case they don't realize what's wrong with the scenario they made. Files::LogError is non-critical, so logging it won't do any harm.

source/StartConditions.cpp
```diff
@@ -47,9 +62,22 @@ void StartConditions::Load(const DataNode &node)
...
 		else
 			conditions.Add(child);
 	}
+	if(name.empty())
+	{
+		if (node.Size() >= 2)
+			name = node.Token(1);
+		else
+			name = "Unnamed start";
+	}
```
Amazinite, line 78: This should be done before parsing the children. The unique name is overwritten by the base name, not the other way around.

source/StartConditions.h
```diff
@@ -15,32 +15,53 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
 	void FinishLoading();
+	// Serialize the basic information of this start
+	void Save(DataWriter &out) const;
 	
 	Date GetDate() const;
-	// Get this start's planet, or New Boston if not set.
+	
 	const Planet &GetPlanet() const;
-	// Get this start's system, or Rutilicus if not set.
 	const System &GetSystem() const;
```
Amazinite, line 47: These comments are technically still true. Might also be worth putting a comment above GetDate: `// Get this start's date, or 16/11/3013 if not set.`

source/StartConditions.h
```diff
@@ -15,32 +15,53 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
+	
+	const std::string &GetName() const;
+	const std::string &GetDescription() const;
+	
 	const Account &GetAccounts() const;
 	const ConditionSet &GetConditions() const;
 	const std::list<Ship> &Ships() const;
 	
+	// Check whether a start scenario is valid
+	// A valid start scenario has a name, a date, a system and a planet
```
Amazinite, line 63: Would no longer be true with my suggestion.
```suggestion
	// A valid start scenario has a valid system, planet, and conversation.
```

Amazinite (changes requested):

source/StartConditions.cpp
```diff
@@ -106,3 +177,19 @@ const list<Ship> &StartConditions::Ships() const
...
+bool StartConditions::IsValid() const
+{
+	if(system && (!planet || !system->IsValid()))
+		return false;
+	
+	if(planet && (!system || !planet->IsValid() || planet->GetSystem() != system))
+		return false;
+	
+	if(!stockConversation && conversation.IsEmpty())
+		return false;
```
tehhowch, line 192: The system & planet checks are overly complex. Note that you'll need to merge in latest `master` for `Conversation::IsValidIntro` to compile.
```suggestion
	// A start must specify a valid system.
	if(!system || !system->IsValid())
		return false;
	
	// A start must specify a valid planet in its specified system.
	if(!planet || !planet->IsValid() || planet->GetSystem() != system)
		return false;
	
	// A start must reference a valid "intro" conversation, either stock or custom.
	if((stockConversation && !stockConversation->IsValidIntro()) || (!stockConversation && !conversation.IsValidIntro()))
		return false;
```

source/StartConditions.h
```diff
@@ -49,6 +73,16 @@ class StartConditions {
 	Account accounts;
 	ConditionSet conditions;
 	std::list<Ship> ships;
+	std::string name = "Unnamed start";
```
tehhowch, line 76:
```suggestion
	std::string name;
```

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+			chosenStartIterator = GameData::StartOptions().begin();
+		else if(GameData::StartOptions().end() - chosenStartIterator < offset)
+			chosenStartIterator = GameData::StartOptions().end() - 1;
+		else
+			chosenStartIterator += offset;
+		
+		chosenStart = &*chosenStartIterator;
+		descriptionWrappedText.Wrap(chosenStart->GetDescription());
+		
+		return false;
```
tehhowch, line 215: Returning `false` from `Panel::KeyDown` (and other event handlers) means that the UI should pass that event to the next Panel in the stack, if the current panel doesn't trap all events. Please re-evaluate all the added event handler methods with this behavior of true/false return in mind.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+			if(chosenStart != &*chosenStartIterator)
+				descriptionScroll = 0;
+			chosenStart = &*chosenStartIterator;
+			descriptionWrappedText.Wrap(chosenStart->GetDescription());
+			break;
+		}
+	}
+	
+	if(chosenStart)
+		descriptionWrappedText.Wrap(chosenStart->GetDescription());
```
tehhowch, line 265: We're wrapping the new description when the selected item changes, and then unconditionally wrapping it again here -- I expect only one of them is needed (probably the conditional one).

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+		
+		listScroll += offset * entryBox.Height();
+		listScroll = min(max(listScroll, 0.), (GameData::StartOptions().size() - 1) * entryBox.Height());
+		
+		if(GameData::StartOptions().begin() - chosenStartIterator < offset)
+			chosenStartIterator = GameData::StartOptions().begin();
+		else if(GameData::StartOptions().end() - chosenStartIterator < offset)
+			chosenStartIterator = GameData::StartOptions().end() - 1;
+		else
+			chosenStartIterator += offset;
```
tehhowch, line 210: Note that `int` is not the guaranteed-correct type for the difference of two iterators.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,292 @@
...
+	if(key == 'b' || key == SDLK_ESCAPE || command.Has(Command::MENU) || (key == 'w' && (mod & (KMOD_CTRL | KMOD_GUI))))
+		GetUI()->Pop(this);
+	else if(key == SDLK_UP || key == SDLK_DOWN || key == SDLK_PAGEUP || key == SDLK_PAGEDOWN) 
+	{
+		int offset = 0;
+		offset += (key == SDLK_DOWN) - (key == SDLK_UP);
+		offset += PAGE_INTERVAL * ((key == SDLK_PAGEDOWN) - (key == SDLK_PAGEUP));
+		
+		listScroll += offset * entryBox.Height();
+		listScroll = min(max(listScroll, 0.), (GameData::StartOptions().size() - 1) * entryBox.Height());
```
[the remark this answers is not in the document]
FranchuFranchu: Which panel does that @tehhowch?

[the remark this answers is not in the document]
FranchuFranchu: I think 556cebb does what you want.

data/start.txt
```diff
@@ -8,7 +8,10 @@
 # WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
 # PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
-start
+start "default"
+	name "Endless Sky"
+	description "The default scenario for Endless Sky"
```
Amazinite, line 13:
```suggestion
	description `You grew up on New Boston, an uninteresting world in the Dirt Belt. You've dreamed of owning a starship ever since your first visit to the spaceport at the age of fifteen. After many long years of working at a textile mill, you've finally saved enough credits to apply for a pilot's license from the Republic.`
	description `    After receiving the license and with freedom from this planet at your fingertips, you travel to the bank to take out a loan on your first ship...`
```

Terin:
The scrolling behavior on StartConditionsPanel is inconsistent with something like LoadPanel. The entries will immediately vanish upon scrolling off, rather than fade away like they do on LoadPanel. You can also scroll when there is room for all the entries on the panel, so when I was clicking between them, I accidentally dragged for a split second, so the top entry disappeared.

I'm not sure if it's a problem with my data entries, but I added Syndicate and Paradise Worlds starts to my start.txt, and it defaults to the last entry with the Syndicate, which also scrolls the other entries up, making it look like there's only one start.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,296 @@
...
+		chosenStartIterator = GameData::StartOptions().end() - 1;
+		listScroll = (chosenStartIterator - GameData::StartOptions().begin()) * entryBox.Height();
+	}
+	
+	const Rectangle firstRectangle = Rectangle::FromCorner(entryListBox.TopLeft(), entryBox.Dimensions());
+	startConditionsClickZones.reserve(GameData::StartOptions().size());
+	for(size_t i = 0; i < GameData::StartOptions().size(); ++i)
+		startConditionsClickZones.emplace_back(firstRectangle + Point(0, i * entryBox.Height()), GameData::StartOptions().begin() + i);
+	
+	descriptionWrappedText.SetAlignment(Alignment::LEFT);
```
oo13, line 84: Why do you use Alignment::LEFT instead of Alignment::JUSTIFIED?

Terin:
The updates from @tehhowch took care of all but one of the scrolling issues I mentioned. Pretty much good to go.

source/StartConditionsPanel.cpp
```diff
@@ -0,0 +1,318 @@
...
+		descriptionBox = startConditionsMenu->GetBox("start description");
+		entriesContainer = startConditionsMenu->GetBox("start entry list");
+		entryBox = startConditionsMenu->GetBox("start entry item bounds");
+		entryTextPadding = startConditionsMenu->GetBox("start entry text padding").Dimensions();
+	}
+	
+	// Default the selection to the most recently loaded scenario, if possible.
+	const auto startCount = scenarios.size();
+	if(startCount >= 1)
+		--startIt;
```
Terin, line 68: The tentative plan with official starts seems to be to have the Dirt Belt, Deep, Paradise Worlds, and Syndicate as separate starts, with the Dirt Belt as the default. Having the list default to the end seems strange in that regard, or even with only plugin starts.

source/StartConditions.cpp
```diff
@@ -47,62 +61,89 @@ void StartConditions::Load(const DataNode &node)
...
+		else if(child.Token(0) == "conversation" && child.Size() >= 2)
+			stockConversation = GameData::Conversations().Get(child.Token(1));
 		else
 			conditions.Add(child);
 	}
+	if(description.empty())
+		description = "(No description provided.)";
 }
-
+
```
Amazinite, line 74: Whitespace.

source/StartConditions.cpp
```diff
@@ -47,62 +61,89 @@ void StartConditions::Load(const DataNode &node)
...
+	
+	if(!GetConversation().IsValidIntro())
+		Files::LogError("Warning: The start scenario \"" + name + "\" has an invalid starting conversation.");
+}
+
+
+
+bool StartConditions::IsValid() const
+{	// A start must specify a valid system.
+	if(!system || !system->IsValid())
```
Amazinite, line 91: Missing new line.

Amazinite:
Unnamed starts don't have their name filled in. Should fill in with their identifier, or if no identifier and no name use (Unnamed Start).
image

source/StartConditionsPanel.cpp
tehhowch: The default-to-end was added to maintain the game's current behavior, which is that the latest-defined `start` controls how the player starts. It's arguably unexpected though, so I've removed it.

source/StartConditionsPanel.cpp
tehhowch: I've removed the left alignment. It's possible some of the justified alignments made @FranchuFranchu 's chosen description text look odd. With justification, the "backstory" description looks good:
image

tehhowch (approved):
Ready to ship:
`feat(mechanics,content)!: Added a "New Pilot" scenario picker, allowing players to choose between game & plugin starts`

Amazinite (approved):

PR #5515, opened 2020-11-22
Fix a bug where some submunitions have wrong velocity.

oo13:
**Bugfix:** This PR fixed a bug where some submunitions have wrong velocity.

## Fix Details
In current master, some submunitions have wrong velocity because the current master ignores "random velocity" of the original projectile and the parameters of the submunition of the submunition. We can notice the submunitions float in space with bad angle.

This PR corrects the bug. A submunition is applied "velocity", "random velocity", and "inaccuracy" parameters over the last velocity instead of "velocity" of the original projectile.

This PR changes the velocity of some submunitions, probably you want to change some "inaccuracy" if you have a large inaccuracy of a submunition and it's the submunition of the submunition or the original projectile have random velocity.

Current Master, Positive Velocity
Peek 2020-11-22 21-00

This PR, Positive Velocity
Peek 2020-11-22 21-02

Current Master, Negative Velocity
Peek 2020-11-22 21-00-n

This PR, Negative Velocity
Peek 2020-11-22 21-02-n

## Testing Done
1. Load pilot [projectile test](https://github.com/endless-sky/endless-sky/files/5579423/projectile_test.zip)
2. Depart.
3. Fire the gun.

## Save File
[projectile_test.zip](https://github.com/endless-sky/endless-sky/files/5579423/projectile_test.zip)

### DIFF
--- a/source/Projectile.cpp
+++ b/source/Projectile.cpp
@@ -76,8 +76,11 @@ Projectile::Projectile(const Projectile &parent, const Weapon *weapon)
 		if(!parent.weapon->Acceleration())
 		{
 			// Move in this new direction at the same velocity.
-			double parentVelocity = parent.weapon->Velocity();
-			velocity += (this->angle.Unit() - parent.angle.Unit()) * parentVelocity;
+			// Velocity may be negative.
+			Point referenceVector = parent.angle.Unit();
+			double parentVelocity = parent.velocity.Y() * referenceVector.Y()
+				+ parent.velocity.X() * referenceVector.X();
+			velocity = this->angle.Unit() * parentVelocity;
 		}
 	}
 	velocity += this->angle.Unit() * (weapon->Velocity() + Random::Real() * weapon->RandomVelocity());


source/Projectile.cpp
```diff
@@ -76,8 +76,11 @@ Projectile::Projectile(const Projectile &parent, const Weapon *weapon)
 		if(!parent.weapon->Acceleration())
 		{
 			// Move in this new direction at the same velocity.
-			double parentVelocity = parent.weapon->Velocity();
-			velocity += (this->angle.Unit() - parent.angle.Unit()) * parentVelocity;
+			// Velocity may be negative.
+			Point referenceVector = parent.angle.Unit();
+			double parentVelocity = parent.velocity.Y() * referenceVector.Y()
+				+ parent.velocity.X() * referenceVector.X();
```
Amazinite, line 82: I'm curious as to why this needs to be done to get the parent velocity. Does parent.velocity.Length() not return the same value? If not, then what's the difference?

oo13: Length() results an absolute value and it cannot decide whether a velocity is positive or negative.
I'm not sure that I may omit handling a negative velocity.

This calculates the part of parallel to the reference vector of the parent's velocity, that is, the signed value of the parent's velocity.

Amazinite: That’s what I thought. Maybe clarify the comment more beyond “Velocity can be negative.”

“To maintain the sign of the velocity, Point::Length can’t be used” or something like that.

source/Projectile.cpp
```diff
@@ -76,8 +76,11 @@ Projectile::Projectile(const Projectile &parent, const Weapon *weapon)
 		if(!parent.weapon->Acceleration())
 		{
 			// Move in this new direction at the same velocity.
-			double parentVelocity = parent.weapon->Velocity();
-			velocity += (this->angle.Unit() - parent.angle.Unit()) * parentVelocity;
+			// Velocity may be negative.
+			Point referenceVector = parent.angle.Unit();
+			double parentVelocity = parent.velocity.Y() * referenceVector.Y()
```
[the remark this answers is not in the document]
oo13: Manual trigonometry, yes.

```parent.velocity``` is not equal to ```parent.weapon->Velocity()``` if this is a submunition of a submunition.
```parent.velocity``` is the real velocity of the parent projectile, but ```parent.weapon->Velocity()``` is the attribute of the parent submunition.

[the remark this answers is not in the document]
Comment (tehhowch) [reply source/Projectile.cpp:81]: We should always use vector math over manual trigonometry. You're projecting the parent projectile/submunition velocity onto its current facing as `x1*x2 + y1*y2`, which is just v1 • v2 and thus should use `Point::Dot` method instead.

**Comment by oo13** [comment]
Thanks, I resolved all reviews.

**Review by Amazinite — APPROVED** [review]

File: source/Projectile.cpp [hunk 3]
```diff
@@ -76,8 +76,10 @@ Projectile::Projectile(const Projectile &parent, const Weapon *weapon)
 		if(!parent.weapon->Acceleration())
 		{
 			// Move in this new direction at the same velocity.
-			double parentVelocity = parent.weapon->Velocity();
-			velocity += (this->angle.Unit() - parent.angle.Unit()) * parentVelocity;
+			// To maintain the sign of the velocity, Point::Length can’t be used.
+			Point referenceVector = parent.angle.Unit();
+			double parentVelocity = referenceVector.Dot(parent.velocity);
+			velocity = this->angle.Unit() * parentVelocity;
```
tehhowch, line 82: I still have some questions on the actual desired behavior (of the code in-general, not just as changed in this PR):

1) What's the end goal, i.e. what _should_ the velocity of a new submunition be, after we've finished creating it? We have a projectile or submunition that is expiring, and is producing submunition(s). At the start of this block scope, a produced submunition `this` is traveling with the same speed and same direction as the dying parent, but is pointed in some different direction. I see two broad descriptions of how we can treat this process:
    1) `this` should be "perfectly emitted," meaning the overall magnitude of the parent velocity is applied to the new submunition's direction, 100%. (This means the game will not preserve player-perceived momentum when creating submunitions. This will be particularly noticeable if multiple submunitions are created and have different directions from each other.)
   2) `this` should be "physically emitted," meaning we apply only some portion of the parent's velocity to it. One possible implementation would be to set the submunition's velocity magnitude to be the same as the parent's projected velocity along the submunition's facing angle, e.g trading some of original momentum for orthogonal momentum. In this mode (absent any random velocity present in the new submunition's `weapon`), I think the player would observe the original velocity of the parent when viewing the motion of all created submunitions as a whole -- think of a projectile that explodes like a firework.

2) Can we deduce the reasoning behind only applying this compensation when the _parent_ projectile/submunition has no acceleration? I am left to wonder if this was meant to be performed only if the _new submunition_'s weapon has no acceleration (and thus the direction in which the submunition is traveling cannot change).

oo13: I understand what you mean:
projectile
i. and ii. are not too different, if "inaccuracy" is ordinarily range.

I guess it's the reason why a projectile that has acceleration does not apply this compensation that the projectile is probably a physical missile on the known physical law and has suitable sprite like a rocket. I wonder he didn't expect a missile firing a beam weapon.

Amazinite: My vote is for "perfectly emitted." As for why it is that this only happens without acceleration, I have no idea.

Ferociousfeind: Lemme toss my hat in the pile
I'd like an attribute which can enable or disable this PR's effects (kind of. The main issue is that a weapon's submunitions, and a submunition's submunitions, act differently) so that you can choose either the unrealistic but very nice to work with PR behavior or inheriting only velocity magnitude and applying that to the submunition's direction, or the more realistic inheriting velocity magnitude and direction, and letting the submunition's own acceleration+drag to do its own moving.

I need this PR for some flashy lightning weapon, so that it could theoretically infinitely branch out (but I will only have it branch out about 6 times) and look like actually tree-like branchy lightning.

PR #5523, opened 2020-11-28
Fix: Changing MCO's person ship to have 12000 hull instead of 1200

Zitchas:
**Bugfix:**

## Fix Details
Dropping from 8700 hull to 1200 hull seems like an extreme weakening of the base ship. Increasing from 8700 to 12000 seems reasonable for the appearance and being a person ship.  So this fix assumes that the 1200 is an error wherein it is missing a 0, and adds that missing zero.

### DIFF
--- a/data/persons.txt
+++ b/data/persons.txt
@@ -1052,7 +1052,7 @@ ship "Marauder Bactrian"
 		category "Transport"
 		"cost" 17600000
 		"shields" 30000
-		"hull" 1200
+		"hull" 12000
 		"required crew" 70
 		"bunks" 93
 		"mass" 650


MCOfficer (approved):
whoops, i wonder when that happened

Zitchas:
I'm not sure, but it is kind of funny that so many of us missed it. You made that section, I PR'd, a bunch of people looked at it...

If nothing else, this is probably an example of why merging stuff sooner rather than later so the larger community actually has it in front of them to look at is a good thing. Hundreds of people playing the game and looking through the files is probably a more reliable filter for one-off mistakes than the dozens or so people that peruse GitHub. (The Github people probably still catch the vast majority, but better to push out to the larger community sooner rather than waiting for perfection from the small group here)

Amazinite:
I thought it was intentional, because that’s exactly what you originally posted for the stats. @MCOfficer

PR #5526, opened 2020-11-30
Fix: Remnant: Astral Job - Cognizance compatabiliy

Zitchas:

## Fix Details
This adds the "inaccessible" attribute to those systems that the Remnant gain access to in the first half of the Cognizance missions.
Then it modifies the Astral research job so that it ignores systems marked with "inaccessible"

It also adds a new mission, "Cognizance 19", which is a very short navigational map data delivery to a random Remnant world, which triggers an event that removes the "inaccessible" tag from those systems.

(a new mission was going to be required one way or another. Either a retroactive "compatibility" mission for those who have already done the Cognizance missions along with a modification to an already existing event, or an invisible mission that triggers this. If I'm adding a new mission, I'd rather one that actually informs the player as to the fact that the Remnant now have maps of these systems, and thus it won't be a shock when the Astral job sends them there.)


Amazinite (changes requested):
I’d suggest removing the system attribute when these systems become accessible instead of having a separate job. The system attribute could also be more generic, perhaps “inaccessible.” That way we don’t need to update these jobs if any future systems exist that can’t be accessed and wouldn’t make sense to have “cognizance” as their system attribute.

Zitchas:
The reason that I tagged these systems specifically with "Cognizance 1" rather than a generic "inaccessible" tag is because there will be a need to identify this particular cluster of systems later on in the mission string. They aren't just generic "inaccessible" systems, and that particular attribute tag will be used for missions beyond these jobs.

tehhowch:
In the event, swap the attributes out. 

Consider the arbitrary `"inaccessible"` attribute a workaround until we implement a way to signal "is it reachable in any way?" for the distance/near filter keywords (rather than just the current "is it hyperlinked to X?").

Zitchas:
OK, switched to swapping out the inaccessibility tag. I can see how that could be useful when new systems are added - particularly if they involve weird happenings like these ones did. 

Also added a new mission "Cognizance 19" which is a very short mission (with no option to refuse) wherein Dawn requests that the player upload their nav data somewhere. That triggers the event that removes the inaccessibility tag, along with a story reason for why now is the time that those systems will be available for those jobs.

data/remnant/remnant missions.txt
```diff
@@ -3729,8 +3768,10 @@ mission "Remnant: Expanded Horizons Astral job"
 	repeat
 	waypoint
 		not attributes "pleiades"
+		not attributes ""inaccessible""
```
tehhowch, line 3771:
```suggestion
		not attributes "inaccessible"
```

data/remnant/remnant missions.txt
```diff
@@ -3729,8 +3768,10 @@ mission "Remnant: Expanded Horizons Astral job"
...
 	to offer
 		has "Remnant: Expanded Horizons Astral 2: done"
+		not "Remnant: Cognizance 18: done"
```
tehhowch, line 3774: Still needed?

data/remnant/remnant missions.txt
```diff
@@ -3466,6 +3466,45 @@ event "remnant: cognizance timer"
...
 
+mission "Remnant: Cognizance 19"
+	name "Deliver nav data"
+	description "Deliver navigational data on the new systems to <destination>."
+	landing
+	source
+		government "Remnant"
+	destination
+		government "Remnant"
+	to offer
```
tehhowch, line 3477: You may get the same planet as source and destination with this limited of a destination filter

data/remnant/remnant missions.txt
```diff
@@ -3466,6 +3466,45 @@ event "remnant: cognizance timer"
...
+		has "Remnant: Cognizance 18: done"
+	on offer
+		conversation
+			`As you finish your landing checks your comlink notifies you that you have an incoming message from Dawn. It promptly displays a short message:`
```
tehhowch, line 3481:
```suggestion
			`As you finish your landing checks, your comlink notifies you of an incoming message from Dawn:`
```

data/remnant/remnant missions.txt
```diff
@@ -3466,6 +3466,45 @@ event "remnant: cognizance timer"
...
+			`	"Greetings, Captain <first>! We are starting to compile updated navigational data for the new systems, and would like a copy of the data your nav system recorded. Stop by <planet> and the starport automated systems will get a copy of the data during the systems check. It is a pretty small data set, so there should be no need to use data crystals. Thank you!"`
```
tehhowch, line 3482:
```suggestion
			`	"Greetings, Captain <first>! We will be compiling updated navigational charts for the new systems and would like a copy of the data your nav system recorded. Stop by <planet> and the starport systems will copy the data during the systems check. It's a pretty small data set, so there's no need to use data crystals. Thank you!"`
```

The logical question a player might ask is, "why didn't those starport systems get the data from me when I landed here?", especially since the player may land on any remnant world to trigger this.

data/remnant/remnant missions.txt
```diff
@@ -3466,6 +3466,45 @@ event "remnant: cognizance timer"
...
+	landing
+	source
+		government "Remnant"
+	destination
+		government "Remnant"
+	to offer
+		has "Remnant: Cognizance 18: done"
+	on offer
+		conversation
+			`As you finish your landing checks, your comlink notifies you of an incoming message from Dawn:`
```
W1zrad, line 3481:
```suggestion
			`As you finish your landing checks, your commlink notifies you of an incoming message from Dawn:`
```
Is this right? I don't know, actually. Just thought I'd point it out.

data/remnant/remnant missions.txt
```diff
@@ -3466,6 +3466,45 @@ event "remnant: cognizance timer"
...
+			`	"Greetings, Captain <first>! We will be compiling updated navigational charts for the new systems and would like a copy of the data your nav system recorded. Stop by <planet> and the starport systems will copy the data during the systems check. It's a pretty small data set, so there's no need to use data crystals. Thank you!"`
```
W1zrad, line 3482:
```suggestion
			`	"Greetings, Captain <first>! We will be compiling updated navigational charts for the new systems and would like a copy of the data your nav system recorded. Stop by <planet>, and the starport systems will copy the data during the systems check. It's a pretty small data set, so there's no need to use data crystals. Thank you!"`
```

data/remnant/remnant missions.txt
```diff
@@ -3466,6 +3466,45 @@ event "remnant: cognizance timer"
...
+			`	The end of the message is the location of a starport on <planet>, presumably where you are supposed to go.`
+				accept
+	on complete
+		payment 100000
+		dialog `As the starport computer downloads the data, your computer displays a message tacked onto the log file: "Thank you for providing your navigational logs. They will be useful to contrast against those of our scouts. A logistics adjustment of <payment> has been authorized for the <ship>."`
```
W1zrad, line 3487:
```suggestion
		dialog `As the starport computer downloads the data, your computer displays a message tacked onto the log file: "Thank you for providing your navigational logs. They will be useful to compare with those of our scouts. A logistics adjustment of <payment> has been authorized for the <ship>."`
```
`contrast against` is fine, I just personally prefer `compare with`. Obviously it sounds like those are antonyms, but they essentially mean the same thing: both imply you're looking for similarities and differences in two "objects". I guess `contrast against` prioritizes differences, while `compare with` prioritizes similarities. If you think `contrast against` is more what you're going for, then feel free to resolve. Up to you.

W1zrad (commented):
Just a comma, a spelling question, and a wording suggestion.

data/remnant/remnant missions.txt
```diff
@@ -3466,6 +3466,46 @@ event "remnant: cognizance timer"
...
+	on offer
+		conversation
+			`As you finish your landing checks, your comlink notifies you of an incoming message from Dawn:`
+			`	"Greetings, Captain <first>! We will be compiling updated navigational charts for the new systems and would like a fresh copy of the data your nav system recorded. We are hoping to retrieve some additional data that is not routinely downloaded by the automated starport systems. Stop by <planet> and the starport systems will copy the data during the systems check. It's a pretty small data set, so there's no need to use data crystals. Thank you!"`
+			`	The end of the message is the location of a starport on <planet>, presumably where you are supposed to go.`
+				accept
+	on complete
+		payment 100000
+		dialog `As the starport computer downloads the data, your computer displays a message tacked onto the log file: "Thank you for providing your navigational logs. They will be useful to contrast against those of our scouts. A logistics adjustment of <payment> has been authorized for the <ship>."`
+		event "remnant: cognizance 1 systems available"
```
Amazinite, line 3489: Stick this above the dialog.

Amazinite (changes requested):
Wizrad's review still needs addressed too.

Zitchas:
Sorry. Going through some RL stuff. Should be able to get back to this and address everything in the next few days.

data/remnant/remnant missions.txt
Zitchas: She'd be looking for both, but differences are probably more intriguing to her: What has changed between your first-on-the-scene info vs the subsequent scouts?

data/remnant/remnant missions.txt
Zitchas: Thanks, fixed!

data/remnant/remnant missions.txt
Zitchas: Fixed. Thanks!

Amazinite (approved):

PR #5527, opened 2020-12-01
ci: add test for memory leaks during file loading

tehhowch:
testing the job

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,6 @@ name: CI
 # - Run unit tests for all platforms (Blocked by XCode build file creation.)
 # - Reduce duplicate code (Blocking: https://github.community/t5/GitHub-Actions/Support-for-YAML-anchors/td-p/30336)
 # - Investigate using scoop as package manager on windows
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -399,3 +398,33 @@ jobs:
       run: chmod +x ./"${ES_BINARY}" && ./"${ES_BINARY}" -v
     - name: Parse Datafiles
       run: ./tests/test_parse.sh ./"${ES_BINARY}"
+
+
+  # TODO: run this only if the push modifies source code
```
MCOfficer, line 403: If i'm not mistaken, this is set on  a per-workflow basis:
https://github.com/tehhowch/endless-sky/blob/master/.github/workflows/ci.yml#L35-L47

If you want to exclude datafiles, you may have to extract this into its own workflow.

tehhowch: There are other actions we can use that will allow us to conditionally execute jobs based on paths modified (I linked one in the top of file in the unit test pr). Haven't had time to check that out, though, but it is something we should do eventually.

PR #5553, opened 2020-12-13
Allow ionization to affect disabled ships

Amazinite:
**Bugfix:** This PR addresses an issue I spotted while working on #4520.

## Fix Details
Ships which are disabled are currently unaffected by ion damage. This PR changes that so that no matter the state of a ship, ion damage affects it.

The issue was that ionization was subtracted from energy just after the energy was increased according to energy generation, but this code was not reached if the ship was disabled. Moved where ionization is subtracted from energy outside of the chunk of code that didn't run if the ship was disabled.

## Testing Done
Attacked a disabled ship with and without this change using a weapon with ion damage while using a tac scanner to check energy levels. Without this change the ship's energy levels don't drop while disabled. With this change they do drop.

## Save File
N/A

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1904,7 +1904,6 @@ void Ship::DoGeneration()
 		
 		double coolingEfficiency = CoolingEfficiency();
 		energy += attributes.Get("energy generation") - attributes.Get("energy consumption");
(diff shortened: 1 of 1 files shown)


tehhowch (approved):
LGTM, good find!

PR #5558, opened 2020-12-14
High Heat Tweaks - Adjusting Remnant and Korath ships for the Ember Waste Environment

Zitchas:
**Content (Tweaks)**

## Summary
With the advent of System Hazards, the Ember Waste will become a much hotter environment. Unlike the current status quo where simply "not being hot" is more than enough to survive there, in the new Ember Waste *all* ships are going to need cooling outfits. It won't be an option, it will be a matter of survival. The PR ensures that all Remnant and Korath ships have a basic amount of cooling.

## Details
- Adds Thermoelectric Coolers to all Remnant ships and variants that don't already have one.
- In order to simply add a cooler to all variants, needed to give the Gull +1 outfit space, and the Starling +4
- The exception is the Puffin, which instead gets the equivalent of a TEC in built-in cooling. (both the active cooling and cooling energy)
- All Korath Raider variants have +1 small heat shunt (in most cases, this means they have 2 instead of 1)

Note: I haven't gone through them in detail to ensure that they are all optimized for the new High Heat environment. This PR is a bit more fundamental in seeking to ensure that they all at least have the basic tools to survive.

## Save File
Any pilot with the appropriate Remnant licenses can purchase the base ships and scanners can be used to see the variants.

## Relevant PRs:
- System Hazards #4931 
- Thermoelectric Cooler Tweak #5557 

## PR Checklist
(No graphics were added or modified)


Amazinite:
I'm thinking we should only tweak those ships which actually need it.
I created a fleet of every Remnant and Ka'het ship, as well as the Raider and Chaser, and these were the results I got when IDLING in an Ember Waste system using #5560.

Ship | Heat reached
----|----
Penguin | >200%
Tern | >200%
Faes'mar | >200%
Selii'mar | >200%
Petrel | >150%
Koarth Chaser | >150%
Gull | >100%
Starling | 90%
Albatross | 70%
ibis | 65%
Pelican | 65%
Korath Raider | 65%
Maeri'het | 50%
Telis'het | 45%
Peregrine | 40%

Zitchas:
Thanks for running that test. This part of the work has been on the backburner while the hazards proper are being discussed and tuned.

Amazinite:
All ships from the Ember Waste with their normal idle heat percentage (i.e. without being under the influence of any hazards), a min and max idle heat percentage as observed over the course of three minutes in the Ember Waste for each ship with 3x speed on (i.e. 9 minutes of hazards observed, which an expected 324 heat hazards spawned over the course of that time), whether the ship normally overheats when not under the influence of any hazards, and whether I got it to overheat while in the Ember Waste. Overheat tests involved me constantly moving and firing the ship.

Active cooling requires energy for to work, so it is important to note that ion storms that knock out a ship's power will cause its cooling to fail and its heat to spike. Spikes during ion storms were recorded, but disregarded for the min-max range. Ships with active cooling have been marked with an asterisk, as well as what heat I witnessed them reach after having their energy drained.

This is ONLY testing stock fits, which means variants might have more or less issues than shown here. 

Overheat testing in the Ember Waste was done without taking ion storms into account. Basically anything with active cooling will overheat if an ion storm knocks out its power.

Ship | Normal Idle % | Ember Waste Idle % (min-max observed) | Overheats normally? | Overheats in Ember Waste?
---|----|----|----|---
Korath Raider | 45% | 59-80% | Yes | Yes
Korath Chaser | 0% | 0-14% | No | No
Albatross* | 56% | 62-76% (91% during ion storm) | Yes (but only JUST barely after a while of constant movement and firing.) | Yes (doesn't require an unthinkable amount of movement and firing)
Peregrine* | 20% | 27-37% (55% during ion storm) | No | No
Ibis* | 38% | 45-58% (>100% during ion storm) | No | No
Starling | 50% | 58-72% | No (but maybe if the PDT was firing constantly alongside constant movement and firing) | Yes (but like the Albatross under normal circumstances, only JUST barely; requires constant movement and firing plus a bad heat storm.)
Pelican* | 40% | 45-55% (>100% during ion storm) | No | No
Gull | 54% | 59-73% | No | Yes (requires a bad storm plus constant movement and firing)
Petrel* | 55% | 59-68% (>100% during ion storm) | No | Yes
Tern* | 34% | 36-45% (>100% during ion storm) | No | No
Penguin | 42% | 47-62% | No | No
Puffin | 56% | 64-81% | No | No
Telis'het* | 37% | 45-51% | No |
Maeri'het* | 36% | 43-49% | No |
Faes'mar* | 14% | 14-22% (>100% during ion storm) | No | No
Selii'mar | 44% | 56-68% | No | No

Ka'het data provided by @Galaucus

Amazinite:
So out of this, the only ships I think that need addressed in terms of heat are the Korath Raider (an extra small heat shunt isn't unthinkable for the Korath to carry around in a heat-heavy environment) and the stock Albatross. I can't say how variants will fare, but you might want to test Remnant variants in combat to see how each one does.

What you really need to look out for is how the Remnant should address the ion storms, because they get wombo comboed by the ion storms knocking out their power and the heat hazard then being able to overheat them when they otherwise wouldn't have.

Zitchas:
Hmmm... Tried this new "rebase" thing as an alternative to updating from Master.... I'm not sure if I broke something or not, but having it list all of Amazinite's recent refactoring as "new stuff that will be merged" on the *Files Changed* tab is... worrysome.

Zitchas:
OK, a few changes here.
1) Amazinite suggested that we remove the visuals for the heat hazard because it was causing his new computer to work a bit harder, and if a new system has to work more because of it, what will older systems do? So, visual is gone. Back to the original concept that the heat hazard is represented by the red haze itself.

2) Giving all the Albatross variants an extra TEC (and giving the hull 4 outfit space to accommodate it because at least one variant was maxed to the brim - including an extra 3 outfit expansions). This is basically the brute force method of solving the problem.

Zitchas:
Initial values for testing purposes. They are intended to be low values, just enough to take the edge off the storms. From a gameplay perspective, these are intended to provide just enough protection to make them noticeable and/or attractive to players, without making these generators better "overall." than others. Basically, they should be really strong contenders for "best generator to use for EW activities" while retaining their current ranking for general use in the rest of the galaxy.

As a side benefit, it'll also make the Remnant a touch less suicidal in their use of EMP torpedoes at close range.

data/remnant/remnant outfits.txt
```diff
@@ -357,6 +360,7 @@ outfit "Aeon Cell"
 	"outfit space" -160
 	"energy generation" 21.9
 	"heat generation" 31.5
+	"ion protection" 0.45
```
Amazinite, line 363: Heliarch ships have ion protection values ranging from 0.1 to 0.3, so values like these look way too high.

Instead of using ion protection, I'd suggest ONLY using ion resistance. Using ion resistance means that weak enough storms can be completely ignored while stronger storms only have a fraction of their ionization defended against.

And given that the main issue that ion storms cause with Remnant ships is the deactivation of their active cooling, maybe we only put the ion resistance on their cooling outfits. Whatever value that is used should also be small. Even a value of 0.05 ion resistance will cause a ship to ignore the ion damage on the very start and end of the weakest storms. How about we start with just that on the thermoelectric coolers? That would mean that 10 coolers (40 tons) would be enough to fight all the ionization of the weakest storms. 

If we want to do slightly more than that, keeping the ion resistance on the crystal capacitor should do.

Zitchas: Regarding values, sure, I'll lower them. How would .3 for the largest one, then .2 for the middle, and .1 for the lowest look?

I had a few reasons for picking the outfits I did: 
1) We're dealing with ionization, which impacts energy production. So it makes sense that energy outfits that are getting hit are the most likely things that the Remnant would have focused on to solve the issue.
2) Remnant power generation is almost universally despised. The few people who use it (especially once they have access to Wanderer stuff), do so for RP reasons, not because they actually are better at anything. The Crystal Capacitor has awesome energy density... But everyone hates the power drain and heat it produces, so they avoid it like the plague. If we're going to put a buff on something, things that are generally poorly liked and underutilized makes more sense. (and the TEC already got a minor buff recently, so might as well share the love a bit.)
3) The three generators are larger items, so it'll require more deliberate choices for people to massively stack the resistance, rather than simply grabbing a bunch of fairly compact and already pretty good TECs.

As for the resistance vs protection, my thought process was:
- Generators produce power, and could be designed to withstand ionization.
- The crystal capacitor already has the lore that it bleeds energy - if it bleeds some ionization as well then that actually becomes a good thing, rather than purely a negative.

Zitchas: I'm not sure why I wanted to go with thirds, but anyway. Current modification is .3, .1, and .033 protection respectively; with crystal capacitors unchanged at .1 resistance.

Amazinite: Yeah, I'd still prefer the Remnant use resistance instead of protection.

Zitchas: 1. I don't see the Heliarch protection being cheapened by these outfits. Built in resistance is still massively superior to having to install sub-par generators. Especially since those ships with built-in protection are also some of the better ships in the game.
2. Warship level anti-ionization built into ships that, depending on the lore you go with, are either:
     - built to go up against enemies that raid them roughly once every century, and just minor raids at that.
     - built to go up against enemies that they "beat" once and expect to retaliate against them someday but they haven't seen for centuries, possibly millennia.
     Neither one of these really sounds like justification for a truly spectacular amount of ionic resistance. The urgency and import just isn't there. I'm not saying we should nerf the Heliarchs, I just don't see the need to have them held up as being ultra-special in this area. Especially not compared to faction(s) that deal with large amounts of ionization on a daily basis. 

They'll *still* be very special, this just opens up the possibility for ships to use sub-optimal generators to give other ships some protection too. Endless Sky is all about giving people choices, right? For a late game player who wants to avoid ionization, giving them the choice between high-powered Heliarch ships and other factions' regular ships sacrificing space to run sub-optimal generators... If I discount my fondness for the Remnant, I'd almost certainly pick the Heliarch.

Zitchas:
OK, as far as "not making the Heliarchs less special" front, just comparing a few ship's stock values:

Starling: Protection 0.133 (at a cost of 127 of a Starling's already very limited 312 space)
Breacher & Neutralizer 0.1  (takes up 0 of 372 and 367 space, respectively)

Small ships: Remnant ship has small advantage in terms of stock protection, but Heliarch has the advantage for players as they can strip everything out and have more space to fill - potentially replacing Coalition power with Remnant and ending up with significantly higher. Aside from guns, the Breacher and Neutralizer could fit the entire Starling loadout onboard to get protection 0.233 and still have 55+ extra space leftover. In contrast, if the Remnant want more ionic protection, they're going to have to sacrifice either weapons or engines, and thus impede their ability to accomplish anything. Sounds like a definite advantage for the Heliarchs in terms of handling Ions while still accomplishing other tasks. 

Albatross: 0.4 (using up 258 of 668 outfit space)
Neutralizer: 0.3 (using up 0 of 885 outfit space)
Pretty much the same thing here.

To me, the story these numbers tell is this: The faction that basically grew up with a legendary enemy that uses ion weaponry built their ships to deal with it, but they haven't had to seriously deal with it for ages. As such, they've hardened their ship systems (the static hull) against it, but haven't bothered to expend resources maintaining protection in their active generation (the part that is likely to wear out over time and have to be replaced more regularly.) Building it into their ships ensures that it is always present and optimization-minded captains can't take it out in favour of other more immediate benefits. (So long-term Heliarch planners prepping for the future basically bypass what the ship captains might want by building it in.)

The Faction that originated in an area without historic enemies that use ionic weaponry but is now living in an environment that has varying large amounts of ionization has hardened their core power systems which lets them customize their ship (including sacrificing utility) to vary from marginal to significant in accordance to their mission parameters, and stock come with significant amounts. This is what I'd expect from a group operating in an environment where the hazard is active and variable. If one has active threats, spending precious outfit space to deal with it makes sense, and they'd presumably have lots of practice at hardening their power generation systems against a threat that occurs so regularly. 

tldr. The faction without an active threat has a static amount of protection on their ships against a threat they anticipate but haven't seen recently. The faction with an active threat has expandable protection on active power generation to protect against things that routinely impact them. Thus the Heliarchs have better ships (built-in protection), but Remnant have an outfits they can stack up to provide better protection.

In short, while it is the same thing (protection), the two factions deal with it differently. It is kind of ironic, admitedly, that the faction that is typically all about being modular has it built into their ships, whereas the faction that tends to build stuff into their ships instead has it modular. I don't think this is a bad bit of diversity to have in the game, and it definitely makes a certain amount of logical sense.

Amazinite:
My concern is not with the amount of space it takes up for each faction, my concern is with the raw numbers themselves. It seems odd to me that a T2.5 faction wouldn't have stronger ion protection than a sub T2 faction.

Zitchas:
That's a valid concern. I'm just pointing out that, to me, it seems odd that a faction for whom dealing with ionic threats is a distant "we think it will happen someday" would have higher ionic protection than a faction for whom dealing with it is a daily (and longstanding) occurrence. The  Heliarch should definitely have better protection in terms of efficiency of space used and whatnot, but not necessarily better overall protection.

data/remnant/remnant outfits.txt
```diff
@@ -357,6 +360,7 @@ outfit "Aeon Cell"
 	"outfit space" -160
 	"energy generation" 21.9
 	"heat generation" 31.5
+	"ion resistance" 1
```
Amazinite, line 363: So an Aeon Cell plus one Crystal Capacitor is enough to completely counteract all ion storms? The worst an ion storm can get is dealing 1.5 ion damage per frame. Aeon Cell + Crystal Capacitor = 1.5 ion resistance. Putting that into perspective, the ion DPS of an Ion Cannon is 0.083, meaning you'd need 18 Ion Cannons just to MATCH the resistance of this combo. And looking at a stock Albatross, which has an Epoch Cell on top of the Aeon Cell and Crystal Capacitor, you'd need 24 Ion Cannons to match the resistance of the ship.

Amazinite: Values I'd suggest:
* Crystal Capacitor: 0.01
* Millennium Cell: 0.025
* Epoch Cell: 0.085
* Aeon Cell: 0.14

Zitchas: Those values look a little low to me, but sure, worth trying out. Making the change.

Amazinite: These are basically the minimum values to look at. We can ratchet them up from here.

PR #5560, opened 2020-12-14
Content: Implementing System Hazards in the Ember Waste & Sagittarius A* systems

Zitchas:
## Summary
With the merging of the System Hazards PR by Amazinite (which hasn't happened as of the time of this PR but is hoped to be soon), we now have the ability to give systems their own unique impacts on their environment. This PR implements this hazard system by making the Ember Wastes some hazards, like their foreboding environment suggests. It also gives the black hole at Sagittarius A* an actual gravitational pull. 

**The Ember Waste**
The hazards in the ember waste have two individual elements:
First is the heat - This has been calibrated to be sufficient to push a collection of medium human warships that have been rigged to idle at roughly 0 heat up to somewhere in the 25-60% range. It will vary due to the nature of the hazard, and may occasionally spike to considerably higher than that, although these spikes should be rare.

~Second is the slow effect - Building on the lore that something catastrophic happened in the Ember Wastes, this ambient slow effect is intended to convey the idea that space/time itself isn't entirely stable, resulting in a reduction in mobility of those ships that operate within it.~ (this has been removed. I like the idea, the implementation (especially of the visible slow effects) not so much.)

Third is the storms. These energy storms are timed to be rare, strong, and not terribly long. They tie into the lore of catastrophic events in the past still lingering in the area, with the results of these storms and their strong double-hit of slow and ionization with a small amount of shield damage. 

The sum total of these (especially the storms) is intended to create an environment that can, on occasion, paralyze many ships through heat and/or energy loss, but shouldn't "lock" a ship permanently - thus the player should always be able to escape, even if they might get knocked out now and then while doing it. 

**Sagittarius A***
The current gravitational force was calibrated against a Starling. It was designed to provide enough force that it could almost trap the player in it when they are flying a starling, but not completely do so. It is *not* intended to simulate an actual black hole ~, so there is no lethal damage - just an attraction~. It should be a fairly strong effect for most ships, but I fully expect that some ships will be able to shrug it off easily while others actually do get stuck. (stuck ships can still hyperjump). As per people's request, Sagittarius now has a deadly gravitational effect. Ships have to be pretty close to the black hole to be affected, but if they remain in its effect for very long it'll munch through ships that aren't prepared to deal with it.

## Details
Note: This content requires Amazinite's "System Hazards" PR in order to function. It can be tested by switching to this branch and using the application pulled from the System Hazards PR.

Adds the following hazards:
- `"Ember Waste Base Heat"`
- `"Ember Waste Base Storm"`
- `"Strong Gravity"`
- `"Deadly Gravity"`
- Accretion disk

Gives all systems in the Ember Waste:
- `"EW Base Heat" 100`
- `"EW Base Storm" 9000`
- Certain "safer" systems have storms at 25000 instead of 9000.

Gives Sagittarius A* Grav Strong 1 and Grav Deadly 1 and accretion disk.

For historical record and so the conversation train makes sense: This originally had another hazard `"EW Base Slow"` which acted like the base heat, except it slowed ships. I liked the idea of it, but as Raven and others point out it both doesn't do much, and has a strong visual effect that reveals cloaked ships like the Archons.

## Save File
For some reason I can't attach files, sorry.  There are no mission or condition triggers required to access this content. Any pilot who has a ship that can access the Ember Waste and/or Sagittarius A* will be able to test this content.

## Related PRs
- System Hazards #4931 
- Thermoelectric Cooler Tweak #5557 
- Tweak Remnant ships for a high heat environment #5558 
- Relative Damage #5569 

## To Do:
- [ ] Run a fleet of Korath Ships through the EW and make sure that they are functional. Ocasional disables due to heat is fine, but they should be able to at least do *some* fighting without overheating...

## PR Checklist
No graphics added or modified in this PR.


ravenshining:
If slowing is a constant effect, that means those purple circles will be showing on your ship at all times.  I'd really rather not see that.  Also, it would spoil cloaked ships, which are common in the wastes.  How about having slow storms, separate from ion storms?

As for Sag A* I'd like there to be some lasting inconvenience at least, like ion damage, and preferably something that kills you but very slowly, like around the rate of a nanobot

tehhowch:
The ember waste is already highly different than all other regions of space - does it really also _need_ to immediately introduce players to 3 kinds of hazards?

I also think a persistent heat-based system hazard that isn't right next to (and _only_ next to) a star makes no sense and would mean there are no hospitable planets in said system. Any such system should also have its hospitable belt radius increased.

Zitchas:
I view it, rather, as making its current aesthetic difference actually substantial. In other words, we've had the warning sign "Caution, danger area!" installed for a few years. These hazards simply implement the dangers the red haze have been warning about. 

Having such a major change in background not actually represent anything that matters to the player is, in my view, both deceptive and counter-productive. Given that we associated red glows with fire and danger, the heat hazard is the one that is the most thematically fitting.

I specifically worked to make it a universal field tied to the system and not to the stars because it *isn't* a star effect. It is a nebula-wide effect. For whatever reason, this area of space is filled with a mix of matter and radiation such that it is - at least in space terms - not the cold hard vacume that we are used to dealing with. Chances are that despite appearances it is actually more of a change to the ambient thermal gradient in the area, such that onboard systems can't radiate or dissipate their heat nearly as effectively as they used to. There's some actual heat out there (the pockets when heat spikes), but mostly it is more of a "heat dissipation doesn't work as well" sort of thing. This heat is explicitly *not* the "standard solar heat" that one would expect to encounter near stars. Likewise, this is tied to source(s) that are not restricted to local space. Kind of like having a big thermonuclear bomb destroy a major city, all the surrounding subburbs are going to be impacted too, even if they aren't technically part of the city where the blast happened.

As far as the habitability goes, well, the Ember Wastes *are* rather sparse in the number of habitable planets, the Graveyard sub-region aside. Many systems don't have *any* planets. Although that being said, since The Graveyard doesn't have the red background, either a) it should gain the red background, or it should lose the mechanic. Possibly a different, not heat related, hazard?

All that aside, the main lore that I have been building for this area of the galaxy is that something *bad* happened here. Not just "oh, we had a war and irradiated a few cities or maybe a planet or two" levels of bad, but the *BAD* that the Drak are worried about. The sort of things that destroy whole star clusters, that warp reality and make physics a little less reliable than normal. The sort of thing that are present day understanding of physics and science says is just plain impossible.

In regards to orbits and habitable zones:
- None of our planets are to scale with their stars.
- Planet orbits definitely are not to scale with their planets. I don't know if they are to scale with their stars, but I'm dubious.
- If an area is deliberately being described as twisting reality and physics, there is little reason to assume that heat output of stars, thermal retention of planets, or other factors that help determine habitability are constant either. 

All that being said, if all that is needed is that the habitable zone and the orbit of inhabited planets is moved out a bit... Sure, I'm happy to do it. If someone could suggest appropriate values, I'll make the change.

In regards to the near constant purple slowing effect... Yeah, I get that. I'm not fond of it. I'm especially not fond of it because it shows up almost immediately even when there is no practical impact of the slowing. Slowing, as an effect, looks spectacular but does little. This PR aside, I'd like to see the slowing effect either have its visual impact reduced or the threshold at which it starts to appear raised. I frequently see ships that - in my view - look like they are sufficiently affected to be crawling through molases, but they're basically functioning at only slightly slower than normal. Might be significant in a long drawn out knife fight or a duel where we are actively dodging shots. But in typical ES quick-and-dirty brawls? Not very useful.

I'm definitely open to considering removing the base slow value and just having the "storms" apply it. And yes, it does reveal cloaked ships. It kind of looks like there are *two* archons in Nenia, at least on my current test pilot.

Arachi-Lover:
I also do think that 3 different hazards at once is excessive, and given raven's point on the visible slowing effect (which Ion damage also has, the lighting zaps that cover the ship afflicted by Ion damage) seems to me it'd be better to maybe just have the heat hazard.

trerri:
For Sagittarius A* it would be cool if the pull forced you to land on the hole, immediately destroying your ship even though some might dislike that.
I really do not care for the proposed hazards for the Wastes, they seem like a constant annoyance. It would be better if the effect was limited to very visible and rare electric storms.

Amazinite:
I'm with @trerri on this one. I'm really not that interested in the nearly constant heat and slowing hazards. I'd keep it to just the ion hazard, which should have a noticeable environmental effect to show when it is active. (With an environmental effect though, you're going to want to drop the range significantly, as otherwise you'd need to generate a lot of effects for there to be anything noticeable.)

Zitchas:
A day of to think about it, and I'm inclined to agree with Raven and a few others: Three hazards is too many, and the visual effects of the Slow are distracting. Given that the "slow" had the least impact on anything (despite being the most visually impactful), it has been eliminated.

The more I think about it, the more I think that is an improvement. Now it is more balanced - with one low-level effect that is pervasive and require attention to manage; and a second - rarer effect that has a strong impact, but can be ignored. (it'll just knock ships out once in a while)

Question regarding timing: Is the Storm too frequent? Not frequent enough?

If the heat is simply a "mere annoyance," then I need to boost it up. This is basically the first time in the game that we have the opportunity to switch things up from "person vs NPC" to being "person vs Environment." As the first such environment, it is intended to be on the easier side of things - The simple novelty of variable external heat effects (and one you can't escape via cloak or afterburners, either) should require deliberate attention. Unlike all the "fixed cost" heat sources in the game, having a variable heat environment that the player may occasionally cross is also likely to make active coolers more useful options, since they tend to be more compact, but only drain energy when needed. That being said, I don't want to ramp the heat up much higher because I don't want it to be a perma-kill thing. I want people to be able to experience it, suffer, escape, and adapt. That, and chances are, this game being what it is, they'll still need to kill things, so I can't expect them to devote all their weapon space to coolers...

LordInsane:
In the absence of sensor interference, I would personally say that being to the level where it can be call an "annoyance" might actually be a _good_ thing (if one doesn't strictly stick to that meaning it'd be only annoying, not gameplay adding). It allows it to be noticeable and force people to consider it without necessarily meaning they have to entirely restructuring their outfit arrangement and, by extension, potentially tactics, leaves it feeling less strange that there's fairly normal planets in there, is just strange enough to stand out as off (as the Ember Waste should), _and_ leaves it open to vary things up in the future simply by having systems that - constantly or perhaps in pulses - do have more dangerous heat levels.

Zitchas:
@trerri @Amazinite You wanted Sagittarius A* to be dangerous? Well, now it is. Just added the "Grav Deadly" hazard, which inflicts a moderate amount of both shield and hull damage (and has 0.25 piercing). I'd like it if some people could go test it and provide some feedback. 

My personal experience with visiting Sagittarius A* with a starling (engines downgraded to just a crucible-class thruster and a forge class steering) was this: 
1) Doing an initial fly-by of the blackhole but not getting within 150 was OK, I could get clear again with some effort. 
2) Then I fly right into the blackhole, and start getting sparks. Shields and hull fluctuating and dropping. Complete stop, appear to be stuck. By rocking back and forth (accelerate, release, then accelerate again) I can gradually build up some speed and get out to the edge of the black sphere part of the image, at which point I can start doing elliptical orbits and something kind of like a slingshot orbit to get more speed. At some point I get outside the range of the damage, and after a minute or two of doing this I escape to safety. Honestly, a lot of fun. It felt very rewarding to escape from it and survive. Also took <5min to do it, and in slower starling than normal, too. (no additional shield or hull regen, either, so just stock defences) This is probably what I'd really like everyone to experience at least once. Feels kind of like the rush of escaping from the Archon that first time.
3) Third time, fly into the blackhole, and sit there watching the sparks and watching how they're damaging my ship. Initially the damage is accumulating, but not very fast. At a certain point (down to about half shields and about 3/4 hull) I decide that maybe I don't actually want to die here, and start trying to repeat my earlier escape. Sadly, I waited too long, and my shields crumble  beneath the weight of the blackhole and I am torn to shreds. I barely managed to get clear of the black sphere part of the image itself before I was disabled.

I intend to go back with a starling with stock engines. (just replacing all the weaponry with ramscoops and a research lab to simulate doing the Astral mission) to see what that feels like. With the more powerful engines I suspect it will be less challenging. But that is as it should be, right? Go properly equipped to deal with the threat, it shouldn't be much of a challenge.

Zitchas:
As far as the general "wide spread" nature of the hazards go, I'm considering removing them from the Graveyard - or at least, from all those graveyard systems that don't have the red haze. Basically, the red haze and the hazard go together.  Escape from one, and you escape from the other. This would reduce the number of systems with the regular heat, and make it more unique to the core Ember Waste systems.

ravenshining:
I'm confused, I've never seen any Graveyard systems that do not have red haze, unless you mean Ak'il?  I just did a fetch and toured the wastes again to make sure - are you talking about some new systems that are in a PR somewhere?

As for the heat, it might be neat to have a sort of pulsating effect to it, so it's constantly varying in intensity.  Having a ships's heat meter going up and down without input might help warn the player something is up, too.

Zitchas:
Hmmm... I'll have to go do a tour myself. Just going through the map file, I'm sure I saw more than one graveyard system without the hazard. Anyway, I'll be happy to be wrong. I'd be sad if they actually didn't have the haze.

@ravenshining Anyway, as far as the heat goes - it *does* pulse. It actually has three separate mechanisms for it.
1) It has a frequency of 100, so there is an average of 100 frames between times it is spawned. When it spawns, it has a random duration between 1 and 200 frames. So roughly half the time it will overlap with the next instance of it happening, which will cause it to stack and push hotter.
2) It has a strength of 0.25 2. This means that every time it spawns, it will pick a random number between these two variables, and that multiplies the effect. So in practice the base heat damage per frame will vary between 10 and 80.
3) I have deliberately omitted the "constant strength" attribute, which means that over the course of the event it will ramp from 10% strength up to 100% strength and back down to 10%. So even in a single event it will have a pulse to it.

The combination of these three variable effects add considerable "randomness" to how it impacts your ship. When I test it, I find that my heat usually goes up for the first little bit (not evenly - sometimes slow, sometimes pretty fast) and then stabilizes somewhere. After that, it goes up and down around that point, sometimes spiking up high enough to disable my ship, but quickly dropping again so the instances of being shutdown rarely last more than a few seconds, and usually not more than once every couple minutes. Of course, this is *heavily* dependant on the ship and how one has it outfitted.  Unless a ship is rigged with significant "overkill" in cooling, though, they should also see some degree of heat increase, and they should see it varying up and down fairly actively all the time.

And is that's an important detail. I wanted this base heat to be high enough that it could - once in a while - overload some ships.

beccabunny:
> Hmmm... I'll have to go do a tour myself. Just going through the map file, I'm sure I saw more than one graveyard system without the hazard.

Can confirm it is only the ex Pug outpost to be blue

Zitchas:
OK, just ignore/forget about that, then.

Amazinite:
Whatever the case, I’m still interested in having an environmental effect for these, and perhaps hit effects as well. While the red haze might show “danger,” it doesn’t show the intensity of hazards currently in the system. Using environmental and hit effects allows the volume of effects to inform the player how bad it currently is. Or the heat hazard I’d imagine these effects to be subtle, perhaps even nearly blending in with the haze as to make a connection between the hazard and the haze, but they should be very noticeable for the job storm given that that is more rare and impactful.

Zitchas:
Sure, but I don't consider having it essential, either. I'd actually rather *not* have another on-the-ship visual effect, actually. I like seeing my ship - I'd much rather have a flashing warning icon next to my heat guage than another ship overlay. I'd love to have something similar for ionization and slow, too, for that matter. Not to mention, I'd like to have those visual effects hidden by cloaks (or better yet, have an attribute that specifies if the effect is hidden or not by the cloak, with the default being "hidden."

Zitchas:
I just did a bit of testing on the World Ships in regards to the heat hazard, and the results seem fairly acceptable. I picked the WS with the highest idle heat and the lowest heat capacity (which turned out to be the Crippled WS). With the current heat hazards, I can spin in circles and fire all my guns for a minute or so before I overheat, and I recover within seconds, and can shoot for another while before overheating again. (this is all on triple speed) I can get a decent amount of fighting/shooting in between shutdowns. And this is the crippled version on top of that. Having them occasionally overheat but recover quickly feels pretty reasonably.

Likewise, for the ion storms.... Yeah, they fry the energy storage, but it doesn't impact production enough to impact engines at all. The WS just keeps on flying. Likewise, the energy meter refills almost before I even notice that it has been hit. That seems good to me.

Next up is testing the Raider.

Zitchas:
Just a note: Once the relative damage PR #5569 is merged, the heat hazard will be modified to use that instead so that all ships are affected relative to their size instead of simply receiving fixed amounts of heat. (larger ships have more surface area to collect heat)

data/hazards.txt
```diff
@@ -0,0 +1,75 @@
...
+	"period" 10
+	"strength" 0.1 2
+	"range" 0 150
+	weapon
+		"hit effect" "nano spark"
+		"shield damage" 50
+		"hull damage" 50
+		"damage dropoff" 0 150
+		"dropoff modifier" 0
+		"piercing" 0.25
```
Amazinite, line 75: How about using a relative hull strength here? These values mean that any ship that falls too close to the black hole will be dead after a minute.
```suggestion
	"constant strength"
	"range" 150
	weapon
		"hit effect" "nano spark"
		"damage dropoff" 0 150
		"dropoff modifier" 0
		"relative hull damage" .00028
		"piercing" 10
```
And I believe this may have been mentioned on Discord, but how about another hazard, "Black Hole Accretion Disk", that extends to the edge of the accretion disk graphic and deals shield damage instead of piercing hull damage, giving the player a warning zone before they actually start to die as they fall into the black hole.

data/map.txt
```diff
@@ -2108,6 +2108,8 @@ system Aescolanus
 	attributes "ember waste" "wormhole"
 	habitable 233.28
 	belt 1323
+	hazard "EW Base Heat" 100
+	hazard "EW Base Storm" 4000
```
Amazinite, line 2112: This means that a storm hazard will appear on average every 4000 frames. Each storm hazard lasts on average (1500+300)/2 = 900 frames. That means that the Ember Waste will have a storm active 900/4000 = 22.5% of the time. That seems a bit frequent to me. The length of the storms seems fine, so how about reducing their spawn frequency to at least 9000. That means that 900/9000 = 10% of the time there'd be a storm.

data/map.txt
```diff
@@ -18250,6 +18306,8 @@ system Pantica
 	attributes "ember waste"
 	habitable 486.68
 	belt 1203
+	hazard "EW Base Heat" 100
+	hazard "EW Base Storm" 25000
```
Amazinite, line 18310: For reference on the last comment, those systems with a lowered spawn rate of the storm will have a storm 900/25000 = 3.6% of the time.

Amazinite (changes requested):

data/hazards.txt
```diff
@@ -0,0 +1,75 @@
...
+#
+# Endless Sky is free software: you can redistribute it and/or modify it under the
+# terms of the GNU General Public License as published by the Free Software
+# Foundation, either version 3 of the License, or (at your option) any later version.
+#
+# Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+# PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+
+hazard "EW Base Heat"
```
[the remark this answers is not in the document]
Zitchas: Sounds good. Changed.

data/map.txt
Zitchas: OK. Seems reasonable. Dropped the frequency down to 9000, being about 10%. Combined with the drop in strength should make them a lot less of a problem, while still being potentially a heavy impact.

data/map.txt
Zitchas: That sounds good to me.

data/hazards.txt
```diff
@@ -0,0 +1,72 @@
...
+	"duration" 1 200
+	"strength" 0.25 2
+	"range" 20000
+	"environmental effect" "heat hazard" 200
+	weapon
+		"relative heat damage" 0.00035
+
+
+
+effect "heat hazard"
```
Amazinite, line 21: New effects should go in effects.txt.

data/hazards.txt
```diff
@@ -0,0 +1,72 @@
...
+	sprite "effect/ew heat/ew heat"
```
Amazinite, line 22: You shouldn't need a nested folder for these sprites.

data/hazards.txt
Zitchas: *nod* I'll move them.

data/hazards.txt
Zitchas: But it makes things so much tidier... (I'd honestly like to  put *all* the effects in their own folders, but I figure some people would object because it just looks like churn. No reason to not start being tidy now, though. I'd be happy to sort them all into folders, though.)

data/map.txt
```diff
@@ -2108,6 +2108,8 @@ system Aescolanus
 	attributes "ember waste" "wormhole"
 	habitable 233.28
 	belt 1323
+	hazard "Ember Waste Base Heat" 100
+	hazard "Ember Waste Base Storm" 9000
```
Amazinite, line 2112: I'd place hazard lines below fleet lines so that the connection between fleet spawns and hazard spawns is easier to make at a glance.

data/map.txt
```diff
@@ -21068,6 +21144,9 @@ system "Sagittarius A*"
 	pos 112 22
 	government Uninhabited
 	habitable 10000
+	hazard "Black Hole Gravity" 1
+	hazard "Black Hole Event Horizon" 1
+	hazard "Black Hole Event Horizon" 10
```
Amazinite, line 21149: Looks like a copy paste moment.
```suggestion
	hazard "Black Hole Accretion Disk" 1
```
If you want to reduce the DPS of the accretion disk then I'd reduce the damage value instead of increasing the spawn time, as that is more deterministic than changing these values here, which result in randomized spawn times with this just being the average time between spawns.

Amazinite (changes requested):

data/hazards.txt
Zitchas: Moved.

data/map.txt
Zitchas: Done. Changed.

Zitchas:
In terms of testing, I flew a starling and took a fleet composed of each of the Raider variants on an expedition out of the EW to Sag A*. (only changes from stock was adding JD to those without and filling empty space with fuel processors or ramscoops) Got them all stuck in the black hole, and then managed to save two of them by flying circles around the black hole to guide them into a slingshot/spiral manouver to escape. 

During the trip to Sag A* got hit by two storms - one in transit and one while idling around Parca refueling. The first one did nothing. Barely even noticed it was there. The second one (which happened in the midst of a fight while my four Raiders were pounding on a lone raider who jumped in) resulted in me getting paralyzed fairly quickly, while the Raiders continued to fight quite a bit longer. They also recovered and resumed their fight faster. it was kind of weird having a battle paused like that. Neat, though.
That was fun, but challenging. Given that adding fun challenges is kind of the point, I judge that a success.

Zitchas:
Is the reduction from 3 space down to 1 intentional? I prefer the 3, as it provides a more clear gap between hazards, not to mention consistency with separating things in other places such as ships and outfits.

Amazinite:
The “standard” is typically 1 line between directly related things (ships and their variants only have one space between them), 2 lines between closely related things (see something like the outfits file), and 3 lines between unrelated things (in the ships file, there are 3 line between ships). The size of the data also matters though. Missions are long and therefore 3 lines is necessary to make it obvious where one ends and the other begins. But these with these hazards effects on the other hand, their definitions are small and therefore 3 lines of white space becomes excessive.

Zitchas:
Reducing the max spike heat? I thought the heat levels were going nicely.

Zitchas:
Ooh, hehe. Now the feedback will get interesting.

Issue #5520, opened 2020-11-26
Strange landing behavior with reverse thrusters


Amazinite:
**Describe the bug**
When attempting to land with the Penguin, which has reverse thrusters built in, the reverse thrusters will fire while the ship is close to the planet, causing you to take longer to land.

**To Reproduce**
Try to land with a Penguin.

**Expected behavior**
The ship shouldn't be fighting itself to land.

**Screenshots**
N/A

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: self-compiled
 - Version: latest commit as of posting (https://github.com/endless-sky/endless-sky/commit/5364605634ff0eabbd3b500c6e3ef53f6554abdb)

**Link to save file**
N/A

**Additional context**
AI use of reverse thrusters was recently updated by #5352.

vitalchip:
What seems to be happening is that on a normal approach the ship reaches the calculated stopping point, cuts thrust and passes the point. This makes the ship turn to face the point and then it decelerates using main engines. When there are reverse thrusters they fire, and instead of smoothly passing the stopping point it gets recalculated due to a decrease in velocity, putting it out in front again. The ship moves forward again to intercept and depending on the magnitude of the reverse thrust they may or may not fire again. This behaviour can also happen right before landing causing the ship to 'slide' sideways under reverse thrust, because its (say) pointing left to intercept the stopping point but firing reverse thrusters will cause it to slow but move right.

vitalchip:
Reverse landings can be improved a bit with a quick code change but they're still slower than a regular landing under most circumstances as the only way I've been able to keep the landing reliable is where forward and reverse thrusters alternate. Anything else I tried just ended up off target at the end and 'circling the drain' to land.

Here's an example, it's what I refer to as the 'shaking your tail feathers' approach.

https://vimeo.com/484002443

vitalchip:
Far less noticable with a small ship like a Penguin.

https://vimeo.com/484006527

I suspect reverse landing could be made faster by lowering the calculated stopping distance for a reverse landing, as it stands it seems to be much longer than required (if I press and hold reverse at the moment the AI first applies reverse thrust I stop long before reaching the planet)  This would require testing over a range of values to be certain of a reliable landing.

vitalchip:
This is the change I made, it just points the ship at its own velocity when reversing, countering drift and stopping the turn around.

[...]

PR #5564, opened 2020-12-18
Fix: Improved reverse thruster behavior for landing

Terin:
**Bugfix:** This PR addresses issue #5520.

## Fix Details
Thanks to the work of Vitalchip, whose work I'm uploading at this PR at their request. For the original issue:

> What seems to be happening is that on a normal approach the ship reaches the calculated stopping point, cuts thrust and passes the point. This makes the ship turn to face the point and then it decelerates using main engines. When there are reverse thrusters they fire, and instead of smoothly passing the stopping point it gets recalculated due to a decrease in velocity, putting it out in front again. The ship moves forward again to intercept and depending on the magnitude of the reverse thrust they may or may not fire again. This behaviour can also happen right before landing causing the ship to 'slide' sideways under reverse thrust, because its (say) pointing left to intercept the stopping point but firing reverse thrusters will cause it to slow but move right.

For the fix:

> Reverse landings can be improved a bit with a quick code change but they're still slower than a regular landing under most circumstances as the only way I've been able to keep the landing reliable is where forward and reverse thrusters alternate. Anything else I tried just ended up off target at the end and 'circling the drain' to land.

And the final implementation:

> This is the change I made, it just points the ship at its own velocity when reversing, countering drift and stopping the turn around.

## Testing Done
Attempted landing on various objects (planets, wormholes, disabled ships) with the patch, with ships that had a variety of reverse thrusters (Penguin, Arfecta, Emerald Sword), and noticed much smoother landing sequences. For a large ship like the Emerald Sword, it may exhibit a "ratcheting" behavior if coming at the target from an angle, but is still faster than the old behavior. Also appropriately works if the ship has weak reverse thrusters, as it will prioritize the quicker turning and forward thrust.

## Performance Impact
No significant change in CPU utilization when attempting to travel with 500 Penguins through a wormhole.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1765,12 +1765,6 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	
 	bool shouldReverse = false;
 	dp = targetPosition - StoppingPoint(ship, targetVelocity, shouldReverse);
-	if(shouldReverse && dp.Length() < radius)
-	{
-		// We can directly use the reverse thrusters to stop at the target.
-		command |= Command::BACK;
-		return false;
-	}
 
 	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .8);
 	if(!isClose || !isFacing)
@@ -1778,7 +1772,10 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	if(isFacing)
 		command |= Command::FORWARD;
 	else if(shouldReverse)
+	{
+		command.SetTurn(TurnToward(ship, velocity));
 		command |= Command::BACK;
+	}
 	
 	return false;
 }
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1766,7 +1766,7 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	bool shouldReverse = false;
 	dp = targetPosition - StoppingPoint(ship, targetVelocity, shouldReverse);
 
-	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .8);
+	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .95);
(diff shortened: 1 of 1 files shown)


Amazinite:
How does this affect boarding, jumping, the hold position escort command, the shift+down hold command for the flagship?

Terin:
Jumping, holding position, and shift+down appear to behave identically. However, boarding/move commands to close-range targets with slow ships may be a regression, as the ratcheting behavior I described gets exacerbated trying to hit that smaller target. I'm going to temporarily close this PR to do some video testing of each, so we can see side-by-side.

Terin:
Vitalchip increased the angle for MoveTo() from 80% to 95%, which greatly improves the worst-case scenario from before, while providing no significant regressions. While it may be marginally slower in a few cases, the noticeable speed increase over the current implementation in the worst case, plus the much better landing movement makes it ready to merge, in my opinion.

source/AI.cpp
```diff
@@ -1765,20 +1765,17 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
...
-	if(shouldReverse && dp.Length() < radius)
-	{
-		// We can directly use the reverse thrusters to stop at the target.
-		command |= Command::BACK;
-		return false;
-	}
 
-	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .8);
+	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .95);
 	if(!isClose || !isFacing)
```
Amazinite, line 1770: When the `if(should reverse...)` bit was added above, this line was changed from `if(!isClose || (!isFacing && !shouldReverse))` to what it is now. Should this line be changed back to what it was?

Amazinite: Yeah, I think this needs readded, as otherwise you could do a command.SetTurn here and down on line 1776.

Amazinite (changes requested):

Amazinite (approved):

PR #5567, opened 2020-12-19
Create a gravitational weapon tag that causes hit force to ignore mass

Amazinite:
**Feature:** This PR implements the feature request discussed as a way to improve system hazards (#4931).

## Feature Details
Adds a new weapon tag `"gravitational"` that causes the hit force of a weapon to not be affected by the mass of the impacted ship. Uses a fixed scaling of 400 mass as to prevent hit force values from needing to be tiny. That is to say that with the `"gravitational"` tag, all ships will be impacted by hit force as if they have a mass of 400.

## UI Screenshots
N/A

## Usage Examples
All ships hit by this weapon or hazard will be impacted by this hit force of -15 in the same way regardless of mass.
```
[outfit | hazard] <name>
	...
	weapon
		...
		"gravitational"
		"hit force" -15
		...
```

## Testing Done
Gave the gravity system hazard from #5560 the "gravitational" tag and entered Sag A* with multiple ships across a wide range of masses. All ships fall toward the center at (roughly) the same rate (with any different in acceleration being caused by the damage dropoff modifier), whereas without this tag lighter ships get pulled toward the center quicker.

## Performance Impact
N/A

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -3011,8 +3011,17 @@ void Ship::TakeHazardDamage(vector<Visual> &visuals, const Hazard *hazard, doubl
 
 // Apply a force to this ship, accelerating it. This might be from a weapon
 // impact, or from firing a weapon, for example.
-void Ship::ApplyForce(const Point &force)
+void Ship::ApplyForce(const Point &force, bool gravitational)
 {
+	if(gravitational)
+	{
+		// Treat all ships as if they have a mass of 400. This prevents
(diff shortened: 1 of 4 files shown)


Zitchas (approved):
I can't speak to the code, but the concept of having an attribute that causes all ships to be acted upon as if they have a fixed 400 mass will be very useful for gravitational stuff.

Zitchas:
Excellent! I'll implement this in #5560 immediately.

PR #5568, opened 2020-12-19
Allow weapons to deal shield/hull damage as a percentage of a ship's total shields/hull

Amazinite:
**Feature:** This PR implements the feature request discussed as a way to improve system hazards (#4931).

## Feature Details
Adds a new weapon tag `"percent scaling"` that causes shield and hull damage to be scaled to deal a percentage of a ship's total shield and hull. A shield damage value of 1 while the percent scaling tag is present will deal damage equal to 1% of a ship's shields upon impact.

## UI Screenshots
N/A

## Usage Examples
All ships hit by this weapon or hazard will die in about 16 seconds (assuming continuous fire, as `100 / (.1 * 60) = 16.7 seconds`) regardless of how much hull they have.
```
[outfit | hazard] <name>
	...
	weapon
		...
		"percent scaling"
		"hull damage" 0.1
		"piercing" 1
		...
```
## Testing Done
Gave `"percent scaling"` to the deadly gravity hazard from #5560, as well as a piercing of 1, a hull damage of 0.1, and a period of 1. Tested on multiple ships, all which which died in roughly 16 seconds regardless of how much hull they had.

## Performance Impact
N/A

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -3642,6 +3642,11 @@ int Ship::TakeDamage(const Weapon &weapon, double damageScaling, double distance
 	
 	double shieldDamage = weapon.ShieldDamage() * damageScaling / (1. + attributes.Get("shield protection"));
 	double hullDamage = weapon.HullDamage() * damageScaling / (1. + attributes.Get("hull protection"));
+	if(weapon.IsPercentScaled())
+	{
+		shieldDamage = shieldDamage / 100. * attributes.Get("shields");
+		hullDamage = hullDamage / 100. * attributes.Get("hull");
+	}
 	double hitForce = weapon.HitForce() * damageScaling / (1. + attributes.Get("force protection"));
 	double fuelDamage = weapon.FuelDamage() * damageScaling / (1. + attributes.Get("fuel protection"));
 	double heatDamage = weapon.HeatDamage() * damageScaling / (1. + attributes.Get("heat protection"));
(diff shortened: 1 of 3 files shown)


petervdmeer (approved):
The code looks fine, but I have some feedback:
- I feel that the name `"percent scaling"` does not convey the meaning of this mechanism well. Did you consider different keywords like `"relative damage"` or `"proportional damage"`?
- The current setup only allows a weapon to have regular damage or to have this type of damage, but no combination. Is that intended?
   - If we would have combinations of both damage types in future weapons, then I suggest to add 2 new keywords; `"relative hull damage"` and `"relative shield damage"` and just use those to list the damages for this feature).

None of those concerns is blocking, so approving already.

Zitchas:
Personally, I would like to request that something like Peter's suggestion to allow combinations. Being able to do something like
```
"shield damage" 50
"relative hull damage" 0.1
```
Has potential applications. For instance, the above weapon I'd interpret as a weapon that will damage shields in the normal fashion, but once it is through the shield it disables ship system piece by piece instead of just bashing away at the hull. 

As far as that goes, "relative heat" I could see using for hazards.

Amazinite:
Closing this since it seems like people would rather have new damage types (`"(hull | shield | heat | fuel) percent damage"` or similar) instead of a single tag that forces percent scaling for only a subset or all of the damage types that could use such a scaling.

PR #5569, opened 2020-12-22
Create "relative" damage types that deal damage as a percentage of a target's stats

Amazinite:
**Feature:** This PR takes the alternative route to #5568.

## Feature Details
This PR creates various new damage types:
* `"relative shield damage"`: Deals shield damage as a percentage of a target's maximum shield value, where 1 = 100% damage.
* `"relative hull damage"`: Deals hull damage as a percentage of a target's maximum hull value.
* `"relative fuel damage"`: Deals fuel damage as a percentage of a target's fuel capacity.
* `"relative heat damage"`: Deal heat damage as a percentage of a target's heat capcaity.
* `"energy damage"`: Deals damage that subtracts directly from a target's energy. Ionization is now the DoT (damage over time) version of this damage type.
* `"relative energy damage"`: Deals energy damage as a percentage of a target's energy capacity.

Also adds a new protection attribute, `"energy protection"`, to combat the brand new energy damage type.
Speaking of which, damage protection attributes protect against these relative damage types in the same way that they do the normal damage types.

## UI Screenshots
Here is a screenshot of the outfitter with a weapon selected that has all damage types applied, showing the new damage types next to the old one. These new damage types are displayed as `% [type] damage`, as `relative [type] damage` does not fit the UI. A number of people who wanted this mechanic were fine with this naming scheme.
image

## Usage Examples
```html
[outfit | hazard] <name>
	...
	weapon
		...
		"energy damage" <amount#>
		"relative shield damage" <amount#>
		"relative hull damage" <amount#>
		"relative energy damage" <amount#>
		"relative fuel damage" <amount#>
		"relative heat damage" <amount#>
		...
```
While energy damage could be a new damage type for normal weapons, I suggest and expect that these relative damage types only be used on hazards, allowing for easier balancing of hazards as to affect all ships in a similar manner instead of heavily damaging smaller ships while leaving larger ships largely unharmed or only slightly harmed by the hazard.

Although there is nothing stopping you from using these on a weapon for a ship, don't expect me to merge any content using these relative damage types in that manner.

## Testing Done
Tested each of the new damage types against a Sparrow and a Raider. The Sparrow and Raider both either die, lose all their energy and fuel at the same time, or gain heat at the same rate depending on the relative damage type used. Energy damage properly subtracts from a ship's stored energy without applying a status effect.

## Performance Impact
N/A

### DIFF
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -105,7 +105,7 @@ tip "    engine capacity:"
 	`How many tons of engines this ship can hold. (To install an engine, you must also have enough total outfit space available.)`
 
 tip "energy protection:"
-	`Protection against weapons taht deal energy damage. If you add more than one outfit with this attribute, each additional one is less effective: a total value of 1 results in a 1 percent reduction in energy damage, while a total value of 11 only results in a 10 percent reduction.`
+	`Protection against weapons that deal energy damage. If you add more than one outfit with this attribute, each additional one is less effective: a total value of 1 results in a 1 percent reduction in energy damage, while a total value of 11 only results in a 10 percent reduction.`
 
 tip "fighter bays:"
 	`The number of ships from the "Fighter" category that this ship can carry.`
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -104,6 +104,9 @@ tip "engine capacity needed:"
 tip "    engine capacity:"
 	`How many tons of engines this ship can hold. (To install an engine, you must also have enough total outfit space available.)`
 
+tip "energy protection:"
+	`Protection against weapons taht deal energy damage. If you add more than one outfit with this attribute, each additional one is less effective: a total value of 1 results in a 1 percent reduction in energy damage, while a total value of 11 only results in a 10 percent reduction.`
+
 tip "fighter bays:"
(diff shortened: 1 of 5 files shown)


petervdmeer:
I'm very happy with the description of the feature and the keywords here.

One small remark on the UI screenshot; The percent sign at the beginning of the line looks a bit odd. If you can find a way to leave out the beginning-percent-sign, then the text would still look fine (since the value is also displayed with a percent sign).

Amazinite:
> One small remark on the UI screenshot; The percent sign at the beginning of the line looks a bit odd. If you can find a way to leave out the beginning-percent-sign, then the text would still look fine (since the value is also displayed with a percent sign).

The issue is that these lines need to be different than the normal damage lines. While without the percent sign these could still be distinguished from normal damage by the percent sign, that would result in the tooltip for say normal hull damage and relative hull damage being the same, since the tooltip looks for the displayed name of the attribute.

As mentioned though, this is mainly intended for use on hazards, in which case one should really never see these damage types displayed in an outfitter. And those who were interested in these separate damage types (as opposed to the single keyword that would change damage to being relative to the capacity of a stat on a ship) were fine with the `"% [type] damage"` labeling.

PR #5570, opened 2020-12-23
Rebalance default Aphid kit to make use of Sand Cells

Terin:
**Content (Tweak)**

## Summary
Now that the Pond Strider PR is merged, and the Hai have access to a power source less intense than a nuclear reactor, I took a crack at reworking the default Aphid outfit loadout. This PR makes the Sand Cell available by default, puts two of them in place of the default Pebble Core, and swaps out the Pulse Turret for a Chameleon Anti-Missile.
 
image

While it is a thematic point of the Hai that they are less relentless in improving efficiency, the Aphid is extremely poorly balanced in terms of its outfit to hull cost (nearly 4 to 1!). This is mainly due to the Pebble Core providing far too much power, and the Pulse Turret giving poor defense, not letting an Aphid outgun a lone Grasshopper.

image

This shouldn't conflict with Nanachi's mission string, either. While 1.5 million credits is certainly a lot more affordable than nearly 3 million, a Heavy Shuttle is still about a third of the cost of the updated Aphid. That's still a proposition that would get the attention of a new pilot like her.

## Details

- Swapped out the Pulse Turret and Pebble Core for a Chameleon Anti-Missile and two Sand Cells, and removed the Williwaw Cooling from the stock Aphid.
- Adjusted Aphid description from "lightly armed" to "comes with basic defenses".
- Added the Sand Cell to the Hai Intermediate and Advanced outfitters, and removed the Pond Strider mission event to add them.
- Adjusted Sand Cell description to refer to "ships that don't need or are unable to carry a nuclear reactor", instead of only the Flea.
- Changed Aphid Anti-Missile variant to an Armed variant, using the original outfit loadout. Swapped instances of each in Hai fleets, as well.
- Adjusted reference to Aphid cost in the Turner Inc. shipyard mission. If this isn't nearly as impressive, we could rewrite that section so Turner compares the Aphid to the Syndicate Freighter, which he would likely view more favorably as someone with Syndicate/Megaparsec connections.

### DIFF
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -917,7 +917,7 @@ mission "Expanding Business [6]"
 			`	"You could call it that," Turner says after sipping from his glass of wine.`
 			
 			label proposition
-			`	"Building an outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Hai sell the Aphid for nearly a million and a half credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this? You look at something we humans sell like a Star Barge, which is hardly worse than an Aphid for only a quarter of the price, and you start to see the problem."`
+			`	"Building an outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Hai sell the Aphid for nearly one and a half million credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this? You look at something we humans sell like a Star Barge, which is hardly worse than an Aphid for only a quarter of the price, and you start to see the problem."`
 			choice
 				`	"You need me to help you escort freighters again?"`
 				`	"Do you want to build a shipyard selling human ships?"`
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -917,7 +917,7 @@ mission "Expanding Business [6]"
 			`	"You could call it that," Turner says after sipping from his glass of wine.`
 			
 			label proposition
-			`	"Building a outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Aphid is nearly a million and a half credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this?"`
+			`	"Building an outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Hai sell the Aphid for nearly a million and a half credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this? You look at something we humans sell like a Star Barge, which is hardly worse than an Aphid for only a quarter of the price, and you start to see the problem."`
 			choice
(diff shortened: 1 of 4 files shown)


data/hai/hai missions.txt
```diff
@@ -917,7 +917,7 @@ mission "Expanding Business [6]"
 			`	"You could call it that," Turner says after sipping from his glass of wine.`
 			
 			label proposition
-			`	"Building a outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Aphid is nearly three million credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this?"`
+			`	"Building a outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Aphid is nearly a million and a half credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this?"`
```
Amazinite, line 920: While we're here, this should be `an outfitter`. Also changing the description here to compare the Aphid to something instead of just stating its cost.
```suggestion
			`	"Building an outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Hai sell the Aphid for nearly a million and a half credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this? You look at something we humans sell like a Star Barge, which is hardly worse than an Aphid for only a quarter of the price, and you start to see the problem."`
```

WarlordMike:
I like this idea. The other advantage with making the Sand Cell available from the word "go" is it benefits newer players who want something better than the Human Fuel Cells but either can't afford or don't need to upgrade straight to a reactor.

I would suggest is making sure the Williwaw Cooling is on the new (Armed) variant, so it's in line with its original incarnation. Also, I would change the statement from "a million and a half" to "one and a half million" in Nanachi's string.

Otherwise, looks good!

data/hai/hai missions.txt
WarlordMike, line 920:
```suggestion
			`	"Building a outfitter selling human goods here on <origin> has had to be the best idea I've ever had, Captain <last>, but recently I thought of an even better idea. You see, most of the humans you see flying around in Hai space are flying decades-old ships. Hai ships are prohibitively expensive for humans. The Aphid is nearly one and a half million credits! This means that demand for ships is extremely high among humans, Captain <last>. Do you see where I'm going with this?"`
```

Suggested conversation edit.

Amazinite (changes requested):

data/hai/hai ships.txt
```diff
@@ -32,26 +32,25 @@ ship "Aphid"
...
-	description "The Aphid is the lightest freighter used by the Hai and was for many millennia the backbone of the Hai inter-planetary economy. Because of how peaceful Hai space has been for much of its history, Aphids are only lightly armed and mainly rely on larger Hai ships for protection."
+	description "The Aphid is the lightest freighter used by the Hai and was for many millennia the backbone of the Hai inter-planetary economy. Because of how peaceful Hai space has been for much of its history, Aphids only come with basic defenses as stock, and rely on larger Hai ships for more serious protection."
 
-ship "Aphid" "Aphid (Anti-Missile)"
+ship "Aphid" "Aphid (Armed)"
 	outfits
-		"Chameleon Anti-Missile"
+		"Pulse Turret"
 		"Pebble Core"
 		"Hai Fissure Batteries"
```
WarlordMike, line 55:
```suggestion
		"Hai Fissure Batteries"
		"Hai Williwaw Cooling"
```

Re-add the Williwaw Cooling to the (Armed) variant so it resembles the ship we once had.

WarlordMike (commented):
Suggested edits contained within.

data/hai/hai ships.txt
Terin: It should still have the Williwaw cooling, it's just cut off underneath the line break.

PR #5572, opened 2020-12-24
Check if a ship is overheated during Ship::TakeDamage

Amazinite:
**Bugfix:** This PR addresses an issue spotted by Terin on Discord.

## Fix Details
Adds a check to see if a ship is overheated at the end of TakeDamage.

Currently in the game, the only time that a ship checks if it is overheated is after it has finished its shield and hull generation. This results in a scenario where a ship that has been overheated will do its shield and hull generation for a frame after taking damage, since TakeDamage overwrites the isDisabled bool that should stop shield and hull generation without checking to see that the ship has been overheated.

## Testing Done
Attacked a Shield Beetle with a flamethrower before and after this change. Before, causing the ship to take damage while overheated would activate its shield generation. After, the ship no longer generates its shields for a frame after taking damage, with everything else working normally. (The ship still properly becomes unoverheated, overheated ships still flash in the HUD.)

## Save File
N/A

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -3719,15 +3719,14 @@ int Ship::TakeDamage(const Weapon &weapon, double damageScaling, double distance
 	if(!wasDestroyed && IsDestroyed())
 		type |= ShipEvent::DESTROY;
 	
-	// Check if the ship is disabled by being overheated. Do this now so that
-	// overheated ships don't count as disabled for the purposes of missions.
-	// Do a full overheat check here in case the overheating was caused by
-	// taking damage.
+	// Inflicted heat damage may also disable a ship, but does not trigger a "DISABLE" event.
 	if(heat > MaximumHeat())
+	{
 		isOverheated = true;
+		isDisabled = true;
+	}
 	else if(heat < .9 * MaximumHeat())
 		isOverheated = false;
-	isDisabled |= isOverheated;
 	
 	return type;
 }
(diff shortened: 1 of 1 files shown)


Terin:
Previous behavior is shown in this video, with the ship losing energy after being overheated, every frame that it takes damage.

https://user-images.githubusercontent.com/4471575/103049250-dc45d480-4556-11eb-9a55-331005c97af9.mp4

tehhowch (commented):
We currently aggregate disabled-by-low-hull, disabled-by-no-crew, and temporarily-disabled-by-heat into `isDisabled`. I'd support a follow-up PR that removes the heat-based contribution entirely, so that "IsDisabled()" implies "won't reactivate without help." That separation would make it possible for AI logic to determine if a ship that currently can't fight is going to resume fighting, at the trivial cost of internal Ship.cpp code checking both isOverheated and isDisabled when necessary.

Amazinite:
>We currently aggregate disabled-by-low-hull, disabled-by-no-crew, and temporarily-disabled-by-heat into isDisabled. I'd support a follow-up PR that removes the heat-based contribution entirely, so that "IsDisabled()" implies "won't reactivate without help." That separation would make it possible for AI logic to determine if a ship that currently can't fight is going to resume fighting, at the trivial cost of internal Ship.cpp code checking both isOverheated and isDisabled when necessary.

Well calling Ship::IsDisabled() if the ship is disabled already only returns true if a ship is disabled by having low crew or low hull. as it doesn't return the state of the isDisabled variable, but instead checks to see if the crew or hull are low.

PR #5574, opened 2020-12-25
Add ability to sort player fleet from info panel

yjhn:
## Summary
This is a reopening of the #5231, since the original developer appears to be inactive.

This includes all commits by @LegitPanda and I also addressed most of the feedback.


source/ShipInfoPanel.cpp
```diff
@@ -644,9 +649,9 @@ void ShipInfoPanel::Rename(const string &name)
 
 
 
-bool ShipInfoPanel::CanDump() const
+bool ShipInfoPanel::CanDump()
 {
-	if(canEdit || shipIt == player.Ships().end())
+	if(panelState.CanEdit() || shipIt == panelState.Ships().end())
```
petervdmeer, line 654: Please consider to change panelState.Ships() to const or to provide an additional const version of the function.
If panelState.Ships() is also available in const version, then CanDump() can also stay const.

petervdmeer (changes requested):
Looks quite good. I have a some review comments.

source/PlayerInfoPanel.cpp
```diff
@@ -654,13 +796,78 @@ void PlayerInfoPanel::DrawFleet(const Rectangle &bounds)
...
+			panelState.Ships().end(),
+			[&](const shared_ptr<Ship> &lhs, const  shared_ptr<Ship> &rhs
+		){
+			return shipComparator(lhs, rhs);
+		});
+	else
+		std::reverse(panelState.Ships().begin() + 1, panelState.Ships().end());
+	
+	// Load the same selected ships from before the sort
+	for(unsigned int i = 0; i < panelState.Ships().size(); i++)
```
tehhowch, line 826:
```suggestion
	for(unsigned int i = 0; i < panelState.Ships().size(); ++i)
```

source/PlayerInfoPanel.cpp
```diff
@@ -654,13 +796,78 @@ void PlayerInfoPanel::DrawFleet(const Rectangle &bounds)
...
+	{
+		auto ship = panelState.Ships()[i];
+		if(selectedShips.count(ship))
```
tehhowch, line 829: For large lists of ships and/or selections, it will be more performant to use `find` and then erase the particular iterator position, rather than use both `count(key)` and `erase(key)`.

source/PlayerInfoPanel.cpp
```diff
@@ -654,13 +796,78 @@ void PlayerInfoPanel::DrawFleet(const Rectangle &bounds)
...
+		{
+			panelState.AllSelected().insert(i);
+			if(lastSelected == ship)
+				panelState.SetSelectedIndex(i);
+			selectedShips.erase(ship);
+		}
+		
+		if(selectedShips.size() == 0)
```
tehhowch, line 837: never use `size()` as an "is it empty" check -- use `empty()`

source/PlayerInfoPanel.cpp
```diff
@@ -558,44 +671,60 @@ void PlayerInfoPanel::DrawFleet(const Rectangle &bounds)
...
-	table.Draw("ship");
-	table.Draw("model");
-	table.Draw("system");
-	table.Draw("shields");
-	table.Draw("hull");
-	table.Draw("fuel");
-	table.Draw("crew");
+	for(auto column = columns.begin(); column < columns.end(); ++column)
+	{
+		const auto currentColumn = *column;
```
tehhowch, line 682: `auto` does not deduce reference types, so without `&` we would make a copy:
```suggestion
		const auto &currentColumn = *column;
```

(it would be better if we can avoid needing to use an explicit iterator at all, but I accept that we may need this style in this loop.)

source/PlayerInfoPanel.cpp
```diff
@@ -372,45 +466,57 @@ bool PlayerInfoPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &comman
...
 		}
 	}
 	else
 	{
 		// If not landed, clicking a ship name takes you straight to its info.
+		panelState.SetSelectedIndex(hoverIndex);
+		
 		GetUI()->Pop(this);
-		GetUI()->Push(new ShipInfoPanel(player, hoverIndex));
+		GetUI()->Push(new ShipInfoPanel(player, move(panelState)));
```
tehhowch, line 511: Prefer `std::move` over an unqualified `move` (this is not strictly necessary but is a way of highlighting important and uncommon calls. Moving an object means we need to take special care with the moved-from object.
```suggestion
		GetUI()->Push(new ShipInfoPanel(player, std::move(panelState)));
```

source/InfoPanelState.cpp
```diff
@@ -0,0 +1,90 @@
...
+{
+	selectedIndex = newSelectedIndex;
+}
+
+
+
+set<int> &InfoPanelState::AllSelected()
+{
+	return allSelected;
+}
```
tehhowch, line 47: It might make a lot more sense to only provide `const &` access to the selection indices, and provide non-const methods e.g. `Select` & `Deselect` that take a ship and / or an index. That refactor will move maintaining this class's invariant, e.g. that when `AllSelected` is empty, SelectedIndex == -1, and otherwise SelectedIndex is a valid entry in `AllSelected`, from its current (many) place(s) in PlayerInfoPanel into this class.

source/InfoPanelState.h
```diff
@@ -0,0 +1,76 @@
...
+	
+	std::set<int> &AllSelected();
+	
+	bool CanEdit() const;
+	
+	int Scroll() const;
+	void SetScroll(int);
+	
+	std::vector<std::shared_ptr<Ship>> &Ships();
+	// This is needed to keep ShipInfoPanel::CanDump() const.
```
tehhowch, line 47: We don't need to justify providing `const &` const-qualified accessors
```suggestion
```

source/PlayerInfo.cpp
```diff
@@ -1011,6 +1012,42 @@ int PlayerInfo::ReorderShips(const set<int> &fromIndices, int toIndex)
 
 
 
+void PlayerInfo::ReorderShips(const vector<shared_ptr<Ship>> newOrder)
+{
+	// Check if the incoming vector contains the same elements
+	bool same = true;
+	size_t aSize = newOrder.size();
```
tehhowch, line 1019: [`std::is_permutation`](https://en.cppreference.com/w/cpp/algorithm/is_permutation)

source/PlayerInfo.cpp
```diff
@@ -1011,6 +1012,42 @@ int PlayerInfo::ReorderShips(const set<int> &fromIndices, int toIndex)
...
+			}
+			if(!same)
+				break;
+		}
+	}
+	
+	if(same)
+		ships = newOrder;
+	else
+		Files::LogError("Cannot reorder ships because the new order does not contain the same ships");
```
tehhowch, line 1046: This should be a fatal error.
```suggestion
		throw runtime_error("Cannot reorder ships because the new order does not contain the same ships");
```
This thrown error will then be caught by the main program loop and generate both an SDL pop-up and an entry in errors.txt.

(You'll need to `#include <stdexcept>` instead of Files.h

data/interfaces.txt
```diff
@@ -18,6 +18,8 @@ color "bright" .75 .75 .75 0.
 color "medium" .5 .5 .5 0.
 color "dim" .25 .25 .25 0.
 color "faint" .1 .1 .1 0.
+color "special" .5 .5 .1 0.
+color "dead" .4 0. 0. 0.
```
tehhowch, line 22: These should be moved to their own "block" and probably also given more scoped names.

tehhowch (changes requested):
There are conflicts to resolve due to changes in `master` re: Table & Font usage.

In addition to the review comments, I have a refactoring suggestion (move class invariant maintenance duties to InfoPanelState, rather than maintaining them in PlayerInfoPanel).

source/InfoPanelState.cpp
tatami4: Current code is relying on `selectedIndex` and `allSelected` being independent of each other. I would prefer not to change this, as it would introduce more complexity (having to know what is going on behind the scenes in InfoPanelState).

source/PlayerInfoPanel.cpp
```diff
@@ -655,13 +767,72 @@ void PlayerInfoPanel::DrawFleet(const Rectangle &bounds)
...
+	
+	for(int i : panelState.AllSelected())
+		selectedShips.insert(panelState.Ships()[i]);
+	panelState.AllSelected().clear();
+	
+	// Move flagship to first position
+	for(auto &ship : panelState.Ships())
+		if(ship.get() == player.Flagship())
+		{
+			iter_swap(&ship, panelState.Ships().begin());
```
tehhowch, line 787: How portable is using a pointer as an iterator here? E.g. will this fail if we change `panelState.Ship()` to not be a vector? Or if some standard library implements std::vector differently?

source/PlayerInfoPanel.h
```diff
@@ -16,8 +16,13 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "Panel.h"
 
 #include "ClickZone.h"
+#include "GameData.h"
+#include "InfoPanelState.h"
 #include "Point.h"
+#include "text/Table.h"
```
tehhowch, line 22: We can move this back to the .cpp file, now that we don't need it for the `Alignment` enum. (A `text/layout.hpp` include will be needed instead.)

source/InfoPanelState.cpp
tehhowch: It feels like they should be strongly coupled, but we can defer extra refactoring until later.

tehhowch (commented):
Some remaining comments

source/PlayerInfoPanel.cpp
```diff
@@ -246,80 +316,89 @@ bool PlayerInfoPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &comman
...
-			size_t moved = allSelected.size();
-			toIndex = min(player.Ships().size() - moved, toIndex);
-			selectedIndex = player.ReorderShips(allSelected, toIndex);
+			size_t moved = panelState.AllSelected().size();
+			toIndex = min(panelState.Ships().size() - moved, toIndex);
+			player.ReorderShips(panelState.Ships());
+			// The order has now changed, so the ships are no longer sorted.
+			panelState.SetCurrentSort(nullptr);
+			
+			int selectedIndex = player.ReorderShips(panelState.AllSelected(), toIndex);
```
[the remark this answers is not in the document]
tatami4: It's needed because if the player sorted the ships, but didn't save the order and later manually reorders them, they should be reordered from the order currently shown on screen (sorted) and not the last saved order.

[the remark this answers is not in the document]
tehhowch: Yes, but we should write code that does not look like it is a bug. If that's as simple  as changing a method name, we should do that.

We don't technically need these two methods to be overloads (we always know which PlayerInfo method we want to call), so let's rename the whole-order one, e.g. `SaveShipOrder` or equivalent.

source/PlayerInfoPanel.cpp
tatami4: changed to std::swap

tatami4:
I am not sure myself about that refactor. It looks like the amount of code multiplied just to be able to do the same thing.

tatami4:
Converting to draft for now until I fix some bugs and weird behoviour.

tatami4:
Hover handling is now done when drawing. This eliminates a delay of one frame between hovering and highlighting the ship under the mouse.

tatami4:
Now ship order saving is done like this: if the player sorts the ships, the order is not saved, but if the player reorders ships by hand, then order currently seen on screen is saved and then the ships  are reordered.
I would like to make so that ship order is only saved when the player clicks "Save order" button. This poses a problem: reordering methods cannot be moved from PlayerInfo to InfoPanelState, because shop depends on them. Any suggestions for how to solve this?

tatami4:
Ship order is never implicitly saved anymore. The order is only saved when "save order" button is clicked.

Zitchas:
@tatami4 Just an FYI, I'm considering merging this into Revamp due to popular interest. That being said, I'm not a coder, so it would be very helpful if you could resolve that conflict in PlayerInfoPanel.cpp. On the same token, bringing it up to even with the current repository would also be very helpful.

Thanks for making this! I hope it can be merged soon.

tatami4:
> @tatami4 Just an FYI, I'm considering merging this into Revamp due to popular interest. That being said, I'm not a coder, so it would be very helpful if you could resolve that conflict in PlayerInfoPanel.cpp. On the same token, bringing it up to even with the current repository would also be very helpful.

Done.

Zitchas:
Alright. Thanks! I'll see about merging it in the next 24h.

Zitchas (approved):
I can't speak for the quality or aesthetics of the code, but I have tested this in-game (and merged it into Revamp and tested it there too), and it works as advertised.

Looks good, works well, and provides a solid QoL enhancement to the Player Info panel. I'm looking forward to having this in the vanilla game.

petervdmeer (approved):
This change looks good. I noticed some code-changes which I consider out-of-scope just for sorting the shipslist, but those out-of-scope changes also seem to be good changes and they are somewhat close to the in-scope parts of this PR, so I'm approving. (I trust that Zitchas tests were good for properly testing this PR.)

petervdmeer:
All review-comments seem to be handled and I did a short retest; the sorting appears to be working as advertised.

tehhowch (commented):
I don't know how relevant these review comments I had pending still are. Would have been good to verify before merging.

petervdmeer:
> I don't know how relevant these review comments I had pending still are. Would have been good to verify before merging.

Good comments! Unfortunately I cannot see pending comments until they are posted.

Can you mark the comments that don't need follow-up as "resolved"? Then I will do a follow-up for the ones that do need a follow-up.

tehhowch:
I'm aware that pending comments are not visible to anyone else until the review is posted. Generally what will happen is I review a large fraction of some PR and come to the point where I need to run the code to verify what I think I am seeing is accurate. I don't want to post comments that are misleading or incorrect, or suggest an alternative that won't work. So a review will sit pending until I get the time to evaluate my comments locally¹. And that can end up being quite a while due to all of life's more important callings.

I'll make a better habit of the review request functionality: if I've requested a review from myself, I have comments pending or have seen that I need to investigate. It's also reasonable to intuit that a PR for which I have requested changes and then not followed up with after additional commits has some pending commentary.

¹I do almost all of my reviews and comments (including this one) via the GitHub mobile app.

quyykk:
> Now ship order saving is done like this: if the player sorts the ships, the order is not saved, but if the player reorders ships by hand, then order currently seen on screen is saved and then the ships are reordered.

Honestly I liked that behavior more and it matches the previous behavior too. Thoughts?

PR #5575, opened 2020-12-26
feat: Split arrival distance into hyperdrive and jump-drive version.

petervdmeer:
**Feature:** This PR implements the Jump drive part of the feature request detailed and discussed in #4334

## Feature Details
This PR splits the keywords for arrival distances in systems into a hyperdrive-version, a jumpdrive version and a common syntax to set both.
This feature only works for systems. Setting arrive distances relative to a planet is not implemented yet.

## UI Screenshots
N/A

## Usage Examples
Use the following keywords on relevant systems:
To set a default arrival distance for all (jumpdrive, hyperdrive):
```
arrival <nr>
```

To set a default arrival distance, but a specific one for hyperdrives:
```
arrival <nr>
   link <nr>
```

To set an arrival distance for jump drives only:
```
arrival
   jump <nr>
```

To set an arrival distance for jumpdrive and hyperdrive only (different from first example in case we get a 3rd drive type):
```
arrival
   link <nr> 
   jump <nr> 
```

## Testing Done
Tested with a flight through Acamar, Aldebaran, Capella and Eteron which all had some form of travel distances set.

## Performance Impact
Calculation is just an addition and only performed once per ship per entering of a system. No noticeable performance impact expected.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -21145,7 +21145,7 @@ system "Sagittarius A*"
 	government Uninhabited
 	habitable 10000
 	haze _menu/haze-full
-	arrival 500 jump
+	arrival 500
 	hazard "Black Hole Gravity" 1
 	hazard "Black Hole Event Horizon" 1
 	hazard "Black Hole Accretion Disk" 1
--- a/data/map.txt
+++ b/data/map.txt
@@ -21145,6 +21145,7 @@ system "Sagittarius A*"
 	government Uninhabited
 	habitable 10000
 	haze _menu/haze-full
+	arrival 500 jump
 	hazard "Black Hole Gravity" 1
 	hazard "Black Hole Event Horizon" 1
 	hazard "Black Hole Accretion Disk" 1
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1464,7 +1464,12 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 			Point target;
 			// Except when you arrive at an extra distance from the target,
 			// in that case always use the system-center as target.
-			double extraArrivalDistance = currentSystem->ExtraArrivalDistance();
+			double extraArrivalDistance = 0;
+			if(isUsingJumpDrive)
+				extraArrivalDistance = currentSystem->ExtraJumpArrivalDistance();
+			else
+				extraArrivalDistance = currentSystem->ExtraHyperArrivalDistance();
+			
 			if(extraArrivalDistance == 0)
 			{
 				if(targetPlanet)
@@ -1482,7 +1487,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 			
 			if(isUsingJumpDrive)
 			{
-				position = target + Angle::Random().Unit() * 300. * (Random::Real() + 1.);
+				position = target + Angle::Random().Unit() * (300. * (Random::Real() + 1.) + extraArrivalDistance);
 				return;
 			}
 			
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1464,11 +1464,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 			Point target;
 			// Except when you arrive at an extra distance from the target,
 			// in that case always use the system-center as target.
-			double extraArrivalDistance = 0;
-			if(isUsingJumpDrive)
-				extraArrivalDistance = currentSystem->ExtraJumpArrivalDistance();
-			else
-				extraArrivalDistance = currentSystem->ExtraHyperArrivalDistance();
+			double extraArrivalDistance = isUsingJumpDrive ? currentSystem->ExtraJumpArrivalDistance() : currentSystem->ExtraHyperArrivalDistance();
 			
 			if(extraArrivalDistance == 0)
 			{
--- a/source/System.cpp
+++ b/source/System.cpp
@@ -319,7 +319,7 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
 			for(const DataNode &grand : child)
 			{
 				if(grand.Size() < 2)
-					child.PrintTrace("Warning: Keywords below arrival always need a value behind the keyword:");
+					grand.PrintTrace("Skipping distance specification without jump type identifier:");
 				else if(grand.Token(0) == "link")
 					extraHyperArrivalDistance = grand.Value(1);
 				else if(grand.Token(0) == "jump")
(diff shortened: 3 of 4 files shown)


source/Ship.cpp
```diff
@@ -1464,7 +1464,12 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 			Point target;
 			// Except when you arrive at an extra distance from the target,
 			// in that case always use the system-center as target.
-			double extraArrivalDistance = currentSystem->ExtraArrivalDistance();
+			double extraArrivalDistance = 0;
+			if(isUsingJumpDrive)
+				extraArrivalDistance = currentSystem->ExtraJumpArrivalDistance();
+			else
+				extraArrivalDistance = currentSystem->ExtraHyperArrivalDistance();
```
Amazinite, line 1471:
```suggestion
			double extraArrivalDistance = isUsingJumpDrive ? currentSystem->ExtraJumpArrivalDistance() : currentSystem->ExtraHyperArrivalDistance();
```

source/System.cpp
```diff
@@ -310,7 +310,35 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+			}
+			if(loadHyperDistance)
+				extraHyperArrivalDistance = child.Value(valueIndex);
+			// We only load additional jump distances if they are above 0.
+			// Jump drives use a circle around the target for targeting,
+			// so a value below 0 doesn't have a meaning in the same way
+			// as that a hyper-jump-distance negative values would be beyond
+			// the target.
+			if(loadJumpDistance && child.Value(valueIndex) >= 0)
+				extraJumpArrivalDistance= child.Value(valueIndex);
```
Amazinite, line 340:
```suggestion
				extraJumpArrivalDistance = child.Value(valueIndex);
```

source/System.cpp
```diff
@@ -310,7 +310,35 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+						child.PrintTrace("Warning: Unknown arrival type \"" + child.Token(i) + "\" for arrival distance:");
+				}
+			}
+			if(loadHyperDistance)
+				extraHyperArrivalDistance = child.Value(valueIndex);
+			// We only load additional jump distances if they are above 0.
+			// Jump drives use a circle around the target for targeting,
+			// so a value below 0 doesn't have a meaning in the same way
+			// as that a hyper-jump-distance negative values would be beyond
+			// the target.
```
Amazinite, line 338: I imagine the behavior of a negative value is just the same as the same magnitude positive value, so why explicitly prevent the use of negative values if the end behavior doesn't actually cause any issues? Seems to me like it'd make more sense to just allow both.

Amazinite (changes requested):
👍 

Mind adding `arrival 500 jump` to Sag A* in this PR? It's the only system I was planning to add it do, and since it's just a one line change then might as well deal with it here.

source/System.cpp
petervdmeer: The behaviour for Hyperdrives is very different, in that a negative value will cause the ship to overshoot the target.

What I can do for Jump-drives is to indeed make negative and positive values act in the same way.

petervdmeer: Updated to allow negative values, but convert them to positive internally (so that the ring width calculation is not affected).

data/map.txt
```diff
@@ -21145,6 +21145,7 @@ system "Sagittarius A*"
 	government Uninhabited
 	habitable 10000
 	haze _menu/haze-full
+	arrival 500 jump
```
tehhowch, line 21148: To me, a more natural syntax would be `[<descriptor>] <#value>`, or a child node sequence e.g.

```sh
arrival [<#default value for all types>]
  <type> <#value for just this type> # <type> may be e.g. "link", "jump", "planet"
  ...
```

Changing to this descriptor-first format can likely simplify the parsing code, too.

source/System.h
```diff
@@ -213,10 +215,13 @@ class System {
...
-	// target (system center or planet) when entering this system through a
-	// hyperspace link.
-	double extraArrivalDistance = 0.;
+	// The amount of additional distance that ships will arrive away from the
+	// system center when entering this system through a hyperspace link.
+	double extraHyperArrivalDistance = 0.;
+	// The amount of additional distance that ships will arrive away from the
+	// system center when entering this system through a jumpdrive jump.
+	double extraJumpArrivalDistance = 0.;
+
```
tehhowch, line 224:
```suggestion
```

data/map.txt
petervdmeer: The keywords/syntax used in PR are based on the descriptions in this [RFC](https://github.com/EndlessSkyCommunity/endless-sky/pull/51). (Sorry that I forgot to include the link.)

Should we discuss the syntax in the RFC PR instead of here?

petervdmeer: Updated to the child node sequence syntax.

source/System.cpp
```diff
@@ -310,7 +310,29 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+				if(grand.Size() < 2)
+					child.PrintTrace("Warning: Keywords below arrival always need a value behind the keyword:");
+				else if(grand.Token(0) == "link")
+					extraHyperArrivalDistance = grand.Value(1);
+				else if(grand.Token(0) == "jump")
+					// Negative jump distances work the same as positive jump
+					// distances.
+					// Jump drives use a circle around the target for targeting,
+					// so a value below 0 doesn't have the same meaning as for hyper
+					// drives where negative values would be beyond the target.
```
Amazinite, line 330: I think it makes more sense to move this comment to the first occurrence of setting this variable, i.e. at line 317.

Amazinite (approved):
Aside from moving a comment, looks good to me. The syntax now seems more appropriate.

source/System.cpp
tehhowch: I would remove these comments entirely, and include this clarifying information in the associated header's variable declarations.

tehhowch (commented):
minor comments

source/System.cpp
```diff
@@ -310,7 +310,29 @@ void System::Load(const DataNode &node, Set<Planet> &planets)
...
+				extraHyperArrivalDistance = child.Value(1);
+				extraJumpArrivalDistance = fabs(child.Value(1));
+			}
+			for(const DataNode &grand : child)
+			{
+				if(grand.Size() < 2)
+					child.PrintTrace("Warning: Keywords below arrival always need a value behind the keyword:");
+				else if(grand.Token(0) == "link")
+					extraHyperArrivalDistance = grand.Value(1);
+				else if(grand.Token(0) == "jump")
```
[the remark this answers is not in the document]
petervdmeer: Your suggestion looks nice, I'll use it.

PR #5580, opened 2020-12-28
Tweak: Giving the Remnant Engines a Niche

### LINKED ISSUE #5579: Request: Giving Remnant Engines a Niche
**Content (Tweaks)**

## Summary:
Remnant engines are significantly overshadowed by Korath, Deep Atomics, and Hai Atomic engines, all of which the player has easy access to by the time they reach the Remnant. Also overshadowed by the Wanderers, whose performance isn't as good but has other benefits.

This draft PR is for testing out some options for making the Remnant engines better in a way that gives them certain areas where they will be desireable to have, without significantly impacting the "raw power superiority" that the Korath and Atomics currently hold.

In this first iteration, this is accomplished by giving Remnant engines both slow protection and slow resistance. The former reduces how much slow damage they take, and the later speeds up how fast they dissipate it. 

## Problem:
Remnant engines have long been avoided within the player-base due to the fact that by the time the character has access to them, they certainly have access to and are probably used to using Deep and/or Hai atomics. About the same time period that they gain access to Remnant thrusters and steering, they also are exposed to all the Korath stuff, which is basically the top of our engine charts. 

Now, Remnant engines aren't actually bad. In terms of thrust/space, they are fairly high on the chart: They fall behind three steps of Korath stuff, the top two Hai thrusters, and the best Deep thruster. The Smelter is 8th in terms of thrust/space, and the top spot is only about 28% more thrust/space. (and the Remnant thrusters are clustered fairly tightly together in 8, 10, 13, and 35th place. I'm basically ignoring the anvil for this, because it is a combined outfit.)

In terms of total thrust, however, the smelter drops down to 9th, and the distance between it and first expands significantly (first place has basically +100% the thrust of the Smelter). On top of that, the spread of Remnant engines expands a lot, so they now occupy 9th, 19th, 29th, and 32nd. 

So the competition basically boils down to this:
1) If you prioritize acceleration and top speed above all else, then:
- If you have space for it, Korath are the best. (needs a lot of cooling, but again, you have space, right?)
- Next up is Deep and Hai that also need a fair bit of space, but not as much as Korath stuff and much less cooling)
2) If you are making a more general purpose ship, then:
- Wanderer thrusters aren't great in terms of total thrust or their direct space efficiency; but the fact that they *provide* cooling instead of requiring it gives them a strong edge in that regards anyway.
- Using smaller Korath or atomics (due to their high thrust/space) are often useful.

Caveats:
- I don't particularly want to make any large changes to the balance between engines. 
- None of the changes should make the Remnant an "all round best" engine.
- Having a strong utilization case for a limited or specific use and/or situation would be ideal. (In other words, being great at something fairly specific while still being "average" for most other things)

**Related Issue Links**
Particularly relevant is Amazinite's overview of Thrusters provided as part of a "best of" list: https://github.com/endless-sky/endless-sky/issues/2852#issuecomment-318910181
Discussion of Reverse Thrusters: https://github.com/endless-sky/endless-sky/issues/3252
An idea for tweaking Remnant thrusters by making them much more efficient but also giving them cooling inefficiency which makes them problematic to stack: https://github.com/endless-sky/endless-sky/issues/4688

The Remnant engines to receive some kind of buff sufficient to make them desirable in at least a few specific use cases, probably ones related to either:
a) Use in the Ember Waste's newly implemented hazards environment, or 
b) Long-term flight endurance.

**Describe ideas you've considered**
1. The Remnant value mobility. Better to not be hit than to be able to tank the hit. They are very familiar with weapons that impede mobility, ergo, maybe they have developed their engines to resist that impediment and/or recover from it more quickly?
2. Alternatively, what if Remnant engines were capable of operating in reverse of significant strength? Basically, every engine includes a built in thrust-reversal system.
3. Continuing that line of thought, what if their afterburner - instead of enabling them to leap forward, actually was built to accelerate them backwards? Thus they could fly forward to attack, then leap backwards out of the way of danger. It'd be even more interesting/effective if it applied thrust at right angles to the ship's axis. As in, the 'afterburner' pushed the ship sideways in a strafing motion.
4. A significant hazard in the Ember Waste is the ion storms, and a secondary hazard is heat. (the latter often magnified by the former disabling active heat systems). Now, Wanderers pretty much have a lock on "engines that dissipate heat instead of make it," so what about having engines that can run on radically less energy? The less energy they use, the less likely they are to overload the ship systems during an ion storm, and the less Remnant mobility will impacted. 
5. Other perks? For instance, giving Remnant engines built-in ramscoops, minor amounts of hull and/or shield repair, fuel generation, etc.
6. Giving them emergency motive power? Not sure if it could be implemented on the same outfits, but basically giving Remnant engines a second "thrust" value that is a tiny percentage of its normal value, but doesn't use any energy at all to operate. Thus even when they are completely drained of energy they'd still be able to effectuate a very slow escape.

Zitchas:
**Content (Tweaks)**

## Summary:
Remnant engines are significantly overshadowed by Korath, Deep Atomics, and Hai Atomic engines, all of which the player has easy access to by the time they reach the Remnant. Also overshadowed by the Wanderers, whose performance isn't as good but has other benefits.

This PR is for testing out some options for making the Remnant engines better in a way that gives them certain areas where they will be desirable to have, without significantly impacting the "raw power superiority" that the Korath and Atomics currently hold.

## Lore:
The Remnant hold the philosophy that survival is paramount over all else. It is *much* better to run away if necessary rather than stay and fight to the death. Their civilization began by running away from a war, and continues to prioritize escape over all else. This is shown in how much energy they devote to engines, as well as cloaks.

Just as Remnant prioritize running away, they are also the most energetic of all factions in regards to exploration and reverse-engineering, particularly with an eye to ensuring they know all the potential escape routes and what threats might be encountered along them. In this case, two of the longer range escape routes are currently up through Korath space, and South-West into Coalition space. Both of these areas have enemies with slowing weapons. 

More importantly, though, is that they are realistic enough to acknowledge that if they can steal weapons from someone, chances are others can to. It would be foolish to assume that any particular tech or ability is going to remain the unique purview of any one group. As such, while having the ability to outright nullify/ignore slowing effects is not a high priority for them, incorporating techniques for recovering from it faster than normal most definitely is. This is basically the equivalent of keeping your ship running at 0 heat even if you don't expect to be fighting enemies with heat based weapons. They know how at least one kind of slowing effects work, and they adjust their engine designs accordingly. 

## Current changes:
Thrusters:
- All thrusters have built-in reverse thrusters equal to 2/3 their regular thrust strength.
- Small amount of slow resistance (range from 0.02 for the anvil, 0.01 for the crucible, and 0.04 for the smelter)

Steering:
- Increased turn speed by 10% by Amazinite's suggestion
- Tiny amount of slow resistance (range from 0.005 for the crucible up to 0.02 for the smelter. Half the thruster values)

In terms of numbers if a ship with a set of smelters got hit with a single shot from 4 inhibitors, that'd give 2 slowing. It would then take that ship 34 frames to fully dissipate the slow from that single burst. 
If the same ship got hit with a single shot from a detainer (which gives 40 slowing), it would take 667 frames to fully dissipate the shot (that's just over 6.5 seconds). This does not seem unreasonable to me.
Given that both weapons actually fire quite quickly and single shot scenarios are rare, these numbers basically translate to the Remnant not building up slowness as quickly, and dissipating it faster afterwards. They'll still suffer from it while it is actually hitting them.

## Related PR:
This PR directly addresses and implements one potential solution for Issue #5579 
Fixes #5579 

## Save File
 - Any pilot with a basic Remnant license can test this content.

## PR Checklist


Amazinite:
As with #5558, it should be either protection or resistance, not both. As far as the slowing attributes themselves go though, I'm not sure it fits too well. It does go along with the idea of the Remnant valuing mobility, but I don't think it would make sense for them to spend energy on a characteristic that will be seldom used by them, unless they happen to be fighting the Sestor a lot. So I'd drop those changes entirely.

The in-built reverse thrusters though, I'm definitely on board with. How about going through the Remnant ships and giving them small reverse engine flares like you did with the Penguin to fit the theme? Might also make sense to ditch the in-built reverse thrust on the Penguin though if all their engines will be getting them.

Zitchas:
I can do small reverse thrust flares, but I'm hesitant because when the reverse thrust mechanics were added and I did the penguin there was a fair amount of complaints about some people not wanting the extra activity of flares showing up everywhere. 

Now, the tiny steering flares I did for the penguin? I like those. Nice and faint, can be ignored. But making reverse thruster flares that small doesn't really fit.

data/remnant/remnant outfits.txt
```diff
@@ -529,6 +542,10 @@ outfit "Smelter-Class Thruster"
 	"thrust" 76.8
 	"thrusting energy" 10.10
 	"thrusting heat" 10.2
+	"reverse thrust" 46.08
+	"reverse thrusting energy" 6.06
+	"reverse thrusting heat" 1.5
+	"slowing resistance" 2.0
```
Amazinite, line 548: We just got through how resistance values like this for ionization result in being able to completely ignore large amounts of ionization. You should REALLY start using much smaller resistance values whenever you add them to stuff. (This value means that a single Detainer's slowing will be completely removed after only 35 frames from when the first shot in a burst hits. That's just over half a second. And if you look at having TWELEVE Inhibitors, this is enough resistance to bleed off the slowing in 3 frames.)

Whenever using a resistance, you should be looking at the values of the weapons that use the related status effect. For example, look at the Inhibitor and Detainer for slowing damage. The Inhibitor deals 0.5 damage while the Detainer deals 40. A resistance of 2 means a single Inhibitor never does anything to you, while a Detainer only mildly inconveniences you for 20 frames, since resistance is done every frame.

I'm still not on board with slowing resistance on Remnant engines anyway, though.

Amazinite (changes requested):

data/remnant/remnant outfits.txt
Zitchas: Reduced the values to 1/5 of this. That being said, a large engine from the faction that is a) devoted to running away, and b) does have an understanding of how slowing effects work? To me, that makes perfect sense that such an engine would be able to shrug off a single inhibitor shot without effect. They probably get  hit by them all the time during training exercises and combat training. Makes a lot of sense that they'd build their stuff to be resistant to their own stuff.

Anyway, reduced that from 2 down to 0.4.

data/remnant/remnant outfits.txt
```diff
@@ -580,11 +599,16 @@ outfit "Smelter-Class Steering"
 	"turn" 1980.0
 	"turning energy" 5.21
 	"turning heat" 6.2
+	"slowing resistance" 0.4
```
Amazinite, line 602: Still seems way too high to me (Smelter thrusters + steering = bleed all slowing damage from 4 Inhibitors in 3 frames, Crucible thrusters + steering = bleed all slowing from 4 inhibitors in 9 frames), and still also don't want it.

Let's just drop slowing resistance on all the engines. Then just so the steering aren't losing out by not having reverse thrust like the thrusters do, how about we instead bump their turn rates a bit? An increase in turn for all steering by just 10% (no change to any other stats) would mean that Remnant steering...
* would be the BEST purchasable steering in terms of heat production vs turn rate, beat only by Pug steering, which can not be purchased.
* would become a little better on energy drained vs turn rate, as right now they're already the worst engines on that front, but this change would make them a little more bearable.
* would have the best small steering, best medium steering, and second best large steering in terms of turn rate vs mass.

Amazinite (changes requested):

data/remnant/remnant outfits.txt
```diff
@@ -478,6 +479,10 @@ outfit "Anvil-Class Engine"
 	"turn" 336.0
 	"turning energy" 1.05
 	"turning heat" 1.1
+	"reverse thrust" 8.1
+	"reverse thrusting energy" 1.23
+	"reverse thrusting heat" 1.17
```
Amazinite, line 484: Even without adding the flare points on Remnant ships, you still need a reverse flare sprite and reverse flare sound.
```suggestion
	"reverse thrust" 8.1
	"reverse thrusting energy" 1.23
	"reverse thrusting heat" 1.17
```

data/remnant/remnant outfits.txt
Zitchas: OK, I'll take a look at making some changes tomorrow. Would it be possible for you to share that spreadsheet you were showing in the livestream in your folder of spreadsheets? 

I still want them to have some slow resistance, but I also want it to be reasonable. And your numbers there make it sound like it is still too high.

data/remnant/remnant outfits.txt
Zitchas: Alright, I'll do that next.

data/remnant/remnant outfits.txt
Zitchas: Alright. Reduced the slow resistance to 10% of those values. So assuming your math is correct (which I assume it is), bleeding that slow will take 90 frames instead of 9. And increased the steering values by 5%.

data/remnant/remnant outfits.txt
Zitchas: Done. Now to test.

data/remnant/remnant outfits.txt
```diff
@@ -476,12 +477,19 @@ outfit "Anvil-Class Engine"
 	"thrust" 13.5
 	"thrusting energy" 2.05
 	"thrusting heat" 1.95
-	"turn" 336.0
+	"turn" 352.8
 	"turning energy" 1.05
 	"turning heat" 1.1
+	"reverse thrust" 8.1
+	"reverse thrusting energy" 1.23
+	"reverse thrusting heat" 1.17
```
Amazinite, line 485: You're still missing the reverse flares and sounds.
And I'm still not interested in slowing resistance on these engines.

Amazinite (changes requested):

data/remnant/remnant outfits.txt
Zitchas: Sorry, for some reason forgot that. Will be done momentarily.

The Remnant are paranoid about being able to escape. They got slow generating weapons from other aliens. If they can do it, best to assume that others can too. Either way, the amount of resistance is small but still enough to give a surprise edge in escaping from things using slow weapons against them. (0.04 for the smelter)

data/remnant/remnant outfits.txt
```diff
@@ -476,12 +477,22 @@ outfit "Anvil-Class Engine"
 	"thrust" 13.5
 	"thrusting energy" 2.05
 	"thrusting heat" 1.95
-	"turn" 336.0
+	"turn" 369.6
```
Amazinite, line 480: I suggested the increased turn speed as an alternative to the slowing resistance, so if you aren't going to budge on the slowing resistance then revert the turn speed changes on the engines.

Amazinite (changes requested):

data/remnant/remnant outfits.txt
Zitchas: Right now, there are two changes to Remnant steering. The buff to turning speed, and the addition of slowing resistance. As things stand right now, the buff to turning is a solid performance increase; while the slowing resistance is basically just flavor. 

One of the big reasons for having that slowing is to hint at the fact that the Remnant are preppers. As in, those survivalists who prep for even unlikely things. 
If the Remnant steering honestly could use a buff, then the buff to the steering should go through on its own merits.
As a separate issue, whether or not they have any slowing resistance should basically be a question of "do we want them to have this flavor or not." It is not, after all, a powerful or particularly useful thing, right now. But even a tiny amount of slowing resistance does mean that they have something that no other family of engines has. 

According to the math, a single inhibitor canon does 230.8 slowing damage per second, and the current smelter steering will dissipate 1.2 slowing per second. Which suggests to me that the slowing resistance there might have some benefit in helping a ship recover after it has gotten out of the line of fire, but probably won't even be noticeable while under the effect of it. Which kind of reinforces the fact that it is basically a lore / flavor thing, not a balance thing. Meanwhile, the Heliarch Repulsor does 600 slowing per second, and the Detainer does 4000 slowing damage per second. So the Steering's slowing resistance is even more "pure flavor" against them.

Zitchas:
Amazinite's explanation on Discord about the revised resistance values:

> Crucible steering: 0.008
> Crucible thrusters: 0.012
> About 3% lower max slowness on a Crucible set, decays to 0 after 209 frames while no resistance is still at 90% speed.
> 
> Forge steering: 0.016
> Forge thrusters: 0.024
> About 7% lower max slowness on a Forge set, decays to 0 after 140 frames while no resistance is still at 83% speed.
> 
> Smelter steering: 0.032
> Smelter thrusters: 0.048
> About 15.5% lower max slowness on a Smelter set, decays to 0 after 71 frames while no resistance is still at 71% speed.

Amazinite (approved):

PR #5581, opened 2020-12-29
Adding to the Inhibitor's description

Zitchas:
**Content (Tweak)**

## Summary
Inhibitors currently do .5 slowing damage per shot. In testing, VitalChip found that with their modded values, it took 4 fully loaded Albatrosses (24 inhibitor cannons and 16 turrets) to almost stop a SB. (that was using values of .8 slow for canons and 1.6 for turrets)

Given that the game is supposed to be balanced for solo fleets, this means that with the current value it is functionally impossible for a solo ship to inflict useful amounts of slowing on an enemy.

~~As such, this PR raises the "per shot" slowing damage value for the Inhibitors up to 2. This is still a very weak slowing effect, but should be enough that players can actually use it as a tactical weapon for controlling the flow of battle.~~

~~edit. With Amazinite's numbers, now reduced the value from 2 down to 1.~~
~~edit. Down to 0.9 after testing against a solo Raider.~~

This PR simply adds a brief snippet of description to the canon, clarifying that while they acquired the weapon for its interesting effect; they eventually decided to rebuild it with a view to making a longer ranged support weapon to complement the Thrasher instead of focusing on the slowing effect.

## Save File
Any pilot with access to Remnant tech will be able to test this. 

## Related Issues:
The PR that added the Inhibitor turret, and the only time I've seen someone mention inhibitor's slowing being more noticeable than heat damage. https://github.com/endless-sky/endless-sky/pull/4920

## PR Checklist


tehhowch:
"Slowing" and "nearly immobilizing" are different goals. And why compare Remnant weapons' slowing impact to a ship the Remnant have never encountered before? The natural targets for comparison should be the Korath Chaser and Raider.


If the slowing is to be increased _at all_ the already-too-good inhibitors must be made slower, larger, and lower-ranged.

Amazinite:
Slowing damage multiplies the turn rate and acceleration of a ship by `1 / (1 + .05 * slowness)`. A single inhibitor cannon hitting a target as fast as possible will within 11 seconds max out a target at about 4.04 slowness. (By this point the slowness applied vs the slowness bled equals out.) That means a single inhibitor cannon is enough to slow any ship down to 82% its max acceleration and turn rate in only 11 seconds. Four inhibitor cannons (equal to the slowing damage of just one with this PR) in that same time results in a slowness of 16.15, which causes a ship to have 55% the acceleration and turn rate. That seems plenty powerful enough to me as is, but making just one inhibitor cannon deal that much slowness is overkill. Just a single 16 ton weapon shouldn't be enough to nearly half the maneuverability of a ship.

Zitchas:
Already too good? hardly. 

Fair, the best target to balance against is the Raider.

Admittedly, immobilizing a ship should require a lot of work. But significant and noticeable slowing should be possible without needing to bring a fleet to bear on a target.

Thank you for the numbers, Amazinite. Based on that, 2 does sounds too powerful, so I'll drop it down to 1.

That being said, what you say sounds pretty good - for a duel. In ES we very rarely ever have duels. Combat is usually a large mass of ships thrashing around where it is challenging to maintain fire on a single target for significant lengths of time. Virtually the only time I have duels is against random raider spawns or wrapping up the tail end of battles when I'm finishing off lone survivors. Not the time when finesse and tactical manoeuvring are really paramount anymore.

Zitchas:
Actually, an alternative to boosting the Inhibitor would be decreasing the "natural" bleed rate at which slowing is lost. Reducing that by 25-50% would be an acceptable alternative. Actually, that'd be more fun in some respect, since that means that weaving in and out of battle could allow a player to gradually "build up" the slowness on a target. That'd make for a fun mechanic, as a target ship is gradually slowed compared to their escorts / rest of the fleet, and gradually falls farther out of it and more vulnerable to being picked off. This is basically what I was hoping to do with an increase in the inhibitor's slowing ability. But if simply reducing the rate at which people lose slowness is an alternative, then that'd be even better.

vitalchip:
Leaving aside the figures for a moment, the Inhibitor just doesn't _feel_ like it's doing much slowing. Technically a ship with 6 cannon and 4 turrets is able to reduce the thrust on an enemy ship down to about 25% but that's if you can keep continuos fire on it and even at that, the opposing ship doesn't seem terribly affected. 

Compared to something like the Ion cannon it's 'undramatic'  Pitching a Shield Beetle with 4 ion cannons against an Albatross with 6 Inhibitors and 4 Inhibitor turrets.. the Shield Beetle is slowed but still well able to advance while the Albatross is very quickly stripped of energy and rendered quite helpless. As 'party trick' weapons on either side the Inhibitor feels like a poor relation, particularly when considering that the ship has a higher overall mass of weaponry, much higher firing energy and many more hardpoints required.

I would suggest making the Inhibitor larger, slower firing and heavier hitting per shot but that would cause issues for the smaller ships currently using them. 

One thing I would not consider is reducing the range. We have quite enough poor range primaries in the game already without hobbling one of the few exceptions.

Possible solutions?

1. Capital ship version, larger, slower firing and harder hitting on each shot. (I currently run 3 versions.. Capital/Standard/Drone)

2. Trade some damage/hit force for a little more slowing on the existing weapon. (There isn't a lot of damage to play with though)

3. Increase the slowing significantly but change the bleed off so that there is a more pronounced slowing when the weapon hits but over a period of time the overall slowing is comparable to now. (Will  affect other weapons but easily remedied)

There's nothing intrinsically wrong with the Inhibitors, they're considerably better than many people seem to think (once turrets are available) but they just lack a bit of drama which I like to see from 'speciality' weapons.

Zitchas:
What these tests are making clearer is that it isn't actually the amount of slowness that the inhibitors impart that is the problem. That's actually probably fine. The problem is that slowness dissipates too quickly, meaning that in the course of a real battle where contact is made and lost with the enemy on a regular basis, it feels that the slowness practically vanishes almost immediately after contact is lost. This is ideal behaviour for a ship stuck in a tractor beam, where the effect lasts as long as it is in the beam. It is not good behaviour for an effect based on a projectile weapon where it is expected that the effect lasts long enough to be tactically significant.

This gives rise to a few ideas:
1) Would it be possible to have an "instant" attribute on weapons (I could see this applying to slow, heat, and ionization) that basically specifies that this damage only lasts while the ship is being hit, and vanishes when the ship is no longer being hit?
2) Would it be possible to either 
   i) reduce the drain rate for slow in general, 
   ii) make it so that the slow from projectiles like the inhibitor drain more slowly than other forms of it, or 
   iii) implement a new kind of slow that drains slower. (so we'd have both a fast dissipating slow and a slow dissipating slow.)

Option 1) would make for some interesting tactical possibilities, such as using tactical disruption ships that don't actually do damage, but have a "instant slow" weapon that would act kind of like the tackler ships do in EVE-online. Similar but more powerful effects could be achieved using instant-heat or instant-ion. It'd make for interesting progression seeing the Remnant use instant-slow to tackle ships, while the Quarg use instant-ion (and it'd make the Heliarch fear of ionization more understandable/reasonable). Perhaps pug or golden-age Korath (aka the high tier ones from the height of their empire) might use instant heat.
Of course, none of theses weapons should actually max out the relevant stat. That would be OP. But having a beam that basically gave someone a static +500 heat (not +500/sec. Just whatever their normal heat is +500, and that lasts as long as the beam is in contact, and vanishes as soon as it is gone.) Cold ships shrug it off, warm ships get pushed up enough that a pilot has to be careful to avoid other sources of heat. Hot ships get paralyzed.

Amazinite:
Copying from Discord.

>Important to note is that the slowing damage of Inhibitors is really a side effect. It's practically a buffed particle cannon. The slowing damage isn't its main draw or purpose, so they shouldn't be looked at from the perspective of "They aren't immobilizing my target? Must be bad."
>
>If your goal is to basically immobilize a target, that's where the Detainer comes in. The Detainer deals trash shield and hull damage for its size, but has considerable amounts of slowing damage.
>
>So the purpose of the Detainer is for slowing targets. That purpose is not shared with the Inhibitor, whose purpose is damage dealing with slowing on the side.
>
>A similar comparison might be between the Korath Repeater and the Flamethrower. The Repeater deals lots of damage while also having some heat. The Flamethrower deals little damage while outputting a lot of heat. You wouldn't say buff the Repeater's heat damage because it isn't all that impactful, because the heat damage isn't the main purpose of the weapon.

TL;DR you're trying to shove the Inhibitor into a role it wasn't designed for and getting bad results because of that.

Zitchas:
Thanks. And as several people commented after that, the counter response is basically:
"The entire description talks specifically about the slow effect. Makes it abundantly clear that the *slow* was the whole point of the effort. 

But I guess it is a moot point: If both you AND Tehhowch are both against it, it ain't happening. So I'll revamp this PR to change the description a bit so that the slow isn't the focal point anymore and solve the issue by changing lore instead.

Zitchas:
In response to all the feedback from the developers, this PR has been changed. It no longer changes any of the slowing values, and simply adds a clarifying statement to the Inhibitor Cannon explaining why it doesn't have the amount of slowing that many people expect it to have. Quite simply: The Remnant decided to focus on shaping it into a longer range support weapon to complement the thrasher rather than focus on the slowing aspects. 

Given the community comments, while the developers have valid points about balance and how good it performs numerically, there is a broad perception that "slowing" is the focus of this weapon, a task at which it is disappointing. This PR now addresses this by expanding the description slightly to state that the Remnant chose to develop a longer range support weapon instead of focusing on the slow effect. This should help people have a better idea of what sort of weapon they are getting when they buy an inhibitor.

data/remnant/remnant outfits.txt
```diff
@@ -40,6 +40,7 @@ outfit "Inhibitor Cannon"
 		"hit force" 39
 		"slowing damage" .5
 	description `Nearly a century ago, a Remnant scout exploring space north of the Ember Waste came under fire from an alien fleet armed with weapons that interfered with the scout's engines. The story says that the Remnant captain calmly told her crew, "We aren't going home until we steal that weapon." That alien technology became the basis of the Inhibitor Cannon.`
+	description `	While the Remnant were intrigued by the metaphysics that enabled the slowing effect, they decided that they needed to focus on developing a longer ranged weapon to compliment the close-quarters Thrasher. As such, while the inhibitor has but a mere shadow of its inspiration's warping effects, it is a very effective weapon.`
```
Amazinite, line 43: At this point you're downplaying the slowing effect, which, again, I think is a lot stronger than people are giving it credit for, even if it doesn't immobilize the target.
```suggestion
	description `	Although not powerful enough to fully immobilize a ship, the Inhibitor is still a formidable weapon. The Inhibitor is excellent at controlling engagement distances, keeping targets at range by both pushing them backward with the impact of each shot while also slowing their approach by interfering with their engines.`
```

Amazinite (changes requested):

data/remnant/remnant outfits.txt
```diff
@@ -40,6 +40,7 @@ outfit "Inhibitor Cannon"
 		"hit force" 39
 		"slowing damage" .5
 	description `Nearly a century ago, a Remnant scout exploring space north of the Ember Waste came under fire from an alien fleet armed with weapons that interfered with the scout's engines. The story says that the Remnant captain calmly told her crew, "We aren't going home until we steal that weapon." That alien technology became the basis of the Inhibitor Cannon.`
+	description `	Although the Remnant appreciated the original weapon, they were disappointed that it was too big to fit on most of their ships. Instead of letting their effort go to waste, they instead opted to put their newfound knowledge to use creating a new weapon suitable for their more nimble craft. Thus, the Inhibitor is excellent at controlling engagement distances, keeping targets at range by both pushing them backward with the impact of each shot while also slowing their approach by interfering with their engines.`
```
Amazinite, line 43: Dropping "new" because it's repetitive with "newfound" 
```suggestion
	description `	Although the Remnant appreciated the original weapon, they were disappointed that it was too big to fit on most of their ships. Instead of letting their effort go to waste, they instead opted to put their newfound knowledge to use creating a weapon suitable for their more nimble craft. Thus, the Inhibitor is excellent at controlling engagement distances, keeping targets at range by both pushing them backward with the impact of each shot while also slowing their approach by interfering with their engines.`
```

PR #5583, opened 2020-12-29
feat: Track child speeds to improve grouped fleet movement

petervdmeer:
**Feature:** This PR implements some support for features discussed in #3381 and #4471

## Feature Details
This PR adds functionality to parent ships to track the maximum speeds of their escorts and take those escort speeds into account in their behaviour.

Ships landing on a planet in a friendly environment will now use 90% of the speed of their slowest escort for landing.

#3381 is relevant since it deals with linger behaviour.
#4471 is relevant since formation flying require formation leaders to adjust their speeds to allow slower escorts to stay in formation.

## UI Screenshots
N/A

## Save File / Usage Example
The attached [save file](https://github.com/endless-sky/endless-sky/files/5752146/Fast.Leader.txt) can be used to test out this feature using the following test-sequence:
- Depart
- Slowly fly away from the planet.
- Land on auto-pilot (press L)

Without this PR (for example on master, commit-id 5bbe1d8c8dd21f53719482f1bf0e98aeb45f7923) you will notice your ship land as fast as possible.
With this PR, you will notice your ship land at a speed just under the speed of your slow escorts (Star Barges).

## Testing Done
I used the usage example above as a method of testing.

## Performance Impact
Some MoveToPlanet() functions got some additional calculations to decide which speed to use during movement.


tehhowch (changes requested):
In addition to the comments, this will need to account for carried ships, which also use the SetParent method when launching or loading.

source/AI.cpp
```diff
@@ -1620,7 +1632,9 @@ void AI::MoveEscort(Ship &ship, Command &command) const
 	{
 		ship.SetTargetSystem(nullptr);
 		ship.SetTargetStellar(parent.GetTargetStellar());
-		MoveToPlanet(ship, command);
+		// Escorts always move at the fastest speed possible in service of their parent.
+		double cruiseSpeed = 0.;
+		MoveToPlanet(ship, command, cruiseSpeed);
```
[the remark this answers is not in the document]
petervdmeer: Yes, this PR currently only changes the behaviour of the parent. The escorts both in master as well as currently in this PR just fly to the planet as fast as they can when the parent is landing.

I agree that it would seem better to have the escorts stick to the parent. I'll have a look if changing the escorts behaviour could be combined into this PR or if it is better to create a separate PR for that (the code overlap looks minimal).

Amazinite:
It's permissible to only launch with reverse thrusting, but Ship::MaxVelocity, which Ship::CruiseVelocity uses, doesn't check reverse thrust. What's the behavior of having an all-reverse-thrust escort?

source/Ship.cpp
```diff
@@ -3510,8 +3530,22 @@ void Ship::RemoveEscort(const Ship &ship)
...
+	// be able to speed up.
+	if(ship.MaxVelocity() == escortsVelocity)
+	{
+		escortsVelocity = MaxVelocity();
+		it = escorts.begin();
+		for( ; it != escorts.end(); ++it)
+		{
+			Ship *escort = it->lock().get();
+			if(escort)
+				escortsVelocity = fmin(escortsVelocity, escort->MaxVelocity());
```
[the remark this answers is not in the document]
petervdmeer: `min` it is then, since `min` is also used all over the rest of the codebase.

source/Ship.cpp
```diff
@@ -3510,8 +3530,22 @@ void Ship::RemoveEscort(const Ship &ship)
...
 			escorts.erase(it);
-			return;
+			break;
 		}
+	
+	// If we remove the slowest esort from a fleet, then the fleet might
+	// be able to speed up.
+	if(ship.MaxVelocity() == escortsVelocity)
+	{
+		escortsVelocity = MaxVelocity();
```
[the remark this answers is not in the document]
petervdmeer: Good catch! This was an issue from some refactoring. I'll fix this.

source/Ship.cpp
```diff
@@ -2981,6 +2981,19 @@ double Ship::MaxReverseVelocity() const
 
 
 
+// The optimal speed for this ship when not moving in a hurry.
+// If the ship has escorts, then the speed returned will allow for all escorts
+// to catch up with this ship.
+double Ship::CruiseVelocity() const
+{
+	if(escortsVelocity < 1.)
```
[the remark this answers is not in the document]
petervdmeer: Defaulted to -1.

source/AI.cpp
```diff
@@ -1789,7 +1805,13 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	if(!isClose || !isFacing)
 		command.SetTurn(TurnToward(ship, dp));
 	if(isFacing)
-		command |= Command::FORWARD;
+	{
+		// We set full forward power when we don't have a cruise-speed, when we are below
+		// cruise-speed or when we need to do course corrections.
+		bool facesMovementDirection = fabs((Angle(velocity) - angle).Degrees()) < 30.;
```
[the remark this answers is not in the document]
petervdmeer: I used the same implementation as isFacing was using a few lines up.

source/Ship.cpp
```diff
@@ -3510,8 +3530,28 @@ void Ship::RemoveEscort(const Ship &ship)
 		if(it->lock().get() == &ship)
 		{
 			escorts.erase(it);
-			return;
+			break;
 		}
+	
+	// If we remove the slowest esort from a fleet, then the fleet might
```
tehhowch, line 3536:
```suggestion
	// If we remove the slowest escort from a fleet, then the fleet might
```

source/Ship.cpp
```diff
@@ -2981,6 +2981,19 @@ double Ship::MaxReverseVelocity() const
...
 
+// The optimal speed for this ship when not moving in a hurry.
+// If the ship has escorts, then the speed returned will allow for all escorts
+// to catch up with this ship.
+double Ship::CruiseVelocity() const
+{
+	if(escortsVelocity < 0.)
+		return MaxVelocity();
+	else
+		return min(MaxVelocity(), 0.9 * escortsVelocity);
```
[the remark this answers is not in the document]
petervdmeer: Done (into a separate function.

source/Ship.cpp
```diff
@@ -3510,8 +3530,28 @@ void Ship::RemoveEscort(const Ship &ship)
...
+			break;
 		}
+	
+	// If we remove the slowest esort from a fleet, then the fleet might
+	// be able to speed up.
+	if(ship.MaxVelocity() == escortsVelocity)
+	{
+		escortsVelocity = -1.;
+
+		it = escorts.begin();
```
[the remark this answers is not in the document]
petervdmeer: I moved the full calculation (erase and tracking) into a single function.

source/AI.cpp
```diff
@@ -1521,7 +1521,16 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
 	}
 	else if(ship.GetTargetStellar())
 	{
-		MoveToPlanet(ship, command);
+		// If our allies are clearly stronger than our enemies in this system,
+		// then move at more civilized speeds when going to a planet.
+		// A cruiseSpeed of 0. is interpreted as "as fast as possible".
+		double cruiseSpeed = 0.;
```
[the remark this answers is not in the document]
petervdmeer: Done

petervdmeer:
> In addition to the comments, this will need to account for carried ships, which also use the SetParent method when launching or loading.

Carried ships are now handled by being ignored in the escorts velocity calculation.

petervdmeer:
> It's permissible to only launch with reverse thrusting, but Ship::MaxVelocity, which Ship::CruiseVelocity uses, doesn't check reverse thrust. What's the behavior of having an all-reverse-thrust escort?

I just did a short test on master (5bbe1d8c8dd21f53719482f1bf0e98aeb45f7923) with escorts with all-reverse thrusters; those escorts just drift away until they have a reason to do reverse thrust, after which they typically drift away even faster.

For this PR I added code to ignore escorts that report 0 MaxVelocity in the escort-velocity calculation to ensure that such escorts don't cause negative impact on the parents movement speed.

source/Ship.cpp
```diff
@@ -3511,46 +3508,55 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
-			if(escort)
-			{
-				if(escortsVelocity < 0.)
-					escortsVelocity = escort->MaxVelocity();
-				else
-					escortsVelocity = min(escortsVelocity, escort->MaxVelocity());
+		auto escort = it->lock().get();
+		if(escort){
+			if(escort == &ship){
+				escorts.erase(it);
```
tehhowch, line 3529: `it` is invalidated by `erase`, so incrementing after erasure is a bug. You'll need to use `it = escorts.erase(it);` to restore `it` to a valid state (which means you'll need to tweak your iteration strategy here, too).

source/Ship.cpp
```diff
@@ -3511,46 +3508,55 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+{
+	// Cache the maximum speed that the parent should stay below to keep
+	// all escorts together.
+	// We don't cache the speeds of escorts that have 0 velocity to avoid
+	// having the parent get stuck when one of the escorts doesn't have
+	// regular thrust.
+	// We also don't cache the speeds of carried ships, since they are often
+	// docked and the carrier waits for them during docking already.
+	double eV = ship.MaxVelocity() * 0.9;
+	if(!ship.CanBeCarried() && eV > 0.)
```
tehhowch, line 3554: move the definition of `eV` inside the if-block.

Note that you could also account for carried escort state in a similar manner to how we handle `carriedMass` -- update on launch & carry.

source/AI.cpp
```diff
@@ -1523,8 +1523,8 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
 	{
 		// If our allies are clearly stronger than our enemies in this system,
 		// then move at more civilized speeds when going to a planet.
-		// A cruiseSpeed of 0. is interpreted as "as fast as possible".
-		double cruiseSpeed = 0.;
+		// A cruiseSpeed of below 0 is interpreted as "as fast as possible".
```
tehhowch, line 1526:
```suggestion
		// A negative cruising speed indicates the ship should move as quickly as possible.
```

source/Ship.cpp
```diff
@@ -3498,20 +3508,56 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+				RegisterEscort(ship);
 		}
+		++it;
+	}
+}
+
+
+
+// Store relevant cached data for the given escort.
+void Ship::RegisterEscort(const Ship &ship)
```
tehhowch, line 3544: I'm not sold on the name, but 🤷‍♂️

tehhowch (changes requested):

source/Ship.cpp
petervdmeer: I already changed the loop to use `while` instead of `for` to allow the iterator itself to be replaced and changed the code not to advance the iterator on erase, but I indeed forgot to assign the return-value of the erase action.
Assignment is now also added. Do you see other issues with the loop as it is now?

source/Ship.cpp
petervdmeer: I agree the name is not great, but I didn't have inspiration yet to come up with something better.

source/Ship.cpp
petervdmeer: Moved the definition of eV.

I considered updating on launch & carry. For deploy this would be fine; if the deployed ship is slower than the current lowest escort-speed, then update the esocrt-speed to the lower value. For docking it could be a bit trickier; if you have a large number of slow carried escorts, then each docking ship could trigger a loop to determine the next slowest escort.

source/Ship.cpp
```diff
@@ -3498,20 +3508,56 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+			if(escort == &ship){
+				it = escorts.erase(it);
+				continue;
+			}
+			else
+				// Re-scan existing escorts to cache data like the
+				// escorts velocity.
+				RegisterEscort(ship);
 		}
+		++it;
```
tehhowch, line 3537: This will double increment on erasure, which means we will increment the `end()` iterator on removing the last escort, or skip the next ship in line.

(Minor style nit on the `if(escort){` too.)

petervdmeer: Isn't the double increment avoided by the `continue` statement on line 3530?

petervdmeer: I'll fix the style issue.

tehhowch: Ah, i missed that line somehow. Indeed, it is.

An alternate formulation that reduces some indentation would be

```c++
for(auto it = escorts.begin(); it != escorts.end(); )
{
  auto escort = it.lock().get();
  if(escort == &ship)
    it = escorts.erase(it);
  else
  {
    if(escort)
      RegisterEscort(ship);
    ++it;
  }
}
```

source/Ship.cpp
```diff
@@ -3498,20 +3508,58 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+void Ship::RegisterEscort(const Ship &ship)
+{
+	// Cache the maximum speed that the parent should stay below to keep
+	// all escorts together.
+	// We don't cache the speeds of escorts that have 0 velocity to avoid
+	// having the parent get stuck when one of the escorts doesn't have
+	// regular thrust.
+	// We also don't cache the speeds of carried ships, since they are often
+	// docked and the carrier waits for them during docking already.
+	if(!ship.CanBeCarried() && ship.MaxVelocity() > 0.)
```
tehhowch, line 3555: Sorry, I wasn't really clear with the previous suggestion -- since MaxVelocity is a computed quantity, can we only compute it once, after we know it's not a carried ship?

source/AI.cpp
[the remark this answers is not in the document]
petervdmeer: Handled in #5621

source/AI.cpp
```diff
@@ -1789,7 +1798,13 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	if(!isClose || !isFacing)
 		command.SetTurn(TurnToward(ship, dp));
 	if(isFacing)
-		command |= Command::FORWARD;
+	{
```
tehhowch, line 1801: This section was updated in master - mind pulling in latest and verifying the behavior is as expected still?

petervdmeer: I did some more testing; when a ship has a velocity perpendicular to the target planet course, then the 30 degree margin results in the ship spiralling to its target planet. It doesn't look very bad, but it looks a bit unprofessional to see a ship spiral for landing.
I think this PR needs a little bit more work to fix this.

petervdmeer: I merged master so that anybody can run the test on the latest version.

petervdmeer: Spiralling is fixed. Ships now calculate the delta-vector they need to apply to get to their cruise-vector (cruise-speed in the direction of the target).

petervdmeer:
Moved to ready for review. This PR might benefit from some additional testing with other ships than the sparrow I tested with in the above savegame.

tehhowch (commented):
Didn't playtest it yet, just code notes.

source/AI.cpp
```diff
@@ -2281,7 +2309,7 @@ void AI::DoSurveillance(Ship &ship, Command &command, shared_ptr<Ship> &target)
 	else if(ship.GetTargetStellar())
 	{
 		// Approach the planet and "land" on it (i.e. scan it).
-		MoveToPlanet(ship, command);
+		MoveToPlanet(ship, command, ship.CruiseVelocity());
```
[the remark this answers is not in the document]
petervdmeer: Agreed. Changed back to original code.

source/AI.cpp
```diff
@@ -1782,12 +1791,31 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
...
+	// Calculate target vector required to get where we want to be.
+	Point tv = dp;
+	bool hasCruiseSpeed = (cruiseSpeed > 0.);
+	if(hasCruiseSpeed){
+		// The ship prefers a velocity at cruise-speed towards the target, so we need
+		// to compare this preffered velocity to the current velocity and apply the
+		// delta to get to the preferred velocity.
+		tv = (dp.Unit() * cruiseSpeed) - velocity;
+		// If we are moving close to our preferred velocity, then face towards the target.
+		if(tv.Length() < .1)
```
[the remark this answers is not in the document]
petervdmeer: The variable `slow` seems to have a meaning to determine if a ship is considered to be "in position".
The constant here has a different meaning more in the sense of "how much are we on course towards our target". I could change this to a variable, but the meaning then would be about the amount of sloppiness in moving towards the target.

petervdmeer:
I did some more testing and found some issues; some angry Archons (that I didn't know about as player) were slowing down my flagship during landing. The speed calculation now only takes into account player-owned ships.

Zitchas:
I definitely think this makes the game look better. It helps fleets look and act more like fleets, and that's a really good thing.

This also looks like "ground work" for the formations PR, which I am eagerly awaiting. I have really enjoyed using it in Revamp, and even when not using it, I think it makes the universe feel much more coherent and interesting.

source/Ship.cpp
```diff
@@ -2608,6 +2608,10 @@ void Ship::Disable()
 void Ship::Destroy()
 {
 	hull = -1.;
+	
+	shared_ptr<Ship> oldParent = parent.lock();
+	if(oldParent)
+		oldParent->RegisterEscorts();
```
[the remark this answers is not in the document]
petervdmeer: Yes, I noticed this as well and I considered to add `if(hull > -1)` before the `hull = -1`, but the "overkill" values don't actually seem to be used, so it didn't really seem necessary.

source/Ship.cpp
```diff
@@ -3585,20 +3599,73 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+void Ship::RegisterEscort(const Ship &ship)
+{
+	// Cache the maximum speed that the parent should stay below to keep
+	// all escorts together.
+	// We don't cache the speeds of escorts that have 0 velocity to avoid
+	// having the parent get stuck when one of the escorts doesn't have
+	// regular thrust.
+	// We also don't cache the speeds of carried ships, since they are often
+	// docked and the carrier waits for them during docking already.
+	if(!ship.CanBeCarried() && !ship.IsDestroyed() && ship.IsYours())
```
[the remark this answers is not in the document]
petervdmeer: If we want to consider friendly mission ships, then we could also check the goals for the NPC (kill, save, etc), but I'm affraid that that could still be tricky.
For now I just changed it to check for the same government.

source/Ship.cpp
```diff
@@ -3669,20 +3683,73 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+	// docked and the carrier waits for them during docking already.
+	if(!ship.CanBeCarried() && !ship.IsDestroyed() && government == ship.GetGovernment())
+	{
+		double eV = ship.MaxVelocity() * 0.9;
+		if(eV > 0.)
+		{
+			if(escortsVelocity <= 0.)
+				escortsVelocity = eV;
+			else
+				escortsVelocity = min(escortsVelocity, eV);
```
quyykk, line 3750: Quick observation: Can't you just store the slowest escort? That way, you only need to loop through the whole escort list when that slowest escort dies/gets removed.

source/Ship.cpp
```diff
@@ -2605,6 +2605,10 @@ void Ship::Disable()
 void Ship::Destroy()
 {
 	hull = -1.;
+	
+	shared_ptr<Ship> oldParent = parent.lock();
+	if(oldParent)
+		oldParent->TuneForEscorts();
```
quyykk, line 2611: If you want: In `Ship::Move` the ship checks whether it's dead and then does various death related stuff. If you put there `SetParent(nullptr);`, you'll get this behavior for free, since `SetParent` removes escorts from the old parent.

source/AI.cpp
```diff
@@ -1792,12 +1801,32 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
...
-
-	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .95);
+	
+	// Calculate target vector required to get where we want to be.
+	Point tv = dp;
+	bool hasCruiseSpeed = (cruiseSpeed > 0.);
+	if(hasCruiseSpeed)
+	{
+		// The ship prefers a velocity at cruise-speed towards the target, so we need
+		// to compare this preffered velocity to the current velocity and apply the
```
petervdmeer, line 1811:
```suggestion
		// to compare this preferred velocity to the current velocity and apply the
```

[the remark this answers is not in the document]
petervdmeer: Fixed

source/Ship.cpp
petervdmeer: Good idea! Thanks for the suggestion. I'll add that.

source/Ship.cpp
petervdmeer: Ships can also be destroyed due to a Mission-Action. So moving it there might cause that case to be missed. (But I guess I could replace this code here by `SetParent(nullptr)`.

petervdmeer: Person ships are recylced (using the Ship->Restore() function), so it looks like we are not allowed to break the parent/child relations on destroyed ships (at least not on person ships).

quyykk: Ah didn't know that, thanks!

source/Ship.cpp
petervdmeer: I added this to reduce some looping.

Zitchas:
Question: is this a strict implementation of speed matching? Was adaptive matching considered?

Just wondering, as this sounds like a contrast to the EV series, where (if I recall correctly) the game actually upped the speed of escorts while flying in formation by <10% as a means of enabling good formation flying without having to limit the player.

petervdmeer:
> Question: is this a strict implementation of speed matching?

It is not strict, players can still choose to fly at 100% of their speed, even if they are flying in formation. But the auto-pilot code for landing will use 90% of the speed of the slowest escort when landing (if there is no hostile overwhelming force in the system). Players can also overrule the landing speed by increasing speed manually before starting the auto-pilot landing.

> Was adaptive matching considered?

I did consider the alternative of giving escorts a speed advantage when following a fast lead but decided against it:
- It doesn't feel fair to make ships faster than they are.
- It seems fair that fast lead-ships should slow-down a bit to allow their escorts to catch up.

> Just wondering, as this sounds like a contrast to the EV series, where (if I recall correctly) the game actually upped the speed of escorts while flying in formation by <10% as a means of enabling good formation flying without having to limit the player.

I never played EV, so I didn't know about this. Something that I did consider for formation flying is giving escort ships additional means of slowing down (automatically applying some slowing down if they "overshoot" their formation-position). The slowing-down option doesn't feel as unfair as making ships faster than they are, and it will make the formation flying behavior less restless.
But this is all out of scope for this PR; this PR is just about having fast lead-ships slow-down to allow their escorts to follow their lead (regardless of if those escorts are flying in formation or are just station-keeping to try to stay with the lead-ship).

source/Ship.h
```diff
@@ -425,6 +426,9 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	// Add or remove a ship from this ship's list of escorts.
 	void AddEscort(Ship &ship);
 	void RemoveEscort(const Ship &ship);
+	// Cache relevant data for all escorts or the given escort.
+	void TuneForEscorts();
+	void TuneForEscort(std::shared_ptr<Ship> &ship);
```
tehhowch, line 431:
```suggestion
	void TuneForEscort(const std::shared_ptr<Ship> &ship);
```
The given ship isn't modified

source/AI.cpp
```diff
@@ -1810,11 +1819,31 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
...
+	Point tv = dp;
+	bool hasCruiseSpeed = (cruiseSpeed > 0.);
+	if(hasCruiseSpeed)
+	{
+		// The ship prefers a velocity at cruise-speed towards the target, so we need
+		// to compare this preferred velocity to the current velocity and apply the
+		// delta to get to the preferred velocity.
+		tv = (dp.Unit() * cruiseSpeed) - velocity;
+		// If we are moving close to our preferred velocity, then face towards the target.
+		if(tv.Length() < .1)
```
tehhowch, line 1832: We can use the squared version just as easily:
```suggestion
		if(tv.LengthSquared() < .01)
```

source/AI.cpp
```diff
@@ -1810,11 +1819,31 @@ bool AI::MoveTo(Ship &ship, Command &command, const Point &targetPosition, const
 	bool shouldReverse = false;
 	dp = targetPosition - StoppingPoint(ship, targetVelocity, shouldReverse);
 
-	bool isFacing = (dp.Unit().Dot(angle.Unit()) > .95);
+	// Calculate target vector required to get where we want to be.
+	Point tv = dp;
```
tehhowch, line 1823: Should we be adjusting the "stopping point" instead of calculating where we want to be "twice"?

tehhowch (commented):
I believe we also need to handle cruise velocity updates when a fleet flagship is captured, since RemoveEscort is not called for performance reasons

source/Ship.cpp
```diff
@@ -3788,20 +3805,91 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+		auto escort = it.lock();
+		if(escort)
+			TuneForEscort(escort);
+	}
+}
+
+
+
+// Store relevant cached data for the given escort.
+void Ship::TuneForEscort(std::shared_ptr<Ship> &ship)
```
petervdmeer, line 3875:
```suggestion
void Ship::TuneForEscort(const std::shared_ptr<Ship> &ship)
```

source/Ship.cpp
```diff
@@ -3194,6 +3198,16 @@ double Ship::MaxReverseVelocity() const
 
 
 
+// The optimal speed for this ship when not moving in a hurry.
+// If the ship has escorts, then the speed returned will allow for all
+// non-carried escorts to catch up with this ship.
+double Ship::CruiseVelocity() const
+{
+	return escortsVelocity >= 0. ? min(MaxVelocity(), escortsVelocity) : MaxVelocity();
```
petervdmeer, line 3206:
```suggestion
	return escortsVelocity > 0. ? min(MaxVelocity(), escortsVelocity) : MaxVelocity();
```

source/Ship.cpp
```diff
@@ -3788,20 +3805,91 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+void Ship::TuneForEscort(std::shared_ptr<Ship> &ship)
+{
+	// Cache the maximum speed that the parent should stay below to keep
+	// all escorts together.
+	// We don't cache the speeds of escorts that have 0 velocity to avoid
+	// having the parent get stuck when one of the escorts doesn't have
+	// regular thrust.
+	// We also don't cache the speeds of carried ships, since they are often
+	// docked and the carrier waits for them during docking already.
+	if(ship && !ship->CanBeCarried() && !ship->IsDestroyed() && government == ship->GetGovernment())
```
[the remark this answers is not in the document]
petervdmeer: Agreed, that check is not needed.

source/Ship.cpp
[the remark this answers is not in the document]
petervdmeer: No, that wasn't intended.

source/Ship.cpp
```diff
@@ -3788,20 +3805,91 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+	// Check if we have a valid slowest escort. If we have one, then we don't
+	// need to rescan for the next slowest escort.
+	shared_ptr<Ship> slowest = slowestEscort.lock();
+	if(slowest && !slowest->IsDestroyed() && government == slowest->GetGovernment())
+		return;
+
+	// Reset this value, we will re-scan the list of escorts to set the
+	// escorts velocity based on remaining active escorts.
+	escortsVelocity = -1.;
+	slowestEscort.reset();
```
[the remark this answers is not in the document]
petervdmeer: A struct is more readable, and if we do this, then I feel it is nicer to also add the vector with the escorts themselves to the struct (since modifications to that vector should also always trigger a check for the speeds).

source/Ship.cpp
```diff
@@ -3250,7 +3264,10 @@ int Ship::TakeDamage(vector<Visual> &visuals, const DamageDealt &damage, const G
 		hullDelay = max(hullDelay, static_cast<int>(attributes.Get("disabled repair delay")));
 	}
 	if(!wasDestroyed && IsDestroyed())
+	{
 		type |= ShipEvent::DESTROY;
+		Destroy();
```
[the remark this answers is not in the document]
petervdmeer: Yes, that is indeed nicer.

source/AI.cpp
petervdmeer: The "stopping point" is still the same before and after this PR; the only difference is that we now also check at what speed we want to get there (instead of always going full-speed).
In the situation where a ship is moving towards its target, but at a higher speed than the cruiseSpeed, then we would want the ship to actively slow down; so the calculation of the desired speed can result in a vector opposite of the vector directly towards the target to ensure the ship slows down.
Maybe I'm missing something, but I don't see which part of the calculation can be optimized to avoid a double calculation.

source/Ship.cpp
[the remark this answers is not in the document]
tehhowch: Yes, that would be in somewhat in line with implementing a "ship < squadron < fleet < government" AI hierarchy I mentioned in a previous PR / discord.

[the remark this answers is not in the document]
petervdmeer: I added the struct (and I also tried a different approach by an extended vector class that tracks the cruise-speed, but such extended vector would result in some code-duplication for the government code, so I decided to go for the struct).

source/Ship.cpp
```diff
@@ -3788,20 +3804,101 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+void Ship::TuneForEscorts()
+{
+	// Check if we have a valid slowest escort. If we have one, then we don't
+	// need to rescan for the next slowest escort.
+	shared_ptr<Ship> slowest = escorts.slowest.lock();
+	if(slowest && !slowest->IsDestroyed() && government == slowest->GetGovernment())
+		return;
+
+	// Reset this value, we will re-scan the list of escorts to set the
+	// escorts velocity based on remaining active escorts.
```
tehhowch, line 3859: This is a "what" comment rather than a "why," and doesn't offer anything the preceding comment or the invoked method comment(s) dont already offer.
```suggestion
```

source/Ship.cpp
```diff
@@ -2851,14 +2852,16 @@ void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
 		if(bay.ship)
 			bay.ship->WasCaptured(capturer);
 	// If a flagship is captured, its escorts become independent.
-	for(const auto &it : escorts)
+	for(const auto &it : escorts.list)
 	{
 		shared_ptr<Ship> escort = it.lock();
 		if(escort)
 			escort->parent.reset();
 	}
```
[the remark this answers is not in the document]
petervdmeer: If `Escorts` becomes a resource-managing class, then I might as well move the TuneForEscort(s) functions into that class (and just make the MinimumSpeed() available as method on the class). But doing this properly would imply that the push_back and erase functions are also to be encapsulated.

Do you think it's valuable to also to do this in this PR?

source/Ship.cpp
```diff
@@ -3907,20 +3915,97 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
 // cues and try to stay with it when it lands or goes into hyperspace.
 void Ship::AddEscort(Ship &ship)
 {
-	escorts.push_back(ship.shared_from_this());
+	auto sp = ship.shared_from_this();
+	escorts.list.push_back(std::move(sp));
+	TuneForEscort(ship);
```
tehhowch, line 3920:
```suggestion
	escorts.list.push_back(ship.shared_from_this());
	TuneForEscort(ship);
```

source/Ship.cpp
```diff
@@ -3907,20 +3915,97 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
...
+	{
+		double eV = ship.MaxVelocity() * 0.9;
+		// We don't cache the speeds of escorts that have 0 velocity to avoid
+		// having the parent get stuck when one of the escorts doesn't have
+		// regular thrust.
+		if(eV > 0. && (escorts.cruiseVelocity <= 0. || (eV < escorts.cruiseVelocity)))
+		{
+			escorts.cruiseVelocity = eV;
+			auto sp = make_shared<Ship>(ship);
+			escorts.slowest = sp;
```
tehhowch, line 3996: Invoking the Ship copy constructor is _definitely not_ what you want.
That will lead to a totally new shared_ptr, rather than incrementing the refcount:

```suggestion
			escorts.slowest = ship.shared_from_this();
```

petervdmeer:
The latest updates caused a conflict between the code, LTO (Link-Time-Optimization) and some tests.

So we either need to either:
- Fix the tests to run without LTO.
- Merge this PR with less optimal code (remove the inner-class)
- Or disable the affected test for now.

I prefer to fix the tests to run without LTO, but that will likely take a long time...

quyykk:
>I prefer to fix the tests to run without LTO, but that will likely take a long time...

I mean, it's nice that the tests don't have a dependency on SDL and co. because of LTO, but if the choice is between linking the tests with SDL and co. anyways, and not having tests, I'd rather we do the first.

Sure it's not great, because we don't want tests to link with SDL and co., but it's the pragmatic solution and we can always work towards eliminating this dependency in the future.

warp-core (commented):
Only running the flagship at 90% the speed of the slowest escort means that the escorts can end up a decent way ahead of me when traversing a system. So, instead of the flagship rushing off ahead of the escorts, the escorts are rushing away from me.

Also, I've opened a PR to your branch to bring this up to date and resolve conflicts, @petervdmeer.

Amazinite:
This is currently our oldest open PR now! Please try to finalize and merge it.

petervdmeer:
> Only running the flagship at 90% the speed of the slowest escort means that the escorts can end up a decent way ahead of me when traversing a system. So, instead of the flagship rushing off ahead of the escorts, the escorts are rushing away from me.

The escorts behavior should have been addressed in (related PR) https://github.com/endless-sky/endless-sky/pull/5621, so either you tested with a branch where 5621 wasn't merged yet, or you found some situation in which 5621 doesn't work as intended.

> Also, I've opened a PR to your branch to bring this up to date and resolve conflicts, @petervdmeer.

Thanks! I merged it.

source/AI.cpp
```diff
@@ -1546,7 +1546,16 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
...
 	else if(ship.GetTargetStellar())
 	{
-		MoveToPlanet(ship, command);
+		// If our allies are clearly stronger than our enemies in this system,
+		// then move at more civilized speeds when going to a planet.
+		// A negative cruising speed indicates the ship should move as quickly as possible.
+		double cruiseSpeed = -1.;
+		auto ait = allyStrength.find(ship.GetGovernment());
+		auto eit = enemyStrength.find(ship.GetGovernment());
+		if(ait != allyStrength.end() && eit != enemyStrength.end() && ait->second > eit->second * 2)
```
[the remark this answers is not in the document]
petervdmeer: Yes, I'll update the code.

petervdmeer:
> This is currently our oldest open PR now! Please try to finalize and merge it.

This PR still needs a solution for dealing with the LTO failure (as discussed around 12 Nov 2022).

source/Ship.cpp
```diff
@@ -4357,3 +4440,15 @@ double Ship::CalculateDeterrence() const
...
+
+
+Ship::Escorts::~Escorts()
+{
+	for(const auto &it : list)
+	{
+		shared_ptr<Ship> escort = it.lock();
+		if(escort)
+			escort->parent.reset();
+	}
```
quyykk, line 4453: Two questions:

1. This would seem more appropriate to do in the Ship destructor no?
2. This seems unnecessary at first glance, because the AI resets the parent pointer if the parent is destroyed and the parent pointer is a weak ptr, so it will be null automatically when the ship is destroyed.

warp-core:
@petervdmeer regarding escorts rushing off ahead of the flagship, the following was recorded with the current continuous (https://github.com/endless-sky/endless-sky/commit/26de3d8dbd156d4f6fade3da4bf0ac22b423d197):

https://www.youtube.com/watch?v=C1KsBHkmbXI

And, this is with this PR:

https://www.youtube.com/watch?v=UTuFirYPslk

In both cases, the faster ships rush off ahead of me, so that part isn't a result of this PR, but perhaps made more obvious by it.

warp-core:
I'm closing this because it doesn't appear to be going anyway.
@petervdmeer @quyykk feel free to reopen if you agree on and implement a fix.

tibetiroka (dismissed):
This change only works if your ship is faster than the escorts; the escorts will not limit their speed to match yours. This is an issue with regular ships as well as deployed fighters, which will always be faster than their carrier.

In my (limited) testing, it just doesn't handle right. For instance, if I have a mixed fleet with a Carrier, some fighters and Berserkers (with a Berserker flagship), the fleet ends up looking like this:

image

It might make more sense to have a fleet-wide cruise speed calculated and applied to every ship.

petervdmeer:
> This change only works if your ship is faster than the escorts; the escorts will not limit their speed to match yours. This is an issue with regular ships as well as deployed fighters, which will always be faster than their carrier.

Yes, that is correct; this was part of a set of 2 PRs, where the other PR (https://github.com/endless-sky/endless-sky/pull/5621) would ensure that escorts stay close to the parent during landing. That other PR was merged on 9th January 2021, but that functionality was apparently lost by an update in some later PR.

You don't want to limit the speed of all ships, you just want the parent to slow down to the slowest escort, and then have the escorts stay close to the parent.

Zitchas:
... Just to be clear, you are waiting on yourself?

petervdmeer:
> ... Just to be clear, you are waiting on yourself?

Yes this PR is waiting for me. Unless someone beats me to restoring the functionality of #5621.

petervdmeer:
> This change only works if your ship is faster than the escorts; the escorts will not limit their speed to match yours. This is an issue with regular ships as well as deployed fighters, which will always be faster than their carrier.

This should be fixed by now; escorts should follow their flagship until the flagship is actually landing.

source/Ship.cpp
petervdmeer: 1 This function is also executed when a flagship gets captured. (And makes the escorts independent).
2 Sure, but it seems relevant when the flagship gets captured.

tibetiroka (changes requested):
Actually, back to the change request.

Currently, the cruise speed is calculated for a parent and its escorts only when escorts are added/removed. This means that the cruise speed is essentially a constant, and fails to take into account the current fleet status. The cruise speed calculation will include ships that are:
 - not in the system
   - even if they can't get to the system
 - are ordered to hold position, instead of landing with the flagship
 
and probably any other condition that I haven't tested, such as being disabled.

Looking at the PR, it's approach seems to be centered around caching the cruise speed, which might make it difficult to fix these issues.

source/AI.cpp
```diff
@@ -2035,7 +2035,14 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
 	}
 	else if(ship.GetTargetStellar())
 	{
-		MoveToPlanet(ship, command);
+		// If our allies are clearly stronger than our enemies in this system,
+		// then move at more civilized speeds when going to a planet.
+		// A negative cruising speed indicates the ship should move as quickly as possible.
+		double cruiseSpeed = -1.;
```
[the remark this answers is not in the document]
petervdmeer: You're right; zero should also work and is probably more clear. Changed it to zero.

petervdmeer:
As discussed with tibetiroka on discord; This PR now only contains the fix to let fast escorts stay close to the flagship during landing. This is probably the most common situation, and flagships just don't slow down at this moment..

Zitchas:
That seems good enough for the player, but for NPCs I think having the flagship actively take escorts into account and not out pace them is important to have.

tibetiroka (approved):
The current feature works. Thanks!

PR #5584, opened 2020-12-29
Human outfits graphical upgrade

beccabunny:

## Summary
This PR, which is quite possibly the biggest PR I've ever worked on (more than the Ka'het in term of work it needed) features a complete graphical upgrade to almost every human outfit plus some korath ones that need it, totally 123. I don't expect all of them to get merged, and I will remove the ones deemed not worthy to change; but I believe most of these could get merged. Stuff like projectiles and hardpoints have not been added, and will be made once it is decided what should get removed and what not. Here's screenshots at max resolution to see what I'm talking about.
<details>
<summary>Screenshots</summary>
<br>

Screenshot (1528)

Screenshot (1529)

Screenshot (1530)

Screenshot (1531)

Screenshot (1532)

Screenshot (1533)

Screenshot (1534)

Screenshot (1535)

Screenshot (1536)

Screenshot (1537)

Screenshot (1538)

Screenshot (1539)

Screenshot (1540)

Screenshot (1541)

Screenshot (1542)

Screenshot (1543)

Screenshot (1544)

Screenshot (1545)

Screenshot (1546)

Screenshot (1547)

Screenshot (1548)

Screenshot (1549)

Screenshot (1550)
</details>

## PR Checklist
 - [x] I uploaded the necessary image, blend, and texture assets here: https://drive.google.com/file/d/1qKP6allR1GL4RkKBQMa2bjXCAmq-6Ueg/view?usp=sharing
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/158


Anarchist2:
~the meteor storage isn't even a box anymore smh~

beccabunny:
Anarchist makes a good point here, it is something I was asking myself as well. What would be the best course of action? Personally, I think it wouldn't be bad to take the outfits with names that don't fit much anymore to depreciated outfits.txt and give them new names, as well as changing every occurrence in the game to the new one

Terin:
Comments on a few of the items:

Bunks/Expansions: Looking great. Despite me liking the EV screwdrivers, they were immediately readable to me with the colors to match. The bunk rooms look great as well, helping explain why these would need 20 tons of outfit space. (although, in the case of the small bunk room, those could use a description change from "hanging a hammock next to the engines" to "finding the floor space to actually tuck this room" or something)

Shield Regenerators: The Red/Green emitters got a lot darker compared to the originals, they could probably glow a bit more to match.

Hyperdrive/Scram Drive: The ducts on top of the drives are a bit distractingly low-tech compared to the brass tubes of before. While I'm not a big fan of steampunk aesthetics, it communicated that these drives were weird tech that nothing else in human space was like.

AA/BB Fuel Cells: The central heat sink part of these fuel cells are much closer in size than the originals, which has a weird effect of making the AA cell look more advanced than the BB one. The AA heat sink should be shrunk a good amount, and maybe the cylinders to the right made thinner/smaller in general.

beccabunny:
Adopted the changes Terin suggested both to the tiny fuel cell and to the regenerators

Zitchas:
For ease of comparison, I have made a plug-in with the vanilla stuff.

https://drive.google.com/file/d/1v5C4O-terqCPGN_uKonmmVcuBuioJTdP/view?usp=sharing

The plug-in creates a duplicate of every human outfit, named "originalnamehere (vanilla)" with a corresponding image name "originalimagename (vanilla)". It then adds every outfit that is found in an outfitter to the "basic" and "common" outfitter lists. 

Two caveats:
1) This is *not* an all-access plug-in. It doesn't grant licenses. If something needs an event that isn't simply adding a new outfitter to a place to access, then you still need that event. All this plug-in does is put every human outfitter outfit list into the basic and common ones. There might be exceptions. If people point them out, I'll try to include them. I'm not going hunting for them, though.
2) This is *not* save friendly. Installing any outfit marked ` (vanilla)` on your ship *will* vanish when you uninstall the plug-in again. Likewise, all such vanilla outfits will not be recognized by missions that refer to specific outfits (like the Remnant tech retrieval missions, or the JD mission with the Pug events, luxury or brig jobs, etc.)
3) No, I will *not* be adding mission compatibility to this plug-in. It is a "quick and dirty comparison tool" for the sake of helping review this PR. It will not be maintained after this PR is done.

**To Use**
1. Install this plug-in in your plug-in folder as per normal.
2. Use continuous, Github Desktop, or whatever your preferred method of testing PRs is to switch to this PR.
3. Launch the game and look in a human outfitter.

Amazinite (changes requested):
Posted a review on the high DPI PR (because I forgot I was looking at it for the higher res images).

beccabunny:
I tried to address most of Derpy's comments in the review,  leaving aside only the ones I thought weren't worth it. These include: 

- The Gatling now has the targeting detector
- The Brig image is now a prison cell
- Spines of the HD/SD enlarged
- Some changes to the Stack Core, including more tubes
- Changes to the Fission that include the control rods being more prominent and cross shaped
- More stuff around the main body of the Fusion
- Lots of changes to the Armageddon core making it more like the original
- Renamed cargo and outfit expansion

Regarding the atomic problem, given that the Hai and Deep engines are mostly identical (or at least very, very similar) and that in general I really prefer human stuff to look the most realistic, what I'd like to do is keeping the original design for the Hai engines (even when/if I start the alien stuff), as it has a quite alien look and simply use the ones added by the PR as the Deep ones, which was the original plan. If that isn't fine let me know

Last but not least, I added the rest of the Korath outfits that needed something to the PR, being the Grab-strike,  Banisher, Fire-lance, Command Center, Control Transceiver and all the Piercer related stuff

Red-57:
Is it possible that you could post the assets in such a way that people don't have to download the entire thing to get at the one blend file that they want, maybe have separate downloads for each category.

Zitchas:
I would like to raise the idea of swapping the colors of the Deep atomics and Hai atomics. The rest of the Hai outfitter uses a green palette, so I always assume that the green atomics are the Hai ones. Since we're revamping all the human graphics and changing their shape too, this seems like a good opportunity to change the color to blue; then use the green alien-looking atomics as the Hai ones. 

I created an issue #5598 for discussion about this.

Amazinite (changes requested):
Maybe move the command center image up a few pixels, as it is currently ever so slightly cut off at the bottom corner now.

And the Fire-Lance could maybe use with a bit more yellow near the muzzle, as right now you've removed a fair amount of it. What form that takes? Making another set of fins behind the current ones so that there's four fins in the front like there was before might help. Could also add fins on the top and bottom that are more like the ones on the middle (i.e. less spiky).
Just a shitty sketch-up of what I'm thinking, with either the top fin or the side fins.
image
This one is entirely up to you though. I won't fuss about it.

Otherwise, I think everything looks good. I've ultimate decided to not mess with changing the colors on the atomic engines.

Amazinite (approved):

Zitchas (changes requested):
This restock of outfit images is a clear upgrade in the quality and detail of all the images provided, and a very nice improvement. 

My only concern is that most images seem to be both muted and darker. By this I mean that the colors are dimmer and less intense, and there is frequently more shadows. I don't know if the shadows are actually darker, but they certainly *feel* darker to me. It is as if the entirety of human outfitters all suddenly switched from ambient lighting to spotlights. Things definitely feel like everything is closer to being drab roughed-up-industrial sort of aesthetic than it used to be. The muting of colors isn't necessarily bad - it does look good, and it helps make alien stuff stand out a bit more. The shadows, on the other hand, I feel are a negative. I don't think we gain anything by having so much darkness.

Take the brig, for example. The main point of the outfit is the reinforced security doors, and yet in the image that whole wall and the doors are in shadow and difficult to make out. 

The gattling gun is another example of an image that has become much darker than it used to be; and this is intensified by having the body of the gattling gun switched to being a very dark grey material.

In terms of image design, however, I do have a concern:

The Large Rader Jammer: This outfit used to point outwards, which makes sense for an omni-directional effect. The new image has all three dishes pointing inwards, as if focussed on a single point located within the outfit. This makes me think of a microwave oven or some kind of scientific scanner for doing detailed scans of some small object located at the point of conjunction. I'd like to see them angled outwards again, or at least shifted so that their focus point is somewhere far outside the outfit instead of right in it. When you get right down to it, an omni-directional setup like the old radar jammer or either of the outfit scanners. The new radar jammer's focused direction would be a better look for the outfit scanner, which is a targeted effect.

My suggestion would be to swap bases: Have the Large Radar Jammer on the central pillar base of the outfit scanner, so all the dishes are pointed outwards; while the outfit scanner is put on the large radar jammer's base, and thus the dishes are focused on a single point. This would make their images convey the behaviour of the outfit.

In terms of reviews, this is basically a "98% approved, 2% request changes" sort of thing. If this all got merged as-is, I wouldn't be upset. But I do think fixing the Large Radar Jammer would be a definite good thing.  As for the shadows... Not sure what (if anything) can be done about them. Those aren't a deal breaker, just something I wanted to put out there for everyone to think about.

Zitchas:
And yes, definitely like having that yellow glow on the Ramscoop. It doesn't really look like an internal glow, but at least there is *a* glow.

copyright
```diff
@@ -372,6 +372,69 @@ Files:
...
+ images/outfit/small?radiothermal*
+ images/outfit/small?thermionic*
+ images/outfit/stack?core*
+ images/outfit/surveillance?pod*
+ images/outfit/banisher*
+ images/outfit/command?center*
+ images/outfit/fire-lance*
+ images/outfit/piercer*
+ images/outfit/piercer?launcher*
+ images/outfit/jorath?piercer?storage*
```
Amazinite, line 437:
```suggestion
 images/outfit/korath?piercer?storage*
```

Amazinite, line 437:
```suggestion
 images/outfit/korath?piercer?storage*
```

Zitchas (approved):
While my concerns about the darkness are still there, I don't think it is a blocking issue. (While a few others have commented about it, I don't think anyone feels strongly about it, myself included. ) And my concerns about the radar jammer have been resolved. 

Thus, approving.

Amazinite (commented):
On Zitchas' comment about images being a bit too dark:
* The Brig could use some extra lighting, as much of the image is cast in its own shadow. The command center for example has a glow along the back wall so that the stuff against the wall isn't cast in shadow. Maybe do something similar here.
* The body of the Gatling Gun is itself a darker color, and much of the detail gets lost in that darkness. Maybe bring the Gatling Gun back to a more grey color.
* The ion engines use a darker blue for the ends, and the steering is cast in more shadow now that the steering engines are bent instead of angled. Maybe use a lighter/brighter blue for the engines, and add some extra light for the shadowed parts of the steering.
image
* Maybe add a warm red glow in the nozzles of the plasma engines to contrast with the blue glow of the ion engines. The old plasma engines had the nozzles in shadow, but given that these new ones have the nozzles packed more closely together, especially so on the huge engines, the shadow becomes one big mass. Could maybe do the same with the afterburner.
* Aside from the nozzles, maybe recolor the plasma engines from the brownish color you have them to something closer to the yellowish-orange of the old plasma engines to make them a bit more colorful.
* These parts on the water cooling.
image
* The old cargo scanner was angled down more. Maybe do the same on the new cargo scanner, which should hopefully shine some more light on this area that's currently in shadow.
image
 
Other comments:
* The Flamethrower tip could use something a bit more. Perhaps return the fins that the old one had, or make the tip look more like the nozzle of an IRL flamethrower where the barrel flares out then comes back in.
image
* The liquid nitrogen and helium outfits had these cooling fins on them previously, but on the new images these have been made smaller to the point where they're hardly noticeable. Perhaps make these fins more prominent by reducing the number and making the ones that remain stick out more.
image

Terin:
Throwing a couple more in at the last second, as well:

image

**Hyperdrive/Scram Drive:** Like I mentioned earlier, the large, industrial ducts compared to the old tubing:
- Draw attention away from the more unique/anomalous elements of the drives by being continuous regions of light/color, especially since the previously-exposed tops of the cylinders/boxes are covered with the duct connections
- Are themselves a bit too mundane compared to the strange tubing it had before.

I'd like to see them replaced with a couple of those smaller ducts you used to replace the tubing in other places, at the very least. On the scram drive in particular, the big ducts block a good chunk of the back cylinder, so smaller ones would free up some visual space and help draw attention to the main parts of the drives.

It may make less mechanical sense to have such small connections/vents, but the Hyperdrive is meant to be a very out-of-place piece of human tech, which was communicated with the strange tubes before (on top of everything else)

image

**Shields 2/3:** While comparing the two side-by-side, you can see that the tier 3 shields have more emitters/is larger. However, that pylon on top of the tier 2 shields adds a sort of complexity that makes it feel like it should be more powerful than the tier 3 shields, at first glance. Perhaps that top pylon could be moved to the tier 3 ones, which would put it in continuity with the tier 4 shields (before the tier 5 ones go "Fuck it, pile on as many chunky emitters you can fit")

This one is definitely tricky since the original shields also could be tough to tell between, and shields are usually sequential in shops anyways. But if moving that pylon over to the tier 3 shields is doable, that'd go a long way.

WarlordMike:
And as long as the train's runnin' here...

The particle cannon comes out looking a bit blurry in the outfitter view. Maybe making the yellow squares on the sides just a bit thinner (so the gaps are wider) and have them not stick out quite as far would help with this. In addition, the contrast between the colored bits and the rest of the cannon is a bit muted. If you up the saturation on the colored bits, it might help draw the contrast out a bit better.

As an aside, the fin on the grab strike that Derpy mentioned in his commentary from the High DPI PR (https://github.com/endless-sky/endless-sky-high-dpi/pull/158) wasn't addressed. 
image

Other than that, I'd say just tackle the few things Derpy and Terin pointed out, and all should be good!

beccabunny:
Since I promised it, here's the reasoning behind what I have not done in the last commit between the suggestions:
- The reddish glow suggestion made the plasmas too korath, even more than now, and I did not particularly like the outcome; I went with white enviromental lighting like everywhere else
- the flamethrower already had two of those things derpy mentioned, so it didn't really feel necessary
- the hyperdrive: I quite disagree with it having to look out of place, after all it still is a 9 centuries old technology at the time of the game's start. It is advanced, sure, but still something that every spaceship has. I think the current sprite is fine for that.
- I kinda agree with the shields one, and I did try it along with the rest of the suggestions, but it was just not possible unfortunately. MZ did the even numbered shields as uv spheres and the uneven numbered ones as icospheres, which makes it hard to take specific details from one to the other while still making sense. In this case, it just didn't look good on the icospheres while it was necessary on the uv ones since the further you go from the middle of the sprite, the smaller and indistinguishable the triangles become
- on the regenerators, my view is quite the opposite: the original was incredibly and completely randomly messy, making some even suggest going with a complete redesign instead because of how the old one just didn't work. In the new ones, I decided to keep most of the old ones' qualities while still making it at least radially symmetrical, and I definetly want that to be visible
- Grab strike: derpy's comment actually wasn't ignored, I rejected it since the fin had literally no purpose, was not particularly aestechically pleasing and did not fit with the rest of the korath. The only reason to put it there would have been because the old one had it, and the previous points greatly outweight this one

PR #5586, opened 2020-12-30
Fix: Add visit messages to Emerald Sword missions

Terin:
## Issue
Currently, landing on the Valley of the Damned without boarding the Emerald Sword, and/or having the required crew/cargo on an escort without a jump drive will silently prevent you from completing the mission, with the former being much more likely, as a post-Free World pilot could be traveling around with a large fleet.
## Fix Details
This adds an On Visit dialog to The Emerald Sword 2, which refers to the crew, cargo, and needing to board the Emerald Sword. It also adds a more generic On Visit dialog for the trip back.

## Testing Done
Created a save file with a Shuttle flagship with two jump drives, and a Bactrian escort with only a hyperdrive, which will make you eligible to receive the mission. Jumping to the target system and landing will silently softlock you without the patch, and will trigger the On Visit message with the patch.
image

## Save File
This save file can be used to verify the bugfix. The bug will occur when using dc253c5, and will not occur when using this branch's build.
[Emerald Sword.txt](https://github.com/endless-sky/endless-sky/files/5752704/Emerald.Sword.txt)


data/sheragi/archaeology missions.txt
```diff
@@ -1305,6 +1305,9 @@ mission "Sheragi Archaeology: The Emerald Sword 3"
...
 			`	After many long hours walking around, the lights of the ship finally flicker to life, along with sirens bringing attention to every other damaged subsystem, startling everyone. One of the specialized crew members comes to you, explaining that they finally managed to turn on the power plant and fix enough of the ship to move it around, but that they need to land it in order to bring in the heavier equipment and fix the rest of it.`
 
+	on visit
+		dialog `Upon landing on this strange world, you realize not everything is accounted for! Better make sure the Emerald Sword has been boarded, and no crew or cargo is on escorts out of jump drive range.`
```
Amazinite, line 1309:
```suggestion
		dialog `Upon landing on this strange world, you realize not everything is accounted for! Better make sure the Emerald Sword has been boarded, and no crew or cargo is on escorts out of the system.`
```

PR #5585, opened 2020-12-30
New indigenous lifeform in the Ember Wastes: the Embershade

Galaucus:
https://user-images.githubusercontent.com/68889791/103321740-e6048780-49ef-11eb-9d3a-b51fd8b1ae4f.mp4

## Summary
Void-dwelling creatures native to the Ember Wastes, the Embershade are ray-like entities which generate a natural cloaking field. Generally shy, they prefer to give intruding ships a wide berth and remain under cloak whenever possible. During the occasional ion storm that washes through the Wastes they become visible.

My intention with the Embershades is to add a ghostly presence to the otherwise-empty wastes; I want the player to know in the back of their mind that there's something there in the system with them, but they can't quite see it. The overall experience should be mildly disconcerting: a dark silhouette and scant luminescent patches should evoke the feeling of seeing a shadow moving in the sea, the creature's true form indeterminate. 

## Save File
[This plugin can be used to preview the new critters. For testing purposes they spawn in Segesta, Stercutus, Caeculus, and Peragenor.](https://cdn.discordapp.com/attachments/266345072554016768/793624413206609950/Embershades.7z)

## PR Checklist
 - [x] Initial mockup and testing complete
 - [x] Assets recreated in blender (blender work by @Ferociousfeind) , textured, and animated
 - [ ] New personalities implemented by @InfiniteDonuts 's PR #5698  
 - [x] Fine-tuning of stats, spawn rate, and systems
 - [x] I updated the copyright attributions, or decline to claim copyright of any assets produced or modified
 - [x] I uploaded the necessary image, blend, and texture assets here: https://drive.google.com/drive/folders/1UvahLxZrVPUZ5OXumWGBWuN-NvWx-qE5?usp=sharing
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/173

PR #5587, opened 2020-12-30
Creating three new personalities [WORK IN PROGRESS]

InfiniteDonuts:
@Galaucus (aka Lake from the Discord server) is creating a new Embershade life-form (PR can be viewed [here](https://github.com/endless-sky/endless-sky/pull/5585)), and asked me to create three new personalities: `shy`, `elusive`, and `roving`.

1. Shy
The ship attempts to avoid non-friendly vessels, retreating to a safe minimum distance if any come too close.

2. Elusive
Ship prioritizes maintaining cloak regardless of fuel cost. When attacked it will prioritize disengaging from combat and raising cloak.

3. Roving
A general purpose "mill about aimlessly" behavior to put a bit of life into fleets inhabiting empty systems. 

Shy and elusive, taken together, is intended ensure that when ion storms force ships to decloak they don't do so right under the player. 

Roving is meant to add movement to indigenous creatures who otherwise have no goal to pursue in the system where they're found.

## PR Checklist

- [x] Create `roving` personality
- [ ] Create `shy` personality
- [x] Create `elusive` personality

## Updates

- Update 12/30/2020: I am done with the Roving personality. Here is a link to a video demonstrating it: https://www.youtube.com/watch?v=KA4oTleb2n4&feature=youtu.be
- Update 1/01/2020: Finished implementing Amazinite's and tehhowch's suggested changes. Roving area is now a circle.

## Acknowledgements

- tehhowch, Terin, and Vitalchip for helping me with code.
- Lake for getting me going and applying the needed pressure to motivate me.
- Beccabunny, Amazinite (aka Derpy Horse), and tehhowch for giving helpful feedback.


Amazinite:
You usually open an issue to discuss things, not a PR with no functionality.
I'd suggest checking out https://github.com/endless-sky/endless-sky/issues/3381. I found it while looking around after a discussion on shy/elusive personalities.

source/AI.cpp
```diff
@@ -2595,6 +2600,22 @@ void AI::DoScatter(Ship &ship, Command &command)
...
+// Rove; i. e. wander aimlessly around the system
+void AI::Rove(Ship &ship, Command &command) 
+{
+	// If this is a carried ship (like a fighter), or if the ship is disabled or destroyed, don't rove.
+	if(ship.CanBeCarried() || ship.IsDisabled() || ship.IsDestroyed())
+		return;
+
+	// Create a random point and move to it.
+	int areaSize = 10000;
+	Point destination(Random::Real()*areaSize-areaSize/2, Random::Real()*areaSize-areaSize/2);
```
Amazinite, line 2612: This makes a square. You should be making a circle.
```suggestion
	Point destination = Angle::Random().Unit() * (areaSize * sqrt(Random::Real()));
```

source/AI.cpp
```diff
@@ -874,6 +874,11 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 			// is to convert completed MOVE_TO orders into HOLD_POSITION orders.
 			UpdateOrders(*it);
 		}
+
+		// Roving takes precedence over leaving the system, but the ship will default to attacking an enemy if there is one.
+		if(isPresent && personality.IsRoving() && !isStranded && !target)
+			Rove(*it, command);
+
```
Amazinite, line 881: You're splitting this if/else structure in half by putting this here. This also doesn't seem like the most appropriate place for this. Seems like it'd make more sense to me to move this up to where other personalities are checked. Lines 702 to 764 check if the ship has the swarming, harvests, surveillance, or mining personalities, then tells them to do something special based on that, so I'd imagine this would make more sense up there.

source/AI.cpp
```diff
@@ -2595,6 +2600,22 @@ void AI::DoScatter(Ship &ship, Command &command)
...
+{
+	// If this is a carried ship (like a fighter), or if the ship is disabled or destroyed, don't rove.
+	if(ship.CanBeCarried() || ship.IsDisabled() || ship.IsDestroyed())
+		return;
+
+	// Create a random point and move to it.
+	int areaSize = 10000;
+	Point destination(Random::Real()*areaSize-areaSize/2, Random::Real()*areaSize-areaSize/2);
+	Point velocity(0, 0);
+	MoveTo(ship, command, destination, velocity, ship.MaxVelocity(), ship.MaxVelocity());
```
Amazinite, line 2614: Points are created as (0,0), so no need to specify it. Also little need to create a variable you never actually use.
```suggestion
	MoveTo(ship, command, destination, Point(), ship.MaxVelocity(), ship.MaxVelocity());
```

source/AI.cpp
```diff
@@ -2595,6 +2600,22 @@ void AI::DoScatter(Ship &ship, Command &command)
 
 
 
+// Rove; i. e. wander aimlessly around the system
+void AI::Rove(Ship &ship, Command &command) 
+{
+	// If this is a carried ship (like a fighter), or if the ship is disabled or destroyed, don't rove.
+	if(ship.CanBeCarried() || ship.IsDisabled() || ship.IsDestroyed())
+		return;
```
Amazinite, line 2608: This is redundant with code that must be executed before reaching this point; a destroyed or disabled ship will never get past line 570 in Step. And this check to see if the ship can be carried should probably be done before figuring out if we should rove, as otherwise a carried ship might end up receiving no commands, as it enters the Rove function but gets kicked right back out and doesn't get sent anywhere else (at least how you currently have it set up).

source/AI.cpp
InfiniteDonuts: It has to be placed after the code for navigating/jumping out of the system, or it will not take precedence, correct?

source/Personality.cpp
```diff
@@ -78,7 +81,10 @@ namespace {
 		{"opportunistic", OPPORTUNISTIC},
 		{"target", TARGET},
 		{"marked", MARKED},
-		{"launching", LAUNCHING}
+		{"launching", LAUNCHING},
+		{"shy", SHY},
+		{"elusive", ELUSIVE},
+		{"roving", ROVING}
```
tehhowch, line 87:
```suggestion
		{"roving", ROVING},
```

tehhowch (changes requested):
In addition to the concerns raised by Amazinite:

What's the purpose of these 3 personalities? (What functional effect is each personality _intended_ to have? Are these personalities meant to combine in a specific way when mixed with any other particular personalities?)  

Right now, you say you're making these for a different PR, and that PR says you're implementing the personalities, but neither of you is specifying the desired-yet-unavailable functionality, the _reason we need these in the first place_.

Without this information, no one can gauge if your implementation satisfies the specification.

source/Personality.cpp
Galaucus: Shy: The ship attempts to avoid non-friendly vessels, retreating to a safe minimum distance if any come too close.
Elusive: Ship prioritizes maintaining cloak regardless of fuel cost. When attacked it will prioritize disengaging from combat and raising cloak.
Roving: A general purpose "mill about aimlessly" behavior to put a bit of life into fleets inhabiting empty systems. 

Shy and elusive, taken together, is intended ensure that when ion storms force ships to decloak they don't do so right under the player. 

Roving is meant to add movement to indigenous creatures who otherwise have no goal to pursue in the system where they're found.

source/AI.cpp
InfiniteDonuts: I for one, Derpy, am extremely interested in keeping code organized. But for some reason (which I haven't figured out yet) if I _don't_ put the roving if statement there, then the test ships either don't rove and just leave the system (without a continue statement), or else they just drift in space (with a continue statement).

source/AI.cpp
```diff
@@ -2595,6 +2610,34 @@ void AI::DoScatter(Ship &ship, Command &command)
...
+void AI::DoElusive(Ship &ship, Command &command)
+{
+	// Cloak if you can.
+	if(ship.Attributes().Get("cloak"))
+		command |= Command::CLOAK;
+	
+	// Cloak if your parent is cloaked.
+	const shared_ptr<const Ship> &parent = ship.GetParent();
+	if(parent && parent->Commands().Has(Command::CLOAK))
+		command |= Command::CLOAK;
```
tehhowch, line 2623: The second part of this adds no value -- if the ship is capable of cloaking, then it will already have set the cloak command. If it isn't capable of cloaking, then it won't have the command set, and setting it because the parent is attempting to cloak is meaningless.

tehhowch (changes requested):
If `personality elusive` just means "cloak whenever possible," then "elusive" is a bad name for the desired functionality. If the goal is a way for a content creator to specify "ships with this flag always cloak whenever possible", that personality token should be named appropriately, e.g. "cloaking" or "always cloaking" or "hiding".  
Further, if that is the only behavior of the personality, then the personality trait should be checked during normal parts of the cloaking code, i.e. `DoCloak`, rather than in its own section.

I have similar concerns about "shy," because "shy" and "timid" are synonyms. It's almost certain one will be used when the behavior of the other was intended.

>Shy: The ship attempts to avoid non-friendly vessels, retreating to a safe minimum distance if any come too close.

"evasive" would fit better, IMO. Note that as described, this personality will have no impact on movement unless the ship encounters hostiles in the system (merchant ships and the player ships are friendly, hai and merchants are friendly, etc). What's the expected behavior of shy/evasive ships w.r.t. to hostile ships? If it's "fight to survive," then this sounds like a personality flag that has no actual behavior of its own, i.e. one that we do not need at all (because it does nothing). If it's "don't fight," then that's the "pacifist" or "forebearing" flag. If it is "run away," that's the "fleeing" flag.

source/AI.cpp
```diff
@@ -762,6 +762,14 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 			else
 				it->SetTargetAsteroid(nullptr);
 		}
+
+		// Handle roving ships
+		if(personality.IsRoving() && it && !it->CanBeCarried() && !parent && !isStranded && !target && isPresent)
```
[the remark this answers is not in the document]
InfiniteDonuts: Ships that have a parent will NOT rove and will instead return to default behavior (e.g., following its parent around).

source/AI.cpp
InfiniteDonuts: gotcha

Zitchas:
The "shy" personality was intended to be a step farther than that. Basically, instead of the usual "avoid hostiles and ignore everyone else" behaviour that is the current standard, the "Shy" personality would be "avoid hostile and everyone that isn't explicitly friendly to it." In other words, as far as the "avoidance" behaviour is concerned, it treats neutrals, passives, and everyone that isn't an actual ally as an enemy to be avoided. Xenophobic might be another way to describe it.

It might be simpler to define it as simply "treats anyone not of the same government as an enemy for the purposes of trying to avoid them."

I see the "shy" behavior as being primarily a non-combat behavior. If engaged in combat I'd expect it to fight as normal (or as dictated by other personalities). 

Also, why close this?

Galaucus:
> ~Also, why close this?~

From what I understand the files got corrupted. Donuts intends to try again.

Zitchas:
Yeah, I read the reason on the discord. Shame to see hard work get messed up. Anyway, hopefully Donuts can get everything back up and running shortly.

PR #5591, opened 2020-12-31
Build flag to remove support for music and libmad

thomasballinger:
## Summary
Adds a build flag for building with no music, removing the need for libmad.

Since only one music track is included in the game, `sounds/ambient/machinery` which plays on space stations, this might be useful for creating builds without mad in environments where mad is not available.

## Testing Done
I built this locally on a mac, works both ways.

Since this PR introduces no automated testing for this codepath, someone could inadvertently break something. I'm hoping this will eventually be addressed by adding a CD artifact of a web build, the motivation for this PR.

## Performance Impact
I can only imagine it helps!

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -34,6 +34,7 @@ if 'steamrt' in chroot_name:
 opts = Variables()
 opts.AddVariables(
 	EnumVariable("mode", "Compilation mode", "release", allowed_values=("release", "debug", "profile")),
+	EnumVariable("music", "Whether to use music", "on", allowed_values=("on", "off")),
 	PathVariable("BUILDDIR", "Directory to store compiled object files in", "build", PathVariable.PathIsDirCreate),
 	PathVariable("BIN_DIR", "Directory to store binaries in", ".", PathVariable.PathIsDirCreate),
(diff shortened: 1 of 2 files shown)


tehhowch:
This seems like a really hacky way to disable music. Wouldn't it be better to make loading libmad optional, and avoid doing so when the compiled binary receives the command line parameter e.g. `--no-music`?

Then we don't need any ifdefs at all, and systems that won't attempt to load the dynamic library won't need the library in their execution environment.

A different take on the ifdef approach here would be to ifdef the entire music.cpp and music.h files, and their associated includes / method calls. Then the compiler wouldn't ever see calls to `Music::____`, and there wouldn't be anything at all in music.cpp or music.h

Only the first approach would make it into the upstream here, but the second way could simplify the patch you need to use for the web fork. Though since you would need to modify audio initialization, perhaps not.

thomasballinger:
That sounds great.

Do you happen to know of an example of dynamically loading a library I should follow when giving this a shot?

thomasballinger:
This ifdefs method is motivated by the [-DNO_THREADS PR](https://github.com/endless-sky/endless-sky/compare/master...thomasballinger:no-threads?expand=1), so I've opened https://github.com/endless-sky/endless-sky/pull/5594 for further discussion.

PR #5592, opened 2020-12-31
Experimental web build CD GitHub action

thomasballinger:
Not sure this actually works, I probably need to test this in my own repository since running unreviewed actions seems like a security risk.


thomasballinger:
I'll play with this on my own fork

PR #5594, opened 2021-01-01
No music and fewer threads

thomasballinger:
I found thinking about https://github.com/endless-sky/endless-sky/pull/5591 confusing without the -DNO_THREADS commit as well, so combining these here. I'm looking for any comments about this approach or other potential approaches. At this point it's a given that littering the code ifdefs like this is unlikely to be accepted, so I'm interested in high-level feedback. Finding a subset of changes that could be merged to make the web build patch smaller would also be useful.

@tehhowch has already left several comments about this.
> It's basically correct to assume that removing threads is not upstreamable. What you could/should do is work out a way to convert our use of threads into something compatible with e.g. WebWorkers, so you could upstream a conversion to "EsThread" from std::thread, and in the web build you'd replace the EsThread implementation with WebWorker stuff

and re music on https://github.com/endless-sky/endless-sky/pull/5591
> This seems like a really hacky way to disable music. Wouldn't it be better to make loading libmad optional, and avoid doing so when the compiled binary receives the command line parameter e.g. --no-music?
>
> Then we don't need any ifdefs at all, and systems that won't attempt to load the dynamic library won't need the library in their execution environment.
> 
> A different take on the ifdef approach here would be to ifdef the entire music.cpp and music.h files, and their associated includes / method calls. Then the compiler wouldn't ever see calls to `Music::____`, and there wouldn't be anything at all in music.cpp or music.h
>
> Only the first approach would make it into the upstream here, but the second way could simplify the patch you need to use for the web fork. Though since you would need to modify audio initialization, perhaps not.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -34,6 +34,7 @@ if 'steamrt' in chroot_name:
 opts = Variables()
 opts.AddVariables(
 	EnumVariable("mode", "Compilation mode", "release", allowed_values=("release", "debug", "profile")),
+	EnumVariable("music", "Whether to use music", "on", allowed_values=("on", "off")),
 	PathVariable("BUILDDIR", "Directory to store compiled object files in", "build", PathVariable.PathIsDirCreate),
 	PathVariable("BIN_DIR", "Directory to store binaries in", ".", PathVariable.PathIsDirCreate),
 	PathVariable("DESTDIR", "Destination root directory, e.g. if building a package", "", PathVariable.PathAccept),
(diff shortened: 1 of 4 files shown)


thomasballinger:
I'm interested in @tehhowch's suggestion of dynamically loading libmad. Maybe mocking out a header to get it to compile would be useful? I need to learn about how the build process will work, does it sound right that the only thing needed to make things compile is a libmad header file?

thomasballinger:
By using Emscripten's [pthreads support](https://emscripten.org/docs/porting/pthreads.html) support, this update to the PR cuts out many of the ES_NO_THREADS ifdefs. Compared with the no threads version, this runs a loads a little slower and takes a bit more memory. 

The introduction of pthreads and [asyncify](https://emscripten.org/docs/porting/asyncify.html) to this port brings the changes that would required for a web build down to the following:
- [OpenGL ES Support](https://github.com/endless-sky/endless-sky/pull/5588)
- [RGBA textures instead of BGRA](https://github.com/endless-sky/endless-sky/pull/5153)
- [IFDEFs removing music and some uses of threads](https://github.com/endless-sky/endless-sky/pull/5594) (this PR)
- some build file changes (no PR, see [this commit](https://github.com/endless-sky/endless-sky/commit/cb659e10e1d58616239a360a1463ffd1986462fe) for a preview

There are [other pieces](https://github.com/endless-sky/endless-sky/compare/master...thomasballinger:thread-experiment?expand=1) (mostly a bunch of JavaScript) used for [test-endless-web.com](https://test-endless-web.com/), but these are the only changes required to create a web build.

I'd love to hear suggestions for direction: is there a refactor that would make the optional use of threads more palatable? @tehhowch's suggestion of dynamically loading libmad.a is still preferable to these ifdefs I can pursue that direction next.

thomasballinger:
Closing, and I'll post something smaller soon.

PR #5595, opened 2021-01-01
Big cpp changes for a web build

thomasballinger:
Since discussion in PRs seems to work well, here's the whole* set of changes to the cpp code that I'm using to make a web build. As mentioned in https://github.com/endless-sky/endless-sky/pull/5594 it's a given that these changes won't be accepted as-is, but I'd love to hear what concerns folks have and hear ideas for working around them. Any discussion welcome! 

The biggest changes are not using threads (already covered in https://github.com/endless-sky/endless-sky/pull/5594, and might be able to be overcome with experimental [Emscripten pthreads support](https://emscripten.org/docs/porting/pthreads.html) I need to look into) and changes to main.cpp to make the forever loop into iterative coroutine-like thing that can be called 30 or 60 times/second.

*If you want to test this, use https://github.com/thomasballinger/endless-sky/tree/es-wasm-reformatted; this PR doesn't include things like https://github.com/endless-sky/endless-sky/pull/5153 and https://github.com/endless-sky/endless-sky/pull/5588 that are required for a web build to work and a bunch of other build script changes.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -35,6 +35,7 @@ opts = Variables()
 opts.AddVariables(
 	EnumVariable("mode", "Compilation mode", "release", allowed_values=("release", "debug", "profile")),
 	EnumVariable("music", "Whether to use music", "on", allowed_values=("on", "off")),
+	EnumVariable("threads", "Whether to use threads", "on", allowed_values=("on", "off")),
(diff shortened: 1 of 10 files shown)


thomasballinger:
Closing this, now with ASYNCIFY working the diff is about half the size, with nearly all the changes to main (making while(true) loop iterative) gone!

PR #5597, opened 2021-01-01
feat: Implement play time tracking for pilots

Terin:
## Feature Details
This PR introduces playtime tracking per-pilot, as shown on the main menu panel, or load panel. This is useful for players who tracked playtime through Steam but want to switch to a continuous build, who want to see how long it's been between snapshots on their pilot, or even rough speedruns.

It adds a loadTime and playTime integers to the PlayerInfo class, where loadTime tracks when the player last loaded or unpaused their game, and playTime is incremented on saves or pauses, based on the difference between when that action happens and loadTime. This is tracked as a "playtime" attribute integer in saves, and is formatted into a Years:Days:Hours:Minutes:Seconds style in-game.

## UI Screenshots
Before/After
image

A case filling out all units of time
image

## Testing Done
- Loaded into the game with a pilot with playtime on record, and without
- Swapped between MenuPanel and LoadPanel to see no differences
- Verified play time was saved on take off, when pausing/loading into a different save on planet, and when quitting the game on planet
- Created new pilots to verify time started from 0s properly

One unresolved issue remains, where I do not know how to detect if the player is "paused" (on MenuPanel) while in main.cpp. This occassionaly results in an extra second being added to playTime when quitting from there, due to lag while quitting. If someone knows of a conditional statement to check that, then UpdatePlayTime() can be run only when "unpaused", preventing that bug.

## Save File
This zip contains multiple snapshots of one pilot, each with play time that is about to tick over into a new unit (minutes, hours, etc.). It also contains ones with no play time on record, so you can verify it will get properly tracked and written to the save after running from it.
[Playtime Test Pilot.zip](https://github.com/endless-sky/endless-sky/files/5759484/Playtime.Test.Pilot.zip)

## Performance Impact
None noticeable. The most intensive calculation is updating loadTime each frame on MenuPanel, just calls time().


source/PlayerInfo.cpp
```diff
@@ -95,13 +95,14 @@ void PlayerInfo::New()
 	
 	SetSystem(start.GetSystem());
 	SetPlanet(start.GetPlanet());
+    UpdateLoadTime();
 	accounts = start.GetAccounts();
```
tehhowch, line 99: Spaces

source/text/Format.cpp
```diff
@@ -89,6 +89,32 @@ string Format::Credits(int64_t value)
 
 
 
+// Convert a time in seconds to years/days/hours/minutes/seconds
+std::string Format::PlayTime(int time)
```
tehhowch, line 93: We probably should use a different parameter name, given that elsewhere we're using `time` to refer to a function.

source/text/Format.cpp
```diff
@@ -89,6 +89,32 @@ string Format::Credits(int64_t value)
...
+{
+	string result;
+	static const vector<char> SUFFIX = {'s', 'm', 'h', 'd', 'y'};
+	static const vector<int> PERIOD = {60, 60, 24, 365};
```
tehhowch, line 97: We should probably use `std::array` instead of `std::vector` here.
(We can clean up the other similar instances in this file in a follow-up PR.)

tehhowch (changes requested):
1. Add unit tests for the new format method. You'll need to create `tests/src/text/test_format.cpp`
2. There's some unintended formatting changes in a couple spots.
3. It's not strictly required, but I wonder if we should use `std::time` instead of just `time` as the function reference.

source/text/Format.cpp
Terin: Would playTime work, if that's also the name of the variable being (indirectly) passed? Trying to think of alternatives that aren't too wordy.

tehhowch: Yes, that would work, or e.g. `timeVal`. My concern was that here we are declaring a function parameter named `time` of type `int`, while in other code we are using a variable named `time` that is a function with a particular signature.

The names can be equivalent with without technical issue (the compiler can deduce what `time` should mean in both cases), but my concern was for the sake of a user that is looking at the code.

tehhowch (commented):
Thanks for writing unit tests! As a field read from a save, the input playtime could be bad, too - negatives or unrepresentable numbers. We should test the behavior of these cases. Would you mind writing some benchmark tests too? (There are some for the Random functions you can work from. I'm thinking one benchmark for "normal" playtimes, around 40 - 400 hrs, and one for absurd times, i.e. `numeric_limits<int>::max()`.)

source/text/Format.h
```diff
@@ -13,6 +13,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #ifndef ES_TEXT_FORMAT_H_
 #define ES_TEXT_FORMAT_H_
 
+#include <array>
```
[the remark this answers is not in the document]
Terin: Ah, was going by `#include <vector>` being in the header file, as well.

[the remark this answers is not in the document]
tehhowch: Yeah, `map` & `vector` are necessary here because they are return values or by-value parameter types used in this class's method signatures. Unlike them, `std::array` is used as an implementation detail and doesn't need to be advertised to other classes.

Terin:
All refactored now and much cleaner, especially when I realized being "paused" simply meant menuPanels having anything in it. Got some benchmarking tests in as well, but wasn't sure if I got the final syntax correct, so please let me know if anything needs to be fixed there. Besides some obnoxious whitespace issues caused by VS Code, I think it should be good to go!

source/PlayerInfo.cpp
```diff
@@ -1438,6 +1440,19 @@ bool PlayerInfo::TakeOff(UI *ui)
 
 
 
+void PlayerInfo::AddPlayTime(double timeVal)
```
tehhowch, line 1443: This should take a `chrono` duration type, to prevent ambiguity.

tests/src/text/test_format.cpp
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+// #endregion unit tests
+
+// #region benchmarks
+#ifndef CATCH_CONFIG_ENABLE_BENCHMARKING
+TEST_CASE( "Benchmark Format::PlayTime", "[!benchmark][random]" ) {
+	BENCHMARK( "Format::PlayTime() with a value under an hour" ) {
+		int randTime = Random::Int(3600)
```
tehhowch, line 76: For more-comparable benchmarks (e.g. between different CI runs), we should remove explicit randomness

tests/src/text/test_format.cpp
```diff
@@ -0,0 +1,90 @@
...
+		THEN( "0s is returned " ) {
+			CHECK( Format::PlayTime(-300) == "0s");
+		}
+	}
+}
+
+// #endregion unit tests
+
+// #region benchmarks
+#ifndef CATCH_CONFIG_ENABLE_BENCHMARKING
```
tehhowch, line 73:
```suggestion
#ifdef CATCH_CONFIG_ENABLE_BENCHMARKING
```

tests/src/text/test_format.cpp
```diff
@@ -0,0 +1,90 @@
...
+TEST_CASE( "Benchmark Format::PlayTime", "[!benchmark][random]" ) {
```
tehhowch, line 74:
```suggestion
TEST_CASE( "Benchmark Format::PlayTime", "[!benchmark][format]" ) {
```

tehhowch (approved):
I let the game idle for several hours and didn't see a noticeable drift between wall clock and recorded playtime.


Some final comments / questions, though:

PR #5600, opened 2021-01-02
add a dialog 'unable to switch vsync state' to settings

tatami4:
This PR adds a dialog to Settings that pops ups when VSync state cannot be changed.
I added this popup because otherwise the UI appears unresponsive.

### DIFF
--- a/source/PreferencesPanel.cpp
+++ b/source/PreferencesPanel.cpp
@@ -173,7 +173,10 @@ bool PreferencesPanel::Click(int x, int y, int clicks)
 					while(Preferences::ZoomViewOut()) {}
 			}
 			else if(zone.Value() == VSYNC_SETTING)
-				Preferences::ToggleVSync();
+			{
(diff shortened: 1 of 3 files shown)


source/PreferencesPanel.cpp
```diff
@@ -173,7 +173,10 @@ bool PreferencesPanel::Click(int x, int y, int clicks)
 					while(Preferences::ZoomViewOut()) {}
 			}
 			else if(zone.Value() == VSYNC_SETTING)
-				Preferences::ToggleVSync();
+			{
+				if(!Preferences::ToggleVSync())
+					GetUI()->Push(new Dialog("Unable to change VSync state"));
+			}
```
tehhowch, line 179:
```suggestion
			else if(zone.Value() == VSYNC_SETTING && !Preferences::ToggleVSync())
				GetUI()->Push(new Dialog("Unable to change VSync state. (Your system's graphics settings may be controlling it instead.)"));
```

PR #5601, opened 2021-01-03
Apply automatic crew rehiring after completing transport missions.

yobbo2020:
Desired crew member count is now stored per flagship, and set when hiring/firing rather than on takeoff. It thus sensibly persists over both save/load and fleet reordering.

-----------------------
**Feature:** I didn't make a separate issue for this, I just implemented it.

## Feature Details
Added internal behaviour to automatically rehire crew when space has been freed up after completing a transport mission. Desired crew complement is set whenever the player hires / fires crew. Previously it was set on takeoff, hence failing to restock crew after transport missions. I also changed it to be stored as a per-ship variable in stead of globally for the player, so the player's desired crew complement will be sensibly remembered per-flagship, and will also persist over save / reload.

## Usage Examples
It's completely automatic and just attempts to keep the number of crew set when hiring / firing whenever it can do so. As before it only applies if the relevant preference is set.

## Testing Done
I played a significant portion of the Free Worlds campaign without noticing a problem with the modified behaviour. It can be easily tested by hiring a full or nearly-full crew complement then doing a transport mission, accepting when it says it needs to fire some of them. I also tested with multiple simultaneous missions, and it rehired correctly as soon as bunks were freed.

## Performance Impact
none


source/Ship.cpp
```diff
@@ -797,6 +799,8 @@ void Ship::Save(DataWriter &out) const
 		
 		cargo.Save(out);
 		out.Write("crew", crew);
+		if(Preferences::Has("Rehire extra crew when lost"))
```
petervdmeer, line 802: Let's only write this value out when it is non-zero (and let's write it out regardless of if the Preference was set).

source/HiringPanel.cpp
```diff
@@ -110,11 +110,13 @@ bool HiringPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, b
 	if(key == 'h' || key == '=' || key == SDLK_RETURN || key == SDLK_SPACE)
 	{
 		player.Flagship()->AddCrew(min(maxHire, Modifier()));
+		player.Flagship()->SetDesiredCrew(player.Flagship()->Crew());
 		player.UpdateCargoCapacities();
 	}
 	else if(key == 'f' || key == '-' || key == SDLK_BACKSPACE || key == SDLK_DELETE)
 	{
 		player.Flagship()->AddCrew(-min(maxFire, Modifier()));
+		player.Flagship()->SetDesiredCrew(player.Flagship()->Crew());
```
petervdmeer, line 119: If the player reduces the `DesiredCrew` to the `Minimum Required Crew`, then let's set the `DesiredCrew` value to 0 (since the player is then likely not interested in hiring more crew than needed).

source/Ship.h
```diff
@@ -305,6 +305,9 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	void AddCrew(int count);
 	// Check if this is a ship that can be used as a flagship.
 	bool CanBeFlagship() const;
+	// Get and set the number of desired crew members when this is a flagship.
```
petervdmeer, line 308: We might want to use this for non-flagships as well in the future (if defense from capturing becomes relevant again).

```suggestion
	// Desired crew members for this ship. Having more than the minimum crew increases the odds when capturing and when defending from capture. (Currently only used for flagships.)
```

petervdmeer (changes requested):
This looks like a good extension to make the "Rehire extra crew when lost" preference work as it was intended.

source/Ship.h
```diff
@@ -545,6 +551,7 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	Point acceleration;
 
 	int crew = 0;
+	int desiredCrew = 0;
```
tehhowch, line 554: This doesn't belong on the ship model. It's a player-only consideration, and should be kept in the player model.

source/HiringPanel.cpp
```diff
@@ -106,11 +106,20 @@ bool HiringPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, b
...
 		player.UpdateCargoCapacities();
 	}
 	else if(key == 'f' || key == SDLK_MINUS || key == SDLK_KP_MINUS || key == SDLK_BACKSPACE || key == SDLK_DELETE)
 	{
 		player.Flagship()->AddCrew(-min(maxFire, Modifier()));
+		// If the ship has more than the minimum required crew, then set the desired crew
+		// to the value the player requested. If the ship has the exact amount (or even
+		// less), then the planer apparently doesn't want to hire any extra crew (and
+		// thus we set the desired crew value to zero).
```
tehhowch, line 118: This is overly verbose for the code that follows. Comments need not say what the code will do; they need to provide the context of the _why_.

(I think resolving the other feedback will probably remove the need for this comment at all.)

tehhowch (changes requested):
@petervdmeer some feedback for you to address

Amazinite:

Issue #4714, opened 2019-12-10
Optional individual turret tracking setting for each turret type


starscollide5:
**Is your feature request related to a problem? Please describe.**

After turret tracking nerf some turrets turn too slow to be effective in opportunistic mode and require focused mode to be effective - while others remain perfectly viable in opportunistic mode. 

On flagships with mostly turret hardpoints, having different types of turrets e.g. 2x fast turrets with mediocre damage for fast targets and 2x slow but powerful turrets for main target - to compensate for lack of gun hardpoints - is essential, but on either setting one of the turret flavors - and overall ship efficiency - suffers greatly. 

The situation is particularly dramatic for Korath Disruptor turret, making it pretty much useless in opportunistic mode for solo players or small fleets - when player's flagship usually needs protection from drones and fighters, thus fast turrets in opportunistic mode are a necessity. 

Turrets that particularly benefit from focused mode:
* Dual Sunbeam Turret
* Electron Turret
* Korath Disruptor
* Plasma Turret
* Sunbeam Turret

Turrets that particularly benefit from opportunistic mode:
* Heavy Laser Turret 
* Korath Repeater Turret
* Laser Turret
* Moonbeam Turret
* Proton Turret
* Pug Gridfire Turret
* Pug Zapper Turret
* Pulse Turret
* Quad Blaster Turret
* Quarg Skylance
* Thrasher Turret

Turrets that benefit from either mode depending on the situation:
* Bombardment Turret
* Heliarch Attractor
* Heliarch Repulsor
* Korath Banisher
* Korath Grab-Strike
* Korath Slicer Turret

**Describe the solution you'd like**

I propose "custom" turret tracking setting activating advanced menu with individual turret tracking setting for each turret type. 

**Additional context**

This change will not affect overall balance of the game in any noticeable way, but will help and improve usable outfits diversity for players who fly solo or with fleets small enough that more effective turret control on flagship can make any impact. Particularly, this feature would make Korath Disruptor less useless in described circumstances, and make turrets with slow tracking, and consequently ships with mostly turrets hardpoints, less undesirable compared to their gun counterparts.

Zitchas:
Just to clarify, you are asking for:

- An additional attribute for turrets that identifies them as either focused or opportunistic
- An additional setting for opportunistic firing that tells the game to look at this attribute and use it to determine whether or not a turret behaves opportunistically or not.

[...]

Issue #4412, opened 2019-07-04
Automatic Firing - On Demand option


ZBok:
The current AI governing the "Automatic Firing" option is outstanding. It allows separate weapons with different ranges to be fired independently when they are in range of a target, and only when the ship/turrets are properly aimed.

For example, if player prefers a kiting style with Particle Cannons, but with some closer range turrets equipped for anti-fighter and extra brawl DPS, this option allows for the player to kite at long range, firing only the Particle cannons, minimising energy/heat by not firing the blasters until the target is in range of these weapons too.

However, Automatic Firing takes away a lot of the feel of combat and generally feels lacklustre. It also introduces small problems like not firing some guns (especially on a ship where the guns have large transverse separation) at very small targets, even if the gun inaccuracy makes them likely to hit, and also sometimes the AI will decide to fire on a disabled hostile ship just after it has been boarded. These are however just minor problems.

I propose a third option in the "Automatic Firing" setting - "on fire key"

Making the options "on/off/on fire key"

This would cause the Fire key to behave like Automatic firing, as described above, but on demand.

Amazinite:
>**Related Issue Links**
None I know of.

https://github.com/endless-sky/endless-sky/issues/166 ?

ChamEV:
So essentially, your ship only attacks when you hold down fire and ships are in range.  As a third option for automatic firing that would probably work.

jafdy:
#1950

ChamEV:
This might be controversial, but what if it was implemented so that it could only be set by editing your preference file.  If set, it would show up in preferences, but if you clicked change in preferences, it would toggle between on and off.

Tadrix:
> So essentially, your ship only attacks when you hold down fire and ships are in range. As a third option for automatic firing that would probably work.

I'd recommend that in such a scenario the fire key becomes a toggle: fire at will/hold fire.

ZBok:
I see now that in #1950 MZ mentions this kind of option would be too complicated, with that being noted should this just be closed or is further discussion warranted?

jafdy:
I think this is effectively a duplicate (but with a much better write up).

ZBok:
@Tadrix I would suggest not making it a toggle. The primary reason for my posting this feature request is that Automatic Firing, while in some ways a solution to the initial problem, takes away from the combat - it feels much nicer when you are actually giving the command to fire yourself.

[...]

PR #5606, opened 2021-01-03
artwork: Add new icons for hardpoint firingmode settings.

petervdmeer:
**Feature:** This PR implements icons for the feature discussed in #4714

## Feature Details
This PR adds 6 new icons:
- firingmode_focussed.png; for focussed turret mode as discussed in #4714
- firingmode_gated.png; for focussed turret mode, but where the ship only fires if it can hit an enemy/target
- firingmode_directed.png; for focussed turret mode, but where the ship automatically fires
- firingmode_opportunistic.png; for opportunistic turret mode as discussed in #4714
- firingmode_semi_automatic.png; for opportunistic turret mode, but where the ship only fires if it can hit an enemy/target
- firingmode_full_automatic.png; for opportunistic turret mode, but where the ship fires fully automatically

The icons are primarily meant for turrets, but the 3 focussed icons could also be applied to guns.

## UI Screenshots
Here is a screenshot of the icons used on a player-screen. For the final version I expect the icons to be on the other side of the text (closer to the ship and align with all other icons):
example_icons_usage

## Usage Examples
Implement the feature as discussed in #4714 and use the icons as provided here.

## Testing Done
The UI screenshot above can be considered a graphical test.

## Performance Impact
N/A


Amazinite:
> For the final version I expect the icons to be on the other side of the text (closer to the ship and aligned with all other icons below and above it):

:+1: Beat me to it.

Amazinite:
Have high DPI versions of these?

petervdmeer:
> Have high DPI versions of these?

Not yet, but I can create high DPI versions.

> I think the "progression" of these images should be the same for both focused and opportunistic fire. That is, for the focus fire we have a circular crosshair, a circular crosshair with a ship in the middle, then a squared crosshair with a ship in the middle.

Yes, that was my original idea as well.

> For opportunistic fire we have an arc of fire, an arc of fire with a ship in the middle, and then... a squared off arc of fire. We don't have that same ship in the middle graphic on the final opportunistic image. It looks like right now that's because having it would obscure the arc of fire part of the image, but perhaps the solution is to not have the square off part obscure the arc of fire so much.

This is exactly why I removed the ship in the middle, but I agree that it would be nicer to indeed use the same progression.

> Outside of that though, I'm not really sure what the purpose of the new four firing modes are. Isn't "only fire when you can hit" already the behavior? Is the purpose of this to allow the player to have some turrets be AI controlled (auto-fire is on) while others are player controlled (auto-fire is off)?

Yes, the purpose of this is indeed to allow some turrets/guns to be fully AI controlled (auto-fire), some turrets/guns to be on-demand AI controlled (as requested here https://github.com/endless-sky/endless-sky/issues/4412) and some/guns turrets to be fully manually controlled.

petervdmeer:
Added HDPI versions and changed the icons to show the same progression for focused and for opportunistic fire.

Amazinite:
I'm inclined to say that we shouldn't merge these icons until the feature they're intended for is actually working. I suspect that having a working feature would also invite more feedback on the images themselves, as, at least here, there hasn't been much in the way of feedback. Would you mind if this PR were closed?

Zitchas:
Perhaps more appropriate to shift it to "draft" as these are definitely desired things and (hopefully, eventually) needed... Just not ready yet. Which seems like the definition of draft.

PR #5607, opened 2021-01-04
Increase deterrence from relative damage weapons

Terin:
## Feature Details
This PR modifies the damage value used in deterrence calculations, by letting weapons that deal relative damage add to your fleet's DPS calculate it based on the vitals of the ship that's carrying it. While this is theoretically game-able by having it change based on what ship it's installed on, it would still let a plugin creator use small values of relative damage without completely blowing out the piracy threat scale, and would be better than any constants we could come up with.

## UI Screenshots
image
image

## Testing Done
As shown above, I tested by using two weapons (Deterrence from Omnis, which does 100% damage 60 times a second, and Deterrence Mini, which does 10% damage 1 time a second) on two different types of ships, the Star Barge and the Archon. I parked and unparked various combinations of those ships to see the fleet threat increase appropriately.

## Save File
This zip has a save file with said ships, and a plugin which provides the outfits, so you can see the deterrence changes yourself.
[relative-deterrence.zip](https://github.com/endless-sky/endless-sky/files/5762768/relative-deterrence.zip)

## Performance Impact
N/A

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -41,6 +41,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include <algorithm>
 #include <cmath>
 #include <ctime>
+#include <iostream>
 #include <sstream>
 
 using namespace std;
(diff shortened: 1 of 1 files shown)


petervdmeer (approved):
The change looks very logical. You might want to lose some brackets since multiplication takes precedence over addition, but I'm personally also okay with the brackets staying as-is.

Looks good to me.

PR #5609, opened 2021-01-04
FTL Fuel Cost Attributes

Ferociousfeind:
**Feature:** This PR aims to allow hyperdrives and jump drives to consume more or less fuel based on two factors, in two ways.

## Feature Details
This feature adds five attributes-
- "drive mass exponent"
- "drive mass reference"
- "drive distance exponent"
- "drive distance reference"
- "jump startup fuel"

which refer to the ship's mass, and the distance from the ship to its destination system.

The reference attributes are anchors- they specify where changes will not affect fuel cost. Uh, sort of? If your ship has a total mass of 500, and the "drive mass reference" is 250, with a "drive mass exponent" of 4, then the jump your about to make will cost `(500/250)^4`=`2^4`=`16` times as much as it normally would- likely, 1600 fuel, instead of 100.
Generalized, the mass of your ship, or distance to your destination, is divided by the mass/distance reference, that number is raised to the power of the mass/distance exponent, and that value is multiplied to your fuel consumption. Because the exponent isn't simply a boolean taking either 0 (no effect) or 1 (which could've been a linear or exponential relationship), it can handle more cases than a simple boolean could've. Because the default values for the exponent, if otherwise not specified, is 0, this change should not impact vanilla ES in any way.

The exponent value can be set in any number of ways, including...

- Exponential: "exponent" > 1, doubling the related attribute more than doubles fuel use
- Linear: "exponent" = 1, doubling the related attribute doubles fuel use
- Exponential Decay: 0 < "exponent" < 1, to double the fuel use, you need to more than double the related attribute
- None: "exponent" = 0, regular, default behavior, mass/distance does not affect fuel use

The attributes are multiplicative against each other, e.g. having a 2x multiplier from mass and a 3x multiplier from distance gives a total (2*3=)6x multiplier on fuel use, instead of (2+3=)5x

Edit:
Jump startup fuel is a secondary "jump fuel" attribute which is unaffected by either the mass of your ship, or the distance to your destination. I suggest you, uhh, don't use it instead of regular jump fuel.

## UI Screenshots
N/A

## Usage Examples
Here is an example outfit. As explained in the description, doubling the mass of your ship with this equipped will triple fuel costs. At 2000 tons, for example, it will cost 900 fuel per jump. At 4000 tons, 2700 fuel. At 8000 tons, 8100 fuel! Though, I suspect there aren't that many ships weighing 8000 tons. 
```
outfit "Mass Drive"
  category "Systems"
  "mass" 10
  "hyperdrive" 1
  "jump fuel" 100
  "drive mass reference" 500
  "drive mass exponent" 1.585
  description "This hyperdrive is designed to use very little fuel at low ship masses, but as your ship's mass increases, its fuel use increases as well.
  description "Avoid using this drive at high masses- doubling the mass of your ship will triple fuel use per jump, and that will add up quickly."
```
(at ~7830 tons, this ship will consume ~7830 fuel, by the way)

Using "jump distance" does not interference with "drive distance reference" and "drive distance exponent", and the combination of the four attributes will not interfere with each other- they're built to work together!

## Testing Done
I have brought many ships- heavy and light- into space, and had them jump near and far, to observe that the fuel usage is what is expected. I've also used edited drives which lack one of the two pairs of new attributes (however I have not actually tried one half of the pairs without the other half, it is irresponsible to do so anyways). I've selected multiple nearby systems with different distances and observed that the fuel gauge properly updates to reflect fuel consumption if you jumped in that moment.

Edit:
The PR works exactly as intended, both mass and distance work to increase or decrease fuel use at every point that Ship::BestFuel is called, which is pretty much everywhere. If either of the attributes is negative (or zero), they are set to their minimum values (0 for exponents, 0.01 for references) so that they do not catastrophically break the game.

## Performance Impact
"Play stupid games, win stupid prizes"- because this adds a long formula into the otherwise-sleek Ship::BestFuel method, if you have hundreds of JDs installed and they all have these four attributes, your game might chug a little as you're pressing J. (Having a large jump range does worse to the game than absurd exponents do anyways.)


Zitchas:
I feel like your explanation of how these variables work needs a graph. If I'm understanding it correctly, this makes fuel costs something like:
y = ax^b
where:
y = fuel cost to jump
x = mass of the ship
a = constant that dictates how fast it ramps up as mass increases
b = constant that dictates whether it is a line or a curve, and if a curve, how steep of a curve.

I think I only vaguely understand what is going on here, but I like the idea.

Ferociousfeind:
Yeah, that's pretty much how it works, though be sure to put parentheses around (ax), since the reference applies before the exponent, like this:
image
y= fuel use
x= _specified_ fuel use in the file
r1= drive [name] reference
e1= drive [name] exponent

With example data, reference of 200 tons, exponent of 0.5, the curve would look like this (kudos to desmos for displaying this)
image

Ferociousfeind:
> Isn't this making the drives (of whatever flavor) nothing more that reaction/thrust drives? Isn't using a hyperlink drive basically "hitching a ride" on the hyperlink between the stars, and thus it's the link that is providing the movement? And a JumpDrive is sort of like creating a miniature, unstable wormhole? Again, the SHIP/DRIVE isn't doing the movement, but rather taking advantage of something else that isn't constrained by the mass of the ship(s). The early spaceships have such reaction/thrust drives that are affected by mass and inertia. But it also meant that Captain Greatgrand Child was the one that landed the ship, no?

Um... no. First of all, this is an entirely optional change that, as it is, changes nothing about vanilla. It puts another set of four little tools in the content creator's toolbox.

Secondly, the way the hyperdrive and jump drive function isn't well-known. The lore literally states that less than a dozen humans understand how the hyperdrive works. And, it can be assumed that that's a similar situation for the Jump Drive as manufactured by the Korath. (Chances are, none of them know how it works, they're just following the instructions.)

Thirdly, if that is exactly how they work, don't you think a more massive ship will be more demanding on the hyperlink? Wouldn't a larger ship demand more from the same-sized JD to produce a larger wormhole?

I'd love to say your opinion is in the minority, but... that's not true, I've gotten plenty of this exact complaint in the past when I suggested it before.

CaptKrevor:
> I'd love to say your opinion is in the minority, but... that's not true, I've gotten plenty of this exact complaint in the past when I suggested it before.<

Maybe because a reaction drive that has to push mass requires leaving virtually the same amount of mass behind to move the mass that needs to be moved?  Then there's that pesky light speed thing that requires even more push the closer you get to it.  Just pointing out that if you want to make things are realistically based on physics as possible, then physics needs to be taken into account as much as possible.

However, IF you want to get into fiction (of the science sort), then get into it.  *I'm* going to go with the theory that hyperdrives and Jumpdrives and other such mechanisms aren't actually pushing anything themselves.  They are achieving a phenomenon that is doing it for us. Thus allowing us to gracefully sidestep physics as we know it today.  *leaves 2 cents behind*

Ferociousfeind:
>  They are achieving a phenomenon that is doing it for us.

See what I already wrote:
> Thirdly, if that is exactly how they work, don't you think a more massive ship will be more demanding on the hyperlink? Wouldn't a larger ship demand more from the same-sized JD to produce a larger wormhole?

CaptKrevor:
I'm talking about a trigger.  The trigger of a .22 caliber pistol is much the same as a trigger of....say....an M60 machine gun.  WHAT they do once their trigger is pulled is considerably different.  I never had to pull the trigger harder just because the weapon was larger.

In this game, I see the hyperlink, scramdrive and jumpdrives simply being triggers for different modes of travel.  All of the drives are of a single size for each of their respective mechanisms.

I could see you pushing for being able to jump FURTHER with more fuel being consumed, but both the hyperlink and scramdrives are dependent upon the existing hyperlinks that already exist between the mapped stars.  THAT is not going to change for the most part.  That's an established (fictionally) functioning system.  There aren't different sized engines that can be installed in different sized ships.  Imagine a cruiser sized engine installed in a clipper. :)  *shrugs*

Ferociousfeind:
I don't see the JDs and HDs as triggers, I see them as entire guns. Takes a bigger gun to throw a bigger bullet further.

MasterOfGrey:
The fact that the scram drive is larger and uses more fuel does imply that there is some kind of energy scaling involved in the effort of initiating whatever science-fiction it is that allows us to sidestep established physics and disappear into superluminal highways. i.e. It's not just a trigger, or scram would represent a bigger more intensive trigger.
However, they're both independent of mass and so is jump drive, and per MZ's lore dump the warp and weft are related in some astrophysical sense even though they are of course very different (in my head I conceive of them as orthogonal dimensions of reality). This implies that, at the very least, the specific equations for how to escape lightspeed physics are much easier to develop/infer than the generalised equations which might let you exploit specific mass cases for reduced fuel use.

Additionally, having options in the creators toolkit is good, if only so that mods and plugins can pursue stranger and more diversely interesting things.

However, in line with my musings, I would be against applying these to any drives in vanilla unless they:
1. Are at least T2.
2. Explain something about an optimal mathematical efficiency that causes mass to cancel out (since hyper and scram are completely independent of mass), but which can be exploited for given cases with the right advanced know-how.

Ferociousfeind:
~~Fuck the lore, HD now consumes more fuel if your ship is heavier, and JD if you're jumping a further distance~~

Ferociousfeind:
I'm contemplating locking the values to >= 0, since negative reference definitely will break things, and negative exponent demolishes the pathfinding algorithm. I only actually ever intended them to be all positive anyways. However, the first implementation I imagined (in ship::BestFuel, if the variable is < 0, set it to the default) seems like it'd violate a lot of style choices. 

Instead, it feels best to set the variable to the default if it's caught _during loading_ less than 0, and drop a notice in the errors.txt file. Would that be the most proper implementation?

Ferociousfeind:
Oh dear, I appear to have broken something again. Allow me to properly trouble-shoot, and properly playtest, and all the important stuff.

Ferociousfeind:
I have diagnosed the problem- I do too much Ruby!! One missing semicolon. Will build, test, and then push to github shortly, all of the people who closely watch #5609

Ferociousfeind:
Additionally, a super quick aside, _this_ is why you cannot trust a compiler to make changes to your code to make it compile! it was complaining not about the line missing a semicolon, but the line three lines down, `error: expected ',' or ';' before 'if'`. Even if most of the time the compiler might be able to point you to the intended code form, it is not perfect and cannot always discern what is wrong, just that it is not right.

data/tooltips.txt
```diff
@@ -157,6 +157,18 @@ tip "disruption resistance heat:"
 
 tip "drag:"
 	`Limits the maximum speed of a ship.`
+	
+tip "drive mass reference:"
+	`The mass of your ship when this drive will function as advertised.`
```
Hurleveur, line 162:
```suggestion
	`If your ship has this mass, this drive will function as advertised.`
```
Could be changed another way, but the previous way seemed to indicate that the mass was being changed by the drive in some way.

source/Ship.cpp
```diff
@@ -3946,12 +3946,39 @@ double Ship::BestFuel(const string &type, const string &subtype, double defaultF
...
+
+		double fuel = baseAttributes.Get("jump fuel");
+
+		double mass = Mass();
+		double driveMassExp = 0;
+		double driveMassRef = 400;
+		double driveDistExp = 0;
+		double driveDistRef = 100;
+
+		if((baseAttributes.Get("drive mass reference") || baseAttributes.Get("drive mass exponent")) && mass > 0)
```
Hurleveur, line 3958:
```suggestion
		if((baseAttributes.Get("drive mass reference") || baseAttributes.Get("drive mass exponent")))
```
Mass cannot be negative.

source/Ship.cpp
```diff
@@ -3946,12 +3946,39 @@ double Ship::BestFuel(const string &type, const string &subtype, double defaultF
...
+		{
+			driveDistRef = max(baseAttributes.Get("drive distance reference"), .01);
+			driveDistExp = max(baseAttributes.Get("drive distance exponent"), 0.);
+			fuel *= pow((jumpDistance / driveDistRef), driveDistExp);
+		}
+
+		// if a "startup" fuel is provided, add that to the above formula.
+		// It's a constant fuel consumption regardless of mass or distance of jump.
+		if(baseAttributes.Get("jump startup fuel") > 0)
+			fuel += baseAttributes.Get("jump startup fuel");
```
Hurleveur, line 3974:
```suggestion
		fuel += baseAttributes.Get("jump startup fuel");
```
I think its best to do it this way, since jump startup fuel cannot be negative. If you want to keep on doing it your way though, at least store it in a variable instead of querying it twice.

source/Ship.cpp
```diff
@@ -3966,6 +3993,34 @@ double Ship::BestFuel(const string &type, const string &subtype, double defaultF
...
+				{
+					driveDistRef = max(it.first->Get("drive distance reference"), .01);
+					driveDistExp = max(it.first->Get("drive distance exponent"), 0.);
+					fuel *= pow((jumpDistance / driveDistRef), driveDistExp);
+				}
+
+				// if a "startup" fuel is provided, add that to the above formula.
+				// It's a constant fuel consumption regardless of mass or distance of jump.
+				if(it.first->Get("jump startup fuel") > 0)
+					fuel += it.first->Get("jump startup fuel");
```
Hurleveur, line 4021: You're using the exact same logic twice, try to place it in a lambda (do contact me on the discord if you need help for that)
I already applied the suggestions from above on it.

The only information you need here is fuel so that's what the lambda should be returning.

It should be like this, and declared before you used it, on line 3949:
```cpp
auto calculateFuelCost = [this, jumpDistance](const Outfit &outfit) -> double
{
	double fuel = outfit.Get("jump fuel");
	// If the jump drive or hyperdrive has an attached
	// (mass/distance) (reference/exponent), calculate
	// the fuel about to be used for this jump.
	double mass = Mass();
	double driveMassExp = 0;
	double driveMassRef = 400;
	double driveDistExp = 0;
	double driveDistRef = 100;

	if((outfit.Get("drive mass exponent") || outfit.Get("drive distance exponent")))
	{
		driveMassRef = max(outfit.Get("drive mass reference"), .01);
		driveMassExp = max(outfit.Get("drive mass exponent"), 0.);
		fuel *= pow((mass / driveMassRef), driveMassExp);
	}
	if((outfit.Get("drive mass reference") || outfit.Get("drive distance reference")) && jumpDistance > 0.)
	{
		driveDistRef = max(outfit.Get("drive distance reference"), .01);
		driveDistExp = max(outfit.Get("drive distance exponent"), 0.);
		fuel *= pow((jumpDistance / driveDistRef), driveDistExp);
	}

	// if a "startup" fuel is provided, add that to the above formula.
	// It's a constant fuel consumption regardless of mass or distance of jump.
	fuel += outfit.Get("jump startup fuel");

	return fuel;
};
```
then you can call it here, using this: `double fuel = calculateFuelCost(*it.first)`

Hurleveur (changes requested):
Nice new changes, may be a bit too complicated for most players but first lets just address this and it'll be easier to change what needs to be changed later.
I'm thinking the `exponent` in the names needs to be changed by something clearer, but idk quite what yet.

Wouldn't it just be better and simpler to have a jump cost per mass option, and only that?
By that I mean something easy to understand for players: cost per 100 mass.
Maybe someone actually wants to use this though, both could be complementary, and if so could be done in different PRs for sure, but I personally see very little use for something so complex.

warp-core (commented):
Haven't looked too closely at this, but the new attribute names and tooltips don't seem too great.
Also, dividing the fuel par into chunks based on how many jumps the ship has does not seem good. This could result in the number of bars halving or doubling going from a hyperspace jump to a jump drive jump, and that's before considering that the rest of this PR supports jump fuel varying based on the distance of the next jump.

Amazinite (commented):
I think this is overengineered and we'd probably be better off with just simply having a hyperdrive/jump drive fuel usage multiplier of some sort that we add to ship hulls. Fuel usage wouldn't strictly be tied to mass due to ships being able to be outfitted differently, but it'd be much more understandable. 

Alternatively, if we still want to keep the increased fuel per mass idea, the attribute could go from being a flat multiplier to a cost per 100 mass. So an interceptor might have a fuel cost/100 tons of say 0.5, meaning that an interceptor with a mass of 200 would have a fuel cost of 1, multiplied by the base hyperdrive's 100 fuel cost to result in 100 fuel usage. But should the interceptor change its outfits around and become say 250 tons, the fuel cost becomes 250/100*0.5*100 = 125 fuel usage. But a medium warship might have a fuel cost/100 tons of 1 and a mass of 1000, meaning its fuel cost becomes 1000/100*1*100 = 1000 fuel per jump; 8 to 10 times more than the interceptor despite only being 4 or 5 times heavier.

I speak of adding the above fuel cost/100 tons attribute to ships, but given how attributes work it could just as easily be made specific to the drive used. So the hyperdrive could be given say 0.5 fuel cost/100 tons with a base cost of 100 while the jump drive could be given 1 fuel cost/100 tons with a base cost of 200.

And if we wanted to mess with cost per distance as well, that'd simply be another attribute, fuel cost/100 map units or something like that.

Whatever form this attribute takes though, I'm still of the opinion that we shouldn't put this to use in vanilla. Higher fuel costs for larger ships would require higher fuel capacities, which breaks the fuel usage weapon economy. Even barring that, I don't see what we would gain from a gameplay perspective by doing this.

Ferociousfeind:
The attributes need better names. And ideally, instead of displaying the values directly (they mean nothing to the player), a graph of mass (or distance) versus fuel usage of the drive (perhaps if there are distance parameters it displays a graph of distance versus fuel use, with mass factored in, and otherwise displays the fuel use of your ship's current mass). I don't know how it'd work with multiple ships selected, that's generally an enigma. Probably `min - max` form, and the graph would be particularly hard to communicate "a range of values" info through.

Ideally, the fuel notches would move down along the fuel gauge as you're engaging hyperspace, to show that one full bar of fuel is being used (down at the bottom) while preserving any partially-filled fuel segment at the top, but I do not think the segmented bar rendering code gives that capability.

If, really, _really,_ only a linear mass -> fuel use relationship is wanted, I could craft a sister PR which does only that- with attribute names like "jump fuel", "jump mass", and "jump penalty" (in fuel units per ton, perhaps normalized to fuel units per 100 tons). It would not be hard to make, given the experience making this one gave me. (But it sure would be demoralizing! 5609, my baby!)

Hurleveur (approved):
Syntax wise this looks good to me.

Ferociousfeind:
There, I've done the deed, now it's only lines.

Amazinite (changes requested):
While looking at this, I've realized that this throws a bit of a wrench in #7378; that PR makes ships cache their fuel costs so that they don't need to be recalculated potentially multiple times a frame as they can be at current. But this PR would make it so that a ship's mass could potentially change its jump cost, invalidating the cached value, meaning I'd need to recalculate a ship's jump cost every time its mass changes. Since a ship's mass includes cargo, that'd mean any time a ship loses cargo (e.g. through dumping it) or gains cargo (e.g. through picking it up), I have to rerun the caching calculation. Is that really something we want to have happen, especially with a feature that I don't see us using in vanilla?

source/Ship.cpp
```diff
@@ -4048,6 +4048,27 @@ double Ship::BestFuel(const string &type, const string &subtype, double defaultF
...
+		double fuel = outfit.Get("jump fuel");
+
+		// If any of these are not defined, they default to 0, so nothing will break.
+		fuel += .01 * outfit.Get("jump mass cost") * (mass - outfit.Get("jump base mass"));
+
+		// It is possible to generate a negative jump fuel, but I
+		// do not think this will mess with things in any major way,
+		// so I will not stop it.
+		return fuel;
+	};
```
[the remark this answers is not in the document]
Ferociousfeind: I would imagine most drives would have 0 mass = some nonzero fuel cost, rather than nonzero mass = zero fuel cost, but if you manage to get to zero fuel cost, I'd say shift the "default the fuel cost" value to -1, not 0, since that is an explicitly disallowed value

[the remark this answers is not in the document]
Amazinite: That would require more changes farther down the line, which I'd rather not dip in to.

source/Ship.cpp
```diff
@@ -4139,6 +4139,22 @@ double Ship::BestFuel(const string &type, const string &subtype, double defaultF
...
+		// Mass cost is the fuel cost per 100 tons of ship mass. The jump base mass of a drive reduces the
+		// ship's effective mass for the jump mass cost calculation. A ship with a mass below the drive's
+		// jump base mass is allowed to have a negative cost relative to the base jump fuel cost.
+		double cost = .01 * outfit.Get("jump mass cost") * (mass - outfit.Get("jump base mass"));
+		cost += outfit.Get("jump fuel");
+
+		// Prevent the cost of jumping from dropping too low.
+		// Put a floor at 1, as a floor of 0 would be forced back up to the default cost, or
+		// be assumed later on to mean you can't jump.
+		return max(1., cost);
```
Amazinite, line 4155: The way I had it is how it should be. If "jump fuel" is 0 and "jump mass cost" is 0 then this should return 0, not 1.

Ferociousfeind:
I've addressed the immediate concerns, though I can't exactly account for a change that isn't in the game yet... (#7378), so I'll have to at least wait for a decision on that. (Or this gets merged first, and that PR has to deal with the fallout)

source/Ship.cpp
```diff
@@ -4139,6 +4139,23 @@ double Ship::BestFuel(const string &type, const string &subtype, double defaultF
...
+
+	auto CalculateFuelCost = [mass, defaultFuel](const Outfit &outfit) -> double
+	{
+		double baseCost = outfit.Get("jump fuel");
+		if(baseCost <= 0.)
+			baseCost = defaultFuel;
+		// Mass cost is the fuel cost per 100 tons of ship mass. The jump base mass of a drive reduces the
+		// ship's effective mass for the jump mass cost calculation. A ship with a mass below the drive's
+		// jump base mass is allowed to have a negative mass cost.
+		double massCost = .01 * outfit.Get("jump mass cost") * (mass - outfit.Get("jump base mass");
```
Amazinite, line 4152:
```suggestion
		double massCost = .01 * outfit.Get("jump mass cost") * (mass - outfit.Get("jump base mass"));
```

Amazinite (approved):

Issue #5582, opened 2020-12-29
Don't reset multiple keybinds that are overloaded to a single key


Amazinite:
**Is your feature request related to a problem? Please describe.**
It is currently possible to overload multiple keybinds to the same key. For the duration of your play session, pressing the overloaded key will cause both commands to be sent. But once the game is closed, the game will reset one of the keybinds to its default to avoid having two keys be overloaded. This is unfortunate, as it can be easier to have multiple commands bound to the same key, such as having both primary and secondary fire on the same key.

**Related Issue Links**
#3972 Having overloading be allowable behavior could make it easier to find setups for other input methods.

**Describe the solution you'd like**
Do not reset keybinds to their default if they are overloaded. We should keep the current UI behavior though where overloaded keybinds get highlighted, although perhaps using yellow instead of red to show that it is a warning, not an error.
![image](https://user-images.githubusercontent.com/17688683/103253602-b6457900-4936-11eb-82aa-921f85a9bd82.png)

**Describe alternatives you've considered**
Nothing.

**Additional context**
N/A

yobbo2020:
This also annoyed me so i made a PR. Although i didn't change the colour of the warning, it at least saves and loads however you set it.

PR #5611, opened 2021-01-05
feat(settings): Allow conflicting keybindings to be saved (#5582).

yobbo2020:
**Feature:** This PR implements the feature request detailed and discussed in issue #5582 

## Feature Details
Keybindings that conflict are saved as is, in stead of only saving one of the conflicting keybindings.

Previously key bindings were saved per-key. I changed it to save per-command. Save file format remains unchanged.

## UI Screenshots
No change. Conflicts are still marked in red.

## Usage Examples
No change to existing usage.

## Testing Done
My preferred setup with both cloaking and afterburner bound to the same key now saves and loads correctly.

## Performance Impact
none

### DIFF
--- a/source/Command.cpp
+++ b/source/Command.cpp
@@ -156,11 +156,11 @@ void Command::SaveSettings(const string &path)
 {
 	DataWriter out(path);
 	
-	for(const auto &it : commandForKeycode)
+	for(const auto &it : keycodeForCommand)
 	{
-		auto dit = description.find(it.second);
+		auto dit = description.find(it.first);
 		if(dit != description.end())
-			out.Write(dit->second, it.first);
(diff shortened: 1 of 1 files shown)


petervdmeer (approved):
I had a short look at the code; this seems to be matching with the way LoadSettings retrieves the keybindings through keycodeForCommand and with the way SetKey also uses keycodeForCommand.

Looks good to me.

Amazinite (changes requested):
There's a comment further up in this file that is no longer valid. Lines 128-129.
```
 	// Each command can only have one keycode, but misconfigured settings can
	// temporarily cause one keycode to be used for two commands.
```

Amazinite:
Superseded by #5683.

PR #5617, opened 2021-01-08
Post-Wanderer Pond Strider compatibility

FixItYondu:

## Summary
I noticed that the Pond Strider mission chain cannot be started if one has moved past that stage in the Wanderer campaign. This means that anyone with an older save which has moved past that stage in Wanderer campaign cannot unlock the Pond Strider or the Flea. This seemed like something of an oversight to me, so I created a couple of missions which would serve the same purpose as the Wanderer missions which triggered the chain. 

## Save File
[Jimmy Jongo~Pond Strider Test.txt](https://github.com/endless-sky/endless-sky/files/5784765/Jimmy.Jongo.Pond.Strider.Test.txt)
This save file should be adequate to play through the changes since there is no combat required in the mission chain. It has the Wanderer campaign done so it is representative of the saves this PR aims to fix.

### DIFF
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -21,28 +21,28 @@ mission "First Contact: Hai"
 				`	(Sure.)`
 				`	(Not right now.)`
 					defer
-			
+
 			`	You walk up to a human merchant who is busy haggling with one of the aliens. They interrupt their conversation when they see you approaching, and the human gestures for you to join them. "You look totally lost," he says. "First time down the rabbit hole?"`
 			`	You nod. The alien reaches out a paw to shake hands with you. "Welcome to Hai space," it says. "We are people of peace, and you are free to visit us or even live among us. Our worlds provide enough bounty for all."`
 				goto next
-			
+
 			label unfettered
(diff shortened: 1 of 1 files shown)


Amazinite:
This was completely intentional.

FixItYondu:
I see. Is there any reason why this is so? It seems a bit unfair to those who wish to have access to the ship on an older save.

Amazinite:
The way I was writing the story, it didn't make much sense to me to have a mission just randomly offer from this section of the story if you've already gotten past it. That isn't to say that an alternative intro is out of the question. Aside from the formatting issues here though, I'd take a different approach to setting this up to remove the need for some much duplicated text.

Terin:
The developments in the Wanderers story don't really change the world enough to put this off the table. The Unfettered gain more worlds, sure, but running into the wall of the Arfectas would likely still cause them to lash out back into Hai space proper with the Solifuges, for the foreseeable future. I don't think we've really moved past that part of the story, it feels just like another extension of new content coming out.

Amazinite:
Handled this with https://github.com/endless-sky/endless-sky/commit/36c86927b2b234fa550b78c729002348b0315759, which avoids having duplicated conversation text.

PR #5620, opened 2021-01-08
feat/test: Testframework steps apply, assert, branch, inject, label and watchdog

petervdmeer:
**Feature:** This PR implements part of the test-framework as discussed in #4308

## Feature Details
This PR contains the implementation for test-steps apply, assert, branch, inject, label and watchdog.
This PR also contains additional error-printing in case a test fails.

## UI Screenshots
```
TAP version 13
# ***********************************************
# ***         ES Autotest-runner              ***
# ***********************************************
# Using Test Anything Protocol for reporting test-results
1..3
ok 1 "Test-Framework - Conditions Arithmetic and Loops"
ok 2 "Test-Framework - Empty Test Sequence"
ok 3 "Test-Framework - Empty Testcase"
```

## Usage Examples
See the test "Test-Framework - Conditions Arithmetic and Loops" in data/tests/tests_framework.txt for usage of the new keywords.

## Testing Done
Most new keywords were automatically tested by the framework, `inject` was tested in a previous version of the framework and `watchdog` was tested manually (by decreasing the timeout in an existing test).

## Performance Impact
Overhead from the test-framework is expected to be minimal, except for tests that have really large loops inside them. I would expect such tests to be caught during review.


source/Test.cpp
```diff
@@ -87,21 +87,111 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
...
+				{
+					step.nameOrLabel = child.Token(1);
+					if(jumpTable.find(step.nameOrLabel) != jumpTable.end())
+					{
+						child.PrintTrace("Error: duplicate label");
+						status = Status::BROKEN;
+						return;
+					}
+					else
+						jumpTable[step.nameOrLabel] = steps.size()-1;
```
tehhowch, line 161:
```suggestion
						jumpTable[step.nameOrLabel] = steps.size() - 1;
```

source/Test.cpp
```diff
@@ -87,21 +87,111 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
...
+				break;
+			case TestStep::Type::NAVIGATE:
+				child.PrintTrace("Error: Not yet implemented step type navigation");
+				status = Status::BROKEN;
+				return;
+			case TestStep::Type::WATCHDOG:
+				if(child.Size() < 2)
+					step.watchdog = 0;
+				else
+					step.watchdog = child.Value(1);
```
tehhowch, line 172: For most of the game code, we do DataNode size checks in the "meets criteria" phrasing first.
```suggestion
				step.watchdog = child.Size() >= 2 ? child.Token(1) : 0;
```

source/Test.cpp
```diff
@@ -192,12 +281,83 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+				{
+					// Lookup the data and inject it in the game or into the environment.
+					const TestData* testData = (GameData::TestDataSets()).Get(stepToRun.nameOrLabel);
+					if(!testData->Inject())
+						Fail(context, player, "injecting data failed");
+				}
+				++(context.stepToRun);
+				break;
+			case TestStep::Type::INPUT:
+				if(context.unprocessedInput){
```
tehhowch, line 334:
```suggestion
				if(context.unprocessedInput)
				{
```

source/Test.cpp
```diff
@@ -210,10 +370,36 @@ const string &Test::StatusText() const
...
+	else
+		testFailMessage += ".";
+	
+	// Add generic logging of the failed test.
+	Files::LogError(testFailMessage);
+	// Log variables at the moment of failure.
+	const map<string, int64_t> &conditions = player.Conditions();
+	if(!conditions.empty())
+		for(const auto &it : conditions)
+			Files::LogError("Condition: \"" + it.first + "\" = " + to_string(it.second));
```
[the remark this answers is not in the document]
petervdmeer: I like your idea of printing only the conditions that this test cares about, but that will be quite some work to implement. What I did for now was to only print the conditions that start with `test: `. This will also limit the amount of conditions that will be printed and gives us room to implement your idea in a later PR.

Just printing the first N conditions is very annoying if the condition somebody wants to see is at N+1.

I also changed the code to roll-up to a huge string and then have a single call for the conditions.

petervdmeer:
I did some additional clean-up of unused code and consider this PR now ready for review/merge.

source/Test.cpp
```diff
@@ -192,12 +274,81 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+	context.unprocessedInput = false;
+	context.branchesSinceGameStep.clear();
 	
-	// Exit with error on a failing testStep.
-	const string &stepTypeName = STEPTYPE_TO_TEXT.at(stepToRun.stepType);
-	
-	Fail("Test step " + to_string(context.stepToRun) + " (" + stepTypeName + ") failed");
+	bool continueGameLoop = false;
+	while(context.stepToRun < steps.size() && !continueGameLoop)
+	{
```
tehhowch, line 283: This should be a do-while construct (since we just bounds-checked stepToRun against steps.size() before entering it).

source/Test.cpp
```diff
@@ -192,12 +274,81 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+		const TestStep &stepToRun = steps[context.stepToRun];
+		switch(stepToRun.stepType)
+		{
+			case TestStep::Type::APPLY:
+				stepToRun.conditions.Apply(player.Conditions());
+				++(context.stepToRun);
+				break;
+			case TestStep::Type::ASSERT:
+				if(!stepToRun.conditions.Test(player.Conditions()))
+					Fail(context, player, "asserted false");
```
tehhowch, line 299: Should we pass the conditionset that failed rather than a failure message in this case? This would allow us to print it and only the associated player conditions.

(We can defer this to a follow-up if desired)

source/Test.cpp
```diff
@@ -192,12 +274,81 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+				++(context.stepToRun);
+				break;
+			case TestStep::Type::BRANCH:
+				// If we encounter a branch entry twice, then resume the gameloop before the second encounter.
+				if(context.branchesSinceGameStep.count(context.stepToRun))
+				{
```
tehhowch, line 305: Whats the reasoning for this?

source/Test.cpp
```diff
@@ -192,12 +274,81 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+			case TestStep::Type::INPUT:
+				if(context.unprocessedInput)
+				{
+					continueGameLoop = true;
+					break;
+				}
+				// Give the relevant inputs here.
+				Fail(context, player, "Input not implemented");
+				// Make sure that we run a gameloop before the next input.
+				context.unprocessedInput = true;
```
tehhowch, line 335: Would it be sensible to set the "run the game loop" flag here, rather than waiting for the next input or other similar condition?

source/Test.cpp
petervdmeer: Let's defer this to follow-up. I'm okay with the idea of only printing the conditions used in the (whole) test, but only printing the conditions from a single step could provide an incomplete picture (if variables around the step are relevant).

source/Test.cpp
petervdmeer: Yes, let's process the input immediately after it was given.

source/Test.cpp
petervdmeer: This deals with loops that wait for some condition to become true. If we don't regularly return to the gameloop from such "wait loops", then the game would never have the chance to proceed with actions that can make the condition true.

petervdmeer: I added this explanation as a comment in the code.

petervdmeer:
All review-comments appear to be processed.

PR #5624, opened 2021-01-10
make zoom work with numpad keys

tatami4:
Should fix #4944.

PR #5627, opened 2021-01-11
make numpad + and - work in trading and hiring panels

tatami4:
Haven't found any issues mentioning this.

### DIFF
--- a/source/HiringPanel.cpp
+++ b/source/HiringPanel.cpp
@@ -107,7 +107,7 @@ bool HiringPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, b
 	if(!player.Flagship())
 		return false;
 	
-	if(key == 'h' || key == SDLK_EQUALS || key == SDLK_KP_PLUS || key == SDLK_RETURN || key == SDLK_SPACE)
+	if(key == 'h' || key == SDLK_EQUALS || key == SDLK_KP_PLUS || key == SDLK_PLUS || key == SDLK_RETURN || key == SDLK_SPACE)
 	{
 		player.Flagship()->AddCrew(min(maxHire, Modifier()));
 		player.UpdateCargoCapacities();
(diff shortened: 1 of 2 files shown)


MCOfficer (approved):
I checked and these are in fact the last occurences of this pattern.

```
$ rg "key == '[-=+]'"
source\TradingPanel.cpp
226:    else if(key == '=' || key == SDLK_RETURN || key == SDLK_SPACE)
228:    else if(key == '-' || key == SDLK_BACKSPACE || key == SDLK_DELETE)
source\HiringPanel.cpp
110:    if(key == 'h' || key == '=' || key == SDLK_RETURN || key == SDLK_SPACE)
115:    else if(key == 'f' || key == '-' || key == SDLK_BACKSPACE || key == SDLK_DELETE) 
```

We could arguably this about replacing `key == '\t'` as well, but i suppose that's out of scope for now.

tatami4:
Why would we need to replace `\t`?

source/HiringPanel.cpp
```diff
@@ -107,12 +107,12 @@ bool HiringPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, b
 	if(!player.Flagship())
 		return false;
 	
-	if(key == 'h' || key == '=' || key == SDLK_RETURN || key == SDLK_SPACE)
+	if(key == 'h' || key == SDLK_EQUALS || key == SDLK_KP_PLUS || key == SDLK_RETURN || key == SDLK_SPACE)
```
Amazinite, line 110: #5624 replaced `key == '='` with `key == SDLK_PLUS || key == SDLK_KP_PLUS || key == SDLK_EQUALS`. Here you've dropped `key == SDLK_PLUS`. Shouldn't that be used here as well? (And in the trading panel instance.)

tatami4: I'm not sure. I don't think that the game registers the SDLK_PLUS key, because it should only work with shift + '='.

MCOfficer:
> 
> 
> Why would we need to replace `\t` (Tab)?

Maybe there are some exotic layouts where the tab key is located elsewhere, like the `-` or `=`.

PR #5629, opened 2021-01-12
Enough fuel renders as just one solid bar of fuel

Ferociousfeind:
**Feature:** The fuel bar will not break into a million pieces if you have a very large amount of fuel

## Feature Details
If you have more than 1200 fuel (12 bars) on your ship, the fuel bar renders as one instead of 13 or 600 bars, or whatever crazy amount you may have.

## UI Screenshots
If you've ever flown a ship with only 100 fuel, you know what the bar will look like.

## Testing Done
Y'all are magicians testing these builds without some sort of game launcher that doesn't require you to compile the game by hand. Will heckin' test on ships with low fuel, and very high fuel, and fuel values not divisible by 100, after I make the pull request and ESL2 can grab this PR number. You know how it be.

## Performance Impact
Adds one check to a process I believe should occur once per frame, but I don't think checking the inequality of a variable to a constant is very intensive, this shouldn't impact anything.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -633,8 +633,10 @@ void Engine::Step(bool isActive)
 		info.SetString("location", currentSystem->Name());
 	info.SetString("date", player.GetDate().ToString());
 	if(flagship)
-	{ //if there is a lot of fuel, display a solid homogenous bar instead of many notches in the bar
-		if ( flagship->Attributes().Get("fuel capacity") <= 1200)
+	{
+		const double fuelCap = 2000.;
+		// if there is a lot of fuel, display a solid homogenous bar instead of many notches in the bar
+		if ( flagship->Attributes().Get("fuel capacity") <= fuelCap)
 		{
 			info.SetBar("fuel", flagship->Fuel(),
 				flagship->Attributes().Get("fuel capacity") * .01);
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -635,11 +635,12 @@ void Engine::Step(bool isActive)
 	if(flagship)
 	{
 		const double fuelCap = 2000.;
-		// if there is a lot of fuel, display a solid homogenous bar instead of many notches in the bar
-		if ( flagship->Attributes().Get("fuel capacity") <= fuelCap)
+		// If there is a lot of fuel, display a solid homogenous bar,
+		// instead of many notches in the bar.
+		if(flagship->Attributes().Get("fuel capacity") <= fuelCap)
 		{
 			info.SetBar("fuel", flagship->Fuel(),
-				flagship->Attributes().Get("fuel capacity") * .01);
+				    flagship->Attributes().Get("fuel capacity") * .01);
 		} else {
 			info.SetBar("fuel", flagship->Fuel());
 		}
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -633,9 +633,14 @@ void Engine::Step(bool isActive)
 		info.SetString("location", currentSystem->Name());
 	info.SetString("date", player.GetDate().ToString());
 	if(flagship)
-	{
-		info.SetBar("fuel", flagship->Fuel(),
-			flagship->Attributes().Get("fuel capacity") * .01);
+	{ //if there is a lot of fuel, display a solid homogenous bar instead of many notches in the bar
+		if ( flagship->Attributes().Get("fuel capacity") <= 1200)
+		{
+			info.SetBar("fuel", flagship->Fuel(),
+				flagship->Attributes().Get("fuel capacity") * .01);
+		} else {
+			info.SetBar("fuel", flagship->Fuel());
+		}
 		info.SetBar("energy", flagship->Energy());
 		double heat = flagship->Heat();
 		info.SetBar("heat", min(1., heat));
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -634,16 +634,14 @@ void Engine::Step(bool isActive)
 	info.SetString("date", player.GetDate().ToString());
 	if(flagship)
 	{
-		const double fuelCap = 2000.;
+		const double MAX_FUEL_DISPLAY = 5000.;
+		double fuelCap = flagship->Attributes().Get("fuel capacity");
 		// If there is a lot of fuel, display a solid homogenous bar,
 		// instead of many notches in the bar.
-		if(flagship->Attributes().Get("fuel capacity") <= fuelCap)
-		{
-			info.SetBar("fuel", flagship->Fuel(),
-				    flagship->Attributes().Get("fuel capacity") * .01);
-		} else {
+		if(fuelCap <= MAX_FUEL_DISPLAY )
+			info.SetBar("fuel", flagship->Fuel(), fuelCap* .01);
+		else
 			info.SetBar("fuel", flagship->Fuel());
-		}
 		info.SetBar("energy", flagship->Energy());
 		double heat = flagship->Heat();
 		info.SetBar("heat", min(1., heat));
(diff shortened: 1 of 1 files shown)


source/Engine.cpp
```diff
@@ -633,9 +633,14 @@ void Engine::Step(bool isActive)
 		info.SetString("location", currentSystem->Name());
 	info.SetString("date", player.GetDate().ToString());
 	if(flagship)
-	{
-		info.SetBar("fuel", flagship->Fuel(),
-			flagship->Attributes().Get("fuel capacity") * .01);
+	{ //if there is a lot of fuel, display a solid homogenous bar instead of many notches in the bar
+		if ( flagship->Attributes().Get("fuel capacity") <= 1200)
```
tatami4, line 637: Maybe declare this as a constant at the top of the file. I think code should contain as little magic numbers as possible.

Ferociousfeind: Probably true

tehhowch:
For the record, opening PRs against this repo just so you can obtain a version of the game compiled for you is unnecessary -- open a PR on your own fork and download the zip file from the CD workflow. Doing otherwise is rude and abuses the project's allocated resources for your own personal benefit.

In the future, workflow runs & artifacts associated with PRs that are not intended for merging **will be deleted.**

tehhowch (changes requested):
Style violations

How did you come up with 1200 fuel as the limit for rendering delimiters? 1300 capacity is still very readable: 
image

I would say 2000 fuel is probably where we should stop adding measurement lines to the bar, but that's still just an arbitrary number.
image

Ferociousfeind:
I'm open to changing the magic number. Before nudging it up to 1200, it was at 1000, it can go higher if people can still read it at those values. This is probably one of those numbers that would go into that gameConstants configuration file I've heard speak of.

(And, I don't open PRs not meant to be merged into the main game. If it's exclusively for me and other people aren't interested I just don't go through with doing code changes. But I suppose wrangling with my own personal version is a touch more appropriate than drafting a PR, testing with ESL2, and then marking it for actual merging once I'm super confident...)

Amazinite:
1000 | 2000 | 3000 | 4000 | 4500 | 5000 | 5500 | 6000 | 6500 | 7000 | 7500 | 8000 | 8500 | 9000 | 10000 | 20000 | 30000
----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|----|
 |  |  |  |  |  |  |  |  |  |  |  |  |  |  |  | 

The strange gradient and having the fuel bar basically completely disappear is likely a function of your vertical resolution. These screenshots were taken in windowed mode on a 1920x1080 monitor with 1000 pixels of vertical space in the window. Stuff doesn't really start to look too off until 5000 when you start to get a gradient caused by the bars not all being the same height. I'd consider that still readable, though. It's not until 6500 where readability goes out the window. So if we're going to have any magic numbers, I'd say either make the bar solid if the fuel is greater than 6000, as at least at that point the bar is still readable, if a bit small.

gunqqer:
Given this is resolution dependent, would it be possible to base the magic number on the resolution? Although that may be a bit more work than is really worth it.

Amazinite:
Yeah, I was about to say that you could try and use the resolution to determine the cutoff point if you want to get fancy with it. Here's that same 6500 fuel that was unreadable before, only this time I'm at 150% main zoom factor, the highest my game will go. (So it isn't vertical space, but zoom factor that we want to use to determine the cutoff.) I'd consider this readable at 150% zoom.
image

Could people with fancier monitors find the point where the fuel bar becomes unreadable?

Ferociousfeind:
I feel like calling for the screen's current resolution is a little bit outside my skillset as of now... (changing how much fuel a drive consumes was disturbingly easy to me and I still have anxiety I've screwed multiple things up in doing it) but adjusting an unchanging constant that is now already there is within my reach!

By "unreadable" I meant moreso "too many divisions to easily, instinctually tell the difference between". You can immediately tell if your new ship has 200, 300, 400, 500, or 600 fuel without even glancing towards the yellow bar, that's well within innate recognizability. While I get the point that at some point huge amounts of fuel will have weird effects on the bar, I want to stop it from dividing long before it gets to that point, on any monitor. I figured, originally, no ship has, stock, more than 1200 fuel, and beyond that point you're not splitting hairs over a single unit of 100 fuel. I bumped that magic number up to 2000 because yeah, 1200 fuel you can easily see, and knowing that amount is fine. And if it feels right and works, I could bump the maximum up to 4000 too.

How small are monitors going to get, realistically? At what amount of fuel does the fuel bar become uselessly split into pieces? I'd like to put the maximum rendered amount at 75-80% that value.

Amazinite:
> You can immediately tell if your new ship has 200, 300, 400, 500, or 600 fuel without even glancing towards the yellow bar...
> I figured, originally, no ship has, stock, more than 1200 fuel, and beyond that point you're not splitting hairs over a single unit of 100 fuel...

Bump it up to where I suggested. You're thinking about glancing at the fuel bar when it's full or nearly so. But I'm looking at readability in terms of "Can I make out how many jumps I have left when I'm low." I don't care how much fuel I have left if I can hold 1200 and at a glance see that I have 1100. What I care about is if I can glance at my fuel gauge and see if I should land soon, and I could glance at that 6000 fuel bar (even if it's absurd to have that much and ever need to jump that much in one go) and say "Oh, I only have 3/4/5/6 jumps left."

source/Engine.cpp
```diff
@@ -634,8 +634,16 @@ void Engine::Step(bool isActive)
...
+		const double fuelCap = 2000.;
+		// If there is a lot of fuel, display a solid homogenous bar,
+		// instead of many notches in the bar.
+		if(flagship->Attributes().Get("fuel capacity") <= fuelCap)
+		{
+			info.SetBar("fuel", flagship->Fuel(),
+				    flagship->Attributes().Get("fuel capacity") * .01);
+		} else {
+			info.SetBar("fuel", flagship->Fuel());
+		}
```
Amazinite, line 646:
```suggestion
		const double MAX_FUEL_DISPLAY = 5000.;
		double fuelCap = flagship->Attributes().Get("fuel capacity");
		// If there is a lot of fuel, display a solid homogenous bar,
		// instead of many notches in the bar.
		if(fuelCap <= MAX_FUEL_DISPLAY )
			info.SetBar("fuel", flagship->Fuel(), fuelCap* .01);
		else
			info.SetBar("fuel", flagship->Fuel());
```
The MAX_FUEL_DISPLAY variable should then be moved to below RADAR_SCALE at the top of the file.

Amazinite (changes requested):

source/Engine.cpp
```diff
@@ -634,8 +635,13 @@ void Engine::Step(bool isActive)
 	info.SetString("date", player.GetDate().ToString());
 	if(flagship)
 	{
-		info.SetBar("fuel", flagship->Fuel(),
-			flagship->Attributes().Get("fuel capacity") * .01);
+		double fuelCap = flagship->Attributes().Get("fuel capacity");
+		// If there is a lot of fuel, display a solid homogenous bar,
+		// instead of many notches in the bar.
+		if(fuelCap <= MAX_FUEL_DISPLAY )
```
Amazinite, line 641:
```suggestion
		if(fuelCap <= MAX_FUEL_DISPLAY)
```

source/Engine.cpp
```diff
@@ -634,8 +635,13 @@ void Engine::Step(bool isActive)
 	info.SetString("date", player.GetDate().ToString());
 	if(flagship)
 	{
-		info.SetBar("fuel", flagship->Fuel(),
-			flagship->Attributes().Get("fuel capacity") * .01);
+		double fuelCap = flagship->Attributes().Get("fuel capacity");
+		// If there is a lot of fuel, display a solid homogenous bar,
+		// instead of many notches in the bar.
+		if(fuelCap <= MAX_FUEL_DISPLAY)
+			info.SetBar("fuel", flagship->Fuel(), fuelCap* .01);
```
Amazinite, line 642:
```suggestion
		// If the flagship has a large amount of fuel, display a solid bar.
		// Otherwise, display a segment for every 100 fuel.
		if(fuelCap <= MAX_FUEL_DISPLAY)
			info.SetBar("fuel", flagship->Fuel(), fuelCap * .01);
```

PR #5632, opened 2021-01-12
feat: Testframework: Print OpenGL, graphical environment variables and running services

petervdmeer:
**Feature:** This PR implements additional graphical debugging info relevant for the discussion at [Discord](https://discord.com/channels/251118043411775489/285540320823869441/798236048844390400)

## Feature Details
This PR adds printing of OpenGL-info, DISPLAY and XDG variables and running known X-services.

## UI Screenshots
N/A

## Usage Examples
This is part of the CI test-suite. Check the logs and see the output when required.

## Testing Done
This is part of the CI-system, so will be tested as part of the CI-runs.

## Performance Impact
The printing itself is likely very fast, but the output could be a bit too verbose (in which case we should follow-up with only printing the information when we detect a display issue in the test).

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -2,10 +2,14 @@
 
 # Helper function to print debugging data for failures on graphical environment.
 function print_graphics_data () {
+#	echo "# ***********************************************"
+#	echo "# OpenGL versions & available extensions:"
+#	echo "# ***********************************************"
+#	echo "$(glxinfo | grep -E "OpenGL|GL_" | sed "s/^/# /")"
 	echo "# ***********************************************"
-	echo "# OpenGL versions & available extensions:"
+	echo "# OpenGL version:"
 	echo "# ***********************************************"
-	echo "$(glxinfo | grep -E "OpenGL|GL_" | sed "s/^/# /")"
+	echo "$(glxinfo | grep -E "OpenGL" | sed "s/^/# /")"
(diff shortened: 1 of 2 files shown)


MCOfficer:
Thank you <3

You could actually keep pushing empty commit until the integration test fails :D

tehhowch:
Pushing empty commits might not trigger much in the way of CI jobs, but testing race conditions by rerunning a number of PRs' checks is certainly something we can do, both here or in a fork.

petervdmeer:
My second commit has shortened the amount of logging a bit. I was hoping that we could merge this PR and then also have this logging coverage for CI-tests that run "in the wild".

You could still run the tests multiple times in a fork or in a PR (but could do that just with master).

tests/run_tests.sh
```diff
@@ -1,5 +1,27 @@
 #!/bin/bash
 
+# Helper function to print debugging data for failures on graphical environment.
+function print_graphics_data () {
+#	echo "# ***********************************************"
```
tehhowch, line 5: Let's include the requirement for the env var to print this stuff - the CI passes it, and then it's opt-in for users running this manually.
```suggestion
function print_graphics_data () {
	if [[ -z ${PRINT_GLXINFO} ]]; then
		return;
	fi
#	echo "# ***********************************************"
```

Issue #5637, opened 2021-01-15
Flagship on load screen can mismatch with active flagship


Terin:
**Describe the bug**
The top ship in the player's fleet will always be the ship shown on the load panel, even if your active flagship is lower on the list.

**To Reproduce**
1. Have a fleet of multiple ships
2. Have another ship hold position outside of the system and land, or land and park another ship
3. Move said ship above your active flagship
4. Observe the active ship change when clicking on "Load/Save..." or "Back to Menu"

**Expected behavior**
The displayed ship will be your active flagship on both panels.

**Screenshots**
![image](https://user-images.githubusercontent.com/4471575/104705466-8db3d200-56df-11eb-85c0-effad816d9ef.png)

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Self-compiled
 - Version: 6de8d68

**Link to save file**
[Test Pilot.txt](https://github.com/endless-sky/endless-sky/files/5819464/Test.Pilot.txt)

PR #5638, opened 2021-01-15
Fix: Display only active flagship on load panel

Terin:
**Bugfix:** This PR addresses issue #5637

## Fix Details
In the SavedGame class, the displayed ship is the first ship grabbed. This fix causes it to clear the entry if it finds that ship is parked, or not in the same system as the player, which will line up with the proper flagship.

## Testing Done
Tested that the flagship matches when the top ship is in another system, and if it is parked.

## Save File
This save file can be used to verify the bugfix. The bug will occur when using 6de8d68, and will not occur when using this branch's build.
[Test Pilot.txt](https://github.com/endless-sky/endless-sky/files/5819464/Test.Pilot.txt)


source/SavedGame.cpp
```diff
@@ -65,6 +65,18 @@ void SavedGame::Load(const string &path)
 					shipName = child.Token(1);
 				else if(child.Token(0) == "sprite" && child.Size() >= 2)
 					shipSprite = SpriteSet::Get(child.Token(1));
+				else if(child.Token(0) == "system" && child.Size() >= 2)
+				{
+					if(child.Token(1) != system) {
+						shipSprite = NULL;
```
Amazinite, line 71: We use “nullptr” around these parts.

source/SavedGame.cpp
```diff
@@ -65,6 +65,21 @@ void SavedGame::Load(const string &path)
 					shipName = child.Token(1);
 				else if(child.Token(0) == "sprite" && child.Size() >= 2)
 					shipSprite = SpriteSet::Get(child.Token(1));
+				// If the selected ship is parked or in another system, it cannot be the active
+				// flagship, so clear the sprite and continue the loop until it is found
```
Amazinite, line 69:
```suggestion
				// flagship, so clear the sprite and continue the loop until it is found.
```
Could the first ship in the list potentially be in system but disabled, therefore meaning it can't be the flagship?

Terin: Ah, good catch. Since all the code to check if a ship is disabled is private in the Ship class, I'm adding a tag for Ship::Write() which'll write out `disabled`. This won't get read when loading a full save, since the game will detect that from the full attributes, but it'll quickly let it find out when doing the quick read.

Terin:
Per suggestion from tehhowch, updated the code to directly detect which ship in the list is the flagship, and write that to the broader save file, rather than writing new attributes to ships.

source/SavedGame.cpp
```diff
@@ -36,6 +36,8 @@ void SavedGame::Load(const string &path)
 	if(file.begin() != file.end())
 		this->path = path;
 	
+	int flagshipIterator = 1;
+	int flagshipTarget = 1;
```
Amazinite, line 40: If the player has ships but no flagship (can happen if you sold your flagship and there are no other in-system ships, or all in-system ships are automata when you save) then this means the first ship is used as the flagship. Perhaps the flagshipTarget should be initialized as 0 or -1.

tehhowch (changes requested):

source/SavedGame.cpp
Terin: Without a default value, any saves pre-patch will appear with no ship. Keeping that default so that pre-patch saves could (possibly incorrectly) show a ship would probably be better than nothing at all. Then we'll set `flagship -1` when writing out the save with no detected flagship. That seems like the solution that causes the least issues.

source/SavedGame.cpp
```diff
@@ -57,8 +59,15 @@ void SavedGame::Load(const string &path)
 					break;
 				}
 		}
+		else if(node.Token(0) == "flagship" && node.Size() >= 2)
+			flagshipTarget = node.Value(1);
 		else if(node.Token(0) == "ship" && !shipSprite)
 		{
+			if(flagshipIterator != flagshipTarget)
+			{
+				flagshipIterator++;
```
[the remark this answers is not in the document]
Terin: Sounds good with the postfix. However, what do you mean about splitting the usage and increment? Line 66 is the "usage", and line 68 is the increment. Unless you mean splitting them even further than that?

Terin:
Any further action to take on this? Tested with the latest commit merged, which fixes the issue, while the main branch still shows it happening.

source/SavedGame.cpp
```diff
@@ -48,6 +50,8 @@ void SavedGame::Load(const string &path)
 			planet = node.Token(1);
 		else if(node.Token(0) == "playtime" && node.Size() >= 2)
 			playTime = Format::PlayTime(node.Value(1));
+		else if(node.Token(0) == "flagship index" && node.Size() >= 2)
+			flagshipTarget = node.Value(1);
```
quyykk, line 54: Is this the first "dependent" node? As in, it needs to appear before any `"ship"` node or else it doesn't work?

Terin: It'll need to appear up top, yeah. However, if a save file does not have that node (like with an old save file), it'll default to the original behavior.

source/SavedGame.cpp
```diff
@@ -57,7 +61,7 @@ void SavedGame::Load(const string &path)
 					break;
 				}
 		}
-		else if(node.Token(0) == "ship" && !shipSprite)
+		else if(node.Token(0) == "ship" && !shipSprite && ++flagshipIterator == flagshipTarget)
```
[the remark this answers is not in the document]
Terin: If anything, that's more accurate. If a player somehow has a ship with no sprite as their flagship, they are certainly breaking the game, but it doesn't stop being their flagship.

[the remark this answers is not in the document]
quyykk: What if there is no "flagship" index? Then the first ship is used but it might not have a sprite => whereas on master it would use the second ship.

But anyways, then you should remove the code related to the sprite thingy if it isn't used anymore. Cause right now it's dead code.

source/PlayerInfo.cpp
```diff
@@ -2794,7 +2794,19 @@ void PlayerInfo::Save(const string &path) const
...
 		out.Write("travel destination", travelDestination->TrueName());
-	
+	// Detect which ship number is the current flagship, for showing on LoadPanel
+	if(flagship)
+	{
+		for(auto it = ships.begin(); it != ships.end(); ++it)
+		{
+			if(*it == flagship)
+				out.Write("flagship index", distance(ships.begin(), it));
+		}
```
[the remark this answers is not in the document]
Terin: Based on what tehhowch said earlier, and looking into it myself, it seems like `std:find` returns a matching element, but not the index, which is needed to find the correct flagship. Is there a way to make it return one?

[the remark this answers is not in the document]
quyykk: `std::find` returns an iterator. So you can do

```cpp
auto it = find(ships.begin(), ships.end(), flagship);
if(it != ships.end())
    out.Write("flagship index", it - ships.begin());
```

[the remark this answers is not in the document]
tehhowch: Can also write the UUID as e.g. `"flagship ID"`.

source/PlayerInfo.cpp
```diff
@@ -2794,6 +2794,19 @@ void PlayerInfo::Save(const string &path) const
...
+	}
+	else
+	{
+		out.Write("flagship index", -1);
+	}
```
Amazinite, line 2809: Remove unnecessary braces. Also, break out of the loop once we have the flagship index
```suggestion
	if(flagship)
	{
		for(auto it = ships.begin(); it != ships.end(); ++it)
			if(*it == flagship)
			{
				out.Write("flagship index", distance(ships.begin(), it));
				break;
			}
	}
	else
		out.Write("flagship index", -1);
```

Amazinite (changes requested):

source/PlayerInfo.cpp
```diff
@@ -2860,6 +2860,18 @@ void PlayerInfo::Save(const string &path) const
 		out.Write("travel", system->Name());
 	if(travelDestination)
 		out.Write("travel destination", travelDestination->TrueName());
+	// Detect which ship number is the current flagship, for showing on LoadPanel
```
Amazinite, line 2863:
```suggestion
	// Detect which ship number is the current flagship, for showing on LoadPanel.
```

PR #5642, opened 2021-01-16
Change a few imperial units to SI

prophile:
This is in line with established canon.

## Summary
Changes as above, I don't think in-game screenshots would be more illuminating than the diff.


data/human/hails.txt
```diff
@@ -870,7 +870,6 @@ phrase "friendly civilian"
 		"Where do astronauts hang out? At the space bar!"
 		"What is a Syndicate salesman's favorite drink? Whatever the Syndicate tell him it is!"
 		"Why does the Navy hate neighbours? Because the grass is greener on the other side!"
-		"What has two heads, six feet, two arms and one tail? A person on a horse!"
```
tatami4, line 873: Why was this removed? "Feet" here is plural of "foot", not a measurement unit.

prophile: This is a pun, because a person is six feet (units of measurement) tall but there are six feet (body parts) in this configuration. The former is a non-SI unit though :)

MCOfficer: way to remove a good pun.

Zitchas (changes requested):
They all look good except the removal of the hail pun/joke. Un-delete that, and I think this will be a good change.

data/human/hails.txt
Zitchas: The reference to non-SI units here is, honestly, actually optional. The joke *can* be read and understood completely literally with the simple understanding of six feet only referring to the six feet that a horse + human would have. Given that the imperial unit reference is simply an extra layer of entertainment, I believe that this should be left in.

data/wanderer/wanderers start.txt
```diff
@@ -17,7 +17,7 @@ mission "First Contact: Wanderer (Before Hai)"
 	on offer
 		conversation
 			`All the other worlds belonging to this species have appeared unwilling to let you land, but here you find a slightly warmer welcome. A cloud of small, highly maneuverable patrol ships takes off from the planet as you approach and herds you onto one particular landing pad where, after a few minutes, several of the aliens approach your ship.`
-			`	This species is birdlike, but with bodies nearly as large as yours, and with wings that tower above you. They come swooping down from the sky, circle your ship to inspect it, and then land near the hatch and wait for you to open it. They do not appear to be armed, but their talons are three or four inches long.`
+			`	This species is birdlike, but with bodies nearly as large as yours, and with wings that tower above you. They come swooping down from the sky, circle your ship to inspect it, and then land near the hatch and wait for you to open it. They do not appear to be armed, but their talons are substantial.`
```
Amazinite, line 20: Why not retain a measurement and say something like “eight or nine centimeters long” both here and below.

data/human/hails.txt
Zitchas: Worth noting that, at six feet tall for a rider on a horse, either the person or the horse is quite small. Possibly both. A short horse is 1.4m (4.5 feet) at the shoulder. Most people have more than 1.5 feet of height in their torse+head.

Amazinite: Yeah, the “six feet” in this pun has nothing to do with measurements of length.

prophile: Readded in e0d71968

data/wanderer/wanderers start.txt
prophile: SI convention which I think is followed here usually prefers millimetres over centimetres and "twenty or thirty millimetres" long didn't sound natural to me, happy to defer to you if you'd prefer that wording though?

Amazinite: Well it's not so much that the game follows "SI convention" than it just uses metric measurements, and just as "three or four inches long" would be used in normal speech, I don't see why "eight or nine centimeters long" wouldn't.

prophile: Places that, uh, drink the metric kool-aid thoroughly like Japan rarely use centimetres, which are mostly a phenomenon of anglophone countries picking the closest amount to an inch. That said, just let me know what wording you want.

Amazinite: I'd say just use centimeters. We use American English spellings, so not like we don't already break from more "proper" standards.

Zitchas (approved):
This looks good.

data/wanderer/wanderers start.txt
prophile: All English spelling systems are equally broken anyway :)

Alright, wilco

MCOfficer (approved):
lgtm, ty

data/wanderer/wanderers start.txt
LordInsane: Speaking from personal experience, there's definitely usage of centimetres outside the context of translating inches in at least some metric-heavy countries, especially in situations like the one in the quote where it is a quick estimate.

prophile: fde3f78e

PR #5643, opened 2021-01-16
Tweak: Looking at the Penguin's mobility in light of the Remnant Engines tweak

Zitchas:
## Issue:
It was pointed out that, with the change to Remnant engines, the Penguin's reverse thrust (which was already very strong, for its size) is now significantly higher than its normal thrust. This is a bug, from my point of view. At the very least, it should be reduced down to equal-to normal thrust. 

## Original purpose:
The original reason why the Penguin had reverse thrust at all was to illustrated the fact that - as a living creature - the Penguin is capable of moving around on its own without needing any external engines of any kind. It isn't particularly *fast* about doing it, but it can. Kind of like how a person can move around on their own, but they'll be much faster if you give them a jetpack.

**Options**
A: reduce the reverse thrust to make the total equal to the normal stock thrust, and leave it at that.
B: reduce the thrust as in A, but then add in equal regular thrust and turning to more properly represent that it can move on its own.
C: leave as is, and have the penguin have a rather impressive reverse-thrust capability.

The fix below opts for option B as the most satisfying.

## Fix Details
- Reduced reverse thrust from 16 to 5.2
- Added built-in thrust of 5.2
- Added built-in turning equal to the same ratio: 80.0

These changes give the base Penguin (without any engines or steering at all):
Max speed: 249.6
Acceleration: 126.5/145.1
Turning: 56.8/64

How I got these numbers: 
1. I reduced the reverse thrust value down to the point where it adds enough reverse thrust to bring the total stock ship up to its previous value of 16
2. I used the ratio of the old to new reverse thrust to calculate appropriate heat and energy 
3. I gave it equal regular thrust to the reverse thrust.
4. I used that ratio with the stock turning speed to get a base "natural creature" turning speed.
5. Adjusted the heat and energy accordingly. 

**Results**
Overall, this makes the Penguin a faster ship, but brings its reverse thrust speed down to about 75% of its stock thrust. When stripped of engines and steering, the penguin can accelerate and decelerate enough to not be frustrating; although the turning *is* slow enough to be frustrating. Basically, it is functional without engines, but not really a fun experience. 

On the flip side, with engines, it is a rather fun and zippy ship that feels fast enough to actually dodge stuff. Not just big lumbering antimatter bolts, but actual likely-to-be-encountered-weapons-fire.

## Testing Done
Purchased a new penguin and tried flying it around after each adjustment.

## Questions for Discussion:

1. The Penguin is two things: A prototype "living" ship, and an upgrade to the gascraft as a research vessel. Do these changes fit that?
2. How do these values feel? Are they too much? Not enough?
3. These stats are intended to reflect the Penguin's natural movement abilities. Are the heat & energy costs for them reasonable? Should they be reduced or eliminated because they *are* actually the creatures natural systems?
4. Currently, the Penguin is effectively disabled in Storms. I tried eliminating the energy/heat costs on the natural systems in hopes that it would mean the Penguin could move using them during storms even with its main engines shut down, but our game mechanics don't support that currently.


Amazinite (commented):
I'd drop the thrust/reverse thrust to 4.2 and having the turn at 80 like you had it seems like it would be better. Leaving off the energy and heat is up to you.

PR #5646, opened 2021-01-16
test/ci: Disable vsync by default on integration test CI-runs

petervdmeer:
**Bugfix:** This PR addresses vsync warnings during integration test CI runs

## Fix Details
This fix adds `vsync 0` to the default preferences of CI-tests.

## Testing Done
Testing in CI is done as part of this PR.

## Save File
N/A

### DIFF
--- a/tests/config/preferences.txt
+++ b/tests/config/preferences.txt
@@ -23,3 +23,4 @@
 "help: outfitter 3" 1
 "help: stranded" 1
 "help: trading" 1
+vsync 0


petervdmeer:
Closing: This PR provided us with the information we needed. (Can re-open when it is needed again.)

petervdmeer:
Closed wrong PR.

tehhowch:
This doesn't stop the game from attempting to set the vsync state to that which was requested, which means there will still be the unsupported operation logs. Is the intent to remove those as well?

petervdmeer:
> This doesn't stop the game from attempting to set the vsync state to that which was requested, which means there will still be the unsupported operation logs. Is the intent to remove those as well?

Yes, the intent was to remove the unsupported operation logs, but it looks like the autodetection always runs regardless of the preference setting.

petervdmeer:
Closing; This PR is not solving the issue I wanted to solve with it.

PR #5645, opened 2021-01-16
Not for merging! Endurance test for integration tests.

petervdmeer:
**Feature:** This PR contains an endurance test for checking the stability of integration tests.

## Feature Details
Not for merging to master/main!
Runs the main test-loop 20 times.
Runs the outer test 5 times.

## Testing Done
This is by definition a test.

## Performance Impact
Uses 20 * 5 = 100 times more integration test CI-capacity than a normal test-run.

PR #5647, opened 2021-01-16
tests: Retry test if it fails on framework/Xvfb issue

petervdmeer:
**Feature:** This PR implements a retry mechanism for the CI integration tests if the tests fails on a known framework issue (#4308, #5645)

## Feature Details
If the testframework detects the error-message `SDL message: "No available video device"`, then it will retry to run the test (max 5 times). This SDL errormessage would (usually) not be related to the test itself, but looks like an intermittent issue between endless-sky, Xvfb and Ubuntu 20.04. We should still handle the `SDL message: "No available video device"` error in the future, this PR only acts as a workaround to prevent CI for unrelated PRs from failing on this error-message.

## UI Screenshots
N/A

## Usage Examples
Automatically runs as part of the CI-jobs.

## Testing Done
Tested with succesfull tests, failing tests and tests failing with the specific error-message.

## Performance Impact
N/A

### DIFF
--- a/tests/run_tests.sh
+++ b/tests/run_tests.sh
@@ -28,9 +28,74 @@ function print_graphics_data () {
 
 
 
+# Helper function that checks for retryable issues in the error-log given
+# as parameter (and returns them at stdout).
 function detect_retryable_issues () {
 	grep -Fx -f "${HERE}/retryable_issues.txt" "$1"
 }
+
+
+
+# Helper function that executes a single test-run.
+# Parameters:
+# $1 Test(name)
+# Return values:
+# 0 = success
+# 1 = failure
+# 2 = fatal failure (should terminate all testing)
+# 3 = recoverable failure (could retry)
+function run_single_testrun () {
+	TEST="$1"
+	
+	# Setup environment for the test
+	ES_CONFIG_PATH=$(mktemp --directory)
+	if [ ! $? ]
+	then
+		echo "not ok Couldn't create temporary directory"
+		return 2
+	fi
+	
+	ES_SAVES_PATH="${ES_CONFIG_PATH}/saves"
+	mkdir -p "${ES_CONFIG_PATH}"
+	mkdir -p "${ES_SAVES_PATH}"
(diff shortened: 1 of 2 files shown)


tests/run_tests.sh
```diff
@@ -25,6 +25,12 @@ function print_graphics_data () {
 }
 
 
+
+function detect_known_issues () {
```
tehhowch, line 29: Should we distinguish known issues from _retryable_ issues?

tests/run_tests.sh
```diff
@@ -25,6 +25,12 @@ function print_graphics_data () {
...
+	cat "$1" | grep "SDL message: \"No available video device\""
```
tehhowch, line 30: Let's make this easily extendable, for e.g. 0 to N different issues, so a PR to add/remove one is a single line.

tehhowch (commented):
Should we break up the test loop into functions? That can make it easier to implement testrunner logic.

tests/run_tests.sh
petervdmeer: Yes, we should. I renamed those to retryable issues.

I believe tests should still fail on known unretryable issues, so I didn't add anything specific for those.

tests/run_tests.sh
petervdmeer: Agreed. I moved the issues list to a separate file.

petervdmeer:
> Should we break up the test loop into functions? That can make it easier to implement testrunner logic.

Yes, I have split off a function for `a single run of a test` and for `a complete test`.

tehhowch (commented):
Add `local` where appropriate, then lgtm

tests/run_tests.sh
```diff
@@ -25,6 +27,96 @@ function print_graphics_data () {
...
+# Helper function that executes a single test-run.
+# Parameters:
+# $1 Test(name)
+# Return values:
+# 0 = success
+# 1 = failure
+# 2 = fatal failure (should terminate all testing)
+# 3 = recoverable failure (could retry)
+function run_single_testrun () {
+	TEST="$1"
```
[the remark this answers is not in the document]
petervdmeer: done

PR #5651, opened 2021-01-17
Save human cultural data for later

Amazinite:

## Summary
There are currently two missions in the game that have the player travel to Alexandria to get a copy of human cultural data, once for the Hai and once for the Remnant. There's no reason though that the player should have to make this same trip twice when they could have easily just saved a copy of the cultural data for themselves.

This PR makes it so that if you retrieve the cultural data for either the Hai or the Remnant, a condition "human cultural data" is saved. This condition is then used when you are asked by the other group to retrieve this data, allowing the player to skip over a trip to Alexandria. By using this condition instead of checking for the `: done` condition of the opposite mission, this also allows this information to easy be used in the future should anyone else make use of it.

## Save File
No.

## PR Checklist
 - [x] Test that this actually works as intended.
 - [ ] Get approval from @Zitchas since this alters Remnant content.

### DIFF
--- a/data/remnant/remnant missions.txt
+++ b/data/remnant/remnant missions.txt
@@ -2205,6 +2205,43 @@ event "remnant: wait for plume"
 
 
 
+conversation "ately cultural data"
+	branch return
+		has "Remnant: Salvage 1B: done"
+	`As you step out into the icy chill of Viminal, you get a message from Taely, the engineering prefect, asking if you could meet her in the cafeteria.`
+	choice
+		`	(Accept her invitation.)`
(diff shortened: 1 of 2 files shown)


Zitchas (approved):
OK, if I'm understanding the flow correctly, there's now a split:
Salvage 1 and 1B are for when the player hasn't got the historical data; and
Salvage 2 is for when the player has the data, with the option of if the player starts out with the data, bypass 1 and 1B and go straight into 2.

Looks good to me. The conversation fits. A solid improvement for the gameplay/realism time-saving as well.

PR #5654, opened 2021-01-19
Fix off system escort depart checking

### LINKED ISSUE #5165: [bug] Off-system Escorts fly away to other systems with only steering Engines
Unparking off-system escorts that are not equipped with thruster engines does not stop them from jumping to other systems.

1. Park an escort.
2. Sell escorts' thruster engines (so that the ship only has steering engines).
3. Take off and jump to a far away system.
4. Land and unpark escort.
5. Takeoff and wait for escorts to start jumping systems.

Expected behavior:
Game should not let you unpark escorts and takeoff while any off-system escorts fail flight checks (engines, power, heat).

Here is a screenshot: You can see the ship description that it does not have any thruster engines, and in the lower left you can see 4 ships of similar specs.
image

System:
 - OS:Windows 10 Pro 64-bit build 18363
 - Game Source: Steam
 - Version: 0.9.12

Save file:
When you load the save file, once you takeoff, you can immediately see the thruster-less ships on the bottom left jumping systems.
[Ember Serebryakova.txt](https://github.com/endless-sky/endless-sky/files/4943801/Ember.Serebryakova.txt)

Additional info:
When I unparked these ships, I was on Glaze in the Aldhibain system.
Issue looks related to #5133

Temporary fix:
I managed to herd the ships back to a spaceport with an outfitter. 

Apparently, once you take off from a planet, the ships are able to jump ONCE, after which they are unable to come to a stop and continue to float away. However, if you land and takeoff again they will be able to do another jump. You can repeat this until the ships come to a planet with an outfitter. 

If, however, the ships run out of fuel in a system with a habitable planet before they can reach an outfitter, you can go to that system, land (and hence refuel them), park the escorts(because the game wont let you takeoff if the escorts dont have engines), go to another system, and unpark them. Then you can continue to land and takeoff repeatedly to make the thruster-less ships jump.

It can get worse however. If the ships run out of fuel in an uninhabited system, you can not refuel them by landing, and you cannot simply board them either because when you jump into the system, they are drifting off at such a speed that they cannot be boarded. The solution here is to select the escort in question(by clicking on them) and disable them by firing at them. Disabling ships removes any velocity they have so they come a stop. This way you can board them to repair and refuel them.

yjhn:
should fix #5165 and #5708

test save: [Ember.Serebryakova.txt](https://github.com/endless-sky/endless-sky/files/5866628/Ember.Serebryakova.txt)


data/conversations.txt
```diff
@@ -22,22 +22,42 @@ conversation "intro"
...
+	
+conversation "flight check: overheating escort!"
+	scene "scene/engine"
+	`You've heard enough horror stories about the sort of accidents that happen when starship systems malfunction to be completely convinced of the necessity of a pre-flight safety check. As you listen to the complex symphony of hums, rattles, and clicks of your ship warming up, you run through a mental checklist. Computers: check. Life support: check. Engines: check. Steering: check.`
+	`	It is all going smoothly until the captain of the <ship> contacts you and tells you that their ship was producing so much heat that they didn't think it would be able to travel anywhere without overheating. How embarrassing. You'll have to tell the captain of the <ship> to remain parked until you arrive to fix it.`
```
petervdmeer, line 63: Nice message, but it doesn't work for ships in the same system as the flagship (since the player already arrived there). Also for some new messages above.

tehhowch (changes requested):
I'm not a fan of the current approach here.

Rather than show a big conversation scene if any of the player's active ships in other systems are in an "error" state, just:

1. display a dialog summarizing that some escorts will be parked on takeoff, since they can't fly safely
2. auto-park said escorts upon confirmation of takeoff intent
3. perform a flight check when unparking a ship in a remote system.

tatami4:
Well, I'm not a fan of it either.

source/Ship.h
```diff
@@ -148,7 +148,7 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	// Get this ship's cost.
 	int64_t Cost() const;
 	int64_t ChassisCost() const;
-
+
```
Hurleveur, line 151:
```suggestion

```

source/PlayerInfo.cpp
```diff
@@ -776,7 +776,7 @@ map<const shared_ptr<Ship>, vector<string>> PlayerInfo::FlightCheck() const
 	
 	auto flightChecks = map<const shared_ptr<Ship>, vector<string>>{};
 	for(const auto &ship : ships)
-		if(ship->GetSystem() == system && !ship->IsDisabled() && !ship->IsParked())
+		if(ship->GetSystem() && !ship->IsDisabled() && !ship->IsParked())
```
Hurleveur, line 779: I think that ships unable to launch in other systems should be treated slightly differently.
Maybe just add a check later on if the ship is in the same system and if not specify which system the ship is from?

Hurleveur (changes requested):
Lore and stuff isn't my area of expertise, but I personally liked the conversations with problems to launch, especially from an immersion standpoint.
Maybe we could just make it give the big dialog only if the ship is in the same system, and the quick one otherwise?
All whitespace changes should be removed anyway, and this looks like a good fix.

Amazinite:

Marking this PR as orphaned as there hasn't been activity from OP in some time, but not closing because it's a bug fix that we should have. Either someone should open a new PR with these changes finalized, or one of the devs could finalized this PR themselves.

Amazinite:
I keep getting an authentication failure when pushing to specifically this branch, so I'm just going to fork it and reopen it under a new PR.

PR #5664, opened 2021-01-24
Fix: Cognizance system data sorting

Zitchas:
**Bugfix:** This PR addresses the issue that the new systems in the Ember Waste's data isn't sorted correctly.

## Fix Details
Moving system attributes into the following order:
- Pos
- government
- attributes
- haze
- link
- asteroid
    - first by type, then by size.
- minables
- hazards
- objects
    - From innermost to outermost.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -8007,6 +8007,7 @@ system Denebola
 		period 2740.45
 
 system Diespiter
+	pos 332 210
 	asteroids "large metal" 76 3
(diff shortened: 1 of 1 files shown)


Amazinite (changes requested):
The planets also need to be sorted by distance, and the asteroids need to be sorted by small/medium/large rock, then small/medium/large metal.

Issue #3663, opened 2018-03-21
Give players less of a need to seek information outside the game


Amazinite:
A vast majority of players likely don't look outside of the game for extra information about it, mainly concerning official player's manuals and the story section of the wiki. Preferably, there should be little to no need to look outside the game for such information. Right now a player can find out most of the stuff that the player's manual has through playing the game, and there are tidbits of information about the regions, organizations, factions, and history of the game, but not everything from the wiki can be found in game, e.g. much of the historical timeline. 

My proposal is to make more use of the logbook to put this information into the game. For example, upon starting the game, in addition to the small bits of info that the player gets about the Republic, Syndicate, and pirates, the player could receive a "Pilot's Manual" special log entry that gives the player basic information from the player's manual. As for the historical timeline, a player could receive a data containing a special log entry for "History" that goes through the information in the wiki's historical timeline. I've already written such a set of special log entries as a proof of concept, putting the historical timeline into a history book-like format: [codex.txt](https://github.com/endless-sky/endless-sky/files/1833105/codex.txt). The only thing really needed for this is to find out where the player actually gains access to this data.

The acquiring of a timeline of human history in game would also open up the possibility to acquiring the history of other factions in game, meaning we could finally solve this problem from a while back by putting the information in game where it has to be earned instead of on the wiki where it can be a major spoiler.
> [I'm not sure if I should add some of that timeline and species information to the wiki, or if it'd present way too much of a spoiler for new players.](https://github.com/endless-sky/endless-sky/issues/2540#issuecomment-302979219)

ghoulavenger:
I like it on one hand, and on the other not so much. From a storytelling perspective that information is very useful since it helps you develop the setting further and maintain internal consistency. From a gameplay perspective, you only need to know how one story relates to another, so that the history you need to know can and should be introduced only as it pertains to the story being told.

For example, in most Elder Scrolls titles you have a bunch of lore books that you can read at any time. None of them do you have to read to make sense of any of the story, since it is transitioned to you in pretty much every questline where it is even remotely important. This makes the game fluid and responsive. I already feel like Endless Sky attempts to do this, by filling in the most important information into the quests. The only part where Endless Sky fails in this regard is that you can completely skip (and/or miss) the quests.

[...]

Issue #2540, opened 2017-05-22
Typo in Rim Archaeology 1?


Amazinite:
https://github.com/endless-sky/endless-sky/blob/master/data/transport%20missions.txt#L997
> We've heard rumors from the Quarg that an alien civilization used to exist in the Rim, a hundred of thousand years ago.

Shouldn't it be "hundreds of thousands of years ago" or "a hundred thousand years ago," because "a hundred of thousand years ago" doesn't appear to be right.

MessyMix:
Seems like a typo to me.

endless-sky:
Good catch. Fixed in 1abd93d.

The Sheragi discovered spaceflight around 100,000 years ago, about the time of the peak of the Hai empire, around the same time that the ancestors of modern humans began "banging rocks together." The Sheragi civilization survived less than a millennium after that.

The Hai discovered spaceflight about 120,000 years ago, and their civil wars stretched from about 95,000 to 90,000 years ago. 90,000 years ago, the Hai pulled back from "human" space and from the ruined worlds to the north of them, and 80,000 years ago the Wanderers moved into that northern territory.

The Korath, Saryds, Kimek, Arachi, and finally the humans, discovered spaceflight 40, 17, 14, 10, and 1 thousand years ago, respectively. Note that that means that new spacefaring species have been appearing more and more frequently in recent history, which is part of why the Archons have been having so much trouble holding things together lately.

Amazinite:
Holy lore.

Amazinite:
So the Unfettered are chill with telling us they discovered space flight 10 times earlier than they actually did? (One says over a million years now gone in "First Contact: Unfettered.") Aight, that's another lie detected.

Amazinite:
>80,000 years ago the Wanderers moved into that northern territory.

Let it be known that 1 Wanderer cycle = 8 years, as in "Wanderers Conversation" they say they've been at their current location for 10,000 cycles.

Bladewood:
https://github.com/endless-sky/endless-sky/commit/ff8315834429f12e891db5358e50339adfcf78aa nope

Amazinite:
Lol, time to change my own time line.

Amazinite:
[...]

PR #5665, opened 2021-01-26
Update references to population sizes of human worlds

Amazinite:
**Content**

## TL;DR (Isn't that what a summary is suppose to be?)

I simulated population sizes of all the worlds in human space. Some planets have mentions of their population size, but the numbers I ended up with don't match the descriptions. Instead of lowering the total number of people as given by MZ at roughly 100 billion, and instead of bringing up the population density of lesser inhabited planets by an insane amount just so I could reach that final number without clashing with the planet descriptions that existed, I've decided to change some descriptions according to what I simulated, essentially canonizing the numbers I came up with.

Link to my numbers is [here](https://docs.google.com/spreadsheets/d/1dWKimAiiXrlVCibdjyiFNqtQJdiXJqB7F0Cs9La0Z60/edit?usp=sharing).

The sheet also includes settlement dates for planets where they could be easily known. No, I'm not going to finish creating settlement dates for all planets right now. Yes, I'll eventually do it later, and we can canonize that too.

## (The not so) Summary
A while back now, I asked MZ for population numbers on the various species in the game, and he gave some rough numbers [here](https://github.com/endless-sky/endless-sky/issues/2540#issuecomment-302986796). A question that that has begged though is how those roughly 100 billion humans are spread out across the galaxy. Only a few planet descriptions actually mention anything concrete about population size of the planet, those being Valhalla, Midgard, Thule, Sunracer, Clark, Heartland, New Sahara, Sinter, Prime, Zug, and Thunder. We then learn from the Shuttle intro how many people Earth has. So great! That's 12 planets out of... 157 accounted for. Some other planet descriptions give numbers on a subset of the population, like millions of Amish on Harmony and tens of thousands of homeless on Bourne, but without more information on the proportion of the population within those groups, the descriptions really only give us lower bounds.

There are some more things we can look to, like how Earth is more than twice as populous as the next closest planet, and Chiron is the second most populous planet, but for the most part we don't have much in the way of figuring out how populated various regions of space are. At best we get a description on many worlds of them being sparsely populated or only having a single settlement or only having towns and no cities or being overpopulated. Well what defines a world as sparsely populated? How many people can a single settlement hold? How many people need to be in a region before it's considered a city? How many people does your average town have? How many people does a planet have to be considered overpopulated? Is overpopulation a factor of resources per person or of number of people? We might look to how we'd define these terms on Earth today, but I imagine that cities of the future are far larger and more impressive than those today, so it takes more to be considered a major city in the Republic.

I've wanted for a while now to try and make out how many people are on each planet, but it's only recently after some discussion on Discord that I actually cared to get around to it. Over the course of three days I read through every planet description in human space, as well as looked at all the attributes on all the planets, in order to get a rough idea of the ranking of the planets. Some planets were easy, like I mentioned above, providing a rough number for me. Most though mentioned nothing of population and only described the state of the planet. Using those descriptions of how nice of a place the planet was, and in some cases giving me a timeline of when the planet was settled, I was able to roughly rank all the planets in terms of population by assigning them random population numbers. These original numbers I came up with only came out to about 60 billion, but using the rankings I was able to create a function that would give me population numbers for a planet depending on its rank (although I did have to fiddle with it by often skipping ranks, e.g. most populous planet 20 and most populous planet 21 might have actually been considered rank 20 and rank 22 in the function, and toward the end planet 140 was actually like rank 2000 in the function, but whatever the case I got a distribution I like and a total value that is roughly around the original number of 100 billion).

For the curious, the function I used was an altered curve that describes the current populations of countries on Earth, scaled to fit Earth at about 20 billion. Seeing as how a good number of planets are named `"New <country>"` or `"New <city>"`, comparing planets to countries doesn't seem too out there.

My final spreadsheet for the population of planets can be found [here](https://docs.google.com/spreadsheets/d/1dWKimAiiXrlVCibdjyiFNqtQJdiXJqB7F0Cs9La0Z60/edit?usp=sharing). The top of this spreadsheet includes the population of groups of humans, the first being humanity as a whole, then going down into regions of humanity, some of which overlap. The "no planet of residence" line is a random number that I pulled out that refers to the number of people who basically live on ships, so merchant captains and their crew, or militia members, or maybe even Navy and Syndicated Security members that spend most of their time amongst the stars. The sheet then sorts the planets into regions, sorting the regions alphabetically and sorting the planets within the regions by population size.

As you might be able to see though, I've changed the population sizes of some planets from how they were described. I already mentioned how my original numbers came out to about 60 billion. Well in order to get 40 billion more people, I would have had to create many, many planets with close to or over a billion people on them, and planets described as sparsely populated would have a lot more people on them than I would expect. So instead of trying to force 40 billion into the upper limit that we had, which was 10 billion on Earth and 5 billion at most on any other planet, I decided to tweak some of the numbers we had. This includes bumping Earth's population from 10 billion to 20 billion, which allowed me to drag some other planets up with that, such as increasing Chiron's population from about 5 billion to about 10 billion, and Moonshake's population from the 3.5 or 4 billion that I originally had to about 7 billion. All these changes allowed for me to reach a total population that was close to 100 billion.

While this PR only changes descriptions to match my spreadsheet, everything from my spreadsheet can't be found in game. This is where #3663 gets involved, which entails providing information like this to the player in-game in an in-universe way. My idea as to how this data might be provided in-universe is by saying that this is census data. Perhaps the Republic does a census every 20 or 25 years, and 3000 was when the last census was done. (You can also see that I simulated what the population growth would need to be to hit 100 billion people by 3013, game start, and the answer is an average of 0.195% per year for all of humanity.) The player could then probably get this census data from Alexandria. Other than that, perhaps we create spaceport news that only appears on an exact planet for every planet, giving the player more information on how life is on each planet in the game, including mentions of the population size (perhaps not exact, but rough numbers).

## Save File
N/A

## PR Checklist
N/A

### DIFF
--- a/data/intro missions.txt
+++ b/data/intro missions.txt
@@ -161,7 +161,7 @@ mission "Intro [3 Transport]"
 	on offer
 		conversation
 			`The spaceport on <origin> is so big and crowded that it takes a long time to find James. The sheer number of people here is almost overwhelming. Homeless men wrapped in old army blankets beg money from well-dressed businessmen. Soldiers in Navy uniforms walk by in groups of ten or twenty, all marching in unison. Starship captains haggle with merchants over trade goods and jobs. Flashes of light come from various landing bays as welders make repairs.`
-			`	When you find James, he's standing by a window, looking out at the city, watching the starships taking off and landing in a continuous stream. "This is it," he says, "where our race began. The world that all the tourists want to visit, and that no one wants to live on. Ten billion people, more industry and exports than any other place in the galaxy. But underneath it all, poverty of a completely different sort from what you've got in the Dirt Belt."`
+			`	When you find James, he's standing by a window, looking out at the city, watching the starships taking off and landing in a continuous stream. "This is it," he says, "where our race began. The world that all the tourists want to visit, and that no one wants to live on. Twenty billion people, more industry and exports than any other place in the galaxy. But underneath it all, poverty of a completely different sort from what you've got in the Dirt Belt."`
 			`	After taking one last glance out the window, he says, "Anyway, we're nearing my destination. Are you willing to take me on one last trip? I need to get to <destination>, and of course I'll pay you quite well to take me there."`
 			choice
 				`	"Of course."`
--- a/data/map.txt
+++ b/data/map.txt
@@ -26657,7 +26657,7 @@ planet "Chosen Nexus"
 planet Clark
 	attributes "dirt belt" forest factory
 	landscape land/valley3
-	description `Clark is a recently settled world, and most of its surface remains covered in dense forests and uninhabited mountains. Because clearing new land is so difficult, agriculture here has not grown far beyond the level of subsistence farming. But the temperate climate has drawn several million settlers, most of whom work in factories that have been planted here by off-world corporations to take advantage of the cheap labor.`
+	description `Clark is a recently settled world, and most of its surface remains covered in dense forests and uninhabited mountains. Because clearing new land is so difficult, agriculture here has not grown far beyond the level of subsistence farming. But the temperate climate has drawn tens of millions of settlers, most of whom work in factories that have been planted here by off-world corporations to take advantage of the cheap labor.`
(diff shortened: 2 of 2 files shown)


Bardagh:
Much respect and appreciation.

Issue #5540, opened 2020-12-10
Combinations of disables and plunders personality don't work as expected


Amazinite:
**Describe the bug**
The behavior that one might expect of the disables and plunders personality in certain combinations isn't what is seen in game. To recap, the following are the descriptions of these personalities from the wiki:
* `disables`: tries to disable enemy ships rather than destroying them.
* `plunders`: will board and pillage enemy ships before destroying them.

Personality | Expected | Actual | Issue
--|--|--|--
disables | The ship won't attack disabled enemies. | The ship won't attack disabled enemies. | None.
plunders | The ship will leave enemies disabled until boarding them, then destroys them. | The ship will leave enemies disabled until boarding them, then take pot shots at them after finished boarding, sometimes not destroying them, instead of outright destroying them. | Ships with plunders alone should always finish off their target instead of just taking pot shots at them like merchants do to disabled ships.
disables + plunders | The ship will leave enemies disabled, even after boarding them. | The ship will leave enemies disabled until boarding them, then take pot shots at them after finished boarding. | A ship with the disables personality should never attempt to fire on disabled ships. The combination of disables + plunders matches the behavior of plunders alone.

**To Reproduce**
See table. Simply test different fleets with the varying personalities.

**Expected behavior**
See table.

**Screenshots**
N/A

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Self-compiled
 - Version: latest commit as of posting (https://github.com/endless-sky/endless-sky/commit/671f9b6b2d4c156420e7537190040f445392e938)

**Link to save file**
N/A

**Additional context**
N/A

PR #5667, opened 2021-01-27
Ships with "disables" personality no longer kill plundered ships

tatami4:
**Bugfix:** This PR addresses issue #5540

This creates another problem: endless plundering (since the plundered ships are no longer killed, every ship entering the system that can plunder it does so).

## Save File
[Craig.McCauley.txt](https://github.com/endless-sky/endless-sky/files/5881207/Craig.McCauley.txt)

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2876,7 +2876,8 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 		currentTarget.reset();
 	
 	// Only fire on disabled targets if you don't want to plunder them.
-	bool spareDisabled = (person.Disables() || (person.Plunders() && ship.Cargo().Free()));
+	bool plunders = (person.Plunders() && ship.Cargo().Free());
+	bool disables = person.Disables();
 	
 	// Don't use weapons with firing force if you are preparing to jump.
 	bool isWaitingToJump = ship.Commands().Has(Command::JUMP | Command::WAIT);
@@ -2949,7 +2950,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 		{
 			// NPCs shoot ships that they just plundered.
 			bool hasBoarded = !ship.IsYours() && Has(ship, currentTarget, ShipEvent::BOARD);
-			if(currentTarget->IsDisabled() && spareDisabled && (!hasBoarded || person.Disables()) && !disabledOverride)
+			if(currentTarget->IsDisabled() && (disables || (plunders && !hasBoarded)) && !disabledOverride)
 				continue;
 			// Don't fire secondary weapons at targets that have started jumping.
 			if(weapon->Icon() && currentTarget->IsEnteringHyperspace())
@@ -2984,7 +2985,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 		{
 			// NPCs shoot ships that they just plundered.
 			bool hasBoarded = !ship.IsYours() && Has(ship, target, ShipEvent::BOARD);
-			if(target->IsDisabled() && spareDisabled && (!hasBoarded || person.Disables()) && !disabledOverride)
+			if(target->IsDisabled() && (disables || (plunders && !hasBoarded)) && !disabledOverride)
 				continue;
 			
 			Point p = target->Position() - start;
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1401,10 +1401,6 @@ void AI::MoveIndependent(Ship &ship, Command &command) const
 			MoveTo(ship, command, target->Position(), target->Velocity(), 40., .8);
 			command |= Command::BOARD;
 		}
-		else if(hasBoarded && ship.GetPersonality().Disables())
-		{
-			// Ships with "disables" personality don't destroy their targets after boarding.
-		}
 		else
 			Attack(ship, command, *target);
 		return;
@@ -2953,7 +2949,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 		{
 			// NPCs shoot ships that they just plundered.
 			bool hasBoarded = !ship.IsYours() && Has(ship, currentTarget, ShipEvent::BOARD);
-			if(currentTarget->IsDisabled() && spareDisabled && !hasBoarded && !disabledOverride)
+			if(currentTarget->IsDisabled() && spareDisabled && (!hasBoarded || person.Disables()) && !disabledOverride)
 				continue;
 			// Don't fire secondary weapons at targets that have started jumping.
 			if(weapon->Icon() && currentTarget->IsEnteringHyperspace())
@@ -2988,7 +2984,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 		{
 			// NPCs shoot ships that they just plundered.
 			bool hasBoarded = !ship.IsYours() && Has(ship, target, ShipEvent::BOARD);
-			if(target->IsDisabled() && spareDisabled && !hasBoarded && !disabledOverride)
+			if(target->IsDisabled() && spareDisabled && (!hasBoarded || person.Disables()) && !disabledOverride)
 				continue;
 			
 			Point p = target->Position() - start;
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -1090,13 +1090,6 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 	if(!gov || ship.GetPersonality().IsPacifist())
 		return target;
 	
-	// Ships with 'plunders' personality always destroy the ships they have boarded.
-	const Personality &person = ship.GetPersonality();
-	shared_ptr<Ship> oldTarget = ship.GetTargetShip();
-	if(oldTarget && oldTarget->IsTargetable() && oldTarget->IsDisabled() && person.Plunders()
-			&& !person.Disables() && Has(ship, oldTarget, ShipEvent::BOARD))
-		return oldTarget;
-	
 	bool isYours = ship.IsYours();
 	if(isYours)
 	{
@@ -1117,11 +1110,17 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 	if(!maxRange)
 		return target;
 	
+	const Personality &person = ship.GetPersonality();
+	shared_ptr<Ship> oldTarget = ship.GetTargetShip();
 	if(oldTarget && !oldTarget->IsTargetable())
 		oldTarget.reset();
 	if(oldTarget && person.IsTimid() && oldTarget->IsDisabled()
 			&& ship.Position().Distance(oldTarget->Position()) > 1000.)
 		oldTarget.reset();
+	// Ships with 'plunders' personality always destroy the ships they have boarded.
+	if(oldTarget && oldTarget->IsDisabled() && person.Plunders()
+			&& !person.Disables() && Has(ship, oldTarget, ShipEvent::BOARD))
+		return oldTarget;
 	shared_ptr<Ship> parentTarget;
 	bool parentIsEnemy = (ship.GetParent() && ship.GetParent()->GetGovernment()->IsEnemy(gov));
 	if(ship.GetParent() && !parentIsEnemy)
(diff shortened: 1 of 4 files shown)


source/AI.cpp
```diff
@@ -2949,7 +2949,7 @@ void AI::AutoFire(const Ship &ship, Command &command, bool secondary) const
 		{
 			// NPCs shoot ships that they just plundered.
 			bool hasBoarded = !ship.IsYours() && Has(ship, currentTarget, ShipEvent::BOARD);
-			if(currentTarget->IsDisabled() && spareDisabled && !hasBoarded && !disabledOverride)
+			if(currentTarget->IsDisabled() && spareDisabled && (!hasBoarded || person.Disables()) && !disabledOverride)
```
Amazinite, line 2952: `spareDisabled = person.Disables() || (person.Plunders() && ship.Cargo().Free())`, so we basically have `person.Disables()` in here twice. 

Reducing the booleans to letters, we have `A && (B || (C && D)) && (!E || B)`. (Ignoring disabledOverride since it's a player command thing.) That has the same truth table as `A && (B || (C && D && !E)`.

So maybe replace the `spareDisabled` variable with a `plunders` and `disables` variable, where `bool plunders = person.Plunders() && ship.Cargo().Free();` and `bool disables = person.Disables();`. Then this would become the following here and below:
```suggestion
			if(currentTarget->IsDisabled() && (disables || (plunders && !hasBoarded) && !disabledOverride)
```

Amazinite:
> This creates another problem: endless plundering (since the plundered ships are no longer killed, every ship entering the system that can plunder it does so).

Well if we have disabled + plunders then that should be the behavior, so I don't see that as a problem. If the problem is who has that behavior, then we should drop disabled from their personality.

tatami4:
Most of the Remnant fleets are disabling + plundering.

Zitchas:
So, what's the end result here?
 
`Disabling` = Ships with this never kill ships.
`plundering` = Ships with this try to disable ships, plunder them, and then once all ships in the system with plundering in the system have had a chance, the last one destroys it. (So, basically plundering is equal to the current disabling+plundering)

tatami4:
> plundering = Ships with this try to disable ships, plunder them, and then once all ships in the system with plundering in the system have had a chance, the last one destroys it. (So, basically plundering is equal to the current disabling+plundering)

Ships with `plundering` personality destroy the ship they have boarded right after boarding it (will be once I finish this), they do not wait for others to board.
And yes, `disabling` + `plundering` before this PR will be equal to `plundering` after this PR and `disabling` will probably only be useful for pirates.

Zitchas:
I feel like I'm missing something here. I regularly watch whole fleets of Remnant cycle through boarding a ship, and it seems like it isn't until the last one has boarded that they destroy the ship. I'd like to preserve that behaviour, one way or another.

tatami4:
> I feel like I'm missing something here. I regularly watch whole fleets of Remnant cycle through boarding a ship

Strange. I don't see anything in the code that would do this.

Amazinite:
> I feel like I'm missing something here. I regularly watch whole fleets of Remnant cycle through boarding a ship, and it seems like it isn't until the last one has boarded that they destroy the ship. I'd like to preserve that behaviour, one way or another.

Per #5540, there is currently no difference between plunders and disables + plunders in game right now, that is that the ship will disable its target, plunder it, then destroy it. Adding the disables personality does nothing because ships with plunders already leave ships disabled to be plundered later, and it isn’t changing their behavior of destroying plundered ships.

This PR makes it so that disables + plunders works as expected, never intentionally destroying a disabled ship due to the disabled personality, but plundering disabled ships due to the plunders personality. The plunders personality should be unchanged, meaning that any ship that had just plunders will remain the same, and any ship that had disables + plunders should probably lose disables if we want to retain the old behavior.

tatami4:
Both issues mentioned in #5540 should be fixed now.

source/Ship.cpp
```diff
@@ -2177,15 +2177,14 @@ shared_ptr<Ship> Ship::Board(bool autoPlunder)
...
-		SetTargetShip(shared_ptr<Ship>());
 		
 		// Pause for two seconds before moving on.
 		pilotError = 120;
 	}
 	
 	// Stop targeting this ship (so you will not board it again right away).
-	SetTargetShip(shared_ptr<Ship>());
+	if(personality.Disables() || !personality.Plunders())
+		SetTargetShip(shared_ptr<Ship>());
```
Amazinite, line 2187: Which AI ships will even reach this point in the code that don't have the plunders personality? The only time ships board each other are if a ship is assisting a friendly, or if a ship has the plunders personality and is boarding a hostile. The friendly assisting is handled above this and exits the function. So wouldn't it be the case that the only ships to reach this point are ships with plunders, meaning this if statement is always true for non-player ships? 
```suggestion
	if(isYours)
		SetTargetShip(shared_ptr<Ship>());
```
The comment even says so *you* will not board it again right away, as the target wasn't reset by this point if you're the one doing the boarding since autoPlunder is false for the player.

source/AI.cpp
```diff
@@ -1090,6 +1090,13 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 	if(!gov || ship.GetPersonality().IsPacifist())
 		return target;
 	
+	// Ships with 'plunders' personality always destroy the ships they have boarded.
+	const Personality &person = ship.GetPersonality();
+	shared_ptr<Ship> oldTarget = ship.GetTargetShip();
+	if(oldTarget && oldTarget->IsTargetable() && oldTarget->IsDisabled() && person.Plunders()
+			&& !person.Disables() && Has(ship, oldTarget, ShipEvent::BOARD))
+		return oldTarget;
+
```
Amazinite, line 1099: Why do this all the way up here? I'd move these first two lines back to where they were, then put the plunders check below the timid check. (This also lets us drop the `oldTarget->IsTargetable()` part since that's done below. We know from lines 1115-1116 that for anything below that point if oldTarget exists, oldTarget is targetable.)

Amazinite:
How about adding `disables` to small pirate fleets?
Looking through the game so far, the only fleets that spawn with disables + plunders (ignoring any mission fleets) are the Unfettered, the Remnant, and Korath raids. This personality combo will make it so that these fleets never destroy the ships they plunder from, so does it makes sense for these fleets to act this way?

In the case of the Unfettered, I'd say yes. The Unfettered fight for the excitement of it. I don't imagine that the death of every Hai they can possibly cause is their goal.

In the case of the Korath, I could go either way. I think it makes sense that in a target left alive is a target you can plunder again in the future, and the Korath need every bit of supplies that they can get. But on the other hand, the Korath probably don't care at all for who they're plundering from, so they might just finish them off after they're done plundering, which could also help to keep the Korath hidden by leaving less people around that know about them, if that's something they even care about.

In the case of the Remnant, you could make the same argument as the Korath keeping the combo; a target left alive is a target to plunder for another day. But the Remnant canonically plunder Korath outfits, so they're leaving ships inoperable after plundering, in which case I think it makes more sense to finish them off. There's also the fact that the player plundering Raiders if very lucrative, so I don't think we want to make it any easier than it already is.

As for adding the disables personality to small pirate fleets though, this would be more of a difficulty thing. Having small pirate fleets leave the player disabled means you don't need to save scum as much in the early game if you are being disabled by such fleets. The alternative would be to give pirates from certain regions the disables personality. Perhaps all southern pirates disable, but northern pirates don't, and only small core pirates disable.

Zitchas:
Transposed thoughts on the Remnant / Korath situation:

Ideally, I think I'd like it to be along the lines of "Disable ships, plunder them, then maybe destroy them if the ship's been hanging around disabled for more than a few minutes."

**My thoughts from a lore perspective:**
1) Fundamentally, the Remnant are about survival, and protecting themselves. This implies destroying ships that represent a significant hazard to them. Most notably in missions this involves Korath that seem to have switched to just killing stuff instead of looting; and ships that are doing recon on them (aka learning too much about them.)

2) The Remnant and the Korath have a tribalistic trade system going on: The Korath provide ships and outfits; the Remnant provide base resources, minimally processed materials, and other commodities. Basically the two meet in orbit and fight it out; with whoever fights better getting the better deal out of that particular day's trading. This implies that the Remnant are *NOT* trying to kill every last Korath ship that they see, since if they kill everything, then the Korath won't get any resources, and the Remnant's source of fresh samples of ships and outfits will dry up very quickly.

**And from a mechanics perspective.**
3) If the Remnant simply disable, plunder, and abandon Korath ships; then eventually their systems are going to get clogged with disabled Korath ships all over the place. We need some way to clear things out on a regular basis. (Either Korath ships repairing and fleeing, getting repaired and fleeing, or giving the ion storms a small amount of hull damage so ships aren't likely to survive more than a couple of them if they are already destroyed).

4) I don't think it is good, desirable, or realistic to have Remnant systems turn into lucrative feeding grounds for players.

**Ideas**; (for future stuff beyond this PR)
5) Would it be possible to have a "repair on plunder" behaviour paired with a "If repaired don't fight just run" behaviour? My thought is that the most realistic thing for the Remnant (and, for that matter, the Korath) to do is to board enemy ships, take outfits, repair the Korath ship, and leave; while the Korath ship immediately leaves the system. Remnant wouldn't take the Hyperdrive, and hyperjumping doesn't take any energy, and they are already motionless. Once out of system we can lore that they are picked up by other Korath ships that provide the necessary repairs for them to limp home.

Of course, that's kind of complex AI behaviour, since it has a bunch of parts to it:
a) start off hostile as per usual, with the "disables" and "plunder" pattern.
b) After a target has been disabled and plundered, repair the ship despite the fact that it is an enemy.
c) Ignore the ship that it just repaired, despite being an enemy.
d) Ships that have this personality that *have been* disabled and boarded, if repaired by an enemy ship, refuse to fight and immediately flee the system.

Come to think of it, sounds like a great behavior (minus the plundering part) for the FW and Navy. The whole oathkeeper thing of "let's release them again on oath they don't fight us" could be interesting to see in a battle: Disable the navy ship, board them, get their oaths, repair, navy ship runs away.

**Short term proposition:**
- Let's go with the simple "disables" behavior for both Remnant and Korath.
- Let's add a small amount of hull damage to the ion storm. Enough so that 2 bad storms or 3-4 average ones will deal enough damage to destroy a disabled Raider. 

(For what it is worth, the "Investigate Ion Storms" mission that I have PR'd already adds a small amount of hull damage as a more reliable means of killing the ion timer ship. I'll probably need to up the damage a bit, but probably not by a whole lot. )

Long term, it'd be nice to have that rather complex AI pattern.

Amazinite:
:-1: on using hazards as a means of garbage control. It can take a lot of time for 3-4 storms to happen, *ESPECIALLY* in the Remnant systems where we've reduced the storm frequency. I'd just ditch disables on Remnant fleets.

source/Ship.cpp
tatami4: The target is only reset here now, so it has to be reset in all cases except when the ship is `plundering` and not `disables`.

Zitchas:
That's not really a problem, though. Unlike Mesuket or other places where people like to hang out to catch disabled ships, the Raiders aren't really that common. For instance, in Arculus, the timing for a Korath Raid is 40000 while the time for a storm is 25000. So in general, there should be at least one storm per raid. Having the player able to find one Raider disabled isn't a problem; we just don't want a pile of them to stack up. And if storms happen more often than raids, then I doubt we'd ever see more than 3 Raiders disabled there. Could be more if they get a bunch of weak storms, though.

(In terms of values, I was thinking 0.15 hull damage. That translates to a total of 45 damage on a very weak short storm, 225 on a very weak but max length storm, and 675 on the harshest of storms.)

tatami4:
By the way, the wiki entry for `plundering` personality should be changed  to "will disable, board and pillage enemy ships before destroying them."

source/AI.cpp
```diff
@@ -1117,6 +1117,10 @@ shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
 	if(oldTarget && person.IsTimid() && oldTarget->IsDisabled()
 			&& ship.Position().Distance(oldTarget->Position()) > 1000.)
 		oldTarget.reset();
+	// Ships with 'plunders' personality always destroy the ships they have boarded.
+	if(oldTarget && oldTarget->IsDisabled() && person.Plunders()
+			&& !person.Disables() && Has(ship, oldTarget, ShipEvent::BOARD))
```
Amazinite, line 1122: Check personalities before checking the status of the target as is done for the timid check.
```suggestion
	if(oldTarget && person.Plunders() && !person.Disables() 
			&& oldTarget->IsDisabled() && Has(ship, oldTarget, ShipEvent::BOARD))
```

PR #5669, opened 2021-01-28
Change Kestrels to use new bay syntax

Anarchist2:
Change bay definition of Kestrels to match the change in 79260f206e28370f1a26f1b32b8011babe3794cd.

### Additional note
Also missing are bay definitions for the Prototype Kestrel. Should bays be added to this variant, or be held back for the finalised Kestrel?

### DIFF
--- a/data/human/kestrel.txt
+++ b/data/human/kestrel.txt
@@ -308,8 +308,8 @@ ship "Kestrel"
 		"Hyperdrive"
 	engine -14 177
 	engine 14 177
-	fighter -46 106 under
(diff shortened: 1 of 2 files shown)


Zitchas:
I don't know. How "prototype" do we think it is? Are bays something that wouldn't have been added yet? (which does seem reasonable, given the prototypical nature. But on the other hand, the conversation makes it sound like it is virtually complete, too.)

I might opt to leave them out, just for simplicity.

petervdmeer (approved):

PR #5673, opened 2021-01-28
Void Sprite 3d remake

beccabunny:

## Summary
Currently, void sprites are slightly edited 2d pics of the Glaucus Atlanticus, an RL animal, which means lighting won't be as consistent and having angles that aren't plain top view is impossible. This PR remakes them in blender and adds their thumbnails too
Gifs:
ezgif com-gif-maker
ezgif com-gif-maker2

## PR Checklist
[void sprite.zip](https://github.com/endless-sky/endless-sky/files/5893871/void.sprite.zip)
 - [x] high DPI https://github.com/endless-sky/endless-sky-high-dpi/pull/164


Amazinite:
Any reason why the adult is bigger than it was before?

beccabunny:
The undersized image was roughly 760x760, so to avoid rerendering it just to have it scale down exactly I went for 380x380. If it is very necessary i can redo it

beccabunny:
The original xfcs were already more than large enough to have 2x even before; their omission was deliberate, likely to avoid using too much memory just for the void sprites 2x with so many frames

Zitchas:
If anyone is running out of memory, they shouldn't be using the high-dpi plug-in. Or they can cull whichever parts of the plug-in they get the least benefit out of. For instance, I'd rather cull all the coalition and hai ships than the Void Sprites.

Amazinite:
Size is no issue.

beccabunny:
Resized the larger void sprite to be closer to the original and created a high DPI PR https://github.com/endless-sky/endless-sky-high-dpi/pull/164

Zitchas:
Overall, I'm liking them. I just spent five or ten minutes hanging around Nenia watching them, and yeah, overall, I think I'm looking forward to merging them and having them as a permanent part of the game.

One thing I have noticed, however, is that the Infant VS animation looks jagged. As in, it is slow enough that I can visually pick out each frame instead of seeing it as something in motion. Digging back into the image files, I note that the infant has 9 frames versus the adult's 13. I don't recall this happening previously, but it does appear that this is matching the number of frames it used to have. So I'm going to make myself a plug-in that adds a parallel set of void sprites with the old animations so I can compare them simultaneously,

Zitchas (approved):
Looks beautiful!

Thank you for adding in the double stripe to more closely match the original Void Sprites. And thanks to Amazinite for rendering the extra frames so they both have a very smooth and organic feel to them!

Amazinite:
[Here's](https://github.com/endless-sky/endless-sky/files/5897449/sprites.zip) the new gimp files with all the frames..

PR #5677, opened 2021-01-30
Player's ships do not land on planets they cannot land on

### LINKED ISSUE #5663: Can land Fleet on Nesqueron; change ships
Mechanics issue: 
At the mission after defeating the Korath fleet attacking the void sprites, one can land a Penguin on Nasqueron along with a fleet. Can then change into a different vessel to engage Korath, scoop Jump Drives, etc. 

**To Reproduce**
Steps to reproduce the behaviour:
1. After the Korath fleet has been defeated, bring the Penguin as the flagship. Land on Nasqueron.
2. Move a different ship into the flagship position.
3. Take off in a different vessel.

**Expected behaviour**
Seems to me the fleet should be unable to land on the gas giant ... 
Jan 8 nightly build
[Speedy Wilson~To Nasqueron 2.txt](https://github.com/endless-sky/endless-sky/files/5862419/Speedy.Wilson.To.Nasqueron.2.txt)

yjhn:
**Bugfix:** This PR addresses issue #5663

## Fix Details
1. Player can no longer reorder ships when landed on planets which have "requires: gaslining" attribute.
2. Player's ships no longer launch from planets they cannot land on.

## Testing Done
Landed on a normal planet, reordered ships.
Landed on a gas giant, was unable to reorder ships.
Launched from a gas giant, ships were waiting in the system.
Launched from a normal planet, ships launched together.

## Save File:
[Harry Seldon.txt](https://github.com/endless-sky/endless-sky/files/5898074/Harry.Seldon.txt)


Zitchas:
Question: When it says "cannot reorder ships when landed on a gas planet," does it mean we cannot reorder *any* ships, or just that we can't reorder ships that do not have the gaslining trait?

Some thoughts: (Note: I'm not a coder. If this PR is actually doing the following, great!)
- "gaslining" isn't a hardcoded attribute, as best I can tell. It is just a random attribute that we can assign ships or outfits that happens to be required to land on certain planets. We could in theory add various other similar traits. "Corrosive Environment," "flotation devices," etc.
- Restrictions on resorting shouldn't be tied to any one attribute, but rather a more general "if this planet requires an attribute, the player should only be able to sort (and swap flagships, cargo, etc) between ships that have the required attribute."

source/PlayerInfoPanel.cpp
```diff
@@ -100,7 +101,8 @@ namespace {
 
 
 PlayerInfoPanel::PlayerInfoPanel(PlayerInfo &player)
-	: player(player), canEdit(player.GetPlanet())
+	: player(player), canEdit(player.GetPlanet()
+			&& !player.GetPlanet()->Attributes().count("requires: gaslining"))
```
petervdmeer, line 105: I prefer this to be handled in a more generic way; maybe something like `(player.GetPlanet().IsUnrestricted() || player.GetPlanet().HasSpaceport())` instead of the gaslining attributes check. Such a more generic solution also works if we get things like lava-blob-planets ("requires: volcanic/lava shields").

source/Engine.cpp
```diff
@@ -266,8 +266,6 @@ void Engine::Place()
 		if(!ship->IsParked() && ship->GetSystem())
 			ships.push_back(ship);
 	
-	// Add NPCs to the list of ships. Fighters have to be assigned to carriers,
-	// and all but "uninterested" ships should follow the player.
```
petervdmeer, line 270: Is this comment now no longer valid?

tatami4:
> Question: When it says "cannot reorder ships when landed on a gas planet," does it mean we cannot reorder any ships, or just that we can't reorder ships that do not have the gaslining trait?

Cannot reorder any ships, since it doesn't make sense for the player to be able to change flagships when "landed" on a gas planet (since you cannot actually land on it).
Planets on which you cannot really land could have an attribute `no surface` or something similar to avoid hardcoding specific attributes which do not always apply only to gas planets.

> Restrictions on resorting shouldn't be tied to any one attribute, but rather a more general "if this planet requires an attribute, the player should only be able to sort (and swap flagships, cargo, etc) between ships that have the required attribute."

It would be pretty complex to allow reordering some ships but not the others. We could simply disallow changing the flagship to a ship not landed on this planet.

source/Engine.cpp
```diff
@@ -286,7 +284,7 @@ void Engine::Place()
 		planetRadius = object->Radius();
 	}
 	
-	// Give each non-carried, special ship we just added a random heading and position.
+	// Give each non-carried player's ship a random heading and position.
```
petervdmeer, line 287: This function also seems to place non-player NPC ships.

source/Engine.cpp
```diff
@@ -311,6 +309,8 @@ void Engine::Place()
 			}
 			else if(hasOwnPlanet)
 				pos = object->Position() + angle.Unit() * Random::Real() * object->Radius();
+			else
```
petervdmeer, line 312: I feel that this else needs some more conditions, since we only want this new behaviour for ships that are owned by the player or escorting the player and that cannot land with the player.

petervdmeer (changes requested):

source/PlayerInfoPanel.cpp
petervdmeer: See also Zitchas comment above...

source/Engine.cpp
tatami4: Yes, since this function only directly deals with player's ships.

source/Engine.cpp
tatami4: Look at lines 263-267, there the ship list is filled only with player's ships.

tatami4: At least I hope the list is not modified elsewhere.

tatami4: Yep, you are correct, it does add NPCs.

source/PlayerInfoPanel.cpp
tatami4: `IsUnrestricted` cannot be used here since the player cannot reorder ships only if he is "landed" on a planet on which you cannot really land (for example, a gas planet) and `IsUnrestricted` does not tell if that's the case.

tatami4: It would be better if there was a specific attribute which tells if you can actually land on a planet.

petervdmeer: Yes, such a special attribute would be even better.

source/Engine.cpp
```diff
@@ -298,7 +299,7 @@ void Engine::Place()
 		{
 			const Personality &person = ship->GetPersonality();
 			bool hasOwnPlanet = ship->GetPlanet();
-			bool launchesWithPlayer = (ship->IsYours() || planet->CanLand(*ship))
+			bool launchesWithPlayer = ((ship->IsYours() && planet->CanLand(*ship)) || planet->CanLand(*ship))
```
tehhowch, line 302: If the landable nature is always required, then it doesn't matter if the ship is yours or not:
```suggestion
			bool launchesWithPlayer = planet->CanLand(*ship)
```

source/Engine.cpp
```diff
@@ -266,12 +266,13 @@ void Engine::Place()
...
-	// Add NPCs to the list of ships. Fighters have to be assigned to carriers,
-	// and all but "uninterested" ships should follow the player.
 	shared_ptr<Ship> flagship = player.FlagshipPtr();
 	
 	// Update the active NPCs for missions based on the player's conditions.
 	player.UpdateMissionNPCs();
+	
+	// Add NPCs to the list of ships. Fighters have to be assigned to carriers,
+	// and all but "uninterested" ships should follow the player.
 	for(const Mission &mission : player.Missions())
```
tehhowch, line 276: Why move the comment?

source/PlayerInfoPanel.cpp
tehhowch, line 105: What is the purpose of this change? If the player is landed on a planet, then they're landed on a planet and can edit their fleet ordering, always. That behavior is not to be conditional.

If there is an issue where the player can take some ship that is not landed and then make that ship into their flagship and launch from it, the place to handle that is not here.

tehhowch (changes requested):
Use descriptive PR titles, please.

source/Engine.cpp
tatami4: Because the comment should be directly above the code block that it describes. I think it most accurately fits here, since only calls to `Place` actually add NPCs to the ship list.

source/PlayerInfoPanel.cpp
tatami4: I think reordering ships while "landed" on a gas planet should not be allowed since you cannot really land on a gas planet. If this is allowed, then I see no reason to disallow reordering the fleet mid-flight (except for changing the flagship, of course).

source/PlayerInfoPanel.cpp
```diff
@@ -23,6 +23,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "text/layout.hpp"
 #include "LogbookPanel.h"
 #include "MissionPanel.h"
+#include "Planet.h"
```
petervdmeer, line 26: This include no longer seems needed.

source/PlayerInfo.cpp
```diff
@@ -1137,14 +1137,14 @@ void PlayerInfo::Land(UI *ui)
 		ship->UnloadBays();
 	
 	// Ships that are landed with you on the planet should fully recharge
-	// and pool all their cargo together. Those in remote systems restore
-	// what they can without landing.
+	// and pool all their cargo together. Those that are in remote systems
+	// or cannot land on your planet restore what they can without landing.
```
petervdmeer, line 1141: What if there is another friendly planet in the system that they can land on? Wouldn't we expect our friendly ships to land there?
(Outfitters also work for ships that are parked on a different planet but in the same system.)

We can keep this for a follow-up PR.

source/Engine.cpp
tehhowch: This comment is about putting npcs into flight and assigning them the correct parent pointer, which the above code lines contribute to.

source/PlayerInfo.cpp
```diff
@@ -743,7 +743,7 @@ const shared_ptr<Ship> &PlayerInfo::FlagshipPtr()
 	if(!flagship)
 	{
 		for(const shared_ptr<Ship> &it : ships)
-			if(!it->IsParked() && it->GetSystem() == system && it->CanBeFlagship())
+			if(!it->IsParked() && it->GetSystem() == system && planet->CanLand(*it) && it->CanBeFlagship())
```
tehhowch, line 746: There's no guarantee the player is landed when this function is called, which means we would then dereference a null pointer.

source/PlayerInfo.cpp
```diff
@@ -1137,14 +1137,14 @@ void PlayerInfo::Land(UI *ui)
...
-	// what they can without landing.
+	// and pool all their cargo together. Those that are in remote systems
+	// or cannot land on your planet restore what they can without landing.
 	bool hasSpaceport = planet->HasSpaceport() && planet->CanUseServices();
 	UpdateCargoCapacities();
 	for(const shared_ptr<Ship> &ship : ships)
 		if(!ship->IsParked() && !ship->IsDisabled())
 		{
-			if(ship->GetSystem() == system)
+			if(ship->GetSystem() == system && planet->CanLand(*ship))
```
tehhowch, line 1147: A problem with this is that the player may own carrier ships that have the required attribute, but the carried ships do not. Those ships should still be treated as though they are landed on the planet.

source/PlayerInfo.cpp
tatami4: I think that in practice flagship is null only when you are landed, since the flagship appears to only be reset when you land on a planet and do stuff with ships (reorder, sell, disown, etc.).
What do you think about changing line 743 to `if(!flagship && planet)` ?

source/PlayerInfo.cpp
```diff
@@ -741,14 +741,23 @@ Ship *PlayerInfo::Flagship()
...
 	if(!flagship)
-	{
 		for(const shared_ptr<Ship> &it : ships)
 			if(!it->IsParked() && it->GetSystem() == system && it->CanBeFlagship())
 			{
-				flagship = it;
-				break;
+				if(planet)
+				{
+					if(planet->CanLand(*it))
```
petervdmeer, line 749: Not critical, but I think this `if` and the `else` below can be avoided by using `if(!planet || planet->CanLand(*it))` on line 747 (new numbering). And then it is even possible to combine the conditions it with the expression above on line 754 (new numbering).

source/PlayerInfo.cpp
```diff
@@ -1136,24 +1145,62 @@ void PlayerInfo::Land(UI *ui)
...
 	for(const shared_ptr<Ship> &ship : ships)
 		if(!ship->IsParked() && !ship->IsDisabled())
 		{
 			if(ship->GetSystem() == system)
 			{
-				ship->Recharge(hasSpaceport);
-				ship->Cargo().TransferAll(cargo);
-				if(!ship->GetPlanet())
-					ship->SetPlanet(planet);
+				if(planet->CanLand(*ship) || (ship->IsCarried() && ship->GetParent()->GetSystem() == system
```
petervdmeer, line 1158: All carried ships have been unloaded on line 1145 (new numbering), so I expect that they will not have a parent at this point.
Also, carried ships don't have a system while carried if I remember correctly, so if the ship would still have been carried, then the conditions in the if-statement on line 1156 would have returned false for such carried ships.

source/Ship.h
```diff
@@ -460,6 +462,8 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	bool isSpecial = false;
 	bool isYours = false;
 	bool isParked = false;
+	// Whether this ship is carried at the moment.
+	bool isCarried = false;
```
petervdmeer, line 466: Is it needed to have this variable? I would expect that there is already other code in the game that checks if a ship is carried and such code was then able to determine if a ships was carried without the need for this variable.

source/PlayerInfo.cpp
```diff
@@ -1136,24 +1145,62 @@ void PlayerInfo::Land(UI *ui)
...
+						&& planet->CanLand(*ship->GetParent())))
+				{
+					ship->Recharge(hasSpaceport);
+					ship->Cargo().TransferAll(cargo);
+					if(!ship->GetPlanet())
+						ship->SetPlanet(planet);
+				}
+				// Ships that cannot land with the flagship choose the most suitable planet
+				// in the system.
```
petervdmeer, line 1167: This code looks similar to the code that searches for planets for refueling. Are the pieces of code sufficiently similar that it would make sense to go to a common search function that is used in both situations?

petervdmeer:
This PR by itself looks quite okay, but it has a merge-conflict that needs to be resolved and 1 open review comment (since January).
Moving this PR to draft so that the author can fix those two and then move the PR to non-draft again.

Amazinite:
Ditto reasoning from #5654.

Amazinite:
Closing, as #8256 picks this up.

PR #5682, opened 2021-02-01
Fix: Missing link in Remnant: Postverta Reveal event

### LINKED ISSUE #5680: Link from Statina to Egeria is not saved.
**Describe the bug**
In the player's save file there is no link from Statina to Egeria even though the player has visited both systems.

**To Reproduce**
don't know, I just saw it when testing one of my PRs (probably happens every time as I have it on two saves)

**Expected behavior**
There is a link from Statina to Egeria.

 - OS: openSUSE Linux
 - Game Source: self-compiled
 - Version: eb1af7d47351fb5bc940664397e85329a1e25ba0

**Link to save file**
[John Brown.txt](https://github.com/endless-sky/endless-sky/files/5903296/John.Brown.txt)
relevant section (from line 12014):
```	
system Statina
	link Postverta
	link Diespiter
```

this one has not visited Statina yet: 
[Stuck.Remnant.txt](https://github.com/endless-sky/endless-sky/files/5903301/Stuck.Remnant.txt)
relevant section (from line 40657):
```	
system Statina
	link Postverta
	link Diespiter
```

Zitchas:
**Bugfix:** This PR addresses issue #5680 

## Bug Details
The `Remnant: Postverta Reveal` event specifies links for several systems in the cluster. However, this specification replaces the system's old links instead of adding to them. In this context, the event is missing one link. It is worth noting that Egeria's links are not modified by the event, and it already has a link to Statina. Thus after the event it retains the link, but Statina loses it. This, in effect, makes the link a one-way connection that will work from Egeria to Statina (and thus be visible on the map), while not functional in the reverse.

This is actually kind of a neat and fitting behavior... If there were a number of instances of it so that the player could tell that it was intentional. A one-off instance like this looks like the error that it is. A good reminder for the future, however, that one-way links are possible and that the Ember Waste would be a good place to use them; but there needs to be sufficient of them that the player can tell that they are intentional. Like the worm-hole train is clearly intentional, for instance.

## Fix Details

This adds the missing link to the event, and moves the postverta link to the end of the list so that they are in proper alphabetic order.

fixes #5680 

## Save File
This is a revised version of the file used to demonstrate the bug in 5680. It has had the relevant section of the Cognizance missions removed so they can be replayed, and thus the fix demonstrated.
[John.Brown.txt](https://github.com/endless-sky/endless-sky/files/5904334/John.Brown.txt)

Thanks to tatami4 for spotting this bug!


data/remnant/remnant missions.txt
```diff
@@ -3951,35 +3951,30 @@ event "remnant: postverta reveal"
...
 			period 389
+	system Levana
+		add object "wormhole barren alpha"
 			sprite planet/wormhole-red
-	system Egeria
-		add object wormhole_barren
 			distance 2313
 			period 389
+	system Prosa
+		add object "wormhole barren beta"
```
Amazinite, line 3970: You'll need to rename the planet definitions in map.txt as well. And object names should be capitalized; "Wormhole Barren Beta," not "wormhole barren beta."

Zitchas: Thanks for the reminder. Done.

data/map.txt
```diff
@@ -29808,11 +29808,11 @@ planet Zug
...
+planet "Wormhole Barren Alpha"
 	attributes "requires: quantum keystone"
 	description ``
-planet wormhole_barren2
+planet "Wormhole Barren Beta"
 	attributes "requires: quantum keystone"
 	description ``
-planet wormhole_link
+planet "Wormhole Link"
 	description ``
```
Amazinite, line 29818: Red wormholes should mean that a stone is required for landing, yet this wormhole doesn't require one despite being red.

data/remnant/remnant missions.txt
```diff
@@ -3946,40 +3946,35 @@ event "remnant: postverta reveal"
...
+		link Statina
+	system Prosa
+		add object "Wormhole Barren Beta"
 			sprite planet/wormhole-red
-		
+			distance 2313
+			period 389
+	link Postverta Diespiter
+	link Postverta Prosa
+	link Postverta Statina
```
Amazinite, line 3976: I still think you should have the systems linked from game start and have Postverta's starting position be a little more up and to the right so that it's out of jump range at game start. You would still move Postverta to `pos 297 269` with this event, though. 

I don't see unvisiting the systems as necessary, since they can't be visited in the first place before this event, but if you do what I suggested then it might make sense to unvisit Postverta as well.

Amazinite (changes requested):
Should we move all the systems to the region I suggested in this PR? I can provide position that looks good to me along with a screenshot on how it looks.

data/map.txt
Zitchas: Thanks. Will fix this.

PR #5683, opened 2021-02-01
Conflicting keybindings are saved

### LINKED ISSUE #5582: Don't reset multiple keybinds that are overloaded to a single key

tatami4:
Conflicting keybindings are saved to player's save file instead of reverted to default.
This is a reopening of #5611 with red keybinding conflict color changed to yellowish and the comment corrected.
Fixes #5582.

### DIFF
--- a/source/PreferencesPanel.cpp
+++ b/source/PreferencesPanel.cpp
@@ -296,7 +296,7 @@ void PreferencesPanel::DrawControls()
 	const Color &bright = *GameData::Colors().Get("bright");
 	
 	// Check for conflicts.
-	Color warning(.5f, .3f, 0.f, 0.f);
+	const Color &warning = *GameData::Colors().Get("warning conflict");
 	
 	Table table;
 	table.AddColumn(-115, {230, Alignment::LEFT});
--- a/source/PreferencesPanel.cpp
+++ b/source/PreferencesPanel.cpp
@@ -296,7 +296,7 @@ void PreferencesPanel::DrawControls()
 	const Color &bright = *GameData::Colors().Get("bright");
 	
 	// Check for conflicts.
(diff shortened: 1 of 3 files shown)


source/PreferencesPanel.cpp
```diff
@@ -296,7 +296,7 @@ void PreferencesPanel::DrawControls()
 	const Color &bright = *GameData::Colors().Get("bright");
 	
 	// Check for conflicts.
-	Color red(.3f, 0.f, 0.f, .3f);
+	Color warning(.5f, .3f, 0.f, 0.f);
```
Amazinite, line 299: This is a pretty harsh yellow compared to the duller red we had before. 
image
image
How about (.315f .27f .12f 1.f) instead?
image
And while we're here, how about defining this color in interfaces.txt like the other colors above? (Speaking of the interface file, the suggested color is actually a 50% brighter version of the "warning back" color already defined there.)

Amazinite (changes requested):
Looks good. Overloaded controls still work and save after quitting the game instead of being overwritten. Only issue is that the color is way too harsh.

Issue #4124, opened 2019-01-15
A fleet jump command is desperately needed


hrs111111:
This is essentially reopening https://github.com/endless-sky/endless-sky/issues/1026

Problem:
Manually long pressing J for multiple seconds, and manually waiting until all ships align to release J, in every system is very tiring, and takes much fun out of the game. A player should instead be able to autopilot to the destination system as long as no threats spawn on the route.

Solution:
Have a key/command that you can press just once to do automatic fleet jumps until the fleet reaches the destination system. A fleet jump is a jump where all ships wait until the slowest ship has aligned and is ready and able to jump, and then jumps together as a single group.

MCOfficer:
If it wasn't on the other end of the keyboard, i'd make a case for shift-j or ctrl-j. Otherwise: K?

jafdy:
This falls under the umbrella of too much automation. Very simply it would allow you to walk away from the game and when you return it has got to the next planet without any input from you.

hrs111111:
Why would it be a bad thing? Anything that is not engaging to the player should be automated away!
(Isn't it the same principle when pressing J to use autopilot when you have only one ship.)

hrs111111:
As a counter point, currently a player has to hold down J for 5 - 10 seconds at each jump, which is very taxing. At the very least, holding J down can be replaced with a single key press.

Secondly, https://github.com/endless-sky/endless-sky/pull/2648 changes that would have been merged in except for merge conflicts and minor issues, If I understand correctly.

Pointedstick:
> This falls under the umbrella of too much automation. Very simply it would allow you to walk away from the game and when you return it has got to the next planet without any input from you.

That's exactly what we already have when you're in a single ship: open map, click on planet, close map, hit `J`, go get a cup of coffee. I don't see why this should be acceptable when you're alone, but unacceptable when you have a fleet.

jafdy:
I don't think it is. I liked the old way where you at the very least needed to be there for the last system.

Pointedstick:
The point is, there's no logical reason for it to be inconsistent between single ship vs fleet. If we want to reduce the automation factor, it should be a purposeful design decision, not the side effect of a bug or oversight.

AlbertNewton:
Taking a while to cross the galaxy only really happens if you're in a slow ship with a normal hyperdrive. And even still, unless you have huge fuel capacity you're going to need to manually land and take off again. Not much chance of a coffee there, Pointedstick.

[...]

PR #5686, opened 2021-02-03
Add fleet jump functionality

yjhn:
**Feature:** This PR implements the feature request detailed and discussed in issue #4124.

## Feature Details
Fleet jump is a command that is given to the player's flagship. The flagship then waits for all escorts to get ready before jumping. Command is activated using shift+Jump.

## UI Screenshots
None, should add settings UI box explaining fleet jump.

## Testing Done
jumped using fleet jump
initiated fleet jump, then jumped using regular jump
initiated fleet jump, cancelled it
initiated normal jump, then changed it to fleet jump

## Performance Impact
When fleet jump is used, every frame fleet readiness to jump is checked.


petervdmeer:
> This is currently implemented in a rather hacky way. It could be implemented as a normal command, but it would probably require more code changes.

Nice to see this feature getting implemented.

I had a look at the code and I really prefer to have this implemented through an (additional) normal command (on the flagship). Implementing this in a non-hacky way is probably a little bit more work, but it really helps for future changes to keep the codebase well-structured and well-readable.

tatami4:
Fleet jump is now a command.

keys.txt
```diff
@@ -20,6 +20,7 @@
 "View player info" 105
 "Toggle fullscreen" 1073741892
 "Toggle fast-forward" 1073741881
+"Fleet: Hyperspace jump" 107
```
petervdmeer, line 23: I'm happy that fleet jump is now a separate command, but I would really like it to be triggered using shift+J instead of a separate key.

tatami4: I see two issues with this approach:
1. It would no longer be independently customizable
2. Shift and J are quite far apart (at least for me, since I use right Shift rarely)

I wouldn't want to do this.

petervdmeer: For me it feels logical that shift+J is used for fleet-jump if the J-key is used for jumping, similar to shift-R(select nearest ship) relates to the R-key(select nearest hostile).
I would expect the fleet-jump command do be given at rather peaceful moments, so that it is a bit more keypresses than regular J is also fine for me.

Another option is to remove shift completely and do something like this:
- Pressing J starts a normal jump (as is done today).
- Pressing J and waiting for the fleet to jump activates fleet-jump. (keep the effect of waiting not just for the first jump, but propagate it).

petervdmeer: The game uses quite a number of buttons already, I prefer to keep all jump-related actions under 1 (configurable) key.

tatami4:
Fleet jump is now activated using shift+Jump. The problem - UI does not show it.

petervdmeer:
Merged master, fixed some compile and merge-issues and briefly retested with Jump, Fleet jump and Fleet Jump for "Jump Drive". Seems to work as advertised.

Zitchas:
Awesome! Looking forward to trying this out. Informing the players, though, seems problematic.

petervdmeer:
> Awesome! Looking forward to trying this out. Informing the players, though, seems problematic.

Adding some hints, or updating the preferences section with something like this (but better aligned and also with an 2x image) would indeed be good:
preferences_section

yjhn:
> Adding some hints, or updating the preferences section with something like this (but better aligned and also with an 2x image) would indeed be good

I wanted to do this, but couldn't figure out how to create a matching image (with those fancy ornaments on the sides and changing side colours). It's really unfortunate that all game menu UI is based on images and those images are not easy to create. This makes it really hard to do any UI changes :frowning_face:.

Issue #5556, opened 2020-12-14
Create a gamerules file that allows for the easy manipulation of various currently-hardcoded values


Amazinite:
**Is your feature request related to a problem? Please describe.**
There are currently a number of hardcoded values that it might be interested to be able to fiddle with, such as as the values that dictate depreciation. There are also mechanics that have been removed in the past that some players might like to have in their game, while others might not. Death benefits comes to mind as one such removed mechanic. Ideally, instead of completely removing such mechanics, it might be better to have them be able to be toggled on and off.

**Related Issue Links**
#2200 attempted to create a file similar to this.
#5554 includes an example of the type of values that one might want to change in a data file.
#1386 Having customizable gamerules would allow for creating different difficulty levels.
#5512 Such gamerules would ideally be able to be modified by start conditions so that changing difficulties doesn't require going into the data files to switch things around.

**Describe the solution you'd like**
Create a class similar to preferences named gamerules. Gamerules would contain a `map<string, double>` where the string is the name of the gamerule and the double is the value. Gamerules would be statically accessed, as preferences is now, whenever it is needed. Using a double for the value means that we can have gamerules that do require doubles, gamerules that require ints (cast the double to and int), and gamerules that require bools (take 0 as false and anything else as true).

What this would not include is the customization of strings that are currently hardcoded, such as outfit and ship categories.

**Describe alternatives you've considered**
Do what #5555 does and make use of the interface class.

**Additional context**
Potential gamerules:
* Basic stuff:
  * Depreciation constants. (#2200)
  * Debris lifespan.
  * Author spawn time.
  * Base hit force relative to mass.
  * Base ramscoop value.
* More advanced stuff (e.g. AI changes):
  * The weights that AI give to various targets? (AI::FindTarget) 
  * Changing the amount of health a fighter has left before it retreats to its carrier.
  * The system fence distance. 
  * Default jump/hyperdrive arrival distance.
    * Also default departure distance (when it becomes a thing).
  * Constants dictating the strength of ramscoops/solar panels given the distance from the system center. (min power, max power, distance where strength = 1)
  * Cooling inefficiency constants.
  * Loan constants (for the player taking loans from the bank).
  * Default constants for the minimum hull equation (when a ship becomes disabled).
  * The range at which missile boat behavior is used.
* Toggle mechanics (stuff that use to exist or has been requested and denied):
  * Whether death benefits are a thing. (Removed mechanic)
[...]

PR #5689, opened 2021-02-04
Greatly increase flotsam lifetimes and add drag

Amazinite:
**Feature:** This PR implements a feature that's been talked about for a while now.

## Feature Details
Flotsams are outfits or commodities that have been jettisoned from ships or dropped by destroyed minable asteroids. These flotsams have been criticized as dying too quickly, and in the case of asteroids their flotsams have been criticized as moving too quickly to pick up.

This PR does two things:
1) Greatly increases the lifetime of flotsams. Commodity flotsams use to last anywhere from 6 to 11 seconds for a full box. Full commodity boxes now last between 2 and 3 minutes. Outfit flotsams on the other hand last 10 times longer than they use to. For example, ore flotsams use to last about 8 seconds. They now last about 80. (Flotsam lifetimes range from about 80 to 85 seconds on average depending on the type of minable, with Yottrite being the only suggested at about 65 seconds on average.) And in terms of more expensive outfits, a Triple Plasma Core that was jettisoned used to last 1.35 seconds. It now lasts about 13.5 seconds.
2) Added drag to flotsams. Flotsams now lose .1% of their velocity every frame. This means that after about 11 seconds, flotsams will have dropped to half their original velocity.

Questions:
* Should the loss in velocity be decreased? Longer lived flotsams like ores and commodities will always come to a stop with the current value.
* Should the loss in velocity change depending on the flotsam? For example, should all flotsams die after reaching say 25% velocity, meaning flotsams that are worth more slow down quicker?

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
Went mining for ores. Flotsams from minables last much longer than they use to, and not picking them up allows them to slowly come to a stop.
Jettisoned some commodities and observed that they lasted even longer than the minables did.
Jettisoned a Triple Plasma Core and observed that it lasted much shorter than the commodities or ores.

## Performance Impact
None.

### DIFF
--- a/source/Flotsam.cpp
+++ b/source/Flotsam.cpp
@@ -95,7 +95,7 @@ void Flotsam::Place(const Body &source, const Point &dv)
 void Flotsam::Move(vector<Visual> &visuals)
 {
 	position += velocity;
-	velocity *= 0.999;
+	velocity *= drag;
 	angle += spin;
 	--lifetime;
 	if(lifetime > 0)
--- a/source/Flotsam.cpp
+++ b/source/Flotsam.cpp
@@ -35,7 +35,7 @@ const int Flotsam::TONS_PER_BOX = 5;
 Flotsam::Flotsam(const string &commodity, int count)
 	: commodity(commodity), count(count)
 {
-	lifetime = Random::Int(300) + 360;
+	lifetime = Random::Int(3600) + 7200;
 	// Scale lifetime in proportion to the expected amount per box.
 	if(count != TONS_PER_BOX)
 		lifetime = sqrt(count * (1. / TONS_PER_BOX)) * lifetime;
@@ -47,8 +47,8 @@ Flotsam::Flotsam(const Outfit *outfit, int count)
 	: outfit(outfit), count(count)
 {
 	// The more the outfit costs, the faster this flotsam should disappear.
-	int lifetimeBase = 300000000 / (outfit->Cost() * count + 1000000);
-	lifetime = Random::Int(lifetimeBase) + lifetimeBase + 60;
+	int lifetimeBase = 3000000000 / (outfit->Cost() * count + 1000000);
+	lifetime = Random::Int(lifetimeBase) + lifetimeBase + 600;
 }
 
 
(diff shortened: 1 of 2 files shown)


MasterOfGrey:
To answer the questions and thereby carry over my discord comments for the record, I will apply two answers to each. The ideal case that I think would be perfect (and do not have any real expectation of seeing), and my opinion on the practical case I think would have low impact on the proposed changes:
- Should the loss in velocity be decreased? Longer lived flotsams like ores and commodities will always come to a stop with the current value.
  - Ideally: I think that there should be a lower limit where drag stops applying, and that at all times there should be a vector applied to simulate the effect of it moving through the gravity well (either with an actual constant vector applied towards the sun or with a fudged vector to make it circle). I think that particularly for longer-lived ones like commodities this would look really good, and also you'd be able to get moving debris fields that all carry some echo of what the ship that was destroyed for them to exist was doing at the time. (Would look way cool in like Wah Ki, Cardax, or Automata space).
  - Practically: I think those things coming to a stop eventually isn't really a problem, and the current implementation works perfectly well with the 'space is a fluid' behaviour of ships in ES.
- Should the loss in velocity change depending on the flotsam? For example, should all flotsams die after reaching say 25% velocity, meaning flotsams that are worth more slow down quicker?
  - Ideally: If it could use the mass of the flotsam as an input to determining drag that would be really cool and fit in especially well with the "ideally" for the previous question.
    - Please god no lets not tie velocity to _value_ though. I will cry.
  - Practically: It's not at all necessary, it'd really just be a nice-to-have, but I think just using a different fixed drag value for each category of commodities/minables/outfits would at least allow a sharp-eyed player to prioritise what to chase if they're picking up from a destroyed ship.

Zitchas:
Longer lived flotsam? Awesome! 

Having generic mineables last almost forever makes sense to me. Asteroids have been around for millions of years. Doesn't make sense that something blasted off them would suddenly melt away into space in mere seconds or minutes.

On the other hand, having things like, well, reactors blow up or vanish or whatnot after a short while also makes sense. I guess it'll come down to playtesting to see how it feels. I don't think raw numbers really convey the right feeling. 

That being said, having everything last longer both makes sense, sounds good, and is something I've wanted for ages now, so, thanks!

Amazinite:
Well #5556 is one of the things I want to work on for the next major release, so if these values are too high then we should be able to more easily tweak them later instead of needing to always change the source code and recompile.

PR #5697, opened 2021-02-06
Lower amount of missile Furies in small southern pirates

Terin:
Currently, a small southern pirate fleet can spawn anything from a single Sparrow, to three Furies with two Meteor Missile Launchers each. While there should be some variety, this is too much for the lowest end, and is reduced to two ships in this PR.

### DIFF
--- a/data/human/fleets.txt
+++ b/data/human/fleets.txt
@@ -1697,7 +1697,7 @@ fleet "Small Southern Pirates"
 	variant 1
 		"Fury" 3
 	variant 1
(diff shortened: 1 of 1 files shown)


Amazinite:
I will note that I've called out that there are issues with pirate fleet strengths in the past. #4158
I plan to eventually take a more comprehensive look at all the pirate fleets and fleet variants to balance them out more, but no harm in taking baby steps now. Missile spam variants are definitely an issue to deal with.

PR #5698, opened 2021-02-06
feat(mechanics): Add four new fleet personalities

InfiniteDonuts:
## Summary
@Galaucus (aka Lake from the Discord server) asked me to help him with his Embershade life-form (PR is #5933) by creating three new personalities: `roving`, `evasive`, and `hiding`. Having been seeking a programming task for a while, I happily agreed!

Here's a brief rundown of the personalities I intend to add:
- `roving`: ships will just wander aimlessly around the system, not leaving. Outlined in #5864.
- `evasive`: ships will stay out of weapons range of any enemies and won't engage them. Outlined in #5866.
- `cloaking`: ships will cloak regardless of fuel cost. Outlined in #5865.
- `bored`: ships with this personality will have a 1/6000 chance of despawning. They will prioritize hyperspacing away, but if they cannot, they will land instead.

## PR Checklist

- [x] `roving`
- [x] `cloaking`
- [x] `evasive`
- [x] `bored`


Zitchas:
Just to clarify a bit: 
**Roving**
Sounds good. Perhaps just leave off the part about attacking. Probably best if this personality just focuses on changing one thing and leaves the rest to the default AI and other personality traits. And the core thing here is `"ships will just wander aimlessly around the system, not leaving."` So I'd just shorten it to that. 

In practice, the goal of "roving" is that ships with this personality will wander around the whole system, irregardless of the location of stars, planets, or other ships; and that it will generally ignore wormholes and hyperjumping.

**Shy**
Shy currently sounds like a mix of normal behaviour and pacifist. I think what we actually are aiming for with that personality is `"Stays outside of weapons range of **all** ships that aren't part of its fleet."`

The point is, after all, having a personality that stays away from everything except members of its own group. This would keep it away from everyone, and make it act like "packs." Again, I'd take off the part about combat, since the core shy behaviour is unique here, but the combat part is something we can easily get by tacking on "pacifist." Having it just be the core behavior means we can have reclusive ships that avoid everyone else - but will still press home the attack when they get provoked. Could be useful behaviour to have.

**Elusive**
A behavior that pushes the NPC to cloak whenever possible. Simple, and tot the point. Question, though: the Archons currently do this. What makes them do it? Is this duplicating something?

beccabunny:
> A behavior that pushes the NPC to cloak whenever possible. Simple, and tot the point. Question, though: the Archons currently do this. What makes them do it? Is this duplicating something

Everything cloaks whenever possible when no enemies are present, no matter the personality

Zitchas:
Ah, so the key detail is that it continues to try to cloak even when enemies are present.

Galaucus:
> Ah, so the key detail is that it continues to try to cloak even when enemies are present.

My intention for elusive is that the ship prioritizes breaking off of combat and entering cloak when made aggressive. It will only fight when cloaking is impossible and it's been provoked. As Embershades have no weapons this should mean that they should always attempt to flee and cloak no matter the circumstances.

InfiniteDonuts:
Correct. An elusive AI will disregard fuel expenses and enemies nearby and choose to remain cloaked as much as possible.

tehhowch:
My concerns from the first PR are still applicable and need addressing:

If `personality elusive` just means "cloak whenever possible," then "elusive" is a bad name for the desired functionality. If the goal is a way for a content creator to specify "ships with this flag always cloak whenever possible", that personality token should be named appropriately, e.g. "cloaking" or "always cloaking" or "hiding".  
Further, if that is the only behavior of the personality, then the personality trait should be checked during normal parts of the cloaking code, i.e. `DoCloak`, rather than in its own section.

I have similar concerns about "shy," because "shy" and "timid" are synonyms. It's almost certain one will be used when the behavior of the other was intended.

>Shy: The ship attempts to avoid non-friendly vessels, retreating to a safe minimum distance if any come too close.

"evasive" would fit better, IMO. Note that as described, this personality will have no impact on movement unless the ship encounters hostiles in the system (merchant ships and the player ships are friendly, hai and merchants are friendly, etc). What's the expected behavior of shy/evasive ships w.r.t. to hostile ships? If it's "fight to survive," then this sounds like a personality flag that has no actual behavior of its own, i.e. one that we do not need at all (because it does nothing). If it's "don't fight," then that's the "pacifist" or "forebearing" flag. If it is "run away," that's the "fleeing" flag.

Zitchas:
Tehhowch's comments regarding evasive illustrate my point, I think. We don't want them to simply avoid hostiles, we want them to avoid *everyone that isn't a part of their fleet.* I'm not positive, but I think fleets are treated as a unit in terms of personality, right? So ships in an NPC fleet are never going to avoid each other, am I correct? In which case, `Evasive`'s description could be simplified to "tries to stay away from everyone."

InfiniteDonuts:
Update: changed the names of shy and elusive to evasive and hiding, respectively.

InfiniteDonuts:
Update: All three personalities *should* be done now. Key word *should*. Tweaking is still required, and it kind of looks like my latest commit (fixing the object-copying problem) may have broken `evasive`.

Galaucus:
We're going to need some way to entice the critters to eventually jump out of the system instead of infinitely roaming - right now they just pile up endlessly because they have ways in to the system, but no way out. Ideally the jump in/jump out frequency would be either equal or slightly skewed in favor of jumping out, so as to make encounters more rare.

InfiniteDonuts:
Also, there's an issue with the embershades roving inside an enemy's range, I'll get to work on fixing that stuff.

arkhne:
I just wanted to add some encouragement here, as I'd love to use them for #5847 (and other plans).

tehhowch:
De-milestoning the PR, because the PR may or may not be what ships the functionality. I'll review it later, but my current takeaway is that the tags are being developed in a too-coupled state. Each personality tag needs to have a distinct purpose _on its own_ that doesn't already exist. Each proposed tag should be fleshed out in its own issue or RFC that proposes how that tag will interact when combined with the other existing tags. Those issues / RFCs would then be milestoned instead of one implementation.

InfiniteDonuts:
For now, I've hardcoded a 1/3000 chance of leaving into the roving personality. I can still abstract it into a new one if you guys like.

InfiniteDonuts:
I just created the Bored personality, which abstracts away the 'chance of leaving per frame' functionality.

Pointedstick:
Nitpick that you can feel to ignore: Could we rename "hiding" to something less generic? I could see people mistakenly apply it to ships without cloaking abilities. Maybe just something as simple as "cloaking".

InfiniteDonuts:
@Pointedstick Done.

Pointedstick:
Cool, thanks! Do you have permission to edit https://github.com/endless-sky/endless-sky/wiki/ShipPersonalities to document the new personalities if and when this is merged?

InfiniteDonuts:
> 
> 
> Cool, thanks! Do you have permission to edit https://github.com/endless-sky/endless-sky/wiki/ShipPersonalities to document the new personalities if and when this is merged?

No, I don't think so

Pointedstick:
OK, I'll do that once this is merged.

tehhowch:
This is not ready to merge, btw

source/AI.cpp
```diff
@@ -705,6 +711,39 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 			targetDistance = target->Position().Distance(it->Position());
 		
 		// Behave in accordance with personality traits.
+		if(isPresent && personality.IsCloaking() && !isStranded)
+		{
```
quyykk, line 715: Drop the braces for single line statements.

source/AI.cpp
```diff
@@ -520,6 +520,12 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 			value = min(FENCE_MAX, value + FENCE_DECAY + 1);
 		}
 	
+	// How many roving ships there are in the system.
+	unsigned int rovingShipCount = 0;
+	for(const auto &ship: ships)
```
quyykk, line 525: Iterating through every ship *again* is less than ideal. But you're in luck: the for statement a few lines above *also* iterates through every ship. Fold them please. :)

source/AI.cpp
```diff
@@ -705,6 +711,39 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
...
+				Refuel(*it, command);
+			}
+			
+			it->SetCommands(command);
+			continue;
+		}
+		
+		// Evasive ships will run away from other ships
+		if(isPresent && personality.IsEvasive() && !isStranded)
+		{
```
quyykk, line 739: drop braces here too and above and elsewhere.

source/AI.cpp
```diff
@@ -1089,7 +1146,7 @@ bool AI::HasHelper(const Ship &ship, const bool needsFuel)
 
 // Pick a new target for the given ship.
 shared_ptr<Ship> AI::FindTarget(const Ship &ship) const
-{
+{
```
quyykk, line 1149: accidental change

source/AI.cpp
```diff
@@ -2573,6 +2630,107 @@ bool AI::DoCloak(Ship &ship, Command &command)
 
 
 
+bool AI::DoRoving(Ship &ship, Command &command, unsigned int rovingShipCount)
+{
+	const System* system = ship.GetSystem();
+	if((rovingShipCount > 9 && system))
```
quyykk, line 2636: drop the extra parens.

source/AI.cpp
```diff
@@ -2573,6 +2630,107 @@ bool AI::DoCloak(Ship &ship, Command &command)
 
 
 
+bool AI::DoRoving(Ship &ship, Command &command, unsigned int rovingShipCount)
+{
+	const System* system = ship.GetSystem();
```
quyykk, line 2635: `System *system`

source/AI.cpp
```diff
@@ -2573,6 +2630,107 @@ bool AI::DoCloak(Ship &ship, Command &command)
...
+		}
+		else if (system->IsInhabited(&ship))
+		{
+			for(const StellarObject& planet: system->Objects())
+				if(CanRefuel(ship, &planet))
+				{
+					ship.SetTargetStellar(&planet);
+				}
+			Refuel(ship, command);
+			return true;
```
quyykk, line 2651: This is doing the same thing as the bored personality. Maybe refactor this into another function?

source/AI.cpp
```diff
@@ -2573,6 +2630,107 @@ bool AI::DoCloak(Ship &ship, Command &command)
...
+
+
+bool AI::DoEvasive(Ship &ship, Command &command)
+{
+	// Did I have to move my targetPosition?
+	bool movedTarget = false;
+	
+	// First, figure out which ships are not of the same government. Ignore disabled/destroyed ships.
+	vector<shared_ptr<Ship>> enemies;
+	for(const auto &otherShip : GetShipsList(ship, false))
```
quyykk, line 2675: Would it be useful to limit the max range where enemies can be that are considered?

source/HailPanel.cpp
```diff
@@ -72,7 +72,7 @@ HailPanel::HailPanel(PlayerInfo &player, const shared_ptr<Ship> &ship)
 	else if(ship->IsDisabled())
 	{
 		const Ship *flagship = player.Flagship();
-		if(!flagship->JumpsRemaining() || flagship->IsDisabled())
```
quyykk, line 75: Why this change?

source/Personality.cpp
```diff
@@ -49,7 +49,11 @@ namespace {
 	const int TARGET = (1 << 25);
 	const int MARKED = (1 << 26);
 	const int LAUNCHING = (1 << 27);
-	
+	const int ROVING = (1 << 28);
+	const int CLOAKING = (1 << 29);
+	const int EVASIVE = (1 << 30);
+	const int BORED = (1 << 31);
```
quyykk, line 55: Seems like we're hitting the limit of personalities. interesting

InfiniteDonuts: Yeah, Derpy mentioned that in one of his streams. Might wanna change that to a `long` instead of just an `int`?

source/HailPanel.cpp
InfiniteDonuts: That might be from when I was trying to make evasive ships mute... I'll fix it

source/AI.cpp
InfiniteDonuts: Probably not, because it's dependent on enemy weapons range, not just proximity.

InfiniteDonuts:
@quyykk Should have all of those nitpicking things finished, but I'm afraid to resolve them in case I didn't

Pointedstick:
> This is not ready to merge, btw

I'm aware. :)

source/AI.cpp
```diff
@@ -2631,35 +2618,19 @@ bool AI::DoCloak(Ship &ship, Command &command)
...
+	if(!target || MoveTo(ship, command, target, Point(), v, v))
     {
         Point newTarget = Angle::Random().Unit() * Random::Real() * MAX_DISTANCE_FROM_CENTER;
         ship.SetTargetPosition(newTarget);
     }
     
+    const System *system = ship.GetSystem();
+	if(rovingShipCount > 9 && system)
+		return DoBored(ship, command, system);
+
```
quyykk: You forgot to fix the indentation here. Also, you can resolve all my comments :)

InfiniteDonuts: Hm.

InfiniteDonuts:
Idk why, but my code got an error after I resolved the merge conflicts (probably something that changed and then was caught when the tests re-ran). Should be good to go now, but it would probably be a good idea to review.

source/Personality.h
```diff
@@ -63,6 +63,9 @@ class Personality {
 	bool IsMining() const;
 	bool Harvests() const;
 	bool IsSwarming() const;
+	bool IsRoving() const;
+	bool IsCloaking() const;
+	bool IsBored() const;
```
Hurleveur, line 68: IsBored() should be at the same location in the cpp as here; but its after IsEvasive

source/AI.cpp
```diff
@@ -2594,6 +2639,113 @@ bool AI::DoCloak(Ship &ship, Command &command)
...
+	for(const auto &enemy : enemies)
+	{
+		// Ignore disabled/destroyed ships.
+		if(enemy->IsDisabled() || enemy->IsDestroyed())
+			continue;
+		
+		// Find out their weapons range.
+		for(const auto &weapon : enemy->Weapons())
+			maxRange = max(maxRange, weapon.GetOutfit()->Range());
+		maxRange *= 2;
```
Hurleveur, line 2685: why do we multiply it by 2? I think multiplying it by 1.5 would be reasonable, or maybe we should instead just add what we consider to be a safe distance to it; like idk, say 500

Hurleveur (changes requested):
Looking good overhaul, just some small things that could be improved.

Also I'm puzzled by the limit of roving ships in any given system; it doesn't correspond to the PR's description saying `ships will just wander aimlessly around the system, not leaving`.

I would like if they wouldn't just stay in the same system except if they have the Staying personality, there is no point in limiting the possibilities offered by the personality any more than we have to. And if a content creator wants 10 ships wandering around the system without moving well just let him.

Just a thought, should there be a warning when you choose personalities that don't work together? Doesn't have to be in this PR but for instance I don't see staying going with bored; or evasive with swarming, and it could lead to weird behaviours.

source/AI.cpp
InfiniteDonuts: I played around with the numbers and found that if the range is multiplied by 2, they have enough time to turn around and run. I tried 1.5 and 1.2, both were too slow.

Hurleveur (commented):
I will do a more serious review in the future probably but being in vacation rn I'll just give the suggestion I had already pending

source/AI.cpp
Hurleveur: Could you add a comment explaining it?
What I was suggesting was defining a safe distance (say 500) and adding it to the range, because rn if a ship has huge range your ship will go too far, and if it uses firethrower you'll be too close.

source/AI.cpp
```diff
@@ -780,6 +807,15 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 				it->SetTargetAsteroid(nullptr);
 		}
 
+		// Ships with the Roving personality should aimlessly scoot about the system.
+		if(isPresent && personality.IsRoving() && !target && !isStranded)
```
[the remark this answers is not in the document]
Zitchas: This is sort of the key thing in ensuring large numbers of ships don't build up in the system. We can't always be sure that the ships have an HD. Although that being said, I would very much like to have a Max value just on flowers in general, but that is probably a topic for another PR.

This kind of reveals a problem with the current fleet system: we want a very high chance of having the specific entities in the systems they are in, usually with several of them - but never a lot of them.

source/AI.cpp
```diff
@@ -2594,6 +2639,113 @@ bool AI::DoCloak(Ship &ship, Command &command)
...
+	const Point target = ship.GetTargetPosition();
+	const auto v = ship.MaxVelocity();
+	if(!target || MoveTo(ship, command, target, Point(), v, v))
+	{
+		Point newTarget = Angle::Random().Unit() * Random::Real() * MAX_DISTANCE_FROM_CENTER;
+		ship.SetTargetPosition(newTarget);
+	}
+	
+	const System *system = ship.GetSystem();
+	if(rovingShipCount > 9 && system)
```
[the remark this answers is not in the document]
Hurleveur: I mostly have a problem with this because the number is hardcoded and I cant find any logic behind it
Would be meat to idk define it by government with a default of 9 for instance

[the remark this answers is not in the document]
Zitchas: That is a good idea. Perhaps even better would be being able to specify the number of ships in the fleet definition.

source/AI.cpp
```diff
@@ -2594,6 +2639,113 @@ bool AI::DoCloak(Ship &ship, Command &command)
...
+			enemies.push_back(make_shared<Ship>(*otherShip));
+	
+	// Loop through enemies and figure out their range and how to evade them.
+	vector<Point> paths;
+	double maxRange = 0.;
+	const Point pos = ship.Position();
+	for(const auto &enemy : enemies)
+	{
+		// Ignore disabled/destroyed ships.
+		if(enemy->IsDisabled() || enemy->IsDestroyed())
```
Hurleveur, line 2679:
```suggestion
	// Loop through enemies and figure out their range and how to evade them.
	vector<Point> paths;
	double maxRange = 0.;
	const Point pos = ship.Position();
	for(const auto &otherShip : GetShipsList(ship, false))
		if(otherShip->GetGovernment() == ship.GetGovernment() ||
				(enemy->IsDestroyed() || enemy->IsDisabled()))
			continue
```
We don't need the vector

Hurleveur:
It would be nice to have the number of roving ships dependant on a government attribute, because I imagine it should be quite different for the remnants or the Korath, for instance.
In order to do it you should really inspire yourself of how its done for the mining personality, near line 737 of AI.cpp

source/AI.cpp
InfiniteDonuts: I tried adding a few numbers to the range like you suggested, so far adding 1,000 seems to give them enough time to get out of the way. I'll push the changes once I resolve the other suggestions you made.

InfiniteDonuts:
My most sincere and heartfelt apologies to anyone who is actually waiting on this to get done.

Zitchas:
No rush, but I am very happy to see the work on this continue. Looking forward to having these in the game.

source/Personality.h
Hurleveur: Maybe I was not clear: the methods should be in the same place in the cpp as in the hpp
Isbored should just be after isevasive here, as for the cpp (except if you changed the order in the cpp)

Hurleveur:
I was thinking the same! Feel free to contact me in the discord if smth wasnt clear enough

Amazinite:

Closing this PR as it has been marked as waiting on OP for a few months now. If you return and finalize changes by getting to any outstanding reviews and fixing merge conflicts, then we can reopen it.

PR #5699, opened 2021-02-06
mechanics: Change Dictionary to templated class.

petervdmeer:
**Feature:** This PR changes the Dictionary class to a templated class (after a discussion on Discord).

## Feature Details
This PR changes the Dictionary class to become a template. Templating the Dictionary class will allow it to be used for other datatypes than doubles as values, for example for usage in the ConditionsStore in #5244.

The change looks relatively large, but that is because code had to be moved from the Dictionary.cpp file to the Dictionary.h file (and because `std::` had to be added at a number of locations.

The Intern function is still shared among all templated instances of Dictionary (this could save a little bit of memory compared to each class having its own version).

## UI Screenshots
N/A

## Usage Examples
Use `Dictionary<type>` instead of `Dictionary` in your code.

## Testing Done
- Executed the automated unit tests.
- Started the game and looked at shipyard, outfitter and ships-overview.

## Performance Impact
Template classes should have the same performance as regular classes.

### DIFF
--- a/source/Dictionary.cpp
+++ b/source/Dictionary.cpp
@@ -20,33 +20,9 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 using namespace std;
 
 namespace {
-	// Perform a binary search on a sorted vector. Return the key's location (or
-	// proper insertion spot) in the first element of the pair, and "true" in
-	// the second element if the key is already in the vector.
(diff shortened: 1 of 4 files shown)


petervdmeer:
Changed to draft, awaiting review of #5748 first.

petervdmeer:
Closing this PR based on the Interner sharing concerns as discussed in #5748.

PR #5700, opened 2021-02-07
Tweak: Buffing the Heavy Shuttle

Zitchas:
**Content (Tweak)**

## Summary
This PR bumps the Heavy Shuttle up slightly. The intent is to move it up so that instead of being an "almost immediately after you get going" upgrade to being a more solid "part way between the shuttle and the Scout" sort of ship.

Quoting Amazinite from Discord:
> Start at 450k for the Sparrow, then within the short range of 730k to 870k you have three ships to choose from: the Wasp, Fury, or Beserker. Moving up slightly more you get the Hawk at 1.1m, then you jump to the heavier category with at least 2m. 
> But, start at 350k for the Shuttle, then you can upgrade to the Heavy Shuttle by 490k, but then you have to go all the way to the 1.3m to 1.6m range to get either the Scout or Bounder.
> And start at 400k for the Star Barge, and as described you can't upgrade until the 1.3m to 1.4m with the Clipper and Freighter.
> 
> We don't have the same category spread with transports and freighters (no separate transport categories, and only two freighter categories instead of four combat categories), but the early-game progression of the Shuttle and Star Barge certainly aren't on par with the Interceptor.

## Details

Improving the Heavy Shuttle from being a "slight upgrade" to the shuttle, to being a respectable upgrade.
- Bunks from 8 to 12
- Outfit space from 120 to 130
- Weapon space from 10 up to 15
- Engine Capacity from 60 up to 65
- Mass from 110 down to 85
- Drag from 2.1 down to 1.9
- Hull cost from 320k to 470k (total purchase price goes from 489k up to 639k)

In terms of comparison, I think it is worth noting that - as the first upgrade for the shuttle, and significantly pricier than both star barge and sparrow - it is just on-par with them in terms of outfit space and has less weapon space. Compared to the two starter ships, the heavy shuttle is clearly better for transport running, and can perform decently for cargo, although the Starbarge still holds an edge to it. 

The balance between the new heavy shuttle and other ships is interesting. For instance, the Scout. Which has the same amount of bunks, but does have 20 extra cargo, +100 outfit space, and some other benefits. Is this still a good benefit? Should the Scout get a price drop? (I kind of think it deserves a little bit of one, but I'm not sure.)


WarlordMike:
Looks awesome, Zitchas. About the only improvement I might suggest would be bumping up the Heavy Shuttle's Cargo, as well, though not by much (about 5 tons extra would be a good amount).

It'll still be less than the Star Barge in that regard, but it will give considerable incentive to upgrade from the Shuttle, as a result.

northeast-eph:
Since the Heavy Shuttle is getting more weapons space, maybe Becca's "If you ever feel useless, just remember that the people at Betelgeuse Shipyards spent months while developing the Heavy Shuttle trying to add one more gun port to it, without ever thinking about giving it more weapon space." hail should be removed.

Amazinite:
Well I would suggest not even increasing the weapon space. The Heavy Shuttle should be a better shuttle than *the* Shuttle, not a better interceptor. Also watching the mass, drag, and engine space changes; I'd expect the Heavy Shuttle, being heavier, to be slower than the Shuttle, the tradeoff being the increased bunk space so that you can do transporty things better.

northeast-eph:
I did some testing between a Heavy Shuttle with the boosted stats and one with the current stats. I tested a Beam Laser, dual blasters, and a Meteor launcher on the current Heavy Shuttle, and a Heavy Laser, dual modified blasters, a Javelin pod, and a Sidewinder launcher on the boosted one. The only weapons that I found were practical on either one were Sidewinders and Meteors. Any other weapon configuration wasn't able to take out anything other than a Sparrow, with the Beam Laser being incapable of even that. 
Using a Sidewinder launcher on the one with boosted stats, and a Meteor launcher with extra ammo on the current one, identical loadouts otherwise (Hai engines+power, Human shield generator, hyperdrive), I fought a variety of Human ships using Omnis, all with the stock configurations. The only ship there was a practical difference with was a Navy Gunboat, due to the better tracking on the Sidewinders. The one with boosted stats was able to disable it by using all 45 sidewinders, while the current one was unable to drop it's shields with up to 60 Meteors. Anything weaker I was able to disable or destroy with either ship, and anything larger or with anti-missile was too durable for either to take out. Both were able to maintain a safe distance easily, though, and outmaneuver any enemy missiles. I can definitely see a player being able to disable and capture a Gunboat with a single Heavy Shuttle being a problem, although it does seem like a problem that would solve itself due to turning the Republic hostile.
I also did some tests for engines and steering between the boosted Heavy shuttle and a standard Shuttle. Using various combinations of Hai and Human engines, I was only able to have either better speed, or better turning on the Heavy Shuttle, but with significant compromises to the other. With the original mass and drag values but keeping the extra engine space, I wasn't able to get the speed or acceleration higher than the standard Shuttle, but I could almost match them with about half the turning. I was still able to get better turning, but it required using the smallest set of Ion thrusters, which gave it abysmal speed.

I hope this is in some way helpful for balance purposes.

Zitchas:
Thanks for the testing and the feedback. It is always appreciated.

I have tweaked the hull mass from 85 up to 90. 

This basically means that if you take the shuttle (which is a fairly old design) and update it with the latest technological advances and refinements and an extra 20 tons of improvements, which results in a 260% price increase, you get a ship that has:

- +100 shields
- +100 hull
- +4 bunks
- +100 fuel
- +20 cargo
- +10 outfit space
- +1 hardpoint (note: There are only two weapons the player could install two of: The blaster and the modified blaster)
- +5 weapon capacity
- +5 engine capacity

PRHeavyShuttlevsShuttle

Pointedstick (approved):
This looks merge-ready to me. I think the slight increase to weapon space makes sense since it provides the opportunity to fit a Sidewinder launcher, which is much more useful on a ship of this size than any other weapon, given that its use it anti-fighter defense. That's exactly the kind of ship that's going to be dangerous to a Heavy Shuttle.

@Amazinite?

Amazinite:
Seems unnecessary given that #5772 and #5779 are open and will likely just overwrite whatever changes are done by this.
And alongside my quote from the PR description, I'd rather address the progression of all the starting ships be looked at all together instead of only addressing one and leaving the other(s) hanging.

Whatever the case, I'm not touching anything until v0.9.13 is released.

Zitchas:
The main argument *for* this would be that it gives new pilots a clear potential progression for the transport side of things right now. This could potentially be useful for new players in particular for the next six months to a year that is likely to take for 0.9.15/0.9.16 be released. Given that we could fix this at least partially right now, I don't think we should wait potentially a year to get the general playerbase an upgrade path here.

And the larger systemic changes described in 5772 and 5779 are the sorts of things that might be worth being 0.9.15 all by themselves.

Pointedstick:
This seems reasonable, and #5772 and #5779 are never going to be merged in their current forms. Mega-PRs like that are doomed to become idea graveyards. I'm going to merge this right now, and we can investigate overhauling the progression of the starting ships in follow-up MRs, if needed. No need to do everything all at once, and this isn't exactly a game-breaking change. :)

PR #5702, opened 2021-02-07
Void sprite corrections

### LINKED ISSUE #5693: New void sprites mismatch with existing effects
**Describe the bug**
#5673 changed the void sprites, but not their death effects.

**Screenshots**
effect (first frame):
void sprite adult death~0
new void sprite (first frame):
void sprite adult-00

beccabunny:

## Summary
The stuff that was forgotten when the void sprites visuals were updated. #5693

 - [x] I uploaded the necessary image, blend, and texture assets here: [void sprite corrections.zip](https://github.com/endless-sky/endless-sky/files/5939329/void.sprite.corrections.zip)
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/166


Zitchas:
The two scenes look great; but the death images look really bright & pink compared to the old ones. That being said, I'll reserve judgment on those until I have the chance to go test it. Some Void Sprites may be harmed in the process. :(

Zitchas (changes requested):
The two scene changes look good.

The Void Sprite deaths are good on shape, however the increased brightness and shift from the old purple to the new pink color doesn't look too good. Could you revert to the previous color scheme please?

beccabunny:
Better?

Anarchist2:
Shouldn't you also change the infant death animation?

beccabunny:
It is basically identical to what is was before, so it would be useless. No disparity is noticeable in game

Zitchas (changes requested):
The darker pink is definitely better than the initial light pink; but it still isn't as good as the original purple color. The shape, however, looks great.

I'd just like to point out that the Void Sprites aren't exploding with a reactor melt-down like ships do, there isn't anything internal to them that should be *bright*. Currently, the pinks give me the impression of something bright compared to the Void Sprite that is doing the exploding, especially against the dark of space. I think this is part of why the dark color looks better.

Amazinite:
I'll take the landscape and scene image, but the death animation could use some work, either to look more like the old one (what I'd prefer) or have the infant death animation match the new adult death animation (which Zitchas seems interested in).

Zitchas:
Perhaps elaborating a bit might help, as there are a few items that change here. The goal, in my view, is to have something that is exactly like the old death animation, but with an updated shape.
So:
1) color: The main color for the old animations is a purple. A fairly dark one. This contrasts heavily with the pink color which is both brighter and a brighter color.
2) Aura: The old ones generally had an aura around them that fades off, whereas the new ones currently  feel like a harder edge with less aura.

PR #5509, opened 2020-11-20
feat(testing): add a unit testing framework

tehhowch:
**Feature:** This PR integrates the Catch2 testing framework, allowing extended tests of our development code.

## Feature Details
 - integrates Catch2
 - Reworks the build pipelines for linux and windows
   - linux and Windows builds invoke link-time optimization (LTO), reducing the binary size
   - replaces Windows makefile rules with the Scons logic
 - Code::Blocks project is now a workspace environment
   - Users will need to update their CB compiler toolchain settings to use `gcc-ar` and/or `gcc-ranlib` to enable LTO support.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
It tests itself 😉

## Performance Impact
No performance impact to the game binary is expected. Linking a release binary will take longer, due to LTO.

This PR does not write all the tests we could write yet, nor is it a comprehensive solution to automated testing of the game, but does enable  us to improve confidence in existing code and proposed refactors.

PR #5704, opened 2021-02-07
Update readme-developer to Markdown, centralizing build instructions

Terin:
Since #5509 was merged, fairly major changes were made to the build process, especially in Windows, that were not fully or clearly reflected. We've had two different documents with broadly similar instructions, but important differences. This PR addresses that by directing the user to readme-developer, updated to use Markdown, so we have a centralized, human-readable location for build instructions we can update more easily.

It also overhauls the Windows build instructions, which have had many issues causing people to be unable to follow them. These updates include:

- Direct link to the MinGW Online Installer, with instructions on what version/settings to use, rather than a zip folder you are expected to unzip into Program Files.
- Explicit sections for CodeBlocks and Scons instructions
- Specific path for MinGW in CodeBlocks
- Moving a critical step to call gcc-ar.exe instead of ar.exe from the end of the document to the middle
- Directions on how to add MinGW tools to the path for Scons
- General improved flow between instructions, verified on a clean VM, and with other users compiling for the first time
- (Potentially) removing the need to copy separate libraries from MinGW, and including them in the precompiled libraries zip.

## Questions
First, for that last bullet, I tested including the MinGW libraries from version 8.1.0 in ``C:\dev64\bin\``, and they work on all versions down to 5.1.0 (where previous versions failed to compile, regardless). We can remove a major, complicated step of the Windows build process this way. However, since these are precompiled libraries and not vettable source code, should a trusted reviewer update [this file](https://github.com/endless-sky/endless-sky.github.io/blob/master/win64-dev.zip)? Or can I make that PR myself?

Second, should we drop (instruction) support for 32-bit Windows? It's 2021, the last person I saw using 32-bit Windows was someone who accidentally pirated that version of Windows 10, when Microsoft basically provides the 64-bit version in all other cases, for free. And according to the [Steam Hardware Survey](https://store.steampowered.com/hwsurvey?platform=pc), 0.18% of users  are on 32-bit Windows 10, of which potential developers for ES would be even more disproportionate. Dropping it will result in cleaner instructions, and I think we would be safe relegating 32-bit Windows to the same place as ARM support. And even then, ARM support may get more demand in the future, while 32-bit Windows will only get less and less, if any is even left.


Zitchas:
Thanks for putting this together. As someone who used to build Windows versions in the past but hasn't really managed to do a "nice" build (aka one that didn't require having a command line window open while running ES, which scons builds seem to require), I look forward to testing out these updated instructions.

Terin:
Hope it'll help, I've gotten good feedback from others so far. Might be able to update this document later with instructions on how to integrate Scons with something like VS Code, as well. That way, you don't have to be committed to CodeBlocks, but also don't need to have a command line open every time.

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+
+Two methods of building the game on Windows are available: via the Code::Blocks IDE, and via Scons on a command line. Both options require a MinGW toolchain, and precompiled libraries provided below.
+
+MinGW provides the tools used to compile the source code into the game's executable. For most cases, you should download the [MinGW-W64 Online Installer](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/installer/mingw-w64-install.exe). When installing, use the following settings:
+
+- Version: `8.1.0`
+- Architecture: `x86_64`
+- Threads: `posix`
+- Exception: `seh`
+- Build revision: `0`
```
tehhowch, line 66:
```suggestion
```
We shouldn't specify the revision number

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+
+
+
+## Windows
+
+Two methods of building the game on Windows are available: via the Code::Blocks IDE, and via Scons on a command line. Both options require a MinGW toolchain, and precompiled libraries provided below.
+
+MinGW provides the tools used to compile the source code into the game's executable. For most cases, you should download the [MinGW-W64 Online Installer](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/installer/mingw-w64-install.exe). When installing, use the following settings:
+
+- Version: `8.1.0`
```
tehhowch, line 62:
```suggestion
- Version: `8.1.0` (or greater; this document assumes 8.1.0)
```

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+```
+
+The produced test binary will be invoked with a reasonable set of arguments. To run any test benchmarks, you will need to invoke the test binary directly. Refer to the build pipeline definitions for the current required arguments.
+More information on unit tests can be found in the project's "tests" directory.
+
+
+
+## Windows
+
+Two methods of building the game on Windows are available: via the Code::Blocks IDE, and via Scons on a command line. Both options require a MinGW toolchain, and precompiled libraries provided below.
```
tehhowch, line 58:
```suggestion
Two methods of building the game on Windows are supported. The official build is done on the command-line with the Scons program, but the game may also be built using the Code::Blocks IDE. Both require a MinGW toolchain and the precompiled libraries linked below.
```

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+
+MinGW provides the tools used to compile the source code into the game's executable. For most cases, you should download the [MinGW-W64 Online Installer](https://sourceforge.net/projects/mingw-w64/files/Toolchains%20targetting%20Win32/Personal%20Builds/mingw-builds/installer/mingw-w64-install.exe). When installing, use the following settings:
+
+- Version: `8.1.0`
+- Architecture: `x86_64`
+- Threads: `posix`
+- Exception: `seh`
+- Build revision: `0`
+
+Endless Sky will also require precompiled libraries to build and run. Download them [here](http://endless-sky.github.io/win64-dev.zip), extract the zip, then move the `dev64` folder to `C:\` in order to build the game without further configuration.
```
tehhowch, line 68:
```suggestion
Endless Sky requires precompiled libraries to compile and play: [Download link](http://endless-sky.github.io/win64-dev.zip)

The zip can be extracted anywhere on your filesystem; to minimize additional configuration you can move the `dev64` folder to `C:\`.
```

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+
+With the workspace open, you can double-click a project in the left-hand menu to activate that project, binding the toolbar & keyboard shortcuts for "Build," "Run,", "Build & Run," etc. to it, no matter which file is being edited.
+
+For example, when making lots of changes to the game, you will generally have the "endless-sky-lib" project active, so that you can ensure changes compile without needing to fully link things together into the actual game binary. After making changes, you would then activate the "EndlessSky" project and use the "Build & Run" option with the "Release" target, which will fully link the compiled code into the small, performant executable.
+
+You can continue making tweaks to files that belong to the "endless-sky-lib" project without reactivating it, so that "Build & Run" will still launch the game with your latest tweaks. Should you experience some show-stopper bug in your modified game, you would then change to the "Debug" build target to help determine the source(s) of the issue.
+
+Before the game will run the first time, however, you will need to skip down to the instructions at **Running the game** below.
+
+### Building with Scons
```
tehhowch, line 90: Scons section should come first, as the official build.

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+### Building with CodeBlocks
+
+If you want to build the game with a visual IDE, then you can use [CodeBlocks](https://codeblocks.org/downloads/26), the main supported IDE for Endless Sky on Windows. As you have already installed a compiler via MinGW, you should select a version without one.
+
+After you install it, CodeBlocks will likely detect a "GNU GCC Compiler" from MinGW. Select this as your compiler, then take these steps to configure it:
+
+- Go to Settings > Compiler > Global Compiler Settings > Toolchain Executables. Set the **Compiler's installation directory** to `C:\Program Files\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin`, if you followed the default MinGW installation process above.
+- Modify **Linker for static libs** from `ar.exe` to `gcc-ar.exe`.
+
+After this, the game should be ready to compile. Use CodeBlocks to open EndlessSky.workspace in the directory where you've cloned the game's code. This workspace consists of three projects: one for the majority of source code, one for the game binary, and one for the unit tests. As with most IDE-based projects, you must explicitly add new files to the respective project for them to be compiled.
```
tehhowch, line 80:
```suggestion
After this, the game should be ready to compile. Use CodeBlocks to open the "EndlessSky.workspace" file in the directory where you've cloned the game's code. This workspace consists of three projects: one for the majority of source code, one for the game binary, and one for the unit tests. As with most IDE-based projects, you must explicitly add new files to the respective project for them to be compiled.
```

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Building with Scons
+
+If you instead want to build the game from the command line via Scons, you will need [Python 3.8 or later](https://www.python.org/downloads/). When installing, be sure to select the "Add to PATH" checkbox.
+
+Afterwards, you will need to add your MinGW installation to your path manually. To do so on Windows 10, to to **Settings** > **System** > **About** > **System info** > **Advanced system settings** > **Environment Variables**. From each, select **Path** under System variables, and click Edit.
+
+Click New, and then if you used a default MinGW installation, enter `C:\Program Files\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin` as the path, then hit OK to exit out of these windows.
+
+From there, restart your command line if you had one open, and simply `cd` to your Endless Sky directory, then type `mingw32-make.exe -f .winmake`. To speed up the build process, you can add a `-jX` flag to that command, where X is the number of CORES/THREADS? your processor has.
```
tehhowch, line 98:
```suggestion
From there, restart your command line if you had one open, and simply `cd` to your Endless Sky directory, then type `mingw32-make.exe -f .winmake`. (`make`'s standard `-j N` flag is supported, to increase the number of active parallel build tasks to N.)
```

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+
+### Building with a third-party IDE
```
tehhowch, line 100:
```suggestion
### Building with other IDEs
```

readme-developer.md
```diff
@@ -0,0 +1,158 @@
...
+  - libturbojpeg.dll
+  - libwinpthread-1.dll
+  - OpenAL32.dll
+  - SDL2.dll
+  - soft_oal.dll
+  - zlib1.dll
+
+For the **bolded** libraries, those will need to be included from your MinGW installation, instead. By default, you can go to `C:\Program Files\mingw-w64\x86_64-8.1.0-posix-seh-rt_v6-rev0\mingw64\bin`, and copy those libraries to your Endless Sky directory to overwrite the current ones.
+
+**TODO:** Incorporate these libraries into the provided zip, so the user only has to copy from one source.
```
tehhowch, line 122: This won't be done

tehhowch (changes requested):
As mentioned on Discord, we won't be updating the precompiled library versions.

The next changes to the libraries will be when we incorporate unicode support, and when we upgrade the C++ language version.

readme-developer.md
Terin: What should a user put in for N? Is it based on threads or cores? I'm guessing threads, but just want to be sure.

tehhowch: Neither, as with all parallelism the choice of "how many jobs" is dependent on both the system and the amount of work to be done.

Terin: Can we offer a guideline? This is meant for new developers, and the workload is relatively fixed. It doesn't take more than 5 minutes on my terrible laptop. So using a quad-core processor as an example, like an i5-6600k, how many jobs would you run?

readme-developer.md
Terin: The only reason mentioned on Discord was we couldn't be sure what version of MinGW/g++ the user was using. Now, we specify for new developers to use 8.1.0. And even in that case, I went back and test 8.1.0, 7.1.0, 6.3.0, and 5.1.0. All compiled without a hitch on both my current installation and on a VM, and the game ran with only needing to copy from one directory. Is there some other reason I'm missing?

readme-developer.md
MCOfficer: the most common use case is maxing out your CPU, which is achieved by N = Number of logical cores. You may need to clarify this use case to make tehh happy.

oo13:
Do you mean we can drop GCC4.8 support? I hope to drop GCC4.8 which does not support std::regex. I imagined Steam required GCC4.8...
GCC4.8 seems to compile a program including std::regex, but the result is a junk.

readme-developer.md
tehhowch: We shouldnt need to provide any guideline on the value. Explaining how to invoke parallel builds should be enough to get developers started, either learning about "make" and the `-j` flag (because they're new to compiling in general) or just to building (e.g. because they're new to `scons` and didn't know it supports parallel builds, too), or experimenting with different values (because they're curious about the possible speedups).

As the intended audience of this document is people familiar with building software programs, or at minimum intent on learning to build this program, I think it is not our place to explain how other programs work - those programs have entire websites devoted to them with that purpose!

tehhowch:
> Do you mean we can drop GCC4.8 support? I hope to drop GCC4.8 which does not support std::regex. I imagined Steam required GCC4.8...
> GCC4.8 seems to compile a program including std::regex, but the result is a junk.

Early implementations of `std::regex` indeed have bugs. I can't speak to the toolchain used to compile the Steam binaries, though.

Terin:
I tried to compile using MinGW 4.8.5 and it was no good, yeah. Maybe it's an issue with the toolchain overall and not g++, or some specific configuration thing I'm missing, but something definitely is off with that early of a version now.

readme-developer.md
Terin: We don't _need_ to, sure, but the whole point of a guide is to provide guidance. If MCO is right, and the most common use case is to put in your CPU's logical processors for the number of tasks, then I see no harm mentioning that. If not, and you really need to do lots of in depth reading or experimentation to find anything close to useful, then it's fine as-is.

MasterOfGrey:
I tested the windows build instructions after a very long day at 11pm at night, having not tried to do anything like this in 10 years (and even then with a whole different setup for a single brief event).
Bar a couple minor issues (resolved and incorporated pre-PR) I can attest that the windows build instructions are fairly foolproof.

tatami4:
https://github.com/endless-sky/endless-sky.github.io/issues/4

readme-developer.md
Zitchas: As a new developer for whom scons is a misspelling of a tasty bread product rather than something to do with games, having the guide spell out common flags and how to use them would be appreciated. It is a guide, after all, and having it be all-inclusive as to what it takes for me to build things quickly and effectively would be much appreciated.

The main thing, I suppose, is to not make any assumptions about our developer's knowledge.

PR #5707, opened 2021-02-07
feat: Print firing and hit forces in weapons table overview

petervdmeer:
**Feature:** This PR implements the feature as described in the title

## Feature Details
Firing force and hit-force can be important weapons characteristics to take into account when balancing ships.
This adds those two forces to the weapons table overview so that content creators can easily use them for balancing ships and weapons.

## UI Screenshots
N/A

## Usage Examples
Run `endless-sky --weapons` to get the overview.

## Testing Done
I used the newly printed values to choose weapons for an NPC ship I was creating.

## Performance Impact
None, only affects the printing function which is not run in-game.

### DIFF
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -1260,8 +1260,9 @@ void GameData::PrintShipTable()
 void GameData::PrintWeaponTable()
 {
 	cout << "name" << '\t' << "cost" << '\t' << "space" << '\t' << "range" << '\t'
-		<< "energy/s" << '\t' << "heat/s" << '\t' << "shield/s" << '\t' << "hull/s" << '\t'
-		<< "homing" << '\t' << "strength" << '\n';
+		<< "energy/s" << '\t' << "heat/s" << '\t' << "move/s" << '\t'
+		<< "shield/s" << '\t' << "hull/s" << '\t' << "push/s" << '\t'
+		<< "homing" << '\t' << "strength" <<'\n';
 	for(auto &it : outfits)
 	{
 		// Skip non-weapons and submunitions.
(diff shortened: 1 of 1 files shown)


source/GameData.cpp
```diff
@@ -1260,8 +1260,9 @@ void GameData::PrintShipTable()
 void GameData::PrintWeaponTable()
 {
 	cout << "name" << '\t' << "cost" << '\t' << "space" << '\t' << "range" << '\t'
-		<< "energy/s" << '\t' << "heat/s" << '\t' << "shield/s" << '\t' << "hull/s" << '\t'
-		<< "homing" << '\t' << "strength" << '\n';
+		<< "energy/s" << '\t' << "heat/s" << '\t' << "move/s" << '\t'
```
tehhowch, line 1263: "move" seems non-intuitive. I wonder if we need to provide description lines before or after the table, to enable us to use terse / unintuitive column headers.

petervdmeer: Maybe "recoil"?

I prefer not to add lines before or after the table itself; the current table exports nicely into a tsv (tab separated values) file by using `endless-sky --weapons > weapons.tsv` and can then easily be imported into Excel or OpenOffice Calc.
If we want to give descriptions, then a markdown file for the wiki might be more appropriate (or we could switch from terse headers to less terse headers).

tehhowch: IMO, header & footer rows don't harm the ability to import data into spreadsheet software, especially when the data delimiter isn't common punctuation.

If we introduce them, I recommend we use header rows over footer rows, so the column descriptors can be visible when the column headers are. Probably we do not need to annotate all column names, either.

petervdmeer, line 1263:
```suggestion
		<< "energy/s" << '\t' << "heat/s" << '\t' << "recoil/s" << '\t'
```

petervdmeer: For now I changed the text to `recoil` to keep the header tense but hopefully less unintuitive.

PR #5713, opened 2021-02-08
Change "Northwestern" to "Northeastern" in "FW Rand 1"

Anarchist2:
Rand and Oblivion are on the North-eastern fringe of the Dirt Belt, not the north-western.


MCOfficer (approved):
This has been here since 824edf21801bcde8ccde3c7135964d1386c1aec5 :smile:

PR #5714, opened 2021-02-09
Fix: Premature haze culling

### LINKED ISSUE #5711: Bug: Haze is culled too early when zoomed in
**Describe the bug
When playing at a high zoom level, the haze sprites in flight will sometimes cull too early, causing them to vanish before they're fully off screen.

**To Reproduce**
1. Enter flight, and zoom in all the way
2. Slowly fly and observe haze sprites leaving the screen
3. Notice some of them vanishing before fully off screen

**Expected behavior**
The haze sprites will stay on screen for all zoom levels

**Screenshots**
https://user-images.githubusercontent.com/4471575/107163263-416e5180-696e-11eb-8eaf-e0b71370231f.mp4

 - OS: Windows 10 Pro
 - Game Source: Self-compiled
 - Version: commit hash a9512d6

**Link to save file**
N/A

Terin:
**Bugfix:** This PR fixes #5711.

## Fix Details
There is an obscure interaction with the zoom variable under the StarField class, where a value of roughly 1.2 or higher causes the haze sprites to start to cull early. While a more elegant fix is likely possible, this PR simply sets the zoom value to the lowest zoom of .25, drawing excess sprite off-screen to ensure none will be culled in the problematic zoom levels before they are well off-screen.

## Testing Done
Careful observation of haze sprites at the edge of the screen when at the maximum zoom, while zooming in and out, and with an upcoming PR for parallax scaling.

## Performance Impact
On my Acer Chromebook C720, a laptop with a weak GPU, I compared GPU usage with and without the patch. In empty space, there was a usage of 9-10% zoomed all the way in, and 18-19% zoomed out, with and without the patch. Drawing the hundreds/thousands of stars is significantly more GPU-hungry than the few dozen haze sprites, so going with this slightly brute-force fix should have practically no performance downside.


source/StarField.cpp
```diff
@@ -200,6 +200,9 @@ void StarField::Draw(const Point &pos, const Point &vel, double zoom) const
 	else
 		transparency = 0.;
 
+	// Set zoom to a higher level than stars, so the haze won't get culled prematurely
```
Amazinite, line 203:
```suggestion
	// Set zoom to a higher level than stars to avoid premature culling.
```

Amazinite (approved):
Flying around with this change and not spotting any premature culling. Definitely an improvement.

PR #5716, opened 2021-02-09
Use hendrikmuhs/ccache-action for managing ccache in GitHub Actions

janisozaur:
**Feature:** Improve ccache setup

## Feature Details
The previous setup did not really restore anything so ccache wasn't utilised properly.
This action:
1. ensures ccache is installed
2. configures ccache to have 500MB size
3. enables compression
4. displays stats
5. automatically pulls/pushes the cache once it's done with using it

## Testing Done
Implemented it already across multiple other repositories

## Performance Impact
CI jobs should finish faster. Detailed times can only be known after the caches get populated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,6 @@ jobs:
       matrix:
         os: [ubuntu-20.04, ubuntu-18.04, ubuntu-16.04]
     env:
-        CCACHE_DIR: ./ccache/
         CXX: ccache g++
         AR: gcc-ar
         RANLIB: gcc-ranlib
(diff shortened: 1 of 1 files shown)


MCOfficer (approved):
I'm not opposed.

One semantic that is being changed is the cache - previously, it was as specific as possible to optimize the cache (a branch would have its own cache, and only fall back to the main cache if it wasn't available), while it's now being shared between *all* actions. That being said, it shouldn't make a noticable difference.


> CI jobs should finish faster. Detailed times can only be known after the caches get populated

Not sure why you'd think that. In fact, it's a bit slower because there's an extra round of apt update & install.

janisozaur:
> In fact, it's a bit slower

I asked a computer and it said it's faster: https://www.wolframalpha.com/input/?i=3m+50s+%3C+2m+37s

I usually trust computers with such calculations.

es_ccache_initial_run
es_ccache_cached_run

MCOfficer:
The 3m50s are a run without any established cache, the 2m37s already have one. So you're comparing runs without ccache to runs with ccache.

My comment was comparing your ccache ones to those on master. But upon having a closer look, the fluctuations between runners have more impact than the additional `apt update` and `apt install`, so this PR doesn't change anything in terms of performance.

janisozaur:
I've updated my commit so the ccache action gets executed after the extra repos get removed from apt.

I see now the ccache setup used currently does actually work - perhaps it was a matter of the build I chose for inspecting initially, but I remember it explicitly stated something akin to "no cache found" and I wrongly assumed there's something wrong (probably) with the keys in general. I've checked another job with current setup and it said it does indeed restore correct cache.

As ccache's compression is on by default, this PR doesn't really change much other than doing a minor cleanup by replacing the custom step into a marketplace action and producing the cache stats after a run is done and the change to cache key.

tehhowch:
I don't see the value of this PR; we already cache things successfully.

PR #5717, opened 2021-02-09
Player name entry improvements

Terin:
## Feature Details
This PR makes a couple of adjustments for the name entry function in the conversation panel. First, both text boxes will be visible at once, one with a slightly darker color, so it is apparent you will need to make two entries at the end of the intro conversation, and moving between them looks better. Second, it provides a flicker functionality, where either pressing enter with one completed text box, or entering an invalid character, will cause the selected text box to jump between two colors for a moment, letting the player know something is not correct.

## UI Video
https://user-images.githubusercontent.com/4471575/107439223-588c7b00-6af7-11eb-9ad3-24cfca22ce4b.mp4

## Testing Done
Entered text in the first name field and attempted to hit enter multiple times, then vice versa. Also attempted each forbidden character, and verify they flickered as well.


data/interfaces.txt
```diff
@@ -18,6 +18,7 @@ color "bright" .75 .75 .75 0.
 color "medium" .5 .5 .5 0.
 color "dim" .25 .25 .25 0.
 color "faint" .1 .1 .1 0.
+color "extrafaint" .05 .05 .05 0.
```
Amazinite, line 21: EIther two words "extra faint" (no other color definition smashes the words together like this), or choose a different word. How about this?
```suggestion
color "dark" .05 .05 .05 0.
```

source/ConversationPanel.cpp
```diff
@@ -112,6 +112,7 @@ void ConversationPanel::Draw()
 	const Color &dim = *GameData::Colors().Get("dim");
 	const Color &grey = *GameData::Colors().Get("medium");
 	const Color &bright = *GameData::Colors().Get("bright");
+	const Color &extrafaint = *GameData::Colors().Get("extrafaint");
```
Amazinite, line 115:
```suggestion
	const Color &dark = *GameData::Colors().Get("dark");
```

source/ConversationPanel.cpp
```diff
@@ -142,15 +143,23 @@ void ConversationPanel::Draw()
...
-			FillShader::Fill(center, fieldSize, selectionColor);
+			// Color selected text box, or flicker if user attempts an error
+			if(flickerTime % 6 > 3)
+				FillShader::Fill(center, fieldSize, dim);
+			else
+				FillShader::Fill(center, fieldSize, selectionColor);
+			if(flickerTime)
+				--flickerTime;
+			// Fill non-selected text box with dimmer color
+			FillShader::Fill(unselected, fieldSize, extrafaint);
```
Amazinite, line 162:
```suggestion
			FillShader::Fill(unselected, fieldSize, dark);
```

source/ConversationPanel.cpp
```diff
@@ -142,15 +143,23 @@ void ConversationPanel::Draw()
...
 				continue;
 			}
 			
-			// Fill in whichever entry box is active right now.
-			FillShader::Fill(center, fieldSize, selectionColor);
+			// Color selected text box, or flicker if user attempts an error
+			if(flickerTime % 6 > 3)
+				FillShader::Fill(center, fieldSize, dim);
+			else
+				FillShader::Fill(center, fieldSize, selectionColor);
```
Amazinite, line 158: I'd simplify this with a ternary operator.
```suggestion
			FillShader::Fill(center, fieldSize, (flickerTime % 6 > 3) ? dim : selectionColor);
```

Amazinite (changes requested):

data/interfaces.txt
Terin: Hah, damn it. Was wracking my brain for a word darker than faint, and it was right there all along.

source/ConversationPanel.h
```diff
@@ -112,6 +112,8 @@ template <class T>
 	// The current choices being presented to you:
 	std::list<Paragraph> choices;
 	int choice;
+	// Flicker time, set if the player enters invald input for a pilot's name
```
Amazinite, line 115:
```suggestion
	// Flicker time, set if the player enters invalid input for a pilot's name.
```

source/ConversationPanel.cpp
```diff
@@ -142,15 +143,20 @@ void ConversationPanel::Draw()
...
 			// Handle mouse clicks in whatever field is not selected.
 			if(side != choice)
 			{
 				AddZone(Rectangle(center, fieldSize), [this, side](){ this->ClickName(side); });
 				continue;
 			}
 			
-			// Fill in whichever entry box is active right now.
-			FillShader::Fill(center, fieldSize, selectionColor);
+			// Color selected text box, or flicker if user attempts an error
```
Amazinite, line 154:
```suggestion
			// Color selected text box, or flicker if user attempts an error.
```

source/ConversationPanel.cpp
```diff
@@ -142,15 +143,20 @@ void ConversationPanel::Draw()
...
+			FillShader::Fill(center, fieldSize, (flickerTime % 6 > 3) ? dim : selectionColor);
+			if(flickerTime)
+				--flickerTime;
+			// Fill non-selected text box with dimmer color
```
Amazinite, line 158:
```suggestion
			// Fill non-selected text box with dimmer color.
```

Amazinite (changes requested):

tehhowch (commented):
Rather than give the ConversationPanel even more responsibility in an area that's not really its job, let's extract the "(re)name the player" functionality into a new Panel whose sole job is to draw 2 "validated" inputs at a given position and then pass their values to a callback on exit.

In theory we wouldn't need a whole new panel class for this, but we don't have an intermediate abstraction like an `InputZone` that wraps a `ClickZone` with keypress handling.

Terin:
Are there any others examples of UI classes that call children which scroll with them? I've been trying chip away at this in a few ways, but creating a whole new class was outside the scope of this PR when I made it.

tehhowch:
> children which scroll with them?

You can either disable scrolling in the new panel, or bind a const reference to the provided draw position anchor (which should be a class member variable in the host panel and not a temporary) and let the host panel handle scrolling by not trapping all events and returning false in the scroll handle.

E.g.
```c++
class NewPanel : public Panel {
public:
  NewPanel(const Point &anchor, Validator v)
    : drawAnchor(anchor), validationFn(v)
  {}
  // ...
  bool Scroll(...) override { return false; }
  bool TrapAllEvents() const override { return false; }
  // ...
private:
  const Point &drawAnchor;
  Validation validationFn;
  // ...
};
```
The associated interfaces.txt definition would then be combined with the draw anchor to position the input(s) appropriately.

Terin:
Re-opening after talking with @quyykk / @warp-core, where even they weren't sure how to proceed with refactoring this code into a new class. When/if we wanted to do that factor, the additional code in this PR is simple enough to include in that, so I'm now looking to get this merged.

quyykk (approved):
Even though it's not great that the conversation panel does yet another thing, I'm not sure how to do this with a panel.

> In theory we wouldn't need a whole new panel class for this, but we don't have an intermediate abstraction like an InputZone that wraps a ClickZone with keypress handling.

this sounds much better, but can always be done later

warp-core (approved):

PR #5722, opened 2021-02-15
Tweak: Adding Reverse Thrusters to the stock Star Queen

Zitchas:

## Situation:
The Star Queen is a luxury liner. It is not designed for fast travel, but rather to transport wealthy clients in luxury. This implies a strong connection to the cruise liners prevalent today. In playing with the Star Queen, it feels like it spins far too quickly for a ship of its bulk, particularly when one thinks in terms of swinging the passengers around inside.

At the same time, the Star Queen's design is perfectly made for reverse thrusters to look good. 

Given this situation, this PR merely downgrades the steering by one category, and adds a set of Reverse thrusters to its stock layout. In order to accommodate this change, the Star Queen loses one missile launcher. The atomic reverse thrusters might be slightly on the strong side, but not significantly so. It enables the Star Queen to have a very graceful glide into port and pull to a stop. A much more befitting arrival for the wealthy who are her clientele.

## Details
Changes the following on the Star Queen:
- Adds reverse flares
- Adds reverse engines
- Removes one missile launcher in order to fit the reverse thruster.

Note:
- Human space effectively only has a single reverse thruster. While having white atomic reverse thrusters with blue primary thrusters doesn't look *bad,* it would be nice to have the possibility of matching them. 
- Likewise, this doesn't impact the limits for the player. If a player wants to strip out the reverse thrusters and swap back to a faster turning steering, they can. So no impact on player usage.

## PR Checklist


data/human/ships.txt
```diff
@@ -3133,8 +3133,8 @@ ship "Star Queen"
 		"heat dissipation" .65
 		"fuel capacity" 700
 		"cargo space" 60
-		"outfit space" 360
-		"weapon capacity" 120
+		"outfit space" 366
+		"weapon capacity" 124
```
Amazinite, line 3137: Remove something to make room instead of increasing the space. The Star Queen isn't meant for combat so ditch a Sidewinder.

Amazinite (changes requested):

data/human/ships.txt
Zitchas: That makes excellent sense. Now that you mention it, having a luxury cruise ship carrying 135 missiles seems... weird. A carrier only has 150, and that's a Heavy Warship. Central missile tube remove, and missiles dropped correspondingly dropped to 90.

Zitchas:
@Anarchist2 @Amazinite  As it currently stands, there is no change to the base hull now, only a change in stock configuration so that it relies on reverse thrust to come to a stop instead of turning. Does this resolve your 👎 ?

thrashie (approved):
Love the idea, will turn a little bit slower but reverse thruster should compensate.

Pointedstick (approved):
Makes total sense to me.

BaddestApple:
Turning is for peasants...I like this change.

Zitchas:
@Amazinite I know in theory all these tweaks were put on hold for the imminent merging, and I suppose it still is imminent. But might it be possible to get your feedback on it, given that your earlier review has been resolved, and another Dev has approved it?

PR #5727, opened 2021-02-18
Fix mission map gift

LukeMarlin:
**Bugfix:** This PR aims at starting a conversion about "Local Map" requirement and gift for missions.

## Context
I recently started playing and didn't appreciate the manual map buying: go to the outfitter, scroll to the bottom, select map, then buy. This routine is tedious, and due to the low cost of maps, I wanted to make a plugin that would do that automatically for me. The thing is, "Local Map" is a special outfit. When giving it as a mission reward, it doesn't behave properly: it actually equips a `Local Map` outfit to the ship, and that's it. It does not update the map, and you can even sell it.
When it comes to requirement, it also looks for an actual `Local Map` outfit on the ship, where I expected it to check if the player had mapped nearby systems.

Once I faced this, I checked the source and played with it. This PR is the result of that process. It makes the `Local Map` gift work, and possibly the requirement (I'm not sure yet, but I guess it should). I do _not_ expect this to be merged, I am merely showing my updated source to ask some questions:

* Would that kind of change be desirable?
* If yes, would it be better to have `HasMapped` return a percentage instead? With my code, if you have visited 6 systems around you and you'd have a mission requiring you to have visited 6 systems, it could fail because the sytem would look around with a specific order.
* Would it be interesting to be able to require stuff like: has player mapped X systems around system Y? I could imagine recon missions with this
* Would it be interesting to be able to give the `Local Map` of a specific system? Possible use case: upon accepting a mission the player could be given the local map of the destination for example (there is the `visit` directive, but since systems can be updated, by events, it might be a bad idea to hardcode `visit` directives to all nearby systems)
* More general thought: I don't feel like the current mapping system is adding to the gameplay. I played an hour or so, and ended up trying to automate it. Might be just me, but it looks like a chore quite quickly, because it's cheap, there's almost no reason not to buy maps. Could it be changed to a completely automatic process in the core game? (I could give it a try)

## Fix Details
Fix for map gifting is to move the Mapping code outside from the buying action to the equipping outfit action. 
Fix for map requirement is to add a special case in Mission condition checking

Note: Sorry for the long text, hopefully this doesn't come as rude or spammy, I'm really trying to share my experience and ideas here but I'm having trouble making all this synthetic. (side note: I'm not a cpp dev)

## Usage example
Inside a mission specification:
```
mission test
    <snip>
    on offer
        require "Local Map" 3 # Requires to have mapped at least the three "closest" systems
        # require  "Local Map" 0 # Requires to not have mapped the entire mapSize

    source
        outfits "Local Map"

    on complete
        payment -1000
        outfit "Local Map" # Will actually make the map available to the player
```

## Testing done

I've used the snippet above to ensure that giving the map revealed the system and did _not_ install a map in the ship
I've used the snippet above to ensure that requirement worked: I set it to 0, and was only offered the mission on non fully visited systems. I've then set it to 3, then 2 and 4 to check that I was offered the mission only when I had visited at least the respective number of systems around (by checking the order that internal code uses, which might be confusing!)

## Performance impact

I did not check, but I don't think there is any given the amount of code and the complexity of it :)


petervdmeer:
If giving a "Local Map" causes it to be installed, then that sounds like a bug, or at least like a very undesirable feature. Fixing this by making the `give "Local Map"` behave the same as the `buy "Local Map"` sounds good to me.

I expect the `give "Local Map"` to often be used in situations where a content writer wants to make sure that the local area is known, so letting the `give "Local Map"` also succeed if the area was already known to the player makes a lot of sense.

I didn't look into the details of your code yet.

The ability to check how much of the area is explored could be also interresting. I would put that in a separate PR though, since it looks sufficiently different from this `give "Local Map"` fix.

I feel that giving "Local Maps" of remote locations should not be done through the "Local Maps" outfit, but could better be done by specifing a (target/origin) system and a range. For me this also feels sufficiently different to go into another separate PR (and you might want to open an issue first to discuss if this feature is desired).

If you want to automate, then you could probably write a plugin with a repeating mission that automatically gives a "Local Map" on every landing (after the fix in this PR here). I don't expect this to land in the main game though, many other players seem to be okay with manual exploring, buying maps or leaving some of the universe unexplored.

LukeMarlin:
Thanks for the response @petervdmeer !
It's been a while now, but yeah, I tried and am sure that it just installs it.

The plugin I intended to write was supposed to do exactly what you said, but yeah, didn't end well due to this behaviour :)

I didn't dive back into the code yet, this PR was also there to open the discussion. If you think that the idea of making the "give map" behaviour the same as "buy map", I could check back what I wrote, rebase it on master en ensure it's okay, then make the PR a normal PR instead of draft :smile:

petervdmeer:
I think that making the "give map" behavior the same as "buy map" is a good idea.

LukeMarlin:
Alright, I did a rebase, tested again my changes and it looks OK:

- Buying works as before
- Gifting a map works as discussed
- Requiring "local map 0" triggers when the player has not explored `mapSize` systems around
- Requiring "local map n" triggers when the player has explored `n` systems around

I've read the contributing and part of style guide (mostly spacing, hence my latest commit), I might have missed something, please tell me if that's the case and I'll fix right away.

Some questions from the original post still stand, but as stated they might deserve separate PRs. I'll just keep one of them as current implementation of requiring "local map n" might not be obvious: `HasMapped` currently looks for systems in a specific order, meaning that if you have visited 3 systems and a mission requires you to map only 2, if you didn't visit the _correct_ ones, condition would fail. Do you think we should provide a different method for that?

Last note: I didn't dive into tests yet, as again, I'm no CPP dev so I'm just going to ask directly: is it practical to write tests for these changes? If yes, could you point me to a file I could use as a quickstart? :)

source/OutfitterPanel.cpp
```diff
@@ -363,15 +362,9 @@ void OutfitterPanel::Buy(bool alreadyOwned)
 		int mapSize = selectedOutfit->Get("map");
 		if(mapSize > 0)
 		{
-			if(!HasMapped(mapSize))
```
petervdmeer, line 366: Why removal of the check if the player already has this distance mapped?

source/MissionAction.cpp
```diff
@@ -419,6 +431,25 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
 	
 	for(const auto &it : requiredOutfits)
 	{
+		// Special case: map
+		int mapSize = it.first->Get("map");
+		if(mapSize > 0)
```
petervdmeer, line 436: Can you show (in the text-description of the PR) how this would look in a datafile written by some content-writer? (Maybe also for the gift section.) (This usage information is normally done in the "how to use" section from the standard template for PRs.)

tehhowch (commented):
As you mention, the likely intent of the mission author is that the player doesn't know X of the nearby Y systems. So it would indeed be good to investigate changes to the mapping algorithm.

However, there is no harm in the current construct, as a map-based mission will be able to trigger when the player would be able to buy a map of the same size. So this investigation can be deferred to a different PR.

 (It may end up being desirable for "visited X of Y neighboring systems" to be a testable condition, rather than a `require` or `outfit` ownership check.)


As far as tests for the code you're adding, you wouldn't be able to use our current framework for unit tests. You could work with Peter on writing some acceptance tests that involve offering (or not offering) missions using test pilots with the requisite system knowledge. Additional eyes and ideas on that framework are helpful.

source/OutfitterPanel.cpp
LukeMarlin: It seemed to me like a duplicate check, as `::CanBuy` already does it: https://github.com/endless-sky/endless-sky/blob/master/source/OutfitterPanel.cpp#L316

Are there are flows where `::Buy` is used alone without first going through `::CanBuy`?

source/MissionAction.cpp
LukeMarlin: Sorry about that. I think that I originally reduced the template because it was just a draft and simply wanted to open the subject. When transforming it, I wasn't proposed the template again. I've updated the original post

LukeMarlin:
> As far as tests for the code you're adding, you wouldn't be able to use our current framework for unit tests. You could work with Peter on writing some acceptance tests that involve offering (or not offering) missions using test pilots with the requisite system knowledge. Additional eyes and ideas on that framework are helpful.

I'm unsure where to start with this. I've tried to locate where these tests are, are they the ones ran with `run_tests.sh` (or headless)? I've tried it and it seems there are only 3, mostly empty ones (from data/tests/tests_framework.txt). It does look nice as it seems to be declarative, but I've no idea where to start with this as I didn't find useful examples. I'd gladly give it a try though if anyone can provide some pointers! :smile:

petervdmeer:
Thanks for already looking in the test(s); PR #5736 extends the framework with quite a number of additional tests (and framework functionality). It should be possible to write a test that does the following:
- Test-step 1: Load a savegame that has:
  - a mission on a target planet that always offers on landing.
  - inside that mission a gift of a map.
  - a setup that ensures that the mission does not need any conversation or input on offering.
  - navigation set to this target planet.
  - planets on the navigation route set as visited.
  - planets around the target planet not visited yet.
  - a flagship that is capable of traveling to the target planet.
- Test-step 2: Verify that planets around the target planet are not visited (using an `assert` together with the has-visited condition).
- Test-steps 3 and 4: Execute the "depart function" and the "jump function" so that the ship flies to the target planet.
- Test-steps ~5: Have the system execute the "check-loop" that checks if the ship already landed.
- Test-step ~6: Check on/after landing that the planets around the target planet are now visited.

petervdmeer:
Is this text `require "Local Map" 3 # Requires to have mapped at least the three "closest" systems` correct? Or should it say `#Requires to have map at least all systems within the given distance/radius`?

source/MissionAction.cpp
petervdmeer: > Some questions from the original post still stand, but as stated they might deserve separate PRs. I'll just keep one of them as current implementation of requiring "local map n" might not be obvious: `HasMapped` currently looks for systems in a specific order, meaning that if you have visited 3 systems and a mission requires you to map only 2, if you didn't visit the _correct_ ones, condition would fail. Do you think we should provide a different method for that?

If this only checks if the player at least knows <nr> systems nearby, then I would expect the check to count all known nearby systems and succeed if the known systems is at least this number (I don't think we want to have a hidden dependency on specific orders).

source/OutfitterPanel.cpp
tehhowch: Consider it like server-side validation on the web. Even if the client-side does it, it's a good idea to verify assumptions before acting.

source/MissionAction.cpp
tehhowch: @petervdmeer I think for this "outfits" usage, the mismatch is fine, as it means the player can (or can't) buy the named map outfit in the outfitter. The mission writer isn't able to say `require map <system size> <outfit count>`, so the _mission_ never explicitly states how many systems the player has or hasn't visited, just how many of the map outfit are needed (which should always be 1, 0 or omitted).

source/MissionAction.cpp
```diff
@@ -419,6 +431,25 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
...
+			if(it.second == 0)
+			// Condition is to _not_ have mapped the whole local map
+			{
+				if(player.HasMapped(mapSize))
+					return false;
+			}
+			else
+			// Condition is to have mapped at least the given amount of systems
+			{
+				if(!player.HasMapped(it.second))
```
tehhowch, line 447: This should use the map size of the given outfit, not an arbitrary number.

source/MissionAction.cpp
petervdmeer: @tehhowch My concern is more with the "specific order", and "if you didn't visit the correct ones" remarks in the text that I quoted. If that is still an issue, then we would be introducing ambiguity in the keywords that are added here. (I don't like ambiguity for keywords, keywords should be very clearly and exactly defined.)

tehhowch: Yeah, my original review missed that this PR is proposing that `require <outfit> <integer>` behave differently when outfit has a "map" attribute vs. not.

That proposal won't be accepted - `<integer>` in a `require` or `outfit` line under an `on ___` action _must_ be the count of that outfit.

Implementing a queryable condition for map knowledge would be where we allow "player knows x of y nearest systems"

LukeMarlin: Yes, currently the existing functions related to map use a specific order.
So I see two possibilities:
* Changing this PR so that `require LocalMap 1` requires a complete map and `require LocalMap 0` requires a non-complete map.
* Making new functions that do the proper checks, by counting visited systems with no defined order.

I feel like writing a separate check would allow for a better result as it allows more interesting flows. I'd even go with percentages so that `mapSize` can vary without actually affecting existing missions. I'd keep `outfit localMap` as it is in that PR: it just "buys" a map as expected.

What do you think could be the next steps?

petervdmeer: Changing this PR so that `require "Local Map" 1` requires the local map to be complete and `require "Local Map" 0` requires the local map to be incomplete seems to match best with the `give "Local Map"` funcationality also in this PR, so I would recommend to go for this option.

The new functions with the new checks would indeed allow for interesting flows, but [as I said before](https://github.com/endless-sky/endless-sky/pull/5727#issuecomment-827139676), I would suggest to keep those for a separate PR; the new functions will trigger a lot more discussion on desired functionality and exact syntax, while the "Local Map" changes feel more like bug-fixes on keywords already in use (and currently misbehaving).

LukeMarlin: Thank you for the quick reply! I'll try to work this out with tests this week :)

source/OutfitterPanel.cpp
LukeMarlin: Restored with d2a39cd

source/MissionAction.cpp
LukeMarlin: Changed behaviour to a binary one as asked:
`require "Local Map" 0`: requires an incomplete map
`require "Local Map" X` where X > 0: requires a complete map

source/MissionAction.cpp
```diff
@@ -419,6 +431,25 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
 	
 	for(const auto &it : requiredOutfits)
 	{
+		// Special case: map
```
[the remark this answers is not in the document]
LukeMarlin: Changed comments to include full stop, and actual sentences.

source/MissionAction.cpp
LukeMarlin: Done!

LukeMarlin:
I started doing the test, but after a quick talk with @petervdmeer, it seems that what I need isn't there yet, as we're missing either #5017 or #5412.

Once we have an `has-visited` keyword or equivalent, I should be able to finish the test quickly :)

source/OutfitterPanel.cpp
```diff
@@ -359,19 +358,15 @@ void OutfitterPanel::Buy(bool alreadyOwned)
 	int modifier = Modifier();
 	for(int i = 0; i < modifier && CanBuy(alreadyOwned); ++i)
 	{
-		// Special case: maps.
+		// Maps are not transferrable; when bought, they make nearby planets known to the player.
```
tehhowch, line 361: We can keep the existing comment:
```suggestion
		// Special case: maps.
```

source/MissionAction.cpp
```diff
@@ -231,6 +231,25 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
...
+				if(player.HasMapped(mapSize))
+					return false;
+			}
+			else
+			// Condition is to have mapped the whole local map
+			{
+				if(!player.HasMapped(mapSize))
+					return false;
+			}
+			continue;
```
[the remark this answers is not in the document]
LukeMarlin: Indeed, thanks

source/OutfitterPanel.cpp
```diff
@@ -359,19 +358,15 @@ void OutfitterPanel::Buy(bool alreadyOwned)
...
-			if(!HasMapped(mapSize))
-			{
-				DistanceMap distance(player.GetSystem(), mapSize);
-				for(const System *system : distance.Systems())
-					if(!player.HasVisited(*system))
-						player.Visit(*system);
-				int64_t price = player.StockDepreciation().Value(selectedOutfit, day);
-				player.Accounts().AddCredits(-price);
-			}
+			if (player.HasMapped(mapSize))
```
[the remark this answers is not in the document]
LukeMarlin: Done

source/OutfitterPanel.cpp
LukeMarlin: Done

tehhowch (approved):

PR #5728, opened 2021-02-18
Don't mention insurance report in Sparrow description

Pointedstick:
It's established in some of the spaceport flavor text that ship
insurance doesn't exist. Therefore this text about insurance reports
indicating such-and-such about the Sparrow's longevity doesn't make
sense. Let's change it to something else.


Amazinite:
> It's established in some of the spaceport flavor text that ship insurance doesn't exist.

Where?

> Reports leaked to the media...

Reports from where?

Zitchas:
Personally, I always assumed that it was referring to life insurance, loan insurance, or something similar. The sort of thing that would be of zero benefit to the pilot, since we'd be dead, but would be of benefit to the bank or large corporations.

petervdmeer:
And I was thinking that the insurance agent was just working for one insurance company; other insurance companies might still sell spaceship insurances.

Amazinite (approved):

PR #5730, opened 2021-02-19
Only display empty planet error message if moving slowly

Terin:
## Feature Details
This PR adds an additional check when looking to set a landing error message over an empty planet, based on the ship's speed. Currently, if a player wants to land on a planet while traveling at high speeds, but is traveling over an empty planet, they'll mainly hear a beep, and might not even see the error message before realizing their autopilot hasn't engaged.

Now, it'll activate the standard planet targeting and landing unless the player is moving slow enough, as a way to express interest in those planets/objects. This behavior is mostly in-place already if a job destination is in-system, as your auto-pilot will already have a target, but this applies it to all cases. It is currently based on having a speed less than 80 pixels/second, which is just below the lowest top speeds of stock ships (such as the Skein and Bulk Freighter).

## Testing Done
Pressed "Land" while traveling at top speed over a gas giant in a Bulk Freighter, and observed "normal" landing procedure. Then, attempted to land on same gas giant while stopped, and saw error message as expected.


source/AI.cpp
```diff
@@ -3321,7 +3321,7 @@ void AI::MovePlayer(Ship &ship, const PlayerInfo &player, Command &activeCommand
 			else if(object.HasSprite())
 			{
 				double distance = ship.Position().Distance(object.Position());
-				if(distance < object.Radius())
+				if(distance < object.Radius() && ship.Velocity().Length() < (80. / 60.))
```
Amazinite, line 3324: Any specific reason behind choosing a velocity of 80 (as seen in game)? Why not some factor of a ship's max velocity? If we stick with the single value that is considered low speed for all ships then it's probably worth defining this as a variable so that it's a bit more clear what exactly the value represents.

Terin: >It is currently based on having a speed less than 80 pixels/second, which is just below the lowest top speeds of stock ships (such as the Skein and Bulk Freighter).

Went with this reasoning for it, with it being basically stationary. I'll throw it in a variable, though.

source/AI.h
```diff
@@ -193,6 +193,9 @@ template <class Type>
 	
 	bool escortsAreFrugal = true;
 	bool escortsUseAmmo = true;
+
+	// The minimum speed before landing will consider non-landable objects
```
Amazinite, line 197:
```suggestion
	// The minimum speed before landing will consider non-landable objects.
```

Issue #5731, opened 2021-02-20
Bug: Escorts path through systems with the 'Hidden' tag


Zitchas:
**Describe the bug**
Systems with the Hidden tag are supposed to effectively non-existant. However, despite the fact that the player can't jump to said systems, the player's escorts *can* do so.

**To Reproduce**
1. Go to Levana with some escorts that have Jump Drives.
2. Tell them to hold position.
3. Return to Arculus
4. Use the Gather command to order the escorts to return.
5. Watch the location of the escorts, and note that they travel through Postverta.

It is worth noting that this also works in reverse.

**Expected behavior**

**Screenshots**
This screenshot is of my escorts flying through Postverta to go from Arculus to Levana.
![Ships in Postverta](https://user-images.githubusercontent.com/32169904/108585923-75e7f300-7308-11eb-89be-9ce7733915f4.png)

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Git-Hub artifacts, continuous
 - Version: continuous as of hash 1eec2a2 (February 12th 2021)

**Additional context**
Originally reported by upvoteotter on the Endless Sky Community discord. Thank you for spotting it!

Amazinite:
#6127 would resolve this.

PR #5732, opened 2021-02-20
Fix: ships no longer use hidden systems

yjhn:
**Bugfix:** This PR addresses issue #5731.

## Fix Details
Hidden systems can only be reached via hyperspace links.

## Testing Done
Steps from #5731 did not reproduce the issue anymore.

## Save File
[Local God.txt](https://github.com/endless-sky/endless-sky/files/6014933/Local.God.txt)

### DIFF
--- a/source/DistanceMap.cpp
+++ b/source/DistanceMap.cpp
@@ -96,9 +96,9 @@ const System *DistanceMap::Route(const System *system) const
 	auto it = route.find(system);
 	return (it == route.end() ? nullptr : it->second.next);
 }
-	
-	
-	
+
+
+
 // Get a set containing all the systems.
 set<const System *> DistanceMap::Systems() const
 {
@@ -203,10 +203,6 @@ void DistanceMap::Init(const Ship *ship)
 		Edge top = edges.top();
 		edges.pop();
 		
-		// If the system is hidden, it cannot be used.
-		if(top.next->Hidden())
-			continue;
-		
 		// Source is only defined when given a ship and a destination system.
 		// Once we have a route between them, stop searching for more routes.
 		if(top.next == source)
@@ -299,11 +295,16 @@ void DistanceMap::Add(const System &to, Edge edge)
 
 
 
-// Check whether the given link is travelable. If no player was given in the
-// constructor then this is always true; otherwise, the player must know
-// that the given link exists.
+// Check whether the given link is travelable. If it is a link to the hidden
+// system, it is only travelable if it is a hyperspace link. Otherwise, if no
+// player was given in the constructor then this is always true; otherwise,
+// the player must know that the given link exists.
 bool DistanceMap::CheckLink(const System &from, const System &to, bool useJump) const
 {
+	// Hidden systems are only reachable via hyperspace links.
+	if(useJump && to.Hidden())
+		return false;
+	
 	if(!player)
 		return true;
 	
--- a/source/DistanceMap.cpp
+++ b/source/DistanceMap.cpp
@@ -35,7 +35,7 @@ DistanceMap::DistanceMap(const System *center, int maxCount, int maxDistance)
 
 // If a player is given, the map will only use hyperspace paths known to the
 // player; that is, one end of the path has been visited. Also, if the
-// player's flagship has a jump drive, the jumps will be make use of it.
+// player's flagship has a jump drive, the jumps will use it.
 DistanceMap::DistanceMap(const PlayerInfo &player, const System *center)
 	: player(&player)
 {
@@ -202,6 +202,10 @@ void DistanceMap::Init(const Ship *ship)
 	{
 		Edge top = edges.top();
 		edges.pop();
+		
+		// If the system is hidden, it cannot be used.
+		if(top.next->Hidden())
+			continue;
 		
 		// Source is only defined when given a ship and a destination system.
 		// Once we have a route between them, stop searching for more routes.
--- a/source/GameData.cpp
+++ b/source/GameData.cpp
@@ -626,15 +626,7 @@ void GameData::Change(const DataNode &node)
 	else if(node.Token(0) == "news" && node.Size() >= 2)
 		news.Get(node.Token(1))->Load(node);
 	else if(node.Token(0) == "link" && node.Size() >= 3)
-	{
-		System *source = systems.Get(node.Token(1));
-		System *dest = systems.Get(node.Token(2));
-		source->Link(dest);
-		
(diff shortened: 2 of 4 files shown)


Zitchas:
I just tested the first part of this, namely the part ensuring that escorts do not use hidden systems for pathfinding. This fixes #5731 .

That being said, the second part regarding removing the `hidden` attribute when the system gets linked is inappropriate. The point of the hidden attribute is to specifically prevent a system from being revealed. 

As used in the case of Postverta, it is actually intended to be used as basically an "off" switch that prevents Postverta from being reached *by any means* until the appropriate event which removes the `hidden` attribute. This is used because it is simpler (and in theory more reliable) to simply tell the game that the system is 'off`' than to have to create and/or remove hyperlinks between systems.

When functioning correctly, the `hidden` attribute allows all the hyperlinks to be in place for the game's initial error checking, but treated by the AI, map, and navigational purposes as if it wasn't there.

Zitchas (changes requested):
I just tested the first part of this, namely the part ensuring that escorts do not use hidden systems for pathfinding. This fixes #5731 .

That being said, the second part regarding removing the `hidden` attribute when the system gets linked is inappropriate. The point of the hidden attribute is to specifically prevent a system from being revealed. 

As used in the case of Postverta, it is actually intended to be used as basically an "off" switch that prevents Postverta from being reached *by any means* until the appropriate event which removes the `hidden` attribute. This is used because it is simpler (and in theory more reliable) to simply tell the game that the system is 'off`' than to have to create and/or remove hyperlinks between systems.

When functioning correctly, the `hidden` attribute allows all the hyperlinks to be in place for the game's initial error checking, but treated by the AI, map, and navigational purposes as if it wasn't there.

tatami4:
Well, I could not think of any other reasonable way for the pathfinding (DistanceMap) to ignore "hidden" systems when they have not been seen by the player, but allow routing through them when they have been seen. DistanceMap in most circumstances does not know anything about the player, so it cannot know if the player has seen the system or not.

Zitchas:
If a system has the `hidden` attribute, it should *never* be visible, routable, or a valid target for any kind of hyperjump.

tatami4:
My mistake, I didn't realize that "hidden" is removed via event.

Zitchas:
No problem. Right now, I think that Postverta is literally the only two systems in the game with the `hidden` attribute. And while one of the two has no links to it, and thus can never be accidentally revelead by the current fault in the code, Postverta *does* have links to it, which makes this problematic.

Due to the way I have the story built, the systems connected to Postverta are intended to be "present, but hard to access." And are, in fact, impossible to access in the current map but will likely be accessible via very roundabout routes in the future when the other half of the galaxy is filled in. Postverta itself, however, is intended to *never* be accessible no matter what changes happen to the map. It is literally a system appearing where no sensor readings had detected one being until now.

Zitchas:
Don't delete that old code yet... Depending on what happens with #5734 , that functionality might still be useful.

For instance, if a coder decides to implement the `invisible` tag as described in #5734 , then having the `hidden` tag removed from systems as the player visits them, and as such opens them up to be used by the escort pathfinding AI, would be a good thing to have included. 

If no such tag is implemented, though, then the hidden tag shouldn't be changed outside of mission events.

source/DistanceMap.cpp
```diff
@@ -203,6 +203,10 @@ void DistanceMap::Init(const Ship *ship)
 		Edge top = edges.top();
 		edges.pop();
 		
+		// If the system is hidden, it cannot be used.
+		if(top.next->Hidden())
```
Amazinite, line 207: Ships should still be able to make use of hidden systems if they’re linked to another system. I believe this completely disallows travel to hidden systems, which is not the intention of hidden.

tatami4: That's what the first version of this PR did, but according to Zitchas those systems should be unusable when they are hidden.

Amazinite: No, originally this PR completely ignored the hidden tag if a system was linked. That’s different than a system which is hidden only being accessible via links.

Amazinite: It might be easier to just allow all ships to pathfind through a hidden system if the player has seen it. It's not as though the player will be able to tell that NPCs are only now using a system that the player has found. That shouldn't be achieved by holding onto a reference to the player, though. Might it be easier for systems to also keep track of whether they've been seen/visited by the player? I believe that would also help to simplify #5415.

Zitchas: Fair enough. However, if the player hasn't mapped a system - especially one marked as hidden - I don't think escorts should be using it. Escorts should be using the player's map and sticking with known routes, not finding more efficient secret ones.

On the flip side, if escorts are heading off through unexplored systems on their own, they should be revealing the player's map, since, well, the player is getting live updates as to their location, and those are ships that the player owns, so it makes sense that they'd get the nav info.

Zitchas: Maybe? I don't know the code well enough to know if tracking "Have you, system, been seen by the player" makes more sense than asking "Have you, player, seen this system?"

That being said, tracking "found" status by system might make more sense if we're going to have more stuff that is stored that way. For instance:
- What is stored here?
- What ships are active here?
- What ships are parked here?

And we could extend that farther to perhaps tracking certain missions (like the static defending Archon missions) being moved from the player to the system (or planet). And going farther afield, perhaps this could be useful in generating some population "caps" for systems.

Ex. If a certain system has a "population" of 30 merchant ships, and the player kills a significant number of them, it might lower the population for a while, and as such there would be less merchants braving that system.

tatami4:
Ships now only travel to hidden systems via hyperspace links. Two notes:
1. This applies to player's ships.
2. Hidden systems can be reached through wormholes.

MasterOfGrey:
So far as I'm concerned, the attribute was added to avoid having to do the classic workaround where a system is put out of reach and then relocated by event to make it "appear" and become accessible to the player.

If the intended use of the attribute has since changed then we need an attribute that does the original intent.

A system attributed with the original intent should be impossible to either find or travel to (including by your escorts), unless there is there is either a wormhole to it, or the attribute is removed by event.

Amazinite:
I don't believe that the change here matches the intention of a hidden system. As I mentioned on one of the review comments, a hidden system can be found. Once a hidden system is found, you should be able to jump drive to it. I believe that #6127 would be a better PR to resolve #5731, as that PR would make it so that the player's ships don't use hidden system unless the player already knows about them.

PR #5733, opened 2021-02-20
Ignore some of player's input while landing

tatami4:
**Bugfix:** This PR addresses issue where you could see the text "Disengaging autopilot" if you tried to, for example, jump while landing (you cannot jump while landing).

## Fix Details
Ignore all the inputs while the flagship is landing.

## Testing Done
Tried to pause the game while landing, it still works.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -353,7 +353,7 @@ void AI::UpdateKeys(PlayerInfo &player, Command &activeCommands)
 	
 	if(!flagship || flagship->IsDestroyed())
 		return;
-
+	
 	if(activeCommands.Has(Command::STOP))
 		Messages::Add("Coming to a stop.");
 	
@@ -398,7 +398,7 @@ void AI::UpdateKeys(PlayerInfo &player, Command &activeCommands)
 		newOrders.target = player.FlagshipPtr();
 		IssueOrders(player, newOrders, "gathering around your flagship.");
 	}
-
+	
 	// Get rid of any invalid orders. Carried ships will retain orders in case they are deployed.
 	for(auto it = orders.begin(); it != orders.end(); )
 	{
@@ -531,9 +531,7 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
 		
 		if(it.get() == flagship)
 		{
-			// Player cannot do anything if the flagship is landing.
-			if(!flagship->IsLanding())
-				MovePlayer(*it, player, activeCommands);
+			MovePlayer(*it, player, activeCommands);
 			continue;
 		}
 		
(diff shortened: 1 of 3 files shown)


Zitchas:
Just to clarify: At what point does the ignoring of inputs start.

For clarity, I see the landing process as such:
1. Hit L to tell the ship to land.
2. Game picks a landing target.
3. (optional) player hits L repeatedly to rotate through valid landing targets.
4. auto-pilot engages
5. auto-pilot begins manouvers to approach the planet.
6. auto-pilot slows the ship to an appropriate landing speed, more or less coming to a stop over the target.
7. the ship descends onto the planet.
8. the ship shrinks out of sight and we are taken to the landing window.

In this sequence, where does the game start ignoring inputs?

My personal opinion is that it should be possible to disengage the autopilot at any point up until 7. And as far as that goes, I'd actually like it if we *could* interrupt the descent and thus trigger an emerging animation again. I know there's been more than a few times that I went to land and then at the very last second realized that a) escort(s) with the rush cargo hadn't arrived yet and I'll fail if I land and take off again to wait for them, or b) I just noticed a disabled ship I wanted to interact with, etc.

tatami4:
Inputs are ignored from the point when the flagship starts shrinking.

Zitchas:
OK. Thanks for the clarification. While I'd love the ability to interrupt and pull out of landing again, it makes sense from a simplicity perspective that we don't add that functionality in. As such, having the game ignore inputs and thus no display erroneous messages during the descent seems like a solid improvement.

tatami4:
Landing / departing can now be cancelled.

tehhowch:
> Landing / departing can now be cancelled.

Landing and departing are not to be cancelled once started.

tatami4:
Why?

tehhowch:
> Why?

It's an explicit decision to either land or depart. When deciding to land, the player has plenty of time to think about the decision as their flagship routes to the planet and slows down. Taking any sort of movement action here is sufficient to prevent landing.

In terms of relationship with other parts of the gameplay experience, we don't allow ships to do anything while in a transition animation except continue that transition animation. The player's ship should always play by the same rules.

Zitchas:
How many transition animations do we have? The only two that come to mind are landing, take-off, hyperjump, and jump drive. While Hyperjump and Jump Drive I can see having potential lore reasons for preventing interruptions, landing is a much more mundane activity that I think would be reasonable to allow interruptions.

I know there's been many times when my bounty target or some juicy looking pirate has jumped into the system just as I started the landing animation to descend onto the planet. I'd love to be able to interrupt at that point.

Issue #5734, opened 2021-02-20
Bug: Hidden doesn't actually prevent someone from going to a system


Zitchas:
**Describe the bug**
As originally understood, I believed the `hidden` tag on Postverta was supposed to basically act like an "off" switch that prevents it from being visible or accessible, period.

However, upon reading how it has subsequently been defined in the wiki mapdata page, I have realized that my understanding was erroneous, and it now merely conceals a system from indirect detection and does *not* actually prevent it from being accessed.

**Expected behavior**
What I expected was that, so long as a system has the `hidden` tag, it is hidden from view on the map, hyperlinks to and from it are not shown, arrows on the minimap do not point to it, and it cannot be targeted for any form of hyperjumping.

**Proposed Solutions**
**1.** 
A new system attribute `invisible` is introduced, which acts as an "off" switch that prevents a system from being seen or accessed in any way, and hides any hyperjump or wormhole links connecting to it.

**2.**
The `hidden` attribute is corrected to provide the expected functionality.

**3.**
I PR a fix that severs all the links to Postverta in the starting map, and have those links be re-established by the event that removes the `hidden` tag. As there will be no links to the system, the `hidden` tag will effectively act like an `invisible` tag.

*preference*
Option 1 is the preferred option, as I can see situations where the currently described `hidden` tag could have very interesting uses for encouraging gameplay and rendering exploration via JD more challenging, without actually locking out an area like the `invisible` functionality would.

Option 2 Is reasonable. As an advantage, it already exists in the code and just needs to be adjusted to provide all the intended functionality. As a downside, it loses the usefulness for creating areas that are harder to explore, but can be "revealed" through careful exploration without needing events.

Option 3's main strength is that I can implement it immediately, since that is just a content change, and requires no changes to the mechanics. Thus it does not require any work on the part of the Devs.

**Background**
This bug was revealed through testing for PR #5732 , which itself is a fix for issue #5731 . Testing revealed that - should someone manage to gain access to any of the other systems in the cluster introduced in the Cognizance PR, they could then follow the links to Postverta and access it. This is contrary to the story and the lore, which specifically revolves around the system being completely cut-off from *all* surrounding systems, not just concealed from the currently existing systems. Just to be clear, this is an important plot point that will be a recurring detail in future missions; not just a gating mechanism.

**Conclusion**
[...]

PR #5735, opened 2021-02-20
Fix: Making Postverta Properly Hidden

Zitchas:
**Bugfix:** This PR addresses the gameplay element that revealed issue #5734 

## Problem:

- Postverta is described in the mission and lore as a system that appears where there wasn't one previously. This description comes from the Remnant, who have the best scanners available, long experience in the Ember Waste, a very strong astronomical program, and a keen awareness of their environment.
- Currently, this is implemented by giving Postverta the `hidden` tag, which renders it invisible and thus can't be jumped to via JD.
- Unfortunately, if the player can somehow gain access to any of the other Cognizance systems (such as Egeria, Prosa, Statina, or Diespiter) they will be able to follow the hyperdrive links to reveal Postverta, and thereafter be able to see it and jump to it whenever they want.
- This placement is intentional, but the ability for Postverta to be discovered is not. The bulk of the Cognizance systems are intended to fall into the category of "hard to get to, but not-impossible. The player could conceivably reach here at some point in the future." Postverta itself is supposed to *not be there.* This is not "just" a gate keeping mechanism, but rather an actual plot point that will be investigated in future missions. Likewise, this is *not* simply a case of "Postverta is hard to see."

## Fix Details
Changes:
- Removes all links between Postverta and other systems.
- Adjust the Postverta Reveal event to create those links
- Also unvisits all systems changed by this event.

## Testing Done
Repeatedly tested this flying a starling with both hyperdrive and jump drive. Verified that every link is a bidirectional hyperdrive link.

## Save File
This save file can be used to verify the bugfix. The bug will occur when using {{insert commit hash / version}}, and will not occur when using this branch's build.
{{attach a save file that can be used to verify your bugfix. It MUST have no plugin requirements}}

### DIFF
--- a/data/remnant/remnant missions.txt
+++ b/data/remnant/remnant missions.txt
@@ -3948,8 +3948,18 @@ mission "Remnant: Cognizance 13"
 	
 
 event "remnant: postverta reveal"
-	system Postverta
-		remove hidden
+	unvisit Caeculus
+	unvisit Egeria
+	unvisit Levana
+	unvisit Prosa
+	unvisit Statina
+	system Caeculus
+		add object "Wormhole Barren Beta"
+			sprite planet/wormhole-red
+			distance 2400
+			period 391
+	system Diespiter
+		add link Postverta
 	system Egeria
 		add object "Wormhole Barren Alpha"
 			sprite planet/wormhole-red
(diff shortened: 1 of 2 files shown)


data/remnant/remnant missions.txt
```diff
@@ -3948,8 +3948,20 @@ mission "Remnant: Cognizance 13"
...
+	unvisit Caeculus
+	unvisit Diespiter
+	unvisit Egeria
+	unvisit Levana
+	unvisit Prosa
+	unvisit Statina
+	link Postverta Diespiter
+	link Postverta Prosa
+	link Postverta Statina
+	system Caeculus
```
Amazinite, line 3960: I'd keep the system order where I had it. You've changed it to "five systems are being changed, here they are in alphabetical order," whereas previously I had "five systems are being changed in three different ways, here are those three ways [removing hidden, adding wormhole alpha, adding wormhole beta]."

Amazinite (changes requested):

data/remnant/remnant missions.txt
Zitchas: Ah, so that was the order. Sure, I will switch sorting orders.

PR #5736, opened 2021-02-20
feat: Integration Test Framework: Add input loading for key, pointer(mouse) and command.

petervdmeer:
**Feature:** This PR implements the key and command input as discussed in issue #4308

## Feature Details
This adds the loading for (keyboard) key, (mouse) pointer and command inputs.
For pointers only the loading was added, while keys and commands can already be used in tests.

## UI Screenshots
N/A

## Usage Examples
See data/tests/tests_framework.txt for usage examples.

## Testing Done
Added a new savegame test-dataset and a number of testcases.

## Performance Impact
No impact expected.


data/tests/tests_framework.txt
```diff
@@ -50,3 +50,181 @@ test "Test-Framework - Conditions Arithmetic and Loops"
...
+		# Set desired travel plan.
+		navigate
+			travel "Alpha Centauri"
+			travel "Sol"
+			travel "Alpha Centauri"
+		# Launch.
+		input
+			key d
+		input
+		input
```
tehhowch, line 215: What do empty inputs signify?

source/Test.cpp
```diff
@@ -74,6 +74,34 @@ namespace{
 			})
 			+ "\", or \"" + lastValidIt->second + '"';
 	}
+	
+	// Send an SDL_event to on of the UIs.
```
tehhowch, line 78:
```suggestion
	// Send an SDL_event to one of the UIs.
```

source/Test.cpp
```diff
@@ -74,6 +74,34 @@ namespace{
...
+	bool EventToUI(UI &menuOrGamePanels, const SDL_Event &event)
+	{
+		return menuOrGamePanels.Handle(event);
+	}
+	
+	// Send an keyboard input to one of the UIs.
+	bool KeyInputToUI(UI &menuOrGamePanels, const char* keyName, bool shift=false, bool ctrl=false, bool alt=false)
```
tehhowch, line 85: Rather than a number of boolean flags, use a bit field, like how the game tests for shift, Ctrl, alt, etc.

source/Test.cpp
```diff
@@ -332,8 +436,39 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+				{
+					// We need to send the command through the top gamepanel, and it needs to be active.
+					if(gamePanels.IsEmpty())
+						Fail(context, player, "panel with engine not present, and can only send commands to the engine");
+					
+					if(gamePanels.Root() != gamePanels.Top())
+						Fail(context, player, "engine not active due to panel on top, and can only send commands to the engine");
+					
+					// Both get as well as the cast can result in a nullpointer. In both cases we
+					MainPanel* mainPanel = dynamic_cast<MainPanel*> (gamePanels.Root().get());
```
tehhowch, line 449:
```suggestion
					auto mainPanel = dynamic_cast<MainPanel *>(gamePanels.Root().get());
```

source/Test.cpp
```diff
@@ -332,8 +436,39 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+				if(stepToRun.command)
+				{
+					// We need to send the command through the top gamepanel, and it needs to be active.
+					if(gamePanels.IsEmpty())
+						Fail(context, player, "panel with engine not present, and can only send commands to the engine");
+					
+					if(gamePanels.Root() != gamePanels.Top())
+						Fail(context, player, "engine not active due to panel on top, and can only send commands to the engine");
+					
+					// Both get as well as the cast can result in a nullpointer. In both cases we
```
tehhowch, line 448: I think this comment is missing some text 😉

data/tests/tests_framework.txt
```diff
@@ -50,3 +50,181 @@ test "Test-Framework - Conditions Arithmetic and Loops"
...
+		# Enter trading
+		input
+			pointer
+				# Go from middle of the screen (to allow many resolutions)
+				X 0.5 relative
+				Y 0.5 relative
+				# Go for the center of the trading button
+				X -411
+				Y 92
+				click left
```
tehhowch, line 122: How robust is this going to be to e.g. interfaces.txt changes? If a plugin shifts things around, will all of its users be unable to use automated tests without seeing these fail?

tehhowch (commented):
It's probably a requirement to develop a system for reusable test sequences in order to compose robust tests that are also readable and comprehensible.

The current direction seems to lead into an entirely imperative test sequence, where the tests must explicitly control every aspect, and a small change (like animating panel transitions and disabling certain input handling during said transitions) will result in a slew of failing tests that broke because of something they didn't really care about.

I'm not opinionated on how you approach the solution to this tight coupling, but resolving it is a prerequisite to making test contributions a requirement for new code.

data/tests/tests_framework.txt
petervdmeer: The ratio of "the risk of the test breaking" vs "the benefit of testing mouse-clicks here" is indeed not favorable for this test. This part of the game is better tested using key-presses only. I'll remove this test.

We could increase the stability of the tests a little by giving tests their own data(sets) files, but in general it is possible that changes to the game also require changes to the relevant tests (tests test against current expected behavior, which can be different from future expected behavior).

data/tests/tests_framework.txt
petervdmeer: I was debugging something here and expected that the game needed more than 1 step to perform the complete launch.
The test-loops return control for 1 game-step after an input statement or when encountering a branch for a second time. An input statement can be used to let the game to do a step.

petervdmeer: I just checked, the game does indeed need more than 1 step to perform the launch (in the current code-base).

petervdmeer:
> It's probably a requirement to develop a system for reusable test sequences in order to compose robust tests that are also readable and comprehensible.
>
> The current direction seems to lead into an entirely imperative test sequence, where the tests must explicitly control every aspect, and a small change (like animating panel transitions and disabling certain input handling during said transitions) will result in a slew of failing tests that broke because of something they didn't really care about.

I also noticed that some test-constructs (like the "loading screen", "load", "depart" sequence) get used fairly often. For a future PR I planned to add some form of functions/injectable-sequences or the ability to let tests call other tests. This would allow such common sequences to be placed on a single place and called by name instead of getting copied to each test-case.

Older versions of the test-code had explicit keywords `load` and `launch` for those common actions and just performed low-level calls bypassing the UI, but I removed those based on [an earlier review comment](https://github.com/endless-sky/endless-sky/pull/4370#discussion_r363046193) I feel that the functions/injectable-sequences/tests-calling-tests is probably a better approach.

data/tests/tests_framework.txt
petervdmeer: Added a comment to explain this better.

petervdmeer:
I made a start with adding a CALL functionality to allow tests to re-use other tests to handle common functionality.

petervdmeer:
Merged 2 navigation tests from Terin into this PR.

petervdmeer:
> It's probably a requirement to develop a system for reusable test sequences in order to compose robust tests that are also readable and comprehensible.

I added the "call" functionality that allows a (small) single test (like for example savegame loading or launching from a planet) to be re-used by other (larger) tests.

source/Test.cpp
```diff
@@ -84,6 +107,70 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
 
 
 
+void Test::TestStep::LoadInput(const DataNode &node)
+{
+	for(const DataNode &child : node)
+	{
+		if(child.Token(0) == "key")
+		{
+			for(int i=1; i < child.Size(); ++i)
```
tehhowch, line 116:
```suggestion
			for(int i = 1; i < child.Size(); ++i)
```

source/Test.cpp
```diff
@@ -84,6 +107,70 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
...
+		else if(child.Token(0) == "pointer")
+		{
+			for(const DataNode &grand: child){
+				if(grand.Token(0) == "X")
+				{
+					if(grand.Size() < 2)
+						grand.PrintTrace("Warning: Pointer X axis input without coordinate:");
+					else{
+						XValue = grand.Value(1);
+					}
```
tehhowch, line 139:
```suggestion
					else
						XValue = grand.Value(1);
```

source/Test.cpp
```diff
@@ -84,6 +107,70 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
...
+				}
+				else if(grand.Token(0) == "Y")
+				{
+					if(grand.Size() < 2)
+						grand.PrintTrace("Warning: Pointer Y axis input without coordinate:");
+					else{
+						YValue = grand.Value(1);
+					}
```
tehhowch, line 147:
```suggestion
					else
						YValue = grand.Value(1);
```

source/Test.cpp
```diff
@@ -84,6 +107,70 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
...
+				}
+				else if(grand.Token(0) == "click")
+					for(int i=1; i < grand.Size(); ++i)
```
tehhowch, line 150:
```suggestion
					for(int i = 1; i < grand.Size(); ++i)
```

source/Test.cpp
```diff
@@ -84,6 +107,70 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
...
+					{
+						if(grand.Token(i) == "left")
+							clickLeft = true;
+						else if(grand.Token(i) == "right")
+							clickRight = true;
+						else if(grand.Token(i) == "middle")
+							clickMiddle = true;
```
tehhowch, line 157: Is it expected that multiple clicks can be submitted on the same step?
```
click left middle right
```

source/Test.cpp
```diff
@@ -84,6 +107,70 @@ Test::TestStep::TestStep(Type stepType) : stepType(stepType)
...
+						else
+							grand.PrintTrace("Warning: Unknown click/button:");
```
tehhowch, line 159: The trace will print the entire line, which per above comment may contain multiple click buttons. If multiple are allowed, include the unknown token to improve the message.

source/main.cpp
```diff
@@ -272,8 +272,8 @@ void GameLoop(PlayerInfo &player, const Conversation &conversation, const string
 		((!isPaused && menuPanels.IsEmpty()) ? gamePanels : menuPanels).StepAll();
 		
 		// All manual events and processing done. Handle any test inputs and events if we have any.
-		if(testContext.testToRun)
-			testContext.testToRun->Step(testContext, menuPanels, gamePanels, player);
+		if(!testContext.testToRun.empty())
+			testContext.testToRun[testContext.testToRun.size() - 1]->Step(testContext, menuPanels, gamePanels, player);
```
tehhowch, line 276: `.back()`?

tehhowch (commented):
Some style nits suggestions included.

Thanks for working on this!

data/tests/tests_framework.txt
```diff
@@ -50,3 +50,115 @@ test "Test-Framework - Conditions Arithmetic and Loops"
...
+	description "Test-Framework - Test for calls to other functions"
+	sequence
+		apply
+			"test: callLoopControl" = "test: callLoopControl" + 1
+		branch loopBreak
+			"test: callLoopControl" > 10
+		# Use recursive call to get callLoopControl above 10.
+		call "Test-Framework - Recursive Calls"
+		label loopBreak
+
```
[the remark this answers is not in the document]
petervdmeer: That depends a bit on the bug that would trigger the issue, but the most likely error-state would be a continuously growing test-call-callstack with "Test-Framework - Recursive Calls" functions continuously recursively calling itself.

I will check if I can add a timeout so that it would fail on the timeout instead of looping forever.

source/Test.cpp
```diff
@@ -128,6 +215,16 @@ void Test::LoadSequence(const DataNode &node)
...
+			case TestStep::Type::CALL:
+				if(child.Size() < 2)
+				{
+					status = Status::BROKEN;
+					child.PrintTrace("Error: Invalid use of \"call\" without name of called (sub)test:");
+					return;
+				}
+				else
+					step.nameOrLabel = child.Token(1);
+				break;
```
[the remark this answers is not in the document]
petervdmeer: The current approach gives a "test not found" at test-execution when a test was missing, while registering with GameData would give a "test not loaded" (but found since we registered it) during test-execution. There could be a minor performance benefit because the translation from string-to-test would happen at test-loadtime instead of test-runtime, but I don't think it makes a lot of difference.

Let's keep this as-is for now. (We can always improve this in a later PR, and then we might want to include the INJECT referencing below as well.)

data/tests/tests_framework.txt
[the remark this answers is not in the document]
petervdmeer: I added an initialize part before the test which sets a watchdog/timeout.

data/tests/tests_framework.txt
```diff
@@ -50,3 +50,115 @@ test "Test-Framework - Conditions Arithmetic and Loops"
...
+	status active
+	description "Tests if the in-game panels can be controlled by keypresses"
+	sequence
+		# Create/inject the savegame and load it.
+		inject "Three Earthly Barges"
+		call "Load First Savegame"
+		# Enter and leave jobs
+		input
+			key j
+		input
```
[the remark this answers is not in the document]
petervdmeer: I added a TODO, since the functionality to inspect panels is not present in the framework yet.

data/tests/tests_framework.txt
```diff
@@ -50,3 +50,115 @@ test "Test-Framework - Conditions Arithmetic and Loops"
...
+
+test "Test-Framework - Calling other functions"
+	status active
+	description "Test-Framework - Test for calls to other functions"
+	sequence
+		call "Test-Framework - UI Keyboard controls"
+		call "Test-Framework - Empty Test Sequence"
+		call "Test-Framework - Empty Testcase"
+		call "Test-Framework - Conditions Arithmetic and Loops"
+
```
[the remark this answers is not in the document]
petervdmeer: Should we already move the framework tests out (to a separate plugin) in this PR? Or is it okay to do this in a later PR (when the test-framework itself is mostly finished?)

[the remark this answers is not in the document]
tehhowch: Later is fine

source/Test.cpp
```diff
@@ -329,33 +457,65 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
+					// We need to send the command through the top gamepanel, and it needs to be active.
+					if(gamePanels.IsEmpty())
+						Fail(context, player, "panel with engine not present, and can only send commands to the engine");
+					
+					if(gamePanels.Root() != gamePanels.Top())
+						Fail(context, player, "engine not active due to panel on top, and can only send commands to the engine");
+					
+					// Both get as well as the cast can result in a nullpointer. In both cases we
+					// will fail the test, since we expect the MainPanel to be here.
+					auto mainPanel = dynamic_cast<MainPanel *>(gamePanels.Root().get());
```
thomasballinger, line 474: Do I understand right that the goal of this dynamic_cast is to return a null pointer if `gamePanels.Root().get()` returns either null or an instance of a panel other than MainPanel? Would it be ok to ignore the wrong type failure case and use a static_cast?

I'm interested because I had been compiling [without cpp runtime support](https://emscripten.org/docs/optimizing/Optimizing-Code.html#c-rtti) as an optimization until this first dynamic_cast in the codebase. This isn't a reason to change it — compiling with Emscripten is explicitly not a goal — but if I understand right I can open a PR since it's also the preferred style in this codebase to avoid dynamic_cast.

petervdmeer: Ok for me to change this to a static cast.

petervdmeer: And if this is the only location that uses run-time type support, then feel free to open a second PR to compile without run-time support.

thomasballinger: Great, thanks @petervdmeer, I opened https://github.com/endless-sky/endless-sky/pull/6210 to remove.

I don't know that we want to add `-fno-rtt` and would want someone with more c++ experience to make that suggestion — but if more experienced folks want that then that'd be great!

PR #5745, opened 2021-02-23
Fix: Prevent jump commands from being cleared while landing

### LINKED ISSUE #5744: Bug: Autopiloting is disabled traveling through wormholes
**Describe the bug**
Introduced by #5733 / 26c1490, when setting up a route that travels through wormholes, your autopilot will stop after having landed on a wormhole, requiring you to hit jump/land to continue.

**To Reproduce**
1. Loading a pilot with a Ultima Thule explored/Terminus explored with a quantum keystone.
2. Set a route to a system on the other side.
3. Hit jump to begin the autopilot sequence through that route.
4. Observe the autopilot being disabled after traveling through the wormhole.

**Expected behavior**
The autopilot will continue to move the ship to the next jump/wormhole after traversing the first one.

 - OS: Windows 10 Pro
 - Game Source: Self-compiled
 - Version: 774cd9f

**Link to save file**
Pilot with route set through Terminus
[Wormhole Travel.txt](https://github.com/endless-sky/endless-sky/files/6025599/Wormhole.Travel.txt)

Terin:
**Bugfix:** This PR fixes #5744.

## Fix Details
Thanks to Vitalchip, jumping has been removed from the list of commands cleared when in the process of landing. This keeps the jumping command intact while traveling through wormholes, while still not triggering the message fixed in #5733.

**TODO:** Verify possible regression that Vitalchip experienced, where a jump can be queued up on landing, does not occur on other computers.

## Testing Done
Traversed wormhole route with only a single button press. Checked that "Disengaging autopilot" does not trigger while pressing "Jump" while landing.

## Save File
This save file can be used to verify the bugfix. The bug will occur when using 774cd9f, and will not occur when using this branch's build.
[Wormhole Pilot.txt](https://github.com/endless-sky/endless-sky/files/6025599/Wormhole.Travel.txt)

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -342,7 +342,7 @@ void AI::UpdateKeys(PlayerInfo &player, Command &activeCommands)
 	autoPilot |= activeCommands;
 	// While the flagship is landing, it cannot do most actions.
 	if(flagship && flagship->IsLanding())
-		autoPilot.Clear(Command::JUMP | Command::LAND | Command::BOARD | Command::STOP);
+		autoPilot.Clear(Command::LAND | Command::BOARD | Command::STOP);
 	else if(activeCommands.Has(AutopilotCancelCommands()))
 	{
 		bool canceled = (autoPilot.Has(Command::JUMP) && !activeCommands.Has(Command::JUMP));


Rob59er:
Works as described.

Terin:
Are you able to reproduce the issue that Vitalchip found? Where if you press jump while in the landing animation, it queues up a jump right when you take off?

I am unable to reproduce that on either my Windows or Linux machines.

Rob59er:
I was able to reproduce it on the Steam version running 0.9.12, not on the most recent commit, nor this pull request.

vitalchip:
I'm still getting the jump on departure issue. I've typed the code chages from PR 5744 into my build again and as merged there is no queued jump when I depart ( after pressing J during landing zoom ) but when I remove the  JUMP command from the autopilot.clear list I do get a 'stored' jump. To be honest, this is what I'd expect as the jump command isn't being cleared.

I'll admit to being somewhat sleep deprived this week so I hop for some lattitute if I'm making some obvious error here.

I also wonder why, since we're trying to prevent player inputs during the final stage of landing.. we don't just do that directly, in engine?

landing

Is there any player input that we need during the landing zoom?  I can't think of one.

tatami4:
> Is there any player input that we need during the landing zoom? I can't think of one.

The player _could_ activate cloak or gather escorts, for example.

tehhowch:
An avenue to investigate would be to avoid adding the various commands to the existing autopilot commands, rather than clearing them after adding in new commands.

Terin:
Closing since the original change was reverted, and we'll likely make sure from the jump it doesn't have this issue when re-implemented

PR #5748, opened 2021-02-25
feat (internal): Move the String Interner from Dictionary to its own class

petervdmeer:
**Feature:** This PR moves the String Interner from Dictionary to its own class

## Feature Details
The String Interner inside the dictionary class can also be usefull outside the dictionary class for strings like for example the ship types, bay types and condition-strings.

## UI Screenshots
N/A

## Usage Examples
See Dictionary.cpp for usage of the StringInterner.

## Testing Done
I added a new unit-test for the StringInterner class.

## Performance Impact
I don't expect any performance impact on the lookup speed in the Dictionary class and the interning speed was already listed as slower in favor of faster lookup speed (and should not be a lot slower, it only adds a static method call if the element is not found).

If we use this functionality for ship-types and bay-types, then the comparisons there might gain a minor speed improvement, since we will be able to use char-pointer comparisons instead of string-comparisons in operations like matching (carried)ships to bays.


tehhowch:
One issue is that exposing the interning `set` for use by others means it will take on more values, which means slower behavior for _all_ consumers the more different consumers there are. This will become especially true when those consumers are on multiple threads and the lock contention rises.

I'm not sure I see a benefit to doing this, especially when the programmer cost of providing this ability to other scenarios is pretty low (two very short static methods).

What would be more useful would be adding benchmarks for the Dictionary class prior to its modification. (You'll want to extend the class with a static method that clears out and shrinks anything with static lifetime so that the benchmarks are reliable.)

petervdmeer:
I agree that interning a new string will be slower when the interner holds more different values, but usage of the interned strings should be at the same speed as before, since the char-pointer that represents the interned string just points to chars and doesn't care which set is the official container.

Interning of strings would typically happen while the game is loading, so if we are optimizing for "faster interning behavior due to smaller sets" then we are mostly optimizing for load-time speed, we are optimizing the [slower construction time](https://github.com/endless-sky/endless-sky/blob/ae40b10f03738cc4d82cfff8c25ab5a7f005ddce/source/Dictionary.h#L23) of Dictionaries.

Lock contention could indeed become worse and could already be an issue today if multiple ships and/or outfits are loading in parallel. If most interned strings typically get used at many places (for example once per ship), then we could reduce the impact by adding a read-check if a string was already interned and combine it with [many-readers/single-writer locking](https://stackoverflow.com/questions/19915152/c11-multiple-read-and-one-write-thread-mutex) to increase parallelism (by having less exclusive (write) locks at the cost of more expensive writing when writing is actually needed).

I see some benefits in using a single interner;
- Usage of string-interning anywhere will just be a oneliner. (Just call the intern function whenever you want to intern some string.)
- Reduction of duplicate code (even if it is just two very short static methods for every scenario).
- If there is overlap in values for different scenarios, then the usage of a single interner class will result in a smaller memory footprint compared to scenarios each having their own interner.
- (interned)string-variables from different scenarios can be tested for equality just by comparing the char-pointer itself instead of doing strcmp. (Which could be faster at run-time.)

I'm happy with all the possibilities we are getting at automated function and performance testing. My time for computer-games is a bit limited at the moment, so I don't know if I can look into benchmarking any time soon.

petervdmeer:
Closing this PR; there are currently a lot of other higher priority issues to address and as Tehhowch mentioned; we can always do a custom interner per situation where interning would be beneficial.

petervdmeer:
Re-opening; PR #7165 gives good reasons on why we would want to merge this PR or #7165.

petervdmeer:
Seems we also have [a branch by Amazinite](https://github.com/endless-sky/endless-sky/issues/6329#issuecomment-963768703) that is trying to improve performance around string comparisons.

Amazinite:
That branch was (in part) opened as #6804. I've since closed that PR, but there are parts of it that we could probably make use of even on top of a performance improvement to the Dictionary class.

quyykk:
> while my PR allowed re-use of the char-pointer for doing the comparisons without needing a full string compare.

How so? you would still need to do a string compare because the interned string's address isn't known at runtime. Or am I missing something?

petervdmeer:
> > while my PR allowed re-use of the char-pointer for doing the comparisons without needing a full string compare.
> 
> How so? you would still need to do a string compare because the interned string's address isn't known at runtime. Or am I missing something?

Sorry, I left out an important detail; when comparing 2 interned strings, then the string's addresses should be the same. This would only work when comparing 2 interned strings, if one is interned and the other is not, then a full string compare would still be needed.

quyykk:
> for use by others means it will take on more values, which means slower behavior for all consumers the more different consumers there are. 

Yeah I definitely agree, we should only intern strings for which it makes sense to do so. Let's just use it for outfit attributes?

> [many-readers/single-writer locking](https://stackoverflow.com/questions/19915152/c11-multiple-read-and-one-write-thread-mutex)

yet another reason use C++17 ([std::shared_mutex](https://en.cppreference.com/w/cpp/thread/shared_mutex))

petervdmeer:
> > for use by others means it will take on more values, which means slower behavior for all consumers the more different consumers there are.
> 
> Yeah I definitely agree, we should only intern strings for which it makes sense to do so. Let's just use it for outfit attributes?

That is fine, we can add other interners later, for other sets of strings.

> > [many-readers/single-writer locking](https://stackoverflow.com/questions/19915152/c11-multiple-read-and-one-write-thread-mutex)
> 
> yet another reason use C++17 ([std::shared_mutex](https://en.cppreference.com/w/cpp/thread/shared_mutex))

Agreed. The code at [WikiPedia](https://en.wikipedia.org/wiki/Readers%E2%80%93writer_lock#Using_a_condition_variable_and_a_mutex) doesn't look too complex though, I'm sure that we can also implement it ourselves using c++11 primitives if we need to.

source/StringInterner.h
```diff
@@ -0,0 +1,29 @@
...
+*/
+
+#ifndef STRING_INTERNER_H_
+#define STRING_INTERNER_H_
+
+#include <string>
+
+
+
+class StringInterner
```
quyykk, line 20: This should definitely have a comment.

tests/unit/src/test_stringInterner.cpp
```diff
@@ -0,0 +1,64 @@
...
+		std::string ahAh = "ah ah";
+		std::string blaBla = "bla bla";
+		std::string daDa = "da da";
+		const char *blaBlaPtr = nullptr;
+
+		WHEN( "the string is interned" ) {
+			blaBlaPtr = StringInterner::Intern(blaBla);
+			THEN( "it still represents the same string" ) {
+				// This is a string comparison (since blaBla is a string).
+				CHECK(blaBla == blaBlaPtr);
```
quyykk, line 38: The whitespace is missing after and before the parens. `CHECK( bla... )`.

tests/unit/src/test_stringInterner.cpp
```diff
@@ -0,0 +1,64 @@
...
+			}
+			THEN( "interning twice results in the same char pointer twice" ) {
+				const char *blaBlaPtr2 = StringInterner::Intern(blaBla);
+				// This is a pointer comparison (since both arguments are char-pointers).
+				CHECK(blaBlaPtr2 == blaBlaPtr);
+			}
+			THEN( "interning other strings still results in the same char pointer" ) {
```
quyykk, line 45: You don't have any tests that verify that different strings have different interned pointers.

source/StringInterner.cpp
```diff
@@ -0,0 +1,41 @@
...
+Foundation, either version 3 of the License, or (at your option) any later version.
+
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+#include "StringInterner.h"
+
+#include <cstring>
```
quyykk, line 15:
```suggestion
```

unused include

tests/unit/src/test_stringInterner.cpp
petervdmeer: I added a test for this.

source/StringInterner.h
petervdmeer: I added one.

mike-f1:
I don't know much about "other scenario", so probably I'm going to say something totally wrong here...
Anyway, my 2 main doubts are:
* doing 7685 I saw that `Dictionary::Get` is used only(?) with string literals, like `something.Get("attribute_I_want_to_ask")` and those string literals can never be "interned", so there's a need to change call sites. If that is going to be `something.Get(StringInterner::Intern("attribute_I_want_to_ask"))` I don't see how the full comparison can be skipped: wouldn't it get the string literal every time? And wouldn't it have to deal with the comparison every time then?
* I think that a hidden advantage of having different methods (one is read/write and the second is read-only) is that that read-only doesn't need a lock, which is by itself little time but if this is going to be on multiple threads, then that lock can slow down things a lot... There's a way to keep read-only/rw methods?

petervdmeer:
> * doing 7685 I saw that `Dictionary::Get` is used only(?) with string literals, like `something.Get("attribute_I_want_to_ask")` and those string literals can never be "interned", so there's a need to change call sites. If that is going to be `something.Get(StringInterner::Intern("attribute_I_want_to_ask"))` I don't see how the full comparison can be skipped: wouldn't it get the string literal every time? And wouldn't it have to deal with the comparison every time then?

You are correct, avoiding the interning would require a different pattern, something like:
```
const static char* ATTRIBUTE_TO_ASK = StringInterner::Intern("attribute_I_want_to_ask");
something.Get(ATTRIBUTE_TO_ASK);
```
Where the const char* pointers are initialized at load-time. (I'm not sure if they should be static.)

> * I think that a hidden advantage of having different methods (one is read/write and the second is read-only) is that that read-only doesn't need a lock, which is by itself little time but if this is going to be on multiple threads, then that lock can slow down things a lot... There's a way to keep read-only/rw methods?

The multiple-readers/single-writer feels more appropriate, where we will not have any slowdowns in reading, as long as there is no writer updating the attributes (and writing should typically only happen during loading).
Or is there a way to perform reading in a fully threadsafe way, even when a writer is active?

mike-f1:
That pattern (which, I should say, led to maintenance issue: in general my feeling is that it should be used only where really necessary) can be enhanced changing the returned type: instead of a pointer (`char *` which is 64bits) you can make a (constexpr-essed) wrapper around an `uintXX_t` where XX is the needed size (eg if keys are less than 65536 then you can use a `uint16_t`), so you can pack more of them in a class. This way you are also using the type system to prevent type mismatches. Obviously, you need to change the call sites :disappointed:

In this case, I suggest something like:

string_wrapper attribute_I_want_to_ask StringInterner::Intern("attribute_I_want_to_ask");

> (I'm not sure if they should be static.)

Those I described yes, and I think also yours because having each instance with a different set of 'shortcuts'
is really difficult to track (what if in two different context you redefine a value?)

petervdmeer:
@mike-f1 Yes, I believe that using a different data-type will be useful for the "in flight attributes" that we can implement as part of https://github.com/endless-sky/endless-sky/issues/8710 .

I also feel that using a generic StringInterner has value in the mean-time, for things like the attributes not used in flight. (And transforming the type there might be a next step, I'm just not sure yet of the costs-vs-benefits ratio for the non-in-flight attributes.)

petervdmeer:
I updated the code to use a multiple-readers/single-writer mutex.

tibetiroka (approved):

tibetiroka:
I was looking to extract a templated FlatMap class from Dictionary (somewhat similarly to #5699, just with easier migration to std::flat_map in even newer c++ standards), for use in #9114, amongst other things. Is this PR in its final state now?

petervdmeer:
> Is this PR in its final state now?

I consider this PR ready for use in #9114. (And the comments from tehhowch and quyykk above suggest to me that they are also okay with using this PR for #9114).

tibetiroka:
Nice. I'll merge it in the unstable release.

source/StringInterner.h
```diff
@@ -0,0 +1,36 @@
...
+#include <string>
+
+
+
+// This class stores a set of interned strings. Interning can be a slow operation during string creation/interning, but
+// it will allow fast char-pointer based comparisons when comparing two interned strings (because interning ensures that
+// each interned string only appears once in the set). Full string compares will still be needed when comparing interned
+// strings to non-interned strings.
+class StringInterner
+{
```
TomGoodIdea, line 28:
```suggestion
class StringInterner {
```

PR #5752, opened 2021-02-27
Add 'invisible' system tag

yjhn:
**Feature:** This PR implements the feature request detailed in issue #5734

## Feature Details
`invisible` systems cannot be seen on map or jumped to

NOTE: this PR does not prevent you from using wormholes to go the invisible system.

## Testing Done
Marked Arculus `invisible` and it wasn't drawn on map or reachable via hyperspace. I didn't test if the attribute is removed correctly, but it should work.

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -8012,6 +8012,7 @@ system Diespiter
 	attributes "ember waste" "inaccessible"
 	habitable 534.375
 	haze _menu/haze-red
+	link Postverta
 	link Statina
 	asteroids "small rock" 65 3
 	asteroids "medium rock" 14 1
@@ -19454,12 +19455,15 @@ system Porrima
 			period 11.4255
 
 system Postverta
-	hidden
+	invisible
 	pos 297 269
 	government Uninhabited
 	attributes "ember waste" "inaccessible"
 	habitable 135
 	haze _menu/haze-red
+	link Diespiter
+	link Prosa
+	link Statina
 	asteroids "medium rock" 4 2
 	asteroids "large rock" 4 3
 	asteroids "medium metal" 52 1
@@ -19632,6 +19636,7 @@ system Prosa
 	attributes "ember waste" "inaccessible"
 	habitable 2372.76
 	haze _menu/haze-red
+	link Postverta
 	link Vaticanus
 	asteroids "medium metal" 11 2
 	asteroids "large metal" 2 2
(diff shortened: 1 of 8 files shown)


yjhn:
> Isn't the actual goal of this to make the systems be unreachable until the tag has been removed?

Yes, the goal is to make the system with invisible` tag unreachable.
I didn't know the tag was removed, I haven't been keeping up with recent developments.
If it indeed is removed, then this PR can be closed as it is not useful anymore.

> Also just to be sure, clicking on the map at the exact location of the system to set a destination to it will not work?

Haven't checked this scenario, but as far as I know this does not (and should not) work.

Hurleveur:
Sorry I wasn't clear enough, no it won't get removed, what I meant is since what this does is making the system unreachable, that could be an appropriate name for it, instead of invisible.

Hurleveur (approved):
Code wise this looks good, I was just thinking it could be named `unreachable` instead of `invisible`, but I'm not good on naming stuff.
Also probably needs some testing, but syntax wise I dont see any issues, and that is what I worry about which is why I am approving this.
The change in links should probably be checked; what would be good is to test this with 
the Postverta mission chain

yjhn:
>  I was just thinking it could be named unreachable instead of invisible

The system is not actually unreachable - you can reach it via wormholes if it has any.

warp-core:
The changes to the Postverta and associated events seem kind of unnecessary, and are also super out of date.
The change to Zubenelhakrabi and associated event are totally unnecessary: that system is intended to only be accessible by jump drive so using this new property doesn't reduce changes that need to be made later.

The name is not great (`invisible` and `hidden` are synonyms), and there is a lot of overlap with the existing `hidden` property.

Perhaps instead of doing everything `hidden` does and then also hiding hyperspace links, a new property that *only* hides hyperlinks could be created? That is, the system would still show up on the map if it is within the visual range, and would still be reachable by jump drive (assuming jump ranges are not limited) but any hyperlinks connecting to it would not be visible or usable while the property is present on the system unless it has already been visited (which could be achieved either by jump drive or wormhole, or possibly `relocation`, when that's in.)

Having both `hidden` and the property I suggest here would result in the behaviour your `invisible` property currently adds.

Amazinite:

Marking this PR as orphaned as there hasn't been activity from OP in some time, but not closing because it's a minor feature that we should have. Either someone should open a new PR with these changes finalized, or one of the devs could finalized this PR themselves.

Amazinite:
Same auth error on this one as on #5654.

Issue #5751, opened 2021-02-26
Add non-directional weapon hazard/projectile "impact effect".


petervdmeer:
**Is your feature request related to a problem? Please describe.**
The `"hit effect"` for projectile based weapons creates visual effects at the position where a projectile hits/touches a ship.
The `"hit effect"` for hazard based weapons creates visual effects (CreateSparks) all over a ship when hitting/touching.

**Related Issue Links**
I was looking at #3878 when I noticed this difference.
PR #4931 introduced Hazards, so it also seems related.

**Describe the solution you'd like**
This request is to change the hazard type `"hit effect"` to `"impact effect"` to allow both projectile weapons and hazards to specify both effects.
This could for example be used by towing weapons to create some sort of grid over the ship that is being towed.

**Describe alternatives you've considered**
I'm not sure about the name `"impact effect"`, but couldn't come up with a better name (maybe `"effect on impacted target"`, but that is a bit long).

**Additional context**
Ship::TakeDamage() is used both for projectile weapons as for impact from ship explosions. If we remove some more differences between hazards and projectiles, then we might be able to also use Ship::TakeDamage() for hazards (and no longer need Ship::TakeHazardDamage()).

Amazinite:
`"spark effect"`?

petervdmeer:
`"spark effect"` sounds okay. `"spark hit effect"` might then even be better if we also want to add `"spark live effect"` and `"spark fire effect"`  someday.

tehhowch:
I think "target effect" is easier to reason about than "spark effect".

Possibly "affected target effect," or similar.

I believe there will be some visual discontinuities without extended code support, since there is currently no coupling of the visuals generated by effects to whatever caused the effect to be triggered. That is, if getting hit with a weapon with this new effect type creates a multi-frame visual, only the first frame is guaranteed to be drawn correctly. On all subsequent frames, the visual and the "source" of the visual, the ship or asteroid, can move with differing velocities.

I've noted (offline) that we can refactor the current hardcoded effect generation (disruption, slowing, etc) into generalized "effect emitters" that store particular effects to spawn and also the rate with which to spawn their respective visuals. A weapon would then be able to transfer an effect source to targets when they are hit, allowing e.g. the slowing caused by a system hazard to be differentiated from the slowing caused by a Detainer, or the aforementioned "towing grid" to always draw over the ship it affects.

petervdmeer:
I expect that eventually we might even want to have both:
[...]

PR #5753, opened 2021-02-27
Rename "hit effect" for hazards to "target effect".

petervdmeer:
**Feature:** This PR replaces the "hit effect" keyword for hazards by "target effect" as discussed in #5751

## Feature Details
This is just the rename to differentiate between "hit effect" (effect at point of impact, for projectile weapons) and "target effect" (as effects over a ship shape, for hazards) and some changes to make the hazard-weapon-terminology more in line with the projectile-weapon-terminilogy.
There are a number of other improvements being discussed in #5751, but the discussion was just started so I want to give a bit more time to discuss before creating other improvements.

## UI Screenshots
N/A (only the keyword changed)

## Usage Examples
See `data/hazards.txt` for the replaced keyword.

## Testing Done
Flown to Sagittarius A* and got my ship destroyed there in a cloud of nano particles.

## Performance Impact
None expected.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -21,7 +21,6 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include "text/Format.h"
 #include "GameData.h"
 #include "Government.h"
-#include "Hazard.h"
 #include "Mask.h"
 #include "Messages.h"
 #include "Phrase.h"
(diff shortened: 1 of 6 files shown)


Zitchas:
Sure, I don't see a problem with this. Keeps things a bit more separate.

petervdmeer:
> Sure, I don't see a problem with this. Keeps things a bit more separate.

The actual goal is to make things more integrated; enable the ability to use the hazard keyword on projectile weapons; and enable the ability to use the projectile keyword on hazards. (And that is only possible if the keywords don't overlap.)

I hope that this will also help us to re-use more code between projectile weapon damage and hazard damage. (Thus also reducing a number of separate code-paths to single code paths that handle both.)

source/Ship.cpp
```diff
@@ -3048,16 +3047,16 @@ int Ship::TakeDamage(const Projectile &projectile, bool isBlast)
 
 
 
-// This ship just got hit by the given hazard. Take damage according to what
+// This ship just got hit by an hazard. Take damage according to what
```
Amazinite, line 3050:
```suggestion
// This ship just got hit by a hazard. Take damage according to what
```

source/Ship.h
```diff
@@ -323,9 +322,9 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	// not necessarily its primary target.
 	// Blast damage is dependent on the distance to the damage source.
 	int TakeDamage(const Projectile &projectile, bool isBlast = false);
-	// This ship just got hit by the given hazard. Take damage according to what
+	// This ship just got hit by an hazard. Take damage according to what
```
Amazinite, line 325:
```suggestion
	// This ship just got hit by a hazard. Take damage according to what
```

Amazinite (commented):
So I'm assuming a future change might allow normal weaponry to also make use of this target effect.

petervdmeer:
> So I'm assuming a future change might allow normal weaponry to also make use of this target effect.

Yes, that is the next step.

PR #5763, opened 2021-03-01
Fix: Dealing with the Puffin availability and problems with the Lost in the EW mission

### LINKED ISSUE #5760: Quantum keystone mission offered with Puffin
**Describe the bug**
[If in Remnant space you sell your ship for a Puffin, then you are offered a quantum keystone as it is assumed you have none (as the Penguin has one 'built-in'. ]

**To Reproduce**
Steps to reproduce the behaviour:
1. Do the Remnant missions to scan void sprites and gain access to Puffin.
2. Sell your ship & buy the Puffin.
3. Land and take off a few time in Remnant space...
4. Get given a keystone when you don't need one.

Or use the save file below:
1. Take off and land on a populated Remant planet.
2. Keystone is offered..

**Expected behavior**
As the stock Puffin can transverse worm holes you shouldn't need to be offered a keystone.

 - OS: [Mint Linux (based on Ubuntu "Focal")]
 - Game Source: [self-compiled]
 - Version: [commit hash 78fc7d8"]
**Link to save file**
[https://pastebin.com/TLJbdMA9 ]

**Additional context**

Zitchas:
**Bugfix:** This PR addresses issue #5760 
fixes #5760 

## Fix Details
- Adds license requirement to the Penguin. (players can no longer buy restricted Remnant outfits, including Key Stones, via buying and stripping multiple Puffins)
- ~~Gives the puffin a Quantum Key Stone.~~ (already in game by Amazinite doing a commit)
- VS 2 now gives the player a Puffin on accept.
- Changed mission text slightly to remove the implication that Plume works for a separate entity from the Remnant as a whole.
- Reduced payment to take into account that the player is getting a Puffin for free.

## Testing Done
Repeatedly flying around with and without ships that have keystones, particularly the puffin.


tehhowch:
I don't think this is a great way to fix the issue.

Amazinite:
I'd rather just leave the key stone on the ship.

Zitchas:
Well, as far as that goes, @FranchuFranchu is working on a way to get what they term "pseudo conditions" available for testing, which could be used to test this more elegantly. (https://github.com/FranchuFranchu/endless-sky/tree/pseudo-variables)

Likewise, finishing off and implementing your flagship conditions PR (https://github.com/endless-sky/endless-sky/pull/4528) would also enable the tools to fix this properly.

Amazinite:
I'd still rather just leave the key stone on the ship.

Zitchas:
We could also leave it as-is. After all the Puffin *doesn't* have a keystone outfit. So the lady technically isn't wrong.

Amazinite:
The whole point of the mission is that the player is stuck in Remnant space, not that the player simply lacks a keystone, so that's not ideal.

Zitchas:
I'm kind of curious - if we leave it as is, how many people will trigger the mission? It really should only happen if someone sells their current ship in order to buy the puffin. Otherwise they'll still have their former flagship with whatever it used to get to the Remnant in the first place.

Amazinite:
If we want more people to experience a mission then we shouldn’t be doing that be erroneously causing it to offer, as, like I said, then the circumstances that the mission expects don’t match the player’s reality.

Zitchas:
OK, the original problem was that it was felt to be inappropriate to let people have unrestricted access to the Quantum Key Stone by being able to buy puffins to get it. I can see a few different solutions to this.

1. Have the puffins equipped with Quantum Keystones. Given the antique nature of the puffins, keystones could be a stand-in for the whatever means the original Remnant used to get into the waste. 
2. Have the puffins available for purchase not equipped with anything, and have Plume give the player a single Key Stone, stating that they'll need it for the Puffin. (missions that use the puffin would use a variant that does have a Key Stone)
3. Instead of having the player buy the puffin, have them given a Puffin instead. Puffins could then be treated as normal as a license restricted ship. Thus the player gets the puffin they need to do the mission, but doesn't gain any access to purchasing more Remnant stuff until they get the license. And them having access to a single Key Stone provided with the puffin is no more a problem than them getting access to any of the other tech on the ship.

Of these three options, I like option 3 the best.

Zitchas:
OK, I have now implemented option 3.

For what it is worth, the way this changes the feel of the mission, I think this change is moving out of the "wrangle a bug" territory into "actually make things fit better" territory.

The Remnant are results driven, and not a for-profit institution. In retrospect, it doesn't make sense that they'd put research on hold so that the player can go drum up some cash. 

Also, to me, the reference to "the university" seemed to imply that there was a separate institution within the Remnant that was a research/education institute. This is not the case. *all* Remnant are expected to be learning, researching, and developing things. They don't have a stand alone institution dedicated to it. I'd much rather Plume emphasize the collective ownership.

movingpictures:
Maybe I'm making this overly simple - but could you not have an "integrated keystone" in the Puffin, the way the Starling has an integrated Cloak?

Zitchas:
You can, and it is, in fact. However, the game isn't testing for an attribute to be present on  your ship, but rather whether or not an outfit is present.

Amazinite:
I've simply readded the key stone outfit in https://github.com/endless-sky/endless-sky/commit/ab998384084383b5654fe33da597d0866e530962. We can discuss the merits of gifting the Puffin to the player later on.

Zitchas:
Fixed the conflict. Probably one of the most inconsequential conflicts I've ever had the pleasure to deal with. A conflict between a new whiteline and an old whiteline...

Pointedstick (approved):
Seems mrege-ready.

PR #5768, opened 2021-03-04
Paragraph breaks added to 100 planet and spaceport descriptions...

skilaa:
## Summary
Added 1-2 paragraph breaks each to 100 planet and spaceport descriptions for the sake of readability. All changes have been confirmed not to cause text box overflow in-game.

Additionally, these 27 descriptions received typo fixes or extremely minor rewording for the sake of clarity or flow:
Canyon spaceport
Chiron planet
Delve spaceport
Flood planet
Flood spaceport
Foundry spaceport
Gemstone planet
Geyser spaceport
Hephaestus spaceport
Humanika spaceport
Icefall planet
Lichen spaceport
New Austria planet
New Austria spaceport
New Boston spaceport
New Greenland spaceport
New Wales spaceport
Nifel planet
Nifel spaceport
Nimbus spaceport
Placer planet
Relic spaceport
Splashdown planet
Thrall planet
Trinket planet
Tundra spaceport
Vail spaceport

### DIFF
--- a/data/map.txt
+++ b/data/map.txt
@@ -26141,7 +26141,8 @@ planet Ada
 	landscape land/canyon4
 	description `Ada is the home planet of Lovelace Labs, one of the most highly respected designers of starship systems and weaponry. Ada was originally settled as a mining planet, and the mines here still produce a steady stream of metal to supply the Labs and the many other industrial companies here.`
 	description `	From above, the planet looks almost lifeless and barren. Most of the major cities on Ada are built into the walls of the canyons, where they are protected from the high winds and occasional dust storms that sweep across the surface.`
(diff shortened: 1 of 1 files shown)


W1zrad (commented):
Interesting idea. I don't have time to look at everything right now, but I've done a quick glance. It looks like what you're trying to do is space different qualities or subjects into different paragraphs, but what makes it sort of strange is that each paragraph is only a couple sentences long now, which makes the amount of separation feel excessive (at least to me). Still, I'm not entirely opposed to it; that spacing isn't inherently a bad thing. I also think the spacing looks more excessive here on Github than it would in-game just because of the text size and font, so maybe it's not as big a concern as I think.

skilaa:
> I also think the spacing looks more excessive here on Github than it would in-game just because of the text size and font, so maybe it's not as big a concern as I think.

Absolutely. Judging edits to the in-game display (justified text blocks wrapped circa every 30 characters) based on how they look displayed in a text file on github is missing the point. Check out how they look in the game before voting yea or nay.

(This isn't an all-or-nothing edit. Any paragraph breaks that are too choppy can easily be omitted without losing them all.)

PR #5774, opened 2021-03-06
Vareti & Telis Engines: Minor Buff

DeBlister:
## Summary
The current Vareti Engine Block is less efficient than the Telis Engine Nacelles in terms of:
- Engine Efficacy
- Engine Efficiency*
- Outfit Efficacy (Turn only)
- Outfit Efficiency* (Turn only)
- Thrust/Turn per Energy/Heat

*using approximate energy and cooling costs

This PR slightly buffs the Vareti Engine Block to bring its performance in-line with the general trend of larger engines in a given series as well as its in-game description (larger, but more efficient).  The thrusting/turning values are untouched, but the following were altered:
-  Engine size (4.2% reduction; 1.25x ratio vs 1.20x ratio previously)
- Overall energy requirement (16.4% reduction; current version is ~14% more expensive than Telis)
- Turning heat (14.9% reduction; current version is ~22% more expensive than Telis)

The result is an overall better engine than the smaller Telis; it still leaves better, larger options for engine efficacy and outfit efficacy/efficiency.


beccabunny (approved):
Might be worth reducing the outfit space it takes a little accordingly

DeBlister:
Thanks for the feedback. :) I dropped it down to 168; it does begin to encroach on Bufaer and Stellar thrusters in terms of outfit efficiency, but it is still rather behind in the turn department so it should be OK there.  I don't know if I would reduce it further, though.

DeBlister:
I actually was intending to reduce the outfit space to 168 but then restored it to 171 mostly due to the ratio on Telis (~1.35x).  Looking at the other Ka'het engines, however, it's Telis that's out of line, not Vareti (they average ~1.26x).  Telis could use an outfit space buff to be honest, maybe to 102 or 103.  Not sure if I should combine that here or start another PR.

Zitchas (approved):
Looks good to me. And you could do the Telis ones in here too. Might as well do the two engines together. If they were part of a different faction set I'd say split them up in different PRs, but as they are all Graveyard stuff...

DeBlister:
I settled on 101 space.  I don't have anything else to change as of now; efficiencies and efficacies now seem to be in line with each other.

DeBlister:
Just increased the turn on Vareti by 2%.  It had less efficacy than Telis before, and now has virtually the same efficacy.  It still isn't the best at turning (efficacy or efficiency) so I don't think that's an issue but Telis could be reduced and Vareti could be restored down if the consensus is otherwise.

PR #5775, opened 2021-03-07
Fix: Independents hostile before FW

### LINKED ISSUE #5758: Independents hostile before Men becomes Independent
**Describe the bug**

The "Independent" faction officially appears when the former pirate System Men becomes independent during the FW campaign. Unfortunately, you can encounter "Independent" ships earlier, if you do the "City-Ship-License / Project Hawking" storyline before choosing sides in the FW campaign. Eliminating those ships in the Deep will make you hostile to the not-yet-founded independent faction in the south.

**To Reproduce**

During the "project Hawking" story, you will be followed by ships flying pirate coulours, but using "independent" FFI. Landing on Midgard, one of them will bombard you, regardless wether you killed them before or not. IF you killed them, however, your reputation with the "independent" will stay negative when later the "independent" will officially be founded.

**Expected behavior**

"Project Hawking" obviously has only to do with the Deep and Belzebub, but is nothing about the Independence of System Men. The Independent Faction of System Men should not care about what you did to those annoying ships in the Deep.

**Additional context**
while using the "independent" faction to have those ships fly "pirate colours" without being identified as actual pirates, might be a nice idea, your reputations with the "Independent" faction should be reset once that faction is officially founded.

clifflocked:
**Bugfix:** This PR addresses issue #5758
fixes #5758 

## Fix Details
This PR makes the Independent ships that follow you be Independent (Killable), so that they do not interfere with the Southern Independent planet & ships.

## Testing Done
I replaced all modified files, since no source code was changed.

### DIFF
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -216,7 +216,7 @@ government "Independent (Killable)"
 	"player reputation" 10
 	bribe .05
 	fine 0
-	# Nothing you do should anger the Killable Independents, as the
+	# Nothing you do should anger the Killable Independents, as they
 	# should not reveal themselves to you.
 	"penalty for"
 		disable 0
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -214,6 +214,8 @@ government "Independent (Killable)"
 	color .78 .36 .36
 	
 	"player reputation" 10
+	bribe .05
+	fine 0
 	# Nothing you do should anger the Killable Independents, as the
 	# should not reveal themselves to you.
 	"penalty for"
--- a/data/governments.txt
+++ b/data/governments.txt
@@ -214,8 +214,13 @@ government "Independent (Killable)"
 	color .78 .36 .36
 	
 	"player reputation" 10
-	"bribe" .05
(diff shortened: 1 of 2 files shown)


data/governments.txt
```diff
@@ -214,8 +214,15 @@ government "Independent (Killable)"
 	color .78 .36 .36
 	
 	"player reputation" 10
-	"bribe" .05
-	"fine" 0
+	bribe .05
+	fine 0
+	# Nothing you do should anger the Killable Independents, as the
```
Anarchist2, line 219:
```suggestion
	# Nothing you do should anger the Killable Independents, as they
```

Amazinite:
Note that the independent killable government is only used in two places right now, one where the goal is to kill the NPC and the other where the goal is to evade the NPC, and killing an NPC counts as evading it. The NPCs from this mission on the other hand don't fit into either objective. They aren't necessary to kill and optionally killing them doesn't gain you anything. So I don't see why we would add then to the killable government.

PR #5776, opened 2021-03-09
Fix dragAcceleration formula

MasterOfGrey:
**Bugfix:** This PR addresses the issue I posted in the Bug-Reporting discord channel

## Fix Details
DragAcceleration formula updated to correctly treat drag as an acceleration in the same units.

The error size is small (~1.65%) but compounds to create significant problems when being used for predictive things such as stopping distances with reverse thrust.

See the excel sheet posted in bug-reporting for the full math explanation.

## Testing Done
Nil - fix arose from testing and finding weird behaviour in related maths

## Save File
Nil

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1775,7 +1775,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 	if(acceleration)
 	{
 		acceleration *= slowMultiplier;
-		Point dragAcceleration = acceleration - velocity * (attributes.Get("drag") / mass);
+		Point dragAcceleration = acceleration - velocity * (attributes.Get("drag") / (mass/60));
 		// Make sure dragAcceleration has nonzero length, to avoid divide by zero.
 		if(dragAcceleration)
 		{
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1775,7 +1775,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 	if(acceleration)
 	{
 		acceleration *= slowMultiplier;
-		Point dragAcceleration = acceleration - velocity * (attributes.Get("drag") / (mass/60));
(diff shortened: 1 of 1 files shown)


source/Ship.cpp
```diff
@@ -1775,7 +1775,7 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 	if(acceleration)
 	{
 		acceleration *= slowMultiplier;
-		Point dragAcceleration = acceleration - velocity * (attributes.Get("drag") / mass);
+		Point dragAcceleration = acceleration - velocity * (attributes.Get("drag") / (mass/60));
```
Amazinite, line 1778: 😔 No one ever reads the [style guide](https://endless-sky.github.io/styleguide/styleguide.xml).
```suggestion
		Point dragAcceleration = acceleration - velocity * (attributes.Get("drag") / mass / 60.);
```

Amazinite: So acceleration is m/s^2. Velocity is m/s. Drag is kg*m/s^2, as it is a force. Mass is kg. The added 60 is s. That means in terms of units we get:
* [m/s^2] - [s] * [m/s] * [kg*m/s^2] / [kg]
* = [m/s^2] - [s] * [m/s] * [m/s^2]
* = [m/s^2] - [s] * [m^2/s^3]
* = [m/s^2] - [m^2/s^2]
Cool.

MasterOfGrey: Sorry, first time code edit. My B.
I went with the division because then it lines up conceptually better with converting displayed acceleration into per-frame acceleration and it's specifically that element that's out of step - even though it doesn't make any mathematical difference.

vitalchip:
Dividing the mass by 60 here will increase the drag by a factor of 60 and top speeds will crash through the floor, no?

MasterOfGrey:
> Dividing the mass by 60 here will increase the drag by a factor of 60 and top speeds will crash through the floor, no?

You're welcome to investigate the spreadsheet I uploaded in bug-reporting. In actual fact our top speeds have been higher than they're reported by the game, by 1.65%. The impact is not large because it's still a proportion of velocity thing.

thrashie (approved):

PR #5772, opened 2021-03-05
[Combat Revamp] Vast number of Combat & Balance changes intended to test our Options

Zitchas:
**Content (Artwork / Missions / Jobs)**

## Summary
This PR is *not* intended for merging in the near future, although if people like the changes it can be considered for such. More likely, this could be the base for the separate "Combat Changes" branch that Amazinite has mentioned wanting to do in the past.

## Thoughts

This PR includes a lot of changes. Broadly speaking, it is focused on mobility related aspects of combat. I'm looking at making the feel of the various ship categories more diverse, with fighters, drones, and interceptors actually having a unique level of mobility unmatched by any of the warship tiers. One prime example of this is dodging. It should be quite feasible now for any fast ship, and interceptors in particular, to dodge quite a few projectiles. The AI isn't going to benefit much from that, but I hope the player will. Dodging always felt very rewarding for me in the EV/O/N series, and has always been a sore point for me in ES because a) successfully dodging stuff didn't feel very meaningful, and b) dodging stuff is much, much harder. 

I would just like to encourage everyone to *NOT* read the details listed below. I know everyone has their own things they want to see changed, and there are some strong feelings about what is and is not appropriate to change. I'd like to argue that how it feels in-game should take priority over *all* our out-of-game meta analysis. There will be a time and place for it, but at the very start of a big overall is a time for first-impressions and feel rather than nitty gritty details. Please, please just switch to this in the launcher, fire up a brand new pilot, and just play the game for a while. See how things feel. After you've got a couple hours of flying around and getting into battles, *then* come back and look over the details.

Again, please skip reading the detail list, and just go try it out. The entire purpose of this is to give us a way to experience the changes we've been talking about, and I fully expect it to change in the future as tweaks, changes, and other ideas are added to it.

I strongly recommend trying to ignore your preconceptions. Because so many things have changed, when purchasing a new ship or deciding on engines; actually take a look at the performance of the ship options available to you first. 

**Please Note:**
Petervdmeer's Formations have been merged into this as of March 6th. You can change the formation of your own fleet by using **shift-G** to rotate through the available formations. The list of messages will tell you which formation you just switched to. Please note that formations are selection sensitive. As in, just like you can give different commands to different sets of ships in your fleet; you can use different formations for different sets. 

Temporary controls to set formations. For default controls the following:

    shift-G to change the formation on selected ships (or the fleet)
    shift-F to move selected ships (or the fleet) to a higher ring
    shift-H to move selected ships (or the fleet) to a lower ring

[...]

Issue #5271, opened 2020-09-01
There is a lack of progression in speed related to ship sizes


Amazinite:
**Is your feature request related to a problem? Please describe.**
Currently in game there is a progression as you increase the size of your ship. Generally, the bigger the ship within a tier, the more HP it has, the more DPS it can output, and the more outfit/weapon/engine space it has. One would also think though that the bigger your ship, the slower it goes. While it is roughly the case that larger ships are slower than smaller ships, it is very easy to equip a larger ship with engines that make it as fast as or faster than a much smaller ship. Therefore, there is currently no real progression in speed related to ship size. Increasing the size of your ship yields no downsides, as the one expected downside of lower speed does not exist.

**Related Issue Links**
Other issues that either speak directly to or are related to the balance between large and small ships:
#1932 was closed because turrets were given turn rates. While this helped to give smaller ships a chance, the issue of speed still exists.
#4034 The issue with smaller ships vs larger ships is related to the issue of fighters being underwhelming at the moment.
#4407 suggested using multipliers to make generators that are more attractive to smaller ships, but I suggested that we instead simply make smaller ships regenerate faster. (This is also something mentioned in #1932.)
#4697 discusses ways to make smaller ships have more of a chance against larger ships. 

**Describe the solution you'd like**
Rebalance the mass and drag of ships to have a much more drastic jump between categories of ships.

For example, right now the difference in the average mass of a human interceptor versus a human light warship is only +80%. From light to medium warship it's +87%, and from medium to heavy warship it's +126%.
In terms of drag, the progression is +127% from interceptor to light warship, +86% from light warship to medium warship, and +57% from medium warship to heavy warship.

By increasing the jump in mass and drag as you go up a ship category, we'd make it so that larger ships handle noticeably slower than smaller ships, being the trade off for more HP and DPS.

**Describe alternatives you've considered**
Another point of concern might be the progression in power between various types of engines. A ship with atomic engines will be going MUCH faster than a ship with ion engines of the same size. Perhaps we want to reduce the difference between the slowest and fastest engines. Currently, a tiny atomic thruster has nearly double the thrust per ton as a tiny ion thruster. (87% more to be exact.) This difference in power continues all the way up the size progression of engines. (The huge atomic thruster has 81% more thrust per ton than the huge ion thruster.)

[...]

Issue #4697, opened 2019-12-03
Generally increase weapon inaccuracies to improve the chances of smaller ships to dodge


Amazinite:
**Is your feature request related to a problem? Please describe.**
Weapon inaccuracies are currently generally very low, which means that even at max range the cone of fire is small enough that the inaccuracy doesn't even come into play, even for small ships. That is, even at the maximum range, nearly every shot will still connect due to the size of ships versus the cone of fire.

**Related Issue Links**
#4034

**Describe the solution you'd like**
Increase weapon inaccuracies across the board, except for on homing weapons and lasers. This will make it so that the cone of fire comes more into play at longer ranges, meaning that it may be possible for smaller ships to actually dodge shots because of the inaccuracy of the enemy.

jafdy:
I don't think this really does "doge", more "miss".

Zitchas:
I like this, although on the flip side if the shots are more randomly spread out across a cone, they become harder to dodge. 

Given that the AI doesn't really dodge, this seems like something that will make life easier for the AI, new players, and those who don't dodge; and make things harder for those who do dodge.

Pointedstick:
I generally agree.

ravenshining:
I'm sceptical about this idea.  Inaccuracy in a projectile weapon may well make it *harder* to dodge, since you have to avoid a larger area of the battlefield.  And, if projectile weapons are allowed to spray wildly, it makes much less sense for certain projectiles such as the Gatling to be relegated to short range - my headcanon about why a ballistic slug vanishes after a very short range is that the gun is firing very inaccurately, so the low range is acting as a proxy for lack of focus.

I rather think lowering turret turn rates would be a more meaningful boon to smaller ships, while being able to be applied to both projectiles and beams.

Zitchas:
That reminds me, that there's at least a few weapons that should probably have significantly increased lifetime, but with a "range override" at their current max range, to represent that they continue to be dangerous for much farther out, even if it is impossible to consistently hit anything at those ranges.

On the flip side of this idea, I think I might be leaning more towards keeping accuracy as-is for the most part, with only specific weapons that are noted for having poor accuracy being bumped up. Then have the general inaccuracy of governments increased. 

So a blaster on a navy ship might have not-great-but-not-terrible accuracy, while a blaster on a merchant ship might be basically an inacurate cone of fire.


[...]

PR #5779, opened 2021-03-11
Combat Test

Amazinite:
**Content (Balance)**

# Summary
This PR will constitute a number of changes that seek to address the various issues under the combat test milestone. I will update this PR with ideas as I work on them, adding a new explanation for each change to this summary as I go. Feedback on each idea is greatly appreciated.
Note that this PR will specifically focus on data changes. I will not be implementing any new mechanics in this PR.

The reason for having one big PR to work on these changes are that some of the balance issues that have been brought up over the years are far more fundamental than just "X weapon/ship feels too strong," and are instead issues like "this entire combat mechanic is misbalanced." As such, resolving such issues will require a large number of changes to be done at once. With that in mind, much of what I'm seeking to do here is build multiple changes on top of each other in order to resolve some of the issues in the combat test milestone, meaning that what is seen so far might not be exactly what we're looking for as a result of needing something else layered on top of it. For example, at first I have only made changes to the balance of engines in human space. This alone won't completely resolve the issue of ship speeds with relation to ship size, but later changes, like tweaking the mass and drag of various ships, will build on top of this one to get us closer to our goal.

This is intended to be a more precise version of #5772--which is more scattershot right now--making smaller changes to get feedback on before building on top of them instead of tossing a bunch of stuff together at once. (Not to say that this way is better, though, just taking a different approach so that we can try to offset any issues that the opposite approach might have.)

This PR will start off by only making changes to human space, then grow from there as general changes are agreed upon. Starting with just human space means that we have a smaller number of things that we need to look at at any one time, making the testing of these changes simpler. We can then refine the changes in human space and use that as a jumping off point to change things in the rest of the game. Making sure that balance changes are near-perfect in human space first also means we're focusing on the part of the game where everyone starts off, ensuring a more pleasant first time experience.

# Explanation of changes

### Addressing #5271: there is a lack of speed progression with ship size

<details><summary>Summary of this issue:</summary>

Much of what these changes seek to fix are described in the above issue, but the TL;DR of it is that larger ships are much too fast, giving smaller ships very little room to be useful. Large ships simple don't have high enough drag and mass, and the largest engines are too powerful, resulting in basically every ship being able to fly around like an interreceptor. Therefore, these changes seek to drastically slow down larger ships while slightly increasing the speed of smaller ships, and at the same time bringing the weakest and strongest engines closer together in power so that the slowest reasonable fitting of a ship (slowest reasonable being using weak engines that fill the space instead of say just slapping on an X1050 ion engine and leaving loads of empty engine space) isn't immensely slower than the fastest reasonable fitting of a ship (fastest reasonable being a ship that has a fair mix of turn rate and max speed instead of dumping all your engine space into one or the other). The goal of this is to have large ships actually feel large, while at the same time giving small ships much more room in the game's combat by giving them the ability to run circles around larger ships or simply outrun them without fear of a heavy warship with the best engines possible catching up to an interceptor with mediocre engines.

For the sake of brevity (although I still go rather in-depth), the following sections only describe the changes made to human ships and outfits, but the same or a similar process has been applied to all other ships and outfits in the game to achieve the desired goal.

</details>

<details><summary>Rebalance engines:</summary>

One of the issues with the speeds of ships right now is that each ship has a vast spread of potential speeds that it can have. This is because while an ion engine might provide a certain amount of thrust, an atomic engine of the same size is currently providing nearly double that amount. The end result of this is that the spread of speeds that each ship can get is so wide that the fastest of the large ships can be considerably faster than the slowest of the small ships. Given that most NPCs aren't running around with the strongest possible engines, this is part of where the issue of large ships being faster than small ships. All things being equal, the smaller ships would generally be faster than the larger ships, but all things are not equal where the fitting of ships is concerned. 

In order to combat this, I suggest bringing the slowest and fastest engines much closer together. This PR seeks to nerf atomic engines by bringing them down to roughly 80% their current thrust, bringing them more in line with ion and plasma engines.

The following image is a scatterplot showing the current human thrusters with mass on the x axis and thrust/ton on the y axis. The blue dots are ion engines, orange plasma, and green atomics. As you can see, atomic engines are much stronger than both ion and plasma engines.
image
The result of bringing atomic thrusters down to 80% their current thrust is the following scatterplot. Now, instead of looking that atomic thrusters are two or three steps above plasma thrusters, the gap between plasma and atomic thrusters is now more comparable to the gap between ion and plasma thrusters. In this way, atomic thruster are no longer far above and beyond anything else you can buy in human space, instead simply being another reasonable upgrade to the previous option. Given that the speed gained from upgrading to atomics is no longer so absurd, this might also make the use of ion and plasma engines more acceptable. Another result of this change is making afterburners more attractive to use. Taking the normal human afterburner as an example, it currently has about 50% more thrust/ton when compared to the largest human atomic thrusters. With this change the afterburner now has nearly double the thrust/ton, making it a more attractive option without even changing it as a result of giving it more breathing room. 
image

Along with nerfing atomic engines to 80% their current value, this PR also buffs all steering by roughly 25%. The reasoning is largely the same as the nerf to atomic engines: bring all engines closer together, only this time going in the opposite direction of buffing the weakest instead of nerfing the strongest. But what we also gain from buffing all steering is more wiggle room with how we tweak the mass of ships. Resolving #5271 doesn't only need to be a matter of nerfing large ships, but can also be a matter of buffing small ships. Turn rate is determined by turn divided by mass, so in order to make ships turn faster we can either increase the turn rate or reduce the mass. The issue with trying to make small ships turn faster though is that small ships already have such low masses that there is hardly any room to make them lower, while there is plenty of room to increase the mass of large ships. So in order to make small ships faster, we need to instead increase the amount of turn, hence the buffing of steering instead of nerfing it. The relationship between ship mass and this change will be discussed further in the explanation for the rebalancing of ship masses.

The following scatterplot shows the current human steering.
image
And this scatterplot is with the changes from this PR. Note that the 80% nerf to atomics and the 25% buff to all steering counteract each other, resulting in no change to atomic steering.
image

</details>

<details><summary>Rebalance ship masses:</summary>

As discussed in the explanation for the previous commit, the turn rate of a ship is determined by its total turn divided by its mass. By simply increasing the turn value of all steering engines, we have increased the turn value of all ships. This gives us room to increase the mass of certain ships in order to bring their turn rates back down.

What I've ultimately done with these mass changes is reduce the mass of smaller combat vessels like drones, fighters, and interceptors, generally kept light warships at the same or around the same mass, generally increased the mass of medium warships such that the turn rates of medium warships are about what they were before the steering changes, and generally increased the mass of heavy warships such that they are substantially heavier and therefore slower turning than they were before.

The masses of all freighters and transport ships have remained unchanged. This means that these ships will all receive a flat 25% buff to their turn rates when using ion and plasma steering. My reason for not making changes to these ships is because freighters are typically already much slower than similarly sized warships, and transport ships aren't all that useful in combat enough that their speed is an important factor. If anything, leaving the masses of these ships unchanged and therefore effectively buffing them compared to warships that have had their masses increased means that freighters and transport ships are now more easily able to escape combat, as more formidable warships are now relatively slower, and freighters and transports can use their buffed turn rate to more easily escape the system.

Light warships are now 90% more massive than interceptors on average, medium warships 91% more massive than light warships on average, and heavy warships about 200% more massive than medium warships on average. Previously this relationship was 66% heavier from interceptor to light, 68% heavier from light to medium, and 86% heavier from medium to heavy.

I will note though that heavy freighters are still exceptionally slow, slower even than heavy warships, largely as a result of their small engine spaces. This may need addressed in the future if we see fit, but as of right now I don't see a need to look into the speed of heavy freighters.

Exceptions to the above changes:
* The Auxiliary, a transport ship, was balanced as a heavy warship.
* The Raven received a significantly greater mass increase than other light warships, as I feel that it was far too light for its capabilities.
* The Type-F Haulers (Nest, Roost, Skein) have been unchanged, as they are balanced relative to the Hauler I, II, and III, and increasing their masses would make them much, much too slow, as they are already very sluggish ships, especially when carrying fighters.
* The Protector and Vanguard had their mass balanced more like medium warships. That is, their increase in mass was such that they turn at about the same rate as they did before. This is because these are the lightest two heavy warships, and at least in regards to the Protector, generally not that well regarded.
* The Dreadnought is designed as a comparable ship to the Cruiser, yet it lacks any drone bays. As such, I have increased the Dreadnought's mass to a greater degree than the Cruiser's to account for the difference in bays. This should not be too much of an issue, as the Dreadnought also comes with more engine space than the Cruiser.
* The Rainmaker, a light warship, has had its mass reduced rather than increased because even with the engine changes this ship felt far too sluggish. 
* The Splinter's mass has been left untouched for reasons similar to the Protector being balanced differently, namely that I get the sense that it's a lesser used ship and could use some love. At the very least the description does mention that it's suppose to be a speedy ship, so leaving it untouched means that it has been buffed relative to other medium warships.

A side effect of increasing mass to target turn rates is that any ship that has had its mass increased now accelerates slower than before given the same thrusters, but that is not necessarily the focus of these changes.

The following graphs show the mass of all human ships in order before and after these changes. The mass value plotted is specifically the mass of the ship's hull plus the mass of the outfits if every bit of outfit space was taken up. This does not account for cargo space or the mass of any carried ships, although I did take carried ships into mind for those ships where their bays are an important part of how the ship is used.

Before: (The final point in the first graph is a dummy point for scaling purposes. In the second graph, that point is the Bactrian.)
image

After:
image

</details>

<details><summary>Rebalance ship drags:</summary>

The other component of ship speeds are their max speeds, determined by thrust divided by drag. In reducing the strength of atomic thrusters, all ships have been made slower. But much like the changes to turn values on engines were followed up by changes to mass values of ships, I have followed up the changes to thrust values on engines with changes to drag values of ships.

My goal with the drag changes is to create more clear lines between categories of ships; fighter-sized, interceptor-sized, light warship-sized, medium warship-sized, and heavy warship-sized. This means that transport and freighter ships have been placed into these size categories as I saw appropriate and adjusted as such, although freighters specifically are still generally slower than similarly sized warships.
* Ships treated as interceptors: Flivver, Scout, Shuttle, Heavy Shuttle, Arrow, Bounder, Star Barge
* Ships treated as light warships: Freighter, Argosy, Clipper, Blackbird, Star Queen
* Ships treated as medium warships: Hauler I-III
* Ships treated as heavy warships: Behemoth, Bulk Freighter, Auxiliary

Fighter-sized vessels have had their drags roughly cut in half, meaning that their max speeds are now roughly doubled. Exceptions to this include the Boxwing and Dropship, which are non-combat vessels, and the Finch, which is the only human fighter capable of equipping atomic engines. Fighter-sized vessels have very little room for customization, and therefore the speed you get is the speed you keep when buying them. Given that, in order to have fighter-sized vessels be faster than almost every interceptor-sized vessel in almost every circumstance, these ships had to have their drag significantly reduced. 

Interceptor-sized vessels have had their drags reduced by roughly 10-20%. Exceptions to this include the Star Barge, which had its drag reduced by closer to 30%, the Flivver, which had its drag reduced by 40%, and the Wasp, Shuttle, and Bounder, which I felt were already in good places. This puts the max speed of interceptor-sized vessels generally lower than fighter-sized vessels, although atomic engines can often bring these ships to speeds close to or exceeding that of fighters.

As I got further up in the sizes of ships, changes became less blanket for the entire category and more directed toward specific ships. Among the light warship-sized vessels, the Headhunter and Clipper, already rather speedy ships, remained the same, while the Raven, Gunboat, Blackbird, and Star Queen lost about 10% of their drag, with the Quicksilver, Argosy, Modified Argosy, and Freighter losing 25-30%, and the Rainmaker losing 50% of its drag. These changes bring light warship-sized vessels closer together in terms of their max speeds, while still being slower than interceptor-sized vessels by a decent amount.

Medium warship-sized vessels I felt were already in a decent enough spot. As such, the main changes were to swap the Splinter and Osprey in terms of their max speeds given a balanced loadout (about +7% drag for the Osprey and -9% drag for the Splinter), and up the speed of the Bastion by a fair amount (nearly 20% less drag) so that it is closer to medium warship levels than heavy warship levels. I've also opted to buff the HP on the Osprey slightly as a tradeoff for the slower speed, making it more of a "Free Worlds Frigate." The main standout among the medium warship-sized changes are those of the Hauler I-III and their Type F versions. Being freighter and freighter-derived ships, these six ships were already very sluggish. I felt as though they were far too sluggish though. As such, I've reduced their drags by about 25-40%. They're still very slow ships, but this should give them a bit more of a reason to be used.

And finally, heavy warship-sized vessels were also already in a decent spot to me. The main changes I made were not to slow down heavy warships as a category, but to bring the speeds of heavy warships closer together. As such, the Auxiliary has gained about 7% drag, the Falcon, Cruiser, and Dreadnought about 10%, and the Bactrian lost about 5% drag, the Vanguard about 10%, and the Behemoth, Bulk Freighter, and Protector lost about 30%.

While there still might be some overlap between categories--especially if you sacrifice turn rate for max speed on larger ships and in cases where a ship naturally straddles the line between categories, such as the Modified Argosy or Bastion being heavy for their size or the Corvette being fast for its size--given a ship of a certain size with balanced turn and max speed, it should generally be faster than almost all ships larger than it and slower than almost all ships smaller than it. 

In terms of comparing across categories, heavy warships are now about three to four times as slow as interceptors, whereas before this value was closer to half. Hopefully this is the type of difference in speed given size that people are expecting.

</details>

<details><summary>Results:</summary>

The following table shows the max speed and turn rate of all human ships as their default when you buy them, as well as a the speed and turn of a customized version using atomic engines where they fit. If the ship couldn't fit atomic engines (only fighters and drones had this issue) then I just fit whatever I could. Some ships may have the same engines for the default and custom, either because nothing else can fit or because the ship already had atomics in it. These customized columns are mainly what I looked at to determine the mass and drag changes that I made, using them as representations of what a customized ship balancing its max speed and turn rate would look like.

You may notice that in some cases the custom speed or turn rate is slower than the default. This may be because of one of two reasons.
1) The turn rate displayed is given a ship that has used up 100% of its outfit space. This is to give a minimum possible turn rate under generally normal circumstances (i.e. in combat not weighed down by fighters or cargo). Some default loadouts have a lot of empty space though, which is why a fully kitted out ship with atomics might be slower than a default ship with plasma or ion engines.
2) For certain ships, only making use of atomic engines actually leaves a fair amount of empty engine space, where slightly larger plasma or ion engines might fill the space better. As a result of filling more space, the plasma or ion engines could potentially be a better choice in terms of speed. The fact that this scenario can occur sounds to me like the changes to atomic engines were a good one, making them not always the no-brainer option (even if in most cases they still are the way to go for high speed loadouts).

Name | Category | Default Speed | Default Turn | Custom Speed | Custom Turn
---|---|---|---|---|---
Arrow | Transport | 780 (A250) | 134.7 (A255) | 780 (A250) | 134.7 (A255)
Auxiliary | Transport | 203.5 (X4700+X1700) | 49.8 (X5200) | 320.1 (A520+A120) | 63.6 (A865)
Blackbird | Transport | 472 (Impala) | 122.2 (Impala) | 502.7 (A370) | 120.4 (A375)
Bounder | Transport | 358.4 (X3700) | 129.5 (X3200) | 716.8 (2*A250) | 199.8 (A375)
Flivver | Transport | 1270 (A120) | 71 (X1200) | 1270 (A120) | 138.4 (A125)
Heavy Shuttle | Transport | 373 (X2700) | 103 (X2200) | 716.8 (A250) | 179.2 (A255)
Scout | Transport | 442 (X3700) | 144.3 (X3200) | 754 (A370) | 216.7 (A375)
Shuttle | Transport | 405.9 (X2700) | 125.6 (X2200) | 780 (A250) | 216.9 (A255)
Star Queen | Transport | 331.5 (X3700) | 73.4 (X3200) | 565.5 (A370) | 115 (A375)
AVERAGE | Transport | 515.1 | 105.9 | 711.8 | 153.9
Argosy | Light Freighter | 263.5 (Greyhound) | 68.7 (Greyhound) | 539.9 (A370) | 68.7 (A255)
Clipper | Light Freighter | 298.4 (Greyhound) | 103.1 (Greyhound) | 611.4 (A370) | 100.5 (A255)
Freighter | Light Freighter | 172.5 (X2700) | 47.4 (X2200) | 331.5 (A250) | 84.1 (A255)
Hauler | Light Freighter | 169.8 (Greyhound) | 125.6 (Impala) | 204 (A250) | 119.2 (A375)
Star Barge | Light Freighter | 211.8 (X1700) | 63.5 (X1200) | 448.2 (A120) | 117.6 (A125)
AVERAGE | Light Freighter | 223.2 | 81.7 | 427.0 | 98.0
Behemoth | Heavy Freighter | 158.8 (X3700) | 45.1 (X3200) | 270.9 (A370) | 39.3 (A255)
Bulk Freighter | Heavy Freighter | 117.3 (X3700) | 36.7 (X3200) | 200.2 (A370) | 32.2 (A255)
Hauler II | Heavy Freighter | 138 (Greyhound) | 105.5 (Impala) | 165.8 (A250) | 100.7 (A375)
Hauler III | Heavy Freighter | 116.2 (Greyhound) | 91 (Impala) | 139.6 (A250) | 87.2 (A375)
AVERAGE | Heavy Freighter | 132.6 | 69.6 | 194.1 | 64.9
Combat Drone | Drone | 1000 (X1700) | 153.8 (X1200) | 1000 (X1700) | 164.4 (X1200)
Surveillance Drone | Drone | 1125 (X1700) | 176.5 (X1200) | 1125 (X1700) | 190.5 (X1200)
AVERAGE | Drone | 1,062.5 | 165.2 | 1,062.5 | 177.5
Barb | Fighter | 600 (X1050) | 82.8 (X1050) | 600 (X1050) | 82.8 (X1050)
Boxwing | Fighter | 406.8 (X1050) | 103.5 (X1050) | 610.2 (X1700) | 141.2 (X1200)
Dagger | Fighter | 900 (X1700) | 110.1 (X1200) | 900 (X1700) | 110.1 (X1200)
Dropship | Fighter | 450 (X1700) | 100 (X1200) | 952.5 (A120) | 188.2 (A125)
Finch | Fighter | 800 (Chipmunk) | 133.7 (Chipmunk) | 1058.3 (A120) | 162.2 (A125)
Lance | Fighter | 857.1 (X1700) | 96 (X1200) | 857.1 (X1700) | 96 (X1200)
AVERAGE | Fighter | 669.0 | 104.4 | 829.7 | 130.1
Berserker | Interceptor | 392 (X2700) | 170.1 (X3200) | 753.3 (A250) | 152.7 (A255)
Fury | Interceptor | 518.8 (X2200) | 103.9 (X2200) | 997 (A250) | 187.4 (A255)
Hawk | Interceptor | 661.1 (Greyhound) | 136 (Greyhound) | 794 (A250) | 141.2 (A255)
Sparrow | Interceptor | 711.1 (Chipmunk) | 113.3 (Chipmunk) | 940.7 (A120) | 134.4 (A125)
Wasp | Interceptor | 575 (X2700) | 111.5 (X2200) | 1105 (A250) | 201.1 (A255)
AVERAGE | Interceptor | 571.6 | 127.0 | 918.0 | 163.4
Gunboat | Light Warship | 473.6 (X3700) | 52 (X2200) | 807.9 (A370) | 93.7 (A255)
Headhunter | Light Warship | 424.6 (Greyhound) | 103.3 (Greyhound) | 870 (A370) | 109 (A255)
Modified Argosy | Light Warship | 250.9 (Greyhound) | 75.6 (X3200) | 301.4 (A250) | 115.4 (A375)
Quicksilver | Light Warship | 552 (Greyhound) | 112.1 (Greyhound) | 663 (A250) | 114.5 (A255)
Rainmaker | Light Warship | 363.2 (X2700) | 65.3 (X2200) | 697.9 (A250) | 65.3 (A125)
Raven | Light Warship | 401.8 (X3700) | 90.8 (X3200) | 685.5 (A370) | 143.0 (A375)
AVERAGE | Light Warship | 411.0 | 83.2 | 671.0 | 106.8
Aerie | Medium Warship | 323.4 (X3700) | 66.3 (X3200) | 551.7 (A370) | 103.4 (A375)
Bastion | Medium Warship | 206.2 (Impala) | 57.8 (Impala) | 272.5 (A370) | 96.9 (A525)
Corvette | Medium Warship | 378.9 (X3700) | 75.5 (X3200) | 646.3 (A370) | 113.5 (A375)
Firebird | Medium Warship | 294.7 (X3700) | 56.5 (X3200) | 502.7 (A370) | 84.1 (A375)
Frigate | Medium Warship | 255 (X3700) | 51.9 (X3200) | 435 (A370) | 81.3 (A375)
Manta | Medium Warship | 282.1 (X3700) | 63.7 (X3200) | 481.3 (A370) | 99.3 (A375)
Mule | Medium Warship | 232.6 (X3700) | 43.8 (X3200) | 396.8 (A370) | 90.7 (A375+A125)
Nest | Medium Warship | 169.8 (Greyhound) | 92.4 (Impala) | 348 (A370) | 50.9 (A255)
Osprey | Medium Warship | 337.1 (Impala) | 85.4 (Impala) | 646.8 (A520) | 84.1 (A375)
Roost | Medium Warship | 138 (Greyhound) | 74.1 (Impala) | 282.8 (A370) | 41.6 (A255)
Skein | Medium Warship | 116.2 (Greyhound) | 62.1 (Impala) | 238.1 (A370) | 35.2 (A255)
Splinter | Medium Warship | 363.3 (X3700) | 75.9 (X3200) | 619.7 (A370) | 110 (A375)
AVERAGE | Medium Warship | 258.1 | 67.1 | 451.8 | 82.6
Bactrian | Heavy Warship | 166.7 (X4700) | 39.5 (X5200) | 249.4 (A520) | 50.7 (A865)
Carrier | Heavy Warship | 276.3 (X5700) | 24.3 (X4200) | 290.5 (A520+A250) | 59.5 (A865)
Cruiser | Heavy Warship | 229.7 (X4700) | 33.5 (X4200) | 343.8 (A520) | 63.2 (A525+A255)
Dreadnought | Heavy Warship | 367 (Orca) | 42.2 (Orca) | 343.8 (A520) | 70.9 (A865+A125)
Falcon | Heavy Warship | 287 (Impala) | 43.5 (Impala) | 515.7 (A520) | 73.2 (A525)
Kestrel | Heavy Warship | 370.4 (Orca) | 57.4 (Orca) | 467.5 (A520+A250) | 84.9 (A865)
Leviathan | Heavy Warship | 174.5 (X3700) | 46.9 (X4200) | 297.6 (A370) | 65.1 (A525)
Protector | Heavy Warship | 186.8 (X3700) | 33.1 (X3200) | 318.6 (A370) | 51.1 (A375)
Vanguard | Heavy Warship | 184.2 (X3700) | 60.4 (X4200) | 314.2 (A370) | 84.2 (A525)
AVERAGE | Heavy Warship | 249.2 | 42.3 | 349.0 | 67.0

</details>

<details><summary>Created an "acceleration multiplier" attribute and used to triple the acceleration of all fighters and drones:</summary>

As is probably evident by the results summary above, my changes so far have been entirely aimed toward the balancing of max speeds and turn rates on ships, but there's still one more movement stat that we can balance: acceleration. Despite me not looking at it thus far, acceleration isn't a stat to be ignored, as the acceleration of a ship can go a long way to changing how sluggish the ship feels. 

While max speed is a factor of thrust / drag and turn rate is a factor of turn / mass, acceleration is a factor of thrust / mass. This means that we have two places where thrust comes into play and two where mass comes into play. The result of this is that balancing all three movement factors, max speed, turn rate, and acceleration, is rather difficult to do with just tweaking the attributes of engine thrusts and turns and ship drags and masses.

Over on #5772, people have suggested that a high acceleration is highly desirable for fighters and drones, accelerations at least three times what I had here, although they reached that acceleration through other means (in-built thrust, increase engine spaces) that I don't want to apply here. In order to reach such accelerations by only altering the four attributes that I mentioned above, I'd need to triple the thrust of all engines while tripling the drag of all ships in order to keep the max speeds the same while tripling accelerations for all ships, with the aim of doing so just for fighters and drones. This led to accelerations for other ships being far too high, though. In order to remedy this and get the accelerations back to where they were, I'd need to triple the mass of all ships on top of the changes I had already made. But since mass is tied to turn rate, I'd need to then triple the turn of all engines to keep the turn rates where I want them, but without having touched fighters or drones in order to keep their accelerations high, this would then mean triple turn rates on fighters. And all that work would *just* be to increase fighter and drone accelerations by a substantial amount. If we wanted to tweak the acceleration of other ships, say to have heavy warships all take at least 5 seconds to accelerate to full speed, it'd become increasingly difficult to maintain the same balance everywhere else while only touching a single category of ships.

Given that, excluding acceleration, all the attributes are generally where I want them, I decided to not go down this path of massively ramping up all the stats across the board just to the end of increasing fighter and drone accelerations. An alternative route would be to instead drop the mass of fighters and drones to a third of what they are now, but that isn't feasible without creating new outfits that are smaller than what fighters and drones currently use, which I don't particularly see as an effective route to take as it'd require us to make an entirely new outfit ecosystem for just fighters and drones. At that point I feel like we might as well go down the "fighters as outfits" route discussed in #4034. This route also has the issue of interceptors being described as fighters with hyperdrives; that is to say, fighters should only be slightly smaller than interceptors, not a third their size or smaller.

Given that I didn't want to take either of the above routes, I've created a new attribute "acceleration multiplier" that functions similarly to "hull repair multiplier" and other similar attributes. That is, this attribute will multiply a ship's acceleration by the equation `1 + "acceleration multiplier"`. I have given all fighters and drones an acceleration multiplier of 2, which results in 1 + 2 = 3 times the acceleration on these ships, without needing to touch the mass and drag of every other ship or the thrust and turn of every engine in the game.

In short, this new "acceleration multiplier" attribute allows us to tweak the max speed, acceleration, and turn rate of ships independently of one another, making balancing all ships across all three movement categories much simpler.

</details>

### Addressing making systems larger

<details><summary>Summary of this issue:</summary>

The size of systems isn't necessarily directly related to combat, but it certainly has an effect on the feel of combat. Systems are currently rather small in size, and battles tend to devolve into small clumps of ships just orbiting each other while they fire all their weapons. By opening up the size of systems more, we give ourselves more room to tweak the AI and expand the sizes of battles so that they aren't all about ships clumping together and blasting each other.

</details>

<details><summary>Doubled all distances:</summary>

Doubled the distance of all stellar objects, of any asteroid belts, of all habitable ranges, of the system fence, and of the effectiveness of ramscoops and solar panels.

This currently also includes ringworlds, so ringworlds will need to be addressed in some way, either by not touching their distances, or by increasing the size of the ringworld graphic.

</details>

### Miscellaneous changes

The following are a list of miscellaneous changes that I have made along the way of testing all the other changes made in this PR. The reasons for these changes will be described in each section, but generally their purpose is to clean up a rough edge that either already existed in the game or that came up as a result of other changes made, but the change didn't otherwise fit into any category above.

<details><summary>Doubled (almost) all shield generation and hull repair:</summary>

In making larger ships slower, they're now less able to escape from smaller ships that will now almost always be able to chase them down. As such, they'll now be taking more of a beating than before. Instead of increasing their HP, I'm opting to increase shield generation rates across the board by 100%.

I also generally feel like shield generation was too low to begin with. Without these changes, a default Sparrow takes 100 seconds to regenerate its shields from 0, and a default Cruiser takes nearly 400 seconds. So cutting down shield regeneration times should not only allow larger ships to stand up against the increased fire they'll be taking, but also let smaller ships more effectively do hit and run tactics, as needing nearly 2 minutes between attacks is far too much for an interceptor.

This same 100% buff has been applied to all hull repair.

Some exceptions have been made, including to Pug ships and Archons, which already have in-built repair rates that are FAR above what any other faction in the game has been capable of.

Shields Up is now real.

</details>

<details><summary>Nerfed the heavy rocket's damage by about 30%:</summary>

The power of the heavy rocket when used by pirates is a pain point for many given its ability to instantly disable many smaller ships when hit by two rockets at once, a loadout that is not uncommonly seen on ships like Hawks and Clippers. Some ships can even be outright destroyed by two heavy rockets, and light warships will still be heavily damaged from two hits.

Given this, I've lowered the damage of the heavy rocket by about 30%. Being hit by a heavy rocket will still be rather painful, but hopefully not as painful as before. Some small ships like the Shuttle may still be taken out by two heavy rockets, but a Sparrow will be capable of surviving two heavy rockets given that it has taken no other damage. This change has also brought the damage of a single heavy rocket below that of a single torpedo, which should hopefully further cement the torpedo as an anti-capital ship weapon while the heavy rocket is intended for firing on groups of targets.

In order to compensate for the lower damage, I increased the blast radius by 50%, but left the trigger radius the same so that a rocket is no harder or easier to dodge than before.

</details>

<details><summary>Returned to hit force only being based on a ship's mass:</summary>

A few years ago a change was made so that hit force (and firing force) on a ship was altered so that smaller ships feel less force than before while larger ships feel more force than before. (e61c25f) For the mass values that the game used, this hasn't been too bad of a change, but with the massively increase mass of ships in this PR, this mechanic made it so that large ships were being pushed far too much by hit force weapon or from the firing force of their own weapons, even with a change I made so that hit force was altered by 20% of a 500 ton fixed mass instead of 30% of a 400 ton fixed mass as was the case before.

Under this PR right now, the lightest ship in the game, the Surveillance Drone with a mass of 63, feels 18% less hit force because of this mechanic as compared to if hit force was only based on the mass of the ship. Then as you get closer to 500 mass, the % reduction in felt hit force gets smaller. On the opposite end, once you get to 1000 tons of mass (about the mass of a Mule or Behemoth under this PR), you're feeling +20% hit force, and it only increases from there. The Protector, the lightest heavy warship at 1400 tons, is feeling +36% hit force. The Cruiser at 2600 tons is feeling +84% hit force. The KIV at about 4000 tons is feeling +140% hit force. The Emerald Sword at 4350 is feeling +154% hit force. Then the Heron at 28600 tons is feeling +1124% hit force. So really, all this mechanic does right now is reduce the hit force on smaller ships by a negligible amount while MASSIVELY increasing the hit force on larger ships. 

We basically have three courses of action here: return to the old mechanic from several years ago where hit force is not altered by any fixed mass, change the fixed mass and weighting used (I found that changing from 20% of 500 to 20% of 2000 yielded better results), or change how hit force is altered to using a separate equation (@vitalchip suggested something to the effect of `force / (.3 * 400 + .7 * currentMass)` instead of `force * (.3 / 400 + .7 / currentMass)`, reducing the effective mass of a ship by a certain degree and using that to change the force instead of adding an additional fixed force). For now I've taken the simple route and simply removed the mechanic entirely, although with #5978 we could make it so that the mechanic remains possible to apply, just with the fixed mass having a weighting of 0%.

</details>

### TO BE ADDRESSED

<details><summary>Issues that I plan to address but haven't worked on yet. These sections will be moved up as they're worked on:</summary>

### Addressing #4039

This one can be done largely independently of other issues being looked at right now, so I'll tackle it here.

Part of this issue is also the sense of the density of ships in systems, so adding an arrival distance to all systems as is done in #5772 could be something we do here. Although I'd choose 1250 instead of 1500, as that is the point at which solar collection and ramscoops are 100% effective, being less effective further away and more effective closer in. If we want to have ships jump in from even further away, then we should also increase this distance.

I've also thought to perhaps double the distance of all planetary objects (i.e. not changing ringworlds) and double all habitable ranges, or something around that. That would give us room to reasonably increase the arrival distance further.

### Addressing #4697

Also decrease turn rates of turrets, again. Or at the very least decrease the turn rates on laser turrets.

### Addressing #3372 and #3264

I could drag #5793 and #5796 into this one.

### Addressing #4158 + human fleet strengths in general

This one will probably need to be done after all other outfit changes are made as PRs like #5793 and #5796 will change the dynamic between categories of ships.

https://steamcommunity.com/app/404410/discussions/2/3090011896393896520/

### Addressing #4077

I think this one will be solved as a result of all other changes listed here.

### Addressing #4034

Again, should hopefully be solved or at least made better by other changes here. Additional ideas might be to shrink the size of fighter sprites to make them harder to hit and/or increase the number of fighters that can be carried at once.

### Addressing the pirate life

Being a pirate for the player is very difficult right now. Some of the above changes would help then (reducing number of ships entering a system, increasing the sizes of systems), but more work could be done on top of that. My current ideas right now are to expand the size of pirate space by adding more systems around pirate regions, as well as adding more uninhabited landable planets across human space to give the player save points, and perhaps have "hidden" pirate outposts for the player to refuel on.

</details>


Amazinite:
I imagine the Rainmaker was left intentionally slow as an anti-frustration measure given its missile boat behavior, but I've significantly buffed its speed with the changes here. Given that the Navy campaign is in the works, I'd rather the Rainmaker be a ship worth using for the player who has joined the Navy rather than a ship we intentionally cripple in order to make things easier on the player fighting the Navy.

arkhne:
As somebody who has experimented with increased mass (and vanilla engines), including having flown a ship with 4k mass on 150t of Engines (think Heron, except that uses about 600t of engines stock), I can say that these mass numbers are definitely a step in the right direction.
I did not touch drag as, given vanilla system sizes, it's a pretty irrelevant stat.

However, you may recall that I've often been one of those advocating that we "not go too far" in massifying, and despite that, I have to disagree with your buff to Steering outfits. The problem with large ship speed is not Accel/Max (see above about system sizes), the problem is turn rates. The proposed changes here are mostly aimed at accel and max speeds, the steering buff acts to counteract the meaningful impact of mass changes, resulting in much less net effect than is needed.

As a result, I feel that, rather that fixing the problem, this compounds and cements its place in the meta. Increasing system sizes (including changes as proposed/started by Ferociousfeind) would go a long way to making accel and max speed stats relevant (some of Vitalchip's AI changes could also help with engagement ranges and general ship combat, too). However, fot such changes to be truly meaningful, the difference in ships does not need to be in Thrusters, but in Turn rates.

Stepping aside from Code/AI changes, I can approve of trying to solve at least some of these issues in a content only fashion (I even started on this myself, but some bugs in how ES handles some things in plugins has been sapping my motivation). More meaningful mass is definitely an important factor, and really helps with the big-ship feel.

Amazinite:
The buffs to steering ONLY applied to ion and plasma steering as a means of bringing all engines closer together, so any ship that has had its mass increased will now turn SLOWER when equipped with atomic steering as compared to before because of atomic steering remaining unchanged. And even with ion and plasma steering being buffed, the degree to which the mass of heavy warships has been increased is greater in this PR than the increase in steering, and about equal for medium warships, so the only ships which will actually see a buff to their steering rate are ships smaller than medium warships.

arkhne:
> >  resulting in much less net effect than is needed.
> 
> The buffs to steering ONLY applied to ion and plasma steering as a means of bringing all engines closer together, so any ship that has had its mass increased will now turn slower when equipped with atomic steering as compared to before because of atomic steering remaining unchanged. And even with ion and plasma steering being buffed, the degree to which the mass of heavy warships has been increased is greater in this PR than the increase in steering, and about equal for medium warships, so the only ships which will actually see a buff to their steering rate are ships smaller than medium warships.

I was not suggesting that it had no effect, just less effect than is needed. Also, I do wholly agree with bringing in more balance between engine types (I had a spreadsheet dedicated to just this, but abandoned it due to lack of interest). 
I was done with the first phase of tweaking, and the next was going to be addressing more in-game realities of it (such as credit:thrust, and how it impacts specific stock ships, especially those in the storyline).
In my spreadsheet, I had specifically buffed Ions, while nerfing Atomics (the Plasmas were basically used as the middle-ground point for human engines).

> 
> So the steering changes haven't really counteracted the mass changes in terms of turn rates because of how the peak has only been brought down for larger ships; I just also brought the trough up a bit so that the differences between the fastest and the slowest aren't as drastic. So if you simply think that the mass changes weren't enough, then we can look at increasing them further.

In a vaccuum, the mass changes on the whole are, as I said, very close to what I came up with, meaning I pretty much agree with them. I think it needs more eyes than just yours (or mine) to go through and look at specific ships and debate on whether they need more or less (for instance, in my rebalance, I'd massed Kestrel to a slightly less extent than other Heavies, because it is meant to be agile).

> 
> Since you've messed around with ideas on your own end, what turn rates would you consider acceptable for heavy warships and other categories? The results table in the PR descriptions shows the current values for all ships, default and with a custom set of atomics, and heavy warships have turn rates typically between 60 and 90 right now with atomics. Are you thinking of turn rates closer to the say 30 to 60 range? This is where the Behemoth and Bulk Freighter lie right now, between 30 and 40 when using atomic steering.

I think that the *general* Heavy Warship turn rates need to be around 40-60, with outliers in the Agile department using the best Engines (or particularly heavy ships using worse engines). At 60-80 turn rates, Heavies still don't really have a downside, at 90 Turn rate, there's no reason to not use a Heavy. If a ship can't reach 50 Turn with Korath engines, and it isn't an outlier for Lore reasons, then that is going to be a problem for most players (I know some are more than fine with 20-30 turn rate).

I find that 40-60 for Heaviers is a compromise between between creating a disparity between the Lighest and Heaviest ships, without impacting the nyoom factor to the point that it becomes painful for, what I would assume to be the average player (not speed obsessed, nor speed adversed).

Beyond this, a goal of 70-100 Turn rate for Mediums is probably also a good spot, with 100-150 for Lights being a little on the high side, but perfectly achievable. This leaves room for Fighters, Interceptors, and Drones to have more agility, especially if we consider giving them some innate thrusters (it doesn't have to be much, but it has to be an accel bonus without mass increase, since the majority of a Fighter/Drone's mass is in outfits).

Vitalchip's suggestions for mass and speed (based on testing AI changes and the speed disparity needed between Fighters and Heavies) is a little harsher than the numbers I have given here, but not so much that it has no effect on the overall outcome of combat.

I like that the changes here are a little more thorough and less scatter-shot than the Revamp (although this has given room for me to be of more help there), and I definitely don't want to sound too negative of it all, I just want to try and help get a good outcome (perspective matters, one person, no matter how knowledgable, cannot perceive everything). Like I said, this is much closer to where we need to be than vanilla, so it's a step in the right direction.

Zitchas:
I haven't had the chance to test any of these changes; but I do like the looks of them.

> I like that the changes here are a little more thorough and less scatter-shot than the Revamp (although this has given room for me to be of more help there), and I definitely don't want to sound too negative of it all, I just want to try and help get a good outcome (perspective matters, one person, no matter how knowledgable, cannot perceive everything). Like I said, this is much closer to where we need to be than vanilla, so it's a step in the right direction.

People shouldn't view the "Combat Revamp" and the "Combat Test" as competitors. It'd be more accurate to say that the "Combat Revamp" is the "Brainstorming/Experiential testing," while "Combat Test" is the carefully thought out accumulation of changes. They both have their uses, and the biggest use for the first one is simply to break the ice and show what is possible.

Amazinite:
I will admit that when making these changes, I purposefully sought to not go too far on reducing heavy warship turn rates, so to hear that people think that things are on the right track but just shy of what's desired is good news.

source/Ship.cpp
```diff
@@ -3103,9 +3103,9 @@ void Ship::ApplyForce(const Point &force, bool gravitational)
 		return;
 	
 	// Reduce acceleration of small ships and increase acceleration of large
-	// ones by having 30% of the force be based on a fixed mass of 400, i.e. the
-	// mass of a typical light warship:
```
Zitchas, line 3107: Huh. The things I never knew because I don't know the engine code. So acceleration isn't purely based on thrust and mass, and actually has an artificial limiter arbitrarily slowing down the fast ships?

Now I'm curious what simply removing this would do. Some of the changes done in both Revamp and Test are based around small ships needing to be a bit faster relevant to big ships.

arkhne:
> I will admit that when making these changes, I purposefully sought to not go too far on reducing heavy warship turn rates, so to hear that people think that things are on the right track but just shy of what's desired is good news.

This was precisely the situation I was going for, too. Good to see that we had the same approach, and both fell so close together.

source/Ship.cpp
Amazinite: This only applied to acceleration caused by hit force weapons.

Zitchas: Ah, that's good. Thanks for the explanation.

Zitchas:
I'm not seriously playing this right now, but I have tried out a bunch of ships in World Forge. 

Starting off with a fleet of Heavy Freighters, I had an amusing duality experience. My first thought was "these things still turn far too fast." But then I loaded them up to the brim with cargo, and now they feel more appropriate. And I suppose that makes sense, in retrospect: The point of heavy freighters is to haul heavy cargo - just considering them empty would be like trying to analyze a warship's performance without any weapons on it. Worth doing, but it is only a small piece of the picture.

Anyway, a Leviathan with full cargo holds feels appropriately weighty. Now to get something to shoot me with a few heavy rockets and see how it is impacted.

Zitchas:
Just a note: I liked (most of) your changes so much I just merged your branch in to the Combat Revamp. Some points that I wasn't as fond of and used a hybrid of:
- Skipped the drones and fighters and stayed with my previous changes to them.
- A number of freighters got diverging treatments from Amazinite and myself: I upped mass slightly, but didn't change drag. Amazinite didn't change mass, but reduced drag. I combined the two so they have my slightly higher mass, but also got Amazinite's lower drag. So, higher max speed, but slower acceleration. The following ships got this treatment:
  - Behemoth
  - Bulk Freighter
  - Hauler II
  - Hauler III
  - Star Queen
- Heavy Shuttle kept my change to bunks (up to 10, but Amazinite's changes to mass and drag (he had higher mass, but lower drag)).
- Given the Osprey's description as being built along the same lines as the blackbird (which is an agile light warship), which implies that the Osprey should be a high-tech light and agile medium warship, I think that it isn't light or agile enough in compared to the rest of the human medium warships. Leaving with Amazinite's changes for now, but I suspect I'll be dropping both mass and drag a little bit later on. (I had it at 284 and 5.9; Amazinite's is at 400 and 6.3)

ravenshining:
My only concern with this as it is currently applied (nevermind planned) is that it way decrease support for the idea of buffing fighters by making them smaller and more plentiful on carriers.  I'd rather see that happen than this, in fact, several months ago I was anticipating it would happen while I was remaking the Raider sprite; I specifically sized the bays to fit two 75% sized Chasers.

Amazinite:
I don't necessarily see this as being mutually exclusive with any further changes to fighters to reduce their sizes and/or increase their numbers. When compared to the existing sizes and speeds, would a smaller ship not make sense to also be made faster?

Something I've also mentioned before is that I think that this actually helps open the door for increasing fighter numbers more. By making fighters ever so slightly lighter while also increasing the mass of larger ships, which include the carriers in the game, the negative effects of increasing the number of fighters on carriers (more carried fighters = more mass = worse handling carriers while fighters are docked) becomes less drastic given that each additional fighter is now much less of the overall mass of the ship, whereas without any of the mass changes from this PR, increasing the number of fighters on carriers would make them very slow and perhaps unpleasant to use compared to other ships while the fighters are docked. (While in combat your fighters will likely be deployed, most of the time when you're flying around they'll be docked, so we want the speed of a laden carrier to be reasonable.)

Amazinite:
@arkhne I just gave the heavy warships another bump in mass. How does that feel now?

Zitchas:
Overall, a very nice set of changes. I like the way things are looking here.

The one exception is probably heavy rockets. I continue to be of the opinion that the heavy rocket / torpedo split shouldn't be "one is for groups and the other anti-capital" but rather "one is for direct-fire capital-killing, and the other is for long-range capital hunting. In my view, the main purpose of a heavy rocket is to have something in which one can sacrifice *all* the guidance (and some of the propulsion) in order to make the single biggest damage-dealing thing possible. Virtually useless at range or on slow ships,  but extremely lethal in the hands of a fast ship. These are basically the "dumb fire" old-school torpedos used by torpedo boats or subs back during the WW era.

And yes, a direct hit by a heavy rocket *should* obliterate a shuttle. Or a starbarge, or a sparrow. It should have a similar impact to a FGM-148 Javelin hitting... well, basically any civilian vehicle in the present day. 

The problem with heavy rockets isn't how deadly they are, it is with how *common* they are. Capital-killer rockets shouldn't be turning up in loadouts of anything that isn't intended to punch a hole in medium warships or bigger.

A secondary problem is how fast they are (both in terms of fire rate and movement). This is why I slowed down both in Revamp. Heavy rockets in Revamp harken back to the classic EV where they were literally the definition of "weapon that will kill you if it hits but is slow enough that pretty much any light ship with an alert pilot can dodge it." This shifts the focus from tanking the rocket to *dodging* the rocket. And dodging is something that is far, far too rare in ES. Yes, I know the AI is terrible at dodging, but that's a different problem. There is no reason why we can't encourage the *player* to dodge things, and the first step to that is giving them things that *can* be individually dodged. And there is precious little of that in ES, as virtually everything seems designed to overwhelm defenses with numbers.

Amazinite:
> I continue to be of the opinion that the heavy rocket / torpedo split shouldn't be "one is for groups and the other anti-capital" but rather "one is for direct-fire capital-killing, and the other is for long-range capital hunting.

Well the problem with that is that the stated purpose of the heavy rocket as per its description is the exact opposite of this; it *is* the heavy rocket for groups and the torpedo for anti-capital. So changing the stated purpose of the heavy rocket is a separate issue entirely.

And if we were to change the stated purpose of the torpedo, I feel like making it less common would then make less sense. If the heavy rocket is designed for ganking larger ships, why *wouldn't* we see it so often on smaller ships? But if the heavy rocket is designed for taking out groups of smaller ships, then it naturally concludes that you wouldn't see it as often on smaller ships and would see it more often on larger ships, especially because a smaller ship is more at risk of hurting itself from its own blast radius that is explicitly stated as being to damage smaller ships.

In any case, the commonality of certain weaponry, heavy rocket included, is something I will be addressing (see "Addressing #4158 + human fleet strengths in general" and "Addressing #4077" sections).

Zitchas:
> Remnant thrusters have received a roughly 25% nerf.

I'll reserve judgment until I actually test it and see how they feel in conjunction with everything else, but I can't say I like the sounds of it. It's going to be hard for the Remnant to *run* with a 25% nerf...

Amazinite:
Note that all other high-performing thrusters have received similar nerfs (~15-25% nerfs for human atomic, Hai atomic, Wanderer, and Korath thrusters), so Remnant thrusters are still as about as powerful as they were before in terms of their ranking versus other engines. Also keep in mind that Remnant ships can still be given lower drag values so that they're generally faster moving ships, and you still have the Bellows afterburner that has remain unchanged that is now comparatively stronger than typical thrusters given that the highest performing thrusters have been nerfed.

Zitchas:
Without testing it I can't say for sure, I'd be inclined to swap that around to 5x the mass and 4x the drag. I seriously intended the Heron to be an example of what I felt a true "cityship" like the World Ship *should* be like. Of course, it is built to be an evacuation ship and capable of running, which is a bit different from the World Ship; but running isn't the same as dogfighting. There should be no dogfighting in the Heron.

Zitchas:
Follow-up: I just did some testing, and my suggestion would be to 5x the mass, but leave the drag as-is or maybe even drop it down to 20. The acceleration is painful (the bellows are pretty much a requirement), but it can still get to a fast enough speed to outrun most things. But stopping or changing directions? That requires some planning.

tldr: 5x the mass for the Heron looks good to me. The drag, not so much.

Amazinite:
High max speed but low acceleration isn't a balance paradigm that we work by and frankly isn't one I want to dip into, especially because I've been spending loads of work making sure that larger ships have lower max speeds than smaller ships, so I don't want to suddenly have one exception in the Heron where it can be as fast as a light warship if you wait long enough.

Zitchas:
Well, to be honest, I like the idea of the Heron being a rule-breaker. In EVN there was a trio of Vellos ships and the Raven that completely broke with the way every other ship in the game worked, which was really neat. The point is, not everything needs to work the same way. Especially when the exception is a person ship.

Amazinite:
Alright, that's the last of the engines and ships. I won't make a massive table like I did for the human ships because that was a pain to compile, but here are some general stats about the different categories for all ships in the game:

Category | Avg. Mass (old/new) | % Change | Avg. Drag (old/new) | % Change
----|----|----|----|----
Transport* | 495 / 612 | 23.64%** | 4.13 / 3.84 | -6.94%
Light Freighter | 438 / 457 | 4.3% | 4.98 / 3.19 | -36%
Heavy Freighter | 1178 / 1541 | 30.9% | 12.53 / 10.48 | -16.3%
Drone | 104 / 103 | -1.2% | 0.70 / 0.59 | -15.8%
Fighter | 141 / 143 | -1.25% | 0.78 / 1.0 | -22%
Interceptor | 291 / 294 | 1.2%*** | 2.01 / 1.67 | -17%
Light Warship | 538 / 604 | 12.24% | 4.21 / 3.74 | -11.1%
Medium Warship | 870 / 1070 | 23% | 6.93 / 6.46 | -6.81%
Heavy Warship | 1601 / 2639 | 64.79% | 12.27 / 12.60 | 2.71%

\* Excluding the Heron.
** The increase in average transport mass is largely because of the Auxiliary. Transport ship masses were largely untouched.
*** The increase in average interceptor mass is because I increased the mass of the two T2 interceptors by a fair amount. All human interceptors have received a reduction in mass.

And as for engines
Thrusters | Avg. Thrust/Engine Space (old/new) | % chance
---|---|---
Ion* | 1747.8 / 1747.8 | 0.00%
Wanderer | 2521.0 / 2143.7 | -14.97%
Plasma | 2236.2 / 2236.2 | 0.00%
Coalition | 2793.8 / 2507.5 | -10.25%
Hai* | 3400.8 / 2543.4 | -25.21%
Atomic | 3239.9 / 2556.3 | -21.10%
Remnant * | 3431.1 / 2571.8 | -25.04%
Pug | 2604.3 / 2604.3 | 0.00%
Ka'het** | 3678.5 / 2833.9 | -22.96%
Korath | 3742.4 / 3088.4 | -17.47%

Steering | Avg. Turn/Engine Space (old/new) | % chance
---|---|---
Ion* | 1037.6 / 1306.6 | 25.92%
Wanderer | 1467.0 / 1539.0 | 4.91%
Coalition | 1620.6 / 1620.6 | 0.00%
Plasma | 1323.1 / 1684.9 | 27.34%
Pug | 1477.1 / 1845.9 | 24.96%
Atomic | 1867.9 / 1867.9 | 0.00%
Hai* | 1961.5 / 1961.5 | 0.00%
Ka'het** | 2345.4 / 2070.2 | -11.73%
Remnant* | 2244.0 / 2131.6 | -5.01%
Korath | 2155.2 / 2155.2 | 0.00%

\* Excluding the tiny hybrid engines.
** Ka'het engine space used is as a factor of how much of the entire engine space of a Ka'het engine unit is dedicated to thrusters or steering instead of taking the entire engine space for thrusting or steering.

Amazinite:
There could definitely be more tweaks to engines and ships relating to speeds from here, but I think that's the bulk of the work done for the direction I want this to go.

Zitchas:
Overall, looks good. I'm kind of disappointed that fighters aren't ending up significantly faster. I'm generally of the opinion that fighters as a category should all be >1000 unless there is a good lore reason for them to not be. For instance, those that aren't actually intended to be fighters, but are just deployable weapon launchers/cargo/passengers.

source/Ship.cpp
```diff
@@ -3111,9 +3111,9 @@ void Ship::ApplyForce(const Point &force, bool gravitational)
...
-	acceleration += force * (.3 / 400. + .7 / currentMass);
+	// ones by having 20% of the force be based on a fixed mass of 500, about the
+	// mass of a human light warship:
+	acceleration += force * (.2 / 500. + .8 / currentMass);
```
Amazinite, line 3116: I'm tempted to simply remove having a fraction of hit force be based on a fixed mass altogether.

Given the values I'm currently using, the lightest ship in the game, the Surveillance Drone with a mass of 63, feels 18% less hit force because of this equation. Then as you get closer to 500 mass, the % reduction in felt hit force gets smaller. On the opposite end, once you get to 1000 tons of mass (about the mass of a Mule or Behemoth under this PR), you're feeling +20% hit force, and it only increases from there. The Protector, the lightest heavy warship at 1400 tons, is feeling +36% hit force. The Cruiser at 2600 tons is feeling +84% hit force. The KIV at about 4000 tons is feeling +140% hit force. The Emerald Sword at 4350 is feeling +154% hit force. Then the Heron at 28600 tons is feeling +1124% hit force.

So really, all this does is reduce the hit force on smaller ships by a negligible amount while MASSIVELY increasing the hit force on larger ships. So either we could just outright remove this weighting (with #5978 we could keep the ability to have it and just weight the left side 0), or massively up the fixed mass. If we increased the fixed mass from 500 to 2000, the Surveillance Drone would go from feeling 18% less force to 19.5% less force, so on the low end it really doesn't change anything, but on the higher end a 4000 ton ships will only experience +20% hit force, and the Heron will feel +266% hit force. The Heron, or any ship that is wildly more massive than anything else, will always be an outlier under a system like this, but a fixed mass of 2000 makes more sense to me than 500, as it's the balance point where the smallest and largest ships feel a similar change in hit force (about -20% on the low end and about +20% on the high end).

Zitchas: I rather like the idea of just removing the modifier entirely. Having the hit force modifier makes sense if the assumption is that all the ships are going to be clustered in the middle and one is trying to balance all these similar ships to be affected fairly similarly.

Honestly, though, a) that's boring, and b) unrealistic. If a surveillance drone gets whacked by the force of a heavy rocket (and by some miracle survives), I *expect* it to get booted across the system. This is not a problem as far as I'm concerned. On the other hand, when a world-ship gets hit by a heavy rocket, it should barely register at all.

vitalchip: I'm all  for removing the fractional application of force but just to put an alternative on the table, we could fix the formula. It's incorrect.

Currently:  acceleration += force * (.3 / 400. + .7 / currentMass);

Should be: acceleration += force / (.3 * 400 + .7 * currentMass);

I suspect an accedental transposing of * and / way back and not until fairly recently has there been a ship with enough mass to really highlight the fault.

Run the numbers and see what you think.

Amazinite: > It's incorrect.

No, the formula is working as intended.
image

Zitchas: For comparison, I have removed the hit force adjustment from Revamp entirely. So anyone that wants to see how impacts feel in various ships without any formulas artificially changing things can do so.

vitalchip: I'm aware of the code comment and it is somewhat ambiguous but I think if you run some numbers for both you'll see.

Amazinite:
Pushed the most recent commit just because I was doing the work for it and people said they were interested in seeing higher accelerations for fighters, but I'm not too much a fan of the change as a whole. Even if tripled acceleration works out on fighters and drones, acceleration now feels far too high on everything else. And in order to use the existing system to lower the acceleration on everything else, I'd need to *triple* the mass of everything just to get them back to the accelerations they had before, which I think felt fine. But tripled mass also means tripled turn on all steering, which will make fighters and drones absolutely spin like tops. It's an absolutely impossible balancing act between thrust, turn, mass, drag, turn rates, max speeds, and accelerations. So as cheaty as it might be, I'm just going to create an acceleration multiplier attribute that we stick onto fighters.

Zitchas:
Well, triple the mass is is good, in my view. And slowing down turn rates I think is actually a positive. Reducing it to a third of normal might be overboard, but reducing things *by* a third would be very beneficial for most of the "heavy" ships in the game.

That being said, it is kind of amusing to note that in Revamp I ended up just *giving* every fighter and drone the built-in equivalent of its stock engine setup. (I also doubled the engines that they come with stock). Which works out to 3x the acceleration, but without needing to triple the mass of other stuff.

Amazinite:
> And slowing down turn rates I think is actually a positive. Reducing it to a third of normal might be overboard, but reducing things by a third would be very beneficial for most of the "heavy" ships in the game.

Remember that I've already slowed the turn rates down on larger ships, and more than doubled the mass of some, so tripled mass without anything else would be a net 6x mass and absolutely abysmal turn rates potentially in the single digits for some ships.

>That being said, it is kind of amusing to note that in Revamp I ended up just giving every fighter and drone the built-in equivalent of its stock engine setup. (I also doubled the engines that they come with stock).

Well I'm not really interested in giving fighters built in engines, or increasing their engine space, therefore requiring fighters to be reoutfitted and making fighters even more expensive than they are right now, which I don't think is something we want when replacing fighters is a concern.

Zitchas:
Sorry, I wasn't saying that you should follow Revamp's route; rather I meant it was interesting that we both ended up in the same place of needing to give fighters more power than they currently have. We solved that in different ways, which is just fine. Your method is probably simpler and more elegant.

On the physics side, having different "acceleration multipliers" for different ships actually makes sense. Different hulls are going to have different performance traits. To use a real-world example: Strapping an F1 racecar engine into a volkswagon cargo van isn't going to get the same sort of power & efficiency as it had in the racecar designed for speed. In-game, a ship like the Flivvr or most fighters is going to have a structure that is much better capable of handling and making use of high power engines than, say, a freighter. 

I could definitely see such a multiplier basically representing, as a holistic concept, how well designed the ship is for handling high-g acceleration and sudden shifts in velocity. The bigger the ship, and especially the less a ship is designed for massive changes in velocity, the more cautious it is going to have to be in applying thrust. In an ultra-realistic simulator style game this might be more accurately represented in having a "stress guage" (kind of like Vne on aircraft or the redline on the tachometer on cars) that the player has to be careful to stay below; but for ES it makes more sense to just have said limit built into the controls under the premise that our onboard systems (and engineer, if we have one) just won't let our ship run the engines at levels that will actually break the ship.

The more I think about, the more I am of the opinion that such a multiplier actually decent sense, and certainly contributes more "realism" than our drag value does.

The only tricky part, really, is explaining to the player that the force described by the engine entry isn't a universal thing. 1 N of force on a Fivver is not going to get the same result, even on a per-ton basis, as on a bulk freighter. The one is optimized for speed, the other is optimized for bulk. Most players don't know the formulas anyway, so adding one extra variable shouldn't be too much of a problem.

vitalchip:
An acceleration multiplier is certainly a very decent 'quick and dirty fix' for the problem and I'm not averse to such things for testing but not as a 'solution'.  we've had a lot of attempts at balancing speed/acceleration/turning/mass etc and if we need to resort to tricks then what's clear is we don't yet have it right.

Drones and Fighters suffer from the sheer mass and inefficiency of the outfits available, this is a direct result of trying to use the same generators/weapons/shields/engines for capital ships and fighters, one end of the scale is always going to lose.

The problem can be resolved properly but it'll take a full re-evaluation of how ships operate. We would need to start from the bottom and work up. Fighters need better thrust/mass engines and better energy/mass generators to nake them viable and not end up as 300 ton bloat ships. Once that's done we need to look at the bigger ships and stop giving them 'freebies' which let them use far lower resources than they should. 

For example a ship with 800 crew, somehow, uses no more energy for 'day to day' operations than a ship with 1 crew. So the larger ship is effectively getting free generation, whereas it should require to install that generation capacity, in turn leaving less outfit space for (eg) engines.  Having a higher generation/mass ratio and therefore a lower engine/mass ratio is one way we get the acceleration/turning differential between small and large ships.

TLDR: We're trying to build gliders out of parts off a 747 and wondering why it flies poorly.

Zitchas:
And we can't even look to EV/O/N for inspiration in this case, because those games actually sidestepped the whole issue by simply having all the movement attributes be built right into the hull.

Actually, one idea that might help with this a bit: What if ships actually had a limit on number of engines/steering installed? For discussion purposes, what if all ships in human space (being the baseline) were strictly limited to 1 engine and 1 steering, along with the current engine space limits. The justification is that a ship (based on the sprite) has a fixed number of nacels and hardpoints. Just like we can't "just strap an extra gun to the hull, even if we have weapon space available" why can we "just strap another engine to it, provided we have space?

If a ship is limited to a single engine, then we could have smaller engines that are more efficient to provide the extra thrust efficiency we want in smaller ships without worrying about it the usual concern of "why wouldn't big ships just use a dozen small engines instead?" Because they only have a single engine hardpoint. Sure, they *could* stick a single high efficiency engine on, but that takes up their available nacelle, so they'll probably be moving pretty slowly due to the mass and whatnot. It also means that there's no "filling up the extra engine space" with other engines if the selected engine doesn't take it all up. 

Extending this farther, it might actually make Coalition ships feel more unique, as they are built for a modular system, so they probably *would* have several engine hardpoints to use. Of course, I'm sure certain ships, particularly alien/faction ships, might have more than one engine hardpoint as something that differentiates them, but it probably wouldn't be standard for this to be the case. 

Such a mechanic would also serve to de-emphasize the sheer "thrust per space" efficiency value of engines as being the overriding metric for comparison. If you have 70 space, it doesn't matter if the 50 space engine has massively better thrust to space than the  70 space engine, if the 65 sized engine still provides more total thrust, since otherwise that 20 space is just going to be unused. (although it might make afterburner use more common, since a 50 engine + afterburner could be very viable when  a 50 engine + a 20 engine is no longer an option. Not necessarily a bad thing)

ravenshining:
in too much pain to really elaborate right now, but this whole acceleration multiplier nonsense seems to me to be just that, nonsense.  magically the same thruster exerts more force because it's installed in a finch instead of a sparrow?  it makes no sense.  even less sense than drag.

i'd much rather see fighters and drones get some efficient-but-inbuilt thrusting capability.  i know you don't like that, but it makes sense for a fighter to be built around a specific engine rather than trying to shoehorn a modular engine into a fighter.  we could do some combination of inbuilt and modular engines -  say the fighter has its integrated engine, and you can in addition strap on a booster underwing instead of an extra missile launcher or something.

generally I've never felt that acceleration is a problem in the first place, i think most ships accelerate too fast.  if anything I think fighters turn too slowly, and there a mass reduction would serve both purposes without having to invent a new arbitrary multiplier.

Zitchas:
If one can accept that the effective acceleration of a ship isn't just a basic application of thrust & mass, but rather a calculation of one's onboard systems that takes into account how the ship is designed and laid out, then it becomes quite reasonable. All structural members aren't equal, and one's onboad computer would take that into account. It is very likely, to the point of actually being a "I'd be surprised if it *wasn't* that way" sort of thing, I would expect the structural design of a starbarge to be significantly different than for a finch. In a real-life comparison, the strength of the structural elements in the wing of a cesna 182is a lot lower than in that of a F-15. It basically boils down to a cesna 165 just wasn't designed to be able to handle changes in velocity that an F-16 can. (For reference, a Cesna 182's flight load factor must remain [between -1.76 and 4.4g.](https://longislandaviators.com/wp-content/uploads/2018/08/1978-Pilots-Operating-Handbook-Cessna-152.pdf) The F-15, on the other hand, is described as being limited to 9g for extended events with instantaneous g's [up to 13.5g](https://www.f-16.net/forum/viewtopic.php?t=10663) depending on configuration.

That being said, I do agree that it seems weird that two ships that share so much in common wouldn't have the same (or at least very similar) acceleration profiles. As such, I think that most interceptors should also have this acceleration multiplier, perhaps at a reduced value. (say, 1.75 perhaps) After, many interceptors are basically described as "fighters with a hyperdrive," and the ones that are such should have matching abilities. Naturally, however, the increase mass on the interceptor would result in the end acceleration being a little lower, which makes sense, but it shouldn't be *massively* lower like the absence of the multiplier would result in.

ravenshining:
> If one can accept that the effective acceleration of a ship isn't just a basic application of thrust & mass, but rather a calculation of one's onboard systems that takes into account how the ship is designed and laid out, then it becomes quite reasonable. All structural members aren't equal, and one's onboad computer would take that into account.

That's what we have engine space limitations for.  A total thrust limit would make more sense to model what you are describing, artificially limiting thrust regardless of how much thrust is applied makes no sense.

If I have a spaceframe that can handle 100,000 thrust, and I install a thruster that makes 120,000 thrust, it would make sense to for the ship's computer to limit the thrust to 120,000.  It would *not* make sense for the ship's computer to limit a 60,000 thrust engine to 50,000 on the same frame.

I can think of only two ships in the game for which this kind of mulitplier would make any sense - the Fury and the Osprey, because their engines fire at an angle.  If you want to invent a whole new mechanic to give those two and only those two ships a thrust penalty in exchange for a turning boost, sure, though that feels even more bloaty.

> Naturally, however, the increase mass on the interceptor would result in the end acceleration being a little lower,

That's what we have mass for!

Zitchas:
Two questions for you:
1) Would you be able to split out the acceleration multiplier code into a separate PR that could be merged independently?

2) What sort of timeline do you anticipate for this? I'd love to see it merged sooner rather than later. (If this was being merged in the near future I'd close my fighter engine change PR)

Zitchas:
@Amazinite Would it be possible for you to bring this up to speed with master?

a) I'd like to try this out with all the changes that have taken place since June,
b) the auto-builds have expired, so I can't download the executable with your changes in it anymore.

Thank you!

Amazinite:
I’ll update this (and #6050) in a bit. 

As far as my timeline for completion goes, I wanted to first knock out my sprite scale and variants in variants PRs, then move on to the conversation actions and gamerules PRs. The former two PRs just need fully reviewed by tehhowch, the conversation actions PR needs a rewrite that I don’t anticipate being too difficult, and the gamerules PR basically just needs any form of feedback on the initial implementation before I move forward; I’m a little surprised that the concept got a lot of praise on Discord and when I posted the issue there was a lot of chatter there, but the PR is a graveyard. Even still though, that PR definitely needs polished up anyway. 

How long all that takes? That’s largely up to tehhowch and others reviewing my PRs. I’d like to focus on those PRs before getting to the balance ones though since I feel like they’re simpler to deal with; they can only add to the game experience (some in some very exciting ways, allowing us to do things we’ve never been able to do before but have been long requested), while any sort of balance changes alter the game experience from what we’re use to. 

Once I finally get to the combat test and economy test PRs though, I do have intentions to break them up into smaller PRs where possible for easier reviewing and merging. The various mining changes from #6050 for example could be their own PRs. 

Since we do have a number of competing PRs up on changing combat balance, I’d ask that you mark your latest PRs as drafts. (e.g. increasing gun ports on smaller ships, speeding up fighters.) I’d rather not have a balance PR slip past me while I’m taking a bit of a break from things after I spent so long on this PR. Not that I’m saying that my PRs will be the ones that merge, as I know you’ve spent plenty of your own time on this front and I’m sure we could pick pieces from each of our works for the final results, just that since I have stake in this I’d prefer if I’m able to be a part of the conversation on balance changes before anything is merged.

Amazinite:
The changes to stars threw up like 150 merge conflicts with the map file, so instead of manually fixing them all I just chose to keep the map file changes from this PR. Before merging though what I'll need to do is go through the vanilla map file and make the same changes I did to it as before in order to get a proper map file for this PR.

Zitchas:
Thanks for updating!

I've marked #6247 as draft. That being said, I actually like **this** PR better than mine. Adding in that acceleration multiplier is a very elegant solution, not to mention a simpler one that retains the player's ability to customize everything. I'm not as confident in the 2x multiplier used on the fighters, but that's much easier to change than the whole built in engine aspect. In some respects, I actually kind of consider #6247 a bandaid solution to improve fighter combat while we wait for a more complete solution such as this one. The main strength of built-in engines is that they allow us to nail down certain ships as having specific strengths. Like having fast fighters with poor turning (bombers) or slow fighters with great turning (deployable turret style things) or other combinations.

Saturnalia9:
> I think that the _general_ Heavy Warship turn rates need to be around 40-60, with outliers in the Agile department using the best Engines (or particularly heavy ships using worse engines). At 60-80 turn rates, Heavies still don't really have a downside, at 90 Turn rate, there's no reason to not use a Heavy. If a ship can't reach 50 Turn with Korath engines, and it isn't an outlier for Lore reasons, then that is going to be a problem for most players (I know some are more than fine with 20-30 turn rate).
> 
> I find that 40-60 for Heaviers is a compromise between between creating a disparity between the Lighest and Heaviest ships, without impacting the nyoom factor to the point that it becomes painful for, what I would assume to be the average player (not speed obsessed, nor speed adversed).
> 

First time participating on anything in GitHub, hello! I feel like 60-80 turn rate is way too high for HWs. I like the idea of some of them going down to 20-30 for the absolute largest, with the rest being between 30-50 even. Having too much turning on a HW takes away some of the HW specialty, in that they are usually tanky and are often known to have many turrets. For something like the Vanguard and SB, I could see it having above-average turning since they have so many forward-facing ports.

As mentioned in Discord (https://discord.com/channels/251118043411775489/308902312741568512/946086370940686356), I do think this could be an idea, taking HW turn rate into consideration with turret turn radii as net turning.

As for the rest of the PR, I like the ideas!

Amazinite:
Closing since this PR is outdated and I plan to start breaking pieces off of this shortly.

PR #5784, opened 2021-03-13
Fix clunky wording in FWC Pug 4B

Anarchist2:
The dialogue used during FWC Pug 4B when you accept Freya's request to check with the Quarg is particularly clunky.

Freya asks the player that they should try to go to the Quarg, the player responds with yes, then Freya asks the player to trust her on this, even though the player has already said they wanted to go, then the player replies saying that "it can't hurt to try", despite having already explicitly saying that they wanted to go.

This PR changes it so that when the player accepts Freya's request, she only thanks the player and expresses her hope in getting useful answers.

### DIFF
--- a/data/human/free worlds checkmate.txt
+++ b/data/human/free worlds checkmate.txt
@@ -1876,7 +1876,7 @@ mission "FWC Pug 4B"
 			`	"Well," she says, "you're the captain, I'll leave that choice up to you."`
 				accept
 			label detour
-			`	"Thanks," she says, "let's just hope that we can get some real answers."`
+			`	"Thanks," she says, "Let's just hope that we can get some real answers."`
 				accept
 	on visit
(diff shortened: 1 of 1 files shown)


Zitchas (approved):
That makes it flow better.

The old lines sound like they actually belong the branch where the player says we should continue to Arabia.

If I'm following the mission correctly, the player is always sent to the Quarg regardless; with the only difference being that one dialog option has the player agreeing to it, and the other has Freya saying that she'll leave the choice up to them. 

The replaced lines sound like something that belong on the "I'll leave it up to you." path.


Alternative interpretation: After the player agrees, Freya is trying to convince herself that it is the right choice, and the player's statement about the Quarg is them reaffirming their dialog choice to go talk to the Quarg.

Pointedstick:
Agreed. Good change.

PR #5791, opened 2021-03-14
Fix: Remnant: Keystone Deliveries and Research sequencing.

Zitchas:
**Bugfix:** This PR addresses an issue reported by EjoThims on the Discord

Somehow, EjoThims managed to get all the way through the Remnant story-line on continuous, up to and including completing the current Cognizance and Astral research missions, and then starting on the Crystal Research mission set, and only then triggering one of the early crystal delivery missions.

Two problem: 
1) That the Crystal Research mission can be triggered before the Keystone Deliveries
2) That the Crystal Delivery missions can block each other even without being accepted or completed.

## Fix Details
1) Crystal Research now requires that one of the three Keystone deliveries is completed.
~2) Keystone delivery missions now only block each other if they are active or done.~

Part 2 has been removed to be considered and dealt with in another PR, perhaps including branches or alternate missions.


Amazinite:
I don't think that the changes to the offer conditions on the initial keystone missions make much sense. Even if there are slight differences between the three potential introductions, they're still all written as if it's your first time. And declining a mission typically means it won't show up again, so it would seem weird to me if I were to decline this mission then later get nearly identical missions that don't reference me previously declining the offer from the same person.

Leaving that aside though, I don't even think that this is the best route for the stated purpose of giving the player a second chance. Some of the versions of this mission won't offer of certain things have already occurred. You get a version of the mission if you've haven't met the Hai, if you have but don't have a Hai keystone, and if you have and do have a Hai keystone. A player who declines the mission where they haven't met the Hai has two potential extra chances with those other two versions of the mission, but someone who shows up to Remnant space after having met the Hai only gets one extra chance if they show up without a Hai keystone, and I'd say realistically they're given no extra chances if they decline the mission where they do have a keystone, as in order to get the other mission they'd need to remove their keystone, which they can't replace with a key stone until they get a license, which requires this mission. Or you can do the broken jump drive missions and potentially replace your keystone with a key stone, which would then cause the other version of this mission to offer much later on, which seems odd to me.

The update to the offer conditions for the later mission are fine, though. So I'd say let's just make that change then if we want to provide second chances for these missions we can look into that later with a more fleshed out second chance system like we did with the second chance on the blood test a while back now.

Zitchas:
OK, I can do that. basically splitting this in two.

That being said, I wish I knew exactly how EjoThims managed to complete nearly *all* the Remnant content without getting the basic keystone retrieval.. .

Zitchas:
OK, made the change. Now this just adds the keystone missions as a prerequisite to the Research mission.

PR #5794, opened 2021-03-14
Ships get repaired to MinimumHull() + some extra

Amazinite:
**Feature:** This PR implements the feature request I thought of because I'm amazing.

## Feature Details
Currently when a ship is repaired, it gets repaired to MinimumHull(). Given that a ship becomes disabled if its hull is less than MinimumHull(), being repaired to this point means that any non-zero positive hull damage will immediately cause the ship to become disabled again. This can be especially frustrating when trying to assist a ship in the middle of combat, only to see it immediately become disabled again.

This PR changes it so that ships now get repaired to MinimumHull() * 1.5, giving some buffer room after a ship is repaired before it becomes disabled again.
Ships typically become disabled after reaching 15-45% hull remaining, with smaller ships becoming disabled sooner. Using MinimumHull() * 1.5 means that ships will get repaired to 22.5-67.5%, so smaller ships will gain more hull as a percentage of their max when being repaired compared to larger ships.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
Repaired some ships to observe that they were repaired to some level greater than just their minimum hull.
Before repairs.
image
After repairs.
image

## Performance Impact
N/A

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2175,7 +2175,7 @@ shared_ptr<Ship> Ship::Board(bool autoPlunder)
 		SetShipToAssist(shared_ptr<Ship>());
 		SetTargetShip(shared_ptr<Ship>());
 		bool helped = victim->isDisabled;
-		victim->hull = max(victim->hull, victim->MinimumHull());
+		victim->hull = max(victim->hull, victim->MinimumHull() * 1.5);
 		victim->isDisabled = false;
 		// Transfer some fuel if needed.
 		if(!victim->JumpsRemaining() && CanRefuel(*victim))
@@ -2689,7 +2689,7 @@ double Ship::TransferFuel(double amount, Ship *to)
 void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
 {
 	// Repair up to the point where this ship is just barely not disabled.
-	hull = max(hull, MinimumHull());
+	hull = max(hull, MinimumHull() * 1.5);
 	isDisabled = false;
 	
 	// Set the new government.
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2175,7 +2175,7 @@ shared_ptr<Ship> Ship::Board(bool autoPlunder)
 		SetShipToAssist(shared_ptr<Ship>());
 		SetTargetShip(shared_ptr<Ship>());
 		bool helped = victim->isDisabled;
-		victim->hull = max(victim->hull, victim->MinimumHull() * 1.5);
+		victim->hull = min(max(victim->hull, victim->MinimumHull() * 1.5), victim->attributes.Get("hull"));
 		victim->isDisabled = false;
 		// Transfer some fuel if needed.
 		if(!victim->JumpsRemaining() && CanRefuel(*victim))
@@ -2689,7 +2689,7 @@ double Ship::TransferFuel(double amount, Ship *to)
 void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
 {
 	// Repair up to the point where this ship is just barely not disabled.
-	hull = max(hull, MinimumHull() * 1.5);
+	hull = min(max(hull, MinimumHull() * 1.5), attributes.Get("hull"));
 	isDisabled = false;
 	
 	// Set the new government.


source/Ship.cpp
```diff
@@ -2689,7 +2689,7 @@ double Ship::TransferFuel(double amount, Ship *to)
 void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
 {
 	// Repair up to the point where this ship is just barely not disabled.
```
clifflocked, line 2691: This isn't extremely important, but I would change the comment here to something else.
```suggestion
	// Repair the ship so that it has a bit more than the minimum required hull, so that it isn't disabled immediatly afterwards.

Comment (clifflocked) [reply source/Ship.cpp:2691]: Also, "just barely not" is a bit wordy IMO.

File: source/Ship.cpp [hunk 2]
```diff
@@ -2175,7 +2175,7 @@ shared_ptr<Ship> Ship::Board(bool autoPlunder)
 		SetShipToAssist(shared_ptr<Ship>());
 		SetTargetShip(shared_ptr<Ship>());
 		bool helped = victim->isDisabled;
-		victim->hull = max(victim->hull, victim->MinimumHull());
+		victim->hull = max(victim->hull, victim->MinimumHull() * 1.5);
```
jerith, line 2178: Should this be capped at the maximum hull value as well? We don't want ships with more than 67% minimum hull to end up with more hull than they started with...

Amazinite: During Ship::DoGeneration there exists a check to prevent the hull (or shields) of a ship from being above the maximum, so it seemed redundant to me to also clamp the hull to below the maximum value here.
https://github.com/endless-sky/endless-sky/blob/1af961037f9133fe7a4b8fdce3033b2173db2025/source/Ship.cpp#L2015-L2018

source/Ship.cpp
Amazinite: Well I feel like ships are still just barely not disabled even after this change. They're just not on a knifes edge between disabled and not. \:)

source/Ship.cpp
tehhowch: It's likely appropriate to add the clamp both here and in `TakeDamage`, to preserve the class invariant externally (that is, a ship cannot ever have more "X" than "MaxX").

If no other class could possibly observe or influence X between when it is above MaxX and when it gets clamped within `MinX <= X <= MaxX`, then it's fine. Generally that is only true within a method call (and doesn't hold in general if exceptions are possible).

ravenshining:
Does this apply to freshly captured ships as well?  I wouldn't want to make that easier.

Do we want this constant?  Might it be better for it to be a random addition, or based on some property of the rescuer or rescuee?

jerith:
> Does this apply to freshly captured ships as well? I wouldn't want to make that easier.

I haven't read the relevant code, but from my normal gameplay I assume a captured ship is repaired immediately after capture.

In many cases, leaving captured ships disabled would make capturing easier, because they they wouldn't immediately fly off into a barrage of enemy fire and get themselves killed. While it *could* make sense for captured ships to get the existing "one point above disabled" repair to keep them as fragile as possible, I think the "capturing ships is too easy" problem is better solved in other ways (having the capture operation take some time, dropping shields on both capturer and captured, etc.) that are out of scope for this work.

> Do we want this constant? Might it be better for it to be a random addition, or based on some property of the rescuer or rescuee?

I like the random idea (maybe 1.3 to 1.6 instead of a flat 1.5?), but the difference is unlikely to be noticeable in most cases unless the high end is overpowered. Some kind of "mobile repair" outfit that increases the repair amount and maybe boosts shields or whatever might be interesting, but that would require more thought in terms of balancing and such.

ravenshining:
> Some kind of "mobile repair" outfit that increases the repair amount and maybe boosts shields or whatever might be interesting, but that would require more thought in terms of balancing and such.

Hull mass could be one way - something like one of these:
- MinimumHull() * ( 1 + ((rescuer hull mass) - 100) / 2000)
- MinimumHull() * ( 1 + (0.1 * (rescuer hull mass) - 10) / (rescuee hull mass))
So a player in one of the starter ships would get no bonus,  whereas if you've got a heavy warship you're more likely to be able to perform a substation repair.

tehhowch (approved):
One possibly negative effect of this change stems from the increase to post-repair `Health()`. A player's escorts that were just repaired may return to battle faster than they would have before due to the higher contribution of hull than shields, which means they may actually end up disabled again more quickly than before depending on their outfits and the player's behavior due to the comparative lack of shielding.

That said, there's a ton of variables at play in that particular scenario, and I expect the lion's share of situations to convey a net benefit to the player.

Issue #5425, opened 2020-10-09
Diversified H2H Equipment Behaviour in Boarding and Capture


MasterOfGrey:
### Is your feature request related to a problem? Please describe.
Boarding is a simple activity, and while no one is keen to turn it into some damn mini-game, the current nature is such that there are H2H options and tactics that are simply *always* superior.
The intent of this is to propose a relatively low-effort change which would address the lack of diversity and still ensure that the *work* involved in capturing is still done in preparation and planning, rather than making the actual act of capturing a more drawn out and exhausting process.

### Related Issue Links
 There is some derived content from here: [Issue 2392](https://github.com/endless-sky/endless-sky/issues/2392)

### Describe the solution you'd like
Essentially this becomes a system of attributes that apply modifiers to the base strength of different H2H weapons.
These modifiers either exists as part of the ship template, or as things added by particular outfits.

**Your first set of modifiers are Ship terrain:**
Corridor Width:
- Narrow
- Standard
- Wide
- Open
(These range from one individual passing at a time in a budget vessel like the Scout, to attacking something like a worldship that is likely to have big open spaces both for storage, mass transit and recreation purposes.)

Design Layout:
- Twisted
- Straight

Ventilation Capacity:
- Poor
- Adequate
- Good

**Your second set of modifiers are Ship/Crew Defences:**
Defences:
- Ventable Sections
- Combat Environmental Suits
- Defensive Alcoves
- Security Embrasures

**Impact on Weapons:**
These modifiers then affect the base strength of weapons. 
All modifiers are conceived of as negative for the principle case. Meaning, a H2H outfit's best effectiveness is it's base stats, circumstances that counter it provide penalties.
It is coneivable that a weapon could receive a positive modifier for something, but no good examples of this spring to mind.
However, the effect of the modifier should be multiplicative/additive in a way described below.

Examples:
- Nerve Gas will be fully effective against narrow corridors, with a twisted layout, and poor ventilation
- - It will be less effective under the inverse, and additionally reduced by the ship having the defensive qualities: 
- - - Ventable Sections
- - - Combat Environmental Suits
.
- Laser Rifles will be fully effective against standard corridors with straight layout.
- - They will have no modifier from ventilation capacity.
- - They will be less effective the further from standard they get and with twisted layout, and additionally reduced by the ship having the defensive qualities: 
- - - Defensive Alcoves
- - - Security Embrasures

[...]

PR #5796, opened 2021-03-16
refactor(content,weapons): Rebalance human lasers and blasters 

Pointedstick:
This is a standalone weapon balance change inspired by https://github.com/endless-sky/endless-sky/pull/5772 and https://github.com/endless-sky/endless-sky/pull/5779, intended to be merged ahead of them.

Ever since I started playing Endless Sky, Laser and Blaster weapons have felt wrong to me, causing me to immediately gravitate to other weapons, even creating the Proton Turret to satisfy my desire for a human turret with no Lasers or Blasters on it. The reason for this was the muddiness of these weapons: both generally did the same things with had similar functional characteristics; it felt like where was rarely a good reason to choose one over the other.

Over the years, the situation has been somewhat improved to differentiate these weapons' ranges and damage profiles. However now Lasers are in an extremely weird spot where they do a ton of hull damage but consume little energy, which is 100% backwards from how lasers behave according to the laws of physics. It makes them feel like nonsense cartoon weapons.

---

This PR differentiates Lasers and Blasters in a way that makes sense:

- Lasers now consume 50% more energy and do 30% more damage to shields, but 1/3 as much damage to armor - almost none at all!
- Blasters now have a balanced damage profile, but at a 17% lower level than their current shield damage, and with lower energy consumption too. To compensate, their inaccuracy is doubled.

The net effect of these changes is to strongly differentiate the weapons and give them distinct battle roles and styles of play. Laser weapons are now used for cracking open shields, but other weapons must be used against the target's exposed hull or else the target can simply jump away to safety! And their perfect accuracy now matters more, because Blasters are less accurate, particularly against very small fast ships due to the reduced accuracy. Blasters by are now true jack-of-all-trades weapons, but their low damage output makes them the master of none.

All told, gun combat is now slower, more tactical, and more forgiving:
- Because Blaster weapons now do less damage to shields, Human shield generators can absorb more of the damage with their regen.
- Because Laser weapons now take much much longer to disable a ship than they did before, ships under laser fire now have much more time to escape - particularly if they have relatively strong Hull values, as many freighter class ships do.
- Both gun types now benefit from the addition or partial substitution of missile weapons for different reasons: for Lasers, missiles can do much-needed hull damage, while for Blasters, missiles offer a much-needed net DPS boost to overcome the target's shield regen.

Because these changes are fairly self-contained, ships' default weapon loadouts need no adjusting, and extensive playtesting should not be required. I did a bunch with various ships with various loadouts in various stages of the Free Worlds campaign, and found things to be fine, with battles overall taking a bit longer and being more fun.

---

Human mining ships and fleets are adjusted because using Laser weapons against asteroids no longer makes any sense. Ultimately, some sort of dedicated mining weapon may be desirable to create.

### DIFF
--- a/data/human/fleets.txt
+++ b/data/human/fleets.txt
@@ -883,18 +883,12 @@ fleet "Human Miners"
 		confusion 20
 		timid frugal mining harvests
 	variant 5
-		"Sparrow"
-	variant 1
+		"Sparrow (Miner)"
+	variant 5
 		"Fury"
-	variant 2
-		"Fury (Miner)"
 	variant 1
-		"Hawk"
-	variant 2
 		"Hawk (Miner)"
 	variant 1
-		"Headhunter"
-	variant 2
 		"Headhunter (Miner)"
 
 fleet "Small Free Worlds"
--- a/data/human/variants.txt
+++ b/data/human/variants.txt
@@ -1096,19 +1096,6 @@ ship "Fury" "Fury (Missile)"
 		"Hyperdrive"
 
 
-ship "Fury" "Fury (Miner)"
-	outfits
-		"Heavy Laser" 2
-		"nGVF-CC Fuel Cell"
-		"Supercapacitor"
-		"D14-RN Shield Generator"
-		"Asteroid Scanner"
-		"X2700 Ion Thruster"
-		"X2200 Ion Steering"
-		"Hyperdrive"
-
-
-
 ship "Gunboat" "Gunboat (Mark II)"
 	outfits
 		"Electron Beam" 2
@@ -1242,7 +1229,7 @@ ship "Hawk" "Hawk (Speedy)"
 
 ship "Hawk" "Hawk (Miner)"
 	outfits
-		"Heavy Laser" 2
+		"Modified Blaster" 2
 		"S3 Thermionic"
 		"LP036a Battery Pack"
 		"D41-HY Shield Generator"
@@ -1977,6 +1964,19 @@ ship "Sparrow" "Sparrow (Gatling)"
 		"Chipmunk Plasma Thruster"
 		"Chipmunk Plasma Steering"
 		"Hyperdrive"
+
+
+
+ship "Sparrow" "Sparrow (Miner)"
+	outfits
+		"Energy Blaster" 2
+		"nGVF-AA Fuel Cell"
+		"LP036a Battery Pack"
+		"D14-RN Shield Generator"
+		"Chipmunk Plasma Thruster"
+		"Chipmunk Plasma Steering"
+		"Hyperdrive"
+
 
 
 ship "Sparrow" "Sparrow (Patrol)"
--- a/data/human/weapons.txt
+++ b/data/human/weapons.txt
@@ -20,15 +20,15 @@ outfit "Energy Blaster"
 		sprite "projectile/blaster"
 		sound "blaster"
 		"hit effect" "blaster impact"
-		"inaccuracy" 3
+		"inaccuracy" 6
 		"velocity" 10.625
 		"lifetime" 48
 		"reload" 12
-		"firing energy" 9.6
+		"firing energy" 7
 		"firing heat" 30
-		"shield damage" 9.6
-		"hull damage" 6
-	description "Although not the most accurate or damaging of weapons, the Energy Blaster is popular because it is small enough to be installed on even the tiniest of ships. One blaster is not enough to do appreciable damage to anything larger than a fighter, but a ship equipped with several of them becomes a force to be reckoned with."
+		"shield damage" 8
+		"hull damage" 8
+	description "Although not the most accurate or damaging of weapons, the Energy Blaster remains popular because it is equally effective against shields and armor, and its size and power needs are so small that it can be installed on even the tiniest of ships. One blaster is not enough to do appreciable damage to anything larger than a fighter, but a ship equipped with many of them becomes a force to be reckoned with - though it may require supplemental cooling systems to avoid overheating in the middle of a fight."
 
 outfit "Blaster Turret"
 	category "Turrets"
@@ -45,15 +45,15 @@ outfit "Blaster Turret"
 		"hardpoint offset" 9.
 		sound "blaster"
 		"hit effect" "blaster impact"
-		"inaccuracy" 3
+		"inaccuracy" 6
 		"turret turn" 3.7
 		"velocity" 10.625
 		"lifetime" 48
 		"reload" 6
-		"firing energy" 9.6
+		"firing energy" 7
 		"firing heat" 30
-		"shield damage" 9.6
-		"hull damage" 6
+		"shield damage" 8
+		"hull damage" 8
 	description "A Blaster Turret is a pair of Energy Blasters mounted on a rotating platform, allowing it to fire in any direction. Sophisticated software algorithms allow the turret to correct for a target's motion, making the Blaster Turret effective even against small, quickly moving fighters. Because it needs a special mounting point with a 360 degree field of view, not all ships are capable of being equipped with a turret."
 
 outfit "Quad Blaster Turret"
@@ -71,15 +71,15 @@ outfit "Quad Blaster Turret"
 		"hardpoint offset" 9.
 		sound "blaster"
 		"hit effect" "blaster impact"
-		"inaccuracy" 3
+		"inaccuracy" 6
 		"turret turn" 3.2
 		"velocity" 10.625
 		"lifetime" 48
 		"reload" 3
-		"firing energy" 9.6
+		"firing energy" 7
 		"firing heat" 30
-		"shield damage" 9.6
-		"hull damage" 6
+		"shield damage" 8
+		"hull damage" 8
 	description "A Quad Blaster Turret carries four guns on a single turret mount, doubling the firepower of an ordinary Blaster Turret for ships with enough space to install one and energy to drive it."
 
 effect "blaster impact"
@@ -103,14 +103,14 @@ outfit "Modified Blaster"
 		sprite "projectile/mod blaster"
 		sound "mod blaster"
 		"hit effect" "blaster impact"
-		"inaccuracy" 4
+		"inaccuracy" 8
 		"velocity" 10
 		"lifetime" 48
 		"reload" 12
-		"firing energy" 10.8
+		"firing energy" 10
 		"firing heat" 42
-		"shield damage" 12
-		"hull damage" 8
+		"shield damage" 10
+		"hull damage" 10
 	description "This is an Energy Blaster that has been tampered with to increase its power slightly, at the cost of reduced range and higher energy consumption. Modified Blasters also produce an extreme amount of heat when firing, increasing the risk that your ship will overheat."
 
 outfit "Modified Blaster Turret"
@@ -128,15 +128,15 @@ outfit "Modified Blaster Turret"
 		"hardpoint offset" 9.
 		sound "mod blaster"
 		"hit effect" "blaster impact"
-		"inaccuracy" 4
+		"inaccuracy" 8
 		"turret turn" 3.5
 		"velocity" 10
 		"lifetime" 48
 		"reload" 6
-		"firing energy" 10.8
+		"firing energy" 10
 		"firing heat" 42
-		"shield damage" 12
-		"hull damage" 8
+		"shield damage" 10
+		"hull damage" 10
 	description "This is a turreted version of the Modified Blaster, which is an Energy Blaster modified for greater power at the cost of higher energy consumption and heat. These turrets are popular with pirates and others who are trying to cram as much firepower into their ships as possible."
 
 
@@ -157,11 +157,12 @@ outfit "Beam Laser"
 		"velocity" 300
 		"lifetime" 1
 		"reload" 1
-		"firing energy" .5
+		"firing energy" .75
 		"firing heat" 1.2
-		"shield damage" 1
-		"hull damage" 1.3
-	description "In the early part of the space era, the settlements in the region of space known as the Deep developed in relative isolation from the rest of human space. One result of that isolation is that their weapons technology mostly uses beam weapons, instead of the energy projectile weapons that are more common elsewhere."
+		"shield damage" 1.3
+		"hull damage" .4
+	description "In the early part of the space era, the settlements in the region of space known as the Deep developed in relative isolation from the rest of human space. One result of that isolation is that their weapons technology mostly uses beam weapons, instead of the energy projectile weapons that are more common elsewhere. Like most examples of Deep technology, these weapons are extremely efficient, though power-hungry."
+	description "	Laser weapons do a great deal of damage to shields, but they are practically useless against armor."
 
 outfit "Laser Turret"
 	category "Turrets"
@@ -184,11 +185,12 @@ outfit "Laser Turret"
 		"velocity" 300
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1
+		"firing energy" 1.5
 		"firing heat" 2.4
-		"shield damage" 2
-		"hull damage" 2.6
+		"shield damage" 2.6
+		"hull damage" .8
 	description "Laser Turrets are hated by fighter pilots because it is nearly impossible to dodge them once you are within their reach. This turret carries a pair of lasers and can swivel almost instantly to fire on new targets as they approach. Laser Turrets are especially useful when mounted on slow-moving freighters to fend off packs of small pirate vessels."
+	description "	Laser weapons do a great deal of damage to shields, but they are practically useless against armor."
 
 effect "beam laser impact"
 	sprite "effect/laser impact"
@@ -217,11 +219,12 @@ outfit "Heavy Laser"
 		"velocity" 400
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1
+		"firing energy" 1.5
 		"firing heat" 2.1
-		"shield damage" 1.7
-		"hull damage" 2.4
+		"shield damage" 2.2
+		"hull damage" .7
 	description "The Heavy Laser is an upgraded Beam Laser with a significantly longer range and higher power. It is mostly intended for larger ships, where energy and space are plentiful, but some pilots consider a single Heavy Laser to be a worthwhile alternative to a pair of Beam Lasers, because the longer range makes up for the fact that it does not quite deal twice as much damage."
+	description "	Laser weapons do a great deal of damage to shields, but they are practically useless against armor."
 
 outfit "Heavy Laser Turret"
 	category "Turrets"
@@ -244,11 +247,12 @@ outfit "Heavy Laser Turret"
 		"velocity" 400
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 2
+		"firing energy" 3.0
 		"firing heat" 4.2
-		"shield damage" 3.4
-		"hull damage" 4.8
+		"shield damage" 4.4
+		"hull damage" 1.4
 	description "For ships with enough space to install one, the Heavy Laser Turret is a powerful weapon, equally useful for driving off fighters and for bombarding larger targets with continuous fire without having to worry about pointing your ship in the right direction."
+	description "	Laser weapons do a great deal of damage to shields, but they are practically useless against armor."
 
 effect "heavy laser impact"
 	sprite "effect/heavy laser impact"
@@ -278,10 +282,10 @@ outfit "Electron Beam"
 		"velocity" 450
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 1.5
+		"firing energy" 2.25
 		"firing heat" 2.7
-		"shield damage" 2.4
-		"hull damage" 3.3
+		"shield damage" 3.1
+		"hull damage" 1.0
 	description "The Electron Beam is a recent development by the Deep Sky labs, a more powerful weapon with a design similar to their perennially popular laser guns."
 
 outfit "Electron Turret"
@@ -305,10 +309,10 @@ outfit "Electron Turret"
 		"velocity" 450
 		"lifetime" 1
 		"reload" 1
-		"firing energy" 3.0
+		"firing energy" 4.5
 		"firing heat" 5.4
-		"shield damage" 4.8
-		"hull damage" 6.6
+		"shield damage" 6.2
+		"hull damage" 2.0
 	description "This turret places two of Deep Sky's recently developed electron beam weapons onto a rotating turret, providing unsurpassed accuracy and power for shooting down fast-moving targets."
 
 effect "electron impact"


Amazinite:
> intended to be merged ahead of them.

:-1:
I mention in the description of #5779 how it's probably worth having one big balance PR given how large in scope (and interconnected) some of our balance issues are. What more balanced blaster and laser weaponry looks like right now could be different than what more balanced blaster and laser weaponry looks like alongside #5779, #5772, or any other balance PRs. So even if we have multiple PRs opened looking to address various balance points (e.g. #5793 has been opened as well), I don't think we should be looking at any of them in a vacuum or saying that we should merge certain changes ahead of others. I think our approach should be "Are all of these ready and look good when combined together? Okay, let's merge them all together."

As far as the changes themselves go just taking a quick glance at it, not a fan of completely flipping lasers on their head to now deal more shield damage than hull damage. Among the guns in human space, lasers are the only hull damage heavy weapons, with the plasma cannon and blasters being more shield damage heavy while the particle cannon and proton gun are more balanced. With this change human space would have no weapons that deal higher hull damage than shield damage outside of the gatling gun.

Where this especially becomes an issue is in changing how mining works, which you've explicitly had to address because of these changes to the damage profile of lasers. I feel that this change VERY negatively effects the mining experience, as it makes mining with an asteroid scanner almost mandatory. An asteroid scanner is necessary in order to target an asteroid, but when you have lasers than instantly hit a point in space then you don't necessarily need to target the asteroid given that you can just aim right at it to hit it. But if blasters become the mining weapon of choice then you need to manually lead the asteroid you're trying to mine if you don't have a scanner to target it, which can be rather frustrating, especially when mining asteroids that are going at high speeds. And I don't think that the introduction of a dedicated "mining laser" is the solution to the problem that this creates. That not only increases the barrier to entry for mining by having to swap out your Sparrow's beam lasers for mining lasers instead of simply being able to use the default fitting, but it also goes against the original design philosophy of minables that you shouldn't need dedicated hardware in order to mine asteroids.

Pointedstick:
This is a pretty self-contained change IMO. I don't think it requires redoing the rest of the world to make sense and improve the combat experience. In general, batching together large changes just makes the whole thing more likely to drown in disagreements and never get implemented. Smaller atomic changes are preferred for this reason.

---

I admit I didn't think much about the mining use case because I find it to be un-fun and tedious and never do it. It wasn't in the base game and was a somewhat recent addition. But you raise a valid point. However from a physics perspective, using lasers to do large amounts of damage to physical matter makes no sense. We don't mine with lasers today; we use lasers for tasks where precision is important. This isn't to say that future lasers wouldn't be more powerful, but the physics would still be against it. Mining demands brute force, not precision.

However maybe we could say that space mining is a precision task and make lasers do 10x damage to only asteroids or something, on the basis that it's specialized for this task (if you don't like the idea of a dedicated mining weapon). That's what the EVN mining laser did.

tehhowch:
I also don't agree that this is a "fast" or particularly self-contained change. I also feel that changing the energy consumption of staple weaponry means we _do_ need to evaluate the energy budget of stock ships. 

This change will also push players away from all-laser fleets, which players will find frustrating for the capture meta. Those that stick with all-laser fleets will find combat harder, especially against common NPC fleets that employ ammunition-based combat.

Pointedstick:
That's the point: you should have to put some thought into your loadout rather than just spamming a single obviously-best weapon. And I thought it was generally accepted that the "capture meta" was overall a bad thing that still needed to be toned down.

tehhowch:
I'm not in favor of the capture-is-king meta, but balancing content isn't how we resolve it.

Forcing players to use a mixture of weapons is bad. Letting them organically choose to do so is good. Forcing mixtures with wildly different time-of-flight is very frustrating since your ship cannot aim both at the same target unless you use turrets for one type and fixed emplacements for the other (or just turrets for everything). NPCs will only fire what may hit, but the player does not have this granularity and will fire everything at once, regardless.

Forcing use of ammo-based weaponry exacerbates the frustration of resupplying ammo, especially compared to the opponents the player will have, which are effectively not resource-constrained.

gunqqer:
Mixing weapons on 1 ship is inherently difficult to use (unless 1 is a secondary weapon that has tracking). Trying to contend with different ranges, different velocities, different turning rates, all are not very fun to do so the best solution is to sit on top of the enemy (or point blank) which defeats the purpose of all thtese combat changes.

On the point of missile: I like them. Until I run out. If I am using cheap human missiles it isn't as bad, but if I use Typhoons I have to go to the Deep to resupply and if I use any alien missile I may have to treck across the galaxy. I've been using a shield beetle with 8 thunderheads to do Remnant missions by choice and resupplying is really annoying to do. Not just because of the treck, but because of the cost. I have to have a lot of ammo racks to take enough ammo to last me some time so when I go to rebuy ammo the popup saying "Do you want to resupply 700 missiles for 1 million credits" I sorta regret using missiles instead of turrets. If I was forced to use missiles.

Being able to pick a single weapon is always going to be easier to work with and less frustrating that working with possibly clashing types. And don't try to fix the capture meta by nerfing beam weapons. That won't actually solve the problem. It only makes it more frustrating, it does not stop it.

Zitchas:
I definitely agree that human space needs a bit more thought behind the weapons, and a change like this that encourages mixing is generally speaking a good thing. Lasers have long enjoyed a "lasers are best" reputation. It often seems like the only weapons that really get talked about are the laser (and heavy laser if you can fit it) for general combat, and plasma canons if you are specifically trying to overheat things. I know others do come up now and then, but these two do dominate.

Lasers already enjoy an accuracy and "instant-effect" advantage over everything else - I don't think they need to be Queens of the Battlefield in all regards on top of that. 

As to whether lasers should be better against shields or better against armor... That's hard to say. Lasers generally aren't the most effective things against large solid objects; but it is really difficult to compare that known against the theoretical properties of shields. Would they be more effective than the discrete chunk of energy that is a blaster against a shield? Less so? I don't know. 

I generally view the blaster's role as being "the cheapest of the cheap guns used to fill up empty gun ports when nothing else will fit." with a lore use case of being weapon-of-choice for discount security outfits and drone/fighter squadrons with an expected high-casualty rate. No sense wasting expensive guns on things that will die in the first 30secs of combat, after all. The one exception is when faced with a situation where neither dps nor accuracy are needed, and you simply need to fill space with shots, such as anti-fighter or anti-missile fire. Similarly, they'd probably make great "spot the cloaked ship with weapons fire" tools.

arkhne:
> I'm not in favor of the capture-is-king meta, but balancing content isn't how we resolve it.
> 
> Forcing players to use a mixture of weapons is bad. Letting them organically choose to do so is good. Forcing mixtures with wildly different time-of-flight is very frustrating since your ship cannot aim both at the same target unless you use turrets for one type and fixed emplacements for the other (or just turrets for everything). NPCs will only fire what may hit, but the player does not have this granularity and will fire everything at once, regardless.
> 
> Forcing use of ammo-based weaponry exacerbates the frustration of resupplying ammo, especially compared to the opponents the player will have, which are effectively not resource-constrained.

This is a lot of my big concerns with rebalancing in one post.
Whether or not you like the current meta, changing it is going to aggrevate a lot of people.
Having (viable) options is great, and should be the goal of any balancing effort, but creating deficiencies just forces the issue, which definitely negatively impacts the player experience.

That all said, at some point, the trigger will need to be pulled, and it wont please everybody, the best we can hope for is that the changes at least make logical sense, and allow players to adapt rather than forcing a sudden and dramatic change. Such a drastic change as this PR is more of the latter kind.

PointedStick, for you I am sure that this is such a small change that it really doesn't impact you, as you yourself said - you don't use blasters and lasers - other people do. That is an important distinction to make when considering the wide-reaching affects of changes. What is normal for you, may not be normal for others.

Tehhowch also made a good point about how changing energy or heat requirements of weapons means that you need to look at stock loadouts. As it is, stock loadouts (particually in the variants department) are often pushing their limits as it is, making changes that affect them can totally invalidate their entire setup.

daveeater:
Correct me if I misunderstand, but would nerfing the Deep's beam weapons not also require changing the lore concerning the Deep, the influence of the Pug on the Deep, and the role of the Deep/Pug in winning the Alpha wars? If lasers are nerfed to the point that they're "meh" as opposed to blasters...how do they turn the tide in the Alpha wars?

What's the point of the Pug populating humans in the Deep in... the 13th century?... if the contribution of that intervention is an ineffective weapon that doesn't really change anything?

To  me, this proposed change feels like an attempt to "improve" early-game battles, at the expense of the story.  It risks alienating players whose enjoyment of the game is not heavily due to battles, as oppose to exploration/discovery.

Or am I over-reacting here?

Pointedstick:
This proposal doesn't nerf lasers at all; it makes them 30% stronger against shields, which are the most important thing to be strong against in human space. What it does is remove them from the realm of being a "no-brainer" weapon, and makes blasters competitive.

---

We're all aware of the issues with Endless Sky's combat system, but I feel like we're too timid to fix it for fear of upsetting the balance. For years we have dug around the edges here, tweaked things there, but in the end we've failed to actually fix it. Sometimes when you have a big problem, you have to go big to fix it, or else you'll be tweaking this and that forever without ever arriving at a real solution. And this isn't that big of a change at all.

I've said my piece. Do with this what you will.

arkhne:
> This proposal doesn't nerf lasers at all; it makes them 30% stronger against shields, which are the most important thing to be strong against in human space. What it does is remove them from the realm of being a "no-brainer" weapon, and makes blasters competitive.

This is 100% absolute BS. This is 100% a nerf. Not only have you increased the energy cost by 50% (that right there is a nerf already), you've given a 30% increase in its lesser damage stat at the cost of 70% of its strongest stat, for a net loss 26% DPS (nerf again). 26% less DPS at an increase of 50% cost is an almost 40% nerf.

I don't think you know what the word 'nerf' means.

MasterOfGrey:
> It often seems like the only weapons that really get talked about are the laser (and heavy laser if you can fit it) for general combat

This is at least 80% on account of it being a matter of convenience to casually capture ships for bonus profit. If we overhaul boarding to actually use some logic and involve requiring you to pick a loadout for the type of capturing you're trying to do ala #5425 (which I am learning to code in order to have a crack at since it won't be on anybody else's list any time soon), then I think a lot of the conversation about lasers would chill. For straight up being able to wreck stuff there are other weapons that are just as good, or at least are competitive and take up less space which you can then use to tank better.

gunqqer:
The reason why I like beam weapons so much is because of the fact they are instant. There's no lead, there's no stopping a second before they're disabled as not to destroy. As MasterOfGrey said, as soon as you change how the capture meta works, then lasers are much easier to debate. For straight up destruction missiles are one of the better bets (espeically with the current proposed rebalances) but then you can't get money from taking outfits or capturing.

Amazinite:
> The reason why I like beam weapons so much is because of the fact they are instant.

This has me thinking. Right now the speeds of various projectiles in human space are as follows

Weapon | Projectile 
---|---
Electron Beam | 450 (Instant impact)
Heavy Laser | 400 (Instant impact)
Beam Laser | 300 (Instant impact)
Particle Cannon | 30
Proton Gun | 24
Gatling Gun | 16
Sidewinder Missile | 14
Plasma Cannon | 12
Meteor Missile | 11 
Javelin | 10
Modified Blaster | 8.5
Energy Blaster | 8
Heavy Rocket | 8
Torpedo | 6
Typhoon | 5

I think we need to reassess the velocity balance of weapons and think about why it is that each weapon travels the speed it does.

Well lasers are fast because they're photons traveling at the speed of light.
The particle cannon is fast because it's described as throwing particles at near relativistic speeds. and the proton gun is similar to that.
The torpedo, Typhoon, and heavy rockets are slow because they're massive projectiles.

But why is it that the gatling gun fires bullets nearly as faster than anything but lasers and the particle and proton guns?
And why are the blaster projectiles so slow? Given the name "energy blaster," I would think that the projectile itself is akin to a laser in that it is a collection of photons, so should they actually be traveling at high speeds? (Although for gameplay/visual purposes, we wouldn't want them to have an instant impact.)

So I think we should rebalance the speeds of projectiles as follows: Lasers at the top with instant impacts, energy blasters as the second fastest projectiles in human space, the three cannons (particle, plasma, proton) below that, then all secondary weapons filling some space below the cannons. I would expect the Javelin, Meteor Missile, and Sidewinder to all drop in projectile speed, which would help in terms of making missiles easier to dodge or even simply outrun.
And speaking of outrunning missiles, part of #5779 is increasing the speed of smaller ships, and since smaller ships lack turrets, their ability to dodge or outrun missiles is paramount to survival against them when getting hit by one deals so much damage.

arkhne:
Because I feel this is an area that is only going to get worse in the future, I will bring up why *I* tend to use Beam weapons if talking about human space only. Inaccuracy. I absolutely hate it, almost as much as I hate the entire missile system in the game. Anything with an Inaccuracy >= 4 isn't go on my ship, and I have a *distinct* preference for <1.

What's worse is that I feel like the answer to every combat rebalance discussion ends up the same, the only thing anybody seems to get some sort of consensus on is increasing inaccuracy, and it most definitely makes me feel pretty disheartened regarding the future of combat in ES.

When you remove the inaccurate weapons, and the missiles, pretty much the only options left tend to be Beam weapons. I don't even care about the capping aspect of using beam weapons, it is 100% an inaccuracy thing.

Protons and Blasters most definitely need to come down in the inaccuracy thing, with them sitting right up against my tolerance level, I'm almost always going to choose something else. If Lasers are nerfed, then I'm just going to use Plasma Cannons. It might be a "no-brainer" to take lasers in human territory, but for me, that's because the options are slim.

I don't know of anybody who uses Blasters. I can name a few people who uses Missiles, Plasma Cannons, and even Particle Cannons and Proton Cannons (even beyond human space). I don't even know of anybody using *Modified Blasters*, so it isn't just the case of people not bothing with the weakest option.

If Blasters and Modified Blasters had their inaccuracy halved, I would genuinely use them in the early game. And even consider the Quad Blaster Turret in the mid-game (although it has to compete with the Hai Pulse Turret at that point).

This, combined with Derpy's suggestions for projectile speeds would definitely help the feel of human combat for me. The ability to outrun missiles (particularly in smaller ships) would be excellent, right now, it feels like most ships genuinely aren't capable of it, even with Atomics. Combined even further with mass increases to further differentiate between ship classes, and we're looking at a much more interesting state of affairs for me.

gunqqer:
So to sum up the last few comments: beams are consistant, so if you want people to use blasters, make them more consistant.

If blasters hits more consistantly that'd also fix the problem of wanting to sit ontop/point blank in relation to the enemy (The AI would need to be updated in general to make it not do this, however)

arkhne:
> So to sum up the last few comments: beams are consistant, so if you want people to use blasters, make them more consistant.
> 
> If blasters hits more consistantly that'd also fix the problem of wanting to sit ontop/point blank in relation to the enemy (The AI would need to be updated in general to make it not do this, however)

Vitalchip was working on getting the AI to be smarter. They've made some excellent progress imho, but a lot of it isn't really compatible with vanilla right now (Such as Fighters needing over 1k accel and Heavies needing to turn <50). I *think* their strafing thrust and rangefinding would work with the current setup in vanilla though.

MasterOfGrey:
> I would think that the projectile itself is akin to a laser in that it is a collection of photons

I always thought of Blasters as a sort of light plasma - can pick up a fair bit of oomph but lacks the mass to impart significant heat (and would rapidly disperse on impact).
I compared this to the more advanced Plasma weapon, which would ideally use a heavier elements for its make-up that would be able to carry more heat, and not lose all of their heat and instantly disperse upon impact after recombining.

In any event the Particle and Proton cannons should be *much* faster than the plasma cannon, and I would argue that the gatling gun should have a slower fire rate but faster travel speed than either Blaster or Plasma.

Zitchas:
I generally adhered to the belief that the main purpose of the blaster is to be something that you upgrade from. Basically, the first rite-of-passage for new pilots who have made enough money to sell their blaster and get something - anything - better. It is noteworthy only in terms of its price. That being said, as a "cheap stock weapon," it should be a reasonable all-rounder. Not super inaccurate, nor particularly weak against anything, but not  particularly *good* against anything either. Anything it can do, something else can do better. It is the "average" weapon, the baseline. Other things may be worse than it in one or two respects, but only alongside areas where they are significantly better.

From there, logic and lore suggest a whole range of places that such a weapon would be used:
- Low-end security firms that just need their forces to have *something,* but who lack the skill to make good use of better weapons. (and likely lack the budget for them too.)
- Weak pirates, low-end criminals, and the desperate-turned-criminal would all tend to have the cheapest weapons available.
- Merchants & others who just want to be able to say that they have a gun and either don't know or don't care about weaponry. 
- Minions of all kinds: Drones, fighters, interceptors, and ships piloted by people considered disposal and likely to die in the course of duty by the people who actually own the hardware. No-one wants to stick a plasma canon on a fighter just to watch it get obliterated in the first 20seconds of combat, after all. 

In EVN this was the case, but not to be confused with the light blaster - the weapon that was worse in almost every way except rate of fire - and thus earned it a special place as weapon of choice for anti-fighter stuff. (also made it quite popular for interceptors and fighters with lots of hardpoints who wanted to be able to mount *lots* of weapons but lacked the space to do so with bigger weapons.)

arkhne:
> I generally adhered to the belief that the main purpose of the blaster is to be something that you upgrade from. Basically, the first rite-of-passage for new pilots who have made enough money to sell their blaster and get something - anything - better. It is noteworthy only in terms of its price. That being said, as a "cheap stock weapon," it should be a reasonable all-rounder. Not super inaccurate, nor particularly weak against anything, but not particularly _good_ against anything either. Anything it can do, something else can do better. It is the "average" weapon, the baseline. Other things may be worse than it in one or two respects, but only alongside areas where they are significantly better.

In the Blaster's case, it's significantly worse in all regards. It's on the high end of inaccuracy (not only for human space, but *all* space). It's on the slow end of projectiles, makes it even less accurate in practice. "Average" would be a significant buff to the Blaster. It's a little inefficient in the heat department, but that's quite a small price to pay, and makes sense (cheap construction is cheap).
With a buff to Projectile Speed and Inaccuracy, the Blaster could do exactly that, and actually allow said rookie pilots to gain experience in piloting skill, rather than gain experience in luck :P

> In EVN this was the case, but not to be confused with the light blaster - the weapon that was worse in almost every way except rate of fire - and thus earned it a special place as weapon of choice for anti-fighter stuff. (also made it quite popular for interceptors and fighters with lots of hardpoints who wanted to be able to mount _lots_ of weapons but lacked the space to do so with bigger weapons.)

Sounds to me like a Blaster/Modified Blaster Scenario (only reversed). I could see the Modified Blasters being buffed to the "Average" weapon, and the Blaster being Buffed to be "Below Average".

Looking at Derpy's table, I'd say that Velocity wise, I'd put them between 12 and 20 Velocity, as to precisely where, that's for somebody smarter. I'd also suggest an inaccuracy somewhere between 1.8 and 2.8.

Pointedstick:
Maybe this is a personality thing, but I find it much more interesting wen you have a set of choices where each choice has a prominent disadvantage that you have to figure out a way to compensate for, than it is to have a set of choices where everything is good, and the differences are in what each choice is outstanding in. I find this to be boring.

arkhne:
> Maybe equip some missiles too.

Have you noticed how every single suggestion, post, and argument you make includes this? Have you considered, perhaps, that there are many players who do *not* want to use missiles to compensate for *everything*? Perhaps because your solution to fixing every problem with your ship being "add missiles" means that the problem is with the missiles?

If you want to nerf beam weapons, sure, go ahead, you have my support, I'll even suggest that you start by nerfing shield and hull damage both by 20%, and increase energy and heat use by 20-50%. This will bring them in line with other non-secondary weapons. However, doing so *will* need you to change *every* single stock and variant ship that uses lasers, and this is why *I* wont be making that PR.

Swapping Shield vs Hull specialization of a weapon requires looking at the weapon ecosystem of the race in question, and in this case, as has been said above, you leave humans distinctly lacking in hull-specializing weapons. This is not a good thing, even if, as you say, human space has a shield emphasis. This is most definitely *not* a good thing. And creates exactly what you should *not* be doing - removing options.

I'm all for balancing the options out, and making all options viable, I just think that *this* approach is flawed and creates more problems than it solves. It's great to talk about it, and discuss the consequences or benefits of it, but I don't see this as being an improvement to the game.

gunqqer:
The problem is twofold. You first have the issue with lore: Blasters are cheap, weak, easily available weapons, so trying to make them not that means you also have to change the lore on the blasters and then implement a new cheap, weak, easily available weapon to replace it. A Star Barge isn't going to have a quality weapon on it.
Then you have the problem of balancing/meta. The reason why lasers are used so much is because of their consistancy. You won't miss, you stop firing instantly. So making lasers less viable and forcing people to use blasters for damage means people either get annoyed at the blaster for killing their target or missing, or they get annoyed at not doing enough damage. 
And on the topic of missiles: I like them, sorta. Asteroids can nullify them way too easily. They are expensive. Even if you are fine with their credit cost, they aren't sold everywhere (especially alien missiles). But the good thing about them is that they are effective (when they hit) against almost anything all around. So saying "Just add missiles" is like how sugary cereal adverts say "part of a complete breakfast." The breakfast was already complete by having the missiles, the sugary cereal of blasters or lasers doesn't add much.

Quantumshark:
Realism-wise, surely lasers would be the most energy efficient weapons for the damage dealt, compared to a projectile which is continually dissipating and/or losing energy to its surroundings as it travels. For the particle cannon, this would be less true, but realistically that's about the only human weapon that could reasonably be more energy-efficient than lasers. As for energy blasters, we would really need some explanation as to how they actually work before making any arguments about them based on realism.

Balance-wise, adjusting the energy use of any equipment should not be done without reviewing the energy of every ship and variant that uses that equipment. The stock Sparrow, for example, now runs out of energy when firing both weapons and engines in 15.2 seconds, compared to 27.8 seconds previously. Also of note, the Navy use a lot of lasers during the main story while the Free Worlds use them relatively little (particularly later on with the electron beam), so the effects of this change on the balance of that would need to be considered as well.
I dislike the emphasis on trying to persuade players to use missiles. If players are choosing not to use missiles, nerfing other weapons to try and force them into using missiles is not going to improve their experience. Fundamentally, missiles have an inherent balance problem because their main downsides only affect the player and not NPCs. While NPCs can run out of ammunition, this doesn't occur very often, nor do they have to worry about paying for it. Trying to force players into using missiles by making them a superior choice, either by buffing them or nerfing other weapons, just emphasises this.
Capturing-wise, I always saw the shield/hull damage ratio of lasers as the factor to balance out the ease of disabling enemies with them. As I see it, you have a choice between lasers, which you can stop firing the instant a ship is disabled but which damage hull more quickly than shields, or the various projectile weapons, which (generally speaking) damage shields more than hull, but you have to worry about the projectiles continuing for a few seconds after you stop firing. In that sense, this change would remove the one downside of lasers as a capturing weapon
I do agree that some changes to the energy blasters would be good, since right now they seem basically to just fill the role of being the smallest, cheapest and weakest weapons. They are inaccurate, deal minimal damage and are incredibly inefficient in both heat and energy, and even the smallest ships have enough space to just run beam lasers instead.

Vedvod:
The problem with realism and lasers is that 
a) lasers travel at the speed of light, and appear instantaneous, you realistically shouldn't see a laser shot, only it's effects. 
b) lasers would consume a lot of energy to deal appreciable damage and be used in a fight.
c) a laser's main damage is in applying a lot of heat to a very pointed area, so depending on how shields work, this may be either a shields bypass or a complete block.

Therefore, we can reasonably accept that Endless Sky does not use "realistic" laser weaponry.

Pointedstick:
Well realism is a sliding scale, right? There's no sound propagation in space either but nobody is running to remove all sound from the game. ES has never been a super hard sci-fi game, which means that some divergence from realism can be expected in the name of fun. But it's also not so soft and squishy a sci-fi game as in the Star Wars universe, in which basically anything can be handwaved away to do whatever the plot requires. ES is somewhere in the middle. So we want weapons to broadly and generally feel somewhat true to life when extrapolated to a far future sci-fi space battle context, but we don't need to tear out our hair making them 100% conform to a hard sci-fi adaptation or making them un-fun in the process. It's enough to get the broad strokes right.

So ships burn fuel, there's no inertia, missiles explode, and so on. Nobody knows what a laser would do to an energy shield because we humans haven't invented energy shields yet, so we can make up whatever we want. But to me, making lasers high-hull-damage weapons is going just a bit too far because of the basic way that lasers work. A focused cutting beam logically can't do as much instantaneous damage to a physical surface at a great distance as an explosive or a high velocity projectile can. We can make up whatever details we want, but to me it violates the laws of physics just a bit too much to say that laser weapons are more effective against a ship's hull than projectiles and explosives. It implies a completely different type of weapon compared to today's lasers, but this is never established in the game's canon.

Amazinite:
Closing in favor of #6675, which takes a more aggressive approach to buffing blasters and a less aggressive approach to nerfing lasers. Let's see how that works out, then in the future if we want to further differentiate the weapon types or tweak things more we can do so.

Issue #5797, opened 2021-03-16
Ship definitions require useless token to work


Terin:
**Describe the bug**
When creating a Person ship definition, or adding a ship to a player start, it must follow the syntax `ship "<ship type>" "<token>"`. In the case of the Persons file, that second token is used assuming it is a name. However, the value of that token is completely ignored, only looked to see if it exists. It does not name the person ship/starting ship after it, but it also breaks functionality if the token is absent.

**To Reproduce**
1. Create a start ship/person definition, such as `ship "Shuttle" "Three Skeleton Key"`.
2. Load the new start/wait for person ship to appear.
3. Notice the name being blank/using the person name rather than the given name.

ALTERNATIVELY: Enter a blank name, and the ship will not appear in either case.

As of the current commit, the Subsidurial person ship has been broken in part due to it missing this second token.

**Expected behavior**
For start ships:
- Entering a token there should give the starting ship that name.
- Having no token should give the ship a random name.

For person ships:
- Entering a token should override the person ship name.
- Having no token should give the person ship the person's name.

There may be other applications of this style of ship definition in the game, but these are the two that come to mind that I can test.

**System (please complete the following information):**
 - OS: Windows 10 Pro
 - Game Source: Self-compiled
 - Version: 6359f3a

**Link to plugin**
[ship-token.zip](https://github.com/endless-sky/endless-sky/files/6149229/ship-token.zip)
This is a plugin which comes with two identical starts that should give a ship. The "Broken" one gives no ship and leaves you stranded, while the "Working" one has a `ship "Shuttle" "Three Skeleton Key"` value, which gives you a shuttle with a blank name.

tehhowch:
Nitpick: If it's required for the content to be parsed, it isn't entirely "useless."

Its necessity is a side effect of how the game currently interprets ship definitions and how the game interprets person definitions. There's no "bug" here, just tech debt to be resolved that arose from disparate development cycles, evolving designs, and wholly manual testing.

As far as ships gifted by starts go, the game does warn about unsupported syntax and then avoids giving the player what would be a completely broken ship. I've already documented that supporting a stock fleet definition is an eventuality. (You'll note that the wiki for starts does not say you can use the syntax you are using.)

PR #5798, opened 2021-03-16
Fixed issues that prevented Subsidurial from appearing

Terin:
**Bugfix:** This PR is a followup to #5104.

## Fix Details
This PR addresses two issues that prevented the Subsidurial from spawning. First, it adds a "jump speed" attribute to it, which is required for a functional hyperdrive to jump in. Second, it adds the second (vestigial) token to the ship definition necessary to spawn the ship, an issue described in #5797.

Also includes minor formatting fixes to the persons file.

## Testing Done
Removed the person ship spawning limiter, which causes all person ships to spawn in at once, and saw the Subsidurial finally coming in with all the rest (while in an uninhabited system).
image


tehhowch:
It isn't a vestigial token if it has a purpose btw, even if that purpose is solely getting the game to parse the content.

PR #5799, opened 2021-03-16
Added hails to Subsidurial

Rob59er:
## Summary
The Subsidurial now has a variety of different "moos" when hailed.
## Testing Done
mooingcow

### DIFF
--- a/data/persons.txt
+++ b/data/persons.txt
@@ -847,7 +847,56 @@ person "Subsidurial"
 		"never disabled"
 	system
 		government "Uninhabited"
-
+person "Subsidurial"
+	government "Indigenous Lifeform"
+	frequency 1000000000
+	personality
+		timid unconstrained coward appeasing uninterested mining harvests
+	ship "Subsidurial" "Subsidurial"
+		"never disabled"
+	system
+		government "Uninhabited"
+	phrase
+		word
+			"M"
+			"m"
+		phrase
+			"ooo"
+		phrase
+			"ooo"
+		word
+			"."
+			"..."
+			"!"
+			"?"
+			"!?"
+	phrase
+		word
+			"M"
+		phrase
+			"OOO"
+		phrase
+			"OOO"
+		word
+			"."
+			"..."
+			"!"
+			"?"
+			"!?"
+phrase "ooo"
+	word
+		"o"
+		"oo"
+		"ooo"
+		"oooo"
+		"ooooo"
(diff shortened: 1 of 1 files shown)


Zitchas:
If the Subsidurial was actually given a variety of interesting and truly alien things, the sort of things that you'd expect an alien space creature to make, sure. The phrase generation system could even be used to make a large variety of different effects.

But just variations on "Moo?" Sorry, not a fan.

Rob59er:
I am open to suggestions.

MCOfficer:
If anyone has time to flesh out the subsidurial and write alien dialog for it, i'll take it. Until then, moos will do just fine :D

clifflocked:
> If anyone has time to flesh out the subsidurial and write alien dialog for it, i'll take it. Until then, moos will do just fine :D

Not sure how good this is, but I spent some time making variations on "space cow": (also, is the reference to Coalition Longcows spoilery? Probably not, but I want to know,)
```
phrase "spacecow"
	word
		"Cosmos"
		"Cosmic"
		"Spatial"
		"Void"
		"Universe"
		"Space"
		"SpAcE"
	word
		" "
	word
		"cow"
		"cOw"
		"CoW"
		"cOW"
		"herd"
		"oxen"
		"bull"
		"longhorn"
		"longcow"
		"mooer"
	word
		"."
		"!"
		"?"
		"!?"```

Amazinite:
I don't see why the subsidurial should have hails at all. Is it not another space animal akin to the void sprite? Why would it be able to communicate to the player?

Zitchas:
That's my thought. My previous comment was intended to suggest things like, for instance, it generating random comm static or bursts of radiation that comms might pick up and interpret as various sounds that the human brain might interpret as sound effects, but probably aren't.

It quite obviously does not have a comms system, and while I could see someone "putting it on the vidscreen" to use Star Trek parlance; but that would quite simply just be pulling up an image of it, not establishing two-way communication.

Rob59er:
Personally I was imagining it as just you viewing the cow when you hailed it, the ability to communicate with it just being a by product of the game since (as far as I am aware) one is not able to hide the other buttons. I suppose a middle ground would be allowing the subsidurial to "speak" in the chat, but not being hail-able. Which I suppose could be achieved be introducing the mute personality, however I'd have to do a bit more digging to ascertain the feasibility of that.

PR #5808, opened 2021-03-22
Minor coalition mission fixes

jerith:
**Bugfix:** This PR doesn't address an existing issue.

## Fix Details
I noticed two minor problems with the Saryd Couple mission chain while playing a recent continuous build. This PR fixes both:
 * a typographical error: `alhough` -> `although`
 * the couple can't afford their stated purchases: increased budget from `1,500,000` to `two and a half million`

The budget probably needs a little explanation. The Arach Courier's ship data has `"cost" 723000`, which is about half of their original 1.5M budget (the other half being spent on the land). However, that's just the base hull and the actual purchase price of the ship includes the installed outfits, for a total of ~1.2M. In addition, they still need to pay the 325k mission reward -- leaving them with a significant shortfall even before shopping for furniture.

By increasing their budget to 2.5M and inflating property prices a bit, we get the following breakdown:
 * 1.25M for the land (still "over half" the budget, barely)
 * 1.2M for the ship
 * -300k from selling the propulsion systems, presumably they'd keep the power and repair modules
 * 325k for the human taxi service
 * a little left over for furnishings and the Saryd equivalent of ramen

I also rewrote the amount in words to make it feel more like an approximation. (`2.5 million` would also be fine, but `2,500,000` seems way too precise.)

## Testing Done
Minor edits to existing conversation text only, no testing necessary.

## Save File
No save provided.

### DIFF
--- a/data/coalition/coalition missions.txt
+++ b/data/coalition/coalition missions.txt
@@ -1139,7 +1139,7 @@ mission "Saryd Couple 3"
 	destination "Saros"
 	on offer
 		conversation
-			`The couple leaves your ship and heads to a real estate agency. From what they told you during the trip, they have about 1,500,000 credits they can spend, which they claim is in line with the standard properties in the most valued (and most sought-after) locations here, and that they should still have enough money to pay you afterwards.`
+			`The couple leaves your ship and heads to a real estate agency. From what they told you during the trip, they have about two and a half million credits they can spend, which they claim is in line with the standard properties in the most valued (and most sought-after) locations here, and that they should still have enough money to pay you afterwards.`
 			`	They return from consulting, the both of them with disappointed looks on their faces. Inturi starts looking at the ships landing and departing while Aunaris comes talk to you.`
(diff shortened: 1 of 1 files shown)


data/coalition/coalition missions.txt
```diff
@@ -1139,7 +1139,7 @@ mission "Saryd Couple 3"
 	destination "Saros"
 	on offer
 		conversation
-			`The couple leaves your ship and heads to a real estate agency. From what they told you during the trip, they have about 1,500,000 credits they can spend, which they claim is in line with the standard properties in the most valued (and most sought-after) locations here, and that they should still have enough money to pay you afterwards.`
+			`The couple leaves your ship and heads to a real estate agency. From what they told you during the trip, they have about two and a half million credits they can spend, which they claim is in line with the standard properties in the most valued (and most sought-after) locations here, and that they should still have enough money to pay you afterwards.`
```
jerith, line 1142: Searching the data for `million` and `,000` found only one other instance of the latter being used for a value of a million or more, and that was an exact payment amount. For the former, words (`two and a half million`) occur slightly more than digits (`2.5 million`) although none of the existing cases had fractional amounts (so `ten million` vs `10 million`).

Feel free to change the formatting to whatever version makes the most sense, though -- I care far more about the size of the budget than its textual representation. :-)

roadrunner56 (approved):
Looks good to me.

PR #5809, opened 2021-03-22
Make some Remnant news comprehensible after completing sign studies

jerith:

## Summary

It bothered me a little on my recent playthrough that after going to all the trouble of learning Remnant sign language well enough for complex mission-related conversations, I was still unable to understand random people in spaceports. This PR makes the `remnant incomprehensible` news block conditional on not having learned to sign and adds a `remnant comprehensible` news block (conditional on having learned to sign) that makes the formerly-incomprehensible chatter somewhat more comprehensible.

This isn't particularly imaginative, but it fixes a minor continuity issue and can serve as a starting point for further improvements to Remnant news. I considered having several variants that became progressively more comprehensible through the "learn sign" mission chain, but there's a very short window of opportunity to actually see spaceport news between those missions and it didn't seem worthwhile to add stuff that almost nobody would ever see.

## Save File
No save provided. (I don't currently have access to a computer that can play CI builds.)

## PR Checklist

### DIFF
--- a/data/remnant/remnant news.txt
+++ b/data/remnant/remnant news.txt
@@ -368,6 +368,8 @@ news "remnant comprehensible"
 		word
 			"make out"
 			"catch"
+			"see"
+			"discern"
 		word
 			" is that they are "
 		word
--- a/data/remnant/remnant news.txt
+++ b/data/remnant/remnant news.txt
@@ -360,7 +360,11 @@ news "remnant comprehensible"
 		word
 			". "
 		word
-			"From this angle, all you can "
+			"From this angle, "
+			"Without them facing you, "
+		word
+			"all you can "
+			"what you can "
 		word
 			"make out"
 			"catch"
--- a/data/remnant/remnant news.txt
+++ b/data/remnant/remnant news.txt
@@ -213,6 +213,8 @@ news "remnant security advisory"
 news "remnant incomprehensible"
 	location
 		government "Remnant"
+	to show
+		not "remnant: sign studies complete"
 	name
 		word
 			"Spaceport"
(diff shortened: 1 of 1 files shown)


Zitchas:
I haven't reviewed the implementation, but at a quick glance this looks like a good addition. Thanks!

movingpictures:
Good pickup! I was going to bring this forward, but felt it too trivial to do so.

data/remnant/remnant news.txt
```diff
@@ -303,3 +305,71 @@ news "remnant cafeteria"
...
+			"each other"
+			"another Remnant"
+			"other Remnant"
+			"a Remnant at the information desk"
+			"a dockworker"
+			"something or someone you can't see"
+		word
+			". "
+		word
+			"From this angle, all you can "
```
ravenshining, line 363:
```suggestion
			"From this angle, "
			"Without them facing you, "
		word
			"all you can "
			"what you can "
```

data/remnant/remnant news.txt
```diff
@@ -303,3 +305,71 @@ news "remnant cafeteria"
...
+		word
+			"make out"
+			"catch"
```
ravenshining, line 366:
```suggestion
			"catch"
			"see"
			"discern"
```

ravenshining (approved):

data/remnant/remnant news.txt
```diff
@@ -213,6 +213,8 @@ news "remnant security advisory"
 news "remnant incomprehensible"
 	location
 		government "Remnant"
+	to show
+		not "remnant: sign studies complete"
```
Zitchas, line 217:
```suggestion
```

I think it would be better to have these be a mix. Instead of basically eliminating the incomprehensible and replace it with the somewhat comprehensible list, I think it is better to leave the existing Remnant news as-is. That way, the news will become a mix with sometimes the player does understand them, sometimes they don't.  We could then add additional comprehensible news set triggered at the end of the Cognizance missions. This would result in a gradient of slowly increasing amount of comprehensible news over time.

data/remnant/remnant news.txt
```diff
@@ -303,3 +305,77 @@ news "remnant cafeteria"
...
+			"catch"
+			"see"
+			"discern"
+		word
+			" is that they are "
+		word
+			"discussing a recent salvage operation"
+			"making lunch plans"
+			"arguing about engine efficiencies"
+			"gossipping about a visitor from outside the Ember Waste"
```
roadrunner56, line 379:
```suggestion
			"gossiping about a visitor from outside the Ember Waste"
```

data/remnant/remnant news.txt
roadrunner56: Have to agree here, the player isn't completely fluent, so having news messages appear where they don't understand something makes sense.

roadrunner56 (changes requested):
Besides the small spelling mistake and Zitchas' comment, the only other thing I would suggest for this PR is to add more variety to the types of conversation subjects the Remnant can have. Having only four means that it won't be long until the player sees the same topic repeated over and over again.

Amazinite (changes requested):
Looks good to me, aside from the other review comments you haven't yet gotten to.

Amazinite:
Anyone home?

data/remnant/remnant news.txt
jerith: While both "gossipping" and "gossiping" are acceptable spellings, it seems the former is in fewer spellcheck dictionaries. Thanks for the correction. :-)

jerith:
Heya. Sorry, life got busy and pulled me away from this for a while. :-/

Thanks for the updates and for merging.

PR #5813, opened 2021-03-26
Choose sensible destinations for remnant tech retrieval missions

jerith:
**Bugfix:** This PR doesn't address an existing issue.

Upon showing a Remnant engineer my pair of S-970 Regenerators, I was asked to please deliver them to Esquiline. While it's possible that the Remnant have some legitimate reason for building an impact testing facility on a gas giant, actually delivering them in any gaslined ship available to the player requires lots of cargo expansions and imported power/engines small enough to fit in the remaining outfit space. (I'm pretty sure it can be done in a Penguin. It can definitely be done in a Petrel, although there's no in-game indication that Petrels are gaslined. It definitely can't be done in a Puffin or a Tern. It would be easy to do in a Heron, but only if you're @Zitchas. I assume it can be done in a fleet of gaslined ships, but I thought it would be impolite deliver of a dismantled pile of parts.)

## Fix Details
I fixed all the tech retrieval missions by adding `attributes outfitter` to the destination selector, based on the broken jump drive missions having that. I considered adding `planet "Aventine" "Caelian" "Viminal"` instead, but that felt uglier.

## Testing Done
I haven't tested the fix, but I spent *way* too long fiddling with gaslined ships while trying to find a way to complete the buggy mission. (By the time I noticed the problem, my oldest available save already had the mission accepted.)

## Save File
No save provided.

### DIFF
--- a/changelog
+++ b/changelog
@@ -31,6 +31,7 @@ Version 0.9.13:
       * The Remembrance Day jobs now properly have a deadline. (@Amazinite)
       * Shortened the descriptions of some jobs as to not overflow the UI. (@Amazinite)
       * Added missing "variant" keyword to NPC fleets in several Wanderer missions. (@tehhowch)
+      * Remnant "tech retrieval" missions have more appropriate destinations. (@jerith)
     * Engine bugs:
       * Missions with illegal cargo that fail when scanned can no longer be scanned again before landing. (@Amazinite)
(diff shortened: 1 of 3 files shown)


Zitchas (approved):
This seems reasonable to me. In the future there will be more gaslined Remnant ships, and missions similar to these that *should* be able to go to Esquiline and other places. But it fits with the lore and motivations that the Remnant have right now that the don't send the player to any of the alternate locations for these early missions.

PR #5818, opened 2021-03-30
Ka'het MHD Launcher!

beccabunny:

## Summary
This PR adds a new, odd weapon for the Ka'het: an emergency launcher installed only in their unmanned drones and that shoots out its MHD generator if the Mar are under attack, creating a big explosion which in the best case leaves the drone without means of creating energy and in the worst one kills the drone as well. This is also a weird weapon in the fact that it doesn't actually take up weapon space, only a large amount of outfit space (32 outfit per ammunition). The explosion isn't quite nuke level, and the ammo is completely unguided. Also has a small "firing hull".
immagine

## PR Checklist
 - [x] I uploaded the necessary image, blend, and texture assets here: [mhd launcher.zip](https://github.com/endless-sky/endless-sky/files/6227245/mhd.launcher.zip)
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/171


Amazinite (commented):
Looks great. Just have two comments on ways we could change this. (The former I'd like to have, the latter is more up to you.)

> The explosion isn't quite nuke level, and the ammo is completely unguided.

That's some expensive ammunition. I certainly wouldn't be opposed to making the explosion nuke level, or perhaps even stronger than a nuke. It's not like you can just easily buy more MHDs, whereas having nukes unlocked means you can buy more nuke ammo whenever you want.

> Also has a small "firing hull".

I will note that a ship's hull must be high enough such that any "firing hull" weapon doesn't cause it to become disabled, so having firing hull could actually prevent a drone from using the MHD when it might need it the most.

PR #5819, opened 2021-03-31
Use correct vocabulary for hull repair in Ka'het ships

Anarchist2:
Changing `"hull repair"` to `"hull repair rate"`.

Also to note: The Ka'het ships have shield heat without shield regeneration, and hull repair rate without any cost to it. Could someone tell me whether that was intentional or not?


beccabunny:
> Could someone tell me whether that was intentional or not?

Intentional yes

PR #5823, opened 2021-04-04
Adjust fuel energy mechanics

Ferociousfeind:
## Feature Details
This PR changes "fuel consumption" to scale with the amount of energy capacity left to fill in a frame (with a cap at 100% strength), and to shut off if the batteries are already full, for a similar in-universe reason that active cooling scales with heat level. Your scarce hyperspace fuel is not consumed if no energy is being used and the batteries are full, to save on fuel.

It also adds a decidedly higher-tech "regenerative fuel consumption" which can run in reverse, consuming energy and heat to produce fuel. Both attributes will play nice with each other, having 5 of each will let you consume 10 fuel per frame, and if your batteries are being vastly overcharged, you can make back up to 5 fuel per frame.

Adding more "fuel energy" to the mix will actually decrease the amount of fuel you would get in return for some amount of extra energy produced, so that is left as-is, but more "fuel heat" will cause more heat to be sucked out of your ship when your batteries are overcharged, so "fuel heat" takes the maximum between its calculated value and 0 (AKA, it will never go negative and act as a cooling outfit).

## Testing Done
Flew a ship with ramscoop and fuel-consuming energy generation. Simply flying around kept fuel at relatively full. Firing weapons and thrusting and turning all together, depleting the battery, led to fuel slowly draining. Once the batteries refilled, the draining stopped, and the ramscoop started slowly refilling fuel.

Flew a ship with _no_ ramscoop, and a hybrid of regular energy generation and fuel-consuming energy generation. After jumping to deplete fuel, fuel started refilling even though I had no fuel generation or ramscoop. This was because the regular energy generation was overcharging the battery, and so the formula determining how much energy to produce to move "energy" to equal "energy capacity" was negative, consuming energy and heat, and producing fuel. The fix, only considering fuel-consuming energy generation if "energy" is already less than "energy capacity". Now, that only happens to "regenerative fuel consumption".

Flew a ship with no fuel consumption, expecting that to crash the game (by putting a 0 in a denominator) but nothing went wrong. I have not tested "fuel energy" and/or "fuel heat" without "fuel consumption", though, so watch out for that, that might cause a crash. It is not an intended situation, though, so I wouldn't worry about it (after all, you can have weapons use zero-outfit-space outfits as ammo, and the outfitter will cause the game to freeze when calculating how much to restock).

## Performance Impact
This should have no strong performance impact, it's one extra division being done each frame while there is fuel aboard a ship to use. Among the dozens or hundreds of others, I think the game will be fine.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2054,13 +2054,18 @@ void Ship::DoGeneration()
 		heat -= coolingEfficiency * attributes.Get("cooling");
 		
 		// Convert fuel into energy and heat only when the required amount of fuel is available.
-		if(attributes.Get("fuel consumption") <= fuel)
+		if((attributes.Get("fuel consumption") + attributes.Get("regenerative fuel consumption")) <= fuel)
 		{
 			// Also, only convert fuel if it won't create more power than the batteries can hold.
-			if(energy < attributes.Get("energy capacity"))
+			double scale = min(1., (attributes.Get("energy capacity") - energy) / attributes.Get("fuel energy"));
+			// "regenerative fuel consumption" can run in reverse
+			if (scale < 0)
+				fuel -= scale * attributes.Get("regenerative fuel consumption");
+			else
+				fuel -= scale * (attributes.Get("fuel consumption") + attributes.Get("regenerative fuel consumption"));
+			// the only case where this fails is if no fuel consumption has occurred in the first place
+			if(scale >= 0 || attributes.Get("regenerative fuel consumption"))
 			{
-				double scale = min(1., (attributes.Get("energy capacity") - energy) / attributes.Get("fuel energy"));
-				fuel -= scale * attributes.Get("fuel consumption");
 				energy += scale * attributes.Get("fuel energy");
 				heat += scale * attributes.Get("fuel heat");
 			}
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2055,10 +2055,12 @@ void Ship::DoGeneration()
 		
 		// Convert fuel into energy and heat only when the required amount of fuel is available.
 		if(attributes.Get("fuel consumption") <= fuel)
-		{	
-			fuel -= attributes.Get("fuel consumption");
-			energy += attributes.Get("fuel energy");
-			heat += attributes.Get("fuel heat");
+		{
+			// Also, only convert fuel if it won't create more power than the batteries can hold.
+			scale = min(1., (attributes.Get("energy capacity") - energy) / attributes.Get("fuel energy"));
+			fuel -= scale * attributes.Get("fuel consumption");
+			energy += scale * attributes.Get("fuel energy");
+			heat += scale * attributes.Get("fuel heat");
 		}
 		
 		// Apply active cooling. The fraction of full cooling to apply equals
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2058,6 +2058,7 @@ void Ship::DoGeneration()
 		{
 			// Also, only convert fuel if it won't create more power than the batteries can hold.
 			double scale = min(1., (attributes.Get("energy capacity") - energy) / attributes.Get("fuel energy"));
+			scale = max(-1., scale);
 			// "regenerative fuel consumption" can run in reverse
 			if (scale < 0)
 				fuel -= scale * attributes.Get("regenerative fuel consumption");
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2057,10 +2057,13 @@ void Ship::DoGeneration()
 		if(attributes.Get("fuel consumption") <= fuel)
 		{
 			// Also, only convert fuel if it won't create more power than the batteries can hold.
-			scale = min(1., (attributes.Get("energy capacity") - energy) / attributes.Get("fuel energy"));
-			fuel -= scale * attributes.Get("fuel consumption");
-			energy += scale * attributes.Get("fuel energy");
-			heat += scale * attributes.Get("fuel heat");
+			if(energy < attributes.Get("energy capacity"))
+			{
+				double scale = min(1., (attributes.Get("energy capacity") - energy) / attributes.Get("fuel energy"));
+				fuel -= scale * attributes.Get("fuel consumption");
+				energy += scale * attributes.Get("fuel energy");
+				heat += scale * attributes.Get("fuel heat");
+			}
 		}
 		
 		// Apply active cooling. The fraction of full cooling to apply equals
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2003,6 +2003,20 @@ void Ship::DoGeneration()
 	// maximum capacity for the rest of the turn, but must be clamped to the
 	// maximum here before they gain more. This is so that, for example, a ship
 	// with no batteries but a good generator can still move.
+
+	// Run fuel consumption in reverse to get rid of excess energy from the last frame
+	// so that energy CAN be available for the frame, but if it isn't used, then the
+	// fuel consumption retroactively didn't happen, and we just don't talk about it.
+	// This also has the effect of shutting off fuel consumption completely if another
+	// energy-generating alternative can cover energy usage for the previous frame.
+	if(attributes.Get("fuel consumption") && energy > attributes.Get("energy capacity"))
+	{
+		double scale = min(1., (energy - attributes.Get("energy capacity")) / attributes.Get("fuel energy"));
+		energy -= scale * attributes.Get("fuel energy");
+		heat -= scale * attributes.Get("fuel heat");
+		fuel += scale * attributes.Get("fuel consumption");
+	}
+
 	energy = min(energy, attributes.Get("energy capacity"));
 	fuel = min(fuel, attributes.Get("fuel capacity"));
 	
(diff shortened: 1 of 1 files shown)


Amazinite:
Isn't "regenerative fuel consumption" a contradiction of terms? And isn't it also, you know, exactly what "fuel generation" is? Wouldn't it make more sense to just use "fuel generation" use energy and produce heat if it's present? And instead of hijacking two preexisting attributes, wouldn't it make more sense to create `"fuel generation energy/heat"`?

In the case of fuel generation, there's already a function to do the heavy lifting for you on this one.
```c++
double fuelGeneration = attributes.Get("fuel generation");
double fuelGenEnergy = attributes.Get("fuel generation energy") / fuelGeneration ;
double fuelGenHeat = attributes.Get("fuel generation heat") / fuelGeneration ;
DoRepair(fuel, fuelGeneration, attributes.Get("fuel capacity"), energy, fuelGenEnergy , fuel, 0., heat, fuelGenHeat);
```

And I totally haven't tested this, but I think you can also use the same function with "fuel consumption" to only add to your energy when it's needed.
```c++
double fuelEnergy = attributes.Get("fuel energy");
double fuelConsumption = attributes.Get("fuel consumption") / fuelEnergy ;
double fuelHeat = attributes.Get("fuel heat") / fuelEnergy;
DoRepair(energy, fuelEnergy, attributes.Get("energy capacity"), energy, 0., fuel, fuelConsumption, heat, fuelHeat);
```

tehhowch:
No other energy generation mechanisms account for the current state of the batteries, so why should this be different?  
Wouldn't this also break no-battery builds?

I'm also not a fan of this proposed "regenerative fuel consumption" at all, and if you'd like to add it, it should be in a separate PR that details why it is needed.

Zitchas:
Based on our discussion, this is part of several larger ideas.

The purpose is to introduce alternative reactors. Instead of simply having a single reactor economy where the only trade offs are energy per space and energy per heat, with the joke that is solar offering a little energy per distance to the star. This would introduce energy per unit of fuel. 

This has two applications:
1) In vanilla ES, this could be used to create generators that offer an alternate trade off. Perhaps they are the same size as other reactors, but generate virtually no heat in exchange for consuming their fuel. Or perhaps they are much smaller than other reactors, but generate the same amount of heat. (Basically, trading either space used or heat produced for fuel consumption.

2) Alternatively, some people have expressed a dislike that reactors have no inputs and can run forever without fuel. Literally something for nothing. This mechanic would enable these people to use plug-ins or data-file editing to replace the current reactors with equivalents that do use fuel.

Lore:
Normally, burning fuel to make power is a one-way process, like running a diesel generator. No amount of pushing energy back into the generator will create fuel. So, the first version takes this into account, and has the limiter so that if the ship has extra energy (perhaps they have a dual solar panel + fuelled generator build so the one powers them near stars, and the other when away from stars). So normally, when the other power source is sufficient to run the ship and fill the batteries, the fuelled generator turns off. This is why the battery check needs to be there. If the battery is full, then there is no need to run the generator.

It makes sense, though, that more advanced factions/civilizations might be using something more efficient and complex than a diesel generator. Perhaps they have a process whereby they can generate power from fission, which is basic nuclear power, but they have figured out how to collect the results of the reaction and - when they have access to a lot of power, they can fuse those results back together, effectively reversing the process. 

In mechanical terms, the former is the T1 fuelled generator, the later is the T2+ fuelled generator.

Ferociousfeind:
Hmm. You're right, hijacking pre-existing attributes is bad etiquette. There may be outfits that want to constantly consume fuel to provide an unrelated attribute (for example, a fuel-guzzling shield capacitor quickly comes to mind)

The idea was originally to stop "fuel consumption" from eating fuel if the batteries were already full, but my implementation led to a weird edge case where it'd run in reverse to get batteries to exactly 100%, also producing fuel and reducing heat in the process. I thought that was cool! But it'd need to be properly addressed, so that's why there's a billion commits in this PR.

I called it "regenerative fuel consumption" in reference to regenerative braking, usually using fuel, but capable of producing fuel using excess energy. The reason I've proposed this, and not regular energy generation, to respond to battery levels, is because regular energy generation is accompanied by heat generation, which a bunch of other outfits also use just to produce general heat.

I'm not sure how it reacts to no-battery builds, I honestly should've actually tested that out to see how it'd work (maybe it's not a problem?), but Zitchas proposed (on discord) simply leaving it be, letting no-battery fuel-consuming builds be generally unviable (a two-ton supercapacitor solves the problem, too). The other potential solution is, perhaps, simulating energy capacity, and then only generating enough energy, consuming enough fuel, to bring the simulated battery back to full, and then in reality that will have generated only enough energy to sustain the ship for that frame, with 0 excess.

I'm going to tinker with "doRepair" and better attribute names, and then commit changes to this PR once I've settled on something less egregious. Perhaps we can discuss implementation casually over discord? ~~I'll waffle about ultimately nothing less~~

source/Ship.cpp
```diff
@@ -2049,27 +2063,22 @@ void Ship::DoGeneration()
...
-				energy += scale * attributes.Get("fuel energy");
-				heat += max(0., scale * attributes.Get("fuel heat"));
-			}
+		if(attributes.Get("fuel consumption") <= fuel)
+		{
+			double fuelEnergy = attributes.Get("fuel energy");
+			double fuelConsumption = attributes.Get("fuel consumption") / fuelEnergy;
+			double fuelHeat = attributes.Get("fuel heat") / fuelEnergy;
+			DoRepair(energy, fuelEnergy, attributes.Get("energy capacity") + fuelEnergy, energy, 0., fuel, fuelConsumption, heat, fuelHeat);
 		}
```
Amazinite: DoRepair already ensures that the attributes are within allowable bounds, so you don't need the if statement. Also, after some more thinking, it might be better to choose the max between the energy capacity and the fuel energy instead of adding the two. This way we're only overcharging the ship when we need to (i.e. no wasted fuel). This also effectively removes the need for the conversion of energy back into fuel, at least for battery loadouts.
```
DoRepair(energy, fuelEnergy, max(attributes.Get("energy capacity"), fuelEnergy), energy, 0., fuel, fuelConsumption, heat, fuelHeat);
```

source/Ship.cpp
```diff
@@ -2003,6 +2003,20 @@ void Ship::DoGeneration()
...
+	// This also has the effect of shutting off fuel consumption completely if another
+	// energy-generating alternative can cover energy usage for the previous frame.
+	if(attributes.Get("fuel consumption") && energy > attributes.Get("energy capacity"))
+	{
+		double scale = min(1., (energy - attributes.Get("energy capacity")) / attributes.Get("fuel energy"));
+		energy -= scale * attributes.Get("fuel energy");
+		heat -= scale * attributes.Get("fuel heat");
+		fuel += scale * attributes.Get("fuel consumption");
+	}
+
```
Amazinite, line 2019: You should still be able to make use of DoRepair for this, just with switching around some of the values.

```c++
DoRepair(fuel, fuelConsumption, attributes.Get("fuel capacity"), energy, fuelEnergy, fuel, 0., heat, -fuelHeat);
```

Given that we'd be making use of these attributes in two spots then (once for energy to fuel then a second time for fuel to energy), you'd want to move the variables from lower down (L2078-2080 right now) up to here and make them const.
```c++
const double fuelEnergy = attributes.Get("fuel energy");
const double fuelConsumption = attributes.Get("fuel consumption") / fuelEnergy;
const double fuelHeat = attributes.Get("fuel heat") / fuelEnergy;
```

Also, a comment as simple as `// Convert any excess energy back into fuel if able to.` would do.

But I think that this is an entirely separate issue that should be handled elsewhere, as while I can get on board with fuel consumption and generation being bounded by energy and heat costs just as shield and hull repair are bounded, this whole "convert excess energy back into fuel" thing is a new mechanic entirely.

ravenshining:
I absolutely love this overall idea, except it doesn't make sense to me that producing fuel would decrease heat - bit of a thermodynamic problem there.  I can't wait to incorporate this mechanic into my own plugins!

As for tehhowch's concerns about precedent, other energy generation do not consume a limited resource.  I live off grid, I don't run my generator when my house batteries are full!   Provided it doesn't break batteryless ships, I see this as a marked improvement - in fact, I'd like to see the mechanic carried over to other generation types, have them reduce (though ideally, not eliminate) their heat output when batteries are full.

ravenshining:
Ran same tests.  The changes to fuel consumption were a blast!  I had a lot of fun there, everything working as expected, including batteryless ships.

Fuel generation, on the other hand, I couldn't get to work as I expected.  I had to set `"fuel generation"` to a negative value in order to get the outfit to generate fuel, and no matter what I entered to `"fuel generation heat"` - I tried low values at first, then kept adding zeroes and switching signs - I couldn't get it to generate heat while producing fuel.  On the plus side, it did generate fuel in a batteryless ship when another power source was provided.

Ferociousfeind:
> 
> 
> Ran some tests. The changes to fuel consumption were a blast! I had a lot of fun there, everything working as expected, including batteryless ships.
> 
> Fuel generation, on the other hand, I couldn't get to work as I expected. I had to set `"fuel generation"` to a negative value in order to get the outfit to generate fuel, and no matter what I entered to `"fuel generation heat"` - I tried low values at first, then kept adding zeroes and switching signs - I couldn't get it to generate heat while producing fuel. On the plus side, it did generate fuel in a batteryless ship when another power source was provided.

Oh, I didn't catch this. I'll have to double-check the code to see what went wrong. Possibly, the way I ran fuel consumption in reverse, is eating up all the fuel generation by running THAT in reverse too.

Ferociousfeind:
There is probably a better way to do this, I should redo this a better way. It is probably also out of date and that part of ship.cpp has been reshuffled and changed in different ways.

PR #5826, opened 2021-04-05
Fix: Spelling mistake in Remnant: Keystone Research 1 - payed -> paid

Zitchas:

## Fix Details
The word "paid" is mis-spelled as "payed" in Remnant: Keystone Research 1.

Thanks to Bardagh on discord for spotting it.

Note, payed is actually a properly spelled word, but it has nothing to do with buying things. It has to do with sealing ships.

### DIFF
--- a/data/remnant/remnant missions.txt
+++ b/data/remnant/remnant missions.txt
@@ -3088,7 +3088,7 @@ mission "Remnant: Keystone Research 1"
 			`	"Greetings, <first>! I trust you are settling in among us?" he signs using simple gestures. "I picked up a rumor that you are learning our language, too." You sign the affirmative as you sit down. "Good, that is useful." He pauses, then continues, "I never introduced myself before. My name is Aeolus, and I help manage resource logistics for our fleets." He eyes you for a moment, then ventures, "You came in here with the appearance of looking for something. Is there something on your mind?"`
 			choice
 				`	"How is the outfit business going?"`
-				`	"That last job payed rather well. Do you have any more like it?"`
+				`	"That last job paid rather well. Do you have any more like it?"`
 					goto crystaljobs
 			`	"Well, fairly well. Stocks are up, there are quite a few new outfits in the current, and some recent supply line disruptions have been resolved," he gestures appreciatively. "I am told that we have you to thank for some of that.`
 			`	"Those 'quantum keystones' you brought back have been quite interesting. The process of dismantling them for analysis yielded some intriguing results." He notices your expression. "Oh, you were expecting me to sell them, weren't you?"`


data/remnant/remnant missions.txt
```diff
@@ -3088,7 +3088,7 @@ mission "Remnant: Keystone Research 1"
 			`	"Greetings, <first>! I trust you are settling in among us?" he signs using simple gestures. "I picked up a rumor that you are learning our language, too." You sign the affirmative as you sit down. "Good, that is useful." He pauses, then continues, "I never introduced myself before. My name is Aeolus, and I help manage resource logistics for our fleets." He eyes you for a moment, then ventures, "You came in here with the appearance of looking for something. Is there something on your mind?"`
 			choice
 				`	"How is the outfit business going?"`
-				`	"That last job payed rather well. Do you have any more like it?"`
+				`	"That last job paid rather well. Do you have any more like it?"`
 					goto crystaljobs
 			`	"Well, fairly well. Stocks are up, there are quite a few new outfits in the current, and some recent supply line disruptions have been resolved," he gestures appreciatively. "I am told that we have you to thank for some of that.`
```
tehhowch, line 3093: "in the current" seems incomplete.

jerith: I read it as "in the pipeline".

jerith (approved):

data/remnant/remnant missions.txt
Zitchas: As Jerith says, it is meant as an analogue to "in the pipeline," except more focused on flows and streams rather than tubes. A vocabulary hint that the Remnant view progress and development as a sometimes unpredictable flow rather than a tightly controlled process. Something that is in a pipeline is pretty much guaranteed, barring a major failure of the pipe. A river flow or air current, on the other hand, are generally predictable but frequently have unforseen accelerations or changes. Remnant are consciously pushing innovation, and accept that new things - even if they are predictable - may yet have interesting twists, deviations, and changes before they arrive.

Likewise, I realize that "in the current" is a bit jarring to our modern ears, as it is both an unfamiliar expression and embodies a non-typical acceptance of uncertainty in planned or upcoming developments.  The fact that it is jarring is a benefit, as it emphasizes the fact that - appearances to the contrary - this is *not* "just another group of humans."

tehhowch: It's forced and awkward, IMO.

"Current" appears after two words that often demarcate a prepositional phrase, and the sequence of these words, "in the current", are almost always followed by a noun. There is no other nautical context to suggest the interpretation of `current` as a noun rather than an adjective, and the player just asked how things are going _right now_, so the time-based definition of the word is especially relevant. The reader's most likely reasoning for the sentence feeling off is "whoever wrote this forgot a word."

Since the player can choose to address this really awkward phrasing shortly after, I agree we don't need to change it here.

PR #5837, opened 2021-04-13
refactor(content): Buff civilian Coalition ships

Arachi-Lover:
----------------------
**(Balance)**

## Summary
It is odd that the Coalition, a group that's been independent and at peace for six thousand years, and which the player can only access once they have a Jump Drive, produces freighters and transports with little advantage over their human counterparts other than slightly more HP, it so I tried my hand at buffing them.
The ship that bugged me the most was the Arach Hulk, the largest freighter in the Coalition, having less cargo space than the Bulk Freighter. I'm not discounting HP as an advantage, but I don't think anyone buys freighters primarily for how much damage they can take.

The buffs are the following:

- Saryd Light Freighters got a 10% increase to bunks and cargo space.
- Saryd Heavy Freighters got a 20% increase to bunks and cargo space.
- All Kimek ships got a 50% increase to bunks, and a 10% increase to cargo space.
- Arach Light Freighters got a 20% increase to cargo space.
- Arach Heavy Freighters got a 40% increase to cargo space.

I don't expect the changes to be perfect as I just made up what the % increases should be in my head, so feel free to suggest other values or even other ways of buffing them.

### DIFF
--- a/data/coalition/coalition ships.txt
+++ b/data/coalition/coalition ships.txt
@@ -24,7 +24,7 @@ ship "Arach Courier"
 		"drag" 2.9
 		"heat dissipation" .65
 		"fuel capacity" 300
-		"cargo space" 78
+		"cargo space" 94
 		"outfit space" 148
 		"weapon capacity" 39
 		"engine capacity" 40
@@ -69,7 +69,7 @@ ship "Arach Freighter"
 		"drag" 10.6
 		"heat dissipation" .55
 		"fuel capacity" 500
-		"cargo space" 310
+		"cargo space" 372
 		"outfit space" 349
 		"weapon capacity" 132
 		"engine capacity" 92
@@ -119,7 +119,7 @@ ship "Arach Hulk"
 		"drag" 18.9
 		"heat dissipation" .50
 		"fuel capacity" 500
-		"cargo space" 541
+		"cargo space" 757
 		"outfit space" 520
 		"weapon capacity" 208
 		"engine capacity" 148
@@ -173,7 +173,7 @@ ship "Arach Spindle"
 		"drag" 10.8
 		"heat dissipation" .55
 		"fuel capacity" 500
-		"cargo space" 280
+		"cargo space" 392
 		"outfit space" 434
 		"weapon capacity" 138
 		"engine capacity" 148
@@ -225,7 +225,7 @@ ship "Arach Transport"
 		"drag" 7.2
 		"heat dissipation" .60
 		"fuel capacity" 400
-		"cargo space" 190
+		"cargo space" 228
 		"outfit space" 281
 		"weapon capacity" 96
 		"engine capacity" 70
@@ -1331,12 +1331,12 @@ ship "Kimek Briar"
 		"shields" 3600
 		"hull" 1700
 		"required crew" 3
-		"bunks" 25
+		"bunks" 38
 		"mass" 180
 		"drag" 3.8
 		"heat dissipation" .75
 		"fuel capacity" 700
-		"cargo space" 21
+		"cargo space" 23
 		"outfit space" 245
 		"weapon capacity" 89
 		"engine capacity" 60
@@ -1377,12 +1377,12 @@ ship "Kimek Spire"
 		"shields" 14200
 		"hull" 4900
 		"required crew" 19
-		"bunks" 197
+		"bunks" 296
 		"mass" 570
 		"drag" 8.4
 		"heat dissipation" .60
 		"fuel capacity" 800
-		"cargo space" 64
+		"cargo space" 70
 		"outfit space" 489
 		"weapon capacity" 256
 		"engine capacity" 129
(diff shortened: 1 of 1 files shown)


Zitchas:
As described, the changes look good.

That being said, you raise a good point: I don't think anyone is travelling to Coalition space to buy a bulk-freighter for combat purposes. It might be worth taking a look at the pure combat capabilities for the civilian ships (ex. armor, shields, regen, and weaponry) in the context of the Coalition's enforced "non-combat" capabilities. Might be worth trimming them a little bit in those regards. Not much, but a little bit.

But yeah, if one is making the effort to go the safest and least-dangerous part of the galaxy to buy civilian ships, the driving factor should be civilian capabilities. So it makes sense to buff them like this.

Arachi-Lover:
When the campaigns come, the first few combat missions for the Resistance will have them using *only* civilian Coalition ships against Heliarch ships. Long story short the civilian ships need to be able of combating Heliarch ships to an extent, and the gap is already way too big as it is, so I don't want any reduction to their combat stats.
It wouldn't make sense in the lore either, as their civilian ships have gun ports/turret mounts, and employ regen outfits, so that they *can* survive and fight the Quarg, or any other threat, if needed.

Zitchas:
That makes sense. If the resistance has been subtly growing for years they would have been subtly pushing for more combat capable ships too. (Resistance wants civilian ship to fight the Heliarchs, Heliarchs want combat capable civilian ships to fight the Quarg... I see that as being limited a bit by the lack of active threats in their space right now, but I suppose countered by the active push.

OK. Either way, having more bunks and/or cargo shifts the focus a bit away from the combat nature, so that's a good thing.

arkhne:
I don't want to take away from what's being done here (because I generally approve, the Kimek ships in particular needed a buff), however, after improving these stats, I do believe that you will need to increase their mass, as they were already quite light for their capabilities (even in the current setup where everything is quite light).

I'm not certain that making all Coalition ships lighter was (or was not) a deliberate design choice, I could see a small difference being due to their "superior alloys and construction techniques", however, I personally believe that the difference was already too big. With these changes, I am definitely concerned about this.

I can take a closer look at the numbers later, and see how they stack up in-game, but for right now, this is all the concerns that I have. I do agree that we shouldn't be touching their combat capabilities, if you account for when they become available, your civilian fleets are probably going to want/need some bulk behind them (not neccessarily weapons, but the ability to take a punishment), and this is what made them attractive over the better cargo/bunk options for smaller fleets (or solo ships, even).

Amazinite:
Given that we already have multiple PRs open looking to address the speeds of *all* ships, I don't see the use in spending time on a separate PR to address the speeds of just Coalition ships relative to all other ships, a change that would be overwritten anyway by the aforementioned PRs that address all ships at once.

arkhne:
> Given that we already have multiple PRs open looking to address the speeds of _all_ ships, I don't see the use in spending time on a separate PR to address the speeds of just Coalition ships relative to all other ships, a change that would be overwritten anyway by the aforementioned PRs that address all ships at once.

I strongly disagree with this attitude on the matter.
There isn't even a consensus on how to go about that change, and here we are, making the situation worse, without any care for the consequences of the actions, and speaking out against any suggestion to address the issue.

In this case, I have to strongly disagree with this change, since it negatively impacts the state of the game. No attribute of any ship should be touched until the sacred ground of existing PRs are resolved, since they're going to be used against any logical discussion or change.

ravenshining:
Maybe if this were accompanied by an enlarging of the freighter sprites, or filling them out more.  The Arach Hulk is just not that big compared to a Bulk Freighter.

Amazinite:
I think that looking at the size of the Hulk is definitely in order. Here we have the Deep River (1140 cargo), Hulk (541->757 cargo), and Bulk Freighter (600 cargo) next to each other. The Hulk looks like it has less space for cargo already, and we'd be buffing it here. So I don't think that a bigger Hulk sprite is out of the question.
image

Arachi-Lover:
I believe scaling up the paint and shadow layers in the Gimp files for the Arach ships would lead to lower image quality, so the problem with changing their size would be needing to repaint those layers in Gimp to match the old ones, something I don't trust myself to do properly.

Pointedstick (approved):
This seems pretty reasonable to me in its current form. You know me; I'm always in favor of merging and iterating rather than seeking (and failing to achieve) perfection on the first go-around. :)

Any strong objections to merging now? If so, can we make some tweaks to make this merge-ready quickly?

Zitchas (approved):
Looks good to me.

The only quibble I might have is possibly bumping the mass up a little bit alongside the increased space, particularly for bunks. Not much, mind you, just a little. Like, maybe +1 per three bunks, and +1 per 20 cargo. Ish. Rough ballpark sort of numbers. Shouldn't significantly change them, but I think it is good to reflect the mass that passenger and cargo frameworks take up.

Pointedstick:
@Arachi-Lover, what do you think about that?

Arachi-Lover:
Given there are multiple PRs open that change the mass of ships in attempting to rebalance speed, I don't see the use in changing the mass of the Coalition ships here, especially if the the change would be as minimal as suggested, given the mass would be changed again once one of said PRs (or at least the mass changes in them) was implemented.

Pointedstick:
All right, let's ship it as-is.

PR #5838, opened 2021-04-13
Coalition Fuel Module

Arachi-Lover:
----------------------
**Content (Artwork / Outfit)**

## Summary
Adds a new outfit to the Coalition, the Fuel Module, which is just a fuel storage outfit.
In terms of space efficiency for fuel stored, the Fuel Pod is at a 12.5, the Fuel Processor at a 14.8, and the Fuel Module at a 21.
Also replaces the Fuel Pods in both Coalition outfitters with the Fuel Module.

Edit: Assets: https://drive.google.com/drive/folders/1jBsJq7xuI559WA--H09r5t_msT9N12l6?usp=sharing


Terin:
Having the hefty chunk of fuel feels great, a couple of those and I can safely get from deep Saryd space to deep Arach space with no trouble. The cost seems too high, though. It's the most expensive piece of civilian Coalition tech, barring the large Shield/Hull modules, for what seem like standard containers. Maybe closer to the 250k price range?

data/coalition/coalition outfits.txt
```diff
@@ -171,6 +171,18 @@ outfit "Cooling Module"
...
+outfit "Fuel Module"
+	category "Systems"
+	licenses
+		Coalition
+	cost 250000
+	thumbnail "outfit/fuel module"
+	"mass" 19
+	"outfit space" -19
+	"fuel capacity" 400
+	description "In their early days of space exploration, the Saryds developed these to travel far away without worrying too much about their fuel."
```
ravenshining, line 183: How about a nod to some of the lore?  Plus a minor stylistic suggestion...
```suggestion
	description "In their early days of space exploration, when Coalition space was neither well settled nor well connected, the Saryds developed the fuel module to travel far from Saros without worrying about becoming stranded without fuel."
```

data/coalition/coalition outfits.txt
```diff
@@ -171,6 +171,18 @@ outfit "Cooling Module"
 	"cooling energy" 1.4
 	description "This cooling module consumes energy. To conserve power, it does not ramp up to full strength unless a ship is very close to overheating."
 
+outfit "Fuel Module"
+	category "Systems"
+	licenses
+		Coalition
+	cost 250000
```
ravenshining, line 178: Seems a bit expensive for what it is to me.

```suggestion
	cost 135000
```

data/coalition/coalition outfits.txt
Arachi-Lover: What lore is this a nod to?

ravenshining: I have no idea what you plan on taking or leaving from the Document, but I recall it saying that before the war with the Quarg, Coalition space had fewer hyperlinks.  That's what I'm referencing with the first edit.  The remaining three just make it sound better without actually adding anything to the message.

Arachi-Lover: It does nothing but harm (or at least be an annoyance) to the Saryds if their space was initially cut to be a miniature maze, with fewer links than there are currently, and there would be no point in the Archons keeping it that way, or keeping the "clusters" of each species isolated form one another, after the Quarg gave the Saryds Jump Drives, so I'm probably just gonna ignore that.
I'll apply the other three changes in a bit.

ravenshining: Right, that's your prerogative :-)

Personally I wasn't imagining an isolated maze though, just fewer hyperlinks, more like Human space is - Coalition space is extremely well-connected by comparison.  I imagine that some time in the future, once humans figure out the tech behind hyperlink creation, Human space will be just as well connected.

roadrunner56 (approved):
Looks good, glad to see more Coalition outfits making it into the game.

Arachi-Lover:
As per suggested by @arkhne on the discord, I added some `fuel protection` to the Fuel Module, and increased its price again, the justification for the price now being the protection offered. 
I also added a bit more to the description to acnowledge the attribute.
As I note in the commit where I added it, since there is nothing that deals fuel damage currently, it is more of a flavor stat, but I don't know how fuel damage might be used in the future, and the protection value might be too high, so I would like some thoughts on the changes.

roadrunner56:
Now that I think about it, it doesn't really make sense for a faction like the Coalition to have such a massive fuel pod, considering how the vast majority of their systems are inhabited. I could see the Heliarchs keeping it around in case of a Quarg invasion, though, for the same reason that civilian Coalition ships still have gun ports. Maybe put something in its description about that. 
Of course, you're the authority on Coal lore, so it's entirely up to you.

Arachi-Lover:
That's acnowledged in the description, "Nowadays, nearly every system on the Coalition has refuelling facilities, but some captains still find use in the extra jumps these modules offer."
As for it making sense to have something that's not terribly useful, there is no combat in Coalition space, the Shield and Repair modules exist and are widely used, and the Small Shield Module's description is similar to the above portion of the Fuel Module's, in that it acnowledges the outfit isn't exactly necessary: "Although combat is almost unheard of in Coalition space, almost all of their ships are able to recharge their shields when necessary."
All in all this outfit is just some flavor for the Coalition, give them a more efficient fuel storage item which also makes it so they don't use the same universal Fuel Pods sold in other places. It's also to have some more civilian outfits for this faction that outnumbers humanity and has a pretty sizable territory, something I've struggled to do for a while, since in thinking up new outfits I can't give them any weapons, and I don't want to have "Largerer Shield Module" or "Medium Battery Module" since those sound silly and don't really help with increasing variety.

Terin:
From a gameplay/space trucking perspective, one or two of those lets you go from say, the edge of Saryd space to the edge of Arach/Kimek space, without stopping to fuel anyone. Very handy if you happen to have a few fast courier missions, and a pilot could swap out their shield/hull modules for that extra distance, so it seems like a fine fit.

roadrunner56:
Yeah that's fair. Coalition space needs more outfits, and a fuel pod seems like a good fit.

Amazinite:
Not particularly wowed by the outfit image, but it's not like you can do too much with a fuel pod.

Issue #927, opened 2016-04-04
Enhancement: outfit variants


LocalGod79:
it's been raised before that the Hai and Lovelace sell the same Fusion Generator.  There are other examples of outfits that seem to merely be "similar enough to x to be represented by it", and lose potential character (and player immersion) as a result.

similar to ship variants, perhaps slightly different spec outfits that carry the same name - the Syndicate sells stuff that falls closer to the lower end of their quality threshold to distributers in far flung markets, or the Hai/Lovelace fusion generators, or the Deep has begun using a new material for their Laser lenses, and the old stock is still being cleared from outfitters carrying only their "basic" offerings, or the 'prototype' ebeams on the upper tier Marauder variants...just a couple of applications off the top of my head...

i haven't thought this idea out very thoroughly, so forgive me if there are potentially fatal flaws in it.

DingusShingleton:
Actually, there is precedent within the game already.... Modified blasters! 

I think it's a cool idea, although it risks becoming an uncontrollable leviathan of outfit variants that could overwhelm newer players and introduce a lot of complication into what is now a relatively straightfoward system. 

Although it would be really cool to load up a bunch of advanced weaponry and ship it to systems that don't have access to it. Lord of war style.

LocalGod79:
i considered that you could just make additional outfits that carry slightly different names...

you could look at the hyper/scram/jump as variants of the same outfit, as well...

i'm trying to think of ways to increase the depth while maintaining the existing simplicity - as i've said, this idea is only half-baked, and might just be more fluff...

regarding 'lord of war'; i've always thought that if you were selling outfits at venues that didn't normally carry them, they should carry a slightly higher price tag - rudimentary outfit merchantism.  that said, margins would need to be pretty small, given the base per-ton value of outfits in general...perhaps this is another issue entirely, though...opening now...

Wrzlprnft:
Am i getting your request right?
A way to implement two outfits with the same stats, same name but different sprites with a minimum of code like variants do?

LocalGod79:
that was the original idea...

perhaps in my initial expression, a brainstorm took hold of me and expressed more of a stat direction.

endless-sky:
The truth is that I was being lazy when I created the Hai, which is why they have so many outfits in common with humans. I suppose you could claim that a few of those outfits were Hai technology that humans copied and explain it that way, but really they should have gotten their own engines, generators, etc.

LocalGod79:
[...]

PR #5849, opened 2021-04-17
Outfit Variants

Rotnum:
**Feature:** This PR implements the feature request detailed and discussed on Discord and in (closed, for some reason) issue #927

## Feature Details
Outfit Variants!

Variant outfits inherit all attributes from their parents, excluding the name, description, license information, stuff related to thrust flares, and stuff related to jump visuals / sounds, and weapons.

For the most part variants are treated like perfectly normal outfits. In missions, however, a `require` will check not only for the provided outfit but all of its variants.

For reasons this PR also includes the ability to define display names independent of an outfit's actual in-data name.

## UI Screenshots
variant names 2 electric boogaloo

## Usage Examples
Variants are declared in the same way as ship variants:
```
outfit <outfit name> <variant name>
	# Normal outfit stuff.
```

By default, a variant displays the same name as its parent. To change this, use a `"display name"` attribute to declare a new name.
Outfits with a unique display name will also show their parent's name in parentheses below. To disable this, set a `"hide parent name"` attribute to 1.

In order to check for all of the variants of an outfit in a mission, you can use normal syntax:
```
mission "Cool mission"
	# Normal mission stuff
	on offer
		require "Some outfit with variants" 3
```

## Testing Done
I have created variants and made sure they are serialized on ships correctly.
I have verified they (and missions which make use of them) serialize correctly.

## Performance Impact
Missions which use the `require` node will iterate over all of the outfit's variants, if it has any.


source/Outfit.cpp
```diff
@@ -118,11 +118,25 @@ const vector<string> Outfit::CATEGORIES = {
...
+		
+		const Outfit *parent = GameData::Outfits().Get(node.Token(1));
+		
+		// Copy data over from the parent outfit.
+		category = parent->Category();
+		thumbnail = parent->Thumbnail();
+		cost = parent->Cost();
+		mass = parent->Mass();
+		attributes = parent->Attributes();
+	}
```
Amazinite, line 139: We can't guarantee that this information is available, as the parent might not be loaded by this point. (That is, the variant might be defined before the parent.) If this is to be done, then we need a FinishLoading function that is only called after everything is loaded, akin to what Ship.h/cpp has.

source/MissionAction.cpp
```diff
@@ -430,6 +430,14 @@ bool MissionAction::CanBeDone(const PlayerInfo &player, const shared_ptr<Ship> &
 				{
 					available += ship->Cargo().Get(it.first);
 					available += ship->OutfitCount(it.first);
+					
+					// Check for this outfit's variants
+					if (GameData::OutfitVariants().find(it.first) != GameData::OutfitVariants().end())
```
Amazinite, line 435: Read up on the [style guide](https://endless-sky.github.io/styleguide/styleguide.xml); use `if(`, not `if (`, and comments should be formatted like statements (i.e. add periods to the end of your comments).

source/Outfit.cpp
```diff
@@ -118,11 +118,25 @@ const vector<string> Outfit::CATEGORIES = {
 
 void Outfit::Load(const DataNode &node)
 {
-	if(node.Size() >= 2)
+	if(node.Size() == 2)
```
Amazinite, line 121: I would think that outfit variants would behave similarly to ship variants. That is, outfit variants should display with the same name as their parent. You'd then save the variant name to a separate string that is used when a ship saves its outfits or a mission saves its require and outfit nodes, just as ship variants do when a mission saves its give ship nodes.

source/GameData.cpp
```diff
@@ -1061,7 +1069,15 @@ void GameData::LoadFile(const string &path, bool debugMode)
...
 		else if(key == "outfit" && node.Size() >= 2)
-			outfits.Get(node.Token(1))->Load(node);
+		{
+			// Allow outfits to inherit attributes from parents
+			const string &name = node.Token((node.Size() > 2) ? 2 : 1);
+			outfits.Get(name)->Load(node);
+			
+			// If this outfit is a variant add it to the map of variants
+			if(node.Size() > 2)
+				outfitVariants[outfits.Get(node.Token(1))].emplace_back(outfits.Get(name));
```
Amazinite, line 1079: I'm not too keen on this method of making `require` work, as it's not as if we keep track of all ship variants in the same way. Not certain on how else it could be done at the moment, though.

Amazinite (commented):
Just taking a very quick look at this and commenting on the implementation.

MCOfficer:
You can merge master now to fix those pesky CI errors.

source/Outfit.cpp
```diff
@@ -244,6 +263,29 @@ void Outfit::Load(const DataNode &node)
...
+		
+	// Copy data over from the parent outfit.
+	if(!hasCategory)
+		category = parent->Category();
+	if(!hasThumbnail)
+		thumbnail = parent->Thumbnail();
+	if(!hasCost)
+		cost = parent->Cost();
+	if(!hasMass)
+		mass = parent->Mass();
```
Amazinite, line 278: These booleans are redundant with other methods you could use to check if these have been set. I've talked about looking to how ship variants are handled and emulating that here, and in that case there aren't any additional booleans checking if various attributes have been set or not.
https://github.com/endless-sky/endless-sky/blob/c9a79a3080fc7a52d1123385f5fdafc451fc73f3/source/Ship.cpp#L490-L500
```suggestion
	if(category.empty())
		category = parent->Category();
	if(!thumbnail)
		thumbnail = parent->Thumbnail();
	if(!cost)
		cost = parent->Cost();
	if(!mass)
		mass = parent->Mass();
```

source/Outfit.cpp
```diff
@@ -244,6 +263,29 @@ void Outfit::Load(const DataNode &node)
 
 
 
+void Outfit::FinishLoading()
+{
+	const Outfit *parent = GameData::Outfits().Get(parentName);
```
Amazinite, line 268: This outfit might not have a parent, in which case you're grabbing a null object and referencing it if this outfit for some reason is missing any of the following attributes. This can actually happen in vanilla, as H2H weapons have no mass, in which case you're referencing a non-existent parent outfit for its mass. I imagine that's not good.

You should be checking if a parent exists before referencing it. Again, look to how ship does it.
https://github.com/endless-sky/endless-sky/blob/c9a79a3080fc7a52d1123385f5fdafc451fc73f3/source/Ship.cpp#L476-L486

source/Outfit.cpp
```diff
@@ -244,6 +263,29 @@ void Outfit::Load(const DataNode &node)
...
+	if(!hasThumbnail)
+		thumbnail = parent->Thumbnail();
+	if(!hasCost)
+		cost = parent->Cost();
+	if(!hasMass)
+		mass = parent->Mass();
+	
+	for(const auto &it : parent->Attributes())
+	{
+		if(!attributes.Has(it.first))
```
Amazinite, line 282: You shouldn't be adding new functions to the dictionary class. Dictionary::Get can be used exactly like how you're using this Dictionary::Has function that you created.
```suggestion
		if(!attributes.Get(it.first))
```
And looking to ship once more, ship only overwrites any attribute if the variant has no attributes. Something similar should be done here where only if attributes is empty do the parent attributes get used (and in that case you can simply assign the parent attribute object to the variant attribute object instead of iterating through all the attributes).
https://github.com/endless-sky/endless-sky/blob/c9a79a3080fc7a52d1123385f5fdafc451fc73f3/source/Ship.cpp#L496-L497

source/Outfit.cpp
```diff
@@ -118,17 +118,27 @@ const vector<string> Outfit::CATEGORIES = {
...
 		name = node.Token(1);
 		pluralName = name + 's';
 	}
+	else if(node.Size() >= 3)
+	{
+		name = node.Token(2);
+		parentName = node.Token(1);
+		pluralName = name + 's';
+		isVariant = true;
+	}
```
Amazinite, line 132: This isVariant boolean is also unnecessary. `if(!parentName.empty())` and `if(isVariant)` are functionally equivalent without the need of a new variable. And as I mentioned in my previous review, outfit variants should share the display name of their parent, so this should be something more like this:
```suggestion
	if(node.Size() >= 2)
	{
		name = node.Token(1);
		pluralName = name + 's';
		if(node.Size() >= 3)
			variantName = node.Token(2);
	}
```
This will require other changes elsewhere, such as the creation of a TrueName function that returns variantName if it exists, otherwise returning name. This function would then be used when a ship is saving its outfits so that we don't accidentally convert a variant outfit to its parent.

Amazinite (changes requested):

source/Outfit.cpp
tehhowch: They do allow setting the respective items to values that implicitly convert to `false`

source/Outfit.cpp
tehhowch: I would prefer a `Get` is called at some point to facilitate reference checking. If `Get` is used, it will never return a null pointer, but it may return a "garbage" pointer to an Outfit that has not been defined.

(Even if `Find` is used and the returned pointer is not null, the outfit may not yet have been loaded, so a defined-ness check is still required.)

source/Outfit.cpp
tehhowch: The intent here is likely to facilitate "intentional zero" values being supported.

tehhowch:
Not diving into a review at the moment.

A few notes about this approach:
- it is not compatible with modifying outfits via events (which is functionality I don't want us to add) as the attribute copy is done during the game's initial loading phase, rather than an indirection followed at time of use. (It could be made compatible by maintaining an inheritance tree and propagating updates to leaf nodes, but this will be gross.)
- it (currently) can't handle variant variants well, and should likely explicitly forbid such behavior. (If that rabbit hole is desirable and useful, it is backwards compatible to enable that later.)
- I don't know that I agree with Amazinite about how a variant should be represented in the UI. _Probably_ it should, but maybe it is sufficient for the variants to appear next to each other in e.g. the outfitter.

This functionality may be worth putting together into an RFC that weighs several implementations against what the functionality is intended to solve

source/Outfit.cpp
```diff
@@ -118,17 +118,27 @@ const vector<string> Outfit::CATEGORIES = {
...
-	if(node.Size() >= 2)
+	if(node.Size() == 2)
 	{
 		name = node.Token(1);
 		pluralName = name + 's';
 	}
+	else if(node.Size() >= 3)
+	{
+		name = node.Token(2);
+		parentName = node.Token(1);
```
tehhowch, line 129: One review comment: if tracking this, use a `GameData::Outfits().Get` to store the pointer, rather than a string that is later used to get the pointer.

source/Outfit.cpp
Rotnum: That's correct.

ravenshining:
> * I don't know that I agree with Amazinite about how a variant should be represented in the UI. _Probably_ it should, but maybe it is sufficient for the variants to appear next to each other in e.g. the outfitter.

I don't agree myself. I think it would be nice to see variant names in the outfitter... though I can see a case for not showing them in scans.  Like, when you scan a ship, you can only tell that a ship has *a* jump drive, but when you take it apart (at the outfitters), then you can see that it has a Korath jump drive or something.

Zitchas:
> I don't agree myself. I think it would be nice to see variant names in the outfitter... though I can see a case for not showing them in scans. Like, when you scan a ship, you can only tell that a ship has a jump drive, but when you take it apart (at the outfitters), then you can see that it has a Korath jump drive or something.

This makes me wish that we could attach some kind of detail level to things. Then we'd have "normal" scans that just show that a ship has, for example, a hyperdrive. Then we could have "scientific analysis" scans that give the variant name instead. Perhaps we could have outfits that individually have the inscrutable tag which don't show up at all on normal scans, and show up as inscrutable on the scientific scans, and actually reveal their name when scanned with an "advanced" scanner. 

(names are just quick placeholders. It basically amounts to normal tech scanner, fancy tech scanner, and super advanced tech scanner)

Pointedstick:
@Amazinite, are your code concerns addressed now?

Functionality-wise, I'm not sold yet and need some convincing. :) This seems like a lot of code for something that I can't see a clear benefit to. Why not just create new outfits? We already have things like the Modified Blaster which is a variant of the Blaster, and (on the ship side) the Modified Argosy which is a variant of the Argosy. What's the win in defining things like these using a new variant system compared to the simpler status quo of just making a new outfit or ship? Sell me on this!

Zitchas:
Reason pretty much boils down to meeting mission requirements,without needing a lot of OR statements for every alternative variant. I think.

Pointedstick:
If that's it, I don't really see the win, TBH. Do we have any actual examples where this is a problem?

Zitchas:
The infamous HD swap missions, since a lot of people want racial HD (and other universal outfits too). But it also means people have to be careful using plugin outfits that may block certain missions. So it is a compatibility thing too.

Amazonite keeps selling people "we can have racial HD once someone gets outfit variants working nicely."

Pointedstick:
Oh, so it's just so that each race's hyperdrive can look different? If so, that suggests a far simpler approach: let outfits' sprites and descriptions be overridden when offered by different outfitters.

Zitchas:
No, it's not just that - that's just the very first use-case we're likely to see. I'm sure people would like to see variations on them in actual performance, too.

Actually, the first use case would be the scram drive, which is technically just a hyperdrive with some extra motion tolerance. 

Regular HD:
- Scram (bigger but has motion tolerance)
- Fuel efficiency (bigger but uses less fuel per jump)
- Combination of the above?
- Some T4 race that has hybrid HD+JD
- What about smaller but uses 3x as much fuel?
- Energy per jump instead of (or in addition to) fuel?

Just spit balling ideas here. But anyway, the main point is to be able to have whole families of outfits that are basically their race's flavor of said outfit. Probably, for the most part, pretty minor tweaks - just enough to feel different in addition to the appearance and description - but still have them all count as 1 <insert main outfit here>. 

It also fits the paradign of ships. We do that for ships, so it makes sense to be able to do it for outfits too.

Amazinite:
> are your code concerns addressed now?

Not really. I’ve repeatedly suggested copying how ship variants are handled yet it looks like some things for outfit variants here are still done in the reverse of how ship variants do it or in a separate way entirely. If this way is somehow better then that should be explained, but otherwise for the sake of consistency I think that ship and outfit variants should be handled similarly internally, which would make it easier to understand the code for both instead of having to learn one system for ship variants and another for outfit variants.

source/Outfit.cpp
```diff
@@ -261,6 +302,13 @@ const string &Outfit::Name() const
 
 
 
+const string &Outfit::DisplayName() const
```
quyykk, line 305: The convention is `Name()` returns the display name and `TrueName()` returns the name in the data files. We should stay consistent :)

petervdmeer:
Closing this PR; It looks abandoned and I see quite a bit of open work to get this finalized;
- I see merge-conflicts and feedback from tehhowch, Amazinite, Pointedstick and quyykk that hasn't been addressed in more than 8 months. Some of the feedback is quite a bit more than trivial text-changes.
- Nobody made a comment on testing yet, but a niche feature like this would likely also need some automated tests to ensure that the functionality works correctly, and keeps working correctly.

@Rotnum; thanks for contributing and please feel free to re-open this PR (or create another PR) when you want to contribute again.

PR #5852, opened 2021-04-18
Create Scrappy variant for the Heliarch Interdictor

Arachi-Lover:
----------------------
**Content (Variant)**

## Summary
I noticed the Heliarch Interdictor was the only Heliarch ship (with the exception of the Rover, which has it as the stock variant) to lack a Scrappy (solely Bombardment Cannons and Turrets) variant, so I made one for it.
This also adds the variant to the Heliarch fleet spawns.

### DIFF
--- a/data/coalition/coalition ships.txt
+++ b/data/coalition/coalition ships.txt
@@ -660,6 +660,28 @@ ship "Heliarch Interdictor" "Heliarch Interdictor (Missile)"
 		"Large Steering Module"
 		"Hyperdrive"
 
+ship "Heliarch Interdictor" "Heliarch Interdictor (Scrappy)"
+	outfits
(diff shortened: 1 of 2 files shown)


Zitchas:
I haven't tested it out  yet, but at least on the surface this looks like a solid addition.

Arachi-Lover:
I gave the build a quick test in Omnis, the old spin around while firing all weapons. There weren't any energy problems, and heat sat at about 80% at the high point. Of course, that doesn't include the energy costs should the ship's regen systems be active, or the heat, but while I don't recall the exact values, I do remember the sum of all heat values was some hundreds below max heat, and total energy production was just a little bit over total energy costs.
That case is the same for many Heliarch variants, so I don't expect this one to run into any particular problems.

PR #5858, opened 2021-04-18
fix(content): Bar off Wanderer space landing anywhere after killing Arfectas

Terin:
**Content (Mission)**

## Summary
Currently, the player can destroy or capture one or both Arfectas spawned by "Wanderers Ap'arak 3: Pug". Despite this being a terrible act to the Wanderers, due to the way the mission is structure, you can only be penalized if you land on the planet where the mission is spawned, which is out of the way and only has one required landing in the rest of the Wanderers campaign.

From a suggestion from @beccabunny, this PR restructures that mission into two `npc save` missions. Upon both being failed, a third mission will be prompted on landing anywhere in Wanderer space, giving you the reputation hit and kicking you out. This is backwards-compatible with current saves, requiring both missions to be failed, and the second mission only activating if you haven't done the follow-up mission.

Possible tweaks to this PR could be having the mission fail if either of the Arfectas die (although the impression I was given was we wanted to make it both), or having the mission be received as a message anywhere in space, to ensure you can't run from the reputation hit. I believe that keeping the player from landing anywhere in Wanderer space is pressure enough, though.

## Testing Done
Used the included save file to land on Varu Tev'kei, spawning both Arfectas. Destroyed one and landed, and noticed no change. Destroyed the second and landed on a random Wanderer planet, which triggered the mission to make them hostile.

Also used an old save with active Arfectas, and ensured that a third did not spawn.

## Save File
This save file can be used to play through the new mission content. Jump to Ap'arak and land to spawn the Arfectas, which should be hostile:
[Arfecta Test.txt](https://github.com/endless-sky/endless-sky/files/6331600/Arfecta.Test.txt)

### DIFF
--- a/data/wanderer/wanderers start.txt
+++ b/data/wanderer/wanderers start.txt
@@ -2751,7 +2751,6 @@ mission "Wanderers Ap'arak 3: Pug 2"
 mission "Wanderers Pug Hostile"
 	landing
 	invisible
-	deadline 1
 	source
 		government "Wanderer"
 	to offer
@@ -2760,7 +2759,7 @@ mission "Wanderers Pug Hostile"
 	on offer
 		conversation
 			`Having defeated the Pug ships guarding Wanderer space, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
-				launch
+				flee
 		"reputation: Wanderer" <?= -1
 
 mission "Wanderers Ap'arak 3"
--- a/data/wanderer/wanderers start.txt
+++ b/data/wanderer/wanderers start.txt
@@ -2757,11 +2757,11 @@ mission "Wanderers Pug Hostile"
 	source
 		government "Wanderer"
 	to offer
-		has "Wanderers Ap'arak 3: Pug: failed"
+		has "Wanderers Ap'arak 3: Pug 1: failed"
 		has "Wanderers Ap'arak 3: Pug 2: failed"
 	on offer
 		conversation
-			`Having defeated the Pug ships guarding Wanderer space, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
+			`You receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
 				flee
 		set "Wanderers Ap'arak 3: Pug: done"
 		"reputation: Wanderer" <?= -1
--- a/data/wanderer/wanderers start.txt
+++ b/data/wanderer/wanderers start.txt
@@ -2727,13 +2727,41 @@ mission "Wanderers Ap'arak 3: Pug"
 	source "Varu Tev'kei"
 	to offer
 		has "Wanderers Ap'arak 2: done"
-	npc kill
+	npc save
 		personality staying unconstrained uninterested
 		government "Pug (Wanderer)"
 		ship "Pug Arfecta" "Hork Ekekek Ptui"
+	to complete
+		never
+
+mission "Wanderers Ap'arak 3: Pug 2"
+	landing
+	invisible
+	source "Varu Tev'kei"
+	to offer
+		has "Wanderers Ap'arak 2: done"
+		not "Wanderers Ap'arak 3: done"
+	npc save
+		personality staying unconstrained uninterested
+		government "Pug (Wanderer)"
 		ship "Pug Arfecta" "Graak Sput Kchoo"
-	on complete
-		dialog `Shortly after defeating the Pug ships, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
+	to complete
+		never
+
+mission "Wanderers Pug Hostile"
+	priority
+	landing
+	invisible
+	deadline 1
+	source
+		government "Wanderer"
+	to offer
+		has "Wanderers Ap'arak 3: Pug: failed"
+		has "Wanderers Ap'arak 3: Pug 2: failed"
+	on offer
+		conversation
+			`Having defeated the Pug ships guarding Wanderer space, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
+				launch
 		"reputation: Wanderer" <?= -1
 
 mission "Wanderers Ap'arak 3"
--- a/data/wanderer/wanderers start.txt
+++ b/data/wanderer/wanderers start.txt
@@ -2749,7 +2749,6 @@ mission "Wanderers Ap'arak 3: Pug 2"
 		never
 
 mission "Wanderers Pug Hostile"
-	priority
 	landing
 	invisible
 	deadline 1
--- a/data/wanderer/wanderers start.txt
+++ b/data/wanderer/wanderers start.txt
@@ -2721,6 +2721,8 @@ mission "Wanderers Ap'arak 2"
 
 
 
+# TODO: When "complete anywhere" functionality is available, 
+# condense the following three missions into a single mission
 mission "Wanderers Ap'arak 3: Pug"
 	landing
 	invisible
--- a/data/wanderer/wanderers start.txt
+++ b/data/wanderer/wanderers start.txt
@@ -2723,12 +2723,13 @@ mission "Wanderers Ap'arak 2"
 
 # TODO: When "complete anywhere" functionality is available, 
 # condense the following three missions into a single mission
-mission "Wanderers Ap'arak 3: Pug"
+mission "Wanderers Ap'arak 3: Pug 1"
 	landing
 	invisible
 	source "Varu Tev'kei"
 	to offer
 		has "Wanderers Ap'arak 2: done"
+		not "Wanderers Ap'arak 3: offered"
 	npc save
 		personality staying unconstrained uninterested
 		government "Pug (Wanderer)"
@@ -2742,7 +2743,7 @@ mission "Wanderers Ap'arak 3: Pug 2"
 	source "Varu Tev'kei"
 	to offer
 		has "Wanderers Ap'arak 2: done"
-		not "Wanderers Ap'arak 3: done"
+		not "Wanderers Ap'arak 3: offered"
 	npc save
 		personality staying unconstrained uninterested
 		government "Pug (Wanderer)"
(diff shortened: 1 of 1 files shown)


data/wanderer/wanderers start.txt
```diff
@@ -2727,13 +2727,41 @@ mission "Wanderers Ap'arak 3: Pug"
...
+		personality staying unconstrained uninterested
+		government "Pug (Wanderer)"
 		ship "Pug Arfecta" "Graak Sput Kchoo"
-	on complete
-		dialog `Shortly after defeating the Pug ships, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
+	to complete
+		never
+
+mission "Wanderers Pug Hostile"
+	priority
```
Amazinite, line 2752: `priority` is redundant with `landing`.

data/wanderer/wanderers start.txt
```diff
@@ -2727,13 +2727,41 @@ mission "Wanderers Ap'arak 3: Pug"
...
+	deadline 1
+	source
+		government "Wanderer"
+	to offer
+		has "Wanderers Ap'arak 3: Pug: failed"
+		has "Wanderers Ap'arak 3: Pug 2: failed"
+	on offer
+		conversation
+			`Having defeated the Pug ships guarding Wanderer space, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
+				launch
```
Amazinite, line 2764: The `launch` keyword is take off and accept. Use `flee` for take off and decline behavior, which removes the need for the deadline.

Amazinite (changes requested):

data/wanderer/wanderers start.txt
```diff
@@ -2727,13 +2727,39 @@ mission "Wanderers Ap'arak 3: Pug"
...
+	landing
+	invisible
+	source
+		government "Wanderer"
+	to offer
+		has "Wanderers Ap'arak 3: Pug: failed"
+		has "Wanderers Ap'arak 3: Pug 2: failed"
+	on offer
+		conversation
+			`Having defeated the Pug ships guarding Wanderer space, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
```
tehhowch, line 2761: The opening here ("Having defeated..., you receive") is awkward and should be rephrased. I'd argue the text flows better if we drop the lead-in entirely, so we can explore a text structure that makes the message seem more like a formal letter from a Wanderer ambassador.

tehhowch (commented):
I think we had not gotten around to cleaning up this mission sequence because it works best with "complete-anywhere" semantics, which are still a future functionality. Ideally, the player would defeat both Arfectas and then promptly suffer the consequences. They already can't land in-system after the attack, so on the next jump the penalty mission with the formal missive would offer and the player would be locked out entirely.

So, probably best to codify that with a to-do comment on this last mission.

data/wanderer/wanderers start.txt
```diff
@@ -2721,19 +2721,47 @@ mission "Wanderers Ap'arak 2"
...
+	to complete
+		never
+
+mission "Wanderers Ap'arak 3: Pug 2"
+	landing
+	invisible
+	source "Varu Tev'kei"
+	to offer
+		has "Wanderers Ap'arak 2: done"
+		not "Wanderers Ap'arak 3: done"
```
tehhowch, line 2745: I think this is incorrect for existing pilots, and will spawn an additional Arfectas.

It should be an "offered" check instead, and the first mission should be both renamed and also given this check. To maintain compatibility, the penalty mission needs to set this particular "done" condition.

data/wanderer/wanderers start.txt
Terin: Part of that line is because the original is "Shortly after defeating the Pug ships, you receive a message...", which sounded fine. The other part is someone could jump away from the system, or even leave Wanderer space entirely before coming back months later. That preamble helps remind a player of what they did from the jump, and dropping it entirely feels really short.

data/wanderer/wanderers start.txt
Terin: It seems like it would only spawn an additional Arfecta if someone had a save between Ap'arak 2 and 3, and I tested on my 0.9.13 pilot with no additional Arfecta spawns. I'll tighten up the requirements as much as possible, but it seems unlikely players will have a specific save there unless they're actively following this PR.

data/wanderer/wanderers start.txt
```diff
@@ -2721,19 +2721,49 @@ mission "Wanderers Ap'arak 2"
...
+	to complete
+		never
+
+mission "Wanderers Pug Hostile"
+	landing
+	invisible
+	source
+		government "Wanderer"
+	to offer
+		has "Wanderers Ap'arak 3: Pug: failed"
```
tehhowch, line 2760:
```suggestion
		has "Wanderers Ap'arak 3: Pug 1: failed"
```

data/wanderer/wanderers start.txt
tehhowch: Honestly, I had completely missed that this was for a different mission than the one it was replacing, which was designed to generally not be "done."

The change to use "offered" was still necessary, but indeed much less impactful than my original perception.

data/wanderer/wanderers start.txt
tehhowch: Yeah, just identifying some awkward phrasing. I didn't have any better alternatives, which is why I suggested perhaps changing the format to be totally different.

2phunkey4u:
Story-wise, wouldn't the elimination of the two Arfectas even have the consequence of Wanderer space being completely overrun by the Unfettered Hai eventually?

data/wanderer/wanderers start.txt
```diff
@@ -2721,19 +2721,49 @@ mission "Wanderers Ap'arak 2"
...
+		government "Wanderer"
+	to offer
+		has "Wanderers Ap'arak 3: Pug 1: failed"
+		has "Wanderers Ap'arak 3: Pug 2: failed"
+	on offer
+		conversation
+			`You receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
+				flee
+		set "Wanderers Ap'arak 3: Pug: done"
 		"reputation: Wanderer" <?= -1
```
Amazinite, line 2767: Just shuffling the action order for style purposes.
```suggestion
		set "Wanderers Ap'arak 3: Pug: done"
		"reputation: Wanderer" <?= -1
		conversation
			`You receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
				flee
```

data/wanderer/wanderers start.txt
```diff
@@ -2721,19 +2721,49 @@ mission "Wanderers Ap'arak 2"
...
+		government "Pug (Wanderer)"
 		ship "Pug Arfecta" "Graak Sput Kchoo"
-	on complete
-		dialog `Shortly after defeating the Pug ships, you receive a message from Sobari Tele'ek: "We are [aware? incredulous?] of reports that you have [exterminated, eradicated] our Keepers in the Ap'arak system. This is unacceptable! From now on, you are no longer welcome in our [territory, space]."`
+	to complete
+		never
+
+mission "Wanderers Pug Hostile"
+	landing
+	invisible
```
Amazinite, line 2756: Redundant with the mission never actually accepting.
```suggestion
```

data/wanderer/wanderers start.txt
```diff
@@ -2721,19 +2721,49 @@ mission "Wanderers Ap'arak 2"
...
+# TODO: When "complete anywhere" functionality is available, 
+# condense the following three missions into a single mission
+mission "Wanderers Ap'arak 3: Pug 1"
 	landing
 	invisible
 	source "Varu Tev'kei"
 	to offer
 		has "Wanderers Ap'arak 2: done"
-	npc kill
+		not "Wanderers Ap'arak 3: offered"
```
Amazinite, line 2732: I think it'd technically be safer to not offer if the old mission was offered.
```suggestion
		not "Wanderers Ap'arak 3: Pug: offered"
```

data/wanderer/wanderers start.txt
```diff
@@ -2721,19 +2721,49 @@ mission "Wanderers Ap'arak 2"
...
+	to complete
+		never
+
+mission "Wanderers Ap'arak 3: Pug 2"
+	landing
+	invisible
+	source "Varu Tev'kei"
+	to offer
+		has "Wanderers Ap'arak 2: done"
+		not "Wanderers Ap'arak 3: offered"
```
Amazinite, line 2746:
```suggestion
		not "Wanderers Ap'arak 3: Pug: offered"
```

Amazinite (changes requested):

Amazinite:
> Story-wise, wouldn't the elimination of the two Arfectas even have the consequence of Wanderer space being completely overrun by the Unfettered Hai eventually? They were rearguards after all, IIRC.

Well story-wise this makes it so that eliminating either Arfecta cuts you out of the Wanderer campaign, so not like we'd need to consider the consequences of your actions that the Wanderers face when you can't experience them.

2phunkey4u:
> 
> Well story-wise this makes it so that eliminating either Arfecta cuts you out of the Wanderer campaign, so not like we'd need to consider the consequences of your actions that the Wanderers face when you can't experience them.

Well, of course the player could choose to eliminate the Arfectas after completing the Wanderer campaign. Either way, I would still argue that eliminating the Arfectas could - in the long-run - lead to Wanderer space (the space before the Eye, to be precise) becoming Unfettered space, as far as the Wanderer story implies at the point of the Pug intervention. This could even lead to a chain reaction of domain shifts: Wanderers completely migrate into Kor Sestor space, Unfettered occupy all of Wanderer space (but not beyond the Eye), maybe ceding Northern Hai space in return (maybe due to otherwise overstretching their frontiers or something like that).

EjoThims:
> > Well story-wise this makes it so that eliminating either Arfecta cuts you out of the Wanderer campaign, so not like we'd need to consider the consequences of your actions that the Wanderers face when you can't experience them.
> 
> Well, of course the player could choose to eliminate the Arfectas after completing the Wanderer campaign. Either way, I would still argue that eliminating the Arfectas could - in the long-run - lead to Wanderer space (the space before the Eye, to be precise) becoming Unfettered space, as far as the Wanderer story implies at the point of the Pug intervention. This could even lead to a chain reaction of domain shifts: Wanderers completely migrate into Kor Sestor space, Unfettered occupy all of Wanderer space (but not beyond the Eye), maybe ceding Northern Hai space in return (maybe due to otherwise overstretching their frontiers or something like that).

I could see this alternate path being the play out in a short Unfettered campaign with the double Arfecta fight as like the "boss fight" climax.

Terin:
It could be a neat little easter egg of sorts if someone does that, although there are tons of places where a player can fail a mission that would result in drastically different outcomes for the universe, so it's really easy to get down that rabbit hole of what-ifs. Maybe once we know how the Wanderers campaign goes, it could be something added in later.

Amazinite (approved):

PR #5860, opened 2021-04-18
Various typo fixes

AraCaputDraco:
## Fix Details

This PR fixes a small number of typos. Changes include switching an occurrence of "you" inside the logbook to "me" as I feel this would be more appropriate. However, it might be better to rephrase that sentence completely, because no other log entry I know of addresses the player directly.

This is my first contribution on GitHub, so any advice is always appreciated.


tehhowch (approved):
Lgtm! Thanks for the PR, and welcome

PR #5868, opened 2021-04-20
Adjust description of the Remnant: Tech Retrieval missions

AraCaputDraco:
-----------------------
**Bugfix:** This PR addresses no established issue.

I noticed that the description of the Remnant: Tech Retrieval missions shows the payment even when no exact amount has been mentioned within the preceding dialogue. Also the mission about Electron Beams did not tell the player where to deliver them.

## Fix Details
Removed exact payment values when not specified before. Also added a sentence about the destination for the Electron Beams to the conversation. I'm open for feedback on the exact wording though.

## Testing Done
Played the mission regarding Electron Beams.


MasterOfGrey:
Part of me feels like the better solution might be to have the preceding conversations actually specify the amount @Zitchas .
But if that would mess up the feel of those conversations then yes I think this is good.

Zitchas (approved):
Good catch. These improve the feel and fit of these missions. Thanks!

PR #5871, opened 2021-04-20
3 new bounty missions reupload

samoja12:
Added 3 new bounty missions. This is a reupload from a few days ago, i accidentally edited the master so i had to reset it.

----------------------
**Content (Jobs)**

## Summary
{{added 3 new bounty missions with lower level rewards and requirements, and lower level ships.}}

### DIFF
--- a/data/human/jobs.txt
+++ b/data/human/jobs.txt
@@ -2575,6 +2575,143 @@ mission "Recycle garbage"
 		payment
 		payment 20000
 
+mission "Bounty Hunting (Very Tiny)"
+	name "Wanted corsair near <system>"
+	description "A small pirate fighter named the <npc> has been attacking merchants near the <system> system. Destroy it and return to <planet> for payment (<payment>)."
+	repeat
+	job
+	to offer
+		"combat rating" > 50
+		random < 40
+	source
+		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Independent"
+		attributes "rim" "south" "north" "dirt belt" "core" "frontier"
+	npc kill
+		government "Bounty"
+		personality heroic staying nemesis target
+		system
+			distance 1 2
+		fleet
+			names "pirate"
+			variant
+				"Sparrow"
+			variant
+				"Wasp"
+			variant
+				"Wasp (Gatling)"
+			variant
+				"Wasp (Javelin)"
+			variant
+				"Wasp (Meteor)"
+			variant
+				"Wasp (Plasma)"
+			variant
+				"Wasp (Proton)"
+			
+		dialog phrase "generic hunted bounty eliminated dialog"
+	on visit
+		dialog phrase "generic bounty hunting on visit"
+	on complete
+		payment 20000
+		dialog phrase "generic bounty hunting payment dialog"
+
+mission "Bounty Hunting (Tiny)"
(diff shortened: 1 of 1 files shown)


Terin:
Reupload of #5848, specifically.

ravenshining:
And as with the closed version, I would like to again say that
a) These missions should be available to players who own an appropriate ship but do not have combat rep, and
b) These missions should have a maximum combat rep to cut down on job board spam

samoja12:
> 
> 
> And as with the closed version, I would like to again say that
> a) These missions should be available to players who own an appropriate ship but do not have combat rep, and
> b) These missions should have a maximum combat rep to cut down on job board spam

I can't see why we should restrict this to low level players when we don't do the same with transport missions. The same thing about ship requirements, higher level bounties are not ship restricted nor are escort missions.This missions are supposed to be in line with existing missions, i have no plans to introduce arbitrary limitations.

ravenshining:
a) Again, is *not* a restriction - it makes the mission *more* available to low-level players.

b) Is in line with existing combat jobs - the smallest escort missions do not appear once the player has earned some reputation.

samoja12:
> 
> 
> a) Again, is _not_ a restriction - it makes the mission _more_ available to low-level players.
> 
> b) Is in line with existing combat jobs - the smallest escort missions do not appear once the player has earned some reputation.

It is a restriction because what if i want to go back to playing in a light warship at any point? In this scenario i no longer get access to missions that are appropriate for me. Besides maybe i don't upgrade my warship and invest all my earnings in a trading fleet. In this case my CR rises but i am still flying the same ship.

ravenshining:
That might be a valid point against b) - it would be nice if we could tailor available missions to the player's current fleet or resources - but I think you are completely missing what I'm suggesting with point a)

The way you have it, the missions do not trigger until you have reached a minimum combat rating. I am suggesting that if the player has an interceptor or warship, they should be able to skip the rep requirement.  Take the player with 0 rep and a sparrow - your missions would not appear for them as you have written them, because they do not have enough reputation, I am suggesting that by virtue of owning a sparrow they should be eligible for the tiny bounty, despite their lack of reputation.

samoja12:
> 
> 
> That might be a valid point against b) - it would be nice if we could tailor available missions to the player's current fleet or resources - but I think you are completely missing what I'm suggesting with point a)
> 
> The way you have it, the missions do not trigger until you have reached a minimum combat rating. I am suggesting that if the player has an interceptor or warship, they should be able to skip the rep requirement. Take the player with 0 rep and a sparrow - your missions would not appear for them as you have written them, because they do not have enough reputation, I am suggesting that by virtue of owning a sparrow they should be eligible for the tiny bounty, despite their lack of reputation.

Fair enough, but i don't think we can use else/if statements, i was thinking of removing or lowering the first mission requirement, but i don't think i can do it so it's either a rep or a ship type, if you know how to do it then by all means i would love to do that.

ravenshining:
I gave you an example in the PR you closed, it just needs an or statement.

Pointedstick (changes requested):
These missions should have a maximum combat rating too, or else they'll always be visible even once the player is flying around in a fleet of heavy warships.

PR #5872, opened 2021-04-20
Creates  new Sparrow variants and adds them to fleets

samoja12:
----------------------
Adds 8 new Sparrow variants and adds them to merchant and pirate fleets across the galaxy. I made an effort to maintain the overall number of Sparrows consistent. For pirates it's slightly increased but for merchants it's the same, i skipped the fleets where the spawn was low from the beginning. 

Note: If both of my pull requests happen to be merged i can add this variants to the smallest level bounty.

### DIFF
--- a/data/human/fleets.txt
+++ b/data/human/fleets.txt
@@ -27,26 +27,154 @@ fleet "Small Southern Merchants"
 		"Clipper (Heavy)"
 	variant 10
 		"Clipper (Speedy)"
-	variant 10
+	variant 2
 		"Star Barge" 2
 		"Sparrow"
-	variant 10
+	variant 1
+		"Star Barge" 2
+		"Sparrow (Blaster)"
+	variant 1
+		"Star Barge" 2
+		"Sparrow (Modified)"
+	variant 1
+		"Star Barge" 2
+		"Sparrow (Meteor)"
+	variant 1
+		"Star Barge" 2
+		"Sparrow (Sidewinder)"
+	variant 1
+		"Star Barge" 2
+		"Sparrow (Heavy_laser)"
+	variant 1
+		"Star Barge" 2
+		"Sparrow (Gatling)"
+	variant 1
+		"Star Barge" 2
+		"Sparrow (Rocket)"
+	variant 1
+		"Star Barge" 2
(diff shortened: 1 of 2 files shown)


Amazinite:
Why?

samoja12:
> 
> 
> Why?

There are currently no Sparrow variants, i noticed that when i was making the 3 missions, i just wanted to introduce a bit of diversity into the Sparrow lineup to make things more interesting in the early game, especially for the Sparrow start. My overall goal with both of this PRs is to make a combat start more viable.

Pointedstick:
This seems sane to me. More diversity in NPC ship loadouts only serves to make the universe seem richer and more alive.

Looks like this needs a merge conflict resolved, and also the CI has found a formatting error. Can you fix those? Also have you tested the Rocket Sparrow variant to make sure that it doesn't blow itself up most of the time?

Zitchas:
Also worth noting that the recently merged weapon pods gives you a lot of new options for outfit sets.

Amazinite:
> More diversity in NPC ship loadouts only serves to make the universe seem richer and more alive.

Seeing the same ship not always have the same loadout certainly helps for immersion, but I wouldn’t say that this many variants really has the same effect. I feel like just a few variants for each ship like we have now suffices, but at a certain point adding more variants just becomes bloat as you probably don’t really perceive the difference between a ship with 3 variants or a ship with 10, and I feel like this PR reaches that point. Do we really need 10 different versions of the Sparrow to be able to appear, accounting for half a thousand new lines in the fleet file? Then just think about how massive this file would be if we aimed to have similar numbers of variants for other ships.

At the very least I don’t think that a change like this makes sense without a mechanic like #4704, which could help to massively cut down on the number of new lines necessary for increasing the number of variants a ship has that get spawned.

Zitchas:
I'm inclined to agree with Amazinite. Three variants of a ship are enough for "general feel." Others can be added as needed, of course. But probably aren't really needed for general atmosphere.

Pointedstick:
Counterpoint: the length of the file is irrelevant to the player, and only of marginal interest to the developer (I say this as a paid developer).

But OK, if this isn't popular enough to merge, let's close the PR.

Issue #5830, opened 2021-04-06
Quarg Wardragon Classification Issue


McloughlinGuy:
**Is your feature request related to a problem? Please describe.**
The Quarg Wardragon is classified as a heavy warship, despite it's description saying otherwise.

**Related Issue Links**
 #756 is similar, surprising they didn't notice this while correcting the previous error.

**Describe the solution you'd like**
Change it's classification to light warship.

**Describe alternatives you've considered**
Or you could change the description to say heavy warship.

**Additional context**
I had a text file which remedied the issue, but it wouldn't load.

McloughlinGuy:
I am addressing this in a pull request (I don’t know how to link them) so I am closing it.

MCOfficer:
You already have

![image](https://user-images.githubusercontent.com/22377202/115955358-41560a80-a4f6-11eb-9fde-413cf633875f.png)


![image](https://user-images.githubusercontent.com/22377202/115955356-3d29ed00-a4f6-11eb-9bc2-232ad99c4ed6.png)

Ideally you'd want to keep the issue open until the fix is merged, because in theory your fix could be rejected and the issue could go back to being on hold.

McloughlinGuy:
Ok.

PR #5880, opened 2021-04-22
Quarg Wardragon Classification Fix

McloughlinGuy:
Fixes the error in which the quarg wardragon is classified as a heavy warship

-----------------------
**Bugfix:** This PR addresses issue #5830

## Fix Details
I swapped out the word heavy for light in the classification.

## Testing Done
I modified the sales file so that the basic ships category included the wardragon. I then chekced to see if my fix worked, and it did work.
## Save File
[Blue Susatonia.txt](https://github.com/endless-sky/endless-sky/files/6361230/Blue.Susatonia.txt)

-----------------------
**Feature:** This PR implements the feature request detailed and discussed in issue #{{insert number}}

## Feature Details
It changes the classification.

## UI Screenshots
N/A (won’t load)

## Usage Examples
Currently N/A

## Testing Done
I added the Quarg Wardragon to the sales file under basic ships, so I could see if the change worked. It did.

## Performance Impact
N/A


McloughlinGuy:
note: I forgot to put a title

MCOfficer:
You can edit titles
image

McloughlinGuy:
Thanks

ravenshining:
-`quarg ships.txt` needs to live in the `data/quarg/` folder, you have it in the root folder which results in this PR making two copies of the file.
-  I didn't see any discussion on #5830 as to where to put it.  The Wardragon seems too heavy and too full of turrets to class as a light warship to me, personally I'd go for a compromise solution and change both title *and* description to class it as a medium warship.
- If the Wardragon is downcategorised, the Skylark should be, too, as they are nearly identical ships

Arachi-Lover:
The thumbs down is in reaction to the second point only.
The Quarg are going to be rebalanced at some point, and as I see it the changes made in # 3665 bring down the Wardragon to pretty much about where it should be as a T3 LW, and since then the various buffs to the Heliarchs should make it so a Punisher beats a Wardragon, which I believe should be the case, given it's only half a tier below it, so it should be more than a "match."
Continuing with that, I think Punishers should still pose some trouble to Quarg MWs imo, but should not be capable of beating them alone, and should be completely outmatched by a Quarg HW.
As for turrets, the Met Par Tek 53 has a total of six turret mounts and three gun ports, so even though it is an automated ship, there's precedent for LWs with many turrets, and I'd say it's reasonable to assume it could still function with four turrets, were it not automated (as in losing all gun ports and two turret mounts to compensate for the crew's living quarters, more than half the hardpoints and whatever mechanisms would be in place to allow guns and turrets to be equipped in them).

McloughlinGuy:
The skylark should be changed to a medium warship, as it is roughly the same as the arfecta in size and strength. I did a test once, and found that the arfecta and skylark have equal strength.

McloughlinGuy:
On second thought, changing classifications is something for another issue.

McloughlinGuy:
> 
> 
>     * `quarg ships.txt` needs to live in the `data/quarg/` folder, you have it in the root folder which results in this PR making two copies of the file.
I don't know how to fix that.

Amazinite:
Aside from the fact that this PR is improperly formatted, I'd really rather not hassle with the categorization of Quarg ships. They've admittedly been just placeholders for many years now and we plan to eventually completely overhaul them. Until then though, there's really no use in fiddling with their stats in any way.

Issue #5879, opened 2021-04-22
Check out this one simple trick to absolutely eat someone's storage with a single image file


Amazinite:
**Describe the bug**
The game currently writes errors encountered when loading the game to an error log. That's all fine and dandy, but it's currently possible to cause this error log to balloon in size with the inclusion of a single image file to the game.

**To Reproduce**
1. Rename or create an image file with a name that denotes a very high frame number. i.e. (<name>~X.png) where X is a very large number.
2. Launch the game.
3. Observe how your error file gets filled with line upon line of `Sprite "<path>": missing frame X.`
4. If the image you chose also has a single 2x version present, you'll also get `Sprite "<path>": missing @2x frame X` for every frame, doubling the number of errors.

**Expected behavior**
Under no circumstances should a single file be capable of filling the error log with so much junk.

**Screenshots**
After renaming `typhoon~3.png` to `typhoon~3000000.png` and launching the game, my error log grew to nearly a third of a gig in size. Notepad++ can't even function opening a text file of that size. Two more zeros and I would have eaten all the remaining space on my SSD and then some; not even sure what would happen after the game ran out of memory to write to.
![image](https://user-images.githubusercontent.com/17688683/115754675-4b8ad400-a351-11eb-9f23-ffea3bfea445.png)

**System (please complete the following information):**
 - OS: Windows 10.
 - Game Source: Self-compiled.
 - Version: Latest commit as of posting (e19c8e2).

**Link to save file**
I'd link the error file to be funny, but GitHub has a 10mb upload limit.

**Additional context**
The following code is the source of all the errors. If an image file is named with a very high frame number, the game expects all frames up to that number to exist, and throws an error for each and every one. 
https://github.com/endless-sky/endless-sky/blob/576f7a72678ae0fafe8a24c9494f34c62f51c847/source/ImageSet.cpp#L171-L186

Suggested that the error log instead list the *number* of frames missing instead of individually listing each frame that is missing.

Amazinite:
This is like an ASAP type of thing to fix to me. Not that I expect anyone to maliciously use this, but it's certainly a possibility, and I only caught it because someone for some reason had a 3mb error log filled with "missing frame" errors for thousands of lines, so clearly at least one plugin has run into this before, and to be honest this isn't the type of thing that I expect you to accidentally cause to happen without immediately fixing it if you noticed it.

MCOfficer:
The reasonable thing is to report this in ranges, i.e.
```
[...]

PR #5881, opened 2021-04-22
fix(logging): log missing frames more concisely

ashdnazg:
-----------------------
**Bugfix:** This PR addresses issue #5879

## Fix Details
Instead of logging an error for every missing frame, log an error just for the first gap in the frames and how many missing frames are there in addition, for example:
```
Sprite "projectile/typhoon": missing frame 2 and 299996 more.
Sprite "projectile/typhoon": missing @2x frames 0-4999 and 294999 more.
```

## Testing Done
Renamed typhoon~... and the @2x to a bunch of scenarios and checked the results


source/ImageSet.cpp
```diff
@@ -45,6 +47,44 @@ namespace {
...
+		{
+			message += "frame " + to_string(firstGapStart);
+		}
+		else
+		{
+			message += "frames " + to_string(firstGapStart) + "-" + to_string(firstGapEnd);
+		}
+		size_t extraMissing = totalMissing - (firstGapEnd - firstGapStart + 1);
+		if(extraMissing)
+			message += " and " + to_string(extraMissing) + " more";
```
tehhowch, line 84: I think it's fine to not print the exact missing range. The content creator has their filesystem view and can use that to find gaps once they know one exists.

So e.g. "missing frame X (<totalmissing>)" is fine by me.

tehhowch (commented):
Thanks for the PR!

I just have some minor nits, a variable rename request, and a request to simplify the output string, which should simplify the code a bunch too.

source/ImageSet.cpp
```diff
@@ -45,6 +47,44 @@ namespace {
 		// are part of the sprite name, not a frame index.
 		return (IsBlend(path[pos]) ? pos : end);
 	}
+
+	// Log an error if frames are missing in one of the paths vectors.
+	void LogIfMissingFrames(const vector<string>& paths, size_t expectedCount, string prefix, bool is2x)
```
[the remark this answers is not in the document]
ashdnazg: ah yes, I didn't like this either!

source/ImageSet.cpp
```diff
@@ -45,6 +47,44 @@ namespace {
...
+	{
+		size_t firstGapStart = paths.size();
+		size_t firstGapEnd = expectedCount - 1;
+		const size_t totalFrames = min(paths.size(), expectedCount);
+		size_t totalMissing = expectedCount - totalFrames;
+		for(size_t i = 0; i < totalFrames; ++i)
+		{
+			if(paths[i].empty())
+			{
+				totalMissing++;
```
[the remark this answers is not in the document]
ashdnazg: irrelevant now

source/ImageSet.cpp
ashdnazg: That simplifies the code significnatly indeed

ashdnazg:
pushed a simplified version

MCOfficer:
> ▶ Try to keep lines to about 80 characters in length, and wrap anything over 120 characters.

http://endless-sky.github.io/styleguide/styleguide.xml

source/ImageSet.cpp
```diff
@@ -45,6 +47,21 @@ namespace {
 		// are part of the sprite name, not a frame index.
 		return (IsBlend(path[pos]) ? pos : end);
 	}
+
+	// Log an error if frames are missing in one of the paths vectors.
+	void LogIfMissingFrames(const vector<string>& frameData, size_t expectedCount, string prefix, bool is2x)
```
tehhowch, line 52:
```suggestion
	void LogIfMissingFrames(const vector<string> &frameData, size_t expectedCount, const string &prefix, bool is2x)
```

source/ImageSet.cpp
```diff
@@ -45,6 +47,21 @@ namespace {
...
+	{
+		auto missingLambda = [](const string &s) { return s.empty(); };
```
tehhowch, line 54:
```suggestion
		auto isMissing = [](const string &s) noexcept -> bool { return s.empty(); };
```

source/ImageSet.cpp
```diff
@@ -45,6 +47,21 @@ namespace {
...
+		const auto totalEmpty = count_if(frameData.begin(), frameData.end(), missingLambda);
+		const auto totalMissing = expectedCount - min(frameData.size(), expectedCount) + totalEmpty;
+		if(!totalMissing)
+			return;
+
+		auto firstMissingIt = find_if(frameData.begin(), frameData.end(), missingLambda);
```
tehhowch, line 60: We make fewer trips through if we

1) find pseudo end PE, begin + min(expected, frameData.size)
2) find first missing frame FM in range [begin, pe)
3) fm == pe? Warn if pe != end and it isn't 2x (logic error), but always exit.
4) count missing in [fm, pe) and log it.

What do you think of this?

tehhowch (commented):
Looking better 👍

source/ImageSet.cpp
ashdnazg: good catch

source/ImageSet.cpp
ashdnazg: tbh, IMO this is a relatively minor optimisation in non-performance-heavy code, so readability should be prioritised higher.

ashdnazg: I changed it slightly to make a single pass and still be mostly readable (I think).

tehhowch (approved):
I fixed up some style nits and avoided counting 2x frames that the game already says it'll ignore:

e.g. if we touch `images/effects/static+10.png` and `images/effects/static+50@2x.png`, we'll get this output:
```
endless-sky (fix-one-simple-trick)$ ./endless-sky -p
Sprite "effect/static": 40 extra frames for the @2x sprite will be ignored.
Sprite "effect/static": missing frame 7 (3 missing in total).
Sprite "effect/static": missing @2x frame 7 (4 missing in total).
Parse completed.
```

PR #5888, opened 2021-04-25
Add tribute to Quarg planets/ringworlds

10010101001:
Add a challenge for players that have a very strong endgame fleet.


petervdmeer:
Sounds like fun, but this feels inappropriate for the Quarg; the Quarg are quite pacifistic and I feel they are not that easily intimidated.

Amazinite:
Note that the only tribute fleet that uses Quarg ships right now is actually a human inhabited planet that is under the protection of the Quarg. The Quarg otherwise don't really care for your threats, I imagine especially when you demand tribute to a ringworld. If no one but the Pug before you have managed to damage a ringworld, what do you think you'll be able to threaten them with?

Even ignoring that, if we were to allow demanding tribute from the Quarg, the thresholds would be much higher than these.

Plip-Plop:
I think the Quarg would run away if you could pose a serious threat to them. They would probably send a medium sized fleet to defend themselves, and if you continued to threaten them and harm the ringworld they would just pack up and leave to places you cannot reach. This is all assuming you could even make them feel threatened.

Pointedstick:
I don't see us doing this, sorry.

PR #5892, opened 2021-04-26
Suitable bulk-style sprite for Class-C Freighter, splitting of bulk freighter

ravenshining:
**Content (Artwork)**

## Summary

There are four issues with the new bulk freighter sprite:

1. It doesn't look like it can fit 12 drones in its 24 cargo containers.
2. Unlike MZ's bulk freighter, the new sprite clearly depicts a container ship, not a bulk carrier.
3. Some people complain its silhouette is too rectangular and lacks the interesting nooks and crannies of MZ's design.
4. It changes the turret arrangement for saves made in 0.9.12 and earlier

This PR adds an alternate sprite using the same backbone as the new bulk freighter, but replaces the 24 intermodal cargo containers with 12 large general-purpose holds, complete with hatchways like MZ's sprite, and with added gearing.  Each of the 12 holds has been sized to fit a combat drone, whether at 100% size exiting through unseen ventral or side hatches or at a possible future smaller scale exiting through the dorsal hatches.

For semantic and compatibility reasons, the present sprite has been renamed the "Container Transport," since a container ship is what it seems to depict.

Addressing concerns raised by the current sprite's artist and in reflection of real-world trends, the Container Transport will spawn much more commonly - anywhere from twice to five times as much - but only between heavily populated planets and in Syndicate space.  The new bulk freighter will not spawn as often overall, but it will still spawn over and between smaller Republic  worlds or worlds that are described as having limited spaceport facilities, and more often in Hai space where ships that are independently owned and not plying the major trade routes are more likely to wind up.

Considering the Syndicate's predilection for modular designs, I have kept the new five-down-the-spine turret arrangement, however, to restore compatibility with old saves and for variety, I would be delighted to change the Bulk/Class-C model to use MZ's original rectangular turret arrangement or any other throwbacks if people would prefer.

## PR Checklist
 - [x] I uploaded the necessary image, blend, and texture assets here:https://drive.google.com/file/d/1jRYbyPd-6sCNgLYvWUo8Dr13WT-QLTQF/view?usp=sharing
 - [x] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: https://github.com/endless-sky/endless-sky-high-dpi/pull/175
    
-----------------------
## UI Screenshots

Screenshot_20210508_203255

## Testing Done
I have sat in various systems to ensure that both variants appear, but the Container Transport appears more often.

## Performance Impact
N/A


ravenshining:
For comparison, here's an alternative model with rectangular turret config as in 0.9.12 on the left, with the model in this PR with spinal turrets as the 0.9.13 bulk freighter on the right:

Screenshot_20210425_182425

Now that I've made it, I'm personally partial to the one on the left, especially for variety, but I'll just leave that here until I get some feedback.

Amazinite:
I'll note that by moving the current bulk freighter to a new ship and replacing the "Bulk Freighter" name with this new one, you're already going to swap the sprite of whatever ships currently exist in save files. Given the similar dimensions that shouldn't be an issue, but moving the turret hardpoints around would make it an issue. So if you want the 2:1:2 turret set up instead of the 1:1:1:1:1 set up, I'd suggest making that change to the now-named "Combined Freighter"--a name which I'll also add right now that I'm not too fond of; is having them both named "Bulk Freighter" and using variants like the Auxiliary does out of the question? That way the existing Bulk Freighter wouldn't need to change, avoiding changing the ships people already own. We could perhaps also differentiate the stats a bit more between the two.

ravenshining:
> is having them both named "Bulk Freighter" and using variants like the Auxiliary does out of the question? That way the existing Bulk Freighter wouldn't need to change, avoiding changing the ships people already own. 

Except that is exactly what the recent change to the Bulk Freighter did a few months ago.  Anyone who owned a bulk freighter bought on continuous prior to then, or bought in 0.9.12 or earlier has now had their turrets messed up, and when I pointed out that this would happen at the time you said it was not a big issue.

Adding a 2:1:2 and having that be the default version would *fix* a lot of ships people already own, and only mess with bulk freighters purchased very recently.

The other problem with that is that the containery version would still be called a "Bulk Freighter," which is exactly what I am trying to fix with this PR.  Using variants causes the same problem, with the added problem of now only only one version would be available in the shipyard.  Having both be ships and moving the current one as I have done solves both these issues.

> "Combined Freighter"--a name which I'll also add right now that I'm not too fond of

Yeah, to be honest I agree.  Other options:
- Intermodal Freighter
- Modular Freighter
- Containerised Freighter
- Container Freighter
- Container Transport

On Discord last night, I believe you were gravitating toward the last while I was gravitating towards the first.

> We could perhaps also differentiate the stats a bit more between the two.

Ah, with pleasure!  I was worried my modest tweaks to hull, shields, drag, and heat dissipation would be too much.  I've considered giving them different loadouts, perhaps equipping some Bulk Freighters with ramscoops, plasma engines, and meteor missiles since they'll be spotted further afield than the more common containerised freighter.  I was also thinking having the 2:1:2 turret arrangement would be good for this purpose.

Most people on Discord last night seemed to prefer the 1:1:1:1:1 turret arrangement, but on questioning them further, I think people were considering the ship in isolation and not as something that will be operating alongside the existing 1:1:1:1:1 freighter, so I am still leaning towards 2:1:2 for now, unless that means it's a no-go for the renaming.

Finally a number of people pointed out that my sprite is too colourful, or at least too saturated.  While I was initially resistant (as the first person to point this out was just pulling my leg, leading me to not take others seriously), I have come to agree.  I wasn't really accounting for swizzle when I made it, I was perhaps too worried about using the same shades of red as other ships use, and one of the tactics I was using to reduce the colour doesn't come out in the 1x sprite.  So, I will tone down or possible reverse some of the colouring.

ravenshining:
If I were to make the proposed changes (2:1:2 turrets, reduce colour, choose better name for the other sprite) *and* limit the scope of this PR, maybe it would be a good idea to put this in 0.9.14, to limit the damage being done by the current bulk freighter sprite.

That way, there won't be any changing of anyone's ships or turrets who have remained on stable releases, only people buying bulk freighters on continuous and beta for a very short period of time will be affected.

Then, I'll follow up this PR with a second for 0.9.15 with any of the changes from this one you don't feel would be suitable for stable.

43897689431805741092:
I believe the use of the term 'bulk' to describe this freighter was probably originally used in the layman context of 'bulk quantities', essentially just saying it's a big freighter. There aren't cargo types in this game.

Amazinite:
Yeah, I think you're reading too much into the name by saying "That bulk freighter isn't actually a bulk freighter."

ravenshining:
https://en.wikipedia.org/wiki/Bulk_freighter

Whether MZ was ignorant of that or not doesn't really matter, though he *did* design something that looked like a bulk freighter. Sure, most players won't know the difference - but get it wrong when you don't have to?  And why on earth would straightening out definitions be a point *against* a PR?

ravenshining:
Did some more work tonight taking into account feedback recieved on Discord, haven't commited yet:
Screenshot_20210427_232113

RecursiveParadox:
Not that it really matters, but IRL bunkers and boxships look very different - much more like the ones in this PR. I like these new ships. And it open up the possibility of differentiating between types of cargo in the future for more diverse missions.

And FWIW there was a period in shipping from about 1975-1995 or so where you did have "combined" ships called OBOs ("ore/bulk/oil"), and you still have ships that take both bulk or unitized cargoes as well as containerized cargoes called General Cargo vessels, although these tend to be on the small side.

ravenshining:
Thanks!

I'm not sure "General Cargo Vessel" really works as a name though, unless we wanted to call one "Bulk Freighter (OBO)" and the other "Bulk Freighter (General Cargo)," which seems a lot clumsier to me than "Bulk Freighter" and "Container Transport."

If we absolutely *have* to keep the current sprite as the "bulk freighter" - a completely nonsensical path with no real benefits - then we'll have to have a better name for *this* bulk carrier... and I really can't think of anything more appropriate.

Meanwhile, I've been touching up the sprite and my latest revision has received messages of approval from everyone who didn't like the first revision.  I really think 2:1:2 is the way to go to make this a visually distinctive ship, on top of the benefit of repairing saves with bulk freighters acquired on 0.9.12 and earlier.

I'm toying with the idea of some markings like MZ's pre-Blender Bulk Freighter had, though I am noncommittal on this and can easily remove them if they are not liked.

Screenshot_20210428_224840

ravenshining:
Along with the updates made on the other PR, I had a go at making the stats more divergent, using the 3D Print add-on in Blender to compare the volumes of various components to generate some numbers and ideas.  In particular, the cargo holds on the Bulk Freighter are *much* bigger, though it might not look that way from the top-down perspective - which I think plays nicely in to its role as the baseline for the Class C.  Everything has been done with lore and role in mind:

- Bulk Frieghter has more cargo space than the Container Transport - the CT trades off space for more efficient loading
- Bulk Freighter has more fuel than the Container Transport - the BF services more remote locations than the latter
- Bulk Freighter has less drag than the Container Transport - thanks to the BF's smooth hull
- Container Transport has better heat dissipation than the Bulk Freighter - the CT has many more fins
- Container Transport is cheaper than the Bulk Freighter - the CT uses more ~~arrays~~ standardised parts
- Container Transport requires less crew than the Bulk Freighter - the BF needs crane operators

data/human/fleets.txt
```diff
@@ -655,6 +699,36 @@ fleet "Large Northern Merchants"
 	variant 1
 		"Scout (Speedy)" 3
 
+fleet "Northern Container Transports"
```
Amazinite, line 702: Not a fan of such a specific fleet addition like this. Can't such fleet variants just exist in the merchant fleets that spawn where this fleet is being added?

data/human/ships.txt
```diff
@@ -812,7 +812,8 @@ ship "Bulk Freighter"
 	explode "large explosion" 35
 	explode "huge explosion" 20
 	"final explode" "final explosion large"
-	description "The Bulk Freighter is the ultimate cargo ship, able to carry as much cargo as four ordinary Syndicate Freighters. It is also, however, an incredibly slow ship and difficult to maneuver, especially when loaded down with cargo. Some freight caravans use ordinary freighters instead of bulk freighters just to avoid the hassle of constantly having to wait for the Bulk Freighter to properly align itself for hyperspace travel."
+	description "The Bulk Freighter is the ultimate cargo ship, able to carry as much cargo as four ordinary Syndicate Freighters. Its on-board gearing allows it to service even the most primitive colonies, though dock workers in busier spaceports often joke that it would be simpler to carry smaller freighters within its gigantic holds."
+	description "It is also an incredibly slow and difficult to maneuver ship, especially when loaded down with cargo. Some freight caravans use ordinary freighters instead of bulk freighters just to avoid the hassle of constantly having to wait for the Bulk Freighter to properly align itself for hyperspace travel."
```
Amazinite, line 816:
```suggestion
	description "	It is also an incredibly slow and difficult to maneuver ship, especially when loaded down with cargo. Some freight caravans use ordinary freighters instead of bulk freighters just to avoid the hassle of constantly having to wait for the Bulk Freighter to properly align itself for hyperspace travel."
```

data/human/ships.txt
```diff
@@ -1057,6 +1070,66 @@ ship "Combat Drone"
...
+	leak "leak" 40 50
+	leak "flame" 30 80
+	leak "big leak" 50 30
+	explode "small explosion" 15
+	explode "medium explosion" 25
+	explode "large explosion" 35
+	explode "huge explosion" 20
+	"final explode" "final explosion large"
+	description "The Container Transport is a newer, more popular configuration of the Bulk Freighter that replaces the outboard loose cargo holds with intermodal containers for more efficient turnaround in larger, better-equipped ports."
+	description "As with its geared cousin, however, it remains a slow and difficult to maneuver ship, especially when loaded down with cargo. Some freight caravans use ordinary freighters instead of bulk freighters just to avoid the hassle of constantly having to wait for the Bulk Freighter to properly align itself for hyperspace travel."
```
Amazinite, line 1129:
```suggestion
	description "	As with its geared cousin, however, it remains a slow and difficult to maneuver ship, especially when loaded down with cargo. Some freight caravans use ordinary freighters instead of bulk freighters just to avoid the hassle of constantly having to wait for the Bulk Freighter to properly align itself for hyperspace travel."
```

data/human/variants.txt
```diff
@@ -454,7 +454,6 @@ ship "Bulk Freighter" "Bulk Freighter (Proton)"
 		"Scram Drive"
 
 
-
```
Amazinite, line 457: Please don't eat the whitespace.

data/human/variants.txt
```diff
@@ -566,6 +565,80 @@ ship "Clipper" "Clipper (Speedy)"
...
+		"Meteor Missile" 60
+		"Fission Reactor"
+		"LP072a Battery Pack"
+		"D94-YV Shield Generator"
+		"Outfits Expansion"
+		"Laser Rifle"
+		"A250 Atomic Thruster"
+		"A375 Atomic Steering"
+		"Scram Drive"
+
```
Amazinite, line 641: You need an extra line of whitespace down here.

Amazinite (changes requested):

Amazinite:
Closing this as #5949 was merged. Further discussions on differentiating the two ships can happen later.

PR #5894, opened 2021-04-26
Make Ka'Het ships take damage if overheated

10010101001:
Since Ka'Het are huge living organisms, overheating them should cause them to die. This PR makes Ka'Het ships take damage if their heat level reaches 120%.


Vedvod:
If Ka'Het are killed from heat, this makes weapons that deliver heat (such as plasma?) very effective against them, so I also suggest some form of very efficient cooling to counteract this, but not so much that it is difficult to overheat them, just at a level that plasma or whatever doesn't make it a super easy fight.

beccabunny:
120% really seems too low. I suggest making the value 200% and eventually buffing their active cooling, which is easily reachable anyway with heat weapons

10010101001:
Buffed Ka'Het heat dissipation by 10%

data/ka'het/ka'het ships.txt
```diff
@@ -13,6 +13,7 @@ ship "Maeri'het"
 	sprite "ship/maeri'het"
 	thumbnail "thumbnail/maeri'het"
 	"never disabled"
+	"cookable"
```
petervdmeer, line 16: I like the keyword `cookable`, but I also feel that a configurable factor is more appropriate to allow some creatures to die at a factor 1.2 (like is done in this PR), while other creatures could die at a factor of 3. (something like `"fatal heat factor" 1.2` and `"fatal heat factor" 3.``.

source/Ship.cpp
```diff
@@ -2009,7 +2011,12 @@ void Ship::DoGeneration()
 	
 	heat -= heat * HeatDissipation();
 	if(heat > MaximumHeat())
-		isOverheated = true;
```
tehhowch, line 2012: The ship is overheated, so this should be set even for "cookable" ships.

source/Ship.cpp
```diff
@@ -2009,7 +2011,12 @@ void Ship::DoGeneration()
...
+		if(cookable && heat > MaximumHeat()*1.2) {
+			hull = MinimumHull() - 1;
```
tehhowch, line 2015: Cooking is a process, not instantaneous.
Instead of a step change in the hull value at some magic heat level, the hull should decrease proportionally to the excess heat.

(Also, unconditionally setting the hull to some value just below the minimum could _repair_ the hull when taking heat damage, if the hull is more damaged than that value.)

tehhowch (changes requested):

source/Ship.cpp
10010101001: Done

10010101001:
Set Ka'Het cook heat to 150%

source/Ship.cpp
10010101001: The point is the organism inside the shell dies, not the hull taking damage but there is no way to permanently disable a ship without lowering its hull below the minimum as far as I know

petervdmeer: Hull is used more as a "representation of the health of the total ship/creature" than as a reference to just the outer shell.
We don't differentiate in the game between ships that have all internal systems blown out (what you would expect with piercing ammunition) and ships that have the outer hull blown to pieces. (All ships that get destroyed act like the latter and explode as if they were fully blown to pieces.)

data/ka'het/ka'het ships.txt
```diff
@@ -22,7 +22,8 @@ ship "Maeri'het"
 		"bunks" 1
 		"mass" 390
 		"drag" 3.8
-		"heat dissipation" .4
+		"heat dissipation" .8
+		"cook temperature" 1.2
```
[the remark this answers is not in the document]
10010101001: Done

source/Ship.cpp
10010101001: Done

tehhowch (changes requested):

source/Ship.cpp
```diff
@@ -2009,7 +2009,12 @@ void Ship::DoGeneration()
 	
 	heat -= heat * HeatDissipation();
 	if(heat > MaximumHeat())
+	{
 		isOverheated = true;
+		if(CookHeatFactor() && heat > MaximumHeat()*CookHeatFactor()) {
+			hull -= CookDamageFactor()* heat / MaximumHeat();
```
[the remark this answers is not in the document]
10010101001: Fixed

data/ka'het/ka'het ships.txt
```diff
@@ -22,7 +22,9 @@ ship "Maeri'het"
...
+		"cook damage" 230
```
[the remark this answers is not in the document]
10010101001: Done

source/Ship.cpp
```diff
@@ -2009,7 +2009,12 @@ void Ship::DoGeneration()
 	
 	heat -= heat * HeatDissipation();
 	if(heat > MaximumHeat())
+	{
 		isOverheated = true;
+		if(CookHeatFactor() && heat > MaximumHeat()*CookHeatFactor()) {
```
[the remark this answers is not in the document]
10010101001: Done

source/Ship.h
```diff
@@ -306,6 +306,10 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	double MaximumHeat() const;
 	// Calculate the multiplier for cooling efficiency.
 	double CoolingEfficiency() const;
+	// Get the heat level at which a ship is instantly disabled.
+	double CookHeatFactor() const;
+	// Get the damage multiplier for hull damage when overheating.
+	double CookDamageFactor() const;
```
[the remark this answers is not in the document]
10010101001: Removed

quyykk (commented):
@Amazinite Is the content side (attribute names and general feature) ok?

source/Ship.cpp
```diff
@@ -2009,7 +2009,13 @@ void Ship::DoGeneration()
 	
 	heat -= heat * HeatDissipation();
 	if(heat > MaximumHeat())
+	{
 		isOverheated = true;
+		float cookHeat = attributes.Get("cook temperature");
+		if(cookHeat && Heat() > cookHeat) {
+			hull -= attributes.Get("cooking rate")* Heat();
+		}
+	}
```
Amazinite, line 2018:
```suggestion
	if(heat > MaximumHeat())
	{
		isOverheated = true;
		double heatRatio = heat / (MaximumHeat() * (1. + attributes.Get("overhead damage threshold"));
		if(heatRatio > 1.)
			hull -= attributes.Get("overheat damage") * heatRatio;
	}
```

Amazinite (changes requested):
> @Amazinite Is the content side (attribute names and general feature) ok?

`"cook rate"` -> `"overheat damage"`: the damage taken per frame that a ship is beyond it's overheat damage threshold (appearing in game as per second). The more overheated a ship is beyond its threshold, the more damage it takes.
`"cook temperature"` -> `"overheat damage threshold"`: the percentage *beyond* being overheated at which point a ship starts taking overheat damage. (1.2 would be changed to .2 in the attributes.)

Zitchas:
Just to clarify, this is something that we can apply to any ship via attributes? I can think of many situations where overheating could cause damage.

I know some would argue (and thus I suspect we'll see in plugins) a situation where *every* ship takes damage from overheating, just with the levels varying. Some things (Korath, Automatons) would probably be very tolerant of overheating; whereas a lot of other things would start taking damage at lower levels, with living ships and space life forms possibly being very sensitive to it.

quyykk:
> Just to clarify, this is something that we can apply to any ship via attributes?

Yes exactly.

> Might also be worth tackling the actual implementation on the Ka'het ships in a separate PR.

For simplicity I'll let them stay in this PR if that's ok.

Amazinite (approved):

Amazinite:
@quyykk Look good?

data/tooltips.txt
```diff
@@ -347,6 +347,12 @@ tip "outfit space needed:"
 tip "outfit space free:"
 	`How much space this ship has for outfits (including weapons and engines).`
 
+tip "overheat damage rate:"
+	`How much hull damage this ship will take per second that it is overheated.`
```
tehhowch, line 351: Should this mention that the value scales with the amount of excess heat?

tehhowch (commented):
Making this damage scale by the excess heat will make flamethrowers even more powerful against whichever ships receive this attribute. Should it be a flat rate instead?

data/tooltips.txt
Amazinite:
```suggestion
	`How much hull damage this ship will take per second that it is overheated. Overheat damage scales with how far past the overheat threshold the ship is.`
```

Amazinite:
> Making this damage scale by the excess heat will make flamethrowers even more powerful against whichever ships receive this attribute. Should it be a flat rate instead?

The damage scaling does improve the effectiveness of flamethrowers, but I think that because of the fact that flamethrowers are limited by fuel and the chosen overheat damage values in this specific case are so low that it isn't too big of an issue. Using 8 flamethrowers and 3.1k fuel, I was able to push a stationary Vareti'het HW escort to about 400% heat before running out of fuel, which is about 3.33x overheat damage. By the time the ship had recovered it still had about 1/3rd of its hull left. So by wasting all my fuel and being 100% accurate by firing on a stationary target, I didn't even take down a heavy warship. Doing the same thing on a Maeri'het LW escort I was able to get it to over 1400% heat (11.67x damage) and it died before it could cool down, and by conserving fuel I was able to kill it in a longer time while using 2/3rds of my fuel. So with the current values, it still takes a significant amount of fuel to either only severely damage a heavy warship or knock out a light warship, and that's with 100% accuracy; in practice, particularly with a more nimble light warship, the actual values will end up much lower with some fuel being wasted when you miss. You'll still be reliant on conventional weaponry for taking on more than one ship or when fighting larger ships that can deal with heat damage better, unless you're completely speccing for high fuel and ramscoops, at which point you're likely able to spec for conventional weaponry that can take down a ship quicker anyway: using 8 plasma cannons instead of flamethrowers, I was able to take down Ka'het ships before they even overheated while not taking too much longer than using flamethrowers.

Could it be a flat rate? Sure, but I find it more interesting if the damage scales up, and as of right now that scaling doesn't appear to be too much of an issue.

quyykk (approved):

PR #5897, opened 2021-04-26
Allow player to speak with all researchers on Emerald Sword

obvious429:
## Summary
Sends player back to dialogue choices to speak with other researchers instead of skipping to the Emerald Sword being operational. 

There's a lot of interesting dialogue and lore that can be read when you talk with the researchers after you first board the Emerald Sword, and I think you should be able to talk with all of them instead of having to reload a save or look at the mission txt.

## Save File
[Test Pilot~Emerald Sword 2.txt](https://github.com/endless-sky/endless-sky/files/6379196/Test.Pilot.Emerald.Sword.2.txt)

image


Amazinite:
We intentionally left it so that for now you only get to speak with one member of the crew per run. If we were to make it so that you could speak with everyone, what you should be doing is removing the choices for the people you already spoke to, not allowing the player to repeatedly pick one. That is currently possible, but it requires a lot of branching. Given that #5440 is open though and seemingly functional, I decided to not mess with all the branching that would be needed to allow you to speak with every crew member while also hiding already picked choices and instead wait until we merge that PR.

PR #5913, opened 2021-04-29
feat(content): Add pirate raid tutorial

Terin:
**Content (Tutorial)**

## Summary
This PR adds a tutorial when the player first encounters a pirate raid, spawned by having too much defenseless cargo, prompted by [this Steam post](https://steamcommunity.com/app/404410/discussions/0/3130541756134242226/). It pauses the game to alert the player the pirates in system are caused by their actions, and having too much cargo space, not just cargo. It then suggests four courses of action:

- Park an offending escort
- Sell an offending escort
- Arm the freighters with more offensive weaponry
- Buy combat escorts

This should hopefully make this mechanic much clearer, than the "Your fleet has attracted the interest of a Pirate raiding party" that disappears in a few seconds in the corner, and the little-context percentage in the Player Info Panel.

### DIFF
--- a/data/help.txt
+++ b/data/help.txt
@@ -124,3 +124,7 @@ help "trading"
 
 help "friendly disabled"
 	`This ship is disabled and needs your help to patch them up! Pressing <Board selected ship> will fly you to their ship and get them up and running again. Just be sure to do it when no one is trying to kill you, or they could get caught in the crossfire!`
+
+help "pirate raid"
+	`Your fleet has attracted the interest of a raiding party! By having too much cargo space in your fleet without enough defenses, word will get out that you're an easy target, and raiders have the possibility of waiting in ambush each jump you make.`
+	`The chance of drawing attention can be seen on the Player Info Panel with <View player info>. To reduce that chance, when on a planet, you can use the Player Info Panel to park ships with lots of cargo space, by selecting them and clicking the Park button at the bottom, or by selling them. Alternatively, you can arm your freighters with offensive weaponry, or buy interceptor or warship escorts to deter the raiders in the first place.`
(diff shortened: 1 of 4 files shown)


Amazinite (commented):
Starting on the implementation, this certainly looks like it'll work, but for future reference we shouldn't be adding public variables to classes (outside of minor classes that are contained within larger ones, like the Bay class in Ship), so the public bool is a no no, and should instead be private and accessed by MainPanel by a function in Engine. Something something good object oriented programming practice. Otherwise though, this looks good.

The main reason for the thumbs down though is that we actually already explain pirate raids to the player in-universe twice (and I'd prefer to explain things in-universe instead of with a tutorial where possible, and pirate raids are definitely one of those in-universe things), the first time being during the James tutorial (but only if you're doing the Star Barge start) and the second time when your pirate attraction is greater than 50%, triggering a new mission that warns you about pirate raids and how to avoid them. I went with a fairly high value for that mission just to start, but I'm certainly open to changing the value on that warning mission to be much lower.

Terin:
Yeah, ended up forgetting about that mission until Anarchist pointed it out on the Discord. That mission is pretty good in-universe, although the two words about "park them" will probably mean nothing to the target players of that mission, unless they have veteran players to turn around and ask.

Maybe for future PRs, could have it so the Hunted Warning mission has a chance to spawn at lower amounts (around 20% or so?), with a more explicit dialog option about parking ships a la the Quantum Key Stone missions. But yeah, in the meantime, too much overlap with this here.

Amazinite:
I meant to mention this earlier but forgot: the issue with the raid message being easy to miss is also related to https://github.com/endless-sky/endless-sky/issues/3361.

PR #5922, opened 2021-04-30
ShipInfoDisplay: Break out heat/energy table values into their own variables

jostephd:
This PR implements the feature request detailed and discussed in ~~issue #{{insert number}}~~ this PR, but see #4215 for a listing of related issues

## Feature Details
Refactor the code that fills the ShipInfoDisplay heat/energy table to have each table cell's numeric value available in a local variable. This will make it easier to write a future patch that adds a `sum` row to the table without duplicating the calculations.

If adding rows isn't acceptable (see https://github.com/endless-sky/endless-sky/pull/3835#issuecomment-440531655), I'll add that row in a local patch (which I'll post it on my fork if there's interest), but even in that case it'd be nice if the PR was accepted so I would have fewer rebase conflicts with that patch.

## UI Screenshots
No difference before/after the patch.

## Usage Examples
N/A

## Testing Done
Loaded the Shipyard before/after the change and observed no difference.

## Performance Impact
None. The compiled, optimized code will be the same.

## Other
The final two whitespace-only hunks are intentional. They make it so the code for each table row starts with an empty line and a `attributesHeight += 20` line (except the first row).

### DIFF
--- a/source/ShipInfoDisplay.cpp
+++ b/source/ShipInfoDisplay.cpp
@@ -270,29 +270,30 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 	// Skip a spacer and the table header.
 	attributesHeight += 30;
 	
+	const double idleEnergyPerFrame = attributes.Get("energy generation")
+		+ attributes.Get("solar collection")
+		+ attributes.Get("fuel energy")
+		- attributes.Get("energy consumption")
+		- attributes.Get("cooling energy");
+	const double idleHeatPerFrame = attributes.Get("heat generation") 
+		+ attributes.Get("solar heat")
+		+ attributes.Get("fuel heat")
+		- ship.CoolingEfficiency() * (attributes.Get("cooling") + attributes.Get("active cooling"));
 	tableLabels.push_back("idle:");
-	energyTable.push_back(Format::Number(
-		60. * (attributes.Get("energy generation")
-			+ attributes.Get("solar collection")
-			+ attributes.Get("fuel energy")
-			- attributes.Get("energy consumption")
-			- attributes.Get("cooling energy"))));
-	double efficiency = ship.CoolingEfficiency();
-	heatTable.push_back(Format::Number(
-		60. * (attributes.Get("heat generation") 
-			+ attributes.Get("solar heat")
-			+ attributes.Get("fuel heat")
-			- efficiency * (attributes.Get("cooling") + attributes.Get("active cooling")))));
+	energyTable.push_back(Format::Number(60. * idleEnergyPerFrame));
+	heatTable.push_back(Format::Number(60. * idleHeatPerFrame));
+
 	attributesHeight += 20;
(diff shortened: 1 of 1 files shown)


petervdmeer (approved):
This change seems to put the results of calculations into their own local variables, exactly as the description says.
Looks good to me, also as a stand-alone change, since it slightly improves readability.

jostephd:
Thanks, folks. As to the `sum` patch I mentioned in the OP, I'm trying it as a `headroom` patch instead, like this:

2021-05-05-211559_232x135_scrot

144-96-12 = 6 and 744-174-174 = 396. (Would work with armed ships too.)

<details>

<summary>
Add `headroom` line
</summary>

```diff
diff --git a/data/tooltips.txt b/data/tooltips.txt
index 24fec9d9e..7fdce2a36 100644
--- a/data/tooltips.txt
+++ b/data/tooltips.txt
@@ -431,6 +431,9 @@ tip "charging shields:"
 tip "max:"
 	`Energy storage capacity and maximum safe level for heat generation. Energy capacity allows a ship to temporarily draw more energy than it produces. The heat maximum is the highest your heat generation can be without eventually causing your ship to overheat.`
 
+tip "headroom:"
+	`For energy, how much energy generation exceeds energy consumption. For heat, how much maximal heat dissipation exceeds heat generation. The higher the values, the better.`
+
 
 
 # Outfit categories:
diff --git a/source/ShipInfoDisplay.cpp b/source/ShipInfoDisplay.cpp
index 35239834b..5fb1d7a42 100644
--- a/source/ShipInfoDisplay.cpp
+++ b/source/ShipInfoDisplay.cpp
@@ -327,6 +327,30 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 	tableLabels.push_back("max:");
 	energyTable.push_back(Format::Number(maxEnergy));
 	heatTable.push_back(Format::Number(maxHeat));
+
+	attributesHeight += 20;
+	tableLabels.push_back("headroom:");
+	energyTable.push_back(Format::Number(
+				// ignore maxEnergy
+				60. * (
+					idleEnergyPerFrame
+					 - movingEnergyPerFrame
+					 - firingEnergy
+					 - (shieldEnergy + hullEnergy)
+				)
+				));
+	// TODO(josteph): probably incorrect; see #3704 #3835 #4215 etc
+	//                e.g., https://github.com/endless-sky/endless-sky/pull/3704#issuecomment-385273226
+	heatTable.push_back(Format::Number(
+				maxHeat -
+				60. * (
+					idleHeatPerFrame
+					 + movingHeatPerFrame
+					 + firingHeat
+					 + (shieldHeat + hullHeat)
+				)
+				));
+
 	// Pad by 10 pixels on the top and bottom.
 	attributesHeight += 30;
 }
```

(This diff is also https://github.com/jostephd/endless-sky/tree/ShipInfoDisplay-add-headroom)

</details>

I'm sure there's room for improvement. Anyway, should I open a PR for this?

PR #5925, opened 2021-05-01
Fix punctuation in Hai mission

Ornok:
Fixed punctuation in `Nanachi 4`:
`While [clause 1], and [clause 2]. [clause 3].` -> `While [clause 1] and [clause 2], [clause 3].`


roadrunner56 (approved):

tehhowch:
Seems like an even better fix is cleaning up the grammar entirely. This whole `While...` sentence is awkward.

Amazinite:
How about this?
> It's certain that for a pirate fleet to take down a Leviathan so easily they must be incredibly powerful. This could be extremely dangerous; you're sure that if Nanachi tries to get to Prime by herself she'll just get herself killed.

I just dropped While and split it into two sentences.

tehhowch:
I don't think the sentence is even needed, as other characters' speech and the player's response choices mention its apparent strength.

Issue #4752, opened 2020-01-22
Priority and minor keywords don't apply to landing missions


Amazinite:
**Describe the bug**
`priority` and `minor` keywords don't apply to `landing` missions. 

**To Reproduce**
See #4310. Putting `priority` on "FW Pug 2C: Jump Drive" and landing on Hephaestus with a JD and an HD will still cause "FW Pug 2C: Hyperdrive" to offer first. 

**Expected behavior**
`priority` and `minor` should still apply to `landing` missions, meaning that `priority` and `landing` missions will only offer with other `priority` `landing` missions, and no other mission will offer if the mission is `minor` and `landing`.

**System (please complete the following information):**
- OS: Windows 10
- Game Source: Self-compiled
- Version: Up to date as of this issue (eeeb523).

**Link to save file**
Here is a save file with a ship that has both an HD and a JD one jump away with Hephaestus with the necessary conditions for "FW Pug 2C" to offer. Add `priority` to your "FW Pug 2C: Jump Drive" and land on Hephaestus.
[Test Pilot.txt](https://github.com/endless-sky/endless-sky/files/4094619/Test.Pilot.txt)

**Additional context**
Should this actually be changed, or should it just be specified in the CreatingMissions page that `minor` and `priority` only apply to spaceport offered missions?

tehhowch:
I think the design idea was that `landing` is to be used for follow-up missions, in which the player is already on their quest, and it would be weird if something interrupted the flow. Which does make it difficult in this situation, where we want to select a particular mission "overload".

A probable workaround for missions of the same series is to name them such that the missions which should be checked before others comes first alphabetically. This is an artifact of the current implementation for storing missions, which is in a hashmap keyed by the name of the mission.

Amazinite:
If priority can be emulated with alphabetizing the mission names how you want them to offer, I'd say this issue doesn't really need fixed. (The wiki should still be updated to reflect this, though.)

Amazinite:
As outlined by https://github.com/endless-sky/endless-sky/pull/5930#issuecomment-830863457, it would appear that the intention was to at least have `priority` and `landing` mix, as the first intro mission has both tags, and the wiki describes the intro mission as having priority to avoid having any other (plugin) missions potentially offer at the same time, which does not occur at the moment; you could have any number of missions offer alongside the intro and the priority tag will do nothing.

Amazinite:
Fixed by #11416.

PR #5930, opened 2021-05-02
Warn about minor or priority being mixed with landing for missions

Amazinite:
**Feature:** This PR implements a feature that should stop my dumb ass from making changes like https://github.com/endless-sky/endless-sky/commit/62f0226bd1483c0d6ec02b01c258ac9f2993c826 again.

## Feature Details
This PR adds a warning that gets written to the error file if a mission has either the minor or priority tag and also has the landing tag, as minor and priority don't effect landing missions.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
Let's see what the test parses turn up.

## Performance Impact
N/A

### DIFF
--- a/data/coalition/coalition missions.txt
+++ b/data/coalition/coalition missions.txt
@@ -229,7 +229,6 @@ mission "Coalition: Lost: Offer"
 mission "Coalition Yottrite 1"
 	name "More Yottrite"
 	description "You were offered great payment for bringing three more samples of the rare mineral yottrite to <destination>."
-	minor
 	landing
 	to offer
 		has "Coalition: Contributor: done"
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -273,7 +273,6 @@ mission "Ask the Hai about the Unfettered"
 
 mission "Unfettered Jump Drive 1"
 	name "More Jump Drives"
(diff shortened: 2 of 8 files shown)


Amazinite:
Well well well. It would appear I'm not the only one who's made this mistake.

Amazinite:
I'll note that the wiki mentions that `priority` is only used for the intro mission as a way to prevent any other missions from offering, but it's also a landing mission that appears in the special case situation of after you've just bought your first ship, so really it's the `landing` that's making it offer while `priority` has no effect. Given that MZ's expectation was clearly that mixing `priority` and `landing` would work, should we reassess #4752 and perhaps allow these tags to be mixed and have an effect? (Multiple landing missions where they're all minor = only one offers. Multiple landing missions where some are minor and some have no extra tag = only the non-minor missions offer. Multiple landing missions where some are minor, some have no tag, and one or more have priority = only the priority mission(s) offers).

tehhowch (approved):

tehhowch:
I'm also fine with extending the behavior of minor and landing. It's fine to include this warning now, and update them to work as expected in a later release.

Amazinite:
In that case I'll reopen that issue and we can merge this one now.

Issue #5939, opened 2021-05-06
Planetary storage: Show stored outfits by planet


jostephd:
**Is your feature request related to a problem? Please describe.**
#4443 added planetary outfits storage and a listing of _how many_ outfits are stored on which planet, but there's no way to see _which_ outfits are stored without landing there. For example, if there's 3 Beam Lasers and 4 batteries stored, all a player can know without landing there is the overall count, `7`.

**Related Issue Links**
#4443

**Describe the solution you'd like**
Add some way to see how the which outfits, and how many of each, are stored on a given planet. This should be accessible in-flight and while landed anywhere.

**Describe alternatives you've considered**
@petervdmeer on #4443 suggested:

> * Get an overview list in the player-info screen section of where (which) outfits are stored.

I think this information could be added to the outfitter section of the map display. The list of outfits, when a planet is selected, could show how many copies of that outfit are stored there. Also, a "Show only stored outfits" mode could be added alongside the "Show only outfits for sale in this system" mode, the yellow button here:

![2021-05-06-183306_126x54_scrot](https://user-images.githubusercontent.com/24784687/117348248-83d5fa80-ae99-11eb-9655-68f8bc845276.png)

**Additional context**
Here's a screenshot of the map/outfitter view. Under "8 tons of weapon space" there could be an additional line like "4 Beam Lasers are stored on this planet". I'm not sure what color the text should be in, for an outfit that's in storage but not in stock.

![2021-05-06-183516_433x299_scrot](https://user-images.githubusercontent.com/24784687/117348525-cf88a400-ae99-11eb-8cb9-25b93b3a2742.png)

jostephd:
> I think this information could be added to the outfitter section of the map display. The list of outfits, when a planet is selected, could show how many copies of that outfit are stored there. Also, a "Show only stored outfits" mode could be added alongside the "Show only outfits for sale in this system" mode, the yellow button here:

Just realized: the outfitter display is per-system, but the planetary storage is per-planet. So, simply adding this to MapOutfitterPanel won't let players know _which planet in the system_ the outfits are stored in. However, https://github.com/endless-sky/endless-sky/pull/4443#issuecomment-678489280 (which references #5210) makes it sound like the outfitter might one day start to treat the planetary storage as "per-system" storage (doesn't require a data file format change), so perhaps adding this to MapOutfitterPanel would be a good direction after all.

jostephd:
https://github.com/endless-sky/endless-sky/pull/5941

Issue #5210, opened 2020-08-05
Interplanetary shuttle? (Access to ships on other planets within the same system)


ira210:
When in a system with multiple planets and one or more of your ships is on a different one of them from you there are several unintended behaviours: firstly you can outfit ships not on the same planet as you even if the planet they are on has no outfitter or hasn’t got stock of the same outfits; secondly if you park all your ships then unpark one on a different planet in the same system there is no visual indication that you move planet however you can launch using a flagship on another planet and your flagship launches from that planet. This is easy demonstrable by going to the Sol system with at least two ships, parking all but one on mars then landing your flagship on earth or Luna, you will be able to access the ships on mars from the outfitter and shipyard and launch in one of them should you choose to. 
The way I think this should be resolved is by making any changes to a ships attributes while it is in the same system as the player automatically re-locate to the players planet. There could even be a tooltip for the first time this happens to say that travel between planets is easy and often done on whatever ship is available so if a ship is leaving one planet for another pilots often catch a lift and bring one or two passengers for free. 
As an alternative the game could be kept the way it is which is not bad, ideally with an added tooltip about the cheep was of in system travel and maybe even implementing a slight fee for doing this all though that is not necessary. It would also be possible to remove this entirely however this would not be the best option as it makes the game un-necessarily difficult and confusing. I don’t think this is worth more than half an hour of anyone’s time to fix although others may differ so if this seem hard don’t bother it isn’t two huge but would be nice for immersion to have at least a tooltip

vitalchip:
Confirmed behaviour as described in 0.9.11 and 0.9.12. I believe this should be fixed rather than explained away.

Amazinite:
#2375

> I don't see much point in fixing this. Especially because if taken to its logical conclusion, you wouldn't be able to outfit those parked ships either, even though you can outfit ships that were holding position in system (and not disabled) when you landed. This is another case of game play trumping realism.

tehhowch:
[To quote myself:](https://discord.com/channels/251118043411775489/536900466655887360/740550375316389919)
> It's an artifact of the game allowing you to work with ships as long as they are in-system, coupled with each ship knowing where it landed. The fix is to only allow choosing a flagship on your given planet, but this may be allowable for the same reason we allow outfitting ships that are landed/parked on the other in-system planets: it's to limit annoyance

[...]

PR #5942, opened 2021-05-07
feat(mechanics): Show stored outfits count by planet

jostephd:
**Feature:** This PR implements the feature request detailed and discussed in https://github.com/endless-sky/endless-sky/issues/5939#issuecomment-834559751

## Feature Details
Add a way to show the number of stored outfits by planet. In master, the map panel shows the number of stored outfits _by system_ (and #5941 doesn't change this), so there's no way to see which planets in a given system store outfits, or how many, without landing on them again.

## UI Screenshots
2021-05-07-171702_208x143_scrot

## Testing Done
Went to a system with two planets with outfitters, stored some outfits on each, took the screenshot.

(edit: Also test the case of having no flagship: https://github.com/endless-sky/endless-sky/pull/5942#pullrequestreview-668208190)

## Performance Impact
N/A


petervdmeer:
Instead of this change I prefer to see the Annoyance-avoidance feature implemented (that allows the player to access all stored outfits in a system from all planets in that system, similar to what is done for ships). And if we go that route, then I feel that we wouldn't need the change from this PR since it would not matter to players which planet the outfits are stored on, as long as they know which system they need to go to.

MasterOfGrey:
> that allows the player to access all stored outfits in a system from all planets in that system, similar to what is done for ships

Wait, you could re-outfit a ship on a different planet in the same system? That sounds hella hacky

petervdmeer:
> Wait, you could re-outfit a ship on a different planet in the same system?

Yes, that is possible.

> That sounds hella hacky

Code-wise it's just filtering the total ships list by `planet->System` instead of filtering on just `planet` when outfitting. Other parts of the code seem more hacky than this.

vitalchip:
Just for balance, I find the fact that you *can* outfit ships on other in system planets to be (mildly) annoying. If the flagship is on planet A and an escort on B it should not be possible to outfit/sell/change flagship.

jostephd:
> Instead of this change I prefer to see the Annoyance-avoidance feature implemented (that allows the player to access all stored outfits in a system from all planets in that system, similar to what is done for ships). And if we go that route, then I feel that we wouldn't need the change from this PR since it would not matter to players which planet the outfits are stored on, as long as they know which system they need to go to.

I wouldn't oppose this, but so long as the annoyance-avoided feature hasn't been implemented, I think this change is a reasonable stopgap. If it does get implemented (jury seems to be out on this, over on #5210 for ships?), I wouldn't be offended if this PR is reverted outright.

MasterOfGrey:
> Code-wise it's just filtering the total ships list by `planet->System` instead of filtering on just `planet` when outfitting. Other parts of the code seem more hacky than this.

To be clear, I meant this as being hacky in a gameplay sense, not in terms of the actual code itself.

source/MapPanel.cpp
```diff
@@ -1190,11 +1192,20 @@ void MapPanel::DrawTooltips()
...
 				tooltip += "\n";
-			tooltip += to_string(t.second) + (t.second == 1 ? " stored outfit" : " stored outfits");
+
+			unsigned sum = 0;
+			for(const auto& it : t.second)
+				sum += it.second;
+
+			tooltip += to_string(sum) + (sum == 1 ? " stored outfit" : " stored outfits");
+
+			if(t.second.size() > 1)
```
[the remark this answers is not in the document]
jostephd: Done in e4caa9aca02796801e43968b1f813b6a54eb28b9.

source/MapPanel.cpp
```diff
@@ -1173,11 +1186,12 @@ void MapPanel::DrawTooltips()
 	// Create the tooltip text.
 	if(tooltip.empty())
 	{
-		pair<pair<int, int>, int> t = escortSystems.at(hoverSystem);
+		pair<pair<int, int>, map<const Planet *, unsigned>> t = escortSystems.at(hoverSystem);
```
jostephd, line 1189: `escortSystems` only gets populated when the `Show escort systems on map` preference is set. Thus, the extra information about outfits is currently conditional on that preference, too. I'm not sure what the best way to fix this is.

petervdmeer:
> I wouldn't oppose this, but so long as the annoyance-avoidance feature hasn't been implemented, I think this change is a reasonable stopgap. If it does get implemented (jury seems to be out on this, over on #5210 for ships?), I wouldn't be offended if this PR is reverted outright.

Good to hear. I hope you also wouldn't be offended if the annoyance-avoidance feature gets implemented before this gets merged. (#5961)
(And #5961 is also not merged yet, so the devs might still choose to put this PR in 0.9.14 as stopgap and keep the annoyance-avoidance feature for 0.9.15.)

jostephd:
> Good to hear. I hope you also wouldn't be offended if the annoyance-avoidance feature gets implemented before this gets merged. (#5961)

Not in the least! If this PR becomes unnecessary, just close it. It won't be stepping on my toes or anything.

source/MapPanel.cpp
```diff
@@ -63,10 +63,10 @@ namespace {
 	//  - first        -> ships
 	//  - first.first  -> ships (in-flight)
 	//  - first.second -> ships (parked)
-	//  - second       -> outfits (in storage)
+	//  - second       -> map of planets to number of outfits on that planet
 	void TallyEscortsAndOutfits(const vector<shared_ptr<Ship>> &escorts,
 		const std::map<const Planet *, CargoHold> &outfits,
-		map<const System *, pair<pair<int, int>, int>> &locations)
+		map<const System *, pair<pair<int, int>, map<const Planet *, unsigned>>> &locations)
```
tehhowch, line 69: I think it is possibly more useful to display a tool-tip in the orbits scene, when the player hovers over the individual planet, rather than further expand the system-level tool-tip with planet-specific information.

source/MapPanel.h
```diff
@@ -125,7 +125,7 @@ class MapPanel : public Panel {
 	// for use in determining which governments are in the legend.
 	std::map<const Government *, double> closeGovernments;
 	// Systems in which your (active and parked) escorts and stored outfits are located.
-	std::map<const System *, std::pair<std::pair<int, int>, int>> escortSystems;
+	std::map<const System *, std::pair<std::pair<int, int>, std::map<const Planet *, unsigned>>> escortSystems;
```
tehhowch, line 128: If we keep this tooltip at the system level, I think it makes sense to introduce a nested class, e.g.

```c++
class SystemTooltipData {
public:
  unsigned parkedShips = 0;
  unsigned activeShips = 0;
  std::map<const Planet *, unsigned> outfits;
}
```
Then this field is just `std::map<const System *, SystemTooltipData>` instead of some unwieldy mess of `.first`s and `.second`s that is only comprehensible inside an IDE.

source/MapPanel.cpp
MasterOfGrey: Gross

jostephd: @tehhowch The orbits scene also shows unlandable planets, and isn't currently used for anything, so players might not realize it has a tooltip with useful information. In contrast, the system's tooltip is already used, and describes a clickable thing that even has graphical annotations that indicate when the tooltip will contain specific information. So I think putting the tooltip on the system is more useful than putting it on the orbits.

jostephd: @MasterOfGrey What is and why?

source/MapPanel.h
jostephd: As part of this PR, or as a separate PR?

source/MapPanel.cpp
petervdmeer: I like tehhowch idea. The argument that the orbits scene is not actively used is actually a good argument to start using it for more information like this.
We could also put a special marking on planets in the orbits scene that have stored items to make them more explicit/discoverable; maybe the same marker as [on systems](https://github.com/endless-sky/endless-sky/pull/4443#issuecomment-583619017)?

petervdmeer: For systems I would keep the total number of outfits stored in the pop-up, or maybe a listing of stored outfits per type (and then use tehhowch idea to give the per planet information).

jostephd: > The argument that the orbits scene is not actively used is actually a good argument to start using it for more information like this.

:thinking:

For discoverability, perhaps we could update [the help.txt "UI tutorial" messages](https://github.com/endless-sky/endless-sky/blob/13d22b84a8bd01070ab2705144511c7c69129c1f/data/help.txt#L69-L82) to point out the existence of orbits scene tooltips.

>  We could also put a special marking on planets in the orbits scene that have stored items to make them more explicit/discoverable;

Wouldn't this require a few more additions to the orbits scene, though? For starters, adding something to distinguish landable objects from other objects?

> maybe the same marker as [on systems](https://github.com/endless-sky/endless-sky/pull/4443#issuecomment-583619017)?

Should the "This system is a mission destination/stopover" markers also be applied to individual planets in the orbits scene, then? Especially in case you have missions to, say, both Luna and Earth? (And for that matter, it might be nice to have a subtle indication in the _system's_ mission destination markers that the destinations are not all on the same planet, but that's getting off-topic.)

Sorry if I sound like I have a mild case of featuritis. I'm just trying to run with @tehhowch's idea and see where it leads. If I had featuritis, I'd be proposing to use the orbits scene to indicate the government of each planet (#3768) :) 

> For systems I would keep the total number of outfits stored in the pop-up, or maybe a listing of stored outfits per type (and then use tehhowch idea to give the per planet information).

How about just the number of outfits and their total market value? Though there's a small can of worms here regarding _which_ value to show: the "As new" price, the depreciated-as-of-today sale price, or the depreciated-as-of-reaching-that-system sale price.

petervdmeer: The habitable (~= landable) planets already seem to stand-out with a blue color in the orbits scene.
Making the landable objects in the orbits scene match the "legend" selection (like showing friendly/hostile and showing governments when selected) sounds nice, but for a separate PR.
Adding the "mission destination/stopover" markers also to the orbit scene also sounds nice, but also for a separate PR.
Adding some tooltips could also be nice.

I would expect that storing outfits would typically be done to re-use them later. So when looking for stored outfits, then the total number and the exact types are interesting to a player. The value of the outfit(s) does not sound like something useful to display for stored outfits (especially since you can already lookup the "new price" value in the outfitter screen.)

jostephd: Mostly agreed, just one thing:

> I would expect that storing outfits would typically be done to re-use them later. So when looking for stored outfits, then the total number and the exact types are interesting to a player. The value of the outfit(s) does not sound like something useful to display for stored outfits (especially since you can already lookup the "new price" value in the outfitter screen.)

I proposed the number and total value of outfits as the tooltip for the system; just a summary so the tooltip doesn't get too crowded. The tooltip of the planet could give the exact list of outfits, even broken out by category (same categories as in the outfitter screen). Though it might be better to instead add a filter to the map outfitter view that shows stored outfits (like #5941 does, but without showing outfits for sale in the list at the same time; i.e., a filter that shows _only_ semidim items, rather than both semidim and dim items).

source/MapPanel.h
tehhowch: It would make sense to add a `SystemTooltipData` class when adding a `PlanetTooltipData` class. I think we all can agree that a map of nested pairs is not readable without IDE features.

source/MapPanel.cpp
petervdmeer: Sounds good, but I still feel that there is little value in showing the outfits total value here. Showing the total amount per planet (as done earlier in this thread) per system feels more valuable to me and there will usually not be more than 2 outfitters in a system anyway, so the risk of overcrowding also seems small.

source/MapPanel.h
petervdmeer: Yes, the nested classes look a lot more readable. The question from jostephd is however not yet answered, so hereby my view on that;
The readability of the affected code was not good before thisPR, but also gets a little worse by the change in this PR, so the addition of the nested classes should be done in a separate PR that gets merged earlier than this PR (which would give you a merge-conflict for this PR), or that should be addressed inside this PR (which would make this PR a little bit bigger).
I feel that the change is not too large, so combining with this PR is probably preferred from an Changelog/review perspective (but if you feel that the change is sufficiently large to deserve it's own PR, then please feel free to choose a separate PR instead.)

source/MapPanel.cpp
petervdmeer: Is rename to `Show escorts and stored outfits on map` an option?

source/MapPanel.cpp
jostephd: OK, so let's the keep the system tooltip as it stands, then. I'll look into implementing a per-planet tooltip, unless y'all prefer to merge the systems tooltip first and do the planets tooltip in a follow-up PR. (That might make sense if there's a release around the corner.)

source/MapPanel.cpp
jostephd: Combining escorts and stored outfits into a single setting sounds good; however:

- That string would overflow the space allotted to it in the preferences screen.

- If we rename the option, we'll need to teach the code to read the old name from existing preferences.txt files for backward compatibility. We might even continue to _write_ it, too, if we want to support downgrading to older versions of endless-sky without losing a change to the setting. (In theory, we could make this option the only one whose UI name is different to its preferences.txt name, but I don't think we want to do that.)

- In the longer term, introducing an abstraction between the UI name and the preferences.txt name of a setting wouldn't be the worst idea.

source/MapPanel.h
jostephd: Thanks, @petervdmeer. From a review perspective, I think the nested class refactoring should be a separate commit _after_ merging to master, which means it should be a separate PR, because this project squashes PRs when merging them. So I'll do a separate PR. But I'd prefer to do it as a _follow-up_ PR so I don't have to resolve conflicts, please. (I'm perfectly capable of rebasing and resolving conflicts, but it takes time, which I don't have an abundance of.)

Or if you prefer, I can squash this PR to two commits (one for the tooltips changes and one for the pairs/class refactoring) so it can be merged by rebasing rather than by squashing.

petervdmeer: Good considerations, but please keep in mind that me and most other reviewers also don't have an abundance of time available. If your approach would result in a PR that is still efficiently reviewable, then I don't see much issues with your approach. But if your approach would make this PR (more) difficult to review, then that could become a bottleneck to get this PR towards merging.

I don't have permissions to merge PRs on this repository, but I feel that you are better of creating 2 separate PRs if you feel strongly about the commits needing to be visible separately on master. (Rebase merges will take a little bit more effort from the devs and the devs already seem to be quite busy.)

tehhowch: It makes sense to me to implement both tooltip classes in the same PR. That said, if implementing orbits-based tooltips wouldn't be touching any of the same code, splitting it could make sense as there won't be much in the way of conflicts.

W.r.t. dealing with squash merges of a precursor feature branch, I've utilized [`git post-squash`](https://github.com/nomeata/git-post-squash) at work a few times to good effect.

source/MapPanel.cpp
MasterOfGrey: @jostephd sorry, my comment had been at tehhowch's suggestion of a tool-tip in the orbits scene

source/MapPanel.h
jostephd: > Good considerations, but please keep in mind that me and most other reviewers also don't have an abundance of time available.

Of course. I am not trying to save time for myself at the reviewers' expense. On the contrary, I proposed what, in my opinion, would _save_ time for reviewers.

> If your approach would result in a PR that is still efficiently reviewable, then I don't see much issues with your approach. But if your approach would make this PR (more) difficult to review, then that could become a bottleneck to get this PR towards merging.

Personally, I think putting the refactoring and the tooltip-adding in the same PR, and then squash-merging it, would make the resulting commit on master more difficult to review than if the two parts were separate commits on master. That's why I proposed what I did. However, it's not my call. If y'all prefer that both changes be made in the same PR, I can do that.

> It makes sense to me to implement both tooltip classes in the same PR. That said, if implementing orbits-based tooltips wouldn't be touching any of the same code, splitting it could make sense as there won't be much in the way of conflicts.

Implementing everything in the same PR feels to me a little like making the perfect the enemy of the good. Consider, there are several pieces to this puzzle:

1. Convert the nested pairs to a class.

2. Add systems tooltips.

3. Add orbits tooltips.

This PR does (2), and that part is basically ready (except that https://github.com/endless-sky/endless-sky/pull/5942#pullrequestreview-659178759 is outstanding). This PR doesn't add the new feature in (3), and neither improves on the existing technical debt in (1) nor regresses it. So if it were up to me, I'd have considered merging this PR as soon as https://github.com/endless-sky/endless-sky/pull/5942#pullrequestreview-659178759 were resolved. I mean, it's not like adding the systems tooltip before the other parts has any obvious downside, not that I can see, anyway, regardless of whether orbits tooltips do or don't touch the same code as the other parts.

Anyway, I've made my point. Please let me know how you'd like to proceed, whether I should add (1) and (3) to this PR, or whether this PR can be merged first and (1) and (3) done subsequently.

> W.r.t. dealing with squash merges of a precursor feature branch, I've utilized [`git post-squash`](https://github.com/nomeata/git-post-squash) at work a few times to good effect.

Thanks, but it sounds like a plain `git rebase --onto=foo bar baz` could do it just as well.

tehhowch: We're in a code freeze currently, waiting for a Steam release. No timeline on when it will lift, as MZ has time for the release whenever he has time for the release.

So if one approach lets you use your current free time to good effect, that's another argument for it.

petervdmeer: @jostephd You seem to have thought about this well. I also made my point and I don't have a strong preference, please feel free to choose the approach you feel is best.

jostephd: @tehhowch @petervdmeer Alright. I've added the refactoring to this PR, and partially squashed this PR so the refactoring is a separate commit from the tooltip adding. Thanks for the constructive discussion.

Next on my list is to figure out the preferences issue (https://github.com/endless-sky/endless-sky/pull/5942#pullrequestreview-659178759).

source/MapPanel.cpp
jostephd: So, the preferences screen looks like this:

2021-05-19-002708_261x249_scrot

The option names don't currently have tooltips.

The option names must fit on a single line. (They're drawn with Table::Draw(std::string&), which renders the string on a single line.)

This being the case, I think it'll be easiest to add a separate `Show stored outfits on map` preference. Renaming the preference, aside from the backward compatibility preferences.txt reading consideration, would require teaching `Table` to have multiline cells. That seems like it would be a lot of effort for little gain, or possibly even a lot of effort for _negative_ gain, should support for multiline cells make it more difficult to extend Table in other ways in the future.

jostephd: > This being the case, I think it'll be easiest to add a separate `Show stored outfits on map` preference.

Done and force pushed. Should hopefully be good to merge now? The orbits tooltips needn't block adding the systems tooltip, I think.

source/MapPanel.cpp
jostephd: > I think it is possibly more useful to display a tool-tip in the orbits scene, when the player hovers over the individual planet, rather than further expand the system-level tool-tip with planet-specific information.

Looking into this. The system tooltips are handled by MapPanel::Hover and MapPanel::DrawTooltips. The former is already marked `override` to enable MapSalesPanel::Hover. I guess I should add a MapDetailPanel::Hover, then. What should it do if the x,y coordinates are outside the orbits scene? Should it just `return false` so UI::Handle can loop on, or should it call MapPanel::Hover as MapSalesPanel::Hover does?

jostephd:
Summary: systems tooltips are implemented and (IMHO) ready to merge; looking into implementing per-planet tooltips in the orbits scene (https://github.com/endless-sky/endless-sky/pull/5942#discussion_r637606361).

source/MapPanel.cpp
```diff
@@ -1173,28 +1192,38 @@ void MapPanel::DrawTooltips()
...
 	if(tooltip.empty())
 	{
-		pair<pair<int, int>, int> t = escortSystems.at(hoverSystem);
+		MapPanel::SystemTooltipData t = escortSystems.at(hoverSystem);
+
 		if(hoverSystem == &playerSystem)
 		{
-			--t.first.first;
-			if(t.first.first || t.first.second || t.second)
+			--t.activeShips;
```
petervdmeer, line 1199: Not a bug you introduced, but I just realized that this can make activeShips negative if the player has no flagship. (And I just tested this, it shows -1 escorts if I open the map after I just sold my flagship with the original implementation.)

source/MapPanel.cpp
```diff
@@ -1173,28 +1192,38 @@ void MapPanel::DrawTooltips()
...
 				tooltip += "\n";
-			tooltip += to_string(t.second) + (t.second == 1 ? " stored outfit" : " stored outfits");
+
+			unsigned sum = 0;
+			for(const auto& it : t.outfits)
+				sum += it.second;
+
+			tooltip += to_string(sum) + (sum == 1 ? " stored outfit" : " stored outfits");
+
+			if(HasMultipleLandablePlanets(hoverSystem))
```
petervdmeer, line 1224: Maybe also add an OR-condition check if t.outfits size is larger than 1 (in case one of the planets becomes un-landable but still has outfits).

petervdmeer (approved):
Looks pretty good. I have 2 minor review-comments, but I don't consider them sufficiently critical not to approve.

source/MapPanel.cpp
jostephd: Good catch. Added without testing.

jostephd:
Both comments addressed, I think. Thanks again @petervdmeer for reviewing.

source/MapPanel.cpp
```diff
@@ -1173,28 +1192,39 @@ void MapPanel::DrawTooltips()
...
+
+			unsigned sum = 0;
+			for(const auto& it : t.outfits)
+				sum += it.second;
+
+			tooltip += to_string(sum) + (sum == 1 ? " stored outfit" : " stored outfits");
+
+			if(HasMultipleLandablePlanets(hoverSystem) || t.outfits.size() > 1)
+				for(const auto& it : t.outfits)
+					tooltip += "\n - " + to_string(it.second) + " on " + it.first->Name();
```
tehhowch, line 1227: This is pretty much the only part of the PR I don't particularly like, since it is putting the planet-specific data onto the system tooltip.

That said, it won't cause any crazy-looking tooltips in vanilla, since it requires X outfitters in the system to get X new lines (and we have much ghastlier UI issues with 3+ _landable_ objects).

I'm fine with this as a stopgap, but do want to see the per-planet system-level callout get removed in the follow up that augments the orbits UI.

source/MapPanel.cpp
```diff
@@ -102,6 +105,16 @@ namespace {
 	const int HOVER_TIME = 60;
 	// Length in frames of the recentering animation.
 	const int RECENTER_TIME = 20;
+
+	bool HasMultipleLandablePlanets(const System *system)
+	{
+		const auto &stellarObjects = system->Objects();
+
+		const auto &isRegularPlanet = [] (const StellarObject &stellarObject) {
```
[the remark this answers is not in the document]
jostephd: `const` because I always use that for variables that aren't written to. `&` because without that it always looks to me like it does a copy, but I'll drop it.

source/MapPanel.cpp
```diff
@@ -102,6 +105,16 @@ namespace {
...
+			return stellarObject.HasValidPlanet() && !stellarObject.GetPlanet()->IsWormhole();
```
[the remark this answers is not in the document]
jostephd: Added

source/MapPanel.cpp
```diff
@@ -102,6 +105,16 @@ namespace {
 	const int HOVER_TIME = 60;
 	// Length in frames of the recentering animation.
 	const int RECENTER_TIME = 20;
+
+	bool HasMultipleLandablePlanets(const System *system)
```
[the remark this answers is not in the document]
jostephd: Changed

jostephd:
Implemented some of the suggestions. The rest I don't have time to do right now, I'll do them later and update this PR.

source/MapPanel.cpp
petervdmeer: When this becomes an issue, then we can always change the tooltip to "You have <nr> outfits stored over <nr> planets in this system."
The issue is for now not very likely, because it will only be an issue when there are many outfitters in a system, and when the player decides to store outfits on many places.

source/MapPanel.cpp
```diff
@@ -102,6 +105,16 @@ namespace {
...
+	{
+		const auto &stellarObjects = system->Objects();
+
+		const auto &isRegularPlanet = [] (const StellarObject &stellarObject) {
+			return stellarObject.HasValidPlanet() && !stellarObject.GetPlanet()->IsWormhole();
+		};
+		return 1 < std::count_if(stellarObjects.begin(), stellarObjects.end(), isRegularPlanet);
```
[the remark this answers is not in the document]
petervdmeer: I uploaded a fix for the double count (that also caused the comparison to be removed).

source/MapPanel.cpp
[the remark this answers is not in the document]
petervdmeer: This code got replaced.

petervdmeer:
Merged master and processed open review comments.

source/MapPanel.cpp
```diff
@@ -102,6 +105,22 @@ namespace {
 	const int HOVER_TIME = 60;
 	// Length in frames of the recentering animation.
 	const int RECENTER_TIME = 20;
+
+	bool HasMultipleLandablePlanets(const System &system)
+	{
+		const Planet *firstPlanet = nullptr;
+		for(auto &stellarObject : system.Objects())
```
tehhowch, line 112:
```suggestion
		for(auto &&stellarObject : system.Objects())
```
No need to make a copy in order to have a mutable reference.

quyykk: I don't see the copy here, both lines are in fact functionally identical. Am I missing something?

PR #5946, opened 2021-05-08
Allow firing turrets and fixed guns separately

10010101001:
Allow binding different keys to firing turrets and fixed guns (can fire together if both actions are bound to the same key)


petervdmeer:
I would prefer to go a step further and get firing settings per turret and per gun (as discussed in https://github.com/endless-sky/endless-sky/issues/4714) and allow the player to set groups of weapons firing together.

10010101001:
That would be better for another PR

Pointedstick:
If we do this, I think we either need a setting to allow people to go back to the previous behavior, or else a keybinding to fire both guns and turrets. I'll admit there were times when I did want to choose which one got fired, but most of the time I want both blazing away. Endless Sky battles using non-missile weaponry are so fast and conducted as such short range that there's generally no time for this kind of micromanagement.

So I would probably leave the current keybinding for "fire everything" and then add two new ones: one for "fire guns only" and another one for "fire turrets only". Does that make sense to you?

Amazinite:
> If we do this, I think we either need a setting to allow people to go back to the previous behavior, or else a keybinding to fire both guns and turrets.

It's already possible to bind two actions to the same key (#5582), allowing for "fire everything" behavior.

Pointedstick:
OK, so if we do that, then both should be bound to the same key by default. :) I can't tell from code inspection if that's been done though.

petervdmeer:
> So I would probably leave the current keybinding for "fire everything" and then add two new ones: one for "fire guns only" and another one for "fire turrets only". Does that make sense to you?

For most of the ships I fly I do see a lot of value in being able to only fire a subset of the armanent installed on my ship. This would allow me to install a set of weapons that when fired together would overheat the ship or drain all energy, but that would be very practical when only ever fired in subsets.
The division over subsets would for me not be between guns and turrets, but is usually more between "long range weapons", "strong defensive weapons" and "short range disabling weapons". (Depending on whether I'm attacking from a distance, defending to survive or busy capturing/boarding some hostile ship.)

With things like non-forward facing guns and the many different types of turrets we have, it still doesn't make much sense to me to just differentiate between guns and turrets.
Assigning weapon hardpoints to fire on the "primary fire key" and/or to fire on "selections for secondary fire" still makes a lot more sense to me than adding another keybind.

Hecter94:
@10010101001 
Is this still something you're interested in working on? It looks like there were some changes requested and hasn't been much activity on it lately.

Amazinite:

TomGoodIdea:
unorphaning because of #11045

PR #5951, opened 2021-05-10
Fixed the description of Remnant: Cognizance 1

Ornok:
The description was telling the player to visit the planets in Pantica instead of Nenia.


Zitchas (approved):
Thanks for catching this! Using text replacement tags isn't always the best path.

PR #5953, opened 2021-05-10
Grammar fixes in coalition dscriptions and missions

Ornok:
This PR fixes grammar issues in coalition missions, planet and outfit descriptions.

In "Coalition Yottrite 8":
`the Arach who you brought several Yottrite samples to.` -> `the Arach to whom you brought several Yottrite samples.`

The other fixes are about the Coalition species substantives and adjectives (mostly about the Arachi). From what I've read so far, If I understand correctly, they can be summarized like this:

Noun (Singular)|Noun (Plural)|Adjective
---|---|---
Arach|Arachi|Arach
Kimek|Kimek|Kimek
Saryd|Saryds|Saryd

There are other occurrences of the plural noun used istead of the adjective for the Arachi that I haven't changed because they are in mission names: ~~`Arachi Rum`~~, `Arachi Orphans` and `Arachi Cartoonist` 1 and 2. Changing those would lead to compatibility issues.

**Edit**: In both cases, changing the mission names to `Arach Orphans` and `Arach Cartoonist` will lead to those missions being offered after the update, even if they were offered before, and even if they're still active.
In the case of `Arachi Cartoonist` 1 and 2, updating the name while `Arachi Cartoonist 1` is active will prevent  `Arachi Cartoonist 2` from being offered, which might confuse the player when completing `Arachi Cartoonist 1`.


Zitchas:
Which part are you referring to as the mission name?

The part that is labeled `name` is just a display name. That can be changed without impact on anything, as it is just what is displayed to the player.

On the other hand, the first line of the mission, which looks like this: `mission "Coalition: Contributor"`, yes, that will impact missions. Typically in simple mission structures where subsequent missions just rely on the previous one being marked "done", it will just result in the player getting to redo that mission. Not really much of a problem. If the mission is using, setting, or modifying other conditions, however, that could really mess things up.

Ornok:
> Which part are you referring to as the mission name?

In the original post I was referring to both, as I wasn't sure which one was marked as offered or done in the save file .
In my second post I called "display name" the `name` attribute and "actual name" the name given in `mission "<mission name>"`.

To sum it up::
- `Saryd Students 1 and 2`: Won't be affected since "Arachi Rum" only appears in the dispaly name.
- `Arachi Orphans`: is a single mission, so the only  issue is that it could be offered a second time if offered before the change.
- `Arachi Cartoonist 1 and 2` : 2 successive missions. Could be offered a second time, and making the change while "Arachi Cartoonist 1" is active will prevent "Arachi Cartoonist 2" from being offered when landing on Prime. From the player's perspective, the mission just disappears for no apparent reason.


Of course none of this is game breaking, and in the case of "Arachi Cartoonist" it will probably affect very few players (if any).

Ornok:
Alright, I think I've been overthinking this...
This last commit replaces `Arachi` by `Arach` everywhere it's needed except the mission names (in `mission <name>`). Since all these missions have a display name , the player will never see the actual name.

That means everything the player sees is fixed, and there won't be any compatibility issue, however small they may be.

Arachi-Lover (approved):
Nice catches.

Issue #5952, opened 2021-05-10
"Lagrange Ring: Cargo" mentions "the nearby system, Enif" even though it can be offered in Enif


Ornok:
In the `on offer` conversation of "Lagrange Ring: Cargo", the Quarg mentions "the nearby system, Enif". This sounds weird when the mission is offered on Lagrange in the Enif System.

Here is the complete sentence:
`Our ringworld, under construction in the nearby system, Enif, is to have a special open section for humans to visit soon.`

The simplest solution would be to exclude Lagrange as a possible source by using `source "Enif" 1 1` instead of  `source "Enif" 1`.
But since it makes sense to get the mission on the Lagrange station, I'd rather keep Enif as a source and rephrase that sentence.

movingpictures:
In the same mission, the language refers to "humane" materials. FYI.

PR #5955, opened 2021-05-11
Adjust phrasing in the mission Lagrange Ring: Cargo

Arachi-Lover:
----------------------
**Phrasing fix**

## Summary
Spotted by Ornok in https://github.com/endless-sky/endless-sky/issues/5952, the mission can offer on Lagrange itself, so I went with the second option they gave, rephrasing the bit that could pose an issue.
Don't suppose it's the best rephrasing as I just glanced at it for some seconds, so feel free to present better ones.


SubstandardZeal:
Another option is something like
> `Our <system> ringworld, under construction, is...`

PR #5956, opened 2021-05-11
Update a log entry for the optional FW Quarg mission

Arachi-Lover:
----------------------
**Logbook update**

## Summary
The Quarg never claim the Pug could defeat the Drak, and I think the part of the Quarg being there refers to the ringworld breaking, at which point it's not really a claim, but explaining how they were defeated once?
Either way, just updated the entries, cut mention of the Quarg you talk to claiming what it never claims.


Amazinite:
I'll note that I believe that MZ has said that log entries may provide added information that wasn't given within a mission conversation, which I believe was the intention here, but I have enough of an issue with the claim in itself that I think it's worth removing. Even if the Quarg thought that the Pug could defeat the Drak, that is 1) something I don't imagine the Quarg would actually admit to and 2) something that I think we should keep open ended instead of having factions outright claim that one can defeat the other, unless we have instances of some saying the Drak could beat the Pug, and others saying the Pug could beat the Drak, but in that case the Quarg should be making the former claim, not the latter.

PR #5960, opened 2021-05-13
Allow selective firing the Nullifier

10010101001:
Players will find it frustrating to have to manually hit the first shot with a Nullifier without auto aim. This issue can be addressed by making it a secondary weapon, like the Dragonflame.


Ornok:
I don't have any issue with aiming the Nullifier manually.
However, it can be a pain to use alongside other weapons if you don't use auto-fire. Because of its slow rate of fire, you need to hold your fire (for all primary weapons) while aiming to avoid wasting a shot. 

But I think that's a broader issue about being able to use weapons separately, which you're already trying to fix in #5946.

PR #5968, opened 2021-05-15
fix(content): various minor typos

infinitewarp:
**Bugfix:** This PR doesn't address any known specific issues. It just cleans up some typos I noticed in a recent playthrough.

## Fix Details
I've included three separate commits for three different minor text corrections. If you disagree with some but agree with others, I'll gladly rebase, drop or amend, and re-push with whichever parts you want to keep.

1. In most of the game text, whenever starting a sentence with a prepositional phrase, the established pattern is to follow that phrase by a comma before proceeding. 16ce09f8 adds a missing comma in one of these cases.
2. The description for "Starting Rubin" says the station is "on orbit" but I'm pretty sure this is meant to say it's "in orbit". The latter is much more typically used to describe the location of one thing orbiting another, whereas the former is far less common and is more often used when referring to an action that occurs while orbiting. 278fc0c2 changes "on" to "in".
3. In some of the "First Contact: Kuwaru Efreti" mission text, one of the Quargs is talking about the Korath. Where it says, "for they wield diverse weapons and strange," I believe it means to say that the _weapons_ are diverse and strange. Otherwise, "strange" isn't referring to anything. 201207da swaps the order to read "for they wield weapons diverse and strange" which I think reads much clearer while otherwise leaving the unusual sentence structure of the Quarg.

## Testing Done
Edited local text files and verified that the game still loads them normally.


Zitchas (approved):
Looks like a good set of fixes to me. Thanks for spotting these!

PR #5970, opened 2021-05-15
feat: Allow automatic determination of bay types (low performance and feature limited plugins only).

petervdmeer:
**Feature:** This PR implements the feature request to allow more bay-types, but only for plugins that can accept the performance hit and additional carried-ship limitations.

## Feature Details
The types of bays are no longer checked against their own list, but checked against the list of valid categories. This will in theory allow any ship-type to be used as carried ship, but it will also set performance issues and the carried-ship-limitations on any ship that could theoretically be carried somewhere.
This is not for usage with existing categories in the main/vanilla game due to the performance issues and carried-ship-limitations,  but this can allow some plugin writers to experiment with additional carried ship types.

PR #3502 is related in that that PR attempts to make the categories lists themselves configurable (which would allow new carried categories without the negative impact of making existing non-carried categories carried). I was actually looking at #3502 when I decided to create this PR.

## UI Screenshots
From a test-ship:
additional_bay

## Usage Examples
Set the category for some bay on some ship to a different category than "Fighter" or "Drone" and notice that the new category can be carried in the ship.

## Testing Done
Edited a save-game and replaced 2 "Fighter" bays by an "Interceptor" bay (and one by an "Anterceptor" bay to trigger the error-message).

## Performance Impact
I expect minimal impact (as long as this feature is not used for other carried categories than Fighter and Drone in the vanilla/main game). I don't know if c++ has optimizations for const-sets which we might be missing now that BAY_TYPES is no longer a const set.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -147,10 +147,7 @@ const vector<string> Ship::CATEGORIES = {
 
 
 // Set of ship types that can be carried in bays.
-const set<string> Ship::BAY_TYPES = {
-	"Drone",
-	"Fighter"
-};
+set<string> Ship::BAY_TYPES = {};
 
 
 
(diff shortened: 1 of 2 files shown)


Amazinite:
I actually have a fixed version of #3502 that I’ve been sitting on. What would the merits of this be over that?

petervdmeer:
> I actually have a fixed version of #3502 that I’ve been sitting on. What would the merits of this be over that?

How do you handle the validation of bays in your fixed version? Do bays have their own "bays" list of allowed values in "data/categories.txt"? Or do you validate bays directly against the "ships" list in "data/categories.txt"?

Amazinite:
Just posted the PR for you to check out: #5971

petervdmeer:
Closing; I prefer to see #5971 merged before this PR.

Issue #4610, opened 2019-10-20
Feature Request: Bay 


Zitchas:
**Is your feature request related to a problem? Please describe.**
Under the current system, any and all drones can be stored in any available drone bay, with the result that all drones in the game must be built to take into account that they may be stored in any drone bay. As a result, all drones operate under an artificially limiting criteria that realistically should not be applying to races in other parts of the galaxy. Currently, this has the practical effect that all drones (including Wanderer, Korath, and potential Remnant and future races) must fit in a 120ton box.

Despite the fact that humans consider drones to be highly expendable and sacrificial ships that are little more than a mobile gun platform, there is no reason that other races would have the same view. For example, it makes sense that a race that places a higher value on the life and welfare of its members might consider drones to be the front-line fightercraft instead of sacrificial glass-cannons; and build them accordingly. 

**Describe the solution you'd like**
I would like two optional attributes to be introduced for bays, one general and one specific.
- The first is the ability to specify that a bay is sized to fit up to a specified mass.
- The second is the ability to specify that a bay is limited to a certain category or categories.
- The third is the ability to specify that a bay is limited to a specific ship or ships.

The first one is the one that should be generally used, and it would limit the bay to accepting ships under a specified mass.

The second one would specify one or more categories of ships that could be accepted into a bay, and would commonly be used to differentiate between bays that have connections for human piloted ships and those that do not. (drone versus fighter)

The last would be fairly rare, specialty cases where a ship is designed with some kind of feature allowing it to carry a specific ship an no other ship could reasonably be able to.

Examples:
```
	bay -72 -46
              "bay mass" 80
```
One possible implementation, showing a drone bay that has a limit of 80 (which in practice limits it to the combat drone and surveillance drone, although a stripped down fighter might be able to fit).

petervdmeer:
Bays are already limited to a specific category, currently the category Fighter or Drone.

Bays for a specific ship or specific ships can also be implemented by putting a ship in a unique category and then creating a bay for that category. Change #4391 was created to allow ES to support more than 2 bay category types.

Placing additional restrictions on bays (like a maximum mass) will make the code for assigning carried ships to bays a lot more complex.
Imagine we have:
- One ship with a drone bay that has a mass restriction of 80
- One ship with a drone bay with no mass restriction
- One combat drone (with a mass of 80)
- One Korath drone (with a mass of 120)

[...]

PR #5971, opened 2021-05-15
Allow outfit and ship categories to be customized in the data files

Amazinite:
**Feature:** This PR is an updated version of the feature request detailed and discussed in PR #3502. The issue that that PR had was that whether a category was considered valid in its use by missions was subject to load order; the category would always need to be loaded by the mission that uses it. Thanks to the changes made by tehhowch, the check to see if a category is valid has been deferred until instantiation of a mission, so little had to be done in order to get this functional and I mainly just tidied up the code, so most of the credit still goes to @Hadron1776.

## Feature Details
This PR adds a new root object "category" that has three names that will be used: "ship", "bay type", and "outfit". The children of a category are then added to a list of the respective name that was provided. In the case of adding to the ship category, ships with a matching category attribute will then display in the shipyard under that category. The same is true with the outfit category, allowing outfits to display in the outfitter with that category. The bay type category is special in that it is what determines whether a ship can be carried or not, and therefore whether a bay can define that category as a carried ship.

Multiple plugins are capable of adding to the categories, as new definitions of a "category" don't overwrite the previous list. Names added to a category that already exist will have their placement in the list moved from wherever they previously were to the bottom of the list.

### Adding a new fighter type.

The following would add a new ship category "Gunship" and also allow it to be carried.
```
category "ship"
	"Gunship"

category "bay types"
	"Gunship"
```
<details><summary>Screenshots</summary>

image
</details>

### Definition order follows display order.

This is the default outfit category order. But what if we wanted to rearrange this?
```
category "outfit"
	"Guns"
	"Turrets"
	"Secondary Weapons"
	"Ammunition"
	"Systems"
	"Power"
	"Engines"
	"Hand to Hand"
	"Special"
```
<details><summary>Screenshots</summary>

image
</details>

The following is a rearranged category definition.
```
category "outfit"
	"Special"
	"Guns"
	"Secondary Weapons"
	"Ammunition"
	"Turrets"
	"Power"
	"Systems"
	"Engines"
	"Hand to Hand"
```
<details><summary>Screenshots</summary>

image
</details>

### Adding an already existing name.

Combining the load order mechanic with the mechanic that redefined names get moved to the end of the list, if we wanted the gunship category from the first example to appear above fighters and drones, instead or redefining all ship category names, we can just define the gunship category then redefine the fighter and drone categories so that they're shuffled down the list.
```
category "ship"
	"Gunship"
	"Fighter"
	"Drone"
```
<details><summary>Screenshots</summary>

image
</details>

## Testing Done
All the above usage examples were tested and work as expected. Alongside testing the gunship category, I changed a Skein to carry 5 gunships and 1 interceptor. The gunships could properly dock with the Skein, while the interceptor bays properly threw a warning about an undefined bay type.
<details><summary>Screenshots</summary>

image
image
image
image
</details>

## Performance Impact
N/A


source/ShipInfoDisplay.cpp
```diff
@@ -249,7 +249,7 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 	}
 	
 	// Print the number of bays for each bay-type we have
-	for(auto bayType : Ship::BAY_TYPES)
+	for(auto bayType : GameData::Category("bay type"))
```
petervdmeer, line 252: One of the reasons to use set as datatype for bay-types was to ensure that those categories were printed alphabetically. Do we still want to ensure printing in alphabetical order here?

source/LocationFilter.cpp
```diff
@@ -293,7 +293,7 @@ bool LocationFilter::IsValid() const
 	if(!shipCategory.empty())
 	{
 		// At least one desired category must be valid.
-		auto categories = set<string>(Ship::CATEGORIES.begin(), Ship::CATEGORIES.end());
+		const auto categories = GameData::CategorySet("ship");
```
tehhowch, line 296: this copies the set.

source/GameData.cpp
```diff
@@ -929,6 +931,23 @@ const string &GameData::Rating(const string &type, int level)
...
+{
+	return categories[type];
+}
+
+
+
+// Lists for categories as a set.
+const set<string> &GameData::CategorySet(const string &type)
+{
+	static const set<string> categorySet(categories[type].begin(), categories[type].end());
```
tehhowch, line 945: This doesn't work how you want. If the first call is for ships, the call for outfits will also return the result from ships.

It's likely simpler to not have this `CategorySet` method at all.

source/GameData.cpp
```diff
@@ -1136,6 +1155,19 @@ void GameData::LoadFile(const string &path, bool debugMode)
...
 				list.push_back(child.Token(0));
 		}
+		else if(key == "category" && node.Size() >= 2)
+		{
+			vector<string> &categoryList = categories[node.Token(1)];
+			for(const DataNode &child : node)
+			{
+				// If a given category already exists, it will be
+				// moved to the back of the list.
+				const auto &it = find(categoryList.begin(), categoryList.end(), child.Token(0));
```
tehhowch, line 1165: This binds a reference to an rvalue (the result of `find`). Instead, just accept the value:
```suggestion
				const auto it = find(categoryList.begin(), categoryList.end(), child.Token(0));
```

source/Ship.cpp
```diff
@@ -3174,7 +3159,7 @@ bool Ship::CanCarry(const Ship &ship) const
 
 void Ship::AllowCarried(bool allowCarried)
 {
-	canBeCarried = allowCarried && BAY_TYPES.count(attributes.Category()) > 0;
+	canBeCarried = allowCarried && GameData::CategorySet("bay type").count(attributes.Category()) > 0;
```
petervdmeer, line 3162: Converting to set looks a little bit inefficient for a single lookup, I suggest to use find directly on the vector (also on other places in this PR).
```suggestion
	bayCategory = GameData::CategorySet("bay type");
	canBeCarried = allowCarried && (find(bayCategory.begin(), bayCategory.end(), attributes.Category()) != bayCategory.end());
```

source/Ship.cpp
```diff
@@ -692,16 +665,28 @@ void Ship::FinishLoading(bool isNewInstance)
...
+	// invalid bays. Add a default "launch effect" to any remaining internal bays if
+	// this ship is crewed (i.e. pressurized).
+	string warning;
+	for(auto it = bays.begin(); it != bays.end(); ++it)
+	{
+		Bay &bay = *it;
+		if(!GameData::CategorySet("bay type").count(bay.category))
+		{
+			bays.erase(it);
+			--it;
```
tehhowch, line 678: `--it` does bad things when at the front of the vector. You'll need to have an empty 3rd statement in the for-loop initialization to avoid the double-increment when erasing.
```suggestion
			it = bays.erase(it);
```

tehhowch (changes requested):
We should use a scoped enum instead of strings for the types of categories, as I don't foresee the use of user-defined _types_ of categories.

source/Ship.cpp
```diff
@@ -692,16 +665,28 @@ void Ship::FinishLoading(bool isNewInstance)
...
-	// Add a default "launch effect" to any internal bays if this ship is crewed (i.e. pressurized).
-	for(Bay &bay : bays)
+	// Ensure that all defined bays are of a valid category. Remove and warn about any
+	// invalid bays. Add a default "launch effect" to any remaining internal bays if
+	// this ship is crewed (i.e. pressurized).
+	string warning;
+	for(auto it = bays.begin(); it != bays.end(); ++it)
+	{
+		Bay &bay = *it;
+		if(!GameData::CategorySet("bay type").count(bay.category))
```
petervdmeer, line 675: Should we also check allowed "bay types" against allowed "categories"? (To avoid having bays for which no ship type exists?)

Amazinite: I don't see it as necessary. A ship or outfit can technically have any category that it wants without causing any issues. The only purpose of ship and outfit categories is for displaying in a shipyard or outfitter. Bay types on the other hand are being used to define the gameplay features of a ship category. So while it likely wouldn't be the case in practice (as then you couldn't but the ship), it's conceivable to have a category that is undefined in the shipyard but defined as a carried ship. (Although I can imagine in practice wanting a one-off carried ship that you never plan to have in a shipyard and therefore never bother giving a defined ship category to, just the defined bay type category.)

Amazinite:
> We should use a scoped enum instead of strings for the types of categories, as I don't foresee the use of user-defined types of categories.

Would we also do the same for the ratings?

source/Ship.cpp
Amazinite: find_if?

source/ShipInfoDisplay.cpp
Amazinite: Was it particularly for the purposes of alphabetical ordered printing that a set was used, or was it just that Ship::BAY_TYPES was a set and therefore bays printed alphabetically? I think that data-defined ordering makes a bit more sense, because then bay categories can be given a "priority" in which they're printed akin to how ship categories are given priority in the shipyard. Although I do believe that the way this PR is set up right now, this will flip the way drone and fighter bays are ordered so that it lists fighters then drones instead of drones then fighters, but changing that is just a matter of flipping the definition order in the bay types category.

source/Ship.cpp
Zitchas: I like the sounds of that. So we could conceivably use this to have a "Fighter Bay" that can hold "Small Attack Craft" (which would include all fighters AND drones, and maybe a few other tiny ships too), but "Drone Bay" would still only fit drones.

That's a feature people have been asking for for ages.

Amazinite: That's not what this means at all. Ships can still only have a single category, and a ship's category must match the bay for it to be able to dock. Peter was asking if the category for the bay should also be a valid category in the shipyard in order for the bay to be valid, and I said that it shouldn't matter whether the category for the bay is a valid shipyard category.

tehhowch:
> > We should use a scoped enum instead of strings for the types of categories, as I don't foresee the use of user-defined types of categories.
> 
> Would we also do the same for the ratings?

We could, but that's outside the scope of this PR.

source/Ship.cpp
Zitchas: Oh. That's disappointing. It'd be nice if we could assign more than one category to ships.

tehhowch (commented):
Does it make sense to support outfitter / shipyard semantics, i.e. removing items from the lists of categories?

source/ShipInfoDisplay.cpp
petervdmeer: My main concern is that we make an explicit choice about the order in which bays are printed (in the ships overview). I originally made sure that the bays were ordered alphabetically, but if we today decide that the order is explicitly defined by the data-files then that is also fine.
(And yes, you might want to change the order in the data-file for Fighters and Drones.)

source/Ship.cpp
Zitchas: OK, then conversely, it'd be nice if a bay could be assigned multiple categories. I'm not picky about which way it happens.

Amazinite:
> Does it make sense to support outfitter / shipyard semantics, i.e. removing items from the lists of categories?

I don't think so. If a plugin doesn't want certain outfit or ship categories in the shop, they should just not sell items in those categories.

source/CategoryTypes.h
```diff
@@ -0,0 +1,24 @@
...
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+#ifndef CATEGORY_TYPES_H_
+#define CATEGORY_TYPES_H_
+
+enum class CategoryType : int {
+	SHIP,
+	BAY_TYPE,
```
tehhowch, line 18: Might as well just be `Bay` or `BAY`, since it will always be preceded by `CategoryType`

source/ShipInfoDisplay.cpp
```diff
@@ -249,7 +250,7 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 	}
 	
 	// Print the number of bays for each bay-type we have
-	for(auto bayType : Ship::BAY_TYPES)
+	for(auto bayType : GameData::Category(CategoryType::BAY_TYPE))
```
tehhowch, line 253:
```suggestion
	for(auto &&bayType : GameData::Category(CategoryType::BAY_TYPE))
```
Might as well fix the extra unconditional copy that was here.

tehhowch (commented):
My thought re: supporting add & remove is that some plugins (e.g. total conversions) may want to avoid displaying certain categories of ships or outfits.

Then again, I'm not sure allowing removal of the "bay type" members is a good idea.

Amazinite:
> My thought re: supporting add & remove is that some plugins (e.g. total conversions) may want to avoid displaying certain categories of ships or outfits.

I feel like it'd be more sensible to clear the shipyard and outfitter in that case rather than remove the category from being shown.

source/Ship.cpp
petervdmeer: @Zitchas You seem to be discussing feature request #4610, but that request looks very unrelated to this PR. The changes to support #4610 would also be quite different from the changes here, and there also doesn't seem to be consensus on the feature in #4610.
The feature in this PR is also requested often and there seems to be a lot of consensus on it, so let's let this PR keep its current focus.

source/ShipInfoDisplay.cpp
Amazinite: I've changed the bay ordering to alphabetical so that it's the same as it was before.

petervdmeer (approved):

PR #5972, opened 2021-05-15
Allow choices of phrases to have weights

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #4570.

## Feature Details
This PR allows choices of a phrase to have weights associated with them akin to how variants of a fleet can have weights (or from #4704, how ships of a subvariant can have weights).
Some vocabulary:
```
phrase "<name>" # This is a sentence of a phrase.
	word # This is a part of the sentence.
		"<word>" # This is a choice of the part.
	phrase # This is also a part.
		"<phrase>" # This is also a choice.
```

## UI Screenshots
N/A

## Syntax + Usage Examples

The syntax for this mechanic is as follows:
```
phrase "<name>"
	word
		"<word>" <weight>
		...
	phrase
		"<phrase>" <weight>
		...
	...
```

The following would be a pirate phrase with the choices weighted according to how many potential outcomes each subphrase can have, resulting in this phrase having an equal change to choose any name it can generate.
```
phrase "pirate"
	phrase
		"pirate adjectives that don't work before singular nouns" 15
		"pirate adjectives that don't work as titles" 66
		"pirate adjectives that work as titles" 50
	word
		" "
	phrase
		"sinister names" 70
		"bad outcomes" 50
```

## Testing Done
Created a new phrase "testing" that chooses from one of two phrases, the first with one name and the second with four.
```
phrase "testing"
	phrase
		"the first one"
		"the second one"

phrase "the first one"
	word
		"Three Skeleton Key"

phrase "the second one"
	word
		"Meme School"
		"Blackbeard"
		"Literally Hitler"
		"Figuratively Hitler"
```
I then spawned 1000 Sparrows from a mission using the "testing" phrase for their names and counted up the results.
I then repeated this test with 1000 Sparrows using weighted choices.
```
phrase "testing"
	phrase
		"the first one" 1
		"the second one" 4
```
This was the result of using the unweighted and the weighted name phrase. Expected results are in parentheses. There is still randomness in which outcome gets chosen, but the outcomes are clearly more evenly weighted than previously.

Name                | Unweighted | Weighted
--------------------|------------|----------
Three Skeleton Key  | 483 (500)        | 210 (200)
Meme School         | 118 (125)        | 194 (200)
Blackbeard          | 152 (125)        | 194 (200)
Literally Hitler    | 114 (125)        | 200 (200)
Figuratively Hitler | 133 (125)        | 202 (200)

## Performance Impact
N/A


source/Phrase.cpp
```diff
@@ -151,6 +153,18 @@ Phrase::Choice::Choice(const DataNode &node, bool isPhraseName)
...
 
+// Pick a random choice from this part given their weights.
+const Phrase::Choice &Phrase::Part::Get() const
+{
+	unsigned index = 0;
+	for(int choice = Random::Int(total); choice >= choices[index].weight; ++index)
+		choice -= choices[index].weight;
+	
+	return choices[index];
+}
```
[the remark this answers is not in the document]
Amazinite: It is something I’d thought of, so I’m certainly not opposed to the idea.

source/Phrase.cpp
```diff
@@ -108,7 +108,9 @@ Phrase::Choice::Choice(const DataNode &node, bool isPhraseName)
 	// The given datanode should not have any children.
 	if(node.HasChildren())
 		node.begin()->PrintTrace("Skipping unrecognized child node:");
-
+	
+	weight = (node.Size() >= 2 ? node.Value(1) : 1);
+	
 	if(isPhraseName)
```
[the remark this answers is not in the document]
Amazinite: DataNode::Value already validates that the token requested can be converted to a value and will return 0 and print a trace if it isn't, so I've simply clamped negative values to 1.

source/Phrase.cpp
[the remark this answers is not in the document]
tehhowch: It would make some level of sense to do so here, as you're taking code used in one place and using it in another. Why defer testing that code until after it's in more widespread use? (Deferring tests until later is a great way to never write tests at all.)

On the other hand, refactoring the fleet variant-picking code to use it is something that could easily be deferred to a follow-up PR.

source/Phrase.cpp
```diff
@@ -108,7 +108,9 @@ Phrase::Choice::Choice(const DataNode &node, bool isPhraseName)
 	// The given datanode should not have any children.
 	if(node.HasChildren())
 		node.begin()->PrintTrace("Skipping unrecognized child node:");
-
+	
+	weight = (node.Size() >= 2 ? min(1, static_cast<int>(node.Value(1))) : 1);
```
tehhowch, line 112:
```suggestion
	weight = max<int>(1, node.Size() >= 2  ? node.Value(1) : 1);
```

source/Phrase.cpp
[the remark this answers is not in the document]
Amazinite: I imagine this will be merged before #4704, but so long as that PR is something we want merged, I'll probably just handle Fleet (and Variant) using WeightedList in that PR, since I'd need to update that PR to use it regardless.

Amazinite:
Retested the outcome of 5000 names using the same setup as described in the PR description after the WeightedList change.

Name | Unweighted | Weighted
Three Skeleton Key | 2453 (2500) | 975 (1000)
Meme School | 681 (625) | 1059 (1000)
Blackbeard | 652 (625) | 985 (1000)
Literally Hitler | 630 (625) | 1005 (1000)
Figuratively Hitler | 584 (625) | 976 (1000)

source/WeightedList.h
```diff
@@ -0,0 +1,76 @@
...
+	
+	typename std::vector<Type>::iterator begin() { return choices.begin(); }
+	typename std::vector<Type>::const_iterator begin() const { return choices.begin(); }
+	typename std::vector<Type>::iterator end() { return choices.end(); }
+	typename std::vector<Type>::const_iterator end() const { return choices.end(); }
+	
+	void clear() { choices.clear(); };
+	int size() const { return choices.size(); };
+	bool empty() const { return choices.empty(); };
+	Type &back() const { return choices.back(); };
```
tehhowch, line 40:
```suggestion
	const Type &back() const { return choices.back(); };
```
Provide a non-const version as well

source/WeightedList.h
```diff
@@ -0,0 +1,76 @@
...
+	
+	template <class ...Args>
+	void emplace_back(Args&&... args);
+	
+	
+private:
+	mutable std::vector<Type> choices;
```
tehhowch, line 47:
```suggestion
	std::vector<Type> choices;
```
Mutable implies that this class can modify its list of options at will and it makes no difference to the class's external state. I would disagree with that implication, as changing the options changes the core state of this class.

source/WeightedList.h
```diff
@@ -0,0 +1,76 @@
...
+	typename std::vector<Type>::iterator end() { return choices.end(); }
+	typename std::vector<Type>::const_iterator end() const { return choices.end(); }
+	
+	void clear() { choices.clear(); };
+	int size() const { return choices.size(); };
+	bool empty() const { return choices.empty(); };
+	Type &back() const { return choices.back(); };
+	
+	template <class ...Args>
+	void emplace_back(Args&&... args);
```
tehhowch, line 43: Have this return `Type &` (which is the signature of emplace_back in newer versions of the standard library).

source/WeightedList.h
```diff
@@ -0,0 +1,76 @@
...
+	
+	
+private:
+	mutable std::vector<Type> choices;
+	int total = 0;
```
tehhowch, line 48: I think it is fair to use an unsigned type when negative values are not sensible at all
```suggestion
	std::size_t total = 0;
```

source/WeightedList.h
tehhowch, line 40: Specify the noexcept-ness of this code (you can look at what `<vector>` declares for these operations.)

source/WeightedList.h
```diff
@@ -0,0 +1,76 @@
...
+
+
+template <class Type>
+template <class ...Args>
+void WeightedList<Type>::emplace_back(Args&&... args)
+{
+	// Type is responsible for all weights being >= 1.
+	choices.emplace_back(args...);
+	total += choices.back().weight;
+}
```
tehhowch, line 60: And if they don't, we should throw some kind of "bad argument" exception or not add the input.

tehhowch (commented):
Looking good!

Template class also needs unit tests 🙂

tehhowch (commented):
The [Catch2 tutorial](https://github.com/catchorg/Catch2/blob/v2.x/docs/tutorial.md#test-cases-and-sections) might explain what I mean in the tests better, about when things are run.

Amazinite:
Should I change it so that the weight of a type is accessed by a function instead of a public variable? (i.e. choice.Weight() instead of choice.weight.) Right now we'd only be using this with phrase choices and fleet variants, which both have public variables named weight, but changing to accessing by a function would expand the potential uses of WeightedList, and phrase choices and fleet variants could still have their public variable, we'd just be accessing weight through a function for the purposes of the WeightedList.

source/Phrase.cpp
```diff
@@ -108,7 +107,9 @@ Phrase::Choice::Choice(const DataNode &node, bool isPhraseName)
 	// The given datanode should not have any children.
 	if(node.HasChildren())
 		node.begin()->PrintTrace("Skipping unrecognized child node:");
-
+	
+	weight = max<int>(1, node.Size() >= 2  ? node.Value(1) : 1);
```
Amazinite, line 111:
```suggestion
	weight = max<int>(1, node.Size() >= 2 ? node.Value(1) : 1);
```

PR #5974, opened 2021-05-15
Expand the number of normal swizzles from 7 to all possible 27

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #1779.

## Feature Details
As the title suggests, this PR expands the number of normal swizzles from 7 to 27.
Swizzles work by swapping the RGB color channels of an image. While there are 6 permutations of RGB that contain all three colors and therefore the highest color space that are all available in game, there are 21 more permutations of RGB that make use of less color channels but none the less result in a sensible set of colors on a ship. See https://github.com/endless-sky/endless-sky/issues/1779#issuecomment-256505771 for MZ's explanation on this (and why he doesn't think we should have more swizzle, but it's notable that the pirate swizzle already doesn't contain all three color channels, so I'd say that simply expanding it to the rest and letting content creators decide what they prefer isn't too far fetched, although I think we'll certainly try to avoid blander looking swizzles for vanilla content).

In any case, this is a feature that repeatedly gets requested, has been attempted multiple times before, and is really no issue to implement. Let's just get it over with.

## Screenshots
I have no idea what order these are in according to the numbers for each swizzle in this PR because I made this image forever ago for #1779, but this image gives you a glimpse at what the various new swizzles might look like on a Bactrian.
image

## Usage Examples
This just allows the `swizzle` on a government or ship to go from 0-8 to 0-28. Go wild.

## Testing Done
Set the `swizzle` tag on a ship to various values and confirmed that the ship has the correct colors for the swizzle. Cloaking also still works, which makes use of swizzle 27.

## Performance Impact
N/A


Zitchas:
Awesome!

Rob59er:
A small scattering of the new swizzles, they look good.
image
image
image
image

unjown:
hmmmmmmmmmmmmmmmm why we not merging?

MCOfficer:
> 
> 
> hmmmmmmmmmmmmmmmm why we not merging?

Apparently no code-savy maintainer has had time to review it yet. Also we have a stable release coming up very soon, and we try to limit high-impact changes for these, so I'm not sure if we want to hurry this into the release this quickly. Patience.

Zitchas:
Well, this strikes me as being something that would fall under the "big code change" rubric, and thus not something we want to push into a "stable" release anyway. But hopefully it gets merged pretty quickly once we're on the 0.9.15 development.

petervdmeer (approved):
The code-change looks consistent and correct. I didn't test this myself, but I feel that the tests done by Amazinite and Rob59er are sufficient for this feature.

PR #5975, opened 2021-05-16
Created four new damage types that deal DoT to shields, hull, fuel, and heat

Amazinite:
**Feature:** This PR implements a feature request discussed on Discord.

## Feature Details
This PR adds four new damage types that deal damage over time, akin to ion damage, against hull, shields, fuel, and heat.

* `"discharge damage"`: deals damage over time to a ship's shields. Unlike other special damage types, discharge damage is 100% effective against a ship regardless of shield level.
* `"corrosion damage"`: deals damage over time to a ship's hull. Requires that a ship's shield fraction be >0% to deal damage. (i.e. the shields must be down, the weapon must be piercing, or the ship must be disrupted.)
* `"leak damage"`: deals damage over time to a ship's fuel. Requires that a ship's shield fraction be >0% to deal damage.
* `"burning damage"`: deal damage over time to a ship's heat.
* Each of these damage types have also been given their respective protection, resistance, and resistance cost attributes. These damage types are affected by the shields being up just like other status effects and special damage types are.

The following table highlights various ship stats along with the instantaneous and DoT damage types that effect them. Those in italics are the ones that were missing but have been added.

"Stat" effected | Instantaneous damage | Damage over time
--|--|--
Shields | shield damage | *discharge damage*
Hull | hull damage | *corrosion damage*
Energy | energy damage | ion damage
Fuel | fuel damage | *leak damage*
Heat | heat damage | *burn damage*
Speed | hit force? (a bit of a stretch) | slowing damage
Shield integrity | piercing | disruption damage

## UI Screenshots
N/A

## Usage Examples
These are used as any other damage type might be.

## Testing Done
Shot at some ships with these damage types on weapons. Work like a charm.

## Performance Impact
N/A

## PR Checklist
N/A


BlazingDiesel:
In regards to corrosion damage, I would have assumed by the name alone that it would only do damage on the hull and therefore not work if shields are active. I imagined that "corrosion" by definition would have to be applied on some physical substance like metal and so would only affect the hull. This is my take on it, anyway.

Amazinite:
An alternative take would be that while a ship's shields are up, corrosion damage doesn't stack, but once the shields are down corrosion damage is able to stack up. Then even if the shields return, the corrosion damage that remains is still able to damage the ship.

Zitchas:
I was thinking the same thing: Corrosion should have no effect when used on a ship with its sheilds up. However, If it can be applied to the ship without shields (either the shields are down or using a weapon to open up a hole to get it through) then it would continue to work while shields are up. In other words, shields block it from being applied, but don't block it from continuing to work.

Alternative idea: What if corrosion didn't actually do any damage, but instead made the hull more vulnerable to being damaged?

EjoThims:
Would piercing or disruption automatically scale corrosion then? Or would that require extra coding?

arkhne:
Corrosion has already been discussed, so I'll skip that one.
Personally, I think leak damage should also require the hull to be damaged by the attack that inflicted it. Unless it's magically teleporting black holes inside the ships, in which case it should probably ignore shields anyway.

Amazinite:
Corrosion and leak damage now function similarly to hull damage. That is, you only receive hull damage if your shields are down, you've been hit by a piercing weapon, or if you have shield disruption.

petervdmeer:
Would you mind to include https://github.com/endless-sky/endless-sky/pull/4404 in this PR? (I would like to see 4404 get to merged, but I'm afraid that the original author has given up on that PR.)

Amazinite:
This PR is specifically about new damage over time types, so I likely won't be including that PR. I had asked a few people previously if they'd like me to pick that PR back up, though, so if there's interest in it I'd certainly fork off of that PR and finish it up like I did with the customizable categories.

Amazinite:
I'm up for merging this without any new graphics to allow people to make use of the new damage types, but should they be implemented in vanilla in the future we'd require new graphics for them. (#3599 could frankly be merged as well for similar reasons.)

source/Ship.cpp
```diff
@@ -3160,10 +3237,28 @@ int Ship::TakeDamage(vector<Visual> &visuals, const Weapon &weapon, double damag
...
+	// will clamp it to a maximum value at the beginning of the next frame.
+	energy = max(0., energy);
+	fuel = max(0., fuel);
+	heat = max(0., heat);
+	
 	// Recalculate the disabled ship check.
 	isDisabled = true;
 	isDisabled = IsDisabled();
 	
+	// Inflicted heat damage may also disable a ship, but does not trigger a "DISABLE" event.
```
petervdmeer, line 3253: The DISABLE event is below on line 3254, so it looks like overheating now does trigger the DISABLE event.

data/tooltips.txt
```diff
@@ -23,6 +23,21 @@ tip "atmosphere scan:"
...
+tip "burn resistance:"
+	`How much burn damage this ship can resist per second.`
+
+tip "burn resistance energy:"
+	`Energy consumed per second when resisting burn damage.`
+
+tip "burn resistance fuel:"
+	`Fuel consumed per second when resisting burn damage.`
+
+tip "burn resistance heat:"
```
petervdmeer, line 38: Should we add some checks that if the burn damage is less than the generated heat for resisting, that we then don't apply the resistance? (Same for fuel / leak)

data/tooltips.txt
```diff
@@ -536,6 +608,18 @@ tip "firing slowing / second:"
 tip "firing disruption / second:"
 	`Disrupts the ships shields after firing this weapon, allowing damage from hostile weapons to "leak through" to the hull even if shields are up. Shield disruption wears off over time.`
 
+tip "firing discharge / second:"
+	`Amount of shield damage over time by this weapon per second when firing.`
```
petervdmeer, line 612: Reduces shields for some time after this weapon is fired. The effect wears off over time. (also for the other texts)

petervdmeer:
Btw, good to keep "hit force over time" for a separate PR. I would expect such hit-force over time damage to be a vector, not a scalar (i.e. you get imact from the weapon on the left side of your ship and you keep having a hit force-over-time towards the right for some time, and impact with hit force from the other side would (partially) cancel out the first).

source/OutfitInfoDisplay.cpp
```diff
@@ -90,6 +102,10 @@ namespace {
 		{"ion resistance", 2},
 		{"disruption resistance", 2},
 		{"slowing resistance", 2},
+		{"discharge resistance", 2},
+		{"corrosion resistance", 2},
+		{"leak resistance", 2},
+		{"burn resistance", 2},
```
petervdmeer, line 108: I like all the new damage types, they really add something to the game, but I'm a bit worried about all the protection/defense attributes. Can we explain a casual player the difference between `"discharge protection"`, `"discharge resistance"`, `"disruption protection"`, `"disruption resistance"`, `"piercing protection"`, `"piercing resistance"` and `"shield protection"`? (and this is just the set of attributes for shield-related defenses/protections, the player also needs to balance against all the available outfits with protections for hull, fuel, heat, energy and balance against weapons and steering/forward thrust when outfitting ships.

data/tooltips.txt
Amazinite: Remember that burn damage decays with time such that a burn damage of 1 applies 100 heat over time, so if any such check were to be made then we would want to not apply resistance if `burn resistance heat > 100 * burn resistance`, as any amount of heat applied for resisting burn < 100 heat / 1 burn is actually less heat overall. Such a check doesn't really seem necessary to me, though. Whether a burn resistance outfit is actually worth it should be down to implementation.

source/OutfitInfoDisplay.cpp
Amazinite: > Can we explain a casual player the difference between...

Well to be honest, I expect that by the time a player even has to worry about juggling every possible resistance or protection attribute, or even just a reasonable enough subset, that they'll be knowledgeable enough in the game's mechanics to know what's best for what occasion, as these aren't exactly basic attributes and should really only be implemented on advanced tech that takes some time to get ahold of; the only such defensive stat even available to the player in game right now is ion resistance on Remnant generators (with ion protection being on the hulls of Heliarch ships, but that isn't shown to the player in game).

petervdmeer: I consider myself to be an experienced player and I just don't "feel" the protection and resistance attributes; It feels to me "out of lore" as if we took the various properties and just put the words "protection" and "resistance" behind it. I'm particularly unhappy with the "resistance" word, since it sounds like a "static" property to me, while it is used as an "active" property (that costs fuel, energy and generates heat).

Is it possible to get some more variation in the terminology? For example by:
- Renaming `"burn protection"` to something like `"thermal insulation"` (it is passive and reduces the effects of burning/heating without actively countering heat.)
- Renaming `"burn resistance"` to something like `"fire suppression"` or `"flame suppression"` (or some other term that we refer to when we talk about sprinklers in buildings or systems in cargo-planes to automatically put out fires).
- Keeping either `"corrosion protection"` or `"corrosion resistance"` for the passive-property (I like resistance better there) and find something more appropriate for the "active anti-corrosion" (maybe `"hull ablation"`, `"hull decontamination"` or just `"decontamination"`)?
- Rename `"leak protection"` to something like `"puncture resistance"` (as is often used for tires).
- Rename `"leak resistance"` to something like `"self sealing"`, `"leak repair"` or just `"sealing"`.
- Rename `"discharge protection"` to something like `"shield stability"`. (With as tooltip: "Higher shield stability will reduce the initial effects of discharge damage.")
- Rename `"discharge resistance"` to something like `"shield stabilization"`? (With as tooltip: "Higher shield stabilization will allow a ship to recover faster from discharge effects.")

petervdmeer:
> Btw, good to keep "hit force over time" for a separate PR. I would expect such hit-force over time damage to be a vector, not a scalar (i.e. you get imact from the weapon on the left side of your ship and you keep having a hit force-over-time towards the right for some time, and impact with hit force from the other side would (partially) cancel out the first).

Thinking about this a bit more gives some more gaps in your table:

Stat Affected: _"Ships Position"_
Instantaneous damage: "hit force"
Damage over time: _"displacement (force)"_
The Displacement force could be thought of a projectile that attaches to a ship as if it is an additional thruster (and keeps pushing the ship in the direction of where the impact was).


Stat Affected: "Speed"
Instantaneous damage: _"stopping force"_
Damage over time: "slowing damage"

source/OutfitInfoDisplay.cpp
Amazinite: I’d rather not have the names for attributes be different for each damage type despite the mechanics used being the same. Right now all one needs to understand is how protection works and how resistance works, then look to the attribute name to see which damage type each applies to. If the attribute names were unique per damage type then it’d be an exercise in figuring out which unique attribute name uses which mechanic for which damage type.

data/tooltips.txt
petervdmeer: Good point.

source/OutfitInfoDisplay.cpp
petervdmeer: I would expect that we create a table somewhere for content-creators to show how certain damages relate to certain defensive attributes. (Similar to the table in your PR, but with the protection and resistance attributes added and things like "cooling" and "active cooling" also added in the heat row.)

For players I would actually expect them to look at Outfits (not at stats directly) and there I expect the player to be more immersed in the game if the defensive attributes are not directly named after the damage that they counter. (In the same was that `"cooling"` sounds much better to me than `"heat removal"`, even though `"heat removal"` would be a better name for the property when it comes to game mechanics).

data/tooltips.txt
```diff
@@ -641,6 +737,18 @@ tip "firing slowing / shot:"
 tip "firing disruption / shot:"
 	`Each shot disrupts the target's shields, allowing hostile weapon damage to "leak through" to the hull even if shields are up. Shield disruption wears off over time.`
```
tehhowch, line 738: Doesn't this stat affect the current ship, not the target?

data/tooltips.txt
```diff
@@ -641,6 +737,18 @@ tip "firing slowing / shot:"
...
+	`Amount of shield damage over time by this weapon per shot when firing.`
+
+tip "firing corrosion / shot:"
+	`Amount of hull damage over time by this weapon per shot when firing.`
+
+tip "firing leak / shot:"
+	`Amount of fuel damage over time by this weapon per shot when firing.`
+
+tip "firing burn / shot:"
+	`Amount of heat damage over time by this weapon per shot when firing.`
```
tehhowch, line 750: These should maybe parrot the not-over-time tooltips more closely, e.g. "Heat damage incurred per shot."

tehhowch (commented):
Minor nits.


@petervdmeer I would consider "impulse" a better name for force-over-time damage. I don't think that we should consider anything as "damaging a ships position in space"; rather we have velocity damage (hit force, "impulse", and slowing).
I also don't think it makes sense to have damage that specifically targets the length of the ships velocity vector when we already have an instantaneous damage for velocity (hit force) that will simultaneously "damage" the speed.

data/tooltips.txt
Amazinite: Do these look good now?

source/Ship.h
```diff
@@ -519,6 +519,10 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	double ionization = 0.;
 	double disruption = 0.;
 	double slowness = 0.;
+	double discharge = 0.;
+	double corrosion = 0.;
+	double leakage = 0.;
+	double burning = 0.;
```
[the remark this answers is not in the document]
Amazinite: Is that good?

tehhowch:
> The main open for this PR seems to be the naming of the protection and resistance attributes. Do you also have a suggestion there on how to proceed?

We should maintain a uniformity of "processing" around these items, as that makes it significantly easier to move to a configuration in which we dynamically register these kinds of attributes without needing to hard-code aliases.

So e.g. `${attr} energy`, `${attr} heat`, ... and so on could be dynamically queried if we implement a system in which the data files register values of `attr` and the scenarios in which they can appear.
As previously mentioned, consistent naming also improves teachability and discovery, especially for those whose primary language is not English.

petervdmeer (approved):
Ok, let's get this merged then.

tehhowch (approved):

Issue #1150, opened 2016-05-27
escape pod flotsom


wambat1:
there has been the suggestion that escape pods be added to ship destruction animations. would it be possible to make these more than animation? make them a kind of collectible flotsam? the escape pod could become a kind of special commodity, with planets paying bounty or compensation based on the type of crew. each faction would produce its own escape pod commodity, and each faction would value each escape pod commodity differently. these could be hidden from the normal commodity list.

in the alternative, current commodities could be used, but merely have their name changed, similar to mission commodities. escape pods likely to be more valuable, like rich paradise world passengers, would equate to luxury goods, while pirate slaves would be equivalent to bulk food, and others falling in-between. that way variable prices are already assigned so less work, maybe.

jafdy:
Escape pods is on the development plan so people can play a "hardcore" type mode. 
As for this, I think it would be well worth adding to the game at some point.

toilethinges:
Adding to this idea, since it's now on the projects page:

> in the alternative, current commodities could be used, but merely have their name changed, similar to mission commodities. escape pods likely to be more valuable, like rich paradise world passengers, would equate to luxury goods, 

Perhaps outfits defined in the fleet data, so for example, "Large Republic" drops crew "republic".

Amazinite:
Noting related issues:
https://github.com/endless-sky/endless-sky/issues/362
https://github.com/endless-sky/endless-sky/issues/654
https://github.com/endless-sky/endless-sky/issues/1150
https://github.com/endless-sky/endless-sky/issues/5845
https://github.com/endless-sky/endless-sky/pull/5977

Issue #362, opened 2015-12-03
hardcore mode: escape pods, permanent death, etc.


Nikwigi:
I posted these in steam as well. but I thought here might be good too. 

Auto landing options
Back in evn some of the most fun was the stress of land while under fire. that sense of relief after making it. auto landing takes much of that threat away. maybe a simple toggle in settings for this would be nice.

Resale price settings
I'd like to see loss when you sell your ships and such. Once you get a nice ship it gets pointless fast capturing pirate ships and selling them back at full price. Also buying and selling your ships to get to this point has no sense of risk if you buy a ship at any point you simply sell as you move your way up the list. 
Might be part of the reason why Escape Velocity only let you own one ship.
I'd love to be able to pick a percentage like easy normal hard. Something set at the start would be best so you cant change it as you please. Maybe just a slider bar. from 10%-100%. This would add point to the more hardcore players. Maybe two sliders for Outfits and and Ships resale price. so the real hardcore players can even set ship resale to 0% if they so please.

Death permanent setting.
Adding the setting and escape pods would add point for those hardcore players.

Thunderforge:
I would be in favor of auto-landing as a preference, with the default to on.

Difficulty settings would make sense, but I think a lot more discussion needs to happen about what that means. The most obvious way would be weighting damage in favor of either the player or the enemies. A more advanced way would be to implement a system of [slotting](http://gamedev.stackexchange.com/a/112041/52205) on easier difficulties, where the AI sometimes ignores the player or does something stupid in order to give the player a break.

As for death permanent as a setting, I always felt that was something that nobody actually used. Besides, someone being hardcore can just delete their save if they really want to do that. But if others would actually use it, I would not object to adding it in.

Nikwigi:
Yeah I only knew a few that did I used it once in a while.  And while I agree with easy normal and hard for damage settings. I wouldn't want resale value tied in with it. I hate to have to play hard mode just to have my resale prices where I want them.

endless-sky:
The auto-landing AI only takes control if you are not moving your ship, so anyone who wants to land manually can do so.

Having ships depreciate is a reasonable idea, but it's already part of a separate feature request. (#428)

So, I'm shifting this issue to focus on "hardcore mode." That will mainly involve adding:
* Escape pods
* Ideally, actually show your escape pod getting rescued and brought to a planet.
* Either they must always drop you on a planet with ships for sale, or there must also be a way to "book a flight" to another planet, as a passenger.

[...]

PR #5977, opened 2021-05-16
Super rough implementation of escape pods (probably not finishing any time soon)

Amazinite:
**Feature:** This PR implements the escape pod feature request from issues #362 and #1150.

## Feature Details
This PR is a VERY rough potential implementation of escape pods. Think of this like my initial PR for system hazards that started out no where near finished but through further discussion and development grew into something pretty good. Although I will note that in this case, this PR actually has most of the functionality that I'm looking for out of escape pods, and it's moreso the code implementation that I think really needs looked at and tidied up.

This PR adds a new attribute that can be placed on outfits or ships:
* `"escape pod" "<model name>"`: The model name of a ship that the ship that has this attribute will spawn when it is in the process of exploding. Should the player's flagship be the ship that is destroyed, the player's flagship changes to the first escape pod that is ejected from their dying flagship. 

I chose to go with this implementation of the player taking control of the escape pod as opposed to other ideas such as having the escape pod eject as an aesthetic thing but have the player be teleported to a safe location due to immersion reasons. Outside of death, at no point does the game take control from the player and move them where it wants to. The player is always in control of where they're going. I believe that that should also extend to the escape pod, having the player be the one that controls the escape pod and gets themself to safety. And to that end, escape pods as I'm implementing them right now are ENTIRELY for the purposes of allowing the player to continue playing after their flagship was destroyed without needing to reload.

Questions of having mission cargo and passengers be moved onto escape pods, or what would happen if you eject in an escape pod but don't have the credits for a new ship, or any other questions around escape pods are for later. Right now this is just a matter of implementing a system by which the player ejects from their flagship when it is destroyed and is capable of flying themself to safety.

I'm sure I could explain more, but I've had at least three long conversations about this on Discord and I can't really rewrite out everything I think about escape pods in general and my specific idea of how they should work. If you have any questions, just ask and I'll answer. I'm also opening this now in an unfinished state because of recent discussions on the matter of escape pods, and just to show what I have for those who are interested.

## UI Screenshots
N/A

## Usage Examples
This outfit would add a Shuttle as an escape pod to the ship it is installed on, causing that ship to spawn a Shuttle when it dies. If the player was in said ship, they would begin piloting the escape pod.
```
outfit "Shuttle Escape Pod"
	...
	"escape pod" "Shuttle"
	...
```

## Testing Done
Yes.

## Performance Impact
N/A


tehhowch:
I don't think I like escape pods just being a special-case attribute, and I definitely do not agree with escape pods contributing 0 cost and 0 mass to the ship in which they are hosted.

movingpictures:
Presuming an escape pod is a simple  vessel with basic navigation and no hyperdrive, how is an escape pod different from a fighter other than the obvious, and moreover, could a fighter-equipped ship also use fighters as an escape? (Oddly, thinking Star Wars here, where Darth Vader escapes on a tie fighter just before the death star goes blooie.)

Amazinite:
> I don't think I like escape pods just being a special-case attribute

What sort of alternative do you suggest?

Another idea I had was for an escape pod bay that doesn’t get deployed when you deploy fighters, so ships could only have a certain number of escape pods dictated by the bay count, and this escape pod bay would eject its escape pod as the ship explodes just like regular carried ships. So in that case, purchasing an escape pod would just be purchasing another ship. The problem with that though actually becomes the UI; ships that already have drones and fighters don’t have room in the ship info panel for another line that would be added by escape pod bays. 

> I definitely do not agree with escape pods contributing 0 cost and 0 mass to the ship in which they are hosted. 

In the case of an escape pod as an outfit, the outfit is what holds the cost and mass, and in the case of a built-in escape pod, the ship’s stats should reflect any built-in outfits, so under this implementation escape pods can certainly contribute cost and mass to their carrier.

petervdmeer:
Why was Ship chosen as the escape-pod-in-flight implementation? Using flotsam as base for escape-pods-in-flight (as suggested in #1150) feels more appropriate to me than using ships:
- Flotsam is a simpler class (which should be beneficial for performance if large amounts of pods get released during large battles).
- Both flotsam and escape-pods are things that are expected to be picked up. (And storing them in a cargo-hold after pickup also seems appropriate.)
- We want escape pods (at least temporary) to be unaffected by weapons fire.
- I feel that it is very inappropriate for escape pods to have hyperdrives and/or jumpdrives. (They feel more like "life rafts" that would be deployed when the main ship goes under and that helps the crew of the main ship to stay alive longer while waiting for rescue.)
- Both flotsam and escape pods would just appear at a ships destruction.
- I consider it unlikely that we want to arm escape pods with weapons.
- Flotsam currently contains a commodity and/or outfits, adding "crew/passengers/persons" doesn't seem like a big step.

Amazinite:
> Why was Ship chosen as the escape-pod-in-flight implementation?

From the PR description.

> I chose to go with this implementation of the player taking control of the escape pod as opposed to other ideas such as having the escape pod eject as an aesthetic thing but have the player be teleported to a safe location due to immersion reasons. Outside of death, at no point does the game take control from the player and move them where it wants to. The player is always in control of where they're going. I believe that that should also extend to the escape pod, having the player be the one that controls the escape pod and gets themself to safety.

So to this end, I chose Ship because the player is already capable of controlling a Ship, but I'd need to add controls to a Flotsam, or add that "fade to black" situation that I explicitly want to avoid.

Comparing this to other media, I'm looking at escape pods more like the Capsules of EVE Online. Capsules are ejected as a ship dies and are jump-capable vessels for the player to flee to the safety of a friendly station.

And using Ship allows escape pods to be other, smaller ships and not just always some special escape pod flotsam. For example, a Bactrian when it explodes might be capable of using Shuttles as escape pods.
```
outfit "Shuttle Escape Pod"
	...
	"escape pod" "Shuttle"
	...
```

As far as armed escape pods go, there's no telling what alien factions we might have that are crazy enough to lose their ship but continue fighting in a smaller vessel. And before even talking about escape pods, I've imagined an alien ship that upon destruction releases a number of smaller ships that continue to attack the player. In the past I would have achieved this by adding a personality that prevents NPC ships from deploying their fighters, but under this PR such ships could just be a number of built-in "escape pods."

So while Flotsam might be simpler to make use of, I think that making use of Ship allows for more interesting possibilities.

source/Ship.cpp
```diff
@@ -2720,6 +2738,25 @@ void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
...
+	SetParent(ejector->GetParent());
+	// This ship behaviors like its ejector did.
+	isSpecial = ejector->isSpecial;
+	isYours = ejector->isYours;
+	personality = ejector->personality;
+	
+	// Escape pods won't get targeted for the first 3 seconds after being ejected.
+	// Escape pods are very fragile ships, and this gives them a chance to escape.
+	if(attributes.Category() == "Escape Pod")
+		untargetable = 180;
```
petervdmeer, line 2755: I would expect untargetable ships to also be unable to use their weapons (as long as they are untargetable).

source/Ship.cpp
```diff
@@ -2720,6 +2738,25 @@ void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
...
+	// If the ejecting ship had a parent, use that as the escape pod's parent.
+	SetParent(ejector->GetParent());
+	// This ship behaviors like its ejector did.
+	isSpecial = ejector->isSpecial;
+	isYours = ejector->isYours;
+	personality = ejector->personality;
+	
+	// Escape pods won't get targeted for the first 3 seconds after being ejected.
+	// Escape pods are very fragile ships, and this gives them a chance to escape.
+	if(attributes.Category() == "Escape Pod")
```
petervdmeer, line 2754: I prefer not to put special meaning to categories here, especially if we also want Fighters and Drones to be possible candidates to be used as escape pods.
Alternatives is to put the "untargetable" property on escape pods directly, or maybe even add an "untargetable after escape launch" attribute to each ship that can be used as escape pod.

petervdmeer:
I'm okay with escape-pods being ships (your explanation sounds quite good), but I like to see the the mechanisms behind it as generic as possible.
Avoiding the "fade to black" situation is good and choosing ship as base for Escape Pods indeed gives us a bit more options.

I also like to have the option of escape-pods that act as "local life-rafts that need to be picked up". (If escape pods are ships, and we make this generic, then the request is to have an option to`rescue passengers and crews from stranded ships` or to ensure that `many ships have the bays to pickup ships of the "Escape Pod" category`.) => I'm very much okay to to keep this for a separate PR.

petervdmeer:
> And using Ship allows escape pods to be other, smaller ships and not just always some special escape pod flotsam. For example, a Bactrian when it explodes might be capable of using Shuttles as escape pods.

Would it be possible to implement this first (in a separate PR?) It feels as if only the code to change the flagship from this PR needs to be transferred into a new PR to make this possible (plus a one-liner in the launch-code).

source/Ship.cpp
Amazinite: I'd prefer to go with the attribute route instead of having all escape pods be untargetable.

Amazinite:
Relaying some information from a discussion about #7644 on the direction we should take escape pods:
* Create a new keyword that is a child of `bay` on `ship` that denotes a bay as being used for escape pods. Any bays with this keyword will be unable to be deployed. Only during the destruction of the ship will carried ships in such bays be released.
  * Should bays marked like this be guaranteed to be released, unlike drones and fighters where there's no guarantee that they'll deploy before the ship explodes?
* While a ship is exploding and releasing its docked ships, if the exploding ship is the player's flagship then transfer the player's flagship to the first ship that escapes.

The combination of these two changes would constitute what we're looking for in an escape pod. Additional content or mechanics could come later, e.g. a dedicated "escape pod" ship category with "escape pod" bays that we place on various ships, or allowing mission passengers and cargo to be transferred to escape pods while the mothership is exploding.

PR #5978, opened 2021-05-16
feat(mechanics): Customizable game constants through "gamerules"

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #5556.

## Feature Details
This PR provides a potential implementation of a gamerules file, a file which contains a variety of game constants which are able to be tweaked within a data file instead of requiring changing the source code and recompiling.

So far this PR turns the following game mechanics into gamerules:
* The various constants used for deprecation.
* Various constants related to flotsams, namely their drag values and constants used for determining their lifetimes.
* The constants used to determine hit force scaling based on the mass of the ship.
* The person ship spawnrate.

## UI Screenshots
N/A

## Usage Examples
The following are the current constants used for deprecation.
```
gamerule
	"depreciation: full" 0.10
	"depreciation: daily" 0.997
	"depreciation: grace period" 7
	"depreciation: max age" 1000
```
If a plugin wanted to increase the grace period timer, they could do the following:
```
gamerule
	"depreciation: grace period" 30
```
This overrides the previous definition of this constant while leaving all other constants untouched. 

## Testing Done
Altered the various gamerule values that I've currently implemented and observed how they properly alter the game mechanic that they're tied to.

## Performance Impact
N/A


quyykk:
Some thoughts:

- I don't really like having `GameData` to Yet Another Thing<sup>tm</sup>. I suggest making a `GameRules` class that is then passed to `GameData` for easy access.
- There really should be typo checking and out-of-bounds checking.
- Do we really need to access gamerules with strings? An enum might be cleaner and less error prone, or you could add functions like `BaseMass()` to the `GameRules` class mentioned above.

Amazinite:
I think that should be good for now. We can add more gamerule constants to tweak after the system is finalized and merged.

Amazinite:
Closing since it still feels a ways off before I'll be focusing on finishing this.

PR #5979, opened 2021-05-16
Add a tag for systems that hides them from view unless in visible range

Amazinite:
**Feature:** This PR implements a feature I thought of after b3ef517.

## Feature Details
This PR adds a new system tag:
* `shrouded`: This system can not be seen unless it is within view range or linked to an unshrouded visited system.

As a reminder, hidden's definition is the following:
* `hidden`: This system can not be seen unless it is visited or linked to a visited system.

Combining hidden and shrouded then results in the following behavior:
* This system can not be seen unless it is linked to an unshrouded visited system.

The exact behavior that is desired is that `shrouded` systems are "forgotten" when not within view. This means that a conceivable region of shrouded systems would require the player to remember the system layout and to manually jump system to system, but a system that is forgotten when no in view could have other uses as well.

## Usage Examples + Screenshots
I changed Sol, Alpha Centauri, and a number of other systems to be shrouded.
```
system Sol
	shrouded
	...
```
<details><summary>The following are a collection of screenshots of me flying through these shrouded systems. Observe how those systems that are shrouded and within range are seen, but the system name and their links to other shrouded system are forgotten after leaving.</summary>

image
image
image
image
</details>

<details><summary>I then added hidden to all these systems so that they are both shrouded and hidden. The behavior is mostly the same, except now I can't see those systems which are within range but not linked to any systems I can see.</summary>

image
image
</details>

## Testing Done
Tested using the usage example.

## Performance Impact
N/A

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -1835,6 +1835,16 @@ void PlayerInfo::CheckReputationConditions()
 // they have actually visited it).
 bool PlayerInfo::HasSeen(const System &system) const
 {
+	// Shrouded systems are only seen if they're currently in view range, or
+	// if a visited system that is non-shrouded is linked to it.
+	if(system.Shrouded())
+		return (system.VisibleNeighbors().count(this->system) || &system == this->system
+			|| any_of(visitedSystems.begin(), visitedSystems.end(), 
+				[&system](const System *s) noexcept -> bool
+				{
+					return (!s->Shrouded() && s->Links().count(&system));
+				}));
+	
 	if(seen.count(&system))
 		return true;
 	
@@ -1862,7 +1872,8 @@ bool PlayerInfo::HasSeen(const System &system) const
 // Check if the player has visited the given system.
 bool PlayerInfo::HasVisited(const System &system) const
 {
-	return visitedSystems.count(&system);
+	// Shrouded systems are only considered visited if you're currently in them.
+	return (visitedSystems.count(&system) && !system.Shrouded()) || (system.Shrouded() && &system == this->system);
 }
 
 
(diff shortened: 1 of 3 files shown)


tehhowch:
I don't see how this makes sense from a player perspective. It is literally the job of your ship's navigation systems to remember where you've been and what you've seen there. Once visited, any system and its details should remain known to the player. To seemingly delete this information automatically is a huge departure from what the player is taught through normal gameplay.

We currently offer the "unvisit" functionality to events as a way to implicitly indicate to the player that a once-known area has changed and should be returned to. How do you propose we avoid confusion between the two systems?

MasterOfGrey:
> How do you propose we avoid confusion between the two systems?

From my understanding this was almost exclusively built with the intent to be used for strange Ember Waste shenanigans?

Amazinite:
> How do you propose we avoid confusion between the two systems?

Well I would expect the context of the situation to provide enough info to understand what is going on. `unvisit` is only used by events, and so far we're only using it whenever something major happens, and that major event is made clear to the player. (e.g. systems literally disappearing from the hyperspace network in the FW campaign, the Eye opening in the Wanderer campaign.) `shrouded` on the other hand would cause the system to disappear as soon as the player moves away from it, something they should notice, and a couple trips back and forth should make it clear that the system is not simply being repeatedly unvisited, but instead has something else going on with it.

Zitchas:
The best use for something like this would be combined with something like sensor static (I know, we don't have that in ES. The EV series did, though, and used it to good effect, so it is something else I'd like to consider)

This could be used to create clusters of systems that for some reason can't be reliably mapped. Maybe they are systems that are orbiting a black hole; or under strong disruptive forces of some kind? Maybe they are just so filled with background noise or surrounded by haze that the nav software can't get a solid fix on where they are located? The core systems of Nebula would be a good place for these.

Galaucus:
I could see this being useful to create a "randomly generated" cluster of systems that rely on missions to populate them and cycle contents during repeat visits. It's kind of a hard sell for anything that's meant to be static, though.

Amazinite:
Closing as this isn't a new mechanic I'm particularly enthusiastic about and is mainly something I thought of as something that might be cool to have but wasn't necessary and threw together in a short amount of time. I'll be keeping the branch up though so I'll likely bring this PR back at some point in the future.

Issue #5976, opened 2021-05-16
Heat damage is exceedingly powerful when compared to other special damage types


Amazinite:
**Is your feature request related to a problem? Please describe.**
Heat damage is an extremely powerful form of damage that doesn't lose much effectiveness, if any, as you increase in tier. A small secondary weapon unlocked during the FW campaign, the flamethrower, is one of the best weapons for punching up and taking down larger ships. (Not on its own, of course, and to an extend that's the whole purpose of the weapon given how the FW use it to take down Navy capitals, but the degree to which it is successful is a bit much.)
Other special damage types by comparison aren't nearly as effective. Disruption is probably the second best in its usefulness in disabling ships without destroying them, slowing being more of an annoyance, and ionization being a worse version of heat damage in terms of its utility as a special damage type.

Something should probably be done about heat damage, either in looking to alter the effectiveness of the weapons that use it (mainly the flamethrower), or in changing how heat damage itself works.

**Related Issue Links**
Idk.

**Describe the solution(s) you'd like**
The following are a number of things I've seen discussed in relation to this issue:
* Have overheating a ship not outright disable it, but instead give it a chance to become disabled for a brief moment, akin to low crew behavior. This would attack heat damage's major strength of completely taking a ship out of combat. Ion damage can also do this, but it is much more difficult to achieve, and often times you only prevent a ship from using some of its systems instead of outright disabling it, so at least a heavily ionized ship can fight back to some degree or try to run if it doesn't have energy for its weapons.
* Just nerf the flamethrower's heat damage. I don't know by how much, but nerf it.
* Buff ion damage. As described, ion damage is the closest damage type to heat damage, but worse. This would probably occur in conjunction with a nerf to heat damage so that we aren't just creating two overpowered damage types. Buffing ion damage might entail simply increasing ion damage on all ion weapons. Another idea that came up though that I like was allowing ion damage to bypass the 50% debuff to special damage that shields give, so while ion damage is always effective against ships, heat damage is more effective against de-shielded ships. This could also create some inter-play between these two damage types; hit a ship with ion damage to help sap its shields then hit it with heat damage to practically disable it when the shields are down.

**Additional context**
[...]

PR #5982, opened 2021-05-17
Rebalancing heat damage

Quantumshark:
Potential solution to #5976 
Initial proposal for rebalancing of heat damage.
Heat damage has been halved across the board, and reduced by 75% on the flamethrower.

### DIFF
--- a/data/human/weapons.txt
+++ b/data/human/weapons.txt
@@ -506,7 +506,7 @@ outfit "Plasma Cannon"
 		"hit force" 30
 		"shield damage" 63
 		"hull damage" 25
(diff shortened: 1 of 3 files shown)


MasterOfGrey:
While this could work in the short-term, I don't believe this is an effective balancing solution long-term.

Amazinite:
Closing because I don't think this is the way to go on heat damage. At the very least, the Wanderer and Korath weapons don't need their heat damage addressed. Even the plasma cannons aren't too bad, as it's mainly the flamethrower that's the issue.

PR #5984, opened 2021-05-19
Fix Grammar in Scar's Hideout's description

Anarchist2:
From `Scar's Hideout is a hollowed out asteroids` to `Scar's Hideout is a hollowed out asteroid`.

Also, should it be `hollowed out asteroid` or `hollowed-out asteroid`?


tehhowch:
Should be `hollowed-out`, yes

Issue #5934, opened 2021-05-03
Addition of "Turn Fuel" and "Thrusting Fuel" attributes


VaelophisNyx:
**Is your feature request related to a problem? Please describe.**
Problem is only a lack of attributes of use to plugin developers

**Related Issue Links**
https://discord.com/channels/251118043411775489/431496424992014347/838770238778638377
current discussion of this on the Discord server

**Describe the solution you'd like**
Simply adding these two attributes to allow fuel to be used by thrusting and steering would be all that is needed here.

**Additional context**
I doubt this would be much use in vanilla ship outfits, but, could still be nice to add.

Plip-Plop:
Do you mean hyperdrive fuel?

GalacticRuler:
Indeed, being consistent with other uses of attributes for fuel use.

PR #5986, opened 2021-05-19
Thrusting/Turning Fuel

Ferociousfeind:
**Feature:** This PR implements the feature request detailed and discussed in issue #5934 

## Feature Details
Adds three attributes, "thrusting fuel", "reverse thrusting fuel", and "turning fuel", which consume that amount of fuel per frame while turning/thrusting.

## UI Screenshots
N/A

## Usage Examples
These are the outfit edits I used to test them out:
```
outfit "Greyhound Plasma Thruster"
	category "Engines"
	"cost" 45000
	thumbnail "outfit/small plasma thruster"
	"mass" 34
	"outfit space" -34
	"engine capacity" -34
	"thrust" 18.4
	"thrusting energy" 1.8
	"thrusting fuel" 0.18
	"thrusting heat" 3.4
	"flare sprite" "effect/plasma flare/small"
		"frame rate" 6
	"flare sound" "plasma small"

outfit "Greyhound Plasma Steering"
	category "Engines"
	"cost" 36000
	thumbnail "outfit/small plasma steering"
	"mass" 26
	"outfit space" -26
	"engine capacity" -26
	"turn" 492
	"turning energy" 0.9
	"turning fuel" 0.09
	"turning heat" 2.1
	"steering flare sprite" "effect/plasma flare/small"
		"frame rate" 6
	"steering flare sound" "plasma small"
```

## Testing Done
Equipped a Behemoth with the above outfits instead of its default ion thrusters, took it out for a spin. Fuel is consumed when jumping, and when turning or thrusting. And the small amount of inherent ramscoop means you're not _entirely_ stranded if you run out of fuel.

## Performance Impact
N/A


source/Ship.cpp
```diff
@@ -1721,6 +1721,9 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 			double cost = attributes.Get("turning energy");
 			if(energy < cost * fabs(commands.Turn()))
 				commands.SetTurn(commands.Turn() * energy / (cost * fabs(commands.Turn())));
+			cost = attributes.Get("turning fuel");
+			if(fuel < cost * fabs(commands.Turn()))
+				commands.SetTurn(commands.Turn() * fuel / (cost * fabs(commands.Turn())));
```
Amazinite, line 1726: I don't think this will work out.

What this is currently doing is limiting the turn rate by the energy cost, *then* by the fuel cost, instead of limiting by both at once. So say you require 10 energy and 10 fuel but you only have 1 energy and 5 fuel available. What *should* happen is that you use 1 energy and 1 fuel for 10% turn power as that's what you have available. I believe that what this is doing is seeing that you only have enough energy for 10% turn power and only enough fuel for 50% turn power, so it'd stack those two limits and give you 5% power because of you're calling SetTurn twice and the second call is getting the turn rate from the previous call because of the Turn call.

What you should be doing is checking your available energy and fuel, then taking the minimum of the two to set your turn. And while we're here, I think it would also make sense to allow heat to be used as a resource (i.e. check if heat + turn heat > 0 and limit the turn if it isn't), as we've done that with weapons and repairs.

This same issue appears to be the case for thrusting.

Ferociousfeind: That's a good point, heat-wise, though I believe because we're decreasing "cost" by our available energy, it'll only check if the amount of thrust/turn we can produce energy-wise will violate our fuel usage too, e.g.

If we use 10 energy and 10 fuel, and we have 1 energy and 5 fuel, we check to see if 10>1, and set our usage to 10%. Then, we check if 10*0.1>5, and see that since we're consuming only 1 fuel out of a total of 5, things check out. The same sort of deal going on in DoRepair(), but it's been heavily streamlined and generalized. The afterburner situation is different, and sort of assumes that one frame is the indivisible unit of afterburner production, require enough fuel ans energy for a full frame of full-blast afterburner.

Tomorrow I'll check for negative thrusting/turning heat and only allow movement if your heat will remain >= 0

Ferociousfeind:
Oh no, I broke it without properly testing it first
What did I do

Ferociousfeind:
ah fuck I left in "heatCost" instead of reducing it back to "cost" fuck

Ferociousfeind:
oh... it appears I have fucked it up again...

Ferociousfeind:
this time I fixed it by properly replacing all the "cost"s with the proper "energyCost", "fuelCost" and "heatCost"

Ferociousfeind:
oh yeah by the way, #6408 obsoletes this

Issue #4656, opened 2019-11-10
Expand stealth objective to be able to only fail on specific government scans


Amazinite:
**Is your feature request related to a problem? Please describe.**
Currently, illegal goods are either all considered illegal by a faction, or a faction considers nothing illegal. This same all-or-nothing approach also extends to the `stealth` mission objective (don't get scanned before completing all other objectives and reaching the destination). While there are good reasons behind why outfits (and ships) don't have the ability to only be deemed illegal by some governments (mainly the complexity of giving that information to the player), that same reasoning can not be made for why the `stealth` mission objective (and perhaps even `illegal` for missions) is also an all-or-nothing mechanic. This is because missions have the ability to explain to the player who exactly it is they should avoid being scanned by.

**Related Issue Links**
#1429

**Describe the solution you'd like**
Have the stealth objective be able to limit who you're scanned by (and perhaps where you're scanned) to determine if the missions should fail, instead of having the mission fail due to any scans. Perhaps government jurisdiction could in some way be extended to the `stealth` objective to specify what governments and/or systems that the player is scanned by/in cause the mission to fail. 
This behavior should also be extended to `illegal` for missions, having the fine only be applied by certain governments or in certain systems.

**Usage example**
You're in touch with the Resistance against the Heliarchs in Coalition space, and they want you to bring them some weapons from outside Coalition space. The Heliarchs will definitely want to question you if you're carrying around a bunch of weapons and munitions in your cargo hold while traveling to or around Resistance areas, so you might slap `stealth` and `illegal` onto the mission. But, this now means that you'll fail the mission for being scanned by Navy ships or be fined for having weapons in your cargo by human planets, which breaks immersion for the mission since only the Heliarchs are suppose to care about your cargo.

Zitchas:
I definitely think this would be very useful for making more immersive missions.

Zitchas:
Given a mission I just wrote for the main Remnant storyline, this just bumped up from "nice to have" to "almost essential to get this mission to function as intended."

That being said, a slight tweak to the above idea would be awesome: The ability to specify that the stealth restriction about being scanned applies to all scanning entities *except* for a specific government. 

For instance, a Remnant mission to pick up some suspicious cargo and bring it back to them. In this case, the mission should fail if anyone *except* the Remnant scan the player.

I suspect that many missions fall into this sort of use-case: The giver of the mission already knows about it and won't fail the mission, but anyone else *will* fail it.

Zitchas:
[...]

Issue #1429, opened 2016-08-12
Ability to make specific governments find specific things illegal


Amazinite:
Currently, the illegal attribute is a blanket thing. If something is marked illegal, then everyone who cares about illegal things will fine you for it. But what if you could choose which things were considered illegal by different governments?

Example situation:
Let's say we have two items. Item A and item B, as well as three governments, 1, 2 and 3. 

Government 1 hates all illegal things, and will fine you for possession of both A and B. 
But government 2 is a little more laid back. They only thinks that A should be illegal, but B is completely fine. 
And government 3 is the exact opposite of 2, finding B illegal but saying A is fine.

Right now only nerve gas is illegal (at least in terms of outfits), but this feature could probably open up more possibilities. My first thought is a drug transportation job that you have to do through Republic space. These cargo items would be illegal in Republic space, but may be completely legal in Syndicate or Free World space where the destination may be. They would pay a lot of money so that the more you get fined, the less and less money you gain off of the job. (So pretty much job versions of the current drug running missions with a twist.)

KiLEdEnNis:
great idea, could also add things where alien outfits are ok for them but illegal in human space

endless-sky:
That's how ship licenses used to work - for example the Republic required you to have a "Freight license" but the Free Worlds did not. It ended up being super annoying because you would buy a ship legally in one part of human space, then fly to another region and get fined for operating a ship without a license. The same would hold true with illegal outfits - it would be annoying to get fined every time you accidentally enter a different region of space.

Right now, some governments just have a much lower security level than others, which I think is a reasonable compromise.

Also, note that if different things were legal in different regions of space, there would need to be some way to communicate that to the player so that they know where they're allowed to use it and where they're not. That's introducing a lot of complexity for a feature that's going to see relatively limited use.

KiLEdEnNis:
what does the "security" thing do anyway?

endless-sky:
"security" is the probability that they'll check your ship for illegal outfits / cargo when you land.

Amazinite:
Closing because #1452 was milestoned.

tehhowch:
One way to do this would be with a data structure in Politics that pairs an illegal outfit with a LocationFilter. this outfit would then be illegal in any system, planet, or to any ship that matches the LocationFilter.
To display this to the player would be a matter of reworking the Shipyard & Outfitter view in the Map Panel:
 - opening the respective panel (i.e. no ship / outfit selected)
[...]

PR #5988, opened 2021-05-21
Only lower reputation due to atrocities if governments are friendly

Amazinite:
**Bugfix:** This PR addresses an issue with demanding tribute affecting reputation with governments.

## Fix Details
A while ago, demanding tribute from a planet was changed to be considered an atrocity. 7675eb7
It has been noted though that since then, demanding tribute from a pirate planet will lower your reputation with the Republic, despite the fact that the Republic has a negative attitude toward pirates. 

As it turns out, committing an atrocity is currently causing your reputation to drop to 0 for *all* governments that have an attitude toward the government you committed an atrocity against, whether that be a positive or negative attitude.
This PR changes it so that atrocities only drop your reputation to 0 with governments that are friendly toward the government you committed an atrocity against.

## Testing Done
While having a positive reputation toward the Republic, demanded tribute from a pirate planet with and without this change. Without this change my reputation with the Republic was dropped to 3. With this change my reputation was unchanged after demanding tribute.

## Save File
Your own save file will probably have a friendly republic and hostile pirates to test this on.

### DIFF
--- a/source/Planet.cpp
+++ b/source/Planet.cpp
@@ -579,6 +579,9 @@ string Planet::DemandTribute(PlayerInfo &player) const
 		for(const auto &gov : toProvoke)
 			gov->Offend(ShipEvent::PROVOKE);
 		// Terrorizing a planet is not taken lightly by it or its allies.
+		// TODO: Use a distinct event type for the domination system and
+		// expose syntax for controlling its impact on the targeted government
+		// and those that know it.
 		GetGovernment()->Offend(ShipEvent::ATROCITY);
 		return "Our defense fleet will make short work of you.";
 	}
--- a/source/Politics.cpp
+++ b/source/Politics.cpp
@@ -106,7 +106,7 @@ void Politics::Offend(const Government *gov, int eventType, int count)
 			// friendly without the player's behavior toward one of them
 			// influencing their reputation with the other.
 			double penalty = (count * weight) * other->PenaltyFor(eventType);
-			if(eventType & ShipEvent::ATROCITY)
+			if(eventType & ShipEvent::ATROCITY && weight > 0)
 				reputationWith[other] = min(0., reputationWith[other]);
 			
 			reputationWith[other] -= penalty;


tehhowch:
I disagree.

Why wouldn't the Republic think less of you if you start demanding money from others and threaten them with death if they don't comply?

That your reputation begins increasing with them again afterwards is more an artifact of the reputation system.

Amazinite:
Why focus on the Republic example? If we have two governments, government A and government B, who are hostile with one another, and the player is friendly with government A while being hostile to government B, why should government A be mad at you for demanding tribute from government B? Government A could think that government B should be wiped off the map, yet the current atrocity mechanic makes it such that government A will dislike you demanding tribute from government B.

And it isn't just the Republic who your reputation is being reset with. Deep Security, the Free Worlds, merchants, the militia, the neutral government, and the Syndicate all also have their reputation set to 0 then bumped back up from there. Perhaps an argument can be made that some of those would also care about you demanding tribute from pirates, but would all of them?

And looking to the future of us potentially making outfit legality (#1429) or mission cargo legality (#4656) differ by government, if government A finds outfit/mission cargo X completely legal but government B considers it an atrocity, why should government A's reputation be wiped out by me getting scanned with X by government B?

tehhowch:
I went for the republic example because it's the closest to what your post suggested.

In the broader context of modifying the legality system, we would be updating this handling anyway, so I don't believe acting now achieves anything significant for those later efforts.

On the matter at hand, we're just reminded again that the tribute system is an afterthought. That it uses atrocity (also an underdeveloped concept) to make things harder is a hack that helps drive up the difficulty for the intended audience, end-game players looking for a fight.

A better solution is to develop a new event type value, a default reputation mapping for it, and a government-specific mapping. Only then will authors be able to achieve the granularity they want

Terin:
Atrocities are definitely a loosely defined system. However, this fix is simple enough that I don't see the need to have it wait for a theoretical future revamp.

tehhowch:
IMO the intent of atrocities is that the player has done something so bad that _everyone_ disapproves of it having been done. In that regard, committing an atrocity should influence the player's reputation with everyone privy to the details of the event, which is conceivably everyone who witnesses the event firsthand, and everyone who can be notified by the witnesses via intersystem  communications.

We don't currently support this visibility-based reaction system, and use the reputation system as a proxy instead. Given that we already have a guard on this behavior, preventing atrocities from affecting reputation with governments whose weight is less than .05, I'm less opposed to the specific change in this PR.

I'd prefer aligning on atrocities as something deemed horrific even to those that dislike the victims (i.e. committing war crimes should not _improve_ your reputation, except maybe with those that also enjoy committing similar crimes), but recognize we need an improved system to support customizing these relationships.

MasterOfGrey:
Putting aside the fact that tehowch's definition of atrocity is hilariously US/western-centric...

The existing behaviour is actively unintuitive behaviour for the major cases that come up, and actively limiting to possible narratives.

It would definitely be nice to have a broader overhaul of the reputation, atrocity, and illegality system for improved government-specific granularity. In the _interim_ though I definitely think this qualifies as a bug that should be fixed. It's quite destructive to keep it while waiting for an overhaul that could be years away.

Terin:
Yeah, out all the possibilities we could have, this is a really unintuitive and broken way to go about it. It'd only be worse if it made the entire galaxy hostile, but even then, that'd make a little more sense. Even if we did introduce in-universe atrocities soon (which it seems we don't have plans to any time soon), I don't think anyone would really complain that the Free Worlds didn't go hostile because the Republic scanned one. I know you don't like tribute, but this is still just a blatant, easily fixable bug.

Zitchas:
A) The illegality / atrocity / reputation system needs an overhaul.

B) As far as I'm concerned, I agree with Tehhowch here that the point of labeling something an atrocity isn't to identify something as an "extra bad crime," but rather to identify it as something so heinous that virtually everyone despises. it. (unless, of course, they are also the sort of person who likes that sort of thing)

Demanding tribute isn't just "I killed a whole bunch of them," it is literally extorting money from someone under threat of orbital bombardment, demonstrated by your ability to lay waste to their defensive fleet. 

To use a real world example, I'm fairly sure if some random private citizen acquired a large private military force and extorted money out of a small country that very few like and many hate via the process of subjecting it to massive bombardments that destroyed its military infrastructure, I'm fairly sure that most countries would be really upset by it; Despite the fact that most countries would be happy if said uncooperative country vanished from the globe, none of them are likely to support or condone the sort of behaviour that could be an active threat to themselves at some future date. 

In fact, we even see this in player behaviour. Sure, everyone starts off just tributing the pirate planets. But once they've started, how many then move on to doing so to virtually every other planet available? Once they have tasted the power, they almost always want more. 

So it makes sense that even governments that are hostile to the government that the player demanded tribute from would be upset by their actions - it is an implicit threat to their own security, and contributes to the snow-ball effect of <first> accumulating more resources that they can then use to move on to bigger targets.

This is, historically, one of the reasons why states have fought so hard to maintain a monopoly on violence. (aka restrictions on who can use weaponry, particularly large scale weapons). Because even if someone outside of state control is using it on a desirable target, it is really hard to stop it from going out of control. See the various problems with mercenary groups, including those hired by governments and frequently going "off the rails" for examples of these problems.

In short, not a new problem, and certainly one that military strategists in most governments are probably going to be aware of.

Amazinite:
I think that in either case there’s an issue then. Either atrocities shouldn’t hurt your reputation with governments who have a negative attitude toward the people you committed an atrocity against, or committing an atrocity should make everyone hate you, which isn’t the case since positive rep just gets dropped to some low value but doesn’t actually go negative. (And perhaps committing an atrocity should stop your rep from increasing at all with the effected governments for the day, since even if demanding tribute from a pirate planet gave you negative Republic rep, successfully conquering the planet would make them like you again.) And if we want to go the angle of demanding tribute being an atrocity in human space, I’ll note that the FW, Republic, and Syndicate don’t have any attitude toward each other that would make demanding tribute from one cause you to become hated by the others; it’s only pirates that they would all care about.  

I think at this point I definitely agree with “atrocity” being able to be a negative hit to all knowing parties, but for the time being I feel like it makes more sense to make this change and then expand upon the atrocity/legality/reputation system later, since there’s no indication in game that demanding tribute from pirates will hurt your Republic reputation, so it certainly looks like a bug right now.

tehhowch (approved):
In case it was missed above:
> Given that we already have a guard on this behavior, preventing atrocities from affecting reputation with governments whose weight is less than .05, I'm less opposed to the specific change in this PR.

Zitchas:
OK, seems reasonable then.

Amazinite:
Should we leave a TODO comment here?

Zitchas:
In other words, I think there should be a difference between "we hate them and would like to see their government destroyed" and "We enjoy the idea of all their people dying an agonizing death as their nervous system is slowly disintegrated by poison, or watching their civilians suffocate to death as their domes are shattered by orbital bombardment."

Normal reputational changes is the former, atrocity is the later.

But I also agree that this distinction isn't particularly clear, especially in terms of how it is reflected in the reputation changes. So this change is fine for now, but we really need a revamp to the whole illegality/atrocity/reputation system.

tehhowch:
> Should we leave a TODO comment here?

```c++
// TODO: Use a distinct event type for the domination system and expose syntax for
// controlling its impact on the targeted government and those that know it.
.
```
Probably that belongs in Planet.cpp and/or Government.cpp


To address what Grey mentions: we don't currently have a way to represent an outfit or action that is interpreted by a single government as punishable-by-death, which other governments find acceptable. Accepting @Eloraam 's legality system proposal (#5141) should enable that, though additional changes are still needed for this "reputation propagation" section.

PR #951, opened 2016-04-14
Support planet "language" and attributes "no trading", "no jobs", "no banking", and "no hiring".

jmhorjus:
A planet can now have a "language" (i.e. "language Esperanto")  which will automatically disable banking, hiring, and jobs unless the player condition "language: Esperanto" is defined.  If a planet does not define a language, the language of the planet's government will be used (or the language of the system's government, if the planet does not define a government).  Planets need not define languages individually; normally language will still be defined only by the government, and you will still get the full effect.  Setting "language clear" on a planet can disable any language requirement for banking/hiring/jobs that might have been inherited from a planet/system government. 

Trading, outfitter, and shipyard availability is unaffected by language (universal currency or barter economy I guess?).  

Planets can also set "no trading", "no jobs", "no banking", or "no hiring" as attributes.  By default, all these things are still all available on inhabited planets that have a spaceport, just as they were before.

PR #5995, opened 2021-05-22
Feature: "fuel cache" Attribute

InfiniteDonuts:
## Summary
@Galaucus (aka Lake) is creating a pirate storyline, and asked me for help implementing a new attribute for planets, called `"fuel cache"`. Basically, it enables a planet to refuel without having a spaceport -- the idea is that pirates have hidden fuel caches on uninhabited planets.

## Testing Done
Playtested the feature, shown in the below pics:

Modded Omnis to have a system with a fuel cache'd planet:
image
image

Fuel before landing:
image

Fuel after landing:
image

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2649,7 +2649,12 @@ void Ship::Recharge(bool atSpaceport)
 	if(atSpaceport)
 	{
 		crew = min<int>(max(crew, RequiredCrew()), attributes.Get("bunks"));
-		fuel = attributes.Get("fuel capacity");
+	}
+	if(GetPlanet())
+	{
+		const auto planetAttrs = GetPlanet()->Attributes();
+		if(planetAttrs.find("fuel cache") != planetAttrs.end() || atSpaceport)
+			fuel = attributes.Get("fuel capacity");
 	}
 	pilotError = 0;
 	pilotOkay = 0;


Amazinite:
Being honest, I feel like planet attributes are a little bloated in what they do. Their main purpose has been for use in location filters for missions, and `uninhabited` has been a special attribute for I don't know how long that shows a planet as uninhabited on the map and removes all services, but then attributes that follow the scheme `"require: <attribute>"` were given the ability to restrict landing access (and also even telling if the planet is landable in the first place) to ships with the attribute(s) that the planet attributes describe. This tying of ship attributes required to land to planet attributes actually makes the system of needing a certain thing in order to land difficult to expand, at least insofar as we continue making use of attribute to define such a system. So I've thought of breaking the mechanics of `"require: <attribute>"` off into its own node under `planet`, allowing for example the ability to require X or Y attributes in order to land, whereas the current system must require all attributes mentioned. (Such a new node could also be used to define the type of behavior that requiring an attribute shows, such as #3646's `prohibits` special attribute that shows the planet as inhabited but doesn't allow you to land if you have the listed attribute(s), whereas `requires` shows the planet as uninhabited and unlandable if you don't meet the requirements)

In the same way, I think that tying attributes to planet services (e.g. #951) should be avoided, and instead we should create a new planet node that defines the services that a planet offers. If we want these to automatically generate planet attributes then that can be done (like how a planet with a shipyard, outfitter, or spaceport automatically generates the `shipyard`, `outfitter`, and `spaceport` attributes for missions to make use of), but I don't think that the functionality itself should be tied to the attributes.

InfiniteDonuts:
Seeing as this is a very small change, it shouldn't be too hard to adapt the fuel cache to whatever new system you propose.

tehhowch:
Are these "fuel caches" available to anyone who lands there? If so, I would expect them to be taken or monitored, meaning they wouldn't offer anything of use to a rogue actor.

Similarly, unless these attributes are applied to a planet that is normally considered for landing (and planets without a spaceport are generally not), the ai won't ever attempt to land on these planets.

In other words, I think this use case needs more concept development before any code changes are considered for merges.

Galaucus:
> Are these "fuel caches" available to anyone who lands there? If so, I would expect them to be taken or monitored, meaning they wouldn't offer anything of use to a rogue actor.
> 
> Similarly, unless these attributes are applied to a planet that is normally considered for landing (and planets without a spaceport are generally not), the ai won't ever attempt to land on these planets.
> 
> In other words, I think this use case needs more concept development before any code changes are considered for merges.

Intended implementation is that they're monitored by relatively tight knit pirate clans, with the location doled out to the player upon reaching a certain level of trust with them. Story related missions will add them to already existing uninhabited landables which will be scattered throughout human space. The ultimate intention is to make navigating through hostile territory as a pirate less of a "jump eight systems, run out of fuel, die, reload" slog.

Galaucus:
> Similarly, unless these attributes are applied to a planet that is normally considered for landing (and planets without a spaceport are generally not), the ai won't ever attempt to land on these planets.

For the primary intended application this is all well and good. Fuel caches will - at least by my plans - only be attached to unremarkable asteroids, upon the surface of barren worlds or moons, or otherwise in places where nobody would be going as part of their normal business. 

It might not work for other use cases, such as making a forward outpost/ammo dump/fuel cache for, say, a Republic Navy expedition. Though I suspect that would end up being a full-service spaceport regardless.

InfiniteDonuts:
I am going to push the changes to the pirate campaign fork where they can use it

PR #5996, opened 2021-05-23
Sail-like Turning and Thrusting Attributes

Ferociousfeind:
**Feature:** This PR implements a feature spoken about briefly (by me) on the discord server, and which may be very useful to Vaelophis Nyx, working on her Binary Order, with all those sail-bearing wooden and ironclad spaceships.

## Feature Details
This PR adds two attributes, "sail thrust" and "sail turn" which scale with direction.
"sail thrust" outputs identically to "thrust" while your ship faces directly towards or away from the system center. It outputs 0 thrust while perpendicular to the system center.
"sail turn" outputs identically to "turn" while your ship faces directly towards or against your current velocity. It outputs 0 turn while perpendicular to ship velocity.

Due to this, ships with ONLY "sail thrust" and "sail turn" are a little dysfunctional, as the AI doesn't understand the attribute. Be sure that AI ships always have a little bit of vanilla thrust and turn, or else they could get stuck trying to turn with no wind in their sails.

## UI Screenshots
N/A
no new UI

## Usage Examples
Exactly like "thrust" and "turn", as in these example outfits:
```
outfit "Greyhound Plasma Thruster"
	category "Engines"
	"cost" 45000
	thumbnail "outfit/small plasma thruster"
	"mass" 34
	"outfit space" -34
	"engine capacity" -34
	"thrust" 18.4
	"thrusting energy" 1.8
	"thrusting heat" 3.4
	"sail thrust" 18.4
	"flare sprite" "effect/plasma flare/small"
		"frame rate" 6
	"flare sound" "plasma small"

outfit "Greyhound Plasma Steering"
	category "Engines"
	"cost" 36000
	thumbnail "outfit/small plasma steering"
	"mass" 26
	"outfit space" -26
	"engine capacity" -26
	"turn" 250
	"turning energy" 0.9
	"turning heat" 2.1
	"sail turn" 250
	"steering flare sprite" "effect/plasma flare/small"
		"frame rate" 6
	"steering flare sound" "plasma small"
```

## Testing Done
Used the above outfits on a stock Headhunter, spun around, flew around, jumped from system to system automatically, and with the base turn and base thrust, the AI can jump the Headhunter from system to system just fine.

## Performance Impact
N/A
I don't believe the dot products are very computation intensive. Additionally, if a ship has no "sail turn", the block adding the sail's turn on that frame doesn't run to begin with. If a ship has no "sail thrust", the sail's thrust code doesn't run, either.


petervdmeer:
I like the `"sail thrust"`, it makes sense to have a type of engines that depend on external factors. If you make the effectiveness range from max 100% effective to min 5% effective, then we don't need to always add regular thrust as well (since we don't get the dis-functionality at 0% effective).
Did you consider to make the "sail thrust" also dependent on the amount of solar wind and on the distance from the center?

The `"sail turn"` feels less good to me. It seems conceptually different from "sail thrust" (so I would expect a different keyword with the current implementation) and I don't really see the fun in making small turns more efficient than full turns (especially since we always need full turns for landing and for jumping).
I could however imagine that ships that are more efficient at turning when they are close to the center of the system could be fun (such behavior would give a combat advantage to ships being more in the center or a combat disadvantage for ships more out of the center).

Zitchas:
This sounds like a neat mechanic. Looking forward to trying it out.

Ferociousfeind:
I would like this PR to be somewhat larger-scope than just two attributes, I really wanted to have a whole bunch of different thrusting/turning types which can play nice with one another. So I've labeled it a draft, and will put it back under construction soon.

data/tooltips.txt
```diff
@@ -368,6 +368,9 @@ tip "movement, full / no cargo:"
 tip "thrust:"
 	`Thrust produced by engines. Higher thrust increases a ship's acceleration and top speed. The more mass a ship has, the more thrust is needed to achieve the same acceleration.`
 
+tip "sail thrust:"
+	`Thrust produced by engines, dependant on direction. Sail-style thrusters function optimally when facing directly towards or away from the gravitational center of a system, but facing perpendicularly to the center produces nearly zero thrust.`
```
samrocketman, line 372: Sails (in real life), tend to be most effective when perpendicular to the wind for sloops.  When with an aft wind (rear pushing forward) spinnakers are effective.  The least effective direction for sail power is a headwind across the bow.  To travel into a headwind a sailboat travels diagonally to the wind to perform windward tacks.  This forms a zig-zag pattern while moving up-wind.

If we generically apply this to space; the least effective direction is into the sun.  Circle around the sun for the most effective and moving away from the sun will get you thrust but the further away from the center you get the less thrust.

Ferociousfeind: I wish I knew the mechanics and reasons for this, but I can't personally imagine (and this is personal incredulity yes) how pointing perpendicular to the direction of wind movement would be the most effective or efficient direction of travel...

If there is a better way to do this, I could refactoring all this code. I think a cross product would simplify some of the math considerably, for one.

samrocketman: The sails slightly angle to form an air foil which creates a lower pressure in front while the higher pressure wind pushes up the rear is my understanding.  Worth noting that a sail boat can angle its sails for optimal propulsion regardless of travel direction to move forward so they're fairly efficient except for the head wind.

I tried using terms you could google such as:
- sloop
- windward tack
- leeward tack
- spinnaker

There's a lot more to operating a sail boat but I didn't think it relevant to your PR.

quyykk: Relevant: https://space.stackexchange.com/questions/370/can-you-tack-against-the-sun-using-a-solar-sail

samrocketman: We also have drag in space in ES as well as top speed in space.  So we could explain away the sailing game mechanic the same way we explain drag and top speed.  https://discord.com/channels/251118043411775489/688532441324847150/895606038563278871

Amazinite:
This mechanic doesn't really seem fitting for the game, and feels like it'd be a pain to maintain. I spoke with Ferocious about it and he says he's fine with letting this one go.

PR #5997, opened 2021-05-23
Link on Linux to new OpenGL ABI

sevu:
In the last years the distributions rolled out GLVND.
Since then, libGL.so is a wrapper to other libraries implementing OpenGL.
[A explanation about the changes can be seen here, with a nice diagram.](https://github.com/NVIDIA/libglvnd#architecture)

Together with that was libOpenGL.so added, which is the new library to link to.
On Windows / Mac nothing changed.

## Testing Done
I ran the game on the Intel and Nvidia card and compared the GPU Usage. The stronger Nvidia card displays less percent of workload. So it does have an impact, or in other words: hardware acceleration works.


MCOfficer:
Looking at the CI failures, it seems like this hasn't been rolled out to 16.04 yet - or do we need to install some form of backport package?

Also @tehhowch since 16.04 hit EOL a month ago, perhaps it's time to drop support?

tehhowch:
Yep, this wont be possible to merge until we drop 16.04 support, which will be 0.9.15 at the earliest, and September by the latest (when GitHub drops support).

I'll want to review exactly which large distros _don't_ support this library but do support our current configuration -- if there are a lot then it wouldn't make sense to totally drop this configuration.

sevu:
For Debian the state is, on stretch, the oldstable, ~2017, one may need [this package](https://packages.debian.org/de/stretch-backports/libopengl0) from the backports. Or my not, if one has a [nvidia card](https://packages.debian.org/de/stretch/libopengl0-glvnd-nvidia).

Can you add this to the  0.9.15 milestone, so it won't be forgotten among the other 172 PRs and can be reconsidered at the time?

tehhowch:
I doubt this is viable for the Steam build, too, which is based on Ubuntu 12.

quyykk:
> What's the state of this PR now that we have dropped 16.04 support (https://github.com/endless-sky/endless-sky/commit/aeb612f7c4cfa42729b83d64970ac69a63c8e8c7).

What tehhowch said. I just checked and indeed GLVND is not available on scout.

quyykk:
@sevu Can you update to the latest master? And, to make it continue to link on Steam just do something like this: `"GL" if 'steamrt_scout in chroot_name else "OpenGL"` (same thing for OpenGL ES).

sevu:
Thanks for coming back to this. Updated.

SConstruct
```diff
@@ -114,11 +114,16 @@ elif is_windows_host:
 		"glew32.dll",
 		"opengl32",
 	])
-else:
+elif 'steamrt_scout' in chroot_name:
 	env.Append(LIBS = [
 		"GL",
```
quyykk, line 119: I was thinking of doing it like this: `"GL" if 'steamrt_scout' in chroot_name else "OpenGL"` to avoid duplication. (And we can do the same thing for GLES above).

quyykk (approved):

PR #5999, opened 2021-05-24
Remove incorrect section from Merchant News

Anarchist2:
Since shipyards only sell the fully outfitted versions of ships, the claim that the Falcon is more expensive than the Leviathan is technically incorrect.


Terin:
Could alternatively have the Leviathan fans talk about how the Falcon requires too much crew, and the Falcon fans talk about how the Leviathan doesn't have enough bunks for boarding. But only if you feel up for more writing, just dropping that second part would work fine as well.

Amazinite:
How about "and they're more expensive to make," instead of entirely removing this line about the price difference? The production cost would more clearly point to the hull price as opposed to the sale price with outfit costs included. And then we could have someone talking about the Falcon saying it might be more expensive to make but it's cheaper to buy or something to that effect.

PR #6009, opened 2021-05-26
Document install process for Apple ARM processors

lukearndt:
I just managed to get Endless Sky building on a computer that has an
Apple M1 processor, which has the new ARM architecture. While I didn't
figure out how to make it compile an ARM release build, I was able to
create an x86_64 build that runs on my machine via Rosetta.

This commit adds some information to the developer readme to help out
anyone else who wants to develop the game from a new macOS device.

### DIFF
--- a/readme-developer.txt
+++ b/readme-developer.txt
@@ -85,12 +85,30 @@ To build Endless Sky with native tools, you will first need to download Xcode fr
 
 Next, install Homebrew (from http://brew.sh).
 
+ARM Processors:
+
+  If you have an ARM processor, such as the Apple M1, you will need to install an x86_64 version of Homebrew. You can do this by running the normal install command through an `arch` terminal:
+
+  $ arch -x86_64 zsh
+
+  You may find it useful to define an alias that you can use to interact with your x86_64 homebrew:
+
+  alias brex="arch -x86_64 /usr/local/Homebrew/bin/brew"
+
+  Use this alias instead of `brew` in any subsequent homebrew commands.
+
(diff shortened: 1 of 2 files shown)


tehhowch:
There are no plans for any M1 support or official ARM builds since I don't have any plans to buy an ARM-powered machine (or Mac in general). It seems like the only real change to the dev process is to explicitly state that you need x86-64 versions of the required libraries, which makes sense given that we explicitly build only an x86-64 target.

vitalchip:
Thanks for this, I'm going to buy an M1 (or maybe it'll be an M2) machine at the tail end of the year, I'll keep this tucked away as a reminder.

readme-developer.txt
```diff
@@ -85,12 +85,30 @@ To build Endless Sky with native tools, you will first need to download Xcode fr
...
 
+ARM Processors:
+
+  If you have an ARM processor, such as the Apple M1, you will need to install an x86_64 version of Homebrew. You can do this by running the normal install command through an `arch` terminal:
+
+  $ arch -x86_64 zsh
+
+  You may find it useful to define an alias that you can use to interact with your x86_64 homebrew:
+
+  alias brex="arch -x86_64 /usr/local/Homebrew/bin/brew"
```
petervdmeer, line 96: A readme like this should describe the minimal amount of steps to compile for a platform. Power-user optimizations like defining an alias are not needed for power-users that already know about aliases, and non-power-users could easily be confused by an alias definition.

PR #6020, opened 2021-05-28
Fix indentation in MZ's Finches

Anarchist2:
MZ's Finches had their definitions indented one layer too deep, which may have caused them to be able to be disabled.


Zitchas (approved):
Looks good to me. Thanks for catching this!

10010101001:
Oh well it was fun when it lasted

tehhowch:
While a fix to the unnecessary overindentation, this change won't achieve the end goal. "Uncapturable" and "never disabled" need to be set on the actual ships in the person definition, not on the root-level ship template definitions.

Issue #6016, opened 2021-05-27
Amend Command Center outfit description


2phunkey4u:
**Describe the bug**
The descr of the Command Center says: "more may be necessary if the ship has turrets or other systems that require crew." However, I could one ever buy and install a single one on a captured 349 which was carrying turrets (despite sufficient free outfit space). As a result, the ship, when used as a flagship, would show a crew of -6 and move erratically.

**Expected behavior**
[**_obsolete, see below_**] Be able to outfit any many command centers as necessary or have the command center provide more bunks or have turrets have a crew req of 0 when mounted on automata ships.

beccabunny:
You need bunks

2phunkey4u:
Okay, that's cool, but should be somehow made more explicit in the Command Center descr text.

beccabunny:
It is. "The command center only **provides a single bunk** for the captain, so more *[bunks]* may be necessary if the ship has turrets or other systems that require crew."

2phunkey4u:
I have trouble getting into the PR thing, so I just post the amended file here:
[korath outfits.txt](https://github.com/endless-sky/endless-sky/files/6559131/korath.outfits.txt)
(line 204)

PR #6021, opened 2021-05-28
Clarify Command Center Description

Anarchist2:
This PR specifies that more bunks are needed for the Command Center on a ship with turrets, removing the confusion found in #6016.


Terin:
Looks good, I think capping an automata was the first time I realized turrets required crew, and something like this will help reduce that period of confusion.

PR #6038, opened 2021-06-03
Change highest Heliarch rank

Arachi-Lover:
----------------------
**Word change**

## Summary
Currently, the Heliarchs have "elders" as their members of highest rank. The Hai also have "elders" as their ruling body. The Wanderers as well.
To not have the rank be used so much by very different factions, this changes the instances of "elders" when referring to the Heliarchs to "consuls."
Already updated the Coalition News PR for the instances there as well, I believe don't have any such instances in my other PRs.

It was more discussed in the discord, but in short the ranking ladder I planned would be Cadet -> Agent -> Magister -> Tribune -> Arbiter -> Consul. Only the first two and final one (currently "elder") are used in the game, other three could pop up in missions.

I think I got all instances of "elder" updated, but do tell if I missed any.


Zitchas (approved):
Looks good. I like the idea of differentiating the races/factions a bit more, and having the heliarchs use different terminology helps significantly with that.

More broadly, I like the series of ranks you proposed. The Coalition and the Heliarchs are a relatively massive empire/government, so it makes sense that it would have more hierarchical depth than the simple "regular heliarch" and "elder heliarch" implies.

McloughlinGuy:
Should we change the highest rank of the Wanderers as well?

Zitchas:
> 
> Should we change the highest rank of the Wanderers as well?

It's been a while since I looked into the Wanderers. At the very least, it would be worth taking a look at them and figuring out their roles and titles. I would suggest, however, that those be done in a different issue or PR so that this one can remain focused on the Coalition.

Amazinite:
> Should we change the highest rank of the Wanderers as well?

I don't see this as necessary. Remember that the Wanderers communicate to us via the Hai language. The Hai call their leadership elders. The Wanderers, speaking in Hai, would likely use the same term.

PR #6040, opened 2021-06-05
Grammatical error in Ka'het Shield Restorer

Jugosloven1612:
I propose either "...required that the Builders develop..." or "...required the Builders to develop..."


Amazinite:
Didn't recognize the name so was going to add you to the credits but realized you were already there. Looks like it's been a while since your last PR. :)

PR #6047, opened 2021-06-06
Automatically select newly captured ships

Amacita:
The first thing most players want to do after capturing a ship is to order it away from combat so that it doesn't get killed. As a simple quality of life improvement, a ship should automatically be selected after it is captured, making it faster and easier to issue those urgent, life-saving orders.

Currently, to issue orders to a newly captured ship, the user has to first find the ship on the screen and click it, which doesn't sound too difficult, but a few facts must be kept in mind. First, when there are already a lot of ships nearby, it is very easy to click a different ship by mistake. Second, often the first thing newly captured ships do is aggro towards a group of enemies that will almost instantly destroy it. Third, in particularly chaotic or dangerous battles, newly captured ships can die very quickly; by the time a user has selected the captured ship, it is often too late to save it. Every millisecond counts.

For these reasons, I propose that a ship be automatically selected after capturing it.

### DIFF
--- a/source/BoardingPanel.cpp
+++ b/source/BoardingPanel.cpp
@@ -395,6 +395,9 @@ bool BoardingPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command,
 					}
 				isCapturing = false;
 				
+				// Select the newly captured ship.
+				player.SelectShip(victim.get(), false);
+				
 				// Report this ship as captured in case any missions care.
 				ShipEvent event(you, victim, ShipEvent::CAPTURE);
 				player.HandleEvent(event, GetUI());


tehhowch:
_Unconditionally_ select a ship you just captured? Not going to be accepted.

Doing that makes the assumption that the player either has no selected ships, or that any ships the player has selected already are supposed to avoid combat. The game engine _does not have_ and _can not obtain_ enough information to know what the player wants to do with the set of currently selected ships in order to determine if it should also select a newly captured ship.

The player should be able to drag-select a box around their flagship to select this newly- captured ship and order it elsewhere.

FWIW, escorts in 0.9.13 should be a bit more intelligent about rushing into combat when severely damaged (#3369), though it's possible the recent change to repair past the just-not-disabled hull threshold (https://github.com/endless-sky/endless-sky/pull/5794#pullrequestreview-643248482) nullifies that "have I repaired at least somewhat" logic.

beccabunny:
> Doing that makes the assumption that the player either has no selected ships, or that any ships the player has selected already are supposed to avoid combat. The game engine does not have and can not obtain enough information to know what the player wants to do with the set of currently selected ships in order to determine if it should also select a newly captured ship.

Currently, the game deselects your currently selected ship when you board something and also stops the boarding attempt entirely if you select something else between pressing b and arriving there, so this doesn't apply

Amazinite:
What tehhowch said. Plus if things are running too quick for you to select your newly captured ship because you have fast forward on, that's partially what the fast forward interrupt setting is for to ensure that you come out of the capture panel on 1x speed, giving you more time to select and command your ships than if you were always on 3x speed.

PR #6062, opened 2021-06-10
Nuclear Missile Hardpoints

Azure3141:
**Nuclear Hardpoints**
This gives nuclear missiles a hardpoint sprite. Although it is generally not recommended to do this for guns, I felt that nuclear weapons, as a single-shot weapon, were unique enough to justify a distinction between ordinary guns and missiles. In addition to this, the limited capacity and destructive nature of nuclear weapons can make it very useful for the player to know if a ship is equipped with them, and how many that ship has in reserve.

Overall, this is a very minor PR, but it makes nuclear missiles more visibly noticeable, even before they are fired.

UPDATE: After feedback, I made a custom sprite for the hardpoints instead of reusing the nuclear missile projectile sprite. I think it turned out much better, but am always open to feedback or suggestions to improve it.

**Screenshots**

image

**Testing**
Cheated in several human (and one Remnant) ships, along with nuclear weapons to equip them with

**Performance Impact**
N/A


Amazinite:
I like the idea, but these screenshots don't look that good.

Azure3141:
> 
> 
> I like the idea, but these screenshots don't look that good.

I'm not 100% happy with the way they look on some smaller ships, but until we get a way to offset the hardpoint sprite itself, or have it display under the ship sprite (which I think someone is working on), this is about the best they can be. Overall, since nuclear missiles appear on only a few ships, I think the visual distinction the hardpoints provide makes up for it.

Amazinite:
> but until we get a way to offset the hardpoint sprite itself

This would be solved by creating a new sprite instead of reusing the projectile, so I don't see a need to create a way to shift the sprite.

> (which I think someone is working on)

Do you think or do you know, because if no one's actually working on it, I could probably knock it out real quick.

Amazinite:
> This is a good idea, but I'm not sure how well it would work with ships that have gun ports in different positions. Cutting off the back half of the projectile sprite would work for some ships that have their gun hardpoints at the foward edge of their ship sprites, for example, but would look awkward for any gun ports placed on a slanted edge.

No, the hardpoint sprite would still be the full projectile, you'd just have the center of the image be further forward on the projectile.

The cross further down is the current image center. If the image center were further up, then the hardpoint would appear further back on the ship.
image

Azure3141:
> No, the hardpoint sprite would still be the full projectile, you'd just have the center of the image be further forward on the projectile.
> 
> The cross further down is the current image center. If the image center were further up, then the hardpoint would appear further back on the ship.
> image

Wouldn't this result in the sprite overlapping with the ship sprite, making it look as though the missiles were awkwardly bolted on top of the ship? This would work with underside hardpoint rendering, but with the current system I think it would just look weird.

Amazinite:
> Wouldn't this result in the sprite overlapping with the ship sprite, making it look as though the missiles were awkwardly bolted on top of the ship? This would work with underside hardpoint rendering, but with the current system I think it would just look weird.

Well allowing for hardpoint sprites rendered under the ship is exactly what I'm suggesting to do along side using a unique image for the nuclear missile hardpoint sprite.

Amazinite:
I think it looks pretty bad.

EjoThims:
Really like the concept, but agree that this implementation just isn't quite there.

Even without "under" display, you could just take the top 2/3 or 1/2 of the missile for the hardpoint sprite.

Zitchas:
Just for a starting point, changing it so that it is just the nose of the missile sticking out of the gunport would be a significant improvement

Amazinite:
Closing in favor of #6114.

PR #6071, opened 2021-06-14
Atomic engine sound lowering/softening

ravenshining:
----------------------
**Content (Sound)**

## Summary
This is a set of revised atomic engine sounds.  The aim was to produce sounds that are less distracting and grating, as the current atomics are so painful for me to listen to I eschew them entirely.

I'm no expert, so here's what I did with the current sound:

1. I duplicated the huge atomic sound into two tracks and ran a low and high pass filter on them, respectively, at 650 Hz.
2. The lower frequency track was Paulstretched x6 with a resolution of 0.151 s
3. The higher frequency track was sped up by about 1.54x, Paulstretched by 1.3x, and repeated 3x for the same total length
4. The high and low tracks were mixed back together
5. I duplicated the track again, gave the second track some Cathedral reverb, and mixed back down to mono
6. Deleted the beginning and ending 1/3rds
7. Normalised to -12.5 dB LUFS
8. Produced smaller variants according to the pattern

I find the result to sound much smoother and, as I hoped, less distracting than the original, as well as subjectively quieter while still seeming to be louder than plasma or ion.

## PR Checklist
 - [x] I have updated the copyright for assets modified

-----------------------
## Testing Done

Flying about with ships of each class of atomic engine installed.


Zitchas (approved):
OK, I'm not familiar with the audio terms above.

I tried out the vanilla effects, then tried out this PR. My impression is that the atomic engine in this PR is a little bit rougher, maybe a touch louder, and perhaps most importantly, doesn't pulse as much. (vanilla atomics have a very steady and noticeable pulse to them, in my opinion)

While I do not have a strong opinion on the matter, I do like these revised sounds and in particular I very much appreciate the virtual elimination of the pulsing.

One thing that I note (and this is probably why I haven't noticed the pulse in the atomic engines before now) is that I play ES with the volume low enough that launching and weapons' fire sound effects are at a comfortable level. And when they are at a comfortable level, the engine sounds are so faint as to be overwhelmed by general background noise from my computer, outside, etc. I can only actually hear it when I wear headphones with the volume set high enough that the standard launching noise is uncomfortable.

ravenshining:
Running that many filters I probably did induce some distortion which I guess is what you call "rougher," but my main goal of making it "smoother" was the elimination of that "pulse," which is what I find so grating and distracting.

Funny that you find these louder, though loudness is subjective and affected by things like age, sex, and lifetime noise exposure, particularly when dealing with pitch changes as this does, not to mention the quality of one's sound system.  I think I tend to be more sensitive to higher frequencies than most people, and have a decent stereo system as a side effect of taking music classes at university,  Even ion engines are loud enough to drown out my local ambient noise.

Issue #6077, opened 2021-06-16
Ships displaying slower acceleration/turn rate in flight compared to landed when weighed down by cargo


Amazinite:
**Describe the bug**
Ships that have any cargo space used are displaying accelerations and turn rates while in flight that are inaccurate compared to what is displayed while landed. What exactly seems to be occurring is that for acceleration and turn rate, cargo is being counted twice.

**To Reproduce**
1. Observe the movement stats of a ship when landed, particularly the acceleration and turn rate when cargo is full vs empty.
2. Launch with the ship with no cargo and observe how the stats shown when landed are accurate to the stats when in flight with no cargo (viewed via the ship info panel in your player info menu).
3. Land and fill your cargo hold with commodities.
4. Launch with a full cargo hold and observe how the stats shown when in flight are now less than the stats shown when landed with full cargo.

**Expected behavior**
A ship's stats shown when landed should be accurate to what they are in flight.

**Screenshots**
A Hauler III while landed. Notice that the acceleration and turn when full are stated to be 64.2 and 70.2 respectively.
![image](https://user-images.githubusercontent.com/17688683/122298340-7d339d80-ceb1-11eb-9966-aad30ad64dd2.png)

A Hauler III in flight while on full cargo. Notice how the acceleration and turn are slower than reported while landed.
![image](https://user-images.githubusercontent.com/17688683/122299068-5e81d680-ceb2-11eb-8117-9142b8fa12d7.png)

A Hauler III in flight while on half cargo. Notice how the acceleration and turn are equal to what they should be if on full cargo.
![image](https://user-images.githubusercontent.com/17688683/122298910-2da1a180-ceb2-11eb-8ae0-52c3be57992e.png)

So it seems like cargo is getting double counted for some reason for the purposes of acceleration and turn when in flight, despite the fact that the mass of the ship is accurate to the empty mass + current cargo used. The above screenshots are with filling my cargo with commodities, but using outfits of equal mass yields the same results.

Whether these displayed stats are accurate to how the ship actually performs, I'm not sure.

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Self-compiled
 - Version: I'm actually using #5779 right now, but this bug was also reported on [Steam](https://steamcommunity.com/app/404410/discussions/1/3076503022314202472/).

**Link to save file**
N/A

**Additional context**
N/A

Amazinite:
I think I've found the culprit, and luckily this is just a display error.

[...]

PR #6078, opened 2021-06-16
Don't count cargo mass twice while in flight in ShipInfoDisplay

Amazinite:
**Bugfix:** This PR addresses issue #6077.

## Fix Details
Renamed emptyMass to baseMass in ShipInfoDisplay, as baseMass is a more accurate description of what Ship::Mass returns. Then, if in flight, fullMass is now equal to baseMass, as baseMass already accounts for cargo space used.

## Testing Done
Tested the same scenarios as shown in #6077 (half cargo used and full cargo used) and observed results that are accurate with what is shown when landed.

## Save File
N/A

### DIFF
--- a/source/ShipInfoDisplay.cpp
+++ b/source/ShipInfoDisplay.cpp
@@ -169,7 +169,7 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 		attributeValues.push_back(Format::Number(attributes.Get("hull")));
 	}
 	attributesHeight += 20;
-	double emptyMass = attributes.Get("mass");
+	double emptyMass = attributes.Mass();
 	double currentMass = ship.Mass();
 	attributeLabels.push_back(isGeneric ? "mass with no cargo:" : "mass:");
 	attributeValues.push_back(Format::Number(isGeneric ? emptyMass : currentMass));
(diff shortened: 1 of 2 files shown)


Amazinite:
Looks like GitHub Desktop has some new features to reorder and squash commits, then force push. Interesting.

Zitchas:
Just out of curiosity, how many terms of mass are there in use? This might be a situation where the various terminology already established by the transportation/shipping industry might be useful. dry mass versus wet mass, GVW, etc.

petervdmeer (commented):
The code-change itself looks correct, but I'm affraid that this code is based on assumptions that might not hold in the future. If we ever change change the code to keep ships loaded while they are landed, then we could get new display-bugs here because of somebody forgetting to update the code here.

I would prefer that we retrieve both the `currentMass` and the `emptyMass` (and maybe also the `fullyLoadedMass`) from the ship directly (currentMass is what is now in the baseMass variable) and emptyMass could be gotten through ship.Attributes()["Mass"] and use the `emptyMass` for printing in case of generic, while we use the `currentMass` when displaying the "in flight" version.

Amazinite:
@petervdmeer does that look good?

source/ShipInfoDisplay.cpp
```diff
@@ -169,9 +169,10 @@ void ShipInfoDisplay::UpdateAttributes(const Ship &ship, const Depreciation &dep
 		attributeValues.push_back(Format::Number(attributes.Get("hull")));
 	}
 	attributesHeight += 20;
-	double emptyMass = ship.Mass();
+	double emptyMass = attributes.Get("mass");
```
petervdmeer, line 172:
```suggestion
	double emptyMass = attributes.Mass();
```
Sorry, my mistake in my previous comment, Mass is a special case and stored directly, not in the map (similar to Cost()).

petervdmeer:
> @petervdmeer does that look good?

Yes, it looks just like I expected, but I just noticed that one of my remarks was a little bit off (see my latest comment).

petervdmeer (approved):
Looks good to me.

Issue #1428, opened 2016-08-12
"Suicidal" fleet personality


Amazinite:
Currently, the AI will not fire a weapon if it will end up hurting itself in the process, usually through blast radii. The suicidal personality would override this, and maybe also cause the ships in the fleet to try and get as close as possible to their target, opening themselves to more fire.

Amazinite:
A less "triggering" word could probably be "reckless."

KiLEdEnNis:
suicidal reminds me of an idea I had; small ships/drones that can self-destruct for a big AoE damage (similar to a ship with a weapon with a huge radius, hit force & damage), but suicide bomb drones should get a special unremovable weapon that blows themself up but does huge damage
that would be cool for suicidal, but reckless is also a good idea

endless-sky:
It's relatively easy right now to trick the AI into thinking it won't damage itself with a weapon, by putting the blast radius in a submunition rather than the main projectile.

petervdmeer:
I feel multiple personalities could be appropriate, for example:
- `"kamikaze"` for ships that damage themselves as long as they can do more damage to an enemy.
- `"suicidal"` for ships that fire weapons that damage themselves anytime.
- `"reckless"` for ships that take large risks.
- `"take with me"` for ships that try to take enemies with them when they are about to die.

Zitchas:
I'd suggest "careless" instead of "suicidal", but otherwise, yeah, I like the sounds of those.

tehhowch:
> I feel multiple personalities could be appropriate, for example:
> - `"kamikaze"` for ships that damage themselves as long as they can do more damage to an enemy.
> - `"suicidal"` for ships that fire weapons that damage themselves anytime.
> - `"reckless"` for ships that take large risks.
> - `"take with me"` for ships that try to take enemies with them when they are about to die.

I think "kamikaze" better suits the "take with me" behavior, while something relating to attrition suits the first behavior.

gunqqer:
I think `kamikaze` is for when they'll just do as much damage as possible if it does more to the enemy no matter their own health, but `take with me` will only do that if they are at low health. So like, if they're about to die they will launch a nuke at point blank range, but at full health they won't.

Amazinite:
[...]

PR #6084, opened 2021-06-24
Allow weapons to disregard whether the firing hull would cause the wielding ship to become disabled

Amazinite:
**Feature:** This PR implements a feature request discussed in relation to #5818.

## Feature Details
This PR adds a new weapon attribute:
* `"disregard firing hull"`: A tag that allows weapons to fire regardless of whether the firing hull cost would cause the firing ship to become disabled.

## UI Screenshots
N/A

## Usage Examples
```
outfit "Ka'het Emergency Deployer"
	...
	weapon
		...
		"disregard firing hull"
		"firing hull" 60
		...
```

## Testing Done
None at all.

## Performance Impact
N/A

### DIFF
--- a/source/Weapon.cpp
+++ b/source/Weapon.cpp
@@ -50,6 +50,8 @@ void Weapon::LoadWeapon(const DataNode &node)
 			isParallel = true;
 		else if(key == "gravitational")
 			isGravitational = true;
+		else if(key == "disregard firing hull")
+			disregardFiringHull = true;
 		else if(child.Size() < 2)
 			child.PrintTrace("Skipping weapon attribute with no value specified:");
 		else if(key == "sprite")
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -3409,8 +3409,10 @@ bool Ship::CanFire(const Weapon *weapon) const
 	if(fuel < weapon->FiringFuel() + weapon->RelativeFiringFuel() * attributes.Get("fuel capacity"))
 		return false;
 	// We do check hull, but we don't check shields. Ships can survive with all shields depleted.
-	// Ships should not disable themselves, so we check if we stay above minimumHull.
-	if(hull - MinimumHull() < weapon->FiringHull() + weapon->RelativeFiringHull() * attributes.Get("hull"))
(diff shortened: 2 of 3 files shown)


source/Weapon.cpp
```diff
@@ -50,6 +50,8 @@ void Weapon::LoadWeapon(const DataNode &node)
 			isParallel = true;
 		else if(key == "gravitational")
 			isGravitational = true;
+		else if(key == "disregard firing hull")
```
petervdmeer, line 53: I feel that this shouldn't be a weapon property, but an AI personality property; Some AI ships should fire weapons that can disable or destroy themselves, while other AI ships should never fire weapons when those would destroy themselves.

Amazinite: So #1428? I suppose MZ's comment is now out of date given that you can't trick a ship into using a firing hull weapon that can disable itself.

Zitchas: I can see use cases for it either way. 

If we have a situationwhere such a weapon is found on a lot of ships, but only some ships are going to be sufficiently desperate to use it, then it should be a personality thing.

If it is only going to be installed on ships that are by definition desperate enough to use it, then just have the attribute on the weapon.

That being said, if the default behaviour is "do not use a weapon unless they can do so and survive," what about an attribute for "will use this weapon regardless of its effect, but only after all other secondaries are used up and/or it looks like I am about to die anyway?"

petervdmeer: > So #1428? I suppose MZ's comment is now out of date given that you can't trick a ship into using a firing hull weapon that can disable itself.

Yes, let's go for #1428. "being able to trick the AI" sounds to me more like a (short-term) workaround to get something done than a (long-term) desired solution.

petervdmeer: > That being said, if the default behaviour is "do not use a weapon unless they can do so and survive," what about an attribute for "will use this weapon regardless of its effect, but only after all other secondaries are used up and/or it looks like I am about to die anyway?"

I just posted a comment on #1428 about having more than 1 relevant personality. (#1428 seems to be the relevant ticket to discuss that).

Amazinite:
Closing in favor of a ship personality that covers more cases than just firing hull.

PR #6087, opened 2021-06-25
build: Add continuous target for v0.9.15

Terin:
**Build:** Adds the v0.9.15 branch to CD for continuous

-----------------------

## Details
This PR updates the workflow files to target the v0.9.15 branch in addition to master, for the v0.9.15 branch. From my testing, whenever any branch is run, it'll look to the workflows in that branch for what to do. This will let v0.9.15 affect the continuous release, without any changes to master. This will let people download the current changes via the release page, or ESLauncher2.

Once v0.9.15 is ready to merge when MZ lives, this change can be reverted.

## Testing Done
Push this change to v0.9.15 on my repo, then pushed another commit, which was properly added to the continuous release once CI/CD ran.

https://github.com/Terin/endless-sky/releases/tag/continuous

### DIFF
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -3,7 +3,6 @@ name: CD
 on:
   push:
     branches:
-      - master
       - v0.9.15
   pull_request:
     types:
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - v0.9.15
   pull_request:
     types:
       - synchronize
--- a/.github/workflows/cd.yaml
+++ b/.github/workflows/cd.yaml
@@ -3,6 +3,7 @@ name: CD
 on:
   push:
     branches:
+      #- master
       - v0.9.15
   pull_request:
     types:


.github/workflows/cd.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
```
MCOfficer, line 6:
```suggestion
#      - master
```

Let's disable this, otherwise once MZ commits on master, the (then outdated) master will be published as continuous.

MCOfficer (changes requested):

.github/workflows/cd.yaml
Terin: We would need to modify the cd.yml in the master branch for that to work, because it looks to the workflows in whatever branch runs.

I figured it wouldn't be too much of a problem, since we'd be a few hours out from merging 0.9.15 back in once MZ makes the release. If one of the maintainers wants continuous to be completely removed from master on both branches, though, I'll rebase the PR to that (can a PR be applied to multiple branches?).

MCOfficer: Right, that's inconvenient :/

Either we deem those few hours acceptable, or we gotta change this on master as well (which might give a small merge conflict later)

Terin:
@tehhowch , we have no idea how long it will take for MZ to return. Easy access to the continuous builds are pretty important for the lifeblood of this community, which is made up of people who are not familiar with git/code.

At the very least, could I open a "PR" that is just current with v0.9.15, so people using ESL2 can just grab that build and keep up with it. Unless you can guarantee than MZ is back in like, the next week or two, we need to operate as if he's gone indefinitely.

Amazinite:
I've opened such a pr: #6088. We'd open a new PR for v0.9.15 eventually, so no harm in doing it early so that the launcher can make use of it.

PR #6089, opened 2021-06-27
Fix Coalition news location attributes

Anarchist2:
The current Kimek Farmer news uses the `or` syntax of the attributes line, instead of the intended `and`.


Anarchist2:
Closing in favour of Terin's PR.

PR #6090, opened 2021-06-27
Added government filters to all Coalition news

Terin:
**Bugfix:** This PR addresses an issue from the Discord

## Fix Details
image
From dankdmitron on the Discord, due to the news attribute filter only checking to see if one of the attributes matches at a given spaceport, and not all of them, the "kimek farmer" news item could show up on any worlds with the "farming" attribute. This is mostly noticeable on something like Vara Rakak in Wanderer space, which didn't have news until just today.

This PR adds a government filter to the rest of the Coalition news items. It is currently redundant on the rest, since arach/kimek/saryd/longcow doesn't show up outside of Coalition space, but it helps to be on the safe side.

## Testing Done
Went to Vara Rakak without the Wanderer news, and saw it was empty, and saw the kimek farmer news item show up in Coalition space.

### DIFF
--- a/data/coalition/coalition news.txt
+++ b/data/coalition/coalition news.txt
@@ -36,6 +36,7 @@ news "heliarch propaganda"
 
 news "longcow rancher"
 	location
+		government "Coalition"
 		attributes "arach" "longcows"
 	name
 		word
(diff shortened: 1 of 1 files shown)


Amazinite:
Already fixed with 9d82cba (which actually used the change from #6089 and this one).

Issue #6074, opened 2021-06-15
Captured fighters don't dock in free escort bays


sodium70mg:
**Describe the bug**
Captured fighters and drones don't seem to recognize that a player's escort has available bays if the bays are at least half occupied already. For example, if the player ship has no bays and the player's escort, a Skein, has 3 of its 6 fighter bays occupied, then any fighter ships that are captured won't be able to dock with the Skein in a reasonable amount of time. It looks like the fighter's `AI::Step` function checks to see if a bay is available and if the check fails then the fighter selects a random ship to escort for a while. If the randomly selected ship is the Skein, then the fighter docks as expected. Until that random selection it's possible to order each non-fighter, non-drone escort to hold position at different spaced out locations and watch the captured fighter periodically fly from ship to ship as it selects them at random even though there are free bays in the fleet.

There's still a delay before the fighter docks even when the Skein has all bays free. This seems to be caused by the captured fighter always starting as an escort of the player ship even when the player ship has no bays.

From looking at the code a bit, I think the issue is in Ship.cpp in the `Ship::CanCarry` method. The `free` value is decremented in a loop based on number of escorts matching the bay category but the occupied bays are already accounted for since they aren't free. This causes the docked escorts to be counted twice. I think the countdown in the loop should start with the total number of matching bays for an accurate count.

**To Reproduce**
1. Pilot a ship with no bays but have a ship such as the Skein with half its bays full as an escort
2. Capture a fighter
3. Wait for the fighter's `AI::Step` function to run
4. Watch the fighter select a ship at random to escort instead of the Skein (though it might get lucky)

**Expected behavior**
Fighters and drones should be able to identify escort ships with available bays even when the bays are nearly full.

**Screenshots**
none

**System (please complete the following information):**
 - OS: Linux 5.12.10-arch1-1 SMP PREEMPT x86_64 GNU/Linux
 - Game Source: Steam beta, self-compiled (both)
 - Version: v0.9.13

**Link to save file**
https://gist.github.com/sodium70mg/7d2b8351eec52650a12f506334ee47af#file-fighter-tester-txt

This file has a decent fleet with beam weapons heading into Kor Ak'Mari which is an easy farming spot for Korath Chasers.  There's a skein escort ready to go and 3 barbs parked on the planet that can be un-parked to take up half the bays for demonstration/comparison purposes.

Of course this could be reproduced anywhere as long as the conditions described apply.

**Additional context**
N/A

petervdmeer:
[...]

PR #6091, opened 2021-06-27
Fix captured fighters don't dock in free escort bays #6074

benflodge:
**Bugfix:** This PR addresses issue #6074 

## Fix Details
When calculating the free bays bays in a ship the escorts in bays are not counted twice, this allows escorts of the correct category to dock in the ships free bays.

Notice: Once a captured ship has docked in a bay it will not be able to launch until its heath is over 75% (https://github.com/endless-sky/endless-sky/blob/master/source/AI.cpp#L122).

## Testing Done
This fix was tested with the save file provided

## Save File
This save file can be used to verify the bugfix. The bug will occur when using e093e13037e2ca92db97306ae88cb1934fb86f94, and will not occur when using this branch's build.
https://gist.github.com/sodium70mg/7d2b8351eec52650a12f506334ee47af#file-fighter-tester-txt


source/Ship.cpp
```diff
@@ -3164,7 +3179,7 @@ bool Ship::CanCarry(const Ship &ship) const
 	for(const auto &it : escorts)
 	{
 		auto escort = it.lock();
-		if(escort && escort->attributes.Category() == category)
+		if(escort && escort->attributes.Category() == category && !IsShipInBays(escort))
```
petervdmeer, line 3182: I feel that changing `BaysFree(...)` on line 3160 (3175 in the new situation) to `BaysTotal(...)` is a more efficient way of  fixing the bug, since the IsShipInBays-function-call introduces an additional loop per escort-ship, while changing line 3160 would actually reduce the amount of code that is run a tiny bit. (And this code runs in the real-time part of the game, so efficiency and performance are relevant here.)

petervdmeer (changes requested):

source/Ship.cpp
petervdmeer: And maybe also add a check here if `(escort ==ship)`, to make sure that CanCarry(...) doesn't count the current ship double.

benflodge: Yes that would be cleaner and more performant.
I was unsure if existing escorts had priority over others, however it looks like the escorts just take the first available space.

benflodge: Thanks for the feedback @petervdmeer, I have pushed an update: https://github.com/endless-sky/endless-sky/pull/6091/commits/20979e5a75c459c669896e7a0a42c54fcff9a420

petervdmeer (approved):
Looks good to me

vitalchip:
I wonder if it's worth also adding && !escort->IsDestroyed() to the check as currently when one of your carried ships gets killed it doesn't release the bay until it gets pruned.

i.e:

if(escort && escort.get() != &ship && escort->attributes.Category() == category && !escort->IsDestroyed())

petervdmeer:
> I wonder if it's worth also adding && !escort->IsDestroyed() to the check as currently when one of your carried ships gets killed it doesn't release the bay until it gets pruned.
> 
> i.e:
> 
> if(escort && escort.get() != &ship && escort->attributes.Category() == category && !escort->IsDestroyed())

Yes, that is smart, and maybe also add `!escort->IsDisabled()` to free up space from disabled escorts.

benflodge:
That sounds like a good plan. As far as I can tell `IsDisabled` should be enough to tell if a ship is disabled or destroyed, however its cheaper to check `IsDestroyed` before checking `IsDisabled` as it does a bit more computation.

tehhowch:
I agree with dropping destroyed escorts from the list, but not with dropping disabled ones.

Other than that, looking good!

benflodge:
Ok I've have pushed an update now to remove destroyed escorts, but not the disabled ones.

petervdmeer (approved):

Issue #4851, opened 2020-02-26
Tweak disable threshold formula.


MasterOfGrey:
In the early game I find it particularly frustrating that medium warships in particular are quite hard to capture and prone to being annihilated by any passing shot.

Obviously since [#3448](https://github.com/endless-sky/endless-sky/issues/3448) the formula has either been made smoother, or #3448 wasn't entirely correct, and we now have:

=FLOOR(HullMax*MAX(0.15,MIN(0.45,10/SQRT(HullMax))),1)

This is in theory fair, but the amount of hull remaining, while scaling somewhat logically by ship size, does not scale very well with the amount of damage that's likely to be incoming at the point of being disabled.

It also seems to me a tad odd that heavy human warships are capable of holding out as long as much more advanced alien warships (relatively).

I propose tweaking the inflection point slightly with this version:

=FLOOR(HullMax*MAX(0.15,MIN(0.45,14/SQRT(HullMax))),1)

The distinction is subtle, but this means the curve for remaining hull points actually doesn't flatten out and hit the 15% fixed slope until about the same max hull value as the Bactrian (technically slightly more than it). This would mean that all T1 human ships sit somewhere on the curve, making medium warships a little more likely to survive battle in a disabled state, without significantly impacting the survival rates of things like Leviathans or other craft on the small/big ends of the scales (there is some impact but it's minor), and also thematically suits the idea that there's a certain threshold of engineering that Humanity is only just approaching, to be able to keep functioning in such advanced damaged states.

Amazinite:
I'd take it.

PR #6092, opened 2021-06-27
Change Default MinimumHull Formula

Ferociousfeind:
**Feature:** This PR implements the feature request detailed and discussed in issue #4851, #3448.

## Feature Details
Changes the calculation of a ship's default minimum hull from a mix of 15%, 45%, and 10/sqrt(hull amount), to a continuous curve moving from 75% to 10% roughly over the spectrum of human ship values. A table to give a rough feel for the change (it is not a very large change in actuality)
100 hull = 69.1 minimum
200 hull = 128.3 minimum
300 hull = 180 minimum
400 hull = 225.7 minimum
500 hull = 266.7 minimum
625 hull = 312.5 minimum (the 50% mark)
1000 hull = 425 minimum
1500 hull = 540 minimum
2000 hull = 633.3 minimum
3000 hull = 787.5 minimum
3333.3 hull = 833.3 minimum (the 25% mark)
4000 hull = 920 minimum
5000 hull = 1041.7 minimum
10000 hull = 1590.1 minimum
25000 hull = 3125 minimum (the 12.5% mark)

## UI Screenshots
image
The Wardragon has enough hull to breach the 15% line (its minimum hull is at 11.27%)
image
The Gull has 2500 hull, which equates to 714.3 hull, or 28.57%

## Usage Examples
N/A

## Testing Done
Purchased various ships and checked their displayed max hull and the size of the "disable" ring in-flight against the chart. For a moment I was concerned because I was checking shield values and they weren't matching up, but indeed the formula is working as intended.

## Performance Impact
Depending on how intensive a series of min()s and max()s are... MinimumHull() isn't called that often either way.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -3614,7 +3614,8 @@ double Ship::MinimumHull() const
 		return absoluteThreshold;
 	
 	double thresholdPercent = attributes.Get("threshold percentage");
-	double minimumHull = maximumHull * (thresholdPercent > 0. ? min(thresholdPercent, 1.) : max(.15, min(.45, 10. / sqrt(maximumHull))));
+	double transition = 1 / (1 + 0.001 * maximumHull);
+	double minimumHull = maximumHull * (thresholdPercent > 0. ? min(thresholdPercent, 1.) : 0.1 * (1. - transition) + 0.75 * transition);
 
 	return max(0., floor(minimumHull + attributes.Get("hull threshold")));
 }
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -3614,8 +3614,8 @@ double Ship::MinimumHull() const
 		return absoluteThreshold;
 	
 	double thresholdPercent = attributes.Get("threshold percentage");
-	double transition = 1 / (1 + 0.0006 * maximumHull);
-	double minimumHull = maximumHull * (thresholdPercent > 0. ? min(thresholdPercent, 1.) : 0.05 * (1. - transition) + 0.75 * transition);
+	double transition = 1 / (1 + 0.0005 * maximumHull);
+	double minimumHull = maximumHull * (thresholdPercent > 0. ? min(thresholdPercent, 1.) : 0.1 * (1. - transition) + 0.5 * transition);
 
 	return max(0., floor(minimumHull + attributes.Get("hull threshold")));
 }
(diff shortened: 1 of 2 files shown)


MasterOfGrey:
Can confirm this satisfactorily addresses #4851

Ferociousfeind:
oops, renamed a bunch of my branches and apparently that broke the connection between the branch and the PR

source/Ship.cpp
```diff
@@ -3614,7 +3614,8 @@ double Ship::MinimumHull() const
 		return absoluteThreshold;
 	
 	double thresholdPercent = attributes.Get("threshold percentage");
-	double minimumHull = maximumHull * (thresholdPercent > 0. ? min(thresholdPercent, 1.) : max(.15, min(.45, 10. / sqrt(maximumHull))));
+	double transition = 1 / (1 + 0.0006 * maximumHull);
+	double minimumHull = maximumHull * (thresholdPercent > 0. ? min(thresholdPercent, 1.) : 0.05 * (1. - transition) + 0.75 * transition);
```
Amazinite, line 3618: I like these numbers better.
```suggestion
	double transition = 1 / (1 + 0.0005 * maximumHull);
	double minimumHull = maximumHull * (thresholdPercent > 0. ? min(thresholdPercent, 1.) : 0.1 * (1. - transition) + 0.5 * transition);
```
A similar impact on ships in the middle with less of an impact for ships on the high end compared to the previous numbers and almost no impact compared to the current equation for ships on the lowest end.

Ships from 0 to 600 hull get disabled at near the same point as the current equation.
Ships from 600 to 14000 hull get disabled sooner. (The peak change is that ships around 4500 hull have a disabled threshold about 50% higher.)
Ships above 14000 hull get disabled later.

Amazinite (changes requested):
I like the form of the equation as a replacement for the existing one, but I'm not as on board with the exact constants used.

Amazinite:
We'll also need to change the hull tooltip along side this.
```
tip "hull:"
	`Total hull strength. Small ships are disabled when their hull strength drops below 45%; larger ships are disabled at 15%.`
```
Instead of giving exact numbers, perhaps just mention that smaller ships are more easily disabled (as a fraction of their total hull) than larger ships.

Ferociousfeind:
Oh, I see. Running these numbers through desmos, that formula seems much less drastic than my change. I'll take it! I'll also adjust the tooltip for hull. At low hull, the minimum hull is around 50%, and the lowest minimum hull is 15% based on the numbers

MasterOfGrey:
Ah yes, nice edits.

Ferociousfeind:
Heeeeey wait, did that not push the changes through? I thought that "commit suggestion" would, you know, commit the suggestion to the PR. It seems it actually did not do this. Let me go manually do it...

PR #6093, opened 2021-06-28
feat: Fines in mission actions

petervdmeer:
**Feature:** This PR implements the **fines** part of the feature requested and discussed in issue https://github.com/endless-sky/endless-sky/issues/1216

## Feature Details
This PR adds a keyword `fine` that can be used in `on accept`, `on complete`, `on <...>` mission actions.
Attached is a data-file [fines.txt](https://github.com/endless-sky/endless-sky/files/6728590/fines.txt) with 4 missions that can be used to test this feature (the missions are quite silent).

This PR only implements the `fine` keyword, it doesn't implement keywords to add a customized mortgage. I consider it an implementation detail that fines are handled as mortgages, so I consider it appropriate to use a separate keywords for fines and for mortgages (the implementation in https://github.com/endless-sky/endless-sky/pull/5073, minus the types, would then be suitable to handle the mortgages part).

## UI Screenshots
Here is a screenshot of all the fines from the datafile above applied:
bank_panel_fines

## Usage Examples
Example usage:
```
mission ...
  ...
  on ...
    fine 800000
```

## Testing Done
I tested this feature with the 4 missions from the datafile above. Including a check if instantiated mission action fines are written to the savegame.

## Performance Impact
N/A

### DIFF
--- a/source/MissionAction.cpp
+++ b/source/MissionAction.cpp
@@ -230,6 +230,15 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			if(child.Size() >= 3)
 				paymentMultiplier += child.Value(2);
 		}
+		else if(key == "fine")
+		{
+			if(child.Size() < 2)
+				child.PrintTrace("Skipping invalid \"fine\" without amount entry:");
+			else
+				fine += child.Value(1);
+			if(child.Size() > 2)
+				child.PrintTrace("Skipping non-defined parameters after \"fine\" value entry:");				
+		}
 		else if(key == "event" && hasValue)
 		{
 			int minDays = (child.Size() >= 3 ? child.Value(2) : 0);
@@ -320,6 +329,8 @@ void MissionAction::Save(DataWriter &out) const
 			out.Write("require", it.first->Name(), it.second);
 		if(payment)
 			out.Write("payment", payment);
+		if(fine)
+			out.Write("fine", fine);
 		for(const auto &it : events)
 		{
 			if(it.second.first == it.second.second)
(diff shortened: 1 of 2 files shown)


source/MissionAction.cpp
```diff
@@ -230,6 +230,21 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
 			if(child.Size() >= 3)
 				paymentMultiplier += child.Value(2);
 		}
+		else if(key == "fine")
```
Amazinite, line 233: Since a value is necessary, why not check hasValue? The only other keywords that don't check hasValue have some sort of default behavior when no value is given.
```suggestion
		else if(key == "fine" && hasValue)
```

source/MissionAction.cpp
```diff
@@ -230,6 +230,21 @@ void MissionAction::Load(const DataNode &node, const string &missionName)
...
+			else
+			{
+				int64_t loadedFine = child.Value(1);
+				if(loadedFine > 0)
+					fine += loadedFine;
+				else
+					child.PrintTrace("Skipping invalid \"fine\" with non-positive value:");					
+			}
+			if(child.Size() > 2)
+				child.PrintTrace("Ignoring non-defined parameters after \"fine\" value entry:");
```
Amazinite, line 246: I don't believe we typically warn about excess parameters elsewhere.

petervdmeer: We might want to add more parameters in the future, so I removed the check.

source/MissionAction.cpp
petervdmeer: Good suggestion, changed as suggested.

Amazinite (approved):

Issue #4100, opened 2018-12-28
Ability to reliably transfer cargo between ships in fleet.


Joeppie:
As far as I can tell, It is not reliably possible to transfer cargo between vessels of your fleet.

For instance, if I am using my warship for mining iron, and my fleet has capacity to spare, I cannot properly get this transferred to another vessel in my fleet. This can help avoid making a landing necessary, which is an inefficient logistic operation that cuts into my profit margins.

What I can do is press [i] for ship info, then jettison a small amount of cargo, which can then be picked up by my other vessels. Sadly, however, this is not reliable, and there is a 50-50 chance your cargo is then lost to space, as you can no longer pick it up.

There are two possible options to resolve this, from my current understanding;

1. implement a button that lets you transfer cargo between fleet vessels, similar to drop cargo, but which then presents a panel to transfer the cargo to another vessel. (alternately, it could just try to cram as much of the cargo onto any other vessel and keep whatever flows over; this might be the most user-friendly implementation and save developer time)
2. somewhat less elegant; make jettisoned cargo not disappear as quickly, and reliably jettison in one direction, optionally making it possible to pick it up after a short duration by the vessel which dropped it.

A possible work around might be if it was possible to change flagships while in space; even if cargo was jettisoned in the most imbecilic manner, a skilled pilot would still be able to safely retrieve it from aboard another vessel. This gave me an idea: I will submit it as a feature request if no similar issues exist; escape pods for captains/crew (it might be fun if exploding ships jettisoned escape pods, yielding 'passengers' that would be okay with going to any (friendly?) port)

Zitchas:
While option 1 would be good, I would really like to see 2 implemented anyway. I've always thought cargo disappears far too quickly.

Bladewood:
Just fly directly over your escort and dump cargo there.

Joeppie:
> Just fly directly over your escort and dump cargo there.

This is not a possibility; the cargo, when dumped, takes a random direction, and worse, if your escort misses it, it will likely disappear. Not even using [h] to order the fleet to  halt can stop that.

tehhowch:
Unlikely. Improving the current system has been brought up before, and the inertia in changing it is that it is simply too easy to make money plundering if you can guarantee the use of your flagship and your fleet cargo space. If plundering has a different limit (cargo space is currently the only limit), a better transfer system is feasible.

Here's a related issue -- there may be others, and probably also some steam threads, too.
[...]

PR #6095, opened 2021-06-30
feat: Ignore flotsam that was explicitly dumped by same government

petervdmeer:
**Feature:** This PR implements a feature to improve the appeasing/flotsam dump behavior

## Feature Details
This PR let's ship ignore (not pickup) flotsam that was explicitly dumped by other ships of their own government.

Only explicit dumps (for example to appease pirates) are ignored; ships still pickup flotsam from same-gov ships that explode.
Explicit dumps are triggered in AI.cpp (for NPCs) and in ShipInfoPanel (by the player).

Related to https://github.com/endless-sky/endless-sky/issues/4100 since this PR removes the long-life flotsam option as mentioned in that PR.

## UI Screenshots
N/A

## Usage Examples
Player playing as a pirate: intimidate merchants and notice that the flotsam that they dump doesn't get collected by other merchants.
Player playing as merchant: dump cargo and notice that your escorts don't pick it up.
Attached is a savegame [Flotsam Dumper.txt](https://github.com/endless-sky/endless-sky/files/6743608/Flotsam.Dumper.txt) to test this.

## Testing Done
- Loaded attached savegame.
- Departed
- Shot some asteroids.
- Picked up the flotsam (by own ships and by cargo escort ships).
- Wait for cargo ships to fully surrounded my shuttle.
- Dumped cargo.
- Noticed that my escorts didn't pickup the flotsam dumped by me.

## Performance Impact
No significant impact expected.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -2046,7 +2046,8 @@ void Engine::DoCollection(Flotsam &flotsam)
 	for(Body *body : shipCollisions.Circle(flotsam.Position(), 5.))
 	{
 		Ship *ship = reinterpret_cast<Ship *>(body);
-		if(!ship->CannotAct() && ship != flotsam.Source() && ship->Cargo().Free() >= flotsam.UnitSize())
+		if(!ship->CannotAct() && ship != flotsam.Source() && ship->Cargo().Free() >= flotsam.UnitSize() &&
+			ship->GetGovernment() != flotsam.SourceGovernment())
 		{
 			collector = ship;
 			break;
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -2047,7 +2047,7 @@ void Engine::DoCollection(Flotsam &flotsam)
 	{
 		Ship *ship = reinterpret_cast<Ship *>(body);
 		if(!ship->CannotAct() && ship != flotsam.Source() && ship->GetGovernment() != flotsam.SourceGovernment()
-			&& ship->Cargo().Free() >= flotsam.UnitSize()
+			&& ship->Cargo().Free() >= flotsam.UnitSize())
 		{
 			collector = ship;
 			break;
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -2046,8 +2046,8 @@ void Engine::DoCollection(Flotsam &flotsam)
 	for(Body *body : shipCollisions.Circle(flotsam.Position(), 5.))
 	{
 		Ship *ship = reinterpret_cast<Ship *>(body);
-		if(!ship->CannotAct() && ship != flotsam.Source() && ship->Cargo().Free() >= flotsam.UnitSize() &&
-			ship->GetGovernment() != flotsam.SourceGovernment())
+		if(!ship->CannotAct() && ship != flotsam.Source() && ship->GetGovernment() != flotsam.SourceGovernment()
+			&& ship->Cargo().Free() >= flotsam.UnitSize()
 		{
 			collector = ship;
 			break;
--- a/source/Ship.h
+++ b/source/Ship.h
@@ -366,8 +366,8 @@ public:
 	CargoHold &Cargo();
 	const CargoHold &Cargo() const;
 	// Display box effects from jettisoning this much cargo.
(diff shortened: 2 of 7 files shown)


Zitchas:
I've always wished I could do something like make a ship both appeasing and harvesting without having them chase their own stuff. 

Thought: Maybe ships should pick their own deliberately dumped cargo back up; but only if there are no hostiles and nothing else to pick up? So, for merchants, it'd be kind of a "well, the pirates I was trying to appease got blown up, so I'll go back and pick up my cargo" sort of thing.

Amazinite (changes requested):
As far as a change for NPCs goes I think this is fine, as you wouldn't want a merchant to dump cargo in order to distract pirates only to have another nearby merchant accidentally pick it all up, but this completely removes the ability for a player to purposefully dump their cargo in order to have their escorts pick it up, which is not something that I think we should be restricting. While not a common action, it is something I make use of from time to time, and I'm sure there are others that make use of dumping flagship cargo for their escorts to pick up far more often than I do for plundering purposes.

petervdmeer:
> <...> this completely removes the ability for a player to purposefully dump their cargo in order to have their escorts pick it up, which is not something that I think we should be restricting. While not a common action, it is something I make use of from time to time, and I'm sure there are others that make use of dumping flagship cargo for their escorts to pick up far more often than I do for plundering purposes.

I know, I also used this when plundering more than that could fit in my cargo-hold, but I understood from comments on #4100 that this type of dump-transferring was considered to be making plundering too powerful.

I changed the code to still allow players to dump cargo to-be-picked up by their escorts.

Amazinite:
If plundering took time and if transferring cargo between ships were to take time then I could see us implementing proper cargo transferring and causing player dumped cargo to not be able to be picked up by escorts. But so long as dumping cargo is our only option for transferring outfits, even if it does make plundering more powerful, I don’t think it’s something we should remove the ability to do.

source/Engine.cpp
```diff
@@ -2046,7 +2046,8 @@ void Engine::DoCollection(Flotsam &flotsam)
 	for(Body *body : shipCollisions.Circle(flotsam.Position(), 5.))
 	{
 		Ship *ship = reinterpret_cast<Ship *>(body);
-		if(!ship->CannotAct() && ship != flotsam.Source() && ship->Cargo().Free() >= flotsam.UnitSize())
+		if(!ship->CannotAct() && ship != flotsam.Source() && ship->Cargo().Free() >= flotsam.UnitSize() &&
+			ship->GetGovernment() != flotsam.SourceGovernment())
```
Amazinite, line 2050: Might be considered a nitpick, but I'd shuffle the order of operations here to check if the governments don't match before checking if you have enough free space. (`can I act AND is this not mine AND is this not an ally's AND do I have space` instead of `can I act AND is this not mine AND do I have space AND is this not an ally's`.)
```suggestion
		if(!ship->CannotAct() && ship != flotsam.Source() && ship->GetGovernment() != flotsam.SourceGovernment()
			&& ship->Cargo().Free() >= flotsam.UnitSize()
```

source/Ship.h
```diff
@@ -366,8 +366,8 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	CargoHold &Cargo();
 	const CargoHold &Cargo() const;
 	// Display box effects from jettisoning this much cargo.
-	void Jettison(const std::string &commodity, int tons);
-	void Jettison(const Outfit *outfit, int count);
+	void Jettison(const std::string &commodity, int tons, bool wasDumped = false);
+	void Jettison(const Outfit *outfit, int count, bool wasDumped = false);
```
Amazinite, line 370: Given that wasDumped is technically inaccurate now (as the player can dump cargo but not have the flotsam's government be set), should this variable be renamed? Would something like saveGov or recordGov be more appropriate?

Amazinite (changes requested):

source/Ship.h
petervdmeer: renamed to wasAppeasing to reflect the intention of the ship that was dumping the cargo.

Amazinite (approved):

PR #6098, opened 2021-07-02
feat: Unify weapons and hazards Ship::TakeDamage() function

petervdmeer:
**Feature:** This PR implements the second part of the feature request detailed and discussed in issue #5751

## Feature Details
This PR makes the `"target effect"` as discussed in #5751 also available for weapons (it was previously only available for hazards).
This PR also unifies the `Ship::TakeDamage()` function to a single function, but hazards don't make use of projectiles, so projectile-bound effects will still not be available for hazards.
I didn't include any provisions in this PR for the effect emitters [as discussed in 5751](https://github.com/endless-sky/endless-sky/issues/5751#issuecomment-787093802). This PR only removes some differences between hazards and weapons.

## UI Screenshots
N/A

## Usage Examples
The `"target effect"` keyword should now also work for regular weapons (giving the same effect as for hazards).

## Testing Done
Added "target effect" "nano sparks" to the Energy Blaster, installed some Energy Blasters on my ship and noticed that the effect was indeed visible when an escort got hit by the weapon.
Flown to Sagittarius A and checked that the hazard is still damaging and giving nano sparks.
Fired at some merchant in human space to verify that Merchants and Syndicate both would be provoked.

## Performance Impact
Expected to be minimal, most TakeDamage() code was already shared between weapons and hazards. Only a few additional checks are now shared.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1986,14 +1986,16 @@ void Engine::DoCollisions(Projectile &projectile)
 				if(isSafe && projectile.Target() != ship && !gov->IsEnemy(ship->GetGovernment()))
 					continue;
 				
-				int eventType = ship->TakeDamage(visuals, projectile, ship != hit.get());
+				int eventType = ship->TakeDamage(visuals, projectile.GetWeapon(), 1.,
+					projectile.DistanceTraveled(), projectile.Position(), projectile.GetGovernment(), ship != hit.get());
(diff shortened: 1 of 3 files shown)


petervdmeer:
Converted to Draft (because some more testing is needed).

petervdmeer:
Ready for review; I've done relevant testing.

PR #6100, opened 2021-07-03
Swizzle the Emerald Sword (And Black Diamond)

kestrel1110:
## Summary
This pull request adds an emerald swizzle to the Emerald Sword. When I play, after getting the Emerald Sword through the mission string, I am always disappointed at the colors changing to my current swizzle. I often simply go into the save file and change it, but it becomes a pain to constantly have to change the pilot file. This pull request makes it so that the ship will permanently have the original Sheragi swizzle.

## Save File
This save file can be used to easily view the change:
[emeraldsword swizzletest.txt](https://github.com/endless-sky/endless-sky/files/6758993/emeraldsword.swizzletest.txt)

## Screenshot
Swizzled Emerald Sword   Black Diamond

## PR Checklist
 - [N/A] I uploaded the necessary image, blend, and texture assets here: {{insert link to assets}}
 - [N/A] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: {{insert PR link}}

### DIFF
--- a/credits.txt
+++ b/credits.txt
@@ -198,7 +198,8 @@ contributed to Endless Sky:
   jozef-mitro
   Jugosloven1612
   Karirawri
-  kestrel1110 (Bereskatuket)
+  kestrel1110
+    Bereskatuket
   KevinKorzekwa
   kilobyte
   kkuchta
--- a/credits.txt
+++ b/credits.txt
@@ -198,6 +198,7 @@ contributed to Endless Sky:
   jozef-mitro
   Jugosloven1612
   Karirawri
+  kestrel1110 (Bereskatuket)
   KevinKorzekwa
   kilobyte
   kkuchta
--- a/credits.txt
+++ b/credits.txt
@@ -199,7 +199,6 @@ contributed to Endless Sky:
   Jugosloven1612
   Karirawri
   kestrel1110
-    Bereskatuket
   KevinKorzekwa
   kilobyte
   kkuchta
--- a/data/sheragi/sheragi ships.txt
+++ b/data/sheragi/sheragi ships.txt
@@ -11,6 +11,7 @@
 ship "Emerald Sword"
 	sprite ship/emeraldsword
 	thumbnail "thumbnail/emeraldsword"
+	swizzle 3
 	attributes
 		category "Heavy Warship"
 		"cost" 100000000
@@ -131,6 +132,7 @@ ship "Emerald Sword"
 ship "Emerald Sword" "Emerald Sword (Fixed)"
 	sprite ship/emeraldsword
 	thumbnail "thumbnail/emeraldsword"
+	swizzle 3
 	outfits
 		"Dragonflame Cannon"
 		"Particle Waveform Turret" 6
--- a/data/sheragi/sheragi ships.txt
+++ b/data/sheragi/sheragi ships.txt
@@ -158,6 +158,7 @@ ship "Emerald Sword" "Emerald Sword (Fixed)"
 ship "Black Diamond"
 	sprite "ship/blackdiamond"
 	thumbnail "thumbnail/blackdiamond"
+	swizzle 9
 	attributes
 		category "Fighter"
 		"cost" 1000000


Terin:
I really liked that swizzle on the original Emerald Sword, yeah. I think @Karirawri mentioned it was meant to be a sort of rusted/faded color from what it used to have, but either way, I'd really like it to have some distinct color, yeah. We've got more options to choose from as well, so it might be worth scanning through those.

Amazinite:
Should this not also apply to the fighters?

credits.txt
```diff
@@ -198,6 +198,7 @@ contributed to Endless Sky:
   jozef-mitro
   Jugosloven1612
   Karirawri
+  kestrel1110 (Bereskatuket)
```
Amazinite, line 201: All the names under this part of the credits are GitHub `@`s. It seems like it'd be a bit cluttered if we started allowing people to put multiple names, especially because it could easily lead to the credits overflowing the UI.

kestrel1110: Yeah I think I'll do that, it's just that I've always gone by Bereskatuket at my main name so it seemed appropriate.
But I do recognize that there should be more organization, so I'll do this.

kestrel1110:
> Should this not also apply to the fighters?

@Amazinite I was actually waiting for your pull request that added many more swizzles, #5974 to get merged, as I would give them swizzle 9, which makes them black.
I just noticed that it is merged, so I'll be updating the pr with that.

Karirawri:
The current swizzle was indeed supposed to be a "faded" version of the original green that it would have, but if we go forward with that, I would recommend giving the player an option to either keep the paint as it is or repainting it with their faction's color, since there's nothing preventing it from being repainted anyway, and not everyone likes that shade of green.

Amazinite:
If a choice were to be made, then it would need to be made either in the boarding conversation of "Sheragi Archaeology: The Emerald Sword 3" or the offer conversation of "Sheragi Archaeology: Epilogue". If done in the former then we'd need two missions separate from "Sheragi Archaeology: Epilogue" to gift the ships to the player when that mission offers. If done in the latter then the missions would gift the ships to the player when "Sheragi Archaeology: Epilogue" completes, in which case the Emerald Sword would still need to be an NPC ships for "Sheragi Archaeology: Epilogue" that you escort to the final destination.

In either case, we'd need to do some rewriting to fit in the question of "so what color do you want it in?" The problem that I see is that I think that topic is the most appropriate for the on complete conversation of "Sheragi Archaeology: Epilogue", not either of the two conversations where it'd need to go *if* we still wanted to give the ship to the player after they completed "Sheragi Archaeology: The Emerald Sword 3" or "Sheragi Archaeology: Epilogue". So if we want the player to be able to choose which paint job they get while still having everything flow nicely, we'd need to add an additional mission after the epilogue. Perhaps Foster keeps the Emerald Sword for study for a little bit longer, then tells you when it's ready to pick up. We don't need the wait period to be long, but I don't think we want it super short either. Perhaps 30 days is a reasonable time.

Zitchas:
This whole topic is, to me, a reminder that we need some way to *partially* change a ship's color without changing *all* of it.

As the text of the FW mission goes, while you are soldier for a government, there is a whole host of reasons why your ships *must* be marked such to identify you as an official soldier/military member of said government. No amount of "but this ship is called X" or "It'd look cooler if it stayed this other color" changes any of those reasons.

On the other hand (and this is true for *all* ships, not just the Emerald Sword), there is no reason why the entire ship needs to be painted. The simple addition of a green stripe down one side (or even better, a green logo) or some other similar pattern would be more than enough to identify the ships while allowing them to maintain the majority of their coloration as normal.

60second GIMPing to illustrate the concept:
emeraldsword (green stripe)
hai shield beetle(green stripe)

I'm sure our artists can work out something that looks a bit better than that. A darker green and a higher transparency would probably look better. (this is done using "darken only" on GIMP, with transparancy at 70
%). And especially using something that combines with the shadows to keep the shadowed parts much darker.  But the point is to have a graphical addition that is layered *on top* of whatever swizle is present. So swizzle does its thing, then the stripe/pattern is put on overtop. So no matter what the swizzle is, the ship has a colored marker of the appropriate government type.

We could even let players import their own patterns and pick their own colors. Yes, I know this adds graphical workload on top of everything, but even it halves the numbers of ships we can have in flight, having average computers start to lag at 500 ships on screen instead of at 1000 ships on screen is perfectly reasonable. Especially for a game that is balanced around solo-ship flight and *not* having a large fleet. (and even carriers that *should* have lots of ships... don't.)

Issue #773, opened 2016-02-09
Suppress 'your ship is moving erratically' message from escorts?


brendanjones:
Annoyingly, I can't recreate this. At least I have a screenshot. :) 

For a while when I first bought my Bactrian, I was getting 'your ship is moving erratically because...' warnings, even though I had +200 crew. My movement was not affected.

Screenshot for proof:

![screen shot 2016-02-06 at 5 49 43 pm](https://cloud.githubusercontent.com/assets/16049594/12912921/f5c38486-cf1b-11e5-9730-29e68f86fdf8.png)

You can see the faint 'your ship is moving erratically' text at the bottom of the pic

natangold:
Weird. You can get the message if your escorts are understaffed (and those escorts will fly wonky), but your Bactrian is full up, so you must've landed, which should have staffed all of your escorts. Unless your escorts were either out of system or warping in at the moment you landed, in which case, they'd continue as they were.

endless-sky:
Could one of those escorts in a different system be one that you captured with not enough crew to pilot it?

brendanjones:
No idea. How does an escort not have enough crew to pilot it? Is it when you capture a ship and your own crew numbers are so low that the crew has to be split across the two ships - both with not enough crew? I don't think I would have done that (because it annoys the crap out of me when you can't control your ship. It's faster in real time to land with just enough crew, refill your crew, then take off and start killing again), but I can't rule it out.

jmhorjus:
I've had this as well - I believe it turned out to be a disabled escort in another system that didn't have enough crew and was spamming the message all the time.  

Perhaps this should only be displayed when your flagship is short on crew? Escorts usually manage their own crew just fine whenever you get them onto a planet.  An abandoned escort in another system shouldn't cause this message.

jmhorjus:
Confirmed:  I can reproduce this with the above mentioned procedure: 
1.) Capture a ship causing both to be short on crew.
2.) Let that ship be disabled. 
3.) Land.   

The messages will continue until the disabled escort is rescued or parked (which you can still do, even though it's disabled).  
Interestingly, after you've landed the disabled escort actually appears to gain enough crew to run, but does not stop spamming the "not enough crew" message.  

Clear solution is to restrict this message to only trigger on the players flagship.

endless-sky:
[...]

Issue #4239, opened 2019-03-30
Main Flight Interface & HUD ease of comprehension Changes


Zitchas:
**Is your feature request related to a problem? Please describe.**
The current in-flight user interface relies heavily on a combination of text and sound, where the graphical elements are limited to targeting and scaling variables such as energy, heat, shields, etc.

Unfortunately, many people have trouble keeping track of all these elements, especially when required to read large amounts of text to keep track of what is going on. 

This is particularly relevant for people who:
 - Are not native English speakers.
 - Have conditions (such as dyslexia) that affect reading ability.
 - Have trouble multi-tasking.
 - For some reason cannot hear the sound.
 - For some reason have to play the game without sound.

**Related Issue Links**
 #4240

**Describe the solution you'd like**
1) The amount of text that scrolls through the message list in the bottom left corner should be reduced significantly.
2) Commonly repeated things that don't have useful differences in data should be converted to an icon or other GUI element that are easily identifiable and longer lasting.
3) User-controllable fade settings controlling how long messages last before fading. 
4) The scrolling text should primarily be for messages that are unique, contain variable specific information, and/or low-priority. For instance, ship hails, crew salaries, interest, and fines all fit into this category.

Zitchas:
#4240

comnom:
>The amount of text that scrolls through the message list in the bottom left corner should be reduced significantly.

There's some low hanging fruit here, specifically in regards to player initiated combat actions.

* Cloaking
User initiated, and has a visual cue.

* Launching fighters
User initiated, and has a visual cue.

* Player scanning
User initiated, and has a visual cue.

* NPC scanning
Has it's own issue.

These are all simple, repeatable actions that don't need a message in my book.

The only other I see is dumping cargo, but that logs finacial info in the message. It's a bit out of place even taking that into account, because the action occurs in the ship panel but the message is displayed in the main panel.

Zitchas:
Those are all good candidates for cleaning up. 

In terms of fighters, I think keeping the notification for recalling fighters, since that acknowledgement when they may (probably likely) be off screen is useful. But definitely don't need the launching one.

petervdmeer:
I would also like to be able to lookup what the ships in my fleet are doing.

If we could get icons for holding position, gathering, escorting and a few others, then we would not need to provide feedback for such actions in the text field (and the information would also be available when the text-message is gone).

Here an example of icons giving status information:
[...]

PR #6101, opened 2021-07-04
Don't spam "does not have enough crew" message until it clears the queue.

ravenshining:
**Feature:** This PR reduces message log spam from ships with not enough crew.

## Feature Details

The `<ship> is moving erratically because you do not have enough crew to pilot it.` messages have been marked as unimportant, so they will not redisplay at the bottom of the screen until they are cleared from the message queue - that is, after about 17 seconds, or sooner if there are many messages from other sources coming in.

~~In practise, a message from a severely shorthanded vessel will still be shown at almost all times, but~~  <- This is what I expected, but ES is bit lazy about cleaning the queue.  Rather, if you are in a quiet area of space you might not see any messages at all until a different messages is triggered, then you'll get the crew message(s).

As expected, however, it will not compete with other messages for display in the queue, and it will scroll up in an orderly fashion, rather than having ships chaotically competing for the front position.

**Reasoning**
In the current game, when one captures a ship and is either severely shorthanded as a result or there are many captured ships, the message log quickly becomes flooded with complaints that there are not enough crew - up to one per ship every half second - drowning out all other messages.  While perhaps realistic - there may not be enough crew members to handle communications - this is extremely annoying and detracts from gameplay, as the player may easily miss more important messages.

**Related Issue Links**
Issue #773 suggested removing this message from escort ships entirely, but MZ said it was worth having the message show.  Rather than eliminating the message entirely, this PR simply reduces its frequency.

Thinking along 773's same lines, and retaining a nod to realism, I could instead only mark escort ships' complaints as unimportant.  This would still disrupt your communications when your flagship is short-handed, but at least escorts would not do this to you.  My preference however is to mark all such messages as unimportant as this PR does currently.

I've also tried and tested adding a cooldown to the messages that is only as long as it takes for a message to start fading (about 10 seconds, which is shorter than the time it takes for a message to be erased from the queue).  Although it worked, that method required adding more lines of code; this PR's method is much simpler.

## Testing Done
Capturing a ship and then basking in the newfound readability of the message log.

## Performance Impact
If anything I'd expect this to be mildly positive.

### DIFF
--- a/changelog
+++ b/changelog
@@ -48,6 +48,7 @@ Version 0.9.14:
   * User interface:
     * The "interrupt fast-forward" setting now defaults to off instead of on. (@Terin)
     * Fixed a misaligned UI element in the map panel. (@Terin)
+    * Messages about low crew ships are no longer considered "important," meaning they will no longer be pushed to the bottom of the message list every time they trigger and drown out other messages. (@ravenshining)
   * Under the hood:
     * Combined checking if a stellar object has a planet and if that planet is valid into a single method. (@tehhowch)
     * Having multiple missing frames for a sprite are now logged to the error file on a single line instead of generating a new line for each missing frame. (@ashdnazg, @tehhowch)
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1700,9 +1700,9 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 	{
 		pilotError = 30;
 		if(parent.lock() || !isYours)
-			Messages::Add("The " + name + " is moving erratically because there are not enough crew to pilot it.",false);
+			Messages::Add("The " + name + " is moving erratically because there are not enough crew to pilot it.", false);
 		else
-			Messages::Add("Your ship is moving erratically because you do not have enough crew to pilot it.",false);
+			Messages::Add("Your ship is moving erratically because you do not have enough crew to pilot it.", false);
 	}
 	else
 		pilotOkay = 30;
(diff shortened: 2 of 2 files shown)


Amazinite:
#4239 seems like a relevant issue relating to this PR, especially the point on converting messages to HUD elements. In the future a low crew ship might have a special icon next to it on your HUD showing that it is low crew (perhaps flashing when the low crew is causing erratic movement). Your flagship could similarly have some sort of special HUD marker showing the same thing. Perhaps we'd still display a message, say the first time after each departure that any player ship gains low crew we provide the message to make it easily known what the HUD element is portraying, but we wouldn't display a new message every time the ship begins moving erratically.

As far as this issue goes as a step toward clearing up the in-flight messages though, seems like a fair enough change to me. In my experience, if my flagship is low on crew then probably so is the ship I just captured, meaning that I currently get two messages basically permanently stuck on my screen telling me that my ships are moving erratically. I understand the reasoning behind why the hails are currently considered "important" (i.e. they keep refreshing at the bottom of the messages whenever triggered), that being that the message is suppose to coincide with the exact frame where a ship starts moving erratically, but given that low crew ships are likely to trigger the erratic movement, especially very low crew ones, continually pushing a new message every time the ship's pilot error triggers is definitely overkill. I think it's fair to assume that if I see the message once alongside my ship moving erratically, then shortly afterward my ship starts moving erratically again, I don't need the message to be dragged to the bottom of the list again.

ravenshining:
Oooh, yeah, I love the idea of a HUD icon, status bar, or tinting, rather than the intermittent messaging.

PR #6106, opened 2021-07-08
Allow the map to be opened when at the end of a conversation

Amazinite:
**Bugfix:** This PR addresses an issue with attempting to open the map while at the end of a conversation.

## Fix Details
The map is typically capable of being opened by the player while in the middle of a conversation, except in two scenarios:
1. The conversation is currently at a name entry field.
2. The conversation is at its end.

Judging by a comment in ConversationPanel.cpp, only the former scenario is intended to block the player from opening the map. As such, this PR changes it so that the map is allowed to be opened while at the end of a conversation.

## Testing Done
Attempted to open the map at the end of a conversation with and without this change. With this change, the map is able to be opened, while without it the map can't be opened.

Tested that the behavior of pressing `m` while at a name entry field is unchanged.

## Save File
N/A

### DIFF
--- a/source/ConversationPanel.cpp
+++ b/source/ConversationPanel.cpp
@@ -202,8 +202,9 @@ void ConversationPanel::Draw()
 bool ConversationPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, bool isNewPress)
 {
 	// Map popup happens when you press the map key, unless the name text entry
-	// fields are currently active.
-	if(command.Has(Command::MAP) && !choices.empty())
+	// fields are currently active. The name text entry fields are active if
+	// choices is empty and we aren't at the end of the conversation.
+	if(command.Has(Command::MAP) && (!choices.empty() || node < 0))
 		GetUI()->Push(new MapDetailPanel(player, system));
 	if(node < 0)
 	{


source/ConversationPanel.cpp
```diff
@@ -202,8 +202,9 @@ void ConversationPanel::Draw()
 bool ConversationPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, bool isNewPress)
 {
 	// Map popup happens when you press the map key, unless the name text entry
-	// fields are currently active.
-	if(command.Has(Command::MAP) && !choices.empty())
+	// fields are currently active. The name text entry fields are active if
+	// choices is empty and we aren't at the end of the conversation.
+	if(command.Has(Command::MAP) && (!choices.empty() || node < 0))
```
petervdmeer, line 207: Is it possible to map the Command::MAP to SDLK_RETURN or SDLK_KP_ENTER or to the 'd' key? (and then have the boardingPannel from the Exit-function called on line 214) to open on top of the map?

Amazinite: Not quite sure what you mean by that. Are you saying that if the map is opened at the end of the conversation, it should also exit out of the conversation entirely?

petervdmeer: What I meant is what happens if the player changes his/her preferences to use one of the keys from line 212 as a key to open the map. I was wondering if that could lead to a situation where the MapDetailPanel() gets pushed here on line 208 and the Exit() function on line 214 also doing something after that like pushing a BoardingPanel())

Amazinite: Not sure, but if there are issues with that then there were issues before as well.

petervdmeer: Agreed, that would not be a new issue, so we can address that separately.

petervdmeer (approved):

PR #6107, opened 2021-07-08
Don't leak the destination of invisible missions from the special marker

Amazinite:
**Bugfix:** This PR addresses issue with the destination of invisible missions being able to be known by opening the map during an invisible mission's on offer conversation.

## Fix Details
It is currently permissible for invisible missions to have on offer conversations. During an on offer conversation, the player is capable of opening the map and viewing a special marker placed at the mission's destination. This marker is created regardless of the visibility of the mission. Invisible missions otherwise don't create map markers of any kind, and the special map marker should be no different.

This PR makes it so that invisible missions do not pass the destination system to any MissionActions, which normally pass the destination to any created Dialogs or ConversationPanels, which only use the destination system otherwise for passing to a MapDetailPanel that displays the special marker, meaning special markers will no longer be created for invisible missions.

## Testing Done
Created an invisible mission with an on offer conversation and opened the map during the conversation with and without this change. With this change, no special marker is created. Without this change, a special marker is created at the mission's destination.

## Save File
N/A

### DIFF
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -950,6 +950,12 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 	// If this trigger has actions tied to it, perform them. Otherwise, check
 	// if this is a non-job mission that just got offered and if so,
 	// automatically accept it.
+	// Actions that are performed only receive the mission destination
+	// system if the mission is visible. This is because the purpose of
+	// a MissionAction being given the destination system is for drawing
+	// a special marker at the destination if the map is opened during any
+	// mission dialog or conversation. Invisible missions don't show this
+	// marker.
 	if(it != actions.end())
 		it->second.Do(player, ui, (destination && isVisible) ? destination->GetSystem() : nullptr, boardingShip, IsUnique());
 	else if(trigger == OFFER && location != JOB)
--- a/source/Mission.cpp
+++ b/source/Mission.cpp
@@ -951,7 +951,7 @@ bool Mission::Do(Trigger trigger, PlayerInfo &player, UI *ui, const shared_ptr<S
 	// if this is a non-job mission that just got offered and if so,
 	// automatically accept it.
 	if(it != actions.end())
-		it->second.Do(player, ui, destination ? destination->GetSystem() : nullptr, boardingShip, IsUnique());
+		it->second.Do(player, ui, (destination && isVisible) ? destination->GetSystem() : nullptr, boardingShip, IsUnique());
 	else if(trigger == OFFER && location != JOB)
 		player.MissionCallback(Conversation::ACCEPT);
 	


Amazinite:
I'll also note that although it's technically a bit of a misuse of `invisible`, akin to #5930, using `invisible` on missions that never actually accept is a common content creator "error," and I believe the expectation is that missions that never actually accept but also are marked invisible won't display this special marker.

petervdmeer:
> This PR makes it so that invisible missions do not pass the destination system to any MissionActions, which normally pass the destination to any created Dialogs or ConversationPanels, which only use the destination system otherwise for passing to a MapDetailPanel that displays the special marker, meaning special markers will no longer be created for invisible missions.

Is it also not allowed to use the destination in Dialogs? (I feel that that should be permitted, because the content creator has an explicit choice to use the destination or not use the destination in a Dialog).

Is it possible to do the check on drawing of the special marker later (for example very close to the code that actually draws the marker)?

Amazinite:
> Is it also not allowed to use the destination in Dialogs? (I feel that that should be permitted, because the content creator has an explicit choice to use the destination or not use the destination in a Dialog).

You're thinking of text replacements, right? This are done during instantiation and have nothing to do with the passing of the destination system to a mission action through the Do function. MissionAction::Do gets the destination system only for the purposes of passing it on to the MapDetailsPanel that creates the special marker.

> Is it possible to do the check on drawing of the special marker later (for example very close to the code that actually draws the marker)?

As described, there's really no need since MissionAction::Do only gets the destination for the purposes of passing it along until the special marker is drawn anyway. Having the check be done close to the drawing of the special marker is originally what I looked into doing, but it involves passing the state of a Mission's isVisible bool all the way to MapDetailsPanel, which seemed too complex when we could just never pass the destination to the MissionAction to begin with.

petervdmeer:
I was indeed thinking of the text replacements, and you are right, those are indeed not affected since they are initialized earlier.

Still, the location of the check in this PR feels not intuitive. That doesn't matter much today, but if we need to change something in the code-path around this code in a few months or a few years from now, then this might make things more difficult by then.

Amazinite:
> Still, the location of the check in this PR feels not intuitive.

Well to me it feels like it'd be more cluttered to propagate the mission's visibility from the Mission to the MissionAction to the Dialog or ConversationPanel and finally to the MapDetailsPanel just to stop the special marker from showing, when all that's really necessary is to cut it off at the source and just use the missions' visibility to determine if the MissionAction even needs the destination to begin with instead of passing along a system that never gets used.

petervdmeer:
I feel the clutter also comes from the software-architecture; the code around MissionActions, Dialogs and ConversationPanels seems to filled with side-effects.
What we could do for now is to go with your suggestion, but with a very clear comment above it. Maybe something like: "Don't send the destination for invisible missions, to avoid those destinations being shown with a special marker when the player switches from a ConversationPanel to a MapPanel."

petervdmeer:
Such comment will help us later if we do decide to cleanup the code around here.

Amazinite:
@petervdmeer Is that good?

PR #6110, opened 2021-07-09
feat(UI): Provide more detail/terminology for combat rating

Terin:
## Feature Details
This PR expands the combat rating section in the Player Info Panel. Since the start, combat rating on the game's end is a number, usually in the thousands when a player has fought enough, which is tracked as a condition for various missions or tribute. This is surfaced to the player, also as "Combat Rating", but as the log of the internal number. This results in a drastically different value than what is expected for missions, leading to lots of confusion when a guide says combat rating of 15 is required, and they go "Oh no, but I only have 3!", when they meet the requirement just fine.

With help from Derpy, this breaks down the terminology to the player more, and surfaces the internal tracking number:

- Combat Rank: The currently surfaced Combat Rating
- Combat Title: The string associated with the current Combat Rating
- Combat XP: The internal combat rating number
- XP required for next rank: Shows the next log rollover, for how much it'll take for the next rank. This becomes `MAX` once the maximum combat rating, as defined in data, is reached.

This should make it much clearer to people asking for help/reading guides, where it can be said "You'll need 15 combat XP, or be at about combat rank 3." It also lets someone who's interested in chasing a higher combat rating for bragging rights to have a better idea of where they're at.

Ideally, I'd prefer if we could replace `combat rating` in the data with `combat xp`, and maybe with an additional `combat rank` as a shorthand, but that would break plenty of data and plugins. Maybe combat rating could be kept for backwards compatibility, and combat XP/rank would be used going forward, but that's a separate issue/PR.

## UI Screenshots
image

## Testing Done
Seeing the combat rating properly roll over when a new rank is reached, and making sure MAX is displayed for max combat rating, depending on what's used in data.

## Performance Impact
N/A

### DIFF
--- a/source/PlayerInfoPanel.cpp
+++ b/source/PlayerInfoPanel.cpp
@@ -507,9 +507,8 @@ void PlayerInfoPanel::DrawPlayer(const Rectangle &bounds)
 		table.DrawGap(5);
 		
 		table.DrawTruncatedPair("rank:", dim,
-			to_string(combatLevel), dim, Truncate::MIDDLE, false);
-		table.DrawTruncatedPair("title:", dim,
-			combatRating, dim, Truncate::MIDDLE, false);
+			"(" + to_string(combatLevel) + ") - " + combatRating,
+			dim, Truncate::MIDDLE, false);
 		table.DrawTruncatedPair("experience:", dim,
 			Format::Number(combatExperience), dim, Truncate::MIDDLE, false);
 		bool maxRank = (combatRating == GameData::Rating("combat", combatLevel + 1));
--- a/source/PlayerInfoPanel.cpp
+++ b/source/PlayerInfoPanel.cpp
@@ -512,12 +512,10 @@ void PlayerInfoPanel::DrawPlayer(const Rectangle &bounds)
 			combatRating, dim, Truncate::MIDDLE, false);
 		table.DrawTruncatedPair("experience:", dim,
 			Format::Number(combatExperience), dim, Truncate::MIDDLE, false);
-		if(combatRating == GameData::Rating("combat", combatLevel + 1))
-			table.DrawTruncatedPair("    for next rank:", dim,
-				"MAX", dim, Truncate::MIDDLE, false);
-		else
-			table.DrawTruncatedPair("    for next rank:", dim,
-				Format::Number(ceil(exp(combatLevel + 1))), dim, Truncate::MIDDLE, false);
+		bool maxRank = (combatRating == GameData::Rating("combat", combatLevel + 1));
+		table.DrawTruncatedPair("    for next rank:", dim,
+				maxRank ? "MAX" : Format::Number(ceil(exp(combatLevel + 1))),
+				dim, Truncate::MIDDLE, false);
 	}
 	
 	// Display the factors affecting piracy targeting the player.
--- a/source/PlayerInfoPanel.cpp
+++ b/source/PlayerInfoPanel.cpp
@@ -495,7 +495,8 @@ void PlayerInfoPanel::DrawPlayer(const Rectangle &bounds)
 		bright, Truncate::MIDDLE, true);
 	
 	// Determine the player's combat rating.
-	int combatLevel = log(max<int64_t>(1, player.GetCondition("combat rating")));
+	int combatExperience = player.GetCondition("combat rating");
+	int combatLevel = log(max<int64_t>(1, combatExperience));
 	const string &combatRating = GameData::Rating("combat", combatLevel);
 	if(!combatRating.empty())
 	{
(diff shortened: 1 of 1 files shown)


source/PlayerInfoPanel.cpp
```diff
@@ -505,8 +506,18 @@ void PlayerInfoPanel::DrawPlayer(const Rectangle &bounds)
...
+		table.DrawTruncatedPair("title:", dim,
+			combatRating, dim, Truncate::MIDDLE, false);
+		table.DrawTruncatedPair("experience:", dim,
+			Format::Number(combatExperience), dim, Truncate::MIDDLE, false);
+		if(combatRating == GameData::Rating("combat", combatLevel + 1))
+			table.DrawTruncatedPair("    for next rank:", dim,
+				"MAX", dim, Truncate::MIDDLE, false);
+		else
+			table.DrawTruncatedPair("    for next rank:", dim,
+				Format::Number(ceil(exp(combatLevel + 1))), dim, Truncate::MIDDLE, false);
```
Amazinite, line 520: Seems like it'd be better to trim this down to the following:
```suggestion
		bool maxRank = (combatRating == GameData::Rating("combat", combatLevel + 1));
		table.DrawTruncatedPair("    for next rank:", dim,
				maxRank ? "MAX" : Format::Number(ceil(exp(combatLevel + 1))),
				dim, Truncate::MIDDLE, false);
```

Amazinite (changes requested):

source/PlayerInfoPanel.cpp
Terin: Some day I'll internalize ternary operators.

Amazinite (approved):

source/PlayerInfoPanel.cpp
```diff
@@ -505,8 +506,16 @@ void PlayerInfoPanel::DrawPlayer(const Rectangle &bounds)
...
 		
-		table.DrawTruncatedPair(combatRating, dim,
-			"(" + to_string(combatLevel) + ")", dim, Truncate::MIDDLE, false);
+		table.DrawTruncatedPair("rank:", dim,
+			to_string(combatLevel), dim, Truncate::MIDDLE, false);
+		table.DrawTruncatedPair("title:", dim,
+			combatRating, dim, Truncate::MIDDLE, false);
+		table.DrawTruncatedPair("experience:", dim,
+			Format::Number(combatExperience), dim, Truncate::MIDDLE, false);
+		bool maxRank = (combatRating == GameData::Rating("combat", combatLevel + 1));
```
Amazinite, line 515: I will say that it's technically possible for us to have two ranks in a row with the same title, which under this line would cause the experience to the next rank to display as MAX, so a better check for max rank might be to see if `experience >= 2000000000`, which is the max rating as defined by PlayerInfo::HandleEvent.

petervdmeer:
The information displayed is technically correct, but I'm not too happy with displaying all 4 pieces of information (rank, title, experience and next level) to players of ES. Most of the information feels like too much internal detail for our regular, non-expert ES players.

Having rank and experience both listed on the UI feels a bit redundant.

Amazinite:
Would deleting "rank" and replacing "for next rank" with "for next title" help? We'd be completely removing the numbered "rank" from the UI with this change then, which might not be a bad thing. The "experience" is the more important part in terms of getting missions to offer anyway, and the rank is always tied to the title, so instead of saying "I'm at combat rating X," people would then just say "My current xp is X" or "My current title is X."

source/PlayerInfoPanel.cpp
Terin: I set it up that way in case someone modifies the list of titles to have less/more (if possible) titles than vanilla. If we changed it to that number, and someone had a list with less titles, the game would keep going "New title coming up!" and not give anything.

Both are non-vanilla modifications, and especially if we're removing rank #, I'd just want to base it on someone who changed the number of titles, rather someone who theoretically puts two identical titles in a row before getting new ones.

Terin:
I don't think that's too much information. It's simply a leveling system, except we provide a name for each level. It's a useful shorthand for say, the Kestrel mission, to say you need rank 9 rather than 8193+ combat XP.

But if I did have to choose that or the old system, I'd prefer looking up the "force to be reckoned with" title than the confusion we have with combat rating right now.

petervdmeer:
A problem with just displaying a title (without any numbers) would be that a player might not know if a title like "force to be reckoned with" is above or below a required title "respected enemy", so we probably would want to still display one of the numbers.

A concern with removing "rank" is that we are basically changing the code to accomodate for guides that were using the wrong number (guides should have used rank, not the numbers they could find in the data-files). Other guides that were using the correct number (the rank) would now need to be updated.

For progress towards the next level we could also consider to show a percentage (this will probably work better when we get to the really large numbers). And I wonder if we should move some of this information to the "hover over text" because there is also a lot of useful other information that we also want to show on this screen.

Terin:
>A concern with removing "rank" is that we are basically changing the code to accomodate for guides that were using the wrong number (guides should have used rank, not the numbers they could find in the data-files). Other guides that were using the correct number (the rank) would now need to be updated.

The reason that's been such a weird point of contention is some missions using arbitrary combat values like 50 or 200, while others use one that correspond with rank cutoffs (like the Kestrel, top-tier Marauders, or the bounty hunters). That inconsistency is why I want to provide both pieces of information in game, so that in case someone creating a guide going through mission data doesn't know/want to pop the numbers into ln(), someone else reading that guide isn't completely lost.

>For progress towards the next level we could also consider to show a percentage (this will probably work better when we get to the really large numbers).

A percentage would be a bit weird with how Endless Sky works, since we're looking at total XP, not really level. If I came back to a save I hadn't been at in a while, and it says I'm 80% of the way to the next level, would that mean my total XP is 80% of the way to that new level? Or the XP since the last level is 80% to that? Implementation-wise, we'd pick the latter, but that still might not be clear to other players.

>And I wonder if we should move some of this information to the "hover over text" because there is also a lot of useful other information that we also want to show on this screen.

We could merge rank and title into a single line, sort of like the original, could be used to save space, though. (with the order/formatting being up for change)

image

All the other categories are able to collapse neatly already though. Unless there's plans for more PRs that fit information there, having all the licenses on display doesn't need too much space to fill.

Amazinite:
That looks good to me.

petervdmeer:
Looks good to me as well.

PR #6112, opened 2021-07-10
Fix licensing for the Large Cogeneration Module.

Arachi-Lover:
----------------------
**Fix (Outfit Licensing)**

## Summary
The Large Cogeneration Module's license definition wasn't in the right format, this just sets it so that it'll require the license to purchase it.

Issue brought up by Amacita on the discord server.


tehhowch:
As of 251fa5a2b758970679a4496da01eebbae0451870 this syntax is now supported.

PR #6113, opened 2021-07-11
Draw any gun hardpoints under the ship (by default)

Amazinite:
**Feature:** This PR implements a feature described in #6062.

## Feature Details
This PR makes it so that any gun hardpoint with a hardpoint sprite will now have the sprite rendered under the ship (by default). Previously, all hardpoint sprites were rendered on top of the ship. While this makes sense for turret hardpoints, which hardpoint sprites were designed for, this doesn't make as much sense if you wanted a gun hardpoint sprite, which is not disallowed from being done.

~~In the future it may be that we allow any hardpoint sprite to arbitrarily be drawn above or below the ship regardless of hardpoint type, but I chose to just draw all gun hardpoint sprites under the ship for the sake of simplicity; all we need right now are gun hardpoints drawn under ships, and allowing turret hardpoints under the ship would introduce issues that would be rather complex to resolve.~~ I've gone ahead and allowed any gun/turret hardpoint to be defined as being drawn under or over, as we need to keep the ability for over drawn guns for the Shooting Star person ship, while under drawn guns are needed for #6062/#6114. While this includes the ability to create under drawn turrets, I still think that actually doing so has issues that need ironed out. I don't plan on tackling such issues in this PR though, so until we tackle them in the future, under drawn turrets shouldn't be in vanilla content.

One issue is that a turret hardpoint drawn under the ship could potentially rotate in such a way that the hardpoint is hidden, while the projectile for the turret would still be drawn on top of the ship, giving the appearance of projectiles coming from nowhere. Vital attempted to resolve this issue by creating two projectile draw layers, one drawn below all ships and one drawn above all ships. The main issue I have with this solution though is that it would allow projectiles to be hidden by ships. That is, an enemy ship could potentially fire at you with a projectile drawn below all ships, only for that projectile to be hidden by other ships friendly to it and suddenly come out from under another ship and hit you with little warning. An alternative solution that was discussed was to clip the projectile sprite from the origin up until it is no longer within the collision mask of the firing ship. This though has issues with ships with holes in them (as holes are not currently cut out of collision masks, so a projectile would be clipped despite being in a region that should make it visible) and with ship geometry where a projectile sprite leaves the collision mask of a ship but then reenters it. (The following image shows an example of this, with the red outlines being a ship's collision mask, the blue square being the hardpoint, and the green lines being projectiles. The bottom-right projectile leaves the collision mask but then reenters it; how would such a projectile be clipped?)
image

## Screenshots
The following image shows nuclear missiles with hardpoint sprites. Notice how the nuke hardpoint sprites render under the ships, while the turret sprites are still rendered over the ships.
image

## Usage Examples
Just define a hardpoint sprite for a gun as you would have before and it will automatically be drawn under the ship.
New over/under keywords can also be added to weapon hardpoint definitions, similar to how engine and bay hardpoint definitions work.
```
	# guns are under by default
	gun x y
		over
	# turrets are over by default
	turret x y
		under
```

## Testing Done
See screenshots.

## Performance Impact
N/A

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -2241,8 +2241,8 @@ void Engine::AddSprites(const Ship &ship)
 	if(ship.IsSteering() && !ship.SteeringEnginePoints().empty())
 		DrawFlareSprites(ship, draw[calcTickTock], ship.SteeringEnginePoints(), ship.Attributes().SteeringFlareSprites(), Ship::EnginePoint::UNDER);
 	
-	drawObject(ship);
-	for(const Hardpoint &hardpoint : ship.Weapons())
+	auto drawHardpoint = [&drawObject, &ship](const Hardpoint &hardpoint) -> void
+	{
 		if(hardpoint.GetOutfit() && hardpoint.GetOutfit()->HardpointSprite().HasSprite())
 		{
 			Body body(
@@ -2253,6 +2253,15 @@ void Engine::AddSprites(const Ship &ship)
 				ship.Zoom());
 			drawObject(body);
 		}
+	};
+	
+	for(const Hardpoint &hardpoint : ship.Weapons())
+		if(!hardpoint.IsTurret())
+			drawHardpoint(hardpoint);
+	drawObject(ship);
+	for(const Hardpoint &hardpoint : ship.Weapons())
+		if(hardpoint.IsTurret())
+			drawHardpoint(hardpoint);
 	
 	if(ship.IsThrusting() && !ship.EnginePoints().empty())
 		DrawFlareSprites(ship, draw[calcTickTock], ship.EnginePoints(), ship.Attributes().FlareSprites(), Ship::EnginePoint::OVER);
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -2255,14 +2255,12 @@ void Engine::AddSprites(const Ship &ship)
 		}
 	};
 	
-	// Gun hardpoints are drawn under the ship, while turret hardpoints are drawn
-	// over it.
 	for(const Hardpoint &hardpoint : ship.Weapons())
-		if(!hardpoint.IsTurret())
+		if(hardpoint.IsUnder())
 			drawHardpoint(hardpoint);
 	drawObject(ship);
 	for(const Hardpoint &hardpoint : ship.Weapons())
-		if(hardpoint.IsTurret())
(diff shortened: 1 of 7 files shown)


source/Engine.cpp
```diff
@@ -2253,6 +2253,15 @@ void Engine::AddSprites(const Ship &ship)
 				ship.Zoom());
 			drawObject(body);
 		}
+	};
+	
+	for(const Hardpoint &hardpoint : ship.Weapons())
```
petervdmeer, line 2258: Maybe add a comment here that this draws the hardpoints under the ship (and below a comment that it draws the hardpoints on top of the ship).

petervdmeer (commented):
Code itself looks good, already approving (with a minor comment that can help for future changes).

petervdmeer (approved):

beccabunny:
This breaks the Shooting Star

vitalchip:
As mentioned I created new draw layers so that projectiles, fire effects and live effects coming from 'under' hardpoints are all drawn under ships. Don't agree it's a problem though.

Turrets which stick out from under the ship at certain points on their rotation and vanish beneath the ship at others look great but only when the smoke/flash/beams/bolts match.

petervdmeer:
> As mentioned I created new draw layers so that projectiles, fire effects and live effects coming from 'under' hardpoints are all drawn under ships. Don't agree it's a problem though.

I agree with Derpys assessment in this PRs text that the "under" layer is a problem when the projectile would appear under more than just the firing ship.

> Turrets which stick out from under the ship at certain points on their rotation and vanish beneath the ship at others look great but only when the smoke/flash/beams/bolts match.

Yes, but I feel that the "clipping" solution as mentioned by Derpy is a more correct solution for this (to make sure that projectiles are always drawn on top of other ships).

petervdmeer:
> This breaks the Shooting Star

Sounds like we need to add the "under" and "over" keywords for gunports already (and add the "over" keyword to the gun-port of the shooting star).

vitalchip:
Again, disagree. I like seeing the projectiles and the live effects (missile trails) going both over and under ships, it adds something, for me anyway. Although it's a 2D top down game there is *some* 3D "feeling" and that enhances it, in my view.

That said, I am interested in seeing how the clipping method looks, beams would be the most trouble, I think.

petervdmeer (approved):
Latest update looks good.

PR #6123, opened 2021-07-18
refactor(internal): Force authors to consciously decide the importance level of a Message

salqadri:
Changed the second argument to Messages::Add from an optional bool called isImportant that defaulted to _true_ to being a required enum class with two options, Low, which maps to what was previously 'false', and High, which maps to what the default case was. Whilst the interface now effectively allows the addition of new importance levels in future, the underlying data type in the implementation is still a bool. No changes were made to the behavior of the code; the ONLY difference is that it is now explicit in the code, to force authors to make a more conscious decision on whether their message should be considered High importance or Low. Future work, IF ANY, could entail revisiting the Importance::High cases (which is almost every case), or adding additional importance levels, or adding more UI / Audio impact to importance levels.

----------------------
**Code Enhancement**

## Summary
One pain I experienced while playing the game was that important messages like escorts getting destroyed were getting lost. I noticed that there is already an Open issue about it from 4 years ago (#3361). In that issue, one comment suggests adding a priority concept to Messages. So I decided to look at the code, and I found that there is already a boolean called isImportant in the Messages::Add method, however, it is defaulted to 'true', and pretty much every application of the Add method was using the default. On reviewing them, I felt that most likely the authors of those instances perhaps did not consciously choose to mark them as important. This PR is NOT about fixing those concerning cases. Instead, this PR is purely about making sure that in future if any author decides to add a Message, they must be required to make a conscious decision on how important their message is. As such, the parameter is no longer optional. Furthermore, the parameter is now an enum class, ensuring that the usage of the API will make the intended importance crystal clear. It also is now High / Low instead of True / False, just in-case in future someone decides to add more levels.

Note that this is my first PR. I've tried to follow the style guide (namely making sure that I wrap at 120). I've seen the contribution.md file. But if I missed anything or if you don't like anything, let me know; happy to change.

## Usage Examples
`Messages::Add("Disengaging autopilot.", Messages::Importance::High);`

## Testing Done
Played the game and tested both a High and Low importance message. They work, though they do not seem visually different, presumably because the game has no visual impact of message importance.

## Performance Impact
N/A

### DIFF
--- a/source/Messages.h
+++ b/source/Messages.h
@@ -34,7 +34,7 @@ public:
 		std::string message;
 	};
     
-    enum class Importance {
+    enum class Importance : uint_least8_t {
         High,
         Low
     };
--- a/source/Messages.h
+++ b/source/Messages.h
@@ -15,6 +15,7 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include <string>
 #include <vector>
+#include <cstdint>
 
 
 
--- a/source/Messages.h
+++ b/source/Messages.h
@@ -33,10 +33,15 @@ public:
 		int step;
 		std::string message;
 	};
+    
+    enum class Importance {
+        High,
+        Low
+    };
 	
 public:
-	// Add a message to the list.
-	static void Add(const std::string &message, bool isImportant = true);
+	// Add a message to the list along with its level of importance
+	static void Add(const std::string &message, Importance importance);
 	
 	// Get the messages for the given game step. Any messages that are too old
 	// will be culled out, and new ones that have just been added will have
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -160,14 +160,14 @@ namespace {
 		{
 			for(Ship *ship : toDeploy)
 				ship->SetDeployOrder(true);
-			Messages::Add("Deployed " + to_string(toDeploy.size()) + " carried ships.");
+			Messages::Add("Deployed " + to_string(toDeploy.size()) + " carried ships.", Messages::Importance::High);
 		}
 		// Otherwise, instruct the carried ships to return to their berth.
 		else if(!toRecall.empty())
 		{
 			for(Ship *ship : toRecall)
 				ship->SetDeployOrder(false);
-			Messages::Add("Recalled " + to_string(toRecall.size()) + " carried ships");
+			Messages::Add("Recalled " + to_string(toRecall.size()) + " carried ships", Messages::Importance::High);
 		}
 	}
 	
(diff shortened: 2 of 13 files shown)


source/Messages.h
```diff
@@ -33,10 +33,15 @@ class Messages {
 		int step;
 		std::string message;
 	};
+    
+    enum class Importance {
```
tehhowch, line 37: Let's request a smaller type, e.g. `int_least8_t`.

tehhowch (commented):
Looks good so far!
Along with this change, you could consider modifying the layout of `incoming` to support multiple importance levels.

Zitchas:
I'm curious: What is the impact of a message being set to be not important? Is it displayed for a shorter time? Does it start out faded out compared to the important ones?

Amazinite:
> I'm curious: What is the impact of a message being set to be not important? Is it displayed for a shorter time? Does it start out faded out compared to the important ones?

Say you have messages A, B. First message A appears, then message B appears. So in the list of hails, you have message A on top of message B. Then message A is sent again. If A is important, the older version of A is removed from the list and readded to the bottom of the list so that we now have message B on top of message A. If A is unimportant, it sees that message A is already on the list and therefore doesn't send it again, keeping message A on top of message B.

Zitchas:
OK, Thanks for the explanation. Sounds like a good change.

source/Messages.h
salqadri: Great suggestion. I made it uint_least8_t.

salqadri:
> Looks good so far!
> Along with this change, you could consider modifying the layout of `incoming` to support multiple importance levels.

@tehhowch In this same PR? I was hoping to keep this PR be focused on just the interface change to keep code changes as minimal as possible, especially as it already is impacting many files.

tehhowch:
> 
> 
> > Looks good so far!
> > Along with this change, you could consider modifying the layout of `incoming` to support multiple importance levels.
> 
> In this same PR? I was hoping to keep this PR be focused on just the interface change to keep code changes as minimal as possible, especially as it already is impacting many files.

The only impact of that change would be to the Messages class, in the implementations of `Messages::Add` and `Messages::Get`. It's a change that is logically consistent with the change you are making--send arbitrary levels of importance--resulting in _storing_ arbitrary levels of importance.

I agree that changing the existing message importance categorization and expanding the discrete number of importances that the UI supports are both well-suited to a different PR.

salqadri:
> > > Looks good so far!
> > > Along with this change, you could consider modifying the layout of `incoming` to support multiple importance levels.
> > 
> > 
> > In this same PR? I was hoping to keep this PR be focused on just the interface change to keep code changes as minimal as possible, especially as it already is impacting many files.
> 
> The only impact of that change would be to the Messages class, in the implementations of `Messages::Add` and `Messages::Get`. It's a change that is logically consistent with the change you are making--send arbitrary levels of importance--resulting in _storing_ arbitrary levels of importance.
> 
> I agree that changing the existing message importance categorization and expanding the discrete number of importances that the UI supports are both well-suited to a different PR.

@tehhowch Done. Please take a look.

PR #6124, opened 2021-07-18
Fix a bug where causes an undefined behavior.

oo13:
**Bugfix:** This PR addresses issue causes an undefined behavior in Job Board.

It's undefined behavior to compare iterators that refer to elements of different containers. (§ 24.2.5 [forward.iterators#2])

The debug library on GCC (activated by -D_GLIBCXX_DEBUG) aborts the program when you enter Job Board from the planet panel.

## Fix Details
This PR makes the function receive a parameter for the comparison.

## Testing Done
I added option -D_GLIBCXX_DEBUG, and entered the Job Board.


tehhowch (approved):
And here we were, entertaining dragons.

Thanks!

PR #6125, opened 2021-07-18
Fix flickering in the shop panel when going up with nothing selected.

### LINKED ISSUE #5966: Shipyard flickers when first selecting a ship using the keyboard
**Describe the bug**
In the shipyard screen, if no ship is selected and you press up arrow on the keyboard, the display flickers for a moment.

**To Reproduce**
1. Start a new pilot.
2. Get to the point where you choose an initial ship.
3. Press up arrow repeatedly so you iterate through four states: no ship selected and each of the starting ships selected in turn. Whenever you change from "no ship is selected" to "Sparrow is selected", the display flickers. More precisely, there is a point in which only the `Interceptor` category and the Sparrow are shown, the other categories and ships aren't shown at that point.

**Expected behavior**
The shipyard screen shouldn't flicker when the selection is changed.

**Screenshots**

https://user-images.githubusercontent.com/24784687/118292573-4d673380-b4c8-11eb-8148-bb92e168a4f2.mp4

What the video shows: first, no ship is selected. Then, for a single frame, only Interceptor/Sparrow are shown, the other headings and ships aren't. Then, all ships are listed but the Sparrow is selected.

 - OS: Linux
 - Game Source: self-compiled
 - Version: 36cc1d9e49248c7a01ad96c77dc4fd73c9df90c4

**Link to save file**
N/A

**Additional context**
This doesn't happen when selecting a ship with the mouse, or when changing the selection with the keyboard once a ship is already selected.

ES runs in an 1820x980window. It seems the shipyard is just a few pixels taller than the window: when the Shuttle is selected the "Scroll down" arrow is white, and when the Sparrow is selected the "Scroll up" arrow is white.

quyykk:
**Bugfix:** This PR fixes #5966

## Fix Details

The wrong sign was used when calculating the amount of scroll to apply when nothing is selected and the Up key is pressed. This resulted in way too much scroll - the last entry moved to the very top of the screen and then in the next frame got immediately moved back down - which caused the flickering.

## Testing Done

Verified that the scenario in the issues is fixed and that this didn't introduce any other bugs.

### DIFF
--- a/source/ShopPanel.cpp
+++ b/source/ShopPanel.cpp
@@ -1118,7 +1118,7 @@ void ShopPanel::MainUp()
 	if(it == zones.end())
 	{
 		--it;
-		mainScroll = max(0., mainScroll + it->Center().Y() + start->Center().Y());
+		mainScroll = max(0., min(maxMainScroll, mainScroll + it->Center().Y() + start->Center().Y()));
 		selectedShip = it->GetShip();
 		selectedOutfit = it->GetOutfit();
 		return;
--- a/source/ShopPanel.cpp
+++ b/source/ShopPanel.cpp
@@ -1047,7 +1047,7 @@ void ShopPanel::MainLeft()
 	if(it == zones.end())
 	{
 		--it;
-		mainScroll += it->Center().Y() - start->Center().Y();
+		mainScroll = maxMainScroll;
 		selectedShip = it->GetShip();
 		selectedOutfit = it->GetOutfit();
 		return;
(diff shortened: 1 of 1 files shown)


tehhowch:
With this PR as of 6c06b05d3d6e04790cf14fbdf65f258e75d1eeeb, I still see a very strong flicker in the outfitter, when using F11-fullscreen (1920x1080), and when running non-maximized windowed mode.

You can use this save to repro.
[outfitter scroll test.txt](https://github.com/endless-sky/endless-sky/files/6837196/outfitter.scroll.test.txt)

Thanks for picking this up!

quyykk:
Thanks! Fixed in latest commit.

tehhowch (approved):
There were a few other keypress scenarios that triggered overscrolling, which should be addressed now

PR #6127, opened 2021-07-18
Make escorts only pathfind through systems that the player has visited.

### LINKED ISSUE #6096: Non-jumpdrive ships navigate through yet-discovered wormhole
**Describe the bug**
In the Remnant mission stack, when the player is offered the job to drop off sensors in Parca, if the player has a blend of jump-drive and non-jumpdrive ships, the non-jump ships will use the wormhole path, even if the player hasn't discovered it yet.

**To Reproduce**
From Viminal, prior to first visit to Parca, navigate there with a blend of jumpdrive equipped and hyperdrive equipped ships.  
At Stercutus, jump to Levana -> Coluber -> Parca.

Hyperdrive vessels will follow to Parca.

Attached savefile has a jumpdrive equipped Bactrian and three Shield Beetles, plus two hyper-drive SBs.
[Speedy Smith~wormhole.txt](https://github.com/endless-sky/endless-sky/files/6751331/Speedy.Smith.wormhole.txt)

OS is Ubuntu 20.04,
Behaviour verified on three different game versions

1) the @Amazinite mining turret/adjusted commodities pull request, June 15 version. 
2) Distro 0.9.12
3) Distro 0.9.13

quyykk:
**Bugfix:** This PR fixes #6096 

## Fix Details

I made `AI` aware of `PlayerInfo` so that it knows which systems the player has visited or not. `DistanceMap` got a new constructor so that you can add a `PlayerInfo` to its pathfinding for a ship that is not the flagship.

Now escorts will not travel through systems that the player hasn't visisted.

## Testing Done
I used the save file to reproduce the issue and fix it.

## Save File
The save file from the linked issue above.


quyykk:
Is this solution/PR not wanted? I mean I get that passing a `PlayerInfo` to `AI` is not pretty but I'm not really seeing an alternative. Maybe adding a way for `GameData` to access HasSeen/HasVisited but that's not exactly better either. Or is this fix not wanted? Basically I need some form of feedback :)

Zitchas:
In my experience, the Devs are pretty quick to slap a "unlikely" label on stuff they don't think will get into the game. And usually say something to that effect. So it's most likely that they just haven't been able to get to it yet; or haven't felt that they have the time to really dig into this as of yet.

pepoluan:
The game hasn't already done this??

Ahaha, I thought the game already did this, which is why whenever a new area opened up I rode solo through the space using a JD+HD combo (to save fuel) to map systems in the area...

petervdmeer (dismissed):
Nice; the actual pathfinding code was already there; the only change that was needed was to pass the PlayerInfo for player-owned escorts.

source/DistanceMap.h
```diff
@@ -41,7 +41,9 @@ class DistanceMap {
 	// Calculate the path for the given ship to get to the given system. The
 	// ship will use a jump drive or hyperdrive depending on what it has. The
 	// pathfinding will stop once a path to the destination is found.
-	DistanceMap(const Ship &ship, const System *destination);
+	// If a player is given, the path will only include systems that the
+	// player has visisted.
+	DistanceMap(const Ship &ship, const System *destination, const PlayerInfo *player = nullptr);
```
tehhowch, line 46: Probably we should refactor `DistanceMap` to take an optional set of known systems, rather than a player object.

(We can do that in a sooner-rather-then-later follow-up though.)

source/DistanceMap.h
```diff
@@ -41,7 +41,9 @@ class DistanceMap {
 	// Calculate the path for the given ship to get to the given system. The
 	// ship will use a jump drive or hyperdrive depending on what it has. The
 	// pathfinding will stop once a path to the destination is found.
-	DistanceMap(const Ship &ship, const System *destination);
+	// If a player is given, the path will only include systems that the
+	// player has visisted.
```
tehhowch, line 45: The "visisted" typo appears here and in the other file.

tehhowch (commented):
What should we do about a hyperdrive-only escort that has no known route to the flagship system because the player used their JD to get there and didn't buy maps?

Should the player get a message every time they take off that they need to go back / explore sideways?
Should we provide a new escort pip color in the map to highlight "stranded" escorts, and a tooltip as well?
Should we let the escort ship explore on its own, perhaps adding the systems it visits to the player's "seen" systems?
Should we let the escort ship "cheat" and take the shortest path between known paths?
Should we restrict the map view to known-accessible-via-hyperspace-lane-only systems, placing an information "fog-of-war" on star clusters not accessible by hyperlink from the player's current system (no trade info, etc.), completely cutting the player off from their stranded escorts?

quyykk:
> What should we do about a hyperdrive-only escort that has no known route to the flagship system because the player used their JD to get there and didn't buy maps?

We already have that problem: What happens currently if the flagship system has no links? The same behavior should happen in this case. I think the escort just stays in the system they were in. I think that's fine.

> Should the player get a message every time they take off that they need to go back / explore sideways?

No. But we can maybe add something to the escort info display on the lower right to indicate that the escort is stranded.

>  Should we provide a new escort pip color in the map to highlight "stranded" escorts, and a tooltip as well?

I think that's a good idea.

> Should we let the escort ship explore on its own, perhaps adding the systems it visits to the player's "seen" systems?

No, that would be weird.

> Should we let the escort ship "cheat" and take the shortest path between known paths?

Nope.

> Should we restrict the map view to known-accessible-via-hyperspace-lane-only systems, placing an information "fog-of-war" on star clusters not accessible by hyperlink from the player's current system (no trade info, etc.), completely cutting the player off from their stranded escorts?

Maybe, maybe not, but in any case I think that's for another PR if we were to do that.

petervdmeer:
"Waiting on OP" due to open review-comments and unresolved merge-conflict.

source/DistanceMap.h
quyykk: We need the player object as well to check if a wormhole was visited by the player.

tehhowch: That's required by the current implementation, yes. If we're passing in the knowledge of usable systems, passing in the knowledge of usable planets/wormholes is probably expected.

tehhowch (commented):
Last thing for this PR

Amazinite:

Closing this PR because Quyykk himself has said that he's not getting to it soon and that he doesn't think it's too important of a change. I'll also note that there has been some discussion of reviving #5979, which would conflict with this change and cause player escorts to become stuck in shrouded systems because of those system being removed from the visited list. If we are to make both changes, then each change needs to keep the other in mind.

Zitchas:
Just a gentle nudge that this - while not a priority - would be much appreciated fix for escort behavior.

source/AI.h
```diff
@@ -180,6 +185,7 @@ template <class Type>
 
 private:
 	// Data from the game engine.
+	const PlayerInfo &player;
```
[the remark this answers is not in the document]
quyykk: @tehhowch I'm having a bit of trouble trying to avoid the `player` field. I would need to pass around 3 sets (seen systems, visited systems, visited planets) around to avoid using `player`, which isn't that great imo.

So, what would you recommend?

ziproot:
It looks like this is waiting on reviewers. There are no merge conflicts and no unaddressed review comments as far as I can tell.

PR #6130, opened 2021-07-19
Carries dock on their respective bay instead of boarding the ship with optional effects.

quyykk:
**Feature:** This PR implements the feature request detailed and discussed in issue #957 and #944

## Feature Details
Currently, recalling carries (fighters/drones) makes them board the ship. This PR changes that and makes them board on the bay. Every carry gets a bay where it always lands. This is also enables the use of retrieve effects that are created when the carry boards its bay.

## Video

Here's how it is currently (with launch effect).

https://user-images.githubusercontent.com/85879619/126198176-166997e7-c032-4906-9abf-02dba25e2aa1.mov

And here with this PR. You can also see the retrieve effect.

https://user-images.githubusercontent.com/85879619/126198307-57c5ffdf-4e7f-465b-bba4-0141488e167c.mov

## Usage Examples

Retrieve effects are defined exactly as launch effects. Here's an example:

```
bay Fighter -47 -24
    under
    "launch effect" "basic launch" 
    "retrieve effect" "basic launch" 
```

## Testing Done

I tested recalling fighters in various scenarios and defining retrieve effects. Here is a test save: 
[swizzing.txt](https://github.com/endless-sky/endless-sky/files/6842650/swizzing.txt)

## Performance Impact
The retrieve effects maybe but apart from that, no.


petervdmeer:
Hi, welcome as a new contributor!
The movies look nice.

I do have some concerns about corner-cases; did you also test:
- What happens when an carried ship gets destroyed? (Does the carriers bay get freed up for use by another carried ship?)
- What happens when an carried ship gets captured? (Does the carriers bay get freed up for use by another carried ship?)
- What happens when a carrier gets destroyed? (Do deployed carried ships find a new carrier as parent? Do deployed carried ships stay alive?)
- What happens when a carrier gets captured? (Do deployed carried ships find a new carrier as parent? Do deployed carried ships still keep their original owner?)
- What happens when a carrier leaves to a different system? (Similar to destroyed; do the carried ships find a new carrier?)
- If newly captured carried ships find a carrier to dock with?

quyykk:
Thanks for the feedback!

It turns out none of the scenarios you mentioned worked out. I managed to fix some of them but then the code became a bit of a mess. So I rewrote it. The only thing left to test is capturing, because I'm having a hard time creating a proper save file for that scenario.

> What happens when an carried ship gets destroyed? (Does the carriers bay get freed up for use by another carried ship?)

The bay is freed and any other carry without a free bay uses that bay.

> What happens when an carried ship gets captured? (Does the carriers bay get freed up for use by another carried ship?)

**untested**: The bay is freed.

> What happens when a carrier gets destroyed? (Do deployed carried ships find a new carrier as parent? Do deployed carried ships stay alive?)

The carried ships all search for a new free bay and use one if any. The deployed ships stay alive.

> What happens when a carrier gets captured? (Do deployed carried ships find a new carrier as parent? Do deployed carried ships still keep their original owner?)

**untested**: Any deployed ships keep their original owner and search for a new free bay. Any undeployed ships are owned by the capturer and can get immediately deployed/used.

> What happens when a carrier leaves to a different system? (Similar to destroyed; do the carried ships find a new carrier?)

The carried ships find a new carrier to dock. Here there is a difference to the current behavior: If you return to the same system and recall, some carries don't dock to the jumping carrier. With this PR they do.

> If newly captured carried ships find a carrier to dock with?

**untested**: They do.

quyykk:
Sometimes a carried ship becomes "stranded". It just stands there doing nothing. What happens is that it can't find a new parent because it's unlucky:

https://github.com/endless-sky/endless-sky/blob/a2dc407dd265638e5d469b53041c690e7aa724b1/source/AI.cpp#L779

Can I remove this randomness? It will cause any carried ship that escorts a carrier without a bay to always search for a new parent. Will probably have an impact on performance but I don't know how much and if it's worth it.

source/Ship.h
```diff
@@ -61,12 +61,14 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
...
 		
 		// Copying a bay does not copy the ship inside it.
-		Bay(const Bay &b) : point(b.point), category(b.category), side(b.side), facing(b.facing), launchEffects(b.launchEffects) {}
+		Bay(const Bay &b) : point(b.point), category(b.category), side(b.side), facing(b.facing), launchEffects(b.launchEffects), retrieveEffects(b.retrieveEffects) {}
 		Bay &operator=(const Bay &b) { return *this = Bay(b); }
 		
 		Point point;
 		std::shared_ptr<Ship> ship;
 		std::string category;
+		// If this bay isn't reversed by a carry.
```
petervdmeer, line 70: typo: reversed -> reserved

source/Ship.h
```diff
@@ -448,6 +456,8 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	// Characteristics of this particular ship:
 	std::string name;
 	bool canBeCarried = false;
+	// The bay this ship docks too, if any.
+	Bay *reservedBay = nullptr;
```
petervdmeer, line 460: The parent-ship is handled by a weak-pointer. Is it safe to use a regular pointer here?

source/Ship.h
```diff
@@ -61,12 +61,14 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
...
 		// Copying a bay does not copy the ship inside it.
-		Bay(const Bay &b) : point(b.point), category(b.category), side(b.side), facing(b.facing), launchEffects(b.launchEffects) {}
+		Bay(const Bay &b) : point(b.point), category(b.category), side(b.side), facing(b.facing), launchEffects(b.launchEffects), retrieveEffects(b.retrieveEffects) {}
 		Bay &operator=(const Bay &b) { return *this = Bay(b); }
 		
 		Point point;
 		std::shared_ptr<Ship> ship;
 		std::string category;
+		// If this bay isn't reversed by a carry.
+		bool free = true;
```
petervdmeer, line 71: It feels a bit redundant to track the occupied/free state in a separate variable (but I also don't have an immediate suggestion on how to improve on this).

petervdmeer (changes requested):
I didn't review the full PR, but I do have some comments on the parts that I did review.

source/Ship.cpp
```diff
@@ -1813,13 +1824,19 @@ void Ship::Move(vector<Visual> &visuals, list<shared_ptr<Flotsam>> &flotsam)
 	// moment, its boarding target should be its parent ship.
 	if(CanBeCarried() && !(target && target == GetShipToAssist()))
 		target = GetParent();
-	if(target && !isDisabled)
+	if(target && !isDisabled && commands.Has(Command::BOARD))
```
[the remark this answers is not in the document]
quyykk: Only for boarding. You can see how the two `if` statements after updating `isBoarding` all require that it is true.

**But**: It seems like `isBoarding` is updated only in that part of the code, so I moved it back or else `isBoarding` would always be `true` after it being true once. Thanks!

source/AI.cpp
```diff
@@ -835,13 +835,24 @@ void AI::Step(const PlayerInfo &player, Command &activeCommands)
...
-				MoveTo(*it, command, parent->Position(), parent->Velocity(), 40., .8);
-				command |= Command::BOARD;
-				it->SetCommands(command);
-				continue;
+
+				if(const auto *bay = it->GetBay())
+				{
+					Point exitPoint = parent->Position() + parent->Facing().Rotate(bay->point);
+					if(MoveTo(*it, command, exitPoint, parent->Velocity(), 15., .3))
+						command |= Command::BOARD;
```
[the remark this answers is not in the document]
quyykk: For the first question it was to prevent the carry docking before being above the bay (because you can board the ship from anywhere).

For the second - no. We want the carry to move to its bay.

source/Ship.cpp
```diff
@@ -2744,6 +2756,12 @@ void Ship::WasCaptured(const shared_ptr<Ship> &capturer)
 	}
 	// This ship should not care about its now-unallied escorts.
 	escorts.clear();
+
+	// Fighters should flee a disabled ship, but if the player manages to capture
+	// the ship before they flee, the fighters are captured, too.
+	for(Bay &bay : bays)
+		if(bay.ship)
+			bay.ship->WasCaptured(shared_from_this());
```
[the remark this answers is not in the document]
quyykk: So that the carrier can immediately get control over the carries.

source/Ship.h
quyykk: Now that I think about it, no. Because it points to a `std::vector` which if any bays get added later (due to an event) will invalidate any pointers. I'll change this to a `std::list`.

We can't use a `weak_ptr` here because the bays aren't `shared_ptr`.

source/Ship.h
quyykk: We kinda have to. We need to have a way of saying: this bay is used by a ship but that ship isn't docked to it.

source/Ship.cpp
```diff
@@ -3266,39 +3276,75 @@ bool Ship::CanBeCarried() const
...
+	if(visuals)
+		for(const Effect *effect : dockBay->retrieveEffects)
+			visuals->emplace_back(*effect, ship->Position(), ship->Velocity(), ship->Facing());
+	return true;
+}
+
+
+
+// Track the given escort as a carry in one bay while in flight.
+void Ship::CarryInFlight(const shared_ptr<Ship> &escort)
```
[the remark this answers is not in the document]
quyykk: `Carry` is when the carry boards the ship. Specifically, it does things like settings planets and system to `nullptr`, storing cargo inside its parent, ...

We need to have a way for carries to tell the parent that they want a bay. Previously this wasn't needed carries that need a bay would just 1) as the parent if there are free bays and 2) then board. This is not possible anymore since the carry (and the ship) need to now beforehand which bay they will dock too. Hence this function.

source/Ship.cpp
```diff
@@ -3266,39 +3276,75 @@ bool Ship::CanBeCarried() const
...
-			ship->isReversing = false;
-			ship->isSteering = false;
-			ship->commands.Clear();
-			// If this fighter collected anything in space, try to store it
-			// (unless this is a player-owned ship).
-			if(!isYours && cargo.Free() && !ship->Cargo().IsEmpty())
-				ship->Cargo().TransferAll(cargo);
-			// Return unused fuel to the carrier, for any launching fighter that needs it.
-			ship->TransferFuel(ship->fuel, this);
+	dockBay->ship = ship;
```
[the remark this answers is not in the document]
quyykk: Pretty much. `dockBay` aka `reservedBay` is the bay that the ship uses to dock. It only gets updated when `Carry*` is called or if the ship is being added as escort. 

**But**: I missed a case when the carrier gets docked. Any carries still flying will still hold a pointer to the old captured ship.

source/Ship.cpp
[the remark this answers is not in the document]
petervdmeer: Sounds like `ReserveBay()` is a better name for this function then (since the carried ship is not actually carried after this function, but only seems to have reserved a bay).

source/Ship.h
petervdmeer: I guess we could also ask the carried-ship if it is docked. (But performance wise we would then need to delay the bay-reservation until a ship actually wants to dock.)

quyykk: Meh. I could do it but I really don't think it's worth it. And also that bool is not *that* redundant imo :)

source/Ship.cpp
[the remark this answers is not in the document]
quyykk: I agree! Thanks for the suggestion.

source/Ship.h
petervdmeer, line 460: I'm still not happy by the direct pointer from one ship to anothers ship bay; all the other code around ships referring other ships is using weak_pointers which suggest that all other code is always assuming that the other ship could disappear any Engine::Step() (which is probably a valid assumption).

If the bays are in a vector, then it should be possible to refer to the bay by using the index-number in the vector. (Such an index number would be similar to a weak pointer in that it is possible to verify if the bay-index is valid before using the bay-index.)

The child/parent relation is relatively strictly defined, so I expect that it is possible to make this code work with a direct pointer, but if we really go down that path, then we need a lot more test-coverage before I consider this ready for merging.

source/Ship.cpp
```diff
@@ -322,13 +322,19 @@ void Ship::Load(const DataNode &node)
...
 				{
-					// Load in the effect(s) to be displayed when the ship launches.
+					// Load in the effect(s) to be displayed when the ship launches/boards.
 					if(grand.Token(0) == "launch effect" && grand.Size() >= 2)
 					{
 						int count = grand.Size() >= 3 ? static_cast<int>(grand.Value(2)) : 1;
 						const Effect *e = GameData::Effects().Get(grand.Token(1));
 						bay.launchEffects.insert(bay.launchEffects.end(), count, e);
 					}
+					else if(grand.Token(0) == "retrieve effect" && grand.Size() >= 2)
```
petervdmeer, line 332: This PR contains 2 changes; adding the `"retrieve effect"` and the functionality to let the carried ship dock on the bay location instead of the ships center.
I have the impression that the `"retrieve effect"` is a lot simpler to test and to review, so you might want to consider to put that in a separate PR (to try to get it merged ahead of the docking at the bay location).

quyykk: I can't. The retrieve effect is generated at the point where the carry docks at a specific bay, which I can't do without the rest of the PR. I could also remove it and then PR it later when this gets (hopefully) merged.

source/Ship.h
quyykk: I think it's pretty safe, since `reservedBay` is really only used when we know that the escort has a parent and the bay is also part of the parent. But I'll try to find another solution; index sound goods thought, thanks.

source/Ship.cpp
petervdmeer: Without this PR the game also knows at the moment of docking which bay is assigned to which carried ship (since the ship gets assigned to a bay at the moment of docking and drawn there (for external bays) from the moment of docking onward).

But I didn't look into the details and if this is difficult, then I'm also okay with just keeping it as part of this PR.

quyykk: Yes that is true. I'll clarify: I can make the effect happen on a specific bay, but then it looks off since carries still board the ship and not the dock.

petervdmeer: I agree, the effect works best if both PRs are merged.

Zitchas:
That certainly looks nice!

source/Ship.h
```diff
@@ -358,13 +362,17 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 	// Check if this is a ship of a type that can be carried.
 	bool CanBeCarried() const;
 	// Move the given ship into one of the bays, if possible.
-	bool Carry(const std::shared_ptr<Ship> &ship);
+	bool Carry(const std::shared_ptr<Ship> &ship, std::vector<Visual> *visuals = nullptr);
```
tehhowch, line 365: Let's drop the default parameter

```suggestion
	bool Carry(const std::shared_ptr<Ship> &ship, std::vector<Visual> *visuals);
```

I think ideally we would have two methods, one called from external contexts with visuals available and the other called from an internal context where visuals do not make sense to display (such as the preflight loading of carried ships post-recharge), but there is some additional refactoring needed to reach that state.

source/Ship.h
```diff
@@ -427,7 +435,7 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
 private:
 	// Add or remove a ship from this ship's list of escorts.
 	void AddEscort(Ship &ship);
-	void RemoveEscort(const Ship &ship);
+	void RemoveEscort(Ship &ship);
```
tehhowch, line 438: Should the reservedBay field be mutable?

source/Ship.cpp
```diff
@@ -3372,39 +3388,75 @@ bool Ship::CanBeCarried() const
...
+	ship->SetSystem(nullptr);
+	ship->SetPlanet(nullptr);
+	ship->SetTargetSystem(nullptr);
+	ship->SetTargetStellar(nullptr);
+	ship->SetParent(shared_from_this());
+	ship->isThrusting = false;
+	ship->isReversing = false;
+	ship->isSteering = false;
+	ship->commands.Clear();
+	// If this fighter collected anything in space, try to store it
```
tehhowch, line 3426: Do we need to clear the fire commands here, too?

(I.e. a bug we missed in that other PR)

source/Ship.cpp
```diff
@@ -3372,39 +3388,75 @@ bool Ship::CanBeCarried() const
...
+		ship->Cargo().TransferAll(cargo);
+	// Return unused fuel to the carrier, for any launching fighter that needs it.
+	ship->TransferFuel(ship->fuel, this);
+
+	// Update the cached mass of the mothership.
+	carriedMass += ship->Mass();
+
+	// Apply any retrieve effects if applicable (for example, no effect should
+	// be generated when launching from a planet).
+	if(visuals)
```
tehhowch, line 3438:
```suggestion
	if(visuals && !dockBay.retrieveEffects.empty())
```

source/Ship.cpp
```diff
@@ -3789,19 +3852,41 @@ const vector<weak_ptr<Ship>> &Ship::GetEscorts() const
 void Ship::AddEscort(Ship &ship)
 {
 	escorts.push_back(ship.shared_from_this());
+
+	// Check to see if the escort wants and can be carried.
+	// The escort is maybe already carried.
+	if(!ship.CanBeCarried() || ship.reservedBay)
```
tehhowch, line 3858: Should "ignore it" be the behavior when AddEscort is called more than once? Or should it be "last parent wins"?

I'm curious the scenarios that result in multiple ships attempting to claim the same fighter

tehhowch (changes requested):
First pass, haven't playtested

quyykk:
Thanks for the review! Apparently Github didn't think it was necessary to notify me of it so I discovered it when I randomly clicked on this PR 5 minutes ago. I will address your comments soonish.

petervdmeer:
"Waiting on OP" for the soonish update.

Amazinite:

Closing this PR because Quyykk himself has said that he's not getting to it soon and that he doesn't think it's too important of a change.

quyykk:
I've tested this PR again, seems to still work :smile:. However, I still would like to do some more tests to catch any edge cases.

source/Ship.h
```diff
@@ -404,12 +408,16 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
...
+	// Reserve a bay for the given escort.
+	void ReserveBay(const std::shared_ptr<Ship> &escort);
 	// Empty the bays. If the carried ships are not special ships that are
 	// saved in the player data, they will be deleted. Otherwise, they become
 	// visible as ships landed on the same planet as their parent.
 	void UnloadBays();
 	// Get a list of any ships this ship is carrying.
-	const std::vector<Bay> &Bays() const;
+	const std::list<Bay> &Bays() const;
+	// Returns the bay this ship docks too if any.
```
petervdmeer, line 419: Typo: too => to

source/Ship.cpp
```diff
@@ -4739,6 +4812,23 @@ void Ship::CreateSparks(vector<Visual> &visuals, const Effect *effect, double am
...
+
+// Move the given ship into one of the bays (if possible) and generate any visuals.
+bool Ship::Carry(const shared_ptr<Ship> &ship, vector<Visual> &visuals)
+{
+	bool success = Carry(ship);
+
+	// Apply any retrieve effects. If the ship was carried succesfully then
+	// it was assigned a bay with the effects to generate.
+	if(success && !ship->reservedBay->retrieveEffects.empty())
+		for(const Effect *effect : ship->reservedBay->retrieveEffects)
```
Amazinite, line 4824: No difference in behavior, I imagine.
```suggestion
	if(success)
		for(const Effect *effect : ship->reservedBay->retrieveEffects)
```

source/Ship.cpp
```diff
@@ -3132,18 +3156,26 @@ void Ship::UnloadBays()
 			bay.ship->SetPlanet(landingPlanet);
 			bay.ship->UnmarkForRemoval();
 			bay.ship.reset();
+			bay.free = true;
```
Amazinite, line 3159: When a fighter is unloaded from a bay, doesn't its reserved bay remain the bay that it was unloaded from? In that case, shouldn't the bay not be considered free?

source/Ship.h
```diff
@@ -404,12 +408,16 @@ class Ship : public Body, public std::enable_shared_from_this<Ship> {
...
+	const Bay *GetBay() const;
```
Amazinite, line 420: I think it'd make more sense to call this `GetReservedBay()`.

tibetiroka:
Closing as orphaned since there are unresolved comments from March, and the CI failures haven't been resolved. Feel free to reopen this once the comments are addressed.

Issue #4794, opened 2020-02-07
Add ability to review on-screen message history.


WildLynxDev:
Problem:
Mission failed! 
When, how? It was before I landed at the previous system or after? Dam it!

Also:
That Merchant captain said something witty, but it scrolled up before I could read it completely.

Solution:
Store all messages at temporary file, clear it on savegame load and game start, add keyboard controls for scrolling messages back [Page Up], forth [PgDown] and to the last [End].

jafdy:
Mission failed messages should be solvable by adding a dialog on landing.

Amazinite:
Ref: #3361

Zireael07:
Most games with many messages (RPGs, roguelikes) have a 'message/log history' screen where you can scroll back at your leisure. This often, but not always overlaps with storing the log in a file, as mentioned by OP.

tehhowch:
Discussion should be redirected to the original issue for duplicates.

WildLynxDev:
Mission failure is only a one of the reasons. Sometimes random passing ships leave comments and they scroll up too fast to read, especially if you trying to dodge missiles at the some time.

PR #6131, opened 2021-07-19
Highlight critical messages

### LINKED ISSUE #3361: Mission failure by ship loss is very hard to spot

salqadri:
-----------------------
**Feature:** This PR resolves #3361. Critical messages like mission failures are now highlighted in bright red, and then slowly fade away as other messages do.

## Feature Details
This is achieved by introducing a new importance level, called Highest, and exposing it to the game engine via a new field in the Messages::Entry class. I've also gone ahead and applied it to some messages that I deemed should be highlighted in such a way, such as mission failures. In the process I also took the luxury of adding fixed named colors to the Color class, as I wanted to be able to set the color of the message to `Color::Red`.

## UI Screenshots
MissionFailRed

## Usage Examples
`Messages::Add("Mission failed.", Messages::Importance::Highest);`

## Testing Done
I launched the game, gave myself a mission that spanned the cargo of my combined fleet, went to a pirate system and let some of my escorts be killed and verified that the resulting mission failure message was highlighted Red and faded away without being repeated, and all other non-critical messages remained as they were.

## Performance Impact
N/A

-----

Update: Based on the highest vote in a Poll on Discord, the color has now been set to Orange. Also, I now take the color from the interfaces.txt file so we can easily change this in future. You can even define different colors for each importance level via the interfaces.txt file.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -955,7 +955,7 @@ void Engine::Draw() const
 				break;
 		}
 		if(!color)
-			color = GameData::Colors().Get("messages importance default");
+			color = GameData::Colors().Get("message importance default");
 		messageLine.Draw(messagePoint, color->Additive(alpha));
 	}
 	
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -942,14 +942,23 @@ void Engine::Draw() const
 		if(messagePoint.Y() < messageBox.Top())
 			break;
 		float alpha = (it->step + 1000 - step) * .001f;
-		Color color;
-		if (it->importance == Messages::Importance::Highest) {
-			color = Color::Red.Additive(alpha);
+		const Color* color;
+		// This switch will make the compiler force us to handle any new importance levels we add in future
+		switch (it->importance) {
+			case Messages::Importance::Highest:
+				color = GameData::Colors().Get("message importance highest");
+				break;
+			case Messages::Importance::High:
+				color = GameData::Colors().Get("message importance high");
+				break;
+			case Messages::Importance::Low:
+				color = GameData::Colors().Get("message importance low");
+				break;
 		}
-		else {
-			color = Color(alpha);
+		if (color == nullptr) {
+			color = GameData::Colors().Get("messages importance default"); // Will not be null
 		}
-		messageLine.Draw(messagePoint, color);
+		messageLine.Draw(messagePoint, color->Additive(alpha));
 	}
 	
 	// Draw crosshairs around anything that is targeted.
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -942,22 +942,20 @@ void Engine::Draw() const
 		if(messagePoint.Y() < messageBox.Top())
 			break;
 		float alpha = (it->step + 1000 - step) * .001f;
-		const Color* color;
-		// This switch will make the compiler force us to handle any new importance levels we add in future
+		const Color *color = nullptr;
 		switch (it->importance) {
 			case Messages::Importance::Highest:
-				color = GameData::Colors().Get("message importance highest");
+				color = GameData::Colors().Find("message importance highest");
 				break;
 			case Messages::Importance::High:
-				color = GameData::Colors().Get("message importance high");
+				color = GameData::Colors().Find("message importance high");
 				break;
 			case Messages::Importance::Low:
-				color = GameData::Colors().Get("message importance low");
+				color = GameData::Colors().Find("message importance low");
 				break;
 		}
-		if (color == nullptr) {
-			color = GameData::Colors().Get("messages importance default"); // Will not be null
-		}
+		if(!color)
+			color = GameData::Colors().Get("messages importance default");
 		messageLine.Draw(messagePoint, color->Additive(alpha));
 	}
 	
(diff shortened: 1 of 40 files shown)


source/Engine.cpp
```diff
@@ -942,7 +942,13 @@ void Engine::Draw() const
...
 			break;
 		float alpha = (it->step + 1000 - step) * .001f;
-		Color color(alpha, 0.f);
+		Color color;
+		if (it->importance == Messages::Importance::Highest) {
+			color = Color::Red.Additive(alpha);
+		}
+		else {
+			color = Color(alpha);
+		}
```
Amazinite, line 951: The color used shouldn't be hardcoded, and should instead be defined in the interfaces file as to allow it to be easily customized. (And by extension, we shouldn't be hardcoding any colors into the Color class as you're doing.)
https://github.com/endless-sky/endless-sky/blob/d712e94e3923a81dd1ab14afcc2339ee86790740/data/interfaces.txt#L92-L95

salqadri: Ok makes sense, thanks; will update.

salqadri: Done

Zitchas:
I definitely like this idea, and I know that I, personally, would much prefer to have the highest alert level be either bright red or orange. (I'm good either way).

The one caveat is that I would also like there to be alternate options, if possible. Having the critical messages be white, but bold and/or a slightly bigger text size would be very valuable. While I wouldn't use it myself, I know there are people that have a lot of trouble seeing red / shades of red, or can only barely be able to do so. Thus having an alternative that still makes the text stand out but doesn't rely on color to do so would be very helpful in terms of accessibility.

salqadri:
> I definitely like this idea, and I know that I, personally, would much prefer to have the highest alert level be either bright red or orange. (I'm good either way).
> 
> The one caveat is that I would also like there to be alternate options, if possible. Having the critical messages be white, but bold and/or a slightly bigger text size would be very valuable. While I wouldn't use it myself, I know there are people that have a lot of trouble seeing red / shades of red, or can only barely be able to do so. Thus having an alternative that still makes the text stand out but doesn't rely on color to do so would be very helpful in terms of accessibility.
> 

Accessibility is an interesting point that I hadn't considered. Better to tackle that more holistically though in a separate ticket.

salqadri:
Results of poll on the color for critical messages:

image

source/Engine.cpp
```diff
@@ -942,8 +942,23 @@ void Engine::Draw() const
 		if(messagePoint.Y() < messageBox.Top())
 			break;
 		float alpha = (it->step + 1000 - step) * .001f;
-		Color color(alpha, 0.f);
-		messageLine.Draw(messagePoint, color);
+		const Color* color;
```
tehhowch, line 945: Always initialize pointers:

```suggestion
		const Color *color = nullptr;
```

source/Engine.cpp
```diff
@@ -942,8 +942,23 @@ void Engine::Draw() const
...
+		// This switch will make the compiler force us to handle any new importance levels we add in future
```
tehhowch, line 946:
```suggestion
```
Unnecessary comment

source/Engine.cpp
```diff
@@ -942,8 +942,23 @@ void Engine::Draw() const
...
+		switch (it->importance) {
+			case Messages::Importance::Highest:
+				color = GameData::Colors().Get("message importance highest");
+				break;
+			case Messages::Importance::High:
+				color = GameData::Colors().Get("message importance high");
+				break;
+			case Messages::Importance::Low:
+				color = GameData::Colors().Get("message importance low");
```
tehhowch, line 955: if you want to fall back to a default, these should use `.Find` instead of `.Get`, as `Set<T>::Get` will always return a valid pointer.

source/Engine.cpp
```diff
@@ -942,8 +942,23 @@ void Engine::Draw() const
...
+				break;
+		}
+		if (color == nullptr) {
+			color = GameData::Colors().Get("messages importance default"); // Will not be null
+		}
```
tehhowch, line 960: Typo + styleguide corrections:
```suggestion
		if(!color)
			color = GameData::Colors().Get("message importance default");
```

tehhowch (changes requested):

source/Engine.cpp
salqadri: Done.

source/Engine.cpp
salqadri: Done.

source/Engine.cpp
salqadri: Done.

source/Engine.cpp
salqadri: Done.

bluss:
Thanks for fixing this! Would it make sense for these to fade slower than regular messages?

Zitchas:
Probably a little bit, yes.

tehhowch:
> Thanks for fixing this! Would it make sense for these to fade slower than regular messages?

@bluss @Zitchas

That's better addressed in a different manner, e.g. https://github.com/endless-sky/endless-sky/issues/4794

Issue #5457, opened 2020-10-17
Allow for more wormhole mechanics by defining wormholes separately from normal objects


Amazinite:
**Is your feature request related to a problem? Please describe.**
Currently, wormholes are simply two or more objects of the same name in separate systems. Wormholes are always two-way, unless you give one of the wormholes no sprite. More than two wormholes of the same name always link to each other in a looping fashion (see the Ember Waste).

There is little control over the behavior of a wormhole, but in order to add new mechanics to wormholes, we would currently need to add those mechanics to all objects as well, or at least the mechanics would need to be a child of object and be ignored if the object is not also a wormhole.

**Related Issue Links**
N/A

**Describe the solution you'd like**
Instead of defining wormholes as objects, define them as wormholes. Wormhole nodes of a system could have attributes that an object could not, while having the same properties that objects have.
```
system <name>
	...
	wormhole "Alpha Wormhole"
		sprite <path>
		distance <x>
		period <x>
		[offset <x>]
		<wormhole properties>
```

**Describe alternatives you've considered**
Instead of having a different root name, perhaps we could create a `wormhole` child of `object`, similar to a `weapon` child of `outfit`, that defines additional wormhole behavior.
```
system <name>
	...
	object "Alpha Wormhole"
		sprite <path>
		distance <x>
		period <x>
		[offset <x>]
		wormhole
			<wormhole properties>
```

**Additional context**
This issue is more about making wormholes a distinct class separate from objects than what exactly we would do with that. 

One suggestion was to have an `exit` keyword that defines which other wormhole a specific wormhole leads to. This would mean that wormholes would no longer share names in order to define themselves as such, and would also allow more control over how multiple interconnected wormholes act. One could perhaps have three wormholes that all lead to a single wormhole, but that single wormhole leads to somewhere entirely different, instead of requiring that interconnected wormholes create a loop. One could also create unidirectional wormholes by omitting any such `exit` keyword. 

Another suggestion was for a `linked` keyword that defines a wormhole as showing a link on the map, as opposed to the current behavior of a wormhole needing a planet with a description in order to show a link.

These are ideas to discuss after any change in the syntax of wormholes as a whole, though, although any such discussion should come shortly after that is done since an attribute like `exit` changes the behavior of how we'd define wormholes.

ravenshining:
A clarification of my `exit` suggestion:

[...]

Issue #5311, opened 2020-09-11
Create "locked name" and "locked outfits" flags for ships


Amazinite:
**Is your feature request related to a problem? Please describe.**
In the future when it is possible to both give and take ships from the player, it may be preferable that while the player has command over the given ship, they are unable to edit it in various ways.

**Related Issue Links**
#5302
#4399

**Describe the solution you'd like**
Create two new keywords for `ship` similar to `"never disabled"`.
* `"locked name"`: Prevents the player from renaming the ship.
* `"locked outfits"`: Prevents the player from selling outfits from or installing outfits onto the ship. (Perhaps ammo can be an exception.)

```
ship "Argosy" "Spiteful King"
	"locked name"
	"locked outfits"
	...
	attributes
		...
```

tehhowch:
Rather than introduce multiple new root-level nodes, these should be under a common node that introduces a permissions specification.
We could then implement a simple permissions interface, extended by at least `Outfit` and `Ship`, that reports things like "CanBuy," "CanSell," "CanRename," "CanOutfit," "CanDisown," etc. Classes that interact with outfits and ships, namely the shop panels and info panels, would then first look at the entity's available permissions for a given action (such as whether renaming is allowed).

e.g.
```
ship "Argosy" "Spiteful King"
  ...
  permissions
    deny all
    outfit
```
Would not let you do any "permissioned" actions, except outfit the ship.
```
permissions
  not outfit
```
Would be the inverse, where you can do anything you want with the ship except customize its outfits.
(We would need to specially handle whether restocking ammo counts as customizing outfits, as the simplest implementation would disallow it.)

 An empty or missing node would be the same as the current state, where the player is allowed full control over the item.

petervdmeer:
The keyword `lock` sounds better to me than `permissions` and I'm not in favor of a 2-stage approach where you first deny everything and then start allowing some things. Just a simple 1-level approach where you specify all that is locked sounds easier to me as content creator.

So then it would look like:
```
lock
	name
	outfits
	sale
	disown
	unflag
	government "Syndicate"
```

Locking outfits would prevent adding or removing any outfits, while we might also want to only lock some outfits to the ship. That would look like:
```
lock
	outfit
		"Super Heavy Laser Cannon" 2
		"High Power Low Energy Engine" 1
```
What we could do for ammo is that you can lock it (as outfit) to prevent ammo from being sold, but to allow expended and additional ammo to be bought.

[...]

PR #6132, opened 2021-07-21
feat(wormholes): Make wormholes first class citizens

quyykk:
**Feature:** This PR implements the feature request detailed and discussed in issue #5457.

## Feature Details
This PR makes wormholes first class citizens!

- Wormholes are defined using a `wormhole` (see below).
- Wormholes no longer abuse some planet properties for themselves

## Usage Example

Here's how to define a wormhole:

```
wormhole <name>
        "display name" <name>
        mappable
        link <from> <to>
```

- `<name>` the name of the wormhole.
- `display name` the display name for this wormhole ("???" by default).
- `mappable` whether the arrow on the map is shown (for example, Deneb's wormhole doesn't have this)
- `link`: Specifies that a wormhole in `<from>` will transport you to `<to>` (you can have any number of this).

To make a planet a wormhole simply use the `"wormhole"` attribute like so:

```
planet <name>
    wormhole <wormhole>
```

# Implementation Details

Created a new `Wormhole` class to do Wormhole things instead of putting it all in `Planet`.

## Testing Done
Here's all what I've done:

- Load old save file to confirm that any wormholes already visited are preserved.
- Gone through the Ember Waste wormholes to confirm that wormholes still work correctly.
- Made several save steps on a clean save to confirm that wormholes are saved correctly.
- Went through the Deneb wormhole to make sure that still works.
- Landed on Ringworlds to make sure that didn't break.

## Performance Impact

Negligible.


Zitchas:
This sounds very useful. It would be very helpful to not have wormholes and planets conflated anymore.

Terin:
Could this support giving wormholes a display name? Something like The Eye in the Wanderer campaign being marked as "???" when it is _extremely_ obvious that it's The Eye is a lot. It'd be cool if others could be updated as you learned more about their area, like the Ember Threshold or Wormhole Alpha, but The Eye is the big one.

quyykk:
> Could this support giving wormholes a display name?

Done. That property is called `"pretty name"` because well the wormhole technically already has a name and the display name is just for displaying. Just add `"pretty name" "The Eye"` to the eye wormholes if you want them to be named :)

Zitchas:
Can this "pretty name" be set using events? It would be great to have a wormhole appear as the normal "??" initially, but then have an event change/add its name later on as situations change.

On a similar tangeant, althought not relevant to this particular issue, I've often wanted to have *planets* labeled with ?? until the player lands on them or other criteria are met.

quyykk:
Yeah you can change every property.

Terin:
Looks good! Would be very handy for naming Ember Wastes wormholes throughout the story there, yeah.

>and the display name is just for displaying.

What do you mean by this, though? You've got `name`, which is used in data, then `display name` for how it looks in-game. That's the syntax used with governments, as well, unless there's a third place with a name I'm not considering.

Also seems like it's not compiling on my end, but that could just be an issue with my setup. Hopefully we can get workflows approved for you before too long.

source/Cast.h
```diff
@@ -0,0 +1,113 @@
...
+
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+#ifndef CAST_H_
+#define CAST_H_
+
+#include <cassert>
```
petervdmeer, line 16: Casting should be a rare operation. If we need this class then we are casting too much.

source/CoreStartData.cpp
```diff
@@ -47,7 +47,7 @@ void CoreStartData::Save(DataWriter &out) const
 	out.BeginChild();
 	{
 		out.Write("system", system->Name());
-		out.Write("planet", planet->TrueName());
+		out.Write("planet", planet->Name());
```
petervdmeer, line 50: TrueName() was defined as the name to be used in the datafiles, why change to Name() here?

quyykk:
> That's the syntax used with governments, as well

Oh I didn't know. I'll rename it then.

> Also seems like it's not compiling on my end

Are you on macOS? I don't know how to add the additional files I created to the project that's why it fails :(.

source/DistanceMap.cpp
```diff
@@ -182,7 +184,7 @@ void DistanceMap::Init(const Ship *ship)
 		{
 			bool hasWormhole = false;
 			for(const StellarObject &object : ship->GetSystem()->Objects())
-				if(object.HasSprite() && object.HasValidPlanet() && object.GetPlanet()->IsWormhole())
+				if(object.HasSprite() && object.HasValidLandable() && isa<Wormhole>(object.GetLandable()))
```
petervdmeer, line 187: I feel that Wormholes are not properly separated off in this PR. Maybe the amount of casts can be reduced by adding an HasValidWormhole() to object (similar to HasValidPlanet())?

petervdmeer (changes requested):
I didn't review the full PR, but I do already have some review comments.

source/CoreStartData.cpp
quyykk: Cause it was unnecessary. A planet's name is the name used in the datafiles. I could change it back, it doesn't really matter since `TrueName` just calls `Name`.

source/DistanceMap.cpp
quyykk: I guess I could! Then I can also get rid of the `assert`s I believe if I introduce `HasValidWormhole` and `GetWormhole` and then the same for planets. Thanks!

source/Cast.h
quyykk: I don't actually call the function that asserts all that often (maybe like 5 overall). But as stated below, I'll remove them :)

Terin:
>Are you on macOS? I don't know how to add the additional files I created to the project that's why it fails :(.

Nope, Windows 10 in that case, although Linux seems to work just fine.

petervdmeer:
> Are you on macOS? I don't know how to add the additional files I created to the project that's why it fails :(.

There is a CI-job that will generate the updated XCode-project for you (once you get approval for the CI jobs). Just look at the error-message and download the updated XCode-project file from there after the CI jobs have run.

Azure3141:
Would it be possible to add a way to specify the color of the wormhole link? e.g. red wormhole links for the ember waste wormholes, and purple ones for the Pug created ones.

quyykk:
Sure I can do that, but I think it's a bit out of scope for this PR. Maybe create a feature request first to see what others think?

Azure3141:
I think it would fit well in this PR, since you're already overhauling the wormhole system. Adding a coloring feature separately could mean it would have to be redone entirely if this gets merged, and I don't think the feature is too out of scope since you've already added in bonus features like hiding bidirectional arrows without exploring a wormhole both ways.

quyykk:
> Adding a coloring feature separately could mean it would have to be redone entirely if this gets merged

That change is like 10 lines code. If this gets merged it's trivial to rewrite it. Really.

> added in bonus features like hiding bidirectional arrows without exploring a wormhole both ways

You're right; it's hypocritical of me. I'll just say this: Recoloring parts of the map isn't possible right now (making things appear differently on the map) AFAIK, which means that what you're proposing is novel. The 2 things I added (renaming and the bidirectional arrow thingy) are not novel, at least not to the same extent as yours IMO. sorry

source/PlayerInfo.cpp
```diff
@@ -157,7 +157,24 @@ void PlayerInfo::Load(const string &path)
 		else if(child.Token(0) == "travel" && child.Size() >= 2)
 			travelPlan.push_back(GameData::Systems().Get(child.Token(1)));
 		else if(child.Token(0) == "travel destination" && child.Size() >= 2)
+		{
```
quyykk, line 160: This statement is duplicated. I'd rather put refactor this into a function, but I don't know where to put such a function.

data/tests/tests_wormhole_navigation.txt
```diff
@@ -232,9 +233,14 @@ test-data "Ember Wastes Navigation"
 		visited Edusa
 		visited Arculus
 		visited Aescolanus
-		"visited planet" "Ember Threshold"
-		"visited planet" "Ember Wormhole"
-		"visited planet" "Remnant Wormhole"
+		"visited wormhole" "Ember Threshold (Terminus)"
+		"visited womrhole" "Ember Threshold (Cardea)"
```
Terin, line 237:
```suggestion
		"visited wormhole" "Ember Threshold (Cardea)"
```
Should fix the unit test error

source/Wormhole.h
```diff
@@ -0,0 +1,73 @@
+/* Planet.h
```
prophile, line 1: Copypasta

data/tests/tests_wormhole_navigation.txt
quyykk: <3

tehhowch (changes requested):
My biggest issue is with the datafile syntax change. It's not acceptable.

There wasn't much discussion in the other issues about different solutions. I feel we could very easily keep the existing syntax and achieve the desired result (detailed control over how the game handles wormholes) by introducing an optional `wormhole` node.

In this optional node, the user would be able to set the display name, the system traversal order, whether or not a wormhole should be displayed on the map, and so on. This node is optional in that the game would process all planets after files have been loaded and generate defaulted wormhole entries for those that weren't provided by the user.

I also don't agree that we should create the base Landable class. I think it's perfectly fine for some "deep space" wormhole to have a defense fleet, enable the player to land "inside" of it, and so on. To me what defines a wormhole is, _you land in one system and depart in another_. There's not even a requirement that you always exit into the same system from the same entry point.

Obviously a bunch of that is just not possible at the moment, and this PR isn't the place to just go implement all of it. I do think it reasonable to address the display name,  traversal order, and map arrow state, though.

source/GameData.cpp
```diff
@@ -333,9 +344,46 @@ void GameData::CheckReferences()
...
+				// Ugly workaround to support old style visited planet that were
+				// actually wormholes not planets.
+				planetsThatAreActuallyWormholesMaybe.push_back(&it);
+			}
+
+		// Loop through every system to try to discover planets that are actually wormholes.
+		for(auto &&system : systems)
+			for(auto &&planet : planetsThatAreActuallyWormholesMaybe)
+			{
+				string name = planet->first + " (" + system.first + ")";
```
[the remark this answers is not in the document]
quyykk: It's not requiring it. It's how the new wormholes are named. Should I instead hardcode the existing wormhole names? Because I really don't see what else I can do.

source/Wormhole.h
```diff
@@ -0,0 +1,73 @@
...
+
+	const Wormhole *Exit() const;
+	const System *ExitSystem() const;
+
+	void FinishLoading(const Set<Wormhole> &wormholes);
+
+	// Whether the wormhole link shows up on the map.
+	bool IsLinked() const;
+	// Whether NPCs will attempt to land on this wormhole.
+	bool AllowNPCs() const;
```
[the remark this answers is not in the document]
quyykk: I don't understand what you mean. This is already in the code (whether the "planet"/wormhole has a spaceport).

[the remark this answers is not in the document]
tehhowch: Sure, but not explicitly like you're making it here - more an artifact of how the game chooses landing sites.

Context ref: https://github.com/endless-sky/endless-sky/issues/5311#issuecomment-691045117

source/MapDetailPanel.cpp
```diff
@@ -667,15 +667,15 @@ void MapDetailPanel::DrawOrbits()
...
 	for(const StellarObject &object : selectedSystem->Objects())
 	{
 		if(object.Radius() <= 0.)
 			continue;
 		
 		Point pos = orbitCenter + object.Position() * scale;
-		if(object.HasValidPlanet() && object.GetPlanet()->IsAccessible(player.Flagship()))
-			planets[object.GetPlanet()] = pos;
+		if(object.HasValidLandable() && object.GetLandable()->IsAccessible(player.Flagship()))
+			landables[object.GetLandable()] = pos;
```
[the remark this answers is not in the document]
quyykk: Which two different collections do you mean?

source/Wormhole.h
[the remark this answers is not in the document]
quyykk: Ah so should I remove it then? Because if I do then NPCs won't land on wormholes anymore.

source/Wormhole.cpp
```diff
@@ -0,0 +1,162 @@
...
+	auto it = links.find(from);
+	return it == links.end() ? from : it->second;
+}
+
+
+
+const vector<const System *> &Wormhole::Systems() const
+{
+	return *systems;
+}
```
tehhowch, line 147: What do we need these systems for, that we cannot get from `links`? Or is keeping it for outsiders more useful than having them consume `links`?

source/MapPanel.cpp
```diff
@@ -991,14 +992,14 @@ void MapPanel::DrawWormholes()
 	
 	// Avoid iterating each StellarObject in every system by iterating over planets instead. A
 	// system can host more than one set of wormholes (e.g. Cardea), and some wormholes may even
-	// share a link vector. If a wormhole's planet has no description, no link will be drawn.
+	// share a link vector.
 	for(auto &&it : GameData::Planets())
 	{
```
tehhowch, line 997: With a root-node approach, there will be a `GameData::Wormholes` collection you would iterate through here, rather than the planets collection.

source/MapPanel.cpp
```diff
@@ -991,14 +992,14 @@ void MapPanel::DrawWormholes()
...
 		const Planet &p = it.second;
-		if(!p.IsValid() || !p.IsWormhole() || !player.HasVisited(p) || p.Description().empty())
+		if(!p.IsValid() || !p.IsWormhole() || !player.HasVisited(p) || !p.GetWormhole().IsLinked())
 			continue;
 		
-		const vector<const System *> &waypoints = p.WormholeSystems();
+		const vector<const System *> &waypoints = p.GetWormhole().Systems();
 		const System *from = waypoints.back();
 		for(const System *to : waypoints)
 		{
```
tehhowch, line 1005: This doesn't obey the links ordering.

tehhowch (commented):
To build on my comments, the load algorithm I'm envisioning is something like this:
1. Parse content as before. When a named object appears in multiple systems, it will still trip the is-a-wormhole logic.
2. Also load data from a new root-level "wormhole" token into a class object owned by a `Wormholes` collection in GameData
  1. It shall be named the same as the associated planet
    - if no valid `Planet` exists with its name, it shall be marked invalid and a diagnostic issued by CheckReferences
  2. It shall support customizations like what you've done here
  3. It shall support being modified by GameEvents, like what you've implicitly have done here by making it a child node of `planet`
  4. It shall be considered invalid if its link routes contain invalid systems.
3. After data has loaded, and after events have applied, the collection of planet objects shall be inspected for wormholes.
  1. Any wormholes that don't exist in the Wormholes collection shall be added to it:
   - The associated planet will receive a pointer to this new object.
   - "linked" will be appropriately derived from the planet's description.
   - `links` routing table shall be derived from the order of the planet's systems and create a closed loop.
  2. Any auto-added wormholes created on initial load shall trigger a diagnostic, so creators can update their content.
  3. It is optional to issue a diagnostic for wormholes auto-added via events, as these would always trigger for existing pilots.
  4. Event-based changes to wormhole planets should trigger associated changes to the modern wormhole representation, e.g. `remove description`.

So this loading this file
```
# file a
system A
  object A* # meh, a planet
    ...

system C
  object A* # look ma, a wormhole!
    ...

system B
  object A* # ooh it's a circular one
```
Would result in a wormhole collection containing a wormhole object under the name A*, and it would trigger a diagnostic since there was no `planet A*` definition.
```
planet A*
```
would remove the diagnostic, and the resulting wormhole would not appear on the map (empty description). It would have the links table
```
link A C
link C B
link B A
```

I'm sure I missed something in the above description, let me know if something is unclear.

*Thanks for working on this!*

data/map.txt
```diff
@@ -27028,21 +27028,35 @@ planet "Elek Kartanu"
 
 planet "Ember Graveyard"
 	attributes "requires: quantum keystone"
-	description ""
+	wormhole
+		linked
+		link Ossipago Polerius
+		link Polerius Ossipago
```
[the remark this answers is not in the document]
quyykk: > as that definitely isn't backwards compatible

No?

> there's no way to inject the required data into already-processed or pending GameEvents

There's no need to do that. The Eye and the Pug Wormhole work, because the wormhole needs to be added as an object to the systems in question so that you can use them. Without the event, then the planets are just unused.

source/Wormhole.cpp
```diff
@@ -0,0 +1,162 @@
...
+	links.erase(from);
+
+	auto it = links.begin();
+	using value_type = decltype(links)::value_type;
+	while(find_if(it, links.end(), [&from](const value_type &pair)
+				{
+					return pair.second == from;
+				}) != links.end())
+			it = links.erase(it);
+}
```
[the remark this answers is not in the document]
quyykk: You're right! I'll change the implementation to follow invalid links instead of removing them. Which means that in your example the round-trip will still work. If someone removes A* from B, then going into the wormhole in A will go to B which is invalid so it will go to C (cause B -> C).

source/Wormhole.cpp
quyykk: Good question. This method is indeed unnecessary.

quyykk: Turns out I do need it lol haha

quyykk:
I changed it like you said (root level wormhole and co). I still need to test event changes, but they should work (I think).

I also made it so that if you don't specify any links, it generates them automatically. What do you think?

quyykk:
I fixed every bug (I hope). I verified that wormholes are correctly autogenerated (including in events), both one-way and normal wormholes.

The only thing missing at this point is some refactoring I want to do and I believe when events are happen for the first time the wormholes don't generate (they do after you reload your save).

quyykk:
This PR is now Ready To Review<sup>TM</sup>. I found several glaring architectural mistakes that tehhowch would have been against and fixed them. The implementation is now pretty straightforward and clean. Nothing to do anymore on my end.

data/map.txt
```diff
@@ -27093,22 +27093,40 @@ planet "Elek Kartanu"
 
 planet "Ember Graveyard"
 	attributes "requires: quantum keystone"
-	description ""
+
+wormhole "Ember Graveyard"
+	linked
```
petervdmeer, line 27098: The `linked` keyword doesn't sound right to me; every wormhole is linked to somewhere (otherwise it wouldn't be a wormhole). Maybe `mappable` (as in "explored and able to be shown on a map") would be a better keyword, since it controls if the link is shown on a map or not.

data/wanderer/wanderers.txt
```diff
@@ -876,8 +876,11 @@ event "wanderers: more systems lost"
 		spaceport `The raft of algae that supports the spaceport is probably at least a dozen meters thick, but flexible enough that it bends as the ocean swells pass underneath, causing individual buildings to rise up or tilt slightly relative to their neighbors. Now that this world has been captured by the Unfettered, nearly all the buildings are empty; it appears that the Hai are not too fond of the idea of inhabiting a water world, so they are using it mostly as a military base.`
 
 planet "The Eye"
-	description ""
 	spaceport ""
```
petervdmeer, line 879: Maybe add a newline here? (Probably not needed for the parser, but it just looks a little bit nicer.)

source/Planet.cpp
```diff
@@ -231,6 +233,28 @@ void Planet::Load(const DataNode &node)
...
+	{
+		wormhole = wormholes.Get(name);
+		wormhole->LoadFromPlanet(this);
+		Files::LogError("Warning: auto generation of wormhole \"" + name
+				+ (wormhole->IsValid() ? "\"." : "\" failed."));
+	}
+
+	// If the wormhole was autogenerated we need to update it to
+	// match the planet's state.
+	if(wormhole && wormhole->IsAutogenerated())
```
petervdmeer, line 249: Why the `IsAutogenerated()` tracking and the separate if-statement? The only situation in which `IsAutoGenerated()` would be true is if LoadFromPlanet() was used to create the wormhole, but the only place where that happens is on line 242 just above here. So we can handle all the effects already above (we could even do the check if the description is empty within the LoadFromPlanet function and handle the GenerateLinks also within the LoadFromPlanet).

source/Wormhole.cpp
```diff
@@ -0,0 +1,207 @@
...
+#include "Planet.h"
+#include "System.h"
+
+#include <algorithm>
+
+using namespace std;
+
+
+
+// Load a planet's description from a file.
```
petervdmeer, line 26: Not a planet's description.

source/Wormhole.cpp
```diff
@@ -0,0 +1,207 @@
...
+				links[from] = to;
+		}
+		else if(key == "linked")
+			linked = !remove;
+		else if(key == "display name")
+		{
+			if(remove)
+				name = "???";
+			else if(hasValue)
+				name = value;
```
petervdmeer, line 83: else child.PrintTrace("Something about display name key given without value")

source/Planet.cpp
```diff
@@ -231,6 +233,28 @@ void Planet::Load(const DataNode &node)
...
+	{
+		wormhole->SetLinked(!description.empty());
```
petervdmeer, line 251: SetLinked() and GenerateLinks() are actually already handled in LoadFromPlanet().

source/Wormhole.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+
+void Wormhole::SetLinked(bool linked)
+{
+	this->linked = linked;
+}
+
+
+
+void Wormhole::GenerateLinks()
```
petervdmeer, line 194: Would it make sense to make the code in this function part of `LoadFromPlanet()`? (Or at least rename it to `"GenerateLinksFromPlanets()`).

petervdmeer (changes requested):
Sorry for my late review-feedback. The overall change looks good and I'm happy with the change being implemented here.

source/Planet.cpp
quyykk: `FinishedLoading` can be called multiple times. If a old-school wormhole (aka a planet) is updated, then we also need to update its wormhole object.

source/Planet.cpp
quyykk: Same comment as above. This is needed when a planet is updated.

source/MapPanel.cpp
```diff
@@ -1014,24 +1016,17 @@ void MapPanel::DrawWormholes()
...
-	for(auto &&it : GameData::Planets())
+	// A system can host more than one set of wormholes (e.g. Cardea), and some wormholes may even
+	// share a link vector.
+	for(auto &&it : GameData::Wormholes())
 	{
-		const Planet &p = it.second;
-		if(!p.IsValid() || !p.IsWormhole() || !player.HasVisited(p) || p.Description().empty())
+		const Planet *p = it.second.GetPlanet();
+		if(!p->IsValid() || !player.HasVisited(*p) || !it.second.IsLinked())
 			continue;
```
tehhowch, line 1025: Do we need to check wormhole validity here too?


(+reference semantics)

source/Planet.cpp
tehhowch, line 249: Should this be `else if`, or done before the one-time fallback?

tehhowch (commented):
>  I found several glaring architectural mistakes that tehhowch would have complained about and fixed them. 

Indeed there were 😅

source/GameData.cpp
```diff
@@ -332,10 +334,28 @@ void GameData::CheckReferences()
 	for(const auto &it : phrases)
 		if(it.second.Name().empty())
 			Warn("phrase", it.first);
+	// Update references to wormholes.
+	for(auto &&it : wormholes)
+		planets.Get(it.second.GetPlanet()->Name())->AssignWormhole();
```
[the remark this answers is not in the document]
quyykk: Currently the syntax to define wormhole is tightly coupled to exactly one planet (`wormhole <planet name>`). That'll be weird if we ever support more than one planet. So, I propose that we decouple the wormhole name from the planet name and add a new attribute to a planet called `wormhole <wormhole>` that binds the given wormhole to the planet.

What do you think?

source/Planet.cpp
```diff
@@ -231,6 +233,28 @@ void Planet::Load(const DataNode &node)
 
 
 
+void Planet::FinishLoading(Set<Wormhole> &wormholes)
+{
+	// If this planet is in multiple systems, then it is a wormhole.
+	if(!wormhole && systems.size() > 1)
+	{
+		wormhole = wormholes.Get(name);
+		wormhole->LoadFromPlanet(this);
```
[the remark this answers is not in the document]
quyykk: That part will only get executed once, since after the first execution `this->wormhole != nullptr`. Calling LoadFromPlanet multiple times is safe yes.

source/Wormhole.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+
+void Wormhole::LoadFromPlanet(const Planet *planet)
+{
+	this->planet = planet;
+	linked = !planet->Description().empty();
+	GenerateLinks();
+	isAutogenerated = true;
+	isDefined = true;
+}
```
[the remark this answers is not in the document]
quyykk: Only in Wormhole::IsValid should be fine I think.

source/GameData.cpp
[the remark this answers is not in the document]
quyykk: Ok I have now implemented what I wrote above. Seems like a better solution tbh and it makes it almost trivial to support the multiple planets one wormhole feature.

data/map.txt
```diff
@@ -30091,21 +30132,41 @@ planet Winter
...
+	mappable
+	link Caeculus Prosa
+	link Prosa Caeculus
 
 planet "Wormhole Link"
 	attributes "requires: quantum keystone"
-	description ``
+	wormhole "Wormhole Link"
+
+wormhole "Wormhole Link"
```
tehhowch, line 30166: We should consider demonstrating that the planet and wormhole can now be named separately. This one is a good candidate IMO.

source/Planet.cpp
```diff
@@ -231,6 +240,25 @@ void Planet::Load(const DataNode &node)
...
+// Legacy wormhole do not have an associated Wormhole object so
+// we must auto generate one if we detect such legacy wormhole.
+void Planet::FinishLoading(Set<Wormhole> &wormholes)
+{
+	// If this planet is in multiple systems, then it is a wormhole.
+	if(!wormhole && systems.size() > 1)
+	{
+		wormhole = wormholes.Get(TrueName());
+		wormhole->LoadFromPlanet(this);
+		Files::LogError("Warning: auto generation of wormhole \"" + name + "\".");
```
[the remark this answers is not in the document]
quyykk: > Let's also print this deprecation for the event case, too.

I mean that's already the case since if a wormhole is created through an event then it goes through the code above already or do you mean something else?

data/map.txt
quyykk: I honestly have no idea what I should name it. Do  you have any recommendations?

source/Planet.cpp
```diff
@@ -231,6 +240,25 @@ void Planet::Load(const DataNode &node)
...
+	{
+		wormhole = wormholes.Get(TrueName());
+		wormhole->LoadFromPlanet(*this);
+		Files::LogError("Warning: deprecated automatic generation of wormhole \"" + name + "\" from a multi-system planet.");
+	}
+	// If the wormhole was autogenerated we need to update it to
+	// match the planet's state.
+	else if(wormhole && wormhole->IsAutogenerated())
+		wormhole->LoadFromPlanet(*this);
+}
```
tehhowch, line 258: My prior comment about logging for the event case was to have logs emitted when we hit this else if branch. I'm ok with not, but knowing "when" the branch is hit could be beneficial.

tehhowch (commented):
I swear I'll actually boot this soon 😅

source/Planet.cpp
quyykk: I can provide a warning if you'd like but I'm not sure it would be useful because you already get a warning for auto-generated wormholes and if you fix that error then you won't see the update warnings anyways.

Hecter94:
@tehhowch 

@Azure3141 was looking into adding custom colours to wormhole links. Would it be appropriate to wait until this PR is merged before doing so, or it can it be done separately? 
If you'd prefer he wait, do we have an idea for when this PR might be ready to merge?

source/Wormhole.cpp
```diff
@@ -0,0 +1,212 @@
+/* Wormhole.cpp
+Copyright (c) 2021 quyykk
+
+Endless Sky is free software: you can redistribute it and/or modify it under the
+terms of the GNU General Public License as published by the Free Software
+Foundation, either version 3 of the License, or (at your option) any later version.
+
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
```
tibetiroka, line 10:
```suggestion
PARTICULAR PURPOSE. See the GNU General Public License for more details.

You should have received a copy of the GNU General Public License along with
this program. If not, see <https://www.gnu.org/licenses/>.
```
The standard copyright header was changed recently.

source/Wormhole.h
```diff
@@ -0,0 +1,70 @@
+/* Wormhole.h
+Copyright (c) 2021 quyykk
+
+Endless Sky is free software: you can redistribute it and/or modify it under the
+terms of the GNU General Public License as published by the Free Software
+Foundation, either version 3 of the License, or (at your option) any later version.
+
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
```
tibetiroka, line 10:
```suggestion
PARTICULAR PURPOSE. See the GNU General Public License for more details.

You should have received a copy of the GNU General Public License along with
this program. If not, see <https://www.gnu.org/licenses/>.
```

tibetiroka (changes requested):

petervdmeer (approved):

source/Wormhole.cpp
```diff
@@ -0,0 +1,215 @@
+/* Wormhole.cpp
+Copyright (c) 2021 quyykk
```
tibetiroka, line 2:
```suggestion
Copyright (c) 2021 by quyykk
```

source/Wormhole.h
```diff
@@ -0,0 +1,73 @@
+/* Wormhole.h
+Copyright (c) 2021 quyykk
```
tibetiroka, line 2:
```suggestion
Copyright (c) 2021 by quyykk
```

tibetiroka (changes requested):

Issue #236, opened 2015-11-22
Feature Request: Mission Sorting


Highrs:
I would like a way to filter out my missions on the job board, by things such as: Reward, passenger or cargo, nearness and weather there is a time limit.

natangold:
Dear lord, yes. Filtering and sorting would be great for either/both 'current missions' and 'available missions'. Alpha by type, alpha by <planet>, numeric by <payment>, numeric by <cargo> / <bunks>, and most importantly numeric by profit (i.e., <payment> divided by <cargo> / <bunks>.

tehhowch:
Seems like this will require an additional "wide" sprite, as one column of the sorted table will have to hold the name of the mission, and the remaining numeric columns only have to be as wide as their names. For the names, offering every numeric component of a mission for sorting isn't sensible, and creating "complex" sorting criteria such as payment/ton/jump tends strongly towards handholding.

I'd lean towards sort options of
 - Payment (could be total, e.g. accept + stopover + complete, or just complete)
 - Due In (deadline - today)
 - Distance (would be the same distance used for auto payment calculation)
 - Used Space (cargo + outfits received - outfits given)
 - Used Bunks

The wide panel could remember that it is open, and the sort used last, so that it does not reset each time the panel is opened or returned to.

For graphics, much like choosing the map color, an indicator would show which sort is being used, and the choice of indicator would indicate ascending vs descending. Simplest would be to scale the collapse indicator used in the Shipyard and outfitter for categories.

This sort process would be restricted to missions that have the JOB category - story and spaceport missions would not be sortable. This limitation seems sensible since non-job missions are often composed from many parts.

warp-core:
#6915?

quyykk:
yes

PR #6138, opened 2021-07-25
Sort available jobs alphabetically.

quyykk:
**Feature:** #236 is related

## Feature Details

<del>This is a pretty simple change: Sort the "available jobs" by payment. The jobs with the smallest payments are at the top while the job with the highest payment is at the very bottom.</del>

After feedback, this PR sorts the list alphabetically.

<hr>

<details>
<summary>Background information</summary>

Sorting jobs by payment/type/distance/... is requested lots of times (see #236). Yesterday I bought some good cargo/transport ships and did lots of jobs. It was pretty annoying to go through the whole list and find out which job paid the most.

This PR defaults jobs by payment and only that. I really don't think all the other sorts make any sense:

- By Cargo Space: cargo space is pretty much proportional to payment, so sorting by payment also sorts by cargo size.
- By Passengers: same as with cargo space.
- By Distance: You can see the distance on the map. It's way easier to get an overview that way than sorting jobs by distance.
- By Due Date: I can't think of a reason why someone would use this sort.

I don't think it's worth to introduce new UI just for all the sort options, since IMO they're going to be unused anyway.
</details>

Thoughts?

## Testing Done

Went through the jobs list and confirmed that the list is indeed sorted.

## Performance Impact

Negligible.

### DIFF
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2651,11 +2651,10 @@ void PlayerInfo::CreateMissions()
 		}
 	}
 
-	// Sort missions on the job board by payment.
+	// Sort missions on the job board alphabetically.
 	availableJobs.sort([](const Mission &lhs, const Mission &rhs)
 	{
-		return lhs.GetAction(Mission::COMPLETE).Payment()
-			< rhs.GetAction(Mission::COMPLETE).Payment();
+		return lhs.Name() < rhs.Name();
 	});
 }
 
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -2650,6 +2650,13 @@ void PlayerInfo::CreateMissions()
 				++it;
 		}
 	}
+
+	// Sort missions on the job board by payment.
+	availableJobs.sort([](const Mission &lhs, const Mission &rhs)
+	{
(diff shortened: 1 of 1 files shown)


Terin:
image

This change would be GREATLY appreciated. The jobs on the board are organized very arbitrarily on the player's end, since we have no look as the internal mission name. "Transport farmers/workers" in the image fit the alphabetical structure, but then "Transport family" is all the way in the middle of the list because it's actually "Family" internally. Sorting by payment by default would be much easier to understand on the player's end, and be VERY useful to not have to pick around for the higher paying jobs.

It's been 6 years since that issue opened, it could be more years before someone comes around with that perfect sorting system. If they do, then this code could be reverted, but in the meantime, we would get a ton out of it.

petervdmeer:
I feel that alphabetic sorting on `DisplayName` would make more sense; Alphabetic sorting looks more organized than the list here and looking through the list to find the jobs that best match my fleet is part of the fun of the game.

quyykk:
Sorting by payment is a bit controversial. I changed it to sort alphabetically, which is better than the current more or less random order.

> browsing through the list to find the jobs that best match my fleet is part of the fun of the game

Yeah I understand.

PR #6139, opened 2021-07-25
Test: Add minimal unit-test for account.cpp

petervdmeer:
**Feature:** This PR implements a minimal unit-test for Account.cpp

## Feature Details
This test was part of https://github.com/endless-sky/endless-sky/pull/5244 (and was a bit larger there), but it moved out of scope for PR 5344. I'm submitting this test as separate PR because the test itself might still be useful.

## UI Screenshots
N/A

## Usage Examples
N/A (just run the test as part of the unit tests)

## Testing Done
Automatically tested as part of CI.

## Performance Impact
N/A

### DIFF
--- a/tests/src/test_account.cpp
+++ b/tests/src/test_account.cpp
@@ -18,16 +18,21 @@ namespace { // test namespace
 
 // #region unit tests
 SCENARIO( "Creating an Account" , "[Account][Creation]" ) {
-	GIVEN( "initial values" ) {
+	GIVEN( "an account" ) {
 		Account account;
-		THEN( "storing and retrieving data should work" ) {
-			REQUIRE( account.Credits() == 0. );
+		REQUIRE( account.Credits() == 0. );
+		REQUIRE( account.TotalDebt() == 0 );
+		WHEN( "money is added" ){
 			account.AddCredits(100);
-			REQUIRE( account.Credits() == 100 );
-
-			REQUIRE( account.TotalDebt() == 0 );
+			THEN( "the balance is increased" ){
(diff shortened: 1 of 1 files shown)


tests/src/test_account.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+// #region mock data
+// #endregion mock data
+
+
+
+// #region unit tests
+SCENARIO( "Creating an Account" , "[Account][Creation]" ) {
+	GIVEN( "initial values" ) {
+		Account account = Account();
```
quyykk, line 22: `Account account;`?

tests/src/test_account.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+
+
+// #region unit tests
+SCENARIO( "Creating an Account" , "[Account][Creation]" ) {
+	GIVEN( "initial values" ) {
+		Account account;
+		THEN( "storing and retrieving data should work" ) {
+			REQUIRE( account.Credits() == 0. );
+			account.AddCredits(100);
```
tehhowch, line 25: This would align better with bdd as e.g.

```
Given an account
When money is added
Then the balance is increased
```
And
```
Given an account
When a fine is levied
Then debt is incurred
```

So the code would have a root `given` + account definition, then sibling `when`s that `require` a certain balance / debt (since their follow-up `then` won't be valid if the requires fail) and perform the modification, and lastly `then`s whose `check`s verify the balance / debt exist and are correct.

tehhowch (commented):
Ooh, tests!

PR #6141, opened 2021-07-26
Fix bug when accepting/aborting mission not paning.

quyykk:
**Bugfix:** couldn't find an issue for this

## Fix Details

When going up/down the mission list, the map updates to show where exactly the destination is. This doesn't happen when accepting/aborting a job. This PR fixes this behavior. Here's a video of the bug. Notice how the map doesn't pan and still shows the initial destination as destination for the other jobs enough though that is not the case.

https://user-images.githubusercontent.com/85879619/127031569-b3530d8d-96c0-450b-a522-9ae74b7d7ca1.mp4

## Testing Done

Tested that the map now pans when accepting/aborting.

### DIFF
--- a/source/MissionPanel.cpp
+++ b/source/MissionPanel.cpp
@@ -321,20 +321,8 @@ bool MissionPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, 
 	else
 		return MapPanel::KeyDown(key, mod, command, isNewPress);
 	
-	// To reach here, we changed the selected mission. Scroll the active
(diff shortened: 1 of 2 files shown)


petervdmeer:
I'm not sure if I consider the original behavior a bug; if I explicitly click on a mission, then I consider it very nice that the map centers on the destination.
If I "accidentally" select a mission (because that is a side-effect of aborting and/or accepting), then I consider it very nice that the UI just executes the abort by removing the mission and doesn't do anything else. (I might mentally not yet be busy with selecting the next mission.)

PR #6142, opened 2021-07-27
Add a simple unit test for DistanceMap.

quyykk:
**Feature:** This PR implements a simple unit test for `DistanceMap`.

## Feature Details

Right now it tests only for hyperdrive travel. 

**Unfortunately**: I had to link against SDL/Opengl for the test program because it won't link otherwise. It's probably caused by `System` but I don't really know what to do to avoid this.

## Testing Done

This is tested as part of the test infrastructure.

## Performance Impact
None.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -136,8 +136,8 @@ test = env.Program(
 	source=RecursiveGlob("*.cpp", testBuildDirectory) + sourceLib,
 	 # Add Catch header & additional test includes to the existing search paths
 	CPPPATH=(env.get('CPPPATH', []) + [pathjoin('tests', 'include')]),
-	# Do not link against the actual implementations of SDL, OpenGL, etc.
-	LIBS=[],
+	# Unfortunately link against the actual implementations of SDL, OpenGL, etc.
+	LIBS=[game_libs],
 	# Pass the necessary link flags for a console program.
 	LINKFLAGS=[x for x in env.get('LINKFLAGS', []) if x not in ('-mwindows',)]
 )
--- a/tests/data/systems.txt
+++ b/tests/data/systems.txt
@@ -1,0 +1,142 @@
+R"(
+system System1
+	pos 0 0
+	link System2
+	link System3
+	link System4
+
+system System2
+	pos 50 0
+	link System1
+	link System5
+	link System11
+
+system System3
+	pos 0 25
+	link System1
+	link System19
+
+system System4
+	pos -50 0
+	link System1
+	link System7
+	link System8
+
+system System5
+	pos 60 10
+	link System2
+	link System6
+	link System12
+
+system System6
+	pos 45 25
+	link System5
+	link System14
+
+system System7
+	pos -125 0
+	link System4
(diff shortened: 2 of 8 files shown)


SConstruct
```diff
@@ -136,8 +136,8 @@ test = env.Program(
 	source=RecursiveGlob("*.cpp", testBuildDirectory) + sourceLib,
 	 # Add Catch header & additional test includes to the existing search paths
 	CPPPATH=(env.get('CPPPATH', []) + [pathjoin('tests', 'include')]),
-	# Do not link against the actual implementations of SDL, OpenGL, etc.
-	LIBS=[],
+	# Unfortunately link against the actual implementations of SDL, OpenGL, etc.
+	LIBS=[game_libs],
```
tehhowch, line 140: I'm not keen to make this change.

tests/data/systems.txt
```diff
@@ -0,0 +1,142 @@
...
+	link System27
+
+system System27
+	pos 90 -50
+	link System27
+
+system System28
+	pos -15 -200
+	link System10
+)"
```
tehhowch, line 142: This is a lot of systems for a single test. It would be significantly better for a small set of systems to be used per tested scenario.

tests/src/helpers/system-factory.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include "../../../source/Planet.h"
+#include "../../../source/Set.h"
+#include "../../../source/System.h"
+
+#include <string>
+
+
+
+// Method to convert text input into consumable Systems.
+Set<System> AsSystems(std::string text)
```
tehhowch, line 26: All pointers to `Planet` objects are dangling as soon as this method returns.

tests/src/helpers/system-factory.cpp
```diff
@@ -0,0 +1,55 @@
...
+			planets.Get(node.Token(1))->Load(node);
+	}
+
+	return systems;
+}
+
+
+
+// Convert the text to a single system.
+System AsSystem(std::string text)
```
tehhowch, line 46: Same here.

tehhowch (commented):
We aren't in a place where this class can be unit tested yet.

SConstruct
quyykk: Well me neither. What do you propose?

quyykk:
> We aren't in a place where this class can be unit tested yet.

I agree. What is there to do to reach that point? do you have something in mind?

tehhowch:
We need to completely refactor the data loading strategy to remove usage of unrestricted globals. I haven't drawn up what the architecture will look like for that approach yet, but I do have an idea of how I will go about it.

A complementary approach will be to evaluate various dependency-mocking libraries and determine what it would look like to integrate with them. We have hard requirements:
1) this project does not use cmake
2) the integration must be compatible with the Code::Blocks workspace (and ideally also XCode, but our macOS support is essentially frozen and as such our current test framework doesn't exist on that platform).

PR #6148, opened 2021-08-01
Remove accidental white supremacy from White Sun

Galaucus:
Turns out that the energy generation on this is 1488, which paired with "white sun" is fairly sketch. This is a very minor tweak that makes the reactor not-Hitlery.

This changes the regen/sec from 1488 to 1485.

### DIFF
--- a/data/wanderer/wanderer outfits.txt
+++ b/data/wanderer/wanderer outfits.txt
@@ -353,7 +353,7 @@ outfit "White Sun Reactor"
 	thumbnail "outfit/white sun"
 	"mass" 127
 	"outfit space" -127
-	"energy generation" 24.8
+	"energy generation" 24.7
 	"heat generation" 32
 	"energy capacity" 6700
 	description "The White Sun is a large nuclear reactor that the Wanderers use for powering their Strong Wind warships."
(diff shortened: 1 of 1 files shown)


MasterOfGrey:
Can we buff it instead to 1502?

Galaucus:
> 
> 
> Can we buff it instead to 1502?

That would be 25.033 repeating, which is kinda' awkward.

MasterOfGrey:
> > Can we buff it instead to 1502?
> 
> That would be 25.033 repeating, which is kinda' awkward.

1501? I think better to buff slightly and get clear away from it than tweak a nerf.

petervdmeer (approved):

Zitchas (approved):
The current modified value of 24.85 results in 1491. Unlike the previous 1488 number, 1491 has no hits when searched for in conjunction with "white supremacy," and instead turns up mainly historical info related to events of 1941 and a book on the subject. 

It also turns up a native american sketch comedy troop "Their comedy sketches depict contemporary Native American life in the United States, using humor and satire to explore issues such as stereotypes and racism (internal and external), tribal politics, and the conflict between tradition and modernity." (copied from wikipedia).

In terms of breaking up the triple set of white supremacist references, I think this should be adequate (and kind of appropriate too).

PR #6152, opened 2021-08-01
Build tests in debug mode

janisozaur:
This changes CI to build tests in debug mode to ensure they build correctly.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -156,9 +156,9 @@ jobs:
           ${{ matrix.os }}-ccache-
     - name: Compile tests
       if: steps.cache-tests.outputs.cache-hit != 'true'
-      run: scons -Qj $(nproc) build-tests
+      run: scons -Qj $(nproc) mode=debug build-tests
     - name: Run tests
-      run: scons -Qj $(nproc) test
+      run: scons -Qj $(nproc) mode=debug test
     - name: Run benchmarks
       run: ./tests/endless-sky-tests [!benchmark]
 


tehhowch:
What is gained by compiling and executing tests in debug mode? That the CI jobs build and execute the test framework in release mode was intentional--consumers of the game don't play debug builds.

quyykk:
Also, this seems to have linked just fine apparently but I get undefined references to SDL. Debug mode isn't built with LTO which is why this doesn't happen for me in release.

janisozaur:
@tehhowch sorry for late response, haven't noticed your comment. The benefit it brings is that developers (e.g. yours truly) use tests in debug mode to step through the code with debugger.

@quyykk no, it didn't link because there are some oddly specific rules to what causes the tests to be built and run, and evidently modifying the tests doesn't trigger running them.

tehhowch:
To be technically accurate, this PR did not modify how tests are built, it modified the CI job itself, and the path filter step doesn't equate workflow changes with changes to the game code.

I won't accept the current change--CI _will_ execute tests against optimized code--but will accept a PR that links the tests with game libraries iff building in debug mode, with the understanding that said linking will be removed as soon as appropriate mocks are configured.

This PR would then add that conditional link, and add a new line to the CI tests job to build the "build-tests" target in the debug mode.

petervdmeer:
@janisozaur Adding a test that ensures that the debug builds (and debug test builds) remain working could be valuable, especially since the debug test-builds appear to be broken due to the SDL-linking issues that quyykk noticed in Aug 2021 (and because those issues still seem to exist in master today).

But it should be an _additional_ test, not a change to the existing CI tests (because we do really want to CI-test the builds that are used by the players of the game). The debug-test-builds also need to be fixed before we could merge an additional test for debug builds.

Please feel free to submit a new PR that incorporates the review-feedback here, but I'm closing this PR, because it is by now more than 7 months idle, it is not going to be merged in its current implementation, and because it is taking up space in the open PRs list.

PR #6155, opened 2021-08-03
feat: Use RGBA texures instead of BGRA for better portability

thomasballinger:
Alternative to https://github.com/endless-sky/endless-sky/pull/5153 at @petervdmeer's suggestion: let's just use RBGA everywhere! Unlike that PR, this does not eliminate the need for libjpeg-turbo.


tehhowch (approved):
LGTM, with a comment update needed on line 239.

PR #6159, opened 2021-08-04
Build flag to remove support for music and libmad

thomasballinger:
Reopening https://github.com/endless-sky/endless-sky/pull/5591

## Summary
Adds a build flag for building with no music, removing the need for libmad.

The gameplay effect is small: only one music track is included in the game, `sounds/ambient/machinery` which plays on space stations.

My motivation for this is [a web build of Endless Sky](https://github.com/thomasballinger/endless-web/pull/16/commits), for which I've not yet been able to get libmad working.

## Discussion

@tehhowch on this patch previously:

> This seems like a really hacky way to disable music. Wouldn't it be better to make loading libmad optional, and avoid doing so when the compiled binary receives the command line parameter e.g. `--no-music`?
> 
> Then we don't need any ifdefs at all, and systems that won't attempt to load the dynamic library won't need the library in their execution environment.
> 
> A different take on the ifdef approach here would be to ifdef the entire music.cpp and music.h files, and their associated includes / method calls. Then the compiler wouldn't ever see calls to `Music::____`, and there wouldn't be anything at all in music.cpp or music.h
> 
> Only the first approach would make it into the upstream here, but the second way could simplify the patch you need to use for the web fork. Though since you would need to modify audio initialization, perhaps not.

Dynamically loading libmad sounds like the approach to pursue here, but I'm creating this PR before I have that for discussion.

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -35,6 +35,7 @@ opts = Variables()
 opts.AddVariables(
 	EnumVariable("mode", "Compilation mode", "release", allowed_values=("release", "debug", "profile")),
 	EnumVariable("opengl", "Whether to use OpenGL or OpenGL ES", "desktop", allowed_values=("desktop", "gles")),
+	EnumVariable("music", "Whether to use music", "on", allowed_values=("on", "off")),
 	PathVariable("BUILDDIR", "Directory to store compiled object files in", "build", PathVariable.PathIsDirCreate),
 	PathVariable("BIN_DIR", "Directory to store binaries in", ".", PathVariable.PathIsDirCreate),
 	PathVariable("DESTDIR", "Destination root directory, e.g. if building a package", "", PathVariable.PathAccept),
@@ -117,13 +118,17 @@ else:
 		"GLEW",
 	])
 
-# libmad is not in the Steam runtime, so link it statically:
-if 'steamrt_scout_i386' in chroot_name:
-	env.Append(LIBS = File("/usr/lib/i386-linux-gnu/libmad.a"))
-elif 'steamrt_scout_amd64' in chroot_name:
-	env.Append(LIBS = File("/usr/lib/x86_64-linux-gnu/libmad.a"))
-else:
-	env.Append(LIBS = "mad")
+if env["music"] == "off":
+	flags += ["-DES_NO_MUSIC"]
+
+if env["music"] == "on":
+	# libmad is not in the Steam runtime, so link it statically:
+	if 'steamrt_scout_i386' in chroot_name:
+		env.Append(LIBS = File("/usr/lib/i386-linux-gnu/libmad.a"))
+	elif 'steamrt_scout_amd64' in chroot_name:
+		env.Append(LIBS = File("/usr/lib/x86_64-linux-gnu/libmad.a"))
+	else:
+		env.Append(LIBS = "mad")
 
 
 binDirectory = '' if env["BIN_DIR"] == '.' else pathjoin(env["BIN_DIR"], env["mode"])
--- a/source/Music.cpp
+++ b/source/Music.cpp
@@ -14,7 +14,9 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 
 #include "Files.h"
 
+#ifndef ES_NO_MUSIC
 #include <mad.h>
+#endif // ES_NO_MUSIC
 
 #include <algorithm>
 #include <cstring>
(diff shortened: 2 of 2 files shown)


thomasballinger:
Another approach is compiling libmad for the web, like [was done for libjpeg-turbo](https://github.com/thomasballinger/endless-web/pull/16/commits/7dcb40a44b331b037fc7241006f62a469fb385f0#diff-ae6e5d9cf228a8cbfebc566a3e997bd9dd27bf9c72ac1c9ea097a8d00f46743fR134-R141). 

So far compiling libmad for the web is giving me trouble, `emconfigure ./configure; emmake make` gets me errors. Digging into libmad will be something for later if necessary, I'll start with mocking it out as cleanly as possible. In terms of content, in the base game there's currently a single mp3 that places on space stations and no other music in the game.

image

And another approach could be mocking out the libmad interface with noops and just linking a different implementation. I'll look into this a bit.

thomasballinger:
Reading about dymamic linking, I'm gathering that
- we'll need different codepaths for Windows vs Mac/Linux (dlopen vs LoadLibrary)
- dynamic linking is a bit scary, it moves failures from build time to runtime (but we have CI that should catch problems, so maybe this is fine)
- we'll still need a header file

given this, I'm going to look more at compiling libmad for the web.

thomasballinger:
Here's a new two-part proposal:
1. command line flag for turning off music as suggested by @tehhowch above, see https://github.com/endless-sky/endless-sky/pull/6163
2. mocked out libmad.h so that music code compiles — and runs! without doing anything — in the web build. Here's a proof of concept, but I'd move this to its own header file: https://github.com/thomasballinger/endless-sky/commit/97829c19c8be8cc40bc9c8de4b7f102e509d0c2b

I don't know if step 2, a mocked out libmad header file, would ever make sense to merge but there's a clean interface here so I think step 1 doesn't depend on step 2 being a good idea.

petervdmeer:
What is the default software to use for WebAssembly when playing music?
This PR now focusses on removing music-support, but would it make sense to also look into alternative music support functions? Similar to OpenGL-embedded vs OpenGL-desktop as was done for the graphics)?

thomasballinger:
Hm great point! I may be thinking too small in an effort to make progress here.

This could be reimplemented in JavaScript if the mp3s can be made available there: playing a streaming mo3 file is part of the web platform, currently there's even a mp3 that plays during loading on a https://play-endless-web.com. A more direct port might be possible: still let SDL2 control output, but use a javascript mp3 decoding library.

This could be implemented more normally (no JS FFI) with another mp3 library. mpg123 is licensed under LGPL, compatible with Endless Sky's GPLv3. There's [a library](https://github.com/kepstin/aurora-mpg123.js) that seems to be compiling it with Emscripten, so this looks doable.

This can probably be done by compiling libmad to Emscripten! I spent a couple hours with ./configure and make in that code yesterday and ended up with something compiling that didn't quite work yet:

```
libmad-0.15.1b.tar.gz:
    wget https://netactuate.dl.sourceforge.net/project/mad/libmad/0.15.1b/libmad-0.15.1b.tar.gz
libmad-0.15.1b: libmad-0.15.1b.tar.gz
    tar xzvf libmad-0.15.1b.tar.gz
libmad-0.15.1b/asdf: libmad-0.15.1b
    cd libmad-0.15.1b; emconfigure ./configure
    # get rid of some compiler optimizations
    cd libmad-0.15.1b; sed -i '' -e 's/-fforce-mem -fforce-addr -fthread-jumps -fcse-follow-jumps -fcse-skip-blocks -fexpensive-optimizations -fregmove -fschedule-insns2 -fstrength-reduce//g' Makefile
    # less assembly plz
    cd libmad-0.15.1b; sed -i '' -e 's/FPM = -DFPM_INTEL/FPM = -DFPM_DEFAULT/g' Makefile
    # dunno what this is
    cd libmad-0.15.1b; sed -i '' -e 's/allow_undefined_flag="-flat_namespace -undefined suppress"/allow_undefined_flag=""/g' libtool
    cd libmad-0.15.1b; emmake make
```
(the weird see syntax is a mac thing, remove `'' -e` for gnu sed)

The problem here is mostly that I don't know what I'm doing! It seems likely libmad can be compiled by emscripten by tweaking some settings, I've just been stuck on it.

---
I bet the right approach is compiling libmad, but I've been stuck on this for a bit.

thomasballinger:
Here's that fake mad header described in step 2. above, to make this more concrete: https://github.com/endless-sky/endless-sky/pull/6164/files#diff-5106b806bbafcfe9906574bc8364912094b17baa3e0f4529cd97731ee152a012

thomasballinger:
The purpose of this PR was to minimize the changes necessary for an Emscripten build, so since Emscripten support is [unlikely to be merged](https://github.com/endless-sky/endless-sky/pull/6164) this isn't needed.

PR #6160, opened 2021-08-04
feat: Add offset and facing angle to submunitions

### LINKED ISSUE #3878: Feature request: hardpoint offset (or similar) for submunitions
A while back on the server, I and a few others discussed the idea of hardpoint offset for submunitions. (It turns out it doesn't actually work when you try it.)

We decided to [bring this up on the forum](https://groups.google.com/forum/#!category-topic/endless-sky/feature-requests/KVGzPyYNCfg), although apparently it didn't receive much attention. I think it's worthy enough of mention here.

The hypothetical use could range to something like this:
```
outfit "Warpstone Launcher"
	category "Secondary Weapons"
	cost 160000
	thumbnail "outfit/torpedo launcher"
	"mass" 13
	"outfit space" -25
	"weapon capacity" -25
	"gun ports" -1
	"warpstone capacity" 30
	weapon
		sprite "projectile/torpedo"
			"frame rate" 2
			"random start frame"
		sound "torpedo"
		ammo "Warpstone Torpedo"
		icon "icon/torpedo"
		"fire effect" "torpedo fire"
		"hit effect" "large explosion"
		"die effect" "blaster impact"
		"inaccuracy" 5
		"velocity" 6
		"lifetime" 600
		"reload" 120
		"firing energy" 2
		"firing heat" 45
		"acceleration" .15
		"drag" .025
		"turn" 1.5
		"homing" 3
		"optical tracking" .8
		"missile strength" 30
		"split range" 120
		"submunition" "Warpstone Submunition"
	description "Warpstone torpedoes momentarily alter the fabric of the universe in order to teleport behind (or into) their intended targets when they come into range, utilizing the briefly stressed form of the local space as protection against anti-missile fire."

outfit "Warpstone Submunition"
	weapon
		sprite "projectile/torpedo"
			"frame rate" 2
			"random start frame"
		"hardpoint offset" 360
		sound "blaster"
		"fire effect" "blaster impact"
		"hit effect" "large explosion"
		"die effect" "tiny explosion"
		"velocity" 6
		"lifetime" 120
		"acceleration" .15
		"drag" .025
		"turn" 1.5
		"homing" 3
		"optical tracking" .8
		"shield damage" 450
		"hull damage" 450
		"hit force" 300
		"missile strength" 70
```
Unfortunately, all this does in the current game is make a more "direct" finisher.

### LINKED ISSUE #5376: Feature Request: Submunitions at an angle.
I would like to create a weapon where submunitions fire at a precise angle from the direction/path of the parent projectile.
Submunitions currently by default fire at exactly the same direction/path as the parent.

This feature should allow for new types of weapons that steer for intermediate locations besides their intended target, that split very precisely or even weapons that travel around their target and then hit from the other side.

**Related Issue Links**
- [hardpoint offset](https://github.com/endless-sky/endless-sky/issues/3878) mentions a submunition fired at 180 degrees (back at a target that was passed).
- [range override](https://github.com/endless-sky/endless-sky/issues/4478) seems related because submunitions fired at an angle will have a reduced range compared to submunitions travelling in a direct line.
- I had expected to find many other feature requests requesting a feature like this, but I didn't find others.

I would like to see the keyword `facing` to be usable on submunitions (or some better solution that accomplishes this). And an `offset <x> <y>` keyword added to set an offset relative to the origin munition (to support #3878 as discussed below). An example of the keyword being used:
```
outfit "Left Splitting Mirv"
	weapon
		"submunition" "Mirv Fragment" 1
			facing 266
			offset -10 -6
		"submunition" "Mirv Fragment" 1
			facing 268
			offset -10 -3
		"submunition" "Mirv Fragment" 1
			facing 270
			offset -10 0
		"submunition" "Mirv Fragment" 1
			facing 272
			offset -10 3
		"submunition" "Mirv Fragment" 1
			facing 274
			offset -10 6
```

The current alternative used for most submunitions where an effect like this is desired is to give an `inaccuracy` on the submunition itself, but that doesn't give the precission that I like to have.

**Additional context**
I hope the description is already sufficiently clear; please let me know if I should add some more details.

petervdmeer:
**Feature:** This PR implements the feature request detailed and discussed in issues #5376 and #3878

## Feature Details
Closes #5376
Closes #3878
This PR adds an offset and an angle for the submunitions as requested in the 2 linked issues. The newly introduced keywords in this PR are `facing` (to select the angle for the submunition) and `offset` (to set the offset compared to the explosion location) which both are to be used under the `submunition` keyword in the datafiles that describe submunitions.

Chosen implementation is to set the facing and offset keywords in the submunition sections on the "parent-projectile" to allow the same "submunition" to be used on different angles and offsets from a parent projectile (without each angle or offset requiring its own definition).

## UI Screenshots
Attached is [mirv.txt](https://github.com/endless-sky/endless-sky/files/6933484/mirv.txt), a datafile that can be used for testing,. Install the datafile and then just buy the new "Left Splitting Mirv" weapon and some sidewinders at any basic outfitter. Here is a screenshot of the "Left Splitting Mirv" weapon in action:
triple_left

## Usage Examples
Here is an example (taken from #5376):
```
outfit "Left Splitting Mirv"
	weapon
		"submunition" "Mirv Fragment" 1
			facing 268
			offset -10 -3
		"submunition" "Mirv Fragment" 1
			facing 270
			offset -10 0
		"submunition" "Mirv Fragment" 1
			facing 272
			offset -10 3
```

## Testing Done
I've tested with the provided datafile and I fired a number of other weapons.

## Performance Impact
Dying projectiles now need to deal with 2 additional variables when creating submunitions; and Angle and a Point. I don't expect any noticeable impact from this.

### DIFF
--- a/source/Projectile.cpp
+++ b/source/Projectile.cpp
@@ -62,7 +62,7 @@ Projectile::Projectile(const Ship &parent, Point position, Angle angle, const We
 
 
 Projectile::Projectile(const Projectile &parent, const Point &offset, const Angle &angle, const Weapon *weapon)
-	: Body(weapon->WeaponSprite(), parent.position + parent.velocity + offset, parent.velocity, parent.angle + angle),
+	: Body(weapon->WeaponSprite(), parent.position + parent.velocity + parent.angle.Rotate(offset), parent.velocity, parent.angle + angle),
 	weapon(weapon), targetShip(parent.targetShip), lifetime(weapon->Lifetime())
 {
 	government = parent.government;
@@ -90,7 +90,7 @@ Projectile::Projectile(const Projectile &parent, const Point &offset, const Angl
 
 
 // Ship explosion.
-Projectile::Projectile(const Point &position, const Weapon *weapon)
+Projectile::Projectile(Point position, const Weapon *weapon)
 	: weapon(weapon)
 {
 	this->position = position;
@@ -112,7 +112,7 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 					visuals.emplace_back(*it.first, position, velocity, angle);
 			
 			for(const auto &it : weapon->Submunitions())
-				for(int i = 0; i < it.count; ++i)
+				for(size_t i = 0; i < it.count; ++i)
 					projectiles.emplace_back(*this, it.offset, it.facing, it.weapon);
 		}
 		MarkForRemoval();
--- a/source/Projectile.cpp
+++ b/source/Projectile.cpp
@@ -93,7 +93,7 @@ Projectile::Projectile(const Projectile &parent, const Point &offset, const Angl
 Projectile::Projectile(Point position, const Weapon *weapon)
 	: weapon(weapon)
 {
-	this->position = position;
+	this->position = std::move(position);
 }
 
 
--- a/source/Projectile.cpp
+++ b/source/Projectile.cpp
@@ -61,8 +61,8 @@ Projectile::Projectile(const Ship &parent, Point position, Angle angle, const We
 
 
 
-Projectile::Projectile(const Projectile &parent, const Weapon *weapon)
-	: Body(weapon->WeaponSprite(), parent.position + parent.velocity, parent.velocity, parent.angle),
+Projectile::Projectile(const Projectile &parent, Point offset, Angle angle, const Weapon *weapon)
+	: Body(weapon->WeaponSprite(), parent.position + parent.velocity + offset, parent.velocity, parent.angle + angle),
 	weapon(weapon), targetShip(parent.targetShip), lifetime(weapon->Lifetime())
 {
 	government = parent.government;
@@ -112,8 +112,8 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 					visuals.emplace_back(*it.first, position, velocity, angle);
 			
 			for(const auto &it : weapon->Submunitions())
-				for(int i = 0; i < it.second; ++i)
-					projectiles.emplace_back(*this, it.first);
+				for(int i = 0; i < it.count; ++i)
+					projectiles.emplace_back(*this, it.offset, it.facing, it.outfit);
 		}
 		MarkForRemoval();
 		return;
--- a/source/Projectile.cpp
+++ b/source/Projectile.cpp
@@ -35,7 +35,7 @@ namespace {
 
 
 
-Projectile::Projectile(const Ship &parent, const Point &position, const Angle &angle, const Weapon *weapon)
+Projectile::Projectile(const Ship &parent, Point position, Angle angle, const Weapon *weapon)
 	: Body(weapon->WeaponSprite(), position, parent.Velocity(), angle),
 	weapon(weapon), targetShip(parent.GetTargetShip()), lifetime(weapon->Lifetime())
 {
--- a/source/Projectile.cpp
+++ b/source/Projectile.cpp
@@ -35,7 +35,7 @@ namespace {
 
 
 
-Projectile::Projectile(const Ship &parent, Point position, Angle angle, const Weapon *weapon)
+Projectile::Projectile(const Ship &parent, const Point &position, const Angle &angle, const Weapon *weapon)
 	: Body(weapon->WeaponSprite(), position, parent.Velocity(), angle),
 	weapon(weapon), targetShip(parent.GetTargetShip()), lifetime(weapon->Lifetime())
 {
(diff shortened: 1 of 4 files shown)


tehhowch (commented):
Looks good! Doesn't prevent us from adding inaccuracy or other things we can specify for the initial projectile's creation.

Some minor comments

thomasballinger:
https://user-images.githubusercontent.com/458879/128304336-eaec54a3-cfd4-4e4f-80e9-80fc4b5771b3.mov

source/Weapon.h
```diff
@@ -33,6 +35,18 @@ class Sprite;
 // copy of them, and storing them as class variables instead of in a map of
 // string to double significantly reduces access time.
 class Weapon {
+public:
+	struct Submunition{
+		const Outfit *outfit;
```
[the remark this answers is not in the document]
petervdmeer: > You may also consider not allowing a default constructor, and initializing the pointer and count as const values, but that can sometimes not play nicely with the standard library, so if it doesn't compile that's fine.

std::vector seems to be okay with it; as long as the struct has a move constructor and a move assignment operator.

But the usage of Outfit for (base)attributes in the copy-able Ship class seems to give the requirement that Outfit (and its parent class Weapon with its member variable submunitions) need to be copyable.

source/Weapon.h
```diff
@@ -33,6 +35,22 @@ class Sprite;
...
+		explicit Submunition(const Weapon *weapon, std::size_t count) noexcept
+			: weapon(weapon), count(count) {};
+		
+		const Weapon *weapon;
+		// The direction and offset at which the submunition will be fired from
+		// the original projectile. The direction and offset are relative to the
+		// point and direction at which the original projectile dies.
+		Angle facing;
+		Point offset;
+		int count;
```
tehhowch, line 50: Initialization is needed for the default constructor, plus comments scoped to the respective variable:
```suggestion
		const Weapon *weapon = nullptr;
		std::size_t count = 0;
		// The angular offset from the source projectile, relative to its current facing.
		Angle facing;
		// The base offset from the source projectile's position, relative to its current facing.
		Point offset;
```

source/Projectile.cpp
```diff
@@ -61,8 +61,8 @@ Projectile::Projectile(const Ship &parent, Point position, Angle angle, const We
 
 
 
-Projectile::Projectile(const Projectile &parent, const Weapon *weapon)
-	: Body(weapon->WeaponSprite(), parent.position + parent.velocity, parent.velocity, parent.angle),
+Projectile::Projectile(const Projectile &parent, const Point &offset, const Angle &angle, const Weapon *weapon)
+	: Body(weapon->WeaponSprite(), parent.position + parent.velocity + offset, parent.velocity, parent.angle + angle),
```
tehhowch, line 65: The new offset is used as an absolute offset, without considering the orientation of the projectile. Is that intended?
If not, it could be addressed here or at the sites where this constructor gets called.

source/Projectile.cpp
```diff
@@ -35,7 +35,7 @@ namespace {
 
 
 
-Projectile::Projectile(const Ship &parent, Point position, Angle angle, const Weapon *weapon)
+Projectile::Projectile(const Ship &parent, const Point &position, const Angle &angle, const Weapon *weapon)
```
tehhowch, line 38: The underlying Body constructor takes its point and angle parameters by value. While we don't quite forward these by-value parameters, a follow-up (profiled) PR can consider changing this constructor.
```suggestion
Projectile::Projectile(const Ship &parent, Point position, Angle angle, const Weapon *weapon)
```

source/Projectile.cpp
```diff
@@ -90,7 +90,7 @@ Projectile::Projectile(const Projectile &parent, const Weapon *weapon)
 
 
 // Ship explosion.
-Projectile::Projectile(Point position, const Weapon *weapon)
+Projectile::Projectile(const Point &position, const Weapon *weapon)
 	: weapon(weapon)
 {
 	this->position = position;
```
tehhowch, line 96: The "better" change would be to `std::move` the by-value parameter here, to use move assignment rather than copy assignment.

source/Projectile.cpp
petervdmeer: No, that was not intended; the orientation of the projectile should be considered. I'll fix that.

source/Projectile.cpp
petervdmeer: Doesn't the compiler automatically use the move-assignment when it detects that the given parameter will never be used again? (If that is the case, then I'll just put back the original code since that is more readable.)

tehhowch: It is forbidden to automatically move from an "lvalue", which is effectively any variable with a local name.
The compiler can optimize to elide _copies_ (e.g. return value optimization) in a variety of cases, but assignment isn't one of them.

So when one has a local variable and basically the whole point is to put it into something else (and never use that local variable again _without first assigning back to it_), we can signal that move ops are allowable via `std::move`. There can be scenarios where this inhibits optimizations (like RVO), and in many of them a compiler warning is generated.

petervdmeer: Thanks, very clear! Then adding `std::move` is indeed a good change; I added it.

PR #6163, opened 2021-08-05
--no-music flag to disable music

thomasballinger:
**Feature:** Command line flag to disable music, an alternative to https://github.com/endless-sky/endless-sky/pull/6159.

Motivation: I'd like to be able to turn music off to avoid needing to run these threads in the web build and to avoid needing to implement mp3 decoding (because I'm having trouble compiling libmad on that platform). 

## Feature Details
`--no-music` command line flag prevents music threads from ever being started, short circuits cleanup code, and prevents music from being playing.

## UI Screenshots
N/A

## Usage Examples
`$ /Users/tomb/Library/Developer/Xcode/DerivedData/EndlessSky-asymldgrudnqsdclwufttpzdiyop/Build/Products/Debug/Endless\ Sky.app/Contents/MacOS/Endless\ Sky --no-music`

## Testing Done
Visited Kraken Station with and without the flag, station music track did not play with the flag and did play without it.

## Performance Impact
I didn't notice anything, although this probably uses less CPU. Not strictly relevant but related to performance: I was able to bump down the threadpool size in the web build when using this flag.

### DIFF
--- a/source/Audio.cpp
+++ b/source/Audio.cpp
@@ -114,10 +114,13 @@ namespace {
 }
 
 
+// always set in Init anyway 
+bool Audio::musicEnabled = false;
 
 // Begin loading sounds (in a separate thread).
-void Audio::Init(const vector<string> &sources)
+void Audio::Init(const vector<string> &sources, bool enableMusic)
 {
+	musicEnabled = enableMusic;
 	device = alcOpenDevice(nullptr);
 	if(!device)
 		return;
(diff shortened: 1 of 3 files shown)


thomasballinger:
The purpose of this PR was to make running an Emscripten build a little more efficient. Since Emscripten support is [unlikely to be merged](https://github.com/endless-sky/endless-sky/pull/6164) this isn't needed.

If there's ever a player request for something like this it could still be useful, but the player-friendly way to implement this is probably something that can be adjusted without restarting the process.

PR #6164, opened 2021-08-05
emscripten build

thomasballinger:
**Feature:** Emscripten build for web browsers

This diff is much smaller than [previous attempts](https://github.com/endless-sky/endless-sky/pull/5595) but there's still plenty to discuss, including whether this is a feature that makes sense to merge at all, regardless of how simple the PR is.

There are at least two remaining hacks (I'm fine with them, but should call attention to them in case others aren't):
* loading sound files synchronously instead of in a thread
* no-op libmad replacement header — https://github.com/endless-sky/endless-sky/pull/6163 could be used to avoid running this code, but without that PR the music code runs and does nothing

If this were a good fit it could be followed up by [a bit more work](https://github.com/thomasballinger/endless-web/pull/17/commits) to make a nicer-looking web build.

## UI Screenshots
image

## Usage Examples
See [developer-readme](https://github.com/endless-sky/endless-sky/commit/e76d42d36f9a23c9a9080db24cbe8643c228ad2e) for full instructions, the gist is to install emsdk and run `scons -j 8 mode=emcc opengl=gles` then `emrun --serve_after_close --browser firefox endless-sky.html`

## Testing Done
Still compiles and runs when run with XCode on a mac.

## Performance Impact
No changes to the normal build, everything is compile-time.

The performance of the web build is significantly slower than the normal build, particularly loading. Performance is similar to https://play-endless-web.com/.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,18 @@ scons-local*
 .sconsign.dblite
 /*.d
 
+# emscripten build generated files
+endless-sky.html
+endless-sky.data
+endless-sky.js
+endless-sky.worker.js
+endless-sky.wasm
+endless-sky.wasm.map
+
+# libjpeg-turbo
+2.1.0.tar.gz
+libjpeg-turbo-2.1.0
+
 # Test Reports
 test-report.xml
 
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,90 @@
+EMSCRIPTEN_ENV := $(shell command -v emmake 2> /dev/null)
+
+clean:
+	rm -f endless-sky.html
+	rm -f endless-sky.js
+	rm -f endless-sky.worker.js
+	rm -f endless-sky.data
+	rm -f endless-sky.wasm
+	rm -f endless-sky.wasm.map
+	rm -f lib/emcc/libendless-sky.a
+2.1.0.tar.gz:
+	wget https://github.com/libjpeg-turbo/libjpeg-turbo/archive/refs/tags/2.1.0.tar.gz
+libjpeg-turbo-2.1.0: 2.1.0.tar.gz
+	tar xzvf 2.1.0.tar.gz
+# | means libjpeg-turbo-2.1.0 is a "order-only prerequisite" so creating the file doesn't invalidate the dir
+libjpeg-turbo-2.1.0/libturbojpeg.a: | libjpeg-turbo-2.1.0
+ifndef EMSCRIPTEN_ENV
+	$(error "emmake is not available, activate the emscripten env first")
+endif
+	cd libjpeg-turbo-2.1.0; emcmake cmake -G"Unix Makefiles" -DWITH_SIMD=0 -DCMAKE_BUILD_TYPE=Release -Wno-dev
+	cd libjpeg-turbo-2.1.0; emmake make
+dev: endless-sky.html
+	emrun --serve_after_close --serve_after_exit --browser chrome --private_browsing endless-sky.html
+
+COMMON_FLAGS = -O3 -flto\
+		-s USE_SDL=2\
+		-s USE_LIBPNG=1\
+		-pthread\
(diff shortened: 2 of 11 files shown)


thomasballinger:
The test failure is the `FakeMad.h` not being in Xcode, I'll wait to hear what folks think of this approach before fixing.

tehhowch:
I don't think adding a new compiler toolchain makes sense for this repo, especially one that is both less performant and less lacking functionality.

A user able to choose either compiler has little reason to choose the emcc build over gcc. Anyone able to deploy the web build elsewhere to play it remotely should be sufficiently skilled to add your fork as a remote and track your branch.

To simplify the SConstruct maintenance you need to do, you may benefit from creating / using a toolchain (which would conceivably handle a number of the flags you set. Structuring to use multiple SConscipt may benefit.

thomasballinger:
Making a separate SConstruct file sounds good, I'll give that that shot.

Regarding a new compiler toolchain in the repo making sense, I'm obviously biased but here's my take:

Emscripten is useful for sharing code with creator tools on the web, code like the file parser and loading logic. When I [reviewed @petervdmeer's formation proposal](https://observablehq.com/@ballingt-personal/review-of-formation-flying-rfc-for-endless-sky) last year I had to reimplement the parsing and formation code in JavaScript, but with Emscripten I could reuse the implementation.

It would be useful to have Emscripten config in this repo instead of in a fork so we can have one-click preview builds in pull requests and for checking whether a given commit breaks the web build (although this needn't block merges). I'd be all for drawing the line at the HTML and not having an official web build, but letting Emscripten config and CI jobs for preview builds and regression testing of Emscripten to live in this repo would be nice.

thomasballinger:
If this isn't a good fit now, I'd hope to check in again about it in the future. I think tooling like showing what the galaxy looks like given a set of plugins or visualizing the probabilities of various ships being in a system or interactive positioning of hardpoints or a GUI mission builder would be made easier to create and more reliable by making sure cpp source files build with Emscripten.

This is separate from producing the artifact of a fancy web build, which I'm also interested in upstreaming as a long term solution to running Endless Sky on unsupported platforms but is more explicitly a separate project and extends beyond the Endless Sky cpp sources.

Urls that specify a build, a save file, and a set of plugins would be nice for debugging but that infrastructure doesn't make sense for this repo; but it does depend on commits being buildable with Emscripten, something that is easiest ensure with the Emscripten build process in this repo.

tehhowch:
That's the thing--I do not want to limit what can be done with Endless Sky to that which can also be done in Emscripten + web. Upstreaming the toolchain & building it on PR would do exactly that.

Note that enabling other tools to build from ES's code is part of why things compile to a library now. Long-term development plans will see the static library partner with one or more dynamic libraries. The obvious first choice is a library that handles converting basic in-memory representations to and from the ES data format. (It doesn't make sense to make a dynamic library that hands out game objects _as needed by the game_, since other tools will have different ideas about what to do with those objects.)

thomasballinger:
OK that's compelling, adding a CI task / building on PR would direct future development in ways you don't want to direct it; the web build failing when someone adds a dependency or changes the threading model or adds VR support would make it harder to do those things, and if the build were made "official" then it would make some of them impossible! I wouldn't want the build to be official or to block other development, but it's true that I don't just want them debugging and tooling purposes — I also want to know when the web build is broken. I can understand how a failing CI task can influence PRs and how just the presence of the files in the repo not only implies support and requires maintenance, but also alters the project's purpose.

I might bring up Emscripten again once these planned dynamic libraries that split out tooling exist, since I'd want those to be compiled for the web. It's great to hear about plans for those.

If there's ever interest in getting full web builds into the repo, let me know and I'd love to help get them in!

thomasballinger:
Reopening to update with a separate build toolchain

thomasballinger:
Hopefully this gets out of sync more slowly, thanks for the suggestion @tehhowch

Issue #5936, opened 2021-05-05
Swizzle changes don't apply to carried ships


beccabunny:
**Describe the bug**
If an event changes the swizzle of the player, after departing the swizzle of normal ships is modified accordingly but not for carried ships (drones and fighters) until those are launched

**To Reproduce**
1. Buy some ships capable of carrying fighters/drones and said fighters/drones
2. Use an event to change your swizzle
3. Depart

**Expected behavior**
Carried ships should change swizzle

**Screenshots**
![Screenshot (619)](https://user-images.githubusercontent.com/45330165/117139141-30d24980-adac-11eb-868e-a99c8a1bfca8.png)
After departing having changed the swizzle from 5 to 0
![Screenshot (620)](https://user-images.githubusercontent.com/45330165/117139143-32037680-adac-11eb-914e-7ae50f126f93.png)
After launching the fighters

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: ESLauncher2
 - Version: lastest nightly

movingpictures:
For the ... unwashed and uninformed .... whatdaheck is a swizzle?

tehhowch:
Possibly related to #1462 or #2295

I'm not positive on the order of events you're having take place; mind throwing a repro save together?

beccabunny:
here you go [tommaso becca.txt](https://github.com/endless-sky/endless-sky/files/6434814/tommaso.becca.txt)

steps I followed to make it:
-started with swizzle 0 since I already had that
-brought 1 roost and 4 finches
-used omnis' mission for simplicity to change my swizzle to 4, which has this code:

```
mission "swizzle 4"
	name `Swizzle- 4`
	description `Changes ships appearance to have a color swizzle of R->B->G->R. E.g. Syndicate; purple.`
	job
	repeat
	source
		government "cheat"
	on accept
		event "swizzle 4"

event "swizzle 4"
	government "Escort"
		swizzle 4```
what happens is:
-departing the roost changes its swizzle to 4 but the fighters are still at 0
-releasing fighters now they too have swizzle 4

vitalchip:
I think this is probably because swizzle gets set during Ship::Place() which wiould usually be when departing the planet but for carried ship it'll be during Ship::Launch()

Could swizzle be set during Ship::FlightCheck() ?

tehhowch:
[...]

PR #6168, opened 2021-08-07
Don't update carry swizzle when undocking/launching.

quyykk:
**Bugfix:** This PR addresses the issue mentioned in  #5936 

## Fix Details

When a carry gets placed into the system, we do not update its swizzle. The carry's swizzle is only updated when the carrier launches, i.e. when departing from a planet.

# Testing Done

Can't test this since you can't deploy captured fighters, but this fix is fairly straightforward - there shouldn't be any bugs.


tehhowch:
There's nothing that says a ship which can be carried _must_ be carried upon departing a planet. In fact, the game was recently modified to enable this exact scenario (you depart alongside ships for which you have no bays, rather than selling them off).

petervdmeer:
I understood that the `Ship::Place(...)` function also gets used for placing NPCs, which includes NPCs that can have carried ships. (The "ships spawning for the first time" as mentioned in [tehhowch comment on 5936](https://github.com/endless-sky/endless-sky/issues/5936#issuecomment-894531371) ). Do those carried ships of those NPCs still have the correct swizzle?

And for deploying captured ships; the only limitation that I know against deploying is low hull on carried ships and I think that that can be overcome by having hull-repair on your carriers.

Did you have a look if it is possible to write some unit-tests (with the catch framework) that test if swizzles are set correctly for various ships in various situations?

tehhowch:
> Did you have a look if it is possible to write some unit-tests (with the catch framework) that test if swizzles are set correctly for various ships in various situations?

That isn't possible at the moment; the only way to get non-default ship objects is to invoke `Ship::Load`, which will pull in all globals and dependencies.

Issue #6156, opened 2021-08-03
Accessing alien space after FW campaign through jump drive feels like getting spoilers


werner291:
**Is your feature request related to a problem? Please describe.**

After completing the FW campaign (A ton of fun! Thanks a lot, it provided me with hours of fun!), my ship is left with the Jump Drive on board.

Quite happy with my new equipment (It kinda feels like "Hey, you finished the game, so just go explore!"), I start jumping to otherwise inaccessible systems, getting to Hai space, or the Ember wastes, or Coalition space... only to discover that at least for Hai and Ember Wastes, somebody clearly put effort into writing ways to get there without the Jump drive, through Keystone Gems, or discovering that wormhole.

The whole "hyperspace links" idea is so useful for controlling the progression of the player's exploration, and the jump drive basically circumvents that by letting players basically go wherever they want in a more-or-less straight line.

**Related Issue Links**

I didn't find similar issues.

**Describe the solution you'd like**

Eh... Good question, sounds like whatever you do might upset the game balance post-drive-acquisition. Perhaps, if there was some alien territory where somebody put effort into writing quests to access them, accessing them through jump drives could somehow be harder?

Perhaps one way could be that the player needs "coordinates" or something of the system they want to jump to, which may be acquired by either visiting it through their own means, or by getting these through a quest. This way, the jump drive still lets them hop, skip and jump wherever they want in known space, but story writers still get to constrain the player's movements.

**Describe alternatives you've considered**

The Quarg seem to really enjoy boxing species into their own little corner of the galaxy. Maybe have them somehow oppose the player just using the Jump drive willy-nilly? Could be a source of conflict, maybe present them more clearly as the antagonists that other species (and humans, apparently) perceive them as.

DeciduousBeard:
Regarding your suggestion that you could require coordinates for a system before accessing it with a jump drive:
1) There are places that can be accessed in no other way but by a jump drive. It is unrealistic to have a whole storyline of missions just to find out the coordinates of every one of these systems.
2) The whole idea of a jump drive is that it allows you to explore further.
3) Jump drives are still limited, as they only work within a certain range. You can't just jump anywhere. A hyperdrive lets you go places with a hyperspace link, a jump drive lets you go places within a certain distance of your existing system.
4) Most critically: Changing the functionality of a jump drive would break many existing storylines.

[...]

PR #6170, opened 2021-08-08
Add a case in "Hai: First Contact" for players who don't enter Hai space

Anarchist2:
## Summary
The discussion in #6156 reminded me of how, during my first playthrough, I entered Hai space using a JD and was thoroughly confused as to why Eliot was able to claim that he was born in human space despite being in an area that was 'clearly' only accessible through a jump drive, and I didn't figure out the wormhole even existed until much later. This PR ensures that the player is aware of the wormhole's existence regardless of entry method, and also adds a possible incentive to land on the Unfettered worlds.

### DIFF
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -39,6 +39,12 @@ mission "First Contact: Hai"
 			
 			label next
 			`	You glance quickly at the human merchant to see his expression, wondering if these aliens can indeed be so benign, but his expression is calm and untroubled. "It's true," he says, "look around you. These buildings have stood for thousands of years. Hai society is so advanced, they need almost no resources to maintain it."`
+			branch wormhole
+				has "Discovered Ultima Thule: failed"
+			choice
+				`	"How did humanity reach this place? Did they use jump drives?"`
+			`	The human merchant raises an eyebrow. "A jump drive? Can't say I'm familiar with what that is. In case you aren't in the know, though, there's a wormhole in Hai space that connects back to a secluded part of human space. Don't bother trying to figure out why and how it's there, because not even the Hai know."`
+			label wormhole
 			choice
 				`	"Why do you allow humans to settle here?"`
 				`	"What do you ask for in return for letting humans settle here?"`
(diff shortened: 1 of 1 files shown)


roadrunner56 (approved):

data/hai/hai missions.txt
```diff
@@ -65,8 +71,13 @@ mission "First Contact: Hai"
 			`	"So I can travel anywhere I want in your territory?" you ask the Hai.`
 			
 			label travel
+			branch jump
+				not "Discovered Ultima Thule: failed"
 			`	"Yes," it says, "but be careful. The north is the territory of some renegade Hai who are misled in their ways: bandits and pirates, living off what they steal from the rest of us. If you travel among them, you will not be safe."`
 				goto end
+			label jump
+			`	"Yes," it says, "but be careful. The north is the territory of some renegade Hai who are misled in their ways: bandits and pirates, living off what they steal from the rest of us. And if what you say about having a jump drive is true, I would be doubly wary. I have heard rumors that the 'Unfettered Hai' are soliciting captains who land on their worlds to give them jump drives."`
```
Pointedstick, line 79: The fact that the old Hai only tells you this if you arrived by jump drive doesn't make sense because he has no way of knowing that this would be relevant information to you because he doesn't know that you arrived via jump drive.

I think it would make more sense if he always tells you about the Unfettered soliciting jump drives, and if you don't have one yourself yet, you can ask him "What's a jump drive?" or whatever.

data/hai/hai missions.txt
```diff
@@ -93,6 +104,14 @@ event "label hai space"
 	galaxy "label hai"
 		sprite "label/hai"
 
+mission "Discovered Ultima Thule"
+	landing
+	invisible
+	to complete
+		never
+	on enter "Ultima Thule"
+		fail
```
tehhowch, line 113: Why not set a custom condition in the `on enter`? Then you don't have to deal with what looks like a negative case and treat that as a positive.

Pointedstick (approved):
LGTM now. @tehhowch?

data/hai/hai missions.txt
```diff
@@ -93,6 +100,14 @@ event "label hai space"
...
 		sprite "label/hai"
 
+mission "Discovered Wormhole Alpha"
+	landing
+	invisible
+	to complete
+		never
+	on enter "Ultima Thule"
+		set "wormhole alpha found"
+
```
tehhowch, line 110: We'll still want the mission to go away after the condition gets set 😉
```suggestion
	on enter "Ultima Thule"
		set "wormhole alpha found"
		fail

```

tehhowch (approved):
This mission is ripe for removal when we support querying visited planets in conditions ( #3735 ). Other than that, looks good.

Issue #5832, opened 2021-04-08
Haze is always appears as blue until loading has finished


beccabunny:
**Describe the bug**
Although haze on the main menu is supposed to be dependant of the system you are in (e.g. red in the ember waste), it always appears as the default blue haze until loading of the game has finished.

**To Reproduce**
1. Have a save file with your pilot landed somewhere in the ember waste or graveyard
2. Open the game
3. See as the haze in the menu appears as blue until ES finishes loading

**Expected behavior**
Best solution would be having it appear as red from the beginning, but that is obviously not possible. Maybe show no haze until it is loaded?

**System (please complete the following information):**
 - OS: Windows 10
 - Game Source: Github
 - Version: Current continous

MCOfficer:
Not showing the haze would arguably be worse. It would be nice to fade between the blue and red backdrop once the loading is done.

Regardless, this is a nit at best. For convenience, here's a pilot stationed on Nenia.
[asd adf.txt](https://github.com/endless-sky/endless-sky/files/6278824/asd.adf.txt)

PR #6174, opened 2021-08-09
Add crossfade effect when haze is changing

pakyinwww:
**Bugfix:** This PR addresses issue #5832 

## Fix Details
Current Flow
The haze changes sharply without any fading effect.

New flow
1. While the haze changes, draw both the new haze(cloak=1) and old haze(cloak=0).
2. The cloak values of new haze keeps decreasing while that of old haze keeps increasing.
3. While the cloak value of new haze reaches 0, stop drawing the old haze.

The speed of fade effect can be adjusted by FADE_PER_FRAME in StarField.cpp

## Testing Done
Start the game - fade in 
Loading a gamesave which uses different haze - crossfade

## Save File
This save is provided by @MCOfficer in the issue #5832 . It can be used to test the crossfade effect.
[asd.adf.txt](https://github.com/endless-sky/endless-sky/files/6956730/asd.adf.txt)


source/StarField.cpp
```diff
@@ -26,6 +26,8 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
 #include <algorithm>
 #include <cmath>
 #include <numeric>
+#include <iostream>
+#include <string>
```
MCOfficer, line 30: At first glance, these don't appear to be used.

source/StarField.h
```diff
@@ -13,11 +13,14 @@ PARTICULAR PURPOSE.  See the GNU General Public License for more details.
...
 #define STAR_FIELD_H_
 
+#include "DrawList.h"
 #include "Shader.h"
+#include "Sprite.h"
 
 #include "gl_header.h"
 
 #include <vector>
+#include <string>
```
MCOfficer, line 23: At first glance, this doesn't appear to be used.

source/StarField.cpp
```diff
@@ -153,12 +169,27 @@ void StarField::Draw(const Point &pos, const Point &vel, double zoom) const
...
+		hazeCloak -= FADE_PER_FRAME;
+	else
+		hazeCloak = 0.;
 	
 	// Any object within this range must be drawn. Some haze sprites may repeat
 	// more than once if the view covers a very large area.
 	Point size = Point(1., 1.) * haze.front().Radius();
 	Point topLeft = pos + (Screen::TopLeft() - size) / zoom;
 	Point bottomRight = pos + (Screen::BottomRight() + size) / zoom;
+	DrawHaze(haze,topLeft,bottomRight,hazeCloak,drawList);
```
MCOfficer, line 183:
```suggestion
	DrawHaze(haze, topLeft, bottomRight, hazeCloak, drawList);
```

source/StarField.cpp
```diff
@@ -153,12 +169,27 @@ void StarField::Draw(const Point &pos, const Point &vel, double zoom) const
...
+	if(hazeCloak > 0.)
+		DrawHaze(prevHaze,topLeft,bottomRight, 1 - hazeCloak,drawList);
```
MCOfficer, line 185:
```suggestion
		DrawHaze(prevHaze, topLeft, bottomRight, 1 - hazeCloak, drawList);
```

MCOfficer (changes requested):
A few things that caught my eye - not proficient in C++ so might be wrong on the `include`s.

pakyinwww:
Thank you for your correction. That's my first time to create pull request.

MCOfficer:
> 
> 
> Thank you for your correction. This is my first time to create pull request for the opensource project.

In that case - welcome! Allow me to share a few tips ;) 

[You can resolve conversations such as my comments above](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/about-pull-request-reviews#resolving-conversations) to indicate they're dealt with. I'm not sure why I as reviewer can't do that, but it is what it is.

And if somebody proposes changes, you have the option of [accepting them directly in the web interface](https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request). That's a matter of preference though.

For now you'll have to wait until one of our actual code monkeys comes around to check this PR, and we'll see what they have to say about it.

source/StarField.cpp
```diff
@@ -82,6 +87,19 @@ void StarField::SetHaze(const Sprite *sprite)
 	if(!sprite)
 		sprite = SpriteSet::Get("_menu/haze");
 	
+	if(isFading){
+		return;
+	}
```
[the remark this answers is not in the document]
pakyinwww: At first I used isFading to delay the sprite change because additional call during fading may cause sharp change of the haze. (Perhaps it rarely happens as the fade effect is short in time.)

But I checked again that, as you mentioned, isFading only blocks the additional call and doesn't delay the sprite change. So I will remove the isFading first.

pakyinwww:
Thank you so much for the review.

source/StarField.cpp
```diff
@@ -153,27 +166,43 @@ void StarField::Draw(const Point &pos, const Point &vel, double zoom) const
...
-	for(const Body &it : haze)
+	DrawHaze(haze, topLeft, bottomRight, transparency, drawList);
+	if(transparency > 0.)
+		DrawHaze(prevHaze, topLeft, bottomRight, 1 - transparency, drawList);
+
+	drawList.Draw();
+}
+
+
+void StarField::DrawHaze(const std::vector<Body> targetHaze, const Point targetTopLeft, const Point targetBottomRight, const double targetTransparency, DrawList& targetDrawList) const
```
tehhowch, line 190: Note that without the `&` to declare the parameter is a reference, this still creates a copy of the input vectors. Also note that it isn't helpful to rename all parameters--doing so causes "churn"--just those parameters that are _class member variables_, like `haze`. The other parameters here were local variables created in the calling function: there's no chance the parameter can be renamed without also renaming the existing usages.

```suggestion
void StarField::DrawHaze(const std::vector<Body> &targetHaze, const Point &topLeft, const Point &bottomRight, double transparency, DrawList &drawList) const
```

You could actually go further, and make this function into a lambda or a function in the anonymous namespace, because it doesn't reference any class-specific information--only what is given to it.

tehhowch:
One issue that this PR has is that the haze is changed not only when the player launches the game, but every time the player changes systems, whether that is by hyperdrive, jump drive, or wormhole.

If the 'hyperdrive flash' setting is enabled, the flash timing is pretty similar to the haze fade and the fade isn't really noticeable, especially when performing a hyperdrive jump.
The haze transition can definitely be noticed for jump drive and wormhole transitions. Do we want to keep the instantaneous haze swap for all system transitions? My gut says yes, but I'm open to feedback.

source/StarField.cpp
```diff
@@ -153,12 +171,30 @@ void StarField::Draw(const Point &pos, const Point &vel, double zoom) const
...
 	// Any object within this range must be drawn. Some haze sprites may repeat
 	// more than once if the view covers a very large area.
 	Point size = Point(1., 1.) * haze.front().Radius();
 	Point topLeft = pos + (Screen::TopLeft() - size) / zoom;
 	Point bottomRight = pos + (Screen::BottomRight() + size) / zoom;
+	DrawHaze(haze, topLeft, bottomRight, hazeCloak, drawList);
+	if(hazeCloak > 0.)
+		DrawHaze(prevHaze, topLeft, bottomRight, 1 - hazeCloak, drawList);
+
+	drawList.Draw();
```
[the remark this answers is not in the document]
pakyinwww: I tried to use BatchDrawList but I am sorry that I was not able to do it. I am still studying how BatchDrawList actually works.

tehhowch (approved):
Will let others weigh in on the haze fade when the player changes systems via in-game travel before merging.

source/StarField.cpp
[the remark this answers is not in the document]
tehhowch: Yes, that's fine. I briefly looked at it and the batch shader is not configured to support sprite transparency / cloaking.

Given the very limited amount of time haze fades in and out relative to the time haze is just being drawn at full opacity, a future PR could certainly optimize the base case to use the batch shader, or to use the engine draw stack's existing list rather than a new one.

PR #6175, opened 2021-08-10
fix(ui): Capitalize words after whitespace in cargo strings

### LINKED ISSUE #6173: Mission cargo on ship info panel can be improperly capitalized
**Describe the bug**
image
Mission cargo names that have a non-alphabet character in the middle of letters (like an apostrophe) will cause the subsequent letter to be improperly capitalized.

**To Reproduce**
1. Accept a mission with cargo that has an apostrophe or non-alphabet character in it, or modify a save to change the name of one.
2. Open ship info panel, observe cargo.

**Expected behavior**
Characters after at least apostrophes will not be capitalized. Maybe only characters after spaces will?

 - OS: Windows 10 Pro
 - Game Source: Self-compiled
 - Version: 34fbe23

**Link to save file**
[asdf asdf.txt](https://github.com/endless-sky/endless-sky/files/6952435/asdf.asdf.txt)

**Additional context**
Line 300 in `text/Format.cpp` is the culprit, as it sets the bool that causes the next character to be capitalized if the current is not an alphabet letter. Not putting up a PR since it's late, and my first instinct is to only set it if the current character is a space. Chances are someone else knows how that might break something, and/or can come up with a much more elegant solution.

Terin:
**Bugfix:** This PR fixes #6173.

## Fix Details
This PR causes Format::Capitalize() to only set the bool to capitalize a character when it matches a space, rather than any alphabet letter. The only non-letter characters in commodities.txt are hyphens, apostrophes, and the number 3 for 3D printers. This fixes any issues with apostrophes.

image

By setting it to not capitalize after hyphens, there is a slight issue in terms of fully-hyphenated words, where we get High-performance Automobiles instead of High-Performance Automobiles. However, this leaves us with more proper De-ionizers or By-products, which seems preferable to the other hyphenated words.

image

## Testing Done
Accepting missions with non-alphabet characters, and observing proper capitalization on them.

## Save File
[asdf asdf.txt](https://github.com/endless-sky/endless-sky/files/6952435/asdf.asdf.txt)

### DIFF
--- a/source/text/Format.cpp
+++ b/source/text/Format.cpp
@@ -298,7 +298,7 @@ string Format::Capitalize(const string &str)
 	bool first = true;
 	for(char &c : result)
 	{
-		if(!isalpha(c))
+		if(c == ' ')
 			first = true;
 		else
 		{
--- a/source/text/Format.cpp
+++ b/source/text/Format.cpp
@@ -298,7 +298,7 @@ string Format::Capitalize(const string &str)
 	bool first = true;
 	for(char &c : result)
 	{
-		if(c == ' ')
+		if(isspace(c))
 			first = true;
 		else
 		{
--- a/tests/src/text/test_format.cpp
+++ b/tests/src/text/test_format.cpp
@@ -93,6 +93,29 @@ SCENARIO("A player-entered quantity can be parsed to a number", "[Format][Parse]
 	}
 }
 
+SCENARIO("A string is capitalized", "[Format][Capitalize]") {
+	GIVEN( "The string 'magnesium'" ) {
+		THEN( "capitalizes to 'Magnesium'" ) {
+			CHECK( Format::Capitalize("magnesium") == "Magnesium" );
+		}
+	}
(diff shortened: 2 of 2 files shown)


source/text/Format.cpp
```diff
@@ -298,7 +298,7 @@ string Format::Capitalize(const string &str)
 	bool first = true;
 	for(char &c : result)
 	{
-		if(!isalpha(c))
+		if(c == ' ')
```
tehhowch, line 301: Rather than just _space_, I would argue for things like tab or newline, too.
```suggestion
		if(isspace(c))
```
(We should really also cast `c` to `unsigned char` before using these functions, to avoid undefined behavior, but that's something to be addressed more fully by @oo13 in #4123.)

oo13: FYI
I pushed the fix (6fbd9743a292c50668401bb9028b9f41137246b2) in #4123.
It's picked from my translatable branch.

tehhowch:
Unit tests?

PR #5781, opened 2021-03-12
content: Add some human missions about skydiving/spacediving.

petervdmeer:
**Content (Artwork / Missions / Jobs)**

## Summary
Add some new missions about skydivers. They might look the same at first glance, but there are important differences between the jobs.

## Save File
Still to be provided (it is also possible to test this by removing the random number conditions and just accept the jobs one by one as they appear in human space).

## PR Checklist
N/A

PR #6181, opened 2021-08-10
fix(ui,content): Shorten spacediving job descriptions to avoid overflow in the Missions panel

### LINKED ISSUE #6179: Spacediving description can overflow
image

Need to find a way to cut this down.

petervdmeer:

## Summary
Shortened the texts around skydiving as suggested in #6179, because the textbox was overflowing with the texts currently in the game.

Fixes #6179

## Save File
N/A

## PR Checklist
 N/A

### DIFF
--- a/data/human/transport missions.txt
+++ b/data/human/transport missions.txt
@@ -3216,7 +3216,7 @@ mission "Spacediving professionals"
 	job
 	repeat
 	name "Spacediving at <planet>"
-	description "Head to <destination> with <bunks> professionally equipped skydivers and drop them from the edge of space.  They've agreed to pay <payment> when they land."
+	description "Head to <destination> with <bunks> professionally equipped skydivers and drop them from the edge of space. They've agreed to pay <payment> when they land."
 	passengers 5 10
 	source
 		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Independent"
@@ -3238,7 +3238,7 @@ mission "Spacediving professionals accident"
 	job
 	repeat
 	name "Spacediving at <planet>"
-	description "Head to <destination> with <bunks> professionally equipped skydivers and drop them from the edge of space.  They've agreed to pay 120000 credits when they land."
+	description "Head to <destination> with <bunks> professionally equipped skydivers and drop them from the edge of space. They've agreed to pay 120000 credits when they land."
 	passengers 5 10
 	source
 		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Independent"
@@ -3260,7 +3260,7 @@ mission "Spacediving professionals disaster"
 	job
 	repeat
 	name "Spacediving at <planet>"
-	description "Head to <destination> with <bunks> professionally equipped skydivers and drop them from the edge of space.  They've agreed to pay 120000 credits when they land."
+	description "Head to <destination> with <bunks> professionally equipped skydivers and drop them from the edge of space. They've agreed to pay 120000 credits when they land."
 	passengers 5 10
 	source
 		government "Republic" "Free Worlds" "Syndicate" "Neutral" "Independent"
--- a/data/human/transport missions.txt
+++ b/data/human/transport missions.txt
@@ -3216,7 +3216,7 @@ mission "Spacediving professionals"
 	job
 	repeat
 	name "Spacediving at <planet>"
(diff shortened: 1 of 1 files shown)


data/human/transport missions.txt
```diff
@@ -3238,7 +3238,7 @@ mission "Spacediving professionals accident"
 	job
 	repeat
 	name "Spacediving at <planet>"
-	description "Head to <destination> with <bunks> skydivers, then drop them from the edge of space as you land. They are wearing thermal spacesuits, durable parachutes, and oxygen bottles to ensure their survival. Extreme sports like skydiving typically carry more risks than regular passenger transports. Payment is 120000."
+	description "Head to <destination> with <bunks> skydivers and drop them from the edge of space. They look professional, but extreme sports like skydiving typically carry more risks than regular passenger transports. Payment is 120000."
```
Terin, line 3241:
```suggestion
	description "Head to <destination> with <bunks> skydivers and drop them from the edge of space. They look professional, but extreme sports like skydiving typically carry more risks than regular passenger transports. Payment is 120000 credits."
```

data/human/transport missions.txt
```diff
@@ -3260,7 +3260,7 @@ mission "Spacediving professionals disaster"
...
```
Terin, line 3263:
```suggestion
	description "Head to <destination> with <bunks> skydivers and drop them from the edge of space. They look professional, but extreme sports like skydiving typically carry more risks than regular passenger transports. Payment is 120000 credits."
```

tehhowch (changes requested):
This phrase isn't quite right: "extreme sports like skydiving typically carry more risks than regular passenger transports."
Why would skydivers care about the risk of transporting passengers? They _are_ passengers, and skydiving doesn't involve the same things as transporting passengers.

Is the intent to signal that _the player_ takes on more risk by _participating in/facilitating_ extreme sports, rather than just transporting these same passengers from spaceport A to spaceport B?

petervdmeer:
> Is the intent to signal that _the player_ takes on more risk by _participating in/facilitating_ extreme sports, rather than just transporting these same passengers from spaceport A to spaceport B?

Yes, that is the intent.

data/human/transport missions.txt
```diff
@@ -3216,7 +3216,7 @@ mission "Spacediving professionals"
 	job
 	repeat
 	name "Spacediving at <planet>"
-	description "Head to <destination> with <bunks> skydivers, then drop them from the edge of space as you land. They are wearing thermal spacesuits, durable parachutes, and oxygen bottles to ensure their survival. Extreme sports like skydiving typically carry more risks than regular passenger transports. Payment is <payment>."
+	description "Head to <destination> with <bunks> skydivers and drop them from the edge of space. They look like professionals. Extreme sports jobs typically carry more risks than regular transport jobs. Payment is <payment>."
```
tehhowch, line 3219: Maybe something like this? It suggests, without overtly stating as much, that one might not get paid in full.
```suggestion
	description "Head to <destination> with <bunks>{{qualifier}} skydivers and drop them from the edge of space. They've agreed to pay <payment> when they land."
```
Qualifiers could be "professionally equipped," "professional-looking," "amateurish", "poorly equipped," or none at all, to provide concise textual flavor.

I think it's okay to not state that there is a risk of being "fined" by these missions, too.

petervdmeer: The texts here have been discussed quite a bit in the [original PR](https://github.com/endless-sky/endless-sky/pull/5781) and your suggestion on the risk seems to match the outcome of that discussion quite well.

The qualifiers do have a bit more meaning than just the "textual flavor"; the idea is that you can estimate the risk a bit better if you read the descriptions carefully. So if we use the qualifier here, then there should be separate qualifiers for the pros and for the amateurs.

data/human/transport missions.txt
```diff
@@ -3216,7 +3216,7 @@ mission "Spacediving professionals"
 	job
 	repeat
 	name "Spacediving at <planet>"
-	description "Head to <destination> with <bunks> skydivers and drop them from the edge of space. They look like professionals. Extreme sports jobs typically carry more risks than regular transport jobs. Payment is <payment>."
+	description "Head to <destination> with <bunks> professionally equipped skydivers and drop them from the edge of space.  They've agreed to pay <payment> when they land."
```
tehhowch, line 3219: Was the two-spaces sentence separation intended?

petervdmeer: No, that was a search/replace artifact! I just fixed them. :-)

Issue #6182, opened 2021-08-11
Better target prioritization on anti-missiles


wjbotham:
**Is your feature request related to a problem? Please describe.**
I noticed while flying recklessly through Mereti minefields in Mesuket that my anti-missile turrets target projectiles seemingly arbitrarily. Instead of targeting nearby mines which are an immediate threat, they spend some of their time targeting mines at the edge of their range. It seems that they should have a more intelligent algorithm for prioritizing targets that are a more immediate threat to their ship.

This video demonstrates the behavior:
https://user-images.githubusercontent.com/547660/128970027-4f4154f0-1a08-4c93-a786-41bf693e9e8b.mp4

If I understand the code correctly, anti-missile turrets currently prioritize targeting the *oldest* projectile in range, which probably works in many situations but is less effective if e.g. two missiles miss a target and loop around, reversing their order by the time they make their second fly-by.

**Related Issue Links**
I wasn't able to find any discussion of anti-missile target prioritization on Discord or GitHub.

**Describe the solution you'd like**
The naive solution is for anti-missile turrets to preferentially target the closest missiles.

I implemented this solution ([see here](https://github.com/wjbotham/endless-sky/pull/1])) to get a video demonstrating this behavior:
https://user-images.githubusercontent.com/547660/128969988-5f2a442a-07eb-4188-8b57-99f0ade7d6e0.mp4

**Describe alternatives you've considered**
More complicated solutions might check a missile's relative velocity to determine how immediate of a threat it is (for example, a missile that has narrowly missed and must loop around).

**Additional context**
This would be a minor nerf of the Korath Minelayer and perhaps some other missiles that are fired in extremely large numbers to overwhelm or distract anti-missiles.

PR #6183, opened 2021-08-11
Make anti-missile turrets shoot nearer projectiles first

wjbotham:
## Feature
This PR changes anti-missile targeting behavior so that anti-missile turrets always target the closest enemy missile. This is an attempt to address https://github.com/endless-sky/endless-sky/issues/6182 .

The current prioritization, which I believe implicitly targeted the oldest projectile first, looks like this:
https://user-images.githubusercontent.com/547660/128966896-044a3f09-9087-4b2d-b521-41bded67f971.mp4

The prioritization implemented here looks like this (note how the closest mines are targeted first):
https://user-images.githubusercontent.com/547660/128966907-1f5afb45-dff1-42ab-bdf4-e53e45fce147.mp4

## Feature Details
To do this, I changed the collision detection function to merely record which missiles are valid antimissile targets and used a subsequent function call to actually make the antimissile attacks.

## Testing Done
I've played the old and new code and observed ships (both player and NPC) with multiple anti-missile turrets reacting to a variety of missiles to verify that anti-missile turrets still function and that no crashes occur.

## Save File
This save file can be used to test the behavior:
[AntiMissilePriorityTest.txt](https://github.com/endless-sky/endless-sky/files/6965755/AntiMissilePriorityTest.txt)

The player's ship in this save file has extra hit points, multiple anti-missile turrets, and a cloaking device. You can either cloak and watch how Sestor anti-missiles interact with Mereti mines, or you can fly through the minefields yourself to see how it looks from the player's perspective.

## Performance Impact
Keeping projectiles sorted by distance introduces performance concerns; addressing those concerns is still in progress.


petervdmeer:
> <...> Each time the function runs, it needs to create a new list of projectiles sorted by each projectile's distance from the ship.
> 
> In order to avoid sorting the actual global list of projectiles and potentially creating side effects, I used a list of pointers to create a sorting order on top of the projectiles. There may be a better solution. This solution certainly _looks_ ugly to me.

I'm afraid that the proposed implementation here would significantly slow down the game because of the large amount of additional memory-write operations to build the sorted vectors of all projectiles per anti-missle ship.

wjbotham:
Closing this while I re-implement it and make style/whitespace fixes.

quyykk:
You don't need to close this if you want to make changes.

wjbotham:
@quyykk I'm not sure how long it'll take to have the PR ready for feedback again and didn't want to clutter the PR list in the meantime. Does this mean "we prefer not to close PRs in this situation" or just "we don't require you to close PRs in this situation"?

quyykk:
I'd leave it open because it shows that you're still working on it. If I see a closed PR, I assume that it has been abandoned in some way. I say leave it open :)

wjbotham:
Gotcha, thanks for the heads up!

petervdmeer:
> I'm not sure how long it'll take to have the PR ready for feedback again and didn't want to clutter the PR list in the meantime. 

If you choose to keep it open in that case, then please press the "convert to Draft" button. The Draft status seems to match quite well with work in progress that is not yet finished.

source/Engine.cpp
```diff
@@ -2020,19 +2023,36 @@ void Engine::DoCollisions(Projectile &projectile)
...
+		// If the projectile did not hit anything, check for anti-missile
+		// systems in range.
 		for(Ship *ship : hasAntiMissile)
 			if(ship == projectile.Target() || gov->IsEnemy(ship->GetGovernment()))
-				if(ship->FireAntiMissile(projectile, visuals))
+				if(ship->IsInAntiMissileRangeOf(projectile))
 				{
-					projectile.Kill();
-					break;
+					antiMissileTargets[ship].push_back(&projectile);
```
tehhowch, line 2032: You can greatly improve the performance of the sort comparator used in `DoAntiMissile` by computing the projectile-ship distance once and storing a `std::pair<Projectile *, double>:
```suggestion
					antiMissileTargets[ship].emplace_back(&projectile, projectile.Position().DistanceSquared(ship.Position()));
```

If you invert the order of the pair, you can use a default-comparator std::set instead of a vector, and get sort-on-insertion due to `std::less`'s behavior comparing `std::pair` by the first and then the second element.

Probably it will be faster to keep the structure as `map<ship *, vector<pair<projectile *, double>>>` and explicitly sort after all insertions, rather than use a collection that sorts on insertion, like `map<ship *, set<pair<double, projectile *>>>`, but one should profile to be certain of that.

source/Engine.cpp
```diff
@@ -2020,19 +2023,36 @@ void Engine::DoCollisions(Projectile &projectile)
...
 
 
+// Fire anti-missile shots as appropriate for this ship, prioritizing nearby
+// projectiles over distant ones. Like DoCollisions, this function adds visuals
+// directly to the main visuals list. If this is multi-threaded in the future,
+// that will need to change.
+void Engine::DoAntiMissile(Ship *ship)
+{
+	sort(antiMissileTargets[ship].begin(), antiMissileTargets[ship].end(),
+			[ship](Projectile *i, Projectile *j) -> bool
```
tehhowch, line 2045: The idiomatic variable names for sorted elements are `a` and `b` :)
Should you end up keeping the sort step, please ensure you declare the comparator's arguments as `const`, and use a reference capture, e.g.
```
[&ship](const Projectile *a, const Projectile *b) -> bool
```

source/Engine.cpp
```diff
@@ -2020,19 +2023,36 @@ void Engine::DoCollisions(Projectile &projectile)
...
+			{
+				return i->Position().Distance(ship->Position()) < j->Position().Distance(ship->Position());
+			});
+	for(Projectile *projectile : antiMissileTargets[ship])
+		if(ship->FireAntiMissile(*projectile, visuals))
+			projectile->Kill();
```
tehhowch, line 2051: It's technically changing the behavior to do so, but I'm leaning towards allowing it: in your proposed algorithm, multiple pointers to the same `Projectile` can be handed out to any number of nearby ships. Thus, if one ship kills a projectile, some other ship will still have a (valid) pointer to the object, and will consider firing antimissile at that same `Projectile` too.

The game's original algorithm processed each projectile and, once it was killed, moved onto the next.

I'm fine with it, since the proposed algorithm removes that implicit information sharing that happens on a single frame, making the ship AI just a bit more fallible, too.
So, what we should do is add a comment that states it is intentional that we allow the ships to "waste" their antimissile fire on a just-killed projectile.

source/Engine.cpp
```diff
@@ -2020,19 +2023,36 @@ void Engine::DoCollisions(Projectile &projectile)
...
+	antiMissileTargets.erase(ship);
```
tehhowch, line 2052: Why do this? Call `clear()` instead, before populating the map.

source/Ship.cpp
```diff
@@ -2401,6 +2401,15 @@ bool Ship::FireAntiMissile(const Projectile &projectile, vector<Visual> &visuals
 
 
 
+bool Ship::IsInAntiMissileRangeOf(const Projectile &projectile)
+{
+	if(projectile.Position().Distance(position) > antiMissileRange)
+		return false;
+	return true;
```
tehhowch, line 2408: If instead of returning a boolean, you return a magic value (e.g. `-1`) for not-in-range, and the distance(squared) between the ship and the projectile when in range, we can reduce the number of distance calculations that need to be done in the new algorithm.

source/Engine.cpp
wjbotham: I could also have ships check that a projectile has lifespan>0 before wasting their antimissiles, then remove the check later if people actually want this change of behavior. (I'm personally more comfortable erring on the side of changing fewer behaviors.)

source/Engine.cpp
wjbotham: I removed the sort step but will keep this in mind if I put it back in after profiling!

Pointedstick:
This makes tons of logical sense, but have you considered the gameplay implications? Presumably this change makes anti-missile turrets better at protecting ships, even it if it's just by a little bit. This would imply the need for changes to commensurately reduce the strength of anti-missile fire or increase missile hardness in the interests of maintaining the balance status quo.

...Unless of course making anti-missile turrets more protective is a design goal. But then that needs to be discussed and considered. :)

source/Ship.cpp
```diff
@@ -2380,13 +2380,15 @@ bool Ship::Fire(vector<Projectile> &projectiles, vector<Visual> &visuals)
 
 
 // Fire an anti-missile.
-bool Ship::FireAntiMissile(const Projectile &projectile, vector<Visual> &visuals)
+bool Ship::FireAntiMissile(double distanceSquared, const Projectile &projectile, vector<Visual> &visuals)
 {
-	if(projectile.Position().Distance(position) > antiMissileRange)
+	if(distanceSquared > pow(antiMissileRange, 2))
```
tehhowch, line 2385: Does it still make sense to perform this check in this method?

tehhowch (commented):
One of the (unstated) arguments I had in favor of using an explicitly sorted `std::vector` is that we can change the sort easily, on a per-hardpoint basis, to implement e.g. anti-missile turn rates. That's a big enough change that it should be entirely separate from this PR, and would constitute a significant nerf to anti-missiles. That eventuality, along with enabling "projectile overkill" when multiple ships target the same projectile, should alleviate any minor benefits conveyed by shooting the nearest projectile first.

source/Engine.cpp
```diff
@@ -2020,15 +2026,30 @@ void Engine::DoCollisions(Projectile &projectile)
...
 			if(ship == projectile.Target() || gov->IsEnemy(ship->GetGovernment()))
-				if(ship->FireAntiMissile(projectile, visuals))
-				{
-					projectile.Kill();
-					break;
-				}
+			{
+				const double distanceSquared = ship->IsInAntiMissileRangeOf(projectile);
+				if (distanceSquared >= 0)
+					antiMissileTargets[ship].emplace(distanceSquared, &projectile);
```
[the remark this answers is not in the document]
wjbotham: Did you mean to change `distanceSquared` to `distance`? The former is more technically accurate.

source/Ship.cpp
```diff
@@ -2380,13 +2380,15 @@ bool Ship::Fire(vector<Projectile> &projectiles, vector<Visual> &visuals)
...
 		return false;
 	if(CannotAct())
 		return false;
-	
+	if(projectile.IsDead())
+		return false;
+
```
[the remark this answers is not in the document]
wjbotham: I am moving the check to Engine rather than removing it, not because I disagree with removing it, but just because I (as a brand new person bumbling around in the source purely for fun) do not feel it's my place to make that behavior change and would prefer to kick that decision to someone else.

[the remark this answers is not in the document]
tehhowch: You can safely consider the request of a maintainer the "decision from someone else."

source/Engine.cpp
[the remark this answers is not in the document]
tehhowch: Yes, dropping the const qualifier and changing the name of the very-short-lived local variable were both intentional. That it is a squared distance rather than a "regular' distance is an implementation detail we have chosen to make for the sake of speed, and if we remove the superfluous distance check from `Ship::FireAntiMissile`, we don't even really need to care, as we don't have to pass it to the method at all.

And to be even more truthful to it's purpose, it should probably be called a "rank"--an astute mathematician would remove the `if` statement here, since a squared distance _is always_ greater than or equal to 0.

Amazinite:

PR #6193, opened 2021-08-16
Scanner Changes

Ferociousfeind:
**Feature:** Nobody asked for this feature :)

## Feature Details
Changes scanners to be more... dynamic. The hope is to make scanning less a check for being able to approach a certain ship, and more a minigame to stay close enough to a ship, with strong enough scanners, to make sure the scan completes. I've also increased the constant "SCAN_TIME" from 60 frames to 600 frames, so that the minigame lasts 10x longer than usual (though there are lots of processes to speed it up dramatically too, so there is on average not very much change).

**Do note:** the following text describes a previous setup and the PR has since been tweaked to be more in-line with others' perceptions of scanning. See the changes to the source code (ship.cpp) for the actual changes. There, comments describe what's happening.

## UI Screenshots
This does not change any UI, but I seriously considered changing how an in-progress scan is represented. In a future expansion to this, my next steps would be to create a semi-visible circle around the scanning ship (either you, or someone scanning you) so that you can tell exactly where you should go to either better complete or better escape a scan. Scanning strength also diminishes past 50% range, so even better would be radiating circles, which fade in brightness and slow in radiating speed past the 50% mark, to intuitively communicate the idea that distance changes scanning strength. Also, perhaps changing the brightness to reflect the reduced strength due to relative motion would help communicate scanner strength too. And an actually good scanning sound, maybe like a sonar ping? Come onnn, don't knock it 'till you try it~

## Usage Examples
Before, you'd use `[cargo/outfit] scan speed` of 1 to make a scanner that takes 60 frames to scan. You'd use 4 if you wanted it to take 30 frames, or 16 if you wanted it to take 15 frames. Now, scan speed is linear (not proportional to the square root of the value), but also impacted by scanner range (more powerful scanners with the same "speed" take longer to finish scanning), and the size of the ship being scanned (larger ships take longer to scan than smaller ones), and the relative speed of the target (it is basically impossible to scan a ship whizzing past you, not just because you have a short window, but also because the results are going to be all blurry and useless).

That is to say, as you might've noticed in the changes, to achieve roughly the same result as before, multiply the scanner speed you want by the square root of the scanner power. (I actually simply multiplied the speed by the power, but if you want the in-game scanners to even more closely resemble their pre-change forms, I could change that 25 to a 5, and that 9 to a 3. I'm not sure on the surveillance pod though. ~~I...might need to change the salvage scanner too, I hadn't considered that, oops.~~)

## Testing Done
Took an Albatross out, equipped with four outfit scanners and a cargo scanner, scanned attacking pirate ships. Because they have virtually no cargo space, the one cargo scanner can complete its scan much more quickly than the four outfit scanners, unless you are maintaining a distance from the attacking Sparrow or Hawk or whatever. I also tried scanning shuttle passersby, and the occasional Cruiser passing by as well. If you only have a few tons dedicated to scanning and aren't quick on the draw, you won't finish your scan before a ship leaves. Or, if you have a lot of scanners onboard, you will have no trouble scanning whatever you like, as long as you aren't careless. It does take some tact.

Ships that have zero outfit space, or zero cargo space, will return instantly (the game does not crash due to some divide-by-zero error) even if the ship has outfits installed that happen to have zero outfit space. As well, though I use `baseAttributes.Get("outfit space")`, I actually use `attributes.Get("cargo space")` so that cargo expansions will increase cargo scan time, while outfits expansions (or otherwise using your outfit space) will not impact outfit scan times. Notice, it is cargo space, not USED cargo space, which determines cargo scan time. An empty freighter takes as long as a full freighter to scan. This is the case for outfit space too, because I fetch the target's hull's outfit space, and not its current outfit space, as its "current" outfit space is how much space is free, which will usually be very low.

I brought out a Bulk Freighter and tried to cargo-scan it. Without a sufficiently-strong scanner (note the addition of `* 0.999` to the scan decay lines) you will be unable to scan particularly large ships. However, because the scan-dissipation is so weak, you will likely not encounter this wall. Instead, it will likely manifest as scanning taking somewhat longer as you approach 100% scan, to make the scanning process ever-so-slightly more climactic than normal. It's a real fight to ascertain someone else's information without their knowledge or consent, of course.

This also means that turning and running is a valid strategy against gunboats trying to scan you to find contraband- if you keep moving, you will be more difficult to scan, and if you leave their scan range, even better!

Anyways, I think this adds some depth to scanning.

## Performance Impact
These changes do not impact performance.


Zitchas:
Sounds interesting! Having a bit more variety, challenge, and potential to escape would all be welcome.

Note that the Albatross you were using also has a bunch of built-in scanning, unless you modded that.

How does scanner interference play into this?

Ferociousfeind:
I hadn't considered interference... For now, it remains as a simple % chance for scanning to fail after you finish scanning. Perhaps it could reduce scanner effectiveness, either (likely) by dividing the current resultant scanner strength by (1+) the target's interference value. Alternatively, subtract the interference value from the resultant scanner strength for every frame of scanning ships. I'd prefer the former, though the latter is also a potential option. (Both would come with removing the % chance to fail, too.)

Zitchas:
The Albatross have the following:
```
		"outfit scan power" 18
		"outfit scan speed" 1
		"tactical scan power" 39
```

I think it'd probably fit best if interference was a slowing effect of some kind, because I find that more visual and more appealing than a chance of failure.

Ferociousfeind:
```
"outfit scan power" 18
"outfit scan speed" 1
```
That'll do it. According to the changes I've made, that means the scanner will take roughly sqrt(18) times longer to scan, so to boost it to a reasonable speed, you'd change the 1 to a 4 or 5 (likely, 4.25)

I'll see to gathering the target's interference value and decreasing scan strength by that!

source/Ship.cpp
```diff
@@ -2263,36 +2263,48 @@ int Ship::Scan()
...
-	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, const int event) -> void
+	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, double depth, const int event) -> void
 	{
 		if(elapsed < SCAN_TIME && distance < scannerRange)
 		{
 			startedScanning |= !elapsed;
 			activeScanning = true;
+			elapsed +=
+				min(1., 2. * (scannerRange - distance) / scannerRange)
+				* (5. / (5. + (Velocity() - target->Velocity()).Length()))
```
tehhowch, line 2297: Do you require _length_? Or is the squared length acceptable, too? (It is faster to compute than the non-squared version when using vector quantities.)

source/Ship.cpp
```diff
@@ -2263,36 +2263,48 @@ int Ship::Scan()
...
+				* (depth / (depth + target->attributes.Get("scan interference")))
+				* (speed / depth);
```
tehhowch, line 2299: As written, this suggests that a machine works through content at a slower rate the more content there is, when the statement should be that the time taken is proportional to the amount of content.
One could reasonably assume that a higher scanning _strength_ would work through more content-to-be-scanned faster.

tehhowch (dismissed):
Not a fan of the current proposal as-is. However I am in agreement that the current scans do not take much time at all and should likely take longer.

Note that increasing the scan durations will further increase the amount of time gunboats are swarming the player, and make "Corporate Espionage" missions that much harder than they already are.

source/Ship.cpp
```diff
@@ -2263,36 +2263,48 @@ int Ship::Scan()
 	if(!cargoDistance && !outfitDistance)
 		return 0;
 	
-	// Scanning speed also uses a square root, so you need four scanners to get
-	// twice the speed out of them.
-	double cargoSpeed = sqrt(attributes.Get("cargo scan speed"));
+	// Scanning speed does not use a square root, so that increasing both power and speed 
+	// proportionally will rise both range and speed. (The longer the range, the slower the scanner.)
```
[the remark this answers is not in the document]
Zitchas: I agree with Tehhowch on this. I'd like to see more variety in scanners with some taking longer than others, but forcing the issue by tying the speed to the power and then providing a second variable to reduce it back down doesn't seem like the best way to achieve that.

source/Ship.cpp
```diff
@@ -2055,9 +2055,9 @@ void Ship::DoGeneration()
 	// uninterrupted scanning is required to successfully scan its target.
 	// Only apply the decay if not already done scanning the target.
 	if(cargoScan < SCAN_TIME)
-		cargoScan = max(0., cargoScan - 1.);
+		cargoScan = max(0., cargoScan * 0.999 - 1.);
 	if(outfitScan < SCAN_TIME)
-		outfitScan = max(0., outfitScan - 1.);
+		outfitScan = max(0., outfitScan * 0.999 - 1.);
```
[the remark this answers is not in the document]
Ferociousfeind: Playing around with the idea of low-power scanners simply being unable to scan very large ships. Also, it makes the tail end of the scan take somewhat longer than the start, which one could describe as "climactic", but it's not _very_ climactic.

Perhaps a better implementation would be to test how long it would take to scan (assuming no interference via movement) and immediately return "no, this scan's not gonna work" if it's below a certain threshold (in this case, if the increase to `elapsed` is less than `SCAN_TIME * (1. - 0.999)`), and to return to linear decay.

source/Ship.cpp
```diff
@@ -2263,36 +2263,48 @@ int Ship::Scan()
...
+	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, double depth, const int event) -> void
 	{
 		if(elapsed < SCAN_TIME && distance < scannerRange)
 		{
 			startedScanning |= !elapsed;
 			activeScanning = true;
+			elapsed +=
+				min(1., 2. * (scannerRange - distance) / scannerRange)
+				* (5. / (5. + (Velocity() - target->Velocity()).Length()))
+				* (depth / (depth + target->attributes.Get("scan interference")))
```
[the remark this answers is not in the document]
Ferociousfeind: Understandable. However, the way interference works now doesn't feel very impactful... This is a very tangible, and potentially very easy to take advantage of, implementation, but you're right in that it doesn't make much sense physically (and indeed it would be pretty obvious something was up if you knew how long it usually takes to scan something). I'll have to think up a better implementation, if this doesn't fly.

source/Ship.cpp
```diff
@@ -2263,36 +2263,48 @@ int Ship::Scan()
...
-		outfitSpeed = 1.;
+		outfitSpeed = 100. / outfitDistance;
 	
 	// Check how close this ship is to the target it is trying to scan.
 	double distance = (target->position - position).Length();
 	
+	// Check the target's outfit and cargo space, a larger ship takes longer to scan.
+	// Normalized around 200 tons of cargo/outfit space.
+	double outfits = target->baseAttributes.Get("outfit space") / 200.;
+	double cargo = target->attributes.Get("cargo space") / 200.;
```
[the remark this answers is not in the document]
Ferociousfeind: There are four ways I saw I could do this-

**base-attribute outfit space and base-attribute cargo space**. This would return the hull's original values, and make scanning times essentially an inherent trait of hulls. Scan times would not be impacted by outfits installed.

**base-attribute outfit space and attribute cargo space**. This would return the hull's original outfit space, but its current cargo space. Because cargo space isn't actually modified by cargo being held, cargo scan times can be increased by cargo expansions, but not by loading up. Outfit space is still tied to the hull, meaning that packing your ship full of outfits (reducing attributes.Get("outfit space") close to zero) will not reduce outfit scan times. This is the solution I picked, because the two attributes behave as similarly as possible here.

**attribute outfit space and base-attribute cargo space**. This one hits both issues I avoid with the above solution- cargo expansions do not impact cargo scan time, but installing ordinary outfits DOES decrease outfit scan time.

**attribute outfit space and attribute cargo space**. This is the solution I picked first, because I hadn't considered fetching a ship's base attributes. It indeed runs into the issue that a ship with 0 outfit space remaining will be instantly scanned.

As I understand it, your solution would be much more rigorous, by taking base outfit space, plus all the outfit space added by outfits, while ignoring outfits which subtract outfit space? I feel like looping through the outfit list of the target ship could get computationally-expensive, but it _would_ let outfit expansions increase outfit scan times... (I am also not sure I know the best technique to loop through the target's outfit list in the first place)

Sorry this one's a huge rambling mess, I did think this one through, I seem to have forgotten including justification in a comment.

source/Ship.cpp
Ferociousfeind: Squared length would have increasing velocity exponentially reduce scan time, which could be good, but it doesn't sound very realistic... It would be more mechanically impactful _and_ computationally cheaper, though.

source/Ship.cpp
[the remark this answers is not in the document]
Zitchas: If scan completion time varies based on mass of target, speed of scanner, and interference, it wouldn't bemuch of a red flag. Especially if there is a variety of scanners available.

Ferociousfeind:
I am still unsure on where planetary scans happen, or where scans will fail based on the amount of scan interference you have, so I can't make changes on those parts.

Galaucus:
Does this do anything with scan failure chance, or is that still an unchanged feature?

Ferociousfeind:
> Does this do anything with scan failure chance, or is that still an unchanged feature?

Nope. I am not entirely sure where the failure happens, but wherever it is I haven't touched it.

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
 			if(elapsed >= SCAN_TIME)
 				result |= event;
 		}
 	};
-	doScan(cargoScan, cargoSpeed, cargoDistance, ShipEvent::SCAN_CARGO);
-	doScan(outfitScan, outfitSpeed, outfitDistance, ShipEvent::SCAN_OUTFITS);
-
+	doScan(cargoScan, cargoSpeed, cargoDistance, cargo, ShipEvent::SCAN_CARGO);
+	doScan(outfitScan, outfitSpeed, outfitDistance, outfits, ShipEvent::SCAN_OUTFITS);
+
```
Amazinite, line 2387:
```suggestion

```

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
 		if(elapsed < SCAN_TIME && distance < scannerRange)
 		{
 			startedScanning |= !elapsed;
 			activeScanning = true;
+			// (a/b) * (c/d) * (e/f) is more expensive than the equivalent (a*c*e) / (b*d*f)
+			elapsed += (scannerRange - distance) * speed * blur
+				/ ((sqrt(speed) + distance) * scannerRange * depth);
 			// To make up for the scan decay above:
-			elapsed += speed + 1.;
+			elapsed ++;
```
Amazinite, line 2380:
```suggestion
			++elapsed;
```

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
+	double outfitSpeed = attributes.Get("outfit scan speed");
 	if(!outfitSpeed)
 		outfitSpeed = 1.;
 
 	// Check how close this ship is to the target it is trying to scan.
-	double distance = (target->position - position).Length();
+	// To normalize 1 "scan power" to reach 100 pixels, divide this square distance by 100^2, or 10000.
+	// Because we use distance squared, to reach 200 pixels you need 4 "scan power".
+	double distance = (target->position - position).LengthSquared() * .0001;
+	double blur = 5. / (5. + (Velocity() - target->Velocity()).LengthSquared());
```
Amazinite, line 2359: Why take the velocity into account? This isn't described in the PR description or by any comment. Having bigger ships or ships that are further away take longer to scan I can understand, but I don't see why we'd have ships with bigger velocity differences take longer.

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
+	// Normalized around 200 tons of cargo/outfit space.
+	double outfits = target->baseAttributes.Get("outfit space") * .005;
+	double cargo = target->attributes.Get("cargo space") * .005;
 
 	// Check if either scanner has finished scanning.
 	bool startedScanning = false;
 	bool activeScanning = false;
 	int result = 0;
-	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, const int event) -> void
+	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, double depth, const int event) -> void
```
Amazinite, line 2370:
```suggestion
	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, const double depth, const int event) -> void
```

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
 	{
 		if(elapsed < SCAN_TIME && distance < scannerRange)
 		{
 			startedScanning |= !elapsed;
 			activeScanning = true;
+			// (a/b) * (c/d) * (e/f) is more expensive than the equivalent (a*c*e) / (b*d*f)
+			elapsed += (scannerRange - distance) * speed * blur
+				/ ((sqrt(speed) + distance) * scannerRange * depth);
 			// To make up for the scan decay above:
```
Amazinite, line 2379: Knock "above" off this comment since scan decay is done in a separate function, so the direction of "above" seemed off to me.
```suggestion
			// To make up for the scan decay:
```
The comment above the scan decay saying 1 power = 1 second to scan also needs changed.

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
-	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, const int event) -> void
+	auto doScan = [&](double &elapsed, const double speed, const double scannerRange, double depth, const int event) -> void
 	{
 		if(elapsed < SCAN_TIME && distance < scannerRange)
 		{
 			startedScanning |= !elapsed;
 			activeScanning = true;
+			// (a/b) * (c/d) * (e/f) is more expensive than the equivalent (a*c*e) / (b*d*f)
+			elapsed += (scannerRange - distance) * speed * blur
+				/ ((sqrt(speed) + distance) * scannerRange * depth);
```
Amazinite, line 2378: I feel like we need some comments here to describe exactly what's going on instead of just saying "It's done this way because the alternative is slower." What alternative? What are a, b, c, d, e, and f? How exactly are each of the inputs suppose to alter the elapsed time?

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
 		outfitSpeed = 1.;
 
 	// Check how close this ship is to the target it is trying to scan.
-	double distance = (target->position - position).Length();
+	// To normalize 1 "scan power" to reach 100 pixels, divide this square distance by 100^2, or 10000.
+	// Because we use distance squared, to reach 200 pixels you need 4 "scan power".
+	double distance = (target->position - position).LengthSquared() * .0001;
+	double blur = 5. / (5. + (Velocity() - target->Velocity()).LengthSquared());
+
+	// Check the target's outfit and cargo space, a larger ship takes longer to scan.
```
Amazinite, line 2361:
```suggestion
	// Check the target's outfit and cargo space. A larger ship takes longer to scan.
```

source/Ship.cpp
```diff
@@ -2334,44 +2334,57 @@ int Ship::Scan()
...
 		cargoSpeed = 1.;
-	double outfitSpeed = sqrt(attributes.Get("outfit scan speed"));
+
+	double outfitSpeed = attributes.Get("outfit scan speed");
 	if(!outfitSpeed)
 		outfitSpeed = 1.;
 
 	// Check how close this ship is to the target it is trying to scan.
-	double distance = (target->position - position).Length();
+	// To normalize 1 "scan power" to reach 100 pixels, divide this square distance by 100^2, or 10000.
```
Amazinite, line 2356:
```suggestion
	// To normalize 1 "scan power" to reach 100 pixels, divide this square distance by 100^2, or 10,000.
```

Amazinite (changes requested):
Interested in scanning being a little more interesting than holding a button for a fraction of a second, but I don't think that the velocity difference between the scanning and target ship should influence the scan speed. You could probably also patch this up with some more explanatory comments.

data/human/outfits.txt
```diff
@@ -248,9 +248,9 @@ outfit "Surveillance Pod"
 	"mass" 5
 	"outfit space" -5
 	"outfit scan power" 16
-	"outfit scan speed" 2
+	"outfit scan speed" 32
 	"cargo scan power" 4
-	"cargo scan speed" 2
+	"cargo scan speed" 8
```
Hurleveur, line 253:
```suggestion
	"cargo scan speed" 18
```
Same question here.

Hurleveur (commented):
Looking really nice!
I think the big comment could be a little bit clearer (not everyone has a math thesis) but other than that this seems good to me. Maybe a similar result could be achieved with simpler maths but I'm no expert on that

data/human/outfits.txt
```diff
@@ -248,9 +248,9 @@ outfit "Surveillance Pod"
 	"mass" 5
 	"outfit space" -5
 	"outfit scan power" 16
-	"outfit scan speed" 2
+	"outfit scan speed" 32
```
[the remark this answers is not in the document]
Ferociousfeind: It's been a while, I don't know. The process of making this PR was messy, and I explored multiple ideas in the process of creating it. Some things in the outfits stuck that maybe shouldn't have. (Who knows what happens between 1am and 3am?) The balance of scanning outfits would have to be entirely re-evaluated, I think.

If you think these values are good though, I'll push them to the PR.

source/Ship.cpp
```diff
@@ -2334,43 +2333,67 @@ int Ship::Scan()
...
+			/* Division is more expensive to calculate than multiplication,
+			so rearrange the formula to minimize divisions.
+
+			"(scannerRange - distance) / scannerRange"
+			This scales linearly from 1 at 0px distance to 0 at "scannerRage" distance.
+
+			"speed / (sqrt(speed) + distance)"
+			This doubles scan time over the square root of "speed" in distance,
+			while also boosting base scan speed by the square root of "speed".
+			The bonus to scan speed is reduced to 1 at ("speed" - sqrt("speed)) distance.
```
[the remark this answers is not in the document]
Ferociousfeind: Sorry, you're right, it's obscure. (It took a bit of finagling to figure out the exact behavior of it after I had derived it from the behaviors I wanted at close and very far ranges. (It gives a qualitative difference to the speed of short-range and long-range scanners, because the speed bonus does not scale at the same rate that the regular speed boost is given. It's behavior you have to play with the functions in a graphing calculator to understand. I'll simplify the comment!

Hurleveur:
In order to fix the spellcheck you just need to merge with master.

Hurleveur (approved):
My concerns have been addressed.
This is a very good improvement in my opinion, to finally not have scanning be an insta-click thing.
With this and the new evasive scan personality, the remnants will be clearly non wanting to be scanned.

source/Ship.cpp
```diff
@@ -2501,43 +2500,67 @@ int Ship::Scan()
...
+
+			"speed / (sqrt(speed) + distance)"
+			This gives a modest speed boost at no distance (speed/sqrt(speed), and
+			the boost tapers off to 0 at arbitrary distance.
+
+			"1 / depth"
+			This doubles scan time for cargo holds or outfit sections that are twice as large. */
+			elapsed += ((scannerRange - distance) * speed) / (scannerRange * (sqrt(speed) + distance) * depth);
+			// To make up for the scan decay:
+			++elapsed;
```
Amazinite, line 2557: I'd just combine these two lines.
```suggestion
			// The final +1 is to make up for scan decay that occurred this frame.
			elapsed += ((scannerRange - distance) * speed) / (scannerRange * (sqrt(speed) + distance) * depth) + 1;
```

source/Ship.cpp
```diff
@@ -2501,43 +2500,67 @@ int Ship::Scan()
...
+
+			"(scannerRange - distance) / scannerRange"
+			This line hits 1 at distace = 0, and 0 at distance = scannerRange.
+			This is a hard cap on scanning range.
+
+			"speed / (sqrt(speed) + distance)"
+			This gives a modest speed boost at no distance (speed/sqrt(speed), and
+			the boost tapers off to 0 at arbitrary distance.
+
+			"1 / depth"
```
Amazinite, line 2553: What if there's no depth? It doesn't look like a depth of 0 is handled at the moment.
There should probably be a floor to the depth fraction. Perhaps anything with less than 10 cargo or outfit space is treated as if it had 10.

source/Ship.cpp
```diff
@@ -2501,43 +2500,67 @@ int Ship::Scan()
...
 	// Check how close this ship is to the target it is trying to scan.
-	double distance = (target->position - position).Length();
+	// To normalize 1 "scan power" to reach 100 pixels, divide this square distance by 100^2, or multiply by 0.0001.
+	// Because we use distance squared, to reach 200 pixels away you need 4 "scan power".
+	double distance = (target->position - position).LengthSquared() * .0001;
+
+	// Check the target's outfit and cargo space. A larger ship takes longer to scan.
+	// Normalized around 200 tons of cargo/outfit space.
+	double outfits = target->baseAttributes.Get("outfit space") * .005;
+	double cargo = target->attributes.Get("cargo space") * .005;
```
Amazinite, line 2529: From the previous comment.
```suggestion
	double outfits = min(10., target->baseAttributes.Get("outfit space")) * .005;
	double cargo = min(10., target->attributes.Get("cargo space")) * .005;
```

Amazinite (changes requested):

source/Ship.cpp
Ferociousfeind: by god, you're right. I suppose the previous sentiment was- "if there is zero cargo (or outfit) space, then `elapsed` will roll over to positive infinity, and the scan will instantly complete" which caused no real behavioral problems. At least, I didn't pick up on anything out of the ordinary. However, avoiding division by zero is a noble goal.

source/Ship.cpp
Ferociousfeind: notably- using "min" here will have it return only numbers between 0 and 10- we want the maximum between our number, or some suitably small number, rather than the minimum. I'll be using `max(10., "outfit space")`

Zitchas:
How does scanner power affect scan speed?

For that matter, if we have a specific scan speed attribute, could we just simplify this to power only affecting range, and speed only affects speed.

Could the original post please get an updated accurate description of what is going on?

data/tooltips.txt
```diff
@@ -75,10 +75,10 @@ tip "cargo scan:"
 	`Lets you scan your target ship's cargo from up to this distance away.`
 
 tip "cargo scan power:"
-	`The range of this cargo scanner is proportional to the square root of this value.`
+	`The range of this cargo scanner is proportional to the square root of this value. The speed of the scanner also decreases in proportion to the square root of this value.`
```
Zitchas, line 78: Why does power affect speed? Does this mean that if we want two scanners to have different powers but the same actual speed, we have to do a weird balancing act where we adjust the speed faster as the power goes up to maintain the same actual speed?

For that matter, is this decrease corresponding to a decrease in the square root? Or to an increase in the square root?

data/tooltips.txt
```diff
@@ -339,10 +339,10 @@ tip "outfit scan:"
 	`Lets you scan your target ship's outfits from up to this distance away.`
 
 tip "outfit scan power:"
-	`The range of this outfit scanner is proportional to the square root of this value.`
+	`The range of this outfit scanner is proportional to the square root of this value. The speed of the scanner also decreases in proportion to the square root of this value.`
```
Zitchas, line 342: Like with cargo, does this mean that as power goes up, speed goes down?

Zitchas (changes requested):
These two comments are unclear as to what is actually going here, or what power had to do with speed when we have an actual speed attribute.

Amazinite:
> How does scanner power affect scan speed?
>
> For that matter, if we have a specific scan speed attribute, could we just simplify this to power only affecting range, and speed only affects speed.

That's effectively the case as is: "scan speed" determines your maximum possible scan speed and "scan power" determines your maximum range. Beyond that, the extra math is just to get an effect where the scan time is quicker up close as compared to afar. Although as usual with Ferocious, it's non-linear. Compared to linear scaling, the scaling used here is stronger than linear from (0, 1/3rd of the range] then weaker for the final 2/3rds, with this scaling approaching 50% effectiveness compared to linear scaling at the tip of the range.

data/tooltips.txt
Ferociousfeind: An artifact of an earlier form of the PR, most likely. Speed is uncoupled from range in this build, but speed decreases with distance (and so naturally if you expect higher ranges you'll spend more time scanning)

I'll fix the tips to the best of my ability, but I'm actually not very good at writing them. I think in a specific way.

Ferociousfeind: As an example of artifacts from previous builds- scanner strength no longer depends on relative velocity to your target. I really should fix these tips.

data/tooltips.txt
```diff
@@ -75,10 +75,10 @@ tip "cargo scan:"
 	`Lets you scan your target ship's cargo from up to this distance away.`
 
 tip "cargo scan power:"
-	`The range of this cargo scanner is proportional to the square root of this value. The speed of the scanner also decreases in proportion to the square root of this value.`
+	`The range of this cargo scanner is proportional to the square root of this value. At increasing distances, the scanner's speed will naturally decline.`
 
 tip "cargo scan speed:"
-	`The time it takes for this scanner to complete is inversely proportional to this value. Many factors can impact the speed of a scan, like the size of the target ship, its distance to the scanner, or its velocity relative to the scanner.`
+	`The time it takes for this scanner to complete, all things equal, is inversely proportional to this value. Multiple factors can impact the speed of a scan, like the size of the target ship or its distance to the scanner.`
```
Zitchas, line 81: Definitely clearer than before, although it might be better yet to have a simpler lead off.
```suggestion
	`The higher this value, the faster a scan will complete. The time it takes for this scanner to complete, all things equal, is inversely proportional to this value. Multiple factors can impact the speed of a scan, like the size of the target ship or its distance to the scanner.`
```

data/tooltips.txt
```diff
@@ -339,10 +339,10 @@ tip "outfit scan:"
 	`Lets you scan your target ship's outfits from up to this distance away.`
 
 tip "outfit scan power:"
-	`The range of this outfit scanner is proportional to the square root of this value. The speed of the scanner also decreases in proportion to the square root of this value.`
+	`The range of this outfit scanner is proportional to the square root of this value. At increasing distances, the scanner's speed will naturally decline.`
 
 tip "outfit scan speed:"
-	`The time it takes for this scanner to complete is inversely proportional to this value. Many factors can impact the speed of a scan, like the size of the target ship, its distance to the scanner, or its velocity relative to the scanner.`
+	`The time it takes for this scanner to complete, all things equal. is inversely proportional to this value. Multiple factors can impact the speed of a scan, like the size of the target ship or its distance to the scanner.`
```
Zitchas, line 345:
```suggestion
	`The higher this value, the faster a scan will complete. The time it takes for this scanner to complete, all things equal. is inversely proportional to this value. Multiple factors can impact the speed of a scan, like the size of the target ship or its distance to the scanner.`
```

Zitchas (dismissed):
Just to make it a little clearer and understandable at a glance. I'm not set in the specific wording, however having a simple and clear statement as the initial statement makes it much clearer and understandable for a wider audience. (Your existing statement, while specific, is fairly complex and not great from the perspective of accessible vocabulary. It is good for offering additional detail, not do much for the opening line.)

warp-core (commented):
Is it still necessary to raise the speed values on the scanners now that increased scan range doesn't reduce speed even if actual distance is unchanged?

As for the variable names, I'd suggest either adding `Squared` or `Metric` to the existing ones.

Amazinite:
Here's a linear equation if we wanted to change to that:
```c++
elapsed += (-0.01 * sqrt(speed * distanceSquared / power) + sqrt(speed)) / depth;
```

data/tooltips.txt
Ferociousfeind: I think it may be better to rapidly prototype tooltips on discord, so we can quickly settle on something good without the obligatory formality of github. I'm thinkin', like...

`The higher this value, the faster a scan will complete. Multiple factors can impact the speed of a scan, bla bla bla etc` because if it's too long of a tooltip it may be difficult to parse out. (even if the longest tooltip takes less than 5 seconds to read out. humans simply have low tolerance for visually-intimidating text.)

source/Ship.cpp
```diff
@@ -2269,9 +2269,8 @@ void Ship::DoGeneration()
...
 	isDisabled = hull < MinimumHull() || (!crew && RequiredCrew());
 
-	// Whenever not actively scanning, the amount of scan information the ship
-	// has "decays" over time. For a scanner with a speed of 1, one second of
-	// uninterrupted scanning is required to successfully scan its target.
+	// Whenever not actively scanning, the amount of
+	// scan information the ship has "decays" over time.
 	// Only apply the decay if not already done scanning the target.
 	if(cargoScan < SCAN_TIME)
 		cargoScan = max(0., cargoScan - 1.);
```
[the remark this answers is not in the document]
Ferociousfeind: that's a good point, it'd take up to 10 seconds for an interrupted scan to decay. Maybe that was intended at some point but it's up for debate now. How long should a scan take to decay?

source/Ship.cpp
```diff
@@ -2501,43 +2500,67 @@ int Ship::Scan()
...
 
 	// Bail out if this ship has no scanners.
 	if(!cargoDistance && !outfitDistance)
 		return 0;
 
-	// Scanning speed also uses a square root, so you need four scanners to get
-	// twice the speed out of them.
-	double cargoSpeed = sqrt(attributes.Get("cargo scan speed"));
+	// Scanning speed will be reduced by the square-distance to the target, which is
+	// close to above scanner range.
```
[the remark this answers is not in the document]
Ferociousfeind: it's not clear and I'll remove it. You get the idea from all the usage of length-squared

data/human/outfits.txt
```diff
@@ -252,9 +252,9 @@ outfit "Surveillance Pod"
 	"mass" 5
 	"outfit space" -5
 	"outfit scan power" 16
-	"outfit scan speed" 2
+	"outfit scan speed" 50
 	"cargo scan power" 4
-	"cargo scan speed" 2
+	"cargo scan speed" 18
```
Amazinite, line 257: There are other scanners in the game beyond just these human ones that would need adjusted.

data/tooltips.txt
```diff
@@ -75,10 +75,10 @@ tip "cargo scan:"
 	`Lets you scan your target ship's cargo from up to this distance away.`
 
 tip "cargo scan power:"
-	`The range of this cargo scanner is proportional to the square root of this value.`
+	`The range of this cargo scanner is proportional to the square root of this value. At increasing distances, the scanner's speed will naturally decline.`
```
Amazinite, line 78: Although distance to target does decrease scan speed, I think factors that effect speed should only be mentioned in the speed tooltip.
```suggestion
	`The range of this cargo scanner is proportional to the square root of this value.`
```

data/tooltips.txt
```diff
@@ -75,10 +75,10 @@ tip "cargo scan:"
...
 
 tip "cargo scan speed:"
-	`Determines how long it takes to perform a cargo scan. A total value of 1 means one second, and the time decreases in proportion to the square root of this value, e.g. 4 means half a second.`
+	`The higher this value, the faster a scan will complete. Multiple factors can impact the speed of a scan, like the size of the target ship or its distance to the scanner.`
```
Amazinite, line 81:
```suggestion
	`The maximum speed of this cargo scanner is proportional to the square root of this value. The observed scan speed will be impacted by multiple factors. Targets which are farther away or have a larger cargo hold will take longer to scan.`
```

data/tooltips.txt
```diff
@@ -339,10 +339,10 @@ tip "outfit scan:"
 	`Lets you scan your target ship's outfits from up to this distance away.`
 
 tip "outfit scan power:"
-	`The range of this outfit scanner is proportional to the square root of this value.`
+	`The range of this outfit scanner is proportional to the square root of this value. At increasing distances, the scanner's speed will naturally decline.`
```
Amazinite, line 342:
```suggestion
	`The range of this outfit scanner is proportional to the square root of this value.`
```

data/tooltips.txt
```diff
@@ -339,10 +339,10 @@ tip "outfit scan:"
...
 
 tip "outfit scan speed:"
-	`Determines how long it takes to perform an outfit scan. A total value of 1 means one second, and the time decreases in proportion to the square root of this value, e.g. 4 means half a second.`
+	`The higher this value, the faster a scan will complete. Multiple factors can impact the speed of a scan, like the size of the target ship or its distance to the scanner.`
```
Amazinite, line 345:
```suggestion
	`The maximum speed of this outfit scanner is proportional to the square root of this value. The observed scan speed will be impacted by multiple factors. Targets which are farther away or have a larger outfit capacity will take longer to scan.`
```

source/Ship.cpp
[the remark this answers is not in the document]
Amazinite: I'm fine with having a longer scan decay time.

source/Ship.cpp
```diff
@@ -2501,43 +2500,66 @@ int Ship::Scan()
...
-	// twice the speed out of them.
-	double cargoSpeed = sqrt(attributes.Get("cargo scan speed"));
+	double cargoSpeed = attributes.Get("cargo scan speed");
 	if(!cargoSpeed)
 		cargoSpeed = 1.;
-	double outfitSpeed = sqrt(attributes.Get("outfit scan speed"));
+
+	double outfitSpeed = attributes.Get("outfit scan speed");
 	if(!outfitSpeed)
 		outfitSpeed = 1.;
```
Amazinite, line 2517: Aren't these speed variables also squared?

source/Ship.cpp
```diff
@@ -2501,43 +2500,66 @@ int Ship::Scan()
...
 
 	// Check how close this ship is to the target it is trying to scan.
-	double distance = (target->position - position).Length();
+	// To normalize 1 "scan power" to reach 100 pixels, divide this square distance by 100^2, or multiply by 0.0001.
+	// Because this uses distance squared, to reach 200 pixels away you need 4 "scan power".
+	double distanceSquared = (target->position - position).LengthSquared() * .0001;
+
+	// Check the target's outfit and cargo space. A larger ship takes longer to scan.
+	// Normalized around 200 tons of cargo/outfit space.
+	// Also make sure this never hits 0.
```
Amazinite, line 2526:
```suggestion
	// A ship with less than 10 tons of outfit space or cargo space takes as long to
	// scan as one with 10 tons. This avoids small sizes being scanned instantly, or
	// causing a divide by zero error at sizes of 0.
```

data/human/outfits.txt
warp-core: Given that increased power doesn't inherently reduce actual scan speed anymore, are these adjustments still necessary anyway?

Amazinite: I suppose we could get away with leaving the scan speeds where they're at.

Right now this cargo scanner takes about 0.7 seconds to scan any ship. If its stats were left alone, then you're looking at 7 seconds at the quickest with against a ship with a cargo space of 200. But given that increased distance reduces scan speed, you might actually get more like 14 seconds against a 200 cargo space ship, and smaller ships would be even quicker. And even if 14 seconds seems too long to stay on target, at least the decay rate is much less proportional to the total scan time needed.

data/human/outfits.txt
```diff
@@ -214,7 +214,7 @@ outfit "Cargo Scanner"
 	"mass" 1
 	"outfit space" -1
 	"cargo scan power" 9
-	"cargo scan speed" 1
+	"cargo scan speed" 9
```
Amazinite, line 217:
```suggestion
	"cargo scan speed" 1
```

data/human/outfits.txt
```diff
@@ -224,7 +224,7 @@ outfit "Outfit Scanner"
 	"mass" 2
 	"outfit space" -2
 	"outfit scan power" 25
-	"outfit scan speed" 1
+	"outfit scan speed" 25
```
Amazinite, line 227:
```suggestion
	"outfit scan speed" 1
```

data/human/outfits.txt
Amazinite, line 257:
```suggestion
	"cargo scan speed" 2
```

Amazinite (approved):

PR #6196, opened 2021-08-17
Feature - Attempting to salvage Painted-Sky's expanded hardpoint support code.

Zitchas:
**Feature:** This is an attempt to see if we can salvage #4635 

Note: I'm not really a coder, so I'm not sure how much coding I can do on this, but it seemed pretty close to done, and I think the capacities it introduces are beneficial, particularly for plug-in support. First off, though, is seeing if it will build and pass tests...

## Feature Details
Re-implements Painted-Sky's code changes to allow large numbers of hardpoints in an up-to-date codebase.

See https://github.com/endless-sky/endless-sky/pull/4635 for conversation and development history.

## UI Screenshots
*images to come*

## Usage Examples
Mainly for plug-in support (Looking at you, Star Wars plugins with Destroyers with hundreds of turrets), although it could be used for - lacking a better word - boss fights in missions. Facing a world-ship that was actually Heron+ sized could realistically make use of 32+

## Testing Done
(to come)

## Performance Impact
Extreme numbers of turrets is expected to have an impact, but this change shouldn't affect current vanilla performance significantly.

### DIFF
--- a/source/Command.cpp
+++ b/source/Command.cpp
@@ -275,6 +275,10 @@ void Command::Clear()
 void Command::Clear(Command command)
 {
 	state &= ~command.state;
+	for(size_t i = 0; i < 8; i++)
+	{
+		firing_weps[i] &= ~(command.firing_weps[i]);
+	}
 }
 
 
@@ -283,6 +287,10 @@ void Command::Clear(Command command)
 void Command::Set(Command command)
 {
 	state |= command.state;
+	for(size_t i = 0; i < 8; i++)
+	{
+		firing_weps[i] |= command.firing_weps[i];
+	}
 }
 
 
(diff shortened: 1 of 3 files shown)


tehhowch:
Please at least attempt to have a PR merged into their repository, rather than PR their work as yours. That means we can benefit from all past review comments in context, and the primary author of the PR is preserved.

One issue I have with the original code is that it explodes the size (12x larger!) of our already-doing-too-much `Command` class, which combines player keyboard inputs with ship-level instructions. As mentioned in the original PR, this change makes this class very cache-unfriendly.

Zitchas:
OK, I have made a PR to painted-sky-master that brings it up to master and deals with the conflict, along with a letter of sorts to Painted-Sky expressing my hope that they will return to the project to work on this. (given that they understand this better than I do)

As for the workload of the `Command` class, I'm not sure what I can do about that. I'll try to take a look at it and see if I can figure out something (I'm guessing moving some of this stuff into its own class?) but I'm not sure if I have the skill to do so.

Pointedstick:
Cool, closing this. We can re-open the other MR as needed.

PR #6197, opened 2021-08-17
Make burst counts of Torpedo and Typhoon Pods equal their ammo capacity

Anarchist2:
Every missile pod except the two mentioned above have burst counts equal to their ammo capacity.


Zitchas (approved):
Good catch. Must have been missed when the two pods were dropped in capacity.

Thanks for digging into them!

PR #6199, opened 2021-08-18
Ramscoop (and Solar Collection) Changes

Ferociousfeind:
**Feature:** I am requesting this feature, all on my own!

## Feature Details
This is meant to accompany #5609 and the HD/SD/JD changes that would optionally accompany. Without that, this kind of just buffs high-mass ships... But it would've been outside of the scope of #5609 to do all three- framework for fuel consumption changes, the data file changes themselves, and the changes to solar/fuel collection- in one PR. 

This makes ramscoops much stronger, but only when within like 500 pixels of the center. At 0 pixels, you get 2x effectiveness. At 200 pixels, 1x. At about 263 pixels you're already down to 50%, followed by 25% at 325 pixels. You drop to less than 1% strength (up from a max of 200%, mind you) at 760 pixels away. It drops off very quickly.
(The formula is `y = 2 / (1 + (x/200)^4)`, which you can plug into desmos or similar to see how distance affects ramscoop capability.)

## UI Screenshots
N/A

## Usage Examples
Instead of taking the multiplier you want out of a ramscoop, and squaring it to get the "ramscoop" you should put on it, since the multiplier scales (nearly) linearly up to a point dependent on the ship, you can simply use the multiplier you want on it as the "ramscoop" on the outfit.
Solar power tapers off more quickly, but not by very much. It also tapers off to 0%, instead of 20% (if you want to have 20% strength infinitely far, simply give the outfit 20% energy generation...)

## Testing Done
The values in the code are indeed within a sane scale, and not wildly scaled too far up or down (like they were for a short while during workshopping it all together), I can tell because I've equipped an Albatross (it feels pretty good to fly) with a few ramscoops and a bunch of flamethrowers, so I can watch the numbers (and not just the bar) tick up as I approach the center, and taper off relatively quickly as I leave the center. If you cannot see the star, likely you are getting so little fuel that it doesn't matter at all.

## Performance Impact
This shouldn't impact performance, but I don't quite know enough to tell for sure. I have not tested a thousand ships at once in both Master and ramscoop-changes to see performance differences, though I suspect there is nearly no negative impact.

P.S. Apologies for dumping two PRs into the pile basically simultaneously, I have had this one in the works for a while, just needing to be finished, while scanner-changes I made the changes and then immediately PR'd.


source/Ship.cpp
```diff
@@ -2042,10 +2042,16 @@ void Ship::DoGeneration()
 		// close to the star. Carried fighters can't collect fuel or energy this way.
 		if(currentSystem)
 		{
-			double scale = .2 + 1.8 / (.001 * position.Length() + 1);
-			fuel += currentSystem->SolarWind() * .03 * scale * (sqrt(attributes.Get("ramscoop")) + .05 * scale);
+			double ramscoop = max(attributes.Get("ramscoop"), .05);
+			double scale = 0.5 * pow(Mass(), 2. / 3.); // Reusing "scale" instead of making a new temporary variable.
```
quyykk, line 2046: Might be worth measuring this. `pow` is expensive. Also imo I don't see the value of the comment.

source/Ship.cpp
```diff
@@ -2042,10 +2042,16 @@ void Ship::DoGeneration()
...
 			
-			double solarScaling = currentSystem->SolarPower() * scale;
+			ramscoop = scale - (scale * scale) / (scale + ramscoop);
```
quyykk, line 2048: equivalent to `scale * ramscoop / (scale + ramscoop)`.

Pointedstick:
Why though? What problem is this solving, or how this this change improve the game?

Ferociousfeind:
Ah shoot, I could've sworn... I typed this PR up once, and then came back and it was all gone, so I had to type it back up again, and clearly I missed one part.

It turns ramscooping into an **active** task that you **do**, instead of a **passive** event that **happens** to you. The area immediately around a star is a bunch of free fuel! But everyone else knows that too, so you will find a lot of commotion, good and bad.

I did not like the current ramscoop mechanic, because it feels like "fuel generation but remarkably finicky and unreliable" while this feels more intentional... more of a reward for getting close to the star, than a punishment for moving away from the star, does that make sense? ("X but worse" extends to solar collection too, but it feels like that's just the gimmick of solar panels in space, they _are_ energy generation but unreliable)

And, the specifics of it are tuned to higher fuel usage in general on larger ships, which isn't currently the case, hence the dependency on #5609.

source/Ship.cpp
Ferociousfeind: Oh, I hadn't realized! I thought this was the lowest, simplest form this could take, but apparently not! Tossing it into desmos, I can confirm they are equivalent.

source/Ship.cpp
Ferociousfeind: The comment was because I believe the standard practice would've been to make one variable for each value that is going to be used later, but I could accomplish this using two variables, which feels better for juggling resources more tightly instead of making a mess and not using a bunch of variables more than once. (It is justification for what might be hard to comprehend if you're just scrolling through the code. ~~It's a bad sign when you feel the need to justify yourself, isn't it?~~)

This should be equivalent to the cube-root of the mass squared, would, say, `val = [cube-root](mass); val *= val;` really be faster than `val = pow(mass, 2/3)`? Otherwise, there is no good cheap approximation afaik... No way to cheaply replicate this with more primitive operations, right?

Ferociousfeind: I suppose a crude approximation would be 0.05 * Mass(), but it'll undershoot below 1000 mass, and overshoot above 1000 mass. And at very low mass, you'll lose a lot of ramscoop you would've gotten...

Pointedstick:
Thanks for the explanation. We'll probably need some time to think it over.

quyykk:
I wrote a benchmark to compare master and PR ([clang](https://quick-bench.com/q/ROhvhh8hXjEbaj6elyi1d5gyYok) and [gcc](https://quick-bench.com/q/MWxdVN_9iK9Ox26roPFGcgGzq5I)). This PR is 1.4 slower than master with gcc and 1.5 times slower than with clang.

Ferociousfeind:
Oh dear, that really is slow. It'd be better if that `
`mass^(2/3)` was cached instead of recalculated each frame, but I'm also searching the web for good workarounds. `cbrt()` I found, combined with then *= itself would give the square of the cube root, and I think that'd be faster because `cbrt()` is designed specifically for this one situation, and `val *= val` is pretty quick too. Alternatively, I could be talked into using `0.05 * mass` I think, but you'd have to justify shortchanging very small ships and overcompensating large ships...

This is, of course, assuming that this mass^(2/3) is the last piece really tying things up. I don't wan to commit unnecessarily now that this is a PR (using up repository resources compiling and testing a half-solution) but I've changed a few things around as per your review, I just have to finalize it, get a little feedback on whether or not 0.05 * mass works, and then commit it all to streamline ramscooping.

Amazinite:
> It turns ramscooping into an **active** task that you **do**, instead of a **passive** event that **happens** to you. (this is the wording I remember using before my draft was callously shredded) The area immediately around a star is a bunch of free fuel! But everyone else knows that too, so you will find a lot of commotion, good and bad.
> 
> I did not like the current ramscoop mechanic, because it feels like "fuel generation but remarkably finicky and unreliable" while this feels more intentional... more of a reward for getting close to the star, than a punishment for moving away from the star, does that make sense? ("X but worse" extends to solar collection too, but it feels like that's just the gimmick of solar panels in space, they _are_ energy generation but unreliable)

I don't see what we'd actually gain from making using your ramscoop a more "active" task. If anything, I'd simply see this change making things more annoying given the absolutely pitiful fuel gains anywhere except right over the star, and since you haven't actually increased the max multiplier above 2x, this is just an insane nerf to ramscoops.

For those who haven't graphed it, here's the current ramscoop scale equation in red compared to the proposed scale equation in blue from a distance of 0 to the system fence. At the current distance of 1250 where your scale is 100%, you'd have a scale of 0.13% under this change. Past only a distance of about 340, you're already receiving less fuel under the proposed change than you could ever receive under the current equation.
image

I think that using a ramscoop necessarily *should* be a passive task *most of the time*. Outside of a select few weapons, fuel is only used for travel. Installing at least one ramscoop helps to make sure you're never stranded for too long, but installing multiple can help to gain an extra jump or two over the course of a journey by doing nothing given the passive nature of ramscoops. But as soon as you run out of fuel entirely, or should you be using one of those select few weapons that require fuel, or even be using a cloak, using a ramscoop *becomes* an active task. You already must necessarily be as close to the system center as you can be in order to maximize your fuel gains, which either helps you minimize the amount of time until you can travel again or maximizes the up time of your fuel consuming weapons or cloak. So given that the behavior of using ramscoops can already be either passive or active depending on the situation, I don't think that we should force them into always being an "active" thing by always requiring that the player basically be right on top of the system center lest they get absolutely nothing in terms of fuel gains.

And I imagine this was part of the intention given the description of "you know there's free fuel in the center, but so does everyone else," but I honestly wouldn't want to be in a situation where I don't have enough fuel to travel but I'm also in combat. Normally under such circumstances I'd just stall out my opponent until I have enough fuel to leave, but under these changes when you have to be so close to the system center to gain any reasonably amount of fuel, you can't simply stall your opponent. (There simply isn't enough range to do donuts around the system and not get shot while still gaining fuel.) Now if the player is capable of fighting then this is fine, but if I was capable of fighting then I wouldn't be trying to stall my opponent in the first place, so really I see this change hurting the early game the most by forcing smaller ships into combat scenarios where they have to fight over the system center in order to even leave the system.

> And, the specifics of it are tuned to higher fuel usage in general on larger ships, which isn't currently the case, hence the dependency on #5609.

I will note that having the *ability* to change the amount of fuel a ship requires does not mean that we'll necessarily make it a vanilla feature. As a concept of what one could do with the game engine I am not against #5609, but as a game mechanic for the vanilla experience I am as I feel that it unnecessarily complicates things compared to the simple to understand fuel mechanics we currently have. More complicated mechanics can be interesting, but not always, and overcomplicating actions as simple as how much fuel you need to travel a certain distance can be more burdensome than fun.

Another issue is the fact that you're making two changes at once. The first change is to change the scale to make using a ramscoop more "active" as you see it, and the second change is to make fuel gains be based off of ship mass. Making multiple changes at once is generally a no-no, and in this case you're definitely not helping yourself by tying the "more active ramscoop" behavior to the "more mass = more fuel because of #5609" behavior.

If we split the two changes being made though and only look at the change to steepen the ramscoop curve, I think that simply tweaking the constants of the existing equation would be simpler than creating a new one. It isn't as harsh of a dropoff (which will necessarily be the case given the form of the base equation being used), but the same effect of making the use of a ramscoop more active (but not necessarily always active) can basically be created by just tweaking the constants of the existing equation. As with before, red is the current equation and blue is the proposed equation, but now orange is a potential equation that changes the scale behavior from 2x at the max, 0.2x at the min, and 1x at a distance of 1250 to 3x at the max, 0x at the min, and 1x at a distance of 200.
image

Such a change to the constants of the existing equation could easily be made at any time by anyone when using #5978. (I don't currently have changing such constants implemented into that PR, but I already have the code changes that could allow it to be done.) Any change to have fuel gains be based off of ship mass could similarly be a toggled mechanic under #5978.

Ferociousfeind:
Hmm... there must've been some miscommunication... this change makes ramscoops _orders of magnitude more powerful._

Looking at the profiles pre- and post-change for 1 "ramscoop" on a ship with an arbitrarily-high mass (since ramscoop gets limited based on mass), the post-change curve has a maximum of 2.5, compared to pre-change which has been normalized to 1.1 (or, 1, +0.1 because of the "even ships with no ramscoop can scoop" safeguard in place)
image

And if we increase ramscoop, the disparity becomes exponentially greater (but I also have to increase the arbitrary high mass to keep the dampening effect mass has completely invisible). Here is a ship with 25 "ramscoop", the pre-change line has a maximum of 5.1, and the post-change line has a maximum of 62.5, or 12.25x stronger.
(A ship of a reasonable 800 tons only gets up to 39.55, or 7.75x stronger.)
image

This effect manifests due to a few things I appear to have poorly explained...

`y=x^(1/2)` expands very, very slowly. for x=1, y=1. But for x=100, y=10. For x=1000000, y=1000. I have replaced this with effectively `y=x`, while roping in another variable (mass) to restrict the growth of the line instead. If we analyze mass instead of ramscoop, the equation is `y=x^(2/3)`, which simply grows faster than `y=x^(1/2)`. At x=8, y=4. At x=64, y=16. At x=1000, y=100. At x=1000000, y=10000.

I have also changed a constant, there was a `0.03` in the fuel gain line, which I changed to a `0.075`, which is 2.5x larger, which also contributes to the higher fuel intake. In the end, it comes down to a comparison on how mass effects the ramscoop multiplier, (on a graph of y = multiplier, x = "ramscoop", with a slider for mass) At arbitrarily high mass, all the other changes into account, the maximum fuel you could get increases with "ramscoop" directly proportionally. Before, since mass never came into the equation, it has no effect and the multiplier rises with "ramscoop" along the y=sqrt(x) line. At lower masses, you'll notice that as "ramscoop" approaches infinity, the multiplier approaches some finite value. Perhaps they could use some tweaking, or perhaps the capping could simply be removed ("outfit space works well enough", one could say, though the reason I have included the mass limitation is because it is analogous to ships having only so much surface area...)
image


TL;DR: Noooo! You get a lot of fuel _really_ fast in exchange for only being able to scoop when _very_ close to the system center!
This is _**not**_ a nerf to ramscoops. ☹️ Ramscoops as they currently are, are really slow, and I have made that not the case.

Amazinite:
Well from your own words in the PR description it's weaker.
> At 0 pixels, you get 2x effectiveness. At 200 pixels, 1x. At about 263 pixels you're already down to 50%, followed by 25% at 325 pixels. You drop to less than 1% strength (up from a max of 200%, mind you) at 760 pixels away. It drops off very quickly.
> (The formula is `y = 2 / (1 + (x/200)^4)`, which you can plug into desmos or similar to see how distance affects ramscoop capability.)

2x at 0 is the same, 1x at a distance of 200 is closer than 1x at a distance of 1250, and the existing equation never drops below 20% efficiency. This was what I was looking at. But like I said, mixing features in a single PR is generally a no-no, exactly because of miscommunications like this.

All that aside, my critique of forcing you to need to go to the system center in order for ramscoops to actually be of use leading to weak ships being forced into combat still stands. And there's also the fact that we already have a balance situation where larger ships are much more powerful than smaller ships and there's almost no reason to use a smaller ship at all. We have multiple open PRs and issues looking to change that. Yet this PR would just add another way that makes larger ships more powerful than smaller ones by increasing their fuel gains. I understand that that is a result of the intention to have larger ships also require more fuel, but, leaving aside that I already mentioned that I'm not really interested in that idea, more fuel for larger ships necessarily means that you'll be able to more effectively use fuel consuming equipment like weapons or cloaks. So what you're creating is a rich get richer scenario where ships need to be able to hold the center of the system to be able to refuel, ships that refuel faster can potentially be stronger by using that fuel on offensive weapons, and the ships that are already the best in combat and therefore the best to be able to hold the center of a system under normal circumstances now receive a bigger boost to their offensive capabilities as a result of holding the system center.

Amazinite:
Closing this since, as described, I don't see this as a desirable change.

PR #6200, opened 2021-08-18
feat(UI): Highlight raid alert message

Terin:
image

This PR sets the raid alert message's importance to Highest, highlighting it orange along with the mission failure or hyperspace alerts. The Raid Warning tutorial mission is still important for specifically laying out why it happens, but now that we have the capability to highlight messages, this seems like one of the more important ones.

### DIFF
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1232,7 +1232,7 @@ void Engine::EnterSystem()
 				{
 					raidFleet->Place(*system, newShips);
 					Messages::Add("Your fleet has attracted the interest of a "
-							+ raidGovernment->GetName() + " raiding party.", Messages::Importance::High);
+							+ raidGovernment->GetName() + " raiding party.", Messages::Importance::Highest);
 				}
 	}
 	


Zitchas:
Seems good. How often might this message get triggered? Any possibility that it gets spammed?

Terin:
It just triggers (at most) once per system, which is when a raid fleet would be spawned in.

Zitchas (approved):
OK, in which case this looks like probably a good change. - although it does make me think that maybe there should be a way for people to change the priority? After the nth time I have gotten a warning about my fleet attracting pirates, it isn't going to be a high priority anymore.

tehhowch:
Yep, my thoughts exactly. This scenario is possibly even desired by some players, so I'm not certain it should scream at players.

That said, we can easily roll it back if it isn't popular.

Terin:
Messages fade pretty quickly still, and even more so if fast-forwarding. I'd be more worried if it was like, flashing, but in this case it's just a short orange message that goes away in a few seconds. Enough to draw the eye if you aren't sure what's happening, but still pretty easy to ignore if you want to.

Zitchas:
Another idea for beyond the scope of this PR: Specify a "count" on messages, after that many counts it fades down a tier. Highest -> High, High -> normal. Just an idea.

PR #6210, opened 2021-08-28
Remove dynamic cast in test

thomasballinger:
**Refactor** This PR makes a small change to https://github.com/endless-sky/endless-sky/pull/5736 to remove the `dynamic_cast`, currently the only feature in the codebase requiring C++ runtime type info support, thereby allowing `-fno-rtt` to be used.

This seems like a decent place to do a dynamic cast instead: in a test! I have no significant experience in C++ codebases to have an opinion here. I'm just removing this because it's convenient to me and it follows the coding style guide to avoid dynamic casts.

## Testing Done
Compiles.

## Performance Impact
Nothing observed. Reading about `dynamic_cast` vs `static_cast`, there could be something small here.

### DIFF
--- a/source/Test.cpp
+++ b/source/Test.cpp
@@ -469,11 +469,11 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
 					if(gamePanels.Root() != gamePanels.Top())
 						Fail(context, player, "engine not active due to panel on top, and can only send commands to the engine");
 					
-					// Both get as well as the cast can result in a nullpointer. In both cases we
-					// will fail the test, since we expect the MainPanel to be here.
-					auto mainPanel = dynamic_cast<MainPanel *>(gamePanels.Root().get());
+					// The get can result in a nullpointer which will will fail the test
+					// since we expect the MainPanel to be here.
+					auto mainPanel = static_cast<MainPanel *>(gamePanels.Root().get());
 					if(!mainPanel)
-						Fail(context, player, "root gamepanel of wrong type when sending command");
+						Fail(context, player, "no root gamepanel when sending command");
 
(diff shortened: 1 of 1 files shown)


quyykk:
We do need to be careful here. If the root panel isn't actually a `MainPanel` then we'll get UB here unlike with `dynamic_cast`.

But it seems like for `gamePanels` there is only `MainPanel` as root possible, as far as I can tell. If that is indeed the case and I didn't miss something, then this change would indeed be safe.

source/Test.cpp
```diff
@@ -469,11 +469,11 @@ void Test::Step(Context &context, UI &menuPanels, UI &gamePanels, PlayerInfo &pl
...
-					// will fail the test, since we expect the MainPanel to be here.
-					auto mainPanel = dynamic_cast<MainPanel *>(gamePanels.Root().get());
+					// The get can result in a nullpointer which will will fail the test
+					// since we expect the MainPanel to be here.
+					auto mainPanel = static_cast<MainPanel *>(gamePanels.Root().get());
 					if(!mainPanel)
-						Fail(context, player, "root gamepanel of wrong type when sending command");
+						Fail(context, player, "no root gamepanel when sending command");
 
 					mainPanel->GiveCommand(stepToRun.command);
```
petervdmeer, line 478: If we are concerned about type-safety here, then we could also choose to make the GiveCommand() method a virtual protected method on Panel. Doing that will result in us not needing to do any cast here at all (since `gamePanels.Root()` will return a shared_ptr to Panel.

petervdmeer: I just tried this approach in https://github.com/endless-sky/endless-sky/pull/6262 .

tehhowch:
Addressed by #6262

PR #6214, opened 2021-08-30
test: Add time information to Ember Wastes Navigation test

petervdmeer:
**Feature:** This PR test both the navigation code and the time conditions

## Feature Details
This PR adds time information to the navigation test to verify that the steps take the amount of time that was expected.
This PR also tests the time conditions code itself (which is useful if we move the time conditions to derived automatic conditions as is done in #5244).

## UI Screenshots
N/A

## Usage Examples
Just run the test.

## Testing Done
Just ran the test (in combination with the code from #5244).

## Performance Impact
N/A (doesn't take a lot of additional test-time either)

### DIFF
--- a/data/tests/tests_wormhole_navigation.txt
+++ b/data/tests/tests_wormhole_navigation.txt
@@ -118,7 +118,7 @@ test-data "Ember Wastes Navigation"
 	category "savegame"
 	contents
 		pilot Bobbi Bughunter
-		date 16 11 3013
+		date 16 12 3012
 		system "Tania Australis"
 		planet Ingot
 		clearance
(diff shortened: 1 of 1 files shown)


data/tests/tests_wormhole_navigation.txt
```diff
@@ -350,87 +350,179 @@ test "Tests - Ember Wastes Navigation"
...
 			travel Limen
 			travel Mora
+		# Our test-savegames don't have the date conditions serialized. Need to
+		# have passed at least one date-transition to get the date conditions set
+		# (until PR https://github.com/endless-sky/endless-sky/pull/5244 removes
+		# this need to serialize the date conditions completely).
+		#assert
+		#	year == 3012
+		#	month == 12
+		#	day == 16
```
tehhowch, line 360: As a future improvement to the integration testing framework, consider how you might identify _which assert_ in a group of them did not pass.

Getting the specific assert text tokens will probably be easier in the generic condition case than getting the substituted token values, and is likely "enough" when combined with the associated failure context to deduce the problem.

petervdmeer: Yes, your other idea of printing the values of the variables used in the assert/condition could also be helpful here (since currently only the "test: ..." conditions get printed). Still on the todo list somewhere.

Issue #5649, opened 2021-01-16
Bug: Using a plug-in to add outfit space also grants extra weapons


Zitchas:
**Describe the bug**
When I am using a plug-in that adds outfit space to the standard existing Albatross, when I purchase an Albatross (which is listed in the shipyard as having 6 inhibitor canons, 2 turrets, 1 PDT) but then look at the player info screen, and it shows all 7 canons slots filled with inhibitor canons, and the turret slots filled with an extra thrasher turret and PDT. The outfit list still says exactly what the shipyard said it should have. Selling *all* weapons and taking off and firing results in me shooting, and I have the extra weapons in their spots, and they fire. Reloading the game causes the game to flag the extra outfits as "not existing in the outfit list" and removes them.

**To Reproduce**
1. Make a plug-in with the following code:
```
ship "Albatross"
    add attributes
        "outfit space" 34
```
or
```
ship "Albatross"
    attributes
        "outfit space" 34
```
1B. Run Endless Sky.
2. Go to a Remnant planet
3. Buy an Albatross
4. Before leaving shipyard, note that the shipyard lists 6 inhibitor canons, 2 thrasher turrets, and 1 PDT.
5. Leave shipyard
6. Check player info. Note that the Player Info outfit lists shows the correct numbers of outfits.
7. Note that the ship pictogram showing weapon locations has all seven canon slots filled with inhibitor canons. Likewise, 3 of the turret slots have thrasher turrets, and 2 of the turret slots have PDT.
8. Launch, and verify that all the weapons do actually work.
9. Land, and sell all the weapons.
10. take off, and note that the extra canon and two turrets are functional
11. Reload the game, note that the game gives errors related to those three outfits, and removes them

**Expected behavior**
The official "add attribute" code, used as described in the "plugin creation section" should not cause extra weapons to be added to the target ship.

**Screenshots**
![Uh-guns-turrets](https://user-images.githubusercontent.com/32169904/104824382-63acfd80-580e-11eb-94aa-645cf33064a4.PNG)

**System (please complete the following information):**
 - OS: Windows 10 Pro
 - Game Source: github compiled version on 2021-Jan-15
 - Version: 

**Link to save file**
[e.g. hastebin.com, direct upload, GitHub Gist]

**Additional context**
Possibly related problem: #4413

vitalchip:
[...]

PR #6216, opened 2021-08-31
Fix hardpoints not being cleared when overwriting a ship's outfits.

quyykk:
**Bugfix** 

When overwriting an existing ship by only changing its outfits, the outfits in the `armament` do not get cleared. This can result in ships having the wrong outfits in their armament (and if you look at the info screen you can see the wrong weapons). This results in a ship that cannot fire. If you reload your save this issue is recognized by the game with a warning and fixed (and there'll be a stray outfit entry in your outfit list).

This PR makes it so that every outfit is uninstalled from every hardpoint when only a `"outfit"`/`"gun"`/`"turret"` node is present.

sparrow

## Fix Details

There is another possible implementation for a bugfix:  In `FinishLoading` remove any hardpoint outfits that aren't in the ship's outfit list. In contrast to this PR if the outfit list changes but the weapons don't, the original hardpoint assignments will be preserved. I don't know whether this is desirable or not.

## To Reproduce

I made a plugin for this. Unzip this in your plugins folder and create a new pilot and buy the "Wanderer Sparrow", an invalid sparrow with 2 Beam Lasers in its outfit list and 2 Sunbeams in its armament.
[broken.zip](https://github.com/endless-sky/endless-sky/files/7085166/broken.zip)


Amazinite:
Ref: #5649

quyykk:
Thanks! The linked issue is now fixed.

We only actually need to clear `equipped` when we actually modify outfits or hardpoints. If not, then there is no reason at all to clear it.

tehhowch:
>  if the outfit list changes but the weapons don't, the original hardpoint assignments will be preserved. I don't know whether this is desirable or not.

It is indeed desirable for the hardpoint assignments to be preserved.

quyykk:
Is there anything left do to? I would really like to see this bug fixed.

tehhowch:
I need to pull down and verify the fixes. Until then, mind merging the base / rebasing onto latest, to get an updated CI run?

source/Ship.cpp
```diff
@@ -404,6 +404,17 @@ void Ship::Load(const DataNode &node)
 				else
 					grand.PrintTrace("Skipping invalid outfit count:");
 			}
+			
+			// Verify we have at least as many installed outfits as were identified as "equipped."
+			// If not (e.g. a variant definition), ensure FinishLoading equips into a blank slate.
+			if(!hasArmament)
+				for(const auto &it : equipped)
+					if(static_cast<int>(outfits.count(it.first)) < it.second)
```
quyykk, line 412: .... I'm dumb. @tehhowch this code has an error that I didn't notice (as usual). Sometimes I'm blind when it comes to errors in my code. 

`outfits` is a map so `count` returns either `1` or `0.... It should `find` and then use the value of the pair to compare.

tehhowch: Yeah, I missed it too when checking the bugs that were fixed by the new logic. I had even taken notice of the need for the static cast due to the type difference, but missed that it was due to comparing the wrong things. Knowing what you were intending to do, plus the wonderful language semantics (_Of course `outfits.count` counts the outfits, that's the point of the condition comparison..._), was just enough to get us both. 😅

Issue #6220, opened 2021-09-01
--resources flag behavior is confusing on mac (and linux?)


thomasballinger:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

When endless-sky is started with the `--resources /foo/bar` flag on mac, the resource directory used is `/foo/bar/../Resources/` which causes `/foo/bar/../Resources/` and `/foo/bar/../` to be checked for a `credits.txt` file before `/foo/bar/credits.txt` is checked. This is due to resource finding hacks being applied even for flag-specified resources:

https://github.com/endless-sky/endless-sky/blob/be21998143b7577bf8381d34d55f385d01c8d030/source/Files.cpp#L104-L107

I haven't run into this bug but I was confused when reading the code, and it could manifest if a user had a resources directory with a credits.txt as a sibling.

I believe there are similar issues with linux based on the code, but I haven't tested this:

https://github.com/endless-sky/endless-sky/blob/be21998143b7577bf8381d34d55f385d01c8d030/source/Files.cpp#L94-L103

### Steps to Reproduce

```
$ mkdir -p fakeresources/data fakeresources/images fakeresources/sounds
$ touch fakeresources/credits.txt
$ /Users/tomb/Library/Developer/Xcode/DerivedData/EndlessSky-asymldgrudnqsdclwufttpzdiyop/Build/Products/Debug/Endless\ Sky.app/Contents/MacOS/Endless\ Sky --resources fakeresources/
# game launches, using specified (empty) resources

$ touch credits.txt
$ /Users/tomb/Library/Developer/Xcode/DerivedData/EndlessSky-asymldgrudnqsdclwufttpzdiyop/Build/Products/Debug/Endless\ Sky.app/Contents/MacOS/Endless\ Sky --resources fakeresources/
# Unable to find the resource directories! dialog appears
```

### Expected Behavior

the `../Resources` hack is only used when no `--resources asdf` flag is specified

### Operating System

MacOS Big Sur 11.5.1

### Game Source

Built from source

### Game Version

commit SHA be21998143b7577bf8381d34d55f385d01c8d030

quyykk:
Turns out that the hack that is `../Resources/` is not needed at all. #7929 removed it, so this is now fixed.

PR #6222, opened 2021-09-01
Don't try to fix resources location when --resources is used.

thomasballinger:
**Bugfix:** This PR resolves https://github.com/endless-sky/endless-sky/issues/6220

## Fix Details
Only apply most adjustments to the resources path when it has not been explicitly specified on the command line.

Still `FixWindowsSlashes(resources)` and append a trailing slash if it is missing.

## Testing Done
I tested the new mac behavior against the repro described in https://github.com/endless-sky/endless-sky/issues/6220 and a credits.txt file in the parent directly no longer causes resource loading to fail. I have not tested the linux changes.

### DIFF
--- a/source/Files.cpp
+++ b/source/Files.cpp
@@ -61,25 +61,21 @@ namespace {
 	
 	void SetResourceDirectory(string &resources)
 	{
-		bool specifiedWithArg = resources.empty();
-		if(resources.empty())
-		{
-			// Find the path to the resource directory. This will depend on the
-			// operating system, and can be overridden by a command line argument.
-			char *str = SDL_GetBasePath();
-			if(!str)
-				throw runtime_error("Unable to get path to resource directory!");
-			
-			resources = str;
-			SDL_free(str);
+		// Find the path to the resource directory. This will depend on the
+		// operating system, and can be overridden by a command line argument.
+		char *str = SDL_GetBasePath();
+		if(!str)
+			throw runtime_error("Unable to get path to resource directory!");
 
-			// Create the "plugins" directory if it does not yet exist, so that it is
-			// clear to the user where plugins should go.
-			{
-				char *str = SDL_GetPrefPath("endless-sky", "plugins");
-				if(str != nullptr)
-					SDL_free(str);
-			}
+		resources = str;
+		SDL_free(str);
+
+		// Create the "plugins" directory if it does not yet exist, so that it is
+		// clear to the user where plugins should go.
+		{
+			char *str = SDL_GetPrefPath("endless-sky", "plugins");
+			if(str != nullptr)
+				SDL_free(str);
 		}
 
 #if defined _WIN32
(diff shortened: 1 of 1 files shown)


source/Files.cpp
```diff
@@ -91,28 +95,32 @@ void Files::Init(const char * const *argv)
...
 #endif
+	}
 	// If the resources are not here, search in the directories containing this
 	// one. This allows, for example, a Mac app that does not actually have the
 	// resources embedded within it.
 	while(!Exists(resources + "credits.txt"))
 	{
 		size_t pos = resources.rfind('/', resources.length() - 2);
-		if(pos == string::npos || pos == 0)
+		if(pos == string::npos || pos == 0 || resourcesSpecifiedWithArg)
```
tehhowch, line 123: Mind splitting the Init implementation up, e.g. into private / namespaced helper methods `SetResourceDirectory(string&,bool` and `SetConfigDirectory(string&)`?

Then we can immediately bail if a resource or config dir is specified but invalid, with a more meaningful runtime_error message.

source/Files.cpp
```diff
@@ -57,109 +57,123 @@ namespace {
...
-		char *str = SDL_GetBasePath();
-		if(!str)
-			throw runtime_error("Unable to get path to resource directory!");
-		
-		resources = str;
-		SDL_free(str);
-	}
+		bool specifiedWithArg = resources.empty();
+		if(resources.empty())
+		{
```
tehhowch, line 65: My earlier comment was more to the tune of
```cpp
if(!resources.empty())
{
  // Add / if needed
  if(!Files::Exists(resources + "credits.txt"))
    throw runtime_error("provided resources directory \"" + resources + "\" is not valid");
}
else
{
  // Go looking.
}
// Verify data, images, sounds, etc.
```

source/Files.cpp
```diff
@@ -57,109 +57,123 @@ namespace {
...
+			SDL_free(str);
+
+			// Create the "plugins" directory if it does not yet exist, so that it is
+			// clear to the user where plugins should go.
+			{
+				char *str = SDL_GetPrefPath("endless-sky", "plugins");
+				if(str != nullptr)
+					SDL_free(str);
+			}
+		}
```
tehhowch, line 82: This should probably still happen even if the game is always started with a resource path.

(It seems more suited to config, as well)

thomasballinger: Are you thinking create plugins directory wherever config is set to (my preference) or always at this default location? re moving to config, whoops, yep will do.

tehhowch: If a config directory is provided, then yes, we'd want to ensure the expected directory structure exists in it.

thomasballinger: Ok, so by default we'll use SDL_GetPrefPath("endless-sky", "plugins"); to create the directory, but if a custom config is specified we'll check for plugins and create a folder there if necessary using system specific calls.

thomasballinger:
The pattern is now

```
if (resources.empty())
	SetResourceDirectory(string &resources) // go looking
else
{
	// fix up the string (windows \ -> /, add a trailing /)
}
// throw runtime error if resources do not exist
```
however credits.txt for resources is not longer enforced explicitly set directories, it's only used for discovery.

Does this pattern work, or would you prefer I swap the conditional to `if (!resources.empty()`?

Should credits.txt in resources be required when `--resources` is manually specified? (currently only data, images, and sounds are required)

source/Files.cpp
```diff
@@ -58,70 +58,60 @@ namespace {
...
+
 #if defined _WIN32
-	FixWindowsSlashes(resources);
+		FixWindowsSlashes(resources);
 #endif
-	if(resources.back() != '/')
-		resources += '/';
+		if(resources.back() != '/')
+			resources += '/';
```
tehhowch, line 85: This is a pattern repeated a couple times (fixing slashes if windows and then ensuring the string ends in `/`), mind extracting to a helper function?

tehhowch (commented):
I like to think of the check for credits.txt in the resources directory as a "you must acknowledge those who built the content / engine when using it," but it's really not a robust check at all.

quyykk:
#7929

PR #6223, opened 2021-09-02
fix(ui): Decimal number display bug

warp-core:
**Bugfix:**

## Fix Details
Numbers with a non-zero hundredths digit but a zero in the tenths column were displaying without the decimal point or the zero in the tenths column:

## Testing Done
Seen in game where the value should be 107.09 (turn = 18420, mass = 172, 18420/172 = ~107.093) but displays as 1079.
image
Also tested directly with Format::Number before and after change. Before change, input of 107.09 would have result as '9' at the point of calling FormatInteger, after the change, it is '90.', as expected.

## Save File
[warp core~3014-01-31 number display bug save.txt](https://github.com/endless-sky/endless-sky/files/7095560/warp.core.3014-01-31.number.display.bug.save.txt)

### DIFF
--- a/source/text/Format.cpp
+++ b/source/text/Format.cpp
@@ -132,6 +132,8 @@ string Format::Number(double value)
 	
 	// Only show decimal places for numbers between +/-10'000.
 	double decimal = modf(value, &value);
+	// Fix any floating point inaccuracy.
+	decimal = round(decimal * 100.) / 100.;
 	if(decimal && value < 10000)
 	{
 		int tenths = static_cast<int>(decimal * 10.);
@@ -143,11 +145,13 @@ string Format::Number(double value)
 			
 		int hundredths = static_cast<int>(decimal * 10.);
 		// Values up to 1000 may have two decimal places.
-		if(value < 1000)
+		if(value < 1000 && hundredths)
 		{
-			if(hundredths)
-				result += static_cast<char>('0' + hundredths);
+			result += static_cast<char>('0' + hundredths);
 		}
+		//Values of 1000 or more should not display any decimal places if there are no tenths regardless of the hundredth values.
+		else
+			hundredths = 0;
 		if(tenths || hundredths)
 		{
 			result += static_cast<char>('0' + tenths);
--- a/source/text/Format.cpp
+++ b/source/text/Format.cpp
@@ -135,19 +135,20 @@ string Format::Number(double value)
 	if(decimal && value < 10000)
 	{
 		int tenths = static_cast<int>(decimal * 10.);
+		
+		decimal *= 10.;
+		decimal -= static_cast<int>(decimal);
+		// Fix any floating point inaccuracy.
+		decimal = round(decimal * 100.) / 100.;
+			
+		int hundredths = static_cast<int>(decimal * 10.);
 		// Values up to 1000 may have two decimal places.
 		if(value < 1000)
 		{
-			decimal *= 10.;
-			decimal -= static_cast<int>(decimal);
-			// Fix any floating point inaccuracy.
-			decimal = round(decimal * 100.) / 100.;
-			
-			int hundredths = static_cast<int>(decimal * 10.);
 			if(hundredths)
 				result += static_cast<char>('0' + hundredths);
 		}
-		if(tenths)
+		if(tenths || hundredths)
 		{
 			result += static_cast<char>('0' + tenths);
 			result += '.';
--- a/source/text/Format.cpp
+++ b/source/text/Format.cpp
@@ -133,7 +133,7 @@ string Format::Number(double value)
 	// Only show decimal places for numbers between +/-10'000.
 	double decimal = modf(value, &value);
 	// Fix any floating point inaccuracy.
-	decimal = round(decimal * 100.) / 100.;
+	decimal = round(decimal * 1000.) / 1000.;
 	if(decimal && value < 10000)
 	{
 		int tenths = static_cast<int>(decimal * 10.);
--- a/tests/src/text/test_format.cpp
+++ b/tests/src/text/test_format.cpp
@@ -137,15 +137,20 @@ TEST_CASE( "Format::Number", "[Format][Number]") {
 		CHECK( Format::Number(0.51) == "0.51" );
 		CHECK( Format::Number(0.56) == "0.56" );
 		CHECK( Format::Number(0.871) == "0.87" );
+		CHECK( Format::Number(0.072) == "0.07" );
 	}
 	SECTION( "Decimals between 10 and 100" ) {
 		CHECK( Format::Number(44.1234) == "44.12" );
 		CHECK( Format::Number(94.5) == "94.5" );
+		CHECK( Format::Number(10.1) == "10.1" );
+		CHECK( Format::Number(10.01) == "10.01" );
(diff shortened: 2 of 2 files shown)


tehhowch (changes requested):
Needs to update the unit tests.

warp-core:
> Needs to update the unit tests.

Just to check, this means add some of the things I encountered giving incorrect results to the unit tests?

quyykk:
> Just to check, this means add some of the things I encountered giving incorrect results to the unit tests?

Yes. And any other number you'd think would be a good idea to test.

warp-core:
Another question: is the desired behaviour for it to never round up, none of the unit tests test for rounding up but I think my new line 136 causes it to round up the hundredths value if the thousandths digit is 5 or above. The old code would round up depending on the ten thousandths digit (my code can be easily changed to behave the same by updating line 136 to multiply and divide by 1000) but this doesn't seem like desirable behaviour, even if it is only an issue in edge cases (rounding up by the ten thousandths digit will only be visible if the thousandths digit is a 9).

tehhowch:
It's fine if the new code rounds up differently.

warp-core:
I did some further testing, running the following code with the updated version of Format:Number (just 'Number' in this example) and it doesn't report any issues:

```c++
for (int64_t integerPart = 0; integerPart < 999; integerPart++) {
		for (int64_t tenths = 0; tenths < 9; tenths++) {
			for (int64_t hundredths = 0; hundredths < 9; hundredths++) {
				for (int64_t trailingDigits = 0; trailingDigits < 9; trailingDigits++) {
					double value = static_cast<double>(integerPart) + (static_cast<double>(tenths) / 10) + (static_cast<double>(hundredths) / 100) + (static_cast<double>(trailingDigits) / 1000);
					string expectedResult;
					if (!hundredths) {
						if (tenths) {
							expectedResult += static_cast<char>('0' + tenths);
							expectedResult += ".";
						}
					}
					else {
						expectedResult += static_cast<char>('0' + hundredths);
						expectedResult += static_cast<char>('0' + tenths);
						expectedResult += ".";
					}
					string result = Number(value);
					if (result != expectedResult) {
						cout << "Value: " << value << " gives: " << result << " instead of: " << expectedResult << "\n";
						
					}
				}
			}
		}
	}

	for (int64_t integerPart = 1000; integerPart < 9999; integerPart++) {
		for (int64_t tenths = 0; tenths < 9; tenths++) {
			for (int64_t hundredths = 0; hundredths < 9; hundredths++) {
				for (int64_t trailingDigits = 0; trailingDigits < 9; trailingDigits++) {
					double value = static_cast<double>(integerPart) + (static_cast<double>(tenths) / 10) + (static_cast<double>(hundredths) / 100) + (static_cast<double>(trailingDigits) / 1000);
					string expectedResult;
					if (tenths) {
						expectedResult += static_cast<char>('0' + tenths);
						expectedResult += ".";
					}
					string result = Number(value);
					if (result != expectedResult) {
						cout << "Value: " << value << " gives: " << result << " instead of: " << expectedResult << "\n";

					}
				}
			}
		}
	}

	for (int64_t integerPart = 10000; integerPart < 99999; integerPart++) {
		for (int64_t tenths = 0; tenths < 9; tenths++) {
			for (int64_t hundredths = 0; hundredths < 9; hundredths++) {
				for (int64_t trailingDigits = 0; trailingDigits < 9; trailingDigits++) {
					double value = static_cast<double>(integerPart) + (static_cast<double>(tenths) / 10) + (static_cast<double>(hundredths) / 100) + (static_cast<double>(trailingDigits) / 1000);
					string expectedResult;
					string result = Number(value);
					if (result != expectedResult) {
						cout << "Value: " << value << " gives: " << result << " instead of: " << expectedResult << "\n";

					}
				}
			}
		}
	}
```

warp-core:
It looks like values with a decimal component greater than 0.9995 will go wrong (both before and after my changes). How should this be handled? Truncate (no round up) or add one to the integral part and display no decimal places?

waterhouse:
> It looks like values with a decimal component greater than 0.9995 will go wrong (both before and after my changes). How should this be handled? Truncate (no round up) or add one to the integral part and display no decimal places?

I encountered this as well in my writeup on https://github.com/endless-sky/endless-sky/issues/6235 (search for "one final issue"), and I think my proposed solution handles both the subject of this PR and the rounding issues in a smooth way: on all codepaths, it rounds exactly once and does so prior to extracting any decimals.  What do you think of it?  (If you like it, feel free to use any or all of it.)
```
// Convert the given number to a string, with a reasonable number of decimal
// places. (This is primarily for displaying ship and outfit attributes.)
string Format::Number(double value)
{
	if(!value)
		return "0";
	
	string result;
	bool isNegative = (value < 0.);
	value = fabs(value);
	
	// Numbers <1000 get rounded to two decimal places.
	// Numbers 1000-10,000 get rounded to one decimal place.
	// Numbers over 10,000 get rounded to an integer.
	// Note that rounding a number <1000 may yield 1000; that's fine here.
	int integer;
	int tenths = 0;
	int hundredths = 0;
	if(value < 1000.)
	{
		int scaled_value = static_cast<int>(round(value * 100.));
		integer = scaled_value / 100;
		tenths = (scaled_value % 100) / 10;
		hundredths = scaled_value % 10;
	}
	else if(value < 10000.)
	{
		int scaled_value = static_cast<int>(round(value * 10.));
		integer = scaled_value / 10;
		tenths = scaled_value % 10;
	}
	else // >=10000
	{
		integer = static_cast<int>(round(value));
	}
	
	if(hundredths)
	{
		result += static_cast<char>('0' + hundredths);
	}
	if(tenths || hundredths)
	{
		result += static_cast<char>('0' + tenths);
		result += '.';
	}
	
	// Convert the number to a string, adding commas if needed.
	FormatInteger(integer, isNegative, result);
	return result;
}
```

warp-core:
That looks pretty good and largely behaves well but I'm encountering a few issues while testing it, some values don't round up as expected:
image
I'll have it dump all the incorrect lines to a file and put that here shortly.

warp-core:
[fails.txt](https://github.com/endless-sky/endless-sky/files/7124108/fails.txt)
The code I used:
[TestingEndlessSkyDecimalDisplayBug code.txt](https://github.com/endless-sky/endless-sky/files/7124111/TestingEndlessSkyDecimalDisplayBug.code.txt)

Strangely, I don't seem to see this behaviour when passing some of those values directly (comment out line 375 and choose manual test when running the application to try it yourself). I'm not sure what's going on there but that doesn't seem like a major issue anyway, certainly not worse than what we have now.
If you want to create a PR for your solution, I think you should (I'll close this one if you do).

tehhowch:
> I guess I'm changing the behavior to "round" rather than "truncate".  Any objections?

No objections. You'll need to update tests accordingly.

> Anyway, yeah, I plan to make a PR.  By the way, how do I run the tests defined in tests/src/text/test_format.cpp?

Linux: `scons test`
Windows: `scons test`, if you configure your environment variables like CI, otherwise open the codeblocks workspace file and run the tests target.
macOS: sorry, not supported. If you'd like to refactor the xcode project to enable the same workflow as Windows and Linux, that'd be great.

waterhouse:
Ok, I made a PR: https://github.com/endless-sky/endless-sky/pull/6240

warp-core:
Awesome!

PR #6226, opened 2021-09-03
Correct Heavy Rocket Pod Burst Count

Anarchist2:
Looks like the heavy rocket pod burst count also had its ammo count changed without the burst count being corrected.


tehhowch:
@Zitchas intentional or oversight?

Zitchas (approved):
Thanks for spotting this!

PR #6230, opened 2021-09-04
Mark dialogs as uninterruptible.

quyykk:
**Feature:** This PR implements the feature request detailed and discussed nowhere.

This PR is very simple: It marks dialog panels as uninterruptible. This means that dialogs are able to be closed using escape, which was also the original intent of the code. But this didn't work because the panel was "interruptible", meaning that the game treated the panel as a panel that can be suspended to show the menu (which doesn't really makes sense for a dialog).

As a heavy user of the escape key, this has been annoying me since forever, since dialogs everywhere are always closeable using escape.


petervdmeer (approved):

tehhowch:
Should this apply to mission offer dialogs, too? I feel there are probably some who appreciate being able to save a snapshot at that moment.

quyykk:
No. It's just a dialog (not a conversation). IMO if someone wants to accept the mission they accept it and then exit the game. From a UI perspective you generally can't save when a dialog is open so I really don't think anybody is using that feature. Also that feature doesn't exist. I just tested this with a drug running mission dialog - exiting the game doesn't accept the mission and so when loading the save again the mission is gone.

oo13:
We can escape a spaceport mission and load the game to go back before offering it.

quyykk:
> Should this apply to mission offer dialogs, too? I feel there are probably some who appreciate being able to save a snapshot at that moment.

Done. I tested this with the spaceport drug running mission. Pressing escape correctly goes to the main menu instead of closing the dialog.

PR #6233, opened 2021-09-06
Added display ID setting.

RPBCACUEAIIBH:
**Feature:** Not sure if it was requested or not, I just fixed an annoyance... (having to drag the window over each time to the other display...)

## Feature Details
You can now specify the display on which you want the game to launch... Until now, it launched on display 0 regardless.

## Usage Examples
~~Go to preferences, and click on Display ID to change it. Changes will take place at next start, so exit and start again.~~
On @tehhowch's request this setting is hidden and automatic, it should now open on whichever display you last closed it.

## Testing Done
I only tested it on my dual display setup under ubuntu linux. It seems to work perfectly to me, but I advise caution...

## Performance Impact
I guess it adds couple of miocroseconds at start. ;)
...and now also a few microseconds before exit. ;)

### DIFF
--- a/source/GameWindow.cpp
+++ b/source/GameWindow.cpp
@@ -91,9 +91,17 @@ bool GameWindow::Init()
 		return false;
 	}
 	
+	/* This is a failsafe in case a display got removed.
+	Can't check for SDL_GetNumVideoDisplays() inside Screen::SetDisplayID()
+	because SDL_Init(SDL_INIT_VIDEO) is called here in GameWindow::Init()
+	after Preferences::Load() in main and SDL_GetNumVideoDisplays()
+	always returns 0. */
+	if (Screen::GetDisplayID() >= SDL_GetNumVideoDisplays())
+		Screen::SetDisplayID(0);
+	
 	// Get details about the current display.
 	SDL_DisplayMode mode;
-	if(SDL_GetCurrentDisplayMode(0, &mode))	{	
+	if(SDL_GetCurrentDisplayMode(Screen::GetDisplayID(), &mode))	{	
 		ExitWithError("Unable to query monitor resolution!");
 		return false;
 	}
@@ -121,6 +129,14 @@ bool GameWindow::Init()
 		windowHeight = min(windowHeight, Screen::RawHeight());
 	}
 	
+	// Get window bounds by ID...
+	SDL_Rect gameWindow;
+	SDL_GetDisplayBounds (Screen::GetDisplayID (), &gameWindow);
+	gameWindow.x += (gameWindow.w - windowWidth) / 2; // Center horizontally
+	gameWindow.y += (gameWindow.h - windowHeight) / 2; // Center vertically
+	gameWindow.w = windowWidth;
+	gameWindow.h = windowHeight;
+	
 	// Settings that must be declared before the window creation.
 	Uint32 flags = SDL_WINDOW_OPENGL | SDL_WINDOW_RESIZABLE | SDL_WINDOW_ALLOW_HIGHDPI;
 
@@ -130,8 +146,8 @@ bool GameWindow::Init()
 		flags |= SDL_WINDOW_MAXIMIZED;
 	
 	// The main window spawns visibly at this point.
-	mainWindow = SDL_CreateWindow("Endless Sky", SDL_WINDOWPOS_UNDEFINED, 
-		SDL_WINDOWPOS_UNDEFINED, windowWidth, windowHeight, flags);
+	mainWindow = SDL_CreateWindow("Endless Sky", gameWindow.x, 
+		gameWindow.y, gameWindow.w, gameWindow.h, flags);
 		
 	if(!mainWindow){
 		ExitWithError("Unable to create window!");
--- a/source/GameWindow.cpp
+++ b/source/GameWindow.cpp
@@ -418,6 +418,25 @@ void GameWindow::ToggleFullscreen()
 
 
 
+int GameWindow::CurrentDisplayID()
+{
+	int posX = 0, posY = 0, ID = 0;
+	SDL_GetWindowPosition(mainWindow, &posX, &posY);
+	for(int i = 0; i < SDL_GetNumVideoDisplays(); i++)
+	{
+		SDL_Rect currentDisplay;
+		SDL_GetDisplayBounds(i, &currentDisplay);
+		if(posX >= currentDisplay.x && posX < currentDisplay.w && posY >= currentDisplay.y && posY < currentDisplay.h)
+		{
+			ID = i;
+			break;
+		}
+	}
+	return ID;
+}
+
+
+
 bool GameWindow::HasSwizzle()
 {
 	return hasSwizzle;
--- a/source/GameWindow.cpp
+++ b/source/GameWindow.cpp
@@ -91,17 +91,17 @@ bool GameWindow::Init()
 		return false;
 	}
 	
-	/* This is a failsafe in case a display got removed.
-	Can't check for SDL_GetNumVideoDisplays() inside Screen::SetDisplayID()
-	because SDL_Init(SDL_INIT_VIDEO) is called here in GameWindow::Init()
-	after Preferences::Load() in main and SDL_GetNumVideoDisplays()
-	always returns 0. */
-	if (Screen::GetDisplayID() >= SDL_GetNumVideoDisplays())
+	// This is a failsafe in case a display got removed.
+	// Can't check for SDL_GetNumVideoDisplays() inside Screen::SetDisplayID()
+	// because SDL_Init(SDL_INIT_VIDEO) is called here in GameWindow::Init()
+	// after Preferences::Load() in main and SDL_GetNumVideoDisplays()
+	// always returns 0.
+	if(Screen::GetDisplayID() >= SDL_GetNumVideoDisplays())
 		Screen::SetDisplayID(0);
 	
 	// Get details about the current display.
 	SDL_DisplayMode mode;
-	if(SDL_GetCurrentDisplayMode(Screen::GetDisplayID(), &mode))	{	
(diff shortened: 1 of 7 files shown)


source/GameWindow.cpp
```diff
@@ -91,9 +91,17 @@ bool GameWindow::Init()
 		return false;
 	}
 	
+	/* This is a failsafe in case a display got removed.
```
tehhowch, line 94: No block comments

source/GameWindow.cpp
```diff
@@ -121,6 +129,14 @@ bool GameWindow::Init()
 		windowHeight = min(windowHeight, Screen::RawHeight());
 	}
 	
+	// Get window bounds by ID...
+	SDL_Rect gameWindow;
+	SDL_GetDisplayBounds (Screen::GetDisplayID (), &gameWindow);
+	gameWindow.x += (gameWindow.w - windowWidth) / 2; // Center horizontally
+	gameWindow.y += (gameWindow.h - windowHeight) / 2; // Center vertically
```
tehhowch, line 136: No inline comments.

source/PreferencesPanel.cpp
```diff
@@ -57,6 +57,7 @@ namespace {
 	const string SCROLL_SPEED = "Scroll speed";
 	const string FIGHTER_REPAIR = "Repair fighters in";
 	const string SHIP_OUTLINES = "Ship outlines in shops";
+	const string DISPLAY_ID = "Display ID (Nees restart.)";
```
tehhowch, line 60: The last used display can be a "hidden" preference, rather than one that is important enough to be shown to users in the UI

source/GameWindow.cpp
```diff
@@ -91,9 +91,17 @@ bool GameWindow::Init()
 		return false;
 	}
 	
+	/* This is a failsafe in case a display got removed.
+	Can't check for SDL_GetNumVideoDisplays() inside Screen::SetDisplayID()
+	because SDL_Init(SDL_INIT_VIDEO) is called here in GameWindow::Init()
+	after Preferences::Load() in main and SDL_GetNumVideoDisplays()
+	always returns 0. */
+	if (Screen::GetDisplayID() >= SDL_GetNumVideoDisplays())
```
tehhowch, line 99:
```suggestion
	if(Screen::GetDisplayID() >= SDL_GetNumVideoDisplays())
```
(Style nit)

source/GameWindow.cpp
```diff
@@ -121,6 +129,14 @@ bool GameWindow::Init()
 		windowHeight = min(windowHeight, Screen::RawHeight());
 	}
 	
+	// Get window bounds by ID...
+	SDL_Rect gameWindow;
+	SDL_GetDisplayBounds (Screen::GetDisplayID (), &gameWindow);
```
tehhowch, line 134:
```suggestion
	SDL_GetDisplayBounds(Screen::GetDisplayID(), &gameWindow);
```

source/GameWindow.cpp
```diff
@@ -91,9 +91,17 @@ bool GameWindow::Init()
...
+	because SDL_Init(SDL_INIT_VIDEO) is called here in GameWindow::Init()
+	after Preferences::Load() in main and SDL_GetNumVideoDisplays()
+	always returns 0. */
+	if (Screen::GetDisplayID() >= SDL_GetNumVideoDisplays())
+		Screen::SetDisplayID(0);
+	
 	// Get details about the current display.
 	SDL_DisplayMode mode;
-	if(SDL_GetCurrentDisplayMode(0, &mode))	{	
+	if(SDL_GetCurrentDisplayMode(Screen::GetDisplayID(), &mode))	{
```
tehhowch, line 104:
```suggestion
	if(SDL_GetCurrentDisplayMode(Screen::GetDisplayID(), &mode)) {	
```

source/GameWindow.cpp
```diff
@@ -130,8 +146,8 @@ bool GameWindow::Init()
 		flags |= SDL_WINDOW_MAXIMIZED;
 	
 	// The main window spawns visibly at this point.
-	mainWindow = SDL_CreateWindow("Endless Sky", SDL_WINDOWPOS_UNDEFINED, 
-		SDL_WINDOWPOS_UNDEFINED, windowWidth, windowHeight, flags);
+	mainWindow = SDL_CreateWindow("Endless Sky", gameWindow.x, 
+		gameWindow.y, gameWindow.w, gameWindow.h, flags);
```
tehhowch, line 150: Does sending windowpos_undefined for x and y not render in the center of the screen when using a non-primary display?

tehhowch (changes requested):
Thanks for the PR!

source/GameWindow.cpp
RPBCACUEAIIBH: It does, but that assumes display 0.
For example if you have 2x1920x1080 resolution diplays side by side, the x coordinate of display 1 content starts at 1920 when full screen, and even higher if not full screen. hance the gameDisplay.x += rather then gameDisplay.x =
Same for y if you have displays stacked vertically.

source/PreferencesPanel.cpp
RPBCACUEAIIBH: Not sure what do you mean by hidden preference... Can you point me to an example?

MCOfficer:
```suggestion
	const string DISPLAY_ID = "Display ID (Needs restart)";
```

Amazinite: > Not sure what do you mean by hidden preference... Can you point me to an example?

Not all preferences are visible from the in-game preferences menu. Some preferences must be changed by going to the preferences.txt file. For example, the window size that the game starts at is a preference not visible in game, meaning that if you wanted to manually change it (without opening the game and just dragging the edges of the window, that is) then you'd need to go to the preferences.txt file to do so.

And I agree with tehhowch's suggestion that the display ID should be a hidden preference. The game should remember what display the game was last on when it was closed, and therefore reopen in the same display, but you shouldn't need to be able to change the display that the game opens in from the in-game preferences menu; the method of choosing the display ID preference should be simply closing the game in the display you want the game to reopen in, just as choosing the window size is a matter of closing the game at the window size you want.

tehhowch: Yep.

It'll require reacting to window move events and setting the display ID in the background, rather than by a field in the preferences panel.

RPBCACUEAIIBH: Ok, that is not simple, cause this does not get the display ID, it kinda starts on display 0 by default and you can change it in settings. Maybe SDL_GetDisplayBounds() can be used to find out which display I'm on before exit...

RPBCACUEAIIBH: Done!

source/GameWindow.cpp
tehhowch: Seems like we can instead set `SDL_WINDOWPOS_CENTERED` and avoid the extra code. Or does that also assume display 0, or does that stretch across displays?

RPBCACUEAIIBH: hmm... I don't know, but I'll get back with an answer.

source/GameWindow.cpp
```diff
@@ -402,6 +418,25 @@ void GameWindow::ToggleFullscreen()
...
+	{
+		SDL_Rect currentDisplay;
+		SDL_GetDisplayBounds(i, &currentDisplay);
+		if(posX >= currentDisplay.x && posX < currentDisplay.w && posY >= currentDisplay.y && posY < currentDisplay.h)
+		{
+			ID = i;
+			break;
+		}
+	}
+	return ID;
```
tehhowch, line 435: Does it make sense to react to the "window moved" event (which sends the new XY as event data) to set the new display event?
We'd cache the bounds of each display on startup so we can simply update.
We could then also re-use the same position the next time the game launches, rather than always recentering.

source/GameWindow.cpp
RPBCACUEAIIBH: Yup, that also only works for display 0.

source/GameWindow.cpp
RPBCACUEAIIBH: Well there's a failsafe, which checks if you actually have the display on which it has to launch the window, and if it's removed it would default back to display 0. That wouldn't work if you just launch on XY coordinates where it was closed...

tehhowch: Triggering the failsafe should reset the coordinates

RPBCACUEAIIBH: Yes but how do you trigger this if you only have coordinates?
Screen::GetDisplayID() >= SDL_GetNumVideoDisplays()

tehhowch: That's why I mention caching the bounds of the available displays. We will still need the display ID in order to get the correct display mode and high-dpi information. The event handler would update the stored coordinates and check if the display changed. The display change handler would update the display ID, high DPI state, and window bounds (and any other display state I'm not thinking of).

RPBCACUEAIIBH: Yeah, I think I'll leave that to you guys. It's getting late, and tomorrow I have other things to do... I kinda just fixed what annoyed me, and don't really want to go down that rabbit hole. Feel free to do whatever you want with the code. :)

petervdmeer:
This is a nice comfort improvement for people with multiple displays and their primary (gaming) display not being at Display 0. The majority of our playerbase will however not have a primary display at non-zero, so this PR only benefits a very small subset of our playerbase.

I also read in the comments on this PR that this PR is a bit stuck; the original author indicates that it's up to "us guys" to proceed on this PR, and I also know that "us guys" still have a lot of other PRs to address (which will benefit a larger set of the playerbase).

Closing this PR; I don't see this going to completion anytime soon.

@RPBCACUEAIIBH Thanks for contributing, and please feel free to re-open the PR (or another PR) if you like to contribute again.

PR #6239, opened 2021-09-08
New Human Weapons

Brick63:
**New Weapons For humans**

The Addition of 2 New weapons to be mainly used by the Navy

The Broadsider, a Dual railgun turret that big and Bulky
broadsider@2x

And the Humble Barrage Launcher, that is small and dangerous
barrage launcher@2x

Since this is my first in a while, I dont know anything for the balance of these things, and probably messed something up, but i want to make these available for use by people that can do things better than me in terms of balance and missions. 

Edit; Assets

[Barrage Asset packet.zip](https://github.com/endless-sky/endless-sky/files/7189952/Barrage.Asset.packet.zip)

[Broadsider Assets.zip](https://github.com/endless-sky/endless-sky/files/7190167/Broadsider.Assets.zip)


data/human/weapons.txt
```diff
@@ -200,6 +200,71 @@ effect "beam laser impact"
...
+		"turret turn" 1.0
+		"velocity" 28
+		"lifetime" 1
+		"reload" 50
+		"firing energy" 198
+		"firing heat" 93
+		"firing force" 11
+		"hit force" 44
+		"submunition" bslug 1
+		description "As part of the Navy's 'New Age Initiative,' the team at Geminus has developed larger and more powerful weaponry to function on their new ships. The Broadsider is one such weapon created for the Battleship and meant to hit harder than anything humanity has to offer."
```
Terin, line 250:
```suggestion
	description "As part of the Navy's 'New Age Initiative,' the team at Geminus has developed larger and more powerful weaponry to function on their new ships. The Broadsider is one such weapon created for the Battleship and meant to hit harder than anything humanity has to offer."
```
Gets the checks to pass.

Brick63: THX M8E

Pointedstick:
Do we have ships for this yet? Feels kinda random TBH

Zitchas:
He does indeed have at least a ship or two from what I've seen on the discord. 

While I can't speak to any plans, my opinion is that these weapons (and probably at least one of the ships I've seen) would most likely be human unlocks sometime after the automatons run amok across northern human space. In other words, *after* the main  human campaign. So, probably sometime during the Wanderer campaign. Might make human space a touch more interesting to have new things happen while the player is off helping the Wanderers.

Pointedstick:
Perhaps. I think that stuff needs to be in place first. Adding these weapons to the game's data files without anything being ready to use them does not make sense IMO.

Brick63:
I was Adding these so they can be used  by others who are better than me at making missions and balancing stats, taking a similar role to the unused Hai Ships and outfits, though to a lesser extent. I wouldn't be against a mission chain being added to this PR, But I know I'm not the one who should create it.

Pointedstick:
Then it doesn't need to be a PR, right? You can provide these assets to someone who makes that mission. Is anyone working on that? Can you find someone?

data/human/weapons.txt
```diff
@@ -1300,6 +1365,65 @@ effect "nuke residue slow"
...
+	ammo "Barrage Missile"
+	description "A safe Storage for the Barrage Missiles."
+
+outfit "Barrage Launcher"
+	category "Secondary Weapons"
+	cost 5000
+	thumbnail "outfit/barrage launcher"
+	"mass" 3
+	"outfit space" -3
+	"weapon capacity" -3
```
W1zrad, line 1392: Outside of Zitchas's new fighter pods, the smallest missile launcher is 10 tons. I get that this is supposed to be compact, but only 3 tons seems excessively small. Could we make it a bit larger? Totally open to suggestions on a new number though.

Zitchas: I think that might be about the same size as the javelin mini pods... (The fighter sized one that only holds 40 javelins)

W1zrad: Ah, gotcha. In that case I would suggest adding "Pod" to the name for the sake of clarity, but it's not a huge thing. Feel free to resolve

W1zrad:
Did a little testing and noticed something: on small ships (especially when using more than one launcher), these missile sprites feel quite large, especially compared to others missile sprites already in use. Here's a few pics for comparison (fired from a Berserker):

Javelins:
Screenshot (12)

Heavy:
Screenshot (11)

Barrage:
Screenshot (10)

The [trail, exhaust, fire] (what's that called? can't remember) behind it is especially large, but the rocket itself is also pretty big. Was it intentional or an oversight? If intentional, what was your thought process?

Brick63:
Definitely not intentional, I forget what I used as a size reference originally. But I'll fix it here soon.

data/human/weapons.txt
Brick63: Well it's made to be pod sized, so I don't think pod is required in the name, especially since there is no larger variants, at least not currently

Zitchas:
I'm currently working on an integration event & mission to introduce and add in the Broadsider. 

In the meantime, I'd like just like to comment that the sound effect for it is a bit high-pitched. This feels (and the name sounds) like the sort of weapon that should have more of a deep artillery fire sort of impact when it is being shot.

In terms of balance, I'd be inclined to slow it down a little bit. Just slightly, but in all three areas of firing speed, projectile speed, and turret turning speed. 

And a comment for Derpy or whoever really digs into the weapon balance side of things: The integration stuff I'm writing is placing it as an unlock tied to the end of the Sestor invasion of human space, which is part of Wanderer Chapter 2; and is framed in terms of "the Navy's been working on a long-range artillery for quite a while and making some progress, but the Sestor invasion has provided them with motivation to work harder and get something out the door now; Significantly boosted funding in response to the newly discovered "alien threat;" and a whole pile of salvaged alien tech to inspire new ideas and hints about possibilities. (specifically, it uses a timed trigger that starts the Broadsider reveal sequence 30-120 days after the end of the Sestor invasion)

So I see this as fitting into the overall balance as a late-game weapon that - while it shouldn't be able to compete with the sunbeam or moonbeam in terms of shear damage output, it should really shine in "long range anti-capital" warfare. This is a weapon for Carriers, Dreadnoughts, and humanity's future heavy warships (battleship?). These are ships that should *not* be racing to close with enemies or chasing down little ships. Rather, these are ships that should be laying waste to the enemy's capital ships from outside of missile range.

kestrel1110:
I just wish there was a gun version of the broadsider. Would be cool to have a navy ship with those on fixed guns at 90 degrees like old sailing ships...

Amazinite (commented):
I'm not particularly interested in these new weapons, at least at the moment.

I recall talking a fair bit about the Barrage Launcher a while back now, but at this point the niche that the launcher could have filled is basically taken up by the mini launcher pods from #5793, so there isn't much of a place for it to fit in now.

As for the Broadsider, in terms of the niche it would fill it certainly has room to fit in, but the problem comes from where in the story it fits in. I don't really see this as being a weapon available at game start, nor do I see it as something that would be developed during the FW campaign. The best place I see something like this fitting in is after the FW campaign, particularly after the Sestor invasion, which Zitchas actually went down the route of with his PR that implements this (#6268). The problem there though is that I wouldn't see just this one weapon coming out as a result of the Sestor invasion; I and others have tossed around the idea of the Navy doing some massive upgrades to its arsenal after a second alien invasion of human space. So insofar as there are plans to do a big arsenal improvement of the Navy, particularly as a part of a big storyline, I'd rather hold off on something like this for now and instead tackle the entire arsenal upgrade at once as opposed to merging something early and basically forcing any future plans to deal with this. (Zitchas has proposed merging it now as a placeholder until we get a bigger story set up, but I'm not really a fan of the idea of having "placeholder story.") So given that I don't really see these weapons fitting in right now, I'll be closing this and the two mission PRs.

Aside from that, the Broadsider could use some visual work. It looks very top/front heavy and like it could fall over. I'd set the barrels a little further back on the mount so that it looks like the center of gravity is more over the middle.

Zitchas:
To be honest, my intention behind the idea of "placeholder story" was not so much to have the story itself be placeholder, but rather that mission's pre-requisites be a placeholder. So for now it triggers after the second invasion, and when the appropriate "main campaign" story is developped, then the missions' prerequisite are swapped to be an offshoot of an appropriate mission during the sequence. So the actual mission never changes, just its timing is adjusted as needed to line up with the missions.

Issue #6235, opened 2021-09-07
Bugs in Format::Number mess up displayed ship stats


waterhouse:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

In the attached pilot file, on the most recent git version, you can see the turning stat of my Blackbird flagship being displayed as "95.35 / 1066"—that is, with no cargo, the ship should have a turn rate of 1066.  (Screenshot in the screenshots section of the report.)  That number is absurd; it's probably at least missing a decimal point after the 106.

I encountered the problem in the current git version, 439387647c27869017ef357c4097d2907a7d6d21, and confirmed that the problem was introduced in commit 1873fd9ee9a9ad6a4fd210fe03f8d5a6fc7abc25 (PR: https://github.com/endless-sky/endless-sky/pull/4304 ).  On the prior commit, fcb4a9700bf1a4abb9d9e1e94440e9b4598219d6, the turn rate is displayed as "95.4 / 106.1".  I therefore suspect that the 1066 is supposed to be 106.06.

If you sell the KP-6 Photovoltaic Arrays, you can see it go to "99.36 / 1115" and then to "103.73 / 116.54".
If you then buy another LP072a Battery Pack, it goes to "1007 / 111.93".  I'd guess the 1007 should be 100.07.  (Buying another yields "96.65 / 107.68", and 100.07 is close to the middle, although it's not the exact arithmetic mean.)

While I'm at it: With 1 battery pack, no photovoltaic arrays, and 1 laser turret, the acceleration is displayed as "243.2 / 274.65", a different number of digits displayed.  Maybe it's cutting off the trailing 0 from 243.20; I guess that's intentional, just mentioning it.

This could probably be unit tested (alongside other tests introduced by 1873fd9ee9a9ad6a4fd210fe03f8d5a6fc7abc25) in tests/src/text/test_format.cpp.  I spent a few minutes and didn't manage to figure out how to run those tests.

If we look at the definition of Format::Number:
```
string Format::Number(double value)
...
        // Only show decimal places for numbers between +/-10'000.
        double decimal = modf(value, &value);
        if(decimal && value < 10000)
        {
                int tenths = static_cast<int>(decimal * 10.);
                // Values up to 1000 may have two decimal places.
                if(value < 1000)
                {
                        decimal *= 10.;
                        decimal -= static_cast<int>(decimal);
                        // Fix any floating point inaccuracy.
                        decimal = round(decimal * 100.) / 100.;
                        
[...]

PR #6240, opened 2021-09-09
Fix decimal display in ship stat screens (issue #6235)

waterhouse:
**Bugfix:** This PR addresses issue #6235

## Fix Details
Change Format::Number to (a) round exactly once at the start, and (b) correctly insert the decimal place and tenths digit in the case where the tenths is zero but the hundredths is nonzero.

## Testing Done
Opening the save file and checking that the numbers on the Blackbird look reasonable (particularly the turning speed); some playing with the fix; and the unit tests in tests/src/text/test_format.cpp.

## Save File
This save file can be used to verify the bugfix. The bug will occur when using anything from 1873fd9ee9a9ad6a4fd210fe03f8d5a6fc7abc25 to the current commit (e.g. 439387647c27869017ef357c4097d2907a7d6d21 ) , and will not occur when using this branch's build.

[JC Penney copy.txt](https://github.com/endless-sky/endless-sky/files/7125535/JC.Penney.copy.txt)

### DIFF
--- a/source/text/Format.cpp
+++ b/source/text/Format.cpp
@@ -130,32 +130,43 @@ string Format::Number(double value)
 	bool isNegative = (value < 0.);
 	value = fabs(value);
 	
-	// Only show decimal places for numbers between +/-10'000.
-	double decimal = modf(value, &value);
-	if(decimal && value < 10000)
+	// Numbers <1000 get rounded to two decimal places.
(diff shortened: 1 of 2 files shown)


tehhowch:
I went a slightly different route to fix the issue: 5e89884bce9b04c890bd1f43b1d20db04243846e

Issue #6097, opened 2021-07-02
Functionality for Derelict Missions


Terin:
**Is your feature request related to a problem? Please describe.**
A large trope for space exploration stories is coming across derelict ships. Either a merchant that was left for dead, a strangely undisturbed hulk, or something much weirder. Our predecessor, EV Nova, had support for these, and it was fun to not know if a ship you were boarding was someone in distress, or a pirate trap.

Endless Sky is so close to supporting this, with a whole data file related to boarding missions, but every one of those relies on a ship that's disabled by pirates or the like. We cannot jump into a system to find a disabled ship, unless it is directly part of a mission.

**Related Issue Links**
#614 - Brief mention of the concept, although in a very specific context.
#4122 - Rescue missions on derelicts, but only when prompted by a mission beforehand.

**Describe the solution you'd like**
Three solutions come to mind:

1. Allow fleets with the `derelict` personality to naturally spawn in the system disabled, but only on takeoff/jumping into a system. This would let us populate areas with different chances of generic derelicts based on flavor for the universe, or have a mission temporarily add a special derelict fleet with a high chance to a specific system. Boarding missions could then be based off of these ships having a specific government and ship category.
2. A `reveal` mission action, that turns a mission from invisible to visible. That way, you can pick up an invisible mission which spawns an `npc assist` ship at some system, and once you board it, it will reveal the mission to take survivors back to a given location or such. However, you won't be able to pick up actual passengers/cargo from these, so it'd be fairly hacky.
3. Some kind of override attribute that allows a boarding mission to spawn on mission NPC ships. That way, an invisible mission can spawn a specific ship of the derelict government at a random location, and if you come across it, one of these override missions from the Derelict government can trigger.

I can start a PR for one of these solutions, if we can come to an agreement for which is the most viable/least disruptive.

Galaucus:
Regarding option 2, what's preventing us from allowing boarding missions to give cargo or passengers?

Terin:
The issue is that option 2 isn't a boarding mission, it's an npc assist mission. As far as I'm aware, you can't have a mission that directs you to a disabled ship, and retrieve actual cargo/passengers off it. That can only come from boarding missions.

quyykk:
> most viable/least disruptive

that's definitely option 1 imo. I do have a question though:

Should they spawn more on the outskirts of the system, or randomly across the whole system or something else? Should this be changeable?

Galaucus:
[...]

PR #6241, opened 2021-09-10
feat(mechanics): Allow fleets to spawn derelicts

Terin:
**Feature:** This PR implements the feature request detailed and discussed in issue #6097

## Feature Details
As said in the original issue, missions in Endless Sky have a lot of power to provide interesting situations while boarding disabled ships. However, this is limited only to naturally spawned ships, which were either disabled by yourself or an enemy of theirs.

This PR lets naturally spawned fleets make use of the `derelict` personality. If a ship from that fleet spawns as the player enters from/takes off into that system, it will be disabled and stationary. Then, no other ships from that fleet will be allowed to jump in.

Content creators can then have events which add a derelict fleet to a given system, of a specific government type that are the only ones eligible for specific boarding missions. We may be able to add derelicts to human space broadly, as well, for more generic derelict boarding missions.

## UI Screenshots
N/A

## Usage Examples
```
fleet "Civilian Derelicts"
	government "Derelict"
	personality
		derelict
	variant 20
		...

system ...
	fleet "Civilian Derelicts" 10000
```

## Testing Done
Having derelict fleets be built into a system, having them added with an event, and confirmed that derelict NPC still function without issue.

## Performance Impact
N/A

### DIFF
--- a/source/Fleet.cpp
+++ b/source/Fleet.cpp
@@ -300,7 +300,7 @@ const Government *Fleet::GetGovernment() const
 // Choose a fleet to be created during flight, and have it enter the system via jump or planetary departure.
 void Fleet::Enter(const System &system, list<shared_ptr<Ship>> &ships, const Planet *planet) const
 {
-	if(!total || variants.empty())
+	if(!total || variants.empty() || personality.IsDerelict())
 		return;
 	
 	// Pick a fleet variant to instantiate.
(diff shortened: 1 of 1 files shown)


Pointedstick (approved):
LGTM

tehhowch:
This is technically a bug fix - you can set `personality derelict` on fleets already, it just doesn't do anything outside of NPCs. 😉

PR #6242, opened 2021-09-10
Fix minor typo

nathan-b:
This is just a very minor text correction.

## PR Checklist
 - [x] I decline to claim copyright of any assets produced or modified


MCOfficer (approved):
I'm not sure if the old version was wrong; you can describe a condition by means of a verb, similar to "the sooner we have these ships working". But if it sounds wrong to some people and the alternative has no downside, why not change it?

nathan-b:
@MCOfficer  "The sooner we have these ships functioning" would also have been fine. "The sooner these ships can function" sounds a bit less natural to my ears, but would also work. If you prefer either of these versions, I'm fine to modify the PR.

Zitchas (approved):
I think this is a good fix, and probably the most minimal change that gets it sounding better.

PR #6244, opened 2021-09-11
[Combat/Fighters] Expanding System Arrival Distance

Zitchas:
**Content (Systems & Map)**

## Summary
- Arrival Distance in all systems set to 150 farther out than either the belt or habitable zone, whichever is farther.
- Note: There are some exceptions. A small number of systems have belts or habitable zones in the multiple thousands. These were hand-set to 150 outside the furthest orbit or other arbitrary distances that felt appropriate.

List of all systems with arrival distances over 4k:
Line | Range
---|---
Line9710 | 8790 (Esix, EW)
Line7226 | 7420 (Convector, EW)
Line16225 | 6305 (Meblumem, Coalition)
Line10657 | 5575 (Feroteri, Korath)
Line3193 | 5480 (Almaaz, North Pirate)
Line21446 | 5000 (Sadr, South Uninhabited)
Line8818 | 4470 (Edusa, EW)
Line5505 | 4231 (Betelgeuse, North Republic
Line25393 | 4171 ("Wah Oh", Hai)
Line19013 | 4070 (Persian, Syndicate, uninhabited)
Line21381 | 4025 (Sadr, South, Uninhabited)

## Expanded
These changes were originally implemented in #5772 and generally received favourably in conjunction with the changes to ship mass, agility, and the improvements to fighters.

The purpose of this is change is threefold:
1) Have ships arrive farther out, such that there is meaningful amounts of *space* travel between systems, not just jump+landing sequence.
2) Have opposing fleets arrive sufficiently far out that there is time for anticipation before they actually engage with each other. This also means that there is time for fighters to race out to engage before the main fleets. Likewise, it means that long range options are meaningful, and there is more combat going on before the inevitable knife-fight develops. 
3) Have distinguishable flight paths show up as players can see the paths of ships arriving from various directions which allows them to actually see where ships are coming from, and thus visually show where the adjoining systems are.

While I would be thrilled to see this merged in the near-term, I expect that it is more likely to happen in conjunction with one or more related changes, including Amazinite's mass changes and fighter speed improvements.

With the advent of derelict ship missions and other mechanic improvements, we are moving into an era of Endless Sky where we can (and hopefully, will) have much more action happening in space. It isn't just land, get job, take off, fight, go to destination system, land, repeat. Each system doesn't have to be just a single combat node, but rather a system where ships arrive from various directions and may or may not engage with other ships within the system. The player should be flying around investigating things within systems. Likewise, having an arrival distance farther out makes smuggling a little bit more possible since scanning patrol ships will have a bit farther to fly to get to the player. All these open up new possibilities and options for expanding the depth of the game.

## Save File
This change can be experience by simply playing the game. No special save required.

## PR Checklist
  
--------------------------------------------------------

 @Amazinite (In discord 2022-Feb-22)
> I think the thing with small PRs is that there needs to be a clear aim with where to go next.
> If a small PR makes a random balance change in a vacuum without regard to other changes being discussed despite it being relevant to those other changes, that's when I'd say it's "not doing enough."

Where do we go from here & stuff being discussed
- Stand Of AI improvements
- Formations
- System Expansion
- System Enlargement
- Departure distances and angles

Expanding the arrival distances of systems has, for its core goal, establishing a degree of separation between ships. Instead of having everything arriving directly into the system core, ships arrive at the periphery. 

**Formations:**
As implemented in https://github.com/endless-sky/endless-sky/pull/4471 , (and a partial ready-for-merging implementation here: https://github.com/endless-sky/endless-sky/pull/6570 ) I find that formations complement these Arrival Distances really well. Having fleets moving together in formation - especially over the longer distances created by this - look really good. I consider these to be a very strong pairing that go well together. Extensive testing of arrival distances combined with Formations yielded a lot of positive comments, and one minor bug: Namely that arriving fleets needed to have a tiny pause to gather before heading into the system, otherwise they don't actually collect. I believe that was resovled, however, and thereafter it looked beautiful seeing the larger merchant fleets ponderously approaching planets with a couple of interceptor escorts. The one caveat is that most fleets in ES aren't big enough to make most of the fleet formations look good. That being said, the "Vee" formation and a few of the other simple formations work *very* nicely.

**Ship differentiation:**
As such, there is space for different ship configurations to shine, including ships optimized for long-range support. To this end, it supports efforts to differentiate ship classes by allowing ships to be sub-optimal at close-quarters-combat by ensuring that most battles will pass through extreme, long, and medium ranges first. Some ships will still shine in the CQC, whereas others (like missile boats) can dominate at long ranges. We could even introduce artillery style weaponry (slow firing, medium speed, heavy hitting, extremely slow turn speeds for turrets) and follow EVN's much-admired route with the Auroran Carriers that were primarily equipped with extreme range artillery to support their flocks of fighters. While ineffective against light ships, said artillery was extremely potent at decimating larger ships who had been pinnned down or distracted by their fighter wings & light warship escorts.

**Stand Off AI Improvements:**
The point in Stand-off tactics is to have ships that ignore their short range weapons (for instance, laser turrets that they have for anti-fighter defense) to stay at a safe and effective range for artillery or other long range weaponry. There's not much point in having AI that will prioritize its artillery like this if the ships never (or rarely) ever have the chance to use it because they start too close to the enemy for it to be relevant. AI that makes more strategic use of distance is better served by (and more useful to have) having ships operating at larger varieties of ranges.

Basically speaking, arrival distances will make the current lack of a good stand-off AI more visible, and will make it much more valuable and impact-full once we do have it. Until that is implemented, alternative measures will need to be taken. The primary three options are a)make ships that we want to be long-range ships *only* equipped with long range weapons. No backup weapons for anti-fighter or whatnot. b) Give their primary long range weapon have enough kick to counteract their forward thrust temporarily to keep them away from the fight. c) cripple their speed to the point that even though they are *trying* to join in the deathball and make use of their short range backup weapons, they just never (or at least, very rarely) ever actually get there.

**System Expansion:**
Some people have been talking about adding more planets to systems, once https://github.com/endless-sky/endless-sky/pull/6355 has been merged. This arrival distance will likely result in more streams of traffic from the edges of the system in towards the various planets instead of just having a general chaos of ships going every which way in the center. These streams of traffic is already one of the reasons why I think arrival distances are such a good benefit, as I find they make patterns of ships more visible.

**System Enlargement:**
There has been talk about just scaling up systems. VitalChip has mentioned, for instance, of having systems normally exceeding 10k in size. I'm not entirely sure how these arrival distance expansions. I suspect that the more the system is scaled up, the more this arrival distance is likely to be diluted. That being said, I'd say these are probably complimentary changes. A system scaling might even want to scale these arrival distances up at the same time.

**Departure distances & Angles:**
The flip side of arrival distances. Once we have recognizable incoming flows of traffic, it will likely make having outgoing hyperjump limits feel more reasonable. Likewise, hyperdrive angles would be much easier to keep track of if there are flows of traffic created by these arrival distances.


Terin:
The arrival distance alone felt like a massive improvement for making systems cluster into less of a death ball, and I'd really love to see it merged.

Zitchas:
These changes look even better with https://github.com/endless-sky/endless-sky/pull/4471 too, as that magnifies how the fleets look. (and it gives the player more of a chance to notice that they are, in fact, moving around in fleets.)

Pointedstick:
I gave this a whirl with a few of my pilots as well as a new game. Observations:

1. The galaxy feels more sparse in a lot of systems. Ships enter and exit off camera quite frequently.
2. The game is a bit slower. It takes longer to travel to planets and stations especially in multi-planet/station systems
3. The galaxy is now quite a bit safer. Pirates warp in far away from you, giving you the practical option to escape in the early game, if you can't or don't feel like fighting.
4. Escort jobs are now probably too easy, especially the ones where the majority of pirates warp in, as opposed to when they're waiting for you at in orbit at the source or destination planets. You can now pretty easily ignore the warping in pirates. They probably need to be adjusted so that their fleets are mostly or entirely composed of ships that are fast, or will become faster after other changes. The same will probably go for other similar missions and jobs where enemy ships warp in and chase you down. They've gotta work for it now, which means more care needs to be paid that they are equipped accordingly
5. Big fleet engagements (I test with some of the showpiece Free Worlds vs Republic battles) don't go that differently. Yes, the fleets spawn farther away, but they still thrust right into each others' faces within about 5 seconds. The net result is that long-range missiles have *slightly* more of an initial impact, and you have a *bit* more time to position yourself and your escorts, but in the end it just turns into a big brawl in the center of the system anyway. I think we will need AI and/or ship loadout changes here to make a significant difference in this issue.

This is not a negative review. However I feel like this change doesn't really stand alone and needs at least a few changes added to it to be merge-able. In particular, changes to the fleet compositions for missions where enemies warp in and chase you down. They become too easy to avoid with this PR in its current state.

Zireael07:
1,2,3 are to me very good things.

Zitchas:
@Pointedstick Thanks for testing it out and posting detailed thoughts about how it affects the game. I appreciate the detailed feedback. 

I agree with Zireael07 that 1,2, and 3 are, overall, good things. It is worth noting that people are working on creating asteroid belts and other localized hazards, which would make the system landscape more diverse internally, and not just between systems, and that would play very nicely with larger systems.

As for 4.... I find escort missions are trivial either way (except when the threat is already waiting in orbit). Having fighters with the speed boost from #6247 would help here, as we could use pursuing carriers to launch fighters that would race ahead to try to pin down fleeing ships. Which is, admittedly, a very appropriate and tactical use for fighters that I would like to see more of in-game.

5. I agree. This isn't a complete solution. Revamp does a bit better with all sorts of different pieces thrown in there, but the AI is really the centerpiece for making battles into something other than an in-your-face-brawl. But at least it does give us a few moments to appreciate the upcoming battle before we are in it. The final space battle from Serenity comes to mind with the two big fleets bearing down on each other before becoming a chaotic brawl. Right now, ES just has the brawl. Adding in system arrival distances gives us the few seconds for the dramatic approach of the two fleets.  AI, fighter speed changes, mass rebalances across the spectrum... Each of those will hopefully help differentiate things and reduce the brawl a bit.

Pointedstick:
I agree that overall the escort missions need to be impeoved even in their current state. But this PR makes that problem even worse. Let's fix it or else return to status-quo levels of problematicness in this PR. :) That should hopefully unblock it.

Zitchas:
OK, I'll see what I can do. 

This would be easier if we had bay outfits so we could do something like, say, an argosy (for carrying the loot) that happens to have a bay for one or two daggers. The daggers go out and pin down (maybe even disable?) the target merchant, then the argosy loots and runs. (I know that sounds overpowered given our current bay numbers, but I hope that in the future true carriers are all in the 10+ carried ships realm, so a freighter outfitting one or two bays for raiding (or self-defense) wouldn't be unreasonable or overpowered.

Pointedstick:
Perhaps there could even be a "light carrier" Modified Argosy variant in the future.

Zitchas:
On the plus side, expanded systems make for a very valid use-case for marauder engine variants, which, while fun to fly, have historically been considered much less useful than the more combat-focused variants.

Zitchas:
Just some thoughts for improving system combat.
For the death-ball:
- Probably one of the most significant changes we could do to spread things out is stripping close-quarter weapons off ships that really shouldn't be in CQC. Most notably, the carrier. Carriers should be *exclusively* armed with fighters, missiles, long-range artillery, and anti-missiles. (ideally we'd have some EVN-style automated anti-missile style anti-fighter turrets too, but that's a code change). And by "long range artillery" I mean things like the AMC that are slow-firing, slow-travel, and heavy-hitting. The sort of things that will decimate ships that it hits, but are effectively useless against anything that isn't slow as molasses or pinned down by fighters/tacklers.
- If Carriers and other support ships don't have any CQC weapons, I *think* the AI will be better about not pushing them straight into melee. 

For the escort missions, I can think of only one way to actually solve escort missions:
1. implement mission code changes that require the player to actually *escort* the ships. For instance, having the hiring ships follow the optimum route to the target, and have them set so that they **will not** hyperjump if the player isn't in the system with them. Thus, the player has to follow the escorted ships around instead of just zipping to the end and waiting for them to show up.

Improvements, however, could include:
- Implementing departure distances so that the player and escorted ships need to survive to the edge of the system in order to escape. No more hyperjumping directly from take-off. (this would make blockades actually a thing, too. It would make a lot of military campaign storylines make a lot more sense than they do currently)
- Implementing pirate raiding "blockades" into escort missions so that the player has to at least deal with a single fleet of pirates at the destination system while waiting for the escorted ships to arrive.

My ultimate thought, though, is this:
Compared to virtually all our other ways of making money, escort missions are the bottom of the barrel. We can make more money faster by commodity running, capturing, salvaging, mission running (particularly pushing up into the very large luxury passenger jobs)... Escort missions might be easy, but they're not that numerous, so it's pretty much one-off cash bonuses now and then; not something that really threatens the economics of the game.

Zitchas:
I also had a thought regarding the entry distance: What if, instead of having a fixed arrival distance of 1500, it was instead set to something more formulaic based on the system?

For instance, setting the arrival distance to whichever of "belt" or "habitable" distance is greater, +50?

For a lot of systems, this would be somewhere in the 1500-2000 range, but would impart a bit more variety to the galaxy.

I'm not going to do this by hand, though, so if we want this, someone will need to make a script to do it.

Zitchas:
Just a quick update on this: 1010Todd made a script based on something Terin did which does exactly what I talked about: Sets the arrival distance of the system to either habitable+50 or belt+50, whichever is greater, for every system in the game. 

As expected, most of the systems are in the 1400-2000 range, although there's more than a few outliers that are significantly shorter or larger than that. That being said, flying around testing it out, and the arrival distances seem pretty good. Maybe a bit on the short side, overall. 

The only problem that really needs resolving is that the script appends the arrival distance near the end of the system definition, just in front of the objects. So we need to figure out a way to sort it back into the right spot.

Zitchas:
Alright, I like the feel of the +150 version. Compared to the version people may have tested earlier, the systems in this version will be more varied. Perhaps most importantly, arrival should now almost always be just outside the orbit of the habitable planets, if any, or just outside the belt, whichever is further. We have virtually zero visually distinguishable belts in the game, but in testing this out on [Cognizance 2](https://github.com/endless-sky/endless-sky/pull/6392) which does have some very visible belts defined in some parts of the story, I found that it felt really meaningful to have my ship arrive from hyperspace just outside the asteroid belt. Makes sense, really. When careening blindly out of hyperspace I probably don't want to land in amongst the asteroids...

This definitely improves the feel of hyperspace arrival and makes it seem more connected to the system.

HTMLPhoton:
Yes please.
As somebody who is working on the AI, it's a pretty big issue that most ships just kinda start on top of each other, and having the AI distangling themselves just looks really weird. Increasing the amount of "space" within a system by increasing the arrival ranges get's a big thumbs up from me.

Zitchas:
@Pointedstick I haven't yet done changes to fleets for missions, as those are honestly more intimidating than the systems themselves.

That being said, it has come up (as mentioned by HTMLPhoton) that providing this "working distance" to start with provides a really valuable space for those working on the AI to improve things. 

Given the already-broken nature of the escort missions, I'll make some basic changes to push all the pirate risks to either the starting planet or destination planet; but I really think that holding this PR back in order to include changes to these other missions may actually be harming more valuable development. I really don't think the escort missions are sufficiently important to be a roadblock to getting better combat AI.

Zitchas:
OK, I just dug through a bunch of the escort missions (not all, but a bunch of the jobs in the Deep, Dirt Belt and frontiers. And I note that most of these missions actually have quite a few pirates being spawned. Most of them being spawned in the destination system. A lot of them have a mix including most or all of:
- Some spawned in the destination and tagged to stay there.
- Some spawned in the destination system as arriving there.
- Several spawned there and wandering.
- More rarely, spawned in the source system.

So the change that I was planning on making to these missions to make them a little more dangerous is actually already done (marking some of the fleets start in the destination system, and stay there). I don't think it is a good idea to put "staying" in the rest of them, because that would quickly shift the problem from "wandering pirates have trouble finding the prey" to "how are pirates managing to have an entire navy-rivalling fleet here to blockade this world?" Basically raising the threat from "trivial" to "major battle." Which I consider to be going from the too easy side of the scale up to the too-hard end.

For a few of the deep ones, they take the alternate approach of specifying a small number of pirates in most of the systems that the player might go to for the mission, and have them stay in each one. 

Basically, this survey through the escort jobs has left me more convinced than ever that there is no good fix possible from the content side. This is something where we really need to look at fixing the way escort jobs are handled, which is something for another PR.

This leaves us with two choices:
1. We accept that escort missions are broken, and merge this so that the AI people have some actual range to work with in their combat AI work; and we discuss how to make escort missions more meaningful.
2. This gets frozen until the undeterminate future when a solution is determined for the escort jobs is done.

I personally favor option 1, as escort jobs are currently a pretty low priority. They're not on anyone's "top money makers" list, they're not on anyone's "fun jobs to take or make" list either. (and given how unpopular escort missions are in the gaming world in general, that's unlikely to ever change no matter how good our mechanics are). So while they are a source of easy money, they aren't a source of very much of it, and no-one has cared over the past many years to do anything about it. As such, I really don't think that requiring a solution should block active development into areas that have significant agreement as being high priorities for fixing.

**Summary:** 
Combat AI is under active development, and having ships start out a bit more spread out has been highlighted as something that enables a lot more possible behaviors, variety, and intelligence in our combat AI. This should be expedited as it helps a core area of gameplay, and has only a minor negative of making something already broken a little bit more so.

Pointedstick:
I'll rescind my objection, but I don't have time to engage in potentially controversial PRs these days, sorry.

Zitchas:
I understand. I mainly just wanted to get your clearance in regards to the objections. 

I will bring up the escort mission problem with our coders and see if anyone has some ideas of how to fix it; but I suspect that it too might actually be an AI sort of issue.

Thank you!

EjoThims (approved):
Played with it in Revamp and through this PR, not only like the visual of ships having to actually fly in to land, but it makes mission fleet battles seem a lot more reasonable, at least to start.

I like this is an -absolutely- necessary starting point for smarter AI changes, especially, though. If they're still jumping in right on top of each other, there's just really no point in having them do anything other than the suicidal death balls they do now.

vitalchip:
Definitely support extending arrival distances. My only concern is that as a change which is most likely to happen once it doesn't go nearly far enough.

For at least a year I've had arrival distance set at 16,000 and I'd never go back to under 10,000 now. This does need to be in conjunction with expanding actual system sizes though, at a very minimum the fence needs moved out.

Pointedstick made some comments which I also agree with, this isn't something which stands alone but it's the right foundation for improvement.

When jumping into a system and landing there's more transit time but I expect this to be tempered as we also expand the systems themselves and as a result give ships higher top speeds, since there will be 'somewhere to go' that makes speed relevant.I would also expect the Ai to be altered such that it uses full speed when running for a planet or escaping but sets a more considered approach speed for combat, so we don't just get ships nose diving into the fight.

Longer transit times from jump in to landing also mean that sometimes pirate ships will launch from the planet and move to intercept while also jumping in behind and you get two attacks, mostly the situation is less dangerous but at times it can be worse. It's a fair balance and again, makes fast pirate ships useful.

Another improvement is to change how the fleets spawn. At the moment we have launching from planet, jumping in and "in action" fleets (those are there when you arrive)  Since we know the system arrival distance and the link vectors we can spawn the "in action" fleets randomly across the variuos arrival points. This creates ambush fleets and might naturally lead us to new tactics such as deploying energy mines at hyperspace exit points.

There's an issue with jump drives, their arrival is a random angle at the arrival radius, this means they will often jump into empty areas of space, far from any action and from their own fleet ships. This isn't really noticed at small radii like the standard 300-600 distance and even in the low 1,000s.  At large arrival distances it can split your fleet to all corners.

My current solution to this is two fold. 1. I can synchronise a jump drive to arrive at the hyperspace exit point, along with my hyperspacing ships 2. Fleet jumping ships arrive in a radius around the parent, not the system. So fleets stay together better.

I don't want to go into every problem and possible solution, as I said this is a start and we'll need to address various things on the way to make it 'right', what's important is that it has to start somewhere.

To round things off I'll just re state my initial point, my only real concern here is that it's too cautious, we may get only one chance to make such a change and I hope it's not squandered on a change that is too small.

NaitNait:
Played this for most of a run, really enjoyed the increased distance; however, there were a few downsides. 

The Remnant Korath bounties were annoying to catch, you pretty much had to anticipate the direction they were coming or quickly zip over to draw their attention to shoot at you, distracting them from jumping. 

Space trucking was a bit slower but safer since enemies wouldn't appear very close. 

Jump driving into high-risk systems is more dangerous due to the fleet jumping in far from another.

Zitchas:
Amazinite pointed out that a few systems have some really extreme habitable and/or belt distances, so I went through manually and checked almost all the systems that had a value over 4k. Some of them, as he said, were over 20k. Everything over 8k has been reduced significantly, and most stuff over 4k has been reduced somewhat. 

Now, nothing is over 10k, and only 11 systems are over 4k.

For any of those systems that were inhabited, I first reduced the arrival distance to 150 beyond the outermost orbit, and then checked to see how reasonable that felt. In a few cases pulled it in a bit tighter, but that was usually enough to feel reasonable. 

Here are the top 11, sorted by distance, along with their line number.
Line9710 | 8790 (Esix, EW)
Line7226 | 7420 (Convector, EW)
Line16225 | 6305 (Meblumem, Coalition)
Line10657 | 5575 (Feroteri, Korath)
Line3193 | 5480 (Almaaz, North Pirate)
Line21446 | 5000 (Sadr, South Uninhabited)
Line8818 | 4470 (Edusa, EW)
Line5505 | 4231 (Betelgeuse, North Republic 
Line25393 | 4171 ("Wah Oh", Hai)
Line19013 | 4070 (Persian, Syndicate, uninhabited)
Line21381 | 4025 (Sadr, South, Uninhabited)

Zitchas:
excerpts from VitalChip:
> Definitely support extending arrival distances. My only concern is that as a change which is most likely to happen once it doesn't go nearly far enough.
> 
> For at least a year I've had arrival distance set at 16,000 and I'd never go back to under 10,000 now. This does need to be in conjunction with expanding actual system sizes though, at a very minimum the fence needs moved out.
> ...
> When jumping into a system and landing there's more transit time but I expect this to be tempered as we also expand the systems themselves and as a result give ships higher top speeds, since there will be 'somewhere to go' that makes speed relevant.I would also expect the Ai to be altered such that it uses full speed when running for a planet or escaping but sets a more considered approach speed for combat, so we don't just get ships nose diving into the fight.
> ...
> I don't want to go into every problem and possible solution, as I said this is a start and we'll need to address various things on the way to make it 'right', what's important is that it has to start somewhere.
> 
> To round things off I'll just re state my initial point, my only real concern here is that it's too cautious, we may get only one chance to make such a change and I hope it's not squandered on a change that is too small.

I definitely like the idea of having bigger systems, but I'm personally of the opinion that there's two separate (but definitely interrelated) ideas at play here. I think that we can implement arrival distances now, get people used to the idea that in-system space is actually a thing to be crossed and not just something that is only relevant for combat. This can be implemented now, and fits very well with our current content.

Vastly expanding the size of our systems sounds really interesting, but I feel that we would need vastly more content, including significantly more planets, stations, and moons in order to really be doable. For instance, if we expanded Sol to have a number of stations, planets, etc reminiscent of all the human activity we see in *The Expanse*, I think that'd be awesome. (Not *everything* of course, but at least the 6-10 biggest ports-of-call, anyway). Whenever we do get to the point of being able to implement so much activity in-system, I think we will be able to simply scale up the arrival distances along with it.

Amazinite:
I think I'd prefer the arrival distances to be `max(1250, min(habitable range, 5000))`. As outlined in #6303, asteroid belts aren't really thick rings of asteroids that would impede hyperspace travel, as they're actually rather diffuse, so I'd rather not have belt radii influence arrival distances. The minimum distance of 1250 is chosen because that's the distance at which ramscoops and solar panels are 100% effective (with close ranges being >100% and farther ranges being <100%). The maximum distance of 5000 is chosen as that's half the system fence distance. Then the reference point of the habitable range between those two points is chosen because the player will want to generally appear in a system close to the inhabited planet in that system, and entering at the habitable range will do that.

EjoThims:
Having it be a baked in formula instead of merely a listed definition definitely helps with comparability for plug ins and new content.

Is that a change simple enough to just edit and merge this PR?

Or should it be done separately and this could instead transition to any manual overrides we want to see included in current vanilla systems?

Amazinite:
The current default behavior is that if there's an arrival distance then arrive at roughly that distance from the system center, or if there's no arrival distance then arrive near the selected planet, or if there's no selected planet then arrive near the first object with a spaceport, or if there's no spaceport then arrive near the system center. To make the suggested arrival distance equation the default would mean to completely change the behavior to just "Arrive at roughly the arrival distance from the system center," i.e. no more having no arrival distance exit you near spaceports.

It wouldn't be difficult to do, there's just a question of if it's something we want to do.

Zitchas:
Having a formulaic method of automatically determining an arrival distance based on system attributes for systems that don't have a distance specified could be good; but only if overridden by having a specified arrival distance.

While the formula that Amazinite provides is a great idea for having built-in for handling systems that don't have a set arrival distance (plug-ins, for instance); I would prefer to use set arrival distances such as those established here for existing systems in-game. We do have some very big systems, after all, including some that exceed the 5000.

And as far as the fence is concerned, I'd personally like to see it moved to an attribute on a per-system basis. So default is 10k, but an individual system can set it higher (or lower) as necessary.

PR #6245, opened 2021-09-11
fix(content) Overhaul illegal good, drug, and slave missions

Pointedstick:
This PR corrects several issues with the Pirate smuggling missions:

1. The mission descriptions listed the amount of the fine, allowing
   players to precisely calculate whether the mission could be worth it.
2. The fine amounts were much too low, often being less than the
   payment! This the player could get caught and fined and still earn a
   profit.
3. Being fined did not fail the mission! There is no possible universe
   in which the police fine you for possession of illegal goods but
   don't confiscate the goods. It just makes no sense.
4. The payments were too low given the risk entailed (if you're on the
   side of the pirates) or the sunk costs involved (if you had to bribe
   your way onto the planet), especially now that being fined fails the
   mission, and *especially* for the slave smuggling missions, for
   which the consequence is death AKA losing the game.
5. Some descriptions unnecessarily talked about "contacts at the
   location" and "landing away from the spaceport", which were just
   flavor text, but could confuse the player into thinking there was
   something special they needed to do, when this was not the case.
6. Despite some descriptions saying "the pirates won't be too happy to
   hear about this", you didn't actually lose any Pirate reputation for
   failing the mission by getting your cargo confiscated.

These issues are all fixed now. In the process, a few awkward wordings
are improved, and some standard sentences are moved to standard phrases.

The net result is that pirate jobs are now much more high risk/high
reward, as they should be. And it is now more economically feasible to
bribe your way onto pirate planets purely to take some pirate jobs for
role-playing purposes.

### DIFF
--- a/data/dialog phrases.txt
+++ b/data/dialog phrases.txt
@@ -111,7 +111,7 @@ phrase "generic pirate fleet battle on visit"
 
 phrase "generic completed pirate cargo mission"
 	word
-		`You find the contact at the edge of the spaceport. Your cargo of <commodity> is removed from your ship and you collect your payment of <payment>.`
+		`You find the contact at the edge of the spaceport. Your cargo of <commodity> is removed from your ship, and you collect your payment of <payment>.`
 
 phrase "generic completed highly illegal pirate cargo mission"
 	word
--- a/data/dialog phrases.txt
+++ b/data/dialog phrases.txt
@@ -108,3 +108,23 @@ phrase "generic fleet bounty hunting on visit"
 phrase "generic pirate fleet battle on visit"
 	word
 		`You've landed on <planet>, but there are still pirates circling overhead. You should take off and help finish them off.`
+
+phrase "generic completed pirate cargo mission"
+	word
+		`You find the contact at the edge of the spaceport. Your cargo of <commodity> is removed from your ship and you collect your payment of <payment>.`
+
+phrase "generic completed highly illegal pirate cargo mission"
+	word
+		`You find the contact outside of the main city. Your cargo of <commodity> is removed from your ship and you collect your payment of <payment>.`
+
+phrase "generic aborted pirate passenger mission"
+	word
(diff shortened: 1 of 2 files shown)


data/dialog phrases.txt
```diff
@@ -108,3 +108,23 @@ phrase "generic fleet bounty hunting on visit"
 phrase "generic pirate fleet battle on visit"
 	word
 		`You've landed on <planet>, but there are still pirates circling overhead. You should take off and help finish them off.`
+
+phrase "generic completed pirate cargo mission"
+	word
+		`You find the contact at the edge of the spaceport. Your cargo of <commodity> is removed from your ship and you collect your payment of <payment>.`
```
W1zrad, line 114:
```suggestion
		`You find the contact at the edge of the spaceport. Your cargo of <commodity> is removed from your ship, and you collect your payment of <payment>.`
```
Just need a comma here b/c two independent clauses.

data/dialog phrases.txt
```diff
@@ -108,3 +108,23 @@ phrase "generic fleet bounty hunting on visit"
...
+
+phrase "generic completed highly illegal pirate cargo mission"
+	word
+		`You find the contact outside of the main city. Your cargo of <commodity> is removed from your ship and you collect your payment of <payment>.`
```
W1zrad, line 118:
```suggestion
		`You find the contact outside of the main city. Your cargo of <commodity> is removed from your ship, and you collect your payment of <payment>.`
```
Same here!

W1zrad (commented):
Sweet! Just did a little proofreading.

data/dialog phrases.txt
Pointedstick: Yeah, this was a straight copy of the original material, but good call.

W1zrad (approved):
Thanks! Solid changes, looking forward to seeing them implemented!

Zitchas (approved):
I like the concept, and the implementation looks good. Having more sensible reasons (and means) to bribe my way onto pirate worlds is always good.

PR #6246, opened 2021-09-11
feat(mechanics): Allow boarding missions from NPC ships with flag

### LINKED ISSUE #6097: Functionality for Derelict Missions

Terin:
**Feature:** This PR implements the feature request detailed and discussed in, and would resolve #6097 

## Feature Details
The other part of half of the coin for the linked issue, this PR lets content creators give the `offers` personality to NPC ships in missions. This overrides the default blocker against NPC/"Special" ships from generating boarding missions, enabling a player to create bespoke derelicts (or ships that you could disable yourself) which can give unique boarding missions.

For example, we could have a mission similar to "Hunted" which periodically, invisibly spawns a derelict within range of the player, and if they happen to stumble upon it, they can get a selection of missions through there. Alternatively, some mission could ask you to destroy a target, but if you disable and board it first, you can get some alternative or bonus outcome.

The implementation checks for the `offers` personality when boarding, and then sets a `boardedOfferer` bool on the ship. Ideally, it would be cleaner to handle it in just the NPC class, but the BoardingMission function under PlayerInfo only checks the ship. As far as I'm aware, you cannot work up from a ship to check if it's an NPC, so it would need to be handled through ships/personalities. The `boardedOfferer` flag is then serialized as `boarded`, if a player saves in that time.

## UI Screenshots
N/A

## Usage Examples
```
mission "Derelict Generator"
	invisible
	landing
	repeat
	deadline 30
	to offer
		random < 3
	npc assist
		government "Derelict"
		personality derelict offers
		system
			distance 3 5
		ship "Freighter" "S.S. Ghost Ship"



mission "Ghost Ship"
	assisting
	destination "Earth"
	to offer
		source
			government "Derelict"
	...
```

## Testing Done
- Ensured ships with the `offers` personality could receive boarding missions
- Boarded once, then tried to board again to confirm that they would not generate a second mission
- Boarded once, saved and reloaded, and tried to board again to see no new mission
- Went to an older `derelict` mission to confirm no boarding missions would spawn off them

## Performance Impact
N/A

### DIFF
--- a/source/Personality.cpp
+++ b/source/Personality.cpp
@@ -49,6 +49,7 @@ namespace {
 	const int TARGET = (1 << 25);
 	const int MARKED = (1 << 26);
 	const int LAUNCHING = (1 << 27);
+	const int OFFERS = (1 << 28);
 	
 	const map<string, int> TOKEN = {
 		{"pacifist", PACIFIST},
@@ -78,7 +79,8 @@ namespace {
 		{"opportunistic", OPPORTUNISTIC},
 		{"target", TARGET},
 		{"marked", MARKED},
-		{"launching", LAUNCHING}
+		{"launching", LAUNCHING},
+		{"offers", OFFERS}
 	};
 	
 	const double DEFAULT_CONFUSION = 10.;
@@ -331,6 +333,13 @@ bool Personality::IsMarked() const
 bool Personality::IsMute() const
 {
 	return flags & MUTE;
+}
+
+
+
+bool Personality::Offers() const
+{
+	return flags & OFFERS;
 }
 
 
--- a/source/Personality.h
+++ b/source/Personality.h
@@ -69,6 +69,7 @@ public:
 	bool IsTarget() const;
 	bool IsMarked() const;
 	bool IsMute() const;
+	bool Offers() const;
 	
 	// Current inaccuracy in this ship's targeting:
 	const Point &Confusion() const;
--- a/source/PlayerInfo.cpp
+++ b/source/PlayerInfo.cpp
@@ -1615,9 +1615,12 @@ Mission *PlayerInfo::MissionToOffer(Mission::Location location)
 // relationship with the player. If none offer, return nullptr.
 Mission *PlayerInfo::BoardingMission(const shared_ptr<Ship> &ship)
 {
-	// Do not create missions from existing mission NPC's, or the player's ships.
-	if(ship->IsSpecial())
+	// Do not create missions from existing mission NPC's, the player's ships,
+	// or ships with the "Offers" personality that have already been boarded.
+	if(ship->IsSpecial() && (!ship->GetPersonality().Offers() || ship->IsBoardedOfferer()))
 		return nullptr;
+	if(ship->GetPersonality().Offers())
+		ship->SetBoardedOfferer();
 	// Ensure that boarding this NPC again does not create a mission.
 	ship->SetIsSpecial();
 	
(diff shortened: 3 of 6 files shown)


tehhowch:
At its core this is using a mission to offer other missions, which we already do today to great effect.

Why is this particular approach needed? It seems like this is addressing a symptom, rather than actually addressing the root issue: missions only trigger on landing, or upon boarding some ship that is not a part of any mission chain, so people find it hard to build a mission chain from a derelict.

Expanding the number of places that can spawn missions is planned (such as when entering a system or scanning something), as is allowing missions to opt-in to completing in-flight rather than upon landing.

I think a better goal is to work in those directions, or ensuring that the existing NPC "boarding conversation" (shown when you complete an npc's success criteria) is able to set the dominos in motion for the next mission in the chain, whether that's an event, condition toggle, etc.

Terin:
> At its core this is using a mission to offer other missions, which we already do today to great effect.

Yeah, it's a tried and true approach to designing specific chains of events in Endless Sky, and this would extend it to boarding NPC ships, letting us make use of the existing mission infrastructure, being able to filter by government, check reputation, ensure the player has enough cargo/passenger space before giving a mission, etc.

> Expanding the number of places that can spawn missions is planned (such as when entering a system or scanning something), as is allowing missions to opt-in to completing in-flight rather than upon landing.

Sure. I don't see why this PR isn't playing towards that goal, though. This is expanding where a mission can spawn, on an NPC ship flagged by a content creator. It is a roundabout implementation, due to how PlayerInfo::BoardingMission can't talk to NPC, but it accomplishes it.

> I think a better goal is to work in those directions, or ensuring that the existing NPC "boarding conversation" (shown when you complete an npc's success criteria) is able to set the dominos in motion for the next mission in the chain, whether that's an event, condition toggle, etc.

Those sound like useful tools to have access to, but how would that affect letting someone create missions from unique derelict ships? Is it planned for events to basically place missions into a player's list? Since otherwise, I cannot see how that specifically addresses the issue of derelicts. Like I mentioned above, the mission infrastructure provides a lot to a content creator, in a way that events or conditions couldn't really replicate in the moment you board a ship.

Issue #4034, opened 2018-11-25
The Dead Horse: Fighter/Drone Balance


Startome:
Hyratel and I feel that the current system for the use of fighters and drones is unbalanced, and has led to many players abandoning the idea of using them altogether.

When the NPC ships spawn with a full complement, players often find fighters to be very interesting. But when they try to use them, find that they die very quickly in a battle, and that a standard warship would have been a better choice than a fighter.

Also the need to spend money to replace them, where a standard warship never would have died in the first place, is also quite a deterrent.

The need to spend money to replenish them, and the fact they often die before doing any damage in a large battle, essentially rules them out as a combat effective option. A Captured capital ship makes a much better meat shield.

Especially considering that most players customize their ships, including fighters. The loss of a single fighter can take quite some time to replace, and they are often lost far faster than they can be rebuilt.
So we propose adjusting the fighter system to be more appealing to the player, and more practical, without breaking the game’s balance. This being to expand interest, and most importantly, strategies available.

I know this idea has been suggested in the past, however only as a quick request for some kind of solution. Not that I have read, has a concrete proposal been brought forth.

The most important part of our proposal is to add a module that allows carrier ships to gradually, at high energy and heat cost (and perhaps resources, depending on the ease of coding and balance), to regenerate their fighters both during and after battles. This module would take up a substantial amount of space. The outfit space, combined with the energy and heat demands, would provide the player with two options:

Only equip one manufacturing module, greatly reducing the rate at which the carrier could deploy its fighters, while maintaining weapons. Effectively serving a multirole purpose.

Focus on fighters, allowing the carrier to field fighters at a slightly less tedious rate, but requiring it to serve a true support role, as carriers traditionally do.

We believe a balanced starting point with the numbers would be for the module to occupy 50 general outfit space, while consuming 500 energy and generating 650 heat. This seems fairly realistic for a manufacturing facility, and balances the demands on the player. Meaning early game they have to make sacrifices, as they do with all builds. And late game, as more efficient alien generators, shield mods, etc, are unlocked, they can truly optimize their ships to be the most effective.

The manufacturing module could produce 2 mass per second. This would allow current fighter models to be redeployed (after the initial bay occupant is destroyed in battle) after 52-74 seconds, depending on which fighter. (the boxwing would be lowest, but isn’t truly a combat unit, so is not included)
Current drones would be redeployed at 34-59 seconds, depending on the model.

[...]

PR #6247, opened 2021-09-11
[Combat/Fighters] Content: Fighter engine speeds

Zitchas:

## Summary
Current vanilla fighters & drones do not have manoeuvring capabilities traditionally associated with fighters & drones as depicted in virtually *any* sci-fi universe or modern day naval environments. This lack of capabilities is so pronounced that stock fighters can sometimes be even outperformed by well-outfitted heavy warships. This problem has been described and discussed at length in many places, particularly in https://github.com/endless-sky/endless-sky/issues/4034 .

## Details
Each fighter & Drone has been given built-in thrust equal to 4x the stock engine, and built-in steering equal to 1.5x the stock steering. They have also had an amount of outfit space removed equal to that taken up by the stock propulsion.

The exception is the Surveillance Drone, because I didn't want to make it too hard to catch or avoid, which is at 3x the stock engine and 1.5x the stock steering.

## Impact
This means that fighters & drones are all universally very fast. Even the slowest can easily outrun most freighters and warships, and the fastest can catch virtually anything. 
In terms of combat, this means that running away from fighters & drones is not an option for most ships, and afterburners are probably going to be needed for shaking them off long enough to jump. That being said, fighters & drones will probably be very useful for screening your ship and taking out incoming fighters.
On the carrier side of things, this massive speed increase should make it feasible for carriers to launch their fighters at long range, and have them go, unload at their target, then recall them to reload. Ideally (probably requiring the increased hyperjump arrival distances and Derpy's mass adjustment), they would do this several times instead of diving into combat themselves. Carriers aren't supposed to be front-line combatants.

This also has the impact of making fighters a little bit cheaper to buy, as they no longer have the cost of a stand-alone engine outfit included in their purchase price. Given people's reluctance to use fighters based on restock cost, I don't see this slight reduction in price to be a bad thing.

## Related stuff:
This PR is intended to work in conjunction with the Weapon Pods PR. That PR reduced fighter's DPS by limiting their arsenal to small numbers of secondary weapons. This speed boost means that even with a limited arsenal, they will have a high chance of reaching their target to make use of it.

This PR is one step amongst many with the goal of making fighter-based combat its own unique part of the game with its own strengths and weaknesses, which includes the following: Not all of these will necessarily be implemented, but hopefully many of them will be:
- More fighters per carrier (at least on dedicated carrier ships. a couple fighters is fine on a cruiser-that-happens-to-have-a-bay. But a dedicated carrier? True carriers should be looking at a minimum of double the fighters they currently carry.)
- Fighters that actually have brains ( @vitalchip 's AI improvements for fast ship manouvering)
- Hyperjump arrival range set to at *least* +1000 from what it is now, possibly even +2000.
- More gunports per fighter
- Single-munition-per-hardpoint capabilities.

## PR Checklist


Arachi-Lover:
This is the second PR from you that makes the Stalker fast.
The Stalker is meant to be slow, not just slower than other fighters, actually slow, it's what balances out its high HP pool.
What part of that do you not understand?

Zitchas:
Apparently that didn't register. Is it a fighter, or is it a warship? If it is supposed to be slow (not just compared to fighters, but slow compared to most ships, then I'm not sure it belongs in the fighter category). The rest of the stats certainly look more like a one person warship rather than a fighter.

Arachi-Lover:
You mean like how the Petrel, a Fighter, has greater outfit and weapon space than the Sparrow, an Interceptor, while having more hardpoints and more customizability in general? Is it a Fighter, or is it an Interceptor?
Get what I mean? If we dictate all Fighters must be done in the fast, squishy type, they become uniform. If we dictate all Heavy Warships must be slow and tanky, they become uniform. I believe that leads to dull and predictable combat, every faction has ships exactly like every other faction, they just have stats downscaled or upscaled accordingly to Tier, and use a different skin.
Whether a tanky and slow Fighter, or a Fighter with much more customizability than an Interceptor, or something else, more variety helps make different factions not feel like the same but with more or less stats.

Zitchas:
Fair enough. Sorry, I didn't realize it was supposed to be that slow compared to *everything*. It makes perfect sense to have it massively slow compared to other fighters. Diversity is good.

vitalchip:
Ship classes should match their abilities, heavy warships should indeed all be heavy and not particularly fast, fighters should indeed be small, nimble and fast. If not then put it in another category. If you make a turret which doesn't turn then it's a gun.

Zitchas:
Well, that tends to be the case, but exceptions exist. And that being said, in some sci-fi there are deployed ships that are basically self-propelled turrets. As in the hull provides the aiming instead of a turret mechanism, but otherwise it is basically a slow (or in some cases basically stationary) ship that focuses on spinning and shooting.

vitalchip:
A fighter should have the extant properties of, well, a fighter. Until recently we were limited in ships classes due to the code, so we had to put up with, for example, the boxwing being a 'fighter' when it is not. Now, however we can properly categorize ships by their properties, so we should.

Pointedstick:
I approve of the goal, and I can appreciate the amount of work you've clearly put into this, but I'm not sure I'm sold on the implementation. Bundling the engines reduces customization potential for players, and also reduces the reason for the smallest engines to exist in the first place. It feels like an inelegant fix. Why can't we fix this speed issue by reducing these ships' mass and/or drag?

I'm totally on board with the idea that *Interceptors* have bundled systems as a side effect of making them small enough to carry onboard. But I don't think the concept works as well for Fighters, especially if it's only done for one system.

Zitchas:
> I approve of the goal, and I can appreciate the amount of work you've clearly put into this, but I'm not sure I'm sold on the implementation. 

Thanks, I appreciate that. :)

> Bundling the engines reduces customization potential for players, and also reduces the reason for the smallest engines to exist in the first place. It feels like an inelegant fix. Why can't we fix this speed issue by reducing these ships' mass and/or drag?

I don't have numbers right now, but I tried reducing mass  & drag when I was working on Revamp to get this same effect. I recall something about not really having enough room to work. If it was just top speed that was the issue, then reducing drag does work. Unfortunately, from my conversations with VitalChip (who made the excellent AI in Revamp that made Fighters actually effective and survive longer), it turns out that top speed isn't nearly as important as acceleration is. The reason is that for fighters to survive, they need to be able to do things like circle a target ship faster than the target ship can turn. Currently, that means that at least the best fighters need to have accelerations somewhere >1000.

This screenshot is comparing the Proton Barb on the left with the regular barb on the right. The proton barb's hull has been reduced to 0 mass, and dropped from 0.8 drag to 0.1.
Barb vs no mass barb
As you can see, dropping the drag was very effective in boosting top speed (300 up to 2400!), but reducing the hull from 20 mass to 0 mass only boosted the acceleration from 137 up to 169, which is nowhere near high enough to run circles around anything. I'm not even sure it could stay ahead of Leviathan's turning speed, let alone any of the more modern ships.

Basically, we run into the problem that the hull is less than 20% of the total mass of the Barb, so its capacity to affect mobility is correspondingly very limited.

Worth noting: When Amazinite was faced with the same problem in #5779 , he created a new mechanic `"acceleration multiplier"` attribute that could be added to ships, which quite simply multiplied the ship's available thrust, and then gave virtually all fighters `"acceleration multiplier" 2.`  (doubling the thrust, in my experience, isn't enough to enable circling manoeuvres with present day stock heavy ship statistics, but it *might* be with the rest of Amazinite's changes, given that the heavy warship category on average got a +65% mass increase.)

> I'm totally on board with the idea that _Interceptors_ have bundled systems as a side effect of making them small enough to carry onboard. But I don't think the concept works as well for Fighters, especially if it's only done for one system.

I did this for fighters and not interceptors for a few reasons.

1. Fighters don't have any means of leaving the system other than being carried, so they are virtually never used as a flagship. Reducing customizability for a ship that is so rarely a flagship isn't as big a deal as reducing customizability for a ship (such as an interceptor) that the player can be expected to fly.
2. Fighters almost universally are equipped with the smallest engine available to their faction, and usually can't fit anything bigger. So there's really not much choice here.
3. One of the big complaints about fighters is the fact that they are expensive to replace. Removing the engine cost from fighters makes them all a bit cheaper. 
4. On a similar note, people also complaint about the annoyance of re-outfitting fighters. The fact that they have such short lifespans and are annoying to re-outfit after each loss means that players are unlikely to spend much time running around the galaxy to pick up alternate engines every time they buy a new fighter. And given the above comment on engines, finding an engine from a different faction or outfitter-supplier is pretty much the only way to change engines.
4. Taking a look at realism, in modern fighters engines are probably one of the least likely parts to be altered, and are often customized specifically for that aircraft. They aren't drop-in components that can be interchanged with a wide variety of aircraft. Weapons, on the other hand, are swapped all the time based on the mission profile. 

As a result of these points, I made the decision to remove one degree of customization that was almost irrelevant anyway, while leaving the most relevant part of the fighter available for players to change. This change also means that the fighter's mobility statistics can be balanced directly and with greater fine tuning. For example: In the base game, due to the modularity of everything, we have no control over how acceleration compares to turning, so it is impossible for us to make a fighter that is specifically a bomber. After all, it is dirt-simple to simply re-adjust the engine loadout and have our bomber returned to air-superiority-fighter levels of agility. If the engines are all built-in, then we can establish that a particular model has, say, great acceleration but poor turning. (this is, incidentally, also great for forcing the AI into doing strafing runs with long turn around times, which typically boost survivability). Likewise, we can force the inverse of that with things like the Stalker, which are apparently supposed to be basically turrets that can be launched. Things like these would typically have high turn to facilitate targeting, but extremely low thrust.

As a side note, several people on Discord suggested building in about 2min flight time worth of batteries as a way to further reduce the mass of the ship. I haven't decided either way on it, since it does get into more of the weapons systems stuff, but there's also a good argument to be made that batteries are basically the fuel tank for fighters, and should be a built-in system.

Pointedstick:
Sorry, I confused fighters and interceptors again. My bad. I'm a lot more okay with this change, then.

However, I still think we are avoiding solving the problem in a better generic way. Because if we hardcode the engine state for Fighters, then the huge difference in acceleration and speed compared to their larger Interceptor cousins becomes a bit silly. The Finch is supposed to be 80% the same as a Sparrow, for example, so why should it be basically four times faster?

Perhaps what we need to do alter how the game calculates acceleration for small ships, to give them a boost relative to their sizes. This would entail code changes, but it might be the simplest and most far-reaching way to do what's needed here.

If we do that, I would still be on board with hardcoding the engines on fighters, but not in a way that arbitrarily makes them four times as effective as the game's own physics say they should be. :) I think we should first try to change the physics instead.

Pointedstick:
Side note:

Real-world bombers are only less maneuverable because they are physically larger so that they can carry more ordnance. If you take a fighter-sized plane and stuff it with ground-attack munitions, you have what's known as a strike fighter: a limited payload bomber in all but name, but with the maneuverability of a fighter. So there's nothing conceptually wrong with that idea.

I do really like that idea of removing fighters' reactors and giving them batteries only. It makes perfect sense, as they can just return to the mothership to recharge. Perhaps they could also have a small solar panel as an emergency backup to let them limp home if need be.

movingpictures:
I've been following discussions on fighters for a while now, and perhaps this is the time to voice my views.... and a solution that may work with what's been suggested there. 

The issue I have with fighters is one that others have voiced: they are essentially weak, and only NPCs (who don't have to invest energy to replace lost assets - speaking of Navy and Korath automata for the most part) seem to benefit from them due to large numbers.

By the time the player can afford vessels with extensive fighter banks (maybe the Aerie is the exception), the battles are so damn nasty, involving big capital ships, that fighters have almost no impact on a battle, other than to drain resources. 

The pod system was a step in the right direction, because now at least, a fighter might be able to launch a torpedo or something that does *some* damage, but even then, that's not worth the investment(s) to make 'em tick.

The best way to solve this problem, in my mind, is to tinker with accuracy of weapons in relation to the size and speed of the vessel.  WHen you think about it, in real world and various depictions, fighters are exceptionally difficult to hit because of their flitty/fast behaviour. 

Adjusting the velocity of a fighter isn't addressing this failing in the AI. However, if the combat mechanisms were tweaked so that larger, slower vessels were easier to target/hit, while faster/smaller vessels more likely to invoke an outright miss, that may solve the issue. Doing so would also breathe a little life into smaller vessels (thinking interceptors), which has some merit.

Zitchas:
That is good feedback. Kind of tying into that, in a three revamp test pr we got AI that does actually Dodge and circle targets to stay away from the line of fire. It does wonders to make fighters survive longer... But it absolutely requires that fighters have massive acceleration (this PR is sort of the bare minimum it takes for that AI to work).

But by bit, fighters are improving...

Pointedstick:
Reducing the speed and accuracy of projectiles in general would help with that. Fighters would get hit a lot less, but big slow capital ships (at least capital ships *should* be big and slow...) would get hit about the same amount. Lasers would still  be a problem though.

Zitchas:
@peteryager Actually, a mass across-the-board rebalance to mass is in the works, thanks to Amazinite in PR #5779 . However, mass changes don't solve everything, fighters still need more oomph.

As far as that goes, I generally think that the big ships need to be slower. Not deathly slow, but slower, anyway. Fighters and drones, on the other hand, need to be *much* faster, hence this PR.

peteryager:
> As far as that goes, I generally think that the big ships need to be slower. Not deathly slow, but slower, anyway. Fighters and drones, on the other hand, need to be _much_ faster, hence this PR.

It does seem a bit odd that the fastest ships in the game are often heavy warships, with a few exceptions (Corvette, Marauder Arrow).  MWs should consistently be faster than HWs, LWs faster than MWs, and interceptors faster than LWs.

I'm not actually sure if fighters should consistently be faster than interceptors.  Maybe a little bit because they're a bit lighter, but fighters being 4x faster than interceptors while Kestrels and Falcons run circles around interceptors seems a bit funny.

Zitchas:
> 
> It does seem a bit odd that the fastest ships in the game are often heavy warships, with a few exceptions (Corvette, Marauder Arrow). MWs should consistently be faster than HWs, LWs faster than MWs, and interceptors faster than LWs.
> 
> I'm not actually sure if fighters should consistently be faster than interceptors. Maybe a little bit because they're a bit lighter, but fighters being 4x faster than interceptors while Kestrels and Falcons run circles around interceptors seems a bit funny.

Fighters and interceptors... Yeah, they should probably be in the same general area. On average, I think fighters should be a little faster, but some of the interceptors should have the edge.

Same, I think, goes for most categories, except more extreme.

Fighters & drones vs interceptors = Pretty close between the groupings, some interceptors are faster.

Fighters & drones vs light warships = Less even, but a decent number of light warships should be at least able to make them work for it.

Fighters & drones vs medium warships = Only the fastest mediums in the galaxy should be able to outperform average fighters, and probably - at most - one or two are capable of outperforming them all.

Fighters & drones vs heavies = the vast majority of heavy (anything) shouldn't even have a *hope* of catching even below average fighters, even when optimized for speed. Only the fastest specialty heavies (meaning, even the hull itself was designed to be highly engine-focused build.) should be able to hold a candle to the fastest fighters.

TheCodex6824:
While I agree that more thrust would help fighters a lot (especially the ones that are stuck with ion engines), I think the more natural way to fix things would be to address the fact that the little thrust already there gets cancelled out by the (proportionally) sheer amount of mass that outfits contribute over the hull. Just adding a generator over a bare ship can double the mass. The new missile pod weapons help this a bit, but it isn't enough as pretty much everything adds enough mass to cause a problem when added together. It probably doesn't make sense to have specific outfits only for fighters, but some of the low end outfits can have their masses reduced a bit to make it more manageable. It might make sense to have built-in batteries / generator as well, although that does mean the player can't change it later.

This would have the bonus of also helping interceptors and other light ships out a bit (as they share a lot of the low-end outfits), while still not touching the bigger ships.

Amazinite:
Closing this, as making the engines of fighters integral to the fighters themselves isn't the path I think we should be taking to improve fighters.

PR #6250, opened 2021-09-14
Add warning when undefined hazards are used.

quyykk:
**Bugfix:** This PR addresses an issue that I randomly found out.

## Fix Details
I was looking at `GameData::CheckReferences` a bit and then realized that hazards aren't checked. So, this PR adds the following warning when an undefined hazard is used:

```
Warning: hazard "test" is referred to, but not fully defined.
```

## Testing Done

Added an undefined hazard and confirmed that the warning is emitted.


tehhowch (approved):
👍

PR #6252, opened 2021-09-15
Changed conditions for calling DoRepair for carried ships.

vitalchip:

----------------------

## Summary
Current code does not recharge carried ships with either excess energy or excess fuel since min(0,<value>) will never return greater than zero.

## Save File
N/A

## PR Checklist
 N/A
  
  
-----------------------

Unreported bug, found while playing with battery only fighters.

## Fix Details
Code could be changed to max(0, excess energy/excess fuel) and pass the value regardless but I feel it is better to check for excess energy/fuel and then only call DoRepair if they exist, potentially saving multiple calls when there was no excess available to begin with.

## Testing Done
Tested before and after with battery only fighters, before saw no progressive recharge from the carrier. After saw immediate recharging of all fighters from carriers excess.

## Save File
No save file, any carried ship with only batteries will show the problem.

-----------------------
**Feature:** This PR implements the feature request detailed and discussed in issue #{{insert number}}

## Feature Details
N/A

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
N/A

## Performance Impact
No performance impact expected or observed.

### DIFF
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -2004,7 +2004,7 @@ void Ship::DoGeneration()
 				Ship &ship = *it.second;
 				if(energyRemaining > 0.)	
 					DoRepair(ship.energy, energyRemaining, ship.attributes.Get("energy capacity"));	
-				if(fuelRemaining >0.)	
+				if(fuelRemaining > 0.)	
 					DoRepair(ship.fuel, fuelRemaining, ship.attributes.Get("fuel capacity"));
 			}
 		}
--- a/source/Ship.cpp
+++ b/source/Ship.cpp
@@ -1997,12 +1997,14 @@ void Ship::DoGeneration()
 			
 			// Now that there is no more need to use energy for hull and shield
 			// repair, if there is still excess energy, transfer it.
-			double energyRemaining = min(0., energy - attributes.Get("energy capacity"));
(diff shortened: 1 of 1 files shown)


source/Ship.cpp
```diff
@@ -1997,13 +1997,15 @@ void Ship::DoGeneration()
...
+			double energyRemaining = energy - attributes.Get("energy capacity");
+			double fuelRemaining = fuel - attributes.Get("fuel capacity");
 			for(const pair<double, Ship *> &it : carried)
 			{
 				Ship &ship = *it.second;
-				DoRepair(ship.energy, energyRemaining, ship.attributes.Get("energy capacity"));
-				DoRepair(ship.fuel, fuelRemaining, ship.attributes.Get("fuel capacity"));
+				if(energyRemaining > 0.)	
+					DoRepair(ship.energy, energyRemaining, ship.attributes.Get("energy capacity"));	
+				if(fuelRemaining >0.)
```
quyykk, line 2007:
```suggestion
				if(fuelRemaining > 0.)	
```

quyykk (commented):
Looks good! Also can you please format your PR correctly?

tehhowch:
One of those where you go, "this was broken for _how long_?"
Nice find!

PR #6257, opened 2021-09-17
feat(content): Add barrage launcher missions

W1zrad:
## Summary
Adds a mission chain to unlock the Barrage Launcher added by #6239 (and is thus dependent on it). In short: you meet Horace Lovelace, head of missile launcher development at Lovelace labs (intended to be a little strange and off-putting without being a Barmy Edward clone), and help him test the new Barrage Launcher, a Navy-funded weapon developed in response to the Automata attack in human space.
The structure is very similar to the Kraz Cybernetics missions the player follows during the FW campaign.
I'm not sure where the best place to put these missions is, so I've created a new file for the time being. If someone has a better idea, I'm open to suggestions!

A note on testing: Some tests failing are unavoidable (as far as I know) because this PR refers to - but doesn't include - outfits from #6239.

## To-Do:
- [ ] Add variants for Republic fleets after the Barrage Launcher becomes available. I'd like to wait for (a) any balancing on the outfit to be done first and ~~(b) the Broadsider mission chain to be finished (currently being worked on by Zitchas)~~ before I create new variants. I'd expect any variants would need to be altered after ~~those two things are~~ that's done anyways.
- [X] Relocate file to `humanity interludes`
- [X] Create fleets that utilize variants

### DIFF
--- a/data/human/humanity interludes.txt
+++ b/data/human/humanity interludes.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2021 by W1zrad
+# Copyright (c) 2021 by Zitchas
 #
 # Endless Sky is free software: you can redistribute it and/or modify it under the
 # terms of the GNU General Public License as published by the Free Software
@@ -134,7 +134,9 @@ mission "Barrage Available"
 	to offer
 		has "event: barrage launcher available"
 	on offer
-		dialog `You receive a message from Horace Lovelace. "Captain <first>- The Barrage Launcher is now available for purchase. Lovelace Labs thanks you for your assistance. -H"`
+		conversation
+			`You receive a message from Horace Lovelace as you land.`
+			`	"Captain <first>- The Barrage Launcher is now available for purchase. Lovelace Labs thanks you for your assistance. -H"`
 		fail
 
 event "barrage launcher available"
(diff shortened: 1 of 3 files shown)


ayoch:
Seems like data/weapons.txt needs to be updated. data/sales.text also? Anything else?

W1zrad:
> Seems like data/weapons.txt needs to be updated. data/sales.text also? Anything else?

`weapons.txt.` is updated in #6239, which this PR is dependent on. Updating `sales.txt` shouldn't be necessary because of the event on line 140 of `unlock missions.txt`, I believe. If you think of anything else throw let me know though; it's certainly possible I missed something.

ayoch:
> > Seems like data/weapons.txt needs to be updated. data/sales.text also? Anything else?
> 
> `weapons.txt.` is updated in #6239, which this PR is dependent on. Updating `sales.txt` shouldn't be necessary because of the event on line 140 of `unlock missions.txt`, I believe. If you think of anything else throw let me know though; it's certainly possible I missed something.

Fair enough. I was just trying to figure out what caused those three tests to fail.

Amazinite:
As mentioned in #6239, the Barrage Launcher doesn't really have a place after #5793, so I'll be closing this PR.

Issue #6143, opened 2021-07-28
Remeberance Day celebration pasenger delivery not dated(/rush)?


mirror176:
Got the task: "Brink 5 passengers to New Sahara in the Phecda system by for the Remembrance Day celebration on February 19th. Payment is 100,000 credits."
Should this task be dated/rush? Getting to a site for a scheduled celebration seems like a candidate unless this task gets thrown around any time and/or its thought the passengers didn't care when they would be there for. "Bring Arthur and his Hai friend, Kiru, to Vinci in the Miaplacidus system for a gaming convention" would also seem like the same kind of idea. Had a cotton sheet delivery for same celebration event also not dated while others to Earth for Earth Day celebration were dated/rush.

Pointedstick:
All "Remembrance Day celebration" jobs have the `deadline` tag, so if this was a bug, it's since been fixed.

Pointedstick:
For the Hai mission in question, see https://github.com/endless-sky/endless-sky/pull/6261

PR #6261, opened 2021-09-20
fix(content): Add deadline for "Hiding in plain sight" mission

Pointedstick:
**Bugfix:** This PR addresses something brought up in issue https://github.com/endless-sky/endless-sky/issues/6143

## Fix Details
Add deadline tag to the "Hiding in plain sight" mission, since presumably the convention is taking place at a specific date and you need to get theme there by then.

### DIFF
--- a/data/hai/hai missions.txt
+++ b/data/hai/hai missions.txt
@@ -1100,12 +1100,13 @@ mission "Expanding Business: shipyard complete"
 
 
 mission "Hiding in Plain Sight"
-	description "Bring Arthur and his Hai friend Kiru to <destination> for a gaming convention."
+	description "Bring Arthur and his Hai friend Kiru to <destination> for a gaming convention by <date>"
 	minor
 	source
 		government "Hai"
 	destination "Vinci"
 	passengers 2
+	deadline
 	to offer
 		random < 30
 		has "First Contact: Hai: offered"
@@ -1116,7 +1117,7 @@ mission "Hiding in Plain Sight"
 				`	"I do. Why do you want to go there?"`
 				`	"I don't. You'll have to ask someone else."`
 					decline
-			`	The human responds, "There's a massive convention on <planet> for a game called 'Boundless Frontiers.' The convention has a huge cosplaying scene, and I think Kiru would fit right in."`
+			`	The human responds, "There's a massive convention on <planet> on <date> for a game called 'Boundless Frontiers.' The convention has a huge cosplaying scene, and I think Kiru would fit right in."`
 			`	"Arthur and I have <payment> to give you if you bring us to the <planet>," Kiru says. "We can find our own way back to the Hai space, as well."`
 			choice
 				`	"Alright, I can take you there."`


Zitchas (approved):
The mission is explicitly trying to get someone to a convention. Makes sense to me that it would have a specific date and not be "just another passenger that can be ferried around until convenient."

waterhouse:
I just accepted the "Hiding in plain sight" mission, and it had chosen a deadline of *two days ago*.  So when my ship took off, it immediately failed.

I tried it again (edited "hai missions.txt" to increase the probability of offering the mission), and got exactly the same result.  This is on current git, 3cfedcd99151d0ac80efa73cc701ddf970c86a28 (plus a commit I use to make it build on my machine).

Savefile attached.  Open it, take off, take one jump to the Wah Oh system, land, then go to the bar ("Trading" button) and you have a 30% chance of being offered the "Hiding in plain sight" mission (again, edit "hai missions.txt" to increase it).  Accept it, and observe that the current date is 12 Oct 3014 while the mission deadline is 10 Oct 3014.
[Miles Edgeworth copy previous.txt](https://github.com/endless-sky/endless-sky/files/7271100/Miles.Edgeworth.copy.previous.txt)

I don't know if it's caused by this—as long as "deadline" with no arguments is valid syntax, which it seems to be (used relatively widely), it doesn't _seem_ like there's anything wrong with this commit.  But maybe there's an underlying problem with the implementation of "deadline"?

I'll plan to make a full bug report unless someone has insights here.

waterhouse:
Some dumb printf-ing:
```
	if(deadlineBase || deadlineMultiplier) {
		result.deadline = player.GetDate() + deadlineBase + deadlineMultiplier * jumps;
		printf("deadlineBase %d deadlineMultiplier %d jumps %d\n", deadlineBase, deadlineMultiplier, jumps);
	}
```
Yields this:
```
deadlineBase 24 deadlineMultiplier 0 jumps -2
deadlineBase 0 deadlineMultiplier 2 jumps 4
deadlineBase 0 deadlineMultiplier 2 jumps 4
deadlineBase 0 deadlineMultiplier 2 jumps -1
```
Assuming the Hiding in Plain Sight mission comes from that last deadline calculation, today + 0 + 2 * -1 = today - 2 = 2 days ago, which perfectly explains the symptoms.

Probably, the fact that "jumps" can be a negative number and yet be used in these calculations is a problem.  That seems to deserve its own issue.

Doesn't look like Mission.cpp has changed any relevant code in a long time...
How about the Days function? ... Seems like not.
Has this just been broken for a long time and no one noticed?  Seems surprising.

I'll go to sleep soon, won't do anything further tonight.

Amazinite:
Deadlines, like payments, don’t calculate through wormholes. An exact deadline/payment must be given.

Pointedstick:
Gross. I guess we should manually specify a month or something. Wanna submit a pull request for that?

waterhouse:
Ok, I'm writing up a PR that gives 30 days.  I think I'll also file an issue for the underlying problem.  Will post them here.

PR #6262, opened 2021-09-21
feat: Remove Runtime Type Information from code and g++ compilation

petervdmeer:
**Feature:** This PR implements the feature request detailed and discussed in issue #6210

## Feature Details
Class Test.cpp had a dynamic_cast, while the ES style guide suggests to never use runtime-type-information (and it explicitly mentions not to use dynamic_casts).
This PR removes the usage of Runtime Type Information (rtti) in Test.cpp by removing the need to dynamic_cast by making the called method a virtual method from the base class.
This PR also removes rtti support by adding a flag to compilation by g++ to avoid including rtti support. Disabling support for rtti should help in avoiding any (accidental) usage of rtti in the future.

## UI Screenshots / Usage Examples
N/A

## Testing Done
Compiled with the `-no-rtti` flag before making the changes, got a proper error-message on the dynamic cast.

## Performance Impact
The compiled binary is expected to become a few kilobytes smaller, no significant other impact expected.


quyykk:
It's a bit dumb but if we add another compilation flag to `SConstruct` we also need to add it to the other project files (CodeBlocks and Xcode). Turns out the `-Wold-style-cast` flag is also missing from the two projects from what I can tell :)

PR #6266, opened 2021-09-22
Break all projectile targeting when you bribe a government

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #2277.

## Feature Details
This PR changes the mechanics of bribing such that when you bribe a government, your projectiles and the projectiles of the government you bribed lose targeting and collision with one another. This prevents the frustrating scenario where you bribe a government that you are in combat with, and just after you have bribed them they become hostile again because projectiles that were in flight when you bribed the government hit their ships, once again provoking them into hostility.

Part of the code in this PR is copied over from #2878, particularly the mechanism for HailPanel using a callback through MainPanel to reach Engine.

## UI Screenshots
N/A

## Usage Examples
N/A

## Testing Done
Engaged in combat with a pirate who was using missiles while I was using missiles as well. Just before our missiles hit each other, I bribed the pirate and observed how the missiles that were about to impact passed right through each ship. This would be the case for all projectiles, not just missiles.

## Performance Impact
N/A (The projectile list is iterated through only once after bribing to break the targeting on the relevant projectiles. The projectile list is iterated through every frame for collision detection, so adding one more iteration for a single frame after bribing doesn't harm performance.)

## Additional Context
Should this change be seen as too powerful in allowing the player to cheat death (i.e. by bribing the government they're fighting just before receiving a killing blow), then a potential change could be to increasing the size of bribes with each subsequent bribe given to a specific government, at the very least within a single day but potentially over the course of a period of days, or simply increase the bribe costs in general.

It is important to note though that later game hostiles are unable to be bribed, so perhaps bribing being highly forgiving for early game combat is desirable.

### DIFF
--- a/source/Engine.h
+++ b/source/Engine.h
@@ -88,6 +88,10 @@ public:
 	void RClick(const Point &point);
 	void SelectGroup(int group, bool hasShift, bool hasControl);
 	
+	// Break targeting on all projectiles between the player and the given
+	// government; gov projectil0es stop targeting the player and player's
+	// projectiles stop targeting gov.
+	void BreakTargeting(const Government *gov);
 	
 private:
 	void EnterSystem();
--- a/source/Engine.cpp
+++ b/source/Engine.cpp
@@ -1127,6 +1127,24 @@ void Engine::SelectGroup(int group, bool hasShift, bool hasControl)
 
 
 
+// Break targeting on all projectiles between the player and the given
+// government; gov projectiles stop targeting the player and player's
+// projectiles stop targeting gov.
+void Engine::BreakTargeting(const Government *gov)
+{
+	const Government *playerGov = GameData::PlayerGovernment();
+	for(Projectile &projectile : projectiles)
+	{
+		const Government *projectileGov = projectile.GetGovernment();
+		const Government *targetGov = projectile.TargetGovernment();
+		if((projectileGov == playerGov && targetGov == gov)
+			|| (projectileGov == gov && targetGov == playerGov))
+			projectile.BreakTarget();
+	}
+}
+
+
+
 void Engine::EnterSystem()
 {
 	ai.Clean();
--- a/source/HailPanel.cpp
+++ b/source/HailPanel.cpp
@@ -40,8 +40,8 @@ using namespace std;
 
 
 
-HailPanel::HailPanel(PlayerInfo &player, const shared_ptr<Ship> &ship)
-	: player(player), ship(ship), sprite(ship->GetSprite()), facing(ship->Facing())
+HailPanel::HailPanel(PlayerInfo &player, const shared_ptr<Ship> &ship, function<void(const Government *)> callback)
+	: player(player), ship(ship), callback(callback), sprite(ship->GetSprite()), facing(ship->Facing())
 {
 	SetInterruptible(false);
 	
@@ -233,7 +233,11 @@ bool HailPanel::KeyDown(SDL_Keycode key, Uint16 mod, const Command &command, boo
 	bool shipIsEnemy = (ship && ship->GetGovernment()->IsEnemy());
 	
 	if(key == 'd' || key == SDLK_ESCAPE || key == SDLK_RETURN || (key == 'w' && (mod & (KMOD_CTRL | KMOD_GUI))))
+	{
+		if(callback)
+			callback(bribed);
 		GetUI()->Pop(this);
+	}
 	else if(key == 't' && hasLanguage && planet)
 	{
 		if(GameData::GetPolitics().HasDominated(planet))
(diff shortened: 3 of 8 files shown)


Pointedstick:
...Or perhaps an equally frustrating scenario where you bribe them, but their in-flight missiles still kill you. Nice.

source/Engine.h
```diff
@@ -88,6 +88,11 @@ class Engine {
 	void RClick(const Point &point);
 	void SelectGroup(int group, bool hasShift, bool hasControl);
 	
+	// Break targeting on all projectiles between the player and the given
+	// government; gov projectil0es stop targeting the player and player's
```
Amazinite, line 92:
```suggestion
	// government; gov projectiles stop targeting the player and player's
```

tehhowch:
So the previous PR, but just a pure buff to bribing?

Amazinite:
To my understanding, the previous PR only stopped projectile tracking, but projectile collision after bribing could still happen because the target government was not cleared, so #2878 still has the issue of becoming hostile because of collisions immediately after bribing.

And yes, this change goes both ways so that the player also doesn't need to dodge projectiles from the government they just bribed.

Amazinite:
If the concern is that buffing bribing without an associated nerf is too much, I do address that in the PR description:

> It is important to note though that later game hostiles are unable to be bribed, so perhaps bribing being highly forgiving for early game combat is desirable.

While there are multiple PRs open that seek to change game balance such that the early game isn't so unforgiving, particularly for interceptors and perhaps light warships, I don't imagine any of those will be done soon, and being insta-ganked in early game combat is are persistent complaint about the game. That's part of the reason why I've made it so that bribing breaks collision against the player. If the player doesn't want to deal with combat in the early game, then swiftly bribing pirates should perhaps *actually* remove the player from combat by breaking collision.

The other side of having collision break against the bribed government is entirely for the purpose of removing the frustrating scenario of attempting to bribe a government but being stuck in a cycle of bribing and provoking because your escorts have projectiles that are still in flight when you bribed. It's been pointed out that this behavior further incentivizes the use of lasers, a balance issue we've been wrestling with for a while, as lasers will instantly stop after bribing. While I don't imagine that everyone will suddenly start using projectile weapons and missiles after this change, this does take a step toward knocking lasers off their pedestal. 

I also think that it'd be better to buff now then nerf later as needed after seeing the results of the bribe buff as opposed to rolling a buff and a nerf together. Only change a single variable at a time and all that. But I certainly have some potential nerfs to bribing in mind should a nerf be seen as necessary after this change. The PR description mentions the idea of perhaps increasing the cost of bribes for each time you bribe within a given day/some period of time (this nerf particularly targets repeatedly bribing for easier ship capturing), but we could also increase the size of the bribe based off of how much hull the player's flagship has left until they are disabled. For example, if you bribe before your shields are down then you'd have no additional bribe multiplier, but should you bribe while your hull is 10% away from being disabled, or even when you're disabled, then the bribe size could increase be a significant amount, say 2 to 3 times higher for bribing while you're down, costing the player more for using bribes to survive as opposed to using them to simply avoid combat.

Pointedstick:
I don't think it's too much.

vitalchip:
I'm confused as to how the collisions are avoided?

When you set targetGovernment to nullptr; Won't this mean pGov evaluates to false in CollisionSet and the projectile will collide with anything?

 			// Check if this projectile can hit this object. If either the
			// projectile or the object has no government, it will always hit.
			const Government *iGov = it->body->GetGovernment();
			if(it->body != target && iGov && pGov && !iGov->IsEnemy(pGov))
				continue;

Zitchas (approved):
This is for the objective and currently described behavior of bribing also breaking projectile targeting, possibly even removing the possibility of said projectiles impacting at all. 

Having a "kill switch" to terminate missiles at the last second is a long standing facet of sci-fi, and the proposed benefits to early-game combat (especially making bribery a more useful way of dealing with pirates) is very worthwhile.

(I don't have a strong grasp of the code, so this approval doesn't apply to that)

Amazinite:
@vitalchip I'm not clearing the projectile's government, I'm clearing it's *target*Government.

Projectile is a Body.
https://github.com/endless-sky/endless-sky/blob/12163511ccd42853073800f31b8952d6aac8488e/source/Projectile.h#L37

Body::GetGovernment returns the government that fired the projectile. (i.e. the government that is being checked in the collision detection.)
https://github.com/endless-sky/endless-sky/blob/12163511ccd42853073800f31b8952d6aac8488e/source/Body.cpp#L190-L193

But Projectile::BreakTarget clears the *targetGovernment* not the *government*.
https://github.com/endless-sky/endless-sky/blob/12163511ccd42853073800f31b8952d6aac8488e/source/Projectile.cpp#L332-L337

vitalchip:
Right, I see the difference.  I still can't work out why clearing the targetGovernment stop the projectile from colliding but I'll take another look tomorrow, just for my curiosity and understanding.

Hyugat:
This is a very welcome addition. The broken bribing system was what really killed my enjoyment of permadeath runs for a while. I think that, for the most part, exploiting this system shouldn't be too much of an issue, since if you bribe a ship 10, 15, or however many times it takes to cheese capturing it, you've probably spent more money than that ship is worth anyways. It seems to me like a [suboptimal, unsustainable] strategy.

PR #6267, opened 2021-09-24
Correctly draw one way wormholes.

quyykk:
## Fix Details
Currently one way wormholes are drawn just like two way wormholes: There are 2 arrows. But that's wrong, because for a one way wormhole one direction of travel is not possible, but just by looking at the map it doesn't seem like the wormhole is one way.

This PR fixes this by not drawing the arrow for one way wormholes. Here's how a one-way wormhole looks like with this PR:
oneway

### DIFF
--- a/source/MapPanel.cpp
+++ b/source/MapPanel.cpp
@@ -1002,7 +1002,7 @@ void MapPanel::DrawWormholes()
 		const System *from = waypoints.back();
 		for(const System *to : waypoints)
 		{
-			if(player.HasVisited(*from) && player.HasVisited(*to))
+			if(from->FindStellar(&p)->HasSprite() && player.HasVisited(*from) && player.HasVisited(*to))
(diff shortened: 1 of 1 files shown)


Zitchas:
Umm, where is it not working? Is there a situation where these don't work? The set of one-way wormholes that I see on regular basis show their directionality quite clearly.

Directional wormholes

Amazinite:
“One way” in this context means a wormhole leads to an object with no sprite, meaning it can’t be landed on from the other direction.

quyykk:
Yes the circular wormholes work. I'm talking about true one way wormholes, so wormholes that transport you to a system that doesn't include a wormhole. like in the image. There are no true one way wormholes in the game but I think becca plans to add simple one way wormholes in Patir II and those would be drawn incorrectly.

Zitchas:
Ah, OK. Thanks for the explanation.

Issue #4654, opened 2019-11-09
Datatype hulls separate from ships


petervdmeer:
**Describe the solution you'd like**
This feature-request is to have hulls as a separate data-type/in-game-object from a ship. Ships in Endless Sky are currently a set of outfits, hull properties, a hull image, hardpoints and active stats. After this change, a ship should refer to a hull definition and a set of outfits.

The new datatypes would look like:
- **Hull**: Frame of the ship, with properties like hardpoints, maximum hull, maximum shields, but no actual amount of hull, no actual shields and no weapons or outfits.
- **Ship**: Ship in active service. Hull, Outfits, Weapons and active stats like actual amount of hull, amount of shields and amount of fuel.

Making this change will make it easier to implement the following features:
- Hulls "in stock", similar to how outfits can be "in stock".
- Player buying ships with the same config as already in the fleet (useful for replenishing swarms of fighters, avoiding the need to manually re-configure each new fighter to the desired config).
  - This should only work on planets where all parts for the ships/fighters are "in stock", "in cargo", "in storage" and/or "for sale".
- Allow a shipyard to sell variants of ship-models (with the same hull).
  - Could be restricted to only selling variants for which the outfits are also available.
- Allow buying empty hulls (if the player intents to reconfigure the stock hull/ship completely).
- Updating hardpoint and engine locations without affecting savegames (since hulls with the hardpoint and engine definitions will be stored in the game-files).
- Transporting small hulls (for example for fighters) as cargo.

**Describe alternatives you've considered**
I considered if we should also have a data-type model (a hull with weapons and outfits, but no active stats). But I didn't see a very strong reason to do that.

**Additional context**
This feature looks non-backwards compatible; I considered three ways to implement this feature:
- Fully backwards incompatible, just don't load existing savegames with ships without hull definitions.
- Temporary backwards compatible; translate existing ships to hull-definition (numbered) and new ship-definition.
  - This requires some hull definitions to be stored in savegames.
- Fully, longer term, backwards compatible; support ships with and without hull.
  - This would lead to some ships behaving different on sale than others.

PR #6269, opened 2021-09-25
feat: Add Equals function to Dictionary (for future de-duplication of ship.baseAttributes)

petervdmeer:
**Feature:** This PR implements a first step towards re-using Ship.baseAttributes among ships.

## Feature Details
The `Ship.baseAttributes` will be identical for most ships of the same Model, but currently every ship in the game contains a copy of all the baseAttributes. Active ships in the game also contain an `const Outfit *explosionWeapon` pointer to the baseAttributes of their Models in GameData to ensure that explosions can be properly shown without pointer-exceptions after the ship-instance dies.

This PR is about adding Equals functions to relevant classes (at initial upload only to the Dictionary class, but I expect Weapon and Outfit to also get an Equals function in this PR or in follow-up PRs). The Equals function should help to centrally store the BaseAttributes without duplication and with the ability to keep a stable pointer to BaseAttributes from the ship (even after it dies, gets removed and still is exploding).

## UI Screenshots /  Usage Examples
N/A

## Testing Done
I added an unit-test for the Equals function in Dictionary.

## Performance Impact
No impact expected in the RealTime section of the game; BaseAttributes don't seem to be used in the RealTime sections of the game, since there the total set of Attributes including the data from installed outfits seems to be relevant.

### DIFF
--- a/source/Dictionary.h
+++ b/source/Dictionary.h
@@ -32,6 +32,9 @@ public:
 	double Get(const char *key) const;
 	double Get(const std::string &key) const;
 	
+	// Compare two dictionaries to see if they contain exactly the same data.
+	bool Equals(Dictionary &other);
+	
 	// Expose certain functions from the underlying vector:
 	using std::vector<std::pair<const char *, double>>::empty;
 	using std::vector<std::pair<const char *, double>>::begin;
--- a/source/Dictionary.h
+++ b/source/Dictionary.h
@@ -33,7 +33,7 @@ public:
 	double Get(const std::string &key) const;
 	
 	// Compare two dictionaries to see if they contain exactly the same data.
-	bool Equals(Dictionary &other);
+	bool Equals(const Dictionary &other) const noexcept;
 	
 	// Expose certain functions from the underlying vector:
 	using std::vector<std::pair<const char *, double>>::empty;
--- a/source/Dictionary.cpp
+++ b/source/Dictionary.cpp
@@ -93,7 +93,7 @@ double Dictionary::Get(const string &key) const
 
 
 // Compare two dictionaries to see if they contain exactly the same data.
-bool Dictionary::Equals(Dictionary &other)
+bool Dictionary::Equals(const Dictionary &other) const noexcept
 {
 	auto it = begin();
 	auto ito = other.begin();
--- a/source/Dictionary.cpp
+++ b/source/Dictionary.cpp
@@ -89,3 +89,19 @@ double Dictionary::Get(const string &key) const
 {
 	return Get(key.c_str());
 }
+
+
+
+// Compare two dictionaries to see if they contain exactly the same data.
+bool Dictionary::Equals(Dictionary &other)
+{
+	auto it = begin();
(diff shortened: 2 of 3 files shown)


source/Dictionary.h
```diff
@@ -32,6 +32,9 @@ class Dictionary : private std::vector<std::pair<const char *, double>> {
 	double Get(const char *key) const;
 	double Get(const std::string &key) const;
 	
+	// Compare two dictionaries to see if they contain exactly the same data.
+	bool Equals(Dictionary &other);
```
tehhowch, line 36:
```suggestion
	bool Equals(const Dictionary &other) const noexcept;
```

tehhowch (commented):
Sorry, but I don't really see the purpose of this method at the moment.
While currently no ship can have a custom explosion, that is something I (and others) would like to change.

For example, destroying a ship using some arbitrary weapon might result in a differently colored explosion, or one with mutated characteristics. To reach that scenario, we will need to pass the `Weapon` from a ship instance by value, rather than a pointer to a base instance.
(This will also allow us to restore the "hidden" nature of many person ships, as their custom ship models will no longer need to be declared as root-level nodes in order to have a non-empty explosion.)


Changes to how we instantiate ship instances should result in the removal of the "baseAttributes" field, too, as there will be a distinction between a Ship instance and a Ship "template".

source/Dictionary.h
petervdmeer: Yes, should be const. I'll update this in the next commit.

petervdmeer:
> Sorry, but I don't really see the purpose of this method at the moment.

No problem. I expect this function to be useful when we want to to de-duplication when we automatically translate "full ships" into "ship templates" and "ship instances". But we didn't even decide if we want to do such automatic translation and if we then also didn't decide if we want to do de-duplication there, so I guess this PR is a bit premature.

> While currently no ship can have a custom explosion, that is something I (and others) would like to change.

I am also a member of this "others" group. :smiley:
  
> For example, destroying a ship using some arbitrary weapon might result in a differently colored explosion, or one with mutated characteristics. To reach that scenario, we will need to pass the `Weapon` from a ship instance by value, rather than a pointer to a base instance.

Yes, passing the "Weapon" by value is one of the options I considered, but I also came up with other options:
- Keeping the ship object alive until after the "Weapon" finishes. (So you can just pass a pointer to the Weapon in the ship itself.)
- Storing the "Weapon" itself outside of the Ship classes (in GameData) so that it can stay alive after the ship dies. (This is the option that this PR relates to.)

> Changes to how we instantiate ship instances should result in the removal of the "baseAttributes" field, too, as there will be a distinction between a Ship instance and a Ship "template".

I agree and I had some similar ideas. When we go this route then we will also need to put some thought into dealing with existing savegames, since those savegames currently contain mostly ship "templates", while they should contain "ship instances" in the future. I listed some options on how we could deal with this in the "additional context" part of https://github.com/endless-sky/endless-sky/issues/4654 (where chassis can be read as "template" and "active ship" as "ship instance").

PR #6270, opened 2021-09-28
Projectile/Submunition additions/changes

EjoThims:
**Feature:** This RFC is to discuss potential/future changes and additions to projectile and submunition behavior.

I'm going to go ahead and list a few different things I can think of having seen discussed, but the creation of this RFC was prompted specifically by curiosity about weapons creating lasting zones of damage, akin to common spell effects like firewall or blizzard.

List can be expanded/trimmed as we figure out how feasible certain behaviors are with what changes or as new suggestions come up.

Names are descriptive place holders, not any kind of stone set mandate.

In general, even where already possible with submunitions, I think we could benefit from base access to them, to reduce repetitive strings of dozens (or even hundreds) of submunitions, for both creation ease and performance issues.

## Feature Details

Pulse AoE- repetitive damage in area, like repeat explosions. Offsets or inaccuracy could change point of origin, and individual explosions need not be same size as zone they spawn within. This is blizzard when using both of those modifications.

Continuous AoE - constant, consistent damage in area (though drop off over time would still be a potential modification), dealt on initial entry as well as at regular times. This is firewall.

AoE Wave - a moving zone of damage that is not stopped by collision. Preferably an attribute that could make either of the above mobile instead of static.

Piercing - a projectile that can pierce X number of targets/frames/whatever measure. Good for hitting multiple ships and/or the same ship multiple times.

Repeat - A projectile that repeats itself X times. A shortcut for looped submunitions that (hopefully) will also avoid the problem of all submunitions in a looped chain going off when enemy is hit early. Modifications could include allowing repeat from initial origin or projectile's current location.

Repeat End - A projectile that spawns at the end of a repeat chain. A shortcut for specific change after a submunition loop.

Charge - projectile that only fires X frames after firing input, preferably from current location of ship or parent projectile.

Spawn - creates a submunition after X frames without current projectile dying

Spawning - creates a submunition -every- X frames without current projectile dying

## Usage Examples

TBD

## Performance Impact

TBD


EjoThims:
I clearly did that wrong.

Closed for now.

PR #6275, opened 2021-09-29
Galaxy Improvements

quyykk:
## Feature Details

- Only one galaxy is rendered at a time.
- Systems are always centered around 0, 0 inside their galaxy.
- Switch between galaxies using `<-` and `->` keys.
- Intergalactic travel works and is correctly rendered using a dashed wormhole arrow (see gif).
- Galaxies now inherit from `Body` so they support animations (labels can also be animated).
- Better syntax for labels
- Plugins that define 2 (or more!) galaxies next to each other are broken as follows: The player will only be able to see systems from one galaxy at a time, while the content creator of such a plugin intended every system to be visible at the same time.

    I do not know of any plugin that uses 2 galaxies next to each other, but that doesn't really mean anything. Systems that have neighbors in another galaxy will be warned about. Any links between galaxies will appear incomplete in the map panel though.

- One change this PR introduces is that galaxies can no longer be separated by links or jumps. This means that they have to be separated by a wormhole (which is what every plugin that I know of does anyways). But it's still a change from the current behavior.

What this PR doesn't do because I don't know whether/how I should:

- UI buttons to switch galaxies for the map.
- Display the current galaxy's name in the map somewhere?
- Might be useful for mission to specify a source/destination galaxy?

## UI Screenshots

Here's a bad gif:

galaxy

## Usage Examples

```
galaxy Test
    pos x y # backwards compatibility
    sprite <name>
        ....
    label <name>
        pos x y
        sprite <name>
            ....
    ....

system SomeSystem
    galaxy Test
```

## Testing Done

- Tested various travel paths between galaxies.
- Tested switching galaxies using the keyboard.
- Tested going through a mission list correctly switches galaxies.
- Reverted `data/` to verify that this PR is indeed backwards compatible.

## Performance Impact

Small - only for the features listed above.


Arachi-Lover:
Bit of a spoiler function isn't it? Even if all shown was a blank void, wouldn't be hard to figure out that pressing some buttons made the Milky Way vanish then reappear was due to other galaxies existing.
Or does it not allow the player to view what they haven't discovered yet (ie a new pilot seeing Pleiades despite never having left Ruticulus)?

quyykk:
If you haven't visited any system in galaxy X, then you have no way of going to galaxy X. So the answer is no, new players cannot see the pug galaxy if they haven't gone through the wormhole first.

data/coalition/coalition missions.txt
```diff
@@ -91,12 +91,15 @@ mission "Discovered Coalition Space"
...
-		sprite "label/kimek"
-	galaxy "label arachi"
-		sprite "label/arachi"
+	galaxy "Milky Way"
+		label "saryds"
+			sprite "label/saryds"
+	galaxy "Milky Way"
+		label "kimek"
+			sprite "label/kimek"
+	galaxy "Milky Way"
```
Amazinite, line 100: Can't you be able to just add all three labels at once? If not, then you should be capable of doing that; we don't have to remention the fleet we're adding a variant to for each new variant added by an event, for example.
```suggestion
		label "kimek"
			sprite "label/kimek"
```

Amazinite:
> Galaxies now inherit from `Body` so they support animations

This also means that they can make use of the new `scale` attribute (#4588), closing #5497.

quyykk:
Just realized that plugins might have 2 (or more) galaxies right next to each other, which this PR would break. I'll have to think about how to best support that.

beccabunny:
> Just realized that plugins might have 2 (or more) galaxies right next to each other, which this PR would break. I'll have to think about how to best support that.

Is there any reason they couldn't make them simply a single galaxy spaced apart, as it is now?

quyykk:
> Is there any reason they couldn't make them simply a single galaxy spaced apart, as it is now?

No since they can merge the two galaxy images to create one. But this PR should be backwards compatible, so I'll have to support a use case like that.

Plip-Plop:
I remember being extremely disappointed when I traveled through Deneb for the first time and realized that both galaxies were within scroll distance of each other.

tehhowch (commented):
Some more thoughts

Amazinite:

Closing this PR because Quyykk has said that he's not getting to it soon.

Issue #424, opened 2015-12-09
Projectile that is not destroyed when it damages ships


thenerdfreak:
Maybe a decently powerful railgun or something that could penetrate multiple ships

endless-sky:
Right now, the only way a projectile can damage multiple ships is if it has a blast radius. Making a projectile that could pass through one ship (damaging it) and then go on to damage another would mean every projectile would need to remember what ships it has already hit so that it cannot damage a single ship twice. That's doable, but maybe not worth adding support for unless there's a big demand for it.

luiges90:
what about having the projectile damages any ship that is overlapping with it every frame, and will not get destroyed upon hitting any ship?

(Yes large ship will take more damage that way, but this is one simple way of doing similar effect the OP mentioned)

brick96:
It makes sense that a bigger ship would take more damage since the shot would go through alot more of the ship

jafdy:
Nice idea but I thing I would be inclined to put it as a main gun and use it as an artillery weapon.

thenerdfreak:
I was thinking more of a huge beam cannon that fired a single shot that instantly pierced all ships it went through (no need for modifications to the engine if it was instant right?) and then had a recharge time

Tufton:
>
> I was thinking more of a huge beam cannon that fired a single shot that
> instantly pierced all ships it went through (no need for modifications to
> the engine if it was instant right?) and then had a recharge time


Lux Ult

fcfort:
How does this piercing rail gun make sense within the game? Does it only pierce if shields are down? Is it going right through all the shields and hull of every ship it pierces? Does it destroy every ship it pierces? How is it not destroying all these ships when it leaves holes in all of them?

thenerdfreak:
@fcfort I was thinking ignored shields but did decent hull damage, and imparts a lot of heats (Think particles moving through the ship, not something blasting a big hole in it)

Pointedstick:
That would be a ridiculously unbalanced no-brainer of a weapon.

thenerdfreak:
It's downside would be extremely slow fire rate and moderately low accuracy

thenerdfreak:
And to further nerf it, 120 pts of weapon space :p

brick96:
A big weapon, on a big ship, perfect for large targets but weak against smaller ships. The particle cannons are alot like rail guns now that I think about it

frameworld:
[...]

Issue #4225, opened 2019-03-10
Velocity at high ranges causes the game to crash


Variance-Moment:
**Describe the bug**
When firing a weapon with a velocity above 3200 or so the game crashes.

**To Reproduce**
1. Equip a weapon with a velocity above 3200
2. Fire it
3. Game crashes

**Expected behavior**
The weapon to fire without consequence.

**System (please complete the following information):**
 - OS: Windows 10 Home, version 1803 build 17134.590
 - Game Source: nightly
 - Version: nightly, commit hash 1cb4da159165aa9cdb70677f7d4706c43af3b889

jafdy:
Do you mean range or velocity? They aren't the same thing, plus range is a product of velocity and lifetime so is unlikely to cause a crash (although it might still be correlated).

Variance-Moment:
> 
> 
> Do you mean range or velocity? They aren't the same thing, plus range is a product of velocity and lifetime so is unlikely to cause a crash (although it might still be correlated).

I meant velocity, sorry.

Amazinite:
Are you sure this is what is causing the crash? Could you paste the data of the weapon you think is causing this crash? I just set the beam laser to fire at a velocity of 3200 and I didn't get a crash.

Amazinite:
Doubled the velocity to 6400 and my game crashed. If you are using a plugin weapon with a sprite that spans that 3200 range, perhaps this has something to do with drawing the projectile at such a distance, not the velocity itself.

Velocity | Crash
---------|------
3200 | No
6400 | Yes
4800 | Yes
4000 | No
4400 | Yes, BUT only when firing up or down, not left or right.

Yeah, I think it has something to do with drawing the projectile.

Amazinite:
So it seems like we have some sort of edge case at 2^12 velocity. @Variance-Moment Can you replicate my results, with 3200 specifically not crashing the game, but anything seemingly higher than 2^12 giving a clear crash? To note, my laptop screen is 1920x1080 and 15.6".

jafdy:
The physical size of the screen won't matter. You could try telling windows (or whatever OS you use) that you have a 720p screen and see if that effects it, if it is an effect then I would expect to see it crash at around 2730 (~4096*2/3). The 2^12 hypothesis seems more likely.

Variance-Moment:
I seem to get similar results:

Velocity | Result
-------- | ------
4096 | Immediate crash
4095 | Crashes after I begin to turn my ship
[...]

PR #6276, opened 2021-09-30
Allow projectiles to penetrate through ships, dealing damage multiple times

Amazinite:
**Feature:** This PR implements the feature request detailed and discussed in issue #424/part of https://github.com/EndlessSkyCommunity/endless-sky/pull/58.

## Feature Details
This PR adds a new weapon mechanic that allows projectiles to persist after impacting a ship, allowing them to either impact a single ship multiple times or impact multiple ships over the projectile's lifetime.
* `"penetration count"`: the number of hits that a projectile is capable of making before dying. For example, the default penetration count of 1 means that the projectile will die after a single hit, but a penetration of 2 means that the projectile would be able to hit the first target, damaging it, then go on to either hit the same target again on a later frame or a second target on the same frame for a total of two hits. Both ships and asteroids will consume penetrations as projectiles impact them.

Given a penetrating projectile with a low velocity, the most likely scenario for such a projectile is to impact the same ship multiple times over multiple frames. The following image shows an Energy Blaster with penetration firing at an Emerald Sword. Observe the multiple hit effects being created and the projectiles exiting out the other side of the ship, meaning that they still have available hits.
image
But firing at the ship from a different angle the projectiles no longer exit the other side of the ship, as they run out of penetrations while traveling through the ship.
image

Penetrating projectiles can also impact multiple ships on the same frame, particularly useful for low-lifetime/high-velocity projectiles such as lasers. The following image shows an Emerald Sword firing its Dragonflame through a hostile World-Ship (observe the hit effects on the World-Ship) and into a Cruiser on the other side (observe the health on the targeted Cruiser).
image

Penetration will also apply to a projectile's trigger radius, allowing it to trigger multiple times over its lifetime, although a trigger radius projectile will only ever trigger at most once per frame.

## Usage Examples
`"penetration count"` is a weapon attribute:
```
outfit "Energy Blaster"
	...
	weapon
		"shield damage" 9.6
		"hull damage" 6
		"penetration count" 10
```

## Testing Done
Tested the following after the refactor:
- [x] Ships can still be damaged.
- [x] Minables can still be damaged.
- [x] Shooting normal asteroids doesn't break the game.
- [x] Ship explosions still damage nearby ships.
- [x] Trigger radius still works.
- [x] Phasing weapons still work.
- [x] Projectiles with no defined penetration can only damage a single ship before dying.
- [x] Projectiles with a defined penetration of 0 will penetrate through an effectively infinite number of ships.
- [x] Projectiles with a defined penetration >1 will penetrate through the specified number of ships before dying.
- [x] Projectiles with multiple penetrations can impact multiple ships on the same frame.
- [x] Projectiles can penetrate through asteroids, minables, and ships all at once.
- [x] If a projectile with no penetration fires at a ship that is hiding behind an asteroid, the asteroid blocks the damage.
- [x] Projectiles with damage dropoff still work.
- [x] Lasers with damage dropoff still work.
- [x] Damage dropoff and penetration can be combined.

## Performance Impact
Non-penetrating projectiles shouldn't have any noticeable performance change.
Penetrating projectiles won't short-circuit collision detection, although this would only potentially cause issues with a large number of very fast (laser-like) projectiles.


source/Engine.cpp
```diff
@@ -1930,113 +1930,133 @@ void Engine::HandleMouseClicks()
...
-	
-	// If this "projectile" is a ship explosion, it always explodes.
-	if(!gov)
-		closestHit = 0.;
-	else if(projectile.GetWeapon().IsPhasing() && projectile.Target())
-	{
-		// "Phasing" projectiles that have a target will never hit any other ship.
-		shared_ptr<Ship> target = projectile.TargetPtr();
-		if(target)
+	while(projectile.Penetrations() >= 0 && hasHit)
```
quyykk, line 1938: You can:

- Rename `hasHit` to `allowMultipleHits` or something better.
- Turn this loop into a `do {} while` to avoid the akward `= false;` at the beginning of the loop.

source/Engine.cpp
```diff
@@ -1930,113 +1930,133 @@ void Engine::HandleMouseClicks()
...
-		}
-		
-		if(hit)
-			DoGrudge(hit, gov);
 	}
-	else if(projectile.MissileStrength())
+	
+	// If the projectile is still alive, give the anti-missile systems
+	// a chance to shoot it down.
+	if(projectile.Penetrations() >= 0 && projectile.MissileStrength())
```
quyykk, line 2058: A `Projectile::IsDead()` function (or some other name) seems appropriate. Also doesn't it make more sense to use `lifetime` instead to check for death?

Amazinite: Well a projectile with <0 penetrations or <0 lifetime is dead, so I don't see why one would be more valid than the other if they're both the case for dead projectiles.

source/Engine.cpp
Amazinite: >Rename `hasHit` to `allowMultipleHits` or something better.

`allowMultipleHits` makes less sense to me, as what we want to do is break out of the loop if there's nothing left to it, hence the `hasHit` name.

source/Engine.cpp
quyykk: I guess, but what I though was to not set penetrations to `< 0` at all and to only use `lifetime`.

Plip-Plop:
The projectiles should not penetrate a shielded ship. Isn't that the point of the shield?

Amazinite:
We're limited by the 2D perspective of the game, but despite the name, projectiles with penetration might not necessarily (in-universe) be boring through the ship they're impacting (which one would expect to deal damage directly to the hull, but without `"piercing"` that is not the case). Perhaps penetrating but non-piercing projectiles skim across the shields of a ship, not being completely disintegrated by the shields upon first impact but instead impacting it multiple times and therefore dealing damage to the shields multiple times. The `"penetration"` keyword still seems accurate though given the 2D perspective; the projectile is penetrating through the ship's collision mask while normally it would just be stopped at the mask's edge, regardless of whether that means it's boring through the ship or skimming the shields. Those are just in-universe explanations for how a penetrating projectile does what it does.

Terin:
Are projectiles being unable to pierce asteroids a technical thing? Since it'd be very strange if like, the Dragonflame Cannon could blast through multiple Kar Ik Vots, but be stopped by a dinky space rock. Especially since there are systems *very* dense with them.

Amazinite:
>Are projectiles being unable to pierce asteroids a technical thing?

No, it's a design choice. However imperfect they are at this role, asteroids are intended to be able to be used as cover, so outside of `"phasing"` weapons which are entirely intended for passing through asteroids (and other ships without damaging them), I wouldn't want any projectiles passing through asteroids to hit ships on the other side. I think it wouldn't be very straightforward to allow penetrating projectiles to go through asteroids though given the tiled nature of asteroids.

Terin:
Asteroids provide cover by taking stacks of penetration off. If a weapon is balanced around hitting multiple times for multiple instances of damage, if it hits asteroids on the way, it'll do a lot less damage.

If a weapon's balanced on just quickly piercing through and guaranteeing to hit a target, then it'll probably do less damage than a traditional projectile. Asteroids provide value by making a player consider if they should take a higher yield weapon that they can play well to make hit, or choosing a better arena, or if it'd be worth it to go for the guaranteed shots.

Just universally saying "No, asteroids are completely impermeable rocks" as part of the design takes a pretty big factor out of that consideration. On the tech end, though, I completely understand if that's a whole other undertaking. I'd just want the option for asteroid piercing to be open in a future PR.

EjoThims:
Phasing and penetration both would make a projectile that only hits the target ship, but can still hit it multiple times, yes?

Amazinite:
Yes.

yafflehk:
Asteroids are solid rock whereas a ship is essentially a series of voids separated by thin walls - It makes sense that an asteroid would stop weapons designed for use against relatively fragile ships.

Amazinite:
Turns out that since the asteroid tile size is 4096x4096, you'd need a projectile with at least that much velocity in order to even have a chance of hitting the same asteroid twice in the same frame in two different tiles, but that velocity is so high that it would cause issues (#4225) and should really never be used in the first place, so we can ignore the case of a projectile hitting the same asteroid twice in a frame, therefore greatly simplifying what I thought would need to be done with projectiles penetrating asteroids.

quyykk:
Would it make sense to allow a weapon to have unlimited penetrations as well? Maybe with `penetrations -1`?

Amazinite:
I'd thought of that, but I'm not sure why one wouldn't just provide an extremely high penetration value instead. Plug in a million penetrations and that should be effectively infinite for the purposes of gameplay.

Amazinite:
Did it anyway.

source/Projectile.cpp
```diff
@@ -268,17 +270,20 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
...
 {
-	clip = intersection;
 	distanceTraveled += dV.Length() * intersection;
 	for(const auto &it : weapon->HitEffects())
 		for(int i = 0; i < it.second; ++i)
 		{
 			visuals.emplace_back(*it.first, position + velocity * intersection, velocity, angle, hitVelocity);
 		}
-	lifetime = -100;
+	if(!weapon->UnlimitedPenetration() && --penetrations < 0)
```
quyykk, line 282: If you check `penetrations-- == 0` you don't need the additional check for unlimited penetrations.

quyykk (approved):
The improvement you mentioned during the stream is definitely better than currently so after doing that I believe this PR is ready to go. I don't have any more comments.

Recalculating the DamageProfile on every hit is not ideal, but it shouldn't be a problem.

tehhowch (changes requested):
Is it fair to say that we are basically taking the existing DoCollisions logic and executing that in a loop, once for each allowed hit of the projectile?

If so, we can do much better, with only a single call to `CollisionSet::Line`, and having that return the list of `Body *` to maybe hit, in the order they should be hit, just like `CollisionSet::Circle`. Probably you will still want to calculate the damage profile for each asteroid & ship, especially if e.g. a projectile is to decrease in effectiveness each time it hits an entity.

To handle interleaved hits between ships and asteroids in a single frame, the return value probably needs to include distance information, or we need to put all entities into the same collision set. (That will require us to remove the need for casting from `Body *` to `Ship *` by making the needed Ship methods virtual on the Body class or [introducing an intermediate class which asteroids, minables, and ships extend](https://github.com/endless-sky/endless-sky/pull/6622#issuecomment-1070902769)).

Also, I _would_ expect a projectile that enters the "same" body more than once in a frame (e.g. Model 512, tiled asteroid) to do damage / consume a hit multiple times. Masks with open spaces or multiple parts are likely something to handle in a follow-up, though.

Amazinite:
> If so, we can do much better, with only a single call to `CollisionSet::Line`, and having that return the list of Body * to maybe hit, in the order they should be hit, just like `CollisionSet::Circle`.

Would the number of remaining hits be handled by Engine::DoCollisions (e.g. get the list of all bodies that are hit and only deal damage to bodies in the list until we run out of hits) or CollisionSet::Line (e.g. bail once we've reached the max hits)? And should we keep a CollisionSet::Line function that always returns the first hit for those uses of Line outside of DoCollision that wouldn't require the ability to see if multiple ships could be hit in the same frame (e.g. CollisionSet::Line for the first hit and CollisionSet::LineList for all hits)?

Amazinite:
Webhook test.

Amazinite:
Closing as I don't plan to get to the requested refactor any time soon (in the next couple months). If anyone wants to pick this up in that time and finish it then be my guest, but I do plan to return to this PR if no one else handles it.

RisingLeaf:
I will try to pick this up

Pshy0:
Instead of a counter of hits, you could use a formula that takes the ship's max hull or shields into account. So for 2 ships of the same size, the one with stronger hull and better shields has higher chances of blocking the shot early.

Amazinite:
Retested the following:
- [x] No penetration count defined on the weapon (default count of 1)
- [x] Finite penetration count >1
- [x] "Infinite" penetration count with an explicit 0

As mentioned in https://github.com/endless-sky/endless-sky/pull/9079#issuecomment-1676436427, leaving any potential performance improvements for later.

RisingLeaf:
btw, I basically implemented what tehhowch said here:

> If so, we can do much better, with only a single call to CollisionSet::Line, and having that return the list of Body * to maybe hit, in the order they should be hit, just like CollisionSet::Circle. Probably you will still want to calculate the damage profile for each asteroid & ship, especially if e.g. a projectile is to decrease in effectiveness each time it hits an entity.
> To handle interleaved hits between ships and asteroids in a single frame, the return value probably needs to include distance information

on #9079 if you want to have a look, but as you said such things can also wait for later

Amazinite:
I'll give it a look when I get the chance. If you've already done it, probably best to just mix everything together.

Amazinite (commented):
Spotted a bug in this PR and in #9079 when combining projectile penetration with damage dropoff. Also trying to combine Rising's CollisionSet changes with the projectile penetration behavior from this PR. Both issues are proving to be more difficult than I expected. This is going to require a decently large refactor of how collisions are handled. 

I'm planning to create a new Collision class that does some of the work that DoCollision is currently doing. The responsibility of DoCollision will be to take every collision that occurred for a given projectile in a given frame into a vector of Collisions sorted by increasing range at which the Collisions occurred. It will then iterate through this list of Collisions until the projectile has died or there are no more Collisions left to calculate. This will require additional refactoring of CollisionSet (closestHit becomes pointless to keep track of) and AsteroidField (AsteroidField::Collide should return all impacted bodies, and shouldn't have the responsibility of dealing damage to Minables).

source/Projectile.cpp
```diff
@@ -283,17 +285,21 @@ void Projectile::Move(vector<Visual> &visuals, vector<Projectile> &projectiles)
 
 
 // This projectile hit something. Create the explosion, if any. This also
-// marks the projectile as needing deletion.
+// marks the projectile as needing deletion if it has run out of hits.
 void Projectile::Explode(vector<Visual> &visuals, double intersection, Point hitVelocity)
 {
-	clip = intersection;
 	distanceTraveled += dV.Length() * intersection;
```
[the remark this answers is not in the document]
Hurleveur: I may be misunderstanding but cant we just reset the distance once we're done with a target? (Or like just distanceTraveled = ...)

If that doesnt work I'd just have said dropoff and penetration on a laser are not compatible given we dont even intend to use it?
Caus if you want to combine it I dont see how it makes sense that distance affects the projectile but going trough a ship doesnt buy anyway...

Amazinite:
Requesting testers to make sure I didn't break the game in a horrible manner.

source/Projectile.cpp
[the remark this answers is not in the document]
Amazinite: I've changed it so that Projectile::GetInfo adds the intersection range to the distance traveled when telling the DamageProfile how far the projectile went for damage dropoff calculations. This way the projectile's distanceTraveled doesn't need to be updated and the projectile doesn't need an extra variable.

source/Engine.cpp
```diff
@@ -2217,26 +2221,35 @@ void Engine::DoCollisions(Projectile &projectile)
...
 
 				// Only directly targeted ships get provoked by blast weapons.
-				int eventType = ship->TakeDamage(visuals, damage.CalculateDamage(*ship, ship == hit.get()),
+				int eventType = ship->TakeDamage(visuals, damage.CalculateDamage(*ship, ship == hit),
 					targeted ? gov : nullptr);
 				if(eventType)
 					eventQueue.emplace_back(gov, ship->shared_from_this(), eventType);
 			}
+			// Blast radius projectiles can only explode once in a frame.
+			break;
```
Amazinite, line 2230:
```suggestion
```

source/AsteroidField.h
```diff
@@ -88,6 +91,9 @@ class AsteroidField {
 
 	CollisionSet asteroidCollisions;
 	CollisionSet minableCollisions;
+
+	// Vector for returning the result of CollideAsteroids.
+	mutable std::vector<Collision> result;
```
quyykk, line 96: This doesn't seem to need to be `mutable`?

source/Collision.h
```diff
@@ -0,0 +1,52 @@
...
+
+#ifndef COLLISION_H_
+#define COLLISION_H_
+
+#include "Body.h"
+#include "CollisionType.h"
+
+#include <list>
+#include <memory>
+#include <vector>
```
quyykk, line 24: All of those are unused.

source/Collision.h
```diff
@@ -0,0 +1,52 @@
...
+
+
+
+// Represents a collision between a projectile and a ship, asteroid, or minable.
+// Contains information such as the distance
```
quyykk, line 29:
```suggestion
// Contains information such as the distance.
```

source/Collision.h
```diff
@@ -0,0 +1,52 @@
...
+class Collision {
+public:
+	// Initialize a Collision, recording the Body that was hit, the type of
+	// object that the Body is (nothing, a Ship, a Minable, or an Asteroid),
+	// and the range that the Body was hit at.
```
quyykk, line 34: That comment should probably be split up and used to describe the data members below.

source/CollisionType.h
```diff
@@ -0,0 +1,30 @@
...
+#ifndef COLLISION_TYPE_H_
+#define COLLISION_TYPE_H_
+
+// Types of objects that projectiles are able to collide with.
+// Each CollisionSet has a CollisionType that it keeps track of.
+enum class CollisionType : int {
+	NONE,
+	SHIP,
+	MINABLE,
+	ASTEROID
```
quyykk, line 25:
```suggestion
	ASTEROID,
```

source/AsteroidField.h
Amazinite: ¯\\_(ツ)\_/¯ I copied what CollisionSet did.

source/Collision.h
Amazinite: Ah, yeah. I was using them before but didn't clean them up when I changed how Collision worked.

source/Collision.h
```diff
@@ -0,0 +1,52 @@
...
+
+	Body *HitBody();
+	CollisionType GetCollisionType() const;
+	double IntersectionRange() const;
+
+	bool operator<(const Collision &rhs) const;
+
+
+private:
+	Body *hit = nullptr;
```
RisingLeaf, line 45: Ok stupid me, we cant make this a const, right?

source/CollisionSet.cpp
```diff
@@ -164,25 +167,51 @@ void CollisionSet::Finish()
...
 
-// Check for collisions with a line, which may be a projectile's current
+// Get all collisions for the given projectile. Collisions are not necessarily
+// sorted by distance.
+const vector<Collision> &CollisionSet::LineAll(const Projectile &projectile) const
+{
+	// What objects the projectile hits depends on its government.
+	const Government *pGov = projectile.GetGovernment();
+
+	// Convert the start and end coordinates to integers.
```
RisingLeaf, line 191: This comment seems wrong? Point stores doubles iirc

source/CollisionSet.cpp
```diff
@@ -164,25 +167,51 @@ void CollisionSet::Finish()
...
+{
+	return LineAll(from, to, pGov, target, false)[0];
+}
+
+
+
+// Get all collisions along a line, given that the "all" boolean is true.
+// Otherwise, return the first (closest) collision.
+const vector<Collision> &CollisionSet::LineAll(const Point &from, const Point &to,
+		const Government *pGov, const Body *target, bool all) const
```
RisingLeaf, line 214: Couldn't we also pass an amount? So that we get only the collisions we need?

source/Collision.h
Amazinite: I'm not sure. I only didn't make it const because the original return type of CollisionSet was a non-const Body pointer.

source/CollisionSet.cpp
Amazinite: We could stop after N collisions, but how would we know those are the closest N collisions?

RisingLeaf: true I forgot

source/CollisionSet.cpp
Amazinite: Yeah, not sure what this comment meant. It's copied from line 176 (and that function could probably just call this one).

source/Collision.h
Amazinite:
```
Engine.cpp:2202:35: error: 'reinterpret_cast' from type 'const Body*' to type 'Ship*' casts away qualifiers
 2202 |                         shipHit = reinterpret_cast<Ship *>(hit)->shared_from_this();
      |                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Engine.cpp:2243:33: error: 'reinterpret_cast' from type 'const Body*' to type 'Minable*' casts away qualifiers
 2243 |                                 reinterpret_cast<Minable *>(hit)->TakeDamage(projectile);
```

RisingLeaf: Ok 👍

source/CollisionSet.cpp
Amazinite: Updated the comment.

RisingLeaf (approved):
like it, need it, NOW

source/AsteroidField.h
Amazinite: Made the collision functions const, so `mutable` is being used now.

EndlessSkyLib.cbp
```diff
@@ -96,14 +96,18 @@
...
 		<Unit filename="source/CollisionSet.cpp" />
 		<Unit filename="source/CollisionSet.h" />
+		<Unit filename="source/CollisionType.h" />
 		<Unit filename="source/Color.cpp" />
 		<Unit filename="source/Color.h" />
 		<Unit filename="source/Command.cpp" />
 		<Unit filename="source/Command.h" />
 		<Unit filename="source/ConditionSet.cpp" />
 		<Unit filename="source/ConditionSet.h" />
+		<Unit filename="source/ConditionType.h" />
```
Amazinite, line 110:
```suggestion
```

source/CollisionType.h
```diff
@@ -0,0 +1,30 @@
...
+this program. If not, see <https://www.gnu.org/licenses/>.
+*/
+
+#ifndef COLLISION_TYPE_H_
+#define COLLISION_TYPE_H_
+
+// Types of objects that projectiles are able to collide with.
+// Each CollisionSet has a CollisionType that it keeps track of.
+enum class CollisionType : int {
+	NONE,
```
Amazinite, line 22:
```suggestion
	// The NONE type represents ship explosions and projectiles
	// tripped by their trigger radius.
	NONE,
```

source/CollisionSet.cpp
```diff
@@ -164,25 +166,26 @@ void CollisionSet::Finish()
...
 {
 	// What objects the projectile hits depends on its government.
 	const Government *pGov = projectile.GetGovernment();
 
-	// Convert the start and end coordinates to integers.
+	// Convert the projectile to a line represented by its start and end points.
 	Point from = projectile.Position();
 	Point to = from + projectile.Velocity();
-	return Line(from, to, closestHit, pGov, projectile.Target());
+	return Line(from, to, pGov, projectile.Target(), projectile.HitsRemaining() == 1);
```
Amazinite, line 180:
```suggestion
	return Line(from, to, pGov, projectile.Target(), projectile.HitsRemaining() != 1);
```

PR #6278, opened 2021-10-02
refactor(internal): Replace public inheritance with composition in MissionAction refactor

tehhowch:
## Feature Details
Public inheritance should really only be used when the derived classes are to be used via the base class interface. #5157 added public inheritance (_object A_ IS-A _object B_) between MissionAction and the new "GameAction" class solely for code reuse, when composition (_object A_ HAS-A _object B_) is the preferred approach.

Inheritance also enabled atypical access by other classes (e.g. Mission) to reach inside of MissionAction and consume internals that it has no business knowing about (the full GameAction interface). No code actually did this, but the encapsulation was broken, which made this possible to do more easily.

To additionally distinguish the new capabilities in conversations, the keyword "action" is permitted wherever "apply" was permitted. For legacy support, "apply" is still permitted. It's possible we may want to restrict "apply" nodes to their previous role, rather than allowing them to trigger actions, but the only real reason we might want that is to allow for condition names that would otherwise be treated by the GameAction class as some other directive. However, conditions generally consist of multiple tokens with some mathematical operator, or an extremely reduced keyword set that doesn't have future utility to the GameAction class itself (e.g. `set <single token>`), so the likelihood of a custom condition that overlaps with GameAction keyword syntax is extremely low.

## Testing Done
Game still loads.

## Performance Impact
n/a

### DIFF
--- a/source/Conversation.h
+++ b/source/Conversation.h
@@ -27,11 +27,11 @@ class Sprite;
 
 
 
-// Class representing a conversation, generally occurring when the you are asked to
+// Class representing a conversation, generally occurring when the player is asked to
 // accept or decline a mission. The conversation can take different paths depending
 // on what responses you choose, leading you to accept, decline, or (rarely) to be
 // killed. A conversation can also branch based on various condition flags that
-// are set for the player, and can also modify those flags.
+// are set for the player, or even trigger various changes to the game's state.
 class Conversation {
 public:
 	// The possible outcomes of a conversation:
@@ -64,9 +64,8 @@ public:
 	// Check if the actions in this conversation are valid.
 	std::string Validate() const;
 	
-	// Do text replacement throughout this conversation and instantiate
-	// any GameActions. This returns a new Conversation object with
-	// things like the player's name filled in.
+	// Generate a new conversation from this template, filling in any text
+	// substitutions and instantiating any actions.
 	Conversation Instantiate(std::map<std::string, std::string> &subs, int jumps = 0, int payload = 0) const;
 	
 	// The beginning of the conversation is node 0. Some nodes have choices for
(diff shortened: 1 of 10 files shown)


source/Conversation.h
```diff
@@ -75,9 +74,9 @@ class Conversation {
 	bool IsChoice(int node) const;
 	int Choices(int node) const;
 	bool IsBranch(int node) const;
-	bool IsApply(int node) const;
+	bool IsAction(int node) const;
 	const ConditionSet &Branch(int node) const;
```
Amazinite, line 78: A slight nitpick, but if Apply is now GetAction then I'd think Branch would now be GetBranch.

Amazinite (approved):
Well that's more or less what I went for originally. :)

source/Conversation.h
```diff
@@ -93,9 +92,9 @@ class Conversation {
 		// choice can be merged into what came before it, to simplify things.
 		explicit Node(bool isChoice = false) noexcept : isChoice(isChoice), canMergeOnto(!isChoice) {}
 		
-		// For applying condition changes or branching based on conditions:
-		ConditionSet conditions;
-		// For applying actions:
+		// The condition expressions that determine the next node to load.
+		ConditionSet branch;
```
petervdmeer, line 96: Another slight nitpick; The ConditionSet is only part of the branch, the labels/nodes to jump to are also part of branching, so `branchCondition` might be a better name for the variable.

petervdmeer (approved):
Looks like a good improvement. :-)

source/Conversation.h
tehhowch: You're not wrong, getters generally shouldn't have an imperative tone that makes them read like a non-getter. I tried to not stray too far into things that weren't modified by the other PR so I didn't do things like that, adding `const noexcept` to all the getters, etc.

source/Conversation.h
tehhowch: Maybe, but to me that falls more into adding the type of the variable to the name of the variable, which I generally feel is redundant.

In some future refactor, where an instantiated `Conversation` exposes only an abstract `Node` interface (rather than `IsAction`, `IsBranch`, etc.), we would have some TU-only private class `Branch` that implements the node interface. That class would better reflect your comment here, in that a branch is not just the tested expressions but also the jump destinations.

PR #6279, opened 2021-10-03
Async loading of data files.

quyykk:
## Feature Details

Right now only images and sounds are loaded in separate threads. This PR makes it so that data files are also loaded in a separate thread. The advantage is that now the game always starts immediately, no matter how many files the game needs to read. It also means that the game loads a bit quicker, since images and sounds now load in parallel to the data files, whereas previously it was first the data files and then the images and sounds.

<details>
<summary>Architecture Changes</summary>

- Introduces a `GameObjects` class for game objects (all sets, ...).
- Functions related to objects have been moved to `GameObjects` instead of being in `GameData`.
- All of `Print*` functions in `GameData` have been moved into `main.cpp`.
- Using those print functions or `-p` doesn't load any images/sounds anymore, only data.
- `GameData::Progress` is a simple accessor.
- `MenuPanel` is responsible for the various `CheckReferences` calls that `Progress` previously made.
- `MenuPanel` is also responsible for loading the player file at the start of the game.
- `SpriteQueue` gets a function so that `Progress` doesn't load anything anymore.

</details>

## UI Screenshots

Here's the screen you get if the data files, images and sounds haven't loaded completely yet. When the circle finishes, the menu appears.

async

## Testing Done

Verified that loading still works as well as reloading (without any warnings). The command line functions also work.

## Performance Impact
Potential small increase if you have lots of data files to load.


thomasballinger:
🎉 for the GameObjects instance, not a big singleton! this looks to me like progress toward @tehhowch's glorious future where data can be read from files but not yet fully loaded, allowing enabling/disabling plugins from game menus.

source/GameObjects.cpp
```diff
@@ -0,0 +1,458 @@
...
+						make_move_iterator(list.end()));
+			}
+
+			const int size = static_cast<int>(files.size()) + 1;
+			for(const auto &path : files)
+			{
+				LoadFile(path, debugMode);
+
+				auto val = progress.load(memory_order::memory_order_relaxed);
+				while(!progress.compare_exchange_weak(val, val + 1. / size));
```
tehhowch, line 128: Atomics are another feature not yet used and thus likely to be confusing to new contributors, making this worthy of an additional comment.

Further, all empty loop bodies should be noticeable, e.g. put a comment in it stating it is intentionally empty.

(Side note: am I correct in thinking the only failures to update would be due to spurious failures?)

tehhowch (changes requested):
Excellent idea, though I have some additional suggestions.

Also, the integration tests must be passing, and the Code::Blocks project needs to be updated too.

source/MenuPanel.cpp
```diff
@@ -70,6 +74,28 @@ void MenuPanel::Step()
...
+		Audio::CheckReferences();
+		// All sprites with collision masks should also have their 1x scaled versions, so create
+		// any additional scaled masks from the default one.
+		GameData::GetMaskManager().ScaleMasks();
+	}
+	else if(alpha == 1.f)
+	{
+		// If the game isn't loaded yet, upload any images to the GPU.
+		GameData::ProcessSprites();
+	}
```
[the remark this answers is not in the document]
quyykk: I actually wanted to do this anyways after this PR :) nice

source/GameObjects.h
```diff
@@ -0,0 +1,113 @@
...
+	// Apply the given change to the universe.
+	void Change(const DataNode &node);
+	// Update the neighbor lists and other information for all the systems.
+	// This must be done any time that a change creates or moves a system.
+	void UpdateSystems();
+
+	double Progress();
+
+	// Check for objects that are referred to but never defined.
+	void CheckReferences();
```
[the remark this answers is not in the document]
quyykk: yeah unfortunately.

source/GameObjects.cpp
quyykk: > (Side note: am I correct in thinking the only failures to update would be due to spurious failures?)

yes.

source/GameObjects.cpp
```diff
@@ -0,0 +1,458 @@
...
+Endless Sky is distributed in the hope that it will be useful, but WITHOUT ANY
+WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+PARTICULAR PURPOSE.  See the GNU General Public License for more details.
+*/
+
+#include "GameObjects.h"
+
+#include "Audio.h"
+#include "BatchShader.h"
+#include "Color.h"
```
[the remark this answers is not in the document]
quyykk: Yes initially, I had forward declarations in the header. But that turned out to be a bad idea since anyone including the header would need to include allll of the other headers too. So, I decided to just `#include` them directly in this header.

quyykk:
Yeah ok, I have no idea why the integration test "Saving and Reloading" is failing. All of my guesses turned out to be wrong and now I'm out of ideas. It's not even failing at the "reload" phase :/ (but at the Alpha Centauri -> Earth jump that it doesn't do for some reason).

Addressed all of your other comments though.

tehhowch (commented):
"LoadingPanel" and "LoadPanel": one of them should be renamed.

@petervdmeer mind helping out with the debugging of the integration tests?

source/UI.cpp
```diff
@@ -28,6 +28,9 @@ using namespace std;
 // of them handles it. If none do, this returns false.
 bool UI::Handle(const SDL_Event &event)
 {
+	// Handle any queued push or pop commands.
+	PushOrPop();
+
 	bool handled = false;
```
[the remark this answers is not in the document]
quyykk: No. That's the cleanest version I can think of to fix one problem with the integration tests. The problem is:

1. When `LoadingPanel` is done loading the game data, it pops itself and pushes a new `MenuPanel`.
2. Later, we reach `Test::Step`, which executes the first step of the test to run (since the data is loaded).
3. In the next frame, SDL events are processed and sent to the menu panels.
4. But the menu panels are empty! The `MenuPanel` that is added only gets actually pushed inside the `StepAll` (or after handling any input processing).
5. So the first input step of the Test is lost.

source/MenuPanel.cpp
```diff
@@ -123,13 +122,12 @@ void MenuPanel::Draw()
...
+	alpha -= .02f;
 	if(alpha > 0.f)
 	{
 		Angle da(6.);
 		Angle a(0.);
-		for(int i = 0; i < progress; ++i)
+		for(int i = 0; i < 60; ++i)
 		{
 			Color color(.5f * alpha, 0.f);
 			PointerShader::Draw(Point(), a.Unit(), 8.f, 20.f, 140.f * alpha, color);
```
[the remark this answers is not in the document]
quyykk: My eyes can't perceive a difference. The transition between the two panel animations is immediate. Or do you mean something else?

quyykk:
So, I found out why the integration tests fail: During the execution of the tests `PlayerInfo::conditions` is suddenly empty, which breaks any test that rely on conditions. What I find interesting is that this happens on my machine only when I use the thread sanitizer. If I don't they pass. CI always fails those test.

It looks like a race or some uninitialized read I think, but I have tried every sanitizer and not one of them report anything wrong. With my luck it's probably some obvious oversight. Still investigating...

petervdmeer:
The tests fail on my Debian Linux system with the same failures as on the Ubuntu Linux CI-tests.

The failure for "Conditions and Arithmetic and Loops" is interesting since it is actually a very simple test that just performs some assignments and a loop.
What is even more interesting is that the "Calling other functions" test passes (and that test calls the "Calling other functions" test).

source/LoadingPanel.cpp
```diff
@@ -0,0 +1,90 @@
...
+		// e.g. due to capitalization errors or other typos.
+		SpriteSet::CheckReferences();
+		Audio::CheckReferences();
+		// All sprites with collision masks should also have their 1x scaled versions, so create
+		// any additional scaled masks from the default one.
+		GameData::GetMaskManager().ScaleMasks();
+
+		GameData::FinishLoading();
+
+		player.LoadRecent();
```
petervdmeer, line 64: Is it possible that this player.LoadRecent() call happens while a Test already started? (This call clears all data in player.)

petervdmeer: I created PR https://github.com/quyykk/endless-sky/pull/1 which appears to solve the issue for the `"Conditions Arithmetic and Loops"` test. The fix is not very elegant, I hope you can come up with something better.

quyykk: You're right. I only checked to see if this is called multiple times. Thank you!

quyykk: I fixed it by given `LoadingPanel` the authority to say when the data has loaded.

petervdmeer: I also have some fixes for the other integration test in https://github.com/quyykk/endless-sky/pull/2. That PR is just for merging directly.

quyykk: Thank you

source/GameObjects.h
```diff
@@ -0,0 +1,113 @@
...
+
+
+class Sprite;
+
+
+
+// Class representing every game object.
+class GameObjects {
+public:
+	Set<Color> colors;
```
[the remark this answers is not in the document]
quyykk: If someday we use this class for tests then we'll need to have public access to it. So that's why I just made the class public so avoid a potential refactor later.

source/MenuPanel.cpp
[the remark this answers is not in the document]
quyykk: WOW how did I misunderstand your comment. I'll post the gif on discord later.

PR #6291, opened 2021-10-15
test: Add tests for selling of ships.

petervdmeer:
**Feature:** This PR implements some new tests for testing selling of ships

## Feature Details
This PR adds some new tests that test selling of ships.

## UI Screenshots
N/A

## Usage Examples
Just run the tests (in CI or in a local clone)

## Testing Done
Executed the new tests

## Performance Impact
N/A


data/tests/tests_common.txt
```diff
@@ -205,3 +583,21 @@ test "Depart"
...
+	sequence
+		# Goto shipyard
+		input
+			key "s"
+		# Sell the first ship
+		input
+			key "s"
+		input
+			key "Return"
+		# Go back to main menu
```
quyykk, line 600: Technically not the main menu, but the planet menu or whatever it is called.

data/tests/tests_ship_sale.txt
```diff
@@ -0,0 +1,49 @@
...
+		# Create/inject the savegame and load it.
+		inject "Fighters and Carriers and Escorts"
+		call "Load First Savegame"
+		call "Sell first ship"
+		call "Depart"
+		assert
+			and
+				"ships: Fighter" == 1
+				"ships: Heavy Warship" == 1
+				"ships: Medium Warship" == 1
```
quyykk, line 14: I'm used to test always test preconditions as well and it's a good practice too I believe. So that there are 2 fighters, 1 HW and 1 MW after loading the first save game. Same for the other tests below.

petervdmeer: Yes, that is a good practice, but the "update_auto_conditions" functions currently don't run directly after loading, and they also don't run after buying or selling ships. So the numbers can only be checked after departing (the numbers are not correct before departing).

data/tests/tests_common.txt
petervdmeer: Good point, I'll use planet menu.

quyykk (approved):

petervdmeer:
Approved by quyykk; merging.

PR #6303, opened 2021-10-24
Allow systems to have multiple asteroid belts

Amazinite:
**Feature:**

## Feature Details
Currently if a system defines multiple asteroid belt radii, the belt distance that is loaded last is the one that gets used for the system. This PR makes it so that multiple asteroid belts can be defined for a single system, with each asteroid belt radius being able to have a weight to it akin to the weighting of variants in a fleet, or choices in a phrase. These weights dictate the chance that an asteroid is placed with that belt radius.

The new syntax is as follows:
```
system <name>
	belt <radius> [<weight>]
```

If a system defines no belt radius, the default radius of 1500 is still used. If a belt defines no weight, a default weight of 1 is used.

## Screenshots
The following screenshots show the results of having 500 minable asteroids across various different belt configurations.

Here is a single belt radius of 500, showing how a low belt radius results in a tightly packed asteroid field.
image
Here is a single belt radius of 3000, showing how a high belt radius results in a looser packing of asteroids.
image
Here is a belt radius of 500 and a belt radius of 3000, each with a weight of 1. See how the inner belt is less dense compared to the single belt of 500, since the asteroids are being spread 50/50 across both belts.
image
Here is a belt radius of 500 with a weight of 1 and a belt radius of 3000 with a weight of 6, resulting in a roughly even density of asteroids between the inner and outer belts given that the weighting of the outer belt relative to the inner belt is equivalent to the ratio between the perimeters of the circles as defined by the belt radii.
image

## Usage Examples
The above screenshots used the following belt definitions, in order.

Single belt distances without any weightings are still possible to do.
```
belt 500
```
```
belt 3000
```
If multiple belts are defined, they are both used. If no weightings are given, a default weighting of 1 is used.
```
belt 500
belt 3000
```
Should belts be given weights, some belts will be more populated than others. The follows weights would mean that 6 out of every 7 asteroids appears in the belt radius of 3000, while the remaining 1 of 7 appears in the belt radius of 500.
```
belt 500 1
belt 3000 6
```

## Testing Done
Tested what is seen in the above screenshots and usage examples.
Also tested 10 human miner fleets in a system with three belts and the miners were still able to find and destroy minable asteroids without an issue.

## Performance Impact
N/A

### DIFF
--- a/source/System.cpp
+++ b/source/System.cpp
@@ -623,7 +623,7 @@ double System::AsteroidBeltRadius() const
 
 
 
-// Get the list of asteroid belt radii.
+// Get the list of asteroid belts.
 const WeightedList<System::Belt> &System::AsteroidBelts() const
 {
 	return belts;
--- a/source/System.h
+++ b/source/System.h
@@ -125,7 +125,7 @@ public:
 	double HabitableZone() const;
 	// Get the radius of an asteroid belt.
 	double AsteroidBeltRadius() const;
-	// Get the list of asteroid belt radii.
+	// Get the list of asteroid belts.
 	const WeightedList<Belt> &AsteroidBelts() const;
 	// Get how far ships can jump from this system.
 	double JumpRange() const; 


Amazinite:
The problem with that syntax is that it wouldn't be backwards compatible with the syntax in this PR. I was imagining something more like the following:
```
belt <radius> [<weight>]
	"eccentricity" [<min>] <max>
	"periapsis" <min> <max>
	"apoapsis" <min> <max>
```
Those are the actual attributes that the Minable class currently uses to determine the orbit.
I'm not sure right now what form the values of the periapsis and apoapsis would take; they'd either be multiples of the radius or distances of their own that the game converts into multiples of the radius.
Fiddling with those three attributes is how you'd alter the thickness of the belt; if all we had was a "thickness" attribute then I have no idea how that would convert into the values of those three attributes that actually get used.

I'm also imagining we could have the following two attributes:
```
	"offset" [<min>] <max>
	"center" <x> <y>
```
"offset" would be a random value added to the "radius". In the image I posted of eccentricity 0, I believe you'd never get anything other than that uniform circle no matter how you altered the periapsis and apoapsis. "offset" is how you'd introduce some randomness to a low eccentricity belt.
"center" would just be where the belt is centered, defaulted to 0,0. That could allow belts orbiting other objects in the system. Alternatively, belts could be defined on systems like hazards are, meaning that belts could be defined as children of objects, which would become the belt's center.

Zitchas:
OK, I can see how that would work. So long as we can actually make a belt, I'm happy.

If I'm not mistaken, that set of values could actually generate elliptical orbits, couldn't it? That's useful, too.

Amazinite:
Orbits are already elliptical.

Zitchas:
Admittedly, I don't mine much. I don't find it as fun in ES as using the asteroid miner in EVN. (And feels much slower in a "Progress per real world hour" sense).

That being said, I have never distinguished an actual orbit to any asteroid in game. So whether it is circular or elliptical, simply having visually discernible and distinct orbits is a massive improvement as far as I'm concerned.

PR #6325, opened 2021-11-07
fix(integration-test): Adjust EW Navigation test dates to avoid tutorial prompt

Terin:
As it's currently implemented, the "lost" tutorial prompt checks if you fly too far from the center of a system, based on how many days off from your starting date you are. The Ember Wastes Navigation pilot was set to fly around late 3012-3013 in #6214, which compared to the default start date, leaves you about -330 days out from the start.

While this currently doesn't affect vanilla content, something in the Ember Wastes may change in the future that would trip that tutorial prompt. According to Peter, there's no particular reason for the 3012-3013 dates, so this PR gets ahead of that. Perhaps we may want to suppress tutorial missions while running integration tests in the future as well, but this is a simpler fix at the moment.

### DIFF
--- a/data/tests/tests_wormhole_navigation.txt
+++ b/data/tests/tests_wormhole_navigation.txt
@@ -118,7 +118,7 @@ test-data "Ember Wastes Navigation"
 	category "savegame"
 	contents
 		pilot Bobbi Bughunter
-		date 16 12 3012
+		date 16 12 3013
 		system "Tania Australis"
 		planet Ingot
 		clearance
@@ -355,13 +355,13 @@ test "Tests - Ember Wastes Navigation"
 		# (until PR https://github.com/endless-sky/endless-sky/pull/5244 removes
 		# this need to serialize the date conditions completely).
 		#assert
-		#	year == 3012
+		#	year == 3013
 		#	month == 12
 		#	day == 16
 		call "Depart"
 		# Give jump command.
 		assert
-			year == 3012
+			year == 3013
 			month == 12
 			day == 17
 		input
@@ -371,7 +371,7 @@ test "Tests - Ember Wastes Navigation"
 		branch notMora1
 			not "flagship system: Mora"
 		assert
-			year == 3012
+			year == 3013
 			month == 12
 			day == 18
 		watchdog 12000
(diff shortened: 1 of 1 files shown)


thomasballinger:
Sounds fine to me. FYI there's a preferences file that prevents these at https://github.com/endless-sky/endless-sky/blob/master/tests/config/preferences.txt, guess it's not being used for integration tests?

Terin:
The "Lost" tutorials don't use the standard tutorial interface. Not sure if there's a technical reason for that, but maybe that could be changed if it's a simple fix.

thomasballinger:
Hm, it looks like there are a couple of these: (basics 1 and basics 2) and the lost messages. The basics 1 and 2 are only shown as messages, so that's pretty different behavior.

https://github.com/endless-sky/endless-sky/blob/a0b4ef43f72a5c3b8e13766b28198074250d3524/source/Engine.cpp#L1292-L1299
    
The other is this batch of lost messages, which I don't understand. Since they're still only offered to players on their first four days, it seems ok to me to [make them use the normal `doHelp()`](https://github.com/endless-sky/endless-sky/pull/6328). There are two changes in behavior here:

- if you read these messages, thing they're pretty funny, then jump to another system and try to read them again (assuming still within the first 4 days of a new pilot) you'll no longer be able to
- the game no longer advances a single frame before pausing because a dialog is up

Either solution is fine with me, if there are any issues with https://github.com/endless-sky/endless-sky/pull/6328 then this seems fine.

Terin:
Yeah, properly handling the tutorials should prevent it from breaking. Having the test cases use dates we actually have in game should probably also be done as well, though, unless it is explicitly testing that.

petervdmeer:
> Sounds fine to me. FYI there's a preferences file that prevents these at https://github.com/endless-sky/endless-sky/blob/master/tests/config/preferences.txt, guess it's not being used for integration tests?

This preferences file was created specifically for the integration tests and is used by them.

petervdmeer (approved):

Terin:
Any further action to take with this? Just a date change to put it in-line with normal ES dates, unless we want this want to explicitly use off-standard dates.

PR #4484, opened 2019-08-21
Displayed ship info refinements

stechio:
First of all, thanks for this little gem of a game! It nostalgically reminds me of an analogous (yet humbler) project I developed in the 90s as a boy -- despite it featured no fancy OpenGL nor polished hull renderings, its physics and AI were somewhat comparable (2d vector motion, collision detection, shielding, targeting, fleet handling, maps, viewport, ... I spotted it on ubuntu repo yesterday and got it a try (shamelessly unlocking its full set by tweaking the all-content plugin ;-) ).

Here I'm proposing a dead-simple **improvement to the readability of the displayed ship info** (see comparison screenshot below): the current implementation looks a bit convoluted and inconsistent (all those slashes don't convey uniform semantics, label terminology is a bit fuzzy and unnecessarily verbose, unqualified measurement numbers are indistinguishable from discrete ones, etc.).

This PR sorts the things out according to these principles:

- **main values** must stand out in relation to ancillary info;
- **measurements**, whenever convenient, must be explictly qualified with their respective units;
- **nautical terminology** must be preferred over generic (mass without cargo -> displacement, cargo -> deadweight) to improve its expressiveness (hovering tooltip text is there for clarification).

I also added the model name on top.

endless-sky_feat_shipinfo-refine

PR #6326, opened 2021-11-07
Displayed ship info refinements from @stechio.

quyykk:
**Feature:** This PR is a continuation from #4484 from @stechio.

## Feature Details
This PR improves the attribute list for ships by changing how they are displayed. See the screenshots for comparisons. All credit goes to @stechio.

## UI Screenshots

Left is continuous, right is this PR.

pelican-shipyard-master pelican-shipyard-stechio


Zitchas:
All in all, this looks good.

The one detail that I'm not fond of is the change from "Required Crew/Bunks" to "Crew (min-max)".

I know many people are capture focused, but to many of us (and, arguably, to the lore) bunks aren't "places to put crew that might happen to be empty so you can put a passenger there", but rather "places to put passengers that you can put a soldier if you want."

The difference is slight, but given that there is no mechanical benefit to having additional crew (aside from capturing), I think it is more accurate to say that bunks are things that hold people that you just happen to be able to fill with crew.

Basically, I prefer the look of having it list "Here's how many required crew" and "here's the max bunks". It makes more sense to me.

tehhowch (changes requested):
Please ensure the commit is properly attributed with a co-author trailer. One way to do that would be to squash the original work into a single commit and then commit the merge & additional feedback resolution

quyykk:
Shoot I pushed the wrong branch. Anyways, I squashed their commits into one and then merged into master. Now it says for that commit "stechio authored and quyykk committed". Is that okay?

Amazinite:
I'd just do what you had just before you squashed it: maintain the commit history of the old PR. That's what I did for #5971 when picking up #3502.

quyykk:
There you go :)

quyykk:
> The one detail that I'm not fond of is the change from "Required Crew/Bunks" to "Crew (min-max)".

I agree. How about `crew (required - total)`? `crew (required - free)`? `crew (required - max)`? I like the first one.

Zitchas:
> > The one detail that I'm not fond of is the change from "Required Crew/Bunks" to "Crew (min-max)".
> 
> I agree. How about `crew (required - total)` or `crew (required - max)`?

I don't see either of those as really being changes. It's not so much the format that I'm concerned about, but rather the "Bunks" reference. We have outfits that talk about bunks, we have missions and jobs that talk about needing x bunks, we have `<bunks>` as a replacement text. I think as a matter of consistency and clarity to the player we should continue to refer to the maximum number of people onboard as bunks.

quyykk:
Well, what would you suggest? To keep with the new formatting `bunks (required - total)`?

Zitchas:
Just looking at the pictures, and it strikes me that the most appropriate format could well be that of the other space related things:

`bunks free    31/75`

After all, the number of required crew is irrelevant, right? What players are really concerned about is generally one of two values:

1. How many free bunks to put passengers in, or
2. What is the total number of bunks to attack an enemy ship with.

quyykk:
It now looks like this, as proposed by Zitchas. The relevant change is the "bunks free" line. It takes some getting used to, but I think it's an improvement.

stechio bunks free

data/tooltips.txt
```diff
@@ -329,8 +329,8 @@ tip "radar jamming:"
 tip "ramscoop:"
 	`Replenishes fuel by harvesting the stellar wind. Ramscoops are more effective when closer to a star. If you add more than one ramscoop, each additional one is less effective: you need four ramscoops to achieve double the recharge rate that one ramscoop provides.`
 
-tip "required crew / bunks:"
-	`How many crew members are required to operate this ship, and how many total bunks are available. Extra bunks can be used for passengers or extra crew.`
+tip "bunks free:"
```
Amazinite, line 332: Not too keen on this change. While it does flip the order to display the number of bunks free, which is more useful for running jobs than the number of required crew, this now doesn't communicate why some bunks are taken up. I think we should just leave this line how it was, displaying the number of required crew and the total number of bunks.

Amazinite (changes requested):

Issue #6329, opened 2021-11-08
Analyzing performance bottlenecks


quyykk:
If you have lots of ships (on my laptop ~400), the game starts to noticeably slow down. Even though the game isn't designed to be played with massive fleets, doing it anyways can uncover performance bottlenecks. But more importantly, having 400 ships in a given system (because of a massive battle for example) has the exact same performance problems. 

So that is exactly what I did. With a fleet of 460 Albatross' armed with only a single gun, I tributed Prime. This does mean any auto aiming turret code isn't included in this benchmark.

For profiling I used [gperftools](https://github.com/gperftools/gperftools), for ES compiled with LTO, `-g` and no optimizations. If I turn on optimizations the profiling data is borderline useless. So, keep this in mind while reading the list below.

Below you'll find a list of possible performance issues that I found that should be investigated. Some are probably not really relevant or fixable, since I didn't compile with optimizations. So, keep this in mind.

1. `AI::GetShipsList` return value's destructor took a whole 2% of execution time because of all the shared pointers being destroyed. And that was only the call in `AI::AutoFire`.
2. 11.5% of execution time is spent on string comparisons in `Outfit::Get`. A lot of `Ship` functions that should be very fast (like `Ship::IsDisabled`) take a lot of time solely because of string comparisons.
3. `Point::operator-` is at 5.9%. A bit surprising. `Point::Point` at 1.8% and `Point::Cross` at 2.2%.
4. `CollisionSet::Line` (26.1%) spends a lot of time inserting into a set and calling `Mask::Collide`.
5. `Ship::IsTargetable` is at 2.2% only due to `AI::GetShipsList`.
6. `BatchDrawList::AddVisual` is at 3.5% with most of that time spent allocating a vector.
7. `Ship::GetSystem` is at 4.1% because of `AI::DoScatter`.
8. `Mask::Collide` in total takes 15.6%. Most of that time is spent in `Mask::Intersection`.
9. `AI::DoScatter` is at 11.7%. At first glance this seems a bit much.

I probably missed some problems and obviously I didn't test every performance intensive scenario. But I'll continue searching and will add any other possible performance bottleneck I find to this issue.

### Steps to Reproduce

1. Download [gperftools](https://github.com/gperftools/gperftools) (Unix only sorry).
2. Modify the `SConstruct` file to link the executable with `-lprofiler` and compile with `-O0`.
[...]

PR #6330, opened 2021-11-08
Replace some unnecessary shared_ptrs with regular pointers in AI.

quyykk:
**Feature:** This PR fixes a performance bottleneck found in #6329.

## Feature Details

`AI::GetShipsList` returned a vector of `shared_ptr` which is unnecessary since there is no reason why a function that returns ships based on a filter should take ownership of the ships in question.

`AI::governmentRosters`, `AI::enemyLists` and `AI::allyLists` were also changed to use a simple `Ship *` instead of a `shared_ptr` since for them it is also unnecessary to incur the cost of ownership.

I'm sure there are still more cases of `AI` using `shared_ptr`s unnecessarily, but I didn't want to change too many things. And since only the fixes in this PR appeared on the profiling data, it seems like modifying the rest is at this point in time premature. but idk

## Performance Impact

Improvement. For N ships in the system, this PR saves the cost of increasing and decreasing at least 2N shared pointers per frame.

### DIFF
--- a/source/AI.cpp
+++ b/source/AI.cpp
@@ -2527,7 +2527,7 @@ bool AI::DoCloak(Ship &ship, Command &command)
 		// Otherwise, always cloak if you are in imminent danger.
 		static const double MAX_RANGE = 10000.;
 		double range = MAX_RANGE;
-		const Ship *nearestEnemy;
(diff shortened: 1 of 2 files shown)


tehhowch (approved):
Great change, thanks!

PR #5626, opened 2021-01-11
Content: "Remnant: Expanded Horizons Storms 1"

Zitchas:
**Content (Missions)**

## Summary
This adds a single mission wherein Dawn recruits the pilot to take her and a research laboratory to take measurements of an incoming ion storm.

## Details
- This PR adds one mission, that requires the player to have a laboratory onboard and have already completed Astral 2, which is the mission where First takes Dawn to do research on Sagittarius A*. This mission involves travelling out into the Ember Waste to Aescolanus, waiting around there for an ion storm to happen, taking some measurements, and then returning to the source planet.
- This PR also gives the Ion Storm hazard 0.01 hull damage as a means of reliably killing the timer ship in the event that the player happens to be cloaked or otherwise manages to evade the timer ship's weapons. This translates to a max of 45 hull damage on a ship that starts out without shields in a long intense storm, which is barely noticeable for most ships that are in the EW.

## Lore
Real stellar info:
 - Info about standard background temperature of 2.B1 degrees. 
 - Average density of the interstellar medium is F4240 particles per cubic meter 

Ember Waste Lore:
- Density of E8D4A51000 particles per cubic meter. (density of dense stellar clouds in real life)
- Background temp of those particles being 2710 degrees (which is the temperature at which Hydrogen will emit enough glow to produce the red glow)
- There is a problem of unaccounted energy coming from somewhere.

Remnant Lore:
- The Remnant track significant amounts of data about the Ember Wastes and document it to provide space forecasts for their pilots.
- The Remnant somehow got an image of the Ember Wastes as it appeared a very long time in the past, when it was significantly more energetic (the hydrogen was glowing purple / ultraviolet back then at 61A8 degrees.)
- The Remnant have determined that the ion storms are one of the means by which energy is transferred into the particles of the Ember Waste... But no word on where the energy is coming *from*. Acquiring this image is hinted at being secret and/or an accomplishment.

## Save File
This save file can be used to play through the new mission content:
 - Currently having problems attaching files. Will try an alternate method later.

## PR Checklist
 - No graphics are added during this PR.
 - ~[X] I updated the copyright attributions, or decline to claim copyright of any assets produced or modified~
 - ~[X] I uploaded the necessary image, blend, and texture assets here: {{insert link to assets}}~
 - ~[X] I created a PR to the [endless-sky-high-dpi repo](https://github.com/endless-sky/endless-sky-high-dpi) with the `@2x` versions of these art assets: {{insert PR link}}~

PR #6339, opened 2021-11-13
Fix: can not -> cannot

Zitchas:
## Fix
Replaces six instances of `can not` with `cannot`.

## Reasoning

As per both Merriam Webster and Grammerly, while both are technically correct, typically `can not` is only used when the `not` is part of a different phrase, as in the example `Now I can not only smell the pie, but I can also see it.` In pretty much all other cases `cannot` is the more correct spelling.

## References
https://www.merriam-webster.com/words-at-play/cannot-vs-can-not-is-there-a-difference
https://www.grammarly.com/blog/cannot-or-can-not

## Commentary
Pointed out by Wedge009 here: https://github.com/endless-sky/endless-sky/pull/5626#discussion_r748682390

Prompted by their comment, I took a look through all the instances of `can not` in the data files (only six of them, total), as well as checking [Merriam Webster's explanation](https://www.merriam-webster.com/words-at-play/cannot-vs-can-not-is-there-a-difference) for it. Between the Merriam definition and "what sounds good to me" I'm inclined to agree with Wedge009 and swap these all of these for `cannot`. That being said, I'm not 100% confident on all of these (the coalition one in particular), but I'm including them all for discussion.

Some other thoughts:
- Torza's `I can not get angry at you` does make a kind of sense. He can "not get angry with you." But he could. "not get angry" is a kind of phrase in and of itself. This is one where we could leave it as `can not`, but switching it to cannot sounds better.
- On the other hand `we can not allow another situation` doesn't work so well. Where Torza might or might not get angry, the Remnant are speaking in unequivocal terms that the situation *must not* occur again. For them, it's not a case of they "can" "not have it happen again", but rather a "must" not happen again. And in that sense, "cannot" has a better impact and conveyance of intent than the more ambivalent sounding "can not".

(Doing this as a PR instead of an issue to make it easier for people to see each of the instances in-context)

## Save File
n/a

### DIFF
--- a/data/coalition/coalition.txt
+++ b/data/coalition/coalition.txt
@@ -1106,7 +1106,7 @@ phrase "hostile disabled heliarch"
 		" "
 	word
 		", you are"
-		"at the mercy of, I can not believe I am"
+		"at the mercy of, I cannot believe I am"
 	word
 		"."
 		"!"
--- a/data/human/human missions.txt
+++ b/data/human/human missions.txt
@@ -2460,7 +2460,7 @@ mission "Paradise Fortune 4"
 			
 			label conversation
 			`	You meet with Alondo at his office in the Free Worlds Senate building. After he and Diana exchange greetings, you explain the situation to him.`
-			`	"You what?" Alondo responds. "You have to turn her in. The war may be over, Captain, but there is still tension between the Republic and Free Worlds that we can not have a situation like this exacerbating."`
+			`	"You what?" Alondo responds. "You have to turn her in. The war may be over, Captain, but there is still tension between the Republic and Free Worlds that we cannot have a situation like this exacerbating."`
 			`	"You can't do that!" Diana yells in protest. "I thought the Free Worlds was supposed to be all about going against the Republic, but you're just going to give in?"`
 			`	"Look here, kid," Alondo says in a stern tone. "The Free Worlds was never about being against the Republic, it was only ever about being against Parliament. There's a big difference there. What you've done is a crime, even to the Free Worlds. We are not about to gain a reputation of harboring criminals of the Republic."`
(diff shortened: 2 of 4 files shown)


Ferociousfeind:
I would leave "can not" alone if it were an example of Remnant speech- are they not already boycotting other contractions? They speak slowly and steadily, with a measured cadence, something which "can not" would reinforce in my opinion.

Cannot is a step in the direction of "can't", which is stylistically forbidden from Remnant dialog, isn't it? Even in the case of "it can not happen again", there may be stress on the "not", to make it obvious that the most important part is _voluntarily preventing_ it from happening again.

Wedge009:
Thank you, I appreciate it.

> I would leave "can not" alone if it were an example of Remnant speech- are they not already boycotting other contractions? They speak slowly and steadily, with a measured cadence, something which "can not" would reinforce in my opinion.

As I alluded to in #5626, the idea of trying to convey speech patterns through something as subtle as unconventional spacing is something we also discussed regarding the drake species (reptilian creatures often mistaken for dragons) in The Battle for Wesnoth. I argued, as I do here, that such subtle cues would be too easily confused for a spelling error by the player (as I have) rather than conveying an actual speech characteristic. So I consider that favouring conventional 'cannot' would be an instance of real-world consideration over game/lore subtleties for the sake of avoiding confusion.

I believe we worked out a solution where, since the drakes are characterised by archaic speech patterns, words are reordered to avoid putting 'can' and 'not together in succession. In a similar vein to the Heliarch, I think.

Regardless, _if_ you were to go down that route of formalising Remnant speech with 'can not' instead of the more conventional 'cannot', you would need to change seven instances of 'cannot' as spoken by Remnant characters in `remnant missions.txt`.

PS I just noticed that the commit message seems to be back to front... but no matter, as long as the contents achieved the intention. (:

PR #6343, opened 2021-11-15
feat(content): Additonal Disabled Hails

LepRyot:
A handful of hostile hails that use the shakespearean insults toward the beginning rather than at the end. Plus one more to the existing list
Also some new disabled hails that are general enough for just about any human faction, but I think fit the Free Worlds slightly more so (in between the formal Navy and melancholy merchants). They're weighted to be shorter so longer responses would feel more novel

Think I made an extraneous step here

### DIFF
--- a/data/human/hails.txt
+++ b/data/human/hails.txt
@@ -3011,7 +3011,7 @@ phrase "disabled hostile"
 	word
 		" "
 	word
-		"Look what you did to my ship."
+		"Look what you did to my ship!"
 		"This is all your fault."
 		"This isn't over."
 		"You got lucky this time."
--- a/data/human/hails.txt
+++ b/data/human/hails.txt
@@ -3000,7 +3000,6 @@ phrase "hostile disabled"
 phrase "disabled hostile"
 	word
 		"You piece of"
-		"Son of a"
 	word
 		" "
 	phrase


data/human/hails.txt
```diff
@@ -2962,6 +2996,34 @@ phrase "hostile disabled"
...
+		" "
+	phrase
+		"shakespearean insults"
+	word
+		"!"
+		"."
+	word
+		" "
+	word
+		"Look what you did to my ship."
```
W1zrad, line 3014:
```suggestion
		"Look what you did to my ship!"
```
Up to you. I think this seems like something that would be exclaimed rather than stated.

W1zrad (approved):
Looks good to me! Had one small suggestion, but new phrases are always good!

data/human/hails.txt
LepRyot: Wanted to make three statements with each of the ending punctuation marks, simply as arbitrary as that. But if it helps to read better, there's no issue with what I intended

data/human/hails.txt
```diff
@@ -2962,6 +2996,34 @@ phrase "hostile disabled"
 		"."
 		"!"
 
+# Shakespearean insults.
+phrase "disabled hostile"
+	word
+		"You piece of"
+		"Son of a"
```
petervdmeer, line 3003: I feel that "Son of a" doesn't work for some the shakespearean insults, for example "Son of a bagage" or "Son of a wagtail" sounds a bit weird.

Amazinite, line 3003:
```suggestion
```

